PR #8587, opened 2022-12-16
fix(sdk): add special handling for PipelineTaskFinalStatus to get_parameter_type

### LINKED ISSUE #8586: [sdk] Can't compile component that uses PipelineTaskFinalStatus: Enum value for ParameterTypeEnum must be an int, but got <class 'NoneType'> None.
### Environment

*  KFP version: 2.0.0b8

*  KFP SDK version: 2.0.0b8

*  All dependencies version: 2.0.0b8

### Steps to reproduce

When compiling a component that uses `PipelineTaskFinalStatus`, only the given parameter type is special cased during type checking, leading to an exception:

```python
from kfp.dsl import component, PipelineTaskFinalStatus

@component
def test(status: PipelineTaskFinalStatus):
    pass

Compiler().compile(test, package_path='test.yaml')  # TypeError: Enum value for ParameterTypeEnum must be an int, but got <class 'NoneType'> None.
```

### Expected result

It should be possible to compile components that use `PipelineTaskFinalStatus` to yaml.

### Materials and Reference

---

suned:

fixes #8586 by introducing special handling for `PipelineTaskFinalStatus` everywhere

**Checklist:**

### DIFF
--- a/sdk/python/kfp/components/types/type_utils.py
+++ b/sdk/python/kfp/components/types/type_utils.py
@@ -131,7 +131,9 @@ def get_parameter_type(
     Raises:
       AttributeError: if type_name is not a string type.
     """
-
+    # Special handling for PipelineTaskFinalStatus, treat it as Dict type.
+    if is_task_final_status_type(param_type):
+        param_type = 'dict'
     if type(param_type) == type:
         type_name = param_type.__name__
     elif isinstance(param_type, dict):
@@ -193,10 +195,6 @@ def verify_type_compatibility(
     Raises:
         InconsistentTypeException if types are incompatible and TYPE_CHECK==True.
     """
-
-    # Special handling for PipelineTaskFinalStatus, treat it as Dict type.
-    if is_task_final_status_type(given_type):
-        given_type = 'Dict'
 
     types_are_compatible = False
     is_parameter = is_parameter_type(str(given_type))


connor-mccarthy:
Thanks, @suned! I think there's a legitimate use-case for this. It's somewhat of a `feat(sdk)` rather than a `fix(sdk)`, since we've never permitted compiling standalone exit tasks AFAIK (it's possible I'm missing something -- if so, please let me know).

A couple testing requests to get this merged:
- Can you add a test for the case of a compiled component shown in the issue?
- Can you add a test for the case of a compiled pipeline?
- For each, can you test that re-loading the compiled component/pipeline (`kfp.components.load_component_from_file`) and composing in another pipeline works as expected?

A good place for these tests is probably `compiler_test.py`.

suned:
@connor-mccarthy Sounds good, I'll add the test cases :fire:

> I think there's a legitimate use-case for this. It's somewhat of a `feat(sdk)` rather than a `fix(sdk)`, since we've never permitted compiling standalone exit tasks AFAIK (it's possible I'm missing something -- if so, please let me know)

I see. In our current setup (`kfp==1.8.17`) we've been relying on `kfp.v2.dsl.component` using the `output_component_file` parameter to generate component yaml, which works fine with `PipelineTaskFinalStatus` and so we've come to rely on compiling components that use it 😅 

I realize that `output_component_file` is deprecated in `kfp==2.0.0b8`, but since our setup is very much based on sharing components via yaml files, upgrading to `2.0.0` this issue has become something of a blocker for us.

connor-mccarthy:
@suned, I see. If this is supported in v1, then it is indeed a regression in v2! Thanks for catching this and explaining.

If you're able to add tests per the [comment](https://github.com/kubeflow/pipelines/pull/8587#issuecomment-1355154699) above, I'm happy to review and merge your change quickly.

connor-mccarthy:
@suned, on second thought, I have another PR (better support for optional inputs) where these tests would fit in well, since `PipelineTaskFinalStatus` falls into the category of optional inputs.

Let's merge this PR and I'll add tests and more changes on top.

connor-mccarthy:
/lgtm
/approve

Thank you, @suned

connor-mccarthy (approved):

PR #8631, opened 2022-12-27
feat(sdk): support for fan-in from `dsl.ParallelFor`

connor-mccarthy:
Enables fan-in from ParallelFor. This PR includes logic for fanning-in parameters and artifacts, but fanning-in artifacts is temporarily gated until it is supported by a KFP-conformant backend.

Sample usage:
```python
@dsl.pipeline
def math_pipeline() -> int:
     with dsl.ParallelFor([1, 2, 3]) as f:
         t = double(num=f) # 2, 4, 6, respectively

     return add(nums=dsl.Collected(t.output)).output # 12
```

**Checklist:**


sdk/python/kfp/components/for_loop.py
```diff
@@ -272,3 +272,45 @@ def _get_name_override(self, loop_arg_name: str, subvar_name: str) -> str:
             The name of this loop arg variable.
         """
         return f'{loop_arg_name}{self.SUBVAR_NAME_DELIMITER}{subvar_name}'
+
+
+class Collected(pipeline_channel.PipelineChannel):
```
connor-mccarthy, line 277: Core changes for this PR

sdk/python/kfp/compiler/compiler_utils.py
```diff
@@ -325,9 +331,149 @@ def get_inputs_for_all_groups(
...
+        if group.group_type in [
+                tasks_group.TasksGroupType.CONDITION,
+                tasks_group.TasksGroupType.EXIT_HANDLER,
+        ]:
+            raise InvalidTopologyException(
+                f'Tasks cannot depend on an upstream task inside {group.__class__.__name__} that is not a common ancestor of both tasks. Task {consumer_task_name} depends on upstream task {upstream_groups[-1]}.'
+            )
+
+
+def get_outputs_for_all_groups(
```
connor-mccarthy, line 374: Core changes for this PR

sdk/python/kfp/compiler/compiler_test.py
```diff
@@ -453,27 +454,6 @@ def my_pipeline(val: bool):
 
                 dummy_op(msg=producer_task.output)
 
-    def test_valid_data_dependency_loop(self):
```
connor-mccarthy, line 456: This is tested in numerous other places

sdk/python/kfp/components/for_loop.py
```diff
@@ -272,3 +272,45 @@ def _get_name_override(self, loop_arg_name: str, subvar_name: str) -> str:
...
+    """For collecting outputs from dsl.ParallelFor.
+
+    Args:
+        output: The output of an upstream task within a dsl.ParallelFor sub-DAG.
```
connor-mccarthy, line 281: Note that this is forward compatible with allowing >1 output to be passed by changing `__init__` to

```python
def __init__(self, output, *args):
    self.outputs = [output] + list(args)
```

sdk/python/kfp/compiler/pipeline_spec_builder.py
```diff
@@ -1483,8 +1508,14 @@ def create_pipeline_spec(
     pipeline_spec.root.CopyFrom(
         _build_component_spec_from_component_spec_structure(component_spec))
 
+    # TODO: add validation of returned outputs -- it's possible to return
+    # an output from a task in a condition group, for example, which isn't
+    # caught until submission time using Vertex SDK client
```
connor-mccarthy, line 1513: Note: This issue existed prior to this PR. The following pipeline can currently be compiled, but may not return the output `int` it promises.

```
@dsl.pipeline
def p(x: int) -> int:
    with dsl.Condition(x == 1):
        return c().output
```

sdk/python/kfp/compiler/compiler_test.py
```diff
@@ -2652,5 +2631,169 @@ def my_pipeline(x: Input[Artifact] = Artifact(
...
+        def add(nums: List[int]) -> int:
+            return sum(nums)
+
+        @dsl.pipeline
+        def my_pipeline(text: str = ''):
+            with dsl.ParallelFor([1, 2, 3]) as f:
+                with dsl.Condition(text == 'text'):
+                    t = double(num=f)
+
+            with dsl.Condition(text == 'text'):
```
chensun, line 2721: What if the condition was associated with loop parameter? How would user collect, or if they could collect them after the loop?
In real world scenarios, the condition should probably depends on loop parameters, otherwise, to implement the logic you have, it would be more nature and efficient to write the condition outside the loop:
```python
with dsl.Condition(text == 'text'):
    with dsl.ParallelFor([1, 2, 3]) as f:
        t = double(num=f)
```

sdk/python/kfp/compiler/compiler_test.py
```diff
@@ -2652,5 +2631,169 @@ def my_pipeline(x: Input[Artifact] = Artifact(
...
+                r'Tasks cannot depend on an upstream task inside Condition that is not a common ancestor of both tasks\. Task add depends on upstream task double\.'
+        ):
+
+            @dsl.pipeline
+            def my_pipeline(text: str = ''):
+                with dsl.Condition(text == 'text'):
+                    with dsl.ParallelFor([1, 2, 3]) as f:
+                        t = double(num=f)
+
+                with dsl.Condition(text == 'text'):
```
chensun, line 2745: This is not the best example, in this case, why would the user ever write the second condition? It's equivalent to
```python
if a:
  foo()
if a:
  bar()
```

sdk/python/kfp/compiler/compiler_test.py
```diff
@@ -2652,5 +2631,169 @@ def my_pipeline(x: Input[Artifact] = Artifact(
...
+
+        with self.assertRaisesRegex(
+                compiler_utils.InvalidTopologyException,
+                r'Tasks cannot depend on an upstream task inside Condition that is not a common ancestor of both tasks\. Task add depends on upstream task double\.'
+        ):
+
+            @dsl.pipeline
+            def my_pipeline(text: str = ''):
+                with dsl.Condition(text == 'a'):
+                    with dsl.Condition(text == 'b'):
```
chensun, line 2766: The second condition would never hit. Again, we should be realistic on the sample code, even if it's the invalid case--it should have one and only one "error", but not multiple mistakes which could distract the focus.  
Secondly, given the indentation level of `add`, the first condition actually has nothing to do with the error, yet it could be a red herring.

sdk/python/kfp/compiler/compiler_test.py
```diff
@@ -2652,5 +2631,169 @@ def my_pipeline(x: Input[Artifact] = Artifact(
...
+
+        with self.assertRaisesRegex(
+                compiler_utils.InvalidTopologyException,
+                'Tasks cannot depend on an upstream task inside ExitHandler that is not a common ancestor of both tasks\. Task add depends on upstream task double\.'
+        ):
+
+            @dsl.pipeline
+            def my_pipeline(text: str = ''):
+                with dsl.ExitHandler(exit_comp()):
+                    with dsl.ParallelFor([1, 2, 3]) as f:
```
chensun, line 2793: Since you listed out combinations of exithandler-parrallelfor and condition-parallelfor, I'm curious what would be the case  for parrallelfor-parallelfor, it would be interested to see what are valid cases vs. invalid cases under nested loops.

sdk/python/kfp/compiler/compiler_utils.py
```diff
@@ -274,7 +274,13 @@ def get_inputs_for_all_groups(
                             channels_to_add.pop()
                             if not channels_to_add:
                                 break
-
+            elif isinstance(channel, for_loop.Collected):
```
chensun, line 277: From a code-readability perspective, the original if-else was channel produced by task (task_name being non-empty value) vs. channel not produced by task (task_name being empty value). Order-wise, It's better not to insert the new case in the middle of the two. And with these three cases, it's better to explicitly check each.

sdk/python/kfp/compiler/compiler_utils.py
```diff
@@ -274,7 +274,13 @@ def get_inputs_for_all_groups(
...
+                # Don't create compiler-injected inputs for fan-in.
+                # This is data passing in the opposite direction: instead of
+                # passing inputs from a parent task group to a child task
+                # group/task (what this function handles), instead surface
+                # outputs from inner tasks to one or more parent groups
```
chensun, line 282: I couldn't think through this clearly. If we bypassing the injection, would the following case work? 
```python
with dsl.ParallelFor([1, 2, 3]) as f:
    t = double(f)
with dsl.Condition(...):
    add(nums=dsl.Collected(t.output))
```
What would be the "punching-hole" input shown in the condition DAG component spec?

sdk/python/kfp/compiler/compiler_test.py
```diff
@@ -2652,5 +2631,169 @@ def my_pipeline(x: Input[Artifact] = Artifact(
...
+        def double(num: int) -> int:
+            return 2 * num
+
+        @dsl.component
+        def add(nums: int) -> int:
+            return nums
+
+        with self.assertRaisesRegex(
+                RuntimeError,
+                r'Tasks cannot depend on an upstream task inside a ParallelFor that is not a common ancestor of both tasks\. Task add depends on upstream task double\. Did you mean to use dsl\.Collected to fan-in outputs from a ParallelFor sub-DAG?'
```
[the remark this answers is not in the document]
connor-mccarthy: I agree, this can be clearer and more descriptive for users.

sdk/python/kfp/compiler/compiler_test.py
```diff
@@ -2652,5 +2631,169 @@ def my_pipeline(x: Input[Artifact] = Artifact(
...
+
+        @dsl.component
+        def double(num: int) -> int:
+            return 2 * num
+
+        @dsl.component
+        def add(nums: List[int]) -> int:
+            return sum(nums)
+
+        # this error will only be surfaced if the annotation is correct,
```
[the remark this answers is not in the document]
connor-mccarthy: Agreed. This comment doesn't add much. Removed.

sdk/python/kfp/compiler/compiler_utils.py
connor-mccarthy: Good catch. This is unnecessary logic. This was holdover code from a previous iteration of my work on this PR. The "hole punching" is done via the `if channel.task_name` condition.

I've removed this. [This test case](https://github.com/kubeflow/pipelines/blob/f51729ef1665eeafe35aa1fbd92529a9dafca833/sdk/python/test_data/pipelines/parallelfor_fan_in/parameters_with_conditional_consumer.py) verifies that compilation works correctly for the case you are describing.

sdk/python/kfp/compiler/compiler_utils.py
connor-mccarthy: Makes sense, though no longer relevant per [other comment](https://github.com/kubeflow/pipelines/pull/8631#discussion_r1092326763).

sdk/python/kfp/compiler/compiler_test.py
connor-mccarthy: I believe [this test case](https://github.com/kubeflow/pipelines/blob/f51729ef1665eeafe35aa1fbd92529a9dafca833/sdk/python/test_data/pipelines/parallelfor_fan_in/parameters_with_conditional_consumer.py#L20) covers the pipeline logic you're desribing.

sdk/python/kfp/compiler/compiler_test.py
connor-mccarthy: Updated to be more realistic.

sdk/python/kfp/compiler/compiler_test.py
connor-mccarthy: I've added test cases for this for completeness in the `compiler_test.py` file.

The actual YAML for these cases is tested extensively in the `read_write_test.py` tests.

sdk/python/kfp/compiler/compiler_utils.py
```diff
@@ -325,9 +326,191 @@ def get_inputs_for_all_groups(
...
+        group = group_name_to_group[group_name]
+        if group.group_type == tasks_group.TasksGroupType.FOR_LOOP:
+            break
+        if group.group_type in [
+                tasks_group.TasksGroupType.CONDITION,
+                tasks_group.TasksGroupType.EXIT_HANDLER,
+        ]:
+            raise InvalidTopologyException(
+                f'{ILLEGAL_CROSS_DAG_ERROR_PREFIX} When using dsl.{for_loop.Collected.__name__} to fan-in outputs from a task within a dsl.{tasks_group.ParallelFor.__name__} context, the dsl.{tasks_group.ParallelFor.__name__} context manager cannot be nested within a dsl.{group.__class__.__name__} context manager unless the consumer task is too. Task {consumer_task_name} consumes from {upstream_groups[-1]} within a dsl.{group.__class__.__name__} context.'
+            )
```
chensun, line 366: IIUC, there's no other type of group we could possible meet here, right? In that case, we should be explicit on raising unexpected type error for the final else. 
That also means we would never need to look at `upstream_groups[1]` and beyond. Should we be explicit and pass in only what we need, which is always `upstream_groups[0]` or null?

sdk/python/test_data/pipelines/parallelfor_fan_in/nested_with_artifacts.py
```diff
@@ -0,0 +1,37 @@
+from typing import List
```
chensun, line 1: I would add these only at the time we support fan-in artifacts. At this moment, this are dead code that serves no real purpose.

sdk/python/test_data/test_data_config.yaml
```diff
@@ -144,6 +144,26 @@ pipelines:
...
+      name: math_pipeline
+      execute: false
+    # TODO: uncomment to support artifact fan-in
+    # - module: parallelfor_fan_in/simple_with_artifacts
+    #   name: math_pipeline
+    #   execute: false
+    # TODO: uncomment to support artifact fan-in
+    # - module: parallelfor_fan_in/return_collected_artifacts_from_nested
+    #   name: math_pipeline
+    #   execute: false
```
chensun, line 163: Remove until needed. The commented section is also incomplete--it doesn't reference `nested_with_artifacts` which presents in the PR, but references `return_collected_artifacts_from_nested` which doesn't present in the PR.

sdk/python/kfp/compiler/compiler_utils.py
```diff
@@ -325,9 +326,191 @@ def get_inputs_for_all_groups(
...
+
+    # unlike inputs, which will be surfaced as component input parameters,
+    # consumers of surfaced outputs need to have a reference to what the parent
+    # component calls them when they surface them, which will be different than
+    # the producer task name and channel name (the information contained in the
+    # pipeline channel)
+    # for this reason, we use _additional_input_name_for_pipeline_channel here
+    # to set the name of the surfaced output once
+    from kfp.compiler.pipeline_spec_builder import \
+        _additional_input_name_for_pipeline_channel
```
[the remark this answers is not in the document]
connor-mccarthy: `pipeline_spec_builder` imports `compiler_utils`, so `compiler_utils` cannot import `pipeline_spec_builder` without forming a circular import.

I agree on the prefix. I also changed the import to import the module, not the function.

I also agree that importing within a function is an anti-pattern. This is a compiler util, so I've opted to move this there and import into `pipeline_spec_builder`, resolving the circular import.

sdk/python/kfp/compiler/compiler_test.py
```diff
@@ -2652,5 +2649,204 @@ def my_pipeline(x: Input[Artifact] = Artifact(
...
+
+        @dsl.pipeline
+        def my_pipeline():
+            with dsl.ParallelFor([1, 2, 3]) as v1:
+                with dsl.ParallelFor([1, 2, 3]) as v2:
+                    t = add_two_ints(num1=v1, num2=v2)
+
+                x = add(nums=dsl.Collected(t.output))
+
+
```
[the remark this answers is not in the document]
connor-mccarthy: It does. Right now it results in a fairly ambiguous error:
```python
...
    group = group_name_to_group[group_name]
KeyError: 'double'
```

I've added better topology validation and error messages for this case, as well as two test cases.

sdk/python/kfp/compiler/compiler_utils.py
connor-mccarthy: Yes, that's a good catch. I added some additional validation to this function that warrants keeping the function signature and provided inputs as-is, but I've modified the logic to better reflect the cases we're looking for to align with what you're describing.

sdk/python/test_data/pipelines/parallelfor_fan_in/nested_with_artifacts.py
connor-mccarthy: Agreed. Updated.

sdk/python/test_data/test_data_config.yaml
connor-mccarthy: Agreed. Updated. (And good catch. Demonstrates the reason for not leaving the comment in this PR.)

sdk/python/kfp/compiler/compiler_utils.py
```diff
@@ -325,9 +326,191 @@ def get_inputs_for_all_groups(
...
+) -> None:
+    """Checks that a dsl.Collected object is being used results in an
+    unambiguous pipeline topology and is therefore legal.
+
+    Args:
+        consumer_task_name: The name of the consumer task.
+        upstream_groups: The names of the producer task's upstream groups, ordered from outermost group at beginning to producer task at end. This is produced by produced by _get_uncommon_ancestors.
+        group_name_to_group: Map of group name to TasksGroup, for fast lookups.
+    """
+    # illegal if the producer has a parent conditional outside of its outermost for loop, since the for loop may or may or may not be executed
```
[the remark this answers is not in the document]
connor-mccarthy: Fixed

sdk/python/kfp/compiler/compiler_utils.py
```diff
@@ -325,9 +326,191 @@ def get_inputs_for_all_groups(
...
+        group.name: [group.name for group in group.groups] +
+        [task.name for task in group.tasks] for group in all_groups
+    }
+
+    outputs = collections.defaultdict(dict)
+
+    # handle dsl.Collected consumed by tasks
+    for task in pipeline.tasks.values():
+        for channel in task.channel_inputs:
+            if isinstance(channel, for_loop.Collected):
```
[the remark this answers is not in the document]
connor-mccarthy: Accepted and done.

chensun (approved):
/lgtm
/approve

Thanks!

PR #8634, opened 2022-12-28
feat(backend): Add support for recurring run v2 api

Linchin:
Added support for recurring runs in v2 api. Added related tests too.

**Checklist:**


backend/src/apiserver/server/util.go
```diff
@@ -331,10 +378,18 @@ func validatePipelineManifest(pipelineManifest string) error {
 	return nil
 }
 
-func getPipelineVersionIdFromResourceReferences(resourceManager *resource.ResourceManager, resourceReferences []*api.ResourceReference) string {
+func validatePiplineSpec(pipelineSpec structpb.Struct) error {
```
chensun, line 381: typo: Pipeline

backend/src/apiserver/main.go
```diff
@@ -111,6 +111,7 @@ func startRpcServer(resourceManager *resource.ResourceManager) {
 	apiV1beta1.RegisterAuthServiceServer(s, server.NewAuthServer(resourceManager))
 
 	apiV2beta1.RegisterExperimentServiceServer(s, sharedExperimentServer)
+	apiV2beta1.RegisterRecurringRunServiceServer(s, server.NewRecurringRunServer(resourceManager, &server.RecurringRunServerOptions{CollectMetrics: *collectMetricsFlag}))
```
chensun, line 114: We should have a consistent pattern here. Why it was a sharedExperimentServer but two individual recurring run servers?

backend/src/apiserver/server/recurring_run_server.go
```diff
@@ -0,0 +1,344 @@
...
+	"github.com/prometheus/client_golang/prometheus/promauto"
+	"github.com/robfig/cron"
+	authorizationv1 "k8s.io/api/authorization/v1"
+)
+
+// Metric variables. Please prefix the metric names with recurring_run_server_.
+var (
+	// Used to calculate the request rate.
+	createRecurringRunRequests = promauto.NewCounter(prometheus.CounterOpts{
+		Name: "recurring_run_server_create_requests",
```
chensun, line 37: Creating new counter will make us lose track of the new APIs until we fix our monitoring side. How did we handle this with experiment service? If we were reusing the same counter, maybe keep consistency for now and separate v1 and v2 in a batch effort.

backend/src/apiserver/resource/model_converter.go
```diff
@@ -224,8 +224,15 @@ func (r *ResourceManager) ToModelJob(jobInterface interface{}, manifest string,
 		modelJob.MaxConcurrency = apiRecurringRun.MaxConcurrency
 		modelJob.NoCatchup = apiRecurringRun.NoCatchup
 		modelJob.ServiceAccount = apiRecurringRun.ServiceAccount
-		namespace, err := r.GetNamespaceFromExperimentID(apiRecurringRun.ExperimentId)
-		modelJob.Namespace = namespace
+		if apiRecurringRun.Namespace != "" {
```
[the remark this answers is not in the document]
Linchin: I think you are right. I will revert this part.

backend/src/apiserver/resource/resource_manager.go
```diff
@@ -1350,7 +1350,16 @@ func (r *ResourceManager) GetNamespaceFromExperimentID(experimentID string) (str
 	if err != nil {
 		return "", util.Wrap(err, "Failed to get namespace from experiment ID.")
 	}
-	return experiment.Namespace, nil
+	namespace := experiment.Namespace
+
+	if len(namespace) == 0 {
```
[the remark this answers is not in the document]
Linchin: This check was copied from [this method](https://github.com/kubeflow/pipelines/blob/943fb6bdff54608aecf2390eb94a680e8dad62fa/backend/src/apiserver/resource/resource_manager.go#L1388). Perhaps this was to ensure experiments always have namespace in multi-user mode?

backend/src/apiserver/server/test_util.go
```diff
@@ -1,4 +1,4 @@
-// Copyright 2018 The Kubeflow Authors
+// Copyright 2018-2022 The Kubeflow Authors
```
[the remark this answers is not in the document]
Linchin: This file indeed had some changes this year. I guess it's ok then to update the copyright year.

backend/src/apiserver/server/util.go
Linchin: Thank you, this typo has been fixed.

backend/src/apiserver/main.go
Linchin: Because in v1, the recurring run server is called job server, so there are two servers with different names here. For runs it will be the same with experiment - we will have sharedRunServer.

backend/src/apiserver/server/recurring_run_server.go
Linchin: With experiment, we indeed used the same counters. I will change the code so we use the same counter for both v1 and v2 apis for now.

backend/src/apiserver/server/test_util.go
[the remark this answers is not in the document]
chensun: If the year wasn't updated during previous change, it's fine. There's nothing needs to be fixed. Adding the current year is optional per https://opensource.google/documentation/reference/copyright#the_year, which I've referenced in the past. 

Having such a diff in PR would always incur the questions on whether it's needed and correct. Since we already spend quite some time on this trivial thing, you can keep this change this time. In the future, let's avoid such unnecessary changes.

backend/src/apiserver/server/recurring_run_server.go
Linchin: Sorry, I commented the part when removing the recurring run counter. I will remove the whole thing.

backend/src/apiserver/resource/resource_manager.go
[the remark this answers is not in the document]
Linchin: I agree that we shouldn't have duplicated logic here. I updated the v1 function here for ease of deprecation.

backend/src/apiserver/main.go
Linchin: Thinking about readability and maintenance, I think the current is more readable (because V1 and V2 methods aren't mixed up), and takes less time to deprecate v1 due to two separate files. I also think it's worthwhile to call the server "Recurring Run Server" because it is more in line with the new name. What do you think?

chensun (approved):
/lgtm
/approve

Thanks!

PR #8639, opened 2022-12-29
fix(sdk): fix deserializing v1 component yaml defaults

connor-mccarthy:
Fixes two errors related to handling of v1 component YAML defaults:
- Boolean types with defaults as strings (e.g., `"True"`/`"true"`/`"False"`/`"false"`) could be loaded in, but could not be compiled.
- Struct/Array types were loaded in, but not compiled correctly (compiled as a `google.protobuf.Value.string_value` instead of `google.protobuf.Value.struct_value` and `google.protobuf.Value.list_value`, respectively.

This PR fixes both and relocates all v1 component YAML default handling logic to the v1 component YAML deserialization code from v2 IR YAML compilation code.

**Checklist:**
- [x] The title for your pull request (PR) should follow our title convention. [Learn more about the pull request 
title convention used in this repository](https://github.com/kubeflow/pipelines/blob/master/CONTRIBUTING.md#pull-request-title-convention).

### DIFF
--- a/sdk/python/kfp/components/structures_test.py
+++ b/sdk/python/kfp/components/structures_test.py
@@ -921,5 +921,267 @@ class TestRetryPolicy(unittest.TestCase):
         self.assertEqual(retry_policy_proto.backoff_max_duration.seconds, 3600)
 
 
+class TestDeserializeV1ComponentYamlDefaults(unittest.TestCase):
+
+    def test_uppercase_T_True(self):
+        comp_text = textwrap.dedent("""\
+inputs:
+  - { name: val, type: Boolean, default: "True" }
+implementation:
+  container:
+    image: python:3.7
+    command:
+      - sh
+      - -c
+      - |
+        echo $0
+      - { inputValue: val }
+""")
+        comp = components.load_component_from_text(comp_text)
+        self.assertEqual(comp.component_spec.inputs['val'].default, True)
+        self.assertEqual(
+            comp.pipeline_spec.root.input_definitions.parameters['val']
+            .default_value.bool_value, True)
+
+    def test_lowercase_t_true(self):
+        comp_text = textwrap.dedent("""\
+inputs:
+  - { name: val, type: Boolean, default: "true" }
+implementation:
+  container:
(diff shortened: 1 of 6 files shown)


sdk/python/kfp/components/structures_test.py
```diff
@@ -921,5 +921,267 @@ def test_to_proto(self):
...
+        comp = components.load_component_from_text(comp_text)
+        self.assertEqual(comp.component_spec.inputs['val'].default, True)
+        self.assertEqual(
+            comp.pipeline_spec.root.input_definitions.parameters['val']
+            .default_value.bool_value, True)
+
+    def test_uppercase_T_True_no_quotes(self):
+        comp_text = textwrap.dedent("""\
+inputs:
+  - { name: val, type: Boolean, default: True }
```
chensun, line 969: We should remove such test cases that technically incorrect, even though we can handle them.
Users often follow examples they could find, sometimes test cases could be a source of examples.

sdk/python/kfp/components/types/type_utils.py
```diff
@@ -60,6 +62,52 @@
...
+    'bool': bool_cast_fn,
+    'boolean': bool_cast_fn,
+    'dict': try_loading_json,
+    'list': try_loading_json,
+    'jsonobject': try_loading_json,
+    'jsonarray': try_loading_json,
+}
+
+
+def deserialize_v1_component_yaml_default(type_: str, default: Any) -> Any:
```
chensun, line 96: nit: (optional) this method may deserve its own unit test.

connor-mccarthy: Done.

sdk/python/kfp/components/structures_test.py
connor-mccarthy: Done.

chensun (approved):
/lgtm
/approve

Thanks!

connor-mccarthy:
`kubeflow-pipelines-sdk-execution-tests` failures are unrelated to this PR. Manually merging.

Issue #8502, opened 2022-11-28
[feature] S3 Integration for Artifact Storage (IRSA)


ryansteakley:
## S3 Integration for Artifact Storage (IRSA)
As of Kubeflow pipelines 2.0-alpha6 release, using IRSA for artifact storage in S3 is not fully supported(details later in the doc). There has been a long-running thread with high-interest for this functionality in the[ kubeflow pipelines github](https://github.com/kubeflow/pipelines/issues/3405). These proposal aims to address it.
 
To support using IRSA with S3, we want to add the[ AWS JS SDK credentials](https://docs.aws.amazon.com/AWSJavaScriptSDK/v3/latest/modules/_aws_sdk_credential_providers.html) module to the frontend container image and update the[ aws-helper](https://github.com/kubeflow/pipelines/blob/4bb57e6723b7a5c2eb685536e5a293aea87bd3a1/frontend/server/aws-helper.ts) to use credentialProviderChain function from the AWS JS SDK.

**Please read the following [design document](https://docs.google.com/document/d/1kWbenzENgj2SE8oKshiu_MMzZuBqfZ8Iv1LQSmwJD7E/edit?usp=sharing&resourcekey=0-DxxoE1CFqJkUeSYpEBKfmg) to better understand the reasoning and use-case behind this feature**.

---
/area frontend 
Love this idea? Give it a 👍.

surajkota:
Hi @james-jwu, @zijianjoy Can you please tag Jo Li/people from frontend side to review this proposal

cc @chensun

surajkota:
@jlyaoyuli

surajkota:
Hi @gkcalat, I believe @ryansteakley is interested to contribute this functionality. Shall we go ahead with the PR?
@jlyaoyuli let us know if you have any comments

ryansteakley:
resolved by #8651 merged, thanks everyone!

PR #8651, opened 2023-01-04
feat(frontend): Implement  aws-js-sdk crendentials to support IRSA for s3

ryansteakley:

#8502 details most of the changes and reasoning in a design document.

Removing most of the aws-helper code that pertains to grabbing the credentials from the ec2 instance metadata as this code is already handled as part of the `credentialProviderChain` that is being imported here.

**Checklist:**


ryansteakley:
@james-jwu @chensun @zijianjoy

ryansteakley:
Looks like the prow job is running with node 12 installed, when it should move to using node 14? https://github.com/GoogleCloudPlatform/oss-test-infra/blob/b3128a944261abbbc3e23b7db974f95ea6c93192/prow/prowjobs/kubeflow/pipelines/kubeflow-pipelines-presubmits.yaml#L9

mvoitko (approved):

ryansteakley:
Let me fix this unit-test looks like this is expected now since the prow has no credentials and it has exhausted all options. Missed this locally since my local env must have had some lingering creds somewhere.

frontend/server/minio-helper.ts
```diff
@@ -39,22 +39,21 @@ export interface MinioClientOptionsWithOptionalSecrets extends Partial<MinioClie
 export async function createMinioClient(config: MinioClientOptionsWithOptionalSecrets) {
   if (!config.accessKey || !config.secretKey) {
     try {
-      if (await awsInstanceProfileCredentials.ok()) {
-        const credentials = await awsInstanceProfileCredentials.getCredentials();
```
zijianjoy, line 43: In the design doc, you mention that we will fail back to existing solution if the IRSA doesn't work. It doesn't look like the case in this block. Can we retain the existing solution?

frontend/server/minio-helper.ts
```diff
@@ -39,22 +39,21 @@ export interface MinioClientOptionsWithOptionalSecrets extends Partial<MinioClie
...
-        if (credentials) {
-          const {
-            AccessKeyId: accessKey,
-            SecretAccessKey: secretKey,
-            Token: sessionToken,
-          } = credentials;
-          return new MinioClient({ ...config, accessKey, secretKey, sessionToken });
-        }
-        console.error('unable to get credentials from AWS metadata store.');
+      const credentials = fromNodeProviderChain();
```
zijianjoy, line 42: Is it possible to detect the source of object storage is AWS first, and perform IRSA only when it is AWS S3? (Pattern like "arn:aws:s3:::$mybucket/*") If it is generic S3 API, then we shouldn't use IRSA solution.

ryansteakley: @zijianjoy  The code that used to be in this block was checking for AWS metadata store credentials. The fromNodeProviderChain() will try to access IRSA credentials then fallback and one of those would be AWS metadata store credentials which retains the exsiting functionality. 

If the option is preferred, then what I can do is call the existing https://github.com/kubeflow/pipelines/blob/834d966033ebb8284981e86b831c2bb086ef30ba/frontend/server/aws-helper.ts#L54 and have this IRSA code live as an if statement above the rest of the code in this method. i.e (if isS3Endpoint) ->my new code

zijianjoy: Hello Ryan, thank you for replying to my comment, Sounds good on using `isS3Endpoint` , I would also suggest adding a comment to explain that which logic area is specific to AWS, and which logic is generic S3.

frontend/server/minio-helper.ts
```diff
@@ -37,6 +38,25 @@ export interface MinioClientOptionsWithOptionalSecrets extends Partial<MinioClie
  * @param config minio client options where `accessKey` and `secretKey` are optional.
  */
 export async function createMinioClient(config: MinioClientOptionsWithOptionalSecrets) {
+  // This logic is AWS S3 specific
+  if (isS3Endpoint(config.endPoint)) {
```
surajkota, line 42: is endpoint always set when running on an AWS by default (I dont remember us setting this as of now)?
or this becomes a required field now (this would be a breaking change)?

frontend/server/aws-helper.test.ts
```diff
@@ -118,6 +118,10 @@ describe('isS3Endpoint', () => {
     expect(isS3Endpoint('s3.cn-north-1.amazonaws.com.cn')).toBe(true);
   });
 
+  it('checks a valid s3 fips GovCloud endpoint', () => {
```
surajkota, line 121: +1

Can you add one for private link endpoint or does it have a similar format?

frontend/server/minio-helper.test.ts
```diff
@@ -19,7 +19,6 @@ import { createMinioClient, isTarball, maybeTarball, getObjectStream } from './m
 import { V1beta1JobTemplateSpec } from '@kubernetes/client-node';
 
 jest.mock('minio');
-jest.mock('./aws-helper');
```
surajkota, line 22: why is this removed? (I have no context on these tests, just curious)

frontend/server/minio-helper.ts
surajkota: @zijianjoy What is an example of generic S3 which has `s3://` in the source/URL?
https://github.com/kubeflow/pipelines/blob/master/frontend/server/handlers/artifacts.ts#L33

frontend/server/minio-helper.test.ts
ryansteakley: Was removed in the initial implementation since I had removed all the previous mocks that used aws-helper. Now that am reverting those changes back this will be necessary once again good catch.

frontend/server/aws-helper.test.ts
ryansteakley: Will add one for private-link endpoint, all amazon s3 endpoints follow the same format of having s3 prefix in the string and then amazonaws.com afterwards.

frontend/server/minio-helper.ts
ryansteakley: When `createMinioClient` is called in the `s3` case it passes in the aws `artifactsConfig` which contains endpoint, that resolves to the default value of `s3.amazonaws.com` or the env variable of `AWS_S3_ENDPOINT.` I'm not sure what you mean by breaking change?

surajkota: https://github.com/kubeflow/pipelines/blob/88a9a31d86d34e61ac0cce9ed22f3455d91c5427/frontend/server/configs.ts#L128

Got it, thanks

frontend/server/minio-helper.ts
```diff
@@ -37,6 +38,25 @@ export interface MinioClientOptionsWithOptionalSecrets extends Partial<MinioClie
...
+      }
+    } catch (err) {
+      console.error('Unable to get credentials from AWS credential provider chain: ', err);
+    }
+  }
+
+  // This logic is S3 generic
   if (!config.accessKey || !config.secretKey) {
     try {
       if (await awsInstanceProfileCredentials.ok()) {
```
surajkota, line 62: This code is outdated and should be removed since there is an updated implementation for AWS S3. New implementation supports multiple ways to provide credentials including better security practices.

Better to discuss it in a separate PR

surajkota (approved):
Good work

ryansteakley:
@zijianjoy Can you take another look?

surajkota:
friendly reminder @zijianjoy since next release might be planned soon. We want to test this as part of distribution testing. Thanks

jlyaoyuli:
Hi @surajkota @ryansteakley ,
This is Joe from KFP frontend, I will review the PR as soon as possible. Thank you!

jlyaoyuli:
/lgtm
/approve
Thank you @ryansteakley and @surajkota for this implementation.

PR #8652, opened 2023-01-05
feat(frontend): Support cloning recurringRun in KFP v2

jlyaoyuli:
1. Enable cloning a recurring run with same `runtimeConfig` and `run-trigger`  from the original recurring run.
2. Enable user change `runtimeConfig` and `run-trigger` from cloned recurring run.

https://user-images.githubusercontent.com/56132941/212144640-cee6287b-5b1a-412a-bf67-f814353e6b60.mov


frontend/src/pages/NewRunV2.tsx
```diff
@@ -131,30 +143,50 @@ function NewRunV2(props: NewRunV2Props) {
...
+  const pipelineDetailsUrlfromRecurringRun = originalRecurringRunId
+    ? RoutePage.PIPELINE_DETAILS.replace(
+        ':' + RouteParams.pipelineId + '/version/:' + RouteParams.pipelineVersionId + '?',
+        '',
+      ) + urlParser.build({ [QUERY_PARAMS.cloneFromRecurringRun]: originalRecurringRunId })
+    : '';
+
+  const pipelineDetailsUrl = pipelineDetailsUrlfromRun
+    ? pipelineDetailsUrlfromRun
+    : pipelineDetailsUrlfromRecurringRun;
```
zijianjoy, line 179: NIT: These logic can be embed in line 427 by abstracting it into a function.

frontend/src/pages/NewRunSwitcher.tsx
```diff
@@ -43,7 +45,36 @@ function NewRunSwitcher(props: PageProps) {
...
+    pipelineManifestFromRecurringRun = apiRecurringRun.pipeline_spec?.pipeline_manifest;
+  }
+
+  if (getRunSuccess && apiRun) {
+    pipelineManifestFromRun = apiRun.run?.pipeline_spec?.pipeline_manifest;
+  }
+
+  let pipelineManifest = pipelineManifestFromRun
+    ? pipelineManifestFromRun
+    : pipelineManifestFromRecurringRun;
```
zijianjoy, line 77: The logics for determining `templateString` has been scattered across multiple places in this file (here and below). Suggest to put all the `API fetching` logics in one session, and combine all `templateString` related logics at the end.

frontend/src/pages/NewRunSwitcher.tsx
```diff
@@ -43,7 +45,36 @@ function NewRunSwitcher(props: PageProps) {
...
+  );
+
+  let pipelineManifestFromRecurringRun: string | undefined;
+  let pipelineManifestFromRun: string | undefined;
+
+  if (getRecurringRunSuccess && apiRecurringRun) {
+    pipelineManifestFromRecurringRun = apiRecurringRun.pipeline_spec?.pipeline_manifest;
+  }
+
+  if (getRunSuccess && apiRun) {
```
zijianjoy, line 71: For discussion: How do you think about this change?

```
let pipelineManifest = apiRun?.run?.pipeline_spec?.pipeline_manifest;
if (getRecurringRunSuccess &&   apiRecurringRun?.pipeline_spec?.pipeline_manifest) {
    pipelineManifest = apiRecurringRun.pipeline_spec.pipeline_manifest;
}
```

frontend/src/pages/NewRunV2.tsx
```diff
@@ -78,6 +78,8 @@ interface RunV2Props {
   namespace?: string;
   existingRunId: string | null;
   apiRun: ApiRunDetail | undefined;
+  originalRecurringRunId: string | null;
```
zijianjoy, line 81: Question: Can we use `originalRecurringRunId?: string` instead?

frontend/src/pages/NewRunV2.tsx
```diff
@@ -78,6 +78,8 @@ interface RunV2Props {
...
+  apiRecurringRun: ApiJob | undefined;
```
zijianjoy, line 82: I think `apiRecurringRun` and `apiRun` should be exclusive to each other, that means if one is not empty, another one should be empty. Otherwise, a lot of inner logics of this component will yield error. For example `hasVersionId()`

frontend/src/pages/NewRunV2.tsx
```diff
@@ -91,16 +93,24 @@ type NewRunV2Props = RunV2Props & PageProps;
...
     return true;
   }
   let hasVersionType: boolean = false;
-  if (apiRun.run?.resource_references) {
+  if (apiRun && apiRun.run?.resource_references) {
     apiRun.run.resource_references.forEach(value => {
       hasVersionType = hasVersionType || value.key?.type === ApiResourceType.PIPELINEVERSION;
     });
   }
+  if (apiRecurringRun && apiRecurringRun.resource_references) {
```
zijianjoy, line 109: If `apiRecurringRun` exists, can we clean up any resource_references set by `apiRun`? Otherwise resource_references are mixed and conflicted.

frontend/src/pages/NewRunV2.tsx
```diff
@@ -109,6 +119,8 @@ function NewRunV2(props: NewRunV2Props) {
   const {
     existingRunId,
     apiRun,
+    originalRecurringRunId,
+    apiRecurringRun,
```
zijianjoy, line 123: We should yield error if both run and recurring run exist.

frontend/src/pages/NewRunV2.tsx
```diff
@@ -131,30 +143,50 @@ function NewRunV2(props: NewRunV2Props) {
...
   const [errorMessage, setErrorMessage] = useState('');
   const [isParameterValid, setIsParameterValid] = useState(false);
-  const [isRecurringRun, setIsRecurringRun] = useState(false);
-  const [trigger, setTrigger] = useState<ApiTrigger>();
-  const [maxConcurrentRuns, setMaxConcurrentRuns] = useState('10');
+  const [isClone, setIsClone] = useState(false);
+  const [isRecurringRun, setIsRecurringRun] = useState(apiRecurringRun !== undefined);
+  const [trigger, setTrigger] = useState(apiRecurringRun ? apiRecurringRun.trigger : undefined);
+  const [maxConcurrentRuns, setMaxConcurrentRuns] = useState(
+    apiRecurringRun ? apiRecurringRun.max_concurrency : '10',
```
zijianjoy, line 150: We need to check if `apiRecurringRun.max_concurrency` is valid as well.

frontend/src/pages/NewRunV2.tsx
```diff
@@ -131,30 +143,50 @@ function NewRunV2(props: NewRunV2Props) {
   const [isStartingNewRun, setIsStartingNewRun] = useState(false);
   const [errorMessage, setErrorMessage] = useState('');
   const [isParameterValid, setIsParameterValid] = useState(false);
-  const [isRecurringRun, setIsRecurringRun] = useState(false);
-  const [trigger, setTrigger] = useState<ApiTrigger>();
-  const [maxConcurrentRuns, setMaxConcurrentRuns] = useState('10');
+  const [isClone, setIsClone] = useState(false);
+  const [isRecurringRun, setIsRecurringRun] = useState(apiRecurringRun !== undefined);
+  const [trigger, setTrigger] = useState(apiRecurringRun ? apiRecurringRun.trigger : undefined);
```
zijianjoy, line 148: We need to check if `apiRecurringRun.trigger` is valid as well.

frontend/src/pages/NewRunV2.tsx
```diff
@@ -131,30 +143,50 @@ function NewRunV2(props: NewRunV2Props) {
...
+  const [maxConcurrentRuns, setMaxConcurrentRuns] = useState(
+    apiRecurringRun ? apiRecurringRun.max_concurrency : '10',
+  );
   const [isMaxConcurrentRunValid, setIsMaxConcurrentRunValid] = useState(true);
-  const [catchup, setCatchup] = useState(true);
+  const initialCatchup =
+    apiRecurringRun && apiRecurringRun.no_catchup !== undefined
+      ? !apiRecurringRun.no_catchup
+      : true;
+  const [catchup, setCatchup] = useState(initialCatchup);
```
zijianjoy, line 157: NIT: replace `catchup` with `needCatchUp`

frontend/src/pages/NewRunV2.tsx
```diff
@@ -131,30 +143,50 @@ function NewRunV2(props: NewRunV2Props) {
...
+  const pipelineDetailsUrl = pipelineDetailsUrlfromRun
+    ? pipelineDetailsUrlfromRun
+    : pipelineDetailsUrlfromRecurringRun;
   const isTemplatePullSuccess = templateString ? true : false;
-  const apiResourceRefFromRun = apiRun?.run?.resource_references
-    ? apiRun.run?.resource_references
-    : undefined;
+  const apiResourceRefFromRun = apiRun?.run?.resource_references;
+  const apiResourceRefFromRecurringRun = apiRecurringRun?.resource_references;
+  const existResourceRef = apiResourceRefFromRun
```
zijianjoy, line 183: Looks like `apiResourceRefFromRun` takes higher priority than `apiResourceRefFromRecurringRun`, while it is not the case in `NewRunSwitcher` file. Let's be clear about which one takes higher priority, or even better, makes sure that `apiRun` and `apiRecurringRun` don't exist at the same time.

frontend/src/pages/NewRunV2.tsx
```diff
@@ -167,6 +199,13 @@ function NewRunV2(props: NewRunV2Props) {
     // eslint-disable-next-line react-hooks/exhaustive-deps
   }, []);
 
+  // Determine if it is cloning (run / recurring run)
+  useEffect(() => {
+    if (apiRun || apiRecurringRun) {
+      setIsClone(true);
+    }
+  }, [apiRun, apiRecurringRun]);
```
zijianjoy, line 207: If we write a function which combine `apiRun` and `apiRunRecurringRun` into one `CloneOrigin` object, which contains the information like `isRecurringRun`. Then we can simplify the main logic in `NewRunV2`.

zijianjoy (commented):
Thank you Joe for the change! I think we can improve the logic a bit by abstracting the logic of cloning from recurringRun or Run to a function. As a result, we don't need to write too many if/else conditions in the main NewRunV2 component.

frontend/src/pages/NewRunV2.tsx
jlyaoyuli: If we change to undefined, we have to add an inline logic `originalRecurringRunId={originalRecurringRunId ? originalRecurringRunId : undefined}  in `NewRunSwitcher`, because the type of returned object from urlParser.get() is `string | null`. I tend to keep the current interface definition.

frontend/src/pages/NewRunV2.tsx
jlyaoyuli: Yes, they should be exclusive to each other.

frontend/src/pages/NewRunV2.tsx
jlyaoyuli: If `apiRecurringRun` exists, `apiRun` will be empty.

frontend/src/pages/NewRunV2.tsx
```diff
@@ -131,30 +167,38 @@ function NewRunV2(props: NewRunV2Props) {
   const [isStartingNewRun, setIsStartingNewRun] = useState(false);
   const [errorMessage, setErrorMessage] = useState('');
   const [isParameterValid, setIsParameterValid] = useState(false);
-  const [isRecurringRun, setIsRecurringRun] = useState(false);
-  const [trigger, setTrigger] = useState<ApiTrigger>();
-  const [maxConcurrentRuns, setMaxConcurrentRuns] = useState('10');
+  const [isClone, setIsClone] = useState(false);
```
zijianjoy, line 170: `isClone` should not be a state from my opinion, it is because user will enter from different entrypoint if it is a clone, and they cannot select pipeline and pipeline version if it is clone. So `isClone` should be determined by the time of first rendering.

frontend/src/pages/NewRunV2.tsx
```diff
@@ -93,21 +93,33 @@ type NewRunV2Props = RunV2Props & PageProps;
...
-): boolean {
-  if (!apiRun && !apiRecurringRun) {
+type CloneOrigin = {
+  isClone: boolean;
+  isRecurring: boolean;
+  run: ApiRunDetail | undefined;
+  recurringRun: ApiJob | undefined;
+};
+
+function getCloneOrigin(apiRun: ApiRunDetail | undefined, apiRecurringRun: ApiJob | undefined) {
```
zijianjoy, line 103: NIT: We can use `function getCloneOrigin(apiRun?: ApiRunDetail, apiRecurringRun?: ApiJob)` to be concise.

frontend/src/pages/NewRunV2.tsx
```diff
@@ -167,38 +181,42 @@ function NewRunV2(props: NewRunV2Props) {
...
+    cloneOrigin.recurringRun?.no_catchup !== undefined
+      ? !cloneOrigin.recurringRun.no_catchup
       : true;
   const [needCatchup, setNeedCatchup] = useState(initialCatchup);
-  const [clonedRuntimeConfig, setClonedRuntimeConfig] = useState<PipelineSpecRuntimeConfig>({});
+
+  const initClonedRuntimeConfig = cloneOrigin.isRecurring
+    ? cloneOrigin.recurringRun?.pipeline_spec?.runtime_config
+    : cloneOrigin.run?.run?.pipeline_spec?.runtime_config;
+  const [clonedRuntimeConfig] = useState<PipelineSpecRuntimeConfig>(
```
zijianjoy, line 204: Do we need `useState` here? If not, we can remove it.

frontend/src/pages/NewRunV2.tsx
```diff
@@ -167,38 +181,42 @@ function NewRunV2(props: NewRunV2Props) {
...
-  const apiResourceRefFromRun = apiRun?.run?.resource_references;
-  const apiResourceRefFromRecurringRun = apiRecurringRun?.resource_references;
-  const existResourceRef = apiResourceRefFromRun
-    ? apiResourceRefFromRun
-    : apiResourceRefFromRecurringRun;
+  const existResourceRef = cloneOrigin.isRecurring
+    ? cloneOrigin.recurringRun?.resource_references
+    : cloneOrigin.run?.run?.resource_references;
 
   // TODO(jlyaoyuli): support cloning recurring run with query parameter from isRecurring.
```
zijianjoy, line 217: This TODO can be removed now.

frontend/src/pages/NewRunV2.tsx
```diff
@@ -279,34 +338,24 @@ function NewRunV2(props: NewRunV2Props) {
...
-            max_concurrency: maxConcurrentRuns || '1',
-            no_catchup: !catchup,
-            trigger: trigger,
-          }
-        : {
-            enabled: false,
-            max_concurrency: undefined,
-            no_catchup: undefined,
-            trigger: undefined,
-          },
```
zijianjoy, line 308: Question: Why do we remove this logic?

frontend/src/pages/NewRunV2.tsx
jlyaoyuli: Yes, we need it. Otherwise, users are not able to change the parameters.

frontend/src/pages/NewRunV2.tsx
jlyaoyuli: The declaration of `run` object and `recurring run` object is independent in v2, so we don't have to consider the one-off `run` condition here in `recurring run` definition.

zijianjoy: Thank you for the explanation, I think this change actually makes newRun to always include recurring run related fields. Look at the definition of `Object.assign()`: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/assign

jlyaoyuli: Thank you for pointing this out!

zijianjoy:
/lgtm
/approve

Thank you Joe for making this change!

PR #8656, opened 2023-01-05
chore(sdk): clean up client.py

droctothorpe:
This PR tidies up the type hints in the client module.

**Checklist:**

### DIFF
--- a/sdk/python/kfp/client/client.py
+++ b/sdk/python/kfp/client/client.py
@@ -23,7 +23,7 @@ import tarfile
 import tempfile
 import time
 from types import ModuleType
-from typing import Any, Callable, Dict, List, Optional, Union
+from typing import Any, Dict, List, Optional
 import warnings
 import zipfile
 
@@ -303,15 +303,8 @@ class Client:
                 namespace)
         return config
 
-    def _is_inverse_proxy_host(self, host):
-        if host:
-            return re.match(r'\S+.googleusercontent.com/{0,1}$', host)
-        if re.match(r'\w+', host):
-            warnings.warn(
-                f'The received host is {host}, please include the full endpoint'
-                ' address (with ".(pipelines/notebooks).googleusercontent.com")'
-            )
-        return False
+    def _is_inverse_proxy_host(self, host: str) -> bool:
+        return bool(re.match(r'\S+.googleusercontent.com/{0,1}$', host))
 
     def _is_ipython(self) -> bool:
         """Returns whether we are running in notebook."""
@@ -559,10 +552,11 @@ class Client:
             resource_reference_key_id=namespace,
             filter=filter)
 
-    def get_experiment(self,
-                       experiment_id=None,
-                       experiment_name=None,
-                       namespace=None) -> kfp_server_api.ApiExperiment:
+    def get_experiment(
+            self,
+            experiment_id: Optional[str] = None,
+            experiment_name: Optional[str] = None,
+            namespace: Optional[str] = None) -> kfp_server_api.ApiExperiment:
         """Gets details of an experiment.
 
         Either ``experiment_id`` or ``experiment_name`` is required.
@@ -1023,7 +1017,7 @@ class Client:
 
     def create_run_from_pipeline_func(
         self,
-        pipeline_func: Union[Callable[..., Any], base_component.BaseComponent],
+        pipeline_func: base_component.BaseComponent,
         arguments: Optional[Dict[str, Any]] = None,
         run_name: Optional[str] = None,
         experiment_name: Optional[str] = None,
@@ -1082,7 +1076,7 @@ class Client:
     def create_run_from_pipeline_package(
         self,
         pipeline_file: str,
-        arguments: Optional[Dict[str, str]] = None,
+        arguments: Optional[Dict[str, Any]] = None,
         run_name: Optional[str] = None,
         experiment_name: Optional[str] = None,
         namespace: Optional[str] = None,
@@ -1362,9 +1356,9 @@ class Client:
 
     def upload_pipeline(
         self,
-        pipeline_package_path: str = None,
-        pipeline_name: str = None,
-        description: str = None,
+        pipeline_package_path: str,
+        pipeline_name: Optional[str] = None,
+        description: Optional[str] = None,
     ) -> kfp_server_api.ApiPipeline:
         """Uploads a pipeline.
 
@@ -1531,7 +1525,7 @@ class Client:
 
 
 def _add_generated_apis(target_struct: Any, api_module: ModuleType,
-                        api_client: kfp_server_api.ApiClient):
+                        api_client: kfp_server_api.ApiClient) -> None:
     """Initializes a hierarchical API object based on the generated API module.
 
     PipelineServiceApi.create_pipeline becomes


sdk/python/kfp/client/client.py
```diff
@@ -1082,7 +1081,7 @@ def create_run_from_pipeline_func(
     def create_run_from_pipeline_package(
         self,
         pipeline_file: str,
-        arguments: Optional[Dict[str, str]] = None,
```
connor-mccarthy, line 1085: Is it possible to submit a run with `arguments={'number': 1}`?

connor-mccarthy (commented):
One question for you

sdk/python/kfp/client/client.py
droctothorpe: I’m able to get the function to execute even when the `arguments` dictionary includes an int. [This](https://www.kubeflow.org/docs/components/pipelines/v2/author-a-pipeline/component-io/#inputs-1) documentation indicates that non-string values are an accepted pattern.

> Valid input parameter annotations include str, int, float, bool, dict, list.

connor-mccarthy: In that case, perhaps the original annotation is correct?

droctothorpe: Also, these two functions just wrap `run_pipeline`, which uses the `Dict[str, any]` syntax in [its type hints](https://github.com/kubeflow/pipelines/blob/master/sdk/python/kfp/client/client.py#L736).

`run_pipeline` invokes `_create_job_config`, which also uses [Any](https://github.com/kubeflow/pipelines/blob/master/sdk/python/kfp/client/client.py#L736),

droctothorpe: After a few more hops, it eventually winds up at [this](https://github.com/kubeflow/pipelines/blob/master/backend/api/v1beta1/go_http_client/job_model/pipeline_spec_runtime_config.go#L21) Go struct, which has a value of `interface{}`, i.e. an empty interface, which matches _all_ types.

droctothorpe: Just to clarify and add some context, [create_run_from_pipeline_func](https://github.com/kubeflow/pipelines/blob/master/sdk/python/kfp/client/client.py#L1027) in this _same_ module uses `Dict[str, Any]` as opposed to `Dict[str, str]`. So two functions in the same module that essentially do the same thing have different type hints. If that's not intentional, then they should probably be brought into alignment, which begs the question, which one is right? Since they both wrap around functions (and ultimately types) that use `Dict[str, Any]`, it seems like that's the standard and `create_run_from_pipeline_package` just has a small discrepancy that needs to be brought into alignment. Maybe I'm missing something though. As my wife can attest, I'm wrong all the time.

connor-mccarthy: Thanks for explaining. I misunderstood :). This change LGTM!

connor-mccarthy: I believe the TODO comment is about type validation, however (i.e., make sure the values submitted are the correct types based on the pipeline signature), not about the type annotation. Can you re-add the comment?

droctothorpe: Added the comments back! Thanks, @connor-mccarthy!

connor-mccarthy: Can you also update the title since the comments aren't removed? Also, since the TODOs are unrelated, this is a fairly small PR. Are there other needed `client.py` type annotation updates that we can bundle in to this PR?

droctothorpe: Good call. Happy to tackle that.

droctothorpe: Updated about 20 type hints and modified the commit and PR title. Happy to address any other feedback.

sdk/python/kfp/client/client.py
```diff
@@ -303,7 +303,7 @@ def _load_config(
                 namespace)
         return config
 
-    def _is_inverse_proxy_host(self, host):
+    def _is_inverse_proxy_host(self, host: str):
```
connor-mccarthy, line 306: nit: return annotation

Can you scan for this throughout?

sdk/python/kfp/client/client.py
```diff
@@ -694,9 +694,9 @@ def _override_caching_options(self, pipeline_obj: dict,
 
     def list_pipelines(
         self,
-        page_token: str = '',
-        page_size: int = 10,
-        sort_by: str = '',
+        page_token: Optional[str] = '',
+        page_size: Optional[int] = 10,
+        sort_by: Optional[str] = '',
```
connor-mccarthy, line 699: `typing.Optional[<type>]` == `Union[<type>, None]` [[ref](https://docs.python.org/3/library/typing.html#typing.Optional)], which is not the same as a parameter having a default value.

Since the default is not `None` and (as far as I know) we don't permit passing `None`, the original annotation is correct.

connor-mccarthy (changes requested):
Can you also run `mypy` over this file? It's not critical to resolve every `mypy` error message,  but a before/after `mypy` output comparison is a good way to assess how much progress is made and if there is other low-hanging fruit to further reconcile inconsistencies.

sdk/python/kfp/client/client.py
droctothorpe: TIL. Thanks! I wonder if there is a way to specify optional-ness but exclude None, because it is optional from a UX standpoint, but maybe that doesn't matter?

droctothorpe:
> > Can you also run `mypy` over this file? It's not critical to resolve every `mypy` error message, but a before/after `mypy` output comparison is a good way to assess how much progress is made and if there is other low-hanging fruit to further reconcile inconsistencies.
> 
> Copy that. Thanks, Connor. Plan to tackle on Monday.

Scratch that, heh. I don't know how to stop working. Addressed feedback. A few minor questions that I'll ask alongside the relevant code.

sdk/python/kfp/client/client.py
droctothorpe: Done! 👍

sdk/python/kfp/client/client.py
```diff
@@ -1057,7 +1057,10 @@ def create_run_from_pipeline_func(
             ``RunPipelineResult`` object containing information about the pipeline run.
         """
         #TODO: Check arguments against the pipeline function
-        pipeline_name = pipeline_func.name
+        if isinstance(pipeline_func, base_component.BaseComponent):
+            pipeline_name = pipeline_func.name
+        else:
+            pipeline_name = pipeline_func.__name__
```
droctothorpe, line 1063: This one is a bit odd. The [type hint](https://github.com/kubeflow/pipelines/blob/master/sdk/python/kfp/client/client.py#L1026) specifies a union:
> pipeline_func: Union[Callable[..., Any], base_component.BaseComponent],

`BaseComponent` has a `name` attribute. A generic callable doesn't. Which is why mypy complained about this. Not sure this is the optimal solution and totally open to alternatives.

sdk/python/kfp/client/client.py
```diff
@@ -303,9 +303,9 @@ def _load_config(
                 namespace)
         return config
 
-    def _is_inverse_proxy_host(self, host):
+    def _is_inverse_proxy_host(self, host: str) -> bool:
         if host:
-            return re.match(r'\S+.googleusercontent.com/{0,1}$', host)
+            return bool(re.match(r'\S+.googleusercontent.com/{0,1}$', host))
```
droctothorpe, line 308: `re.match` does not return a bool, but it does return a truthy object if a match is found, and `None` (falsey) if a match is not found. Let me know if you want to be more explicit about this, e.g.:
```
if re.match(r'\S+.googleusercontent.com/{0,1}$', host) is None:
    return False
```

connor-mccarthy: Perhaps it makes sense to expand with an explicit `None` check, since it does not require future readers to know about the truthiness/falsiness properties of `re.Match` objects.

sdk/python/kfp/client/client.py
connor-mccarthy: In this case, I believe the annotation should be `base_component.BaseComponent` only.

The `Callable` annotation is a relic from v1, where pipelines were pure functions. In v2, pipelines created with the `@dsl.pipeline` decorator are instances of `GraphComponent`, which is a subclass of `BaseComponent`.

connor-mccarthy: In this case, I believe the annotation should be `base_component.BaseComponent` only.

The `Callable` annotation is a relic from v1, where pipelines were pure functions. In v2, pipelines created with the `@dsl.pipeline` decorator are instances of `GraphComponent`, which is a subclass of `BaseComponent`.

connor-mccarthy (changes requested):

sdk/python/kfp/client/client.py
```diff
@@ -303,14 +303,10 @@ def _load_config(
...
 
-    def _is_inverse_proxy_host(self, host):
+    def _is_inverse_proxy_host(self, host: str) -> bool:
         if host:
-            return re.match(r'\S+.googleusercontent.com/{0,1}$', host)
-        if re.match(r'\w+', host):
-            warnings.warn(
-                f'The received host is {host}, please include the full endpoint'
-                ' address (with ".(pipelines/notebooks).googleusercontent.com")'
-            )
```
droctothorpe, line 313: AFACT, this code is never reached. The only time that we reach line 309 is when `host` is `""` or `None`. Since both are falsey, this condition never evaluates as `True`. As a result, the subsequent warning on lines 310-313 is never executed.

We may want to handle `host` being `""` or `None` with a simple `else` aligned to the `if host` conditional, but the warning should probably be reworded in that case.

sdk/python/kfp/client/client.py
```diff
@@ -303,14 +303,10 @@ def _load_config(
...
+            if re.match(r'\S+.googleusercontent.com/{0,1}$', host) is not None:
```
droctothorpe, line 308: Used the negative to consolidate the `return False` statements into one, i.e. we return `False` if the regex doesn't match _or_ if the host is falsey.

sdk/python/kfp/client/client.py
```diff
@@ -1362,9 +1358,9 @@ def _get_workflow_json(self, run_id: str) -> dict:
 
     def upload_pipeline(
         self,
-        pipeline_package_path: str = None,
-        pipeline_name: str = None,
-        description: str = None,
+        pipeline_package_path: str,
```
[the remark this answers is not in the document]
droctothorpe:
```
❯ python -u "/Users/inn487/code/kubeflow/pipelines/test_args.py"
/Users/inn487/code/kubeflow/pipelines/sdk/python/kfp/client/client.py:141: FutureWarning: This client only works with Kubeflow Pipeline v2.0.0-alpha.0 and later versions.
  warnings.warn(
  File "/Users/inn487/code/kubeflow/pipelines/test_args.py", line 53, in <module>
    print(client.upload_pipeline())
  File "/Users/inn487/code/kubeflow/pipelines/sdk/python/kfp/client/client.py", line 1378, in upload_pipeline
    response = self._upload_api.upload_pipeline(
  File "/opt/homebrew/Caskroom/miniconda/base/envs/kfp-upstream/lib/python3.8/site-packages/kfp_server_api/api/pipeline_upload_service_api.py", line 69, in upload_pipeline
    return self.upload_pipeline_with_http_info(uploadfile, **kwargs)  # noqa: E501
  File "/opt/homebrew/Caskroom/miniconda/base/envs/kfp-upstream/lib/python3.8/site-packages/kfp_server_api/api/pipeline_upload_service_api.py", line 132, in upload_pipeline_with_http_info
    raise ApiValueError("Missing the required parameter `uploadfile` when calling `upload_pipeline`")  # noqa: E501
kfp_server_api.exceptions.ApiValueError: Missing the required parameter `uploadfile` when calling `upload_pipeline`
```

sdk/python/kfp/client/client.py
connor-mccarthy: Casting `re.Match` or `None` to a boolean and `-> bool` SGTM.

connor-mccarthy (approved):
/lgtm
/approve

Thanks for this, @droctothorpe!

connor-mccarthy:
Manually merging, as `kubeflow-pipelines-sdk-execution-tests` failures are unrelated to these changes.

PR #8681, opened 2023-01-16
fix(frontend): Show pipeline details for an existing recurring run without pipeline / pipeline version.

jlyaoyuli:
Enable pipeline details logic to retrieve `template string` from pipeline manifest in recurring run and then using it to configure DAG in pipeline details page.

Before:

https://user-images.githubusercontent.com/56132941/213337299-b7570a94-0768-42b5-afd4-8827e6544956.mov

After:

https://user-images.githubusercontent.com/56132941/213337382-1d71f6fe-a40b-41b6-a337-a75f680b85c5.mov

### DIFF
--- a/frontend/src/pages/PipelineDetails.test.tsx
+++ b/frontend/src/pages/PipelineDetails.test.tsx
@@ -29,6 +29,7 @@ import TestUtils from '../TestUtils';
 import * as WorkflowUtils from 'src/lib/v2/WorkflowUtils';
 import { PageProps } from './Page';
 import PipelineDetails from './PipelineDetails';
+import { ApiJob } from '../apis/job';
 
 describe('PipelineDetails', () => {
   const updateBannerSpy = jest.fn();
@@ -40,6 +41,7 @@ describe('PipelineDetails', () => {
   const getPipelineVersionSpy = jest.spyOn(Apis.pipelineServiceApi, 'getPipelineVersion');
   const listPipelineVersionsSpy = jest.spyOn(Apis.pipelineServiceApi, 'listPipelineVersions');
   const getRunSpy = jest.spyOn(Apis.runServiceApi, 'getRun');
+  const getRecurringRunSpy = jest.spyOn(Apis.jobServiceApi, 'getJob');
   const getExperimentSpy = jest.spyOn(Apis.experimentServiceApi, 'getExperiment');
   const deletePipelineVersionSpy = jest.spyOn(Apis.pipelineServiceApi, 'deletePipelineVersion');
   const getPipelineVersionTemplateSpy = jest.spyOn(
@@ -52,21 +54,32 @@ describe('PipelineDetails', () => {
   let testPipeline: ApiPipeline = {};
   let testPipelineVersion: ApiPipelineVersion = {};
   let testRun: ApiRunDetail = {};
+  let testRecurringRun: ApiJob = {};
 
-  function generateProps(fromRunSpec = false): PageProps {
+  function generateProps(fromRunSpec = false, fromRecurringRunSpec = false): PageProps {
+    let params = {};
+    if (!fromRunSpec && !fromRecurringRunSpec) {
+      params = {
+        [RouteParams.pipelineId]: testPipeline.id,
+        [RouteParams.pipelineVersionId]:
+          (testPipeline.default_version && testPipeline.default_version!.id) || '',
+      };
+    }
+
+    let search = '';
+    if (fromRunSpec) {
+      search = `?${QUERY_PARAMS.fromRunId}=test-run-id`;
+    } else if (fromRecurringRunSpec) {
+      search = `?${QUERY_PARAMS.fromRecurringRunId}=test-recurring-run-id`;
+    }
+
     const match = {
       isExact: true,
-      params: fromRunSpec
-        ? {}
-        : {
-            [RouteParams.pipelineId]: testPipeline.id,
-            [RouteParams.pipelineVersionId]:
-              (testPipeline.default_version && testPipeline.default_version!.id) || '',
-          },
+      params: params,
       path: '',
       url: '',
     };
-    const location = { search: fromRunSpec ? `?${QUERY_PARAMS.fromRunId}=test-run-id` : '' } as any;
+    const location = { search } as any;
     const pageProps = TestUtils.generatePageProps(
       PipelineDetails,
       location,
@@ -112,12 +125,21 @@ describe('PipelineDetails', () => {
       },
     };
 
+    testRecurringRun = {
+      id: 'test-recurring-run-id',
+      name: 'test recurring run',
+      pipeline_spec: {
+        pipeline_id: 'run-pipeline-id',
+      },
+    };
+
     getPipelineSpy.mockImplementation(() => Promise.resolve(testPipeline));
     getPipelineVersionSpy.mockImplementation(() => Promise.resolve(testPipelineVersion));
     listPipelineVersionsSpy.mockImplementation(() =>
       Promise.resolve({ versions: [testPipelineVersion] }),
     );
     getRunSpy.mockImplementation(() => Promise.resolve(testRun));
+    getRecurringRunSpy.mockImplementation(() => Promise.resolve(testRecurringRun));
     getExperimentSpy.mockImplementation(() =>
       Promise.resolve({ id: 'test-experiment-id', name: 'test experiment' } as ApiExperiment),
     );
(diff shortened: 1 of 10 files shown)


zijianjoy:
General question regarding this PR before looking into detail: Shouldn't we also support V2 pipeline detail: http://go/kfp-repo/blob/master/frontend/src/pages/PipelineDetailsV2.tsx as well?

frontend/src/pages/PipelineDetails.tsx
```diff
@@ -59,6 +61,12 @@ interface PipelineDetailsState {
   versions: ApiPipelineVersion[];
 }
 
+type origin = {
```
zijianjoy, line 64: Use capital for first letter in `Origin`.

frontend/src/pages/PipelineDetails.tsx
```diff
@@ -248,34 +290,72 @@ class PipelineDetails extends Page<{}, PipelineDetailsState> {
...
+            }
+          } catch (err) {
+            await this.showPageError(
+              `Failed to parse pipeline spec from run with ID: ${recurringRunDetails.id}.`,
+              err,
+            );
+            logger.error(
+              `Failed to parse pipeline spec JSON from run with ID: ${recurringRunDetails.id}.`,
+              err,
+            );
```
zijianjoy, line 341: This part is similar to the logics above, suggest to simply by consolidating similar logics into function. (And try to reduce the amount of if/else nested block)

frontend/src/pages/PipelineDetails.tsx
```diff
@@ -186,9 +194,38 @@ class PipelineDetails extends Page<{}, PipelineDetailsState> {
...
+            experiment.id!,
+          ),
+        },
+      );
+    } else {
+      breadcrumbs.push({ displayName: 'All runs', href: RoutePage.RUNS });
+    }
+    breadcrumbs.push({
+      displayName: origin.isRecurring ? origin.recurringRun!.name! : origin.run!.run?.name!,
+      href: RoutePage.RUN_DETAILS.replace(':' + RouteParams.runId, runId),
```
zijianjoy, line 220: I think this href will need to generate different URL based on whether it is recurring run or run. Right? RecurringRun doesn't have a run ID.

frontend/src/pages/PipelineDetails.tsx
```diff
@@ -248,34 +290,72 @@ class PipelineDetails extends Page<{}, PipelineDetailsState> {
...
-          displayName: runDetails.run!.name!,
-          href: RoutePage.RUN_DETAILS.replace(':' + RouteParams.runId, fromRunId),
-        });
+
+        const relatedExperimentId = RunUtils.getFirstExperimentReferenceId(recurringRunDetails);
+        let experiment: ApiExperiment | undefined;
+        if (relatedExperimentId) {
+          experiment = await Apis.experimentServiceApi.getExperiment(relatedExperimentId);
+        }
+        this.buildBreadCumb(breadcrumbs, origin, fromRecurringRunId, experiment);
```
zijianjoy, line 350: Why are we passing `fromRecurringRunId` to `runId` argument? RecurringRunId and RunId are ID for different database tables.

frontend/src/pages/PipelineDetails.test.tsx
```diff
@@ -52,21 +54,32 @@ describe('PipelineDetails', () => {
   let testPipeline: ApiPipeline = {};
   let testPipelineVersion: ApiPipelineVersion = {};
   let testRun: ApiRunDetail = {};
+  let testRecurringRun: ApiJob = {};
+
+  function generateProps(fromRunSpec = false, fromRecurringRunSpec = false): PageProps {
+    let params = {};
+    if (!fromRunSpec && !fromRecurringRunSpec) {
```
zijianjoy, line 61: NIT: Suggest to add some comment here to explain that: if no `fromXXX` parameter is provided, it means KFP UI expects to show Pipeline detail with pipeline version ID.

jlyaoyuli:
Also fix the `No graph to show` bug for run of recurringRun in RunList.

Before:

https://user-images.githubusercontent.com/56132941/214452686-ed56c3d3-19d5-4dda-8fae-f3c0b8a317bb.mov

After:

https://user-images.githubusercontent.com/56132941/214452726-fbc89416-30f0-4780-824c-b50835b615bd.mov

frontend/src/pages/PipelineDetails.tsx
```diff
@@ -199,51 +231,68 @@ class PipelineDetails extends Page<{}, PipelineDetailsState> {
...
-    // If fromRunId is specified, load the run and get the pipeline template from it
-    if (fromRunId) {
+    // If fromRunId or fromRecurringRunId is specified, load the run and get the pipeline template from it
+    if (origin) {
+      const errorTextAdjective = origin.isRecurring ? 'recurring ' : '';
       try {
-        const runDetails = await Apis.runServiceApi.getRun(fromRunId);
+        origin.recurringRun = origin.isRecurring
+          ? await Apis.jobServiceApi.getJob(origin.recurringRunId!)
+          : undefined;
```
chensun, line 240: Should it be `null` instead? My impression is `null` is intentional empty, which is preferred in explicit initialization.

frontend/src/pages/PipelineDetails.tsx
```diff
@@ -199,51 +231,68 @@ class PipelineDetails extends Page<{}, PipelineDetailsState> {
     let selectedVersion: ApiPipelineVersion | undefined;
     let versions: ApiPipelineVersion[] = [];
 
-    // If fromRunId is specified, load the run and get the pipeline template from it
-    if (fromRunId) {
+    // If fromRunId or fromRecurringRunId is specified, load the run and get the pipeline template from it
+    if (origin) {
+      const errorTextAdjective = origin.isRecurring ? 'recurring ' : '';
```
chensun, line 236: nit: for better readability, I would have

```javascript
const msgRunOrRecurringRun = origin.isRecurring ? 'recurring run' : 'run';
```
and down below 
```
`Failed to parse pipeline spec from ${msgRunOrRecurringRun} with ID:`
```

chensun (commented):
Thanks, Joe. I left a few nitpick comments, but overall LGTM.

frontend/src/pages/PipelineDetails.tsx
```diff
@@ -59,6 +61,14 @@ interface PipelineDetailsState {
   versions: ApiPipelineVersion[];
 }
 
+type Origin = {
+  isRecurring: boolean;
```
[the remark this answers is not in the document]
jlyaoyuli: Correct! The `isRecurring` is not necessary; however, the inline expression will become a little weird like `id = runId ? runId : recurringRunId`. Thus, we choose to add a flag here, which is consistent to previous implementation in `NewRunV2`

frontend/src/pages/PipelineDetails.tsx
jlyaoyuli: Probably the inline expression is a little confusing. We don't have to make the field empty intentionally. I will change this part as
```
if (origin.isRecurring) {
    origin.recurringRun = await Apis.jobServiceApi.getJob(origin.recurringRunId!);
} else {
    origin.run = await Apis.runServiceApi.getRun(origin.runId!);
}
```
(i.e. only define the field when it's needed)

frontend/src/pages/PipelineDetails.tsx
```diff
@@ -186,9 +201,26 @@ class PipelineDetails extends Page<{}, PipelineDetailsState> {
     return this.load();
   }
 
+  private getOrigin() {
+    const urlParser = new URLParser(this.props);
+    const fromRunId = urlParser.get(QUERY_PARAMS.fromRunId);
+    const fromRecurringRunId = urlParser.get(QUERY_PARAMS.fromRecurringRunId);
+
+    if (fromRunId && fromRecurringRunId) {
+      throw new Error('The existence of run and recurring run should be exclusive.');
```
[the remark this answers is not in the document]
jlyaoyuli: The condition that run and recurringRun are both empty is the general situation that user check the pipeline details directly (no clone process). It should not throw an error.

frontend/src/pages/PipelineDetails.tsx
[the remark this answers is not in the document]
chensun: I personally don't feel `id = runId ? runId : recurringRunId` being weird. It can be also written as 
```javascript
id = rundId || recurringRunId;
``` 
which might look cleaner?

Anyway, this is a nitpick, feel free to ignore if you prefer a dedicated flag.

frontend/src/pages/PipelineDetails.tsx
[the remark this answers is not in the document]
chensun: Ah, that makes sense. Thanks for the explanation.

frontend/src/pages/PipelineDetails.tsx
```diff
@@ -199,51 +231,67 @@ class PipelineDetails extends Page<{}, PipelineDetailsState> {
     let selectedVersion: ApiPipelineVersion | undefined;
     let versions: ApiPipelineVersion[] = [];
 
-    // If fromRunId is specified, load the run and get the pipeline template from it
-    if (fromRunId) {
+    // If fromRunId or fromRecurringRunId is specified, load the run and get the pipeline template from it
+    if (origin) {
+      const msgRunOrRecurringRun = origin.isRecurring ? 'recurring ' : '';
```
chensun, line 236: nit: `? 'recurring run' : 'run'`
make the msg self-contained, so you can have `${msgRunOrRecurringRun}` instead of `${msgRunOrRecurringRun}run` below.

chensun (commented):
/lgtm

chensun (approved):
/lgtm
/approve

Thanks!

PR #8743, opened 2023-01-20
feat(backend): add support for environment variables. Fixes #8704

### LINKED ISSUE #8704: feat(backend): Support environment variables

Linchin:
* Added support for environment variables at task level
* Added compiler test

**Note:**
* Need to update driver image at [argo.go](https://github.com/kubeflow/pipelines/blob/8d87bec4661fc3fb82a513f9ea37129a93f4e2ce/backend/src/v2/compiler/argocompiler/argo.go#L110).

Fixes #8704.

**Checklist:**

### DIFF
--- a/backend/src/v2/driver/driver.go
+++ b/backend/src/v2/driver/driver.go
@@ -315,8 +315,8 @@ func makePodSpecPatch(
 
 	// Convert environment variables
 	userEnvVar := make([]k8score.EnvVar, 0)
-	for _, envVar := range container.Env {
-		userEnvVar = append(userEnvVar, k8score.EnvVar{Name: envVar.Name, Value: envVar.Value})
+	for _, envVar := range container.GetEnv() {
+		userEnvVar = append(userEnvVar, k8score.EnvVar{Name: envVar.GetName(), Value: envVar.GetValue()})
 	}
 
 	userCmdArgs := make([]string, 0, len(container.Command)+len(container.Args))
--- a/backend/src/v2/driver/driver.go
+++ b/backend/src/v2/driver/driver.go
@@ -313,6 +313,12 @@ func makePodSpecPatch(
 		return "", fmt.Errorf("failed to make podSpecPatch: %w", err)
 	}
 
+	// Convert environment variables
+	userEnvVar := make([]k8score.EnvVar, 0)
+	for _, envVar := range container.Env {
+		userEnvVar = append(userEnvVar, k8score.EnvVar{Name: envVar.Name, Value: envVar.Value})
+	}
+
 	userCmdArgs := make([]string, 0, len(container.Command)+len(container.Args))
 	userCmdArgs = append(userCmdArgs, container.Command...)
 	userCmdArgs = append(userCmdArgs, container.Args...)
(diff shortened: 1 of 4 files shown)


backend/src/v2/driver/driver.go
```diff
@@ -313,6 +313,12 @@ func makePodSpecPatch(
 		return "", fmt.Errorf("failed to make podSpecPatch: %w", err)
 	}
 
+	// Convert environment variables
+	userEnvVar := make([]k8score.EnvVar, 0)
+	for _, envVar := range container.Env {
```
gkcalat, line 318: NIT: can we use `container.GetEnv()` provided by the proto compiler?

backend/src/v2/driver/driver.go
```diff
@@ -313,6 +313,12 @@ func makePodSpecPatch(
...
+		userEnvVar = append(userEnvVar, k8score.EnvVar{Name: envVar.Name, Value: envVar.Value})
```
gkcalat, line 319: NIT: similarly, can we use `.GetName()` and `.GetValue()` instead?

gkcalat (commented):
Hi @Linchin! Thank you!

- Do we need to implement the same changes in `DAG` and `RootDAG` or env variables will only be available for container `driverType`? If it covers other cases, should we consider adjusting the [cmd interface](https://github.dev/kubeflow/pipelines/blob/8d87bec4661fc3fb82a513f9ea37129a93f4e2ce/backend/src/v2/cmd/driver/main.go#L141)?
- I am not sure, but do we need to bump the drive image? For example, [here](https://github.dev/kubeflow/pipelines/blob/8d87bec4661fc3fb82a513f9ea37129a93f4e2ce/backend/src/v2/compiler/argocompiler/argo.go#L110)?
- Should we adjust or add a new test that would fail if env variables are not set & read properly?

Linchin:
> Hi @Linchin! Thank you!
> 
> * Do we need to implement the same changes in `DAG` and `RootDAG` or env variables will only be available for container `driverType`? If it covers other cases, should we consider adjusting the [cmd interface](https://github.dev/kubeflow/pipelines/blob/8d87bec4661fc3fb82a513f9ea37129a93f4e2ce/backend/src/v2/cmd/driver/main.go#L141)?

No, we only need to change the env variables in drivers.


> * I am not sure, but do we need to bump the drive image? For example, [here](https://github.dev/kubeflow/pipelines/blob/8d87bec4661fc3fb82a513f9ea37129a93f4e2ce/backend/src/v2/compiler/argocompiler/argo.go#L110)?

Yes, we will have to, since the driver image is changed.

> * Should we adjust or add a new test that would fail if env variables are not set & read properly?

I think the tests are done in the compiling process in sdk.

backend/src/v2/driver/driver.go
Linchin: This makes sense, since `container.GetEnv()` covers the case when container is nil. I will change the code. I don't think it's much a problem though, since the rest of the function also directly accessed the fields.

backend/src/v2/driver/driver.go
Linchin: Similarly, see [reply](https://github.com/kubeflow/pipelines/pull/8743#discussion_r1083062553).

Linchin:
Use new PR #8886 after rebase.

PR #8765, opened 2023-01-27
feat(backend): support optional and default parameters. Fixes #8716

### LINKED ISSUE #8716: feat(backend): Support optional input parameters

Linchin:
* Default value for parameters is already supported
* Added support for optional parameters
Fixes #8716.

**Checklist:**


gkcalat (changes requested):
Thank you @Linchin!

It seems that the sample run is actually in `Failed` state. Can you check locally if it gets into succeeded state?

backend/src/v2/driver/driver.go
```diff
@@ -671,7 +671,7 @@ func resolveInputs(ctx context.Context, dag *metadata.DAG, iterationIndex *int,
 				}
 			}
 			value, hasValue := inputs.GetParameterValues()[name]
-			if !hasValue {
+			if !hasValue && inputsSpec.GetParameters()[name].IsOptional == false {
```
chensun, line 674: With this change, aren't we missing the opportunity to get default value when `IsOptional == true`?

samples/v2/component_with_optional_inputs_test.py
```diff
@@ -0,0 +1,59 @@
...
+    component_op_dict = tasks['component-op'].get_dict()
+
+    t.assertEqual(
+        {
+            'name': 'component-op',
+            'inputs': {
+                'parameters': {
+                    'input1': 'Hello',
+                    'input2': 'World',
+                    'input3': ''
```
chensun, line 44: `None` is different from empty string, should we expect absent of `input3` here?

backend/src/v2/driver/driver.go
Linchin: This is a really good point! I should change the logic here.

gkcalat: Would it be more clear? WDYT?
```golang
if !hasValue {
   value = spec.GetDefaultValue()
   if inputsSpec.GetParameters()[name].IsOptional {
      value = structpb.NewStringValue("")
   }
   if value == nil {
      return fmt.Errorf("no value or default value provided for non-optional parameter %q", name)
   }
}
```

samples/v2/component_with_optional_inputs.py
```diff
@@ -0,0 +1,42 @@
...
+
+from kfp import compiler
+from kfp import dsl
+from kfp.dsl import component
+
+
+@component
+def component_op(
+    input1: str = 'default value',
+    input2: Optional[str] = None,
```
gkcalat, line 24: Do we support other types? If yes,m should we add them: `int`, `float`, `list`, `dict`, `set`, etc.

samples/v2/component_with_optional_inputs.py
```diff
@@ -0,0 +1,42 @@
...
+    input3: Optional[str] = None,
```
gkcalat, line 25: Would `None` have the same behavior as `''` (this may answer Chen's questions below)? If not, should we add an input with default value `''` to verify the behavior is as expected?

backend/src/v2/driver/driver.go
```diff
@@ -671,13 +671,42 @@ func resolveInputs(ctx context.Context, dag *metadata.DAG, iterationIndex *int,
 				}
 			}
 			value, hasValue := inputs.GetParameterValues()[name]
+			/*
```
gkcalat, line 674: NIT: remove dead code?

backend/src/v2/driver/driver.go
Linchin: Thank you @gkcalat for suggesting this. I'm already using a different logic handling this, following this truth table:
| optional | has value | action                                    |
|----------|-----------|-------------------------------------------------|
| +        | +         | use provided value                              |
| +        | -         | use default value, if no default, then use None |
| -        | +         | use provided value                              |
| -        | -         | use default value, if no default, then error    |
When a parameter is optional, it is still possible that there is a default value, see [example](https://github.com/kubeflow/pipelines/blob/69969475bc09ea0d48115ea48a0b0041043c1d1c/samples/test/lightweight_python_functions_v2_with_outputs.py#L72).

samples/v2/component_with_optional_inputs.py
Linchin: Thank you! I don't think `None` is the same as empty string `''` in Python, and the input type `string` is covered in `input1`.

backend/src/v2/driver/driver.go
Linchin: Thank you, this was kept for testing purposes. I will remove this later.

backend/src/v2/driver/driver.go
chensun: > 
> optional | has value | action
> -- | -- | --
> | + | - |  use default value, if no default, then use None |

What does this `None` mean? Is it Python `None`, or Golang `nil`, or something else?

chensun (commented):
To better understand the topic, I'd suggest start with the sample you included in this PR.
- Compile it, inspect the produced IR
- Run it on Vertex Pipelines, examine the runtime behavior
- Run in on KFP, compare the behavior with Vertex

It would help clarify the concepts and expected behaviors.

Linchin:
> What does this `None` mean? Is it Python `None`, or Golang `nil`, or something else? 

@chensun My understanding is the `None` here applies in Python, because ultimately the pipeline is executed in Python, thus the input being `None`. Its representation in backend should support this `None` value.

chensun:
> > What does this `None` mean? Is it Python `None`, or Golang `nil`, or something else?
> 
> @chensun My understanding is the `None` here applies in Python, because ultimately the pipeline is executed in Python, thus the input being `None`. Its representation in backend should support this `None` value.

Chatted offline, but for visibility post here again in case anyone had the same misunderstanding or confusion.
- Pipeline is a set of components and that's all--there's nothing in between of pipelines and components. In some occasions, you may find a pipeline has some python code in it, for example, below is still a valid pipeline defintion:
```python
@dsl.pipeline
def my_pipeline():
    print('hello')
    my_component(input='world')
```
But the `print` statement would not be executed at the pipeline runtime, instead it will be resolved (executed and then discarded) at pipeline compilation time. So again, a pipeline is nothing but a set of component(s). 

- Components are always executed in containers. It's language-agnostic. Even if we know the container is running some Python script, e.g. a component created as a lightweight Python component, the data passing is always through container command-line interface, where everything being passed is a string.

backend/src/v2/driver/driver.go
Linchin: See [note](https://github.com/kubeflow/pipelines/pull/8765#issuecomment-1439774421).

samples/v2/component_with_optional_inputs_test.py
Linchin: Yes, `input3` should be absent here. For more information, see [discussion](https://github.com/kubeflow/pipelines/pull/8765#discussion_r1109030724).

samples/v2/component_with_optional_inputs.py
Linchin: Thank you, I added more types.

backend/src/v2/driver/driver.go
```diff
@@ -671,13 +671,26 @@ func resolveInputs(ctx context.Context, dag *metadata.DAG, iterationIndex *int,
...
-			if !hasValue {
+
+			// Handle when parameter does not have input value
+			if !hasValue && inputsSpec.GetParameters()[name].IsOptional == false {
+				// when parameter is not optional, value comes from default value, otherwise error
 				if spec.GetDefaultValue() == nil {
-					return fmt.Errorf("input parameter %q is required", name)
+					return fmt.Errorf("no value or default value provided for non-optional parameter %q", name)
 				}
 				inputs.GetParameterValues()[name] = spec.GetDefaultValue()
```
chensun, line 681: I think this is probably dead code. If a parameter has a default value then optional must be true. 
"no value + not optional" goes to error directly.

backend/src/v2/driver/driver.go
```diff
@@ -671,13 +671,26 @@ func resolveInputs(ctx context.Context, dag *metadata.DAG, iterationIndex *int,
...
 				value = spec.GetDefaultValue()
+			} else if !hasValue && inputsSpec.GetParameters()[name].IsOptional == true {
+				// when parameter is optional, value comes from default value, otherwise we delete this parameter from input list, because nothing is being passed
```
chensun, line 684: comment isn't accurate, "optional + no value" means values comes from default. 
Can we avoid the deletion part--only added it when needed instead of add first then delete later if applicable?

backend/src/v2/driver/driver.go
Linchin: This is true. I will remove this.

backend/src/v2/driver/driver.go
Linchin: Thank you for catching this, I updated the logic and comments. I also removed the deletion part, since it doesn't seem to exist in the first place.

backend/src/v2/driver/driver.go
```diff
@@ -671,13 +671,22 @@ func resolveInputs(ctx context.Context, dag *metadata.DAG, iterationIndex *int,
...
+			// Handle when parameter does not have input value
+			if !hasValue && inputsSpec.GetParameters()[name].IsOptional == false {
+				// when parameter is not optional and there is no input value, report error
+				return fmt.Errorf("no value provided for non-optional parameter %q", name)
+			} else if !hasValue && inputsSpec.GetParameters()[name].IsOptional == true {
+				// when parameter is optional and there is no input value, value comes from default value
+				if spec.GetDefaultValue() != nil {
+					inputs.GetParameterValues()[name] = spec.GetDefaultValue()
+					value = spec.GetDefaultValue()
+				} else {
```
chensun, line 684: I don't recall if we reached consensus on this, but I called out in our group chat that this entire block can be omitted--only add value to executor input when users provided a value, leave default value to be filled in inside the container regardless whether it's `None` or something else. 
If you decide to keep this block at this moment. Can you please add a TODO comment above say we might want to revisit whether to have this block at all, so that we don't forgot all the context.

chensun (commented):
/lgtm

gkcalat:
/lgtm

@Linchin, this needs to be rebased.

gkcalat (approved):

backend/src/v2/driver/driver.go
chensun: To be clear, for Python-function component, whether we pass the default value or not makes no difference, and I was proposing to not pass. For customer component though, it depends on whether user code inside the container handles the default value. My argument was they should, thus passing default value here is not needed and should be avoided unless necessary. The other argument was some existing custom container may not work in this way, if we don't pass default value, it could break them. So I could see going either way at this moment. Starting with not passing default value is fine, and we could change it if too many users are on the other side.

backend/src/v2/driver/driver.go
```diff
@@ -678,13 +678,12 @@ func resolveInputs(ctx context.Context, dag *metadata.DAG, iterationIndex *int,
 				}
 			}
 			value, hasValue := inputs.GetParameterValues()[name]
+
+			// Skip when hasValue == false, which means this parameter does not have an input value
 			if !hasValue {
```
chensun, line 683: Don't you still need the isOptional check? "no value and not optional" should throw an error.

backend/src/v2/driver/driver.go
Linchin: I agree, both sides have their reasonings. I will add a note here about both possible choices.

backend/src/v2/driver/driver.go
Linchin: Thank you, I added this check back.

backend/src/v2/driver/driver.go
```diff
@@ -678,13 +678,24 @@ func resolveInputs(ctx context.Context, dag *metadata.DAG, iterationIndex *int,
...
-				if spec.GetDefaultValue() == nil {
-					return fmt.Errorf("input parameter %q is required", name)
-				}
-				inputs.GetParameterValues()[name] = spec.GetDefaultValue()
-				value = spec.GetDefaultValue()
+				continue
```
gkcalat, line 684: I might be wrong, but wouldn't this make the two if-statements below unreachable? Should we remove the code and leave the comment there explaining the reasoning behind the decision to not replace the parameters to their default values?

Linchin: Thank you, I should have removed this.

chensun (approved):
/lgtm
/approve

PR #8767, opened 2023-01-30
fix(test): Add frontend integration test manual run

gkcalat:

The goal of this PR is to improve debugging experience for the frontend integration test.

1. Changes the flow of the FE e2e test. The test waits for 90 seconds before opening run details page to verify run's status.
1. Adds local debugging option that would be of a great value for those who face failed tests.
1. Uploads test artifacts to the GCS bucket (this was missing, although artifacts are created).

**Checklist:**

### DIFF
--- a/test/frontend-integration-test/README.md
+++ b/test/frontend-integration-test/README.md
@@ -1,0 +1,24 @@
+# Frontend integration test
+
+This test gets triggered by the end-to-end testing workflows.
+
+## Local run
+
+1. Deploy KFP on a k8s cluster and enable port forwarding. Replace the default namespace `kubeflow` below if needed.
+
+    ```bash
+    POD=`kubectl get pods -n kubeflow -l app=ml-pipeline-ui -o jsonpath='{.items[0].metadata.name}'`
+    kubectl port-forward -n kubeflow ${POD} 3000:3000 &
+    ```
+
+1. Build the container with the tests:
+
+    ```bash
+    docker build . -t kfp-frontend-integration-test:local
+    ```
+
+1. Run the test with enabled networking (**this exposes your local networking stack to the testing container**):
+
+    ```bash
+    docker run --net=host kfp-frontend-integration-test:local --remote-run true
+    ```
--- a/test/frontend-integration-test/helloworld.spec.js
+++ b/test/frontend-integration-test/helloworld.spec.js
@@ -72,10 +72,6 @@ describe('deploy helloworld sample run', () => {
     $('#experimentDescription').setValue(experimentDescription);
 
     $('#createExperimentBtn').click();
-
-    browser.waitUntil(() => {
-      return new URL(browser.getUrl()).hash.startsWith('#/runs/new');
-    }, waitTimeout);
   });
 
   it('creates a new run in the experiment', () => {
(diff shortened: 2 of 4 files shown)


test/frontend-integration-test/README.md
```diff
@@ -0,0 +1,24 @@
+# Frontend integration test
```
chensun, line 1: Thank you for adding the additional running mode, and the documenting the usage.

/cc @jlyaoyuli in case frontend team has a similar or better way to run FE integration test in custom deployment.

test/frontend-integration-test/helloworld.spec.js
```diff
@@ -135,31 +140,35 @@ describe('deploy helloworld sample run', () => {
...
   it('waits for run to finish', () => {
-    let status = getValueFromDetailsTable('Status');
-
     let attempts = 0;
-    const maxAttempts = 60;
+    const maxAttempts = 90;
 
-    // Wait for a reasonable amount of time until the run is done
-    while (attempts < maxAttempts && status.trim() !== 'Succeeded') {
+    while (attempts < maxAttempts) {
```
chensun, line 151: Without the other condition, we are simply pausing 90 seconds before entering the run detail page followed by opening the "Config" tab. So this chunk of change is equivalent to adding a `browser.pause(90 * 1000);` after line 137 (old) / 142 (new). 
Also in my local testing, the "Config" tab for v1 does auto-refresh itself--the "Details" tab for v2 run however doesn't auto-refresh until I click into some blank area on the page. 
So I'm not confident this would solve the test failure in your other PR. 
Can we revert the change in this file and merge only the enabling remote-run change? You can have the one line change of  adding `browser.pause(90 * 1000);` in your other PR to see if it helps.

test/frontend-integration-test/helloworld.spec.js
```diff
@@ -135,31 +140,35 @@ describe('deploy helloworld sample run', () => {
...
-      status = getValueFromDetailsTable('Status');
       attempts++;
     }
 
-    assert(attempts < maxAttempts, `waited for ${maxAttempts} seconds but run did not succeed. ` +
-      'Current status is: ' + status);
+    browser.execute('document.querySelector(".tableRow a").click()');
+    browser.waitUntil(() => {
+      return new URL(browser.getUrl()).hash.startsWith('#/runs/details');
+    }, waitTimeout);
```
chensun, line 159: Waiting for the run details page to load is technically correct but doesn't change anything in the current setup, because `$('button=Config').waitForVisible();` below would also ensure the page to load--unless there's no config button on run list page.

test/frontend-integration-test/helloworld.spec.js
gkcalat: Good point. Reverted.

test/frontend-integration-test/helloworld.spec.js
gkcalat: To add my 5 cents: waiting for the page is more explicit. I reverted the change as you suggested above.

chensun: Absolutely agree explicit is better, and that's what I meant by "technically correct". For this PR, I think it's better to focus on what it claims, the effective change in the `helloworld.spec.js` can be reduced into one-line change of waiting for 90 seconds before clicking into "config" tab, and I was a bit skeptical whether that would fix the test failure we saw. For that reason I suggested we don't make this not-so-relevant change. 
If we were to change this file in a meaningful way again, say in your large PR, I would support waiting explicitly on the page load.

chensun:
/lgtm
/approve

Thanks!

/cc @zijianjoy @jlyaoyuli

PR #8779, opened 2023-02-02
feat(sdk): Dedupe pipeline spec

JOCSTAA:
This PR adds a feature to clean up the IR YAML file, removing component specs that are duplicates of each which is a result of a component being called by multiple tasks

**Checklist:**

### DIFF
--- a/sdk/python/kfp/compiler/compiler_test.py
+++ b/sdk/python/kfp/compiler/compiler_test.py
@@ -3961,7 +3961,7 @@ class TestDedupePipelineSpec(unittest.TestCase):
         def my_pipeline():
             task1 = print_op()
             task2 = print_op()
-            use_secret_as_env(task1, 'jaja', {'haha': 'haha'})
+            use_secret_as_env(task1, 'test', {'test-key': 'test-feature'})
             task3 = print_op()
 
         pipeline_spec = my_pipeline.pipeline_spec
--- a/sdk/python/kfp/compiler/compiler_test.py
+++ b/sdk/python/kfp/compiler/compiler_test.py
@@ -3111,245 +3111,6 @@ class TestOutputDefinitionsPresentWhenCompilingComponents(unittest.TestCase):
                       comp.pipeline_spec.root.output_definitions.parameters)
 
 
-class TestDedupePipelineSpec(unittest.TestCase):
-
-    def test_basic_would_dedupe(self):
-
-        @dsl.component
-        def print_op(message: str = 'default'):
-            print(message)
-
-        @dsl.component
-        def identity(message: str) -> str:
-            return (message)
-
-        @dsl.pipeline()
-        def my_pipeline(pipeline_input: str = 'pipeine_input'):
-            task1 = print_op()
-            task2 = print_op(message='constant')
-            task3 = print_op(message=pipeline_input)
-
-            upstream_task = identity(message='upstream_task_output')
-            task4 = print_op(message=upstream_task.output)
-
-        pipeline_spec = my_pipeline.pipeline_spec
-
-        self.assertEqual(len(pipeline_spec.components), 2)
-
-        # check that primary component spec and corresponding executor exists
-        self.assertIn('comp-print-op', pipeline_spec.components)
-        self.assertIn(
-            'exec-print-op', pipeline_spec.deployment_spec.fields['executors']
-            .struct_value.fields)
-
-        # check that the second component spec and corresponding executor does not exist
-        self.assertNotIn('comp-print-op-2', pipeline_spec.components)
-        self.assertNotIn(
-            'exec-print-op-2', pipeline_spec.deployment_spec.fields['executors']
-            .struct_value.fields)
-        self.assertNotIn('comp-print-op-3', pipeline_spec.components)
-        self.assertNotIn(
-            'exec-print-op-3', pipeline_spec.deployment_spec.fields['executors']
-            .struct_value.fields)
-        self.assertNotIn('comp-print-op-4', pipeline_spec.components)
-        self.assertNotIn(
-            'exec-print-op-4', pipeline_spec.deployment_spec.fields['executors']
-            .struct_value.fields)
-
-        # check that primary component contains the correct executor label
-        self.assertEqual(
-            pipeline_spec.components['comp-print-op'].executor_label,
-            'exec-print-op')
-
-        # check tasks point to corrrect component spec reference
-        self.assertEqual(
-            pipeline_spec.root.dag.tasks['print-op'].component_ref.name,
-            'comp-print-op')
-        self.assertEqual(
-            pipeline_spec.root.dag.tasks['print-op-2'].component_ref.name,
-            'comp-print-op')
-        self.assertEqual(
-            pipeline_spec.root.dag.tasks['print-op-3'].component_ref.name,
-            'comp-print-op')
-        self.assertEqual(
-            pipeline_spec.root.dag.tasks['print-op-4'].component_ref.name,
-            'comp-print-op')
-
-        # check task specific input/ output information is not lost
-        self.assertEqual(
-            pipeline_spec.root.dag.tasks['print-op-2'].inputs
-            .parameters['message'].runtime_value.constant.string_value,
-            'constant')
-        self.assertEqual(
-            pipeline_spec.root.dag.tasks['print-op-3'].inputs
-            .parameters['message'].component_input_parameter, 'pipeline_input')
-        self.assertEqual(
-            pipeline_spec.root.dag.tasks['print-op-4'].inputs
-            .parameters['message'].task_output_parameter.producer_task,
-            'identity')
-        self.assertEqual(
-            pipeline_spec.root.dag.tasks['print-op-4'].inputs
-            .parameters['message'].task_output_parameter.output_parameter_key,
-            'Output')
-
-    def test_task_specific_executor_case_would_not_dedupe(self):
-
-        @dsl.component
-        def print_op(message: str = 'default'):
-            print(message)
-
-        @dsl.pipeline()
-        def my_pipeline():
-            task1 = print_op()
-            task2 = print_op().set_cpu_limit('4').set_memory_limit('14Gi')
-
-        pipeline_spec = my_pipeline.pipeline_spec
-
-        self.assertEqual(len(pipeline_spec.components), 2)
-
-        # check both tasks component spec and executors exist
-        self.assertIn('comp-print-op', pipeline_spec.components)
-        self.assertIn(
-            'exec-print-op', pipeline_spec.deployment_spec.fields['executors']
-            .struct_value.fields)
-        self.assertIn('comp-print-op-2', pipeline_spec.components)
-        self.assertIn(
-            'exec-print-op-2', pipeline_spec.deployment_spec.fields['executors']
-            .struct_value.fields)
-
-        # check both component specs contains the correct executor label
-        self.assertEqual(
-            pipeline_spec.components['comp-print-op'].executor_label,
-            'exec-print-op')
-        self.assertEqual(
-            pipeline_spec.components['comp-print-op-2'].executor_label,
-            'exec-print-op-2')
-
-        # check tasks point to corrrect component spec reference
-        self.assertEqual(
-            pipeline_spec.root.dag.tasks['print-op'].component_ref.name,
-            'comp-print-op')
-        self.assertEqual(
-            pipeline_spec.root.dag.tasks['print-op-2'].component_ref.name,
-            'comp-print-op-2')
-
-        # check information of resource spec is not lost
-        self.assertEqual(
-            4.0, pipeline_spec.deployment_spec.fields['executors'].struct_value
-            .fields['exec-print-op-2'].struct_value['container']
-            ['resources'].fields['cpuLimit'].number_value)
-        self.assertAlmostEqual(
-            15.032385536, pipeline_spec.deployment_spec.fields['executors']
-            .struct_value.fields['exec-print-op-2'].struct_value['container']
-            ['resources'].fields['memoryLimit'].number_value)
-
-    def test_task_specific_configurations_would_dedupe(self):
-
-        @dsl.component
-        def print_op(message: str = 'default'):
-            print(message)
-
-        @dsl.pipeline()
-        def my_pipeline():
-            task1 = print_op()
-            task2 = print_op(message='different').set_display_name('name')
-
-        pipeline_spec = my_pipeline.pipeline_spec
-
-        self.assertEqual(len(pipeline_spec.components), 1)
-
-        # check one component spec and executor exist
-        self.assertIn('comp-print-op', pipeline_spec.components)
-        self.assertIn(
-            'exec-print-op', pipeline_spec.deployment_spec.fields['executors']
-            .struct_value.fields)
-        self.assertNotIn('comp-print-op-2', pipeline_spec.components)
-        self.assertNotIn(
-            'exec-print-op-2', pipeline_spec.deployment_spec.fields['executors']
-            .struct_value.fields)
-
-        # check tasks point to correct component spec
-
-        self.assertEqual(
-            pipeline_spec.root.dag.tasks['print-op'].component_ref.name,
-            'comp-print-op')
-        self.assertEqual(
-            pipeline_spec.root.dag.tasks['print-op-2'].component_ref.name,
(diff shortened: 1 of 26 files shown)


connor-mccarthy (changes requested):
Thanks, @JOCSTAA! Could you add some tests to cover some of the different interesting cases for this solution (e.g., IR with pipeline-in-pipeline, resource requirements, etc.)?

sdk/python/kfp/compiler/compiler_test.py
```diff
@@ -2652,5 +2652,114 @@ def my_pipeline(x: Input[Artifact] = Artifact(
...
+        # check information of resource spec is not lost
+        self.assertEqual(
+            4.0, pipeline_spec.deployment_spec.fields['executors'].struct_value
+            .fields['exec-print-op-2'].struct_value['container']
+            ['resources'].fields['cpuLimit'].number_value)
+        self.assertEqual(
+            15.032385536, pipeline_spec.deployment_spec.fields['executors']
+            .struct_value.fields['exec-print-op-2'].struct_value['container']
+            ['resources'].fields['memoryLimit'].number_value)
+
```
connor-mccarthy, line 2762: These tests look good. Does it also make sense to have a test for the pipeline-in-pipeline case? And `ComponentSpec`-level configuration, such as display name?

sdk/python/kfp/compiler/pipeline_spec_builder.py
```diff
@@ -1570,6 +1571,135 @@ def write_pipeline_spec_to_file(pipeline_spec: pipeline_spec_pb2.PipelineSpec,
...
+
+                    for _, task_spec in pipeline_spec.root.dag.tasks.items():
+                        if task_spec.component_ref.name == initial:
+                            task_spec.component_ref.name = component
+
+        for initial, new_details in changed_names.items():
+            del pipeline_spec.components[initial]
+            component_name, component_spec = new_details
+            pipeline_spec.components[component_name].CopyFrom(component_spec)
+
```
connor-mccarthy, line 1699: I'm curious if this function can be simplified by maintaining a map of (unique) serialized executor bodies to the executor label we want to use for it. Pseudocode is probably better than a description for conveying the idea:

```python
unique_executors = {}  # maps body to label
for component_name, component_body in components.items():
    if component.executor_label:
        executor_body = executors[executor_label]
        if executor_body in unique_executors:
            executor_body_str = serialize_executor_body(executor_body)
            unique_executors[executor_body] = correct_executor_label
            update_pipeline_spec(pipeline_spec, component_name, executor_label)
```

WDYT?

connor-mccarthy (changes requested):

sdk/python/kfp/compiler/pipeline_spec_builder.py
JOCSTAA: This simplification would change a lot of the logic that has already been written, I don't know if its worth trying to build the logic from scratch on top of this

sdk/python/kfp/components/graph_component.py
```diff
@@ -85,3 +89,140 @@ def pipeline_spec(self) -> pipeline_spec_pb2.PipelineSpec:
...
+                                component_spec.executor_label]
+
+                        corresponding_executor_name = component_spec.executor_label
+                        corresponding_executor_name = corresponding_executor_name[:
+                                                                                  -2]
+                        corresponding_executor_name = utils.make_name_unique_by_adding_index(
+                            name=corresponding_executor_name,
+                            collection=pipeline_spec.deployment_spec
+                            .fields['executors'].struct_value.fields.keys(),
+                            delimiter='-')
```
connor-mccarthy, line 207: Can you help me understand what this section does?

connor-mccarthy (changes requested):

sdk/python/kfp/compiler/compiler_test.py
```diff
@@ -3180,5 +3180,197 @@ def pipeline_component(input_list: Input[List[Artifact]]):
             .input_definitions.artifacts['input_list'].is_artifact_list, True)
 
 
+class TestDedupePipelineSpec(unittest.TestCase):
+
+    def test_basic(self):
```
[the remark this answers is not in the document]
JOCSTAA: Can you please clarify what you mean by this?

sdk/python/kfp/components/graph_component.py
```diff
@@ -85,3 +89,140 @@ def pipeline_spec(self) -> pipeline_spec_pb2.PipelineSpec:
...
+                pipeline_spec.components.items()):
+            if component_spec.executor_label and component_name not in components_with_clones:
+                original_components_executor_spec = pipeline_spec.deployment_spec.fields[
+                    'executors'].struct_value.fields[
+                        component_spec.executor_label]
+                for executor_name, executor_spec in sorted(
+                        pipeline_spec.deployment_spec.fields['executors']
+                        .struct_value.fields.items()):
+                    corresponding_component_name = 'comp' + executor_name[4:]
+                    if executor_name != component_spec.executor_label and executor_spec == original_components_executor_spec and corresponding_component_name not in components_with_clones and pipeline_spec.components[
```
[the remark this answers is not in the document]
JOCSTAA: Oh right, It wont

sdk/python/kfp/components/graph_component.py
```diff
@@ -85,3 +89,140 @@ def pipeline_spec(self) -> pipeline_spec_pb2.PipelineSpec:
...
+                component]
+
+            last_delimiter = component.rfind('-')
+            if len(component) > last_delimiter + 1 and component[
+                    last_delimiter + 1:].isnumeric():
+                component = component[:-2]
+                component = utils.make_name_unique_by_adding_index(
+                    name=component,
+                    collection=pipeline_spec.components.keys(),
+                    delimiter='-')
```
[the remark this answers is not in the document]
JOCSTAA: I believe this was a duplicate code block, it has been removed

sdk/python/kfp/components/graph_component.py
JOCSTAA: This cleans up the yaml file component spec names, so there are no missing numbers for vacant name spaces caused by deduped component spec names

sdk/python/kfp/components/graph_component.py
```diff
@@ -85,3 +89,165 @@ def pipeline_spec(self) -> pipeline_spec_pb2.PipelineSpec:
...
+        calling on the same components.
+
+        component specs are only deduped when the executor specs are the
+        same, this means factors like two tasks called on the same
+        component with differing resource specs would not be deduped,
+        deduping would still occur if the caller tasks differ by their
+        input specifications since this not affect the executor spec of
+        the components
+        """
+        from dataclasses import dataclass
```
connor-mccarthy, line 106: Per [Google Python Style Guide](https://google.github.io/styleguide/pyguide.html#22-imports), import packages/modules, not classes/functions directly.

sdk/python/kfp/components/graph_component.py
```diff
@@ -85,3 +89,165 @@ def pipeline_spec(self) -> pipeline_spec_pb2.PipelineSpec:
...
+
+        @dataclass
+        class clones_data_structure:
```
connor-mccarthy, line 109: Can we make the class name [PascalCase](https://pylint.pycqa.org/en/1.7/user_guide/options.html#:~:text=camelCase-,PascalCase,-UPPER_CASE)?

sdk/python/kfp/components/graph_component.py
```diff
@@ -85,3 +89,165 @@ def pipeline_spec(self) -> pipeline_spec_pb2.PipelineSpec:
...
+
+            def __init__(self, component_name, component_spec, executor_name,
```
connor-mccarthy, line 111: Dataclasses will take case of this for you. See [example](https://realpython.com/python-data-classes/#basic-data-classes).

sdk/python/kfp/components/graph_component.py
```diff
@@ -85,3 +89,165 @@ def pipeline_spec(self) -> pipeline_spec_pb2.PipelineSpec:
...
+                         executor_spec) -> None:
+                self.component_name = component_name
+                self.component_spec = component_spec
+                self.executor_name = executor_name
+                self.executor_spec = executor_spec
+
+            def get_data(self) -> list:
```
connor-mccarthy, line 118: Can we access the data directly from the dataclass instance using its attributes, instead of returning a list with `.get_data()`? The motivations for using a dataclass are to (a) avoid heterogenous lists and (b) avoid depending on the position of data in a list for obtaining its value. The `.get_data() approach doesn't quite address these.

sdk/python/kfp/components/graph_component.py
```diff
@@ -85,3 +89,165 @@ def pipeline_spec(self) -> pipeline_spec_pb2.PipelineSpec:
...
+                return [
+                    self.executor_name, self.component_spec, self.executor_spec
+                ]
+
+        components_with_clones = {}
+
+        def _collect_and_process_duplicates():
```
connor-mccarthy, line 125: Can we refactor these functions to pass in and return variables, instead of modifying non-local objects?

sdk/python/test_data/pipelines/lightweight_python_functions_pipeline.yaml
```diff
@@ -239,4 +239,4 @@ root:
       message:
         parameterType: STRING
 schemaVersion: 2.1.0
-sdkVersion: kfp-2.0.0-beta.8
+sdkVersion: kfp-2.0.0-beta.12
```
connor-mccarthy, line 242: Can we remove the diff in this file since it's unrelated to the PR?

sdk/python/kfp/compiler/compiler_test.py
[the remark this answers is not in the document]
connor-mccarthy: Discussed offline.

sdk/python/kfp/compiler/compiler_test.py
```diff
@@ -3180,5 +3180,232 @@ def pipeline_component(input_list: Input[List[Artifact]]):
...
+            pipeline_spec.root.dag.tasks['print-op-2'].component_ref.name,
+            'comp-print-op-2')
+
+        # check information of resource spec is not lost
+        self.assertEqual(
+            4.0, pipeline_spec.deployment_spec.fields['executors'].struct_value
+            .fields['exec-print-op-2'].struct_value['container']
+            ['resources'].fields['cpuLimit'].number_value)
+        self.assertEqual(
+            15.032385536, pipeline_spec.deployment_spec.fields['executors']
```
connor-mccarthy, line 3311: For float comparison, I think we should use [`assertAlmostEqual`](`https://docs.python.org/3/library/unittest.html#unittest.TestCase.assertAlmostEqual`).

sdk/python/kfp/compiler/compiler_test.py
```diff
@@ -3180,5 +3180,232 @@ def pipeline_component(input_list: Input[List[Artifact]]):
...
+            print(msg)
+
+        @dsl.pipeline(name='graph-component')
+        def graph_component(msg: str):
+            print_op(msg=msg)
+
+        @dsl.pipeline(name='test-pipeline')
+        def my_pipeline():
+            task1 = graph_component(msg='hello')
+            task2 = graph_component(msg='hello')
```
connor-mccarthy, line 3371: What happens if we reuse `print_op` again inside `my_pipeline`? Can we add this to the test case?

sdk/python/kfp/compiler/compiler_test.py
```diff
@@ -3180,5 +3180,232 @@ def pipeline_component(input_list: Input[List[Artifact]]):
...
+            task2 = print_op()
+            task3 = print_op()
+            task4 = print_op().set_cpu_limit('4').set_memory_limit('14Gi')
+
+        pipeline_spec = my_pipeline.pipeline_spec
+
+        # Check that only print-op-1 and print-op-2 exist instead of print-op-1 and print-op-4
+        self.assertIn('comp-print-op', pipeline_spec.components)
+        self.assertIn('comp-print-op-2', pipeline_spec.components)
+        self.assertNotIn('comp-print-op-4', pipeline_spec.components)
```
connor-mccarthy, line 3407: Good thinking on these tests. Do we also need to check the executor label?

sdk/python/kfp/components/graph_component.py
```diff
@@ -85,3 +89,165 @@ def pipeline_spec(self) -> pipeline_spec_pb2.PipelineSpec:
...
+                        for task_spec in component_spec.dag.tasks.values():
+                            if task_spec.component_ref.name in clones:
+                                task_spec.component_ref.name = component_name
+
+        def _clean_up_component_spec_names():
+            # clean up other component spec names
+            if not components_with_clones:
+                return
+            changed_names = {}
+            for component_name, component_spec in sorted(
```
connor-mccarthy, line 197: Is sorting important here? Ideally we sort the data structures once then use them throughout or avoid sorting altogether when it's not required by the implementation.

sdk/python/kfp/components/graph_component.py
```diff
@@ -85,3 +89,165 @@ def pipeline_spec(self) -> pipeline_spec_pb2.PipelineSpec:
...
+                    pipeline_spec.components.items()):
+                if component_name not in components_with_clones and component_spec.executor_label:
+                    last_delimiter = component_name.rfind('-')
+                    if len(
+                            component_name
+                    ) > last_delimiter + 1 and component_name[last_delimiter +
+                                                              1:].isnumeric():
+                        initial = component_name
```
connor-mccarthy, line 205: Consider extracting some of the subroutines into helper functions throughout. For example, this could be `get_name_from_component_ref` or something similar.

sdk/python/kfp/compiler/compiler_test.py
```diff
@@ -3180,5 +3180,232 @@ def pipeline_component(input_list: Input[List[Artifact]]):
...
+        @dsl.component
+        def print_op(message: str = 'default'):
+            print(message)
+
+        @dsl.pipeline()
+        def my_pipeline(pipeline_input: str = 'pipeine_input'):
+            task1 = print_op()
+            task2 = print_op()
+            task3 = print_op()
+            task4 = print_op().set_cpu_limit('4').set_memory_limit('14Gi')
```
connor-mccarthy, line 3400: Is it possible there's a bug here? If I reorder as the following, the tests fail (non-deterministically):

```python
        @dsl.pipeline()
        def my_pipeline(pipeline_input: str = 'pipeine_input'):
            task1 = print_op()
            task4 = print_op().set_cpu_limit('4').set_memory_limit('14Gi')
            task2 = print_op()
            task3 = print_op()
```

connor-mccarthy (changes requested):

sdk/python/kfp/compiler/compiler_test.py
JOCSTAA: Nothing particularly new happens, it adds a new task spec for `task3` `print_op` but still dedupes all `print_op ` component specs and executor_specs into one, would you still recommend i add it to the test case, if so what would you recommend i test for about it?

sdk/python/kfp/components/graph_component.py
JOCSTAA: This information is carried in the `pipeline_spec` proto,  in the first function the sorting was not done to the proto itself but to the dict copy extracted using `.items()`, so we would need to extract another dict copy and sort it again in the new function

sdk/python/kfp/compiler/compiler_test.py
JOCSTAA: Good find! It is a bug that happens because the the `_clean_up_component_spec_names` function actually updates the name while still in the for loop that collects the names to be changed

sdk/python/kfp/compiler/compiler_test.py
connor-mccarthy: I think it makes sense to test a root pipeline with the structure below. The interesting cases are where `comp-1` is deduped and the case where it is not. The same assertions you have for other tests make sense here too.

```txt
root-pipeline
    inner-pipeline
        comp-1
    comp-1
```

sdk/python/kfp/components/graph_component.py
connor-mccarthy: Sounds good. For cases it's needed, let's keep it in. If there are cases we can remove sorting we should to avoid suggesting that it's part of the algorithm's implementation.

PR #8804, opened 2023-02-06
fix(backend): add retry run in v2 api. Fixes #8735

### LINKED ISSUE #8735: fix(backend): KFP v2 API - missing retry run API

Linchin:
* Added retry run in v2 api proto
* Added support for retry run in v2 api in the backend
* Unit test

Fixes https://github.com/kubeflow/pipelines/issues/8735

**Checklist:**


backend/src/apiserver/server/run_server.go
```diff
@@ -499,7 +499,17 @@ func (s *RunServer) terminateRun(ctx context.Context, runId string, experimentId
...
+// Applies common logic on v1beta1 and v2beta1 API.
+func (s *RunServer) retryRun(ctx context.Context, runId string, experimentId string) error {
+	err := s.canAccessRun(ctx, runId, &authorizationv1.ResourceAttributes{Verb: common.RbacResourceVerbRetry})
+	if err != nil {
+		return util.Wrap(err, "Failed to authorize the request")
+	}
+	return s.resourceManager.RetryRun(ctx, runId)
+}
+
+// Retarts a run.
```
chensun, line 512: s/Retarts/Terminates/

backend/src/apiserver/server/run_server.go
```diff
@@ -499,7 +499,17 @@ func (s *RunServer) terminateRun(ctx context.Context, runId string, experimentId
 	return s.resourceManager.TerminateRun(ctx, runId)
 }
 
-// Terminates a run.
+// Restarts a run. Not exported.
```
chensun, line 502: Maybe use "retry" to be more aligned with the method name?

"Not exported." is not needed and obvious since the function starts with the lowercase letter.

backend/src/apiserver/server/run_server.go
```diff
@@ -661,6 +668,21 @@ func (s *RunServer) TerminateRun(ctx context.Context, request *apiv2beta1.Termin
 	return &empty.Empty{}, nil
 }
 
+// Restarts a run.
```
chensun, line 671: s/Restarts/Retries

backend/api/v2beta1/run.proto
```diff
@@ -436,4 +443,13 @@ message ReadArtifactRequest {
 message ReadArtifactResponse {
   // Byte array of the artifact content.
   bytes data = 1;
+}
+
+message RetryRunRequest {
+ // The ID of the parent experiment.
+ string experiment_id = 1;
+
+
```
gkcalat, line 452: NIT: remove empty line

backend/api/v2beta1/run.proto
```diff
@@ -436,4 +443,13 @@ message ReadArtifactRequest {
...
+ // The ID of the run to be Retried.
```
gkcalat, line 453: NIT: `Retried` -> `retried`

backend/src/apiserver/server/run_server.go
Linchin: Thank you so much! I fixed this and other comments in this file.

backend/src/apiserver/server/run_server.go
Linchin: Thank you, fixed.

backend/src/apiserver/server/run_server.go
Linchin: Thank you, fixed.

backend/src/apiserver/server/run_server.go
```diff
@@ -499,7 +499,17 @@ func (s *RunServer) terminateRun(ctx context.Context, runId string, experimentId
 	return s.resourceManager.TerminateRun(ctx, runId)
 }
 
-// Terminates a run.
+// Restarts a run. Not exported.
+// Applies common logic on v1beta1 and v2beta1 API.
+func (s *RunServer) retryRun(ctx context.Context, runId string, experimentId string) error {
```
gkcalat, line 504: `experimentId` not being used here. For v2, we can verify the provided `experimentId` with the one retrieved from `GetRun`. If they don't match or if `GetRun` returns an error, fail quickly. For v1, we simply read the `experimentId` from the same `GetRun` result as `experimentId` will be empty.

The `experimentId` can then be passed to the resource manager and instead of `getNamespaceFromRunId` we can call `GetNamespaceFromExperimentId` to save one extra call there.

The same logic is [already applied](https://github.com/kubeflow/pipelines/blob/master/backend/src/apiserver/server/run_server.go#L492-L498) to `TerminateRun`.

WDYT? 

/CC @chensun

backend/src/apiserver/server/run_server_test.go
```diff
@@ -1502,3 +1502,42 @@ func TestReadArtifacts_Succeed(t *testing.T) {
...
+		"param2": &structpb.Value{Kind: &structpb.Value_BoolValue{BoolValue: true}},
+		"param3": &structpb.Value{Kind: &structpb.Value_ListValue{ListValue: v2RuntimeListParams}},
+		"param4": &structpb.Value{Kind: &structpb.Value_NumberValue{NumberValue: 12}},
+		"param5": &structpb.Value{Kind: &structpb.Value_StructValue{StructValue: v2RuntimeStructParams}},
+	}
+
+	pipelineSpecStruct := &structpb.Struct{}
+	yaml.Unmarshal([]byte(v2SpecHelloWorld), pipelineSpecStruct)
+
+	run := &apiv2beta1.Run{
```
gkcalat, line 1528: NIT: add a comment explaining why this run will achieve the `FAILED` state.

backend/src/apiserver/server/run_server_test.go
```diff
@@ -1502,3 +1502,42 @@ func TestReadArtifacts_Succeed(t *testing.T) {
...
+			PipelineSpec: pipelineSpecStruct,
+		},
+		RuntimeConfig: &apiv2beta1.RuntimeConfig{
+			Parameters:   v2RuntimeParams,
+			PipelineRoot: "model-pipeline-root",
+		},
+	}
+	run, err := server.CreateRun(nil, &apiv2beta1.CreateRunRequest{Run: run})
+	assert.Nil(t, err)
+
```
gkcalat, line 1541: NIT: verify that the state is `FAILED`

backend/src/apiserver/server/run_server_test.go
```diff
@@ -1502,3 +1502,42 @@ func TestReadArtifacts_Succeed(t *testing.T) {
...
+	_, err = server.RetryRun(nil, &apiv2beta1.RetryRunRequest{RunId: run.RunId})
```
gkcalat, line 1542: NIT: add a case with retried a terminated run

backend/src/apiserver/resource/resource_manager.go
```diff
@@ -719,15 +719,15 @@ func (r *ResourceManager) RetryRun(ctx context.Context, runId string) error {
 		return util.Wrapf(err, "Failed to retry run %s", runId)
 	}
 
+	if namespace == "" {
```
gkcalat, line 722: Good catch! Thank you!

gkcalat (commented):
Thank you @Linchin. I left a few comments. Please, correct me where I was wrong.

backend/api/v2beta1/run.proto
Linchin: Thank you, removed.

backend/src/apiserver/server/run_server.go
chensun: Good catch, we shouldn't even pass experimentId here.
Retry is just to retry as-is, to allow changing experiment and more, users can use clone.
https://github.com/kubeflow/pipelines/blob/b3de27a7d01ee7febef58b92abc52ffb1f49b4a7/backend/api/v1beta1/run.proto#L180-L183

backend/api/v2beta1/run.proto
```diff
@@ -436,4 +443,13 @@ message ReadArtifactRequest {
 message ReadArtifactResponse {
   // Byte array of the artifact content.
   bytes data = 1;
+}
+
+message RetryRunRequest {
+ // The ID of the parent experiment.
+ string experiment_id = 1;
```
chensun, line 450: Remove `experiment_id`
Not presented in v1 either:

https://github.com/kubeflow/pipelines/blob/b3de27a7d01ee7febef58b92abc52ffb1f49b4a7/backend/api/v1beta1/run.proto#L180-L183

chensun (changes requested):

backend/api/v2beta1/run.proto
Linchin: Thank you for catching this! Fixed.

backend/api/v2beta1/run.proto
Linchin: The `experiment_id` is added to be consistent with other v2 methods in this proto, including [GetRunRequest](https://github.com/kubeflow/pipelines/blob/b3de27a7d01ee7febef58b92abc52ffb1f49b4a7/backend/api/v2beta1/run.proto#L345), [TerminateRunRequest](https://github.com/kubeflow/pipelines/blob/b3de27a7d01ee7febef58b92abc52ffb1f49b4a7/backend/api/v2beta1/run.proto#L379), [ArchiveRunRequest](https://github.com/kubeflow/pipelines/blob/b3de27a7d01ee7febef58b92abc52ffb1f49b4a7/backend/api/v2beta1/run.proto#L398), [UnarchiveRunRequest](https://github.com/kubeflow/pipelines/blob/b3de27a7d01ee7febef58b92abc52ffb1f49b4a7/backend/api/v2beta1/run.proto#L406), and [DeleteRunRequest](https://github.com/kubeflow/pipelines/blob/b3de27a7d01ee7febef58b92abc52ffb1f49b4a7/backend/api/v2beta1/run.proto#L414). All these messages include an `experiment_id` field because there is an `experiment_id` section in the API path ([example](https://github.com/kubeflow/pipelines/blob/b3de27a7d01ee7febef58b92abc52ffb1f49b4a7/backend/api/v2beta1/run.proto#L73)). I agree that here `experiment_id` is redundant, although it follows [AIP design principle](https://google.aip.dev/122). However, I do think we need to maintain consistency among all methods across the same service, i.e., either all methods within RunService should have `experiment_id` as part of the path, or no method does.

chensun: Synced offline, given that experiment is a parent resource of run, we need to keep the experiment_id here. But the current implementation of ignoring experiment id should be fixed as well. This is not a blocker for the PR, consider the same pattern exists in GetRuns, GetJobs, etc.--although we do have validation that could complain "the run_id doesn't belong to the experiment" for those APIs, while such validation is missing for the retry run. 
We can fix the validation across all the APIs to check if "run id can be find in the given experiment id" post beta0 release.

chensun (approved):
/lgtm
/approve

Thanks!

backend/src/apiserver/server/run_server.go
Linchin: Added logic to verify the experiment id is correct.

backend/api/v2beta1/run.proto
Linchin: Before seeing your comment, I added logic to verify the experiment id is correct.

gkcalat:
Thank you, @Linchin!

/lgtm
/approve

PR #8805, opened 2023-02-06
fix(backend): fix run, job from empty manifest. Closes #8802.

### LINKED ISSUE #8802: [backend] Can't create a run from pipeline with empty PipelineSpec (stored as a file in OS)
### Environment

KFP standalone

*  KFP version: 

1. Deploy KFP.
2. Try to create a run from the sample pipelines. 
3. Error message dialog will show the error validating the manifests.

### Expected result

A run.

### Materials and Reference

We need to handle the case with PipelineSpec field is empty in the backend.

**Note**: a fix is in progress.

---

gkcalat:

- Adding to read manifest from object store for pipeline versions with empty PipelineSpec field
- Test cases for creating runs and a job in the upgrade e2e test
- Fixes #8802 

**Checklist:**

### DIFF
--- a/backend/src/apiserver/resource/resource_manager.go
+++ b/backend/src/apiserver/resource/resource_manager.go
@@ -387,7 +387,7 @@ func (r *ResourceManager) CreateRun(ctx context.Context, run *model.Run) (*model
 	run.ExperimentId = expId
 	run.Namespace = expNs
 
-	// Fetch pipeline version from pipeline spec
+	// Fetch pipeline version based on pipeline spec (do not create anything yet)
 	var wfTemplate *template.Template
 	pipelineVersion, err := r.fetchPipelineVersionFromPipelineSpec(run.PipelineSpec, run.DisplayName, run.Description, run.Namespace)
 	if err != nil {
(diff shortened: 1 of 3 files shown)


backend/src/apiserver/resource/resource_manager.go
```diff
@@ -1064,19 +1068,28 @@ func (r *ResourceManager) CreateJob(ctx context.Context, job *model.Job) (*model
...
+		manifest = pipelineVersion.PipelineSpec
+	}
 	// Get manifest from either of the two places:
 	// (1) raw manifest in pipeline_spec
 	// (2) pipeline version in resource_references
 	// And the latter takes priority over the former when the manifest is from pipeline_spec.pipeline_id
 	// workflow/pipeline manifest and pipeline id/version will not exist at the same time, guaranteed by the validation phase
 	var tmpl template.Template
+	// This should only happen when creating a run from an existing pipeline or pipeline version
 	if wfTemplate == nil {
```
chensun, line 1081: My general feeling reading the code, logic seems to be right, but the code is hard to follow. 
We check `if manifest == ""` multiple times in different stages, and also check `if wfTemplate == nil` and patch it separately. `manifest` and `wfTemplate` seems to me the same content that serves different interfaces, they should probably be finalized at the same place. For example, I could imaging we read `manifest` from different possible sources, and generate `wfTemplate` at the end. There's probably some room for refactoring to make the logic easier to follow.

chensun (approved):
/lgtm
/approve

Thanks!

backend/src/apiserver/resource/resource_manager.go
gkcalat: I agree. There are a few other places that could be improved within the resource manager. I will try to improve it in the later PRs before GA.

PR #8824, opened 2023-02-08
fix(frontend): Change the prompt and error message of parameter field into more readable text

jlyaoyuli:
Previously, KFP directly use proto type as error message of parameter field. The main purpose of this change is change the type in the text more readable, which aligns with MP's design.

The changes as follow:
```
NUMBER_INTEGER -> integer
NUMBER_DOUBLE -> double
BOOLEAN -> boolean
STRING -> string
LIST -> list
STRUCT -> map / dict
```

Before:
Screenshot 2023-02-08 at 11 36 02 AM

After:
Screenshot 2023-02-08 at 11 37 39 AM

### DIFF
--- a/frontend/src/components/NewRunParametersV2.tsx
+++ b/frontend/src/components/NewRunParametersV2.tsx
@@ -60,6 +60,15 @@ interface NewRunParametersProps {
   setIsValidInput?: (isValid: boolean) => void;
 }
 
+const protoMap = new Map<string, string>([
+  ['NUMBER_DOUBLE', 'float/double'],
+  ['NUMBER_INTEGER', 'integer'],
+  ['STRING', 'string'],
+  ['BOOLEAN', 'boolean'],
+  ['LIST', 'list'],
+  ['STRUCT', 'map/dict'],
+]);
+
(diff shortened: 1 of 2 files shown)


frontend/src/components/NewRunParametersV2.tsx
```diff
@@ -60,6 +60,15 @@ interface NewRunParametersProps {
   setIsValidInput?: (isValid: boolean) => void;
 }
 
+const protoMap = new Map<string, string>([
+  ['NUMBER_DOUBLE', 'double'],
+  ['NUMBER_INTEGER', 'integer'],
+  ['STRING', 'string'],
+  ['BOOLEAN', 'boolean'],
+  ['LIST', 'list'],
+  ['STRUCT', 'map / dict'],
```
zijianjoy, line 69: NIT: I am wondering what is the context of these types? Are we referring to them as Python class?

jlyaoyuli: I think we are referring to python class!

zijianjoy (commented):
/lgtm
/approve

Thank you Joe!

PR #8838, opened 2023-02-10
feat(sdk): Ignore upstream failure

JOCSTAA:
Introduces a new syntax for pipeline tasks to consume outputs from the upstream task while at the same time ignoring if the upstream tasks succeeds or not. Similar to an exit handler but with the included ability to parse inputs to the exit task.

sample syntax
```
@dsl.pipeline()
def my_pipeline(sample_input1: str = 'message'):
    task = fail_op(message=sample_input1)
    clean_up = print_op(message=task.output).ignore_upstream_failure()
```

**Checklist:**

### DIFF
--- a/sdk/python/kfp/compiler/compiler_test.py
+++ b/sdk/python/kfp/compiler/compiler_test.py
@@ -2895,7 +2895,6 @@ class TestValidIgnoreUpstreamTaskSyntax(unittest.TestCase):
 
         @dsl.component
         def fail_op(message: str) -> str:
-            """Fails."""
             import sys
             print(message)
             sys.exit(1)
@@ -2903,7 +2902,6 @@ class TestValidIgnoreUpstreamTaskSyntax(unittest.TestCase):
 
         @dsl.component
         def print_op(message: str = 'default'):
-            """Prints a message."""
             print(message)
 
         @dsl.pipeline()
@@ -2912,20 +2910,50 @@ class TestValidIgnoreUpstreamTaskSyntax(unittest.TestCase):
             clean_up_task = print_op(
                 message=task.output).ignore_upstream_failure()
 
+        self.assertEqual(
+            my_pipeline.pipeline_spec.root.dag.tasks['print-op'].trigger_policy
+            .strategy, 2)
+        self.assertNotEqual(
+            my_pipeline.pipeline_spec.root.dag.tasks['fail-op'].trigger_policy
+            .strategy, 2)
+
+    def test_component_with_no_input_permitted(self):
+
+        @dsl.component
+        def fail_op(message: str) -> str:
+            import sys
+            print(message)
+            sys.exit(1)
+            return message
+
+        @dsl.component
+        def print_op():
+            print('default')
+
+        @dsl.pipeline()
+        def my_pipeline(sample_input1: str = 'message'):
+            task = fail_op(message=sample_input1)
+            clean_up_task = print_op().ignore_upstream_failure()
+
+        self.assertEqual(
+            my_pipeline.pipeline_spec.root.dag.tasks['print-op'].trigger_policy
+            .strategy, 2)
+        self.assertNotEqual(
+            my_pipeline.pipeline_spec.root.dag.tasks['fail-op'].trigger_policy
+            .strategy, 2)
+
     def test_clean_up_on_wrapped_pipeline_permitted(self):
 
         @dsl.component
-        def fail_op(message: str = 'msg') -> str:
-            """Fails."""
+        def fail_op(message: str = 'message') -> str:
             import sys
-            print(msg)
+            print(message)
             sys.exit(1)
-            return input_string
+            return message
 
         @dsl.component
-        def print_op(message: str = 'msg'):
-            """Prints a message."""
-            print(msg)
+        def print_op(message: str = 'message'):
+            print(message)
 
         @dsl.pipeline
         def wrapped_pipeline(message: str = 'message') -> str:
@@ -2943,12 +2971,18 @@ class TestValidIgnoreUpstreamTaskSyntax(unittest.TestCase):
             compiler.Compiler().compile(
                 pipeline_func=my_pipeline, package_path=package_path)
 
+        self.assertEqual(
+            my_pipeline.pipeline_spec.root.dag.tasks['print-op'].trigger_policy
+            .strategy, 2)
+        self.assertNotEqual(
+            my_pipeline.pipeline_spec.root.dag.tasks['fail-op'].trigger_policy
+            .strategy, 2)
+
     def test_clean_up_task_with_no_default_value_for_upstream_input_blocked(
             self):
 
         @dsl.component
         def fail_op(message: str) -> str:
-            """Fails."""
             import sys
             print(message)
             sys.exit(1)
@@ -2956,13 +2990,11 @@ class TestValidIgnoreUpstreamTaskSyntax(unittest.TestCase):
 
         @dsl.component
         def print_op(message: str):
-            """Prints a message."""
             print(message)
 
         with self.assertRaisesRegex(
                 ValueError,
-                r'requires a default value to make sure the Exit handler never fails.'
-        ):
+                r'requires a default value to make sure it never fails.'):
 
             @dsl.pipeline()
             def my_pipeline(sample_input1: str = 'message'):
@@ -2975,7 +3007,6 @@ class TestValidIgnoreUpstreamTaskSyntax(unittest.TestCase):
 
         @dsl.component
         def fail_op(message: str) -> str:
-            """Fails."""
             import sys
             print(message)
             sys.exit(1)
@@ -2983,7 +3014,6 @@ class TestValidIgnoreUpstreamTaskSyntax(unittest.TestCase):
 
         @dsl.component
         def print_op(message: str):
-            """Prints a message."""
             print(message)
 
         @dsl.pipeline()
@@ -2992,11 +3022,17 @@ class TestValidIgnoreUpstreamTaskSyntax(unittest.TestCase):
             clean_up_task = print_op(
                 message=sample_input1).ignore_upstream_failure()
 
+        self.assertEqual(
+            my_pipeline.pipeline_spec.root.dag.tasks['print-op'].trigger_policy
+            .strategy, 2)
+        self.assertNotEqual(
+            my_pipeline.pipeline_spec.root.dag.tasks['fail-op'].trigger_policy
+            .strategy, 2)
+
     def test_clean_up_task_with_no_default_value_for_constant_permitted(self):
 
         @dsl.component
         def fail_op(message: str) -> str:
-            """Fails."""
             import sys
             print(message)
             sys.exit(1)
@@ -3004,13 +3040,19 @@ class TestValidIgnoreUpstreamTaskSyntax(unittest.TestCase):
 
         @dsl.component
         def print_op(message: str):
-            """Prints a message."""
             print(message)
 
         @dsl.pipeline()
         def my_pipeline(sample_input1: str = 'message'):
             task = fail_op(message=sample_input1)
             clean_up_task = print_op(message='sample').ignore_upstream_failure()
+
+        self.assertEqual(
+            my_pipeline.pipeline_spec.root.dag.tasks['print-op'].trigger_policy
+            .strategy, 2)
+        self.assertNotEqual(
+            my_pipeline.pipeline_spec.root.dag.tasks['fail-op'].trigger_policy
+            .strategy, 2)
 
 
 if __name__ == '__main__':
--- a/sdk/python/kfp/compiler/compiler_test.py
+++ b/sdk/python/kfp/compiler/compiler_test.py
@@ -2966,17 +2966,35 @@ class TestValidIgnoreUpstreamTaskSyntax(unittest.TestCase):
             clean_up_task = print_op(
                 message=task.output).ignore_upstream_failure()
 
-        with tempfile.TemporaryDirectory() as tempdir:
-            package_path = os.path.join(tempdir, 'pipeline.yaml')
-            compiler.Compiler().compile(
-                pipeline_func=my_pipeline, package_path=package_path)
-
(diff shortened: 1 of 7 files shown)


sdk/python/kfp/components/pipeline_task.py
```diff
@@ -474,6 +475,28 @@ def my_pipeline():
...
+                if (input_spec_name == input_name) and (isinstance(
+                        argument_value, pipeline_channel.PipelineChannel)) and (
+                            not input_spec.optional) and (
+                                argument_value.task_name in pipeline_context
+                                .Pipeline.get_default_pipeline().tasks):
+                    raise ValueError(
+                        f'Exit task {self.name} requires a default value to make sure the Exit handler never fails.'
+                    )
+
+        self.ignore_upstream_failure_tag = True
```
connor-mccarthy, line 498: Maybe prefix with underscore? We don't do this consistently for all other `PipelineTask` attributes, but probably better if we do.

sdk/python/kfp/components/pipeline_task.py
```diff
@@ -474,6 +475,28 @@ def my_pipeline():
             self._task_spec.dependent_tasks.append(task.name)
         return self
 
+    def ignore_upstream_failure(self) -> 'PipelineTask':
+        """If called, the pipeline task will run when any specified upstream
+        tasks complete, even if unsuccessful.
+
+        This method effectively turns the caller task into an exit task
+        if the caller task has upstream dependencies.
+        """
```
connor-mccarthy, line 484: I think we should probably add the same `Returns:` section found in other task methods to the docstring. WDYT?

sdk/python/kfp/components/pipeline_task.py
```diff
@@ -474,6 +475,28 @@ def my_pipeline():
...
+                        argument_value, pipeline_channel.PipelineChannel)) and (
+                            not input_spec.optional) and (
+                                argument_value.task_name in pipeline_context
+                                .Pipeline.get_default_pipeline().tasks):
+                    raise ValueError(
+                        f'Exit task {self.name} requires a default value to make sure the Exit handler never fails.'
+                    )
+
+        self.ignore_upstream_failure_tag = True
+
```
connor-mccarthy, line 499: Missing `return self`

sdk/python/kfp/components/pipeline_task.py
```diff
@@ -474,6 +475,28 @@ def my_pipeline():
...
+        tasks complete, even if unsuccessful.
+
+        This method effectively turns the caller task into an exit task
+        if the caller task has upstream dependencies.
+        """
+
+        from kfp.components import pipeline_context
+        for input_spec_name, input_spec in self.component_spec.inputs.items():
+            for input_name, argument_value in self._inputs.items():
+                if (input_spec_name == input_name) and (isinstance(
```
connor-mccarthy, line 489: Can we index into the inputs with `self._inputs[input_spec_name]` to avoid the nested for loop?

sdk/python/kfp/components/pipeline_task.py
connor-mccarthy, line 498: Maybe prefix with underscore? We don't do this consistently for all other PipelineTask attributes, but probably better if we do.

sdk/python/kfp/components/pipeline_task.py
connor-mccarthy, line 499: Missing `return self`

sdk/python/kfp/components/pipeline_task.py
connor-mccarthy, line 489: Can we index into the inputs with `self._inputs[input_spec_name]` to avoid the nested for loop?

sdk/python/kfp/components/pipeline_task.py
connor-mccarthy, line 484: I think we should probably add the same `Returns:` section found in other task methods to the docstring. WDYT?

sdk/python/kfp/compiler/compiler_test.py
```diff
@@ -2889,5 +2889,129 @@ def math_pipeline():
...
+        def print_op(message: str):
+            """Prints a message."""
+            print(message)
+
+        @dsl.pipeline()
+        def my_pipeline(sample_input1: str = 'message'):
+            task = fail_op(message=sample_input1)
+            clean_up_task = print_op(message='sample').ignore_upstream_failure()
+
+
```
connor-mccarthy, line 3015: For these tests, can you assert that the expected trigger policy is present for the expected task and not for the other tasks?

sdk/python/kfp/compiler/pipeline_spec_builder.py
```diff
@@ -300,6 +300,11 @@ def build_task_spec_for_task(
                 'str, int, float, bool, dict, and list.'
                 f'Got {input_value} of type {type(input_value)}.')
 
+        if task.ignore_upstream_failure_tag:
```
connor-mccarthy, line 303: I think this is over-indented one level and is called once for each input.

sdk/python/kfp/compiler/compiler_test.py
connor-mccarthy, line 3015: Can you add a read/write test for this? This helps catch side-effects in compilation that are missed by just checking that the compilation succeeds and a narrow check for a single field (in this case, the trigger policy field). This is as much to check the correctness of the current PR as it to prevent side-effects from future PRs that don't play nice with these changes.

Another benefit of the read/write tests is that it validates that any fields written are preserved when the component is loaded back in. For most features this isn't a concern, but it's good to have at least one test case for each feature to be sure.

sdk/python/kfp/compiler/compiler_test.py
```diff
@@ -2889,5 +2889,129 @@ def math_pipeline():
...
+            """Fails."""
+            import sys
+            print(msg)
+            sys.exit(1)
+            return input_string
+
+        @dsl.component
+        def print_op(message: str = 'msg'):
+            """Prints a message."""
+            print(msg)
```
connor-mccarthy, line 2928: Should be `message`. Same with `message` and `input_string` above.

sdk/python/kfp/compiler/compiler_test.py
```diff
@@ -2889,5 +2889,129 @@ def math_pipeline():
...
+        def fail_op(message: str) -> str:
+            """Fails."""
+            import sys
+            print(message)
+            sys.exit(1)
+            return message
+
+        @dsl.component
+        def print_op(message: str = 'default'):
+            """Prints a message."""
```
connor-mccarthy, line 2906: nit: can remove the docstrings unless they are there to help explain the behavior being tested

sdk/python/kfp/compiler/compiler_test.py
connor-mccarthy: For this comment (and the above about undefined variables): while this doesn't affect the correctness of the test case or throw an exception, it has the potential to be misleading. In general, the test cases should introduce nothing unique other than the specific behavior being tested.

sdk/python/kfp/compiler/compiler_test.py
```diff
@@ -2889,5 +2889,129 @@ def math_pipeline():
...
+            print(message)
+
+        @dsl.pipeline()
+        def my_pipeline(sample_input1: str = 'message'):
+            task = fail_op(message=sample_input1)
+            clean_up_task = print_op(
+                message=task.output).ignore_upstream_failure()
+
+    def test_clean_up_on_wrapped_pipeline_permitted(self):
```
connor-mccarthy, line 2915: Good idea adding this test

connor-mccarthy (changes requested):
Thanks, @JOCSTAA. Can you also add a release note to `./sdk/RELEASE.md`?

sdk/python/kfp/components/pipeline_task.py
JOCSTAA: Yea i think thats a good idea! I also included a sample

sdk/python/kfp/compiler/pipeline_spec_builder.py
JOCSTAA: Good catch! Thanks

sdk/python/kfp/components/pipeline_task.py
```diff
@@ -474,6 +475,42 @@ def my_pipeline():
...
+        """
+
+        for input_spec_name, input_spec in (self.component_spec.inputs or
+                                            {}).items():
+            argument_value = self._inputs[input_spec_name]
+            if (isinstance(argument_value, pipeline_channel.PipelineChannel)
+               ) and (not input_spec.optional) and (argument_value.task_name
+                                                    is not None):
+                raise ValueError(
+                    f' Task {self.name} requires a default value to make sure it never fails.'
```
connor-mccarthy, line 507: It can be useful to construct error messages first with what the requirement is, then what specific user action didn't adhere to the requirement. WDYT about the following?

`f'Tasks can only use .ignore_upstream_failure() if all input parameters that accept arguments created by an upstream task have a default value, in case the upstream task fails to produce its output. Input parameter task {self.name}'s {input_spec_name} argument is an output of an upstream task {argument_value.task_name}, but {input_spec_name} has no default value.'`

Perhaps there's a less verbose way to put this. Just want to provide a little more instruction for the user.

You can also use `!r` to normalize formatting of the interpolated strings, similar to [this](https://github.com/kubeflow/pipelines/blob/a45d72d75dd99c9cbb8dc4706d380e4ceb9ec47e/sdk/python/kfp/components/pipeline_task.py#L319).

connor-mccarthy (changes requested):

sdk/python/kfp/compiler/compiler_test.py
```diff
@@ -2889,5 +2889,170 @@ def math_pipeline():
...
+        @dsl.pipeline
+        def my_pipeline(sample_input1: str = 'message'):
+            task = wrapped_pipeline(message=sample_input1)
+            clean_up_task = print_op(
+                message=task.output).ignore_upstream_failure()
+
+        with tempfile.TemporaryDirectory() as tempdir:
+            package_path = os.path.join(tempdir, 'pipeline.yaml')
+            compiler.Compiler().compile(
+                pipeline_func=my_pipeline, package_path=package_path)
```
connor-mccarthy, line 2972: I think compilation is not necessary here

connor-mccarthy, line 2972: I think compilation is not necessary here (similar to test case above). I did not catch this on the previous passes -- my mistake.

sdk/python/kfp/compiler/compiler_test.py
```diff
@@ -2889,5 +2889,170 @@ def math_pipeline():
...
+
+        self.assertEqual(
+            my_pipeline.pipeline_spec.root.dag.tasks['print-op'].trigger_policy
+            .strategy, 2)
+        self.assertEqual(
+            my_pipeline.pipeline_spec.root.dag.tasks['fail-op'].trigger_policy
```
connor-mccarthy, line 2978: I believe this should check `wrapped-pipeline`. `fail-op` isn't present in the root dag (`my_pipeline.pipeline_spec.root.dag.tasks`), so the proto is returning the "zero"/default value of `pipeline_spec_pb2.PipelineTaskSpec` and giving a false negative test result because the trigger policy asserted here is the zero/default trigger policy.

sdk/python/kfp/compiler/compiler_test.py
```diff
@@ -2889,5 +2889,170 @@ def math_pipeline():
...
+            .strategy, 0)
+
+    def test_clean_up_on_wrapped_pipeline_permitted(self):
+
```
connor-mccarthy, line 2946: Perhaps also add the test case of `.ignore_upstream_failure()` called on a pipeline (`GraphComponent`).

sdk/python/kfp/components/pipeline_task.py
```diff
@@ -474,6 +475,42 @@ def my_pipeline():
...
+        """
+
+        for input_spec_name, input_spec in (self.component_spec.inputs or
+                                            {}).items():
+            argument_value = self._inputs[input_spec_name]
+            if (isinstance(argument_value, pipeline_channel.PipelineChannel)
+               ) and (not input_spec.optional) and (argument_value.task_name
+                                                    is not None):
+                raise ValueError(
+                    f'Tasks can only use .ignore_upstream_failure() if all input parameters that accept arguments created by an upstream task have a default value, \n\tin case the upstream task fails to produce its output. Input parameter task {self.name}`s {input_spec_name!r} argument is an output of an upstream task {argument_value.task_name}, but {input_spec_name} has no default value.'
```
connor-mccarthy, line 507: nit: if going to use `!r`, prefer using for all of the interpolated strings that represent user code symbols/names

connor-mccarthy: Separately, what is `\n\t` for here?

connor-mccarthy (changes requested):
Last few comments. Then let's merge this!

sdk/python/kfp/components/pipeline_task.py
JOCSTAA: > Separately, what is \n\t for here?

It gives a cleaner print of the error on the terminal

connor-mccarthy: I think that might depend on the width of the terminal. What happens if you make your terminal more narrow? This is what mine looks like:

image

JOCSTAA: Ok! Ill just remove the tag

connor-mccarthy:
/lgtm
/approve

Thank you, @JOCSTAA! This is a great feature.

connor-mccarthy:
`kubeflow-pipelines-sdk-execution-tests` failures are unrelated to these changes. Manually merging.

divyank000:
Can we expect this feature in this release cycle ?

connor-mccarthy:
@divyank000, yes, this will be included in the next v2.x.x beta release.

PR #8850, opened 2023-02-15
feature(backed): Status-IR.part-1: Add updating tasks and run statuses to api server

gkcalat:

1. Added `child_tasks` with `task_id` and `pod_name` (NodeId) to the `Run` proto
2. Introduced associated changes to the DB schema 
3. Added creating and updating entries to the `tasks` table
4. Added `JOIN` between runs and tasks
5. Added a flow to report tasks along the workflows
6. Simplified `ReportServer` since v1 and v2 proto message have the same structure and content
7. Added the corresponding converter functions
8. Added unit tests

In the next contributions:
1. Add MLMD client to persistence agent
2. Enable updating the task-level information by from MLMD
3. Reporting tasks as a separate call to `ReportService`

**Checklist:**


backend/api/v2beta1/run.proto
```diff
@@ -271,45 +271,65 @@ message PipelineTaskDetail {
...
   string task_id = 2;
 
+  // Name of the corresponding pod assigned by the orchestration engine.
+  // Also known as node_id.
+  string pod_name = 3;
+
   // User specified name of a task that is defined in
   // [Pipeline.spec][].
-  string display_name = 3;
+  string display_name = 4;
```
chensun, line 280: Protobuf field number can't be changed. We'll have to add new field with the next available number.

https://stackoverflow.com/questions/65230623/safeness-of-changing-proto-field-number

backend/api/v2beta1/swagger/kfp_api_single_file.swagger.json
```diff
@@ -1678,19 +1678,6 @@
         }
       }
     },
-    "protobufListValue": {
```
chensun, line 1681: Curious why do we see such deletion in generated code? Does that mean the last released api client doesn't really match our proto definition?

backend/api/v2beta1/run.proto
gkcalat: Thank you. Fixed.

backend/api/v2beta1/swagger/kfp_api_single_file.swagger.json
gkcalat: I don't know. I also noticed that `PipelineSpec *ProtobufStruct` changed to `PipelineSpec interface{}`

chensun (commented):
Left a few comments, overall LGTM.

Thanks!

backend/src/apiserver/storage/task_store.go
```diff
@@ -275,3 +331,160 @@ func (s *TaskStore) GetTask(id string) (*model.Task, error) {
...
+			task.UUID = id.String()
+		}
+		if task.CreatedTimestamp == 0 {
+			if task.StartedTimestamp == 0 {
+				now := s.time.Now().Unix()
+				task.StartedTimestamp = now
+				task.CreatedTimestamp = now
+			} else {
+				task.CreatedTimestamp = task.StartedTimestamp
+			}
```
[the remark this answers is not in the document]
gkcalat: No, it was going to replace the timestamp only when `StartedTimestamp` had been set already.

I changed it to explicitly check whether the `StartedTimestamp` is earlier that `now`. This should be safer going forward.
`StartedTimestamp` is set by the orchestration engine or the persistence agent.

The purpose is to be consisten with other resources on how we timestamp them. One aparent advantage is that it gives us a way to estimate how long it took for a a task to be schedulled and to be executed.

backend/src/apiserver/storage/task_store.go
```diff
@@ -275,3 +331,160 @@ func (s *TaskStore) GetTask(id string) (*model.Task, error) {
...
+	if original.Fingerprint == "" {
+		original.Fingerprint = patch.Fingerprint
+	}
+	if original.Name == "" {
+		original.Name = patch.Name
+	}
+	if original.ParentTaskId == "" {
+		original.ParentTaskId = patch.ParentTaskId
+	}
+	if original.State.ToV2() == model.RuntimeStateUnspecified && patch.State.ToV2() != model.RuntimeStateUnspecified {
```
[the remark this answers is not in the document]
gkcalat: `patch.State` could be any string (e.g. `NEW_ARGO_STATUS`). `ToV2()` will return `RuntimeStateUnspecified` for invalid or missing strings. My intention was to keep the original string if the new state is semantically not different from `RuntimeStateUnspecified`. Adding `ToV2()` when reassigning the state, however, lets us to remove the second condition. We just need to keep `ToV2()` function updated (and correct).

backend/src/apiserver/storage/task_store_test.go
```diff
@@ -273,3 +288,232 @@ func TestTaskStore_GetTask(t *testing.T) {
 		})
 	}
 }
+
+func TestTaskStore_patchWithExistingTasks(t *testing.T) {
+	db, taskStore := initializeTaskStore()
+	defer db.Close()
+
+	task1 := &model.Task{
```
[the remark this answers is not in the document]
gkcalat: Thank you. Fixed.

backend/src/apiserver/server/api_converter.go
```diff
@@ -1348,11 +1366,14 @@ func toModelRun(r interface{}) (*model.Run, error) {
 			},
 		},
 		RunDetails: model.RunDetails{
+			State:                   state.ToV2(),
```
chensun, line 1369: Not to be addressed now, but another incidence of the `ToV2()` vs. `ToV1()` vs. "nocall" confusion.

backend/src/apiserver/server/api_converter.go
```diff
@@ -1615,90 +1628,164 @@ func toApiRunDetailV1(r *model.Run) *apiv1beta1.RunDetail {
...
 	}, nil
 }
 
+// Converts API tasks details into their internal representations.
+// Supports both v1beta1 and v2beta1 API.
+func toModelTasks(t interface{}) ([]*model.Task, error) {
+	if t == nil {
+		return nil, nil
+	}
+	switch t := t.(type) {
```
chensun, line 1727: This method probably has a too generic or flexible interface:
- case []*apiv2beta1.PipelineTaskDetail
- case *apiv2beta1.Run
- case []*apiv1beta1.Task
- case *util.Workflow:
- case workflowapi.Nodes:

Do we need a single method that handles all these types of inputs? 
And the method it calls into, which is `toModelTask`, again handles multiple types at once:
- case *apiv1beta1.Task:
- case *apiv2beta1.PipelineTaskDetail
- case workflowapi.NodeStatus

backend/src/apiserver/server/api_converter_test.go
```diff
@@ -2306,3 +2314,2356 @@ func TestToApiRecurringRun(t *testing.T) {
...
+			"nil",
+			nil,
+			model.RuntimeStateUnknownV1,
+			model.RuntimeStateUnspecified,
+			false,
+			"",
+		},
+		{
+			"V2 int32=7",
+			int32(7),
```
chensun, line 2514: Is this a possible value we could get? Is 7 the enum value?
And we expect a filed state for v1 yet canceled state for v2?

backend/src/apiserver/server/api_converter_test.go
```diff
@@ -2306,3 +2314,2356 @@ func TestToApiRecurringRun(t *testing.T) {
...
+
+func Test_toApiRuntimeStatus(t *testing.T) {
+	tests := []struct {
+		name        string
+		modelStatus *model.RuntimeStatus
+		want        *apiv2beta1.RuntimeStatus
+	}{
+		{
+			"nil",
+			nil,
```
chensun, line 2846: Is `nil` a valid state? Does it result to some error at a later point?

backend/src/apiserver/server/api_converter_test.go
```diff
@@ -2306,3 +2314,2356 @@ func TestToApiRecurringRun(t *testing.T) {
...
+					"b2": {
+						ArtifactIds: []int64{4, 5, 6},
+					},
+				},
+				ParentTaskId: "8",
+				StateHistory: []*apiv2beta1.RuntimeStatus{
+					{
+						UpdateTime: &timestamppb.Timestamp{Seconds: 9},
+						State:      apiv2beta1.RuntimeState_PAUSED,
+						Error:      util.ToRpcStatus(util.NewInvalidInputError("Sample error2")),
```
chensun, line 3098: What would be a real world example for `Sample error2`? and what's its relation with the `Sample error` above on line 3082? 
Given it's an `NewInvalidInputError`, do we expect the task to even have `StartTime` and `EndTime`?

gkcalat: Good point. Removed the errors. Thank you.

backend/src/apiserver/server/api_converter_test.go
gkcalat: Correct. It will throw an error at a later time. Generally, `toApi*` functions return `nil` when they encounter errors. The calling functions then check if the returned pointer is `nil` and returns an error.

backend/src/apiserver/server/api_converter_test.go
gkcalat: You are right. These test cases are completely unneccessary. I removed them and simplified the function too.

backend/src/apiserver/resource/resource_manager_test.go
```diff
@@ -2862,8 +2922,15 @@ func TestReportWorkflowResource_ScheduledWorkflowIDNotEmpty_NoExperiment_Success
 			CreatedAtInSec:          11,
 			ScheduledAtInSec:        11,
 			FinishedAtInSec:         0,
-			Conditions:              "Error",
+			Conditions:              "Unknown",
```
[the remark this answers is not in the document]
gkcalat: Yes. This is expected as we move to V2-style states. Both `Error` and `Unknown` map to `RuntimeStateUnspecified`.

backend/src/apiserver/resource/resource_manager.go
```diff
@@ -1247,18 +1256,22 @@ func (r *ResourceManager) ReportWorkflowResource(ctx context.Context, execSpec u
 		state = model.RuntimeState(string(exec.ExecutionPhase(model.RunTerminatingConditionsV1))).ToV2()
 	}
 	// If run already exists, simply update it
-	if _, err := r.GetRun(runId); err == nil {
-		if updateError := r.runStore.UpdateRun(runId, string(state.ToV1()), execStatus.FinishedAt(), execSpec.ToStringForStore(), state.ToString()); updateError != nil {
-			return util.Wrapf(err, "Failed to report a workflow for existing run %s during updating the run. Check if the run entry is corrupted", runId)
+	run, updateError := r.GetRun(runId)
+	if updateError == nil {
+		run.State = state
+		run.Conditions = string(state.ToV1())
```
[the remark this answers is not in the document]
gkcalat: `Conditions` is the old field that carries the same meaning. The name originates from Argo's `Conditions` field. Let me sync with you offline on this to make sure we sort this out.

backend/src/apiserver/resource/resource_manager.go
```diff
@@ -1208,20 +1207,30 @@ func (r *ResourceManager) DeleteJob(ctx context.Context, jobId string) error {
...
+	if err != nil {
+		return nil, util.Wrap(err, "Failed to create or update tasks")
+	}
+	return tasks, nil
+}
+
 // Reports a workflow CR.
 // This is called by the persistence agent to update runs.
-func (r *ResourceManager) ReportWorkflowResource(ctx context.Context, execSpec util.ExecutionSpec) error {
+func (r *ResourceManager) ReportWorkflowResource(ctx context.Context, execSpec util.ExecutionSpec) (util.ExecutionSpec, error) {
```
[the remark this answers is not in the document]
gkcalat: Thank you. Fixed.

backend/src/apiserver/server/api_converter.go
```diff
@@ -1255,6 +1262,15 @@ func toModelRun(r interface{}) (*model.Run, error) {
 		serviceAcc = apiRunV1.GetServiceAccount()
 	case *apiv2beta1.Run:
 		apiRunV2 := r
+		if temp, err := toModelTasks(apiRunV2.GetRunDetails().GetTaskDetails()); err == nil {
```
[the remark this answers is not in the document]
gkcalat: Good catch! Thank you. I fixed this.

backend/src/apiserver/server/api_converter.go
gkcalat: For `toModelTask`, we simply provide interface for converting from v1, v2, and argo types. All three are external to the backend.

For `toModelTasks`:
`[]*apiv2beta1.PipelineTaskDetail`: v2
`*apiv2beta1.Run`: I agree. It is unnecessary. Removed.
`[]*apiv1beta1.Task`: v1
`*util.Workflow`: this is convenient to use as argo's Node do not ave parent `runId`, `Namespace`, and `CreatedTimestamp`. This branch takes care of these fields and wraps around the last case.
`workflowapi.Nodes`: argo

backend/src/apiserver/storage/task_store.go
```diff
@@ -80,6 +97,32 @@ func (s *TaskStore) CreateTask(task *model.Task) (*model.Task, error) {
 	}
 	newTask.UUID = id.String()
 
+	if newTask.CreatedTimestamp == 0 {
+		if newTask.StartedTimestamp == 0 {
```
chensun, line 101: Asked the question before, and I'm still very confused, are there other places where we set create/started time? I recall the answer was yes? But then is it right that we have multiple sources to set created/started time? In what scenarios would we use one source than another?

backend/src/apiserver/resource/resource_manager.go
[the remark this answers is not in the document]
gkcalat: I agree. I propose to clean this up at a later stage because changing now may require updating tests.

backend/src/apiserver/server/api_converter.go
gkcalat: `[]*apiv1beta1.Task` is not being used as we don't have task-level status update in v1. I will add it later once I figure out usage of `TaskService` in v1.

backend/src/apiserver/server/api_converter_test.go
```diff
@@ -2306,3 +2314,2319 @@ func TestToApiRecurringRun(t *testing.T) {
...
+func Test_toModelTasks_v2(t *testing.T) {
+	argV2 := []*apiv2beta1.PipelineTaskDetail{
+		{
+			RunId:       "2",
+			TaskId:      "1",
+			DisplayName: "task",
+			CreateTime:  &timestamppb.Timestamp{Seconds: 4},
+			StartTime:   &timestamppb.Timestamp{Seconds: 5},
+			EndTime:     &timestamppb.Timestamp{Seconds: 6},
+			// ExecutorDetail *PipelineTaskExecutorDetail `protobuf:"bytes,7,opt,name=executor_detail,json=executorDetail,proto3" json:"executor_detail,omitempty"`
```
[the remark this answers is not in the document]
gkcalat: Thanks! Fixed.

backend/src/apiserver/server/api_converter_test.go
```diff
@@ -2306,3 +2314,2319 @@ func TestToApiRecurringRun(t *testing.T) {
...
+				"b2": {
+					ArtifactIds: []int64{4, 5, 6},
+				},
+			},
+			ParentTaskId: "8",
+			StateHistory: []*apiv2beta1.RuntimeStatus{
+				{
+					UpdateTime: &timestamppb.Timestamp{Seconds: 9},
+					State:      apiv2beta1.RuntimeState_PAUSED,
+					Error:      util.ToRpcStatus(util.NewInvalidInputError("Sample error2")),
```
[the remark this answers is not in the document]
gkcalat: Thanbk you. Fixed.

backend/src/apiserver/server/api_converter.go
```diff
@@ -2424,7 +2514,7 @@ func toApiRuntimeStatus(s *model.RuntimeStatus) *apiv2beta1.RuntimeStatus {
 func toApiRuntimeStatuses(s []*model.RuntimeStatus) []*apiv2beta1.RuntimeStatus {
 	statuses := make([]*apiv2beta1.RuntimeStatus, 0)
 	if s == nil {
-		return statuses
+		return nil
```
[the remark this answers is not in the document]
gkcalat: Thank you. Moved it up.

backend/src/apiserver/server/api_converter_test.go
```diff
@@ -2306,3 +2314,2319 @@ func TestToApiRecurringRun(t *testing.T) {
...
+		},
+		{
+			State: model.RuntimeStateCancelling,
+		},
+		{
+			Error: util.NewInvalidInputError("Invalid input: %s", "sample value"),
+		},
+		{
+			UpdateTimeInSec: 100,
+		},
```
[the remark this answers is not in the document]
gkcalat: This is a good point. Removed.

backend/src/apiserver/storage/task_store.go
gkcalat: Let's clarify:
`CreatedTimestamp` - time of creating a task entry in DB
`StartedTimestamp` - time of starting an execultion (pod?)
`FinishedTimestamp` - end of execution
`FinishedTimestamp` - `StartedTimestamp` = execution duration (compute time)
`StartedTimestamp` - `CreatedTimestamp` = queue time

The constraint we have:
`StartedTimestamp` <= `StartedTimestamp` <= `FinishedTimestamp`

Tasks can be created when we report workflows or via `TaskService` (reports a individual task). Argo's workflows don't have task-level `CreatedTimestamp`, but have `StartedTimestamp`. `TaskService` may create new tasks without timestamps. This logic take care of this situation and makes sure that we obey the constraint. 

These timestamps can be updated by Persistance Agent later via `CreateOrUpdateTasks` call.

backend/src/apiserver/server/report_server.go
```diff
@@ -30,52 +31,52 @@ type ReportServer struct {
...
-	if err != nil {
-		return nil, err
+		return nil, util.Wrap(err, "Failed to report task details")
 	}
 	return &empty.Empty{}, nil
 }
 
-func (s *ReportServer) ReportScheduledWorkflow(ctx context.Context,
-	request *apiv2beta1.ReportScheduledWorkflowRequest,
+func (s *ReportServer) ReportWorkflowV1(ctx context.Context,
```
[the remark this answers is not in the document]
gkcalat: This is correct. The v2 method is a part of v2 API. I propose to include this integration and deprecation to another PR where I will adjust the persistence agent.

backend/src/apiserver/server/api_converter_test.go
```diff
@@ -2306,3 +2314,2319 @@ func TestToApiRecurringRun(t *testing.T) {
...
+							Namespace:         "",
+							PipelineName:      "",
+							RunId:             "run2",
+							MLMDExecutionID:   "14",
+							CreatedTimestamp:  11,
+							StartedTimestamp:  12,
+							FinishedTimestamp: 13,
+							Fingerprint:       "",
+							Name:              "this is task",
+							ParentTaskId:      "task2",
```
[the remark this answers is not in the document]
gkcalat: This makes sense. I adjusted the test cases to make them closer to reality.

backend/src/apiserver/server/api_converter.go
gkcalat: Merged the `workflowapi.Nodes` and `*util.Workflow`

backend/src/apiserver/resource/resource_manager_test.go
```diff
@@ -1545,6 +1545,13 @@ func TestCreateRun_ThroughPipelineID(t *testing.T) {
 			ScheduledAtInSec:        5,
 			Conditions:              "Pending",
 			WorkflowRuntimeManifest: util.NewWorkflow(expectedRuntimeWorkflow).ToStringForStore(),
+			StateHistoryString:      "[{\"UpdateTimeInSec\":6,\"State\":\"PENDING\"}]",
+			StateHistory: []*model.RuntimeStatus{
```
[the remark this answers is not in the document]
gkcalat: This is a good point. I removed this field from the tests. However, I propose to keep the model as is. As we migrate to [`gorm` v2](#8852), we can drastically simplify this using built-in [serializers](https://gorm.io/docs/serializer.html).

chensun (approved):
/lgtm
/approve

Thanks for you attention to the feedback and the prompt responses.

PR #8871, opened 2023-02-20
feat(sdk): using component's pip_index_urls for Dockerfile generation. Fixes #8816

### LINKED ISSUE #8816: [feature] Support pip_index_urls in containerized python components
### Feature Area

/area sdk

/area components

So far the argument `pip_index_urls` is used only for lightweight python component. But when `target_image` is provided and therefore containerized python component can be generated with `kfp component build`, the argument is not considered - it is not a part of generated Dockerfile.
This is something that could be relatively easily (I think) implemented and I'd be happy to give it a try if there's interest.

Similarly as with lightweight python component, one might be using internal pypi index for internal libraries to be used in containerized components.

Yes. After running `kfp component build` modify generated Dockerfile (add --extra-index-url to the pip line), run docker build directly from commandline (not using `kfp` anymore) and then push the image.

---

Love this idea? Give it a 👍.

b4sus:
As described in issue #8816, pip_index_urls can be used for `pip install` command inside of generated `Dockerfile` - handy for containerized python components.
The PR is reusing same code for generating the index url options string, which is used for lightweight python component.


sdk/python/kfp/cli/component.py
```diff
@@ -214,6 +215,13 @@ def _load_components(self):
             raise sys.exit(1)
         logging.info(f'Using target image: {self._target_image}')
 
+        pip_index_urls = []
+        for comp in self._components:
+            if comp.pip_index_urls is not None:
+                pip_index_urls.extend(comp.pip_index_urls)
+        if len(pip_index_urls) > 0:
```
connor-mccarthy, line 222: Per [Google Python styleguide 2.14](https://google.github.io/styleguide/pyguide.html#214-truefalse-evaluations), use implicit true for checking non-empty sequences

```suggestion
        if pip_index_urls:
```

sdk/python/kfp/cli/component.py
```diff
@@ -214,6 +215,13 @@ def _load_components(self):
...
+            self._pip_index_urls = list(dict.fromkeys(pip_index_urls))
```
connor-mccarthy, line 223: Can you explain what's happening here? It seems like this converts a list to dict then back to list, but it's possible I'm missing something.

sdk/python/kfp/cli/component_test.py
```diff
@@ -443,6 +448,66 @@ def test_docker_file_is_created_correctly(self):
...
+        self.assert_file_exists_and_contains(
+            'Dockerfile',
+            textwrap.dedent('''\
+                # Generated by KFP.
+
+                FROM python:3.7
+
+                WORKDIR /usr/local/src/kfp/components
+                COPY runtime-requirements.txt runtime-requirements.txt
+                RUN pip install --index-url https://pypi.org/simple --trusted-host https://pypi.org/simple  --extra-index-url https://example.com/pypi/simple --trusted-host https://example.com/pypi/simple --no-cache-dir -r runtime-requirements.txt
```
connor-mccarthy, line 505: (not your code; feel free to ignore) Looks like string joiner implementation results in two spaces between the first entry and the second (before the first `--extra-index-url`).

connor-mccarthy (changes requested):
@b4sus, this looks great! Thank you for the contribution.

Just a few small nitpicks. Could you also add a release note to `./sdk/RELEASE.md` that describes this as a feature to support `pip index_urls` in Containerized Python Components?

sdk/python/kfp/cli/component.py
b4sus: It is a little trick to get rid of possible duplicates preserving order (dict guarantees insertion order since 3.7, see [here](https://docs.python.org/3/library/stdtypes.html?highlight=dict#dict.popitem).
I will remove this as it is a bit too defensive and it is not done for any other list arguments I've seen - to be consistent.

sdk/python/kfp/cli/component_test.py
b4sus: Fixed

sdk/python/kfp/cli/component.py
connor-mccarthy: Neat! Thank you for explaining and good thinking.

I actually think it makes sense to keep this deduping. I can imagine containerizing many components that each use the same single URL. We don't need to list that package registry multiple times. This is probably the most realistic use case, since combining the `pip_index_urls` of components that each use a different registry wouldn't really make sense.

And the order preserving is good, too. There's no logical total ordering across multiple components, but it's easy enough to preserve a partial ordering, so let's do it.

The entries in the requirements file (`runtime-requirements.txt`) are deduplicated (via `set()`) and have some of the same concerns/considerations that arise from combining the component parameters for multiple components, so there's one precedent for deduping.

WDYT?

connor-mccarthy: Neat! Thank you for explaining and good thinking.

I actually think it makes sense to keep this deduping. I can imagine containerizing many components that each use the same single URL. We don't need to list that package registry multiple times. This is probably the most realistic use case, since combining the `pip_index_urls` of components that each use a different registry wouldn't really make sense.

And the order preserving is good, too. There's no logical total ordering across multiple components, but it's easy enough to preserve a partial ordering, so let's do it.

The entries in the requirements file (`runtime-requirements.txt`) are deduplicated (via `set()`) and have some of the same concerns/considerations that arise from combining the component parameters for multiple components, so there's one precedent for deduping.

WDYT?

b4sus: You're right, the deduplication is actually necessary when more components in a file share the same pypi url, dockerfile needs it only once - what is I guess expected use-case - company having own pypi index.
I've put it back in :)
Preserving order is nice to have for unit tests, otherwise set would be fine (like in runtime-requirements case).

connor-mccarthy (approved):
/lgtm
/approve

Thank you, @b4sus! This is a valuable contribution.

connor-mccarthy:
/remove-lgtm
/remove-approve

Just caught this: looks like the release notes entry is under `2.0.0-beta.13`, which was already released yesterday. Can you bump the note to the "Current Version" section?

connor-mccarthy:
/lgtm
/approve

Thanks again, @b4sus!

PR #8886, opened 2023-02-23
feat(backend): add support for environment variables. Fixes #8704

### LINKED ISSUE #8704: feat(backend): Support environment variables

Linchin:
* Added support for environment variables at task level
* Added compiler test

**Note:**
* Need to update driver image at [argo.go](https://github.com/kubeflow/pipelines/blob/8d87bec4661fc3fb82a513f9ea37129a93f4e2ce/backend/src/v2/compiler/argocompiler/argo.go#L110).

Fixes #8704.

**Checklist:**


backend/src/v2/compiler/argocompiler/testdata/pipeline_with_env.yaml
```diff
@@ -0,0 +1,298 @@
+apiVersion: argoproj.io/v1alpha1
```
chensun, line 1: There's no change to the argo compiler, it dumps IR into metadata fields. So I'm not sure if this json to yaml compilation is worth adding a separate test set. And we probably don't want to grow this folder with every single feature unless there's some argo compiler specifics to test.

backend/src/v2/driver/driver.go
```diff
@@ -313,6 +313,12 @@ func makePodSpecPatch(
 		return "", fmt.Errorf("failed to make podSpecPatch: %w", err)
 	}
 
+	// Convert environment variables
+	userEnvVar := make([]k8score.EnvVar, 0)
+	for _, envVar := range container.GetEnv() {
+		userEnvVar = append(userEnvVar, k8score.EnvVar{Name: envVar.GetName(), Value: envVar.GetValue()})
```
chensun, line 319: Curious if env has a dynamic value, e.g. an output from an upstream component, what would `envVar.GetValue()` return?

samples/v2/pipeline_with_env.py
```diff
@@ -0,0 +1,60 @@
...
+name: Check env
+implementation:
+  container:
+    image: alpine
+    command:
+    - sh
+    - -c
+    - |
+      set -e -x
+      if [ "$ENV2" == "val2" ]; tjhen
```
chensun, line 37: typo: then

This is also an indicator the test wasn't even run.
You need to add to the list: https://github.com/kubeflow/pipelines/blob/d17916fc6a6c9ecee322466105561c0feb99662d/samples/test/config.yaml#L118-L128

Linchin: Thank you, just fixed this.

backend/src/v2/compiler/argocompiler/testdata/pipeline_with_env.yaml
Linchin: This is a great point. I will remove this test case.

backend/src/v2/driver/driver.go
Linchin: I'm not sure about this, but I don't think the function here resolves the dynamic value. It likely returns a placeholder,  because the input comes from `*pipelinespec.PipelineDeploymentConfig_PipelineContainerSpec`, which doesn't seem to have been changed.

chensun (approved):
/lgtm
/approve

PR #8910, opened 2023-03-01
chore(frontend): Regenerate TypeScript API client for the latest changes (add listAllRuns).

jlyaoyuli:
The main purpose of this PR is to generate the latest typescript API client to use the `listAllRuns` API in run list page.


zijianjoy:
/lgtm


Interestingly, the pipeline_spec field is changed from Struct to any here. I wonder how it impacts the frontend adopting v2 API.

PR #8917, opened 2023-03-02
feat(backend): support task level retry

Linchin:
Added support for task level retry.

**Checklist:**


backend/src/v2/compiler/argocompiler/container.go
```diff
@@ -121,6 +122,24 @@ func (c *workflowCompiler) addContainerDriverTemplate() string {
 			Resources: driverResources,
 		},
 	}
+	// Add retry policy
+	if retryPolicy != nil {
+		maxRetryCount := intstr.FromInt(int(retryPolicy.GetMaxRetryCount()))
+		// During the converstion from float64 to int, the fraction is discarded (truncated towards 0).
+		// See https://go.dev/ref/spec#Conversions
+		backOffFactor := intstr.FromInt(int(retryPolicy.GetBackoffFactor()))
```
gkcalat, line 130: I am probably missing something. Why do we need to use `int()` and  `intstr.FromInt()` for these parameters (`int32` and `double` with default values of `0`)?

gkcalat, line 130: We need to check the value of `retryPolicy.GetBackoffFactor()` and replace it with `2` if is set to `0.0`. Ref: https://github.com/kubeflow/pipelines/blob/master/api/v2alpha1/pipeline_spec.proto#L542-L544

backend/src/v2/compiler/argocompiler/container.go
```diff
@@ -121,6 +122,24 @@ func (c *workflowCompiler) addContainerDriverTemplate() string {
...
+		if maxRetryCount.IntValue() > 0 {
+			// If maxRetryCount == 0, that means retry policy is empty. In this case we leave the whole field empty.
+			t.RetryStrategy = &wfapi.RetryStrategy{
+				Limit: &maxRetryCount,
+				Backoff: &wfapi.Backoff{
+					Duration:    retryPolicy.GetBackoffDuration().String(),
+					Factor:      &backOffFactor,
+					MaxDuration: retryPolicy.GetBackoffMaxDuration().String(),
```
gkcalat, line 138: We need to check the value of `retryPolicy.GetBackoffMaxDuration()` and replace it with `1h` or `3600s` if it is nil or set to `0`. Ref: https://github.com/kubeflow/pipelines/blob/master/api/v2alpha1/pipeline_spec.proto#L546-L549

backend/src/v2/compiler/argocompiler/container.go
```diff
@@ -121,6 +122,24 @@ func (c *workflowCompiler) addContainerDriverTemplate() string {
...
+		maxRetryCount := intstr.FromInt(int(retryPolicy.GetMaxRetryCount()))
+		// During the converstion from float64 to int, the fraction is discarded (truncated towards 0).
+		// See https://go.dev/ref/spec#Conversions
+		backOffFactor := intstr.FromInt(int(retryPolicy.GetBackoffFactor()))
+		if maxRetryCount.IntValue() > 0 {
+			// If maxRetryCount == 0, that means retry policy is empty. In this case we leave the whole field empty.
+			t.RetryStrategy = &wfapi.RetryStrategy{
+				Limit: &maxRetryCount,
+				Backoff: &wfapi.Backoff{
+					Duration:    retryPolicy.GetBackoffDuration().String(),
```
gkcalat, line 136: Should we check if `retryPolicy.GetBackoffDuration()` is nil and replace with `0s`?

gkcalat (commented):
Thank you @Linchin!

Do you think we can test this functionality by checking if the compiler adds the necessary `RetryPolicy` to the argo spec? A few unit-tests with various values in `pipelinespec.PipelineTaskSpec_RetryPolicy` could be enough?

PR #8921, opened 2023-03-03
fix(backend): Deprecate experiment_id in Run API v2

gkcalat:

As we removed the parent resource from Run API URLs, this PR marks as deprecated the parent `experiment_id` fields when it's unnecessary. This assumes that `run_id` is unique, which [is true](https://github.com/kubeflow/pipelines/blob/e568148264a5166258bd8ff5d0f9e6c417c0cc9b/backend/src/apiserver/model/run.go#L208).

**Checklist:**


chensun (approved):
/lgtm

PR #8926, opened 2023-03-03
fix(backend): Allow runs and recurring runs without creator pipeline and pipeline version

gkcalat:

- Addresses #8930 
- Removed creation of a pipeline or pipeline version when creating a run or recurring run
- Simplified the logic
- Removed unused unit-tests
- [x] Manually tested

**Checklist:**


backend/src/apiserver/resource/resource_manager_test.go
```diff
@@ -1573,19 +1573,17 @@ func TestCreateRun_ThroughWorkflowSpecV2(t *testing.T) {
...
-			PipelineVersionId:    "123e4567-e89b-12d3-a456-426655440000",
-			PipelineName:         "hello-world-3",
+			PipelineName:         "hello-world-2",
 			PipelineSpecManifest: v2SpecHelloWorld,
 		},
 		RunDetails: model.RunDetails{
-			CreatedAtInSec:   5,
-			ScheduledAtInSec: 5,
+			CreatedAtInSec:   3,
+			ScheduledAtInSec: 3,
```
chensun, line 1581: Curious why do we have these diffs, including the timestamp changes as well as the `PipelineName` changes? It's not very clear to me how they are related to the code change.

gkcalat: Thank you for reviewing! I slightly adjusted the logic here.

When a run is created with manifest (pipeline spec or argo), the backend will try its best to find existing pipeline and pipeline version based on the manifest and run's name and namespace. Depending on what is found, the returned run might get existing `PipelineId`, `PipelineName`, and, optionally, `PipelineVersionId`.

Why is this important? To avoid duplicating manifests in `run_details` table and add references to the existing resources when it's possible. Users will be able to see the graph and manifest definition for both cases, but when we have a reference to a pipeline version, they will see the existing resource references.

To answer you question:
1. Timestamps are now changed since we no longer creating a new pipeline and a new pipeline version (-2 ticks in this case)
2. PipelineName is removed because no existing pipeline was found (I changed this since your last review to make it empty if no existing pipeline is found)

I added two test cases for submitting a run with manifest that matches to an existing (1) pipeline and (2) both pipeline and pipeline version. I also adjusted the comment to explain the logic here.

chensun: This is definitely not a slight adjustment. The bug we are trying to fix is due to an unintended behavior change, and goal is to simply restore the old behavior. We should not make yet another behavior change to fix an unintended behavior change, no matter how fancy that change could be. 
It's fine and appreciated to propose improvements. However, such changes should be conducted separately from bug fixing, let alone user-facing behavior change usually require a design review. 

Just to throw some quick thoughts/questions on your proposed change:
1. Why is it important to avoid such duplication? V1 has been doing the same for a long time, do we have any feedback this is causing issues, or have we done any analysis on how much storage would the duplication costs in general?
2. When users upload pipeline versions( templates), we don't do any find-and-dedupe. But we do this find-and-dedupe when users create runs using pipeline spec? What if there're multiple matches, which pipeline version should we link the run to?
3. A user submits a run using pipeline spec but the run was "smartly" linked to an identical pipeline version (template) the user may have forgot about. What if the pipeline version was deleted afterwards? Do we try searching for another match and update the link? Or do we removing the link? Or leave the broken link as-is?

chensun (changes requested):
See https://github.com/kubeflow/pipelines/pull/8926#discussion_r1125607596

backend/src/apiserver/resource/resource_manager_test.go
gkcalat: Thank you @chensun. This is a good point. I removed the logic.

backend/src/apiserver/resource/resource_manager.go
```diff
@@ -399,9 +399,14 @@ func (r *ResourceManager) CreateRun(ctx context.Context, run *model.Run) (*model
...
 	if pipelineVersion == nil {
-		pipelineVersion, wfTemplate, err = r.createPipelineFromSpecIfNoExisting(manifest, run.Namespace, run.DisplayName, run.Description)
-		if err != nil {
-			return nil, util.Wrap(err, "Failed to create a run due to error fetching pipeline version from pipeline spec")
+		if temp, err := template.New([]byte(manifest)); err != nil {
+			return nil, util.Wrap(err, "Failed to create a run due to error reading manifest")
+		} else {
+			wfTemplate = &temp
+		}
+		pipelineVersion = &model.PipelineVersion{
```
chensun, line 407: I think we can, and probably should, avoid creating this `model.PipelineVersion` object even though this is only in-memory represeentation, because all we need is its manifest part--This is a partially initialized object with `pipelineVersion.PipelineId`, `pipelineVersion. UUID`, and `pipelineVersion.Name` all being empty, and the `pipelineVersion.Description` which is initialized here but not consumed afterwards. 

Following [this description of v1 behavior](https://github.com/kubeflow/pipelines/blame/c4a2a35b5c8ceee8188fab715f348c40ee8b3e00/backend/src/apiserver/resource/resource_manager.go#L685-L690), the code would be cleaner and easier to follow if it's structured as follows:


```golang
if pipelineVersion != nil {
   manifest = pipelineVersion.PipelineSpec
   run.PipelineSpec.PipelineId = pipelineVersion.PipelineId
   run.PipelineSpec.PipelineVersionId = pipelineVersion.UUID
   ...
} else {
   // no need to create `model.PipelineVersion`
   manifest = run.PipelineSpec.PipelineSpecManifest || run.PipelineSpec.WorkflowSpecManifest // this is not valid go syntax, but you got the idea.
   ...
}
```

WDYT?

gkcalat: Thank you for taking another look at this! I further simplified the logic and placed it in a separate function as it's used in `CreateJob` method.

PR #8920, opened 2023-03-03
fix(backend): Revert the url changes in run API v 2

gkcalat:
**Description of your changes:**

- Removed the parent experiment from URLs. E.g. `/apis/v2beta1`~~/experiments/{experiment_id}~~`/runs`

**Checklist:**
- [x] The title for your pull request (PR) should follow our title convention. [Learn more about the pull request title convention used in this repository](https://github.com/kubeflow/pipelines/blob/master/CONTRIBUTING.md#pull-request-title-convention).

PR #8927, opened 2023-03-03
chore(frontend): Regenerate typescript API client to match the revert change in PR #8920

jlyaoyuli:
Update TS API client to match the change in https://github.com/kubeflow/pipelines/pull/8920.


zijianjoy:
/lgtm
/approve

Thank you Joe!

PR #8937, opened 2023-03-07
feat: add wait parameter to enable and disable functions in recurring…

andreafehrman:
- Add a `sleep_duration` parameter (defaulted to 0) to add a wait time for the `enable()` and `disable()` function within `recurring_run.py`

### DIFF
--- a/sdk/python/kfp/cli/recurring_run.py
+++ b/sdk/python/kfp/cli/recurring_run.py
@@ -15,6 +15,7 @@
 from typing import Any, Dict, List, Optional
 
 import click
+import time
 from kfp import client
 from kfp.cli import output
 from kfp.cli.utils import parsing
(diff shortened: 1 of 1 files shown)


connor-mccarthy (commented):
Thanks, @andreafehrman! For my own education, what is the current behavior without the sleep? Is there a race condition, failure, etc.?

Also, is there a benefit to to parameterizing `sleep_duration` v hard-coding a reasonable constant?

connor-mccarthy:
FYI, @andreafehrman, we can safely ignore `kubeflow-pipelines-sdk-execution-tests` failures

andreafehrman:
Hi @connor-mccarthy, unfortunately we weren't able to test due to mismatching versions on our side with the server version and v2 that has the enable and disable functionality. It looks like a post and then a get so the get could be returning an old state if the post is asynchronous which is why a wait could be necessary. Is this what you had in mind when adding the TODO regarding the wait? 

I

connor-mccarthy:
@andreafehrman, thanks for the response. In truth, I don't recall the motivation for the TODO (not a very good TODO, it turns out).

I'd like to investigate the issue before adding a `sleep` call, if possible. If we do add a `sleep` call, sounds good on not exposing the `sleep_duration` parameter to the user.

andreafehrman:
Ok no problem @connor-mccarthy, thanks for looking into this! We can put this issue on the back-burner for now, per further research if you think that'd be best. I can push to my branch in the meantime to alter the PR to just sleep if that would be helpful.

PR #8939, opened 2023-03-07
chore(backend): Remove API dependdency from filter and list packages. Closes #8934

### LINKED ISSUE #8934: chore(backend) Remove API dependency in backend apiserver's filter package
The api-server currently depends on v1beta1 API packed with regards to filtering. The current proposal is to convert the incoming filtering options to their internal representation and pass it down the line. As the API message has been changed in favor of [AIP-140](https://google.aip.dev/140) and [AIP-126](https://google.aip.dev/126) conventions and we still need to support both v1 and v2 API, this design is a good option.

Also mentioned previously:

1.
              Hi @gkcalat, @Linchin 
              I already brought up this topic in another PR (see [here](https://github.com/kubeflow/pipelines/pull/8480#issuecomment-1325214714) for context), but I want to comment on it one more time. 
              As it has been chosen that the backend is going to use just one of the `filter api` - because `v1.Filtet.proto` and `v2.Filetr.proto` are the same and because they are not directly used in any gRPC call, but serialized encoded (used behind an interface{}) - they are interchangeable. 
              So, my question is: isn't it better to move toward using                             **v2**:`github.com/kubeflow/pipelines/backend/api/v2beta1/go_client.Filter` now, instead of keeping the entire backend code depended on `v1.Filter`? Isn't it better to do it now, instead of later?

_Originally posted by @difince in https://github.com/kubeflow/pipelines/pull/8633#discussion_r1087895977_

2.
              @Linchin , please note that **v2** `ListExperiments` function  uses `api/v1beta1` **Filter**.  
              If you go deeper into this `validatelistOptions` function ... you will get to the  [filter.go](https://github.com/kubeflow/pipelines/blob/master/backend/src/apiserver/filter/filter.go) and [list.go](https://github.com/kubeflow/pipelines/blob/master/backend/src/apiserver/list/list.go) which both use 
              `api "github.com/kubeflow/pipelines/backend/api/v1beta1/go_client"`

_Originally posted by @difince in https://github.com/kubeflow/pipelines/pull/8480#discussion_r1030557680_

gkcalat:

- Closes #8934
- Removed API filter in `filter`
- Updated constructors of the filter object
- Added a method to update keys in the filter after creation
- Removed dependency on API packages in `list` package
- Refactored list handlers in `server` package
- Added a few unit tests to cover `v2beta1` API cases
- Fixed typos in the comments inside `filter.proto`
- [x] Manually tested `v2beta1` listing calls (Runs and Pipelines)

**Checklist:**


chensun (approved):
/lgtm
/approve

chensun:
Test failure will be solved by https://github.com/kubeflow/pipelines/pull/8951
There's a deadlock between the two, force merging this PR first.

PR #8978, opened 2023-03-13
test: Adding v2 initialization tests 

chensun:
Part of https://github.com/kubeflow/pipelines/issues/6168
New tests ran successfully
image

**Checklist:**


test/initialization-test/run_test.sh
```diff
@@ -57,8 +61,11 @@ fi
 GITHUB_REPO=kubeflow/pipelines
 BASE_DIR=/go/src/github.com/${GITHUB_REPO}
 JUNIT_TEST_RESULT=junit_InitializationTestOutput.xml
-TEST_DIR=backend/test/initialization
-
+if [ -n "$TEST_V2_API" ]; then
+  TEST_DIR=backend/test/v2/initialization
```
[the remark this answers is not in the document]
chensun: In that way, `go test ./...` on v1 test folder would also run v2 tests because it's a subdirectory. 
Technically subdirectories can still be filtered out, but I don't feel like overcomplicating the command, which could be error-prone.

backend/src/common/client/api_server/experiment_client_fake.go
```diff
@@ -9,7 +23,6 @@ import (
 )
 
 const (
-	ExperimentForDefaultTest     = "EXPERIMENT_DEFAULT"
```
gkcalat, line 12: To confirm: was this an unused constant?

chensun: correct

backend/src/common/client/api_server/v2/experiment_client.go
```diff
@@ -0,0 +1,217 @@
...
+import (
+	"fmt"
+	"time"
+
+	"github.com/go-openapi/runtime"
+	"github.com/go-openapi/strfmt"
+	apiclient "github.com/kubeflow/pipelines/backend/api/v2beta1/go_http_client/experiment_client"
+	params "github.com/kubeflow/pipelines/backend/api/v2beta1/go_http_client/experiment_client/experiment_service"
+	model "github.com/kubeflow/pipelines/backend/api/v2beta1/go_http_client/experiment_model"
+	"github.com/kubeflow/pipelines/backend/src/common/client/api_server"
```
gkcalat, line 26: Would it be better to move v1 clients into `common/client/v1` and v2 clients into `/common/client/v2`, while keeping the shared utils in `/common/client`?

gkcalat: As we are going to support both v1 and v2 API once integration is completed, should we make a single package for both API clients? If yes, we may want to store v2-related code in the same folder with the v1?

backend/src/common/client/api_server/v2/experiment_client.go
```diff
@@ -0,0 +1,217 @@
...
+	Get(params *params.GetExperimentParams) (*model.V2beta1Experiment, error)
+	List(params *params.ListExperimentsParams) ([]*model.V2beta1Experiment, int, string, error)
+	ListAll(params *params.ListExperimentsParams, maxResultSize int) ([]*model.V2beta1Experiment, error)
+	Archive(params *params.ArchiveExperimentParams) error
+	Unarchive(params *params.UnarchiveExperimentParams) error
+}
+
+type ExperimentClient struct {
+	apiClient      *apiclient.Experiment
+	authInfoWriter runtime.ClientAuthInfoWriter
```
gkcalat, line 48: Can we remove it? What is it being used for?

backend/src/common/client/api_server/v2/experiment_client.go
```diff
@@ -0,0 +1,217 @@
...
+	ctx, cancel := context.WithTimeout(context.Background(), apiServerDefaultTimeout)
+	defer cancel()
+
+	// Make service call
+	parameters.Context = ctx
+	response, err := c.apiClient.ExperimentService.CreateExperiment(parameters)
+	if err != nil {
+		return nil, util.NewUserError(err,
+			fmt.Sprintf("Failed to create experiment. Params: '%+v'. Body: '%+v'", parameters, parameters.Body),
+			fmt.Sprintf("Failed to create experiment '%v'", parameters.Body.DisplayName))
```
gkcalat, line 94: NIT: combine these as `fmt.Sprintf("Failed to create experiment '%v'. Params: '%+v'. Body: '%+v'", parameters.Body.DisplayName, parameters, parameters.Body)`

backend/src/common/client/api_server/v2/experiment_client.go
```diff
@@ -0,0 +1,217 @@
...
+	ctx, cancel := context.WithTimeout(context.Background(), apiServerDefaultTimeout)
+	defer cancel()
+
+	// Make service call
+	parameters.Context = ctx
+	response, err := c.apiClient.ExperimentService.GetExperiment(parameters)
+	if err != nil {
+		return nil, util.NewUserError(err,
+			fmt.Sprintf("Failed to get experiment. Params: '%+v'", parameters),
+			fmt.Sprintf("Failed to get experiment '%v'", parameters.ExperimentID))
```
gkcalat, line 112: NIT: combine together?

backend/src/common/client/api_server/v2/experiment_client.go
```diff
@@ -0,0 +1,217 @@
...
+
+	return nil
+}
+
+func (c *ExperimentClient) ListAll(parameters *params.ListExperimentsParams, maxResultSize int) (
+	[]*model.V2beta1Experiment, error) {
+	return listAllForExperiment(c, parameters, maxResultSize)
+}
+
+func listAllForExperiment(client ExperimentInterface, parameters *params.ListExperimentsParams,
```
gkcalat, line 158: For my education, why do we need `listAllForExperiment` function and the ` to to wrap this function?

backend/src/common/client/api_server/v2/experiment_client.go
chensun: > Would it be better to move v1 clients into common/client/v1 and v2 clients into /common/client/v2, while keeping the shared utils in /common/client?

Yes, I can do that, it would just produce more diff with not very obvious benefits IMO. But I don't have a strong opinion. I can make the change.

> As we are going to support both v1 and v2 API once integration is completed, should we make a single package for both API clients? If yes, we may want to store v2-related code in the same folder with the v1?

We will support both v1 and v2 API in the foreseeable near term, and deprecate v1 API later. And the clients are only used for testing only, so I don't think it's worth the effort to combine them into a single package.

gkcalat (commented):
Thank you @chensun!
I left a few comments below.

backend/src/common/client/api_server/v2/experiment_client.go
chensun: Removed. Will revisit whether to add/remove this for other APIs later.

backend/src/common/client/api_server/v2/experiment_client.go
chensun: Separating clients (even only experiment client) to v1 and v2 folder creates diffs for the entire integration test suite. I'll keep this PR clean to initialization test at this moment, and rethink whether to move v1 clients when adding v2 integration tests.

test/initialization-test/run_test.sh
```diff
@@ -25,6 +25,7 @@ usage()
     echo "usage: run_test.sh
     --results-gcs-dir GCS directory for the test results. Usually gs://<project-id>/<commit-sha>/initialization_test
     [--namespace      k8s namespace where ml-pipelines is deployed. The tests run against the instance in this namespace]
+    [--test-v2-api run test using v2 API]
```
Linchin, line 28: nit: make the space consistent with other options.

backend/src/common/client/api_server/v2/experiment_client.go
Linchin, line 1: Just wondering, would it be better if we create a folder `api_server_v2` that is parallel to `api_server` to store all the clients?

backend/test/v2/test_utils.go
```diff
@@ -0,0 +1,96 @@
...
+	api_server "github.com/kubeflow/pipelines/backend/src/common/client/api_server/v2"
+	"github.com/pkg/errors"
+	"github.com/stretchr/testify/assert"
+	"k8s.io/client-go/tools/clientcmd"
+	clientcmdapi "k8s.io/client-go/tools/clientcmd/api"
+)
+
+func WaitForReady(namespace string, initializeTimeout time.Duration) error {
+	operation := func() error {
+		response, err := http.Get(fmt.Sprintf("http://ml-pipeline.%s.svc.cluster.local:8888/apis/v2beta11/healthz", namespace))
```
Linchin, line 36: Please correct me if I'm wrong, is there supposed to be two 1's  in the path `/v2beta11/`?

backend/test/v2/test_utils.go
```diff
@@ -0,0 +1,96 @@
...
+
+func GetClientConfig(namespace string) clientcmd.ClientConfig {
+	loadingRules := clientcmd.NewDefaultClientConfigLoadingRules()
+	loadingRules.DefaultClientConfig = &clientcmd.DefaultClientConfig
+	overrides := clientcmd.ConfigOverrides{Context: clientcmdapi.Context{Namespace: namespace}}
+	return clientcmd.NewInteractiveDeferredLoadingClientConfig(loadingRules,
+		&overrides, os.Stdin)
+}
+
+func GetExperiment(name string, description string, namespace string) *experiment_model.V2beta1Experiment {
```
Linchin, line 64: This function doesn't seem to carry out the logic that retrieves experiments from the server. Is it intentional at this stage?

test/initialization-test/run_test.sh
chensun: Good catch. Fixed.

backend/test/v2/test_utils.go
chensun: Looks like this method is not used yet, removed it from the PR. 
To answer your question, looking at the v1 integration test, it's intended to "create" an API model object. So maybe the name is a bit misleading.

backend/test/v2/test_utils.go
chensun: Good catch. Fixed.

backend/src/common/client/api_server/v2/experiment_client.go
chensun: I don't see much difference between the two. Having a `v2` folder seems to be more common in our current code base. Also per https://github.com/kubeflow/pipelines/pull/8978/files/2f6e940718a1ebf469a5ac0834279c738285fefe#r1135920013, we might move v1 clients to a parallel folder later.

backend/test/v2/test_utils.go
Linchin: I see. Maybe we can rename this function and add a test for `GetExperiment()` method.

backend/src/common/client/api_server/v2/experiment_client.go
Linchin: Good point. Also either way we will have to rename/move the folder when we deprecate v1.

gkcalat:
/lgtm
/approve

Thank you, @chensun!

PR #8981, opened 2023-03-13
feat: add NodeSelector message to kfp-kubernetes proto

connor-mccarthy:

**Checklist:**
- [x] The title for your pull request (PR) should follow our title convention. [Learn more about 
the pull request title convention used in this repository](https://github.com/kubeflow/pipelines/blob/master/CONTRIBUTING.md#pull-request-title-convention).


connor-mccarthy:
/assign @gkcalat

@gkcalat, when you review, can you let me know if you think you the `NodeSelector` message ought to be wrapped in an outer parent message? I'm curious if you think it's likely we have future features that should be grouped alongside `NodeSelector`.

FYI, @chensun

gkcalat (approved):
Thank you @connor-mccarthy.

`NodeSelector` is located [at the top level](https://pkg.go.dev/k8s.io/api/core/v1#PodSpec) of `PodSpec`. I believe the current implementation is sufficient.

If we get user requests for [Affinity](https://pkg.go.dev/k8s.io/api/core/v1#Affinity), it can be added a separate message. It is a lot more functional compared to NodeSelector, but will also require more knowledge about compute cluster from SDK users.

PR #8983, opened 2023-03-13
feat(backend): support yaml with platform-specific specs

### LINKED ISSUE #8758: feat(backend): Accept pipeline YAML that contains platform-specific sections

Linchin:

* Backend reads the additional YAML documents with platform specific specs
* Added tests to verify that backend reads the yaml correctly
* Moved pipeline string for template testing into a file

Fixes #8758 
**Checklist:**


backend/src/apiserver/template/v2_template.go
```diff
@@ -217,3 +255,16 @@ func (t *V2Spec) RunWorkflow(modelRun *model.Run, options RunWorkflowOptions) (u
...
 	return executionSpec, nil
 }
+
+func isKubernetesExecutorConfig(template []byte) bool {
+	var platformSpec pipelinespec.PlatformSpec
+	templateJson, err := yaml.YAMLToJSON(template)
+	if err != nil {
+		return false
+	}
+	err = protojson.Unmarshal(templateJson, &platformSpec)
```
gkcalat, line 265: I believe `yaml.YAMLToJSON(template)` will take the first part of the yaml file. So I expect this function always fail for input like:
```
<YAML1>
---
<YAML2>
```

This is probably expected, but we might need to add comment explaining this behavior.

backend/src/apiserver/template/v2_template.go
```diff
@@ -107,37 +110,72 @@ func (t *V2Spec) GetTemplateType() TemplateType {
...
+		// this is unexpected, cannot convert proto message to JSON
+		return nil
+	}
+
+	bytesExecutorConfig, err := protojson.Marshal(t.platformSpec)
 	if err != nil {
 		// this is unexpected, cannot convert proto message to JSON
 		return nil
 	}
+	bytes := append(bytesSpec, bytesExecutorConfig...)
```
gkcalat, line 178: To better understand how it works: if I take `string(bytes)`, would I get something

```
<YAML1>
---
<YAML2>
```

or 

```
 YAML1YAML2
```

Should we prefer the former over the later? WDYT?

go.mod
gkcalat, line 1: You might need to regenerate the license files due to this change.

backend/src/apiserver/template/template_test.go
```diff
@@ -300,3 +246,51 @@ func TestModelToCRDTrigger_Cron(t *testing.T) {
...
+	if err != nil {
+		t.Error(err)
+	}
+	return string(res)
+}
+
+func TestIsKubernetesExecutorConfig(t *testing.T) {
+	template := loadYaml(t, "testdata/pipeline_with_volume.yaml")
+	splitTemplate := strings.Split(template, "\n---\n")
+	assert.True(t, isKubernetesExecutorConfig([]byte(splitTemplate[1])))
```
gkcalat, line 261: Can we also add `assert.False` to make sure the function works both ways.

backend/src/apiserver/template/v2_template.go
```diff
@@ -107,37 +110,72 @@ func (t *V2Spec) GetTemplateType() TemplateType {
...
+	decoder := goyaml.NewDecoder(bytes.NewReader(template))
+	for {
+		var value map[string]interface{}
+		// Break at end of file
+		if decoder.Decode(&value) != nil {
+			break
+		}
+		valueBytes, err := goyaml.Marshal(value)
+		if err != nil {
+			return nil, util.NewInvalidInputErrorWithDetails(ErrorInvalidPipelineSpec, fmt.Sprintf("unable to marshal this yaml document: %s", err.Error()))
```
gkcalat, line 123: I am probably wrong, but I expected something like this:
```golang
for _, templateSpec := strings.Split(string(template), "\n---\n") {
  if isPipelineSpec([]byte(templateSpec)){
      // <throw error if PipelineSpec has been already parse (duplicate PipelineSpec)>
      // <parse into *PipelineSpec>
  } else if isKubernetesExecutorConfig([]byte(templateSpec)){
      // <throw error if PlatformSpec has been already parse (duplicate PlatformSpec)>
      // <parse into *PlatformSpec>
  }
}
// <throw error if PipelineSpec has not been found in the input string>
```

gkcalat (commented):
Thank you @Linchin!

I left a few comments (some of them may be important)

backend/src/apiserver/template/v2_template.go
Linchin: I think both ways can parse the input of multiple yaml files, but I think you have a good point that we should verify that both inputs are unique. I will change the code accordingly.

backend/src/apiserver/template/v2_template.go
Linchin: This function is only called with individual yaml docs, because the input is already split in `NewV2SpecTemplate()`. I'm not sure if we still need to add comment in this case.

backend/src/apiserver/template/template_test.go
Linchin: Good idea. I will add this.

Linchin: done.

backend/src/apiserver/template/v2_template.go
Linchin: done.

backend/src/apiserver/template/template_test.go
```diff
@@ -300,3 +246,63 @@ func TestModelToCRDTrigger_Cron(t *testing.T) {
...
+
+// Verify that the V2Spec object created from Bytes() method is the same as the original object.
+// The byte slice may be slightly different from the original input during the conversion.
+func TestBytes_V2(t *testing.T) {
+	template := loadYaml(t, "testdata/pipeline_with_volume.yaml")
+	templateV2Spec, _ := New([]byte(template))
+	templateBytes := templateV2Spec.Bytes()
+	newTemplateV2Spec, err := New(templateBytes)
+	assert.Nil(t, err)
+	assert.Equal(t, templateV2Spec, newTemplateV2Spec)
```
gkcalat, line 307: If I got it right, the purpose of `Bytes()` method is to return the []byte of the original spec YAML?

Then, we need to compare `assert.Equal(t, template, string(templateBytes))`

backend/src/apiserver/template/v2_template.go
gkcalat: Thank you adding the errors!

Could you also clarify why do we need to use:
```golang
decoder := goyaml.NewDecoder(bytes.NewReader(template))
for {
	var value map[string]interface{}
	// Break at end of file
	if decoder.Decode(&value) != nil {
		break
	}
...
}
```
instead of this:

```golang
for _, templateSpec := strings.Split(string(template), "\n---\n") {
      templateBytes := []byte(templateSpec)
...
}
```

backend/src/apiserver/template/v2_template.go
gkcalat: That makes sense. My comment was a nit.

backend/src/apiserver/template/template_test.go
Linchin: I understand that it would be best to have `Bytes()` return the exact same content as the spec YAML, but with the marshaling and unmarshaling, it is quite inevitable that there will be some small changes, such as single quote being replaced by double quote. That's why I resorted to comparing the unmarshaled struct. I also think it's good practice to compare the struct due to the flakiness of the serialized format.

backend/src/apiserver/template/v2_template.go
```diff
@@ -26,11 +27,13 @@ import (
...
 	"google.golang.org/protobuf/encoding/protojson"
+	goyaml "gopkg.in/yaml.v2"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 )
 
 type V2Spec struct {
-	spec *pipelinespec.PipelineSpec
+	spec         *pipelinespec.PipelineSpec
+	platformSpec *pipelinespec.PlatformSpec
 }
```
gkcalat, line 37: It seems that we have integration tests checking for the content of V2Spec, so they also need to be updated.

Linchin: Thank you, I just fixed it.

backend/src/apiserver/template/v2_template.go
gkcalat: Thank you, @Linchin!

gkcalat:
/lgtm

Thank you, @Linchin!

backend/src/apiserver/template/v2_template.go
```diff
@@ -217,3 +271,16 @@ func (t *V2Spec) RunWorkflow(modelRun *model.Run, options RunWorkflowOptions) (u
 	executionSpec.SetPodMetadataLabels(util.LabelKeyWorkflowRunId, options.RunId)
 	return executionSpec, nil
 }
+
+func isKubernetesExecutorConfig(template []byte) bool {
```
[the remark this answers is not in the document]
gkcalat: What is the purpose of making sure the `kubernetes` exists? As we allow pipeline spec without platform spec here, we can probably let it pass with a non-kubernetes platform spec. WDYT?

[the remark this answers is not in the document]
chensun: We don't need to make sure `kubernetes` exists, but we do need to prevent mis-recognize `other` platform specs as `kubernetes` spec. The current change in the PR would treat any platform spec as `kubernetes` spec.

backend/src/apiserver/template/template_test.go
```diff
@@ -300,3 +246,76 @@ func TestModelToCRDTrigger_Cron(t *testing.T) {
...
+	protojson.Unmarshal(expectedSpecJson, &expectedSpec)
+	expectedTemplate := &V2Spec{
+		spec: &expectedSpec,
+	}
+	templateV2Spec, err := New([]byte(template))
+	assert.Nil(t, err)
+	assert.Equal(t, expectedTemplate, templateV2Spec)
+}
+
+func TestNewTemplate_WithExecutorConfig(t *testing.T) {
```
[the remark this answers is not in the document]
Linchin: Good point, I will rename the function.

backend/src/apiserver/template/v2_template.go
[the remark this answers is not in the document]
Linchin: A question that may be unrelated: should we keep all platforms' configs in memory, or only pick out the k8s related one?

backend/src/apiserver/template/v2_template.go
```diff
@@ -107,43 +110,94 @@ func (t *V2Spec) GetTemplateType() TemplateType {
...
+			}
+			match, _ := regexp.MatchString("[a-z0-9]([-a-z0-9]*[a-z0-9])?(\\.[a-z0-9]([-a-z0-9]*[a-z0-9])?)*", spec.GetPipelineInfo().GetName())
+			if !match {
+				return nil, util.NewInvalidInputErrorWithDetails(ErrorInvalidPipelineSpec, "invalid v2 pipeline spec: name should consist of lower case alphanumeric characters, '-' or '.', and must start and end with an alphanumeric character")
+			}
+			if spec.GetRoot() == nil {
+				return nil, util.NewInvalidInputErrorWithDetails(ErrorInvalidPipelineSpec, "invalid v2 pipeline spec: root component is empty")
+			}
+			v2Spec.spec = &spec
+		} else if isKubernetesExecutorConfig(valueBytes) {
```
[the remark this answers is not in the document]
Linchin: Thank you, I just updated it.

backend/src/apiserver/template/v2_template.go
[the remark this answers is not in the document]
Linchin: I added a check for the key "Kubernetes".

backend/src/apiserver/template/template_test.go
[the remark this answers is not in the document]
Linchin: done.

backend/src/apiserver/template/v2_template.go
```diff
@@ -217,3 +271,17 @@ func (t *V2Spec) RunWorkflow(modelRun *model.Run, options RunWorkflowOptions) (u
...
+	var platformSpec pipelinespec.PlatformSpec
+	templateJson, err := yaml.YAMLToJSON(template)
+	if err != nil {
+		return false
+	}
+	err = protojson.Unmarshal(templateJson, &platformSpec)
+	if err != nil {
+		return false
+	}
+	_, ok := platformSpec.Platforms["kubernetes"]
```
chensun, line 285: The function name should match what it checks, per this line of code, only `kubernetes` platform spec should return `True`, if that's the intention, the function name should reflect `kubernetes` as well.

Linchin: Thank you! I just updated the function name.

chensun (approved):
/lgtm
/approve

Thanks!

Issue #8997, opened 2023-03-16
[feature] Add support for ad hoc triggering of recurring runs to the KFP SDK


droctothorpe:
### Feature Area

/area sdk

### What feature would you like to see?

Ad hoc trigger support for recurring runs in the KFP SDK (and eventually the UI).

### What is the use case or pain point?

Currently, there is no way to ad hoc trigger a recurring run in KFP. A common user experience is creating a recurring run then validating it works as expected by creating a run. If the cron expression is set to once a day, it's not reasonable to wait until the cron expression is satisfied.

Argo Workflows makes this turn key in its UI and CLI. At the moment, this is not possible in KFP.

### Is there a workaround currently?

The only workaround currently is to take some of the parameters passed to `create_recurring_run` and manually pass them to `create_run_from_pipeline_package`. Part of the problem with this approach, in addition to the fact that it's not obvious or intuitive, is that there parameters of these functions vary in subtle ways. For example, `create_recurring_run` accepts an `experiment_id`, whereas `create_run_from_pipeline_func` accepts an `experiment_name`. 

---

Love this idea? Give it a 👍.

chensun:
Replied in https://github.com/kubeflow/pipelines/pull/8998#issuecomment-1473287386, and we should probably continue the discussion here, as it likely can't be solved by SDK alone.

> A common user experience is creating a recurring run then validating it works as expected by creating a run.

I wonder if the common scenario should be the other way around. Are we usually create a run first then turn it into a recurring once verified the run works?

droctothorpe:
I think validation is just one use case for ad hoc executions of recurring runs, but there are many others.

As a frame of reference, the Argo Workflows UI supports push-button ad hoc executions:

![](https://i.imgur.com/sWREakU.png)

The Argo Workflows CLI also supports ad hoc execution of CronWorkflows:

```bash
argo submit --from cronwf/my-cron-wf
```

chensun:
> I think validation is just one use case for ad hoc executions of recurring runs, but there are many others.

For my education, care to elaborate on some "others"?
"A has some feature thus we should have the same" isn't a convincing argument IMO. I'm not denying the value of the feature, but it just feels unclear to me, and I would like to learn more from a user's perspective.

Anyway, if you would like to proceed, please follow the [contribution guidelines](https://github.com/kubeflow/pipelines/blob/master/CONTRIBUTING.md#contribution-guidelines) for the next steps.

droctothorpe:
[...]

PR #8998, opened 2023-03-16
feat(sdk): implement trigger_recurring_run method

droctothorpe:
This PR implements a `trigger_recurring_run` method in the `Client` class of the SDK.

This method enables ad hoc triggering of recurring runs.

This PR also adds `experiment_id` as a parameter to `create_run_from_pipeline_func` and `create_run_from_pipeline_package` to provide a consistent UX across these functions and `create_recurring_run` and to avoid a potentially unnecessary call to `create_experiment` in the bodies of `create_run_from_pipeline_func` and `create_run_from_pipeline_package`.

We've implemented unit tests to provide coverage for the new functionality. We've also manually validated functionality against the latest release of KFP.

Here's an example of the UX:

```python
rr = client.create_recurring_run(
    pipeline_package_path="recurring_run.yaml",
    experiment_id=experiment_id,
    cron_expression="0 0 * * *", # Once a day at midnight.
    job_name="recurring_run",
    params={},
)

run = client.trigger_recurring_run(rr.id)
```

Relevant issue: https://github.com/kubeflow/pipelines/issues/8997

**Checklist:**

### DIFF
--- a/sdk/python/kfp/client/client.py
+++ b/sdk/python/kfp/client/client.py
@@ -18,6 +18,7 @@ import datetime
 import json
 import logging
 import os
+import random
 import re
 import tarfile
 import tempfile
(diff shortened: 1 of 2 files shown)


connor-mccarthy:
@droctothorpe, what do you think about whether or not this feature should be implemented at the BE API level, rather than via the SDK client only?

connor-mccarthy (commented):
@droctothorpe, what do you think about whether or not this feature should be implemented at the BE API level, rather than via the SDK client only?

droctothorpe:
> @droctothorpe, what do you think about whether or not this feature should be implemented at the BE API level, rather than via the SDK client only?

Before we submitted this PR, I said, "Brace yourselves, we might be asked to fix this on the BE instead!" 😂 😂 😂 

Figured a fix in the SDK would be less disruptive since it's basically downstream of everything, but happy to tackle on the BE if that's preferable.

connor-mccarthy:
cc @chensun on discussion of whether to implement in BE and, if so, how to coordinate

chensun:
This SDK implementation only mimics triggering an ad-hoc recurring run, but in reality the one-off run has nothing to with recurring run aside from sharing the same pipeline template and runtime parameters--on the UI for example, it won't have a link to the origin recurring run, and listing run instances of a recurring run won't include such one-off runs. 

So I think if we decided this is an important feature that we want native KFP support, we should do it on the API level. Otherwise, it would always be a client-side hack. Such hack is simple enough to be part of user code/library, it probably shouldn't be part of KFP SDK.

PR #8999, opened 2023-03-16
fix: support setting task dependencies via kfp.kubernetes.mount_pvc

connor-mccarthy:
Fixes a bug where `kfp.kubernetes.mount_pvc` doesn't set task dependencies when `pvc_name` is produced by an upstream task.

**Checklist:**
- [x] The title for your pull request (PR) should follow our title convention. [Learn more about the pull request 
title convention used in this repository](https://github.com/kubeflow/pipelines/blob/master/CONTRIBUTING.md#pull-request-title-convention).

### DIFF
--- a/sdk/python/kfp/components/pipeline_channel.py
+++ b/sdk/python/kfp/components/pipeline_channel.py
@@ -97,6 +97,14 @@ class PipelineChannel(abc.ABC):
         # so that serialization and unserialization remain consistent
         # (i.e. None => '' => None)
         self.task_name = task_name or None
+        from kfp.components import pipeline_context
+
+        default_pipeline = pipeline_context.Pipeline.get_default_pipeline()
+        if self.task_name is not None and default_pipeline is not None and default_pipeline.tasks:
+            self.task = pipeline_context.Pipeline.get_default_pipeline().tasks[
+                self.task_name]
+        else:
+            self.task = None
 
     @property
     def full_name(self) -> str:
--- a/kubernetes_platform/python/kfp/kubernetes/volume.py
+++ b/kubernetes_platform/python/kfp/kubernetes/volume.py
@@ -64,7 +64,7 @@ def CreatePVC(
 
 def mount_pvc(
     task: PipelineTask,
-    pvc_name: str,
+    pvc_name: Union[str, 'PipelineChannel'],
     mount_path: str,
 ) -> PipelineTask:
     """Mount a PersistentVolumeClaim to the task's container.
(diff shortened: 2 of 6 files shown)


sdk/python/kfp/components/pipeline_channel.py
```diff
@@ -97,6 +97,14 @@ def __init__(
         # so that serialization and unserialization remain consistent
         # (i.e. None => '' => None)
         self.task_name = task_name or None
+        from kfp.components import pipeline_context
+
+        default_pipeline = pipeline_context.Pipeline.get_default_pipeline()
+        if self.task_name is not None and default_pipeline is not None and default_pipeline.tasks:
+            self.task = pipeline_context.Pipeline.get_default_pipeline().tasks[
+                self.task_name]
```
chensun, line 105: This block of code feels a bit hacky to me. Maybe it's fine, I just cannot imagine if it may break in certain scenarios. 
A possible alternative, would it be possible to pass in the task object reference explicitly when constructing a PipelineChannel?

chensun (commented):
/lgtm

sdk/python/kfp/components/pipeline_channel.py
connor-mccarthy: I agree this is a bit awkward as currently implemented. I attempted constructing `PipelineChannel` from a `PipelineTask` instance, but ran into this troublesome section with `extract_pipeline_channels_from_any` [here](https://github.com/kubeflow/pipelines/blob/master/sdk/python/kfp/compiler/pipeline_spec_builder.py#L231).

`extract_pipeline_channels_from_any` needs to be able to create `PipelineChannel` instances from the "special string" used to represent the channel. This is possible in the proposed state where `PipelineChannel` accepts a `PipelineTask`, so long as the default pipeline (`Pipeline.get_default_pipeline()`) is available to look up the task instance from the task name which is extracted from the string.

The default pipeline is available for `extract_pipeline_channels_from_any` when it's called [here](https://github.com/kubeflow/pipelines/blob/328243b6ea51b86b01b29f9c4dde7e9fa6fc6c38/sdk/python/kfp/components/pipeline_task.py#L144), but not when it's called [here](https://github.com/kubeflow/pipelines/blob/328243b6ea51b86b01b29f9c4dde7e9fa6fc6c38/sdk/python/kfp/compiler/pipeline_spec_builder.py#L231), making it hard to modify `PipelineChannel` as suggested.

Of the two options, I think the current implementation is preferred.

I will merge per LGTM, but watch this thread in case it makes sense to follow up with another change.

PR #9001, opened 2023-03-16
chore(frontend): Integrate recurring run functions with KFP v2 API

jlyaoyuli:
Part of #8703 

The main purpose of this PR is integrate recurring run functionalities with v2 API.

Primary proto difference between v1 and v2

from `job` to `recurringRun`
from `id` to `recurring_run_id`
from `name` to `display_name`
`pipeline_spec` only store either `workflow_manifest` or `pipeline_manifest` in v2
`resource_reference `is deprecated in v2
`runtime_config` is moved out from `pipeline_spec`
**boolean** `enabled` is replaced with **enum** `V2beta1RecurringRunStatus`


frontend/src/components/Trigger.tsx
```diff
@@ -20,7 +20,7 @@ import FormControlLabel from '@material-ui/core/FormControlLabel';
 import MenuItem from '@material-ui/core/MenuItem';
 import * as React from 'react';
 import { classes, stylesheet } from 'typestyle';
-import { ApiTrigger } from '../apis/job';
```
zijianjoy, line 23: Is this trigger being used by both V1 and V2 recurring run? Do we want to preserve a separate trigger for V1?

jlyaoyuli: v1 and v2 trigger has the same functionality, so we don't have to preserve the trigger for v1.

frontend/src/pages/PipelineDetails.test.tsx
```diff
@@ -244,8 +265,8 @@ describe('PipelineDetails', () => {
               ),
             },
             {
-              displayName: testRun.run!.name,
```
zijianjoy, line 247: We might want to preserve V1 test because we still support V1 pipeline. Can you make a copy of this test for v1?

frontend/src/pages/PipelineDetails.test.tsx
```diff
@@ -258,11 +279,10 @@ describe('PipelineDetails', () => {
     'shows all runs breadcrumbs, and "Pipeline details" as page title when the pipeline ' +
       'comes from a recurring run spec that has an experiment',
     async () => {
-      testRecurringRun.resource_references = [
```
zijianjoy, line 261: Same for here and below, we need to keep V1 unit test.

frontend/src/pages/PipelineDetails.tsx
```diff
@@ -235,24 +239,49 @@ class PipelineDetails extends Page<{}, PipelineDetailsState> {
     if (origin) {
       const msgRunOrRecurringRun = origin.isRecurring ? 'recurring run' : 'run';
       try {
+        // TODO(jlyaoyuli): change to v2 API after v1 is deprecated
         if (origin.isRecurring) {
-          origin.recurringRun = await Apis.jobServiceApi.getJob(origin.recurringRunId!);
+          origin.v1RecurringRun = await Apis.jobServiceApi.getJob(origin.recurringRunId!);
+          origin.v2RecurringRun = await Apis.recurringRunServiceApi.getRecurringRun(
+            origin.recurringRunId!,
```
zijianjoy, line 246: Won't one of these API call fail as always? There should be only V1 or V2 recurring runs, and the API call with different version should fail.

frontend/src/pages/PipelineDetails.tsx
```diff
@@ -235,24 +239,49 @@ class PipelineDetails extends Page<{}, PipelineDetailsState> {
...
+        }
+
+        // v2 run or recurring run with pipeline_spec
+        let pipelineManifest = '';
+        if (origin.v2Run?.pipeline_spec) {
+          pipelineManifest = JsYaml.safeDump(origin.v2Run.pipeline_spec);
+        }
+        if (origin.v2RecurringRun?.pipeline_spec) {
+          pipelineManifest = JsYaml.safeDump(origin.v2RecurringRun.pipeline_spec);
         }
```
zijianjoy, line 273: This piece of logic is getting complicated. I suggest to abstract it to a function. Ultimately what it tries to do is simple: Get pipeline template from run or recurring run.

frontend/src/pages/PipelineDetails.tsx
```diff
@@ -235,24 +239,49 @@ class PipelineDetails extends Page<{}, PipelineDetailsState> {
...
+          templateStrFromOrigin = response.template || '';
+        }
+
+        // v2 run or recurring run with pipeline_spec
+        let pipelineManifest = '';
+        if (origin.v2Run?.pipeline_spec) {
+          pipelineManifest = JsYaml.safeDump(origin.v2Run.pipeline_spec);
+        }
+        if (origin.v2RecurringRun?.pipeline_spec) {
+          pipelineManifest = JsYaml.safeDump(origin.v2RecurringRun.pipeline_spec);
```
zijianjoy, line 272: It is better to be explicit about the order here, what is the priority between templates from run vs recurringrun?

frontend/src/pages/PipelineDetails.tsx
```diff
@@ -235,24 +239,49 @@ class PipelineDetails extends Page<{}, PipelineDetailsState> {
...
         } else {
-          origin.run = await Apis.runServiceApi.getRun(origin.runId!);
+          origin.v1Run = await Apis.runServiceApi.getRun(origin.runId!);
+          origin.v2Run = await Apis.runServiceApiV2.getRun(origin.runId!);
+        }
+
+        // v2 run or recurring run with pipeline_version_id
+        const pipelineVersionIdFromOrigin = origin.isRecurring
+          ? origin.v2RecurringRun?.pipeline_version_id
+          : origin.v2Run?.pipeline_version_id;
```
zijianjoy, line 256: We should remember the decision here so you don't need to check for both `origin.v2Run?.pipeline_spec` and `origin.v2RecurringRun?.pipeline_spec` in the logic below.

frontend/src/pages/RunDetails.tsx
```diff
@@ -626,7 +626,7 @@ class RunDetails extends Page<RunDetailsInternalProps, RunDetailsState> {
                     <div>
                       <DetailsTable
                         title='Run parameters'
-                        fields={workflowParameters.map(p => [p.name, p.value || ''])}
```
zijianjoy, line 629: Why deleting `|| ''`?

frontend/src/pages/RunDetailsRouter.tsx
```diff
@@ -58,35 +36,11 @@ export default function RunDetailsRouter(props: RunDetailsProps) {
     {},
   );
 
-  // Get Recurring run ID (if it is existed) for retrieve recurring run detail (PipelineSpec IR).
```
zijianjoy, line 61: Will this break `V1 run created by recurring run` functionality?

frontend/src/pages/RecurringRunDetailsV2.test.tsx
```diff
@@ -0,0 +1,469 @@
...
+import TestUtils from 'src/TestUtils';
+import { V2beta1RecurringRun, V2beta1RecurringRunStatus } from 'src/apisv2beta1/recurringrun';
+import { Apis } from 'src/lib/Apis';
+import { PageProps } from './Page';
+import { RouteParams, RoutePage, QUERY_PARAMS } from 'src/components/Router';
+import { shallow, ReactWrapper, ShallowWrapper } from 'enzyme';
+import { ButtonKeys } from 'src/lib/Buttons';
+
+describe('RecurringRunDetailsV2', () => {
+  let tree: ReactWrapper<any> | ShallowWrapper<any>;
```
zijianjoy, line 30: Prefer react testing library over enzyme.

frontend/src/pages/RecurringRunDetailsRouter.tsx
```diff
@@ -0,0 +1,80 @@
...
+
+  const pipelineVersionId = v2RecurringRun?.pipeline_version_id;
+
+  const { data: templateStrFromVersionId } = useQuery<string, Error>(
+    ['PipelineVersionTemplate', pipelineVersionId],
+    async () => {
+      if (!pipelineVersionId) {
+        return '';
+      }
+      // TODO(jlyaoyuli): temporarily use v1 API here, need to change in pipeline API integration.
```
zijianjoy, line 59: What is blocking us from using V2 API?

frontend/src/pages/RecurringRunDetailsRouter.tsx
```diff
@@ -0,0 +1,80 @@
...
+      const template = await Apis.pipelineServiceApi.getPipelineVersionTemplate(pipelineVersionId);
+      return template?.template || '';
+    },
+    { enabled: !!pipelineVersionId, staleTime: Infinity, cacheTime: Infinity },
+  );
+
+  const templateString = pipelineManifest ? pipelineManifest : templateStrFromVersionId;
```
zijianjoy, line 66: NIT: `const templateString = pipelineManifest ?? templateStrFromVersionId`

frontend/src/pages/RecurringRunDetailsV2.test.tsx
```diff
@@ -0,0 +1,469 @@
...
+          end_time: new Date(2018, 10, 9, 8, 7, 6),
+          start_time: new Date(2018, 9, 8, 7, 6),
+        },
+      },
+    };
+    getRecurringRunSpy.mockImplementation(() => cronTestRecurringRun);
+
+    render(
+      <CommonTestWrapper>
+        <RecurringRunDetailsV2 {...generateProps()} />
```
zijianjoy, line 134: We should use `RecurringRunDetailsRouter` instead of `RecurringRunDetailsV2` here, because RecurringRunDetailsRouter file doesn't have test coverage.

frontend/src/pages/RecurringRunDetailsRouter.tsx
jlyaoyuli: Because we haven't integrated pipeline/pipeline version call with v2 API.

test/frontend-integration-test/helloworld.spec.js
```diff
@@ -24,7 +24,7 @@ const runWithoutExperimentName = 'helloworld-2-' + Date.now();
 const runWithoutExperimentDescription =
   'test run without experiment description ' + runWithoutExperimentName;
 const waitTimeout = 5000;
-const outputParameterValue = 'Hello world in test';
```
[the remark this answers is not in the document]
jlyaoyuli: The failed e2e test makes me check the case again, and I found that the `outputParameterValue` should be changed to "hello world".

The run created by the example yaml file:

image

frontend/src/lib/Buttons.ts
```diff
@@ -523,7 +523,7 @@ export default class Buttons {
       [id],
       'Do you want to delete this recurring run config? This action cannot be undone.',
       useCurrentResource,
-      jobId => Apis.jobServiceApi.deleteJob(jobId),
+      recurringRunId => Apis.recurringRunServiceApi.deleteRecurringRun(recurringRunId),
```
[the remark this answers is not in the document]
jlyaoyuli: Yes, it works.

frontend/src/pages/ExperimentDetails.tsx
```diff
@@ -319,15 +320,17 @@ export class ExperimentDetails extends Page<{}, ExperimentDetailsState> {
       // Fetch this experiment's jobs
       try {
         // TODO: get ALL jobs in the experiment
-        const recurringRuns = await Apis.jobServiceApi.listJobs(
+        const recurringRuns = await Apis.recurringRunServiceApi.listRecurringRuns(
```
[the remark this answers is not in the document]
jlyaoyuli: Yes, it works. The response of v2 list recurring run API request include both v1 and v2 recurring runs.

frontend/src/pages/NewRunV2.tsx
```diff
@@ -295,31 +279,11 @@ function NewRunV2(props: NewRunV2Props) {
   const newRunMutation = useMutation((run: V2beta1Run) => {
     return Apis.runServiceApiV2.createRun(run);
   });
-  const newRecurringRunMutation = useMutation((apiJob: ApiJob) => {
-    return Apis.jobServiceApi.createJob(apiJob);
+  const newRecurringRunMutation = useMutation((recurringRun: V2beta1RecurringRun) => {
+    return Apis.recurringRunServiceApi.createRecurringRun(recurringRun);
   });
 
   const startRun = () => {
```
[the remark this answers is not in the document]
jlyaoyuli: We have experiment_id in `newRun` which is assign to `newRecurringRun` later.

frontend/src/pages/PipelineDetails.test.tsx
jlyaoyuli: For build breadcrumb part, we only use v2 getRun() / getRecurringRun() request, so I think we only need v2 test.

frontend/src/pages/PipelineDetails.tsx
jlyaoyuli: This is solved after wrapping the logics into a helper function!

frontend/src/pages/RunDetailsRouter.tsx
jlyaoyuli: Simply answer is no.

Actually, **v1 never use the recurring run id here**. Previously, existing run can be either v1 or v2 run. We use `getRecurringRunId` helper to **get v2 recurring run id** and use it to retrieve the pipeline spec in recurring run body. However, when I work on the run integration, v2 run is no longer has resource reference. The helper should be removed at that moment rather than mis-embedding with v1 run. Therefore, it is just a misconception that the change might affect v1 run.

frontend/src/pages/RecurringRunsManager.tsx
```diff
@@ -114,20 +114,22 @@ class RecurringRunsManager extends React.Component<RecurringRunListProps, Recurr
...
+  public _enabledCustomRenderer: React.FC<CustomRendererProps<V2beta1RecurringRunStatus>> = (
+    props: CustomRendererProps<V2beta1RecurringRunStatus>,
   ) => {
     const isBusy = this.state.busyIds.has(props.id);
     return (
       <BusyButton
-        outlined={props.value}
-        title={props.value === true ? 'Enabled' : 'Disabled'}
+        outlined={props.value === V2beta1RecurringRunStatus.ENABLED}
+        title={props.value === V2beta1RecurringRunStatus.ENABLED ? 'Enabled' : 'Disabled'}
```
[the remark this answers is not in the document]
jlyaoyuli: We call v2 listRecurringRuns() API only, which means that **no matter it is v1 or v2 recurring run, it will be in v2 format**. In addition, the `enabled` in v1 is change into `RecurringRunStatus` by the backend implementation. Therefore, the current logic is also applicable to v1 use case.

frontend/src/pages/RecurringRunDetailsV2.tsx
```diff
@@ -0,0 +1,206 @@
...
+import { commonCss, padding } from 'src/Css';
+import { KeyValue } from 'src/lib/StaticGraphParser';
+import { formatDateString, errorToMessage, enabledDisplayStringV2 } from 'src/lib/Utils';
+import { triggerDisplayString } from 'src/lib/TriggerUtils';
+
+interface RecurringRunConfigState {
+  run: V2beta1RecurringRun | null;
+}
+
+class RecurringRunDetailsV2 extends Page<{}, RecurringRunConfigState> {
```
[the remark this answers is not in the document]
jlyaoyuli: Will create a backlog item for this implementation.

frontend/src/pages/RecurringRunDetailsV2.test.tsx
jlyaoyuli: Will create a backlog item for this implementation.

frontend/src/pages/RunDetails.tsx
jlyaoyuli: The main reason why I deleted this is trying to fix e2e test, though it's not the root cause. However, after checking the code base, I found that in the `Parameter` definition, `value` is an optional field. It means that we don't have to put `''` as a default input, so I didn't add it back.

Ref: https://github.com/kubeflow/pipelines/blob/master/frontend/src/third_party/mlmd/argo_template.ts#L212-L231

frontend/src/components/Trigger.tsx
```diff
@@ -42,13 +42,13 @@ import { logger } from 'src/lib/Utils';
...
   maxConcurrentRuns?: string;
   catchup?: boolean;
-  trigger?: ApiTrigger;
+  trigger?: V2beta1Trigger;
 };
 
 interface TriggerProps {
   initialProps?: TriggerInitialProps;
   onChange?: (config: {
-    trigger?: ApiTrigger;
```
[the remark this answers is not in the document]
jlyaoyuli: This is a brilliant idea!

frontend/src/pages/PipelineDetails.tsx
jlyaoyuli: No, both of the API call won't fail. The pipeline details here is like a router, so we should make both v1 and v2 API call and determine which version the run is, and then get the template string via either `pipeline_spec` or `pipeline_version_id`

frontend/src/lib/Buttons.ts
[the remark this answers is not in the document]
zijianjoy: Sorry, my question is whether V1 recurring runs are being deleted if we use V2 API deleteRecurringRun?

test/frontend-integration-test/helloworld.spec.js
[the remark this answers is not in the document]
zijianjoy: That is interesting, I think the run parameter `hello world` hasn't changed since the beginning: https://github.com/kubeflow/pipelines/blob/master/test/frontend-integration-test/helloworld.yaml#L25. Why does it cause issues only recently?

frontend/src/pages/RecurringRunsManager.tsx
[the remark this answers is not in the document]
zijianjoy: To my knowledge: If this is V1 Recurring run, and we use V2 API to get the recurring run, how is `run parameters` being returned by the backend? Are `Run Parameters` being returned within `RuntimeConfig`?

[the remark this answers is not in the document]
jlyaoyuli: This one is to be determined. Need more information from the backend team, but I think the required field `id`, `name`, `enabled`, are preserved in corresponding field in v2 API response.

frontend/src/pages/PipelineDetails.test.tsx
jlyaoyuli: If the parameters required in some logics are retrievable from both v1 and v2 API response, we then select v2.

chensun (approved):
/lgtm
/approve

Thanks!

PR #9004, opened 2023-03-17
feat(sdk): add experiment_id parameter to create run methods

droctothorpe:
Currently, `create_run_from_pipeline_func` and `create_run_from_pipeline_package` accept an `experiment_name` parameter and lookup the experiment ID with a backend call via `create_experiment`.

This tiny PR lets users pass the experiment ID directly as an optional alternative, circumventing the extra call, and improving alignment between these methods and `create_recurring_run`.

It also decouples the experiment handling logic from the run name handling logic. They were interleaved previously. 

**Checklist:**

### DIFF
--- a/sdk/python/kfp/client/client.py
+++ b/sdk/python/kfp/client/client.py
@@ -507,7 +507,7 @@ class Client:
             return result.pipelines[0].id
         elif len(result.pipelines) > 1:
             raise ValueError(
-                f'Multiple pipelines with the name: {name} found, the name needs to be unique'
+                f'Multiple pipelines with the name: {name} found, the name needs to be unique.'
             )
         return None
 
@@ -572,7 +572,7 @@ class Client:
         namespace = namespace or self.get_user_namespace()
         if experiment_id is None and experiment_name is None:
             raise ValueError(
-                'Either experiment_id or experiment_name is required')
+                'Either experiment_id or experiment_name is required.')
         if experiment_id is not None:
             return self._experiment_api.get_experiment(id=experiment_id)
         experiment_filter = json.dumps({
@@ -916,7 +916,7 @@ class Client:
         if all([interval_second, cron_expression
                ]) or not any([interval_second, cron_expression]):
             raise ValueError(
-                'Either interval_second or cron_expression is required')
+                'Either interval_second or cron_expression is required.')
         if interval_second is not None:
             trigger = kfp_server_api.ApiTrigger(
                 periodic_schedule=kfp_server_api.ApiPeriodicSchedule(
@@ -1025,6 +1025,7 @@ class Client:
         pipeline_root: Optional[str] = None,
         enable_caching: Optional[bool] = None,
         service_account: Optional[str] = None,
+        experiment_id: Optional[str] = None,
     ) -> RunPipelineResult:
         """Runs pipeline on KFP-enabled Kubernetes cluster.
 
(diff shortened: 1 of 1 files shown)


connor-mccarthy:
@gkcalat @chensun, I seem to recall some discussion around particular parts of the API having trouble with requests that specify an experiment ID. It's possible I'm misremembering. Could you let me know if this sounds familiar?

/ok-to-test

droctothorpe:
FWIW, the `run_pipeline` invocation receives an experiment ID, it's just preceded by an ID lookup via `create_experiment` when the experiment name is passed.

sdk/python/kfp/client/client.py
```diff
@@ -1020,6 +1020,7 @@ def create_run_from_pipeline_func(
         pipeline_func: base_component.BaseComponent,
         arguments: Optional[Dict[str, Any]] = None,
         run_name: Optional[str] = None,
+        experiment_id: Optional[str] = None,
```
chensun, line 1023: New parameters should be added to the end to not break existing users who might call the method using positional arguments.

sdk/python/kfp/client/client.py
```diff
@@ -1111,22 +1116,27 @@ def create_run_from_pipeline_package(
...
-            KF_PIPELINES_DEFAULT_EXPERIMENT_NAME, None)
-        overridden_experiment_name = os.environ.get(
-            KF_PIPELINES_OVERRIDE_EXPERIMENT_NAME, experiment_name)
-        if overridden_experiment_name != experiment_name:
-            warnings.warn(
-                f'Changing experiment name from "{experiment_name}" to "{overridden_experiment_name}".'
-            )
-        experiment_name = overridden_experiment_name or 'Default'
+
+        if not experiment_id:
```
chensun, line 1120: Should probably check that `experiment_id` and `experiment_name` can't be specified at the same time.

sdk/python/kfp/client/client.py
droctothorpe: Good call. Didn't think of that! Resolved.

sdk/python/kfp/client/client.py
droctothorpe: Resolved. Modeled after similar ValueErrors in `client.py`. Noticed that about half of the ValueErrors were missing closing periods so added that as well.

droctothorpe:
Addressed the yapf finding. Are the other test failures unrelated?

sdk/python/kfp/client/client.py
```diff
@@ -1104,29 +1108,39 @@ def create_run_from_pipeline_package(
...
-            KF_PIPELINES_OVERRIDE_EXPERIMENT_NAME, experiment_name)
-        if overridden_experiment_name != experiment_name:
-            warnings.warn(
-                f'Changing experiment name from "{experiment_name}" to "{overridden_experiment_name}".'
-            )
-        experiment_name = overridden_experiment_name or 'Default'
+
+        if (experiment_name is not None) and (experiment_id is not None):
+            raise ValueError(
+                'You cannot specify both experiment_name and experiment_id.')
```
chensun, line 1122: nit: Better also mention this in the docstring.

chensun (commented):
/lgtm

sdk/python/kfp/client/client.py
droctothorpe: Non-ironically love nits. Resolved!

chensun (approved):
/lgtm
/approve

PR #9024, opened 2023-03-22
fix(backend): backend accepts pipeline spec struct as well as nested

### LINKED ISSUE #9016: [backend] Backend api should accept pipelineSpec in both current format and a nested struct
### Environment

N/A
*  KFP version: 
Beta 0

*  KFP SDK version: 
N/A

Currently, the backend API V2 accepts pipeline spec as a struct (see [here](https://github.com/kubeflow/pipelines/blob/cc30d1aa85eb25fb20a6d2531956b99e5a9ec04d/backend/api/v2beta1/run.proto#L167-L168)). Since we want to support platform-specific features, we will turn this generic struct into a nested format:  `Struct{pipeline_spec: <Struct>, kubernetes_config: <Struct>, ...}`.

Related: #8983 

---

Linchin:
Updated the backend so the API can accept pipeline spec directly (see [here](https://github.com/kubeflow/pipelines/blob/8c91abdc0cb7e9f2c2bf7af66b2de02d66415b88/backend/api/v2beta1/run.proto#L167-L168)), as well as a sub struct, along with platform specific configs like this:
```
Struct{"pipelineSpec": <Struct>, "platforms": <Struct>}
```
Added unit test too.

fixes #9016 

**Checklist:**


backend/src/apiserver/server/api_util.go
```diff
@@ -114,11 +116,32 @@ func getJobIdFromResourceReferencesV1(resourceRefs []*apiv1beta1.ResourceReferen
...
+// Converts pipeline spec in the form of structpb.Struct to a yaml string.
+// Handles three possible cases:
+// 1) pipeline spec passed directly
+// 2) pipeline spec along with platform specific configs, each as a sub struct
+// 3) pipeline spec as a sub struct, with platform specific field empty
+func pipelineSpecStructToYamlString(s *structpb.Struct) (string, error) {
+	var bytes []byte
+	var err error
+	if _, ok := s.GetFields()["pipelineSpec"]; ok {
+		bytes, err = yaml.Marshal(s.GetFields()["pipelineSpec"])
```
gkcalat, line 128: We can remove the additional call to `s.GetFields()["pipelineSpec"]`:
```golang
	if spec, ok := s.GetFields()["pipelineSpec"]; ok {
		bytes, err = yaml.Marshal(spec)
```

backend/src/apiserver/server/api_util.go
```diff
@@ -114,11 +116,32 @@ func getJobIdFromResourceReferencesV1(resourceRefs []*apiv1beta1.ResourceReferen
...
-func protobufStructToYamlString(s *structpb.Struct) (string, error) {
-	bytes, err := yaml.Marshal(s)
-	if err != nil {
-		return "", util.Wrap(err, "Failed to convert a protobuf struct to a yaml string")
+// Converts pipeline spec in the form of structpb.Struct to a yaml string.
+// Handles three possible cases:
+// 1) pipeline spec passed directly
+// 2) pipeline spec along with platform specific configs, each as a sub struct
+// 3) pipeline spec as a sub struct, with platform specific field empty
+func pipelineSpecStructToYamlString(s *structpb.Struct) (string, error) {
```
gkcalat, line 124: Could you please add unit-tests to check the behavior for these three cases and some expected errors (nils, maybe)?

backend/src/apiserver/server/api_util.go
```diff
@@ -137,14 +160,57 @@ func GetResourceReferenceFromRunInterface(r interface{}) ([]*apiv1beta1.Resource
...
+// The string may contain multiple yaml documents, where there has to be one doc for pipeline spec,
+// and there may be a second doc for platform specific spec.
+// If platform spec is empty, then return pipeline spec directly. Else, return a struct with
+// pipeline spec and platform spec as subfields.
+func yamlStringToPipelineSpecStruct(s string) (*structpb.Struct, error) {
+	if s == "" {
+		return nil, nil
+	}
+	var pipelineSpec structpb.Struct
+	var platformSpec structpb.Struct
```
gkcalat, line 173: We pass both `pipelineSpec` and `platformSpec` as pointers. So, we can probably initialize them as `*structpb.Struct`?

backend/src/apiserver/server/api_util.go
```diff
@@ -137,14 +160,57 @@ func GetResourceReferenceFromRunInterface(r interface{}) ([]*apiv1beta1.Resource
...
+	hasPipelineSpec := false
+	hasPlatformSpec := false
```
gkcalat, line 175: NIT: if we initialize `pipelineSpec` and `platformSpec` as pointers, we might be able to remove these booleans and check if `pipelineSpec == nil` or `platformSpec == nil`

backend/src/apiserver/server/api_util.go
```diff
@@ -137,14 +160,57 @@ func GetResourceReferenceFromRunInterface(r interface{}) ([]*apiv1beta1.Resource
...
+		} else {
+			hasPipelineSpec = true
+			jsonBytes, err := yaml.YAMLToJSON([]byte(yamlString))
+			if err != nil {
+				return nil, util.Wrap(err, "Failed to convert pipelineSpec yaml string into a protobuf struct")
+			}
+			err = protojson.Unmarshal(jsonBytes, &pipelineSpec)
+			if err != nil {
+				return nil, util.Wrap(err, "Failed to convert pipelineSpec yaml string into a protobuf struct")
+			}
```
gkcalat, line 198: Should we throw an error if `pipelineSpec` has already been found in the YAML? Something like `util.NewInvalidInputError("Duplicate pipeline spec has been found in the provided YAML file")`? WDYT?

backend/src/apiserver/server/api_util.go
Linchin: Good catch! Just updated.

backend/src/apiserver/server/api_util.go
```diff
@@ -114,11 +116,32 @@ func getJobIdFromResourceReferencesV1(resourceRefs []*apiv1beta1.ResourceReferen
...
+func pipelineSpecStructToYamlString(s *structpb.Struct) (string, error) {
+	var bytes []byte
+	var err error
+	if _, ok := s.GetFields()["pipelineSpec"]; ok {
+		bytes, err = yaml.Marshal(s.GetFields()["pipelineSpec"])
+		if err != nil {
+			return "", util.Wrap(err, "Failed to convert pipeline protobuf struct to a yaml string")
+		}
+		if _, ok = s.GetFields()["platforms"]; ok {
+			bytesPlatforms, err := yaml.Marshal(s.GetFields()["platforms"])
```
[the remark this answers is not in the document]
Linchin: Good catch! Just updated.

backend/src/apiserver/server/api_util.go
Linchin: I initially did the same, but it looks like `protojson` does not support unmarshalling to a nil pointer. See example [here](https://go.dev/play/p/zu7W47kmOgk).

backend/src/apiserver/server/api_util.go
Linchin: Similarly, see [reply](https://github.com/kubeflow/pipelines/pull/9024#discussion_r1145251806).

backend/src/apiserver/server/api_util.go
```diff
@@ -137,14 +160,57 @@ func GetResourceReferenceFromRunInterface(r interface{}) ([]*apiv1beta1.Resource
...
+	hasPlatformSpec := false
+
+	yamlStrings := strings.Split(s, "\n---\n")
+	for _, yamlString := range yamlStrings {
+		if template.IsPlatformSpecWithKubernetesConfig([]byte(yamlString)) {
+			hasPlatformSpec = true
+			jsonBytes, err := yaml.YAMLToJSON([]byte(yamlString))
+			if err != nil {
+				return nil, util.Wrap(err, "Failed to convert pipelineSpec yaml string into a protobuf struct")
+			}
```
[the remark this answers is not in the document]
Linchin: I'm not sure if it is necessary, similar to the comments [here](https://github.com/kubeflow/pipelines/pull/9024#discussion_r1145256995), the yaml docs are created from the struct, which has only one `platforms` field.

backend/src/apiserver/server/api_util.go
Linchin: This is a really good point! I just added more tests.

backend/src/apiserver/server/api_util.go
```diff
@@ -137,14 +160,57 @@ func GetResourceReferenceFromRunInterface(r interface{}) ([]*apiv1beta1.Resource
...
+		return &pipelineSpec, nil
+	} else {
+		pipeline := &structpb.Struct{
+			Fields: map[string]*structpb.Value{
+				"pipelineSpec": structpb.NewStructValue(&pipelineSpec),
+				"platforms":    structpb.NewStructValue(&platformSpec),
+			},
+		}
+		return pipeline, nil
 	}
```
gkcalat, line 214: This is probably intended behavior, but I wanted to call this out. For a YAML file like this
```yaml
<pipelineSpec>
---
<empty or invalid (e.g. missing `kubernetes` key) platformSpec>
```
We will ge
```golang
return &pipelineSpec, nil
```
So we intentionally dropping data with `IsPlatformSpecWithKubernetesConfi()` and treat it as the old-style `pipelineSpec`.

backend/src/apiserver/server/api_util.go
```diff
@@ -114,11 +116,33 @@ func getJobIdFromResourceReferencesV1(resourceRefs []*apiv1beta1.ResourceReferen
...
+// Converts pipeline spec in the form of structpb.Struct to a yaml string.
+// Handles three possible cases:
+// 1) pipeline spec passed directly
+// 2) pipeline spec along with platform specific configs, each as a sub struct
+// 3) pipeline spec as a sub struct, with platform specific field empty
+// Note: this method does not verify the validity of pipeline spec.
+func pipelineSpecStructToYamlString(s *structpb.Struct) (string, error) {
+	var bytes []byte
+	var err error
+	if spec, ok := s.GetFields()["pipelineSpec"]; ok {
```
gkcalat, line 128: The key might need to be changed to `pipeline_spec`.

backend/src/apiserver/server/api_util.go
```diff
@@ -114,11 +116,33 @@ func getJobIdFromResourceReferencesV1(resourceRefs []*apiv1beta1.ResourceReferen
...
+		bytes, err = yaml.Marshal(spec)
+		if err != nil {
+			return "", util.Wrap(err, "Failed to convert pipeline protobuf struct to a yaml string")
+		}
+		if platforms, ok := s.GetFields()["platforms"]; ok {
```
gkcalat, line 133: The key might need to be changed to `platform_spec`.

backend/src/apiserver/server/api_util.go
Linchin: Thank you for pointing this out! I will update the code accordingly.

backend/src/apiserver/server/api_util.go
Linchin: That's right. With `platformSpec` empty, the backend cannot tell how `pipelineSpec` was passed in.

backend/src/apiserver/server/api_util.go
```diff
@@ -137,14 +161,57 @@ func GetResourceReferenceFromRunInterface(r interface{}) ([]*apiv1beta1.Resource
...
+			hasPlatformSpec = true
+			jsonBytes, err := yaml.YAMLToJSON([]byte(yamlString))
+			if err != nil {
+				return nil, util.Wrap(err, "Failed to convert platformSpec yaml string into a protobuf struct")
+			}
+			err = protojson.Unmarshal(jsonBytes, &platformSpec)
+			if err != nil {
+				return nil, util.Wrap(err, "Failed to convert platformSpec yaml string into a protobuf struct")
+			}
+		} else {
```
gkcalat, line 190: Should we also call `IsPipelineSpec` here? If `IsPlatformSpecWithKubernetesConfig` returns `false` instead of `true`, we will return error from `err = protojson.Unmarshal(jsonBytes, &pipelineSpec)`, which may be less convenient to debug for a user. WDYT?

Linchin: This was what I did initially, but we had test cases with minimum pipeline spec, which means we didn't check the validity of pipeline spec here. I'm not sure if we should have a behavior change here.

backend/src/apiserver/server/api_util_test.go
gkcalat, line 1: NIT1: it's better to have separate tests for each function in unit-testing. The issue with stochastic order of the fields after marshalling can be solved by using `assert.Contains()` for example.

NIT2: We probably need to test `yamlStringToPipelineSpecStruct` separately for valid cases 1) a single pipeline spec yaml, 2) pipeline and platform specs, 3) pipeline spec and empty or invalid platform spec; as well as invalid cases: 1) single platform spec, 2) invalid or empty pipeline spec with a platform spec. Current tests set it to consume the result of `pipelineSpecStructToYamlString()`, which makes the tests too fragile and less sensitive to specific changes.

chensun (commented):
+1 on @gkcalat's comments.

/lgtm

Thanks!

backend/src/apiserver/server/api_util_test.go
Linchin: Thank you for your suggestions!
For NIT 1: The issue with comparing marshalled strings is less about the stochastic order, but more about the special characters are being handled differently in the original string vs. marshalled string. For example, in the original string, line breaks are represented by `\n`, while in the marshalled string, line breaks are represented by a literal line break. The `assert.Equal()` function judges them to be different. Using `assert.Contains()` feels more suitable for situation such as checking whether an error message includes certain language, and it would take many `assert.Contains()`s to do a comprehensive check. I also understand that we can just check certain parts of the yaml (using `assert.Contains()`), but I'm not sure how that compares to the complete unmarshalled struct.

For NIT 2: I don't think we checked for validity of the pipeline spec at the api level. It is checked when it's being compiled and I think it's a good idea to keep it this way. And ideally the `yamlStringToPipelineSpecStruct()` only consumes the output of `pipelineSpecStructToYamlString ()`, so it's unlikely for these corner cases to exist.

chensun:
/test kubeflow-pipelines-samples-v2

Although not a required test for this PR, just to give it another try and see if there might be real issues we have missed.

PR #9058, opened 2023-03-28
chore(frontend): Regenerate TS client for backend changes on createPipelineVersion

jlyaoyuli:
Backend add the missing parameters `body` in createPipelineVersion. Frontend then need to regenerate new TS client.


chensun (approved):
/lgtm
/approve

PR #9060, opened 2023-03-29
fix(sdk): resolve kubernetes libary CVE

connor-mccarthy:

**Checklist:**


connor-mccarthy:
`kubeflow-pipelines-sdk-python311` test failure will be resolved by https://github.com/GoogleCloudPlatform/oss-test-infra/pull/1921

sdk/python/requirements.txt
```diff
@@ -1,8 +1,8 @@
 #
-# This file is autogenerated by pip-compile with python 3.9
-# To update, run:
+# This file is autogenerated by pip-compile with Python 3.7
```
connor-mccarthy, line 2: This should be 3.7, not 3.9

chensun: This doesn't really matter, as the file has no effect on the end users. Only SDK developers may possibly--but not necessarily, like I personally don't--rely on this file to replicate the exact dev environment with precise dependencies. If the developer uses a different Python version, this file may not even work. 
So the Python version here just gives the context of how this snapshot is generated. It doesn't necessarily mean developers must be on one version than the other.

chensun (approved):
/lgtm
/approve

PR #9064, opened 2023-03-30
fix(backend): fix UploadPipeline/UploadPipelineVersion v2 API returns v1 object.

chensun:
While integrating with v2 API, we noticed that UploadPipeline API always return v1 object (with resource references).
/cc @jlyaoyuli 

**Checklist:**

### DIFF
--- a/backend/src/apiserver/server/pipeline_upload_server.go
+++ b/backend/src/apiserver/server/pipeline_upload_server.go
@@ -70,6 +70,14 @@ type PipelineUploadServer struct {
 	options         *PipelineUploadServerOptions
 }
 
+func (s *PipelineUploadServer) UploadPipelineV1(w http.ResponseWriter, r *http.Request) {
+	s.uploadPipeline("v1beta1", w, r)
+}
+
+func (s *PipelineUploadServer) UploadPipeline(w http.ResponseWriter, r *http.Request) {
+	s.uploadPipeline("v2beta1", w, r)
+}
+
 // Creates a pipeline and a pipeline version.
 // HTTP multipart endpoint for uploading pipeline file.
 // https://www.w3.org/Protocols/rfc1341/7_2_Multipart.html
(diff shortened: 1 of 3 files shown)


gkcalat:
Thank you @chensun!

/lgtm
/approve

gkcalat (approved):

backend/src/apiserver/server/pipeline_upload_server.go
```diff
@@ -235,12 +260,15 @@ func (s *PipelineUploadServer) UploadPipelineVersion(w http.ResponseWriter, r *h
 
 	w.Header().Set("Content-Type", "application/json")
 	marshaler := &jsonpb.Marshaler{EnumsAsInts: false, OrigName: true}
-	createdPipelineVersion := toApiPipelineVersionV1(newPipelineVersion)
-	if createdPipelineVersion == nil {
-		s.writeErrorToResponse(w, http.StatusInternalServerError, util.NewInternalServerError(errors.New("Failed to convert internal pipeline version representation to its v1beta1 API counterpart"), "Failed to create a pipeline version due to error converting it to API"))
+	if api_version == "v1beta1" {
+		err = marshaler.Marshal(w, toApiPipelineVersionV1(newPipelineVersion))
```
Linchin, line 264: In the previous code, we have a check for the output of `toApiPipelineVersionV1()` to be `nil' or not. I wonder if we should keep it.

backend/src/apiserver/server/pipeline_upload_server_test.go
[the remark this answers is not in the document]
chensun: `UploadPipelineVersion()` is highly similar to `UploadPipeline()`, but that isn't a good excuse at all. I agree with you we should have better test coverage here. I will follow up on improving the test coverage shortly.

backend/src/apiserver/server/pipeline_upload_server.go
chensun: `toApiPipelineVersionV1()` per its [current implementation](https://github.com/kubeflow/pipelines/blob/01c87f8a032e70a6ca92cdbefa974a7da387f204/backend/src/apiserver/server/api_converter.go#L432) never returns nil. So the previous code wasn't useful.

PR #9072, opened 2023-03-30
fix(backend): support reading large doc from URL; Fixes #7652

### LINKED ISSUE #7652: [backend] Unable to upload a specific testing pipeline via URL
For the user journey test of upload pipeline. The testing pipeline https://github.com/kubeflow/pipelines/blob/master/sdk/python/kfp/compiler_cli_tests/test_data/xgboost_sample_pipeline.yaml was not able to be uploaded via URL.

Linchin:
Although we support [larger file sizes now](https://github.com/kubeflow/pipelines/blob/4e85352ed484db84dca5414f32c4a0c0eebf7c85/backend/src/apiserver/common/const.go#L62), the size for reading from URL is still constrained due to [package bufio's limit](https://pkg.go.dev/bufio#pkg-constants). This PR updated the related function for loading files from URLs, as well as added related unit tests.

Fixes #7652

**Checklist:**

### DIFF
--- a/backend/src/apiserver/server/util.go
+++ b/backend/src/apiserver/server/util.go
@@ -29,17 +29,24 @@ import (
 )
 
 func loadFile(fileReader io.Reader, MaxFileLength int) ([]byte, error) {
-	reader := bufio.NewReader(fileReader)
-	pipelineFile := make([]byte, MaxFileLength+1)
-	size, err := reader.Read(pipelineFile)
-	if err != nil && err != io.EOF {
-		return nil, util.NewInvalidInputErrorWithDetails(err, "Error read pipeline file")
+	reader := bufio.NewReaderSize(fileReader, MaxFileLength)
+	var pipelineFile []byte
+	for {
+		currentRead := make([]byte, bufio.MaxScanTokenSize)
+		size, err := reader.Read(currentRead)
+		pipelineFile = append(pipelineFile, currentRead[:size]...)
+		if err == io.EOF {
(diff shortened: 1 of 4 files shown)


Linchin:
/test kubeflow-pipeline-backend-test
The backend test fail is caused by some flakiness in string comparison.

backend/src/apiserver/server/util.go
```diff
@@ -29,17 +29,24 @@ import (
...
 
 func loadFile(fileReader io.Reader, MaxFileLength int) ([]byte, error) {
-	reader := bufio.NewReader(fileReader)
-	pipelineFile := make([]byte, MaxFileLength+1)
-	size, err := reader.Read(pipelineFile)
-	if err != nil && err != io.EOF {
-		return nil, util.NewInvalidInputErrorWithDetails(err, "Error read pipeline file")
+	reader := bufio.NewReaderSize(fileReader, MaxFileLength)
+	var pipelineFile []byte
+	for {
```
gkcalat, line 34: Let's add a TODO comment to improve the performance by setting a custom buffer size instead of 64KB. With `MaxFileLength` of 32MB we may end up with 500 iterations in this loop instead of reading the file in a single step.

gkcalat (commented):
Thank you, @Linchin!

Aside from the small NIT below it looks good to me.

gkcalat (approved):
Thanks for adding the comment!

/lgtm
/approve

PR #9075, opened 2023-03-31
fix(backend): Fix CreatePipelineVersion API

chensun:
- Bring back `code_source_url` as an optional input field.
- Mark `pipeline_spec` as an output field.
- Remove an error comment in v1 API proto.
- Fix api to model converter to read only input fields.
- Fix `createPipelineVersion` to read pipeline spec only from pipeline URL.
- Refactor `toModelPipelineVersion` so it accepts only `PipelineVersion` api objects.

Manually tested the change works for both v1 and v2 `CreatePipelineVersion` API.

**Checklist:**


backend/api/v1beta1/pipeline.proto
gkcalat, line 1: Have you generated the client libraries? Shouldn't the removed comment result in a diff?

chensun: Good catch, I intentionally didn't regenerate v1beta1 as there were a bunch of unrelated changes (protobuf.Struct related). I plan to do the v1beta1 regeneration separately, potentially after the first bug bash even to avoid potential noises. 
I can postpone this comment removal as well, but feels it's highly relevant to the current PR, and likely a culprit for the previous misunderstanding.

backend/api/v2beta1/pipeline.proto
```diff
@@ -173,7 +173,10 @@ message PipelineVersion {
   // one package file location.
   Url package_url = 6;
 
-  // Required input field. Specifies the pipeline spec for the pipeline version.
+  // Optional input field. Pipeline version code source URL.
```
gkcalat, line 176: NIT: add a comment explaining differences between `package_url` and `code_source_url`.

gkcalat (commented):
Hi @chensun. Thank you for catching and fixing these inconsistencies in the BE logic.

backend/api/v1beta1/pipeline.proto
chensun: `Pipeline.parameters` is actually an **output**. Fixed the server logic on this. 
Also updated the comment for url and code_source_url. PTAL.

chensun:
/test kubeflow-pipeline-backend-test

Should have been fixed by #9079

gkcalat:
> /test kubeflow-pipeline-backend-test
> 
> Should have been fixed by #9079

The failure is irrelevant to this PR and it is caused by [these lines](https://github.com/kubeflow/pipelines/blob/master/backend/src/apiserver/server/api_converter.go#L1751). Let me fix it in a separate PR as this is my change. The sill be similar to #9079.

backend/src/apiserver/server/pipeline_server.go
```diff
@@ -588,9 +589,9 @@ func NewPipelineServer(resourceManager *resource.ResourceManager, options *Pipel
 // Creates a pipeline version from. Not exported.
 // Applies common logic on v1beta1 and v2beta1 API.
 func (s *PipelineServer) createPipelineVersion(ctx context.Context, pv *model.PipelineVersion) (*model.PipelineVersion, error) {
-	// Fail if pipeline spec is missing
-	if (pv.PipelineSpec == "") && (pv.CodeSourceUrl == "") && (pv.PipelineSpecURI == "") {
-		return nil, util.NewInvalidInputError("Failed to create a pipeline version due to missing pipeline spec")
+	// Fail if pipeline URL is missing
+	if pv.PipelineSpecURI == "" {
```
Linchin, line 593: What happens when the pipeline is not provided through URL? For example, when it is uploaded directly.

gkcalat: This is a great question.

Uploading is a special case where we parse the content of a request body into `PipelineSpec` field. We won't have a URL in that case and will directly call `ResourceManager`, which handles both cases.

backend/src/apiserver/server/api_converter.go
```diff
@@ -316,113 +309,38 @@ func toApiPipelines(pipelines []*model.Pipeline) []*apiv2beta1.Pipeline {
...
-		parameters = ""
 		pipelineId = apiPipelineVersionV2.GetPipelineId()
-		codeUrl = apiPipelineVersionV2.GetPackageUrl().GetPipelineUrl()
+		pipelineUrl = apiPipelineVersionV2.GetPackageUrl().GetPipelineUrl()
+		codeUrl = apiPipelineVersionV2.GetCodeSourceUrl()
 		description = apiPipelineVersionV2.GetDescription()
 	default:
 		return nil, util.NewUnknownApiVersionError("PipelineVersion", p)
 	}
 	return &model.PipelineVersion{
```
Linchin, line 337: Nit: Is there a specific reason we are not using `toModelPipelineVersion()` here?

chensun: This code is inside `toModelPipelineVersion`, it can't recursively calling itself without a stop condition.

backend/src/apiserver/server/pipeline_server.go
chensun: upload won't hit this code path, this `pipeline_server` only has `Create` methods, while `Upload` belongs to `pipeline_upload_server`: https://github.com/kubeflow/pipelines/blob/master/backend/src/apiserver/server/pipeline_upload_server.go

backend/src/apiserver/server/api_converter.go
Linchin: Sorry, I misread. XD I thought this was inside `createPipelineVersion()`.

backend/src/apiserver/server/pipeline_server.go
Linchin: I see! Thank you for clarifying, I mixed up these two APIs.

PR #9059, opened 2023-03-29
feat(backend): updates compiler to include kubernetes-specific specs in argo workflow

Linchin:
**Description of your changes:**
* Updated argo compiler to intake kubernetes-specific spec and incorporate it in the generated argo workflow
* Added compiler test

**Checklist:**
- [ ] The title for your pull request (PR) should follow our title convention. [Learn more about the pull request title convention used in this repository](https://github.com/kubeflow/pipelines/blob/master/CONTRIBUTING.md#pull-request-title-convention).

Issue #58, opened 2018-11-05
The Experiments/All runs tab should probably be the default tab in Experiments.


Ark-kun:
It's probably going to be the most frequently used tab. Especially in cases where the user only has one or two experiments.

PR #9079, opened 2023-03-31
test(backend): fix flaky backend test

Linchin:
Since #9059, the backend test has been rather flaky. After some investigation, it turned out that the tasks inside the root dag had a random order, and an added unit test in #9059 exposed this issue. This PR fixed the issue by ordering the tasks alphabetically.

**Checklist:**

### DIFF
--- a/backend/src/v2/compiler/argocompiler/dag.go
+++ b/backend/src/v2/compiler/argocompiler/dag.go
@@ -16,6 +16,7 @@ package argocompiler
 
 import (
 	"fmt"
+	"sort"
 	"strings"
 
 	wfapi "github.com/argoproj/argo-workflows/v3/pkg/apis/workflow/v1alpha1"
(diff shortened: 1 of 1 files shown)


backend/src/v2/compiler/argocompiler/dag.go
```diff
@@ -48,7 +49,18 @@ func (c *workflowCompiler) DAG(name string, componentSpec *pipelinespec.Componen
...
-	for taskName, kfpTask := range dagSpec.GetTasks() {
+	tasks := dagSpec.GetTasks()
+	// Iterate through tasks in deterministic order to facilitate testing.
+	// Note, order doesn't affect compiler with real effect right now.
+	// In the future, we may consider using topology sort when building local
+	// executor that runs on pipeline spec directly.
+	keys := make([]string, 0, len(tasks))
+	for key := range tasks {
+		keys = append(keys, key)
+	}
```
chensun, line 60: I think this chunk can be just

```golang
keys := dagSpec.GetTasks().Keys()
```

chensun: My bad, ignore me then.

chensun (approved):
/lgtm
/approve

Thanks for the quick fix.

backend/src/v2/compiler/argocompiler/dag.go
Linchin: With 
> ```go
> keys := reflect.ValueOf(dagSpec.GetTasks()).MapKeys()
> ```
It doesn't feel very readable? it also introduces a new package. The variable `tasks` is used again at line #58 so I don't think it makes a huge difference.

PR #9084, opened 2023-03-31
chore(backend): fix flakiness in toModelTasks function

gkcalat:

Caused flakiness in the backend test  in #9075 due to random order of the keys in the underlying `map[string]`

Similar to #9079.

**Checklist:**

### DIFF
--- a/backend/src/apiserver/server/api_converter.go
+++ b/backend/src/apiserver/server/api_converter.go
@@ -17,6 +17,7 @@ package server
 import (
 	"encoding/json"
 	"fmt"
+	"sort"
 	"strconv"
 
 	workflowapi "github.com/argoproj/argo-workflows/v3/pkg/apis/workflow/v1alpha1"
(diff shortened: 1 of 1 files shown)


backend/src/apiserver/server/api_converter.go
```diff
@@ -1747,8 +1748,15 @@ func toModelTasks(t interface{}) ([]*model.Task, error) {
 		runId := execSpec.ExecutionObjectMeta().Labels[util.LabelKeyWorkflowRunId]
 		namespace := execSpec.ExecutionNamespace()
 		createdAt := execSpec.GetCreationTimestamp().Unix()
+		// Get sorted node names to make the results repeatable
+		nodeNames := make([]string, 0, len(execSpec.Status.Nodes))
```
Linchin, line 1752: Just wondering, for cases when `execSpec.Status == nil` or `execSpec.Status.Nodes == nil`, are they possible to happen? If so, I think it's a good idea to cover these edge cases.

gkcalat: I believe we don't need this as the `Status` is node pointer. `nodeNames` slice will get appropriate capacity and length. Here is an example: https://go.dev/play/p/5xcfq5TT7b4

Issue #9087, opened 2023-04-02
[backend] Resource limits and requests not working in KFP v2


gkcalat:
### Environment

*  How did you deploy Kubeflow Pipelines (KFP)?
Standalone KFP from master

*  KFP SDK version: 
2.0.0-beta.13

### Steps to reproduce

1. Create a pipeline version with `.set_memory_limit(A).set_memory_request(B).set_cpu_limit(C).set_cpu_request(D)`
2. Create a run from the pipeline
3. Check the corresponding pod's `spec.requests`

### Expected result

```yaml
    resources:
      limits:
        cpu: C
        memory: A
      requests:
        cpu: D
        memory: B
```

### Materials and Reference

```yaml
    resources:
      limits:
        cpu: 500m
        memory: 512Mi
      requests:
        cpu: 10m
        memory: 32Mi
```

The values are hardcoded and do not depend on what we set as `A`, `B`, `C`, and `D`.

They seem to come from `workflow-controller-configmap`. The change was probably introduced in #5695.

---

Impacted by this bug? Give it a 👍.

PR #9088, opened 2023-04-02
feat(backend): add support for resource requests. Closes #7047. Closes #6354.

### LINKED ISSUE #6354: v2 - support resource requests and limits
* [x] https://github.com/kubeflow/pipelines/pull/7045
* [x] #7043
* [ ] #7047

### LINKED ISSUE #7047: feat(backend): support resource requests
KFP pipeline spec only has resource limit fields right now, consider also adding support for request fields.
https://github.com/kubeflow/pipelines/blob/1e032f550ce23cd40bfb6827b995248537b07d08/api/v2alpha1/pipeline_spec.proto#L632-L653

gkcalat:

- Added setting of resource requests (CPU and memory) to KFP v2 driver
- Added a unit test
- [x] Tested manually on Standalone KFP @ GKE 1.24
    - not functional due to #9087
    - Worked as expected after removing resources in `workflow-controller-configmap`
- Hold until #9089

**Checklist:**


Linchin:
Overall the PR looks good. Just some thoughts: 
* Can we add a sample test for this feature so we can test it end to end?
* Maybe build the driver image and upload it, and update the SHA identification so we don't have to create a separate PR for that.

gkcalat:
> Can we add a sample test for this feature so we can test it end to end?
> Maybe build the driver image and upload it, and update the SHA identification so we don't have to create a separate PR for that.

Thank you @Linchin!

I might be wrong on this, but it might be difficult to verify provisioned resources from within a pod itself. Please, let me know if there is a way to test resource requests. The manual testing is probably suffice in this case.

I also updated the driven image. It looks like we recently introduced hardcoded driver image in a bunch of test cases. This should probably be fixed in a separate PR as it makes the backend testing to detect irrelevant changes.

/cc @chensun

Linchin:
@gkcalat Thank you! I don't think these are big issues, so I will approve the PR. Just for the purpose of brainstorming, I checked [how v1 samples tested resource requests](https://github.com/kubeflow/pipelines/blob/5de4d7865442600e1c071d061df8c8f951411179/samples/core/resource_spec/resource_spec_test.py). Looks like it looked for a failure caused by OOM. I'm curious if there is a more straightforward way to verify this.
/lgtm
/approve

gkcalat:
> Interesting, is v1 sample test still run in presubmit? I thought the config map removed by https://github.com/kubeflow/pipelines/pull/9089 would break v1 as well. Was that not the case? @gkcalat

It sees that the sample pipelines have limits/requests already. So before #9089, these values were replaced by smaller default values. Now, they are used as is. So, I guess the CPU and memory in the sample tests are small enough to cause OOM.

chensun (approved):
/lgtm
/approve

Thanks!

PR #9101, opened 2023-04-04
chore(frontend): Change pipeline_version_id field in run / recurring run

jlyaoyuli:
In the latest backend changes, `pipeline_id` is also added in the return value of run and recurring run. In addition, `pipeline_id` and `pipeline_version_id` are wrapped as `pipeline_version_reference`. The main purpose of this PR is modifying frontend implementation to match the changes.

### DIFF
--- a/frontend/src/pages/NewRunSwitcher.tsx
+++ b/frontend/src/pages/NewRunSwitcher.tsx
@@ -92,8 +92,8 @@ function NewRunSwitcher(props: PageProps) {
   const pipelineVersionId =
     pipelineVersionIdParam ||
     apiPipeline?.default_version?.id ||
-    v2Run?.pipeline_version_id ||
-    recurringRun?.pipeline_version_id;
+    v2Run?.pipeline_version_reference?.pipeline_version_id ||
+    recurringRun?.pipeline_version_reference?.pipeline_version_id;
 
   const { isFetching: pipelineVersionIsFetching, data: apiPipelineVersion } = useQuery<
     ApiPipelineVersion,
--- a/frontend/src/pages/NewRunV2.test.tsx
+++ b/frontend/src/pages/NewRunV2.test.tsx
@@ -103,7 +103,10 @@ describe('NewRunV2', () => {
     finished_at: new Date('2021-05-18T21:01:23.000Z'),
     run_id: TEST_RUN_ID,
     display_name: 'Run of v2-xgboost-ilbo',
-    pipeline_version_id: ORIGINAL_TEST_PIPELINE_VERSION_ID,
+    pipeline_version_reference: {
+      pipeline_id: ORIGINAL_TEST_PIPELINE_ID,
+      pipeline_version_id: ORIGINAL_TEST_PIPELINE_VERSION_ID,
+    },
     runtime_config: { parameters: { intParam: 123 } },
     scheduled_at: new Date('2021-05-17T20:58:23.000Z'),
     state: V2beta1RuntimeState.SUCCEEDED,
@@ -116,7 +119,10 @@ describe('NewRunV2', () => {
     finished_at: new Date('2022-08-12T21:01:23.000Z'),
     run_id: 'test-clone-ui-run-id',
     display_name: 'Clone of Run of v2-xgboost-ilbo',
-    pipeline_version_id: NEW_TEST_PIPELINE_VERSION_ID,
+    pipeline_version_reference: {
+      pipeline_id: NEW_TEST_PIPELINE_ID,
+      pipeline_version_id: ORIGINAL_TEST_PIPELINE_VERSION_ID,
+    },
     runtime_config: { parameters: { intParam: 123 } },
     scheduled_at: new Date('2022-08-12T20:58:23.000Z'),
     state: V2beta1RuntimeState.SUCCEEDED,
@@ -152,7 +158,10 @@ describe('NewRunV2', () => {
     created_at: new Date('2021-05-17T20:58:23.000Z'),
     description: 'V2 xgboost',
     display_name: 'Run of v2-xgboost-ilbo',
-    pipeline_version_id: ORIGINAL_TEST_PIPELINE_VERSION_ID,
+    pipeline_version_reference: {
+      pipeline_id: ORIGINAL_TEST_PIPELINE_ID,
+      pipeline_version_id: ORIGINAL_TEST_PIPELINE_VERSION_ID,
+    },
     recurring_run_id: TEST_RECURRING_RUN_ID,
     runtime_config: { parameters: { intParam: 123 } },
     trigger: {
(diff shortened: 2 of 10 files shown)


frontend/src/pages/NewRunSwitcher.tsx
```diff
@@ -92,8 +92,8 @@ function NewRunSwitcher(props: PageProps) {
   const pipelineVersionId =
     pipelineVersionIdParam ||
     apiPipeline?.default_version?.id ||
```
chensun, line 94: Would this be changed with the pipeline api integration? I wonder if this change can or should be part of the pipeline api integration change to avoid touching this code twice.

frontend/src/pages/PipelineDetails.tsx
```diff
@@ -226,7 +226,7 @@ class PipelineDetails extends Page<{}, PipelineDetailsState> {
     // existing run or recurring run have two kinds of resources that can provide template string
 
     // 1. Pipeline version id
-    const pipelineVersionId = existingObj.pipeline_version_id;
+    const pipelineVersionId = existingObj.pipeline_version_reference?.pipeline_version_id;
```
chensun, line 229: Do we also need `pipeline_id` for this page?

frontend/src/pages/RunList.test.tsx
```diff
@@ -453,7 +464,9 @@ describe('RunList', () => {
 
   it('shows pipeline version name', async () => {
     mockNRuns(1, {
-      pipeline_version_id: 'testversion1',
+      pipeline_version_reference: {
+        pipeline_version_id: 'testversion1',
```
chensun, line 468: Whenever `pipeline_version_id` present, `pipeline_id` should present as well to make `pipeline_version_reference` a valid object, even if your test might not need to check the other part.

frontend/src/pages/NewRunSwitcher.tsx
jlyaoyuli: If we put this changes in the pipeline API integration, the functionality will break at this moment.

frontend/src/pages/RunList.test.tsx
jlyaoyuli: Agree with your point; however, (Correct me if I am wrong), I think making both changes (adding `pipeline_id` in the implementation and test) in pipeline API integration rather separating into two PRs is better to manage the code changes.

frontend/src/pages/RunDetailsRouter.tsx
```diff
@@ -40,7 +40,7 @@ export default function RunDetailsRouter(props: RunDetailsProps) {
     pipelineManifest = JsYaml.safeDump(v2Run.pipeline_spec);
   }
 
-  const pipelineVersionId = v2Run?.pipeline_version_id;
+  const pipelineVersionId = v2Run?.pipeline_version_reference?.pipeline_version_id;
```
zijianjoy, line 43: Do we need to change https://github.com/kubeflow/pipelines/blob/d53307551ffe133895f9475419dd436755cffa31/frontend/src/pages/RunDetailsRouter.tsx#L52 to provide pipelineId as well?

frontend/src/pages/NewRunV2.test.tsx
```diff
@@ -116,7 +119,10 @@ describe('NewRunV2', () => {
     finished_at: new Date('2022-08-12T21:01:23.000Z'),
     run_id: 'test-clone-ui-run-id',
     display_name: 'Clone of Run of v2-xgboost-ilbo',
-    pipeline_version_id: NEW_TEST_PIPELINE_VERSION_ID,
+    pipeline_version_reference: {
+      pipeline_id: NEW_TEST_PIPELINE_ID,
+      pipeline_version_id: ORIGINAL_TEST_PIPELINE_VERSION_ID,
```
zijianjoy, line 124: Why are we using *Original* ORIGINAL_TEST_PIPELINE_VERSION_ID while the pipeline_id field is *New* `NEW_TEST_PIPELINE_ID`?

frontend/src/pages/NewRunV2.tsx
```diff
@@ -288,11 +301,13 @@ function NewRunV2(props: NewRunV2Props) {
       description: runDescription,
       display_name: runName,
       experiment_id: apiExperiment?.id,
-      // pipeline_spec and pipeline_version_id is exclusive.
-      pipeline_spec: !(existingRun?.pipeline_version_id || existingPipelineVersion?.id)
+      // pipeline_spec and pipeline_version_reference is exclusive.
+      pipeline_spec: !(pipelineVersionRefClone || pipelineVersionRefNew)
```
zijianjoy, line 305: Correct me if I am wrong: `pipelineVersionRefClone` is an object while `pipeline_version_id` is a string. And `{}` is truly instead of falsy: https://stackoverflow.com/questions/21743921/javascript-is-an-empty-object-a-falsy-one.

frontend/src/pages/RunDetailsRouter.tsx
jlyaoyuli: Yes, this changes is needed and it will be included in the pipeline API integration.

frontend/src/pages/NewRunV2.test.tsx
jlyaoyuli: Good catch!

frontend/src/pages/NewRunV2.tsx
jlyaoyuli: Thanks for the reminder! If we get a run / recurring created by `pipeline_spec`, the `pipeline_version_reference` in the API return value is undefined rather than `{}`.

zijianjoy: Sounds good, thank you for confirming Joe!

frontend/src/pages/NewRunV2.tsx
```diff
@@ -288,11 +301,13 @@ function NewRunV2(props: NewRunV2Props) {
...
         ? JsYaml.safeLoad(templateString || '')
         : undefined,
-      pipeline_version_id: existingRun?.pipeline_version_id || existingPipelineVersion?.id,
+      pipeline_version_reference: cloneOrigin.isClone
```
zijianjoy, line 308: Since pipeline_spec and pipeline_version_reference need to be exclusive, I think the condition check here for pipeline_version_reference should be: 

```
pipeline_version_reference: (pipelineVersionRefClone || pipelineVersionRefNew) ? 
              ( cloneOrigin.isClone ? pipelineVersionRefClone  : pipelineVersionRefNew ) : 
               undefined,
```

Is that right?

frontend/src/pages/NewRunSwitcher.tsx
chensun: Looking back, backend should probably do it in a non-breaking way. Anyway, the fact that this functionality was broken has happened at the time the backend PR merged. Since, apparently we have no test coverage on the feature, which we should improve post-GA, and we are not doing any release until pipeline API integration is done. It's okay to fix it later. 
I don't have a strong opinion on whether to combine or separate the changes, you can make the call.

frontend/src/pages/RunList.test.tsx
chensun: This seems to be a unit test that you're providing a test input, if you just add one line of `pipeline_id` here, would it require any other change? I might be wrong, but I guess it won't break anything, and you have a valid test input that you don't need to change again soon after.

PR #9106, opened 2023-04-05
feat(backend): support pvc creation, deletion and mount; fixes #8708

### LINKED ISSUE #8708: feat(backend): Support volume mount

Linchin:
This PR adds support for creating PVC, deleting PVC and mounting them in tasks.
* Updated the argo compiler, so the tasks of creating/deleting PVC only has a driver, but not a launcher/executor
* Updated the driver to handle Kubernetes configs input, and carry out the corresponding creation/deletion/mounting of PVCs
* Build the new driver image, uploaded, and updated the image address

Note:
Because the package `kfp/kubernetes` is not published yet, there's no samples test added for now.

**Checklist:**


backend/src/v2/driver/driver.go
```diff
@@ -9,17 +22,21 @@ import (
...
 	"github.com/kubeflow/pipelines/backend/src/v2/cacheutils"
 	"github.com/kubeflow/pipelines/backend/src/v2/component"
 	"github.com/kubeflow/pipelines/backend/src/v2/config"
 	"github.com/kubeflow/pipelines/backend/src/v2/expression"
 	"github.com/kubeflow/pipelines/backend/src/v2/metadata"
+	"github.com/kubeflow/pipelines/kubernetes_platform/go/kubernetesplatform"
 	pb "github.com/kubeflow/pipelines/third_party/ml-metadata/go/ml_metadata"
 	"google.golang.org/protobuf/encoding/protojson"
 	"google.golang.org/protobuf/types/known/structpb"
-	k8score "k8s.io/api/core/v1"
```
gkcalat, line 21: NIT: avoid renaming of package aliases as it creates irrelevant code diff and can confuse developers

backend/src/v2/compiler/argocompiler/dag.go
```diff
@@ -1,17 +1,16 @@
-// Copyright 2021 The Kubeflow Authors
+// Copyright 2021-2023 The Kubeflow Authors
 //
 // Licensed under the Apache License, Version 2.0 (the "License");
 // you may not use this file except in compliance with the License.
 // You may obtain a copy of the License at
 //
-//      https://www.apache.org/licenses/LICENSE-2.0
+//	https://www.apache.org/licenses/LICENSE-2.0
```
gkcalat, line 7: NIT: revert?

backend/src/v2/component/launcher_v2.go
```diff
@@ -1,3 +1,16 @@
+// Copyright 2021-2023 The Kubeflow Authors
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//	http://www.apache.org/licenses/LICENSE-2.0
```
gkcalat, line 7: NIT: should we replace the `\tab` with spaces to be consistent with other license headers?

backend/src/v2/driver/driver.go
```diff
@@ -1,3 +1,16 @@
+// Copyright 2021-2023 The Kubeflow Authors
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//	https://www.apache.org/licenses/LICENSE-2.0
```
gkcalat, line 7: NIT: same. To be clear, these are super-tiny NITs

backend/src/v2/compiler/argocompiler/argo.go
```diff
@@ -326,3 +326,8 @@ var launcherResources = k8score.ResourceRequirements{
 const (
 	tmplEntrypoint = "entrypoint"
 )
+
+var dummyImages = map[string]bool{
+	"argostub/createpvc": true,
+	"argostub/deletepvc": true,
```
gkcalat, line 332: I wonder if we can use more arbitrary container names. While, it's a very wild scenario, it's still possible for someone to create a docker hub repo called `argostub` and upload containers `createpvc` and `deletepvc`. WDYT?

Besides, should we add comments and document it somewhere?

backend/src/v2/compiler/argocompiler/dag.go
Linchin: Thank you, I will revert this. I'm not sure why these white spaces were replaced though.

Linchin: Looks like it is caused by VSCode's formatting tools:
https://github.com/microsoft/vscode-go/issues/1930

backend/src/v2/compiler/argocompiler/argo.go
Linchin: I agree that these cases may happen, although not with a very high likelihood. These dummy image names seem to be the sole identification that this component is not a normal user defined component, but a special component for a specific function. I think the user will have to avoid such container names and we definitely should make this clear in the docs. I will add comments here and a TODO that we need to include these restrictions in the documentation.

backend/src/v2/driver/driver.go
```diff
@@ -285,7 +333,28 @@ func Container(ctx context.Context, opts Options, mlmd *metadata.Client, cacheCl
...
+		glog.Infof("Published this execution (createpvc).")
+	}
+	if opts.Container.Image == "argostub/deletepvc" {
+		if err := mlmd.PublishExecution(ctx, createdExecution, nil, nil, pb.Execution_COMPLETE); err != nil {
+			return execution, fmt.Errorf("failed to publish execution: %w", err)
+		}
+		glog.Infof("Published this execution (deletepvc).")
+	}
+
+	execution.PodSpecPatch, err = makePodSpecPatch(opts.Container, opts.Component, executorInput, execution.ID, opts.PipelineName, opts.RunID, opts.KubernetesConfig, opts.Task, dag, dagTasks)
```
gkcalat, line 357: I don't have enough context here, but do you think we can skip the `makePodSpecPatch` for both `createpvc` and `deletepvc`?

Linchin: Good catch! I think we can skip it, since this patch is intended for the launcher container, which will not be created in this case. I also verified it manually.

backend/src/v2/component/launcher_v2.go
Linchin: I will replace the tabs with 6 spaces, as is consistent with other copyright declarations.

backend/src/v2/driver/driver.go
Linchin: I will replace the tabs with 6 spaces, as is consistent with other copyright declarations.

backend/src/v2/driver/driver.go
```diff
@@ -209,6 +232,10 @@ func Container(ctx context.Context, opts Options, mlmd *metadata.Client, cacheCl
 		return nil, err
 	}
 	glog.Infof("parent DAG: %+v", dag.Execution)
+	dagTasks, err := mlmd.GetExecutionsInDAG(ctx, dag, pipeline)
+	if err != nil {
+		return nil, err
+	}
```
gkcalat, line 238: It looks like we only need this call to MLMD when we have a `pvc_mount` with `pvc_reference` set via `task_output_parameter`. I feel this may be a pretty rare scenario, which means we should probably put it inside an if-statement.

backend/src/v2/driver/driver.go
gkcalat, line 357: I feel the `makePodSpecPatch` function is getting overloaded with additional arguments. Taking this concern along with [this comment](https://github.com/kubeflow/pipelines/pull/9106/files#r1159150523), can we set the generation of the pod-spec patch with platform-specific features into a separate function:
1. Call `makePodSpecPatch` that returns `*corev1.PodSpec`
2. Call `extendPodSpecPatch` that takes the output of `makePodSpecPatch` and appends platform-specific changes and marshals the patch into a string

backend/src/v2/driver/driver.go
```diff
@@ -940,3 +1021,203 @@ func generateOutputURI(root, artifactName string, taskName string) string {
...
 }
+
+var accessModeMap = map[string]corev1.PersistentVolumeAccessMode{
+	"ReadWriteOnce":    corev1.ReadWriteOnce,
+	"ReadOnlyMany":     corev1.ReadOnlyMany,
+	"ReadWriteMany":    corev1.ReadWriteMany,
+	"ReadWriteOncePod": corev1.ReadWriteOncePod,
+}
+
+func CreatePVC(inputs *pipelinespec.ExecutorInput_Inputs, namespace string) (pvcName string, err error) {
```
gkcalat, line 1032: NIT: add comment describing the function

backend/src/v2/driver/driver.go
```diff
@@ -940,3 +1021,203 @@ func generateOutputURI(root, artifactName string, taskName string) string {
...
+	// Create the PVC in the cluster
+	createdPVC, err := k8sClient.CoreV1().PersistentVolumeClaims(namespace).Create(context.Background(), pvc, metav1.CreateOptions{})
+	if err != nil {
+		return "", fmt.Errorf("failed to create pvc: %w", err)
+	}
+	glog.Infof("Created PVC %s\n", createdPVC.ObjectMeta.Name)
+	return createdPVC.ObjectMeta.Name, nil
+}
+
+func DeletePVC(inputs *pipelinespec.ExecutorInput_Inputs, namespace string) error {
```
gkcalat, line 1126: NIT: add a comment describing the function

backend/src/v2/driver/driver.go
```diff
@@ -940,3 +1021,203 @@ func generateOutputURI(root, artifactName string, taskName string) string {
...
+	err = k8sClient.CoreV1().PersistentVolumeClaims(namespace).Delete(context.TODO(), pvcName, metav1.DeleteOptions{})
+	if err != nil {
+		return fmt.Errorf("failed to delete pvc %s: %v", pvcName, err)
+	}
+
+	glog.Infof("Deleted PVC %s\n", pvcName)
+	return nil
+}
+
+func makeVolumeMountPatch(pvcMount []*kubernetesplatform.PvcMount, task *pipelinespec.PipelineTaskSpec, dag *metadata.DAG, dagTasks map[string]*metadata.Execution) ([]corev1.VolumeMount, []corev1.Volume, error) {
```
gkcalat, line 1160: NIT: add a comment describing the function

backend/src/v2/driver/driver.go
gkcalat, line 1: Could you please add unit-tests for the new functions to verify that they function correctly as well as the old functions with specified volumes (e.g. verifying that volumes and volumeMounts are correctly generated in the pod-spec patch)?

backend/src/v2/driver/driver_test.go
gkcalat, line 1: Could you please add unit-tests for the new functions to verify that they function correctly as well as the old functions with specified volumes (e.g. verifying that `volumes` and `volumeMounts` are correctly generated in the pod-spec patch)?

gkcalat (commented):
Thank you, @Linchin!

I left a few suggestions with performance/readability optimizations. Great job!

backend/src/v2/driver/driver.go
Linchin: Thank you for bringing this up. I changed this because the more conventional package name is `corev1` or `v1`. But for the purpose of this PR, I will change it back.

backend/src/v2/driver/driver.go
```diff
@@ -217,6 +244,27 @@ func Container(ctx context.Context, opts Options, mlmd *metadata.Client, cacheCl
 	if err != nil {
 		return nil, err
 	}
+
+	// When the container image is a dummy image, there is no launcher for this task.
+	// This happens when this task is created to implements a Kubernetes-specific configuration, i.e., there is nothing to run.
+
+	// This task creates a PVC.
+	var pvcName string
+	if opts.Container.Image == "argostub/createpvc" {
```
chensun, line 253: There are a number of places we use the string value `"argostub/createpvc"` directly. We should create a constant for it.

backend/src/v2/driver/driver_test.go
chensun: +1, the amount of changes appeared in `driver.go` definitely requires more test than we saw here.

backend/src/v2/driver/driver.go
Linchin: Good point! I moved this part to inside an if-statement for creating PVC.

backend/src/v2/driver/driver.go
```diff
@@ -217,6 +244,27 @@ func Container(ctx context.Context, opts Options, mlmd *metadata.Client, cacheCl
 	if err != nil {
 		return nil, err
 	}
+
+	// When the container image is a dummy image, there is no launcher for this task.
+	// This happens when this task is created to implements a Kubernetes-specific configuration, i.e., there is nothing to run.
+
+	// This task creates a PVC.
```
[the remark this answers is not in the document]
Linchin: Good point. I will remove this.

backend/src/v2/driver/driver.go
Linchin: I will create a constant in the `driver` package for now, but I think in the long run it would be nice to declare them in the proto, and import from there.

backend/src/v2/driver/driver_test.go
Linchin: Thank you, I added some unit tests.

backend/src/v2/driver/driver.go
Linchin: Good point. I added some comments for this.

backend/src/v2/driver/driver.go
Linchin: For this function, I think its name is pretty self-explanatory.  WDYT?

backend/src/v2/driver/driver.go
Linchin: Same as above, adding comments describing this function feels quite redundant.

chensun (commented):
One comment, otherwise LGTM.

backend/src/v2/compiler/argocompiler/argo.go
```diff
@@ -326,3 +326,12 @@ var launcherResources = k8score.ResourceRequirements{
 const (
 	tmplEntrypoint = "entrypoint"
 )
+
+// Here is the collection of all special dummy images that the backend recognizes.
+// User need to avoid these image names for their self-defined components.
+// These values are in sync with the values in SDK to form a contract between BE and SDK.
+// TODO(lingqinggan): clarify these in documentation for KFP V2.
```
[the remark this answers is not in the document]
Linchin: When I wrote the comments, I was thinking about the scenario @gkcalat brought up, where the user uses an image name that's the same as our dummy image names. It feel necessary for the users to be informed of their image names choices.

I also thought it would be a good idea to have a single source of all the constants we are using here. However, it feels that neither package argo compiler nor package driver is a good place to be the place to hold these constants. By defining the costants in either places, it introduces an unnecessary dependency between the driver and the compiler. I think it's a better idea to define them in the `kubernetesplatform` proto, and all packages, both backend and sdk, all import from the single source.

backend/src/v2/driver/driver.go
```diff
@@ -400,11 +492,25 @@ func makePodSpecPatch(
...
+	dagTasks map[string]*metadata.Execution,
+) error {
+	// Get volume mount information
+	volumeMounts, volumes, err := makeVolumeMountPatch(kubernetesConfig.GetPvcMount(), dag, dagTasks)
 	if err != nil {
-		return "", fmt.Errorf("JSON marshaling pod spec patch: %w", err)
+		return fmt.Errorf("failed to extract volume mount info: %w", err)
 	}
-	return string(podSpecPatchBytes), nil
+	// TODO(gkcalat): add nodeSelector once it is added to platform-specific features in PipelineSpec
```
gkcalat, line 510: Thank you moving my comment here!

backend/src/v2/driver/driver.go
```diff
@@ -9,21 +22,30 @@ import (
...
+	"k8s.io/apimachinery/pkg/api/resource"
 	k8sres "k8s.io/apimachinery/pkg/api/resource"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/client-go/kubernetes"
 	"k8s.io/client-go/rest"
 )
 
+const (
+	createPVCImage = "argostub/createpvc"
+	deletePVCImage = "argostub/deletepvc"
```
gkcalat, line 46: NIT: should we use the constant `dummyImages` defined in `argocompiler` package instead of defining them here too?

gkcalat (commented):
Hi @Linchin!

Thank you for addressing the comments!

/lgtm

backend/src/v2/compiler/argocompiler/argo.go
```diff
@@ -118,7 +118,7 @@ func Compile(jobArg *pipelinespec.PipelineJob, kubernetesSpecArg *pipelinespec.S
 		wf:        wf,
 		templates: make(map[string]*wfapi.Template),
 		// TODO(chensun): release process and update the images.
-		driverImage:   "gcr.io/ml-pipeline-test/kfp-driver@sha256:690e078d675f9cc91b25465aaabd6bf8d7fa2b745741ad36ac2fce0ce217c11c",
+		driverImage:   "gcr.io/ml-pipeline-test/kfp-driver@sha256:0a26d38f18d538e2df30d1f1ca44ac77deff7bd1d4bf35246bb12bae1bf72967",
```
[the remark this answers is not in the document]
gkcalat: This PR has been merged. You may need to rebase and update the driver image once again.

backend/src/v2/driver/driver.go
Linchin: There is a similar discussion here:
https://github.com/kubeflow/pipelines/pull/9106#discussion_r1161012631

backend/src/v2/cmd/driver/main.go
```diff
@@ -140,13 +149,14 @@ func drive() (err error) {
...
-		DAGExecutionID: *dagExecutionID,
-		IterationIndex: *iterationIndex,
+		PipelineName:     *pipelineName,
+		RunID:            *runID,
+		Namespace:        namespace,
+		Component:        componentSpec,
+		Task:             taskSpec,
+		DAGExecutionID:   *dagExecutionID,
+		IterationIndex:   *iterationIndex,
+		KubernetesConfig: kubernetesConfig,
```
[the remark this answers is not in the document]
Linchin: Good point! I will move this.

chensun (approved):
/lgtm
/approve

Thanks!

Linchin:
Thank you @gkcalat ! Overall this PR looks good to me.  Just left a few small comments.

backend/src/v2/driver/driver.go
```diff
@@ -400,11 +492,25 @@ func makePodSpecPatch(
...
+	// Get volume mount information
+	volumeMounts, volumes, err := makeVolumeMountPatch(kubernetesConfig.GetPvcMount(), dag, dagTasks)
 	if err != nil {
-		return "", fmt.Errorf("JSON marshaling pod spec patch: %w", err)
+		return fmt.Errorf("failed to extract volume mount info: %w", err)
 	}
-	return string(podSpecPatchBytes), nil
+	// TODO(gkcalat): add nodeSelector once it is added to platform-specific features in PipelineSpec
+	podSpec.Volumes = volumes
+	podSpec.Containers[0].VolumeMounts = volumeMounts
```
gkcalat, line 512: I wanted to confirm with you that we intentionally hard-coding the volumes to be mounted only to the first container. It might be okay to assume that we always have a single user container and that it goes first for every pod.

chensun: Yes, I think we can assume one user container for now.

PR #9112, opened 2023-04-06
feat(SDK)!: SDK client v2beta1 API integration. Fixes #8706

### LINKED ISSUE #8706: feat(SDK): KFP v2 API integration

chensun:
Migrate SDK client to use v2beta1 API.

Manually tested all public method with most commonly used arguments.

**Checklist:**

### DIFF
--- a/sdk/python/kfp/cli/recurring_run.py
+++ b/sdk/python/kfp/cli/recurring_run.py
@@ -135,7 +135,7 @@ def create(ctx: click.Context,
         raise ValueError(either_option_required)
     if not experiment_id:
         experiment = client_obj.create_experiment(experiment_name)
-        experiment_id = experiment.id
+        experiment_id = experiment.experiment_id
 
     # Ensure we only split on the first equals char so the value can contain
     # equals signs too.
--- a/sdk/python/kfp/cli/recurring_run.py
+++ b/sdk/python/kfp/cli/recurring_run.py
@@ -159,7 +159,7 @@ def create(ctx: click.Context,
         version_id=version_id)
     output.print_output(
         recurring_run,
-        output.ModelType.JOB,
+        output.ModelType.RECURRING_RUN,
         output_format,
     )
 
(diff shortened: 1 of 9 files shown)


sdk/python/kfp/client/client.py
```diff
@@ -55,22 +57,20 @@
...
-    def __init__(
-            self, spec: kfp_server_api.ApiPipelineSpec,
-            resource_references: kfp_server_api.ApiResourceReference) -> None:
-        self.spec = spec
-        self.resource_references = resource_references
+@dataclasses.dataclass
+class _JobConfig:
+    pipeline_spec: dict
+    pipeline_version_reference: kfp_server_api.V2beta1PipelineVersionReference
+    runtime_config: kfp_server_api.V2beta1RuntimeConfig
```
connor-mccarthy, line 64: Good thinking on this change.

chensun:
Fixes https://github.com/kubeflow/pipelines/issues/8706.

Test failures are expected because we haven't released `kfp-server-api==2.0.0b1` yet, which we plan to do in the next couple of days.

sdk/python/kfp/client/client.py
```diff
@@ -176,7 +176,8 @@ def __init__(
             header_value=self._context_setting.get(
                 'client_authentication_header_value'))
         _add_generated_apis(self, kfp_server_api, api_client)
-        self._job_api = kfp_server_api.JobServiceApi(api_client)
+        self._recurring_run_api = kfp_server_api.RecurringRunServiceApi(
+            api_client)
```
connor-mccarthy, line 180: Per offline conversation, consider aliasing existing uses of "job" in the client's public API to use "recurring_run" instead if "recurring run" is the new canonical name for this object.

sdk/python/kfp/client/client.py
```diff
@@ -495,16 +491,16 @@ def get_pipeline_id(self, name: str) -> Optional[str]:
         """
         pipeline_filter = json.dumps({
             'predicates': [{
-                'op': _FILTER_OPERATIONS['EQUALS'],
-                'key': 'name',
+                'operation': _FILTER_OPERATIONS['EQUALS'],
```
connor-mccarthy, line 494: Would the key change from `'op'` to `'operation'` be a breaking change?

sdk/python/kfp/client/client.py
```diff
@@ -517,79 +513,78 @@ def list_experiments(
...
-                (see `filter.proto message <https://github.com/kubeflow/pipelines/blob/777c98153daf3dfae82730e14ff37bdddc334c4d/backend/api/filter.proto>`_). For a list of all filter operations ``'op'``, see `here <https://github.com/kubeflow/pipelines/blob/777c98153daf3dfae82730e14ff37bdddc334c4d/sdk/python/kfp/client/client.py#L37-L45>`_. Example:
+                (see `filter.proto message <https://github.com/kubeflow/pipelines/blob/cb7d9a87c999eb1d2280959e5afbeee9e270ef3d/backend/api/v2beta1/filter.proto>`_). Example:
 
                   ::
 
                     json.dumps({
                         "predicates": [{
-                            "op": _FILTER_OPERATIONS["EQUALS"],
-                            "key": "name",
+                            "operation": "EQUALS",
```
connor-mccarthy, line 532: Curious: why remove `_FILTER_OPERATIONS` here?

connor-mccarthy (commented):
/lgtm

sdk/python/kfp/client/client.py
```diff
@@ -55,22 +57,20 @@
...
+    pipeline_spec: dict
+    pipeline_version_reference: kfp_server_api.V2beta1PipelineVersionReference
+    runtime_config: kfp_server_api.V2beta1RuntimeConfig
 
 
 class RunPipelineResult:
 
     def __init__(self, client: 'Client',
-                 run_info: kfp_server_api.ApiRun) -> None:
+                 run_info: kfp_server_api.V2beta1Run) -> None:
```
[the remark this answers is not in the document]
chensun: No, not until the next major refresh `v3`. The prefix comes from the proto package name (last section), e.g.: https://github.com/kubeflow/pipelines/blob/56f1c80c7eb51aaf0dabac4514e6c036f7f17ba5/backend/api/v2beta1/run.proto#L18

Minor API changes, hopefully should be backward compatible won't introduce changes to the package name.

sdk/python/kfp/client/client.py
chensun: Unfortunately yes, this is a breaking change. I guess when we made the API changes, we missed that SDK users could pass in this as a string. 
Technically we could avoid breaking SDK users by loading the JSON back to a dict and do a string replacement on behalf of the user, but I feel a bit hesitate to introduce such a hack that alters user provided value. Hopefully this breakage would be easy to discover and easy to fix. WDYT?

sdk/python/kfp/client/client.py
chensun: This is in the docstring part, and `_FILTER_OPERATIONS` isn't exposed, the example is not really runnable--it needs to be at least `kfp.client._FILTER_PERATIONS[...]` to be runnable. 
Also the API works with either the enum name or the enum value, if users has to know the key `EQUALS`, there's really no point for them to convert the key to its enum value here.

sdk/python/kfp/client/client.py
connor-mccarthy: Agreed that this is a very minor breaking change. I'm okay with not fixing. It will be easy to add that hack if it turns out to be a pain point for users. Plus, we don't promise backward compatibility from the main namespace to KFP SDK v2 anyway and the client is associate most with the main namespace in KFP v1.

sdk/python/kfp/cli/recurring_run.py
```diff
@@ -202,73 +202,75 @@ def list(ctx: click.Context, experiment_id: str, page_token: str, max_size: int,
...
-        output.ModelType.JOB,
+        response.recurring_runs or [],
+        output.ModelType.RECURRING_RUN,
         output_format,
     )
 
 
 @recurring_run.command()
-@click.argument('job-id')
+@click.argument('recurring-run-id')
```
chensun, line 212: Chose to break the CLI interface because it's painful to support both old and new parameters in the code path, and I expect CLI users usually use it in a more interactive way, the cli helper makes it easier to discover the parameter changes. Thoughts? @connor-mccarthy

connor-mccarthy: SGTM. What do you think about adding a section to the website docs beneath this one? https://github.com/kubeflow/website/blob/2ca1e2cf634cdd5311f68d06d25674aaf084a87b/content/en/docs/components/pipelines/v2/migration.md#cli-output-change

Probably would only need a sentence or two and I don't think we need to enumerate every break.

connor-mccarthy (commented):
/lgtm

sdk/python/kfp/cli/recurring_run.py
chensun: SG. Will follow up on the documentation changes.

connor-mccarthy (commented):
/lgtm

Thanks for this big update and the associated testing!

PR #9118, opened 2023-04-07
fix(backend): Fix missing parameters in jobs and recurring runs. Closes #9012.

### LINKED ISSUE #9012: [backend] ApiParameter in v1 recurring run is missing.
V1 one-off run with `ApiParameter`

Screenshot 2023-03-17 at 2 00 55 PM

V1 recurring run from same pipeline is missing `ApiParameter`
Screenshot 2023-03-20 at 9 20 30 AM

gkcalat:

Issue: #9012 

When a recurring run is created, backend creates an entry in the `jobs` table and a `ScheduledWorkflow` CRD in k8s cluster. The CRD is then reconciled by persistence agent, which calls the api-server to update the information in the database and create new entries in the `run_details` table. We have to distinguish between v1 `parameters` and `runtime_config.parameters` because the former one is `[]Parameters`, but the later is `map[string]*sturctpb.Value`. When serialized, they have different formats, which is why we store them in separate columns in the DB.

There were several root causes of the missing parameters in recurring runs:
   1. As we switched from v1 `parameters` to `runtime_config.parameters`, we changed it for both argo and v2 templates. For v1 runs without `runtime_config` this resulted in v1 `parameters` being overwritten with empty `runtime_config.parameters`. In v2 runs this resulted in `runtime_config.parameters` being written to the v1 `parameters`.
   2. When creating `ScheduledWorkflow` for v1 template we did not specify `TypeMeta`, which should store Argo metadata (`APIVersion`  and `Kind`).
   3. When serializing parameters stored in `ScheduledWorkflow`, we did not have logic for handling v1 `parameters` vs `runtime_config.parameters`.
   4. We did not have logic in the backend for distinguishing between v1 and v2 templates when writing to the DB.
   5. When converting `Runs` and `Jobs` to their API counterparts, we had inconsistencies in the control flow logic.
   6. We did not have unit-tests for the scenarios described above.

These fixes and a few readability improvements were added in this PR.

- [x] Manually tested on standalone KFP from this branch:
    - [x] v1 template (both `v1beta1` and `v2beta1` API)
    - [x] v2 template (both `v1beta1` and `v2beta1` API)

**Checklist:**


backend/src/apiserver/template/argo_template.go
```diff
@@ -122,16 +122,21 @@ func (t *Argo) ScheduledWorkflow(modelJob *model.Job) (*scheduledworkflow.Schedu
...
 	crdTrigger, err := modelToCRDTrigger(modelJob.Trigger)
 	if err != nil {
 		return nil, err
 	}
 
 	scheduledWorkflow := &scheduledworkflow.ScheduledWorkflow{
+		TypeMeta: metav1.TypeMeta{
+			APIVersion: "argoproj.io/v1alpha1",
+			Kind:       "Workflow",
+		},
```
chensun, line 139: Technically this isn't correct. ScheduledWorkflow itself is not Argo workflow, only part of it (spec.workflow) is.
Should we use our own kind instead, and add it to both v1 and v2 template?

backend/src/common/util/scheduled_workflow.go
```diff
@@ -102,11 +105,23 @@ func (s *ScheduledWorkflow) ConditionSummary() string {
...
+	var params interface{}
 	if s.ScheduledWorkflow.Spec.Workflow == nil {
-		params = make([]swfapi.Parameter, 0)
-	} else {
+		return "", nil
+	}
+	if s.IsV1() {
 		params = s.ScheduledWorkflow.Spec.Workflow.Parameters
+	} else {
+		paramsMap := make(map[string]*structpb.Value, 0)
```
Linchin, line 115: nit: Can we create one less variable by directly using `params = make(map[string]*structpb.Value, 0)`?

backend/src/common/util/scheduled_workflow.go
```diff
@@ -126,3 +141,10 @@ func (s *ScheduledWorkflow) ToStringForStore() string {
 	}
 	return string(swf)
 }
+
+func (s *ScheduledWorkflow) IsV1() bool {
```
Linchin, line 145: Not a strong opinion, but I wonder if we should add a method IsV2() for v2 scheduledworkflows.

backend/src/apiserver/storage/job_store_test.go
```diff
@@ -474,6 +474,7 @@ func TestCreateJob(t *testing.T) {
 		K8SName:     "pp1",
 		Namespace:   "n1",
 		PipelineSpec: model.PipelineSpec{
+			Parameters:   "sample params",
```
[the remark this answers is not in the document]
gkcalat: Fixed. Thank you!

backend/src/apiserver/template/argo_template.go
gkcalat: Good point. Fixed. Thank you!

backend/src/common/util/scheduled_workflow.go
gkcalat: `params` is an `interface{}`. If I assign it to an empty `map`, Golang compiler complains when I try to add elements to it later. I guess this is good enough for now.

backend/src/common/util/scheduled_workflow.go
gkcalat: Thank you. I changed this in response to Chen's suggestion above. You can call `GetVersion()` which covers v1, v2 and beyond.

backend/src/common/util/scheduled_workflow.go
Linchin: I see! Thank you for explaining.

Linchin:
Thank you @gkcalat !
/lgtm
/approve

chensun (approved):
/lgtm

Thanks!

PR #9121, opened 2023-04-07
feat(sdk): Support setting cpu/memory requests.

chensun:

**Checklist:**


sdk/python/kfp/components/pipeline_task.py
```diff
@@ -246,21 +246,69 @@ def set_caching_options(self, enable_caching: bool) -> 'PipelineTask':
...
+                <https://kubernetes.io/docs/tasks/configure-pod-container/assign-cpu-resource/#specify-a-cpu-request-and-a-cpu-limit>`_.
+
+        Returns:
+            Self return to allow chained setting calls.
+        """
+        cpu = self._validate_cpu_request_limit(cpu)
+
+        if self.container_spec is None:
+            raise ValueError(
+                'There is no container specified in implementation')
```
connor-mccarthy, line 288: Does it make sense to convert this to a more user-facing message indicating that cpu can only be set on single-step components, not pipelines used as a component?

connor-mccarthy: Also, maybe it makes sense to move this shared check into the validate method (or a separate validation method) for DRY code? Not sure what's best, but want to mention it.

connor-mccarthy (commented):
/lgtm

Thanks, @chensun.

sdk/python/kfp/components/pipeline_task.py
```diff
@@ -246,6 +246,13 @@ def set_caching_options(self, enable_caching: bool) -> 'PipelineTask':
         self._task_spec.enable_caching = enable_caching
         return self
 
+    def _ensure_container_spec_exists(self) -> None:
```
connor-mccarthy, line 249: Micro improvement suggestion: Could also use a `method_name: str` param to which you pass `self.set_memory_limit.__name__` (or whatever the method is) to include the method name in the error message.

connor-mccarthy (commented):
/lgtm

sdk/python/kfp/components/pipeline_task.py
chensun: Can't really reference `self.func` from within the same `func`. Opted to use inspect to achieve the same goal.

connor-mccarthy: SG. `inspect` works!

Here's what I had in mind. Curious, where did things break down?


```python
class Cls:
    def method(self):
        self.printer(self.method.__name__)

    def printer(self, name):
        print(name)

Cls().method()
>>> 'method'
```

connor-mccarthy (commented):
/lgtm

sdk/python/kfp/components/pipeline_task.py
chensun: Okay, that works, I thought I tried the same and got the error but maybe I missed something else.
Anyway, I think I'll keep the inspect approach in favor of there's less effort for the caller to pass in values.

PR #9123, opened 2023-04-07
chore(frontend): Integrate pipeline / pipeline version functions with KFP v2 API

jlyaoyuli:
Part of #8703


frontend/src/lib/Buttons.ts
```diff
@@ -176,6 +176,30 @@ export default class Buttons {
     return this;
   }
 
+  public deletePipelineVersionV2(
+    // getSelectedIds: () => string[],
```
chensun, line 180: remove?

frontend/src/lib/Buttons.ts
```diff
@@ -176,6 +176,30 @@ export default class Buttons {
     return this;
   }
 
+  public deletePipelineVersionV2(
```
chensun, line 179: Similar question to the what I asked on `PipelinesDialogV2`, do we need both v1 and v2 mehtods, when and where would the v1 button be used?

frontend/src/lib/Buttons.ts
```diff
@@ -654,6 +702,102 @@ export default class Buttons {
       callback(unsuccessfulIds, !unsuccessfulIds.length);
     }
   }
+
+  private _dialogActionHandlerMultiId(
```
chensun, line 706: nit: This method seems to be mostly duplicate code with the existing method `_dialogActionHandler`, and similarly for the next one `_dialogClosedMultiId`. Would it be possible to reduce the duplication by merging the two? Say you can extend the existing method to accept multiple Ids, the existing method would only use the first one and ignore the second value which would be empty/nil for them. And the newly added `_deletePipelineVersionV2` would take advantage of both Ids. WDYT?

frontend/src/lib/Buttons.ts
```diff
@@ -654,6 +702,102 @@ export default class Buttons {
...
+    selectedResourceAndParentIds: Map<string, string>[],
```
chensun, line 707: nit: If I'm not mistaken, each map would have only one key:value pair. so you only need a tuple array: `[string, string][]`. And the possible value would be something like:
```typescript
[['pv-1', 'p-A'], ['pv-2', 'p-A'], ['pv-8', 'p-C']]`
```
Alternatively, if you use map, you only need `Map<string, string>` but not an array of map. And the possible value would be something like:
```typescript
{'pv-1' => 'p-A', 'pv-2' => 'p-A', 'pv-8' => 'p-C'}`
```

frontend/src/pages/NewPipelineVersion.test.tsx
```diff
@@ -42,39 +41,17 @@ describe('NewPipelineVersion', () => {
...
+    display_name: 'original mock pipeline name',
   };
 
   let MOCK_PIPELINE_VERSION = {
-    id: 'original-run-pipeline-version-id',
-    name: 'original mock pipeline version name',
+    pipeline_version_id: 'original-run-pipeline-version-id',
+    display_name: 'original mock pipeline version name',
     description: 'original mock pipeline version description',
-    resource_references: [
```
chensun, line 69: While removing resource-reference, you need to keep the pipeline ID info under `pipeline_id`, so the mock should look like:
```typescript
let MOCK_PIPELINE_VERSION = {
    pipeline_version_id: 'original-run-pipeline-version-id',
    display_name: 'original mock pipeline version name',
    description: 'original mock pipeline version description',
    pipeline_id: 'original-run-pipeline-id',
}
```

frontend/src/pages/NewRunV2.test.tsx
```diff
@@ -266,30 +260,24 @@ describe('NewRunV2', () => {
...
-    getPipelineVersionTemplateSpy.mockImplementation(() =>
-      Promise.resolve({ template: v2YamlTemplateString }),
-    );
+
     render(
       <CommonTestWrapper>
         <NewRunV2
           {...generatePropsNewRun()}
-          existingRunId='e0115ac1-0479-4194-a22d-01e65e09a32b'
+          existingRunId={null}
```
chensun, line 272: What's the reason for this change?

frontend/src/pages/NewRunV2.test.tsx
```diff
@@ -500,12 +465,13 @@ describe('NewRunV2', () => {
...
 
-      const getPipelineSpy = jest.spyOn(Apis.pipelineServiceApi, 'getPipeline');
-      getPipelineSpy.mockImplementation(() => NEW_TEST_PIPELINE);
-
       const expectedPipeline = await screen.findByText(NEW_TEST_PIPELINE_NAME);
       fireEvent.click(expectedPipeline);
 
+      await waitFor(() => {
+        expect(getPipelineSpy).toHaveBeenCalled();
+      });
```
chensun, line 473: For my education, why do we need to add this code now?

frontend/src/pages/PipelineDetails.tsx
chensun, line 1: For my education, is `PipelineDetails.tsx` for rending v1 (argo workflow) pipelines, while `PipelineDetailsV2.tsx` for rending v2 (IR) pipelines?

frontend/src/pages/PipelineDetails.tsx
```diff
@@ -242,20 +269,66 @@ class PipelineDetails extends Page<{}, PipelineDetailsState> {
...
+    } else {
+      // Get the latest version if no version id
+      let listVersionsResponse: V2beta1ListPipelineVersionsResponse;
+      let latesetVersion: V2beta1PipelineVersion;
+      try {
+        listVersionsResponse = await Apis.pipelineServiceApiV2.listPipelineVersions(
+          pipelineId,
+          undefined,
+          undefined,
+          'created_at desc',
```
chensun, line 292: nit: you can also include `limit 1` to reduce the size of the returned payload.

frontend/src/pages/PipelineDetails.tsx
```diff
@@ -434,38 +526,46 @@ class PipelineDetails extends Page<{}, PipelineDetailsState> {
 
     const [graph, reducedGraph, graphV2] = await this._createGraph(templateString);
 
+    // TBD(jlyaoyuli): If we allow upload v1 version to v2 pipeline or vice versa,
```
chensun, line 529: I think with v1beta1 API, we have been allowing uploading both Argo and IR pipelines, so we can't change the behavior. However, in v2beta1 API, we don't need to allow uploading Argo pipelines, we can only accept IR pipelines. 
Not sure if this is the current implementation thought. @gkcalat to confirm.

frontend/src/pages/PipelineDetails.tsx
```diff
@@ -476,15 +576,51 @@ class PipelineDetails extends Page<{}, PipelineDetailsState> {
...
+        versionId,
+      );
+      this.props.history.replace({
+        pathname: `/pipelines/details/${this.state.v2Pipeline.pipeline_id}/version/${versionId}`,
+      });
+
+      const [graph, reducedGraph, graphV2] = await this._createGraph(
+        selectedVersionPipelineTemplate,
+      );
+      if (isFeatureEnabled(FeatureKey.V2_ALPHA) && graphV2.length > 0) {
```
chensun, line 610: Question orthogonal to the PR: as we're going into GA, do we still need to support this feature key? 
/cc @zijianjoy

frontend/src/pages/ResourceSelectorV2.tsx
```diff
@@ -0,0 +1,157 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+// Separate the resource selector between v1 and v2 to avoid breaking current v1 behavior
+// TODO(jlyaoyuli): consider to merge 2 selectors together (change updatedSelection() in v1)
```
chensun, line 18: +1. 
What you have is fine for now. But I do think avoiding code duplication is an important factor to maintain a healthy code base. As discussed offline, the alternative could be having converters to/from `BaseResource` like the api_converter in the BE code. 
We can target such cleanups post-GA.

frontend/src/components/navigators/PipelineVersionCard.tsx
```diff
@@ -92,19 +90,17 @@ export function PipelineVersionCard({
...
+          <Description description={pipeline.description || 'empty pipeline description'} />
 
           {/* selectedVersion is always populated by either selected or pipeline default version if it exists */}
           {selectedVersion && selectedVersion.description ? (
             <>
               <div className='text-gray-900 mt-5'>
-                {selectedVersion.id === apiPipeline.default_version?.id
-                  ? 'Default Version Description'
-                  : 'Version Description'}
+                {selectedVersion.pipeline_version_id === 'Version Description'}
```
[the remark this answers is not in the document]
jlyaoyuli: Great catch! This one should be fixed!

frontend/src/pages/NewRunV2.test.tsx
jlyaoyuli: The `existingRunId` will exist only when `existingRun` is existing, so the previous one is incorrect although it didn't lead to failed tests.

frontend/src/pages/PipelineDetails.tsx
jlyaoyuli: `PIpelineDetails.tsx` is like a **router** which determines to render `PipelineDetailsV1.tsx` (Argo) or `PipelineDetailsV2.tsx` (IR). Maybe we should consider to rename the file to avoid misconception.

frontend/src/lib/Buttons.ts
jlyaoyuli: Merged!

frontend/src/lib/Buttons.ts
jlyaoyuli: Changed to the second options above.

frontend/src/pages/NewRunV2.test.tsx
jlyaoyuli: Simple answer: this is related to ResourceSelectorV2. 

In ResourceSelector(v1), the component will return the selected resource (base resource) back to its parent component (which is PipelineDialog(v1) here). Because `ApiPipeline` has same field name (`id`, `name` etc.), the returned base resource from ResourceSelector can be used directly in PipelineDialog(v1) (note: PipelineDialog(v1) will return the `ApiPipeline` to its parent component NewRun(v1) afterward.

As for V2beta1Pipeline, it has different name field as base resource, so we discuss to create a **converter** which works well when convert `apiListResponse` to base resource. However, it will miss some information when convert back from base resource to API object (pipeline, version, experiment). There are two solutions for this bug: 
1. Add all missing info as optional field in base resource interface in ResourceSelectorV2, then convert it back to API object.
2. (Currently using) return `selectedId` rather than `selected base resource` back to its parent component (PipelineDialogV2). Once the pipeline dialog get the returned selectId from resource selector, it then call getPipeline() API again to retrieve the selectedPipeline which is returned back to parent component (NewRunV2). Also, other selectors like experiment and version in NewRunV2 are using the same logic.

frontend/src/lib/Buttons.ts
jlyaoyuli: This is solved after merging the logic of dialogHandler.

frontend/src/components/PipelinesDialogV2.tsx
```diff
@@ -0,0 +1,172 @@
...
+import Button from '@material-ui/core/Button';
+import Dialog from '@material-ui/core/Dialog';
+import DialogActions from '@material-ui/core/DialogActions';
+import { classes } from 'typestyle';
+import { padding, commonCss } from 'src/Css';
+import DialogContent from '@material-ui/core/DialogContent';
+import ResourceSelectorV2 from 'src/pages/ResourceSelectorV2';
+import { Apis, PipelineSortKeys } from 'src/lib/Apis';
+import { Column } from './CustomTable';
+import { V2beta1Pipeline } from 'src/apisv2beta1/pipeline';
```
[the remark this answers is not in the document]
jlyaoyuli: The reason I choose to separate v1 and v2 here is to avoid changes on v1 implementation. We can definitely merge the logic together, but I am not sure if changing v1 code base (cause some risks of regression) is preferred at this moment.

frontend/src/pages/PipelineDetails.tsx
jlyaoyuli: The current implementation is making v1 and v2 exclusive to avoid mix of Argo and IR under single pipeline (folder).

frontend/src/pages/ResourceSelectorV2.tsx
```diff
@@ -0,0 +1,157 @@
...
+      this.setState(newState as any, cb);
+    }
+  }
+
+  protected _selectionChanged(selectedIds: string[]): void {
+    if (!Array.isArray(selectedIds) || selectedIds.length !== 1) {
+      logger.error(`${selectedIds.length} resources were selected somehow`, selectedIds);
+      return;
+    }
+    this.props.selectionChanged(selectedIds[0]);
```
jlyaoyuli, line 113: /cc @chensun

frontend/src/components/PipelinesDialogV2.tsx
```diff
@@ -0,0 +1,172 @@
...
+          return {
+            nextPageToken: response.next_page_token || '',
+            resources: response.pipelines?.map(p => convertPipelineToResource(p)) || [],
+          };
+        }}
+        columns={props.pipelineSelectorColumns}
+        emptyMessage='No pipelines found. Upload a pipeline and then try again.'
+        initialSortColumn={PipelineSortKeys.CREATED_AT}
+        selectionChanged={async (selectedId: string) => {
+          const selectedPipeline = await Apis.pipelineServiceApiV2.getPipeline(selectedId);
```
jlyaoyuli, line 86: /cc @chensun

chensun (approved):
/lgtm
/approve 

Thanks!

frontend/src/pages/ResourceSelectorV2.tsx
jlyaoyuli: Cleanups for this one is in #9693

PR #9140, opened 2023-04-12
feat(sdk): Support submitting pipeline with platform config.

chensun:

**Checklist:**


chensun:
rebase and review after https://github.com/kubeflow/pipelines/pull/9112

sdk/python/kfp/client/client.py
```diff
@@ -1666,3 +1627,71 @@ def validate_pipeline_resource_name(name: str) -> None:
...
+
+    def _safe_load_yaml(stream: TextIO) -> _PipelineDoc:
+        docs = yaml.safe_load_all(stream)
+        pipeline_spec_dict = None
+        platform_spec_dict = {}
+        for doc in docs:
+            if pipeline_spec_dict is None:
+                pipeline_spec_dict = doc
+            else:
+                platform_spec_dict.update(doc)
```
connor-mccarthy, line 1657: nit: Would this logic be simpler and equivalent if we check if `len(list(docs))` is 1 or 2, then construct `_PipelineDoc`? Something like [this](https://github.com/kubeflow/pipelines/blob/master/sdk/python/kfp/components/structures.py#L1052). IIUC, this is done implicitly in the current implementation via the expectation that `pipeline_spec_dict` will be `None` on iteration 1 and `not None` on iteration 2.

sdk/python/kfp/client/client.py
```diff
@@ -1666,3 +1627,71 @@ def validate_pipeline_resource_name(name: str) -> None:
...
+    else:
+        raise ValueError(
+            f'The package_file {package_file} should end with one of the '
+            'following formats: [.tar.gz, .tgz, .zip, .yaml, .yml].')
+
+
+def _override_caching_options(
+    pipeline_spec: pipeline_spec_pb2.PipelineSpec,
+    enable_caching: bool,
+) -> None:
```
connor-mccarthy, line 1689: +1 on extracting these to functions

connor-mccarthy (commented):
/lgtm

Thanks, @chensun!

sdk/python/kfp/client/client.py
chensun: I was thinking it could be more than 2, so I implemented a dict merge logic for platform configs spread across multi docs, e.g.:
```yaml
<IR>
---
platforms:
    kubernetes:
        ...
---
platforms:
   vertex:
       ...
---
platforms:
    aws:
       ....
```
This should be a legitimate structure if we consider the "overlay"-like concept. So I should probably change [this](https://github.com/kubeflow/pipelines/blob/master/sdk/python/kfp/components/structures.py#L1052) as well. WDYT?

chensun: Okay, I guessing per current compiler implementation, it would only produce two docs even there're multiple platforms:
```yaml
<IR>
---
platforms:
   kubernetes:
        ...
   vertex:
       ...
   aws:
       ....
```
I'm thinking maybe in future we should consider the more than 2 docs structure. The benefits in my opinion is if I took a platform template, only care about one platform config, it's slightly easier to strip off other platform configs.

connor-mccarthy: SGTM. I don't see a strong need for this currently, but I'm definitely open to permitting >1 platform config document in the future.

I think there is probably a slight maintenance cost associated with having more than one way to represent the same information, but this cost is probably minimal.

chensun: SG. I'll keep both loading code as-is for now. Both can handle 2 docs correctly at this moment, and once we have a clear needs on whether we need multi-doc support, we may consolidate the two methods.

connor-mccarthy: SGTM 👍🏼

PR #9150, opened 2023-04-13
fix(backend): fix volume bug

Linchin:
With the last PR for volume mount (#9106), when a task writes to the PVC, the later tasks cannot find the same file. This was caused by not specifying `PersistentVolumeClaim` in the volume field in [`PodSpec`](https://pkg.go.dev/k8s.io/api/core/v1#PodSpec). 

* Fix bug that files written to the pvc are not found in the later tasks
* Add a v2 samples test
* Update the driver image hash

Note: 
- GCE storage does not support access mode "ReadWriteMany". Use "ReadWriteOnce" instead. We need to specify this in the documentation.  [(reference)](https://stackoverflow.com/questions/57798267/kubernetes-persistent-volume-access-modes-readwriteonce-vs-readonlymany-vs-read#:~:text=3.,modes%20are%20ReadWriteOnce%20and%20ReadOnlyMany.)
**Checklist:**


samples/test/config.yaml
```diff
@@ -130,6 +130,8 @@
   path: samples.v2.component_with_optional_inputs_test
 - name: pipeline_with_env
   path: samples.v2.pipeline_with_env_test
+  - name: pipeline_with_volume
```
gkcalat, line 133: Redundant code indentation?

chensun (approved):
/lgtm
/approve

backend/src/v2/driver/driver.go
gkcalat, line 1: Thank you, @Linchin!

I got a few questions that would help me better understand how this feature works.

It looks like we only need to provide the name and mount path of a pvc in order to generate a podSpec patch. That assumes that `pvc` properties such as class-type, access modes, etc. are set at creation. Is that correct?

What about a case when someone wants to use an existing `pvc` in `read-only` mode (e.g. reading a training dataset or an artifact from another pipeline)? Would creating a pvc fail in that case?

/lgtm
/approve

gkcalat (approved):

Linchin:
@gkcalat, thank you for your comments, and great questions!
>  That assumes that pvc properties such as class-type, access modes, etc.
are set at creation. Is that correct?
Yes, that is correct.

>  What about a case when someone wants to use an existing pvc in read-only mode
(e.g. reading a training dataset or an artifact from another pipeline)?
Would creating a pvc fail in that case?
In this case, the pvc already exists, so we don't have to create a new one.
It is possible to mount a pvc as read-only mode, using the `ReadOnly` field
[here](
https://pkg.go.dev/k8s.io/api/core/v1#PersistentVolumeClaimVolumeSource),
but our kubernetes package did not expose this option.

PR #9184, opened 2023-04-19
chore(backend): change optional parameter enforce rule

Linchin:
Changes:
* Updates rules to resolve optional parameters
* Updates driver image hash

Tests:
* Manually verified that the error message does not exist when `isOptional == false` and default value provided.
* Manually verified that there is an error message when when `isOptional == false` and default value is not provided.

Description:
In #8765, we enforced a rule that when the user does not provide a value for a parameter, this parameter has to be optional. If it is not marked as optional, driver will report an error. Code is [here](https://github.com/kubeflow/pipelines/blob/de060a9816d0e0331e60a057a0c785cac0df18c8/backend/src/v2/driver/driver.go#L683-L685).
In order to accommodate older versions of pipeline spec where `isOptional` field does not exist, we no longer enforce the same rule. Instead, driver looks for a default value. If the default value exists, we use it. If the default value does not exist, the driver reports an error.

**Checklist:**


gkcalat (approved):
Hi @Linchin!

Thank you for the quick fix and testing it manually.

/lgtm
/approve

PR #9189, opened 2023-04-20
fix(backend): Cleanup pipeline creation and upload logic

gkcalat:

- Added deletion of parent pipeline when a pipeline version creation fails
- Improved error codes and messages (e.g. invalid pipeline URL)
- Fixed unit-tests

**Checklist:**


backend/src/apiserver/server/pipeline_server.go
```diff
@@ -191,11 +191,14 @@ func (s *PipelineServer) CreatePipelineV1(ctx context.Context, request *apiv1bet
 	}
 
 	// A new pipeline has been created, but pipeline version creation failed after validation
+	// This block should not be executed when an existing pipeline is found
```
chensun, line 194: This code is inside `CreatePipelineV1`. Shouldn't we change this behavior only in the v2 method, and keep v1 behavior as-is--whatever that is in v1.8.* before we refactor?

backend/src/apiserver/server/pipeline_server.go
```diff
@@ -191,11 +191,14 @@ func (s *PipelineServer) CreatePipelineV1(ctx context.Context, request *apiv1bet
...
 	if perr == nil {
```
chensun, line 195: Assuming this logic will be move to v2 method, this check should be `if perr != nil`, right?

backend/src/apiserver/server/pipeline_server.go
```diff
@@ -191,11 +191,14 @@ func (s *PipelineServer) CreatePipelineV1(ctx context.Context, request *apiv1bet
...
-		return toApiPipelineV1(createdPipeline, &model.PipelineVersion{}), nil
+		if cleanupErr := s.deletePipeline(ctx, createdPipeline.UUID); cleanupErr != nil {
```
chensun, line 196: While the reverting logic is what have discussed and it should be fine in probably most if not all cases, a better solution would be make creating pipeline and pipeline version a DB transaction, so that no reverting would ever be needed. WDYT? 
While it should be technically feasible, I'm not sure the implementation cost. So I'm fine to have a reverting logic for now if the transaction logic requires too much effort to change.

backend/src/apiserver/server/pipeline_server.go
chensun: Another question: the issue reported was upload, will `upload` method eventually call into such `create`, or we need to change `upload` method as well?

backend/src/apiserver/server/pipeline_server.go
```diff
@@ -610,6 +613,9 @@ func (s *PipelineServer) createPipelineVersion(ctx context.Context, pv *model.Pi
 
 	resp, err := s.httpClient.Get(pipelineUrl.String())
 	if err != nil || resp.StatusCode != http.StatusOK {
+		if err == nil {
```
Linchin, line 616: Looks like we can simplify the logic here into `if err == nil && resp.StatusCode != http.StatusOK`?

backend/src/apiserver/server/pipeline_server.go
Linchin: > the issue reported was upload, will upload method eventually call into such create, or we need to change upload method as well?
The upload method directly calls resource manager, and a cleaning logic is added as well.

backend/src/apiserver/server/pipeline_server_test.go
```diff
@@ -243,22 +264,9 @@ func TestCreatePipelineV1_ExistingPipeline(t *testing.T) {
...
-	assert.NotNil(t, newPipelineVersion)
-	assert.Equal(t, newPipelineVersion.PipelineId, existingPipeline.GetId())
-	assert.Equal(t, newPipelineVersion.PipelineId, createdPipeline.GetId())
-
-	assert.Equal(t, "", existingPipeline.GetDefaultVersion().GetId())
-	assert.Equal(t, 0, int(existingPipeline.GetDefaultVersion().GetCreatedAt().Seconds))
-
-	assert.Equal(t, newPipelineVersion.UUID, createdPipeline.GetDefaultVersion().GetId())
-	assert.Less(t, 0, int(createdPipeline.GetDefaultVersion().GetCreatedAt().Seconds))
+	assert.NotNil(t, err)
```
Linchin, line 267: From the result of the test, this looks like a bebavior change with `CreatePipelineV1`. I understand the new behavior may make more sense, but I'm not sure if we should change how v1 works at this point.

backend/src/apiserver/server/pipeline_server.go
gkcalat: Good point thank you.

Let me implement a single-transaction call for creating v1 pipelines.

backend/src/apiserver/server/pipeline_server_test.go
gkcalat: Agree on the behavior change. I will try to avoid it in the next iteration, but it might be infeasible due to the difference in `Pipeline` across v1 and v2. Besides, some of these changes were introduced due to refactoring and API changes.

backend/src/apiserver/server/pipeline_server.go
gkcalat: In this case, `perr` is `nil` only when a new pipeline is created - no existing pipelines with such (name, namespace)

Linchin:
All looks good, just one small thing: maybe add a test for the newly added function `CreatePipelineAndPipelineVersion()`?

backend/src/apiserver/server/pipeline_server.go
gkcalat: It turned out out that we actually need to change the `CreatePipelineV1` in order to be consistent with the old v1 behavior. I may not be exposed via SDK or UI, but I propose to maintain it the same to stay consistent. The change was probably introduced in the recent refactoring. I changed it back.

Linchin:
Thank you @gkcalat! 
/lgtm
/approve

PR #9192, opened 2023-04-20
fix(frontend): Render graph for yaml file containing platform_spec

jlyaoyuli:
Previously, we directly use the return value from safeLoad(), which is PipelineSpec, to draw the graph. However, after the concept `platform_spec` is introduced, the return value from safeLoad() will become a nested structure, and the PipelineSpec we use to draw graph is stored in `pipeline_spec` field.

### DIFF
--- a/frontend/src/lib/v2/WorkflowUtils.ts
+++ b/frontend/src/lib/v2/WorkflowUtils.ts
@@ -19,7 +19,8 @@ import { convertFlowElements } from 'src/lib/v2/StaticFlow';
 import * as WorkflowUtils from 'src/lib/v2/WorkflowUtils';
 import { Workflow } from 'src/third_party/mlmd/argo_template';
 
-export const PIPELINE_SPEC = 'pipeline_spec';
+// This key can be used to retrieve 'pipeline_spec' in yaml file containing 'platform_spec'
+export const PIPELINE_SPEC_TEMPLATE_KEY = 'pipeline_spec'; 
 
 export function isV2Pipeline(workflow: Workflow): boolean {
   return workflow?.metadata?.annotations?.['pipelines.kubeflow.org/v2_pipeline'] === 'true';
@@ -35,7 +36,7 @@ export function isArgoWorkflowTemplate(template: Workflow): boolean {
 export function isTemplateV2(templateString: string): boolean {
   try {
     const template =
-      jsyaml.safeLoad(templateString)[PIPELINE_SPEC] ?? jsyaml.safeLoad(templateString);
+      jsyaml.safeLoad(templateString)[PIPELINE_SPEC_TEMPLATE_KEY] ?? jsyaml.safeLoad(templateString);
     if (isArgoWorkflowTemplate(template)) {
       return false;
     } else if (isFeatureEnabled(FeatureKey.V2_ALPHA)) {
@@ -53,7 +54,7 @@ export function isTemplateV2(templateString: string): boolean {
 export function convertYamlToV2PipelineSpec(template: string): PipelineSpec {
   // If pipeline_spec exists in the return value of safeload, which means the original yaml contains
   // platform_spec, then the pipeline spec is stored in pipeline_spec field.
-  const pipelineSpecYAML = jsyaml.safeLoad(template)['pipeline_spec'] ?? jsyaml.safeLoad(template);
+  const pipelineSpecYAML = jsyaml.safeLoad(template)[PIPELINE_SPEC_TEMPLATE_KEY] ?? jsyaml.safeLoad(template);
   const ts_pipelinespec = PipelineSpec.fromJSON(pipelineSpecYAML);
   if (!ts_pipelinespec.root || !ts_pipelinespec.pipelineInfo || !ts_pipelinespec.deploymentSpec) {
     throw new Error('Important infomation is missing. Pipeline Spec is invalid.');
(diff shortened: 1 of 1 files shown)


frontend/src/lib/v2/WorkflowUtils.ts
```diff
@@ -48,7 +49,9 @@ export function isTemplateV2(templateString: string): boolean {
 
 // Assuming template is the JSON format of PipelineSpec in api/v2alpha1/pipeline_spec.proto
 export function convertYamlToV2PipelineSpec(template: string): PipelineSpec {
-  const pipelineSpecYAML = jsyaml.safeLoad(template);
+  // If pipeline_spec exists in the return value of safeload, which means the original yaml contains
+  // platform_spec, then the pipeline spec is stored in pipeline_spec field.
+  const pipelineSpecYAML = jsyaml.safeLoad(template)['pipeline_spec'] ?? jsyaml.safeLoad(template);
```
zijianjoy, line 54: 'pipeline_spec' is a constant which is defined by the IR. Recommend to define a global variable for `pipeline_spec` to be used consistently across the FE.

frontend/src/lib/v2/WorkflowUtils.ts
```diff
@@ -48,7 +49,9 @@ export function isTemplateV2(templateString: string): boolean {
 
 // Assuming template is the JSON format of PipelineSpec in api/v2alpha1/pipeline_spec.proto
 export function convertYamlToV2PipelineSpec(template: string): PipelineSpec {
-  const pipelineSpecYAML = jsyaml.safeLoad(template);
+  // If pipeline_spec exists in the return value of safeload, which means the original yaml contains
+  // platform_spec, then the pipeline spec is stored in pipeline_spec field.
```
zijianjoy, line 53: This is a temporary solution which omit platform-specific proto. However, KFP frontend should support kubernetes platform-specific proto because it contains information that defines this pipeline. Recommend creating a bug to follow up on this item.

jlyaoyuli: /cc @chensun

chensun: James was correct that we need to read in kubernetes specific config from the `platform_spec` section. This would be needed to support volume for example:
https://github.com/kubeflow/pipelines/blob/6a55449ffb5b3d935b32033dc9a13b1852c059a4/kubernetes_platform/python/test/snapshot/data/create_mount_delete_dynamic_pvc.yaml#L180-L181

frontend/src/lib/v2/WorkflowUtils.ts
```diff
@@ -69,7 +74,8 @@ export function isPipelineSpec(templateString: string) {
     return false;
   }
   try {
-    const template = jsyaml.safeLoad(templateString);
+    const template =
+      jsyaml.safeLoad(templateString)['pipeline_spec'] ?? jsyaml.safeLoad(templateString);
```
zijianjoy, line 78: We should use the global constant to replace `'pipeline_spec'` here.

frontend/src/lib/v2/WorkflowUtils.ts
```diff
@@ -32,7 +34,8 @@ export function isArgoWorkflowTemplate(template: Workflow): boolean {
 
 export function isTemplateV2(templateString: string): boolean {
   try {
-    const template = jsyaml.safeLoad(templateString);
+    const template =
+      jsyaml.safeLoad(templateString)[PIPELINE_SPEC] ?? jsyaml.safeLoad(templateString);
```
zijianjoy, line 38: I feel that this line should be abstracted to a function so it can be reused everywhere. (And with a test case to demonstrate how this logic differentiate single yaml file vs multiple yaml files. This piece of behavior can be hard to understand because it relies on an unusual API behavior.

frontend/src/lib/v2/WorkflowUtils.ts
```diff
@@ -19,6 +19,8 @@ import { convertFlowElements } from 'src/lib/v2/StaticFlow';
 import * as WorkflowUtils from 'src/lib/v2/WorkflowUtils';
 import { Workflow } from 'src/third_party/mlmd/argo_template';
 
+export const PIPELINE_SPEC = 'pipeline_spec';
```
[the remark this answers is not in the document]
jlyaoyuli: Great suggestion!

frontend/src/lib/v2/WorkflowUtils.ts
```diff
@@ -19,6 +19,9 @@ import { convertFlowElements } from 'src/lib/v2/StaticFlow';
 import * as WorkflowUtils from 'src/lib/v2/WorkflowUtils';
 import { Workflow } from 'src/third_party/mlmd/argo_template';
 
+// This key can be used to retrieve 'pipeline_spec' in yaml file containing 'platform_spec'
```
zijianjoy, line 22: Why is this key being used to retrieved `'platform_spec'`? This key is for retrieving pipeline definition.

jlyaoyuli: I mean for the yaml file that contains 'platform_spec', we should use this key to retrieve `pipeline_spec`.

zijianjoy:
/lgtm
/approve

Thank you Joe for the fix!

PR #9193, opened 2023-04-20
fix(frontend): fail to clone SDK-created run / recurring run

jlyaoyuli:
This is a regression bug caused by #9123. Main reason is the incorrect in-line expression used to determine template string. This PR fixes the logic and add short comment to describe the source of template string and specify the priority!


frontend/src/pages/NewRunSwitcher.tsx
```diff
@@ -147,10 +147,13 @@ function NewRunSwitcher(props: PageProps) {
...
 
+  // Three possible source for template string (priority from high to low)
+  // 1. pipelineManifest: pipeline_spec stored in run or recurring run
+  // 2. templateStrFromSpec: pipeline_spec stored in pipeline_version
+  // 3. templateStrFromTemplate(v1): template in the response of getPipelineVersionTemplate()
   const templateString =
-    pipelineManifest ?? isTemplateV2(templateStrFromSpec)
-      ? templateStrFromSpec
-      : templateStrFromTemplate;
+    pipelineManifest ??
```
zijianjoy, line 155: Can you add a test to validate the priority in assigning templateString?

PR #9228, opened 2023-04-24
feat(backend): add parameter verification before submitting the workflow

Linchin:
Adds a verification function that ensures the parameters are valid before the workflow is submitted to the client.

* Add the verification function
* Update the tests
* Add a test case 

**Checklist:**


backend/src/apiserver/server/job_server_test.go
```diff
@@ -415,7 +415,7 @@ func TestCreateJob_V2(t *testing.T) {
 
 	// Test all parameters types converted to model.RuntimeConfig.Parameters, which is string type
 	v2RuntimeParams := map[string]*structpb.Value{
-		"param1": {Kind: &structpb.Value_StringValue{StringValue: "world"}},
+		"text":   {Kind: &structpb.Value_StringValue{StringValue: "world"}},
```
chensun, line 418: nit: This change shouldn't be necessary? Looks a bit odd the list is `text`, `param2`, `param3`, ...

backend/src/apiserver/server/job_server_test.go
```diff
@@ -877,6 +878,9 @@ func TestCreateRecurringRun(t *testing.T) {
 		PipelineSource: &apiv2beta1.RecurringRun_PipelineSpec{PipelineSpec: pipelineSpecStruct},
 		RuntimeConfig: &apiv2beta1.RuntimeConfig{
 			PipelineRoot: "model-pipeline-root",
+			Parameters: map[string]*structpb.Value{
+				"text": &structpb.Value{Kind: &structpb.Value_StringValue{StringValue: "world"}},
```
chensun, line 882: nit: this can be simplified as `"text": structpb.NewStringValue("world"),`

same for below

backend/src/apiserver/resource/resource_manager_test.go
```diff
@@ -4089,8 +4074,8 @@ root:
   inputDefinitions:
     parameters:
       text:
-        type: STRING
-schemaVersion: 2.0.0
+        parameterType: STRING
```
chensun, line 4077: This change is incomplete, the component spec above would have the same input parameter type definition.

backend/src/apiserver/server/pipeline_upload_server_test.go
```diff
@@ -624,7 +624,7 @@ root:
   inputDefinitions:
     parameters:
       text:
-        type: STRING
-schemaVersion: 2.0.0
+        parameterType: STRING
```
chensun, line 627: ditto: incomplete IR change

backend/src/apiserver/template/v2_template.go
```diff
@@ -256,7 +259,9 @@ func (t *V2Spec) RunWorkflow(modelRun *model.Run, options RunWorkflowOptions) (u
 		return nil, util.Wrap(err, "Failed to convert to PipelineJob RuntimeConfig")
 	}
 	job.RuntimeConfig = jobRuntimeConfig
-
+	if err = t.validatePipelineJobInputs(job); err != nil {
+		return nil, util.Wrap(err, fmt.Sprintf("invalid pipeline job: %s", err.Error()))
```
chensun, line 263: nit: invalid pipeline job inputs

backend/src/apiserver/template/v2_template.go
```diff
@@ -306,3 +311,55 @@ func IsPlatformSpecWithKubernetesConfig(template []byte) bool {
 	_, ok := platformSpec.Platforms["kubernetes"]
 	return ok
 }
+
+func (t *V2Spec) validatePipelineJobInputs(job *pipelinespec.PipelineJob) error {
+	// The pipeline requires no input
```
chensun, line 316: If the pipeline has no input, we should validate that runtime config also doesn't contain any parameter.

backend/src/apiserver/template/v2_template.go
```diff
@@ -306,3 +311,55 @@ func IsPlatformSpecWithKubernetesConfig(template []byte) bool {
...
+			if !param.GetIsOptional() && param.GetDefaultValue() == nil {
+				return fmt.Errorf("parameter %s is not optional, yet has neither default value nor user provided value", name)
+			}
+		} else {
+			// Verify the parameter type is correct
+			switch param.GetParameterType() {
+			case pipelinespec.ParameterType_PARAMETER_TYPE_ENUM_UNSPECIFIED:
+				return fmt.Errorf("input parameter %s has unspecified type", name)
+			case pipelinespec.ParameterType_NUMBER_DOUBLE, pipelinespec.ParameterType_NUMBER_INTEGER:
+				if _, ok := input.GetKind().(*structpb.Value_NumberValue); !ok {
```
chensun, line 338: I wonder if we can actually get the kind of an input value. Can you test this via UI, when you put a number in the input box, do this get numbervalue here or stringvalue?

backend/src/apiserver/template/v2_template.go
```diff
@@ -306,3 +311,55 @@ func IsPlatformSpecWithKubernetesConfig(template []byte) bool {
...
+					return fmt.Errorf("input parameter %s has type double or integer, but the input parameter is not of Value_NumberValue type", name)
```
chensun, line 339: nit: would probably avoid using `Value_NumberValue` in the user facing error message.

backend/src/apiserver/template/v2_template.go
```diff
@@ -306,3 +311,55 @@ func IsPlatformSpecWithKubernetesConfig(template []byte) bool {
...
+			case pipelinespec.ParameterType_LIST:
+				if _, ok := input.GetKind().(*structpb.Value_ListValue); !ok {
+					return fmt.Errorf("input parameter %s has type list, but the input parameter is not of Value_ListValue type", name)
+				}
+			case pipelinespec.ParameterType_STRUCT:
+				if _, ok := input.GetKind().(*structpb.Value_StructValue); !ok {
+					return fmt.Errorf("input parameter %s has type struct, but the input parameter is not of Value_StructValue type", name)
+				}
+			case pipelinespec.ParameterType_TASK_FINAL_STATUS:
+				return fmt.Errorf("input parameter %s has type TASK_FINAL_STATUS, which is invalid for root component", name)
```
chensun, line 358: Can we raise a not implemented error instead? We may support this later

backend/src/apiserver/server/job_server_test.go
Linchin: It is necessary because the pipeline calls for a parameter `text`.

backend/src/apiserver/server/job_server_test.go
Linchin: Good point! I will change this.

backend/src/apiserver/template/v2_template.go
Linchin: Thank you! Updated.

backend/src/apiserver/template/v2_template.go
Linchin: I think as long as the users provided the parameters we need, it's ok if there's more. This makes our verification simpler and gives more leeway for the users. WDYT?

backend/src/apiserver/template/v2_template.go
Linchin: I definitely can. Could you help me understand how a task output can be the input of root? I always thought the root has to run first, so there can be no task that generates an output before root.

backend/src/apiserver/server/pipeline_upload_server_test.go
Linchin: Updated.

backend/src/apiserver/resource/resource_manager_test.go
Linchin: Thank you! Just updated.

backend/src/apiserver/server/pipeline_upload_server_test.go
Linchin: Done.

backend/src/apiserver/template/v2_template.go
Linchin: I understand that `Value_NumberValue` is more of some internal info, but it feels helpful that that users know what the required type is. Certainly it would be much better if the types of parameter and inputs are represented in the same way.

backend/src/apiserver/template/v2_template.go
chensun: You're right, this error message is fine. The not implemented error should be raised in some other places.

backend/src/apiserver/template/v2_template.go
Linchin: I did borrow this logic from Vertex. They only verify that the needed parameters are provided, and do not care about extra parameters users give. But I think you have a good point that this may affect the speed of users' debugging.

backend/src/apiserver/template/v2_template.go
Linchin: Thank you, I just verified it.

backend/src/apiserver/template/v2_template.go
Linchin: Sorry, I didn't make it very clear here. On the left it should say 
<pre>
input parameter %s <b>requires</b> type double or integer
</pre>
This is the type the pipeline calls for. This type maps to a `structpb.Value` "subtype", which is the user provided value, such as double/integer mapping to `Value_NumberValue`.

backend/src/apiserver/resource/resource_manager_test.go
Linchin: Sorry, I will add this back.

backend/src/apiserver/template/v2_template.go
Linchin: Missed this error message :) will add it.

backend/src/apiserver/server/job_server_test.go
Linchin: I think you are right. I have changed the parameter name to `param1`.

backend/src/apiserver/resource/resource_manager_test.go
```diff
@@ -231,7 +231,7 @@ func initWithJobV2(t *testing.T) (*FakeClientManager, *ResourceManager, *model.J
 		PipelineSpec: model.PipelineSpec{
 			PipelineSpecManifest: v2SpecHelloWorld,
 			RuntimeConfig: model.RuntimeConfig{
-				Parameters:   "{\"param1\":\"world\"}",
+				Parameters:   "{\"text\":\"world\"}",
```
chensun, line 234: I missed this earlier, but `parameters` belongs to schemaVersion 2.0.0, and is deprecated. Starting from schema 2.1.0, it should be `parameter_values` instead.
https://github.com/kubeflow/pipelines/blob/02ae214223070694624f634e6140a3d0aef92fe5/api/v2alpha1/pipeline_spec.proto#L29-L43

chensun (commented):
/lgtm

backend/src/apiserver/resource/resource_manager_test.go
Linchin: Good catch!! I will change accordingly.

backend/src/apiserver/template/v2_template.go
```diff
@@ -306,3 +312,75 @@ func IsPlatformSpecWithKubernetesConfig(template []byte) bool {
...
+	// but we still need to verify that no extra parameters are provided.
+	requiredParams := make(map[string]*pipelinespec.ComponentInputsSpec_ParameterSpec)
+	if t.spec.GetRoot().GetInputDefinitions() != nil {
+		requiredParams = t.spec.GetRoot().GetInputDefinitions().GetParameters()
+	}
+
+	runtimeConfig := job.GetRuntimeConfig()
+	if runtimeConfig == nil {
+		if len(requiredParams) != 0 {
+			return fmt.Errorf("pipeline requiring input has empty runtime config")
```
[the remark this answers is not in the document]
Linchin: Good point. I will add this.

backend/src/apiserver/resource/resource_manager_test.go
Linchin: Since this will be a breaking change for UI and SDK, we won't implement for now.

chensun (approved):
/lgtm
/approve 

Thanks!

PR #9231, opened 2023-04-25
fix(frontend): Missing pipeline info while cloning recurring run.

jlyaoyuli:
In the new run page, we allow user to change experiment selection. With the updated experiment, search string in URL is changed as well. Previously, the scenario of cloning recurring run is not handled, which lead to the missing pipeline information issue after changing experiment.

Before:

https://user-images.githubusercontent.com/56132941/234348697-7d5a2d46-ca27-4450-8959-7b8c10c89c24.mov

After:

https://user-images.githubusercontent.com/56132941/234348765-b6b22b63-c3e3-49f6-8c54-ebda24b4113d.mov


chensun (approved):
/lgtm
/approve

PR #9255, opened 2023-04-27
fix(frontend): fail to clone SDK-created run / recurring run

jlyaoyuli:
This is a regression bug caused by https://github.com/kubeflow/pipelines/pull/9123. Main reason is the incorrect in-line expression used to determine template string. This PR fixes the logic and add short comment to describe the source of template string and specify the priority!

### DIFF
--- a/frontend/src/pages/NewRunV2.test.tsx
+++ b/frontend/src/pages/NewRunV2.test.tsx
@@ -238,12 +238,15 @@ describe('NewRunV2', () => {
   const updateDialogSpy = jest.fn();
   const updateSnackbarSpy = jest.fn();
   const updateToolbarSpy = jest.fn();
-  function generatePropsNewRun(): PageProps {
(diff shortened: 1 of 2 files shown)


frontend/src/pages/NewRunV2.test.tsx
```diff
@@ -40,6 +40,8 @@ import { convertYamlToV2PipelineSpec } from 'src/lib/v2/WorkflowUtils';
 import NewRunV2 from './NewRunV2';
 import { PageProps } from './Page';
 import * as JsYaml from 'js-yaml';
+import NewRunSwitcher from './NewRunSwitcher';
+import * as features from '../features';
```
zijianjoy, line 44: NIT: Use absolute path for imports.

zijianjoy:
/lgtm
/approve

Good work Joe!

PR #9257, opened 2023-04-27
fix(frontend): Add name space pre-check function for getArtifactServiceGetter().

jlyaoyuli:
Currently, the frontend server doesn't check namespace before getting the getArtifact() request, which might cause potential risk of attack from hacker. The main purpose of this PR is adding a pre-check function to make sure that namespace follows the naming rule of k8s.

Ref: https://kubernetes.io/docs/concepts/overview/working-with-objects/names/#dns-label-names

### DIFF
--- a/frontend/server/handlers/artifacts.ts
+++ b/frontend/server/handlers/artifacts.ts
@@ -383,11 +383,18 @@ function getNamespaceFromUrl(path: string): string | undefined {
   return params.get('namespace') || undefined;
 }
 
+function isValidNameSpace(namespace: string) {
+  return (
+    namespace.length <= 63 && namespace.match('[a-z0-9]([-a-z0-9]*[a-z0-9])?')?.includes(namespace)
+  );
+}
+
 // `new URL('/path')` doesn't work, because URL only accepts full URL with scheme and hostname.
 // We use the DUMMY_BASE_PATH like `new URL('/path', DUMMY_BASE_PATH)`, so that URL can parse paths
 // properly.
 const DUMMY_BASE_PATH = 'http://dummy-base-path';
 
 export function getArtifactServiceGetter({ serviceName, servicePort }: ArtifactsProxyConfig) {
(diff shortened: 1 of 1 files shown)


frontend/server/handlers/artifacts.ts
```diff
@@ -383,11 +383,18 @@ function getNamespaceFromUrl(path: string): string | undefined {
...
+
 // `new URL('/path')` doesn't work, because URL only accepts full URL with scheme and hostname.
 // We use the DUMMY_BASE_PATH like `new URL('/path', DUMMY_BASE_PATH)`, so that URL can parse paths
 // properly.
 const DUMMY_BASE_PATH = 'http://dummy-base-path';
 
 export function getArtifactServiceGetter({ serviceName, servicePort }: ArtifactsProxyConfig) {
-  return (namespace: string) => `http://${serviceName}.${namespace}:${servicePort}`;
+  return (namespace: string) =>
+    isValidNameSpace(namespace) ? `http://${serviceName}.${namespace}:${servicePort}` : '';
```
zijianjoy, line 399: If this is not valid, we should throw exception to tell users about it, right? Can we throw exception here instead of returning empty string?

frontend/server/handlers/artifacts.ts
```diff
@@ -383,11 +383,18 @@ function getNamespaceFromUrl(path: string): string | undefined {
   return params.get('namespace') || undefined;
 }
 
+function isValidNameSpace(namespace: string) {
```
zijianjoy, line 386: Add a test in https://github.com/kubeflow/pipelines/blob/6a55449ffb5b3d935b32033dc9a13b1852c059a4/frontend/server/integration-tests/artifact-proxy.test.ts#L11 to validate this change.

frontend/server/handlers/artifacts.ts
chensun, line 399: This url format itself, with namespace at the end, looks rather odd to me. Is this something defined by us? If so, can we change it?

PR #9287, opened 2023-05-02
[WIP] test: Add v2 upgrade test

chensun:
Part of https://github.com/kubeflow/pipelines/issues/6168

**Checklist:**

### DIFF
--- a/backend/test/v2/integration/upgrade_test.go
+++ b/backend/test/v2/integration/upgrade_test.go
@@ -513,7 +513,11 @@ func (s *UpgradeTests) VerifyCreatingRunsAndJobs() {
 			PipelineID:        pipelineVersions[0].PipelineID,
 			PipelineVersionID: pipelineVersions[0].PipelineVersionID,
 		},
-		RuntimeConfig:  &recurring_run_model.V2beta1RuntimeConfig{},
+		RuntimeConfig: &recurring_run_model.V2beta1RuntimeConfig{
+			Parameters: map[string]interface{}{
+				"url": "gs://ml-pipeline-playground/shakespeare1.txt",
+			},
+		},
 		MaxConcurrency: 10,
 		Mode:           recurring_run_model.RecurringRunModeENABLE,
 	}}
--- a/backend/test/v2/integration/upgrade_test.go
+++ b/backend/test/v2/integration/upgrade_test.go
@@ -513,11 +513,7 @@ func (s *UpgradeTests) VerifyCreatingRunsAndJobs() {
 			PipelineID:        pipelineVersions[0].PipelineID,
 			PipelineVersionID: pipelineVersions[0].PipelineVersionID,
 		},
-		RuntimeConfig: &recurring_run_model.V2beta1RuntimeConfig{
-			Parameters: map[string]interface{}{
-				"url": "gs://ml-pipeline-playground/shakespeare1.txt",
-			},
-		},
+		RuntimeConfig:  &recurring_run_model.V2beta1RuntimeConfig{},
 		MaxConcurrency: 10,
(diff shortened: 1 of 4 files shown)


backend/test/v2/integration/upgrade_test.go
```diff
@@ -0,0 +1,601 @@
...
+		s.resourceNamespace)
+	require.Nil(t, err)
+	assert.True(t, len(runs) >= 1)
+	assert.Equal(t, "hello world", runs[0].DisplayName)
+
+	/* ---------- Get hello world run ---------- */
+	helloWorldRunDetail, err := s.runClient.Get(&runParams.GetRunParams{RunID: runs[0].RunID})
+	require.Nil(t, err)
+	assert.Equal(t, "hello world", helloWorldRunDetail.DisplayName)
+	assert.Equal(t, "this is hello world", helloWorldRunDetail.Description)
```
Linchin, line 394: I think we verified more fields of the returned run in v1 ([ref](https://github.com/kubeflow/pipelines/blob/ccbac97a7bc3f3d65151fd46d7714162e15cbfc5/backend/test/integration/upgrade_test.go#L568)). We can mask the fields that vary in each test with definite values, and then make the comparison.

backend/test/v2/integration/upgrade_test.go
```diff
@@ -0,0 +1,601 @@
...
+		CreatedAt:      recurringRun.CreatedAt,
+		UpdatedAt:      recurringRun.UpdatedAt,
+		Trigger:        recurringRun.Trigger,
+		Status:         recurringRun.Status,
+	}
+
+	assert.Equal(t, expectedRecurringRun, recurringRun)
+}
+
+func (s *UpgradeTests) VerifyCreatingRunsAndJobs() {
```
Linchin, line 455: nit: function name should use recurring run

backend/test/v2/integration/upgrade_test.go
chensun: The v1 code actually doesn't verify more, if you look closely, the pattern is basically:

```golang
func check(t, actual) {
   expected = Run{
       Foo: actual.Foo,
       Bar: actual.Bar,
   }
   assert.Equal(t, expected, actual)
}
```

In a word, they are constructing the `expected` using the `actual`, then compare the two. There's actually little verification value provided with this pattern.

backend/test/v2/integration/upgrade_test.go
chensun: Good catch. Fixed.

backend/test/v2/integration/upgrade_test.go
Linchin: I think you are right, much of the fields are overwritten by the returned values. But the v1 code also verifies that some fields are not empty. Maybe we can also check if fields like pipeline spec, exp id etc. are non-empty?

chensun: pipeline spec would be empty in this case because the run was created using pipeline version id, which is a behavior change from v1. We do have API integration test that verifies when pipeline spec shouldn't be empty (for run created using pipeline spec).
For experiment id, I can certainly verify that it's not empty, but I think it's no different from verifying any other field being non-empty and we don't do this extensive non-empty check for other fields. In v1, the specialty with experiment id is it belongs to resource reference, so it makes sense to verify that is constructed correctly. In v2, we no longer have resource references.

Linchin: Thank you for explaining! This makes a lot of sense.

Linchin:
Thank you @chensun!
/lgtm
/approve

PR #9297, opened 2023-05-03
fix(conformance): enable healthz test in full Kubeflow.

james-jwu:
This change will also be ported to release-2.0-alpha branch to support conformance tests for KF 1.7, which is using alpha.7 version of KFP.

### DIFF
--- a/backend/src/common/client/api_server/v1/healthz_client.go
+++ b/backend/src/common/client/api_server/v1/healthz_client.go
@@ -17,6 +17,7 @@ package api_server
 import (
 	"fmt"
 
+	"github.com/go-openapi/runtime"
 	"github.com/go-openapi/strfmt"
 	apiclient "github.com/kubeflow/pipelines/backend/api/v1beta1/go_http_client/healthz_client"
 	params "github.com/kubeflow/pipelines/backend/api/v1beta1/go_http_client/healthz_client/healthz_service"
@@ -31,7 +32,8 @@ type HealthzInterface interface {
 }
 
 type HealthzClient struct {
-	apiClient *apiclient.Healthz
+	apiClient      *apiclient.Healthz
+	authInfoWriter runtime.ClientAuthInfoWriter
 }
 
 func NewHealthzClient(clientConfig clientcmd.ClientConfig, debug bool) (*HealthzClient, error) {
@@ -45,6 +47,20 @@ func NewHealthzClient(clientConfig clientcmd.ClientConfig, debug bool) (*Healthz
 	// Creating upload client
 	return &HealthzClient{
 		apiClient: apiClient,
+	}, nil
+}
+
+func NewKubeflowInClusterHealthzClient(namespace string, debug bool) (
+	*HealthzClient, error) {
+
+	runtime := NewKubeflowInClusterHTTPRuntime(namespace, debug)
+
+	apiClient := apiclient.New(runtime, strfmt.Default)
+
+	// Creating experiment client
+	return &HealthzClient{
+		apiClient:      apiClient,
+		authInfoWriter: SATokenVolumeProjectionAuth,
 	}, nil
 }
 
--- a/backend/src/common/client/api_server/v1/healthz_client.go
+++ b/backend/src/common/client/api_server/v1/healthz_client.go
@@ -53,14 +53,14 @@ func NewHealthzClient(clientConfig clientcmd.ClientConfig, debug bool) (*Healthz
 func NewKubeflowInClusterHealthzClient(namespace string, debug bool) (
 	*HealthzClient, error) {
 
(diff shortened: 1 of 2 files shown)


backend/test/integration/healthz_api_test.go
```diff
@@ -43,10 +44,27 @@ func (s *HealthzApiTest) SetupTest() {
...
+			return api_server.NewHealthzClient(clientConfig, *isDebugMode)
+		}
+	}
+
 	var err error
-	s.healthzClient, err = api_server.NewHealthzClient(clientConfig, false)
+	s.healthzClient, err = newHealthzClient()
 	if err != nil {
 		glog.Exitf("Failed to get healthz client. Error: %v", err)
 	}
```
gkcalat, line 70: NIT: as we don't use `newHealthzClient` function anywhere else, consider simplifying this to something more readable:
```golang
var err error
if *isKubeflowMode {
    s.resourceNamespace = *resourceNamespace
    s.healthzClient, err = api_server.NewKubeflowInClusterHealthzClient(s.namespace, *isDebugMode)
} else {
    clientConfig := test.GetClientConfig(*namespace)
    s.healthzClient, err = api_server.NewHealthzClient(clientConfig, *isDebugMode)
}
if err != nil {
    glog.Exitf("Failed to get healthz client. Error: %v", err)
}
```

backend/src/common/client/api_server/v1/healthz_client.go
```diff
@@ -17,6 +17,7 @@ package api_server
 import (
 	"fmt"
 
+	"github.com/go-openapi/runtime"
```
gkcalat, line 20: Should we add an alias to this import to avoid confusion with the variables `runtime` inside the `NewKubeflowInClusterHealthzClient` and `NewHealthzClient` functions?

backend/test/integration/healthz_api_test.go
```diff
@@ -26,8 +26,9 @@ import (
 
 type HealthzApiTest struct {
 	suite.Suite
-	namespace     string
-	healthzClient *api_server.HealthzClient
+	namespace         string
+	resourceNamespace string
```
gkcalat, line 30: I don't have the full context, so to confirm, is it being used somewhere during the test?

backend/test/integration/healthz_api_test.go
```diff
@@ -43,10 +44,27 @@ func (s *HealthzApiTest) SetupTest() {
...
 	s.namespace = *namespace
-	clientConfig := test.GetClientConfig(*namespace)
+
+	var newHealthzClient func() (*api_server.HealthzClient, error)
+
+	if *isKubeflowMode {
+		s.resourceNamespace = *resourceNamespace
+
+		newHealthzClient = func() (*api_server.HealthzClient, error) {
+			return api_server.NewKubeflowInClusterHealthzClient(s.namespace, *isDebugMode)
```
gkcalat, line 56: Just to confirm, what's the difference between `s.resourceNamespace` and `s.namespace`? Should we use the former in full Kubeflow mode?

gkcalat (commented):
Thank you @james-jwu!
I left a few questions to clarify things as I don't have full context here. Overall it looks good to me though.

backend/test/integration/healthz_api_test.go
james-jwu: s.namespace is the namespace where kfp is installed (i.e. kubeflow). s.resourceNamespace is where the KFP resources (e.g. experiments) will be created (i.e. kf-conformance).

The API client is created against s.namespace so it can talk to ml-pipeline server.

backend/test/integration/healthz_api_test.go
james-jwu: See the comment on namespace vs. resourceNamespace below. The latter isn't used in this test, but all other tests has are styled the same way so I thought it might be good for consistency sake to include resourceNamespace.

backend/src/common/client/api_server/v1/healthz_client.go
james-jwu: Sounds good. I will do it in another PR because other clients need to be updated as well.

backend/test/integration/healthz_api_test.go
james-jwu: Sounds good. Can I fix it in another PR, because all other tests are using the same style which need to be updated as well.

gkcalat:
Thanks @james-jwu 

/lgtm
/approve

backend/test/integration/healthz_api_test.go
james-jwu: Offline discussion with @gkcalat: will not fix this in order to keep the style consistent with other test file.

Issue #9350, opened 2023-05-08
[frontend] No console output when graph too large


droctothorpe:
### Environment
*  How did you deploy Kubeflow Pipelines (KFP)?
Onto an AWS-backed K8s cluster.
*  KFP version: 
1.8.1

### Steps to reproduce
Create a pipeline with more than 1000 edges or nodes.

### Expected result
There should be console indicating that graphs beyond a certain size are not supported.

### Materials and Reference
[This conditional](https://github.com/kubeflow/pipelines/blob/master/frontend/src/lib/StaticGraphParser.ts#L292-L295) prevent graphs that exceed a certain size from rendering, but there's no error message in the console indicating what the underlying problem is. This makes debugging difficult for end users and us as Kubeflow platform maintainers.

Impacted by this bug? Give it a 👍.

PR #9351, opened 2023-05-08
fix(frontend): Show error dialog for large graphs (#9350)

droctothorpe:
This PR simply adds console log messages to help identify and debug large graph problems. There's no way for a user to identify the underlying problem currently without poking around the typescript source. We maintain a KF platform and this comes up _all_ the time. Proper console output should help both end users and platform maintainers. Happy to update this to throw errors instead if that makes more sense, I just wanted to avoid modifying the function signature to be as minimally invasive as possible.

Relevant issue: https://github.com/kubeflow/pipelines/issues/9350

**Checklist:**

### DIFF
--- a/frontend/src/lib/StaticGraphParser.ts
+++ b/frontend/src/lib/StaticGraphParser.ts
@@ -288,10 +288,14 @@ export function createGraph(workflow: Workflow): dagre.graphlib.Graph {
  *
  * @param graph The dagre graph object
  */
-export function transitiveReduction(graph: dagre.graphlib.Graph): dagre.graphlib.Graph | undefined {
-  // safeguard against too big graphs
-  if (!graph || graph.edgeCount() > 1000 || graph.nodeCount() > 1000) {
-    return undefined;
+export function transitiveReduction(graph: dagre.graphlib.Graph): dagre.graphlib.Graph {
+  // Safeguard against excessively large graphs
+  if (!graph) {
+    throw new Error("Graph object doesn't exist.");
+  } else if (graph.nodeCount() > 1000) {
+    throw new Error('Graph node count exceeds 1000.');
+  } else if (graph.edgeCount() > 1000) {
+    throw new Error('Graph edge count exceeds 1000.');
   }
 
   const result = graphlib.json.read(graphlib.json.write(graph));
--- a/frontend/src/pages/PipelineDetails.tsx
+++ b/frontend/src/pages/PipelineDetails.tsx
@@ -46,7 +46,7 @@ import RunUtils from 'src/lib/RunUtils';
 import * as StaticGraphParser from 'src/lib/StaticGraphParser';
 import { compareGraphEdges, transitiveReduction } from 'src/lib/StaticGraphParser';
 import { URLParser } from 'src/lib/URLParser';
-import { logger } from 'src/lib/Utils';
+import { errorToMessage, logger } from 'src/lib/Utils';
 import { Page } from './Page';
 import PipelineDetailsV1 from './PipelineDetailsV1';
 import PipelineDetailsV2 from './PipelineDetailsV2';
@@ -666,8 +666,13 @@ class PipelineDetails extends Page<{}, PipelineDetailsState> {
         const template = JsYaml.safeLoad(templateString);
         if (WorkflowUtils.isArgoWorkflowTemplate(template)) {
           graph = StaticGraphParser.createGraph(template!);
-
-          reducedGraph = graph ? transitiveReduction(graph) : undefined;
+          let reducedGraph;
+          try {
+            reducedGraph = transitiveReduction(graph);
+          } catch (err) {
+            const errorMessage = await errorToMessage(err);
+            await this.showErrorDialog('Graph rendering failed', errorMessage);
+          }
           if (graph && reducedGraph && compareGraphEdges(graph, reducedGraph)) {
             reducedGraph = undefined; // disable reduction switch
           }
--- a/frontend/src/pages/RunDetails.test.tsx
+++ b/frontend/src/pages/RunDetails.test.tsx
@@ -140,7 +140,9 @@ describe('RunDetails', () => {
 
     testRun = {
       pipeline_runtime: {
-        workflow_manifest: '{}',
+        workflow_manifest: JSON.stringify({
+          status: { nodes: { node1: { id: 'node1' } } },
+        }),
       },
       run: {
         created_at: new Date(2018, 8, 5, 4, 3, 2),
@@ -458,6 +460,7 @@ describe('RunDetails', () => {
   });
 
   it('renders an empty run', async () => {
+    testRun.pipeline_runtime!.workflow_manifest = '{}';
     tree = shallow(<RunDetails {...generateProps()} />);
     await TestUtils.flushPromises();
     expect(tree).toMatchSnapshot();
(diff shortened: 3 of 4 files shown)


frontend/src/lib/StaticGraphParser.ts
```diff
@@ -290,9 +290,16 @@ export function createGraph(workflow: Workflow): dagre.graphlib.Graph {
  */
 export function transitiveReduction(graph: dagre.graphlib.Graph): dagre.graphlib.Graph | undefined {
   // safeguard against too big graphs
-  if (!graph || graph.edgeCount() > 1000 || graph.nodeCount() > 1000) {
+  if (!graph) {
+    console.log("Error: Graph not provided.");
```
zijianjoy, line 294: I think throwing an exception and showing an error dialog on the UI makes this error more apparent. Not everybody goes to developer console on browser if they encounter such issue. How do you think about using this function? https://github.com/kubeflow/pipelines/blob/270257512a0f217b598bf82ce6c5a283dcc458f4/frontend/src/pages/Page.tsx#L83

droctothorpe: Sounds good. My only question is as follows: given that showErrorDialog is a method of Page, and we are not subclassing Page in this file, do you have a recommended approach for accessing showErrorDialog? All other invocations of showErrorDialog take place in classes that extend Page. Would it be strange to import and instantiate Page for the sole purpose of invoking showErrorDialog?

zijianjoy: Since `RunDetail` and `PipelineDetail` pages are the two callers of this function: https://github.com/search?q=repo%3Akubeflow%2Fpipelines%20transitiveReduction&type=code, you can throw exception in `transitiveReduction()`, then catch them in `RunDetail` and `PipelineDetail` pages. Once you are in `RunDetail` and `PipelineDetail` pages, you can call `showErrorDialog`.

droctothorpe: Thanks for the guidelines, @zijianjoy! Really appreciate it.

Pushed up the changes. Happy to make any other modifications you think are necessary.

We ran into the following failures when executing the unit tests locally. It doesn't look related to me, but I'm not sure.
```
Summary of all failing tests
 FAIL  src/pages/RunDetails.test.tsx (33.131 s)
  ● RunDetails › calls retry API when retry dialog is confirmed

    TypeError: Cannot read property 'onClick' of undefined

      297 |     const call = updateDialogSpy.mock.calls[0][0];
      298 |     const confirmBtn = call.buttons.find((b: any) => b.text === 'Retry');
    > 299 |     await confirmBtn.onClick();
          |                      ^
      300 |     expect(retryRunSpy).toHaveBeenCalledTimes(1);
      301 |     expect(retryRunSpy).toHaveBeenLastCalledWith(testRun.run!.id);
      302 |   });

      at Object.onClick (src/pages/RunDetails.test.tsx:299:22)

  ● RunDetails › shows an error dialog when retry API fails

    TypeError: Cannot read property 'onClick' of undefined

      325 |     const call = updateDialogSpy.mock.calls[0][0];
      326 |     const confirmBtn = call.buttons.find((b: any) => b.text === 'Retry');
    > 327 |     await confirmBtn.onClick();
          |                      ^
      328 |     expect(updateDialogSpy).toHaveBeenLastCalledWith(
      329 |       expect.objectContaining({
      330 |         content: 'Failed to retry run: test-run-id with error: ""mocked error""',

      at Object.onClick (src/pages/RunDetails.test.tsx:327:22)

  ● RunDetails › calls terminate API when terminate dialog is confirmed

    TypeError: Cannot read property 'onClick' of undefined

      377 |     const call = updateDialogSpy.mock.calls[0][0];
      378 |     const confirmBtn = call.buttons.find((b: any) => b.text === 'Terminate');
    > 379 |     await confirmBtn.onClick();
```

zijianjoy: It is likely due to the mock test run is empty in this line: https://github.com/kubeflow/pipelines/blob/a2d1f89aa1274b6b57770151043bbf6794f74a84/frontend/src/pages/RunDetails.test.tsx#L143. You will need to fulfill it with some valid mock data, otherwise the UI will show the error dialog, which is expected because that is what your PR is trying to achieve.

droctothorpe: Copy that. Thanks, @zijianjoy! I've been slammed with other stuff but will try to resolve tomorrow. Thanks for your patience!

frontend/src/lib/StaticGraphParser.ts
```diff
@@ -288,10 +288,14 @@ export function createGraph(workflow: Workflow): dagre.graphlib.Graph {
...
-    return undefined;
+export function transitiveReduction(graph: dagre.graphlib.Graph): dagre.graphlib.Graph {
+  // Safeguard against excessively large graphs
+  if (!graph) {
+    throw new Error("Graph object doesn't exist.");
+  } else if (graph.nodeCount() > 1000) {
+    throw new Error('Graph node count exceeds 1000.');
+  } else if (graph.edgeCount() > 1000) {
+    throw new Error('Graph edge count exceeds 1000.');
   }
```
zijianjoy, line 299: Have you tested these two scenarios locally?

droctothorpe: Not yet. Can do though. Hopefully it doesn't clobber my local K8s, heh.

droctothorpe: Hey, @zijianjoy. I tested this locally. It seems to be working as expected. The only problem is that trying to spawn a pipeline with 1000 components crashed my local k8s cluster. I temporarily reduced the condition from 1000 to 4 for validation purposes. When I submit the following pipeline:

```python
def ran() -> None:
    from random import random
    print(random(1, 1000))

ran_component = kfp.components.create_component_from_func(ran)

@kfp.dsl.pipeline(name="thousand")
def large_pipeline():
    step = ran_component()
    step = ran_component()
    step = ran_component()
    step = ran_component()
```

I get the following result in the GUI:

zijianjoy: Thank you for the validation. I was under the impression that frontend shouldn't load the large graph that exceeded threshold. But looks like it is still loading the large graph with your change. Can we avoid loading the large graph?

frontend/src/pages/PipelineDetails.tsx
```diff
@@ -666,8 +666,13 @@ class PipelineDetails extends Page<{}, PipelineDetailsState> {
         const template = JsYaml.safeLoad(templateString);
         if (WorkflowUtils.isArgoWorkflowTemplate(template)) {
           graph = StaticGraphParser.createGraph(template!);
```
zijianjoy, line 668: If I understand it correctly, you would probably like to implement the validation for `graph` here instead of `reducedGraph`, because it is the default rendering approach. and the `transitiveReduction()` is like a post-render action to check whether the graph has been too large.

droctothorpe:
Sincere apologies on the delay here. I have priority whiplash from shifting focus to so many other things. I can make this more of a priority.

sachdevayash1910:
frontend docker build for this fails:
```shell
ub.com/request/request/issues/3142
[2/4] Fetching packages...
error workbox-webpack-plugin@6.6.1: The engine "node" is incompatible with this module. Expected version ">=16.0.0". Got "14.21.3"
error Found incompatible module
``` 
Is this something which has a quick fix? Changing the base image version to 16 did not help either
```shell
npm ERR! code E403
npm ERR! 403 403 Forbidden - GET https://artifactory.twitter.biz:443/api/npm/js-virtual/html-tags/-/html-tags-3.1.0.tgz
npm ERR! 403 In most cases, you or one of your dependencies are requesting
npm ERR! 403 a package version that is forbidden by your security policy, or
npm ERR! 403 on a server you do not have access to.
```

PR #9373, opened 2023-05-10
fix(backend): Fix authorization logic and remove redundant DB calls

gkcalat:

- Fixed inconsistencies in authorization requests (`Verb`, `Group`, `Version`, `Name`, `Namespace`, `Resource`) across all servers
- Enforced the same pattern for checking access to minimize future bugs across all servers
- Added better support for Shared pipelines in multi-user mode
- Fixed bugs and erroneous test cases for multi-user behavior

- Other improvements:
   - Removed redundant calls to DB
   - Improved error messages and types

- Tested (WIP)
   - [x] Kubeflow 1.7.0 on GKE
   - [x] Standalone KFP @ master on GKE

/hold until #9391 gets merged

**Checklist:**


backend/src/apiserver/resource/resource_manager.go
```diff
@@ -1792,12 +1795,3 @@ func (r *ResourceManager) GetTask(taskId string) (*model.Task, error) {
 	}
 	return task, nil
 }
-
-// Fetches run metric entries for a given run id.
-func (r *ResourceManager) GetRunMetrics(runId string) ([]*model.RunMetric, error) {
-	metrics, err := r.runStore.GetMetrics(runId)
```
Linchin, line 1798: If we are removing the `GetRunMetrics()` method, looks like this is also the only use case for `GetMetrics()` in run storage. Shall we remove this method too?

backend/src/apiserver/server/experiment_server.go
```diff
@@ -328,26 +328,23 @@ func (s *ExperimentServer) canAccessExperiment(ctx context.Context, experimentID
...
-			resourceAttributes.Namespace = experiment.Namespace
-		}
+		namespace = experiment.Namespace
 		if len(resourceAttributes.Name) == 0 {
 			resourceAttributes.Name = experiment.Name
 		}
 	}
-
+	if s.resourceManager.IsEmptyNamespace(resourceAttributes.Namespace) {
+		if len(namespace) == 0 {
```
Linchin, line 343: Since `namespace` is a string, maybe we can just compare it with an empty string instead?

backend/src/apiserver/server/experiment_server.go
```diff
@@ -328,26 +328,23 @@ func (s *ExperimentServer) canAccessExperiment(ctx context.Context, experimentID
 		// Skip authorization if not multi-user mode.
 		return nil
 	}
-
+	namespace := ""
```
Linchin, line 331: I see that the value of `resourceAttributes.Namespace` takes the order of 
1. `resourceAttributes.Namespace` itself if non-empty
2. `experiment.Namespace` if this is non-empty
3. error if both `resourceAttributes.Namespace` and `experiment.Namespace` are empty

I wonder if we should add a verification that `resourceAttributes.Namespace == experiment.Namespace` when they are both provided.

backend/src/apiserver/server/job_server.go
```diff
@@ -90,22 +89,19 @@ func (s *JobServer) createJob(ctx context.Context, job *model.Job) (*model.Job,
 	if err != nil {
 		return nil, util.Wrapf(err, "Failed to create a run due to invalid experimentId and namespace combination")
```
Linchin, line 90: nit: recurring run

backend/src/apiserver/server/job_server.go
```diff
@@ -403,33 +399,38 @@ func (s *JobServer) canAccessJob(ctx context.Context, jobID string, resourceAttr
 		// Skip authorization if not multi-user mode.
```
Linchin, line 399: Shall we also verify that `namespace` is empty if we are not in multi-user mode?

backend/src/apiserver/server/pipeline_server.go
```diff
@@ -113,17 +113,14 @@ func (s *PipelineServer) createPipeline(ctx context.Context, pipeline *model.Pip
 	}
 	pipeline.Namespace = s.resourceManager.ReplaceNamespace(pipeline.Namespace)
```
Linchin, line 114: Similar to above, here we just replaces the namespace with empty string if we are not in multi-user mode. Should we report an error instead?

backend/src/apiserver/server/pipeline_server.go
```diff
@@ -985,66 +972,56 @@ func (s *PipelineServer) GetPipelineVersionTemplate(ctx context.Context, request
...
-	if err != nil {
-		return util.Wrapf(err, "Failed to access pipeline version %s. Check if you have access to namespace %s", versionId, resourceAttributes.Namespace)
+		pipelineId = pipelineVersion.PipelineId
 	}
-	return nil
+	return s.canAccessPipeline(ctx, pipelineId, resourceAttributes)
 }
 
 // Checks if a user can access a pipeline.
+// Adds parent namespace of pipeline id is not empty,
```
Linchin, line 994: nit: typo

backend/src/apiserver/server/run_server.go
```diff
@@ -366,28 +339,16 @@ func (s *RunServer) DeleteRunV1(ctx context.Context, request *apiv1beta1.DeleteR
...
-		}
-		resourceAttributes := &authorizationv1.ResourceAttributes{
-			Namespace: existingRun.Namespace,
-			Verb:      common.RbacResourceVerbList,
-		}
-		if err := s.canAccessRun(ctx, "", resourceAttributes); err != nil {
-			return nil, util.Wrapf(err, "Failed to report run metrics due to authorization error. Check if you have permission to access namespace %s", existingRun.Namespace)
+	// Verify that the run exists for single user mode.
+	// Multi-user model will verify this when checking authorization above.
+	if !common.IsMultiUserMode() {
```
Linchin, line 349: Here in `reportRunMetrics()` we verify that the run exists in single-user mode. Should we do the same for delete/archive/unarchive?

backend/src/apiserver/server/job_server.go
```diff
@@ -90,22 +89,19 @@ func (s *JobServer) createJob(ctx context.Context, job *model.Job) (*model.Job,
 	if err != nil {
 		return nil, util.Wrapf(err, "Failed to create a run due to invalid experimentId and namespace combination")
 	}
+	if common.IsMultiUserMode() && namespace == "" {
```
Linchin, line 92: Shall we also verify that `namespace == ""` in single-user mode?

Linchin:
Hi @gkcalat, thank you and good job! I left some comments to clarify the access verification logic.

backend/src/apiserver/server/experiment_server.go
```diff
@@ -328,26 +328,23 @@ func (s *ExperimentServer) canAccessExperiment(ctx context.Context, experimentID
...
-		}
+		namespace = experiment.Namespace
 		if len(resourceAttributes.Name) == 0 {
 			resourceAttributes.Name = experiment.Name
 		}
 	}
-
+	if s.resourceManager.IsEmptyNamespace(resourceAttributes.Namespace) {
+		if len(namespace) == 0 {
+			return util.NewInvalidInputError("An experiment cannot have an empty namespace in multi-user mode")
```
chensun, line 344: Does empty namespace means multi-user mode? It's not super clear to me that this code path is only hit with multi-user mode.

backend/src/apiserver/server/experiment_server_test.go
```diff
@@ -341,21 +341,21 @@ func TestCreateExperimentV1_Multiuser(t *testing.T) {
...
-			"Invalid - wrong owner",
+			"Invalid - wrong owner type",
 			DefaultFakeIdEight,
 			&apiv1beta1.Experiment{
 				Name:        "exp6",
 				Description: "first experiment",
 				ResourceReferences: []*apiv1beta1.ResourceReference{
 					{
-						Key:          &apiv1beta1.ResourceKey{Type: apiv1beta1.ResourceType_EXPERIMENT, Id: "exp2"},
+						Key:          &apiv1beta1.ResourceKey{Type: apiv1beta1.ResourceType_EXPERIMENT, Id: "ns1"},
```
chensun, line 351: This diff looks fishy and possibly an indication of bugs somewhere. `ns1` is we used for namespace value, while this resource is of type experiment.

backend/src/apiserver/server/job_server.go
```diff
@@ -403,33 +399,38 @@ func (s *JobServer) canAccessJob(ctx context.Context, jobID string, resourceAttr
...
-		if len(resourceAttributes.Namespace) == 0 {
-			if len(job.Namespace) == 0 {
-				return util.NewInternalServerError(
-					errors.New("Empty namespace"),
-					"The job doesn't have a valid namespace",
-				)
+		if job.Namespace == "" {
+			experiment, err := s.resourceManager.GetExperiment(job.ExperimentId)
+			if err != nil {
+				return util.NewInvalidInputError("recurring run %v has an empty namespace and the parent experiment %v could not be fetched: %s", jobID, job.ExperimentId, err.Error())
```
chensun, line 411: Why would a recurring run has no parent experiment? I think that would be a case of internal server error.

backend/src/apiserver/server/pipeline_server.go
```diff
@@ -985,66 +972,56 @@ func (s *PipelineServer) GetPipelineVersionTemplate(ctx context.Context, request
...
+		pipeline, err := s.resourceManager.GetPipeline(pipelineId)
 		if err != nil {
 			return util.Wrapf(err, "Failed to access pipeline %s. Check if it exists and have a namespace assigned", pipelineId)
 		}
-		// Allow users to access pipelines in the default namespace.
-		if s.resourceManager.IsEmptyNamespace(namespace) {
-			return nil
+		resourceAttributes.Namespace = pipeline.Namespace
+		if len(resourceAttributes.Name) == 0 {
+			resourceAttributes.Name = pipeline.Name
```
chensun, line 1008: Why this change? Is `len(resourceAttributes.Name) == 0` a legit case, and expected to be patched here?

backend/src/apiserver/server/job_server.go
```diff
@@ -90,22 +89,19 @@ func (s *JobServer) createJob(ctx context.Context, job *model.Job) (*model.Job,
...
 	job.Namespace = namespace
-	// Validate authorization
-	if common.IsMultiUserMode() {
-		if job.Namespace == "" {
-			return nil, util.NewInternalServerError(util.NewInvalidInputError("Recurring run cannot have an empty namespace in multi-user mode"), "Failed to create a recurring run due to empty namespace")
-		}
-		resourceAttributes := &authorizationv1.ResourceAttributes{
-			Namespace: job.Namespace,
-			Verb:      common.RbacResourceVerbCreate,
-			Resource:  common.RbacResourceTypeJobs,
```
chensun, line 103: Is it intentional that `Resource` is deleted from the new code?

backend/src/apiserver/server/pipeline_server.go
```diff
@@ -113,17 +113,14 @@ func (s *PipelineServer) createPipeline(ctx context.Context, pipeline *model.Pip
 	}
 	pipeline.Namespace = s.resourceManager.ReplaceNamespace(pipeline.Namespace)
 	// Check authorization
-	if common.IsMultiUserMode() {
-		resourceAttributes := &authorizationv1.ResourceAttributes{
-			Namespace: pipeline.Namespace,
-			Verb:      common.RbacResourceVerbCreate,
-			Group:     common.RbacPipelinesGroup,
-			Resource:  common.RbacResourceTypePipelines,
```
chensun, line 121: ditto on missing `Resource` in the new code.

backend/src/apiserver/server/pipeline_server.go
```diff
@@ -377,14 +369,12 @@ func (s *PipelineServer) GetPipelineByName(ctx context.Context, request *apiv2be
...
-			Namespace: namespace,
-			Verb:      common.RbacResourceVerbGet,
-		}
-		if err := s.checkAccess(ctx, resourceAttributes); err != nil {
-			return nil, nil, 0, "", util.Wrapf(err, "Failed to list pipelines due to authorization error. Check if you have read permission to namespace %v", namespace)
-		}
+	resourceAttributes := &authorizationv1.ResourceAttributes{
+		Namespace: namespace,
+		Verb:      common.RbacResourceVerbList,
+	}
```
chensun, line 375: It looks very random that resourceAttributes may have 2 to 4 attributes set and it varies in different places. Don't we need `Resource` and `Name` here and in all other places?

backend/src/apiserver/server/run_server.go
```diff
@@ -118,23 +117,18 @@ func (s *RunServer) createRun(ctx context.Context, run *model.Run) (*model.Run,
...
-			Resource:  common.RbacResourceTypeRuns,
-			Name:      run.DisplayName,
-		}
-		if err := s.resourceManager.IsAuthorized(ctx, resourceAttributes); err != nil {
-			return nil, util.Wrapf(err, "Failed to create a run due to authorization error. Check if you have write permissions to namespace %s", run.Namespace)
-		}
+	resourceAttributes := &authorizationv1.ResourceAttributes{
+		Namespace: run.Namespace,
+		Verb:      common.RbacResourceVerbCreate,
+	}
```
chensun, line 129: Again, here's a case that the old code sets 5 attributes in `ResourceAttributes`, while the new code sets only 2. Seems not right?

chensun (commented):
The title of this PR should be more explicit about the fix, this is definitely not only a refactoring.

chensun (commented):
The title of this PR should be more explicit about the fix, this is definitely not only a refactoring.

backend/src/apiserver/common/const.go
```diff
@@ -17,7 +17,7 @@ package common
 const (
 	RbacKubeflowGroup    = "kubeflow.org"
 	RbacPipelinesGroup   = "pipelines.kubeflow.org"
-	RbacPipelinesVersion = "v1beta1"
+	RbacPipelinesVersion = "v2beta1"
```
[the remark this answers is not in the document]
gkcalat: This is a version of k8s resource. We currently don't have version-level access control for users, so it doesn't break anything.
However, I agree that it should be a separate PR and may require additional discussion.

backend/src/apiserver/server/experiment_server.go
```diff
@@ -285,7 +285,7 @@ func (s *ExperimentServer) ListExperiments(ctx context.Context, request *apiv2be
 }
 
 func (s *ExperimentServer) deleteExperiment(ctx context.Context, experimentId string) error {
-	err := s.canAccessExperiment(ctx, experimentId, &authorizationv1.ResourceAttributes{Verb: common.RbacResourceVerbArchive})
+	err := s.canAccessExperiment(ctx, experimentId, &authorizationv1.ResourceAttributes{Verb: common.RbacResourceVerbDelete})
```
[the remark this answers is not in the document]
gkcalat: Yes. We can potentially add more fin-grained control to [FakeSubjectAccessReviewClient](https://github.com/gkcalat/pipelines/blob/master/backend/src/apiserver/client/subject_access_review_fake.go#L25-L34). This will require additional changes to all multi-user unit tests.

Another option is to develop end-to-end test for Full Kubeflow with fine-grained RBAC.

backend/src/apiserver/resource/resource_manager.go
gkcalat: Created a separate PR and added @Linchin's suggestion there: https://github.com/kubeflow/pipelines/pull/9391

backend/src/apiserver/server/job_server.go
gkcalat: In a single-user mode, we simply skip authorization. So no check are needed.

backend/src/apiserver/server/run_server.go
gkcalat: As the comment suggests, the `canAccessRun()` verifies that for multi-user mode. So we are removing a redundant call to the DB here.

backend/src/apiserver/server/experiment_server.go
gkcalat: I agree that both are okay and in fact they get compiled to the same result. I do support @Linchin's comment as it is probably more explicit.

backend/src/apiserver/server/experiment_server.go
gkcalat: `canAccessExperiment()` [returns immediately](https://github.com/kubeflow/pipelines/blob/master/backend/src/apiserver/server/experiment_server.go#L326-L330) if it's a single-user mode.

backend/src/apiserver/server/experiment_server_test.go
gkcalat: It was intended to test a wrong resource type (`ResourceType_EXPERIMENT` instead of `ResourceType_NAMESPACE`), hence I changed it to `ns1`. Let me revert this change.

backend/src/apiserver/server/job_server.go
gkcalat: Fixed

backend/src/apiserver/server/job_server.go
gkcalat: Yes. The resource, group, and version are added in `canAccessJob`.

backend/src/apiserver/server/job_server.go
gkcalat: IMO, we should not. I see a standalone deployment of KFP as a mode where a user has access to everything. So specifying a `Namespace` is still a valid request, but we ignore it.

backend/src/apiserver/server/job_server.go
gkcalat: Both `namespace` and `experimentId` are coming from the user, so `NewInvalidInputError` makes sense when someone specifies a non-existing id, for example. But, that shouldn't happen. Let me change it to the server error.

backend/src/apiserver/server/pipeline_server.go
gkcalat: I might be wrong on this, but see my opinion [here](https://github.com/kubeflow/pipelines/pull/9373/files#r1191532729)

backend/src/apiserver/server/pipeline_server.go
gkcalat: See [here](https://github.com/kubeflow/pipelines/pull/9373/files#r1191530425)

backend/src/apiserver/server/pipeline_server.go
gkcalat: The way we check authorization actually became more systemic with this PR. We follow the same pattern everywhere and reuse a lot of code. Depending on what we are trying to do (`Verb`) we may or may not specify `Name` and `Namespace`.

In this particular case, it's because we are trying to list all pipelines in a given namespace. Hence no `Name`. The rest is being added in the `canAccessPipeline`.

backend/src/apiserver/server/pipeline_server.go
gkcalat: If the `pipelineId` is not empty, the request targets a specific resource. Hence we add the name if it's missing. This is to avoid multiple calls to `GetRun`. Is some cases, like `GetPipelineByNameV1` we already have a name.

backend/src/apiserver/server/run_server.go
gkcalat: See [here](https://github.com/kubeflow/pipelines/pull/9373/files#r1191530425)

backend/src/apiserver/server/experiment_server.go
gkcalat: Thank you. I agree. This shouldn't happen. Whenever an `experimentId` is present, the `resourceAttributes.Namespace` should be come from the experiment. Let me implement it.

backend/src/apiserver/server/job_server.go
chensun: I see, so probably a cleaner way is to always construct the full `ResourceAttributes` object, and reduce all the `canAccessXXX` functions into `canAccessResource`, WDTY? This can be out-of-scope.

chensun (commented):
/lgtm

Thanks!

PR #9374, opened 2023-05-10
fix(backend): Add pipeline spec name validation and relax display name validation

gkcalat:

- Added validation of the pipeline name in `spec.pipeline-info.name` that matches the MLMD requirements for the following calls:
  - Upload pipeline (v1)
  - Create pipeline (v1)
  - Upload pipeline version
  - Create pipeline version
- Added a few unit tests related to name validation
- Relaxes requirements for pipeline's `DisplayName`
- Removed irrelevant unit tests
- Removed logic for overwriting `spec.pipeline-info.name`

/CC @jlyaoyuli 

**Checklist:**

### DIFF
--- a/backend/src/apiserver/common/utils.go
+++ b/backend/src/apiserver/common/utils.go
@@ -16,7 +16,18 @@ package common
 
 import (
 	"fmt"
+	"regexp"
 	"strings"
+
+	"github.com/kubeflow/pipelines/backend/src/common/util"
+)
+
+const (
+	// This regex expresses the following constraints:
+	// * Allows lowercase letters and numbers at the beginning
+	// * Allows lowercase letters, numbers, and "-" in everywhere else
+	// * Sets max length to be 128 characters
+	pipelineNamePattern = "^[a-z0-9][a-z0-9-]{0,127}$"
 )
 
 // CreateArtifactPath creates artifact resource path.
@@ -88,4 +99,21 @@ func PatchPipelineDefaultParameter(text string) (string, error) {
 		text = strings.Replace(text, key, value, -1)
 	}
 	return text, nil
+}
+
+// Validates a pipeline name to match MLMD requirements.
+func ValidatePipelineName(pipelineName string) error {
+	if pipelineName == "" {
+		return util.NewInvalidInputError("pipeline's name cannot be empty")
+	}
+	if len(pipelineName) > 128 {
+		return util.NewInvalidInputError("pipeline's name must contain no more than 128 characters")
+	}
+	if matched, err := regexp.MatchString(pipelineNamePattern, pipelineName); err != nil {
+		return util.NewInternalServerError(
+			err, "failed to compile pattern '%s'", pipelineNamePattern)
+	} else if !matched {
+		return util.NewInvalidInputError("pipeline's name must contain only lowercase alphanumeric characters or '-' and must start with alphanumeric characters")
+	}
+	return nil
 }
--- a/backend/src/apiserver/resource/resource_manager.go
+++ b/backend/src/apiserver/resource/resource_manager.go
@@ -308,6 +308,9 @@ func (r *ResourceManager) UpdatePipelineDefaultVersion(pipelineId string, versio
 // Creates a pipeline, but does not create a pipeline version.
 // Call CreatePipelineVersion to create a pipeline version.
 func (r *ResourceManager) CreatePipeline(p *model.Pipeline) (*model.Pipeline, error) {
+	if p.Name == "" {
+		return nil, util.NewInvalidInputError("pipeline's name cannot be empty")
+	}
 	// Create a record in KFP DB (only pipelines table)
 	newPipeline, err := r.pipelineStore.CreatePipeline(p)
 	if err != nil {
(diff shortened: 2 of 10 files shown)


backend/src/apiserver/server/pipeline_server.go
```diff
@@ -592,6 +601,23 @@ func (s *PipelineServer) validatePipelineVersionBeforeCreating(p *model.Pipeline
...
+// Validates a pipeline name to match the requirements. This function matches the constraints in the frontend.
+func validatePipelineName(pipelineName string) error {
+	if pipelineName == "" {
+		return util.NewInvalidInputError("pipeline's name cannot be empty")
+	}
+	if len(pipelineName) > 100 {
+		return util.NewInvalidInputError("pipeline's name must contain no more than 100 characters")
+	}
+	if matched, err := regexp.MatchString(pipelineNamePattern, pipelineName); err != nil {
+		return util.NewInternalServerError(
```
chensun, line 613: This should never happen, our unit test can already proof it has no problem compiling the regex.

backend/src/apiserver/server/pipeline_server.go
```diff
@@ -592,6 +601,23 @@ func (s *PipelineServer) validatePipelineVersionBeforeCreating(p *model.Pipeline
 	return util.NewInvalidInputError("Pipeline version must have a pipeline spec or a valid source code's URL. PipelineSpec: %s. PipelineSpecURI: %s. CodeSourceUrl: %s. At least one of them must have a valid pipeline spec", p.PipelineSpec, p.PipelineSpecURI, p.CodeSourceUrl)
 }
 
+// Validates a pipeline name to match the requirements. This function matches the constraints in the frontend.
+func validatePipelineName(pipelineName string) error {
+	if pipelineName == "" {
+		return util.NewInvalidInputError("pipeline's name cannot be empty")
+	}
+	if len(pipelineName) > 100 {
```
chensun, line 609: As discussed, this check should be removed.

chensun (commented):
Is this PR still needed? I think we landed on removing the pipeline name validation at pipeline creation time and no override to pipeline_spec.pipeline_info.name.

backend/src/apiserver/server/pipeline_server.go
gkcalat: Updated the PR.

backend/src/apiserver/server/pipeline_server.go
gkcalat: True, this error is very unlikely. I still believe having this error message gives more options for troubleshooting issues caused by factors that out of our control like future changes or bugs in `regexp` package or sporadic runtime issues.

gkcalat:
> Is this PR still needed? I think we landed on removing the pipeline name validation at pipeline creation time and no override to pipeline_spec.pipeline_info.name.

Yes, we still need it. I updated the PR description too.

backend/src/apiserver/server/pipeline_server.go
chensun: > future changes or bugs in `regexp` package 

This if happens would be captured by our presubmit.

> or sporadic runtime issues.

For the natural of what it does, it would be very unlikely to have flakiness

I still prefer cleaner code to the little chance of this code ever providing any value, but no strong opinion.

backend/src/apiserver/resource/resource_manager.go
```diff
@@ -340,9 +343,29 @@ func (r *ResourceManager) CreatePipelineAndPipelineVersion(p *model.Pipeline, pv
...
+	// 1. pipeline spec for v2 pipelines and v2-compatible pipeline must comply with MLMD requirements
+	// 2. display name must be non-empty
+	pipelineSpecName := ""
+	if tmpl.IsV2() {
+		pipelineSpecName = tmpl.V2PipelineName()
+		if err := common.ValidatePipelineName(pipelineSpecName); err != nil {
+			return nil, nil, err
+		}
+	}
+	if pv.Name == "" && p.Name == "" {
```
chensun, line 357: Would this ever happen? not required input for v1 API?
Anyway, this contradicts to the comment above: "2. display name must be non-empty"

chensun (commented):
/lgtm

backend/src/apiserver/resource/resource_manager.go
gkcalat: I am probably missing something here. What we are doing here:
1. Validate `spec.pipeline-info.name`
2. If `DisplayName` is empty, replace it with `spec.pipeline-info.name`
3. If both are empty, throw an error.

`DisplayName` can be empty. We can't rely on the validation in UI or SDK. Hence we have the logic to validate it here.

backend/src/apiserver/common/utils.go
```diff
@@ -89,3 +100,20 @@ func PatchPipelineDefaultParameter(text string) (string, error) {
 	}
 	return text, nil
 }
+
+// Validates a pipeline name to match MLMD requirements.
+func ValidatePipelineName(pipelineName string) error {
+	if pipelineName == "" {
+		return util.NewInvalidInputError("pipeline's name cannot be empty")
+	}
+	if len(pipelineName) > 128 {
```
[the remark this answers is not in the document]
gkcalat: One of the goals I was chasing was to get as close as possible to the error messages that are [shown to the users in UI right now](https://github.com/kubeflow/pipelines/blob/master/frontend/src/pages/NewPipelineVersion.tsx#L660-L674).

PR #9380, opened 2023-05-10
fix(frontend): Remove redundant "(optional)" text in the optional input fields.

jlyaoyuli:
We already use asterisk to indicate required field, so `(optional)` can be removed.

Before:
Screenshot 2023-05-10 at 1 45 10 PM

After:
Screenshot 2023-05-10 at 1 44 24 PM


chensun (approved):
/lgtm
/approve

PR #9383, opened 2023-05-10
fix(frontend): Rename button in pipeline/run details side panel

jlyaoyuli:
Currently, we use `Open workflow` on the button for expanding Sub-DAG. The term `Workflow` is a little confusing, so we change to `Open Sub-DAG`.

Before:
Screenshot 2023-05-10 at 2 56 30 PM

After:
Screenshot 2023-05-10 at 2 56 58 PM


chensun (approved):
/lgtm
/approve

PR #9397, opened 2023-05-11
fix(backend): prepublish execution in the driver for create/delete pvc

Linchin:
During the beta.2 testing, we found that we could not retrieve logs for createpvc / deletepvc pods. This is because in these tasks we omitted the launcher, which prepublishes and publishes the execution to the MLMD. Currently, the driver only publishes but does not prepublish, and prepublishing is crucial in retrieving the log because it provides the pod name and pod id. 

This PR 
* Adds the prepublishing process for createpvc / deletepvc tasks
* Combines all the actions for createpvc in driver into one function, and similarly for deletepvc
* Ensures that prepublishing and publishing happen even if createpvc / deletepvc fails (except when k8s client fails to initiate)
* Updates the driver and launcher image hash

Note:
Cannot write unit tests due to lack of fake client for MLMD.

Tests:
Tested manually that the driver logs are displayed for createpvc/deletepvc.

**Checklist:**


backend/src/v2/compiler/argocompiler/testdata/hello_world.yaml
```diff
@@ -50,7 +50,7 @@ spec:
       - '{{inputs.parameters.kubernetes-config}}'
       command:
       - driver
-      image: gcr.io/ml-pipeline/kfp-driver@sha256:60d6d090937d7de393b3740520f4a08fcf85c3419bbdb10bd266cd446d8fe44a
+      image: gcr.io/ml-pipeline-test/kfp-driver@sha256:577221911abd8e6be65d7afad1bdb8d5e4544a11822489f9f5edb72d6e7f87de
```
chensun, line 53: When this PR is ready for merge, can you build `gcr.io/ml-pipeline/kfp-driver` and `gcr.io/ml-pipeline/kfp-launcher` using these Dockerfile instead?
- https://github.com/kubeflow/pipelines/blob/master/backend/Dockerfile.driver
- https://github.com/kubeflow/pipelines/blob/master/backend/Dockerfile.launcher

backend/src/v2/driver/driver.go
```diff
@@ -1118,6 +1082,136 @@ var accessModeMap = map[string]k8score.PersistentVolumeAccessMode{
...
+		"name": structpb.NewStringValue(pvcName),
+	}
+	status = pb.Execution_COMPLETE
+	return nil
+}
+
+// deletePVCandPublish() deletes the PVC and publishes the execution to MLMD, regardless of the
+// creation succeeds or not. However, if we failed to initiate the k8s client, we cannot
+// publish the execution.
+func deletePVCandPublish(
```
chensun, line 1132: The two methods create and delete are mostly duplicate, can them be reduced into one?

chensun (commented):
@Linchin did you test this manually? Would it show the driver log for createpvc/deletepvc?

Linchin:
@chensun, yes, I tested manually.

backend/src/v2/driver/driver.go
Linchin: I think you have a good point here. If we add more driver only functions in the future, it would be cumbersome to have one function for each. I will change the code accordingly.

chensun (commented):
/lgtm

Thanks!

(pending image rebuild and update)

backend/src/v2/compiler/argocompiler/testdata/hello_world.yaml
Linchin: Thank you, I updated the driver/launcher images.

chensun:
Tests passed in previous run and was retriggered due to a FE PR merged. 
Skipping the test and merge as is.

PR #9414, opened 2023-05-12
feat(backend): disable cache when volumes are involved

Linchin:
Disables cache when volumes are involved in driver, including the following operations:
* Create PVC
* Delete PVC
* Mount PVC

Tested locally that cache is disabled. 

**Note:** Need to update driver image hash when the PR is approved.

**Checklist:**

### DIFF
--- a/backend/src/v2/driver/driver.go
+++ b/backend/src/v2/driver/driver.go
@@ -1377,7 +1377,7 @@ func isVolumeOp(image string, k8sExecCfg *kubernetesplatform.KubernetesExecutorC
 	if image == "argostub/createpvc" || image == "argostub/deletepvc" {
 		return true
 	}
-	if len(k8sExecCfg.GetSecretAsVolume()) > 0 || len(k8sExecCfg.GetPvcMount()) > 0 {
+	if len(k8sExecCfg.GetPvcMount()) > 0 {
 		return true
 	}
 	return false
--- a/backend/src/v2/driver/driver.go
+++ b/backend/src/v2/driver/driver.go
@@ -261,6 +261,11 @@ func Container(ctx context.Context, opts Options, mlmd *metadata.Client, cacheCl
 		executorInput.Outputs = provisionOutputs(pipeline.GetPipelineRoot(), opts.Task.GetTaskInfo().GetName(), opts.Component.GetOutputDefinitions())
 	}
 
+	isVolumeOp := isVolumeOp(opts.Container.Image, opts.KubernetesExecutorConfig)
+	if isVolumeOp {
+		glog.Infof("Task {%s} disables cache due to it's volume related operation", opts.Task.GetTaskInfo().GetName())
+	}
+
 	ecfg, err := metadata.GenerateExecutionConfig(executorInput)
 	if err != nil {
 		return execution, err
@@ -271,7 +276,7 @@ func Container(ctx context.Context, opts Options, mlmd *metadata.Client, cacheCl
 	ecfg.IterationIndex = iterationIndex
 	ecfg.NotTriggered = !execution.WillTrigger()
 
-	if execution.WillTrigger() && opts.Task.GetCachingOptions().GetEnableCache() {
+	if execution.WillTrigger() && opts.Task.GetCachingOptions().GetEnableCache() && !isVolumeOp {
 		glog.Infof("Task {%s} enables cache", opts.Task.GetTaskInfo().GetName())
 		fingerPrint, err := getFingerPrint(opts, executorInput)
 		if err != nil {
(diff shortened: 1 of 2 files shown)


backend/src/v2/driver/driver.go
```diff
@@ -271,7 +276,7 @@ func Container(ctx context.Context, opts Options, mlmd *metadata.Client, cacheCl
 	ecfg.IterationIndex = iterationIndex
 	ecfg.NotTriggered = !execution.WillTrigger()
 
-	if execution.WillTrigger() && opts.Task.GetCachingOptions().GetEnableCache() {
+	if execution.WillTrigger() && opts.Task.GetCachingOptions().GetEnableCache() && !isVolumeOp {
```
chensun, line 279: This creates a discrepancy between what's instructed by IR and the actual behavior. 
I wonder if we should implement this disable caching logic through SDK instead. @connor-mccarthy, thoughts?

connor-mccarthy: Just to be sure I understand, would the alternative in the SDK be:
(a) to instruct users to not use cache? (action item: add documentation)
(b) have the [`kubernetes.mount_pvc` method](https://kfp-kubernetes.readthedocs.io/en/kfp-kubernetes-0.0.1/source/kubernetes.html#kfp.kubernetes.mount_pvc) call [`PipelineTask.set_caching_options(False)`](https://kubeflow-pipelines.readthedocs.io/en/latest/source/dsl.html#kfp.dsl.PipelineTask.set_caching_options) under the hood when it mounts the volume? (action item: implement this logic in `kfp-kubernetes`)

Is there another option I'm not thinking of?

chensun: I was thinking option b, for `CreatePVC`, `DeletePVC`, and maybe for `mount_pvc` as well. I wonder if we can disable caching by default, but if users intend to use cache, they can still do so. e.g.:
```python
# caching disabled 
CreatePVC(...)
```
```python
# caching enabled
CreatePVC(...).set_caching_option(True)
```

I think this should be easy to do from SDK side with `CreatePVC` and `DeletePVC`?

Not so sure about `mount_pvc`, also not sure if we should change the caching setting for it. The argument for not doing it for `mount_pvc` is that any component taking a GCS url as an input is in the same situation--url may be the same value but the content may be different--and we don't change the caching setting for those. 

For `mount_pvc`, if we don't do it from SDK side, maybe we can count the pvc name into caching key calculation. That way, at least caching would be disabled when the pvc name is different. WDYT? @Linchin

connor-mccarthy: I like this approach in principle. Thinking about technical feasibility...

For `mount_pvc`, I think this is fairly easy because we have [access to the `task`](https://github.com/kubeflow/pipelines/blob/5b680a2bb2462ab491f41d3ea6f0c045dbceaa96/kubernetes_platform/python/kfp/kubernetes/volume.py#L66).

This seems harder for `CreatePVC` and `DeletePVC`, since they are components and only the user has access to their associated tasks, which are created in user code (not library code). Perhaps we can override the `__call__` to both create the task and call the task's `.set_caching_options` method before returning it, but implementation seems non-trivial (and maybe not possible... I'd need to do some research) to implement since we're overriding the call of an instance (the `PythonComponent` instance returned by the `@dsl.container_component`), not a subclass of `BaseComponent`.

For `CreatePVC` and `DeletePVC`, is there a simpler solution I'm not thinking of?

chensun: Okay, I think you're right.

> For CreatePVC and DeletePVC, is there a simpler solution I'm not thinking of?

Simpler solution would be opening up a component level cache setting in `BaseComponent`, which would be passed down as default cache setting on task. 
```python
@dsl.container_component(enable_cache=False)
def CreatePVC(...):
    ...
```

and such component level cache setting could make sense in certain scenarios, for example:
```python
@dsl.component(enable_cache=False)
def generate_random_number(...) -> float:
      ...
```
This becomes a more general question on whether we should expose a cache setting on component interface. I think we can follow up on that later. 
At this moment, maybe the AIs could be:
- No SDK change, but instruct users to not use cache through documentation
- BE adds volume name in cache key calculation, so `mount_pvc` with different names would not reuse cache.

WDYT about the AIs? @connor-mccarthy @Linchin

connor-mccarthy: >  but instruct users to not use cache through documentation

+1 on this for `CreatePVC` and `DeletePVC`. I prefer we do not begin a new pattern of adding task configurations parameters to the component decorators unless we get more evidence for its usefulness.

> BE adds volume name in cache key calculation, so `mount_pvc` with different names would not reuse cache.

Is it easier and/or preferred for the SDK to make this change (call `.set_caching_options` in the `mount_pvc` function)? This approach makes it (slightly) more transparent to the user (can inspect authoring reference docs and source easily) and is also overridable in authoring code.

Linchin: I think the PVC name, if generated by previous tasks, is already accounted for in the cache key generation, as the PVC name is part of the input. Additionally, currently the `createpvc` and `deletepvc` did not enable cache, because cache is usually created in launcher, and I didn't add that to `createpvc` and `deletepvc` which only runs a driver. So currently the only time cache is used is when `mount_pvc` uses a static pvc name. I can add the cache function, but I'm not sure if it's necessary given we will advise users to not use cache.

Linchin:
Changed design and used #9504 instead.

PR #9416, opened 2023-05-13
fix(backend): Add creation of pipeline and pipeline version atomically

gkcalat:

- Added a new endpoint to create a pipeline and pipeline version in a single transaction
- Existing flow: `/api/v2beta1/pipelines/upload` -> `UploadPipeline` -> `Pipeline` (creates a pipeline version too)
- This addition: `/api/v2beta1/pipelines/create` -> `CreatePipelineAndVersion` -> `Pipeline` (creates a pipeline version too)

**Checklist:**


backend/src/apiserver/server/pipeline_server.go
```diff
@@ -573,6 +573,94 @@ func NewPipelineServer(resourceManager *resource.ResourceManager, options *Pipel
...
+		createPipelineVersionRequests.Inc()
+	}
+
+	// URL is a required input for CreatePipelineAndVersion call
+	pipelineUrlStr := request.GetPipelineVersion().GetPackageUrl().GetPipelineUrl()
+	if pipelineUrlStr == "" {
+		return nil, util.NewInvalidInputError("Failed to create a pipeline due to missing pipeline URL")
+	}
+
+	// Get pipeline description
```
chensun, line 590: nit: such comment doesn't add value as it only repeats what the code says.

backend/src/apiserver/server/pipeline_server.go
```diff
@@ -573,6 +573,94 @@ func NewPipelineServer(resourceManager *resource.ResourceManager, options *Pipel
...
+	}
+
+	// Create a pipeline version
+	pipelineVersion := &model.PipelineVersion{
+		Name:            pipeline.Name,
+		PipelineSpecURI: pipelineUrlStr,
+		Description:     pipeline.Description,
+		Status:          model.PipelineVersionCreating,
+	}
+	pipelineUrl, err := url.ParseRequestURI(pipelineUrlStr)
```
chensun, line 634: The rest of the code are much duplicate with 
https://github.com/kubeflow/pipelines/blob/64b4b87f4b10d1b339e4df27784b65097946592c/backend/src/apiserver/server/pipeline_server.go#L578

Consider reusing the function.

backend/api/v2beta1/pipeline.proto
```diff
@@ -95,6 +95,14 @@ service PipelineService {
     };
   }
 
+  // Creates a new pipeline and a new pipeline version in a single transaction.
+  rpc CreatePipelineAndVersion(CreatePipelineAndVersionRequest) returns (Pipeline) {
+    option (google.api.http) = {
+      post: "/apis/v2beta1/pipelines/create"
```
chensun, line 101: This feels anti-pattern to include the verb in the path, should it be `"/apis/v2beta1/pipelineandversion"` instead?

backend/api/v2beta1/pipeline.proto
```diff
@@ -95,6 +95,14 @@ service PipelineService {
     };
   }
 
+  // Creates a new pipeline and a new pipeline version in a single transaction.
+  rpc CreatePipelineAndVersion(CreatePipelineAndVersionRequest) returns (Pipeline) {
```
chensun, line 99: Given that we have the test infra now, when adding new API endpoint, we should also add the integration test, either in 
https://github.com/kubeflow/pipelines/blob/master/backend/test/integration/pipeline_api_test.go
or https://github.com/kubeflow/pipelines/blob/master/backend/test/integration/pipeline_version_api_test.go

backend/src/apiserver/server/pipeline_server.go
```diff
@@ -573,6 +573,94 @@ func NewPipelineServer(resourceManager *resource.ResourceManager, options *Pipel
...
+	if description == "" {
+		description = request.GetPipelineVersion().GetDescription()
+	}
+
+	// Convert the input request. Fail fast if either pipeline or pipeline version is corrupted.
+	pipeline, err := toModelPipeline(request.GetPipeline())
+	if err != nil {
+		return nil, util.Wrap(err, "Failed to create a pipeline due to pipeline conversion error")
+	}
+	pipeline.Description = description
```
Linchin, line 601: The `description` may come from pipeline version, if the description field of pipeline is empty. Should we change the pipeline description in this case, since this will change the data user enter?

backend/src/apiserver/server/pipeline_server.go
```diff
@@ -573,6 +573,94 @@ func NewPipelineServer(resourceManager *resource.ResourceManager, options *Pipel
...
+
+	// Get pipeline name
+	displayName := request.GetPipeline().GetDisplayName()
+	if displayName == "" {
```
Linchin, line 605: Similar to above, if pipeline name is empty, should we just report an error instead of using the name of pipeline version?

backend/src/apiserver/server/pipeline_server.go
```diff
@@ -573,6 +573,94 @@ func NewPipelineServer(resourceManager *resource.ResourceManager, options *Pipel
 	return &PipelineServer{resourceManager: resourceManager, httpClient: http.DefaultClient, options: options}
 }
 
+// Creates a pipeline and a pipeline version in a single transaction.
+// Supports v2beta1 behavior.
+func (s *PipelineServer) CreatePipelineAndVersion(ctx context.Context, request *apiv2beta1.CreatePipelineAndVersionRequest) (*apiv2beta1.Pipeline, error) {
```
Linchin, line 578: Although the resources created here include both pipeline and pipeline version, we only return the pipeline. Should we return both of them here?

gkcalat: Thank you. This is to be consistent with `UploadPipeline`, which also returns a pipeline. We can consider changing the API to be closer to AIP requirements, but I would rather leave it to post-GA.

backend/src/apiserver/server/pipeline_server.go
gkcalat: This is a good point. Currently, UI does not have a way to set `DisplayName` and `Description` for a pipeline version. The fall-back logic is fairly confusing, so let me replace it with overwriting both fields in `PipelineVersion` from `Pipeline`. I will add a comment in the proto to be clear.

backend/src/apiserver/server/pipeline_server.go
gkcalat: I [proposed above](https://github.com/kubeflow/pipelines/pull/9416#discussion_r1196978863) to ignore `DisplayName` and `Description` fields from `PipelineVersion` and overwrite with whatever is provided in the `Pipeline` message.

backend/src/apiserver/server/pipeline_server.go
gkcalat: Sounds good. Let me fix it.

backend/src/apiserver/server/pipeline_server.go
gkcalat: Sounds good. Let me fix it.

backend/api/v2beta1/pipeline.proto
gkcalat: Sounds good. Let me fix it.

chensun (approved):
/lgtm
/approve

Thanks!

PR #9433, opened 2023-05-16
feat(frontend): make container name configurable

yhwang:
Add an UIConfig param to specify the container name in a pod where logs are retrieved. The default container name for the Argo is `main`. However, in Tekton, the main container name contains the `step-` prefix. Make this configurable to support other orchestrations.

**Checklist:**

### DIFF
--- a/frontend/server/app.ts
+++ b/frontend/server/app.ts
@@ -196,7 +196,7 @@ function createUIServer(options: UIConfigs) {
     registerHandler(
       app.get,
       '/k8s/pod/logs',
-      getPodLogsHandler(options.argo, options.artifacts, options.pod),
+      getPodLogsHandler(options.argo, options.artifacts, options.pod.logContainerName),
     );
   }
 
@@ -222,7 +222,7 @@ function createUIServer(options: UIConfigs) {
     registerHandler(
       app.get,
       '/k8s/pod/logs',
-      getPodLogsHandler(options.argo, options.artifacts, options.pod),
+      getPodLogsHandler(options.argo, options.artifacts, options.pod.logContainerName),
     );
   }
 
--- a/frontend/server/app.ts
+++ b/frontend/server/app.ts
@@ -193,7 +193,11 @@ function createUIServer(options: UIConfigs) {
       }),
     );
   } else {
-    registerHandler(app.get, '/k8s/pod/logs', getPodLogsHandler(options.argo, options.artifacts));
+    registerHandler(
+      app.get,
+      '/k8s/pod/logs',
+      getPodLogsHandler(options.argo, options.artifacts, options.pod),
+    );
   }
 
   if (options.artifacts.streamLogsFromServerApi) {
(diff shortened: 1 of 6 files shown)


yhwang:
appreciate the review @jlyaoyuli

frontend/server/workflow-helper.test.ts
```diff
@@ -83,7 +83,7 @@ describe('workflow-helper', () => {
       mockedGetPodLogs.mockResolvedValueOnce('pod logs');
 
       const stream = await getPodLogsStreamFromK8s('podName', 'namespace');
-      expect(mockedGetPodLogs).toBeCalledWith('podName', 'namespace');
+      expect(mockedGetPodLogs).toBeCalledWith('podName', 'namespace', undefined);
```
jlyaoyuli, line 86: Why is this mock called with ('podName', 'namespace', **undefined**) rather than ('podName', 'namespace', **'main'**)

yhwang: I guess default='main' is inside `getPodLogs`. when calling `getPodLogs` from `getPodLogsStreamFromK8s`, it doesn't specify the containerName. therefore, it's `undefined` here.

jlyaoyuli: To my understanding, the unspecified "optional" parameter can be `undefined`. It seems to me that if the `containerName` is not specified, the `getPodLogs` should be called with default value which is  **'main'** rather than `undefined`, because it is a **required** parameter. 

Would you mind changing the third parameter from `undefined` to **'main'** and see if the test still pass?

jlyaoyuli:
/approve

Thanks for the contribution, @yhwang!

frontend/server/app.ts
```diff
@@ -193,7 +193,11 @@ function createUIServer(options: UIConfigs) {
       }),
     );
   } else {
-    registerHandler(app.get, '/k8s/pod/logs', getPodLogsHandler(options.argo, options.artifacts));
+    registerHandler(
+      app.get,
+      '/k8s/pod/logs',
+      getPodLogsHandler(options.argo, options.artifacts, options.pod),
```
zijianjoy, line 199: Why is the `container name` being set via `options`? And the options name is `pod` instead of `container`. Won't you want to set container name as a request parameter?

yhwang: `options.pod` here is an object option for grouping the pod-related options, like `options.argo` and `options.artifacts`. All three of them are groups of options. currently the definition of options.pod is:
```
pod: {
    logContainerName: string;
}
```
Most of the time the container name for retrieving logs should be fixed. So in my change, I added an extra config in `UIConfigs` to specify the container name. And I grouped it under the `pod` sub-group.

frontend/server/workflow-helper.test.ts
yhwang: let me having the default value for both `getPodLogsStreamFromK8s` and `getPodLogs` and update the test case to avoid the confusion. :-)

frontend/server/app.ts
yhwang: oh, so you want me to just pass the `pod.logContainerName`, as the third argument but not the pod sub-group?

zijianjoy: Sounds good on both comments, and yes you confirmed correct about the latter comment. I am wondering if there is any other flag you want to introduce under `options.pod`?  If not, maybe just passing `options.pod.logContainerName` to `getPodLogsHandler` is sufficient?

yhwang: Sure! It only passes the `options.pod.logContainerName` now.

yhwang:
again appreciate your reviews @jlyaoyuli and @zijianjoy

PR #9449, opened 2023-05-17
fix(backend): Fix url encoding issue. Closes #9448

### LINKED ISSUE #9448: [backend] URL escaped parameter parsing error
### Environment

Standalone KFP from `master`

### Steps to reproduce

Try to upload a pipeline with `%` + any other special character in the pipeline name

### Expected result

Creating a new pipeline and a new pipeline version

### Materials and Reference

Error parsing unescaped params

/CC @jlyaoyuli 

---

gkcalat:

- Removed redundant parsing of unescaped strings that were causing errors

**Checklist:**

### DIFF
--- a/backend/src/apiserver/server/pipeline_server.go
+++ b/backend/src/apiserver/server/pipeline_server.go
@@ -16,7 +16,6 @@ package server
 
 import (
 	"context"
-	"fmt"
 	"net/http"
 	"net/url"
 	"path"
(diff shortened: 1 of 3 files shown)


Linchin:
I'm a little confused here. I remember that we only allow lowercase alphanumeric characters with dash in between, so user should fail at the UI when using pipeline name like `$S`. Am I missing something?
image

gkcalat:
> wercase alphanumeric characters with dash in between, so user should fail at the UI when using pipeline name like `$S`. Am I missing something?

Thank you. We decided to allow users to set any `DisplayName`. This is already implemented and checked into master.

This PR fixes a bug in the backend. Somehow we are try to decode already decoded parameters from URL. This causes issues when a parameter contains a `%` character. The decoded expects the following two symbols to be HEX (0-F), otherwise it returns an error. In a worse case, it will mutate user-defined DisplayName.

For cases, when the `DisplayName` is missing, the backend will try to replace it with pipeline's name from `spec.pipeline-info.name`. The `spec.pipeline-info.name` is validated and needs to follow the requirements from MLMD, as it's used for `Context`.

chensun (commented):
/lgtm
/approve

Thanks!

PR #9450, opened 2023-05-17
fix(frontend): Incorrect page title after switching to another pipeline version.

jlyaoyuli:
The pipeline version name in page title should be changed as well after switching to another pipeline version.

Before:

https://github.com/kubeflow/pipelines/assets/56132941/ec699e27-d07e-41f1-9a55-e9c2347d7432

After:

https://github.com/kubeflow/pipelines/assets/56132941/2df6f260-3ea0-4374-86fe-cfcf672055a3


chensun (approved):
/lgtm
/approve

PR #9461, opened 2023-05-19
feature(frontend): Support non-English characters and special characters in pipeline "display" name

jlyaoyuli:
Based on #9449

Previously, we don't support run name, which is partly from pipeline name, with non-English characters and special characters. A name precheck for pipeline display name (user-typed input) is set to avoid invalid name.
Currently, there's no limitation for run name any more. The precheck for pipeline name can be eliminated as well.

Screenshot 2023-05-19 at 10 17 05 AM
Screenshot 2023-05-19 at 10 18 06 AM

Both of run and pipeline can accept any characters

### DIFF
--- a/frontend/src/pages/NewPipelineVersion.test.tsx
+++ b/frontend/src/pages/NewPipelineVersion.test.tsx
@@ -118,7 +118,7 @@ describe('NewPipelineVersion', () => {
     it('creates pipeline version is default when landing from pipeline details page', () => {
       tree = shallow(
         <TestNewPipelineVersion
-          {...generateProps(`?${QUERY_PARAMS.pipelineId}=${MOCK_PIPELINE.id}`)}
+          {...generateProps(`?${QUERY_PARAMS.pipelineId}=${MOCK_PIPELINE.pipeline_id}`)}
         />,
       );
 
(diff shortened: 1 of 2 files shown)


frontend/src/pages/NewPipelineVersion.test.tsx
```diff
@@ -118,7 +118,7 @@ describe('NewPipelineVersion', () => {
     it('creates pipeline version is default when landing from pipeline details page', () => {
       tree = shallow(
         <TestNewPipelineVersion
-          {...generateProps(`?${QUERY_PARAMS.pipelineId}=${MOCK_PIPELINE.id}`)}
+          {...generateProps(`?${QUERY_PARAMS.pipelineId}=${MOCK_PIPELINE.pipeline_id}`)}
```
chensun, line 121: IIRC, this change is unrelated to the change--which seems to be only removing a name check? 
So question was/is this test actually working?

chensun (commented):
To clarify, this is only to allow non-ASCII characters in pipeline **display name**. The pipeline name more commonly referred to, which is user provided value when pipeline was defined and stored in IR, still needs to meet certain criteria. 
```python
@dsl.pipeline(name=...)
def my_pipeline(...):
     ...
```
Title should be explicit on "display  name" to avoid the confusion.

frontend/src/pages/NewPipelineVersion.test.tsx
jlyaoyuli: Great question! The short answer is: Yes, they are working. 

The `MOCK_PIPELINE.id` caused `pipelineId` to be `undefined`; however, those unit tests are validating the behavior which is independent to the `pipelineId`. Thus, the tests work even though we accidentally forget to fix it.

chensun (approved):
/lgtm
/approve

Thanks!

PR #9467, opened 2023-05-19
fix(frontend): Toolbar in pipeline details (from run) should only have "Clone run" button

jlyaoyuli:
We have 2 possible source for pipeline details (from version and from run). For the details from run (clone existing run or SDK-created run), action should be `Clone run` rather `Create run`. Also, we shouldn't allow `Upload version` in this case, because pipeline is not existed.

Pipeline details from version:

https://github.com/kubeflow/pipelines/assets/56132941/94ca9ab1-1617-4d82-8eb9-69da9c801538

Pipeline details from run (SDK-created run)

https://github.com/kubeflow/pipelines/assets/56132941/0133fa5e-a609-48ef-acb7-a121081a0a4f

Pipeline details from run (clone existing run)

https://github.com/kubeflow/pipelines/assets/56132941/c780a086-6a52-4687-8022-014585924e47

Pipeline details from run (after fix)

https://github.com/kubeflow/pipelines/assets/56132941/118341c2-0cad-4736-84ab-a99cdf28677d

### DIFF
--- a/frontend/src/pages/PipelineDetails.test.tsx
+++ b/frontend/src/pages/PipelineDetails.test.tsx
@@ -590,7 +590,7 @@ describe('PipelineDetails', () => {
     expect(newExperimentBtn).toBeDefined();
   });
 
-  it("has 'create run' toolbar button if viewing an embedded pipeline", async () => {
+  it("has 'clone run' toolbar button if viewing an embedded pipeline", async () => {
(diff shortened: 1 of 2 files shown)


chensun (approved):
/lgtm
/approve

Thanks!

chensun (approved):
/lgtm
/approve

Thanks!

PR #9440, opened 2023-05-16
feature(frontent): Support both "Flat" and "Group" view for Execution list

jlyaoyuli:
We supported `Flat` view and pagination feature for the execution list in our last rc release. However, we hope that the original `Group` view can be preserved as well. The main purpose of this PR is adding a little tab on the top of execution list for users to switch between both views.

[Demo]

https://github.com/kubeflow/pipelines/assets/56132941/c57d97c9-855d-40d5-9489-19e58ea4c205

PR #9470, opened 2023-05-22
feature(frontent): Support both "Flat" and "Grouped" view for Artifact list

jlyaoyuli:
Similar to #9440


zijianjoy (commented):
/lgtm
/approve

Good work Joe!

PR #9485, opened 2023-05-23
feat(backend): Update cache-server to only check if Pod is Succeeded. Fixes #9454

### LINKED ISSUE #9454: [feature] Cache Server to cache tasks as they complete instead of waiting on the entire pipeline.
### Feature Area

We make **heavy** use of Kubeflow's caching feature.
However, it doesn't play very well with some of the config you can set in Argo's `workflow-controller-configmap`. We run a large amount of pipelines at scale and as the number of pipelines increased the load on our EKS-Control plane also increased to the point where ~24k containers existed in one of our clusters and pretty much brought the entire cluster to a stand-still.
- After some debugging the root cause was found to be the interplay between the above configmap and the `cache-server`.
- To improve things we altered the `workflow-controller-configmap` to the following:
- ```
    spec:
        parallelism: 50  # Parallelism limits the max total parallel pods that can execute at the same time in a workflow
        # PodGC describes how to delete completed pods as they complete: https://argoproj.github.io/argo-workflows/fields/#podgc
        podGC:
            strategy: OnPodSuccess # Strategy is the strategy to use. One of "OnPodCompletion", "OnPodSuccess", "OnWorkflowCompletion", "OnWorkflowSuccess"
        # TTLStrategy limits the lifetime of a Workflow that has finished execution depending on if it Succeeded or Failed                                    
        ttlStrategy:
            secondsAfterSuccess: 60  # Time to live after workflow is successful
            secondsAfterFailure: 10800  # Time to live after workflow fails
            # secondsAfterCompletion: 604800 # Time to live after workflow is completed, replaces ttlSecondsAfterFinished

- This caused an issue with caching: Pods were being cleaned up before the `workflow` itself was marked as completed and so the pod was **not** added to the cache-db

The root cause of this problem is the following in `backend/src/cache/server/watcher.go`

```
    func isPodCompletedAndSucceeded(pod *corev1.Pod) bool {
	    return pod.ObjectMeta.Labels[ArgoCompleteLabelKey] == "true" && pod.Status.Phase == corev1.PodSucceeded
    }
```
- This means the watcher won't deem a pod ready to be cached until the entire pipeline has completed.

Ideally the above `isPodCompletedAndSucceeded` would simply check the whether or no the pod itself is successful, if yes we don't care about the rest of the pipeline and that pod/tasks metadata can be added to the cache-db

This would greatly increase the scalability of caching and also help to reduce the load on the Control Plane for people running pipelines at large scales as they will be able to clean pods up at a greater rate and still benefit from caching.

Not really, we've currently made the above change to our Fork of Kubeflow and are running this in our clusters

---

Love this idea? Give it a 👍.

CosyOranges:
This is because to cache tasks we shouldn't care if the entire pipeline has completed, once a Pod is Succeeded it must also have completed and can be entered into the cache database.

The change here is to change the CacheServer to only watch for `Succeeded` Pods
- From the `Succeeded` status it can be inferred the Pod is also `Completed`

The check for whether the Argo Workflow has completed `ObjectMeta.Labels[ArgoCompleteLabelKey] == "true"` has been removed as caching of a task shouldn't be dependent on the whole pipeline.
- This should speed up the rate of caching and it will also allow users to make use of Argos ability to do Pod Garbage Collection `onPodSuccess` instead of relying on the Time to Live of the workflow
    - Ultimately this will allow users to keep a "cleaner" cluster by having fewer `Succeeded` pods hanging around for long running workflows

**Checklist:**

### DIFF
--- a/backend/src/cache/server/watcher.go
+++ b/backend/src/cache/server/watcher.go
@@ -47,7 +47,7 @@ func WatchPods(ctx context.Context, namespaceToWatch string, clientManager Clien
 			}
 			log.Printf((*pod).GetName())
 
-			if !isPodCompletedAndSucceeded(pod) {
+			if !isPodSucceeded(pod) {
 				log.Printf("Pod %s is not completed or not in successful status.", pod.ObjectMeta.Name)
 				continue
 			}
@@ -108,8 +108,8 @@ func WatchPods(ctx context.Context, namespaceToWatch string, clientManager Clien
 	}
 }
 
-func isPodCompletedAndSucceeded(pod *corev1.Pod) bool {
-	return pod.ObjectMeta.Labels[ArgoCompleteLabelKey] == "true" && pod.Status.Phase == corev1.PodSucceeded
+func isPodSucceeded(pod *corev1.Pod) bool {
+	return pod.Status.Phase == corev1.PodSucceeded
 }
 
 func isCacheWriten(labels map[string]string) bool {
--- a/backend/src/cache/server/watcher_test.go
+++ b/backend/src/cache/server/watcher_test.go
@@ -25,7 +25,7 @@ func TestIsPodSucceeded(t *testing.T) {
 	runningPod.Status.Phase = "Running"
 	podIncomplete := isPodSucceeded(fakePod)
 	assert.False(t, podIncomplete)
-	// Assign pod as succeeded
+	// Assign pod as succeeded.
 	completedPod := *fakePod.DeepCopy()
 	completedPod.Status.Phase = "Succeeded"
 	podComplete := isPodSucceeded(&completedPod)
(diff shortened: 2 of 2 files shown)


CosyOranges:
I've also added a `watcher_test.go` file with some very basic unit tests... tbh this was mostly for my own practice but also as a "suggestion". Happy to remove if redundant!

chensun:
@CosyOranges I think this change might not work as intended, our caching using pipeline name as MLMD context, at least for v2 caching, not sure about v1 which is what you're changing here. In that case if pipeline didn't succeed, there might not be a valid context entry--not certain though. 

/cc @Linchin if she might know better for v2 caching.

CosyOranges:
Hi @chensun yeah I wasn't too sure about the V2 side of things 😅 , we've been running this at a very large scale on the V1 side with things working as expected

PR #9497, opened 2023-05-23
fix(frontend): Run name should be empty if no pipeline is selected

jlyaoyuli:
Before:

https://github.com/kubeflow/pipelines/assets/56132941/d29145ba-04e4-4645-be8b-bb4c862914a5

After:

https://github.com/kubeflow/pipelines/assets/56132941/81ecdd19-1aab-437d-a050-a7fb9d72a044


zijianjoy:
/lgtm
/approve

Thank you for the fix!

PR #8831, opened 2023-02-09
feat(frontend): Support namespaced pipelines from the UI. Closes #5084

tasos-ale:
The purpose of this PR is to introduce namespaced pipelines in the pipelines frontend.
Design doc: https://docs.google.com/document/d/1fM4y2L1IVqVj-iiNjYFRRktdCh7FQXgU2XpaYLaqt-A/edit?resourcekey=0-kd5loyP7w3PBD0ug6ECmLQ#heading=h.x9snb54sjlu9

The pipelines list page now has two tabs, one that displays the private pipelines and one for shared.

Screenshot from 2023-02-09 12-31-07

Screenshot from 2023-02-09 12-31-38

The new run page now includes radio buttons that help create a private or shared pipeline.

Screenshot from 2023-02-09 12-32-17

Screenshot from 2023-02-09 12-32-41

Similar to the pipelines list, the choose pipeline dialog has two tabs.

Screenshot from 2023-02-09 12-33-07

Screenshot from 2023-02-09 12-33-34

Finally, the upload pipeline dialog includes radio buttons for a private and shared pipeline.

Screenshot from 2023-02-09 12-33-56

In single-user mode, the pipelines list page and the choose pipeline dialog have no tabs. The radio buttons should also be hidden.

Part of #4197
Closes #5084

/cc @jlyaoyuli @zijianjoy @chensun @StefanoFioravanzo @elikatsis

PR #9499, opened 2023-05-23
fix(frontend): Choose button is disable even if pipeline is already selected (after uploading a new pipeline)

jlyaoyuli:
The enable / disable logic of choose button in pipeline version box was changed in #8831 , and cause the bug.

Before:

https://github.com/kubeflow/pipelines/assets/56132941/5aae95ef-e96f-47b3-984b-1308f43d9e96

After:

https://github.com/kubeflow/pipelines/assets/56132941/608cdf2f-2f1c-4906-8bce-f1f1fe8cd3f9


chensun (approved):
/lgtm
/approve

Thanks!

PR #9511, opened 2023-05-25
fix(frontend): Make pipeline definition in YAML tab more readable

jlyaoyuli:
Before:
Screenshot 2023-05-25 at 11 29 51 AM

After:
Screenshot 2023-05-25 at 11 30 12 AM

### DIFF
--- a/frontend/src/components/PipelineSpecTabContent.tsx
+++ b/frontend/src/components/PipelineSpecTabContent.tsx
@@ -31,9 +31,14 @@ export function PipelineSpecTabContent(props: PipelineSpecTabContentProps) {
     <Editor
       value={
         isTemplateV2(props.templateString)
-          ? jsyaml.safeDump(jsyaml.safeLoad(props.templateString || ''))
-          : props.templateString || ''
-      } // Use safeLoad and then safeDump to make sure the v2 PipelineSpec is in Yaml Form.
+          ? props.templateString || ''
+          : jsyaml.safeDump(jsyaml.safeLoad(props.templateString || ''))
+      }
+      // V2(YAML-formatted):
+      //    Directly use it to render in <PipelineSpecTabContent>
+      // V1(JSON-formatted):
+      //    safeLoad() convert string to object first, and safeDump() changes the object
+      //    to yaml format and then render it in <PipelineSpecTabContent>
       height={editorHeightWidth}
       width={editorHeightWidth}
       mode='yaml'
--- a/frontend/src/components/PipelineSpecTabContent.tsx
+++ b/frontend/src/components/PipelineSpecTabContent.tsx
@@ -17,7 +17,6 @@
 import jsyaml from 'js-yaml';
 import React from 'react';
 import { isSafari } from 'src/lib/Utils';
-import { isTemplateV2 } from 'src/lib/v2/WorkflowUtils';
 import Editor from './Editor';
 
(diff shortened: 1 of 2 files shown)


frontend/src/pages/PipelineDetailsV1.tsx
```diff
@@ -260,7 +261,9 @@ const PipelineDetailsV1: React.FC<PipelineDetailsV1Props> = ({
         )}
         {selectedTab === 1 && !!templateString && (
           <div className={css.containerCss} data-testid={'spec-yaml'}>
-            <PipelineSpecTabContent templateString={templateString || ''} />
+            <PipelineSpecTabContent
+              templateString={jsyaml.safeDump(jsyaml.safeLoad(templateString)) || ''}
```
zijianjoy, line 265: We should add some documentations here to explain why do we dump and load the templateString. People who read this change cannot find any reference in tests, either. 

In the documentation, explaining the followings is important:
- The input templateString can be a `json formatted` string
- SafeLoad can convert the string to object
- SafeDump will change the object to yaml format
- yaml format is acceptable by PipelineSpecTabContent

frontend/src/pages/PipelineDetailsV1.tsx
```diff
@@ -260,7 +261,14 @@ const PipelineDetailsV1: React.FC<PipelineDetailsV1Props> = ({
...
         {selectedTab === 1 && !!templateString && (
           <div className={css.containerCss} data-testid={'spec-yaml'}>
-            <PipelineSpecTabContent templateString={templateString || ''} />
+            {/* The templateString is from getPipelineTemplated() and it's JSON-formatted.
+              In order to make consistent with v2 pipeline behavior, we decide to render
+              it in yaml format.
+              safeLoad() convert string to object first, and safeDump() changes the object
+              to yaml format which is also acceptable by <PipelineSpecTabContent> */}
+            <PipelineSpecTabContent
+              templateString={jsyaml.safeDump(jsyaml.safeLoad(templateString)) || ''}
```
zijianjoy, line 270: Looks like we need to change this part to include V1 template instead, how do you think? https://github.com/kubeflow/pipelines/blob/master/frontend/src/components/PipelineSpecTabContent.tsx#L32-L36

frontend/src/components/PipelineSpecTabContent.tsx
```diff
@@ -31,9 +31,14 @@ export function PipelineSpecTabContent(props: PipelineSpecTabContentProps) {
     <Editor
       value={
         isTemplateV2(props.templateString)
-          ? jsyaml.safeDump(jsyaml.safeLoad(props.templateString || ''))
-          : props.templateString || ''
-      } // Use safeLoad and then safeDump to make sure the v2 PipelineSpec is in Yaml Form.
+          ? props.templateString || ''
+          : jsyaml.safeDump(jsyaml.safeLoad(props.templateString || ''))
```
zijianjoy, line 35: How about just removing the V2 template check?

`value={jsyaml.safeDump(jsyaml.safeLoad(props.templateString || '')}`

We can keep the comments below to explain difference between V1 and V2.

jlyaoyuli: If we remove the v2 template check, we shouldn't say "Directly use"(?)

PR #9519, opened 2023-05-26
fix(frontend): Unexpected page crash when clicking artifact node.

jlyaoyuli:
This issue is because of the uncaught error (from parseStoragePath() helper) in ArtifactPreview.
Also, instead of displaying error banner for unsupported artifact URI, we decide to directly render it.

Before:

https://github.com/kubeflow/pipelines/assets/56132941/1d370f5f-6e69-4d3c-90fb-9bc689bb51ce

After:

https://github.com/kubeflow/pipelines/assets/56132941/a0730837-4df5-445e-bd38-cca0a56b2179

### DIFF
--- a/frontend/src/components/ArtifactPreview.tsx
+++ b/frontend/src/components/ArtifactPreview.tsx
@@ -23,6 +23,7 @@ import WorkflowParser, { StoragePath } from 'src/lib/WorkflowParser';
 import { stylesheet } from 'typestyle';
 import Banner from './Banner';
 import { ValueComponentProps } from './DetailsTable';
+import { logger } from 'src/lib/Utils';
 
 const css = stylesheet({
   root: {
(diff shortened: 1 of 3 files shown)


frontend/src/components/ArtifactPreview.tsx
```diff
@@ -74,9 +79,7 @@ const ArtifactPreview: React.FC<ArtifactPreviewProps> = ({
   );
 
   if (!storage) {
-    return (
-      <Banner message={'Can not retrieve storage path from artifact uri: ' + value} mode='error' />
-    );
+    return value ? <div>{value}</div> : <div>No artifact URI is found.</div>;
```
chensun, line 82: nit: Artifact **URI** should always be available, it may not be a storage URL though. From that perspective, the original message seems okay except the `mode='error'` part? Up to you do decide what's the best non-error message here.

chensun (commented):
/lgtm

Thanks!

chensun (commented):
/lgtm

Thanks!

chensun (commented):
/lgtm

Thanks!

chensun (approved):
/lgtm
/approve

Thanks!

Issue #9553, opened 2023-06-02
[backend] Data too long for column 'StateHistory' at row 1


chensun:
### Environment

*  How did you deploy Kubeflow Pipelines (KFP)?
Standalone KFP

*  KFP version: 
2.0.0-rc.1

*  KFP SDK version: 
N/A

### Steps to reproduce
While doing some random debugging, I found the following error log in api-server pod:

```
I0602 05:14:55.623705       7 error.go:278] Error 1406: Data too long for column 'StateHistory' at row 1
[...]

PR #9554, opened 2023-06-02
fix(backend): fix column size of `StateHistory` in `task` table. Fixes: #9553

chensun:
Fixes: #9553

**Checklist:**

### DIFF
--- a/backend/src/apiserver/model/task.go
+++ b/backend/src/apiserver/model/task.go
@@ -33,7 +33,7 @@ type Task struct {
 	Name               string           `gorm:"column:Name; default:null"`
 	ParentTaskId       string           `gorm:"column:ParentTaskUUID; default:null"`
 	State              RuntimeState     `gorm:"column:State; default:null;"`
-	StateHistoryString string           `gorm:"column:StateHistory; default:null;"`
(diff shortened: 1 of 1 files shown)


gkcalat:
Thanks, @chensun 

/lgtm
/approve

backend/src/apiserver/model/task.go
```diff
@@ -33,7 +33,7 @@ type Task struct {
 	Name               string           `gorm:"column:Name; default:null"`
 	ParentTaskId       string           `gorm:"column:ParentTaskUUID; default:null"`
 	State              RuntimeState     `gorm:"column:State; default:null;"`
-	StateHistoryString string           `gorm:"column:StateHistory; default:null;"`
+	StateHistoryString string           `gorm:"column:StateHistory; default:null; size:65535;"`
```
chensun, line 36: @gkcalat, this actually feels a bit inaccurate and unreliable that `size:65535` maps to `longtext` in MySQL, because `text` also accomdates the 65535 size limit [1]. 
I opt to doing it this way in favor of much smaller change and less risky, but I feel like we might want to migrate all the columns that we design to be `longtext` to use a more explicit and reliable gorm size or type.

https://dev.mysql.com/doc/refman/8.0/en/storage-requirements.html#data-types-storage-reqs-strings

PR #9560, opened 2023-06-02
fix(frontend): Incorrect page text for empty pipeline.

jlyaoyuli:
For the case that API call returns error, the page text should be **"No graph to show"** rather than **"Currently loading pipeline information"**.

For example, an empty pipeline will get error from getPipelineVersion() call.

Before:

https://github.com/kubeflow/pipelines/assets/56132941/b1e0a9b5-b89a-4ecd-b9f3-af2f9001395e

After:

https://github.com/kubeflow/pipelines/assets/56132941/d1d3c4e4-2e63-4630-8189-d0c13d30e680

### DIFF
--- a/frontend/src/pages/PipelineDetails.test.tsx
+++ b/frontend/src/pages/PipelineDetails.test.tsx
@@ -14,7 +14,7 @@
  * limitations under the License.
  */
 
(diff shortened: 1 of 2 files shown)


chensun (approved):
/lgtm
/approve

Thanks!

PR #9576, opened 2023-06-06
feat(backend): Added metrics to be collected from failed/successful workflows

MGSousa:

- Added metrics related to failed and successful workflow runs to be exported to Prometheus in ResourceManager.
- These new metrics can be filtered by each Namespace (`profile`) and by DisplayName (`workflow`)
- Improvements in how the gauge registers the count by using a Collector.
    - Checks a specific metric when deleting a run if has already reached the minimum value of 0.

Tests are also updated to reflect the new changes

**Checklist:**

### DIFF
--- a/backend/src/apiserver/resource/resource_manager.go
+++ b/backend/src/apiserver/resource/resource_manager.go
@@ -53,7 +53,7 @@ var (
 		// display in which Kubeflow namespace the runs were triggered
 		"profile",
 
-		// display latest workflow name
+		// display workflow name
 		"workflow",
 	}
 
--- a/backend/src/apiserver/resource/resource_manager.go
+++ b/backend/src/apiserver/resource/resource_manager.go
@@ -49,11 +49,31 @@ import (
 
 // Metric variables. Please prefix the metric names with resource_manager_.
 var (
+	extraLabels = []string{
+		// display in which Kubeflow namespace the runs were triggered
+		"profile",
+
+		// display latest workflow name
+		"workflow",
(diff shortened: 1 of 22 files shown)


MGSousa:
Any feedback related to this feature?

backend/src/apiserver/resource/resource_manager.go
```diff
@@ -49,11 +49,31 @@ import (
 
 // Metric variables. Please prefix the metric names with resource_manager_.
 var (
+	extraLabels = []string{
+		// display in which Kubeflow namespace the runs were triggered
+		"profile",
```
chensun, line 54: What's the rationale for naming this `profile` instead of `namespace`?

backend/src/apiserver/resource/resource_manager.go
```diff
@@ -49,11 +49,31 @@ import (
...
+
+		// display latest workflow name
```
chensun, line 56: Why is `latest` necessary here?

chensun (commented):
/lgtm

Just a couple questions.

backend/src/apiserver/resource/resource_manager.go
MGSousa: I chose this since it is better to understand that the profile will always be the namespace that Kubeflow uses for specific workflow runs, not a generic namespace from the K8s Cluster.

chensun (approved):
/lgtm
/approve

Thanks!

PR #9578, opened 2023-06-06
fix(frontend): Pipeline version is not updated in new run page

jlyaoyuli:
In the pipeline details page, users can switch between different versions; however, pipeline version is not updated after clicking `Create run` button.

Before:

https://github.com/kubeflow/pipelines/assets/56132941/72285dc8-6d2a-429e-83a8-77fa73727c87

After:

https://github.com/kubeflow/pipelines/assets/56132941/f575cf0a-ace0-4eb1-8802-7c2032e895e0

### DIFF
--- a/frontend/src/pages/PipelineDetails.tsx
+++ b/frontend/src/pages/PipelineDetails.tsx
@@ -578,7 +578,6 @@ class PipelineDetails extends Page<{}, PipelineDetailsState> {
         v => v.pipeline_version_id === versionId,
       );
       const pageTitle = this.state.v1Pipeline.name?.concat(' (', selectedVersionV1?.name!, ')');
-      this.props.updateToolbar({ pageTitle });
 
       const selectedVersionPipelineTemplate = await this._getTemplateString(
         this.state.v1Pipeline.id!,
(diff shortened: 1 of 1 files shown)


chensun (approved):
/lgtm
/approve

Thanks!

frontend/src/pages/PipelineDetails.tsx
```diff
@@ -587,6 +586,8 @@ class PipelineDetails extends Page<{}, PipelineDetailsState> {
       this.props.history.replace({
         pathname: `/pipelines/details/${this.state.v1Pipeline.id}/version/${versionId}`,
       });
+      this.props.updateToolbar(this.getInitialToolbarState());
+      this.props.updateToolbar({ pageTitle });
```
zijianjoy, line 590: Thank you Joe for the change, can you alternate the `pageTitle` before calling `updateToolbar`?

That is:
```
let toolbarState = this.getInitialToolbarState();
toolbarState.pageTitle = pageTitle;
this.props.updateToolbar(toolbarState);
```

In this case, we avoid the intermediate state in toolbar rendering.

PR #9591, opened 2023-06-07
fix(backend): let components add default values

### LINKED ISSUE #9572: [backend] input parameter with default value not resolved at runtime.
### Environment

KFP standalone
*  KFP version: 
2.0.0rc.1
*  KFP SDK version: 
2.0.0rc.1

### Steps to reproduce

Run the following pipeline: 
https://github.com/kubeflow/pipelines/blob/63936c07735521b1244e1e528a9baf80d40b296d/sdk/python/test_data/pipelines/xgboost_sample_pipeline.py
https://github.com/kubeflow/pipelines/blob/63936c07735521b1244e1e528a9baf80d40b296d/sdk/python/test_data/pipelines/xgboost_sample_pipeline.yaml

### Expected result

Expect pipeline to run successfully.

### Actual result

Pipeline failed due to input parameter with default value not resolved at runtime:
image

Linchin:
Background: #9572 
fixes #9572 

Adds a logic in the launcher to add parameter default values to the executor input, if there is no user value. Tested manually that the default values are resolved now.

Note: 
* Need to update launcher image once approved
* The mentioned sample still doesn't work, because "IfPresent placeholder" is not implemented yet
 
**Checklist:**

### DIFF
--- a/backend/src/v2/component/launcher_v2.go
+++ b/backend/src/v2/component/launcher_v2.go
@@ -260,7 +260,23 @@ func (l *LauncherV2) publish(
 	return l.metadataClient.PublishExecution(ctx, execution, outputParameters, outputArtifacts, status)
 }
 
-func executeV2(ctx context.Context, executorInput *pipelinespec.ExecutorInput, component *pipelinespec.ComponentSpec, cmd string, args []string, bucket *blob.Bucket, bucketConfig *objectstore.Config, metadataClient *metadata.Client, namespace string, k8sClient *kubernetes.Clientset) (*pipelinespec.ExecutorOutput, []*metadata.OutputArtifact, error) {
+func executeV2(
+	ctx context.Context,
+	executorInput *pipelinespec.ExecutorInput,
+	component *pipelinespec.ComponentSpec,
+	cmd string,
+	args []string,
+	bucket *blob.Bucket,
+	bucketConfig *objectstore.Config,
+	metadataClient *metadata.Client,
+	namespace string,
+	k8sClient *kubernetes.Clientset,
+) (*pipelinespec.ExecutorOutput, []*metadata.OutputArtifact, error) {
+
+	// Add parameter default values to executorInput, if there is not already a user input.
+	// This process is done in the launcher because we let the component resolve default values internally.
+	executorInput = addDefaultParams(executorInput, component)
+
 	executorOutput, err := execute(ctx, executorInput, cmd, args, bucket, bucketConfig, namespace, k8sClient)
 	if err != nil {
 		return nil, nil, err
@@ -695,4 +711,17 @@ func prepareOutputFolders(executorInput *pipelinespec.ExecutorInput) error {
 	}
 
 	return nil
+}
+
+func addDefaultParams(
+	executorInput *pipelinespec.ExecutorInput,
+	component *pipelinespec.ComponentSpec,
+) *pipelinespec.ExecutorInput {
+	for name, value := range component.GetInputDefinitions().GetParameters() {
+		_, hasInput := executorInput.GetInputs().GetParameterValues()[name]
+		if value.GetDefaultValue() != nil && !hasInput {
+			executorInput.GetInputs().GetParameterValues()[name] = value.GetDefaultValue()
+		}
+	}
+	return executorInput
 }


chensun:
Thank you @Linchin for the quick fix. Is there a way we can unit test the change?

Linchin:
I think it would be best to include this use case as part of the samples test, what do you think?

chensun:
> I think it would be best to include this use case as part of the samples test, what do you think?

Sample test is an integration or e2e test, it doesn't replace the need to have unit test. Ideally every code path should be unit tested. This one is a clear example that it can and thus should be unit tested--it's single function you can send inputs and verify outputs as expected. The longer we delay improving test coverage, the more tech debt we accumulate. I would strongly suggest we take a stab at adding missing UTs with all new changes.

backend/src/v2/component/launcher_v2.go
```diff
@@ -696,3 +712,16 @@ func prepareOutputFolders(executorInput *pipelinespec.ExecutorInput) error {
...
 }
+
+func addDefaultParams(
+	executorInput *pipelinespec.ExecutorInput,
+	component *pipelinespec.ComponentSpec,
+) *pipelinespec.ExecutorInput {
+	for name, value := range component.GetInputDefinitions().GetParameters() {
+		_, hasInput := executorInput.GetInputs().GetParameterValues()[name]
+		if value.GetDefaultValue() != nil && !hasInput {
+			executorInput.GetInputs().GetParameterValues()[name] = value.GetDefaultValue()
```
chensun, line 723: As you update `executorInput` in place, there could be potentially no difference whether you add default values here vs. earlier at this point: https://github.com/kubeflow/pipelines/blob/c0640d7a033b6fa2133b92c2a459d9eb3dabee32/backend/src/v2/driver/driver.go#L824

And if I read the `execute` function correctly, this change reverts our earlier decision on not passing in default value explicitly. Did you see such behavior change when testing pipeline with default value?

Linchin: We had some discussion offline on this. The executor input displayed at the UI comes from the file that driver writes, so changing it in the launcher does not affect that. However, it is still not advisable to alter `executorInput`, in case it may affect later functionalities. Thus I created a deep copy of `executorInput` for this purpose. I also added a few unit tests, although they may bring some flakiness for developers outside of Google, due to authentication issue caused by opening bucket.

backend/src/v2/component/launcher_v2.go
```diff
@@ -673,7 +714,7 @@ func localPathForURI(uri string) (string, error) {
 func prepareOutputFolders(executorInput *pipelinespec.ExecutorInput) error {
 	for name, parameter := range executorInput.GetOutputs().GetParameters() {
 		dir := filepath.Dir(parameter.OutputFile)
-		if err := os.MkdirAll(dir, 0644); err != nil {
+		if err := os.MkdirAll(dir, 0777); err != nil {
```
chensun, line 717: Why changing the permission?

backend/src/v2/component/launcher_v2.go
```diff
@@ -696,3 +737,27 @@ func prepareOutputFolders(executorInput *pipelinespec.ExecutorInput) error {
...
+	executorInput *pipelinespec.ExecutorInput,
+	component *pipelinespec.ComponentSpec,
+) (*pipelinespec.ExecutorInput, error) {
+	// Make a deep copy so we don't alter the original data
+	executorInputWithDefaultMsg := proto.Clone(executorInput)
+	executorInputWithDefault, ok := executorInputWithDefaultMsg.(*pipelinespec.ExecutorInput)
+	if executorInputWithDefault.GetInputs().GetParameterValues() == nil {
+		executorInputWithDefault.Inputs.ParameterValues = make(map[string]*structpb.Value)
+	}
+	if !ok {
```
chensun, line 752: This check should follow right after when ok got the value.

backend/src/v2/component/launcher_v2_test.go
```diff
@@ -0,0 +1,127 @@
...
+			bucket, err := blob.OpenBucket(nil, "gs://ml-pipeline-test")
+			assert.Nil(t, err)
+			bucketConfig, err := objectstore.ParseBucketConfig("gs://ml-pipeline-test/pipeline-root/")
+			assert.Nil(t, err)
+			addNumbersExecutorArgs := []string{"-c", "python3 -m venv .venv && source .venv/bin/activate\nexport CLUSTER_SPEC=\"{\\\"task\\\":{\\\"type\\\":\\\"workerpool0\\\",\\\"index\\\":0,\\\"trial\\\":\\\"TRIAL_ID\\\"}}\"\nif ! [ -x \"$(command -v pip)\" ]; then\n    python3 -m ensurepip || python3 -m ensurepip --user || apt-get install python3-pip\nfi\n\nPIP_DISABLE_PIP_VERSION_CHECK=1 python3 -m pip install --quiet     --no-warn-script-location 'kfp==2.0.0-rc.1' && \"$0\" \"$@\"\nprogram_path=$(mktemp -d)\nprintf \"%s\" \"$0\" > \"$program_path/ephemeral_component.py\"\npython3 -m kfp.components.executor_main                         --component_module_path                         \"$program_path/ephemeral_component.py\"                         \"$@\"\n", "\nimport kfp\nfrom kfp import dsl\nfrom kfp.dsl import *\nfrom typing import *\n\ndef add_numbers(a: int, b: int) -> int:\n    return a + b\n\n", "--executor_input", "{{$}}", "--function_to_execute", "add_numbers"}
```
chensun, line 114: Given this args string, which contains only `{{$}}` but not something like ``{$.inputs.parameters['name']}`, the test doesn't actually cover the bug you're trying to fix, right?

chensun (commented):
Thank you @Linchin for adding UTs. 
Left a few comments, and there seems to be some build error in presubmit.

backend/src/v2/component/launcher_v2.go
Linchin: Thank youfor catching this! Added the initialization to the wrong line. I will fix this.

backend/src/v2/component/launcher_v2.go
Linchin: I'm not entirely sure why, but if we keep the previous permission, the code doesn't have the permission to write to these created folders.

chensun: We can check back on it later.

chensun (commented):
/lgtm 

@Linchin Please go ahead and update the launcher image. Thanks!

chensun (approved):
/lgtm
/approve

Thanks!

PR #9594, opened 2023-06-08
fix(frontend): Splitting logsDetails into lines based on CR and LF. Fixes #9593

### LINKED ISSUE #9593: [frontend] tqdm logs not displayed correctly
### Environment

*  How did you deploy Kubeflow Pipelines (KFP)? manifests

*  KFP version: [2.0.0-rc.1](https://github.com/kubeflow/pipelines/releases/tag/2.0.0-rc.1)

### Steps to reproduce

Here is an example of kubeflow components:
```yaml
name: Example
description: Pass

implementation:
  container:
    image: python:alpine3.18
    command:
      - /bin/sh
    args:
      - "-c"
      - "pip install tqdm && seq 9999999 | tqdm --bytes | wc -l"
```

A simple pipeline
```python
from kfp.components import load_component_from_file
import kfp.dsl as dsl

example_op = load_component_from_file("example.yaml")

@dsl.pipeline
def example():
    example_op()
```
### Expected result

When fetching logs from the pod using **kubectl**, we get the following logs:

image

However in the kubeflow logs tab:

image

The logs are not returning in lines.

In the logs endpoint (`pipeline/k8s/pod/logs?podname=...`) we get:
image

## Investigation

Using notepad++, we can see the differences between the ends of lines:

CR = Carriage Return (\r)
LF = Line Feed (\n)

image

---

axel7083:

Fixing #9593

Replacing `.split('\n')` with `.split(/[\r\n]+/)` to ensure lines are properly split. 

## Tests:

After fixing the split argument:
image

### DIFF
--- a/frontend/src/components/tabs/RuntimeNodeDetailsV2.tsx
+++ b/frontend/src/components/tabs/RuntimeNodeDetailsV2.tsx
@@ -202,7 +202,7 @@ function TaskNodeDetail({
             )}
             {!logsBannerMessage && (
               <div className={commonCss.pageOverflowHidden} data-testid={'logs-view-window'}>
-                <LogViewer logLines={(logsDetails || '').split('\n')} />
+                <LogViewer logLines={(logsDetails || '').split(/[\r\n]+/)} />
               </div>
             )}
           </div>


zijianjoy:
/lgtm
/approve

Thank you for the contribution!

PR #9599, opened 2023-06-09
fix(frontend): Display experiment name if experiment is selected

jlyaoyuli:
When the user enter the new run page from experiment details page, it means that the experiment is selected. The text field of experiment name should be automatically filled.

Before:

https://github.com/kubeflow/pipelines/assets/56132941/5c087a45-579a-46b0-aa39-1758c8403448

After:

https://github.com/kubeflow/pipelines/assets/56132941/e4a56290-f79f-4d72-bdd0-b9ab943bbccd


chensun (approved):
/lgtm
/approve 

Thanks!

Issue #6845, opened 2021-11-01
[backend] severe perf problem in archive experiment API


Bobgy:
### Environment

*  KFP version: master

### Steps to reproduce

1. On a KFP cluster with many runs/experiments.
2. Use archive experiment API on any experiment, observe that the API takes >1 minute to process and may timeout + cause other queries to fail. Observed failure look like https://github.com/kubeflow/pipelines/issues/6815#issue-1037408014

The issue caused severe flakiness in test infra: https://github.com/kubeflow/pipelines/issues/6815, and was mitigated by skipping the archive experiment step in test script: https://github.com/kubeflow/pipelines/pull/6843 (We should revert the https://github.com/kubeflow/pipelines/pull/6843 after fixing the perf problem)

### Expected result

There's no perf problem with experiment archive API.

### Materials and Reference

This is caused by a known TODO item to improve SQL perf: https://github.com/kubeflow/pipelines/blob/74c7773ca40decfd0d4ed40dc93a6af591bbc190/backend/src/apiserver/storage/experiment_store.go#L266

I experimented with a SQL query improvement in https://github.com/kubeflow/pipelines/issues/6815#issuecomment-955935043, we still need to investigate how to turn that into our golang backend.

---

Impacted by this bug? Give it a 👍. We prioritise the issues with the most 👍.

lfloeer:
We just today got hit by this issue (Kubeflow 1.3 deployment). Would really appreciate if this could get addressed in a future version. Right now, we have to tell our users to not archive experiments. An unfortunate side effect of the performance issue is, that since this results in an `UPDATE` statement, the `run_details` and `resource_references` table are locked during this operation, leading to other failed API requests, e.g. updating a single run, etc.

gkcalat:
Fixed by @difince.

/close

PR #6843, opened 2021-11-01
test: skip archiving experiments because of perf problem

Bobgy:
**Description of your changes:**
Mitigates https://github.com/kubeflow/pipelines/issues/6815#issuecomment-955938098

**Checklist:**
- [ ] The title for your pull request (PR) should follow our title convention. [Learn more about the pull request title convention used in this repository](https://github.com/kubeflow/pipelines/blob/master/CONTRIBUTING.md#pull-request-title-convention).

Issue #6815, opened 2021-10-27
[testing] kfp-ci cluster related tests flaky


Bobgy:
[example error log](https://oss-prow.knative.dev/view/gs/oss-prow/pr-logs/pull/kubeflow_pipelines/6746/kubeflow-pipelines-samples-v2/1450527267473068032#1:build-log.txt%3A224)

> HTTP response headers: HTTPHeaderDict({'Content-Type': 'application/json', 'Date': 'Tue, 19 Oct 2021 18:22:34 GMT', 'Vary': 'Origin', 'X-Content-Type-Options': 'nosniff', 'X-Frame-Options': 'SAMEORIGIN', 'X-Powered-By': 'Express', 'X-Xss-Protection': '0', 'Transfer-Encoding': 'chunked', 'Set-Cookie': 'S=cloud_datalab_tunnel=HMzjedbUtOfvCUxatmKXGynHsJuhMxw5of6HG3PVShE; Path=/; Max-Age=3600'})
HTTP response body: {"error":"Failed to create a new run.: InternalServerError: Failed to store run v2-sample-test-zcmzl to table: Error 1205: Lock wait timeout exceeded; try restarting transaction","code":13,"message":"Failed to create a new run.: InternalServerError: Failed to store run v2-sample-test-zcmzl to table: Error 1205: Lock wait timeout exceeded; try restarting transaction","details":[{"@type":"type.googleapis.com/api.Error","error_message":"Internal Server Error","error_details":"Failed to create a new run.: InternalServerError: Failed to store run v2-sample-test-zcmzl to table: Error 1205: Lock wait timeout exceeded; try restarting transaction"}]}

Bobgy:
Here's current DB size:

```
mysql> select count(*) from run_details;
+----------+
| count(*) |
+----------+
|   150004 |
+----------+
1 row in set (5.98 sec)
mysql> select count(*) from experiments;
+----------+
| count(*) |
+----------+
|    65422 |
+----------+
1 row in set (0.07 sec)
```

Some queries on run seem to run for a very long time.

I'm trying to use `show full processlist` to figure out which ones take a long time.

Bobgy:
This query stays in preparing state for more than 1 minute:
```
UPDATE run_details SET StorageState = ? WHERE UUID in (SELECT ResourceUUID FROM resource_references as rf WHERE (rf.ResourceType = ? AND rf.ReferenceUUID = ? AND rf.ReferenceType = ?))
```

[...]

PR #9680, opened 2023-06-23
fix(backend): Fix performance issue within a mysql request

difince:
One of the requests used for Archiving Experiments is rewritten to use inner join for better performance, instead of using nested select. The fix leverage 'SQLDialect' interface, because the new Mysql request is not supported by SQLite Dialect (used for testing). 
There are other MySQL requests that use nested requests in Archive Experiment, but only the most problematic one reported by this [comment](https://github.com/kubeflow/pipelines/issues/6815#issuecomment-955935043) (bobgy) is fixed in the PR. 
The other requests are not optimized, as the `resource_reference` table will be deprecated soon - once `v2beta1` is available. 

Issue: https://github.com/kubeflow/pipelines/issues/6845

**Note**: Archive experiments need to be enabled back in test infra. The change done in this [PR](https://github.com/kubeflow/pipelines/pull/6843) needs to be reverted once this PR got merged. 

**Checklist:**

### DIFF
--- a/backend/src/apiserver/storage/db.go
+++ b/backend/src/apiserver/storage/db.go
@@ -23,6 +23,8 @@ import (
 	"github.com/VividCortex/mysqlerr"
 	"github.com/go-sql-driver/mysql"
 	sqlite3 "github.com/mattn/go-sqlite3"
+
+	"github.com/kubeflow/pipelines/backend/src/apiserver/model"
 )
 
 // DB a struct wrapping plain sql library with SQL dialect, to solve any feature
@@ -59,6 +61,9 @@ type SQLDialect interface {
 
 	// Inserts new rows and updates duplicates based on the key column.
 	Upsert(query string, key string, overwrite bool, columns ...string) string
+
+	// Updates a run_details based on join (production) or inner select (test).
+	UpdateRunDetailsStorageState(expId string) (string, []interface{})
 }
 
 // MySQLDialect implements SQLDialect with mysql dialect implementation.
@@ -86,6 +91,13 @@ func (d MySQLDialect) Concat(exprs []string, separator string) string {
 func (d MySQLDialect) IsDuplicateError(err error) bool {
 	sqlError, ok := err.(*mysql.MySQLError)
 	return ok && sqlError.Number == mysqlerr.ER_DUP_ENTRY
+}
+
+// TODO(gkcalat): deprecate resource_references table once we migration to v2beta1 is available.
+func (d MySQLDialect) UpdateRunDetailsStorageState(expId string) (string, []interface{}) {
+	var args []interface{}
+	args = append(args, model.StorageStateArchived.ToString(), model.RunResourceType, expId, model.ExperimentResourceType)
+	return "UPDATE run_details as runs JOIN resource_references as rf ON runs.UUID = rf.ResourceUUID SET StorageState = ? WHERE rf.ResourceType = ? AND rf.ReferenceUUID = ? AND rf.ReferenceType = ?", args
 }
 
 // SQLiteDialect implements SQLDialect with sqlite dialect implementation.
@@ -129,6 +141,13 @@ func (d SQLiteDialect) Upsert(query string, key string, overwrite bool, columns 
 func (d SQLiteDialect) IsDuplicateError(err error) bool {
 	sqlError, ok := err.(sqlite3.Error)
 	return ok && sqlError.Code == sqlite3.ErrConstraint
+}
+
+// TODO(gkcalat): deprecate resource_references table once we migration to v2beta1 is available.
+func (d SQLiteDialect) UpdateRunDetailsStorageState(expId string) (string, []interface{}) {
+	var args []interface{}
+	args = append(args, model.StorageStateArchived.ToString(), model.StorageStateArchived.ToString(), model.RunResourceType, expId, model.ExperimentResourceType, expId)
+	return "UPDATE run_details SET StorageState = ? WHERE StorageState <> ? AND UUID in (SELECT ResourceUUID FROM resource_references as rf WHERE (rf.ResourceType = ? AND rf.ReferenceUUID = ? AND rf.ReferenceType = ?)) OR ExperimentUUID = ?", args
 }
 
 func NewMySQLDialect() MySQLDialect {
--- a/backend/src/apiserver/storage/db.go
+++ b/backend/src/apiserver/storage/db.go
@@ -93,7 +93,7 @@ func (d MySQLDialect) IsDuplicateError(err error) bool {
 
 // UpdateFromOrJoin TODO(gkcalat): deprecate resource_references table once we migration to v2beta1 is available.
 func (d MySQLDialect) UpdateWithJointOrFrom(targetTable, joinTable, setClause, joinClause, whereClause string) string {
-	return fmt.Sprintf("UPDATE %s JOIN %s ON %s SET %s WHERE %s", targetTable, joinTable, joinClause, setClause, whereClause)
+	return fmt.Sprintf("UPDATE %s LEFT JOIN %s ON %s SET %s WHERE %s", targetTable, joinTable, joinClause, setClause, whereClause)
 }
 
 // SQLiteDialect implements SQLDialect with sqlite dialect implementation.
--- a/backend/src/apiserver/storage/db.go
+++ b/backend/src/apiserver/storage/db.go
@@ -23,8 +23,6 @@ import (
 	"github.com/VividCortex/mysqlerr"
 	"github.com/go-sql-driver/mysql"
 	sqlite3 "github.com/mattn/go-sqlite3"
-
-	"github.com/kubeflow/pipelines/backend/src/apiserver/model"
 )
 
 // DB a struct wrapping plain sql library with SQL dialect, to solve any feature
@@ -62,8 +60,8 @@ type SQLDialect interface {
 	// Inserts new rows and updates duplicates based on the key column.
 	Upsert(query string, key string, overwrite bool, columns ...string) string
 
-	// Updates a run_details based on join (production) or inner select (test).
-	UpdateRunDetailsStorageState(expId string) (string, []interface{})
+	// Updates a table using UPDATE with JOIN (mysql/production) or UPDATE FROM (sqlite/test).
+	UpdateWithJointOrFrom(targetTable, joinTable, setClause, joinClause, whereClause string) string
 }
 
 // MySQLDialect implements SQLDialect with mysql dialect implementation.
@@ -93,11 +91,9 @@ func (d MySQLDialect) IsDuplicateError(err error) bool {
 	return ok && sqlError.Number == mysqlerr.ER_DUP_ENTRY
 }
 
-// TODO(gkcalat): deprecate resource_references table once we migration to v2beta1 is available.
-func (d MySQLDialect) UpdateRunDetailsStorageState(expId string) (string, []interface{}) {
-	var args []interface{}
-	args = append(args, model.StorageStateArchived.ToString(), model.RunResourceType, expId, model.ExperimentResourceType)
-	return "UPDATE run_details as runs JOIN resource_references as rf ON runs.UUID = rf.ResourceUUID SET StorageState = ? WHERE rf.ResourceType = ? AND rf.ReferenceUUID = ? AND rf.ReferenceType = ?", args
+// UpdateFromOrJoin TODO(gkcalat): deprecate resource_references table once we migration to v2beta1 is available.
+func (d MySQLDialect) UpdateWithJointOrFrom(targetTable, joinTable, setClause, joinClause, whereClause string) string {
+	return fmt.Sprintf("UPDATE %s JOIN %s ON %s SET %s WHERE %s", targetTable, joinTable, joinClause, setClause, whereClause)
 }
 
 // SQLiteDialect implements SQLDialect with sqlite dialect implementation.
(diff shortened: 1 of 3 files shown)


backend/src/apiserver/storage/db.go
```diff
@@ -59,6 +61,9 @@ type SQLDialect interface {
 
 	// Inserts new rows and updates duplicates based on the key column.
 	Upsert(query string, key string, overwrite bool, columns ...string) string
+
+	// Updates a run_details based on join (production) or inner select (test).
+	UpdateRunDetailsStorageState(expId string) (string, []interface{})
```
Linchin, line 66: Hi @difince, thank you so much for your help in improving the query performance! It looks like the existing methods in this interface `SQLDialect` are mostly common database operations (except `IsDuplicateError()`). Can we do something similar here? We can use [`UpdateJoin()`](https://stackoverflow.com/questions/19270259/update-with-join-in-sqlite) or [`UpdateFrom()`](https://www.sqlite.org/lang_update.html#upfrom). Also according to [here](https://stackoverflow.com/a/69831549/8216546), SQLite supports UPDATE FROM starting v3.33. If we are using a version later than that, we don't have to use nested selects with SQLite either.

juliusvonkohout: Then maybe it would make sense to improve all database queries. We are also experience heavy slowdowns with the listruns database query.

difince: Hello @Linchin Thank you for your review and comments. 
I rewrote the SQLite request to use the UPDATE-FROM construct. This made the code a bit clearer, and I was able to abstract away (at some degree) the implementation within `SQLDialect` - trying to make it a bit more like a `common database operation`

I was unable to make the SQLite works with UpdateJoin. Most of the examples I browsed were using inner select, including the one referenced in this link - [UpdateJoin](https://stackoverflow.com/questions/19270259/update-with-join-in-sqlite). I have checked the SQLite version we work with - `3.39.4`
Still, if we take a closer look at the diagram of SQLite's official documentation about [Update Statement ](https://www.sqlite.org/lang_update.html)- and goes deeper in `join-clause` and `join-operator` - it looks like there is some support - but I really didn't make it work and didn't find any reference in the internet. 

Still, my last commit improves the code.

difince: Here are the current versions of the requests, if this will make the review easier
 
**MySQL**:
`UPDATE run_details JOIN resource_references ON run_details.UUID = resource_references.ResourceUUID SET StorageState = "STORAGESTATE_ARCHIVED" WHERE resource_references.ResourceType = "RUN" AND resource_references.ReferenceUUID = "<experiment_id>" AND resource_references.ReferenceType = "Experiment"
`

**SQLite**:
`UPDATE run_details SET StorageState = "STORAGESTATE_ARCHIVED" FROM resource_references WHERE run_details.UUID = resource_references.ResourceUUID AND resource_references.ResourceType = "RUN" AND resource_references.ReferenceUUID = "<experiment_id>" AND resource_references.ReferenceType = "Experiment"`

backend/src/apiserver/storage/db.go
```diff
@@ -88,6 +91,11 @@ func (d MySQLDialect) IsDuplicateError(err error) bool {
 	return ok && sqlError.Number == mysqlerr.ER_DUP_ENTRY
 }
 
+// UpdateFromOrJoin TODO(gkcalat): deprecate resource_references table once we migration to v2beta1 is available.
+func (d MySQLDialect) UpdateWithJointOrFrom(targetTable, joinTable, setClause, joinClause, whereClause string) string {
+	return fmt.Sprintf("UPDATE %s JOIN %s ON %s SET %s WHERE %s", targetTable, joinTable, joinClause, setClause, whereClause)
+}
```
gkcalat, line 97: Could we limit this function's scope to "UPDATE %s JOIN %s ON %s" and let users set the remaining via `squirrel.SetMap` and `squirrel.Where`? That way, this function will only implement the missing part.

backend/src/apiserver/storage/db_test.go
```diff
@@ -103,11 +103,35 @@ func TestSQLiteDialect_Upsert(t *testing.T) {
 }
 
 func TestMySQLDialect_Upsert(t *testing.T) {
-	sqliteDialect := NewMySQLDialect()
-	actualQuery := sqliteDialect.Upsert(`insert into table (uuid, name, namespace) values ("a", "item1", "kubeflow"),("b", "item1", "kubeflow")`, "namespace", true, []string{"uuid", "name"}...)
+	mysqlDialect := NewMySQLDialect()
```
gkcalat, line 106: Thanks for catching this!

backend/src/apiserver/storage/experiment_store.go
```diff
@@ -309,32 +309,6 @@ func (s *ExperimentStore) ArchiveExperiment(expId string) error {
 			"Failed to create query to archive experiment %s. error: '%v'", expId, err.Error())
 	}
 
-	// TODO(gkcalat): deprecate resource_references table once we migration to v2beta1 is available.
```
gkcalat, line 312: We should apply the same change to `jobs` (Recurring Runs) down below.

backend/src/apiserver/storage/experiment_store.go
```diff
@@ -309,32 +309,6 @@ func (s *ExperimentStore) ArchiveExperiment(expId string) error {
...
-			"StorageState": model.StorageStateArchived.ToString(),
-		}).
-		Where(sq.NotEq{"StorageState": model.StorageStateArchived.ToString()}).
-		Where(fmt.Sprintf("UUID in (%s) OR ExperimentUUID = '%s'", filteredRunsSql, expId), filteredRunsArgs...).
-		ToSql()
-	if err != nil {
-		return util.NewInternalServerError(err,
-			"Failed to create query to archive the runs in an experiment %s. error: '%v'", expId, err.Error())
-	}
-
```
gkcalat, line 337: Should we remove similar lines for jobs (recurring runs) down below?

backend/src/apiserver/storage/db.go
gkcalat: NIT: replace `JOIN` with explicit `INNER JOIN` for better readability

backend/src/apiserver/storage/experiment_store.go
```diff
@@ -388,7 +362,15 @@ func (s *ExperimentStore) ArchiveExperiment(expId string) error {
...
-	_, err = tx.Exec(updateRunsSql, updateRunsArgs...)
+	var arguments []interface{}
+	arguments = append(arguments, model.StorageStateArchived.ToString(), model.RunResourceType, expId, model.ExperimentResourceType)
+	query := s.db.UpdateWithJointOrFrom(
+		"run_details",
+		"resource_references",
+		"StorageState = ?",
+		"run_details.UUID = resource_references.ResourceUUID",
+		"resource_references.ResourceType = ? AND resource_references.ReferenceUUID = ? AND resource_references.ReferenceType = ?")
+	_, err = tx.Exec(query, arguments...)
```
gkcalat, line 373: This could be a little more readable if we limit the scope of `UpdateWithJointOrFrom` to only `UPDATE %s INNER JOIN %s ON %s`:

```golang
query, args, err := s.db.UpdateWithJointOrFrom(
		"run_details",
		"resource_references",
                "run_details.UUID = resource_references.ResourceUUID").
                SetMap(...).
                Where(...).
                ToSql()
```

gkcalat (commented):
Thank you @difince for the contribution!
I left a few comments.

backend/src/apiserver/storage/experiment_store.go
difince: That is a great idea, but at least for now, the implementation does not look trivial to me. Squirrel heavily uses `builder` design pattern, so to be able to call `SetMap` and `Where` - IMO t I will need to extend squirrel library - by creating a new `StatementBuilder`, which still needs to build different requests based if MySQL or SQLite ... Let me think

gkcalat: That makes sense. Thank you. Perhaps another alternative is to opt for gorm v2.

backend/src/apiserver/storage/db.go
gkcalat: > Could we limit this function's scope to "UPDATE %s JOIN %s ON %s" and let users set the remaining via `squirrel.SetMap` and `squirrel.Where`? That way, this function will only implement the missing part.

Please, ignore this, according to the discussion below.

One last comment: we recently updated the [code styling requirements](https://github.com/kubeflow/pipelines/blob/master/backend/README.md#code-style). TL;DR; it is recommended to run `pre-commit` with the provided config, which includes a golang linter.

difince: Thank you for referencing this. I have just started using pre-commit.

backend/src/apiserver/storage/experiment_store.go
difince: Done - Jobs now use the new queries

backend/src/apiserver/storage/db.go
```diff
@@ -88,6 +91,11 @@ func (d MySQLDialect) IsDuplicateError(err error) bool {
 	return ok && sqlError.Number == mysqlerr.ER_DUP_ENTRY
 }
 
+// UpdateFromOrJoin TODO(gkcalat): deprecate resource_references table once we migration to v2beta1 is available.
+func (d MySQLDialect) UpdateWithJointOrFrom(targetTable, joinTable, setClause, joinClause, whereClause string) string {
+	return fmt.Sprintf("UPDATE %s JOIN %s ON %s SET %s WHERE %s", targetTable, joinTable, joinClause, setClause, whereClause)
```
gkcalat, line 96: I just realized that this actually should be `LEFT JOIN` because we are replacing the `OR` op in the `WHERE` clause with this.

difince: Yep! Thanks for catching this!

backend/src/apiserver/storage/db.go
```diff
@@ -131,6 +139,11 @@ func (d SQLiteDialect) IsDuplicateError(err error) bool {
 	return ok && sqlError.Code == sqlite3.ErrConstraint
 }
 
+// UpdateFromOrJoin TODO(gkcalat): deprecate resource_references table once we migration to v2beta1 is available.
+func (d SQLiteDialect) UpdateWithJointOrFrom(targetTable, joinTable, setClause, joinClause, whereClause string) string {
+	return fmt.Sprintf("UPDATE %s SET %s FROM %s WHERE %s AND %s", targetTable, setClause, joinTable, joinClause, whereClause)
```
gkcalat, line 144: It seems to be equivalent to `INNER JOIN`. If so, this will be different from MySQL.

backend/src/apiserver/storage/db.go
difince: Re-opening this conversation... 
I just realized that the below statement is not correct. 
> I just realized that this actually should be `LEFT JOIN` because we are replacing the `OR` op in the `WHERE` clause with this.

This is the original request:
```UPDATE run_details SET StorageState = 'ARCHIVED' WHERE StorageState <> 'ARCHIVED' AND UUID in (SELECT ResourceUUID FROM resource_references as rf WHERE (rf.ReferenceUUID = "d86e5315-09d7-4c8a-982e-4de004b2e4b6" AND rf.ResourceType = 'Run' AND rf.ReferenceType = 'Experiment')) OR ExperimentUUID = 'd86e5315-09d7-4c8a-982e-4de004b2e4b6'; ``` 
created by the following code.
``` 
	// TODO(gkcalat): deprecate resource_references table once we migration to v2beta1 is available.
	// TODO(jingzhang36): use inner join to replace nested query for better performance.
	filteredRunsSql, filteredRunsArgs, err := sq.Select("ResourceUUID").
		From("resource_references as rf").
		Where(sq.And{
			sq.Eq{"rf.ResourceType": model.RunResourceType},
			sq.Eq{"rf.ReferenceUUID": expId},
			sq.Eq{"rf.ReferenceType": model.ExperimentResourceType},
		}).ToSql()
	if err != nil {
		return util.NewInternalServerError(err,
			"Failed to create query to filter the runs in an experiment %s. error: '%v'", expId, err.Error())
	}
	updateRunsSql, updateRunsArgs, err := sq.
		Update("run_details").
		SetMap(sq.Eq{
			"StorageState": model.StorageStateArchived.ToString(),
		}).
		Where(sq.NotEq{"StorageState": model.StorageStateArchived.ToString()}).
		Where(fmt.Sprintf("UUID in (%s) OR ExperimentUUID = '%s'", filteredRunsSql, expId), filteredRunsArgs...).
		ToSql()
	if err != nil {
		return util.NewInternalServerError(err,
			"Failed to create query to archive the runs in an experiment %s. error: '%v'", expId, err.Error())
	}
```

This is the new Optimized MySQL query: 

```UPDATE run_details JOIN resource_references ON run_details.UUID = resource_references.ResourceUUID SET StorageState = "STORAGESTATE_ARCHIVED" WHERE resource_references.ResourceType = "RUN" AND resource_references.ReferenceUUID = "<experiment_id>" AND resource_references.ReferenceType = "Experiment" ```

By adding `LEFT JOIN` we won't directly achieve `OR ExperimentUUID = 'experiment_ID'` from the original query, because:
 1. In general, LEFT JOIN intends to return the entire LEFT table (we just do not want this) and 
 2. The  WHERE clause of the new query is too specific: ` WHERE resource_references.ResourceType = "RUN" AND resource_references.ReferenceUUID = "<experiment_id>" AND resource_references.ReferenceType = "Experiment"`. 
Even if we add an `OR ExperimentUUID = 'experiment_ID'` statement it won't work because of how we JOIN both Tables. Plus there is already a dedicated query that updates all `run_details` based just on `experimentUUID` see (ref A) : https://github.com/kubeflow/pipelines/blob/4f74148343397297bbb89aa7d24cedf7874fb70f/backend/src/apiserver/storage/experiment_store.go#L338-L346 

In short :), After I took a look at how Run and Jobs are created IMO, it would be a bug if we have in the database run_details record without a corresponding `resource_reference` record with `ResourceType= RUN` and `ReferenceType=Experiment`. 
So I suggest keeping the INNER JOIN version of the optimized request and even removing the  (*ref A) - IMO it has been added for "just in case"

backend/src/apiserver/storage/db.go
difince: See my above comment

backend/src/apiserver/storage/db.go
Linchin: @juliusvonkohout thank you for bringing this up. Let's keep the scope of this PR to what it is now. Would you create an issue on the listruns query speed, and if there are any other operations with similar speed issue, add them too?

Linchin:
Thank you @difince! 
/lgtm
/approve
closes #6845

difince:
Hi @Linchin I haven't removed the "LEFT" from the JOIN as I expect some comments :)
Even with the LEFT JOIN the query is still doing the right thing -  the performance is much worse. So, I will open a new PR to remove it!

backend/src/apiserver/storage/db.go
gkcalat: Thank you @difince for a thorough analysis. I agree.

gkcalat:
> Hi @Linchin I haven't removed the "LEFT" from the JOIN as I expect some comments :) Even with the LEFT JOIN the query is still doing the right thing - the performance is much worse. So, I will open a new PR to remove LEFT JOIN and instead iser INNER!

The PR was merged by accident. In the next PR, could you please replace the join with `INNER JOIN` and add a TODO with a note to switch to filtering by `experiment_id` once we deprecate `resource_references`?

Thank you!

PR #9693, opened 2023-06-29
chore(frontend): Refactor resource selector for sharable usage in both v1 and v2

jlyaoyuli:
The `error` type in v2beta1 API return values is not a `string` but `GooglerpcStatus`, which makes the `BaseResource` interface in `ResourceSelector` is not usable in v2 functionalities. `ResourceSelectorV2` was created to temporarily support v2 behavior. The main purpose for this PR is to refactor the logic of `ResourceSelector` (return **selected id** rather than **selected resource**) to make it usable for both v1 and v2.

Ref: 
https://github.com/kubeflow/pipelines/blob/master/backend/api/v2beta1/pipeline.proto#L160
https://github.com/kubeflow/pipelines/blob/master/frontend/src/pages/ResourceSelector.tsx#L35
https://github.com/kubeflow/pipelines/pull/9123


frontend/src/pages/ResourceSelector.test.tsx
```diff
@@ -177,20 +177,4 @@ describe('ResourceSelector', () => {
       RESOURCES[1].id,
     ]);
   });
-
-  it('logs error if selected resource ID is not found in list', async () => {
```
jlyaoyuli, line 181: This test is not applicable for new logic.

chensun (approved):
/lgtm
/approve

Thanks!

Issue #8074, opened 2022-07-27
[feature] Add authorization to all functions in ReportServer


difince:
### Feature Area
There are two API-server endpoints that still miss authorization - [ReportWorkflow](https://github.com/kubeflow/pipelines/blob/d48776fb2d9752ca48af14ea6256791d2b167a71/backend/src/apiserver/server/report_server.go#L32) and [ReportScheduledWorkflow](https://github.com/kubeflow/pipelines/blob/d48776fb2d9752ca48af14ea6256791d2b167a71/backend/src/apiserver/server/report_server.go#L45). This is a security issue. Each endpoint should validate that the user has permission to call them. 

**persistent-agent** service calls these endpoints. Once authorization is enabled, the persistent-agent need to authorize itself by providing user information in the request headers.

This issue is a related/ follow-up issue to PR https://github.com/kubeflow/pipelines/pull/7819

 /area backend 

### Is there a workaround currently?
no. A security issue exists in the current implementation.

---

Love this idea? Give it a 👍.

difince:
cc: @juliusvonkohout

juliusvonkohout:
For the next KFP meeting ;-)

zijianjoy:
Hello @difince , the `ReportWorkflow` and `ReportScheduledWorkflow` are used by single persistent-agent instance for monitoring the status of workflow. persistent-agent itself cannot and shouldn't authenticate as a user. 
/assign @chensun

difince:
Thank you @zijianjoy for your feedback. How services are supposed to authenticate themself?

juliusvonkohout:
> Hello @difince , the `ReportWorkflow` and `ReportScheduledWorkflow` are used by single persistent-agent instance for monitoring the status of workflow. persistent-agent itself cannot and shouldn't authenticate as a user. /assign @chensun

By default every kubeflow user can hijack them, since these endpoints are unauthenticated. So first they need ANY kind of authentication.

juliusvonkohout:
> Thank you @zijianjoy for your feedback. How services are supposed to authenticate themself then? Any suggestion? I guess this reflects on the implementation of #7819 as well ?

@chensun (@zijianjoy college at google) already approved #7819 so i do not think that it is affected. Readartifact etc. is really called by users for a proper reason. Reportworkflow might only be abused for no reason.

difince:
Fixed by #9699 
[...]

PR #9699, opened 2023-07-03
feat(backend) Enable auth between pesistence agent and pipelineAPI (ReportServer)

difince:
1. Add authentication and authorization logic to PipelineAPI PeportServer's methods - [reportworkflow](https://github.com/kubeflow/pipelines/blob/a634eef3ec541ee64eb0220d5db12b82f682479e/backend/src/apiserver/server/report_server.go#L47C24-L47C38) & [reportSheduledWorkflow](https://github.com/kubeflow/pipelines/blob/a634eef3ec541ee64eb0220d5db12b82f682479e/backend/src/apiserver/server/report_server.go#L78C24-L78C47) and
2. Force Persistence Agent to authenticate itself through Service Account Token Volume Projection when calling the above methods.

Issue: https://github.com/kubeflow/pipelines/issues/8074

**Checklist:**

### DIFF
--- a/backend/src/agent/persistence/client/pipeline_client.go
+++ b/backend/src/agent/persistence/client/pipeline_client.go
@@ -18,6 +18,7 @@ import (
 	"context"
 	"fmt"
 	"os"
+	"strings"
 	"time"
 
 	"github.com/kubeflow/pipelines/backend/src/apiserver/common"
@@ -46,11 +47,13 @@ type PipelineClient struct {
 	timeout             time.Duration
 	reportServiceClient api.ReportServiceClient
 	runServiceClient    api.RunServiceClient
+	tokenRefresher      TokenRefresherInterface
 }
 
 func NewPipelineClient(
 	initializeTimeout time.Duration,
 	timeout time.Duration,
+	tokenRefresher TokenRefresherInterface,
 	basePath string,
 	mlPipelineServiceName string,
 	mlPipelineServiceHttpPort string,
@@ -71,13 +74,19 @@ func NewPipelineClient(
 	return &PipelineClient{
 		initializeTimeout:   initializeTimeout,
 		timeout:             timeout,
+		tokenRefresher:      tokenRefresher,
 		reportServiceClient: api.NewReportServiceClient(connection),
 		runServiceClient:    api.NewRunServiceClient(connection),
 	}, nil
 }
 
 func (p *PipelineClient) ReportWorkflow(workflow util.ExecutionSpec) error {
-	ctx, cancel := context.WithTimeout(context.Background(), time.Minute)
+	pctx := context.Background()
+	if os.Getenv("MULTIUSER") == "True" {
+		pctx = metadata.AppendToOutgoingContext(pctx, "Authorization",
+			"Bearer "+p.tokenRefresher.GetToken())
+	}
+	ctx, cancel := context.WithTimeout(pctx, time.Minute)
 	defer cancel()
 
 	_, err := p.reportServiceClient.ReportWorkflowV1(ctx, &api.ReportWorkflowRequest{
@@ -91,6 +100,15 @@ func (p *PipelineClient) ReportWorkflow(workflow util.ExecutionSpec) error {
 			// * there is something wrong with the workflow
 			// * the workflow has been deleted by someone else
 			return util.NewCustomError(err, util.CUSTOM_CODE_PERMANENT,
+				"Error while reporting workflow resource (code: %v, message: %v): %v, %+v",
+				statusCode.Code(),
+				statusCode.Message(),
+				err.Error(),
+				workflow.ToStringForStore())
+		} else if statusCode.Code() == codes.Unauthenticated && strings.Contains(err.Error(), "service account token has expired") {
+			//if unauthenticated because SA token is expired, re-read/refresh the token and try again
+			p.tokenRefresher.RefreshToken()
+			return util.NewCustomError(err, util.CUSTOM_CODE_TRANSIENT,
 				"Error while reporting workflow resource (code: %v, message: %v): %v, %+v",
 				statusCode.Code(),
 				statusCode.Message(),
@@ -110,7 +128,12 @@ func (p *PipelineClient) ReportWorkflow(workflow util.ExecutionSpec) error {
 }
 
 func (p *PipelineClient) ReportScheduledWorkflow(swf *util.ScheduledWorkflow) error {
-	ctx, cancel := context.WithTimeout(context.Background(), time.Minute)
+	pctx := context.Background()
+	if os.Getenv("MULTIUSER") == "True" {
+		pctx = metadata.AppendToOutgoingContext(pctx, "Authorization",
+			"Bearer "+p.tokenRefresher.GetToken())
+	}
+	ctx, cancel := context.WithTimeout(pctx, time.Minute)
 	defer cancel()
 
 	_, err := p.reportServiceClient.ReportScheduledWorkflowV1(ctx,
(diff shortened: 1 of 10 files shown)


backend/src/agent/persistence/client/pipeline_client.go
```diff
@@ -96,6 +105,15 @@ func (p *PipelineClient) ReportWorkflow(workflow util.ExecutionSpec) error {
 				statusCode.Message(),
 				err.Error(),
 				workflow.ToStringForStore())
+		} else if statusCode.Code() == codes.Unauthenticated && strings.Contains(err.Error(), "service account token has expired") {
+			//if unauthenticated because SA token is expired, re-read/refresh the token and try again
+			p.tokenRefresher.RefreshToken()
+			return util.NewCustomError(err, util.CUSTOM_CODE_TRANSIENT,
```
chensun, line 111: Although I see you're following an existing pattern, for my education how would user discover such error? Would this be surfaced back by apiserver?

backend/src/agent/persistence/client/token_refresher.go
```diff
@@ -0,0 +1,77 @@
...
+	"sync"
+	"time"
+)
+
+type TokenRefresherInterface interface {
+	GetToken() string
+	RefreshToken()
+}
+
+const SA_TOKEN_FILE = "/var/run/secrets/kubeflow/tokens/persistenceagent-sa-token"
```
chensun, line 15: There's a single instance of persistence agent. How would this token file work? Won't multi users share the same token file?

difince: Thank you for the review! 

Currently, every pod has an occasioned Service Account and SA token dynamically loaded at `/var/run/secrets/kubernetes.io/serviceaccount/token`.This token has a target audience set to - used to authenticate against the Kube API server -  "https://kubernetes.default.svc.cluster.local"
```
{
  "aud": [
    "https://kubernetes.default.svc.cluster.local"
  ],
  "exp": 1720267473,
  "iat": 1688731473,
  "iss": "https://kubernetes.default.svc.cluster.local",
  "kubernetes.io": {
    "namespace": "kubeflow",
    "pod": {
      "name": "ml-pipeline-persistenceagent-5f9978c49d-npqkn",
      "uid": "2b132bfc-bd86-460d-95ca-00b8c55d7949"
    },
    "serviceaccount": {
      "name": "ml-pipeline-persistenceagent",
      "uid": "d64595e1-53a6-4ada-be90-10398316ba38"
    },
    "warnafter": 1688735080
  },
  "nbf": 1688731473,
  "sub": "system:serviceaccount:kubeflow:ml-pipeline-persistenceagent"
}
```
The new SA Token introduced in this PR and made available to persistenceagent pod via the Service Account Token Volume Projection is time-limited, audience-bound token. Both tokens has the same identity (peristenceagent-pod) but have different audience group. Kubelet takes care to rotate the new token. If the pod recreated , the new token will be recreated too. This makes the need of having "refresh token " logic.  If we decrypt it it looks like this:  
```
{
  "aud": [
    "pipelines.kubeflow.org"
  ],
  "exp": 1688734989,
  "iat": 1688731389,
  "iss": "https://kubernetes.default.svc.cluster.local",
  "kubernetes.io": {
    "namespace": "kubeflow",
    "pod": {
      "name": "ml-pipeline-persistenceagent-5f9978c49d-npqkn",
      "uid": "2b132bfc-bd86-460d-95ca-00b8c55d7949"
    },
    "serviceaccount": {
      "name": "ml-pipeline-persistenceagent",
      "uid": "d64595e1-53a6-4ada-be90-10398316ba38"
    }
  },
  "nbf": 1688731389,
  "sub": "system:serviceaccount:kubeflow:ml-pipeline-persistenceagent"
}
  ```
  Your question a valid. If only persistence-agent should be able to call PipelineAPI's report (scheduled) workflows endpoints and it is just a single service..   is there a need to care about MULTIUSER mode here?

File: manifests/kustomize/base/pipeline/ml-pipeline-persistenceagent-deployment.yaml [hunk 3]
```diff
@@ -36,4 +36,15 @@ spec:
...
+            name: persistenceagent-sa-token
       serviceAccountName: ml-pipeline-persistenceagent
+      volumes:
+        - name: persistenceagent-sa-token
+          projected:
+            sources:
+              - serviceAccountToken:
+                  path: persistenceagent-sa-token
+                  expirationSeconds: 3600
+                  audience: pipelines.kubeflow.org
```
difince, line 50: This is the place instructing Kubernetes to create a new token with a specific audience, expirity time & the path where the token to be loaded

backend/src/apiserver/server/report_server.go
```diff
@@ -136,6 +158,20 @@ func validateReportScheduledWorkflowRequest(swfManifest string) (*util.Scheduled
 	return swf, nil
 }
 
+func (s *ReportServer) canAccessWorkflow(ctx context.Context, executionName string, resourceAttributes *authorizationv1.ResourceAttributes) error {
+	if !common.IsMultiUserMode() {
```
difince, line 162: Having in mind the above comment - IMO, there is no need to do this check `IsMultiUser`

backend/src/apiserver/server/report_server.go
```diff
@@ -136,6 +158,20 @@ func validateReportScheduledWorkflowRequest(swfManifest string) (*util.Scheduled
...
+		// Skip authz if not multi-user mode.
+		return nil
+	}
+	resourceAttributes.Group = common.RbacPipelinesGroup
+	resourceAttributes.Version = common.RbacPipelinesVersion
+	err := s.resourceManager.IsAuthorized(ctx, resourceAttributes)
```
difince, line 168: Here I do not care about the "namespace"

backend/src/agent/persistence/client/pipeline_client.go
difince: Hi! Sorry for my late response. 
The user could observe the status of the Run and inspect the logs of the persistent agent. The errors are not directly displayed on the UI. 
If a "transient" error occurs the Persistent Agent will retry with an exponential delay

backend/src/agent/persistence/client/token_refresher.go
chensun: Thank you for such detailed reply. 

> With my current implementation, only the Persistence-agent will be able to call these two KF pipeline endpoints. No one else has permission to do that.

> only with these two KFP endpoints there is no need to care about the MULTIUSER mode value - instead, we always should validate that persistent-agent makes the requests. The real user and the namespace are validated when the run/jobs are created.

This sounds good to me!

chensun: Other than this, is there a way we can add some unit tests to the change? Otherwise, we could accidentally break it without any alert.

juliusvonkohout:
@difince is any help needed?

backend/src/agent/persistence/client/pipeline_client.go
difince: Hi @chensun 
I have added unit tests for the TokenRefresher. For the need of the unit tests I created the `FileReader interface`.

backend/src/agent/persistence/client/token_refresher.go
```diff
@@ -0,0 +1,78 @@
...
+
+func (tr *tokenRefresher) StartTokenRefreshTicker() error {
+	err := tr.RefreshToken()
+	if err != nil {
+		return err
+	}
+
+	ticker := time.NewTicker(*tr.seconds)
+	go func() {
+		for range ticker.C {
```
difince, line 55: Another thing, I was unsure if I should use the `stopCh`, so I removed it. WDYT?  Please inspect the changes introduced by my third [commit](https://github.com/kubeflow/pipelines/pull/9699/commits/bc9e1e05d35127a0ce8e5fd1e92a3b5e85ca9ac6).

backend/src/agent/persistence/client/pipeline_client.go
difince: Here, If an error occurs while refreshing/reading the SA token I just log the error and let the worker thread retry. Any idea how to handle this gracefully?

backend/src/agent/persistence/client/token_refresher.go
difince: Unit tests added

chensun (approved):
/lgtm
/approve

Thanks!

juliusvonkohout:
@difince maybe you have to rebase to https://github.com/kubeflow/pipelines/pull/9885 if the test fails again

PR #9712, opened 2023-07-06
chore(frontend): Remove unnecessary getVersionTemplate() call in PipelineDetails

jlyaoyuli:
For v1 pipeline, we can get its template from `pipeline_spec` in the response of v2 getPipelineVersion() call. It means that the getVersionTemplate() call is not necessary. Thus, remove it to reduce the time cost.

### DIFF
--- a/frontend/src/pages/PipelineDetails.test.tsx
+++ b/frontend/src/pages/PipelineDetails.test.tsx
@@ -53,10 +53,6 @@ describe('PipelineDetails', () => {
   const getV2RecurringRunSpy = jest.spyOn(Apis.recurringRunServiceApi, 'getRecurringRun');
   const getExperimentSpy = jest.spyOn(Apis.experimentServiceApiV2, 'getExperiment');
   const deletePipelineVersionSpy = jest.spyOn(Apis.pipelineServiceApiV2, 'deletePipelineVersion');
-  const getPipelineVersionTemplateSpy = jest.spyOn(
-    Apis.pipelineServiceApi,
-    'getPipelineVersionTemplate',
-  );
   const createGraphSpy = jest.spyOn(StaticGraphParser, 'createGraph');
 
   let tree: ShallowWrapper | ReactWrapper;
@@ -198,10 +194,6 @@ describe('PipelineDetails', () => {
         display_name: 'test experiment',
       } as V2beta1Experiment),
     );
-    // getTemplateSpy.mockImplementation(() => Promise.resolve({ template: 'test template' }));
-    getPipelineVersionTemplateSpy.mockImplementation(() =>
-      Promise.resolve({ template: 'test template' }),
-    );
     createGraphSpy.mockImplementation(() => new graphlib.Graph());
   });
 
@@ -214,7 +206,6 @@ describe('PipelineDetails', () => {
 
   it('shows pipeline name in page name, and breadcrumb to go back to pipelines', async () => {
     tree = shallow(<PipelineDetails {...generateProps()} />);
-    await getPipelineVersionTemplateSpy;
     await TestUtils.flushPromises();
     expect(updateToolbarSpy).toHaveBeenLastCalledWith(
       expect.objectContaining({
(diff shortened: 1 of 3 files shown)


frontend/src/pages/PipelineDetails.test.tsx
```diff
@@ -507,20 +514,6 @@ describe('PipelineDetails', () => {
     },
   );
 
-  it('uses an empty string and does not show error when getTemplate response is empty', async () => {
```
zijianjoy, line 510: Can we change the implementation of this test instead of removing it? We want to validate the behavior when `Apis.pipelineServiceApiV2.getPipelineVersion` doesn't return a valid pipeline_spec. (for example: pipeline_spec is empty)

frontend/src/pages/PipelineDetails.test.tsx
```diff
@@ -536,34 +529,51 @@ describe('PipelineDetails', () => {
...
-    await getV1PipelineSpy;
-    await TestUtils.flushPromises();
-    expect(updateBannerSpy).toHaveBeenCalledTimes(2); // Once to clear banner, once to show error
+  it('shows load error banner when failing to get pipeline version', async () => {
+    TestUtils.makeErrorResponse(getV2PipelineVersionSpy, 'No pipeline version is found');
+    render(<PipelineDetails {...generateProps()} />);
+
+    await waitFor(() => {
+      // one for selected Version, another for template string
+      expect(getV2PipelineVersionSpy).toHaveBeenCalledTimes(2);
```
zijianjoy, line 538: Can we improve this by only querying the same API once? It feels strange to me that `updateBanner` will be called twice in the following validation. It is probably a bug that needs fixing because banner has been updated twice.

frontend/src/pages/PipelineDetails.test.tsx
```diff
@@ -550,6 +550,69 @@ describe('PipelineDetails', () => {
...
+      render(<PipelineDetails {...generateProps()} />);
+
+      await waitFor(() => {
+        expect(getV2PipelineVersionSpy).toHaveBeenCalled();
+        // empty template string from empty pipeline_spec and it won't call createGraph()
+        expect(createGraphSpy).toHaveBeenCalledTimes(0);
+      });
+
+      // No errors
+      expect(updateBannerSpy).toHaveBeenCalledTimes(1); // Once to clear banner
```
zijianjoy, line 578: I think an `pipeline_spec = undefined` is an error situation, how do we communicate with users that we have received no pipeline_spec? Silently showing no graph seems a lack of guidance.

frontend/src/pages/PipelineDetails.test.tsx
```diff
@@ -550,6 +550,69 @@ describe('PipelineDetails', () => {
     );
   });
 
+  it(
+    'uses an empty string and does not show error ' +
+      'when pipeline_spec in the response of getPipelineVersion() is undefined',
```
zijianjoy, line 555: Probably we should call this `shows error when pipeline_spec in the response of getPipelineVersion() is undefined`

frontend/src/pages/PipelineDetails.test.tsx
```diff
@@ -550,6 +550,69 @@ describe('PipelineDetails', () => {
...
+        pipeline_version_id: 'test-pipeline-version-id',
+        pipeline_spec: {}, // invalid pipeline_spec
+      });
+      render(<PipelineDetails {...generateProps()} />);
+
+      await waitFor(() => {
+        expect(getV2PipelineVersionSpy).toHaveBeenCalled();
+      });
+
+      expect(updateBannerSpy).toHaveBeenCalledTimes(2); // Once to clear banner, once to show error
```
zijianjoy, line 605: I thought that there are 3 calls in total before another PR of your fix. Is the issue already fixed?

frontend/src/pages/PipelineDetails.test.tsx
jlyaoyuli: Add log error for debug

zijianjoy:
/lgtm
/approve

Thank you Joe for the improvement!

PR #9728, opened 2023-07-12
chore(frontend): Avoid redundant v2 API call in PipelineDetails.

jlyaoyuli:
Part of #9731 
Previously, we update the URL with `pipelineVersionId` obtained in getSelectedVersion(), and it will cause to re-render the page, which brings out several redundant API calls. The PR is removing the URL update logic to avoid those calls. Also, the getPipelineVersion() in _getTemplateString() is not necessary either, so it's removed as well.

### DIFF
--- a/frontend/src/pages/PipelineDetails.test.tsx
+++ b/frontend/src/pages/PipelineDetails.test.tsx
@@ -54,6 +54,7 @@ describe('PipelineDetails', () => {
   const getExperimentSpy = jest.spyOn(Apis.experimentServiceApiV2, 'getExperiment');
   const deletePipelineVersionSpy = jest.spyOn(Apis.pipelineServiceApiV2, 'deletePipelineVersion');
   const createGraphSpy = jest.spyOn(StaticGraphParser, 'createGraph');
+  const PIPELINE_VERSION_ID = 'test-pipeline-version-id';
 
   let tree: ShallowWrapper | ReactWrapper;
   let testV1Pipeline: ApiPipeline = {};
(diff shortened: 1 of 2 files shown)


frontend/src/pages/PipelineDetails.tsx
```diff
@@ -296,21 +296,18 @@ class PipelineDetails extends Page<{}, PipelineDetailsState> {
...
         if (listVersionsResponse.pipeline_versions) {
-          latesetVersion = listVersionsResponse.pipeline_versions[0];
-          // Append version id to URL for create run (new run switcher call getPipelineVersion)
-          this.props.history.replace({
-            pathname: `/pipelines/details/${pipelineId}/version/${latesetVersion.pipeline_version_id}`,
-          });
-          return latesetVersion;
+          selectedVersion = listVersionsResponse.pipeline_versions[0];
+        } else {
+          return;
```
zijianjoy, line 301: Can we `return undefined`  in this case?

frontend/src/pages/PipelineDetails.tsx
```diff
@@ -296,21 +296,18 @@ class PipelineDetails extends Page<{}, PipelineDetailsState> {
...
         }
-        return undefined;
       } catch (err) {
         this.setStateSafe({ graphIsLoading: false });
         await this.showPageError('Cannot retrieve pipeline version list.', err);
         logger.error('Cannot retrieve pipeline version list.', err);
         return;
```
zijianjoy, line 307: Can we `return undefined` in this case?

frontend/src/pages/PipelineDetails.tsx
```diff
@@ -296,21 +296,18 @@ class PipelineDetails extends Page<{}, PipelineDetailsState> {
         );
 
         if (listVersionsResponse.pipeline_versions) {
```
zijianjoy, line 298: We should probably also check that `len(listVersionsResponse.pipeline_versions) > 0`

frontend/src/pages/PipelineDetails.tsx
```diff
@@ -523,10 +520,7 @@ class PipelineDetails extends Page<{}, PipelineDetailsState> {
           logger.error('Cannot retrieve pipeline versions.', err);
           return;
         }
-        templateString = await this._getTemplateString(
-          pipelineId,
-          v2SelectedVersion ? v2SelectedVersion.pipeline_version_id! : v1SelectedVersion?.id!,
-        );
```
zijianjoy, line 529: Can we remove `v1SelectedVersion` as well? Just curious if we can rely on v2 API as much as we can.

frontend/src/pages/PipelineDetails.test.tsx
```diff
@@ -535,12 +535,12 @@ describe('PipelineDetails', () => {
 
     await waitFor(() => {
       // one for selected Version, another for template string
```
zijianjoy, line 537: This comment needs update as well.

frontend/src/pages/PipelineDetails.tsx
```diff
@@ -296,21 +296,18 @@ class PipelineDetails extends Page<{}, PipelineDetailsState> {
         );
 
         if (listVersionsResponse.pipeline_versions) {
-          latesetVersion = listVersionsResponse.pipeline_versions[0];
-          // Append version id to URL for create run (new run switcher call getPipelineVersion)
-          this.props.history.replace({
-            pathname: `/pipelines/details/${pipelineId}/version/${latesetVersion.pipeline_version_id}`,
-          });
-          return latesetVersion;
+          selectedVersion = listVersionsResponse.pipeline_versions[0];
```
zijianjoy, line 299: Can you add a test to verify that: ` If there are more than 1 pipeline version, the first version is being used`.  The way to verify that the first version is being used is to check the API call to getPipelineVersion is using the first version ID in this request.

frontend/src/pages/PipelineDetails.tsx
jlyaoyuli: Plan to do it in next PR if it's workable : )

frontend/src/pages/PipelineDetails.test.tsx
zijianjoy: The comment here needs to be updated.

zijianjoy:
/lgtm

Only one minor comment, thank you Joe!

Issue #9746, opened 2023-07-18
[frontend] artifacts [unknown] pipeline/workspace


axel7083:
### Environment

*  How did you deploy Kubeflow Pipelines (KFP)? manifests
*  KFP version: 2.0.0

### Description

No matter how I start my run, (from UI or SDK) I always get the `[unknown]` in the Artifact panel for pipeline/workspace.

![image](https://github.com/kubeflow/pipelines/assets/42176370/305c20ac-5bd2-488d-9081-99d541f6d7d6)

### Expected result

The name of the pipeline.

---

Impacted by this bug? Give it a 👍.

khituras:
I saw your PR for that. Did you find any workaround in the meantime?

axel7083:
> I saw your PR for that. Did you find any workaround in the meantime?

Sadly, no, looking for the pipeline name was kinda hard as explained in the PR, maybe a metadata was not added properly or a CustomProperties. But from the current element, I inspected the networks logs in the current version, and there is not the pipeline name attached.

AndersBennedsgaard:
@HumairAK or @axel7083 would you mind re-opening this issue? I also see artifacts as `[unknown]`, so I don't think it has been fixed

thesuperzapper:
I can confirm that this issue affects Kubeflow Pipelines 2.1.0.

rimolive:
I tested the code changes in https://github.com/kubeflow/pipelines/pull/9772, but we should definitely think of a proposal that involves backend changes too. The code in that PR makes frontend response times go crazy, specially the grouped list, which took me ~19 minutes to render with the PR change.

HumairAK:
AFAIK this is not resolved it

/reopen

rvolchek:
🏁 flagged for review

PR #9772, opened 2023-07-21
feat(frontend): Adding logic to fetch pipeline name in ArtifactsList

axel7083:
Fixes https://github.com/kubeflow/pipelines/issues/9746

Here is a small fix for removing the `getResourcePropertyViaFallBack` inside `pages/ArtifactList` responsible of getting the pipeline/workspace name from Artifact.

From what I can see when using V2, nor the Artifact and execution hold the Pipeline name, or workspaces.

To work around, I created a function taking called `getPipelineNameByArtifact`, the only place I could find the pipeline name was in the Context, to get the context I had to:
- Get Events By Artifact
- Take the execution id in the event
- GetContextByExecutionID
- Getting pipeline-name from context name.

### DIFF
--- a/frontend/src/mlmd/MlmdUtils.ts
+++ b/frontend/src/mlmd/MlmdUtils.ts
@@ -252,12 +252,7 @@ export async function getContextByExecution(
   return result[0];
 }
 
-export async function getContextsByExecution(execution: Execution): Promise<Context[]> {
-  const executionId = execution.getId();
-  if (!executionId) {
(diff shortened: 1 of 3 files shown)


khituras:
Could someone look at this? It's unfortunate that all the artifacts are put into the `[unknown]` category.

zijianjoy:
/ok-to-test

+ @jlyaoyuli Hello Joe, would you like to take a look at this Artifact list PR?

PR #9798, opened 2023-07-28
feat(backend): add postgres initialization

Linchin:
Part 1 of supporting PostgreSQL:

* Adds driver for PostgreSQL
* Supports initializing PostgreSQL
* Adds integration tests for MySQL or PostgreSQL initialization
* Moves `client_manager.go` to its own package and folder for ease of testing
* Allows users to select which type of database they want to run integration test against
* Fixes a typo in the PostgreSQL deployment guide
* Makes PostgreSQL no longer load balancer

**Checklist:**


Linchin:
test kubeflow-pipeline-e2e-test

backend/test/integration/run_tests_locally.sh
```diff
@@ -32,7 +32,22 @@ case "$response" in
         ;;
 esac
 
-echo "Starting integration tests..."
-command="go test -v ./... -namespace ${NAMESPACE} -args -runIntegrationTests=true -isDevMode=true"
-echo $command "$@"
-$command "$@"
+if [ "$1" == "postgres" ]; then
+    echo "Starting PostgreSQL DB port forwarding..."
+    kubectl -n "$NAMESPACE" port-forward svc/postgres-service 5432:5432 --address="127.0.0.3" & PORT_FORWARD_PID=$!
```
zijianjoy, line 37: What is this part used for ? `PORT_FORWARD_PID=$!`

backend/test/integration/run_tests_locally.sh
```diff
@@ -32,7 +32,22 @@ case "$response" in
...
+    $command "$@"
+else 
+    echo "Starting MySQL DB port forwarding..."
+    kubectl -n "$NAMESPACE" port-forward svc/mysql 3306:3306 --address=localhost & PORT_FORWARD_PID=$!
+    # wait for kubectl port forward
+    sleep 10
+    echo "Starting integration tests..."
+    command="go test -v ./... -namespace ${NAMESPACE} -args -runIntegrationTests=true -isDevMode=true"
+    echo $command "$@"
+    $command "$@"
```
zijianjoy, line 52: NIT: How about using `set -x` at the top of file to print command automatically, instead of echoing here?

backend/test/integration/run_tests_locally.sh
```diff
@@ -32,7 +32,22 @@ case "$response" in
...
-command="go test -v ./... -namespace ${NAMESPACE} -args -runIntegrationTests=true -isDevMode=true"
-echo $command "$@"
-$command "$@"
+if [ "$1" == "postgres" ]; then
+    echo "Starting PostgreSQL DB port forwarding..."
+    kubectl -n "$NAMESPACE" port-forward svc/postgres-service 5432:5432 --address="127.0.0.3" & PORT_FORWARD_PID=$!
+    # wait for kubectl port forward
+    sleep 10
+    echo "Starting integration tests..."
+    command="go test -v ./... -namespace ${NAMESPACE} -args -runIntegrationTests=true -isDevMode=true -runPostgreSQLTests=true"
```
zijianjoy, line 41: Looks like only this line is different between `if` and `else` condition, can we append the `-runPostgreSQLTests=true` to the command if this is postgresql, so we can reduce some duplication.

backend/src/apiserver/client_manager/client_manager.go
```diff
@@ -35,25 +37,33 @@ import (
...
+	pipelineBucketName = "MINIO_PIPELINE_BUCKET_NAME"
+	pipelinePath       = "MINIO_PIPELINE_PATH"
+
+	mysqlServiceHost       = "DBConfig.MySQLConfig.Host"
+	mysqlServicePort       = "DBConfig.MySQLConfig.Port"
+	mysqlUser              = "DBConfig.MySQLConfig.User"
+	mysqlPassword          = "DBConfig.MySQLConfig.Password"
+	mysqlDBName            = "DBConfig.MySQLConfig.DBName"
+	mysqlGroupConcatMaxLen = "DBConfig.MySQLConfig.GroupConcatMaxLen"
+	mysqlExtraParams       = "DBConfig.MySQLConfig.ExtraParams"
```
zijianjoy, line 53: It might cause a breaking change to our users for moving the db config to MySQLConfig. @chensun is it okay to make this change?

backend/src/apiserver/client_manager/client_manager.go
```diff
@@ -208,16 +218,12 @@ func (c *ClientManager) Close() {
...
-	switch driverName {
-	case "mysql":
-		arg = initMysql(driverName, initConnectionTimeout)
-	default:
-		glog.Fatalf("Driver %v is not supported", driverName)
-	}
+func InitDBClient(initConnectionTimeout time.Duration) *storage.DB {
+	// Allowed driverName values:
+	// 1) To use MySQL, use `mysql`
+	// 2) To use PostgreSQL, use `pgx`
```
zijianjoy, line 224: Why do we choose `pgx` for postgresql in `DBDriverName` flag? Is there an alternative which is easier for people to understand?

backend/src/apiserver/client_manager/client_manager.go
```diff
@@ -250,8 +256,18 @@ func initDBClient(initConnectionTimeout time.Duration) *storage.DB {
...
+	if ignoreAlreadyExistError(response.Error) != nil {
+		glog.Fatalf("Failed to initialize the databases. Error: %s", response.Error)
+	}
+
+	var textFormat string
+	switch driverName {
+	case "mysql":
+		textFormat = "longtext not null"
+	case "pgx":
+		textFormat = "text"
```
zijianjoy, line 268: We might need to define some `const` for value like `text` and `longtext not null`.

backend/src/apiserver/client_manager/client_manager.go
```diff
@@ -310,52 +347,74 @@ func initDBClient(initConnectionTimeout time.Duration) *storage.DB {
...
+			common.GetStringConfigWithDefault(postgresPassword, ""),
+			common.GetStringConfigWithDefault(postgresHost, "postgresql"),
+			"postgres",
+			uint16(common.GetIntConfigWithDefault(postgresPort, 5432)),
+		)
+		dbName = common.GetStringConfig(postgresDBName)
+	default:
+		glog.Fatalf("Driver %v is not supported, use \"mysql\" for MySQL, or \"pgx\" for PostgreSQL", driverName)
+	}
+	glog.Infof("lingqinggan sqlConfig: %s\n", sqlConfig)
```
zijianjoy, line 405: Removing your name from logging? :)

backend/src/apiserver/client_manager/client_manager.go
```diff
@@ -310,52 +347,74 @@ func initDBClient(initConnectionTimeout time.Duration) *storage.DB {
...
+			"",
+			common.GetStringConfigWithDefault(mysqlGroupConcatMaxLen, "1024"),
+			common.GetMapConfig(mysqlExtraParams),
+		)
+		sqlConfig = mysqlConfig.FormatDSN()
+		dbName = common.GetStringConfig(mysqlDBName)
+	case "pgx":
+		sqlConfig = client.CreatePostgreSQLConfig(
+			common.GetStringConfigWithDefault(postgresUser, ""),
+			common.GetStringConfigWithDefault(postgresPassword, ""),
```
zijianjoy, line 396: Do we want to set a default user like mysql does?

backend/src/apiserver/client_manager/client_manager.go
```diff
@@ -264,40 +280,61 @@ func initDBClient(initConnectionTimeout time.Duration) *storage.DB {
...
-	response = db.Model(&model.PipelineVersion{}).
-		AddForeignKey("PipelineId", "pipelines(UUID)", "CASCADE" /* onDelete */, "CASCADE" /* update */)
-	if response.Error != nil {
-		glog.Fatalf("Failed to create a foreign key for PipelineId in pipeline_versions table. Error: %s", response.Error)
-	}
-	response = db.Model(&model.Task{}).
-		AddForeignKey("RunUUID", "run_details(UUID)", "CASCADE" /* onDelete */, "CASCADE" /* update */)
-	if response.Error != nil {
-		glog.Fatalf("Failed to create a foreign key for RunUUID in task table. Error: %s", response.Error)
+	switch driverName {
```
zijianjoy, line 303: I think extracting the different behavior of each driver to their own module is helpful. But feel free to defer it to future PRs.

backend/Dockerfile.persistenceagent
```diff
@@ -27,9 +27,13 @@ RUN ./hack/install-go-licenses.sh
 # First, make sure there's no forbidden license.
 RUN go-licenses check ./backend/src/agent/persistence
 RUN go-licenses csv ./backend/src/agent/persistence > /tmp/licenses.csv && \
-    diff /tmp/licenses.csv backend/third_party_licenses/persistence_agent.csv && \
     go-licenses save ./backend/src/agent/persistence --save_path /tmp/NOTICES
 
+# RUN go-licenses csv ./backend/src/agent/persistence > /tmp/licenses.csv && \
+#    diff /tmp/licenses.csv backend/third_party_licenses/persistence_agent.csv && \
+#    go-licenses save ./backend/src/agent/persistence --save_path /tmp/NOTICES
```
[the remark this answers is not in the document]
Linchin: Thank you for pointing this out. I was temporarily disabling them, because the generated license files apparently changed although I didn't change this part of code. I intend to double check if we need to update all the license files.

backend/src/apiserver/client/sql.go
```diff
@@ -15,13 +15,14 @@
...
 import (
+	"bytes"
 	"fmt"
 
 	"github.com/go-sql-driver/mysql"
 )
 
-func CreateMySQLConfig(user, password string, mysqlServiceHost string,
-	mysqlServicePort string, dbName string, mysqlGroupConcatMaxLen string, mysqlExtraParams map[string]string,
+func CreateMySQLConfig(user, password, mysqlServiceHost, mysqlServicePort,
```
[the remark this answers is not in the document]
Linchin: I don't have a strong preference, I'm doing this solely for style purpose. Since all these parameters are of string type, it feels more succinct to declare their type together.

backend/src/apiserver/client/sql.go
```diff
@@ -44,3 +45,26 @@ func CreateMySQLConfig(user, password string, mysqlServiceHost string,
 		AllowNativePasswords: true,
 	}
 }
+
+func CreatePostgreSQLConfig(user, password, postgresHost, dbName string, postgresPort uint16,
```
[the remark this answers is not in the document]
Linchin: I think it would be great to improve security, but I wonder if it is necessary, since the DB and the backend are hosted in the same cluster.

backend/test/integration/run_tests_locally.sh
Linchin: Thank you for catching this! I should have added the function that cleans up the process on exit.

backend/src/apiserver/client_manager/client_manager.go
chensun: Seems to be fine, as I don't think `backend/src/apiserver/config/config.json` is user-facing?

backend/src/apiserver/client_manager/client_manager.go
```diff
@@ -35,25 +37,33 @@ import (
...
+	mysqlPassword          = "DBConfig.MySQLConfig.Password"
+	mysqlDBName            = "DBConfig.MySQLConfig.DBName"
+	mysqlGroupConcatMaxLen = "DBConfig.MySQLConfig.GroupConcatMaxLen"
+	mysqlExtraParams       = "DBConfig.MySQLConfig.ExtraParams"
+
+	postgresHost     = "DBConfig.PostgreSQLConfig.Host"
+	postgresPort     = "DBConfig.PostgreSQLConfig.Port"
+	postgresUser     = "DBConfig.PostgreSQLConfig.User"
+	postgresPassword = "DBConfig.PostgreSQLConfig.Password"
+	postgresDBName   = "DBConfig.PostgreSQLConfig.DBName"
```
chensun, line 59: The newly added Postgres configs seems so generic and completely covered by existing configs for mysql already. Do we need to separate them?

backend/src/apiserver/client_manager/client_manager.go
```diff
@@ -12,15 +12,17 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-package main
+package clientmanager
```
chensun, line 15: What's the reason for moving this file?

backend/src/apiserver/client_manager/client_manager.go
chensun: +1, `pgx` feels weird, albeit I saw that's the name of the go package you're using. I think `postgres` would be still better here.

backend/src/apiserver/client_manager/client_manager.go
```diff
@@ -506,3 +584,12 @@ func backfillExperimentIDToRunTable(db *gorm.DB) error {
 	`)
 	return err
 }
+
+// Returns the same error, if it's not "already exists" related.
+// Otherwise, return nil.
+func ignoreAlreadyExistError(err error) error {
+	if err != nil && strings.Contains(err.Error(), "exists") {
```
chensun, line 591: This method seems very fragile. I imagine there's could be multiple scenarios matching keyword "exists".
By the way, why do we need to ignore this now? Was it an issue for MySQL in the past or does only applies to Postgres?

backend/src/apiserver/client_manager/client_manager.go
```diff
@@ -376,14 +434,34 @@ func initMysql(driverName string, initConnectionTimeout time.Duration) string {
 	b.MaxElapsedTime = initConnectionTimeout
 	err = backoff.Retry(operation, b)
 
+	if err != nil {
+		err = fmt.Errorf("lingqinggan sqlConfig: %s\n error message: %w", sqlConfig, err)
```
chensun, line 438: nit: fix error message.

backend/test/integration/README.md
```diff
@@ -5,7 +5,22 @@
...
+The default integration test will test the default Database, MySQL.
+
 1. Configure kubectl to connect to your kfp cluster.
 2. Run the following for all integration tests: `NAMESPACE=<kfp-namespace> ./run_tests_locally.sh`.
 3. Or run the following to select certain tests: `NAMESPACE=<kfp-namespace> ./run_tests_locally.sh -testify.m Job`.
    Reference: https://stackoverflow.com/a/43312451
+
+### Run database tests with PostgreSQL
+
+To run this test, you need to first deploy the PostgreSQL images on your Kubernetes cluster. For how to deploy,
```
chensun, line 17: Is it in your next step to automate the test? We can rely on manual test.

backend/src/apiserver/client_manager/client_manager.go
zijianjoy: Sounds good, thanks for confirming! We are still keeping the user-facing flag unchanged: https://github.com/kubeflow/pipelines/blob/master/backend/src/cache/client_manager.go#L112-L115

backend/src/apiserver/client_manager/client_manager.go
Linchin: I added this because MySQL simply ignores if the foreign key already exist, but PostgreSQL reports an error. A more robust way I can think of is to use an if/else branch for the two SQLs and handle them separately.

backend/src/apiserver/client_manager/client_manager.go
Linchin: I used this because we are using the package `pgx` as the PostgreSQL driver, and it is needed in the `sql.Open()` function to create the correct DB instance. See the example [here](https://pkg.go.dev/github.com/jackc/pgx/stdlib). I agree it may be quite confusing to people, but I was weighting this against being consistent with the driver name we are using, so I eventually opted for making it clear in the comments to reduce the confusion. WDYT?

backend/src/apiserver/client_manager/client_manager.go
Linchin: I agree that they are pretty much the same fields here, but I think it is possible that there can be different values for the two types of DB, as well as different configs available, so it's safer to have one set of values for each of them. For example, the default host name and port number for MySQL and PostgreSQL are different.

backend/src/apiserver/client_manager/client_manager.go
Linchin: Good point! I moved it because I added integration test for functions here, but it's impossible to test stuff in "main" from a different package in Golang. Reference: https://stackoverflow.com/questions/53997114/how-to-test-the-main-package-in-golang-from-a-test-package

backend/src/apiserver/client_manager/client_manager.go
Linchin: I agree. This part was more of a patch work to make it work for now.

backend/src/apiserver/client_manager/client_manager.go
Linchin: Yes, thanks for finding this! I should add a default user here.

backend/src/apiserver/client_manager/client_manager.go
Linchin: Yes, thank you :)

backend/src/apiserver/client/sql.go
```diff
@@ -44,3 +45,26 @@ func CreateMySQLConfig(user, password string, mysqlServiceHost string,
...
+	var b bytes.Buffer
+	if dbName != "" {
+		fmt.Fprintf(&b, "database=%s ", dbName)
+	}
+	if user != "" {
+		fmt.Fprintf(&b, "user=%s ", user)
+	}
+	if password != "" {
+		fmt.Fprintf(&b, "password=%s ", password)
+	}
```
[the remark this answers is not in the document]
Linchin: `passfile` sounds like a great way to enhance security. Could you tell me more about it? My guess is this will allow users to pass a file containing the password as part of the deployment config. Maybe this can be added together with other configs users can set.

backend/src/apiserver/client/sql.go
```diff
@@ -44,3 +45,26 @@ func CreateMySQLConfig(user, password string, mysqlServiceHost string,
...
+	if postgresHost != "" {
+		fmt.Fprintf(&b, "host=%s ", postgresHost)
+	}
```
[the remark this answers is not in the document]
Linchin: I wonder if this is necessary, since the host address here is simply for communications inside the cluster. Could you tell me more why this is beneficial? Thank you!

backend/test/integration/run_tests_locally.sh
Linchin: added

backend/test/integration/run_tests_locally.sh
Linchin: Thank you for finding this. The command for port forwarding is also slightly different for these two DBs. About 4 lines above. I moved the more repetitive lines outside so the script is more succinct.

backend/src/apiserver/client_manager/client_manager.go
Linchin: Could you help me understand why it is better to use `const` values here? Thank you.

backend/src/apiserver/client_manager/client_manager.go
Linchin: done

backend/src/apiserver/client_manager/client_manager.go
Linchin: Thank you, removed.

backend/test/integration/README.md
Linchin: Ideally we should, and also include PostgreSQL set of tests in the test infra.

backend/src/apiserver/client/sql.go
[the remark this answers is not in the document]
zijianjoy: Makes sense, I am just curious about this change, thank you for the explanation!

backend/src/apiserver/client/sql.go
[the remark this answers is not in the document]
zijianjoy: That is right! So the `passfile` identify the location path to a password, so you don't need to expose the secret via deployment. https://www.postgresql.org/docs/current/libpq-pgpass.html When passing this parameter to postgresql, the `password` field can be omitted.

We can add this configuration in the coming PR, if this makes sense to you.

backend/src/apiserver/client/sql.go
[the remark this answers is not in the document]
zijianjoy: Sure thing, `hostaddr` is helpful when we are connecting to external database, for example: CloudSQL. Yes we can use `host` which is handy for `Unix-domain communication`, but `hostaddr` allows postgresql client to make `TCP/IP communication` instead.

backend/src/apiserver/client_manager/client_manager.go
zijianjoy: Okay I think @Linchin 's point is reasonable, we just need to make sure we document this parameter is being used as  a sql driver flag: https://github.com/golang/go/wiki/SQLDrivers.

backend/src/apiserver/client_manager/client_manager.go
zijianjoy: DB dialects should be stored in a centralized place, so it helps reviewing, modifying and extending. For example: If we integrate with a new DB type in the future, I hope that I can view the existing syntax differences easily in one place, instead of trying to catch those adhoc variables during implementation. 

Therefore, my recommendation is to define a const for `MYSQL_TEXT_FORMAT` and `PGX_TEXT_FORMAT` in a centralized file: https://github.com/kubeflow/pipelines/blob/master/backend/src/apiserver/storage/db.go

backend/src/apiserver/client_manager/client_manager.go
```diff
@@ -597,8 +593,11 @@ func backfillExperimentIDToRunTable(db *gorm.DB) error {
 
 // Returns the same error, if it's not "already exists" related.
 // Otherwise, return nil.
-func ignoreAlreadyExistError(err error) error {
-	if err != nil && strings.Contains(err.Error(), "exists") {
+func ignoreAlreadyExistError(driverName string, err error) error {
+	if driverName == "pgx" && err != nil && strings.Contains(err.Error(), "already exists") {
+		return nil
+	}
+	if driverName == "mysql" && err != nil && strings.Contains(err.Error(), "database exists") {
```
zijianjoy, line 600: We can define consts for each error message keyword `already exists` and `database exists` in a centralized place: https://github.com/kubeflow/pipelines/blob/master/backend/src/apiserver/storage/db.go

manifests/kustomize/third-party/postgresql/base/kustomization.yaml
```diff
@@ -2,7 +2,8 @@ apiVersion: kustomize.config.k8s.io/v1beta1
 kind: Kustomization
 resources:
 - pg-deployment.yaml
+- pg-pv.yaml
```
zijianjoy, line 5: I shouldn't need a PV resource, it is automatically created by PVC when it is bound to a deployment resource.

backend/test/integration/run_tests_locally.sh
Linchin: Thank you! I updated the file.

backend/src/apiserver/client_manager/client_manager.go
Linchin: Added

backend/src/apiserver/client_manager/client_manager.go
Linchin: Thank you so much for the thourought explanation! I added the constants in `backend/src/apiserver/client/sql.go` instead, because this is a SQL related dependency that client_manager is already using.

manifests/kustomize/third-party/postgresql/base/kustomization.yaml
Linchin: I'm not sure about this. I added the PV file because the PVC had trouble creating due to lack of PV. Maybe I was doing something wrong during my deployment. Could you give me more reference on this? Thank you.

Linchin:
part of #9813

backend/src/apiserver/client/sql.go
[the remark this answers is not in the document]
zijianjoy: This allows scenario like connecting to a managed database service. We can discuss about this in future implementation.

manifests/kustomize/third-party/postgresql/base/kustomization.yaml
zijianjoy: You can look at the `Events` for the PVC to learn more about why is PV not created. Maybe the disk driver has something wrong when provisioning a PV. But we shouldn't create PV resource manually in codebase. Refer  to MySQL manifest folder: https://github.com/kubeflow/pipelines/tree/master/manifests/kustomize/third-party/mysql/base

backend/src/apiserver/client_manager/client_manager.go
```diff
@@ -402,7 +402,7 @@ func initDBDriver(driverName string, initConnectionTimeout time.Duration) string
 		dbName = common.GetStringConfig(mysqlDBName)
 	case "pgx":
 		sqlConfig = client.CreatePostgreSQLConfig(
-			common.GetStringConfigWithDefault(postgresUser, ""),
+			common.GetStringConfigWithDefault(postgresUser, "user"),
 			common.GetStringConfigWithDefault(postgresPassword, ""),
```
zijianjoy, line 406: Set password default value as `password` here? reference: https://github.com/kubeflow/pipelines/blob/master/manifests/kustomize/third-party/postgresql/base/pg-secret.yaml#L7

backend/test/integration/run_tests_locally.sh
```diff
@@ -14,6 +14,7 @@
 # limitations under the License.
 
 set -e
+set -x
```
zijianjoy, line 17: These two commands can be combined `set -ex`

backend/src/apiserver/client_manager/client_manager.go
Linchin: Good point! Added the password.

backend/test/integration/run_tests_locally.sh
Linchin: Thank you for catching this! Just fixed it.

manifests/kustomize/third-party/postgresql/base/kustomization.yaml
Linchin: You are right about this. I tried a fresh deployment without the PV, and it passed the tests. The previous error was likely caused by something else. I removed the pv declaration.

zijianjoy (commented):
/lgtm

Great work on this PR, Lingqing!
Waiting for Chen for a final approval.

chensun (commented):
/lgtm

Thanks!

chensun (approved):
/approve

Thanks!

iciclespider:
I get the following error when the db driver name is configured to "gpx":

client_manager.go:428] sql: unknown driver "pgx" (forgotten import?)

According to this, it appears to be a missing import which will register the needed driver: https://stackoverflow.com/a/52791919

PR #9738, opened 2023-07-15
chore(sdk): extract DSL into `kfp-dsl` package

connor-mccarthy:
**Description of your changes:**

**Checklist:**
- [x] The title for your pull request (PR) should follow our title 
convention. [Learn more about the pull request title convention used in this repository](https://github.com/kubeflow/pipelines/blob/master/CONTRIBUTING.md#pull-request-title-convention).

PR #9802, opened 2023-07-30
feat(sdk): Expose typing in kfp python sdk. Fixes #8826

### LINKED ISSUE #8826: [sdk] Types for kfp and kfp.dsl are not found by mypy
### Environment

All dependencies version:

* kfp 1.8.12 and 1.8.18
* mypy 1.0.0

I am managing my dependencies with poetry.  

### Steps to reproduce

```
import kfp.dsl as dsl
```

Run `mypy` 

Produces error:

```
file.py:1: error: Skipping analyzing "kfp.dsl": module is installed, but missing library stubs or py.typed marker  [import]
```

### Expected result

Type stubs are included in project.  

### Materials and Reference

It looks like this issue was addressed in #7451 but it doesn't seem to work for me.  

---

hmc-cs-mdrissi:
Expose typing metadata to users by adding py.typed placemarker file. [PEP 561
](https://peps.python.org/pep-0561/) explains it more, but core is by default a typed library does not expose it's type hints unless py.typed is present.

Should resolve #8826 

**Checklist:**

### DIFF
--- a/sdk/python/MANIFEST.in
+++ b/sdk/python/MANIFEST.in
@@ -1,2 +1,3 @@
 include requirements.in
 include kfp/registry/context/*.json
+include kfp/py.typed


connor-mccarthy (commented):
Thanks for this! I'm happy to support this effort and think it would be a great improvement for users. A few questions:

1) Per "For namespace packages (see PEP 420), the py.typed file should be in the submodules of the namespace, to avoid conflicts and for clarity.", it seems we may need this in the new `kfp-dsl` package. WDYT? See #9738 for context.

2) What other changes are needed to make this fully ready? My sense is that, if any, these should come first, since including this marker file signals endorsement that the KFP SDK's type annotations are a fully ready "feature". A few things that come to mind:
- Type checking is optional and not validated as a presubmit. We probably ought to update our types and change this first.
- When a user authors a component `def comp(foo: str): ...`, foo is typed as a string but, in practice, accepts a string or `PipelineChannel` (pipeline input or upstream task output). Are you aware of a solution to this?

hmc-cs-mdrissi:
> Per "For namespace packages (see PEP 420), the py.typed file should be in the submodules of the namespace, to avoid conflicts and for clarity.", it seems we may need this in the new kfp-dsl package. WDYT? See https://github.com/kubeflow/pipelines/pull/9738 for context.


Yes I think kfp-dsl should have `py.typed` file. I can add that to this pr.




> Type checking is optional and not validated as a presubmit. We probably ought to update our types and change this first.

I agree here that running a type checker (mypy/pyright) in CI would make sense.  This should probably be separate pr. I may be able to look at that later this week, but probably not today.


> When a user authors a component def comp(foo: str): ..., foo is typed as a string but, in practice, accepts a string or PipelineChannel (pipeline input or upstream task output). Are you aware of a solution to this?

The decorator component can be typed like,

```python
def component(func: Callable[..., R]) -> Callable[..., R]:
  ...
```

This will erase input signature so comp(channel) works but it will also allow comp(5.0) (float instead of str) to pass type checking. There is no way currently to have decorator change a function input signature to keep old types + add new thing like PipelineChannel. I've seen other decorators that have same issue (ray's remote decorator) and there's tradeoff of either preserve old signature without allowing PipelineChannel (false positives) or erase input signature allowing Pipelinechannel + other stuff (false negatives). I'd probably make latter choice leaning that few false negatives are better for users then false positives.

connor-mccarthy:
Thanks for sharing how the component decorator can be typed. I haven't tested, but if it works as described this LGTM.

For completing this end-to-end, I recommend the following steps:
1. Update the annotations in the KFP SDK so that mypy checks pass
2. Add a script to invoke mypy ([see existing pattern](https://github.com/GoogleCloudPlatform/oss-test-infra/blob/e9adafab3c4720962d8f7dac1064a9a7aa901e36/prow/prowjobs/kubeflow/pipelines/kubeflow-pipelines-presubmits.yaml#L216))
3. Add a presubmit to invoke the script from the prior step ([see existing pattern](https://github.com/GoogleCloudPlatform/oss-test-infra/blob/e9adafab3c4720962d8f7dac1064a9a7aa901e36/prow/prowjobs/kubeflow/pipelines/kubeflow-pipelines-presubmits.yaml#L216)). Use the flag `optional: true` to prevent blocking on this new presubmit ([example](https://github.com/GoogleCloudPlatform/oss-test-infra/blob/e9adafab3c4720962d8f7dac1064a9a7aa901e36/prow/prowjobs/kubeflow/pipelines/kubeflow-pipelines-presubmits.yaml#L254)).
4. Create another "dummy" PR to invoke the mypy presubmit. Confirm it passes.
5. Remove the `optional: false` flag from the presubmit config.
6. Submit the current PR (#9802), with `kfp-dsl` types too.

PR #9811, opened 2023-08-02
test(frontend): Improve v1 frontend-integration-test

jlyaoyuli:
Previously, we use `tab` key to switch between buttons and text fields. The main purpose of this PR is assign an ID to each input field and directly select it via the ID, which make the integration test more easily to maintain.

### DIFF
--- a/test/frontend-integration-test/helloworld.spec.js
+++ b/test/frontend-integration-test/helloworld.spec.js
@@ -183,6 +183,11 @@ describe('deploy helloworld sample run', () => {
     );
   });
 
+  it('displays run description inputs correctly', async () => {
+    const descriptionValue = await getValueFromDetailsTable('Description');
+    assert.equal(descriptionValue, runDescription, 'run message is not shown correctly');
+  });
+
   it('displays run inputs correctly', async () => {
     const paramValue = await getValueFromDetailsTable('message');
     assert.equal(paramValue, outputParameterValue, 'run message is not shown correctly');
--- a/test/frontend-integration-test/helloworld.spec.js
+++ b/test/frontend-integration-test/helloworld.spec.js
@@ -118,7 +118,7 @@ describe('deploy helloworld sample run', () => {
 
     // await browser.keys('Tab');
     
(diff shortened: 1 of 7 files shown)


test/frontend-integration-test/helloworld.spec.js
```diff
@@ -49,7 +54,9 @@ describe('deploy helloworld sample run', () => {
     await $('#localPackageBtn').click();
     const remoteFilePath = await browser.uploadFile('./helloworld.yaml');
     await $('#dropZone input[type="file"]').addValue(remoteFilePath);
-    await $('#newPipelineName').setValue(pipelineName);
```
jlyaoyuli, line 52: Here is a hidden bug: setValue() won't clear default input value.

test/frontend-integration-test/helloworld.spec.js
```diff
@@ -99,22 +106,14 @@ describe('deploy helloworld sample run', () => {
       reverse: true,
     });
 
-    await browser.keys('Tab');
+    await $('#runNameInput').click();
```
zijianjoy, line 109: We should also `clearDefaultInput()` for the `#runNameInput` first, correct?

test/frontend-integration-test/helloworld.spec.js
```diff
@@ -99,22 +106,14 @@ describe('deploy helloworld sample run', () => {
...
-    // Skip over service account help button
-    await browser.keys('Tab');
-    // Skip over "service account" textbox
-    await browser.keys('Tab');
-    // Skip over "Run Type" radio button
-    await browser.keys('Tab');
-
-    await browser.keys('Tab');
+    
+    await $('input#newRunPipelineParam0').click();
```
zijianjoy, line 115: NIT: We can add a comment to explain what we expect this parameter name is for, for example: This is pipeline parameter `message`.

zijianjoy (commented):
Good work Joe! There are two comments only.

zijianjoy (commented):
/lgtm
/approve

Good work Joe!

Issue #9704, opened 2023-07-05
[e2e test] Docker daemon connection is failed


jlyaoyuli:
Error log:
https://oss.gprow.dev/view/gs/oss-prow/pr-logs/pull/kubeflow_pipelines/9694/kubeflow-pipeline-e2e-test/1674873953513902080

chensun:
Suspect it's caused by GKE upgrade. 

The last succeeded test run was on June 29, [with GKE version 1.25.8-gke.1000](https://oss.gprow.dev/view/gs/oss-prow/pr-logs/pull/kubeflow_pipelines/9693/kubeflow-pipeline-e2e-test/1674513454611828736#)
Then it starts to fail since June 30, [with GKE version 1.26.5-gke.1200](https://oss.gprow.dev/view/gs/oss-prow/pr-logs/pull/kubeflow_pipelines/9694/kubeflow-pipeline-e2e-test/1674873953513902080#)

Looking through the K8s 1.26 release note, I couldn't identify a specific deprecation that could affect us. However, [this Argo doc](https://argoproj.github.io/argo-workflows/walk-through/docker-in-docker-using-sidecars/) warned us that docker-in-docker (dind) using sidecar--that [we're currently using for building test images](https://github.com/kubeflow/pipelines/blob/4a57ab69ddc0abf08daa73473646492bad7ab38d/test/e2e_test_gke_v2.yaml#L271-L278)--are likely to break with future upgrades. It's advised to seek alternatives to building image via dind.

PR #9841, opened 2023-08-09
chore(test): change hard-coded GKE version to 1.25

Linchin:
e2e test failing due to 1.25.10-gke.1200 is no longer supported. ~~Now change the pinned version to 1.25.10-gke.2100, which is what GKE would auto-upgrade to from an 1.25 version.~~

Use version 1.25 instead to allow auto upgrade, as @chensun suggested in the comments.

Error message:
https://oss.gprow.dev/view/gs/oss-prow/pr-logs/pull/kubeflow_pipelines/9798/kubeflow-pipeline-upgrade-test/1689342821942693888#1:build-log.txt%3A129

GKE release note:
https://cloud.google.com/kubernetes-engine/docs/release-notes#2023-r17_version_updates

**Checklist:**

### DIFF
--- a/test/deploy-cluster.sh
+++ b/test/deploy-cluster.sh
@@ -92,7 +92,9 @@ else
   # reference: https://github.com/kubeflow/pipelines/issues/6696
   # Hard-coded GKE to 1.25.10-gke.1200 (the latest 1.25 in STABLE channel). Reference: 
   # https://github.com/kubeflow/pipelines/issues/9704#issuecomment-1622310358
-  gcloud container clusters create ${TEST_CLUSTER} --image-type cos_containerd --release-channel stable --cluster-version 1.25.10-gke.1200 ${SCOPE_ARG} ${NODE_POOL_CONFIG_ARG} ${WI_ARG}
+  # 08/09/2023 update: 1.25.10-gke.1200 no longer supported, use 1.25.10-gke.2100 instead. Reference:
+  # https://cloud.google.com/kubernetes-engine/docs/release-notes-nochannel#2023-r17_version_updates
+  gcloud container clusters create ${TEST_CLUSTER} --image-type cos_containerd --release-channel stable --cluster-version 1.25.10-gke.2100 ${SCOPE_ARG} ${NODE_POOL_CONFIG_ARG} ${WI_ARG}
 fi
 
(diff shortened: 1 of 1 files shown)


test/deploy-cluster.sh
```diff
@@ -92,7 +92,9 @@ else
   # reference: https://github.com/kubeflow/pipelines/issues/6696
   # Hard-coded GKE to 1.25.10-gke.1200 (the latest 1.25 in STABLE channel). Reference: 
   # https://github.com/kubeflow/pipelines/issues/9704#issuecomment-1622310358
-  gcloud container clusters create ${TEST_CLUSTER} --image-type cos_containerd --release-channel stable --cluster-version 1.25.10-gke.1200 ${SCOPE_ARG} ${NODE_POOL_CONFIG_ARG} ${WI_ARG}
+  # 08/09/2023 update: 1.25.10-gke.1200 no longer supported, use 1.25.10-gke.2100 instead. Reference:
```
chensun, line 95: Can we still specify 1.25 explicitly? In that way, when the default changed to 1.26+, the error message would be more clear and strait-forward instead of being "daemon connection failed".

https://cloud.google.com/kubernetes-engine/versioning#specifying_cluster_version

gkcalat (commented):
Thanks, @Linchin

/lgtm

Linchin:
Test results:
* Using stable channel (1.27) still doesn't work. We may need to address the dind issue to resolve this
* 1.25.10-gke.2100 test passed. Now try using GKE version 1.25 directly to allow auto upgrade, so the test does not break when GKE no longer support 1.25.10-gke.2100.

test/deploy-cluster.sh
Linchin: Thank you for providing the info on ways we can specify cluster version, it's really helpful! Using 1.25 would make the test cluster much more reliable. I will change the version and rerun the test.

gkcalat (approved):
/lgtm

Linchin:
Note: is related to #9704

bsikander:
@Linchin is there a compatability matrix between kubeflow pipelines and kubernetes version?
I have kfp 1.8.5 and GKE 1.26 and my deployment of kfp doesn't work

Issue #9820, opened 2023-08-04
[sdk] KFP SDK 2.1.2 breaks when upgraded from an older version.


chensun:
### Environment

*  KFP version:
N/A

*  KFP SDK version: 
2.1.2

*  All dependencies version:
N/A

### Steps to reproduce

- Step 1: `pip3 install kfp==2.0.1`
- Step 2: `pip3 install kfp==2.1.2`
- Step 3: `import kfp`

### Expected result

import kfp should work.

### Actual result

```
>>> import kfp
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/Library/Frameworks/Python.framework/Versions/3.9/lib/python3.9/site-packages/kfp/__init__.py", line 23, in <module>
    from kfp import components
  File "/Library/Frameworks/Python.framework/Versions/3.9/lib/python3.9/site-packages/kfp/components/__init__.py", line 27, in <module>
    from kfp.components.load_yaml_utilities import load_component_from_file
  File "/Library/Frameworks/Python.framework/Versions/3.9/lib/python3.9/site-packages/kfp/components/load_yaml_utilities.py", line 20, in <module>
    from kfp.dsl import structures
  File "/Library/Frameworks/Python.framework/Versions/3.9/lib/python3.9/site-packages/kfp/dsl/structures.py", line 25, in <module>
    from kfp.dsl import placeholders
  File "/Library/Frameworks/Python.framework/Versions/3.9/lib/python3.9/site-packages/kfp/dsl/placeholders.py", line 22, in <module>
    from kfp.dsl.types import type_utils
  File "/Library/Frameworks/Python.framework/Versions/3.9/lib/python3.9/site-packages/kfp/dsl/types/type_utils.py", line 239, in <module>
    expected_spec: Union[structures.InputSpec, structures.OutputSpec],
AttributeError: partially initialized module 'kfp.dsl.structures' has no attribute 'InputSpec' (most likely due to a circular import)
```

chensun:
We already yanked KFP SDK 2.1.2, so that users' won't upgrade to this version unless they pin the version specifically in the command.

In case you run into this issue, and stuck with a bad environment. The remedy steps are:
1. Uninstall both `kfp` and `kfp-dsl` package: `pip3 uninstall -y kfp kfp-dsl`
2. Reinstall an previous version, e.g.: `pip3 install kfp==2.0.1`.

Issue #8370, opened 2022-10-14
[sdk] Protobuf 4 Support


alanhdu:
### Environment

The latest `kfp` version pins `protobuf < 4.0`:

https://github.com/kubeflow/pipelines/blob/71c8e77d8a583c629e6f83dcd629fee8133f6d2a/sdk/python/requirements.in#L27

What would it take for `kfp` to support `protobuf==4.23`?  It looks like Python's protobuf package went from `3.20 -> 4.21` and other packages are now starting to require `protobuf > 4`.

The list of breaking changes is available at https://developers.google.com/protocol-buffers/docs/news/2022-05-06#python-updates.

stepanhruda:
Any progress on this? New packages are starting to depend on protobuf 4, making dependency resolution impossible.

connor-mccarthy:
We plan to release at least one version of `kfp==2.x.x` that depends on protobuf 3 before bumping to depend on protobuf 4.

lrodriguezesc:
Any ETA here?

anuvedantham:
bumping this up, any ETA?

connor-mccarthy:
After more investigation, I'm not sure that we can bump major versions within KFP SDK v2. Doing so would require removing support for protobuf 3 per the protobuf [cross-version runtime guarantees](https://protobuf.dev/support/cross-version-runtime-guarantee/#major). This would prevent some users from upgrading KFP SDK versions. For a other smaller dependencies with greater compatibility across major versions, this might be acceptable. This is not the case for protobuf, unfortunately.

Please feel free to provide additional information if there is something I'm missing. I will close for now.

mynewestgitaccount:
> Please feel free to provide additional information if there is something I'm missing.

@connor-mccarthy I feel like there must be something being missed here.

A number of packages have successfully managed to allow simultaneous support of `protobuf` 3 and 4 (e.g. take a look at [`google-cloud-pubsub`'s requirements](https://github.com/googleapis/python-pubsub/blob/main/setup.py#L44)), so I don't follow why it's an impossibility for `kfp`.

> This would prevent some users from upgrading KFP SDK versions.

[...]

PR #9846, opened 2023-08-10
chore(sdk): undo creation of kfp-dsl namespace package

connor-mccarthy:

**Checklist:**
- [x] The title for your pull request (PR) should follow our title 
convention. [Learn more about the pull request title convention used in this repository](https://github.com/kubeflow/pipelines/blob/master/CONTRIBUTING.md#pull-request-title-convention).


mynewestgitaccount:
I noticed a number of PRs that I assume are meant to be tests of potential ways to handle #9820, and I really hope you won't choose this one as the "solution". `kfp-dsl` in components is a great way to (partially) sidestep the dependency resolution issues caused by the `protobuf < 4` pin in the main package (ref: #8370), and it would be a shame to remove that.

<hr>

P.S. I'd go so far as to say that even #9845 would be preferable to this. Nobody likes to get an error, but if a user is already doing an upgrade, adding an additional step to that upgrade shouldn't be a big deal.

PR #9894, opened 2023-08-17
feat(sdk): support dsl.If, dsl.Elif, and dsl.Else

connor-mccarthy:
Supports authoring pipelines that use `dsl.If`, `dsl.Elif`, and `dsl.Else`. For example:

```python
from kfp import dsl

@dsl.pipeline
def flip_coin_pipeline():
    flip_coin_task = flip_coin()
    with dsl.If(flip_coin_task.output == 'heads'):
        print_and_return(text='Got heads!')
    with dsl.Elif(flip_coin_task.output == 'tails'):
        print_and_return(text='Got tails!')
    with dsl.Else():
        print_and_return(text='Draw!')
```

Note that `dsl.If` is merely an alias for `dsl.Condition`. `dsl.Elif` and `dsl.Else` are new functionality.

**Checklist:**
- [x] The title for your pull request (PR) should follow our title convention. [Learn more about the 
pull request title convention used in this repository](https://github.com/kubeflow/pipelines/blob/master/CONTRIBUTING.md#pull-request-title-convention).

### DIFF
--- a/sdk/python/kfp/compiler/compiler_utils.py
+++ b/sdk/python/kfp/compiler/compiler_utils.py
@@ -119,7 +119,18 @@ def get_parent_groups(
     return (tasks_to_groups, groups_to_groups)
 
 
-# TODO: do we really need this?
+def get_channels_from_condition(
+    operations: List[pipeline_channel.BinaryOperation],
+    collected_channels: list,
+) -> None:
+    """Append to collected_channels each pipeline channels used in each operand
+    of each operation in operations."""
+    for operation in operations:
+        for operand in [operation.left_operand, operation.right_operand]:
+            if isinstance(operand, pipeline_channel.PipelineChannel):
+                collected_channels.append(operand)
+
+
 def get_condition_channels_for_tasks(
     root_group: tasks_group.TasksGroup,
 ) -> Mapping[str, Set[pipeline_channel.PipelineChannel]]:
@@ -139,16 +150,13 @@ def get_condition_channels_for_tasks(
         current_conditions_channels,
     ):
         new_current_conditions_channels = current_conditions_channels
-        if isinstance(group, tasks_group.Condition):
+        if isinstance(group, tasks_group._ConditionBase):
             new_current_conditions_channels = list(current_conditions_channels)
-            if isinstance(group.condition.left_operand,
-                          pipeline_channel.PipelineChannel):
-                new_current_conditions_channels.append(
-                    group.condition.left_operand)
-            if isinstance(group.condition.right_operand,
-                          pipeline_channel.PipelineChannel):
-                new_current_conditions_channels.append(
-                    group.condition.right_operand)
+            get_channels_from_condition(
+                group.condition,
+                new_current_conditions_channels,
+            )
+
         for task in group.tasks:
             for channel in new_current_conditions_channels:
                 conditions[task.name].add(channel)
@@ -661,8 +669,9 @@ def get_dependencies(
                 dependent_group = group_name_to_group.get(
                     uncommon_upstream_groups[0], None)
 
-                if isinstance(dependent_group,
-                              (tasks_group.Condition, tasks_group.ExitHandler)):
+                if isinstance(
+                        dependent_group,
+                    (tasks_group._ConditionBase, tasks_group.ExitHandler)):
                     raise InvalidTopologyException(
                         f'{ILLEGAL_CROSS_DAG_ERROR_PREFIX} A downstream task cannot depend on an upstream task within a dsl.{dependent_group.__class__.__name__} context unless the downstream is within that context too. Found task {task.name} which depends on upstream task {upstream_task.name} within an uncommon dsl.{dependent_group.__class__.__name__} context.'
                     )
--- a/sdk/python/kfp/compiler/compiler_utils.py
+++ b/sdk/python/kfp/compiler/compiler_utils.py
@@ -123,8 +123,8 @@ def get_channels_from_condition(
     operations: List[pipeline_channel.BinaryOperation],
     collected_channels: list,
 ) -> None:
-    """Append to collected_channels each pipeline channels used in each operand
-    of each operation in operations."""
+    """Appends to collected_channels each pipeline channels used in each
+    operand of each operation in operations."""
     for operation in operations:
         for operand in [operation.left_operand, operation.right_operand]:
             if isinstance(operand, pipeline_channel.PipelineChannel):
@@ -153,7 +153,7 @@ def get_condition_channels_for_tasks(
         if isinstance(group, tasks_group._ConditionBase):
             new_current_conditions_channels = list(current_conditions_channels)
             get_channels_from_condition(
-                group.condition,
+                group.conditions,
                 new_current_conditions_channels,
             )
 
--- a/sdk/python/kfp/compiler/compiler_utils.py
+++ b/sdk/python/kfp/compiler/compiler_utils.py
@@ -120,7 +120,7 @@ def get_parent_groups(
 
 
 def get_channels_from_condition(
-    operations: List[pipeline_channel.BinaryOperation],
+    operations: List[pipeline_channel.ConditionOperation],
     collected_channels: list,
 ) -> None:
     """Appends to collected_channels each pipeline channels used in each
--- a/sdk/python/kfp/dsl/pipeline_channel.py
+++ b/sdk/python/kfp/dsl/pipeline_channel.py
@@ -24,17 +24,20 @@ from kfp.dsl.types import type_utils
 
 
 @dataclasses.dataclass
-class ConditionOperator:
-    """Represents a condition expression to be used in dsl.Condition().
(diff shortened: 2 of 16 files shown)


sdk/python/test_data/pipelines/if_elif_else_complex.py
```diff
@@ -0,0 +1,60 @@
...
+    print(text)
+    return text
+
+
+@dsl.pipeline
+def flip_coin_pipeline(confirm: bool):
+    int_task = int_zero_through_three()
+    flip_coin_task = flip_coin()
+
+    with dsl.If(flip_coin_task.output == 'heads'):
```
chensun, line 41: nit: for a complex exmaple, how about some nesting with loops and/or complex conditions containing logical operators, numerical ranges, etc.?

sdk/python/kfp/compiler/compiler_utils.py
```diff
@@ -119,7 +119,18 @@ def _get_parent_groups_helper(
     return (tasks_to_groups, groups_to_groups)
 
 
-# TODO: do we really need this?
+def get_channels_from_condition(
+    operations: List[pipeline_channel.BinaryOperation],
+    collected_channels: list,
+) -> None:
+    """Append to collected_channels each pipeline channels used in each operand
```
chensun, line 126: nit: Appends

(for consistent style)

sdk/python/kfp/dsl/pipeline_channel.py
```diff
@@ -24,17 +24,20 @@
 
 
 @dataclasses.dataclass
-class ConditionOperator:
-    """Represents a condition expression to be used in dsl.Condition().
+class BinaryOperation:
```
chensun, line 27: `negate` is a unary operator, so the name here seems not accurate.

sdk/python/kfp/dsl/tasks_group.py
```diff
@@ -138,9 +140,31 @@ def __init__(
...
+    Elif, Else).
+
+    Args:
+        condition: A list of binary operations to be combined via conjunction.
+        name: The name of the condition group.
+    """
+
+    def __init__(
+        self,
+        condition: List[pipeline_channel.BinaryOperation],
```
chensun, line 154: nit: should it be named `conditions` to reflect a list?

sdk/python/kfp/dsl/tasks_group.py
```diff
@@ -150,22 +174,151 @@ class Condition(TasksGroup):
...
         task1 = my_component1(...)
-        with Condition(task1.output=='pizza', 'pizza-condition'):
+        with dsl.If(task1.output=='pizza', 'pizza-condition'):
             task2 = my_component2(...)
     """
 
     def __init__(
         self,
-        condition: pipeline_channel.ConditionOperator,
+        condition,
```
chensun, line 183: nit: type annotation?

sdk/python/kfp/dsl/tasks_group.py
```diff
@@ -150,22 +174,151 @@ class Condition(TasksGroup):
...
+        self,
+        condition,
+        name: Optional[str] = None,
+    ) -> None:
+        prev_cond = pipeline_context.Pipeline.get_default_pipeline(
+        ).get_last_tasks_group()
+        if not isinstance(prev_cond, (Condition, If, Elif)):
+            # prefer pushing toward dsl.If rather than dsl.Condition for syntactic consistency with the if-elif-else keywords in Python
+            raise InvalidControlFlowException(
+                'dsl.Else can only be used following an upstream dsl.If or dsl.Elif.'
```
chensun, line 244: `dsl.Elif` can only be...

sdk/python/kfp/dsl/tasks_group.py
```diff
@@ -150,22 +174,151 @@ class Condition(TasksGroup):
...
+        with dsl.If(task1.output=='pizza', 'pizza-condition'):
+            task3 = my_component3(...)
+
+        with dsl.Elif(task2.output=='pasta', 'pasta-condition'):
+            task4 = my_component4(...)
+    """
+
+    def __init__(
+        self,
+        condition,
```
chensun, line 236: nit: type annotation?

sdk/python/test_data/pipelines/if_elif_else.py
```diff
@@ -0,0 +1,51 @@
...
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+from kfp import compiler
+from kfp import dsl
+
+
+@dsl.component
+def flip_coin() -> str:
```
chensun, line 19: nit: maybe rename this function? "flipping a coin" typically expects a binary result, and the addition of `draw` makes the example a bit odd.

sdk/python/kfp/dsl/pipeline_context.py
```diff
@@ -189,6 +189,12 @@ def pop_tasks_group(self):
         """Removes the current TasksGroup from the stack."""
         del self.groups[-1]
 
+    def get_last_tasks_group(self) -> Union['tasks_group.TasksGroup', None]:
```
chensun, line 192: Why not just use `Optional[x]` than `Union[x, None]`?

sdk/python/kfp/dsl/pipeline_channel.py
connor-mccarthy: I've been thinking of this class as a binary operation comprised of an `operator` between a `left_operand` and a `right_operand`, with the `negate` flag as a secondary consideration on top of the `BinaryOperation` instance's core data (the other three fields). It seems a bit too heavyweight to split out these abstractions entirely into `BinaryOperation` and `UnaryOperation`/`Negation` classes. I'd prefer to either (a) keep the name as-is, (b) rename to something like `BinaryOperationWithNegation`, or (c) rename to `Operation`. I think I (weakly) prefer those options in that order. Do you have a preference?

sdk/python/kfp/dsl/pipeline_context.py
connor-mccarthy: Updated.

sdk/python/kfp/dsl/tasks_group.py
connor-mccarthy: I intentionally removed this, since I personally find it a bit confusing and I suspect users do too. See the [current reference docs](https://kubeflow-pipelines.readthedocs.io/en/master/source/dsl.html#kfp.dsl.Condition) which use this annotation.

While annotation is technically correct, it isn't informative to the user and exposes an [implementation detail](https://github.com/kubeflow/pipelines/blob/760c1589edbe58bbd77611222a66a17b371a0d08/sdk/python/kfp/dsl/pipeline_channel.py#L151-L167). We don't actually want the user to provide a `ConditionOperator` (name changed in this PR)... we want them to provide a comparative expression involving one or more pipeline channels (i.e., to think of it as a "future" boolean).

WDYT?

sdk/python/kfp/dsl/tasks_group.py
connor-mccarthy: See [comment above](https://github.com/kubeflow/pipelines/pull/9894#discussion_r1320240559)

sdk/python/kfp/dsl/tasks_group.py
connor-mccarthy: Thank you.

sdk/python/test_data/pipelines/if_elif_else.py
connor-mccarthy: Updated.

sdk/python/test_data/pipelines/if_else.py
```diff
@@ -0,0 +1,42 @@
+# Copyright 2023 The Kubeflow Authors
```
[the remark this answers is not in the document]
connor-mccarthy: I'd slightly prefer to keep this test case to separately test that the compilation logic for `dsl.Else` is correct both following a `dsl.If` (single condition to negate) and following a `dsl.If` + `dsl.Elif` (>1 condition to negate).

sdk/python/test_data/pipelines/if_elif_else_complex.py
connor-mccarthy: Good idea. Updated to compose more elements of the DSL together.

sdk/python/kfp/dsl/pipeline_channel.py
chensun: I agree not to split the class into two. But the naming is still a bit of as I read the class name I won't think of `negate` being part of it. What do you think about `ConditionOperation`?

connor-mccarthy: SG. Updated.

chensun (approved):
/lgtm
/approve

Thanks!

Issue #9904, opened 2023-08-18
[backend] Persistence agent failing `kubeflow-pipeline-mkp-test`


zijianjoy:
### Environment

Currently `/test kubeflow-pipeline-mkp-test` is failing on HEAD.

### Steps to reproduce

In https://github.com/kubeflow/pipelines/commit/cb18d00bbbaed9cd77fc50dce739ed62c72b2356#diff-7df17678529866306ef6665ffbeeda377a1e869aca24856f7cf8a99891877fe8, service account token projection is introduced to enable Auth. However, it is failing the `kubeflow-pipeline-mkp-test` test because persistent agent is unable to find the service account token. Error message as below:

![image](https://github.com/kubeflow/pipelines/assets/37026441/fe516db0-f9eb-4613-86ad-51bd81bf10e4)

cc @difince @chensun 

---

Impacted by this bug? Give it a 👍.

zijianjoy:
### Exploration

I granted `roles/iam.serviceAccountTokenCreator` to the GCE default service account: https://cloud.google.com/iam/docs/service-account-permissions#token-creator-role, and retried the mkt-test. If this permission unblocks the test, then we will need to do one of the followings:

1. Make SA token requirement configurable (which means it can become optional), it allows us to exclude volume token projection for KFP standalone mode. We still need to clearly document the permission needed in the full kubeflow mode.
2. Update all existing clusters to grant this additional permission. Put up a warning to all users about this breaking change and how should they unblock the KFP deployment. Update documentations for KFP standalone and full Kubeflow.

difince:
From the log message, it looks like in GCE the SA token hasn't been created. 
I would like to clearly state that in the current (the new) implementation, PA( Persistence Agent) authentication is not related to the multiuser mode at all. No matter the multiuser mode, PA is the only one that has the permissions and the right identity to call these two endpoints (no matter the namespace the reported workflow belongs to). 
As I do not have any knowledge of GCE .. I will totally count on your decision and if really needed I will make it configurable ..

juliusvonkohout:
Yes, in a normal kubeflow there is a serviceaccount

```
        volumeMounts:
          - mountPath: /var/run/secrets/kubeflow/tokens
            name: persistenceagent-sa-token
      serviceAccountName: ml-pipeline-persistenceagent
      volumes:
        - name: persistenceagent-sa-token
          projected:
            sources:
              - serviceAccountToken:
                  path: persistenceagent-sa-token
[...]

PR #9908, opened 2023-08-18
fix(manifests): Update persistence agent manifests in marketplace helm chart

zijianjoy:

Fix https://github.com/kubeflow/pipelines/issues/9904. Syncing manifest changes to marketplace. Resolving presubmit issue.

**Checklist:**


manifests/gcp_marketplace/chart/kubeflow-pipelines/templates/pipeline.yaml
```diff
@@ -149,6 +149,19 @@ rules:
...
+      - pipelines.kubeflow.org
+    resources:
+      - scheduledworkflows
+      - workflows
+    verbs:
+      - report
+  - apiGroups:
+      - ''
+    resources:
+      - namespaces
```
difince, line 162: @zijianjoy Glad you catch the root cause of the failing tests! 
Without any prior knowledge of GCE and its tests, I want to mention that I see this "copy and paste" of config files as an error-prone place .. that could probably be automated. 
Glad that you catch the missing:
```
- apiGroups:
  - ''
  resources:
  - namespaces
  verbs:
  - get
 ```
It was actually missing for quite a lot of time.
Something that is still missing is: 
https://github.com/kubeflow/pipelines/blob/5b59e4a76fa4a086ebb4e6b590557bdd72e86cb5/manifests/kustomize/base/pipeline/ml-pipeline-persistenceagent-deployment.yaml#L26-L31

I guess that the GCE tests are run in multi-user mode = false, otherwise, I would expect `readArtifact` and `ReportRunMetric` to be failing. Is this the case?

difince (changes requested):

manifests/gcp_marketplace/chart/kubeflow-pipelines/templates/pipeline.yaml
zijianjoy: Thank you @difince for the information. Yes these GCE tests are in KFP standalone mode. I agree with you that we should automate the process of generating marketplace helm chart. 

In fact, when looking at the charts in gcp_marketplace folder, I sense that they are generated by kustomize at the first place. Unfortunately I cannot find the documented process for how to regenerate. We can consider introducing a kustomize overlay for marketplace, so most of future yaml changes can land on gcp_marketplace automatically.

chensun (approved):
/lgtm
/approve

Thanks!

Issue #9880, opened 2023-08-16
[feature] request a minimum amount of GPU VRAM for scheduling 


axel7083:
### Feature Area

/area sdk

### What feature would you like to see?

When using the GPU Operator a lot of labels are generated[^1] to the nodes such as:
- `nvidia.com/gpu.memory` (Integer) 
- `nvidia.com/cuda.runtime.major`  (Integer)
etc..

It is possible in Kubernetes to add nodeAffinity with comparaison (greater, lower). The following syntax can be used for a pod

```yaml
apiVersion: v1
kind: Pod
metadata:
  name: with-affinity-anti-affinity
spec:
  affinity:
    nodeAffinity:
      requiredDuringSchedulingIgnoredDuringExecution:
        nodeSelectorTerms:
        - matchExpressions:
          - key: nvidia.com/gpu.memory
            operator: Gt
            values:
            - 50
...
```

This would allows to specify the minimum amount of memory required, or the minimum cuda version, or the interval allowed.
The kfp[kubernetes] package have the `add_node_selector` function, that could (probably?) be extended to support this ?

---

Love this idea? Give it a 👍. 

[^1]: [Generated labels](https://github.com/NVIDIA/gpu-feature-discovery#generated-labels)

PR #9913, opened 2023-08-19
feat(backend): Adding NodeAffinity support in v2 driver

axel7083:
Fixes: https://github.com/kubeflow/pipelines/issues/9880

## Description

The current v2 driver support `NodeSelector` spec in PodSpec. Allowing the `kfp[kubernetes]` package to add the `add_node_selector` function.

This is great but Kubernetes offer way more control to handle the Affinity. The main issue with NodeSelector that our team had was the lack of comparaison. We are using a series of servers with different GPU on each of them for different need, and when using the NVIDIA-GPU-Operator, we cannot _request_ a minimum  amount of VRAM. This information is however available on the `nvidia.com/gpu.memory` generated label. 

> Our workaround is to give a table to our users, with the list of GPU we are using, and ask them to use the NodeSelector with the `nvidia.com/gpu.product` label.

## Affinity support

This PR add the [`nodeAffinity`](https://kubernetes.io/docs/reference/kubernetes-api/workload-resources/pod-v1/#NodeAffinity), `nodeName`[^1] and `nodeSelector`[^1] support.

> The [PodAffinity](https://kubernetes.io/docs/reference/kubernetes-api/workload-resources/pod-v1/#PodAffinity), and [PodAntiAffinity](https://kubernetes.io/docs/reference/kubernetes-api/workload-resources/pod-v1/#PodAntiAffinity) are not included in this PR.

## SDK Changes

I moved the existing nodeSelector under the scheduling in the proto file, to match the Kubernetes format and keep everything related to affinity under the same name.

I did not add help functions in the kfp[kubernetes] package for using the new features, the goal here is to keep the current behavior as it is and in a future PR add them (or not depending on feedback here).

## Tests

I wrote unit test for the go driver-v2, and updated the tests for the `kubernetes_plateform` python package. The tests run properly.

> I did not test it in a real cluster with multiple nodes and with an existing pipeline.

[^1]: In the [Scheduling](https://kubernetes.io/docs/reference/kubernetes-api/workload-resources/pod-v1/#NodeAffinity) section.


axel7083:
@Linchin , @gkcalat since I changed the go package github.com/kubeflow/pipelines/kubernetes_platform and the driver depending on it, how to handle that here ? Do I make two different PRs ? One for the kubernetes_platform alone, and another one for the driver ?

Linchin:
@axel7083 Yes, that should resolve the current test failure. Thank you for this great work!

Linchin:
@connor-mccarthy, after merging the first PR, do we need to release an SDK version for the second PR to work?

connor-mccarthy:
@Linchin, I think the order of submission and release would actually be the reverse. We should add BE support, then add the ability to access those features via the `kfp-kubernetes` extension package.

axel7083:
> @Linchin, I think the order of submission and release would actually be the reverse. We should add BE support, then add the ability to access those features via the `kfp-kubernetes` extension package.

I am not sure to understand, the proto file is located in `kubernetes_platform/proto` and this packagewould be needed by the driver ? So it makes sense to add it first (sadly?)

Linchin:
> > @Linchin, I think the order of submission and release would actually be the reverse. We should add BE support, then add the ability to access those features via the `kfp-kubernetes` extension package.
> 
> I am not sure to understand, the proto file is located in `kubernetes_platform/proto` and this packagewould be needed by the driver ? So it makes sense to add it first (sadly?)

I agree with you. This was the order we added the first batch of kubernetes-specific features.

AlexandreBrown:
Hello, great work!
Out of curiosity, will these changes be usable via an sdk upgrade or does it require backend changes ?

Thanks

connor-mccarthy:
I think the BE needs to support this first. Then, the KFP SDK's extension package `kfp-kubernetes` (effectively a client of the BE, for purposes of this feature), can add support for authoring these features. Without implementing support in the BE first, the "client" would be sending requests to the "server" with fields the "server" does not know how to handle.
I defer to the KFP BE on if/how/when this should be supported.

I defer to the KFP BE on if/how/when this should be supported.

axel7083:
@Linchin  any idea of the following steps then ?

axel7083:
> Is this still being worked on? This feature is absolutely needed for us if we want to migrate from V1 to V2. Would ne great to have it :)

Not on my side.

phred-unity:
> > Is this still being worked on? This feature is absolutely needed for us if we want to migrate from V1 to V2. Would ne great to have it :)
> 
> Not on my side anymore.

Thanks for the update! So if I understand correctly, V2 doesn't support setting tolerations / nodeSelector yet and no one is actively working on it?

axel7083:
> > > Is this still being worked on? This feature is absolutely needed for us if we want to migrate from V1 to V2. Would ne great to have it :)
> > 
> > 
> > Not on my side anymore.
> 
> Thanks for the update! So if I understand correctly, V2 doesn't support setting tolerations / nodeSelector yet and no one is actively working on it?

This is probably a good summarize :/

rimolive:
@axel7083 Are you still working on this PR? We introduced DCO checks in this repo so make sure you signed-off all commits.

axel7083:
> @axel7083 Are you still working on this PR? We introduced DCO checks in this repo so make sure you signed-off all commits.

No

PR #9951, opened 2023-09-01
chore(frontend): Improve structure of NewExperimentFC.

zijianjoy:

- Remove the need for `setErrorMessage`, directly set the error message in html based on experiment name availability.
- Remove the need for `useEffect` for `setLatestVersion`, replaced with `useQuery` because `getLatestVersion()` is an API call.
- Add explanation for places where we need to disable `react-hooks/exhaustive-deps`.
- `setErrMsgFromApi` to undefined if the create experiment call is successful.
- Extended timeout for NewRun test since jest has been upgraded implicitly to 27. It is the version which `modern` fake timer has become default.

**Checklist:**

### DIFF
--- a/frontend/src/pages/NewRun.test.tsx
+++ b/frontend/src/pages/NewRun.test.tsx
@@ -1699,7 +1699,7 @@ describe('NewRun', () => {
       await screen.findByDisplayValue('prefilled value 1');
       await screen.findByLabelText('param-2');
       await screen.findByDisplayValue('prefilled value 2');
-    });
+    }, 10000);
 
     it('trims whitespace from the pipeline params', async () => {
       tree = shallow(<TestNewRun {...generateProps()} />);
--- a/frontend/src/pages/functional_components/NewExperimentFC.tsx
+++ b/frontend/src/pages/functional_components/NewExperimentFC.tsx
@@ -14,21 +14,21 @@
  * limitations under the License.
  */
 
-import React, { useEffect, useState } from 'react';
-import BusyButton from 'src/atoms/BusyButton';
 import Button from '@material-ui/core/Button';
-import Input from 'src/atoms/Input';
+import React, { useEffect, useState } from 'react';
+import { useMutation, useQuery } from 'react-query';
+import { commonCss, fontsize, padding } from 'src/Css';
 import { V2beta1Experiment } from 'src/apisv2beta1/experiment';
+import { V2beta1PipelineVersion } from 'src/apisv2beta1/pipeline';
+import BusyButton from 'src/atoms/BusyButton';
+import Input from 'src/atoms/Input';
+import { QUERY_PARAMS, RoutePage } from 'src/components/Router';
 import { Apis } from 'src/lib/Apis';
-import { PageProps } from 'src/pages/Page';
-import { RoutePage, QUERY_PARAMS } from 'src/components/Router';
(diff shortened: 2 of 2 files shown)


frontend/src/pages/functional_components/NewExperimentFC.tsx
```diff
@@ -97,17 +94,11 @@ export function NewExperimentFC(props: NewExperimentFCProps) {
...
-
-  // Handle the error when createExperiment() is failed
-  useEffect(() => {
-    if (!experimentName) {
-      setErrorMessage('Experiment name is required');
-    } else {
-      setErrorMessage('');
-    }
-  }, [experimentName]);
+  }, [experimentResponse, latestVersion]);
```
jlyaoyuli, line 101: I think once the `pipelineId` is changed, the `latestVersion` will change too (cause re-render as well). Not sure if it's good to include `latestVersion` in the dependency array.

frontend/src/pages/functional_components/NewExperimentFC.tsx
```diff
@@ -118,8 +109,9 @@ export function NewExperimentFC(props: NewExperimentFCProps) {
         title: 'Experiment creation failed',
       });
     }
+    // Do not rerun this effect if updateDialog callback has changes to avoid re-rendering.
```
jlyaoyuli, line 112: Is this line of comment used to avoid rerun the useEffect()?

frontend/src/pages/NewRun.test.tsx
```diff
@@ -1699,7 +1699,7 @@ describe('NewRun', () => {
       await screen.findByDisplayValue('prefilled value 1');
       await screen.findByLabelText('param-2');
       await screen.findByDisplayValue('prefilled value 2');
-    });
+    }, 10000);
```
jlyaoyuli, line 1702: Can I ask what this change is used for?

frontend/src/pages/functional_components/NewExperimentFC.tsx
zijianjoy: `pipelineId` is not the only reason that causes `latestVersion` to change. `latestVersion` is being pulled from `useQuery`, and the query can be executed any time and multiple times during the page is visible. Since we are using  `latestVersion` in this `useEffect` code block, I added the `latestVersion` to the dependency array.

frontend/src/pages/functional_components/NewExperimentFC.tsx
zijianjoy: This comment is an explanation for disabling eslint `react-hooks/exhaustive-deps`. Usually we try to fulfill the requirement of eslint by adding all required dependencies to the dependency list. But in this case, we are skipping some dependency, this line of comment is to explain what variable we have skipped and the reason.

frontend/src/pages/NewRun.test.tsx
zijianjoy: Explained in PR description https://github.com/kubeflow/pipelines/pull/9951#issue-1877927798:

```
Extended timeout for NewRun test since jest has been upgraded implicitly to 27. It is the version which modern fake timer has become default.
```

I think a better fix is to explicitly pin the jest version, and add a jest config file for the fake timer. Since some effort needs to be invested for that, I want to narrow the PR focus here. (This workaround is for passing the presubmit test)

PR #9953, opened 2023-09-02
chore(frontend): Refactor ExperimentDetails to functional component

jlyaoyuli:
Add alternative functional component for experiment details page.


frontend/src/pages/ExperimentDetails.tsx
```diff
@@ -435,7 +437,11 @@ const EnhancedExperimentDetails: React.FC<PageProps> = props => {
     return <Redirect to={RoutePage.EXPERIMENTS} />;
   }
 
-  return <ExperimentDetails {...props} />;
+  return isFeatureEnabled(FeatureKey.FUNCTIONAL_COMPONENT) ? (
```
chensun, line 440: My first reaction opening this PR was why there is no code removal for a refactoring change. Then I saw this feature flag. 
Generally speaking, a refactoring should achieve the goal of reducing complexity and lowering development and maintenance cost. By making pure additions of code behind a feature flag doesn't achieve the goal and even increases the tech debt. 
We should probably sync offline and put up a plan for the refactoring work.

chensun (changes requested):

PR #9969, opened 2023-09-11
fix(sdk): make component names more unique when merging specs

stijntratsaertit:
When having having multiple levels of nesting in pipelines (especially when using a similar pipeline), components have a greater chance to conflict in naming and resulting in wrong references in the YAML. 

Using the following change, names of already renamed components are also taken into account when altering a component name.

**Checklist:**


connor-mccarthy:
@stijntratsaertit, can you describe what you mean by more unique? Have you encountered a key collision?

stijntratsaertit:
Hi, I indeed encountered a key collision, more specifically when dealing with multiple for loops. It seems to go wrong when it's merging the specs. In VertexAI Pipelines, the UI was rendering wrong components within a for loop and in YAML I was able to see that wrong reference because the componentRef of a for-loop was pointing to a different component.

The names will now be more unique in the sense that it will now generate a new name when the same name exists in the current subgroup components (this was my initial change).

I have now gone back to it and noticed an additional change would be needed. The renaming should be done starting from the last generated name mapping, in order to avoid a case where unintentional renames were being done as well.

stijntratsaertit:
I was able to provide the test case based on the case I'm dealing with. Test fails on master, but succeeds on this branch. I'll pick up linting soon (todo).

For the failing `kubeflow-pipelines-sdk-execution-tests` test, would you have any hint how I could troubleshoot this one? I'm not familiar with the setup (details and such)

sdk/python/kfp/compiler/compiler_test.py
```diff
@@ -792,6 +792,55 @@ def my_pipeline() -> NamedTuple('Outputs', [
...
+
+        @dsl.pipeline()
+        def pipeline_1(run_pipeline_2: int = 1, run_pipeline_3: int = 1):
+            with dsl.Condition(run_pipeline_2 == run_pipeline_2):
+                with dsl.Condition(run_pipeline_2 == 1):
+                    pipeline_2()
+
+            with dsl.Condition(run_pipeline_3 == run_pipeline_3):
+                with dsl.Condition(run_pipeline_3 == 1):
+                    pipeline_3()
```
connor-mccarthy, line 819: Is this a minimal example? If not, could we reduce this to the minimal example required to reproduce the problematic behavior? This will help us ensure we don't regress and will reduce the likelihood that the test is removed as redundant or unclear in the future.

Also, consider renaming the test method to reflect the behavior under test and, if needed, adding a docstring to describe in more detail.

connor-mccarthy:
@stijntratsaertit, feel free to ignore `kfp-kubernetes-execution-tests`. These are unrelated to your changes.

connor-mccarthy (changes requested):

sdk/python/kfp/compiler/compiler_test.py
stijntratsaertit: Should be much better now, thanks for the push

sdk/python/kfp/compiler/compiler_test.py
```diff
@@ -882,6 +882,45 @@ def my_pipeline() -> NamedTuple('Outputs', [
...
+                component_refs = [[
+                    x.get('componentRef', {}).get('name')
+                    for x in task.values()
+                ]
+                                  for task in tasks]
+                all_component_refs = [
+                    item for sublist in component_refs for item in sublist
+                ]
+                counted_refs = collections.Counter(all_component_refs)
+                self.assertEqual(1, max(counted_refs.values()))
```
connor-mccarthy, line 922: In my local testing, this passes both before and after the changes. I also observe that the YAML is not different. Is it possible there is something missing from this test case?

stijntratsaertit: My bad - went to harsh on cutting components to a minimal version.

stijntratsaertit:
As I pulled master in to update the staleness of the PR, issues will show when attempting to fix the DCO check. What would be the correct approach to resolve this in my case?

PR #8232, opened 2022-09-01
feat(sdk): add compilation logic for google artifact types (support custom artifact types pt. 2)

connor-mccarthy:
**Description of your changes:**
Adds compilation logic to support custom artifact types (currently only `google.` types supported) in lightweight Python components and importer components. 

* Changes the in-memory representation of artifacts so that the full schema_title is preserved. This requires changes to v1 deserialization logic to maintain v1 read to v2 write idempotency.
* Preserves schema_title and schema_version on artifact class annotations for compilation.
* Only permit `system.` and `google.` artifact types
* Does not include lightweight Python component runtime handling of custom artifacts.

**Checklist:**
- [x] The title for your pull request (PR) should follow our title convention. [Learn more about the pull request title 
convention used in this repository](https://github.com/kubeflow/pipelines/blob/master/CONTRIBUTING.md#pull-request-title-convention).

PR #10005, opened 2023-09-18
chore(sdk): test observability, refactorings, and cleanup

connor-mccarthy:
Some opportunities for improvement I noticed while working on the component I/O changes. Some of these changes will improve the readability of the diff on that PR.

**Checklist:**
- [x] The title for your pull request (PR) should follow our title 
convention. [Learn more about the pull request title convention used in this repository](https://github.com/kubeflow/pipelines/blob/master/CONTRIBUTING.md#pull-request-title-convention).

### DIFF
--- a/sdk/python/kfp/dsl/component_factory.py
+++ b/sdk/python/kfp/dsl/component_factory.py
@@ -35,6 +35,7 @@ from kfp.dsl.types import type_annotations
 from kfp.dsl.types import type_utils
 
 _DEFAULT_BASE_IMAGE = 'python:3.7'
+SINGLE_OUTPUT_NAME = 'Output'
 
 
 @dataclasses.dataclass
@@ -209,7 +210,6 @@ def extract_component_interface(
     description: Optional[str] = None,
     name: Optional[str] = None,
 ) -> structures.ComponentSpec:
-    single_output_name_const = 'Output'
 
     signature = inspect.signature(func)
     parameters = list(signature.parameters.values())
@@ -280,10 +280,9 @@ def extract_component_interface(
         if passing_style in [
                 type_annotations.OutputAnnotation, type_annotations.OutputPath
         ]:
-            if io_name == single_output_name_const:
+            if io_name == SINGLE_OUTPUT_NAME:
                 raise ValueError(
-                    f'"{single_output_name_const}" is an invalid parameter name.'
-                )
+                    f'"{SINGLE_OUTPUT_NAME}" is an invalid parameter name.')
             io_name = _maybe_make_unique(io_name, output_names)
             output_names.add(io_name)
             if type_annotations.is_artifact_class(parameter_type):
@@ -359,8 +358,7 @@ def extract_component_interface(
                 output_spec = structures.OutputSpec(type=output_type_struct)
                 outputs[name] = output_spec
         elif signature.return_annotation is not None and signature.return_annotation != inspect.Parameter.empty:
-            output_name = _maybe_make_unique(single_output_name_const,
-                                             output_names)
+            output_name = _maybe_make_unique(SINGLE_OUTPUT_NAME, output_names)
             # Fixes exotic, but possible collision:
             #   `def func(output_path: OutputPath()) -> str: ...`
             output_names.add(output_name)
--- a/sdk/python/kfp/dsl/component_factory.py
+++ b/sdk/python/kfp/dsl/component_factory.py
@@ -401,7 +401,7 @@ def extract_component_interface(
 
     def parse_docstring_with_return_as_args(
             docstring: Union[str,
-                             None]) -> Union[docstring_parser.Docstring, None]:
+                             None]) -> Optional[docstring_parser.Docstring]:
         """Modifies docstring so that a return section can be treated as an
         args section, then parses the docstring."""
         if docstring is None:
@@ -661,7 +661,7 @@ def create_graph_component_from_func(
 def get_pipeline_description(
     decorator_description: Union[str, None],
     docstring: docstring_parser.Docstring,
-) -> Union[str, None]:
+) -> Optional[str]:
     """Obtains the correct pipeline description from the pipeline decorator's
     description argument and the parsed docstring.
 
--- a/sdk/python/kfp/dsl/types/type_annotations.py
+++ b/sdk/python/kfp/dsl/types/type_annotations.py
@@ -17,7 +17,7 @@ These are only compatible with v2 Pipelines.
 """
 
 import re
-from typing import List, Type, TypeVar, Union
+from typing import Any, List, Type, TypeVar, Union
 
 from kfp.dsl.types import artifact_types
 from kfp.dsl.types import type_annotations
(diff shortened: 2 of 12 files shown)


sdk/python/kfp/dsl/types/type_annotations.py
```diff
@@ -233,13 +233,37 @@ def is_artifact_class(artifact_class_or_instance: Type) -> bool:
...
-        type_var, '__args__') and type_annotations.is_artifact_class(
-            type_var.__args__[0])
-    return is_list_or_list_generic and contains_artifact
+    """Checks if an object is a list of list of artifacts annotation (e.g.,
+    List[Artifact], List[Dataset])"""
+    return is_generic_list(annotation) and issubclass_of_artifact(
+        get_inner_type(annotation))
+
+
+def get_inner_type(annotation: Any) -> Union[Any, None]:
```
chensun, line 245: nit: `Optional[Any]`?

sdk/python/kfp/dsl/types/type_annotations.py
```diff
@@ -233,13 +233,37 @@ def is_artifact_class(artifact_class_or_instance: Type) -> bool:
...
+    """Return the inner type of a generic annotation.
```
chensun, line 246: nit: Returns

sdk/python/kfp/dsl/types/type_annotations.py
```diff
@@ -233,13 +233,37 @@ def is_artifact_class(artifact_class_or_instance: Type) -> bool:
...
+    return None
+
+
+def issubclass_of_artifact(obj: Any) -> bool:
+    """Checks if an object is a class and a subclass of a dsl.Artifact."""
+    return type(obj) == type and issubclass(obj, artifact_types.Artifact)
+
+
+def is_generic_list(annotation: Any) -> bool:
+    # handles typing generics and built-in generics for python>=3.9
```
chensun, line 266: It seems to me his code also covers python version below 3.9, right?

sdk/python/kfp/dsl/component_factory.py
```diff
@@ -35,6 +35,7 @@
 from kfp.dsl.types import type_utils
 
 _DEFAULT_BASE_IMAGE = 'python:3.7'
+SINGLE_OUTPUT_NAME = 'Output'
```
chensun, line 38: Do you expect this to be used as a "public" constant? If not, consider prefix with an underscore.

https://google.github.io/styleguide/pyguide.html#3162-naming-conventions

sdk/python/test_data/pipelines/pipeline_with_google_artifact_type.py
```diff
@@ -38,7 +40,7 @@ def __init__(self, name: str, uri: str, metadata: dict) -> None:
         def path(self) -> str:
             return self.uri.replace('gs://', '/')
 
-    class VertexDataset:
+    class VertexDataset(dsl.Artifact):
```
chensun, line 43: Would `dsl.Dataset` make more sense here?

sdk/python/kfp/dsl/types/type_annotations_test.py
```diff
@@ -222,5 +224,89 @@ class NotArtifact:
         self.assertFalse(type_annotations.is_artifact_class(NotArtifact))
 
 
+class MyArtifact(dsl.Artifact):
+    pass
+
+
+class NotArtifact:
```
chensun, line 231: We currently treat arbitrary unknown types as artifact types. This seems to be a change of behavior?

sdk/python/kfp/dsl/types/type_annotations.py
connor-mccarthy: It does

sdk/python/kfp/dsl/component_factory.py
connor-mccarthy: Thanks for raising this. Curious what you think about this...

This is in a private module (none of its objects are included in the reference docs), so my sense is that a leading underscore would be redundant. This is the pattern we've tended toward in the v2 codebase (e.g., private functions tend to not feature a leading underscore).

image

I opted to not strip the leading underscore from `_DEFAULT_BASE_IMAGE` since it is currently unused outside of this module, so it is in a sense "private" to the module.

Not a strong feeling at all, so happy to make a change if desired.

sdk/python/test_data/pipelines/pipeline_with_google_artifact_type.py
connor-mccarthy: I think that might falsely suggest some form of artifact polymorphism. My preference is to inherit from the base class which contains the minimum set of functionality (`.path`, constructor) that we're interested in inheriting, without suggesting that it gets "dataset" properties.

sdk/python/kfp/dsl/types/type_annotations.py
connor-mccarthy: Thank you. Will try to catch these going forward.

Also updated this throughout the codebase.

sdk/python/kfp/dsl/types/type_annotations_test.py
connor-mccarthy: For the purposes of this PR and these unit tests, I think this is not a behavior change. `issubclass_of_artifact(NotArtifact)` should be `True` as it does not subclass artifact.

Separately, I think the behavior of "falling back" to an artifact may no longer be the current state.

For example, this pipeline does not compile:

```python
@dsl.component
def comp(string: 'Artifact'):
    print(string)
```

Based on the error message, I think this was lost some time ago in https://github.com/kubeflow/pipelines/pull/8232.

chensun: > `issubclass_of_artifact(NotArtifact)` should be `True` as it does not subclass artifact.

You meant `False`? More importantly, would `NotArtifact` compile into artifact or parameter? I think it's the former, right? If that's the case, I think there's an issue here.

In v1, the way to annotate an artifact would be

```python
@create_component_from_func
def comp(string: InputPath('Artifact')):
    with open(...) as f:
         print(f.read())
```
IIRC, while `Input/Output` accepts only subclasses of `dsl.Artifact`, `InputPath/OutputPath` should accepts everything with or without quotes. And we maintain this backward compatibility in v2. If this is no longer the current state, this seems to be an uncaught regression, and we should have a better understanding of the impact of this change. 

Maintaining v1 compatibility aside, it seems this behavior (whether it's new from this PR, or unintentionally changed in the past) is incompatible with our future plan of supporting common Python types (`pd.DataFrame` for example)? My impression was those types would be compiled into artifacts?

sdk/python/test_data/pipelines/pipeline_with_google_artifact_type.py
chensun: That's fine. We can revisit this later. IIRC, `dsl.Dataset` doesn't have any special properties other than being rendered with a different icon in Vertex. It's likely out of scope here, but we should sort out a clean path for our artifact usage so that we would have certain answer/guidline to questions like this.

sdk/python/kfp/dsl/component_factory.py
chensun: Makes sense to keep both as-is.

sdk/python/kfp/dsl/types/type_annotations_test.py
connor-mccarthy: Yes, thank you -- I meant `False`.

Neither of the following compiles successfully. Let's discuss this more synchronously.

```python
@dsl.component
def comp(string: dsl.InputPath('Artifact')):
    pass
```

```python
@dsl.component
def comp(string: dsl.InputPath(NotArtifact)):
    pass
```

Future looking, unless there's something I'm missing, I think that we are still able to add special support for `pd.DataFrame` (and similar) in the future, even if this is not currently supported.

chensun: Synced offline, and this is now resolved after renaming the class name used in test.

chensun (approved):
/lgtm
/approve

Thanks!

PR #10019, opened 2023-09-22
chore(sdk): update python to 3.9

davidspek:
Based on the work from https://github.com/kubeflow/pipelines/pull/10012, this PR updates all the sdk components to Python 3.9.

**Checklist:**

/cc @connor-mccarthy


sdk/python/setup.py
```diff
@@ -92,13 +92,14 @@ def read_readme() -> str:
...
         'Programming Language :: Python :: 3.11',
+        'Programming Language :: Python :: 3.12',
         'Topic :: Scientific/Engineering',
         'Topic :: Scientific/Engineering :: Artificial Intelligence',
         'Topic :: Software Development',
         'Topic :: Software Development :: Libraries',
         'Topic :: Software Development :: Libraries :: Python Modules',
     ],
-    python_requires='>=3.7.0,<3.12.0',
+    python_requires='>=3.7.0,<3.13.0',
```
connor-mccarthy, line 102: Since the upper bound Python version is decoupled from changing the recommended minimum Python version, let's make both changes in the same PR.

Additionally, this upper bound is a bit risky since we don't actually know that the _current_ KFP SDK on `master:HEAD` (and subsequent releases) will be compatible with Python 3.12, since it's not generally available yet nor tested on a pre-release version. Once Python 3.12 is released, users of KFP SDK 2.3.0 (the next release), for example, might try to upgrade to Python 3.12 and find it does not work.

sdk/python/test_data/README.md
```diff
@@ -11,6 +11,7 @@ These tests require golden snapshots of pipelines and components against with co
 
 ```bash
 for f in sdk/python/test_data/pipelines/*.py ; do echo "$f" && python3 "$f" ; done
+for f in sdk/python/test_data/pipelines/parallelfor_fan_in/*.py ; do echo "$f" && python3 "$f" ; done
```
connor-mccarthy, line 14: Thanks for catching this!

sdk/python/kfp/dsl/component_factory.py
```diff
@@ -147,7 +147,7 @@ def _get_packages_to_install_command(
                 install_parts=[
                     f'kfp=={kfp.__version__}',
                     '--no-deps',
-                    'typing-extensions>=3.7.4,<5; python_version<"3.9"',
+                    'typing-extensions>=3.7.4,<5; python_version<"3.13"',
```
connor-mccarthy, line 150: Can you explain why this is required? It seems typing extensions shouldn't be needed in versions greater than 3.9, but this says it is needed in Python 3.9-3.12 inclusive.

sdk/python/tests/compiler/testdata/artifact_passing_using_volume.yaml
```diff
@@ -115,7 +115,7 @@ spec:
                 pass
             with open(output_file, 'w') as f:
                 f.write(_output_serializers[idx](_outputs[idx]))
-      image: python:3.7
+      image: python:3.9
```
connor-mccarthy, line 118: These tests under `sdk/python/tests/` are associated with KFP SDK v1, so let's not bother to update them. This is undocumented knowledge, apologies. We're moving toward cleaning this up.

connor-mccarthy (changes requested):
Thanks, @DavidSpek! Just a few comments.

I think updating the test code (which is ultimately what the KFP SDK changes will boil down to) is great for setting an example for which Python version to use. I think there are a few other high impact changes we can make, if you're interested:

1) Updating the [user documentation](https://www.kubeflow.org/docs/components/pipelines/v2/components/lightweight-python-components/#base_image) to suggest not relying on the Python 3.7 default.
2) Updating the [reference documentation](https://kubeflow-pipelines.readthedocs.io/en/master/source/dsl.html#kfp.dsl.component) via the [doctrings](https://github.com/kubeflow/pipelines/blob/623e3094466485319bf1bdbd2585017172bf492e/sdk/python/kfp/dsl/component_decorator.py#L31) to suggest the same.
3) Add a compile time warning in the component decorator if users specify a Python 3.7 image. I recommend we not do this yet, since due to its intrusiveness, but keep it in mind as a future option.

Issue #10105, opened 2023-10-16
[feature] Improve Logging in the KFP components


DharmitD:
### Feature Area

/area backend

### What feature would you like to see?

We need to enable logging for KFP components (APIServer, Persistence Agent, Scheduled Workflow) through manifest parameters and inject through environment variables in all pods to enable debugging.
This would involve updating the main.go for each of these components, in order to have a log-level parameter that could either be set to info, error, debug, etc levels, and also updating respective components' dockerfiles.

### Is there a workaround currently?

---

Love this idea? Give it a 👍.

chensun:
> logging for KFP components (APIServer, Persistence Agent, Scheduled Workflow)

The cluster admin would have access to such logging, I don't think KFP users needs to debug through such information.

rimolive:
> > logging for KFP components (APIServer, Persistence Agent, Scheduled Workflow)
> 
> The cluster admin would have access to such logging, I don't think KFP users needs to debug through such information.

I think this issue needs clarification. The idea is to give users a way to easily specify the logging level for KFP components if needed. Instructions on how to do that are in the code, so either clarify in the KFP documentation how to set the logging level or add an env var in each KFP component Deployment to do that.

DharmitD:
@chensun following up on this thread, any thoughts / suggestions based on @rimolive 's explanation above?

Could we move this task to "ready" from "needs more info"?

PR #10288, opened 2023-12-05
Enable logging for KFP components

DharmitD:
Resolves https://github.com/kubeflow/pipelines/issues/10105

**Checklist:**


backend/Dockerfile.scheduledworkflow
```diff
@@ -46,4 +46,4 @@ COPY --from=builder /tmp/NOTICES /third_party/NOTICES
 
 ENV NAMESPACE ""
 
-CMD /bin/controller --logtostderr=true --namespace=${NAMESPACE}
+CMD /bin/controller --logtostderr=true --namespace=${NAMESPACE} --log-level=info
```
Tomcli, line 49: same here

backend/Dockerfile
```diff
@@ -82,4 +82,4 @@ RUN sed -E "s#/(blob|tree)/master/#/\1/${COMMIT_SHA}/#g" -i /config/sample_confi
 EXPOSE 8888
 
 # Start the apiserver
-CMD /bin/apiserver --config=/config --sampleconfig=/config/sample_config.json -logtostderr=true
+CMD /bin/apiserver --config=/config --sampleconfig=/config/sample_config.json -logtostderr=true --log-level=info
```
Tomcli, line 85: same here

backend/src/apiserver/main.go
```diff
@@ -29,6 +29,7 @@ import (
 	"strings"
 	"time"
 
+    log "github.com/sirupsen/logrus"
```
Tomcli, line 32: Seems like your Golang code indent is off and the backend are failing during pod start up. Can you lint the golang code and verify the backend can start without any issue? Thank you.

backend/Dockerfile.persistenceagent
```diff
@@ -50,4 +50,4 @@ ENV TTL_SECONDS_AFTER_WORKFLOW_FINISH 86400
 # NUM_WORKERS indicates now many worker goroutines
 ENV NUM_WORKERS 2
 
-CMD persistence_agent --logtostderr=true --namespace=${NAMESPACE} --ttlSecondsAfterWorkflowFinish=${TTL_SECONDS_AFTER_WORKFLOW_FINISH} --numWorker ${NUM_WORKERS}
+CMD persistence_agent --logtostderr=true --namespace=${NAMESPACE} --ttlSecondsAfterWorkflowFinish=${TTL_SECONDS_AFTER_WORKFLOW_FINISH} --numWorker ${NUM_WORKERS} --log-level=info
```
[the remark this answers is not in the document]
DharmitD: Done, updated to have the log level parameterized in Dockerfiles.

backend/src/apiserver/main.go
DharmitD: Done, rectified the indentation error and the backend tests seem to be passing now.

Tomcli (changes requested):

DharmitD:
@Tomcli @chensun e2e and upgrade tests are failing due to `ml-pipelines` pod rollback errors. Could you please take a look? Thanks.

backend/src/apiserver/main.go
```diff
@@ -77,6 +77,17 @@ func main() {
 		}
 	}
 
+    logLevel := os.Getenv("LOG_LEVEL")
```
Tomcli, line 80: I believe all the golang file are lint with tab not 4 white spaces. It still seems like your indent is a bit off.

backend/src/agent/persistence/main.go
```diff
@@ -18,6 +18,7 @@ import (
 	"flag"
 	"time"
 
+    "os"
```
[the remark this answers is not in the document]
DharmitD: Updated to correct indentation, could you please take a look?

backend/src/apiserver/main.go
DharmitD: Done, rectified the indentation errors.

backend/Dockerfile
```diff
@@ -82,4 +82,4 @@ RUN sed -E "s#/(blob|tree)/master/#/\1/${COMMIT_SHA}/#g" -i /config/sample_confi
 EXPOSE 8888
 
 # Start the apiserver
-CMD /bin/apiserver --config=/config --sampleconfig=/config/sample_config.json -logtostderr=true
+CMD /bin/apiserver --config=/config --sampleconfig=/config/sample_config.json -logtostderr=true --log-level=${LOG_LEVEL}
```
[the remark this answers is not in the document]
DharmitD: @Tomcli @chensun I updated manifests to have the variable in deployment files, but looks like the CI tests are failing due to APIServer, Persistence Agent and Scheduled Workflow pods crash looping. Do I need to add this var somewhere else in the manifests as well? A ConfigMap maybe?

Could the pods be crash looping due to some other reasons not related to changes in this PR?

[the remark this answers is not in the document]
Tomcli: @DharmitD I think you can add the new value in the configmap here with the default value `info`.
https://github.com/kubeflow/pipelines/blob/master/manifests/kustomize/base/installs/generic/pipeline-install-config.yaml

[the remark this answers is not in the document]
DharmitD: Done, added the var with the default value `info` in this configmap.

DharmitD:
> @DharmitD in my local environment I see your code is failing because there aren't any default log level value in the deployment. I think you would need to define a default log level value for the e2e test to pass.
> 
> ```
> time="2023-12-12T17:57:21Z" level=info msg="Location: UTC"
> flag provided but not defined: -log-level
> Usage of /bin/controller:
>   -alsologtostderr
>     	log to standard error as well as files
>   -clientBurst int
>     	Maximum burst for throttle from this client. (default 10)
>   -clientQPS float
>     	The maximum QPS to the master from this client. (default 5)
>   -kubeconfig string
>     	Path to a kubeconfig. Only required if out-of-cluster.
>   -log_backtrace_at value
>     	when logging hits line file:N, emit a stack trace
>   -log_dir string
>     	If non-empty, write log files in this directory
>   -logtostderr
>     	log to standard error instead of files
>   -master string
>     	The address of the Kubernetes API server. Overrides any value in kubeconfig. Only required if out-of-cluster.
>   -namespace string
>     	The namespace name used for Kubernetes informers to obtain the listers.
>   -stderrthreshold value
>     	logs at or above this threshold go to stderr
>   -v value
>     	log level for V logs
>   -vmodule value
>     	comma-separated list of pattern=N settings for file-filtered logging
> ```

Do you mean directly providing the default value in the deployment manifests instead of `$(LOG_LEVEL)`? Or, is there any .env file where you generally specify default values / environment variables that the manifests could utilize?

backend/src/agent/persistence/main.go
```diff
@@ -86,6 +86,17 @@ func main() {
 		log.Fatalf("Error building schedule clientset: %s", err.Error())
 	}
 
+	logLevel := os.Getenv("LOG_LEVEL")
```
Tomcli, line 89: So because here you are parsing the `LOG_LEVEL` from environment variable in your Golang code, you haven't define any new argument inputs for the controller startup function.

Tomcli: Flags for persistent agent are defined over here. You need to do it for every microservices.
https://github.com/kubeflow/pipelines/blob/master/backend/src/agent/persistence/main.go#L132-L150

Once the flag is parsed to logLevel, you can use the logLevel variable directly instead of loading it using `os.Getenv`

Tomcli: A better way to debug this is just to run the binary itself on your local machine, then you can catch the errors from the main and init functions.

DharmitD: > So because here you are parsing the `LOG_LEVEL` from environment variable in your Golang code, you haven't define any new argument inputs for the controller startup function.

Yes, the original thought behind this PR was to specify log levels using an environment variable with it's value specified via manifests. I believe just adding that and these changes to the go function and not passing it as an argument to the command should make this work. Is there a reason why we prefer having this value passed in as an argument?

DharmitD: Done, updated each of the components to have the logLevel value picked via flags.

Again, just like how there's inconsistency on libraries (glog/logrus for eg.) across component go files, the variable definition / flag creation bit is also different for each of the three components. We should probably have a follow up task to ensure consistency across all components.

Tomcli: Sorry @DharmitD I misunderstood your intend. You have the `--log-level=${LOG_LEVEL}` in your docker entrypoint so I thought you want to introduce the log level as a flag. 

I agree we should update all the backend components to use the same libraries and definitions to make things easier. Would you mind create that as an issue so we can pick it up when we have bandwidth? Thank you.

backend/Dockerfile
```diff
@@ -82,4 +83,4 @@ RUN sed -E "s#/(blob|tree)/master/#/\1/${COMMIT_SHA}/#g" -i /config/sample_confi
...
```
Tomcli, line 86: You need to update the docker entrypoint arguments to use the flag name you defined.
```suggestion
CMD /bin/apiserver --config=/config --sampleconfig=/config/sample_config.json -logtostderr=true --logLevel=${LOG_LEVEL}
```

backend/Dockerfile.persistenceagent
```diff
@@ -49,5 +49,6 @@ ENV TTL_SECONDS_AFTER_WORKFLOW_FINISH 86400
 
 # NUM_WORKERS indicates now many worker goroutines
 ENV NUM_WORKERS 2
+ENV LOG_LEVEL info
 
-CMD persistence_agent --logtostderr=true --namespace=${NAMESPACE} --ttlSecondsAfterWorkflowFinish=${TTL_SECONDS_AFTER_WORKFLOW_FINISH} --numWorker ${NUM_WORKERS}
+CMD persistence_agent --logtostderr=true --namespace=${NAMESPACE} --ttlSecondsAfterWorkflowFinish=${TTL_SECONDS_AFTER_WORKFLOW_FINISH} --numWorker ${NUM_WORKERS} --log-level=${LOG_LEVEL}
```
Tomcli, line 54:
```suggestion
CMD persistence_agent --logtostderr=true --namespace=${NAMESPACE} --ttlSecondsAfterWorkflowFinish=${TTL_SECONDS_AFTER_WORKFLOW_FINISH} --numWorker ${NUM_WORKERS} --logLevel=${LOG_LEVEL}
```

backend/Dockerfile.scheduledworkflow
```diff
@@ -45,5 +45,6 @@ COPY --from=builder /tmp/licenses.csv /third_party/licenses.csv
 COPY --from=builder /tmp/NOTICES /third_party/NOTICES
 
 ENV NAMESPACE ""
+ENV LOG_LEVEL info
 
-CMD /bin/controller --logtostderr=true --namespace=${NAMESPACE}
+CMD /bin/controller --logtostderr=true --namespace=${NAMESPACE} --log-level=${LOG_LEVEL}
```
Tomcli, line 50:
```suggestion
CMD /bin/controller --logtostderr=true --namespace=${NAMESPACE} --logLevel=${LOG_LEVEL}
```

Tomcli (changes requested):

backend/src/agent/persistence/main.go
DharmitD: Done, created an issue for this: https://github.com/kubeflow/pipelines/issues/10360

backend/src/crd/controller/scheduledworkflow/main.go
```diff
@@ -102,6 +113,7 @@ func initEnv() {
 func init() {
 	initEnv()
 
+	flag.StringVar(&logLevel, "", "Defines the log level for the application.")
```
Tomcli, line 116: Seems like you missed an argument according to the e2e test. I would recommend you try to run the docker image on your local machine to catch the start up errors so you don't have to wait for the e2e test to catch each bug.

```suggestion
	flag.StringVar(&logLevel, "logLevel", "", "Defines the log level for the application.")
```

DharmitD: Apologies, my bad. I had this change saved locally but it didn't get git push'ed. Updated now and the tests have passed as well. :)

Tomcli (commented):
thanks @DharmitD 
/lgtm

chensun (approved):
/approve

DharmitD:
Thank you so much @Tomcli @chensun !

PR #10302, opened 2023-12-12
feat(sdk): add subprocess task handler #localexecution

connor-mccarthy:
Add subprocess task handler to the task dispatcher.

There are basic tests, but will add more comprehensive E2E local execution tests once we support accessing outputs on locally executed tasks. There several other TODOs included throughout.

**Checklist:**

### DIFF
--- a/sdk/python/kfp/dsl/executor.py
+++ b/sdk/python/kfp/dsl/executor.py
@@ -191,7 +191,7 @@ class Executor:
             origin_type = getattr(annotation_type, '__origin__',
                                   None) or annotation_type
             # relax float-typed return to allow both int and float.
-            if origin_type in {int, float}:
+            if origin_type == float:
                 accepted_types = (int, float)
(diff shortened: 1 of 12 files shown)


sdk/python/kfp/dsl/executor.py
```diff
@@ -191,7 +191,7 @@ def handle_single_return_value(self, output_name: str, annotation_type: Any,
             origin_type = getattr(annotation_type, '__origin__',
                                   None) or annotation_type
             # relax float-typed return to allow both int and float.
-            if origin_type == float:
+            if origin_type in {int, float}:
```
chensun, line 194: This seems to be an orthogonal behavior change? And it seems this might be problematic to accept float when the int was expected?

connor-mccarthy: Yes. Thank you for catching that. This LOC was included in error.

I'll flag the correct solution to the problem this intended to solve in the next PR.

Issue #9689, opened 2023-06-29
[feature] remove hard-coded configs in KFP V2 launcher


thesuperzapper:
### Feature Area

/area backend

## Problem

Currently, we are hardcoding a few important configs in KFP V2's launcher and cache code.
This is making it difficult for distributions to support KFP v2 in a generic way.

__Launcher - The minio credentials secret is hard-coded as `mlpipeline-minio-artifact`:__

https://github.com/kubeflow/pipelines/blob/35f750703dcf4d1073a90708375459d1491085dd/backend/src/v2/objectstore/object_store.go#L292

__Launcher - The minio credentials secret keys are hard-coded as `accesskey` and `secretkey`:__

https://github.com/kubeflow/pipelines/blob/35f750703dcf4d1073a90708375459d1491085dd/backend/src/v2/objectstore/object_store.go#L324-L325

__Launcher - The minio endpoint is hard-coded as `minio-service.kubeflow:9000`:__

https://github.com/kubeflow/pipelines/blob/35f750703dcf4d1073a90708375459d1491085dd/backend/src/v2/objectstore/object_store.go#L291

__Cache - The `ml-pipeline` endpoint is hard-coded as `ml-pipeline.kubeflow:8887`:__

https://github.com/kubeflow/pipelines/blob/35f750703dcf4d1073a90708375459d1491085dd/backend/src/v2/cacheutils/cache.go#L26

## Solution

We want to allow the above configs to be changed from their defaults, preferably on a per-namespace basis.

I propose we add these configs to the `ConfigMap/kfp-launcher` which already exists in each profile namespace to set `defaultPipelineRoot`.

 For example, a new `ConfigMap/kfp-launcher` in a profile namespace might look like this:

```yaml
data:
  defaultPipelineRoot: "minio://mlpipeline/v2/artifacts"
 
  ## minio endpoint config
  minioEndpoint: "minio.example.com:9000"
 
  ## minio auth configs
  minioAuthSecret: "mlpipeline-minio-artifact"
  minioAuthSecretAccessKeyKey: "access_key"
  minioAuthSecretAccessKeyKey: "access_key"

  ## ml pipeline endpoint config
  mlPipelineEndpoint: "ml-pipeline.kubeflow:8887"
```

The question now becomes how to propagate these configs to the `kfp-launcher` container in each workflow Pod.
I think there are two options:

- OPTION 1: have the `kfp-launcher` container read the `ConfigMap/kfp-launcher`
[...]

PR #10319, opened 2023-12-14
feat(backend): preserve querystring in pipeline root (fixes #10318)

### LINKED ISSUE #10318: [backend] Query Strings in the Pipeline Root are ignored
To my understanding, using custom artifact stores in pipelines v2 requires configuring the pipelineRoot. 
According to the [blob documentation](https://gocloud.dev/howto/blob/#s3-compatible), an endpoint needs to be configured by using a queryString that gets appended to the pipelineRoot.

However, the kfp driver ignores query strings and assembles wrong bucket names. 

This issue effectively blocks using custom S3 artifact stores for pipelines v2.

### Environment

KFP 2.0.5.

### Steps to reproduce

Set a custom Pipeline Root, e.g. by setting the kfp-launcher's ConfigMap key `defaultPipelineRoot` to `s3://my-bucket/my-dir?endpoint=my.endpoint&s3ForcePathStyle=true`.

### Expected result

We'll see errors like 

> failed to execute component: Failed to open bucket "my-bucket": open bucket s3//my-bucket/my-dir?endpoint=my.endpoint&s3ForcePathStyle=true%2Fmy-task-name%2Fmy-artifact-name: invalid value for query parameter "s3ForcePathStyle": strconv.ParseBool: parsing true/my-task-name/my-artifact-name: invalid syntax

Also, after the driver has been executed, the wrongly concatenated pipelineRoot is visible in the workflow yaml.

---

TobiasGoerke:
The kfp-driver now recognizes query params and assembles bucket names correctly.

See #10318

### DIFF
--- a/backend/src/v2/driver/driver.go
+++ b/backend/src/v2/driver/driver.go
@@ -17,9 +17,7 @@ import (
 	"context"
 	"encoding/json"
 	"fmt"
-	"path"
 	"strconv"
-	"strings"
 	"time"
 
 	"github.com/golang/glog"
@@ -1062,8 +1060,9 @@ func provisionOutputs(pipelineRoot, taskName string, outputsSpec *pipelinespec.C
 		outputs.Artifacts[name] = &pipelinespec.ArtifactList{
 			Artifacts: []*pipelinespec.RuntimeArtifact{
 				{
-					// Split off any query strings first
-					Uri:      fmt.Sprintf("%s/%s", strings.TrimRight(strings.Split(pipelineRoot, "?")[0], "/"), path.Join(taskName, name)),
+					// Do not preserve the query string for output artifacts, as otherwise
+					// they'd appear in file and artifact names.
+					Uri:      metadata.GenerateOutputURI(pipelineRoot, []string{taskName, name}, false),
 					Type:     artifact.GetArtifactType(),
 					Metadata: artifact.GetMetadata(),
 				},
--- a/backend/src/v2/driver/driver.go
+++ b/backend/src/v2/driver/driver.go
@@ -17,9 +17,7 @@ import (
 	"context"
 	"encoding/json"
 	"fmt"
-	"path"
 	"strconv"
-	"strings"
 	"time"
 
 	"github.com/golang/glog"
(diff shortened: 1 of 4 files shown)


TobiasGoerke:
Using custom S3 endpoints now works for me. 

Ready for review!

Tomcli (commented):
/lgtm

TobiasGoerke:
Thanks @Tomcli.
I pushed another minor change, see [https://github.com/TobiasGoerke/pipelines/pull/1/files](https://github.com/TobiasGoerke/pipelines/pull/1/files).

This causes filenames to be stored without the query string and the artifacts in the pipeline yaml don't contain the connection string. 
It's more a cosmetic change but also lets the pipelines ui correctly download the artifacts.

backend/src/v2/driver/driver.go
```diff
@@ -1078,12 +1079,6 @@ func provisionOutputs(pipelineRoot, taskName string, outputsSpec *pipelinespec.C
 	return outputs
 }
 
-func generateOutputURI(root, artifactName string, taskName string) string {
-	// we cannot path.Join(root, taskName, artifactName), because root
```
Tomcli, line 1082: Since now you new function is similar to the old `generateOutputURI` function, it make sense to bring these comments to the changes you have at line 1065 or just update the `generateOutputURI` function.

Tomcli (commented):
thanks @TobiasGoerke 
/lgtm

thesuperzapper:
@chensun @Tomcli @connor-mccarthy is there any chance we can cut a 2.0.6 patch release that has this PR included?

It's pretty important, because external object stores can't be used with KFP v2 without query parameters (in the pipeline root URI).

This will be less important if we ever get https://github.com/kubeflow/pipelines/issues/9689 resolved, but that's a longer term fix.

TobiasGoerke:
> > Thanks @Tomcli. I pushed another minor change, see [here](https://github.com/TobiasGoerke/pipelines/pull/1/files).
> > This causes filenames to be stored without the query string and the artifacts in the pipeline yaml don't contain the connection string. It's more a cosmetic change but also lets the pipelines ui correctly download the artifacts.
> 
> How to use custom S3 endpoints ？Can you give an example? @TobiasGoerke

For v2 pipelines, you'll need to set the `kfp-launcher` ConfigMap's `data.defaultPipelineRoot` to your desired endpoint. Depending on your endpoint, you may now also need to set some env vars for authentication in your pipeline pods and artifacts proxies etc.

PR #10325, opened 2023-12-15
feat(sdk): add local execution output collection #localexecution

connor-mccarthy:
Adds the code that collects and surfaces the outputs from locally executed tasks on the `task.output` and `task.outputs['<key>']`.

This constitutes E2E local execution for the subprocess task handler. For this reason, we also now include E2E tests.

Forthcoming PRs will add the container task handler, logging, and a few more changes.

**Checklist:**

### DIFF
--- a/sdk/python/kfp/dsl/types/artifact_types.py
+++ b/sdk/python/kfp/dsl/types/artifact_types.py
@@ -89,7 +89,8 @@ class Artifact:
             return _MINIO_LOCAL_MOUNT_PREFIX + self.uri[len('minio://'):]
         elif self.uri.startswith('s3://'):
             return _S3_LOCAL_MOUNT_PREFIX + self.uri[len('s3://'):]
-        return None
+        # uri == path for local execution
+        return self.uri
 
     def _set_path(self, path: str) -> None:
         self.uri = convert_local_path_to_remote_path(path)
--- a/sdk/python/kfp/local/e2e_test.py
+++ b/sdk/python/kfp/local/e2e_test.py
@@ -1,0 +1,382 @@
+# Copyright 2023 The Kubeflow Authors
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#      http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+"""E2E local execution tests.
+
+These contain tests of various component definitions/types, tested on each local runner type + configurations.
+
+These can be thought of as local runner conformance tests. The test results should be the same irrespective of the runner.
+"""
+import io
+import sys
+from typing import NamedTuple
+import unittest
+
+from absl.testing import parameterized
+from kfp import dsl
+from kfp import local
+from kfp.dsl import Artifact
+from kfp.dsl import Dataset
+from kfp.dsl import Output
+from kfp.local import testing_utilities
+
+# NOTE when asserting on task.output or task.outputs[<key>]
+# since == is overloaded for dsl.Condition, if local execution is not
+# "hit", then actual will be a channel and actual == expected evaluates
+# to ConditionOperation. Since ConditionOperation is truthy,
+# this may result in a false negative test result. For this reason,
+# we perform an isinstance check first.
+
+ALL_RUNNERS = [
+    (local.SubprocessRunner(use_venv=False),),
+    (local.SubprocessRunner(use_venv=True),),
+]
+
+
+@parameterized.parameters(ALL_RUNNERS)
+class TestLightweightPythonComponentLogic(
+        testing_utilities.LocalRunnerEnvironmentTestCase):
+
+    def test_str_input(self, runner):
+        local.init(runner=runner)
+
+        @dsl.component
+        def identity(x: str) -> str:
+            return x
+
+        actual = identity(x='hello').output
+        expected = 'hello'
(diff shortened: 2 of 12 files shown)


sdk/python/kfp/dsl/types/artifact_types.py
```diff
@@ -89,7 +89,8 @@ def _get_path(self) -> Optional[str]:
             return _MINIO_LOCAL_MOUNT_PREFIX + self.uri[len('minio://'):]
         elif self.uri.startswith('s3://'):
             return _S3_LOCAL_MOUNT_PREFIX + self.uri[len('s3://'):]
-        return None
+        # uri == path for local execution
+        return self.uri
```
connor-mccarthy, line 93: @chensun, WDYT about this change?

It is a behavior change. If a user depends on `.path` being `None`, their code's behavior may change. This doesn't affect code paths that write to `.path` (since you cannot write to `None`), but it would affect code like this:

```python
@dsl.component
def my_comp(a: Artifact):
    if a.path is None: # prev evaluates to True, now would evaluate to False
        persist('some artifact contents', a.uri)
    else:
        persist('some artifact contents', a.path)
```

The alternative is to set some sort of special flag on both input and output artifacts to allow `.path` to return the "local uri" when executing locally:

```python
# implementation code
for k, maybe_artifact in user_kwargs.items():
    if running_locally and isinstance(maybe_artifact, dsl.Artifact):
        maybe_artifact._use_uri_as_path = True

# then change `.path` to respect the `._use_uri_as_path` flag
```

WDYT? The reason for not immediately proceeding with the alternative implementation above is that:
- The current implementation's behavior may be acceptable in a minor version bump with release notes.
- The solution above is somewhat hacky and it's challenging to exhaustively reason about whether it will play nicely with user local execution code or, eventually, input artifact constants.

I can go either way on this.

sdk/python/kfp/local/executor_output_utils.py
```diff
@@ -0,0 +1,213 @@
...
+        executor_output=executor_output,
+        component_spec=component_spec,
+    )
+
+    # collect parameter outputs from executor output
+    output_parameters = {
+        param_name: pb2_value_to_python(value)
+        for param_name, value in executor_output.parameter_values.items()
+    }
+    # process the special case of protobuf ints
```
connor-mccarthy, line 76: This is related to [this previous comment](https://github.com/kubeflow/pipelines/pull/10302#discussion_r1428394031). The current implementation in this PR is the correct way to handle this, IMO.

sdk/python/kfp/dsl/types/artifact_types.py
chensun: > If a user depends on .path being None, their code's behavior may change.

This sounds fine to me. I don't perceive `.path` being `None` an existing scenario until very recent changes for local execution.

sdk/python/kfp/local/e2e_test.py
```diff
@@ -0,0 +1,382 @@
...
+    def test_str_input(self, runner):
+        local.init(runner=runner)
+
+        @dsl.component
+        def identity(x: str) -> str:
+            return x
+
+        actual = identity(x='hello').output
+        expected = 'hello'
+        self.assertIsInstance(actual, str)
```
chensun, line 59: nit: seems a bit redundant to me as the equality test next should cover this as well. 
But I don't have strong opinion, feel free to keep it.

sdk/python/kfp/local/executor_output_utils.py
```diff
@@ -0,0 +1,213 @@
...
+    outputs_typed_int = [
+        output_param_name for output_param_name, parameter_spec in
+        component_spec.output_definitions.parameters.items()
+        if parameter_spec.parameter_type ==
+        pipeline_spec_pb2.ParameterType.ParameterTypeEnum.NUMBER_INTEGER
+    ]
+    output_parameters = cast_ints_as_floats_to_ints(
```
chensun, line 83: nit: naming of `cast_ints_as_floats_to_ints` is a bit hard to follow. one possibly cleaner way is to combine this and the code block above to have some function like `casting_protobuf_numbers`.

chensun (commented):
/lgtm

only a few nitpicks, feel free to ignore

sdk/python/kfp/local/e2e_test.py
connor-mccarthy: I added [a comment](https://github.com/kubeflow/pipelines/blob/4e3699e4dcbbe3a84204213c79c71d9c66c4d418/sdk/python/kfp/local/e2e_test.py#L33-L38) to the top of the file on why this is needed. Added to the top, since it applies for every assertion.

sdk/python/kfp/local/executor_output_utils.py
connor-mccarthy: Updated in 958362c3353224d4bda0ad588ede1f8a406d60b3.

I also noticed that this casting is needed on the input side. [3f86da7](https://github.com/kubeflow/pipelines/pull/10325/commits/3f86da7b1a53e25c6b1c234a2460e8df084fef88) adds casting when going from the ExecutorInput to JSON. This is consistent with the remote BE behavior.

This is needed because of the case like:

```python
@dsl.component
def my_comp(x: int) -> int:
    return x
```

If `x` goes in as a `float` (default representation in protobuf Value), it will be returned as a `float` and throw a runtime exception, since it expects an `int` return. Also related to [previous comment](https://github.com/kubeflow/pipelines/pull/10302#discussion_r1428394031).

chensun (approved):
/lgtm
/approve

Thanks!

connor-mccarthy:
Merging per [prior LGTM](https://github.com/kubeflow/pipelines/pull/10325#pullrequestreview-1785052799). Changes since the LGTM include rebasing and removing 2 duplicate LOC causing a false positive test failure.

`kfp-kubernetes-execution-tests` test failures are independent of this PR. The pipeline is hanging in a "Running" state, despite inner tasks all succeeding. Investigating separately.

PR #10394, opened 2024-01-12
feat(frontend): Automatically fit DAG canvas to elements on change

LotemAm:
Automatically fit DAG canvas to elements on change

### DIFF
--- a/frontend/src/pages/v2/DagCanvas.tsx
+++ b/frontend/src/pages/v2/DagCanvas.tsx
@@ -44,6 +44,8 @@ export default function DagCanvas({
   setFlowElements,
 }: DagCanvasProps) {
   const instanceRef = useRef<OnLoadParams | null>(null);
+  const prevLayer = useRef(layers);
+
   const onLoad = (reactFlowInstance: OnLoadParams) => {
     reactFlowInstance.fitView();
     instanceRef.current = reactFlowInstance;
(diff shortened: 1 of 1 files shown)


frontend/src/pages/v2/DagCanvas.tsx
```diff
@@ -59,6 +61,12 @@ export default function DagCanvas({
     }
   });
 
+  useEffect(() => {
+    if (instanceRef.current) {
+      instanceRef.current.fitView();
+    }
+  }, [elements]);
```
zijianjoy, line 68: Can you share a video snippet how this might work in action? I am afraid about Drag-and-drop or element change can cause unexpected movement to the canvas.

LotemAm: Before:

https://github.com/kubeflow/pipelines/assets/19858141/9e3a318a-69e6-4880-a238-42bfc948fc44



After:
https://github.com/kubeflow/pipelines/assets/19858141/9fd0074f-3001-40a9-b176-0ec42b35c749

LotemAm: I added a check to make sure this happens only when the layers change, because I saw a timed GET request causing a re-render

zijianjoy: Thank you @LotemAm , I don't think we should automatically adjust the canvas when a node is dragged. There is a button on the bottom left to manually do so. I would like to keep this canvas focus functionality initiated by users.

LotemAm: @zijianjoy In my last commit I have changed the behavior to have the automatic fit happen only on layer change, so it no longer happens when node is dragged

zijianjoy: Thank you for the change! Approving the PR.

Issue #9366, opened 2023-05-09
[sdk] Bug when trying to iterate a list of dictionaries with ParallelFor


lucasvbalves:
### Environment

*  KFP SDK version: 
kfp==2.0.0b16
*  All dependencies version:
kfp==2.0.0b16
kfp-pipeline-spec==0.2.2
kfp-server-api==2.0.0b1

### Steps to reproduce

When running the code snippet below the following error is raised:

`kfp.components.types.type_utils.InconsistentTypeException: Incompatible argument passed to the input 'val_a' of component 'add': Argument type 'STRING' is incompatible with the input type 'NUMBER_INTEGER'`

```python
@dsl.component()
def add(val_a: int, val_b: int) -> int:
    return val_a + val_b


@dsl.pipeline()
def model_training_pipeline() -> None:
    with dsl.ParallelFor(
        items=[{"a": 1, "b": 10}, {"a": 2, "b": 20}], parallelism=1
    ) as item:
        task = add(val_a=item.a, val_b=item.b)


compiler.Compiler().compile(
    pipeline_func=model_training_pipeline, package_path="/app/pipeline.yaml"
)
```

### Expected result

According to the [ParallelFor documentation](https://kubeflow-pipelines.readthedocs.io/en/2.0.0b16/source/dsl.html#kfp.dsl.ParallelFor), the code sample above should compile without errors. The `add` component should receive the values of the dictionaries as integer arguments.

### Materials and Reference

The code snippet below is a modification of the code snippet above, changing the `add` component to accept string arguments. 

```python
@dsl.component()
def add(val_a: str, val_b: str) -> int:
    return int(val_a) + int(val_b)


@dsl.pipeline()
def model_training_pipeline() -> None:
    with dsl.ParallelFor(
        items=[{"a": 1, "b": 10}, {"a": 2, "b": 20}], parallelism=1
    ) as item:
        task = add(val_a=item.a, val_b=item.b)


compiler.Compiler().compile(
    pipeline_func=model_training_pipeline, package_path="/app/pipeline.yaml"
)
```

The pipeline compiles without errors with this modification, however it fails to run in Google Vertex Pipelines. The `add` component doesn't even run and throws the following error in the UI:

`Failed to evaluate the expression with error: INVALID_ARGUMENT: Failed to parseJson from string.; Failed to evaluate the parameter_expression_selector.`

As the component's code is not even executed, it seems that the problem occurs when executing the DAG.

[...]

PR #10414, opened 2024-01-19
Fix type issues in dict for ParallelFor

rickyxie0929:
fix(sdk): fixes type issues for dict in ParallelFor Fixes https://github.com/kubeflow/pipelines/issues/9366

**Checklist:**

### DIFF
--- a/components/google-cloud/setup.py
+++ b/components/google-cloud/setup.py
@@ -69,10 +69,10 @@ setuptools.setup(
         ] + [
             "commonmark==0.9.1",
             "autodocsumm==0.2.9",
-            "sphinx==5.0.2",
+            "sphinx>=5.0.2,<6.0.0",
             "sphinx-immaterial==0.9.0",
-            "sphinx-rtd-theme==1.0.0",
-            "m2r2==0.3.3",
+            "sphinx-rtd-theme==2.0.0",
+            "m2r2==0.3.3.post2",
             "sphinx-notfound-page==0.8.3",
         ],
     },


sdk/python/kfp/dsl/for_loop.py
```diff
@@ -64,6 +64,25 @@ def _get_subvar_type(type_name: str) -> Optional[str]:
     return match['value_type'].lstrip().rstrip() if match else None
 
 
+def _get_first_element_type(item_list: ItemList) -> str:
```
connor-mccarthy, line 67: FYI, no action needed.

The `_` prefix makes sense here since it follows the pattern in the module, but FYI that this mechanism for declaring a symbol private is not needed since it is already in a private module.

We know this is a private module because the public modules are only those included in our [reference docs](https://kubeflow-pipelines.readthedocs.io/en/master/index.html):

image

sdk/python/kfp/dsl/for_loop.py
```diff
@@ -64,6 +64,25 @@ def _get_subvar_type(type_name: str) -> Optional[str]:
...
+    """Returns the type of the first element of ItemList.
+
+    Args:
+        item_list: It is the a list which for now the type is Union[int, float, str, Dict[str, Any]]]
+    Returns:
+        - A string representing the type of the first element (e.g., "int", "Dict[str, int]").
```
connor-mccarthy, line 73:
```suggestion
        A string representing the type of the first element (e.g., "int", "Dict[str, int]").
```

sdk/python/kfp/dsl/for_loop.py
```diff
@@ -180,7 +199,7 @@ def from_raw_items(
         return LoopArgument(
             items=raw_items,
             name_code=name_code,
-            channel_type=type(raw_items[0]).__name__,
+            channel_type=_get_first_element_type(raw_items),
```
connor-mccarthy, line 202: Can you add an E2E compilation test to [this directory](https://github.com/kubeflow/pipelines/tree/master/sdk/python/test_data/pipelines) and [this config](https://github.com/kubeflow/pipelines/blob/5b7f67acdcbd81d612a3deb39823f28ac6a56c6e/sdk/python/test_data/test_data_config.yaml#L19) to demonstrate that this solves the specific `dsl.ParallelFor` issue?

To see how these files/config are used, take a look at the test entrypoint [here](https://github.com/kubeflow/pipelines/blob/5b7f67acdcbd81d612a3deb39823f28ac6a56c6e/sdk/python/kfp/compiler/read_write_test.py#L166).

sdk/python/kfp/dsl/for_loop.py
```diff
@@ -180,7 +199,7 @@ def from_raw_items(
...
         )
```
connor-mccarthy, line 204: Can you update the [release notes](https://github.com/kubeflow/pipelines/blob/master/sdk/RELEASE.md) to include this fix?

sdk/python/kfp/dsl/for_loop.py
```diff
@@ -64,6 +64,25 @@ def _get_subvar_type(type_name: str) -> Optional[str]:
...
+    Returns:
+        - A string representing the type of the first element (e.g., "int", "Dict[str, int]").
+    """
+    first_element = item_list[0]
+    if isinstance(first_element, dict):
+        key_type = type(list(
+            first_element.keys())[0]).__name__  # Get type of first key
+        value_type = type(list(
+            first_element.values())[0]).__name__  # Get type of first value
+        return f'dict[{key_type}, {value_type}]'
```
connor-mccarthy, line 81: What you have works great. Question: WDYT about making this `typing.Dict` instead of `dict`?

I notice for this pipeline...

```python
@dsl.component
def dict_printer(x: Dict[str, int]) -> Dict[str, int]:
    return x


@dsl.pipeline
def my_pipeline(a: Dict[str, int]):
    dict_printer(x=a)
```

...the value of the channel holding `a` is type `"typing.Dict[str, int]"`. It seems like the logic downstream of your changes is robust to both `"typing.Dict[str, int]"` and `"dict[str, int]"`, but perhaps it would be good to be consistent?

I'll defer to your preference/understanding, since both are correct implementations.

sdk/python/kfp/dsl/for_loop.py
```diff
@@ -64,6 +64,25 @@ def _get_subvar_type(type_name: str) -> Optional[str]:
...
+    else:
+        return type(first_element).__name__
```
connor-mccarthy, line 83: I think there are two other cases we need to handle:

1) The dict is outputted by an upstream component
```python
@dsl.component
def identity(x: int) -> int:
    return x

@dsl.component
def dict_maker() -> Dict[str, int]:
    return [{'a': 1}, {'b': 2}]

@dsl.pipeline
def my_pipeline():
    t = dict_maker()
    with dsl.ParallelFor(t.output) as item:
        identity(x=item.a)
```

2) Same as above, but the component that outputs a dict is returned from a component
```python
@dsl.component
def identity(x: int) -> int:
    return x

@dsl.component
def dict_maker() -> Dict[str, int]:
    return [{'a': 1}, {'b': 2}]


from kfp import compiler
from kfp import components

compiler.Compiler().compile(dict_maker, 'output.yaml')
loaded_dict_maker = components.load_component_from_file('output.yaml')

@dsl.pipeline
def my_pipeline():
    t = loaded_dict_maker()
    with dsl.ParallelFor(t.output) as item:
        identity(x=item.a)
```

In this case, the inner types to the Dict generic (`str` and `int`) are lost during compilation. See the output definition for the compiled component:

```textproto
parameters {
  key: "Output"
  value {
    parameter_type: STRUCT
  }
}
```

This means we're limited in our ability to provide robust type checking once we reload this component. We should be forgiving and let the user compile an erroneous pipeline if they attempt to, rather than block their ability to do this at all (current state). Worst case is that it will fail on the first attempt to run the pipeline and they'll realize the type error (similar to normal Python).

connor-mccarthy (changes requested):
Thanks, @rickyxie0929! This looks great. Just some nitpicks and two other cases I thought of that we'll need to handle.

sdk/python/kfp/dsl/for_loop.py
rickyxie0929: For now the ItemList is `ItemList = List[Union[int, float, str, Dict[str, Any]]]`. I changed the `isinstance(first_element, dict)` to `isinstance(first_element, Dict)` to make it consistent.

sdk/python/kfp/dsl/for_loop.py
rickyxie0929: Synced offline. Added the fix for the case 2 which is the dict is outputted by an upstream component.

PR #10416, opened 2024-01-21
feat(kubernetes_platform): Update kubernetes_platform go package to include imagePullPolicy.

revit13:
Description of your changes:
Part of https://github.com/kubeflow/pipelines/issues/9714 and https://github.com/kubeflow/pipelines/issues/9980

Update kubernetes_platform go package to include image pull policy. Once this is ready, I will also update the backend driver, compiler, and SDK with this new feature.

Checklist:

- [x] The title for your pull request (PR) should follow our title convention. [Learn more about the pull request title convention used in this repository](https://github.com/kubeflow/pipelines/blob/master/CONTRIBUTING.md#pull-request-title-convention).

Issue #9980, opened 2023-09-13
[feature] Enable Setting Image Pull Policy in V2 SDK


PhilippeMoussalli:
### Feature Area

/area sdk

### What feature would you like to see?
KFP V1 supported setting the image pull policy [link](https://github.com/kubeflow/pipelines/blob/4bee3d8dc2ee9c33d87e1058bac2a94d899dd4a5/sdk/python/kfp/deprecated/dsl/_container_op.py#L518)

This feature is currently not available in KFP V2

### What is the use case or pain point?  

We used to set image pull policy to always in V1 to avoid cases when outdated images are pulled. 

### Is there a workaround currently?  

No current workaround.  

---

Love this idea? Give it a 👍.

rimolive:
cc @Tomcli @chensun I believe this issue can be closed as #10416 is merged. Can you confirm?

Tomcli:
Yes, #10417 should close this issue.
/close

Issue #9714, opened 2023-07-07
[feature] Editing Pod's spec.


stefanondisponibile:
### Feature Area

/area sdk

### What feature would you like to see?

I would like to understand what is the plan or the best practices for customizing the the Pod or container spec using `kfp`.
[Is this feature going to be deprecated together with the rest of the v1 sdk](https://github.com/search?q=repo%3Akubeflow%2Fpipelines+%22add_pod_annotation%22&type=code)?

### What is the use case or pain point?

In my case, I'm trying to add some _annotations_ (e.g. like it's defined [here](https://github.com/kubeflow/pipelines/blob/3b6201779c7db58c6cbc5fae1e02cac5420280af/samples/core/pipeline_transformers/pipeline_transformers.py#L27)). The pain point is that I don't feel like using a deprecated feature.

Having some control over the final manifest is critical, especially in cases where you need sidecars, annotations, etc. The lack of this flexibility makes it difficult to deal with critical requirements such as secrets handling or cluster-specific customizations.

[This](https://kubeflow-pipelines.readthedocs.io/en/sdk-2.0.1/source/dsl.html#kfp.dsl.PipelineTask) in particular:

> (`PipelineTask`) Replaces ContainerOp from kfp v1. Holds operations available on a task object, such as .after(), .set_memory_limit(), .enable_caching(), etc.

That `etc` is critical, we'll users be able to manipulate the final manifest? e.g. `set_annotations`?

### Is there a workaround currently?

Not quite? Other than using deprecated APIs?

Related issues:

* #4020 

---

Love this idea? Give it a 👍.

PR #10417, opened 2024-01-21
feat(Backend + SDK): Update kfp backend and kubernetes sdk to support ImagePullPolicy

revit13:
Description of your changes:
Part of Part of https://github.com/kubeflow/pipelines/issues/9714 and https://github.com/kubeflow/pipelines/issues/9980

Update kfp backend and kubernetes sdk to support image pull policy.
To be rebased and update go.mod after #10416 

Checklist:

 The title for your pull request (PR) should follow our title convention. [Learn more about the pull request title convention used in this repository](https://github.com/kubeflow/pipelines/blob/master/CONTRIBUTING.md#pull-request-title-convention).


Tomcli:
rebase and update go.mod after #10416

Tomcli (changes requested):
@revit13 Can you update the go.mod and the related files like in this PR? Thanks.
https://github.com/kubeflow/pipelines/pull/10427#pullrequestreview-1868388426

backend/src/v2/driver/driver.go
```diff
@@ -475,6 +475,12 @@ func extendPodSpecPatch(
 		podSpec.Containers[0].VolumeMounts = append(podSpec.Containers[0].VolumeMounts, volumeMounts...)
 	}
 
+	// Get image pull policy
+	if kubernetesExecutorConfig.GetImagePullPolicy() != "" {
```
Tomcli, line 479: Maybe we can also set kubernetesExecutorConfig.GetImagePullPolicy() as a variable, then check whether the value is one of ['Always', 'Never', 'IfNotPresent']?

There could be a case where someone can compose the IR without the SDK, so we also wants to check the value is valid before passing into the pod spec. If the value is not valid, we should raise an error instead.

revit13: @Tomcli Done. Thanks

chensun:
@revit13 Can you please rebase and resolve the conflicts?

kubernetes_platform/python/kfp/kubernetes/image.py
```diff
@@ -38,11 +38,33 @@ def set_image_pull_secrets(
...
 
     return task
+
+
+def set_image_pull_policy(task: PipelineTask, policy: str) -> PipelineTask:
+    """Set image pull policy for the container.
+
+    Args:
+        task: Pipeline task.
+        policy: One of `Always`, `Never`, `IfNotPresent`.
```
chensun, line 57: nit: using enum could be less error-prone, and also works with IDE code auto-complete. No strong preference though.

chensun (approved):
/lgtm
/approve

Hammad-NobleAI:
Putting this comment here... we need this facility through the Python API. Do we have any idea when the next release might be?

PR #10423, opened 2024-01-23
feat(sdk): support local pipeline execution

connor-mccarthy:
Enables running a pipeline locally:
```python
from kfp import dsl
from kfp import local

local.init(runner=local.SubprocessRunner())


@dsl.component
def add(a: int, b: int) -> int:
    return a + b


@dsl.pipeline
def my_pipeline(num: int) -> int:
    t1 = add(a=num, b=2)
    t2 = add(a=t1.output, b=3)
    return t2.output


task = my_pipeline(num=1)
print(f'Got result: {task.output}')
```

Currently:
- Only sequential pipelines are supported. Control flow features (`dsl.Condition`, `dsl.ParallelFor`, and `dsl.ExitHandler`) are not supported yet.
- `dsl.importer`is not supported yet.
- Pipeline-in-pipeline is not supported yet.
- Passing data via f-strings is not supported.

**Checklist:**

### DIFF
--- a/sdk/python/kfp/dsl/pipeline_task.py
+++ b/sdk/python/kfp/dsl/pipeline_task.py
@@ -28,6 +28,7 @@ from kfp.dsl import placeholders
 from kfp.dsl import structures
 from kfp.dsl import utils
 from kfp.dsl.types import type_utils
+from kfp.local import pipeline_orchestrator
 from kfp.pipeline_spec import pipeline_spec_pb2
 
 _register_task_handler = lambda task: utils.maybe_rename_for_k8s(
@@ -190,13 +191,20 @@ class PipelineTask:
         from kfp.local import task_dispatcher
 
         if self.pipeline_spec is not None:
-            raise NotImplementedError(
-                'Local pipeline execution is not currently supported.')
-
-        self._outputs = task_dispatcher.run_single_task(
-            pipeline_spec=self.component_spec.to_pipeline_spec(),
-            arguments=args,
-        )
+            self._outputs = pipeline_orchestrator.run_local_pipeline(
+                pipeline_spec=self.pipeline_spec,
+                arguments=args,
+            )
+        elif self.component_spec is not None:
+            self._outputs = task_dispatcher.run_single_task(
+                pipeline_spec=self.component_spec.to_pipeline_spec(),
+                arguments=args,
+            )
+        else:
+            # user should never hit this
+            raise ValueError(
+                'One of pipeline_spec or component_spec must not be None for local execution.'
+            )
         self.state = TaskState.FINAL
 
     @property
--- a/sdk/python/kfp/local/dag_orchestrator.py
+++ b/sdk/python/kfp/local/dag_orchestrator.py
@@ -1,0 +1,316 @@
+# Copyright 2024 The Kubeflow Authors
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#      http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+"""Code for locally executing a DAG within a pipeline."""
+from typing import Any, Dict, Optional, Tuple
+
+from kfp.local import config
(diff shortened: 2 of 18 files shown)


sdk/python/kfp/local/dag_orchestrator.py
```diff
@@ -0,0 +1,316 @@
...
+                    key,
+                    output,
+                )
+        else:
+            raise ValueError(f'Got unknown task status: {task_status.name}')
+
+    return status.Status.SUCCESS, None
+
+
+def bind_defaults_to_dag_arguments(
```
chensun, line 121: nit: this name is a bit misleading as function binds both user inputs and default values (when user inputs are not provided).

sdk/python/kfp/local/dag_orchestrator.py
```diff
@@ -0,0 +1,316 @@
...
+) -> None:
+    """Validates that an ExecutorSpec is a supported executor for local
+    execution.
+
+    Args:
+        executor: The ExecutorSpec to validate.
+    """
+    if executor.WhichOneof('spec') == 'importer':
+        raise ValueError(
+            "Importer is not yet supported by local pipeline execution. Found 'dsl.importer' task in pipeline."
```
chensun, line 224: nit: per message this should be `NotImplementedError`?

sdk/python/kfp/local/dag_orchestrator.py
```diff
@@ -0,0 +1,316 @@
...
+        kind = parameter_selector_spec.WhichOneof('kind')
+        if kind == 'value_from_parameter':
+            value_from_parameter = parameter_selector_spec.value_from_parameter
+            outputs[root_output_key] = io_store.get_task_output(
+                value_from_parameter.producer_subtask,
+                value_from_parameter.output_parameter_key,
+            )
+        elif kind == 'value_from_oneof':
+            raise ValueError(
+                "'dsl.OneOf' is not yet supported in local execution.")
```
chensun, line 258: nit: NotImplementedError

sdk/python/kfp/local/graph_utils.py
```diff
@@ -0,0 +1,79 @@
...
+    allows for a general topological_sort sort implementation.
+
+    Args:
+        tasks: The tasks in the pipeline.
+
+    Returns:
+        An adjacency list of tasks name to a list of upstream tasks. The key task depends on all value tasks being executed first.
+    """
+    return {
+        task_name: task_details.dependent_tasks
```
chensun, line 50: nit: This seems to be reversed order of key and value comparing to traditional adjacency list definition. Though you get the result right by reserving the sorted result at the end.

sdk/python/kfp/local/graph_utils_test.py
```diff
@@ -0,0 +1,290 @@
...
+            for k, v in COMPLEX_TASK_TOPOLOGY.items()
+        }
+        actual = graph_utils.topological_sort_tasks(tasks)
+        expected = ['printer', 'divide', 'add-2', 'multiply', 'add']
+        self.assertEqual(actual, expected)
+
+
+SIMPLE_TASK_TOPOLOGY = {
+    'identity': {
+        'cachingOptions': {
```
chensun, line 97: nit: for shorter code and better readability, unused/unrelated fields can be omitted.

chensun (commented):
/lgtm

sdk/python/kfp/dsl/pipeline_task.py
```diff
@@ -190,13 +191,20 @@ def _execute_locally(self, args: Dict[str, Any]) -> None:
...
-
-        self._outputs = task_dispatcher.run_single_task(
-            pipeline_spec=self.component_spec.to_pipeline_spec(),
-            arguments=args,
-        )
+            self._outputs = pipeline_orchestrator.run_local_pipeline(
+                pipeline_spec=self.pipeline_spec,
+                arguments=args,
+            )
+        elif self.component_spec is not None:
```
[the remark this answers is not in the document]
connor-mccarthy: I think this LOC should not be hit for the case of a pipeline in a pipeline, since [`execute_locally`](https://github.com/kubeflow/pipelines/blob/8c7b5b2bf56beef42511bf640d35b2c040389cc9/sdk/python/kfp/dsl/pipeline_task.py#L182) would be `False` (passed in [here](https://github.com/kubeflow/pipelines/blob/8c7b5b2bf56beef42511bf640d35b2c040389cc9/sdk/python/kfp/dsl/base_component.py#L104)). For condition and iteration `DagSpec`s, there is no associated `PipelineTask` so this LOC would also not be hit.

Is there another case you had in mind?

In my WIP implementation of pipeline-in-pipeline local execution, I've also not hit an issue here.

PR #10441, opened 2024-01-29
feat(sdk): Support dsl.ParallelFor over list of Artifacts

KevinGrantLee:

This PR adds support over dsl.ParallelFor over tasks that output lists of artifacts.

ex.
```
@dsl.component
def make_artifacts(...) -> List[Artifact]
    ...

@dsl.pipeline
def my_pipeline:
    make_artifacts_task = make_artifacts()
    with dsl.ParallelFor(items=make_artifacts_task.output) as item:
        print_artifact_name(var_artifact=item)
```

This PR does not support dsl.ParallelFor over a raw list of Artifacts.

**Checklist:**

### DIFF
--- a/sdk/RELEASE.md
+++ b/sdk/RELEASE.md
@@ -3,6 +3,7 @@
 ## Features
 * Support local execution of sequential pipelines [\#10423](https://github.com/kubeflow/pipelines/pull/10423)
 * Support local execution of `dsl.importer` components [\#10431](https://github.com/kubeflow/pipelines/pull/10431)
+* Support local execution of pipelines in pipelines [\#10440](https://github.com/kubeflow/pipelines/pull/10440)
 * Support dsl.ParallelFor over list of Artifacts [\#10441](https://github.com/kubeflow/pipelines/pull/10441)
 
 ## Breaking changes
--- a/sdk/RELEASE.md
+++ b/sdk/RELEASE.md
@@ -2,6 +2,7 @@
 
 ## Features
 * Support local execution of sequential pipelines [\#10423](https://github.com/kubeflow/pipelines/pull/10423)
+* Support local execution of `dsl.importer` components [\#10431](https://github.com/kubeflow/pipelines/pull/10431)
 * Support dsl.ParallelFor over list of Artifacts [\#10441](https://github.com/kubeflow/pipelines/pull/10441)
 
 ## Breaking changes
--- a/sdk/RELEASE.md
+++ b/sdk/RELEASE.md
@@ -3,6 +3,8 @@
 ## Features
 * Support local execution of sequential pipelines [\#10423](https://github.com/kubeflow/pipelines/pull/10423)
 
+* Support dsl.ParallelFor over list of Artifacts [\#10437](https://github.com/kubeflow/pipelines/pull/10437)
+
 ## Breaking changes
 
 ## Deprecations
--- a/sdk/RELEASE.md
+++ b/sdk/RELEASE.md
@@ -3,7 +3,7 @@
 ## Features
 * Support local execution of sequential pipelines [\#10423](https://github.com/kubeflow/pipelines/pull/10423)
 
-* Support dsl.ParallelFor over list of Artifacts [\#10437](https://github.com/kubeflow/pipelines/pull/10437)
+* Support dsl.ParallelFor over list of Artifacts [\#10441](https://github.com/kubeflow/pipelines/pull/10441)
 
 ## Breaking changes
 
--- a/sdk/RELEASE.md
+++ b/sdk/RELEASE.md
@@ -2,7 +2,6 @@
 
 ## Features
 * Support local execution of sequential pipelines [\#10423](https://github.com/kubeflow/pipelines/pull/10423)
-
 * Support dsl.ParallelFor over list of Artifacts [\#10441](https://github.com/kubeflow/pipelines/pull/10441)
 
 ## Breaking changes
--- a/sdk/python/kfp/compiler/compiler_test.py
+++ b/sdk/python/kfp/compiler/compiler_test.py
@@ -3947,7 +3947,7 @@ class ExtractInputOutputDescription(unittest.TestCase):
             string: str,
             in_artifact: Input[Artifact],
         ) -> Outputs:
-            """Pipeline description. Returns.
+            """Pipeline description. Returns
 
             Args:
                 string: Return Pipeline input string. Returns
@@ -4410,9 +4410,7 @@ class TestDslOneOf(unittest.TestCase):
     # To help narrow the tests further (we already test lots of aspects in the following cases), we choose focus on the dsl.OneOf behavior, not the conditional logic if If/Elif/Else. This is more verbose, but more maintainable and the behavior under test is clearer.
 
     def test_if_else_returned(self):
-        """Uses If and Else branches, parameters passed to dsl.OneOf, dsl.OneOf
-        returned from a pipeline, and different output keys on dsl.OneOf
-        channels."""
+        """Uses If and Else branches, parameters passed to dsl.OneOf, dsl.OneOf returned from a pipeline, and different output keys on dsl.OneOf channels."""
 
         @dsl.pipeline
         def roll_die_pipeline() -> str:
@@ -4473,9 +4471,7 @@ class TestDslOneOf(unittest.TestCase):
         )
 
     def test_if_elif_else_returned(self):
-        """Uses If, Elif, and Else branches, parameters passed to dsl.OneOf,
-        dsl.OneOf returned from a pipeline, and different output keys on
-        dsl.OneOf channels."""
+        """Uses If, Elif, and Else branches, parameters passed to dsl.OneOf, dsl.OneOf returned from a pipeline, and different output keys on dsl.OneOf channels."""
 
         @dsl.pipeline
         def roll_die_pipeline() -> str:
@@ -4550,9 +4546,7 @@ class TestDslOneOf(unittest.TestCase):
         )
 
     def test_if_elif_else_consumed(self):
-        """Uses If, Elif, and Else branches, parameters passed to dsl.OneOf,
-        dsl.OneOf passed to a consumer task, and different output keys on
-        dsl.OneOf channels."""
+        """Uses If, Elif, and Else branches, parameters passed to dsl.OneOf, dsl.OneOf passed to a consumer task, and different output keys on dsl.OneOf channels."""
 
         @dsl.pipeline
         def roll_die_pipeline():
@@ -4629,9 +4623,7 @@ class TestDslOneOf(unittest.TestCase):
         )
 
     def test_if_else_consumed_and_returned(self):
-        """Uses If, Elif, and Else branches, parameters passed to dsl.OneOf,
-        and dsl.OneOf passed to a consumer task and returned from the
-        pipeline."""
+        """Uses If, Elif, and Else branches, parameters passed to dsl.OneOf, and dsl.OneOf passed to a consumer task and returned from the pipeline."""
 
         @dsl.pipeline
         def flip_coin_pipeline() -> str:
@@ -4704,8 +4696,7 @@ class TestDslOneOf(unittest.TestCase):
         )
 
     def test_if_else_consumed_and_returned_artifacts(self):
-        """Uses If, Elif, and Else branches, artifacts passed to dsl.OneOf, and
-        dsl.OneOf passed to a consumer task and returned from the pipeline."""
+        """Uses If, Elif, and Else branches, artifacts passed to dsl.OneOf, and dsl.OneOf passed to a consumer task and returned from the pipeline."""
 
         @dsl.pipeline
         def flip_coin_pipeline() -> Artifact:
@@ -4866,8 +4857,7 @@ class TestDslOneOf(unittest.TestCase):
                                      print_task_2.outputs['a'])
 
     def test_deeply_nested_consumed(self):
-        """Uses If, Elif, Else, and OneOf deeply nested within multiple dub-
-        DAGs."""
+        """Uses If, Elif, Else, and OneOf deeply nested within multiple dub-DAGs."""
 
         @dsl.pipeline
         def flip_coin_pipeline(execute_pipeline: bool):
@@ -4966,8 +4956,7 @@ class TestDslOneOf(unittest.TestCase):
                                  print_task_2.outputs['a'])
 
     def test_oneof_in_condition(self):
-        """Tests that dsl.OneOf's channel can be consumed in a downstream group
-        nested one level."""
+        """Tests that dsl.OneOf's channel can be consumed in a downstream group nested one level"""
 
         @dsl.pipeline
         def roll_die_pipeline(repeat_on: str = 'Got heads!'):
@@ -5020,8 +5009,7 @@ class TestDslOneOf(unittest.TestCase):
         )
 
     def test_consumed_in_nested_groups(self):
-        """Tests that dsl.OneOf's channel can be consumed in a downstream group
-        nested multiple levels."""
+        """Tests that dsl.OneOf's channel can be consumed in a downstream group nested multiple levels"""
 
         @dsl.pipeline
         def roll_die_pipeline(
--- a/sdk/python/kfp/compiler/compiler_test.py
+++ b/sdk/python/kfp/compiler/compiler_test.py
@@ -3947,7 +3947,7 @@ class ExtractInputOutputDescription(unittest.TestCase):
             string: str,
             in_artifact: Input[Artifact],
         ) -> Outputs:
-            """Pipeline description. Returns
+            """Pipeline description. Returns.
 
             Args:
                 string: Return Pipeline input string. Returns
@@ -4410,7 +4410,9 @@ class TestDslOneOf(unittest.TestCase):
     # To help narrow the tests further (we already test lots of aspects in the following cases), we choose focus on the dsl.OneOf behavior, not the conditional logic if If/Elif/Else. This is more verbose, but more maintainable and the behavior under test is clearer.
 
     def test_if_else_returned(self):
-        """Uses If and Else branches, parameters passed to dsl.OneOf, dsl.OneOf returned from a pipeline, and different output keys on dsl.OneOf channels."""
+        """Uses If and Else branches, parameters passed to dsl.OneOf, dsl.OneOf
+        returned from a pipeline, and different output keys on dsl.OneOf
+        channels."""
 
         @dsl.pipeline
         def roll_die_pipeline() -> str:
@@ -4471,7 +4473,9 @@ class TestDslOneOf(unittest.TestCase):
         )
 
     def test_if_elif_else_returned(self):
-        """Uses If, Elif, and Else branches, parameters passed to dsl.OneOf, dsl.OneOf returned from a pipeline, and different output keys on dsl.OneOf channels."""
+        """Uses If, Elif, and Else branches, parameters passed to dsl.OneOf,
+        dsl.OneOf returned from a pipeline, and different output keys on
+        dsl.OneOf channels."""
 
         @dsl.pipeline
         def roll_die_pipeline() -> str:
@@ -4546,7 +4550,9 @@ class TestDslOneOf(unittest.TestCase):
         )
 
     def test_if_elif_else_consumed(self):
-        """Uses If, Elif, and Else branches, parameters passed to dsl.OneOf, dsl.OneOf passed to a consumer task, and different output keys on dsl.OneOf channels."""
+        """Uses If, Elif, and Else branches, parameters passed to dsl.OneOf,
+        dsl.OneOf passed to a consumer task, and different output keys on
+        dsl.OneOf channels."""
 
         @dsl.pipeline
         def roll_die_pipeline():
@@ -4623,7 +4629,9 @@ class TestDslOneOf(unittest.TestCase):
         )
 
     def test_if_else_consumed_and_returned(self):
-        """Uses If, Elif, and Else branches, parameters passed to dsl.OneOf, and dsl.OneOf passed to a consumer task and returned from the pipeline."""
+        """Uses If, Elif, and Else branches, parameters passed to dsl.OneOf,
+        and dsl.OneOf passed to a consumer task and returned from the
+        pipeline."""
 
         @dsl.pipeline
         def flip_coin_pipeline() -> str:
@@ -4696,7 +4704,8 @@ class TestDslOneOf(unittest.TestCase):
         )
 
     def test_if_else_consumed_and_returned_artifacts(self):
-        """Uses If, Elif, and Else branches, artifacts passed to dsl.OneOf, and dsl.OneOf passed to a consumer task and returned from the pipeline."""
+        """Uses If, Elif, and Else branches, artifacts passed to dsl.OneOf, and
+        dsl.OneOf passed to a consumer task and returned from the pipeline."""
 
         @dsl.pipeline
         def flip_coin_pipeline() -> Artifact:
@@ -4857,7 +4866,8 @@ class TestDslOneOf(unittest.TestCase):
                                      print_task_2.outputs['a'])
 
     def test_deeply_nested_consumed(self):
-        """Uses If, Elif, Else, and OneOf deeply nested within multiple dub-DAGs."""
+        """Uses If, Elif, Else, and OneOf deeply nested within multiple dub-
+        DAGs."""
 
         @dsl.pipeline
         def flip_coin_pipeline(execute_pipeline: bool):
@@ -4956,7 +4966,8 @@ class TestDslOneOf(unittest.TestCase):
                                  print_task_2.outputs['a'])
 
     def test_oneof_in_condition(self):
-        """Tests that dsl.OneOf's channel can be consumed in a downstream group nested one level"""
+        """Tests that dsl.OneOf's channel can be consumed in a downstream group
+        nested one level."""
 
         @dsl.pipeline
         def roll_die_pipeline(repeat_on: str = 'Got heads!'):
@@ -5009,7 +5020,8 @@ class TestDslOneOf(unittest.TestCase):
         )
 
     def test_consumed_in_nested_groups(self):
-        """Tests that dsl.OneOf's channel can be consumed in a downstream group nested multiple levels"""
+        """Tests that dsl.OneOf's channel can be consumed in a downstream group
+        nested multiple levels."""
 
         @dsl.pipeline
         def roll_die_pipeline(
--- a/sdk/python/kfp/compiler/pipeline_spec_builder.py
+++ b/sdk/python/kfp/compiler/pipeline_spec_builder.py
@@ -131,25 +131,28 @@ def build_task_spec_for_task(
     for input_name, input_value in task.inputs.items():
         # Since LoopParameterArgument and LoopArtifactArgument and LoopArgumentVariable are narrower
         # types than PipelineParameterChannel, start with them.
-        if isinstance(
-                input_value,
-            (for_loop.LoopParameterArgument, for_loop.LoopArtifactArgument)):
 
-            component_input_arg = (
+        if isinstance(input_value, for_loop.LoopParameterArgument):
+
+            component_input_parameter = (
                 compiler_utils.additional_input_name_for_pipeline_channel(
                     input_value))
-            if component_input_arg in parent_component_inputs.parameters:
-                pipeline_task_spec.inputs.parameters[
-                    input_name].component_input_parameter = (
-                        component_input_arg)
-            elif component_input_arg in parent_component_inputs.artifacts:
-                pipeline_task_spec.inputs.artifacts[
-                    input_name].component_input_artifact = (
-                        component_input_arg)
-            else:
-                raise RuntimeError(
-                    f'component_input_arg: {component_input_arg} not found. All inputs: {parent_component_inputs}'
-                )
+            assert component_input_parameter in parent_component_inputs.parameters, \
+                f'component_input_parameter: {component_input_parameter} not found. All inputs: {parent_component_inputs}'
+            pipeline_task_spec.inputs.parameters[
+                input_name].component_input_parameter = (
+                    component_input_parameter)
+
+        elif isinstance(input_value, for_loop.LoopArtifactArgument):
+
+            component_input_artifact = (
+                compiler_utils.additional_input_name_for_pipeline_channel(
+                    input_value))
+            assert component_input_artifact in parent_component_inputs.artifacts, \
+                f'component_input_artifact: {component_input_artifact} not found. All inputs: {parent_component_inputs}'
+            pipeline_task_spec.inputs.artifacts[
+                input_name].component_input_artifact = (
+                    component_input_artifact)
 
         elif isinstance(input_value, for_loop.LoopArgumentVariable):
 
@@ -701,21 +704,25 @@ def build_component_spec_for_group(
                         channel.channel_type))
             component_spec.input_definitions.artifacts[
                 input_name].is_artifact_list = channel.is_artifact_list
-        else:
-            # channel is one of PipelineParameterChannel, LoopParameterArgument,
-            # LoopArtifactArgument, or LoopArgumentVariable.
-            if isinstance(channel, for_loop.LoopArtifactArgument):
-                component_spec.input_definitions.artifacts[
-                    input_name].artifact_type.CopyFrom(
-                        type_utils.bundled_artifact_to_artifact_proto(
-                            channel.channel_type))
+        elif isinstance(channel, for_loop.LoopArtifactArgument):
+            component_spec.input_definitions.artifacts[
+                input_name].artifact_type.CopyFrom(
+                    type_utils.bundled_artifact_to_artifact_proto(
+                        channel.channel_type))
 
-                component_spec.input_definitions.artifacts[
-                    input_name].is_artifact_list = channel.is_artifact_list
-            else:
-                component_spec.input_definitions.parameters[
-                    input_name].parameter_type = type_utils.get_parameter_type(
-                        channel.channel_type)
+            component_spec.input_definitions.artifacts[
+                input_name].is_artifact_list = channel.is_artifact_list
+        elif isinstance(channel,
+                        (pipeline_channel.PipelineParameterChannel,
+                         for_loop.LoopParameterArgument,
+                         for_loop.LoopArgumentVariable, for_loop.Collected)):
+            component_spec.input_definitions.parameters[
+                input_name].parameter_type = type_utils.get_parameter_type(
+                    channel.channel_type)
+        else:
+            raise TypeError(
+                f'Expected PipelineParameterChannel, PipelineArtifactChannel, LoopParameterArgument, LoopArtifactArgument, or LoopArgumentVariable, got {type(channel)}'
+            )
 
     for output_name, output in output_pipeline_channels.items():
         if isinstance(output, pipeline_channel.PipelineArtifactChannel):
--- a/sdk/python/kfp/compiler/pipeline_spec_builder.py
+++ b/sdk/python/kfp/compiler/pipeline_spec_builder.py
@@ -796,7 +796,6 @@ def _update_task_spec_for_loop_group(
                 f'Expected LoopParameterArgument or LoopArtifactArgument, got {type(group.loop_argument)}.'
             )
 
-
         # If the loop items itself is a loop arguments variable, handle the
         # subvar name.
         if isinstance(loop_items_channel, for_loop.LoopArgumentVariable):
--- a/sdk/python/kfp/compiler/pipeline_spec_builder.py
+++ b/sdk/python/kfp/compiler/pipeline_spec_builder.py
@@ -129,23 +129,24 @@ def build_task_spec_for_task(
             task._task_spec.retry_policy.to_proto())
 
     for input_name, input_value in task.inputs.items():
-        # since LoopArgument and LoopArgumentVariable are narrower types than PipelineParameterChannel, start with it
-        if isinstance(input_value, for_loop.LoopArgument):
+        # Since LoopParameterArgumen andLoopArtifactArgument and LoopArgumentVariable are narrower
+        # types than PipelineParameterChannel, start with them.
+        if isinstance(input_value, (for_loop.LoopParameterArgument, for_loop.LoopArtifactArgument)):
 
-            component_input_parameter = (
+            component_input_arg = (
                 compiler_utils.additional_input_name_for_pipeline_channel(
                     input_value))
-            if component_input_parameter in parent_component_inputs.parameters:
+            if component_input_arg in parent_component_inputs.parameters:
                 pipeline_task_spec.inputs.parameters[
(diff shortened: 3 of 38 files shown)


sdk/python/kfp/dsl/types/type_annotations.py
```diff
@@ -220,15 +220,26 @@ def get_short_type_name(type_name: str) -> str:
...
     Args:
       type_name: The original type name.
 
     Returns:
       The short form type name or the original name if pattern doesn't match.
     """
-    match = re.match('(typing\.)?(?P<type>\w+)(?:\[.+\])?', type_name)
-    return match['type'] if match else type_name
+    parameter_match = re.match('(typing\.)?(?P<type>\w+)(?:\[.+\])?', type_name)
+    artifact_match = re.match('(system\.|google\.)(?P<type>\w+)@(\d\.\d\.\d)',
```
connor-mccarthy, line 234: I think there might be a simpler way to extract the artifact type: https://github.com/kubeflow/pipelines/blob/7bd31d104bd403a830bf2a455c9c2c0dbf493c4d/sdk/python/kfp/local/importer_handler.py#L137-L142

I also wonder if we need to at all in some places in this PR. Can we just pass the type from the `PipelineArtifactChannel` directly to the `LoopArtifactArgument`?

sdk/python/kfp/dsl/for_loop.py
```diff
@@ -39,8 +43,17 @@ def _get_loop_item_type(type_name: str) -> Optional[str]:
...
+                               type_name)
+    artifact_match = re.match('(system\.|google\.)(?P<type>\w+)@(\d\.\d\.\d)',
+                              type_name)
+
+    if parameter_match:
+        return parameter_match['item_type'].lstrip().rstrip()
+    elif artifact_match:
+        return type_name
+    else:
+        return None
```
connor-mccarthy, line 56: This seems like it may be duplicative of the changes to `type_utils.get_short_type_name`. Is there any opportunity to dedupe?

Separately, this comment may be obsolete due to other comments about how we can extract artifact type information.

sdk/python/kfp/dsl/for_loop.py
```diff
@@ -64,33 +77,46 @@ def _get_subvar_type(type_name: str) -> Optional[str]:
     return match['value_type'].lstrip().rstrip() if match else None
 
 
-class LoopArgument(pipeline_channel.PipelineParameterChannel):
-    """Represents the argument that are looped over in a ParallelFor loop.
+def _make_name(code: str):
```
connor-mccarthy, line 80: missing return annotation

sdk/python/kfp/dsl/for_loop.py
```diff
@@ -64,33 +77,46 @@ def _get_subvar_type(type_name: str) -> Optional[str]:
...
+    """Makes a name for a loop argument from a unique code."""
+    return f'{LOOP_ITEM_PARAM_NAME_BASE}-{code}'
+
+
+def name_is_loop_argument(name: str) -> bool:
```
connor-mccarthy, line 85: Is this used anywhere?

sdk/python/kfp/dsl/for_loop.py
```diff
@@ -99,25 +125,23 @@ def __init__(
             name_code: A unique code used to identify these loop arguments.
                 Should match the code for the ParallelFor ops_group which created
                 these LoopArguments. This prevents parameter name collisions.
-            name_override: The override name for PipelineParameterChannel.
-            **kwargs: Any other keyword arguments passed down to PipelineParameterChannel.
+            name_override: The override name for PipelineChannel.
+            **kwargs: Any other keyword arguments passed down to PipelineChannel.
```
connor-mccarthy, line 129: (Not required for this PR)

AFAICT, since this `**kwargs` is an implementation detail and we know all consumers of this API, there's no reason to have `**kwargs`. It would be nice to remove this by converting to explicit arguments at some point.

sdk/python/kfp/dsl/for_loop.py
```diff
@@ -99,25 +125,23 @@ def __init__(
...
-            super().__init__(name=self._make_name(name_code), **kwargs)
+            super().__init__(name=_make_name(name_code), **kwargs)
         else:
             super().__init__(name=name_override, **kwargs)
 
-        if not isinstance(
-                items,
-            (list, tuple, pipeline_channel.PipelineParameterChannel)):
+        if not isinstance(items,
+                          (list, tuple, pipeline_channel.PipelineChannel)):
```
connor-mccarthy, line 142: Do we still want to enforce that this is a `PipelineParameterChannel` if it's a `LoopParameterArgument`? There may be another reason for loosening this check that I'm not aware of.

sdk/python/kfp/dsl/for_loop.py
```diff
@@ -149,20 +173,16 @@ def __getattr__(self, name: str):
...
-        """Makes a name for this loop argument from a unique code."""
-        return f'{self.LOOP_ITEM_PARAM_NAME_BASE}-{code}'
-
     @classmethod
     def from_pipeline_channel(
         cls,
-        channel: pipeline_channel.PipelineParameterChannel,
-    ) -> 'LoopArgument':
-        """Creates a LoopArgument object from a PipelineParameterChannel
+        channel: pipeline_channel.PipelineChannel,
```
connor-mccarthy, line 179: Should this still be a `PipelineParameterChannel`?

sdk/python/kfp/dsl/for_loop.py
```diff
@@ -172,26 +192,82 @@ def from_raw_items(
...
+            super().__init__(name=name_override, **kwargs)
+
+        self.items_or_pipeline_channel = items
+        self.is_with_items_loop_argument = not isinstance(
+            items, pipeline_channel.PipelineArtifactChannel)
+
+    @classmethod
+    def from_pipeline_channel(
+        cls,
+        channel: pipeline_channel.PipelineChannel,
```
connor-mccarthy, line 259: Same as comments above: `PipelineArtifactChannel`?

sdk/python/kfp/dsl/for_loop.py
```diff
@@ -301,21 +377,30 @@ def __init__(
...
+        # it's "correct" channel during compilation
         if isinstance(output, pipeline_channel.PipelineArtifactChannel):
             channel_type = output.channel_type
             self.is_artifact_channel = True
-            # we know all dsl.Collected instances are lists, so set to true
-            # for type checking, which occurs before dsl.Collected is updated to
-            # it's "correct" channel during compilation
-            self.is_artifact_list = True
+            is_artifact_list = True
+        elif isinstance(output, pipeline_channel.PipelineChannel
```
connor-mccarthy, line 387: Can we simplify by checking if `isinstance(output, pipeline_channel.PipelineParameterChannel)`?

sdk/python/kfp/compiler/pipeline_spec_builder.py
```diff
@@ -128,17 +129,27 @@ def build_task_spec_for_task(
             task._task_spec.retry_policy.to_proto())
 
     for input_name, input_value in task.inputs.items():
-        # since LoopArgument and LoopArgumentVariable are narrower types than PipelineParameterChannel, start with it
-        if isinstance(input_value, for_loop.LoopArgument):
+        # Since LoopParameterArgumen andLoopArtifactArgument and LoopArgumentVariable are narrower
```
connor-mccarthy, line 132:
```suggestion
        # Since LoopParameterArgument and LoopArtifactArgument and LoopArgumentVariable are narrower
```

sdk/python/kfp/compiler/pipeline_spec_builder.py
```diff
@@ -128,17 +129,27 @@ def build_task_spec_for_task(
...
+        # types than PipelineParameterChannel, start with them.
+        if isinstance(
+                input_value,
+            (for_loop.LoopParameterArgument, for_loop.LoopArtifactArgument)):
```
connor-mccarthy, line 136: Can we flatten from (current):
```
if is (loop param, loop artifact):
   if (loop param):
      ...
   elif (loop artifact):
      ...
   else:
      ...
```
to:
```
if is (loop param):
   ...
elif is (loop artifact):
   ...
else:
   ...
```

sdk/python/kfp/compiler/pipeline_spec_builder.py
```diff
@@ -691,11 +702,20 @@ def build_component_spec_for_group(
...
+                    input_name].artifact_type.CopyFrom(
+                        type_utils.bundled_artifact_to_artifact_proto(
+                            channel.channel_type))
+
+                component_spec.input_definitions.artifacts[
+                    input_name].is_artifact_list = channel.is_artifact_list
+            else:
+                component_spec.input_definitions.parameters[
+                    input_name].parameter_type = type_utils.get_parameter_type(
+                        channel.channel_type)
```
connor-mccarthy, line 718: See comment above about flattening nested conditions. Is that possible here?

sdk/python/kfp/compiler/pipeline_spec_builder.py
```diff
@@ -691,11 +702,20 @@ def build_component_spec_for_group(
...
+            # LoopArtifactArgument, or LoopArgumentVariable.
+            if isinstance(channel, for_loop.LoopArtifactArgument):
+                component_spec.input_definitions.artifacts[
+                    input_name].artifact_type.CopyFrom(
+                        type_utils.bundled_artifact_to_artifact_proto(
+                            channel.channel_type))
+
+                component_spec.input_definitions.artifacts[
+                    input_name].is_artifact_list = channel.is_artifact_list
+            else:
```
connor-mccarthy, line 715: Since we're adding more values for `channel` here, I think it makes sense to formalize the comment above (`# channel is one of PipelineParameterChannel, LoopParameterArgument, LoopArtifactArgument, or LoopArgumentVariable.`) into code. We can do this be changing `if/else` to `if/elif/(else, then raise)`. The middle `elif` will explicitly state what we values we expect, which is implicit with `else`.

sdk/python/kfp/compiler/pipeline_spec_builder.py
```diff
@@ -747,13 +767,29 @@ def _update_task_spec_for_loop_group(
...
 
-        loop_arguments_item = f'{input_parameter_name}-{for_loop.LoopArgument.LOOP_ITEM_NAME_BASE}'
+        loop_arguments_item = f'{input_parameter_name}-{for_loop.LOOP_ITEM_NAME_BASE}'
         assert loop_arguments_item == loop_argument_item_name
 
-        pipeline_task_spec.parameter_iterator.items.input_parameter = (
-            input_parameter_name)
-        pipeline_task_spec.parameter_iterator.item_input = (
-            loop_argument_item_name)
+        if type_annotations.is_artifact_subtype(group.channel_type):
```
connor-mccarthy, line 773: See other comments about extracting artifact types.

sdk/python/kfp/compiler/pipeline_spec_builder.py
```diff
@@ -747,13 +767,29 @@ def _update_task_spec_for_loop_group(
         loop_argument_item_name = compiler_utils.additional_input_name_for_pipeline_channel(
             group.loop_argument.full_name)
 
-        loop_arguments_item = f'{input_parameter_name}-{for_loop.LoopArgument.LOOP_ITEM_NAME_BASE}'
+        loop_arguments_item = f'{input_parameter_name}-{for_loop.LOOP_ITEM_NAME_BASE}'
```
connor-mccarthy, line 770: This is a good change!

sdk/python/kfp/dsl/for_loop.py
```diff
@@ -172,26 +192,82 @@ def from_raw_items(
...
+        channel: pipeline_channel.PipelineChannel,
+        is_artifact_list: bool = False,
+    ) -> 'LoopArtifactArgument':
+        """Creates a LooArtifactArgument object from a PipelineChannel
+        object."""
+        return LoopArtifactArgument(
+            items=channel,
+            name_override=channel.name + '-' + LOOP_ITEM_NAME_BASE,
+            task_name=channel.task_name,
+            channel_type=_get_loop_item_type(channel.channel_type) or 'String',
```
connor-mccarthy, line 268: See other comments about getting artifact types from the channel. I think there may be an easier way to do this without `_get_loop_item_type` needing to know how to handle both parameters and artifacts and perform additional string parsing.

sdk/python/kfp/dsl/for_loop.py
```diff
@@ -172,26 +192,82 @@ def from_raw_items(
...
+        self.is_with_items_loop_argument = not isinstance(
+            items, pipeline_channel.PipelineArtifactChannel)
+
+    @classmethod
+    def from_pipeline_channel(
+        cls,
+        channel: pipeline_channel.PipelineChannel,
+        is_artifact_list: bool = False,
+    ) -> 'LoopArtifactArgument':
+        """Creates a LooArtifactArgument object from a PipelineChannel
```
connor-mccarthy, line 262:
```suggestion
        """Creates a LoopArtifactArgument object from a PipelineChannel
```

connor-mccarthy (changes requested):

sdk/RELEASE.md
```diff
@@ -3,6 +3,8 @@
 ## Features
 * Support local execution of sequential pipelines [\#10423](https://github.com/kubeflow/pipelines/pull/10423)
```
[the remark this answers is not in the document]
KevinGrantLee: Done.

sdk/python/kfp/compiler/compiler_test.py
```diff
@@ -4410,7 +4410,9 @@ class TestDslOneOf(unittest.TestCase):
     # To help narrow the tests further (we already test lots of aspects in the following cases), we choose focus on the dsl.OneOf behavior, not the conditional logic if If/Elif/Else. This is more verbose, but more maintainable and the behavior under test is clearer.
 
     def test_if_else_returned(self):
-        """Uses If and Else branches, parameters passed to dsl.OneOf, dsl.OneOf returned from a pipeline, and different output keys on dsl.OneOf channels."""
+        """Uses If and Else branches, parameters passed to dsl.OneOf, dsl.OneOf
+        returned from a pipeline, and different output keys on dsl.OneOf
+        channels."""
```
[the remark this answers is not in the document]
KevinGrantLee: Done.

sdk/python/kfp/dsl/tasks_group.py
```diff
@@ -455,11 +456,18 @@ def __init__(
...
+            if type_annotations.is_artifact_subtype(self.channel_type):
+                self.loop_argument = for_loop.LoopArtifactArgument.from_pipeline_channel(
+                    items,
+                    is_artifact_list=False,
+                )
+            else:
+                self.loop_argument = for_loop.LoopParameterArgument.from_pipeline_channel(
+                    items,)
             self.items_is_pipeline_channel = True
         else:
```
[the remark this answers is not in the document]
KevinGrantLee: Done.

sdk/python/kfp/dsl/tasks_group.py
```diff
@@ -455,11 +456,18 @@ def __init__(
         )
 
         if isinstance(items, pipeline_channel.PipelineChannel):
```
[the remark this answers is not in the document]
KevinGrantLee: Done.

sdk/python/kfp/dsl/pipeline_channel.py
```diff
@@ -73,6 +73,8 @@ def __init__(
         name: str,
         channel_type: Union[str, Dict],
         task_name: Optional[str] = None,
+        is_artifact_list: Optional[bool] = False,
```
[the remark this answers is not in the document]
KevinGrantLee: Done.

sdk/python/kfp/dsl/for_loop.py
KevinGrantLee: Done.

sdk/python/kfp/compiler/pipeline_spec_builder.py
KevinGrantLee: Done.

sdk/python/kfp/dsl/for_loop.py
KevinGrantLee: Done.

sdk/python/kfp/dsl/for_loop.py
KevinGrantLee: Done.

sdk/python/kfp/dsl/for_loop.py
KevinGrantLee: Done.

sdk/python/kfp/dsl/for_loop.py
KevinGrantLee: Done.

sdk/python/kfp/dsl/for_loop.py
KevinGrantLee: Done.

sdk/python/kfp/dsl/for_loop_test.py
```diff
@@ -91,7 +91,8 @@ def test_get_subvar_type(self, dict_type, value_type):
     )
     def test_loop_argument_from_pipeline_channel(self, channel,
                                                  expected_serialization_value):
-        loop_argument = for_loop.LoopArgument.from_pipeline_channel(channel)
+        loop_argument = for_loop.LoopParameterArgument.from_pipeline_channel(
```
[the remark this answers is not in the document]
KevinGrantLee: Done.

sdk/python/kfp/dsl/for_loop.py
KevinGrantLee: It's used in https://github.com/kubeflow/pipelines/blob/7bd31d104bd403a830bf2a455c9c2c0dbf493c4d/sdk/python/kfp/dsl/for_loop_test.py#L152

sdk/python/kfp/dsl/types/type_annotations.py
```diff
@@ -220,15 +220,26 @@ def get_short_type_name(type_name: str) -> str:
       typing.Dict[str, str] -> Dict
```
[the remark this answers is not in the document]
KevinGrantLee: Done.

sdk/python/kfp/compiler/pipeline_spec_builder.py
KevinGrantLee: Done.

sdk/python/kfp/compiler/pipeline_spec_builder.py
KevinGrantLee: Done.

sdk/python/kfp/compiler/pipeline_spec_builder.py
KevinGrantLee: Done.

sdk/python/kfp/dsl/types/type_annotations.py
KevinGrantLee: This method doesnt handle the schema version `@0.0.1` so we would have to pre-parse the string every time to check the channel type and it wouldn't support any gcpc artifact types.

Also the usage is a little different bc that method presupposes the input is a valid artifact class and always returns an artifact class, whereas `type_annotations.get_short_type_name` is more general and works for all input types.

My preference here is to keep `type_annotations.get_short_type_name`.

sdk/python/kfp/dsl/for_loop.py
KevinGrantLee: Deduped via `type_annotations.extract_artifact_subclass`

sdk/python/kfp/dsl/for_loop.py
KevinGrantLee: Simplified this, now `_get_loop_item_type` only handles parameter type checking.

sdk/python/kfp/dsl/for_loop.py
connor-mccarthy: Can we move this to be a function in the test module?

sdk/python/kfp/dsl/pipeline_channel.py
```diff
@@ -83,6 +83,10 @@ def __init__(
             task_name: Optional; The name of the task that produces the pipeline
                 channel. If provided, the task name will be sanitized to be k8s
                 compatible.
+            is_artifact_list: Optional; True if `channel_type` represents a list
+                of the artifact type. Only applies when `channel_type` is an artifact.
+            value: The actual value of the pipeline channel. If provided, the
+                pipeline channel is "resolved" immediately.
```
connor-mccarthy, line 89: Docstrings changes also need to be removed

sdk/python/kfp/dsl/tasks_group.py
```diff
@@ -22,6 +22,7 @@
 from kfp.dsl import pipeline_channel
 from kfp.dsl import pipeline_context
 from kfp.dsl import pipeline_task
+from kfp.dsl.types import type_annotations
```
connor-mccarthy, line 25: Unused import (applies to the `type_annotations` import throughout)

sdk/python/kfp/compiler/pipeline_spec_builder.py
```diff
@@ -690,12 +704,25 @@ def build_component_spec_for_group(
                         channel.channel_type))
             component_spec.input_definitions.artifacts[
                 input_name].is_artifact_list = channel.is_artifact_list
-        else:
-            # channel is one of PipelineParameterChannel, LoopArgument, or
-            # LoopArgumentVariable.
+        elif isinstance(channel, for_loop.LoopArtifactArgument):
```
connor-mccarthy, line 707: Can we bundle this into the condition above, consistent with how we handle parameters below?

`if isinstance(channel, (pipeline_channel.PipelineArtifactChannel, for_loop.LoopArtifactArgument)):`

sdk/python/kfp/compiler/pipeline_spec_builder.py
```diff
@@ -747,13 +774,29 @@ def _update_task_spec_for_loop_group(
...
 
-        loop_arguments_item = f'{input_parameter_name}-{for_loop.LoopArgument.LOOP_ITEM_NAME_BASE}'
+        loop_arguments_item = f'{input_parameter_name}-{for_loop.LOOP_ITEM_NAME_BASE}'
         assert loop_arguments_item == loop_argument_item_name
 
-        pipeline_task_spec.parameter_iterator.items.input_parameter = (
-            input_parameter_name)
-        pipeline_task_spec.parameter_iterator.item_input = (
-            loop_argument_item_name)
+        if group.channel_is_artifact_subtype:
```
connor-mccarthy, line 780: Can we get this by checking the type of `group.loop_argument.items_or_pipeline_channel`? It's useful to avoid expanding the public interface of `group` with additional attributes when the information is available on existing attributes.

connor-mccarthy: [Not your code] Same with `group.items_is_pipeline_channel`, though this is pre-existing so feel free to leave as-is for purposes of this PR.

sdk/python/kfp/dsl/types/type_annotations.py
connor-mccarthy: Is this update still needed with the changed implementation? If I remove this change, tests still pass.

sdk/python/kfp/dsl/for_loop.py
```diff
@@ -172,26 +185,84 @@ def from_raw_items(
...
+
+        self.items_or_pipeline_channel = items
+        self.is_with_items_loop_argument = not isinstance(
+            items, pipeline_channel.PipelineArtifactChannel)
+
+    @classmethod
+    def from_pipeline_channel(
+        cls,
+        channel: pipeline_channel.PipelineArtifactChannel,
+        is_artifact_list: bool = False,
```
connor-mccarthy, line 253: Is `is_artifact_list` already available on `PipelineArtifactChannel`?

connor-mccarthy (changes requested):

sdk/python/kfp/dsl/pipeline_channel.py
KevinGrantLee: Done.

sdk/python/kfp/dsl/tasks_group.py
KevinGrantLee: Done.

sdk/python/kfp/compiler/pipeline_spec_builder.py
KevinGrantLee: Done.

sdk/python/kfp/compiler/pipeline_spec_builder.py
KevinGrantLee: Done, re-added `type_annotations.issubtype_of_artifact()` to check`group.channel_type`

sdk/python/kfp/dsl/for_loop.py
KevinGrantLee: Removed the former.

`is_artifact_list` is not available there, but it will always be false because we don't support nested lists yet. So removed that argument and directly pass `is_artifact_list=False` into the subsequent `PipelineArtifactChannel` constructor.

sdk/python/kfp/dsl/types/type_annotations.py
KevinGrantLee: Removed.

sdk/python/kfp/compiler/pipeline_spec_builder.py
```diff
@@ -683,19 +696,25 @@ def build_component_spec_for_group(
...
                         channel.channel_type))
             component_spec.input_definitions.artifacts[
                 input_name].is_artifact_list = channel.is_artifact_list
-        else:
-            # channel is one of PipelineParameterChannel, LoopArgument, or
-            # LoopArgumentVariable.
+        elif isinstance(channel,
+                        (pipeline_channel.PipelineParameterChannel,
+                         for_loop.LoopParameterArgument,
+                         for_loop.LoopArgumentVariable, for_loop.Collected)):
```
connor-mccarthy, line 710: Is it correct for `for_loop.Collected` be in this tuple of types? `dsl.Collected` can result in a list of artifacts.

I suppose this is consistent with the prior implementation before making the `else` explicit with `elif`, but now that it's there just want to make sure this is correct.

If it is correct, let's use the public object path `dsl.Collected` (rather than `for_loop.Collected`), since this is a better test of user code paths. This style of referring to the public API for public objects is the dominant approach throughout the SDK internals.

sdk/python/kfp/compiler/pipeline_spec_builder.py
```diff
@@ -747,13 +766,29 @@ def _update_task_spec_for_loop_group(
...
+                loop_items_channel)
+
+            pipeline_task_spec.artifact_iterator.items.input_artifact = input_artifact_name
+            pipeline_task_spec.artifact_iterator.item_input = (
+                loop_argument_item_name)
+
+            _pop_input_from_task_spec(
+                task_spec=pipeline_task_spec,
+                input_name=pipeline_task_spec.artifact_iterator.item_input)
+        else:
```
connor-mccarthy, line 783: nit: consider explicit `elif, else (raise)` here too, rather than implicit `else`

sdk/python/test_data/pipelines/pipeline_with_parallelfor_list_artifacts.py
```diff
@@ -0,0 +1,77 @@
+# Copyright 2022 The Kubeflow Authors
```
connor-mccarthy, line 1:
```suggestion
# Copyright 2024 The Kubeflow Authors
```

sdk/python/kfp/dsl/tasks_group.py
```diff
@@ -454,20 +454,29 @@ def __init__(
             is_root=False,
         )
 
-        if isinstance(items, pipeline_channel.PipelineChannel):
-            self.loop_argument = for_loop.LoopArgument.from_pipeline_channel(
+        if isinstance(items, pipeline_channel.PipelineParameterChannel):
+            self.channel_type = str(items.channel_type)
```
connor-mccarthy, line 458: Do we need to set the `.channel_type` attribute? If I comment this out, the compilation test still passes.

sdk/python/kfp/dsl/for_loop.py
```diff
@@ -301,16 +363,17 @@ def __init__(
         output: pipeline_channel.PipelineChannel,
     ) -> None:
         self.output = output
+        # we know all dsl.Collected instances are lists, so set `is_artifact_list``
```
connor-mccarthy, line 366:
```suggestion
        # we know all dsl.Collected instances are lists, so set `is_artifact_list`
```

sdk/python/kfp/dsl/for_loop_test.py
```diff
@@ -89,12 +98,44 @@ def test_get_subvar_type(self, dict_type, value_type):
...
+            'expected_serialization_value':
+                '{{channel:task=task1;name=param1-loop-item;type=system.Artifact@0.0.1;}}',
+        },
+        {
+            'channel':
+                pipeline_channel.PipelineArtifactChannel(
+                    name='output1',
+                    channel_type='system.Dataset@0.0.1',
+                    task_name='task1',
+                    is_artifact_list=False,
```
connor-mccarthy, line 126: Should `is_artifact_list` be `True` since we are going to iterate over it? If yes, should we add validation to catch the case where `is_artifact_list=False`?

As a concrete example, this pipeline happens to compile, but it should not:

```python
@dsl.pipeline
def my_pipeline():
    single_artifact = make_single_artifact(content='foo')
    with dsl.ParallelFor(items=single_artifact.output) as item:
        print_artifact(artifact=item)
```

Note that it does run successfully due to an implementation detail of how we implement artifacts, but we should block this on the compiler-side since it doesn't make sense to iterate over a single `Artifact`.

sdk/python/test_data/pipelines/pipeline_with_parallelfor_list_artifacts.py
```diff
@@ -0,0 +1,77 @@
...
+def make_datasets(
+        texts: List[str] = ['Hello', ',', ' ', 'world!']) -> List[Dataset]:
+    with dsl.ParallelFor(texts) as text:
+        t1 = make_dataset(text=text)
+
+    return dsl.Collected(t1.output)
+
+
+@dsl.component
+def make_artifact(name: str) -> Artifact:
```
connor-mccarthy, line 47: Since this is used as the artifact `content`, perhaps rename this parameter to something other than `name`. The artifact name is a different attribute on the artifact. If you look at the value printed by `print(artifact.name)` in downstream tasks you wont see the value you pass in here.

sdk/python/test_data/pipelines/pipeline_with_parallelfor_list_artifacts.py
```diff
@@ -0,0 +1,77 @@
...
+
+
+@dsl.pipeline
+def make_artifacts(
+        names: List[str] = ['Hello', ',', ' ', 'world!']) -> List[Artifact]:
+    with dsl.ParallelFor(names) as name:
+        t1 = make_artifact(name=name)
+
+    return dsl.Collected(t1.output)
+
```
connor-mccarthy, line 61: I think it would be good to test some of the other pipeline topologies that a user may express. For example, I believe this should run but does not:

```python
@dsl.pipeline
def my_pipeline(
        names: List[str] = ['Hello', ',', ' ', 'world!']) -> List[Artifact]:
    with dsl.ParallelFor(names) as name:
        t1 = make_artifact(name=name)

    with dsl.ParallelFor(dsl.Collected(t1.output)) as item:
        print_artifact_name(artifact=item)
```

We can do so in this file. It also may be helpful to do this in `compiler_test.py`, that way we can split out the test cases into clearly named methods to ensure we're capturing a representative set of topologies.

connor-mccarthy (changes requested):

sdk/python/kfp/compiler/pipeline_spec_builder.py
KevinGrantLee: Yes, it is correct, ok changed all occurrences of `for_loop.Collected` to `dsl.Collected`. There were some other pre-existing references that I changed.

sdk/python/kfp/compiler/pipeline_spec_builder.py
KevinGrantLee: Done.

sdk/python/test_data/pipelines/pipeline_with_parallelfor_list_artifacts.py
KevinGrantLee: Done.

sdk/python/kfp/dsl/tasks_group.py
KevinGrantLee: Missed that, done.

sdk/python/kfp/dsl/for_loop.py
KevinGrantLee: Done.

sdk/python/test_data/pipelines/pipeline_with_parallelfor_list_artifacts.py
KevinGrantLee: Done, renamed to `data`.

sdk/python/test_data/pipelines/pipeline_with_parallelfor_list_artifacts.py
KevinGrantLee: I tried out the corresponding code for parameters in a clean venv (installed kfp from pip) and I also got an error
```
@dsl.component
def str_identity(s: str) -> str:
    return s

@dsl.pipeline
def my_pipeline(): 
  texts = ['Hello', ',', ' ', 'world!']
  with dsl.ParallelFor(texts) as text:
      t1 = str_identity(s=text)

  with dsl.ParallelFor(dsl.Collected(t1.output)) as item:
      str_identity(s=item)
```

Seems like this issue affects parameters as well as artifacts, wdyt about resolving this in the next pr?

sdk/python/kfp/dsl/for_loop_test.py
KevinGrantLee: Done, also noticed same issue affects parameters. I was able to compile the corresponding parameter pipeline, but it failed without a clear error. Will address this is the next PR.


```
@dsl.component
def str_identity(s: str) -> str:
    return s


@dsl.pipeline
def my_pipeline():
single_param = str_identity(s='string')
    with dsl.ParallelFor(items=single_param.output) as item:
        str_identity(s=item)
```

sdk/python/test_data/pipelines/pipeline_with_parallelfor_list_artifacts.py
connor-mccarthy: Good catch! Addressing as a follow-up sounds good.

sdk/python/kfp/dsl/for_loop.py
```diff
@@ -172,26 +175,86 @@ def from_raw_items(
...
+            items, pipeline_channel.PipelineArtifactChannel)
+
+    @classmethod
+    def from_pipeline_channel(
+        cls,
+        channel: pipeline_channel.PipelineArtifactChannel,
+    ) -> 'LoopArtifactArgument':
+        """Creates a LoopArtifactArgument object from a PipelineArtifactChannel
+        object."""
+        assert channel.is_artifact_list is True, 'Cannot iterate over single Artifacts.'
```
connor-mccarthy, line 249: Can we raise a formal exception as we do in other cases (as opposed to an `AssertionError`)? And also add a bit more info to help the user associate the error with their LOC? Since the compilation error messages are typically not thrown from the offending LOC in the user's program, it can be hard discern where the error comes from and what to do about it if we don't provide more info.

For example:

```python
if channel.is_artifact_list:
    raise ValueError("Cannot iterate over single a Artifact using 'dsl.ParallelFor'. Expected a list of artifacts as argument to 'items'.")
```

sdk/python/test_data/pipelines/pipeline_with_parallelfor_list_artifacts.py
connor-mccarthy: Good catch. Fixing in a follow-up PR sounds good!

connor-mccarthy (changes requested):
/lgtm with one additional change

sdk/python/kfp/dsl/for_loop.py
KevinGrantLee: Done.

sdk/python/kfp/dsl/for_loop.py
```diff
@@ -246,7 +246,8 @@ def from_pipeline_channel(
     ) -> 'LoopArtifactArgument':
         """Creates a LoopArtifactArgument object from a PipelineArtifactChannel
         object."""
-        assert channel.is_artifact_list is True, 'Cannot iterate over single Artifacts.'
+        if not channel.is_artifact_list:
```
connor-mccarthy, line 249: Thanks, Kevin. Can we add a test for this?

KevinGrantLee: Done.

sdk/python/kfp/dsl/for_loop_test.py
```diff
@@ -136,6 +136,23 @@ def test_loop_artifact_argument_from_pipeline_channel(
...
+                    name='param1',
+                    channel_type='system.Artifact@0.0.1',
+                    task_name='task1',
+                    is_artifact_list=False,
+                ),
+        },
+    )
+    def test_loop_artifact_argument_from_single_pipeline_channel_raises_error(
+            self, channel):
+        self.assertRaises(ValueError,
```
connor-mccarthy, line 151: Can we use `assertRaisesRegex` to ensure we're catching the correct `ValueError` and that, throughout future changes, we don't "lose" this `ValueError` to a different downstream `ValueError`?

This is the [prevailing pattern](https://github.com/kubeflow/pipelines/blob/b6247fb8e489bb3dd8495fc6b4e9a181755c8896/sdk/python/kfp/dsl/for_loop_test.py#L194) in the SDK source.

connor-mccarthy (changes requested):

sdk/python/kfp/dsl/for_loop_test.py
```diff
@@ -151,6 +151,12 @@ def test_loop_artifact_argument_from_single_pipeline_channel_raises_error(
         self.assertRaises(ValueError,
                           for_loop.LoopArtifactArgument.from_pipeline_channel,
                           channel)
+        with self.assertRaisesRegex(
+                ValueError,
+                'Cannot iterate over a single Artifact using `dsl.ParallelFor`. Expected a list of Artifacts as argument to `items`.'
```
connor-mccarthy, line 153: Since `assertRaisesRegex` expects a regex rather than a literal string, we'll need use a raw string literal and escape all special regex characters:

```python
r'Cannot iterate over a single Artifact using `dsl\.ParallelFor`\. Expected a list of Artifacts as argument to `items`\.'
```

KevinGrantLee: Done

connor-mccarthy (approved):
/lgtm
/approve

Thank you, Kevin!

PR #10452, opened 2024-02-06
fix(sdk): fix bug where `dsl.OneOf` with multiple consumers cannot be compiled

connor-mccarthy:

**Checklist:**

### DIFF
--- a/sdk/python/kfp/compiler/compiler_test.py
+++ b/sdk/python/kfp/compiler/compiler_test.py
@@ -4786,6 +4786,12 @@ class TestDslOneOf(unittest.TestCase):
                 x = dsl.OneOf(print_task_1.outputs['a'],
                               print_task_2.outputs['a'])
                 print_artifact(a=x)
+                # test can be consumed multiple times from same oneof object
+                print_artifact(a=x)
+                y = dsl.OneOf(print_task_1.outputs['a'],
+                              print_task_2.outputs['a'])
+                # test can be consumed multiple times from different equivalent oneof objects
+                print_artifact(a=y)
 
         # hole punched through if
         self.assertEqual(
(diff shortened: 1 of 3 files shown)


sdk/python/kfp/compiler/compiler_test.py
```diff
@@ -4786,6 +4786,12 @@ def flip_coin_pipeline(execute_pipeline: bool):
                 x = dsl.OneOf(print_task_1.outputs['a'],
                               print_task_2.outputs['a'])
                 print_artifact(a=x)
+                # test can be consumed multiple times from same oneof object
+                print_artifact(a=x)
+                y = dsl.OneOf(print_task_1.outputs['a'],
```
chensun, line 4791: This seems to me redundant oneof, `x` and `y` would never differ from each other. In which scenario would we need `y` instead of reusing `x`?

connor-mccarthy: It is redundant, but it should be permitted, similar to how this would be permitted in normal Python:

```python
x = 1
y = 1
add(x, y)
```
even though `add(x, x)` is equivalent due to Python's use of call by value for immutable variables.

Prior to this PR, this would have failed. Since the PR adds a fix, it also adds test coverage.

chensun: It's fine to allow this redundancy. Albeit I think it's also reasonable to block it. Reasons include:
- `dsl.OneOf` is a special syntax that is paired with a conditional branch set. There're restrictions on which inputs it can accept--e.g. it needs to include all exclusive outputs, not something else nor a subset of such outputs. Arguably, its argument list is even redundant and skippable. The special use case warrants special treatment/requirement.
- I had this concern earlier that the pair seems loosely connected given it allows arbitrary code in between of the conditional branch and the oneof statement. Allowing multiple collection seems exposing this loose-pair further.
- Conceptually, the oneof resolution happens only once (at runtime). Enforcing a single oneof matches and highlights the actual behavior, and avoids possible misconception.

chensun (commented):
/lgtm

sdk/python/kfp/compiler/compiler_test.py
connor-mccarthy: Discussed offline and will merge as-is. As far as we can tell, there's no obvious cost to supporting this user code other than increased flexibility, which is both a cost and a benefit.

PR #10459, opened 2024-02-08
fix(backend): fixes "cannot save parameter" error message. Fixes #9678

### LINKED ISSUE #9678: cannot save parameter /tmp/outputs/condition
I've created a KubeFlow pipelines deployment on GKE with the manifests directly from the repository (the ones in the tutorial combined with code examples here led to validation errors). You can see the exact steps I'm taking to deploy the cluster, install KubeFlow, generate the compile / compile / install here: https://github.com/converged-computing/flux-operator-component/tree/add/component#kubeflow-on-gke

The issue I run into is on a run - I don't see any evidence that my workflow is running, but rather a weird error message about a parameter path in `/tmp/outputs` not existing. I also see that although I haven't yet defined any outputs, there is an automatic addition of a metadata one?

```console
I0622 23:49:18.534358      17 main.go:224] output ExecutorInput:{
  "inputs": {
    "parameterValues": {
      "command": "echo hello world",
      "image": "ghcr.io/flux-framework/flux-restful-api:latest",
      "local": true,
      "name": "hello-world-run-123",
      "namespace": "kubeflow",
      "nnodes": 2,
      "project": "llnl-flux"
    }
  },
  "outputs": {
    "outputFile": "/tmp/kfp_outputs/output_metadata.json"
  }
}
```

And the error

```console
time="2023-06-22T23:49:19.209Z" level=info msg="sub-process exited" argo=true error="<nil>"
time="2023-06-22T23:49:19.209Z" level=info msg="/tmp/outputs/pod-spec-patch -> /var/run/argo/outputs/parameters//tmp/outputs/pod-spec-patch" argo=true
time="2023-06-22T23:49:19.210Z" level=info msg="/tmp/outputs/cached-decision -> /var/run/argo/outputs/parameters//tmp/outputs/cached-decision" argo=true
time="2023-06-22T23:49:19.210Z" level=error msg="cannot save parameter /tmp/outputs/condition" argo=true error="open /tmp/outputs/condition: no such file or directory"
```

I am new to KubeFlow so I apologize for my ignorance! I really like the idea, conceptually, and hope that the implementation is able to empower me to build components that work - I'm still trying to get my feet wet for this basic development setup. Thanks!

hbelmiro:
Fixes: https://github.com/kubeflow/pipelines/issues/9678

I propose these changes as an initial solution, but I'm not sure it's the best one. So, feel free to recommend a better one and I'll be happy to implement it.

This PR modifies the driver so it creates the needed files with default values when `IterationCount` and `Condition` are `nil`.

The errors happen because we return an `Execution` without setting `IterationCount` and `Condition` here:

- https://github.com/kubeflow/pipelines/blob/87db18e3a1df08a23a71f872dc8dac6b4bfb9a95/backend/src/v2/driver/driver.go#L175
- https://github.com/kubeflow/pipelines/blob/87db18e3a1df08a23a71f872dc8dac6b4bfb9a95/backend/src/v2/driver/driver.go#L343

Which prevents it from creating the files here:

- https://github.com/opendatahub-io/data-science-pipelines/blob/410a82d45e6c51b259eeccab2e85f9b42d627880/backend/src/v2/cmd/driver/main.go#L194
- https://github.com/opendatahub-io/data-science-pipelines/blob/410a82d45e6c51b259eeccab2e85f9b42d627880/backend/src/v2/cmd/driver/main.go#L204

**Checklist:**

### DIFF
--- a/backend/src/v2/cmd/driver/execution_paths.go
+++ b/backend/src/v2/cmd/driver/execution_paths.go
@@ -1,0 +1,9 @@
+package main
+
+type ExecutionPaths struct {
+	ExecutionID    string
+	IterationCount string
+	CachedDecision string
+	Condition      string
+	PodSpecPatch   string
+}
--- a/backend/src/v2/cmd/driver/main.go
+++ b/backend/src/v2/cmd/driver/main.go
@@ -37,6 +37,9 @@ import (
 
 const (
 	driverTypeArg = "type"
+	ROOT_DAG      = "ROOT_DAG"
+	DAG           = "DAG"
+	CONTAINER     = "CONTAINER"
 )
 
 var (
@@ -160,12 +163,12 @@ func drive() (err error) {
 	var execution *driver.Execution
 	var driverErr error
 	switch *driverType {
-	case "ROOT_DAG":
+	case ROOT_DAG:
 		options.RuntimeConfig = runtimeConfig
 		execution, driverErr = driver.RootDAG(ctx, options, client)
-	case "DAG":
+	case DAG:
 		execution, driverErr = driver.DAG(ctx, options, client)
-	case "CONTAINER":
+	case CONTAINER:
 		options.Container = containerSpec
 		options.KubernetesExecutorConfig = k8sExecCfg
 		execution, driverErr = driver.Container(ctx, options, client, cacheClient)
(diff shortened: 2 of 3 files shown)


hbelmiro:
The failing test is unrelated.

```
AttributeError: module 'kfp_server_api' has no attribute 'ApiGetHealthzResponse'
make: *** [Makefile:18: sample-test] Error 1
```

rimolive:
@chensun @connor-mccarthy @zijianjoy Can we ignore this failed test and just merge?

Tomcli (approved):
/lgtm

nit: Maybe we can have a simple unit test to make sure in all scenarios, it should always write down the outputs for conditions and iterationcount?

hbelmiro:
> nit: Maybe we can have a simple unit test to make sure in all scenarios, it should always write down the outputs for conditions and iterationcount?

@Tomcli any ideas on how to do that?
I only found tests that use the KFP API to invoke tasks and check results. For this particular issue, I'd have to read the files inside the pods. Would you know any existing test that I can use as a reference?

Tomcli:
> > nit: Maybe we can have a simple unit test to make sure in all scenarios, it should always write down the outputs for conditions and iterationcount?
> 
> @Tomcli any ideas on how to do that? I only found tests that use the KFP API to invoke tasks and check results. For this particular issue, I'd have to read the files inside the pods. Would you know any existing test that I can use as a reference?

I don't think there's any test for the main.go. I'm thinking if you can implement a simple check file path exist function to make sure these files are exist after the test case. For the long term implementation, we may want to move this whole logic to use Argo's [execution plugin](https://argo-workflows.readthedocs.io/en/latest/executor_plugins/). This way we don't have to spend extra time to create new pods and write files. 

The test case is nice to have since the main.go can be implemented in a more efficient way in the future.

Tomcli:
@hbelmiro Can you redo your commits with git signoff? Kubeflow is moving to DCO CLA so all the commits from now on needed to be signed on git.

hbelmiro:
@Tomcli tests added. I just had to do a small refactor in `main.go` to make the code more testable.

Tomcli (commented):
thanks @hbelmiro 
/lgtm

PR #10427, opened 2024-01-24
feat(Backend + SDK): Update kfp backend and kubernetes sdk to support ImagePullSecrets

JosepSampe:
**Description of your changes:**
Update kfp backend and kubernetes sdk to support image pull secrets.

Part of https://github.com/kubeflow/pipelines/issues/10262
To be rebased and update go.mod after https://github.com/kubeflow/pipelines/pull/10410

**Checklist:**
- [x] The title for your pull request (PR) should follow our title convention. [Learn more about the pull request title convention used in this repository](https://github.com/kubeflow/pipelines/blob/master/CONTRIBUTING.md#pull-request-title-convention).

Issue #4454, opened 2020-09-02
Change adding V1Toleration, V1Affinity pattern in the container


subodh101:
What?
`add_toleration` and `add_affinity` requires users to add [V1Toleration](https://github.com/kubeflow/pipelines/blob/master/sdk/python/kfp/dsl/_container_op.py#L793) and [V1Affinity](https://github.com/kubeflow/pipelines/blob/master/sdk/python/kfp/dsl/_container_op.py#L804) objects.
Instead, pass the arguments and create the object in the SDK itself.

Why?
[Users](https://kubeflow.slack.com/archives/CE10KS9M4/p1598915504014300) and I myself found it difficult at first as I was expecting this to happen under the hood similar to other `add_*` functions.
Also sometimes people don't even know where to get the `V1Toleration` and `V1Affinity` objects.
Another important reason is that this feature itself is very important when it comes to cost optimization of the workloads. So should be straight forward of use.

How?
- By deprecating the current approach.
- and adding new changes.

Ark-kun:
1. Generally we try not to add such functions ourselves, but rely on existing structures for the maintenance. For example, task.container just uses the V1Container Kubernetes type. We had to add add_toleration and add_affinity to ContainerOp since they belong to the Kubernetes PodSpec, but we/Argo cannot support most of the PodSpec attributes. So we have to manually add support for those attributes.
2. These two functions have either been community We're not using these functions ourselves, so we do not know the user scenarios and do not know how to simplify them. We try to avoid the DSL being too Kubernetes-specific so adding multiple simplified variants of the same advanced function might be questionable.

We'd like to see your proposal for a simplified and more user-friendly version of these functions.

We could then try to add them to either ContainerOp or the extensions (similarly to `kfp.on_prem.mount_pvc`)

PR #10471, opened 2024-02-12
feat(Backend + SDK): Update kfp backend and kubernetes sdk to support tolerations

droctothorpe:
This PR adds support for tolerations to the driver and kubernetes_platform SDK. 

It is a follow up to https://github.com/kubeflow/pipelines/pull/10442, which updated the protobuf and Golang stub.

It follows the excellent example set by https://github.com/kubeflow/pipelines/pull/10427.

Huge thanks to @Tomcli for paving the way and sharing how to validate things end to end.

**Checklist:**


kubernetes_platform/proto/kubernetes_executor_config.proto
```diff
@@ -154,3 +154,11 @@ message ImagePullSecret {
     // Name of the image pull secret.
     string secret_name = 1;
 }
+
+message Toleration {
+    string key = 1;
+    string operator = 2;
+    string value = 3;
+    string effect = 4;
+    optional int64 toleration_seconds = 5;
```
droctothorpe, line 163: This needs to be optional otherwise not passing it results in it defaulting to 0 which leads to unintended effects.

backend/src/v2/driver/driver_test.go
```diff
@@ -671,3 +671,87 @@ func Test_extendPodSpecPatch_ImagePullSecrets(t *testing.T) {
 		})
 	}
 }
+
+func Test_extendPodSpecPatch_Tolerations(t *testing.T) {
```
edmondop, line 675: @droctothorpe as we discussed, we probably want to have a test for the failure case / invalid toleration? What's the behavior for example for a wrong combination of effect and value?

backend/src/v2/driver/driver.go
```diff
@@ -480,6 +480,28 @@ func extendPodSpecPatch(
...
 	}
 
+	if tolerations := kubernetesExecutorConfig.GetTolerations(); tolerations != nil {
+		var k8sTolerations []k8score.Toleration
+
+		glog.Infof("Tolerations passed: %+v", tolerations)
+
+		for _, toleration := range tolerations {
+			if toleration != nil {
+				k8sToleration := k8score.Toleration{
```
edmondop, line 490: I remembered we discussed the need to validate the correctness of this. I would move this out into a function that return a `(k8score.Toleration, error)` to keep the loop simple, and at each iteration either append to the array or return an error and fail early

backend/src/v2/driver/driver.go
```diff
@@ -480,6 +480,28 @@ func extendPodSpecPatch(
...
+					Key:               toleration.Key,
+					Operator:          k8score.TolerationOperator(toleration.Operator),
+					Value:             toleration.Value,
+					Effect:            k8score.TaintEffect(toleration.Effect),
```
edmondop, line 494: This is an enum, what's the behavior if we pass a "randomstring" ?

kubernetes_platform/python/test/snapshot/data/toleration.py
```diff
@@ -0,0 +1,45 @@
...
+from kubernetes.client import V1Toleration
+
+
+@dsl.component
+def comp():
+    pass
+
+
+@dsl.pipeline
+def my_pipeline():
```
edmondop, line 26: is `my_pipeline` a good name for this? I think giving a more meaningful name could help when searching maybe?

kubernetes_platform/python/kfp/kubernetes/__init__.py
```diff
@@ -23,15 +23,17 @@
     'add_node_selector',
     'add_pod_label',
     'add_pod_annotation',
-    'set_image_pull_secrets'
+    'set_image_pull_secrets',
+    'add_tolerations',
```
edmondop, line 27: Can you add that in the right location to respect the order for future modifications?

kubernetes_platform/python/test/unit/test_tolerations.py
```diff
@@ -0,0 +1,193 @@
...
+                            },
+                        }
+                    }
+                }
+            }
+        }
+
+
+@dsl.component
+def comp():
```
edmondop, line 192: is this required?

edmondop (changes requested):

backend/src/v2/driver/driver.go
edmondop: See https://github.com/kubeflow/pipelines/blob/002bfe96903ec1766151a9d6dada7f38303be481/backend/src/v2/driver/driver.go#L1259

droctothorpe: @connor-mccarthy curious to hear your thoughts on this. Should we add additional validation to the driver logic or let the K8s API handle validation? The main advantage of deferring validation to the K8s API is significantly reduced overhead IMO. Could go either way though.

backend/src/v2/driver/driver_test.go
droctothorpe: Contingent on implementing validation ourselves.

kubernetes_platform/python/test/snapshot/data/toleration.py
droctothorpe: This is just following suit with the other snapshots: https://github.com/kubeflow/pipelines/blob/master/kubernetes_platform/python/test/snapshot/data/node_selector.py#L25.

kubernetes_platform/python/kfp/kubernetes/__init__.py
droctothorpe: Mind elaborating? The sequence just mirrors the order in which the function was added (which is true of the ones above it as well).

kubernetes_platform/python/test/unit/test_tolerations.py
droctothorpe: Yup -- since it's referenced by each of the tests, e.g. [here](https://github.com/kubeflow/pipelines/pull/10471/files/002bfe96903ec1766151a9d6dada7f38303be481#diff-d0cc953dcb63eaf93dcc71bef9e5b13e48feecdcca9836476b6c60bf99b5b869R108).

kubernetes_platform/python/kfp/kubernetes/__init__.py
edmondop:
```
 'add_node_selector',
    'add_pod_label',
    'add_pod_annotation',
    'add_tolerations',
    'set_image_pull_secrets'
    'set_image_pull_secrets',
```

droctothorpe: Fixed 👍 .

backend/src/v2/driver/driver.go
droctothorpe: The workflow controller fails when it tries to create the pod:

```
Warning  WorkflowNodeError      6s    workflow-controller  Error node my-pipeline-xxlsx.root.comp.executor: task 'my-pipeline-xxlsx.root.com p.executor' errored: Pod "my-pipeline-xxlsx-1295791212" is invalid: spec.tolerations[0].operator: Unsupported value: "foo": supported values: "Equal", "Exists"
```

I think this ties back to the broader discussion of whether we want to reimplement validation or let the K8s API handle it.

backend/src/v2/driver/driver.go
droctothorpe: @edmondop we should be able to just pull in https://github.com/kubernetes/kubernetes/blob/master/pkg/apis/core/validation/validation.go#L3900 and call it in the driver logic! No duplication necessary. Shout out to @DWSR for the excellent suggestion!

droctothorpe: So `validationTolerations` is in core whereas the podspec declared by Kubeflow uses core/v1. I think validating in the driver doesn't really accomplish anything. The failure is still asynchronous (the run submission will still succeed) and will require parsing the workflow failure to debug. I suggest we defer validation to the K8s API.

droctothorpe:
Rebased and resolved merge conflicts.

kubernetes_platform/python/kfp/kubernetes/__init__.py
```diff
@@ -12,26 +12,28 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-__version__ = '1.1.0'
+__version__ = "1.1.0"
```
connor-mccarthy, line 15: Can we revert the quote style changes? [This pre-commit hook](https://github.com/kubeflow/pipelines/blob/1fcc68121cd030bd5f8301bf965ec969f170ad77/.pre-commit-config.yaml#L15), if used, will later override this.

kubernetes_platform/python/kfp/kubernetes/toleration.py
```diff
@@ -0,0 +1,52 @@
...
+from google.protobuf import json_format
+from kfp.dsl import PipelineTask
+from kfp.kubernetes import common
+from kfp.kubernetes import kubernetes_executor_config_pb2 as pb
+from kubernetes.client import V1Toleration
+
+
+def add_tolerations(
+    task: PipelineTask,
+    tolerations: List[V1Toleration],
```
connor-mccarthy, line 26: This is a departure from the existing pattern, where these task methods (/functions) are parameterized by Python primitives, rather than `kubernetes` objects. Is this preferred to keeping with the existing pattern?

WDYT about `add_toleration` (singular), for consistency as well? This makes parameterizing with primitives a bit more manageable.

kubernetes_platform/python/kfp/kubernetes/toleration.py
```diff
@@ -0,0 +1,52 @@
+# Copyright 2023 The Kubeflow Authors
```
connor-mccarthy, line 1: nit: 2024

kubernetes_platform/proto/kubernetes_executor_config.proto
```diff
@@ -154,3 +155,11 @@ message ImagePullSecret {
     // Name of the image pull secret.
     string secret_name = 1;
 }
+
+message Toleration {
+    string key = 1;
+    string operator = 2;
+    string value = 3;
+    string effect = 4;
```
connor-mccarthy, line 163: Should this be an `enum`? No strong preference from me -- just want to mention. The enum adds better validation at the cost of more maintenance effort.

kubernetes_platform/python/kfp/kubernetes/toleration.py
```diff
@@ -0,0 +1,52 @@
...
+
+    Args:
+        task: Pipeline task.
+        tolerations: A list of V1Tolerations defined using the Kubernetes Python Client.
+
+    Returns:
+        Task object with added tolerations.
+    """
+    tolerations_pb = []
+    for toleration in tolerations:
```
connor-mccarthy, line 38: I see in the unit tests that not all fields are required. Should we add validation for the optional/required fields?

One benefit of parameterizing with Python primitives is we can encode this information directly in the function interface.

connor-mccarthy (changes requested):
Thanks, @droctothorpe!

kubernetes_platform/python/kfp/kubernetes/toleration.py
droctothorpe: > This is a departure from the existing pattern, where these task methods (/functions) are parameterized by Python primitives, rather than kubernetes objects. Is this preferred to keeping with the existing pattern?

We used a Python Kubernetes client object for a couple of reasons.

First and foremost, that's [what v1 of the SDK does](https://github.com/kubeflow/pipelines/blob/1.8.22/sdk/python/kfp/dsl/_container_op.py#L948), and we wanted to provide as consistent of an experience as possible for users migrating from v1 to v2.

Secondly, node selectors and labels are fairly simple key value pairs, whereas tolerations, with five distinct fields, are substantially more complex -- sufficiently complex to justify using the Python Kubernetes client's `Toleration` object.

> WDYT about add_toleration (singular), for consistency as well? This makes parameterizing with primitives a bit more manageable.

Singular makes more sense. Commit to follow.

kubernetes_platform/proto/kubernetes_executor_config.proto
droctothorpe: Good question. Interestingly, the K8s protobuf tolerations message ([link](https://github.com/kubernetes/api/blob/master/core/v1/generated.proto#L5922-L5927)), which this message is modeled after, does **not** use `enum`. I think it might be a little strange if we hold our message to a higher standard of validation than their's, especially given that the K8s API handles validation for these objects on the backend. All that being said, happy to implement as `enum` if you think that makes more sense.

kubernetes_platform/python/kfp/kubernetes/toleration.py
droctothorpe: In the Python Kubernetes client, **none** of the fields are required!

```python
from kubernetes.client import V1Toleration

toleration = V1Toleration()
toleration
```

Output:

```python
{'effect': None,
 'key': None,
 'operator': None,
 'toleration_seconds': None,
 'value': None}
```

In fact, even on the backend, **none** of the fields are required (even the `key` field)! There are a few edge cases, handled [here](https://github.com/kubernetes/kubernetes/blob/master/pkg/apis/core/validation/validation.go#L3900), where if one field is set to a specific value, other fields need to be set to specific values, but no single field is required in all cases.

In light of that, and considering the complexity of the validation, I suggest we just let the K8s API handle the validation, much like we did in V1.

Here's the code we used to test a bunch of different variations of toleration fields if you want to try it out yourself:

```yaml
apiVersion: v1
kind: Pod
metadata:
  name: nginx
  labels:
    env: test
spec:
  containers:
  - name: nginx
    image: nginx
    imagePullPolicy: IfNotPresent
  tolerations:
  - key: "example-key"
    operator: "Equal"
    effect: "NoSchedule"
    tolerationSeconds: 10
```

droctothorpe:
@connor-mccarthy We addressed all feedback. We put it in a discrete commit to make review easier, but happy to squash if that's helpful. Thank you! 🙏

Tomcli:
fyi @droctothorpe your tests will be failing because the backend is pulling the kubernetes_platform package from github upstream instead of local packages.

droctothorpe:
> fyi @droctothorpe your tests will be failing because the backend is pulling the kubernetes_platform package from github upstream instead of local packages. This is why the PRs our team created have to at least separate into two PRs.

Doh. That makes sense. I'll rebase and resolve merge conflicts in the other PR this evening. Thanks, @Tomcli!

droctothorpe:
@Tomcli I rebased and re-opened https://github.com/kubeflow/pipelines/pull/10442. It should be good to go. Does it make sense to add the `kubernetes_platform` package to the `replace` directive either in general or for CI specifically? That would negate the need to submit these kinds of changes in discrete PRs. Just a thought.

kubernetes_platform/python/kfp/kubernetes/toleration.py
connor-mccarthy: > First and foremost, that's [what v1 of the SDK does](https://github.com/kubeflow/pipelines/blob/1.8.22/sdk/python/kfp/dsl/_container_op.py#L948), and we wanted to provide as consistent of an experience as possible for users migrating from v1 to v2.

I see the reason for wanting to align with v1. However, it comes at the (possibly greater) cost of making v2 internally consistency. Not going to block on this, but cc @Tomcli for a second opinion.

droctothorpe: More than happy to refactor. Just let me know. To clarify, the alternative you're referring to would be something along the following lines, right? I could definitely see how this could be more accessible for end users. 

```python
def add_toleration(
    key: Optional[str] = None, 
    operator: Optional[Literal["Equal", "Exists"]] = None,
    value: Optional[str] = None,
    effect: Optional[str] = None,
    toleration_seconds: Optional[int] = None
)
```

Tomcli: I would say it comes down to the user preference. The V1Toleration object is useful to standardize and get new updates from the k8s library automatically, but our users often don't understand the k8s object so we have created some helper function package for them to define these objects.

Having `def add_toleration(task: PipelineTask, tolerations: V1Toleration)` makes more sense to me in terms of v1 alignment. Then the add_toleration function could call another function like `def add_toleration_item` with the proto spec key above so that users can have both options.

Just extra information, the reason V1 was using the k8s object is because the SDK compiler was composing the containerSpec in K8S Python Object back then. Since now we are mapping to our own proto file, it make sense to give both options if we care about both backward compatibility and usability https://github.com/kubeflow/pipelines/issues/4454

kubernetes_platform/python/kfp/kubernetes/toleration.py
edmondop: What is the design rationale behind this approach? Is it calling for a pydantic usage?

kubernetes_platform/python/kfp/kubernetes/toleration.py
droctothorpe: That's a great idea. How about something like this?

```python
def add_toleration_object(
    task: PipelineTask,
    toleration: V1Toleration,
) -> PipelineTask:
    return add_toleration(
        task=task,
        key=toleration.key,
        operator=toleration.operator,
        value=toleration.value,
        effect=toleration.effect,
        toleration_seconds=toleration.toleration_seconds,
    )

def add_toleration(
    task: PipelineTask,
    key: Optional[str] = None,
    operator: Optional[Literal["Equal", "Exist"]] = None,  # https://github.com/kubernetes/kubernetes/blob/master/pkg/apis/core/types.go#L3137
    value: Optional[str] = None,
    effect: Optional[Literal["NoSchedule", "PreferNoSchedule", "NoExecute"] = None,  # https://github.com/kubernetes/kubernetes/blob/master/pkg/apis/core/types.go#L3146-L3148
    toleration_seconds: Optional[int] = None,
):
    msg = common.get_existing_kubernetes_config_as_message(task)
    msg.tolerations.append(
        pb.Toleration(
            key=key,
            operator=operator,
            value=value,
            effect=effect,
            toleration_seconds=toleration_seconds,
        )
    )
    task.platform_config["kubernetes"] = json_format.MessageToDict(msg)

    return task
```

I think that since we need to support 3.7 we can't use Literals, but we can use Unions or Enums if that's preferable.

Tomcli: That sounds good to me. thanks @droctothorpe

connor-mccarthy: Thank you, @Tomcli, for the input!

I do want to highlight that since the user code in v2 is syntactically breaking from the user code in v1, the "backward compatibility" with v1 you mention is really only in a stylistic sense, not a true "breaking changes" syntactic sense. We've _already_ departed from v1 conceptually, so using the k8s object in v2 is actually an _additional_ stylistic inconsistency (within v2) on top of the one that already (between v1 and v2). I still tend to think that (1) true backward compatibility (syntax), (2) internal consistency, and (3) backward stylistic consistency should be valued in that order.

That said, I will not block :). Just want to express a difference of opinion to inform the discussion.

In the spirit of a lower maintenance effort and better UX, I recommend we start with just one style, whichever that is. In my experience, multiple ways to do the same thing doesn't make for a better user experience, since the reason for two approaches is typically less obvious to the user than it is to the maintainer(s). Two approaches suggests that the implementations are functionally different in some way, which they are not.

Thank you both for the good discussion and I approve whichever approach you choose.

kubernetes_platform/proto/kubernetes_executor_config.proto
connor-mccarthy: SGTM! Thanks for looking into it.

kubernetes_platform/python/kfp/kubernetes/toleration.py
droctothorpe: I defer to you on this, Connor, because (a) you're an actual maintainer and I'm just some rando, and (b) your reasoning is unassailable. Just rebased and pushed up an implementation with primitives.

Two quick questions:

1. Is [this](https://github.com/kubeflow/pipelines/pull/10471/files#diff-72e60aa4ca84c37857bc27f55c975ae0f2be7ccabc51a3bf36530ded4332dc55R22-R25) okay? I saw you implemented something similar [in the DSL package](https://github.com/kubeflow/pipelines/blob/6fb997a611118d280325f499491a41799e5948f6/sdk/python/kfp/dsl/__init__.py#L61-L66).

2. The type hints and doc strings point users in the right direction, but they don't enforce operator and effect values. Is that okay? I think we can enforce with [typing.get_args](https://docs.python.org/3/library/typing.html#typing.get_args) if we want to.

kubernetes_platform/python/kfp/kubernetes/toleration.py
```diff
@@ -0,0 +1,81 @@
...
+        effect:
+            Effect indicates the taint effect to match. Empty means match all
+            taint effects. When specified, allowed values are NoSchedule,
+            PreferNoSchedule and NoExecute.
+        toleration_seconds:
+            TolerationSeconds represents the period of time the toleration
+            (which must be of effect NoExecute, otherwise this field is ignored)
+            tolerates the taint. By default, it is not set, which means tolerate
+            the taint forever (do not evict). Zero and negative values will be
+            treated as 0 (evict immediately) by the system.
```
droctothorpe, line 64: These are taken directly from the Python Kubernetes client docstrings.

droctothorpe:
Are the `kfp-kubernetes-execution-tests` failures related to this PR?

Tomcli:
@droctothorpe kubeflow-pipeline-backend-test is the only required test you failed. You can ignore the other non required test

Tomcli:
It seems like you haven't update the go.mod to the version with the new kubernete_platform spec you committed, so your go tests are failing.

droctothorpe:
Thank you so much, @Tomcli! Do you think it would be helpful to have some contribution documentation in the kubernetes_platform subfolder? We'd be happy to add it based on what we learned working on this PR.

droctothorpe:
Hm it looks like `github.com/klauspost/cpuid,https://github.com/klauspost/cpuid/blob/v1.3.1/LICENSE,MIT` was deleted for some reason. Let me try adding it back.

connor-mccarthy (approved):
/lgtm
/approve

Thank you, @droctothorpe!

droctothorpe:
> /lgtm /approve
> 
> Thank you, @droctothorpe!

I was debugging locally because I've been grappling with CI errors throughout the day due to the license checks in the image build and pushed up before I saw that this variant actually worked and you had approved it! Thank you for the lgtm and sorry for making you have to provide another one!

droctothorpe:
Locally, running `make all` from the `backend` directory results in the following in `backend/third_party_licenses/apiserver.csv`:

- Removes `github.com/klauspost/cpuid,https://github.com/klauspost/cpuid/blob/v1.3.1/LICENSE,MIT` 
- Replaces `golang.org/x/sys,https://cs.opensource.google/go/x/sys/+/v0.8.0:LICENSE,BSD-3-Clause` with `golang.org/x/sys/unix,https://cs.opensource.google/go/x/sys/+/v0.8.0:LICENSE,BSD-3-Clause` (note the addition of `unix`).

With this `apiserver.csv`, the license checks in `docker build -f backend/Dockerfile .` succeed **locally**, but they fail in CI. I wonder if this discrepancy is a function of that fact that I'm running `make all` on a mac whereas CI presumably runs on Linux. I wonder if it makes sense to execute `make all` from `backend` in a container, much like `make golang` in `kubernetes_platform`.

droctothorpe:
All required tests are finally passing. Thanks for your patience! CC @connor-mccarthy.

Tomcli (commented):
thanks @droctothorpe 
/lgtm

chensun (approved):

PR #10477, opened 2024-02-13
fix(sdk): fix compilation error when default contains escape sequence

### LINKED ISSUE #10438: [sdk] Newline characters in PythonComponent arguments break YAML compilation
### Environment

*  KFP version: N/A (Python SDK issue)
*  KFP SDK version: v2.5.0

### Steps to reproduce

I noticed that syntactically invalid YAML is produced when compiling a standard Kubeflow PythonComponent, which contains a "\n" string in function arguments. The bug only affects compilation to YAML (JSON works fine). 

### Expected result

Syntactically valid YAML should be produced, and the pipeline should be runnable on Kubeflow.

### Materials and Reference

[Example source Python](https://github.com/ndebuhr/kubeflows-hub/blob/e13dbbbab45c05610533fffdfca21b29003e09ac/split/components.py#L95)
[Example compiled YAML](https://github.com/ndebuhr/kubeflows-hub/blob/e13dbbbab45c05610533fffdfca21b29003e09ac/split/split-on-tokens-openai.yaml#L8)

---

connor-mccarthy:
Fixes #10438.

**Checklist:**

### DIFF
--- a/sdk/python/kfp/compiler/compiler_test.py
+++ b/sdk/python/kfp/compiler/compiler_test.py
@@ -2381,6 +2381,34 @@ class TestYamlComments(unittest.TestCase):
         # test reloaded comments
         self.assertIn(predicted_comment, reloaded_yaml_content)
 
+    def test_comment_with_newline_chars(self):
+
+        @dsl.component
+        def foo(
+            n: str = '\n',
+            r: str = '\r',
+            t: str = '\t',
+        ):
+            pass
+
+        with tempfile.TemporaryDirectory() as tmpdir:
+            pipeline_spec_path = os.path.join(tmpdir, 'output.yaml')
+            compiler.Compiler().compile(
+                pipeline_func=foo,
+                package_path=pipeline_spec_path,
+            )
+
+            with open(pipeline_spec_path, 'r+') as f:
+                reloaded_yaml_content = f.read()
+        expected_comment = textwrap.dedent("""\
+            # PIPELINE DEFINITION
+            # Name: foo
+            # Inputs:
+            #    n: str [Default: '\\n']
+            #    r: str [Default: '\\r']
+            #    t: str [Default: '\\t']""")
+        self.assertTrue(reloaded_yaml_content.startswith(expected_comment))
+
 
 class TestCompileThenLoadThenUseWithOptionalInputs(unittest.TestCase):
 
--- a/sdk/python/kfp/compiler/pipeline_spec_builder.py
+++ b/sdk/python/kfp/compiler/pipeline_spec_builder.py
@@ -2038,6 +2038,9 @@ def extract_comments_from_pipeline_spec(pipeline_spec: dict,
         'outputDefinitions': '# Outputs:'
     }
 
+    def escape_linebreak_characters(text: str) -> str:
+        return text.encode('unicode_escape').decode('ascii')
+
     def collect_pipeline_signatures(root_dict: dict,
                                     signature_type: str) -> List[str]:
         comment_strings = []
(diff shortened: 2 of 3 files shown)


sdk/python/kfp/compiler/compiler_test.py
```diff
@@ -2381,6 +2381,29 @@ def my_pipeline(sample_input1: bool = True,
         # test reloaded comments
         self.assertIn(predicted_comment, reloaded_yaml_content)
 
+    def test_comment_with_newline_chars(self):
+
+        @dsl.component
+        def foo(n: str = '\n', r: str = '\r'):
+            ...
```
chensun, line 2388: nit: Though compiler don't look into the function body, but I'd suggest at least keep it a valid Python syntax, e.g.: use `pass` instead of `...`.

sdk/python/kfp/compiler/pipeline_spec_builder.py
```diff
@@ -2038,6 +2038,9 @@ def extract_comments_from_pipeline_spec(pipeline_spec: dict,
         'outputDefinitions': '# Outputs:'
     }
 
+    def escape_linebreak_characters(text: str) -> str:
+        return text.replace('\n', '\\n').replace('\r', '\\r')
```
chensun, line 2042: I think this fix is incomplete. While this fixes the reported issue, others non-printables like `'\t'` may not break the yaml, but they won't display correctly in the comment string.

sdk/python/kfp/compiler/pipeline_spec_builder.py
```diff
@@ -2079,8 +2082,8 @@ def collect_pipeline_signatures(root_dict: dict,
                 string = '#    ' + signature['name'] + ': ' + signature[
                     'parameterType']
                 if 'defaultValue' in signature:
-                    string += ' [Default: ' + str(
-                        signature['defaultValue']) + ']'
+                    string += ' [Default: ' + escape_linebreak_characters(
```
chensun, line 2085:
```suggestion
                    string += ' [Default: ' + repr(
```

sdk/python/kfp/compiler/pipeline_spec_builder.py
```diff
@@ -2079,8 +2082,8 @@ def collect_pipeline_signatures(root_dict: dict,
...
+                        str(signature['defaultValue'])) + ']'
```
chensun, line 2086:
```suggestion
                       signature['defaultValue']) + ']'
```

sdk/python/kfp/compiler/pipeline_spec_builder.py
chensun: I think this might be a simpler and more complete fix?

chensun (changes requested):

sdk/python/kfp/compiler/compiler_test.py
connor-mccarthy: Done.

sdk/python/kfp/compiler/pipeline_spec_builder.py
connor-mccarthy: Explicitly type casting is needed here, since the default value will not be implicitly casted when string concatenation is performed via addition.

Since converting all uses of addition to f-strings in this function would distract from the intent of the PR, I'll omit this change.

sdk/python/kfp/compiler/pipeline_spec_builder.py
chensun: Actually I'm not sure which one is more faithful, extra quotes vs. extra backslash. 
Extra quotes would be more consistent as it's universally applied on all strings. Extra backslash is less intrusive and only visible on certain cases.

chensun (commented):
/lgtm

PR #10494, opened 2024-02-15
fix(sdk): Prevents dsl.ParallelFor over single parameter from compiling.

KevinGrantLee:

Prevents the following pipeline from compiling
```
@dsl.component
def str_identity(s: str) -> str:
    return s


@dsl.pipeline
def my_pipeline():
  single_param = str_identity(s='string')
  with dsl.ParallelFor(items=single_param.output) as item:
      str_identity(s=item)
```

The component `str_identity` outputs a `str` so routing this to `dsl.ParallelFor` does not make sense.

**Checklist:**


sdk/python/kfp/dsl/for_loop_test.py
```diff
@@ -144,6 +144,24 @@ def test_loop_parameter_argument_from_pipeline_channel(
...
+        },)
+    def test_loop_parameter_argument_from_single_pipeline_channel_raises_error(
+            self, channel):
+        with self.assertRaisesRegex(
+                ValueError,
+                r'Cannot iterate over a single Parameter using `dsl\.ParallelFor`\. Expected a list of Parameters as argument to `items`\.'
+        ):
+            loop_argument = for_loop.LoopParameterArgument.from_pipeline_channel(
+                channel)
+
```
connor-mccarthy, line 164: Can we add a test for this with some example user code to `compiler_test.py`?

sdk/python/kfp/dsl/for_loop.py
```diff
@@ -188,6 +191,13 @@ def from_pipeline_channel(
         compilation progress in cases of unknown or missing type
         information.
         """
+        if not isinstance(channel, for_loop.LoopArgumentVariable):
```
connor-mccarthy, line 194: For my own understanding, is this condition needed for nested loops? Perhaps add a comment to indicate its purpose.

Can we ensure we have test coverage for both cases (i.e., `True` and `False`)?

connor-mccarthy (changes requested):
Thank you, @KevinGrantLee!

sdk/python/kfp/dsl/for_loop_test.py
KevinGrantLee: Done, will add the corresponding test for single artifacts too in the next pr

sdk/python/kfp/dsl/for_loop.py
KevinGrantLee: Per offline convo, ignoring nested loop case (channel is LoopArgumentVariable). Added todo

sdk/python/kfp/dsl/for_loop.py
```diff
@@ -188,6 +191,16 @@ def from_pipeline_channel(
...
         """
+        # TODO: if is LoopArgumentVariable, check if nested items are lists
+        if not isinstance(channel, LoopArgumentVariable):
+            type_name = type_annotations.get_short_type_name(
+                channel.channel_type)
+            parameter_type = type_utils.PARAMETER_TYPES_MAPPING[
+                type_name.lower()]
+            if parameter_type != type_utils.LIST:
+                raise ValueError(
+                    'Cannot iterate over a single Parameter using `dsl.ParallelFor`. Expected a list of Parameters as argument to `items`.'
```
connor-mccarthy, line 202:
```suggestion
                    'Cannot iterate over a single parameter using `dsl.ParallelFor`. Expected a list of parameters as argument to `items`.'
```

Our error messages do not capitalize the "P" in parameter elsewhere in the SDK.

sdk/python/kfp/dsl/for_loop.py
```diff
@@ -188,6 +191,16 @@ def from_pipeline_channel(
         compilation progress in cases of unknown or missing type
         information.
         """
+        # TODO: if is LoopArgumentVariable, check if nested items are lists
```
connor-mccarthy, line 194: Up to you whether to keep or remove the `TODO`, but I don't think this is feasible within the current system. Perhaps this should just be a note about the case we cannot validate, rather than a TODO.

sdk/python/kfp/dsl/for_loop.py
```diff
@@ -16,7 +16,10 @@
 import re
 from typing import Any, Dict, List, Optional, Union
 
+from kfp.dsl import for_loop
```
connor-mccarthy, line 19: This import is unused

connor-mccarthy (changes requested):
Thanks, Kevin!

connor-mccarthy (approved):
/lgtm
/approve

Thank you, @KevinGrantLee!

PR #10502, opened 2024-02-19
feat(backend): Upgrade go version to 1.20

### LINKED ISSUE #10487: [feature] Upgrade go version to 1.19+
### Feature Area
KFP is using go version 1.13 which is very old. Although the image is using go 1.20, we should update the go.mod and the core program to go 1.19+. (Ideally 1.21)

---

Love this idea? Give it a 👍.

Tomcli:
fixes #10487 
- Update go version to 1.20
- Update dependencies such as go clients and cel packages to address critical CVEs while compatible with Argo 3.3.
- With the updated CEL package, CEL expression now can overload between double and integer.

**Checklist:**


backend/Dockerfile
```diff
@@ -13,7 +13,7 @@
 # limitations under the License.
 
 # 1. Build api server application
-FROM golang:1.20.4-buster as builder
+FROM golang:1.21.7-bookworm as builder
```
Tomcli, line 16: debian:stable is now based on the bookworm image. https://www.debian.org/releases/

backend/src/v2/expression/expression_test.go
```diff
@@ -154,8 +154,9 @@ func TestCondition(t *testing.T) {
 		name:      "errorOnTypeMismatch",
 		input:     input,
 		condition: "inputs.parameter_values['num'] == 1",
-		// Note, inputs.parameter_values['num'] is double type, but 1 is integer type.
-		err: "no such overload",
+		// https://github.com/google/cel-spec/blob/master/doc/langdef.md#numbers
+		// overload double and integer is now supported, so the result is true
+		output: true,
```
Tomcli, line 159: New CEL packages now can overload between double and integer type.
https://github.com/google/cel-spec/blob/master/doc/langdef.md#numbers

go.mod
```diff
@@ -4,65 +4,194 @@ require (
...
+	k8s.io/klog/v2 v2.100.1 // indirect
+	k8s.io/kube-openapi v0.0.0-20230515203736-54b630e78af5 // indirect
+	sigs.k8s.io/json v0.0.0-20221116044647-bc3834ca7abd // indirect
+	sigs.k8s.io/structured-merge-diff/v4 v4.2.3 // indirect
+)
+
 replace (
+	k8s.io/api => k8s.io/api v0.25.9
+	k8s.io/apimachinery => k8s.io/apimachinery v0.26.5
+	k8s.io/client-go => k8s.io/client-go v0.25.9
```
Tomcli, line 191: Need to lock in at client-go 0.25.x to support Argo 3.3.

go.mod
```diff
@@ -4,65 +4,194 @@ require (
...
+	google.golang.org/protobuf v1.31.0
 	gopkg.in/yaml.v3 v3.0.1
-	k8s.io/api v0.24.3
-	k8s.io/apimachinery v0.24.3
-	k8s.io/client-go v0.24.3
-	k8s.io/code-generator v0.23.3
-	k8s.io/kubernetes v0.17.9
+	k8s.io/api v0.27.2
+	k8s.io/apimachinery v0.27.3
+	k8s.io/client-go v0.27.2
```
Tomcli, line 53: This is the ideal version we want to achieve for supporting the latest k8s version with the new dependencies. However, both Argo and Tekton have some issues running 0.27+, so we are using replace to lock the k8s version until Argo and Tekton addressed it.

Tomcli:
Hi @chensun, it seems like the presubmit-backend-test go binary is 1.20. Do you know how to upgrade it to 1.21? If it's too much work, I can downgrade my current PR to use go 1.20 as well.

chensun:
> Hi @chensun, it seems like the presubmit-backend-test go binary is 1.20. Do you know how to upgrade it to 1.21? If it's too much work, I can downgrade my current PR to use go 1.20 as well.

I created https://github.com/GoogleCloudPlatform/oss-test-infra/pull/2205

Tomcli:
Hi @chensun, I can't reproduce the error `curl: (6) getaddrinfo() thread failed to start` from the command `'/bin/sh -c curl https://dl.google.com/dl/cloudsdk/release/google-cloud-sdk.tar.gz > /tmp/google-cloud-sdk.tar.gz'` in the integration ci image running on `golang:1.21.7`.

I looked up online and it seems like could be the DNS or firewall issue on the server itself. But this error only happens when in the kubeflow gcp environment running the golang:1.21.7 image. Do you recommend to move the go version back to 1.20 so we don't have to update the integration ci image?

Here is some related issues on the curl getaddrinfo()
https://repost.aws/questions/QU3jvWeM8OTCuCePQQa_WFKw/curl-error-6-getaddrinfo-thread-failed-to-start
https://stackoverflow.com/questions/59911649/fatal-unable-to-access-link-getaddrinfo-thread-failed-to-start

Tomcli:
@chensun I reduced the scope of this PR to only update the go version to 1.20 and the relevant dependencies. This passed the test without breaking the setup of the CI itself.

We can follow up in another PR if we feel like we want to upgrade everything to go 1.21. As for the Tekton backend support, go 1.20 is enough right now.

backend/Dockerfile
Tomcli: No updating dockerfile in this PR anymore

chensun (approved):
/lgtm
/approve

Issue #7345, opened 2022-02-23
[sdk] Can't use create_component_from_func with pip packages when running as non-root


skogsbrus:
### Environment

*  KFP version: 1.7 (KF 1.4)
*  KFP SDK version: 1.6.6
*  All dependencies version:

```
kfp                              1.6.6
kfp-pipeline-spec                0.1.13
kfp-server-api                   1.7.1
```

### Steps to reproduce

Background:

* Due to security concerns it's a bad idea to run containers as root. 
* For composability and maintenance, it's a good idea to define small & modular KFP components.

With this in mind, I wish to report that `create_components_from_func` does not work as expected when the container is run as a non-root user and when the `packages_to_install` parameter is used to add some runtime dependencies.

To reproduce, see attached pipeline definition at the bottom.

When this pipeline is run, the following output is seen in Kubeflow:

```
WARNING: The directory '/.cache/pip' or its parent directory is not owned or is not writable by the current user. The cache has been disabled. Check the permissions and owner of that directory. If executing pip with sudo, you should use sudo's -H flag.
ERROR: Could not install packages due to an OSError: [Errno 13] Permission denied: '/.local'
Check the permissions.
WARNING: The directory '/.cache/pip' or its parent directory is not owned or is not writable by the current user. The cache has been disabled. Check the permissions and owner of that directory. If executing pip with sudo, you should use sudo's -H flag.
ERROR: Could not install packages due to an OSError: [Errno 13] Permission denied: '/.local'
Check the permissions.
Error: exit status 1
```

### Expected result

The correct behaviour here would be for packages to be installed in a location that's writable by non-root users. As a direct consequence, that location would also to have to be added to PYTHONPATH.

With the attached pipeline definition, `kfp.components._python_op._get_packages_to_install_command` today produces the following yaml:

```
      - sh
      - -c
      - (PIP_DISABLE_PIP_VERSION_CHECK=1 python3 -m pip install --quiet --no-warn-script-location
        'tqdm' || PIP_DISABLE_PIP_VERSION_CHECK=1 python3 -m pip install --quiet --no-warn-script-location
        'tqdm' --user) && "$0" "$@"
      - sh
      - -ec
      - |
        program_path=$(mktemp)
        printf "%s" "$0" > "$program_path"
        python3 -u "$program_path" "$@"
      - |
        def hello_world():
            import tqdm
            print("Hello world!")

        import argparse
        _parser = argparse.ArgumentParser(prog='Hello world', description='')
[...]

Issue #5673, opened 2021-05-18
[Launcher] support non-root containers in v2 compatible mode


Bobgy:
Currently, because the launcher writes input artifacts to paths like:

* `/gcs/xxx`
* `/minio/xxx`
* `/s3/xxxx`

These paths are not accessible by non-Root users by default.

When using a component with non-Root image, launcher fails when preparing input/output artifacts.

Because `/gcs/xxx` is currently a contract for KFP v2 python component wrappers, we cannot change to a different path like `/tmp/gcs/xxx` etc.

Ark-kun:
It's possible to specify the user that the container should run under. Backend can auto-set that on all containers.

Bobgy:
@Ark-kun do you mean our backend sets all containers to run as root by default?

Nagarajj:
@Bobgy What is the plan to support non-root containers ? it is difficult to do any PoC with V2 in an enterprise setup due to this. Thanks.

Bobgy:
Hi @Nagarajj, this isn't on my priority list right now.
So welcome contributions on this! Feel free to discuss if you need any help.
Or if there are more people chiming in that this is important, we can re-prioritize.

Bobgy:
I think what needs to be done is basically making sure all the local dirs v2 compatible mode launcher reads from/writes to should be accessible by all non-root users.

@Nagarajj may I confirm do you require all containers to run as non-root? or is it OK for some KFP system containers to be root? e.g. we have a `kfp-launcher` init container that copies the launcher binary to a shared emptyDir volume. Do you want it to be non-root too?

Bobgy:
I'd imagine the easiest solution/workaround is to
* mount another emptyDir volume for `/gcs` folder, so it's accessible by all users (because as mentioned, it's part of container contract -- we cannot change the /gcs path)
* change path of `/minio/xxx` and `/s3/xxx` to the same folder as the emptyDir containing the launcher binary

For best practice, I think we should move the volume with launcher binary to `/var/run/kfp`, because `/var/run` is runtime variable data. https://refspecs.linuxfoundation.org/FHS_3.0/fhs/ch05s13.html

So here's the design:
* an empty dir volume is mounted at `/var/run/kfp` (accessible to all users)
    * kfp launcher binary is copied to `/var/run/kfp/bin/launcher`
[...]

Issue #10397, opened 2024-01-15
[backend]  Unable to create directory in Minio when using Artifacts: Permission denied


jmaunon:
Hi Developers

I have tried to create a simple pipeline using and transfering data using "built-in" `artifacts` approach without success.
Difficult to say what is hapenning but I have found similar issues in other threads. 

Please, if you know a manual patch, let us know. I see `artifacts` a core solution/approach.

cc: @juliusvonkohout , @chensun 

I am aware that there are some issues related, but I do not see a final solution or alternative patch. See: #6530 , https://github.com/kubeflow/manifests/issues/2573, https://github.com/kubeflow/pipelines/issues/7629

### Environment

*  How did you deploy Kubeflow Pipelines (KFP): [https://github.com/kubeflow/manifests](https://github.com/kubeflow/manifests) latest tag: v.1.8.0

*  KFP version: 
According to the `Readme` from manifest repo: KFP 2.0.3
*  KFP SDK version:  2.6.0

### Steps to reproduce

I get a `permission denied` error when using `Artifacts`. 

Snippet of code:

```
@dsl.component(base_image="kubeflownotebookswg/jupyter-pytorch-full:v1.8.0-rc.0")
def download_data(test_path: Output[Dataset]):
    
    import torch
    
    from torchvision.transforms import ToTensor
    from torchvision.datasets import MNIST
    
    mnist_test  = MNIST(".", download=True, train=False, transform=ToTensor())
        
    with open(test_path.path, "wb") as f:
        torch.save(mnist_test,f)

@dsl.pipeline(
    name='mnist',
    description='Detect digits',
)
def run():
    step_1 = download_data()

client.create_run_from_pipeline_func(run)
```

Associated logs:

`failed to execute component: unable to create directory "/minio/mlpipeline/v2/artifacts/mnist/43f760f9-b638-4129-87fe-602e24076beb/download-data" for output artifact "test_path": mkdir /minio: permission denied`

### Expected result

Work without issus

### Materials and Reference

---

Impacted by this bug? Give it a 👍.

juliusvonkohout:
Please use the final 1.8 image, not jupyter-pytorch-full:v1.8.0-rc.0 and join the biweekly KFP meeting to discuss this.

juliusvonkohout:
You should also try to update from KFP 2.0.3 to 2.0.5 first.

jmaunon:
Thans for the reply @juliusvonkohout . I write here my findings:
[...]

PR #10538, opened 2024-03-04
feat(backend): mount EmptyDir volumes for launcher write locations

gregsheremeta:
Launcher writes input artifacts to root paths `/gcs`, `/minio`, and `/s3`. These paths are not accessible by non-root users by default, which is problematic in locked-down Kubernetes installations and/or OpenShift. [`/gcs` is currently a contract for KFP v2 python component wrappers, so the path cannot be changed.](https://github.com/kubeflow/pipelines/issues/5673#issuecomment-901045901)

Mount an EmptyDir scratch volume to these paths to work around this.

Additionally, `/.local` and `/.cache` are written to by pip, so add EmptyDir mounts for those too.

Fixes: #5673
Fixes: #7345

### DIFF
--- a/backend/src/v2/compiler/argocompiler/container.go
+++ b/backend/src/v2/compiler/argocompiler/container.go
@@ -26,11 +26,21 @@ import (
 )
 
 const (
-	volumeNameKFPLauncher = "kfp-launcher"
-	DefaultLauncherImage  = "gcr.io/ml-pipeline/kfp-launcher@sha256:80cf120abd125db84fa547640fd6386c4b2a26936e0c2b04a7d3634991a850a4"
-	LauncherImageEnvVar   = "V2_LAUNCHER_IMAGE"
-	DefaultDriverImage    = "gcr.io/ml-pipeline/kfp-driver@sha256:8e60086b04d92b657898a310ca9757631d58547e76bbbb8bfc376d654bef1707"
-	DriverImageEnvVar     = "V2_DRIVER_IMAGE"
+	volumeNameKFPLauncher   = "kfp-launcher"
+	DefaultLauncherImage    = "gcr.io/ml-pipeline/kfp-launcher@sha256:80cf120abd125db84fa547640fd6386c4b2a26936e0c2b04a7d3634991a850a4"
+	LauncherImageEnvVar     = "V2_LAUNCHER_IMAGE"
+	DefaultDriverImage      = "gcr.io/ml-pipeline/kfp-driver@sha256:8e60086b04d92b657898a310ca9757631d58547e76bbbb8bfc376d654bef1707"
+	DriverImageEnvVar       = "V2_DRIVER_IMAGE"
+	gcsScratchLocation      = "/gcs"
+	gcsScratchName          = "gcs-scratch"
+	s3ScratchLocation       = "/s3"
+	s3ScratchName           = "s3-scratch"
+	minioScratchLocation    = "/minio"
+	minioScratchName        = "minio-scratch"
+	dotLocalScratchLocation = "/.local"
+	dotLocalScratchName     = "dot-local-scratch"
+	dotCacheScratchLocation = "/.cache"
+	dotCacheScratchName     = "dot-cache-scratch"
 )
 
 func (c *workflowCompiler) Container(name string, component *pipelinespec.ComponentSpec, container *pipelinespec.PipelineDeploymentConfig_PipelineContainerSpec) error {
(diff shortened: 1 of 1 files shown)


Tomcli (commented):
Thanks @gregsheremeta, just wondering is there anyway we can reduce the amount of the volume mount definition? Due to the k8s yaml size limitation, we won't able to fit more tasks if the individual container definition is too big.

Tomcli (changes requested):
@gregsheremeta You also need to update the unit tests because you modified the container spec in this PR.

andraxin:
This will not solve the `Permission denied` error reported in https://github.com/kubeflow/pipelines/issues/10397,
unless a version of the v2 launcher that includes [this fix](https://github.com/kubeflow/pipelines/commit/4003e562713bd04fa94387d8b53dfbe3cf31cb12) is used.

_**BTW: where/how do I tell Kubeflow which launcher image to use?**_

That's because the root cause for the original error is that the installer creates
(or used to create, after the fix)   directories with mode `0644`, i.e. `drw-r--r--`,
which makes it impossible for unprivileged users to create subdirectories within,
but doesn't hamper `root`...

### Brief Demo
```console
$ docker run --rm -it alpine
/ # adduser -D demo
/ # su - demo
05decfa077fd:~$ mkdir -m 0644 demo
05decfa077fd:~$ mkdir demo/test
mkdir: can't create directory 'demo/test': Permission denied
05decfa077fd:~$ 
/ # mkdir /home/demo/demo/test
/ # ls -lR /home/demo
/home/demo:
total 4
drw-r--r--    3 demo     demo          4096 Apr 14 11:16 demo

/home/demo/demo:
total 4
drwxr-xr-x    2 root     root          4096 Apr 14 11:16 test

/home/demo/demo/test:
total 0
/ # 
```

BTW, that also explains why creating alternate images with world-read-write
`/minio` pre-created (i.e. `RUN mkdir -m 1777 /minio`) doesn't help.

I tracked this down by [mounting a PVC](https://kfp-kubernetes.readthedocs.io/en/kfp-kubernetes-1.0.0/source/kubernetes.html#kfp.kubernetes.mount_pvc) at `/minio` and observing permissions
on the first level (that does get created).

rimolive:
> unless a version of the v2 launcher that includes [this fix](https://github.com/kubeflow/pipelines/commit/4003e562713bd04fa94387d8b53dfbe3cf31cb12) is used.

This fix was introduced in KFP 2.0.2

andraxin:
> > unless a version of the v2 launcher that includes [this fix](https://github.com/kubeflow/pipelines/commit/4003e562713bd04fa94387d8b53dfbe3cf31cb12) is used.
> 
> This fix was introduced in KFP 2.0.2

Yes, I'm well aware. GitHub shows that quite prominently.
```
fix(backend): OutPutPath dir creation mode Fixes https://github.com/kubeflow/pipelines/issues/7629 (https://github.com/kubeflow/pipelines/pull/9946)
[master](https://github.com/kubeflow/pipelines) ([#9946](https://github.com/kubeflow/pipelines/pull/9946))
[sdk-2.7.0](https://github.com/kubeflow/pipelines/releases/tag/sdk-2.7.0)
[2.0.2](https://github.com/kubeflow/pipelines/releases/tag/2.0.2)
```

BUT, that doesn't really help me or answer my question.
The launcher that's spawned by default in the Kubeflow v1.7.x
deployment I **have to** use apparently pre-dates that; and,
I have no idea how/if Kubeflow can be instructed to grab
a newer launcher with the fix.

rimolive:
Kubeflow 1.7 comes with an alpha release of KFP. You should either fully upgrade your kubeflow installation to 1.8 or you can upgrade individually kfp. Just a reminder that if you choose the later one, you need to use the multi-user manifests, as you probably deployment Kubeflow w/ multi-user support.

andraxin:
> Kubeflow 1.7 comes with an alpha release of KFP. You should either fully upgrade your kubeflow installation to 1.8 or you can upgrade individually kfp. Just a reminder that if you choose the later one, you need to use the multi-user manifests, as you probably deployment Kubeflow w/ multi-user support.

I guess I wasn't sufficiently clear, when I said I **have to** use Kubeflow 1.7.
I cannot upgrade the installation nor apply any manifests at the cluster level.
Applying manifest in my profile's namespace is about as far as I can reach.

My re-stated question becomes: as a Kubeflow **user**, can I somehow instruct
Kubeflow to use a newer launcher image to run my pipelines?

BTW, I _can_ use custom images for my pipeline stages with world-read-write
`/gcs`, `/minio`, and `/s3`, which renders these emptyDir volume mounts moot.
**If** I could only tell Kubeflow to use a launcher with the fix, that is.

rimolive:
> My re-stated question becomes: as a Kubeflow user, can I somehow instruct
Kubeflow to use a newer launcher image to run my pipelines?

If, and only if, you upgrade your KFP installation to at least 2.0.5 release, you can [parametrize the launcher and driver images](https://github.com/kubeflow/pipelines/commit/23a2bc3ba370e28a5ef5f0e0df1e40ec8f7888b9) which include that configuration.

andraxin:
> > My re-stated question becomes: as a Kubeflow user, can I somehow instruct
> > Kubeflow to use a newer launcher image to run my pipelines?
> 
> If, and only if, you upgrade your KFP installation to at least 2.0.5 release, you can [parametrize the launcher and driver images](https://github.com/kubeflow/pipelines/commit/23a2bc3ba370e28a5ef5f0e0df1e40ec8f7888b9).

I see.  Thanks for the info, and your patience! :smirk: 

Given that the code changes you linked appear to be
in backend Golang code, what you call "KFP installation"
probably includes more than just `pip install kfp==2.0.2`
inside a notebook in the existing Kubeflow 1.7 cluster.
Is that correct?

rimolive:
Yes, this is not a simple SDK upgrade or code changes. You will need to upgrade your whole KFP installation.

thesuperzapper:
@chensun @HumairAK @Tomcli we definitely need to prioritize fixing this issue, because it's pretty bad to have a hard requirement on root container images, see https://github.com/kubeflow/pipelines/issues/10397 for an example of user having this problem.

droctothorpe:
FYI: https://github.com/kubeflow/pipelines/issues/10397#issuecomment-2140756532

HumairAK:
@gregsheremeta  is currently out, 

due to the expressed urgency and prioritzation of this, I'm picking this PR up here: https://github.com/kubeflow/pipelines/pull/10857

please forward any further discussions to the pr above

/close

PR #10547, opened 2024-03-07
chore: Change stalebot rules

rimolive:
daysUntilClose is set to 90 days. This is too much for an issue to keep opened as there is daysUntilStale set to 90 days. That would keep an issue opened for 6 months, and now there are +600 issues opened in the repository.

We need to start working on keeping the repository healthy.

**Checklist:**

### DIFF
--- a/.github/stale.yml
+++ b/.github/stale.yml
@@ -2,9 +2,9 @@
 # https://probot.github.io/apps/stale/
 #
 # Number of days of inactivity before an issue becomes stale
-daysUntilStale: 90
+daysUntilStale: 60
 # Number of days of inactivity before a stale issue is closed
-daysUntilClose: 90
+daysUntilClose: 21
 # Issues with these labels will never be considered stale
 exemptLabels:
   - lifecycle/frozen
--- a/.github/workflows/stale.yml
+++ b/.github/workflows/stale.yml
@@ -21,10 +21,8 @@ jobs:
     - uses: actions/stale@v5
       with:
         repo-token: ${{ secrets.GITHUB_TOKEN }}
-        days-before-stale: -1
-        days-before-close: -1
-        days-before-issue-stale: 90
-        days-before-issue-close: 90
+        days-before-stale: 60
+        days-before-close: 21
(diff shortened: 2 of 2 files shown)


rimolive:
cc @Tomcli @chensun @zijianjoy

gregsheremeta:
:+1:  This is a good step towards getting the community back on a healthy track.

.github/stale.yml
```diff
@@ -4,7 +4,7 @@
 # Number of days of inactivity before an issue becomes stale
 daysUntilStale: 90
 # Number of days of inactivity before a stale issue is closed
-daysUntilClose: 90
+daysUntilClose: 7
```
Tomcli, line 7: Maybe we can use the default value of slate bot which is 21 days? The bigger problem here is many issues are not slate anymore when they reopen.
https://github.com/etcd-io/etcd/blob/main/.github/stale.yml#L9

rimolive: We can always use `/lifecycle frozen` for important issues that needs more time, but the fact is that we currently have [~310 issues](https://github.com/kubeflow/pipelines/labels?sort=count-desc) with `lifecycle/stale` and ~90 issues with `lifecycle/frozen`. This number is already unmanageable for the number of contributors we have.

Let me know your thoughts.

rimolive: As for the "many issues are not slate anymore when they reopen", we can also consider setting daysUntilStale to the default 60 days.

Tomcli: Thanks @rimolive, shorten the `daysUntilStale` and having a longer `daysUntilClose` would work better. This way we can warn the issue creator on issues going to stale without closing it too soon.

rimolive: Let me know if these parameters look good:
`daysUntilStale`: 60
`daysUntilClose`: 21

Tomcli (commented):
/lgtm

chensun (approved):
Let's test out these parameters. Thanks for taking the lead!

/lgtm
/approve

PR #8526, opened 2022-12-05
feature(sdk): Upgrade kubernetes library to v19 to get access to Ephe…

abaland:
**Description of your changes:**
Upgrade the SDK v1 kubernetes version specification to allow v19. This allows usage of ephemeral volumes (https://github.com/kubernetes-client/python/blob/master/kubernetes/docs/V1Volume.md > ephemeral) which becomes available with v19.

**Checklist:**
- [x] The title for your pull request (PR) should follow our title convention. [Learn more about the pull request title convention used in this repository](https://github.com/kubeflow/pipelines/blob/master/CONTRIBUTING.md#pull-request-title-convention).

PR #10602, opened 2024-03-21
feat(kubernetes_platform): Update kubernetes_platform go package to include generic ephemerl volume

abaland:
Include Generic Ephemeral Volume resources to kubernetes platform proto.

This add support for generic ephemeral volume (PVC that are tied to the lifecycle of the pod), to be used in KFP SDF V2. It is currently usable in KFP SDK V1 but the functionality disappeared with v2.

Reference for k8s resource: https://kubernetes.io/docs/concepts/storage/ephemeral-volumes/#generic-ephemeral-volumes

Reference for adding this is KFP SDK v1 (it came out of the box with the kubernetes python library upgrade): https://github.com/kubeflow/pipelines/pull/8526

Why ephemeral volume vs creating a PVC via the provided utilities: mainly 2 reasons:
1. At least with AWS EKS, creating PVC for the pipeline and pass it between components causes issues w.r.t to availability zone. When created, PVC do not have an AZ attached, but as soon as the underlying storage is created, the AZ is assigned and it cannot be changed. This means that all future steps that will re-use this PVC MUST be on the same AZ, which has multiple annoyances coming wth it: (a) the default cluster autoscaler can't select AZ when it scales up an autoscaling group that uses multiple AZ -> need to have separate autoscaling groups for each AZ. (b) if a specific instance type has availability issues in a given AZ, we're stuck.
2. An alternative to the above would be to create PVC per component instead of at the pipeline level. This is also annoying for a few reasons: (a) our pipeline has 10 steps, the 10 creations and 10 deletions triples that and make the graph difficult to rid. This ephemeral volume is done under the hood by K8s. (b) if components die halfway, the PVC gets automatically cleaned up with its attached pod -> no need to execute the cleanup

I am currently drafting the PR for the backend and SDK as well (hopefully ready this weekend), but from what I understood, this MR to update the proto first is required to allow the test suite to pass

**Checklist:**


kubernetes_platform/proto/kubernetes_executor_config.proto
```diff
@@ -156,6 +157,23 @@ message ConfigMapAsEnv {
...
+    // Corresponds to ephemeral.volumeClaimTemplate.spec.accessModes field.
+    repeated string access_modes = 3;
+    // Corresponds to ephemeral.volumeClaimTemplate.spec.resources.requests.storage field.
+    string size = 4;
+    // If true, corresponds to omitted ephemeral.volumeClaimTemplate.spec.storageClassName.
+    bool default_storage_class = 5;
+    // Corresponds to ephemeral.volumeClaimTemplate.spec.storageClassName string field.
+    // Should only be used if default_storage_class is false.
+    string storage_class_name = 6;
+}
```
Tomcli, line 175: According to the documentations, ephemeral volume also support metadata and resources (requests, limits, claims). Would you mind adding those spec as well? Thank you.

You can reuse the proto message `PodMetadata` for metadata.

abaland: Wow, that was a super fast review

For resources: I don't think that they support resources other that "requests.storage", which is what the "size" is there (it's a tiny abstraction of the k8s terminology, but i don't find it that confusing).

For metadata: sure, will add that asap.

Tomcli (commented):
/lgtm
/assign @chensun

abaland:
Sorry for the above large change, i pushed to the wrong branch when filing a draft PR....

Tomcli (approved):
/lgtm
Only the non required tests are failing, so we are good to go.

PR #10637, opened 2024-03-28
fix(samples): Fix the loop_output.py example to handle the new parallel loop type requirement

Tomcli:
After PR #10494, all the parallel loop input type has to be `List`. The current loop_output.py example still output as `str` which causes the following errors:
```shell
Cannot iterate over a single parameter using `dsl.ParallelFor`. Expected a list of parameters as argument to `items`.
```

Thus, fix the loop_output.py example to output type `List[str]`.

**Checklist:**

### DIFF
--- a/samples/core/loop_output/loop_output.py
+++ b/samples/core/loop_output/loop_output.py
@@ -13,11 +13,12 @@
 # limitations under the License.
 
 from kfp import compiler, dsl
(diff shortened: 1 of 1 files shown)


connor-mccarthy:
/lgtm
/approve

Thank you, @Tomcli!

PR #10631, opened 2024-03-27
fix(backend): Revert API Generator update and regenerate backend API Proto and Swagger files

gmfrasca:
**Description of your changes:**
Reverts Backend API-Generator updates for v2.1.0 release, which was causing API mismatches and CI/unit test compilation errors 

Fixes #10630 

**Checklist:**
- [x] The title for your pull request (PR) should follow our title convention. [Learn more about the pull request title convention used in this repository](https://github.com/kubeflow/pipelines/blob/master/CONTRIBUTING.md#pull-request-title-convention).

PR #10640, opened 2024-03-28
fix(backend): Update backend common code and integration tests with updated API Service Params

### LINKED ISSUE #10629: [bug] Integration test failed after the 2.1.0 release commit
### Environment

Running unit tests on 2.1.0 branch
*  KFP version:
2.1.0
*  KFP SDK version:
2.7.0

### Steps to reproduce
After the 2.1.0 release, all the API integration tests are failing which causes all the current backend PRs to fail too.
e.g. `FAIL github.com/kubeflow/pipelines/backend/src/common/client/api_server/v1`

Before the [2.1.0 release commit](https://github.com/kubeflow/pipelines/commit/f328f0b588c35cdf1e5b31638fbe3596b2f38413), the backend API didn't change and the integration test works fine.

### Labels

---

gmfrasca:
Fixes #10629 , Supercedes #10631 

The 2.1.0 release updated the API Generator, which in turn updated the signature/names of most of the *Params APIs.  This PR refactors all instances to use the newly generated naming schema

**Checklist:**


Tomcli (commented):
/ok-to-test

Tomcli:
Thanks @gmfrasca, seems like the integration test is passed but the upgrade test is failing. Feel free to let us know if you are blocked by anything in the ci/cd.

gmfrasca:
Update:

So both tests are failing due to an issue with unexpected responses (see example error printout[1]).  I tried tracing the error flow and while doing so noticed that the available schemes were reduced from http AND https to simply http (see example [change in DefaultSchemes](https://github.com/kubeflow/pipelines/commit/f328f0b588c35cdf1e5b31638fbe3596b2f38413#diff-c0210c926b228066f87ac8fe5379104c5c05bbf70bfe0c8f3e7a51730937e105R30) for various `*_client.go`).  These were generated via protobuf so I'm not entirely sure _where_ that is specified but my suspicion is that this may be the root cause of the integration test failures (since they are using live connections/traffic while backend unit tests are not)

[1] errortext:
```
Failed to list experiments. Raw error from the service: Issue calling the service. Use the '--debug' flag to see the HTTP request/response. Raw error from the client: invalid character 'C' looking for beginning of value
```

gmfrasca:
/hold

Tested manually updating the DefaultScheme and Scheme values as I suspected in previous message.  Local testing is promising, so pushing a WIP commit to see how CI fares.

If it passes we need to figure out how to make sure API Generation includes the option

gmfrasca:
alright, looks like all CI is passing as hoped, with just a very minor expected syntax update made. :) 

Given that, I did some more digging and found a change in grpc-gateway[1] that makes the schemes spec default to emptyset (`[]`) rather than `["http", "https"]`, so perhaps that's the root cause be it grpc-gateway or go-swagger making accommodations because of that change (although with the versions pinned I'm not sure how the behavior managed to change here).  So with that said, updating the base `.proto`s to include `schemes: ["http", "https"]` and running the generator appears to correctly generate the rest of the files.

The following commit will have these changes made, but my testing infra has hibernated for the night so this is a shot in the dark that CI remains passing. Assuming it does, I believe this PR is ready for final review and can be unblocked.


[1] https://github.com/grpc-ecosystem/grpc-gateway/commit/7dddcb1aacd290920dcd078eca96d90b9c1b2355

rimolive:
@Tomcli @chensun Can you plesae review?

Tomcli (commented):
/lgtm

chensun (approved):
/lgtm
/approve

Thank you, @gmfrasca!

Issue #10489, opened 2024-02-15
[feature] Add the integration tests (github actions) and corresponding deployment manifests for Tekton runtime


Tomcli:
### Feature Area
Add the integration tests (github actions) and corresponding deployment manifests for Tekton runtime

### Is there a workaround currently?

---

Love this idea? Give it a 👍.

rimolive:
Closing this issue as #10702 is merged

/close

PR #10702, opened 2024-04-16
chore: Add kfp-tekton integration tests and manifests

rimolive:
Solves #10489

Adds the kfp-tekton integration tests and corresponding manifests

**Checklist:**


rimolive:
/hold

Wait for tests to pass before ask for review/approval

Tomcli (changes requested):
few comments on the registry name.

manifests/kustomize/third-party/tekton-custom-task/driver-controller/200-serviceaccount.yaml
```diff
@@ -0,0 +1,9 @@
+apiVersion: v1
```
Tomcli, line 1: FYI everything under the `/manifests/kustomize/third-party/tekton-custom-task/driver-controller` was for the old v2 implementation. We moved all the driver logic to the Tekton custom kfptask to handle driver and task creation logic in the same Tekton controller.

manifests/kustomize/env/platform-agnostic-tekton-multi-user/kustomization.yaml
```diff
@@ -0,0 +1,81 @@
...
+
+# Identifier for application manager to apply ownerReference.
+# The ownerReference ensures the resources get garbage collected
+# when application is deleted.
+commonLabels:
+  application-crd-id: kubeflow-pipelines
+
+images:
+- name: gcr.io/ml-pipeline/api-server
+  newName: quay.io/aipipeline/apiserver
```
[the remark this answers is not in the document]
rimolive: Only this image or all of them?

scripts/deploy/github/re-tag.sh
```diff
@@ -0,0 +1,37 @@
...
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+# source: https://raw.githubusercontent.com/open-toolchain/commons/master/scripts/check_registry.sh
+
+# Remove the x if you need no print out of each command
+set -e
+
+REGISTRY1="${REGISTRY1:-docker.io/aipipeline}"
+REGISTRY2="${REGISTRY2:-quay.io/aipipeline}"
```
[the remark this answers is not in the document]
rimolive: Where this script is used? Maybe it's safe to remove?

manifests/kustomize/env/platform-agnostic-tekton-multi-user/kustomization.yaml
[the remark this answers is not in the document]
Tomcli: All the images under `manifests/kustomize/env/platform-agnostic-tekton-multi-user/kustomization.yaml` and ``manifests/kustomize/env/platform-agnostic-tekton/kustomization.yaml`

scripts/deploy/github/re-tag.sh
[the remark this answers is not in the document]
Tomcli: This script is to retag the local built docker.io images to the gcr.io images. So the local test can use the same manifests but only use the images in the local registry.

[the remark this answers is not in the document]
Tomcli: The renamed images in this script will not push to external, so we don't need to worry about push credentials etc.

.github/workflows/backend.yml
```diff
@@ -0,0 +1,83 @@
...
+  backend-integration:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout code
+        uses: actions/checkout@v4
+      - name: Create k8s Kind Cluster
+        uses: container-tools/kind-action@v2
+        with:
+          cluster_name: kfp-tekton
+          kubectl_version: v1.26.3
```
Tomcli, line 40: I found the deployment is missing some manifests file. It probably due to the recent update to the new kustomize 5.3 version. To fix that, we need to upgrade the kubectl version to 1.32+ or convert all the kubectl -k commands to `kustomize`

rimolive: I changed to 1.29.4 because of the k8s 1.29 support we're planning for Kubeflow 1.9. I'll check the kubectl -k change to kustomize.

rimolive:
Test are failling because some resources don't have namespace explicitly set. Some examples:
* [MySQL](https://github.com/kubeflow/pipelines/blob/master/manifests/kustomize/third-party/mysql/base/mysql-deployment.yaml#L4)
* [ml-metadata Envoy Proxy](https://github.com/kubeflow/pipelines/blob/master/manifests/kustomize/base/metadata/base/metadata-envoy-service.yaml#L6)
* [ml-metadata gRPC Server](https://github.com/kubeflow/pipelines/blob/master/manifests/kustomize/base/metadata/base/metadata-grpc-deployment.yaml#L4)

I tried to find a way to set in the kustomization.yaml level to set namespace in the objects that don't have namespace set but no luck. That being said, I'd like to ask @Tomcli and @chensun :
* Is it ok if I set namespace in these Deployments?
* If not, do you suggest another way to make them deploy in kubeflow namespace without explicitly setting in these objects?

Tomcli (commented):
few comments on fixing the old kustomize code to kustomize 5.2.1

Tomcli (commented):
Also, I think you didn't copy some role permission from the persistent agent, so the end to end test wasn't able to update the run status.

https://github.com/kubeflow/kfp-tekton/blob/v2-integration/manifests/kustomize/base/pipeline/ml-pipeline-persistenceagent-role.yaml#L29-L45

manifests/kustomize/third-party/tekton/upstream/manifests/base/tektoncd-install/tekton-release.yaml
```diff
@@ -0,0 +1,3463 @@
...
+  # Setting this flag to "true" enables populating the "provenance" field in TaskRun
+  # and PipelineRun status. This field contains metadata about resources used
+  # in the TaskRun/PipelineRun such as the source from where a remote Task/Pipeline
+  # definition was fetched.
+  enable-provenance-in-status: "true"
+  # Setting this flag will determine how Tekton pipelines will handle non-falsifiable provenance.
+  # If set to "spire", then SPIRE will be used to ensure non-falsifiable provenance.
+  # If set to "none", then Tekton will not have non-falsifiable provenance.
+  # This is an experimental feature and thus should still be considered an alpha feature.
+  enforce-nonfalsifiability: "spire"
```
rimolive, line 1704: @Tomcli One quick thing: I had to change the value of that feature flag because otherwise kfptask and exit-handler webhooks will keep failling. If you have more details about that flag and if it's ok to keep it set that way, I really appreciate.

rimolive:
> Also, I think you didn't copy some role permission from the persistent agent, so the end to end test wasn't able to update the run status.
> 
> https://github.com/kubeflow/kfp-tekton/blob/v2-integration/manifests/kustomize/base/pipeline/ml-pipeline-persistenceagent-role.yaml#L29-L45

Thank you for that! I'm still trying to figure you why the flip coin pipeline never finishes. At least with that change I can see the pipeline status of RUNNING, but it stays in that status until failure.

manifests/kustomize/third-party/tekton/upstream/manifests/base/tektoncd-install/tekton-release.yaml
Tomcli: spire is an extra server for signing Tekton results and artifacts. It's not really a feature we need for KFP and enable it will reduce the Tekton performance.

Do you have the error message for the kfptask and exit-handler webhooks? We probably just need to update the config for these two webhooks to support k8s 1.29.

.github/workflows/backend.yml
```diff
@@ -0,0 +1,88 @@
...
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout code
+        uses: actions/checkout@v4
+      - name: Create k8s Kind Cluster
+        uses: container-tools/kind-action@v2
+        with:
+          cluster_name: kfp-tekton
+          kubectl_version: v1.29.2
+          version: v0.17.0
```
Tomcli, line 41: I'm not sure kind 1.17 has the bug fixes for the k8s 1.29 cluster deployment. Maybe we either upgrade kind to 0.22 which has k8s 1.29 support or just run `node_image: kindest/node:v1.26.3` to minimize the deployment variable changes.
```suggestion
          version: v0.22.0
```

manifests/kustomize/third-party/tekton/upstream/manifests/base/tektoncd-install/tekton-release.yaml
rimolive: Nevermind, I reverted that change and deployment looks good.

rimolive:
Just a checkpoint comment: Tests are passing, but I still need to address Tommy's comment on Kustomize 5.2.1 support and the re-tag script.

rimolive:
/unhold
cc @Tomcli @chensun

DharmitD (commented):
Skimmed through the integration test results in CI and verified they work as expected.

/lgtm

chensun (approved):

rimolive:
@chensun Looks like e2e test did not finish well, but unrelated to the PR. Should we try run again?

PR #10790, opened 2024-05-06
feat(backend): Remove PipelineSpec Template storage from ObjStore responsibilies. Fixes #10509

### LINKED ISSUE #10509: [feature] Store Pipeline IR in database, not object storage

Currently the Object Store in KFP is largely used for artifacts, except for one outlier, which is the [Pipeline IR](https://github.com/kubeflow/pipelines/blob/2.0.5/backend/src/apiserver/resource/resource_manager.go#L408-L409). 

I agree with the inline comments that this should be stored in the DB just like everything else that's not an artifact. 

Moving this to be stored in db, removes api server's dependency on the object store, and will make it fore future solutions for different artifact store implementations, without having to worry about api server. 

No

### Anything else? 

There's also archive logging, but this seems delegated to the backend engine (currently argo, but soon tekton as well), I'm not sure what to do about this one.

---

Love this idea? Give it a 👍.

gmfrasca:
Fixes #10509 

The PipelineSpec definition is currently stored in two places, DB and ObjStore, creating the potential for a competing source-of-truth problem.  The ObjStore copy doesn't appear to actually be used/retrieved anywhere, so remove it from the list of responsibilities

**Checklist:**

### DIFF
--- a/backend/src/apiserver/resource/resource_manager_test.go
+++ b/backend/src/apiserver/resource/resource_manager_test.go
@@ -1489,6 +1489,18 @@ func TestDeletePipelineVersion_FileError(t *testing.T) {
 	assert.True(t, ok)
 	pipelineStore.SetUUIDGenerator(util.NewFakeUUIDGeneratorOrFatal(FakeUUIDOne, nil))
 	manager.CreatePipelineVersion(pv)
+
+	// Switch to a bad object store
+	manager.objectStore = &FakeBadObjectStore{}
+
+	// Delete the above pipeline_version.
+	err = manager.DeletePipelineVersion(FakeUUIDOne)
+	assert.Nil(t, err)
+
+	// Verify the version in deleting status.
+	version, err := manager.pipelineStore.GetPipelineVersionWithStatus(FakeUUIDOne, model.PipelineVersionDeleting)
+	assert.NotNil(t, err)
+	assert.Nil(t, version)
 }
 
 // Tests DeletePipeline
--- a/backend/src/apiserver/resource/resource_manager_test.go
+++ b/backend/src/apiserver/resource/resource_manager_test.go
@@ -389,24 +389,6 @@ func TestCreatePipeline(t *testing.T) {
 			model:    createPipeline("complex", "", "user1"),
 		},
 		{
-			msg:            "BadObjectStore",
-			badObjectStore: true,
-			template:       testWorkflow.ToStringForStore(),
-			errorCode:      codes.Internal,
-			errorMsg:       "bad object store",
-			model:          createPipeline("BadOS", "", "user1"),
-			// We previously verified that the failed pipeline version
-			// in DB is in status PipelineVersionCreating by faking
-			// the UUID generator, so that we know the created version
-			// UUID in advance.
-			// We cannot verify it using public APIs,
-			// because the API does not expose them unless we know its UUID, but we
-			// cannot know its UUID when create version request failed.
-			// TODO: do we really need to verify this status? or should
-			// the create version request return a UUID when the
-			// pipeline version fails in PipelineVersionCreating state.
-		},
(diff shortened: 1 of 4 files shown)


backend/src/apiserver/resource/resource_manager_test.go
```diff
@@ -1524,18 +1489,6 @@ func TestDeletePipelineVersion_FileError(t *testing.T) {
...
-	manager.objectStore = &FakeBadObjectStore{}
-
-	// Delete the above pipeline_version.
-	err = manager.DeletePipelineVersion(FakeUUIDOne)
-	assert.NotNil(t, err)
-
-	// Verify the version in deleting status.
-	version, err := manager.pipelineStore.GetPipelineVersionWithStatus(FakeUUIDOne, model.PipelineVersionDeleting)
-	assert.Nil(t, err)
-	assert.NotNil(t, version)
```
gregsheremeta, line 1538: it didn't jump out at me why this entire block is being deleted. Is there nothing here test-worthy if the object store storage goes away? Specifically wondering if it makes sense to keep the check of the deleting status.

backend/src/apiserver/server/fakes_test.go
```diff
@@ -271,8 +271,9 @@ func initWithExperimentsAndTwoPipelineVersions(t *testing.T) *resource.FakeClien
 	resourceManager = resource.NewResourceManager(clientManager, &resource.ResourceManagerOptions{CollectMetrics: false})
 	_, err = resourceManager.CreatePipelineVersion(
 		&model.PipelineVersion{
-			Name:       "pipeline_version",
-			PipelineId: DefaultFakeUUID,
+			Name:         "pipeline_version",
+			PipelineId:   DefaultFakeUUID,
+			PipelineSpec: "apiVersion: argoproj.io/v1alpha1\nkind: Workflow",
```
gregsheremeta, line 276: how is this related to this PR?

gmfrasca: If a mock PipelineSpec isn't provided, the Mock ResourceManager will attempt to fetch the spec from ObjectStore which will break as it no longer is uploaded there.  This behavior only exists in the test RM, I believe

backend/src/apiserver/resource/resource_manager_test.go
gmfrasca: yeah, this specific test can probably remain, albeit the expected results are now the exact opposite: the RM should *not* throw an error if the ObjStore is bad, *should* be able to delete the pipelineversion, and the pipelineversion should *not* exist after performing the delete operation (essentially flip the asserts in L1533, L1537, and L1538)

gregsheremeta:
what's the status of this? Just need reviews?

gmfrasca:
> what's the status of this? Just need reviews?

yes, believe so

hbelmiro (commented):
/lgtm

backend/src/apiserver/server/fakes_test.go
HumairAK: yep: https://github.com/kubeflow/pipelines/blob/fbb8b399b32e49f0683d2adf85d89dcbc30e3a31/backend/src/apiserver/storage/pipeline_store.go#L433 this is now fetched from the PipelineSpec column in the DB.

HumairAK:
Confirmed this with pre-existing pipelines, they continue to be fetched successfully. New pipelines as well, and they are not stored in the object store. Great work! 

/lgtm
/approve

HumairAK:
@gmfrasca this now has decoupled api server from the object store, can we update the docs here: 

https://github.com/kubeflow/website/blob/master/content/en/docs/components/pipelines/operator-guides/configure-object-store.md

I think we can basically remove the entire first section


fyi @chensun

PR #10826, opened 2024-05-16
fix(backend): fixes upload pipeline version function. Fixes #10825

### LINKED ISSUE #10825: [sdk] Create version command requires both pipeline_id and pipeline_name
### Environment

*  KFP version:
2.7.0
*  KFP SDK version: 
2.0.5
*  All dependencies version:
kfp                      2.7.0
kfp-kubernetes           1.2.0
kfp-pipeline-spec        0.3.0
kfp-server-api           2.0.5

### Steps to reproduce

```
kfp pipeline create-version -n some-pipeline pipeline.yaml
Usage: kfp pipeline create-version [OPTIONS] PACKAGE_FILE
Try 'kfp pipeline create-version --help' for help.

Error: Missing option '-v' / '--pipeline-version'.
```

### Expected result

The correct behaviour should be allowing creation of the pipeline by providing `pipeline-version` **OR** `pipeline-name` option.

### Materials and Reference

`create-version` function utilize `upload_pipeline_version` under the hood which has potential bug:
[here](https://github.com/kubeflow/pipelines/blob/master/sdk/python/kfp/client/client.py#L1447)

Potential fix:
```python
        if  not any([pipeline_id, pipeline_name]):
            raise ValueError('Either pipeline_id or pipeline_name is required.')
```

rchojn:
Introduce fix to `upload_pipeline_version` function - removed unnecessary `all` condition.

**Checklist:**


rchojn:
@hbelmiro is there a way to see a detailed log of this error?

sdk/python/kfp/client/client.py
```diff
@@ -1444,8 +1444,7 @@ def upload_pipeline_version(
             ``V2beta1PipelineVersion`` object.
         """
 
-        if all([pipeline_id, pipeline_name
```
chensun, line 1447: The all checks ensures that `pipeline_id` and `pipeline_name` aren't provided at the same time--either you upload a brand new pipeline so you need to provide a name or you upload to an existing pipeline as a new version so you need to specify the id.

chensun (changes requested):

sdk/python/kfp/client/client.py
rchojn: @chensun, thanks for the clarification. Here's how the pipeline lifecycle works from my perspective:

1. I upload the pipeline with a specific name using the command: 
   ```
   kfp pipeline create -p pipeline123 package.yaml
   ```
2. After releasing new code, I want to upload a new version of the pipeline. To do this, I follow these steps:
   
   a) I attempt to create a new version using:
   ```
   kfp pipeline create-version -n pipeline123 -v 2024-05-23 package.yaml
   ```
   However, this shows the error: `Either pipeline_id or pipeline_name is required.` This is misleading because the `pipeline_name` was provided. Why is the `pipeline_name` option not available here? Isn't this a potential issue or at least something that needs clarification?

   b) I then use the command:
   ```
   kfp pipeline create-version -p 20540d8f-03f0-4ec2-97d1-3b79d341d713 -v 2024-05-23 package.yaml
   ```
   This works as expected. So, it appears I need to store the `pipeline_id` or obtain it dynamically before deploying the new pipeline version. It seems I can't rely solely on the `pipeline_name`.

Could you please provide some clarification on why the `pipeline_name` option isn't supported for creating a new version and if there's a recommended way to handle this?

Thank you.

---

rchojn: @chensun so I believe, there is no need to fetch additionally `pipeline_id` in `create_version` command. Because of this mechanism, we get both `pipeline_id` and `pipeline_name` which then are passed to `upload_pipeline_version` and it ended up with mentioned error

rchojn: @connor-mccarthy could you please have a look?

Issue #10851, opened 2024-05-27
[test] periodic functional tests fail due to absence of GCP credentials


tmvfb:
Not a backend but rather a testing issue, I'm not able to select any suitable labels.

Periodic functional tests currently fail as connection to GCP test cluster is not possible (https://github.com/kubeflow/pipelines/actions/runs/9247286137/job/25435904107).

I already started fixing the tests, but now I need some community input. Issues to be addressed:

1. GCP test cluster access from GitHub actions. Is there a secret in this repository that provides GCP credentials for the test cluster  that can be used with GitHub actions (possibly [GOOGLE_APPLICATION_CREDENTIALS](https://github.com/prokube-ai/pipelines/tree/master/test/kfp-functional-test))? GitHub CI is trying to access test cluster that is configured here: https://github.com/kubeflow/testing/tree/master/test-infra/kfp.
2. Prow tests [failing](https://testgrid.k8s.io/googleoss-kubeflow-pipelines#kubeflow-pipelines-periodic-functional-test). Python3.7-slim image used to run prow test ([defined here](https://github.com/prokube-ai/pipelines/blob/master/test/kfp-functional-test/kfp-functional-test.sh)) is debian-based and doesn't have `sudo`. GitHub actions need `sudo` as they run on ubuntu. Should be easily fixable by changing test code to something like this:
```bash
if command -v sudo >/dev/null 2>&1; then
    sudo apt-get update -y
    sudo apt --no-install-recommends -y -q install curl
else
    apt-get update -y
    apt --no-install-recommends -y -q install curl
fi
```

---

Impacted by this bug? Give it a 👍.

rimolive:
The periodic funcional tests running in GitHub actions should be using the KinD cluster created by the workflow. GCP is not an option here because this is specific environment and we want to run these tests in a vendor-neutral environment.

That said, ensure that your fixes target the KinD cluster created in the workflow.

tmvfb:
@rimolive thank you, now that makes sense to me! 

From my understanding, running the functional tests against bare kind cluster is not of any use, since we need to actually install kfp on this cluster. I tried to address this in a new PR, tested on a fork, seems to work now.

PR #10859, opened 2024-06-02
test: install kfp on github periodic functional tests workflow

tmvfb:
Hopefully last fix for periodic functional tests. Current problems (https://github.com/kubeflow/pipelines/issues/10851):
1. Tests fail in GH workflows as they try to connect to the same GCP cluster that prow tests run on. The corresponding environment is not configured in GitHub.
2. Tests fail in prow as they try to run `sudo` which is not available on python3.7-slim (runs on debian).

Solutions:
1. For GH actions, use the kind cluster defined in the workflow file to run the test instead of GCP test cluster. This requires actually installing kfp on this kind cluster. I copypasted installation from the [backend test workflow file](https://github.com/kubeflow/pipelines/blob/master/.github/workflows/backend.yml). Let me know if there is a simpler/faster solution (test runs for 15 mins), [standalone kfp deployment](https://www.kubeflow.org/docs/components/pipelines/v1/installation/standalone-deployment/) didn't work for me.
3. Make tests aware of the environment they are running on. If running in GitHub actions (we can check if `KIND_REGISTRY` env var created by kind action is available), tests work without indicating [host](https://github.com/kubeflow/pipelines/blob/master/sdk/python/kfp/client/client.py#L105). For prow tests, run `apt` without `sudo`, as it was [initially](https://github.com/kubeflow/pipelines/blob/7b4ddf658c77bc5d4663ca100bdd2392d2c4f109/test/kfp-functional-test/kfp-functional-test.sh) defined.

**Checklist:**

### DIFF
--- a/.github/workflows/periodic.yml
+++ b/.github/workflows/periodic.yml
@@ -10,13 +10,8 @@ jobs:
     steps:
       - name: Checkout repository
         uses: actions/checkout@v4
-      - name: Create k8s Kind Cluster
-        uses: container-tools/kind-action@v2
-        with:
-          cluster_name: kfp-tekton
-          kubectl_version: v1.29.2
-          version: v0.22.0
-          node_image: kindest/node:v1.29.2
+      - name: Create KFP cluster
+        uses: ./.github/actions/kfp-cluster
       - name: Run Functional Tests
         run: |
           log_dir=$(mktemp -d)
--- a/.github/workflows/periodic.yml
+++ b/.github/workflows/periodic.yml
@@ -12,6 +12,9 @@ jobs:
         uses: actions/checkout@v4
       - name: Create KFP cluster
         uses: ./.github/actions/kfp-cluster
+      - name: Port forward kfp apiserver
+        run: |
+          nohup kubectl port-forward --namespace kubeflow svc/ml-pipeline 8888:8888 &
       - name: Run Functional Tests
         run: |
           log_dir=$(mktemp -d)
--- a/test/kfp-functional-test/kfp-functional-test.sh
+++ b/test/kfp-functional-test/kfp-functional-test.sh
@@ -15,4 +15,4 @@
 
 source_root="$(pwd)"
 python3 -m pip install -r "${source_root}/test/kfp-functional-test/requirements.txt"
-python3 "${source_root}/test/kfp-functional-test/run_kfp_functional_test.py" --host ""
+python3 "${source_root}/test/kfp-functional-test/run_kfp_functional_test.py" --host "http://localhost:8888"  # host configured in workflow file
--- a/test/kfp-functional-test/kfp-functional-test.sh
+++ b/test/kfp-functional-test/kfp-functional-test.sh
@@ -12,11 +12,7 @@
 # WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 # See the License for the specific language governing permissions and
 # limitations under the License.
-sudo apt-get update -y
-sudo apt --no-install-recommends -y -q install curl
+
 source_root="$(pwd)"
-
 python3 -m pip install -r "${source_root}/test/kfp-functional-test/requirements.txt"
-HOST="https://$(curl https://raw.githubusercontent.com/kubeflow/testing/master/test-infra/kfp/endpoint)"
-
-python3 "${source_root}/test/kfp-functional-test/run_kfp_functional_test.py" --host "${HOST}"
+python3 "${source_root}/test/kfp-functional-test/run_kfp_functional_test.py" --host ""


test/kfp-functional-test/kfp-functional-test.sh
```diff
@@ -12,11 +12,16 @@
...
-sudo apt-get update -y
-sudo apt --no-install-recommends -y -q install curl
-source_root="$(pwd)"
 
+source_root="$(pwd)"
 python3 -m pip install -r "${source_root}/test/kfp-functional-test/requirements.txt"
-HOST="https://$(curl https://raw.githubusercontent.com/kubeflow/testing/master/test-infra/kfp/endpoint)"
 
-python3 "${source_root}/test/kfp-functional-test/run_kfp_functional_test.py" --host "${HOST}"
+if [ -n "$KIND_REGISTRY" ]; then
```
rimolive, line 19: When this variable is specified when GH Action is running?

.github/workflows/periodic.yml
```diff
@@ -17,6 +17,10 @@ jobs:
           kubectl_version: v1.29.2
           version: v0.22.0
           node_image: kindest/node:v1.29.2
+      - name: Build images
+        run: ./scripts/deploy/github/build-images.sh
+      - name: Deploy kfp-tekton
+        run: ./scripts/deploy/github/deploy-kfp.sh
```
rimolive, line 23: These new lines are supposed to be only used in kfp-tekton backend tests. Periodic tests don't run any tests against kfp-tekton.

test/kfp-functional-test/kfp-functional-test.sh
```diff
@@ -12,11 +12,16 @@
...
+    # if KIND_REGISTRY is available, we're running GH actions, host is not required
+    python3 "${source_root}/test/kfp-functional-test/run_kfp_functional_test.py" --host ""
+else
```
rimolive, line 22: We expect to remove the Prow jobs to run these tests, I think this confitional is not required.

rimolive (commented):
Thank you for your contributions! I left a few comments, let's work on these items before lgtm/approve

test/kfp-functional-test/kfp-functional-test.sh
tmvfb: This variable gets created when the kind action runs, see [here](https://github.com/container-tools/kind-action). If we don't need prow tests, we can remove this conditional anyway.

.github/workflows/periodic.yml
tmvfb: This one I don't really fully get! We need to define which components of the kfp we actually need in the kind cluster to be able to run the tests. The prow tests run against the cluster that is specified [here](https://github.com/kubeflow/testing/tree/master/test-infra/kfp).

If I try to run GH actions tests against bare kind cluster or against [standalone KFP](https://www.kubeflow.org/docs/components/pipelines/v1/installation/standalone-deployment/), I get a missing endpoint error: 
image

As far as I understand, that means that Client class cannot connect to kfp server. 

A possible solution in this case is that I actually *need* to specify the host address. Could you give a possible hint what the kfp host address available in GH actions can be? I think I already tried:
http://ml-pipeline:8888/
http://ml-pipeline-ui:80

tmvfb: I guess we can run the tests against standalone kfp then? I realised that my tests didn't work since I didn't allow for any time for the kfp pods to come up in the test cluster.

tmvfb:
@rimolive, completely reworked the PR, now we deploy standalone kfp inside GitHub actions kind cluster instead of kfp-tekton. I also simplified shell script considering we won't need prow tests anymore.

rimolive (commented):
One last comment.

test/kfp-functional-test/kfp-functional-test.sh
```diff
@@ -12,11 +12,7 @@
...
-sudo apt-get update -y
-sudo apt --no-install-recommends -y -q install curl
-source_root="$(pwd)"
 
+source_root="$(pwd)"
 python3 -m pip install -r "${source_root}/test/kfp-functional-test/requirements.txt"
-HOST="https://$(curl https://raw.githubusercontent.com/kubeflow/testing/master/test-infra/kfp/endpoint)"
-
-python3 "${source_root}/test/kfp-functional-test/run_kfp_functional_test.py" --host "${HOST}"
+python3 "${source_root}/test/kfp-functional-test/run_kfp_functional_test.py" --host ""
```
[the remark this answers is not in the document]
tmvfb: Pipelines are not available on `http://localhost:8888` inside GitHub actions.

`Client` class is configured to automatically connect to local cluster if no host is specified ([see source code](https://github.com/kubeflow/pipelines/blob/master/sdk/python/kfp/client/client.py#L105)). I would opt for just removing the CLI argument from the code.

P.S. Pipelines should theoretically be available on `ml-pipeline.kubeflow.svc.cluster.local:8888`, but didn't work for me as well. I am not able to identify the host where I can connect to pipelines inside GH actions, or how to configure this correctly, so the easiest option I see is to leave the host empty.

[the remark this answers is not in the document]
tmvfb: @rimolive, this is surely an option, thank you!!

Could you please take a look at my current solution, however? It uses the fact that test `.py` file gets executed from a pod inside cluster. In this case in-cluster `ml-pipeline` service DNS name is used by default, it is reachable from that pod without specifying the *host* argument. This approach is simpler, but is also a little less explicit.

Please let me know if we should better go with port-forwarding instead, already tested that approach, and it also works. And thanks again for the feedback.

rimolive:
Don't forget to sign-off your commits

tmvfb:
> Don't forget to sign-off your commits

Done!

rimolive:
Thank you!

/lgttm

cc @DharmitD @chensun

DharmitD (commented):
changes /lgtm

@tmvfb could you provide a sample run on your fork with these changes?

tmvfb:
@DharmitD, https://github.com/prokube-ai/pipelines/actions/runs/9390604598

DharmitD (commented):
/lgtm

rimolive:
cc @chensun

rimolive:
@chensun Can you approve?

hbelmiro:
@tmvfb can you please rebase?

tmvfb:
> @tmvfb can you please rebase?

@hbelmiro, @DharmitD, done

.github/workflows/periodic.yml
```diff
@@ -13,10 +13,21 @@ jobs:
       - name: Create k8s Kind Cluster
         uses: container-tools/kind-action@v2
         with:
-          cluster_name: kfp-tekton
+          cluster_name: kfp-functional-test
           kubectl_version: v1.29.2
           version: v0.22.0
           node_image: kindest/node:v1.29.2
+      - name: Deploy kfp
```
hbelmiro, line 20: One more thing @tmvfb. Sorry.
Can you please use the new `kfp-cluster` action here? Like what is done in https://github.com/kubeflow/pipelines/blob/2fb4922fc0ecf78f7c8050111171333abdd2812a/.github/workflows/e2e-test.yaml#L26.

hbelmiro (changes requested):

.github/workflows/periodic.yml
tmvfb: @hbelmiro, no problem, edited commits and tested [here](https://github.com/prokube-ai/pipelines/actions/runs/9759012845/job/26934822612)

hbelmiro (commented):
/unhold
/lgtm

PR #10883, opened 2024-06-10
feat(components): Support dynamic machine type paramters in CustomTrainingJobOp

KevinGrantLee:

**Checklist:**


connor-mccarthy:
@KevinGrantLee, can you please address required presubmit checks `DCO` and `kubeflow-pipelines-sdk-yapf`?

sdk/python/kfp/compiler/compiler_utils.py
```diff
@@ -772,3 +772,29 @@ def get_dependencies(
...
+    """
+    if isinstance(data, dict):
+        return {
+            k: recursive_replace(v, old_value, new_value)
+            for k, v in data.items()
+        }
+    elif isinstance(data, list):
+        return [recursive_replace(i, old_value, new_value) for i in data]
+    else:
+        if isinstance(data, pipeline_channel.PipelineChannel):
```
chensun, line 798: This method seems explicitly replacing placeholder from one representation to another. It's not for replacing arbitrary value. So the method name should reflect it's purpose.

sdk/python/kfp/compiler/compiler_utils_test.py
```diff
@@ -53,6 +53,82 @@ def test_additional_input_name_for_pipeline_channel(self, channel,
...
+        },
+        {
+            'data': [{
+                0: 'a',
+                1: 'b',
+                2: 'c',
+                3: ['d']
+            }],
+            'old_value': 'd',
+            'new_value': 'dd',
```
chensun, line 118: The old and new values doesn't seem to be testing the real case scenario?

sdk/python/kfp/compiler/pipeline_spec_builder.py
```diff
@@ -239,70 +327,18 @@ def build_task_spec_for_task(
                         component_input_parameter)
 
         elif isinstance(input_value, str):
-            # Handle extra input due to string concat
```
chensun, line 242: IIRC, this chunk of code is only applicable for string typed inputs, why merging the code and expand it to other input types? Also it's a bit hard to read the diff between the deleted code and the extracted. Can you try make the changes in place without refactoring, and see if it's actually necessary to expand the logic to non-string typed inputs?

chensun (changes requested):

sdk/python/test_data/pipelines/pipeline_with_dynamic_custom_training_job.py
```diff
@@ -0,0 +1,59 @@
...
+    machine_type_task = machine_type()
+    accelerator_type_task = accelerator_type()
+    accelerator_count_task = accelerator_count()
+
+    custom_job.CustomTrainingJobOp(
+        display_name='add-numbers',
+        worker_pool_specs=[{
+            'container_spec': {
+                # doesn't need to be the container under test
+                # just need an image within the VPC-SC perimeter
```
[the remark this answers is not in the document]
KevinGrantLee: I copied this comment from another test pipeline, will remove.

sdk/python/test_data/pipelines/pipeline_with_dynamic_custom_training_job.py
```diff
@@ -0,0 +1,59 @@
...
+
+
+@dsl.component
+def accelerator_type() -> str:
+    return 'NVIDIA_TESLA_P4'
+
+
+@dsl.component
+def accelerator_count() -> int:
+    # This can either be int or int string
```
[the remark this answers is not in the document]
KevinGrantLee: Yes, I also tried compiling and submitting a pipeline where `accelerator_count` returned `'1'` and that pipeline succeeded.

sdk/python/kfp/compiler/compiler_utils_test.py
KevinGrantLee: That testcase was for some simply dummy values just to verify behavior of `recursive_replace()`, I can remove if you think its redundant?

sdk/python/test_data/pipelines/pipeline_with_dynamic_custom_training_job.py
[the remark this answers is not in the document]
KevinGrantLee: I'll remove the comment to avoid confusion because leaving return annotation as int and returning `'1'` causes errors

sdk/python/test_data/pipelines/pipeline_with_dynamic_custom_training_job.py
```diff
@@ -0,0 +1,59 @@
...
+                'image_uri':
+                    ('gcr.io/ml-pipeline/google-cloud-pipeline-components:2.5.0'
+                    ),
+                'command': ['echo'],
+                'args': ['foo'],
+            },
+            'machine_spec': {
+                'machine_type': machine_type_task.output,
+                'accelerator_type': accelerator_type_task.output,
+                'accelerator_count': accelerator_count_task.output,
```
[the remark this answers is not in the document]
KevinGrantLee: Done.

sdk/python/kfp/compiler/pipeline_spec_builder.py
KevinGrantLee: I found that this block of code could be reused for handling PipelineChannels inside of `worker_pool_specs` in addition to handling string typed inputs. Instead of copying the ~50 lines of code, I thought it'd be better to refactor the logic as a separate function `def replace_and_inject_placeholders()`. 

So I could un-refactor and duplicate the logic; I do have a slight preference for this refactoring but can go either way. wdyt?

sdk/python/kfp/compiler/compiler_utils.py
KevinGrantLee: Although here I'm just using this method for replacing placeholders, it can be used for arbitrary values as well so I left the method name general.

KevinGrantLee:
@chensun I verified that nested dags with dsl.Condition() and custom jobs with dynamic (task output and pipeline inputs) machine parameters compile and run successfully.

sdk/python/kfp/compiler/compiler_utils_test.py
KevinGrantLee: Removed testcase.

sdk/python/kfp/compiler/compiler_utils.py
KevinGrantLee: Renamed method to recursive_replace_placeholders

sdk/python/kfp/compiler/pipeline_spec_builder.py
chensun: This name of the extracted method isn't accurate--the code does more than placeholder manipulation but also component input expansion.
The branch logic now reads like this:
```python
if isinstance(input_value, str):
    # shared code
    pipeline_task_spec.inputs.parameters[
                input_name].runtime_value.constant.string_value = input_value
elif isinstance(input_value, (int, float, bool, dict, list)):
    if isinstance(input_value, (dict, list):
          # shared code
    pipeline_task_spec.inputs.parameters[
                input_name].runtime_value.constant.CopyFrom(
                    to_protobuf_value(input_value))
else:
     raise
```
You can achieve the same goal, and even more code reuse, without extracting a shared method by:
```python
if not isinstance(input_value, (str, dict, list, int, float, bool)):
    raise

if isinstance(input_value, (str, dict, list)):
    # shared code

pipeline_task_spec.inputs.parameters[
            input_name].runtime_value.constant.CopyFrom(
                to_protobuf_value(input_value))
```

sdk/python/kfp/compiler/compiler_utils.py
```diff
@@ -772,3 +772,32 @@ def get_dependencies(
...
     return dependencies
+
+
+def recursive_replace_placeholders(
+        data: Union[Dict, List], old_value: str,
+        new_value: str) -> Union[Dict, List]:
+    """Recursively replaces values in a nested dict/list object.
+
+    This method is used to replace PipelineChannel objects with pipeine channel
+    placeholders in a nested object like worker_pool_specs for custom jobs.
```
chensun, line 783: pipeline channel placeholders -> input parameter placeholder

sdk/python/kfp/compiler/pipeline_spec_builder.py
chensun: Aside from the refactoring part, I wonder what's the case for dict and list? In case `CustomTrainingJobOp` is used, what's the `input_value` here?

chensun:
> @chensun I verified that nested dags with dsl.Condition() and custom jobs with dynamic (task output and pipeline inputs) machine parameters compile and run successfully.

Can you add a test case?

sdk/python/kfp/compiler/pipeline_spec_builder.py
KevinGrantLee: If `CustomTrainingJobOp` is used, then we pass in `worker_pool_spec` into `input_value`.

`input_value = [{'container_spec': {'image_uri': 'gcr.io/ml-pipeline/google-cloud-pipeline-components:2.5.0', 'command': ['echo'], 'args': ['foo']}, 'machine_spec': {'machine_type': {{channel:task=machine-type;name=Output;type=String;}}, 'accelerator_type': {{channel:task=accelerator-type;name=Output;type=String;}}, 'accelerator_count': {{channel:task=accelerator-count;name=Output;type=Integer;}}}, 'replica_count': 1}]`

chensun: Thanks for the explanation. So `input_value` would be of type `list` in this case. Including `dict` in the same code path is just for future use cases not entirely necessary at this moment, right? I'm fine to include `dict` now. 

Have you seen my other comment on refactoring above? https://github.com/kubeflow/pipelines/pull/10883/files#r1636822260

sdk/python/kfp/compiler/pipeline_spec_builder.py
```diff
@@ -238,71 +238,77 @@ def build_task_spec_for_task(
...
-                    if channel.task_name in tasks_in_current_dag:
-                        # Dependent task within the same DAG.
-                        pipeline_task_spec.inputs.parameters[
-                            additional_input_name].task_output_parameter.producer_task = (
-                                utils.sanitize_task_name(channel.task_name))
-                        pipeline_task_spec.inputs.parameters[
-                            additional_input_name].task_output_parameter.output_parameter_key = (
-                                channel.name)
+        elif isinstance(input_value, (str, int, float, bool, dict, list)):
+            if isinstance(input_value, (str, dict, list)):
```
chensun, line 242: you can remove this `if isinstance(input_value, (str, dict, list)):` check, `extract_pipeline_channels_from_any` would return an empty list in case float, int, bool.

sdk/python/kfp/compiler/pipeline_spec_builder.py
```diff
@@ -238,71 +238,77 @@ def build_task_spec_for_task(
...
+                            raise RuntimeError(
+                                f'Name collision between existing input name {existing_input_name} and compiler injected input name {additional_input_name}'
+                            )
+
+                    additional_input_placeholder = placeholders.InputValuePlaceholder(
+                        additional_input_name)._to_string()
+
+                    if isinstance(input_value, str):
+                        input_value = input_value.replace(
+                            channel.pattern, additional_input_placeholder)
```
chensun, line 274: This would be covered by `recursive_replace_placeholders`, right?

chensun (changes requested):

sdk/python/kfp/compiler/pipeline_spec_builder.py
KevinGrantLee: Is it fine to remove the inner if check and expand the type annotations?

It would simplify the code but I'm not sure if it makes sense to update the type annotations for `extract_pipeline_channels_from_any` since ints, floats, and bools can't contain pipeline channels.
 https://github.com/kubeflow/pipelines/blob/92c317824a3697d955cb78cc9a85ecac58a3366c/sdk/python/kfp/dsl/pipeline_channel.py#L589

sdk/python/kfp/compiler/pipeline_spec_builder.py
KevinGrantLee: The string case is not covered by `recursive_replace_placeholders` as is, I would need to embed `string.replace()` logic in `recursive_replace_placeholders` if we wanted to get rid of this ifelse block.

I suppose question is if we want to expose this logic in `pipeline_spec_builder:build_task_spec_for_task` or `compiler_utils:recursive_repalce_placeholders`?

sdk/python/kfp/compiler/pipeline_spec_builder.py
chensun: Yes, you can update the payload annotation type.

sdk/python/kfp/compiler/pipeline_spec_builder.py
chensun: I see. It's up to you whether you want to keep it as-is or remove the ifelse block. I don't have a strong preference.

I'm not sure how this is related to your question on exposing the logic.

sdk/python/kfp/compiler/pipeline_spec_builder.py
KevinGrantLee: Done.

sdk/python/kfp/compiler/pipeline_spec_builder.py
KevinGrantLee: Alright, kept as it.

chensun (approved):
/lgtm
/approve

Thanks, @KevinGrantLee !

pthieu:
@KevinGrantLee is it possible to do the same for `worker_pool_specs.container_spec.env`? I have a use-case where I am trying to keep a consistent timestamp used for file naming on my custom training op and testing op afterwards (testing op needs the freshly-trained model), so I pass it in as an environment variable.

KevinGrantLee:
Hi @pthieu, this PR should also enable that usecase - I did some local tests, but can you confirm on your end?

pthieu:
@KevinGrantLee let me try the latest release

pthieu:
@KevinGrantLee think I'll need to wait for the next release as we use this in our CI/CD pipeline. Do you know what the release schedule is? I can do a test and confirm when it's out.

KevinGrantLee:
@pthieu, we're planning to do a release later this week. cc @chensun

alexredplanet:
I believe the same issue is occurring for the ModelBatchPredictOp @pthieu @KevinGrantLee , here is a minimal example: 

```
@dsl.component
def gcs_jsonl_uri(bucket_name: str, file_name: str) -> str:
    return f"gs://{bucket_name}/{file_name}"

@dsl.pipeline
def pipeline(project: str, location: str, model_name: str):

    model = ModelGetOp(project=project, model_name=model_name, location=location)
    gcs_jsonl = gcs_jsonl_uri(bucket_name="example", file_name="example.jsonl")

    batch_predict_job = ModelBatchPredictOp(
        model=model.outputs["model"],
        job_display_name="example",
        gcs_source_uris=[gcs_jsonl.output],
        location=location
    )
```

producing the same error as your issue @pthieu : 
`ValueError: Value must be one of the following types: str, int, float, bool, dict, and list. Got: "{{channel:task=gcs-jsonl-uri;name=Output;type=String;}}" of type "<class 'kfp.dsl.pipeline_channel.PipelineParameterChannel'>"`

KevinGrantLee:
Hi @alexredplanet , I'm reasonably confident that this pr should also fix your case. Once the next kfp release is done, can you retry?

alexredplanet:
Thanks @KevinGrantLee it did indeed fix that issue!

KevinGrantLee:
Hi @pthieu , KFP SDK 2.8.0, you should be able to test https://github.com/kubeflow/pipelines/issues/10902 again

pthieu:
@KevinGrantLee looks like getting the latest worked (at least no errors thrown)

Just need a dependency requirement change on the google pipelines package:
```
The conflict is caused by:
    The user requested kfp==2.8.0
    google-cloud-pipeline-components 2.14.1 depends on kfp<=2.7.0 and >=2.6.0
```

senyuyouren:
I am encountering a TypeError when trying to concatenate a PipelineParameterChannel object with a string in my pipeline component. This issue seems to be specific to kfp==2.8.0, as the same code worked without issues in previous versions.

Error Message:

TypeError: unsupported operand type(s) for +: 'PipelineParameterChannel' and 'str'

Code Snippet:
def minio_training_pipeline(imageurl: str = 'go'):
    import json
    from datetime import datetime

    # Read configuration file
    total_dir = "/mnt/data/AutoMeta"
    opt = {
        'model_name_prefix': imageurl
    }
    
    # Generate model name by concatenating prefix with current date time
    date_time = datetime.now().strftime("%Y%m%d%H%M%S")
    model_name = opt['model_name_prefix'] + date_time  # Error occurs here

senyuyouren:
> Hi @pthieu , KFP SDK 2.8.0 has been released: you should be able to test #10902 again

@KevinGrantLee 

I am encountering a TypeError when trying to concatenate a PipelineParameterChannel object with a string in my pipeline component. This issue seems to be specific to kfp==2.8.0, as the same code worked without issues in previous versions.

Error Message:

TypeError: unsupported operand type(s) for +: 'PipelineParameterChannel' and 'str'

Code Snippet:
def minio_training_pipeline(imageurl: str = 'go'):
    import json
    from datetime import datetime

    # Read configuration file
    total_dir = "/mnt/data/AutoMeta"
    opt = {
        'model_name_prefix': imageurl
    }
    
    # Generate model name by concatenating prefix with current date time
    date_time = datetime.now().strftime("%Y%m%d%H%M%S")
    model_name = opt['model_name_prefix'] + date_time  # Error occurs here

senyuyouren:
> Hi @pthieu , KFP SDK 2.8.0 has been released: you should be able to test #10902 again

@KevinGrantLee 
Version kfp==2.8.0
I seem to have encountered this problem as well. The values passed in as parameters cannot be directly called

Error Message:

TypeError: unsupported operand type(s) for +: 'PipelineParameterChannel' and 'str'

Code Snippet:
def minio_training_pipeline(imageurl: str = 'go'):
    import json
    from datetime import datetime

    # Read configuration file
    total_dir = "/mnt/data/AutoMeta"
    opt = {
        'model_name_prefix': imageurl
    }
    
    # Generate model name by concatenating prefix with current date time
    date_time = datetime.now().strftime("%Y%m%d%H%M%S")
    model_name = opt['model_name_prefix'] + date_time  # Error occurs here

KevinGrantLee:
Hi @senyuyouren , I think you need to cast the PipelineParameterChannel to str.

```
model_name = str(opt['model_name_prefix']) + date_time
```

senyuyouren:
> Hi @senyuyouren , I think you need to cast the PipelineParameterChannel to str.
> 
> ```
> model_name = str(opt['model_name_prefix']) + date_time
> ```

Hi @KevinGrantLee , In the pipeline, the value I passed through was modeled as str (modeled as)+date_time. In the first minio-training pipeline, the parameter value could be obtained normally. However, when the parameter value of modeled as cannot be obtained normally in the second minio-mode_name_test, do you know what the problem is

code: 
from kfp import compiler, dsl
from typing import Dict, Any


@dsl.component(base_image="mt-image-hub-registry.cn-beijing.cr.aliyuncs.com/public-image/python:3.9")
def minio_training_pipeline(imageurl: str, modelname: str, mode_name: str):
    print(f"Image URL: {imageurl}")
    print(f"Model Name: {modelname}")
    print(f"Mode_Name: {mode_name}")


@dsl.component(base_image="mt-image-hub-registry.cn-beijing.cr.aliyuncs.com/public-image/python:3.9")
def minio_mode_name_test(mode_name: str):
    print(f"Mode_Name: {mode_name}")


@dsl.pipeline(name='minio-training-pipeline')
def my_pipeline(imageurl: str, modelname: str, model_name: str = 'xiexieni'):
    import datetime
    date_time = datetime.datetime.now().strftime('%Y-%m-%d')
    model_name = str(model_name) + date_time
    print(model_name)

    minio_training_pipeline(imageurl=imageurl, modelname=modelname, mode_name=model_name).set_caching_options(enable_caching=False)
    minio_mode_name_test( mode_name=model_name).set_caching_options(enable_caching=False).after(minio_training_pipeline)



if __name__ == '__main__':
    pipeline_parameters = {
        'imageurl': 'images.zip',
        'modelname': 'MNXJ'
    }

    compiler.Compiler().compile(
        pipeline_func=my_pipeline,
        package_path='train_pipeline_test.yaml',
        pipeline_parameters=pipeline_parameters
    )
    print("Pipeline YAML has been generated.")


image
image

image
image

PR #10911, opened 2024-06-16
Upgrade go version to 1.21

DharmitD:
Resolves https://github.com/kubeflow/pipelines/issues/10912

**Checklist:**

### DIFF
--- a/.github/workflows/backend.yml
+++ b/.github/workflows/backend.yml
@@ -23,7 +23,7 @@ jobs:
     - name: Install Go
       uses: actions/setup-go@v4
       with:
-        go-version: 1.20.x
+        go-version: 1.21.x
     - name: Checkout code
       uses: actions/checkout@v4
     - name: "run go unit tests"
--- a/api/go.mod
+++ b/api/go.mod
@@ -1,6 +1,6 @@
 module github.com/kubeflow/pipelines/api
 
-go 1.16
+go 1.21
 
 require (
 	google.golang.org/genproto v0.0.0-20211026145609-4688e4c4e024
--- a/backend/api/Dockerfile
+++ b/backend/api/Dockerfile
@@ -13,7 +13,7 @@
 # limitations under the License.
 
 # Generate client code (go & json) from API protocol buffers
-FROM golang:1.20 as generator
+FROM golang:1.21 as generator
 ENV GRPC_GATEWAY_VERSION v1.9.6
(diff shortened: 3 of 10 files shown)


hbelmiro (commented):
/lgtm

DharmitD:
@chensun could you take a look?

gregsheremeta:
yeah seems good to me. I thought about bumping the kfp-kubernetes one myself :)

/lgtm

DharmitD:
@chensun this PR is reviewed and lgtm'd, could you please approve?

cc: @rimolive @HumairAK

DharmitD:
cc @chensun @zijianjoy @james-jwu

chensun (approved):
/lgtm

.github/workflows/backend.yml
```diff
@@ -23,7 +23,7 @@ jobs:
     - name: Install Go
       uses: actions/setup-go@v4
       with:
-        go-version: 1.20.x
+        go-version: 1.21.x
```
AndersBennedsgaard, line 26: It might make sense to use something like this instead:
```suggestion
        go-version-file: './go.mod'
```
This would reduce the number of places where we have to upgrade Go in the future. 
Obviously there are multiple `go.mod`s in this repo, so if there is one that makes more sense than the one in the root of the repository, it could be changed to that instead

AndersBennedsgaard (commented):
I just noticed this PR and got interested. No need to change anything if you don't think it's relevant :smile: 

On another note: should we consider using [Go workspaces](https://go.dev/ref/mod#workspaces) to reduce duplication between projects? That would mean we only have to change the Go version in the `go.work` file at the root of the repo

.github/workflows/backend.yml
HumairAK: Hey @AndersBennedsgaard 
I think these are very valid suggestions, but they are out of scope of this PR which is merely to upgrade the go version. My suggestion is to create an issue for this or propose a separate PR.

HumairAK:
@DharmitD  unfortunately tide is preventing merge due to: `kubeflow-pipeline-e2e-test` 

This might be because you need to rebase the new changes to the github workflows

DharmitD:
Rebased the PR. @HumairAK @gregsheremeta @hbelmiro could any of you lgtm again?

hbelmiro (commented):
/lgtm

Issue #10656, opened 2024-04-03
[feature]  Mounting emptyDir option on kfp v2 workflow-task using kfp-kubernetes


harshad16:
### Feature Area

/area sdk 

### What feature would you like to see?

In `kfp-kubernetes` SDK can we include `add_volume` function.
as in the kfp dsl, we had the option in kfp v1 https://github.com/kubeflow/pipelines/blob/f80bfd4e07565f0a95ccebcf5abd8e9a7ae2627f/sdk/python/kfp/deprecated/dsl/_container_op.py#L934
Use case is to have an option to volumes to the workflow tasks.

### What is the use case or pain point?

In (kfp v2) kfp-kubernetes only provides the `add_pvc` and `mount_pvc` options.
As currently , there is no option to mount ,  `emptyDir`
secretes and configmaps can be mounted separately. 

https://github.com/kubeflow/pipelines/blob/master/kubernetes_platform/python/kfp/kubernetes/volume.py

### Is there a workaround currently?

No. 
use older deprecated version.

---

Love this idea? Give it a 👍.

gregsheremeta:
not stale ... i'm working on it

thesuperzapper:
It looks like there is a PR to add this important feature in https://github.com/kubeflow/pipelines/pull/10913, but I am not sure about its status.

Can one of @gregsheremeta @hbelmiro @droctothorpe @rimolive confirm what we need to do to get that merged?

gregsheremeta:
> confirm what we need to do to get that merged?

We need one of the Google maintainers to review it. Chen is out until July 22. It doesn't appear that Connor works on this project anymore.

haiminh2001:
Hi all, can I ask when will these feature be released ? And I expect that this feature only requires me to update my python KFP SDK, is that true?

gregsheremeta:
The change requires both a backend update and an SDK update. The project doesn't currently have a fixed release schedule, but I'll ask about the next release at this week's KFP community meeting (tomorrow, Wednesday August 28 at 1PM US Eastern Time, [agenda](http://bit.ly/kfp-meeting-notes))

haiminh2001:
@gregsheremeta great! Thank you.

gregsheremeta:
Update from the meeting: @chensun is ready to to both backend and SDK releases. There's currently a failing test blocking him, but as soon as that's fixed, they should be out. I don't have an ETA but shouldn't be too long.

gregsheremeta:
[...]

PR #10892, opened 2024-06-12
feat(kubernetes_platform): Add empty dir mount

gregsheremeta:
**Description of your changes:**
Update kubernetes_platform go package to include the ability to add an EmptyDir mount. 

Part 1 of 2. Will be rebased when the backend part is merged. Posting as a WIP now to get feedback on the fields in the proto message.

Following the example set in #10410.

**Checklist:**
- [x] The title for your pull request (PR) should follow our title convention. [Learn more about the pull request title convention used in this repository](https://github.com/kubeflow/pipelines/blob/master/CONTRIBUTING.md#pull-request-title-convention).

PR #10410, opened 2024-01-18
feat(kubernetes_platform): Update kubernetes_platform go package to include ImagePullSecrets

JosepSampe:
Update kubernetes_platform go package to include ImagePullSecrets.

**Description of your changes:**
Part of https://github.com/kubeflow/pipelines/issues/10262

**Checklist:**
- [x] The title for your pull request (PR) should follow our title convention. [Learn more about the pull request title convention used in this repository](https://github.com/kubeflow/pipelines/blob/master/CONTRIBUTING.md#pull-request-title-convention).

PR #10913, opened 2024-06-17
feat(Backend + SDK): Update kfp backend and kubernetes sdk to support EmptyDir

gregsheremeta:
Update kfp backend and kubernetes sdk to support mounting EmptyDir volumes to task pods.

(Based on #10892 and needs rebase after that is merged first)

Inspired by #10427

Fixes: #10656

**Checklist:**

### DIFF
--- a/backend/src/v2/driver/driver.go
+++ b/backend/src/v2/driver/driver.go
@@ -665,6 +665,33 @@ func extendPodSpecPatch(
 		podSpec.Volumes = append(podSpec.Volumes, ephemeralVolume)
 		podSpec.Containers[0].VolumeMounts = append(podSpec.Containers[0].VolumeMounts, ephemeralVolumeMount)
 	}
+
+	// EmptyDirMounts
+	for _, emptyDirVolumeSpec := range kubernetesExecutorConfig.GetEmptyDirMounts() {
+		var sizeLimitResource *k8sres.Quantity
+		if emptyDirVolumeSpec.GetSizeLimit() != "" {
+			r := k8sres.MustParse(emptyDirVolumeSpec.GetSizeLimit())
+			sizeLimitResource = &r
+		}
+
+		emptyDirVolume := k8score.Volume{
+			Name: emptyDirVolumeSpec.GetVolumeName(),
+			VolumeSource: k8score.VolumeSource{
+				EmptyDir: &k8score.EmptyDirVolumeSource{
+					Medium:    k8score.StorageMedium(emptyDirVolumeSpec.GetMedium()),
+					SizeLimit: sizeLimitResource,
+				},
+			},
+		}
+		emptyDirVolumeMount := k8score.VolumeMount{
+			Name:      emptyDirVolumeSpec.GetVolumeName(),
+			MountPath: emptyDirVolumeSpec.GetMountPath(),
+		}
+
+		podSpec.Volumes = append(podSpec.Volumes, emptyDirVolume)
+		podSpec.Containers[0].VolumeMounts = append(podSpec.Containers[0].VolumeMounts, emptyDirVolumeMount)
+	}
+
 	return nil
 }
 
--- a/kubernetes_platform/python/test/snapshot/data/empty_dir_mounts.py
+++ b/kubernetes_platform/python/test/snapshot/data/empty_dir_mounts.py
@@ -1,0 +1,36 @@
+# Copyright 2024 The Kubeflow Authors
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#      http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+from kfp import dsl
+from kfp import kubernetes
+
+
+@dsl.component
+def comp():
+    pass
+
+@dsl.pipeline
+def my_pipeline():
+    task = comp()
+    kubernetes.empty_dir_mount(
+            task, 
+            volume_name='emptydir-vol-1', 
+            mount_path='/mnt/my_vol_1',
+            medium='Memory',
+            size_limit='1Gi'
+        )
+
+if __name__ == '__main__':
+    from kfp import compiler
+    compiler.Compiler().compile(my_pipeline, __file__.replace('.py', '.yaml'))
--- a/kubernetes_platform/python/test/snapshot/data/empty_dir_mounts.yaml
+++ b/kubernetes_platform/python/test/snapshot/data/empty_dir_mounts.yaml
@@ -1,0 +1,60 @@
+# PIPELINE DEFINITION
+# Name: my-pipeline
+components:
+  comp-comp:
+    executorLabel: exec-comp
+deploymentSpec:
+  executors:
+    exec-comp:
+      container:
+        args:
+        - --executor_input
+        - '{{$}}'
+        - --function_to_execute
+        - comp
+        command:
+        - sh
+        - -c
+        - "\nif ! [ -x \"$(command -v pip)\" ]; then\n    python3 -m ensurepip ||\
+          \ python3 -m ensurepip --user || apt-get install python3-pip\nfi\n\nPIP_DISABLE_PIP_VERSION_CHECK=1\
+          \ python3 -m pip install --quiet --no-warn-script-location 'kfp==2.7.0'\
+          \ '--no-deps' 'typing-extensions>=3.7.4,<5; python_version<\"3.9\"' && \"\
+          $0\" \"$@\"\n"
+        - sh
+        - -ec
+        - 'program_path=$(mktemp -d)
+
+
+          printf "%s" "$0" > "$program_path/ephemeral_component.py"
+
+          _KFP_RUNTIME=true python3 -m kfp.dsl.executor_main                         --component_module_path                         "$program_path/ephemeral_component.py"                         "$@"
+
+          '
+        - "\nimport kfp\nfrom kfp import dsl\nfrom kfp.dsl import *\nfrom typing import\
+          \ *\n\ndef comp():\n    pass\n\n"
+        image: python:3.9
+pipelineInfo:
+  name: my-pipeline
+root:
+  dag:
+    tasks:
+      comp:
+        cachingOptions:
+          enableCache: true
+        componentRef:
+          name: comp-comp
+        taskInfo:
+          name: comp
+schemaVersion: 2.1.0
+sdkVersion: kfp-2.7.0
+---
+platforms:
+  kubernetes:
+    deploymentSpec:
+      executors:
+        exec-comp:
+          emptyDirMounts:
+          - medium: Memory
+            mountPath: /mnt/my_vol_1
+            sizeLimit: 1Gi
+            volumeName: emptydir-vol-1
--- a/kubernetes_platform/python/test/unit/test_empty_dir_mounts.py
+++ b/kubernetes_platform/python/test/unit/test_empty_dir_mounts.py
@@ -1,0 +1,136 @@
+# Copyright 2024 The Kubeflow Authors
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#      http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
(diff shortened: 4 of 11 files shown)


kubernetes_platform/python/test/snapshot/data/empty_dir_mounts.py
gregsheremeta, line 1: I don't know how this is used. I cargo culted it from #10427

kubernetes_platform/python/test/snapshot/data/empty_dir_mounts.yaml
gregsheremeta, line 1: I don't know how this is used. I cargo culted it from #10427

droctothorpe:
Thanks for tackling this! It's an important gap between v1 and v2.

gregsheremeta:
@chensun can I ask for your review, or can you recommend someone to review? thanks.

rimolive:
@gregsheremeta I can help with the backend review, but for SDK we have only @connor-mccarthy. I can review both and see how can we proceed with approvals.

gregsheremeta:
thanks. For what it's worth, the SDK piece is fairly simple, and Chen gave a preliminary lgtm over in #10892. It follows the (already approved and merged) example set in #10410. The key thing to review there is that I included the important fields. When you are doing an emptydir mount in kube, there are four pieces of information required (ref: https://kubernetes.io/docs/reference/kubernetes-api/config-and-storage-resources/volume/#local-temporary-directory) : name, mount path, medium (optional), and size limit (optional)

gregsheremeta:
actually, you may have been referring to the code in `empty_dir.py` :) In which case, kind of the same answer. It has the same pattern as (say) [secrets](https://github.com/kubeflow/pipelines/blob/master/kubernetes_platform/python/kfp/kubernetes/secret.py#L60) but that I have these 4 fields instead.

gregsheremeta:
The failing tests are expected. When #10892 is merged, I'll rebase this, and then the tests will work

@chensun @james-jwu @zijianjoy @connor-mccarthy can I ask for your review, or can you recommend someone to review? thanks.

kubernetes_platform/python/test/snapshot/data/empty_dir_mounts.yaml
```diff
@@ -0,0 +1,60 @@
...
+
+
+          printf "%s" "$0" > "$program_path/ephemeral_component.py"
+
+          _KFP_RUNTIME=true python3 -m kfp.dsl.executor_main                         --component_module_path                         "$program_path/ephemeral_component.py"                         "$@"
+
+          '
+        - "\nimport kfp\nfrom kfp import dsl\nfrom kfp.dsl import *\nfrom typing import\
+          \ *\n\ndef comp():\n    pass\n\n"
+        image: python:3.7
```
hbelmiro, line 35: We no longer use 3.7. Can you use 3.9 since that's the version you're using in the command above?

```suggestion
        image: python:3.9
```

kubernetes_platform/python/test/unit/test_empty_dir_mounts.py
```diff
@@ -0,0 +1,136 @@
...
+from kfp import kubernetes
+
+
+class TestEmptyDirMounts:
+
+    def test_add_one(self):
+
+        @dsl.pipeline
+        def my_pipeline():
+            task = comp()
```
hbelmiro, line 26: Function `comp` doesn't return anything.

kubernetes_platform/python/test/unit/test_empty_dir_mounts.py
```diff
@@ -0,0 +1,136 @@
...
+                    }
+                }
+            }
+        }
+
+    def test_add_two(self):
+
+        @dsl.pipeline
+        def my_pipeline():
+            task = comp()
```
hbelmiro, line 58: Function `comp` doesn't return anything.

kubernetes_platform/python/test/unit/test_empty_dir_mounts.py
```diff
@@ -0,0 +1,136 @@
...
+                    }
+                }
+            }
+        }
+
+    def test_respects_other_configuration(self):
+
+        @dsl.pipeline
+        def my_pipeline():
+            task = comp()
```
hbelmiro, line 99: Function `comp` doesn't return anything.

kubernetes_platform/python/test/snapshot/data/empty_dir_mounts.yaml
gregsheremeta: sure. note that I copied this test from an existing test.

kubernetes_platform/python/test/unit/test_empty_dir_mounts.py
gregsheremeta: not explicitly, but it's wrapped in a `@dsl.component`, and a Component is what is returned and then passed to `empty_dir_mount` on line 28.

kubernetes_platform/python/test/unit/test_empty_dir_mounts.py
gregsheremeta: same as above

kubernetes_platform/python/test/unit/test_empty_dir_mounts.py
gregsheremeta: same as above

thesuperzapper:
@chensun @connor-mccarthy I am not sure why these tests are failing, but I would appreciate your review of this feature, as not being able to mount `emptyDir` to tasks is preventing a number of people updating to KFP v2.

hbelmiro:
@gregsheremeta FYI all the GitHub Actions tests are mandatory. Some of the Prow ones may be optional. You can check which of them are optional [here](https://github.com/GoogleCloudPlatform/oss-test-infra/blob/3738b70f78b971cc281b1791cf8e73cfbe08ae55/prow/prowjobs/kubeflow/pipelines/kubeflow-pipelines-presubmits.yaml).

kubernetes_platform/python/test/snapshot/data/empty_dir_mounts.yaml
gregsheremeta: i'm good if y'all are good!

kubernetes_platform/python/test/snapshot/data/empty_dir_mounts.yaml
gregsheremeta: [done](https://github.com/kubeflow/pipelines/pull/10913/files#diff-cac76397fcf6a375f3865a864a3f6725b023c69b5216fa1ed71a86b334641399R35)

backend/src/v2/driver/driver.go
```diff
@@ -36,6 +36,7 @@ import (
 	"google.golang.org/protobuf/encoding/protojson"
 	"google.golang.org/protobuf/types/known/structpb"
 	k8score "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/api/resource"
```
gregsheremeta, line 39: oops, duplicate import. fixing...

backend/src/v2/driver/driver.go
```diff
@@ -665,6 +665,33 @@ func extendPodSpecPatch(
...
+				},
+			},
+		}
+		emptyDirVolumeMount := k8score.VolumeMount{
+			Name:      emptyDirVolumeSpec.GetVolumeName(),
+			MountPath: emptyDirVolumeSpec.GetMountPath(),
+		}
+
+		podSpec.Volumes = append(podSpec.Volumes, emptyDirVolume)
+		podSpec.Containers[0].VolumeMounts = append(podSpec.Containers[0].VolumeMounts, emptyDirVolumeMount)
```
DharmitD, line 692: The code assumes that `Containers[0]` is always valid and the target container for mounting the volume. If `podSpec.Containers` is empty or if there are multiple containers in the pod, this could lead to errors or incorrect behavior. 
Consider adding a check to ensure the container exists or iterating through all containers if the volume should be mounted to more than one. Something like:

```
if len(podSpec.Containers) == 0 {
    return fmt.Errorf("podSpec has no containers to mount volumes to")
}
```

backend/src/v2/driver/driver.go
```diff
@@ -665,6 +665,33 @@ func extendPodSpecPatch(
 		podSpec.Volumes = append(podSpec.Volumes, ephemeralVolume)
 		podSpec.Containers[0].VolumeMounts = append(podSpec.Containers[0].VolumeMounts, ephemeralVolumeMount)
 	}
+
+	// EmptyDirMounts
+	for _, emptyDirVolumeSpec := range kubernetesExecutorConfig.GetEmptyDirMounts() {
+		var sizeLimitResource *k8sres.Quantity
```
DharmitD, line 671: We could add some logging here to indicate the attempt to create and mount the EmptyDir volume. 
```
log.Infof("Attempting to create EmptyDir volume '%s' and mount it at '%s' with medium '%s'",
        emptyDirVolumeSpec.GetVolumeName(),
        emptyDirVolumeSpec.GetMountPath(),
        emptyDirVolumeSpec.GetMedium(),
    )
   ```
   Just a nice to have, no compulsion :)

Comment (DharmitD) [inline backend/src/v2/driver/driver.go:671]: Could add a check for the podSpec.Volumes slice to see if a volume with the same name already exists. If a duplicate is found, it logs an error and skips the creation of the volume.
```
if emptyDirVolumeSpec.GetSizeLimit() != "" {
        r, err := k8sres.ParseQuantity(emptyDirVolumeSpec.GetSizeLimit())
        if err != nil {
            log.Errorf("Failed to parse size limit '%s' for volume '%s': %v",
                emptyDirVolumeSpec.GetSizeLimit(),
                emptyDirVolumeSpec.GetVolumeName(),
                err)
            continue // Skip this volume since size limit parsing failed
        }
        sizeLimitResource = &r
    }
    ```

backend/src/v2/driver/driver.go
```diff
@@ -665,6 +665,33 @@ func extendPodSpecPatch(
...
+			},
+		}
+		emptyDirVolumeMount := k8score.VolumeMount{
+			Name:      emptyDirVolumeSpec.GetVolumeName(),
+			MountPath: emptyDirVolumeSpec.GetMountPath(),
+		}
+
+		podSpec.Volumes = append(podSpec.Volumes, emptyDirVolume)
+		podSpec.Containers[0].VolumeMounts = append(podSpec.Containers[0].VolumeMounts, emptyDirVolumeMount)
+	}
```
DharmitD, line 693: Again, some logging here would be nice :)

```
log.Infof("Successfully created EmptyDir volume '%s' and mounted it at '%s'",
        emptyDirVolumeSpec.GetVolumeName(),
        emptyDirVolumeSpec.GetMountPath(),
    )
    ```
    Feel free to skip if you think it would be too much verbose.

**Comment by DharmitD** [comment]
Added some nitpick error handling and logging suggestions for the driver code. The Kubernetes Platform / SDK code and the corresponding tests LGTM.

File: backend/src/v2/driver/driver.go (hunk 6 above)
Comment (gregsheremeta) [reply backend/src/v2/driver/driver.go:692]: had to refresh my memory on this :)

I copied this block from the other blocks already in driver. I think you're exactly right and hardcoding the `[0]` is not great, but I want to tackle all of them in a refactor PR that we can do later.

Examples: [1](https://github.com/kubeflow/pipelines/blob/2e6e634de43c42fee88dfe2abfdb0e4155dc4f95/backend/src/v2/driver/driver.go#L487-L488), [2](https://github.com/kubeflow/pipelines/blob/2e6e634de43c42fee88dfe2abfdb0e4155dc4f95/backend/src/v2/driver/driver.go#L505-L506), [3](https://github.com/kubeflow/pipelines/blob/2e6e634de43c42fee88dfe2abfdb0e4155dc4f95/backend/src/v2/driver/driver.go#L565) ... you get the idea.

File: backend/src/v2/driver/driver.go (hunk 7 above)
Comment (gregsheremeta) [reply backend/src/v2/driver/driver.go:671]: +1, gonna do this in a followup overall refactor of driver :)

Comment (gregsheremeta) [reply backend/src/v2/driver/driver.go:671]: 2 lines below we have
```
r := k8sres.MustParse(emptyDirVolumeSpec.GetSizeLimit())
```

`MustParse` will panic if the resource can't be parsed, and that makes sense to me. In other words, I don't think we should keep going and try to run the user's workload if the volume can't be mounted. If the volume can't be mounted, the workload is likely to fail anyway, so better to fail faster.

backend/src/v2/driver/driver.go
gregsheremeta: +1, will do in a followup driver refactor

DharmitD (commented):
/lgtm

HumairAK:
/lgtm
/approve

Thanks!

gregsheremeta:
@chensun this is ready for your final review / merge. Can you take a look?

chensun (approved):

droctothorpe:
Congrats on the merge, @gregsheremeta. Thank you for tackling this meaningful contribution! 🔥

PR #10927, opened 2024-06-20
Add an example about Federated-Learning

sefgsefg:


sefgsefg:
I have to change my code to kfpv2 ?

PR #10947, opened 2024-06-24
add(pipeline/samples/contrib/QPE_pipeline_example): a pipeline example about QPE

Henry551:
Pull Request: Add Quantum Phase Estimation (QPE) Pipeline

This PR is about a Kubeflow pipeline for running the Quantum Phase Estimation (QPE) algorithm using Qiskit. The pipeline is designed to leverage the power of quantum computing for estimating the phases of eigenvalues.


Henry551:
The code I've tested on kubeflow and it worked. I see the AttributeError in the test and said: module 'kfp_server_api' has no attribute 'ApiGetHealthzResponse',  but the environment I use is python=3.11.7, how can I fix it?

hbelmiro:
> @rimolive :The code I've tested on kubeflow and it worked. I see the AttributeError in the test and said: module 'kfp_server_api' has no attribute 'ApiGetHealthzResponse', but the environment I use is python=3.11.7, how can I fix it?

@Henry551 I'm migrating the `kubeflow-pipelines-samples-v2` tests to GitHub Actions and facing this issue too. When I complete this migration the problem should be fixed.

It would be nice to have a test for this new sample you're adding. However, since the existing tests are not working, we need to wait for the migration to GitHub Actions.

Can we put this PR on hold in this meantime?

/hold

hbelmiro (changes requested):
@Henry551 [now that we have tests for samples](https://github.com/kubeflow/pipelines/pull/11048), can you please add a test for this?

Henry551:
@hbelmiro How could I add test for this? Should I add a Github Action?

hbelmiro:
> @hbelmiro How could I add test for this? Should I add a Github Action?

No. You should add a new test case here: https://github.com/kubeflow/pipelines/blob/0d098db41c09c5b65c374f4e110d188b982f544b/samples/v2/sample_test.py

rimolive:
@Henry551 Please sign-off your commits

PR #10949, opened 2024-06-24
feat(samples/contrib/modular_exponentiation_quantum_circuit)

Rhm-B-WT:
Create an example of modular exponentiation quantum circuit in kubflow pipeline

### DIFF
--- a/samples/contrib/modular_exponentiation_quantum_circuit/QPE.sh
+++ b/samples/contrib/modular_exponentiation_quantum_circuit/QPE.sh
@@ -1,0 +1,28 @@
+#!/bin/bash
+#
+# Copyright 2021 The Kubeflow Authors
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#      http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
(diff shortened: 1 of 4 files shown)


Rhm-B-WT:
I've tested the code on kubeflow and it worked. The AttributeError said "module 'kfp_server_api' has no attribute 'ApiGetHealthzResponse'". The environment I used is python 3.11.7. How can I fix it?

hbelmiro:
> I've tested the code on kubeflow and it worked. The AttributeError said "module 'kfp_server_api' has no attribute 'ApiGetHealthzResponse'". The environment I used is python 3.11.7. How can I fix it?

@Rhm-B-WT  I'm migrating the kubeflow-pipelines-samples-v2 tests to GitHub Actions and facing this issue too. When I complete this migration the problem should be fixed.

It would be nice to have a test for this new sample you're adding. However, since the existing tests are not working, we need to wait for the migration to GitHub Actions.

Can we put this PR on hold in the meantime?

/hold

hbelmiro (changes requested):
@Rhm-B-WT [now that we have tests for samples](https://github.com/kubeflow/pipelines/pull/11048), can you please add a test for this?

hbelmiro (changes requested):
@Rhm-B-WT you don't need to create any scripts to run the test. You just need to add your test to https://github.com/kubeflow/pipelines/blob/ba006bddcb83e34bc68c3db15b09fbb0bc317f6c/samples/v2/sample_test.py#L46

Issue #10968, opened 2024-06-28
[backend] The v2 AWF compiler inadvertently limits workflow size to 256KB. 


droctothorpe:
### Environment

*  How did you deploy Kubeflow Pipelines (KFP)?
K8s on AWS using the KFP chart.

*  KFP version: 
2.2.0

*  KFP SDK version: 
2.7.0

### Steps to reproduce

Submit the following pipeline:

```
from kfp import Client, dsl


@dsl.component()
def comp(name: str) -> str:
    # Lorem ipsum below to bloat pipeline/component definition size.
    """_summary_

    Args:
        name (str): _description_

    Returns:
        str: _description_

    Lorem ipsum dolor sit amet, consectetur adipiscing elit, sed do eiusmod tempor incididunt ut labore et dolore magna aliqua. Sodales neque sodales ut etiam. At urna condimentum mattis pellentesque. Mattis molestie a iaculis at erat pellentesque. Facilisis magna etiam tempor orci. Tellus at urna condimentum mattis. Lobortis feugiat vivamus at augue eget arcu dictum. Et malesuada fames ac turpis egestas integer. Sollicitudin nibh sit amet commodo nulla facilisi nullam vehicula. Eros in cursus turpis massa tincidunt dui ut. Varius vel pharetra vel turpis nunc eget.

    Cras semper auctor neque vitae. Adipiscing diam donec adipiscing tristique risus nec feugiat in fermentum. Purus sit amet volutpat consequat mauris. Dictum sit amet justo donec enim diam vulputate. Non blandit massa enim nec dui nunc mattis. Ipsum dolor sit amet consectetur. Vitae tortor condimentum lacinia quis vel eros donec ac. Arcu ac tortor dignissim convallis aenean et. Ut faucibus pulvinar elementum integer enim neque volutpat ac. Nibh venenatis cras sed felis eget velit aliquet sagittis. Pulvinar sapien et ligula ullamcorper. Egestas quis ipsum suspendisse ultrices gravida dictum fusce. Euismod lacinia at quis risus. Quis viverra nibh cras pulvinar. Risus commodo viverra maecenas accumsan lacus vel facilisis volutpat est.

[...]

Issue #4170, opened 2020-07-08
[Limitation] SDK Complier creates bigger YAML exceeding Kubeflow limits quickly


radcheb:
### What steps did you take:
This problem is only noticed in `1.0.0-rc.3`, it does not exist in `0.5.1`.
Given the same pipeline, `1.0.0-rc.3` Complier creates a YAML file far bigger than the same created with `0.5.1` Compiler.
Thus, this may be a problem for huge workflows, that previously run successfully and can no more be run if compiled with new Compiler.

### What happened:
The pipeline compiled with `1.0.0-rc.3` fails to run with the following error: `Workflow is longer than maximum allowed size. Size=1048602`
![image](https://user-images.githubusercontent.com/5963615/86892643-d0796980-c100-11ea-861e-5152458a3b5c.png)

### What did you expect to happen:
The pipeline must run successfully

### Environment:
Kubeflow 1.0
Build commit: `743746b`
python `3.7` + kfp `1.0.0-rc.3`

How did you deploy Kubeflow Pipelines (KFP)?

### Anything else you would like to add: : Reproducing the limit

For this same pipeline:
```python
from kfp import Client
from kfp.components import func_to_container_op
from kfp.dsl import pipeline

def some_func(i: int) -> str:
    msg = f"""{i}
    This is a huge function, with a lot of code.
    Lorem Ipsum is simply dummy text of the printing and typesetting industry. 
    Lorem Ipsum has been the industry's standard dummy text ever since the 1500s, 
    when an unknown printer took a galley of type and scrambled it to make a type specimen book. 
    It has survived not only five centuries, but also the leap into electronic typesetting, 
    remaining essentially unchanged. It was popularised in the 1960s with the release of Letraset sheets 
    containing Lorem Ipsum passages, and more recently with desktop publishing software like Aldus PageMaker 
    including versions of Lorem Ipsum.
    """
    return msg

@pipeline(name="huge pipeline")
def test_pipeline():
    
    component = func_to_container_op(func=some_func, base_image="library/python:3.6")
 
    previous = None
    for i in range(239):
        op = component(i=i)
        if previous:
            op.after(previous)
        previous = op

Compiler().compile(test_pipeline, package_path="toto.yaml")
```
If compiled with kfp `0.5.1` it produces a file of size **625kb** while if compiled with kfp `1.0.0-rc.3` it produces a file of size **1129kb** and thus fails to run in cluster.

[...]

PR #10979, opened 2024-07-03
feat(backend): move comp logic to workflow params

zazulam:
This PR relocates component logic from Workflow Annotations to Workflow Arguments.

In addition, it deduplicates the component logic, dramatically reducing the size of the compiled Workflow (in our preliminary testing we saw as much as a 70% reduction in Workflow size).

They’re still referenced by template tag in the actual Argo Workflow Templates, the tags are just pointing to Workflow Arguments instead of Workflow Annotations now.

The testdata manifests diff neatly illustrates how the compiled Workflow manifest has been improved.

This resolves  https://github.com/kubeflow/pipelines/issues/10968.

Before this PR, the Argo Workflow compiler was storing component logic as Workflow Annotations. Since Kubernetes Annotations cannot exceed 256 KB in size (cumulatively for a given object), this limited the maximum size of any given KFP DAG. Furthermore, when a pipeline referenced a single component multiple times, that component’s logic was duplicated in the Workflow Annotations each time it was referenced, unnecessarily inflating the workflow size.

**Checklist:**


zijianjoy:
/lgtm

@chensun Please review and approve if this makes sense to you.

zijianjoy:
@HumairAK Thank you for volunteering on testing this implementation!

droctothorpe:
@zijianjoy brought up a great question about potential backward compatibility issues.

The IR is the source of truth and is not impacted by this change. The compiled Workflow is a downstream artifact / representation. Whether it compiles to a Workflow with Annotations or a Workflow with Parameters shouldn't impact end users, as long as the source of truth (the IR) stays the same, and it does. If a user has built abstractions on top of KFP that interact directly with the Workflow, they could conceivably be impacted.

backend/src/v2/compiler/argocompiler/argo.go
```diff
@@ -180,69 +184,96 @@ func (c *workflowCompiler) templateName(componentName string) string {
...
+	baseComponentName := componentName
+	componentNameArray := strings.Split(componentName, "-")
+
+	if _, err := strconv.Atoi(componentNameArray[len(componentNameArray)-1]); err == nil {
+		baseComponentName = strings.Join(componentNameArray[:len(componentNameArray)-1], "-")
+
+	}
+
+	return baseComponentName
 }
```
HumairAK, line 277: If I recall correctly, component function names that have underscores also get converted to dashes `-` in the IR `.components` section, example: 

```python
@dsl.component()
def foo_1(name: str) -> str:
    hello_text = f"Hello 2, {name}!"
    print(hello_text)
    return hello_text
```

would result in: 

```yaml
components:
  comp-foo-1:
```

In this case it seems like this function would use the baseComponentName `comp-foo`, which seems incorrect no?

Another alternative could be to introduce a separate sdk field that keeps the component name in check, maybe [here](https://github.com/kubeflow/pipelines/blob/master/api/v2alpha1/pipeline_spec.proto#L88), but then we have to consider backwards incompatibility.

backend/src/v2/compiler/argocompiler/argo.go
```diff
@@ -180,69 +184,96 @@ func (c *workflowCompiler) templateName(componentName string) string {
...
+
+// extractBaseComponentName removes the iteration suffix that the IR compiler
+// adds to the component name.
+func ExtractBaseComponentName(componentName string) string {
+	baseComponentName := componentName
+	componentNameArray := strings.Split(componentName, "-")
+
+	if _, err := strconv.Atoi(componentNameArray[len(componentNameArray)-1]); err == nil {
+		baseComponentName = strings.Join(componentNameArray[:len(componentNameArray)-1], "-")
+
```
HumairAK, line 273: nit: remove whitespace

backend/src/v2/compiler/argocompiler/argo.go
```diff
@@ -180,69 +184,96 @@ func (c *workflowCompiler) templateName(componentName string) string {
...
 
 func (c *workflowCompiler) saveKubernetesSpec(name string, spec *structpb.Struct) error {
-	return c.saveProtoToAnnotation(annotationKubernetesSpec+name, spec)
+	return c.saveProtoToArguments(argumentsKubernetesSpec+name, spec)
 }
 
 func (c *workflowCompiler) useKubernetesImpl(name string) (string, error) {
-	return c.annotationPlaceholder(annotationKubernetesSpec + name)
+	return c.argumentsPlaceholder(argumentsKubernetesSpec + name)
 }
```
HumairAK, line 221: was there a reason not to deduplicate here? 

I would think kubernetesSpec benefit from this in the same way:

e.g: 

```python
    for i in range(2):
        task = bar(name="bar")
        kubernetes.use_secret_as_env(
            task,
            secret_name='aws-s3-creds',
            secret_key_to_env={'AWS_SECRET_ACCESS_KEY': 'AWS_SECRET_ACCESS_KEY'})
```

HumairAK (changes requested):

backend/src/v2/compiler/argocompiler/argo.go
droctothorpe: Yup! Nice catch! This is an edge case that we were thinking of tackling in a discrete PR, since the component naming logic already gets confused -- if you have a component named `foo` and a component named `foo_2` and you invoke `foo` twice, you wind up with components named `comp-foo`, `comp-foo-2`, and `comp-foo-2-2` in the IR. We can include that edge case handling in this PR though. The easiest way to do that would be to raise an exception in the SDK when a component function name culminates in `_<int(s)>`. Any backward compatibility concerns would only take place at the IR compilation stage, which (a) happens client-side and is easily resolved, and (b) would be an exceedingly rare edge case. Open to alternative approaches ofc!

backend/src/v2/compiler/argocompiler/argo.go
droctothorpe: `kubernetesSpec` can vary across multiple instances of a given component, e.g.:

```python
from kfp import dsl, kubernetes

@dsl.component
def foo():
    print("foo")

@dsl.pipeline
def pipeline():
    task1 = foo()
    task2 = foo()
    kubernetes.add_pod_annotation(
        task1,
        annotation_key='run_id',
        annotation_value='123456',
    )
```

In the above example, `task1` will have a distinct `kubernetesSpec` vis a vis `task2`. Deduplicating the `kubernetesSpec` will fail to account for those distinctions.

Caveat about the above: **...I think** 🙃 .

Really appreciate the thorough review, @HumairAK!

backend/src/v2/compiler/argocompiler/argo.go
HumairAK: >  if you have a component named foo and a component named foo_2 and you invoke foo twice, you wind up with components named comp-foo, comp-foo-2, and comp-foo-2-2 

Yeah in this case it looks like it would extract the right base names for each. The edge case I mentioned will occur every time you have components with unique base names with a `_<int>` prefix.

The current approach feels a little hacky to me, because we are trying to rebuild info that should have been provided by the sdk initially. For example, in the alternative I proposed we could have sdk provide something like: 

```yaml
components:
  comp-foo-1:
    componentName: foo
    executorLabel: exec-foo-1
  comp-foo-1-1:
    componentName: foo-1
    executorLabel: exec-foo-1-2
```
And just use the component name `componentName`, instead of inferring it. Due to backwards incompatibility this would require we support this only future KFP versions though.

Anyways, food for thought. It is an edge case as you mentioned that is likely very rare. 

> We can include that edge case handling in this PR though. The easiest way to do that would be to raise an exception in the SDK when a component function name culminates in _<int(s)>.  

I'm personally fine with this, we should provide documentations on component naming maybe [here](https://www.kubeflow.org/docs/components/pipelines/concepts/component/).

I would also suggest adding some simple unit tests to illustrate the handling of the different component name formats.

backend/src/v2/compiler/argocompiler/argo_test.go
```diff
@@ -137,3 +137,30 @@ func load(t *testing.T, path string, platformSpecPath string) (*pipelinespec.Pip
...
 	return job, nil
 }
+
+func Test_extractBaseComponentName(t *testing.T) {
+	tests := []struct {
+		name             string
+		componentName    string
+		expectedBaseName string
+	}{
+		{
```
diegolovison, line 147: Please add more test instructions. Example:
```
{
   componentName:    "component-123",
   ...
},
{
   componentName:    "service-api-v2",
   ...
},
{
   componentName:    "module-3-1",
   ...
},
{
   componentName:    "foo-bar-1-2-3",
   ...
},
```

backend/src/v2/compiler/argocompiler/argo.go
HumairAK: very good point! makes sense to me

backend/src/v2/compiler/argocompiler/argo.go
```diff
@@ -180,69 +184,96 @@ func (c *workflowCompiler) templateName(componentName string) string {
...
 	}
-	if c.wf.Annotations == nil {
-		c.wf.Annotations = make(map[string]string)
+	if c.wf.Spec.Arguments.Parameters == nil {
+		c.wf.Spec.Arguments = wfapi.Arguments{Parameters: []wfapi.Parameter{}}
 	}
-	if _, alreadyExists := c.wf.Annotations[name]; alreadyExists {
-		return fmt.Errorf("annotation %q already exists", name)
+	if c.wf.Spec.Arguments.GetParameterByName(componentName) != nil {
+		return nil
```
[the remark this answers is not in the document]
zazulam: Yes, based on the component name we are inferring we want to know if it already exists in the parameters so we don't add any duplicated information. In this scenario we do not want to raise an error as we just want to store only unique parameters.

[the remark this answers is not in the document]
HumairAK: Makes perfect sense, thanks!

hbelmiro:
@droctothorpe @zazulam FYI due to https://github.com/kubeflow/pipelines/issues/10981#issue-2391320390, we need to trigger `/rerun-all` to run the GitHub Actions tests when a PR comes from a new contributor.

droctothorpe:
Thanks, @hbelmiro! Are there known issues with the sdk execution tests? The prow logs don't provide any debugging info and the links to the actual executions just 403.

```
E   asyncio.exceptions.CancelledError
/usr/local/lib/python3.8/site-packages/pytest_asyncio_cooperative/plugin.py:287: CancelledError
```

HumairAK:
CI aside, the code lgtm
though note, I'm not that familiar with the sdk side of things, I tested the changes and they work as intended: 

image
image
image

I would suggest bringing this up in the upcoming kfp community call so we can run by the maintainers the change to the name requirements

this is great, thanks folks!

/lgtm

droctothorpe:
Thank you so much for validating, @HumairAK! Planning to be at the pipelines community call on Wednesday.

HumairAK:
@chensun can you please take a look? PR functionality looks good, if the maintainers are fine with adding a constraint to component naming see [here for the discussion](https://github.com/kubeflow/pipelines/pull/10979#discussion_r1672595012) then we should be good to merge

backend/src/v2/compiler/argocompiler/argo.go
chensun: I would like to avoid introducing naming constraints if possible. Introducing a naming constraints is a breaking changes. 
The current state of appending component name with index suffix isn't ideal, but that's more of an implementation detail which can be changed without user impact, and we can dedupe them: https://github.com/kubeflow/pipelines/blob/eea7d483793c2b4f8c3aa20b1500535f7fbf50d4/sdk/python/kfp/deprecated/dsl/_component_bridge.py#L675

sdk/python/kfp/dsl/component_factory.py
```diff
@@ -66,6 +66,13 @@ class ComponentInfo():
 
 def _python_function_name_to_component_name(name):
     name_with_spaces = re.sub(' +', ' ', name.replace('_', ' ')).strip(' ')
+    name_list = name_with_spaces.split(' ')
+
+    if name_list[-1].isdigit():
+        raise ValueError(
+            f'Invalid function name "{name}". The function name must not end in `_<int>`.'
```
chensun, line 73: Can we focus on the backend change only without introduce a breaking change?

chensun (changes requested):

backend/src/v2/compiler/argocompiler/argo.go
droctothorpe: Copy that. Thanks, @chensun!

One solution could be to update [ExtractBaseComponentName](https://github.com/kubeflow/pipelines/pull/10979/files#diff-481aa8f29d58257fef3e48fb60fd320f73d50b858018fe6adf7ac583125fc2c6R267) to return the full component name if the IR includes a task with a name in the format of `.*-\d+-\d+`. If you think this makes sense, maybe @zazulam and I can crank it out tomorrow morning before the community call. Just have to confirm with @zazulam about bandwidth. 

Another approach could  preserving dash `-` characters in the IR instead of converting them into underscores `_`. That way we could use `-<index>` in `make_name_unique_by_adding_index` and not have to worry about collisions since python function names cannot include dashes. Sounds easy in theory but maybe including dashes violates expectations in the protobuf / backend.

sdk/python/kfp/dsl/component_factory.py
droctothorpe: Alternative design discussion continuing in [this](https://github.com/kubeflow/pipelines/pull/10979#discussion_r1697542558) thread.

backend/src/v2/compiler/argocompiler/argo.go
HumairAK: > The current state of appending component name with index suffix isn't ideal, but that's more of an implementation detail which can be changed without user impact, and we can dedupe them: 

if we don't address it here, and we fear this breaking change can impact users, then we still need to resolve this as it would introduce a regression for this niche case. @droctothorpe correct me if I'm wrong here, it's been a couple of weeks since I looked at this, but in this case the component can assume the wrong parameters if paired with another component incorrectly.

> One solution could be to update [ExtractBaseComponentName](https://github.com/kubeflow/pipelines/pull/10979/files#diff-481aa8f29d58257fef3e48fb60fd320f73d50b858018fe6adf7ac583125fc2c6R267) to return the full component name if the IR includes a task with a name in the format of .*-\d+-\d+.

so I understand, this sounds like the alternative I proposed [here yes](https://github.com/kubeflow/pipelines/pull/10979#discussion_r1672251048)? If so I like this approach more. We can add it so as not to introduce any regression, and only support this for future sdk/kfp releases.

droctothorpe: You mean [this](https://github.com/kubeflow/pipelines/pull/10979#discussion_r1672595012) specific suggestion, right, @HumairAK? i.e. add component name to the IR spec to eliminate any guess work. I think that might be the ideal solution. Having an explicit map from the task name to the component name is an important piece of information to store in the IR. The only drawback that I can think of is that it theoretically increases the complexity of this PR, but if it's the desired target state, then the complexity is justified. @zazulam can you please mark this conversation as unresolved? Thanks!

HumairAK: >The only drawback that I can think of is that it theoretically increases the complexity of this PR, but if it's the desired target state, then the complexity is justified.

Yeah, this is why I proposed sticking with what we had, given that it does introduces a backwards incompatible change, but for a very niche case, and the fix is relatively straight forward. 

But, if there is concerns around the impact, going IR spec route is the only way I can think of that is least destructive. I suppose we could also preserve the dashes, I'm not super familiar with the reasoning behind why we didn't initially, so my concern there would be possible around unforeseen side effects.

droctothorpe: Thanks for confirming, @HumairAK.

To summarize the three proposed solutions:

1. Update [ExtractBaseComponentName](https://github.com/kubeflow/pipelines/pull/10979/files#diff-481aa8f29d58257fef3e48fb60fd320f73d50b858018fe6adf7ac583125fc2c6R267) to return the full component name if the IR includes a task with a name in the format of .*-\d+-\d+.
2. Preserve dash `-` characters in the IR instead of converting them into underscores `_`. That way we could use `-<index>` in `make_name_unique_by_adding_index` and not have to worry about collisions since python function names cannot include dashes. Sounds easy in theory but including dashes may violate expectations in the protobuf / backend.
3. Add an explicit `componentName` field to each component in the IR so that the backend doesn't have to deduce the "base name."

Do you have a preference between these three approaches, @chensun? Are there other options that we haven't taken into consideration?

droctothorpe:
Continuing the conversation from https://github.com/kubeflow/pipelines/pull/10979#discussion_r1698923786. We decided to run a bunch of different tests to confirm that the proposed edge cases are in fact legitimately edge cases.

We were surprised to discover that **none** of the proposed edge cases actually resulted in errors!

Here's relevant code:

## Test case 1
```python
@dsl.component()
def foo():
    print("foo")


@dsl.component()
def foo_2():
    print("foo_2")


@dsl.pipeline()
def pipeline():
    task1 = foo()
    task2 = foo()
    task3 = foo_2()
```

## Test case 2
```python
@dsl.component()
def foo_1():
    print("foo_1")


@dsl.pipeline()
def pipeline():
    task1 = foo_1()
```

**Both** of these succeeded without issue. **All** of the components executed appropriately and had the expected outputs.

**However**, we did discover one legitimate but _exceedingly_ rare edge case. Despite its unlikeliness, we updated the logic to account for it.

Specifically, if and only if you have two components with the same name except one of them ends in `_1`, e.g. `foo` and `foo_1`, the compiled AWF executes logic from `foo` for **both** tasks. This is only applicable when the suffix ends in the integer 1 specififcally. The other examples demonstrate that higher integers do not cause problems.

We addressed the edge case by making `ExtractBaseComponentName` return the full component name if and only if the component name ends in `_1`.

Ultimately, the root cause of this issue is because of how the IR names tasks in scenarios like this. This should probably eventually be addressed at the IR compilation level, but we elected to resolve it here because it's less disruptive and we want to expedite the fix for this regression. We can always resolve the IR issue in a later PR.

Here was the code that allowed us to finally trigger an actual edge case:

```python
@dsl.component()
def foo():
    print("foo")


@dsl.component()
def foo_1():
    print("foo_1")


@dsl.pipeline()
def pipeline():
    task1 = foo()
    task2 = foo()
    task3 = foo_1()
```

We removed all of the breaking changes to the client / SDK.

We think this is ready for final review. Please feel free to try to replicate the edge cases and let us know if you need any help with that.

cc @chensun @zijianjoy @HumairAK

backend/src/v2/compiler/argocompiler/argo.go
droctothorpe: Implemented and validated a solution. Continuing the thread [here](https://github.com/kubeflow/pipelines/pull/10979#issuecomment-2263795625) because this is too dang nested.

droctothorpe:
The e2e test failures are unrelated to this PR afaict.

hbelmiro:
> The e2e test failures are unrelated to this PR afaict.

Let's try again.
Since we migrated the tests to run on GitHub, we have no known flaky tests.

/rerun-all

hbelmiro:
`/rerun-all` is not working as expected.
@chensun @zijianjoy do you guys have permission to retrigger the `KFP e2e tests / integration-test` using the GitHub UI?

droctothorpe:
A few more potential edge cases just occurred to me. I'm going to test them as soon as my kids go to bed.

droctothorpe:
Argh we jumped the gun. It's not just `foo` + `foo_1`, it's `foo` + `foo_<any int>`. My fault, guys. At least we narrowed down the edge case. Fix TBD.

droctothorpe:
We fixed it! So it turns out that the IR actually stores the function name in the `deploymentSpec/executors/<executor-name>/container/args` field! So the suggestion that you provided [here](https://github.com/kubeflow/pipelines/pull/10979#discussion_r1672595012), @HumairAK, is already sort of indirectly implemented. Here's what the `args` field looks like for a given executor:

```yaml
...
deploymentSpec:
  executors:
    exec-foo-2:
      container:
        args:
        - --executor_input
        - '{{$}}'
        - --function_to_execute
        - foo
```

That function name is **exactly** what we need for deduplication! So we updated the logic to reference that instead, replacing `extractBaseName` with [extractFunctionName](https://github.com/kubeflow/pipelines/pull/10979/files#diff-481aa8f29d58257fef3e48fb60fd320f73d50b858018fe6adf7ac583125fc2c6R283) (this function is where the essence of the fix resides).

We think this is a much more reliable way to evaluate the underlying component function and it leans into data already available in the IR instead of trying to reverse engineer it.

We have thoroughly tested it against **all** of the edge cases described above and it succeeded in all of them.

There's just one catch: some of the really old unit test stubs are not working with this updated pattern. We think the stubs probably just need to be updated, but we'll have to tackle that tomorrow, bandwidth permitting.

droctothorpe:
Flakey unit tests fixed. We optimized the `extractFunctionName` logic a bit. If deduplication fails for any reason, the compiler just reverts to the standard (redundant) approach (while still migrating from annotations to wf spec arguments to circumvent the 256KB limit).

hbelmiro:
@droctothorpe FYI, I just confirmed that the e2e tests are failing :(
https://github.com/kubeflow/pipelines/issues/11069

droctothorpe:
Thanks, @hbelmiro.

Do you have any thoughts on how to drill down into the cause of failure? From the client's perspective, it's a generic `TimeoutError: Run timeout`, which isn't really actionable. The supplied GUI URL, `http://localhost:8888/#/experiments/details/c9b55db2-9ee8-4d49-b0a5-250fb5224618` doesn't resolve (for obvious reason). 

I downloaded the container logs that are made available at the end of `Collect test results` stage, i.e. --

```
Artifact download URL: https://github.com/kubeflow/pipelines/actions/runs/10216869671/artifacts/1769698572
```

and there are no relevant logs in the KFP backend pod or any other pod for that matter, and actually no pod logs for the submitted run!

```
ls -l
total 232
-rw-r--r--@ 1 alex  staff  21097 Aug  2 15:03 cache-deployer-deployment-cf9646b9c-jmp2m.log
-rw-r--r--@ 1 alex  staff    121 Aug  2 15:03 cache-server-56d4959c9-wbjg8.log
-rw-r--r--@ 1 alex  staff  21855 Aug  2 15:03 metadata-envoy-deployment-9c7db86d8-5k776.log
-rw-r--r--@ 1 alex  staff    161 Aug  2 15:03 metadata-grpc-deployment-d94cc8676-jhmck.log
-rw-r--r--@ 1 alex  staff  12289 Aug  2 15:03 metadata-writer-cd5dd8f7-5xpp5.log
-rw-r--r--@ 1 alex  staff    656 Aug  2 15:03 minio-5dc6ff5b96-h5b66.log
-rw-r--r--@ 1 alex  staff   1001 Aug  2 15:03 ml-pipeline-64d6db5897-qmxwl.log
-rw-r--r--@ 1 alex  staff    151 Aug  2 15:03 ml-pipeline-persistenceagent-77947c888d-b9k8g.log
-rw-r--r--@ 1 alex  staff    704 Aug  2 15:03 ml-pipeline-scheduledworkflow-676478b778-42m2r.log
-rw-r--r--@ 1 alex  staff   5935 Aug  2 15:03 ml-pipeline-ui-87b9d4fb6-f8fqq.log
-rw-r--r--@ 1 alex  staff    420 Aug  2 15:03 ml-pipeline-viewer-crd-8574556b89-gf9rz.log
-rw-r--r--@ 1 alex  staff      0 Aug  2 15:03 ml-pipeline-visualizationserver-5d7c54f495-gd2st.log
-rw-r--r--@ 1 alex  staff   5581 Aug  2 15:03 mysql-5b446b5744-2ptld.log
-rw-r--r--@ 1 alex  staff   4701 Aug  2 15:03 workflow-controller-679dcfdd4f-jhs28.log
```

/rerun-required

rimolive:
I believe it's because of this:
Screenshot from 2024-08-05 11-41-56

hbelmiro:
@droctothorpe @zazulam I have a PR open to fix the e2e tests: https://github.com/kubeflow/pipelines/pull/11072
Once it gets merged, please rebase your PR.

droctothorpe:
Thank you so much, @hbelmiro! Rebased from master and force pushed.

chensun (commented):
Overall LGTM.

backend/src/v2/compiler/argocompiler/argo.go
```diff
@@ -180,69 +184,127 @@ func (c *workflowCompiler) templateName(componentName string) string {
...
+	if c.wf.Spec.Arguments.GetParameterByName(componentName) == nil {
+		return "", fmt.Errorf("using component spec: failed to find workflow parameter %q", componentName)
+	}
+
+	return workflowParameter(componentName), nil
+}
+
+// extractFunctionName extracts the function name of a component by looking it
+// up in the pipeline spec.
+func (c *workflowCompiler) extractFunctionName(componentName string) string {
```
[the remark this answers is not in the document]
droctothorpe: So this PR tackles two separate problems:

1. This PR moves the component logic from the metadata annotations (which have a 256KB limit) to the spec (which has a much higher, configurable limit defined by the [etcd request size limit](https://etcd.io/docs/v3.5/dev-guide/limit/#:~:text=requests%20and%20storage-,Request%20size%20limit,bytes%20flag%20for%20etcd%20server.)). This in it of itself solves a major regression in v2.

2. In addition, this PR deduplicates component logic in the compiled AWF manifest. This duplication is a long-standing issue that was carried over from v1 and was reported as early as [2020](https://github.com/kubeflow/pipelines/issues/4170). Since the IR compiler appends an incrementing index to duplicate tasks, the underlying component logic gets duplicated in full in the the compiled AWF manifest for each instance. This is kind of like copying a function in full every time you want to invoke it. Since components can get quite large, with all this duplicate logic, it's easy to hit even the etcd requests size limit. Several of our internal model development teams have run into this exact problem and have been forced to partition large pipelines or minify compiled manifests (i.e. strip verbose annotations and comments), but minification only works in v1 where compilation happens client-side.

> Not every component has a visible function name. Why not just use the original name as-is?

If the `--function_to_execute` is not specified in the executor, then `extractFunctionName` just returns the original (indexed) component name, i.e. "the original name as-is." So that's precisely what happens when the function name is not specified.

At a bare minimum, this PR circumvents the 256KB limit, but it also substantially reduces the size of the compiled manifest (when possible) thanks to deduplication (we saw a 70% file size reduction in our tests).

Apologies if I'm over-explaining anything!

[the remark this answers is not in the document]
chensun: > If the --function_to_execute is not specified in the executor, then extractFunctionName just returns the original (indexed) component name, i.e. "the original name as-is." So that's precisely what happens when the function name is not specified.

Exactly as this function checks multiple cases, but doesn't guarantee any improvement over the original, so It feels to me an over-engineering that doesn't help much. I would suggest we remove it and use the original name as-is for simplicity, if the goal it to guarantee a short length and/or unique name, then we should use some hashing method.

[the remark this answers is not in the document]
droctothorpe: Returning the original name as is would guarantee redundancy in the compiled manifest. Hashing is an interesting approach! We could hash the `IR.deploymentSpec.executors.<executor>.container.command.` and use that as the key. One concern with that approach is that it significantly reduces the human readability of the AWF manifest. Also, could hashing hundreds of large components increase AWF compilation time?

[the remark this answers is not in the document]
droctothorpe: > Not every component has a visible function name. Why not just use the original name as-is?

QQ: Under what conditions does a component not include `--function_to_execute`?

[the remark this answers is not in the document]
chensun: For hashing, I was thinking hash the entire [ComponentSpec](https://github.com/kubeflow/pipelines/blob/e128bdbaf3b2071926e8a5c23041cc692329bb39/api/v2alpha1/pipeline_spec.proto#L86) (and possibly the expanded [ExecutorSpec](https://github.com/kubeflow/pipelines/blob/e128bdbaf3b2071926e8a5c23041cc692329bb39/api/v2alpha1/pipeline_spec.proto#L841) when applicable). That would guarantee uniqueness. I don't think it would add noticable AWF compilation time, but admit I didn't do any experiment.

> QQ: Under what conditions does a component not include --function_to_execute?

Exactly what I was planning to explain. :)
The issue with the current `extraFunctionName implementation` is that it only makes a difference for maybe 50%-60% of the time. IMO, it's not worth the added complexity and it may even hurts AWF readability--if we value that--one needs to look into this implementation to figure out where the name comes from. 

I can list a number of examples where the current approach fails to improve over original name as-is:
- condition, loop, and exit handler all produce DAG component with no executor. e.g.: https://github.com/kubeflow/pipelines/blob/e128bdbaf3b2071926e8a5c23041cc692329bb39/sdk/python/test_data/pipelines/pipeline_with_loops_and_conditions.yaml#L64-L84
- pipeline can be used as component, which is another DAG component without executor, e.g.: https://github.com/kubeflow/pipelines/blob/e128bdbaf3b2071926e8a5c23041cc692329bb39/sdk/python/test_data/pipelines/pipeline_in_pipeline_complex.yaml#L62-L80
- Custom container component doesn't have `--function_to_execute`, and may comes with very generic entry point like `main`, `run`, etc. And large enterprise may sometimes build a shared container with a single entrypoint that takes arguments to execute different component function. e.g: https://github.com/kubeflow/pipelines/blob/e128bdbaf3b2071926e8a5c23041cc692329bb39/sdk/python/test_data/components/container_io.yaml#L20-L28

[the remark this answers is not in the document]
chensun: So my suggestion for this PR is to simply remove the `extractFunctionName` part and use original component name as-is. I think the major concern on the size limit has been solved by moving the spec from annotation to argument. If further size reduction is needed, we can then look into a hashing solution that would work for 100%.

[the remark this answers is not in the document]
droctothorpe: Thank you so much for the thorough response and additional context, @chensun. We really appreciate it.

We POCed the hash-based approach that you suggested before we saw your comment and pushed it up since the work was done anyway. It works like a charm!

We tested against the three edge cases that you posted. Hashing successfully deduplicates all of the corresponding component logic. It's true that duplicate conditionals, DAGs, and nested pipelines are not deduplicated per say, but we confirmed that **their underlying shared components, which are what typically inflate the manifest file size beyond the 1.5 MB etcd limit, are in fact deduplicated.**

For example, [pipeline_in_pipeline_complex.yaml](https://github.com/kubeflow/pipelines/blob/e128bdbaf3b2071926e8a5c23041cc692329bb39/sdk/python/test_data/pipelines/pipeline_in_pipeline_complex.yaml#L62-L80) references `print_op1` three times: once at the root of `my_pipeline` and twice in the `graph_component` sub_pipeline (since it's parallelized).

We confirmed that the compiled AWF manifest consolidates these various tasks under the shared key, `implementations-83cf3e248270dfedd9adf4317961878cb8ebdf352ee8121750c35bfee5d408ae`, and therefore only stores the logic for the corresponding function once.

We think the change may be small enough in scope and lines of code and the underlying problem significant enough to warrant keeping it in this PR. That being said, we don't mind tackling the hash approach in a discrete PR.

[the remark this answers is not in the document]
droctothorpe: I modified the `component_io.yaml` example slightly to demonstrate deduplication (with hashing) for container components that don't list the function name:

```python
from kfp import Client
from kfp import dsl


@dsl.container_component
def container_io(arg: str, output_path: dsl.OutputPath(str)):
    return dsl.ContainerSpec(
        image="python:3.7",
        command=["echo", arg],
        args=["--output_path", output_path],
    )


@dsl.pipeline
def pipeline():
    task1 = container_io(arg="foo")
    task2 = container_io(arg="foo")
    task3 = container_io(arg="bar")


client = Client()
``` 

The resulting AWF manifest looks like this:

```yaml
Spec:
  Arguments:
    Parameters:
      Name:    components-fc838d49ffb7211dd93e608a1671731bc81e22197bbdb2e423df0ac72d451bc9
      Value:   {"executorLabel":"exec-container-io","inputDefinitions":{"parameters":{"arg":{"parameterType":"STRING"}}},"outputDefinitions":{"parameters":{"output_path":{"parameterType":"STRING"}}}}
      Name:    implementations-fc838d49ffb7211dd93e608a1671731bc81e22197bbdb2e423df0ac72d451bc9
      Value:   {"args":["--output_path","{{$.outputs.parameters['output_path'].output_file}}"],"command":["echo","{{$.inputs.parameters['arg']}}"],"image":"python:3.7"}
      Name:    components-root
      Value:   {"dag":{"tasks":{"container-io":{"cachingOptions":{"enableCache":true},"componentRef":{"name":"comp-container-io"},"inputs":{"parameters":{"arg":{"runtimeValue":{"constant":"foo"}}}},"taskInfo":{"name":"container-io"}},"container-io-2":{"cachingOptions":{"enableCache":true},"componentRef":{"name":"comp-container-io-2"},"inputs":{"parameters":{"arg":{"runtimeValue":{"constant":"foo"}}}},"taskInfo":{"name":"container-io-2"}},"container-io-3":{"cachingOptions":{"enableCache":true},"componentRef":{"name":"comp-container-io-3"},"inputs":{"parameters":{"arg":{"runtimeValue":{"constant":"bar"}}}},"taskInfo":{"name":"container-io-3"}}}}}
 ...

The tasks that share logic share the same hash in the `Spec.Arguments.Parameters` so the component logic is not duplicated.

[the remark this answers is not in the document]
Comment (droctothorpe) [reply backend/src/v2/compiler/argocompiler/argo.go]: With respect to the [pipelines_with_loops_and_conditions.yaml](https://github.com/kubeflow/pipelines/blob/e128bdbaf3b2071926e8a5c23041cc692329bb39/sdk/python/test_data/pipelines/pipeline_with_loops_and_conditions.yaml) edge case you mentioned: it has **10** instances of the `print_text` component.

Thanks to the deduplication (through hashing), it stores **just one copy** of the component logic!

```bash
go run backend/src/v2/cmd/compiler/main.go -spec sdk/python/test_data/pipelines/pipeline_with_loops_and_conditions.yaml | grep 'print_text(msg' | wc -l
       1
```

`print_text` is a small component (though there's plenty of fluff around even one line components in the form of pip installs and what not), but it's not uncommon to parallelize massive components; that 10x increase in storage consumption makes reaching the etcd `max-request-bytes` threshold inevitable for our more complex models.

backend/src/v2/compiler/argocompiler/argo.go
```diff
@@ -180,69 +186,131 @@ func (c *workflowCompiler) templateName(componentName string) string {
...
 	}
-	// Reference: https://argoproj.github.io/argo-workflows/variables/
-	return fmt.Sprintf("{{workflow.annotations.%s}}", name), nil
+
+	return workflowParameter(componentName), nil
+}
+
+// hashComponentCommandAndArgs combines and hashes the command and args fields of a
+// given component.
+func (c *workflowCompiler) hashComponentCommandAndArgs(componentName string) string {
```
chensun, line 273: This is very exciting improvement. Thank you, @droctothorpe !
While we are on the right track, a small change is needed to ensure we dedupe correctly. cmds and args are enough. At bare minimum, container image should be part of the hash as well. For completeness, if we only handle components with executor spec, can we hash the entire [PipelineContainerSpec](https://github.com/kubeflow/pipelines/blob/99bd2344011fdceb8f5c44afc361e32c2701b98d/api/v2alpha1/pipeline_spec.proto#L688)? WDYT?

backend/src/v2/compiler/argocompiler/argo.go
```diff
@@ -180,69 +186,131 @@ func (c *workflowCompiler) templateName(componentName string) string {
...
+	log.Debug("componentName: ", componentName)
+	// Return early for root component since it has no command and args.
+	if componentName == "root" {
+		return componentName
+	}
+	if c.executors != nil { // Don't bother if there are no executors in the pipeline spec.
```
chensun, line 279: nitpick: There's opportunity to dedupe DAG component as well. We can hash the entire ComponentSpec (plus ExecutorSpec when available) minus their naming part: component name, executor label, etc. 
We can leave it to future improvements, not a blocker for this PR.

backend/src/v2/compiler/argocompiler/argo.go
droctothorpe: Thank you, @chensun! And thank you for the excellent suggestion! You're absolutely right that hashing the entire `PipelineContainerSpec` makes more sense. We just pushed up a commit implementing that improved approach.

droctothorpe: https://github.com/kubeflow/pipelines/pull/10979/commits/a56df166fe091d2dff22a26d410f66d3fccf8c3e

chensun (approved):
/lgtm
/approve

Thanks you, @zazulam and @droctothorpe !

droctothorpe:
Thanks for the input, design review, and merge, @chensun! 🎉

HumairAK:
much needed change, great work folks! @droctothorpe ++ @zazulam ++

backend/src/v2/compiler/argocompiler/argo.go
HumairAK: @droctothorpe looks like we didn't go this far on this PR yes? would you be willing to open up a new github issue as a follow up, in case someone wants to pick this up?

PR #7725, opened 2022-05-13
feat(backend): isolate artifacts per namespace/profile/user using only one bucket

juliusvonkohout:
Fixes https://github.com/kubeflow/pipelines/issues/4649

Due to newly discovered functionality in argo we will not use multiple buckets but just multiple folders and the following iam policy. Furthermore e.g. AWS has a limit of 100 buckets so we should definitely stick with 1 bucket. This is a follow up of https://github.com/kubeflow/pipelines/pull/7406

By the way I am open to change the folder structure etc. as you prefer.

## One bucket multiple folder namespace isolation

1. change the workflow-controller configmap by adding workflow.namespace to `keyFormat: "private-artifacts/{{workflow.namespace}}/{{workflow.name}}/{{workflow.creationTimestamp.Y}}/{{workflow.creationTimestamp.m}}/{{workflow.creationTimestamp.d}}/{{pod.name}}"`
We do not need the per namespace artifactrepositories configmaps anymore, but users can still use them to change the default artifact storage defined above.

2. We still need to create a minio user in sync.py and and save the individual credentials to the per namespace secret. But we now need only one policy although this policy must still be attached to every user. So just some small changes in sync.py. We do not need to create additional buckets anymore
```
"Resource": [
    "arn:aws:s3:::%s/artifacts/*"  % shared_bucket_name, # old shared artifacts for backwards compatibility
    "arn:aws:s3:::%s/private-artifacts/${aws:username}/*"  % shared_bucket_name, # private artifacts
    "arn:aws:s3:::%s/private/${aws:username}/*"  % shared_bucket_name, # private storage
    "arn:aws:s3:::%s/shared/*"  % shared_bucket_name # shared storage for collaboration
]
```

We exclude `"arn:aws:s3:::%s/pipelines/*"  % shared_bucket_name, # shared pipelines ` because the the apiserver, not the user should manage pipeline  definitions on minio. Hopefully @zijianjoy  will remove them entirely from MinIO soon.
So changing the environment variable https://github.com/kubeflow/pipelines/blob/49cdb42c129dce653b542d4323d7087f67097e2e/backend/src/apiserver/client_manager.go#L43 and splitting  the storage into shared and private pipelines becomes unnecessary.

[...]

PR #10998, opened 2024-07-11
feat(deployment) Replace minio with seaweedfs as object store

pschoen-itsc:

Replace the old minio deployment with seaweedfs as object store. Seaweedfs is licensed under Apache 2.0 and supports ACLs inside the buckets, so you can give read / write permissions for paths only to specific users.

**Checklist:**


pschoen-itsc:
For now I just replaced to minio component in one of the overlays. Don't know what would be the best way to integrate this. New overlays for seaweedfs?
Also how should we design the service? For now I continue using minio-service, because it is used statically a lot in the project. But we can also replace every occurrence.

HumairAK:
This definitely looks like something we should bring up in the KFP meeting for a wider discussion.

juliusvonkohout:
- Lets try a staging are in https://github.com/kubeflow/manifests/tree/master/contrib
- Can you use Kustomize components there https://github.com/kubernetes-sigs/kustomize/blob/master/examples/components.md
- we need to be able add users at runtime in https://github.com/kubeflow/pipelines/blob/master/manifests/kustomize/base/installs/multi-user/pipelines-profile-controller/sync.py
- How can we implement something like https://github.com/kubeflow/pipelines/pull/7725 with seaweedfs (on-deman ACLs and new folders and credentials for users)
- lets use the same service and secret names as before. e.g. minio-secret
- performance numbers

juliusvonkohout:
Ah and I forgot
- simple horizontal scalability by increasing replicas
- can we scale with read-write-once PVCs as well?
- should it be a statefulset or a deployment?
- does it support something similar to minios deprecated gateway mode?

gregsheremeta:
Back in March, I did some exploration into a minio replacement for the KFP community. I came up with a rough list of requirements, and then charted out various alternatives against those requirements.

(Note: these requirements are not authoritative. I mostly came up with them myself as a way to compare alternatives)

Requirements, from a Kubeflow Pipelines perspective:

R0 Can storage and retrieve objects using S3 API
R1 Permissive license (Apache 2.0 is best, GPLv3 might be ok but would prefer to avoid, AGPL is not ok)
R2 Works in disconnected / on-prem environments (backed by local storage, e.g. PVCs)
R3 In cloud environments, can pass through to GCS or S3
R4 Can be automatically installed with a Kubeflow installer / no extra steps required
?? R5 Supports a developer mode where it is lightweight, low overhead (single pod is best).
Does this matter for production use cases? E.g. if I'm a large corporation storing petabytes worth of models, I want clustered object storage or possibly not even on prem.
R6 Doesn't require an operator / can be installed via simple manifests
?? Rn7 Shared server / zero-overhead namespaces
?? Rn8 Private namespace-scoped storage (e.g. global shared bucket with namespace-scoped subfolders)

This is the chart I came up with:
image

@pschoen-itsc would you be so kind as to give a quick look at these requirements and tell us how you think SeaweedFS does at fulfilling them? If we were to add SeaweedFS as a row 8 in that table, what would be in the columns? Thanks.

pschoen-itsc:
> Back in March, I did some exploration into a minio replacement for the KFP community. I came up with a rough list of requirements, and then charted out various alternatives against those requirements.
> 
> (Note: these requirements are not authoritative. I mostly came up with them myself as a way to compare alternatives)
> 
> Requirements, from a Kubeflow Pipelines perspective:
> 
> R0 Can store and retrieve objects using S3 API R1 Permissive license (Apache 2.0 is best, GPLv3 might be ok but would prefer to avoid, AGPL is not ok) R2 Works in disconnected / on-prem environments (backed by local storage, e.g. PVCs) R3 In cloud environments, can pass through to GCS or S3 R4 Can be automatically installed with a Kubeflow installer / no extra steps required ?? R5 Supports a developer mode where it is lightweight, low overhead (single pod is best). Does this matter for production use cases? E.g. if I'm a large corporation storing petabytes worth of models, I want clustered object storage or possibly not even on prem. R6 Doesn't require an operator / can be installed via simple manifests ?? Rn7 Shared server / zero-overhead namespaces ?? Rn8 Private namespace-scoped storage (e.g. global shared bucket with namespace-scoped subfolders)
> 
> This is the chart I came up with: image
> 
> @pschoen-itsc would you be so kind as to give a quick look at these requirements and tell us how you think SeaweedFS does at fulfilling them? If we were to add SeaweedFS as a row 8 in that table, what would be in the columns? Thanks.

R0: Yes (tested)
R1: Yes (Apache 2.0)
R2: Yes
R3: Yes (Documented [1](https://github.com/seaweedfs/seaweedfs/wiki/Gateway-to-Remote-Object-Storage) [2](https://github.com/seaweedfs/seaweedfs/wiki/Configure-Remote-Storage) )
R4: Yes
R5: Yes (Deployment with single container tested, scaling horizontally is documented)
R6: Yes (tested)
R7: Yes
R8: Yes (tested)

pschoen-itsc:
> * we need to be able add users and ACLs at runtime in https://github.com/kubeflow/pipelines/blob/master/manifests/kustomize/base/installs/multi-user/pipelines-profile-controller/sync.py

There is also a CLI to configure S3 access. SeaweedFs uses this in their own helm charts to set some settings after deployment. They create a Job which then connect to the running master / cluster and use the CLI  [1](https://github.com/seaweedfs/seaweedfs/blob/master/k8s/charts/seaweedfs/templates/post-install-bucket-hook.yaml) . So I think we could do something similar with the existing sync.py script.

> * How can we implement something like [feat(backend): isolate artifacts per namespace/profile/user using only one bucket #7725](https://github.com/kubeflow/pipelines/pull/7725) with seaweedfs (on-demand ACLs and new folders and different credentials per namespace per Seaweedfs rest API calls from sync.py)

I would say we generate the credentials in sync.py , then create the secret in the new namespace and create the identity in seaweedfs with the permissions:
- List:kubeflow
- Read:kubeflow/project1/*
- Write:kubeflow/project1/*

where "kubeflow" is the bucket name and "project1" the project name. Listing of objects is only possible to set globally, but as far as I know, this is just a limitation of S3.

> * lets use the same service and secret names as before. e.g. minio-secret

Service is no problem of course. And also secret should work, because when we use the CLI for dynamic user creation, seaweedfs does not have to read the secret. Changing credentials would then require manual intervention.

> * performance numbers

There are some in the official docs (could be biased) and also some more independent testing. It depends on how you set it up, just a single service with all components or you scale it horizontally with multiple filer / volume nodes. They provide helm charts for setting this up, so I could do my own tests if needed.

> * GUI with upload and download functionality (UI around their webdav interface) in the centraldashboard menu
> * Architectural proposal and diagrams

There are multiple options. Depends own desired performance / fault tolerance / resource usage. You have the 3 main services master, volume and filer which can be scaled independently or you can have everything in one container. You could even deploy separate gateways for S3 / WebDAV / etc. instead of let this run on the filer.
[Here](https://github.com/internetarchive/sandcrawler/blob/master/proposals/2020_seaweed_s3.md#testrun-4) they tested with just a single instance writing 100 millions of files with solid performance and resource usage. Downside in this setup would be that after a restart the inspection of the volumes and indexes takes minutes. With a more distributed setup I assume you could reduce this time or even keep the service available the whole time.

> * How active and reliable is the seaweedfs community? is there a license change going on, is there a commercial backer?
> * You need to sign the DCO
> * You should also present this in the KFP meeting

juliusvonkohout:
> I would say we generate the credentials in sync.py , then create the secret in the new namespace and create the identity in seaweedfs with the permissions:
> 
>     * List:kubeflow
> 
>     * Read:kubeflow/project1/*
> 
>     * Write:kubeflow/project1/*
> 
> 
> where "kubeflow" is the bucket name and "project1" the project name. Listing of objects is only possible to set globally, but as far as I know, this is just a limitation of S3.

Can you elaborate a bit more on the global Listing?
I am fine with dropping the global listing, if they can still list their namespace (aws:username) folder.
Can we achieve
```
                            "arn:aws:s3:::%s/artifacts/*"  % shared_bucket_name, # old shared artifacts for backwards compatibility
                            "arn:aws:s3:::%s/private-artifacts/${aws:username}/*"  % shared_bucket_name, # private artifacts
                            "arn:aws:s3:::%s/private/${aws:username}/*"  % shared_bucket_name, # private storage
                            "arn:aws:s3:::%s/shared/*"  % shared_bucket_name # shared storage for collaboration
```
with read/write/list access ?

> 
> > * lets use the same service and secret names as before. e.g. minio-secret
> 
> Service is no problem of course. And also secret should work, because when we use the CLI for dynamic user creation, seaweedfs does not have to read the secret. Changing credentials would then require manual intervention.

Who is then able to modify ACLs?
Can we keep an admin secret in the Kubeflow folder that is used in the sync.py script to change permissions? 


> > * GUI with upload and download functionality (UI around their webdav interface) in the centraldashboard menu
> > * Architectural proposal and diagrams
> 
> There are multiple options. Depends own desired performance / fault tolerance / resource usage. You have the 3 main services master, volume and filer which can be scaled independently or you can have everything in one container. You could even deploy separate gateways for S3 / WebDAV / etc. instead of let this run on the filer. [Here](https://github.com/internetarchive/sandcrawler/blob/master/proposals/2020_seaweed_s3.md#testrun-4) they tested with just a single instance writing 100 millions of files with solid performance and resource usage. Downside in this setup would be that after a restart the inspection of the volumes and indexes takes minutes. With a more distributed setup I assume you could reduce this time or even keep the service available the whole time.

Can we use a single statefulset that starts a pod with all three services and scale this statefulset or is scaling only possible with three statefulsets (master, volume, filer)?


Is 32 GB the file size limit for the time being?

pschoen-itsc:
> > I would say we generate the credentials in sync.py , then create the secret in the new namespace and create the identity in seaweedfs with the permissions:
> > ```
> > * List:kubeflow
> > 
> > * Read:kubeflow/project1/*
> > 
> > * Write:kubeflow/project1/*
> > ```
> > 
> > where "kubeflow" is the bucket name and "project1" the project name. Listing of objects is only possible to set globally, but as far as I know, this is just a limitation of S3.
> 
> Can you elaborate a bit more on the global Listing? I am fine with dropping the global listing, if they can still list their namespace (aws:username) folder. Can we achieve
> 
> ```
>                             "arn:aws:s3:::%s/artifacts/*"  % shared_bucket_name, # old shared artifacts for backwards compatibility
>                             "arn:aws:s3:::%s/private-artifacts/${aws:username}/*"  % shared_bucket_name, # private artifacts
>                             "arn:aws:s3:::%s/private/${aws:username}/*"  % shared_bucket_name, # private storage
>                             "arn:aws:s3:::%s/shared/*"  % shared_bucket_name # shared storage for collaboration
> ```
> 
> with read/write/list access ?

Because ListObjects is a bucket level operation it is not possible to control the permissions via the resource, like you can do it with put / read. In AWS S3 there seems to be a way to implement [this](https://repost.aws/de/knowledge-center/s3-folder-user-access) using IAM policies with conditions. But they also have to first give the user listing permissions for the whole bucket.

juliusvonkohout:
> > > I would say we generate the credentials in sync.py , then create the secret in the new namespace and create the identity in seaweedfs with the permissions:
> > > ```
> > > * List:kubeflow
> > > 
> > > * Read:kubeflow/project1/*
> > > 
> > > * Write:kubeflow/project1/*
> > > ```
> > > 
> > > 
> > >     
> > >       
> > >     
> > > 
> > >       
> > >     
> > > 
> > >     
> > >   
> > > where "kubeflow" is the bucket name and "project1" the project name. Listing of objects is only possible to set globally, but as far as I know, this is just a limitation of S3.
> > 
> > 
> > Can you elaborate a bit more on the global Listing? I am fine with dropping the global listing, if they can still list their namespace (aws:username) folder. Can we achieve
> > ```
> >                             "arn:aws:s3:::%s/artifacts/*"  % shared_bucket_name, # old shared artifacts for backwards compatibility
> >                             "arn:aws:s3:::%s/private-artifacts/${aws:username}/*"  % shared_bucket_name, # private artifacts
> >                             "arn:aws:s3:::%s/private/${aws:username}/*"  % shared_bucket_name, # private storage
> >                             "arn:aws:s3:::%s/shared/*"  % shared_bucket_name # shared storage for collaboration
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > with read/write/list access ?
> 
> Because ListObjects is a bucket level operation it is not possible to control the permissions via the resource, like you can do it with put / read. In AWS S3 there seems to be a way to implement [this](https://repost.aws/de/knowledge-center/s3-folder-user-access) using IAM policies with conditions. But they also have to first give the user listing permissions for the whole bucket.

Can we live without the list operation? is it enough to have read/write?

juliusvonkohout:
@pschoen-itsc  can you create a PR with customize components for kubeflow/manifests/contrib/seaweedfs ? Then we can start there with integration testing.

pschoen-itsc:
> @pschoen-itsc can you create a PR with customize components for kubeflow/manifests/contrib/seaweedfs ? Then we can start there with integration testing.

Yes, will do it tomorrow.

thesuperzapper:
@pschoen-itsc I also want to confirm that SeaweedFS would support IAM policies that restrict object access based on key prefixes?

For example, a profile might use an IAM policy that looks like this, which only allows them to read/write objects under a specific path:

```json
{
  "Version": "2012-10-17",
  "Statement": [
    {
      "Effect": "Allow",
      "Action": [
        "s3:GetBucketLocation",
        "s3:ListBucket"
      ],
      "Resource": [
        "arn:aws:s3:::<BUCKET_NAME>"
      ]
    },
    {
      "Effect": "Allow",
      "Action": [
        "s3:GetObject",
        "s3:PutObject",
        "s3:DeleteObject"
      ],
      "Resource": [
        "arn:aws:s3:::<BUCKET_NAME>/artifacts/<PROFILE_NAME>/*",
        "arn:aws:s3:::<BUCKET_NAME>/v2/artifacts/<PROFILE_NAME>/*"
      ]
    }
  ]
}
```

pschoen-itsc:
> @pschoen-itsc I also want to confirm that SeaweedFS would support IAM policies that restrict object access based on key prefixes?
> 
> For example, a profile might use an IAM policy that looks like this, which only allows them to read/write objects under a specific path:
> 
> ```json
> {
>   "Version": "2012-10-17",
>   "Statement": [
>     {
>       "Effect": "Allow",
>       "Action": [
>         "s3:GetBucketLocation",
>         "s3:ListBucket"
>       ],
>       "Resource": [
>         "arn:aws:s3:::<BUCKET_NAME>"
>       ]
>     },
>     {
>       "Effect": "Allow",
>       "Action": [
>         "s3:GetObject",
>         "s3:PutObject",
>         "s3:DeleteObject"
>       ],
>       "Resource": [
>         "arn:aws:s3:::<BUCKET_NAME>/artifacts/<PROFILE_NAME>/*",
>         "arn:aws:s3:::<BUCKET_NAME>/v2/artifacts/<PROFILE_NAME>/*"
>       ]
>     }
>   ]
> }
> ```

Object access based on key prefixes is supported. So you can set equivalent ACLs for the provided example. There is just no separation of PUT and DELETE object, just `Write`.

juliusvonkohout:
> Currently the list operation is used by KFP at least for retrieving input artifacts. I tested this for [kubeflow/manifests#2826](https://github.com/kubeflow/manifests/pull/2826). If you really just want to PUT and GET objects you don't need it. But that should then be in the implementation of argoexec.

Its fine for cluster.local/ns/kubeflow/sa/ml-pipeline and other admin level service account such as the Kubeflow UI. We just need to prevent that user workloads can access other users artifacts.

umka1332:
Any progress on this PR? Is slightly strange that kubeflow/manifests#2826 was merged, but this wasn't...

pschoen-itsc:
@umka1332 I am still working on integration of profiles with seaweedfs, which poses some unexpected challenges. Also I wanted to contribute that to kubeflow/manifests first to complete the feature there.

juliusvonkohout:
> @umka1332 I am still working on integration of profiles with seaweedfs, which poses some unexpected challenges. Also I wanted to contribute that to kubeflow/manifests first to complete the feature there.

Yes we need the ACLs and user management similar to https://github.com/kubeflow/pipelines/pull/7725 and some more cicd tests to verify that you cannot access other users artifacts, but your own. Adding this is PoC in manifests/contrib is easier and can then when it is done be merged here as well.

juliusvonkohout:
I will plan with the KFP maintainers to upstream https://github.com/kubeflow/manifests/pull/3051. Probably in a separate PR. Before I want to finish the follow up items here https://github.com/kubeflow/manifests/pull/3051#issuecomment-2833371838

juliusvonkohout:
https://github.com/kubeflow/pipelines/pull/11965/files

Issue #10036, opened 2023-09-26
[feature] display main.log as artifact for each step


Linchin:
### Feature Area
/area frontend
/area backend

### What feature would you like to see?
A feature similar to v1 behavior. In the details page of each step, the log appears as one of the artifacts that users can view:
![image](https://github.com/kubeflow/pipelines/assets/12806577/825829d8-f87f-4a02-9f4d-fe8f0c65d561)

In v2, this main log artifact is no longer displayed. It would be great if we could add a similar section to show the logs, or use the log artifact as a source for the "logs" panel after the pod is deleted, instead of directly pulling from the (already deleted) pod.

### What is the use case or pain point?

Because all the logs of completed pods are auto deleted by Kubernetes after 24 hours, the users can no longer access the logs of earlier pipeline runs. 

### Is there a workaround currently?

No workaround.

/cc @kromanow94

---

Love this idea? Give it a 👍.

revolutionisme:
I read that this issue was not included in the previous release as mentioned in this [comment](https://github.com/kubeflow/manifests/issues/2442#issuecomment-1788066478) , just wanted to ask whether we have an update on this or not.

AnnKatrinBecker:
/reopen

Is readding this feature to v2 planned for future releases? It would be very helpful.

HumairAK:
I'm a bit unfamiliar with the v1 version of this, did this come from argo's archive logging feature? 

In v2 kfp aims to be agnostic to the backing engine, so if this is an argo specific feature, we may need to consider a separate feature/proposal on how to handle persistent logging in general for kfp.

TobiasGoerke:
> I'm a bit unfamiliar with the v1 version of this, did this come from argo's archive logging feature?
> 
> In v2 kfp aims to be agnostic to the backing engine, so if this is an argo specific feature, we may need to consider a separate feature/proposal on how to handle persistent logging in general for kfp.

Yes, Argo Workflows stores the logs separately and given only kfp metadata, there currently seems to be no way to obtain the logs' locations. You'd need to access the Argo Workflow object directly to obtain the info.

Including this feature right into kfp would make sense and imo is much needed, as otherwise, users aren't able to view logs after the step's pod has been deleted.

sanchesoon:
Hi @HumairAK What do you think about this ? 
Points:
- DS engineer would be interesting to see logs as artifact, as part of models producer investigation
- logs can have different retention against artifacts: user can clean them to reduce space usage
[...]

PR #11010, opened 2024-07-16
fix(frontend): retrieve archived logs from correct location

droctothorpe:
This PR solves https://github.com/kubeflow/pipelines/issues/10036 by making it possible to inject the AWF `keyFormat`, which determines where archived workflow container logs are saved, into the UI via `configs.ts` / env vars so that the UI retrieves them from the correct location.

Currently, for v2, the UI is looking in the wrong object store path (which is hard-coded) and coming up empty. We could update the hard-coded path, but loading it from `configs.ts` makes it more future proof and flexible in case end users want to modify the `keyFormat` for any reason.

This is an interim solution that's still a huge improvement over the current state of affairs (logs are inaccessible in v2 once the AWF CR has been GCed), but it has two important drawbacks that we want to delineate:

1. It doesn't show logs for the dag driver or the task driver(s). This can be helpful if there are issues in the driver rather than the component code. The only way to debug this currently is to interact directly with the pods via the K8s API. This PR doesn't introduce this limitation. It just doesn't solve for it.
2. If the AWF `keyFormat` is updated in both the `workflow-controller-configmap` and the UI (via `configs.ts` / env vars), the logs of new runs will be accessible, but the logs of old runs will no longer be accessible. In defense of this PR as an interim solution, the logs of NO runs are currently accessible.

In spite of these drawbacks, we still think it may be worthwhile to merge this PR as an interim fix since `keyFormat` is rarely modified and this will solve 99% of use cases for a really significant regression (with, at present, 21 👍 emojis). 

Looking forward to feedback. Thanks!

PS. We can add support for additional AWF template tags, as long as the actual parameters are available to the handler. We can also list supported tags via comments in the configmap.

PPS. I am not a frontend engineer.

**Checklist:**

### DIFF
--- a/frontend/server/configs.ts
+++ b/frontend/server/configs.ts
@@ -91,7 +91,10 @@ export function loadConfigs(argv: string[], env: ProcessEnv): UIConfigs {
     ARGO_ARCHIVE_ARTIFACTORY = 'minio',
     /** Bucket to retrive logs from */
     ARGO_ARCHIVE_BUCKETNAME = 'mlpipeline',
-    /** This should match the keyFormat specified in the Argo workflow-controller-configmap */
+    /** This should match the keyFormat specified in the Argo workflow-controller-configmap.
+     * It's set here in the manifests:
+     * https://github.com/kubeflow/pipelines/blob/7b7918ebf8c30e6ceec99283ef20dbc02fdf6a42/manifests/kustomize/third-party/argo/base/workflow-controller-configmap-patch.yaml#L28
+     */
     ARGO_KEYFORMAT = 'artifacts/{{workflow.name}}/{{workflow.creationTimestamp.Y}}/{{workflow.creationTimestamp.m}}/{{workflow.creationTimestamp.d}}/{{pod.name}}',
     /** Should use server API for log streaming? */
     STREAM_LOGS_FROM_SERVER_API = 'false',
--- a/frontend/server/configs.ts
+++ b/frontend/server/configs.ts
@@ -91,8 +91,8 @@ export function loadConfigs(argv: string[], env: ProcessEnv): UIConfigs {
     ARGO_ARCHIVE_ARTIFACTORY = 'minio',
     /** Bucket to retrive logs from */
     ARGO_ARCHIVE_BUCKETNAME = 'mlpipeline',
-    /** Prefix to logs. */
-    ARGO_ARCHIVE_PREFIX = 'logs',
+    /** This should match the keyFormat specified in the Argo workflow-controller-configmap */
+    ARGO_KEYFORMAT = 'artifacts/{{workflow.name}}/{{workflow.creationTimestamp.Y}}/{{workflow.creationTimestamp.m}}/{{workflow.creationTimestamp.d}}/{{pod.name}}',
     /** Should use server API for log streaming? */
     STREAM_LOGS_FROM_SERVER_API = 'false',
     /** The main container name of a pod where logs are retrieved */
@@ -127,7 +127,7 @@ export function loadConfigs(argv: string[], env: ProcessEnv): UIConfigs {
       archiveArtifactory: ARGO_ARCHIVE_ARTIFACTORY,
       archiveBucketName: ARGO_ARCHIVE_BUCKETNAME,
       archiveLogs: asBool(ARGO_ARCHIVE_LOGS),
-      archivePrefix: ARGO_ARCHIVE_PREFIX,
+      keyFormat: ARGO_KEYFORMAT,
     },
     pod: {
       logContainerName: POD_LOG_CONTAINER_NAME,
@@ -253,7 +253,7 @@ export interface ArgoConfigs {
   archiveLogs: boolean;
   archiveArtifactory: string;
   archiveBucketName: string;
-  archivePrefix: string;
+  keyFormat: string;
 }
 export interface ServerConfigs {
   basePath: string;
--- a/frontend/server/workflow-helper.ts
+++ b/frontend/server/workflow-helper.ts
@@ -61,15 +61,15 @@ export interface SecretSelector {
  * fails.
  */
 export function composePodLogsStreamHandler<T = Stream>(
-  handler: (podName: string, namespace?: string) => Promise<T>,
-  fallback?: (podName: string, namespace?: string) => Promise<T>,
+  handler: (podName: string, createdAt: string, namespace?: string) => Promise<T>,
+  fallback?: (podName: string, createdAt: string, namespace?: string) => Promise<T>,
 ) {
-  return async (podName: string, namespace?: string) => {
+  return async (podName: string, createdAt: string, namespace?: string) => {
     try {
-      return await handler(podName, namespace);
+      return await handler(podName, createdAt, namespace);
     } catch (err) {
       if (fallback) {
-        return await fallback(podName, namespace);
+        return await fallback(podName, createdAt, namespace);
       }
       console.warn(err);
       throw err;
@@ -85,6 +85,7 @@ export function composePodLogsStreamHandler<T = Stream>(
  */
 export async function getPodLogsStreamFromK8s(
   podName: string,
+  createdAt: string,
   namespace?: string,
   containerName: string = 'main',
 ) {
@@ -112,10 +113,10 @@ export const getPodLogsStreamFromWorkflow = toGetPodLogsStream(
  * on the provided pod name and namespace (optional).
  */
 export function toGetPodLogsStream(
-  getMinioRequestConfig: (podName: string, namespace?: string) => Promise<MinioRequestConfig>,
+  getMinioRequestConfig: (podName: string, createdAt: string, namespace?: string) => Promise<MinioRequestConfig>,
 ) {
-  return async (podName: string, namespace?: string) => {
-    const request = await getMinioRequestConfig(podName, namespace);
+  return async (podName: string, createdAt: string, namespace?: string) => {
+    const request = await getMinioRequestConfig(podName, createdAt, namespace);
     console.log(`Getting logs for pod:${podName} from ${request.bucket}/${request.key}.`);
     return await getObjectStream(request);
   };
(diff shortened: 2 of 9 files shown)


thesuperzapper:
@droctothorpe if we are going to do this interim fix can we please change the implementation slightly:

1. Use the go template package (which is what Argo uses) rather than native string replacement, so we:
    - allow for key formats that use non perfect spaces `{{ workflow.name}}`, which is still valid
    - fail to start KFP if an unsupported key format template is passed (e.g. using an unsupported variable/function)
2. Also support `{{ workflow.namespace }}` because this is a very common element of keyFormat to allow for namespace isolation by IAM scoped to object store prefixes

droctothorpe:
We added namespace tag handling (we'll add proper tests post design review) and confirmed that it works as expected, which is a good step towards improved user isolation for log artifacts.

We also added template tag validation. The node server throws an error which bubbles up to the UI and provides clear remediation instructions.

thesuperzapper:
@droctothorpe can you try and make it support arbitrary spaces within the `{{ xxx }}`, probably using a regex like `{{\s*some_token\s*}}`

It's very common to format go templates like`{{ xxxxx }}` and not like `{{xxxx}}`, and it would be helpful if users can use the exact same value as their Argo Workflows `keyFormat` config.

For example, these should all be equivalent:

- `{{aaaa}}`
- `{{ aaaa}}`
- `{{aaaa }}`
- `{{ aaaa }}`
- `{{         aaaa    }}`

thesuperzapper:
@droctothorpe also, just to be clear, does this affect KFP V1 pipelines?

If so, we should support separate `keyFormat` configs for KFP V1 vs V2, because many deployments store their V1 and V2 pipeline logs separately.

droctothorpe:
@thesuperzapper addressed your penultimate comment with `.replace(/\s+/g, '')` since the `keyFormat` should not contain any whitespace anyway.

To your second point, my understanding is that `keyFormat` is defined at the AWF level rather than the KFP level and is set globally for both v1 and v2. This PR just instructs the UI where to look for the logs. The UI has no say over where the logs are actually being saved.

Apologies for the delayed response. Had to pivot to some dask-kubernetes / xgboost issues for two weeks.

thesuperzapper:
@droctothorpe why not just use the approach I described in https://github.com/kubeflow/pipelines/pull/11010#issuecomment-2232342785?

While I agree removing spaces might seem like a solution, if keyFormat ever adds functions (beyond the default ones with every go template) removing spaces will break the inputs to those functions.

Also, while it would be a little crazy, users might want to include spaces in their S3 object store names.

droctothorpe:
> @droctothorpe why not just use the approach I described in [#11010 (comment)](https://github.com/kubeflow/pipelines/pull/11010#issuecomment-2232342785)?
> 
> While I agree removing spaces might seem like a solution, if keyFormat ever adds functions (beyond the default ones with every go template) removing spaces will break the inputs to those functions.
> 
> Also, while it would be a little crazy, users might want to include spaces in their S3 object store names.

This was my rationale:

> Bucket names can consist only of lowercase letters, numbers, dots (.), and hyphens (-).

https://docs.aws.amazon.com/AmazonS3/latest/userguide/bucketnamingrules.html

> Bucket names can only contain lowercase letters, numeric characters, dashes (-), underscores (_), and dots (.). Spaces are not allowed.

https://cloud.google.com/storage/docs/buckets#:~:text=Your%20bucket%20names%20must%20meet,Spaces%20are%20not%20allowed.

https://github.com/minio/minio-go/blob/cc7769f319debdee18b5b304e2b35166b0165fdb/pkg/s3utils/utils.go#L346

frontend/server/configs.ts
```diff
@@ -91,8 +91,8 @@ export function loadConfigs(argv: string[], env: ProcessEnv): UIConfigs {
     ARGO_ARCHIVE_ARTIFACTORY = 'minio',
     /** Bucket to retrive logs from */
     ARGO_ARCHIVE_BUCKETNAME = 'mlpipeline',
-    /** Prefix to logs. */
-    ARGO_ARCHIVE_PREFIX = 'logs',
+    /** This should match the keyFormat specified in the Argo workflow-controller-configmap */
+    ARGO_KEYFORMAT = 'artifacts/{{workflow.name}}/{{workflow.creationTimestamp.Y}}/{{workflow.creationTimestamp.m}}/{{workflow.creationTimestamp.d}}/{{pod.name}}',
```
zijianjoy, line 95: It will be great to add a link to the source of this log path in the comment. Whenever there is a location change in the future, we can still find out by looking at the source of truth.

frontend/server/workflow-helper.ts
```diff
@@ -132,19 +133,48 @@ export function toGetPodLogsStream(
...
+      .replace("{{workflow.creationTimestamp.Y}}", createdAtArray[0])
+      .replace("{{workflow.creationTimestamp.m}}", createdAtArray[1])
+      .replace("{{workflow.creationTimestamp.d}}", createdAtArray[2])
+      .replace("{{pod.name}}", podName)
+      .replace("{{workflow.namespace}}", namespace)
+
+    if (!key.endsWith("/")) {
+      key = key + "/"
+    }
+    key = key + "main.log"
```
zijianjoy, line 155: So the `main.log` is not part of the keyFormat, but it is appended here? Maybe just adding the `main.log` to the keyFormat constant?

zijianjoy (commented):
NIT: Run `npm run format` to pass presubmit test.

frontend/server/workflow-helper.ts
droctothorpe: Thank you so much for the feedback, @zijianjoy! 🙏

My rationale for excluding `main.log` from `keyFormat` was so that the `keyFormat` value there could precisely mirror the `keyFormat` value specified by the backend, which is set (by default) [here](https://github.com/kubeflow/pipelines/blob/7b7918ebf8c30e6ceec99283ef20dbc02fdf6a42/manifests/gcp_marketplace/chart/kubeflow-pipelines/templates/argo.yaml#L864) (internally) and [here](https://github.com/argoproj/argo-workflows/blob/main/docs/workflow-controller-configmap.yaml#L122-L127) (externally), and doesn't include `main.log`, i.e. the container name. I'm concerned that introducing a delta might confuse end users.

It looks like the `main` container name is hardcoded [here](https://github.com/kubeflow/pipelines/blob/7b7918ebf8c30e6ceec99283ef20dbc02fdf6a42/backend/src/v2/driver/driver.go#L458). Since it's hard-coded in the backend, do you still want it to be parameterized it in `config.ts`? If so, would you prefer making the user container name a dedicated parameter in `config.ts` to avoid the aforementioned `keyFormat` delta? Happy to implement either way.

frontend/server/configs.ts
droctothorpe: Copy that! Do you prefer referencing it by:

1. Master branch link:
https://github.com/kubeflow/pipelines/blob/master/manifests/kustomize/third-party/argo/base/workflow-controller-configmap-patch.yaml#L28
3. Permalink: https://github.com/kubeflow/pipelines/blob/7b7918ebf8c30e6ceec99283ef20dbc02fdf6a42/manifests/kustomize/third-party/argo/base/workflow-controller-configmap-patch.yaml#L28
4. The file path relative to `config.ts`:
../../manifests/kustomize/third-party/argo/base/workflow-controller-configmap-patch.yaml

droctothorpe:
I'm going to do some manual validation against v1 to make sure that this didn't introduce any breaking changes.

frontend/server/configs.ts
zijianjoy: Thanks for the question! I am fine with your preference for any of that above, I would `link to the premalink` for the stability of content we are referring.

frontend/server/workflow-helper.ts
zijianjoy: In that case, it is okay not to change the current keyFormat. We can think more about making the log name configurable, but that will be a different task. Thanks for sharing your thinking process!

droctothorpe:
# Manual validation evidence

## KFP v2

The run code:

```python
from kfp import dsl
from kfp.compiler.compiler import Compiler
from kfp import Client


@dsl.component()
def foo():
    from random import random

    print(random())


@dsl.pipeline()
def pipeline():
    task1 = foo()
    task1.set_caching_options(False)


# Compiler().compile(pipeline_func=pipeline, package_path="two.yaml")
client = Client()
```

Logs in the UI:
image

Deleting the workflow:

```bash
> k delete workflows --all && k get workflows
workflow.argoproj.io "pipeline-fzhgs" deleted
No resources found in kubeflow namespace.
```

Logs still available in the UI:
image

## KFP v1

The run code:

```python
from kfp import Client
from kfp import dsl

# from kfp.compiler.compiler import Compiler
from kfp.components import create_component_from_func


def foo():
    from random import random

    print(random())


foo_comp = create_component_from_func(foo)


@dsl.pipeline()
def pipeline():
    task1 = foo_comp()
    task1.set_caching_options(False)


# Compiler().compile(pipeline_func=pipeline, package_path="two.yaml")
client = Client()
client.create_run_from_pipeline_func(pipeline, arguments={})
```

Logs in the UI:
image

Deleting the workflow:

```bash
> k delete workflows --all && k get workflows
workflow.argoproj.io "pipeline-vkq4t" deleted
No resources found in kubeflow namespace.
```

Logs are NOT available in the UI from the logs tab. They must be accessed via output artifacts. This is a known bug that we could probably fix pretty easily, though I recommend doing so in a discrete PR since this PR doesn't introduce this bug, it just doesn't fix it.

## A new regression

I decided to run another manual validation and found an unrelated regression.

In the **master** branch, if you create a run then delete the pod but **preserve** the manifest, it still fails to retrieve logs!

image

Unless I'm somehow misconfiguring something, I think it's parsing the AWF manifest incorrectly. All of the workflow parsing logic [here](https://github.com/kubeflow/pipelines/blob/master/frontend/server/workflow-helper.ts#L160-L191) seems to be deeply misaligned with the actual workflow structure.

I started working on a fix but it was quite complicated and seemed like scope creep so I figured it might make more sense to tackle it in a standalone PR. That being said, the fact that retrieving artifact keys from the workflow doesn't work arguably increases the urgency of this PR we always need to fallback to the final handler now when the pod is GCed.

droctothorpe:
Screw it. @quinnovator 🧙 and I decided to just fix the workflow parsing as part of this PR. Confirmed it works. May have to update tests. Additional commits coming soon.

droctothorpe:
@zijianjoy, @quinnovator and I decided to just fix `getPodLogsStreamFromWorkflow` in this PR. It broke as a result of changes to the AWF manifest structure that were probably introduced a while ago. We updated the interfaces and brought the logic up to speed.

To summarize, the frontend tries to get logs from the following three locations in order:

1. Directly from the K8s API.
2. Extracting the artifact path from the AWF manifest (if the pods are GCed but the AWF manifest is not).
3. Inferring the artifact path directly (if the AWF manifest is GCed).

**Both 2 and 3 were broken! This PR fixes both.** 🎉

zijianjoy:
/lgtm
/approve

Thank you @droctothorpe for the contribution and validation!

Issue #10914, opened 2024-06-17
[kubernetes_platform] Error calling kubernetes.use_secret_as_env and use_secret_as_volume with secret name as parameter


revit13:
We get the following error running the program from the [example ](https://github.com/kubeflow/pipelines/tree/master/kubernetes_platform/python#secret-as-environment-variable) with the change of passing the secret name as a parameter:
```
e-packages/kfp/kubernetes/secret.py", line 48, in use_secret_as_env
    secret_as_env = pb.SecretAsEnv(
TypeError: bad argument type for built-in operation
```

### Environment

*  KFP version: 
2.2.0

### Steps to reproduce

run the following example:

```bash
from kfp import dsl
from kfp import kubernetes

@dsl.component(base_image="python:3.10")
def print_secret():
    import os
    print(os.environ['my-secret'])

@dsl.pipeline
def pipeline(secret_name: str = 'my-secret'):
    task = print_secret()
    kubernetes.use_secret_as_env(task,
                                 secret_name=secret_name,
                                 secret_key_to_env={'password': 'SECRET_VAR'})
```

It also happens when running the following program:

```bash
from kfp import dsl
from kfp import kubernetes

@dsl.component(base_image="python:3.10")
def print_secret():
    with open('/mnt/my_vol') as f:
        print(f.read())

@dsl.pipeline
def pipeline(secret_name: str = 'my-secret'):
    task = print_secret()
    kubernetes.use_secret_as_volume(task,
                                    secret_name=secret_name,
                                    mount_path='/mnt/my_vol')
```

### Expected result

not produce an error

### Materials and Reference

---

Impacted by this bug? Give it a 👍.

roytman:
As a workaround, we add hard-coded names of CM/Secrets in pipelines, so moving to a new CM/Secret requires updating/compilation and deployment of a pipeline.

DharmitD:
@diegolovison and I tested out the first example with kfp sdk version 2.2.0 and kfp-kubernetes version 1.0.0. We noticed  that the environment variable setting is incorrect in the example, find details [here.](https://github.com/kubeflow/pipelines/pull/11013#pullrequestreview-2181291388) With that correct, we were able to successfully run the first example and spin up a Pipeline and a run.

Second scenario is the one that is failing, this is the error observed on attempting to complle the second example:
[...]

PR #10435, opened 2024-01-26
feat(sdk): support f-strings in local pipeline execution

connor-mccarthy:
**Description of your changes:**
Support f-strings in local pipeline execution.

```python
from kfp import dsl
from kfp import local

local.init(runner=local.SubprocessRunner())

@dsl.component
def identity(string: str) -> str:
    return string

@dsl.pipeline
def my_pipeline(string: str = 'baz') -> str:
    op1 = identity(string=f'bar-{string}')
    op2 = identity(string=f'foo-{op1.output}')
    return op2.output

task = my_pipeline()
assert task.output == 'foo-bar-baz'
```

**Checklist:**
- [x] The title for your pull request (PR) should follow our title convention. [Learn more about the pull request title convention used in this repository](https://github.com/kubeflow/pipelines/blob/master/CONTRIBUTING.md#pull-request-title-convention).

PR #5183, opened 2021-02-24
feat(sdk.v2): Parse pipeline param in argument values

chensun:
**Description of your changes:**
To support string concatenation in pipeline params, for instance:
```
@dsl.pipeline(name='pipeline-with-pipelineparam-containing-format')
def my_pipeline(name: str = 'KFP'):
  print_op('Hello {}'.format(name))
```

**Checklist:**
- [x] The title for your pull request (PR) should follow our title convention. [Learn more about the pull request title convention used in this repository](https://github.com/kubeflow/pipelines/blob/master/CONTRIBUTING.md#pull-request-title-convention). 

- [ ] Do you want this pull request (PR) cherry-picked into the current release branch?

    [Learn more about cherry-picking updates into the release branch](https://github.com/kubeflow/pipelines/blob/master/RELEASE.md#cherry-picking-pull-requests-to-release-branch).

PR #11039, opened 2024-07-25
WIP: fix(Backend+SDK): Update kubernetes.use_secret_as_volume to accept secret name dynamically at runtime

DharmitD:
Resolves https://github.com/kubeflow/pipelines/issues/10914 

Attempting to pass the `secret_name` parameter in the dsl compile this way:

```
from kfp import dsl
from kfp import kubernetes

@dsl.component(base_image="python:3.10")
def print_secret():
    with open('/mnt/my_vol') as f:
        print(f.read())

@dsl.pipeline
def pipeline(secret_name: str = 'my-secret'):
    task = print_secret()
    kubernetes.use_secret_as_volume(task,
                                    secret_name=secret_name,
                                    mount_path='/mnt/my_vol')
```

results in this error:
```
File "/home/ddalvi/PycharmProjects/pythonProject2/main.py", line 23, in pipeline
    kubernetes.use_secret_as_volume(task,
  File "/home/ddalvi/miniconda3/envs/kfp_pipe_secret_env/lib/python3.11/site-packages/kfp/kubernetes/secret.py", line 81, in use_secret_as_volume
    secret_as_vol = pb.SecretAsVolume(
                    ^^^^^^^^^^^^^^^^^^
TypeError: bad argument type for built-in operation
```

The secret_name parameter was originally expected to be a string (str). However, in this case, it was being provided as a an object of the `PipelineParameterChannel` class , leading to a type mismatch error when the code tried to use it as a string.
Updated `secret_name` to accept either a `str` or `PipelineParameterChannel` using `Union[str, PipelineParameterChannel]` and added a check to convert `PipelineParameterChannel` to its string representation.

Furthermore,
- Implemented support for dynamically specifying secret names in the `use_secret_as_volume` function.
- Updated driver code to handle secret name substitution at runtime based on input parameters.
- Introduced a `{{secret_name}}` template string representation for secret names in the compiled DSL.
- Added a test to validate secret name template creation in IR.

Refer to [this comment](https://github.com/kubeflow/pipelines/issues/10914#issuecomment-2285049104) to understand why the driver code change is required.

Please refer to a [PoC](https://github.com/opendatahub-io/data-science-pipelines/pull/71) wherein we've listed out different test cases - four different pairs of DSL compile code and resulting pipeline yaml IRs with which we tested this update.

Collaborated with @gregsheremeta for this contribution, it was a joint effort. Thanks Greg for your contributions! :)

**Checklist:**

### DIFF
--- a/backend/src/v2/driver/driver.go
+++ b/backend/src/v2/driver/driver.go
@@ -17,9 +17,11 @@ import (
 	"context"
 	"encoding/json"
 	"fmt"
-	"github.com/kubeflow/pipelines/backend/src/v2/objectstore"
 	"strconv"
+	"strings"
 	"time"
+
+	"github.com/kubeflow/pipelines/backend/src/v2/objectstore"
 
 	"github.com/golang/glog"
 	"github.com/golang/protobuf/ptypes/timestamp"
@@ -536,14 +538,40 @@ func extendPodSpecPatch(
 	// Get secret mount information
 	for _, secretAsVolume := range kubernetesExecutorConfig.GetSecretAsVolume() {
 		optional := secretAsVolume.Optional != nil && *secretAsVolume.Optional
+
+		secretName := secretAsVolume.GetSecretName()
+
+		if strings.HasPrefix(secretName, "{{") && strings.HasSuffix(secretName, "}}") {
+			// Strip the braces
+			key := secretName[2 : len(secretName)-2]
+
+			// Check if the key exists in the parameter inputs map
+			inputParams, _, err := dag.Execution.GetParameters()
+			if err != nil {
+				return fmt.Errorf("failed to get input parameters: %v", err)
+			}
+
+			val, ok := inputParams[key]
+			if !ok {
+				return fmt.Errorf("dynamic secret name key '%s' not found in input parameters", key)
+			}
+
+			secretName = val.GetStringValue()
+			if secretName == "" {
+				return fmt.Errorf("secret name for key '%s' is empty", key)
+			}
+		} else if strings.TrimSpace(secretName) == "" {
+			return fmt.Errorf("secret name is empty or invalid")
+		}
+
 		secretVolume := k8score.Volume{
-			Name: secretAsVolume.GetSecretName(),
+			Name: secretName,
 			VolumeSource: k8score.VolumeSource{
-				Secret: &k8score.SecretVolumeSource{SecretName: secretAsVolume.GetSecretName(), Optional: &optional},
+				Secret: &k8score.SecretVolumeSource{SecretName: secretName, Optional: &optional},
 			},
 		}
 		secretVolumeMount := k8score.VolumeMount{
-			Name:      secretAsVolume.GetSecretName(),
+			Name:      secretName,
 			MountPath: secretAsVolume.GetMountPath(),
 		}
 		podSpec.Volumes = append(podSpec.Volumes, secretVolume)
--- a/kubernetes_platform/python/kfp/kubernetes/secret.py
+++ b/kubernetes_platform/python/kfp/kubernetes/secret.py
@@ -18,7 +18,8 @@ from google.protobuf import json_format
 from kfp.dsl import PipelineTask
 from kfp.kubernetes import common
 from kfp.kubernetes import kubernetes_executor_config_pb2 as pb
-
+from typing import Union
+from kfp.dsl.pipeline_channel import PipelineParameterChannel
 
 def use_secret_as_env(
     task: PipelineTask,
@@ -59,7 +60,7 @@ def use_secret_as_env(
 
 def use_secret_as_volume(
     task: PipelineTask,
-    secret_name: str,
+    secret_name: Union[str, PipelineParameterChannel],
     mount_path: str,
     optional: bool = False,
 ) -> PipelineTask:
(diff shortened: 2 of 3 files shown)


gregsheremeta:
/lgtm

I also verified it, so
/verified
:)

@chensun mind taking a quick look?

amadhusu (approved):

diegolovison (changes requested):
Please create a test for this scenario

gregsheremeta:
> Please create a test for this scenario

good call

kubernetes_platform/python/test/unit/test_secret.py
```diff
@@ -147,6 +147,34 @@ def my_pipeline():
             }
         }
 
+    def test_with_secret_name_param(self):
+        @dsl.pipeline
+        def my_pipeline(secret_name: str = 'my-secret'):
```
diegolovison, line 152: It is missing a test that assert from the value `my-secret`

DharmitD: The problem here is the pipeline spec generated from this compile step has the name stored in as a variable under the `roots` section.
```
root:
  dag:
    tasks:
      print-secret:
        cachingOptions:
          enableCache: true
        componentRef:
          name: comp-print-secret
        taskInfo:
          name: print-secret
  inputDefinitions:
    parameters:
      secret_name:
        defaultValue: my-secret
        isOptional: true
        parameterType: STRING
schemaVersion: 2.1.0
sdkVersion: kfp-2.8.0
---
platforms:
  kubernetes:
    deploymentSpec:
      executors:
        exec-print-secret:
          secretAsVolume:
          - mountPath: /mnt/my_vol
            optional: false
            secretName: secret_name
```

I tried a bunch of combinations to have the `defaultValue` field validated, so we know the name is being passed correctly, but the structure of this tests is such that if I pass in the entire pipelineSpec in the assert it fails:
https://github.com/kubeflow/pipelines/actions/runs/10191867165/job/28193820249#step:4:886

The tests only work when `platform_spec` is extracted and matched.

Note that the assert statement still verifies that the `secret_name` variable is getting passed into the `secretAsVolume`, I believe that's a good enough preliminary test at this point. Wdyt? @diegolovison @gregsheremeta

gregsheremeta: for posterity

> The problem here is the pipeline spec generated from this compile step has the name stored in as a variable under the roots section.

we discovered today that that's not how this works. `secretName` (at the very bottom of the platforms spec) has to say `secretName: my-secret` and not `secretName: secret_name`

gregsheremeta: also for posterity, the above comment is also incorrect :sweat_smile: 

`secretName: secret_name` actually needs to be a key to the input parameter map, so we can grab the parameter by name at runtime. Newly added changes to driver.go show how we do that.

DharmitD:
@chensun could you please take a look and let us know if this templating method looks good to you? Thanks!

DharmitD:
@chensun could you please approve? Thanks!

DharmitD:
@chensun could you please take a look and let us know if this templating method looks good to you? Thanks!

DharmitD:
@chensun bumping :)

cc: @gregsheremeta @rimolive

kubernetes_platform/python/kfp/kubernetes/secret.py
```diff
@@ -75,11 +76,16 @@ def use_secret_as_volume(
...
         Task object with updated secret configuration.
     """
-
     msg = common.get_existing_kubernetes_config_as_message(task)
 
+    val = secret_name
+    # if secret_name is a PipelineParameterChannel, then we don't know what secret to mount until RUNTIME
+    # so, treat is as a map KEY instead of a secret name
+    if isinstance(secret_name, PipelineParameterChannel):
+        val = "{{" + secret_name.name + "}}"
```
HumairAK, line 85: I am uncertain about this approach, we're essentially implementing templating, but what seems like a hacky way. Have we considered other approaches? 

For example, we for names retrieved via parameterschannels, we can include a new field for `SecretAsEnv` called "secret_name_parameter". 

Based on last community call @chensun had a proposal for an alternative approach as well, so I would be interested in hearing about this as well.

HumairAK:
I've tested the implementation and it works as intended, however I think it's worth getting input from @chensun on the approach, I've left a comment [here](https://github.com/kubeflow/pipelines/pull/11039/files#r1734682355) as well.

kubernetes_platform/python/kfp/kubernetes/secret.py
gregsheremeta: > we can include a new field for SecretAsEnv called "secret_name_parameter".

that _does_ feel cleaner to me

backend/src/v2/driver/driver.go
```diff
@@ -536,14 +538,40 @@ func extendPodSpecPatch(
 	// Get secret mount information
 	for _, secretAsVolume := range kubernetesExecutorConfig.GetSecretAsVolume() {
 		optional := secretAsVolume.Optional != nil && *secretAsVolume.Optional
+
+		secretName := secretAsVolume.GetSecretName()
+
+		if strings.HasPrefix(secretName, "{{") && strings.HasSuffix(secretName, "}}") {
```
chensun, line 544: As discussed in the community meeting, we suggest follow the "inject inputs" pattern to accept dynamic runtime value.
Example PRs on "injecting inputs":
https://github.com/kubeflow/pipelines/pull/10435 
https://github.com/kubeflow/pipelines/pull/5183

chensun (changes requested):

HumairAK:
closing in lieu of https://github.com/kubeflow/pipelines/pull/11621

PR #11096, opened 2024-08-13
fix(frontend): fixes filter pipeline text box shows error when typing anything in it. Fixes #10241

### LINKED ISSUE #10241: [frontend] Filter pipeline text box shows error when typing anything in it
### Environment

*  How did you deploy Kubeflow Pipelines (KFP)? kustomization
*  KFP version:  2.0.3

### Steps to reproduce

* Select `Runs` on left sidebar.
* Select `+ Create run` on top right sidebar.
* Select `Choose` in `Pipeline` input box.
* `Type a` in `Filter pipelines` input text.
* An `Error retrieving resources` is poped up, with the following error message

```
List request failed with:
{"error":"Failed to list pipelines due invalid list options: pageToken: , pageSize: 10, sortBy: created_at desc, filter: %7B%22predicates%22%3A%5B%7B%22key%22%3A%22name%22%2C%22op%22%3A%22IS_SUBSTRING%22%2C%22string_value%22%3A%22a%22%7D%5D%7D: Invalid input error: failed to parse valid filter from \"%7B%22predicates%22%3A%5B%7B%22key%22%3A%22name%22%2C%22op%22%3A%22IS_SUBSTRING%22%2C%22string_value%22%3A%22a%22%7D%5D%7D\": unknown field \"op\" in kubeflow.pipelines.backend.api.v2beta1.Predicate","code":3,"message":"Failed to list pipelines due invalid list options: pageToken: , pageSize: 10, sortBy: created_at desc, filter: %7B%22predicates%22%3A%5B%7B%22key%22%3A%22name%22%2C%22op%22%3A%22IS_SUBSTRING%22%2C%22string_value%22%3A%22a%22%7D%5D%7D: Invalid input error: failed to parse valid filter from \"%7B%22predicates%22%3A%5B%7B%22key%22%3A%22name%22%2C%22op%22%3A%22IS_SUBSTRING%22%2C%22string_value%22%3A%22a%22%7D%5D%7D\": unknown field \"op\" in kubeflow.pipelines.backend.api.v2beta1.Predicate","details":[{"@type":"type.googleapis.com/google.rpc.Status","code":3,"message":"failed to parse valid filter from \"%7B%22predicates%22%3A%5B%7B%22key%22%3A%22name%22%2C%22op%22%3A%22IS_SUBSTRING%22%2C%22string_value%22%3A%22a%22%7D%5D%7D\": unknown field \"op\" in kubeflow.pipelines.backend.api.v2beta1.Predicate"}]}
```

### Expected result

There shall be no error message when typing in  `Filter pipelines` input box.

### Materials and Reference

---

ElayAharoni:
fixes filter pipeline text box shows error when typing anything in it.

fix for issue #10241.

just did small change to use the V2 code instead the V1

### DIFF
--- a/frontend/src/pages/__snapshots__/ResourceSelector.test.tsx.snap
+++ b/frontend/src/pages/__snapshots__/ResourceSelector.test.tsx.snap
@@ -29,7 +29,7 @@ exports[`ResourceSelector displays resource selector 1`] = `
     emptyMessage="Test - Sorry, no resources."
     filterLabel="test filter label"
     initialSortColumn="created_at"
-    isCalledByV1={true}
+    isCalledByV1={false}
     reload={[Function]}
     rows={
       Array [
--- a/frontend/src/pages/ResourceSelector.tsx
+++ b/frontend/src/pages/ResourceSelector.tsx
@@ -78,7 +78,7 @@ class ResourceSelector extends React.Component<ResourceSelectorProps, ResourceSe
         {title && <Toolbar actions={toolbarActionMap} breadcrumbs={[]} pageTitle={title} />}
 
         <CustomTable
-          isCalledByV1={true}
+          isCalledByV1={false}
           columns={columns}
           rows={rows}
           selectedIds={selectedIds}


hbelmiro:
@andreyvelich if my understanding is correct, `/ok-to-test` should trigger GitHub Actions after [this change](https://github.com/GoogleCloudPlatform/oss-test-infra/pull/2331/files#diff-61d4ecf8ed653ce1a03b8a3758c9fa61af90500fd05f6580dbafcfef3ec780f0R345). But it didn't.
Screenshot from 2024-08-13 12-57-51

Do you know if there's anything else that we should do, or is that a bug?

cc @rimolive @DharmitD

andreyvelich:
> /ok-to-test

Add `/rerun-all` comment to run the actions

boarder7395:
@ElayAharoni Take a look at my comment on the original issue https://github.com/kubeflow/pipelines/issues/10241#issuecomment-2292023909

While this does fix the create run page, it will break the clone from run page. 

Steps to reproduce with your change.
* Open Runs page.
* Click on an existing pipeline Run 
* Click "Clone Run"
* Click "Choose" on the pipeline (or experiment).
* Enter search text.

HumairAK:
@boarder7395 I was unable to reproduce your issue with clone run, can you describe what your environment a bit more? 

I tried this on standalone and it seemed to work fine: 
Before:
pipeline-search-err-before


After: 
pipeline-search-err-after

With clone run: 
pipeline-search-err-after-clone-run

frontend/src/pages/__snapshots__/ResourceSelector.test.tsx.snap
```diff
@@ -29,7 +29,7 @@ exports[`ResourceSelector displays resource selector 1`] = `
     emptyMessage="Test - Sorry, no resources."
     filterLabel="test filter label"
     initialSortColumn="created_at"
-    isCalledByV1={true}
+    isCalledByV1={false}
```
HumairAK, line 32: While this does work for listing v2 pipelines, I'm guessing it breaks it for v1 pipelines. I'm trying to wrap my head around why this `isCalledByV1` exists, because you can either listPipelines based off v1 api or v2 api, but not both, yet can have v1/v2 pipelines uploaded together in KFP, which makes me think that we actually don't properly support listing/filtering for both kfp v1 and v2 pipelines together 

@zijianjoy does that sound right to you?

either way I would rather we have this work with v2 rather than v1, and discuss a better design approach on how to handle this case for when we have both pipelines uploaded and need to list/filter simultaneously

@zijianjoy wdyt?

boarder7395:
@HumairAK You're right when I was using clone from run with a pipelines created in V1 I ran into the following error:
image

When I use a V2 pipeline the functionality works as expected. I can understand merging this since it does solve the "Create Run" issue, which we've actually implemented on our fork as well, since that is a bigger issue than it breaking the "Clone from V1 Pipeline" page. 

Should I open a separate issue on the broken "Clone from v1 pipeline" search field?

Also what tool are you using for browser recording, that would be super helpful for future issues.

HumairAK:
@boarder7395 
> Should I open a separate issue on the broken "Clone from v1 pipeline" search field? 

If we merge this I think we would need a follow up issue for resolving search fields for when v1 pipelines are present. I've prompted @zijianjoy  for suggestions [here](https://github.com/kubeflow/pipelines/pull/11096#discussion_r1718910841), so let's wait for a response. 

> Also what tool are you using for browser recording, that would be super helpful for future issues.

I'm on linux, and use a tool called [peek](https://github.com/phw/peek), it's a super basic tool, but it certainly fits my needs 😄

frontend/src/pages/__snapshots__/ResourceSelector.test.tsx.snap
zijianjoy: isCalledByV1 was introduced to differentiate API request payload schema between V1 API and V2 API. https://github.com/kubeflow/pipelines/blob/1c7954de261b22fa93905cf11b6b280fdf19c320/frontend/src/components/CustomTable.tsx#L229-L246. Because we have a flag to control whether to show v2 UI by default during the initial migration.

Ideally v2 should support filer for V1 names filtering as well. So it makes sense to me for switching to V2 filter mechanism.

zijianjoy:
/lgtm
/approve

Sounds good to me for fixing V2 scenario first. Please feel free to create a separate issue if mixing with V1 pipeline causing problem.

Issue #10885, opened 2024-06-10
[bug] samples/core/exit_handler/exit_handler_test.py is failing


hbelmiro:
`samples/core/exit_handler/exit_handler_test.py` is failing

![Screenshot from 2024-06-10 14-37-47](https://github.com/kubeflow/pipelines/assets/11776454/64791385-386c-4109-b85c-f4dec05281ef)

Full logs:
```
I0610 14:24:34.625197      26 launcher_v2.go:90] input ComponentSpec:{
  "executorLabel": "exec-fail-op",
  "inputDefinitions": {
    "parameters": {
      "message": {
        "parameterType": "STRING"
      }
    }
  }
}
I0610 14:24:34.625952      26 cache.go:116] Connecting to cache endpoint 10.96.159.44:8887
WARNING: Running pip as the 'root' user can result in broken permissions and conflicting behaviour with the system package manager. It is recommended to use a virtual environment instead: https://pip.pypa.io/warnings/venv
[KFP Executor 2024-06-10 14:24:40,767 INFO]: Looking for component `fail_op` in --component_module_path `/tmp/tmp.fmftAEDVvr/ephemeral_component.py`
[KFP Executor 2024-06-10 14:24:40,767 INFO]: Loading KFP component "fail_op" from /tmp/tmp.fmftAEDVvr/ephemeral_component.py (directory "/tmp/tmp.fmftAEDVvr" and module name "ephemeral_component")
[KFP Executor 2024-06-10 14:24:40,768 INFO]: Got executor_input:
{
    "inputs": {
        "parameterValues": {
            "message": "Task failed."
        }
    },
    "outputs": {
        "outputFile": "/tmp/kfp_outputs/output_metadata.json"
    }
}
Task failed.
/usr/local/lib/python3.7/runpy.py:109: FutureWarning: Python 3.7 has reached end-of-life. KFP will drop support for Python 3.7 on April 23, 2024. To use new versions of the KFP SDK after that date, you will need to upgrade to Python >= 3.8. See https://devguide.python.org/versions/ for more details.
  __import__(pkg_name)
I0610 14:24:40.791065      26 launcher_v2.go:151] publish success.
F0610 14:24:40.791111      26 main.go:49] failed to execute component: exit status 1
```

### Steps to reproduce

* Enable the tests in:
  - https://github.com/kubeflow/pipelines/blob/4dab1914de393ae2c834743c459cd6f188d1a0ab/.github/workflows/e2e-test.yaml#L72

[...]

PR #11100, opened 2024-08-14
fix: re-enable exit hanler test.

liavweiss:
Fix for #10885
In this case, the fail_op task fails as intended, and the ExitHandler task is executed afterward to handle the exit. The log entry "Exit handler has worked!" confirms that the ExitHandler worked as expected after the failure.

It is the "fail" log from the fail-ops task that indicates that the task performed as expected:
image, 
As a result of this failure, we were able to obtain the following result:
image
 "message": "Exit handler has worked!"


hbelmiro (commented):
/lgtm

/hold
@liavweiss can you please add the following to https://github.com/kubeflow/pipelines/pull/11100#issue-2466364709 so the issue is closed automatically when this PR gets merged?

```
Resolves: https://github.com/kubeflow/pipelines/issues/10885
```

liavweiss:
@hbelmiro thanks for the review, I added the issue number to my description.

hbelmiro (approved):
Thank you for this contribution @liavweiss.

PR #11130, opened 2024-08-22
fix(frontend): first time choosing a pipeline definition is VERY slow. Fixes #10897

### LINKED ISSUE #10897: [frontend] first time choosing a pipeline definition is VERY slow
### Environment

*  KFP version: All Kubeflow Pipelines 2.X.X versions.

### Steps to reproduce

The first time you try and select the pipeline definition for a run, its CRAZY SLOW (>60 seconds of loading sometimes).

1. Going to the "Runs" page
2. Clicking "Create Run"
3. Clicking "Choose" pipeline
4. Finding a pipeline and clicking "Use this Pipeline"
5. Observe the crazy slow response.

### Related issues

Possibly related to:

- https://github.com/kubeflow/pipelines/issues/10241

Possibly similar to:

- https://github.com/kubeflow/pipelines/issues/10778

---

ElayAharoni:
Fixes #10897.
i found that whenever a V2 pipeline is chosen for a new run, then there is a check whether v1Template is fetching and because it is not a v1 pipeline then the message is seen for long time.

so basically what i did is check whether the v1 template is fetching only when the pipeline is v1 and not v2, which cause in significant reduce in the loading time.

### DIFF
--- a/frontend/src/pages/NewRunSwitcher.tsx
+++ b/frontend/src/pages/NewRunSwitcher.tsx
@@ -162,7 +162,7 @@ function NewRunSwitcher(props: PageProps) {
     recurringRunIsFetching ||
     pipelineIsFetching ||
     pipelineVersionIsFetching ||
-    v1TemplateStrIsFetching ||
+    (!isTemplateV2(templateString) && v1TemplateStrIsFetching) ||
     experimentIsFetching
   ) {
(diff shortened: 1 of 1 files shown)


ElayAharoni:
please review when you get the chance.
@HumairAK , @zijianjoy

chensun (approved):
/lgtm
/approve

PR #11163, opened 2024-09-03
feat(frontend/backend): Allow the ability to sort experiments by last run creation. Fixes #10884

### LINKED ISSUE #10884: [feature] Allow the ability to sort Experiments by "last run creation"
### Feature Area

/area frontend
/area backend

When experiments list get's rather long, would be very useful to be able to sort the experiments by "last run creation date": 

In the UI this would result in the user being able to click the "Last 5 runs" column header, and sort by this. Thus also getting rid of the "Cannot sort by this column" mouseover message. 

image

Note that this field does not currently exist in the backend api, and would need to be added. This would result in a new column on the Experiments table for which we could sort this field by. This field would need to be updated everytime a run is created, this means when: 

* a single run is manually created
* when a scheduled run is triggered

It is hard to find experiments that have had a recent run because we cannot sort by this field.

No

---

Love this idea? Give it a 👍.

ElayAharoni:
Fixes #10884.

I cherry picked @HumairAK commit for the BE side.
and did the necessary changes to the FE side.
also built the image and tested the behaviour, it works as requested. (attached video of the new behaviour).
https://www.loom.com/share/4ce09c4c87464285a744cff42d42aa2f?sid=e65c93d4-3116-4f05-8066-66fd0115de63


backend/src/apiserver/model/experiment.go
```diff
@@ -15,12 +15,13 @@
...
-	Name           string       `gorm:"column:Name; not null; unique_index:idx_name_namespace;"`
-	Description    string       `gorm:"column:Description; not null;"`
-	CreatedAtInSec int64        `gorm:"column:CreatedAtInSec; not null;"`
-	Namespace      string       `gorm:"column:Namespace; not null; unique_index:idx_name_namespace;"`
-	StorageState   StorageState `gorm:"column:StorageState; not null;"`
+	UUID                  string       `gorm:"column:UUID; not null; primary_key;"`
+	Name                  string       `gorm:"column:Name; not null; unique_index:idx_name_namespace;"`
+	Description           string       `gorm:"column:Description; not null;"`
+	CreatedAtInSec        int64        `gorm:"column:CreatedAtInSec; not null;"`
+	LastRunCreatedAtInSec int64        `gorm:"column:LastRunCreatedAtInSec; not null;"`
```
HumairAK, line 22: @chensun fyi, this change adds a new column to Experiment table to allow for tracking for the last created run. 

Also note here: https://github.com/kubeflow/pipelines/pull/11163/files#diff-62d02d47cfac0eb11fdad7cf760bac31eb64f26e05aec490934c05bfd7949c09R569 which makes an additional db call during run creation calls.

HumairAK:
cc @chensun

backend/src/apiserver/model/experiment.go
chensun: Adding new columns should be fine. Thanks for the callout.

chensun (approved):
/lgtm
/approve

Issue #10041, opened 2023-09-28
[backend] Nested pipeline does not pass artifacts outputs


omriel1:
Hi!
I was trying to run a nested pipeline, where the first pipeline outputs two artifacts (`Dataset`) and the second one consumes them. I was generally the following example from the docs as the motivation: https://www.kubeflow.org/docs/components/pipelines/v2/data-types/artifacts/#using-output-artifacts

Even though the pipeline compiled successfully and in the UI it seems right, it seems like the pipeline fails (a failure sign near the run, but in the graph itself it seems like the first pipeline still running but its outputs are ready. I attached an image below), and there is no indication in the second pipeline for the failure reason.

If I look at the logs of the failing pod I see:
```
31 main.go:76] KFP driver: driver.DAG(pipelineName=parent-pipeline, runID=b11af0ad-3793-4d8a-a13a-43cce00898e4, task="second-inner-pipeline", component="comp-second-inner-pipeline", dagExecutionID=1, componentSpec) failed: failed to resolve inputs: failed to resolve input artifact table_1 with spec task_output_artifact:{producer_task:"first-inner-pipeline" output_artifact_key:"table_1"}: cannot find output artifact key "table_1" in producer task "first-inner-pipeline"
time="2023-09-28T17:18:05.473Z" level=info msg="sub-process exited" argo=true error="<nil>"
time="2023-09-28T17:18:05.534Z" level=error msg="cannot save parameter /tmp/outputs/execution-id" argo=true error="open /tmp/outputs/execution-id: no such file or directory"
time="2023-09-28T17:18:05.536Z" level=error msg="cannot save parameter /tmp/outputs/iteration-count" argo=true error="open /tmp/outputs/iteration-count: no such file or directory"
time="2023-09-28T17:18:05.536Z" level=error msg="cannot save parameter /tmp/outputs/condition" argo=true error="open /tmp/outputs/condition: no such file or directory"
Error: exit status 1
```

The UI:

<img width="830" alt="Screenshot 2023-09-28 at 20 23 54" src="https://github.com/kubeflow/pipelines/assets/80003866/ea3517c7-ecc0-4573-bef1-bce6f2762569">

### Environment

I used Local deployment on kind:
```bash
kind create cluster --name kfp

export PIPELINE_VERSION=2.0.0
kubectl apply -k "github.com/kubeflow/pipelines/manifests/kustomize/cluster-scoped-resources?ref=$PIPELINE_VERSION"
[...]

PR #4371, opened 2020-08-13
Pipeline IR

hongye-sun:
The PR includes a schema (in proto format) to describe the pipeline intermediate representation (IR) as discussed in https://github.com/kubeflow/pipelines/issues/3703.

PR #11196, opened 2024-09-11
feat(backend): implement subdag output resolution

droctothorpe:

This PR:
- Implements input resolution for outputs that come from (multiple layers of) nested DAGs. Without this, the usefulness of pipelines as components is severely compromised. This is achieved by recording the relationship between DAGs and their producer subtasks in MLMD as custom properties, which are consumed and resolved downstream by `resolveInputs` in `driver.go`.
- Begins functional decomposition in `driver.go`, paving the way for small tests.
- Implements extensive large test cases to validate the many different scenarios. The full set of test cases is summarized in a table below. These test cases are not incorporated into CI yet but will be once #11239 is resolved.
- Implements extensive, togglable debug level logs throughout the driver and mlmd client to improve troubleshooting.

Test file | Description 
--- | ---
artifact.py | Passing an artifact output through two nested pipelines.
artifact_cache.py | Same as above but with cache enabled.
mixed_parameters.py | Passing mixed parameters through two nested pipelines.
multiple_artifacts_namedtuple.py | Passing multiple artifacts through two nested pipelines.
multiple_parameters_namedtuple.py | Passing multiple parameters through two nested pipelines.
parameter.py | Passing a parameter through two nested pipelines.
parameter_cache.py | Same as above but with cache enabled.
parameter_oneof.py | Passing a parameter through a nested pipeline using dsl.OneOf.

**Checklist:**

### DIFF
--- a/backend/src/v2/cmd/driver/main.go
+++ b/backend/src/v2/cmd/driver/main.go
@@ -85,6 +85,7 @@ func init() {
 	flag.Set("logtostderr", "true")
 	// Change the WARNING to INFO level for debugging.
 	flag.Set("stderrthreshold", "WARNING")
+	flag.Set("v", "4")
 }
 
 func validate() error {
--- a/backend/src/v2/cmd/driver/main.go
+++ b/backend/src/v2/cmd/driver/main.go
@@ -85,7 +85,8 @@ func init() {
 	flag.Set("logtostderr", "true")
 	// Change the WARNING to INFO level for debugging.
 	flag.Set("stderrthreshold", "WARNING")
-	flag.Set("v", "4")
+	// Enable V(4) logging level for more verbose debugging.
+	// flag.Set("v", "4")
 }
 
 func validate() error {
--- a/backend/src/v2/cmd/driver/main.go
+++ b/backend/src/v2/cmd/driver/main.go
@@ -85,8 +85,6 @@ func init() {
 	flag.Set("logtostderr", "true")
 	// Change the WARNING to INFO level for debugging.
 	flag.Set("stderrthreshold", "WARNING")
-	// Enable V(4) logging level for more verbose debugging.
-	// flag.Set("v", "4")
 }
 
 func validate() error {
--- a/backend/src/v2/component/launcher_v2.go
+++ b/backend/src/v2/component/launcher_v2.go
@@ -148,6 +148,17 @@ func (l *LauncherV2) Execute(ctx context.Context) (err error) {
 			}
 		}
 		glog.Infof("publish success.")
+		// At the end of the current task, we check the statuses of all tasks in the current DAG and update the DAG's status accordingly.
+		// TODO: If there's a pipeline whose only components are DAGs, this launcher logic will never run and as a result the dag status will never be updated. We need to implement a mechanism to handle this edge case.
+		dag, err := l.metadataClient.GetDAG(ctx, execution.GetExecution().CustomProperties["parent_dag_id"].GetIntValue())
+		if err != nil {
+			glog.Errorf("DAG Status Update: failed to get DAG: %s", err.Error())
+		}
+		pipeline, _ := l.metadataClient.GetPipelineFromExecution(ctx, execution.GetID())
+		err = l.metadataClient.UpdateDAGExecutionsState(ctx, dag, pipeline)
+		if err != nil {
+			glog.Errorf("failed to update DAG state: %s", err.Error())
+		}
 	}()
 	executedStartedTime := time.Now().Unix()
 	execution, err = l.prePublish(ctx)
--- a/backend/src/v2/component/launcher_v2.go
+++ b/backend/src/v2/component/launcher_v2.go
@@ -148,8 +148,8 @@ func (l *LauncherV2) Execute(ctx context.Context) (err error) {
 			}
 		}
 		glog.Infof("publish success.")
-		// At the end of the current task, we check the statuses of all tasks in the current DAG and update the DAG's status accordingly.
-		// TODO: If there's a pipeline whose only components are DAGs, this launcher logic will never run and as a result the dag status will never be updated. We need to implement a mechanism to handle this edge case.
+		// At the end of the current task, we check the statuses of all tasks in
+		// the current DAG and update the DAG's status accordingly.
 		dag, err := l.metadataClient.GetDAG(ctx, execution.GetExecution().CustomProperties["parent_dag_id"].GetIntValue())
 		if err != nil {
 			glog.Errorf("DAG Status Update: failed to get DAG: %s", err.Error())
--- a/backend/src/v2/driver/driver.go
+++ b/backend/src/v2/driver/driver.go
@@ -967,6 +967,44 @@ func validateNonRoot(opts Options) error {
 	return nil
 }
 
+// getDAGTasks gets all the tasks associated with the specified DAG and all of
+// its subDAGs.
+func getDAGTasks(
+	ctx context.Context,
+	dag *metadata.DAG,
+	pipeline *metadata.Pipeline,
+	mlmd *metadata.Client,
+	flattenedTasks map[string]*metadata.Execution,
+) (map[string]*metadata.Execution, error) {
+	if flattenedTasks == nil {
+		flattenedTasks = make(map[string]*metadata.Execution)
+	}
+	currentExecutionTasks, err := mlmd.GetExecutionsInDAG(ctx, dag, pipeline)
+	if err != nil {
+		return nil, err
+	}
+	for k, v := range currentExecutionTasks {
+		flattenedTasks[k] = v
+	}
+	for _, v := range currentExecutionTasks {
+		if v.GetExecution().GetType() == "system.DAGExecution" {
+			glog.V(4).Infof("Found a task, %v, with an execution type of system.DAGExecution. Adding its tasks to the task list.", v.TaskName())
+			subDAG, err := mlmd.GetDAG(ctx, v.GetExecution().GetId())
+			if err != nil {
+				return nil, err
+			}
+			// Pass the subDAG into a recursive call to getDAGTasks and update
+			// tasks to include the subDAG's tasks.
+			flattenedTasks, err = getDAGTasks(ctx, subDAG, pipeline, mlmd, flattenedTasks)
+			if err != nil {
+				return nil, err
+			}
+		}
+	}
+
+	return flattenedTasks, nil
+}
+
 func resolveInputs(ctx context.Context, dag *metadata.DAG, iterationIndex *int, pipeline *metadata.Pipeline, task *pipelinespec.PipelineTaskSpec, inputsSpec *pipelinespec.ComponentInputsSpec, mlmd *metadata.Client, expr *expression.Expr) (inputs *pipelinespec.ExecutorInput_Inputs, err error) {
 	defer func() {
 		if err != nil {
@@ -1138,37 +1176,6 @@ func resolveInputs(ctx context.Context, dag *metadata.DAG, iterationIndex *int, 
 		}
 		return inputs, nil
 	}
-	// get executions in context on demand
-	var tasksCache map[string]*metadata.Execution
-	getDAGTasks := func() (map[string]*metadata.Execution, error) {
-		if tasksCache != nil {
-			return tasksCache, nil
-		}
-		tasks, err := mlmd.GetExecutionsInDAG(ctx, dag, pipeline)
-		if err != nil {
-			return nil, err
-		}
-		// TODO: Make this recursive.
-		for _, v := range tasks {
-			if v.GetExecution().GetType() == "system.DAGExecution" {
-				glog.V(4).Infof("Found a task, %v, with an execution type of system.DAGExecution. Adding its tasks to the task list.", v.TaskName())
-				dag, err := mlmd.GetDAG(ctx, v.GetExecution().GetId())
-				if err != nil {
-					return nil, err
-				}
-				subdagTasks, err := mlmd.GetExecutionsInDAG(ctx, dag, pipeline)
-				if err != nil {
-					return nil, err
-				}
-				for k, v := range subdagTasks {
-					tasks[k] = v
-				}
-			}
-		}
-		tasksCache = tasks
-
-		return tasks, nil
-	}
 
 	for name, paramSpec := range task.GetInputs().GetParameters() {
 		glog.V(4).Infof("name: %v", name)
@@ -1199,60 +1206,70 @@ func resolveInputs(ctx context.Context, dag *metadata.DAG, iterationIndex *int, 
 			if taskOutput.GetOutputParameterKey() == "" {
 				return nil, paramError(fmt.Errorf("output parameter key is empty"))
 			}
-			tasks, err := getDAGTasks()
+			tasks, err := getDAGTasks(ctx, dag, pipeline, mlmd, nil)
 			if err != nil {
 				return nil, paramError(err)
 			}
 
-			// The producer is the task that produces the output that we need to
-			// consume.
-			producer, ok := tasks[taskOutput.GetProducerTask()]
+			// If the producer is a DAG, AND its output / producer subtask is
+			// ALSO a DAG, then we need to cycle through this loop until we
+			// arrive at a non-DAG subtask and essentially bubble up that
+			// non-DAG subtask so that its value can be consumed.
+			producerSubTaskMaybeDAG := true
+			for producerSubTaskMaybeDAG {
+				// The producer is the task that produces the output that we need to
+				// consume.
+				producer := tasks[taskOutput.GetProducerTask()]
 
-			glog.V(4).Info("producer: ", producer)
+				glog.V(4).Info("producer: ", producer)
 
-			// Get the producer's outputs.
-			_, producerOutputs, err := producer.GetParameters()
-			if err != nil {
-				return nil, paramError(fmt.Errorf("get producer output parameters: %w", err))
-			}
-			glog.V(4).Info("producer output parameters: ", producerOutputs)
-			// Deserialize them.
-			var producerOutputsMap map[string]string
-			b, err := producerOutputs["Output"].GetStructValue().MarshalJSON()
-			if err != nil {
-				return nil, err
-			}
-			json.Unmarshal(b, &producerOutputsMap)
-			glog.V(4).Info("producerOutputsMap: ", producerOutputsMap)
-
-			// If the producer's output includes a producer subtask, which means
-			// that the producer is a DAG that is getting its output from one of
-			// the tasks in the DAG, then we want to roll up the output from the
-			// producer subtask to the producer, so that the downstream logic
-			// can retrieve it appropriately.
-			if producerSubTask, ok := producerOutputsMap["producer_subtask"]; ok {
-				glog.V(4).Infof(
-					"Overriding producer task, %v, output with producer_subtask, %v, output.",
-					producer.TaskName(),
-					producerSubTask,
-				)
-				_, producerOutputs, err = tasks[producerSubTask].GetParameters()
+				// Get the producer's outputs.
+				_, producerOutputs, err := producer.GetParameters()
+				if err != nil {
+					return nil, paramError(fmt.Errorf("get producer output parameters: %w", err))
+				}
+				glog.V(4).Info("producer output parameters: ", producerOutputs)
+				// Deserialize them.
+				var producerOutputsMap map[string]string
+				b, err := producerOutputs["Output"].GetStructValue().MarshalJSON()
 				if err != nil {
 					return nil, err
 				}
-				glog.V(4).Info("producerSubTask output parameters: ", producerOutputs)
-				// The only reason we're updating this is to make the downstream
-				// logging more accurate.
-				taskOutput.ProducerTask = producerOutputsMap["producer_subtask"]
+				json.Unmarshal(b, &producerOutputsMap)
+				glog.V(4).Info("producerOutputsMap: ", producerOutputsMap)
+
+				// If the producer's output includes a producer subtask, which means
+				// that the producer is a DAG that is getting its output from one of
+				// the tasks in the DAG, then we want to roll up the output from the
+				// producer subtask to the producer, so that the downstream logic
+				// can retrieve it appropriately.
+				if producerSubTask, ok := producerOutputsMap["producer_subtask"]; ok {
+					glog.V(4).Infof(
+						"Overriding producer task, %v, output with producer_subtask, %v, output.",
+						producer.TaskName(),
+						producerSubTask,
+					)
+					_, producerOutputs, err = tasks[producerSubTask].GetParameters()
+					if err != nil {
+						return nil, err
+					}
+					glog.V(4).Info("producerSubTask output parameters: ", producerOutputs)
+					// The only reason we're updating this is to make the downstream
+					// logging more accurate.
+					taskOutput.ProducerTask = producerOutputsMap["producer_subtask"]
+					// Grab the value of the producer output.
+					producerOutputValue, ok := producerOutputs[taskOutput.GetOutputParameterKey()]
+					if !ok {
+						return nil, paramError(fmt.Errorf("cannot find output parameter key %q in producer task %q", taskOutput.GetOutputParameterKey(), taskOutput.GetProducerTask()))
+					}
+					// Update the input to be the producer output value.
+					inputs.ParameterValues[name] = producerOutputValue
+				} else {
+					// The producer subtask is not a DAG, so we exit the loop.
+					producerSubTaskMaybeDAG = false
+				}
 			}
 
-			// Grab the value of the producer output.
-			producerOutputValue, ok := producerOutputs[taskOutput.GetOutputParameterKey()]
-			if !ok {
-				return nil, paramError(fmt.Errorf("cannot find output parameter key %q in producer task %q", taskOutput.GetOutputParameterKey(), taskOutput.GetProducerTask()))
-			}
-			// Update the input to be the producer output value.
-			inputs.ParameterValues[name] = producerOutputValue
 		case *pipelinespec.TaskInputsSpec_InputParameterSpec_RuntimeValue:
 			runtimeValue := paramSpec.GetRuntimeValue()
 			switch t := runtimeValue.Value.(type) {
@@ -1292,7 +1309,7 @@ func resolveInputs(ctx context.Context, dag *metadata.DAG, iterationIndex *int, 
 			if taskOutput.GetOutputArtifactKey() == "" {
 				return nil, artifactError(fmt.Errorf("output artifact key is empty"))
 			}
-			tasks, err := getDAGTasks()
+			tasks, err := getDAGTasks(ctx, dag, pipeline, mlmd, nil)
 			if err != nil {
 				return nil, artifactError(err)
 			}
--- a/backend/src/v2/driver/driver.go
+++ b/backend/src/v2/driver/driver.go
@@ -973,44 +973,6 @@ func validateNonRoot(opts Options) error {
 	return nil
 }
 
-// getDAGTasks gets all the tasks associated with the specified DAG and all of
-// its subDAGs.
-func getDAGTasks(
-	ctx context.Context,
-	dag *metadata.DAG,
-	pipeline *metadata.Pipeline,
-	mlmd *metadata.Client,
-	flattenedTasks map[string]*metadata.Execution,
-) (map[string]*metadata.Execution, error) {
-	if flattenedTasks == nil {
-		flattenedTasks = make(map[string]*metadata.Execution)
-	}
-	currentExecutionTasks, err := mlmd.GetExecutionsInDAG(ctx, dag, pipeline)
-	if err != nil {
-		return nil, err
-	}
-	for k, v := range currentExecutionTasks {
-		flattenedTasks[k] = v
-	}
-	for _, v := range currentExecutionTasks {
-		if v.GetExecution().GetType() == "system.DAGExecution" {
-			glog.V(4).Infof("Found a task, %v, with an execution type of system.DAGExecution. Adding its tasks to the task list.", v.TaskName())
-			subDAG, err := mlmd.GetDAG(ctx, v.GetExecution().GetId())
-			if err != nil {
-				return nil, err
-			}
-			// Pass the subDAG into a recursive call to getDAGTasks and update
-			// tasks to include the subDAG's tasks.
-			flattenedTasks, err = getDAGTasks(ctx, subDAG, pipeline, mlmd, flattenedTasks)
-			if err != nil {
-				return nil, err
-			}
-		}
-	}
-
-	return flattenedTasks, nil
-}
-
 func resolveInputs(ctx context.Context, dag *metadata.DAG, iterationIndex *int, pipeline *metadata.Pipeline, task *pipelinespec.PipelineTaskSpec, inputsSpec *pipelinespec.ComponentInputsSpec, mlmd *metadata.Client, expr *expression.Expr) (inputs *pipelinespec.ExecutorInput_Inputs, err error) {
 	defer func() {
 		if err != nil {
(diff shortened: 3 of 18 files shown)


backend/src/v2/driver/driver.go
```diff
@@ -749,6 +756,27 @@ func DAG(ctx context.Context, opts Options, mlmd *metadata.Client) (execution *E
...
+		glog.V(4).Info("producerSubtask: ", producerSubTask)
+
+		outputParameterMap := map[string]interface{}{
+			"output_parameter_key": outputParameterKey,
+			"producer_subtask":     producerSubTask,
+		}
+
+		outputParameterStruct, _ := structpb.NewValue(outputParameterMap)
+
+		ecfg.OutputParameters = map[string]*structpb.Value{
```
droctothorpe, line 775: Is it okay to store this here?

backend/src/v2/driver/driver.go
```diff
@@ -125,6 +126,8 @@ func RootDAG(ctx context.Context, opts Options, mlmd *metadata.Client) (executio
 			err = fmt.Errorf("driver.RootDAG(%s) failed: %w", opts.info(), err)
 		}
 	}()
+	b, _ := json.Marshal(opts)
+	glog.V(4).Info("RootDAG opts: ", string(b))
```
droctothorpe, line 130: We added a ton of debug level logs to make debugging stuff like this easier for the next person. We need to add some handling in the backend to support toggling level 4 logs in the driver on and off.

backend/src/v2/driver/driver.go
```diff
@@ -1143,19 +1203,56 @@ func resolveInputs(ctx context.Context, dag *metadata.DAG, iterationIndex *int,
...
 			producer, ok := tasks[taskOutput.GetProducerTask()]
-			if !ok {
-				return nil, paramError(fmt.Errorf("cannot find producer task %q", taskOutput.GetProducerTask()))
-			}
-			_, outputs, err := producer.GetParameters()
+
+			glog.V(4).Info("producer: ", producer)
+
+			// Get the producer's outputs.
+			_, producerOutputs, err := producer.GetParameters()
```
droctothorpe, line 1214: This whole producer subtask lookup should be decomposed and called recursively in case the subtask is *also* a DAG!

droctothorpe:
Happy to jump on a call if synchronous questions / feedback is easier. Although concise, these changes are quite convoluted.

backend/src/v2/driver/driver.go
droctothorpe: Just pushed up a commit that implements this and confirmed that it works with subdags of subdags. Here's is the example code to validate it:

```python
from kfp import dsl
from kfp.client import Client

@dsl.component
def small_comp() -> str:
    return "privet"

@dsl.component
def large_comp(input: str):
    print("input :", input)


@dsl.pipeline
def small_matroushka_doll() -> str:
    task = small_comp()
    task.set_caching_options(False)
    return task.output

@dsl.pipeline
def medium_matroushka_doll() -> str:
    dag_task = small_matroushka_doll()
    dag_task.set_caching_options(False)
    return dag_task.output

@dsl.pipeline
def large_matroushka_doll():
    dag_task = medium_matroushka_doll()
    task = large_comp(input=dag_task.output)
    task.set_caching_options(False)
    dag_task.set_caching_options(False)


if __name__ == "__main__":
    client = Client()

    run = client.create_run_from_pipeline_func(
        pipeline_func=large_matroushka_doll,
        enable_caching=False,
    )
```

PS. I hate matroushka dolls, they're so full of themselves.

droctothorpe:
We just pushed up a commit that implements support for multiple layers of nested subdags (i.e. subdags of subdags). We validated that it behaves as expected with the following example code:

```python
from kfp import dsl
from kfp.client import Client

@dsl.component
def small_comp() -> str:
    return "privet"

@dsl.component
def large_comp(input: str):
    print("input :", input)


@dsl.pipeline
def small_matroushka_doll() -> str:
    task = small_comp()
    task.set_caching_options(False)
    return task.output

@dsl.pipeline
def medium_matroushka_doll() -> str:
    dag_task = small_matroushka_doll()
    dag_task.set_caching_options(False)
    return dag_task.output

@dsl.pipeline
def large_matroushka_doll():
    dag_task = medium_matroushka_doll()
    task = large_comp(input=dag_task.output)
    task.set_caching_options(False)
    dag_task.set_caching_options(False)


if __name__ == "__main__":
    client = Client()

    run = client.create_run_from_pipeline_func(
        pipeline_func=large_matroushka_doll,
        enable_caching=False,
    )
```

PS. I hate matroushka dolls, they're so full of themselves.

droctothorpe:
We're going to test (and likely implement some updates) for output artifacts as opposed to parameters.

droctothorpe:
We just added and validated support for output artifacts as well, which addresses https://github.com/kubeflow/pipelines/issues/10041. Here's a screenshot from a pipeline with nested DAGs and output artifacts that executed successfully:

image

Here's the example code:

```python
from kfp import dsl
from kfp.client import Client

@dsl.component
def inner_comp(dataset: dsl.Output[dsl.Dataset]):
    with open(dataset.path, "w") as f:
        f.write("foobar")


@dsl.component
def outer_comp(input: dsl.Dataset):
    print("input: ", input)


@dsl.pipeline
def inner_pipeline() -> dsl.Dataset:
    inner_comp_task = inner_comp()
    inner_comp_task.set_caching_options(False)
    return inner_comp_task.output
    
@dsl.pipeline
def outer_pipeline():
    inner_pipeline_task = inner_pipeline()
    outer_comp_task = outer_comp(input=inner_pipeline_task.output)
    outer_comp_task.set_caching_options(False)


if __name__ == "__main__":
    # Compiler().compile(outer_pipeline, "ignore/subdag_artifacts.yaml")
    client = Client()

    run = client.create_run_from_pipeline_func(
        pipeline_func=outer_pipeline,
        enable_caching=False,
    )
```

There's still a lot more work to be done in terms of testing, decomposition, making the code more consistent and DRY, etc, but **it works** and it did not work before, so hooray for progress.

backend/src/v2/driver/driver.go
```diff
@@ -1188,43 +1214,235 @@ func resolveInputs(ctx context.Context, dag *metadata.DAG, iterationIndex *int,
...
+			glog.V(4).Info("Deserialized outputParametersMap: ", outputParametersMap)
+			subTaskName := outputParametersMap["producer_subtask"]
+			glog.V(4).Infof(
+				"Overriding currentTask, %v, output with currentTask's producer_subtask, %v, output.",
+				currentTask.TaskName(),
+				subTaskName,
+			)
+
+			// Reassign sub-task before running through the loop again.
+			currentTask = tasks[subTaskName]
```
HumairAK, line 1303: might need to handle some of these potential error cases, for example: 
*  if the producer task is not in outputParametersMap
* subtaskname is not in tasks

backend/src/v2/driver/driver.go
```diff
@@ -749,6 +756,33 @@ func DAG(ctx context.Context, opts Options, mlmd *metadata.Client) (execution *E
...
+		producerSubTask := value.GetValueFromParameter().ProducerSubtask
+		glog.V(4).Info("outputParameterKey: ", outputParameterKey)
+		glog.V(4).Info("producerSubtask: ", producerSubTask)
+
+		outputParameterMap := map[string]interface{}{
+			"output_parameter_key": outputParameterKey,
+			"producer_subtask":     producerSubTask,
+		}
+
+		outputParameterStruct, _ := structpb.NewValue(outputParameterMap)
```
HumairAK, line 774: how come you didn't use a [DagOutputParameterSpec](https://github.com/kubeflow/pipelines/blob/83dcf1a60919f5bcc0c644c8fdff94ad686cad07/api/v2alpha1/pipeline_spec.proto#L157) like you did for [DagOutputArtifactSpec](https://github.com/kubeflow/pipelines/pull/11196/files#diff-87ec6fcace7656e2c64443ea8b5386b8d54215eab715ab193fc0e2c1bfe939d1R139)?

backend/src/v2/metadata/client.go
```diff
@@ -576,6 +582,24 @@ func (c *Client) CreateExecution(ctx context.Context, pipeline *Pipeline, config
...
+			},
+		}}
+	}
+	if config.OutputArtifacts != nil {
+		b, err := json.Marshal(config.OutputArtifacts)
+		if err != nil {
+			return nil, err
+		}
+		e.CustomProperties[keyOutputArtifacts] = StringValue(string(b))
+	}
```
HumairAK, line 602: one thing I'm thinking about is how we distinguish between `outputs` for container executions vs DAG executions, since for paramers in container executions they map to the actual resolved values, but for DAG we're simply storing reference values to the producer tasks

similarly for artifacts, for container executions we store the artifact obj store metadata (pulled from artifact properties), but for dag executions we are again, storing the producer metadata, right? 

so I'm wondering if it makes sense to have a separate parameter entirely to distinguish these types, I don't have a concrete suggestion, maybe something like parameter_producer_tasks and artifact_producer_tasks 🤷🏾‍♂️ 


I'm not sure but atm if we use `outputs` then it will probably show up on the UI for executions page under the DAG but show something different than it does for Container executions

backend/src/v2/driver/driver.go
```diff
@@ -749,6 +756,33 @@ func DAG(ctx context.Context, opts Options, mlmd *metadata.Client) (execution *E
...
+
+		ecfg.OutputParameters = map[string]*structpb.Value{
+			value.GetValueFromParameter().OutputParameterKey: outputParameterStruct,
+		}
```
HumairAK, line 778: this looks like you are re writing ecfg.OutputParameters with a new map every iteration, did you mean to update the map instead?

HumairAK (commented):
Hey folks, love that you are doing this, amazing stuff!!


I just had a skim and left some quick thoughts, I see that it's still WIP, so apologies if the comments are premature. Haven't had a chance to try it out yet.

backend/src/v2/driver/driver.go
```diff
@@ -1188,43 +1214,235 @@ func resolveInputs(ctx context.Context, dag *metadata.DAG, iterationIndex *int,
...
+	ctx context.Context,
+	dag *metadata.DAG,
+	pipeline *metadata.Pipeline,
+	mlmd *metadata.Client,
+	flattenedTasks map[string]*metadata.Execution,
+) (map[string]*metadata.Execution, error) {
+	if flattenedTasks == nil {
+		flattenedTasks = make(map[string]*metadata.Execution)
+	}
+	currentExecutionTasks, err := mlmd.GetExecutionsInDAG(ctx, dag, pipeline)
```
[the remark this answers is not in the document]
droctothorpe: 🔥 Right now, the number of calls is equal to the number of nested sub-DAGs. If the call sans filter gets all executions AND executions in sub-DAGs, that could definitely reduce the number of database queries. We'll test it out. Thanks for the suggestion!

backend/src/v2/driver/driver.go
droctothorpe: The logic we added to the mlmd client handles converting the **artifact** struct into a format suitable for storage in the database. Our reasoning was that that was an implementation detail that wasn't really relevant to the driver. That logic holds true for the output parameters as well, but we never refactored it to apply the principle. We'll move that logic out of `driver.go` and into `client.go`. Good call!

backend/src/v2/driver/driver.go
droctothorpe: You're absolutely right! We would likely have hit a wall because of this when we started tested multiple outputs.

Really appreciate the time and effort you took to grok some not particularly grokkable code. Thank you for the feedback, @HumairAK!

HumairAK: Anytime! feel free to @ me here or slack if you'd like another review

boarder7395: Awesome! I spent like 2 hours trying to dig up where the overwriting of outputs was happening! Thank @HumairAK

backend/src/v2/driver/driver.go
```diff
@@ -1188,43 +1214,235 @@ func resolveInputs(ctx context.Context, dag *metadata.DAG, iterationIndex *int,
...
+			// Deserialize the output artifacts.
+			var outputArtifacts map[string]*pipelinespec.DagOutputsSpec_DagOutputArtifactSpec
+			err := json.Unmarshal([]byte(outputArtifactsCustomProperty.GetStringValue()), &outputArtifacts)
+			if err != nil {
+				return err
 			}
-			// TODO(Bobgy): cache results
-			outputs, err := mlmd.GetOutputArtifactsByExecutionId(ctx, producer.GetID())
+			glog.V(4).Infof("Deserialized outputArtifacts: %v", outputArtifacts)
+			artifactSelectors := outputArtifacts["Output"].GetArtifactSelectors()
```
boarder7395, line 1368: Key "Output" only works when a component has a single output. We should loop over the keys here.
```
			for k, value := range outputArtifacts {
				glog.V(4).Infof("k: %v", k)
				glog.V(4).Infof("outputArtifacts[k]: %v", value)
				artifactSelectors := value.GetArtifactSelectors()
				for _, v := range artifactSelectors {
					glog.V(4).Infof("v: %v", v)
					glog.V(4).Infof("v.ProducerSubtask: %v", v.ProducerSubtask)
					glog.V(4).Infof("v.OutputArtifactKey: %v", v.OutputArtifactKey)
					subTaskName = v.ProducerSubtask
					outputArtifactKey = v.OutputArtifactKey
				}
			}
```

boarder7395: Note: This change breaks the lookup loop. I think we'll have to go to full recursion.

backend/src/v2/driver/driver.go
```diff
@@ -1188,43 +1214,235 @@ func resolveInputs(ctx context.Context, dag *metadata.DAG, iterationIndex *int,
...
+			return err
+		}
+		// If the current task is a DAG:
+		if *currentTask.GetExecution().Type == "system.DAGExecution" {
+			// Since currentTask is a DAG, we need to deserialize its
+			// output parameter map so that we can look its
+			// corresponding producer sub-task, reassign currentTask,
+			// and iterate through this loop again.
+			var outputParametersMap map[string]string
+			b, err := outputParametersCustomProperty["Output"].GetStructValue().MarshalJSON()
```
boarder7395, line 1289: Key "Output" only works when a component has a single output. We should loop over the keys here.
```
			for k, value := range outputParametersCustomProperty {
			        var outputParametersMap map[string]string
				b, err := value. GetStructValue().MarshalJSON()
			        if err != nil {
				        return err
			        }
			        json.Unmarshal(b, &outputParametersMap)
				subTaskName := outputParametersMap["producer_subtask"]
				outputArtifactKey := outputParametersMap["output_artifact_key"]
			}
```

Note: This breaks the loop logic.

droctothorpe:
Huge shout out to @boarder7395 for adding support for multiple artifacts and parameters. We merged his commit into this PR. We also added extensive large tests to the `samples/v2/subdagio` directory and updated `sample_test.py` to execute them. For now, because the version of the driver being used in CI doesn't include the necessary changes. That's a separate issue being tracked in #11239. That being said, we've run **all** of the test cases against a local KFP deployment and they all run without an issue. The cases include:


Test file | Description 
--- | ---
artifact.py | Passing an artifact output through two nested pipelines.
artifact_cache.py | Same as above but with cache enabled.
mixed_parameters.py | Passing mixed parameters through two nested pipelines.
multiple_artifacts_namedtuple.py | Passing multiple artifacts through two nested pipelines.
multiple_parameters_namedtuple.py | Passing multiple parameters through two nested pipelines.
parameter.py | Pass a parameter through two nested pipelines.
parameter_cache.py | Same as above but with cache enabled.

We'll address your remaining comments shortly, @HumairAK.

zazulam:
Hey @HumairAK, we managed to address most of your suggestions which required us to add a few more handlings, mainly the case for uses of `ValueFromOneof` in pipelines. Given that `pipelinespec.DagOutputsSpec_DagOutputParameterSpec` has 2 kinds for its potential value we had to include some logic for handling Oneof (previously, we were only assuming logic for `ValueFromParameter`). Since there isn't any true ordering logic for which output to grab for Oneof we opted to pretty much use the output of the first successfully executed task. We definitely want to hear others' thought on this as it different case and I was not sure if it was missing from the backend or if it gets handled solely from the sdk. 

Another piece is that for Oneof we would need to be able to grab the status of the Oneof branches. It seems that they are registered as DAGs in the backend therefore we need to be able to have the sub-DAGs have their status' updated which is a known issue. I drafted a mini solution for this where after a task is successfully published, in the launcher we will attempt to check the statuses of the tasks associated with the current DAG context and update it accordingly. Granted there is the case where the last task could be a DAGExecution and that won't render the update but we set that aside for now to tackle after. We also took the sample of the oneof test and tossed it into the subdag tests scenarios.

Some screenshots of (some of) the subDAG statuses changing 🎉 
 
image
image
image
image

zazulam:
Found this nugget from #4371 😄 
image

droctothorpe:
Bump, @chensun. Hoping this can be reviewed before the next KFP meeting so that we can address any feedback. Thanks! 🙏

droctothorpe:
Just pushed up a [commit](https://github.com/kubeflow/pipelines/pull/11196/commits/93d1f3ce5e092063ca64c3287871e908395d343e) that ensures that all of the large tests that we wrote for the various edge cases addressed by this PR are executed in CI as a condition for merge. This should improve confidence in the changes and protect it from future regressions. Big shout out to @CarterFendley for implementing the PRs (https://github.com/kubeflow/pipelines/pull/11284 and https://github.com/kubeflow/pipelines/pull/11294) that made this possible (and to everyone who reviewed / merged them).

droctothorpe:
The `oneof` test is failing for some reason. In addition, the tests appear to be executed twice (in `KFP Samples / samples` and `Kubeflow Pipelines V2 integration Tests / integration-tests`). Investigating.

droctothorpe:
There's one failing check that we're zeroing in on. This PR should be merged until we address it. Will keep ya'll posted.

droctothorpe:
@chensun @HumairAK all CI tests (including the many subdag IO ones that we implemented) are passing against the latest version of the driver and launcher images. We had to switch back to the recursive `getDAGTasks` implementation in the driver to circumvent [this](https://github.com/kubeflow/pipelines/blob/master/backend/src/v2/metadata/client.go#L743-L744) feature gap, which is fairly significant IMO. The keys in the map are the task names, which are not unique. Any parallelFor will have collisions, as will any nested DAG where the same component is used in both the parent and child DAG. Even if we implement a way to store duplicate task names, the producer subtask custom property that stores the relationship doesn't provide any information that will let us differentiate between which specific task it's looking for! Given the scope of the problem, the fact that it will require modifying how things are stored in MLMD, the size of this PR, and the fact that it already successfully addresses a major feature gap, we think it's probably best that that change be tackled in a follow-up PR (after design review ofc).

chensun:
> @chensun @HumairAK all CI tests (including the many subdag IO ones that we implemented) are passing against the latest version of the driver and launcher images. We had to switch back to the recursive `getDAGTasks` implementation in the driver to circumvent [this](https://github.com/kubeflow/pipelines/blob/master/backend/src/v2/metadata/client.go#L743-L744) feature gap, which is fairly significant IMO. The keys in the map are the task names, which are not unique. Any parallelFor will have collisions, as will any nested DAG where the same component is used in both the parent and child DAG. Even if we implement a way to store duplicate task names, the producer subtask custom property that stores the relationship doesn't provide any information that will let us differentiate between which specific task it's looking for! Given the scope of the problem, the fact that it will require modifying how things are stored in MLMD, the size of this PR, and the fact that it already successfully addresses a major feature gap, we think it's probably best that that change be tackled in a follow-up PR (after design review ofc).

Can you summarize what's the failure mode with the current implementation? Under what circumstances will there be collisions? Task names are unique within a DAG, so in theory there shouldn't be collisions under a parallel loop unless there's bug in implementation, otherwise how do we distinguish which parallel tasks produced which outputs. 
For the same component used both inside and outside a dag, we can avoid collision by prefixing the DAG name.

droctothorpe:
> Can you summarize what's the failure mode with the current implementation? Under what circumstances will there be collisions? Task names are unique within a DAG, so in theory there shouldn't be collisions under a parallel loop unless there's bug in implementation, otherwise how do we distinguish which parallel tasks produced which outputs. For the same component used both inside and outside a dag, we can avoid collision by prefixing the DAG name.

Sure thing. This PR was actually able to address both edge cases introduced by the collisions, we just think that handling the collisions directly would probably make the code cleaner and easier to understand.

The first edge case surfaces as a result of `parallelFor`. The [solution](https://github.com/kubeflow/pipelines/pull/11196/files#diff-9cde79602d658770cae19cfc598a2afa45658f5f4904f038aa673c2eb056ea35R1243-R1252) is to refrain from adding `parallelFor` DAGs to the map.

The second edge case is extremely unlikely.

Suppose you have three DAGs, A, B, and C. A calls B and B calls C. B and C share a component. Because the shared component task is in two separate DAGs, the task name is not incremented by the IR compiler, so it winds up being identical and stored in MLMD with the same value in the task name column.

The issue surfaces when a component in A needs the output of DAG B, which comes from a producer subtask. When `resolveInputs` tries to look up the producer subtask, there are two tasks which match the specified name.

Initially, we tried calling [GetExecutionsInDag](https://github.com/kubeflow/pipelines/pull/11196/files#diff-87ec6fcace7656e2c64443ea8b5386b8d54215eab715ab193fc0e2c1bfe939d1R806) without a filter (to get tasks from all DAGs in the specified context), but that inevitably raised the [collision error](https://github.com/kubeflow/pipelines/pull/11196/files#diff-87ec6fcace7656e2c64443ea8b5386b8d54215eab715ab193fc0e2c1bfe939d1R860) because of the two tasks with the same name.

It's an extremely unlikely edge case but we circumvented it using [recursive DFS](https://github.com/kubeflow/pipelines/pull/11196/files) and overwriting higher level matches in the nested DAG with lower level matches. After making this change, all tests passed.

We think that avoiding collisions with unique keys is a better long term approach but were concerned about making the PR any bigger than it already is.

Prefixing with the DAG name is a great idea! Another approach that just occurred to me, though it's a client-side fix rather than a backend fix, is updating the IR compiler to avoid duplicate task names across layers of nested DAGs, much like the compiler is already doing within a given DAG. 

I hope that's clear. It's a rare, unlikely, and difficult to explain edge case.

droctothorpe:
> Task names are unique within a DAG, so in theory there shouldn't be collisions under a parallel loop unless there's bug in implementation

This did not seem to be the case in our testing. Maybe it's a separate bug or maybe we misconfigured something. Let me find some example code.

droctothorpe:
Given [this](https://github.com/kubeflow/pipelines/blob/6a35ee5144fba1c72badc7c52161d8a49f34804e/sdk/python/test_data/pipelines/pipeline_with_loops.py#L4) pipeline, which is part of the SDK execution tests workflow, if you run the following query on the `metadb` database:

```sql
select * from ExecutionProperty where name = "task_name";
```

You get the following:
image

Note how there are:
- 3 x `for-loop-1` task names.
- 3 x `for-loop-4` task names.

These are from a single run.

droctothorpe:
If this is unintended `parallelFor` behavior, @chensun, let me know and I'll file a corresponding issue. I'll file another issue as well about using unique `task_names` in the `ExecutionProperties` table and mention your <dag name>_<task_name> suggestion.

chensun:
> If this is unintended `parallelFor` behavior, @chensun, let me know and I'll file a corresponding issue. I'll file another issue as well about using unique `task_names` in the `ExecutionProperties` table and mention your `<dag name>_<task_name>` suggestion.
> 
> Note from today's community call: make sure to tag @gmfrasca in the issues.

I see. I don't know on top of my head if this MLMD behavior is problematic or just fine. I was thinking about IR, where `tasks` is a map so the key has to be unique. Based on the screenshot you have, it might be okay that this task name is not unique, because the first column appears to be a unique ID, and I assume we can locate the correct task even with duplicate names, otherwise I'd imagine the entire parallelfor function would be broken.

chensun:
> Prefixing with the DAG name is a great idea! Another approach that just occurred to me, though it's a client-side fix rather than a backend fix, is updating the IR compiler to avoid duplicate task names across layers of nested DAGs, much like the compiler is already doing within a given DAG.

I definitely need to read the code to see what fix you're proposing, but I think client-side fix won't be sufficient, for the parallelfor case, we don't always know the exact number of tasks at compilation time.

backend/src/v2/component/launcher_v2.go
```diff
@@ -148,6 +148,17 @@ func (l *LauncherV2) Execute(ctx context.Context) (err error) {
 			}
 		}
 		glog.Infof("publish success.")
+		// At the end of the current task, we check the statuses of all tasks in the current DAG and update the DAG's status accordingly.
+		// TODO: If there's a pipeline whose only components are DAGs, this launcher logic will never run and as a result the dag status will never be updated. We need to implement a mechanism to handle this edge case.
```
chensun, line 152: > If there's a pipeline whose only components are DAGs

I can't imagine a valid case for this. and I vaguely recall the compiler check that a pipeline must have concrete tasks.

backend/src/v2/cmd/driver/main.go
```diff
@@ -85,6 +85,8 @@ func init() {
 	flag.Set("logtostderr", "true")
 	// Change the WARNING to INFO level for debugging.
 	flag.Set("stderrthreshold", "WARNING")
+	// Enable V(4) logging level for more verbose debugging.
+	// flag.Set("v", "4")
```
chensun, line 89: remove?

backend/src/v2/driver/driver.go
chensun: For my education, I'm not very familiar with v4 log, can't this be glog.debug instead?

backend/src/v2/driver/driver.go
```diff
@@ -1188,43 +1198,311 @@ func resolveInputs(ctx context.Context, dag *metadata.DAG, iterationIndex *int,
...
-			producer, ok := tasks[taskOutput.GetProducerTask()]
+		}
+	}
+
+	return flattenedTasks, nil
+}
+
+// resolveUpstreamParametersConfig is just a config struct used to store the
+// input parameters of the resolveUpstreamParameters function.
+type resolveUpstreamParametersConfig struct {
```
chensun, line 1272: Do you see opportunities to consolidate and merge this with `resolveUpstreamArtifacts...`?
I tend to think the logic should be very similar between the two?

samples/v2/subdagio/__init__.py
```diff
@@ -0,0 +1,8 @@
+from subdagio import artifact
```
chensun, line 1: why do we need an `__init__.py` file in the sample folder?

samples/v2/subdagio/mixed_parameters.py
```diff
@@ -0,0 +1,48 @@
...
+
+    task = crust_comp(x=2, y=dag_task.output)
+    task.set_caching_options(False)
+
+
+if __name__ == '__main__':
+    Compiler().compile(
+        pipeline_func=crust,
+        package_path=f"{__file__.removesuffix('.py')}.yaml")
+    client = Client()
```
chensun, line 47: nit: This is a bit unconventional to include a submission code in the execution body, not sure users would intend to run it remotely.

zazulam:
My thoughts on the ParallelFor tasks is that for` dsl.Collected` is meant to be the utility that the user should be using to obtain the outputs from tasks instantiated within the loop, therefore this solution for obtaining outputs from subdags has no reason to parse through the executions mapped to ParallelFor tasks, which gets registered as a DAG along with its iterations. 

As for the issue with the same task used in multiple dags of a a pipeline, that is where a solution is needed. Here is the corresponding IR from this [pipeline](https://github.com/kubeflow/pipelines/blob/master/sdk/python/test_data/pipelines/pipeline_with_outputs.py). 

```yaml
components:
  comp-inner-pipeline:
    dag:
      outputs:
        artifacts:
          data:
            artifactSelectors:
              - outputArtifactKey: data
                producerSubtask: print-op2
        parameters:
          msg:
            valueFromParameter:
              outputParameterKey: Output
              producerSubtask: print-op1
      tasks:
        print-op1:
          cachingOptions:
            enableCache: true
          componentRef:
            name: comp-print-op1-2
          inputs:
            parameters:
              msg:
                componentInputParameter: msg
          taskInfo:
            name: print-op1
        print-op2:
          cachingOptions:
            enableCache: true
          componentRef:
            name: comp-print-op2
          dependentTasks:
            - print-op1
          inputs:
            parameters:
              msg:
                taskOutputParameter:
                  outputParameterKey: Output
                  producerTask: print-op1
          taskInfo:
            name: print-op2
    inputDefinitions:
      parameters:
        msg:
          parameterType: STRING
    outputDefinitions:
      artifacts:
        data:
          artifactType:
            schemaTitle: system.Artifact
            schemaVersion: 0.0.1
      parameters:
        msg:
          parameterType: STRING
  comp-print-op1:
    executorLabel: exec-print-op1
    inputDefinitions:
      parameters:
        msg:
          parameterType: STRING
    outputDefinitions:
      parameters:
        Output:
          parameterType: STRING
  comp-print-op1-2:
    executorLabel: exec-print-op1-2
    inputDefinitions:
      parameters:
        msg:
          parameterType: STRING
    outputDefinitions:
      parameters:
        Output:
          parameterType: STRING
  comp-print-op2:
    executorLabel: exec-print-op2
    inputDefinitions:
      parameters:
        msg:
          parameterType: STRING
    outputDefinitions:
      artifacts:
        data:
          artifactType:
            schemaTitle: system.Artifact
            schemaVersion: 0.0.1
...
root:
  dag:
    outputs:
      artifacts:
        Output:
          artifactSelectors:
            - outputArtifactKey: data
              producerSubtask: inner-pipeline
    tasks:
      inner-pipeline:
        cachingOptions:
          enableCache: true
        componentRef:
          name: comp-inner-pipeline
        inputs:
          parameters:
            msg:
              runtimeValue:
                constant: world
        taskInfo:
          name: inner-pipeline
      print-op1:
        cachingOptions:
          enableCache: true
        componentRef:
          name: comp-print-op1
        inputs:
          parameters:
            msg:
              componentInputParameter: msg
        taskInfo:
          name: print-op1
  inputDefinitions:
    parameters:
      msg:
        defaultValue: Hello
        isOptional: true
        parameterType: STRING
  outputDefinitions:
    artifacts:
      Output:
        artifactType:
          schemaTitle: system.Artifact
          schemaVersion: 0.0.1
schemaVersion: 2.1.0
sdkVersion: kfp-2.9.0
```

The issue here is that although the IR handles multiple instances of `print-op-1` as `comp-print-op-1` & `comp-print-op1-2`, both instances are registered in MLMD as `print-op-1`. So when we attempt to get all the executions associated with the root dag we stumble upon the duplicate task name error. I'm certain the future solution to this will also solve the `set_display_name` issue that we see in v2.

backend/src/v2/component/launcher_v2.go
zazulam: If that's the case then I'll reword it, as the screenshots of the `oneof` scenario above showcases that there are some instances of dags that seem to not be updated with this functionality. There probably is some logic that can be used to address those edge cases.

samples/v2/subdagio/__init__.py
zazulam: This was so that we can isolate/group the tests used for subdag outputs into a module. We can move the tests up level and remove the `subdagio` module if that is the preferred method.

droctothorpe: So that we can import like [this](https://github.com/kubeflow/pipelines/pull/11196/files/704d349faafd8c21cc978c9dacbda504605de912#diff-aa23d4179ee7858c47fb40841752ade320139212dcb99f9ddb37c9f00127f4f4R30) instead of importing each individual module in `sample_test.py`. If it's preferable, we can remove the init and import the modules individually or take the tests out of the subdirectory like @zazulam suggested.

samples/v2/subdagio/mixed_parameters.py
droctothorpe: The rationale behind this was that it improves the developer experience by making it really easy to kick off a **single** test run when troubleshooting without having to manipulate the contents of `sample_test.py`. I don't think there's any other context in which these modules are run as main. Let me us know if that makes sense / sounds reasonable. If not, we can remove it.

backend/src/v2/driver/driver.go
droctothorpe: Great question! We wondered the same thing (and even tried it out) but the resolution logic is so different that it results in almost no deduplication at the cost of a massive, convoluted function full of nested conditionals. 

That being said, your comment inspired us to consolidate the two config structs that are passed to the two functions. Thanks!

droctothorpe:
@chensun Thank you for the review! We know it's not exactly the easiest PR to review (in terms of scale, complexity, and convolution). Hopefully, we can further decompose `resolveInputs` in subsequent contributions to make future changes like this smaller / more incremental. 

We resolved / responded to all notes. Standing by for further instructions. Thanks!

sdk/python/kfp/compiler/pipeline_spec_builder.py
```diff
@@ -2006,13 +2008,18 @@ def convert_pipeline_outputs_to_dict(
     output name to PipelineChannel."""
     if pipeline_outputs is None:
         return {}
+    elif isinstance(pipeline_outputs, dict):
+        # This condition is required to support pipelines that return NamedTuples.
```
chensun, line 2012: Is this not covered by `isinstance(..., tuple)`?

chensun (commented):
/lgtm

sdk/python/kfp/compiler/pipeline_spec_builder.py
droctothorpe: Good call out. This comment was not clear enough. I just pushed up a [commit](https://github.com/kubeflow/pipelines/pull/11196/commits/17e0c5d9a358a251934fa24fd28e9f2298547570) that expands on it. Some of the nested pipeline tests with namedtuples that we added to the `subdagio` directory were failing until we added this handling to the IR compiler.

droctothorpe: [Here's](https://github.com/kubeflow/pipelines/blob/17e0c5d9a358a251934fa24fd28e9f2298547570/samples/v2/subdagio/multiple_parameters_namedtuple.py) a specific example. If you try to compile IR from this file, it raises the following error:

```
  return component_factory.create_component_from_func(
  File "/Users/alex/code/pipelines/validation.py", line 21, in <module>
    def core() -> NamedTuple('outputs', val1=str, val2=str):  # type: ignore
  File "/Users/alex/code/pipelines/sdk/python/kfp/dsl/pipeline_context.py", line 71, in pipeline
    return component_factory.create_graph_component_from_func(
  File "/Users/alex/code/pipelines/sdk/python/kfp/dsl/component_factory.py", line 708, in create_graph_component_from_func
    return graph_component.GraphComponent(
  File "/Users/alex/code/pipelines/sdk/python/kfp/dsl/graph_component.py", line 71, in __init__
    pipeline_spec, platform_spec = builder.create_pipeline_spec(
  File "/Users/alex/code/pipelines/sdk/python/kfp/compiler/pipeline_spec_builder.py", line 1915, in create_pipeline_spec
    pipeline_outputs_dict = convert_pipeline_outputs_to_dict(pipeline_outputs)
  File "/Users/alex/code/pipelines/sdk/python/kfp/compiler/pipeline_spec_builder.py", line 2022, in convert_pipeline_outputs_to_dict
    raise ValueError(f'Got unknown pipeline output: {pipeline_outputs}')
ValueError: Got unknown pipeline output: {'val1': {{channel:task=core-comp;name=val1;type=String;}}, 'val2': {{channel:task=core-comp;name=val2;type=String;}}}
```

Adding the `dict` conditional resolved this error. Shout out to @zazulam for identifying / correcting it.

chensun (approved):
/lgtm
/approve

Thanks!

HumairAK:
Great work folks!!

PR #11204, opened 2024-09-13
[TESTING] Testing some tests, not actual PR (yet)

CarterFendley:


.github/workflows/kubeflow-pipelines-integration-v2.yml
```diff
@@ -8,9 +8,9 @@ on:
     paths:
       - '.github/workflows/kubeflow-pipelines-integration-v2.yml'
       - 'scripts/deploy/github/**'
-      - 'samples'
-      - 'core'
-      - 'backend'
+      - 'samples/**'
+      - 'core/**'
+      - 'backend/**'
```
hbelmiro, line 13: It seems like we really need this.
Is this PR intended for that? What are you exactly testing?

PR #11209, opened 2024-09-13
Enable or Disable cache for the Compiler

diegolovison:

Support for enabling or disabling caching for the entire pipeline is being introduced. This functionality mirrors the one provided by the `client::run` method (see https://github.com/kubeflow/pipelines/blob/master/sdk/python/kfp/client/client.py#L707). The aim is to offer a similar feature using the same parameter names to prevent confusion.

Note that this PR does not address changing the default value for `structures.TaskSpec::enable_caching`. That functionality will be introduced in a separate PR and is unrelated to this one.

**Checklist:**

### DIFF
--- a/sdk/python/kfp/cli/compile_.py
+++ b/sdk/python/kfp/cli/compile_.py
@@ -22,6 +22,7 @@ from typing import Callable, Dict, Optional
 
 import click
 from kfp import compiler
+from kfp.cli.utils import parsing
 from kfp.dsl import base_component
 from kfp.dsl import graph_component
 
@@ -133,12 +134,19 @@ def parse_parameters(parameters: Optional[str]) -> Dict:
     is_flag=True,
     default=False,
     help='Whether to disable type checking.')
+@click.option(
+    '--enable-caching/--disable-caching',
+    type=bool,
+    default=None,
+    help=parsing.get_param_descr(compiler.Compiler.compile, 'enable_caching'),
+)
 def compile_(
     py: str,
     output: str,
     function_name: Optional[str] = None,
     pipeline_parameters: Optional[str] = None,
     disable_type_check: bool = False,
+    enable_caching: Optional[bool] = None,
 ) -> None:
     """Compiles a pipeline or component written in a .py file."""
     pipeline_func = collect_pipeline_or_component_func(
@@ -149,7 +157,8 @@ def compile_(
         pipeline_func=pipeline_func,
         pipeline_parameters=parsed_parameters,
         package_path=package_path,
-        type_check=not disable_type_check)
+        type_check=not disable_type_check,
+        enable_caching=enable_caching)
 
     click.echo(package_path)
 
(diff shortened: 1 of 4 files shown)


sdk/python/kfp/cli/compile_.py
```diff
@@ -133,12 +134,19 @@ def parse_parameters(parameters: Optional[str]) -> Dict:
     is_flag=True,
     default=False,
     help='Whether to disable type checking.')
+@click.option(
+    '--enable-caching/--disable-caching',
+    type=bool,
```
DharmitD, line 139: Make sure to use `is_flag` functionality here, to maintain precedence (just like `type_check` for instance).

sdk/python/kfp/cli/compile_.py
```diff
@@ -133,12 +134,19 @@ def parse_parameters(parameters: Optional[str]) -> Dict:
     is_flag=True,
     default=False,
     help='Whether to disable type checking.')
+@click.option(
+    '--enable-caching/--disable-caching',
```
DharmitD, line 138: Change this to `--disable-default-caching` to maintain precedence with other flags in this file. (like the `type_check` one)

sdk/python/kfp/cli/compile_.py
```diff
@@ -149,7 +157,8 @@ def compile_(
         pipeline_func=pipeline_func,
         pipeline_parameters=parsed_parameters,
         package_path=package_path,
-        type_check=not disable_type_check)
+        type_check=not disable_type_check,
+        enable_caching=enable_caching)
```
DharmitD, line 161: once you update as per https://github.com/kubeflow/pipelines/pull/11209/files#r1759241709, 
you would need to change this as well to something like:
`enable_caching=not disable_caching_default`

sdk/python/kfp/compiler/compiler.py
```diff
@@ -62,6 +64,12 @@ def compile(
             pipeline_name: Name of the pipeline.
             pipeline_parameters: Map of parameter names to argument values.
             type_check: Whether to enable type checking of component interfaces during compilation.
+            enable_caching: Whether or not to enable caching for the
+                run. If not set, defaults to the compile time settings, which
+                is ``True`` for all tasks by default, while users may specify
+                different caching options for individual tasks. If set, the
+                setting applies to all tasks in the pipeline (overrides the
```
DharmitD, line 71: Not sure if we want to override compile time settings, we just want to have an option to change the caching default. When we add the flag, this needs to happen for instance:

```
When we are done with this feature, this will be true:
@dsl.pipeline(name='iris-training-pipeline')
def my_pipeline():
   task_1 = create_dataset()
   task_2 = create_dataset()
   task_3 = create_dataset()
   task_3.set_caching_options(True)
   # tasks 1 and 2 don’t try to use the cache. Task 3 does try to use the cache.
```
@gregsheremeta could elaborate

PR #11097, opened 2024-08-13
feat(sdk): support dynamic machine type parameters in pipeline task setters

KevinGrantLee:
**Description of your changes:**

Enables dynamic values for cpu_limit, cpu_request, memory_limit, memory_request, accelerator_type, accelerator_limit.

Enables pipelines like
```
@dsl.component
def cpu_limit() -> str:
    return '4000m'

@dsl.component
def sum_numbers(a: int, b:int) -> int:
    return a + b


@dsl.pipeline
def pipeline(
):
    sum_numbers_task = sum_numbers(a=1, b=2)
    sum_numbers_task.set_cpu_limit(cpu_limit().output)
```

**Checklist:**
- [X] The title for your pull request (PR) should follow our title convention. [Learn more about the pull request title convention used in this repository](https://github.com/kubeflow/pipelines/blob/master/CONTRIBUTING.md#pull-request-title-convention).

PR #11240, opened 2024-09-23
chore(sdk): check that requested memory less than or equals according limits

ntny:
**Description of your changes:** raise an error with a description if the requested resources (memory/CPU) exceed the selected limit while defining the pipeline.

examples: 
```
from kfp.dsl import pipeline_task
task = pipeline_task.PipelineTask(...).set_memory_request('2M').set_memory_limit('1M')
```

### DIFF
--- a/sdk/python/kfp/dsl/pipeline_task.py
+++ b/sdk/python/kfp/dsl/pipeline_task.py
@@ -313,6 +313,16 @@ class PipelineTask:
         self._task_spec.enable_caching = enable_caching
         return self
 
+    def _ensure_resource_requests_meet_limits(self) -> None:
+        resources = self.container_spec.resources
+        if (resources.memory_request is not None and
+                resources.memory_limit is not None and
+                self._parse_memory_str_request(resources.memory_request) >
+                self._parse_memory_str_request(resources.memory_limit)):
+            raise ValueError(
+                f'Requested memory: {resources.memory_request} cannot be greater than memory limit: {resources.memory_limit}. '
+                'Check the set_memory_request and set_memory_limit parameters.')
+
     def _ensure_container_spec_exists(self) -> None:
         """Ensures that the task has a container spec."""
         caller_method_name = inspect.stack()[1][3]
@@ -453,6 +463,48 @@ class PipelineTask:
             category=DeprecationWarning)
         return self.set_accelerator_limit(gpu)
 
+    def _parse_memory_str_request(self, memory_str: str) -> float:
+        memory_request = float(0)
+        if memory_str.endswith('E'):
+            memory_request = float(
+                memory_str[:-1]) * constants._E / constants._G
+        elif memory_str.endswith('Ei'):
+            memory_request = float(
+                memory_str[:-2]) * constants._EI / constants._G
+        elif memory_str.endswith('P'):
+            memory_request = float(
+                memory_str[:-1]) * constants._P / constants._G
+        elif memory_str.endswith('Pi'):
+            memory_request = float(
+                memory_str[:-2]) * constants._PI / constants._G
+        elif memory_str.endswith('T'):
+            memory_request = float(
+                memory_str[:-1]) * constants._T / constants._G
+        elif memory_str.endswith('Ti'):
+            memory_request = float(
+                memory_str[:-2]) * constants._TI / constants._G
+        elif memory_str.endswith('G'):
+            memory_request = float(memory_str[:-1])
+        elif memory_str.endswith('Gi'):
+            memory_request = float(
+                memory_str[:-2]) * constants._GI / constants._G
+        elif memory_str.endswith('M'):
+            memory_request = float(
+                memory_str[:-1]) * constants._M / constants._G
+        elif memory_str.endswith('Mi'):
+            memory_request = float(
+                memory_str[:-2]) * constants._MI / constants._G
+        elif memory_str.endswith('K'):
+            memory_request = float(
+                memory_str[:-1]) * constants._K / constants._G
+        elif memory_str.endswith('Ki'):
+            memory_request = float(
+                memory_str[:-2]) * constants._KI / constants._G
+        else:
+            # By default interpret as a plain integer, in the unit of Bytes.
+            memory_request = float(memory_str) / constants._G
+        return memory_request
+
     def _validate_memory_request_limit(self, memory: str) -> str:
         """Validates memory request/limit string and converts to its numeric
         string value.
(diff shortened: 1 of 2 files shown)


HumairAK:
Looks like there are sdk ci failures, @ntny can you take a look?

ntny:
> Looks like there are sdk ci failures, @ntny can you take a look?

Sure, in a few days.

ntny:
For validation purposes, I had to restore the code [fragment ](https://github.com/kubeflow/pipelines/pull/11097/files#diff-0ab2e518eb07ebf6d4e8d5ec14746dd60b0602ec0ea0e16c7cdbe2d825820350L456)for converting requested resource memory labels to float (for comparing limits and requests).

ntny:
CI Check failed due to timeout.
I suppose it might be unrelated to MR

gregsheremeta:
Thanks for the PR.

This kind of thing -- baking Kubernetes logic into KFP -- makes be a bit nervous. I already don't like that the kfp-kubernetes SDK basically "chases" the [official python client](https://github.com/kubernetes-client/python), and the slope here seems slippery.

Using a different example, today we allow mounting a secret as a volume. Is it KFP's responsibility to make sure that the secret name is a valid Kubernetes name, or should we just rely on the Kubernetes server to deal with that? Going further, should we be pre-checking that the secret exists, or should we just let Kubernetes tell us that the volume mount failed at pipeline run time?

I've been mostly thinking we should just leave these sorts of things up to Kubernetes, and treat kfp-kubernetes as basically a "dumb" yaml generator.

This change seems fairly innocuous, but it's more code to maintain, and we're on the hook if the Kubernetes behavior changes someday.

I'd be a little  more receptive to the change if we could somehow use the [official python client](https://github.com/kubernetes-client/python). Did you by any chance look into that?

What do others think?

ntny:
> Thanks for the PR.
> 
> This kind of thing -- baking Kubernetes logic into KFP -- makes me a bit nervous. I already don't like that the kfp-kubernetes SDK basically "chases" the [official python client](https://github.com/kubernetes-client/python), and the slope here seems slippery.
> 
> Using a different example, today we allow mounting a secret as a volume. Is it KFP's responsibility to make sure that the secret name is a valid Kubernetes name, or should we just rely on the Kubernetes server to deal with that? Going further, should we be pre-checking that the secret exists, or should we just let Kubernetes tell us that the volume mount failed at pipeline run time?
> 
> I've been mostly thinking we should just leave these sorts of things up to Kubernetes, and treat kfp-kubernetes as basically a "dumb" yaml generator.
> 
> This change seems fairly innocuous, but it's more code to maintain, and we're on the hook if the Kubernetes behavior changes someday.
> 
> I'd be a little more receptive to the change if we could somehow use the [official python client](https://github.com/kubernetes-client/python). Did you by any chance look into that?
> 
> What do others think?

Thanks, i agree with your points. My reasoning for this PR:
1. pipeline_task already performs similar memory checks on input parameters and it is also  k8s specific validation https://github.com/kubeflow/pipelines/blob/master/sdk/python/kfp/dsl/pipeline_task.py#L457 . I am not sure my check adds anything different.
2. Requesting resources beyond the limits looks abnormal, and this behavior isn’t specific to Kubernetes in my view. It seems to me that any API providing request/limit options should handle this validation. 

But overall, I agree with your arguments about maintaining the code, and it’s up to you to decide here.
Regarding the Python-K8s client, thanks, I will take a look at it

gregsheremeta:
@chensun do you have any thoughts on the above two comments? I'm not sure how much "Kubernetes logic" (for lack of a better term) we should be baking into KFP. Anton has some good points.

ntny:
Hi @gregsheremeta!
Regarding the use of the official Python client for validation at the KFP level, I might be mistaken, but in my view, it is not suitable here. This is because KFP generates its own specification (or worked directly with Argo in previous versions) and does not operate within the Kubernetes context.
The official client, on the other hand, works with Kubernetes entities.
Plus official Python client supports a dry-run pattern as an alternative for validation. I'm not sure if this would be useful for the entire pipeline, but it should be implemented somewhere in the KFP driver in my account.

chensun:
> I've been mostly thinking we should just leave these sorts of things up to Kubernetes, and treat kfp-kubernetes as basically a "dumb" yaml generator.

I personally lean towards keeping SDK as a "dumb" yaml generator, and not baking too much logic into it. Otherwise, SDK might become a bottleneck, and we would need to keep chasing trivial changes from Kubernetes side. 

> pipeline_task already performs similar memory checks on input parameters and it is also k8s specific validation https://github.com/kubeflow/pipelines/blob/master/sdk/python/kfp/dsl/pipeline_task.py#L457 . I am not sure my check adds anything different.

This check is something we inherited from KFP v1 codebase, it's probably much easier to keep it as-is then reasoning the other way around. 
While I agree that client side validation is nice to have and catching error earlier is often desirable UX, I do want to point out that such check has limited usage and the potential downside could be inconsistent UX. For input parameter validation, SDK can only check when the value is available at compilation or submission time, which is not always the case. For example,

```python
some_component().set_memory_limit('8G') # SDK validation can check and throw early error on incorrect format

upstream_task = upstream()
downstream().set_memory_limit(upstream_task.output) # SDK validation cannot do anything with the runtime value
```

I personally prefer having consistent behavior and error message on both cases.

chensun:
> > I've been mostly thinking we should just leave these sorts of things up to Kubernetes, and treat kfp-kubernetes as basically a "dumb" yaml generator.
> 
> I personally lean towards keeping SDK as a "dumb" yaml generator, and not baking too much logic into it. Otherwise, SDK might become a bottleneck, and we would need to keep chasing trivial changes from Kubernetes side.
> 
> > pipeline_task already performs similar memory checks on input parameters and it is also k8s specific validation https://github.com/kubeflow/pipelines/blob/master/sdk/python/kfp/dsl/pipeline_task.py#L457 . I am not sure my check adds anything different.
> 
> This check is something we inherited from KFP v1 codebase, it's probably much easier to keep it as-is then reasoning the other way around. While I agree that client side validation is nice to have and catching error earlier is often desirable UX, I do want to point out that such check has limited usage and the potential downside could be inconsistent UX. For input parameter validation, SDK can only check when the value is available at compilation or submission time, which is not always the case. For example,
> 
> ```python
> some_component().set_memory_limit('8G') # SDK validation can check and throw early error on incorrect format
> 
> upstream_task = upstream()
> downstream().set_memory_limit(upstream_task.output) # SDK validation cannot do anything with the runtime value
> ```
> 
> I personally prefer having consistent behavior and error message on both cases.

HumairAK:
I think this was accidentally reopened? Based on the discussion here and in the community call a few weeks ago, we have decided to keep this level of validation for k8s resources out of the sdk.

PR #10798, opened 2024-05-07
feat(backend): Add Parallelism Limit to ParallelFor tasks. Fixes #8718

gmfrasca:
**Description of your changes:**
Fixes #8718

Adds the `Parallelism` item to a DAG template if specified by a task's IteratorPolicy (ie ParallelFor w/ a parallelism limit).

**Checklist:**
- [x] The title for your pull request (PR) should follow our title convention. [Learn more about the pull request title convention used in this repository](https://github.com/kubeflow/pipelines/blob/master/CONTRIBUTING.md#pull-request-title-convention).

PR #11243, opened 2024-09-24
fix(backend): randomizing output uri path to avoid overwriting. Fixes #10186

### LINKED ISSUE #10186: [backend] Artifacts of Components in ParallelFor / Sub-DAGs Overwritten by Concurrent Iterations
### Environment

Using AWS.

* KFP version:
1.8-rc5

* KFP SDK version:

```python
kfp-kubernetes             1.0.0                  
kfp-server-api             2.0.2
```

### Steps to reproduce

The problem occurs when we run the same component across ParallelFor iterations OR sub-dags. I discovered the error due to observing unexpected behavior when running nested parallelfor loops, although it originates from the sub-dag structure itself and is not specific to ParallelFor loops. 

What happens is that each component overwrites the artifacts of all other components created from the same component template / function because the auto-generated path for S3 is the same across ParallelFor iterations / sub-dags . The path is `s3://{bucket}/{pipeline_name}/{run_id}/{component_name}/{output_path_name} `. This means that if you run the same component for each iteration of the ParallelFor or across duplicate sub-dags, it overwrites itself in S3 (at least that's my theory).

The deceptive part of this is that there is no explicit failure, but rather unexpected results originating from this overwriting of artifacts. Some examples of how this can happen are as follows. My general use-case to test this is simply creating a dataframe with a column filled with a passed string value. You can then see in the input/output artifact preview tab if the artifact has the correct column. For example, in the screenshot below, the value passed for this particular iteration of a ParallelFor is `b`, but the value in the displayed output artifact is `a`.

Screen Shot 2023-10-30 at 6 20 38 PM

**SUB-DAGS**

The compiler to recreate the problem in sub-dags is as follows:
```python
@dsl.component(packages_to_install=['pandas'])
def parent(val: str, output_path: Output[Dataset]):
    import pandas as pd
    data = pd.DataFrame({"a": [val] * 3, "b": [1,2,3]})
    print("data:\n", data, flush=True)
    data.to_csv(output_path.path, index=False)

@dsl.pipeline(pipeline_root="s3://beta-kf-1-8-test")
def sub_dag(val: str):
    parent_component = parent(val=val)
    parent_component.set_caching_options(False)
    parent_component.set_env_variable(name="AWS_REGION", value="us-east-1")

@dsl.pipeline(pipeline_root="s3://beta-kf-1-8-test")
def compile_pipeline(val_1: str = "a", val_2: str = "b"):
    sub_dag(val=val_1)
    sub_dag(val=val_2)
```

**ParallelFor**

The compiler to create the problem in ParallelFor is as follows.
```python
@dsl.component(packages_to_install=['pandas'])
def parent(val: str, output_path: Output[Dataset]):
    import pandas as pd
    data = pd.DataFrame({"a": [val] * 3, "b": [1,2,3]})
    data.to_csv(output_path.path, index=False)

@dsl.pipeline(pipeline_root="s3://beta-kf-1-8-test")
def compile_pipeline(l: list = ["a", "b", "c"]):
    with dsl.ParallelFor(l) as val:
        parent_component = parent(val=val)
        parent_component.set_caching_options(False)
        parent_component.set_env_variable(name="AWS_REGION", value="us-east-1")
```

**Duplicate Components Same Level**
I did test and confirm that duplicating the same component in the same-level of a DAG did not run into this issue as the auto-generated output paths were unique. The compiler for this test is here:
```python
@dsl.component(packages_to_install=['pandas'])
def parent(val: str, output_path: Output[Dataset]):
    import pandas as pd
    data = pd.DataFrame({"a": [val] * 3, "b": [1,2,3]})
    data.to_csv(output_path.path, index=False)

@dsl.pipeline(pipeline_root="s3://beta-kf-1-8-test")
def compile_pipeline(val_1: str = "a", val_2: str = "b"):
    parent_component = parent(val=val_1)
    parent_component.set_caching_options(False)
    parent_component.set_env_variable(name="AWS_REGION", value="us-east-1")

    parent_component_2 = parent(val=val_2)
    parent_component_2.set_caching_options(False)
    parent_component_2.set_env_variable(name="AWS_REGION", value="us-east-1")
```

This issue essentially makes each iteration of a ParallelFor / sub-dag non-unique and not dependent on the specific inputs provided, and the artifacts generated from it non-determinstic. I believe it may be related to the fix introduced in this [PR](https://github.com/kubeflow/pipelines/issues/10149) by @chensun from last week.

### Expected result

The ParallelFor / sub-dag structures should not overwrite other components executed in concurrent ParallelFor iterations / sub-dags. We should be able to refer to the output artifact of a component within an iteration of a ParallelFor or sub-dag and retrieve the artifact that was generated within the same iteration (rather than a parallel one that has different args and thus different outputs).

---

b4sus:
In driver, random string is added when uri paths for output artifacts are generated. This should ensure that when component of certain name is executed in parallel (either with ParallelFor or just simply calling it multiple times in @pipeline), its outputs are always stored to different paths.

### DIFF
--- a/backend/src/v2/driver/driver.go
+++ b/backend/src/v2/driver/driver.go
@@ -272,7 +272,7 @@ func Container(ctx context.Context, opts Options, mlmd *metadata.Client, cacheCl
 		execution.Condition = &willTrigger
 	}
 	if execution.WillTrigger() {
-		executorInput.Outputs = provisionOutputs(pipeline.GetPipelineRoot(), opts.Task.GetTaskInfo().GetName(), opts.Component.GetOutputDefinitions())
+		executorInput.Outputs = provisionOutputs(pipeline.GetPipelineRoot(), opts.Task.GetTaskInfo().GetName(), opts.Component.GetOutputDefinitions(), uuid.NewString())
 	}
 
 	ecfg, err := metadata.GenerateExecutionConfig(executorInput)
@@ -1227,7 +1227,7 @@ func resolveInputs(ctx context.Context, dag *metadata.DAG, iterationIndex *int, 
 	return inputs, nil
 }
 
-func provisionOutputs(pipelineRoot, taskName string, outputsSpec *pipelinespec.ComponentOutputsSpec) *pipelinespec.ExecutorInput_Outputs {
+func provisionOutputs(pipelineRoot, taskName string, outputsSpec *pipelinespec.ComponentOutputsSpec, outputUriSalt string) *pipelinespec.ExecutorInput_Outputs {
 	outputs := &pipelinespec.ExecutorInput_Outputs{
 		Artifacts:  make(map[string]*pipelinespec.ArtifactList),
 		Parameters: make(map[string]*pipelinespec.ExecutorInput_OutputParameter),
@@ -1239,7 +1239,7 @@ func provisionOutputs(pipelineRoot, taskName string, outputsSpec *pipelinespec.C
 				{
 					// Do not preserve the query string for output artifacts, as otherwise
 					// they'd appear in file and artifact names.
-					Uri:      metadata.GenerateOutputURI(pipelineRoot, []string{taskName, name}, false),
+					Uri:      metadata.GenerateOutputURI(pipelineRoot, []string{taskName, outputUriSalt, name}, false),
 					Type:     artifact.GetArtifactType(),
 					Metadata: artifact.GetMetadata(),
 				},


hbelmiro (commented):
/lgtm
/ok-to-test

HumairAK:
Hey @b4sus , thanks for the contribution! 

Can you provide a sample pipeline that illustrates the issue this pr is aiming to resolve? 

At least in the case of a component being re-used, I believe the taskname will have a `-#` suffix, so _should_ already be distinguished from repeated earlier calls. For parallelFor I'd be interested of its impact with https://github.com/kubeflow/pipelines/pull/10798 

cc @gmfrasca

gmfrasca:
@HumairAK - This appears to only impact output artifacts, and only changes the driver behavior when in CONTAINER driver mode, so I don't believe this should have any effect on #10798 in terms of sub-DAG naming schemes, etc.

With that said, I did see that ParallelFor outputs are storing artifacts in the same URI, which is a problem that this PR addresses by adding UUID salts.

gmfrasca (commented):
Tested this out using a ParallelFor task and confirmed each iteration's output artifacts are given unique URIs which are referenced properly in KFP UI.

/lgtm

b4sus:
Hey @HumairAK ,

We noticed the problem when, from one pipeline, we started other pipelines (pipeline as component) using ParallelFor. This is roughly the code:
```python
@dsl.pipeline
def inner_pipeline(date_to_process: str):
    comp1_task = component1(date_to_process = date_to_process)
    comp2_task = component2(comp1_task.outputs["output_df"])

@dsl.pipeline
def main_pipeline(from_date: str, to_date: str):
    prepare_dates_task = prepare_dates_component(from_date = from_date, to_date = to_date)

    with dsl.ParallelFor(items = prepare_dates_task.output, parallelism=4) as date_to_process:
        inner_ppln_task = inner_pipeline(date_to_process = date_to_process)
```

In this case, many inner pipelines were started (more then 4 as parallelism is not yet supported) and problem was that output of component1 was/is written to the same minio location, so overwriting each other. And subsequently couple of component2 tasks get the same input, regardless of the argument (date_to_process), producing the same final output (not visible here in code as it is store directly in component).

HumairAK:
Perfect, thanks guys

tested and works as well: 

before:
Pasted image 20241003162455

after:
image


/lgtm
/approve

wassimbensalem:
> Perfect, thanks guys
> 
> tested and works as well with the following pipeline:
> 
> pipeline.py
> ```python
> from typing import List
> 
> from kfp import dsl, compiler
> from kfp.dsl import Dataset
> from kfp.dsl import Output, InputPath
> 
> @dsl.component(base_image="quay.io/opendatahub/ds-pipelines-ci-executor-image:v1.0")
> def component1(date_to_process: str, output_df: Output[Dataset]):
>     with open(output_df.path, 'w') as f:
>         f.write(date_to_process)
> 
> @dsl.component(base_image="quay.io/opendatahub/ds-pipelines-ci-executor-image:v1.0")
> def component2(dataset_in: InputPath('Dataset')):
>     with open(dataset_in, 'r') as input_file:
>         dataset_one_contents = input_file.read()
>     print(dataset_one_contents)
> 
> @dsl.component(base_image="quay.io/opendatahub/ds-pipelines-ci-executor-image:v1.0")
> def prepare_dates_component() -> List[str]:
>     return ["1", "2", "3", "4", "5", "6"]
> 
> @dsl.pipeline
> def inner_pipeline(date_to_process: str):
>     comp1_task = component1(date_to_process = date_to_process).set_caching_options(enable_caching=False)
>     comp2_task = component2(dataset_in = comp1_task.outputs["output_df"]).set_caching_options(enable_caching=False)
> 
> @dsl.pipeline
> def main_pipeline():
>     prepare_dates_task = prepare_dates_component().set_caching_options(enable_caching=False)
> 
>     with dsl.ParallelFor(items = prepare_dates_task.output, parallelism=4) as date_to_process:
>         inner_ppln_task = inner_pipeline(date_to_process = date_to_process)
> 
> 
> if __name__ == '__main__':
>     compiler.Compiler().compile(main_pipeline, __file__ + '.yaml')
> ```
> 
> before: Pasted image 20241003162455
> 
> after: image
> 
> /lgtm /approve

Should we use the exact same base image ? I ran it without that image and it failed. I also cannot read the artifacts in a loop, any idea why ?

HumairAK:
@wassimbensalem I think you might me encountering a couple of different issues, can you reach out in the cncf k8s platform slack with the errors you are encountering, for more context around this base image you can find the dockerfile [here](https://github.com/opendatahub-io/data-science-pipelines-operator/blob/main/docs/example_pipelines/iris/Dockerfile) 

 this sounds related to [this isssue](https://github.com/kubeflow/pipelines/pull/11627) which was recently resolved in master branch, let us know in slack if that is indeed the case, if not I suggest creating a new issue with a reproducible pipeline and information around the platform you are using as well as kfp/ kfp sdk / k8s version.

PR #11273, opened 2024-10-07
chore: Updating Pytorch-Launcher component to work with pipelines v2

Fiona-Waters:
This PR will resolve https://github.com/kubeflow/training-operator/issues/2068 
I have updated the pytorch launcher component to use v2 constructs.
I have also updated the pytorch launcher component to use kubeflow training-operator **TrainingClient.**

**Checklist:**


Fiona-Waters:
@terrytangyuan maybe you could help with the training-operator related error I am getting here. Any advice would be really appreciated. Thank you.

terrytangyuan:
Can you print out `job` object? It's supposed to be an Job/CRD object with fields like `kind`.

Fiona-Waters:
> Can you print out `job` object? It's supposed to be an Job/CRD object with fields like `kind`.

Sure. Running it quickly locally it looks like this: 
```
job {'api_version': 'kubeflow.org/v1',
 'kind': 'PyTorchJob',
 'metadata': {'annotations': None,
              'creation_timestamp': None,
              'deletion_grace_period_seconds': None,
              'deletion_timestamp': None,
              'finalizers': None,
              'generate_name': None,
              'generation': None,
              'labels': None,
              'managed_fields': None,
              'name': 'pytorchjob',
              'namespace': 'kubeflow',
              'owner_references': None,
              'resource_version': None,
              'self_link': None,
              'uid': None},
 'spec': {'elastic_policy': None,
          'nproc_per_node': None,
          'pytorch_replica_specs': {'Master': {}, 'Worker': {}},
          'run_policy': {'active_deadline_seconds': None,
                         'backoff_limit': None,
                         'clean_pod_policy': 'Running',
                         'scheduling_policy': None,
                         'suspend': None,
                         'ttl_seconds_after_finished': None}},
 'status': None}
```
I can try to log it out when running on KinD later today.

components/kubeflow/pytorch-launcher/src/launch_pytorchjob.py
```diff
@@ -29,16 +30,16 @@ def get_current_namespace():
 
 
 # Patch PyTorchJob APIs to align with k8s usage
-class V1PyTorchJob(V1PyTorchJob_original):
+class V1PyTorchJob(KubeflowOrgV1PyTorchJob):
     def __init__(self, *args, **kwargs):
         super().__init__(*args, **kwargs)
-        self.openapi_types = self.swagger_types
+        # self.openapi_types = self.swagger_types
```
terrytangyuan, line 36: Why is this removed?

components/kubeflow/pytorch-launcher/src/launch_pytorchjob.py
```diff
@@ -116,33 +119,34 @@ def main(args):
         ),
         spec=jobSpec,
     )
+  #  print(job)
+  #  serialized_job = k8s_client.ApiClient().sanitize_for_serialization(job)
+  #  print(serialized_job)
```
terrytangyuan, line 124: I think commenting this out causes the issue. The JSON/dict object needs to be serialized to a job objection.

components/kubeflow/pytorch-launcher/src/launch_pytorchjob.py
Fiona-Waters: I moved to importing KubeflowOrgV1PyTorchJob from kubeflow.training and this does not have a swagger_types attribute. With the previous `from kubeflow.pytorchjob import V1PyTorchJob` there was an error around using floats and ints within the numpy config file.
```
AttributeError: module 'numpy' has no attribute 'float'.
`np.float` was a deprecated alias for the builtin `float`. To avoid this error in existing code, use `float` by itself. Doing this will not modify any behavior and is safe. If you specifically wanted the numpy scalar type, use `np.float64` here.
```

components/kubeflow/pytorch-launcher/src/launch_pytorchjob.py
Fiona-Waters: I've tried it both ways (serialized and not serialized) but unfortunately am still getting the same error.
```
time="2024-10-09T07:42:21.764Z" level=info msg="capturing logs" argo=true
I1009 07:42:21.785974      29 cache.go:116] Connecting to cache endpoint 10.96.101.74:8887
  File "/ml/src/launch_pytorchjob.py", line 202, in <module>
    main(args)
  File "/ml/src/launch_pytorchjob.py", line 182, in main
    training_client.create_job(pytorch_job)
  File "/usr/local/lib/python3.11/site-packages/kubeflow/training/api/training_client.py", line 399, in create_job
    job_kind = str(job.kind)
                   ^^^^^^^^
AttributeError: 'dict' object has no attribute 'kind'
I1009 07:42:22.532704      29 launcher_v2.go:150] publish success.
F1009 07:42:22.532777      29 main.go:49] failed to execute component: exit status 1
time="2024-10-09T07:42:22.765Z" level=info msg="sub-process exited" argo=true error="<nil>"
```

Fiona-Waters:
> > Can you print out `job` object? It's supposed to be an Job/CRD object with fields like `kind`.
> 
> Sure. Running it quickly locally it looks like this:
> 
> ```
> job {'api_version': 'kubeflow.org/v1',
>  'kind': 'PyTorchJob',
>  'metadata': {'annotations': None,
>               'creation_timestamp': None,
>               'deletion_grace_period_seconds': None,
>               'deletion_timestamp': None,
>               'finalizers': None,
>               'generate_name': None,
>               'generation': None,
>               'labels': None,
>               'managed_fields': None,
>               'name': 'pytorchjob',
>               'namespace': 'kubeflow',
>               'owner_references': None,
>               'resource_version': None,
>               'self_link': None,
>               'uid': None},
>  'spec': {'elastic_policy': None,
>           'nproc_per_node': None,
>           'pytorch_replica_specs': {'Master': {}, 'Worker': {}},
>           'run_policy': {'active_deadline_seconds': None,
>                          'backoff_limit': None,
>                          'clean_pod_policy': 'Running',
>                          'scheduling_policy': None,
>                          'suspend': None,
>                          'ttl_seconds_after_finished': None}},
>  'status': None}
> ```
> 
> I can try to log it out when running on KinD later today.

This is what the serialized job looks like. Kind is present.
```
{'apiVersion': 'kubeflow.org/v1', 'kind': 'PyTorchJob', 'metadata': {'name': 'pytorchjob', 'namespace': 'kubeflow'}, 'spec': {'pytorchReplicaSpecs': {'Master': {}, 'Worker': {}}, 'runPolicy': {'cleanPodPolicy': 'Running'}}}.
```

components/kubeflow/pytorch-launcher/src/launch_pytorchjob.py
Fiona-Waters: I've removed these functions altogether as they are no longer required.

components/kubeflow/pytorch-launcher/src/launch_pytorchjob.py
Fiona-Waters: I've been able to resolve this error. Thank you for your help.

Fiona-Waters:
@HumairAK would really appreciate if you could review this if/when you have time, please.

components/kubeflow/pytorch-launcher/requirements.txt
```diff
@@ -1,4 +1,6 @@
 pyyaml
 kubernetes
 kubeflow-pytorchjob
+kubeflow.training
```
terrytangyuan, line 4: Is there minimal required version?

components/kubeflow/pytorch-launcher/src/launch_pytorchjob.py
```diff
@@ -116,33 +107,31 @@ def main(args):
...
+    # remove one of these depending on where you are running this
     config.load_incluster_config()
-    api_client = k8s_client.ApiClient()
-    launcher_client = launch_crd.K8sCR(
-        group=args.jobGroup,
-        plural=args.jobPlural,
-        version=args.version,
-        client=api_client
-    )
+    #config.load_kube_config()
```
terrytangyuan, line 114: This might be needed?

components/kubeflow/pytorch-launcher/sample.py
```diff
@@ -66,124 +103,96 @@ def create_worker_spec(
             },
         }
 
-    worker_spec_output = namedtuple(
-        "MyWorkerOutput", ["worker_spec"]
+    return worker
+
+# container component description setting inputs and implementation
+@dsl.container_component
+def pytorch_job_launcher(
```
terrytangyuan, line 110: I'd imagine code like would be inside the /src directory instead of sample.py file but I am not familiar enough with KFP codebase to support that.

components/kubeflow/pytorch-launcher/src/launch_pytorchjob.py
```diff
@@ -28,18 +29,6 @@ def get_current_namespace():
...
-    def __init__(self, *args, **kwargs):
-        super().__init__(*args, **kwargs)
-        self.openapi_types = self.swagger_types
-
-
-class V1PyTorchJobSpec(V1PyTorchJobSpec_original):
-    def __init__(self, *args, **kwargs):
-        super().__init__(*args, **kwargs)
-        self.openapi_types = self.swagger_types
-
```
terrytangyuan, line 42: Are these no longer needed? Will this change break existing usage?

terrytangyuan:
Thank you for your work on this! I left some comments.

components/kubeflow/pytorch-launcher/sample.py
Fiona-Waters: This sample implementation of a pipeline, uses the image in the pytorch job container and runs launch_pytorchjob.py from the /src directory using the parameters etc that are set out, afaiu.

components/kubeflow/pytorch-launcher/src/launch_pytorchjob.py
Fiona-Waters: Here I have included both `config.load_incluster_config()` and `config.load_kube_config()` for use depending on how it will be run. I can update this to what ever makes most sense for users. This is what is included currently:
```
 # remove one of these depending on where you are running this
    config.load_incluster_config()
    #config.load_kube_config()
```

components/kubeflow/pytorch-launcher/src/launch_pytorchjob.py
Fiona-Waters: I moved to using KubeflowOrgV1PyTorchJob and KubeflowOrgV1PyTorchJobSpec  from kubeflow-training to be inline with the training operator, therefore these classes are not required. I am not sure exactly how to test if it will break existing usage, or if this is relevant - any advice on this would be appreciated. @HumairAK @rimolive could you please advise?

components/kubeflow/pytorch-launcher/requirements.txt
Fiona-Waters: Thank you. I've updated the requirements.txt to include more specific dependencies - see this [commit](https://github.com/kubeflow/pipelines/pull/11273/commits/6d980bd317b5b8285c3a2d4bd83c02983c8ae993)

rimolive:
Thanks for you contribution, @Fiona-Waters !

/lgtm

components/contrib/kubeflow/pytorch-launcher/sample.py
```diff
@@ -66,124 +103,96 @@ def create_worker_spec(
...
+        '--cleanPodPolicy', clean_pod_policy,]
+    if active_deadline_seconds is not None and isinstance(active_deadline_seconds, int):
+        command_args.append(['--activeDeadlineSeconds', str(active_deadline_seconds)])
+    if backoff_limit is not None and isinstance(backoff_limit, int):
+        command_args.append(['--backoffLimit', str(backoff_limit)])
+    if ttl_seconds_after_finished is not None and isinstance(ttl_seconds_after_finished, int):
+        command_args.append(['--ttlSecondsAfterFinished', str(ttl_seconds_after_finished)])
+    
+    return dsl.ContainerSpec(
+        image='quay.io/rh_ee_fwaters/kubeflow-pytorchjob-launcher:v2',
```
HumairAK, line 142: Hrm, not sure if we should be referencing users individual quay accounts here, but I don't know how we have typically handled such cases in the past. 

Options are: 

1. We add a workflow that builds this image and uploads it to ghcr for the kubeflow repo 
2. We just add a placeholder here, and leave a comment telling the user they need to build/upload this themselves 

since we have the build_image.sh script here, I'm leaning towards (2) and leaving a comment to run this script and replace with the repo here

components/contrib/kubeflow/pytorch-launcher/Dockerfile
HumairAK, line 1: A general comment about the location

I'm not quite sure what the historical context of the `contrib` folder was, but given [this PR](https://github.com/kubeflow/pipelines/tree/master/components/contrib) by @chensun, it makes me think that we're just dumping old unmaintained component files here. So my current thinking is that we should restore these specific files back to the  /components/kubeflow/pytorch-launcher/* location

HumairAK (changes requested):

components/kubeflow/pytorch-launcher/src/launch_pytorchjob.py
HumairAK: I would ignore concerns around breaking compatibility, as this component has not been updated in a really long time (and was thus moved to contrib as a result).

components/kubeflow/pytorch-launcher/sample.py
HumairAK: yes that is correct

components/kubeflow/pytorch-launcher/src/launch_pytorchjob.py
HumairAK: I think the current approach makes sense, the pod is running in a container, on k8s, so it makes sense we'd leverage the pod's SA's creds here.

HumairAK:
some nits above, otherwise happy to approve once addressed

Shreyanand (changes requested):

components/contrib/kubeflow/pytorch-launcher/sample.py
Shreyanand: Another option could be to remove the image completely. We could define a lightweight component with [packages_to_install](https://www.kubeflow.org/docs/components/pipelines/user-guides/components/lightweight-python-components/#packages_to_install) and add the `launch_pytorchjob.py` code in there. This way we don't have to build and maintain the image. Happy to be corrected if I'm missing something.

components/contrib/kubeflow/pytorch-launcher/Dockerfile
Fiona-Waters: Thank you for the review @HumairAK 
Yes this sounds good to me. I'll move the files back there.

components/contrib/kubeflow/pytorch-launcher/sample.py
rimolive: So from my understanding this is a sample code. If it's a sample code created by a contributor, I don't see a problem.

My main concern here is if we begin maintaining images from samples it will be a pain for KFP having so many images.

Fiona-Waters: I am happy to update it, and add a comment with instructions on using the build script. Is this sufficient or do we want to define a lightweight component which includes the contents of the launch_pytorchjob.py script instead?

Fiona-Waters: Updated as per point 2. above. https://github.com/kubeflow/pipelines/pull/11273/commits/1c9ae31d90af072b085a3d5d309bcd437eea0763

Shreyanand: Here's what I would recommend doing in sample.py:

1. Remove the command arg from pytorch_job_launcher component and directly add the [main](https://github.com/kubeflow/pipelines/blob/1c9ae31d90af072b085a3d5d309bcd437eea0763/components/kubeflow/pytorch-launcher/src/launch_pytorchjob.py#L81) function code from `launch_pytorchjob.py`. This would remove the argument parsers in `launch_pytorchjob.py` thus simplifying it further. It would also eliminate the need for the user to build an image which is not required for just installing `kubeflow-training`. 
2. Remove the `create_master_spec` and `create_worker_spec` as they don't need be their own component. Use util functions like [1](https://github.com/kubeflow/training-operator/blob/master/sdk/python/kubeflow/training/api/training_client.py#L494) and [2](https://github.com/kubeflow/training-operator/blob/master/sdk/python/kubeflow/training/api/training_client.py#L469) to create these specs directly in the `pytorch_job_launcher` container component.

Fiona-Waters: Please see latest commit https://github.com/kubeflow/pipelines/pull/11273/commits/dbea49b17b1168487a38929ea2ee9c1e019a8453 with refactor as suggested. @Shreyanand @HumairAK @rimolive 
Thanks for the help @Shreyanand :)

Shreyanand: This looks so much cleaner! Thanks for making the changes @Fiona-Waters 💯 I'm gonna test it out but from the looks of it, this should be good to go.

Shreyanand (approved):
/lgtm

The latest commits address all the requested changes.

HumairAK:
Thanks @Fiona-Waters! This is great! 

/lgtm
/approve

Issue #11038, opened 2024-07-25
[bug] SDK Test Failing


DharmitD:
This SDK test is failing: https://github.com/kubeflow/pipelines/blob/master/sdk/python/kfp/cli/component_test.py#L579

```

=================================== FAILURES ===================================
_____________ Test.test_dockerfile_can_contain_custom_kfp_package ______________

self = <kfp.cli.component_test.Test testMethod=test_dockerfile_can_contain_custom_kfp_package>

    def test_dockerfile_can_contain_custom_kfp_package(self):
        component = _make_component(
            func_name='train', target_image='custom-image')
        _write_components('components.py', component)
        package_dir = os.path.dirname(os.path.dirname(self.current_dir))
    
        # suppresses large stdout from subprocess that builds kfp package
        with mock.patch.object(
                subprocess,
                'run',
                new=functools.partial(subprocess.run, capture_output=True)):
            result = self.runner.invoke(
                self.cli,
                [
                    'build',
                    str(self._working_dir), f'--kfp-package-path={package_dir}'
                ],
            )
>       self.assertEqual(result.exit_code, 0)
E       AssertionError: 1 != 0

sdk/python/kfp/cli/component_test.py:597: AssertionError
------------------------------ Captured log call -------------------------------
ERROR    root:component.py:122 Failed to find built KFP wheel under /tmp/tmpbrsaqwfx
=============================== warnings summary ===============================
```

### Steps to reproduce

- run the following command from the repo's root directory: `pytest sdk/python/kfp --ignore=sdk/python/kfp/deprecated --cov=kfp`

/area testing

PR #11285, opened 2024-10-10
test: Fix failing SDK test by replacing mock.patch with subprocess for actual build

DharmitD:
Resolves: https://github.com/kubeflow/pipelines/issues/11038 

- Removed the mock.patch on subprocess.run.
- Added a direct call to subprocess.run to build the KFP package.
- Introduced error handling with a helpful error message if the package build fails.
- Kept the rest of the test logic intact.

The original mock.patch.object wasn’t working because it prevented the real build from happening, which was necessary for the test to pass. Replacing it with a direct call to subprocess.run ensures the KFP package is actually built, resolving the issue. This approach allows the test to reflect the real build environment and provides clearer error reporting if something goes wrong during the build process.

SDK test GHA passed on this PR: https://github.com/kubeflow/pipelines/actions/runs/11284347350/job/31385382544?pr=11285 
Alternatively, you could test locally by running the following command from the repo's root directory: 
```
pytest sdk/python/kfp --ignore=sdk/python/kfp/deprecated --cov=kfp
```

**Checklist:**

### DIFF
--- a/sdk/python/kfp/cli/component_test.py
+++ b/sdk/python/kfp/cli/component_test.py
@@ -13,7 +13,6 @@
 # limitations under the License.
 """Tests for `components` command group in KFP CLI."""
 import contextlib
-import functools
 import os
 import pathlib
 import subprocess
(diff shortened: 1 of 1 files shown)


DharmitD:
ready for review.
cc: @chensun @gregsheremeta @connor-mccarthy @hbelmiro

hbelmiro (commented):
/lgtm

gregsheremeta:
what was the `mock.patch.object` supposed to be doing, why wasn't it working, and how does calling a build via subprocess act as a valid replacement? (The commit message and PR should explain this, too.)

DharmitD:
> what was the `mock.patch.object` supposed to be doing, why wasn't it working, and how does calling a build via subprocess act as a valid replacement? (The commit message and PR should explain this, too.)

Done, added some details on this in the commit message and the PR description.

sdk/python/kfp/cli/component_test.py
```diff
@@ -579,27 +578,28 @@ def test_existing_dockerfile_can_be_overwritten(self):
...
-        "Skipping this test as it's failing. Refer to https://github.com/kubeflow/pipelines/issues/11038"
-    )
     def test_dockerfile_can_contain_custom_kfp_package(self):
         component = _make_component(
             func_name='train', target_image='custom-image')
         _write_components('components.py', component)
         package_dir = os.path.dirname(os.path.dirname(self.current_dir))
 
-        # suppresses large stdout from subprocess that builds kfp package
-        with mock.patch.object(
```
chensun, line 592: I think it is actually the intention to mock subprocess runs, as unit tests should generally be isolated and dependency-free. Subprocesses introduce external dependencies that can make tests flaky and harder to debug.
Why was the mock not working?

diegolovison: +1

DharmitD: The [original error message](https://github.com/kubeflow/pipelines/issues/11038) indicates that the test can't find the built KFP wheel file in the temp directory.
This is because the mock of `subprocess.run` doesn’t perform the actual build process (e.g., running `python setup.py bdist_wheel`), it doesn't create the wheel file.
While the mock ensures `subprocess.run` doesn't actually execute, it fails to account for the necessary file outputs (i.e., `.whl` file). 
In this case mock wasn't building out the wheel file entirely. Replacing by `subprocess.run` did it.

mprahl: I agree that if we can avoid calling a Python build as part of a unit test, we should.

@DharmitD could you consider mocking the subprocess call with a mock function that just creates an empty file with a `.whl` extension in the temporary directory passed to the `--dist-dir` argument? Then the check for the wheel file would succeed.

PR #11295, opened 2024-10-14
test: Resolve ci-check failures in long-running CI tests

VaniHaripriya:
-  Increased the retry count to support long-running tests.
- Added a label check to ensure the ci-check runs only when the ```needs-ok-to-test``` label is absent.
- The workflow is triggered when the ```ok-to-test``` label is present.


hbelmiro (commented):
@VaniHaripriya since we're adding this validation. The workflows should be triggered also when we add the `ok-to-test` label.

hbelmiro (commented):
/lgtm
/unhold

cc @DharmitD

hbelmiro (approved):

PR #11305, opened 2024-10-17
WIP: test: Re-enable KFP Dockerfile-CLI component SDK unit test

DharmitD:
Removed the skip unit test decorator that was previously added to bypass the test due to failure (referencing issue #11038).

The issue reported in #11038 was caused due to the Dockerfile referencing to Python 3.7. With the Python version updated to 3.9 (done as part of [this commit](https://github.com/kubeflow/pipelines/commit/123ed1eb30c26290546429ffce6360d0910b8eee#diff-dd6f2caaca60c10dc9a15fddf7c7e844e853a9c6a8917f036b304be71fbf376d)), the test now passes consistently. This commit re-enables the test to ensure proper validation of Dockerfile creation and KFP package.

**Checklist:**

### DIFF
--- a/sdk/python/kfp/cli/component_test.py
+++ b/sdk/python/kfp/cli/component_test.py
@@ -579,9 +579,6 @@ class Test(unittest.TestCase):
                 COPY . .
                 '''))
 
-    @unittest.skip(
-        "Skipping this test as it's failing. Refer to https://github.com/kubeflow/pipelines/issues/11038"
-    )
     def test_dockerfile_can_contain_custom_kfp_package(self):
         component = _make_component(
             func_name='train', target_image='custom-image')


sdk/python/kfp/cli/component_test.py
```diff
@@ -608,7 +605,7 @@ def test_dockerfile_can_contain_custom_kfp_package(self):
         file_start = textwrap.dedent('''\
                 # Generated by KFP.
 
-                FROM python:3.9
+                FROM python:3.12
```
HumairAK, line 608: based on your comment, it seems like it works for 3.9, if so maybe it's best if we keep this consistent with the rest of the tests and use 3.9

DharmitD: Unfortunately it doesn't. I'd executed a subprocess previously in my virtual environment and then ran this test which is why it went through. 

I replicated this on a fresh venv and [the missing wheel file error](https://github.com/kubeflow/pipelines/issues/11038) still comes up.

Issue #11061, opened 2024-07-31
Remove obsolete README files after migrating tests to GitHub Actions


hbelmiro:
We have several README files with obsolete information about tests.
Once we migrate all tests to GitHub Actions we need to update/remove those files.

### Labels

/area testing

---

Impacted by this bug? Give it a 👍.

PR #11322, opened 2024-10-21
fix(tests): remove redundant integration test wf

droctothorpe:
Identical sample tests are being run in two separate CI workflows for some reason: [kfp-samples](https://github.com/kubeflow/pipelines/blob/master/.github/workflows/kfp-samples.yml) and [kubeflow-pipelines-integration-v2](https://github.com/kubeflow/pipelines/blob/master/.github/workflows/kubeflow-pipelines-integration-v2.yml). The first [invokes](https://github.com/kubeflow/pipelines/blob/c2a77134265dbbe7ca8492e6e429fd3cc60b8419/backend/src/v2/test/sample-test.sh#L28) `sample_test.py`. The second does as [well](https://github.com/kubeflow/pipelines/blob/c2a77134265dbbe7ca8492e6e429fd3cc60b8419/backend/src/v2/test/Makefile#L14) (after pathing through a [Makefile](https://github.com/kubeflow/pipelines/blob/c2a77134265dbbe7ca8492e6e429fd3cc60b8419/backend/src/v2/test/integration-test.sh#L25)). Presumably, we don’t need both. 

I suggest we tackle eliminating `integration-test.sh` and some of the corresponding out of date docs in a follow-up PR. There should be one, clearly documented path for running large tests both in CI and from local.

**Checklist:**


DharmitD (commented):
/lgtm

.github/workflows/kubeflow-pipelines-integration-v2.yml
```diff
@@ -1,39 +0,0 @@
...
-  push:
-    branches:
-      - master
-  pull_request:
-    paths:
-      - '.github/workflows/kubeflow-pipelines-integration-v2.yml'
-      - '.github/resources/**'
-      - 'samples/core/parameterized_tfx_oss/**'
-      - 'samples/core/dataflow/**'
-      - 'backend/**'
```
hbelmiro, line 13: We need to add these paths to https://github.com/kubeflow/pipelines/blob/095a794cd46df470a821d60b1b5289f2564d33f3/.github/workflows/kfp-samples.yml#L8 so `kfp-samples.yml` is triggered by changes on those.

hbelmiro (changes requested):
> I suggest we tackle eliminating integration-test.sh

@droctothorpe can you remove the file in this PR.

> and some of the corresponding out of date docs

FYI, We have an open issue for that: https://github.com/kubeflow/pipelines/issues/11061

.github/workflows/kubeflow-pipelines-integration-v2.yml
droctothorpe: Good call. Done. Thanks, @hbelmiro! 🙏

hbelmiro (commented):
Thank you for addressing my suggestions @droctothorpe.

> > I suggest we tackle eliminating integration-test.sh
> 
> @droctothorpe can you remove the file in this PR.

You probably missed this comment. Can we remove the file in this PR if it doesn't impact anything?

droctothorpe:
> Thank you for addressing my suggestions @droctothorpe.
> 
> > > I suggest we tackle eliminating integration-test.sh
> > 
> > 
> > @droctothorpe can you remove the file in this PR.
> 
> You probably missed this comment. Can we remove the file in this PR if it doesn't impact anything?

Done! 👍

rimolive:
Based on @hbelmiro suggestions that was already addressed, I believe we're good to go.

/lgtm

hbelmiro (commented):
Thank you @droctothorpe.
/lgtm

chensun:
Just to share a bit history, in the past, sample tests and integration tests are different workflows, IIRC, the former runs against the latest release--we do this by maintaining a long running deployment and update it post every release--while the latter runs against the built from master head + the PR being tested--developing on the fly. 
I sense this current GA workflow may have merged, but in the long run, I think there're value to test with both stable (release) and branch head.

PR #11326, opened 2024-10-23
feat(component): execute in a virtual env

leseb:

This commit introduces the ability to run the component inside a virtual environment, which is particularly useful when the Python system site package directory is read-only, causing pip to fail during the installation of kfp and other dependencies.

To bypass this issue, a writable temporary directory is selected to create a virtual environment that inherits the global system site packages. The entire workflow is then executed from within this virtual environment.

Furthermore, a new field, `use_venv`, has been added to the component decorator, with a default value of `False`.

Signed-off-by: Sébastien Han <seb@redhat.com>

**Checklist:**

### DIFF
--- a/components/google-cloud/RELEASE.md
+++ b/components/google-cloud/RELEASE.md
@@ -2,6 +2,7 @@
 * Remove default prediction column names in `v1.model_evaluation.regression_component` component to fix pipeline errors when using bigquery data source.
 * Add reservation_affinition support in `v1.create_custom_training_job_from_component`.
 * Deprecate `preview.custom_job` module.
+* Add a new `use_venv` field to the component decorator, enabling the component to run inside a virtual environment.
 
 ## Release 2.17.0
 * Fix Gemini batch prediction support to `v1.model_evaluation.autosxs_pipeline` after output schema change.
--- a/sdk/python/kfp/dsl/component_decorator.py
+++ b/sdk/python/kfp/dsl/component_decorator.py
@@ -28,7 +28,8 @@ def component(func: Optional[Callable] = None,
               output_component_file: Optional[str] = None,
               install_kfp_package: bool = True,
               kfp_package_path: Optional[str] = None,
-              pip_trusted_hosts: Optional[List[str]] = None):
+              pip_trusted_hosts: Optional[List[str]] = None,
+              use_venv: bool = False):
     """Decorator for Python-function based components.
 
     A KFP component can either be a lightweight component or a containerized
@@ -75,6 +76,9 @@ def component(func: Optional[Callable] = None,
             as that used when this component was created. Component authors can
             choose to override this to point to a GitHub pull request or
             other pip-compatible package server.
+        use_venv: Specifies if the component should be executed in a virtual environment.
+            The environment will be created in a temporary directory and will inherit the system site packages.
+            This is useful in restricted environments where most of the system is read-only.
 
     Returns:
         A component task factory that can be used in pipeline definitions.
@@ -116,7 +120,8 @@ def component(func: Optional[Callable] = None,
             output_component_file=output_component_file,
             install_kfp_package=install_kfp_package,
             kfp_package_path=kfp_package_path,
-            pip_trusted_hosts=pip_trusted_hosts)
+            pip_trusted_hosts=pip_trusted_hosts,
+            use_venv=use_venv)
 
     return component_factory.create_component_from_func(
         func,
@@ -127,4 +132,5 @@ def component(func: Optional[Callable] = None,
         output_component_file=output_component_file,
         install_kfp_package=install_kfp_package,
         kfp_package_path=kfp_package_path,
-        pip_trusted_hosts=pip_trusted_hosts)
+        pip_trusted_hosts=pip_trusted_hosts,
+        use_venv=use_venv)
--- a/sdk/python/kfp/dsl/component_factory.py
+++ b/sdk/python/kfp/dsl/component_factory.py
@@ -57,6 +57,7 @@ class ComponentInfo():
     packages_to_install: Optional[List[str]] = None
     pip_index_urls: Optional[List[str]] = None
     pip_trusted_hosts: Optional[List[str]] = None
+    use_venv: bool = False
 
 
 # A map from function_name to components.  This is always populated when a
@@ -132,6 +133,15 @@ fi
 PIP_DISABLE_PIP_VERSION_CHECK=1 {pip_install_commands} && "$0" "$@"
 '''
 
+# Creates and activates a virtual environment in a temporary directory.
+# The environment inherits the system site packages.
+_use_venv_script_template = '''
+export PIP_DISABLE_PIP_VERSION_CHECK=1
+tmp=$(mktemp -d)
+python3 -m venv "$tmp/venv" --system-site-packages
+. "$tmp/venv/bin/activate"
+'''
+
 
 def _get_packages_to_install_command(
     kfp_package_path: Optional[str] = None,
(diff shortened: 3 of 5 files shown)


leseb:
@HumairAK FYI

sdk/python/kfp/dsl/component_decorator.py
```diff
@@ -75,6 +76,9 @@ def component(func: Optional[Callable] = None,
             as that used when this component was created. Component authors can
             choose to override this to point to a GitHub pull request or
             other pip-compatible package server.
+        use_venv: Specifies if the component should be executed in a virtual environment.
+        The environment will be created in a temporary directory and will inherit the system site packages.
+        This is useful in restricted environments where most of the system is read-only.
```
HumairAK, line 81: nit: can you indent the 2nd/3rd lines here to match the formatting of the previous parameter docstrings

sdk/python/kfp/dsl/component_decorator.py
```diff
@@ -127,4 +132,5 @@ def pipeline():
         output_component_file=output_component_file,
         install_kfp_package=install_kfp_package,
         kfp_package_path=kfp_package_path,
-        pip_trusted_hosts=pip_trusted_hosts)
+        pip_trusted_hosts=pip_trusted_hosts,
+        use_venv=use_venv)
```
HumairAK, line 136: nit: remove newline

sdk/python/kfp/dsl/component_factory.py
```diff
@@ -132,6 +133,14 @@ def make_pip_install_command(
 PIP_DISABLE_PIP_VERSION_CHECK=1 {pip_install_commands} && "$0" "$@"
 '''
 
+# Creates and activates a virtual environment in a temporary directory.
+# The environment inherits the system site packages.
+_use_venv_script_template = '''
+tmp=$(mktemp -d)
```
HumairAK, line 139: worth noting that `mktemp` is not posix compliant, 
https://stackoverflow.com/questions/2792675/how-portable-is-mktemp1

though it's pretty widespread

HumairAK:
Thank you @leseb this is a useful addition imo, some nits and comments above, otherwise this lgtm

cc @chensun

sdk/python/kfp/dsl/component_factory.py
leseb: What's the concern about not being POSIX compliant? Despite Linux, does the project support any other UNIX-like system?

The main executor also uses `mktemp`. We could look into platformdir or xdg base dir if needed.

sdk/python/kfp/dsl/component_decorator.py
leseb: Ack!

sdk/python/kfp/dsl/component_decorator.py
```diff
@@ -116,7 +120,8 @@ def pipeline():
             output_component_file=output_component_file,
             install_kfp_package=install_kfp_package,
             kfp_package_path=kfp_package_path,
-            pip_trusted_hosts=pip_trusted_hosts)
+            pip_trusted_hosts=pip_trusted_hosts,
+            use_venv=use_venv)
```
DharmitD, line 124: Would be nice to have a unit test for this `use_venv` functionality, including an edge case where the environment might have read-only restrictions.

components/google-cloud/RELEASE.md
```diff
@@ -1,5 +1,6 @@
 ## Upcoming release
 * Remove default prediction column names in `v1.model_evaluation.regression_component` component to fix pipeline errors when using bigquery data source.
+* Add a new `use_venv` field to the component decorator, enabling the component to run inside a virtual environment.
```
DharmitD, line 3: Would be nice to document the use of the `use_venv` field in the component decorator to ensure users understand when and how to use it.

leseb: Do you mean this https://github.com/kubeflow/pipelines/pull/11326/files#diff-6a9e82b5686376e314dddd14a9faefcecf870d0b42cb2256d3934f7fbf2031a4R79?

sdk/python/kfp/dsl/component_factory.py
```diff
@@ -132,6 +133,14 @@ def make_pip_install_command(
...
+python3 -m venv "$tmp/venv" --system-site-packages
+source "$tmp/venv/bin/activate"
```
DharmitD, line 141: We might also need a cleanup mechanism to ensure temporary directories used for virtual environments are deleted after use.

leseb: Thanks!

A few questions:

* Is it a must-have or a nice-to-have? 
* What's the rationale for the cleanup? The main executor does not have such mechanism AFAIK. 
* Isn't the life-time of the pod sufficient to handle cleanup?

sdk/python/kfp/dsl/component_decorator.py
leseb: I can work on that. Can you recommend the right place for both unit test (`sdk/python/kfp/dsl/component_decorator_test.py`?) and the edge case? Thanks!

sdk/python/kfp/dsl/component_factory.py
HumairAK: I don't have concern per se, just noting it here for posterity, if we are using this elsewhere, then the precedent is already set and it's out of scope of this PR imo

sdk/python/kfp/dsl/component_decorator.py
HumairAK: the tests that are included seem sufficient to me, is there anything not covered there that you are looking to address @DharmitD ?

DharmitD: I was thinking of an edge case where the venv set up fails, but I agree, existing tests seem sufficient.

sdk/python/kfp/dsl/component_factory.py
HumairAK, line 141: `source` is resulting in: 

```
sh: 12: source: not found
```

Since `source` is not available in shell (it's available in bash/zsh)

Can you replace this with `.` which is available in `sh`?

leseb: Done. Thanks!

sdk/python/kfp/dsl/component_factory.py
```diff
@@ -132,6 +133,14 @@ def make_pip_install_command(
 PIP_DISABLE_PIP_VERSION_CHECK=1 {pip_install_commands} && "$0" "$@"
 '''
 
+# Creates and activates a virtual environment in a temporary directory.
+# The environment inherits the system site packages.
+_use_venv_script_template = '''
```
HumairAK, line 138: because this is being embedded into `_install_python_packages_script_template` 
the new lines are causing this: 

```
PIP_DISABLE_PIP_VERSION_CHECK=1 python3 -m pip install ...
```

Now become: 
```
          PIP_DISABLE_PIP_VERSION_CHECK=1 
          tmp=$(mktemp -d)
          python3 -m venv "$tmp/venv" --system-site-packages
          . "$tmp/venv/bin/activate"
           python3 -m pip install --quiet
```

this is causing the python3 command to ignore it, this can result in following logs that we want to ignore: 

```
[notice] A new release of pip is available: 23.0.1 -> 24.3.1
[notice] To update, run: pip install --upgrade pip
[notice] A new release of pip is available: 23.0.1 -> 24.3.1
[notice] To update, run: pip install --upgrade pip
```


one solution is to `export PIP_DISABLE_PIP_VERSION_CHECK=1` so we apply it to all subprocesses

leseb: Good catch, thanks! Updated :)

leseb: I prefixed the `python3` with `PIP_DISABLE_PIP_VERSION_CHECK=1` to avoid exporting from `_install_python_packages_script_template` since it would have changed all the yamls in the repo.

PR #11340, opened 2024-10-30
feat(SDK): Add SemaphoreKey and MutexName fields to DSL

DharmitD:
This PR introduces `semaphoreKey` and `mutexName` fields in `PipelineConfig` to support pipeline-level concurrency controls in KFP SDK.

This PR should be merged only after https://github.com/kubeflow/pipelines/pull/11384 gets merged. 

**Testing instructions**

Create a Python virtualenv and install the SDK and IR YAML API packages locally:
```
$ python -m venv .venv
$ source .venv/bin/activate
$ pip install wheel setuptools protobuf grpcio grpcio-tools
$ pip install -r sdk/python/requirements-dev.txt
$ pip install -e api/v2alpha1/python
$ pip install -e sdk/python
```
Use the [example code](https://gist.github.com/DharmitD/57a04c25e4211ffb9cf8891ddb34e1ce) to compile
```
$ kfp dsl compile --py main.py --output main.yaml
```
You should be able to compile and find the following snippet in the main.yaml file:
```
---
platforms:
  kubernetes:
    pipelineConfig:
      mutexName: mutex
      semaphoreKey: semaphore
```

**Checklist:**

### DIFF
--- a/sdk/python/kfp/compiler/compiler_test.py
+++ b/sdk/python/kfp/compiler/compiler_test.py
@@ -4336,6 +4336,67 @@ class TestPlatformConfig(unittest.TestCase):
                     pipeline_func=my_pipeline, package_path=output_yaml)
 
 
+class TestPipelineSemaphoreMutex(unittest.TestCase):
+
+    def test_pipeline_with_semaphore(self):
+        """Test that pipeline config correctly sets the semaphore key."""
+        config = PipelineConfig()
+        config.semaphore_key = 'semaphore'
+
+        @dsl.pipeline(pipeline_config=config)
+        def my_pipeline():
+            task = comp()
+
+        with tempfile.TemporaryDirectory() as tempdir:
+            output_yaml = os.path.join(tempdir, 'pipeline.yaml')
+            compiler.Compiler().compile(
+                pipeline_func=my_pipeline, package_path=output_yaml)
+
+            with open(output_yaml, 'r') as f:
+                pipeline_docs = list(yaml.safe_load_all(f))
+
+        platform_spec = None
+        for doc in pipeline_docs:
+            if 'platforms' in doc:
+                platform_spec = doc
+                break
+
+        self.assertIsNotNone(platform_spec,
+                             'No platforms section found in compiled output')
+        kubernetes_spec = platform_spec['platforms']['kubernetes'][
+            'pipelineConfig']
+        self.assertEqual(kubernetes_spec['semaphoreKey'], 'semaphore')
+
+    def test_pipeline_with_mutex(self):
+        """Test that pipeline config correctly sets the mutex name."""
+        config = PipelineConfig()
+        config.mutex_name = 'mutex'
+
+        @dsl.pipeline(pipeline_config=config)
+        def my_pipeline():
+            task = comp()
+
+        with tempfile.TemporaryDirectory() as tempdir:
+            output_yaml = os.path.join(tempdir, 'pipeline.yaml')
+            compiler.Compiler().compile(
+                pipeline_func=my_pipeline, package_path=output_yaml)
+
+            with open(output_yaml, 'r') as f:
+                pipeline_docs = list(yaml.safe_load_all(f))
+
+        platform_spec = None
+        for doc in pipeline_docs:
+            if 'platforms' in doc:
+                platform_spec = doc
+                break
+
+        self.assertIsNotNone(platform_spec,
+                             'No platforms section found in compiled output')
+        kubernetes_spec = platform_spec['platforms']['kubernetes'][
+            'pipelineConfig']
+        self.assertEqual(kubernetes_spec['mutexName'], 'mutex')
+
+
 class ExtractInputOutputDescription(unittest.TestCase):
 
     def test_no_descriptions(self):
--- a/sdk/python/kfp/compiler/pipeline_spec_builder.py
+++ b/sdk/python/kfp/compiler/pipeline_spec_builder.py
@@ -2242,14 +2242,20 @@ def _write_kubernetes_manifest_to_file(
 
 def _merge_pipeline_config(pipelineConfig: pipeline_config.PipelineConfig,
                            platformSpec: pipeline_spec_pb2.PlatformSpec):
-    workspace = pipelineConfig.workspace
-    if workspace is None:
-        return platformSpec
+    config_dict = {}
 
-    json_format.ParseDict(
-        {'pipelineConfig': {
-            'workspace': workspace.get_workspace(),
-        }}, platformSpec.platforms['kubernetes'])
+    workspace = pipelineConfig.workspace
+    if workspace is not None:
+        config_dict['workspace'] = workspace.get_workspace()
+
+    if pipelineConfig.semaphore_key is not None:
+        config_dict['semaphoreKey'] = pipelineConfig.semaphore_key
+    if pipelineConfig.mutex_name is not None:
+        config_dict['mutexName'] = pipelineConfig.mutex_name
+
+    if config_dict:
+        json_format.ParseDict({'pipelineConfig': config_dict},
+                              platformSpec.platforms['kubernetes'])
 
     return platformSpec
 
--- a/sdk/python/kfp/dsl/pipeline_config.py
+++ b/sdk/python/kfp/dsl/pipeline_config.py
@@ -96,5 +96,57 @@ class WorkspaceConfig:
 class PipelineConfig:
     """PipelineConfig contains pipeline-level config options."""
 
-    def __init__(self, workspace: Optional[WorkspaceConfig] = None):
+    def __init__(self,
+                 workspace: Optional[WorkspaceConfig] = None,
+                 semaphore_key: Optional[str] = None,
+                 mutex_name: Optional[str] = None):
         self.workspace = workspace
+        self._semaphore_key = semaphore_key
+        self._mutex_name = mutex_name
+
+    @property
+    def semaphore_key(self) -> Optional[str]:
+        """Get the semaphore key for controlling pipeline concurrency.
+
+        Returns:
+            Optional[str]: The semaphore key, or None if not set.
+        """
+        return self._semaphore_key
+
+    @semaphore_key.setter
+    def semaphore_key(self, value: str):
+        """Set the semaphore key to control pipeline concurrency.
+
+        Pipelines with the same semaphore key will be limited to a configured maximum
+        number of concurrent executions. This allows you to control resource usage by
+        ensuring that only a specific number of pipelines can run simultaneously.
+
+        Note: A pipeline can use both semaphores and mutexes together. The pipeline
+        will wait until all required locks are available before starting.
+
+        Args:
+            value (str): The semaphore key name for controlling concurrent executions.
+        """
+        self._semaphore_key = (value and value.strip()) or None
+
+    @property
+    def mutex_name(self) -> Optional[str]:
+        """Get the mutex name for exclusive pipeline execution.
+
+        Returns:
+            Optional[str]: The mutex name, or None if not set.
+        """
+        return self._mutex_name
+
+    @mutex_name.setter
+    def mutex_name(self, value: str):
+        """Set the name of the mutex to ensure mutual exclusion.
+
+        Pipelines with the same mutex name will only run one at a time. This ensures
+        exclusive access to shared resources and prevents conflicts when multiple
+        pipelines would otherwise compete for the same resources.
+
+        Args:
+            value (str): Name of the mutex for exclusive pipeline execution.
+        """
+        self._mutex_name = (value and value.strip()) or None


api/v2alpha1/pipeline_spec.proto
```diff
@@ -1103,5 +1103,6 @@ message PlatformDeploymentConfig {
 
 // Spec for pipeline-level config options. See PipelineConfig DSL class.
 message PipelineConfig {
-  // TODO add pipeline-level configs
+  string semaphore_name = 1;
```
gregsheremeta, line 1106: add a small comment for each

api/v2alpha1/go/pipelinespec/pipeline_spec.pb.go
```diff
@@ -1,7 +1,7 @@
 // Code generated by protoc-gen-go. DO NOT EDIT.
 // versions:
-// 	protoc-gen-go v1.33.0
-// 	protoc        v3.17.3
+// 	protoc-gen-go v1.35.1
```
gregsheremeta, line 3: what version of protoc do you have installed? I think we'll need to pick one that we all standardize on. I've been using protoc-21.12.

sdk/python/kfp/compiler/pipeline_spec_builder.py
```diff
@@ -2071,11 +2071,11 @@ def write_pipeline_spec_to_file(
 
 def _merge_pipeline_config(pipelineConfig: pipeline_config.PipelineConfig,
                            platformSpec: pipeline_spec_pb2.PlatformSpec):
-    # TODO: add pipeline config options (ttl, semaphore, etc.) to the dict
-    # json_format.ParseDict(
-    #     {'pipelineConfig': {
-    #         '<some pipeline config option>': pipelineConfig.<get that value>,
-    #     }}, platformSpec.platforms['kubernetes'])
+    pipeline_config_json = json_format.ParseDict(
```
gregsheremeta, line 2074: pipeline_config_json isn't used. Notice how ParseDict edits platformSpec (as an input parameter)

gregsheremeta:
couple nitpicks but overall lgtm.

I don't want to merge it until the backend PR is posted and close to shipping.

gregsheremeta (commented):
/hold

lgtm but let's wait until the backend PR is ready to go too

sdk/python/kfp/compiler/pipeline_spec_builder.py
DharmitD: Done, removed `pipeline_config_json`

api/v2alpha1/pipeline_spec.proto
DharmitD: Done, added a comment for both.

HumairAK:
can we add a test where we verify the semaphore setting compiles to the expected IR?

api/v2alpha1/go/pipelinespec/pipeline_spec.pb.go
```diff
@@ -1,7 +1,7 @@
...
+// 	protoc        v4.25.3
```
gregsheremeta, line 4: What command did you run that resulted in these versions getting bumped? They shouldn't get bumped if you're just going into the api folder and running `make golang` :thinking:

gregsheremeta: might be this causing it:
```
$ pip install wheel setuptools protobuf==4.21.1 grpcio grpcio-tools
```

we should not use protobuf==4.21.1 there. What made you pick that version? Is that something I advised you to do? lol

DharmitD: I actually did a conda install protobuf
`conda install -c conda-forge protobuf`
which installed this protoc version in my conda venv
```
protoc --version
libprotoc 25.3
```

I then went ahead and generated this `.pb.go` file, due to which protoc versions got bumped here. Any ideas on how I could regenerate the `.pb.go` without having these protoc versions updated?

DharmitD:
Edit: Updated to have the DSL generated `SemaphoreKey` instead of `SemaphoreName` field, since having a fixed, hardcoded semaphore name and letting users define the semaphore key is the format prescribed in Argo Workflows' synchronization docs [here.](https://argo-workflows.readthedocs.io/en/latest/synchronization/#multiple-locks)

api/v2alpha1/go/pipelinespec/pipeline_spec.pb.go
DharmitD: Resolved this by running `make golang` as per @gregsheremeta 's suggestion. Thanks!

gregsheremeta:
> can we add a test where we verify the semaphore setting compiles to the expected IR?

still need to do this. Here's an example:
https://github.com/kubeflow/pipelines/pull/10913/files#diff-cac76397fcf6a375f3865a864a3f6725b023c69b5216fa1ed71a86b334641399

DharmitD:
> > can we add a test where we verify the semaphore setting compiles to the expected IR?
> 
> still need to do this. Here's an example: https://github.com/kubeflow/pipelines/pull/10913/files#diff-cac76397fcf6a375f3865a864a3f6725b023c69b5216fa1ed71a86b334641399

Yes I'm working on this ATM. I'm adding a unit test to the `compiler_test.py` similar to the platform tests here: https://github.com/kubeflow/pipelines/blob/8a7b0e5bae5338de7430f4c7d407d594ebc3bd95/sdk/python/kfp/compiler/compiler_test.py#L3489

Have a draft ready, testing locally and pushing it in a bit.

DharmitD:
> > > can we add a test where we verify the semaphore setting compiles to the expected IR?
> > 
> > 
> > still need to do this. Here's an example: https://github.com/kubeflow/pipelines/pull/10913/files#diff-cac76397fcf6a375f3865a864a3f6725b023c69b5216fa1ed71a86b334641399
> 
> Yes I'm working on this ATM. I'm adding a unit test to the `compiler_test.py` similar to the platform tests here:
> 
> https://github.com/kubeflow/pipelines/blob/8a7b0e5bae5338de7430f4c7d407d594ebc3bd95/sdk/python/kfp/compiler/compiler_test.py#L3489
> 
> Have a draft ready, testing locally and pushing it in a bit.

Added a unit test to verify semaphore and mutex setting, the test works as expected and on my local venv, but fails on this PR's CI: https://github.com/kubeflow/pipelines/actions/runs/11846932672/job/33015593483?pr=11340

Wondering if it's because the .proto changes haven't been merged into the code base yet. @gregsheremeta could you take a look and confirm?
And if it is so, then should the test be added in a follow up PR?

gregsheremeta:
> ml_pipelines.PipelineConfig" has no field named "semaphoreKey" at "SinglePlatformSpec.pipelineConfig".

So yes. The test can stay here. Tests just won't pass until the proto is merged and this is rebased. It's the normal/usual/expected flow.

DharmitD:
> > ml_pipelines.PipelineConfig" has no field named "semaphoreKey" at "SinglePlatformSpec.pipelineConfig".
> 
> So yes. The test can stay here. Tests just won't pass until the proto is merged and this is rebased. It's the normal/usual/expected flow.

Edit: Removed proto/API changes and created a separate PR for those here: https://github.com/kubeflow/pipelines/pull/11384
So as @gregsheremeta mentioned, once proto changes are merged and this PR is rebased, the test will work.

sdk/python/kfp/compiler/compiler_test.py
```diff
@@ -3847,6 +3847,42 @@ def outer():
                 foo_platform_set_bar_feature(task, 12)
 
 
+class TestPipelineSemaphoreMutex(unittest.TestCase):
```
gregsheremeta, line 3850: split into two

gregsheremeta:
lgtm other than the test nitpick

sdk/python/kfp/compiler/compiler_test.py
DharmitD: Done, split this test into two - one for semaphore and one for mutex.

gregsheremeta:
lgtm

need to rebase when #11384 is merged, and then I'll tag it

DharmitD:
Please note that the test failures are because the PR is waiting for api changes in https://github.com/kubeflow/pipelines/pull/11384 to be merged. Once merged and this PR is rebased, test failures should be resolved and CI should pass.

test/presubmit-tests-sdk.sh
```diff
@@ -15,20 +15,42 @@
...
-python3 -m pip install --upgrade protobuf
+python3 -m pip uninstall protobuf
+python3 -m pip install protobuf
+python3 -m pip install grpcio grpcio-tools
+python3 -m pip install api/v2alpha1/python
+
+pushd api && make clean-python python && popd
 
+python3 -m pip install api/v2alpha1/python
+##python3 -m pip install -r sdk/python/requirements-dev.txt
```
rimolive, line 48: Remove the commented lines

rimolive (commented):
Good job, Dharmit! Just asked for a nitpick change and you'll be good.

test/presubmit-tests-sdk.sh
DharmitD: done, thanks!

DharmitD:
@chensun @connor-mccarthy could you PTAL? Thank you!

DharmitD:
This PR is LGTM'd by @rimolive, rebased and ready for review.

Should be merged after https://github.com/kubeflow/pipelines/pull/11370 is approved and merged.

sdk/python/kfp/compiler/compiler_test.py
```diff
@@ -4258,6 +4258,71 @@ def test_workspace_config_validation(self):
...
+                pipeline_spec = doc
+                break
+
+        # # Debug statement to print the contents of the pipeline_spec
+        # print("Pipeline spec:", pipeline_spec)
+
+        if pipeline_spec:
+            kubernetes_spec = pipeline_spec['platforms']['kubernetes'][
+                'pipelineConfig']
+            assert kubernetes_spec['semaphoreKey'] == 'semaphore'
```
gmfrasca: for consistency, let's use `self.assertEqual()` here

sdk/python/kfp/compiler/compiler_test.py
```diff
@@ -4258,6 +4258,71 @@ def test_workspace_config_validation(self):
...
+            with open(output_yaml, 'r') as f:
+                pipeline_docs = list(yaml.safe_load_all(f))
+
+        pipeline_spec = None
+        for doc in pipeline_docs:
+            if 'platforms' in doc:
+                pipeline_spec = doc
+                break
+
+        # # Debug statement to print the contents of the pipeline_spec
```
gmfrasca: Probably should remove this

sdk/python/kfp/compiler/compiler_test.py
```diff
@@ -4258,6 +4258,71 @@ def test_workspace_config_validation(self):
...
+        # print("Pipeline spec:", pipeline_spec)
+
+        if pipeline_spec:
+            kubernetes_spec = pipeline_spec['platforms']['kubernetes'][
+                'pipelineConfig']
+            assert kubernetes_spec['mutexName'] == 'mutex'
```
gmfrasca:
```suggestion
            self.assertEqual(kubernetes_spec['mutexName'], 'mutex')
```

test/presubmit-tests-sdk.sh
```diff
@@ -30,6 +40,11 @@ if [ "${SETUP_ENV}" = "true" ]; then
   python3 -m pip install --upgrade protobuf
   python3 -m pip install sdk/python
 
+  # API regeneration steps (from your branch)
+  #install kfp api
```
gmfrasca:
```suggestion
  # install kfp api
```

sdk/python/kfp/dsl/pipeline_config.py
```diff
@@ -75,3 +75,29 @@ class PipelineConfig:
 
     def __init__(self, workspace: Optional[WorkspaceConfig] = None):
```
mprahl: Could you add the new options to the init method?

sdk/python/kfp/dsl/pipeline_config.py
```diff
@@ -75,3 +75,29 @@ class PipelineConfig:
...
         self.workspace = workspace
+        self.semaphore_key = None
+        self.mutex_name = None
+
+    def set_semaphore_key(self, semaphore_key: str):
```
mprahl: Could these be actually getter and setter methods like this PR?
https://github.com/kubeflow/pipelines/pull/12234

sdk/python/kfp/compiler/pipeline_spec_builder.py
```diff
@@ -2242,14 +2242,20 @@ def _write_kubernetes_manifest_to_file(
...
+    if workspace is not None:
+        config_dict['workspace'] = workspace.get_workspace()
 
-    json_format.ParseDict(
-        {'pipelineConfig': {
-            'workspace': workspace.get_workspace(),
-        }}, platformSpec.platforms['kubernetes'])
+    if pipelineConfig.semaphore_key is not None:
+        config_dict['semaphoreKey'] = pipelineConfig.semaphore_key
+    if pipelineConfig.mutex_name is not None:
```
gmfrasca, line 2253: can a pipeline have both mutexes and semaphores, or are they (ironically) mutually exclusive?

test/presubmit-tests-sdk.sh
```diff
@@ -16,11 +16,21 @@
 source_root=$(pwd)
 SETUP_ENV="${SETUP_ENV:-true}"
 
+# Download and install protoc binary (from your branch)
+PROTOC_ZIP=protoc-3.19.6-linux-x86_64.zip
```
mprahl: This shouldn't be needed since the workflow already uses the `./.github/actions/protobuf` Action.

sdk/python/kfp/compiler/compiler_test.py
DharmitD: Updated to use `self.assertEqual()`.

sdk/python/kfp/compiler/compiler_test.py
DharmitD: Oops, left it by mistake. Removed it.

sdk/python/kfp/compiler/compiler_test.py
DharmitD: Done.

test/presubmit-tests-sdk.sh
DharmitD: Done

sdk/python/kfp/dsl/pipeline_config.py
DharmitD: Done, added the options to init.

sdk/python/kfp/dsl/pipeline_config.py
DharmitD: Done, updated to use getter and setter methods / Pythonic properties.

test/presubmit-tests-sdk.sh
DharmitD: Done, removed it.

sdk/python/kfp/compiler/pipeline_spec_builder.py
DharmitD: Can have both, or either one of them, or none. Find related documentation [here.](https://argo-workflows.readthedocs.io/en/latest/synchronization/#multiple-locks)

sdk/python/kfp/dsl/pipeline_config.py
```diff
@@ -73,5 +73,64 @@ def set_kubernetes_config(self,
...
+            Optional[str]: The semaphore key, or None if not set.
+        """
+        return self._semaphore_key
+
+    @semaphore_key.setter
+    def semaphore_key(self, value: str):
+        """Set the name of the semaphore to control pipeline concurrency.
+
+        The semaphore is configured via a ConfigMap. By default, the ConfigMap is
+        named "semaphore-config", but this name can be specified through the APIServer
```
mprahl: Technically, the SDK is supposed to be backend agnostic. Could you keep this docstring describing generic behavior of semaphore?

sdk/python/kfp/dsl/pipeline_config.py
```diff
@@ -73,5 +73,64 @@ def set_kubernetes_config(self,
...
+        """Get the mutex name.
+
+        Returns:
+            Optional[str]: The mutex name, or None if not set.
+        """
+        return self._mutex_name
+
+    @mutex_name.setter
+    def mutex_name(self, value: str):
+        """Set the name of the mutex to ensure mutual exclusion.
```
mprahl, line 143: Could you further describe the behavior and not mention Argo Workflows? I think this can be simplified to something like: Pipelines with the same mutex name will only run one at a time.

test/presubmit-tests-sdk.sh
```diff
@@ -30,6 +30,10 @@ if [ "${SETUP_ENV}" = "true" ]; then
   python3 -m pip install --upgrade protobuf
   python3 -m pip install sdk/python
 
+  # install kfp api
```
mprahl: This looks to be duplicating the code just below it.

sdk/python/kfp/compiler/compiler_test.py
```diff
@@ -4258,6 +4258,65 @@ def test_workspace_config_validation(self):
...
+            with open(output_yaml, 'r') as f:
+                pipeline_docs = list(yaml.safe_load_all(f))
+
+        pipeline_spec = None
+        for doc in pipeline_docs:
+            if 'platforms' in doc:
+                pipeline_spec = doc
+                break
+
+        if pipeline_spec:
```
mprahl: This needs to be replaced with an assertion that `pipeline_spec` is not None because the test will pass right now if it wasn't in the compiled output.

Also, I think a better variable name is `platform_spec` based on my understanding of the code.

sdk/python/kfp/compiler/compiler_test.py
```diff
@@ -4258,6 +4258,65 @@ def test_workspace_config_validation(self):
...
```
mprahl: Same comment as the previous test

sdk/python/kfp/dsl/pipeline_config.py
```diff
@@ -73,5 +73,64 @@ def set_kubernetes_config(self,
...
+
+        Note: A pipeline can have both semaphores and mutexes (Argo Workflows v3.6+).
+        The workflow will block until all locks are available.
+
+        Args:
+            value (str): The key used to map to the ConfigMap.
+        """
+        if not value or not value.strip():
+            raise ValueError('Semaphore key is required and cannot be empty')
+        self._semaphore_key = value.strip()
```
mprahl: It's optional so I don't see why we'd prevent the user from making it None later on, though it would be strange to do so.
```suggestion
        self._semaphore_key = (value and value.strip()) or None
```

sdk/python/kfp/dsl/pipeline_config.py
```diff
@@ -73,5 +73,64 @@ def set_kubernetes_config(self,
...
+
+        Note: A pipeline can have both semaphores and mutexes (Argo Workflows v3.6+).
+        The workflow will block until all locks are available.
+
+        Args:
+            value (str): Name of the mutex.
+        """
+        if not value or not value.strip():
+            raise ValueError('Mutex name is required and cannot be empty')
+        self._mutex_name = value.strip()
```
mprahl: Same comment as above
```suggestion
        self._mutex_name = (value and value.strip()) or None
```

sdk/python/kfp/dsl/pipeline_config.py
DharmitD: Done, just added a generic summary.

sdk/python/kfp/dsl/pipeline_config.py
DharmitD: Done, just added a generic summary.

test/presubmit-tests-sdk.sh
DharmitD: I left it during the rebase thinking the tests might need it, but you're right - it is duplicating indeed. Hence, removed it.

sdk/python/kfp/compiler/compiler_test.py
DharmitD: Done, replaced with an assertion and the `platform_spec` var name.

sdk/python/kfp/compiler/compiler_test.py
DharmitD: Done, replaced with an assertion and the platform_spec var name.

sdk/python/kfp/dsl/pipeline_config.py
DharmitD: Done.

sdk/python/kfp/dsl/pipeline_config.py
DharmitD: Done

mprahl (approved):
/approve
/lgtm
/hold for @gmfrasca to review again

sdk/python/kfp/dsl/pipeline_config.py
```diff
@@ -96,5 +96,57 @@ def set_kubernetes_config(self,
...
+    def __init__(self,
+                 workspace: Optional[WorkspaceConfig] = None,
+                 semaphore_key: Optional[str] = None,
+                 mutex_name: Optional[str] = None):
         self.workspace = workspace
+        self._semaphore_key = semaphore_key
+        self._mutex_name = mutex_name
+
+    @property
+    def semaphore_key(self) -> Optional[str]:
```
nsingla, line 108: Do we want this to be a semi builder pattern style class? or would making it a dataclass be good enough?

DharmitD: The current property-based approach is the right choice because:

- It's consistent with existing KFP SDK patterns
- It's Pythonic and follows Python best practices
- It provides validation exactly where needed
- It offers great user experience with familiar syntax

This pattern was specifically requested by @mprahl, in order to stay consistent with the rest of the code in this file.

The builder pattern would be over-engineering for this use case, and switching to dataclass now would be a breaking change without significant benefits.

gmfrasca (approved):
/lgtm
/unhold

sdk/python/kfp/compiler/compiler_test.py
```diff
@@ -4336,6 +4336,67 @@ def my_pipeline():
                     pipeline_func=my_pipeline, package_path=output_yaml)
 
 
+class TestPipelineSemaphoreMutex(unittest.TestCase):
```
nsingla, line 4339: possible for you to create a python component + pipeline file for these tests?
the structure of these tests are changing such that:
1. Component/Pipeline files creates YAML
2. Compile tests compile pipeline/component function
3. Compare the compiled YAML with a known working YAML 
This will prevent any regression from happening in the future

Also, if you can add assertion to the component code, then it will give us coverage on the expectation while the pipeline is running w/o adding a custom logic to verify if the pipeline did what it was supposed to do

DharmitD: @nsingla could you please create a follow up issue for this? Thanks!

PR #11347, opened 2024-10-31
chore: upstream our authorization changes for oauth2-proxy

juliusvonkohout:
This upstreams our changes in https://github.com/kubeflow/manifests/blob/master/apps/pipeline/upstream/base/installs/multi-user/istio-authorization-config.yaml for oauth2-proxy in Kubeflow 1.9/1.9.1. It is an exact copy of the file and must be included in the next KFP release or it will break our synchronization scripts at https://github.com/kubeflow/manifests/blob/master/hack/synchronize-kubeflow-manifests.sh. for the 1.10 release.

closes https://github.com/kubeflow/manifests/issues/2804

@HumairAK @rimolive @kimwnasptd

### DIFF
--- a/manifests/kustomize/base/installs/multi-user/istio-authorization-config.yaml
+++ b/manifests/kustomize/base/installs/multi-user/istio-authorization-config.yaml
@@ -8,10 +8,21 @@ spec:
     matchLabels:
       app: ml-pipeline-ui
   rules:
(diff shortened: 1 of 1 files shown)


HumairAK:
/lgtm
/approve

Thanks @juliusvonkohout

PR #11361, opened 2024-11-06
fix(backend): handle client side HTTP timeouts to fix crashes of metadata-writer. Fixes #8200

### LINKED ISSUE #8200: [backend] Metadata writer pod always restarting
### Environment

Manifests in k8s
*  K8S Version:
1.21
*  KFP version: 
1.8.1/1.8.2/1.8.3/1.8.4

### Steps to reproduce

Hi. 

Since release 1.8.1 (can't be sure about older versions) our metadata-writer pod is always restarting infinitely with the following message error:

```
metadata-writer-78fc7d5bb8-ph9kj                         2/2     Running   299        78d
```

```
  File "/usr/local/lib/python3.7/site-packages/urllib3/response.py", line 697, in _update_chunk_length
    self.chunk_left = int(line, 16)
ValueError: invalid literal for int() with base 16: b''

During handling of the above exception, another exception occurred:

  File "/usr/local/lib/python3.7/site-packages/urllib3/response.py", line 438, in _error_catcher
    yield
  File "/usr/local/lib/python3.7/site-packages/urllib3/response.py", line 764, in read_chunked
    self._update_chunk_length()
  File "/usr/local/lib/python3.7/site-packages/urllib3/response.py", line 701, in _update_chunk_length
    raise InvalidChunkLength(self, line)
urllib3.exceptions.InvalidChunkLength: InvalidChunkLength(got length b'', 0 bytes read)

During handling of the above exception, another exception occurred:

  File "/kfp/metadata_writer/metadata_writer.py", line 157, in <module>
    for event in pod_stream:
  File "/usr/local/lib/python3.7/site-packages/kubernetes/watch/watch.py", line 144, in stream
    for line in iter_resp_lines(resp):
  File "/usr/local/lib/python3.7/site-packages/kubernetes/watch/watch.py", line 48, in iter_resp_lines
    for seg in resp.read_chunked(decode_content=False):
  File "/usr/local/lib/python3.7/site-packages/urllib3/response.py", line 793, in read_chunked
    self._original_response.close()
  File "/usr/local/lib/python3.7/contextlib.py", line 130, in __exit__
    self.gen.throw(type, value, traceback)
  File "/usr/local/lib/python3.7/site-packages/urllib3/response.py", line 455, in _error_catcher
    raise ProtocolError("Connection broken: %r" % e, e)
urllib3.exceptions.ProtocolError: ("Connection broken: InvalidChunkLength(got length b'', 0 bytes read)", InvalidChunkLength(got length b'', 0 bytes read))
```

We already try the most recent versions of version 1.8 (we did not try version 2.0.0). 
The pipelines are working very well, and we don't have any problems till now because of this, but this only happens with this pod. 

This happens in our multiple clusters with multiple installations, so don't look like an issue of a specific cluster. 

### Expected result

The pod should stop restarting. 

---

OutSorcerer:

* Handles client side timeouts (`urllib3.exceptions.ReadTimeoutError`) of `k8s_watch.stream` to prevent crashes of `metadata-writer` pod. Without this `metadata-writer` pod fails in cases when a connection error causes a client timeout. This should fix #8200.
* Adjusts the client side timeout according to [the recommendations of Kubernetes Python client authors](https://github.com/kubernetes-client/python/blob/master/examples/watch/timeout-settings.md). 
  >It is recommended to set this [server] timeout value to a higher number such as 3600 seconds (1 hour).
  >It is recommended to set this [client] timeout value to a lower number (for eg. ~ maybe 60 seconds).
  * The benefit of this for the `metadata-writer` use case is that, currently, with a client side timeout of 2000 seconds (33.3 minutes) if a connection error happens during this period, `metadata-writer` stops doing its job for up to 33.3 minutes. After this change `metadata-writer` will only be disconneted for up to the new client timeout, 60 seconds.
  * Since the client timeout is now smaller than the server timeout, if no events occur within the client timeout period, a `ReadTimeoutError` is thrown and caught even in the absence of errors.
* Makes client side and server side timeout configurable via environment variables.
* Updates Kubernetes Python client to the latest version.

**Checklist:**

### DIFF
--- a/backend/metadata_writer/src/metadata_writer.py
+++ b/backend/metadata_writer/src/metadata_writer.py
@@ -402,4 +402,5 @@ while True:
 
     except urllib3.exceptions.ReadTimeoutError as e:
         # Client side timeout, continue watching.
-        pass
+        continue
+ 
--- a/backend/metadata_writer/src/metadata_writer.py
+++ b/backend/metadata_writer/src/metadata_writer.py
@@ -398,6 +398,7 @@ while True:
                 pods_with_written_metadata[obj.metadata.name] = None
 
         # Server side timeout, continue watching.
+        pass
 
     except urllib3.exceptions.ReadTimeoutError as e:
         # Client side timeout, continue watching.
--- a/backend/metadata_writer/src/metadata_writer.py
+++ b/backend/metadata_writer/src/metadata_writer.py
@@ -19,6 +19,7 @@ import re
 import collections
 import kubernetes
 import yaml
+import urllib3
 from time import sleep
 import lru
 
@@ -30,7 +31,10 @@ pod_name_to_execution_id_size = os.environ.get('POD_NAME_TO_EXECUTION_ID_SIZE', 
 workflow_name_to_context_id_size = os.environ.get('WORKFLOW_NAME_TO_CONTEXT_ID_SIZE', 5000)
 pods_with_written_metadata_size = os.environ.get('PODS_WITH_WRITTEN_METADATA_SIZE', 5000)
 debug_files_size = os.environ.get('DEBUG_FILES_SIZE', 5000)
-
+# See the documentation on settings k8s_watch timeouts:
+# https://github.com/kubernetes-client/python/blob/master/examples/watch/timeout-settings.md
+k8s_watch_server_side_timeout = os.environ.get('K8S_WATCH_SERVER_SIDE_TIMEOUT', 1800)
+k8s_watch_client_side_timeout = os.environ.get('K8S_WATCH_CLIENT_SIDE_TIMEOUT', 60)
 
 kubernetes.config.load_incluster_config()
 k8s_api = kubernetes.client.CoreV1Api()
@@ -150,18 +154,18 @@ while True:
             k8s_api.list_namespaced_pod,
             namespace=namespace_to_watch,
             label_selector=ARGO_WORKFLOW_LABEL_KEY,
-            timeout_seconds=1800,  # Sometimes watch gets stuck
-            _request_timeout=2000,  # Sometimes HTTP GET gets stuck
+            timeout_seconds=k8s_watch_server_side_timeout, 
+            _request_timeout=k8s_watch_client_side_timeout,
         )
     else:
         pod_stream = k8s_watch.stream(
             k8s_api.list_pod_for_all_namespaces,
             label_selector=ARGO_WORKFLOW_LABEL_KEY,
-            timeout_seconds=1800,  # Sometimes watch gets stuck
-            _request_timeout=2000,  # Sometimes HTTP GET gets stuck
+            timeout_seconds=k8s_watch_server_side_timeout, 
+            _request_timeout=k8s_watch_client_side_timeout,
         )
-    for event in pod_stream:
-        try:
+    try:
+        for event in pod_stream:
             obj = event['object']
             print('Kubernetes Pod event: ', event['type'], obj.metadata.name, obj.metadata.resource_version)
             if event['type'] == 'ERROR':
@@ -393,6 +397,12 @@ while True:
 
                 pods_with_written_metadata[obj.metadata.name] = None
 
-        except Exception as e:
-            import traceback
-            print(traceback.format_exc())
+        # Server side timeout, continue watching.
+
+    except urllib3.exceptions.ReadTimeoutError as e:
+        # Client side timeout, continue watching.
+        pass
+
+    except Exception as e:
+        import traceback
+        print(traceback.format_exc())
--- a/backend/metadata_writer/src/metadata_writer.py
+++ b/backend/metadata_writer/src/metadata_writer.py
@@ -397,7 +397,7 @@ while True:
 
                 pods_with_written_metadata[obj.metadata.name] = None
 
-        # Server side timeout, continue watching.
+        # If the for loop ended, a server-side timeout occurred. Continue watching.
         pass
 
     except urllib3.exceptions.ReadTimeoutError as e:
--- a/backend/metadata_writer/src/metadata_writer.py
+++ b/backend/metadata_writer/src/metadata_writer.py
@@ -402,7 +402,3 @@ while True:
     except urllib3.exceptions.ReadTimeoutError as e:
         # Client side timeout, continue watching.
         pass
-
-    except Exception as e:
-        import traceback
-        print(traceback.format_exc())
--- a/backend/metadata_writer/requirements.txt
+++ b/backend/metadata_writer/requirements.txt
@@ -22,7 +22,7 @@ grpcio==1.66.1
     # via ml-metadata
 idna==3.10
     # via requests
-kubernetes==10.0.1
+kubernetes==31.0.0
     # via -r -
 lru-dict==1.3.0
     # via -r -


backend/metadata_writer/src/metadata_writer.py
```diff
@@ -393,6 +397,12 @@ def is_kfp_v2_pod(pod) -> bool:
 
                 pods_with_written_metadata[obj.metadata.name] = None
 
-        except Exception as e:
-            import traceback
-            print(traceback.format_exc())
+        # Server side timeout, continue watching.
```
thesuperzapper, line 400: What is this line meant to be a comment for?

backend/metadata_writer/src/metadata_writer.py
```diff
@@ -150,18 +154,18 @@ def is_kfp_v2_pod(pod) -> bool:
...
             label_selector=ARGO_WORKFLOW_LABEL_KEY,
-            timeout_seconds=1800,  # Sometimes watch gets stuck
-            _request_timeout=2000,  # Sometimes HTTP GET gets stuck
+            timeout_seconds=k8s_watch_server_side_timeout, 
+            _request_timeout=k8s_watch_client_side_timeout,
         )
-    for event in pod_stream:
-        try:
+    try:
+        for event in pod_stream:
```
thesuperzapper, line 168: I am trying to understand why we need to retry the entire iterator on every error and thus create a new one?

Or does the iterator returned by `pod_stream` become "poisoned" when it fails, so calling `__next__` on it will never return a new item in the stream?

backend/metadata_writer/src/metadata_writer.py
```diff
@@ -393,6 +397,12 @@ def is_kfp_v2_pod(pod) -> bool:
...
-        except Exception as e:
-            import traceback
-            print(traceback.format_exc())
+        # Server side timeout, continue watching.
+
+    except urllib3.exceptions.ReadTimeoutError as e:
+        # Client side timeout, continue watching.
+        pass
+
+    except Exception as e:
```
thesuperzapper, line 406: I know we were already catching all these errors before, but I am struggling to see why catching `Exception` won't get us sometimes stuck where we never actually crash.

Especially because you proposed above that we have this try around the for loop, meaning any iteration errors will endlessly retry on a `pod_stream` that may never work?

thesuperzapper (commented):
@OutSorcerer just want to clarify a few things.

backend/metadata_writer/src/metadata_writer.py
OutSorcerer: This comment marks the place in code which is only reached when the server side timeout occurs, it was not meant to comment a particular statement.

Should I remove it?

backend/metadata_writer/src/metadata_writer.py
OutSorcerer: >Or does the iterator returned by pod_stream become "poisoned" when it fails, so calling __next__ on it will never return a new item in the stream?

As I understand, yes, this is what happens here.

In case of [a network error causing a client timeout](https://github.com/kubeflow/pipelines/issues/8200#issuecomment-2333979936) it was leading to an unhandled exception in `metadata-writer`, so Kubertenetes was restarting it and increasing the restart counter. 

In the stack trace the error was happening on this line: https://github.com/kubeflow/pipelines/blob/f8973d27cc3b7328b31222ac26b9a6361f8ed16d/backend/metadata_writer/src/metadata_writer.py#L163

```
  File "/kfp/metadata_writer/metadata_writer.py", line 163, in <module>
    for event in pod_stream:
  File "/usr/local/lib/python3.8/site-packages/kubernetes/watch/watch.py", line 144, in stream
    for line in iter_resp_lines(resp):
  File "/usr/local/lib/python3.8/site-packages/kubernetes/watch/watch.py", line 48, in iter_resp_lines
    for seg in resp.read_chunked(decode_content=False):
  File "/usr/local/lib/python3.8/site-packages/urllib3/response.py", line 857, in read_chunked
    self._original_response.close()
  File "/usr/local/lib/python3.8/contextlib.py", line 131, in __exit__
    self.gen.throw(type, value, traceback)
  File "/usr/local/lib/python3.8/site-packages/urllib3/response.py", line 449, in _error_catcher
    raise ReadTimeoutError(self._pool, None, "Read timed out.")
urllib3.exceptions.ReadTimeoutError: HTTPSConnectionPool(host='34.118.224.1', port=443): Read timed out.
```

backend/metadata_writer/src/metadata_writer.py
OutSorcerer: I wanted to keep the amount of changes small originally.

But in my understanding removing this catch-all except clause improves the code. The unhalded exceptions will still be printed to the console and they will not be hidden from a user anymore. A user will see that restart counter increases and will be able to get the logs by a commnad like `kubectl -n kubeflow logs --previous metadata-writer-6d5b8456-78265`. Kubernetes will restart `metadata-writer` pod so it will continue handling events if it is possible. 

So I made a commit that removes this.

backend/metadata_writer/src/metadata_writer.py
ishaan-mehta: Maybe for clarity we can explain in the comment that the "continue[d] watching" occurs by getting a new stream on the next iteration of the `while` loop — that way it is clear the comment is referring to the logical flow rather than a particular statement.

ishaan-mehta:
#8200 originally was opened for the `InvalidChunkLength` issue on the `metadata-writer` pod, but this seems to address only the read timeout issue introduced in the comments (as far as I can tell) — should this still be marked as fixing #8200?

backend/metadata_writer/src/metadata_writer.py
hbelmiro: That comment indeed seems misplaced.
@OutSorcerer isn't possible to handle the server-side exception the same way you did for the client-side one and keep the original exception handling (`except Exception as e:`)?
Something like:

```python
    try:
        ...
    except <Server-side error> as e:
         # Server side timeout, continue watching.
         pass
    except urllib3.exceptions.ReadTimeoutError as e:
         # Client side timeout, continue watching.
         pass
    except Exception as e:
         import traceback
         print(traceback.format_exc())
```

OutSorcerer: But the behaviour in case of a server-side timeout is that the loop `for event in pod_stream` just finishes normally (because the underlying HTTP request to Kubernetes API also finishes) so it cannot be handled by `try-except`.

hbelmiro: I see. Thanks for clarifying @OutSorcerer.
Then how about the following?

```suggestion
        # If the for loop ended, a server-side timeout occurred. Continue watching.
        pass
```

OutSorcerer: OK, I see, now it will be clear to which line that comment belongs. I made a new commit that adds `pass` under the comment.

OutSorcerer: Updated the comment text to `If the for loop ended, a server-side timeout occurred. Continue watching.` as well.

hbelmiro (commented):
/lgtm

ishaan-mehta:
Is there an estimate on when this could be approved and merged, and/or is there anything I can do to help? Just curious as my deployment is running into the same issue.

thesuperzapper:
@kubeflow/pipelines maintainers can we get some eyes on this important PR (It needs some work, but is important as it fixes a critical issue that prevents KFP metadata-writer working on some Kubernetes distros).

For more context, please see my comment here:
- https://github.com/kubeflow/pipelines/issues/8200#issuecomment-2457945006

But the issue is simply that some TCP sockets are timing out when we are watching Kubernetes resources from python.

backend/metadata_writer/src/metadata_writer.py
```diff
@@ -393,6 +397,9 @@ def is_kfp_v2_pod(pod) -> bool:
 
                 pods_with_written_metadata[obj.metadata.name] = None
 
-        except Exception as e:
-            import traceback
-            print(traceback.format_exc())
+        # If the for loop ended, a server-side timeout occurred. Continue watching.
+        pass
```
HumairAK, line 401: this statement doesn't seem like it's necessary

backend/metadata_writer/src/metadata_writer.py
```diff
@@ -393,6 +397,9 @@ def is_kfp_v2_pod(pod) -> bool:
...
+
+    except urllib3.exceptions.ReadTimeoutError as e:
+        # Client side timeout, continue watching.
+        pass
```
HumairAK, line 405: loooks like we can replace pass with `continue`

backend/metadata_writer/src/metadata_writer.py
```diff
@@ -393,6 +397,9 @@ def is_kfp_v2_pod(pod) -> bool:
 
                 pods_with_written_metadata[obj.metadata.name] = None
 
-        except Exception as e:
-            import traceback
-            print(traceback.format_exc())
```
HumairAK, line 398: do we still want to catch other exceptions that are not client timeouts?

backend/metadata_writer/src/metadata_writer.py
OutSorcerer: Technically yes, and originally there was just a comment here, but @hbelmiro suggested to add `pass` statement because otherwise it was confusing to which line the comment belongs.

OutSorcerer: See https://github.com/kubeflow/pipelines/pull/11361#discussion_r1831807285

backend/metadata_writer/src/metadata_writer.py
OutSorcerer: Originally I wanted to keep this, but removed after a comment by @thesuperzapper, because but with the catch-all except clause removed it will be easier for a user to notice that something is wrong, because Kubernetes will restart metadata-writer pod and its restart counter will be incremented. 

In a case that an exception is more severe then a timeout and really blocks metadata-writer from doing its job, it is probaly better to let user now by crashing and restarting it. With the catch-all except clause in the case it will just infinetely repeat the loop so metadata-writer will appear as normally working but will not do anything.

See the thread after https://github.com/kubeflow/pipelines/pull/11361#discussion_r1831813126

backend/metadata_writer/src/metadata_writer.py
OutSorcerer: Done.

HumairAK:
/approve

Thanks @OutSorcerer !

OutSorcerer:
@thesuperzapper, @hbelmiro, @HumairAK, thanks for reviewing the PR!

Issue #6553, opened 2021-09-13
[feature] Expose argo synchronization options


jmcarp:
### Feature Area

/area sdk

### What feature would you like to see?

Argo includes useful synchronization features for workflows and templates: https://argoproj.github.io/argo-workflows/synchronization/. It would be helpful to expose these in kfp without having to manipulate manifests as yaml.

I'd be happy to work on this if the request makes sense. It seems like workflow and template level synchronization options would live on the pipeline config and container op classes respectively, but let me know if there's a better place to put them.

### What is the use case or pain point?

Allow users to set a semaphore on a workflow or a step.

### Is there a workaround currently?

Edit generated yaml.

---

Love this idea? Give it a 👍. We prioritize fulfilling features with the most 👍.

jmcarp:
@Bobgy @capri-xiyue: if this request makes sense, let me know if there's a good place to add these options in the dsl, and I can write up a patch.

zijianjoy:
Hello @jmcarp , what is the CUJ for setting limitation of synchronization on 1)workflow level, 2) step level?

jmcarp:
Hi @zijianjoy, the typical use case in either case is to prevent concurrent execution of tasks that could otherwise change some shared resource. For example, if there's a race condition between tasks, or between multiple runs of the same task, it's helpful to use argo synchronization to prevent those tasks from running concurrently.

This is a useful argo feature, and it would be helpful to have it in kfp as well.

capri-xiyue:
> Hi @zijianjoy, the typical use case in either case is to prevent concurrent execution of tasks that could otherwise change some shared resource. For example, if there's a race condition between tasks, or between multiple runs of the same task, it's helpful to use argo synchronization to prevent those tasks from running concurrently.
> 
> This is a useful argo feature, and it would be helpful to have it in kfp as well.


Hi @jmcarp , in your use case, can you add synchronization logic for the shared resources? I think when it comes to shared resources, the synchronization logic should be added to the shared resources instead of the workflows.

jmcarp:
I *can* implement my own synchronization, but that's the kind of feature I expect the workflow engine to provide--it's easy to get wrong, so ideally it's implemented once at the workflow layer and not independently by each user of the workflow tool. That's why tools like argo workflows and apache airflow include synchronization primitives, for example.

jmcarp:
I see a 👍 from @Bobgy. Would you accept a patch for this?

kdubovikov:
[...]

PR #11370, opened 2024-11-12
feat(backend): Add Semaphore and Mutex fields to Workflow Spec

DharmitD:
This PR introduces `semaphore` and `mutex` fields in Workflow Spec to support pipeline-level concurrency controls in KFP.

**Checklist:**


gregsheremeta:
add `fixes #6553` to the PR description

gregsheremeta:
> The semaphore is configured via a fixed ConfigMap named semaphore-config

We should edit the kubeflow manifest to deploy a skeleton of this configmap. You can do that in here or in a follow-up PR.

gregsheremeta:
> The Workflow CR now includes a Synchronization field

I would probably delete this line (and maybe edit the PR title), because that reads like things you enhanced on Workflow itself. We're just setting fields on it...

gregsheremeta:
```
platforms:
  kubernetes:
    pipelineConfig:
      mutexName: mutex
      semaphoreKey: semaphore
```
> The expected output should include the semaphore and mutex references:

What does Argo Workflows do when both are set?

A better verification would be to do two separate test pipelines -- one where you use mutex, and one where you use semaphore. And then in addition to verifying the Workflow yaml, _also_ verify that multiple runs are being locked like they should be.

backend/src/apiserver/template/v2_template.go
```diff
@@ -77,9 +77,20 @@ func (t *V2Spec) ScheduledWorkflow(modelJob *model.Job) (*scheduledworkflow.Sche
 		}
 	}
 
+	var pipeline_options argocompiler.Options
```
gregsheremeta, line 80:
```suggestion
	var pipelineOptions argocompiler.Options
```

backend/src/apiserver/template/v2_template.go
```diff
@@ -77,9 +77,20 @@ func (t *V2Spec) ScheduledWorkflow(modelJob *model.Job) (*scheduledworkflow.Sche
...
+	for _, platform := range t.platformSpec.Platforms {
+		if platform.PipelineConfig.SemaphoreKey != "" || platform.PipelineConfig.MutexName != "" {
```
gregsheremeta, line 82: 1. We should specifically check for the "kubernetes" platform here. There could be others, and we would ignore all others
2. Check that if you put some other non-kubernetes platform in there that this line doesn't cause a panic. I don't think it will, but just check :)

backend/src/apiserver/template/v2_template.go
```diff
@@ -77,9 +77,20 @@ func (t *V2Spec) ScheduledWorkflow(modelJob *model.Job) (*scheduledworkflow.Sche
...
+			pipeline_options = argocompiler.Options{
+				SemaphoreKey: platform.PipelineConfig.SemaphoreKey,
+				MutexName:    platform.PipelineConfig.MutexName,
```
gregsheremeta, line 85: Let's add these individually, and only if they are specified in the IR. Users will typically only add one or the other. (So don't add mutex if the user only wants a semaphore, and vice versa.)

backend/src/apiserver/template/v2_template.go
```diff
@@ -300,9 +311,20 @@ func (t *V2Spec) RunWorkflow(modelRun *model.Run, options RunWorkflowOptions) (u
 		}
 	}
 
+	var pipeline_options *argocompiler.Options
```
gregsheremeta, line 314: this looks copied from above. Is there any way to reuse it?

backend/src/v2/compiler/argocompiler/argo.go
```diff
@@ -76,6 +79,14 @@ func Compile(jobArg *pipelinespec.PipelineJob, kubernetesSpecArg *pipelinespec.S
 		}
 	}
 
+	var semaphore_key, mutex_name string
```
gregsheremeta, line 82: camel case -- we're not in python :)

gregsheremeta (changes requested):

backend/src/apiserver/template/v2_template.go
DharmitD: Done. It now iterates over the `platformSpec.Platforms` map and adds a check to ensure that only the "kubernetes" platform is processed. It also includes the nil checks for platform and `platform.PipelineConfig` to prevent nil pointer dereference errors.

backend/src/v2/compiler/argocompiler/argo.go
DharmitD: Done :)

backend/src/apiserver/template/v2_template.go
DharmitD: Done, added these individually.

backend/src/apiserver/template/v2_template.go
DharmitD: It is copied, but the context is different.

First one, `argocompiler.Options` is a non pointer value that represents a copy of the actual data. Second one, `*argocompiler.Options` is a pointer that stores the memory address of the value.

I thought over how this code could be put into a function and called in twice, without having to reuse it. Here's a pseudo code I was thinking might work:

```
func setPipelineOptions(platformSpec map[string]*PlatformSpec, pipelineOptions interface{}) {
	
	for key, platform := range platformSpec {
		if key == "kubernetes" && platform != nil && platform.PipelineConfig != nil {
			if platform.PipelineConfig.SemaphoreKey != "" {
				  SemaphoreKey = platform.PipelineConfig.SemaphoreKey
			}
			if platform.PipelineConfig.MutexName != "" {
			    MutexName = platform.PipelineConfig.MutexName
			}
			break
		}
	}
}
```
and then just calling it each time with the pointer and the non-pointer. But it gets slightly complicated with passing a pointer into a function (needs to be de-referenced).

However, there's a pattern of repetition of the same code between pointer and non-pointer references in this file. For instance [here](https://github.com/kubeflow/pipelines/blob/497bc6b4c5a7f28f158f9616a22420d6982aa04b/backend/src/apiserver/template/v2_template.go#L63-L115), and [here](https://github.com/kubeflow/pipelines/blob/497bc6b4c5a7f28f158f9616a22420d6982aa04b/backend/src/apiserver/template/v2_template.go#L300-L348).

Which is why I would suggest keeping this repetition as is, and we can later refactor this file to remove repeated code by putting them into functions. This could be a separate task out of the scope of this PR/task.

DharmitD: Edit: Built a common helper function `getKubernetesSpec` and updated to have the function get called at both instances.

DharmitD:
Update: Added Semaphore ConfigMap name environment variable to APIServer deployment manifest, in order to provide users flexibility to set a different CM name.
If a CM name isn't set via this env var, "semaphore-config" would be the default name.

backend/src/v2/compiler/argocompiler/argo.go
```diff
@@ -28,6 +29,7 @@ import (
 	"google.golang.org/protobuf/proto"
 	"google.golang.org/protobuf/types/known/structpb"
 	k8score "k8s.io/api/core/v1"
+	v1 "k8s.io/api/core/v1"
```
gregsheremeta, line 32: the same import is just above this one :)

backend/src/v2/compiler/argocompiler/argo.go
```diff
@@ -40,6 +42,16 @@ type Options struct {
 	// optional
 	PipelineRoot string
 	// TODO(Bobgy): add an option -- dev mode, ImagePullPolicy should only be Always in dev mode.
+	SemaphoreKey string
```
gregsheremeta, line 45: I don't love using this Options struct for these fields, but I'm not sure what an alternative would be. I guess it's fine for now and maybe we can figure out a cleaner way down the line.

backend/src/v2/compiler/argocompiler/argo.go
```diff
@@ -119,6 +130,28 @@ func Compile(jobArg *pipelinespec.PipelineJob, kubernetesSpecArg *pipelinespec.S
 			Entrypoint:         tmplEntrypoint,
 		},
 	}
+	
+	if semaphoreKey != "" {
```
gregsheremeta, line 134: no need for this intermediate variable and the block above where they're set. Just use opts.SemaphoreKey

```suggestion
	if opts != nil && opts.SemaphoreKey != "" {
```

DharmitD:
Some screenshots from Argo Workflows UI, showing what the messages look like. Refer to the **Scenarios and Argo Workflows UI Verification** section in the PRs testing instructions to learn more about these messages.

Screenshot from 2024-12-13 14-23-15
Screenshot from 2024-12-13 12-42-10

manifests/kustomize/base/pipeline/ml-pipeline-semaphore-configmap.yaml
```diff
@@ -0,0 +1,5 @@
+kind: ConfigMap
```
gregsheremeta, line 1: I think this lgtm, but I just had a thought. What happens if:
1. I install kfp and I get this configmap created
2. I customize it by adding keys to it
3. I upgrade to the next version of kfp

Is that down the line upgrade going to overwrite my customized configmap with this blank one? It'd be cool if we could test that somehow -- perhaps by manually creating the configmap on a cluster, putting data in it, and then installing kfp.

gregsheremeta (changes requested):
looks good -- just a couple small things left :)

backend/src/v2/compiler/argocompiler/argo.go
DharmitD: Done, removed this import and accordingly also updated the remaining code to use the `k8score` import instead.

backend/src/v2/compiler/argocompiler/argo.go
DharmitD: Done, updated to remove the variable initialization block and used `opts.SemaphoreKey` and `opts.MutexName` directly instead.

manifests/kustomize/base/pipeline/ml-pipeline-semaphore-configmap.yaml
DharmitD: Makes sense. 
To resolve this, I've added a job to the configmap, `semaphore-configmap-init` that runs during the Kustomize deployment process.
It checks for the existence of the semaphore-config ConfigMap using kubectl get.
If the ConfigMap does not exist, it creates one with an empty init key.
If the ConfigMap already exists, the Job skips creation and exits successfully.

gregsheremeta:
lgtm

need to rebase when https://github.com/kubeflow/pipelines/pull/11384 is merged, and then I'll tag it

mprahl (changes requested):
@DharmitD could you please add one or more tests in backend/src/v2/compiler/argocompiler/argo_test.go?

manifests/kustomize/base/pipeline/ml-pipeline-semaphore-configmap.yaml
```diff
@@ -0,0 +1,29 @@
...
+apiVersion: v1
+metadata:
+  name: semaphore-config
+data: {}
+---
+apiVersion: batch/v1
```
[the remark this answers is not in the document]
mprahl: @DharmitD I'm not understanding why the ConfigMap has to exist. What happens if it doesn't exist already? Does Argo Workflow create it?

manifests/kustomize/base/pipeline/ml-pipeline-semaphore-configmap.yaml
mprahl: What applies the manifests that overwrites the fields by default? kubectl apply should just merge (and in this case do nothing) and I'm nearly certain that Argo CD will not remove existing fields unless you configure it to.

manifests/kustomize/base/pipeline/ml-pipeline-semaphore-configmap.yaml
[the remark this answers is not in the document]
DharmitD: The CM is required as the semaphore is configured via a ConfigMap. By default, the ConfigMap is named "semaphore-config", in this case, but a user can define the name through the CM. This Argo Workflows' doc can be helpful in understanding how semaphores are defined via CMs: https://argo-workflows.readthedocs.io/en/latest/synchronization/

> Does Argo Workflow create it?

No, Argo Workflow doesn't create it. We're adding it in the manifests here so users can utilize the CM to define semaphore, or user could create a new one and overwrite this existing CM if they want to define it differently.

manifests/kustomize/base/pipeline/ml-pipeline-semaphore-configmap.yaml
DharmitD: Good point, I see what you mean. Now that I think about it, it’s unlikely that any custom keys added to the ConfigMap would be removed. Given that, I believe we can keep the ConfigMap as defined in the manifests and remove the job. Wdyt @mprahl ?

manifests/kustomize/base/pipeline/ml-pipeline-semaphore-configmap.yaml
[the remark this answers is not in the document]
DharmitD: Added a comment there :)

DharmitD:
> @DharmitD could you please add one or more tests in backend/src/v2/compiler/argocompiler/argo_test.go?

Done, added a test!

manifests/kustomize/base/pipeline/ml-pipeline-semaphore-configmap.yaml
mprahl: @DharmitD that works for me!

backend/src/v2/compiler/argocompiler/argo_test.go
```diff
@@ -154,6 +156,40 @@ func Test_argo_compiler(t *testing.T) {
 
 }
 
+func TestSemaphoreAndMutex(t *testing.T) {
```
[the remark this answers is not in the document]
DharmitD: Updated to add a test that:
- tests actual `Compile()` function in `argo.go`
- Uses real pipeline data (`hello_world.json`)
- Verifies real Argo Workflow output
- Would catch bugs in the compiler logic
- Tests all combinations (both/semaphore-only/mutex-only/neither)

manifests/kustomize/base/pipeline/ml-pipeline-semaphore-configmap.yaml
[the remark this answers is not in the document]
DharmitD: I've removed this file as it's no longer needed. CM creation is now handled via the API Server.

backend/src/apiserver/resource/semaphore_manager.go
```diff
@@ -0,0 +1,86 @@
...
+		}
+		glog.Infof("Added semaphore key %s=1 to existing ConfigMap %s/%s", semaphoreKey, namespace, configMapName)
+	}
+
+	return nil
+}
+
+// getSemaphoreConfigMapName returns the name of the ConfigMap used for semaphores
+func (s *SemaphoreManager) getSemaphoreConfigMapName() string {
+	const defaultConfigMapName = "semaphore-config"
```
gmfrasca: optional: perhaps we should move to the top/outside of this func so that it can be reused/maintained in other places?  (for example, the `component` label)

DharmitD: Done, moved the labels and other constants to the top of this file / outside of the functions, so they can be reused/maintained in other places.

backend/src/apiserver/resource/semaphore_manager.go
```diff
@@ -0,0 +1,106 @@
...
+
+	// ConfigMap exists - only add missing keys, preserve user edits
+	if configMap.Data == nil {
+		configMap.Data = make(map[string]string)
+	}
+
+	// Only add the key if it doesn't exist (preserves user-modified limits)
+	if _, exists := configMap.Data[semaphoreKey]; !exists {
+		configMap.Data[semaphoreKey] = "1" // Default limit of 1
+		_, err = configMapClient.Update(ctx, configMap, metav1.UpdateOptions{})
```
mprahl, line 90: Could you ensure the API server role has the correct permissions to handle ConfigMaps?

backend/src/apiserver/resource/resource_manager.go
```diff
@@ -2048,3 +2058,137 @@ func (r *ResourceManager) GetTask(taskId string) (*model.Task, error) {
...
+
+	// Final conversion to string
+	if str, ok := current.(string); ok {
+		return str, true
+	}
+	return "", false
+}
+
+// extractSemaphoreConfig parses the pipeline manifest to extract semaphore_key and mutex_name
+func (r *ResourceManager) extractSemaphoreConfig(manifest string) (semaphoreKey, mutexName string, err error) {
```
mprahl, line 2121: This is duplicating a lot of code. Let's use `template.NewV2SpecTemplate` to parse the platform spec instead.

backend/src/apiserver/resource/resource_manager.go
```diff
@@ -558,6 +563,11 @@ func (r *ResourceManager) CreateRun(ctx context.Context, run *model.Run) (*model
 	}
 	executionSpec.SetExecutionNamespace(k8sNamespace)
 
+	// Auto-create semaphore ConfigMaps if needed
+	if err := r.ensureSemaphoreConfigMap(ctx, k8sNamespace, manifest); err != nil {
```
mprahl, line 567: Could we abstract this as a new method to the `Template` interface? Perhaps a method called `ConfigurePlatformResources` with the input being the platform spec, a `model.run`, and namespace?

My thought behind this is that `CreateRun` is pipeline engine agnostic but `ensureSemaphoreConfigMap` is very specific to Argo Workflows. So we should have it be part of the interface that pipeline engines must implement.

backend/src/apiserver/resource/resource_manager.go
```diff
@@ -2048,3 +2058,137 @@ func (r *ResourceManager) GetTask(taskId string) (*model.Task, error) {
 	}
 	return task, nil
 }
+
+// ensureSemaphoreConfigMap parses the pipeline manifest and creates semaphore ConfigMaps if needed
+func (r *ResourceManager) ensureSemaphoreConfigMap(ctx context.Context, namespace, manifest string) error {
+	// Parse the manifest to extract semaphore configuration
+	semaphoreKey, mutexName, err := r.extractSemaphoreConfig(manifest)
+	if err != nil {
+		glog.Warningf("Failed to parse semaphore configuration from manifest: %v", err)
```
mprahl, line 2067: Could you help me understand why it's safe to ignore the semaphore parsing issue?

backend/src/apiserver/resource/resource_manager.go
```diff
@@ -2048,3 +2058,137 @@ func (r *ResourceManager) GetTask(taskId string) (*model.Task, error) {
...
+	// Create ConfigMap for semaphore_key if present
+	if semaphoreKey != "" {
+		if err := r.semaphoreManager.EnsureSemaphoreConfigMap(ctx, namespace, semaphoreKey); err != nil {
+			glog.Errorf("Failed to ensure semaphore ConfigMap for key %s: %v", semaphoreKey, err)
+			return nil // Don't fail run creation for ConfigMap issues
+		}
+	}
+
+	// Create ConfigMap for mutex_name if present (same logic, treat as semaphore with limit 1)
+	if mutexName != "" {
```
mprahl, line 2080: I think it'd be better to use Argo Workflows' official mutex support. For example:

```yaml
apiVersion: argoproj.io/v1alpha1
kind: Workflow
metadata:
  generateName: synchronization-wf-level-
  namespace: foo
spec:
  synchronization:
    mutexes:
      - name: bar
```

backend/src/apiserver/resource/semaphore_manager.go
```diff
@@ -0,0 +1,106 @@
...
+
+func NewSemaphoreManager(k8sClient client.KubernetesCoreInterface) *SemaphoreManager {
+	return &SemaphoreManager{k8sClient: k8sClient}
+}
+
+// EnsureSemaphoreConfigMap creates a ConfigMap entry for the given semaphore key with default limit of 1.
+// Only creates new entries - preserves existing user edits.
+func (s *SemaphoreManager) EnsureSemaphoreConfigMap(ctx context.Context, namespace, semaphoreKey string) error {
+	if semaphoreKey == "" {
+		return nil
```
mprahl, line 52: Should this be an error instead?

backend/src/apiserver/resource/semaphore_manager.go
```diff
@@ -0,0 +1,106 @@
...
+	"github.com/kubeflow/pipelines/backend/src/apiserver/client"
+	"github.com/kubeflow/pipelines/backend/src/common/util"
+	v1 "k8s.io/api/core/v1"
+	k8errors "k8s.io/apimachinery/pkg/api/errors"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+// Common constants for semaphore functionality
+const (
+	SemaphoreConfigName = "semaphore-config"
```
mprahl, line 17: Maybe prefix it with `ml-pipeline-` so that it's easily recognized as a KFP thing?

backend/src/apiserver/resource/semaphore_manager.go
```diff
@@ -0,0 +1,106 @@
...
+				Name:      configMapName,
+				Namespace: namespace,
+				Labels:    createKFPLabels(SemaphoreConfigName),
+			},
+			Data: map[string]string{
+				semaphoreKey: "1", // Default limit of 1 (mutex behavior)
+			},
+		}
+		_, err = configMapClient.Create(ctx, configMap, metav1.CreateOptions{})
+		if err != nil {
```
mprahl, line 73: I think we should retry if the error is due to the object already existing to account for race conditions of two pipeline runs being submitted at the same time.

backend/src/apiserver/resource/semaphore_manager.go
```diff
@@ -0,0 +1,106 @@
...
+	// ConfigMap exists - only add missing keys, preserve user edits
+	if configMap.Data == nil {
+		configMap.Data = make(map[string]string)
+	}
+
+	// Only add the key if it doesn't exist (preserves user-modified limits)
+	if _, exists := configMap.Data[semaphoreKey]; !exists {
+		configMap.Data[semaphoreKey] = "1" // Default limit of 1
+		_, err = configMapClient.Update(ctx, configMap, metav1.UpdateOptions{})
+		if err != nil {
```
mprahl, line 91: This ConfigMap may be updating frequently and thus you'll get a conflicts error because the resource version changed since the `Get` call from above. I suggest making this code retry the "get and update" until it succeeds or there is an error that isn't due to conflicts.

backend/src/apiserver/resource/semaphore_manager.go
```diff
@@ -0,0 +1,106 @@
...
+		}
+		glog.Infof("Added semaphore key %s=1 to existing ConfigMap %s/%s", semaphoreKey, namespace, configMapName)
+	}
+
+	return nil
+}
+
+// getSemaphoreConfigMapName returns the name of the ConfigMap used for semaphores
+func (s *SemaphoreManager) getSemaphoreConfigMapName() string {
+	if name := os.Getenv("SEMAPHORE_CONFIGMAP_NAME"); name != "" {
```
mprahl, line 102: Could you make this a viper configuration so it can also be set in the `config.json` file?

mprahl:
@DharmitD could you please add an end to end test that with pipelines testing the semaphore and mutex features?

DharmitD:
As discussed with @mprahl and @HumairAK, the current design requires too much Kubernetes expertise from end users and creates a partially-managed ConfigMap that could be confusing to maintain.
Next Steps:
This PR will be closed in favor of a new implementation using pipeline_run_parallelism: int
The API server will auto-generate semaphore keys as <pipeline_name>/<pipeline_version>
Users will simply specify max concurrent runs (e.g., pipeline_run_parallelism=3) without needing to understand semaphores
The core ConfigMap automation logic and RBAC changes from this PR will be valuable reference for the new implementation.

PR #11075, opened 2024-08-05
feat(api): Add new ResourceSpec proto fields to support dynamic values.

KevinGrantLee:
**Description of your changes:**

Adds the fields resource_cpu_limit, resource_cpu_request, resource_memory_limit, resource_memory_request, resource_accelerator_type, and resource_accelerator_count.

**Checklist:**
- [X] The title for your pull request (PR) should follow our title convention. [Learn more about the pull request title convention used in this repository](https://github.com/kubeflow/pipelines/blob/master/CONTRIBUTING.md#pull-request-title-convention).

PR #11392, opened 2024-11-19
fix(sdk): Backport fixes in kubeflow/pipelines#11075

### LINKED ISSUE #11390: [sdk] Pipelines generated from kfp 2.10 ignore cpu/memory requests/limits
Follow-up from #11374

### Environment

*  KFP version: 2.3.0

*  KFP SDK version: 2.10.0

*  All dependencies version:

### Steps to reproduce
1. Create a python virtual environment

```
python -m venv venv
source venv/bin/activate
```

2. Install kfp 2.10

```
pip install kfp==2.10
```

3. Create the following pipeline with the file name req-test.py

```
from kfp import dsl

@dsl.component()
def pass_component():
    pass

@dsl.pipeline(name='pipeline-requests')
def pipeline_requests():
    task = pass_component()
    task.set_cpu_request('100m').set_memory_request('100Mi').set_cpu_limit('500m').set_memory_limit('500Mi')
```

4. Compile the pipeline

```
kfp dsl compile --py req-test.py --output req-test.yaml
```

5. Upload the pipeline and trigger an execution.

According to @vanHavel user:

```python
After updating to kfp SDK 2.10, we saw pods running with no resources on kfp server 2.3.
It seems that in the generated pipeline spec, the keys for CPU / RAM requests and limits have changed as well (probably in https://github.com/kubeflow/pipelines/pull/11097).
```

### Expected result
The pod should include the resources definition for CPU/Memory, including requests and limits.

### Materials and Reference
When compiling the pipeline with 2.10 it renders the following:

```
        resources:
          resourceCpuLimit: 500m
          resourceCpuRequest: 100m
          resourceMemoryLimit: 500Mi
          resourceMemoryRequest: 100Mi
```

With older version such as 2.9, it renders the following:

```
        resources:
          cpuLimit: 0.5
          cpuRequest: 0.1
          memoryLimit: 0.524288
          memoryRequest: 0.1048576
```

---

rimolive:
Introduced back the functions to convert k8s size values to float, but moved to kfp.dsl.utils

fixes #11390 

**Checklist:**

### DIFF
--- a/sdk/python/kfp/compiler/pipeline_spec_builder.py
+++ b/sdk/python/kfp/compiler/pipeline_spec_builder.py
@@ -648,16 +648,28 @@ def build_container_spec_for_task(
             container_spec.resources.resource_cpu_request = (
                 convert_to_placeholder(
                     task.container_spec.resources.cpu_request))
+            container_spec.resources.cpu_request = compiler_utils.validate_cpu_request_limit_to_float(
+                cpu=convert_to_placeholder(
+                    task.container_spec.resources.cpu_request))
         if task.container_spec.resources.cpu_limit is not None:
             container_spec.resources.resource_cpu_limit = (
                 convert_to_placeholder(task.container_spec.resources.cpu_limit))
+            container_spec.resources.cpu_limit = compiler_utils.validate_cpu_request_limit_to_float(
+                cpu=convert_to_placeholder(
+                    task.container_spec.resources.cpu_limit))
         if task.container_spec.resources.memory_request is not None:
             container_spec.resources.resource_memory_request = (
                 convert_to_placeholder(
                     task.container_spec.resources.memory_request))
+            container_spec.resources.memory_request = compiler_utils.validate_memory_request_limit_to_float(
+                memory=convert_to_placeholder(
+                    task.container_spec.resources.memory_request))
         if task.container_spec.resources.memory_limit is not None:
             container_spec.resources.resource_memory_limit = (
                 convert_to_placeholder(
+                    task.container_spec.resources.memory_limit))
+            container_spec.resources.memory_limit = compiler_utils.validate_memory_request_limit_to_float(
+                memory=convert_to_placeholder(
                     task.container_spec.resources.memory_limit))
         if task.container_spec.resources.accelerator_count is not None:
             container_spec.resources.accelerator.CopyFrom(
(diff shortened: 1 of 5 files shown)


sdk/python/kfp/compiler/pipeline_spec_builder.py
```diff
@@ -648,29 +648,41 @@ def convert_to_placeholder(input_value: str) -> str:
...
+            container_spec.resources.memory_request = compiler_utils.validate_memory_request_limit_to_float(
+                memory=convert_to_placeholder(
+                    task.container_spec.resources.memory_request))
         if task.container_spec.resources.memory_limit is not None:
             container_spec.resources.resource_memory_limit = (
                 convert_to_placeholder(
                     task.container_spec.resources.memory_limit))
+            container_spec.resources.memory_limit = compiler_utils.validate_memory_request_limit_to_float(
+                memory=convert_to_placeholder(
+                    task.container_spec.resources.memory_limit))
```
HumairAK, line 673: I think this is getting a little messy. I would prefer it if we kept the logic of how we validate the deprecated fields similar to how it used to be, and keep it in `class PipelineTask:` the way I suggest doing this is: 

Return [these](https://github.com/kubeflow/pipelines/pull/11097/files#diff-e87964fa6e1275bbe5b60d8cb7446aeb5854514987053d1c875092129b0975e0L242-L247) fields as float types, but rename them as: 

```python
    cpu_request_deprecated: Optional[float] = None
    cpu_limit_deprecated: Optional[float] = None
    memory_request_deprecated: Optional[float] = None
    memory_limit_deprecated: Optional[float] = None
```

Then move `validate_cpu_request_limit_to_float` & `validate_memory_request_limit_to_float` back in to `pipeline_task.py`, and when  `set_cpu_request`, `set_cpu_limit`, `set_memory_request`, `set_memory_limit`  is called, we set the deprecated fields the same as we did before the breaking change, and validate right before assignment. This keeps the logic functioning similarly as it used to, and it keeps all the logic close to gether so it's easier to remove it when we sunset these fields.

HumairAK:
Thanks @rimolive  can you also add this change to: https://github.com/kubeflow/pipelines/blob/master/sdk/RELEASE.md#bug-fixes-and-other-changes? otherwise lgtm

sdk/python/kfp/compiler/pipeline_spec_builder.py
rimolive: I only followed the way it's done for Accelerator fields. So Accelerator should be changed as well?

HumairAK:
@chensun PTAL when you have the chance, this is another api breaking change that was introduced in 2.10

PR #11408, opened 2024-11-26
chore: Added Dockerfiles to .dockerignore

hbelmiro:
Since [we're copying all files into the images](https://github.com/kubeflow/pipelines/blob/9ccec4c7d1aff4d2bfdb20cf4fd1f9d64b8632f4/backend/Dockerfile#L27), when we change anything in a Dockerfile, the container engine invalidates the cached layer at the `COPY . .` point. That makes subsequent builds unnecessarily long having to copy and build everything again, even when no source code was changed.

This PR adds Dockerfiles to `.dockerignore` so they are ignored when running `COPY . .`. That way we leverage the existing caches when no source code was changed.

**Checklist:**

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -3,4 +3,14 @@ bower_components
 dist
 **/node_modules
 backend/build
+backend/Dockerfile
+backend/Dockerfile.cacheserver
+backend/Dockerfile.conformance
+backend/Dockerfile.driver
+backend/Dockerfile.launcher
+backend/Dockerfile.persistenceagent
+backend/Dockerfile.scheduledworkflow
+backend/Dockerfile.viewercontroller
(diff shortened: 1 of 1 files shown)


franciscojavierarceo (approved):

HumairAK:
For the curious on docs for this: 

> You can use the .dockerignore file to exclude the Dockerfile and .dockerignore files. These files are still sent to the builder as they're needed for running the build. But you can't copy the files into the image using ADD, COPY, or bind mounts. [[1]](https://docs.docker.com/build/concepts/context/#syntax)

Since we don't keep any Dockerifles in the root, and all builds specify via `-f`, this should, I think, be a fairly safe change. 

/lgtm
/approve

PR #11458, opened 2024-12-11
fix(ui): request task logs only if pod was created

ntny:

Request the task logs only if the pod was created and the pod name is available.

Scenario: a new task in run is about to start
1. Execution record in mlmd was created with a status of Running (in the -system-container-driver- Pod)
2. -system-container-impl- pod has not yet been started and Launcher has not yet saved the pod name into the MLMD execution 
3. At this point, the Run Details/Logs tab starts pulling logs and may request them before the pod name is available, leading to an error (see attached), which could confuse the user

This issue is resolved by not requesting the log if the pod name is still unavailable. Instead of the error, the user will see an empty console until the pod is created
Снимок экрана 2024-12-11 в 22 34 55

### DIFF
--- a/frontend/src/components/tabs/RuntimeNodeDetailsV2.tsx
+++ b/frontend/src/components/tabs/RuntimeNodeDetailsV2.tsx
@@ -321,15 +321,17 @@ async function getLogsInfo(execution: Execution, runId?: string): Promise<Map<st
     return logsInfo; // Early return if it is from cache.
   }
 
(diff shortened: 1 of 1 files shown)


hbelmiro (commented):
/lgtm
/ok-to-test

hbelmiro:
cc @HumairAK

hbelmiro (commented):
/lgtm

Issue #11296, opened 2024-10-14
[backend] Updating KFP doesn't update existing scheduled runs


hbelmiro:
When we upgrade KFP, the existing `ScheduledWorkflow` doesn't have `spec.workflow.spec` updated.
That leads to subsequent runs being executed with old images and possible missing parameters.

### Workaround
Delete the scheduled run that is failing to execute, and then recreate it. Alternatively, you can duplicate the existing scheduled run.

---

Impacted by this bug? Give it a 👍.

PR #11469, opened 2024-12-16
fix(backend): Synced ScheduledWorkflow CRs on apiserver startup

hbelmiro:
Resolves: https://github.com/kubeflow/pipelines/issues/11296

This PR changes apiserver to patch existing `ScheduledWorkflow` CRs for each `Job` on startup to reflect the current KFP version deployed.

### Testing

1. Create recurring runs
2. Make sure the recurring runs are running
3. Patch their `swf` CRs to force failures
3.1 Get the `swf` CRs

    ```shell
    kubectl get swf
    NAME                             AGE
    heya8gqgg                        27m
    howdyxt5fj                       26m
    runofpipelinewithconditio5zwcj   60m
    ```
    3.2 For each `swf` patch them with an invalid workflow spec to force failures. At this point, the recurring runs will start to fail due to the invalid spec.

    ```shell
    kubectl patch scheduledworkflow heya8gqgg --type='merge' -p '{"spec":{"workflow":{"spec":{"dynamicKey1":"dynamicValue1","dynamicKey2":"dynamicValue2"}}}}'
    ```

4. Build a new apiserver image
5. Edit the `ml-pipeline` deployment to use the new apiserver image
6. The new apiserver pod will fix the `swf` CRs and the recurring runs will run successfully again

See a video demonstrating the test:

https://github.com/user-attachments/assets/a766ed10-3075-4b93-940d-e137fb092395

**Checklist:**

### DIFF
--- a/backend/src/apiserver/main.go
+++ b/backend/src/apiserver/main.go
@@ -28,6 +28,7 @@ import (
 	"os"
 	"strconv"
 	"strings"
+	"sync"
 	"time"
 
 	"github.com/fsnotify/fsnotify"
@@ -106,10 +107,25 @@ func main() {
 	}
 	log.SetLevel(level)
 
+	backgroundCtx, backgroundCancel := context.WithCancel(context.Background())
+	defer backgroundCancel()
+	wg := sync.WaitGroup{}
+	wg.Add(1)
+	go reconcileSwfCrs(resourceManager, backgroundCtx, &wg)
 	go startRpcServer(resourceManager)
+	// This is blocking
 	startHttpProxy(resourceManager)
-
+	backgroundCancel()
 	clientManager.Close()
+	wg.Wait()
+}
+
+func reconcileSwfCrs(resourceManager *resource.ResourceManager, ctx context.Context, wg *sync.WaitGroup) {
+	defer wg.Done()
+	err := resourceManager.ReconcileSwfCrs(ctx)
+	if err != nil {
+		log.Errorf("Could not reconcile the ScheduledWorkflow Kubernetes resources: %v", err)
+	}
 }
 
 // A custom http request header matcher to pass on the user identity
--- a/backend/src/apiserver/resource/resource_manager.go
+++ b/backend/src/apiserver/resource/resource_manager.go
@@ -18,8 +18,10 @@ import (
 	"context"
 	"encoding/json"
 	"fmt"
+	scheduledworkflow "github.com/kubeflow/pipelines/backend/src/crd/pkg/apis/scheduledworkflow/v1beta1"
 	"io"
 	"net"
+	"reflect"
 	"strconv"
 
 	"github.com/cenkalti/backoff"
@@ -565,6 +567,77 @@ func (r *ResourceManager) CreateRun(ctx context.Context, run *model.Run) (*model
 	}
 
 	return newRun, nil
+}
+
+// ReconcileSwfCrs reconciles the ScheduledWorkflow CRs based on existing jobs.
+func (r *ResourceManager) ReconcileSwfCrs(ctx context.Context) error {
+	filterContext := &model.FilterContext{
+		ReferenceKey: &model.ReferenceKey{Type: model.NamespaceResourceType, ID: common.GetPodNamespace()},
+	}
+
+	opts := list.EmptyOptions()
+
+	jobs, _, _, err := r.jobStore.ListJobs(filterContext, opts)
+
+	if err != nil {
+		return util.Wrap(err, "Failed to reconcile ScheduledWorkflow Kubernetes resources")
+	}
+
+	for i := range jobs {
+		select {
+		case <-ctx.Done():
+			return nil
+		default:
+		}
+
+		tmpl, _, err := r.fetchTemplateFromPipelineSpec(&jobs[i].PipelineSpec)
+		if err != nil {
+			return failedToReconcileSwfCrsError(err)
+		}
+
+		newScheduledWorkflow, err := tmpl.ScheduledWorkflow(jobs[i])
+		if err != nil {
+			return failedToReconcileSwfCrsError(err)
+		}
+
+		for {
+			currentScheduledWorkflow, err := r.getScheduledWorkflowClient(jobs[i].Namespace).Get(ctx, jobs[i].K8SName, v1.GetOptions{})
+			if err != nil {
+				if util.IsNotFound(err) {
+					break
+				}
+				return failedToReconcileSwfCrsError(err)
+			}
+
+			if !reflect.DeepEqual(currentScheduledWorkflow.Spec, newScheduledWorkflow.Spec) {
+				currentScheduledWorkflow.Spec = newScheduledWorkflow.Spec
+				err = r.updateSwfCrSpec(ctx, jobs[i].Namespace, currentScheduledWorkflow)
+				if err != nil {
+					if apierrors.IsConflict(errors.Unwrap(err)) {
+						continue
+					} else if util.IsNotFound(errors.Cause(err)) {
+						break
+					}
+					return failedToReconcileSwfCrsError(err)
+				}
+			}
+			break
+		}
+	}
+
+	return nil
+}
+
+func failedToReconcileSwfCrsError(err error) error {
+	return util.Wrap(err, "Failed to reconcile ScheduledWorkflow Kubernetes resources")
+}
+
+func (r *ResourceManager) updateSwfCrSpec(ctx context.Context, k8sNamespace string, scheduledWorkflow *scheduledworkflow.ScheduledWorkflow) error {
+	_, err := r.getScheduledWorkflowClient(k8sNamespace).Update(ctx, scheduledWorkflow)
+	if err != nil {
+		return util.Wrap(err, "Failed to update ScheduledWorkflow")
+	}
+	return nil
 }
 
 // Fetches a run with a given id.
--- a/backend/src/apiserver/client/scheduled_workflow_fake.go
+++ b/backend/src/apiserver/client/scheduled_workflow_fake.go
@@ -66,9 +66,9 @@ func (c *FakeScheduledWorkflowClient) Get(ctx context.Context, name string, opti
 	return nil, k8errors.NewNotFound(k8schema.ParseGroupResource("scheduledworkflows.kubeflow.org"), name)
 }
 
-func (c *FakeScheduledWorkflowClient) Update(context.Context, *v1beta1.ScheduledWorkflow) (*v1beta1.ScheduledWorkflow, error) {
-	glog.Error("This fake method is not yet implemented.")
-	return nil, nil
+func (c *FakeScheduledWorkflowClient) Update(_ context.Context, scheduledWorkflow *v1beta1.ScheduledWorkflow) (*v1beta1.ScheduledWorkflow, error) {
+	c.scheduledWorkflows[scheduledWorkflow.Name] = scheduledWorkflow
+	return scheduledWorkflow, nil
 }
 
 func (c *FakeScheduledWorkflowClient) DeleteCollection(ctx context.Context, options *v1.DeleteOptions, listOptions v1.ListOptions) error {
--- a/backend/src/apiserver/list/list.go
+++ b/backend/src/apiserver/list/list.go
@@ -22,6 +22,7 @@ import (
 	"encoding/base64"
 	"encoding/json"
 	"fmt"
+	"math"
 	"reflect"
 	"strings"
 
@@ -95,6 +96,13 @@ func (t *token) marshal() (string, error) {
 type Options struct {
 	PageSize int
 	*token
+}
+
+func EmptyOptions() *Options {
+	return &Options{
+		math.MaxInt32,
+		&token{},
+	}
 }
 
 // Matches returns trues if the sorting and filtering criteria in o matches that
@@ -213,9 +221,14 @@ func (o *Options) AddSortingToSelect(sqlBuilder sq.SelectBuilder) sq.SelectBuild
 	if o.IsDesc {
 		order = "DESC"
 	}
-	sqlBuilder = sqlBuilder.
-		OrderBy(fmt.Sprintf("%v %v", o.SortByFieldPrefix+o.SortByFieldName, order)).
-		OrderBy(fmt.Sprintf("%v %v", o.KeyFieldPrefix+o.KeyFieldName, order))
+
+	if o.SortByFieldName != "" {
+		sqlBuilder = sqlBuilder.OrderBy(fmt.Sprintf("%v %v", o.SortByFieldPrefix+o.SortByFieldName, order))
+	}
+
+	if o.KeyFieldName != "" {
+		sqlBuilder = sqlBuilder.OrderBy(fmt.Sprintf("%v %v", o.KeyFieldPrefix+o.KeyFieldName, order))
+	}
 
 	return sqlBuilder
 }
--- a/backend/src/apiserver/list/list_test.go
+++ b/backend/src/apiserver/list/list_test.go
@@ -15,6 +15,8 @@
 package list
 
 import (
+	"fmt"
+	"math"
 	"reflect"
 	"strings"
 	"testing"
@@ -643,6 +645,11 @@ func TestAddPaginationAndFilterToSelect(t *testing.T) {
 				},
 			},
 			wantSQL:  "SELECT * FROM MyTable ORDER BY SortField DESC, KeyField DESC LIMIT 124",
+			wantArgs: nil,
+		},
+		{
+			in:       EmptyOptions(),
+			wantSQL:  fmt.Sprintf("SELECT * FROM MyTable LIMIT %d", math.MaxInt32+1),
 			wantArgs: nil,
 		},
 		{
--- a/backend/src/apiserver/resource/resource_manager_test.go
+++ b/backend/src/apiserver/resource/resource_manager_test.go
@@ -3146,6 +3146,35 @@ func TestReportScheduledWorkflowResource_Success_withRuntimeParamsV2(t *testing.
 	assert.Equal(t, expectedJob.ToV1(), actualJob.ToV1())
 }
 
+func TestReconcileSwfCrs(t *testing.T) {
+	store, manager, job := initWithJobV2(t)
+	defer store.Close()
+
+	fetchedJob, err := manager.GetJob(job.UUID)
+	require.Nil(t, err)
+	require.NotNil(t, fetchedJob)
+
+	swfClient := store.SwfClient().ScheduledWorkflow("ns1")
+
+	options := v1.GetOptions{}
+	ctx := context.Background()
+
+	swf, err := swfClient.Get(ctx, "job-", options)
+	require.Nil(t, err)
+
+	// emulates an invalid/outdated spec
+	swf.Spec.Workflow.Spec = nil
+	swf, err = swfClient.Update(ctx, swf)
+	require.Nil(t, swf.Spec.Workflow.Spec)
+
+	err = manager.ReconcileSwfCrs(ctx)
+	require.Nil(t, err)
+
+	swf, err = swfClient.Get(ctx, "job-", options)
+	require.Nil(t, err)
+	require.NotNil(t, swf.Spec.Workflow.Spec)
+}
+
 func TestReportScheduledWorkflowResource_Error(t *testing.T) {
 	store := NewFakeClientManagerOrFatal(util.NewFakeTimeForEpoch())
 	defer store.Close()


hbelmiro:
@HumairAK @chensun can you guys PTAL?

backend/src/apiserver/main.go
```diff
@@ -106,6 +106,13 @@ func main() {
 	}
 	log.SetLevel(level)
 
+	ctx, cancel := context.WithTimeout(context.Background(), 3*time.Minute)
+	defer cancel()
+	err = resourceManager.SyncSwfCrs(ctx)
```
mprahl, line 111: Does this need to be run synchronously? Could it be run in a separate goroutine instead?

backend/src/apiserver/resource/resource_manager.go
```diff
@@ -567,6 +567,79 @@ func (r *ResourceManager) CreateRun(ctx context.Context, run *model.Run) (*model
...
+
+	opts := list.EmptyOptions()
+
+	jobs, _, _, err := r.jobStore.ListJobs(filterContext, opts)
+
+	if err != nil {
+		return util.Wrap(err, "Failed to refresh ScheduledWorkflow Kubernetes resources")
+	}
+
+	for i := 0; i < len(jobs); i++ {
```
mprahl, line 584:
```suggestion
	for i := range jobs {
```

backend/src/apiserver/main.go
mprahl, line 111: The persistence agent seems to already have a reconcile loop for scheduled workflows. If I'm reading the code right, on start up, it'll reconcile everything and then handle creates, updates, and deletes.

Could the migration logic be added to the persistence agent instead?

hbelmiro: It probably could. But that doesn't sound like a persistence agent's responsibility. It doesn't sound like API server's responsibility too, but I think it fits better there since we have a [`jobStore`](https://github.com/kubeflow/pipelines/blob/953426db9895a64c380370be19c716d009a10250/backend/src/apiserver/resource/resource_manager.go#L578).

My original plan was to do it in https://github.com/kubeflow/pipelines/tree/master/backend/src/crd/controller/scheduledworkflow, but we would have to make HTTP calls to the API server. Then we decided to leave it in the API server.

It would be nice to hear others' opinions about that.

mprahl: @hbelmiro I agree that persistence agent isn't the best fit. That controller you linked could be a good fit.

I think we should consider adding a KFP version column to the `jobs` table so that you can skip generating and updating scheduled workflows if the version matches.

If were to pursue using the scheduled workflow controller, here is an idea:

* Have the scheduled workflow controller query the API server health endpoint at start up to get the `tag_name` value to see what version of KFP we are on. In the background, it could keep querying the API server to see if the version changed.
* The scheduled workflow controller's reconcile loop checks an annotation of `pipelines.kubeflow.org/version` and if the value of that annotation doesn't match `tag_name`, then the workflow definition is updated and the annotation value is set to the current version.
* When the API server creates a `ScheduledWorkflow` object, it sets the `pipelines.kubeflow.org/version` annotation.

backend/src/apiserver/main.go
```diff
@@ -106,12 +106,23 @@ func main() {
...
 }
 
+func reconcileSwfCrs(resourceManager *resource.ResourceManager) {
+	ctx, cancel := context.WithTimeout(context.Background(), 3*time.Minute)
+	defer cancel()
+	err := resourceManager.ReconcileSwfCrs(ctx)
+	if err != nil {
+		log.Errorf("Could not reconcile the ScheduledWorkflow Kubernetes resources: %v", err)
+	}
+}
```
mprahl, line 124: I know the rest of the code doesn't do this for separate go routines but I think the correct approach would be something like this to properly wait for the goroutine to finish.
```suggestion
	backgroundCtx, backgroundCancel := context.WithCancel(context.Background())
	defer backgroundCancel()
	wg := sync.WaitGroup{}
	wg.Add(1)
	go reconcileSwfCrs(resourceManager, ctx, &wg)

	go startRpcServer(resourceManager)
	// This is blocking
	startHttpProxy(resourceManager)

	backgroundCancel()
	clientManager.Close()
	wg.Wait()
}

func reconcileSwfCrs(resourceManager *resource.ResourceManager, ctx context.Context, wg *sync.WaitGroup) {
	defer wg.Done()
	err := resourceManager.ReconcileSwfCrs(ctx)
	if err != nil {
		log.Errorf("Could not reconcile the ScheduledWorkflow Kubernetes resources: %v", err)
	}
}
```

backend/src/apiserver/main.go
```diff
@@ -106,12 +106,23 @@ func main() {
...
+	go reconcileSwfCrs(resourceManager)
+
 	go startRpcServer(resourceManager)
 	startHttpProxy(resourceManager)
 
 	clientManager.Close()
 }
 
+func reconcileSwfCrs(resourceManager *resource.ResourceManager) {
+	ctx, cancel := context.WithTimeout(context.Background(), 3*time.Minute)
```
mprahl, line 118: I don't think we need a timeout context here since we don't really care if it takes longer than 3 minutes when it's run asynchronously. A regular background context should be okay.

backend/src/apiserver/main.go
```diff
@@ -106,12 +106,23 @@ func main() {
...
+	defer cancel()
+	err := resourceManager.ReconcileSwfCrs(ctx)
+	if err != nil {
+		log.Errorf("Could not reconcile the ScheduledWorkflow Kubernetes resources: %v", err)
```
mprahl, line 122: I agree with your approach of not exiting with an error code in this case, but I'm curious what others think. My concern is that it'd be hard to know that this failed as a KFP admin but it also doesn't seem warranted to keep the API server from running if this can't succeed.

backend/src/apiserver/resource/resource_manager.go
```diff
@@ -567,6 +567,79 @@ func (r *ResourceManager) CreateRun(ctx context.Context, run *model.Run) (*model
...
+	return util.Wrap(err, "Failed to reconcile ScheduledWorkflow Kubernetes resources")
+}
+
+func (r *ResourceManager) patchSwfCrSpec(ctx context.Context, k8sNamespace string, crdName string, newSpec interface{}) error {
+	if k8sNamespace == "" {
+		k8sNamespace = common.GetPodNamespace()
+	}
+	if k8sNamespace == "" {
+		return errors.New("Namespace cannot be empty when deleting a ScheduledWorkflow Kubernetes resource.")
+	}
```
mprahl, line 617: When is this code needed? It seems the jobs table always has the namespace set.

backend/src/apiserver/resource/resource_manager.go
```diff
@@ -567,6 +567,79 @@ func (r *ResourceManager) CreateRun(ctx context.Context, run *model.Run) (*model
...
+		if err != nil {
+			return failedToReconcileSwfCrsError(err)
+		}
+
+		scheduledWorkflow, err := tmpl.ScheduledWorkflow(jobs[i])
+		if err != nil {
+			return failedToReconcileSwfCrsError(err)
+		}
+
+		err = r.patchSwfCrSpec(ctx, jobs[i].Namespace, jobs[i].K8SName, scheduledWorkflow.Spec)
```
mprahl, line 595: Could you use a ` reflect.DeepEquals` check to compare the desired spec and the current spec to avoid patches when the value is already correct?

backend/src/apiserver/resource/resource_manager.go
```diff
@@ -567,6 +567,79 @@ func (r *ResourceManager) CreateRun(ctx context.Context, run *model.Run) (*model
...
+		"spec": newSpec,
+	}
+
+	patchBytes, err := json.Marshal(patchPayload)
+	if err != nil {
+		return util.NewInternalServerError(err,
+			"Failed to marshal patch spec")
+	}
+
+	_, err = r.getScheduledWorkflowClient(k8sNamespace).Patch(
```
mprahl, line 629: Could you perform an update instead of a patch? That way it'll catch `resourceVersion` mismatches (i.e. ScheduledWorkflow was updated between when the object was retrieved and the update request). It'd be good to also retry the whole job iteration flow in that event. The `IsConflict` function from `"k8s.io/apimachinery/pkg/api/errors"` should help detect that specific case.

backend/src/apiserver/resource/resource_manager.go
```diff
@@ -567,6 +567,79 @@ func (r *ResourceManager) CreateRun(ctx context.Context, run *model.Run) (*model
...
+	}
+
+	for i := range jobs {
+		tmpl, _, err := r.fetchTemplateFromPipelineSpec(&jobs[i].PipelineSpec)
+		if err != nil {
+			return failedToReconcileSwfCrsError(err)
+		}
+
+		scheduledWorkflow, err := tmpl.ScheduledWorkflow(jobs[i])
+		if err != nil {
```
mprahl, line 591: Should this ignore not found errors? I could see the case where the database still has a reference to the scheduled workflow but the object no longer exists on the Kubernetes cluster.

backend/src/apiserver/main.go
mprahl: Sure, no problem! What do you think of the other comnment of adding a KFP version column to the jobs table so that you can skip generating and updating scheduled workflows if the version matches?

hbelmiro: I like that too. But other than https://github.com/kubeflow/pipelines/blob/master/VERSION, I'm not aware of a field/file from where I can get the version. Are you?
Otherwise, maybe @HumairAK and @chensun can recommend something.

backend/src/apiserver/resource/resource_manager.go
hbelmiro: I'm not sure I'm following. Are you suggesting to ignore the error when the CR is not found? I'm already doing that, but not at this line. `tmpl.ScheduledWorkflow(jobs[i])` actually just converts a `Job` object to a `ScheduleWorkflow` object.

backend/src/apiserver/main.go
hbelmiro: @mprahl where do you see it?

backend/src/apiserver/main.go
hbelmiro: Done.

backend/src/apiserver/main.go
hbelmiro: Done.

backend/src/apiserver/resource/resource_manager.go
hbelmiro: You're right. Removed.

backend/src/apiserver/resource/resource_manager.go
hbelmiro: Done.

backend/src/apiserver/resource/resource_manager.go
hbelmiro: Cool :)
Done.

hbelmiro:
This PR depends on:

- https://github.com/kubeflow/pipelines/pull/11480
- https://github.com/kubeflow/pipelines/pull/11481

I included those commits here while the PRs are not merged.

cc @mprahl

backend/src/apiserver/main.go
jgarciao: @mprahl @hbelmiro beware that, in the past, we had also problems with scheduled runs after upgrading RHOAI:

If I recall well, the problem in https://issues.redhat.com/browse/RHOAIENG-10790 was that the kfp-launcher pod needed an additional parameter to be set and the scheduled runs created before the upgrade didn't provide that parameter, so they failed to run. In this case it would have been helpful to rebuild all scheduled runs on kfp api server startup, regardless the kfp api server version.

What about providing an environment variable to control the behavior?
`ML_PIPELINE_SCHEDULEDRUNS_REBUILDPOLICY=[disable,always,...]`

backend/src/apiserver/resource/resource_manager.go
```diff
@@ -567,6 +569,71 @@ func (r *ResourceManager) CreateRun(ctx context.Context, run *model.Run) (*model
...
+
+	opts := list.EmptyOptions()
+
+	jobs, _, _, err := r.jobStore.ListJobs(filterContext, opts)
+
+	if err != nil {
+		return util.Wrap(err, "Failed to reconcile ScheduledWorkflow Kubernetes resources")
+	}
+
+nextJob:
```
mprahl, line 586: I don't think you need gotos. You could change `continue retryJob` to just `continue` and `continue nextJob` to `break`.

backend/src/apiserver/resource/resource_manager.go
```diff
@@ -567,6 +569,71 @@ func (r *ResourceManager) CreateRun(ctx context.Context, run *model.Run) (*model
...
+			if err != nil {
+				return failedToReconcileSwfCrsError(err)
+			}
+
+			newScheduledWorkflow, err := tmpl.ScheduledWorkflow(jobs[i])
+			if err != nil {
+				return failedToReconcileSwfCrsError(err)
+			}
+
+			currentScheduledWorkflow, err := r.getScheduledWorkflowClient(jobs[i].Namespace).Get(ctx, jobs[i].K8SName, v1.GetOptions{})
```
mprahl, line 600: I think the `for {` loop only needs to be around this part since `tmpl` won't change since you aren't refetching `jobs[i]` from the database.

backend/src/apiserver/resource/resource_manager.go
```diff
@@ -567,6 +569,71 @@ func (r *ResourceManager) CreateRun(ctx context.Context, run *model.Run) (*model
...
+			if err != nil {
```
mprahl, line 601: Ignoring a not found error is probably okay here as well.

backend/src/apiserver/resource/resource_manager.go
```diff
@@ -567,6 +569,71 @@ func (r *ResourceManager) CreateRun(ctx context.Context, run *model.Run) (*model
...
+				return failedToReconcileSwfCrsError(err)
+			}
+
+			if !reflect.DeepEqual(currentScheduledWorkflow.Spec, newScheduledWorkflow.Spec) {
+				newScheduledWorkflow.Name = currentScheduledWorkflow.Name
```
mprahl, line 606: I think I'd rather take the approach of replacing the spec on `currentScheduledWorkflow` and updating that. This is to not overwrite any metadata the user assigned to it (i.e. managed by Argo CD label or similar).

backend/src/apiserver/resource/resource_manager.go
hbelmiro: Done.

backend/src/apiserver/resource/resource_manager.go
hbelmiro: Done.

backend/src/apiserver/resource/resource_manager.go
hbelmiro: Done.

backend/src/apiserver/resource/resource_manager.go
```diff
@@ -567,6 +569,71 @@ func (r *ResourceManager) CreateRun(ctx context.Context, run *model.Run) (*model
...
+	opts := list.EmptyOptions()
+
+	jobs, _, _, err := r.jobStore.ListJobs(filterContext, opts)
+
+	if err != nil {
+		return util.Wrap(err, "Failed to reconcile ScheduledWorkflow Kubernetes resources")
+	}
+
+	for i := range jobs {
+		tmpl, _, err := r.fetchTemplateFromPipelineSpec(&jobs[i].PipelineSpec)
```
mprahl, line 587: Optional: On each iteration, you can check if the context was canceled rather than letting the `r.getScheduledWorkflowClient(jobs[i].Namespace).Get` call fail below. It's just a little cleaner but not really needed.

```suggestion
		select {
		case <-ctx.Done():
			return nil
		default:
		}

		tmpl, _, err := r.fetchTemplateFromPipelineSpec(&jobs[i].PipelineSpec)
```

mprahl:
@hbelmiro is there any way to add test coverage for this?

hbelmiro:
> @hbelmiro is there any way to add test coverage for this?

@mprahl I'm not sure it's worth it. I'd have to replicate [this reproducer](https://github.com/kubeflow/pipelines/pull/11469#issue-2743172116), which means recreating the API Server pod. That would probably affect other tests.

mprahl:
> > @hbelmiro is there any way to add test coverage for this?
> 
> @mprahl I'm not sure it's worth it. I'd have to replicate [this reproducer](https://github.com/kubeflow/pipelines/pull/11469#issue-2743172116), which means recreating the API Server pod. That would probably affect other tests. Also, if we implement the version/image check, it will be even harder to test.

I was thinking you could have a test that calls `ReconcileSwfCrs` with the fake client manager. See `backend/src/apiserver/resource/resource_manager_test.go` for examples of where this is used. You should be able to create an invalid `ScheduledWorkflow` in the fake store (in memory SQLite database), call `ReconcileSwfCrs`, and verify it was updated.

backend/src/apiserver/main.go
jgarciao: If the api server is still running users will be able to delete or clone the failing scheduled runs using the UI

hbelmiro: That's a good point @jgarciao.

hbelmiro:
> > > @hbelmiro is there any way to add test coverage for this?
> > 
> > 
> > @mprahl I'm not sure it's worth it. I'd have to replicate [this reproducer](https://github.com/kubeflow/pipelines/pull/11469#issue-2743172116), which means recreating the API Server pod. That would probably affect other tests. Also, if we implement the version/image check, it will be even harder to test.
> 
> I was thinking you could have a test that calls `ReconcileSwfCrs` with the fake client manager. See `backend/src/apiserver/resource/resource_manager_test.go` for examples of where this is used. You should be able to create an invalid `ScheduledWorkflow` in the fake store (in memory SQLite database), call `ReconcileSwfCrs`, and verify it was updated.

@mprahl actually what needs to be updated is the CR. But that's a good idea anyway. I've added the test.

mprahl (approved):

mprahl (approved):
/lgtm

backend/src/apiserver/main.go
hbelmiro: @mprahl this discussion is still pending.

Since we're now [checking if the current `swf` is deeply equal the new one](https://github.com/kubeflow/pipelines/blob/e2070263f71bd7dab7b5d855fa71cce35c23a423/backend/src/apiserver/resource/resource_manager.go#L612), I think we could skip this verification.
We would still be fetching the current `swf` and generating the new one unnecessarily, but I think that's a fair price to avoid adding more complexity.

WDYT?

I'd like to have @jgarciao @HumairAK's opinion on this too.

(My opinion on this is not strong and I'm fine with implementing this if you guys think it's better)

mprahl: My opinion is that the current code is good enough for now and the version metadata isn't needed at the moment. I think the extra version data would be nice but I'm thinking there isn't likely to be very many scheduled workflows, so the extra cost that happens asynchronously seems okay to me.

HumairAK:
this looks good, happy to approve but @hbelmiro can you squash down some of these commit messages?

backend/src/apiserver/main.go
HumairAK: I think given the go routine approach, I think we have implicitly decided that the rest of the api server's operation are not dependent on this sync op, so I don't see the point in handling the error beyond just logging it (which we are already doing in the routine) so imo it should be fine.

backend/src/apiserver/main.go
```diff
@@ -106,10 +107,25 @@ func main() {
 	}
 	log.SetLevel(level)
 
+	backgroundCtx, backgroundCancel := context.WithCancel(context.Background())
+	defer backgroundCancel()
```
HumairAK, line 111: @hbelmiro / @mprahl looks like we are cancelling the context at the end anyway, is this second call needed?

hbelmiro: I think we can remove [this call](https://github.com/kubeflow/pipelines/pull/11469/files#diff-d3ae27a3b5003f91b2cbf384575c21f07dd142165daf842fc79fd8459a844829R118) instead.

mprahl: There's no harm in keeping it since canceling an already canceled context is a no-op. The "benefit" I suppose is if the HTTP server panics, the context would still cancel but the rest of the code isn't going to wait for goroutine to exit cleanly anyways. Feel free to remove this one but we need to keep the other call or else `wg.Wait()` will finish only after `reconcileSwfCrs` is done.

mprahl (approved):
/lgtm

HumairAK:
/approve

thanks guys!

PR #11475, opened 2024-12-19
fix(backend): modelToCRDTrigger was not including periodic schedule correctly

hbelmiro:
This PR fixes the issue where `CronSchedule` and `PeriodicSchedule` structs were incorrectly evaluated as non-empty even when their fields were all nil. The issue arose because direct struct comparison with `(CronSchedule{})` or `(PeriodicSchedule{})` does not account for the semantics of pointer fields being `nil`, `""` or `0`.

**Checklist:**

### DIFF
--- a/backend/src/apiserver/model/cron_schedule.go
+++ b/backend/src/apiserver/model/cron_schedule.go
@@ -1,0 +1,21 @@
+package model
+
+type CronSchedule struct {
+	// Time at which scheduling starts.
+	// If no start time is specified, the StartTime is the creation time of the schedule.
+	CronScheduleStartTimeInSec *int64 `gorm:"column:CronScheduleStartTimeInSec;"`
+
+	// Time at which scheduling ends.
+	// If no end time is specified, the EndTime is the end of time.
+	CronScheduleEndTimeInSec *int64 `gorm:"column:CronScheduleEndTimeInSec;"`
+
+	// Cron string describing when a workflow should be created within the
+	// time interval defined by StartTime and EndTime.
+	Cron *string `gorm:"column:Schedule;"`
+}
+
+func (c CronSchedule) IsEmpty() bool {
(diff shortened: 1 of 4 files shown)


backend/src/apiserver/model/cron_schedule.go
```diff
@@ -0,0 +1,21 @@
+package model
```
mprahl, line 1: I noticed all the other files have copyright headers. Do these new files need those also?

backend/src/apiserver/model/cron_schedule.go
```diff
@@ -0,0 +1,21 @@
...
+
+	// Cron string describing when a workflow should be created within the
+	// time interval defined by StartTime and EndTime.
+	Cron *string `gorm:"column:Schedule;"`
+}
+
+func (c CronSchedule) IsEmpty() bool {
+	return c.CronScheduleStartTimeInSec == nil &&
+		c.CronScheduleEndTimeInSec == nil &&
+		(c.Cron == nil || *c.Cron == "")
```
mprahl, line 20: Why is the empty case checked for the `Cron` field but not the other fields (e.g. `0` for `CronScheduleStartTimeInSec`)?

backend/src/apiserver/model/cron_schedule.go
hbelmiro: That's a good question. I've never come across any indication that they are necessary.
Anyways... Added just in case.

backend/src/apiserver/model/cron_schedule.go
hbelmiro: Good catch!
Added for all fields.

mprahl (approved):
/lgtm

mprahl (approved):
/lgtm

Let's try this again...

PR #11462, opened 2024-12-12
chore:  Add securitycontext for PSS PoC (rootless Kubeflow)

juliusvonkohout:
**Description of your changes:**

upstream what we have in https://github.com/kubeflow/manifests/tree/master/contrib/security/PSS to make PSS enforcable
and enjoy a rootless Kubeflow https://github.com/kubeflow/manifests/issues/2528

**Checklist:**
- [x] You have [signed off your commits](https://www.kubeflow.org/docs/about/contributing/#sign-off-your-commits)
- [x] The title for your pull request (PR) should follow our title convention. [Learn more about the pull request title convention used in this repository](https://github.com/kubeflow/pipelines/blob/master/CONTRIBUTING.md#pull-request-title-convention).

PR #11487, opened 2024-12-26
fix(manifests): Revert PSS changes in manifests

rimolive:
The changes introduced in #11462 broke manifests deployment in OpenShift environments. We need a more coordinated effort to add these PSS requirements as other WGs are requesting.

cc @HumairAK @juliusvonkohout 

**Checklist:**

### DIFF
--- a/manifests/kustomize/base/cache/cache-deployment.yaml
+++ b/manifests/kustomize/base/cache/cache-deployment.yaml
@@ -16,16 +16,6 @@ spec:
     spec:
       containers:
       - name: server
-        securityContext:
-          allowPrivilegeEscalation: false
-          seccompProfile:
-            type: RuntimeDefault
-          runAsNonRoot: true
-          runAsUser: 1000
-          runAsGroup: 0
-          capabilities:
-            drop:
-            - ALL
         image: gcr.io/ml-pipeline/cache-server:dummy
         env:
           - name: DEFAULT_CACHE_STALENESS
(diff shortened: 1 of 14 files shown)


juliusvonkohout:
@rimolive is that not a distribution specific thing, so openshift only and not Kubernetes in general? I ran Kubeflow on openshift before and it is possible to do it rootless. We are really following Kubernetes best practices here according to https://kubernetes.io/docs/concepts/security/pod-security-standards/. So I propose to not revert it, but rather that you post the error message here and we try to fix it for openshift.

juliusvonkohout:
@rimolive if there are confidential manifests involved we can also do a private session.

rimolive:
@juliusvonkohout There are no confidential information involved, no worries! I'm talking specifically about the `runAs*` attributes that broke OpenShift. That's because we make use of random UIDs and setting these values explicitly is not allowed. I can't speak of the other changes, but I also reviewed the PSS requirements and I noticed that it is allowed to not set any of these `runAs*` values. For now, I'll plan to revert only this and see if the rest keeps working in OpenShift I'll make the proper changes.

/hold

juliusvonkohout:
> @juliusvonkohout There are no confidential information involved, no worries! I'm talking specifically about the `runAs*` attributes that broke OpenShift. That's because we make use of random UIDs and setting these values explicitly is not allowed. I can't speak of the other changes, but I also reviewed the PSS requirements and I noticed that it is allowed to not set any of these `runAs*` values. For now, I'll plan to revert only this and see if the rest keeps working in OpenShift I'll make the proper changes.
> 
> /hold

Can you just comment out the runas with a hashtag to #runAs: ...? This way we keep all information for later. It could be that we need to modify some dockerfiles to runasanyuser.

juliusvonkohout:
/hold

@rimolive is this still relevant?

Issue #11500, opened 2025-01-06
[backend] ResourceCpuLimit, ResourceCpuRequest, ResourceMemoryLimit, and ResourceMemoryRequest are ignored


mprahl:
### Environment

*  How did you deploy Kubeflow Pipelines (KFP)?

I used the manifests from the master branch.

*  KFP version: 

Latest from the master branch.

*  KFP SDK version: 

2.11

### Steps to reproduce

The Driver ignores the new fields ResourceCpuLimit, ResourceCpuRequest, ResourceMemoryLimit, and ResourceMemoryRequest. The API fields were added in this commit but the Driver code is only looking at the now deprecated fields:
https://github.com/kubeflow/pipelines/commit/c2f56495b9b1e9eda1b44b6106e12d5290a89ed7#diff-42023372ccec7209826b2d45768cbac5c1181a3a79645c3e56b9527702b29572L29

```json
          "resources": {
            "resourceCpuLimit": "{{$.inputs.parameters['pipelinechannel--cpu_limit']}}",
            "resourceCpuRequest": "{{$.inputs.parameters['pipelinechannel--cpu_request']}}"
          }
```

### Expected result

The new fields should be used first and then fallback to the deprecated fields.

### Materials and Reference

Related SDK commit:
https://github.com/boarder7395/pipelines/commit/10f1dd5e4594a43035db323f43cc78244ce7bf72

Here is an example reproducer:

```json
{
  "components": {
    "comp-generate-text": {
      "executorLabel": "exec-generate-text",
      "outputDefinitions": {
        "parameters": {
          "Output": {
            "parameterType": "STRING"
          }
        }
      }
    },
    "comp-print-text": {
      "executorLabel": "exec-print-text",
      "inputDefinitions": {
        "parameters": {
          "text": {
            "parameterType": "STRING"
          }
        }
      }
    }
  },
  "deploymentSpec": {
    "executors": {
      "exec-generate-text": {
        "container": {
          "args": [
            "--executor_input",
            "{{$}}",
            "--function_to_execute",
            "generate_text"
          ],
          "command": [
            "sh",
            "-c",
[...]

Issue #11375, opened 2024-11-14
chore(backend): use non deprecated accelerator type & count fields in driver


HumairAK:
## Chore description

Driver should use `resource_type` and `resource_count` and move away from deprecated `type` and `count` fields. Api sec for these fields are [here](https://github.com/kubeflow/pipelines/blob/master/api/v2alpha1/pipeline_spec.proto#L770-L783). 

Driver calls for these fields are [here](https://github.com/kubeflow/pipelines/blob/master/backend/src/v2/driver/driver.go#L449-L457). 

This PR and comment for more context: https://github.com/kubeflow/pipelines/pull/11373#issuecomment-2474762540

### Labels

/area backend 

---

Love this idea? Give it a 👍.

mahdikhashan:
I would like to take over this issue (as my first code pr) and start working on it. would it be possible to assign it to me?

mprahl:
@HumairAK can this be closed due to https://github.com/kubeflow/pipelines/pull/11501 being merged?

HumairAK:
yes

PR #11501, opened 2025-01-06
feat(sdk/backend): Add support for placeholders in resource limits

mprahl:

The API introduced new fields prefixed with Resource (e.g. ResourceCpuLimit) to replace the old fields without the prefix. The Driver hadn't been updated to honor those fields but the SDK started using them which led to unexpected behavior.

The Driver now honors both fields but prioritizes the new fields. The SDK now only sets the new fields.

The outcome is that resource limits/requests can now use input parameters.

Note that pipeline_spec_builder.py was doing some validation on the limits/requests being set, but that's already handled in the user facing method (e.g. set_cpu_limit).

Resolves:
https://github.com/kubeflow/pipelines/issues/11500

**Checklist:**

### DIFF
--- a/backend/src/v2/driver/driver.go
+++ b/backend/src/v2/driver/driver.go
@@ -362,6 +362,35 @@ func Container(ctx context.Context, opts Options, mlmd *metadata.Client, cacheCl
 	return execution, nil
 }
 
+// getPodResource will accept the new field that accepts placeholders (e.g. resourceMemoryLimit) and the old float64
+// field (e.g. memoryLimit) and return the resolved value as a Quantity. If the returned Quantity is nil, it was not set
+// by the user. If the new field is set, the old field is ignored.
+func getPodResource(
+	new string, old float64, executorInput *pipelinespec.ExecutorInput, oldFmtStr string,
+) (*k8sres.Quantity, error) {
+	var resolved string
+
+	if new != "" {
+		var err error
+
+		resolved, err = resolvePodSpecInputRuntimeParameter(new, executorInput)
+		if err != nil {
+			return nil, fmt.Errorf("failed to resolve executor input when retrieving pod resource: %w", err)
+		}
+	} else if old != 0 {
+		resolved = fmt.Sprintf(oldFmtStr, old)
+	} else {
+		return nil, nil
+	}
+
+	q, err := k8sres.ParseQuantity(resolved)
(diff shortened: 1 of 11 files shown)


mprahl:
/cc @HumairAK @rimolive 

Note that I labeled this as a feature but it's also a bug fix as mentioned in the Slack thread here:
https://cloud-native.slack.com/archives/C073N7BMLB1/p1735746170303549

Additionally, it seems that this recent unreleased commit added placeholder support to the wrong accelerator type field, so it'd be nice to merge this PR before the next KFP release:
https://github.com/kubeflow/pipelines/pull/11404

tomzx:
I've tested the SDK portion as we're eagerly awaiting the capability of setting cpu/ram/accelerator type/count dynamically in our project.
I can confirm that it works, as in it passes compilation and can be submitted to GCP Vertex AI (for example).

I also see the following in the compiled pipeline configuration for a step using those dynamic properties.
```yaml
        resources:
          accelerator:
            resourceCount: '{{$.inputs.parameters[''pipelinechannel--some-step-accelerator_count'']}}'
            resourceType: '{{$.inputs.parameters[''pipelinechannel--some-step-accelerator_type'']}}'
          resourceCpuLimit: '{{$.inputs.parameters[''pipelinechannel--some-step-cpu'']}}'
          resourceMemoryLimit: '{{$.inputs.parameters[''pipelinechannel--some-step-ram'']}}'
```

HumairAK:
Note that this also resolves: https://github.com/kubeflow/pipelines/issues/11375

Looks like a safe change. Folks pre-kfp 2.4 and post 2.4 will have continued support of old fields, even with pipelines rendered pre sdk 2.10, when submitted. So removing them from the pipeline render makes sense. Recompiling will remove these old fields, which does seem like it's api-changing (the api is implicit here in the yaml even when not defined in the proto) but looks backwards compatible since we'll continue to handle it on driver side. Since initial change completely removed it from vertex side, I do not think this is likely to have downstream effects there, but regardless fyi @chensun.  

I tested this with deprecated and new values, and with parameter values which are correctly resolved. Which is great, thanks @mprahl! 

some nit comments above @mprahl upto you to address them or not, let me know otherwise I can approve 

/lgtm

HumairAK:
/lgtm
/approve

Thanks!

PR #11518, opened 2025-01-15
fix(frontend): fixes Default pipeline input params are missing from the GUI. Fixes #11515

### LINKED ISSUE #11515: [frontend] Default pipeline input params are missing from the GUI
When default pipeline input parameters, such as False or an empty string, are used, the GUI displays a `Missing parameter` error when attempting to run the pipeline. This issue does not occur with a `True` value. It prevents the pipeline from running unless the values are manually filled.
### Environment

Running on openshift 4.16.23

*  KFP version: 
kubeflow image: `gcr.io/ml-pipeline/api-server:2.3.0`

### Steps to reproduce

Run the following pipeline:

```bash
import kfp.compiler as compiler
import kfp.dsl as dsl

def print_params(
    bool_param: bool,
    str_param: str
) -> None:
    print(bool_param)
    print(str_param)

print_params_op = dsl.component_decorator.component(func=print_params, base_image='python:3.9')

@dsl.pipeline(
    name="example",
    description="Pipeline for default values task",
)
def example(bool_param: bool = False, str_param: str = ''):
    print_params_op(bool_param=bool_param, str_param=str_param)


if __name__ == "__main__":
    # Compiling the pipeline
    compiler.Compiler().compile(example, __file__.replace(".py", ".yaml"))
```

Trying to run the pipeline gives an error in the GUI:

Image

### Materials and Reference

---

ElayAharoni:
Fixes #11515.

fixed the issue.
current behavior after the fix with the same pipeline.

image

### DIFF
--- a/frontend/src/components/NewRunParametersV2.tsx
+++ b/frontend/src/components/NewRunParametersV2.tsx
@@ -205,7 +205,7 @@ function NewRunParametersV2(props: NewRunParametersProps) {
     let allParamtersWithDefault = true;
     let errMsg: string[] = [];
     Object.keys(specParameters).forEach(key => {
-      if (specParameters[key].defaultValue) {
+      if (specParameters[key].defaultValue !== undefined) {
         // TODO(zijianjoy): Make sure to consider all types of parameters.
         // Convert default value to string type first to avoid error from convertInput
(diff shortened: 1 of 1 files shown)


HumairAK:
thanks @ElayAharoni 

/lgtm
/approve

PR #11585, opened 2025-02-04
fix(backend) fix run retry for argo workflow v3.4.16

ntny:
**Fix Argo Workflow v3.4.16 Retry**

Argo Workflow upgrade broke the Run Retry

From [upgrade descriotion](https://github.com/kubeflow/pipelines/commit/809d5766fc9ec436ff05c083e9a2ae65ad2667b7): Argo 3.4.16 upgrade introduces a breaking change with inconsistent node.ID vs
  node.Name

However as you can see by [link](https://github.com/argoproj/argo-workflows/blob/v3.4.16/workflow/util/util.go#L879) v3.4.16 of Argo still operates nodes by ID during retry.

Also i added the missing [logic](https://github.com/argoproj/argo-workflows/blob/v3.4.16/workflow/util/util.go#L973)(fix node.Children/node.OutboundNodes) from the Argo Workflow retry

Using the node name instead of the ID caused Argo to fail to identify the task that needed to be retried. As a result, it would restart the entire workflow from the beginning, attempt to recreate the pod for the -system-dag-driver-, fail at this step (since such a pod already existed), and leave the run stuck in the Running state

according logs from the argo workflow-controller
`level=info msg="Failed pod debug-component-pipeline-d6wjn.root-driver (debug-component-pipeline-d6wjn-system-dag-driver-2907082658) creation: already exists" namespace=arpechenin workflow=debug-component-pipeline-d6wjn`

This MR is based on the assumption that when using KFP 2.3.0 & Argo 3.4.16, retries do not work on any installation, not just for some users. It has fixed Run retry on my installation.


juliusvonkohout:
/ok-to-test

CC @rimolive @varodrig @HumairAK @hbelmiro @kubeflow/release-team

juliusvonkohout:
Related to https://github.com/kubeflow/manifests/issues/2952

juliusvonkohout:
@kubeflow/release-team tracked as Kubeflow regression and for Kubeflow 1.10 in https://github.com/kubeflow/manifests/issues/2952

juliusvonkohout:
@HumairAK @rimolive

hbelmiro (commented):
/lgtm

Issue #11584, opened 2025-02-04
[feature] Add support for model artifacts in the "modelcar" format


mprahl:
### Feature Area

/area backend

### What feature would you like to see?

Currently, a user may import a model as an artifact with the KFP SDK like the following example, but it's limited to S3.

```python
model_source_s3_task = dsl.importer(artifact_uri=model_in_s3, artifact_class=dsl.Model)
```

It'd be nice to add support for the [modelcar format](https://kserve.github.io/website/latest/modelserving/storage/oci/#using-modelcars) that is supported in KServe.

The suggestion is for the artifact URI passed to `dsl.importer` to have a prefix of `oci://`. No changes to the SDK should be needed for this. 

The KFP driver would see this and add an init container to the launcher pod which launches the modelcar container image and copies the model to an `emptyDir` volume. The artifact output would be the local path to this `emptyDir` volume in the launcher container. The benefit of doing it this way allows for the global image pull secret or image pull secrets assigned to the KFP pipeline runner service account to pull it without specifying additional credentials.

Once OCI Volume Source has been GA for several Kubernetes releases, we may consider leverage that instead of the init container.

### What is the use case or pain point?

In order to support fine tuning a model stored in an OCI registry in the [modelcar format](https://kserve.github.io/website/latest/modelserving/storage/oci/#using-modelcars) without first copying to S3, it'd be nice to have native support for a model artifact stored in OCI.

### Is there a workaround currently?

The user must copy it to S3 outside of the pipeline or manually pull down the image.

---

Love this idea? Give it a 👍.

HumairAK:
I think this is a great proposal @mprahl, thank you. There are various benefits to this approach beyond pull secrets. The linked Kserve page outlines some of these, but I'll paste them here for clarity: 

```
Reduced Startup Times: By avoiding repetitive downloads of large models, startup delays are significantly minimized.

Lower Disk Space Usage: The feature decreases the need for duplicated local storage, conserving disk space.

Enhanced Performance: Modelcars allows for advanced techniques like pre-fetching images and lazy-loading, improving efficiency.
```

As kserve is an addon component to KF, aligning here just makes sense as well. 

cc @franciscojavierarceo fyi

franciscojavierarceo:
Nice, this is great @mprahl !

PR #11606, opened 2025-02-08
feat(backend): Add support for importing models stored in the Modelcar format (sidecar)

mprahl:

This allows dsl.import to leverage Modelcar container images in an OCI repository. This works by having an init container prepull the image and then adding a sidecar container when the launcher container is running. The Modelcar container adds a symlink to its /models directory in an emptyDir volume that is accessible by the launcher container. Once the launcher is done running the user code, it stops the Modelcar containers.

This approach has the benefit of leveraging image pull secrets configured on the Kubernetes cluster rather than require separate credentials for importing the artifact. Additionally, no data is copied to the emptyDir volume, so the storage cost is just pulling the Modelcar container image on the Kubernetes worker node.

Note that once Kubernetes supports OCI images as volume mounts for several releases, consider replacing the init container with that approach.

This also adds a new environment variable of PIPELINE_RUN_AS_USER to set the runAsUser on all pods created by Argo Workflows.

Resolves:
https://github.com/kubeflow/pipelines/issues/11584

**Checklist:**

### DIFF
--- a/backend/src/v2/component/launcher_v2.go
+++ b/backend/src/v2/component/launcher_v2.go
@@ -127,12 +127,51 @@ func NewLauncherV2(ctx context.Context, executionID int64, executorInputJSON, co
 	}, nil
 }
 
+// stopWaitingArtifacts will create empty files to tell Modelcar sidecar containers to stop. Any errors encountered are
+// logged since this is meant as a deferred function at the end of the launcher's execution.
+func stopWaitingArtifacts(artifacts map[string]*pipelinespec.ArtifactList) {
+	for _, artifactList := range artifacts {
+		if len(artifactList.Artifacts) == 0 {
+			continue
+		}
+
+		// Following the convention of downloadArtifacts in the launcher to only look at the first in the list.
+		inputArtifact := artifactList.Artifacts[0]
+
+		// This should ideally verify that this is also a model input artifact, but this metadata doesn't seem to
+		// be set on inputArtifact.
+		if !strings.HasPrefix(inputArtifact.Uri, "oci://") {
+			continue
+		}
+
+		localPath, err := LocalPathForURI(inputArtifact.Uri)
+		if err != nil {
+			continue
+		}
+
+		glog.Infof("Stopping Modelcar container for artifact %s", inputArtifact.Uri)
+
+		launcherCompleteFile := strings.TrimSuffix(localPath, "/models") + "/launcher-complete"
+		_, err = os.Create(launcherCompleteFile)
+		if err != nil {
+			glog.Errorf(
+				"Failed to stop the artifact %s by creating %s: %v", inputArtifact.Uri, launcherCompleteFile, err,
+			)
+
+			continue
+		}
+	}
+}
+
 func (l *LauncherV2) Execute(ctx context.Context) (err error) {
 	defer func() {
 		if err != nil {
 			err = fmt.Errorf("failed to execute component: %w", err)
 		}
 	}()
+
+	defer stopWaitingArtifacts(l.executorInput.GetInputs().GetArtifacts())
+
 	// publish execution regardless the task succeeds or not
 	var execution *metadata.Execution
 	var executorOutput *pipelinespec.ExecutorOutput
@@ -401,6 +440,7 @@ func execute(
 	if err := downloadArtifacts(ctx, executorInput, bucket, bucketConfig, namespace, k8sClient); err != nil {
 		return nil, err
 	}
+
 	if err := prepareOutputFolders(executorInput); err != nil {
 		return nil, err
 	}
@@ -441,7 +481,7 @@ func uploadOutputArtifacts(ctx context.Context, executorInput *pipelinespec.Exec
 		}
 
 		// Upload artifacts from local path to remote storages.
-		localDir, err := localPathForURI(outputArtifact.Uri)
+		localDir, err := LocalPathForURI(outputArtifact.Uri)
 		if err != nil {
 			glog.Warningf("Output Artifact %q does not have a recognized storage URI %q. Skipping uploading to remote storage.", name, outputArtifact.Uri)
 		} else {
@@ -477,6 +517,31 @@ func uploadOutputArtifacts(ctx context.Context, executorInput *pipelinespec.Exec
 	return outputArtifacts, nil
 }
 
+// waitForModelcar assumes the Modelcar has already been validated by the init container on the launcher
+// pod. This waits for the Modelcar as a sidecar container to be ready.
+func waitForModelcar(artifactURI string, localPath string) error {
+	glog.Infof("Waiting for the Modelcar %s to be available", artifactURI)
+
+	for {
+		_, err := os.Stat(localPath)
+		if err == nil {
+			glog.Infof("The Modelcar is now available at %s", localPath)
+
+			return nil
+		}
+
+		if !os.IsNotExist(err) {
+			return fmt.Errorf(
+				"failed to see if the artifact %s was ready at %s; ensure the main container and Modelcar "+
+					"container have the same UID (can be set with the PIPELINE_RUN_AS_USER environment variable on "+
+					"the API server): %v",
+				artifactURI, localPath, err)
+		}
+
+		time.Sleep(500 * time.Millisecond)
+	}
+}
+
 func downloadArtifacts(ctx context.Context, executorInput *pipelinespec.ExecutorInput, defaultBucket *blob.Bucket, defaultBucketConfig *objectstore.Config, namespace string, k8sClient kubernetes.Interface) error {
 	// Read input artifact metadata.
 	nonDefaultBuckets, err := fetchNonDefaultBuckets(ctx, executorInput.GetInputs().GetArtifacts(), defaultBucketConfig, namespace, k8sClient)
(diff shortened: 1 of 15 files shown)


mprahl:
The last step that I'm aware of that needs to happen is to have the launcher SIGHUP the `sleep infinity` rather than the Argo Exec process.

samples/v2/modelcar_import/modelcar_import.py
```diff
@@ -0,0 +1,75 @@
...
+    return ", ".join(sorted(filesInPath))
+
+
+@dsl.pipeline(name="pipeline-with-modelcar-model")
+def pipeline_modelcar_import(
+    model_uri: str = "oci://registry.domain.local/modelcar:test",
+):
+    model_source_oci_task = dsl.importer(
+        artifact_uri=model_uri, artifact_class=dsl.Model
+    ).set_caching_options(False)
```
mprahl, line 67: @mprahl to test with caching enabled

backend/src/v2/driver/driver.go
```diff
@@ -549,9 +550,141 @@ func initPodSpecPatch(
...
+					"sh",
+					"-c",
+					// $$$$ gets escaped by YAML to $$, which is the current PID
+					// Mostly taken from KServe, but exit 0 once sleep gets a SIGHUP signal so the pod
+					// is seen as exiting successfully:
+					// https://github.com/kserve/kserve/blob/v0.14.1/pkg/webhook/admission/pod/storage_initializer_injector.go#L732
+					fmt.Sprintf(
+						"ln -s /proc/$$$$/root/models \"%s\" && "+
+							"echo \"Running...\" && sleep infinity || "+
+							"if [ $? -eq 129 ]; then echo \"Shutting down...\"; else echo \"Failed with: $?\" && exit $?; fi",
```
mprahl, line 655: Remove the echo here

backend/src/v2/driver/driver.go
```diff
@@ -549,9 +550,141 @@ func initPodSpecPatch(
...
+
+	if podSpec.Containers[0].SecurityContext.Capabilities == nil {
+		podSpec.Containers[0].SecurityContext.Capabilities = &k8score.Capabilities{}
+	}
+
+	// The SYS_PTRACE capability is required for the main container to access the Modelcar container's
+	// filesystem. The SIGHUP capability is required for the main container to stop the Modelcar container
+	// when launcher is exiting.
+	podSpec.Containers[0].SecurityContext.Capabilities.Add = append(
+		podSpec.Containers[0].SecurityContext.Capabilities.Add, "SYS_PTRACE", "SIGHUP",
```
mprahl, line 684: Make sure this isn't a problem in OpenShift.

dandawg (approved):
/lgtm
This is really cool. A couple of side questions came to mind.
1) What happens if the model is not in the /models path? I know this is the default in kserve images, but can we always count on this? I know BentoML and Seldon also create oci images for models. Is the path something we can parameterize?
2) We may need examples/docs, since it may not be immediately clear in which use cases this is valuable--answering the question of when I should use an api like huggingface, when I should use s3, and when I should use oci.

HumairAK:
@mprahl can you rebase?

mprahl:
> /lgtm This is really cool. A couple of side questions came to mind.
> 
> 1. What happens if the model is not in the /models path? I know this is the default in kserve images, but can we always count on this? I know BentoML and Seldon also create oci images for models. Is the path something we can parameterize?
> 2. We may need examples/docs, since it may not be immediately clear in which use cases this is valuable--answering the question of when I should use an api like huggingface, when I should use s3, and when I should use oci.

Good question on 1. For Modelcar, it's always in `/models`. It's part of the [requirements](https://kserve.github.io/website/latest/modelserving/storage/oci/#prepare-an-oci-image-with-model-data). I couldn't find any clear documentation on the format used from BentoML and Seldon, so if there's a need later on to support those, we'll have to figure out what the standard is. Since KServe is part of Kubeflow, we should try to support their packaging options.

On 2, I definitely need to add documentation to the website with examples. If this PR gets merged, I'll try to get documentation in before the next KFP release.

Thanks for the review!

backend/src/v2/component/launcher_v2.go
```diff
@@ -491,17 +556,31 @@ func downloadArtifacts(ctx context.Context, executorInput *pipelinespec.Executor
...
-		localPath, err := localPathForURI(inputArtifact.Uri)
+
+		localPath, err := LocalPathForURI(inputArtifact.Uri)
 		if err != nil {
 			glog.Warningf("Input Artifact %q does not have a recognized storage URI %q. Skipping downloading to local path.", name, inputArtifact.Uri)
+
 			continue
 		}
+
+		// OCI artifacts are handled specially
```
HumairAK, line 574:
```suggestion
		// OCI artifacts are accessed via shared storage with ModelCar
```

HumairAK (changes requested):

HumairAK:
a note for documentation, we'll probably want to somehow express that only input models are supported today, as soon as users try to output this oci model it will fail

sdk/python/kfp/dsl/types/artifact_types.py
```diff
@@ -108,6 +114,14 @@ def convert_local_path_to_remote_path(path: str) -> str:
         return MINIO_REMOTE_PREFIX + path[len(_MINIO_LOCAL_MOUNT_PREFIX):]
     elif path.startswith(_S3_LOCAL_MOUNT_PREFIX):
         return S3_REMOTE_PREFIX + path[len(_S3_LOCAL_MOUNT_PREFIX):]
+    elif path.startswith(_OCI_LOCAL_MOUNT_PREFIX):
+        remotePath = OCI_REMOTE_PREFIX + path[len(_OCI_LOCAL_MOUNT_PREFIX
+                                                 ):].replace('\\/', '/')
```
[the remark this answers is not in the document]
mprahl: This was how `yapf` formatted it but I'll try something nicer.

sdk/python/kfp/dsl/types/artifact_types.py
```diff
@@ -128,6 +142,13 @@ def framework(self) -> str:
     def _get_framework(self) -> str:
         return self.metadata.get('framework', '')
 
+    @property
+    def path(self) -> str:
+        if self.uri.startswith("oci://"):
+            return self._get_path() + "/models"
```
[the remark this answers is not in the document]
mprahl: Yeah, it's a hardcoded path that KServe expects for Modelcar containers. I'll add a note.

backend/src/v2/driver/driver.go
```diff
@@ -549,9 +550,117 @@ func initPodSpecPatch(
...
+		// If there is at least one Modelcar image, then shareProcessNamespace must be enabled.
+		trueVal := true
+		podSpec.ShareProcessNamespace = &trueVal
+
+		image := strings.TrimPrefix(inputArtifact.Uri, "oci://")
+
+		podSpec.InitContainers = append(
+			podSpec.InitContainers,
+			k8score.Container{
+				Name:  "oci-prepull-" + name,
```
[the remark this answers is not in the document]
mprahl: Good catch. I'll sort by input artifact name and then use the index here instead. It'll be deterministic and always valid.

backend/src/v2/component/launcher_v2.go
```diff
@@ -127,12 +127,51 @@ func NewLauncherV2(ctx context.Context, executionID int64, executorInputJSON, co
...
+		if !strings.HasPrefix(inputArtifact.Uri, "oci://") {
+			continue
+		}
+
+		localPath, err := LocalPathForURI(inputArtifact.Uri)
+		if err != nil {
+			continue
+		}
+
+		glog.Infof("Stopping artifact %s", inputArtifact.Uri)
```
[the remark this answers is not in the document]
mprahl: I tried to make this code generic but since we only support Modelcars for now, I'll be more specific.

sdk/python/kfp/dsl/types/artifact_types.py
```diff
@@ -128,6 +141,15 @@ def framework(self) -> str:
     def _get_framework(self) -> str:
         return self.metadata.get('framework', '')
 
+    @property
+    def path(self) -> str:
+        if self.uri.startswith("oci://"):
+            # Modelcar container images always have the model files stored in /models
```
dandawg, line 147: Maybe "are expected to" instead of "always". From what I remember of KServe, it's possible to build a runtime image and put your model in another path--but we would expect it to be in /models.

samples/v2/modelcar_import/modelcar_import.py
```diff
@@ -0,0 +1,75 @@
...
+from kfp import dsl
+from kfp.dsl import component
+
+# In tests, we install a KFP package from the PR under test. Users should not
+# normally need to specify `kfp_package_path` in their component definitions.
+_KFP_PACKAGE_PATH = os.getenv("KFP_PACKAGE_PATH")
+
+
+@dsl.component(kfp_package_path=_KFP_PACKAGE_PATH)
+def get_model_files_list(model: dsl.Input[dsl.Model]) -> str:
```
HumairAK, line 29: optional: this will help with testing for the case [here](https://github.com/kubeflow/pipelines/pull/11606#discussion_r1960792830)

```suggestion
def get_model_files_list(input_model: dsl.Input[dsl.Model]) -> str:
```

PR #11474, opened 2024-12-19
chore: Update image for metacontroller 

biswassri:
**Description of your changes:**

Based on [this](https://github.com/metacontroller/metacontroller/releases/tag/v4.11.21) release the docker image of the metacontroller is being deprecated and need to be updated.

**Checklist:**
- [x] You have [signed off your commits](https://www.kubeflow.org/docs/about/contributing/#sign-off-your-commits)
- [x] The title for your pull request (PR) should follow our title convention. [Learn more about the pull request title convention used in this repository](https://github.com/kubeflow/pipelines/blob/master/CONTRIBUTING.md#pull-request-title-convention).

PR #11608, opened 2025-02-10
fix(backend): the metacontroller is broken since #11474

juliusvonkohout:

This fixes the metacontroller after https://github.com/kubeflow/pipelines/pull/11474 for multi-user KFP as done in https://github.com/kubeflow/manifests/pull/2953/

@hbelmiro 

@tarekabouzeid @biswassri I did not get feedback for a week in https://github.com/kubeflow/manifests/issues/2970#issuecomment-2647615231, so I created https://github.com/kubeflow/pipelines/pull/11608 to at least fix the broken upstream KFP. Afterwards you can still upgrade metacontroller from 2.6.1 to 4.x


hbelmiro (commented):
/lgtm

Issue #6161, opened 2021-07-27
v2 control flow - fan in


Bobgy:
[aggregate outputs as iterator DAG output](https://github.com/kubeflow/pipelines/blob/fef8c03e401a15a9f92c1839fe0f9a5c22f709e1/api/v2alpha1/pipeline_spec.proto#L466-L470)

papagala:
Sorry again for the spam but its a bit confusing since this is still not fixed nor supported as far as I could see by running the `after` (https://github.com/kubeflow/pipelines/issues/10050) method or the `dsl.Collected` (https://github.com/kubeflow/pipelines/issues/8830). 

It seems there is no workaround for fan in at the moment @chensun (sorry for the second mention in a day, just trying to understand this issue)

JohanRuiz9801:
hello @papagala Im searching for a way to gather the results of the tasks inside a ParallelFor loop and my investigation drive me here, but I saw that dsl.Collected still seems to be unsupported. What other alternative do I have? Thank you

PR #11627, opened 2025-02-13
fix(backend): parallelFor resolve upstream inputs. Fixes #11520

### LINKED ISSUE #10050: [sdk] Unable to aggregate results over ParallelFor in Kubeflow V2 using V1 workarounds such as `.after()`
### Environment

*  KFP version: V2 backend. 2.0.1.
*  KFP SDK version: 2.3.0 (failing), 1.8.21 (working)
*  All dependencies version:
```
kfp-server-api             2.0.1 
``` 

### Steps to reproduce

The new Kubeflow V2 backend does not yet support the use of `dsl.Collected()`, which is a functionality I was really looking forward to. I have a process where I run several components within a `dsl.ParallelFor` and then want to aggregate the results of all of these components after the `dsl.ParallelFor` finishes. 

In V1 using SDK version `kfp==1.8.21`, I was able to work around the lack of any released fan-in mechanism by using the `.after()` method. A dummy snippet of code would be as follows, where `comp_in_parfor` executes some process and saves the results to S3, and then `collector` collects the results after the `comp_in_parfor` has finished by reading in the results from S3.
```python
with dsl.ParallelFor(parfor_list) as args:
    # saves out some artifacts to S3 to make DIY fan-in logic
    comp_in_parfor = comp_template(a=args.a, b=args.b)
    
# reads in those same artifacts from S3
 collector = collector_template(...).after(comp_in_parfor)
``` 

However, if I try to do the same thing using `kfp==2.3.0` for the SDK, I get the following error, which means I cannot use the `.after()` method in V2. However, because the new `dsl.Collected` method also does not work in the V2 open-source backend (only Vertex AI from what I can tell), there is no way to fan-in from a ParallelFor, either DIY or properly.
```
kfp.compiler.compiler_utils.InvalidTopologyException: Illegal task dependency across DSL context managers. A downstream task
cannot depend on an upstream task within a dsl.ParallelFor context unless the downstream is within that context too or the
outputs are begin fanned-in to a list using dsl.Collected. Found task square-root which depends on upstream task square-and
sum within an uncommon dsl.ParallelFor context.
``` 
This error can easily be reproduced with this example:
```python
import os
import kfp
from kfp import dsl
from typing import List


@dsl.component
def add(x: float, y: float) -> float:
    return x + y

@dsl.component
def square_root(x: float) -> float:
    return x ** .5

@dsl.pipeline
def dummy(parfor_list: List[float] = [1.2, 2.4, 3.6], b: float = 1.2) -> float:
    with dsl.ParallelFor(parfor_list) as args:
        sq_and_sum_task = add(x=args, y=args)
    square_root_comp = square_root(x=b).after(sq_and_sum_task)
    return square_root_comp.output
```

### Expected result

My expected result can be one of two things:

1. `dsl.Collected()` is properly implemented in V2 KFP backend, rendering the need for `.after()` moot.
2. `.after()` works, so that users can properly lay out the sequential structure of their DAG and workaround the lack of fan-in logic present in kubeflow.

Am I missing a workaround that allows users to fan-in from a ParallelFor in the V2 SDK? If not, is there any way that `.after()` can be restored until `dsl.Collected()` is implemented properly in the backend? Until then, everyone blocked by this [issue](https://github.com/kubeflow/pipelines/issues/3412) will be totally unable to use the V2 Kubeflow with their own DIY fan-in logic. It's a major bummer because V2 has some great functionalities. Namely I'm quite excited to have sub-dags to logically break up my repeated ParallelFor units and reduce UI strain due to overly large DAG structures.

### Materials and Reference

- https://github.com/kubeflow/pipelines/issues/3412
- https://www.kubeflow.org/docs/components/pipelines/v2/pipelines/control-flow/#parallel-looping-dslparallelfor

### LINKED ISSUE #11520: [backend] dsl.ParallelFor loop: cannot resolve the upstream artifact output of a previous pod
### Environment

*  KFP version: master (which I assume is a predecessor to 2.4.0)

*  KFP SDK version: 2.11.0

### Steps to reproduce

Here is a simple pipeline that produces this error:
```python
import kfp
import kfp.kubernetes
from kfp import dsl
from kfp.dsl import Artifact, Input, Output


@dsl.component(base_image="python:3.10")
def split_model_ids(model_ids: str) -> list:
    return model_ids.split(",")

@dsl.component(base_image="python:3.10")
def create_file(file: Output[Artifact], content: str):
    with open(file.path, "w") as f:
        f.write(content)

@dsl.component(base_image="python:3.10")
def read_file(file: Input[Artifact]) -> str:
    with open(file.path, "r") as f:
        print(f.read())

@dsl.pipeline(name="Pipeline", description="Pipeline")
def export_model(
    model_ids: str = "",
):
    model_ids_split_op = split_model_ids(model_ids=model_ids)
    with dsl.ParallelFor(model_ids_split_op.output) as model_id:
        create_file_op = create_file(content=model_id)
        read_file_op = read_file(file=create_file_op.outputs["file"])
        read_file_op.after(create_file_op)

if __name__ == "__main__":
    kfp.compiler.Compiler().compile(export_model, "simple_pipeline.yaml")
```

### Expected result

There were some changes that were made to parallel for loop when running from the master branch. I get this error because of the new kfp driver. This happens because I have a dsl.ParallelFor loop, where it cannot resolve the upstream artifact output of a previous pod. 

### Materials and Reference

Gives this error:
```
│ main I0107 16:08:18.259327      22 driver.go:984] parent DAG input parameters: map[pipelinechannel--split-model-ids-Output-loop-item:string_value:"zs1"], artifa │
│ cts: map[]                                                                                                                                                       │
│ main panic: runtime error: invalid memory address or nil pointer dereference                                                                                     │
│ main [signal SIGSEGV: segmentation violation code=0x1 addr=0x40 pc=0x1f1cd39]                                                                                    │
│ main                                                                                                                                                             │
│ main goroutine 1 [running]:                                                                                                                                      │
│ main github.com/kubeflow/pipelines/backend/src/v2/driver.resolveUpstreamArtifacts({{0x2c074f0, 0x4148aa0}, 0x0, 0xc000809e40, 0xc000c71200, 0xc00012df00, 0xc000 │
│ a718f0, 0xc000befbc0, {0xc00088dde0, 0x9}, ...})                                                                                                                 │
│ main     /go/src/github.com/kubeflow/pipelines/backend/src/v2/driver/driver.go:1450 +0x4b9                                                                       │
│ main github.com/kubeflow/pipelines/backend/src/v2/driver.resolveInputs({0x2c074f0, 0x4148aa0}, 0xc000c71200, 0x0, 0xc00012df00, 0xc000301b90, 0xc000809840, 0xc0 │
│ 00a718f0, 0xc000ba77f0)                                                                                                                                          │
│ main     /go/src/github.com/kubeflow/pipelines/backend/src/v2/driver/driver.go:1221 +0x1b58                                                                      │
│ main github.com/kubeflow/pipelines/backend/src/v2/driver.Container({0x2c074f0, 0x4148aa0}, {{0x7ffe76d8c3f0, 0x16}, {0x7ffe76d8c410, 0x24}, 0xc0009bf310, 0xffff │
│ ffffffffffff, 0x0, {0xc0009a22a0, ...}, ...}, ...)                                                                                                               │
│ main     /go/src/github.com/kubeflow/pipelines/backend/src/v2/driver/driver.go:263 +0x3c9                                                                        │
│ main main.drive()                                                                                                                                                │
│ main     /go/src/github.com/kubeflow/pipelines/backend/src/v2/cmd/driver/main.go:174 +0xb1c                                                                      │
│ main main.main()                                                                                                                                                 │
│ main     /go/src/github.com/kubeflow/pipelines/backend/src/v2/cmd/driver/main.go:77 +0x65                                                                        │
│ main time="2025-01-07T16:08:18.862Z" level=info msg="sub-process exited" argo=true error="<nil>"                                                                 │
│ main time="2025-01-07T16:08:18.862Z" level=error msg="cannot save parameter /tmp/outputs/pod-spec-patch" argo=true error="open /tmp/outputs/pod-spec-patch: no s │
│ uch file or directory"                                                                                                                                           │
│ main time="2025-01-07T16:08:18.862Z" level=error msg="cannot save parameter /tmp/outputs/cached-decision" argo=true error="open /tmp/outputs/cached-decision: no │
│  such file or directory"                                                                                                                                         │
│ main time="2025-01-07T16:08:18.862Z" level=error msg="cannot save parameter /tmp/outputs/condition" argo=true error="open /tmp/outputs/condition: no such file o │
│ r directory"                                                                                                                                                     │
│ main Error: exit status 2   
```

Thank you so much!
---

zazulam:

* Updated `resolveUpstreamArtifacts` to use `getDAGTasks` to allow for it parse through parallelFor DAG contexts to retrieve the appropriate producerTask. This fixes #11520, it seems that the call with the filter to `GetExecutionsInDAG` was not reverted back to the same as it is in `resolveUpstreamParameters` in #11196.

* Updated the argocompiler for the `iteratorTask` section. The additional "-loop" added to the task associated with the parallelFor DAG breaks the dependency validation that the argoworkflows API calls in the submission of the pipeline. This partially resolves #10050 for the `.after()` usage of a parallelTask. The implementation for `dsl.Collected` in the backend will be coming in a follow-up PR.

cc: @droctothorpe 

**Checklist:**

### DIFF
--- a/backend/src/v2/compiler/argocompiler/dag.go
+++ b/backend/src/v2/compiler/argocompiler/dag.go
@@ -369,7 +369,7 @@ func (c *workflowCompiler) iteratorTask(name string, task *pipelinespec.Pipeline
 
 	tasks = []wfapi.DAGTask{
 		{
-			Name:     name + "-loop",
+			Name:     name,
 			Template: loopTmplName,
 			Depends:  depends(task.GetDependentTasks()),
 			Arguments: wfapi.Arguments{
--- a/backend/src/v2/compiler/argocompiler/testdata/multiple_parallel_loops.yaml
+++ b/backend/src/v2/compiler/argocompiler/testdata/multiple_parallel_loops.yaml
@@ -9,32 +9,13 @@ spec:
     - name: components-e7a1060777c9ef84e36a4d54f25d3102abbcbd62d4c8bbb5883c3a2cbcfb5c6d
       value: '{"executorLabel":"exec-print-op","inputDefinitions":{"parameters":{"s":{"parameterType":"STRING"}}}}'
     - name: implementations-e7a1060777c9ef84e36a4d54f25d3102abbcbd62d4c8bbb5883c3a2cbcfb5c6d
-      value: '{"args":["--executor_input","{{$}}","--function_to_execute","print_op"],"command":["sh","-c","\nif
-              ! [ -x \"$(command -v pip)\" ]; then\n    python3 -m ensurepip || python3 -m
-              ensurepip --user || apt-get install python3-pip\nfi\n\nPIP_DISABLE_PIP_VERSION_CHECK=1
-              python3 -m pip install --quiet --no-warn-script-location ''kfp==2.7.0'' ''--no-deps''
-              ''typing-extensions\u003e=3.7.4,\u003c5; python_version\u003c\"3.9\"'' \u0026\u0026
-              \"$0\" \"$@\"\n","sh","-ec","program_path=$(mktemp -d)\n\nprintf \"%s\" \"$0\"
-              \u003e \"$program_path/ephemeral_component.py\"\n_KFP_RUNTIME=true python3 -m
-              kfp.dsl.executor_main                         --component_module_path                         \"$program_path/ephemeral_component.py\"                         \"$@\"\n","\nimport
-              kfp\nfrom kfp import dsl\nfrom kfp.dsl import *\nfrom typing import *\n\ndef
-              print_op(s: str):\n    print(s)\n\n"],"image":"python:3.7"}'
+      value: '{"args":["--executor_input","{{$}}","--function_to_execute","print_op"],"command":["sh","-c","\nif ! [ -x \"$(command -v pip)\" ]; then\n    python3 -m ensurepip || python3 -m ensurepip --user || apt-get install python3-pip\nfi\n\nPIP_DISABLE_PIP_VERSION_CHECK=1 python3 -m pip install --quiet --no-warn-script-location ''kfp==2.7.0'' ''--no-deps'' ''typing-extensions\u003e=3.7.4,\u003c5; python_version\u003c\"3.9\"'' \u0026\u0026 \"$0\" \"$@\"\n","sh","-ec","program_path=$(mktemp -d)\n\nprintf \"%s\" \"$0\" \u003e \"$program_path/ephemeral_component.py\"\n_KFP_RUNTIME=true python3 -m kfp.dsl.executor_main                         --component_module_path                         \"$program_path/ephemeral_component.py\"                         \"$@\"\n","\nimport kfp\nfrom kfp import dsl\nfrom kfp.dsl import *\nfrom typing import *\n\ndef print_op(s: str):\n    print(s)\n\n"],"image":"python:3.7"}'
     - name: components-comp-for-loop-2
       value: '{"dag":{"tasks":{"print-op":{"cachingOptions":{"enableCache":true},"componentRef":{"name":"comp-print-op"},"inputs":{"parameters":{"s":{"componentInputParameter":"pipelinechannel--loop-item-param-1","parameterExpressionSelector":"parseJson(string_value)[\"A_a\"]"}}},"taskInfo":{"name":"print-op"}},"print-op-2":{"cachingOptions":{"enableCache":true},"componentRef":{"name":"comp-print-op-2"},"inputs":{"parameters":{"s":{"componentInputParameter":"pipelinechannel--loop-item-param-1","parameterExpressionSelector":"parseJson(string_value)[\"B_b\"]"}}},"taskInfo":{"name":"print-op-2"}}}},"inputDefinitions":{"parameters":{"pipelinechannel--loop-item-param-1":{"parameterType":"STRUCT"}}}}'
     - name: components-comp-for-loop-4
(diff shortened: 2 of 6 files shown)


zazulam:
Seems like this PR is also being affected by the flaky tests. I ran the failing pipelines locally and all passed.

image

droctothorpe (approved):

droctothorpe:
cc @HumairAK this resolves https://github.com/kubeflow/pipelines/issues/11520.

zazulam:
Adding some details here related to `.after()` scenario, running an `argo lint` on the workflow file generated from the IR can show the actual error [from the argoworkflow api](https://github.com/argoproj/argo-workflows/blob/e79d1c5c527491b05d7fc421e8b903208b8a5f1c/util/sorting/topological_sorting.go#L12-L30) that was being raised in [this comment](https://github.com/kubeflow/pipelines/issues/10050#issuecomment-2301933497). 

image

With removing that `-loop`
image

Only diff in the workflow
image

The example used was from the comment and also added as the test case in the `samples/v2` suite:
```python
from kfp import dsl


@dsl.component
def print_op(message: str) -> str:
    print(message)
    return message

@dsl.component
def reduce_op(message: str) -> str:
    print(message)
    return message[0]


@dsl.pipeline()
def my_pipeline():
    with dsl.ParallelFor([1, 2, 3]):
        one = print_op(message='foo')
    two = print_op(message='bar').after(one)
```

HumairAK:
/lgtm
/approve 

Thanks for the quick turn around on this folks!

zazulam:
@HumairAK I actually started to work on implementing the backend support for `dsl.Collected` 😅 
I just need to add some tests for artifacts and parameters, then this PR should be good to fully resolve #10050.

zazulam:
> @HumairAK I actually started to work on implementing the backend support for dsl.Collected 😅
I just need to add some tests for artifacts and parameters, then this PR should be good to fully resolve https://github.com/kubeflow/pipelines/issues/10050.

I'm going to save the `dsl.Collected` for a separate PR as I learned from #6161 that there exists certain Iterator classes in the pipelinespec (I'm not aware atm if they are being used or set anywhere in the backend). My current solution is not leveraging those classes to fan in the outputs, but I would rather not hold back this PR.

HumairAK:
@zazulam I think separate pr makes sense, if we can keep this one light weight I might be able to cherry pick this for the 2.4.1 patch release I'll make next week so we can address the regression.

PR #11628, opened 2025-02-14
feat(frontend): Add an 'Always use latest version option' for recurring runs. Fixes #11581

### LINKED ISSUE #11581: [feature] Frontend: Allow creating recurring runs that always use the latest pipeline version
### Feature Area

/area frontend

In #11560, the backend made the pipeline version ID optional when creating a recurring run. When the pipeline version ID is not set, the latest pipeline version is always selected for each run.

The user must use the API to create such recurring runs.

You must use the API as a workaround.

---

Love this idea? Give it a 👍.

muzzlol:

This change introduces a checkbox in the New Run V2 page that allows users to create recurring runs that always use the latest version of a pipeline, without needing to specify a pipeline version ID.

The changes include:

- Added a `useLatestVersion` state variable to track the checkbox's state.
- Persisted the `useLatestVersion` preference in `localStorage`.
- Modified the `pipeline_version_reference` in the API payload to only send the `pipeline_id` when `useLatestVersion` is enabled.
- Disabled the pipeline version selector when `useLatestVersion` is enabled.
- Implemented a fallback mechanism to update the Run Name when the checkbox is toggled before a pipeline is selected.

**Checklist:**

### DIFF
--- a/frontend/src/pages/NewRunV2.tsx
+++ b/frontend/src/pages/NewRunV2.tsx
@@ -22,6 +22,7 @@ import {
   InputAdornment,
   FormControlLabel,
   Radio,
+  Checkbox,
 } from '@material-ui/core';
 import React, { useEffect, useState } from 'react';
 import * as JsYaml from 'js-yaml';
@@ -204,6 +205,26 @@ function NewRunV2(props: NewRunV2Props) {
       ? !cloneOrigin.recurringRun.no_catchup
       : true;
   const [needCatchup, setNeedCatchup] = useState(initialCatchup);
+  const [useLatestVersion, setUseLatestVersion] = useState(() => {
+    const saved = localStorage.getItem('useLatestPipelineVersion');
+    return saved === 'true';
+  });
+
+  useEffect(() => {
+    localStorage.setItem('useLatestPipelineVersion', useLatestVersion.toString());
+  }, [useLatestVersion]);
+
+  useEffect(() => {
+    if (useLatestVersion) {
+      // Clear pipeline version when using latest
+      setPipelineVersionName('');
+      setRunName(
+        existingPipeline
+          ? 'Run of ' + existingPipeline.display_name + ' ' + new Date().toISOString()
+          : '',
+      );
+    }
+  }, [useLatestVersion, existingPipeline]);
 
   const clonedRuntimeConfig = cloneOrigin.isRecurring
     ? cloneOrigin.recurringRun?.runtime_config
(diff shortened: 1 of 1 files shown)


frontend/src/pages/NewRunV2.tsx
```diff
@@ -204,6 +205,26 @@ function NewRunV2(props: NewRunV2Props) {
       ? !cloneOrigin.recurringRun.no_catchup
       : true;
   const [needCatchup, setNeedCatchup] = useState(initialCatchup);
+  const [useLatestVersion, setUseLatestVersion] = useState(() => {
+    const saved = localStorage.getItem('useLatestPipelineVersion');
```
mprahl, line 209: My preference is that this preference isn't saved because the user decision is for a particular recurring run they are creating for a particular pipeline, but this preference is a global setting.

It's also not clear that checking this box is causing it to be the default preference in the future.

frontend/src/pages/NewRunV2.tsx
```diff
@@ -505,6 +525,7 @@ function NewRunV2(props: NewRunV2Props) {
               {...props}
               pipeline={existingPipeline}
               pipelineVersionName={pipelineVersionName}
+              disabled={useLatestVersion}
```
mprahl, line 528: When it's disabled, the form input still shows "Pipeline Version *" which makes it seem like it's still required but I can't select anything. Could you change the text to something like "Using latest pipeline version" or remove that text?

mprahl:
Thanks for the contribution @muzzlol ! This is looking good.

Could you also please add tests?

frontend/src/pages/NewRunV2.tsx
muzzlol: @mprahl 
Apologies for the delayed response, for some  reason I didn't receive notifications about new activity in this thread.
Regarding your assessment, I agree that storing this as a global setting isn't ideal, but the implementation  I proposed is largely redundant since the system already fetches and preselects the latest pipeline version by default when user loads the page.
I believe it's best to close this PR for now and potentially revisit with a comprehensive approach if users specifically request this functionality.
for  now the current implementation seems to be the best user experience imo

mprahl: @muzzlol the benefit of the feature of always using the latest pipeline version is that you don't have to change recurring runs if you upload a new version due to something like CVE fixes in the container images used in the pipeline, so I think it's a good feature to expose in the UI.

kaikaila:
Hi @muzzlol @mprahl ,

I fully respect @muzzlol's initial implementation and reasoning. However, since @mprahl has highlighted concrete user benefits for this feature, I'd like to propose a collaborative path forward:  

- Preserve your commits (c09bc4e/d74d043) as the foundation  
- Address feedback (remove `localStorage`, improve UI labels, add tests)  
- Submit as a new PR with co-authorship attribution to @muzzlol 

If you agree, I'll:  
1. Fork your branch and make incremental changes  
2. Add you as a co-author in the final commit  
3. Ensure your original contribution is fully recognized  

This way, we can collectively deliver the feature while respecting all perspectives.  
/assign

muzzlol:
yea @kaikaila sounds good . let me know if you need any help setting up the development enviornment

frontend/src/pages/NewRunV2.tsx
muzzlol: `       
        disabled={useLatestVersion}
        label={useLatestVersion ? 'Using latest pipeline version' : 'Pipeline Version'}
`
@kaikaila you could add somethign like this to adress the this

kaikaila:
Hey everyone, just a quick update! I’ve implemented the fix and am currently resolving some final Git branch/sign-off issues. Expecting to submit the PR later today or early tomorrow. Let me know if there’s anything specific you’d like me to double-check before submitting. Thanks

frontend/src/pages/NewRunV2.tsx
kaikaila: > ` disabled={useLatestVersion} label={useLatestVersion ? 'Using latest pipeline version' : 'Pipeline Version'}` @kaikaila you could add somethign like this to adress the this

Got it. Thanks

Issue #11572, opened 2025-01-30
[bug] Subdag Statuses update incorrectly 


HumairAK:
When using any control flows that require subdags like loops, nested pipelines, etc. the statuses update as soon as the first task within the dag completes. This is incorrect behavior as it should only show loading icon while sub tasks within the dag are still running. 

### Environment

*  How do you deploy Kubeflow Pipelines (KFP)? standalone and multi-user

*  KFP version: 2.4

To find the version number, See version number shows on bottom of KFP UI left sidenav. -->
*  KFP SDK version: 2.11

### Steps to reproduce

* Use a simple dag pipeline like the following: 

```python
@dsl.component()
def fail():
    import sys
    sys.exit(1)

@dsl.component()
def hello_world():
    print("hellow_world")

@dsl.component()
def post_msg():
    print(f"this is a message")

@dsl.pipeline
def pipeline():
    post_msg_task = post_msg().set_caching_options(enable_caching=False)
    with dsl.ExitHandler(exit_task=post_msg_task):
        hello_task = hello_world().set_caching_options(enable_caching=False)
        fail_task = fail().set_caching_options(enable_caching=False)
        fail_task.after(hello_task)


compiler.Compiler().compile(
    pipeline_func=pipeline,
    package_path=__file__.replace('.py', '-v2.yaml'))

```

Or any other control flow that utilizes sub dags, i.e. loops, nested pipelines, exit handlers, etc. 

### Expected result

The subdag status should stay as loading until all tasks are completed. 

### Materials and reference

https://github.com/user-attachments/assets/4da858b9-79dc-4a72-a285-58880d8b253a

### Labels

/area backend

---

Impacted by this bug? Give it a 👍.

izuku-sds:
Hello @HumairAK , can you please mention exact steps you used to deploy kubeflow pipeline. I was trying to deploy it locally `standalone and multi-user` but got problem at last when trying to access ui page or dashboard.
I followed [instructions](https://github.com/kubeflow/manifests?tab=readme-ov-file#install-individual-components) .

PR #11651, opened 2025-02-18
fix(backend): fixes DAG status update to reflect completion of all tasks

VaniHaripriya:
This PR resolves [#11572](https://github.com/kubeflow/pipelines/issues/11572)
Subdags are being updated as soon as the first task within the subdag completes, which is incorrect behavior. The subdag status should only be updated once all tasks have completed.

This fix ensures that the subdag status is updated only after all tasks have been executed.

**Checklist:**

### DIFF
--- a/backend/src/v2/driver/driver.go
+++ b/backend/src/v2/driver/driver.go
@@ -849,6 +849,10 @@ func DAG(ctx context.Context, opts Options, mlmd *metadata.Client) (execution *E
 	ecfg.OutputArtifacts = opts.Component.GetDag().GetOutputs().GetArtifacts()
 	glog.V(4).Info("outputArtifacts: ", ecfg.OutputArtifacts)
 
+	totalDagTasks := len(opts.Component.GetDag().GetTasks())
+	ecfg.TotalDagTasks = &totalDagTasks
+	glog.V(4).Info("totalDagTasks: ", *ecfg.TotalDagTasks)
+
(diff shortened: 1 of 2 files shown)


VaniHaripriya:
Attaching a pipeline run after the fix.

https://github.com/user-attachments/assets/92333fdf-8635-4159-9eb0-8ba034920b55

VaniHaripriya:
> @VaniHaripriya can you create a related issue and elaborate more on what you're seeing? Also if you could provide some screenshots of before & after in this PR it would be very helpful.

@zazulam Here is related issue-https://github.com/kubeflow/pipelines/issues/11572 and it has a video of DAG status before the fix.

VaniHaripriya:
> This makes sense to me, the PR that introduced this bug would check for the task Executions in a given subdag context and attempt to update the subdag's state. However, it was using the `mlmd.GetExecutionsInDAG`, which would only return executions that have already been created rather than the list of tasks defined 🤦‍♂️
> 
> Would it seem more appropriate to store the number of tasks a subdag has rather than the tasks themselves? That way we can avoid multiple calls to the `GetTotalTaskCount` function.

Makes sense to me, I updated to store the number of tasks instead of tasks.

HumairAK:
great work @VaniHaripriya! tested and works, approach lgtm! (thanks @zazulam for the review!) 



/approve

PR #11664, opened 2025-02-24
Chore: dockerhub introduces a stricter rate limit

juliusvonkohout:
See also https://github.com/kubeflow/manifests/issues/3010#issuecomment-2677977953 @hbelmiro @HumairAK and 3.9 is EOL in October so I upgraded it.


manifests/kustomize/base/installs/multi-user/pipelines-profile-controller/deployment.yaml
```diff
@@ -11,7 +11,7 @@ spec:
     spec:
       containers:
       - name: profile-controller
-        image: python:3.9
+        image: public.ecr.aws/docker/library/python:3.12
```
juliusvonkohout, line 14: This one here is the most important change for end users. I am running this for a long time on 3.12

hbelmiro (commented):
@juliusvonkohout @HumairAK, we currently support versions 3.9 to 3.12. If we're dropping support for 3.9, I think we should replace it with 3.10 instead of 3.12. If we use 3.12, we may end up using some feature that still doesn't exist in 3.10.

juliusvonkohout:
> @juliusvonkohout @HumairAK, we currently support versions 3.9 to 3.12. If we're dropping support for 3.9, I think we should replace it with 3.10 instead of 3.12. If we use 3.12, we may end up using some feature that still doesn't exist in 3.10.

Alright, let me replace 3.12 with 3.10

juliusvonkohout:
It seems you have a bug as soon as the python version has 3 digits such as 3.10 instead of 3.9

hbelmiro:
> It seems you have a bug as soon as the python version has 3 digits such as 3.10 instead of 3.9

@juliusvonkohout you need to set `"3.10"` (string) instead of `3.10` (number).

PR #11725, opened 2025-03-05
feat(backend/sdk): enable dsl.Collected for parameters & artifacts

### LINKED ISSUE #10050: [sdk] Unable to aggregate results over ParallelFor in Kubeflow V2 using V1 workarounds such as `.after()`

zazulam:
Jumping off from #11627, these changes resolve #10050 in regards to the using `dsl.Collected` for both parameters and artifacts in pipelines. Currently for artifacts, the executor needs to be updated in the sdk and have a release prior to tests being enabled. As for parameters that should work out of the box with solely backend changes.

This PR introduces a few new methods to help ease with detecting and resolving collected inputs as well as cleaning up some of the shared logic between `resolveUpstreamParameters` & `resolveUpstreamArtifacts`.

A few things to keep in mind:
A **parallelFor** creates `x+1` dags, where `x` is the number of iterations. The extra dag is something I like to refer to as the **parallelFor** **_Head_** dag, being that it only contains the iteration dags within it.

With these changes, the keys within the task map returned from `getDAGTasks` will now contain the parent dag id associated with the task. This is to help maintain uniqueness, also when detecting a parallelFor iteration task/dag, the index will be added to the task name to further prevent any potential collisions and map the appropriate tasks within the same iteration for input/output resolution.

When resolving the inputs, if the current Task is a DAG the driver now first checks if the it is a **parallelFor** by inspecting if there exists an `iteration_count` custom property, as only **parallelFor heads** have that property. If found, the iteration dag names will be added to a queue that the `CollectInputs` will use to start the resolution search.

New methods:
* `CollectInputs` - Performs a BFS on the tasks passed to reach the final producer task iterations and collects the values into the appropriate structure for either Artifacts or Parameters
* `CollectContainerOutput` - Helper function for processing the output of a containerExecution, currently only used in the `CollectInputs`.
*` GetProducerTask` - Performs the check on a task for the potential new producer task and output parameter/artifact key and returns the updated value if found.
* `InferIndexedTaskName` - Used to update the producer task with the appropriate index if the current dag context is a parallelFor iteration dag i.e. resolving task inputs within the loop.
* `getParallelForIterationCount` - Helper to get the number of iterations / determine if a parallelFor head dag.
* `GetParallelForTaskName` - Used to update the task name with the iteration value supplied.
* `GetTaskNameWithDagID` - Used to update the task name with the it's parent's dag id.

**Checklist:**

### DIFF
--- a/backend/src/v2/component/launcher_v2.go
+++ b/backend/src/v2/component/launcher_v2.go
@@ -636,51 +636,53 @@ func downloadArtifacts(ctx context.Context, executorInput *pipelinespec.Executor
 		if len(artifactList.Artifacts) == 0 {
 			continue
 		}
-		inputArtifact := artifactList.Artifacts[0]
-
-		localPath, err := LocalPathForURI(inputArtifact.Uri)
-		if err != nil {
-			glog.Warningf("Input Artifact %q does not have a recognized storage URI %q. Skipping downloading to local path.", name, inputArtifact.Uri)
-
-			continue
-		}
-
-		// OCI artifacts are accessed via shared storage of a Modelcar
-		if strings.HasPrefix(inputArtifact.Uri, "oci://") {
-			err := waitForModelcar(inputArtifact.Uri, localPath)
+		for _, artifact := range artifactList.Artifacts {
+			// Iterating through the artifact list allows for collected artifacts to be properly consumed.
+			inputArtifact := artifact
+			localPath, err := LocalPathForURI(inputArtifact.Uri)
 			if err != nil {
-				return err
+				glog.Warningf("Input Artifact %q does not have a recognized storage URI %q. Skipping downloading to local path.", name, inputArtifact.Uri)
+
+				continue
 			}
 
-			continue
-		}
+			// OCI artifacts are accessed via shared storage of a Modelcar
+			if strings.HasPrefix(inputArtifact.Uri, "oci://") {
+				err := waitForModelcar(inputArtifact.Uri, localPath)
+				if err != nil {
+					return err
+				}
 
-		// Copy artifact to local storage.
-		copyErr := func(err error) error {
-			return fmt.Errorf("failed to download input artifact %q from remote storage URI %q: %w", name, inputArtifact.Uri, err)
-		}
-		// TODO: Selectively copy artifacts for which .path was actually specified
-		// on the command line.
-		bucket := defaultBucket
-		bucketConfig := defaultBucketConfig
-		if !strings.HasPrefix(inputArtifact.Uri, defaultBucketConfig.PrefixedBucket()) {
-			nonDefaultBucketConfig, err := objectstore.ParseBucketConfigForArtifactURI(inputArtifact.Uri)
+				continue
+			}
+
+			// Copy artifact to local storage.
+			copyErr := func(err error) error {
+				return fmt.Errorf("failed to download input artifact %q from remote storage URI %q: %w", name, inputArtifact.Uri, err)
+			}
+			// TODO: Selectively copy artifacts for which .path was actually specified
+			// on the command line.
+			bucket := defaultBucket
+			bucketConfig := defaultBucketConfig
+			if !strings.HasPrefix(inputArtifact.Uri, defaultBucketConfig.PrefixedBucket()) {
+				nonDefaultBucketConfig, err := objectstore.ParseBucketConfigForArtifactURI(inputArtifact.Uri)
+				if err != nil {
+					return fmt.Errorf("failed to parse bucketConfig for output artifact %q with uri %q: %w", name, inputArtifact.GetUri(), err)
+				}
+				nonDefaultBucket, ok := nonDefaultBuckets[nonDefaultBucketConfig.PrefixedBucket()]
+				if !ok {
+					return fmt.Errorf("failed to get bucket when downloading input artifact %s with bucket key %s: %w", name, nonDefaultBucketConfig.PrefixedBucket(), err)
+				}
+				bucket = nonDefaultBucket
+				bucketConfig = nonDefaultBucketConfig
+			}
+			blobKey, err := bucketConfig.KeyFromURI(inputArtifact.Uri)
 			if err != nil {
-				return fmt.Errorf("failed to parse bucketConfig for output artifact %q with uri %q: %w", name, inputArtifact.GetUri(), err)
+				return copyErr(err)
 			}
-			nonDefaultBucket, ok := nonDefaultBuckets[nonDefaultBucketConfig.PrefixedBucket()]
-			if !ok {
-				return fmt.Errorf("failed to get bucket when downloading input artifact %s with bucket key %s: %w", name, nonDefaultBucketConfig.PrefixedBucket(), err)
+			if err := objectstore.DownloadBlob(ctx, bucket, localPath, blobKey); err != nil {
+				return copyErr(err)
 			}
-			bucket = nonDefaultBucket
-			bucketConfig = nonDefaultBucketConfig
-		}
-		blobKey, err := bucketConfig.KeyFromURI(inputArtifact.Uri)
-		if err != nil {
-			return copyErr(err)
-		}
-		if err := objectstore.DownloadBlob(ctx, bucket, localPath, blobKey); err != nil {
-			return copyErr(err)
 		}
 
 	}
--- a/backend/src/v2/component/launcher_v2.go
+++ b/backend/src/v2/component/launcher_v2.go
@@ -139,29 +139,31 @@ func stopWaitingArtifacts(artifacts map[string]*pipelinespec.ArtifactList) {
 		}
 
 		// Following the convention of downloadArtifacts in the launcher to only look at the first in the list.
-		inputArtifact := artifactList.Artifacts[0]
+		for _, artifact := range artifactList.Artifacts {
+			inputArtifact := artifact
 
-		// This should ideally verify that this is also a model input artifact, but this metadata doesn't seem to
-		// be set on inputArtifact.
-		if !strings.HasPrefix(inputArtifact.Uri, "oci://") {
-			continue
-		}
+			// This should ideally verify that this is also a model input artifact, but this metadata doesn't seem to
+			// be set on inputArtifact.
+			if !strings.HasPrefix(inputArtifact.Uri, "oci://") {
+				continue
+			}
 
-		localPath, err := LocalPathForURI(inputArtifact.Uri)
-		if err != nil {
-			continue
-		}
+			localPath, err := LocalPathForURI(inputArtifact.Uri)
+			if err != nil {
+				continue
+			}
 
-		glog.Infof("Stopping Modelcar container for artifact %s", inputArtifact.Uri)
+			glog.Infof("Stopping Modelcar container for artifact %s", inputArtifact.Uri)
 
-		launcherCompleteFile := strings.TrimSuffix(localPath, "/models") + "/launcher-complete"
-		_, err = os.Create(launcherCompleteFile)
-		if err != nil {
-			glog.Errorf(
-				"Failed to stop the artifact %s by creating %s: %v", inputArtifact.Uri, launcherCompleteFile, err,
-			)
+			launcherCompleteFile := strings.TrimSuffix(localPath, "/models") + "/launcher-complete"
+			_, err = os.Create(launcherCompleteFile)
+			if err != nil {
+				glog.Errorf(
+					"Failed to stop the artifact %s by creating %s: %v", inputArtifact.Uri, launcherCompleteFile, err,
+				)
 
-			continue
+				continue
+			}
 		}
 	}
 }
--- a/backend/src/v2/driver/driver.go
+++ b/backend/src/v2/driver/driver.go
@@ -19,11 +19,12 @@ import (
 	"encoding/json"
 	"errors"
 	"fmt"
-	"github.com/kubeflow/pipelines/backend/src/apiserver/config/proxy"
 	"slices"
 	"strconv"
 	"strings"
 	"time"
+
+	"github.com/kubeflow/pipelines/backend/src/apiserver/config/proxy"
 
 	"github.com/kubeflow/pipelines/backend/src/common/util"
 	"github.com/kubeflow/pipelines/backend/src/v2/objectstore"
@@ -628,7 +629,7 @@ func addModelcarsToPodSpec(
 			continue
 		}
 		// Following the convention of downloadArtifacts in the launcher to look at the entire list.
-		for _, artifact := range artifactList.Artifacts {
+		for index, artifact := range artifactList.Artifacts {
 			inputArtifact := artifact
 
 			// This should ideally verify that this is also a model input artifact, but this metadata doesn't seem to
@@ -637,8 +638,9 @@ func addModelcarsToPodSpec(
 				continue
 			}
 
-			modelcarArtifacts[name] = inputArtifact
-			modelcarArtifactNames = append(modelcarArtifactNames, name)
+			artifactName := fmt.Sprintf("%s-%d", name, index)
+			modelcarArtifacts[artifactName] = inputArtifact
+			modelcarArtifactNames = append(modelcarArtifactNames, artifactName)
 		}
 	}
 
@@ -1685,7 +1687,7 @@ func resolveInputParameterStr(
 
 	if typedVal, ok := val.GetKind().(*structpb.Value_StringValue); ok && typedVal != nil {
 		if typedVal.StringValue == "" {
-			return nil, fmt.Errorf("resolving input parameter with spec %s. Expected a non-empty string.", paramSpec)
+			return nil, fmt.Errorf("resolving input parameter with spec %s. Expected a non-empty string", paramSpec)
 		}
 	} else {
 		return nil, fmt.Errorf("resolving input parameter with spec %s. Expected a string but got: %T", paramSpec, val.GetKind())
@@ -1812,16 +1814,20 @@ func resolveUpstreamParameters(cfg resolveUpstreamOutputsConfig) (*structpb.Valu
 		return nil, cfg.err(fmt.Errorf("output parameter key is empty"))
 	}
 
-	// For the scenario where 2 tasks are defined within a ParallelFor and 1 receives the output of the other we must ensure that the downstream task resolves the approriate output of the iteration it is in. With knowing if we are resolving inputs for a task within a ParallelFor DAG we can add the iteration index to the producerTaskName so that we can resolve the correct iteration of that task.
-	isParallelForDAG := cfg.dag.Execution.GetExecution().GetCustomProperties()["iteration_index"] != nil
-	if isParallelForDAG {
-		// This is needed to support tasks within a ParallelFor Loop that do not leverage the iterator values but still consume outputs from ones that do
-		task_iteration_index := cfg.dag.Execution.GetExecution().GetCustomProperties()["iteration_index"].GetIntValue()
-		producerTaskName = metadata.GetParallelForTaskName(producerTaskName, task_iteration_index)
-		glog.Infof("Attempting to retrieve DAG Tasks from a parallelFor DAG")
-	}
-	// Get a list of tasks for the current DAG first.
-	// The reason we use getDAGTasks instead of mlmd.GetExecutionsInDAG without the dag filter is because the latter does not handle task name collisions in the map which results in a bunch of unhandled edge cases and test failures.
+	producerTaskName = metadata.GetTaskNameWithDagID(producerTaskName, cfg.dag.Execution.GetID())
+	// For the scenario where 2 tasks are defined within a ParallelFor and 1
+	// receives the output of the other we must ensure that the downstream task
+	// resolves the approriate output of the iteration it is in. With knowing if
+	// we are resolving inputs for a task within a ParallelFor DAG we can add
+	// the iteration index to the producerTaskName so that we can resolve the
+	// correct iteration of that task.
+	producerTaskName = InferIndexedTaskName(producerTaskName, cfg.dag.Execution)
+	// Get a list of tasks for the current DAG first. The reason we use
+	// getDAGTasks instead of mlmd.GetExecutionsInDAG without the dag filter is
+	// because the latter does not handle task name collisions in the map which
+	// results in a bunch of unhandled edge cases and test failures.
+	glog.V(4).Infof("producerTaskName: %v", producerTaskName)
+	glog.V(4).Infof("outputParameterKey: %v", outputParameterKey)
 	tasks, err := getDAGTasks(cfg.ctx, cfg.dag, cfg.pipeline, cfg.mlmd, nil)
 	if err != nil {
 		return nil, cfg.err(err)
@@ -1834,6 +1840,7 @@ func resolveUpstreamParameters(cfg resolveUpstreamOutputsConfig) (*structpb.Valu
 	glog.V(4).Info("producer: ", producer)
 	glog.V(4).Infof("tasks: %#v", tasks)
 	currentTask := producer
+	subTaskName := producerTaskName
 	// Continue looping until we reach a sub-task that is NOT a DAG.
 	for {
 		glog.V(4).Info("currentTask: ", currentTask.TaskName())
@@ -1863,74 +1870,36 @@ func resolveUpstreamParameters(cfg resolveUpstreamOutputsConfig) (*structpb.Valu
 
 			glog.V(4).Infof("Deserialized dagOutputParametersMap: %v", dagOutputParametersMap)
 
+			// For this section, if the currentTask we are looking for is within
+			// a ParallelFor DAG, this means the actual task that produced the
+			// output we need has multiple iterations so we have to gather all
+			// them and fan them in by collecting them into a list i.e.
+			// kfp.dsl.Collected support.
+			parentDAG, err := cfg.mlmd.GetExecution(cfg.ctx, currentTask.GetExecution().GetCustomProperties()["parent_dag_id"].GetIntValue())
+			if err != nil {
+				return nil, cfg.err(err)
+			}
+			iterations := getParallelForIterationCount(currentTask, parentDAG)
+			if iterations > 0 {
+				parameterList, _, err := CollectInputs(cfg, subTaskName, tasks, outputParameterKey, false)
+				if err != nil {
+					return nil, cfg.err(err)
+				}
+				return parameterList, nil
+			}
 			// Support for the 2 DagOutputParameterSpec types:
 			// ValueFromParameter & ValueFromOneof
-			var subTaskName string
-			switch dagOutputParametersMap[outputParameterKey].Kind.(type) {
-			case *pipelinespec.DagOutputsSpec_DagOutputParameterSpec_ValueFromParameter:
-				subTaskName = dagOutputParametersMap[outputParameterKey].GetValueFromParameter().GetProducerSubtask()
-				outputParameterKey = dagOutputParametersMap[outputParameterKey].GetValueFromParameter().GetOutputParameterKey()
-			case *pipelinespec.DagOutputsSpec_DagOutputParameterSpec_ValueFromOneof:
-				// When OneOf is specified in a pipeline, the output of only 1 task is consumed even though there may be more than 1 task output set. In this case we will attempt to grab the first successful task output.
-				paramSelectors := dagOutputParametersMap[outputParameterKey].GetValueFromOneof().GetParameterSelectors()
-				glog.V(4).Infof("paramSelectors: %v", paramSelectors)
-				// Since we have the tasks map, we can iterate through the parameterSelectors if the ProducerSubTask is not present in the task map and then assign the new OutputParameterKey only if it exists.
-				successfulOneOfTask := false
-				for !successfulOneOfTask {
-					for _, paramSelector := range paramSelectors {
-						subTaskName = paramSelector.GetProducerSubtask()
-						glog.V(4).Infof("subTaskName from paramSelector: %v", subTaskName)
-						glog.V(4).Infof("outputParameterKey from paramSelector: %v", paramSelector.GetOutputParameterKey())
-						if subTask, ok := tasks[subTaskName]; ok {
-							subTaskState := subTask.GetExecution().LastKnownState.String()
-							glog.V(4).Infof("subTask: %w , subTaskState: %v", subTaskName, subTaskState)
-							if subTaskState == "CACHED" || subTaskState == "COMPLETE" {
-
-								outputParameterKey = paramSelector.GetOutputParameterKey()
-								successfulOneOfTask = true
-								break
-							}
-						}
-					}
-					if !successfulOneOfTask {
-						return nil, cfg.err(fmt.Errorf("processing OneOf: No successful task found"))
-					}
-				}
+			subTaskName, outputParameterKey, err = GetProducerTask(currentTask, tasks, subTaskName, outputParameterKey, false)
+			if err != nil {
+				return nil, cfg.err(err)
 			}
 			glog.V(4).Infof("SubTaskName from outputParams: %v", subTaskName)
 			glog.V(4).Infof("OutputParameterKey from outputParams: %v", outputParameterKey)
 			if subTaskName == "" {
 				return nil, cfg.err(fmt.Errorf("producer_subtask not in outputParams"))
 			}
-			// For this section, if the currentTask we are looking for is within a ParallelFor DAG, this means the actual task that produced the output we need has multiple iterations so we have to gather all them  and fan them in by collecting them into a list i.e. kfp.dsl.Collected support.
-			if currentTask.GetExecution().GetCustomProperties()["iteration_count"] != nil {
-				glog.V(4).Infof("currentTask, %v, is a ParallelFor DAG. Attempting to gather all producer_subtask iterations of %w", currentTask.TaskName(), subTaskName)
 
-				parallelForOutputList := make([]*structpb.Value, 0)
-				// Iterate over the parallelFor task iterations and gather the output of the subTask.
-				for i := range currentTask.GetExecution().GetCustomProperties()["iteration_count"].GetIntValue() {
-					// Follow the convention set in mlmd.GetExecutionsInDAG for ParallelFor tasks
-					subTaskIterationName := metadata.GetParallelForTaskName(subTaskName, i)
-					glog.V(4).Infof("subTaskIterationName: %v", subTaskIterationName)
-					subTask, ok := tasks[subTaskIterationName]
-					if !ok {
-						return nil, cfg.err(fmt.Errorf("subTaskName, %v, not in tasks", subTaskName))
-					}
-					_, outputParametersCustomProperty, err := subTask.GetParameters()
-					if err != nil {
-						return nil, err
-					}
-					parallelForOutputList = append(parallelForOutputList, outputParametersCustomProperty[outputParameterKey])
-				}
-				return &structpb.Value{
-					Kind: &structpb.Value_ListValue{
-						ListValue: &structpb.ListValue{
-							Values: parallelForOutputList,
-						},
-					},
-				}, nil
-			}
-
+			// If the sub-task is a DAG, reassign currentTask and run
 			glog.V(4).Infof(
 				"Overriding currentTask, %v, output with currentTask's producer_subtask, %v, output.",
 				currentTask.TaskName(),
@@ -1967,84 +1936,62 @@ func resolveUpstreamArtifacts(cfg resolveUpstreamOutputsConfig) (*pipelinespec.A
 	if taskOutput.GetOutputArtifactKey() == "" {
 		cfg.err(fmt.Errorf("output artifact key is empty"))
 	}
-
-	isParallelForDAG := cfg.dag.Execution.GetExecution().GetCustomProperties()["iteration_index"] != nil
-	if isParallelForDAG {
-		// This is needed to support tasks within a ParallelFor Loop that do not leverage the iterator values but still consume outputs from ones that do
-		task_iteration_index := cfg.dag.Execution.GetExecution().GetCustomProperties()["iteration_index"].GetIntValue()
-		producerTaskName = metadata.GetParallelForTaskName(producerTaskName, task_iteration_index)
-		glog.Infof("Attempting to retrieve DAG Tasks from a parallelFor DAG")
-	}
-
+	producerTaskName = metadata.GetTaskNameWithDagID(producerTaskName, cfg.dag.Execution.GetID())
+	// The main difference between the root ParallelFor DAG and its iteration
+	// DAGs is that the root contains the custom property "iteration_count"
+	// while the iterations contain "iteration_index". We can use this to
+	// determine if we are in a ParallelFor DAG or not. The iteration DAGs will
+	// contain the "iteration_index" which is used to resolve the correct output
+	// artifact for the downstream task within the iteration. ParallelFor
+	// iterations are DAGs themselves, we can verify if we are in a iteration by
+	// confirming that the "iteration_index" exists for the DAG of the current
+	// task we are attempting to resolve. If the dag contains the
+	// "iteration_index", the producerTaskName will be updated appropriately
+	producerTaskName = InferIndexedTaskName(producerTaskName, cfg.dag.Execution)
+	glog.V(4).Infof("producerTaskName: %v", producerTaskName)
 	tasks, err := getDAGTasks(cfg.ctx, cfg.dag, cfg.pipeline, cfg.mlmd, nil)
 	if err != nil {
-		cfg.err(err)
+		return nil, cfg.err(err)
 	}
 
 	producer, ok := tasks[producerTaskName]
 	if !ok {
-		cfg.err(
+		return nil, cfg.err(
 			fmt.Errorf("cannot find producer task %q", producerTaskName),
 		)
 	}
 	glog.V(4).Info("producer: ", producer)
+	glog.V(4).Infof("tasks: %#v", tasks)
 	currentTask := producer
 	outputArtifactKey := taskOutput.GetOutputArtifactKey()
-
-	// Continue looping until we reach a sub-task that is either a ParallelFor task or a Container task.
+	subTaskName := producerTaskName
+	// Continue looping until we reach a sub-task that is either a ParallelFor
+	// task or a Container task.
 	for {
 		glog.V(4).Info("currentTask: ", currentTask.TaskName())
 		// If the current task is a DAG:
 		if *currentTask.GetExecution().Type == "system.DAGExecution" {
 			// Get the sub-task.
-			outputArtifactsCustomProperty := currentTask.GetExecution().GetCustomProperties()["artifact_producer_task"]
-			// Deserialize the output artifacts.
-			var outputArtifacts map[string]*pipelinespec.DagOutputsSpec_DagOutputArtifactSpec
-			err := json.Unmarshal([]byte(outputArtifactsCustomProperty.GetStringValue()), &outputArtifacts)
+			parentDAG, err := cfg.mlmd.GetExecution(cfg.ctx, currentTask.GetExecution().GetCustomProperties()["parent_dag_id"].GetIntValue())
 			if err != nil {
-				return nil, err
+				return nil, cfg.err(err)
 			}
-			glog.V(4).Infof("Deserialized outputArtifacts: %v", outputArtifacts)
-			// Adding support for multiple output artifacts
-			var subTaskName string
-			artifactSelectors := outputArtifacts[outputArtifactKey].GetArtifactSelectors()
-			artifactSelectorValue := artifactSelectors[len(artifactSelectors)-1]
-
-			subTaskName = artifactSelectorValue.ProducerSubtask
-			outputArtifactKey = artifactSelectorValue.OutputArtifactKey
+			iterations := getParallelForIterationCount(currentTask, parentDAG)
+			if iterations > 0 {
+				_, artifactList, err := CollectInputs(cfg, subTaskName, tasks, outputArtifactKey, true)
+				if err != nil {
+					return nil, cfg.err(err)
+				}
+				return artifactList, nil
+			}
+			subTaskName, outputArtifactKey, err = GetProducerTask(currentTask, tasks, subTaskName, outputArtifactKey, true)
+			if err != nil {
+				return nil, cfg.err(err)
+			}
 			glog.V(4).Infof("ProducerSubtask: %v", subTaskName)
 			glog.V(4).Infof("OutputArtifactKey: %v", outputArtifactKey)
-
-			// For this section, it's the same as the one in resolveUpstreamParameters, but for artifacts.
-			//TODO: Conslidate Paremeter and Artifact resolution logic.
-			if currentTask.GetExecution().GetCustomProperties()["iteration_count"] != nil {
-				glog.V(4).Infof("currentTask, %v, is a ParallelFor DAG. Attempting to gather all producer_subtask iterations of %w", currentTask.TaskName(), subTaskName)
-
-				parallelForOutputList := make([]*pipelinespec.RuntimeArtifact, 0)
-				for i := range currentTask.GetExecution().GetCustomProperties()["iteration_count"].GetIntValue() {
-					// Follow the task name convention set in mlmd.GetExecutionsInDAG for ParallelFor tasks
-					subTaskIterationName := metadata.GetParallelForTaskName(subTaskName, i)
-					glog.V(4).Infof("subTaskIterationName: %v", subTaskIterationName)
-					subTask, ok := tasks[subTaskIterationName]
-					if !ok {
-						return nil, cfg.err(fmt.Errorf("subTaskName, %v, not in tasks", subTaskName))
-					}
-					outputArtifacts, err := cfg.mlmd.GetOutputArtifactsByExecutionId(cfg.ctx, subTask.GetID())
-					if err != nil {
-						return nil, err
-					}
-
-					runtimeArtifact, err := outputArtifacts[outputArtifactKey].ToRuntimeArtifact()
-					if err != nil {
-						return nil, cfg.err(err)
-					}
-					parallelForOutputList = append(parallelForOutputList, runtimeArtifact)
-				}
-				return &pipelinespec.ArtifactList{
-					Artifacts: parallelForOutputList,
-				}, nil
-			}
 			// If the sub-task is a DAG, reassign currentTask and run
+			glog.V(4).Infof("currentTask ID: %v", currentTask.GetID())
 			glog.V(4).Infof(
 				"Overriding currentTask, %v, output with currentTask's producer_subtask, %v, output.",
 				currentTask.TaskName(),
@@ -2054,7 +2001,6 @@ func resolveUpstreamArtifacts(cfg resolveUpstreamOutputsConfig) (*pipelinespec.A
 			if !ok {
 				return nil, cfg.err(fmt.Errorf("subTaskName, %v, not in tasks", subTaskName))
 			}
-
 		} else {
 			// Base case, currentTask is a container, not a DAG.
 			outputs, err := cfg.mlmd.GetOutputArtifactsByExecutionId(cfg.ctx, currentTask.GetID())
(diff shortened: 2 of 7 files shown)


zazulam:
Providing some more context, this is the error that occurs for artifacts when a user attempts use either `List[Artifact]`:

```
main     executor = component_executor.Executor(
main   File "/usr/local/lib/python3.9/site-packages/kfp/dsl/executor.py", line 52, in __init__
main     self.assign_input_and_output_artifacts()
main   File "/usr/local/lib/python3.9/site-packages/kfp/dsl/executor.py", line 73, in assign_input_and_output_artifacts
main     self.input_artifacts[name] = [
main   File "/usr/local/lib/python3.9/site-packages/kfp/dsl/executor.py", line 74, in <listcomp>
main     self.make_artifact(
main   File "/usr/local/lib/python3.9/site-packages/kfp/dsl/executor.py", line 118, in make_artifact
main     return create_artifact_instance(
main   File "/usr/local/lib/python3.9/site-packages/kfp/dsl/executor.py", line 378, in create_artifact_instance
main     ) if hasattr(artifact_cls, '_from_executor_fields') else artifact_cls(
main   File "/usr/local/lib/python3.9/typing.py", line 685, in __call__
main     raise TypeError(f"Type {self._name} cannot be instantiated; "
main TypeError: Type List cannot be instantiated; use list() instead
```

or `list[Artifact]`

```
main     return _run_code(code, main_globals, None,
main   File "/usr/local/lib/python3.9/runpy.py", line 87, in _run_code
main     exec(code, run_globals)
main   File "/usr/local/lib/python3.9/site-packages/kfp/dsl/executor_main.py", line 109, in <module>
main     executor_main()
main   File "/usr/local/lib/python3.9/site-packages/kfp/dsl/executor_main.py", line 98, in executor_main
main     executor = component_executor.Executor(
main   File "/usr/local/lib/python3.9/site-packages/kfp/dsl/executor.py", line 52, in __init__
main     self.assign_input_and_output_artifacts()
main   File "/usr/local/lib/python3.9/site-packages/kfp/dsl/executor.py", line 73, in assign_input_and_output_artifacts
main     self.input_artifacts[name] = [
main   File "/usr/local/lib/python3.9/site-packages/kfp/dsl/executor.py", line 74, in <listcomp>
main     self.make_artifact(
main   File "/usr/local/lib/python3.9/site-packages/kfp/dsl/executor.py", line 118, in make_artifact
main     return create_artifact_instance(
main   File "/usr/local/lib/python3.9/site-packages/kfp/dsl/executor.py", line 378, in create_artifact_instance
main     ) if hasattr(artifact_cls, '_from_executor_fields') else artifact_cls(
main TypeError: list() takes no keyword arguments
```

So we use the `get_args` & `get_origin` to determine the actual artifact cls type used within the collection.

I had made a custom image to test with the sdk installed with the executor changes, this is what it looks like:
image

sdk/python/kfp/dsl/executor.py
```diff
@@ -371,15 +372,21 @@ def create_artifact_instance(
...
-        uri=runtime_artifact.get('uri', ''),
-        name=runtime_artifact.get('name', ''),
-        metadata=runtime_artifact.get('metadata', {}),
-    ) if hasattr(artifact_cls, '_from_executor_fields') else artifact_cls(
-        uri=runtime_artifact.get('uri', ''),
-        name=runtime_artifact.get('name', ''),
-        metadata=runtime_artifact.get('metadata', {}),
-    )
+
+    if get_origin(artifact_cls) is list or get_origin(artifact_cls) is List:
```
chensun, line 376: This might not be the right place for the change. Based on the method name and description, it seems meant to be handling single artifact. Looking at where the method is used, it seems to me this method is only used for local execution? and next to its caller `runtime_artifact_to_dsl_artifact` there's also a `artifact_list_to_dsl_artifact` method:
https://github.com/kubeflow/pipelines/blob/8d0ae5381e8366905c90009c56fd0e4807e94f0f/sdk/python/kfp/local/executor_output_utils.py#L181-L189

Can you please double check if this is the right SDK change?

zazulam: @chensun Thanks for double checking that! I was treating the fanned in artifacts each as their own runtimeArtifact since the structure that we need for the [`ExecutorInputs_Inputs_Artifacts`](https://github.com/kubeflow/pipelines/blob/8d0ae5381e8366905c90009c56fd0e4807e94f0f/api/v2alpha1/pipeline_spec.proto#L935-L973) is of `map<string, ArtifactList[RuntimeArtifact] >`.

I can play with updating the pipelinespec to see what changes to the `ExecutorInputs_Inputs_Artifacts` would be needed for this to work smoothly and hopefully avoid the changes to the SDK.

zazulam: After looking into it a bit, I was testing adding proto changes to allow for either an ArtifactList or RuntimeArtifact as oneof the values for Artifacts within an ArtifactList through some wrapper. That route seems to complicate the code around artifacts in both the driver and launcher, so I would prefer avoiding that option. 

Referring to the current change back to the previous example, the value of ExecutorInput that is passed to the main container of the launcher pod contains a proper artifact list.


Because the Artifact's value is already an ArtifactList I would assume we could just add to that for fanning in from the producers in the loop. I think if anything the Schema piece might not be set properly as the compiled yaml does indeed have `isArtifactList: true` set but the handling for the executor in the main container seems to pick up the schema for each artifact as the List[Artifact] from the error above:
```
main   File "/usr/local/lib/python3.9/site-packages/kfp/dsl/executor.py", line 74, in <listcomp>
main     self.make_artifact(
```


ExecutorInput:
```yaml
{
    "inputs": {
        "artifacts": {
            "files": {
                "artifacts": [
                    {
                        "name": "385",
                        "type": {
                            "instanceSchema": ""
                        },
                        "uri": "minio://mlpipeline/v2/artifacts/collected-artifact-pipeline/14159b34-7c25-4324-88de-65cd6b9a5cfe/create-file/cbbc7614-211d-48da-9757-dd5f14153864/file",
                        "metadata": {
                            "display_name": "file",
                            "store_session_info": "{\"Provider\":\"minio\",\"Params\":{\"accessKeyKey\":\"accesskey\",\"disableSSL\":\"true\",\"endpoint\":\"10.96.103.164:9000\",\"fromEnv\":\"false\",\"region\":\"minio\",\"secretKeyKey\":\"secretkey\",\"secretName\":\"mlpipeline-minio-artifact\"}}"
                        }
                    },
                    {
                        "name": "383",
                        "type": {
                            "instanceSchema": ""
                        },
                        "uri": "minio://mlpipeline/v2/artifacts/collected-artifact-pipeline/14159b34-7c25-4324-88de-65cd6b9a5cfe/create-file/b9038899-3f42-443a-bee0-b0385fc0ed7e/file",
                        "metadata": {
                            "display_name": "file",
                            "store_session_info": "{\"Provider\":\"minio\",\"Params\":{\"accessKeyKey\":\"accesskey\",\"disableSSL\":\"true\",\"endpoint\":\"10.96.103.164:9000\",\"fromEnv\":\"false\",\"region\":\"minio\",\"secretKeyKey\":\"secretkey\",\"secretName\":\"mlpipeline-minio-artifact\"}}"
                        }
                    },
                    {
                        "name": "384",
                        "type": {
                            "instanceSchema": ""
                        },
                        "uri": "minio://mlpipeline/v2/artifacts/collected-artifact-pipeline/14159b34-7c25-4324-88de-65cd6b9a5cfe/create-file/7092ff7f-5fb5-464e-8321-e4a34b939cd5/file",
                        "metadata": {
                            "display_name": "file",
                            "store_session_info": "{\"Provider\":\"minio\",\"Params\":{\"accessKeyKey\":\"accesskey\",\"disableSSL\":\"true\",\"endpoint\":\"10.96.103.164:9000\",\"fromEnv\":\"false\",\"region\":\"minio\",\"secretKeyKey\":\"secretkey\",\"secretName\":\"mlpipeline-minio-artifact\"}}"
                        }
                    }
                ]
            }
        }
    },
    "outputs": {
        "parameters": {
            "Output": {
                "outputFile": "/tmp/kfp/outputs/Output"
            }
        },
        "outputFile": "/tmp/kfp_outputs/output_metadata.json"
    }
}
```



Snippet from compiled IR:
```yaml
    outputDefinitions:
      artifacts:
        pipelinechannel--create-file-file:
          artifactType:
            schemaTitle: system.Artifact
            schemaVersion: 0.0.1
          isArtifactList: true
  comp-read-files:
    executorLabel: exec-read-files
    inputDefinitions:
      artifacts:
        files:
          artifactType:
            schemaTitle: system.Artifact
            schemaVersion: 0.0.1
          isArtifactList: true
```

I'll revisit the artifact list generation in the backend to verify that the individual schema is propagating correctly.

zazulam: @chensun it seemed that the artifact creation from the[ list comprehension](https://github.com/kubeflow/pipelines/pull/11725/files#diff-1213dafd823d362320792c2c425f88b763e14be8c349642387132e592b1ccf2eR73-R85) was using the type of the component func kwarg for each artifact, which resulted in the list type propagating to the previous function I was modifying. This change makes more sense, thanks for that callout.

zazulam:
After rebasing CI started to fail.

zazulam:
Tests are finally passing after rebasing again 🙏

zazulam:
I think there should be some consensus on how the UI should depict the fan-in of artifacts since there is a dag UI element associated with them, but we can save that for another PR 😃.

samples/v2/sample_test.py
```diff
@@ -20,16 +20,19 @@
 from typing import List
 import unittest
 
+import collected_parameters
 import component_with_optional_inputs
 import hello_world
 import kfp
 from kfp.dsl.graph_component import GraphComponent
+from modelcar import modelcar
```
mprahl, line 28: Could you please fix the imports? I think as part of a rebase, this PR added duplicate imports for `modelcar` and `pipeline_with_placeholders`.

backend/src/v2/component/launcher_v2.go
```diff
@@ -562,51 +562,53 @@ func downloadArtifacts(ctx context.Context, executorInput *pipelinespec.Executor
 		if len(artifactList.Artifacts) == 0 {
 			continue
 		}
-		inputArtifact := artifactList.Artifacts[0]
```
mprahl, line 565: There are other parts of the code assume the artifacts list is either a size of 0 or 1. See `addModelcarsToPodSpec` as an example. So at the very least, `addModelcarsToPodSpec` probably needs to be adjusted to look at the whole list as well or else `waitForModelcar` may hang forever for a model car that was never added as a side car container.

backend/src/v2/driver/driver.go
```diff
@@ -1645,8 +1646,8 @@ func getDAGTasks(
 			// handled with dsl.Collected.
```
mprahl, line 1646: Is this comment still valid since my understanding is that this PR changes `GetExecutionsInDAG` to return unique task names if ParallelFor is used?

backend/src/v2/driver/driver.go
```diff
@@ -1645,8 +1646,8 @@ func getDAGTasks(
...
 			_, ok := v.GetExecution().GetCustomProperties()["iteration_count"]
 			if ok {
-				glog.Infof("Found a ParallelFor task, %v. Skipping it.", v.TaskName())
-				continue
+				glog.V(4).Infof("Found a ParallelFor task, %v.", v.TaskName())
+				parallelFor = true
```
mprahl, line 1650: Nit: It's hard to read code when a function argument value is overridden. Please consider using a different variable name.

backend/src/v2/driver/driver.go
```diff
@@ -1624,6 +1624,7 @@ func getDAGTasks(
 	pipeline *metadata.Pipeline,
 	mlmd *metadata.Client,
 	flattenedTasks map[string]*metadata.Execution,
+	parallelFor bool,
```
mprahl, line 1627: What does this function argument do?

backend/src/v2/driver/driver.go
```diff
@@ -1777,14 +1785,42 @@ func resolveUpstreamParameters(cfg resolveUpstreamOutputsConfig) (*structpb.Valu
...
+					outputs_list = append(outputs_list, outputParametersCustomProperty[outputParameterKey])
+				}
+				return &structpb.Value{
+					Kind: &structpb.Value_ListValue{
+						ListValue: &structpb.ListValue{
+							Values: outputs_list,
+						},
+					},
+				}, nil
+			} else {
```
mprahl, line 1814: Nit: You don't need the else since you always return in the `if`. It helps with readability in my opinion to fix this.

backend/src/v2/driver/driver.go
```diff
@@ -1777,14 +1785,42 @@ func resolveUpstreamParameters(cfg resolveUpstreamOutputsConfig) (*structpb.Valu
...
+				glog.V(4).Infof(
```
mprahl, line 1815: Optional: In the ParallelFor case, this log message isn't output anymore. Not sure if something equivalent would be helpful.

backend/src/v2/driver/driver.go
```diff
@@ -1850,10 +1897,46 @@ func resolveUpstreamArtifacts(cfg resolveUpstreamOutputsConfig) (*pipelinespec.A
...
+					runtimeArtifact, err := outputArtifacts[outputArtifactKey].ToRuntimeArtifact()
+					if err != nil {
+						cfg.err(err)
+					}
+					outputs_list = append(outputs_list, runtimeArtifact)
+				}
+				return &pipelinespec.ArtifactList{
+					Artifacts: outputs_list,
+				}, nil
+			} else {
```
mprahl, line 1927: Nit: You don't need the else since you always return in the `if`. It helps with readability in my opinion to fix this.

backend/src/v2/driver/driver.go
```diff
@@ -1850,10 +1897,46 @@ func resolveUpstreamArtifacts(cfg resolveUpstreamOutputsConfig) (*pipelinespec.A
...
-			currentTask = tasks[subTaskName]
-			// }
+
+			if currentTask.GetExecution().GetCustomProperties()["iteration_count"] != nil {
+				glog.V(4).Infof("currentTask, %v, is a ParallelFor DAG. Attempting to gather all producer_subtask iterations of %w", currentTask.TaskName(), subTaskName)
+
+				outputs_list := make([]*pipelinespec.RuntimeArtifact, 0)
+				for i := range currentTask.GetExecution().GetCustomProperties()["iteration_count"].GetIntValue() {
+					// Follow the convention set in mlmd.GetExecutionsInDAG for ParallelFor tasks
+					subTaskIterationName := fmt.Sprintf("%s_idx_%v", subTaskName, i)
```
mprahl, line 1907: Consider adding a function that abstracts deriving the task name when iteration count is set. This `fmt.Sprintf` is used a lot in this PR and the comment about fllowing the convention of `GetExecutionsInDAG` isn't always present.

backend/src/v2/driver/driver.go
```diff
@@ -1850,10 +1897,46 @@ func resolveUpstreamArtifacts(cfg resolveUpstreamOutputsConfig) (*pipelinespec.A
...
-			// If the sub-task is a DAG, reassign currentTask and run
-			// through the loop again.
-			currentTask = tasks[subTaskName]
-			// }
+
+			if currentTask.GetExecution().GetCustomProperties()["iteration_count"] != nil {
+				glog.V(4).Infof("currentTask, %v, is a ParallelFor DAG. Attempting to gather all producer_subtask iterations of %w", currentTask.TaskName(), subTaskName)
+
+				outputs_list := make([]*pipelinespec.RuntimeArtifact, 0)
+				for i := range currentTask.GetExecution().GetCustomProperties()["iteration_count"].GetIntValue() {
```
mprahl, line 1905: Neat, I didn't know about this new Go feature of ranging over an int like that

backend/src/v2/driver/driver.go
```diff
@@ -1850,10 +1897,46 @@ func resolveUpstreamArtifacts(cfg resolveUpstreamOutputsConfig) (*pipelinespec.A
 				subTaskName = v.ProducerSubtask
 				outputArtifactKey = v.OutputArtifactKey
 			}
```
mprahl, line 1899: Why is only the last artifact selector used (i.e. the loop doesn't do anything with the intermediate values)? If that's intentional, why loop through the whole thing?

backend/src/v2/driver/driver.go
```diff
@@ -1850,10 +1897,46 @@ func resolveUpstreamArtifacts(cfg resolveUpstreamOutputsConfig) (*pipelinespec.A
...
+						return nil, err
+					}
+
+					runtimeArtifact, err := outputArtifacts[outputArtifactKey].ToRuntimeArtifact()
+					if err != nil {
+						cfg.err(err)
+					}
+					outputs_list = append(outputs_list, runtimeArtifact)
+				}
+				return &pipelinespec.ArtifactList{
```
mprahl, line 1924: Could you update the comment `// Continue looping until we reach a sub-task that is NOT a DAG.`? It seems to apply anymore since ParallelFor is a DAG.

backend/src/v2/driver/driver.go
```diff
@@ -1850,10 +1897,46 @@ func resolveUpstreamArtifacts(cfg resolveUpstreamOutputsConfig) (*pipelinespec.A
...
+						return nil, cfg.err(fmt.Errorf("subTaskName, %v, not in tasks", subTaskName))
+					}
+					outputArtifacts, err := cfg.mlmd.GetOutputArtifactsByExecutionId(cfg.ctx, subTask.GetID())
+					if err != nil {
+						return nil, err
+					}
+
+					runtimeArtifact, err := outputArtifacts[outputArtifactKey].ToRuntimeArtifact()
+					if err != nil {
+						cfg.err(err)
```
mprahl, line 1920: Did you mean to return this error instead?

samples/v2/collected_artifacts.py
```diff
@@ -0,0 +1,74 @@
...
+
+    read_files_op = read_files(files=dag.output)
+    read_files_op.set_caching_options(False)
+
+
+if __name__ == '__main__':
+    # Compiler().compile(pipeline_func=collected_artifacts_pipeline, package_path=f"{__file__.removesuffix('.py')}.yaml")
+    client = kfp.Client()
+    run = client.create_run_from_pipeline_func(
+        collected_artifact_pipeline, arguments={})
```
mprahl, line 74: Consider setting `enable_caching=False` to avoid having to set `set_caching_options` everywhere and accidentally missing one.

samples/v2/collected_artifacts.py
```diff
@@ -0,0 +1,74 @@
...
+    model_ids = 's1,s2,s3'
+    dag = collecting_artifacts(model_ids=model_ids)
+    dag.set_caching_options(False)
+
+    read_files_op = read_files(files=dag.output)
+    read_files_op.set_caching_options(False)
+
+
+if __name__ == '__main__':
+    # Compiler().compile(pipeline_func=collected_artifacts_pipeline, package_path=f"{__file__.removesuffix('.py')}.yaml")
```
mprahl, line 71:
```suggestion
```

droctothorpe:
This is really impressive, @zazulam . Will provide an in-depth review but just want to start with that 😂.

backend/src/v2/component/launcher_v2.go
```diff
@@ -562,51 +562,53 @@ func downloadArtifacts(ctx context.Context, executorInput *pipelinespec.Executor
...
+		for _, artifact := range artifactList.Artifacts {
```
droctothorpe, line 565: Working on the same thing in a different [place](https://github.com/kubeflow/pipelines/pull/11762/files#diff-8ad39c78b8d631f0ca74f79f5090a722aeef939a511aa19b116e361c1a2a01c2R548), heh.

backend/src/v2/driver/driver.go
```diff
@@ -1694,10 +1695,17 @@ func resolveUpstreamParameters(cfg resolveUpstreamOutputsConfig) (*structpb.Valu
 		return nil, cfg.err(fmt.Errorf("output parameter key is empty"))
 	}
 
+	isParallelForDAG := cfg.dag.Execution.GetExecution().GetCustomProperties()["iteration_index"] != nil
```
droctothorpe, line 1698: It might be helpful to include a comment here explaining what this block is doing. In particular, it might be helpful to provide some context on what `iteration_index`, how parallelFor's use it, and why we need to iterate using it.

backend/src/v2/metadata/client.go
```diff
@@ -854,6 +855,15 @@ func (c *Client) GetExecutionsInDAG(ctx context.Context, dag *DAG, pipeline *Pip
 				// When retrieving executions without the parentDAGFilter, the rootDAG execution is supplied but does not have an associated TaskName nor is the parentDagID set, therefore we won't include it in the executionsMap.
 				continue
 			}
+			// Handle for parallelFor subdags & their tasks that consume the values from the iterator.
+			if e.GetCustomProperties()[keyIterationIndex] != nil {
+				taskName = fmt.Sprintf("%s_idx_%d", taskName, e.GetCustomProperties()[keyIterationIndex].GetIntValue())
+
+			} else if dag.Execution.GetExecution().GetCustomProperties()[keyIterationIndex] != nil {
+				// Handle for tasks within a parallelFor subdag that do not consume the values from the iterator as input but rather the output of a task that does.
+				taskName = fmt.Sprintf("%s_idx_%d", taskName, dag.Execution.GetExecution().GetCustomProperties()[keyIterationIndex].GetIntValue())
```
droctothorpe, line 864: To be honest, I don't grok this! It might be helpful to expand the comments a bit.

backend/src/v2/driver/driver.go
zazulam: I had added this parameter before I learned that only ParallelFor DAGs have the custom property `iteration_count` that I can rely on when iterating through the potential tree of DAGs. Will remove it.

backend/src/v2/metadata/client.go
zazulam: From my exploration, there are 2 types of tasks that have varying attributes in a parallelFor DAG, 1 that consume the values that are being looped over and one that consumes the output of another task that does.

example:
```python
...
with dsl.ParallelFor(items=["id_1","id_2","id_3"]) as model_id:
        task_1 = comp_1(id=model_id)
        task_2 = comp_2(result=task_1.output)
```

`task_1` would have the iteration_index custom property set on the Execution associated with it because it is consuming values from the iterator used for the loop, `model_id`. `task_2` does not have that property set on its Execution because it does not consume `model_id`. We could probably update this to always just check the the dag's execution properties, I just tested it in this order so I cannot confirm atm if there may exist a case where only checking the dag can suffice.

As for the actual logic at play, we are attempting to store the "would-be" collisions in the map of execution as such:
```
{
    "task1_idx_0" : &Execution{},
    "task1_idx_1" : &Execution{},
    "task1_idx_2" : &Execution{},
    "task2_idx_0" : &Execution{},
    "task2_idx_1" : &Execution{},
    "task2_idx_2" : &Execution{},
}
```

backend/src/v2/driver/driver.go
zazulam: For the parallelFor there's a log for the attempting the fan-in which I *think* should be fine. If not, can update the log statement.

backend/src/v2/metadata/client.go
zazulam: Not sure if I mentioned this somewhere else, but each iteration of the ParallelFor is its own DAG. So with the pipeline above there's 5 DAG executions 😅 
- The root pipeline the loop is in
- root of ParallelFor loop
- 1st iteration 
- 2nd iteration 
- 3rd iteration

backend/src/v2/driver/driver.go
```diff
@@ -1777,6 +1779,34 @@ func resolveUpstreamParameters(cfg resolveUpstreamOutputsConfig) (*structpb.Valu
 			if subTaskName == "" {
 				return nil, cfg.err(fmt.Errorf("producer_subtask not in outputParams"))
 			}
+			// If the sub-task is a ParallelFor DAG, we need to gather all i.e. dsl.Collected support
+			if currentTask.GetExecution().GetCustomProperties()["iteration_count"] != nil {
+				glog.V(4).Infof("currentTask, %v, is a ParallelFor DAG. Attempting to gather all producer_subtask iterations of %w", currentTask.TaskName(), subTaskName)
+
+				outputs_list := make([]*structpb.Value, 0)
```
zazulam, line 1786: :camel:

backend/src/v2/driver/driver.go
```diff
@@ -1777,6 +1779,34 @@ func resolveUpstreamParameters(cfg resolveUpstreamOutputsConfig) (*structpb.Valu
 			if subTaskName == "" {
 				return nil, cfg.err(fmt.Errorf("producer_subtask not in outputParams"))
 			}
+			// If the sub-task is a ParallelFor DAG, we need to gather all i.e. dsl.Collected support
```
zazulam, line 1782: clarify currentTask in to resolve outputs is within a parallelFor DAG

backend/src/v2/driver/driver.go
```diff
@@ -1694,6 +1688,14 @@ func resolveUpstreamParameters(cfg resolveUpstreamOutputsConfig) (*structpb.Valu
 		return nil, cfg.err(fmt.Errorf("output parameter key is empty"))
 	}
 
+	// For the scenario where 2 tasks are defined within a ParallelFor and 1 receives the output of the other we must ensure that the downstream task resolves the approriate output of the iteration it is in.
+	isParallelForDAG := cfg.dag.Execution.GetExecution().GetCustomProperties()["iteration_index"] != nil
+	if isParallelForDAG {
+		// This is needed to support tasks within a ParallelFor Loop that do not leverage the iterator values but still consume outputs from ones that do
```
zazulam, line 1694: UPdate to clarify this case is used specifically for tasks within the parallelFor loop that consumes the output of earlier tasks in the same loop.

backend/src/v2/driver/driver.go
```diff
@@ -1850,15 +1892,51 @@ func resolveUpstreamArtifacts(cfg resolveUpstreamOutputsConfig) (*pipelinespec.A
...
 			}
-			// If the sub-task is a DAG, reassign currentTask and run
-			// through the loop again.
-			currentTask = tasks[subTaskName]
-			// }
+
+			if currentTask.GetExecution().GetCustomProperties()["iteration_count"] != nil {
+				glog.V(4).Infof("currentTask, %v, is a ParallelFor DAG. Attempting to gather all producer_subtask iterations of %w", currentTask.TaskName(), subTaskName)
+
+				outputs_list := make([]*pipelinespec.RuntimeArtifact, 0)
```
zazulam, line 1899: :camel: rename as well

backend/src/v2/driver/driver.go
zazulam: I updated it to just use the last one, I think when we first added this functionality we were unaware if there is ever a case where there is more that 1 element in the artifactSelector list so we had just opted for the last one since there was precursor.

samples/v2/collected_artifacts.py
zazulam: Done

droctothorpe (approved):
LGTM. Amazing work, @zazulam!

zazulam:
cc: @chensun @HumairAK

backend/src/v2/driver/driver.go
```diff
@@ -1933,23 +1976,58 @@ func resolveUpstreamArtifacts(cfg resolveUpstreamOutputsConfig) (*pipelinespec.A
 			// Adding support for multiple output artifacts
 			var subTaskName string
 			artifactSelectors := outputArtifacts[outputArtifactKey].GetArtifactSelectors()
+			artifactSelectorValue := artifactSelectors[len(artifactSelectors)-1]
```
mprahl, line 1979: Could you add a comment explaining why only the last one is used?

mprahl (changes requested):
A couple small comments but it looks like my other comments were addressed. Let me know when it's ready for review again!

backend/src/v2/driver/driver.go
zazulam: Since we haven't come across pipelines where it did not work I'll add some more testing of different pipeline loop setups to verify that this will always be the case before adding that comment.

backend/src/v2/driver/driver.go
```diff
@@ -627,18 +627,19 @@ func addModelcarsToPodSpec(
...
+			// be set on inputArtifact.
+			if !strings.HasPrefix(inputArtifact.Uri, "oci://") {
+				continue
+			}
 
-		// This should ideally verify that this is also a model input artifact, but this metadata doesn't seem to
-		// be set on inputArtifact.
-		if !strings.HasPrefix(inputArtifact.Uri, "oci://") {
-			continue
+			modelcarArtifacts[name] = inputArtifact
```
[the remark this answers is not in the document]
zazulam: Nice catch, thoughts on modifying the name used to store the artifact in the map to include the index used during the looping?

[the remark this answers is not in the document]
mprahl: @zazulam sorry, my comment in Markdown didn't render as I intended, but yes, I think keys in the format of `<name>-<index>` would be ok.

backend/src/v2/driver/driver.go
```diff
@@ -1934,22 +1994,54 @@ func resolveUpstreamArtifacts(cfg resolveUpstreamOutputsConfig) (*pipelinespec.A
...
+			resolveCollection, iterations := getParallelForIterationCount(currentTask, parentDAG)
+			glog.V(4).Infof("resolveCollection: %v", resolveCollection)
+			glog.V(4).Infof("iterations: %v", iterations)
+			if resolveCollection {
+				starterTaskQueue := make([]string, 0)
+				for i := range iterations {
+					starterTaskName := metadata.UpdateTaskNameWithDagID(currentTask.TaskName(), currentTask.GetID())
+					starterTaskName = metadata.GetParallelForTaskName(starterTaskName, i)
+					starterTaskQueue = append(starterTaskQueue, starterTaskName)
+				}
```
zazulam, line 2021: As mentioned above for Parameters, the detection and setup of the collecting Artifacts can be moved to the same function.

backend/src/v2/driver/driver.go
```diff
@@ -1974,7 +2066,254 @@ func resolveUpstreamArtifacts(cfg resolveUpstreamOutputsConfig) (*pipelinespec.A
...
+								break
+							}
+						}
+					}
+					if !successfulOneOfTask {
+						return "", "", fmt.Errorf("processing OneOf: No successful task found")
+					}
+				}
+			}
+		}
```
zazulam, line 2286: For the logic in `resolveUpstreamParameters`, we can swap to using this function.

backend/src/v2/metadata/client.go
```diff
@@ -846,14 +855,25 @@ func (c *Client) GetExecutionsInDAG(ctx context.Context, dag *DAG, pipeline *Pip
 		glog.V(4).Infof("execs: %v", execs)
 		for _, e := range execs {
 			execution := &Execution{execution: e}
-			taskName := execution.TaskName()
+			taskName := UpdateTaskNameWithDagID(execution.TaskName(), parentDAGID)
+			glog.V(4).Infof("taskName: %s", taskName)
+			glog.V(4).Infof("execution: %s", execution)
```
zazulam, line 860: Every task now stored in the task map will use their parent dag id and if there are iterations then it will be accounted for in the downstream input resolving functions.

samples/v2/collected_artifacts.py
```diff
@@ -0,0 +1,138 @@
...
+        
+        read_models_op = read_models(
+            models=dsl.Collected(read_single_dataset_generate_model_op.outputs['results']))
+        read_models_op.set_caching_options(False)
+
+        read_datasets_op = read_datasets(
+            data=dsl.Collected(single_dag_op.output))
+        read_datasets_op.set_caching_options(False)
+    
+    return dsl.Collected(read_single_dataset_generate_model_op.outputs["results"])
```
zazulam, line 119: This was a nightmare to collect, but hey, it was fun 😃

zazulam (commented):
Commented on some areas after my own look through, let me know any thoughts!

zazulam:
As for the artifact loop that is apart of the sample tests that isn't enabled yet, here's a recording of the UI from my most recent successful run I have.

https://github.com/user-attachments/assets/dfe4d291-70a5-4b29-ab2b-5b0baa4cd219

backend/src/v2/driver/driver.go
zazulam: just added

backend/src/v2/driver/driver.go
```diff
@@ -1829,9 +1837,39 @@ func resolveUpstreamParameters(cfg resolveUpstreamOutputsConfig) (*structpb.Valu
...
+				// after, both it and the subTaskName exist within a struct that
+				// is mapped with the original outputParameterKey. Thus, we need
+				// only 1 of the 2 values for downstream collection of
+				// parameters
+
+				for i := range iterations {
+					starterTaskName := metadata.UpdateTaskNameWithDagID(currentTask.TaskName(), currentTask.GetID())
+					starterTaskName = metadata.GetParallelForTaskName(starterTaskName, i)
+					starterTaskQueue = append(starterTaskQueue, starterTaskName)
+				}
```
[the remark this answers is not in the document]
zazulam: done

backend/src/v2/driver/driver.go
```diff
@@ -1934,22 +1994,54 @@ func resolveUpstreamArtifacts(cfg resolveUpstreamOutputsConfig) (*pipelinespec.A
...
 
-			for _, v := range artifactSelectors {
-				glog.V(4).Infof("v: %v", v)
-				glog.V(4).Infof("v.ProducerSubtask: %v", v.ProducerSubtask)
-				glog.V(4).Infof("v.OutputArtifactKey: %v", v.OutputArtifactKey)
-				subTaskName = v.ProducerSubtask
-				outputArtifactKey = v.OutputArtifactKey
+			glog.Infof("Logging all tasks within TaskMap")
+			for key, task := range tasks {
+				glog.Infof("task key: %v  | task id: %v", key, task.GetID())
```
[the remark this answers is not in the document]
zazulam: done

backend/src/v2/driver/driver.go
```diff
@@ -1845,11 +1883,12 @@ func resolveUpstreamParameters(cfg resolveUpstreamOutputsConfig) (*structpb.Valu
...
 						subTaskName = paramSelector.GetProducerSubtask()
-						glog.V(4).Infof("subTaskName from paramSelector: %v", subTaskName)
+						subTaskNameWithDagId := metadata.UpdateTaskNameWithDagID(subTaskName, currentTask.GetID())
+						glog.V(4).Infof("subTaskName with Dag ID from paramSelector: %v", subTaskNameWithDagId)
 						glog.V(4).Infof("outputParameterKey from paramSelector: %v", paramSelector.GetOutputParameterKey())
-						if subTask, ok := tasks[subTaskName]; ok {
+						if subTask, ok := tasks[subTaskNameWithDagId]; ok {
 							subTaskState := subTask.GetExecution().LastKnownState.String()
-							glog.V(4).Infof("subTask: %w , subTaskState: %v", subTaskName, subTaskState)
+							glog.V(4).Infof("subTask: %w , subTaskState: %v", subTaskNameWithDagId, subTaskState)
```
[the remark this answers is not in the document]
zazulam: `resolveUpstreamParameters` now uses `InspectProducerTask`

backend/src/v2/metadata/client.go
```diff
@@ -281,6 +281,14 @@ func (e *Execution) FingerPrint() string {
 	return e.execution.GetCustomProperties()[keyCacheFingerPrint].GetStringValue()
 }
 
+func UpdateTaskNameWithDagID(taskName string, dagID int64) string {
```
droctothorpe, line 284: nit: I think a docstring that explains **why** this function exists might be nice.

backend/src/v2/metadata/client.go
```diff
@@ -281,6 +281,14 @@ func (e *Execution) FingerPrint() string {
...
+	return fmt.Sprintf("%s_%d", taskName, dagID)
+}
+
+func GetParallelForTaskName(taskName string, iterationIndex int64) string {
```
droctothorpe, line 288: Same as above.

backend/src/v2/metadata/client.go
```diff
@@ -820,9 +828,10 @@ func (c *Client) GetExecutionsInDAG(ctx context.Context, dag *DAG, pipeline *Pip
 	// Documentation on query syntax:
 	// https://github.com/google/ml-metadata/blob/839c3501a195d340d2855b6ffdb2c4b0b49862c9/ml_metadata/proto/metadata_store.proto#L831
 	// If filter is set to true, the MLMD call will only grab executions for the current DAG, else it would grab all the execution for the context which includes sub-DAGs.
+	parentDAGID := dag.Execution.GetID()
```
droctothorpe, line 831: It may be helpful to clarify what this is the parent of.

backend/src/v2/metadata/client.go
droctothorpe: Thanks for clarifying!

backend/src/v2/metadata/client.go
```diff
@@ -846,14 +855,25 @@ func (c *Client) GetExecutionsInDAG(ctx context.Context, dag *DAG, pipeline *Pip
 		glog.V(4).Infof("execs: %v", execs)
 		for _, e := range execs {
 			execution := &Execution{execution: e}
-			taskName := execution.TaskName()
+			taskName := UpdateTaskNameWithDagID(execution.TaskName(), parentDAGID)
```
droctothorpe, line 858: It may be helpful to add a comment here explaining why this is necessary, e.g.:

```
// Sometimes components in nested DAGs have identical task names. We update the task name to include the DAG ID to avoid key collisions in the executions map.
```

backend/src/v2/metadata/client.go
```diff
@@ -846,14 +855,25 @@ func (c *Client) GetExecutionsInDAG(ctx context.Context, dag *DAG, pipeline *Pip
...
+			glog.V(4).Infof("taskName: %s", taskName)
```
droctothorpe, line 859: Do you think it might be helpful to also log the `taskName` before the modification? e.g. `taskName before DAG injection:` and `taskName after DAG injection:`.

backend/src/v2/metadata/client.go
```diff
@@ -846,14 +855,25 @@ func (c *Client) GetExecutionsInDAG(ctx context.Context, dag *DAG, pipeline *Pip
...
 			if taskName == "" {
 				if e.GetCustomProperties()[keyParentDagID] != nil {
 					return nil, fmt.Errorf("empty task name for execution ID: %v", execution.GetID())
 				}
 				// When retrieving executions without the parentDAGFilter, the rootDAG execution is supplied but does not have an associated TaskName nor is the parentDagID set, therefore we won't include it in the executionsMap.
 				continue
 			}
+			// Handle for parallelFor subdags & their tasks that consume the values from the iterator.
+			if e.GetCustomProperties()[keyIterationIndex] != nil {
+				taskName = GetParallelForTaskName(taskName, e.GetCustomProperties()[keyIterationIndex].GetIntValue())
```
droctothorpe, line 870: Similar to [this comment](https://github.com/kubeflow/pipelines/pull/11725/files#r2064178331), I think it would be helpful to explain the **why** behind this, specifically, what problem it solves, i.e. redundant keys in the execution map.

droctothorpe: e.g.:

In parallelized tasks, task names are not unique so we run into key collision errors in the execution map that are not resolved by `UpdateTaskNameWithDagID`. In parallelized tasks, each iteration has a unique `keyIterationIndex` in its custom properties. We therefore update the task name to include this index to ensure there are no key collisions in the executions map.

droctothorpe: Random edge case that just occurred to me (I don't recall if there's validation to protect against task names that end with ints):

What happens if you have a pipeline with a parallelized "foo" task and, separately, outside of the parallelization, but within the same pipeline, a `foo_idx_1` or `foo_idx_2` task? I feel like a monster for even contemplating this possibility.

backend/src/v2/metadata/client.go
```diff
@@ -846,14 +855,25 @@ func (c *Client) GetExecutionsInDAG(ctx context.Context, dag *DAG, pipeline *Pip
...
+
+			} else if dag.Execution.GetExecution().GetCustomProperties()[keyIterationIndex] != nil {
+				// Handle for tasks within a parallelFor subdag that do not consume the values from the iterator as input but rather the output of a task that does.
```
droctothorpe, line 873: Appreciate the comment. The level of nesting and complexity here is **wild** (and unavoidable).

samples/v2/collected_artifacts.py
```diff
@@ -0,0 +1,138 @@
+from typing import List
```
droctothorpe, line 1: This sample is complex enough that a comment that summarizes / justifies it might be worthwhile. How many components does it have total? This is definitely above and beyond but an ASCII diagram might be nice.

backend/src/v2/driver/driver.go
```diff
@@ -1927,61 +1927,83 @@ func resolveUpstreamParameters(cfg resolveUpstreamOutputsConfig) (*structpb.Valu
...
+	glog.V(4).Info("taskOutput: ", taskOutput)
+	producerTaskName := taskOutput.GetProducerTask()
 	if taskOutput.GetProducerTask() == "" {
 		return nil, cfg.err(fmt.Errorf("producer task is empty"))
 	}
 	if taskOutput.GetOutputArtifactKey() == "" {
 		cfg.err(fmt.Errorf("output artifact key is empty"))
 	}
+	producerTaskName = metadata.UpdateTaskNameWithDagID(producerTaskName, cfg.dag.Execution.GetID())
+	// The main difference between the root ParallelFor DAG and its iteration
```
droctothorpe, line 1939: Hol' up. Each iteration has its own DAG? Why?

backend/src/v2/driver/driver.go
```diff
@@ -1927,61 +1927,83 @@ func resolveUpstreamParameters(cfg resolveUpstreamOutputsConfig) (*structpb.Valu
...
+	glog.V(4).Infof("tasks: %#v", tasks)
 	currentTask := producer
 	outputArtifactKey := taskOutput.GetOutputArtifactKey()
 
-	// Continue looping until we reach a sub-task that is NOT a DAG.
+	// Continue looping until we reach a sub-task that is either a ParallelFor task or a Container task.
 	for {
 		glog.V(4).Info("currentTask: ", currentTask.TaskName())
 		// If the current task is a DAG:
 		if *currentTask.GetExecution().Type == "system.DAGExecution" {
```
droctothorpe, line 1971: Could there be cases in this conditional where it's a DAG but not a ParallelFor?

backend/src/v2/driver/driver.go
```diff
@@ -2006,7 +2028,253 @@ func resolveUpstreamArtifacts(cfg resolveUpstreamOutputsConfig) (*pipelinespec.A
 	}
 }
 
-// provisionOuutputs prepares output references that will get saved to MLMD.
+func isCollection(cfg resolveUpstreamOutputsConfig, currentTask *metadata.Execution) ([]string, error) {
```
droctothorpe, line 2031: A docstring for this would be helpful.

backend/src/v2/driver/driver.go
```diff
@@ -2006,7 +2028,253 @@ func resolveUpstreamArtifacts(cfg resolveUpstreamOutputsConfig) (*pipelinespec.A
...
+			starterTaskName = metadata.GetParallelForTaskName(starterTaskName, i)
+			starterTaskQueue = append(starterTaskQueue, starterTaskName)
+		}
+	}
+	return starterTaskQueue, nil
+}
+
+// In order to properly perform artifact/parameter collection across a DAG/tree
+// we perform a breadth first search traversal.
+func CollectInputs(cfg resolveUpstreamOutputsConfig, starterTaskQueue []string, tasks map[string]*metadata.Execution, outputKey string, isArtifact bool) (*structpb.Value, *pipelinespec.ArtifactList, error) {
```
droctothorpe, line 2053:
```suggestion
func CollectInputs(
        cfg resolveUpstreamOutputsConfig,
        starterTaskQueue []string,
        tasks map[string]*metadata.Execution,
        outputKey string,
        isArtifact bool,
) (*structpb.Value, *pipelinespec.ArtifactList, error) {
```

Improved readability for a complex signature.

backend/src/v2/driver/driver.go
```diff
@@ -2006,7 +2028,253 @@ func resolveUpstreamArtifacts(cfg resolveUpstreamOutputsConfig) (*pipelinespec.A
...
+		for i := range iterations {
+			starterTaskName := metadata.UpdateTaskNameWithDagID(currentTask.TaskName(), currentTask.GetID())
+			starterTaskName = metadata.GetParallelForTaskName(starterTaskName, i)
+			starterTaskQueue = append(starterTaskQueue, starterTaskName)
+		}
+	}
+	return starterTaskQueue, nil
+}
+
+// In order to properly perform artifact/parameter collection across a DAG/tree
```
droctothorpe, line 2051: How did v1 handle this? Just relied on AWF to the do the heavy lifting? I think that maybe, given the complexity of this function, it might make sense to expand the docstring and maybe even add parameter descriptions.

backend/src/v2/driver/driver.go
```diff
@@ -2006,7 +2028,253 @@ func resolveUpstreamArtifacts(cfg resolveUpstreamOutputsConfig) (*pipelinespec.A
...
+	resolveTaskNameQueue := starterTaskQueue
+	var currentTask *metadata.Execution
+	var workingSubTaskName string
+	workingOutputKey := outputKey
+	tempWorkingOutputKey := outputKey
+	previousTaskName := tasks[resolveTaskNameQueue[0]].TaskName()
+
+	for key, task := range tasks {
+		glog.Infof("task: %v  || task: %v", key, task)
+	}
```
droctothorpe, line 2067: Is this still necessary? You sure you want it to be `Infof`?

backend/src/v2/driver/driver.go
zazulam: That seems to be the way that loops were handled with the v2 setup and where we first experienced the collisions within `GetExecutionsInDag`, because the name of each of the sub dags (iterations) created from the parallelFor loop dag all had the same name within the ExecutionProperty table.
image

backend/src/v2/driver/driver.go
zazulam: Nope, will remove

backend/src/v2/metadata/client.go
mprahl, line 284: This function name is misleading in my opinion as it makes it sound like something in MLMD is getting updated. How about just removing `Update` or replace with `Get`.

backend/src/v2/driver/driver.go
```diff
@@ -2006,7 +2028,253 @@ func resolveUpstreamArtifacts(cfg resolveUpstreamOutputsConfig) (*pipelinespec.A
...
+	glog.V(4).Infof("currentTask is a ParallelFor DAG. Attempting to gather all nested producer_subtasks")
+	// Set some helpers for the BFS
+	parallelForParameterList := make([]*structpb.Value, 0)
+	parallelForArtifactList := make([]*pipelinespec.RuntimeArtifact, 0)
+	resolveTaskNameQueue := starterTaskQueue
+	var currentTask *metadata.Execution
+	var workingSubTaskName string
+	workingOutputKey := outputKey
+	tempWorkingOutputKey := outputKey
+	previousTaskName := tasks[resolveTaskNameQueue[0]].TaskName()
```
mprahl, line 2063: This should likely check the length of `resolveTaskNameQueue` before accessing its index.

backend/src/v2/driver/driver.go
```diff
@@ -2006,7 +2028,253 @@ func resolveUpstreamArtifacts(cfg resolveUpstreamOutputsConfig) (*pipelinespec.A
...
+				// output set. In this case we will attempt to grab the first
+				// successful task output.
+				paramSelectors := tempOutputParametersMap[tempOutputKey].GetValueFromOneof().GetParameterSelectors()
+				glog.V(4).Infof("paramSelectors: %v", paramSelectors)
+				// Since we have the tasks map, we can iterate through the
+				// parameterSelectors if the ProducerSubTask is not present in
+				// the task map and then assign the new OutputParameterKey only
+				// if it exists.
+				successfulOneOfTask := false
+				for !successfulOneOfTask {
```
mprahl, line 2222: Is this outer loop needed? It seems like the inner loop is what matters.

backend/src/v2/driver/driver.go
```diff
@@ -2006,7 +2028,253 @@ func resolveUpstreamArtifacts(cfg resolveUpstreamOutputsConfig) (*pipelinespec.A
...
+							glog.V(4).Infof("subTask: %w , subTaskState: %v", updatedSubTaskName, subTaskState)
+							if subTaskState == "CACHED" || subTaskState == "COMPLETE" {
+								tempOutputKey = paramSelector.GetOutputParameterKey()
+								successfulOneOfTask = true
+								break
+							}
+						}
+					}
+					if !successfulOneOfTask {
+						return "", "", fmt.Errorf("processing OneOf: No successful task found")
```
mprahl, line 2240: If there isn't a successful task, does Argo Workflows actually get to executing the one of consumer? I'm thinking this would be hard otherwise for the user to trace down why their pipeline failed since this error is in the driver.

backend/src/v2/driver/driver.go
```diff
@@ -627,18 +628,20 @@ func addModelcarsToPodSpec(
 		if len(artifactList.Artifacts) == 0 {
 			continue
 		}
+		// Following the convention of downloadArtifacts in the launcher to look at the entire list.
+		for index, artifact := range artifactList.Artifacts {
```
mprahl, line 632: This also needs to be updated in `stopWaitingArtifacts`.

backend/src/v2/driver/driver.go
```diff
@@ -1927,61 +1927,83 @@ func resolveUpstreamParameters(cfg resolveUpstreamOutputsConfig) (*structpb.Valu
...
 		cfg.err(fmt.Errorf("output artifact key is empty"))
 	}
+	producerTaskName = metadata.UpdateTaskNameWithDagID(producerTaskName, cfg.dag.Execution.GetID())
+	// The main difference between the root ParallelFor DAG and its iteration
+	// DAGs is that the root contains the customer property "iteration_count"
+	// while the iterations contain "iteration_index". We can use this to
+	// determine if we are in a ParallelFor DAG or not. The iteration DAGs will
+	// contain the "iteration_index" which is used to resolve the correct output
+	// artifact for the downstream task within the iteration. ParallelFor
+	// iterations are DAGs themselves, we can verify if we are in a iteration by
```
mprahl, line 1945:
```suggestion
	// iterations are DAGs themselves, we can verify if we are in an iteration by
```

backend/src/v2/driver/driver.go
droctothorpe: I think it's worth adding parameter descriptions to the doc string given the complexity of what this function is doing and the inputs that it receives.

droctothorpe: Consider using named return values as well to improve clarity.

backend/src/v2/driver/driver.go
droctothorpe, line 2063: Consider adding comments clarifying the why behind these variables.

backend/src/v2/driver/driver.go
```diff
@@ -2006,7 +2028,253 @@ func resolveUpstreamArtifacts(cfg resolveUpstreamOutputsConfig) (*pipelinespec.A
...
+	var workingSubTaskName string
+	workingOutputKey := outputKey
+	tempWorkingOutputKey := outputKey
+	previousTaskName := tasks[resolveTaskNameQueue[0]].TaskName()
+
+	for key, task := range tasks {
+		glog.Infof("task: %v  || task: %v", key, task)
+	}
+	for len(resolveTaskNameQueue) > 0 {
+		// The starterQueue contains the first set of child DAGs from the
```
droctothorpe, line 2069: Make sure comment var name aligns with updated var name.

backend/src/v2/driver/driver.go
```diff
@@ -2006,7 +2028,253 @@ func resolveUpstreamArtifacts(cfg resolveUpstreamOutputsConfig) (*pipelinespec.A
...
+}
+
+// In order to properly perform artifact/parameter collection across a DAG/tree
+// we perform a breadth first search traversal.
+func CollectInputs(cfg resolveUpstreamOutputsConfig, starterTaskQueue []string, tasks map[string]*metadata.Execution, outputKey string, isArtifact bool) (*structpb.Value, *pipelinespec.ArtifactList, error) {
+	glog.V(4).Infof("currentTask is a ParallelFor DAG. Attempting to gather all nested producer_subtasks")
+	// Set some helpers for the BFS
+	parallelForParameterList := make([]*structpb.Value, 0)
+	parallelForArtifactList := make([]*pipelinespec.RuntimeArtifact, 0)
+	resolveTaskNameQueue := starterTaskQueue
```
droctothorpe, line 2058: Considering naming this `q` or `traversalQueue` or `BFSQueue` or something like that to make it clear that this is the core structure in the algorithm.

backend/src/v2/driver/driver.go
```diff
@@ -2006,7 +2028,253 @@ func resolveUpstreamArtifacts(cfg resolveUpstreamOutputsConfig) (*pipelinespec.A
...
+	outputArtifactList := &pipelinespec.ArtifactList{
+		Artifacts: parallelForArtifactList,
+	}
+	glog.V(4).Infof("outputParameterList: %v", outputParameterList)
+	glog.V(4).Infof("outputArtifactList: %v", outputArtifactList)
+	return outputParameterList, outputArtifactList, nil
+}
+
+// Helper for determining whether to update the ProducerSubTask / Output[Artifact|Parameter]Key
+func InspectProducerTask(parentTask *metadata.Execution, tasks map[string]*metadata.Execution, subTaskName string, outputKey string, isArtifact bool) (string, string, error) {
```
droctothorpe, line 2173: I recommend putting these parameters on separate lines to reduce the line length of the function signature. It may also help add a more elaborate doc string and maybe even named returns in the signature.

backend/src/v2/driver/driver.go
```diff
@@ -1861,45 +1869,36 @@ func resolveUpstreamParameters(cfg resolveUpstreamOutputsConfig) (*structpb.Valu
 
 			glog.V(4).Infof("Deserialized dagOutputParametersMap: %v", dagOutputParametersMap)
 
+			// For this section, if the currentTask we are looking for is within
+			// a ParallelFor DAG, this means the actual task that produced the
+			// output we need has multiple iterations so we have to gather all
+			// them and fan them in by collecting them into a list i.e.
+			// kfp.dsl.Collected support.
+			starterTaskQueue, err := isCollection(cfg, currentTask)
```
mprahl, line 1877: How about something like this? It's not actually a "queue" until `CollectInputs` is called.
```suggestion
			parallelForTasks, err := isCollection(cfg, currentTask)
```

backend/src/v2/driver/driver.go
```diff
@@ -2006,7 +2028,253 @@ func resolveUpstreamArtifacts(cfg resolveUpstreamOutputsConfig) (*pipelinespec.A
...
+				_, outputParameters, err := currentTask.GetParameters()
+				glog.V(4).Infof("outputParameters: %v", outputParameters)
+				if err != nil {
+					return nil, nil, cfg.err(err)
+				}
+				parallelForParameterList = append(parallelForParameterList, outputParameters[workingOutputKey])
+			}
+			continue
+
+			// Base case 2: currentTask is a DAGExecution within a loop but is NOT a ParallelFor
```
droctothorpe, line 2135: nit: correct indentation

backend/src/v2/driver/driver.go
```diff
@@ -2006,7 +2028,253 @@ func resolveUpstreamArtifacts(cfg resolveUpstreamOutputsConfig) (*pipelinespec.A
...
+			var tempOutputArtifacts map[string]*pipelinespec.DagOutputsSpec_DagOutputArtifactSpec
+			err := json.Unmarshal([]byte(producerTaskValue.GetStringValue()), &tempOutputArtifacts)
+			if err != nil {
+				return "", "", err
+			}
+			glog.V(4).Infof("tempOutputsArtifacts: %v", tempOutputArtifacts)
+			glog.V(4).Infof("outputArtifactKey: %v", outputKey)
+			tempSelectors := tempOutputArtifacts[outputKey].GetArtifactSelectors()
+			if len(tempSelectors) > 0 {
+				tempSubTaskName = tempSelectors[len(tempSelectors)-1].ProducerSubtask
```
mprahl, line 2188: Why is only the last selector used?

backend/src/v2/driver/driver.go
```diff
@@ -2006,7 +2028,253 @@ func resolveUpstreamArtifacts(cfg resolveUpstreamOutputsConfig) (*pipelinespec.A
...
+				glog.V(4).Infof("outputParameters: %v", outputParameters)
+				if err != nil {
+					return nil, nil, cfg.err(err)
+				}
+				parallelForParameterList = append(parallelForParameterList, outputParameters[workingOutputKey])
+			}
+			continue
+
+			// Base case 2: currentTask is a DAGExecution within a loop but is NOT a ParallelFor
+			// Head DAG
```
droctothorpe, line 2136: It might be helpful to expand on the distinction between parallel head DAGs vs parallel iteration DAGs if that hasn't already been clarified elsewhere.

backend/src/v2/driver/driver.go
```diff
@@ -2006,7 +2028,253 @@ func resolveUpstreamArtifacts(cfg resolveUpstreamOutputsConfig) (*pipelinespec.A
...
+				// handle for parallel iteration dag
+				tempSubTaskName = metadata.GetParallelForTaskName(tempSubTaskName, iterationIndex.GetIntValue())
+				glog.V(4).Infof("subTaskIterationName: %v", tempSubTaskName)
+			}
+			glog.V(4).Infof("tempSubTaskName: %v", tempSubTaskName)
+			resolveTaskNameQueue = append(resolveTaskNameQueue, tempSubTaskName)
+			continue
+		}
+
+		for i := range iterations.GetIntValue() {
```
droctothorpe, line 2149: Consider adding an explicit conditional for this or a comment that explains that if the if / else if are not met, then we're definitely dealing with a parallel iteration.

backend/src/v2/driver/driver.go
mprahl, line 2173: How about calling it `GetOutputProducerSubtask` or something? The word inspect was not intuitive to me that it'd return values other than an error.

backend/src/v2/driver/driver.go
```diff
@@ -2006,7 +2028,253 @@ func resolveUpstreamArtifacts(cfg resolveUpstreamOutputsConfig) (*pipelinespec.A
...
+		Artifacts: parallelForArtifactList,
+	}
+	glog.V(4).Infof("outputParameterList: %v", outputParameterList)
+	glog.V(4).Infof("outputArtifactList: %v", outputArtifactList)
+	return outputParameterList, outputArtifactList, nil
+}
+
+// Helper for determining whether to update the ProducerSubTask / Output[Artifact|Parameter]Key
+func InspectProducerTask(parentTask *metadata.Execution, tasks map[string]*metadata.Execution, subTaskName string, outputKey string, isArtifact bool) (string, string, error) {
+	tempSubTaskName := subTaskName
```
mprahl, line 2174: Does it have to be called temp? How about `producerSubTaskName`?

backend/src/v2/driver/driver.go
```diff
@@ -2006,7 +2028,253 @@ func resolveUpstreamArtifacts(cfg resolveUpstreamOutputsConfig) (*pipelinespec.A
...
+				successfulOneOfTask := false
+				for !successfulOneOfTask {
+					for _, paramSelector := range paramSelectors {
+						tempSubTaskName = paramSelector.GetProducerSubtask()
+						// Used just for retrieval since we lookup the task in the map
+						updatedSubTaskName := metadata.UpdateTaskNameWithDagID(tempSubTaskName, parentTask.GetID())
+						glog.V(4).Infof("subTaskName with Dag ID from paramSelector: %v", updatedSubTaskName)
+						glog.V(4).Infof("outputParameterKey from paramSelector: %v", paramSelector.GetOutputParameterKey())
+						if subTask, ok := tasks[updatedSubTaskName]; ok {
+							subTaskState := subTask.GetExecution().LastKnownState.String()
```
mprahl, line 2230:
```suggestion
							subTaskState := subTask.GetExecution().GetLastKnownState().String()
```

backend/src/v2/driver/driver.go
mprahl: Or just call it `tasksToResolve` with a comment explaining the slice can grow during the execution of the function as new tasks are found.

backend/src/v2/driver/driver.go
```diff
@@ -2006,7 +2028,253 @@ func resolveUpstreamArtifacts(cfg resolveUpstreamOutputsConfig) (*pipelinespec.A
...
+		// in question.
+
+		// An issue arises if we update the outputParameterKey but there exists
+		// multiple iterations of the same task and we haven't fully parsed all
+		// iterations. We will encounter a scenario where we will attempt to
+		// extract fields from the struct with the wrong key. Hence, the
+		// condition below. NOTE: This is only an issue for Parameter resolution
+		// and does not interfere with Artifact resolution
+		if currentTask.TaskName() == previousTaskName {
+			workingOutputKey = tempWorkingOutputKey
```
mprahl, line 2091: Should `tempWorkingOutputKey` just be called `previousWorkingOutputKey` or something?

backend/src/v2/driver/driver.go
```diff
@@ -2006,7 +2028,253 @@ func resolveUpstreamArtifacts(cfg resolveUpstreamOutputsConfig) (*pipelinespec.A
...
+		task_iteration_index := dag.GetExecution().GetCustomProperties()["iteration_index"].GetIntValue()
+		producerTaskName = metadata.GetParallelForTaskName(producerTaskName, task_iteration_index)
+		glog.V(4).Infof("TaskIteration - ProducerTaskName: %v", producerTaskName)
+		glog.Infof("Attempting to retrieve outputs from a ParallelFor iteration")
+	}
+	return producerTaskName
+}
+
+// Helper for checking if collecting outputs is required for downstream tasks.
+func getParallelForIterationCount(task *metadata.Execution, dag *metadata.Execution) (bool, int64) {
```
droctothorpe, line 2266: Consider consolidating with `isCollection`.

backend/src/v2/driver/driver.go
mprahl, line 1877: My preference would be to have a function called `GetIterationCount` that if the return value is not 0, then call `CollectInputs` with `currentTask` and `iterationCount` and let `CollectInputs` create `starterTaskQueue`. It's a bit odd that the BFS algorithm is split the way it is where the "queue" is exposed in `resolveUpstreamParameters`.

backend/src/v2/driver/driver.go
```diff
@@ -2006,7 +2028,253 @@ func resolveUpstreamArtifacts(cfg resolveUpstreamOutputsConfig) (*pipelinespec.A
...
+				if err != nil {
+					return nil, nil, cfg.err(err)
+				}
+				parallelForParameterList = append(parallelForParameterList, outputParameters[workingOutputKey])
+			}
+			continue
+
+			// Base case 2: currentTask is a DAGExecution within a loop but is NOT a ParallelFor
+			// Head DAG
+		} else if iterations == nil {
```
mprahl, line 2137: Could you change this to `if` instead of `else if` since the previous if block always calls `continue`?

backend/src/v2/driver/driver.go
zazulam: This was moved from `resolveUpstreamParameters`, I can update it to remove the outer loop.

backend/src/v2/driver/driver.go
```diff
@@ -2006,7 +2028,253 @@ func resolveUpstreamArtifacts(cfg resolveUpstreamOutputsConfig) (*pipelinespec.A
...
+						}
+					}
+					if !successfulOneOfTask {
+						return "", "", fmt.Errorf("processing OneOf: No successful task found")
+					}
+				}
+			}
+		}
+	}
+	if tempSubTaskName != subTaskName {
```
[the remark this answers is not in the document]
zazulam: It's the former, thanks for the suggestion!

backend/src/v2/driver/driver.go
zazulam: Totally make sense, most of this was a consequence of working with artifacts and then parameters and ensuring it all fit together. I'll condense it to have the queue set within the BFS function.

backend/src/v2/driver/driver.go
zazulam: From my experience, there's only 1 selector per output key that we are using.
This is from my artifact test when resolving from the inner for loop:
```
tempOutputsArtifacts: map[Output:artifact_selectors:{producer_subtask:"for-loop-1"  output_artifact_key:"pipelinechannel--read-single-dataset-generate-model-results"}] 

tempOutputsArtifacts: map[pipelinechannel--read-single-dataset-generate-model-results:artifact_selectors:{producer_subtask:"for-loop-2"  output_artifact_key:"pipelinechannel--read-single-dataset-generate-model-results"}]
```

mprahl (approved):
/lgtm
/approve

Awesome work!

PR #11758, opened 2025-03-18
feat(proto): Add TTL fields to KFP IR yaml proto

rimolive:
Add the TTL field to IR yaml proto spec.

**Checklist:**


api/v2alpha1/pipeline_spec.proto
```diff
@@ -1111,4 +1111,7 @@ message PipelineConfig {
 
   // Name of the mutex to ensure mutual exclusion
   string mutex_name = 2;
+
+  // TTL Pipeline configuration
+  int32 pipelineTtl = 3;
```
HumairAK, line 1116: So I think there is agreement we can have this but we should get consensus on naming and description, at the moment at the very least the description in my mind is insufficient and doesn't describe to the user what they may expect, for example: 

If I set this, what am I setting TTL for? Will my pipeline be deleted after this ttl? will my pipeline run? will the DB be emptied out, is the action of ttl same as if I do "pipeline or pipelinerun delete" ? 

At the moment if I were the user I would be confused about which of the above is happening

rimolive: Talking to @mprahl, he suggested to change the field name to `resourceTTL`. Given the description, looks like this name fits better. What do you think?

api/v2alpha1/pipeline_spec.proto
```diff
@@ -1111,4 +1111,9 @@ message PipelineConfig {
 
   // Name of the mutex to ensure mutual exclusion
   string mutex_name = 2;
+
+  // Time to live configuration after the pipeline run is completed. When TTL
+  // is reached, the Kubernetes resource that represents the pipeline run will
+  // be deleted by the underlying pipeline backend.
+  int32 pipelineTtl = 3;
```
mprahl, line 1118:
```suggestion
  // Time to live configuration after the pipeline run is completed for
  // ephemeral Kubernetes resources created by the pipeline run. 
  int32 resourceTtl = 3;

```

HumairAK:
thank you @rimolive 

/lgtm
/approve

Issue #11761, opened 2025-03-19
[backend] Terminating a run causes missing artifacts (logs) on the Kubeflow UI


ameya-parab:
### Environment

*  How did you deploy Kubeflow Pipelines (KFP)?

Deployed using the raw [1.8 manifest](https://github.com/kubeflow/manifests/tree/v1.8.1) on AWS EKS 1.29 cluster

*  KFP version: v1.8.1
*  KFP SDK version: 1.8.22

### Steps to reproduce

When a running workflow is terminated from the UI, the `main.log` Output Artifact is not displayed on the UI, even though the Pod that runs the task pushed the local file to the Log repository (S3 in our case).

```python
import kfp
from kfp import dsl
from kfp.components import create_component_from_func


def sleep_and_print_time_func():

    import time
    from datetime import datetime

    total_time = 300  # 5 minutes in seconds
    interval = 5  # print every 5 seconds

    print(
        "Starting sleep component. Will run for 5 minutes, printing time every 5 seconds."
    )

    for i in range(0, total_time, interval):
        current_time = datetime.now().strftime("%Y-%m-%d %H:%M:%S")
        print(f"Current time: {current_time}")

        # Don't sleep after the last iteration
        if i < total_time - interval:
            time.sleep(interval)

    print("Sleep component completed.")

sleep_and_print_component = create_component_from_func(
    func=sleep_and_print_time_func, base_image="python:3.7"
)


# Define the pipeline
@kfp.dsl.pipeline(
    name="Sleep",
)
def sleep_print_pipeline():
    sleep_task = (
        sleep_and_print_component().set_cpu_request("0.5").set_memory_limit("64Mi")
    )

# Trigger the pipeline
client = kfp.Client()
client.create_run_from_pipeline_func(sleep_print_pipeline, arguments={})
```

### Expected result

Even after the workflow is terminated, the `main.log` file should remain in the task's 'Input/Output' tab.

![Image](https://github.com/user-attachments/assets/5b1a2b4e-fc1b-4c92-96be-a6f9ec7840d1)

### Materials and Reference

---

Impacted by this bug? Give it a 👍.

droctothorpe:
After reviewing this issue, we've decided to close it:

User is on KFP v1.8.1 which is EOL. PR #11762 (merged) implemented logs as artifacts in the current codebase. The underlying artifact handling has been reworked.

If you believe this was closed in error, please feel free to reopen with additional context.

Issue #11357, opened 2024-11-05
[bug] Can't see logs from the UI when Argo Workflow is deleted by Persistence Agent


kimwnasptd:
### Environment
*  How do you deploy Kubeflow Pipelines (KFP)?
    * Using the 1.9 manifests from https://github.com/kubeflow/manifests/commit/a38c2be88fbafb0844c0231f0062e4b3719d4737
    * Using the 1.9.1 manifests from https://github.com/kubeflow/manifests/commit/b68bb751c3e067f4a23bad6b9b2c6720645bf266
*  KFP version: 2.3.0
*  KFP SDK version: 2.9.0

### Steps to reproduce

1. Create an Experiment and a Run from the Data Passing pipeline
2. Update the `TTL_SECONDS_AFTER_WORKFLOW_FINISH` env var in `ml-pipeline-persistence` Deployment to be something short, like 60
3. Wait for the Argo Workflow to succeed
4. After the configured time the persistence agent will mark the workflow as completed and remove the Argo Workflow
5. Access the UI and try to see the logs from one pod
6. It'll fail with `Failed to retrieve pod logs.`
7. When clicking on `Details` there's a popup with `Error response: Could not get main container logs: Error: Unable to retrieve workflow status: [object Object].`

![image](https://github.com/user-attachments/assets/b75c0210-9f18-4c98-a312-f55b882a2f7e)

### Expected result

I would expect to see the Pod logs in the case the workflow is deleted, which are stored in MinIO as part of Argo.

### Materials and reference

I see some references of the same error in https://github.com/kubeflow/pipelines/pull/11010 and https://github.com/kubeflow/pipelines/issues/11339 but am not entirely sure if it's the same.

Note that I'm only using upstream manifests and their example installation, that doesn't deviate the MinIO/Argo installation from what's provided in this repo.

Also, when looking at the requests, even if the Workflow is GCed I see requests for logs to the following URL:
`http://localhost:8080/pipeline/k8s/pod/logs?podname=tutorial-data-passing-h2c74-system-container-impl-306858994&runid=8542d9b2-89ee-47bc-a8fc-7210978115eb&podnamespace=kubeflow-user-example-com&createdat=2024-11-05`

Not sure if this is expected, but seemed weird that it tries to get k8s pod logs when we know the pod doesn't exist in the cluster

### Labels

/area frontend
/area backend

---

Impacted by this bug? Give it a 👍.

juliusvonkohout:
[...]

PR #11762, opened 2025-03-19
feat(backend): implement logs as artifacts

droctothorpe:
This PR updates the backend driver and launcher to support execution logs as proper artifacts. This makes logs available after pod and workflow garbage collection and closes many relevant open issues, such as https://github.com/kubeflow/pipelines/issues/11761, https://github.com/kubeflow/pipelines/issues/11357, and https://github.com/kubeflow/pipelines/issues/10036.

The changes were based on [this design proposal](https://docs.google.com/document/d/1GeMTWGHqHO5YLxr_S77McW6Gl4doMoOTMW2m8mODFsY/edit?usp=sharing) that was approved on a community call.

To our pleasant surprise, the UI does not render log artifacts as nodes in the graph. Here's a screenshot for reference:
image

As you can see, the dataset artifact gets a dedicated node and the log artifact does not.

This PR paves the way for some future simplification in the UI. Currently, it does some complex, convoluted, and poorly tested guess work to try to figure out where the log artifact is stored. As a result, the `logs` tab is broken under various circumstances. Once this PR is merged, we can update the UI's logic to simply fetch the precise, unequivocal path of the logs from MLMD.

We can register and upload DAG driver and container driver logs in a follow up PR.

We added a few simple doc strings to the launcher functions to clarify and disambiguate them, which will hopefully be a prelude to some comprehensive refactoring, clarification, and testing.

**Checklist:**

### DIFF
--- a/backend/src/v2/component/launcher_v2.go
+++ b/backend/src/v2/component/launcher_v2.go
@@ -19,6 +19,7 @@ import (
 	"encoding/json"
 	"errors"
 	"fmt"
+	"io"
 	"os"
 	"os/exec"
 	"path/filepath"
@@ -72,6 +73,7 @@ type kubernetesClient struct {
 	Clientset kubernetes.Interface
 }
 
+// NewLauncherV2 is a factory function that returns an instance of LauncherV2.
 func NewLauncherV2(ctx context.Context, executionID int64, executorInputJSON, componentSpecJSON string, cmdArgs []string, opts *LauncherV2Options) (l *LauncherV2, err error) {
 	defer func() {
 		if err != nil {
@@ -163,6 +165,7 @@ func stopWaitingArtifacts(artifacts map[string]*pipelinespec.ArtifactList) {
 	}
 }
 
+// Execute calls executeV2, updates the cache, and publishes the results to MLMD.
 func (l *LauncherV2) Execute(ctx context.Context) (err error) {
 	defer func() {
 		if err != nil {
@@ -243,6 +246,7 @@ func (l *LauncherV2) Execute(ctx context.Context) (err error) {
 		}
 		return l.cacheClient.CreateExecutionCache(ctx, task)
 	}
+
 	return nil
 }
 
@@ -319,6 +323,8 @@ func (l *LauncherV2) publish(
 	return l.metadataClient.PublishExecution(ctx, execution, outputParameters, outputArtifacts, status)
 }
 
+// executeV2 handles placeholder substitution for inputs, calls execute to
+// execute end user logic, and uploads the resulting output Artifacts.
 func executeV2(
 	ctx context.Context,
 	executorInput *pipelinespec.ExecutorInput,
@@ -427,6 +433,46 @@ func prettyPrint(jsonStr string) string {
 
 const OutputMetadataFilepath = "/tmp/kfp_outputs/output_metadata.json"
 
+// We overwrite this as a DI mechanism for testing getLogWriter. We could make
+// it an explicit input to getLogWriter, but then it would have to be an input
+// in getLogWriter's caller as well.
+var osCreateFunc = os.Create
+
+// getLogWriter returns an io.Writer that can either be single-channel to stdout
+// or dual-channel to stdout AND a log file based on the URI of a log artifact
+// in the supplied ArtifactList. Downstream, the resulting log file gets
+// uploaded to the object store.
+func getLogWriter(artifacts map[string]*pipelinespec.ArtifactList) (writer io.Writer) {
+	logsArtifactList, ok := artifacts["logs"]
+
+	if !ok || len(logsArtifactList.Artifacts) == 0 {
+		return os.Stdout
+	}
+
+	logURI := logsArtifactList.Artifacts[0].Uri
+	logFilePath, err := LocalPathForURI(logURI)
+	if err != nil {
+		glog.Errorf("Error converting log artifact URI, %s, to file path.", logURI)
+		return os.Stdout
+	}
+
+	logFile, err := osCreateFunc(logFilePath)
+	if err != nil {
+		glog.Errorf("Error creating logFilePath, %s.", logFilePath)
+		return os.Stdout
+	}
+
+	return io.MultiWriter(os.Stdout, logFile)
+}
+
+// TODO: Clarify the differences between the various execution functions. Are
+// all of them necessary? Do the existing partitions make sense? Can the
+// function names more effectively encapsulate the division in responsibilities?
+// We have tried to clarify this with some  basic function doc strings, but a
+// deeper refactor might be advisable.
+
+// execute downloads input artifacts, prepares the execution environment,
+// executes the end user code, and returns the outputs.
 func execute(
 	ctx context.Context,
 	executorInput *pipelinespec.ExecutorInput,
@@ -445,17 +491,21 @@ func execute(
 		return nil, err
 	}
 
-	// Run user program.
-	executor := exec.Command(cmd, args...)
-	executor.Stdin = os.Stdin
-	executor.Stdout = os.Stdout
-	executor.Stderr = os.Stderr
+	writer := getLogWriter(executorInput.Outputs.GetArtifacts())
+
+	// Prepare command that will execute end user code.
+	command := exec.Command(cmd, args...)
+	command.Stdin = os.Stdin
+	// Pipe stdout/stderr to the aforementioned multiWriter.
+	command.Stdout = writer
+	command.Stderr = writer
 	defer glog.Flush()
-	if err := executor.Run(); err != nil {
+
+	// Execute end user code.
+	if err := command.Run(); err != nil {
 		return nil, err
 	}
 
-	// Collect outputs from output metadata file.
 	return getExecutorOutputFile(executorInput.GetOutputs().GetOutputFile())
 }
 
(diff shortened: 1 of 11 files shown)


backend/src/v2/component/launcher_v2.go
```diff
@@ -472,47 +522,49 @@ func uploadOutputArtifacts(ctx context.Context, executorInput *pipelinespec.Exec
...
 			continue
 		}
-		// TODO: Support multiple artifacts someday, probably through the v2 engine.
-		outputArtifact := artifactList.Artifacts[0]
 
-		// Merge executor output artifact info with executor input
-		if list, ok := executorOutput.Artifacts[name]; ok && len(list.Artifacts) > 0 {
-			mergeRuntimeArtifacts(list.Artifacts[0], outputArtifact)
-		}
+		for _, outputArtifact := range artifactList.Artifacts {
```
droctothorpe, line 526: This whole block only looks like a change because it's intended. We're just actually iterating through the artifacts now instead of taking the first one from the list.

backend/src/v2/component/launcher_v2.go
```diff
@@ -72,6 +73,7 @@ type kubernetesClient struct {
 	Clientset kubernetes.Interface
 }
 
+// NewLauncherV2 is a factory function that returns an instance of LauncherV2.
```
droctothorpe, line 76: If any of these doc strings are not accurate, we're happy to fix them. They're pretty hand-wavy. We just figured something is better than nothing given the complexity of the call chain.

droctothorpe:
cc @zazulam

droctothorpe:
Some of the sample test are reporting errors. Will investigate.

droctothorpe:
No idea why these sample tests keep getting cancelled.

droctothorpe:
Woot everything passing now 🔥

droctothorpe:
I'm going to provide some additional manual validation evidence from the sample tests for improved confidence in the fix:

hello_world.py:

HumairAK:
amazing! I will try to find time to review this, in the meantime I welcome others to take a look as well and give it a shot

zazulam (approved):
/lgtm

backend/src/v2/driver/driver.go
```diff
@@ -1738,13 +1738,30 @@ func resolveUpstreamArtifacts(cfg resolveUpstreamOutputsConfig) error {
...
-	for name, artifact := range outputsSpec.GetArtifacts() {
+	// TODO: Consider putting this behind a feature flag.
+	//
+	// Add a placeholder for a log artifact that will be written to by the
+	// subsequent executor.
+	artifacts := outputsSpec.GetArtifacts()
+	if artifacts == nil {
+		artifacts = make(map[string]*pipelinespec.ComponentOutputsSpec_ArtifactSpec)
+	}
+	artifacts["logs"] = &pipelinespec.ComponentOutputsSpec_ArtifactSpec{
```
HumairAK, line 1756: "logs" seems ubiquitous and generic enough to me that I could easily foresee users using this as their input parameter to their component. And this would then overwrite that input. I don't have a great solution other than, maybe we can add a suffix or prefix like "executor-logs", "pipeline-logs", etc. further reducing the chances of this happening. 

I was considering having it be `logs-<date>` but it feels more appropriate to keep it static

backend/src/v2/component/launcher_v2.go
```diff
@@ -427,6 +433,46 @@ func prettyPrint(jsonStr string) string {
 
 const OutputMetadataFilepath = "/tmp/kfp_outputs/output_metadata.json"
 
+// We overwrite this as a DI mechanism for testing getLogWriter. We could make
+// it an explicit input to getLogWriter, but then it would have to be an input
+// in getLogWriter's caller as well.
+var osCreateFunc = os.Create
```
HumairAK, line 439: optional: another recurring pattern we have been using is just passing an `opts` argument around which feels a little more extensible/cleaner to me if we need to do something similar elsewhere, I'm not strongly opinionated on this one

backend/src/v2/component/launcher_v2.go
```diff
@@ -427,6 +433,46 @@ func prettyPrint(jsonStr string) string {
 
 const OutputMetadataFilepath = "/tmp/kfp_outputs/output_metadata.json"
 
+// We overwrite this as a DI mechanism for testing getLogWriter. We could make
+// it an explicit input to getLogWriter, but then it would have to be an input
+// in getLogWriter's caller as well.
```
HumairAK, line 438:
```suggestion
// We overwrite this as a DI mechanism for testing getLogWriter.
```

backend/src/v2/component/launcher_v2.go
```diff
@@ -427,6 +433,46 @@ func prettyPrint(jsonStr string) string {
...
+// getLogWriter returns an io.Writer that can either be single-channel to stdout
+// or dual-channel to stdout AND a log file based on the URI of a log artifact
+// in the supplied ArtifactList. Downstream, the resulting log file gets
+// uploaded to the object store.
+func getLogWriter(artifacts map[string]*pipelinespec.ArtifactList) (writer io.Writer) {
+	logsArtifactList, ok := artifacts["logs"]
+
+	if !ok || len(logsArtifactList.Artifacts) == 0 {
+		return os.Stdout
+	}
```
HumairAK, line 450: couple of questions: 

1. can we have more than one artifacts["logs"]? 
2. when can we not have logs? when we don't opt in? 

afaik we haven't made this optional so I'm wondering if we should be explicit here and ensure there is always one and only one "logs" entry? 

(lmk if I have misunderstood this one)

backend/src/v2/component/launcher_v2.go
```diff
@@ -427,6 +433,46 @@ func prettyPrint(jsonStr string) string {
...
+
+	return io.MultiWriter(os.Stdout, logFile)
+}
+
+// TODO: Clarify the differences between the various execution functions. Are
+// all of them necessary? Do the existing partitions make sense? Can the
+// function names more effectively encapsulate the division in responsibilities?
+// We have tried to clarify this with some  basic function doc strings, but a
+// deeper refactor might be advisable.
```
HumairAK, line 472: while I agree with the spirit of this comment, I'm not sure I find this to be an appropriate place for this, a github issue would make more sense to me

backend/src/v2/component/launcher_v2.go
```diff
@@ -472,47 +522,49 @@ func uploadOutputArtifacts(ctx context.Context, executorInput *pipelinespec.Exec
...
 		}
-		// TODO: Support multiple artifacts someday, probably through the v2 engine.
-		outputArtifact := artifactList.Artifacts[0]
 
-		// Merge executor output artifact info with executor input
-		if list, ok := executorOutput.Artifacts[name]; ok && len(list.Artifacts) > 0 {
-			mergeRuntimeArtifacts(list.Artifacts[0], outputArtifact)
-		}
+		for _, outputArtifact := range artifactList.Artifacts {
+			glog.Infof("outputArtifact in uploadOutputArtifacts call: ", outputArtifact)
```
HumairAK, line 527:
```suggestion
			glog.Infof("outputArtifact in uploadOutputArtifacts call: %s", outputArtifact.Name)
```
did you mean to do something like this?

HumairAK (changes requested):

backend/src/v2/component/launcher_v2.go
HumairAK: appreciate this, they look good to me!

backend/src/v2/driver/driver.go
```diff
@@ -1738,13 +1738,30 @@ func resolveUpstreamArtifacts(cfg resolveUpstreamOutputsConfig) error {
...
+// provisionOuutputs prepares output references that will get saved to MLMD.
 func provisionOutputs(pipelineRoot, taskName string, outputsSpec *pipelinespec.ComponentOutputsSpec, outputUriSalt string) *pipelinespec.ExecutorInput_Outputs {
 	outputs := &pipelinespec.ExecutorInput_Outputs{
 		Artifacts:  make(map[string]*pipelinespec.ArtifactList),
 		Parameters: make(map[string]*pipelinespec.ExecutorInput_OutputParameter),
 		OutputFile: component.OutputMetadataFilepath,
 	}
-	for name, artifact := range outputsSpec.GetArtifacts() {
+	// TODO: Consider putting this behind a feature flag.
+	//
```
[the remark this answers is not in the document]
HumairAK: Given that other users may opt instead to have their own logging solutions (which is a common approach), for example via fluentd, etc

I think having it be something that is behind a toggle makes sense, maybe something at the api server level an option, config, or env variable. What do you think? 

I think we should enable it by default on kfp manifests and for release.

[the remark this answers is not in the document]
dandawg: I think a toggle makes sense, perhaps even at the `component` level (as a parameter). That way users aren't forced to persist logs via this artifact (and therefore accumulate data in their object store, and log artifacts in their metadata/database).

I know some underlying system code and Python dependencies will generate verbose logs. The user doesn't always have easy access to control logs at this level (for example, the base image they use is produced by someone else), so they might appreciated being able to toggle it at the component level.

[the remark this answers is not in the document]
dandawg: You may also just be running someone else's component, but not want to persist all of the logs, or dive into their code to change logging.

backend/src/v2/driver/driver.go
dandawg: The idea of a special "Logs" artifact type comes to mind. (perhaps a future PR). KFP could then add special functionality around this (like it does for Metrics or Datasets for example).

backend/src/v2/driver/driver.go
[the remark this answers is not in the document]
HumairAK: good suggestion @dandawg! I think this is a useful feature, though potentially non-trivial and require proto changes (unless we already have some api here we can leverage)

I would say this is an optional feat for this PR depending on @droctothorpe and @cmdevoto , we can also iterate on it and have this be a future enhancement so we can get this one in time for 2.5

droctothorpe:
Thanks for the thorough review, @HumairAK! At Kubecon Europe so may not be able to get to implementing the changes until next week.

backend/src/v2/component/launcher_v2.go
droctothorpe: Technically, passing an options object is more correct. The only problem is that `execute`, which calls `getLogWriter`, and `executeV2`, which calls `execute`, and `Execute`, which calls `executeV2` would all require updated function signatures to handle the `opts` parameter. Many of these functions lack unit tests as well. With that in mind, I think that the global is probably an acceptable compromise for the time being. We can update the function signatures and implement proper unit tests in a separate pass.

backend/src/v2/component/launcher_v2.go
droctothorpe: AFAIK, the values are either 1 (logs are enabled) or 0 (logs are disabled and / or we're executing a run from before this feature was implemented).

That being said, I'll make the check more explicit: `!= 1` instead of `== 0`.

backend/src/v2/component/launcher_v2.go
droctothorpe: Agreed! Deleted.

droctothorpe: https://github.com/kubeflow/pipelines/issues/11791

backend/src/v2/component/launcher_v2.go
droctothorpe: I was logging the whole thing while validating / debugging, but I think `.Name` is sufficient going forward. Updated.

backend/src/v2/driver/driver.go
droctothorpe: Agreed, @HumairAK. Updating to `executor-logs`.

droctothorpe:
Addressed all feedback **except** for putting the feature behind a flag. Will address that in a standalone commit when I'm not at the airport 🙃 .

droctothorpe:
Pushed up the flag. Per the community call, it defaults to true if not set. I had to pass the value from the manifest to the compiler in the API, then the driver, and finally the launcher. Something that should be pretty easy was surprisingly complicated. Shout out to @CarterFendley for reviewing it before submission. He's listed as co-author.

All unit tests are passing locally. Just waiting to see if the e2e tests pass remotely. Here's some manual validation evidence:

`PUBLISH_LOGS` not set:
image

`PUBLISH_LOGS` set to `"true"`:
image

`PUBLISH_LOGS` set to `"false"`:
image

Here's the basic pipeline I used for validation:

```python
import kfp
from kfp import dsl

@dsl.component
def foo():
    print("foo")

@dsl.pipeline(name="publish-logs")
def pipeline():
    task = foo()
    task.set_caching_options(False)

client = kfp.Client()
```

I elected to pass the value through the driver and launcher options objects to be more explicit about the dependency and because it was aligned with the other environment variables. There's a TODO [here](https://github.com/kubeflow/pipelines/blob/master/backend/src/v2/cmd/launcher-v2/main.go#L28) about switching to cobra. We should probably do that at some point IMO.

droctothorpe:
Relevant sticker I received at Kubecon (it's a scratch and sniff!):

image

droctothorpe:
All tests passing. CC @HumairAK. Thanks again for the feedback everyone!

HumairAK:
I would hang that sticker everywhere

HumairAK:
amazing work, thank you folks! 

/lgtm
/approve

HumairAK:
@juliusvonkohout I don't believe it should, this is simply adding a new artifact for every component that writes the logs to it, it's a toggle-able feature, so users are free to disable if they wish to

PR #11766, opened 2025-03-21
docs: add proposal for KFP component expectations and standards

dandawg:
This PR adds a proposal to address the critical need for clear, official standards for Kubeflow Pipelines (KFP) components. The current lack of defined expectations for component structure, quality, usability, and documentation has led to inconsistencies, user confusion, and increased development burden. **This proposal focuses on establishing and implementing official component expectations and standards for the KFP repository to improve the overall KFP user experience for both users and contributors.**

**Checklist:**

### DIFF
--- a/proposals/11766-KFP-component-expectations-and-standards/11766-KFP-component-expectations-and-standards.md
+++ b/proposals/11766-KFP-component-expectations-and-standards/11766-KFP-component-expectations-and-standards.md
@@ -40,13 +40,14 @@ This proposal advocates for the formal definition, documentation, and enforcemen
 * **Create a Component Checklist:** Develop a detailed checklist to guide component developers and reviewers. This checklist will serve as a practical tool to ensure adherence to the defined expectations. Key areas covered by the checklist can include:
 
     * **Python Function Documentation and Accessibility:**  Ensure core Python function is documented and usable both within and outside KFP pipelines.
-    * **Dependency Declarations:**  Explicitly list Python libraries, environment variables, and PyPI index/server requirements.
+    * **Dependency Declarations:**  Explicitly list Python libraries, environment variables, and PyPI index/server requirements. Also, note minimum KFP (SDK or backend) version requirements.
     * **Environment Setup and Requirements:**  Detail cluster dependencies (CRD, ConfigMap, Secret, etc), and system resource expectations (GPU, memory, CPU).
     * **Usage Instructions Across Environments:** Provide clear instructions for setting up and running components in different environments (local, cluster, etc.), considering differing needs for authentication, interacting with outputs, testing, and etc.
     * **Input/Output and Artifact Documentation:**  Clearly document component inputs, outputs, metadata, artifacts, errors, and logging.
     * **Performance and Testing Information:**  Provide guidance on expected runtime, resource usage, and include tests with instructions for execution.
+    * **Ownership:** Make it clear who is responsible for the individual component maintenance (if applicable). Have procedures for transfering ownership and for deprecating abandoned components.
 
-* **Document and Promot Standards:**  Clearly document the defined component expectations and the component checklist in prominent locations, such as:
+* **Document and Promote Standards:**  Clearly document the defined component expectations and the component checklist in prominent locations, such as:
 
     * **Top-level `README.md` in the KFP repository.**
     * **Dedicated section in the KFP website documentation.**
@@ -62,7 +63,7 @@ This proposal advocates for the formal definition, documentation, and enforcemen
 
 To implement this proposal, the following steps are necessary:
 
-1. **Formalize Component Expectations:**  Finalize and officially document the "Component Expectations" after giving time for community feedback and consensus.
+1. **Formalize Component Expectations:**  Give time for community feedback and consensus on this proposal. Finalize and officially document the "Component Expectations".
 2. **Develop the Component Checklist:**  Create a practical Component Checklist based on the outlined areas above, ensuring it is both user-friendly and comprehensive.
 3. **Document Standards and Checklist:**  Publish the Component Expectations and Checklist in prominent locations (repository README/Guide, website documentation).
 4. **Implement Review Process:**  Establish the review process for new component contributions and updates, incorporating the Component Checklist as a key evaluation tool.


proposals/11766-KFP-component-expectations-and-standards/11766-KFP-component-expectations-and-standards.md
```diff
@@ -0,0 +1,74 @@
...
+* **User Confusion and Frustration:** Without clear expectations, users struggle to understand component setup, dependencies, and intended usage. This applies to both components that users build, as well as components they rely on from the community. The lack of clarity increases the learning curve for KFP and hinders efficient pipeline development. Components that are poorly built can also cause problems in runtime environments, and across enterprise ML organizations.
+* **Contribution Barriers:**  Contributors lack clear guidelines on what constitutes a high-quality, acceptable component for the official KFP repository. This ambiguity discourages contributions and arguably has led to a proliferation of inconsistent and subpar components in the past.
+* **Increased Support Burden:**  The lack of standards results in increased support requests from users struggling with poorly documented or unreliable components.
+
+These issues collectively undermine the KFP user experience and hinder the growth of the KFP component ecosystem.
+
+## Proposed Solution: Define and Implement Official Component Expectations and Standards
```
mprahl, line 28: Things that come to mind:

1. We should mention prompt CVE fixes a must. If a component uses container images or dependencies with moderate or higher CVEs and aren't addressed within 2 months, they should be removed from the repo.
2. We need clear ownership of each component. You can't just contribute it and abandon it. If the component doesn't have an owner, then it should go up for "adoption" in the KFP community. If there are no takers, it should be removed.
3. A component is like an API, so there should be clear versioning.
4. The minimum KFP version and SDK should be listed. It'd be cool if we defined standardize metadata around this so the SDK could fail to compile if you are using an unsupported KFP SDK or something.

proposals/11766-KFP-component-expectations-and-standards/11766-KFP-component-expectations-and-standards.md
```diff
@@ -0,0 +1,74 @@
...
+
+    * **Code reviews specifically focusing on adherence to standards.**
+    * **Automated checks to verify documentation and checklist completion.**
+    * **Community feedback and validation of component quality.**
+
+## Implementation Steps
+
+To implement this proposal, the following steps are necessary:
+
+1. **Formalize Component Expectations:**  Finalize and officially document the "Component Expectations" after giving time for community feedback and consensus.
```
mprahl, line 65: I was hoping that this would be part of this KEP so that this KEP establishes the following:

- The KFP community wants clear guidelines and expectations on components.
- The KFP community agrees on what those guidelines are.

From there, it'd be fairly actionable.

proposals/11766-KFP-component-expectations-and-standards/11766-KFP-component-expectations-and-standards.md
dandawg: I love your suggestions on 2-4. I'll see how I can work this in.

On 1, remember that the base image is a parameter in a component, so it is meant to be changeable. The component isn't defined by a specific base image (though it's useful to have a default). Most components in the wild run on (a relatively small set of) public base images (python:3.9, pytorch/pytorch, etc.). As far as standards go, we can stress using reputable/well-established images for defaults by preference. Setting specific CVE fix requirements feels too detailed for standards.

mprahl: @dandawg I think to align with your base image ideas, it'd be good if the components defaulted to a maintained base image when applicable in the event there are dependencies that are hard to install (e.g. compiles C, or requires non-Python dependencies). It could be as simple as setting a standard where the SDK builtin components are functions that return components, so it can prefill some of the component arguments like the base image.

dandawg: @mprahl I believe we're mostly on the same page, but let me know if I'm not getting something. If I understand your comment right, I think this is something already provided for in the dsl.component class. We can set the default image to a recommended generalized public image, and then users can override it if they wish (I believe this is also a plus for Python component functionality over the load yaml route). 

There are some really good public images out there that precisely take care of low level stuff like GPU drivers, CUDA versions, gcc, and the like, as well as providing python dependencies (for example the [deep learning images](https://cloud.google.com/deep-learning-containers/docs/choosing-container#pytorch) from Google).

There may be base images that we want, but that don't exist in the public space. I love the idea of figuring out how we can add support here, but I lean towards deferring this conversation until we have a specific component that we are solving for.

proposals/11766-KFP-component-expectations-and-standards/11766-KFP-component-expectations-and-standards.md
dandawg: I reworded number 1 to be a little more explicit on this. Of course, this proposal itself seeks to be the starting point for the discussion.

proposals/11766-KFP-component-expectations-and-standards/11766-KFP-component-expectations-and-standards.md
dandawg: My latest push some clarity around @mprahl 's points.

mprahl: @dandawg if I'm importing the component from Python, how do you override the base image since the base image is set in the decorator? Looking at the Python SDK code, it also looks like the base image is not set on the `PythonComponent` but is stored in a `ComponentInfo` object in a `REGISTERED_MODULES` dict.

dandawg: @mprahl I had to test this out to remember--which I did. I created a component in a file, imported it via python, changed the image, and then tested on a local cluster that the updated image was used in a run.
```
# import @dsl.component wrapped function from my.py
from my import comp
comp.component_spec.implementation.container.image = "python:3.11-slim-buster"

@pipeline()
 def new_pipe():
     op = comp()

client.create_run_from_pipeline_func(pipeline_func=new_pipe)

```
It's not exactly a friendly command, but that is how you can change the image after importing. This also works with yaml-based components.

droctothorpe:
All for this. Great proposal, Dan!

ryanrusson:
This is a much needed intent! As a user of Kubeflow over the last 5 years, KFP component creation has definitely come a long way. However, the lack of defined, official standards has led to a lot of frustration and misuse by myself and the users of the platforms I have supported.

dandawg:
Thanks for taking a look @droctothorpe and @ryanrusson, and for the validation that this is something that would be helpful in the field. 

I look forward to working more on it once the PR gets accepted.

PR #11771, opened 2025-03-24
Issue 11744

hbelmiro:

**Checklist:**


nithin8702:
This is very useful and waiting for this PR to get merged. Thanks

backend/src/apiserver/resource/resource_manager.go
```diff
@@ -495,7 +498,7 @@ func (r *ResourceManager) CreateRun(ctx context.Context, run *model.Run) (*model
 		RunId: run.UUID,
 		RunAt: run.RunDetails.CreatedAtInSec,
 	}
-	executionSpec, err := tmpl.RunWorkflow(run, runWorkflowOptions)
+	executionSpec, err := tmpl.RunWorkflow(run, runWorkflowOptions, r.proxyConfig)
```
mprahl, line 501: On second thought, it might be beneficial to have it be part of the `Template` interface/objects such as adding a method called `GetProxyConfig` or something so that way the proxy configuration is stored on the template and doesn't need to be passed around.

backend/src/v2/cmd/driver/main.go
```diff
@@ -100,6 +110,15 @@ func validate() error {
...
 	}
+	if *httpProxy == unsetProxyArgValue {
+		return fmt.Errorf("argument --%s must be specified", httpProxyArg)
+	}
+	if *httpsProxy == unsetProxyArgValue {
+		return fmt.Errorf("argument --%s must be specified", httpsProxyArg)
+	}
+	if *noProxy == unsetProxyArgValue {
+		return fmt.Errorf("argument --%s must be specified", noProxyArg)
+	}
```
mprahl, line 121: The error message isn't particularly clear that you can set these to empty values. How about this?
```suggestion
	if *httpProxy == unsetProxyArgValue {
		return fmt.Errorf("argument --%s is required but can be an empty value", httpProxyArg)
	}
	if *httpsProxy == unsetProxyArgValue {
		return fmt.Errorf("argument --%s is required but can be an empty value", httpsProxyArg)
	}
	if *noProxy == unsetProxyArgValue {
		return fmt.Errorf("argument --%s is required but can be an empty value", noProxyArg)
	}
```

backend/src/v2/component/launcher_v2.go
```diff
@@ -178,24 +178,26 @@ func (l *LauncherV2) Execute(ctx context.Context) (err error) {
 	var outputArtifacts []*metadata.OutputArtifact
 	status := pb.Execution_FAILED
 	defer func() {
-		if perr := l.publish(ctx, execution, executorOutput, outputArtifacts, status); perr != nil {
+		if err == nil {
```
mprahl, line 181: Does this still publish failures? If `err` is `nil`, then I think status is always `pb.Execution_COMPLETE` due to it meaning that `executeV2` did not return an error.

backend/src/v2/driver/driver.go
```diff
@@ -429,6 +432,27 @@ func initPodSpecPatch(
 		userEnvVar = append(userEnvVar, k8score.EnvVar{Name: envVar.GetName(), Value: envVar.GetValue()})
 	}
 
+	if proxyConfig.GetHttpProxy() != "" {
+		glog.Infof("Setting http_proxy as %s", proxyConfig.GetHttpProxy())
+		userEnvVar = append(userEnvVar, k8score.EnvVar{Name: "http_proxy", Value: proxyConfig.GetHttpProxy()})
```
mprahl, line 437: Are these environment variable names meant to be lowercase?

backend/src/v2/cmd/driver/main.go
```diff
@@ -70,6 +75,11 @@ var (
 	cachedDecisionPath = flag.String("cached_decision_path", "", "Cached Decision output path")
 	conditionPath      = flag.String("condition_path", "", "Condition output path")
 	logLevel           = flag.String("log_level", "1", "The verbosity level to log.")
+
+	// proxy
+	httpProxy  = flag.String(httpProxyArg, unsetProxyArgValue, "The proxy for HTTP connections.")
```
mprahl, line 80: From my reading of the code, the Driver itself does not use these but uses it as environment variables for the Launcher Pod in the Pod spec patch. Should Driver itself uses these by setting these environment variables after validating the flag values?

I think the only connections it makes are to MLMD and the Kubernetes API but it could be good to set this to be on the safe side.

backend/src/v2/component/launcher_v2.go
hbelmiro: Yeah, I haven't noticed that.
I'll revert it.

backend/src/v2/cmd/driver/main.go
hbelmiro: I'm not sure it's needed. But I agree on adding it for safety.

hbelmiro:
/rerun KFP SDK Execution Tests - K8s v1.31.0

hbelmiro:
/rerun "KFP SDK Execution Tests - K8s v1.31.0"

backend/src/v2/driver/driver.go
hbelmiro: As discussed offline, added both lower and uppercase.

backend/src/v2/component/launcher_v2.go
hbelmiro: Done.

backend/src/apiserver/resource/resource_manager.go
hbelmiro: It will increase coupling. `Template` [is created in `PipelineVersionsWebhook`](https://github.com/kubeflow/pipelines/blob/2efcde5efd3952b91ea79a5ee6dbf064282f719a/backend/src/apiserver/webhook/pipelineversion_webhook.go#L86C20-L86C33), for example. `PipelineVersionsWebhook` doesn't need to know `proxy.Config`, but it would need to depend on it to create a `Template`.
Unless we make it optional and lose the benefit of having the compiler on our side 🤷🏽‍♂️

backend/src/v2/cmd/driver/main.go
```diff
@@ -100,6 +110,15 @@ func validate() error {
...
 	}
+	if *httpProxy == unsetProxyArgValue {
+		return fmt.Errorf("argument --%s is required but can be an empty value", httpProxyArg)
+	}
+	if *httpsProxy == unsetProxyArgValue {
+		return fmt.Errorf("argument --%s is required but can be an empty value", httpsProxyArg)
+	}
+	if *noProxy == unsetProxyArgValue {
+		return fmt.Errorf("argument --%s is required but can be an empty value", noProxyArg)
+	}
```
HumairAK, line 121: why are these required?

backend/src/apiserver/config/config_test.go
```diff
@@ -30,7 +31,7 @@ import (
 
 func fakeResourceManager() *resource.ResourceManager {
 	clientManager := resource.NewFakeClientManagerOrFatalV2()
-	resourceManager := resource.NewResourceManager(clientManager, &resource.ResourceManagerOptions{CollectMetrics: false})
+	resourceManager := resource.NewResourceManager(clientManager, &resource.ResourceManagerOptions{CollectMetrics: false}, proxy.EmptyConfig())
```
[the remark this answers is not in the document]
hbelmiro: No longer applies since we changed how to interact with `proxy.Config`.

backend/src/apiserver/resource/resource_manager.go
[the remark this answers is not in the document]
hbelmiro: No longer applies, since we changed how to interact with `proxy.Config`.

hbelmiro: No longer applies, since we changed how to interact with `proxy.Config`.

backend/src/v2/cmd/driver/main.go
hbelmiro: I made them required to avoid not setting them by mistake.

backend/src/v2/compiler/argocompiler/container.go
```diff
@@ -180,6 +180,9 @@ func (c *workflowCompiler) addContainerDriverTemplate() string {
 		"--pod_spec_patch_path", outputPath(paramPodSpecPatch),
 		"--condition_path", outputPath(paramCondition),
 		"--kubernetes_config", inputValue(paramKubernetesConfig),
+		"--http_proxy", c.proxyConfig.GetHttpProxy(),
+		"--https_proxy", c.proxyConfig.GetHttpsProxy(),
+		"--no_proxy", c.proxyConfig.GetNoProxy(),
```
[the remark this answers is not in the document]
hbelmiro: Same as https://github.com/kubeflow/pipelines/pull/11771#discussion_r2025482767.

backend/src/v2/cmd/driver/main.go
hbelmiro: Also, if they were optional, we would have to run the default logic at these points again.

hbelmiro:
/rerun-workflow "KFP SDK Execution Tests - K8s v1.29.2"

backend/test/v2/resources/env-var.py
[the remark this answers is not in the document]
hbelmiro: Done.

backend/src/v2/compiler/argocompiler/container.go
```diff
@@ -209,6 +212,14 @@ func (c *workflowCompiler) addContainerDriverTemplate() string {
 			Command:   c.driverCommand,
 			Args:      args,
 			Resources: driverResources,
+			Env: []k8score.EnvVar{
+				{Name: "http_proxy", Value: c.proxyConfig.GetHttpProxy()},
+				{Name: "HTTP_PROXY", Value: c.proxyConfig.GetHttpProxy()},
+				{Name: "https_proxy", Value: c.proxyConfig.GetHttpsProxy()},
+				{Name: "HTTPS_PROXY", Value: c.proxyConfig.GetHttpsProxy()},
+				{Name: "no_proxy", Value: c.proxyConfig.GetNoProxy()},
+				{Name: "NO_PROXY", Value: c.proxyConfig.GetNoProxy()},
```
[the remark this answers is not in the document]
hbelmiro: Done.

hbelmiro:
@mprahl @HumairAK I've addressed/answered all your comments. Can you please take another look?

nithin8702:
Hi. Are you adding proxies to all pipeline images driver, launcher, api, etc.

hbelmiro:
> @hbelmiro could you please rebase?

@mprahl done.

backend/src/apiserver/config/proxy/config.go
```diff
@@ -0,0 +1,132 @@
...
+type Config interface {
+	GetHttpProxy() string
+	GetHttpsProxy() string
+	GetNoProxy() string
+	GetEnvVars() []k8score.EnvVar
+}
+
+var (
+	configInstance Config = &nonInitializedConfig{}
+	emptyConfig           = newConfig("", "", "")
```
mprahl, line 37: I would have expected empty config to have the default no proxy set but I see that it's only used in the tests.

backend/src/v2/cmd/compiler/main.go
```diff
@@ -35,18 +43,34 @@ var (
 	launcher     = flag.String("launcher", "", "v2 launcher image")
 	driver       = flag.String("driver", "", "v2 driver image")
 	pipelineRoot = flag.String("pipeline_root", "", "pipeline root")
+	httpProxy    = flag.String(httpProxyArg, unsetProxyArgValue, "The proxy for HTTP connections.")
+	httpsProxy   = flag.String(httpsProxyArg, unsetProxyArgValue, "The proxy for HTTPS connections.")
+	noProxy      = flag.String(noProxyArg, unsetProxyArgValue, "Addresses that should ignore the proxy.")
 )
 
 func main() {
```
mprahl, line 51: I couldn't find where this command is used. If it's user facing only (i.e. backend doesn't call it) then I think we should not require the proxy flags as it impacts UX for the majority of cases.

backend/src/apiserver/template/template_test.go
```diff
@@ -172,6 +173,8 @@ func TestToSwfCRDResourceGeneratedName_EmptyName(t *testing.T) {
 }
 
 func TestScheduledWorkflow(t *testing.T) {
+	proxy.InitializeConfigWithEnv()
```
mprahl, line 176: In tests, it'd probably be safer to initialize with an empty proxy config so that it's deterministic (not dependent on if you have those environment variables set locally).

backend/src/apiserver/config/proxy/config.go
hbelmiro: Maybe we can rename it.
By "Empty", one can think of `newConfig("", "", "")`, which is a valid value (and not really empty), or the default values.
How about `blankConfig`? As a non-native English speaker, I'm not sure how different from `empty` it really is.

backend/src/v2/cmd/compiler/main.go
hbelmiro: It's the DAG driver entry point. It's not user-facing. I tried to make it mandatory only for non-user-facing parts.
The only user-facing part should be the apiserver env vars. If you find any other, please let me know.

mprahl (commented):
A few minor comments but this overall looks good!

backend/src/apiserver/config/proxy/config.go
mprahl: I think `empty` is fine. I think I was equating `empty` to `default` but it's not the same.

backend/src/apiserver/template/template_test.go
mprahl: Yeah, `proxy.EmptyConfig()` so that the user's environment variables don't impact the tests.

mprahl (approved):
/lgtm

hbelmiro:
@HumairAK can you take another look?

nithin8702:
This is much required feature to us as we heavily involved in using corporate proxy. Thanks for this team.

nithin8702:
Could someone please approve this PR

@HumairAK  @mprahl

PR #11774, opened 2025-03-26
feat(backend): Create a validating webhook for the PipelineVersion kind

mprahl:

This supersedes https://github.com/kubeflow/pipelines/pull/11684 with the original commit from @VaniHaripriya and the second commit from me with a few fixes.

As part of implementing https://github.com/kubeflow/pipelines/pull/11551 to introduce a Kubernetes Native API for Pipelines and Pipeline Versions, this PR adds a validating webhook as outlined [here](https://github.com/kubeflow/pipelines/tree/master/proposals/11551-kubernetes-native-api#validating-webhook)

**Checklist:**


backend/src/apiserver/main.go
```diff
@@ -239,6 +282,50 @@ func startHttpProxy(resourceManager *resource.ResourceManager) {
...
+	go func() {
+		defer wg.Done()
+
+		if tlsCertPath != "" && tlsKeyPath != "" {
+			if !common.FileExists(tlsCertPath) || !common.FileExists(tlsKeyPath) {
+				glog.Fatalf("TLS certificate/key paths are set but files do not exist")
+			}
+			glog.Info("Starting the Kubernetes webhook with TLS")
+			err := server.ListenAndServeTLS(tlsCertPath, tlsKeyPath)
+			if err != nil && err != http.ErrServerClosed {
```
HumairAK, line 314:
```suggestion
			if err != nil && !errors.Is(err, http.ErrServerClosed) {
```

so that this can succeed against wrapped errors (credit to goland for this one) 
same [here](https://github.com/kubeflow/pipelines/pull/11774/files#diff-d3ae27a3b5003f91b2cbf384575c21f07dd142165daf842fc79fd8459a844829R321)

backend/src/apiserver/client_manager/client_manager.go
```diff
@@ -71,6 +74,18 @@ const (
 	clientBurst = "ClientBurst"
 )
 
+var scheme *runtime.Scheme
+
+func init() {
```
HumairAK, line 79: there is already a clientmanager init step, is there benefit to doing this out of that context? if so maybe a comment here might help future devs

backend/src/apiserver/webhook/pipelineversion_webhook.go
```diff
@@ -0,0 +1,135 @@
...
+
+	if oldPipelineVersion.Generation != newPipelineVersion.Generation {
+		return nil, newBadRequestError("Pipeline spec is immutable; only metadata changes (labels/annotations) are allowed")
+	}
+
+	return nil, nil
+}
+
+func (p *PipelineVersionsWebhook) ValidateDelete(_ context.Context, _ runtime.Object) (ctrladmission.Warnings, error) {
+	return nil, nil
```
HumairAK, line 122: I'm guessing this is because we don't do any delete validation, still, it's a bit odd seeing this, maybe a quick comment is worthwhile

backend/src/apiserver/webhook/pipelineversion_webhook.go
```diff
@@ -0,0 +1,135 @@
...
+	}
+
+	err = common.ValidatePipelineName(tmpl.V2PipelineName())
+	if err != nil {
+		return nil, newBadRequestError(err.Error())
+	}
+
+	if pipelineVersion.Name != tmpl.V2PipelineName() {
+		return nil, newBadRequestError("The object name must match spec.pipelineSpec.pipelineInformation.name")
+	}
```
HumairAK, line 98: the pipelineversion name does not have to be the same name as the pipeline name, I might be reading this one wrong though -- I also don't see this check in the rest db validation

backend/src/apiserver/main.go
```diff
@@ -72,12 +83,59 @@ func main() {
...
+
+	defer clientManager.Close()
+
+	backgroundCtx, backgroundCancel := context.WithCancel(signals.SetupSignalHandler())
+	defer backgroundCancel()
+
+	wg := sync.WaitGroup{}
+
+	wg.Add(1)
+	webhookServer, err := startWebhook(clientManager.ControllerClient(), &wg)
```
HumairAK, line 118: right now it looks like we are always starting the server, which seems unnecessary when k8s native mode is not enabled, so it makes more sense to me to introduce this flag in this pr first so we can make this conditional, if we are merging this first

backend/src/apiserver/client_manager/client_manager.go
mprahl: It's a common pattern in Kubernetes controllers so that the code panics when the package is imported if there is a coding error. Guaranteed to encounter it if it's at all imported in a test.

backend/src/apiserver/webhook/pipelineversion_webhook.go
mprahl: You're right. I'll remove it. When I originally wrote this, I think I thought it was the pipeline version name but it's just an arbitrary name it seems.

HumairAK:
/lgtm
/approve

tytyty

mprahl:
Sorry @HumairAK I had forgotten to update `deploy-kfp.sh`.

PR #11782, opened 2025-03-31
feat(backend): Add a mutating webhook for the PipelineVersion kind

mprahl:

This mutating webhook adds labels to query pipeline versions by their pipeline ID and pipeline name, and adds an owner's reference to the pipeline it's a part of.

**Checklist:**

### DIFF
--- a/backend/src/apiserver/webhook/pipelineversion_webhook.go
+++ b/backend/src/apiserver/webhook/pipelineversion_webhook.go
@@ -119,14 +119,81 @@ func (p *PipelineVersionsWebhook) ValidateDelete(_ context.Context, _ runtime.Ob
 	return nil, nil
 }
 
-func NewPipelineVersionWebhook(client ctrlclient.Client) (http.Handler, error) {
+func (p *PipelineVersionsWebhook) Default(ctx context.Context, obj runtime.Object) error {
+	pipelineVersion, ok := obj.(*k8sapi.PipelineVersion)
+	if !ok {
+		return &apierrors.StatusError{
+			ErrStatus: metav1.Status{
+				Code:    http.StatusBadRequest,
+				Reason:  metav1.StatusReasonBadRequest,
+				Message: fmt.Sprintf("expected a PipelineVersion object but got %T", pipelineVersion),
+			},
+		}
+	}
+
+	pipeline := &k8sapi.Pipeline{}
+	nsName := types.NamespacedName{Namespace: pipelineVersion.Namespace, Name: pipelineVersion.Spec.PipelineName}
+
+	err := p.Client.Get(ctx, nsName, pipeline)
+	if err != nil {
+		if apierrors.IsNotFound(err) {
+			return newBadRequestError("The spec.pipelineName doesn't map to an existing Pipeline object")
+		}
+
+		return err
+	}
+
+	if pipelineVersion.Labels == nil {
+		pipelineVersion.Labels = map[string]string{}
+	}
+
+	// Labels for efficient querying
+	pipelineVersion.Labels["pipelines.kubeflow.org/pipeline-id"] = string(pipeline.UID)
+	pipelineVersion.Labels["pipelines.kubeflow.org/pipeline"] = pipeline.Name
+
+	trueVal := true
+
+	for i := range pipelineVersion.OwnerReferences {
+		ownerRef := &pipelineVersion.OwnerReferences[i]
+		if ownerRef.APIVersion != k8sapi.GroupVersion.String() || ownerRef.Kind != "Pipeline" {
+			continue
+		}
+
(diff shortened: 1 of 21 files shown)


backend/Makefile
```diff
@@ -93,3 +93,7 @@ kind-load-driver-debug:
 
 .PHONY: kind-build-and-load-driver-debug
 kind-build-and-load-driver-debug: image_driver_debug kind-load-driver-debug
+
+.PHONY: test-webhook
+test-webhook:
+	WEBHOOK_INTEGRATION=true go test -v -run '.+WebhookIntegration$$' "$(CURDIR)/src/apiserver/webhook/..."
```
HumairAK, line 99: this and the integration seems oddly (maybe arguably) placed when we have a `backend/test/integration` location

backend/src/apiserver/webhook/pipelineversion_webhook.go
```diff
@@ -119,14 +119,77 @@ func (p *PipelineVersionsWebhook) ValidateDelete(_ context.Context, _ runtime.Ob
...
+		return err
+	}
+
+	if pipelineVersion.Labels == nil {
+		pipelineVersion.Labels = map[string]string{}
+	}
+
+	// Labels for efficient querying
+	pipelineVersion.Labels["pipelines.kubeflow.org/pipeline-id"] = string(pipeline.UID)
+	pipelineVersion.Labels["pipelines.kubeflow.org/pipeline"] = string(pipeline.Name)
```
HumairAK, line 152:
```suggestion
	pipelineVersion.Labels["pipelines.kubeflow.org/pipeline"] = pipeline.Name
```

backend/src/apiserver/webhook/pipelineversion_webhook_test.go
```diff
@@ -207,3 +217,196 @@ func TestPipelineVersionWebhook_ValidateUpdate_MetadataChangeAllowed(t *testing.
...
+	p, err = pClient.Create(context.TODO(), p, metav1.CreateOptions{})
+	require.NoError(t, err, "Expected no error when creating the pipeline")
+
+	pv = createPipelineVersion(t, pvClient, pv)
+
+	t.Log("Checking that the mutating webhook set the pipeline-id label")
+	require.Equal(t, pv.GetLabels()["pipelines.kubeflow.org/pipeline-id"], string(p.GetUID()))
+
+	t.Log("Checking that the mutating webhook set the pipeline label")
+	require.Equal(t, pv.GetLabels()["pipelines.kubeflow.org/pipeline"], string(p.GetName()))
```
HumairAK, line 389:
```suggestion
	require.Equal(t, pv.GetLabels()["pipelines.kubeflow.org/pipeline"], p.GetName())
```

backend/src/apiserver/webhook/pipelineversion_webhook.go
```diff
@@ -119,14 +119,77 @@ func (p *PipelineVersionsWebhook) ValidateDelete(_ context.Context, _ runtime.Ob
...
+
+		return nil
+	}
+
+	pipelineVersion.OwnerReferences = append(pipelineVersion.OwnerReferences, metav1.OwnerReference{
+		APIVersion: k8sapi.GroupVersion.String(),
+		Kind:       "Pipeline",
+		Name:       pipeline.Name,
+		UID:        pipeline.UID,
+	})
```
HumairAK, line 171: an observation here, I'm wondering how this affects pipeline deletions, currently the db store implementation you can't delete the Pipeline without first having to delete the PipelineVersions, the use of ownerreferences here makes it a non issue for the k8s native store implementation, something to note for @rimolive maybe when doing pipeline deletions

backend/Makefile
mprahl: Whoops, thanks for pointing that out.

backend/src/apiserver/webhook/pipelineversion_webhook.go
mprahl: @HumairAK and @rimolive do you think we should set `blockOwnerDeletion` to prevent the pipeline from being deleted if the pipeline versions are still around? I lean towards no because I like the idea of being able to fully delete a pipeline with one API call but wanted your take.

HumairAK: sigh, I agree with you but I think there has been discussion in the past about adding a separate api or toggle in the pipeline deletion call something like a parameter "cascade=true" to delete all at once, so in this case I'm leaning towards consistency and blocking it, but open to thoughts

mprahl: Okay, I added `blockOwnerDeletion`!

PR #11657, opened 2025-02-20
fix(frontend): fixes optional pipeline inputs. Fixes #11632

izuku-sds:
**Description of your changes:**
This PR resolves #11632
Optional pipeline inputs were incorrectly marked as a required field.

This fix ensures that the optional pipeline inputs are not marked as a required field.

**Checklist:**
- [x] You have [signed off your commits](https://www.kubeflow.org/docs/about/contributing/#sign-off-your-commits)
- [x] The title for your pull request (PR) should follow our title convention. [Learn more about the pull request title convention used in this repository](https://github.com/kubeflow/pipelines/blob/master/CONTRIBUTING.md#pull-request-title-convention).

PR #11788, opened 2025-04-02
Handle optional pipeline inputs in the driver

mprahl:
If the pipeline run is submitted without specifying an optional parameter and there is no default, it was not handled by the driver. The approach taken is explicitly set null for these values and let the driver handle if the component parameter has a default that can be used.

**Checklist:**

### DIFF
--- a/backend/src/v2/driver/driver.go
+++ b/backend/src/v2/driver/driver.go
@@ -17,6 +17,7 @@ package driver
 import (
 	"context"
 	"encoding/json"
+	"errors"
 	"fmt"
 	"slices"
 	"strconv"
@@ -50,6 +51,8 @@ var dummyImages = map[string]string{
 	"argostub/createpvc": "create PVC",
 	"argostub/deletepvc": "delete PVC",
 }
+
+var ErrResolvedParameterNull = errors.New("the resolved input parameter is null")
 
 // TODO(capri-xiyue): Move driver to component package
 // Driver options
@@ -179,6 +182,7 @@ func RootDAG(ctx context.Context, opts Options, mlmd *metadata.Client) (executio
 	if err != nil {
 		return nil, err
 	}
+
 	executorInput := &pipelinespec.ExecutorInput{
 		Inputs: &pipelinespec.ExecutorInput_Inputs{
 			ParameterValues: opts.RuntimeConfig.GetParameterValues(),
@@ -784,7 +788,7 @@ func extendPodSpecPatch(
 	for _, secretAsVolume := range kubernetesExecutorConfig.GetSecretAsVolume() {
 		var secretName string
 		if secretAsVolume.SecretNameParameter != nil {
-			resolvedSecretName, err := resolveInputParameter(ctx, dag, pipeline, opts, mlmd,
+			resolvedSecretName, err := resolveInputParameterStr(ctx, dag, pipeline, opts, mlmd,
 				secretAsVolume.SecretNameParameter, inputParams)
 			if err != nil {
 				return fmt.Errorf("failed to resolve secret name: %w", err)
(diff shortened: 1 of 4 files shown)


mprahl:
/cc @HumairAK

I wanted your take on it since I'm not thrilled with the use of a sentinel but it's the only way I could come up with to handle the null optional pipeline input case since null values are not allowed in other circumstances (see `handleParamTypeValidationAndConversion`).

backend/src/v2/driver/driver.go
```diff
@@ -1521,6 +1550,13 @@ func resolveInputParameter(
 		if !ok {
 			return nil, paramError(fmt.Errorf("parent DAG does not have input parameter %s", componentInput))
 		}
+
+		if v.GetStringValue() == EmptyOptionalPipelineInputValue {
+			// Return a known error in this case and let the caller choose if they can handle an empty optional pipeline
+			// input value.
+			return nil, ErrOptionalPipelineInputValue
+		}
```
HumairAK, line 1558: I don't understand why previously kfp decided to handle this in the launcher, to me it seems inputs should be resolved and submitted as part of executor input, so even component inputs should be handled at driver level

if we did that, we could forego the need for a sentinel right? we can just check if it is unset, and if component input definition provides a default and handle it right here?

backend/src/v2/driver/driver.go
```diff
@@ -179,9 +184,27 @@ func RootDAG(ctx context.Context, opts Options, mlmd *metadata.Client) (executio
...
+
+	for paramName, paramDef := range opts.Component.GetInputDefinitions().GetParameters() {
+		if _, ok := paramValues[paramName]; !ok && paramDef.IsOptional {
+			if paramValues == nil {
+				paramValues = map[string]*structpb.Value{}
+			}
+
+			// Set the defaults for optional parameters not provided in the pipeline run
+			if paramDef.DefaultValue != nil {
+				paramValues[paramName] = paramDef.DefaultValue
```
HumairAK, line 198: it seems to me like this case is handled [here](https://github.com/kubeflow/pipelines/blob/cc1c435f1e06aad3e9b83e57768512a63460b15b/backend/src/v2/compiler/argocompiler/argo.go#L75) right?  since you can only have a default value if it's optional (so the optional check is implicit?) 

I find it weird we're doing this partly in api server and partly in driver, maybe we ought to just do it all in driver? wdyt?

backend/src/v2/driver/driver.go
```diff
@@ -179,9 +184,27 @@ func RootDAG(ctx context.Context, opts Options, mlmd *metadata.Client) (executio
 	if err != nil {
 		return nil, err
 	}
+
+	paramValues := opts.RuntimeConfig.GetParameterValues()
+
+	for paramName, paramDef := range opts.Component.GetInputDefinitions().GetParameters() {
+		if _, ok := paramValues[paramName]; !ok && paramDef.IsOptional {
```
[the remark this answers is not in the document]
mprahl: This code relies on the API server doing the proper validation here:
https://github.com/kubeflow/pipelines/blob/97e57368d1c737d6191dee1ce4aa02fda1d6157a/backend/src/apiserver/template/v2_template.go#L389-L395

backend/src/v2/driver/driver.go
mprahl: Good find! Not sure why I got it in my head that this was needed.

mprahl:
@HumairAK this is ready for review again.

gmfrasca (commented):
/lgtm

HumairAK:
fyi, empty but optional pipeline inputs will be hard to replicate due to a bug in the ui that prevents an unspecified optional string from being entered (I think it's a gh issue somewhere) 

but tested with an empty runtime config and got the expected result

/lgtm
/approve

thanks!

mprahl:
> fyi, empty but optional pipeline inputs will be hard to replicate due to a bug in the ui that prevents an unspecified optional string from being entered (there should be a gh issue somewhere)
> 
> but tested with an empty runtime config and got the expected result
> 
> /lgtm /approve
> 
> thanks!

The UI issue I think is addressed here:
https://github.com/kubeflow/pipelines/pull/11657

PR #11817, opened 2025-04-11
df

droctothorpe:
This PR is intended to gather more information about and test solutions for the MinIO disk space errors that intermittently surface during CI.


droctothorpe:
Trying out the suggestion [here](https://github.com/actions/runner-images/issues/2840#issuecomment-1284059930) which claims to free up 10GB.

PR #11824, opened 2025-04-14
fix(docs): Remove Podman as backend README pre-req

alyssacgoins:

**Checklist:**

### DIFF
--- a/backend/README.md
+++ b/backend/README.md
@@ -10,7 +10,7 @@ This README will help you set up your coding environment in order to build and r
 ## Prerequisites
 Before you begin, ensure you have:
 - [Go installed](https://go.dev/doc/install)
-- Docker or Podman installed (for building container images)
+- Docker installed (for building container images)
 
 Note that you may need to restart your shell after installing these resources in order for the changes to take effect.
 
--- a/backend/README.md
+++ b/backend/README.md
@@ -10,7 +10,7 @@ This README will help you set up your coding environment in order to build and r
 ## Prerequisites
 Before you begin, ensure you have:
 - [Go installed](https://go.dev/doc/install)
-- Docker installed (for building container images)
+- Docker and Podman installed (for building container images)
 
 Note that you may need to restart your shell after installing these resources in order for the changes to take effect.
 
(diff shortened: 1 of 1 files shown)


hbelmiro (commented):
@alyssacgoins I use Podman and it works fine for me. That said, I'm not sure we should proceed with this change.
What kind of issues did you have? Feel free to ping me if you need any help with it.

alyssacgoins:
> @alyssacgoins I use Podman and it works fine for me. That said, I'm not sure we should proceed with this change. What kind of issues did you have? Feel free to ping me if you need any help with it.

Messaged you! @hbelmiro

HumairAK:
@hbelmiro the issue is more to do with spinning up a kfp kind cluster using the make kfp kind command, @alyssacgoins you can build the images with podman just fine, we can keep the original comment but also drop a note like:

"Note that while you can use Podman to build the KFP images, you will still require Docker installed to run the make dev-kind-cluster commands"

hbelmiro:
> @hbelmiro the issue is more to do with spinning up a kfp kind cluster using the make kfp kind command, @alyssacgoins you can build the images with podman just fine, we can keep the original comment but also drop a note like:
> 
> "Note that while you can use Podman to build the KFP images, you will still require Docker installed to run the make dev-kind-cluster commands"

Yeah, I clarified with @alyssacgoins.
The problem is with `kind load docker-image`. This command supports only images built with Docker.
As a workaround, I save the image to a file and load it with `kind load image-archive`.

hbelmiro:
`kind load --name ${CLUSTER_NAME} image-archive <(podman save ${IMAGE_NAME})` instead of `kind load docker-image` does the trick.

alyssacgoins:
> @hbelmiro the issue is more to do with spinning up a kfp kind cluster using the make kfp kind command, @alyssacgoins you can build the images with podman just fine, we can keep the original comment but also drop a note like:
> 
> "Note that while you can use Podman to build the KFP images, you will still require Docker installed to run the make dev-kind-cluster commands"

Updated

backend/README.md
```diff
@@ -10,7 +10,7 @@ This README will help you set up your coding environment in order to build and r
 ## Prerequisites
 Before you begin, ensure you have:
 - [Go installed](https://go.dev/doc/install)
-- Docker or Podman installed (for building container images)
+- Docker and Podman installed (for building container images)
```
hbelmiro, line 13: Only Docker is required.

hbelmiro:
> > `kind load --name ${CLUSTER_NAME} image-archive <(podman save ${IMAGE_NAME})` instead of `kind load docker-image` does the trick.
> 
> @hbelmiro @HumairAK Should I create an issue and add this change? Assuming it wouldn't be placed in this PR.

It would be nice to be agnostic. But this is not urgent.
Just an issue is fine.

backend/README.md
alyssacgoins: Updated

hbelmiro (commented):
/lgtm
/ok-to-test

HumairAK:
/approve
tyty!

PR #11844, opened 2025-04-21
chore(backend): fixed support for Podman in Makefile

hbelmiro:
Fixed support for Podman in Makefile.

**Checklist:**

### DIFF
--- a/backend/Makefile
+++ b/backend/Makefile
@@ -6,7 +6,14 @@ KIND_NAME ?= dev-pipelines-api
 CERT_MANAGER_VERSION ?= v1.16.2
 
 # Container Build Params
-CONTAINER_ENGINE ?= docker
+CONTAINER_ENGINE ?= $(shell \
+	if command -v docker >/dev/null 2>&1; then \
+		echo docker; \
+	elif command -v podman >/dev/null 2>&1; then \
+		echo podman; \
+	fi \
+)
+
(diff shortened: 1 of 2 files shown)


mprahl (approved):
/approve
/lgtm

backend/Makefile
```diff
@@ -6,7 +6,14 @@ KIND_NAME ?= dev-pipelines-api
...
 # Container Build Params
-CONTAINER_ENGINE ?= docker
+CONTAINER_ENGINE ?= $(shell \
+	if command -v docker >/dev/null 2>&1; then \
+		echo docker; \
+	elif command -v podman >/dev/null 2>&1; then \
+		echo podman; \
+	fi \
+)
+
```
HumairAK, line 16: hrmm not sure how I feel about this, this can lead to surprises, I think it's better to be predictable here with default (docker), and explicitly tell the user to switch to podman via env vars etc.

this is fine if we do this _everywhere_ where docker is used and podman can be as easily subbed, but not in a one off scenario for this makefile

hbelmiro: > explicitly tell the user to switch to podman via env vars etc.

@HumairAK I'm not sure I'm following.
Are you saying to run _this very same `if`_ where the default (docker) can't be used? Like in https://github.com/kubeflow/pipelines/pull/11844/files#diff-5024268fbb0b80c2fca407bd63672f2e0e8c2b61e87f3bdd361e6a7a41a9dda3R88?

PR #11849, opened 2025-04-23
feat(components): dynamically compute expected value in Profile Controller: Fixes #11828

### LINKED ISSUE #11828: [feature] Remove hardcoded values from Pipelines Controller´s `sync.py`
### Feature Area

/area components

The Profile Controller `sync.py` has [hardcoded values](https://github.com/kubeflow/pipelines/blob/866ff3556a4454ccb52f1594bbca4167a04c0d3e/manifests/kustomize/base/installs/multi-user/pipelines-profile-controller/sync.py#L128) for each type of component. It may be better to compute the expected values from the `desired_resources` list. This would allow for easier customization of said file.

One may want additional resources by default for each user namespace, for instance a `PodDefault`. Since the aforementioned values are hardcoded, for any modification one needs to manually modify those values.

Manually updating the hardcoded values to reflect any additions to the base resources.

---

Love this idea? Give it a 👍.

grudloffev:

This pull request addresses the issue #11828 regarding hardcoded values in the Profile Controller `sync.py` for each type of component. The changes involve dynamically computing the expected values from the `desired_resources` list. This enhancement allows for easier customization of the file, enabling users to add additional resources by default for each user namespace without manually modifying hardcoded values.

**Checklist:**

### DIFF
--- a/manifests/kustomize/base/installs/multi-user/pipelines-profile-controller/sync.py
+++ b/manifests/kustomize/base/installs/multi-user/pipelines-profile-controller/sync.py
@@ -90,6 +90,27 @@ def get_settings_from_env(controller_port=None,
     return settings
 
 
+def compute_desired_status(attachments, desired_resources):
+    expected_counts = {}
+
+    for resource in desired_resources:
+        kind = resource["kind"]
+        api_version = resource["apiVersion"]
+        key = f"{kind}.{api_version}"
+        if key not in expected_counts:
+            expected_counts[key] = 0
+        expected_counts[key] += 1
+
+    desired_status = {}
+    for key, expected_count in expected_counts.items():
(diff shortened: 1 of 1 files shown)


hbelmiro:
@grudloffev can you please [link the PR to the issue](https://docs.github.com/en/issues/tracking-your-work-with-issues/using-issues/linking-a-pull-request-to-an-issue)?

grudloffev:
FYI this has been tested in my environment, without any issues so far.

grudloffev:
Not sure about the failing test here, in the other PR -which builds on this- all tests are passing

grudloffev:
The error is the same for all retests so it doesn't seem like flakiness. There is a difference with the implementation from the other PR, which may be the reason for the error, also rebased again just in case.

PR #11861, opened 2025-04-25
feat(backend): parameterize retryStrategy  input in Argo workflow

alyssacgoins:
This PR improves how the retryStrategy field is set in the argo workflow spec. When a component with setRetry enabled is included in a pipeline, two single separate container and executor templates are added - retry-system-container-executor and retry-system-container-impl. All four retryStrategy fields (duration, factor maxDuration and limit) are input as parameters into these templates, in order to allow the reuse (previously, separate templates were created for every unique retryStrategy configuration).

If the input pipeline contains no components with setRetry enabled, these two templates will not be created. 

**Checklist:**
- [x ] You have [signed off your commits](https://www.kubeflow.org/docs/about/contributing/#sign-off-your-commits)

### DIFF
--- a/backend/src/v2/compiler/argocompiler/argo.go
+++ b/backend/src/v2/compiler/argocompiler/argo.go
@@ -333,21 +333,25 @@ func hashValue(value interface{}) (string, error) {
 }
 
 const (
-	paramComponent        = "component"      // component spec
-	paramTask             = "task"           // task spec
-	paramContainer        = "container"      // container spec
-	paramImporter         = "importer"       // importer spec
-	paramRuntimeConfig    = "runtime-config" // job runtime config, pipeline level inputs
-	paramParentDagID      = "parent-dag-id"
-	paramExecutionID      = "execution-id"
-	paramIterationCount   = "iteration-count"
-	paramIterationIndex   = "iteration-index"
-	paramExecutorInput    = "executor-input"
-	paramDriverType       = "driver-type"
-	paramCachedDecision   = "cached-decision"   // indicate hit cache or not
-	paramPodSpecPatch     = "pod-spec-patch"    // a strategic patch merged with the pod spec
-	paramCondition        = "condition"         // condition = false -> skip the task
-	paramKubernetesConfig = "kubernetes-config" // stores Kubernetes config
+	paramComponent               = "component"      // component spec
+	paramTask                    = "task"           // task spec
+	paramContainer               = "container"      // container spec
+	paramImporter                = "importer"       // importer spec
+	paramRuntimeConfig           = "runtime-config" // job runtime config, pipeline level inputs
+	paramParentDagID             = "parent-dag-id"
+	paramExecutionID             = "execution-id"
+	paramIterationCount          = "iteration-count"
+	paramIterationIndex          = "iteration-index"
+	paramExecutorInput           = "executor-input"
+	paramDriverType              = "driver-type"
+	paramCachedDecision          = "cached-decision"            // indicate hit cache or not
+	paramPodSpecPatch            = "pod-spec-patch"             // a strategic patch merged with the pod spec
+	paramCondition               = "condition"                  // condition = false -> skip the task
+	paramKubernetesConfig        = "kubernetes-config"          // stores Kubernetes config
+	paramRetryMaxCount           = "retry-max-count"            // limit on number of retries
+	paramRetryBackOffDuration    = "retry-backoff-duration"     // duration of backoff between retries
+	paramRetryBackOffFactor      = "retry-backoff-factor"       // multiplier for backoff duration between retries
+	paramRetryBackOffMaxDuration = "retry-backoff-max-duration" // limit on backoff duration between retries
 )
 
 func runID() string {
--- a/backend/src/v2/compiler/argocompiler/container.go
+++ b/backend/src/v2/compiler/argocompiler/container.go
@@ -267,10 +267,11 @@ func (c *workflowCompiler) containerExecutorTask(name string, inputs containerEx
 		Template: c.addContainerExecutorTemplate(name, refName),
 		When:     when,
 		Arguments: wfapi.Arguments{
-			Parameters: []wfapi.Parameter{
+			Parameters: append([]wfapi.Parameter{
 				{Name: paramPodSpecPatch, Value: wfapi.AnyStringPtr(inputs.podSpecPatch)},
-				{Name: paramCachedDecision, Value: wfapi.AnyStringPtr(inputs.cachedDecision), Default: wfapi.AnyStringPtr("false")},
-			},
+				{Name: paramCachedDecision, Value: wfapi.AnyStringPtr(inputs.cachedDecision), Default: wfapi.AnyStringPtr("false")}},
+				c.getTaskRetryParametersWithValues(name)...,
+			),
 		},
 	}
 
@@ -289,8 +290,8 @@ func (c *workflowCompiler) addContainerExecutorTemplate(name string, refName str
 	nameContainerImpl := "system-container-impl"
 	taskRetrySpec := c.getTaskRetryPolicySpec(name)
 	if taskRetrySpec != nil {
-		nameContainerExecutor = name + "-" + nameContainerExecutor
-		nameContainerImpl = name + "-" + "system-container-impl"
+		nameContainerExecutor = "retry-" + nameContainerExecutor
+		nameContainerImpl = "retry-" + nameContainerImpl
 	}
 	_, ok := c.templates[nameContainerExecutor]
 	if ok {
@@ -299,20 +300,22 @@ func (c *workflowCompiler) addContainerExecutorTemplate(name string, refName str
 	container := &wfapi.Template{
 		Name: nameContainerExecutor,
 		Inputs: wfapi.Inputs{
-			Parameters: []wfapi.Parameter{
+			Parameters: append([]wfapi.Parameter{
 				{Name: paramPodSpecPatch},
-				{Name: paramCachedDecision, Default: wfapi.AnyStringPtr("false")},
-			},
+				{Name: paramCachedDecision, Default: wfapi.AnyStringPtr("false")}},
+				c.addParameterDefault(c.getTaskRetryParameters(name), "0")...,
+			),
 		},
 		DAG: &wfapi.DAGTemplate{
 			Tasks: []wfapi.DAGTask{{
 				Name:     "executor",
 				Template: nameContainerImpl,
 				Arguments: wfapi.Arguments{
-					Parameters: []wfapi.Parameter{{
+					Parameters: append([]wfapi.Parameter{{
 						Name:  paramPodSpecPatch,
-						Value: wfapi.AnyStringPtr(inputParameter(paramPodSpecPatch)),
-					}},
+						Value: wfapi.AnyStringPtr(inputParameter(paramPodSpecPatch))}},
+						c.addParameterInputPath(c.getTaskRetryParameters(name))...,
+					),
 				},
 				// When cached decision is true, the container
 				// implementation template will be skipped, but
@@ -335,12 +338,11 @@ func (c *workflowCompiler) addContainerExecutorTemplate(name string, refName str
 		args = append(args, "--publish_logs", value)
 	}
 	executor := &wfapi.Template{
-		Name:          nameContainerImpl,
-		RetryStrategy: c.getTaskRetryStrategy(name),
+		Name: nameContainerImpl,
 		Inputs: wfapi.Inputs{
-			Parameters: []wfapi.Parameter{
-				{Name: paramPodSpecPatch},
-			},
+			Parameters: append([]wfapi.Parameter{
+				{Name: paramPodSpecPatch}},
+				c.getTaskRetryParameters(name)...),
 		},
 		// PodSpecPatch input param is where actual image, command and
 		// args come from. It is treated as a strategic merge patch on
@@ -448,6 +450,13 @@ func (c *workflowCompiler) addContainerExecutorTemplate(name string, refName str
 			Env:     commonEnvs,
 		},
 	}
+	// If retry policy is set, add retryStrategy to executor
+	if taskRetrySpec != nil {
+		executor.RetryStrategy = c.getTaskRetryStrategyFromInput(inputParameter(paramRetryMaxCount),
+			inputParameter(paramRetryBackOffDuration),
+			inputParameter(paramRetryBackOffFactor),
+			inputParameter(paramRetryBackOffMaxDuration))
+	}
 	// Update pod metadata if it defined in the Kubernetes Spec
 	kubernetesConfigParam := c.wf.Spec.Arguments.GetParameterByName(argumentsKubernetesSpec + refName)
 
@@ -527,38 +536,71 @@ func (c *workflowCompiler) getTaskRetryPolicySpec(name string) *pipelinespec.Pip
 	return taskSpec.RetryPolicy
 }
 
-func (c *workflowCompiler) getTaskRetryStrategy(name string) *wfapi.RetryStrategy {
+func (c *workflowCompiler) getTaskRetryParameters(name string) []wfapi.Parameter {
 	retryPolicy := c.getTaskRetryPolicySpec(name)
 	if retryPolicy == nil {
 		return nil
 	}
+	// Create slice of parameters with "Name" field set.
+	parameters := []wfapi.Parameter{
+		{Name: paramRetryMaxCount},
+		{Name: paramRetryBackOffDuration},
+		{Name: paramRetryBackOffFactor},
+		{Name: paramRetryBackOffMaxDuration}}
+	return parameters
+}
 
-	argoBackOffDuration := "0"
-	backoffDuration := retryPolicy.GetBackoffDuration()
-	if backoffDuration != nil {
-		argoBackOffDuration = strconv.FormatInt(backoffDuration.Seconds, 10)
+func (c *workflowCompiler) getTaskRetryParametersWithValues(name string) []wfapi.Parameter {
+	retryPolicy := c.getTaskRetryPolicySpec(name)
+	if retryPolicy == nil {
+		return []wfapi.Parameter{}
 	}
+	// Create slice of all non-nil retryPolicy fields.
+	parameters := append(
+		[]wfapi.Parameter{},
+		wfapi.Parameter{
+			Name:  paramRetryMaxCount,
+			Value: wfapi.AnyStringPtr(retryPolicy.MaxRetryCount),
+		})
+	if retryPolicy.GetBackoffDuration() != nil {
+		retryBackOffDuration := wfapi.AnyStringPtr(retryPolicy.GetBackoffDuration().Seconds)
+		parameters = append(
+			parameters,
+			wfapi.Parameter{
+				Name:  paramRetryBackOffDuration,
+				Value: retryBackOffDuration})
+	}
+	parameters = append(
+		parameters,
+		wfapi.Parameter{
+			Name:  paramRetryBackOffFactor,
+			Value: wfapi.AnyStringPtr(retryPolicy.GetBackoffFactor()),
+		})
+	if retryPolicy.GetBackoffMaxDuration() != nil {
+		retryBackOffMaxDuration := wfapi.AnyStringPtr(retryPolicy.GetBackoffMaxDuration().Seconds)
+		parameters = append(
+			parameters,
+			wfapi.Parameter{
+				Name:  paramRetryBackOffMaxDuration,
+				Value: retryBackOffMaxDuration})
+	}
+	return parameters
+}
 
-	var argoMaxDuration string
-	backoffMaxDuration := retryPolicy.GetBackoffMaxDuration()
-	if backoffMaxDuration != nil {
-		argoMaxDuration = strconv.FormatInt(backoffMaxDuration.Seconds, 10)
+func (c *workflowCompiler) addParameterDefault(parameters []wfapi.Parameter, defaultValue string) []wfapi.Parameter {
+	// Set the "Default" field of each parameter in input slice with input defaultValue.
+	for i := range parameters {
+		parameters[i].Default = wfapi.AnyStringPtr(defaultValue)
 	}
-	backoff := &wfapi.Backoff{
-		Factor: &intstr.IntOrString{
-			Type:   intstr.Int,
-			IntVal: int32(retryPolicy.GetBackoffFactor()),
-		},
-		MaxDuration: argoMaxDuration,
-		Duration:    argoBackOffDuration,
+	return parameters
+}
+
+func (c *workflowCompiler) addParameterInputPath(parameters []wfapi.Parameter) []wfapi.Parameter {
+	// Format "Value" field of each parameter in input slice as '{{inputs.parameters.[parameter name]}}'
+	for i := range parameters {
+		parameters[i].Value = wfapi.AnyStringPtr(inputParameter(parameters[i].Name))
 	}
-	return &wfapi.RetryStrategy{
-		Limit: &intstr.IntOrString{
-			Type:   intstr.Int,
-			IntVal: retryPolicy.MaxRetryCount,
-		},
-		Backoff: backoff,
-	}
+	return parameters
 }
 
 // Extends the PodMetadata to include Kubernetes-specific executor config.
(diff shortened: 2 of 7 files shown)


backend/src/v2/compiler/argocompiler/container.go
```diff
@@ -580,6 +637,26 @@ func extendPodMetadata(
...
+func (c *workflowCompiler) getTaskRetryStrategyFromInput(maxCount string, backOffDuration string, backOffPolicy string,
+	backOffMaxDuration string) *wfapi.RetryStrategy {
+	backoff := &wfapi.Backoff{
+		Factor: &intstr.IntOrString{
+			Type:   intstr.String,
+			StrVal: backOffPolicy,
+		},
+		MaxDuration: backOffMaxDuration,
+		Duration:    backOffDuration,
+	}
```
gmfrasca, line 649:
```suggestion
func (c *workflowCompiler) getTaskRetryStrategyFromInput(maxCount string, backOffDuration string, backOffFactor string,
	backOffMaxDuration string) *wfapi.RetryStrategy {
	backoff := &wfapi.Backoff{
		Factor: &intstr.IntOrString{
			Type:   intstr.String,
			StrVal: backOffFactor,
		},
		MaxDuration: backOffMaxDuration,
		Duration:    backOffDuration,
	}
```

backend/src/v2/compiler/argocompiler/container.go
```diff
@@ -325,11 +349,18 @@ func (c *workflowCompiler) addContainerExecutorTemplate(name string, refName str
 		args = append(args, "--publish_logs", value)
 	}
 	executor := &wfapi.Template{
-		Name:          nameContainerImpl,
-		RetryStrategy: c.getTaskRetryStrategy(name),
+		Name: nameContainerImpl,
+		RetryStrategy: c.getTaskRetryStrategyFromInput(inputParameter(paramRetryMaxCount),
```
gmfrasca, line 353: since the `getTaskRetryStrategy(name)` function is now unused, we should remove it

alyssacgoins:
> Thanks for the PR @alyssacgoins ! Looks like there's some extra merge and dev commits in the history - Could you please rebase/squash your commits into a single (or more, if logical) commit(s)? Thanks!

Yep, squashed down to one commit :+1:

backend/src/v2/compiler/argocompiler/argo.go
```diff
@@ -331,21 +331,25 @@ func hashValue(value interface{}) (string, error) {
...
+	paramExecutionID             = "execution-id"
+	paramIterationCount          = "iteration-count"
+	paramIterationIndex          = "iteration-index"
+	paramExecutorInput           = "executor-input"
+	paramDriverType              = "driver-type"
+	paramCachedDecision          = "cached-decision"            // indicate hit cache or not
+	paramPodSpecPatch            = "pod-spec-patch"             // a strategic patch merged with the pod spec
+	paramCondition               = "condition"                  // condition = false -> skip the task
+	paramKubernetesConfig        = "kubernetes-config"          // stores Kubernetes config
+	paramRetryMaxCount           = "retry-max-count"            // retry max count
```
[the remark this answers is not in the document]
alyssacgoins: Added some more in-detail comments - is that a good level of info?

backend/src/v2/compiler/argocompiler/container.go
alyssacgoins: Yep that variable naming makes much more sense

backend/src/v2/compiler/argocompiler/container.go
alyssacgoins: Removed :+1:

backend/src/v2/compiler/argocompiler/container.go
```diff
@@ -257,11 +257,25 @@ func (c *workflowCompiler) containerExecutorTask(name string, inputs containerEx
...
-		},
+			Parameters: []wfapi.Parameter{{
+				Name:  paramPodSpecPatch,
+				Value: wfapi.AnyStringPtr(inputs.podSpecPatch),
+			}, {
+				Name:  paramCachedDecision,
+				Value: wfapi.AnyStringPtr(inputs.cachedDecision), Default: wfapi.AnyStringPtr("false"),
+			}, {
+				Name:  paramRetryMaxCount,
+				Value: wfapi.AnyStringPtr(c.getTaskRetryParameterValue(name, paramRetryMaxCount)),
```
mprahl, line 268: Optional: I think it'd be cleaner if instead of `getTaskRetryParameterValue` there was a method that returns a slice of `wfapi.Parameter` with all of the retry parameters that are set. This means it could return an empty slice if `set_retry` wasn't called and we can then rely on default values in the arguments/parameters. This would also simplify the code but not needing to have a switch statement on the parameter name.

If you did this approach, I imagine the parameters would added like so:

```go
task.Arguments = append(task.Arguments, c.getTaskRetryParameterValues(name)...)
```

backend/src/v2/compiler/argocompiler/container.go
```diff
@@ -292,6 +301,10 @@ func (c *workflowCompiler) addContainerExecutorTemplate(name string, refName str
 			Parameters: []wfapi.Parameter{
 				{Name: paramPodSpecPatch},
 				{Name: paramCachedDecision, Default: wfapi.AnyStringPtr("false")},
+				{Name: paramRetryMaxCount},
```
mprahl, line 304: Could you set the defaults here so that the task doesn't need to set them and we can just always rely on the template having the correct defaults?

backend/src/v2/compiler/argocompiler/container.go
```diff
@@ -517,38 +548,30 @@ func (c *workflowCompiler) getTaskRetryPolicySpec(name string) *pipelinespec.Pip
...
-		MaxDuration: argoMaxDuration,
-		Duration:    argoBackOffDuration,
-	}
-	return &wfapi.RetryStrategy{
-		Limit: &intstr.IntOrString{
-			Type:   intstr.Int,
-			IntVal: retryPolicy.MaxRetryCount,
-		},
-		Backoff: backoff,
+	value := "0"
```
mprahl, line 557: It seems like `paramRetryMaxCount` should default to `1`. Is there special behavior if it's set to `0`?

mprahl:
@alyssacgoins have you tested the case where a pipeline fails and `set_retry` isn't set on the component? I'm wondering how the default parameter values of `"0"` are handled by Argo Workflows. Looking at the Argo Workflows code, it executes a different code path if retryStrategy is not nil and since default values are being passed, it is no longer nil due to Argo Workflows usage of pointers on values and thus `omitempty` on JSON doesn't take effect if a default Go value is passed (e.g. "" for string or 0 for int).

mprahl:
I think we should have at least one integration test that uses `set_retry` with all the options to ensure Argo Workflows accepts the values passed in as parameters by the KFP Argo compiler.

alyssacgoins:
> @alyssacgoins have you tested the case where a pipeline fails and `set_retry` isn't set on the component? I'm wondering how the default parameter values of `"0"` are handled by Argo Workflows. Looking at the Argo Workflows code, it executes a different code path if retryStrategy is not nil and since default values are being passed, it is no longer nil due to Argo Workflows usage of pointers on values and thus `omitempty` on JSON doesn't take effect if a default Go value is passed (e.g. "" for string or 0 for int).

I added this test case (samples/v2/pipeline_failure_no_retry.py) - though now I'm wondering if it's necessary because I implemented both a retry-enabled and non-retry-enabled template (therefore retryStrategy would be nil on the template produced by this test)

alyssacgoins:
> I think we should have at least one integration test that uses `set_retry` with all the options to ensure Argo Workflows accepts the values passed in as parameters by the KFP Argo compiler.

Added - samples/v2/pipeline_with_retry_all_args.py

backend/src/v2/compiler/argocompiler/container.go
alyssacgoins: Added - currently deciding how to refactor further.

backend/src/v2/compiler/argocompiler/container.go
alyssacgoins: Updated

backend/src/v2/compiler/argocompiler/container.go
alyssacgoins: Initially I was concerned with defaulting this value to 1, because I was setting a retryStrategy on every single template regardless of whether retry parameters were input - and a default "1" value would potentially cause every single template to retry. However, now that I'm using a 2-template approach, (retry and non-retry) there is no chance of this happening, so a default of "1" makes more sense - I have updated this.

backend/src/v2/compiler/argocompiler/container.go
```diff
@@ -257,10 +263,11 @@ func (c *workflowCompiler) containerExecutorTask(name string, inputs containerEx
 		Template: c.addContainerExecutorTemplate(name, refName),
 		When:     when,
 		Arguments: wfapi.Arguments{
-			Parameters: []wfapi.Parameter{
+			Parameters: append([]wfapi.Parameter{
 				{Name: paramPodSpecPatch, Value: wfapi.AnyStringPtr(inputs.podSpecPatch)},
-				{Name: paramCachedDecision, Value: wfapi.AnyStringPtr(inputs.cachedDecision), Default: wfapi.AnyStringPtr("false")},
-			},
+				{Name: paramCachedDecision, Value: wfapi.AnyStringPtr(inputs.cachedDecision), Default: wfapi.AnyStringPtr("false")}},
+				*c.getTaskRetryParameters(name, templateParameterInstructions{false, false, true})...,
```
mprahl, line 269: This is the only time that `setValue` is set and `setDefault` and `setInputPath` aren't utilized in that code path. For readability, let's include the `setValue` branch directly in in the `containerExecutorTask` method or break it out to its own function.

alyssacgoins: Created separate function `getTaskRetryParametersWithValue()` to handle setValue logic.  I believe this was a cleaner approach than including the setValue logic directly within the  `containerExecutorTask` method, but let me know what you think.

samples/v2/sample_test.py
```diff
@@ -37,6 +37,7 @@
 from kubernetes import client, config, utils
 import yaml
 from modelcar import modelcar
+from samples.v2 import pipeline_with_retry, pipeline_with_retry_all_args, pipeline_with_zero_retry
```
mprahl, line 40: I think this will fix your CI failure.
```suggestion
import pipeline_with_retry, pipeline_with_retry_all_args, pipeline_with_zero_retry
```

backend/src/v2/compiler/argocompiler/container.go
```diff
@@ -527,38 +536,55 @@ func (c *workflowCompiler) getTaskRetryPolicySpec(name string) *pipelinespec.Pip
...
+		if setPath {
+			parameters[i].Value = wfapi.AnyStringPtr(inputParameter(parameters[i].Name))
+		}
+		// If specified, set "Default" field with default parameter value.
+		if setDefault {
+			parameters[i].Default = wfapi.AnyStringPtr("0")
+			if parameterType == paramRetryMaxCount {
+				parameters[i].Default = wfapi.AnyStringPtr("1")
+			}
+		}
```
mprahl, line 562: It seems like getTaskRetryParameters is also setting defaults. I'd prefer either of these options:

1. The parameters are always required and just handle setting default values in `getTaskRetryParameters`.
2. Modify `getTaskRetryParameters` to not return a parameter if it wasn't set in the task.

samples/v2/sample_test.py
```diff
@@ -161,6 +162,9 @@ def test(self):
             TestCase(pipeline_func=modelcar.pipeline_modelcar_test),
             TestCase(pipeline_func=parallel_consume_upstream.loop_consume_upstream),
             TestCase(pipeline_func=parallel_after_dependency.loop_with_after_dependency_set),
+            TestCase(pipeline_func=pipeline_with_retry.retry_pipeline),
+            TestCase(pipeline_func=pipeline_with_retry_all_args.retry_pipeline),
+            TestCase(pipeline_func=pipeline_with_zero_retry.retry_pipeline),
```
mprahl, line 167: How do these tests pass since `run_test_case` asserts that the test ran successfully?

backend/src/v2/compiler/argocompiler/container.go
```diff
@@ -527,38 +536,55 @@ func (c *workflowCompiler) getTaskRetryPolicySpec(name string) *pipelinespec.Pip
...
+		parameterType := parameters[i].Name
+		// If specified, set "Value" field with input parameter path (e.g. '{{inputs.parameters.retry-max-count}}')
+		if setPath {
+			parameters[i].Value = wfapi.AnyStringPtr(inputParameter(parameters[i].Name))
+		}
+		// If specified, set "Default" field with default parameter value.
+		if setDefault {
+			parameters[i].Default = wfapi.AnyStringPtr("0")
+			if parameterType == paramRetryMaxCount {
+				parameters[i].Default = wfapi.AnyStringPtr("1")
```
mprahl, line 560: I think you might have been right in your original choice that `0` is the correct default for this. I misread the code here in Argo Workflows:
```go
if retryStrategy.Limit != nil && limit != nil && int32(len(node.Children)) > *limit {
```

samples/v2/sample_test.py
alyssacgoins: Thanks for pointing that out, updated :+1:

backend/src/v2/compiler/argocompiler/container.go
alyssacgoins: Updated back to a single default value of 0

samples/v2/sample_test.py
alyssacgoins: Facepalm moment here, I've updated the tests so that they don't raise an error (and now my build is passing too :slightly_smiling_face:)

backend/src/v2/compiler/argocompiler/container.go
alyssacgoins: I went with option 1 - if I understood correctly - I separated the logic for setting input parameter path and setting default value - they are each in separate functions now

backend/src/v2/compiler/argocompiler/container.go
```diff
@@ -527,38 +536,61 @@ func (c *workflowCompiler) getTaskRetryPolicySpec(name string) *pipelinespec.Pip
...
+		{Name: paramRetryBackOffMaxDuration}}
+	return &parameters
+}
 
-	argoBackOffDuration := "0"
-	backoffDuration := retryPolicy.GetBackoffDuration()
-	if backoffDuration != nil {
-		argoBackOffDuration = strconv.FormatInt(backoffDuration.Seconds, 10)
-	}
+func (c *workflowCompiler) getTaskRetryParametersWithValues(name string) *[]wfapi.Parameter {
```
mprahl, line 553: We don't need to use a pointer here on the returned slice.
```suggestion
func (c *workflowCompiler) getTaskRetryParametersWithValues(name string) []wfapi.Parameter {
```

backend/src/v2/compiler/argocompiler/container.go
```diff
@@ -527,38 +536,61 @@ func (c *workflowCompiler) getTaskRetryPolicySpec(name string) *pipelinespec.Pip
 	return taskSpec.RetryPolicy
 }
 
-func (c *workflowCompiler) getTaskRetryStrategy(name string) *wfapi.RetryStrategy {
+func (c *workflowCompiler) getTaskRetryParameters(name string) *[]wfapi.Parameter {
```
mprahl, line 539: We don't need to use a pointer here on the returned slice.
```suggestion
func (c *workflowCompiler) getTaskRetryParameters(name string) []wfapi.Parameter {
```

backend/src/v2/compiler/argocompiler/container.go
```diff
@@ -527,38 +536,61 @@ func (c *workflowCompiler) getTaskRetryPolicySpec(name string) *pipelinespec.Pip
...
 	retryPolicy := c.getTaskRetryPolicySpec(name)
 	if retryPolicy == nil {
-		return nil
-	}
+		return &[]wfapi.Parameter{}
```
mprahl, line 542: Optional: You can keep it as returning nil once you stop returning a slice pointer as a nil slice can still be read from (e.g. range over it) just not written to until it's initialized in Go.

backend/src/v2/compiler/argocompiler/container.go
```diff
@@ -527,38 +536,61 @@ func (c *workflowCompiler) getTaskRetryPolicySpec(name string) *pipelinespec.Pip
...
-		argoMaxDuration = strconv.FormatInt(backoffMaxDuration.Seconds, 10)
-	}
-	backoff := &wfapi.Backoff{
-		Factor: &intstr.IntOrString{
-			Type:   intstr.Int,
-			IntVal: int32(retryPolicy.GetBackoffFactor()),
-		},
-		MaxDuration: argoMaxDuration,
-		Duration:    argoBackOffDuration,
+func (c *workflowCompiler) addParameterDefault(parameters []wfapi.Parameter, defaultValue string) *[]wfapi.Parameter {
```
mprahl, line 580: No need to return a pointer here.
```suggestion
func (c *workflowCompiler) addParameterDefault(parameters []wfapi.Parameter, defaultValue string) []wfapi.Parameter {
```

samples/v2/pipeline_with_retry.py
```diff
@@ -0,0 +1,31 @@
+import os
+
+from kfp import dsl
+
+from kfp import compiler
+
+_KFP_PACKAGE_PATH = os.getenv('KFP_PACKAGE_PATH')
+
+@dsl.component(kfp_package_path=_KFP_PACKAGE_PATH)
+def print_op_retries(retries: str) -> str:
```
mprahl, line 10: Optional: It doesn't really print anything anymore :laughing:

backend/src/v2/compiler/argocompiler/container.go
alyssacgoins: Updated

backend/src/v2/compiler/argocompiler/container.go
alyssacgoins: Updated

backend/src/v2/compiler/argocompiler/container.go
alyssacgoins: Makes sense, thanks for pointing that out - reverted back to returning nil

backend/src/v2/compiler/argocompiler/container.go
alyssacgoins: Updated

samples/v2/pipeline_with_retry.py
alyssacgoins: Refactored this component to `verify_retries()` which returns a boolean value

mprahl (approved):
/lgtm

samples/v2/pipeline_with_retry.py
```diff
@@ -0,0 +1,30 @@
...
+@dsl.pipeline()
+def retry_pipeline():
+    task1 = verify_retries(retries="{{retries}}").set_retry(num_retries=2)
+    task2 = print_op(text='test').set_retry(num_retries=2, backoff_duration="1 second", backoff_factor=1, backoff_max_duration="1 second")
+    task3 = print_op(text='test').set_retry(num_retries=0)
+
+if __name__ == '__main__':
+    compiler.Compiler.compile(
+        pipeline_func=retry_pipeline,
+        package_path='pipeline_with_retry.yaml')
```
gmfrasca, line 30: code formatting/lint: add newline to end

backend/src/v2/compiler/testdata/hello_world_with_retry_all_args.json
```diff
@@ -0,0 +1,102 @@
...
+    "sdkVersion": "kfp-1.6.5"
+  },
+  "runtimeConfig": {
+    "parameters": {
+      "text": {
+        "stringValue": "hi there"
+      }
+    }
+  }
+}
```
gmfrasca, line 102: code formatting/lint: add newline to end

backend/src/v2/compiler/argocompiler/container.go
```diff
@@ -527,38 +536,61 @@ func (c *workflowCompiler) getTaskRetryPolicySpec(name string) *pipelinespec.Pip
...
-	}
+func (c *workflowCompiler) getTaskRetryParametersWithValues(name string) []wfapi.Parameter {
+	retryPolicy := c.getTaskRetryPolicySpec(name)
+	if retryPolicy == nil {
+		return []wfapi.Parameter{}
+	}
+	// Create slice of all non-nil retryPolicy fields.
+	parameters := append([]wfapi.Parameter{},
+		wfapi.Parameter{Name: paramRetryMaxCount,
+			Value: wfapi.AnyStringPtr(retryPolicy.MaxRetryCount)})
```
gmfrasca, line 561:
```suggestion
	parameters := append(
	    []wfapi.Parameter{},
		wfapi.Parameter{
		    Name: paramRetryMaxCount,
			Value: wfapi.AnyStringPtr(retryPolicy.MaxRetryCount),
		},
	)
```

backend/src/v2/compiler/argocompiler/container.go
```diff
@@ -527,38 +536,61 @@ func (c *workflowCompiler) getTaskRetryPolicySpec(name string) *pipelinespec.Pip
...
+	if retryPolicy.GetBackoffDuration() != nil {
+		retryBackOffDuration := wfapi.AnyStringPtr(retryPolicy.GetBackoffDuration().Seconds)
+		parameters = append(parameters,
+			wfapi.Parameter{Name: paramRetryBackOffDuration,
+				Value: retryBackOffDuration})
```
gmfrasca, line 566:
```suggestion
		parameters = append(
		    parameters,
			wfapi.Parameter{
			    Name: paramRetryBackOffDuration,
				Value: retryBackOffDuration,
			},
		)
```

backend/src/v2/compiler/argocompiler/container.go
```diff
@@ -527,38 +536,61 @@ func (c *workflowCompiler) getTaskRetryPolicySpec(name string) *pipelinespec.Pip
...
+	}
+	parameters = append(parameters,
+		wfapi.Parameter{Name: paramRetryBackOffFactor,
+			Value: wfapi.AnyStringPtr(retryPolicy.GetBackoffFactor())})
```
gmfrasca, line 570:
```suggestion
	parameters = append(
	    parameters,
		wfapi.Parameter{
		    Name: paramRetryBackOffFactor,
			Value: wfapi.AnyStringPtr(retryPolicy.GetBackoffFactor()),
		},
	)
```

backend/src/v2/compiler/argocompiler/container.go
```diff
@@ -527,38 +536,61 @@ func (c *workflowCompiler) getTaskRetryPolicySpec(name string) *pipelinespec.Pip
...
+	if retryPolicy.GetBackoffMaxDuration() != nil {
+		retryBackOffMaxDuration := wfapi.AnyStringPtr(retryPolicy.GetBackoffMaxDuration().Seconds)
+		parameters = append(parameters,
+			wfapi.Parameter{Name: paramRetryBackOffMaxDuration,
+				Value: retryBackOffMaxDuration})
```
gmfrasca, line 575:
```suggestion
		parameters = append(
		    parameters,
			wfapi.Parameter{
			    Name: paramRetryBackOffMaxDuration,
				Value: retryBackOffMaxDuration,
			},
		)
```

backend/src/v2/compiler/argocompiler/container.go
```diff
@@ -289,8 +290,8 @@ func (c *workflowCompiler) addContainerExecutorTemplate(name string, refName str
 	nameContainerImpl := "system-container-impl"
 	taskRetrySpec := c.getTaskRetryPolicySpec(name)
 	if taskRetrySpec != nil {
-		nameContainerExecutor = name + "-" + nameContainerExecutor
-		nameContainerImpl = name + "-" + "system-container-impl"
+		nameContainerExecutor = "retry" + "-" + nameContainerExecutor
+		nameContainerImpl = "retry" + "-" + "system-container-impl"
```
gmfrasca, line 294:
```suggestion
		nameContainerExecutor = "retry-" + nameContainerExecutor
		nameContainerImpl = "retry-" + nameContainerImpl
```

backend/src/v2/compiler/testdata/hello_world_with_retry_all_args.json
alyssacgoins: Fixed. Applied this change to hello_world_with_retry.json as well

backend/src/v2/compiler/argocompiler/container.go
alyssacgoins: Updated

backend/src/v2/compiler/argocompiler/container.go
alyssacgoins: Updated

backend/src/v2/compiler/argocompiler/container.go
alyssacgoins: Updated

backend/src/v2/compiler/argocompiler/container.go
alyssacgoins: Updated

backend/src/v2/compiler/argocompiler/container.go
alyssacgoins: Updated

gmfrasca (commented):
/lgtm

PR #11866, opened 2025-04-26
fix(sdk): resolve issue when creating pipeline version from pipeline name using the cli. Fixes #11810

### LINKED ISSUE #11810: [sdk] kfp cli: cannot upload new version using `--pipeline-name`
### Environment

*  KFP version: 2.4.1
*  KFP SDK version: master / sdk-2.12.1

### Steps to reproduce

create a pipeline:
```bash
kfp pipeline create --pipeline-name test test.yaml
```

try to upload a new version of the pipeline
```bash
kfp pipeline create-version --pipeline-name test --pipeline-version v2 test.yaml
```

throws error: `Either pipeline_id or pipeline_name is required.`

if using the pipeline id instead it works as expected:
```bash
kfp pipeline create-version --pipeline-id ... --pipeline-version v2 test.yaml
```

### Expected result

new pipeline version should be able to upload when providing `--pipeline-name` instead of `--pipeline-id`

### Materials and Reference

when setting the `--pipeline-name` option it will trigger the branch in https://github.com/kubeflow/pipelines/blob/a680e2230c99a30ee03a992daf1ef499851b9b3b/sdk/python/kfp/cli/pipeline.py#L104 and fetch the pipeline id.

the `pipeline_name` variable is however not unset and passed alongside the `pipeline_id` to `upload_pipeline_version`
in https://github.com/kubeflow/pipelines/blob/a680e2230c99a30ee03a992daf1ef499851b9b3b/sdk/python/kfp/cli/pipeline.py#L113 

however https://github.com/kubeflow/pipelines/blob/5a3dc8473e0a8d9e41b865e622099aff94dd1ce1/sdk/python/kfp/client/client.py#L1504 enforces that not both values are set:

```python
        if all([pipeline_id, pipeline_name
               ]) or not any([pipeline_id, pipeline_name]):
            raise ValueError('Either pipeline_id or pipeline_name is required.')
```

and thus erroneously causing the `Either pipeline_id or pipeline_name is required.` error.

if only the `--pipeline-id` parameter is passed it works as expected.

---

happy to try and upstream a patch if pointed in the right direction :)

chjdev:
When trying to create a new pipeline version using the cli it is checked that not both pipeline id and pipeline name are passed in as parameters. However, in the case of passing in the pipeline name the pipeline id is fetched via the client and then both values are passed to `upload_pipeline_version` defeating the initial check and raising an error.

Simply removing the `pipeline_name` argument from the call to `upload_pipeline_version` fixes this as the `pipeline_id` will always be set.

**Checklist:**

### DIFF
--- a/sdk/python/kfp/cli/pipeline.py
+++ b/sdk/python/kfp/cli/pipeline.py
@@ -110,7 +110,6 @@ def create_version(ctx: click.Context,
         pipeline_package_path=package_file,
         pipeline_version_name=pipeline_version,
         pipeline_id=pipeline_id,
-        pipeline_name=pipeline_name,
         description=description)
     output.print_output(
         version,


zazulam (approved):
Thanks for the fix!
/lgtm

hbelmiro (commented):
/lgtm

PR #11882, opened 2025-04-30
feat(backend & sdk):  add support for pipeline-level parallelism limitation WIP

ntny:
**Use case**: Apply a limit on overall computation within the pipeline. This is the first step, with plans to implement similar limits for overall memory and CPU consumption.
**The goal** is to prevent any single pipeline from consuming all available resources in a cooperative cluster.

`
@dsl.pipeline(pipeline_config=PipelineConfig(3))
        def my_pipeline():
            comp()
            comp()
        print(my_pipeline)
`


zazulam:
@ntny can you rebase this PR so the tests can pull in the proto changes you have

Issue #9813, opened 2023-08-03
Tracking: PostgreSQL support in KFP


zijianjoy:
PostgreSQL request has become the top upvoted issue on KFP repo: https://github.com/kubeflow/pipelines/issues/7512. This issue is for tracking the work of this integration.

- [ ] KFP Backend integration
  - [x] Define the DB config for PostgreSQL DB info
    - [x] Basic connection config
    - [ ] Secure connection config: passfile, hostaddr, ssl mode, certificate, etc.
    - [x] Flag to switch between Postgresql driver and MySQL driver
  - [x] ~~#9859~~
  - [ ] Syntax support of PostgreSQL DB in backend
    - [ ] Adopt a different syntax during initialization. (schema difference)
    - [ ] Adopt a different syntax during execution. (Modification of data)
    - [ ] Explore the dialect difference and develop a control mechanism to enable easy testing on the storage layer. 
  - [ ] Testing
    - [ ] Unit testing for Postgresql behavior.
    - [ ] Functional testing for Postgresql behavior.
    - [ ] E2E testing for Postgresql behavior.
  - [ ] Cache server integration
  - [ ] KFP API server integration
  - [ ] Manifest support for Postgresql
    - [x] #9860
    - [x] #9861
    - [ ] Marketplace KFP (managed CloudSQL instance)

- [ ] MLMD integration
  - [ ] Develop Postgresql integration on MLMD
    - [ ] Investigate https://github.com/google/ml-metadata/issues/194#issuecomment-1975207465
  - [ ] Release new version of MLMD
  - [x] #9848
  - [ ] Manifest support for Postgresql in MLMD
    - [ ] Standalone KFP
    - [ ] Full Kubeflow Kubeflow
    - [ ] Marketplace KFP (managed CloudSQL instance)

zijianjoy:
cc @chensun

EshaanAgg:
Hi! I am Eshaan Aggarwal, an avid Open Source enthusiast from India. I am a web developer proficient in GoLang and PostgreSQL and have recently started learning about Kubernetes. I would love to contribute to this issue and hopefully join Kubeflow as a GSoC '24 mentee. Are there any pre-tests or other beginner-friendly contributions I can make to get acquainted with this project and as a proof of skill?

UditNayak:
[...]

Issue #8851, opened 2023-02-15
chore(backend): Change backend unit-testing store from SQLite to MySQL


gkcalat:
### Feature Area

Problem:

- Backend testing uses SQLite as a fake DB instance
- Real DB is based on MySQL
- Two databases have slight differences, which makes the testing useless

Solution:
- Switch the fake DB to MySQL instances

---

Love this idea? Give it a 👍.

PR #11889, opened 2025-05-03
test (backend) Added containerization support in testing and replaced SQLite with MySQL in the api-server/run_store_test

ntny:
- add [testcontainers-go](https://testcontainers.com/?language=go)
- add [testcontainers-go mtsql](https://golang.testcontainers.org/modules/mysql/)

**Motivation:**
1. While it's lightweight and fast, SQLite is less strict than MySQL in terms of data types, constraints, and SQL dialect compliance. This can lead to false positives — tests passing in SQLite but failing in production when using MySQL.
For example, the attached test passes when using SQLite, but fails on MySQL due to stricter enforcement of constraints.
2. Introducing Testcontainers simplifies testing when [adding](https://github.com/kubeflow/pipelines/issues/9813) PostgreSQL support. Existing tests can be reused for both MySQL and PostgreSQL. 
3. Another nice thing is that you’ll be able to use tests for debugging the generated queries, since they now use the MySQL dialect — the same one as in production.

First approach implementation, applied only to run_details.

**I ignored the test TestListRuns_Pagination_WithSortingOnMetrics** because it generates an invalid query from the MySQL perspective (test was false-positive on SQLite).
**Below are two SQLite queries that will not work using the MySQL dialect**
[query 1 without where by dummymetric](https://github.com/user-attachments/files/20026582/mysql_no_aggregation.txt) contains the problem in the 
```sql
SELECT
  ...
  rd.dummymetric
FROM (...)
GROUP BY rd.UUID
```
Here rd.dummymetric is not included in the GROUP BY clause and is not aggregated, but it is used in the SELECT statement. will only work if the sql_mode is lowered for MySQL and the GROUP_BY_FULL mode is removed, because there is no aggregation for the dummymetric field in the query. Without this adjustment, MySQL will throw an error
this query was generated [here](https://github.com/kubeflow/pipelines/blob/master/backend/src/apiserver/storage/run_store_test.go#L335)

[mysql_and_where.txt](https://github.com/user-attachments/files/20026643/mysql_and_where.txt) doesn't work at all in MySQL. With exception ' Unknown column 'dummymetric' in 'where clause''
the second query was generated [here](https://github.com/kubeflow/pipelines/blob/master/backend/src/apiserver/storage/run_store_test.go#L345)

The SQLite alternatives to the above queries contain the same issues but still work because SQLite performs less strict query validation.

Also, just to double-check, I have reproduced the first query on my own Kubeflow Pipelines setup and it also threw the same error during the tests(dummymetric is just a random alias):
group by issue

Additionally, since I am creating a container and reusing it for performance reasons, I had to add logic for cleaning the database between tests, which complicated the initialization logic in run_details_test


backend/src/apiserver/storage/run_store_test.go
```diff
@@ -246,8 +340,8 @@ func TestListRuns_Pagination(t *testing.T) {
 }
 
 func TestListRuns_Pagination_WithSortingOnMetrics(t *testing.T) {
-	db, runStore := initializeRunStore()
-	defer db.Close()
+	t.Skip("Generate invalid sql query")
```
ntny, line 343: check the PR description for details

ntny:
Hello @HumairAK,
I found the [closed issue](https://github.com/kubeflow/pipelines/issues/8851) with a similar motivation described.
There are several merged commits mostly related to preparation for switching from SQLite testing to MySQL.

However, SQLite tests are still being used in api-server/storage, and it seems that [#8851](https://github.com/kubeflow/pipelines/issues/8851) is still relevant and the current PR partially closed it.
Could you please take a look and correct me if I'm wrong?

backend/src/apiserver/storage/mysql_fake.go
```diff
@@ -0,0 +1,83 @@
+// Copyright 2018 The Kubeflow Authors
```
HumairAK:
```suggestion
// Copyright 2025 The Kubeflow Authors
```

backend/src/apiserver/storage/run_store_test_setup.go
```diff
@@ -0,0 +1,113 @@
+// Copyright 2018 The Kubeflow Authors
```
HumairAK:
```suggestion
// Copyright 2025 The Kubeflow Authors
```

backend/src/apiserver/storage/mysql_fake.go
```diff
@@ -0,0 +1,83 @@
...
+// This behavior can be customized as described here:
+// https://golang.testcontainers.org/features/configuration/#customizing-ryuk-the-resource-reaper
+//
+// IMPORTANT:
+// If your tests run inside a container in your CI environment,
+// you need to provide access to the Docker runtime on the host,
+// typically by mounting the Docker socket (e.g., /var/run/docker.sock)
+// into the test container, so that Testcontainers can manage containers.
+
+func NewMySqlDBOrFatal(container *mysqlcontainer.MySQLContainer, ctx context.Context) *DB {
```
HumairAK: nit: per go coding convention, ctx should be first argument

backend/src/apiserver/storage/mysql_fake.go
```diff
@@ -0,0 +1,83 @@
...
+	gormDb, err := OpenMysqlDb(ctx, container)
+	if err != nil {
+		glog.Fatalf(err.Error())
+	}
+	db, err := ApplyMigration(gormDb)
+	return db
+}
+
+func LaunchMysqlContainer(ctx context.Context) (*mysqlcontainer.MySQLContainer, error) {
+	container, err := mysqlcontainer.Run(ctx, "mysql:8.0")
```
HumairAK: `mysql:8.0` - this feels easy too hardcoded, we should surface this as a flag argument, and default to a constant which can be "mysql:8.0"

backend/src/apiserver/storage/run_store_test_setup.go
```diff
@@ -0,0 +1,113 @@
...
+import (
+	"context"
+	"fmt"
+
+	"github.com/golang/glog"
+	"github.com/kubeflow/pipelines/backend/src/apiserver/model"
+	"github.com/testcontainers/testcontainers-go/modules/mysql"
+)
+
+type RuneStoreMysqlTestEnv struct {
```
HumairAK:
```suggestion
type RunStoreMySQLTestEnv struct {
```

HumairAK: I would similarly update instances of naming of mysql -> mySQL and Mysql -> MySQL

backend/src/apiserver/storage/run_store_test.go
```diff
@@ -43,12 +47,46 @@ func (r RunMetricSorter) Len() int           { return len(r) }
...
+	storeEnv.StopOrFatal(ctx)
+	os.Exit(code)
+}
+
+func initializeStorageAndCloseConnection() (*RunStore, error) {
+	/*
+		This function is used in fail-expected tests,
+		where we intentionally expect an error to be throw
+		due to operations on a closed database connection
+	*/
```
HumairAK: could you update this formatting to adhere to the go doc comment convention [ [1]](https://tip.golang.org/doc/comment#func)

backend/src/apiserver/storage/run_store_test.go
```diff
@@ -43,12 +47,46 @@ func (r RunMetricSorter) Len() int           { return len(r) }
...
+		return nil, fmt.Errorf("failed to close mysql connection: %v", err)
+	}
+	return runStore, nil
+}
+
+func initializeRunStore(db *DB) *RunStore {
+	if db == nil {
+		glog.Fatal("failed to initialize RunStore: db is nil")
+	}
+	storeEnv.CleanStorageOrFatal()
```
HumairAK: it feels unnecessary to pass `storeEnv.Db` as an argument when we then go on to access it globally (e.g. via `storeEnv.CleanStorageOrFatal()` in `initializeRunStore`), couldn't we just access `storeEnv.Db` in the same manner? 

if the reasoning to keep `db` as an arg so we keep `initializeRunStore` agnostic to the dialect, then to me it makes more sense to `storeEnv.CleanStorageOrFatal()` before calling `initializeRunStore`, and if we want to further keep the entire test set agnostic, then you could also opt to include a `StoreEnvironmentInterface` with a `Clean()` method.

backend/src/apiserver/storage/run_store_test_setup.go
```diff
@@ -0,0 +1,113 @@
...
+func NewRuneStoreMysqlSetupOrFatal(ctx context.Context) *RuneStoreMysqlTestEnv {
+	container, err := LaunchMysqlContainer(ctx)
+	if err != nil {
+		glog.Fatal("failed to create mysql container:", err)
+	}
+	gormDb := NewMySqlDBOrFatal(container, ctx)
+	return &RuneStoreMysqlTestEnv{Db: gormDb, MysqlContainer: container}
+}
+
+func (r *RuneStoreMysqlTestEnv) CleanStorageOrFatal() {
```
HumairAK: seems to me all the functions in this file can be unexported (starting with lowercase), since they are all used within the package

backend/src/apiserver/storage/run_store_test.go
HumairAK: If we went with a latter approach, we could also extend it to a sqllite implementation and allow the user to invoke the test optionally via flag/config

I do see some value in keeping sqlite testing as in the future we may want to utilize sqlite for a local kfp server experience. 

@mprahl / @nsingla any thoughts?

backend/src/apiserver/storage/run_store_test_setup.go
HumairAK: most of the implementation here feels pretty generic and re-usable, does it need to be scoped to Runs? it seems like if we update `cleanTables` to clean out all the tables [here](https://github.com/kubeflow/pipelines/pull/11889/files#diff-210f8bf0fba8e19d865e643c3c85bfb489835c6d13e60a1f42918b758fab485fR47), it seems - a lot of the logic can be generalized to be used for the other store tests no?

HumairAK (changes requested):

HumairAK:
@nsingla - are you able to also take a look at this pr?

backend/src/apiserver/storage/run_store_test.go
ntny: @HumairAK  I'll try to clarify :)

The reasoning for passing db is pretty straightforward — I want to reuse the GORM thread pool instead of creating a new one every time.

Yes, originally I planned to use the global storeEnv.Db.
But unfortunately, there are tests  that explicitly close the db connection and check that an error is returned on a closed connection.
If we use the global storeEnv.Db, those tests would close it and break everything else.
list of such tests:
- https://github.com/kubeflow/pipelines/blob/master/backend/src/apiserver/storage/run_store_test.go#L621
- https://github.com/kubeflow/pipelines/blob/master/backend/src/apiserver/storage/run_store_test.go#L687
- https://github.com/kubeflow/pipelines/blob/master/backend/src/apiserver/storage/run_store_test.go#L857
- https://github.com/kubeflow/pipelines/blob/master/backend/src/apiserver/storage/run_store_test.go#L1230
- https://github.com/kubeflow/pipelines/blob/master/backend/src/apiserver/storage/run_store_test.go#L1361

I'm not sure these tests are even that useful, but I want to keep the impact on existing tests minimal in this MR.

Basically, there are two options:

either create a new gorm.DB each time so tests don't interfere with each other,

or (like I'm doing now) only create a new db for the few tests that explicitly close it.

In both cases, I think it's simpler to pass db into initializeRunStore so we can reuse the populating runs logic.

backend/src/apiserver/storage/run_store_test_setup.go
ntny: I designed this class specifically for initializing and re-initializing RunStorage for testing. It knows which tables are involved and how to clean them up properly. 
To be honest, I’m a bit hesitant to extract cleanTables into a shared utility — there’s a non-zero chance someone might start using it outside of tests. Here it’s clearly part of a test setup and used only within this file, so its intent and scope are obvious. Maybe I’m just being paranoid :)

backend/src/apiserver/model/experiment.go
```diff
@@ -14,6 +14,10 @@
 
 package model
 
+func (Experiment) TableName() string {
```
nsingla, line 17: I am pretty sure there is a reason behind it, but curious to know why do we need this when we have https://github.com/kubeflow/pipelines/pull/11889/files#diff-5cc8bd9e86d20e984748f40a0eb53ebc9ca8b4b1d5b30924d4b0b781f6b9156fL66?

backend/src/apiserver/storage/run_store_test_setup.go
```diff
@@ -0,0 +1,123 @@
...
+	}
+	err = cleanTable(db, model.ResourceReference{})
+	if err != nil {
+		return err
+	}
+	return nil
+}
+
+func cleanTable(db *DB, table model.Tabler) error {
+	_, err := db.Exec(fmt.Sprintf("DELETE FROM %v", table.TableName()))
```
nsingla: Not super important since this is just for testing (not a production code), but this is prone to SQL injection. Standard way example: https://github.com/kubeflow/pipelines/blob/c03127d9671ed2ac29350840b4a56c5cf0c227eb/backend/src/apiserver/storage/run_store.go#L492

backend/src/apiserver/storage/run_store_test_setup.go
```diff
@@ -0,0 +1,123 @@
...
+	"flag"
+	"fmt"
+
+	"github.com/golang/glog"
+	"github.com/kubeflow/pipelines/backend/src/apiserver/model"
+	"github.com/testcontainers/testcontainers-go/modules/mysql"
+)
+
+var (
+	mysqlImage = flag.String("mysql-image", "mysql:8.0", "mysql image")
```
nsingla: we can probably leverage this to run these tests against other databases like PostgreSQL, so may be we should make this a bit generic so that its easy to expand in the future (without much refactoring)?

@HumairAK what do you think? Its probably out of scope for this work, but something to consider for the future

nsingla:
**Out of scope for this PR:** we should try to remove the hard coded values in assertions to use the created objects in the setup/initialize phase

backend/src/apiserver/storage/run_store_test_setup.go
HumairAK: right, but it seems like we're going to be repeating a lot of this logic when we add this for the other stores no? can we not just create a `test_util` package in `backend/src/apiserver/storage`, and make this re-usable for all the stores? from what I can tell the only difference here are the tables, we could just extend it to support all tables right [[1]](https://github.com/kubeflow/pipelines/blob/375226aed17be43898578a9b6ab9d6e5b02ff98e/backend/src/apiserver/storage/db_fake.go#L47)?

backend/src/apiserver/storage/run_store_test.go
HumairAK: okay, in that case I think this is fine, I would just either do one of the following: 

* provide function comments on `initializeRunStore` explaining why db is parameterized
* update the code logic to be more explicit in its intent, maybe something like: 
```go
func initializeRunStore(override *DB) *RunStore {
	var db *DB
	// Allow overriding db connection for testing
	// niche cases such as testing connection closures
	// these scenarios require a separate db connection
	// from the main thread pool.
	if override == nil {
		db = storeEnv.Db
	}
```

since the actual intended use case is that we re-use storeEnv.DB for majority of the scenarios iiuc

backend/src/apiserver/storage/run_store_test_setup.go
ntny: Good point. Addressed both comments

backend/src/apiserver/storage/run_store_test.go
mprahl: I'd also be +1 to having an interface where we default to SQLite (in memory preferred) for unit tests, but have the GitHub workflow test using MySQL or other database servers that we support.

nsingla (commented):
Looks like I never submitted by review comments :doh

kaikaila:
Hi @ntny @mprahl @HumairAK

Thanks for the discussion! I'm trying to understand the implications of the Testcontainers approach and had a few questions:

1. **Performance**: How long would unit tests take? both locally and in CI. Has anyone benchmarked this with Testcontainers vs SQLite?

2. **Migration scope**: I noticed this PR applies to run_details tests. Are we thinking this would eventually cover all storage unit tests, or would it be selective? 

3. **CI considerations**: Since Testcontainers needs Docker, I wanted to check if there are any known issues with container startup in GitHub Actions that we should be aware of.

I'll add the metric sorting test case (like TestListRuns_Pagination_WithSortingOnMetrics) to the integration tests in #12379   to ensure we catch these issues regardless of which approach we take.

Let me know what you think - happy to help either way!

nsingla:
I think test containers will provide good value for local development experience, as it should enable unit tests to be able to run with actual database instance.
There is a little overhead with running test containers in GitHub runners but we are already running Docker runners with sdk tests (although not in parallel)

Another value I see with that is that, we can save some workflow runs in CI (as the number of workflows in our CI are increasing and also save some CI time). We can also perform better edge case testing with local database instance as oppose to sqllite (with test containers) and will probably also enable easier performance benchmarking

ntny:
Hi @kaikaila, happy to share my thoughts.

Performance:
Yes, there is a trade-off. 
These tests are slower than in-memory tests, but still fairly fast. I haven’t done formal benchmarks comparing Testcontainers with SQLite, but based on my experience the API server storage tests run quickly enough for practical use. One important upside is that we immediately test against real databases and implicitly cover future DB migrations.
Developers can run these tests locally and catch MySQL/Postgres-specific issues before sending an MR to CI/CD. These tests are much easier to run locally than full integration tests, and in some cases this could even reduce overall CI time. The current in-memory tests, for example, don’t validate SQL syntax nuances at all.

Migration scope:
You’re right  this is not a full migration. This MR is mainly about introducing Testcontainers and validating whether this approach works well for the team. If it does, we can then consider migrating the rest of the storage tests incrementally.

CI considerations / drawbacks:
The main downside I see is that **developers who want to run storage tests locally will now need a container runtime** (Docker, Colima, etc.).

PR #11893, opened 2025-05-06
feat(sdk): validate component variables and function calls in compiler

dandawg:
KFP python-based components require that any functions and variables they use are either imported or defined within the component function. However, this was never validated, making it possible to define pipelines/components that pass formatting/linting checks and kfp compilation, but then fail during runtime.

When a pipeline fails during runtime, it can be a major problem as time and money are wasted on resources and developers are forced to troubleshoot otherwise preventable issues. This new validation solves the problem.

Note that the scope of this feature is not comprehensive validation. However, it does set a foundation for further work in this area.

I added variable and function call validation to compiler to check that variables and functions are either properly imported or defined within components.  Also added tests to make sure the new feature is working.

I added an optional `validate` parameter to `kfp.compiler.Compiler().compile` with default `False`. When set to `True`, the compiler validates that all variables and functions used by any component function in a pipeline, are either defined inside the function, or they are properly imported within their respective components.  The flag is meant to make the feature fully backwards compatible, with the idea that this can be further incorporated later (perhaps on the next major/minor release), by making the flag default to True.

For example, the following code would fail compilation (as expected), since `os` is imported outside of the component (the script is valid Python, so linting doesn't pick it up)
```python

import kfp
from kfp.dsl import component
import os

@component
def mycomp(var1: str):
  # os not imported here as it should be
  os.getcwd()

# compile will fail with validate=True
# saving developer headaches later
Compiler().compile(mycomp, 'mycomp.yaml', validate=True)
```
This PR also created error messaging. The error message when `python comp.py` is run will be:
```python
  return component_factory.create_component_from_func(
  File "/Users/ddowler/Code/Sandbox/temp-validate-component/comp.py", line 15, in <module>
    Compiler().compile(mycomp, 'mycomp.yaml', validate=True) 
    ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/ddowler/Code/Kubeflow/pipelines/sdk/python/kfp/compiler/compiler.py", line 87, in compile
    compiler_utils.validate_component(pipeline_func)
  File "/Users/ddowler/Code/Kubeflow/pipelines/sdk/python/kfp/compiler/compiler_utils.py", line 982, in validate_component
    validate_component_imports(component)
  File "/Users/ddowler/Code/Kubeflow/pipelines/sdk/python/kfp/compiler/compiler_utils.py", line 968, in validate_component_imports
    raise ValueError(
ValueError: Component mycomp uses functions that are neither defined nor imported: {'os.getcwd', 'print', 'os'}
```

Without this feature, it is possible to compile a component that uses functions defined outside its context; despite the compile succeeding, the component fails to run. This can be a huge problem, as a component with a bad function call could be several layers deep in a pipeline, and time and money are wasted before the error hits during runtime.

Without this feature (or setting `validate=False`), the above code example will compile successfully, but the component would fail at runtime since  the `os` module is not properly imported within the component code.
 
Note: I used Cursor to assist with some development. This helped in some areas, however, the AI struggled significantly, in particular with nested Python functions and their closure scopes, as well as with some specific KFP SDK understanding.

**Checklist:**

### DIFF
--- a/sdk/python/kfp/compiler/compiler_utils.py
+++ b/sdk/python/kfp/compiler/compiler_utils.py
@@ -13,16 +13,23 @@
 # limitations under the License.
 """Utility methods for compiler implementation that is IR-agnostic."""
 
+import ast
+import builtins
 import collections
 import copy
+import inspect
+import textwrap
 from typing import DefaultDict, Dict, List, Mapping, Set, Tuple, Union
 
 from kfp import dsl
+from kfp.dsl import base_component
 from kfp.dsl import constants
 from kfp.dsl import for_loop
+from kfp.dsl import graph_component
 from kfp.dsl import pipeline_channel
 from kfp.dsl import pipeline_context
 from kfp.dsl import pipeline_task
+from kfp.dsl import python_component
 from kfp.dsl import tasks_group
 
 GroupOrTaskType = Union[tasks_group.TasksGroup, pipeline_task.PipelineTask]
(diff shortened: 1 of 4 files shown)


sdk/python/kfp/compiler/compiler_utils.py
```diff
@@ -861,3 +868,188 @@ def _memory_to_float(memory: str) -> float:
         memory = float(memory) / constants._G
 
     return memory
+
+
+def validate_component_funcs_and_vars(
+        component: base_component.BaseComponent) -> None:
+    """Validates that all functions/variables used within a python-based
+    component are builtin, defined, or imported within the component."""
+    from kfp.dsl import python_component
```
zazulam, line 877: Do we need this import here since we have the import on [line 32?](https://github.com/kubeflow/pipelines/pull/11893/files#diff-1840ec3a2fcd81e1013e8db7d31f872f5560093eff00791bacad520bbd93ff82R32)

sdk/python/kfp/compiler/compiler_utils.py
```diff
@@ -861,3 +868,188 @@ def _memory_to_float(memory: str) -> float:
...
+
+    # Parse the source code into an AST
+    tree = ast.parse(source)
+
+    # Sets to track variables in the component function
+    defined = set()  # Variables defined within the function
+    imported = set()  # Modules and names imported in the function
+    used = set()  # All names referenced in the function
+
+    class Analyzer(ast.NodeVisitor):
```
zazulam, line 912: I see this is a helper class for traversing the ast, any reason why we have it defined within this function instead of in the module?

zazulam (commented):
Awesome stuff Dan, just had 2 questions

sdk/python/kfp/compiler/compiler_utils.py
dandawg: Good catch. I'll update.

sdk/python/kfp/compiler/compiler_utils.py
dandawg: I defined it there since it is tightly coupled to the function that uses it `validate_component_funcs_and_vars`. Personally, I don't have a strong preference, but I like that this makes it clear the class is only used by the one function. That being said, I could change it without to much of an issue. 

Here's the pros and cons that AI identified:

Pros:

- Encapsulation: The class is only used by validate_component_funcs_and_vars, so keeping it inside the function makes it clear that it's not intended for use elsewhere.
- Namespace Pollution: It doesn't pollute the module's namespace with a class that is only relevant to one function.
- Readability: For readers of the function, it's immediately clear that Analyzer is tightly coupled to the logic of this function.

Cons:

- Testing/Reuse: If you ever want to test or reuse Analyzer elsewhere, it's harder to access.
- Performance: The class is redefined every time the function is called, though in practice this overhead is negligible.

Let me know if you prefer I change it to the module level @zazulam .

zazulam: No worries, was just curious

dandawg:
I'm not sure why a test is failing after a PR that literally just removed one unused import line? @HumairAK did testing get updated? Any idea what I need to do to get it passing again?

PR #11910, opened 2025-05-12
chore(test): fixed log collection in tests

hbelmiro:
Some steps were not collecting logs in case of failures.
This PR fixes that.

**Checklist:**

### DIFF
--- a/.github/workflows/e2e-test.yml
+++ b/.github/workflows/e2e-test.yml
@@ -35,21 +35,27 @@ jobs:
           python-version: 3.9
 
       - name: Create KFP cluster
+        id: create-kfp-cluster
         uses: ./.github/actions/kfp-cluster
         with:
           k8s_version: ${{ matrix.k8s_version }}
+        continue-on-error: true
 
       - name: Forward API port
+        id: forward-api-port
+        if: ${{ (steps.create-kfp-cluster.outcome == 'success' )}}
         run: ./.github/resources/scripts/forward-port.sh "kubeflow" "ml-pipeline" 8888 8888
-        
+        continue-on-error: true
+
       - name: Initialization tests v1
         id: tests
+        if: ${{ (steps.forward-api-port.outcome == 'success' )}}
         working-directory: ./backend/test/initialization
         run: go test -v ./... -namespace kubeflow -args -runIntegrationTests=true
         continue-on-error: true
 
       - name: Collect failed logs
-        if: steps.tests.outcome != 'success'
+        if: ${{ (steps.create-kfp-cluster.outcome != 'success' ) || ( steps.forward-api-port.outcome != 'success' ) || ( steps.tests.outcome != 'success' )}}
         run: |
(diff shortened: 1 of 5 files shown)


ig-raiderler:
The performance optimizations here are quite clever. Nice engineering work!

.github/workflows/e2e-test.yml
```diff
@@ -35,21 +35,27 @@ jobs:
...
       - name: Initialization tests v1
         id: tests
+        if: ${{ (steps.forward-api-port.outcome == 'success' )}}
         working-directory: ./backend/test/initialization
         run: go test -v ./... -namespace kubeflow -args -runIntegrationTests=true
         continue-on-error: true
 
       - name: Collect failed logs
-        if: steps.tests.outcome != 'success'
+        if: ${{ (steps.create-kfp-cluster.outcome != 'success' ) || ( steps.forward-api-port.outcome != 'success' ) || ( steps.tests.outcome != 'success' )}}
```
alyssacgoins, line 58: I noticed that some blocks here (e.g. Forward API port and Initialization tests v1) add an id field whereas other blocks (e.g. Collect Failed logs) do not - is this because some tasks are not labeled with an id, or for some other reason?

.github/workflows/e2e-test.yml
```diff
@@ -77,21 +83,27 @@ jobs:
...
       - name: Initialization tests v2
         id: tests
+        if: ${{ (steps.forward-api-port.outcome == 'success' )}}
         working-directory: ./backend/test/v2/initialization
         run: go test -v ./... -namespace kubeflow -args -runIntegrationTests=true
         continue-on-error: true
 
       - name: Collect failed logs
-        if: steps.tests.outcome != 'success'
+        if: ${{ (steps.create-kfp-cluster.outcome != 'success' ) || ( steps.forward-api-port.outcome != 'success' ) || ( steps.tests.outcome != 'success' )}}
```
alyssacgoins, line 106: More out of curiosity -  what is the reasoning for this additional space between the parentheses "{" and "("?

.github/workflows/e2e-test.yml
hbelmiro: The id is needed when you reference that step in another step.
In this line, for example, we reference `create-kfp-cluster`, `forward-api-port` and `tests`. While `Collect Failed logs` is not referenced anywhere. Adding an id to it wouldn't make any difference.

.github/workflows/e2e-test.yml
hbelmiro: Just bad formatting :)
We could even remove those parentheses.
This is probably spread everywhere in the CI as I copied that `if` and just adjusted as needed.
We can open an issue to fix it later and ask for help. I think someone would be happy to work on it.

hbelmiro: Issue: https://github.com/kubeflow/pipelines/issues/11923

PR #11919, opened 2025-05-20
docs: add GoLand-specific configs to backend Readme

alyssacgoins:
Adds a GoLand-specific configuration for remotely debugging the driver.  This configuration includes an updated environment variable V2_DRIVER_COMMAND as well as the runtime arguments utilized when launching the api server  normally. 

**Checklist:**

### DIFF
--- a/backend/README.md
+++ b/backend/README.md
@@ -35,7 +35,7 @@ docker build -f backend/Dockerfile . --tag $API_SERVER_IMAGE
 ```
 ### Deploying the APIServer (from the image you built) on Kubernetes
 
-First, push your image to a registry that is accessible from your Kubernetes cluster. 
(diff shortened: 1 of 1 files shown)


backend/README.md
```diff
@@ -246,6 +246,41 @@ command to use Delve and the Driver image to use debug image built previously.
 }
 ```
 
+Use the following configuration if using GoLand:
```
HumairAK, line 249: I don't think Goland has a json configuration like vscode right? If so, I think it would be better to just list out the steps explicitly like: 

1. Create a new run configuration
2. Create a new Go Build 
3. Provide the following fields: 
    * Name: blah
    * Environment Variables: etc.

and so on

alyssacgoins: Thanks for pointing that out! Will update

PR #11925, opened 2025-05-21
- fix(launcher): missing executorInput parameter values caused by {{$}} evaluation order

ntny:

The getPlaceholders method [prepends](https://github.com/kubeflow/pipelines/blob/master/backend/src/v2/component/launcher_v2.go#L752) `placeholders["{{$}}"] = string(executorInputJSON) `to the placeholders map to put the full executorInputJSON as a template.

The [code](https://github.com/kubeflow/pipelines/blob/master/backend/src/v2/component/launcher_v2.go#L376) assumes that the {{$}} key will be processed first, since it inserts the entire JSON template into the argument string. Only after this replacement are individual placeholders (like {{$.inputs.parameters['some_param']}}) applied.

However, since Go maps [do not guarantee](https://go.dev/blog/maps#iteration-order) iteration order, the {{$}} key is often processed after other keys  As a result, some or all parameter values may be randomly missing in the final output.
Reproduced in our environment
Снимок экрана 2025-05-22 в 11 08 48

This PR fixes the issue by ensuring that the {{$}} placeholder is processed before other keys, preserving the correct substitution order.

### DIFF
--- a/backend/src/v2/component/launcher_v2.go
+++ b/backend/src/v2/component/launcher_v2.go
@@ -364,26 +364,16 @@ func executeV2(
 	}
 
 	// Fill in placeholders with runtime values.
-	placeholders, err := getPlaceholders(executorInputWithDefault)
+	compiledCmd, compiledArgs, err := compileCmdAndArgs(executorInputWithDefault, cmd, args)
 	if err != nil {
 		return nil, nil, err
-	}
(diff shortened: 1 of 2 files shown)


mprahl (approved):
/lgtm
/approve

Good catch. Thanks for the contribution!

PR #11269, opened 2024-10-03
feat(sdk): Implement Pipeline Configuration with TTL

rimolive:
**Description of your changes:**
This PR introduces a new class `PipelineConfig` in KFP SDK to add pipeline-specific configurations with the first one: Pipeline TTL.

### SDK
* Create a Python virtualenv and install the SDK and IR YAML API packages locally
```
$ python -m venv .venv
$ pushd api
$ make clean-python python
$ popd
$ source .venv/bin/activate
$ pip install wheel setuptools protobuf grpcio grpcio-tools
$ pip install -r sdk/python/requirements-dev.txt
$ pip install -e api/v2alpha1/python
$ pip install -e sdk/python
```
* Use the [example code](https://gist.github.com/rimolive/e0c31ccad1f6c963ec010c455b424afe) to compile
```
$ kfp dsl compile --py main.py --output main.yaml
```
* You should be able to compile and find the following snippet in the `main.yaml` file:
```
---
platforms:
  kubernetes:
    pipelineConfig:
      resourceTtl: 60
```

**Checklist:**
- [X] You have [signed off your commits](https://www.kubeflow.org/docs/about/contributing/#sign-off-your-commits)
- [X] The title for your pull request (PR) should follow our title convention. [Learn more about the pull request title convention used in this repository](https://github.com/kubeflow/pipelines/blob/master/CONTRIBUTING.md#pull-request-title-convention).

PR #11934, opened 2025-05-27
feat(sdk): update PipelineConfig to reflect new workspace Protobuf changes

VaniHaripriya:

 This PR updates the dsl.PipelineConfig class to support configuring a pipeline workspace using the newly introduced WorkspaceConfig and KubernetesWorkspaceConfig message types in pipeline_spec.proto

**Checklist:**

### DIFF
--- a/sdk/python/kfp/compiler/pipeline_spec_builder.py
+++ b/sdk/python/kfp/compiler/pipeline_spec_builder.py
@@ -2063,11 +2063,14 @@ def write_pipeline_spec_to_file(
 
 def _merge_pipeline_config(pipelineConfig: pipeline_config.PipelineConfig,
                            platformSpec: pipeline_spec_pb2.PlatformSpec):
-    # TODO: add pipeline config options (ttl, semaphore, etc.) to the dict
-    # json_format.ParseDict(
-    #     {'pipelineConfig': {
-    #         '<some pipeline config option>': pipelineConfig.<get that value>,
-    #     }}, platformSpec.platforms['kubernetes'])
+    workspace = pipelineConfig.workspace
+    if workspace is None:
+        return platformSpec
+
+    json_format.ParseDict(
+        {'pipelineConfig': {
+            'workspace': workspace.get_workspace(),
+        }}, platformSpec.platforms['kubernetes'])
 
     return platformSpec
 
--- a/sdk/python/kfp/dsl/pipeline_config.py
+++ b/sdk/python/kfp/dsl/pipeline_config.py
@@ -13,11 +13,61 @@
 # limitations under the License.
 """Pipeline-level config options."""
 
+from typing import Any, Dict, Optional
+
+
+class KubernetesWorkspaceConfig:
+    """Configuration for Kubernetes-specific workspace settings.
+
+    Use this to override the default PersistentVolumeClaim (PVC) configuration
+    used when running pipelines on a Kubernetes cluster.
+
+    Attributes:
+        pvcSpecPatch: A dictionary of fields to patch onto the default PVC spec
+                      (e.g., 'storageClassName', 'accessModes').
+    """
+
+    def __init__(self, pvcSpecPatch: Optional[Dict[str, Any]] = None):
+        self.pvcSpecPatch = pvcSpecPatch or {}
+
+    def set_pvcSpecPatch(self, patch: Dict[str, Any]):
+        self.pvcSpecPatch = patch
+
+
+class WorkspaceConfig:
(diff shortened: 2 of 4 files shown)


mprahl:
@VaniHaripriya you'll need to update `_merge_pipeline_config` similar to what @rimolive 's PR is doing here:
https://github.com/kubeflow/pipelines/pull/11269/files#diff-42a6e18d166f33d4a0992ab9609966d24389846045972a99e9f286bba08d4014R2064-R2071

sdk/python/kfp/compiler/pipeline_spec_builder.py
```diff
@@ -2063,12 +2063,24 @@ def write_pipeline_spec_to_file(
 
 def _merge_pipeline_config(pipelineConfig: pipeline_config.PipelineConfig,
                            platformSpec: pipeline_spec_pb2.PlatformSpec):
-    # TODO: add pipeline config options (ttl, semaphore, etc.) to the dict
-    # json_format.ParseDict(
-    #     {'pipelineConfig': {
-    #         '<some pipeline config option>': pipelineConfig.<get that value>,
-    #     }}, platformSpec.platforms['kubernetes'])
+    workspace = pipelineConfig.workspace
```
mprahl, line 2066: Do you need to guard against `pipelineConfig.workspace` being `None`/not set below?

sdk/python/kfp/dsl/pipeline_config.py
```diff
@@ -13,11 +13,35 @@
...
+    def __init__(self, pvc_spec_patch: Optional[Dict[str, Any]] = None):        
+        self.pvc_spec_patch = pvc_spec_patch or {}
+
+    def set_pvc_spec_patch(self, patch: Dict[str, Any]):        
+        self.pvc_spec_patch = patch
+
+class WorkspaceConfig:
+    """Configuration for a shared workspace that persists during the pipeline run."""
+
+    def __init__(self, size: str = "", kubernetes: Optional[KubernetesWorkspaceConfig] = None):
```
mprahl, line 30: I think we should make `size` required.
```suggestion
    def __init__(self, size: str, kubernetes: Optional[KubernetesWorkspaceConfig] = None):
```

sdk/python/kfp/compiler/pipeline_spec_builder.py
```diff
@@ -2063,11 +2063,21 @@ def write_pipeline_spec_to_file(
...
+        return platformSpec
+
+    workspace_dict = {
+        'size': workspace.size,
+    }
+
+    if workspace.kubernetes:
+        workspace_dict['kubernetes'] = {
+            'pvcSpecPatch': workspace.kubernetes.pvc_spec_patch
+        }
```
gmfrasca: I wonder if we should move this patch dict construction into the WorkspaceConfig class itself, such that we would simply call something like `workspace_dict = workspace.get_spec()` which returns this dict.  It would consolidate construction/formatting into the class itself in case we want to expand it later.

That said, this is splitting hairs so strictly optional/not worth blocking PR on

sdk/python/kfp/dsl/pipeline_config.py
```diff
@@ -13,11 +13,56 @@
...
+
+    def set_kubernetes_config(self,
+                              kubernetes_config: KubernetesWorkspaceConfig):
+        self.kubernetes = kubernetes_config
+
 
 class PipelineConfig:
     """PipelineConfig contains pipeline-level config options."""
 
     def __init__(self):
```
mprahl: You'll need to update the constructors to match the design of being able to do this:

```python
@dsl.pipeline(
    name="my-pipeline",
    pipeline_config=dsl.PipelineConfig(
        workspace=dsl.WorkspaceConfig(
          size="250GB",
          kubernetes=dsl.KubernetesWorkspaceConfig(
            pvcSpecPatch={
              "storageClassName": "super-fast-storage",
              "accessModes": ["ReadWriteMany"],
            }
          ),
        ),
    ),
)
```

Design reference:
https://github.com/kubeflow/pipelines/tree/master/proposals/11875-pipeline-workspace#1-pipeline-configuration

sdk/python/kfp/compiler/pipeline_spec_builder.py
mprahl: I think that's a good idea. What do you think @VaniHaripriya ?

VaniHaripriya: That's a good idea @gmfrasca . I will update the  WorkspaceConfig class.

PR #11952, opened 2025-06-02
feat(backend/frontend): Add the name field for pipelines and pipeline versions

mprahl:

The API only accepted display_name, which accepted arbitrary characters but this does not work well in Kubernetes native API mode (Kubernetes pipeline store) since object names must be a valid DNS subdomain.

This commit adds the name field which is required, but for backwards compatibility, the API will accept just name or display_name and duplicate the set value to the unset field. For existing users, they will only see that there is an extra field of "name" on pipelines and pipeline versions that is the same value as "display_name".

The database migration logic adds a new DisplayName column and sets the values to what was in the Name column. Confusingly, the API field "display_name" used to map the a database column of "Name".

The UI now handles both name and display name but for a simpler user experience, the display name form fields when uploading a pipeline are hidden when not in Kubernetes native API mode (Kubernetes pipeline store) since there is no need to have a display name since there is no naming restrictions on the name.

**Checklist:**


backend/test/v2/test_utils.go
```diff
@@ -167,6 +169,30 @@ func DeleteAllPipelineVersions(client *api_server.PipelineClient, t *testing.T,
...
+	for {
+		select {
+		case <-ctx.Done():
+			// This should be a ContextDeadlineExceeded error and causes the test to fail.
+			require.Nil(t, ctx.Err(), "Pipeline versions have not been deleted after 5 seconds")
+
+			return
+		default:
+			pipelineVersions, _, _, err = ListPipelineVersions(client, pipelineId)
+			assert.Nil(t, err)
```
hbelmiro: I'm not sure I entirely understand this.
Do we really want the test to fail in case of an error? Wouldn't be better to just log the error and keep waiting?

mprahl: Yeah, that's fair. I can do that.

frontend/src/pages/PipelineVersionList.tsx
```diff
@@ -82,15 +86,15 @@ class PipelineVersionList extends React.PureComponent<
               this.props.pipelineId,
             ).replace(':' + RouteParams.pipelineVersionId, props.id)}
           >
-            {props.value}
+            {props.value?.display_name}
```
Gkrumbach07: should this default to name if display name is not defined?

frontend/src/pages/PipelineVersionList.tsx
```diff
@@ -68,11 +68,15 @@ class PipelineVersionList extends React.PureComponent<
     };
   }
 
-  public _nameCustomRenderer: React.FC<CustomRendererProps<string>> = (
-    props: CustomRendererProps<string>,
-  ) => {
+  public _nameCustomRenderer: React.FC<
+    CustomRendererProps<{ display_name?: string; name?: string }>
```
Gkrumbach07: shouldn't name always be defined

backend/src/apiserver/client_manager/client_manager.go
```diff
@@ -344,6 +371,20 @@ func InitDBClient(initConnectionTimeout time.Duration) *storage.DB {
 		}
 	}
 
+	if db.HasTable(&model.Pipeline{}) {
```
HumairAK, line 372: is there ever a scenario where we don't have this table? would it be better turn this into a validation check instead and we error out if the table does not exist? (same with version)

backend/src/apiserver/client_manager/client_manager.go
```diff
@@ -344,6 +371,20 @@ func InitDBClient(initConnectionTimeout time.Duration) *storage.DB {
...
+		scope := db.NewScope(&model.Pipeline{})
```
HumairAK, line 373: it's weird to me that newscope does not return an error
I see there is a `scope.HasError()` , could that help here?

backend/src/apiserver/client_manager/client_manager.go
```diff
@@ -320,6 +320,33 @@ func (c *ClientManager) Close() {
...
+			"ALTER TABLE " + quotedTableName + " ADD COLUMN DisplayName VARCHAR(255);",
+		).Exec()
+		scope.Raw("UPDATE " + quotedTableName + " SET DisplayName = Name").Exec()
+		scope.Raw("ALTER TABLE " + quotedTableName + " ALTER COLUMN DisplayName SET NOT NULL").Exec()
+	}
+
+	scope.CommitOrRollback()
+
+	if scope.HasError() {
+		glog.Fatalf("Failed to add DisplayName column to pipelines table. Error(s): %v", db.GetErrors())
```
HumairAK: optional: my preference is that if we are going to crash we should do it higher up the stack if possible, so ideally we return an error here and crash in `InitDBClient`

backend/src/apiserver/server/api_converter.go
```diff
@@ -146,21 +146,35 @@ func toApiExperiments(experiments []*model.Experiment) []*apiv2beta1.Experiment
...
+		name = apiPipeline.GetName()
+		displayName = apiPipeline.GetDisplayName()
 		description = apiPipeline.GetDescription()
 	default:
 		return nil, util.NewUnknownApiVersionError("Pipeline", p)
 	}
+
+	// Previously display_name was the required API field and name didn't exist. Name is now the required API field
+	// but if display_name is provided, we use it as the name.
+	if name == "" {
```
HumairAK, line 167: might be good to enforce that at least one of them is set here as well

backend/src/apiserver/server/pipeline_server.go
```diff
@@ -1044,9 +1063,14 @@ func (s *PipelineServer) canAccessPipeline(ctx context.Context, pipelineId strin
 // This method extract the common logic of naming the pipeline.
```
HumairAK:
```suggestion
// buildPipelineName extracts the common logic of naming the pipeline.
```

frontend/src/pages/PipelineVersionList.tsx
mprahl: Good point. The v2 API currently forces name and display_name to have a value for backwards compatibility but the v1 API only uses name. I'll change it here.

frontend/src/pages/PipelineVersionList.tsx
mprahl: Good point. Fixed!

tools/k8s-native/migration.py
```diff
@@ -132,16 +132,19 @@ def convert_to_k8s_format(pipeline, pipeline_versions, add_prefix, namespace):
...
         }
     }
     k8s_objects.append(pipeline_obj)
 
      # Create a PipelineVersion object for each version
     for i, version in enumerate(versions):
-        version_display_name = version.get("display_name", f"v{i}")
-        pipeline_version_name = get_version_name(pipeline_name, version_display_name, i, add_prefix)        
+        version_name = version.get("display_name", f"v{i}")
+        version_display_name = version.get("display_name", version_name)
```
HumairAK: how come we're not also checking for `name` now? if that is present, isn't that what we want to use? 
similar to the `pipeline_name` above?

backend/src/apiserver/client_manager/client_manager.go
mprahl: @HumairAK the logic is that if the table is not there, then this is a new database and we don't need to perform this migration. We can just rely on `db.AutoMigrate` below which will create the table with the correct columns and constraints.

backend/src/apiserver/client_manager/client_manager.go
mprahl: It's essentially just returning a clone from the `db` instance with a scope to the table. I suppose it could have verified that the table is a valid scope though but it doesn't. `scope.HasError()` I think only gets populated after queries, which is why it's used in `addDisplayNameColumn`.

backend/src/apiserver/server/api_converter.go
mprahl: I thought about that too but it seems that this is called before `buildPipelineName` to allow defaulting to the file name of the pipeline.

tools/k8s-native/migration.py
mprahl: Good catch. Thanks!

backend/src/apiserver/server/api_converter.go
HumairAK: right yeah, saw that later down the road, kk

Issue #10917, opened 2024-06-17
[bug] {{workflow.status}} and {{workflow.failures}} Behavior in KFP v2 dsl.ExitHandler versus v1


tom-pavz:
### Environment

*  How do you deploy Kubeflow Pipelines (KFP)?
`deployKF`
*  KFP version:
`2.1.0`
*  KFP SDK version:
`2.7.0`

### Steps to reproduce

We use `dsl.ExitHanlder` plus Argo Workflow's Workflow Variables to automatically report failed pipeline runs to slack. This was working well in KFP v1, but no longer in v2.

I have a simple hello-world pipeline that has the following `fail` component wrapped in a `dsl.ExitHanlder` with an `exit_task` of `post_msg_to_slack_on_pipeline_fail`.

```
@component
def fail():
    import sys

    sys.exit(1)
```

```
@component
def post_msg_to_slack_on_pipeline_fail():
    import json

    if "{{workflow.status}}" == "Failed":
        json_object = json.loads('{{workflow.failures}}'[1:-1])
        json_formatted_str = json.dumps(json_object, indent=2)
        # send alert to slack
```

```
@pipeline
def pipeline():
    post_msg_to_slack_on_pipeline_fail_task = post_msg_to_slack_on_pipeline_fail_op()

    with ExitHandler(exit_task=post_msg_to_slack_on_pipeline_fail_task):
        hello_task = hello_world_op()
        fail_task = fail_op()
        fail_task.after(hello_task)
```

Starting in KFP v2, at the time of the `exit_task` runtime, `{{workflow.status}}` gets a value of `Running` and the `{{workflow.failures}}` does not get any value and just prints out `{{workflow.failures}}` itself. It seems that for some reason in KFP V2, the failure of the task does not propagate up and out to the pipeline itself by the time the `exit_task` is running. In addition, I notice that the pipeline itself has an "Executed Successfully" status in the UI (see below screenshots of the DAG and sub-DAG), even though one of its tasks failed, which does not seem correct to me. I also notice that in the UI the `exit-handler-1` sub-DAG is stuck in "Running" status, which also seems incorrect.

### Expected result
The expected result, and the behavior we are getting when using v1 of KFP, was that at the time of the `exit_task` runtime, `{{workflow.status}}` got a value of `Failed` and the `{{workflow.failures}}` got a json file with information about the failed tasks of the pipeline, which we could then send to slack.

### Materials and reference

Screenshots of the pipeline and its DAGs' in the UI:
[...]

PR #11953, opened 2025-06-02
feat(backend/sdk): support PipelineTaskFinalStatus input

alyssacgoins:
Add support for exit handlers to take input PipelineTaskFinalStatus.

fixes https://github.com/kubeflow/pipelines/issues/10917

**Checklist:**

### DIFF
--- a/backend/src/v2/compiler/argocompiler/dag.go
+++ b/backend/src/v2/compiler/argocompiler/dag.go
@@ -630,8 +630,6 @@ func addImplicitDependencies(dagSpec *pipelinespec.DagSpec) error {
 				if err := addDep(input.GetTaskOutputParameter().GetProducerTask()); err != nil {
 					return wrap(err)
 				}
-			case *pipelinespec.TaskInputsSpec_InputParameterSpec_TaskFinalStatus_:
-				return wrap(fmt.Errorf("task final status not supported yet"))
 			default:
 				// other parameter input types do not introduce implicit dependencies
 			}
--- a/backend/src/v2/driver/driver_test.go
+++ b/backend/src/v2/driver/driver_test.go
@@ -15,7 +15,10 @@ package driver
 
 import (
 	"encoding/json"
+	"fmt"
 	"testing"
+
+	"github.com/stretchr/testify/require"
 
 	"github.com/kubeflow/pipelines/backend/src/apiserver/config/proxy"
 
@@ -634,4 +637,91 @@ func Test_initPodSpecPatch_resourceRequests(t *testing.T) {
 			}
 		})
 	}
+}
+
+func Test_initPodSpecPatch_inputTaskFinalStatus(t *testing.T) {
+	proxy.InitializeConfigWithEmptyForTests()
+	containerSpec := &pipelinespec.PipelineDeploymentConfig_PipelineContainerSpec{
+		Image:   "python:3.9",
+		Command: []string{"sh", "-ec", "python3 -m kfp.components.executor_main"},
+		Args:    []string{"--executor-input", "{{$}}", "--function_to_execute", "exit-op"},
+	}
+	componentSpec := &pipelinespec.ComponentSpec{
+		Implementation: &pipelinespec.ComponentSpec_ExecutorLabel{ExecutorLabel: "exec-exit-op"},
+		InputDefinitions: &pipelinespec.ComponentInputsSpec{
+			Parameters: map[string]*pipelinespec.ComponentInputsSpec_ParameterSpec{
+				"status": {ParameterType: pipelinespec.ParameterType_TASK_FINAL_STATUS},
+			},
+		},
+	}
+	finalStatusStruct, err := structpb.NewStruct(map[string]interface{}{
+		"state":                   "test-state",
+		"pipelineTaskName":        "test-pipeline-task-name",
+		"pipelineJobResourceName": "test-job-resource-name",
+		"error":                   map[string]interface{}{},
+	})
+	executorInput := &pipelinespec.ExecutorInput{
+		Inputs: &pipelinespec.ExecutorInput_Inputs{
+			ParameterValues: map[string]*structpb.Value{
+				"status": {
+					Kind: &structpb.Value_StructValue{
+						StructValue: finalStatusStruct,
+					},
+				},
+			},
+		},
+	}
+
+	podSpec, err := initPodSpecPatch(
+		containerSpec,
+		componentSpec,
+		executorInput,
+		27,
+		"test",
+		"0254beba-0be4-4065-8d97-7dc5e3adf300",
+		"1",
+		"false",
+		"false",
+	)
+	require.Nil(t, err)
+
+	expectedExecutorInput := map[string]interface{}{
+		"inputs": map[string]interface{}{
+			"parameterValues": map[string]interface{}{
+				"status": map[string]interface{}{
+					"error":                   map[string]interface{}{},
+					"pipelineJobResourceName": "test-job-resource-name",
+					"pipelineTaskName":        "test-pipeline-task-name",
+					"state":                   "test-state",
+				},
+			},
+		},
+	}
+	expectedComponentSpec := map[string]interface{}{
+		"executorLabel": "exec-exit-op",
+		"inputDefinitions": map[string]interface{}{
+			"parameters": map[string]interface{}{
+				"status": map[string]interface{}{
+					"parameterType": "TASK_FINAL_STATUS",
+				},
+			},
+		},
+	}
+	actualExecutorInput := map[string]interface{}{}
+	actualComponentSpec := map[string]interface{}{}
+
+	for i, arg := range podSpec.Containers[0].Command {
+		if arg == "--executor_input" {
+			err := json.Unmarshal([]byte(podSpec.Containers[0].Command[i+1]), &actualExecutorInput)
+			fmt.Println(podSpec.Containers[0].Command[i+1])
+			require.Nil(t, err)
+		}
+		if arg == "--component_spec" {
+			err := json.Unmarshal([]byte(podSpec.Containers[0].Command[i+1]), &actualComponentSpec)
+			require.Nil(t, err)
+		}
+	}
+
+	assert.Equal(t, expectedExecutorInput, actualExecutorInput)
+	assert.Equal(t, expectedComponentSpec, actualComponentSpec)
 }
--- a/backend/src/v2/driver/resolve.go
+++ b/backend/src/v2/driver/resolve.go
@@ -24,6 +24,7 @@ import (
 	"github.com/kubeflow/pipelines/api/v2alpha1/go/pipelinespec"
 	"github.com/kubeflow/pipelines/backend/src/v2/expression"
 	"github.com/kubeflow/pipelines/backend/src/v2/metadata"
+	"google.golang.org/genproto/googleapis/rpc/status"
 	"google.golang.org/protobuf/encoding/protojson"
 	"google.golang.org/protobuf/types/known/structpb"
 )
@@ -211,7 +212,7 @@ func resolveInputs(
 						return typeMismatch("list")
 					}
 				case *structpb.Value_StructValue:
-					if spec.GetParameterType() != pipelinespec.ParameterType_STRUCT {
+					if (spec.GetParameterType() != pipelinespec.ParameterType_STRUCT) && (spec.GetParameterType() != pipelinespec.ParameterType_TASK_FINAL_STATUS) {
 						return typeMismatch("struct")
 					}
 				default:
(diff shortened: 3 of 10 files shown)


samples/v2/pipeline_with_input_status_state.py
```diff
@@ -0,0 +1,28 @@
...
+
+@dsl.component
+def exit_op(user_input: str, status: dsl.PipelineTaskFinalStatus):
+    """Prints pipeline run status."""
+    print(user_input)
+    print('Pipeline status: ', status.state)
+    print('Job resource name: ', status.pipeline_job_resource_name)
+    print('Pipeline task name: ', status.pipeline_task_name)
+    print('Error code: ', status.error_code)
+    print('Error message: ', status.error_message)
```
HumairAK: can we make these asserts so we can test the results are what we expect?

samples/v2/pipeline_with_input_status_state.py
```diff
@@ -0,0 +1,28 @@
...
+
+@dsl.component
+def exit_task():
```
HumairAK: the naming is a bit confusing here, as this is not really an task that runs on exit, but a task that runs within the exit dag, maybe let's call it something like "some_task" instead of "exit_task"

backend/src/v2/driver/resolve.go
```diff
@@ -378,8 +379,24 @@ func resolveInputParameter(
...
-	// case *pipelinespec.TaskInputsSpec_InputParameterSpec_TaskFinalStatus_:
+	case *pipelinespec.TaskInputsSpec_InputParameterSpec_TaskFinalStatus_:
+		finalStatus, err := resolveProducerTaskFinalStatus(ctx, dag, pipeline, mlmd, paramSpec)
+		if err != nil {
+			return nil, err
+		}
+
+		finalStatusMap := map[string]interface{}{
+			"state":                   finalStatus.String(),
+			"pipelineTaskName":        opts.Task.GetTaskInfo().GetName(),
```
mprahl: This should be the producer task. This looks to be the current task.

backend/src/v2/compiler/argocompiler/dag.go
```diff
@@ -631,7 +631,9 @@ func addImplicitDependencies(dagSpec *pipelinespec.DagSpec) error {
 					return wrap(err)
 				}
 			case *pipelinespec.TaskInputsSpec_InputParameterSpec_TaskFinalStatus_:
-				return wrap(fmt.Errorf("task final status not supported yet"))
+				if err := addDep(input.GetTaskFinalStatus().GetProducerTask()); err != nil {
```
mprahl: Is `addDep` required since my understanding is that a task that uses `dsl.PipelineTaskFinalStatus` must be an exit task? Exit tasks are handled specially with Argo Workflow exit handlers.

backend/src/v2/driver/resolve.go
```diff
@@ -378,8 +379,24 @@ func resolveInputParameter(
...
 		}
-	// TODO(Bobgy): implement the following cases
-	// case *pipelinespec.TaskInputsSpec_InputParameterSpec_TaskFinalStatus_:
+	case *pipelinespec.TaskInputsSpec_InputParameterSpec_TaskFinalStatus_:
+		finalStatus, err := resolveProducerTaskFinalStatus(ctx, dag, pipeline, mlmd, paramSpec)
+		if err != nil {
+			return nil, err
+		}
+
+		finalStatusMap := map[string]interface{}{
```
mprahl: Can we use the `PipelineTaskFinalStatus` struct directly?

mprahl: Actually, why not just continue using `finalStatus`?

backend/src/v2/driver/resolve.go
HumairAK: derp yes @alyssacgoins this is correct, you can get the producer from `resolveProducerTaskFinalStatus`

alyssacgoins: Fixed :+1:

backend/src/v2/compiler/argocompiler/dag.go
alyssacgoins: Good point - I hadn't considered that exit handlers have special logic. Removed this line.

backend/src/v2/driver/resolve.go
alyssacgoins: resolveInputParameter returns a structpb.Value because ultimately pipelinespec.ExecutorInput_Inputs.ParameterValues has the type map[string]*structpb.Value. 
Structpb.Value does not take type structpb.PipelineTaskFinalStatus (it only takes double, string bool struct and ListValue), so instead of using structpb.PipelineTaskFinalStatus directly, I represented the value as a StructValue. 
I ended up moving the resolveProducerTaskFinalStatus() work to resolveInputParameter() because I think the logic is easier to follow that way.

samples/v2/pipeline_with_input_status_state.py
alyssacgoins: Updated

samples/v2/pipeline_with_input_status_state.py
alyssacgoins: Updated

alyssacgoins:
Fixes https://github.com/kubeflow/pipelines/issues/10917

backend/src/v2/driver/resolve.go
```diff
@@ -378,8 +379,22 @@ func resolveInputParameter(
...
+		producer, ok := tasks[opts.Task.DependentTasks[0]+"_"+strconv.FormatInt(dag.Execution.GetID(), 10)]
+		if !ok {
+			return nil, fmt.Errorf("producer task, %v, not in tasks", producer.TaskName())
+		}
+		finalStatusStruct, err := structpb.NewStruct(map[string]interface{}{
+			"state":                   producer.GetExecution().GetLastKnownState().String(),
+			"pipelineTaskName":        producer.TaskName(),
+			"pipelineJobResourceName": opts.RunName,
+			"error":                   map[string]interface{}{},
+		})
```
mprahl: Optional and untested: It's less efficient but I think this is safer since we are for sure using the correct fields and it's rendered in a struct that maps to correct JSON fields. I'll leave it up to you to decide which you prefer.

```suggestion
		finalStatus := pipelinespec.PipelineTaskFinalStatus{
			State: producer.GetExecution().GetLastKnownState().String(),
			PipelineTaskName: producer.TaskName(),
			PipelineJobResourceName: opts.RunName,
		}
		finalStatusJSON, err := json.Marshal(finalStatus)
		if err != nil {
			return nil, fmt.Errorf("failed to marshal PipelineTaskFinalStatus: %w", err)
		}

		var finalStatusMap map[string]interface{}
		if err := json.Unmarshal(finalStatusJSON, &finalStatusMap); err != nil {
			return nil, fmt.Errorf("failed to unmarshal JSON of PipelineTaskFinalStatus: %w", err)
		}

		// Convert the finalStatusMap to a structpb.Struct
		finalStatusStruct, err := structpb.NewStruct(finalStatusMap)
		if err != nil {
			return nil, fmt.Errorf("failed to create structpb.Struct: %w", err)
		}
		
		return structpb.NewStructValue(finalStatusStruct), nil
```

backend/src/v2/driver/resolve.go
```diff
@@ -378,8 +379,22 @@ func resolveInputParameter(
...
 			return nil, paramError(fmt.Errorf("param runtime value spec of type %T not implemented", t))
 		}
-	// TODO(Bobgy): implement the following cases
-	// case *pipelinespec.TaskInputsSpec_InputParameterSpec_TaskFinalStatus_:
+	case *pipelinespec.TaskInputsSpec_InputParameterSpec_TaskFinalStatus_:
+		tasks, err := getDAGTasks(ctx, dag, pipeline, mlmd, nil)
+		if err != nil {
+			return nil, err
+		}
+		producer, ok := tasks[opts.Task.DependentTasks[0]+"_"+strconv.FormatInt(dag.Execution.GetID(), 10)]
```
mprahl: Do you need to check the length of `DependenTasks` before using it? I don't see a check earlier in the code.

Also, I think the string formatting is abstracted in `metadata.GetTaskNameWithDagID`. Could you please reuse that function?

alyssacgoins: Updated

backend/src/v2/driver/resolve.go
alyssacgoins: I agree this is a safer approach - implemented the above code. Working out one issue - the "Error" field of PipelineTaskFinalStatus has to be set in order for sdk/python/kfp/dsl/executor (line 331) to process a PipelineTaskFinalStatus input, because each error field (error_code & error_message) are processed as `error_code=value.get('error').get('code', None) `- if the "Error" field is not set it causes a nil pointer error. 
However, setting` Error: &status.Status`  in PipelineTaskFinalStatus causes `json.Marshal(finalStatus)` to fail trying to encode the status type.

backend/src/v2/driver/resolve.go
```diff
@@ -378,8 +379,47 @@ func resolveInputParameter(
...
+		}
+		producer, ok := tasks[metadata.GetTaskNameWithDagID(opts.Task.DependentTasks[0], dag.Execution.GetID())]
+		if !ok {
+			return nil, fmt.Errorf("producer task, %v, not in tasks", producer.TaskName())
+		}
+
+		finalStatus := pipelinespec.PipelineTaskFinalStatus{
+			State:                   producer.GetExecution().GetLastKnownState().String(),
+			PipelineTaskName:        "producer.TaskName()",
+			PipelineJobResourceName: "opts.RunName",
```
mprahl: Why are there quotes around these values?

alyssacgoins: This was for testing purposes and should not have been pushed - I've fixed it now

backend/src/v2/driver/driver_test.go
```diff
@@ -635,3 +635,56 @@ func Test_initPodSpecPatch_resourceRequests(t *testing.T) {
...
+		"test",
+		"0254beba-0be4-4065-8d97-7dc5e3adf300",
+		"1",
+		"false",
+		"false",
+	)
+
+	podSpecString, err := json.Marshal(podSpec)
+	assert.Nil(t, err)
+	assert.Contains(t, string(podSpecString), "--executor_input\",\"{\\\"inputs\\\":{\\\"parameterValues\\\":{\\\"status\\\":{\\\"error\\\":{}, \\\"pipeline_job_resource_name\\\":\\\"test-job-resource-name\\\", \\\"pipeline_task_name\\\":\\\"test-pipeline-task-name\\\", \\\"state\\\":\\\"test-state\\\"}}}}")
```
mprahl: The test failure is because the keys are now camel case. There's also some spacing issues.

My suggestion here is since `podSpec` is an object, you can just parse the `Command` field in the container directly. I'd also suggest parsing the JSON to a map[string]interface{} since you are relying that JSON marshaling will always return the keys in order by doing a JSON string comparison.

alyssacgoins: Updated :+1:

sdk/python/kfp/dsl/executor.py
```diff
@@ -328,16 +328,24 @@ def execute(self) -> Optional[str]:
             # `Optional[str]`. In this case, we need to strip off the part
             # `Optional[]` to get the actual parameter type.
             v = type_annotations.maybe_strip_optional_from_annotation(v)
-
             if v == task_final_status.PipelineTaskFinalStatus:
                 value = self.get_input_parameter_value(k)
+
+                # TODO: PipelineTaskFinalStatus field names formatted in camel case are deprecated.
```
HumairAK:
```suggestion
                # PipelineTaskFinalStatus field names pipelineJobResourceName and pipelineTaskName are deprecated. Support for these fields will be removed at a later date.
```

Can you also add this note to the sdk release notes [here](https://github.com/opendatahub-io/data-science-pipelines/blob/master/sdk/RELEASE.md#deprecations)

samples/v2/pipeline_with_input_status_state.py
```diff
@@ -0,0 +1,22 @@
+
+from kfp import dsl
+
+@dsl.component
+def echo_state(status: dsl.PipelineTaskFinalStatus):
+    assert(status.state == 'COMPLETE')
+    assert('status-state-pipeline' in status.pipeline_job_resource_name)
+    assert(status.pipeline_task_name == 'exit-handler-1')
+    assert(status.error_code is None)
+    assert(status.error_message is None)
```
HumairAK: we shouldn't have these here as it's validating unimplemented behavior, which may give the wrong impression, we can drop a todo comment for future implementer to include error_code and error_message tests once that's been implemented

backend/src/v2/driver/resolve.go
```diff
@@ -378,8 +379,43 @@ func resolveInputParameter(
...
+			return nil, fmt.Errorf("producer task, %v, not in tasks", producer.TaskName())
+		}
+		finalStatus := pipelinespec.PipelineTaskFinalStatus{
+			State:                   producer.GetExecution().GetLastKnownState().String(),
+			PipelineTaskName:        producer.TaskName(),
+			PipelineJobResourceName: opts.RunName,
+			//TODO: Implement fields "Message and "Code" below for Error status.
+			Error: &status.Status{},
+		}
+		finalStatusJSON, err := json.Marshal(finalStatus)
```
HumairAK: consider using `protojson.Marshal(&finalStatus)`, using json.marshal for protomessages is discouraged because of underlying mutex in proto messages failing to copy over

backend/src/v2/driver/resolve.go
```diff
@@ -378,8 +379,43 @@ func resolveInputParameter(
...
+		if err != nil {
+			return nil, fmt.Errorf("failed to marshal PipelineTaskFinalStatus: %w", err)
+		}
+
+		var finalStatusMap map[string]interface{}
+		if err := json.Unmarshal(finalStatusJSON, &finalStatusMap); err != nil {
+			return nil, fmt.Errorf("failed to unmarshal JSON of PipelineTaskFinalStatus: %w", err)
+		}
+
+		// Convert the finalStatusMap to a structpb.Struct
```
HumairAK:
```suggestion
```

samples/v2/pipeline_with_input_status_state.py
alyssacgoins: Yup makes sense.  Updated

backend/src/v2/driver/resolve.go
alyssacgoins: Oh I didn't realize that, thanks for bringing that up! Updated to protojson :+1:

sdk/python/kfp/dsl/executor.py
alyssacgoins: Updated

backend/src/v2/driver/resolve.go
alyssacgoins: Updated

HumairAK:
wooo 
/lgtm
/approve 

thank you!

PR #11956, opened 2025-06-03
adding POC for API tests in goinkgo

nsingla:
Creating a POC for converting Go Unit tests to use Goinkgo + Gomega. This also refactors existing tests and paves way as to how a new test should be written. But basically, we have new requirements to write a new test:

1. Each test file should be based on how API services are grouped here: https://www.kubeflow.org/docs/components/pipelines/reference/api/kubeflow-pipeline-api-spec/ (it makes it easier to understand the grouping and the reasoning behind it)
2. Each endpoint of the service should be tested
3. Every test case per endpoint should be an independent spec (goinkgo spec)
4. Test should cover different possible combinations of the allowed payload - test plan should get reviewed with the team
5. Every method should log what action it's trying to perform and after it performs the action
6. Test Steps should be included in every test for easier reporting (WIP to define this)

New Test Structure for API tests:
```kfp_pipelines
    backend
        src
        test
            v2
                api
                    utils
                    custom_matcher
                    test_file.go
```

**Checklist:**


.github/workflows/api-server-tests.yml
```diff
@@ -0,0 +1,144 @@
...
+
+on:
+  push:
+    branches: [master]
+
+  workflow_dispatch:
+    inputs:
+      test_label:
+        description: "Test label that you want to filter on and run"
+        default: 'ApiServerTests'
```
alyssacgoins, line 17: What are the other possible values other than ApiServerTests?

data/pipeline_files/valid/add_numbers.yaml
```diff
@@ -0,0 +1,86 @@
+# PIPELINE DEFINITION
+# Name: add-numbers
+# Inputs:
+#    a: int
+#    b: int
+# Outputs:
+#    Output: int
+components:
```
alyssacgoins: I noticed the two directories - "valid" and "critical" - what differentiates the two?

.github/workflows/api-server-tests.yml
nsingla: These are basically the labels attached to a ginkgo node, for now, I am adding following labels to Pipeline Upload Tests:
`Positive`, `PipelineUpload`, `ApiServerTests`, `FullRegression`, `Negative`
So we can filter out test run based on the label provided. My suggestion is to use `APIServerTests` label for all API Integration tests, and if anyone wants to run tests manually, then they can always override it and either run specific tests or a super set as well
Gingko also allows you to specify multiple labels with boolean keyword, so its pretty flexible

data/pipeline_files/valid/add_numbers.yaml
nsingla: `critical` is just a subset to run some smoke tests, to capture some very basic pipelines that we can quickly verify if we don;t want the tests to iterate over the whole list of valid pipelines

mprahl:
@nsingla could you please address the CI failure and rebase?

nsingla:
> @nsingla could you please address the CI failure and rebase?

The CI failure is a genuine failure, I think its a performance issue with deleting pipeline versions with cacheEnabled=true and storage=kubernetes

Will check if this needs to be rebased off of master again (did it this morning)

.github/workflows/api-server-tests.yml
alyssacgoins: Gotcha. Is there somewhere in the codebase that these labels are currently or could be documented?

nsingla: we can try to maintain that in this file: https://github.com/kubeflow/pipelines/blob/a73b467037dcd13cb8ab8ac958e6155477a6d6c2/backend/test/v2/api/README.md

nsingla: Done

backend/test/v2/api/utils/pipeline_utils.go
```diff
@@ -0,0 +1,112 @@
...
+func FindPipelineByName(client *api_server.PipelineClient, pipelineName string) bool {
+	requestedNumberOfPipelinesPerPage := 1000
+	params := new(pipeline_params.PipelineServiceListPipelinesParams)
+	params.PageSize = util.Int32Pointer(int32(requestedNumberOfPipelinesPerPage))
+	logger.Log("Get all pipelines")
+	pipelines, size, _, err := client.List(params)
+	Expect(err).NotTo(HaveOccurred())
+	logger.Log("Finding pipeline with name=%s", pipelineName)
+	if size < requestedNumberOfPipelinesPerPage {
+		for _, pipeline := range pipelines {
```
VaniHaripriya, line 103:
```suggestion
		for _, pipeline := range pipelines {
		      if pipeline.DisplayName == pipelineName {
		            return true
		      }
		 }
```

backend/test/v2/api/utils/pipeline_utils.go
```diff
@@ -0,0 +1,112 @@
...
+			if pipeline.DisplayName == pipelineName {
+				return true
+			} else {
```
VaniHaripriya: Could you check this: having return false inside the else block might cause the function to exit after the first non-matching pipeline, instead of iterating through all pipelines.

.github/workflows/api-server-tests.yml
```diff
@@ -0,0 +1,149 @@
+name: API Server Tests
```
VaniHaripriya, line 2: Could you add a short comment at the top explaining the overall purpose of this workflow.

.github/workflows/api-server-tests.yml
```diff
@@ -0,0 +1,149 @@
...
+    needs: build
+    runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        pipeline_store: [ "database", "kubernetes" ]
+        k8s_version: [ "v1.29.2", "v1.31.0" ]
+        cache_enabled: [ "true", "false" ]
+        proxy: [ false, true ]
+      fail-fast: false # So that failure in 1 type of parameterized job does not cause other jobs to terminate prematurely
+    name: KFP API Server tests - K8sVersion=${{ matrix.k8s_version }} pipelineStore=${{ matrix.pipeline_store }} cachEnabled=${{ matrix.cache_enabled }} proxyEnabled=${{ matrix.proxy }}
```
VaniHaripriya:
```suggestion
    name: KFP API Server tests - K8sVersion=${{ matrix.k8s_version }} pipelineStore=${{ matrix.pipeline_store }} cacheEnabled=${{ matrix.cache_enabled }} proxyEnabled=${{ matrix.proxy }}
```

backend/test/v2/api/integration_suite_test.go
```diff
@@ -0,0 +1,132 @@
...
+var pipelineClient *api_server.PipelineClient
+var testStartTimeUTC time.Time
+var testLogsDirectory = "logs"
+var testReportDirectory = "reports"
+var junitReportFilename = "junit.xml"
+var jsonReportFilename = "report.json"
+
+var _ = BeforeSuite(func() {
+	err := os.MkdirAll(testLogsDirectory, 0755)
+	if err != nil {
```
VaniHaripriya: Just curious- any reason for logging the errors and returning here instead of using Expect(err).NotTo(HaveOccurred())?

backend/test/v2/api/integration_suite_test.go
```diff
@@ -0,0 +1,132 @@
...
+	"k8s.io/client-go/kubernetes"
+	"log"
+	"os"
+	"path/filepath"
+	"testing"
+	"time"
+
+	api_server "github.com/kubeflow/pipelines/backend/src/common/client/api_server/v2"
+	"github.com/kubeflow/pipelines/backend/test/v2/api/logger"
+	test "github.com/kubeflow/pipelines/backend/test/v2/api/utils"
```
VaniHaripriya: Could you remove the duplicate imports.

backend/test/v2/api/pipeline_upload_api_test.go
```diff
@@ -0,0 +1,282 @@
...
+func uploadPipeline(pipelineDir string, pipelineFileName string, pipelineName *string) (*model.V2beta1Pipeline, error) {
+	pipelineFile := filepath.Join(pipelineFilesRootDir, pipelineDir, pipelineFileName)
+	uploadParams.SetName(pipelineName)
+	expectedPipeline.DisplayName = *pipelineName
+	logger.Log("Uploading pipeline with name=%s, from file %s", *pipelineName, pipelineFile)
+	return pipelineUploadClient.UploadFile(pipelineFile, uploadParams)
+}
+
+func uploadPipelineAndVerify(pipelineDir string, pipelineFileName string, pipelineName *string) *model.V2beta1Pipeline {
+	createdPipeline, err := uploadPipeline(pipelineDir, pipelineFileName, pipelineName)
```
VaniHaripriya: Could you add tests to cover the immutability of the pipeline spec.

backend/test/v2/api/matcher/pipeline_matcher.go
```diff
@@ -0,0 +1,52 @@
...
+		// Validate namespace only if the mode is kubeflow otherwise everything is in the same namespace and
+		// pipeline object in the response does not even have namespace
+		Expect(actual.Namespace).To(Equal(expected.Namespace), "Pipeline Namespace not matching")
+	}
+	Expect(actual.Description).To(Equal(expected.Description), "Pipeline Description not matching")
+
+}
+
+func MatchPipelineVersions(actual *model.V2beta1PipelineVersion, expected *model.V2beta1PipelineVersion) {
+	Expect(actual.PipelineVersionID).To(Not(Equal(expected.PipelineVersionID)), "Pipeline Version ID is empty")
```
VaniHaripriya: Could you check if the matcher should be Expect(actual.PipelineVersionID).NotTo(BeEmpty()), as the message doesn’t match the current matcher.

backend/test/v2/api/utils/config_utils.go
```diff
@@ -0,0 +1,64 @@
...
+	"k8s.io/client-go/tools/clientcmd"
+	clientcmdapi "k8s.io/client-go/tools/clientcmd/api"
+)
+
+func WaitForReady(initializeTimeout time.Duration) error {
+	operation := func() error {
+		response, err := http.Get("http://localhost:8888/apis/v2beta1/healthz")
+		if err != nil {
+			return err
+		}
```
VaniHaripriya, line 35:
```suggestion
		response, err := http.Get("http://localhost:8888/apis/v2beta1/healthz")
		if err != nil {
			return err
		}
		defer response.Body.Close()
```

backend/test/v2/api/utils/pipeline_version_utils.go
```diff
@@ -0,0 +1,81 @@
...
+	"github.com/kubeflow/pipelines/backend/test/v2/api/logger"
+	. "github.com/onsi/gomega"
+	"sigs.k8s.io/yaml"
+)
+
+// JsonFromYAML - Construct expected Pipeline Spec from the uploaded file
+func JsonFromYAML(pipelineFilePath string) []byte {
+	pipelineSpec, err := os.ReadFile(pipelineFilePath)
+	Expect(err).NotTo(HaveOccurred())
+	jsonSpecFromFile, errDataConvertion := yaml.YAMLToJSON(pipelineSpec)
```
VaniHaripriya:
```suggestion
	jsonSpecFromFile, errDataConversion := yaml.YAMLToJSON(pipelineSpec)
```

backend/test/v2/api/utils/pipeline_version_utils.go
```diff
@@ -0,0 +1,81 @@
...
+	Expect(errDataConvertion).NotTo(HaveOccurred())
```
VaniHaripriya:
```suggestion
	Expect(errDataConversion).NotTo(HaveOccurred())
```

VaniHaripriya (changes requested):

backend/test/v2/api/utils/pipeline_utils.go
nsingla: good catch! fixed

.github/workflows/api-server-tests.yml
nsingla: Done

.github/workflows/api-server-tests.yml
nsingla: Done

backend/test/v2/api/integration_suite_test.go
nsingla: Done

backend/test/v2/api/integration_suite_test.go
nsingla: done

backend/test/v2/api/pipeline_upload_api_test.go
nsingla: Can you please describe the test case you are suggesting? I am already verifying that duplicate name pipeline upload is not allowed, but duplicate pipeline spec with different name is allowed.

backend/test/v2/api/matcher/pipeline_matcher.go
nsingla: :thinking: I actually don;t remember the reasoning behind this assertion, so switching to `Not Be Empty`

backend/test/v2/api/utils/config_utils.go
nsingla: Done

backend/test/v2/api/utils/pipeline_version_utils.go
nsingla: Done

backend/test/v2/api/pipeline_upload_api_test.go
VaniHaripriya: @nsingla Please add a test to cover this scenario- After a pipeline and its version are created, should not allow changes to its spec apart from metadata edits.

backend/test/v2/api/integration_suite_test.go
```diff
@@ -0,0 +1,129 @@
...
+var testReportDirectory = "reports"
+var junitReportFilename = "junit.xml"
+var jsonReportFilename = "report.json"
+
+var _ = BeforeSuite(func() {
+	err := os.MkdirAll(testLogsDirectory, 0755)
+	Expect(err).NotTo(HaveOccurred(), fmt.Sprintf("Error creating testLogsDirectory: %s", testLogsDirectory))
+	err = os.MkdirAll(testReportDirectory, 0755)
+	if err != nil {
+		logger.Log("Error creating reports directory: %s", testReportDirectory)
```
VaniHaripriya: Please remove this block. Since directory creation failure already fails the suite, the logger here will never be reached.

backend/test/v2/api/pipeline_upload_api_test.go
nsingla: But how do I send an update? Fwik, we only have 2 post calls available, and both throws an error if name of the pipeline or pipeline version is the same as existing and I am covering both of these scenarios already. And if the names are different then it's a new pipeline and pipeline version (which is also covered)

backend/test/v2/api/integration_suite_test.go
nsingla: I am assuming you are referring to in case Log directory creation fails, this won't be excuted but this block is for Reports directory. Although I just changed it to use `	Expect(err).NotTo(HaveOccurred(), fmt.Sprintf("Error creating testLogsDirectory: %s", testLogsDirectory))` rather than logging and retunring

backend/test/v2/api/pipeline_upload_api_test.go
nsingla: Discussed offline and explained that the above mentioned scenario will be covered when we add coverage for PipelineService Endpoints [here](https://www.kubeflow.org/docs/components/pipelines/reference/api/kubeflow-pipeline-api-spec/)

VaniHaripriya: As discussed on the call, this test case will be covered under the Pipeline Service Tests.

backend/test/v2/api/integration_suite_test.go
```diff
@@ -0,0 +1,126 @@
...
+		newPipelineUploadClient = func() (*api_server.PipelineUploadClient, error) {
+			return api_server.NewPipelineUploadClient(clientConfig, *isDebugMode)
+		}
+		newPipelineClient = func() (*api_server.PipelineClient, error) {
+			return api_server.NewPipelineClient(clientConfig, *isDebugMode)
+		}
+	}
+
+	pipelineUploadClient, err = newPipelineUploadClient()
+	if err != nil {
```
mprahl: These should have assertions that there are no errors.

backend/test/v2/api/integration_suite_test.go
```diff
@@ -0,0 +1,126 @@
...
+		writeLogFile(specReport)
+	} else {
+		log.Printf("Test passed")
+	}
+})
+
+func TestAPIs(t *testing.T) {
+	RegisterFailHandler(Fail)
+	suiteConfig, reporterConfig := GinkgoConfiguration()
+	suiteConfig.FailFast = false
```
mprahl, line 103: Optional: We could consider failing fast for test runs on PRs.

backend/test/v2/api/integration_suite_test.go
```diff
@@ -0,0 +1,126 @@
...
+	reporterConfig.GithubOutput = true
```
mprahl: Should this be configurable?

backend/test/v2/api/integration_suite_test.go
nsingla: I thought about it as well but I wanted to see all tests run and gather all the failures in the end (as each test is an independent test, and we would like to see which failed and which ones passed)

backend/test/v2/api/integration_suite_test.go
nsingla: we can override it with command line arg, this just acts as a default value, I can remove it and add this to the CI command only?

backend/test/v2/api/matcher/custom_matcher.go
```diff
@@ -0,0 +1,38 @@
...
+
+import (
+	"fmt"
+	"reflect"
+
+	"github.com/onsi/gomega"
+)
+
+// MatchMaps - Iterate over 2 maps and compare if they are same or not
+func MatchMaps(actual interface{}, expected interface{}, mapType string) {
```
mprahl, line 26: Can we use the native `BeEquivalentTo` instead?

backend/test/v2/api/integration_suite_test.go
mprahl: I think remove it and add it to the CI call would be preferable.

I'm surprised the CLI flags override this hardcoded configuration. Could you double check?

backend/test/v2/api/matcher/pipeline_matcher.go
```diff
@@ -0,0 +1,52 @@
...
+
+	"github.com/kubeflow/pipelines/backend/test/v2/api/logger"
+	"github.com/onsi/ginkgo/v2"
+
+	model "github.com/kubeflow/pipelines/backend/api/v2beta1/go_http_client/pipeline_upload_model"
+	. "github.com/onsi/gomega"
+)
+
+func MatchPipelines(actual *model.V2beta1Pipeline, expected *model.V2beta1Pipeline, isKubeflowMode bool) {
+	Expect(actual.PipelineID).To(Not(BeEmpty()), "Pipeline ID is empty")
```
mprahl, line 18: For all these assertions, please use `ExpectWithOffset` so that the assertion failure is reported on the line that called `MatchPipelines` rather than this line here. This applies to other custom matchers.

backend/test/v2/api/matcher/custom_matcher.go
nsingla: I think it uses DeepEquals behind the scene which is not so verbose and makes it hard to know where exactly the difference is, especially problematic for complex data structures. I wanted the comparison to be more verbose and spit out which exact field is not matching

backend/test/v2/api/matcher/pipeline_matcher.go
```diff
@@ -0,0 +1,52 @@
...
+	if !(actualTime.Equal(expectedTime) || actualTime.After(expectedTime)) {
+		logger.Log("Pipeline Version creation time %v is expected to be after test start time %v", actual.CreatedAt, expected.CreatedAt)
+		ginkgo.Fail(fmt.Sprintf("Pipeline Version creation time %v is expected to be after test start time %v", actual.CreatedAt, expected.CreatedAt))
+	}
+	Expect(actual.DisplayName).To(Equal(expected.DisplayName), "Pipeline Display Name not matching")
+	Expect(actual.Description).To(Equal(expected.Description), "Pipeline Description not matching")
+	MatchMaps(actual.PipelineSpec, expected.PipelineSpec, "Pipeline Spec")
+}
+
+func MatchPipelineVersionSpec(actual any, expected map[string]interface{}) {
```
mprahl: This is not used.

backend/test/v2/api/pipeline_upload_api_test.go
```diff
@@ -0,0 +1,282 @@
...
+// ################################################################################################################
+
+var _ = Describe("Verify Pipeline Upload >", Label("Positive", "PipelineUpload", "ApiServerTests", FullRegression), func() {
+
+	/* Critical Positive Scenarios of uploading a pipeline file */
+	Context("Upload a valid critical pipeline file and verify pipeline metadata after upload >", Label(Smoke), func() {
+		var pipelineDir = "valid/critical"
+		criticalPipelineFiles := utils.GetListOfFileInADir(filepath.Join(pipelineFilesRootDir, pipelineDir))
+		for _, pipelineFile := range criticalPipelineFiles {
+			It(fmt.Sprintf("Upload %s pipeline", pipelineFile), func() {
```
mprahl, line 74: See https://onsi.github.io/ginkgo/#dynamically-generating-specs
```suggestion
			pipelineFile := pipelineFile
			It(fmt.Sprintf("Upload %s pipeline", pipelineFile), func() {
```

backend/test/v2/api/pipeline_upload_api_test.go
```diff
@@ -0,0 +1,282 @@
...
+			expectedPipeline.Description = description
+			uploadPipelineAndVerify(pipelineDir, helloWorldPipelineFileName, &pipelineGeneratedName)
+		})
+	})
+
+	/* Positive Scenarios of uploading a pipeline file */
+	Context("Upload a valid pipeline and verify pipeline metadata after upload >", func() {
+		var pipelineDir = "valid"
+		validPipelineFiles := utils.GetListOfFileInADir(filepath.Join(pipelineFilesRootDir, pipelineDir))
+		for _, pipelineFile := range validPipelineFiles {
```
mprahl, line 99:
```suggestion
		for _, pipelineFile := range validPipelineFiles {
			pipelineFile := pipelineFile
```

backend/test/v2/api/pipeline_upload_api_test.go
```diff
@@ -0,0 +1,282 @@
...
+				uploadPipelineAndVerifyFailure(pipelineDir, fileName, &pipelineGeneratedName, "Failed to upload pipeline")
+			})
+		}
+	})
+})
+
+var _ = Describe("Verify Pipeline Upload Version Failure >", Label("Negative", "PipelineUpload", "ApiServerTests", FullRegression), func() {
+	var pipelineDir = "valid/critical"
+
+	/* Positive Scenarios of uploading a pipeline file */
```
mprahl: Did you mean negative scenario here?

backend/test/v2/api/pipeline_upload_api_test.go
```diff
@@ -0,0 +1,282 @@
...
+
+var _ = Describe("Verify Pipeline Upload >", Label("Positive", "PipelineUpload", "ApiServerTests", FullRegression), func() {
+
+	/* Critical Positive Scenarios of uploading a pipeline file */
+	Context("Upload a valid critical pipeline file and verify pipeline metadata after upload >", Label(Smoke), func() {
+		var pipelineDir = "valid/critical"
+		criticalPipelineFiles := utils.GetListOfFileInADir(filepath.Join(pipelineFilesRootDir, pipelineDir))
+		for _, pipelineFile := range criticalPipelineFiles {
+			It(fmt.Sprintf("Upload %s pipeline", pipelineFile), func() {
+				uploadPipelineAndVerify(pipelineDir, pipelineFile, &pipelineGeneratedName)
```
mprahl: Optional: It would be nice if each `It` called `DeferCleanup` so that each test deleted their own pipeline rather than a blanket delete all in `AfterEach`. This could allow running the tests in parallel.

backend/test/v2/api/pipeline_upload_api_test.go
```diff
@@ -0,0 +1,282 @@
...
+const (
+	helloWorldPipelineFileName = "hello-world.yaml"
+	pipelineWithArgsFileName   = "arguments-parameters.yaml"
+)
+
+var pipelineGeneratedName string
+var expectedPipeline *model.V2beta1Pipeline
+var createdPipelines []*model.V2beta1Pipeline
+var testStartTime strfmt.DateTime
+var uploadParams *upload_params.UploadPipelineParams
```
mprahl: I'd rather each test instantiate its own rather than this global that is reinitialized in `BeforeEach`. This will allow for parallel test execution in the future.

backend/test/v2/api/pipeline_upload_api_test.go
```diff
@@ -0,0 +1,282 @@
...
+	// Construct expected pipeline version object for comparison
+	expectedPipelineVersion.Description = descriptionNew
+	expectedPipelineVersion.PipelineSpec = inputFileContent
+	expectedPipelineVersion.DisplayName = pipelineNameNew
+	uploadPipelineVersionAndVerify(pipelineDir, pipelineFileNameWhenChangingVersion, parameters, expectedPipelineVersion)
+}
+
+func uploadPipeline(pipelineDir string, pipelineFileName string, pipelineName *string) (*model.V2beta1Pipeline, error) {
+	pipelineFile := filepath.Join(pipelineFilesRootDir, pipelineDir, pipelineFileName)
+	uploadParams.SetName(pipelineName)
```
mprahl: Is there test coverage for when no name or display name is set? I believe an default name is chosen based on the `pipelineInfo`  field in the pipeline spec.

backend/test/v2/api/pipeline_upload_api_test.go
```diff
@@ -0,0 +1,282 @@
...
+	expectedPipeline.DisplayName = *pipelineName
+	logger.Log("Uploading pipeline with name=%s, from file %s", *pipelineName, pipelineFile)
+	return pipelineUploadClient.UploadFile(pipelineFile, uploadParams)
+}
+
+func uploadPipelineAndVerify(pipelineDir string, pipelineFileName string, pipelineName *string) *model.V2beta1Pipeline {
+	createdPipeline, err := uploadPipeline(pipelineDir, pipelineFileName, pipelineName)
+	logger.Log("Verifying that NO error was returned in the response to confirm that the pipeline was successfully uploaded")
+	Expect(err).NotTo(HaveOccurred())
```
mprahl, line 222: As mentioned in an other comment, please use `ExpectWithOffset` in helper functions.

backend/test/v2/api/pipeline_upload_api_test.go
```diff
@@ -0,0 +1,282 @@
...
+
+	createdPipelineFromDB := utils.GetPipeline(pipelineClient, createdPipeline.PipelineID)
+	Expect(createdPipelineFromDB).To(Equal(*createdPipeline))
+	matcher.MatchPipelines(&createdPipelineFromDB, expectedPipeline, *isKubeflowMode)
+
+	// Validate the created pipeline spec (by API server) matches the input file
+	pipelineVersionFilePath := filepath.Join(pipelineFilesRootDir, pipelineDir, pipelineFileName)
+	validatePipelineSpecs := false
+	var expectedPipelineSpec map[string]interface{}
+	if strings.Contains(pipelineFileName, ".yaml") {
```
mprahl: Optional: Consider reusing `server.ReadPipelineFile` to handle JSON, YAML, and zip differences and `template.NewV2SpecTemplate` to properly parse `PipelineSpec` and `PlatformSpec` cases.

backend/test/v2/api/pipeline_upload_api_test.go
```diff
@@ -0,0 +1,282 @@
...
+		validatePipelineSpecs = true
+		expectedPipelineSpec = utils.ReadYamlFile(pipelineVersionFilePath).(map[string]interface{})
+	} else if strings.Contains(pipelineFileName, ".json") {
+		validatePipelineSpecs = true
+		specFromFile, err := os.ReadFile(pipelineVersionFilePath)
+		Expect(err).NotTo(HaveOccurred())
+		err = json.Unmarshal(specFromFile, &expectedPipelineSpec)
+		Expect(err).NotTo(HaveOccurred())
+	}
+	if validatePipelineSpecs {
```
mprahl: Should it be an error if the file couldn't be parsed rather than just verifying it if it's YAML or JSON?

backend/test/v2/api/pipeline_upload_api_test.go
```diff
@@ -0,0 +1,282 @@
...
+		logger.Log("Verifying that the generated pipeline spec matches the input yaml file")
+		versions := utils.GetSortedPipelineVersionsByCreatedAt(pipelineClient, createdPipeline.PipelineID, nil)
```
mprahl: Can we filter by pipeline version ID instead?

backend/test/v2/api/utils/config_utils.go
```diff
@@ -0,0 +1,72 @@
...
+	"os"
+	"time"
+
+	"github.com/cenkalti/backoff"
+	"github.com/pkg/errors"
+	"k8s.io/client-go/tools/clientcmd"
+	clientcmdapi "k8s.io/client-go/tools/clientcmd/api"
+)
+
+func WaitForReady(initializeTimeout time.Duration) error {
```
mprahl, line 30: Is this used?

backend/test/v2/api/utils/config_utils.go
```diff
@@ -0,0 +1,72 @@
...
+
+func GetClientConfig(namespace string) clientcmd.ClientConfig {
+	loadingRules := clientcmd.NewDefaultClientConfigLoadingRules()
+	loadingRules.DefaultClientConfig = &clientcmd.DefaultClientConfig
+	overrides := clientcmd.ConfigOverrides{Context: clientcmdapi.Context{Namespace: namespace}}
+	return clientcmd.NewInteractiveDeferredLoadingClientConfig(loadingRules,
+		&overrides, os.Stdin)
+}
+
+func GetDefaultPipelineRunnerServiceAccount(isKubeflowMode bool) string {
```
mprahl, line 66: Is this used?

backend/test/v2/api/utils/config_utils.go
```diff
@@ -0,0 +1,72 @@
...
+		return nil
+	}
+
+	b := backoff.NewExponentialBackOff()
+	b.MaxElapsedTime = initializeTimeout
+	err := backoff.Retry(operation, b)
+	return errors.Wrapf(err, "Waiting for ml pipeline API server failed after all attempts.")
+}
+
+func GetClientConfig(namespace string) clientcmd.ClientConfig {
```
mprahl, line 58: Can we reuse `GetClientConfig` from `backend/test/test_utils.go` or is there an import issue?

backend/test/v2/api/utils/file_utils.go
```diff
@@ -0,0 +1,86 @@
...
+	if err != nil {
+		logger.Log("Failed to get current directory, due to %s", err.Error())
+	}
+	return filepath.Join(dir, "..", "..", "..", "../")
+}
+
+// GetPipelineFilesDir Get the directory location of the main list of pipeline files
+func GetPipelineFilesDir() string {
+	projectRootDir := GetProjectRoot()
+	return filepath.Join(projectRootDir, "data", "pipeline_files")
```
mprahl: Could we have the `data` directory be nested under `backend/test/v2` or something?

backend/test/v2/api/utils/file_utils.go
```diff
@@ -0,0 +1,86 @@
...
+	for _, file := range files {
+		if !file.IsDir() {
+			fileNames = append(fileNames, file.Name())
+		}
+	}
+	return fileNames
+}
+
+// ReadYamlFile - Read a YAML file and unmarshall it into a map
+func ReadYamlFile(pipelineFilePath string) interface{} {
```
mprahl: Let's just directly use `server.ReadPipelineFile` and `template.NewV2SpecTemplate`. We don't need to parse YAML files other than compiled pipelines do we?

backend/test/v2/api/utils/kubernetes_utils.go
```diff
@@ -0,0 +1,131 @@
...
+}
+
+// GetPodContainingName - Get the name of the pod with name containing substring
+func GetPodContainingName(client *kubernetes.Clientset, namespace, podName string) *v1.Pod {
+	pods, err := client.CoreV1().Pods(namespace).List(context.Background(), metav1.ListOptions{})
+	if err != nil {
+		logger.Log("Failed to list pods due to: %v", err)
+	}
+	for _, pod := range pods.Items {
+		podNameSplit := strings.Split(pod.Name, "-")
```
mprahl, line 102: Can we use `strings.Contains` instead?

backend/test/v2/api/utils/kubernetes_utils.go
```diff
@@ -0,0 +1,131 @@
...
+	"time"
+
+	"github.com/kubeflow/pipelines/backend/test/v2/api/logger"
+	v1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/client-go/kubernetes"
+)
+
+// ReadContainerLogs - Read pod logs from a specific container
+func ReadContainerLogs(client *kubernetes.Clientset, namespace string, containerName string, follow *bool, sinceTime *time.Time, logLimit *int64) string {
```
mprahl, line 33: Could we just use `ReadPodLogs` directly since `ReadContainerLogs` is only called once for a static pod prefix?

backend/test/v2/api/utils/pipeline_utils.go
```diff
@@ -0,0 +1,110 @@
...
+	logger.Log("Deleting pipeline with id=%s", pipelineId)
+	err := client.Delete(&pipeline_params.PipelineServiceDeletePipelineParams{PipelineID: pipelineId})
+	Expect(err).NotTo(HaveOccurred(), fmt.Sprintf("Error occurred while deleting pipeline with id=%s", pipelineId))
+	logger.Log("Pipeline with id=%s, DELETED", pipelineId)
+}
+
+/*
+Delete all pipelines
+*/
+func DeleteAllPipelines(client *api_server.PipelineClient) {
```
mprahl, line 52: Can these functions be reused from `backend/test/test_utils.go`?

backend/test/v2/api/pipeline_upload_api_test.go
nsingla: Oh! that's really good to know, thanks for catching this

backend/test/v2/api/pipeline_upload_api_test.go
nsingla: copy pasta issue :)

backend/test/v2/api/pipeline_upload_api_test.go
nsingla: `AfterEach` is cleaning up only that test specific data (it not a blanket deleteall, it only deletes the pipelines created by that specific test), these tests are designed to run in parallel

mprahl: Oh, I misread it. Then yes, we should definitely use `DeferCleanup` instead of that global variable of `createdPipelines`.

backend/test/v2/api/pipeline_upload_api_test.go
nsingla: over here, wouldn't we want to actually get the location of failure as well as the caller location?

backend/test/v2/api/pipeline_upload_api_test.go
nsingla: oh! I didn;t know about these methods, will refactor

backend/test/v2/api/utils/config_utils.go
nsingla: this is a copy paste, will be used when we migrate `initialization tests`

backend/test/v2/api/utils/config_utils.go
nsingla: This will be used in PipelineRunTests, again its a copy paste from `integration` package

backend/test/v2/api/utils/file_utils.go
nsingla: I wanted the `data` directory to be globally available for other types of tests as well (it is the location for the `main` list of pipelines from my KEP)

backend/test/v2/api/utils/kubernetes_utils.go
nsingla: I brought this from a future PR, which requires both :)

backend/test/v2/api/utils/pipeline_utils.go
nsingla: these functions were copied from `backend/test/test_utils.go`, just re-organized in a more contextual and modularized way. In the near future, we can delete the duplicate code

backend/test/v2/api/utils/kubernetes_utils.go
nsingla: Let me check, but I believe I had an issue with `strings.Contains`, in a future PR (for E2E tests)

backend/test/v2/api/integration_suite_test.go
nsingla: Done

backend/test/v2/api/integration_suite_test.go
nsingla: Removed from here and added to the CI file

backend/test/v2/api/matcher/pipeline_matcher.go
nsingla: Done

backend/test/v2/api/pipeline_upload_api_test.go
nsingla: Done

backend/test/v2/api/pipeline_upload_api_test.go
nsingla: Fixed

backend/test/v2/api/pipeline_upload_api_test.go
nsingla: Added, I think at the time of the PR, this feature was still in development

backend/test/v2/api/matcher/pipeline_matcher.go
nsingla: removed

backend/test/v2/api/pipeline_upload_api_test.go
nsingla: Done

backend/test/v2/api/pipeline_upload_api_test.go
nsingla: fixed, you are asking many things from my future PR that I have already addressed :)

backend/test/v2/api/pipeline_upload_api_test.go
nsingla: but we don't know the pipeline version in this case, we just upload a pipeline which automatically creates a version but does not return any version info in the response

backend/test/v2/api/utils/config_utils.go
nsingla: This is again an artifact of splitting up the test utils in multiple util files, which I believe is easier to read and use, rather than a giant all encompassing utils file

backend/test/v2/api/utils/file_utils.go
nsingla: Done

backend/test/v2/api/integration_suite_test.go
nsingla: @mprahl do you agree with my approach?

backend/test/v2/api/matcher/custom_matcher.go
nsingla: @mprahl do you agree?

backend/test/v2/api/pipeline_upload_api_test.go
nsingla: chatted offline, that ginkgo runs every node in a separate process, the current approach works fine

backend/test/v2/api/pipeline_upload_api_test.go
nsingla: moved these to a text context object

backend/test/v2/api/pipeline_upload_api_test.go
nsingla: done

backend/test/v2/api/pipeline_upload_api_test.go
nsingla: chatted offline and implemented the suggestion

backend/test/v2/api/utils/config_utils.go
nsingla: @mprahl is it okay to leave these here, because ultimately we will be transitioning all integration tests to this new framework

backend/test/v2/api/utils/config_utils.go
nsingla: @mprahl are you good with my answer?

backend/test/v2/api/utils/file_utils.go
nsingla: @mprahl :+1:  or :-1: ?

backend/test/v2/api/utils/kubernetes_utils.go
nsingla: chatted offline and explained the reasoning behind it

backend/test/v2/api/utils/pipeline_utils.go
nsingla: @mprahl do my reasoning sound good to you?

backend/test/v2/api/integration_suite_test.go
mprahl: Yeah, that's fine.

backend/test/v2/api/matcher/custom_matcher.go
mprahl: Perhaps we can wrap `MatchJSON` or `MatchYAML` instead for a better diff.

backend/test/v2/api/matcher/custom_matcher.go
```diff
@@ -0,0 +1,40 @@
...
+	actualMap, _ := actual.(map[any]interface{})
+	for key, value := range expectedMap {
+		if reflect.TypeOf(value).Kind() == reflect.Map {
+			expectedMapFromValue, _ := value.(map[any]interface{})
+			actualMapFromValue, _ := actualMap[key].(map[any]interface{})
+			MatchMaps(expectedMapFromValue, actualMapFromValue, mapType)
+		}
+		actualStringValue := fmt.Sprintf("%v", actualMap[key])
+		expectedStringValue := fmt.Sprintf("%v", value)
+		gomega.Expect(actualStringValue).To(gomega.Equal(expectedStringValue), fmt.Sprintf("'%s' key's value not matching for type %s", key, mapType))
```
mprahl: If this is a nested map, it'd be hard to know where the key is where as `MatchYAML` tells you the whole path to the key that didn't match.

backend/test/v2/api/utils/config_utils.go
mprahl: Okay, no worries.

backend/test/v2/api/utils/config_utils.go
mprahl: I'd prefer to not duplicate the code where possible.

backend/test/v2/api/utils/file_utils.go
mprahl: Okay, let's just rename it to `test_data` then.

backend/test/v2/api/utils/kubernetes_utils.go
mprahl: I'd prefer `strings.HasPrefix` or a regex match

backend/test/v2/api/utils/pipeline_utils.go
mprahl: You're saying after the refactoring is complete, the duplicated code can be deleted?

nsingla: I don't think we need to wait for the whole refactoring to complete, but once its safe to delete (i.e. all references to that specific duplicate code has been migrated), we can start to delete

backend/test/v2/api/utils/kubernetes_utils.go
nsingla: i think the issue I ran into was that the pod name in RunDetails response may contain unordered words than the actual k8s pod name, which is why I had to split it up and verify that all the words exists in the actual k8s pod name or not

backend/test/v2/api/utils/config_utils.go
nsingla: This will get cleaned up in the future PR, for now I did not want to refactor existing tests when we know those are not long living tests

backend/test/v2/api/matcher/custom_matcher.go
nsingla: done, tests are now using MatchYAML to compare pipeline specs, currently this method is unused but I will leave it there in case it is needed in the future (we can remove it if its still used)

backend/test/v2/api/utils/file_utils.go
nsingla: done

Issue #4649, opened 2020-10-21
[Multi User] Support separate artifact repository for each namespace


Bobgy:
Part of https://github.com/kubeflow/pipelines/issues/1223

Support configuring separate artifact repository for each namespace, so that artifact repository can have namespace separation too.

This is currently not supported in KFP multi user support as explained in https://github.com/kubeflow/pipelines/issues/1223#issuecomment-656507073.

Use your thumbs up to show how important this issue is for you.

We'll most likely need upstream enhancement from argo in https://github.com/argoproj/argo/issues/3307.

Ark-kun:
Last year I've added the `ArtifactRepositoryRef` feature to Argo to support this goal.
The idea is to make it possible to specify the artifact repository information by referencing a ConfigMap.
See https://github.com/argoproj/argo/pull/1350 and https://github.com/argoproj/argo/pull/1821
With both features in place it might even be unnecessary to make any SDK-side changes.
Each namespace can have a different artifact repository ConfigMap with the same name. Then Argo will choose one based on the namespace.
The backend can add the artifactRepositoryRef field when submitting the run.

Bobgy:
If anyone would like to help this issue, confirming `artifactRepositoryRef` usage and whether it supports all the features we want can be a step forward

solarist:
In case of S3, would it make sense to perform the auth for minio using dex? Because otherwise someone has to generate access/secret keys for each namespace -> https://github.com/minio/minio/blob/master/docs/sts/dex.md

marcjimz:
@Bobgy we can pick this up as this will be a requirement for us to go multi-tenant. I think this makes sense as the administrator of the platform would be provisioning the namespaces and with that can provision the scoped configMap. 

I suspect we want to edit the workflow spec more upstream to here: https://github.pwc.com/AI-Lab/kubeflow-pipelines/blob/914ca584adb725db1664d87efa31bdef5a290f1b/backend/src/crd/controller/scheduledworkflow/client/workflow_client.go#L138

Any suggestions on the best place to support the change to the workflow spec?

marcjimz:
@Bobgy we can pick this up as this will be a requirement for us to go multi-tenant. I think this makes sense as the administrator of the platform would be provisioning the namespaces and with that can provision the scoped configMap. 

[...]

PR #11965, opened 2025-06-10
chore: Seaweedfs as Minio replacement

juliusvonkohout:

closes https://github.com/kubeflow/manifests/issues/3119

Thank you everyone!
I am pushing this for 4 years or so and even had google and redhat employees involved, back then even Amazon. It is fundamental for CVEs, maintainabiliy (minio is now stuck for 5 years or so) and hard multi-tenancy as basic requirement for an enterprise platform. We also had approaches there for several years with minio. It started all in 2020 here https://github.com/kubeflow/pipelines/issues/4649 and went via https://github.com/kubeflow/pipelines/pull/7725 (2022) and https://github.com/kubeflow/manifests/pull/2826 (October 2024) to https://github.com/kubeflow/manifests/pull/3051 (2025). Without that experimental and extended tests it would have been very hard to pull of and coordinate. I want to especially highlight @pschoen-itsc who spent his effort here for the public health sector in Germany where many insurances need hard multi-tenancy to process data.

We evaluated many alternatives and now we have something S3 and IAM policy compatible, scalable and with hard multi-tenancy.

@akagami-harsh you can create branches against this PR.

- [x] use seaweedfs
- [x] replace the old sync.py etc.
- [x] refactor our tests to be usable within KFP
- [x] Remove Minio (including /env/azure and other legacy stuff)
- [x] Explain how to use the seaweedfs gateway for AWS/GCP/Azure S3


manifests/kustomize/third-party/seaweedfs/base/pipeline-profile-controller/sync.py
juliusvonkohout, line 1: This file must  completly replace the replace the old default sync.py in another folder. We need to move it.

manifests/kustomize/third-party/seaweedfs/base/pipeline-profile-controller/deployment.yaml
juliusvonkohout, line 1: Also this file must replace the old one.

manifests/kustomize/third-party/seaweedfs/base/seaweedfs/seaweedfs-create-admin-user-job.yaml
```diff
@@ -0,0 +1,68 @@
...
+          privileged: false
+          runAsNonRoot: true
+          # image defaults to root user
+          runAsUser: 1001
+          runAsGroup: 1001
+          capabilities:
+            drop:
+            - ALL
+            add:
+            - NET_BIND_SERVICE
```
juliusvonkohout: why do we need net_bind_service ?

manifests/kustomize/third-party/seaweedfs/base/seaweedfs/seaweedfs-deployment.yaml
```diff
@@ -0,0 +1,73 @@
...
+          privileged: false
+          runAsNonRoot: true
+          # image defaults to root user
+          runAsUser: 1001
+          runAsGroup: 1001
+          capabilities:
+            drop:
+            - ALL
+            add:
+            - NET_BIND_SERVICE
```
juliusvonkohout: why do we need net_bind_service ?

manifests/kustomize/third-party/seaweedfs/base/seaweedfs/seaweedfs-create-admin-user-job.yaml
pschoen-itsc: I think it can be removed. Probably from copy-pasting the deployment below

manifests/kustomize/third-party/seaweedfs/base/seaweedfs/seaweedfs-deployment.yaml
pschoen-itsc: Can be removeed. Seaweed only listens on high ports

juliusvonkohout:
closes https://github.com/kubeflow/pipelines/pull/7725

juliusvonkohout:
We are still missing 
Explain how to use the seaweedfs gateway for AW3/GCP/Azure S3 (Harshvir)
add an architectural diagram here for minio and in general for kubeflow/manifest as from my kubecon presentations and blogs (Julius)

but it is at least ready for a first review @HumairAK

.github/resources/manifests/argo/overlays/proxy/proxy-env.yaml
```diff
@@ -13,4 +13,4 @@ spec:
             - name: HTTPS_PROXY
               value: "http://squid.squid.svc.cluster.local:3128"
             - name: NO_PROXY
-              value: "localhost,127.0.0.1,.svc.cluster.local,kubernetes.default.svc,metadata-grpc-service,0,1,2,3,4,5,6,7,8,9"
+              value: "localhost,127.0.0.1,.svc.cluster.local,kubernetes.default.svc,metadata-grpc-service,10.0.0.0/8,0,1,2,3,4,5,6,7,8,9"
```
juliusvonkohout: what is 10.0.0.0/8 being used for ?

.github/resources/manifests/argo/overlays/no-proxy/workflow-disable-logs-patch.yaml
```diff
@@ -8,7 +8,7 @@ data:
     s3:
```
juliusvonkohout: @HumairAK this file looks in general like a bit of duplication

manifests/kustomize/hack/test.sh
```diff
@@ -35,8 +35,6 @@ kustomization_yamls=(
   "env/gcp"
```
juliusvonkohout, line 35: What about env/gcp @HumairAK

.github/resources/manifests/argo/overlays/proxy/proxy-env.yaml
akagami-harsh: With Squid proxy the ml-pipeline pod was failing with 
https://github.com/kubeflow/pipelines/actions/runs/15985246821/job/45088286866
```
----- LOGS -----
I0630 23:03:47.908058      13 client_manager.go:271] Initializing client manager
I0630 23:03:47.908168      13 client_manager.go:272] Initializing DB client...
I0630 23:03:47.908209      13 config.go:57] Config DBConfig.MySQLConfig.ExtraParams not specified, skipping
I0630 23:03:48.054641      13 client_manager.go:275] DB client initialized successfully
I0630 23:03:48.055979      13 client_manager.go:287] Initializing Object store client...
F0630 23:04:48.093068      13 client_manager.go:653] Failed to check if object store bucket exists. Error: Get "http://10.96.61.232:9000/mlpipeline/?location=": net/http:  
```
So added 10.0.0.0/8 to NO_PROXY to bypass the proxy for all internal cluster communication (10.0.0.0 to 10.255.255.255)

droctothorpe:
This is incredibly comprehensive and impressive, @juliusvonkohout and @akagami-harsh. What obstacles do you anticipate for end users upgrading from Minio to Seawead? Does it make sense to provide migration documentation or automation?

juliusvonkohout:
> This is incredibly comprehensive and impressive, @juliusvonkohout and @akagami-harsh. What obstacles do you anticipate for end users upgrading from Minio to Seawead? Does it make sense to provide migration documentation or automation?

If users really need the old data then the cluster administrator needs to handle it anyway by copying via boto3 from minio to seaweedfs. So that is something we could deal with in follow up PRs. We could provide a job/cronjob that does this automatically. i think even LLMS can write this. 

More interesting is probaly the seaweedfs gateway documentation link that shows how users can connect seaweefdfs to AWS/GCP/Azure/S3 compatible object storage. But also there i prefer a follow up PR. Lets /approve and merge what we have @HumairAK @hbelmiro @droctothorpe and continue in follow up PRs.

HumairAK:
/hold 

hey folks this is on my radar, will get to reviewing this soon - excited for this - thanks for all your hard work

cc @mprahl / @nsingla

.github/resources/scripts/free-disk-space.sh
```diff
@@ -0,0 +1,49 @@
+#!/bin/bash
+set -euo pipefail
+
+# This script frees up disk space on GitHub Actions runners.
+# Several GHA workflows were failing with "no space left on device" errors.
+# This script is only meant to run in GitHub Actions CI environment.
+
+# Safety check: Only run on GitHub Actions
+if [[ "${GITHUB_ACTIONS:-false}" != "true" ]]; then
```
nsingla, line 9: is this going to impact any other run that's running on that runner at the same time?

.github/resources/scripts/kfp-readiness/wait_for_pods.py
```diff
@@ -50,8 +50,17 @@ def get_pod_statuses():
 
 
 def all_pods_ready(statuses):
-    return all(pod_status == 'Running' and ready == total
-               for pod_status, ready, total, _ in statuses.values())
+    def is_pod_ready(pod_name, pod_status, ready, total, waiting_messages):
```
nsingla: `waiting_messages` is unused

.github/workflows/e2e-seaweedfs-test.yml
```diff
@@ -0,0 +1,147 @@
...
+    - name: Create KFP cluster with SeaweedFS
+      id: create-kfp-cluster
+      uses: ./.github/actions/kfp-cluster
+      with:
+        k8s_version: ${{ matrix.k8s_version }}
+      
+    - name: Free up disk space for CI
+      run: ./.github/resources/scripts/free-disk-space.sh
+
+    - name: Install istio
```
nsingla: should this become part of the kustomize file if this is required for seaweedfs?

.github/workflows/e2e-seaweedfs-test.yml
```diff
@@ -0,0 +1,147 @@
...
+      id: install-istio
+      if: ${{ steps.create-kfp-cluster.outcome == 'success' }}
+      run: |
+        kubectl apply -k https://github.com/kubeflow/manifests//common/istio/istio-crds/base?ref=master
```
nsingla: 1 extra slash in `https://github.com/kubeflow/manifests//common/istio/istio-crds/base?ref=master`

.github/resources/scripts/free-disk-space.sh
```diff
@@ -0,0 +1,49 @@
...
+sudo rm -rf /usr/share/dotnet
+sudo rm -rf /opt/ghc
+sudo rm -rf /usr/local/share/boost
+sudo rm -rf /usr/local/lib/android
+sudo rm -rf /usr/local/.ghcup
+sudo rm -rf /usr/share/swift
+
+# Selectively remove large tools from hostedtoolcache while preserving Go, Node, Python
+# Remove these specific large tools that aren't needed for KFP tests
+sudo rm -rf /opt/hostedtoolcache/CodeQL || true
```
nsingla, line 29: Don't we use these to run code analysis?

nsingla (commented):
Have you tried pipelines with big objects (large training data, and big models) in range of few 100 GBs?
And what if someone is running few experiments to fine tune such models? 

This is just to see if there is any performance impact with this move? Looks like there are few issues related to copying large objects/memory issues etc open with Seaweed already (not sure if those apply to us but would be nice to test it out)

.github/resources/scripts/free-disk-space.sh
akagami-harsh: Each job gets a fresh, isolated virtual machine, So no, this will not impact other runs on the same runner.

.github/resources/scripts/free-disk-space.sh
akagami-harsh: The disk cleanup script only removes CodeQL from workflows that don't require it for their test execution. This is safe because CodeQL runs in its own dedicated workflow

.github/workflows/e2e-seaweedfs-test.yml
akagami-harsh: The double slash is actually the correct syntax for kustomize remote references. This is the official format for referencing remote kustomize directories https://github.com/kubernetes-sigs/kustomize/blob/master/examples/remoteBuild.md

nsingla: thanks for the clarification :+1:

.github/resources/scripts/free-disk-space.sh
nsingla: got it

juliusvonkohout:
> Have you tried pipelines with big objects (large training data, and big models) in range of few 100 GBs? And what if someone is running few experiments to fine tune such models?
> 
> This is just to see if there is any performance impact with this move? Looks like there are few issues related to copying large objects/memory issues etc open with Seaweed already (not sure if those apply to us but would be nice to test it out)

It looks very good. If you are interested in numbers there is https://gwdg.de/pdf/Performance_Evaluation_of_Object_Storages_NHR2022_.pdf . Pushing more than 32 GB (uncompressed 60-100GB) into a single file and transferring it for every step in the pipeline from and to S3 is the wrong approach. You should use PVCs for data passing then.

https://github.com/seaweedfs/seaweedfs/wiki/Data-Structure-for-Large-Files

https://github.com/seaweedfs/seaweedfs/blob/master/README.md#storage-size

If you need up to 8 TB for a single file, then you can use the large_disk flavor. But that is for system administrators to decide when customizing kubeflow to special customer needs. https://github.com/seaweedfs/seaweedfs/discussions/5984#discussioncomment-12407046

manifests/kustomize/base/installs/multi-user/pipelines-profile-controller/sync.py
```diff
@@ -74,62 +84,53 @@ def get_settings_from_env(controller_port=None,
...
-                    },
-                }]
-
-
             # Compute status based on observed state.
             desired_status = {
                 "kubeflow-pipelines-ready":
                     len(attachments["Secret.v1"]) == 1 and
-                    len(attachments["ConfigMap.v1"]) == desired_configmap_count and
+                    len(attachments["ConfigMap.v1"]) == 3 and
```
orfeas-k, line 133: Not a full review but does the desired status still needs the following lines?
```
len(attachments["DestinationRule.networking.istio.io/v1alpha3"])                len(attachments["AuthorizationPolicy.security.istio.io/v1beta1"]) == 1
```
I see that `sync.py` doesn't create those istio resources any more

pschoen-itsc: Yes, you are right. I forgot to remove some lines here in the original PR. Deployments, Services and DestionationRule should be removed.

juliusvonkohout: No, i think they can be removed, as the tests show.

.github/workflows/e2e-seaweedfs-test.yml
juliusvonkohout: We want to test the multi-tenancy of the S3 storage, therefore we install Istio. I would keep it separate.

manifests/kustomize/hack/test.sh
juliusvonkohout: @kunal-511 i think env/gcp also needs to go given the seaweedfs gateway support

.github/workflows/e2e-seaweedfs-test.yml
```diff
@@ -0,0 +1,147 @@
...
+        kubectl apply -k https://github.com/kubeflow/manifests//common/istio/istio-namespace/base?ref=master
+        kubectl apply -k https://github.com/kubeflow/manifests//common/istio/istio-install/base?ref=master
```
HumairAK: This and [this](https://github.com/kubeflow/pipelines/pull/11965/files#diff-3eeaad67414bdd388d655745dff212dc422191d3ed51c8ec6b7dfc47ee4faf42R65) reference master in Kubeflow/manifest. This strongly couples two development branches between KFP and KF/Manfiests, if some new change in kf/manifest breask KFP CI this will cause CI failures in KFP. 

Can we pin these to the last stable version? 

cc @mprahl / @droctothorpe

.github/workflows/e2e-test.yml
```diff
@@ -199,6 +202,9 @@ jobs:
           pipeline_store: ${{ matrix.pipeline_store }}
         continue-on-error: true
 
+      - name: Free up disk space
+        run: ./.github/resources/scripts/free-disk-space.sh
+
```
HumairAK: is the a reason for running this twice in a single job?

backend/Dockerfile
```diff
@@ -51,7 +51,7 @@ RUN set -e; \
     < /samples/sample_config.json jq ".pipelines[].file" --raw-output | while read pipeline_yaml; do \
     pipeline_py="${pipeline_yaml%.yaml}"; \
     echo "Compiling: \"$pipeline_py\"" && python3 "$pipeline_py" && echo -n "Output: " && ls "$pipeline_py.yaml"; \
-    done
+done
```
HumairAK: is this a stylistic change? if so my preference is to keep the indent

manifests/kustomize/third-party/seaweedfs/base/seaweedfs/minio-service.yaml
```diff
@@ -2,11 +2,15 @@ apiVersion: v1
 kind: Service
 metadata:
   name: minio-service
```
HumairAK, line 4: what is the value in keeping this around? this seems misleading to have a service named minio-service pointing to something that is not minio

manifests/kustomize/third-party/seaweedfs/base/seaweedfs/seaweedfs-deployment.yaml
```diff
@@ -0,0 +1,71 @@
...
+  namespace: kubeflow
+  labels:
+    app: seaweedfs
+spec:
+  selector:
+    matchLabels:
+      app: seaweedfs
+  strategy:
+    type: Recreate
+  # Single container setup not scalable
```
HumairAK, line 14: can you clarify? are you saying that this setup does not support more than one replicas?

manifests/kustomize/third-party/seaweedfs/OWNERS
```diff
@@ -0,0 +1,6 @@
+approvers:
+#  - pschoen-itsc
+  - juliusvonkohout
+reviewers:
+#  - pschoen-itsc
+  - juliusvonkohout
```
HumairAK: who is pschoen-itsc and why are they commented out? can we remove them?

manifests/kustomize/third-party/seaweedfs/istio/istio-authorization-policy.yaml
```diff
@@ -16,15 +16,17 @@ spec:
     - source:
         principals:
         - cluster.local/ns/kubeflow/sa/ml-pipeline-ui
-  # Allow traffic from User Pipeline Pods, which don't have a sidecar.
+  # Allow traffic to s3 endpoint from User Pipeline Pods, which don't have a sidecar.
+  # Also needed for traffic from seaweedfs init pod. Seaweedfs gives the client an ip to connect to. This can not be
+  # handled well by istio (AuthPolicy). Instead, access to the sensitive ports will be limited by the NetworkPolicy.
   - {}
```
HumairAK, line 22: I'm not too familiar with istio but doesn't this rule effectively allows all requests to seaweedfs, if so doesn't this render the other rules useless?

manifests/kustomize/third-party/seaweedfs/PSS_baseline_enable.sh
HumairAK, line 1: These files: 

pipelines_swfs_install.sh
port_forward_gateway.sh
PSS_baseline_enable.sh
PSS_restricted_enable.sh

Maybe I missed it but, I don't see them being used anywhere, in any CI, or mentioned in any documentation being added. What is the reason for these? use cases?

manifests/kustomize/third-party/seaweedfs/pipelines_swfs_install.sh
```diff
@@ -0,0 +1,21 @@
+#!/bin/bash
```
HumairAK: I'm not sure what the purpose of this file is, and `experimental/seaweedfs/istio` doesn't exist. Is this an accidental add from devleopment?

test/seaweedfs/test_pipeline_v1_seaweedfs.py
```diff
@@ -0,0 +1,57 @@
...
+    
+    def hello_world():
+        print("Hello World from Kubeflow Pipelines V1!")
+        return "Hello World"
+    
+    return func_to_container_op(hello_world)
+
+def hello_world_pipeline():
+    hello_op = hello_world_op()
+    hello_op()
```
HumairAK: since this is testing seaweedfs, we should have artifact passing here, yes?

though as noted elsewhere, I don't think we need these test files to begin with

.github/resources/manifests/argo/overlays/no-proxy/workflow-disable-logs-patch.yaml
HumairAK: can you elaborate?

manifests/kustomize/third-party/seaweedfs/base/seaweedfs/seaweedfs-create-admin-user-job.yaml
```diff
@@ -0,0 +1,66 @@
...
+        image: 'chrislusf/seaweedfs:3.85'
+        env:
+        - name: WEED_CLUSTER_DEFAULT
+          value: "sw"
+        - name: WEED_CLUSTER_SW_MASTER
+          value: "seaweedfs.kubeflow:9333"
+        envFrom:
+        - secretRef:
+            name: mlpipeline-minio-artifact
+        command:
```
HumairAK, line 25: Checking the logs: 


```
kubectl  logs -n kubeflow init-seaweedfs-j2kpf
Waiting for service at http://minio-service.kubeflow:9000/status...
Attempt 1: Service not ready yet, retrying in 5s...
Attempt 2: Service not ready yet, retrying in 5s...
Attempt 3: Service not ready yet, retrying in 5s...
Attempt 4: Service not ready yet, retrying in 5s...
Attempt 5: Service not ready yet, retrying in 5s...
Attempt 6: Service not ready yet, retrying in 5s...
Attempt 7: Service not ready yet, retrying in 5s...
Attempt 8: Service not ready yet, retrying in 5s...
Attempt 9: Service not ready yet, retrying in 5s...
Service at http://minio-service.kubeflow:9000/status is up!
master: seaweedfs.kubeflow:9333 filer: 
I0723 12:07:05.053474 masterclient.go:228 master seaweedfs.kubeflow:9333 redirected to leader 10.244.0.19:9333
master: seaweedfs.kubeflow:9333 filers: [10.244.0.19:8888]
.create bucket under /buckets
created bucket mlpipeline
> > master: seaweedfs.kubeflow:9333 filer: 
I0723 12:07:05.296393 masterclient.go:228 master seaweedfs.kubeflow:9333 redirected to leader 10.244.0.19:9333
master: seaweedfs.kubeflow:9333 filers: [10.244.0.19:8888]
> {
  "identities":  [
    {
      "name":  "kubeflow-admin",
      "credentials":  [
        {
          "accessKey":  "minio",
          "secretKey":  "minio123"
        }
      ],
      "actions":  [
        "Admin"
      ],
      "account":  null
    }
  ],
  "accounts":  []
}
> .%  

```

there looks like credentials are being leaked to stdout logging, can we obscure this?

.github/resources/manifests/argo/overlays/proxy/proxy-env.yaml
HumairAK: it is not guaranteed that clusters internal IP will fall within this range, while this may have worked for your environment, this is not sufficient for all cases

we should instead avoid using IPs all together and default to using the service name, the configuration for the object store can be found here: https://github.com/kubeflow/pipelines/blob/a30e9ead10e0d3facb8ff6fa55cb38956c0d828a/backend/src/apiserver/client_manager/client_manager.go#L628-L629

I think we should just use the dns name like seaweedfs.kubeflow.svc.cluster.local

we should also rename those environment variables to be object store specific since we are getting rid of minio, see [here](https://github.com/kubeflow/pipelines/blob/a30e9ead10e0d3facb8ff6fa55cb38956c0d828a/backend/src/apiserver/client_manager/client_manager.go#L46-L51). These can all be removed, and instead we can leverage the config file via viper [here](https://github.com/kubeflow/pipelines/blob/a30e9ead10e0d3facb8ff6fa55cb38956c0d828a/backend/src/apiserver/client_manager/client_manager.go#L627-L640)

See how accesskey is set [here](https://github.com/kubeflow/pipelines/blob/cc7830812ae3ced24962238dcbf5f113f28c4772/manifests/kustomize/base/pipeline/ml-pipeline-apiserver-deployment.yaml#L111). So for example for host you can add `OBJECTSTORECONFIG_HOST`,  and set this to the seaweedfs service dns name.

.github/workflows/e2e-seaweedfs-test.yml
HumairAK, line 1: There is very little here that actually makes this specific to seaweedFS. This is more of a test for KFP in multi-user mode with the addition of testing namespace isolation in seaweed fs. I think we can just consume this in our existing KFP tests, and just add a new "multi-user" mode matrix to test against. I think we should approach this in the following way: 


1. Move the following steps into an optional confuguration to the [deploy-kfp script](https://github.com/kubeflow/pipelines/blob/9aebb62be1a9412d960fc681787468e1e3ab2001/.github/resources/scripts/deploy-kfp.sh#L19-L18) used in the existing KFP CI. Also as noted elsewhere, we need to keep minio around to allow users to migrate in a release. So the script should support an option to also deploy minio instead of seaweedFS. 


```yaml
    - name: Install istio
    - name: Deploy Platform Agnostic Multi-User
    - name: Install Profile Controller Resources

    # the following might need to be conditionally only for seaweed I think?
    - name: ClusterRole for User Permissions
    - name: Create KF Profile
    - name: ApplyNetwork Policy for Cross-Namespace Access

    - name: Verify Pipeline Integration
```

Remove this workflow, and get rid of `test_pipeline_v1_seaweedfs.py` and `test_pipeline_v2_seaweedfs.py`, this is simply deploying pipelines and verifying success status. We can instead re-use the [kfp-samples](https://github.com/kubeflow/pipelines/blob/master/.github/workflows/kfp-samples.yml) workflow for this. Add a new matrix like: 

```yaml
    strategy:
      matrix:
        include:
          # v1.29.2 we don't test minio
          - k8s_version: "v1.29.2"
            mode: "multi-user"
            storage: "seaweedfs"
          - k8s_version: "v1.29.2"
            mode: "standalone"
            storage: "seaweedfs"

          # v1.31.0 we test minio as well
          # we can remove minio in the release after
          - k8s_version: "v1.31.0"
            mode: "multi-user"
            storage: "minio"
          - k8s_version: "v1.31.0"
            mode: "multi-user"
            storage: "seaweedfs"
          - k8s_version: "v1.31.0"
            mode: "standalone"
            storage: "minio"
          - k8s_version: "v1.31.0"
            mode: "standalone"
            storage: "seaweedfs"
```

use the matrix value to optionally pass multiuser mode & storage mode to the [deploy-kfp action](https://github.com/kubeflow/pipelines/blob/d45b5810eb7dda69a6e4062c69f175e3da98b3bc/.github/actions/kfp-cluster/action.yml#L47-L60)

You will also probably need to update the samples test script to [support token passing](https://github.com/kubeflow/pipelines/blob/c92c2a64e7c80a022dcfcdf0da497c64934ba5af/samples/v2/sample_test.py#L153)

you can also conditionally trigger: 

```yaml
    - name: Test SeaweedFS Namespace Isolation
      id: test-namespace-isolation
      if: ${{ matrix.mode == 'multiuser' && matrix.storage != minio }}
      run: ./test/seaweedfs/namespace_isolation_test.sh
      continue-on-error: true
```

This way we can continuously test not only seaweedFS but multi-user mode against regular KFP ci testing. It will also give more confidence that this is not introducing any breaking changes in KF multi user mode along with seaweedFS. 

cc @mprahl / @nsingla / @droctothorpe

manifests/kustomize/third-party/minio/base/kustomization.yaml
HumairAK, line 1: We should keep minio for a release to allow users to be able to upgrade to newest KFP with minio and then switch over to seaweed FS.

manifests/kustomize/third-party/seaweedfs/base/seaweedfs/mlpipeline-minio-artifact-secret.yaml
HumairAK, line 1: we should stop using a secret name that has `minio` in it if it's being used for seaweedFS

HumairAK: this will need to accompany a change to[ this default](https://github.com/kubeflow/pipelines/blob/cd66b6965c1576d698fa6623cbe7c77cc7e8b5ed/backend/src/v2/config/env.go#L40)

.github/resources/scripts/kfp-readiness/wait_for_pods.py
HumairAK: and `pod_name`

test/seaweedfs/kubeflow-edit-clusterrole.yaml
```diff
@@ -0,0 +1,46 @@
...
+  resources:
+  - deployments
+  - replicasets
+  verbs:
+  - get
+  - list
+  - create
+  - update
+  - patch
+  - delete
```
HumairAK: nit: I would give all files another pass through and add missing newlines at EOF where needed, I see a couple like this

manifests/kustomize/third-party/seaweedfs/base/seaweedfs/minio-service.yaml
pschoen-itsc: We kept it for backward compatibility. I think there are still places in kubeflow where this service is hardcoded (or at least was when @juliusvonkohout recommended that).

manifests/kustomize/third-party/seaweedfs/base/seaweedfs/seaweedfs-deployment.yaml
pschoen-itsc: Yes, that is correct. To scale seaweedfs horizontally you would create different kinds of deployments / stateful sets for S3-Proxy, Filer, Master and Volume server. To keep it simple and low on resource footprint, we opted for the single container version for now. But scaling this Deployment up would not end well, so I added this comment.

manifests/kustomize/third-party/seaweedfs/OWNERS
pschoen-itsc: I made the basic contributions for seaweedfs. If I remember correctly, I could not be owner / reviewer in my first PR here, so it was commented out. So I guess we could fix that now?

.github/workflows/e2e-test.yml
akagami-harsh: No, I'll remove it.

manifests/kustomize/third-party/seaweedfs/base/seaweedfs/minio-service.yaml
juliusvonkohout: Yes we could remove it, but as said this might ease the transition. I would recommend to use a follow up PR for that if needed. Maybe keep it for one kfp release.

.github/workflows/e2e-seaweedfs-test.yml
juliusvonkohout: I asked explicitly to keep it as master since our cicd runs a lot of tests, including the same we added here for seaweedfs before something is merged to master. So master is very stable and can be relied upon. If that is a hard blocker for you we can change it, but I think with master our testing life will be actually easier.

manifests/kustomize/third-party/seaweedfs/base/seaweedfs/seaweedfs-deployment.yaml
juliusvonkohout: Seawedfs can massively scale and from the pdf linked you can see how fast it is. But for very large installations with 1000s of users it could make sense for the cluster administrator to scale horizontally or use the gateway to connect to AWS/GCS etc. But that is really customer specific. We could evaluate in a follow up PR to directly offer the scalable option.

.github/workflows/e2e-seaweedfs-test.yml
juliusvonkohout: The regular KFP with Minio does not support proper hard multi-tenancy and the tests will fail. So I would rather keep the changes you suggest minimal. We do not need to keep minio, only the minio service that points to seaweedfs again.

manifests/kustomize/third-party/seaweedfs/base/seaweedfs/mlpipeline-minio-artifact-secret.yaml
juliusvonkohout: Also that I would rather keep for backwards compatibility now and remove in a follow up PR.

manifests/kustomize/third-party/seaweedfs/OWNERS
HumairAK: Hey @pschoen-itsc awesome work! appreciate your contributions here. In that case I think `reviewers` makes the most sense here, so I'm fine with uncommented that out, but let's remove you from the owners section, as this requires you to first have operated under the capacity of a reviewer for some time.

.github/workflows/e2e-seaweedfs-test.yml
HumairAK: @juliusvonkohout which tests? We don't need to do the namespace isolation tests in the case of minio if that is what you are referring to. 

Consider the user perspective: 

1. I am deploying KFP x.y with minio and no seaweed
2. KFP x.y+1 switches to seaweedfs and minio, with seaweed as the default 
3. I upgrade to KFP x.y+1 but specify the minio, since this is what my x.y used
4. I can now migrate to seaweed fas in x.y+1 

We can remove minio in x.y+2, and users can use x.y+1 as a migration point. If x.y+1 does a hard switch, then users cannot direclty upgrade to migrate and have to do a lot of additional extra deployments of kfp to migrate.

.github/workflows/e2e-seaweedfs-test.yml
mprahl: We can keep it for now and revisit it later.

HumairAK: Okay let's keep it.

.github/resources/manifests/argo/overlays/no-proxy/workflow-disable-logs-patch.yaml
juliusvonkohout: Well if that is fine for you, lets keep it.

manifests/kustomize/third-party/seaweedfs/istio/istio-authorization-policy.yaml
juliusvonkohout: Minio is also open for everyone, since not all workload using Minio/Seaweedfs support istio yet.

manifests/kustomize/third-party/seaweedfs/PSS_baseline_enable.sh
juliusvonkohout: @akagami-harsh please check and use or remove. We defintely should set PSS restricted for the tests.

manifests/kustomize/third-party/seaweedfs/pipelines_swfs_install.sh
juliusvonkohout: @akagami-harsh please check and use or remove.

manifests/kustomize/third-party/seaweedfs/base/seaweedfs/mlpipeline-minio-artifact-secret.yaml
juliusvonkohout: covered here https://github.com/kubeflow/pipelines/pull/11965#discussion_r2228970531

test/seaweedfs/kubeflow-edit-clusterrole.yaml
juliusvonkohout: Maybe you want to steal bash/python/yaml linting in general for KFP https://github.com/kubeflow/manifests/blob/master/.yamllint.yaml  https://github.com/kubeflow/manifests/blob/master/.github/workflows/linting_bash_python_yaml_files.yaml in a follow up PR.

juliusvonkohout:
@akagami-harsh please fix the merge conflicts. @HumairAK I think it is ready for the next review round.

manifests/kustomize/third-party/seaweedfs/OWNERS
```diff
@@ -0,0 +1,6 @@
+approvers:
+#  - pschoen-itsc
+  - juliusvonkohout
+reviewers:
+#  - pschoen-itsc
```
pschoen-itsc:
```suggestion
  - juliusvonkohout
reviewers:
  - pschoen-itsc
```

juliusvonkohout:
@pschoen-itsc you need to raise a PR in kubeflow/internal-acls to become a member

pschoen-itsc:
> @pschoen-itsc you need to raise a PR in kubeflow/internal-acls to become a member

Sure. kubeflow/internal-acls#815

juliusvonkohout:
> > @pschoen-itsc you need to raise a PR in kubeflow/internal-acls to become a member
> 
> Sure. [kubeflow/internal-acls#815](https://github.com/kubeflow/internal-acls/pull/815)

Merged

manifests/kustomize/third-party/seaweedfs/istio/istio-authorization-policy.yaml
HumairAK: the question above looks unaddressed, if we have a wide rule like `{}` why are the other rules needed?

.github/workflows/kfp-samples.yml
HumairAK, line 1: can you also update `on.pull_request.paths` and add: 

```
- manifests/kustomize/third-party/seaweedfs/**
- manifests/kustomize/third-party/argo/base/workflow-controller-configmap-patch.yaml
- manifests/kustomize/third-party/minio/**
```

.github/workflows/kfp-samples.yml
```diff
@@ -21,45 +21,48 @@ jobs:
...
+          - k8s_version: "v1.29.2"
+            mode: "standalone"
+            storage: "seaweedfs"
+
+          - k8s_version: "v1.31.0"
+            mode: "multi-user"
+            storage: "seaweedfs"
+          - k8s_version: "v1.31.0"
+            mode: "standalone"
+            storage: "minio"
```
HumairAK, line 46: we're missing: 

```yaml
          - k8s_version: "v1.31.0"
            mode: "multi-user"
            storage: "minio"
```

.github/resources/manifests/argo/overlays/no-proxy-minio/workflow-disable-logs-patch.yaml
```diff
@@ -0,0 +1,20 @@
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  name: workflow-controller-configmap
+data:
```
HumairAK: can you explain the need for this patch, is it to disable logs? if so can you explain the reasoning? is it to save on space or does log archiving not work anymore?

manifests/kustomize/env/platform-agnostic-minio/kustomization.yaml
HumairAK, line 1: let's put a comment at the top of this file that says something like: 

`Minio deployment manifests in KFP will be removed in 2.16.0, in the future end-users will need to provision their own minio deployments.`

backend/src/v2/metadata/env.go
```diff
@@ -1,13 +1,31 @@
...
 }
 
 func DefaultConfig() *ServerConfig {
+	// If proxy is enabled, use DNS name `metadata-grpc-service.kubeflow:8080` as default.
+	_, isHttpProxySet := os.LookupEnv(proxy.HttpProxyEnv)
+	_, isHttpsProxySet := os.LookupEnv(proxy.HttpsProxyEnv)
+	if isHttpProxySet || isHttpsProxySet {
+		return &ServerConfig{
+			Address: metadataGrpcServiceAddress,
+			Port:    metadataGrpcServicePort,
```
HumairAK: was there a particular reason we have to only do this for proxy? it seems to me we can just always use these default values wether proxy is set or not (same for cache.go and objectstore/config.go)

manifests/kustomize/third-party/seaweedfs/base/seaweedfs/seaweedfs-networkpolicy.yaml
```diff
@@ -0,0 +1,43 @@
...
+metadata:
+  name: default-allow-same-namespace
+  namespace: kubeflow
+spec:
+  podSelector: {}
+  ingress:
+    - from:
+        - podSelector: {}
+  policyTypes:
+    - Ingress
```
HumairAK: can you explain the need for this policy? this does not look KFP specific and would have broader implications

manifests/kustomize/third-party/seaweedfs/base/seaweedfs/mlpipeline-minio-artifact-secret.yaml
HumairAK: I would like us to explore renaming the service and secret names `minio-service` and `mlpipeline-minio-artifact` - @akagami-harsh I'll leave this call to you - you have seen now where these values are configured in the codebase so my thinking is that it should be a small-lift to just use new seaweedfs names and not accrue this future tech debt of renaming. 

I don't think keeping these names the same makes migration easier, in fact it might have the opposite affect if the user deploys KFP + Seaweedfs over their KFP+Minio and think get the wrong impression that their minio-service is still talking to minio when they try to migrate. 

@akagami-harsh if you find that this is a non-trivial task, and it's more than just a ctrl+f and replace, then we can do this as a follow-on PR, preferably before the next release

HumairAK (changes requested):

manifests/kustomize/env/platform-agnostic-minio/kustomization.yaml
juliusvonkohout: And that the default will be seawedfs with a link to to and the gateway documentation

manifests/kustomize/third-party/seaweedfs/base/seaweedfs/seaweedfs-networkpolicy.yaml
juliusvonkohout: That is the default policy used in Kubeflow platform for 2 years or so.

juliusvonkohout: Check kubeflow/manifests/common/networkpolicies

pschoen-itsc: We needed to add the policy below to restrict the access to seaweedfs and with adding that policy we now need something to still allow other traffic (inside the namespace).

HumairAK: that is fine, but since this is not seaweedFS specific, can you make this part of the standard pipelines manifest instead? https://github.com/kubeflow/pipelines/tree/master/manifests/kustomize/base/pipeline

juliusvonkohout:
@HumairAK Regarding "

I would like us to explore renaming the service and secret names minio-service and mlpipeline-minio-artifact - @akagami-harsh I'll leave this call to you - you have seen now where these values are configured in the codebase so my thinking is that it should be a small-lift to just use new seaweedfs names and not accrue this future tech debt of renaming.

I don't think keeping these names the same makes migration easier, in fact it might have the opposite affect if the user deploys KFP + Seaweedfs over their KFP+Minio and think get the wrong impression that their minio-service is still talking to minio when they try to migrate.

@akagami-harsh if you find that this is a non-trivial task, and it's more than just a ctrl+f and replace, then we can do this as a follow-on PR, preferably before the next release
"

I think we should hide this implementation detail from the users in general. I think the data scientists should not worry what is used as S3 backend, it should just be S3 compatible.

backend/src/v2/metadata/env.go
akagami-harsh: I initially kept the ClusterIP method assuming it might have a specific purpose, but yes — we can always use the default DNS values. I’ll update it.

.github/resources/manifests/argo/overlays/no-proxy-minio/workflow-disable-logs-patch.yaml
akagami-harsh: This was copied from the no-proxy config and is identical to [workflow-controller-configmap-patch.yaml](https://github.com/kubeflow/pipelines/blob/master/manifests/kustomize/third-party/argo/base/workflow-controller-configmap-patch.yaml). It looks redundant and doesn’t change logging — I’ll remove it.

samples/v2/sample_test.py
```diff
@@ -146,28 +150,62 @@ def delete_k8s_yaml(namespace: str, yaml_file: str):
         print(f'Exception when deleting from YAML: {e}')
 
 
+def get_auth_token():
```
juliusvonkohout: please use the proper word authentication

manifests/kustomize/third-party/seaweedfs/istio/istio-authorization-policy.yaml
juliusvonkohout: Yes, they are probably not needed. but that is just copying the existing minio authorizationpolicy configuration. That might be something for a separate follow up PR. This one here is getting quite big and hard to maintain.

.github/actions/kfp-cluster/action.yml
```diff
@@ -68,10 +77,22 @@ runs:
 
         if [ "${{ inputs.proxy }}" = "true" ]; then
           ARGS="${ARGS} --proxy"
-        elif [ "${{inputs.cache_enabled }}" = "false" ]; then
+        fi
+        
+        if [ "${{inputs.cache_enabled }}" = "false" ]; then
           ARGS="${ARGS} --cache-disabled"
-        elif [ "${{inputs.pipeline_store }}" = "kubernetes" ]; then
```
juliusvonkohout, line 73: Here the rebase maybe messed something up @akagami-harsh

.github/workflows/kfp-samples.yml
```diff
@@ -25,88 +25,122 @@ jobs:
...
+      continue-on-error: true
+
+    - name: Create KFP cluster
+      id: create-kfp-cluster
+      uses: ./.github/actions/kfp-cluster
+      with:
+        k8s_version: ${{ matrix.k8s_version }}
+        multi_user: ${{ matrix.mode == 'multi-user' }}
+        storage_backend: ${{ matrix.storage }}
+      continue-on-error: true
```
juliusvonkohout: also here the rebase does not look right @akagami-harsh

manifests/kustomize/env/gcp/kustomization.yaml
juliusvonkohout, line 1: I think env/gcp should also go @akagami-harsh

.github/workflows/kfp-samples.yml
```diff
@@ -25,88 +25,122 @@ jobs:
...
+        exit 1
+
+    - name: Collect test results
+      if: always()
+      uses: actions/upload-artifact@v4
+      with:
+        name: kfp-samples-tests-artifacts-k8s-${{ matrix.k8s_version }}
+        path: /tmp/tmp*/*
+
+    - name: Test SeaweedFS Namespace Isolation
```
juliusvonkohout: I do not see this test being triggered @akagami-harsh maybe also a rebase related problem.

manifests/kustomize/env/gcp/kustomization.yaml
akagami-harsh: For now i think we should keep env/gcp because of cloudsql-proxy and inverse-proxy.

.github/workflows/kfp-samples.yml
akagami-harsh: Added in this pr https://github.com/kubeflow/pipelines/pull/12140

HumairAK:
@akagami-harsh I noticed this flakiness occur in this CI failure https://github.com/kubeflow/pipelines/actions/runs/16993993913/job/48180413365?pr=11965 

I noticed the seaweedfs pod was filled with: 

```
2025-08-15T16:40:38.8534567Z I0815 16:39:12.797939 master_grpc_server_volume.go:141 volume grow &{Option:{"collection":"mlpipeline","replication":{},"ttl":{"Count":0,"Unit":0},"preallocate":1073741824,"version":3} Count:0 Force:false Reason:grpc assign}
2025-08-15T16:40:38.8535444Z E0815 16:39:12.801327 volume_grpc_admin.go:59 assign volume volume_id:82  collection:"mlpipeline"  preallocate:1073741824  replication:"000"  version:3: No more free space left
2025-08-15T16:40:38.8536338Z W0815 16:39:12.802934 volume_growth.go:273 Failed to assign volume 82 on topo:DefaultDataCenter:DefaultRack:10.244.0.26:8080: rpc error: code = Unknown desc = No more free space left
2025-08-15T16:40:38.8537464Z I0815 16:39:12.803731 volume_growth.go:120 create 7 volume, created 0: failed to assign volume 82 on topo:DefaultDataCenter:DefaultRack:10.244.0.26:8080: rpc error: code = Unknown desc = No more free space left
2025-08-15T16:40:38.8538277Z I0815 16:39:12.815178 master_grpc_server_volume.go:141 volume grow &{Option:{"collection":"mlpipeline","replication":{},"ttl":{"Count":0,"Unit":0},"preallocate":1073741824,"version":3} Count:0 Force:false Reason:grpc assign}
```

I also notice that the "Free up Disk Space" step took 8 min 39s to complete, which is odd given the others complete much faster. Still, the final disk usage reports the same - so I'm not sure it's related but it was the only one that stood out.

Not sure if this is seaweedFS specific, but looking at the history of this workflow, I couldn't find us encountering this behavior with minio in recent history. Thoughts?

manifests/kustomize/third-party/seaweedfs/base/seaweedfs/seaweedfs-create-admin-user-job.yaml
```diff
@@ -0,0 +1,77 @@
+kind: Job
+apiVersion: batch/v1
+metadata:
+  name: init-seaweedfs
+spec:
+  template:
+    metadata:
+      name: init-seaweedfs
```
HumairAK, line 8: @akagami-harsh / @juliusvonkohout do you know why this init job was required? It seems like it just creates a bucket. The KFP server on startup will already do this if the [bucket doesn't already exist](https://github.com/kubeflow/pipelines/blob/4eac7349764978a7aa57be4978bdd91fda7393d0/backend/src/apiserver/client_manager/client_manager.go#L632). Did this not work?

pschoen-itsc: It creates tje bucket and the admin user for profile management. So I guess it would be possible to remove the bucket creation from this job, but we still need to create an admin user.

juliusvonkohout: Yes i would like to keep it.

juliusvonkohout: And seaweedfs/S3 might also be used by other components, not just KFP. Just for historic reasons it is in kubeflow/pipelines

manifests/kustomize/third-party/seaweedfs/base/seaweedfs/seaweedfs-create-admin-user-job.yaml
akagami-harsh: solved in this pr https://github.com/kubeflow/pipelines/pull/12080

.github/actions/kfp-cluster/action.yml
juliusvonkohout: @akagami-harsh told me that it is intended

manifests/kustomize/third-party/seaweedfs/base/seaweedfs/mlpipeline-minio-artifact-secret.yaml
juliusvonkohout: Lets put this into a follow up PR.

juliusvonkohout:
/lgtm
@HumairAK for approval

akagami-harsh:
/lgtm
@HumairAK for approval

akagami-harsh:
> @akagami-harsh I noticed this flakiness occur in this CI failure https://github.com/kubeflow/pipelines/actions/runs/16993993913/job/48180413365?pr=11965
> 
> I noticed the seaweedfs pod was filled with:
> 
> ```
> 2025-08-15T16:40:38.8534567Z I0815 16:39:12.797939 master_grpc_server_volume.go:141 volume grow &{Option:{"collection":"mlpipeline","replication":{},"ttl":{"Count":0,"Unit":0},"preallocate":1073741824,"version":3} Count:0 Force:false Reason:grpc assign}
> 2025-08-15T16:40:38.8535444Z E0815 16:39:12.801327 volume_grpc_admin.go:59 assign volume volume_id:82  collection:"mlpipeline"  preallocate:1073741824  replication:"000"  version:3: No more free space left
> 2025-08-15T16:40:38.8536338Z W0815 16:39:12.802934 volume_growth.go:273 Failed to assign volume 82 on topo:DefaultDataCenter:DefaultRack:10.244.0.26:8080: rpc error: code = Unknown desc = No more free space left
> 2025-08-15T16:40:38.8537464Z I0815 16:39:12.803731 volume_growth.go:120 create 7 volume, created 0: failed to assign volume 82 on topo:DefaultDataCenter:DefaultRack:10.244.0.26:8080: rpc error: code = Unknown desc = No more free space left
> 2025-08-15T16:40:38.8538277Z I0815 16:39:12.815178 master_grpc_server_volume.go:141 volume grow &{Option:{"collection":"mlpipeline","replication":{},"ttl":{"Count":0,"Unit":0},"preallocate":1073741824,"version":3} Count:0 Force:false Reason:grpc assign}
> ```
> 
> I also notice that the "Free up Disk Space" step took 8 min 39s to complete, which is odd given the others complete much faster. Still, the final disk usage reports the same - so I'm not sure it's related but it was the only one that stood out.
> 
> Not sure if this is seaweedFS specific, but looking at the history of this workflow, I couldn't find us encountering this behavior with minio in recent history. Thoughts?


I think lots of files were deleted in the free-disk space step, creating fragmented free space, there is a [`preallocation` ](https://github.com/seaweedfs/seaweedfs/wiki/Optimization#preallocate-volume-file-disk-spaces)feature in seaweedfs. Preallocation needs large contiguous blocks, not just total free space. logs show preallocate:1073741824 (1GB) - it needs 1GB of contiguous space
```
2025-08-15T16:40:38.8534567Z I0815 16:39:12.797939 master_grpc_server_volume.go:141 volume grow &{Option:{"collection":"mlpipeline","replication":{},"ttl":{"Count":0,"Unit":0},"preallocate":1073741824,"version":3} Count:0 Force:false Reason:grpc assign}
2025-08-15T16:40:38.8535444Z E0815 16:39:12.801327 volume_grpc_admin.go:59 assign volume volume_id:82  collection:"mlpipeline"  preallocate:1073741824  replication:"000"  version:3: No more free space left
2025-08-15T16:40:38.8536338Z W0815 16:39:12.802934 volume_growth.go:273 Failed to assign volume 82 on topo:DefaultDataCenter:DefaultRack:10.244.0.26:8080: rpc error: code = Unknown desc = No more free space left
2025-08-15T16:40:38.8537464Z I0815 16:39:12.803731 volume_growth.go:120 create 7 volume, created 0: failed to assign volume 82 on topo:DefaultDataCenter:DefaultRack:10.244.0.26:8080: rpc error: code = Unknown desc = No more free space left
2025-08-15T16:40:38.8538277Z I0815 16:39:12.815178 master_grpc_server_volume.go:141 volume grow &{Option:{"collection":"mlpipeline","replication":{},"ttl":{"Count":0,"Unit":0},"preallocate":1073741824,"version":3} Count:0 Force:false Reason:grpc assign}
```  
After disk cleanup, the disk likely has plenty of total free space but it's fragmented, i think this would explain why MinIO don't hit this issue

HumairAK:
@akagami-harsh I'm skeptical that this is due to defragmentation, and it's likely due to the storage size we are setting in the pvc for seaweedfs, see [here](https://github.com/kubeflow/pipelines/pull/11965/files#diff-abd78b443c67cdc04d16633ac335e6e93736ff2bfa5c83c88d418ed8ea19dec5). Can we set it to the same size as the minio pvc at 20Gi [here](https://github.com/kubeflow/pipelines/blob/6033718786574cd64b1ebd59e0246e7c5ba1fc88/manifests/kustomize/third-party/minio/base/minio-pvc.yaml#L10)?

akagami-harsh:
> @akagami-harsh I'm skeptical that this is due to defragmentation, and it's likely due to the storage size we are setting in the pvc for seaweedfs, see [here](https://github.com/kubeflow/pipelines/pull/11965/files#diff-abd78b443c67cdc04d16633ac335e6e93736ff2bfa5c83c88d418ed8ea19dec5). Can we set it to the same size as the minio pvc at 20Gi [here](https://github.com/kubeflow/pipelines/blob/6033718786574cd64b1ebd59e0246e7c5ba1fc88/manifests/kustomize/third-party/minio/base/minio-pvc.yaml#L10)?

@HumairAK, Opened a pr to update pvc volume size https://github.com/kubeflow/pipelines/pull/12156

HumairAK:
thanks @akagami-harsh, this lgtm, can you squash your commits and add a meaningful clean commit message?

juliusvonkohout:
> thanks @akagami-harsh, this lgtm, can you squash your commits and add a meaningful clean commit message?

The commits will be automatically squashed on merge with the PR title as commit message, so i would like to avoid any further changes.

image


also given multiple authors it is a bit more complicated.

HumairAK:
I updated the commit message to a more meaningful message - in general I prefer to have the PR authors curate their message as they have more domain knowledge. 

Thank you everyone for all your hard work on this - a much needed change, well done all around !

droctothorpe:
Congrats, everyone! Phenomenal (and overdue) enhancement! You should consider submitting a talk on this to the next Kubeflow Summit in Europe.

PR #11980, opened 2025-06-13
feat(backend): support for optional input parameters in nested pipelines

### LINKED ISSUE #11957: chore(backend):  support for Optional input parameters in nested pipelines
## Chore description

If optional input parameters for a component in a nested pipeline are not provided, the tasks using those inputs fail with the error below:

`failed: failed to resolve inputs: resolving input parameter with spec component_input_parameter:"optional_input": parent DAG does not have input parameter optional_input`

Support needed for nested pipelines with optional input parameters such that the pipeline finishes successfully for cases in which the optional parameters are set and cases in which they are not set. 

### Labels

/area backend

---

Love this idea? Give it a 👍.

alyssacgoins:
Resolves optional input parameters for components of a nested pipeline. RuntimeConfigs updated to include inputs from nested pipeline components in addition to root pipeline. Task-level inputs are resolved with the input passed in by the corresponding component. 
fixes #11957

**Checklist:**

### DIFF
--- a/backend/src/v2/driver/resolve.go
+++ b/backend/src/v2/driver/resolve.go
@@ -29,7 +29,7 @@ import (
 	"google.golang.org/protobuf/types/known/structpb"
 )
 
-var ErrResolvedParameterNull = errors.New("the resolvead input parameter is null")
+var ErrResolvedParameterNull = errors.New("the resolved input parameter is null")
 
 // resolveUpstreamOutputsConfig is just a config struct used to store the input
 // parameters of the resolveUpstreamParameters and resolveUpstreamArtifacts
@@ -142,6 +142,9 @@ func resolveInputs(
 		}
 		return nil
 	}
+	// Track parameters set to nil by the driver (for the case in which optional pipeline input parameters are not
+	// included, and default value is nil).
+	parametersSetNilByDriver := map[string]bool{}
 	handleParamTypeValidationAndConversion := func() error {
 		// TODO(Bobgy): verify whether there are inputs not in the inputs spec.
 		for name, spec := range inputsSpec.GetParameters() {
@@ -179,6 +182,10 @@ func resolveInputs(
 			case pipelinespec.ParameterType_STRING:
 				_, isValueString := value.GetKind().(*structpb.Value_StringValue)
 				if !isValueString {
+					// If parameter was set to nil by driver, allow input parameter to have a nil value.
+					if parametersSetNilByDriver[name] {
+						continue
+					}
 					// TODO(Bobgy): discuss whether we want to allow auto type conversion
 					// all parameter types can be consumed as JSON string
 					text, err := metadata.PbValueToText(value)
@@ -193,6 +200,10 @@ func resolveInputs(
 				}
 				switch v := value.GetKind().(type) {
 				case *structpb.Value_NullValue:
+					// If parameter was set to nil by driver, allow input parameter to have a nil value.
+					if parametersSetNilByDriver[name] {
+						continue
+					}
 					return fmt.Errorf("got null for input parameter %q", name)
 				case *structpb.Value_StringValue:
 					// TODO(Bobgy): consider whether we support parsing string as JSON for any other types.
@@ -269,7 +280,26 @@ func resolveInputs(
 		}
 		return inputs, nil
 	}
+	// A DAG driver (not Root DAG driver) indicates this is likely the start of a nested pipeline.
+	// Handle omitted optional pipeline input parameters similar to how they are handled on the root pipeline.
+	isDagDriver := opts.DriverType == "DAG"
+	if isDagDriver {
+		for name, paramSpec := range opts.Component.GetInputDefinitions().GetParameters() {
+			_, ok := task.Inputs.GetParameters()[name]
+			if !ok && paramSpec.IsOptional {
+				if paramSpec.GetDefaultValue() != nil {
+					// If no value was input, pass along the default value to the component.
+					inputs.ParameterValues[name] = paramSpec.GetDefaultValue()
+				} else {
+					//  If no default value is set, pass along the null value to the component.
+					//	This is analogous to a pipeline run being submitted without optional pipeline input parameters.
+					inputs.ParameterValues[name] = structpb.NewNullValue()
+					parametersSetNilByDriver[name] = true
+				}
 
+			}
+		}
+	}
 	// Handle parameters.
 	for name, paramSpec := range task.GetInputs().GetParameters() {
 		v, err := resolveInputParameter(ctx, dag, pipeline, opts, mlmd, paramSpec, inputParams)
--- a/samples/v2/nested_pipeline_opt_inputs_nil.py
+++ b/samples/v2/nested_pipeline_opt_inputs_nil.py
@@ -1,0 +1,35 @@
+
+from kfp import compiler
+from kfp import dsl
+
+
+@dsl.component()
+def component_str(componentInput: str = None):
+    if componentInput is not None:
+        raise ValueError(f"componentInput should be None but is {componentInput}")
+
+@dsl.component()
+def component_int(componentInput: int = None):
+    if componentInput is not None:
+        raise ValueError(f"componentInput should be None but is {componentInput}")
+
+@dsl.component()
+def component_bool(componentInput: bool = None):
+    if componentInput is not None:
+        raise ValueError(f"componentInput should be None but is {componentInput}")
+
+@dsl.pipeline()
+def nested_pipeline(nestedInputStr: str = None, nestedInputInt: int = None, nestedInputBool: bool = None):
+    component_str(componentInput=nestedInputStr).set_caching_options(False)
+    component_int(componentInput=nestedInputInt).set_caching_options(False)
+    component_bool(componentInput=nestedInputBool).set_caching_options(False)
+
+
+@dsl.pipeline()
+def nested_pipeline_opt_inputs_nil():
+    nested_pipeline().set_caching_options(False)
+
+
+if __name__ == '__main__':
+    compiler.Compiler().compile(pipeline_func=nested_pipeline_opt_inputs_nil, package_path=__file__.replace('.py', '_compiled.yaml'))
+
--- a/samples/v2/nested_pipeline_opt_inputs_parent_level.py
+++ b/samples/v2/nested_pipeline_opt_inputs_parent_level.py
@@ -1,0 +1,58 @@
+from typing import Optional
+
+from kfp import compiler, dsl
+
+
+@dsl.component()
+def component_nil_str_default(componentInput: str = None):
+    if componentInput != 'Input - parent pipeline':
+        raise ValueError(f"componentInput should be 'Input - parent pipeline' but is {componentInput}")
+
+@dsl.component()
+def component_str_default(componentInput: str = 'Input - component'):
+    if componentInput != 'Input - parent pipeline':
+        raise ValueError(f"componentInput should be 'Input - parent pipeline' but is {componentInput}")
+
+@dsl.component()
+def component_nil_int_default(componentInput: int = None):
+    if componentInput != 1:
+        raise ValueError(f"componentInput should be 1 but is {componentInput}")
+
+@dsl.component()
+def component_int_default(componentInput: int = 0):
+    if componentInput != 1:
+        raise ValueError(f"componentInput should be 1 but is {componentInput}")
+
+@dsl.component()
+def component_nil_bool_default(componentInput: bool = None):
+    if componentInput != True:
+        raise ValueError(f"componentInput should be True but is {componentInput}")
+
+@dsl.component()
+def component_bool_default(componentInput: bool = False):
+    if componentInput != True:
+        raise ValueError(f"componentInput should be True but is {componentInput}")
+
+@dsl.pipeline()
+def nested_pipeline_non_nil_defaults(nestedInputStr: str = 'Input - nested pipeline', nestedInputInt: int = 0, nestedInputBool: bool = False):
+    component_str_default(componentInput=nestedInputStr).set_caching_options(False)
+    component_int_default(componentInput=nestedInputInt).set_caching_options(False)
+    component_bool_default(componentInput=nestedInputBool).set_caching_options(False)
+
+@dsl.pipeline()
+def nested_pipeline_nil_defaults(nestedInputStr: str = 'Input - nested pipeline', nestedInputInt: int = None, nestedInputBool: bool = None):
+    component_nil_str_default(componentInput=nestedInputStr).set_caching_options(False)
+    component_nil_int_default(componentInput=nestedInputInt).set_caching_options(False)
+    component_nil_bool_default(componentInput=nestedInputBool).set_caching_options(False)
+
+
+@dsl.pipeline()
+def nested_pipeline_opt_inputs_parent_level(inputStr: str = 'Input - parent pipeline', inputInt: int = 1, inputBool: bool = True):
+    # verifies that the parent pipeline input overrides both nested pipeline-level and component-level default values.
+    nested_pipeline_non_nil_defaults(nestedInputStr=inputStr, nestedInputInt=inputInt, nestedInputBool=inputBool).set_caching_options(False)
+    # verifies that the parent pipeline input overrides both nested pipeline-level & component-level nil default input values.
+    nested_pipeline_nil_defaults(nestedInputStr=inputStr, nestedInputInt=inputInt, nestedInputBool=inputBool).set_caching_options(False)
+
+if __name__ == "__main__":
+    compiler.Compiler().compile(pipeline_func=nested_pipeline_opt_inputs_parent_level, package_path=__file__.replace(".py", "_compiled.yaml"))
+
--- a/backend/src/v2/cmd/driver/main.go
+++ b/backend/src/v2/cmd/driver/main.go
@@ -195,6 +195,7 @@ func drive() (err error) {
 		PipelineLogLevel: *logLevel,
 		PublishLogs:      *publishLogs,
 		CacheDisabled:    *cacheDisabledFlag,
+		DriverType:       *driverType,
 	}
 	var execution *driver.Execution
 	var driverErr error
--- a/backend/src/v2/driver/driver.go
+++ b/backend/src/v2/driver/driver.go
@@ -65,6 +65,8 @@ type Options struct {
 	PublishLogs string
 
 	CacheDisabled bool
+
+	DriverType string
 }
 
 // Identifying information used for error messages
--- a/samples/v2/nested_pipeline_opt_input_child_level.py
+++ b/samples/v2/nested_pipeline_opt_input_child_level.py
@@ -1,0 +1,60 @@
+
+from kfp import compiler
+from kfp import dsl
+
+
+@dsl.component()
+def component_a_str(componentInputStr: str = None):
+    if componentInputStr != 'Input - pipeline':
+        raise ValueError(f"componentInputStr should be 'Input - pipeline' but is {componentInputStr}")
+
+
+@dsl.component()
+def component_b_str(componentInputStr: str = None):
+    if componentInputStr != 'Input 2 - nested pipeline':
+        raise ValueError(f"componentInputStr should be 'Input 2 - nested pipeline' but is {componentInputStr}")
+
+@dsl.component()
+def component_a_int(componentInputInt: int = None):
+    if componentInputInt != 1:
+        raise ValueError(f"componentInputInt should be 1 but is {componentInputInt}")
+
+@dsl.component()
+def component_b_int(componentInputInt: int = None):
+    if componentInputInt != 0:
+        raise ValueError(f"componentInputInt should be 0 but is {componentInputInt}")
+
+@dsl.component()
+def component_a_bool(componentInputBool: bool = None):
+    if componentInputBool != True:
+        raise ValueError(f"componentInputBool should be True but is {componentInputBool}")
+
+@dsl.component()
+def component_b_bool(componentInputBool: bool = None):
+    if componentInputBool != False:
+        raise ValueError(f"componentInputBool should be False but is {componentInputBool}")
+
+@dsl.pipeline()
+def nested_pipeline(nestedInputStr1: str = 'Input 1 - nested pipeline',  nestedInputStr2: str = 'Input 2 - nested pipeline',
+                        nestedInputInt1: int = 0, nestedInputInt2: int = 0,
+                        nestedInputBool1: bool = False, nestedInputBool2: bool = False):
+    component_a_str(componentInputStr=nestedInputStr1).set_caching_options(False)
+    component_b_str(componentInputStr=nestedInputStr2).set_caching_options(False)
+
+    component_a_int(componentInputInt=nestedInputInt1).set_caching_options(False)
+    component_b_int(componentInputInt=nestedInputInt2).set_caching_options(False)
+
+    component_a_bool(componentInputBool=nestedInputBool1).set_caching_options(False)
+    component_b_bool(componentInputBool=nestedInputBool2).set_caching_options(False)
+
+
+@dsl.pipeline()
+def nested_pipeline_opt_input_child_level():
+    # validate that input value overrides default value, and that when input is not provided, default is used.
+    nested_pipeline(nestedInputStr1='Input - pipeline', nestedInputInt1=1, nestedInputBool1=True).set_caching_options(False)
+
+
+
+if __name__ == '__main__':
+    compiler.Compiler().compile(pipeline_func=nested_pipeline_opt_input_child_level, package_path=__file__.replace('.py', '_compiled.yaml'))
+
--- a/samples/v2/sample_test.py
+++ b/samples/v2/sample_test.py
@@ -42,6 +42,9 @@ import two_step_pipeline_containerized
 import yaml
 import pipeline_with_retry
 import pipeline_with_input_status_state
+import nested_pipeline_opt_inputs_parent_level
+import nested_pipeline_opt_inputs_nil
+import nested_pipeline_opt_input_child_level
 
 _MINUTE = 60  # seconds
 _DEFAULT_TIMEOUT = 10 * _MINUTE
@@ -174,6 +177,9 @@ class SampleTest(unittest.TestCase):
                 pipeline_func=collected_parameters.collected_param_pipeline),
             TestCase(pipeline_func=pipeline_with_retry.retry_pipeline),
             TestCase(pipeline_func=pipeline_with_input_status_state.status_state_pipeline),
+            TestCase(pipeline_func=nested_pipeline_opt_inputs_parent_level.nested_pipeline_opt_inputs_parent_level),
+            TestCase(pipeline_func=nested_pipeline_opt_input_child_level.nested_pipeline_opt_input_child_level),
+            TestCase(pipeline_func=nested_pipeline_opt_inputs_nil.nested_pipeline_opt_inputs_nil),
         ]
 
         with ThreadPoolExecutor() as executor:


samples/v2/nested_pipeline_parent_and_child_and_component_optional_input.py
```diff
@@ -0,0 +1,39 @@
...
+@dsl.pipeline()
+def nested_pipeline_and_component_non_nil_defaults(nestedInput: str = 'Input - nested pipeline'):
+    component_non_nil_default(componentInput=nestedInput).set_caching_options(False)
+
+@dsl.pipeline()
+def pipeline(input: str = 'Input - parent pipeline'):
+    # verifies that the parent pipeline input overrides a nested pipeline-level default input value.
+    nested_pipeline_non_nil_defaults(nestedInput=input).set_caching_options(False)
+    # verifies that the parent pipeline input overrides a nested pipeline-level nil default input value.
+    nested_pipeline_non_nil_defaults(nestedInput=input).set_caching_options(False)
```
hbelmiro: According to the comment, it seems you meant:
```suggestion
    nested_pipeline_nil_defaults(nestedInput=input).set_caching_options(False)
```

backend/src/v2/compiler/argocompiler/dag.go
```diff
@@ -230,25 +233,63 @@ func (c *workflowCompiler) task(name string, task *pipelinespec.PipelineTaskSpec
...
+					taskRuntimeConfig[param] = runtimeConfigParams[param]
+				}
+			}
+		}
+		if len(parameters) > 0 || len(artifacts) > 0 {
+			task.Inputs = &pipelinespec.TaskInputsSpec{
+				Artifacts:  artifacts,
+				Parameters: parameters}
+		}
+	}
```
hbelmiro: Can we move all of this, except the `isIteratorTask` declaration, inside the `case *pipelinespec.ComponentSpec_Dag:` block so it runs only if that condition applies?

backend/src/v2/compiler/argocompiler/dag.go
```diff
@@ -230,25 +233,63 @@ func (c *workflowCompiler) task(name string, task *pipelinespec.PipelineTaskSpec
...
-		driver, driverOutputs, err := c.dagDriverTask(driverTaskName, dagDriverInputs{
+		driverInputs := dagDriverInputs{
 			parentDagID:    inputs.parentDagID,
 			component:      componentSpecPlaceholder,
 			task:           taskSpecJson,
 			iterationIndex: inputs.iterationIndex,
-		})
+		}
+		driverInputs.runtimeConfig = &pipelinespec.PipelineJob_RuntimeConfig{ParameterValues: taskRuntimeConfig}
+		driver, driverOutputs, err := c.dagDriverTask(driverTaskName, driverInputs)
```
hbelmiro: nit: How about setting only `driverInputs.runtimeConfig` without changing the other lines?

```go
		driver, driverOutputs, err := c.dagDriverTask(driverTaskName, dagDriverInputs{
			parentDagID:    inputs.parentDagID,
			component:      componentSpecPlaceholder,
			task:           taskSpecJson,
			iterationIndex: inputs.iterationIndex,
/* just add this line */runtimeConfig:  &pipelinespec.PipelineJob_RuntimeConfig{ParameterValues: taskRuntimeConfig},
 		})
```

alyssacgoins: Updated

backend/src/v2/compiler/argocompiler/dag.go
alyssacgoins: Updated

samples/v2/nested_pipeline_parent_and_child_and_component_optional_input.py
alyssacgoins: Updated

backend/src/v2/compiler/argocompiler/argo.go
```diff
@@ -83,7 +83,21 @@ func Compile(jobArg *pipelinespec.PipelineJob, kubernetesSpecArg *pipelinespec.S
 			}
 		}
 	}
-
+	// fill nested pipeline and component default parameters to PipelineJob
+	for _, componentSpec := range spec.Components {
+		if componentSpec.GetInputDefinitions() != nil {
```
mprahl: I think if we check that `componentSpec.GetDag() != nil` then we know it's a nested pipeline of some kind.
```suggestion
		if componentSpec.GetDag() != nil && componentSpec.GetInputDefinitions() != nil {
```

backend/src/v2/compiler/argocompiler/argo.go
```diff
@@ -83,7 +83,21 @@ func Compile(jobArg *pipelinespec.PipelineJob, kubernetesSpecArg *pipelinespec.S
...
+			for param, paramSpec := range componentSpec.GetInputDefinitions().GetParameters() {
+				_, ok = job.RuntimeConfig.ParameterValues[param]
+				if !ok {
+					if paramSpec.GetDefaultValue() != nil {
+						job.RuntimeConfig.ParameterValues[param] = paramSpec.GetDefaultValue()
```
mprahl: What if the `param` from the root pipeline is the same name as this nested pipeline param? Won't it get overridden?

alyssacgoins: If root pipeline and nested pipeline contain a parameter of the same name, the `job.RuntimeConfig.ParameterValues[param]` entry won't get overridden because of the `_, ok = job.RuntimeConfig.ParameterValues[param]` check on lines 90-91 - `job.RuntimeConfig.ParameterValues[param]` only executes if there is no existing entry

backend/src/v2/compiler/argocompiler/argo.go
alyssacgoins: Updated

backend/src/v2/compiler/argocompiler/argo.go
```diff
@@ -83,7 +83,21 @@ func Compile(jobArg *pipelinespec.PipelineJob, kubernetesSpecArg *pipelinespec.S
 			}
 		}
 	}
-
+	// fill nested pipeline default parameters to PipelineJob
+	for _, componentSpec := range spec.Components {
+		if componentSpec.GetDag() != nil && componentSpec.GetInputDefinitions() != nil {
+			for param, paramSpec := range componentSpec.GetInputDefinitions().GetParameters() {
+				_, ok = job.RuntimeConfig.ParameterValues[param]
```
mprahl: Am I reading this correctly that it assumes the pipeline input parameter and the nested pipeline input parameter are the same name?

mprahl:
@alyssacgoins I need to review your PR more indepth when I have a bit more time but I was originally thinking the solution may be something simple like this. To help me review your PR, could you please explain why this is naive? :sweat_smile: 

```diff
diff --git a/backend/src/v2/driver/resolve.go b/backend/src/v2/driver/resolve.go
index 98469d26b..6866ec201 100644
--- a/backend/src/v2/driver/resolve.go
+++ b/backend/src/v2/driver/resolve.go
@@ -270,6 +270,9 @@ func resolveInputs(
 		return inputs, nil
 	}
 
+	// TODO: Verify if this is always accurate.
+	isDagDriver := opts.Container == nil && opts.Task != nil
+
 	// Handle parameters.
 	for name, paramSpec := range task.GetInputs().GetParameters() {
 		v, err := resolveInputParameter(ctx, dag, pipeline, opts, mlmd, paramSpec, inputParams)
@@ -280,6 +283,17 @@ func resolveInputs(
 
 			componentParam, ok := opts.Component.GetInputDefinitions().GetParameters()[name]
 			if ok && componentParam != nil && componentParam.IsOptional {
+				if isDagDriver {
+					if componentParam.GetDefaultValue() != nil {
+						// If this is a nested pipeline, pass along the default value to the component.
+						inputs.ParameterValues[name] = componentParam.GetDefaultValue()
+					} else {
+						// If this is a nested pipeline, pass along the null value to the component.
+						inputs.ParameterValues[name] = structpb.NewNullValue()
+					}
+
+					continue
+				}
 				// If the resolved paramter was null and the component input parameter is optional, just skip setting
 				// it and the launcher will handle defaults.
 				continue
```

alyssacgoins:
> ```diff
> 	// If this is a nested pipeline, pass along the default value to the component.
> +						inputs.ParameterValues[name] = componentParam.GetDefaultValue()
> +					} else {
> +						// If this is a nested pipeline, pass along the null value to the component.
> +						inputs.ParameterValues[name] = structpb.NewNullValue()
> +					}
> ```

In my understanding, the block of code added in this diff is called within the `if err != nil ` and  statement, and only if the error is `ErrResolvedParameterNull`. However, when an input parameter is not located within a task, the error thrown is `"parent DAG does not have input parameter"`. 
I think the diff you've included is definitely part of the overall solution. The block of code added to backend/src/v2/compiler/argocompilerargo.task() is necessary because it adds the necessary nested dag parameters to component task inputs

backend/src/v2/compiler/argocompiler/argo.go
alyssacgoins: Updated

backend/src/v2/driver/resolve.go
```diff
@@ -269,7 +272,23 @@ func resolveInputs(
...
 	}
+	isDagDriver := opts.Container == nil && opts.Task != nil
+	// If dag driver: check executor task inputs against component inputs, and fill in default and null values as necessary.
+	if isDagDriver {
+		for name, paramSpec := range opts.Component.GetInputDefinitions().GetParameters() {
+			_, ok := task.Inputs.GetParameters()[name]
+			if !ok {
+				if paramSpec.GetDefaultValue() != nil {
+					// If no value was input, pass along the default value to the component.
+					inputs.ParameterValues[name] = structpb.NewStringValue(paramSpec.GetDefaultValue().GetStringValue())
```
mprahl: Did you mean to only support string defaults?

I think you'll need to expand your tests to have defaults of different types.

backend/src/v2/driver/resolve.go
```diff
@@ -269,7 +272,23 @@ func resolveInputs(
 		}
 		return inputs, nil
 	}
+	isDagDriver := opts.Container == nil && opts.Task != nil
+	// If dag driver: check executor task inputs against component inputs, and fill in default and null values as necessary.
+	if isDagDriver {
+		for name, paramSpec := range opts.Component.GetInputDefinitions().GetParameters() {
+			_, ok := task.Inputs.GetParameters()[name]
+			if !ok {
```
mprahl:
```suggestion
			if !ok && paramSpec.IsOptional  {
```

backend/src/v2/driver/resolve.go
```diff
@@ -269,7 +272,23 @@ func resolveInputs(
...
+				if paramSpec.GetDefaultValue() != nil {
+					// If no value was input, pass along the default value to the component.
+					inputs.ParameterValues[name] = structpb.NewStringValue(paramSpec.GetDefaultValue().GetStringValue())
+				} else {
+					// If no default value is set, pass along the null value to the component.
```
mprahl:
```suggestion
					// If no default value is set, pass along the null value to the component. This is analogous to a pipeline run being submitted without optional pipeline input parameters.
```

backend/src/v2/driver/resolve.go
```diff
@@ -269,7 +272,23 @@ func resolveInputs(
 		}
 		return inputs, nil
 	}
+	isDagDriver := opts.Container == nil && opts.Task != nil
+	// If dag driver: check executor task inputs against component inputs, and fill in default and null values as necessary.
```
mprahl: Optional:
```suggestion
	// A DAG driver (not Root DAG driver) indicates this is likely the start of a nested pipeline.
	// Handle omitted optional pipeline input parameters similar to how they are handled on the
	// root pipeline.
```

backend/src/v2/driver/resolve.go
```diff
@@ -328,7 +347,7 @@ func resolveInputParameter(
 		}
 		v, ok := inputParams[componentInput]
 		if !ok {
-			return nil, paramError(fmt.Errorf("parent DAG does not have input parameter %s", componentInput))
+			return nil, fmt.Errorf("%w: %s", ErrParentDAGDoesNotHaveInputParameter, componentInput)
```
mprahl: It seems this special error is not being used, so your PR can revert back to the way it was.

alyssacgoins: Updated

backend/src/v2/driver/resolve.go
alyssacgoins: Updated

backend/src/v2/driver/resolve.go
alyssacgoins: Updated

backend/src/v2/driver/resolve.go
alyssacgoins: Updated

backend/src/v2/driver/resolve.go
alyssacgoins: Updated

backend/src/v2/driver/resolve.go
```diff
@@ -269,7 +271,26 @@ func resolveInputs(
...
+	// A DAG driver (not Root DAG driver) indicates this is likely the start of a nested pipeline.
+	// Handle omitted optional pipeline input parameters similar to how they are handled on the root pipeline.
+	if isDagDriver {
+		for name, paramSpec := range opts.Component.GetInputDefinitions().GetParameters() {
+			_, ok := task.Inputs.GetParameters()[name]
+			if !ok && paramSpec.IsOptional {
+				defaultValue, err := structpb.NewValue(paramSpec.GetDefaultValue())
+				if paramSpec.GetDefaultValue() != nil && err != nil {
+					// If no value was input, pass along the default value to the component.
+					inputs.ParameterValues[name] = defaultValue
```
mprahl:
```suggestion
				if paramSpec.GetDefaultValue() != nil && err != nil {
					// If no value was input, pass along the default value to the component.
					inputs.ParameterValues[name] = structpb.NewValue(paramSpec.GetDefaultValue())
```

samples/v2/nested_pipeline_opt_inputs_nil.py
```diff
@@ -0,0 +1,35 @@
...
+
+
+@dsl.component()
+def component_str(componentInput1: str = None):
+    if componentInput1 is not '':
+        raise ValueError(f"componentInput2 should be '' but is {componentInput1}")
+
+@dsl.component()
+def component_int(componentInput1: int = None):
+    if componentInput1 is not '':
```
mprahl: Why is componentInput1 a string?

samples/v2/nested_pipeline_opt_inputs_nil.py
```diff
@@ -0,0 +1,35 @@
...
+        raise ValueError(f"componentInput2 should be '' but is {componentInput1}")
+
+@dsl.component()
+def component_bool(componentInput1: bool = None):
+    if componentInput1 is not '':
```
mprahl: Why is componentInput1 a string?

samples/v2/nested_pipeline_opt_inputs_nil.py
```diff
@@ -0,0 +1,35 @@
...
+        raise ValueError(f"componentInput2 should be '' but is {componentInput1}")
```
mprahl: Typo in the variable name

samples/v2/nested_pipeline_opt_inputs_parent_level.py
```diff
@@ -0,0 +1,58 @@
...
+    if componentInput != 1:
+        raise ValueError(f"componentInput should be 1 but is {componentInput}")
+
+@dsl.component()
+def component_nil_bool_default(componentInput: bool = None):
+    if componentInput != True:
+        raise ValueError(f"componentInput should be True but is {componentInput}")
+
+@dsl.component()
+def component_bool_default(componentInput: bool = False):
```
mprahl, line 32: This looks like it's not used.

samples/v2/nested_pipeline_opt_inputs_nil.py
alyssacgoins: Fixed

samples/v2/nested_pipeline_opt_inputs_parent_level.py
alyssacgoins: Fixed - is used now

samples/v2/nested_pipeline_opt_inputs_nil.py
alyssacgoins: Fixed - replaced empty string with None

samples/v2/nested_pipeline_opt_inputs_nil.py
alyssacgoins: Fixed - replaced empty string with None

backend/src/v2/driver/resolve.go
```diff
@@ -269,7 +271,25 @@ func resolveInputs(
...
+	isDagDriver := opts.Container == nil && opts.Task != nil
+	// A DAG driver (not Root DAG driver) indicates this is likely the start of a nested pipeline.
+	// Handle omitted optional pipeline input parameters similar to how they are handled on the root pipeline.
+	if isDagDriver {
+		for name, paramSpec := range opts.Component.GetInputDefinitions().GetParameters() {
+			_, ok := task.Inputs.GetParameters()[name]
+			if !ok && paramSpec.IsOptional {
+				if paramSpec.GetDefaultValue() != nil {
+					// If no value was input, pass along the default value to the component.
+					inputs.ParameterValues[name], _ = structpb.NewValue(paramSpec.GetDefaultValue())
```
mprahl:
```suggestion
					inputs.ParameterValues[name] = paramSpec.GetDefaultValue()
```

backend/src/v2/driver/resolve.go
```diff
@@ -269,7 +271,25 @@ func resolveInputs(
...
+				} else {
+					//  If no default value is set, pass along the null value to the component.
+					//	This is analogous to a pipeline run being submitted without optional pipeline input parameters.
+					inputs.ParameterValues[name] = structpb.NewNullValue()
+				}
```
mprahl, line 288: After you moved the logic to outside of the `for name, paramSpec := range task.GetInputs().GetParameters() {` loop, we can just omit the null value and I think that'll solve your CI failure.
```suggestion
				}
```

backend/src/v2/driver/resolve.go
```diff
@@ -269,7 +271,25 @@ func resolveInputs(
...
+		for name, paramSpec := range opts.Component.GetInputDefinitions().GetParameters() {
+			_, ok := task.Inputs.GetParameters()[name]
+			if !ok && paramSpec.IsOptional {
+				if paramSpec.GetDefaultValue() != nil {
+					// If no value was input, pass along the default value to the component.
+					inputs.ParameterValues[name], _ = structpb.NewValue(paramSpec.GetDefaultValue())
+				} else {
+					//  If no default value is set, pass along the null value to the component.
+					//	This is analogous to a pipeline run being submitted without optional pipeline input parameters.
+					inputs.ParameterValues[name] = structpb.NewNullValue()
```
mprahl, line 296: I think the CI is also failing because `handleParamTypeValidationAndConversion` is failing due to this being a null value. I think we also need code to allow that to continue if it's a DAG driver.

backend/src/v2/driver/resolve.go
alyssacgoins: Updated

backend/src/v2/driver/resolve.go
alyssacgoins: Updated

backend/src/v2/driver/resolve.go
```diff
@@ -193,7 +194,12 @@ func resolveInputs(
 				}
 				switch v := value.GetKind().(type) {
 				case *structpb.Value_NullValue:
-					return fmt.Errorf("got null for input parameter %q", name)
+					// If isDagDriver is true, allow input parameter to have null value.
+					if isDagDriver {
+						return nil
```
mprahl: We still need to validate the other inputs.
```suggestion
						continue
```

backend/src/v2/driver/resolve.go
```diff
@@ -269,7 +275,25 @@ func resolveInputs(
...
 	}
+	// A DAG driver (not Root DAG driver) indicates this is likely the start of a nested pipeline.
+	// Handle omitted optional pipeline input parameters similar to how they are handled on the root pipeline.
+	if isDagDriver {
+		for name, paramSpec := range opts.Component.GetInputDefinitions().GetParameters() {
+			_, ok := task.Inputs.GetParameters()[name]
+			if !ok && paramSpec.IsOptional {
+				if paramSpec.GetDefaultValue() != nil {
+					// If no value was input, pass along the default value to the component.
+					//inputs.ParameterValues[name], _ = structpb.NewValue(paramSpec.GetDefaultValue())
```
mprahl:
```suggestion
```

backend/src/v2/driver/resolve.go
```diff
@@ -193,7 +194,12 @@ func resolveInputs(
 				}
 				switch v := value.GetKind().(type) {
 				case *structpb.Value_NullValue:
-					return fmt.Errorf("got null for input parameter %q", name)
+					// If isDagDriver is true, allow input parameter to have null value.
+					if isDagDriver {
+						return nil
+					} else {
```
mprahl: Optional: You don't need the else in this case since you're always continuing the loop in the if statement.

backend/src/v2/driver/resolve.go
```diff
@@ -269,7 +275,25 @@ func resolveInputs(
...
+			_, ok := task.Inputs.GetParameters()[name]
+			if !ok && paramSpec.IsOptional {
+				if paramSpec.GetDefaultValue() != nil {
+					// If no value was input, pass along the default value to the component.
+					//inputs.ParameterValues[name], _ = structpb.NewValue(paramSpec.GetDefaultValue())
+					inputs.ParameterValues[name] = paramSpec.GetDefaultValue()
+				} else {
+					//  If no default value is set, pass along the null value to the component.
+					//	This is analogous to a pipeline run being submitted without optional pipeline input parameters.
+					inputs.ParameterValues[name] = structpb.NewNullValue()
```
mprahl, line 296: Could you keep track of which ones were set to null so that `handleParamTypeValidationAndConversion` only does `continue` on those the driver set to null rather than on all null values?

backend/src/v2/driver/resolve.go
```diff
@@ -142,6 +142,7 @@ func resolveInputs(
 		}
 		return nil
 	}
+	isDagDriver := opts.Container == nil && opts.Task != nil
```
mprahl: After thinking about this, I think I'd be more comfortable with being explicit here with something like this:

```diff
diff --git a/backend/src/v2/cmd/driver/main.go b/backend/src/v2/cmd/driver/main.go
index 5833c0473..ea8dbe1b6 100644
--- a/backend/src/v2/cmd/driver/main.go
+++ b/backend/src/v2/cmd/driver/main.go
@@ -195,6 +195,7 @@ func drive() (err error) {
 		PipelineLogLevel: *logLevel,
 		PublishLogs:      *publishLogs,
 		CacheDisabled:    *cacheDisabledFlag,
+		DriverType:       *driverType,
 	}
 	var execution *driver.Execution
 	var driverErr error
diff --git a/backend/src/v2/driver/driver.go b/backend/src/v2/driver/driver.go
index 22c2b4880..1c7940b6f 100644
--- a/backend/src/v2/driver/driver.go
+++ b/backend/src/v2/driver/driver.go
@@ -65,6 +65,8 @@ type Options struct {
 	PublishLogs string
 
 	CacheDisabled bool
+
+	DriverType string
 }
 
 // Identifying information used for error messages
diff --git a/backend/src/v2/driver/resolve.go b/backend/src/v2/driver/resolve.go
index 7142bbbb9..60bd07c82 100644
--- a/backend/src/v2/driver/resolve.go
+++ b/backend/src/v2/driver/resolve.go
@@ -142,7 +142,7 @@ func resolveInputs(
 		}
 		return nil
 	}
-	isDagDriver := opts.Container == nil && opts.Task != nil
+	isDagDriver := opts.DriverType == "DAG"
 	handleParamTypeValidationAndConversion := func() error {
 		// TODO(Bobgy): verify whether there are inputs not in the inputs spec.
 		for name, spec := range inputsSpec.GetParameters() {
```

backend/src/v2/driver/resolve.go
alyssacgoins: Updated

backend/src/v2/driver/resolve.go
alyssacgoins: Fixed

backend/src/v2/driver/resolve.go
alyssacgoins: Oh good point!

backend/src/v2/driver/resolve.go
alyssacgoins: Updated - created slice on line 147 to track the parameters set to null by the driver

backend/src/v2/driver/resolve.go
alyssacgoins: Updated

backend/src/v2/driver/resolve.go
```diff
@@ -142,6 +143,10 @@ func resolveInputs(
 		}
 		return nil
 	}
+	// Track parameters set to nil by the driver (for the case in which optional pipeline input parameters are not
+	// included, and default value is nil).
+	var parametersSetNilByDriver []string
```
mprahl: A better data structure for this is a set since we can get O(1) retrieval and we know there won't be any duplicates. In Go, you typically denote a set with a type of `map[string]bool` so that you can do something like `if parametersSetNilByDriver[name] { continue }`.

backend/src/v2/driver/resolve.go
```diff
@@ -193,7 +198,11 @@ func resolveInputs(
 				}
 				switch v := value.GetKind().(type) {
 				case *structpb.Value_NullValue:
-					return fmt.Errorf("got null for input parameter %q", name)
+					// If isDagDriver is true and parameter was set to nil by driver, allow input parameter to have a
+					// nil value.
+					if isDagDriver && slices.Contains(parametersSetNilByDriver, name) {
+						continue
+					}
```
mprahl: As discussed in Slack:
```suggestion
					if slices.Contains(parametersSetNilByDriver, name) {
						continue
					}

					return fmt.Errorf("got null for input parameter %q", name)
```

backend/src/v2/driver/resolve.go
alyssacgoins: Makes sense! Updated

backend/src/v2/driver/resolve.go
alyssacgoins: Updated

samples/v2/nested_pipeline_opt_inputs_nil.py
```diff
@@ -0,0 +1,35 @@
+
+from kfp import compiler
+from kfp import dsl
+
+
+@dsl.component()
+def component_str(componentInput: str = None):
+    if componentInput is not '':
```
mprahl: I was expecting this to be `None` since the default value of `componentInput` is that. Do you know why it's becoming a string?

alyssacgoins: Currently resolve.go line 182 `case pipelinespec.ParameterType_STRING:` handles input string parameters of type  (in this case, input is type str though set to None) separately than every other type of input - the logic is set up as either `case String` or `default`. The string block (line 182) handles nil string input by creating an empty string. This is definitely something I can update to resolve to nil in the case of driver-set nil strings, if we think that's more logical

mprahl: @alyssacgoins thanks for the explanation! Yes, I think in this case, let's just treat it like we do for other types and let the component's default be used.

alyssacgoins: @mprahl awesome - updated now, added a `if parametersSetNilByDriver[name] {
						continue
					}` block on resolve.go line 186

mprahl (approved):
/lgtm :rocket:

PR #11982, opened 2025-06-16
feat(backend): update Argo Workflow Compiler to create workspace PVCs

VaniHaripriya:
Part of: https://github.com/kubeflow/pipelines/issues/12053
Resolve: https://github.com/kubeflow/pipelines/issues/12058

- This PR is part of the ongoing work on the [Pipeline Run Workspace feature](https://github.com/kubeflow/pipelines/issues/12053).
- This introduces support for defining default PersistentVolumeClaim (PVC) configurations at the API server level to preconfigure workspace settings for users.
- The compiler automatically injects `volumeClaimTemplates` into Argo Workflows when a workspace is defined in the pipeline.
- Defaults from the API server config are used unless overridden by user-defined settings in the pipeline.
Example config.json -
```
{
  "Workspace": {
    "VolumeClaimTemplateSpec": {
      "accessModes": ["ReadWriteMany"],
      "storageClassName": "my-storage"
    }
  }
}
```

**Checklist:**

### DIFF
--- a/backend/src/apiserver/config/config.json
+++ b/backend/src/apiserver/config/config.json
@@ -24,5 +24,14 @@
   "CacheEnabled": "true",
   "CRON_SCHEDULE_TIMEZONE": "UTC",
   "CACHE_IMAGE": "ghcr.io/containerd/busybox",
-  "CACHE_NODE_RESTRICTIONS": "false"
+  "CACHE_NODE_RESTRICTIONS": "false",
+  "Workspace": {
+    "size": "100Gi",
+    "VolumeClaimTemplateSpec": {
+      "accessModes": [
+        "ReadWriteOnce"
+      ],
+      "storageClassName": "standard-csi"
+    }
+  }
 }
--- a/backend/src/apiserver/main.go
+++ b/backend/src/apiserver/main.go
@@ -48,6 +48,7 @@ import (
 	"go.uber.org/zap/zapcore"
 	"google.golang.org/grpc"
 	"google.golang.org/grpc/reflection"
+	corev1 "k8s.io/api/core/v1"
 	ctrlclient "sigs.k8s.io/controller-runtime/pkg/client"
 	ctrllog "sigs.k8s.io/controller-runtime/pkg/log"
 	"sigs.k8s.io/controller-runtime/pkg/log/zap"
@@ -57,6 +58,8 @@ import (
 const (
 	executionTypeEnv = "ExecutionType"
 	launcherEnv      = "Launcher"
+	workspaceConfig  = "workspace"
+	workspaceSize    = "size"
 )
 
 var (
@@ -157,11 +160,20 @@ func main() {
 
 	}
 
+	var pvcSpec *corev1.PersistentVolumeClaimSpec
+	var defaultWorkspaceSize string
+	pvcSpec, defaultWorkspaceSize, err = getPVCSpec()
+	if err != nil {
+		glog.Fatalf("Failed to get Workspace PVC Spec: %v", err)
+	}
+
 	resourceManager := resource.NewResourceManager(
 		clientManager,
 		&resource.ResourceManagerOptions{
-			CollectMetrics: *collectMetricsFlag,
-			CacheDisabled:  !common.GetBoolConfigWithDefault("CacheEnabled", true),
+			CollectMetrics:       *collectMetricsFlag,
+			CacheDisabled:        !common.GetBoolConfigWithDefault("CacheEnabled", true),
+			DefaultWorkspace:     pvcSpec,
+			DefaultWorkspaceSize: defaultWorkspaceSize,
 		},
 	)
 	err = config.LoadSamples(resourceManager, *sampleConfigPath)
@@ -397,4 +409,23 @@ func initConfig() {
 	})
 
 	proxy.InitializeConfigWithEnv()
+}
+
+// getPVCSpec retrieves the default workspace PersistentVolumeClaimSpec and size from the config.
+// These defaults are used for workspace PVCs when users do not specify their own configuration.
+func getPVCSpec() (*corev1.PersistentVolumeClaimSpec, string, error) {
+	workspaceConfig := viper.Sub(workspaceConfig)
+	if workspaceConfig == nil {
+		glog.Info("No workspace config found; proceeding without a default PVC spec")
+		return nil, "", nil
+	}
+	var pvcSpec corev1.PersistentVolumeClaimSpec
+	if err := workspaceConfig.UnmarshalKey("volumeclaimtemplatespec", &pvcSpec); err != nil {
+		return nil, "", fmt.Errorf("failed to unmarshal workspace.volumeclaimtemplatespec: %w", err)
+	}
+	if len(pvcSpec.AccessModes) == 0 || pvcSpec.StorageClassName == nil || *pvcSpec.StorageClassName == "" {
+		return nil, "", fmt.Errorf("invalid workspace.volumeclaimtemplatespec: must specify accessModes and storageClassName")
+	}
+	defaultSize := workspaceConfig.GetString(workspaceSize)
+	return &pvcSpec, defaultSize, nil
 }
--- a/backend/src/apiserver/resource/resource_manager.go
+++ b/backend/src/apiserver/resource/resource_manager.go
@@ -101,8 +101,10 @@ type ClientManagerInterface interface {
 }
 
 type ResourceManagerOptions struct {
-	CollectMetrics bool `json:"collect_metrics,omitempty"`
-	CacheDisabled  bool `json:"cache_disabled,omitempty"`
+	CollectMetrics       bool                              `json:"collect_metrics,omitempty"`
+	CacheDisabled        bool                              `json:"cache_disabled,omitempty"`
+	DefaultWorkspace     *corev1.PersistentVolumeClaimSpec `json:"default_workspace,omitempty"`
+	DefaultWorkspaceSize string                            `json:"default_workspace_size,omitempty"`
 }
 
 type ResourceManager struct {
@@ -375,7 +377,7 @@ func (r *ResourceManager) CreatePipelineAndPipelineVersion(p *model.Pipeline, pv
 	if pipelineSpecURI != "" {
 		pv.PipelineSpecURI = pipelineSpecURI
 	}
-	tmpl, err := template.New(pipelineSpecBytes, r.options.CacheDisabled)
+	tmpl, err := template.New(pipelineSpecBytes, r.options.CacheDisabled, r.options.DefaultWorkspace)
 	if err != nil {
 		return nil, nil, util.Wrap(err, "Failed to create a pipeline and a pipeline version due to template creation error")
 	}
@@ -508,9 +510,10 @@ func (r *ResourceManager) CreateRun(ctx context.Context, run *model.Run) (*model
 	}
 	run.RunDetails.CreatedAtInSec = r.time.Now().Unix()
 	runWorkflowOptions := template.RunWorkflowOptions{
-		RunId:         run.UUID,
-		RunAt:         run.RunDetails.CreatedAtInSec,
-		CacheDisabled: r.options.CacheDisabled,
+		RunId:            run.UUID,
+		RunAt:            run.CreatedAtInSec,
+		CacheDisabled:    r.options.CacheDisabled,
+		DefaultWorkspace: r.options.DefaultWorkspace,
 	}
 	executionSpec, err := tmpl.RunWorkflow(run, runWorkflowOptions)
 	if err != nil {
@@ -1107,7 +1110,7 @@ func (r *ResourceManager) CreateJob(ctx context.Context, job *model.Job) (*model
 			return nil, util.Wrap(err, "Failed to validate the input parameters on the latest pipeline version")
 		}
 
-		tmpl, err := template.New(manifest, r.options.CacheDisabled)
+		tmpl, err := template.New(manifest, r.options.CacheDisabled, r.options.DefaultWorkspace)
 		if err != nil {
 			return nil, util.Wrap(err, "Failed to fetch a template with an invalid pipeline spec manifest")
 		}
@@ -1493,7 +1496,7 @@ func (r *ResourceManager) fetchTemplateFromPipelineSpec(pipelineSpec *model.Pipe
 			return nil, "", util.NewInvalidInputError("Failed to fetch a template with an empty pipeline spec manifest")
 		}
 	}
-	tmpl, err := template.New([]byte(manifest), r.options.CacheDisabled)
+	tmpl, err := template.New([]byte(manifest), r.options.CacheDisabled, r.options.DefaultWorkspace)
 	if err != nil {
 		return nil, "", util.Wrap(err, "Failed to fetch a template with an invalid pipeline spec manifest")
 	}
@@ -1652,7 +1655,7 @@ func (r *ResourceManager) CreatePipelineVersion(pv *model.PipelineVersion) (*mod
 	}
 
 	// Create a template
-	tmpl, err := template.New(pipelineSpecBytes, r.options.CacheDisabled)
+	tmpl, err := template.New(pipelineSpecBytes, r.options.CacheDisabled, r.options.DefaultWorkspace)
 	if err != nil {
 		return nil, util.Wrap(err, "Failed to create a pipeline version due to template creation error")
 	}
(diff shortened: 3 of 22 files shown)


hbelmiro (commented):
/lgtm

backend/src/apiserver/main.go
```diff
@@ -157,11 +158,18 @@ func main() {
 
 	}
 
+	var pvcSpec corev1.PersistentVolumeClaimSpec
+	err = viper.Sub("workspace").UnmarshalKey("volumeclaimtemplatespec", &pvcSpec)
```
HumairAK: workspace is an optional config, so this can be nil, also we should probably validate the unmarshalled value, at the minimum it should have accessmodes and storageclassname, Junie suggested this function, I would suggest something similar as well: 

```go
func getPVCSpec() (*corev1.PersistentVolumeClaimSpec, error) {
    var pvcSpec corev1.PersistentVolumeClaimSpec
    
    workspaceConfig := viper.Sub("workspace")
    if workspaceConfig == nil {
        return nil, fmt.Errorf("workspace configuration not found")
    }
    
    if err := workspaceConfig.UnmarshalKey("volumeclaimtemplatespec", &pvcSpec); err != nil {
        return nil, fmt.Errorf("failed to unmarshal workspace.volumeclaimtemplatespec: %w", err)
    }
        
    return &pvcSpec, nil
}
```

backend/src/apiserver/resource/resource_manager.go
```diff
@@ -101,8 +101,9 @@ type ClientManagerInterface interface {
 }
 
 type ResourceManagerOptions struct {
-	CollectMetrics bool `json:"collect_metrics,omitempty"`
-	CacheDisabled  bool `json:"cache_disabled,omitempty"`
+	CollectMetrics bool                              `json:"collect_metrics,omitempty"`
+	CacheDisabled  bool                              `json:"cache_disabled,omitempty"`
+	DefaultPVCSpec *corev1.PersistentVolumeClaimSpec `json:"workspace,omitempty"`
```
HumairAK: nit: my preference is to use workspace in the naming, maybe just `DefaultWorkspace`, this applies everywhere we are using this terminology

backend/src/v2/compiler/argocompiler/argo.go
```diff
@@ -43,9 +44,15 @@ type Options struct {
 	PipelineRoot string
 	// optional
 	CacheDisabled bool
+	// optional
+	DefaultPVCSpec *k8score.PersistentVolumeClaimSpec
 	// TODO(Bobgy): add an option -- dev mode, ImagePullPolicy should only be Always in dev mode.
 }
 
+const (
+	kfpWorkspaceSize = "10Gi"
```
HumairAK:
```suggestion
	defaultWorkspaceSize = "10Gi"
```

backend/src/apiserver/template/v2_template.go
```diff
@@ -110,7 +112,7 @@ func (t *V2Spec) ScheduledWorkflow(modelJob *model.Job) (*scheduledworkflow.Sche
 
 	var obj interface{}
 	if util.CurrentExecutionType() == util.ArgoWorkflow {
-		obj, err = argocompiler.Compile(job, kubernetesSpec, &argocompiler.Options{CacheDisabled: t.cacheDisabled})
+		obj, err = argocompiler.Compile(job, kubernetesSpec, &argocompiler.Options{CacheDisabled: t.cacheDisabled, DefaultPVCSpec: t.defaultPVCSpec})
 	}
```
HumairAK: nit: avoid long lines for improved readability: 

```go
	if util.CurrentExecutionType() == util.ArgoWorkflow {
		opts := &argocompiler.Options{
			CacheDisabled: t.cacheDisabled,
			DefaultPVCSpec: t.defaultPVCSpec,
		}
		obj, err = argocompiler.Compile(job, kubernetesSpec, opts)
	}
```

backend/src/v2/compiler/visitor.go
```diff
@@ -109,7 +109,7 @@ func (state *pipelineDFS) dfs(name string, component *pipelinespec.ComponentSpec
 		}
 
 		// Add kubernetes spec to annotation
-		if state.kubernetesSpec != nil {
+		if state.kubernetesSpec != nil && state.kubernetesSpec.DeploymentSpec != nil {
 			kubernetesExecSpec, ok := state.kubernetesSpec.DeploymentSpec.Executors[executorLabel]
 			if ok {
 				state.visitor.AddKubernetesSpec(name, kubernetesExecSpec)
```
HumairAK: while we are here, I see that the error from `AddKubernetesSpec` is  not handled, could you add simple handling here?

backend/src/v2/compiler/argocompiler/argo.go
```diff
@@ -93,6 +100,67 @@ func Compile(jobArg *pipelinespec.PipelineJob, kubernetesSpecArg *pipelinespec.S
 			return nil, fmt.Errorf("bug: cloned Kubernetes spec message does not have expected type")
 		}
 	}
+	var volumeClaimTemplates []k8score.PersistentVolumeClaim
+	if kubernetesSpec != nil {
```
HumairAK: nit: if possible it would be nice to move this to a helper function for improved readability

backend/src/v2/compiler/argocompiler/argo.go
```diff
@@ -93,6 +100,67 @@ func Compile(jobArg *pipelinespec.PipelineJob, kubernetesSpecArg *pipelinespec.S
...
+		if pipelineConfig := kubernetesSpec.GetPipelineConfig(); pipelineConfig != nil {
+			if workspace := pipelineConfig.GetWorkspace(); workspace != nil {
```
HumairAK: optional: the triple nested structure feels unnecessary since we don't have any `else` branches, in my opinion you can improve readability by just having one if statement, something like: 

```go
	hasPipelineConfig := kubernetesSpec != nil && kubernetesSpec.GetPipelineConfig() != nil
	hasWorkspace := hasPipelineConfig && kubernetesSpec.GetPipelineConfig().GetWorkspace() != nil
	if hasWorkspace {
		workspace := kubernetesSpec.GetPipelineConfig().GetWorkspace()
		.... = handleWorkspace(...)
```

backend/src/v2/compiler/argocompiler/argo.go
```diff
@@ -93,6 +100,67 @@ func Compile(jobArg *pipelinespec.PipelineJob, kubernetesSpecArg *pipelinespec.S
...
+				sizeStr := workspace.GetSize()
+				if sizeStr == "" {
+					sizeStr = kfpWorkspaceSize
+				}
+
+				if k8sWorkspace := workspace.GetKubernetes(); k8sWorkspace != nil {
+					volumeName := "kfp-workspace"
```
HumairAK: let's make this a constant like we did with the size and use it directly

backend/src/v2/compiler/argocompiler/argo.go
```diff
@@ -93,6 +100,67 @@ func Compile(jobArg *pipelinespec.PipelineJob, kubernetesSpecArg *pipelinespec.S
...
+
+					var pvcSpec k8score.PersistentVolumeClaimSpec
+					if opts.DefaultPVCSpec != nil {
```
HumairAK: you should verify opts is not nil

backend/src/v2/compiler/argocompiler/argo.go
```diff
@@ -93,6 +100,67 @@ func Compile(jobArg *pipelinespec.PipelineJob, kubernetesSpecArg *pipelinespec.S
...
+			if workspace := pipelineConfig.GetWorkspace(); workspace != nil {
+				sizeStr := workspace.GetSize()
+				if sizeStr == "" {
+					sizeStr = kfpWorkspaceSize
+				}
+
+				if k8sWorkspace := workspace.GetKubernetes(); k8sWorkspace != nil {
+					volumeName := "kfp-workspace"
+
+					var pvcSpec k8score.PersistentVolumeClaimSpec
```
HumairAK: I think it would be useful to add some inline code comments to explain that - we only need to use the default pvc spec when the user requests a workspace, so we may populate it with missing defaults (like storage classes, etc.)

backend/src/v2/compiler/argocompiler/argo.go
```diff
@@ -93,6 +100,67 @@ func Compile(jobArg *pipelinespec.PipelineJob, kubernetesSpecArg *pipelinespec.S
...
 		}
 	}
+	var volumeClaimTemplates []k8score.PersistentVolumeClaim
+	if kubernetesSpec != nil {
+		if pipelineConfig := kubernetesSpec.GetPipelineConfig(); pipelineConfig != nil {
+			if workspace := pipelineConfig.GetWorkspace(); workspace != nil {
+				sizeStr := workspace.GetSize()
+				if sizeStr == "" {
+					sizeStr = kfpWorkspaceSize
+				}
```
HumairAK: correct me if I'm wrong here - but what if the size is set via the apiserver config, it doesn't seem like we are handling that case?

backend/src/v2/compiler/argocompiler/argo.go
```diff
@@ -93,6 +100,67 @@ func Compile(jobArg *pipelinespec.PipelineJob, kubernetesSpecArg *pipelinespec.S
...
+
+				if k8sWorkspace := workspace.GetKubernetes(); k8sWorkspace != nil {
+					volumeName := "kfp-workspace"
+
+					var pvcSpec k8score.PersistentVolumeClaimSpec
+					if opts.DefaultPVCSpec != nil {
+						pvcSpec = *opts.DefaultPVCSpec.DeepCopy()
+					}
+					if k8sWorkspacePVCSpec := k8sWorkspace.GetPvcSpecPatch(); k8sWorkspacePVCSpec != nil {
```
HumairAK: optional: a more extensible approach would be to implement a pvcPatch helper that can be more generally used in the codebase under util package, for example from gemini I got: 

```go
import (
	k8score "k8s.io/api/core/v1"
)

// PatchPVCSpec applies fields from `patch` into `base` where patch has non-zero values.
// It mutates and returns the `base` spec.
func PatchPVCSpec(base, patch *k8score.PersistentVolumeClaimSpec) *k8score.PersistentVolumeClaimSpec {
	if base == nil || patch == nil {
		return base
	}

	// Patch AccessModes if non-empty
	if len(patch.AccessModes) > 0 {
		base.AccessModes = patch.AccessModes
	}

	// Patch Selector if provided
	if patch.Selector != nil {
		base.Selector = patch.Selector.DeepCopy()
	}

	// Patch Resources
	if patch.Resources.Requests != nil {
		if base.Resources.Requests == nil {
			base.Resources.Requests = make(k8score.ResourceList)
		}
		for k, v := range patch.Resources.Requests {
			base.Resources.Requests[k] = v
		}
	}

	// Patch VolumeName
	if patch.VolumeName != "" {
		base.VolumeName = patch.VolumeName
	}

	// Patch StorageClassName
	if patch.StorageClassName != nil {
		base.StorageClassName = patch.StorageClassName
	}

	// Patch VolumeMode
	if patch.VolumeMode != nil {
		base.VolumeMode = patch.VolumeMode
	}

	// Patch DataSource and DataSourceRef
	if patch.DataSource != nil {
		base.DataSource = patch.DataSource.DeepCopy()
	}
	if patch.DataSourceRef != nil {
		base.DataSourceRef = patch.DataSourceRef.DeepCopy()
	}

	return base
}

```

(untested)

HumairAK: then you can just call `PatchPVCSpec(opts.DefaultPVCSpec, k8sWorkspacePVCSpec)` 
if you go this route, I would carete a separate k8s_patches.go file in utils package, because I'm pretty sure we'll be adding more such helpers in the future

backend/src/v2/compiler/argocompiler/argo.go
HumairAK: then we can also add unit tests for this code

HumairAK:
@VaniHaripriya can you create a broder github issue to "Implement the workspace feature KFP, so we can link back to it in our this and future prs? and link it in the original pr description

VaniHaripriya:
> @VaniHaripriya can you create a broder github issue to "Implement the workspace feature KFP, so we can link back to it in our this and future prs? and link it in the original pr description

@HumairAK Sure, will create one and add it in the description.

backend/src/v2/compiler/argocompiler/argo.go
VaniHaripriya: Good catch @HumairAK , updated the logic to handle this case.

backend/src/apiserver/main.go
```diff
@@ -398,3 +408,21 @@ func initConfig() {
 
 	proxy.InitializeConfigWithEnv()
 }
+
+// Update getPVCSpec to also return a default size if present in the config
+func getPVCSpec() (*corev1.PersistentVolumeClaimSpec, string, error) {
+	workspaceConfig := viper.Sub("workspace")
```
HumairAK: let's convert all the viper configs to constant values and put them to the top of the file, this includes `workspace`, `size`, and any other such config value (e.g. see `executionTypeEnv`)

backend/src/v2/compiler/argocompiler/argo.go
```diff
@@ -43,9 +45,18 @@ type Options struct {
...
 	CacheDisabled bool
+	// optional
+	DefaultWorkspace *k8score.PersistentVolumeClaimSpec
+	// optional, default workspace size from API server config
+	DefaultWorkspaceSize string
 	// TODO(Bobgy): add an option -- dev mode, ImagePullPolicy should only be Always in dev mode.
 }
 
+const (
+	fallbackWorkspaceSize = "10Gi"
```
HumairAK: to clarify, this is if no size is set in apiserverconfig but we specify workspace in apiserver config right? if so maybe a inline comment here would help clarify this for future devs

backend/src/v2/compiler/argocompiler/argo.go
```diff
@@ -434,3 +461,69 @@ var dummyImages = map[string]bool{
...
+	}
+
+	k8sWorkspace := workspace.GetKubernetes()
+	var pvcSpec k8score.PersistentVolumeClaimSpec
+	// Only use the default PVC spec when the user requests a workspace.
+	// This allows us to populate missing defaults (such as storage class, access modes, etc.)
+	// if the user did not specify them in their workspace config.
+	if opts != nil && opts.DefaultWorkspace != nil {
+		pvcSpec = *opts.DefaultWorkspace.DeepCopy()
+	}
```
HumairAK, line 523: Currently it looks like the code doesn't require admins to configure workspaces to allow users to use it, I think that's fine, just want to make sure that's the intention here - if so an inline code comment indicating as such is probably warranted

backend/src/v2/compiler/argocompiler/argo.go
```diff
@@ -434,3 +461,69 @@ var dummyImages = map[string]bool{
...
+	if err := json.Unmarshal(jsonBytes, &pvcSpec); err != nil {
+		return nil, err
+	}
+	return &pvcSpec, nil
+}
+
+// getWorkspacePVC constructs a PersistentVolumeClaim for the workspace.
+// It uses the default PVC spec (from API server config), and applies any user-specified
+// overrides from the pipeline spec, including the requested storage size.
+func getWorkspacePVC(
```
HumairAK: we should add a couple of unit tests for this, maybe use cusror or gemini to help create some quickly

backend/src/apiserver/config/config.json
```diff
@@ -24,5 +24,14 @@
...
+  "Workspace": {
+    "size": "100Gi",
+    "VolumeClaimTemplateSpec": {
+      "accessModes": [
+        "ReadWriteOnce"
+      ],
+      "storageClassName": "standard-csi"
+    }
+  }
+}
```
HumairAK: add a newline

HumairAK:
@VaniHaripriya I created a subissue for the epic issue you created for the workspace feature, let's have all future issues for future work related to this and create them as "sub-issues" to https://github.com/kubeflow/pipelines/issues/12053

backend/src/v2/compiler/argocompiler/argo.go
HumairAK: Thanks for the change, let's note this also in the user docs when we do that as a follow up item

HumairAK:
/lgtm
/approve

great work!

PR #11997, opened 2025-06-19
Fix(backend): Move version info to version.py for editable installs.

### LINKED ISSUE #11892: [sdk] kfp library has no version attribute when installed in dev mode
The `kfp` library has no `__version__` attribute when installed via `pip install -e kfp/python` for development. This is a problem for local development, because some of the library functions depend on the `__version__` attribute. For example, the `_get_packages_to_install_command` function in [sdk/python/kfp/dsl/component_factory.py](sdk/python/kfp/dsl/component_factory.py) uses it.

### Environment

*  KFP version: NA

*  KFP SDK version: 2.13.0 

*  All dependencies version:

### Steps to reproduce

- Create a virtual env
```bash
python -m venv .venv

source .venv/bin/activate
```
- Install `kfp` for local development
```bash
# from pipelines repo root
pip install -e sdk/python
```
- Try to access the version attribute
```python
# in a python shell
>>> import kfp
>>> print(kfp.__version__)
```
This results in the following error:
```python
  File "<stdin>", line 1, in <module>
AttributeError: module 'kfp' has no attribute '__version__'
```

### Expected result

The command `kfp.__version__` definitely should not fail. Perhaps it should return an empty string?

### Materials and Reference

A workaround for now is to set the version manually to some value (i.e. `kfp.__version__='2.13.0'`), but this is not ideal.

---

therrshan:

This PR fixes the missing `kfp.__version__` attribute when installing the SDK using `pip install -e sdk/python`.

Changes:
- Move version string to a dedicated `version.py` file.
- Update `__init__.py` to import `__version__` from `version.py` with a fallback.
- Update `setup.py` to extract version from `version.py` instead of `__init__.py`.

This makes the version attribute consistently available, including during editable installs.

Fixes #11892 

**Checklist:**
- [ .] You have [signed off your commits](https://www.kubeflow.org/docs/about/contributing/#sign-off-your-commits)
- [ .] The title for your pull request (PR) should follow our title convention. [Learn more about the pull request title convention used in this repository](https://github.com/kubeflow/pipelines/blob/master/CONTRIBUTING.md#pull-request-title-convention).


sdk/python/kfp/__init__.py
HumairAK, line 1: can you also update the https://github.com/opendatahub-io/data-science-pipelines/blob/master/sdk/RELEASE.md and add this to bug fixes (under current version)

sdk/python/kfp/__init__.py
```diff
@@ -16,7 +16,10 @@
 # https://packaging.python.org/guides/packaging-namespace-packages/#pkgutil-style-namespace-packages
 __path__ = __import__('pkgutil').extend_path(__path__, __name__)
 
-__version__ = '2.13.0'
```
HumairAK, line 19: can you also update the release docs (step 2): https://github.com/kubeflow/pipelines/blob/8a402c10a8e3d36f964fb451760319c99a185e8f/RELEASE.md#releasing-kfp-python-packages

sdk/python/kfp/version.py
```diff
@@ -0,0 +1 @@
+__version__ = '2.13.0'
```
HumairAK:
```suggestion
__version__ = '2.13.0'

```

HumairAK (changes requested):

HumairAK:
thanks @therrshan appreciate the change, we'll probably want to do something similar for kubernetes_platform package if you are up for it, though not required for this pr, it can be a follow up as well 

feel free to ping me in slack for a re-review

sdk/python/kfp/__init__.py
therrshan: Wasn't sure if I should update the opendatahub repo since this is on kubeflow/pipelines, but I have done the change on the corresponding file in this repo

PR #12006, opened 2025-06-23
feat(sdk): Add Support for Docker Container Run Arguments

BieniekAlexander:
At the time of this writing, the KFP SDK local pipeline run functionality doesn't support authentication mechanisms, [as documented here](https://www.kubeflow.org/docs/components/pipelines/user-guides/core-functions/execute-kfp-pipelines-locally/). For cloud platforms such as GCP, authentication just involves [supplying authentication files to the runtime filesystem of the container](https://cloud.google.com/docs/authentication/application-default-credentials?hl=en#personal). 

Docker's [container running interface](https://docker-py.readthedocs.io/en/stable/containers.html#docker.models.containers.ContainerCollection.run) already supports everything necessary for the developer to propagate such credentials to the container runtime, so as far as the above is concerned, this PR seeks to collect such runtime configurations in the `DockerRunner` object and propagate them to the `containers.run` invocation.

In writing this, I realize that the implementation only supports supplying the same run configurations to all containers in the pipeline, which has limited flexibility. In comporting with the existing interface for local runs, I don't believe there's a simple means of propagating different run configurations for different containers, but I believe that this proposed update to the existing interface does provide for a lot.

**Checklist:**

**Testing:**
```py
from kfp import local
from kfp import dsl
from docker.types import Mount

local.init(
    runner=local.DockerRunner(
        mounts=[Mount(source="/home/usr/credentials", target="/credentials")],
        environment={"GOOGLE_APPLICATION_CREDENTIALS": "/credentials/google-creds.json"}
    )
)

@dsl.component
def print_credentials():
    # obviously, this is just for proving the functionality
    with open('/credentials/google-creds.json', 'r') as f:
        print(f.read())

# or run it in a pipeline
@dsl.pipeline
def pipeline():
    print_credentials()

pipeline_task = pipeline()
```

### DIFF
--- a/sdk/python/kfp/local/config.py
+++ b/sdk/python/kfp/local/config.py
@@ -45,18 +45,143 @@ class SubprocessRunner:
     use_venv: bool = True
 
 
-@dataclasses.dataclass
 class DockerRunner:
     """Runner that indicates that local tasks should be run as a Docker
     container."""
+    DOCKER_CONTAINER_RUN_ARGS = {
+        # for available run parameters:
+        # https://docker-py.readthedocs.io/en/stable/containers.html
+        'image',
+        'command',
+        'auto_remove',
+        'blkio_weight_device',
+        'blkio_weight',
+        'cap_add',
+        'cap_drop',
+        'cgroup_parent',
+        'cgroupns',
+        'cpu_count',
+        'cpu_percent',
+        'cpu_period',
+        'cpu_quota',
+        'cpu_rt_period',
+        'cpu_rt_runtime',
+        'cpu_shares',
+        'cpuset_cpus',
+        'cpuset_mems',
+        'detach',
+        'device_cgroup_rules',
+        'device_read_bps',
+        'device_read_iops',
+        'device_write_bps',
+        'device_write_iops',
+        'devices',
+        'device_requests',
+        'dns',
+        'dns_opt',
+        'dns_search',
+        'domainname',
+        'entrypoint',
+        'environment',
+        'extra_hosts',
+        'group_add',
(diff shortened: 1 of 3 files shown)


sdk/python/kfp/local/config.py
```diff
@@ -45,19 +46,47 @@ class SubprocessRunner:
...
         try:
             import docker  # noqa
         except ImportError as e:
             raise ImportError(
                 f"Package 'docker' must be installed to use {DockerRunner.__name__!r}. Install it using 'pip install docker'."
             ) from e
 
+        unsupported_args = container_run_args.keys() - {
+                param
+                for config in [docker.types.HostConfig,docker.types.ContainerConfig]
```
HumairAK: it is very common for docker-py to have apy breaking changes on this internal implementation, so this is not always guaranteed to work depending on the docker version since docker.types is not always supported - my suggestion is to hardcode an allow list and check against that, since the args api is publicly supported (unlike docker.types) and less likely to break.

HumairAK:
appreciate the change @BieniekAlexander, the usecase makes sense to me - can you address the ci failures as well as the comment above, feel free to ping me in slack for a second look

sdk/python/kfp/local/config.py
BieniekAlexander: I'm concerned about hardcoded kwargs easily breaking due to changes to the `containers.run` interface, so I went with the above as a means of programattically pulling something. However, I'll take your recommendation here!

HumairAK:
/lgtm
/approve

Thank you @BieniekAlexander !

PR #12020, opened 2025-06-27
docs: Add a proposal for Model Registry integration

mprahl:

This document proposes integrating Kubeflow Pipelines (KFP) with the Kubeflow Model Registry to enable automated model registration during pipeline execution workflows. The integration would provide a simple model.register() API in the KFP SDK that abstracts away Model Registry connection details and authentication, while automatically populating KFP-specific metadata for proper model lineage tracking.

**Checklist:**

### DIFF
--- a/proposals/12020-model-registry-integration/README.md
+++ b/proposals/12020-model-registry-integration/README.md
@@ -1,0 +1,399 @@
+# KEP-12020: Model Registry Integration for Kubeflow Pipelines
+
+<!-- toc -->
+- [Summary](#summary)
+- [Motivation](#motivation)
+  - [Current State and Limitations](#current-state-and-limitations)
+  - [Integration Benefits](#integration-benefits)
+  - [Current Workarounds](#current-workarounds)
+  - [Goals](#goals)
+  - [Non-Goals](#non-goals)
+- [Proposal](#proposal)
+  - [SDK User Experience](#sdk-user-experience)
+    - [API Reference](#api-reference)
+    - [Backend Translation](#backend-translation)
+    - [Cross-Reference Metadata](#cross-reference-metadata)
+  - [Configuration Management](#configuration-management)
+  - [User Stories](#user-stories)
+    - [Story 1](#story-1)
+    - [Story 2](#story-2)
+  - [Risks and Mitigations](#risks-and-mitigations)
+- [Design Details](#design-details)
+  - [Architecture Overview](#architecture-overview)
+  - [Implementation Notes](#implementation-notes)
+    - [Metadata Handling](#metadata-handling)
+    - [Launcher Processing](#launcher-processing)
+  - [API Design](#api-design)
+    - [Model Registry Request Structure](#model-registry-request-structure)
+  - [Security Considerations](#security-considerations)
+- [Test Plan](#test-plan)
+  - [Unit Tests](#unit-tests)
+    - [Configuration Tests](#configuration-tests)
+    - [SDK Tests](#sdk-tests)
+  - [Integration Tests](#integration-tests)
+    - [Successful Scenarios](#successful-scenarios)
+    - [Error Scenarios](#error-scenarios)
+  - [Graduation Criteria](#graduation-criteria)
+- [Implementation History](#implementation-history)
+- [Drawbacks](#drawbacks)
+- [Alternatives](#alternatives)
+  - [Alternative 1: Direct SDK Integration](#alternative-1-direct-sdk-integration)
+<!-- /toc -->
+
+## Summary
+
+The [Kubeflow Model Registry](https://www.kubeflow.org/docs/components/model-registry/) serves as the centralized
+metadata store for the Kubeflow ecosystem, providing comprehensive model cataloging, versioning, and discovery
+capabilities. This proposal introduces seamless integration between Kubeflow Pipelines (KFP) and Model Registry,
+enabling automated model registration as part of pipeline execution workflows.
+
+The integration abstracts away connection details, authentication mechanisms, and Model Registry-specific APIs,
+providing data scientists with a simple interface for model registration through the KFP SDK. This enhancement bridges
+the gap between KFP's artifact storage and Model Registry's cataloging capabilities, creating a unified model management
+experience.
+
+## Motivation
+
+### Current State and Limitations
+
+Kubeflow Pipelines currently maintains its own artifact store for pipeline outputs, which provides basic storage
+functionality but lacks advanced cataloging and versioning features. Models can only be added to this store through:
+
+1. **Pipeline execution outputs** (e.g., `dsl.Output[dsl.Model]`)
+2. **Importer components** for external model registration
+
+This approach has several limitations:
+
+- **No centralized cataloging**: Models are scattered across pipeline runs without unified discovery
+- **Limited versioning**: No structured version management or lineage tracking
+- **Tool-specific isolation**: Models created in KFP are not discoverable by other Kubeflow components
+- **Manual registration overhead**: Users must manually register models outside of pipeline workflows
+
+### Integration Benefits
+
+By enabling seamless Model Registry integration, users will experience:
+
+- **Centralized model catalog**: Single source of truth for all models across the Kubeflow ecosystem
+- **Versioning**: Structured version management with lineage tracking
+- **Cross-tool discovery**: Models become discoverable by other Kubeflow components (e.g. KServe)
+- **Simplified workflows**: One-line model registration within pipeline components
+- **Enhanced governance**: Better model lifecycle management and compliance tracking
+
+### Current Workarounds
+
+Without this integration, users must implement complex workarounds:
+
+1. Add Model Registry infrastructure details in the pipeline (e.g. URLs) through hardcoded values, pipeline input
+   parameters, or mounting a Kubernetes `ConfigMap`/`Secret`.
+1. Mount a Kubernetes `Secret` to access the token.
+1. Know the KFP standards for registering models in Model Registry (e.g. `model_source_kind="kfp"`).
+
+### Goals
+
+1. **Seamless SDK Integration**: Provide a simple, API for model registration within KFP components
+2. **Infrastructure Abstraction**: Hide Model Registry connection details and authentication from user code
+3. **Standardized Metadata**: Automatically populate KFP-specific metadata for proper model lineage tracking
+4. **Error Resilience**: Provide configurable error handling to prevent pipeline failures due to registration issues
+5. **Multi-tenancy Support**: Enable namespace-specific configuration for isolated deployments
+
+### Non-Goals
+
+1. Implement KFP-specific RBAC controls over Model Registry APIs, such as model allowlists for version registration.
+
+## Proposal
+
+### SDK User Experience
+
+The proposed integration introduces a `register()` method on KFP Model artifacts, providing a clean interface for model
+registration:
+
+```python
+@dsl.component()
+def train_model(
+    model: dsl.Output[dsl.Model],
+):
+    # Training logic
+    with open(model.path, "r") as model_file:
+        print("Training the model...")
+
+    # Set model metadata
+    model.name = "my-model-v1.0.0"
+    model.metadata["training_epochs"] = 100
+    model.metadata["accuracy"] = 0.95
+
+    # Register model with Model Registry
+    model.register(
+        model_name="sentiment-classifier",
+        description="BERT-based sentiment classification model",
+        model_format_name="vLLM",
+        model_format_version=None,
+        owner="ml-team",
+        author="data-scientist@company.com",
+        continue_on_error=True,  # Default: True
+    )
+```
+
+#### API Reference
+
+The `model.register()` method accepts the following parameters:
+
+| Parameter              | Type | Required | Default              | Description                                          |
+| ---------------------- | ---- | -------- | -------------------- | ---------------------------------------------------- |
+| `model_name`           | str  | Yes      | -                    | Name of the model in Model Registry                  |
+| `description`          | str  | No       | ""                   | Human-readable model description                     |
+| `model_format_name`    | str  | No       | None                 | Model format (e.g., "PyTorch", "TensorFlow", "ONNX") |
+| `model_format_version` | str  | No       | None                 | Version of the model format                          |
+| `owner`                | str  | No       | "Kubeflow Pipelines" | Model owner/team                                     |
+| `author`               | str  | No       | "Kubeflow Pipelines" | Model author                                         |
+| `continue_on_error`    | bool | No       | True                 | Whether to fail pipeline on registration error       |
+
+#### Backend Translation
+
+The SDK call translates to the following Model Registry API invocation:
+
+```python
+# Equivalent Model Registry client call
+registered_model = registry.register_model(
+    name="sentiment-classifier",
+    description="BERT-based sentiment classification model",
+    owner="ml-team",
+    author="data-scientist@company.com",
+    version="my-model-v1.0.0",  # model.name
+    uri="s3://kfp-artifacts/run-123/model",  # model.uri
+    metadata={
+        "training_epochs": 100,
+        "accuracy": 0.95
+    },  # model.metadata
+    model_format_name="vLLM",
+    model_format_version=None,
+    model_source_id="b6a9dde3-1647-463f-aeb8-5800089c84e8",  # Pipeline run ID
+    model_source_name="sentiment-training-pipeline",  # Pipeline run name
+    model_source_class="pipelinerun",  # KFP-specific identifier
+    model_source_kind="kfp",  # KFP-specific identifier
+    model_source_group="ml-team",  # Pipeline namespace
+)
+```
+
+#### Cross-Reference Metadata
+
+After successful registration, KFP adds metadata to the model artifact for UI cross-referencing. This is a list/array
+since multiple pipeline runs could register the same model.
+
+```python
+model.metadata["registered_models"] = [{
+    "modelName": "sentiment-classifier",
+    "versionName": "my-model-v1.0.0",
+    "versionID": 42,
+    "modelID": 15,
+    "modelRegistryURL": "https://model-registry.example.com:8443/models/15/versions/42",
+}]
+```
+
+The KFP UI should prominently display the model versions and link to the Model Registry UI when viewing the model's
+details.
+
+### Configuration Management
+
+Extend the existing `kfp-launcher` ConfigMap to include Model Registry configuration:
+
+```yaml
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  name: kfp-launcher
+  namespace: kubeflow
+data:
+  # Existing configuration...
+  pipelineRoot: "s3://kfp-artifacts"
+
+  # New Model Registry configuration
+  modelRegistry: |
+    url: https://model-registry.example.com:8443
+    tokenSecretRef: # Follows the same key names for existing configurations
+      secretName: model-registry-auth
+      secretNamespace: model-registry-system  # Defaults to current namespace
+      tokenKey: token
+    caConfigMapRef:  # Optional TLS certificate
+      configMapName: model-registry-ca-bundle
+      configMapNamespace: model-registry-system
+      key: ca-bundle.crt
+    timeout: 30s  # HTTP timeout for registration requests
+    retryAttempts: 3  # Number of retry attempts on failure
+```
+
+### User Stories
+
+#### Story 1
+
+As a data scientist, I would like to register my model to Model Registry in a pipeline without knowing the underlying
+Model Registry infrastructure and APIs, so that I can easily track model versions, share models with my team, and
+maintain a centralized model catalog as part of my automated ML workflows.
+
+#### Story 2
+
+As a data scientist, I would like to access a centralized model catalog that is independent of the specific tool used to
+create the models, enabling easy discovery and simplified model management across different ML workflows.
+
+### Risks and Mitigations
+
+The Model Registry API doesn't have granular RBAC so the `pipeline-runner` service account has full access to the Model
+Registry API.
+
+## Design Details
+
+### Architecture Overview
+
+```
+┌─────────────────┐     ┌──────────────────┐     ┌─────────────────┐
+│   KFP Pipeline  │     │   KFP Launcher   │     │ Model Registry  │
+│                 │     │                  │     │                 │
+│ model.register()│───▶│ Extract metadata │───▶│   API Server    │
+│                 │     │ Register model   │     │                 │
+└─────────────────┘     └──────────────────┘     └─────────────────┘
+                                ▲
+                                │
+                       ┌───────────────────┐
+                       │    kfp-launcher   │
+                       │     ConfigMap     │
+                       │                   │
+                       └───────────────────┘
+```
+
+### Implementation Notes
+
+#### Metadata Handling
+
+The `model.register()` call sets a special metadata field that the launcher processes:
+
+```python
+# SDK sets this metadata field
+model.metadata["_kfp_model_registry_request"] = json.dumps({
+    "model_name": "sentiment-classifier",
+    "description": "BERT-based sentiment classification model",
+    "model_format_name": "vLLM",
+    "model_format_version": None,
+    "owner": "ml-team",
+    "author": "data-scientist@company.com",
+    "continue_on_error": True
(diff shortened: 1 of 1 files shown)


proposals/12020-model-registry-integration/README.md
```diff
@@ -0,0 +1,432 @@
...
+## Design Details
+
+### Architecture Overview
+
+```
+┌─────────────────┐     ┌──────────────────┐     ┌─────────────────┐
+│   KFP Pipeline  │     │   KFP Launcher   │     │ Model Registry  │
+│                 │     │                  │     │                 │
+│ model.register()│───▶│ Extract metadata │───▶│   API Server    │
+│                 │     │ Register model   │     │                 │
```
tarilabs, line 251:
```suggestion
│                 │     │                  │     │                 │
│ model.register()│───▶ │ Extract metadata │───▶ │   API Server    │
│                 │     │ Register model   │     │                 │
```

proposals/12020-model-registry-integration/README.md
```diff
@@ -0,0 +1,432 @@
...
+
+### Non-Goals
+
+1. Implement KFP-specific RBAC controls over Model Registry APIs, such as model allowlists for version registration.
+
+## Proposal
+
+### SDK User Experience
+
+The proposed integration introduces a `register()` method on KFP Model artifacts, providing a clean interface for model
```
tarilabs, line 107: understanding here, this `.register()` is available to the DSP author, so can also choose to selectively register it or not depending on the logic of the pipeline itself, sgtm

proposals/12020-model-registry-integration/README.md
```diff
@@ -0,0 +1,432 @@
...
+    model_source_name="sentiment-training-pipeline",  # Pipeline run name
+    model_source_class="pipelinerun",  # KFP-specific identifier
+    model_source_kind="kfp",  # KFP-specific identifier
+    model_source_group="ml-team",  # Pipeline namespace
+)
+```
+
+#### Cross-Reference Metadata
+
+After successful registration, KFP adds metadata to the model artifact for UI cross-referencing:
```
tarilabs: this will be very helpful to go DSP -> MR, in addition to the opposite direction thanks to the `model_source*` fields above.

proposals/12020-model-registry-integration/README.md
```diff
@@ -0,0 +1,432 @@
...
+N/A
+
+### Risks and Mitigations
+
+1. The Model Registry API doesn't have granular RBAC so the `pipeline-runner` service account has full access to the
+   Model Registry API.
+
+## Design Details
+
+### Architecture Overview
```
tarilabs, line 244: This I'm not sure I fully understands: if you set a model registry client based on the configuration in the context of the KFP code, you can use the Model Registry python client directly: https://model-registry.readthedocs.io/en/latest/#registering-models

Below, it sounds to me the `.register()` is actually manually mapping to a request handled on the Go side.
I'm confused why this translation is required/preferred, can you expand on it please?

tarilabs (commented):
thanks! some comments from my side and one question please

proposals/12020-model-registry-integration/README.md
mprahl: @tarilabs the nuance here is that the user doesn't have direct access to the Model Registry credentials/connection info and that they are essentially logging a request for the model to be registered and the backend logic handles the registration. This is to scope the interface to registering model versions associated with the pipeline step.

proposals/12020-model-registry-integration/README.md
mprahl: Yes, exactly. It's modeled after the InstructLab pipeline so with minor tweaks to the Model Registry dashboard, we should get really nice linking.

proposals/12020-model-registry-integration/README.md
mprahl: That's right! The user could choose to do the registration in a separate evaluation component from the one that generated the model.

proposals/12020-model-registry-integration/README.md
tarilabs: it's indeed KFP choice, but the option I spoke earlier would give User the ability to perform further actions than what is currently designed here, for example remove labels from previous ModelVersions of a RegisteredModel, etc.

proposals/12020-model-registry-integration/README.md
```diff
@@ -0,0 +1,432 @@
...
+    with open(model.path, "r") as model_file:
+        print("Training the model...")
+
+    # Set model metadata
+    model.name = "my-model-v1.0.0"
+    model.metadata["training_epochs"] = 100
+    model.metadata["accuracy"] = 0.95
+
+    # Register model with Model Registry
+    model.register(
```
franciscojavierarceo, line 125: A missing piece of metadata is features/inputs. In the traditional ML space, these are structured and governed in Feast quite well. In the context of an LLM, the representation is not quite as obvious as it's just generic test. But probably some optional key (maybe called `inputs` or `model_context`/`model_features`?) could be useful.

proposals/12020-model-registry-integration/README.md
franciscojavierarceo: Inevitably this should be consolidated by the Kubeflow SDK :)

proposals/12020-model-registry-integration/README.md
```diff
@@ -0,0 +1,432 @@
...
+| `owner`                | str  | No       | "Kubeflow Pipelines" | Model owner/team                                     |
+| `author`               | str  | No       | "Kubeflow Pipelines" | Model author                                         |
+| `continue_on_error`    | bool | No       | True                 | Whether to fail pipeline on registration error       |
+
+#### Backend Translation
+
+The SDK call translates to the following Model Registry API invocation:
+
+```python
+# Equivalent Model Registry client call
```
syntaxsdev, line 155: This all looks good. A few things I'm confused about:

1. We are trying to map this to direct MR client calls and not calls directly to the API correct? This adds in MR as a requirement to the KFP.

2. Model Registry also supports a [upload_artifact_and_register](https://model-registry.readthedocs.io/en/stable/#uploading-local-models-to-external-storage-and-registering-them)  method for doing an atomic store and MR registration. The store is either S3 or OCI.

Do we plan on utilizing that as well? Or are we counting on user to self-store or never store? Imo that breaks workflow.

mprahl: @syntaxsdev I was anticipating direct calls to the REST API for the implementation but I used the MR client as an example that may be easier to understand for the discussion.

mprahl: @syntaxsdev to your second point, Kubeflow Pipelines already handles uploading to S3 as part of it's normal flow of output artifacts, so this would be registering the models in Model Registry in addition to what it's already doing today. In other words, the URI associated with the model version is the S3 URI after Kubeflow Pipelines has uploaded the model.

proposals/12020-model-registry-integration/README.md
mprahl: @franciscojavierarceo that's a good point! I don't see any fields around that in Model Registry and so I think our only option is to add metadata fields. One thing we could consider is automatically adding the input parameters and values of the component as metadata in Model Registry. What do you think about it?

I think we need to think more broadly about adding out of the box cross referencing with Feast.

franciscojavierarceo: Yeah I think `inputs: dict[str, Any] = {}`  probably is the right language to refer to features / context. We could, of course, add all three even though that feels excessive. 

>I think we need to think more broadly about adding out of the box cross referencing with Feast.

I would personally love this, I think we should be Feast and our registry probably serves as a great starting point. I would want to make sure we incorporate enough primitives from GenAI to make sure it's good though, so we don't need to be 100% reliant on Feast.

proposals/12020-model-registry-integration/README.md
syntaxsdev: @mprahl In the future do we see substitution for the Kubeflow SDK? Especially if other endpoints are required or added later.

proposals/12020-model-registry-integration/README.md
```diff
@@ -0,0 +1,432 @@
...
+      tokenKey: token
+    caConfigMapRef:  # Optional TLS certificate
+      configMapName: model-registry-ca-bundle
+      configMapNamespace: model-registry-system
+      key: ca-bundle.crt
+    timeout: 30s  # HTTP timeout for registration requests
+    retryAttempts: 3  # Number of retry attempts on failure
+```
+
+#### Option 2: Introduce LauncherConfig CRD
```
HumairAK: I think this warrants it's own KEP, and my preference is to tackle this as part of a hollistic effor to revamp config management in KFP all together. In general my preference is to use a CR for configs, but I feel it is out of scope of this KEP to introduce this feature. So I lean option 1.

proposals/12020-model-registry-integration/README.md
```diff
@@ -0,0 +1,432 @@
...
+
+After successful registration, KFP adds metadata to the model artifact for UI cross-referencing:
+
+```python
+model.metadata["registered_models"] = [{
+    "modelName": "sentiment-classifier",
+    "versionName": "my-model-v1.0.0",
+    "versionID": 42,
+    "modelID": 15,
+    "modelRegistryURL": "https://model-registry.example.com:8443/models/15/versions/42",
```
HumairAK, line 188: consideration: we could also remove this and build this in code via `modelID` + `versionID` + the url from the config, and keep this metadata more portable

the problem I see with that is, then we would have to possibly make the UI aware of the MR host and specially handle this case

proposals/12020-model-registry-integration/README.md
```diff
@@ -0,0 +1,432 @@
...
+    "registrationTimestamp": "2024-01-15T10:30:00Z",
```
HumairAK: one consideration is when this is evaluated, as the actual _registration_ happens in the launcher, but the user code `model.register()` is executed during the runtime, so it may make sense to calculate the timestamp during the executor run and save it as an artifact metadata field on the model, rather than during launcher post-execution phase

proposals/12020-model-registry-integration/README.md
```diff
@@ -0,0 +1,432 @@
...
+      key: ca-bundle.crt
+    timeout: 30s
+    retryAttempts: 3
+```
+
+### User Stories
+
+#### Story 1
+
+As a data scientist, I would like to register my fine-tuned model to Model Registry in a pipeline without knowing the
```
HumairAK: does it have to be a fine-tuned model? iiuc model & modelversions have a similar relationship like pipelines and pipelineversions in KFP right? will we be able to register new models in this sense? or only register model versions on pre-existing models? if the former than we should maybe say "register my model, or fine-tuned model..."

proposals/12020-model-registry-integration/README.md
```diff
@@ -0,0 +1,432 @@
...
+kind: ConfigMap
+metadata:
+  name: kfp-launcher
+  namespace: kubeflow
+data:
+  # Existing configuration...
+  pipelineRoot: "s3://kfp-artifacts"
+
+  # New Model Registry configuration
+  modelRegistry: |
```
HumairAK, line 210: I'm going back and forth on this, but do we want the api to be such that we can easily support multiple MRs in the future and make this a list? wdyt? 


I think we also already talked about this and decided to support a single MR, but it doesn't sound too bad adding support for multiple: 

```python
    model.register(
        model_name="sentiment-classifier",
        description="BERT-based sentiment classification model",
        ....
        model_registry=some_name, # <------ specify registry here 
    )
```

and make this to a config name: 


```yaml 
apiVersion: v1
kind: ConfigMap
metadata:
  name: kfp-launcher
  namespace: kubeflow
data:

  pipelineRoot: "s3://kfp-artifacts"
  modelRegistries: 
      - name: some_config               # maps to model.register(model_registry=....,...)
        url: https://model-registry.example.com:8443
        ....
      - name: another_config
        url: https://model-registry.example.com:8443
        default: true
```

If you don't specify a config, it picks the default. We can also do this as a follow up as well if we don't want to support this as an mvp.

proposals/12020-model-registry-integration/README.md
```diff
@@ -0,0 +1,432 @@
...
+and maintain a centralized model catalog as part of my automated ML workflows.
+
+#### Story 2
+
+As a data scientist, I would like to access a centralized model catalog that is independent of the specific tool used to
+create the models, enabling easy discovery and simplified model management across different ML workflows.
+
+### Notes/Constraints/Caveats
+
+N/A
```
HumairAK:
```suggestion
```

proposals/12020-model-registry-integration/README.md
```diff
@@ -0,0 +1,432 @@
...
+
+### Risks and Mitigations
+
+1. The Model Registry API doesn't have granular RBAC so the `pipeline-runner` service account has full access to the
```
HumairAK:
```suggestion
The Model Registry API doesn't have granular RBAC so the `pipeline-runner` service account has full access to the
```

proposals/12020-model-registry-integration/README.md
HumairAK: @tarilabs the flow you are describing will always be open to the user, at the end of the day it's just python code that executes within a pipeline.

The question is: Are majority of the users looking for an easy one button push to register a model in majority of the cases (post training/fine-tuning)? 

We surmise the answer is yes, and this makes that process significantly easier. This way the user does not need to reach out to their admin for example, and ask for the URL and the Credentials/Access to the Model Registry they are writing to, we handle that all for them. 

If we find that the answer is no, and that in majority of the instances the user is expected to do a lot more than just register a model in a given run, then we should list these use cases out and discuss them further.

proposals/12020-model-registry-integration/README.md
```diff
@@ -0,0 +1,432 @@
...
+    ModelName         string            `json:"model_name"`
+    Description       string            `json:"description,omitempty"`
+    ModelFormatName   string            `json:"model_format_name,omitempty"`
+    ModelFormatVersion string           `json:"model_format_version,omitempty"`
+    Owner             string            `json:"owner,omitempty"`
+    Author            string            `json:"author,omitempty"`
+    ContinueOnError   bool              `json:"continue_on_error,omitempty"`
+}
+
+type ModelRegistryConfig struct {
```
HumairAK, line 325: we should include more tls options: 

* tlsEnabled (bool)
* tlsVerify  (bool)

proposals/12020-model-registry-integration/README.md
```diff
@@ -0,0 +1,432 @@
...
+                              │
+                              ▼
+                       ┌───────────────────┐
+                       │   Configuration   │
+                       │   (ConfigMap/     │
+                       │   LauncherConfig) │
+                       └───────────────────┘
+```
+
+### Implementation Notes
```
HumairAK, line 262: Are there any UI considerations? Maybe worth mentioning a note if so or not.

proposals/12020-model-registry-integration/README.md
```diff
@@ -0,0 +1,432 @@
...
+
+```
+┌─────────────────┐     ┌──────────────────┐     ┌─────────────────┐
+│   KFP Pipeline  │     │   KFP Launcher   │     │ Model Registry  │
+│                 │     │                  │     │                 │
+│ model.register()│───▶│ Extract metadata │───▶│   API Server    │
+│                 │     │ Register model   │     │                 │
+└─────────────────┘     └──────────────────┘     └─────────────────┘
+                              │
+                              ▼
```
HumairAK: I think the arrow should be reversed since we're reading in the configuration and not writing to

proposals/12020-model-registry-integration/README.md
mprahl: @syntaxsdev I wouldn't expect using the Kubeflow SDK in the future unless there was a compelling reason to. Mostly so that we don't have to call Python code from Go and we don't have to `pip install` the Kubeflow SDK in the user's container image.

proposals/12020-model-registry-integration/README.md
mprahl: I'm torn on this one. I think I'd rather keep the URL as is for convenience so the UI doesn't have to query the `kfp-launcher` ConfigMap per namespace and it allows the URL to still work if the namespace is later configured to a different Model Registry instance.

I could go either way. Do you have a preference?

proposals/12020-model-registry-integration/README.md
HumairAK: hey folks, I'm 100% with you on this being our plan, but I think this is better left as a follow up feature, I have set aside KFP and feast store integration as part of the KFP roadmap, and I think this is better suited for that effort. For now we can advise the user to log the metadata themselves via documentation which we ship as part of the feature with examples.

I don't think we should go with the automatic registration of a kfp component's inputs to the model registration metadata, because it's not always the case that all the relevant inputs are scoped to just the component - it can very well be the case we may miss some inputs that were used in a more upstream task that executed prior. So I think there's also lineage considerations here within (and outside) the pipeline.

proposals/12020-model-registry-integration/README.md
mprahl: That seems to have been an errant auto complete from Cursor lol. I'll just remove it since the timestamp should be in the Model Registry.

proposals/12020-model-registry-integration/README.md
franciscojavierarceo: Yeah, I agree and I think that's very reasonable. It's good for us to be explicit about it and have a consideratino for it but we don't need to solve for it now.

proposals/12020-model-registry-integration/README.md
mprahl: Fair point. That was too specific.

proposals/12020-model-registry-integration/README.md
mprahl: I think `tlsEnabled` is implied based on the protocol set (http or https) in the `URL` field. I'll add `tlsVerify` though.

proposals/12020-model-registry-integration/README.md
mprahl: Added to the Cross-Reference Metadata section.

proposals/12020-model-registry-integration/README.md
mprahl: Will defer on this for now and wait until there's a need for multiple model registries.

proposals/12020-model-registry-integration/README.md
HumairAK: yeah I think let's stick with the url for now to keep the UI decoupled

mprahl:
@syntaxsdev and/or @tarilabs could you please take another look?

proposals/12020-model-registry-integration/README.md
tarilabs:
```suggestion
│                 │     │                  │     │                 │
│ model.register()│───▶ │ Extract metadata │───▶ │   API Server    │
│                 │     │ Register model   │     │                 │
```

still not applied :)

proposals/12020-model-registry-integration/README.md
tarilabs: > This way the user does not need to reach out to their admin for example, and ask for the URL and the Credentials/Access to the Model Registry they are writing to, we handle that all for them.

You could do similarly for having MR auth in every pipelinerun injected, by setting up some configuration with analogous mechanism by the Admin, and without having "user to reach out to their admin" :)

But regardless I see, you want to have an "_opinionated_ for simple Model tracking to MR" without exposing all the capabilities for now.
/lgtm

proposals/12020-model-registry-integration/README.md
```diff
@@ -0,0 +1,399 @@
...
+1. **Configuration Overhead**: Requires per-namespace configuration, though this enables proper multi-tenancy
+2. **Dependency on Model Registry**: Creates dependency on external Model Registry service availability
+3. **API Coupling**: Tight coupling to Model Registry API version and structure
+
+## Alternatives
+
+### Alternative 1: Direct SDK Integration
+
+Instead of launcher-based registration, implement direct Model Registry client integration in the SDK. This would expose
+infrastructure details to user code and require authentication handling in components.
```
tarilabs, line 399: as comment above, this isn't necessarily the case; you can use analogous injection mechanism for the authentication, so that the user can simply focus on using the MR py client.

proposals/12020-model-registry-integration/README.md
tarilabs: > That's right! The user could choose to do the registration in a separate evaluation component from the one that generated the model.

thanks for the confirmation!

tarilabs (commented):
per comments below; I think this is a very valid integration, with the caveat that:

1. it doesn't expose all capabilities of the MR py client, so eventually we will see if KFP user demands them
2. analogousl admin-config, auth-inject mechanism, can be designed to surface the MR py client into the pipeline (py) code, so that the user can just use the client and without having to manually authenticate.

rareddy:
Should this integration be thought about at the Kubeflow SDK level instead of the Model Registry?

proposals/12020-model-registry-integration/README.md
mprahl: @tarilabs it renders nicely in the GitHub UI for me without this change. Maybe the arrow is displaying funny on your end?
Screenshot From 2025-07-14 15-13-19

proposals/12020-model-registry-integration/README.md
mprahl: @tarilabs we want to avoid injecting the credentials for security reasons to avoid leaking the credential to any pipeline. There are still ways for the pipeline to get it with this proposal since the Launcher runs with the same Kubernetes service account as the user's Python code, but it's something I want to fix later on. By going with an approach of injecting the credentials, it makes it harder to to close this gap.

We also don't want the user to being modifying other things in the Model Registry unrelated to the pipeline unless they provide their own credentials.

These are all things we could revisit later.

mprahl:
> Should this integration be thought about at the Kubeflow SDK level instead of the Model Registry?

@rareddy could you please elaborate a bit more? The proposal is to enhance the existing KFP SDK since it already deals with output model artifacts.

rareddy:
>@rareddy could you please elaborate a bit more? The proposal is to enhance the existing KFP SDK since it already deals with output model artifacts.

@mprahl I understand there will be a Kubeflow SDK, which will wrap the Model Registry SDK. I was trying to see if the integration makes sense using the Kubeflow SDK vs directly using the Model Registry SDK. cc: @syntaxsdev

mprahl:
> > @rareddy could you please elaborate a bit more? The proposal is to enhance the existing KFP SDK since it already deals with output model artifacts.
> 
> @mprahl I understand there will be a Kubeflow SDK, which will wrap the Model Registry SDK. I was trying to see if the integration makes sense using the Kubeflow SDK vs directly using the Model Registry SDK. cc: @syntaxsdev

@rareddy we aren't planning on using the Model Registry SDK directly as of now. We will likely just use the REST API. See the discussion about this here:
https://github.com/kubeflow/pipelines/pull/12020#discussion_r2177508328

proposals/12020-model-registry-integration/README.md
syntaxsdev: @mprahl For sure, was originally under the impression that some of this would have been kicked off in the python library.

syntaxsdev:
@mprahl I know this proposal is strictly for model registration, but there is still a storage piece which I assume is going to use existing solutions?

I'd just want to make sure it's known that the MR client does have direct integration with registering + storing artifacts (both to S3/OCI) which also directly integrates with the OCI connection and S3 connections attached to workbench notebooks, which in that case might make sense to streamline if necessary, (there are definitely tradeoffs though) since we also offer an atomic failure solution.

syntaxsdev:
Other than that, looks good. Some table formatting.
/lgtm

mprahl:
> @mprahl I know this proposal is strictly for model registration, but there is still a storage piece which I assume is going to use existing solutions?
> 
> I'd just want to make sure it's known that the MR client does have direct integration with registering + storing artifacts (both to S3/OCI) which also directly integrates with the OCI connection and S3 connections attached to workbench notebooks, which in that case might make sense to streamline if necessary, (there are definitely tradeoffs though) since we also offer an atomic failure solution.

For now, we want to leverage the existing S3 configuration/integration in KFP but it's something we can revisit when KFP and Model Registry are more integrated.

Thanks for the review!

PR #12023, opened 2025-06-29
Add a proposal for the standalone driver implementation based on Agro…

ntny:
… Workflow backend

**Checklist:**


ntny:
Added more detailed information about the implementation)

droctothorpe:
This is a fantastic proposal and POC @ntny. Kudos!

What do you think the LOE would be to provide a **global** driver server instead of leveraging executor plugins to have one per workflow? Maybe it could be a standalone server that we deploy in the `kubeflow` namespace _independent of AWF_.

ntny:
> This is a fantastic proposal and POC @ntny. Kudos!
> 
> What do you think the LOE would be to provide a **global** driver server instead of leveraging executor plugins to have one per workflow? Maybe it could be a standalone server that we deploy in the `kubeflow` namespace _independent of AWF_.

Hi, thanks!
Makes sense.

1. Regarding the global KFP driver server: it might not be entirely clear from the[ diagram](https://github.com/kubeflow/pipelines/pull/12023/files#diff), but the KFP driver server is intended to be a single instance for the entire multi-user cluster. It runs in the kubeflow namespace alongside the ML Pipelines API server, the Argo Workflow Controller, and other core components. I’ll make sure to clarify this in the documentation.

2. In contrast, the agent pod is launched once per pipeline (Argo workflow). It acts as a proxy to the KFP driver server, and this is a limitation of how Argo Workflows currently work with the agent pod.

As for the LOE — this needs further investigation. The underlying issue is tracked [here](https://github.com/argoproj/argo-workflows/issues/7891)

proposals/separate-standalone-driver/README.md
droctothorpe, line 1: These appear as one line in the rendered markdown. Truly a nit, but do you mind making sure it renders more clearly? Thanks!

proposals/separate-standalone-driver/README.md
```diff
@@ -0,0 +1,190 @@
...
+
+Execution Order:
+1. entrypoint
+   Type: DAG
+The root DAG represents the entire pipeline run.
+2. root-driver (template: system-dag-driver)
+   Type: Container
+   Purpose: Initializes the root DAG. Creates an MLMD execution for the DAG.
+Outputs:
+execution-id – ID of the DAG execution (created during root-driver execution)
```
droctothorpe, line 36: These line breaks are visible in the raw markdown but appear to be one line in the rendered markdown. Truly a nit, but mind making sure that they render properly?

proposals/separate-standalone-driver/README.md
```diff
@@ -0,0 +1,190 @@
...
+        parameters:
+          - name: pod-spec-patch
+            jsonPath: $.pod_spec_patch
+          - name: cached-decision
+            jsonPath: $.cached_decision
+```
+### Requirements:
+- Execute a remote call with parameters
+- Read the response
+- Extract parameters from the response
```
droctothorpe, line 120: What component handles the parameter extraction and podspecpatch creation and application? The AWF controller?

proposals/separate-standalone-driver/README.md
```diff
@@ -0,0 +1,190 @@
...
+- [Executor Plugin](https://argo-workflows.readthedocs.io/en/latest/executor_plugins/)
+
+Comparison:
+
+| Feature          | Supports Remote Call | Read the Response | Can Extract Parameters | Notes                        |
+|------------------|----------------------|-------------------|------------------------|------------------------------|
+| HTTP Template    | ✅                   | ✅                | ❌                     |                              |
+| Executor Plugin  | ✅                   | ✅                | ✅                     | Requires plugin installation |
+
+The HTTP template [is not able](https://github.com/argoproj/argo-workflows/issues/13955) to extract parameters from the response and can only use the full response as-is. As a result, it cannot be used in podSpecPatch: '{{inputs.parameters.pod-spec-patch}}' or when: '{{inputs.parameters.cached-decision}} != true'
```
droctothorpe, line 137: Do you mind elaborating on why this is a blocker? What handles the parameter extraction? Why is it necessary?

proposals/separate-standalone-driver/README.md
```diff
@@ -0,0 +1,190 @@
...
+
+There’s a trade-off between running a standalone driver service pod globally or single per workflow. This is a balance between better performance and avoiding a single point of failure.
+Currently, Argo [supports](https://github.com/argoproj/argo-workflows/issues/7891) only one driver pod per workflow option. Both options are based on the Agent pod, which is currently started per workflow — this is a limitation of the current implementation.
+
+### Implementation Based on the Executor Plugin
```
droctothorpe, line 142: I want to make sure I'm picking up what you're putting down.

So there's one agent + server per workflow and all they basically do is make HTTP requests to the centralized driver server, right?

proposals/separate-standalone-driver/README.md
ntny: Sure! 
1. System-container-driver implementations initially try to [find](https://github.com/kubeflow/pipelines/blob/master/backend/src/v2/driver/cache.go#L59) a cache for the execution. If a cache exists, a pod for the execution will not be created and the cache will be reused.
Whether the pod is created or skipped is controlled via a `when` [condition](https://github.com/kubeflow/pipelines/blob/master/backend/src/v2/compiler/argocompiler/testdata/hello_world.yaml#L110) condition in the workflow spec, and this logic is handled by the workflow-controller
A boolean parameter must be explicitly passed to the `when` expression to control this behavior.
2. The same is true for the[pod-psec-path](https://github.com/kubeflow/pipelines/blob/master/backend/src/v2/compiler/argocompiler/testdata/hello_world.yaml#L176) it uses for the override image/cmd of the new pod and should be passed also explicitly.
 
Both parameters are part of the driver response and must be extracted in order to be used in the `when` condition and `pod-spec-patch`.
There might be a workaround, but I haven’t found a better option than using the Argo executor.

proposals/separate-standalone-driver/README.md
ntny: Exactly.
There is one agent pod per workflow in the user namespace. It has a sidecar that hosts a proxy server which makes requests to the centralized driver.

proposals/separate-standalone-driver/README.md
ntny: Yep

ntny:
> This is a fantastic proposal and POC @ntny. Kudos!
> 
> What do you think the LOE would be to provide a **global** driver server instead of leveraging executor plugins to have one per workflow? Maybe it could be a standalone server that we deploy in the `kubeflow` namespace _independent of AWF_.

I'm gonna take ownership of this community request and get back to you later with more details about the LOE.

proposals/separate-standalone-driver/src/driver-plugin/Dockerfile
HumairAK, line 1: let's scope this pr to just the proposal and not any code, can you remove `src/driver-plugin` ?

proposals/separate-standalone-driver/executor-plugin-flow.png
HumairAK, line 1: the diagram is making it seem like there are multiple tasksets CRs being updated, but iirc it's just one per workflow right? and these persist along with the workflow resource, after the workflow completes? if so we should note this behavior

proposals/separate-standalone-driver/README.md
```diff
@@ -0,0 +1,190 @@
...
+            dag_execution_id: '{{inputs.parameters.parent-dag-id}}'
+            iteration_index: '{{inputs.parameters.iteration-index}}'
+            kubernetes_config: '{{inputs.parameters.kubernetes-config}}'
+            pipeline_name: namespace/n1/pipeline/hello-world
+            pod_spec_patch_path: '{{outputs.parameters.pod-spec-patch.path}}'
+            run_id: '{{workflow.uid}}'
+            task: '{{inputs.parameters.task}}'
+            type: CONTAINER
+      outputs:
+        parameters:
```
HumairAK, line 190: there's also `iteration-count` and `execution-id`

proposals/separate-standalone-driver/README.md
```diff
@@ -0,0 +1,190 @@
...
+              jsonPath: $.cached-decision
+          - name: condition
+            valueFrom:
+              default: "true"
+              jsonPath: $.condition
+```
+
+## Conclusion
+This proposal introduces an optimization for Kubeflow Pipelines (KFP) that replaces per-task driver pods with a lightweight standalone service based on Argo Workflows’ Executor Plugin mechanism. It significantly reduces pipeline task startup time by eliminating the overhead of scheduling a separate driver pod for each task — particularly beneficial for large pipelines with multiple steps and caching enabled.
+Instead of launching a new driver pod per task, the driver logic is offloaded to a shared sidecar container (agent pod) within the workflow. This reduces latency in cache lookups and metadata initialization.
```
HumairAK:
```suggestion
Instead of launching a new driver pod per task, the driver logic is offloaded to a shared agent pod that is scheduled per workflow, and completes once the workflow ends. This reduces latency in cache lookups and metadata initialization.
```

let's be explicit in noting this is a separate pod that runs per workflow

proposals/separate-standalone-driver/executor-plugin-flow.png
HumairAK, line 1: why do we need a separate kfp-driver server? the agent pod can just act like an ephemeral service as per the executor example [here](https://argo-workflows.readthedocs.io/en/latest/executor_plugins/#example-a-simple-python-plugin) - instead of many drivers per workflow, we do a driver agent pod per workflow

I have a sample poc implementation this approach, here I make server mode optional, but if we went this route we would make this the only option and eliminate the cmd option in the driver: [[1]](https://github.com/HumairAK/data-science-pipelines/blob/external-driver-2/backend/src/v2/cmd/driver/main.go#L105C9-L105C18), [[2]](https://github.com/HumairAK/data-science-pipelines/blob/external-driver-2/backend/src/v2/driver/server.go#L234)

proposals/separate-standalone-driver/README.md
```diff
@@ -0,0 +1,190 @@
...
+            valueFrom:
+              default: "false"
+              jsonPath: $.cached-decision
+          - name: condition
+            valueFrom:
+              default: "true"
+              jsonPath: $.condition
+```
+
+## Conclusion
```
HumairAK, line 221: what are the authentication consideration, I see in argo's example here [they provide a token](https://argo-workflows.readthedocs.io/en/latest/executor_plugins/#example-a-simple-python-plugin)

can you speak a little bit here about how we plan to authenticate the service whether ephemeral or otherwise

HumairAK (changes requested):

proposals/separate-standalone-driver/README.md
HumairAK, line 221: can you also comment on testing strategies, or some sort of a test plan, cc @nsingla

proposals/separate-standalone-driver/executor-plugin-flow.png
ntny: Nice catch, fixed!

proposals/separate-standalone-driver/README.md
ntny: I might be wrong, but it [looks](https://github.com/kubeflow/pipelines/blob/master/backend/src/v2/compiler/argocompiler/testdata/hello_world.yaml#L86) like iteration-count and execution-id aren’t exists in the container-driver.outputs Argo-generated yaml . . execution-id is probably passed via command line when pod-sec-path is set up in the container-driver.

proposals/separate-standalone-driver/README.md
ntny: Could you please clarify — do you mean a list of specific test cases (which i do have), or more of a general test and release plan? I’m not entirely sure how testing is usually handled in Kubeflow, especially for PRs in general and breaking changes like this in particular. We should definitely think it through carefully — happy to collaborate on that. Do you maybe have any examples of places where a test plan has been written up before? That could really help as a starting point.

HumairAK: Essentially, we need to be test this implementation, and verify it works and doesn't break with future changes. So the Test plan section should elaborate on how we aim to accomplish that, unit tests, integration tests, what we want to be testing for etc. You can see the previous proposals [like this one](https://github.com/kubeflow/pipelines/tree/master/proposals/11875-pipeline-workspace#test-plan), for an example.

proposals/separate-standalone-driver/README.md
HumairAK: you can see all the parameters that are written out in driver [here](https://github.com/kubeflow/pipelines/blob/master/backend/src/v2/cmd/driver/main.go#L227-L235), these are written out to the filesystem but read in as argo parameters for downstream tasks, though I don't remember the exact details of where and when, you can inspect the workflow resource for that, [I think this is one example](https://github.com/kubeflow/pipelines/blob/master/backend/src/v2/compiler/argocompiler/testdata/hello_world.yaml#L286), for iteration count you may need to run a pipeline with a loop to find it in the workflow

proposals/separate-standalone-driver/executor-plugin-flow.png
ntny: @HumairAK  Due to the limited documentation available for the executor plugin, it’s challenging to understand how to properly mount the service account token into the agent sidecar container — or even if this is possible at all. 
This token is necessary for Kubernetes API access https://github.com/kubeflow/pipelines/blob/master/backend/src/v2/driver/k8s.go#L68

When I tried to run the driver inside the agent pod, i ran into the issue.  As a workaround, I have decided to extract the driver into a standalone server and use the agent just as a simple proxy between the workflow controller and the driver server.

proposals/separate-standalone-driver/README.md
ntny: There seems to be some confusion. The [example](https://github.com/kubeflow/pipelines/blob/master/backend/src/v2/compiler/argocompiler/testdata/hello_world.yaml#L286) with the execution-id and the iteration-count comes from the system-dag-driver.
I only provided a sample for the system-container-driver. I think there’s a mix-up, and I need to add a sample for the dag driver as well.
[ExecutionPaths](https://github.com/kubeflow/pipelines/blob/master/backend/src/v2/cmd/driver/main.go#L227-L235) is the output of both the DAG and container drivers; as a result, it contains fields from both container types.

proposals/separate-standalone-driver/executor-plugin-flow.png
ntny: Provided details about the Kubernetes access issue from the agent pod [here](https://github.com/kubeflow/pipelines/pull/12023/files#diff-3f4caa297ff7d1af7412226a6c628d4ec2bfbe515636cdf785857e316d2c63afR29).

proposals/separate-standalone-driver/README.md
ntny: Added a sample for the system-dag-driver next to system-container-driver

proposals/separate-standalone-driver/executor-plugin-flow.png
HumairAK: hey @ntny do you have stdout / or stacktrace of the issue you encountered? I'd be curious to know more, I did hit some issues on openshift as well and I'm curious to know if it's similar, especially if you were working on vanilla k8s. Feel free to reach out to me on slack, my inbox in github tends to be flooded, your bandwidth permitting I would love to see this kep move forward.

ntny: hey @HumairAK sure! Thanks a lot for your engagement!
Yep, it is vanila k8s
When I try to launch the driver as part of the agent pod, I get the following error during the request to the root DAG driver:
`failed to initialize kubernetes client: open /var/run/secrets/kubernetes.io/serviceaccount/token: no such file or directory; invalid configuration: no configuration has been provided, try setting KUBERNETES_MASTER environment variable
`
[source](https://github.com/kubeflow/pipelines/blob/master/backend/src/v2/driver/root_dag.go#L82)

I briefly explained the [reason](https://github.com/kubeflow/pipelines/pull/12023/files#diff-3f4caa297ff7d1af7412226a6c628d4ec2bfbe515636cdf785857e316d2c63afR29)
The executor pod doesn’t mount the service account volume into the sidecar container that hosts the driver.

Planning investigate it further, for now, I’m just grepping through related GitHub issues like this one: https://github.com/argoproj/argo-workflows/issues/13026

Instead of diving deep into the root cause, I’ve decided to implement a lightweight workaround: a proxy (agent pod hosts it as a sidecar container in the workflow namespace) between the controller and the driver (running in the Kubeflow namespace). This is helps to validate that the executor pod–based approach works overall.

ntny: Hi, @droctothorpe @HumairAK 
Update: Actually, we can embed the KFP driver directly inside the pod agent, avoiding the need for additional proxying. The Argo Workflow executor plugin lacks sufficient documentation, so it wasn't obvious at first that.
I have updated the proposal accordingly.
Details are here: https://github.com/kubeflow/pipelines/pull/12023/files#diff-3f4caa297ff7d1af7412226a6c628d4ec2bfbe515636cdf785857e316d2c63afR30
Also, I added the test plan.

proposals/separate-standalone-driver/README.md
```diff
@@ -0,0 +1,226 @@
...
+
+Execution Order:
+1. entrypoint
+   Type: DAG
+The root DAG represents the entire pipeline run.
+2. root-driver (template: system-dag-driver)
+   Type: Container
+   Purpose: Initializes the root DAG. Creates an MLMD execution for the DAG.
+Outputs:
+execution-id – ID of the DAG execution (created during root-driver execution)
```
HumairAK: nit: the newlines don't render the way here as you expect (try a rendered view of the markdown) 

the same applies to the points under "Tasks inside:"

proposals/separate-standalone-driver/README.md
```diff
@@ -0,0 +1,226 @@
...
+
+Tasks inside:
+- hello-world-driver (template: system-container-driver)
+  Purpose: Check for the existence of an execution in the cache, if no prepare the MLMD execution of the hello-world container task, and generate the appropriate pod-spec-patch.
```
HumairAK:
```suggestion
  Purpose: Check for the existence of an execution in the cache, if it does not exist, prepare the MLMD execution of the hello-world container task, and generate the appropriate pod-spec-patch.
```

proposals/separate-standalone-driver/README.md
```diff
@@ -0,0 +1,226 @@
...
+          - name: pod-spec-patch
+            valueFrom:
+              path: /tmp/outputs/pod-spec-patch
+          - name: cached-decision
+            valueFrom:
+              path: /tmp/outputs/cached-decision
+```
+It creates a pod that launches the driver container using the kfp-driver image.
+
+## Alternative
```
HumairAK:
```suggestion
## Proposal
```

proposals/separate-standalone-driver/README.md
```diff
@@ -0,0 +1,226 @@
...
+
+Instead of launching a new driver's pod using a container template, configure the system to send requests to an already running server.
+Something like this (showing both types of containers):
```
HumairAK:
```suggestion
Something like this (showing both types of drivers):
```
you meant drivers right?

proposals/separate-standalone-driver/README.md
```diff
@@ -0,0 +1,226 @@
...
+|------------------|----------------------|-------------------|------------------------|------------------------------|
+| HTTP Template    | ✅                   | ✅                | ❌                     |                              |
+| Executor Plugin  | ✅                   | ✅                | ✅                     | Requires plugin installation |
+
+The HTTP template [is not able](https://github.com/argoproj/argo-workflows/issues/13955) to extract parameters from the response and can only use the full response as-is. As a result, it cannot be used in podSpecPatch: '{{inputs.parameters.pod-spec-patch}}' or when: '{{inputs.parameters.cached-decision}} != true'
+
+There’s a trade-off between running a standalone driver service pod globally or single per workflow. This is a balance between better performance and avoiding a single point of failure.
+Currently, Argo [supports](https://github.com/argoproj/argo-workflows/issues/7891) only one driver pod per workflow option. Both options are based on the Agent pod, which is currently started per workflow — this is a limitation of the current implementation.
```
HumairAK: > this is a limitation of the current implementation.

I would add [this link](https://github.com/argoproj/argo-workflows/issues/7891) here as well

proposals/separate-standalone-driver/README.md
```diff
@@ -0,0 +1,226 @@
...
+
+### Implementation Based on the Executor Plugin
+
+Instead of creating a driver pod for each task, we can reuse a single agent pod via a plugin template:
+[Agent pod](https://github.com/argoproj/argo-workflows/issues/5544) is a unit designed for extension. 
+It can be extended by any server that implements the protocol.
+This server(plugin in Executor plugin terminology) runs as a sidecar alongside the agent pod.
+
+Below is a scheme where, instead of creating a pod for the driver's task, we reuse the Argo Workflow Agent via a plugin
+![img.png](executor-plugin-flow.png)
```
HumairAK, line 150: optional: 

to make it a bit clearer, I would color the following a separate color: 
* workflowtaskset CRD box
* wf-system-container-impl-1 pod 

each should be a different color imo, so we especially highlight that there no new pod for driver in this architecture

proposals/separate-standalone-driver/README.md
```diff
@@ -0,0 +1,226 @@
...
+Currently, Argo [supports](https://github.com/argoproj/argo-workflows/issues/7891) only one driver pod per workflow option. Both options are based on the Agent pod, which is currently started per workflow — this is a limitation of the current implementation.
+
+### Implementation Based on the Executor Plugin
+
+Instead of creating a driver pod for each task, we can reuse a single agent pod via a plugin template:
+[Agent pod](https://github.com/argoproj/argo-workflows/issues/5544) is a unit designed for extension. 
+It can be extended by any server that implements the protocol.
+This server(plugin in Executor plugin terminology) runs as a sidecar alongside the agent pod.
+
+Below is a scheme where, instead of creating a pod for the driver's task, we reuse the Argo Workflow Agent via a plugin
```
HumairAK, line 149: a note on what the workflowtaskset (can be brief) would help as it won't be immediately clear to everyone reading this why this exist/needed or how it is used by argo

proposals/separate-standalone-driver/README.md
```diff
@@ -0,0 +1,226 @@
...
+![img.png](executor-plugin-flow.png)
+
+
+To move from the container template to the Executor Plugin template:
+- patch the [Argo compiler](https://github.com/kubeflow/pipelines/tree/master/backend/src/v2/compiler/argocompiler) to generate a plugin template instead of a container template. Sample: hello-world [adapted](kfp-plugin-flow.png) (see name: system-container-driver)
```
HumairAK:
```suggestion
- Patch the [Argo compiler](https://github.com/kubeflow/pipelines/tree/master/backend/src/v2/compiler/argocompiler) to generate a plugin template instead of a container template. Sample: hello-world [adapted](kfp-plugin-flow.png) (see name: system-container-driver)
```

proposals/separate-standalone-driver/README.md
```diff
@@ -0,0 +1,226 @@
...
+- Namely, replace the templates used in the [container-driver](https://github.com/kubeflow/pipelines/blob/master/backend/src/v2/compiler/argocompiler/container.go#L148) and [dag-driver](https://github.com/kubeflow/pipelines/blob/master/backend/src/v2/compiler/argocompiler/dag.go#L156) section of the compiler
+- Extract the [driver](https://github.com/kubeflow/pipelines/tree/master/backend/src/v2/driver) component into a standalone server.
```
HumairAK:
```suggestion
- Convert the [driver](https://github.com/kubeflow/pipelines/tree/master/backend/src/v2/driver) component into a standalone server that will launch at the start of the workflow and persist throughout the duration of the workflow.
```

proposals/separate-standalone-driver/README.md
```diff
@@ -0,0 +1,226 @@
...
+          - name: component
+          - name: task
+          - name: container
+          - name: parent-dag-id
+          - default: "-1"
+            name: iteration-index
+          - default: ""
+            name: kubernetes-config
+      metadata: {}
+      plugin:
```
HumairAK, line 174:
```suggestion
      metadata: {}
      # this is the key change that specifies use of the executor plugin
      plugin:
```

proposals/separate-standalone-driver/README.md
HumairAK: I would also update the agend pod to explicitly note that this is driver **server** now

proposals/separate-standalone-driver/README.md
```diff
@@ -0,0 +1,226 @@
...
+Add an additional E2E test to verify the behavior of the global driver server.
+
+Additionally, it is nice to have end-to-end (E2E) tests to verify basic functionality. Existing tests should be reused if available. The E2E tests should cover at least the following scenarios:
+- A simple pipeline with a single component, waiting for successful completion of the run.
+- A pipeline with a chain of components passing inputs and outputs between them, waiting for successful completion of the run.
+- A pipeline designed to fail, waiting for the run to end with an error.
+- A pipeline which fails but has retries enabled(pipeline/ and component level), waiting for the run to complete successfully.
+
+## Conclusion
+This proposal introduces an optimization for Kubeflow Pipelines (KFP) that replaces per-task driver pods with a lightweight standalone service based on Argo Workflows’ Executor Plugin mechanism. It significantly reduces pipeline task startup time by eliminating the overhead of scheduling a separate driver pod for each task — particularly beneficial for large pipelines with multiple steps and caching enabled.
```
HumairAK, line 222: another note on caching, with this approach with caching enabled, cached steps now will truly result in no new pods being executed per task, where as before for a component that is cached we would still require a pod to be scheduled, with this approach, an entire workflow that has all components cached will at maximum result in 1 new pod (the agend pod)

proposals/separate-standalone-driver/README.md
```diff
@@ -0,0 +1,226 @@
...
+Integration tests
+Add an additional E2E test to verify the behavior of the global driver server.
+
+Additionally, it is nice to have end-to-end (E2E) tests to verify basic functionality. Existing tests should be reused if available. The E2E tests should cover at least the following scenarios:
+- A simple pipeline with a single component, waiting for successful completion of the run.
+- A pipeline with a chain of components passing inputs and outputs between them, waiting for successful completion of the run.
+- A pipeline designed to fail, waiting for the run to end with an error.
+- A pipeline which fails but has retries enabled(pipeline/ and component level), waiting for the run to complete successfully.
+
+## Conclusion
```
HumairAK, line 221: can you briefly comment on the speed gains if any for the driver step, from what I remember because there is some latency and round trip time due to communication with workflow controller, and the workflow task set reads, it isn't as instant as one might expect - but might still be preferable because it is capped by latency + CR updates/reads, where as previusly the start up time was capped by latency + pod scheduling + container start up time

proposals/separate-standalone-driver/README.md
```diff
@@ -0,0 +1,226 @@
...
+It can be extended by any server that implements the protocol.
+This server(plugin in Executor plugin terminology) runs as a sidecar alongside the agent pod.
+
+Below is a scheme where, instead of creating a pod for the driver's task, we reuse the Argo Workflow Agent via a plugin
+![img.png](executor-plugin-flow.png)
+
+
+To move from the container template to the Executor Plugin template:
+- patch the [Argo compiler](https://github.com/kubeflow/pipelines/tree/master/backend/src/v2/compiler/argocompiler) to generate a plugin template instead of a container template. Sample: hello-world [adapted](kfp-plugin-flow.png) (see name: system-container-driver)
+- Namely, replace the templates used in the [container-driver](https://github.com/kubeflow/pipelines/blob/master/backend/src/v2/compiler/argocompiler/container.go#L148) and [dag-driver](https://github.com/kubeflow/pipelines/blob/master/backend/src/v2/compiler/argocompiler/dag.go#L156) section of the compiler
```
HumairAK: to make this proposal make sense for the future reader, I suggest replacing all references to `master` value in URLs to a hard commit

proposals/separate-standalone-driver/README.md
```diff
@@ -0,0 +1,226 @@
...
+Instead of launching a new driver pod per task, the driver logic is offloaded to a shared agent pod that is scheduled per workflow, and completes once the workflow ends. This reduces latency in cache lookups and metadata initialization.
+However, this approach does not fully eliminate pod scheduling issues: the standalone driver is not a global service, but is instantiated per workflow. Thus, a pod still needs to be scheduled for each workflow run.
+
+## Disadvantages:
+A key limitation of this implementation is that it currently supports only the Argo Workflows backend. The Executor plugin also adds some extra complexity to maintenance and deployment.
+
+## Open Questions:
+- Do we need a fallback mechanism to the per-task driver pods in case the Executor Plugin is not available in some installations? Should KFP continue supporting both execution flows (plugin-based and pod-based drivers) for compatibility?
```
HumairAK, line 230: the executor plugin becomes an infra pod, my opinion here is a hard-no, we should just have a clear means of surfacing this infra level error to user facing UI 

since today we don't do this even for existing driver, I would say it is out of scope, but it would be a big gain for the user if we can do this at some point, 

for example today if driver pod fails to schedule, the user has no way of knowing why the pipeline is hanging

the same would be true if the driver agent service pod is unavailable for any reason 

we should note this here, and that it should be a good follow up to this proposal, but a non-goal

ntny: > for example today if driver pod fails to schedule, the user has no way of knowing why the pipeline is hanging

It's true that when a driver fails or there's an issue that prevents the executor pod from being scheduled, the UI completely loses observability.
However, the Argo Workflow's status contains all the necessary information about driver errors and most of the issues that break observability.
Regarding task-level  drivers (not root drivers), I'm considering upgrading the persistenceagent or creating a new component to track and handle such issues. It should be able to modify MLMD and update the pipeline's node status accordingly.
In the context of implementing a centralized driver for observability, I believe the best we can do is to always return the execution_id (if it was successfully created), even in case of a driver error. This would allow the system to link the failed driver with an execution.
Another challenge is how to surface such driver-level or scheduling-related errors in the UI in a clear and user-friendly way. This will likely require changes on the frontend side as well
I believe enhancing of observability is a separate topic, potentially for a much larger proposal.

Thanks, I’ve seen the other comments. I’ll address them soon.

HumairAK: Yes agreed, I would caution against the following: 

> Regarding task-level drivers (not root drivers), I'm considering upgrading the persistenceagent or creating a new component to track and handle such issues. It should be able to modify MLMD and update the pipeline's node status accordingly.

We intend to remove MLMD in the future, so future design decisions should aim to not leverage MLMD any further. We will likely want to surface this direction in the run object.

juliusvonkohout: With what do you want to replace mlmd ? Do you just want to hide it behind the API or switch it out completely for a multi-tenant compatible substitute?

HumairAK: I will be making a KEP proposal for this effort in a few days

proposals/separate-standalone-driver/README.md
ntny: Honestly, it feels a bit quicker.
I’ll think about how to properly measure it.

proposals/separate-standalone-driver/executor-plugin-flow.png
droctothorpe: Another pattern that may be worth considering is leveraging Kubeflow's PodDefault custom resource mechanism: https://github.com/kubeflow/kubeflow/blob/master/components/admission-webhook/README.md

It's already being used to automatically mount the KFP API token:
https://github.com/kubeflow/manifests/blob/8f1bbf66065a4e4fb9245be2676bf8674478fb55/tests/poddefaults.access-ml-pipeline.kubeflow-user-example-com.yaml#L1-L25

Not sure they're designed to target sidecars but might be worth looking into. Just a thought.

droctothorpe:
I plan to deploy this locally and validate it today. Will let you know if I run into any hiccups.

proposals/separate-standalone-driver/README.md
```diff
@@ -0,0 +1,233 @@
...
+## Motivation
+While using a separate pod for the executor makes sense - since it often handles heavy workloads and benefits from isolation and flexibility - the driver is a lightweight component. It typically performs just a few API calls: checking for cached results and creating an MLMD Execution.
+
+However, running the driver in a separate pod causes several issues:
+
+High overhead: Launching a Kubernetes pod merely to execute a few API calls introduces significant latency. Often, the pod scheduling and startup time outweighs the driver's actual processing time.
+
+Resource availability problems: There's no guarantee the Kubernetes cluster has sufficient resources to schedule the driver pod. If scheduling fails, the pipeline gets stuck. The UI currently doesn't show driver pod scheduling failures, which makes it hard to debug and understand what's going on.
+
+## Current state details
```
droctothorpe, line 24: Does documentation to this effect exist anywhere outside of this KEP? If not, it really should. We need something that clarifies and justifies this architecture beyond the out of date v2 design doc.

proposals/separate-standalone-driver/README.md
juliusvonkohout: Oh yes i like the security stuff :-D

proposals/separate-standalone-driver/executor-plugin-flow.png
juliusvonkohout: KFP should be able to work standalone, so i would not include poddefaults in the mix.

proposals/separate-standalone-driver/README.md
juliusvonkohout: And for what do you need authentication in the first place?

proposals/separate-standalone-driver/executor-plugin-flow.png
droctothorpe: Good point, @juliusvonkohout. Disregard the above, @ntny.

proposals/separate-standalone-driver/README.md
ntny: Based on my experience, I have indeed encountered this issue. From working with the implementation, I understand why it happens. The [HumairAK](https://github.com/HumairAK) [here](https://github.com/kubeflow/pipelines/pull/12023#discussion_r2248239153) is describing a similar problem, which is definitely still relevant.

Unfortunately, I haven’t found any information explaining the historical reasons behind the choice of architecture.

That said, I must admit I didn’t fully grasp this point about needing additional clarification and justification for the architecture beyond the out-of-date v2 design doc. I’d appreciate if you could kindly elaborate on what exactly is expected.

droctothorpe: I meant just fleshing out the great work you did in this "Current state details" section and making it available in the documentation (outside of this KEP). After writing this comment, I decided to make a mermaid diagram that clarifies this flow for end users / contributors, so no worries, I'll tackle this.

ntny: I recently made an internal presentation at my company about how things currently work and the benefits of moving to the new approach.
I still have some slides that might be useful. I can adapt them and publish something — at least covering the current implementation (with a driver per task) — or just share the slides with you directly.

If you think it’s worth doing, let me know what format would be best (e.g. an article somewhere or directly in the docs). Totally sure you’d make an even better version, but happy to share what I have.

droctothorpe: That's very thoughtful, thanks @ntny! Feel free to HMU on the CNCF slack. I'm @droctothorpe on there. I decided a markdown table probably made the most sense given how much information I want to put in some of the cells.

proposals/separate-standalone-driver/README.md
```diff
@@ -0,0 +1,233 @@
...
+[Agent pod](https://github.com/argoproj/argo-workflows/issues/5544) is a unit designed for extension. 
+It can be extended by any server that implements the protocol.
+This server(plugin in Executor plugin terminology) runs as a sidecar alongside the agent pod.
+
+Below is a scheme where, instead of creating a pod for the driver's task, we reuse the Argo Workflow Agent via a plugin
+![img.png](executor-plugin-flow.png)
+
+
+To move from the container template to the Executor Plugin template:
+- patch the [Argo compiler](https://github.com/kubeflow/pipelines/tree/a870b1a325dae0c82c8b6f57941468ee1aea960b/backend/src/v2/compiler/argocompiler) to generate a plugin template instead of a container template. Sample: hello-world [adapted](kfp-plugin-flow.png) (see name: system-container-driver)
```
mprahl: Could you fix the image link to kfp-plugin-flow.png?

proposals/separate-standalone-driver/README.md
```diff
@@ -0,0 +1,233 @@
...
+Two similar features in Argo Workflow can be considered to meet these requirements:
+- [Http Template](https://argo-workflows.readthedocs.io/en/latest/http-template)
+- [Executor Plugin](https://argo-workflows.readthedocs.io/en/latest/executor_plugins/)
+
+Comparison:
+
+| Feature          | Supports Remote Call | Read the Response | Can Extract Parameters | Notes                        |
+|------------------|----------------------|-------------------|------------------------|------------------------------|
+| HTTP Template    | ✅                   | ✅                | ❌                     |                              |
+| Executor Plugin  | ✅                   | ✅                | ✅                     | Requires plugin installation |
```
mprahl, line 135: @ntny could you touch on the authentication/authorization bits gating the plugin? Since the user running the pipeline can run any code with pipeline-runner service account, we need to make sure the side car doesn't accept requests from the user's code somehow.

ntny: @mprahl sure, makes sense, [added](https://github.com/ntny/pipelines/blob/standalone-driver-kep/proposals/separate-standalone-driver/plugin.md#securing-the-driver-sidecar-container) some details.
I'm not entirely sure I got your point about the pipeline-runner SA. The sidecar runs under a [separate](https://github.com/ntny/pipelines/blob/standalone-driver-kep/proposals/separate-standalone-driver/plugin.md#interaction-with-the-kubernetes-api-from-a-sidecar-container) `driver-plugin-executor-plugin` SA and it doesn’t use the pipeline-runner.

proposals/separate-standalone-driver/README.md
ntny: > what are the authentication consideration, I see in argo's example here [they provide a token](https://argo-workflows.readthedocs.io/en/latest/executor_plugins/#example-a-simple-python-plugin)
> 
> can you speak a little bit here about how we plan to authenticate the service whether ephemeral or otherwise

sure, [added](https://github.com/ntny/pipelines/blob/standalone-driver-kep/proposals/separate-standalone-driver/plugin.md#securing-the-driver-sidecar-container)

proposals/separate-standalone-driver/README.md
ntny: Hi @HumairAK 
It is difficult to get precise metrics directly, so I did some experiments
- I have launched pipelines consisting of 3 tasks. All of them were cached beforehand and used from cache.
- I logged the execution time of the driver plugin from the code.
The log shows 4 requests (1 from the root driver, the others from the container driver)

the output:
1st pipeline:
-  INFO:2025-09-02 11:59:26 - execute plugin... 
-  INFO:2025-09-02 11:59:36 - execute plugin...
-  INFO:2025-09-02 11:59:56 - execute plugin...
- INFO:2025-09-02 12:00:16 - execute plugin...
2nd pipeline:
- INFO:2025-09-02 12:10:53 - execute plugin... 
- INFO:2025-09-02 12:11:23 - execute plugin... 
- INFO:2025-09-02 12:11:44 - execute plugin... 
- INFO:2025-09-02 12:12:04 - execute plugin...

The delay of about 10 seconds(default option) is due to the agent’s requeueTime=10s, which is the interval at which the TaskSet reconciliation occurs.
it looks like the requeuetime is [customizable](https://argo-workflows.readthedocs.io/en/latest/environment-variables/) option 

```
time="2025-09-02T12:10:53.864Z" level=info msg="Starting Agent" requeueTime=10s taskWorkers=16 workflow=debug-component-pipeline-pbmzb

time="2025-09-02T12:10:53.877Z" level=info msg="TaskSet Event" event_type=ADDED workflow=debug-component-pipeline-pbmzb
time="2025-09-02T12:10:53.877Z" level=info msg="Processing task" nodeID=debug-component-pipeline-pbmzb-1733769900
time="2025-09-02T12:10:54.733Z" level=info msg="Sending result" message= nodeID=debug-component-pipeline-pbmzb-1733769900 phase=Succeeded requeue=0s

time="2025-09-02T12:11:03.868Z" level=info msg="Processing Patch" workflow=debug-component-pipeline-pbmzb
time="2025-09-02T12:11:03.879Z" level=info msg="Patched TaskSet"
time="2025-09-02T12:11:03.880Z" level=info msg="TaskSet Event" event_type=MODIFIED workflow=debug-component-pipeline-pbmzb
time="2025-09-02T12:11:03.880Z" level=info msg="Task is already considered" nodeID=debug-component-pipeline-pbmzb-1733769900
```

droctothorpe:
@ntny have you explored the possibility of adding driver endpoints and handlers to the central API instead of making it a standalone deployment?

ntny:
Here’s a working [POC](https://github.com/kubeflow/pipelines/pull/12269), including instructions for running it.

proposals/separate-standalone-driver/executor-plugin-flow.png
mprahl: @ntny could you please update proposals/separate-standalone-driver/plugin.md with that change?

droctothorpe:
/lgtm
/approve

Any remaining feedback can be addressed in the PR. Tysm @ntny!

Issue #9682, opened 2023-06-24
[feature] Add ability to specify node affinity & toleration using KFP V2


AlexandreBrown:
### Feature Area

/area sdk

### What feature would you like to see?
A core production use case of KFP involves being able to run CPU and GPU workload on specific nodegroups that are more powerful and different from the nodegroup where Kubeflow is installed and usually they will have autoscaling as well.
In order to achieve this, we used to be able to simply specify which component would run on which node using node affinity + tolerations. This is no longer possible in KFP v2 yet I feel like such a core feature should be supported.

### What is the use case or pain point?
The existing [set_accelerator_type](https://kubeflow-pipelines.readthedocs.io/en/sdk-2.0.1/source/dsl.html?h=add_#kfp.dsl.PipelineTask.set_accelerator_type.accelerator) is far from being flexible enough to allow such use case. Here is a small list of examples that shows that `set_accelerator_type` is not flexible enough to support production use cases :  
- Does not work if the GPU is not one of the few (3) supported GPU : `NVIDIA_TESLA_K80`, `TPU_V3` or `cloud-tpus.google.com/v3`. Otherwise we must use the generic `nvidia.com/gpu` which is not precise hence defeating the purpose of selecting an accelerator.
- If you have 2 nodegroups with the same GPU but one should be reserved for inference and one should be reserved for pipeline exectuion (eg: training) then there is no way to cover such distinction purely based on `set_accelerator_type('nvidia.com/gpu')`.
- this method is only meant to be used for GPU but it is common to want to run CPU workload on specific nodegroups, reasons could include nodegroup isolation (to run workload that wont affect the nodegroup where Kubeflow core pods run) or to allow for more powerful CPU nodegroups to be used for pipeline while Kubeflow would remain on cheaper instances. 

### Is there a workaround currently?  
Users can try to use external tools such as [Kyverno](https://kyverno.io/) to create mutating rules that a webhook can use to add a toleration and/or node affinity/node selector based on some predefined criteria such as a label name and value.  
It's still a pain since it makes it way more involved than being able to use `.add_node_affinity()` and `.add_toleration()` on a component. In fact, we can't even add a label using kfp sdk anymore so matching has to be done on labels that are present by chance (we have no control to explicitly ensure their presence).  
[...]

PR #12028, opened 2025-07-02
feat(sdk) Add Input Parameter support for node affinity

VaniHaripriya:

This change enables the specification of node affinity through input parameters, in alignment with the enhancement discussed in [#9682](https://github.com/kubeflow/pipelines/issues/9682).

**Testing Instructions**

**SDK**

- Create a Python virtualenv , generate the proto files locally and install the SDK.

```
$ python -m venv .venv
$ pushd api
$ make clean-python python
$ popd
$ source .venv/bin/activate
$ pip install --upgrade pip
$ pip install wheel setuptools protobuf grpcio grpcio-tools
$ pip install -r sdk/python/requirements-dev.txt
$ pip install -e api/v2alpha1/python
$ pip install -e sdk/python
$ pip install -e kubernetes_platform/python
```
Use the [example code](https://gist.github.com/VaniHaripriya/5d8ef65e3867abb7b709d8fe2ef1a71d) to compile
```
$ kfp dsl compile --py nodeaffinity.py --output nodeaffinity.yaml
```
You should be able to compile and find the following snippet in the main.yaml file:
```
---
platforms:
  kubernetes:
    deploymentSpec:
      executors:
        exec-print-hello-with-explicit-affinity:
          nodeAffinity:
          - matchFields:
            - key: metadata.name
              operator: In
              values:
              - node-1
              - node-2
            weight: 100.0
        exec-print-hello-with-json-affinity:
          nodeAffinity:
          - nodeAffinityJson:
              runtimeValue:
                constant:
                  preferredDuringSchedulingIgnoredDuringExecution:
                  - preference:
                      matchExpressions:
                      - key: disktype
                        operator: In
                        values:
                        - ssd
                    weight: 100.0
```
**Checklist:**


nsingla:
@VaniHaripriya Possible for you to provide an example of pipeline (sdk component) to verify this?

nsingla:
@VaniHaripriya Thanks for adding the test description, its very clear now as to how someone can add node affinity. However I am still not sure about the validation to see if this actually worked or not :thinking: 
Did you try to create a multi node cluster and verified if node affinity actually worked or not? or may within the single cluster, try to see if the only node does not get selected as it does not meet the criteria?

VaniHaripriya:
> @VaniHaripriya Thanks for adding the test description, its very clear now as to how someone can add node affinity. However I am still not sure about the validation to see if this actually worked or not 🤔 Did you try to create a multi node cluster and verified if node affinity actually worked or not? or may within the single cluster, try to see if the only node does not get selected as it does not meet the criteria?

@nsingla  I confirmed that the node affinity settings are correctly reflected in the generated pipeline YAML and pod specs during runtime. I used a multi-node cluster and set node affinity criteria that do not match the node labels — this caused the pods to remain unscheduled, which confirms that the affinity rules were enforced.

kubernetes_platform/python/kfp/kubernetes/node_affinity.py
```diff
@@ -0,0 +1,93 @@
...
+                key=expr.get("key"),
+                operator=expr.get("operator"),
+                values=expr.get("values", []),
+            )
+    if match_fields:
+        for field in match_fields:
+            affinity_term.match_fields.add(
+                key=field.get("key"),
+                operator=field.get("operator"),
+                values=field.get("values", []),
```
hbelmiro: We could add some validations here to catch errors early.

- Validate that `key` and `operator` are non-empty
- Validate that `operator` has a valid value like `In`, `NotIn`, `Exists`...
  - Valid values: https://kubernetes.io/docs/concepts/scheduling-eviction/assign-pod-node/#operators

Also, add tests for those validations.

kubernetes_platform/python/kfp/kubernetes/node_affinity.py
```diff
@@ -0,0 +1,93 @@
...
+            )
+    if weight is not None:
+        affinity_term.weight = weight
```
hbelmiro: We can validate that `weight` is  1-100.

kubernetes_platform/python/kfp/kubernetes/node_affinity.py
```diff
@@ -0,0 +1,93 @@
...
+    msg.node_affinity.append(affinity_term)
+    task.platform_config["kubernetes"] = json_format.MessageToDict(msg)
+    return task
+
+def add_node_affinity_json(
+    task: PipelineTask,
+    node_affinity_json: Union[pipeline_channel.PipelineParameterChannel, dict],
```
hbelmiro, line 82: We could validate `node_affinity_json` against the schema.
You probably can use something like the following (AI-generated):

```python
from kubernetes import client, config

def validate_node_affinity(node_affinity_json: dict):
    """
    Validates a JSON object against the V1NodeAffinity model.

    Args:
        node_affinity_json: A Python dictionary loaded from your JSON.

    Returns:
        True if the JSON is valid, False otherwise.
    """
    try:
        # Create an empty V1NodeAffinity object to merge into.
        v1_node_affinity = client.V1NodeAffinity()

        # The ApiMerger is used to deserialize Python dicts into Kubernetes models.
        merger = client.api_client.ApiMerger()
        merger.merge(v1_node_affinity, node_affinity_json)

        print("✅ JSON is a valid V1NodeAffinity.")
        return True
    except Exception as e:
        print(f"❌ Invalid V1NodeAffinity JSON: {e}")
        return False
```

And add a test for an invalid JSON.

hbelmiro (commented):
Looks good in general @VaniHaripriya.
I just left some suggestions to validate the input at compile time (SDK) for a better experience.

kubernetes_platform/python/kfp/kubernetes/node_affinity.py
```diff
@@ -0,0 +1,138 @@
...
+                raise ValueError("Each match_field must have a non-empty 'key'.")
+            if not operator:
+                raise ValueError(f"Each match_field for key '{key}' must have a non-empty 'operator'.")
+            if operator not in VALID_OPERATORS:
+                raise ValueError(f"Invalid operator '{operator}' for key '{key}'. Must be one of {sorted(VALID_OPERATORS)}.")
+            affinity_term.match_fields.add(
+                key=key,
+                operator=operator,
+                values=field.get("values", []),
+            )
```
hbelmiro: Sorry for not noticing earlier, but we could create a helper method to eliminate duplicate code here. Both loops do almost the same thing.

kubernetes_platform/python/kfp/kubernetes/node_affinity.py
```diff
@@ -0,0 +1,138 @@
...
+
+    Raises:
+        ValueError: If the dictionary does not conform to the V1NodeAffinity schema.
+    """
+    from kubernetes import client
+
+    try:
+        k8s_model_dict = common.deserialize_dict_to_k8s_model_keys(node_affinity_json)
+        client.V1NodeAffinity(**k8s_model_dict)
+        return True
```
hbelmiro: nit: It doesn't need to return anything (can be void).

backend/src/v2/driver/k8s.go
```diff
@@ -482,6 +482,87 @@ func extendPodSpecPatch(
...
+				}
+
+				k8snodeAffinity, err := json.Marshal(nodeAffinityJSON)
+				if err != nil {
+					return fmt.Errorf("failed to marshal node affinity json: %w", err)
+				}
+				var nodeAffinity k8score.NodeAffinity
+				if err := json.Unmarshal(k8snodeAffinity, &nodeAffinity); err != nil {
+					return fmt.Errorf("failed to unmarshal node affinity json: %w", err)
+				}
```
hbelmiro: I may be missing something, but it doesn't seem right to me. Marshal and then unmarshal?

backend/src/v2/driver/k8s.go
```diff
@@ -482,6 +482,87 @@ func extendPodSpecPatch(
 		podSpec.Containers[0].VolumeMounts = append(podSpec.Containers[0].VolumeMounts, emptyDirVolumeMount)
 	}
 
+	// Get node affinity information
+	if nodeAffinityTerms := kubernetesExecutorConfig.GetNodeAffinity(); len(nodeAffinityTerms) > 0 {
+		var requiredTerms []k8score.NodeSelectorTerm
+		var preferredTerms []k8score.PreferredSchedulingTerm
+
+		for _, nodeAffinityTerm := range nodeAffinityTerms {
```
hbelmiro: It might be worth adding a quick check at the start of the loop to skip empty terms early on:

```go
for i, nodeAffinityTerm := range nodeAffinityTerms {
    if nodeAffinityTerm.GetNodeAffinityJson() == nil && 
       len(nodeAffinityTerm.GetMatchExpressions()) == 0 && 
       len(nodeAffinityTerm.GetMatchFields()) == 0 {
        glog.Warningf("NodeAffinityTerm %d is empty, skipping", i)
        continue
    }
    // rest of processing...
}
```

Right now, we only catch empty explicit terms later in the function, but JSON terms could also be empty and we'd still try to process them. This would save some unnecessary work and give clearer debugging info.

kubernetes_platform/python/test/unit/test_node_affinity.py
```diff
@@ -0,0 +1,593 @@
...
+                                    'nodeAffinityJson': {
+                                        'componentInputParameter': 'affinity_input'
+                                    }
+                                }]
+                            }
+                        }
+                    }
+                }
+            }
+        }
```
hbelmiro: These two tests look exactly the same.

kubernetes_platform/python/test/unit/test_node_affinity.py
```diff
@@ -0,0 +1,593 @@
...
+            kubernetes.add_node_affinity_json(
+                task,
+                node_affinity_json=affinity_input_1,
+            )
+            # This should overwrite the previous JSON
+            kubernetes.add_node_affinity_json(
+                task,
+                node_affinity_json=affinity_input_2,
+            )
+        print(json_format.MessageToDict(my_pipeline.platform_spec))
```
hbelmiro: Do we really want to keep this print, or is it a leftover?

VaniHaripriya: Its a leftover :-)

backend/src/v2/driver/k8s.go
VaniHaripriya: Updated the code to avoid both marshal and then unmarshal.

PR #12035, opened 2025-07-09
docs(KEP): K8s native api test plan (WIP)

nsingla:
**Checklist:**


proposals/11551-kubernetes-native-api/TestPlan.md
```diff
@@ -0,0 +1,87 @@
...
+| (K8s Mode) Create an experiment and a run                                                                                            |                                                                                                                                      |                                                                                                              |
+
+#### Without migration script
+Create few pipelines in DB mode and migrate it by recreating those via CRs
+
+| **Test Case**                                                                                                                                                                  | **Test Steps**                                        | **Expected Result** |
+|--------------------------------------------------------------------------------------------------------------------------------------------------------------------------------|-------------------------------------------------------|---------------------|
```
alyssacgoins: What is the reasoning behind the mostly empty boxes under Expected Result?

proposals/11551-kubernetes-native-api/TestPlan.md
```diff
@@ -0,0 +1,87 @@
...
+| (KFP DB mode) Create pipeline and a pipeline version and re-create the pipeline & version using CRs                                                                            | Match the pipeline and pipeline version (DB vs K8s)   |                     |
```
alyssacgoins: Will the same pipeline(s) be used across all tests - what type of pipeline (ie nested, loops, etc) would be used for testing?

proposals/11551-kubernetes-native-api/TestPlan.md
nsingla: in some cases I did not know what the expected out is, and other I am in the process of updating it

i will add `??` in cases I am not sure of the expected outcome to make it clear

proposals/11551-kubernetes-native-api/TestPlan.md
nsingla: I don't think it matters which pipeline we use for most of the tests as long as we've verified that this test case(s) passed: https://github.com/kubeflow/pipelines/pull/12035/files#diff-ce901021cb6143c487145ee10bf9d944af4337fd356119dea8292577d53c47eaR16

Also, I think I should add this test case under Create New Pipelines in K8s mode as well

proposals/11551-kubernetes-native-api/TestPlan.md
```diff
@@ -0,0 +1,87 @@
...
+
+| **Test Case**                                                                                                                        | **Test Steps**                                                                                                                        | **Expected Result**                                                                                          |
+|--------------------------------------------------------------------------------------------------------------------------------------|---------------------------------------------------------------------------------------------------------------------------------------|--------------------------------------------------------------------------------------------------------------|
+| (KFP DB mode) Create a pipeline and a pipeline version and run migrate script                                                        |                                                                                                                                       | Pipeline and pipeline version should match exactly except for ID?                                            |
```
mprahl: Yes, but the ID should match in the `pipelines.kubeflow.org/original-id` annotation. Additionally, we need to verify that the `spec.pipelineName` field matches to the pipeline name on the pipeline version.

proposals/11551-kubernetes-native-api/TestPlan.md
```diff
@@ -0,0 +1,87 @@
...
+| (KFP DB mode) Create 2 pipelines and 1 pipeline version per pipeline spec and run migrate script                                     |                                                                                                                                       | Pipelines and pipeline versions should match exactly except for IDs?                                         |
+| (KFP DB mode) Create 2 pipeline and 2 pipeline versions with different specs and run migrate script                                  |                                                                                                                                       | Pipelines and pipeline versions should match exactly except for IDs?                                         |
+| (KFP DB mode) Create all pipelines and 1 pipeline version per pipeline spec under "data/pipeline_files/valid" and run migrate script | Run [Pipeline Upload API Tests](https://github.com/nsingla/kfp_pipelines/tree/pipeline_run_tests) without clean up and run migration  | Confirm that the migration script does not error out and is able to migrate all pipelines and their versions |
```
mprahl: What is "data/pipeline_files/valid"?

proposals/11551-kubernetes-native-api/TestPlan.md
```diff
@@ -0,0 +1,87 @@
...
+| (K8s Mode) Create a pipeline in DB mode and switch to k8s mode and try creating the same pipeline without migration                  |                                                                                                                                       | Should disallow duplicate pipeline which hasn't been migrated?                                               |
```
mprahl: It will actually work because no pipelines exist after you switch from DB mode to Kubernetes mode. Perhaps we can remove this test case?

proposals/11551-kubernetes-native-api/TestPlan.md
```diff
@@ -0,0 +1,87 @@
...
+| (K8s Mode) Create a pipeline run in an experiment in DB mode and while the run is in progress switch to k8s mode                     |                                                                                                                                       | The run continue after migration                                                                             |
+| (K8s Mode) Create a pipeline recurring run in an experiment in DB mode and switch to k8s mode                                        |                                                                                                                                       | The run should still exist with the same cron settings                                                       |
```
mprahl: I think these will just work but I'm debating if we want to commit/guarantee this works considering it's just a one time migration and it could be tricky to automate. What do you think @HumairAK ?

proposals/11551-kubernetes-native-api/TestPlan.md
```diff
@@ -0,0 +1,87 @@
+## Feature Summary
+Please refer [this](https://github.com/kubeflow/pipelines/blob/master/proposals/11551-kubernetes-native-api/README.md) for more details
+
+## Test Sections
+### Migration of existing pipelines
+#### With the migration script
```
mprahl, line 6: I think I'd like to consider modifying existing tests (e.g. sample tests) so that rather than directly starting runs from the pipeline specs, pipelines and pipeline versions are created. Then we can have scenarios of running the same tests:

1. DB mode and run the pipelines
2. Upload the pipelines in DB Mode, migrate + additional migration verifications, and then run the pipelines
3. Kubernetes mode only and run the pipelines

This would get us a lot of the coverage in this doc without having to maintain separate tests.

proposals/11551-kubernetes-native-api/TestPlan.md
```diff
@@ -0,0 +1,87 @@
...
+| (K8s Mode) Run an existing pipeline and get run details and match                                                                    |                                                                                                                                       |                                                                                                              |
+| (K8s Mode) Run all migrated pipeline and match the run details                                                                       |                                                                                                                                       |                                                                                                              |
+| (K8s Mode) Create an experiment and a run                                                                                            |                                                                                                                                       |                                                                                                              |
```
mprahl: What aspect are you wanting to cover by running the pipelines? In essence, it's just covering a get request to the Kubernetes API similar to just calling the REST API endpoint for a pipeline version. Perhaps we can simplify this.

proposals/11551-kubernetes-native-api/TestPlan.md
```diff
@@ -0,0 +1,87 @@
...
+|-----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------|---------------------------------------------------------|---------------------------------------------------------------------------------------------------------------------------------------------------------------------|
+| (KFP DB mode) Create pipeline and a pipeline version and re-create the pipeline & version using CRs in K8s mode                                                                         | Match the pipeline and pipeline version (DB vs K8s)     | Pipeline and pipeline version should match exactly except for ID?                                                                                                   |
```
mprahl: Is this using the REST API in both cases and testing the translation of API request to Kubernetes CR?

proposals/11551-kubernetes-native-api/TestPlan.md
```diff
@@ -0,0 +1,87 @@
...
+| (KFP DB mode) Create pipeline and 2 pipeline versions with different specs and and re-create the pipeline & version using CRs (in K8s mode)                                             | Match the pipelines and pipeline versions (DB vs K8s)   | Pipeline and pipeline version should match exactly except for ID?                                                                                                   |
+| (KFP DB mode) Create pipeline and a pipeline version and and re-create the pipeline & version using CRs but with different **pipelineName** than the one in pipelineSpecs (in K8s mode) |                                                         | Should this be allowed? Will webhooks throw a validation error                                                                                                      |
```
mprahl: The pipeline name in the pipeline spec is used as a default value when uploading a pipeline but doesn't need to match the CR name. I think this is a cut candidate.

proposals/11551-kubernetes-native-api/TestPlan.md
```diff
@@ -0,0 +1,87 @@
...
+| (K8s mode) Create an experiment and a pipeline run associating it with a migrated pipeline                                                                                              | Get Experiment and Run details and validate             | Pipeline run should be allowed and the experiment association (to the specific experiment) should be correctly saved in the DB                                      |
+| (K8s mode) Rerun a pipeline run associated with an existing experiment and a migrated pipeline version                                                                                  | Get Experiment and Run details and validate             | Pipeline run should be allowed and the experiment association (to the specific experiment) should stay inplace, but how does it remain the same if ID is different? |
```
mprahl: Experiments aren't impacted by this feature so if you can reduce some of these test cases with that in mind, please do.

proposals/11551-kubernetes-native-api/TestPlan.md
```diff
@@ -0,0 +1,87 @@
...
+| (K8s mode) Create pipeline run (via UI)                                                                                              | Get Run details and validate                                                                                                | Run should be successful                                                |
+| (K8s mode) Create an experiment and a pipeline run  (via UI)                                                                         | Get Experiment and Run details and validate                                                                                 | Run should be associated to the correct experiment and should succeed   |
+| (KFP DB mode) Create all pipelines and 1 pipeline version per pipeline spec under "data/pipeline_files/valid" and run migrate script | Use pipelines defined in [Pipeline Upload API Tests](https://github.com/nsingla/kfp_pipelines/tree/pipeline_run_tests)      | Confirm that you are able to create all pipelines and pipeline versions |
+
+### Custom K8s storage (with custom role binding)
+Should we care about this? from migration script POV?
```
mprahl: Could you elaborate a bit more on this?

proposals/11551-kubernetes-native-api/TestPlan.md
```diff
@@ -0,0 +1,87 @@
...
+
+### Custom K8s storage (with FIPS)
```
mprahl: FIPS support isn't something we pursue upstream so perhaps this would be better for Open Data Hub discussions.

proposals/11551-kubernetes-native-api/TestPlan.md
```diff
@@ -0,0 +1,87 @@
...
+| (K8s mode) Create pipeline and 2 pipeline versions with same spec using CRs       | Match the pipeline and pipeline versions (DB vs K8s)  | Pipeline and Pipeline versions creation should be successful                                       |
+| (K8s mode) Create pipeline and 2 pipeline versions with different specs using CRs | Match the pipelines and pipeline versions (DB vs K8s) | Pipeline and Pipeline versions creation should be successful                                       |
+| (K8s mode) Create pipeline run                                                    | Get Run details and validate                          | Pipeline run should succeed                                                                        |
+| (K8s mode) Create an experiment and a pipeline run                                | Get Experiment and Run details and validate           | Pipeline run should be correctly associated to the created experiment and run should be successful |
+
+### Custom K8s storage (with Kubeflow)
```
mprahl: What does "Custom" mean here?

proposals/11551-kubernetes-native-api/TestPlan.md
```diff
@@ -0,0 +1,87 @@
...
+| (K8s mode) Create pipeline and 2 pipeline versions with same spec using CRs       | Match the pipeline and pipeline versions (DB vs K8s)  | Pipeline and Pipeline versions creation should be successful                                       |
+| (K8s mode) Create pipeline and 2 pipeline versions with different specs using CRs | Match the pipelines and pipeline versions (DB vs K8s) | Pipeline and Pipeline versions creation should be successful                                       |
+| (K8s mode) Create pipeline run                                                    | Get Run details and validate                          | Pipeline run should succeed                                                                        |
+| (K8s mode) Create an experiment and a pipeline run                                | Get Experiment and Run details and validate           | Pipeline run should be correctly associated to the created experiment and run should be successful |
+
+### Custom K8s storage (in ODH)
```
mprahl: Can the coverage from above just be inherited in ODH as we rebase? Since this is upstream, perhaps we can remove this section.

proposals/11551-kubernetes-native-api/TestPlan.md
nsingla: Done

proposals/11551-kubernetes-native-api/TestPlan.md
nsingla: Added expected result to all test cases

proposals/11551-kubernetes-native-api/TestPlan.md
nsingla: oh ok, so we will have a new field called `original-id`, gtk, I will add this to the expected result

as for other verification, my thinking is that, we should validate that all fields in the `spec` should match after the migration

proposals/11551-kubernetes-native-api/TestPlan.md
nsingla: that's where I am keeping full list of  pipeline files, so I thought it will be easy to create 100+ pipelines if there is something to refer to

proposals/11551-kubernetes-native-api/TestPlan.md
nsingla: so duplicate pipeline will be allowed if the migration script hasn't been run? I was thinking may be some customers may be in some limbo mode where they want new pipelines to be in K8s mode but still in the process of migrating over older pipelines, in which case if new ones are created with same name, then the migration will become complex. And also, I believe @VaniHaripriya mentioned that in this case, duplicate won;t be allowed

proposals/11551-kubernetes-native-api/TestPlan.md
nsingla: we don't need to automate it right away, but I think we should still verify it, even if its a manual verification

proposals/11551-kubernetes-native-api/TestPlan.md
nsingla: I just want to make sure that its working (we don;t need to concerned with the status of the run as long as run gets created)

proposals/11551-kubernetes-native-api/TestPlan.md
nsingla: the source of creation of pipeline and pipeline version in DB mode doesn't matter but the idea is to see if the pipeline can be recreated using CR (in case a customer does not want to use migration script and create their own using CRs)

proposals/11551-kubernetes-native-api/TestPlan.md
nsingla: hmmm.... so that means we `pipeline` will have different name than in `pipelineversion`. In that case, we need to:
1. make sure existing run(s) associated to these migrated pipelines still work 
2. You can create new run

proposals/11551-kubernetes-native-api/TestPlan.md
nsingla: Its just trying to make sure that experiments and runs still work as runs are associated with pipeline versions which are getting impacted (its like an internal integration test)
We can probably reduce the scope but assuming that if its works in 1 cluster config, it works in other as well, but we should still verify

proposals/11551-kubernetes-native-api/TestPlan.md
nsingla: I may be overthinking here, but I was thinking that what if customer's K8s storage is a custom DB with custom role bindings. Will the migration script still work? or run into some issues :thinking: 

but then again, this might just be that customer's problem that they themselves need to resolve by running scripts and associated components with the right role

proposals/11551-kubernetes-native-api/TestPlan.md
nsingla: agree, I just wanted to add it here but we can perform the testing when we pull the changes to DSP

proposals/11551-kubernetes-native-api/TestPlan.md
nsingla: I think its a copy pasta issue, will fix the titles :)

proposals/11551-kubernetes-native-api/TestPlan.md
nsingla: Yup, but as mentioned before I just wanted to capture that this needs to be done

proposals/11551-kubernetes-native-api/TestPlan.md
nsingla: Yup, that was my intention as well, infact I have 100s of pipeline yaml here: https://github.com/nsingla/kfp_pipelines/tree/e2e-tests/data/pipeline_files/valid which we can use to create pipelines in bulk

proposals/11551-kubernetes-native-api/TestPlan.md
mprahl: @nsingla the configuration is all or nothing so you can run the migration tool and apply the Kubernetes manifests it generates,  but it won't take effect until the configuration changed.

proposals/11551-kubernetes-native-api/TestPlan.md
mprahl: The migration script uses the REST API so if the user has access to the namespace in KFP, then they'll be able to use the migration script.

proposals/11551-kubernetes-native-api/TestPlan.md
```diff
@@ -0,0 +1,87 @@
...
+| (K8s Mode) Run all migrated pipeline and match the run details                                                                       |                                                                                                                                       |                                                                                                                                                                          |
+| (K8s Mode) Create an experiment and a run                                                                                            |                                                                                                                                       |                                                                                                                                                                          |
+
+#### Without migration script
```
mprahl: I'm not quite following what this is covering.

proposals/11551-kubernetes-native-api/TestPlan.md
nsingla: :thinking: I guess that's another test case we need to test that it does not cause any issues with API server when its trying to look for duplicates

nsingla: Looks like this is outdated, newer test case is:
| (K8s Mode) Create a pipeline in DB mode and switch to k8s mode and try creating the same pipeline without migration                  |                                                                                                                                       | New pipeline with the same name will be created, but what happens to existing runs associated with it? and what happens when we run the migration script?                |

proposals/11551-kubernetes-native-api/TestPlan.md
nsingla: Moved these out

proposals/11551-kubernetes-native-api/TestPlan.md
nsingla: This is to make sure that if someone is using direct CR to migrate pipelines (without using migration script), the behavior is the same - for users who would want to write their custom migration scripts

mprahl: @nsingla do you think we could cut back on this section since to me, it's covering the same as just using CRs directly? The context of it previously being in the database does not add test coverage since the IDs change as part of the migration and thus, existing runs associated with the DB pipeline version IDs do not associate with the new ones created via CRs.

Perhaps we can just add coverage that is to directly define pipelines via CRs rather than through the REST API. I think this is covered well in unit tests but not in integration tests.

nsingla: Done, reduced it to just 4 test cases

proposals/11551-kubernetes-native-api/TestPlan.md
```diff
@@ -0,0 +1,69 @@
...
+| (K8s mode) Create pipeline run (via UI)                                                                                              | Get Run details and validate                                                                                                | Run should be successful                                                |
+| (K8s mode) Create an experiment and a pipeline run  (via UI)                                                                         | Get Experiment and Run details and validate                                                                                 | Run should be associated to the correct experiment and should succeed   |
+| (KFP DB mode) Create all pipelines and 1 pipeline version per pipeline spec under "data/pipeline_files/valid" and run migrate script | Use pipelines defined in [Pipeline Upload API Tests](https://github.com/nsingla/kfp_pipelines/tree/pipeline_run_tests)      | Confirm that you are able to create all pipelines and pipeline versions |
+
+### Custom K8s storage (with custom role binding)
```
mprahl: Let's remove this section.

proposals/11551-kubernetes-native-api/TestPlan.md
mprahl: My preference is to just remove this section entirely and rely on the "New Pipelines" for this.

proposals/11551-kubernetes-native-api/TestPlan.md
nsingla: Removed

proposals/11551-kubernetes-native-api/TestPlan.md
nsingla: Removed, and consolidated a test case with a New Pipelines section test

mprahl (approved):
/approve
/lgtm

PR #12037, opened 2025-07-10
[WIP] Isolate GormDBDataType() impact from varchar length changes

kaikaila:
This PR serves as a control group for PR # 12013 to locate the bug.  This PR isolates and tests the CI impact of introducing `GormDBDataType()` without making any changes to varchar lengths.

### Background
In the main `chore/gorm-v2-migration` PR, both `GormDBDataType()` and varchar length adjustments were introduced simultaneously, making it hard to identify the root cause of recent CI failures.

### This PR Includes:
- Only the replacement of `gorm:"type:xxx"` with `gorm:"gormDBDataType:xxx"` using our helper function
- No change to varchar or string field lengths
- No changes to schema constraints or validations

### Purpose:
To test whether `GormDBDataType()` alone causes sample test or integration test failures under MySQL (default CI dialect).

---

Once this PR is marked `ok to test`, we'll be able to confirm whether this refactor is safe in isolation. If it fails, it provides a clear signal that further investigation into `GormDBDataType()` logic is needed.

### DIFF
--- a/backend/src/apiserver/client/sql.go
+++ b/backend/src/apiserver/client/sql.go
@@ -22,11 +22,9 @@ import (
 )
 
 const (
-	MYSQL_TEXT_FORMAT      string = "longtext not null"
-	MYSQL_TEXT_FORMAT_NULL string = "longtext"
-	MYSQL_EXIST_ERROR      string = "database exists"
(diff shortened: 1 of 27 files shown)


kaikaila:
The root cause is the uncorrect implementation of GormDBDataType()

PR #12049, opened 2025-07-15
feat(backend): Support more than one label & annotations setting per component

### LINKED ISSUE #11077: [backend] kfp-kubernetes sets incorrect pod metadata in pipelines with multiple tasks
### Environment

*  How did you deploy Kubeflow Pipelines (KFP)? From the manifests
*  KFP version: 2.2.0
*  KFP SDK version: 
```
kfp                                2.8.0
kfp-kubernetes                     1.2.0
kfp-pipeline-spec                  0.3.0
kfp-server-api                     2.0.5
```

### Steps to reproduce

Here is a minimal example pipeline.
```
from kfp import dsl
from kfp.kubernetes import add_pod_label, add_pod_annotation

@dsl.component
def a_op() -> int:
    return 1

@dsl.component
def b_op() -> int:
    return 2

@dsl.pipeline
def pipe() -> None:
    a = a_op()
    add_pod_label(a, "task", "a")
    add_pod_annotation(a, "task", "a")
    b = b_op()
    add_pod_label(b, "task", "b")
    add_pod_annotation(b, "task", "b")
```

This is the generated pipeline and platform spec, which looks correct.
<details>
<summary> Click to expand </summary>

```
pipeline_spec:
  components:
    comp-a-op:
      executorLabel: exec-a-op
      outputDefinitions:
        parameters:
          Output:
            parameterType: NUMBER_INTEGER
    comp-b-op:
      executorLabel: exec-b-op
      outputDefinitions:
        parameters:
          Output:
            parameterType: NUMBER_INTEGER
  deploymentSpec:
    executors:
      exec-a-op:
        container:
          args:
            - '--executor_input'
            - '{{$}}'
            - '--function_to_execute'
            - a_op
          command:
            - sh
            - '-c'
            - >

              if ! [ -x "$(command -v pip)" ]; then
                  python3 -m ensurepip || python3 -m ensurepip --user || apt-get install python3-pip
              fi


              PIP_DISABLE_PIP_VERSION_CHECK=1 python3 -m pip install --quiet
              --no-warn-script-location 'kfp==2.8.0' '--no-deps'
              'typing-extensions>=3.7.4,<5; python_version<"3.9"' && "$0" "$@"
            - sh
            - '-ec'
            - >
              program_path=$(mktemp -d)


              printf "%s" "$0" > "$program_path/ephemeral_component.py"

              _KFP_RUNTIME=true python3 -m
              kfp.dsl.executor_main                        
              --component_module_path                        
              "$program_path/ephemeral_component.py"                        
              "$@"
            - |+

              import kfp
              from kfp import dsl
              from kfp.dsl import *
              from typing import *

              def a_op() -> int:
                  return 1

          image: 'python:3.8'
      exec-b-op:
        container:
          args:
            - '--executor_input'
            - '{{$}}'
            - '--function_to_execute'
            - b_op
          command:
            - sh
            - '-c'
            - >

              if ! [ -x "$(command -v pip)" ]; then
                  python3 -m ensurepip || python3 -m ensurepip --user || apt-get install python3-pip
              fi


              PIP_DISABLE_PIP_VERSION_CHECK=1 python3 -m pip install --quiet
              --no-warn-script-location 'kfp==2.8.0' '--no-deps'
              'typing-extensions>=3.7.4,<5; python_version<"3.9"' && "$0" "$@"
            - sh
            - '-ec'
            - >
              program_path=$(mktemp -d)


              printf "%s" "$0" > "$program_path/ephemeral_component.py"

              _KFP_RUNTIME=true python3 -m
              kfp.dsl.executor_main                        
              --component_module_path                        
              "$program_path/ephemeral_component.py"                        
              "$@"
            - |+

              import kfp
              from kfp import dsl
              from kfp.dsl import *
              from typing import *

              def b_op() -> int:
                  return 2

          image: 'python:3.8'
  pipelineInfo:
    name: pipe
  root:
    dag:
      tasks:
        a-op:
          cachingOptions:
            enableCache: true
          componentRef:
            name: comp-a-op
          taskInfo:
            name: a-op
        b-op:
          cachingOptions:
            enableCache: true
          componentRef:
            name: comp-b-op
          taskInfo:
            name: b-op
  schemaVersion: 2.1.0
  sdkVersion: kfp-2.8.0
platform_spec:
  platforms:
    kubernetes:
      deploymentSpec:
        executors:
          exec-a-op:
            podMetadata:
              annotations:
                task: a
              labels:
                task: a
          exec-b-op:
            podMetadata:
              annotations:
                task: b
              labels:
                task: b
```

</details>

However, both pods for the two pipeline tasks get the label and annotation `a`.

I did some more testing, and it seems that for all tasks, the platform spec of the one that is alphabetically first is taken. If you rename the first task to `c_op`, both pods get the label / annotation `b`. This might be because the first platform spec in the YAML is always taken.

### Expected result

The pod metadata is set correctly for all tasks. In the above example, task `a` should have label / annotation `a`, task `b` should have label / annotation `b`.

### Materials and Reference

The pull request that introduced the feature to set labels / annotations: https://github.com/kubeflow/pipelines/pull/10393/

I am not too familiar with the codebase, but it might have something to do with not picking the correct `KubernetesExecutorConfig`.

---

alyssacgoins:
Resolves #11077 
This PR parameterizes pod metadata similarly to how task retryStrategy was parameterized in #11861 , in order to set task-specific metadata.

**Checklist:**


samples/v2/pipeline_with_pod_metadata.py
```diff
@@ -0,0 +1,108 @@
...
+            )
+        validated_metadata_count+=1
+    if label_path_2 is not None:
+        label_val_2 = os.getenv(label_path_2)
+        if label_val_2 is not None and label_val_2 != label_exp_val_2:
+            raise ValueError(
+                f"Pod label is {label_val_2} but is supposed to be {label_exp_val_2}."
+            )
+        validated_metadata_count+=1
+    if validated_metadata_count <1:
```
HumairAK: you can further solidify this by adding an expected count of metadata we are triggering

though tbh I think it's just simpliar to have multiple components for task_a and task_b and hardcode the values like this: 

```python
    annotation_val_1 = os.getenv("POD_TASK_ANNOTATION", "")
    if annotation_val_1 != annotation_exp_val_1:
        raise ValueError(
            f"Pod annotation is {annotation_val_1} but is supposed to be {annotation_exp_val_1}."
        )
```

though that's optional, personally I think this would be easier to read at the cost of some hardcoding

samples/v2/pipeline_with_pod_metadata.py
```diff
@@ -0,0 +1,108 @@
...
+    add_pod_annotation(task_b, 'task-annotation-2', 'annotation-2')
+
+    use_field_path_as_env(task_b, 'POD_TASK_ANNOTATION_1', "metadata.annotations['task-annotation-1']")
+    use_field_path_as_env(task_b, "POD_TASK_ANNOTATION_2", "metadata.annotations['task-annotation-2']")
+
+    # task c is set with no metadata
+    task_c = validate_no_pod_metadata(annotation_path="POD_TASK_ANNOTATION", label_path="POD_TASK_LABEL").set_caching_options(False)
+    use_field_path_as_env(task_c, "POD_TASK_ANNOTATION", "metadata.annotations['task-annotation']")
+    use_field_path_as_env(task_c, "POD_TASK_LABEL", "metadata.labels['task-label']")
+
```
HumairAK, line 247: I think it would be useful to have a case where we have # of annotations and # of labels that are greater than in task_a (and not just annotations like in task_b) but are also different from each other, so something like 3 tasks, 4 annotations

backend/src/v2/compiler/argocompiler/container.go
```diff
@@ -265,16 +266,23 @@ func (c *workflowCompiler) containerExecutorTask(name string, inputs containerEx
 	if inputs.condition != "" {
 		when = inputs.condition + " != false"
 	}
+	refName := task.GetComponentRef().GetName()
+	// Retrieve pod metadata defined in the Kubernetes Spec, if any
+	kubernetesConfigParam := c.wf.Spec.Arguments.GetParameterByName(argumentsKubernetesSpec + refName)
+	k8sExecCfg := &kubernetesplatform.KubernetesExecutorConfig{}
+	if kubernetesConfigParam != nil {
+		jsonpb.UnmarshalString(string(*kubernetesConfigParam.Value), k8sExecCfg)
+	}
```
HumairAK: this code is making some assumptions about values that while they may always be true, we should still add nil checks (and handle errors in the case of unmarshaling), 

here's a suggested update: 

```go
var refName string
if componentRef := task.GetComponentRef(); componentRef != nil {
    refName = componentRef.GetName()
} else {
    // Handle nil component ref case
    return nil, fmt.Errorf("component reference is nil")
}

// Retrieve pod metadata defined in the Kubernetes Spec, if any
kubernetesConfigParam := c.wf.Spec.Arguments.GetParameterByName(argumentsKubernetesSpec + refName)
k8sExecCfg := &kubernetesplatform.KubernetesExecutorConfig{}
if kubernetesConfigParam != nil && kubernetesConfigParam.Value != nil {
    if err := jsonpb.UnmarshalString(string(*kubernetesConfigParam.Value), k8sExecCfg); err != nil {
        return nil, fmt.Errorf("failed to unmarshal kubernetes config: %v", err)
    }
}

```

backend/src/v2/compiler/argocompiler/container.go
```diff
@@ -265,16 +266,23 @@ func (c *workflowCompiler) containerExecutorTask(name string, inputs containerEx
...
-		Template: c.addContainerExecutorTemplate(task),
+		Template: c.addContainerExecutorTemplate(task, k8sExecCfg),
 		When:     when,
 		Arguments: wfapi.Arguments{
 			Parameters: append([]wfapi.Parameter{
 				{Name: paramPodSpecPatch, Value: wfapi.AnyStringPtr(inputs.podSpecPatch)},
 				{Name: paramCachedDecision, Value: wfapi.AnyStringPtr(inputs.cachedDecision), Default: wfapi.AnyStringPtr("false")},
 			},
-				c.getTaskRetryParametersWithValues(task)...,
+				append(c.getPodMetadataParameters(k8sExecCfg.GetPodMetadata(), true), c.getTaskRetryParametersWithValues(task)...)...,
```
HumairAK: nit: this is starting to get a bit hard to read, I would suggest optionally refactoring this to something like: 

```go
parameters := append(
    []wfapi.Parameter{
        {
            Name:  paramPodSpecPatch,
            Value: wfapi.AnyStringPtr(inputs.podSpecPatch),
        },
        {
            Name:    paramCachedDecision,
            Value:   wfapi.AnyStringPtr(inputs.cachedDecision),
            Default: wfapi.AnyStringPtr("false"),
        },
    },
    append(
        c.getPodMetadataParameters(k8sExecCfg.GetPodMetadata(), true),
        c.getTaskRetryParametersWithValues(task)...,
    )...,
)


```

backend/src/v2/compiler/argocompiler/container.go
```diff
@@ -302,6 +310,20 @@ func (c *workflowCompiler) addContainerExecutorTemplate(task *pipelinespec.Pipel
 			nameContainerImpl = "retry-" + nameContainerImpl
 		}
 	}
+	podMetadata := k8sExecCfg.GetPodMetadata()
+	// if pod metadata is set, create a template name including the numbers of annotations and labels.
```
HumairAK, line 337: we should add an inline comment here explaining why we are using this approach (the reason being Argo doesn't support patching of the metadata field for a pod as a string, like podspecpatch, and then link to the github issue you created in upstream argo), prefix the comment with a TODO, so like: 

```
// TODO (agoins): <reasoning>
```

backend/src/v2/compiler/argocompiler/container.go
```diff
@@ -627,7 +644,81 @@ func (c *workflowCompiler) getTaskRetryStrategyFromInput(maxCount string, backOf
...
+						},
+						{
+							Name:  val,
+							Value: wfapi.AnyStringPtr(labels[labelKey]),
+						}}...)
+			} else {
+				parameters = append(parameters, []wfapi.Parameter{{Name: key}, {Name: val}}...)
+			}
+			count++
+		}
```
HumairAK: nit/optional: this logic seems like it's repeating the same functionality as for annotations, can we turn this into a helper or reduce the duplication somehow?

backend/src/v2/compiler/argocompiler/container.go
alyssacgoins: Updated

backend/src/v2/compiler/argocompiler/container.go
alyssacgoins: Updated - applied to similar block on line 352, 371 & 408 as well.

backend/src/v2/compiler/argocompiler/container.go
alyssacgoins: Makes sense - updated

samples/v2/pipeline_with_pod_metadata.py
alyssacgoins: Updated with two more cases - four annotations/3 labels, 0 annotations/3 labels

samples/v2/pipeline_with_pod_metadata.py
alyssacgoins: Updated - added a component per task. Since I created a separate component per task, I decided against adding an expected metadata count because each component verifies task-specific metadata. Let me know what you think

backend/src/v2/compiler/argocompiler/container.go
alyssacgoins: Refactored - created helper method `formatParameterEntries(),` included below `getPodMetadataParameters()`.

backend/src/v2/compiler/argocompiler/container.go
alyssacgoins: ** Updated line numbers are: 357, 376, 413

HumairAK:
/lgtm
/approve

awesome work!

Issue #12050, opened 2025-07-15
Broken documentation link in NewRunParametersV2.tsx


mustachemo:
https://github.com/kubeflow/pipelines/blob/9aa3dfb0d9810e3d3940b728e2e39dad1e956587/frontend/src/components/NewRunParametersV2.tsx#L245

The file frontend/src/components/NewRunParametersV2.tsx contains a broken link on line 245.
Current link:
https://www.kubeflow.org/docs/components/pipelines/overview/pipeline-root/ (returns 404)

The correct link should be:
https://www.kubeflow.org/docs/components/pipelines/concepts/pipeline-root/

Steps to reproduce:
Go to the UI where the pipeline root documentation link is rendered.
Click the link.
Observe that it leads to a 404 page.
Suggested fix:
Update the URL in NewRunParametersV2.tsx to use the correct link.

uday-codes69:
Hi @mustachemo, I’d like to work on this issue. Could you please assign it to me?

I’ll update the broken link in `NewRunParametersV2.tsx` as described and open a PR shortly. Thanks!

droctothorpe:
Ya'll submitted identical PRs 🙃 :

<img width="493" height="99" alt="Image" src="https://github.com/user-attachments/assets/0cf02486-f491-4154-b37f-0081648d2226" />

mustachemo:
what the... this guy really copy pasted my pr

uday-codes69:
Thanks @droctothorpe for pointing this out! I'll look for another issue to work on — feel free to close my PR (#12052) if needed.

PR #12052, opened 2025-07-15
Fix broken pipeline root documentation link

uday-codes69:
Replaces the outdated pipeline root documentation link in NewRunParametersV2.tsx (line 245) with the correct one as suggested in issue #12050


uday-codes69:
Hi @drcotothorpe @mprahl
This is a small fix for a broken documentation link (issue #12050).  
Would be grateful if you could take a look when you get a chance. Thank you!

PR #12061, opened 2025-07-17
fix: backwards compatibility for pipeline spec task_name

zazulam:
This change is to add a check on whether the API server version contains the necessary backend changes to process the `task_name` values on the TaskSpec in the Pipeline IR. Since the `kfp-pipeline-spec` and the `kfp` packages are more tied together I figured the best way to detect the version of the backend was to just check the `kfp-server-api` package version since that is tied the backend version that would handle the new field logic. Please let me know if you think there's a better way to handle this for now. I think there should be a discussion around the approach for handling backwards compatibility, not sure if it should be tied in with the testing/CI discussions.

For reference: https://github.com/kubeflow/pipelines/blob/master/sdk/python/requirements.in#L13-L19

**Checklist:**


droctothorpe:
Thanks for tackling this so quickly, @zazulam!

I think it's a good idea to do some manual validation from local for the following two scenarios:

1. Updated SDK, previous version of API
2. Updated API, previous version of SDK

We can look into automating that kind of backwards compatibility validation in a discrete PR.

zazulam:
> Thanks for tackling this so quickly, @zazulam!
> 
> I think it's a good idea to do some manual validation from local for the following two scenarios:
> 
> 1. Updated SDK, previous version of API
> 2. Updated API, previous version of SDK
> 
> We can look into automating that kind of backwards compatibility validation in a discrete PR.

@droctothorpe, yea I had modified my deployment to have the api server + the driver & launcher point to 2.5.0, the local version of the `kfp-server-api` is still set to 2.5.0 so it was able to launch the pipelines from the v2 samples set where only the ones that failed were relying on what is on master for the backend, like the fixes for collected and display_name.

I'll make sure to test the updated backend with the previous SDK and report back.

zazulam:
After discussing offline with others, changed course and opted to remove the new field from the pipeline spec and the logic in the sdk to add it in and place it in the compiler to hand off to the driver. This way the amount of code necessary for the original unique task name is reduced and sdk backwards compatibility is preserved for 2.13 when the new release will be cut (if this gets merged in by then).

This meant having to recompile all the test cases that were added before to remove the field, hence all the yaml diffs 😅 

It seems to me that the backend test failing `TestCreatePipelineV1_LargeFile` with the error below means there _might_ be some mismatch of a stub that is being tested against the pipeline server.
```
=== RUN   TestCreatePipelineV1_LargeFile
    pipeline_server_test.go:104: 
        	Error Trace:	/home/runner/work/pipelines/pipelines/backend/src/apiserver/server/pipeline_server_test.go:104
        	Error:      	Expected nil, but got: &util.UserError{internalError:(*errors.withStack)(0xc0014ba5b8), externalMessage:"unknown template format", externalStatusCode:0x3}
        	Test:       	TestCreatePipelineV1_LargeFile
    pipeline_server_test.go:105: 
        	Error Trace:	/home/runner/work/pipelines/pipelines/backend/src/apiserver/server/pipeline_server_test.go:105
        	Error:      	Expected value not to be nil.
        	Test:       	TestCreatePipelineV1_LargeFile
--- FAIL: TestCreatePipelineV1_LargeFile (0.17s)
```

zazulam:
Turns out we test against a pipeline IR from master:  
[pipeline](https://github.com/kubeflow/pipelines/blob/master/sdk/python/test_data/pipelines/xgboost_sample_pipeline.yaml#L762-L764) in question from master with TaskName still set.

[test](https://github.com/kubeflow/pipelines/blob/master/backend/src/apiserver/server/pipeline_server_test.go#L87-L102) using said pipeline.

Not sure how to proceed, the fix works it just fails that test.

cc: @droctothorpe @mprahl @HumairAK

droctothorpe:
Nice catch. That definitely shouldn't be pulling from master. How hard would it be to pass in a filepath or dynamically retrieve and inject the branch name into the URL?

zazulam:
> Nice catch. That definitely shouldn't be pulling from master. How hard would it be to pass in a filepath or dynamically retrieve and inject the branch name into the URL?

It shouldn't be too hard, I can give it go.

droctothorpe:
The org has to change as well, which makes filepath seem more compelling:

```
        	Error:      	Expected nil, but got: &util.UserError{internalError:(*errors.withStack)(0xc00115d500), externalMessage:"error fetching pipeline spec from https://raw.githubusercontent.com/kubeflow/pipelines/back-compat-11938/sdk/python/test_data/pipelines/xgboost_sample_pipeline.yaml - request returned 404 Not Found", externalStatusCode:0x3}
```

I'm okay merging even though this specific test is failing since you have delineated why it's failing and it's not related to the content of this PR. 

/lgtm

zazulam:
> The org has to change as well, which makes filepath seem more compelling:
> 
> ```
>         	Error:      	Expected nil, but got: &util.UserError{internalError:(*errors.withStack)(0xc00115d500), externalMessage:"error fetching pipeline spec from https://raw.githubusercontent.com/kubeflow/pipelines/back-compat-11938/sdk/python/test_data/pipelines/xgboost_sample_pipeline.yaml - request returned 404 Not Found", externalStatusCode:0x3}
> ```
> 
> I'm okay merging even though this specific test is failing since you have delineated why it's failing and it's not related to the content of this PR.
> 
> /lgtm

Yea, I'm gonna revert the commit. When this would be merged the build on master should pass since it will be referencing the proper pipeline spec.

mprahl:
@zazulam let me know once you've fixed the tests but this looks good to me. Thanks for fixing this!

zazulam:
Pulled in the changes from master and now the tests are failing from kind not launching & protobuf dep issues. I can't figure out the diff that would be causing the kind issue or the protobuf problem, honestly feels like the CI is just working against me 🙃 

image
image

zazulam:
confirmation bias incoming
image

zazulam:
@droctothorpe @mprahl managed to get the CI passing 🙏

droctothorpe:
Kudos, Mike! This LGTM. Thanks for resolving the problematic test! 

/approve

Issue #11742, opened 2025-03-11
[backend] Cannot get MLMD objects from Metadata store: received initial metadata size exceeds limit


vkatrychenko:
### Environment

*  How did you deploy Kubeflow Pipelines (KFP)?
deployed kubeflow using k8s manifests (standalone kubeflow components): https://github.com/kubeflow/manifests
*  KFP version: 
kfp 2.4.1
*  KFP SDK version: 
kfp-sdk: 2.12.1, 2.11.0
* MySQL server:
8.0.21 Azure flexible mysql (also, tried to use self-hosted mysql v8.0.26)

### Steps to reproduce

Deploy the latest kubeflow ver v1.10.0-rc.2 with dex/oauth2-proxy auth. Set up Microsoft auth method for dex. Then try to run a pipeline.

![Image](https://github.com/user-attachments/assets/ce8a1404-a637-4085-89a3-33202c007b6b)

When running pipelines under a static kf user, everything is working fine.

![Image](https://github.com/user-attachments/assets/34b0f5b3-3817-47d2-9228-eff6fa742a81)

### Expected result

No errors should appear under pipeline-ui.

### Materials and Reference

It seems that the issue occurs due to large azure oidc info according to this thread: https://github.com/kubeflow/pipelines/issues/8733#issuecomment-2627771197

Tried to downgrade/upgrade grpc ml medata service, but did not help.

Do you have a workaround for this?

BTW: we saw this error in v1.9.x as well

---

Impacted by this bug? Give it a 👍.

vkatrychenko:
For context: everything is working fine in kubeflow 1.7.1.

NJ3rsey:
Fixed by adding the argument to `metadata-grpc-deployment`:
```
spec:
  ...
  template:
    ...
    spec:
      containers:
        - args:
            - '--grpc_channel_arguments=grpc.max_metadata_size=16384' <<-- (16kb instead of default 8kb)
```

juliusvonkohout:
> Fixed by adding the argument to `metadata-grpc-deployment`:
> 
> ```
> spec:
>   ...
>   template:
>     ...
>     spec:
>       containers:
>         - args:
>             - '--grpc_channel_arguments=grpc.max_metadata_size=16384' <<-- (16kb instead of default 8kb)
> ```

@hbelmiro @HumairAK should we make this the default, maybe even 32k ? See also  https://github.com/kubeflow/manifests/tree/master/common/oauth2-proxy#known-issues

[...]

PR #12062, opened 2025-07-17
fix(backend): increase max_metadata_size for ml-metadata grpc server

droctothorpe:
Resolves https://github.com/kubeflow/pipelines/issues/11742. 

Kudos to @NJ3rsey for the fix.

**Checklist:**


HumairAK:
would be nice if we could drop a comment on why this is needed, I don't want to block the pr on that though

/lgtm
/approve

Thanks @droctothorpe and the rest for investigating and resolving this quickly!

droctothorpe:
Copy that, @HumairAK. Adding to the description as well.

Some openidc providers use large JWTs / Cookies that can exceed the default gRPC limit of MLMD. This PR increases the limit from 8 kb (the default) to 16 kb.

After patching the limit, I confirmed that this resolved https://github.com/kubeflow/pipelines/issues/11742 for our end users.

PR #12063, opened 2025-07-17
[WIP] chore(backend): Storage-layer refactor & Postgres enablement roadmap

kaikaila:
Superseded by [#12379](https://github.com/kubeflow/pipelines/pull/12379).
Further development and review will continue there.

## Background

Kubeflow Pipelines currently hardcodes MySQL assumptions in its storage layer. To enable PostgreSQL adoption (via pgx driver) and improve long-term maintainability, we need to decouple SQL dialect logic and centralize it into a reusable abstraction.

⸻

## Changes

This PR introduces a major refactor of the backend storage layer and related test workflows:
Dialect abstraction
	•	Introduced DBDialect interface and refactored pipeline/experiment/run/job stores to use it.
	•	Removed legacy SQLDialect, centralized initialization in InitDBClient.
Storage refactor
	•	Updated resource_reference, db_status_store, and cache/storage clients to use the new dialect.
	•	Moved filter logic into dialect package, cleaned up list package.
	•	Adapted db_fake to support dialect abstraction.
Tests & dependencies
	•	Updated unit tests across storage, list, and filter packages.
	•	Updated go.mod dependencies and tidied modules.
	•	Added PostgreSQL GitHub Actions workflow with kustomize deployment and port-forward.

⸻

## CI Status
- In the upstream CI, the failures are infra-related: tests share the same database and pollute each other.
- Running the tests in parallel jobs resolves this issue and results in all tests passing, as shown in this PR's CI results.
- Per maintainer feedback, PR#12251 was closed to make room for the ongoing test overhaul PR#12222.
- This PR includes the fixes from PR#12251, so CI results here are green.

### DIFF
--- a/backend/src/cache/certs/.gitignore
+++ b/backend/src/cache/certs/.gitignore
@@ -1,0 +1,9 @@
+# Ignore certificate files - these should never be committed
+*.pem
+*.key
+*.crt
+*.p12
+*.pfx
+
+# Keep this .gitignore file
+!.gitignore


HumairAK:
@kaikaila truly appreciate the depth of this plan, I think the steps make sense, but something like this should be delivered as a single PR, it is fine if the PR is larged - this way we can merge the feature as whole rather than in parts. I am fine with docs being a separate PR. The CI should also accompany this change so we can verify the results.

kaikaila:
Thanks a lot, @HumairAK  — really appreciate the clarification. I was splitting the work into smaller PRs mainly to keep the review size manageable (since I’d heard ~300–400 LOC is a good target). But if the preference here is to have the full feature in a single PR, that works perfectly for me.

I’ll consolidate the storage refactor changes together with the CI workflow into one PR, perhaps including docs too.

kaikaila:
Hi team,

I'm working on integrating the PostgreSQL (pgx) backend deployment into the main CI by adding a --db-type pgx flag to the [deploy-kfp.sh](code-assist-path:/Users/yunkaili/Documents/personal/open_source/kubeflow_pipeline/kubeflow-pipelines/.github/resources/scripts/deploy-kfp.sh) script, [as suggested](https://github.com/kubeflow/pipelines/pull/12159#discussion_r2298091874).

Previously, I had a working CI workflow that successfully deployed KFP with a PostgreSQL backend like [this](https://github.com/kaikaila/kubeflow-pipelines/actions/runs/17935999311). However, after refactoring this logic into [deploy-kfp.sh](code-assist-path:/Users/yunkaili/Documents/personal/open_source/kubeflow_pipeline/kubeflow-pipelines/.github/resources/scripts/deploy-kfp.sh) and the associated Kustomize overlays, I'm facing a persistent issue where the ml-pipeline-apiserver pod fails to start.

## Current Status & Problem

The ml-pipeline-apiserver pod is stuck in a CrashLoopBackOff loop.
Logs and events show it fails the startup probe due to a connection refused error, and eventually gets terminated with Exit Code 137.
The net-checker pod in our CI confirms that basic network connectivity to the PostgreSQL service (postgres-service.kubeflow.svc.cluster.local:5432) is working correctly.

## Fixes Already Applied

- Correct Driver Name: Ensured all dbType and DBDriverName values in the PostgreSQL-specific Kustomize manifests are correctly set to pgx.
- Code-level Credential Bug: Identified and fixed a bug in [client_manager.go](code-assist-path:/Users/yunkaili/Documents/personal/open_source/kubeflow_pipeline/kubeflow-pipelines/backend/src/apiserver/client_manager/client_manager.go) where hardcoded credentials ("root", "") were used for the second connection attempt, instead of the credentials from the secret.
- Resource Limits: Increased apiserver memory limits to 4Gi to rule out simple OOM issues, though I believe the exit code is a symptom of the connection failure, not the cause.

Despite these fixes, the apiserver still cannot connect to the database. I suspect I'm missing a subtle but critical piece of configuration in how the Kustomize overlays are structured or how [deploy-kfp.sh](code-assist-path:/Users/yunkaili/Documents/personal/open_source/kubeflow_pipeline/kubeflow-pipelines/.github/resources/scripts/deploy-kfp.sh) selects the correct manifests. The layers of configuration are making it difficult to pinpoint the exact misconfiguration.

Since the root cause of the CI failures in `kfp-backend-v1-posgres-tests.yml` and `kfp-backend-v2-posgres-tests.yml` is essentially the same, I’ve been running all my experiments and collecting logs only on the [v1 workflow](https://github.com/kubeflow/pipelines/actions/runs/17974190580/job/51123960410?pr=12063). The v1 lane reflects my latest progress, while the v2 lane is lagging behind.

Could someone with more experience in the KFP manifest structure take a look? I feel like I'm one small configuration change away from solving this.

Thanks for your help.

kaikaila:
I ran the PGX workflow locally to reproduce the CI failure and uncovered two bugs along the way:

The Postgres credentials were split between two secrets. I unified them so the API server and cache pods now read from the same postgres-secret.
The stack never created the mlpipeline / cachedb databases on first boot. I added an init SQL ConfigMap so those databases are provisioned automatically.
After the change I wiped the PVCs, redeployed the overlay, and the databases show up correctly. However the ml-pipeline pod still never becomes Ready. The logs always stop right after the DB client is initialized:

```I0924 23:48:43.101959      11 client_manager.go:270] Initializing client manager
I0924 23:48:43.102145      11 client_manager.go:271] Initializing DB client...
I0924 23:48:44.180760      11 client_manager.go:559] Running AutoMigrate.
I0924 23:48:44.278840      11 client_manager.go:275] DB client initialized successfully
I0924 23:48:44.279659      11 client_manager.go:287] Initializing Object store client...
```
ChatGPT speculates the API server crashes right after “Initializing Object store client…” because it gets OOMKilled before it even starts listening on the health check port. I’ve tried memory limits from 250 Mi up to 4 Gi, but it still exits with code 137. I ran out of hypotheses about the root cause.

An example run is [here](https://github.com/kaikaila/kubeflow-pipelines/actions/runs/17998654376/job/51203168022).

Do you have any suggestions on what the next step should be to debug this?

kaikaila:
## Core issue

the ml-pipeline pod always dies right after logging Initializing Object store client.... The process never starts listening on 8888, so the startup probe keeps hitting “connection refused,” kubelet issues a SIGKILL (exit code 137), and the container crashloops.

## Investigation so far

Instrumented memory with runtime.ReadMemStats and /debug/pprof; heap stays around 9–12 MB during startup, no abnormal growth. Docker stats show the Kind node has ~12 GiB available and only ~3.5 GiB in use → not a Go heap blow-up.

Increased the startup probe window, shortened the profiling sleep to 2 s, rebuilt ARM64 images via buildx, loaded them into Kind, and confirmed Docker Desktop is giving the VM 12 GiB.

Verified minio-service endpoints/ports/secret (SeaweedFS S3 gateway), ran nc from a BusyBox pod — the service is reachable. SeaweedFS logs only show its own startup; the API server never gets far enough to issue S3 calls.

## Result

it’s still unclear why the process is killed during object store initialization. The profiling changes are in the branch

kaikaila:
Currently, both MySQL and PGX setups use the DB superuser for all KFP operations, which is why client_manager.go contains a “create database if not exist” step [here](https://github.com/kubeflow/pipelines/blob/82f6b7d08e9bf7000974bed8d527a6a05bb24031/backend/src/apiserver/client_manager/client_manager.go#L416). 

From a security standpoint, would it be preferable to:
	1.	Move DB creation out of the client manager and into the deployment/init phase (i.e. add a `manifests/kustomize/third-party/postgresql/base/pg-init-configmap.yaml`) and
	2.	Introduce a dedicated restricted user for KFP components, limited to the mlpipeline database?

If the team agrees, I can propose a follow-up PR to refactor accordingly.

kaikaila:
The blocker is the [legacy v2 test workflow](https://github.com/kubeflow/pipelines/actions/runs/18119783695/job/51562394751?pr=12063), which keeps failing red. I can’t track down the source of data pollution. Here’s what I’ve tried so far:

-	Tried adding `t.Log` and `fmt.Fprint` to print debug logs in CI → but the logs never showed up. Why? 

-	Tried local debugging with launch.json and breakpoints → locally everything goes green, I can’t reproduce the failure.

-	Suspect race condition in `recurring_run_api_test.go` during `TearDownSuite`: cleanup deletes runs, but while cleaning scheduled workflows, SWF can create zombie runs and cleanup doesn’t re-delete those runs. So I added an extra cleanup in `run_api_test.go`’s `SetupTest`, but I still see dirty data.

- Also, when I tried skipping the recurring run test as @HumairAK  suggested, the failure pattern changed (e.g. 0 instead of 2 runs, or 5 instead of 0) [here](https://github.com/kaikaila/kubeflow-pipelines/actions/runs/18213773799/job/51859164657).

My current hypothesis is that it’s related to concurrency or test execution order — when the order happens to work out, everything passes; when it doesn’t, the database gets “polluted” and the assertions fail.
The tricky part is that I haven’t been able to identify the exact source of pollution yet — I’ve added cleanup logic, but it doesn’t seem to fully clear the state.

PR #12068, opened 2025-07-19
chore: remove `env/gcp`

akagami-harsh:
part of https://github.com/kubeflow/pipelines/pull/11965

- removed `env/gcp`


akagami-harsh:
looking into if we need to keep inverse proxy and codesql

juliusvonkohout:
/ok-to-test

@madmecodes or @kunal-511 please raise a PR toadd you 4 as kubeflow organisation member in internal-acls. CC @LogicalGuy77

Issue #12038, opened 2025-07-10
[sdk] High severity CVE-2025-4565 in kfp 1.8.22


droctothorpe:
### Environment
*  KFP SDK version: 
* 1.8.22

### Details
https://www.cve.org/CVERecord?id=CVE-2025-4565
`protobuf` < 6.31.1 is the offending dep. It's pinned [here](https://github.com/kubeflow/pipelines/blob/1.8.22/sdk/python/setup.py#L55). Unfortunately, it's also pinned in the sub-dep, [kfp-server-api](https://github.com/kubeflow/pipelines/blob/1.8.22/sdk/python/setup.py#L46).

I know v1 is technically frozen but this is a high CVE and there's still a ton of v1 adoption.

I plan to tackle this myself, bandwidth permitting.

CC @juliusvonkohout 

---

Impacted by this bug? Give it a 👍.

juliusvonkohout:
THank you @droctothorpe having a 1.8.23 release with updated dependencies and python3.12 compatibility will help a lot to ease the transistion to KFP v2.

droctothorpe:
Finally off support. @cmdevoto made some great progress in the interim. Here are some decisions made / plans going forward:

1. We will target the latest version of the backend for the large tests.
2. We will cut a release-kfp-sdk-1.8.23 branch from master and checkout the sdk folder from release-kfp-sdk-1.8.22. This will align our commit history with master and make the subsequent PR more readable.
3. We will conduct an inventory of the current tests.
4. We will disable any tests that are not SDK specific. Anything that tests the backend in isolation is not relevant to a branch that's used exclusively for SDK releases. That being said, we will preserve large tests that test the SDK in concert with the backend.
5. TBD, but we may elect to disable any v2 specific tests and may even delete v2 specific code from the SDK. Anyone using v2 should not be on a 1.* version of the SDK.

droctothorpe:
[Here's](https://github.com/kubeflow/pipelines/pull/12073) the corresponding PR. It's absolutely gargantuan, but our commits will just be a thin layer on top and we can squash them to make future rebasing easier. We'll have to ignore the DCO failure because a lot of the older commits don't adhere to it.

PR #12073, opened 2025-07-24
[WIP] Retrofit CI for V1 SDK

droctothorpe:
Paves the way to resolve https://github.com/kubeflow/pipelines/issues/12038


droctothorpe:
The prow job config on 1.8.22 is stored in an external repository:
https://github.com/GoogleCloudPlatform/oss-test-infra/tree/48b09567c8df28fab2d3f2fb6df86defa12207fb/prow/prowjobs/kubeflow/pipelines

droctothorpe:
Due to how much the 1.8.22 release branch diverges from master (54 extra commits), we shifted strategies to make this PR much more readable.

droctothorpe:
In an effort to make this PR more human readable, we used the following strategy:

1. Cut a `release-kfp-sdk-1.8.23` branch from `master` in the `kubeflow/pipelines` repository to function as a target for this PR.
2. Ran `git checkout origin/release-kfp-sdk-1.8.22 -- sdk` on the `release-kfp-sdk-1.8.23` branch. This aligns our non-SDK code and, crucially, git commit history with the `master` branch.
3. Cut a `retrofit-ci` branch from `release-kfp-sdk-1.8.23`. This is where we will make the actual changes to ensure compatibility between the v1 SDK and the updated CI.

droctothorpe:
In order to fix the component tests, I had to pull in components from `release-kfp-sdk-1.8.22`. To keep this PR human readable, I did that in the `kfp-release-sdk-1.8.23` branch then rebased that commit into this branch.

droctothorpe:
I scrapped the runtime workflow since it was introduced after the 1.8.22 release.

I also scrapped the sample workflow for the following reasons:
- In master, it only targets v2.
- In 1.8.22, it's an unnecessarily complicated sequence of abstractions (prow > bash script > python script > AWF > KFP pipeline) that ultimately covers [only 4 test cases](https://github.com/kubeflow/pipelines/blob/1.8.22/test/sample_test.yaml#L102-L108).

That being said, I'm not comfortable with the limited large test scope, even if that's the precedent set by 1.8.22, so I'm constructing a bunch of test cases from the samples directory and plan to create a workflow that executes them in parallel using pytest-xdist.

TBH, I don't understand why so many layers of abstraction were implemented. When I'm done, the resulting approach could be a good reference for the test refactor, @nsingla. 

In general, I think it would be helpful to have a table that describes and differentiates the various workflows in master.

nsingla:
> I scrapped the runtime workflow since it was introduced after the 1.8.22.
> 
> I also scrapped the sample workflow for the following reasons:
> 
>     * In master, it only targets v2.
> 
>     * In 1.8.22 it's an unnecessarily complicated sequence of abstractions (prow > bash script > python script > AWF > KFP pipeline) that ultimately covers [only 4 test cases](https://github.com/kubeflow/pipelines/blob/1.8.22/test/sample_test.yaml#L102-L108) which have dependencies on GCP.
> 
> 
> That being said, I'm not comfortable with the limited large test scope, even if that's the precedent set by 1.8.22, so I'm constructing a bunch of test cases from the samples directory and plan to create a workflow that executes them in parallel using pytest-xdist.
> 
> TBH, I don't understand why so many layers of abstraction were implemented. When I'm done, the resulting approach could be a good reference for the test refactor, @nsingla.
> 
> In general, I think it would be helpful to have a table that describes and differentiates the various workflows. If there isn't automation around that (based on comments in the workflow files) there should be. Might be a fun side project.

I don't have the context about 1.8.22 but that abstraction is definitely not ideal :) 
thanks for removing it

Also, I think that's a great idea to have a table of workflow references, what gets triggered when and what each workflow does. I'll try to ask an AI tool to generate that for me (and hope for the best, otherwise it will have to be a manual work)

.github/workflows/e2e-test.yml
```diff
@@ -19,393 +19,6 @@ on:
       - '!**/OWNERS'
 
 jobs:
-  initialization-tests-v1:
```
droctothorpe, line 22: I removed all non-SDK tests.

.github/workflows/e2e-test.yml
```diff
@@ -456,20 +69,16 @@ jobs:
...
-
-      - name: Basic sample tests - exit_handler
+      
+      - name: Run large tests
         if: ${{ steps.create-kfp-cluster.outcome == 'success' }}
-        id: sample-test
-        run: python3 ./test/sample-test/sample_test_launcher.py sample_test run_test --namespace kubeflow --test-name exit_handler --expected-result failed --results-gcs-dir output
+        id: large-tests
+        run: |
+          python3 -m pytest test/v1-large -n 20
```
droctothorpe, line 77: I switched to using parallelized pytest for large test execution and increased the test count from 2 to 14.

.github/workflows/sdk-component-yaml.yml
```diff
@@ -48,8 +48,8 @@ jobs:
...
           python3 -m pip install api/v2alpha1/python
 
-      - name: Install requirements
-        run: pip install -r ./test/sdk-execution-tests/requirements.txt
+      - name: Install prerequisites
+        run: pip3 install -r ./test/sample-test/requirements.txt && pip install sdk/python
         
       - name: Run component YAML tests
-        run: ./test/presubmit-component-yaml.sh
+        run: $(pwd)/components/test_load_all_components.sh
```
droctothorpe, line 55: Removed some of the superfluous abstraction.

test/sample-test/run_sample_test.py
```diff
@@ -87,7 +87,7 @@ def run(self):
 
         ###### Create Experiment ######
         response = self._client.create_experiment(self._experiment_name)
-        self._experiment_id = response.experiment_id
+        self._experiment_id = response.id
```
droctothorpe, line 90: Updated this to be v1 compatible, but wound up switching to pytest instead of this home-brewed abstraction anyway.

test/v1-large/condition_test.py
```diff
@@ -0,0 +1,41 @@
+from kfp import components
```
droctothorpe, line 1: This `test/v1-large` folder contains 14 core sample files updated to run as proper large tests executed by pytest. These replace the existing e2e test workflow (wrt the SDK) and sample test workflow, which collectively covered less ground despite being far more complex (and redundant).

droctothorpe:
This PR is ready for review. Please note, it's a merge to the `release-kfp-sdk-1.8.23` branch, not master. Merging it won't impact anyone, but I was hoping for some feedback to be on the safe side before I embark on the protobuf and urllib3 upgrades. Thanks!

CC @zazulam @mprahl @HumairAK @cmdevoto @hbelmiro

.github/workflows/e2e-test.yml
nsingla: Would it more sense to use pytest markers and create test groupings that way rather than in a directory? Directory is good as well, but i think markers provide more flexiibility and directory structure can then be based more on functional differences

droctothorpe: I like that directories create logical groupings that are readable at a glance without having to parse the corresponding files / decorators, but I'm open to the alternative! I don't think it's necessary in the context of this PR / branch since we're just retrofitting CI, i.e. these changes will never get merged into master.

test/sample-test/run_sample_test.py
nsingla: I know it's not a pythonic way or a go way of writing code, but imo, its way easier to read code if you define the initializing object type (otherwise, in this case, I who currently does not have full idea of the code, had to trace back and see which object is being returned by `create_experiment` method
Suggested Change:
`response: kfp_server_api.V2beta1Experiment = self._client.create_experiment(self._experiment_name)`

droctothorpe: The way I refactored the workflows means this file is thankfully no longer used in CI (in this branch). Unfortunately, that's not true for the master branch atm. I think any workflows in master that execute this file should be refactored to use pytest instead, but that's unrelated to this PR.

.github/workflows/e2e-test.yml
nsingla: :+1:

test/sample-test/run_sample_test.py
nsingla: I guess, we can discuss this when we have a PR against master :)

Issue #12077, opened 2025-07-25
[feature] Define and use dsl.WORKSPACE_PATH_PLACEHOLDER for workspace access


VaniHaripriya:
### Feature Area

 /area backend
/area sdk

As part of https://github.com/kubeflow/pipelines/issues/12053, we propose enabling automatic workspace volume mounting in the container driver based on a standard placeholder path.

**Summary**

- Add support in the container driver to detect usage of dsl.WORKSPACE_PATH_PLACEHOLDER or paths starting with /kfp-workspace. 
- When detected, the driver mounts the workspace PVC at /kfp-workspace in the Pod spec. 
- This enables components to access a shared volume without explicitly defining volume mounts. 
---

Love this idea? Give it a 👍.

PR #12078, opened 2025-07-25
feat(backend/sdk): define and use dsl.WORKSPACE_PATH_PLACEHOLDER for workspace access

VaniHaripriya:

This resolves https://github.com/kubeflow/pipelines/issues/12077

**Summary**

**SDK Change:**
    Added a new placeholder constant:
    dsl.WORKSPACE_PATH_PLACEHOLDER = "/kfp-workspace"

 **Container Driver Change:**
  - Detects if any input parameter equals dsl.WORKSPACE_PATH_PLACEHOLDER. 
  - Also detects if any input or output path starts with /kfp-workspace. 
  - If matched, the driver injects the volume mount for the shared workspace PVC into the Pod spec.

**Checklist:**

### DIFF
--- a/backend/src/v2/compiler/visitor.go
+++ b/backend/src/v2/compiler/visitor.go
@@ -111,11 +111,13 @@ func (state *pipelineDFS) dfs(name string, component *pipelinespec.ComponentSpec
 
 		// Add kubernetes spec to annotation
 		if state.kubernetesSpec != nil && state.kubernetesSpec.DeploymentSpec != nil {
-			kubernetesExecSpec, ok := state.kubernetesSpec.DeploymentSpec.Executors[executorLabel]
-			if ok {
-				err := state.visitor.AddKubernetesSpec(name, kubernetesExecSpec)
-				if err != nil {
-					return componentError(fmt.Errorf("failed to add Kubernetes spec for %s: %w", name, err))
+			if state.kubernetesSpec.DeploymentSpec.Executors != nil {
+				kubernetesExecSpec, ok := state.kubernetesSpec.DeploymentSpec.Executors[executorLabel]
+				if ok {
+					err := state.visitor.AddKubernetesSpec(name, kubernetesExecSpec)
+					if err != nil {
+						return componentError(fmt.Errorf("failed to add Kubernetes spec for %s: %w", name, err))
+					}
 				}
 			}
 		}
--- a/backend/src/v2/driver/driver.go
+++ b/backend/src/v2/driver/driver.go
@@ -26,6 +26,7 @@ import (
 	"github.com/kubeflow/pipelines/backend/src/v2/metadata"
 	"github.com/kubeflow/pipelines/kubernetes_platform/go/kubernetesplatform"
 	"google.golang.org/protobuf/encoding/protojson"
+	"google.golang.org/protobuf/types/known/structpb"
 	k8score "k8s.io/api/core/v1"
 	k8sres "k8s.io/apimachinery/pkg/api/resource"
 )
@@ -320,7 +321,74 @@ func initPodSpecPatch(
 
 	addModelcarsToPodSpec(executorInput.GetInputs().GetArtifacts(), userEnvVar, podSpec)
 
+	if needsWorkspaceMount(executorInput) {
+		// Validate that no user volume mounts conflict with the workspace
+		if err := validateVolumeMounts(podSpec); err != nil {
+			return nil, fmt.Errorf("failed to validate volume mounts: %w", err)
+		}
+
+		// Uses Argo template variable to reference the PVC created by volumeClaimTemplates
+		// Argo resolves {{workflow.name}}-kfp-workspace to the actual PVC name at runtime
+		pvcName := "{{workflow.name}}-" + component.WorkspaceVolumeName
+
+		addWorkspaceMount(podSpec, pvcName)
+	}
+
 	return podSpec, nil
+}
+
+// needsWorkspaceMount checks if the component needs workspace mounting based on input parameters and artifacts.
+func needsWorkspaceMount(executorInput *pipelinespec.ExecutorInput) bool {
+	// Check if any input parameter is the workspace path placeholder
+	for _, param := range executorInput.GetInputs().GetParameterValues() {
+		if strVal, ok := param.GetKind().(*structpb.Value_StringValue); ok {
+			if strings.Contains(strVal.StringValue, "{{$.workspace_path}}") {
+				return true
+			}
+
+			if strings.HasPrefix(strVal.StringValue, component.WorkspaceMountPath) {
+				return true
+			}
+		}
+	}
+
+	// Check if any input artifact has workspace metadata
+	for _, artifactList := range executorInput.GetInputs().GetArtifacts() {
+		if len(artifactList.Artifacts) == 0 {
+			continue
+		}
+		// first artifact is used, as the list is expected to contain a single artifact
+		artifact := artifactList.Artifacts[0]
+		if artifact.Metadata != nil {
+			if workspaceVal, ok := artifact.Metadata.Fields["_kfp_workspace"]; ok {
+				if boolVal, ok := workspaceVal.GetKind().(*structpb.Value_BoolValue); ok && boolVal.BoolValue {
+					return true
+				}
+			}
+		}
+	}
+
+	return false
+}
+
+// addWorkspaceMount adds the workspace volume mount to the pod spec if needed.
+func addWorkspaceMount(podSpec *k8score.PodSpec, pvcName string) {
+	workspaceVolume := k8score.Volume{
+		Name: component.WorkspaceVolumeName,
+		VolumeSource: k8score.VolumeSource{
+			PersistentVolumeClaim: &k8score.PersistentVolumeClaimVolumeSource{
+				ClaimName: pvcName,
+			},
+		},
+	}
+
+	workspaceVolumeMount := k8score.VolumeMount{
+		Name:      component.WorkspaceVolumeName,
+		MountPath: component.WorkspaceMountPath,
+	}
+
+	podSpec.Volumes = append(podSpec.Volumes, workspaceVolume)
+	podSpec.Containers[0].VolumeMounts = append(podSpec.Containers[0].VolumeMounts, workspaceVolumeMount)
 }
 
 // addModelcarsToPodSpec will patch the pod spec if there are any input artifacts in the Modelcar format.
@@ -524,4 +592,21 @@ func provisionOutputs(
 	}
 
 	return outputs
+}
+
+func validateVolumeMounts(podSpec *k8score.PodSpec) error {
+	// Validate that no user volume mounts conflict with the workspace mount path or volume name
+	for _, container := range podSpec.Containers {
+		for _, mount := range container.VolumeMounts {
+			if strings.HasPrefix(mount.MountPath, component.WorkspaceMountPath) {
+				return fmt.Errorf("user volume mount at %s conflicts with workspace mount at %s", mount.MountPath, component.WorkspaceMountPath)
+			}
+
+			if mount.Name == component.WorkspaceVolumeName {
+				return fmt.Errorf("user volume mount name %s conflicts with workspace volume name %s", mount.Name, component.WorkspaceVolumeName)
+			}
+		}
+	}
+
+	return nil
 }
--- a/backend/src/v2/driver/driver_test.go
+++ b/backend/src/v2/driver/driver_test.go
@@ -725,3 +725,241 @@ func Test_initPodSpecPatch_inputTaskFinalStatus(t *testing.T) {
 	assert.Equal(t, expectedExecutorInput, actualExecutorInput)
 	assert.Equal(t, expectedComponentSpec, actualComponentSpec)
 }
+
+func TestNeedsWorkspaceMount(t *testing.T) {
+	tests := []struct {
+		name          string
+		executorInput *pipelinespec.ExecutorInput
+		expected      bool
+	}{
+		{
+			name: "workspace path placeholder in parameters",
+			executorInput: &pipelinespec.ExecutorInput{
+				Inputs: &pipelinespec.ExecutorInput_Inputs{
+					ParameterValues: map[string]*structpb.Value{
+						"workspace_path": {
+							Kind: &structpb.Value_StringValue{
+								StringValue: "{{$.workspace_path}}",
+							},
+						},
+					},
+				},
+			},
+			expected: true,
+		},
+		{
+			name: "workspace path prefix in parameters",
+			executorInput: &pipelinespec.ExecutorInput{
+				Inputs: &pipelinespec.ExecutorInput_Inputs{
+					ParameterValues: map[string]*structpb.Value{
+						"file_path": {
+							Kind: &structpb.Value_StringValue{
+								StringValue: "/kfp-workspace/data/file.txt",
+							},
+						},
+					},
+				},
+			},
+			expected: true,
+		},
+		{
+			name: "artifact with workspace metadata",
+			executorInput: &pipelinespec.ExecutorInput{
+				Inputs: &pipelinespec.ExecutorInput_Inputs{
+					Artifacts: map[string]*pipelinespec.ArtifactList{
+						"model": {
+							Artifacts: []*pipelinespec.RuntimeArtifact{
+								{
+									Metadata: &structpb.Struct{
+										Fields: map[string]*structpb.Value{
+											"_kfp_workspace": {
+												Kind: &structpb.Value_BoolValue{
+													BoolValue: true,
+												},
+											},
+										},
+									},
+								},
+							},
+						},
+					},
+				},
+			},
+			expected: true,
+		},
+		{
+			name: "workspace path with subdirectory",
+			executorInput: &pipelinespec.ExecutorInput{
+				Inputs: &pipelinespec.ExecutorInput_Inputs{
+					ParameterValues: map[string]*structpb.Value{
+						"file_path": {
+							Kind: &structpb.Value_StringValue{
+								StringValue: "{{$.workspace_path}}/data",
+							},
+						},
+					},
+				},
+			},
+			expected: true,
+		},
+		{
+			name: "no workspace usage",
+			executorInput: &pipelinespec.ExecutorInput{
+				Inputs: &pipelinespec.ExecutorInput_Inputs{
+					ParameterValues: map[string]*structpb.Value{
+						"text": {
+							Kind: &structpb.Value_StringValue{
+								StringValue: "hello world",
+							},
+						},
+					},
+				},
+			},
+			expected: false,
+		},
+	}
+
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			result := needsWorkspaceMount(tt.executorInput)
+			if result != tt.expected {
+				t.Errorf("needsWorkspaceMount() = %v, want %v", result, tt.expected)
+			}
+		})
+	}
+}
+
+func TestAddWorkspaceMount(t *testing.T) {
+	podSpec := &k8score.PodSpec{
+		Containers: []k8score.Container{
+			{
+				Name: "main",
+			},
+		},
+	}
+
+	pvcName := "test-workflow-kfp-workspace"
+
+	addWorkspaceMount(podSpec, pvcName)
+
+	// Check that volume was added
+	if len(podSpec.Volumes) != 1 {
+		t.Errorf("Expected 1 volume, got %d", len(podSpec.Volumes))
+	}
+
+	volume := podSpec.Volumes[0]
+	if volume.Name != "kfp-workspace" {
+		t.Errorf("Expected volume name kfp-workspace, got %s", volume.Name)
+	}
+
+	if volume.PersistentVolumeClaim == nil {
+		t.Error("Expected PersistentVolumeClaim to be set")
+	}
+
+	if volume.PersistentVolumeClaim.ClaimName != pvcName {
+		t.Errorf("Expected claim name %s, got %s", pvcName, volume.PersistentVolumeClaim.ClaimName)
+	}
+
+	// Check that volume mount was added
+	if len(podSpec.Containers[0].VolumeMounts) != 1 {
+		t.Errorf("Expected 1 volume mount, got %d", len(podSpec.Containers[0].VolumeMounts))
+	}
+
+	volumeMount := podSpec.Containers[0].VolumeMounts[0]
+	if volumeMount.Name != "kfp-workspace" {
+		t.Errorf("Expected volume mount name kfp-workspace, got %s", volumeMount.Name)
+	}
+
+	if volumeMount.MountPath != "/kfp-workspace" {
+		t.Errorf("Expected mount path /kfp-workspace, got %s", volumeMount.MountPath)
+	}
+}
+
+func TestValidateVolumeMounts(t *testing.T) {
+	tests := []struct {
+		name        string
+		podSpec     *k8score.PodSpec
+		expectError bool
+	}{
+		{
+			name: "no conflicting volume mounts",
+			podSpec: &k8score.PodSpec{
+				Containers: []k8score.Container{
+					{
+						Name: "main",
+						VolumeMounts: []k8score.VolumeMount{
+							{
+								Name:      "data",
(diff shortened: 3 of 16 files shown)


backend/src/v2/compiler/visitor.go
```diff
@@ -111,11 +111,13 @@ func (state *pipelineDFS) dfs(name string, component *pipelinespec.ComponentSpec
 
 		// Add kubernetes spec to annotation
 		if state.kubernetesSpec != nil && state.kubernetesSpec.DeploymentSpec != nil {
-			kubernetesExecSpec, ok := state.kubernetesSpec.DeploymentSpec.Executors[executorLabel]
-			if ok {
-				err := state.visitor.AddKubernetesSpec(name, kubernetesExecSpec)
-				if err != nil {
-					return componentError(fmt.Errorf("failed to add Kubernetes spec for %s: %w", name, err))
+			if state.kubernetesSpec.DeploymentSpec.Executors != nil {
```
mprahl, line 114: I don't think this is needed since nil maps in Go can still be accessed safely but it's okay to have it here too.

backend/src/v2/component/launcher_v2.go
```diff
@@ -820,6 +820,9 @@ func getPlaceholders(executorInput *pipelinespec.ExecutorInput) (placeholders ma
 		placeholders[key] = parameter.OutputFile
 	}
 
+	// Add workspace path placeholder
+	placeholders["{{$.workspace_path}}"] = WorkspaceMountPath
```
mprahl: I was originally thinking that this logic would live in the driver and be another switch statement here:
https://github.com/kubeflow/pipelines/blob/a870b1a325dae0c82c8b6f57941468ee1aea960b/backend/src/v2/driver/resolve.go#L387-L406

I think your approach of using `getPlaceholders` may enable more scenarios so that any reference to `{{$.workspace_path}}` will be replaced, not just input parameters. So I think it's better than what I had in mind. :smile:

backend/src/v2/driver/driver.go
```diff
@@ -320,9 +321,80 @@ func initPodSpecPatch(
...
+	}
+
+	// Check if any input parameter is a component output path starting with /kfp-workspace
+	for _, param := range executorInput.GetInputs().GetParameterValues() {
+		if strVal, ok := param.GetKind().(*structpb.Value_StringValue); ok {
+			if strings.HasPrefix(strVal.StringValue, component.WorkspaceMountPath) {
+				return true
+			}
+		}
+	}
```
mprahl:
```suggestion
	for _, param := range executorInput.GetInputs().GetParameterValues() {
		if strVal, ok := param.GetKind().(*structpb.Value_StringValue); ok {
			if strVal.StringValue == "{{$.workspace_path}}" {
				return true
			}

			if strings.HasPrefix(strVal.StringValue, component.WorkspaceMountPath) {
				return true
			}
		}
	}
```

backend/src/v2/driver/driver.go
```diff
@@ -320,9 +321,80 @@ func initPodSpecPatch(
...
 
+	if needsWorkspaceMount(executorInput) {
+		// Validate that no user volume mounts conflict with the workspace
+		if err := validateVolumeMounts(podSpec); err != nil {
+			return nil, fmt.Errorf("failed to validate volume mounts: %w", err)
+		}
+
+		// Uses Argo template variable to reference the PVC created by volumeClaimTemplates
+		// Argo resolves {{workflow.name}}-kfp-workspace to the actual PVC name at runtime
+		pvcName := "{{workflow.name}}-" + component.WorkspaceVolumeName
```
mprahl, line 332: Could you please update the design document to match the implementation for the PVC name? For example:

```diff
diff --git a/proposals/11875-pipeline-workspace/README.md b/proposals/11875-pipeline-workspace/README.md
index 1c355aabc..41803b4ef 100644
--- a/proposals/11875-pipeline-workspace/README.md
+++ b/proposals/11875-pipeline-workspace/README.md
@@ -509,7 +509,7 @@ kind: Workflow
 spec:
   volumeClaimTemplates:
     - metadata:
-        name: kfp-workspace-46f1d52e-c72b-42fc-88ae-789edf7c33fd # The suffix is the run ID
+        name: kfp-workspace
       spec:
         accessModes:
           - ReadWriteMany
```

backend/src/v2/driver/driver.go
```diff
@@ -525,3 +597,17 @@ func provisionOutputs(
 
 	return outputs
 }
+
+func validateVolumeMounts(podSpec *k8score.PodSpec) error {
+	// Check for user-defined volume mounts that conflict with the workspace path
+	for _, container := range podSpec.Containers {
+		for _, mount := range container.VolumeMounts {
+			// Check if this mount path conflicts with the workspace mount path
+			if strings.HasPrefix(mount.MountPath, component.WorkspaceMountPath) {
```
mprahl, line 601: Could you please also check that the mount name doesn't conflict with `WorkspaceVolumeName` as well?

backend/src/v2/driver/resolve.go
```diff
@@ -401,6 +402,8 @@ func resolveInputParameter(
 				v = structpb.NewStringValue(opts.TaskName)
 			case "{{$.pipeline_task_uuid}}":
 				v = structpb.NewStringValue(fmt.Sprintf("%d", opts.DAGExecutionID))
+			case "{{$.workspace_path}}":
+				v = structpb.NewStringValue(component.WorkspaceMountPath)
```
mprahl: I now see you implemented both options (related to https://github.com/kubeflow/pipelines/pull/12078/files#r2242652605). Is there a case where `getPlaceholders` doesn't cover cases here? I think `getPlaceholders` will likely be sufficient.

sdk/python/kfp/local/placeholder_utils.py
```diff
@@ -347,6 +348,8 @@ def resolve_individual_placeholder(
             pipeline_task_id,
         dsl.PIPELINE_ROOT_PLACEHOLDER:
             pipeline_root,
+        dsl.WORKSPACE_PATH_PLACEHOLDER:
+            constants.WORKSPACE_MOUNT_PATH,
```
mprahl: This is for KFP local and we didn't exactly design for it so we can either raise an exception as not implemented, or you could implement it similar to this (untested):

```diff
diff --git a/sdk/python/kfp/local/config.py b/sdk/python/kfp/local/config.py
index 101529f97..c3e59d89e 100755
--- a/sdk/python/kfp/local/config.py
+++ b/sdk/python/kfp/local/config.py
@@ -15,7 +15,8 @@
 import abc
 import dataclasses
 import os
-from typing import Union
+import tempfile
+from typing import Optional, Union
 
 from kfp import local
 
@@ -191,6 +192,7 @@ class LocalExecutionConfig:
         cls,
         runner: SubprocessRunner,
         pipeline_root: str,
+        workspace_root: str,
         raise_on_error: bool,
     ) -> 'LocalExecutionConfig':
         # singleton pattern
@@ -201,6 +203,7 @@ class LocalExecutionConfig:
         self,
         runner: SubprocessRunner,
         pipeline_root: str,
+        workspace_root: str,
         raise_on_error: bool,
     ) -> None:
         permitted_runners = (SubprocessRunner, DockerRunner)
@@ -210,6 +213,7 @@ class LocalExecutionConfig:
             )
         self.runner = runner
         self.pipeline_root = pipeline_root
+        self.workspace_root = workspace_root
         self.raise_on_error = raise_on_error
 
     @classmethod
@@ -224,6 +228,7 @@ def init(
     # annotate with subclasses, not parent class, for more helpful ref docs
     runner: Union[SubprocessRunner, DockerRunner],
     pipeline_root: str = './local_outputs',
+    workspace_root: Optional[str] = None,
     raise_on_error: bool = True,
 ) -> None:
     """Initializes a local execution session.
@@ -237,8 +242,12 @@ def init(
     """
     # updates a global config
     pipeline_root = os.path.abspath(pipeline_root)
+    if workspace_root is None:
+        workspace_root = tempfile.mkdtemp(prefix='kfp-workspace-')
+
     LocalExecutionConfig(
         runner=runner,
         pipeline_root=pipeline_root,
+        workspace_root=workspace_root,
         raise_on_error=raise_on_error,
     )
diff --git a/sdk/python/kfp/local/pipeline_orchestrator.py b/sdk/python/kfp/local/pipeline_orchestrator.py
index 117673688..78698f8a7 100644
--- a/sdk/python/kfp/local/pipeline_orchestrator.py
+++ b/sdk/python/kfp/local/pipeline_orchestrator.py
@@ -13,6 +13,7 @@
 # limitations under the License.
 """Code for locally executing a compiled pipeline."""
 import logging
+import shutil
 from typing import Any, Dict, List
 
 from kfp.local import config
@@ -86,17 +87,21 @@ def _run_local_pipeline_implementation(
     # convert to dict for consistency with executors
     components = dict(pipeline_spec.components.items())
     fail_stack: List[str] = []
-    outputs, dag_status = dag_orchestrator.run_dag(
-        pipeline_resource_name=pipeline_resource_name,
-        dag_component_spec=pipeline_spec.root,
-        executors=executors,
-        components=components,
-        dag_arguments=arguments,
-        pipeline_root=pipeline_root,
-        runner=runner,
-        unique_pipeline_id=placeholder_utils.make_random_id(),
-        fail_stack=fail_stack,
-    )
+    try:
+        outputs, dag_status = dag_orchestrator.run_dag(
+            pipeline_resource_name=pipeline_resource_name,
+            dag_component_spec=pipeline_spec.root,
+            executors=executors,
+            components=components,
+            dag_arguments=arguments,
+            pipeline_root=pipeline_root,
+            runner=runner,
+            unique_pipeline_id=placeholder_utils.make_random_id(),
+            fail_stack=fail_stack,
+        )
+    finally:
+        shutil.rmtree(config.LocalExecutionConfig.instance.workspace_root)
+
     if dag_status == status.Status.SUCCESS:
         status_with_color = logging_utils.format_status(status.Status.SUCCESS)
         with logging_utils.local_logger_context():
diff --git a/sdk/python/kfp/local/placeholder_utils.py b/sdk/python/kfp/local/placeholder_utils.py
index 405b48e85..a9d4ba064 100644
--- a/sdk/python/kfp/local/placeholder_utils.py
+++ b/sdk/python/kfp/local/placeholder_utils.py
@@ -19,6 +19,7 @@ import re
 from typing import Any, Dict, List, Optional, Union
 
 from kfp import dsl
+from kfp.local import config
 
 
 def make_random_id() -> str:
@@ -347,6 +348,8 @@ def resolve_individual_placeholder(
             pipeline_task_id,
         dsl.PIPELINE_ROOT_PLACEHOLDER:
             pipeline_root,
+        dsl.WORKSPACE_ROOT_PLACEHOLDER:
+            config.LocalExecutionConfig.instance.workspace_root,
     }
     for placeholder, value in PLACEHOLDERS.items():
         element = element.replace(placeholder, value)
```

HumairAK: I think we should definitely add local support as part of this, this lgtm - @VaniHaripriya can you test this out and see if it gets the intended result?

backend/src/v2/driver/resolve.go
mprahl: @VaniHaripriya I think we can go ahead and just use this instead of `getPlaceholders` but we'll need to change this logic so that it uses logic to see if the string contains `{{$.workspace_path}}` and replace all those rather than checking for an exact match of the string `{{$.workspace_path}}`. This would support cases of like `os.path.join(dsl.WORKSPACE_PATH_PLACEHOLDER, "my-path")`.

backend/src/v2/compiler/visitor.go
VaniHaripriya: I added the nil check as I encountered a panic during testing. I believe the root cause is that Executors is a protobuf-generated map (map[string]*structpb.Struct).

backend/src/v2/driver/resolve.go
VaniHaripriya: Thank you @mprahl , that makes sense. I'll implement the changes as suggested.

backend/src/v2/driver/driver.go
```diff
@@ -320,9 +321,75 @@ func initPodSpecPatch(
...
+			}
+		}
+	}
+
+	// Check if any input artifact has workspace metadata
+	for _, artifactList := range executorInput.GetInputs().GetArtifacts() {
+		if len(artifactList.Artifacts) == 0 {
+			continue
+		}
+		artifact := artifactList.Artifacts[0]
```
alyssacgoins, line 361: For someone with less knowledge of the executorInput structure, `artifact := artifactList.Artifacts[0]` could read a little confusingly - could be worth adding a comment explaining why it's indexed to the first entry

backend/src/v2/driver/driver.go
```diff
@@ -320,9 +321,76 @@ func initPodSpecPatch(
...
 }
 
+// needsWorkspaceMount checks if the component needs workspace mounting based on input parameters and artifacts.
+func needsWorkspaceMount(executorInput *pipelinespec.ExecutorInput) bool {
+	// Check if any input parameter is the workspace path placeholder
+	for _, param := range executorInput.GetInputs().GetParameterValues() {
+		if strVal, ok := param.GetKind().(*structpb.Value_StringValue); ok {
+			if strVal.StringValue == "{{$.workspace_path}}" {
+				return true
+			}
```
mprahl:
```suggestion
			if strings.Contains(strVal.StringValue, "{{$.workspace_path}}") {
				return true
			}
```

backend/src/v2/driver/driver_test.go
```diff
@@ -725,3 +725,209 @@ func Test_initPodSpecPatch_inputTaskFinalStatus(t *testing.T) {
...
+	if volumeMount.Name != "kfp-workspace" {
+		t.Errorf("Expected volume mount name kfp-workspace, got %s", volumeMount.Name)
+	}
+
+	if volumeMount.MountPath != "/kfp-workspace" {
+		t.Errorf("Expected mount path /kfp-workspace, got %s", volumeMount.MountPath)
+	}
+}
+
+func TestValidateVolumeMounts(t *testing.T) {
```
mprahl, line 878: Could you add a test case for the conflicting volume name of `kfp-workspace`?

backend/src/v2/driver/driver_test.go
```diff
@@ -725,3 +725,209 @@ func Test_initPodSpecPatch_inputTaskFinalStatus(t *testing.T) {
...
 }
+
+func TestNeedsWorkspaceMount(t *testing.T) {
+	tests := []struct {
+		name          string
+		executorInput *pipelinespec.ExecutorInput
+		expected      bool
+	}{
+		{
+			name: "workspace path placeholder in parameters",
```
mprahl, line 736: Could you add a test case for `"{{$.workspace_path}}/subdir"` as well?

backend/src/v2/driver/resolve.go
```diff
@@ -389,20 +391,26 @@ func resolveInputParameter(
...
-				v = structpb.NewStringValue(opts.RunID)
-			case "{{$.pipeline_task_name}}":
-				v = structpb.NewStringValue(opts.TaskName)
-			case "{{$.pipeline_task_uuid}}":
-				v = structpb.NewStringValue(fmt.Sprintf("%d", opts.DAGExecutionID))
-			default:
-				v = val
+
+			if strings.Contains(valStr, "{{$.workspace_path}}") {
+				v = structpb.NewStringValue(strings.ReplaceAll(valStr, "{{$.workspace_path}}", component.WorkspaceMountPath))
```
mprahl, line 398: Optional: You can return here and avoid needing `else`

sdk/python/kfp/local/placeholder_utils_test.py
```diff
@@ -458,5 +458,60 @@ def test_simple(self):
...
+        
+        self.assertIsInstance(result, str)
+        self.assertTrue(result.startswith('/tmp/kfp-workspace-'))
+
+    def test_embedded_workspace_placeholder(self):
+        """Test embedded workspace placeholder resolution."""
+        executor_input_dict = {
+            'inputs': {
+                'parameterValues': {
+                    'file_path': "os.path.join('{{$.workspace_path}}', 'data', 'file.txt')"
```
mprahl: Did you mean this instead?
```suggestion
                    'file_path': os.path.join('{{$.workspace_path}}', 'data', 'file.txt')
```

VaniHaripriya: Yes :slightly_smiling_face:  that 's supposed be with out quotes.

sdk/python/kfp/local/placeholder_utils.py
VaniHaripriya: This has been tested locally and works as expected.

mprahl:
@VaniHaripriya sorry for not mentioning it previously, but could you please also add a simple E2E test? You can just write a simple pipeline and add it here as a sample test:
https://github.com/kubeflow/pipelines/blob/master/samples/v2/sample_test.py

The pipeline can be two components where they both accept `dsl.WORKSPACE_PATH_PLACEHOLDER` and one writes a file to the workspace and the other component reads from it (don't forget the `.after` call to make it execute in the proper order).

Future updates to the workspace feature can just add to that test.

sdk/python/kfp/local/placeholder_utils.py
mprahl: Cool! I tried your branch locally too and KFP local works nicely.

samples/v2/pipeline_with_workspace.py
```diff
@@ -0,0 +1,81 @@
...
+    """A pipeline using workspace functionality with write and read components."""
+    
+    # Write to workspace
+    write_task = write_to_workspace(
+        workspace_path=dsl.WORKSPACE_PATH_PLACEHOLDER
+    )
+    
+    # Read from workspace
+    read_task = read_from_workspace(
+        workspace_path=dsl.WORKSPACE_PATH_PLACEHOLDER
```
mprahl: Could you make this task take `write_task.output` as input and then `read_task` can just read the input path directly? This adds test coverage and then you can remove the `read_task.after` call.

sdk/python/kfp/dsl/placeholders_test.py
```diff
@@ -562,3 +562,15 @@ def container_with_placeholder_in_fstring(
                         'echo',
                         f"another-prefix-{dsl.IfPresentPlaceholder(input_name='text1', then=['val'])}",
                     ])
+
+
+class TestWorkspacePathPlaceholder(parameterized.TestCase):
```
mprahl: I think this test case can be deleted since it's essentially just testing `WORKSPACE_PATH_PLACEHOLDER` is the value that it was set to.

samples/v2/pipeline_with_workspace.py
```diff
@@ -0,0 +1,81 @@
...
+    """Read a file from the workspace."""    
+    import os
+    
+    file_path = os.path.join(workspace_path, "data", "test_file.txt")
+    
+    if os.path.exists(file_path):
+        with open(file_path, "r") as f:
+            content = f.read()
+        print(f"Read content from: {file_path}")
+        print(f"Content: {content}")
```
mprahl, line 44: Could you add an `assert` call to ensure the content is equal to `"Hello from workspace!"`?

sdk/python/kfp/local/config.py
```diff
@@ -15,10 +15,14 @@
...
 import dataclasses
 import os
-from typing import Union
+import tempfile
+from typing import Optional, Union
 
 from kfp import local
 
+# Default workspace path when no custom path is provided
+DEFAULT_WORKSPACE_PATH = '/tmp/kfp-workspace-default'
```
mprahl: I think rather than a default, could we just generate a temporary directory with `kfp-workspace-` as the prefix? Otherwise we can't run multiple local pipelines at the same time that use a workspace.

sdk/python/kfp/local/pipeline_orchestrator.py
```diff
@@ -13,6 +13,8 @@
 # limitations under the License.
 """Code for locally executing a compiled pipeline."""
 import logging
+import os
+import shutil
```
mprahl, line 17: Could you add the following unit test or similar for KFP local?

```diff
diff --git a/sdk/python/kfp/local/pipeline_orchestrator_test.py b/sdk/python/kfp/local/pipeline_orchestrator_test.py
index 1e8703761..84b0757c8 100644
--- a/sdk/python/kfp/local/pipeline_orchestrator_test.py
+++ b/sdk/python/kfp/local/pipeline_orchestrator_test.py
@@ -677,6 +677,50 @@ class TestRunLocalPipeline(testing_utilities.LocalRunnerEnvironmentTestCase):
         task = my_pipeline()
         self.assertEqual(task.output, 'foo-bar-baz')
 
+    def test_workspace_functionality(self):
+        import tempfile
+        
+        # Create temporary directory for workspace
+        with tempfile.TemporaryDirectory() as temp_dir:
+            workspace_root = os.path.join(temp_dir, 'workspace')
+            os.makedirs(workspace_root, exist_ok=True)
+            
+            local.init(
+                local.SubprocessRunner(), 
+                pipeline_root=ROOT_FOR_TESTING,
+                workspace_root=workspace_root
+            )
+
+            @dsl.component
+            def write_to_workspace(text: str, workspace_path: str) -> str:
+                import os
+                output_file = os.path.join(workspace_path, 'output.txt')
+                with open(output_file, 'w') as f:
+                    f.write(text)
+                return output_file
+
+            @dsl.component
+            def read_from_workspace(file_path: str) -> str:
+                with open(file_path, 'r') as f:
+                    return f.read()
+
+            @dsl.pipeline
+            def my_pipeline(text: str = 'Hello workspace!') -> str:
+                # Write to workspace
+                write_task = write_to_workspace(
+                    text=text,
+                    workspace_path=dsl.WORKSPACE_PATH_PLACEHOLDER
+                )
+                
+                # Read from workspace
+                read_task = read_from_workspace(file_path=write_task.output)
+                
+                return read_task.output
+
+            task = my_pipeline(text='Test workspace functionality!')
+            self.assertEqual(task.output, 'Test workspace functionality!')
+            self.assert_output_dir_contents(1, 2)
+
 
 class TestFstringContainerComponent(
         testing_utilities.LocalRunnerEnvironmentTestCase):
```

sdk/python/kfp/local/placeholder_utils.py
```diff
@@ -347,6 +350,10 @@ def resolve_individual_placeholder(
             pipeline_task_id,
         dsl.PIPELINE_ROOT_PLACEHOLDER:
             pipeline_root,
+        dsl.WORKSPACE_PATH_PLACEHOLDER:
+            config.LocalExecutionConfig.instance.workspace_root
+            if config.LocalExecutionConfig.instance else tempfile.mkdtemp(
```
mprahl: Could you raise an exception if `config.LocalExecutionConfig.instance.workspace_root` isn't set rather than defaulting to a temp dir here? The reason is that we only want the default to be set at the beginning of the local pipeline run. Having it here causes a different temporary directory to be created and used for each pipeline step using the workspace.

sdk/python/kfp/local/config.py
```diff
@@ -233,12 +238,17 @@ def init(
     Args:
         runner: The runner to use. Supported runners: kfp.local.SubprocessRunner and kfp.local.DockerRunner.
         pipeline_root: Destination for task outputs.
+        workspace_root: Directory to use as workspace. If None, a temporary directory will be created.
         raise_on_error: If True, raises an exception when a local task execution fails. If False, fails gracefully and does not terminate the current program.
     """
     # updates a global config
     pipeline_root = os.path.abspath(pipeline_root)
+    if workspace_root is None:
```
mprahl, line 246: I just realized we forgot about `DockerRunner`. Could you raise an exception here if `workspace_root` is set and avoid setting a default value if the runner is `DockerRunner`? Then once we merge this PR, could you please create a follow up GitHub/Jira issue to implement this separately?

sdk/python/kfp/local/importer_handler_test.py
```diff
@@ -24,6 +24,13 @@
 
 class TestRunImporter(testing_utilities.LocalRunnerEnvironmentTestCase):
 
+    def setUp(self):
+        """Set up test environment with workspace configured."""
+        super().setUp()
+        # Initialize local environment with workspace for importer tests
+        from kfp import local
+        local.init(runner=local.SubprocessRunner())
+
```
mprahl:
```suggestion
```

PR #9886, opened 2023-08-16
feat(sdk): enable dependency-free runtime install of kfp

connor-mccarthy:
**Description of your changes:**
Compiles Python components and structures the KFP SDK such that the KFP SDK will require no additional runtime dependencies (excluding the standard library backport of `typing` via `typing-extensions` for components running Python 3.7 and Python 3.8).

This:
- improves task startup latency
- reduces errors associated with dependency conflicts between the KFP SDK and user code
- reduces vulnerability to runtime bugs introduced by transitive dependencies

**Checklist:**
- [x] The title for your pull request (PR) should follow our title convention. [Learn more about the pull request title convention 
used in this repository](https://github.com/kubeflow/pipelines/blob/master/CONTRIBUTING.md#pull-request-title-convention).

PR #12081, opened 2025-07-28
fix(sdk): Configurable component compile time imports

cbartram:
Add a configurable compile time import flag for kfp components. When attempting to import a compile time dependency at runtime the following error is thrown: `ImportError cannot import name `PipelineTask` from `kfp.dsl`.

See the following example pipeline which replicates this issue: 

```python
import kfp
from kfp.dsl import component, pipeline

@component(
	base_image="python:3.9-slim"
)
def hello_world() -> str:
        # Compile time import used at runtime
	from kfp.kubernetes import node_selector

	return "Hello, World from Kubeflow Pipelines!"


@pipeline(
	name="hello-world-pipeline",
	description="A simple hello world pipeline for KFP v2"
)
def hello_world_pipeline():
	"""Define the pipeline workflow."""
	hello_task = hello_world()


if __name__ == "__main__":
	kfp.compiler.Compiler().compile(
		pipeline_func=hello_world_pipeline,
		package_path="hello_world_pipeline.yaml"
	)

```

This change makes the `_KFP_RUNTIME=true` portion of the IR spec configurable so that compile time imports can be used at runtime if desired. The `compile_time_imports` flag on the component decorator defaults to false so it won't disrupt any existing pipeline configurations or change existing functionality. It's purely opt-in. Here is an example utilizing it with a component:

```python
import kfp
from kfp.dsl import component, pipeline

@component(
	base_image="python:3.9-slim" ,
        compile_time_imports=True
)
def hello_world() -> str:
        # Now this works as expected
	from kfp.kubernetes import node_selector

	return "Hello, World from Kubeflow Pipelines!"
```

**Checklist:**

### DIFF
--- a/sdk/python/kfp/dsl/component_decorator.py
+++ b/sdk/python/kfp/dsl/component_decorator.py
@@ -29,7 +29,8 @@ def component(func: Optional[Callable] = None,
               install_kfp_package: bool = True,
               kfp_package_path: Optional[str] = None,
               pip_trusted_hosts: Optional[List[str]] = None,
-              use_venv: bool = False):
+              use_venv: bool = False,
+              compile_time_imports: bool = False):
     """Decorator for Python-function based components.
 
     A KFP component can either be a lightweight component or a containerized
@@ -79,6 +80,8 @@ def component(func: Optional[Callable] = None,
         use_venv: Specifies if the component should be executed in a virtual environment.
             The environment will be created in a temporary directory and will inherit the system site packages.
             This is useful in restricted environments where most of the system is read-only.
+        compile_time_imports: When true this will include imports in the kfp.dsl package traditionally used only at
+        compile time. Set this value to true when you need compile time imports at runtime. By default, this is false because skipping compile time imports improves component execution times.
 
     Returns:
         A component task factory that can be used in pipeline definitions.
@@ -121,7 +124,9 @@ def component(func: Optional[Callable] = None,
             install_kfp_package=install_kfp_package,
             kfp_package_path=kfp_package_path,
             pip_trusted_hosts=pip_trusted_hosts,
-            use_venv=use_venv)
+            use_venv=use_venv,
+            compile_time_imports=compile_time_imports
+        )
 
     return component_factory.create_component_from_func(
         func,
@@ -133,4 +138,6 @@ def component(func: Optional[Callable] = None,
         install_kfp_package=install_kfp_package,
         kfp_package_path=kfp_package_path,
         pip_trusted_hosts=pip_trusted_hosts,
-        use_venv=use_venv)
+        use_venv=use_venv,
+        compile_time_imports=compile_time_imports,
+    )
--- a/sdk/python/kfp/dsl/component_decorator_test.py
+++ b/sdk/python/kfp/dsl/component_decorator_test.py
@@ -149,3 +149,21 @@ class TestComponentDecorator(unittest.TestCase):
         # TODO: ideally should be the canonical type string, rather than the specific annotation as string, but both work
         self.assertEqual(comp.component_spec.outputs['Output'].type,
                          'typing.List[str]')
+
+    def test_with_compile_time_imports(self):
+
+        @component(base_image='python:3.9', compile_time_imports=True)
+        def hello_world(text: str) -> str:
+            return text
+
+        self.assertIsInstance(hello_world, python_component.PythonComponent)
+        self.assertTrue("_KFP_RUNTIME=false" in hello_world.component_spec.implementation.container.command[5])
+
+    def test_with_compile_time_imports_defaults_false(self):
+
+        @component(base_image='python:3.9')
+        def hello_world(text: str) -> str:
+            return text
+
+        self.assertIsInstance(hello_world, python_component.PythonComponent)
+        self.assertTrue("_KFP_RUNTIME=true" in hello_world.component_spec.implementation.container.command[5])
--- a/sdk/python/kfp/dsl/component_factory.py
+++ b/sdk/python/kfp/dsl/component_factory.py
@@ -486,7 +486,7 @@ CONTAINERIZED_PYTHON_COMPONENT_COMMAND = [
 
 
 def _get_command_and_args_for_lightweight_component(
-        func: Callable) -> Tuple[List[str], List[str]]:
+        func: Callable, include_compile_time_imports: bool = False) -> Tuple[List[str], List[str]]:
     imports_source = [
         'import kfp',
         'from kfp import dsl',
@@ -507,7 +507,7 @@ def _get_command_and_args_for_lightweight_component(
                     program_path=$(mktemp -d)
 
                     printf "%s" "$0" > "$program_path/ephemeral_component.py"
-                    _KFP_RUNTIME=true python3 -m {EXECUTOR_MODULE} \
+                    _KFP_RUNTIME={'false' if include_compile_time_imports else 'true'} python3 -m {EXECUTOR_MODULE} \
                         --component_module_path \
                         "$program_path/ephemeral_component.py" \
                         "$@"
@@ -548,6 +548,7 @@ def create_component_from_func(
     kfp_package_path: Optional[str] = None,
     pip_trusted_hosts: Optional[List[str]] = None,
     use_venv: bool = False,
+    compile_time_imports: bool = False,
 ) -> python_component.PythonComponent:
     """Implementation for the @component decorator.
 
@@ -584,7 +585,7 @@ def create_component_from_func(
             function_name=func.__name__,)
     else:
         command, args = _get_command_and_args_for_lightweight_component(
-            func=func)
+            func=func, include_compile_time_imports=compile_time_imports)
 
     component_spec = extract_component_interface(func)
     component_spec.implementation = structures.Implementation(
--- a/sdk/python/kfp/dsl/component_factory_test.py
+++ b/sdk/python/kfp/dsl/component_factory_test.py
@@ -337,5 +337,21 @@ class TestPythonEOLWarning(unittest.TestCase):
                 pass
 
 
+class TestCompileTimeImports(unittest.TestCase):
+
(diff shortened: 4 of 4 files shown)


HumairAK:
What is the use case for trying to use  `kfp.kubernetes` in @component? this functionality is intended for the @pipeline decorator - if this is not clear we should address this in documentation.

There is, I think a broader issue your example hints at, which is that a new user would expect imports to be usable in @components when they are declared in the global scope. I think the proper way to support this is to throw an error when users try to use packages that have not been imported at the @component scope, so they know to import them per @component function (and install via packages_to_install, or include in the base image). I believe there was an attempt at this in the past see [here](https://github.com/kubeflow/pipelines/pull/11893).

cbartram:
Hey @HumairAK the use case for this PR is because internally we wrap Spark configurations which allow for cpu and memory requests for the driver and executor pods to be passed through to kubernetes requests and limits. We do this all at runtime both validating the cpu/mem configs and updating the spec's with `set_memory_request` and `set_memory_limit`. 

The same thing applies for updating Spark kubernetes manifests with tolerations for GPU's in a pipeline using `add_toleration` 

For pipeline tasks within the `@pipeline` decorator this works great however, for components which create other pods like Spark executors there is a module import chain that eventually breaks because `kfp.dsl.__init__` does not load all the compile time modules at runtime. 

I believe we can restructure our packages internally to move around this but probably not without duplicating code!

cbartram:
Here is the compile time module load analysis. Summary: Compile-time dependencies add 217.48ms on average. I'd be happy to just remove the condition from `kdp.dsl.__init__` if you feel like the increase in load times is okay. 

KFP DSL Import Time Benchmark - Comprehensive Analysis
Iteration 1/15... Runtime: 36.13ms, Full: 1750.34ms, Overhead: 1714.21ms
Iteration 2/15... Runtime: 4.84ms, Full: 69.32ms, Overhead: 64.48ms
Iteration 3/15... Runtime: 4.57ms, Full: 69.86ms, Overhead: 65.30ms
Iteration 4/15... Runtime: 4.63ms, Full: 97.56ms, Overhead: 92.94ms
Iteration 5/15... Runtime: 10.24ms, Full: 76.67ms, Overhead: 66.43ms
Iteration 6/15... Runtime: 4.78ms, Full: 68.09ms, Overhead: 63.31ms
Iteration 7/15... Runtime: 4.77ms, Full: 202.21ms, Overhead: 197.43ms
Iteration 8/15... Runtime: 17.36ms, Full: 189.75ms, Overhead: 172.40ms
Iteration 9/15... Runtime: 10.04ms, Full: 184.41ms, Overhead: 174.37ms
Iteration 10/15... Runtime: 13.45ms, Full: 116.60ms, Overhead: 103.16ms
Iteration 11/15... Runtime: 9.68ms, Full: 128.75ms, Overhead: 119.07ms
Iteration 12/15... Runtime: 5.93ms, Full: 133.52ms, Overhead: 127.59ms
Iteration 13/15... Runtime: 13.94ms, Full: 128.50ms, Overhead: 114.56ms
Iteration 14/15... Runtime: 5.35ms, Full: 93.83ms, Overhead: 88.49ms
Iteration 15/15... Runtime: 7.91ms, Full: 106.40ms, Overhead: 98.48ms

### STATISTICAL ANALYSIS

Runtime-only Import Statistics:
  Mean:     10.24 ms
  Median:   7.91 ms
  Min:      4.57 ms
  Max:      36.13 ms
  Std Dev:  8.22 ms
  CV:       80.3%

Full Import (Runtime + Compile-time) Statistics:
  Mean:     227.72 ms
  Median:   116.60 ms
  Min:      68.09 ms
  Max:      1750.34 ms
  Std Dev:  423.51 ms
  CV:       186.0%

Compile-time Overhead Statistics:
  Mean:     217.48 ms
  Median:   103.16 ms
  Min:      63.31 ms
  Max:      1714.21 ms
  Std Dev:  416.22 ms
  CV:       191.4%

### PERFORMANCE ANALYSIS
Compile-time dependencies add 217.48ms on average

### INDIVIDUAL MODULE ANALYSIS

Individual module import times (5 iterations each):
  component_decorator      :   0.26ms (min:  0.22, max:  0.35)
  container_component_decorator:   0.26ms (min:  0.20, max:  0.38)
  pipeline_context         :   0.72ms (min:  0.57, max:  0.91)
  tasks_group              :   3.04ms (min:  2.49, max:  3.68)
  pipeline_config          :   0.93ms (min:  0.50, max:  1.61)
  pipeline_task            :   0.91ms (min:  0.77, max:  1.21)
  for_loop                 :   0.55ms (min:  0.48, max:  0.63)
  importer_node            :   0.30ms (min:  0.24, max:  0.37)

Slowest compile-time modules:
  1. tasks_group: 3.04ms (43.6% of total)
  2. pipeline_config: 0.93ms (13.3% of total)
  3. pipeline_task: 0.91ms (13.1% of total)

HumairAK:
looking at Connor's initial pr on this https://github.com/kubeflow/pipelines/pull/9886

I think I agree with the decision on keeping the compiler dependencies not imported by default. He lists a couple of reasons for the design decision here, but I think the overall point is to keep the execution environment as lean and dependency free as possible. Likely because all we need from KFP is the executor logic, and not all of KFP. 

I don't think I have a strong opinion here, but eventually I do want us to move towards removing the KFP dependency entirely from executor, at which point users would be able to just do packages_to_install and we would remove the need for this env var entirely since KFP won't be required at runtime.

I would be interested in the thoughts of other maintainers @droctothorpe / @mprahl  / @zazulam  wdyt?

zazulam:
> looking at Connor's initial pr on this #9886
> 
> I think I agree with the decision on keeping the compiler dependencies not imported by default. He lists a couple of reasons for the design decision here, but I think the overall point is to keep the execution environment as lean and dependency free as possible. Likely because all we need from KFP is the executor logic, and not all of KFP.
> 
> I don't think I have a strong opinion here, but eventually I do want us to move towards removing the KFP dependency entirely from executor, at which point users would be able to just do packages_to_install and we would remove the need for this env var entirely since KFP won't be required at runtime.
> 
> I would be interested in the thoughts of other maintainers @droctothorpe / @mprahl / @zazulam wdyt?

I do agree on removing the KFP dependency from the executor along with that env variable would be ideal. As for this, I think it comes down to internal wrappers that inherit from kfp where the imports within kfp fail because of it does not load these compile time modules. I'm okay with this as long as we start tracking the removal of KFP within the executor, I'll create an issue for it.

zazulam:
created #12090 to track

droctothorpe:
I agree with you @HumairAK that target state should remove the dep altogether. In the meantime, avoiding dep conflicts and CVEs is valuable and worthwhile. Unfortunately, that still leaves users who want to explicitly leverage KFP code inside of a component in a bit of a pickle. There needs to be a way to import the non-runtime dependencies at runtime. With _KFP_RUNTIME hard-coded to True, the conditional always evaluates to False and the imports never happen. Making it configurable through the SDK like @cbartram demonstrated is one approach. If there is a simpler pattern, totally open to it.

HumairAK:
> Making it configurable through the SDK like @cbartram demonstrated is one approach. If there is a simpler pattern, totally open to it.

Sorry, I should have been clearer, yes my preference is that we keep this configurable and off by default, so users have the option to keep their containers as dependency-free as possible. @mprahl thoughts?

mprahl:
Sorry for the late response. I'm fine with keeping it configurable but I think it's a little strange to use the compile time options at runtime. If there's hesitation, an easy workaround is to do this in the pipeline function:

```python
import os
import importlib

os.environ["_KFP_RUNTIME"] = "false"
importlib.reload(kfp.dsl)

import kfp.kubernetes
```

cbartram:
> Sorry for the late response. I'm fine with keeping it configurable but I think it's a little strange to use the compile time options at runtime. If there's hesitation, an easy workaround is to do this in the pipeline function:
> 
> ```python
> import os
> import importlib
> 
> os.environ["_KFP_RUNTIME"] = "false"
> importlib.reload(kfp.dsl)
> 
> import kfp.kubernetes
> ```

@mprahl is it still worth making this configurable in the upstream or are there any other changes you'd like to see? We have a decorator workaround which patches the env var deployed internally but I'll test your snippet of code out as well as it may be a better fit.

mprahl:
@cbartram I think we could consider automating this by parsing if `kfp` is imported in the component. This would be added in the `execute` method in `sdk/python/kfp/dsl/executor.py`. Here some example code I generated with AI so review it thoroughly if you use it:

```python
import dis
from types import CodeType
from typing import Callable
import os

os.environ["_KFP_RUNTIME"] = "true"
import kfp
import kfp.dsl

def _is_kfp_imported_in_callable(func: Callable) -> bool:
    """Returns True if the callable imports `kfp` or any `kfp.*` submodule.

    This inspects Python bytecode for IMPORT_NAME instructions within the
    function and any nested code objects (e.g., inner functions, comprehensions).
    """
    seen_codes = set()

    def walk(code: CodeType) -> bool:
        if code in seen_codes:
            return False
        seen_codes.add(code)

        for instr in dis.get_instructions(code):
            if instr.opname == 'IMPORT_NAME' and isinstance(instr.argval, str):
                mod = instr.argval
                if mod == 'kfp' or mod.startswith('kfp.'):
                    return True

        for const in code.co_consts:
            if isinstance(const, CodeType) and walk(const):
                return True
        return False

    return walk(func.__code__)


def my_func():
    import kfp

def my_func2():
    from kfp import dsl

def my_func3():
    import os

print(_is_kfp_imported_in_callable(my_func))
print(_is_kfp_imported_in_callable(my_func2))
print(_is_kfp_imported_in_callable(my_func3))

if _is_kfp_imported_in_callable(my_func):
    import os
    import importlib
    import kfp
    import kfp.dsl

    os.environ["_KFP_RUNTIME"] = "false"
    importlib.reload(kfp.dsl)
    importlib.reload(kfp)

import kfp.kubernetes
print(kfp.kubernetes)
```

PR #12107, opened 2025-08-04
Fix samples-test workflow

akagami-harsh:
- part of https://github.com/kubeflow/pipelines/pull/11965

### DIFF
--- a/.github/workflows/kfp-samples.yml
+++ b/.github/workflows/kfp-samples.yml
@@ -86,6 +86,8 @@ jobs:
     - name: Run Samples Tests
       id: tests
       if: ${{ steps.forward-api-port.outcome == 'success' }}
+      env:
+        KFP_MULTI_USER: ${{ matrix.mode == 'multi-user' }}
       run: |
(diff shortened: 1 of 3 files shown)


akagami-harsh:
@juliusvonkohout I think i have picked up wrong changes while rebasing, these were causing samples tests to fail, this PR removes them.

.github/workflows/kfp-samples.yml
```diff
@@ -45,42 +45,17 @@ jobs:
       uses: actions/checkout@v2
 
     - name: Free up disk space
-      run: ./.github/resources/scripts/free-disk-space.sh
+      run: |
+        sudo rm -rf "$AGENT_TOOLSDIRECTORY"
+        docker system prune -af --volumes
+        docker image prune -af
```
hbelmiro, line 51: @akagami-harsh @juliusvonkohout, these commands should have been added to the `free-disk-space.sh` script. @akagami-harsh can you please send a follow-up PR?

PR #12121, opened 2025-08-07
chore(CI): Fix invalid references in workflow conditions and update GitHub Actions to use `actions/checkout@v4` consistently

hbelmiro:
Fixed invalid references in workflow conditions and updated GitHub Actions to use `actions/checkout@v4` consistently

**Checklist:**

### DIFF
--- a/.github/workflows/sdk-execution.yml
+++ b/.github/workflows/sdk-execution.yml
@@ -80,6 +80,7 @@ jobs:
 
       - name: Run tests
         id: tests
+        if: ${{ steps.forward-minio-port.outcome == 'success' }}
         env:
           PULL_NUMBER: ${{ github.event.pull_request.number }}
           REPO_NAME: ${{ github.repository }}
--- a/.github/workflows/ci-checks.yml
+++ b/.github/workflows/ci-checks.yml
@@ -13,7 +13,7 @@ jobs:
       pull-requests: write        
     steps:
       - name: Check out the repository
-        uses: actions/checkout@v3
+        uses: actions/checkout@v4
 
       - name: Check for 'needs-ok-to-test' and 'ok-to-test' labels
         id: label_check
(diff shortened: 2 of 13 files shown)


.github/workflows/sdk-execution.yml
```diff
@@ -80,6 +80,7 @@ jobs:
 
       - name: Run tests
         id: tests
+        if: ${{ steps.forward-minio-port.outcome == 'success' }}
```
nsingla, line 90: this won't run if Install Requirements is a failure and probably should not run as well. May be change this to depend on `install-requirements` and change `install-requirements` to depend on `install-protobuf-deps` and `install-protobuf-deps` to depend on `forward-minio-port` and `forward-mlmd-port` to depend on `forward-api-port` ?

hbelmiro: > this won't run if Install Requirements is a failure and probably should not run as well.

Yes, it shouldn't run.

> May be change this to depend on install-requirements

That's what effectively happens. The steps that don't have `continue-on-error: true` will fail the workflow, and none of the subsequent steps will run.

nsingla: Yup, but its good to be explicit about it, just to make it more readable and it will also display better in the workflow UI :)

hbelmiro: I think it will make it harder to read and maintain 😅
And doesn't have any impact on the UI.

droctothorpe:
Some merge conflicts need to be resolved. Otherwise, lgtm!

hbelmiro:
`actions/checkout@v5` was just released. I'm leveraging this PR to update it.

hbelmiro:
@droctothorpe @nsingla can you please take another look?

nsingla:
> `actions/checkout@v5` was just released. I'm leveraging this PR to update it.

Looks like runners were updated very recently that works with this latest `checkout` version :partying_face:

PR #12148, opened 2025-08-15
chore: Add CI for uploading pipelines as Kubernetes objects

mprahl:

Some of the changes are so that the pipeline names are compliant with Kubernetes object names. Note, the first commit is to address a potential bug that can be encountered with an invalid pipeline spec. See the commit for details.

**Checklist:**

### DIFF
--- a/.github/workflows/e2e-test.yml
+++ b/.github/workflows/e2e-test.yml
@@ -260,6 +260,71 @@ jobs:
           name: kfp-api-integration-tests-v2-artifacts-k8s-${{ matrix.k8s_version }}-${{ matrix.pipeline_store }}
           path: /tmp/tmp*/*
 
+  api-integration-tests-v2-compile-k8s:
+    runs-on: ubuntu-latest
+    needs: build
+    strategy:
+      matrix:
+        k8s_version: [ "v1.29.2", "v1.31.0" ]
+    name: API integration tests v2 - compile pipelines with Kubernetes - K8s ${{ matrix.k8s_version }}
+    steps:
+      - name: Checkout code
+        uses: actions/checkout@v5
+
+      - name: Set up Python
+        uses: actions/setup-python@v4
+        with:
+          python-version: 3.9
+
+      - name: Free up disk space
+        run: ./.github/resources/scripts/free-disk-space.sh
+
+      - name: Create KFP cluster
+        id: create-kfp-cluster
+        uses: ./.github/actions/kfp-cluster
+        with:
+          k8s_version: ${{ matrix.k8s_version }}
+          pipeline_store: kubernetes
+          image_path: ${{ needs.build.outputs.IMAGE_PATH }}
+          image_tag: ${{ needs.build.outputs.IMAGE_TAG }}
+          image_registry: ${{ needs.build.outputs.IMAGE_REGISTRY }}
+        continue-on-error: true
+
+      - name: Forward API port
+        id: forward-api-port
+        if: ${{ steps.create-kfp-cluster.outcome == 'success' }}
+        run: ./.github/resources/scripts/forward-port.sh "kubeflow" "ml-pipeline" 8888 8888
+        continue-on-error: true
+
+      - name: Forward MLMD port
+        id: forward-mlmd-port
+        if: ${{ steps.forward-api-port.outcome == 'success' }}
+        run: kubectl -n kubeflow port-forward svc/metadata-grpc-service 8080:8080 &
+        continue-on-error: true
+
+      - name: API integration tests v2
+        id: tests
+        if: ${{ steps.forward-api-port.outcome == 'success' }}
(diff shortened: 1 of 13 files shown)


.github/workflows/e2e-test.yml
```diff
@@ -254,6 +254,67 @@ jobs:
...
+      - name: Set up Python
+        uses: actions/setup-python@v4
+        with:
+          python-version: 3.9
+
+      - name: Create KFP cluster
+        id: create-kfp-cluster
+        uses: ./.github/actions/kfp-cluster
+        with:
+          k8s_version: v1.31.0
```
gmfrasca:
```suggestion
          k8s_version: ${{ matrix.k8s_version }}
```

.github/workflows/e2e-test.yml
```diff
@@ -254,6 +254,67 @@ jobs:
...
+        run: |
+          ./.github/resources/scripts/collect-logs.sh --ns kubeflow --output /tmp/tmp_pod_log.txt
+          exit 1
+
+      - name: Collect test results
+        if: always()
+        uses: actions/upload-artifact@v4
+        with:
+          name: kfp-api-integration-tests-v2-artifacts-compile-kubernetes
+          path: /tmp/tmp*/*
```
gmfrasca, line 326:
```suggestion
          path: /tmp/tmp*/*

```

.github/workflows/e2e-test.yml
```diff
@@ -260,6 +260,71 @@ jobs:
...
+        if: ${{ steps.create-kfp-cluster.outcome != 'success' || steps.forward-api-port.outcome != 'success' || steps.forward-mlmd-port.outcome != 'success' || steps.tests.outcome != 'success' }}
+        run: |
+          ./.github/resources/scripts/collect-logs.sh --ns kubeflow --output /tmp/tmp_pod_log.txt
+          exit 1
+
+      - name: Collect test results
+        if: always()
+        uses: actions/upload-artifact@v4
+        with:
+          name: kfp-api-integration-tests-v2-artifacts-compile-kubernetes
```
gmfrasca: Because this test uses matrix strategy, you should include a differentiator here, like `kfp-api-integration-tests-v2-artifacts-compile-kubernetes-k8s-${{ matrix.k8s_version }}`. It will work fine for now since k8s_version only has 1 item,  but if its ever expanded the following tests will have artifact storage paths collide and cause the test to fail

backend/test/v2/test_utils.go
```diff
@@ -218,3 +218,21 @@ func DeleteAllPipelines(client *api_server.PipelineClient, t *testing.T) {
...
 }
+
+func GetPipelineUploadClient(
+	uploadPipelinesWithKubernetes bool,
+	isKubeflowMode bool,
+	isDebugMode bool,
+	namespace string,
+	clientConfig clientcmd.ClientConfig,
+) (api_server.PipelineUploadInterface, error) {
+	if uploadPipelinesWithKubernetes {
```
gmfrasca, line 229: optional: given this is essentially a Factory pattern, this feels like it should be a enum rather than a series of bool switches. ie take in mode=`(kubeflow|kubernetes|default)`.

backend/src/common/client/api_server/v2/pipeline_upload_client_kubernetes.go
```diff
@@ -0,0 +1,307 @@
...
+	if parameters.DisplayName != nil {
+		displayName = *parameters.DisplayName
+	} else {
+		displayName = name
+	}
+
+	var description string
+	if parameters.Description != nil {
+		description = *parameters.Description
+	}
```
gmfrasca: This block is used several times in this client, perhaps we should centralize it as a func?

.github/workflows/e2e-test.yml
```diff
@@ -260,6 +260,71 @@ jobs:
           name: kfp-api-integration-tests-v2-artifacts-k8s-${{ matrix.k8s_version }}-${{ matrix.pipeline_store }}
           path: /tmp/tmp*/*
 
+  api-integration-tests-v2-compile-k8s:
+    runs-on: ubuntu-latest
+    needs: build
+    strategy:
+      matrix:
+        k8s_version: ["v1.31.0"]
```
nsingla: is the assumption that there is no impact of K8s version? but may be its good to confirm that by adding another k8s version here?

.github/workflows/e2e-test.yml
nsingla: or you can use `${{ github.job }}` which I believe will be the parameterized name

backend/test/v2/test_utils.go
mprahl: I'll skip this for now since it would involve a lot more refactoring in the rest of the test code.

PR #12149, opened 2025-08-15
Fix failing backend test

akagami-harsh:
- part of https://github.com/kubeflow/pipelines/pull/11965
- Remove `Test_GetMinioDefaultEndpoint` cause the corresponding remove `GetMinioDefaultEndpointfunction is also removed in last PR

### DIFF
--- a/backend/src/v2/objectstore/object_store_test.go
+++ b/backend/src/v2/objectstore/object_store_test.go
@@ -17,7 +17,6 @@ package objectstore
 import (
 	"context"
 	"fmt"
-	"os"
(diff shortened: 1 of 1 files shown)


hbelmiro (commented):
If it's part of https://github.com/kubeflow/pipelines/pull/11965, why not include this change in that PR?

akagami-harsh:
> If it's part of #11965, why not include this change in that PR?

This pr is opened against seaweedfs branch, and also I don't have access to seaweedfs branch

PR #12151, opened 2025-08-15
test: Run e2e CI tests using various versions of AWF. Fixes #12150

### LINKED ISSUE #12150: [feature] Test against multiple versions of Argo backend
### Feature Area

/area backend

Run CI checks against several different minor versions of Argo to test compatibility

Some KFP users already have a separate, unrelated Argo Workflows installed and therefore disable deployment of the on-board Workflow Controller to prevent conflicts (vanilla AWF picking up KFP-based Workflows, for example).  Keeping these external AWF installations can be a pain point, which may lead to adoption hesitancy.

Testing multiple versions of Argo in CI can give us some confidence that KFP is compatible using one of a range of minor versions of the AWF dependency, which may alleviate some concerns there.

Besides keeping standalone AWFs and KFP installations on seperate k8s clusters, users can workaround this by configuring their AWF and Workflows to use a InstanceID to isolate the vanilla workflows from KFP, but this is cumbersome (and KFP and its Pipelines, inversely, do not have this concept implemented).

---

Love this idea? Give it a 👍.

gmfrasca:
Add an e2e CI check to test KFP installations deployed with different minor versions of the Argo Workflow Controller.  For this instance, v3.6.7 (current) and v3.5.14 (previous) versions are used, but can be expanded and/or updated in the workflow defintion's input parameter

**Checklist:**

### DIFF
--- a/.github/workflows/e2e-test.yml
+++ b/.github/workflows/e2e-test.yml
@@ -389,6 +389,71 @@ jobs:
           name: kfp-api-integration-tests-v2-with-cache-disabled-artifacts-k8s-${{ matrix.k8s_version }}
           path: /tmp/tmp*/*
 
+  api-integration-tests-v2-argo:
+    runs-on: ubuntu-latest
+    needs: build
+    strategy:
+      matrix:
+        k8s_version: [ "v1.31.0" ]
+        argo_version: [ "v3.5.14" ]
+    name: API Integration tests v2 - Argo ${{ matrix.argo_version }}
+    steps:
+      - name: Checkout code
+        uses: actions/checkout@v5
+      
+      - name: Free up disk space
+        run: ./.github/resources/scripts/free-disk-space.sh
+
+      - name: Set up Python
+        uses: actions/setup-python@v4
+        with:
+          python-version: 3.9
+
+      - name: Create KFP cluster
+        id: create-kfp-cluster
+        uses: ./.github/actions/kfp-cluster
+        with:
+          k8s_version: ${{ matrix.k8s_version }}
+          image_path: ${{ needs.build.outputs.IMAGE_PATH }}
+          image_tag: ${{ needs.build.outputs.IMAGE_TAG }}
+          image_registry: ${{ needs.build.outputs.IMAGE_REGISTRY }}
+          argo_version: ${{ matrix.argo_version }}
+        continue-on-error: true
+
+      - name: Forward API port
+        id: forward-api-port
+        if: ${{ steps.create-kfp-cluster.outcome == 'success' }}
+        run: ./.github/resources/scripts/forward-port.sh "kubeflow" "ml-pipeline" 8888 8888
+        continue-on-error: true
+
+      - name: Forward MLMD port
+        id: forward-mlmd-port
+        if: ${{ steps.forward-api-port.outcome == 'success' }}
+        run: kubectl -n kubeflow port-forward svc/metadata-grpc-service 8080:8080 &
+        continue-on-error: true
+
+      - name: API integration tests v2
+        id: tests
+        if: ${{ steps.forward-mlmd-port.outcome == 'success' }}
+        working-directory: ./backend/test/v2/integration
+        run: go test -v ./... -namespace kubeflow -args -runIntegrationTests=true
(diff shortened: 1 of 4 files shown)


nsingla:
should we also update unit tests to run argo compiler tests with different version of argo dependency?

.github/workflows/e2e-test.yml
```diff
@@ -377,6 +377,62 @@ jobs:
           name: kfp-api-integration-tests-v2-with-cache-disabled-artifacts-k8s-${{ matrix.k8s_version }}
           path: /tmp/tmp*/*
 
+  initialization-tests-v2-argo:
+    runs-on: ubuntu-latest
+    needs: build
+    strategy:
+      matrix:
+        k8s_version: [ "v1.31.0" ]
+        argo_version: [ "v3.6.7", "v3.5.14" ]
```
nsingla: is this `n` and `n-1`? and any reason why k8s version was removed (do you not foresee any  breaking change with k8s version change)? also what about cacheDisabled?

gmfrasca: this is current and the previous versions pinned/used in KFP.  It appears since upgrading from 3.5.14 to 3.6.7 Argo had since released 3.5.15.

this test is only checking one k8s version as adding a second doubles the test matrix which significantly increases runtime and resource/runner requirements.  k8s compatibility is checked in a separate test, so found it to be duplicated work to also check here.  same goes for cachedisabled and external k8s - each option doubles the test matrix size

gmfrasca:
> should we also update unit tests to run argo compiler tests with different version of argo dependency?

@nsingla - the intent of this check is to verify backwards compatibility with previous AWF versions, without requiring multiple build streams of the backend.  updating the code dependencies to check for argo compilation would mean we would also need to include multiple versions of each component image on release (ie apiserver-awf3.6, apiserver-awf3.5, etc) which is outside the scope of this proposal

.github/workflows/e2e-test.yml
```diff
@@ -377,6 +377,62 @@ jobs:
           name: kfp-api-integration-tests-v2-with-cache-disabled-artifacts-k8s-${{ matrix.k8s_version }}
           path: /tmp/tmp*/*
 
+  initialization-tests-v2-argo:
```
mprahl:
```suggestion
  api-integration-tests-v2-argo:
```

.github/workflows/e2e-test.yml
mprahl: This should include a single version since the `api-integration-tests-v2` workflow already covers `v3.6.7`.

mprahl:
@gmfrasca could you also please add the multiple Argo versions to the [kfp-samples.yml](https://github.com/kubeflow/pipelines/blob/master/.github/workflows/kfp-samples.yml) workflow? That has more comprehensive coverage of various Argo Workflow features where as the API integration tests is mostly just testing the REST API with just basic pipeline runs.

.github/workflows/e2e-test.yml
gmfrasca: updated

.github/workflows/e2e-test.yml
gmfrasca: ah good point, updated

.github/workflows/e2e-test.yml
```diff
@@ -389,6 +389,62 @@ jobs:
...
+      - name: Forward API port
+        id: forward-api-port
+        if: ${{ steps.create-kfp-cluster.outcome == 'success' }}
+        run: ./.github/resources/scripts/forward-port.sh "kubeflow" "ml-pipeline" 8888 8888
+        continue-on-error: true
+
+      - name: Initialization tests v2
+        id: tests
+        if: ${{ steps.forward-api-port.outcome == 'success' }}
+        working-directory: ./backend/test/v2/initialization
```
mprahl: I think you may have copied the wrong GitHub workflow since I don't think you meant to run the "initialization" tests but rather the "integration" tests.

gmfrasca: oops, updated (actually did the change last week, commenting now to resolve)

PR #12159, opened 2025-08-20
[WIP] test(backend): add CI workflow for integration tests with Postgres

kaikaila:
This PR adds a new GitHub Actions workflow that runs backend integration tests against Postgres using the pgx driver.

## The workflow

- 	Spins up a KinD cluster.
- 	Installs a default StorageClass (local-path).
- 	Deploys Postgres via manifests/kustomize/third-party/postgresql (service postgres-service).
- 	Bootstraps a test user (user/password) and database (mlpipeline).
- 	Port-forwards to 127.0.0.3:5432.
- Runs backend integration tests with environment variables set for DB_TYPE=postgres and DB_DRIVER=pgx.

It enables the CI to run the same integration suite on Postgres, providing a repeatable safety net for the ongoing storage-layer refactor (e.g. LargeText changes).

## Notes
- Workflow is manual-triggered (workflow_dispatch) with inputs for repository/branch, so maintainers can run tests on fork branches.
- Marked with continue-on-error: true initially, since Postgres tests will fail until storage-layer refactor is completed.
- Follow-ups: improve readiness checks and make the job required once Postgres passes.

## How to run this workflow
- Go to Actions → KFP Backend Postgres Tests → Run workflow.
- Provide the following inputs (example values):

> test_repository: <your-github-id>/kubeflow-pipelines (use your fork repo if testing unmerged code)
> test_ref: feature/replace-str-with-largetext-WIP (branch, tag, or commit SHA to test)
> test_paths: ./backend/test/integration/... (or ./backend/test/initialization if you only want to run smoke tests)

## Reference Run
Here is a successful [run](https://github.com/kaikaila/kubeflow-pipelines/actions/runs/17085693190/job/48449263001) on my fork.

Note: The reference run workflow is not identical to this PR’s version — the fork-specific workflow uses different triggers and a fixed checkout target, while the upstream PR workflow relies solely on workflow_dispatch with parameterized test_repository/test_ref.

### DIFF
--- a/.github/workflows/kfp-backend-postgres-tests.yml
+++ b/.github/workflows/kfp-backend-postgres-tests.yml
@@ -10,6 +10,8 @@ on:
       - ".github/workflows/kfp-backend-postgres-tests.yml"
       - "!**/*.md"
       - "!**/OWNERS"
+env:
+  NAMESPACE: kfp-pgx-test
 jobs:
   postgres-pgx:
     name: Backend Postgres (pgx via Kustomize+KinD)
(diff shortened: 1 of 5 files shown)


shivaylamba (approved):

.github/workflows/kfp-backend-postgres-tests.yml
```diff
@@ -0,0 +1,245 @@
...
+    inputs:
+      test_repository:
+        description: "owner/repo to checkout (empty = upstream)"
+        required: false
+        default: ""
+      test_ref:
+        description: "Git ref (branch/tag/SHA) to test"
+        required: true
+        default: "master"
+      test_paths:
```
HumairAK: this does not seem like it is used

.github/workflows/kfp-backend-postgres-tests.yml
```diff
@@ -0,0 +1,245 @@
+name: KFP Backend Postgres Tests
+
+on:
+  workflow_dispatch:
```
HumairAK: please see the other CI test triggers (e.g. [here](https://github.com/kubeflow/pipelines/blob/9526abb03d603b0d367fd2f13c8a1949388621a5/.github/workflows/kfp-samples.yml#L3-L7)) and emulate those with the paths set to the files affected, we need them to trigger on pushes to master and PR CI

.github/workflows/kfp-backend-postgres-tests.yml
```diff
@@ -0,0 +1,245 @@
...
+    name: Backend Postgres (pgx via Kustomize+KinD)
+    runs-on: ubuntu-latest
+    continue-on-error: true # Temporarily allow failures; switch to false after LargeText/storage changes land
+
+    steps:
+      - name: Checkout target code
+        uses: actions/checkout@v4
+        with:
+          repository: ${{ github.event.inputs.test_repository || github.repository }}
+          ref: ${{ github.event.inputs.test_ref }}
```
HumairAK: I don't think we need this, if you are running the action from a fork it will automatically check out code from there (hence `github.repository`), and there's no reason to run this CI from a remote repository in the KFP repo.

.github/workflows/kfp-backend-postgres-tests.yml
```diff
@@ -0,0 +1,245 @@
...
+      test_paths:
+        description: "Go package paths"
+        required: true
+        default: "./backend/test/initialization"
+
+jobs:
+  postgres-pgx:
+    name: Backend Postgres (pgx via Kustomize+KinD)
+    runs-on: ubuntu-latest
+    continue-on-error: true # Temporarily allow failures; switch to false after LargeText/storage changes land
```
HumairAK, line 19: We should include this Github workflow with the PR that resolve these failures and not prior.

.github/workflows/kfp-backend-postgres-tests.yml
```diff
@@ -0,0 +1,245 @@
...
+        run: |
+          sudo apt-get update
+          sudo apt-get install -y postgresql-client netcat-openbsd jq
+
+      - name: Create namespace
+        env:
+          NAMESPACE: kfp-pgx-test
+        run: kubectl create namespace "$NAMESPACE" || true
+
+      - name: Install default StorageClass for KinD (local-path)
```
HumairAK: we have a `kind` cluster action at `.github/actions/kfp-cluster` which will use the kind action, can we just re-use that here instead of setting up the kind cluster from scratch?

.github/workflows/kfp-backend-postgres-tests.yml
```diff
@@ -0,0 +1,245 @@
...
+          # Sanity check: ensure there is a default StorageClass before deploying PG
+          kubectl get storageclass | tee /dev/stderr | grep "(default)" >/dev/null
+
+      - name: Deploy Postgres via Kustomize
+        working-directory: manifests/kustomize/third-party/postgresql
+        env:
+          NAMESPACE: kfp-pgx-test
+        run: |
+          kustomize build ./base | kubectl -n "$NAMESPACE" apply -f -
+          kubectl -n "$NAMESPACE" get all
```
HumairAK: can we instead update the `deploy-kfp.sh` to take [another flag](https://github.com/kubeflow/pipelines/blob/c11a7358f566be93c65131cf3f3cfbc722b40080/.github/resources/scripts/deploy-kfp.sh#L35) and deploy postgresql option? then use the [kfp deploy cluster action](https://github.com/kubeflow/pipelines/blob/9526abb03d603b0d367fd2f13c8a1949388621a5/.github/workflows/e2e-test.yml#L42). 

This script will also wait for pods to ready so you can skip the `Wait for Postgres pod ready` step as well.

.github/workflows/kfp-backend-postgres-tests.yml
```diff
@@ -0,0 +1,245 @@
...
+            sleep 2
+          done
+          [ -n "${EP:-}" ] || { echo "no endpoints"; cat /tmp/pf.log || true; exit 1; }
+
+          # Wait until local port is open
+          for i in {1..60}; do nc -z 127.0.0.3 "${SVC_PORT}" && break; sleep 2; done
+          nc -z 127.0.0.3 "${SVC_PORT}" || { echo "local port not open"; cat /tmp/pf.log || true; exit 1; }
+
+          # Optional: psql probe (not fatal; default db/user may not exist)
+          PGPASSWORD=kubeflow psql -h 127.0.0.3 -p "${SVC_PORT}" -U kubeflow -d kubeflow -c "SELECT 1;" || true
```
HumairAK: I don't think we need these, the postgresql deployment already has a readiness probe, so if the pod is ready, the svc should be ready. And for the second one, we can assume kubectl port-forward is working as expected.

.github/workflows/kfp-backend-postgres-tests.yml
```diff
@@ -0,0 +1,245 @@
...
+
+      - name: Bootstrap PG objects for tests (discover superuser & create user/db)
```
HumairAK: @kaikaila correct me if I'm wrong but it seems like we are already specifying the super user and db [here](https://github.com/kubeflow/pipelines/blob/e1f0c010f80031ea09af69f9bbedf2e24509605f/manifests/kustomize/third-party/postgresql/base/pg-deployment.yaml#L22-L31), so I'm wondering why this step is needed? we can just switch the db name to the one matching the deployment yes? 

The rest seem like a lot of script to verify/find information we already have, i.e. super user name/password - since we are specifying in the deployment, I'm fine with just using it as is instead of searching for it to simplify the workflow. Wdyt?

.github/workflows/kfp-backend-postgres-tests.yml
kaikaila: Got it. I’ll stop working on this CI-only PR. I’ll include the workflow in the larger PR #12063 and apply your suggestions there. PR #12063 will also make the PGX lane green (with continue-on-error: false).

Issue #12047, opened 2025-07-14
[feature][RFC] Introduce LargeText type to support dialect-aware large string fields


kaikaila:
As part of our effort to support PostgreSQL alongside MySQL, we propose replacing certain string fields in the model layer with a custom LargeText type that is dialect-aware. This change enables proper handling of large text fields in both MySQL (LONGTEXT) and PostgreSQL (TEXT) during schema migration.

⸻

Motivation

Directly using gorm:"type:longtext" in struct tags works fine for MySQL, but fails immediately on PostgreSQL during table creation (AutoMigrate), because PostgreSQL does not recognize longtext as a valid SQL type. Unfortunately, this error prevents any downstream logic (such as dialect-switching in initDBClient) from being executed.

⸻

Proposed Solution

As the [GORM doc recommends](https://gorm.io/docs/data_types.html#GormDataTypeInterface), define a custom type LargeText string and implement GormDBDataType(db, field):
```
func (LargeText) GormDBDataType(db *gorm.DB, field *schema.Field) string {
    switch db.Dialector.Name() {
    case "mysql":
        return "LONGTEXT"
    case "postgres":
        return "TEXT"
    }
    return "TEXT"
}
```

Then update relevant fields from string to LargeText.

⸻

Breaking Change: Risks and Migration Plan

This is a breaking change in Go code. It introduces the following implications:

1. Model-Level Compatibility
	•	Code that previously used string fields must now convert explicitly: model.LargeText(someString)
	•	Any direct use of these fields in Create/Update must be updated to reflect the new type.

2. Third-Party Extension Breakage
	•	If downstream users or plugins have embedded these models, their code may fail to compile due to type mismatches.

3. Tooling Compatibility
	•	JSON serializers, reflection-based tools, or hooks expecting string may need verification or adaptation.

4. Migration & Testing
	•	While the SQL DDL output remains functionally similar (TEXT or LONGTEXT), all migration scripts and integration tests must be re-verified under both MySQL and PostgreSQL.

⸻

Alternatives Considered
	•	❌ Using gorm:"type:longtext" with fallback logic in initDBClient: Not viable, because Postgres rejects invalid types at DDL time, and the fallback logic never runs.
	•	⚠️ Writing dialect-specific tag overrides or invoking Migrator().CreateTable() with dynamic tag rewrites: possible but hacky and unmaintainable.

⸻

Request for Feedback

We invite maintainers and community members to:
	•	 Review the breaking change and its migration impact.
[...]

PR #12163, opened 2025-08-20
refactor(backend): introduce dialect-aware LargeText and replace longtext (+call-site type adjustments). Part of #12063

kaikaila:
Introduce an internal LargeText model type and replace all `gorm: type:longtext` usages. LargeText implements `GormDBDataType()` to map to MySQL=LONGTEXT and Postgres=TEXT. This unblocks Postgres AutoMigrate while keeping MySQL schemas unchanged.

## Motivation

Postgres fails on migrate with type "longtext" does not exist (SQLSTATE 42704). Making the type dialect-aware removes this blocker.

## Scope
Although this PR is “models-first”, some call sites needed type conversions/wiring only so code compiles and tests run. **No query semantics or business logic were changed.**

## Backward compatibility
- MySQL: No schema drift (still LONGTEXT); existing data unaffected.
- Postgres: Fresh installs migrate cleanly (no longtext errors).
- External API / user-visible behavior: No change (internal refactor only).

## Testing
- MySQL backend integration tests remain green (no expected diffs).
- Postgres (pgx) CI job is currently marked allowed to fail; this PR should remove the early longtext migration failure and let tests proceed further.
- Local reproduction (example):

```
# MySQL
DB_TYPE=mysql go test -v ./backend/test/integration/...

# Postgres (pgx)
DB_TYPE=postgres DB_DRIVER=pgx go test -v ./backend/test/integration/... \
  -args -runIntegrationTests=true -isDevMode=true -runPostgreSQLTests=true -localTest=true
```

## Notes for reviewers
- This PR is intentionally models-only to keep review surface small.
- No query/dialect logic changes here; follow-ups will migrate stores to the injected sqBuilder gradually.

## Changelog

Internal refactor; no user-facing change. No release note needed.

Part of #12063.

### Reference

This PR is part of RFC discussion in [#12047](https://github.com/kubeflow/pipelines/issues/12047).

### Technical Appendix
<details>
<summary>Click to expand full investigation summary</summary>

### 🔍 Background & Rationale

Kubeflow Pipelines previously relied on `gorm:"size:65535"` to represent large string fields (e.g., descriptions, manifests, parameters).  
In **GORM v1**, this produced compatible behavior:
- [MySQL](https://github.com/DoOR-Team/gorm/blob/88b21140e68ece527a36f54af8654512d92a4909/dialect_mysql.go#L91): mapped to `LONGTEXT`
- [PostgreSQL](https://github.com/DoOR-Team/gorm/blob/88b21140e68ece527a36f54af8654512d92a4909/dialect_postgres.go#L56): mapped to `TEXT`

However, in **GORM v2**, internal dialect logic has changed:

- [MySQL](https://github.com/go-gorm/mysql/blob/d744156745998478ffa3c5395c3336ae6f0b0c0c/mysql.go#L391) now interprets `size:65535` as `MEDIUMTEXT`, which may be **too small for our payloads**
- [PostgreSQL](https://github.com/go-gorm/postgres/blob/master/postgres.go) still uses `TEXT`, but will **fail** if `type:longtext` is explicitly used (since Postgres doesn’t recognize it)
- Using `VARCHAR(65535)` in [MySQL](https://dev.mysql.com/doc/refman/8.4/en/column-count-limit.html) can exceed row size limits and lead to errors

Thus, the old approach is no longer reliable for cross-database compatibility.

⸻

✅ Chosen Solution: Define a LargeText Custom Type

Following GORM official [recommendation](https://gorm.io/docs/data_types.html#GormDataTypeInterface), We introduce a dialect-aware LargeText type:

```go
type LargeText string

func (LargeText) GormDBDataType(db *gorm.DB, field *schema.Field) string {
	switch db.Dialector.Name() {
	case "mysql":
		return "LONGTEXT"
	case "postgres", "pgx":
		return "TEXT"
	default:
		return "TEXT"
	}
}
```

This approach is :
	•	Ensures correct behavior for both MySQL and PostgreSQL
	•	Avoids manual ALTER COLUMN migrations
	•	Makes the semantic intent (this is a large string field) explicit in code

⸻

🛠 Refactor Scope

We’ve audited existing fields and begun migrating high-risk ones like:
	•	Description
	•	Parameters
	•	PipelineSpecURI

This surfaced ~13 compile-time errors due to type mismatches, primarily in:
	•	Struct literal initializations
	•	Function parameter type checks

Most of these are resolved with explicit conversions like LargeText(x) or string(x).

⸻

🚨 Fallback Plan

If this refactor proves too invasive, we will fall back to a simpler (but less elegant) approach, which currenly is adopted in the [GORM-v1 kfp](https://github.com/kubeflow/pipelines/blob/daac099508865670f41eeeef135fa22f9ec880f1/backend/src/apiserver/client_manager/client_manager.go#L430):
	•	Keep gorm:"type:longtext" for MySQL only fields
	•	Use db.Migrator().AlterColumn(...) during DB init to patch schema post-facto
</details>


mprahl:
@kaikaila is this PR still WIP?

kaikaila:
Hi @mprahl , thanks for checking in!
I’ve just removed the WIP from the title—this PR is now ready for review.

mprahl (approved):
/approve
/lgtm

It's unfortunate that we have to change Go types but it seems there's no other way. Thanks for digging into this!

mprahl:
@kaikaila it looks like there are some merge conflicts. Let me know when they are resolved.

kaikaila:
Thanks a lot @mprahl ! I really appreciate your review and the LGTM.
There was a conflict after master moved forward, so I rebased and resolved it. The CI is re-running now — once it finishes, if everything is good, I'll ping you.

PR #12164, opened 2025-08-20
[WIP] refactor(backend): InitDBClient returns (db, sqBuilder) for centralized SQL builder injection

kaikaila:
What
- Change InitDBClient to return a preconfigured Squirrel StatementBuilderType alongside the DB handle:
InitDBClient(...) -> (db, statementBuilder [, dialect])
- Update the minimal call sites to accept the new return value(s).
- No query/behavior changes in this PR.

Why

Centralize SQL builder configuration (placeholder format, quoting via builder) so that follow-up store refactors can use an injected builder instead of ad-hoc configuration. This is a surgical step that reduces risk and keeps subsequent PRs focused per store.

Scope
- Touches client_manager.go (function signature + return path) and the few callers that receive the new value(s).
- Out of scope: store SQL rewrites, filter extraction, schema/model changes.

Backward compatibility
- MySQL: No behavior change (queries and schema remain identical).
- Postgres (pgx): No functional changes yet; this is prep work only.

Testing
- Backend (MySQL) tests remain green (no behavioral deltas expected).
- Postgres (pgx) CI job (currently allowed to fail) should not regress.

Notes
- This PR is intentionally small to keep review surface minimal.
- Follow-ups will switch individual stores to the injected statementBuilder (and dialect).

Part of #12063.

### DIFF
--- a/backend/src/apiserver/client_manager/client_manager.go
+++ b/backend/src/apiserver/client_manager/client_manager.go
@@ -264,7 +264,7 @@ func (c *ClientManager) init(options *Options) error {
 
 	glog.Info("Initializing client manager")
 	glog.Info("Initializing DB client...")
-	db := InitDBClient(common.GetDurationConfig(initConnectionTimeout))
+	db, _ := InitDBClient(common.GetDurationConfig(initConnectionTimeout))
 	db.SetConnMaxLifetime(common.GetDurationConfig(dbConMaxLifeTime))
 	glog.Info("DB client initialized successfully")
 
@@ -314,7 +314,7 @@ func (c *ClientManager) Close() {
 	c.db.Close()
 }
 
-func InitDBClient(initConnectionTimeout time.Duration) *storage.DB {
+func InitDBClient(initConnectionTimeout time.Duration) (*storage.DB, SQLDialect) {
 	// Allowed driverName values:
 	// 1) To use MySQL, use `mysql`
 	// 2) To use PostgreSQL, use `pgx`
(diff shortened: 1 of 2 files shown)


backend/src/apiserver/client_manager/client_manager.go
```diff
@@ -360,7 +360,12 @@ func InitDBClient(initConnectionTimeout time.Duration) *storage.DB {
 	if err != nil {
 		glog.Fatalf("Failed to retrieve *sql.DB from gorm.DB. Error: %v", err)
 	}
-	return storage.NewDB(newdb, storage.NewMySQLDialect())
+	// TODO(kaikaila):
```
rimolive, line 363: @kaikaila Is this part of another PR or this is still WIP?

kaikaila: @rimolive  Thanks for checking in!

This PR was part of the refactor sequence that culminates in [#12063](https://github.com/kubeflow/pipelines/pull/12063). Per Humair's [guidance](https://github.com/kubeflow/pipelines/pull/12063#issuecomment-3220384042), we moved towards consolidating everything into a single large umbrella PR — so the TODO noted here was eventually completed as part of #12063 directly, rather than in this sub-PR.

If you'd prefer for each sub-PR to be cleaned up and merged independently, I'm totally happy to re-scope — just let me know if you're planning to review this series or prefer that I continue aligning with Humair's single-PR strategy.

Appreciate the feedback either way!

zazulam: I think the guidance for a single large PR is fine, if that is the path you are continuing on can you mark this PR as closed?

PR #12180, opened 2025-08-25
test: Add Integration tests for Kubernetes Native API Migration script

VaniHaripriya:

- Created a GitHub workflow and corresponding tests to verify the Kubernetes Native API migration script.

**Checklist:**


.github/workflows/kfp-kubernetes-native-migration-tests.yaml
```diff
@@ -0,0 +1,188 @@
+name: KFP Kubernetes Native Migration Tests
+
+on:
+  push:
+    branches: [master]
+
+  pull_request:
+    paths:
```
nsingla, line 8: should we add a trigger to trigger when there is a change in the migration script or when the corresponding code changes?

.github/resources/scripts/forward-port.sh
```diff
@@ -23,7 +23,7 @@ APP_NAME=$2
 LOCAL_PORT=$3
 REMOTE_PORT=$4
 
-POD_NAME=$(kubectl get pods -n "$KUBEFLOW_NS" -l "app=$APP_NAME" -o jsonpath='{.items[0].metadata.name}')
+POD_NAME=$(kubectl get pods -n "$KUBEFLOW_NS" -l "app=$APP_NAME" --field-selector=status.phase=Running --sort-by='.metadata.creationTimestamp' -o jsonpath='{.items[-1].metadata.name}')
```
nsingla, line 26: i wonder why this was required to be changed :thinking:

nsingla (commented):
I feel like a lot of the assertions are hardcoded and only verifying the presence of files/pipelines. 
We should probably write these tests in pytest rather than Python Unit test, and use of fixture for set up and tear down can make things much easier.
Can you please add some code comments explaining what the tests are suppose to do? May be add a link it to the test in the proposal per test
Also, curious to know why were these tests written in python as oppose to Golang? 

Since this is probably just 1 off thing, we can live by with the hard coded assertions but we should still add assertions to compare whole objects rather than just the presence of pipelines/runs/versions

.github/resources/scripts/forward-port.sh
VaniHaripriya: This update ensures that the newest running pod is selected. Previously, it sometimes connected to an old pod from database mode when switching to Kubernetes mode, which caused connection errors.

.github/workflows/kfp-kubernetes-native-migration-tests.yaml
VaniHaripriya: "tools/k8s-native/**" - this path takes care of the changes for migration script. Adding few more triggers for corresponding code changes.

nsingla: don't think we need to run migration tests when there is a change `api` every time? may be we should only run when there is a change in the backend code, specifically in the webhooks and in the api server + common code?

.github/workflows/kfp-kubernetes-native-migration-tests.yaml
```diff
@@ -0,0 +1,192 @@
...
+        if: ${{ steps.create-test-pipelines.outcome == 'success' }}
+        id: run-db-mode-tests
+        env:
+          PULL_NUMBER: ${{ github.event.pull_request.number }}
+          REPO_NAME: ${{ github.repository }}
+          KFP_ENDPOINT: "http://localhost:8888"
+        run: python -m pytest ./test/kfp-kubernetes-native-migration-tests/kfp_db_mode_migration_tests.py -v
+        continue-on-error: true
+
+      - name: Install cert-manager
```
alyssacgoins, line 99: What's the reasoning behind installing cert manager after the tests DB mode migration tests are run?

VaniHaripriya: Good question! Before switching to the Kubernetes native API deployment mode, cert-manager must be installed to manage the webhook deployment.

.github/resources/scripts/forward-port.sh
DharmitD, line 26: `{.items[-1].metadata.name}` is invalid JSONPath syntax. JSONPath doesn't support negative indices like Python. You could pipe into `jq` though. Something like:

```
POD_NAME=$(kubectl get pods -n "$KUBEFLOW_NS" -l "app=$APP_NAME" \
  --field-selector=status.phase=Running \
  --sort-by=.metadata.creationTimestamp -o json \
  | jq -r '.items[-1].metadata.name')
```

DharmitD: `jsonpath='{.items[*].metadata.name}'` with a tail command could also work.

test/kfp-kubernetes-native-migration-tests/create_test_pipelines.py
```diff
@@ -0,0 +1,258 @@
...
+import sys
+import os
+from pathlib import Path
+import kfp
+from kfp.client import Client
+
+
+KFP_ENDPOINT = os.environ.get('KFP_ENDPOINT', 'http://localhost:8888')
+
+def serialize_object_for_comparison(obj):
```
nsingla: would json.dump or json.dumps, not suffice ? Any particular reason for this method?

.github/resources/scripts/forward-port.sh
VaniHaripriya: I used `{.items[-1].metadata.name}` to ensure that the newest running pod is selected. Previously, it sometimes connected to an old pod from database mode when switching to Kubernetes mode, which caused connection errors.

DharmitD: My bad, you're right, turns out negative indexing does work!

test/kfp-kubernetes-native-migration-tests/kfp_k8s_mode_migration_tests.py
```diff
@@ -0,0 +1,402 @@
...
+    # Validate experiment structure
+    experiment_details = kfp_client.get_experiment(experiment_id=experiment_id)
+    assert getattr(experiment_details, 'experiment_id', None) == experiment_id, "Experiment should have correct ID"
+    assert getattr(experiment_details, 'display_name', None) == experiment_name, "Experiment should have correct name"
+    experiment_description = getattr(experiment_details, 'description', None)
+    assert experiment_description == "Test experiment created in K8s mode", "Experiment should have correct description"
+    
+    # Validate run structure
+    run_details = kfp_client.get_run(run_id=run_id)
+    assert getattr(run_details, 'run_id', None) == run_id, "Run should have correct ID"
```
nsingla: can we just not do 
`assert run_details.run_id == run_data.run_id`?

nsingla: And similarly everywhere else where its possible?

nsingla (commented):
Rather than dealing with dict attributes, can you convert test data after migration into a python object and use python objects for assertions?

PR #12185, opened 2025-08-26
feat(sdk/backend): Support forwarding task configuration to external workloads

mprahl:

This PR introduces a powerful new feature that allows KFP tasks to forward their task configuration (resources, environment variables, node selectors, tolerations, affinity, volumes) to external workloads like Kubeflow TrainJobs, rather than applying them only to the task pod itself. This is particularly valuable for scenarios where the KFP task is launching another Kubernetes workload (e.g. `TrainJob`, `Job`, etc).

**Checklist:**

### DIFF
--- a/api/v2alpha1/pipeline_spec.proto
+++ b/api/v2alpha1/pipeline_spec.proto
@@ -95,6 +95,8 @@ message ComponentSpec {
   }
   // Supports platform-specific component features.
   repeated SinglePlatformSpec single_platform_specs = 5;
+  // Specifies the task configurations that can be passed through to an external workload.
+  repeated TaskConfigPassthrough task_config_passthroughs = 6;
 }
 
 // A DAG contains multiple tasks.
@@ -442,7 +444,43 @@ message ParameterType {
     // Indicates that a parameter is a TaskFinalStatus type; these types can only accept inputs
     // specified by InputParameterSpec.task_final_status
     TASK_FINAL_STATUS = 7;
+    // Indicates that a parameter is a TaskConfig type; these types are
+    // injected by the backend to provide the configuration set on the task.
+    TASK_CONFIG = 8;
   }
+}
+
+
+// Represents the task configurations that can be passed through to an external workload.
+message TaskConfigPassthroughType {
+  enum TaskConfigPassthroughTypeEnum {
+    // Throwaway default value.
+    NONE = 0;
+    // Indicates that the resource limits and requests should be passed through to the external workload.
+    // Be cautious about also setting apply_to_task=true since that will double the resources required for
+    // the task.
+    RESOURCES = 1;
+    // Indicates that the environment variables should be passed through to the external workload.
+    // It is generally safe to always set apply_to_task=true on this field.
+    ENV = 2;
+    // Indicates that the Kubernetes node affinity should be passed through to the external workload.
+    KUBERNETES_AFFINITY = 3;
+    // Indicates that the Kubernetes node tolerations should be passed through to the external workload.
+    KUBERNETES_TOLERATIONS = 4;
+    // Indicates that the Kubernetes node selector should be passed through to the external workload.
+    KUBERNETES_NODE_SELECTOR = 5;
+    // Indicates that the Kubernetes persistent volumes and ConfigMaps/Secrets mounted as volumes should be
+    // passed through to the external workload. Be sure that when setting apply_to_task=true, the volumes are
+    // ReadWriteMany or ReadOnlyMany or else the task's pod may not start.
+    // This is useful when the task prepares a shared volume for the external workload or defines output artifact
+    // (e.g. dsl.Model) that is created by the external workload.
+    KUBERNETES_VOLUMES = 6;
+  }
+}
+
+message TaskConfigPassthrough {
+  TaskConfigPassthroughType.TaskConfigPassthroughTypeEnum field = 1;
+  bool apply_to_task = 2;
 }
 
 // The spec of a pipeline task.
--- a/backend/src/v2/driver/container.go
+++ b/backend/src/v2/driver/container.go
@@ -27,6 +27,8 @@ import (
 	"github.com/kubeflow/pipelines/backend/src/v2/expression"
 	"github.com/kubeflow/pipelines/backend/src/v2/metadata"
 	pb "github.com/kubeflow/pipelines/third_party/ml-metadata/go/ml_metadata"
+	"google.golang.org/protobuf/encoding/protojson"
+	"google.golang.org/protobuf/types/known/structpb"
 )
 
 func validateContainer(opts Options) (err error) {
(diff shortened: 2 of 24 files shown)


backend/src/v2/driver/driver.go
```diff
@@ -308,17 +324,34 @@ func initPodSpecPatch(
...
+			Command: launcherCmd,
+			Args:    userCmdArgs,
+			Image:   containerImage,
+			Env:     userEnvVar,
 		}},
 	}
 
+	if taskK8sConfig != nil {
+		taskK8sConfig.Resources = res
+		// Set modest limits if the container is being used for passthrough
```
mprahl: Remove the limits being set similarly to driver.

sdk/python/kfp/dsl/task_kubernetes_config.py
```diff
@@ -0,0 +1,40 @@
...
+    """Kubernetes configurations for a task.
+
+    Annotate a component parameter with this type when you want the task's
+    runtime Kubernetes configuration to be forwarded to an external workload,
+    rather than applied to the task's own pod. This is useful when the task
+    launches another Kubernetes resource (for example, a Kubeflow Trainer job).
+
+    Not applied to the task's pod (a modest default is used for resources):
+    - resources
+    - image_pull_secrets
```
mprahl: I think this is incorrect

sdk/python/kfp/dsl/task_kubernetes_config.py
```diff
@@ -0,0 +1,40 @@
...
+import dataclasses
+from typing import Any, Dict, List, Optional
+
+
+@dataclasses.dataclass
+class TaskKubernetesConfig:
+    """Kubernetes configurations for a task.
+
+    Annotate a component parameter with this type when you want the task's
+    runtime Kubernetes configuration to be forwarded to an external workload,
```
mprahl: Show example pipeline in the docstring.

sdk/python/kfp/dsl/task_kubernetes_config.py
```diff
@@ -0,0 +1,117 @@
...
+    - affinity
+    - node_selector
+    - resources
+    - tolerations
+
+    Still applied to the task's pod:
+    - env
+    - image_pull_secrets
+    - volumes (used to share data between the task and the external workload)
+    - volume_mounts
```
HumairAK: Didn't we decide to make this configurable (or all the fields)? 

The `env` I don't care as much about, but the volume mounts can have some implications here with access mode limitations like ReadWriteOnce, so a user may not want this to apply to the task pod

sdk/python/kfp/dsl/executor.py
```diff
@@ -348,6 +349,22 @@ def execute(self) -> Optional[str]:
                     error_message=value.get('error', {}).get('message', None),
                 )
 
+            elif v == TaskKubernetesConfig:
+                # The backend injects this struct under the actual input parameter name.
+                # If missing, pass an empty structure.
+                value = self.get_input_parameter_value(k)
```
HumairAK, line 356: can we update the type returned on `get_input_parameter_value` to be `Optional[Union[str, Dict]` instead of `Optional[str]` ? since for both `TaskKubernetesConfig` and `PipelineTaskFinalStatus` value can be dict

backend/src/v2/driver/driver.go
```diff
@@ -72,6 +72,17 @@ type Options struct {
 	TaskName string // the original name of the task, used for input resolution
 }
 
+type TaskKubernetesConfig struct {
```
HumairAK: can we add an inline comment that this needs to be aligned to the sdk TaskKubernetesConfig dataclass

backend/src/v2/driver/container.go
```diff
@@ -182,6 +183,19 @@ func Container(ctx context.Context, opts Options, mlmd *metadata.Client, cacheCl
 		glog.Info("Cache disabled globally at the server level.")
 	}
 
+	var taskK8sConfig *TaskKubernetesConfig
+	taskK8sConfigInputs := map[string]bool{}
```
HumairAK: curious, what are we doing if we allow multiple component parameters of `TASK_KUBERNETES_CONFIG`? it really seems to only make sense to ever have 1, do we just allow the user to do it anyway or should we error on it (presumably at sdk compile time)

mprahl: I figured there's no good reason to have more than one but there's no good reason to error if the user has more than one so hence, the implementation you see lol.

api/v2alpha1/pipeline_spec.proto
```diff
@@ -442,9 +444,45 @@ message ParameterType {
...
 }
 
+
+// Represents the task configurations that can be passed through to an external workload.
+message TaskConfigPassthroughType {
+  enum TaskConfigPassthroughTypeEnum {
+    // Throwaway default value.
+    NONE = 0;
+    // Indicates that the resource limits and requests should be passed through to the external workload.
+    // Be cautious about also setting apply_to_task=true since that may double apply the
```
HumairAK:
```suggestion
    // Be cautious about also setting apply_to_task=true since that will double apply

```

sdk/python/kfp/dsl/component_decorator.py
```diff
@@ -13,24 +13,29 @@
...
+    target_image: Optional[str] = None,
+    packages_to_install: List[str] = None,
+    pip_index_urls: Optional[List[str]] = None,
+    output_component_file: Optional[str] = None,
+    install_kfp_package: bool = True,
+    kfp_package_path: Optional[str] = None,
+    pip_trusted_hosts: Optional[List[str]] = None,
+    use_venv: bool = False,
+    additional_funcs: Optional[List[Callable]] = None,
+    task_config_passthroughs: Optional[List[Union[TaskConfigPassthrough,
```
HumairAK, line 37: missing this parameter in the docstring

backend/src/v2/driver/driver_test.go
```diff
@@ -635,10 +662,65 @@ func Test_initPodSpecPatch_resourceRequests(t *testing.T) {
...
+
+	containerSpec := &pipelinespec.PipelineDeploymentConfig_PipelineContainerSpec{
+		Image:   "python:3.9",
+		Args:    []string{"--function_to_execute", "add"},
+		Command: []string{"sh", "-ec", "python3 -m kfp.components.executor_main"},
+		Resources: &pipelinespec.PipelineDeploymentConfig_PipelineContainerSpec_ResourceSpec{
+			CpuLimit:      2.0,
+			MemoryLimit:   1.5,
+			CpuRequest:    1.0,
+			MemoryRequest: 0.65,
```
HumairAK: can we use the non-deprecated fields?

HumairAK:
/lgtm
/approve

niceee

PR #12208, opened 2025-09-02
Revert "fix(backend): Guard against panic when MLMD execution publish…

droctothorpe:
…ing fails (#12203)"

This reverts commit a09fc030e9341553bafb68a14bb0789bac8b43b4 because it seems to be causing CI failures in the master branch.


droctothorpe:
Likely just CI flakiness. No need to revert.

PR #12215, opened 2025-09-04
feat(sdk): Add Docker runner support for KFP workspaces

VaniHaripriya:

This PR extends workspace functionality to DockerRunner, enabling Docker containers to access shared workspace storage during local pipeline execution. 

**Checklist:**

### DIFF
--- a/sdk/python/kfp/local/docker_task_handler_test.py
+++ b/sdk/python/kfp/local/docker_task_handler_test.py
@@ -93,55 +93,137 @@ class TestRunDockerContainer(DockerMockTestCase):
 class TestDockerTaskHandler(DockerMockTestCase):
 
     def test_get_volumes_to_mount(self):
-        handler = docker_task_handler.DockerTaskHandler(
-            image='alpine',
-            full_command=['echo', 'foo'],
-            pipeline_root=os.path.abspath('my_root'),
-            runner=local.DockerRunner(),
-        )
-        volumes = handler.get_volumes_to_mount()
-        self.assertEqual(
-            volumes, {
-                os.path.abspath('my_root'): {
-                    'bind': os.path.abspath('my_root'),
-                    'mode': 'rw'
-                }
-            })
-
-    def test_run(self):
-        handler = docker_task_handler.DockerTaskHandler(
-            image='alpine',
-            full_command=['echo', 'foo'],
-            pipeline_root=os.path.abspath('my_root'),
-            runner=local.DockerRunner(),
-        )
-
-        handler.run()
-        self.mocked_docker_client.containers.run.assert_called_once_with(
-            image='alpine:latest',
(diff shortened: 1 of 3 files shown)


sdk/python/kfp/local/docker_task_handler_test.py
```diff
@@ -93,55 +93,137 @@ def test_cwd_volume(self):
...
+                full_command=['echo', 'foo'],
+                pipeline_root=os.path.abspath('my_root'),
+                runner=local.DockerRunner(),
+            )
+            volumes = handler.get_volumes_to_mount()
+            self.assertEqual(
+                volumes, {
+                    os.path.abspath('my_root'): {
+                        'bind': os.path.abspath('my_root'),
+                        'mode': 'rw'
```
DharmitD, line 109: Hard-coded 'mode': 'rw' gives full read-write access to pipeline root. Consider if pipeline components really need write access to the entire directory structure, or if read-only would be safer for most use cases.

sdk/python/kfp/local/docker_task_handler_test.py
```diff
@@ -93,55 +93,137 @@ def test_cwd_volume(self):
...
+            )
+            volumes = handler.get_volumes_to_mount()
+
+            expected_volumes = {
                 os.path.abspath('my_root'): {
                     'bind': os.path.abspath('my_root'),
                     'mode': 'rw'
+                },
+                '/tmp/test-workspace': {
+                    'bind': '/kfp-workspace',
```
DharmitD, line 133: Hard-coded mount path `/kfp-workspace` could conflict with existing container paths. Should this be configurable or at least documented as a reserved path?

sdk/python/kfp/local/docker_task_handler_test.py
VaniHaripriya: We are using 'rw' mode so the pipeline components can write their outputs here and read inputs from other components.

sdk/python/kfp/local/docker_task_handler_test.py
VaniHaripriya: `/kfp-workspace` is the standerdized path for pipeline workspace as per the [KEP](https://github.com/kubeflow/pipelines/tree/master/proposals/11875-pipeline-workspace#container-driver). As per this, the driver will not allow any volume mount paths that conflict with the workspace mount path (`/kfp-workspace`).

sdk/python/kfp/local/docker_task_handler_test.py
mprahl: Agreed. For now, we don't distinguish between read write needs or read only needs for the workspace.

mprahl (approved):
/approve
/lgtm

PR #12220, opened 2025-09-05
feat(backend): Add mounted PVCs to the generated cache key

mprahl:

When a PVC is mounted, there are often side effects (read or write). Tasks further in the DAG may expect those side effects. This can lead to confusing behavior, so the safe approach is use the PVC name in the cache key. So if the PVC name is the same in the next run, caching can still work.

**Checklist:**

### DIFF
--- a/backend/test/v2/integration/cache_test.go
+++ b/backend/test/v2/integration/cache_test.go
@@ -313,7 +313,11 @@ func (s *CacheTestSuite) createRun(pipelineVersion *pipeline_upload_model.V2beta
 	require.Eventually(s.T(), func() bool {
 		pipelineRunDetail, err = s.runClient.Get(&runParams.RunServiceGetRunParams{RunID: pipelineRunDetail.RunID})
 
-		s.T().Logf("Pipeline %v state: %v", pipelineRunDetail.RunID, pipelineRunDetail.State)
+		if err == nil {
+			s.T().Logf("Pipeline %v state: %v", pipelineRunDetail.RunID, *pipelineRunDetail.State)
+		} else {
+			s.T().Logf("Pipeline %v state: %v", pipelineRunDetail.RunID, err.Error())
+		}
 
 		return err == nil && *pipelineRunDetail.State == expectedState
 	}, 2*time.Minute, 10*time.Second)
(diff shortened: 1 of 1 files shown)


backend/src/v2/driver/container.go
```diff
@@ -136,9 +136,40 @@ func Container(ctx context.Context, opts Options, mlmd *metadata.Client, cacheCl
...
+
+		if needsWorkspaceMount(execution.ExecutorInput) {
+			if opts.RunName == "" {
+				return execution, fmt.Errorf("failed to generate fingerprint: run name is required when workspace is used")
+			}
+
+			pvcNames = append(pvcNames, GetWorkspacePVCName(opts.RunName))
+		}
+
+		fingerPrint, cachedMLMDExecutionID, err := getFingerPrintsAndID(execution, &opts, cacheClient, pvcNames)
```
DharmitD, line 172: This seems inconsistent with what you have everywhere in the k8s.go file:

```
fingerPrint, cachedMLMDExecutionID, err := getFingerPrintsAndID(&execution, opts, cacheClient, nil)
```

Is there a reason why 'pvcNames' parameter is included here in container operations and excluded in PVC operations?

mprahl: Good question. I could change it but I left it the way it was because of this existing code in the PVC operations:

```go
		// Add pvcName to the executor input for fingerprint generation
		execution.ExecutorInput.Inputs.ParameterValues[pvcName] = structpb.NewStringValue(pvcName)
```

backend/test/v2/resources/pvc-mount.py
```diff
@@ -0,0 +1,52 @@
...
+@dsl.component
+def consumer(input_text: str) -> None:
+    with open('/data/file.txt', 'r') as f:
+        print(f.read())
+
+
+@dsl.pipeline(name='pvc-mount-pipeline')
+def pvc_mount_pipeline(pvc_name: str):
+    p = producer()
+    c = consumer(input_text='hello').after(p)
```
VaniHaripriya: I think we can remove the input_text parameter for consumer task since its not being used.

PR #12234, opened 2025-09-10
feat(sdk): add compile-time validation for workspace size

VaniHaripriya:

This PR adds validation to the SDK, so that the compilation fails if workspace feature is used without a workspace size being set in the PipelineConfig.

**Checklist:**


sdk/python/kfp/compiler/pipeline_spec_builder.py
```diff
@@ -2032,6 +2032,10 @@ def create_pipeline_spec(
 
     platform_spec = pipeline_spec_pb2.PlatformSpec()
     if pipeline_config is not None:
+        workspace_config = pipeline_config.workspace
+        if workspace_config is not None:
+            if not workspace_config.size or not workspace_config.size.strip():
```
DharmitD: The validation assumes workspace_config.size is either None or a string, but what if someone sets it to a different type?
For instance, setting it as an integer could cause an Attribute error.

The validation should probably be:

```
if not workspace_config.size or (isinstance(workspace_config.size, str) and not workspace_config.size.strip()):
    raise ValueError('Workspace is configured but size is not set.')
    ```

File: sdk/python/kfp/compiler/compiler_test.py [hunk 2]
```diff
@@ -4257,6 +4257,27 @@ def test_workspace_config_validation(self):
             workspace.set_size('   ')
         self.assertIn('required and cannot be empty', str(context.exception))
 
+    def test_compile_fails_when_workspace_config_has_empty_size(self):
```
DharmitD: You could also then add tests for these edge cases, in addition to the empty string case:

- None values
- Whitespace-only strings ("   ")
- Non-string types (integers, etc.)

sdk/python/kfp/compiler/compiler_test.py
```diff
@@ -4257,6 +4257,27 @@ def test_workspace_config_validation(self):
...
+        """Test that compilation should fail if workspace is configured without appropriate size."""
+        empty_sizes = ['', '   ', '\t', '\n', '  \t  \n  ']
+        for size in empty_sizes:
+            with self.subTest(size=repr(size)):
+                workspace_config = WorkspaceConfig(size='10Gi')
+                workspace_config.size = size
+                config = PipelineConfig(workspace=workspace_config)
```
mprahl: Could you add a test for when workspace is never set in the pipeline config? I imagine that will be the most common mistake.

sdk/python/kfp/compiler/pipeline_spec_builder.py
```diff
@@ -2032,6 +2032,10 @@ def create_pipeline_spec(
 
     platform_spec = pipeline_spec_pb2.PlatformSpec()
     if pipeline_config is not None:
+        workspace_config = pipeline_config.workspace
+        if workspace_config is not None:
```
mprahl: Sorry, I don't the original issue was clear.

What we'd like to solve is when a user leverages something like `dsl.WORKSPACE_PATH_PLACEHOLDER` or another workspace feature in their component but they didn't set the workspace size or set an invalid value.

mprahl: Sorry, I don't think the original issue was clear.

What we'd like to solve is when a user leverages something like `dsl.WORKSPACE_PATH_PLACEHOLDER` or another workspace feature in their component but they didn't set the workspace size or set an invalid value.

VaniHaripriya: Sure, I will update the changes.

DharmitD (commented):
/lgtm

sdk/python/kfp/compiler/pipeline_spec_builder.py
```diff
@@ -2058,6 +2073,9 @@ def create_pipeline_spec(
         platform_spec=platform_spec,
     )
 
+    # Validate workspace usage and configuration
+    _validate_workspace_requirements(pipeline_config, pipeline_spec)
```
mprahl: Could we directly look at the inputs instead? I was prompting GPT-5 about this and it suggested. I don't like how all the code is nested there but it does seem nice to look at the inputs directly.

```diff
diff --git a/sdk/python/kfp/dsl/graph_component.py b/sdk/python/kfp/dsl/graph_component.py
index 7b17ec8a2..1e31b7a89 100644
--- a/sdk/python/kfp/dsl/graph_component.py
+++ b/sdk/python/kfp/dsl/graph_component.py
@@ -18,7 +18,7 @@ from typing import Callable, Optional
 import uuid
 
 from kfp.compiler import pipeline_spec_builder as builder
-from kfp.dsl import base_component
+from kfp.dsl import WORKSPACE_PATH_PLACEHOLDER, base_component
 from kfp.dsl import pipeline_channel
 from kfp.dsl import pipeline_config
 from kfp.dsl import pipeline_context
@@ -63,6 +63,43 @@ class GraphComponent(base_component.BaseComponent):
         if not dsl_pipeline.tasks:
             raise ValueError('Task is missing from pipeline.')
 
+        # Detect workspace usage and validate PipelineConfig.workspace
+        # Workspace usage is indicated by the placeholder string token below,
+        # which is later resolved by the backend driver at runtime.
+        _uses_workspace = False
+        for _t in dsl_pipeline.tasks.values():
+            # 1) Scan task input values for the workspace token
+            _inputs = getattr(_t, '_inputs', {}) or {}
+            for _v in _inputs.values():
+                if isinstance(_v, str) and WORKSPACE_PATH_PLACEHOLDER in _v:
+                    _uses_workspace = True
+                    break
+            if _uses_workspace:
+                break
+            # 2) If container task, also scan command/args
+            _spec = getattr(_t, 'container_spec', None)
+            if _spec is not None:
+                _cmd = (_spec.command or [])
+                _args = (_spec.args or [])
+                if any(isinstance(x, str) and WORKSPACE_PATH_PLACEHOLDER in x for x in _cmd):
+                    _uses_workspace = True
+                if not _uses_workspace and any(
+                        isinstance(x, str) and WORKSPACE_PATH_PLACEHOLDER in x for x in _args):
+                    _uses_workspace = True
+            if _uses_workspace:
+                break
+
+        if _uses_workspace:
+            if self.pipeline_config is None or getattr(self.pipeline_config, 'workspace', None) is None:
+                raise ValueError(
+                    'Workspace features are used in this pipeline, but PipelineConfig.workspace is not set. '
+                    'Please set pipeline_config=dsl.PipelineConfig(workspace=dsl.WorkspaceConfig(size="10Gi")).')
+            _workspace_cfg = self.pipeline_config.workspace
+            # WorkspaceConfig enforces non-empty size in its constructor, but validate defensively.
+            if not getattr(_workspace_cfg, 'size', None):
+                raise ValueError(
+                    'Workspace features are used in this pipeline, but PipelineConfig.workspace.size is missing or empty.')
+
         # Making the pipeline group name unique to prevent name clashes with
         # templates
         pipeline_group = dsl_pipeline.groups[0]
```

sdk/python/kfp/dsl/graph_component.py
```diff
@@ -86,6 +92,106 @@ def __init__(
...
+
+        for task in tasks:
+            if hasattr(task, 'inputs') and task.inputs:
+                for input_value in task.inputs.values():
+                    if _check_value_for_workspace_placeholder(input_value):
+                        return True
+
+            if hasattr(task, 'args') and task.args:
+                if _check_value_for_workspace_placeholder(task.args):
+                    return True
```
mprahl: I don't think the type `PipelineTask` ever has an `.args` property, so we can remove this.

sdk/python/kfp/dsl/graph_component.py
```diff
@@ -86,6 +92,106 @@ def __init__(
...
+                if hasattr(task.container_spec,
+                           'args') and task.container_spec.args:
+                    if _check_value_for_workspace_placeholder(
+                            task.container_spec.args):
+                        return True
+                if hasattr(task.container_spec,
+                           'command') and task.container_spec.command:
+                    if _check_value_for_workspace_placeholder(
+                            task.container_spec.command):
+                        return True
```
mprahl: I think based on `needsWorkspaceMount` in `backend/src/v2/driver/driver.go`, it only looks at input parameters and since `task.inputs` would show something like this `{'workspace_path': '{{$.workspace_path}}'}`, we can just use `task.inputs`.

sdk/python/kfp/dsl/graph_component.py
```diff
@@ -86,6 +92,106 @@ def __init__(
...
+            tasks: List of pipeline tasks to check for workspace usage.
+
+        Returns:
+            True if any task uses workspace features, False otherwise.
+        """
+
+        def _check_value_for_workspace_placeholder(value: Any) -> bool:
+            """Recursively check a value for workspace placeholder usage."""
+            if isinstance(value, str):
+                return dsl.WORKSPACE_PATH_PLACEHOLDER in value
```
mprahl: This isn't documented behavior but based on the the `needsWorkspaceMount` function, it'd get mounted so we might as well check for it.
```suggestion
                return dsl.WORKSPACE_PATH_PLACEHOLDER in value or value.startswith(dsl.constants.WORKSPACE_MOUNT_PATH)
```

sdk/python/kfp/dsl/graph_component.py
```diff
@@ -86,6 +92,106 @@ def __init__(
...
+        if pipeline.groups:
+            workspace_used = self._detect_workspace_usage_in_groups(
+                pipeline.groups)
+
+        # If workspace features are used, ensure workspace is configured
+        if workspace_used:
+            if (pipeline_config is None or pipeline_config.workspace is None or
+                    not getattr(pipeline_config.workspace, 'size', None)):
+                raise ValueError(
+                    'Workspace features are used (e.g., dsl.WORKSPACE_PATH_PLACEHOLDER) but PipelineConfig.workspace.size is not set.'
```
mprahl, line 165:
```suggestion
                    'Workspace features are used (e.g. dsl.WORKSPACE_PATH_PLACEHOLDER) but PipelineConfig.workspace.size is not set.'
```

sdk/python/kfp/dsl/pipeline_config.py
```diff
@@ -63,6 +80,10 @@ def get_workspace(self) -> dict:
     def set_size(self, size: str):
```
mprahl, line 88: I think to prevent `size` from being set directly (e.g. `config.size = "invalid"`). We may have to have getter method for size leveraging the `@property` concept.

mprahl: Or perhaps converting `set_size` to a setter method and then have `set_size` just assign it directly

sdk/python/kfp/compiler/compiler_test.py
```diff
@@ -4257,6 +4267,34 @@ def test_workspace_config_validation(self):
             workspace.set_size('   ')
         self.assertIn('required and cannot be empty', str(context.exception))
 
+    def test_compile_fails_when_workspace_placeholder_used_without_workspace_config(
```
mprahl, line 4270: Could you add a test case for the nested groups case? An example might be `dsl.WORKSPACE_PATH_PLACEHOLDER` used inside of a `dsl.If`.

PR #12238, opened 2025-09-11
doc: Add a KEP for Jupyter notebook components

mprahl:

This proposal introduces a first-class `@dsl.notebook_component` decorator that lets users build Kubeflow Pipelines
(KFP) components directly from Jupyter notebooks. The decorator embeds a `.ipynb` file into the component and executes
it at runtime via `nbclient`, with parameters injected as a prepended cell. This provides a simple path for
notebook-centric workflows to run in KFP without requiring separate packaging or bespoke wrappers.

**Checklist:**

### DIFF
--- a/proposals/12238-notebook-component/README.md
+++ b/proposals/12238-notebook-component/README.md
@@ -1,0 +1,298 @@
+# KEP-12238: Jupyter Notebook Components
+
+<!-- toc -->
+
+- [Summary](#summary)
+- [Motivation](#motivation)
+  - [Goals](#goals)
+  - [Non-Goals](#non-goals)
+- [Proposal](#proposal)
+  - [Baseline Feature For Embedded Assets](#baseline-feature-for-embedded-assets)
+  - [SDK User Experience](#sdk-user-experience)
+    - [Example](#example)
+    - [notebook_component Decorator Arguments](#notebook_component-decorator-arguments)
+    - [Behavior Notes](#behavior-notes)
+  - [User Stories](#user-stories)
+  - [Notes/Constraints/Caveats](#notesconstraintscaveats)
+  - [Risks and Mitigations](#risks-and-mitigations)
+- [Design Details](#design-details)
+  - [Security Considerations](#security-considerations)
+  - [Test Plan](#test-plan)
+    - [Unit Tests](#unit-tests)
+    - [Integration tests](#integration-tests)
+  - [Graduation Criteria](#graduation-criteria)
+- [Implementation History](#implementation-history)
+- [Drawbacks](#drawbacks)
+- [Alternatives](#alternatives)
+<!-- /toc -->
+
+## Summary
+
+This proposal introduces a first-class `@dsl.notebook_component` decorator that lets users build Kubeflow Pipelines
+(KFP) components directly from Jupyter notebooks. The decorator embeds a `.ipynb` file into the component and executes
+it at runtime via `nbclient`, with parameters injected as a prepended cell. This provides a simple path for
+notebook-centric workflows to run in KFP without requiring separate packaging or bespoke wrappers.
+
+## Motivation
+
+Many users begin experimentation and development in Jupyter notebooks. Turning those notebooks into pipeline components
+currently requires boilerplate: exporting to Python, writing a wrapper function, or managing custom container images. A
+native notebook component:
+
+- Reduces friction to productionize notebook code in pipelines
+- Preserves the notebook as the source of truth while allowing parameterization
+- Avoids extra build steps by embedding notebook content into the component
+
+### Goals
+
+1. Enable defining a component from a `.ipynb` notebook with a single decorator.
+2. Support parameter injection into the notebook at execution time.
+3. Use the existing Python executor to keep compatibility with existing inputs/outputs concepts.
+
+### Non-Goals
+
+1. Notebook validation or linting beyond JSON and structural checks.
+2. Adding a new pipeline IR or backend executor. This reuses the Python executor.
+
+## Proposal
+
+### Baseline Feature For Embedded Assets
+
+This KEP establishes a baseline, generic capability for Python-function components to embed arbitrary files or
+directories directly into a lightweight component. The goal is to support cases where a component needs a small amount
+of read-only assets (configs, scripts, models, notebooks, etc.) without requiring a custom image.
+
+- Add a new decorator argument to `@dsl.component`:
+  - `embedded_artifact_path: Optional[str]` — path to a file or directory on the authoring machine to embed within the
+    component.
+- Add a new SDK type: `dsl.EmbeddedInput[T]` — a runtime-only input annotation that resolves to an artifact instance of
+  type `T` (e.g. `dsl.EmbeddedInput[dsl.Dataset]`) whose `path` points to the extracted embedded artifact root.
+  - If a directory is embedded, `.path` points to the extracted directory.
+  - If a single file is embedded, `.path` points to that file.
+
+Example:
+
+```python
+@dsl.component(
+    embedded_artifact_path="assets/config_dir",
+)
+def my_component(cfg: dsl.EmbeddedInput[dsl.Dataset], param: int):
+    # cfg.path is a directory when a directory is embedded; a file path if a file is embedded
+    print(cfg.path)
+```
+
+Execution model (lightweight components):
+
+- At compile time, the file/dir is archived (tar + gzip) and base64-embedded into the ephemeral module.
+- At runtime, the embedded artifact is made available by extracting to a temporary location and the `EmbeddedInput[...]`
+  parameter is injected as an artifact with `.path` pointing to the extracted file/dir in a temporary directory. The
+  extracted root is also added to `sys.path` for Python module resolution if it's a directory.
+- `sys.path` precedence: the embedded path/zip is prepended to `sys.path` (before existing entries) to ensure
+  deterministic use of embedded modules when names overlap.
+
+Relationship to `@dsl.notebook_component`:
+
+- The notebook component leverages the same embedding pattern, specializing the runtime helper to execute `.ipynb`
+  content via `nbclient`. If `notebook_path` is a directory, the `.ipynb` file in the directory is executed to allow for
+  utility functions to be embedded. Pipeline compilation fails if there is more than one `.ipynb` file.
+
+### SDK User Experience
+
+#### Example
+
+```python
+from kfp import dsl
+
+
+@dsl.notebook_component(
+    notebook_path="train.ipynb",
+    packages_to_install=["pandas", "scikit-learn", "nbclient"],
+)
+def train_from_notebook(dataset_uri: str, model: dsl.Output[dsl.Model]):
+    dsl.run_notebook(
+        dataset_uri=dataset_uri,
+        output_model_path=model.path,
+    )
+
+
+@dsl.pipeline(name="nb-pipeline")
+
+def pipeline(dataset_uri: str = "s3://bucket/dataset.csv"):
+    train_from_notebook(dataset_uri=dataset_uri)
+```
+
+#### Complex Example
+
+A mixed pipeline with a Python preprocessor, a notebook training step, and a notebook evaluation step:
+
+```python
+from kfp import dsl
+
+
+@dsl.component(
+    base_image="python:3.11-slim",
+    packages_to_install=["pandas==2.2.2"],
+)
+def preprocess(text: str, cleaned_text: dsl.Output[dsl.Dataset]):
+    """Cleans whitespace from input text and writes to cleaned_text."""
+    import re
+
+    cleaned = re.sub(r"\s+", " ", text).strip()
+    with open(cleaned_text.path, "w", encoding="utf-8") as f:
+        f.write(cleaned)
+
+
+@dsl.notebook_component(
+    notebook_path="dev-files/nb_train.ipynb",
+    base_image="registry.access.redhat.com/ubi9/python-311:latest",
+)
+def train_model(
+    cleaned_text: dsl.Input[dsl.Dataset],
+    model: dsl.Output[dsl.Model],
+):
+    """Trains a model from cleaned text and writes model."""
+    import shutil
+
+    # Read the dataset for the notebook since it expects a string
+    with open(cleaned_text.path, "r", encoding="utf-8") as f:
+        cleaned_text_str = f.read()
+
+    # Execute the embedded notebook with kwargs injected as variables
+    dsl.run_notebook(cleaned_text=cleaned_text_str)
+
+    # Translate notebook outputs into KFP outputs
+    nb_model_dir = "/tmp/kfp_nb_outputs/model_dir"
+    shutil.copytree(nb_model_dir, model.path, dirs_exist_ok=True)
+
+
+@dsl.notebook_component(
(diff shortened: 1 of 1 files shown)


proposals/12238-notebook-component/README.md
```diff
@@ -0,0 +1,248 @@
...
+- `notebook_path: str` – New parameter specifying the `.ipynb` file to embed and execute (required).
+- `packages_to_install: Optional[List[str]]` – Same as Python executor except when `None` (default here) it additionally
+  installs `jupyter>=1,<2` and `nbconvert>=7,<8` to avoid major-version bumps; `[]` installs nothing; a non-empty list
+  installs exactly the provided packages.
+
+All other decorator arguments and behaviors are identical to the Python executor.
+
+#### Behavior Notes
+
+- The notebook JSON is compressed using gzip and base64-encoded before embedding into the ephemeral Python module used
```
droctothorpe, line 210: This is exactly what we're doing, albeit in a more general way. It's a very effective pattern.

proposals/12238-notebook-component/README.md
```diff
@@ -0,0 +1,248 @@
...
+
+- Embedded notebooks can bloat the command payload and slow compilation/execution for large files, though gzip
+  compression typically helps.
+- Notebooks are less modular than Python modules for code reuse and testing.
+
+## Alternatives
+
+1. Use `@dsl.component` with manual `nbconvert` calls inside the function. This requires boilerplate and manual
+   packaging of the notebook.
+2. Pre-build a container image containing the notebook and its dependencies, then use `@dsl.container_component`. This
```
droctothorpe, line 297: Having to build and push a new image every time you change one cell would result in a great deal of development friction. The approach described in this KEP is so much smoother.

droctothorpe:
This is a great idea. Data scientists will love it. 

/lgtm

proposals/12238-notebook-component/README.md
droctothorpe: It may be helpful to generalize the file (and wholesale directory) injection pattern and make notebook injection a specific subset of that, since the former can be useful in other contexts.

zazulam: I like the idea of wholesaling a directory but I think it would probably be better for a workspace pvc utility.

mprahl:
Thanks for the reviews! I added a more generic option to the component decorator as well. I got a proof of concept working so it seems doable. I'd like your input on this. :smile:

proposals/12238-notebook-component/README.md
```diff
@@ -0,0 +1,291 @@
...
+native notebook component:
+
+- Reduces friction to productionize notebook code in pipelines
+- Preserves the notebook as the source of truth while allowing parameterization
+- Avoids extra build steps by embedding notebook content into the component
+
+### Goals
+
+1. Enable defining a component from a `.ipynb` notebook with a single decorator.
+2. Support parameter injection into the notebook at execution time.
```
andreyvelich, line 49: Is there any interest to integrate with [`papermil`](https://github.com/nteract/papermill) to support Notebook execution within KFP component ?
It does support input parameters, artifacts, and more which might be useful for KFP context.

We are using it in our Trainer E2Es: https://github.com/kubeflow/trainer/blob/master/hack/e2e-run-notebook.sh#L48-L49

proposals/12238-notebook-component/README.md
```diff
@@ -0,0 +1,291 @@
...
+    bundled_artifact_path="assets/config_dir",
+)
+def my_component(cfg: dsl.BundledInput[dsl.Dataset], param: int):
+    # cfg.path is a directory when a directory is bundled; a file path if a file is bundled
+    print(cfg.path)
+```
+
+Execution model (lightweight components):
+
+- At compile time, the file/dir is archived (tar + gzip) and base64-embedded into the ephemeral module.
```
andreyvelich, line 86: Would it allow us to support imports from Python files which are in the same users workspace ?
E.g. if I have a Notebook, but want to import util functions from another Python file, that I create.

proposals/12238-notebook-component/README.md
```diff
@@ -0,0 +1,291 @@
...
+
+#### Behavior Notes
+
+- The notebook JSON is compressed using gzip and base64-encoded before embedding into the ephemeral Python module used
+  by the Python executor. This reduces command line length and allows for larger notebooks.
+- At runtime, `dsl.run_notebook(**kwargs)` is bound to a helper that:
+  1. Decompresses and parses the embedded notebook into memory
+  2. Injects a cell at the start of the notebook with the input parameters passed to `dsl.run_notebook`
+  3. Executes via `nbconvert.ExecutePreprocessor`
+  4. Streams cell outputs (stdout/stderr and `text/plain` displays)
```
andreyvelich, line 219: @akshaychitneni @zsailer Do you know if we directly use `papermil` to execute the notebook, will it be more memory efficient compare to nbconvert ?

proposals/12238-notebook-component/README.md
zazulam: I believe [Elyra](https://github.com/elyra-ai/elyra) is currently using papermill for executing notebook components in DAGs as well. My only concern with papermill is that the repo seems to be not as active as it used to be.

mprahl: Thanks for the input @andreyvelich! I updated the design document to have the same behavior as papermill for input parameter injection.

I think for artifact handing, I think it makes sense to lean on KFP's existing concepts around that.

proposals/12238-notebook-component/README.md
mprahl: Good point. Added:
If `notebook_path` is a directory, the `.ipynb` file in the directory is executed to allow for utility functions to be bundled. Pipeline compilation fails if there is more than one `.ipynb` file.

proposals/12238-notebook-component/README.md
lresende: Papermil was actually using nbconvert in the past, but checking, they have migrated to more efficient component that we might also investigate? https://github.com/nteract/papermill/blob/53847310586b74fdb2763fe333ab99babee367ed/docs/changelog.md?plain=1#L129

proposals/12238-notebook-component/README.md
lresende: This is mostly the case for this, and this is also how Elyra is supporting this use case.

proposals/12238-notebook-component/README.md
mprahl: It looks like papermill uses nbclient which is lighter weight than nbconvert. I updated the KEP to use nbclient directly.

proposals/12238-notebook-component/README.md
```diff
@@ -0,0 +1,295 @@
...
+
+#### Example
+
+```python
+from kfp import dsl
+
+
+@dsl.notebook_component(
+    notebook_path="train.ipynb",
+    packages_to_install=["pandas", "scikit-learn", "nbconvert", "jupyter"],
```
lresende: Maybe have an index URL as an optional parameter that defaults to public pypi? This allows using private/internal repositories

mprahl: Agreed. `notebook_component` is wrapping the `component` decorator, so it'll have the `pip_index_urls` argument available:
https://github.com/kubeflow/pipelines/blob/ffaeeda3da4f68c6b39231828ab23ff5f2a367e7/sdk/python/kfp/dsl/component_decorator.py#L30

proposals/12238-notebook-component/README.md
shravan-achar: Also, the source of the file/dir must be considered. If running this on a remote server, the files/dir must be local to that remote server. It would be nice to abstract or remove this requirement that authoring machine must have the files.

One option is to prepend different prefixes (protocols) with paths such as s3://, local://, ftp://, git:// etc.

proposals/12238-notebook-component/README.md
```diff
@@ -0,0 +1,295 @@
...
+)
+def my_component(cfg: dsl.BundledInput[dsl.Dataset], param: int):
+    # cfg.path is a directory when a directory is bundled; a file path if a file is bundled
+    print(cfg.path)
+```
+
+Execution model (lightweight components):
+
+- At compile time, the file/dir is archived (tar + gzip) and base64-embedded into the ephemeral module.
+- At runtime, the bundle is safely extracted to a temporary location and the `BundledInput[...]` parameter is injected
```
shravan-achar: Python natively supports importing from zip archives without extraction. Would the extraction be needed for something else?

proposals/12238-notebook-component/README.md
```diff
@@ -0,0 +1,295 @@
...
+  as an artifact with `.path` pointing to the extracted file/dir in a temporary directory.
```
shravan-achar: Also, does the path need to be handled in a way that it is considered before all the system paths? Or should it be after all the system paths?

proposals/12238-notebook-component/README.md
mprahl: Interesting, I didn't know that! I think to avoid detecting if it's a Python module or some data directory, we can just stick extracting it to a temporary directory and prepending `sys.path` to the extracted location.

proposals/12238-notebook-component/README.md
mprahl: Good point! I added a bit about prepending `sys.path` with the directory if the bundled artifact/notebook path at compile time is a directory.

proposals/12238-notebook-component/README.md
```diff
@@ -0,0 +1,301 @@
...
+```
+
+Execution model (lightweight components):
+
+- At compile time, the file/dir is archived (tar + gzip) and base64-embedded into the ephemeral module.
+- At runtime, the bundle is made available by extracteing to a temporary location and the `BundledInput[...]` parameter
+  is injected as an artifact with `.path` pointing to the extracted file/dir in a temporary directory. The extracted
+  root is also added to `sys.path` for Python module resolution if it's a directory.
+- `sys.path` precedence: the bundled path/zip is prepended to `sys.path` (before existing entries) to ensure
+  deterministic use of bundled modules when names overlap.
```
HumairAK: but a BundledInput isn't just limited to python code right (for example in your example you have a dsl.Dataset as the bundled input)? so shouldn't this be optional?

proposals/12238-notebook-component/README.md
```diff
@@ -0,0 +1,301 @@
...
+  - If a directory is bundled, `.path` points to the extracted directory.
+  - If a single file is bundled, `.path` points to that file.
+
+Example:
+
+```python
+@dsl.component(
+    bundled_artifact_path="assets/config_dir",
+)
+def my_component(cfg: dsl.BundledInput[dsl.Dataset], param: int):
```
HumairAK: Do you think `BundledInput` naming might be confusing, given that we can already have Artifacts that are themselves a directory and not a file?

proposals/12238-notebook-component/README.md
```diff
@@ -0,0 +1,301 @@
...
+- Add a new SDK type: `dsl.BundledInput[T]` — a runtime-only input annotation that resolves to an artifact instance of
+  type `T` (e.g. `dsl.Dataset`) whose `path` points to the extracted bundle root.
+  - If a directory is bundled, `.path` points to the extracted directory.
+  - If a single file is bundled, `.path` points to that file.
+
+Example:
+
+```python
+@dsl.component(
+    bundled_artifact_path="assets/config_dir",
```
HumairAK: I can't help but think, what we are really asking for here is a way to do importer that is locally sourced, and that the right place to do this is to do dsl.importer(file_path="assets/config_dir") and then have something like: 

```python
@dsl.component() 
def train_model(my_notebook: dsl.LocalInput[dsl.Notebook]): 
    my_notebook.run(my_param=..) 

@dsl.pipeline() 
def my_pipeline(): 
    importer = dsl.importer(file_path="assets/notebook", artifact_class=dsl.Notebook)
    use_nb_task = use_nb(notebook_input=importer.output)
    
```

Then this makes the design extensible so that in the future we can also just do a dsl.importer on a notebook artifact that is externally sourced rather than local. Other things we could do with this approach is allow the option to "upload" this local artifact into the object store so other components, in other runs, can also import it.

Have you given thought to this method, is this more complicated than doing it via the current approach?

proposals/12238-notebook-component/README.md
```diff
@@ -0,0 +1,301 @@
...
+  3. Executes via `nbclient.NotebookClient`
+  4. Streams cell outputs (stdout/stderr and `text/plain` displays)
+
+For the baseline bundling feature:
+
+- Files/directories are archived (tar+gz) and base64-embedded similarly; by default they are extracted at import time to
+  satisfy the `BundledInput[...]` contract of providing a real filesystem `.path` and to support non-Python assets.
+- For Python import resolution, the bundled path (extracted root or zip archive) is prepended into `sys.path` if it's a
+  directory.
+- `dsl.BundledInput[T]` is not part of the component interface; it is injected at runtime and provides an artifact with
```
HumairAK: Would this `dsl.BundledInput[T]` be of type dsl.Artifact then?

proposals/12238-notebook-component/README.md
```diff
@@ -0,0 +1,301 @@
...
+    # Copy notebook-generated metrics into output parameter
+    with open("/tmp/kfp_nb_outputs/metrics.json", "r") as f_in, open(metrics_json, "w") as f_out:
+        import json
+        json.dump(json.load(f_in), f_out)
+
+
+@dsl.pipeline(name="three-step-nb-mix")
+def pipeline(text: str = "Hello   world"):
+    p = preprocess(text=text).set_caching_options(False)
+    t = train_model(cleaned_text=p.outputs["cleaned_text"]).set_caching_options(False)
```
HumairAK: it probably goes without saying, but any `env` set on the tasks should pass through to the notebook runs, yes?

proposals/12238-notebook-component/README.md
mprahl: Is there harm in always setting it? I suppose we could detect if there's a .py file.

proposals/12238-notebook-component/README.md
mprahl: `T` is just the placeholder we use in the code for any artifact type (could be a dataset, artifact, etc)..

mprahl: Clarified in the latest push.

proposals/12238-notebook-component/README.md
mprahl: Discussed offline and I think this is a better approach for when/if we add functionality to upload artifacts to object store alongside your pipeline version. Renamed to `EmbeddedInput` instead of `BundledInput` to distinguish the behavior.

proposals/12238-notebook-component/README.md
mprahl: Renamed to `EmbeddedInput`.

proposals/12238-notebook-component/README.md
mprahl: Yes, it does!

proposals/12238-notebook-component/README.md
HumairAK: discussed offline, we deemed there's no harm in always setting this

HumairAK:
/lgtm
/approve

EXCITING

PR #12242, opened 2025-09-14
feat: Make artifact proxy opt-in via ARTIFACTS_PROXY_ENABLED configuration

akagami-harsh:
**Description of your changes:**
- Part of https://github.com/kubeflow/pipelines/issues/12239#issuecomment-3284715990
- **Add `ARTIFACTS_PROXY_ENABLED` config**: New environment variable in `pipeline-install-config` (defaults to `"false"`)

**Checklist:**
- [ ] You have [signed off your commits](https://www.kubeflow.org/docs/about/contributing/#sign-off-your-commits)
- [ ] The title for your pull request (PR) should follow our title convention. [Learn more about the pull request title convention used in this repository](https://github.com/kubeflow/pipelines/blob/master/CONTRIBUTING.md#pull-request-title-convention).

PR #12246, opened 2025-09-15
Revert "feat: Make artifact proxy opt-in via ARTIFACTS_PROXY_ENABLED configuration"

akagami-harsh:
Reverts kubeflow/pipelines#12242


akagami-harsh:
@HumairAK I added the triggers with the current changes, but the tests that were failing on the master branch don’t seem to fail now. Should I revert even if the tests are passing?

HumairAK:
hey @akagami-harsh apologies, this looks like it was due to flaky failures, re-running the CI it's not passing

PR #12266, opened 2025-09-22
fix(ci): correct PROXY check in workflow files

### LINKED ISSUE #12265: Fix incorrect PROXY check in workflow files causing CI failures
Problem:
In multiple workflow files (e.g., api-server-tests.yml), the PROXY check is written as:

if [ -n PROXY ]; then
    PROXY=false
fi

This does not correctly evaluate the variable, causing workflows to fail when PROXY is expected to be set.

Expected behavior:
The check should correctly reference the variable using quotes:

if [ -n "$PROXY" ]; then
    PROXY=false
fi

Impact:
Incorrect evaluation of PROXY causes intermittent CI failures. Fixing this ensures workflows run reliably.

aniketpati1121:
Problem: 
In multiple workflow files (like api-server-tests.yml), the PROXY check is written as:

if [ -n PROXY ]; then
    PROXY=false
fi

This is incorrect because it does not properly evaluate the variable, causing the workflows to fail when the PROXY variable is expected to be set.

Expected behavior:
The check should correctly reference the variable using quotes:

if [ -n "$PROXY" ]; then
    PROXY=false
fi

Impact:
CI jobs fail intermittently due to incorrect evaluation of the PROXY variable.

fixes #12265


nsingla (approved):
lgtm

zazulam (commented):
/approve

aniketpati1121:
Thank you all!  
This is my first merged PR in Kubeflow and it feels great 
Looking forward to learning and contributing more 🚀

PR #12082, opened 2025-07-28
feat(backend/frontend): Implement TLS for apiserver HTTP/GRPC

alyssacgoins:
**Description of your changes:**
Implement an option for pod-to-pod TLS protocol at the API server and launcher levels. PR includes an e2e test run with TLS enabled.
**Checklist:**
- [X] You have [signed off your commits](https://www.kubeflow.org/docs/about/contributing/#sign-off-your-commits)
- [X] The title for your pull request (PR) should follow our title convention. [Learn more about the pull request title convention used in this repository](https://github.com/kubeflow/pipelines/blob/master/CONTRIBUTING.md#pull-request-title-convention).

PR #12269, opened 2025-09-22
Central driver POC 

ntny:

POC for https://github.com/kubeflow/pipelines/pull/12023
- Modify Argo compiler: generate a plugin template instead of a container


droctothorpe:
This is EPIC, @ntny! Can't wait to try it out.

ntny:
Hi @HumairAK @droctothorpe  would you mind giving this a try?
It should be pretty straightforward to run the cluster with the agent only without the driver by following the[ instructions](https://github.com/kubeflow/pipelines/pull/12269#issue-3442114495) above.

ntny:
Hi! @nsingla  I made intentional changes to the compiler, and manually updating all specs in test/compiled-workflow would be very time-consuming.
I’ve already used the [following code](https://github.com/kubeflow/pipelines/pull/12269/files#diff-0e968b838baa6b52a64936e66c5f0c50332ebb260d19327ee342a3480f4b1fd0R133) on my side to regenerate specs directly from the test using a special flag (similar to snapshot tests) and then review the diff manually. 
Do you have any concerns about this approach, given your experience with test code and test practices?

nsingla:
> Hi! @nsingla I made intentional changes to the compiler, and manually updating all specs in test/compiled-workflow would be very time-consuming. I’ve already used the [following code](https://github.com/kubeflow/pipelines/pull/12269/files#diff-0e968b838baa6b52a64936e66c5f0c50332ebb260d19327ee342a3480f4b1fd0R133) on my side to regenerate specs directly from the test using a special flag (similar to snapshot tests) and then review the diff manually. Do you have any concerns about this approach, given your experience with test code and test practices?

You don;t need to update it manually, you can run the compiler tests locally with flag:
`ginkgo -v -- -updateCompiledFiles=true`
this should update the workflows

droctothorpe:
Hey, @ntny . Unfortunately, I won't have bandwidth to validate it in the next two weeks but just wanted to let you know that it's on my radar and I will get to it as soon as I can. Maybe someone else will get to it before me. VERY excited about this. Kudos!

ntny:
> Hey, @ntny . Unfortunately, I won't have bandwidth to validate it in the next two weeks but just wanted to let you know that it's on my radar and I will get to it as soon as I can. Maybe someone else will get to it before me. VERY excited about this. Kudos!

Hi, thanks! Sure, absolutely no rush!

droctothorpe:
@ntny, I'm hoping to review and validate this PR this sprint. Can you resolve the conflicts when you have a chance? Thank you and thanks for your patience!

droctothorpe:
Addressed the merge conflicts in a fork and working on validating locally now.

ntny:
@droctothorpe  Sorry for the late reply!
I’ll fix the conflicts by Monday.
Ping me if you need anything.

ntny:
Hi, @droctothorpe !
Pulled the updates. There’s a [MR](https://github.com/kubeflow/pipelines/pull/12572) that hits the plugin limitations.

In this PR, the volume with certificates is [mounted](https://github.com/kubeflow/pipelines/pull/12572/files#diff-88badacfd78f66dfe792bec3763d8564798463dc822b92603a6ae50d990eb317L80) at pipeline compilation via driver.controller.volumes.
The plugin is a global entity with separate settings that do not support managing volumes at the level of a specific compilation.

Possible options:

Dynamic volume management during compilation would require a separate design and discussion.

If volume settings are static, they can be moved to the global level, like the other plugin settings.

In the current POC, tests involving certificates are temporarily skipped.

droctothorpe:
@alyssacgoins curious to hear if you have any thoughts on how to resolve this.

droctothorpe:
> The plugin is a global entity with separate settings that do not support managing volumes at the level of a specific compilation.

Can you elaborate on this, @ntny?

My understanding based on the linked PR  is that it's set globally at the platform level and not variable from one run / component to another. I think that from the compiler's perspective, the volume settings are for all intents and purposes static. 

Apologies if I'm missing something.

ntny:
@droctothorpe Yes, you understood it correctly 🙂
And I agree from the compiler’s perspective it looks static, though it may need double-checking.

ntny:
Recently discovered one more limitation of the global driver:

### Current state:

Recurrent Runs work fine.

Each Run is pinned to a specific driver/launcher image, so updating the global driver/launcher doesn’t break existing Runs.

### Potential problem:

If we move to a global Argo Workflow executor plugin (one driver for all workflows), a new driver/launcher release could be incompatible with existing recurrent Runs.

These Runs are pinned to old image versions of launchers and might stop working with the new global driver.

### Idea / solution:

We might need the ability to configure the driver per workflow spec instead of globally.

That way, we can update the global driver without breaking existing recurrent Runs.

I’ve also created an issue in the Argo Workflows community to get feedback on this proposal : [15234](https://github.com/argoproj/argo-workflows/issues/15234)

@HumairAK, @droctothorpe  cc

droctothorpe:
@ntny, I think the specified edge case exists in KFP in its current state before this PR. 

Either way, I don't think it should block this merge.

I think that the last hurdle is just making sure that the e2e tests in CI in this branch are validating these changes (i.e. referencing the updated driver image).

Also, I'm open to reducing the AWF requeue time to 5s in this PR. Curious to hear if you have any thoughts on that.

droctothorpe:
Going to run a test on the master branch to validate if the edge-case pre-exists this PR.

droctothorpe:
## Edge-case validation notes

### Setup

```bash
git checkout master
kubectl apply -k ./manifests/kustomize/env/platform-agnostic
```

### The recurring run

```python
from kfp import dsl
from kfp import Client
from kfp import compiler


@dsl.component()
def comp():
    print("foo")


@dsl.pipeline
def pipeline():
    comp().set_caching_options(enable_caching=False)

compiler.Compiler().compile(pipeline, "recurring-run.yaml")

client = Client()
client.create_recurring_run(
    experiment_id="ec535e9e-7d80-49da-a2b3-83e30a853aba",
    job_name="validation",
    pipeline_package_path="recurring-run.yaml",
    cron_expression="* * * * *", # Every minute.
)
```

Run completed successfuly:
    
```
NAME↑                           STATUS
validation5tj4q-1-604852182     Succeeded
```

Driver is set to `2.15.0` in the API:

```bash
pipelines master* ⇣ kdd ml-pipeline | grep -i v2_driver
      V2_DRIVER_IMAGE:                       ghcr.io/kubeflow/kfp-driver:2.15.0
```

The driver pods are referencing this image, e.g.:

```
kdp validation5tj4q-1-604852182-system-dag-driver-712789934 | grep -i image: | tail -1
    Image:         ghcr.io/kubeflow/kfp-driver:2.15.0
```

### Update backend

Now let's update the `V2_DRIVER_IMAGE` env var in the API deployment then kick off another execution of the existing recurring run.

Updating to `2.14.4`. 

Confirming that the resulting pod from the pre-existing recurring run references the updated driver image.

```
❯ pipelines master* ⇣ kdp validation5tj4q-2-588074563-system-dag-driver-3124215983 | grep -i image: | tail -1
    Image:         ghcr.io/kubeflow/kfp-driver:2.14.4
```

This is happening in master, not this PR branch, **which confirms that this edge-case is not introduced by this PR**.

backend/src/v2/compiler/argocompiler/argo.go
```diff
@@ -454,7 +452,7 @@ var driverResources = k8score.ResourceRequirements{
 // Launcher only copies the binary into the volume, so it needs minimal resources.
 var launcherResources = k8score.ResourceRequirements{
 	Limits: map[k8score.ResourceName]k8sres.Quantity{
-		k8score.ResourceMemory: k8sres.MustParse("128Mi"),
+		k8score.ResourceMemory: k8sres.MustParse("256Mi"),
```
droctothorpe: For a separate PR I think but we should really make this configurable.

backend/src/v2/compiler/argocompiler/container.go
```diff
@@ -179,62 +166,61 @@ func (c *workflowCompiler) containerDriverTask(name string, inputs containerDriv
 		cached:       taskOutputParameter(name, paramCachedDecision),
 		condition:    taskOutputParameter(name, paramCondition),
 	}
-	return dagTask, outputs
+	return dagTask, outputs, nil
 }
 
-func (c *workflowCompiler) addContainerDriverTemplate() string {
+func (c *workflowCompiler) addContainerDriverTemplate() (string, error) {
```
droctothorpe, line 170: A docstring here clarifying that this function adds an executor plugin would be nice I think.

backend/src/v2/compiler/argocompiler/dag.go
```diff
@@ -295,6 +292,9 @@ func (c *workflowCompiler) task(name string, task *pipelinespec.PipelineTaskSpec
 				kubernetesConfig: kubernetesConfigPlaceholder,
 				taskName:         name,
 			})
+			if err != nil {
```
droctothorpe, line 307: I appreciate the improved error handling and general cleanup and consolidation throughout.

backend/src/v2/compiler/argocompiler/plugin.go
```diff
@@ -0,0 +1,37 @@
...
+package argocompiler
+
+import (
+	"encoding/json"
+	"fmt"
+
+	wfapi "github.com/argoproj/argo-workflows/v3/pkg/apis/workflow/v1alpha1"
+)
+
+func driverPlugin(params map[string]interface{}) (*wfapi.Plugin, error) {
```
droctothorpe, line 26: This is a nit, but I think some more doc strings for some of the net new functions would be helpful. Try to put yourself in the shoes of someone else looking at this 10 months down the line who has no idea what an executor plugin is.

backend/test/compiler/argo_ginkgo_test.go
```diff
@@ -64,16 +64,18 @@ var _ = Describe("Verify Spec Compilation to Workflow >", Label(POSITIVE, Workfl
...
-			compilerOptions:   argocompiler.Options{CacheDisabled: false},
-			envVars:           map[string]string{"CABUNDLE_CONFIGMAP_NAME": "test-configmap-name", "CABUNDLE_KEY_NAME": "test-configmap-key"},
-			pipelineFilePaths: []string{filepath.Join(pipelineFilesRootDir, pipelineDirectory, "mounted_cabundle_configmap.yaml")},
-		},
-		{
-			compilerOptions:   argocompiler.Options{CacheDisabled: false},
-			envVars:           map[string]string{"CABUNDLE_SECRET_NAME": "test-secret-name"},
-			pipelineFilePaths: []string{filepath.Join(pipelineFilesRootDir, pipelineDirectory, "mounted_cabundle_secret.yaml")},
-		},
+		// WARN: Tests are temporarily ignored because volume mount management
```
droctothorpe: Are you planning to re-enable these since the cert is set globally by the API and therefore be propagated down to the driver plugin?

ntny: Sure, I’ve seen all the comments and will address them. Initially, I intended this MR more as a quick POC, just to test the idea.

backend/test/compiler/utils/workflow_utils.go
```diff
@@ -89,6 +91,31 @@ func CreateCompiledWorkflowFile(compiledWorflow *v1alpha1.Workflow, compiledWork
 	return testutil.CreateFile(compiledWorkflowFilePath, [][]byte{fileContents})
 }
 
+func ConfigurePluginSettings(workflow *v1alpha1.Workflow, remove bool) *v1alpha1.Workflow {
```
droctothorpe, line 95: Can you add a docstring / comment clarifying why this is necessary? Thanks!

manifests/kustomize/base/pipeline/ml-pipeline-apiserver-deployment.yaml
```diff
@@ -129,7 +129,7 @@ spec:
             # JSON patch to apply to compiled workflow specifications
             - name: COMPILED_PIPELINE_SPEC_PATCH
               value: "{}"
-          image: ghcr.io/kubeflow/kfp-api-server:dummy
+          image: ntny/kfp-api-server:beta-poc04
```
droctothorpe: This goes without saying but this obviously can't be merged with this. All these images are built from scratch in CI [here](https://github.com/kubeflow/pipelines/blob/master/.github/workflows/image-builds-with-cache.yml) and deployed to the cluster [here](https://github.com/kubeflow/pipelines/blob/master/.github/actions/deploy/action.yml#L59-L73). @nsingla can you confirm? Both of those steps are part of the [e2e test workflow](https://github.com/kubeflow/pipelines/blob/master/.github/workflows/e2e-test.yml) so you should be able to just leave the image as is.

manifests/kustomize/base/pipeline/ml-pipeline-driver-plugin-cm.yaml
```diff
@@ -0,0 +1,23 @@
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  labels:
+    workflows.argoproj.io/configmap-type: ExecutorPlugin
+  name: ml-pipeline-driver-agent
+data:
+  sidecar.automountServiceAccountToken: "true"
+  sidecar.container: |
```
droctothorpe, line 9: Do you mind adding a comment here explaining why it needs to be a sidecar? Thanks!

manifests/kustomize/base/pipeline/ml-pipeline-driver-plugin-cm.yaml
```diff
@@ -0,0 +1,23 @@
...
+    name: driver-plugin
+    ports:
+    - containerPort: 8080
+    resources:
+      limits:
+        cpu: "1"
+        memory: 1Gi
+      requests:
+        cpu: 250m
+        memory: 512Mi
```
droctothorpe, line 32: Have you profiled this at all? I'm worried about scenarios where someone runs many steps in parallel, each of which needs to hit the central driver agent.

manifests/kustomize/base/pipeline/pipeline-runner-role.yaml
```diff
@@ -85,3 +89,44 @@ rules:
...
+
+---
+apiVersion: rbac.authorization.k8s.io/v1
+kind: Role
+metadata:
+  name: configmap-reader
+rules:
+  - apiGroups: [""]
+    resources: ["configmaps"]
+    verbs: ["get", "list", "watch"]  # права на чтение
```
droctothorpe:
```suggestion
    verbs: ["get", "list", "watch"]

File: manifests/kustomize/base/pipeline/ml-pipeline-apiserver-deployment.yaml (hunk 7 above)
Comment (ntny) [reply manifests/kustomize/base/pipeline/ml-pipeline-apiserver-deployment.yaml]: Yes, absolutely. This was done purely to make it easier to run a local PoC. I’ll handle this along with the remaining changes.

File: manifests/kustomize/env/dev/kustomization.yaml [hunk 11]
```diff
@@ -9,7 +9,7 @@ resources:
 - ../gcp/inverse-proxy
 
 images:
-- name: ghcr.io/kubeflow/kfp-api-server
+- name: ghcr.io/kubeflow/kfp-driver
```
droctothorpe, line 14: Why do we no longer need to specify the kfp-api-server image?

manifests/kustomize/third-party/application/application-controller-deployment.yaml
```diff
@@ -31,8 +31,8 @@ spec:
           resources:
             limits:
               cpu: 100m
-              memory: 30Mi
+              memory: 128Mi
```
droctothorpe: Please provide context on why these had to bumped. Thanks!

manifests/kustomize/third-party/argo/installs/namespace/kustomization.yaml
```diff
@@ -16,3 +16,9 @@ patches:
     kind: Deployment
     name: workflow-controller
     version: v1
+- path: workflow-controller-argo-role-patch.json
```
droctothorpe: Can you add a comment explaining why this patch is necessary? Thanks!

alyssacgoins:
> Hi, @droctothorpe ! Pulled the updates. There’s a [MR](https://github.com/kubeflow/pipelines/pull/12572) that hits the plugin limitations.
> 
> In this PR, the volume with certificates is [mounted](https://github.com/kubeflow/pipelines/pull/12572/files#diff-88badacfd78f66dfe792bec3763d8564798463dc822b92603a6ae50d990eb317L80) at pipeline compilation via driver.controller.volumes. The plugin is a global entity with separate settings that do not support managing volumes at the level of a specific compilation.
> 
> Possible options:
> 
> Dynamic volume management during compilation would require a separate design and discussion.
> 
> If volume settings are static, they can be moved to the global level, like the other plugin settings.
> 
> In the current POC, tests involving certificates are temporarily skipped.

Hi @ntny @droctothorpe, 
In October we merged #12082 which allowed users to enable pod-to-pod TLS (which requires certs mounted on the driver and launcher). #12572, referenced above, is a follow-up that allows a user to set env variable `CABUNDLE_SECRET_NAME` on the API server to pass custom certs into the driver/launcher even when pod-to-pod TLS is not enabled (this change handles the edge case in which the object store serves over TLS. The launcher would require custom certs to successfully upload and download artifacts.). To clarify - the volume containing the cert is mounted during workflow compilation [here](https://github.com/kubeflow/pipelines/pull/12572/files#diff-88badacfd78f66dfe792bec3763d8564798463dc822b92603a6ae50d990eb317R57), not during pipeline compilation. 

A solution to adapt your POC to this feature, would be to mount the custom cert(s) onto the central driver when it is initially created if `CABUNDLE_SECRET_NAME` is set.

When you say **hits plugin limitations** - what exactly does that mean here?

backend/test/compiler/argo_ginkgo_test.go
droctothorpe: I wish my quick POCs were this robust and impressive! 😂

ntny:
> ## Edge-case validation notes
> ### Setup
> ```shell
> git checkout master
> kubectl apply -k ./manifests/kustomize/env/platform-agnostic
> ```
> 
> ### The recurring run
> ```python
> from kfp import dsl
> from kfp import Client
> from kfp import compiler
> 
> 
> @dsl.component()
> def comp():
>     print("foo")
> 
> 
> @dsl.pipeline
> def pipeline():
>     comp().set_caching_options(enable_caching=False)
> 
> compiler.Compiler().compile(pipeline, "recurring-run.yaml")
> 
> client = Client()
> client.create_recurring_run(
>     experiment_id="ec535e9e-7d80-49da-a2b3-83e30a853aba",
>     job_name="validation",
>     pipeline_package_path="recurring-run.yaml",
>     cron_expression="* * * * *", # Every minute.
> )
> ```
> 
> Run completed successfuly:
> 
> ```
> NAME↑                           STATUS
> validation5tj4q-1-604852182     Succeeded
> ```
> 
> Driver is set to `2.15.0` in the API:
> 
> ```shell
> pipelines master* ⇣ kdd ml-pipeline | grep -i v2_driver
>       V2_DRIVER_IMAGE:                       ghcr.io/kubeflow/kfp-driver:2.15.0
> ```
> 
> The driver pods are referencing this image, e.g.:
> 
> ```
> kdp validation5tj4q-1-604852182-system-dag-driver-712789934 | grep -i image: | tail -1
>     Image:         ghcr.io/kubeflow/kfp-driver:2.15.0
> ```
> 
> ### Update backend
> Now let's update the `V2_DRIVER_IMAGE` env var in the API deployment then kick off another execution of the existing recurring run.
> 
> Updating to `2.14.4`.
> 
> Confirming that the resulting pod from the pre-existing recurring run references the updated driver image.
> 
> ```
> ❯ pipelines master* ⇣ kdp validation5tj4q-2-588074563-system-dag-driver-3124215983 | grep -i image: | tail -1
>     Image:         ghcr.io/kubeflow/kfp-driver:2.14.4
> ```
> 
> This is happening in master, not this PR branch, **which confirms that this edge-case is not introduced by this PR**.

Thanks! For me, it’s actually surprising that recurring runs are updated in single-user mode. I just noticed that in multi-user mode they are not updated, and I was assuming this was done intentionally by design :)
I opened a [MR](https://github.com/kubeflow/pipelines/pull/12648) for the bug, but we need to think twice about whether it’s worth fixing.

ntny:
> Hi @ntny @droctothorpe, In October we merged #12082 which allowed users to enable pod-to-pod TLS (which requires certs mounted on the driver and launcher). #12572, referenced above, is a follow-up that allows a user to set env variable `CABUNDLE_SECRET_NAME` on the API server to pass custom certs into the driver/launcher even when pod-to-pod TLS is not enabled (this change handles the edge case in which the object store serves over TLS. The launcher would require custom certs to successfully upload and download artifacts.). To clarify - the volume containing the cert is mounted during workflow compilation [here](https://github.com/kubeflow/pipelines/pull/12572/files#diff-88badacfd78f66dfe792bec3763d8564798463dc822b92603a6ae50d990eb317R57), not during pipeline compilation.
> 
> A solution to adapt your POC to this feature, would be to mount the custom cert(s) onto the central driver when it is initially created if `CABUNDLE_SECRET_NAME` is set.
> 
> When you say **hits plugin limitations** - what exactly does that mean here?

Sorry for the mix-up. I meant that we might potentially hit the limitations.

Thanks for clarifying! You’re right, since this is actually a global setting, we can simply move it to the driver level, and it isn’t actually a limitation.

droctothorpe:
Good call out from @mprahl. We want to make sure that the driver pod is not GCed immediately in the event of a failure so that users can access the logs for debugging purposes. Can you validate persistence of the agent pod when the driver is responsible for the failure? Thanks, @ntny!

droctothorpe:
Any updates on this, @ntny? Worried about potential merge conflicts if we drag it out too long.

ntny:
Hi, @droctothorpe!
Current status and blockers:
## Blockers
### Custom Driver cetificate implementation details:
The executor plugin injects user code into the agent pod via sidecar containers. Accordingly, the central driver also runs as a sidecar.
To mount the CA certificate into the agent pod, a secret with the name:
`argo-workflows-agent-ca-certificates`
[must be created](https://github.com/argoproj/argo-workflows/blob/main/workflow/common/common.go#L276). 
Creating this secret automatically creates a volume in the agent pod with the certificate. 
For local testing, I [copied](https://github.com/kubeflow/pipelines/pull/12269/changes#diff-5024268fbb0b80c2fca407bd63672f2e0e8c2b61e87f3bdd361e6a7a41a9dda3R98) the `kfp-api-tls-cert` secret into `argo-workflows-agent-ca-certificates`

We also need to manually define the corresponding volumeMount for the driver sidecar container.
I added the mount path [here](https://github.com/kubeflow/pipelines/pull/12269/changes#diff-c88281b3ec9e53ad44162811817728c3c93d2f1cd27b328bd4e60e0439c83948R24)
The local installation works well.

@alyssacgoins, do you mind elaborating on why mounting custom certs into the driver/launcher via a ConfigMap is needed? I’m not very familiar with this part, and it doesn’t seem like a typical approach.

In our case, the Argo agent pod runs the executor plugin, but we cannot define additional volumes (for example, from a ConfigMap) for that pod. The pod spec for the agent is fully controlled by the Argo Workflows controller. We can only control aspects of the container spec, such as volumeMounts for already existing volumes, but cannot introduce new volumes ourselves. Because of this limitation, we currently cannot mount a CA certificate from a ConfigMap into the driver via the agent/executor plugin path.

However, we can mount certificates from existing secrets using the standard mechanisms provided by the workflow controller.

### e2e tests
- Although the local installation works well and the secrets are created successfully, all TLS-enabled tests in CI e2e fail due to an error when creating the kfp-api-tls-cert secret.
`Warning  FailedMount  15m (x2 over 15m)  kubelet            MountVolume.SetUp failed for volume "tls-certs" : secret "kfp-api-tls-cert" not found`
- In addition, multi-user mode tests also failed with the following [error](https://github.com/kubeflow/pipelines/actions/runs/21593858209/job/62221301412?pr=12269#step:8:4225)
`[FAILED] Pipeline run did not complete, it stayed in RUNNING state`
I suspect this may be related to missing RBAC rules for WorkflowTaskSet, which is used for communication between the agent and the workflow-controller.
This will require some additional time to investigate.
I would greatly appreciate any pointers regarding the first issue (the secret creation problem).

## minor issues
- I opened an [MR](https://github.com/kubeflow/pipelines/pull/12648) with a fix could you please take a look? Without it, the driver and launcher might drift apart in the future. https://github.com/kubeflow/pipelines/pull/12269#issuecomment-3749383581
- Another important point: I noticed that almost all manifests use the stable driver version, which will be incompatible with these changes.
Could someone clarify how this process works?
Is the new driver image built and pushed to kfp-pipelines first, and only then are the manifests updated in the MR before merging?
- Regarding the agent pod removal: all information about its failure is available in the corresponding Argo Workflow.status, but I agree that it does make debugging more complicated. @mprahl

alyssacgoins:
> Hi, @droctothorpe! Current status and blockers:
> 
> ## Blockers
> ### Custom Driver cetificate implementation details:
> The executor plugin injects user code into the agent pod via sidecar containers. Accordingly, the central driver also runs as a sidecar. To mount the CA certificate into the agent pod, a secret with the name: `argo-workflows-agent-ca-certificates` [must be created](https://github.com/argoproj/argo-workflows/blob/main/workflow/common/common.go#L276). Creating this secret automatically creates a volume in the agent pod with the certificate. For local testing, I [copied](https://github.com/kubeflow/pipelines/pull/12269/changes#diff-5024268fbb0b80c2fca407bd63672f2e0e8c2b61e87f3bdd361e6a7a41a9dda3R98) the `kfp-api-tls-cert` secret into `argo-workflows-agent-ca-certificates`
> 
> We also need to manually define the corresponding volumeMount for the driver sidecar container. I added the mount path [here](https://github.com/kubeflow/pipelines/pull/12269/changes#diff-c88281b3ec9e53ad44162811817728c3c93d2f1cd27b328bd4e60e0439c83948R24) The local installation works well.
> 
> @alyssacgoins, do you mind elaborating on why mounting custom certs into the driver/launcher via a ConfigMap is needed? I’m not very familiar with this part, and it doesn’t seem like a typical approach.
> 
> In our case, the Argo agent pod runs the executor plugin, but we cannot define additional volumes (for example, from a ConfigMap) for that pod. The pod spec for the agent is fully controlled by the Argo Workflows controller. We can only control aspects of the container spec, such as volumeMounts for already existing volumes, but cannot introduce new volumes ourselves. Because of this limitation, we currently cannot mount a CA certificate from a ConfigMap into the driver via the agent/executor plugin path.
> 
> However, we can mount certificates from existing secrets using the standard mechanisms provided by the workflow controller.
> 
> ### e2e tests
> * Although the local installation works well and the secrets are created successfully, all TLS-enabled tests in CI e2e fail due to an error when creating the kfp-api-tls-cert secret.
>   `Warning  FailedMount  15m (x2 over 15m)  kubelet            MountVolume.SetUp failed for volume "tls-certs" : secret "kfp-api-tls-cert" not found`
> * In addition, multi-user mode tests also failed with the following [error](https://github.com/kubeflow/pipelines/actions/runs/21593858209/job/62221301412?pr=12269#step:8:4225)
>   `[FAILED] Pipeline run did not complete, it stayed in RUNNING state`
>   I suspect this may be related to missing RBAC rules for WorkflowTaskSet, which is used for communication between the agent and the workflow-controller.
>   This will require some additional time to investigate.
>   I would greatly appreciate any pointers regarding the first issue (the secret creation problem).
> 
> ## minor issues
> * I opened an [MR](https://github.com/kubeflow/pipelines/pull/12648) with a fix could you please take a look? Without it, the driver and launcher might drift apart in the future. [Central driver POC  #12269 (comment)](https://github.com/kubeflow/pipelines/pull/12269#issuecomment-3749383581)
> * Another important point: I noticed that almost all manifests use the stable driver version, which will be incompatible with these changes.
>   Could someone clarify how this process works?
>   Is the new driver image built and pushed to kfp-pipelines first, and only then are the manifests updated in the MR before merging?
> * Regarding the agent pod removal: all information about its failure is available in the corresponding Argo Workflow.status, but I agree that it does make debugging more complicated. @mprahl

@ntny Sure - to clarify first, users can also mount custom certs as a secret. In fact, a Secret is the preferred approach and the `CABUNDLE_SECRET_NAME` env var is prioritized above the `CABUNDLE_CONFIGMAP_NAME` env var during CA bundle configuration [here](https://github.com/kubeflow/pipelines/blob/master/backend/src/v2/compiler/argocompiler/common.go#L67).  The choice to provide a `ConfigMap` option for mounting custom certs on the driver/launcher addresses the cases in which we are mounting a custom CA to be treated as public within the cluster.

droctothorpe:
> I opened an https://github.com/kubeflow/pipelines/pull/12648 with a fix could you please take a look? Without it, the driver and launcher might drift apart in the future. Central driver POC  #12269 (comment)

Left a comment on this. I don't think we have to hold up this PR behind it.

> Another important point: I noticed that almost all manifests use the stable driver version, which will be incompatible with these changes.
> Could someone clarify how this process works?
> Is the new driver image built and pushed to kfp-pipelines first, and only then are the manifests updated in the MR before merging?

Looks like the bump is part of the official release. For example: [this](https://github.com/kubeflow/pipelines/commit/34f401666ab82cdeaf124016de2c9e62dfc8677a#diff-b8ec42e65046bed5636c791664edec2351cef100ac7d2dac2ae1d20c8acf1852R126). You don't have to worry about it. We'll tackle it.

> Regarding the agent pod removal: all information about its failure is available in the corresponding Argo Workflow.status, but I agree that it does make debugging more complicated. @mprahl

The workflow status stores events, not logs though, right? Correct me if I'm wrong. I'm worried about end users having access to the driver logs specifically in the event of a driver failure.

ntny:
Regarding the failing E2E tests,
~~for instance, [KFP API Integration V2 (Legacy) / API integration tests v2 - K8s with database v1.34.0 pod_to_pod](https://github.com/kubeflow/pipelines/actions/runs/21687329737/job/62538230481?pr=12269),
I [added](https://github.com/kubeflow/pipelines/pull/12269/changes/f7d91e6ee2c3d425be9809f9ef5666ef8769675e) debug code to check for the missing secret before and after the tests~~

~~From the output, I see that the secret is available~~
Снимок экрана 2026-02-05 в 19 33 18

~~though in the tests section I see:
MountVolume.SetUp failed for volume "tls-certs": secret "kfp-api-tls-cert" not found
in each pod (kubeflow-api-server, cache-server, etc.).~~
Снимок экрана 2026-02-05 в 19 32 38

Case with a lost certificate is no longer relevant

droctothorpe:
FYI we're aiming for a release on the week of the 23rd. No pressure but would be awesome if we could include this.

ntny:
### Concerns
Despite the PR being tested and working, there are a few concerns:

- We can only mount certificates into the driver via existing Secrets, [not ConfigMaps](https://github.com/kubeflow/pipelines/pull/12269#issuecomment-3820018488).
The Argo agent pod running the executor plugin has its Pod spec fully controlled by the Argo Workflows controller. We cannot define additional volumes (e.g., from a ConfigMap) — only configure aspects of the container spec like volumeMounts for existing volumes. Because of this, mounting custom CA certificates from a ConfigMap into the driver via the executor plugin is not possible, and it’s unlikely that Argo Workflows would accept a PR enabling this.
- We [cannot control](https://github.com/argoproj/argo-workflows/issues/12692) the GС executor agent pod lifecycle. It’s deleted after the workflow finishes regardless of GC settings, so driver logs for troubleshooting are lost. Each driver/launcher task corresponds to a status.node in the workflow; errors appear there. Additional workarounds: persist logs to S3 or the final workflow component.
@droctothorpe

droctothorpe:
image

Mike and I both manually validated and everything looks great! Amazing work, Anton. A few very small tweaks that should def happen in discrete PRs so that we don't hold this one up any longer:

- Enable wrapping of logs in the UI.
- Enabling searching / filtering of logs in the UI.
- Add some kind of info overlay or clickable i icon for system logs since users will have no idea what it is.
- Scale test the driver with extremely parallelized workloads.

Tag us in slack as soon as you rebase and we will merge. Happy to help if there are issues with the rebase.

ntny:
I created a [backup branch](https://github.com/kubeflow/pipelines/pull/13171) backup-central-driver to preserve all changes from this branch before merging with master.
The merge with master caused too many conflicts for a clean rebase, so to preserve the correct history, I extracted the diffs into a separate branch.

I am closing this PR to avoid confusion with the merge. All changes are saved in the backup branch, and it is this branch that should be merged.

PR #12279, opened 2025-09-23
fix(docs/CI): Small documentation updates and CI fixes after test refactoring

mprahl:

See each commit for details.

**Checklist:**

### DIFF
--- a/.github/workflows/api-server-tests.yml
+++ b/.github/workflows/api-server-tests.yml
@@ -106,16 +106,10 @@ jobs:
           fi
 
           PROXY=${{ matrix.proxy }}
-          if [ -n "$PROXY" ]; then
+          if [ -z "${PROXY:-}" ]; then
             PROXY=false
           fi
-          go run github.com/onsi/ginkgo/v2/ginkgo -r -v --cover -p --keep-going --github-output=true --nodes=$NUMBER_OF_NODES -v --label-filter=$TEST_LABEL -- -namespace=$NAMESPACE -useProxy=$PROXY -cacheEnabled=${{ matrix.cache_enabled }}
-        continue-on-error: true
 
-      - name: Collect Pod logs in case of Test Failures
-        id: collect-logs
-        if: ${{ steps.api-tests.outcome != 'success' || steps.create-kfp-cluster.outcome != 'success' }}
-        run: |
           echo "NUMBER_OF_NODES=$NUMBER_OF_NODES" >> $GITHUB_OUTPUT
           echo "TEST_LABEL=$TEST_LABEL" >> $GITHUB_OUTPUT
           echo "NAMESPACE=$NAMESPACE" >> $GITHUB_OUTPUT


AGENTS.md
```diff
@@ -473,7 +520,10 @@ docformatter --check --recursive sdk/python/ --exclude "compiler_test.py"
 - Deploy local cluster (standalone): `make -C backend kind-cluster-agnostic`
 - Deploy local cluster (development) and run the API server in the IDE: `make -C backend dev-kind-cluster`
 - Run SDK tests: `pytest -v sdk/python/kfp`
-- Run backend tests: `go test -v ./backend/...`
+- Run backend unit tests: `go test -v $(go list ./backend/... | grep -v backend/test/)`
+- Run compiler tests: `go test -v ./backend/test/compiler`
```
nsingla: These are also ginkgo tests, although these can run as plain go test but I think we should still run these with ginkgo command

AGENTS.md
```diff
@@ -473,7 +520,10 @@ docformatter --check --recursive sdk/python/ --exclude "compiler_test.py"
...
+- Run API tests: `ginkgo -v --label-filter="Smoke" ./backend/test/v2/api`
+- Run E2E tests: `go test -v ./backend/test/end2end -namespace=kubeflow`
```
nsingla: Same here

AGENTS.md
```diff
@@ -371,13 +414,13 @@ The frontend includes several generated code components:
...
 - Workflows: `.github/workflows/` (build, test, lint, release)
-- Composite actions: `.github/actions/` (e.g., `kfp-k8s`)
+- Composite actions: `.github/actions/` (e.g., `kfp-k8s`, `create-cluster`, `deploy`, `test-and-report`)
 - Typical checks: Go unit tests (backend), Python SDK tests, frontend tests/lint, image builds.
 
 ### Test matrices and variants (Kubernetes, stores, proxy, cache)
 
 - Kubernetes versions: CI runs a matrix across a low and high supported version, commonly `v1.29.2` and `v1.31.0`.
-  - Examples: `e2e-test.yml`, `sdk-execution.yml`, `upgrade-test.yml`, `periodic.yml`, `kfp-kubernetes-execution-tests.yml`, `kfp-webhooks.yml`, `kfp-samples.yml`, `api-server-tests.yml`, and frontend integration in `e2e-test.yml`.
+  - Examples: `e2e-test.yml`, `sdk-execution.yml`, `upgrade-test.yml`, `kfp-kubernetes-execution-tests.yml`, `kfp-webhooks.yml`, `api-server-tests.yml`, `compiler-tests.yml`, `api-integration-tests-v2-with-k8s-native-api.yml`, `integration-tests-v1.yml`, and frontend integration in `e2e-test-frontend.yml`.
```
nsingla: needs to be renamed here as well

nsingla (changes requested):

nsingla (approved):
lgtm

.github/resources/manifests/standalone/minio/kustomization.yaml
```diff
@@ -17,3 +17,4 @@ images:
 
 patches:
 - path: apiserver-env.yaml
+- path: minio-pvc.yaml
```
nsingla: do we need this patch to be applied to `standlaone/proxy-minio`, `standalone/cache-disabled-minio`, `multiuser/cache-disabled-minio` as well

.github/workflows/api-server-tests.yml
```diff
@@ -106,16 +106,9 @@ jobs:
           fi
 
           PROXY=${{ matrix.proxy }}
-          if [ -n "$PROXY" ]; then
+          if [ -z "$PROXY" ]; then
```
nsingla: I believe proxy can also be empty in case matrix doesn't set it (mostly with include), which is why I did a -n

.github/workflows/api-server-tests.yml
```diff
@@ -106,16 +106,9 @@ jobs:
...
             PROXY=false
           fi
-          go run github.com/onsi/ginkgo/v2/ginkgo -r -v --cover -p --keep-going --github-output=true --nodes=$NUMBER_OF_NODES -v --label-filter=$TEST_LABEL -- -namespace=$NAMESPACE -useProxy=$PROXY -cacheEnabled=${{ matrix.cache_enabled }}
```
nsingla, line 112: Good catch

.github/workflows/api-server-tests.yml
mprahl: @nsingla the `-n` returns true if non-zero so I think PROXY was always off prior to this change. I just pushed a change to account for unset variables but I don't think it was needed.

nsingla (approved):

PR #12281, opened 2025-09-24
feat(backend): adding a retry logic to object store s3 configuration

nsingla:
Adding a retry logic to S3 Uploads. This is benefecial especially in case when your upload fails due to some issues like network, or some other type of glitch. And its not really a good UX when you are at 90% done but your upload fails and due to that your whole pipeline fails

**Checklist:**


backend/src/v2/objectstore/object_store.go
```diff
@@ -255,11 +257,14 @@ func createS3BucketSession(ctx context.Context, namespace string, sessionInfo *S
 	if err != nil {
 		return nil, err
 	}
-	config.Credentials = creds
-	config.Region = aws.String(params.Region)
-	config.DisableSSL = aws.Bool(params.DisableSSL)
-	config.S3ForcePathStyle = aws.Bool(params.ForcePathStyle)
-
+	s3Config, err := config.LoadDefaultConfig(context.TODO(),
+		config.WithRetryer(func() aws.Retryer { return retry.AddWithMaxAttempts(retry.NewStandard(), 5) }),
```
ntny: I believe it’s better to move MaxAttempts to config, but it’s up to you.

Also, do you think it makes sense to document in comments which errors are retried and which are not, and to specify the backoff policy in the comments as well?

nsingla: Yup, its definitely better to move it to a config
Also, I plan to add explicit HTTP Error codes for retry with exponential backoff, its just a first pass at retry as I wanted to see if the code changes actually makes sense and/or work :)

nsingla: After reading more about Standard Retry, I feel like this is enough for now, if we run into a situation where we need to extend it, we can revisit it. So for now, I;ve left a code comment explaining the retry and moved the max reties to config

backend/src/v2/objectstore/object_store.go
```diff
@@ -255,11 +257,19 @@ func createS3BucketSession(ctx context.Context, namespace string, sessionInfo *S
 	if err != nil {
 		return nil, err
 	}
-	config.Credentials = creds
-	config.Region = aws.String(params.Region)
-	config.DisableSSL = aws.Bool(params.DisableSSL)
-	config.S3ForcePathStyle = aws.Bool(params.ForcePathStyle)
-
+	s3Config, err := config.LoadDefaultConfig(context.TODO(),
```
mprahl:
```suggestion
	s3Config, err := config.LoadDefaultConfig(ctx,
```

mprahl (approved):
/approve
/lgtm

PR #12296, opened 2025-09-26
chore: fix import order in 15 files (baseline formatting)

kaikaila:
Context:
- Local pre-commit runs `golangci-lint fmt` and reorders imports in 15 legacy files on every commit.
- CI only runs `golangci-lint run --new-from-rev HEAD` and reports 0 issues, so these legacy files stayed as-is in master.

What this PR does:
- Reorders imports in the 6 files to align with the formatter expectations.
- No functional changes; formatting-only baseline to avoid unrelated diffs in future feature PRs.

Notes:
- After this baseline formatting, local pre-commit will no longer pull these files into unrelated commits.

### DIFF
--- a/backend/src/common/client/api_server/v2/experiment_client.go
+++ b/backend/src/common/client/api_server/v2/experiment_client.go
@@ -16,6 +16,7 @@ package api_server_v2
 
 import (
 	"fmt"
+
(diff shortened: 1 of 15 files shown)


mprahl (approved):
/approve
/lgtm

PR #12302, opened 2025-09-29
ci: refactor pre-commit workflow to correctly detect new lint issues

kaikaila:
PR Description

This PR refactors the pre-commit GitHub Action workflow to ensure golangci-lint correctly detects new lint issues introduced in pull requests.

## Changes
	- Replace `--new-from-rev` HEAD with `--new --new-from-merge-base=main` to ensure lint runs against diffs from the branch merge-base.
	- Prevents false green CI runs where lint issues could slip into master.
	- Keeps configuration aligned with local pre-commit behavior.
	-  Fixes import order in 4 files as an example: these violations slipped into master previously because CI was not actually running golangci-lint.

## Context

Previously, the workflow used --new-from-rev HEAD, which effectively compared HEAD against itself, resulting in no lint issues being reported. This configuration caused format and lint errors to be missed in CI.

## Impact
	- CI will now block PRs that introduce new lint violations.
	- No user-facing changes, only CI configuration improvements.

## Reference
[ Golangci lint Documentation](https://golangci-lint.run/docs/configuration/file/)


.github/workflows/pre-commit.yml
```diff
@@ -12,18 +12,22 @@ jobs:
...
+        run: git fetch origin master:refs/remotes/origin/master
+      - uses: actions/setup-python@v3
+      - uses: actions/setup-go@v5
+        with:
+          go-version-file: go.mod
+      - name: golangci-lint
+        uses: golangci/golangci-lint-action@v8
+        with:
+          version: v2.3
+          args: --new --new-from-merge-base=origin/master
```
HumairAK: Can we update this to be more portable and not hardcode master? something like: 

```
args: --new --new-from-merge-base=origin/${{ github.base_ref }}
```

HumairAK: let's also add `--whole-files` so we don't miss side effect issues [[1]](https://golangci-lint.run/docs/welcome/faq/?utm_source=chatgpt.com#why---new-from-xxx-dont-seem-to-be-working-in-some-cases)

kaikaila: > let's also add `--whole-files` so we don't miss side effect issues [[1]](https://golangci-lint.run/docs/welcome/faq/?utm_source=chatgpt.com#why---new-from-xxx-dont-seem-to-be-working-in-some-cases)

The reason I originally set whole-files: false was that I wanted to minimize noise in the PR — I was trying to avoid pulling in a lot of unrelated lint issues so that you wouldn’t have extra review overhead beyond the main purpose of the PR.

Since you don’t mind the additional reports, I’m happy to switch it to true and follow your recommendation. 🙌

kaikaila: After switching to `--whole-files`, golangci-lint reported [19 additional issues ](https://github.com/kubeflow/pipelines/actions/runs/18176608687/job/51743875309?pr=12302) in the 4 files where I only meant to fix import order. I don’t mind fixing these 19 myself, but for larger PRs this could easily blow up the scope — just want to confirm if this is the behavior we want.

HumairAK: yes I think it's fine to resolve these

HumairAK:
/lgtm
/approve

fyi @nsingla

PR #12313, opened 2025-10-02
feat(backend): Adding Multi-CPU architecture support in Dockerfile.* files for supporting Arm64 platform

jtu-ampere:
**Feature:**

By adding TARGETARCH argument to get docker's "--platform" parameters, it supports the 3 Dockerfile files to download the CPU architecture matched packages.

It will build Multi-CPU Architecture Docker Images for ARM/arm64 and x86-64/amd64 platforms.

Here is the example to build a manifest with arm64 and amd64 container images.

`docker buildx build --platform linux/arm64,linux/amd64 -t api-server-temp -f backend/Dockerfile .`

REF: https://itnext.io/building-multi-cpu-architecture-docker-images-for-arm-and-x86-1-the-basics-2fa97869a99b
 
The corresponding change may be added to the GitHub workflow later.
REF: https://docs.docker.com/build/ci/github-actions/multi-platform/

**Checklist:**


zazulam:
@jtu-ampere can you update the associated readme and makefiles for building these images as well

backend/Dockerfile
```diff
@@ -34,12 +34,15 @@ RUN wget https://bootstrap.pypa.io/get-pip.py && python3 get-pip.py
 COPY backend/requirements.txt .
 RUN python3 -m pip install -r requirements.txt --no-cache-dir
 
+ARG TARGETARCH
```
mprahl, line 37: Could you default this to amd64 if the build arg isn't provided?

Also, I don't think we need both an ENV and build arg.

jtu-ampere: Yes, it's correct. 
I will keep `ARG TARGETARCH` and remove the `ENV TARGETARCH`.
For setting its default as `amd64`, based on what I tested on Ubuntu 22.04, the docker (CE version 28.4.0) will detect the local CPU architecture and pass to the TARGETARCH argument when building without buildx or targeting a single platform without specifying --platform.

jtu-ampere: Additional information. When set the default value as `ARG TARGETARCH amd64` in Dockerfile, it won't take the build argument from docker buildx with --platform.

Here are the logs when I run `docker buildx  build --progress=plain  --platform linux/arm64 -t api-server-temp:2.14.3 -f backend/Dockerfile . ` or `docker build --progress=plain -t api-server-temp:2.14.3 -f backend/Dockerfile . ` 
The `TARGETARCH` is fixed to `amd64`. 
Please note that.

#16 [compiler  6/12] RUN echo "Building for architecture: amd64"
#16 0.144 Building for architecture: amd64
#16 DONE 0.2s

#17 [compiler  7/12] RUN if [ "amd64" = "amd64" ]; then      echo "Building for architecture for amd64" ;     fi
#17 0.191 Building for architecture for amd64
#17 DONE 0.2s

juliusvonkohout:
Part of https://github.com/kubeflow/manifests/issues/2745

backend/Dockerfile
```diff
@@ -34,8 +34,9 @@ RUN wget https://bootstrap.pypa.io/get-pip.py && python3 get-pip.py
 COPY backend/requirements.txt .
 RUN python3 -m pip install -r requirements.txt --no-cache-dir
 
+###
```
mprahl:
```suggestion
```

backend/Dockerfile
```diff
@@ -34,12 +34,16 @@ RUN wget https://bootstrap.pypa.io/get-pip.py && python3 get-pip.py
...
+ARG TARGETARCH
+RUN echo "Building for architecture: ${TARGETARCH}"
+
 # Downloading Argo CLI so that the samples are validated
 ENV ARGO_VERSION=v3.6.7
-RUN curl -sLO https://github.com/argoproj/argo-workflows/releases/download/${ARGO_VERSION}/argo-linux-amd64.gz && \
-    gunzip argo-linux-amd64.gz && \
-    chmod +x argo-linux-amd64 && \
-    mv ./argo-linux-amd64 /usr/local/bin/argo
+RUN curl -sLO https://github.com/argoproj/argo-workflows/releases/download/${ARGO_VERSION}/argo-linux-$TARGETARCH.gz && \
```
mprahl, line 42: I think we can make this automatic by parsing the output of the `arch` command. You can just ask AI how to do this.

backend/Dockerfile.driver
```diff
@@ -26,8 +26,9 @@ COPY ./api/go.mod ./api/go.mod
 RUN GO111MODULE=on go mod download
 
 COPY . .
-
-RUN GO111MODULE=on CGO_ENABLED=0 GOOS=linux GOARCH=amd64 go build -tags netgo -gcflags="${GCFLAGS}" -ldflags '-extldflags "-static"' -o /bin/driver ./backend/src/v2/cmd/driver/*.go
+ARG TARGETARCH
+RUN echo "Building for architecture: ${TARGETARCH}"
+RUN GO111MODULE=on CGO_ENABLED=0 GOOS=linux GOARCH=$TARGETARCH go build -tags netgo -gcflags="${GCFLAGS}" -ldflags '-extldflags "-static"' -o /bin/driver ./backend/src/v2/cmd/driver/*.go
```
mprahl: Do we need to set the GOARCH at all? I think it will autodetect when building it with `docker buildx build --platform <platform>`.

backend/Dockerfile.launcher
```diff
@@ -24,8 +24,9 @@ COPY ./api/go.mod ./api/go.mod
 RUN GO111MODULE=on go mod download
 
 COPY . .
-
-RUN GO111MODULE=on CGO_ENABLED=0 GOOS=linux GOARCH=amd64 go build -tags netgo -ldflags '-extldflags "-static"' -o /bin/launcher-v2 ./backend/src/v2/cmd/launcher-v2/*.go
+ARG TARGETARCH
+RUN echo "Building for architecture: $TARGETARCH"
+RUN GO111MODULE=on CGO_ENABLED=0 GOOS=linux GOARCH=$TARGETARCH go build -tags netgo -ldflags '-extldflags "-static"' -o /bin/launcher-v2 ./backend/src/v2/cmd/launcher-v2/*.go
```
mprahl: Same comment as above.

backend/Dockerfile.driver
jtu-ampere: The parameter, GOARCH, was there before I modified the Dockerfile.driver.
Is it needed for the go?

mprahl: Perhaps we can try just removing it and see if it works for you. I think it shouldn't be needed.

backend/Dockerfile
jtu-ampere: **Additional information for backend/Dockerfile:**
Those lines under `#Downloading Argo CLI so that the samples are validated` are in the section of `FROM golang:1.24-bookworm` AS builder which is a temporary image as a builder for providing the built binary to the final container image. It won't change the final container image size.
I have run a quick test as Humair suggested and the image size is same as the one built with `ARG TARGETARCH`.
Hence, it seems my original approach is easy and simple. I would like to keep this way.

mprahl: @jtu-ampere why was this comment marked as resolved?

mprahl: It seems TARGETARCH is correct in most cases.

PR #12347, opened 2025-10-11
feat(sdk): add kubeflow extras and auto-detection for component impor…

### LINKED ISSUE #12027: [feature] Include Kubeflow SDK in KFP SDK
### Feature Area
/area sdk

As discussed in the Kubeflow SDK, we want to include the Kubeflow SDK in the python requirements of the KFP SDK.

This has several benefits:

1. Force a coherent product story for our end users.
2. Force the SDK to resolve package dependency issues earlier 
3. Seed the Kubeflow SDK UI with the existing KFP UI
    - See [this PR](https://github.com/kubeflow/pipelines/pull/11981) as a point of reference and [this github issue](https://github.com/kubeflow/trainer/issues/2648)
4. Force the SDK to prioritize migrating KFP 

`pip install kfp` to include `kubeflow`

From work that was done by Red Hat's @instructlab project (and likely other closed-source projects), the experience of using the Kubeflow Trainer with Kubeflow Pipelines is bad. The projects feel entirely separate even though they are meant to be used together. 

Concretely, the code putting them together can be seen here: https://github.com/opendatahub-io/ilab-on-ocp/blob/main/training/components.py to get clarity on how challenging it is.

Yes, but it's an extraordinary amount of code that I do not expect an end user to actually write. 

Love this idea? Give it a 👍.

Prateekbala:
## Description

This PR implements automatic Kubeflow SDK integration for KFP components as requested in #12027. It adds the ability to automatically detect and install the Kubeflow SDK when components use kubeflow imports.

### 1. Added Kubeflow Extras to setup.py
- Added `kubeflow = ['kubeflow']` extras option
- Included kubeflow in the 'all' extras bundle
- Allows users to install with `pip install kfp[kubeflow]`

### 2. Implemented AST-based Auto-detection
- Added `_detect_kubeflow_imports_in_function()` in `component_factory.py`
- Uses Abstract Syntax Tree parsing to detect kubeflow imports in component functions
- Supports multiple import patterns:
  - `import kubeflow`
  - `import kubeflow.training` (or any submodule)
  - `from kubeflow import ...`

### 3. Automatic Package Installation
- Modified `_get_packages_to_install_command()` to auto-detect kubeflow usage
- Automatically adds 'kubeflow' to packages_to_install when detected
- Only installs when kubeflow imports are actually used in component code
- Respects explicit user-provided packages (no duplication)

### 4. Comprehensive Test Coverage
- Added test cases for all import detection scenarios
- Tests for components with and without kubeflow imports
- Verification that kubeflow is only installed when needed
- Tests for explicit package specification (no auto-detection override)

## Related Issues

Fixes #12027 - Include Kubeflow SDK in KFP SDK


mynewestgitaccount:
There should be a way to explicitly opt out of this auto-detection without adding kubeflow to the `packages_to_install` list, similar to the `install_kfp_package` argument. Autodetecting the kubeflow package based on the AST doesn't account for scenarios where kubeflow is purposely installed in the underlying base image of the component.

It's currently a valid/working use-case to run components without having pip installed in the image at all (using `packages_to_install=[], install_kfp_package=False`), and automatically trying to install kubeflow based on the AST with no way to opt out would break that.

Prateekbala:
Thank you for the thorough review and the important edge case catch. You're absolutely right—my implementation needs adjustments to handle these scenarios properly.

Issue #12352, opened 2025-10-13
[feature] Add download_to_workspace option to dsl.importer


VaniHaripriya:
### Feature Area

/area backend 
/area sdk

### What feature would you like to see?
As a KFP user, I want dsl.importer to support a download_to_workspace=True option, so that artifacts can be downloaded directly into the pipeline workspace. This requires adding a download_to_workspace field to the ImporterSpec protobuf message and updating the corresponding Python class and function. When the input artifact has the metadata field `_kfp_workspace=True` set, then `artifact.path` should return the correct path within the workspace (e.g., prefixed with /kfp-workspace/.artifacts/)

The KFP Importer needs to differentiate artifacts downloaded to the workspace. It should use a new execution type in MLMD (e.g., system.ImporterWorkspaceExecution). Before registering artifacts in MLMD, they should be downloaded to the workspace at a path like /kfp-workspace/.artifacts/. The KFP Importer should fail if the artifact is from OCI (starts with oci://).

The KFP Launcher should skip downloading artifacts to local emptyDir volumes if the artifact has the _kfp_workspace metadata field set to True (indicating it's already in the workspace).

The Container Driver needs to:
1. Set a _kfp_workspace=True metadata field (not persisted to MLMD) on artifacts originating from a system.ImporterWorkspaceExecution task.
2. Mount the KFP workspace volume in the Pod spec patch if an input artifact is from the workspace. 
3.  Disallow user-mounted volumes in or under /kfp-workspace.

---

Love this idea? Give it a 👍.

PR #12353, opened 2025-10-13
feat(backend/sdk):Add download_to_workspace option to dsl.importer

VaniHaripriya:

Resolves [#12352 ](https://github.com/kubeflow/pipelines/issues/12352).

This PR adds download_to_workspace option to dsl.importer to download artifacts into the pipeline workspace and consume them downstream without re-downloading.

**Checklist:**

### DIFF
--- a/api/v2alpha1/pipeline_spec.proto
+++ b/api/v2alpha1/pipeline_spec.proto
@@ -868,11 +868,14 @@ message PipelineDeploymentConfig {
     map<string, ValueOrRuntimeParameter> custom_properties = 4
         [deprecated = true];
 
+    // Whether or not import an artifact regardless it has been imported before.
+    bool reimport = 5;
+
     // Properties of the Artifact.
     google.protobuf.Struct metadata = 6;
 
-    // Whether or not import an artifact regardless it has been imported before.
-    bool reimport = 5;
+    // If true, download artifact into the pipeline workspace.
+    bool download_to_workspace = 7;
   }
 
   // ResolverSpec resolves artifacts from historical metadata and returns them
--- a/backend/src/v2/compiler/argocompiler/importer.go
+++ b/backend/src/v2/compiler/argocompiler/importer.go
@@ -34,7 +34,7 @@ func (c *workflowCompiler) Importer(name string, componentSpec *pipelinespec.Com
 	return c.saveComponentImpl(name, importer)
 }
 
-func (c *workflowCompiler) importerTask(name string, task *pipelinespec.PipelineTaskSpec, taskJSON string, parentDagID string) (*wfapi.DAGTask, error) {
+func (c *workflowCompiler) importerTask(name string, task *pipelinespec.PipelineTaskSpec, taskJSON string, parentDagID string, downloadToWorkspace bool) (*wfapi.DAGTask, error) {
 	componentPlaceholder, err := c.useComponentSpec(task.GetComponentRef().GetName())
 	if err != nil {
 		return nil, err
@@ -45,7 +45,7 @@ func (c *workflowCompiler) importerTask(name string, task *pipelinespec.Pipeline
 	}
 	return &wfapi.DAGTask{
 		Name:     name,
-		Template: c.addImporterTemplate(),
+		Template: c.addImporterTemplate(downloadToWorkspace),
 		Arguments: wfapi.Arguments{Parameters: []wfapi.Parameter{{
 			Name:  paramTask,
 			Value: wfapi.AnyStringPtr(taskJSON),
@@ -62,8 +62,11 @@ func (c *workflowCompiler) importerTask(name string, task *pipelinespec.Pipeline
 	}, nil
 }
 
-func (c *workflowCompiler) addImporterTemplate() string {
+func (c *workflowCompiler) addImporterTemplate(downloadToWorkspace bool) string {
 	name := "system-importer"
+	if downloadToWorkspace {
+		name += "-workspace"
+	}
 	if _, alreadyExists := c.templates[name]; alreadyExists {
 		return name
 	}
@@ -104,6 +107,24 @@ func (c *workflowCompiler) addImporterTemplate() string {
 	if value, ok := os.LookupEnv(PublishLogsEnvVar); ok {
 		args = append(args, "--publish_logs", value)
 	}
+
+	var volumeMounts []k8score.VolumeMount
+	var volumes []k8score.Volume
+	if downloadToWorkspace {
+		volumeMounts = append(volumeMounts, k8score.VolumeMount{
+			Name:      workspaceVolumeName,
+			MountPath: component.WorkspaceMountPath,
+		})
+		volumes = append(volumes, k8score.Volume{
+			Name: workspaceVolumeName,
+			VolumeSource: k8score.VolumeSource{
+				PersistentVolumeClaim: &k8score.PersistentVolumeClaimVolumeSource{
+					ClaimName: fmt.Sprintf("{{workflow.name}}-%s", workspaceVolumeName),
+				},
+			},
+		})
+	}
+
 	importerTemplate := &wfapi.Template{
 		Name: name,
 		Inputs: wfapi.Inputs{
@@ -115,13 +136,15 @@ func (c *workflowCompiler) addImporterTemplate() string {
 			},
 		},
 		Container: &k8score.Container{
-			Image:     c.launcherImage,
-			Command:   c.launcherCommand,
-			Args:      args,
-			EnvFrom:   []k8score.EnvFromSource{metadataEnvFrom},
-			Env:       commonEnvs,
-			Resources: driverResources,
+			Image:        c.launcherImage,
+			Command:      c.launcherCommand,
+			Args:         args,
+			EnvFrom:      []k8score.EnvFromSource{metadataEnvFrom},
+			Env:          commonEnvs,
+			Resources:    driverResources,
+			VolumeMounts: volumeMounts,
 		},
+		Volumes: volumes,
 	}
 
 	// If TLS is enabled (apiserver or metadata), add the custom CA bundle to the importer template.
--- a/backend/src/v2/component/importer_launcher.go
+++ b/backend/src/v2/component/importer_launcher.go
@@ -8,6 +8,7 @@ import (
 
 	"github.com/kubeflow/pipelines/backend/src/common/util"
 
+	"github.com/kubeflow/pipelines/backend/src/v2/config"
 	"github.com/kubeflow/pipelines/backend/src/v2/objectstore"
 
 	pb "github.com/kubeflow/pipelines/third_party/ml-metadata/go/ml_metadata"
@@ -125,12 +126,18 @@ func (l *ImportLauncher) Execute(ctx context.Context) (err error) {
 	if err != nil {
 		return err
 	}
+	// Determine execution type
+	executionType := metadata.ExecutionType(metadata.ImporterExecutionTypeName)
+	if l.importer.GetDownloadToWorkspace() {
+		executionType = metadata.ExecutionType(metadata.ImporterWorkspaceExecutionTypeName)
+	}
+
 	ecfg := &metadata.ExecutionConfig{
 		TaskName:      l.task.GetTaskInfo().GetName(),
 		PodName:       l.launcherV2Options.PodName,
 		PodUID:        l.launcherV2Options.PodUID,
 		Namespace:     l.launcherV2Options.Namespace,
-		ExecutionType: metadata.ImporterExecutionTypeName,
+		ExecutionType: executionType,
 		ParentDagID:   l.importerLauncherOptions.ParentDagID,
 	}
 	createdExecution, err := l.metadataClient.CreateExecution(ctx, pipeline, ecfg)
@@ -253,15 +260,17 @@ func (l *ImportLauncher) ImportSpecToMLMDArtifact(ctx context.Context) (artifact
 	}
 
 	if strings.HasPrefix(artifactUri, "oci://") {
+		// OCI artifacts are not supported when workspace is used
+		if l.importer.GetDownloadToWorkspace() {
+			return nil, fmt.Errorf("importer workspace download does not support OCI registries")
+		}
 		artifactType, err := metadata.SchemaToArtifactType(schema)
 		if err != nil {
 			return nil, fmt.Errorf("converting schema to artifact type failed: %w", err)
 		}
-
 		if *artifactType.Name != "system.Model" {
 			return nil, fmt.Errorf("the %s artifact type does not support OCI registries", *artifactType.Name)
 		}
-
 		return artifact, nil
 	}
 
@@ -283,7 +292,52 @@ func (l *ImportLauncher) ImportSpecToMLMDArtifact(ctx context.Context) (artifact
 	}
 	storeSessionInfoStr := string(storeSessionInfoJSON)
 	artifact.CustomProperties["store_session_info"] = metadata.StringValue(storeSessionInfoStr)
+
+	// Download the artifact into the workspace
+	if l.importer.GetDownloadToWorkspace() {
+		bucketConfig, err := l.resolveBucketConfigForURI(ctx, artifactUri)
+		if err != nil {
+			return nil, err
+		}
+		localPath, err := LocalWorkspacePathForURI(artifactUri)
+		if err != nil {
+			return nil, fmt.Errorf("failed to get local path for uri %q: %w", artifactUri, err)
+		}
+		blobKey, err := bucketConfig.KeyFromURI(artifactUri)
+		if err != nil {
+			return nil, fmt.Errorf("failed to derive blob key from uri %q while downloading artifact into workspace: %w", artifactUri, err)
+		}
+		bucket, err := objectstore.OpenBucket(ctx, l.k8sClient, l.launcherV2Options.Namespace, bucketConfig)
+		if err != nil {
+			return nil, fmt.Errorf("failed to open bucket for uri %q: %w", artifactUri, err)
+		}
+		defer bucket.Close()
+		glog.Infof("Downloading artifact %q (blob key %q) to workspace path %q", artifactUri, blobKey, localPath)
+		if err := objectstore.DownloadBlob(ctx, bucket, localPath, blobKey); err != nil {
+			return nil, fmt.Errorf("failed to download artifact to workspace: %w", err)
+		}
+	}
 	return artifact, nil
+}
+
+// resolveBucketConfigForURI parses bucket configuration for a given artifact URI and
+// attaches session information from the kfp-launcher ConfigMap when available.
+func (l *ImportLauncher) resolveBucketConfigForURI(ctx context.Context, uri string) (*objectstore.Config, error) {
+	bucketConfig, err := objectstore.ParseBucketConfigForArtifactURI(uri)
+	if err != nil {
+		return nil, fmt.Errorf("failed to parse bucket config while resolving uri %q: %w", uri, err)
+	}
+	// Resolve and attach session info from kfp-launcher config for the artifact provider
+	if cfg, err := config.FromConfigMap(ctx, l.k8sClient, l.launcherV2Options.Namespace); err != nil {
+		glog.Warningf("failed to load launcher config while resolving bucket config: %v", err)
+	} else if cfg != nil {
+		if sess, err := cfg.GetStoreSessionInfo(uri); err != nil {
+			glog.Warningf("failed to resolve store session info for %q: %v", uri, err)
+		} else {
+			bucketConfig.SessionInfo = &sess
+		}
+	}
+	return bucketConfig, nil
 }
 
 func (l *ImportLauncher) getOutPutArtifactName() (string, error) {
--- a/backend/src/v2/component/launcher_v2.go
+++ b/backend/src/v2/component/launcher_v2.go
@@ -715,6 +715,12 @@ func downloadArtifacts(ctx context.Context, executorInput *pipelinespec.Executor
 		for _, artifact := range artifactList.Artifacts {
 			// Iterating through the artifact list allows for collected artifacts to be properly consumed.
 			inputArtifact := artifact
+			// Skip downloading if the artifact is flagged as already present in the workspace
+			if inputArtifact.GetMetadata() != nil {
+				if v, ok := inputArtifact.GetMetadata().GetFields()["_kfp_workspace"]; ok && v.GetBoolValue() {
+					continue
+				}
+			}
 			localPath, err := LocalPathForURI(inputArtifact.Uri)
 			if err != nil {
 				glog.Warningf("Input Artifact %q does not have a recognized storage URI %q. Skipping downloading to local path.", name, inputArtifact.Uri)
@@ -857,6 +863,20 @@ func getPlaceholders(executorInput *pipelinespec.ExecutorInput) (placeholders ma
 		// Prepare input uri placeholder.
 		key := fmt.Sprintf(`{{$.inputs.artifacts['%s'].uri}}`, name)
 		placeholders[key] = inputArtifact.Uri
+
+		// If the artifact is marked as already in the workspace, map to the workspace path
+		// with the same shape as LocalPathForURI, but rebased under the workspace mount.
+		if inputArtifact.GetMetadata() != nil {
+			if v, ok := inputArtifact.GetMetadata().GetFields()["_kfp_workspace"]; ok && v.GetBoolValue() {
+				localPath, lerr := LocalWorkspacePathForURI(inputArtifact.Uri)
+				if lerr != nil {
+					return nil, fmt.Errorf("failed to get local workspace path for input artifact %q: %w", name, lerr)
+				}
+				key = fmt.Sprintf(`{{$.inputs.artifacts['%s'].path}}`, name)
+				placeholders[key] = localPath
+				continue
+			}
+		}
 
 		localPath, err := LocalPathForURI(inputArtifact.Uri)
 		if err != nil {
@@ -1008,6 +1028,21 @@ func retrieveArtifactPath(artifact *pipelinespec.RuntimeArtifact) (string, error
 	} else {
 		return LocalPathForURI(artifact.Uri)
 	}
+}
+
+// LocalWorkspacePathForURI returns the local workspace path for a given artifact URI.
+// It preserves the same path shape as LocalPathForURI, but rebases it under the
+// workspace artifacts directory: /kfp-workspace/.artifacts/...
+func LocalWorkspacePathForURI(uri string) (string, error) {
+	if strings.HasPrefix(uri, "oci://") {
+		return "", fmt.Errorf("failed to generate workspace path for URI %s: OCI not supported for workspace artifacts", uri)
+	}
+	localPath, err := LocalPathForURI(uri)
+	if err != nil {
+		return "", err
+	}
+	// Rebase under the workspace mount, stripping the leading '/'
+	return filepath.Join(WorkspaceMountPath, ".artifacts", strings.TrimPrefix(localPath, "/")), nil
 }
 
 func prepareOutputFolders(executorInput *pipelinespec.ExecutorInput) error {
--- a/backend/src/v2/component/launcher_v2_test.go
+++ b/backend/src/v2/component/launcher_v2_test.go
@@ -20,6 +20,7 @@ import (
 	"errors"
 	"io"
 	"os"
+	"path/filepath"
 	"testing"
 
 	"github.com/kubeflow/pipelines/backend/src/v2/cacheutils"
@@ -173,6 +174,29 @@ func Test_executeV2_publishLogs(t *testing.T) {
 
 			}
 		})
+	}
+}
+
+func Test_getPlaceholders_WorkspaceArtifactPath(t *testing.T) {
+	execIn := &pipelinespec.ExecutorInput{
+		Inputs: &pipelinespec.ExecutorInput_Inputs{
+			Artifacts: map[string]*pipelinespec.ArtifactList{
+				"data": {
+					Artifacts: []*pipelinespec.RuntimeArtifact{
+						{Uri: "minio://mlpipeline/sample/sample.txt", Metadata: &structpb.Struct{Fields: map[string]*structpb.Value{"_kfp_workspace": structpb.NewBoolValue(true)}}},
+					},
+				},
+			},
+		},
+	}
+	ph, err := getPlaceholders(execIn)
+	if err != nil {
+		t.Fatalf("getPlaceholders error: %v", err)
+	}
+	actual := ph["{{$.inputs.artifacts['data'].path}}"]
+	expected := filepath.Join(WorkspaceMountPath, ".artifacts", "minio", "mlpipeline", "sample", "sample.txt")
+	if actual != expected {
+		t.Fatalf("placeholder path mismatch: actual=%q expected=%q", actual, expected)
 	}
 }
 
--- a/backend/src/v2/driver/driver.go
+++ b/backend/src/v2/driver/driver.go
@@ -461,7 +461,7 @@ func needsWorkspaceMount(executorInput *pipelinespec.ExecutorInput) bool {
 				return true
 			}
 
-			if strings.HasPrefix(strVal.StringValue, component.WorkspaceMountPath) {
+			if strings.Contains(strVal.StringValue, component.WorkspaceMountPath) {
 				return true
 			}
 		}
--- a/backend/src/v2/driver/driver_test.go
+++ b/backend/src/v2/driver/driver_test.go
@@ -1207,6 +1207,70 @@ func TestWorkspaceMount_PassthroughVolumes_ApplyAndCapture(t *testing.T) {
 	}
 }
 
+func TestWorkspaceMount_TriggeredByArtifactMetadata(t *testing.T) {
+	proxy.InitializeConfigWithEmptyForTests()
+	containerSpec := &pipelinespec.PipelineDeploymentConfig_PipelineContainerSpec{Image: "python:3.9"}
+	componentSpec := &pipelinespec.ComponentSpec{
+		TaskConfigPassthroughs: []*pipelinespec.TaskConfigPassthrough{
+			{
+				Field:       pipelinespec.TaskConfigPassthroughType_KUBERNETES_VOLUMES,
+				ApplyToTask: true,
+			},
+		},
+	}
+
+	// Build an ExecutorInput that does NOT reference workspace path in params,
+	// but contains an input artifact marked as already in workspace.
+	execInput := &pipelinespec.ExecutorInput{
+		Inputs: &pipelinespec.ExecutorInput_Inputs{
+			Artifacts: map[string]*pipelinespec.ArtifactList{
+				"data": {
+					Artifacts: []*pipelinespec.RuntimeArtifact{
+						{
+							Uri: "minio://mlpipeline/sample/sample.txt",
+							Metadata: &structpb.Struct{Fields: map[string]*structpb.Value{
+								"_kfp_workspace": structpb.NewBoolValue(true),
+							}},
+						},
+					},
+				},
+			},
+		},
+	}
+
+	taskCfg := &TaskConfig{}
+	podSpec, err := initPodSpecPatch(
+		containerSpec, componentSpec, execInput,
+		27, "test", "run", "my-run-name", "1", "false", "false", taskCfg, false, false, "",
+		"metadata-grpc-service.kubeflow.svc.local",
+		"8080",
+	)
+	assert.Nil(t, err)
+
+	// Expect workspace volume mounted
+	if assert.Len(t, podSpec.Volumes, 1) {
+		assert.Equal(t, "kfp-workspace", podSpec.Volumes[0].Name)
+		if assert.NotNil(t, podSpec.Volumes[0].PersistentVolumeClaim) {
+			assert.Equal(t, "my-run-name-kfp-workspace", podSpec.Volumes[0].PersistentVolumeClaim.ClaimName)
+		}
+	}
+	if assert.Len(t, podSpec.Containers, 1) {
+		if assert.Len(t, podSpec.Containers[0].VolumeMounts, 1) {
+			assert.Equal(t, "kfp-workspace", podSpec.Containers[0].VolumeMounts[0].Name)
+			assert.Equal(t, "/kfp-workspace", podSpec.Containers[0].VolumeMounts[0].MountPath)
+		}
+	}
+
+	// Expect volumes to be captured in TaskConfig
+	if assert.Len(t, taskCfg.Volumes, 1) {
+		assert.Equal(t, "kfp-workspace", taskCfg.Volumes[0].Name)
+	}
+	if assert.Len(t, taskCfg.VolumeMounts, 1) {
+		assert.Equal(t, "kfp-workspace", taskCfg.VolumeMounts[0].Name)
+		assert.Equal(t, "/kfp-workspace", taskCfg.VolumeMounts[0].MountPath)
+	}
+}
+
 func Test_initPodSpecPatch_TaskConfig_Env_Passthrough_CaptureOnly(t *testing.T) {
 	proxy.InitializeConfigWithEmptyForTests()
 	containerSpec := &pipelinespec.PipelineDeploymentConfig_PipelineContainerSpec{
--- a/backend/src/v2/driver/resolve.go
+++ b/backend/src/v2/driver/resolve.go
@@ -33,6 +33,17 @@ import (
 
 var ErrResolvedParameterNull = errors.New("the resolved input parameter is null")
 
+// setWorkspaceFlag sets the _kfp_workspace metadata flag on the provided
+// runtime artifact when the producing execution is an ImporterWorkspace.
+func setWorkspaceFlag(execution *metadata.Execution, runtimeArtifact *pipelinespec.RuntimeArtifact) {
+	if execution.GetExecution().GetType() == string(metadata.ImporterWorkspaceExecutionTypeName) {
+		if runtimeArtifact.Metadata == nil {
+			runtimeArtifact.Metadata = &structpb.Struct{Fields: map[string]*structpb.Value{}}
+		}
+		runtimeArtifact.Metadata.Fields["_kfp_workspace"] = structpb.NewBoolValue(true)
+	}
+}
+
 // resolveUpstreamOutputsConfig is just a config struct used to store the input
 // parameters of the resolveUpstreamParameters and resolveUpstreamArtifacts
 // functions.
@@ -763,6 +774,8 @@ func resolveUpstreamArtifacts(cfg resolveUpstreamOutputsConfig) (*pipelinespec.A
 			if err != nil {
 				cfg.err(err)
 			}
+			// If produced by workspace importer, Set _kfp_workspace=True
+			setWorkspaceFlag(currentTask, runtimeArtifact)
 			// Base case
 			return &pipelinespec.ArtifactList{
 				Artifacts: []*pipelinespec.RuntimeArtifact{runtimeArtifact},
@@ -980,6 +993,8 @@ func collectContainerOutput(
 		if err != nil {
 			return nil, nil, cfg.err(err)
 		}
+		// If produced by workspace importer, Set _kfp_workspace=True
+		setWorkspaceFlag(currentTask, artifact)
 		glog.V(4).Infof("runtimeArtifact: %v", artifact)
 	} else {
 		_, outputParameters, err := currentTask.GetParameters()
--- a/sdk/python/kfp/compiler/pipeline_spec_builder.py
+++ b/sdk/python/kfp/compiler/pipeline_spec_builder.py
@@ -614,6 +614,10 @@ def build_importer_spec_for_task(
         metadata_protobuf_struct.update(task.importer_spec.metadata)
         importer_spec.metadata.CopyFrom(metadata_protobuf_struct)
 
+    # Emit download_to_workspace if set on the task
+    if getattr(task.importer_spec, 'download_to_workspace', False):
+        importer_spec.download_to_workspace = task.importer_spec.download_to_workspace
+
     if isinstance(task.importer_spec.artifact_uri,
                   pipeline_channel.PipelineChannel):
         importer_spec.artifact_uri.runtime_parameter = 'uri'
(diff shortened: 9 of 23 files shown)


backend/src/v2/component/importer_launcher.go
```diff
@@ -283,6 +293,40 @@ func (l *ImportLauncher) ImportSpecToMLMDArtifact(ctx context.Context) (artifact
...
+		workspaceRoot := filepath.Join(WorkspaceMountPath, ".artifacts")
+		if err := os.MkdirAll(workspaceRoot, 0755); err != nil {
+			return nil, fmt.Errorf("failed to create workspace directory %q: %w", workspaceRoot, err)
+		}
+		bucket, err := objectstore.OpenBucket(ctx, l.k8sClient, l.launcherV2Options.Namespace, bucketConfig)
+		if err != nil {
+			return nil, fmt.Errorf("failed to open bucket for uri %q: %w", artifactUri, err)
+		}
+		defer bucket.Close()
+		if err := objectstore.DownloadBlob(ctx, bucket, workspaceRoot, blobKey); err != nil {
```
mprahl: I don't think the workspace paths are quite correct. This is more in line with the design of having the same path but just under `/kfp-workspace/.artifacts`:

```diff
diff --git a/backend/src/v2/component/importer_launcher.go b/backend/src/v2/component/importer_launcher.go
index 7c11f9fbd..78e8a6d03 100644
--- a/backend/src/v2/component/importer_launcher.go
+++ b/backend/src/v2/component/importer_launcher.go
@@ -4,8 +4,6 @@ import (
 	"context"
 	"encoding/json"
 	"fmt"
-	"os"
-	"path/filepath"
 	"strings"
 
 	"github.com/kubeflow/pipelines/backend/src/common/util"
@@ -310,20 +308,22 @@ func (l *ImportLauncher) ImportSpecToMLMDArtifact(ctx context.Context) (artifact
 				bucketConfig.SessionInfo = &sess
 			}
 		}
+
+		localPath, err := LocalWorkspacePathForURI(artifactUri)
+		if err != nil {
+			return nil, fmt.Errorf("failed to get local path for uri %q: %w", artifactUri, err)
+		}
+
 		blobKey, err := bucketConfig.KeyFromURI(artifactUri)
 		if err != nil {
 			return nil, fmt.Errorf("failed to derive blob key from uri %q while downloading artifact into workspace: %w", artifactUri, err)
 		}
-		workspaceRoot := filepath.Join(WorkspaceMountPath, ".artifacts")
-		if err := os.MkdirAll(workspaceRoot, 0755); err != nil {
-			return nil, fmt.Errorf("failed to create workspace directory %q: %w", workspaceRoot, err)
-		}
 		bucket, err := objectstore.OpenBucket(ctx, l.k8sClient, l.launcherV2Options.Namespace, bucketConfig)
 		if err != nil {
 			return nil, fmt.Errorf("failed to open bucket for uri %q: %w", artifactUri, err)
 		}
 		defer bucket.Close()
-		if err := objectstore.DownloadBlob(ctx, bucket, workspaceRoot, blobKey); err != nil {
+		if err := objectstore.DownloadBlob(ctx, bucket, localPath, blobKey); err != nil {
 			return nil, fmt.Errorf("failed to download artifact to workspace: %w", err)
 		}
 	}
diff --git a/backend/src/v2/component/launcher_v2.go b/backend/src/v2/component/launcher_v2.go
index 95caf375b..ac670ca2d 100644
--- a/backend/src/v2/component/launcher_v2.go
+++ b/backend/src/v2/component/launcher_v2.go
@@ -865,16 +865,13 @@ func getPlaceholders(executorInput *pipelinespec.ExecutorInput) (placeholders ma
 		// If the artifact is marked as already in the workspace, map the workspace path.
 		if inputArtifact.GetMetadata() != nil {
 			if v, ok := inputArtifact.GetMetadata().GetFields()["_kfp_workspace"]; ok && v.GetBoolValue() {
-				bucketConfig, err := objectstore.ParseBucketConfigForArtifactURI(inputArtifact.Uri)
-				if err == nil {
-					blobKey, err := bucketConfig.KeyFromURI(inputArtifact.Uri)
-					if err == nil {
-						localPath := filepath.Join(WorkspaceMountPath, ".artifacts", blobKey)
-						key = fmt.Sprintf(`{{$.inputs.artifacts['%s'].path}}`, name)
-						placeholders[key] = localPath
-						continue
-					}
+				localPath, err := LocalWorkspacePathForURI(inputArtifact.Uri)
+				if err != nil {
+					return nil, fmt.Errorf("failed to get local workspace path for input artifact %q: %w", name, err)
 				}
+				key = fmt.Sprintf(`{{$.inputs.artifacts['%s'].path}}`, name)
+				placeholders[key] = localPath
+				continue
 			}
 		}
 
diff --git a/sdk/python/kfp/dsl/types/artifact_types.py b/sdk/python/kfp/dsl/types/artifact_types.py
index ae737b483..39ae42adb 100644
--- a/sdk/python/kfp/dsl/types/artifact_types.py
+++ b/sdk/python/kfp/dsl/types/artifact_types.py
@@ -94,39 +94,26 @@ class Artifact:
         self._set_path(path)
 
     def _get_path(self) -> Optional[str]:
-        # If the artifact is already present in the pipeline workspace, map to the workspace path.
-        # This is indicated by backend setting metadata['_kfp_workspace'] = True.
-        if self.metadata.get('_kfp_workspace') is True:
-            uri = self.uri or ''
-            for prefix in (RemotePrefix.GCS.value, RemotePrefix.MINIO.value,
-                           RemotePrefix.S3.value):
-                if uri.startswith(prefix):
-                    # Derive the object key relative to the bucket:
-                    # "<bucket>/<key>" -> blob_key == "<key>"
-                    without_scheme = uri[len(prefix):]
-                    parts = without_scheme.split('/', 1)
-                    blob_key = parts[1] if len(parts) == 2 else ''
-                    if blob_key:
-                        return os.path.join(WORKSPACE_MOUNT_PATH, '.artifacts',
-                                            blob_key)
-
-                    return os.path.join(WORKSPACE_MOUNT_PATH, '.artifacts')
+        local_path = self.uri
 
         if self.uri.startswith(RemotePrefix.GCS.value):
-            return _GCS_LOCAL_MOUNT_PREFIX + self.uri[len(RemotePrefix.GCS.value
-                                                         ):]
-        if self.uri.startswith(RemotePrefix.MINIO.value):
-            return _MINIO_LOCAL_MOUNT_PREFIX + self.uri[len(RemotePrefix.MINIO
-                                                            .value):]
-        if self.uri.startswith(RemotePrefix.S3.value):
-            return _S3_LOCAL_MOUNT_PREFIX + self.uri[len(RemotePrefix.S3.value
-                                                        ):]
-        if self.uri.startswith(RemotePrefix.OCI.value):
+            local_path = _GCS_LOCAL_MOUNT_PREFIX + self.uri[len(RemotePrefix.GCS.value                                             ):]
+        elif self.uri.startswith(RemotePrefix.MINIO.value):
+            local_path = _MINIO_LOCAL_MOUNT_PREFIX + self.uri[len(RemotePrefix.MINIO                                                 .value):]
+        elif self.uri.startswith(RemotePrefix.S3.value):
+            local_path = _S3_LOCAL_MOUNT_PREFIX + self.uri[len(RemotePrefix.S3.value                                             ):]
+        elif self.uri.startswith(RemotePrefix.OCI.value):
             escaped_uri = self.uri[len(RemotePrefix.OCI.value):].replace(
                 '/', '_')
-            return _OCI_LOCAL_MOUNT_PREFIX + escaped_uri
+            local_path = _OCI_LOCAL_MOUNT_PREFIX + escaped_uri
+
+        # If the artifact is already present in the pipeline workspace, map to the workspace path.
+        # This is indicated by backend setting metadata['_kfp_workspace'] = True.
+        if self.metadata.get('_kfp_workspace') is True:
+            local_path = os.path.join(WORKSPACE_MOUNT_PATH, ".artifacts", local_path.lstrip("/"))
+
         # uri == path for local execution
-        return self.uri
+        return local_path
 
     def _set_path(self, path: str) -> None:
         self.uri = convert_local_path_to_remote_path(path)
```

backend/src/v2/compiler/argocompiler/importer.go
```diff
@@ -91,6 +91,32 @@ func (c *workflowCompiler) addImporterTemplate() string {
 	if value, ok := os.LookupEnv(PublishLogsEnvVar); ok {
 		args = append(args, "--publish_logs", value)
 	}
+	// Add workspace volume only if the workflow defines a workspace PVC
+	hasWorkspacePVC := false
+	for _, pvc := range c.wf.Spec.VolumeClaimTemplates {
```
mprahl: This always mounts the PVC for all importers if the pipeline uses a workspace. Let's make this more conditional to avoid unnecessary PVC mounting which can cause pod scheduling issues. Something like this:

```diff
diff --git a/backend/src/v2/compiler/argocompiler/dag.go b/backend/src/v2/compiler/argocompiler/dag.go
index 73e3efc8b..c57193222 100644
--- a/backend/src/v2/compiler/argocompiler/dag.go
+++ b/backend/src/v2/compiler/argocompiler/dag.go
@@ -326,7 +326,8 @@ func (c *workflowCompiler) task(name string, task *pipelinespec.PipelineTaskSpec
 				// it's impossible to add a when condition based on driver outputs.
 				return nil, fmt.Errorf("triggerPolicy.condition on importer task is not supported")
 			}
-			importer, err := c.importerTask(name, task, taskSpecJson, inputs.parentDagID)
+
+			importer, err := c.importerTask(name, task, taskSpecJson, inputs.parentDagID, e.Importer.GetDownloadToWorkspace())
 			if err != nil {
 				return nil, err
 			}
diff --git a/backend/src/v2/compiler/argocompiler/importer.go b/backend/src/v2/compiler/argocompiler/importer.go
index 5c509bf4e..096481ee6 100644
--- a/backend/src/v2/compiler/argocompiler/importer.go
+++ b/backend/src/v2/compiler/argocompiler/importer.go
@@ -32,7 +32,7 @@ func (c *workflowCompiler) Importer(name string, componentSpec *pipelinespec.Com
 	return c.saveComponentImpl(name, importer)
 }
 
-func (c *workflowCompiler) importerTask(name string, task *pipelinespec.PipelineTaskSpec, taskJSON string, parentDagID string) (*wfapi.DAGTask, error) {
+func (c *workflowCompiler) importerTask(name string, task *pipelinespec.PipelineTaskSpec, taskJSON string, parentDagID string, downloadToWorkspace bool) (*wfapi.DAGTask, error) {
 	componentPlaceholder, err := c.useComponentSpec(task.GetComponentRef().GetName())
 	if err != nil {
 		return nil, err
@@ -43,7 +43,7 @@ func (c *workflowCompiler) importerTask(name string, task *pipelinespec.Pipeline
 	}
 	return &wfapi.DAGTask{
 		Name:     name,
-		Template: c.addImporterTemplate(),
+		Template: c.addImporterTemplate(downloadToWorkspace),
 		Arguments: wfapi.Arguments{Parameters: []wfapi.Parameter{{
 			Name:  paramTask,
 			Value: wfapi.AnyStringPtr(taskJSON),
@@ -60,11 +60,16 @@ func (c *workflowCompiler) importerTask(name string, task *pipelinespec.Pipeline
 	}, nil
 }
 
-func (c *workflowCompiler) addImporterTemplate() string {
+func (c *workflowCompiler) addImporterTemplate(downloadToWorkspace bool) string {
 	name := "system-importer"
+	if downloadToWorkspace {
+		name += "-workspace"
+	}
+
 	if _, alreadyExists := c.templates[name]; alreadyExists {
 		return name
 	}
+
 	args := []string{
 		"--executor_type", "importer",
 		"--task_spec", inputValue(paramTask),
@@ -91,18 +96,10 @@ func (c *workflowCompiler) addImporterTemplate() string {
 	if value, ok := os.LookupEnv(PublishLogsEnvVar); ok {
 		args = append(args, "--publish_logs", value)
 	}
-	// Add workspace volume only if the workflow defines a workspace PVC
-	hasWorkspacePVC := false
-	for _, pvc := range c.wf.Spec.VolumeClaimTemplates {
-		if pvc.Name == workspaceVolumeName {
-			hasWorkspacePVC = true
-			break
-		}
-	}
 
 	var volumeMounts []k8score.VolumeMount
 	var volumes []k8score.Volume
-	if hasWorkspacePVC {
+	if downloadToWorkspace {
 		volumeMounts = append(volumeMounts, k8score.VolumeMount{
 			Name:      workspaceVolumeName,
 			MountPath: component.WorkspaceMountPath,
```

backend/src/v2/component/importer_launcher.go
```diff
@@ -283,6 +293,40 @@ func (l *ImportLauncher) ImportSpecToMLMDArtifact(ctx context.Context) (artifact
...
+		// Resolve and attach session info from kfp-launcher config for the artifact provider
+		if cfg, cfgErr := config.FromConfigMap(ctx, l.k8sClient, l.launcherV2Options.Namespace); cfgErr != nil {
+			glog.Warningf("failed to load launcher config for workspace download: %v", cfgErr)
+		} else if cfg != nil {
+			if sess, sessErr := cfg.GetStoreSessionInfo(artifactUri); sessErr != nil {
+				glog.Warningf("failed to resolve store session info for %q: %v", artifactUri, sessErr)
+			} else {
+				bucketConfig.SessionInfo = &sess
+			}
+		}
```
mprahl: Let's avoid duplicating this code and move the code from the `Execute` method to a helper function.

backend/src/v2/component/importer_launcher.go
mprahl: Could you also log a message indicating the artifact is being downloaded?

test_data/sdk_compiled_pipelines/valid/critical/pipeline_with_importer_workspace.py
```diff
@@ -0,0 +1,56 @@
...
+            size='1Gi',
+            kubernetes=dsl.KubernetesWorkspaceConfig(
+                pvcSpecPatch={'storageClassName': 'standard'}
+            ),
+        ),
+    ),
+)
+def pipeline_with_importer_workspace() -> str:
+    ds = importer(
+        artifact_uri='minio://mlpipeline/sample/sample.txt',
```
mprahl: Please add a test case that downloads a directory as well.

mprahl:
The CI should be fixed by #12404

nsingla (commented):
I believe we can also add a test to the sdk local execution for this?

api/v2alpha1/pipeline_spec.proto
```diff
@@ -873,6 +873,9 @@ message PipelineDeploymentConfig {
 
     // Whether or not import an artifact regardless it has been imported before.
     bool reimport = 5;
```
nsingla, line 872: Just my ocd, but can we reorder these :)

VaniHaripriya: Done

backend/src/v2/component/importer_launcher.go
```diff
@@ -126,12 +127,17 @@ func (l *ImportLauncher) Execute(ctx context.Context) (err error) {
...
-		PodName:       l.launcherV2Options.PodName,
-		PodUID:        l.launcherV2Options.PodUID,
-		Namespace:     l.launcherV2Options.Namespace,
-		ExecutionType: metadata.ImporterExecutionTypeName,
-		ParentDagID:   l.importerLauncherOptions.ParentDagID,
+		TaskName:  l.task.GetTaskInfo().GetName(),
+		PodName:   l.launcherV2Options.PodName,
+		PodUID:    l.launcherV2Options.PodUID,
+		Namespace: l.launcherV2Options.Namespace,
+		ExecutionType: func() metadata.ExecutionType {
```
mprahl: Optional: No need for this temporary function declaration. This code can live in the main function just prior to defining `ecfg`.

backend/src/v2/component/launcher_v2.go
```diff
@@ -856,6 +862,20 @@ func getPlaceholders(executorInput *pipelinespec.ExecutorInput) (placeholders ma
...
+		// If the artifact is marked as already in the workspace, map to the workspace path
+		// with the same shape as LocalPathForURI, but rebased under the workspace mount.
+		if inputArtifact.GetMetadata() != nil {
+			if v, ok := inputArtifact.GetMetadata().GetFields()["_kfp_workspace"]; ok && v.GetBoolValue() {
+				localPath, lerr := LocalWorkspacePathForURI(inputArtifact.Uri)
+				if lerr != nil {
+					return nil, fmt.Errorf("failed to get local workspace path for input artifact %q: %w", name, lerr)
+				}
+				key = fmt.Sprintf(`{{$.inputs.artifacts['%s'].path}}`, name)
+				placeholders[key] = localPath
```
mprahl, line 876: Could you update `needsWorkspaceMount` so that `if strings.HasPrefix(strVal.StringValue, component.WorkspaceMountPath) {` is `if strings.Contains(strVal.StringValue, component.WorkspaceMountPath) {`?

This is because the placeholder replacement logic is much less strict and we need `needsWorkspaceMount` to match.

backend/src/v2/driver/resolve.go
```diff
@@ -33,6 +33,17 @@ import (
...
 var ErrResolvedParameterNull = errors.New("the resolved input parameter is null")
 
+// setWorkspaceFlag sets the _kfp_workspace metadata flag on the provided
+// runtime artifact when the producing execution is an ImporterWorkspace.
+func setWorkspaceFlag(execution *metadata.Execution, runtimeArtifact *pipelinespec.RuntimeArtifact) {
+	if execution.GetExecution().GetType() == string(metadata.ImporterWorkspaceExecutionTypeName) {
+		if runtimeArtifact.Metadata == nil {
+			runtimeArtifact.Metadata = &structpb.Struct{Fields: map[string]*structpb.Value{}}
+		}
+		runtimeArtifact.Metadata.Fields["_kfp_workspace"] = structpb.NewBoolValue(true)
```
mprahl, line 43: Please update `toMLMDArtifact` to skip setting `_kfp_workspace`. We want this flag to be only available at runtime for the executor to know the correct path, not stored in the MLMD database. The reason is that the state of it being in the `_kfp_workspace` is specific to this pipeline run, not generic to the artifact itself.

We should rely on the execution type of the importer to determine this value at runtime.

test_data/sdk_compiled_pipelines/valid/critical/pipeline_with_importer_workspace.py
```diff
@@ -0,0 +1,152 @@
...
+    if os.path.isdir(path):
+        names = []
+        for root, _, files in os.walk(path):
+            for name in files:
+                names.append(os.path.relpath(os.path.join(root, name), path))
+        names.sort()
+        result = ",".join(names) if names else "EMPTY_DIRECTORY"
+        print(f"Found {len(names)} files: {result}")
+        return result
+    elif os.path.isfile(path):
```
mprahl: It should fail if it's not a directory right?

sdk/python/test/local_execution/local_execution_test.py
```diff
@@ -167,6 +169,12 @@ def idfn(val):
         pipeline_func_args=None,
         expected_output=None,
     ),
+    TestData(
+        name='Importer Workspace',
+        pipeline_func=importer_workspace_pipeline,
```
mprahl, line 180: I don't think the local pipeline runner will honor the `download_to_workspace` option so we probably can remove this for now.

mprahl: If you want to add local pipeline runner, you can try to base it off of this AI generated code. I tried to keep it simple:

```diff
diff --git a/sdk/python/kfp/dsl/types/artifact_types.py b/sdk/python/kfp/dsl/types/artifact_types.py
index 59381592b..ee8824d24 100644
--- a/sdk/python/kfp/dsl/types/artifact_types.py
+++ b/sdk/python/kfp/dsl/types/artifact_types.py
@@ -98,7 +98,7 @@ class Artifact:
     def _get_path(self) -> Optional[str]:
         local_path = self.uri
 
-        if self.custom_path is not '':
+        if self.custom_path:
             local_path = self._get_custom_path()
         elif self.uri.startswith(RemotePrefix.GCS.value):
             local_path = _GCS_LOCAL_MOUNT_PREFIX + self.uri[len(RemotePrefix.GCS
diff --git a/sdk/python/kfp/local/importer_handler.py b/sdk/python/kfp/local/importer_handler.py
index 44c6a407f..47be1e751 100644
--- a/sdk/python/kfp/local/importer_handler.py
+++ b/sdk/python/kfp/local/importer_handler.py
@@ -13,13 +13,19 @@
 # limitations under the License.
 """Code for running a dsl.importer locally."""
 import logging
+import os
+import shutil
+import uuid
 from typing import Any, Dict, Tuple
+from urllib.parse import urlparse
 import warnings
 
 from google.protobuf import json_format
 from kfp import dsl
+from kfp.dsl import constants as dsl_constants
 from kfp.dsl.types import artifact_types
 from kfp.dsl.types import type_utils
+from kfp.local import config
 from kfp.local import logging_utils
 from kfp.local import placeholder_utils
 from kfp.local import status
@@ -28,6 +34,43 @@ from kfp.pipeline_spec import pipeline_spec_pb2
 Outputs = Dict[str, Any]
 
 
+def _get_workspace_root() -> str:
+    cfg = config.LocalExecutionConfig.instance
+    if cfg is None or not cfg.workspace_root:
+        raise RuntimeError(
+            'Workspace not configured. Initialize with workspace_root parameter:\n'
+            "local.init(runner=local.SubprocessRunner(), workspace_root='/path/to/workspace')"
+        )
+    workspace_root = cfg.workspace_root
+    if not os.path.isabs(workspace_root):
+        workspace_root = os.path.abspath(workspace_root)
+    return workspace_root
+
+
+def _copy_local_artifact_to_workspace(source_path: str, workspace_root: str,
+                                      component_name: str) -> str:
+    source_path = os.path.abspath(source_path)
+    if not os.path.exists(source_path):
+        raise FileNotFoundError(source_path)
+
+    destination_dir = os.path.join(workspace_root, 'importer', component_name,
+                                   uuid.uuid4().hex)
+    os.makedirs(destination_dir, exist_ok=True)
+
+    if os.path.isdir(source_path):
+        destination_path = os.path.join(destination_dir,
+                                        os.path.basename(source_path.rstrip(os.sep)))
+        if os.path.exists(destination_path):
+            shutil.rmtree(destination_path)
+        shutil.copytree(source_path, destination_path)
+    else:
+        destination_path = os.path.join(destination_dir,
+                                        os.path.basename(source_path))
+        shutil.copy2(source_path, destination_path)
+
+    return destination_path
+
+
 def run_importer(
     pipeline_resource_name: str,
     component_name: str,
@@ -90,13 +133,40 @@ def run_importer(
     )
     ArtifactCls = get_artifact_class_from_schema_title(
         executor_spec.importer.type_schema.schema_title)
-    outputs = {
-        'artifact': ArtifactCls(
-            name='artifact',
-            uri=uri,
-            metadata=metadata,
-        )
-    }
+    artifact = ArtifactCls(
+        name='artifact',
+        uri=uri,
+        metadata=metadata,
+    )
+
+    if executor_spec.importer.download_to_workspace:
+        parsed_uri = urlparse(uri)
+        scheme = parsed_uri.scheme
+        if scheme not in ('', 'file'):
+            warnings.warn(
+                "download_to_workspace only supports local file paths when running locally. Skipping workspace copy for '%s'."
+                % uri)
+        else:
+            source_path = parsed_uri.path if scheme == 'file' else uri
+            try:
+                workspace_root = _get_workspace_root()
+                host_path = _copy_local_artifact_to_workspace(
+                    source_path, workspace_root, component_name)
+
+                cfg = config.LocalExecutionConfig.instance
+                if isinstance(cfg.runner, config.DockerRunner):
+                    rel_path = os.path.relpath(host_path, workspace_root)
+                    container_path = os.path.join(dsl.constants.WORKSPACE_MOUNT_PATH,
+                                                  rel_path)
+                    artifact.custom_path = container_path
+                else:
+                    artifact.custom_path = host_path
+            except FileNotFoundError:
+                warnings.warn(
+                    f"download_to_workspace expected a local file at '{source_path}', but it was not found. Skipping workspace copy."
+                )
+
+    outputs = {'artifact': artifact}
     with logging_utils.local_logger_context():
         logging.info(
             f'Task {task_name_for_logs} finished with status {logging_utils.format_status(status.Status.SUCCESS)}'
diff --git a/sdk/python/kfp/local/importer_handler_test.py b/sdk/python/kfp/local/importer_handler_test.py
index 56a4b4467..0ceb06478 100644
--- a/sdk/python/kfp/local/importer_handler_test.py
+++ b/sdk/python/kfp/local/importer_handler_test.py
@@ -12,10 +12,13 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 """Tests for importer_handler_test.py."""
+import os
+import pathlib
 import unittest
 
 from google.protobuf import json_format
 from kfp import dsl
+from kfp.local import config as local_config
 from kfp.local import importer_handler
 from kfp.local import status
 from kfp.local import testing_utilities
@@ -94,6 +97,158 @@ class TestRunImporter(testing_utilities.LocalRunnerEnvironmentTestCase):
                          expected_artifact.metadata)
         self.assertEqual(task_status, status.Status.SUCCESS)
 
+    def test_download_to_workspace_local_file(self):
+        workspace_root = os.path.join(self.temp_dir.name, 'workspace')
+        pipeline_root = os.path.join(self.temp_dir.name, 'pipeline-root')
+        os.makedirs(workspace_root, exist_ok=True)
+        os.makedirs(pipeline_root, exist_ok=True)
+        local_config.init(
+            runner=local_config.SubprocessRunner(),
+            pipeline_root=pipeline_root,
+            workspace_root=workspace_root,
+        )
+        source_file = pathlib.Path(self.temp_dir.name) / 'input.txt'
+        source_content = 'hello workspace'
+        source_file.write_text(source_content)
+
+        component_spec_dict = {
+            'inputDefinitions': {
+                'parameters': {
+                    'uri': {
+                        'parameterType': 'STRING'
+                    }
+                }
+            },
+            'outputDefinitions': {
+                'artifacts': {
+                    'artifact': {
+                        'artifactType': {
+                            'schemaTitle': 'system.Dataset',
+                            'schemaVersion': '0.0.1'
+                        }
+                    }
+                }
+            },
+            'executorLabel': 'exec-importer'
+        }
+        component_spec = pipeline_spec_pb2.ComponentSpec()
+        json_format.ParseDict(component_spec_dict, component_spec)
+
+        executor_spec_dict = {
+            'importer': {
+                'artifactUri': {
+                    'runtimeParameter': 'uri'
+                },
+                'typeSchema': {
+                    'schemaTitle': 'system.Dataset',
+                    'schemaVersion': '0.0.1'
+                },
+                'downloadToWorkspace': True
+            }
+        }
+        executor_spec = pipeline_spec_pb2.PipelineDeploymentConfig.ExecutorSpec(
+        )
+        json_format.ParseDict(executor_spec_dict, executor_spec)
+
+        outputs, task_status = importer_handler.run_importer(
+            pipeline_resource_name='my-pipeline',
+            component_name='comp-importer',
+            component_spec=component_spec,
+            executor_spec=executor_spec,
+            arguments={'uri': str(source_file)},
+            pipeline_root=pipeline_root,
+            unique_pipeline_id='123',
+        )
+
+        artifact = outputs['artifact']
+        self.assertEqual(artifact.uri, str(source_file))
+        self.assertTrue(artifact.custom_path)
+        self.assertTrue(artifact.custom_path.startswith(workspace_root))
+        self.assertEqual(pathlib.Path(artifact.custom_path).read_text(),
+                         source_content)
+        self.assertTrue(os.path.exists(artifact.custom_path))
+        self.assertEqual(artifact.path, artifact.custom_path)
+        self.assertEqual(task_status, status.Status.SUCCESS)
+
+        local_config.LocalExecutionConfig.instance = None
+
+    def test_download_to_workspace_local_file_docker(self):
+        workspace_root = os.path.join(self.temp_dir.name, 'workspace')
+        pipeline_root = os.path.join(self.temp_dir.name, 'pipeline-root')
+        os.makedirs(workspace_root, exist_ok=True)
+        os.makedirs(pipeline_root, exist_ok=True)
+        local_config.init(
+            runner=local_config.DockerRunner(),
+            pipeline_root=pipeline_root,
+            workspace_root=workspace_root,
+        )
+        source_file = pathlib.Path(self.temp_dir.name) / 'input.txt'
+        source_content = 'hello docker workspace'
+        source_file.write_text(source_content)
+
+        component_spec_dict = {
+            'inputDefinitions': {
+                'parameters': {
+                    'uri': {
+                        'parameterType': 'STRING'
+                    }
+                }
+            },
+            'outputDefinitions': {
+                'artifacts': {
+                    'artifact': {
+                        'artifactType': {
+                            'schemaTitle': 'system.Dataset',
+                            'schemaVersion': '0.0.1'
+                        }
+                    }
+                }
+            },
+            'executorLabel': 'exec-importer'
+        }
+        component_spec = pipeline_spec_pb2.ComponentSpec()
+        json_format.ParseDict(component_spec_dict, component_spec)
+
+        executor_spec_dict = {
+            'importer': {
+                'artifactUri': {
+                    'runtimeParameter': 'uri'
+                },
+                'typeSchema': {
+                    'schemaTitle': 'system.Dataset',
+                    'schemaVersion': '0.0.1'
+                },
+                'downloadToWorkspace': True
+            }
+        }
+        executor_spec = pipeline_spec_pb2.PipelineDeploymentConfig.ExecutorSpec(
+        )
+        json_format.ParseDict(executor_spec_dict, executor_spec)
+
+        outputs, task_status = importer_handler.run_importer(
+            pipeline_resource_name='docker-pipeline',
+            component_name='docker-importer',
+            component_spec=component_spec,
+            executor_spec=executor_spec,
+            arguments={'uri': str(source_file)},
+            pipeline_root=pipeline_root,
+            unique_pipeline_id='456',
+        )
+
+        artifact = outputs['artifact']
+        expected_prefix = os.path.join(dsl.constants.WORKSPACE_MOUNT_PATH,
+                                       'importer')
+        self.assertTrue(artifact.custom_path.startswith(expected_prefix))
+
+        host_copy_path = os.path.join(workspace_root, 'importer',
+                                      'docker-importer')
+        copied_files = list(pathlib.Path(host_copy_path).rglob('input.txt'))
+        self.assertEqual(len(copied_files), 1)
+        self.assertEqual(copied_files[0].read_text(), source_content)
+        self.assertEqual(task_status, status.Status.SUCCESS)
+
+        local_config.LocalExecutionConfig.instance = None
+
     def test_uri_constant(self):
         component_spec_dict = {
             'inputDefinitions': {

```

VaniHaripriya: Agree, I verified that using the local runner we wont be able to test the download_to_workspace option to validate the PVC creation and downloading artifact to the /kfp-workspace path.

mprahl: @VaniHaripriya for the local support, we can just leverage the existing importer behavior but copy the file/dir to the workspace root like the diff shows.

VaniHaripriya: Sorry I didn't notice the latest comment earlier, I will add the support to run the pipelines using dsl.importer locally.

mprahl: @VaniHaripriya you don't have to in this PR, but I figured you could try it. If it ends up being too hard, we can do it in another issue.

backend/src/v2/component/importer_launcher.go
```diff
@@ -283,9 +292,54 @@ func (l *ImportLauncher) ImportSpecToMLMDArtifact(ctx context.Context) (artifact
...
 
+// resolveBucketConfigForURI parses bucket configuration for a given artifact URI and
+// attaches session information from the kfp-launcher ConfigMap when available.
+func (l *ImportLauncher) resolveBucketConfigForURI(ctx context.Context, uri string) (*objectstore.Config, error) {
+	bucketConfig, err := objectstore.ParseBucketConfigForArtifactURI(uri)
+	if err != nil {
+		return nil, fmt.Errorf("failed to parse bucket config while resolving uri %q: %w", uri, err)
+	}
+	// Resolve and attach session info from kfp-launcher config for the artifact provider
+	if cfg, cfgErr := config.FromConfigMap(ctx, l.k8sClient, l.launcherV2Options.Namespace); cfgErr != nil {
```
alyssacgoins: Is there a formatting or stylistic reason why `cfgErr` and `sessErr` are used instead of `err`? I found it a little confusing, and I was wondering if the syntax choice is designating these errors as special.

backend/src/v2/driver/driver.go
```diff
@@ -461,7 +461,7 @@ func needsWorkspaceMount(executorInput *pipelinespec.ExecutorInput) bool {
 				return true
 			}
 
-			if strings.HasPrefix(strVal.StringValue, component.WorkspaceMountPath) {
+			if strings.Contains(strVal.StringValue, component.WorkspaceMountPath) {
```
alyssacgoins, line 464: Does updating `string.HasPrefix()` to `strings.Contains()` have any side consequences for other uses of this function?

backend/src/v2/driver/driver_test.go
```diff
@@ -1189,6 +1189,68 @@ func TestWorkspaceMount_PassthroughVolumes_ApplyAndCapture(t *testing.T) {
...
+		}
+	}
+	if assert.Len(t, podSpec.Containers, 1) {
+		if assert.Len(t, podSpec.Containers[0].VolumeMounts, 1) {
+			assert.Equal(t, "kfp-workspace", podSpec.Containers[0].VolumeMounts[0].Name)
+			assert.Equal(t, "/kfp-workspace", podSpec.Containers[0].VolumeMounts[0].MountPath)
+		}
+	}
+
+	// Also captured to TaskConfig
```
alyssacgoins: Would you be able to update this sentence to the "Expect _______" format in the comment directly above ("Expect workspace volume mounted")?

test_data/sdk_compiled_pipelines/valid/critical/pipeline_with_importer_workspace.py
```diff
@@ -0,0 +1,148 @@
...
+    
+    # Return outputs for validation
+    Outputs = NamedTuple('Outputs', [('train_result', str), ('dir_result', str)])
+    return Outputs(stage.outputs['train_result'], stage.outputs['dir_result'])
+
+
+if __name__ == '__main__':
+    compiler.Compiler().compile(
+        pipeline_func=pipeline_with_importer_workspace,
+        package_path=__file__.replace('.py', '.yaml'))
```
alyssacgoins: EOF

test_data/sdk_compiled_pipelines/valid/critical/pipeline_with_importer_workspace.py
```diff
@@ -0,0 +1,148 @@
...
+    # Produce a directory artifact and compute its runtime URI
+    dir_writer = write_dir_artifact()
+    dir_uri = get_uri(d=dir_writer.outputs["out_ds"])    
+    
+    # Import and consume inside a nested sub-pipeline
+    stage = import_stage(file_uri=file_uri.output, dir_uri=dir_uri.output)
+    
+    # Return outputs for validation
+    Outputs = NamedTuple('Outputs', [('train_result', str), ('dir_result', str)])
+    return Outputs(stage.outputs['train_result'], stage.outputs['dir_result'])
```
alyssacgoins, line 142: Does the `Outputs()` call validate the outputs here?

backend/src/v2/driver/driver.go
VaniHaripriya: It won’t have any impact, as this is only used to determine whether the task requires mounting the shared workspace PVC.

test_data/sdk_compiled_pipelines/valid/critical/pipeline_with_importer_workspace.py
VaniHaripriya: No.. it just returns them as the pipeline’s results.

backend/src/v2/component/importer_launcher.go
VaniHaripriya: No specific reason, updated them to `err` to keep it consistent.

backend/src/v2/component/launcher_v2_test.go
```diff
@@ -176,6 +177,29 @@ func Test_executeV2_publishLogs(t *testing.T) {
...
+					},
+				},
+			},
+		},
+	}
+	ph, err := getPlaceholders(execIn)
+	if err != nil {
+		t.Fatalf("getPlaceholders error: %v", err)
+	}
+	actual := ph["{{$.inputs.artifacts['data'].path}}"]
```
mprahl, line 196: Good catch for supporting container components.

sdk/python/kfp/local/importer_handler.py
```diff
@@ -90,13 +133,53 @@ def run_importer(
...
+                used_base = 'workspace'
+                try:
+                    host_path = _copy_local_artifact_to_workspace(
+                        source_path, workspace_root, component_name)
+                except PermissionError:
+                    if isinstance(cfg.runner, config.DockerRunner):
+                        # fallback to pipeline_root for Docker
+                        used_base = 'pipeline_root'
+                        host_path = _copy_local_artifact_to_workspace(
+                            source_path, cfg.pipeline_root, component_name)
```
mprahl: I thought workspace_root was mounted as a volume for the DockerRunner. Is this fallback needed?

sdk/python/kfp/local/importer_handler.py
```diff
@@ -90,13 +133,53 @@ def run_importer(
...
+                        container_path = os.path.join(
+                            dsl_constants.WORKSPACE_MOUNT_PATH, rel_path)
+                        artifact.custom_path = container_path
+                    else:
+                        artifact.custom_path = host_path
+                else:
+                    artifact.custom_path = host_path
+            except FileNotFoundError:
+                warnings.warn(
+                    f"download_to_workspace expected a local file at '{source_path}', but it was not found. Skipping workspace copy."
```
mprahl: It might be better to fail the pipeline run in this case since it would cause unexpected issues in subsequent pipeline steps.

sdk/python/kfp/local/importer_handler_test.py
```diff
@@ -94,6 +98,158 @@ def test_uri_from_upstream(self):
...
+        local_config.init(
+            runner=local_config.SubprocessRunner(),
+            pipeline_root=pipeline_root,
+            workspace_root=workspace_root,
+        )
+        source_file = pathlib.Path(self.temp_dir.name) / 'input.txt'
+        source_content = 'hello workspace'
+        source_file.write_text(source_content)
+
+        component_spec_dict = {
```
mprahl: Could we just define this as a pipeline and run it through the normal SDK paths for local pipelines?

sdk/python/kfp/local/importer_handler_test.py
```diff
@@ -94,6 +98,158 @@ def test_uri_from_upstream(self):
...
+        local_config.init(
+            runner=local_config.DockerRunner(),
+            pipeline_root=pipeline_root,
+            workspace_root=workspace_root,
+        )
+        source_file = pathlib.Path(self.temp_dir.name) / 'input.txt'
+        source_content = 'hello docker workspace'
+        source_file.write_text(source_content)
+
+        component_spec_dict = {
```
mprahl: Same comment as above. Let's define a minimal pipeline and run it with the KFP SDK Docker runner directly.

sdk/python/kfp/local/importer_handler_test.py
mprahl: It looks like it's tested with a real pipeline below. Just make sure it runs for subprocess runner and Docker runner. Then these tests can be removed. I have a suspicion some of that fallback logic above is from these tests.

sdk/python/kfp/compiler/pipeline_spec_builder.py
```diff
@@ -614,6 +614,10 @@ def build_importer_spec_for_task(
         metadata_protobuf_struct.update(task.importer_spec.metadata)
         importer_spec.metadata.CopyFrom(metadata_protobuf_struct)
 
+    # Emit download_to_workspace if set on the task
+    if getattr(task.importer_spec, 'download_to_workspace', False):
+        importer_spec.download_to_workspace = task.importer_spec.download_to_workspace
```
mprahl, line 619: We need the pipeline compilation to fail if the pipelineconfig doesn't specify a workspace.

sdk/python/kfp/local/importer_handler.py
VaniHaripriya: The fallback was added because, in CI, the process couldn’t write under workspace_root, so creating <workspace_root>/importer/... raised PermissionError. This error originated [here](https://github.com/kubeflow/pipelines/pull/12353/files#diff-ff42c148abf4b597d5faf044f61dbc0a77e38f14e712e53a523aa7336bf4d57fR58) .

sdk/python/kfp/local/importer_handler_test.py
VaniHaripriya: Removed these tests,  E2E test with existing pipeline using local runner is already added [here](https://github.com/kubeflow/pipelines/pull/12353/files#diff-420e5b82704dfa46ee385b24e5933a91c1e322173387c925eb031d75b3ae469cR172).

test_data/sdk_compiled_pipelines/valid/critical/pipeline_with_importer_workspace.py
```diff
@@ -0,0 +1,148 @@
...
+            for name in files:
+                names.append(os.path.relpath(os.path.join(root, name), path))
+        names.sort()
+        result = ",".join(names) if names else "EMPTY_DIRECTORY"
+        print(f"Found {len(names)} files: {result}")
+        return result
+    
+    raise ValueError(f"Not a directory: {path}")
+
+@dsl.component(base_image='python:3.9')
```
nsingla: I believe we have officially upgraded our python version to 3.11, and just recently, we switched all base images to `registry.access.redhat.com/ubi9/python-311:latest`

test_data/sdk_compiled_pipelines/valid/critical/pipeline_with_importer_workspace.py
nsingla: can we add some assertion to validate that the content is not empty and it actually did download to/from workspace

test_data/sdk_compiled_pipelines/valid/critical/pipeline_with_importer_workspace.py
VaniHaripriya: Updated the base image to `registry.access.redhat.com/ubi9/python-311:latest`

sdk/python/kfp/local/importer_handler_test.py
```diff
@@ -94,6 +94,8 @@ def test_uri_from_upstream(self):
                          expected_artifact.metadata)
         self.assertEqual(task_status, status.Status.SUCCESS)
 
+    # Removed unit-level workspace copy tests; covered by local runner E2E tests.
+
```
mprahl:
```suggestion
```

sdk/python/kfp/local/importer_handler.py
```diff
@@ -90,13 +133,53 @@ def run_importer(
...
+        else:
+            source_path = parsed_uri.path if scheme == 'file' else uri
+            try:
+                workspace_root = _get_workspace_root()
+                cfg = config.LocalExecutionConfig.instance
+                used_base = 'workspace'
+                try:
+                    host_path = _copy_local_artifact_to_workspace(
+                        source_path, workspace_root, component_name)
+                except PermissionError:
```
mprahl: I ran it locally with this small diff and I didn't encounter any permission errors:
```diff
diff --git a/sdk/python/test/local_execution/local_execution_test.py b/sdk/python/test/local_execution/local_execution_test.py
index bf43a8c64..2487d7279 100644
--- a/sdk/python/test/local_execution/local_execution_test.py
+++ b/sdk/python/test/local_execution/local_execution_test.py
@@ -234,7 +234,7 @@ class TestDockerRunner:
         Path(ws_root).mkdir(exist_ok=True)
         Path(pipeline_root).mkdir(exist_ok=True)
         local.init(
-            runner=local.DockerRunner(),
+            runner=local.DockerRunner(user=f'{os.getuid()}'),
             raise_on_error=True,
             workspace_root=ws_root,
             pipeline_root=pipeline_root)
```

I think let's apply that diff and then rather than fallback, we just provide the user with a good error message indicating that there maybe a mismatch between the host user and container and they can either explicitly set the Docker container UID like above or provider more permissive permissions on the host folder. I don't want the behavior to be inconsistent for how workspace artifacts are downloaded.

VaniHaripriya: Thank you @mprahl !! this looks better, updated it.

sdk/python/kfp/local/importer_handler.py
```diff
@@ -90,13 +133,53 @@ def run_importer(
...
+        if scheme not in ('', 'file'):
+            warnings.warn(
+                "download_to_workspace only supports local file paths when running locally. Skipping workspace copy for '%s'."
+                % uri)
+        else:
+            source_path = parsed_uri.path if scheme == 'file' else uri
+            try:
+                workspace_root = _get_workspace_root()
+                cfg = config.LocalExecutionConfig.instance
+                used_base = 'workspace'
```
mprahl: I don't think we need the used_base variable anymore.

VaniHaripriya: Agree..removed the flag and updated the code.

sdk/python/test/local_execution/local_execution_test.py
```diff
@@ -244,11 +252,32 @@ def setup_and_teardown(self):
...
     def test_execution(self, test_data: TestData):
-        if test_data.pipeline_func_args is not None:
-            pipeline_task = test_data.pipeline_func(
-                **test_data.pipeline_func_args)
+        if test_data.name == 'Importer Workspace':
+            ws_root = f'{ws_root_base}_docker'
+            pipeline_root = f'{pipeline_root_base}_docker'
+            if os.path.isdir(ws_root):
+                shutil.rmtree(ws_root, ignore_errors=True)
+            Path(ws_root).mkdir(parents=True, exist_ok=True)
```
mprahl: I think this is still handled in setup_and_teardown so this might be duplicating.

sdk/python/test/local_execution/local_execution_test.py
```diff
@@ -244,11 +252,32 @@ def setup_and_teardown(self):
...
+            local.init(
+                runner=local.DockerRunner(user=f'{os.getuid()}'),
```
mprahl: Can we enhance TestData to accept a new optional arg of `docker_runner_args` to make this reusable?

mprahl (approved):
/approve
/lgtm

PR #12356, opened 2025-10-13
[feature] Improve typing for @dsl.component to support strict LSPs

### LINKED ISSUE #12354: [feature] Fix typing of component-decorator to avoid errors in strict LSPs
### Feature Area

/area sdk

/area components

I'd like to see the `kfp` SDK improved with better typing for the `@dsl.component` decorator, as it seems to give both inconsistent and incorrect types.

Pointing to the [discussion I brought up](https://github.com/kubeflow/pipelines/discussions/12318), it seems like the decorator provides incorrect types. If using a stricter LSP such as Pyright (e.g. more strict than VSCode default of Pylance), this leads to many errors during development, as well as failure of autocomplete suggestions.

I'll give one example, where the use of parentheses (or lack thereof) changes the type of the decorated function;

Image

Image

I am slower at developing due to lack of autocomplete & "Go to Definition" behaviour, making it harder to read the source code and understand what objects are actually being used where.

I completely disable all type-checking for KFP specifically. I also mainly copy-paste exact examples from docs, as my IDE cannot help much with customisation due to lack of proper typing.

---

Love this idea? Give it a 👍.

aniketpati1121:
### Summary  fixes #12354 
This PR improves typing for the @dsl.component decorator in Kubeflow Pipelines SDK to fix autocomplete and type errors in strict LSPs like Pyright.

### Before
- func argument was declared as `Optional[Callable]` → IDE could not infer argument or return types.
- Autocomplete and go-to-definition did not work properly.
- Strict LSP errors appeared.

### After
- func argument declared as `Optional[Callable[P, R]]` with ParamSpec and TypeVar.
- Return type annotated as `Callable[P, R]`.
- functools.partial handling ensures decorator with parentheses works correctly.
- IDE autocomplete, go-to-definition, and strict type checks work properly.


droctothorpe:
Can you rebase and address conflicts?

PR #12357, opened 2025-10-13
fix(test/backend): Fix pod to pod TLS

mprahl:

It seems that the E2E tests were not running due to syntax errors in the workflows. By fixing those, it uncovered some issues in the pod to pod TLS implementation.

See each commit for more details.

**Checklist:**

### DIFF
--- a/backend/src/v2/compiler/argocompiler/importer.go
+++ b/backend/src/v2/compiler/argocompiler/importer.go
@@ -86,6 +86,19 @@ func (c *workflowCompiler) addImporterTemplate() string {
 	if c.cacheDisabled {
 		args = append(args, "--cache_disabled")
 	}
+	if c.mlPipelineTLSEnabled {
+		args = append(args, "--ml_pipeline_tls_enabled")
+	}
+	if common.GetMetadataTLSEnabled() {
+		args = append(args, "--metadata_tls_enabled")
+	}
+
+	setCABundle := false
+	if common.GetCaBundleSecretName() != "" && (c.mlPipelineTLSEnabled || common.GetMetadataTLSEnabled()) {
+		args = append(args, "--ca_cert_path", common.TLSCertCAPath)
+		setCABundle = true
+	}
+
 	if value, ok := os.LookupEnv(PipelineLogLevelEnvVar); ok {
 		args = append(args, "--log_level", value)
 	}
@@ -113,7 +126,7 @@ func (c *workflowCompiler) addImporterTemplate() string {
 	}
 
 	// If TLS is enabled (apiserver or metadata), add the custom CA bundle to the importer template.
-	if c.mlPipelineTLSEnabled || common.GetMetadataTLSEnabled() {
+	if setCABundle {
 		ConfigureCustomCABundle(importerTemplate)
 	}
 	c.templates[name] = importerTemplate


.github/actions/deploy/action.yml
```diff
@@ -60,7 +60,7 @@ runs:
     - name: Load Docker Images
       shell: bash
       run: |
-        APPS=("apiserver" "driver" "launcher" "scheduledworkflow" "persistenceagent" "frontend")
+        APPS=("apiserver" "driver" "launcher" "scheduledworkflow" "persistenceagent" "frontend" "metadata-writer")
```
nsingla, line 63: just for curiosity, what's the reason for adding this component?

.github/workflows/api-server-tests.yml
```diff
@@ -102,9 +102,11 @@ jobs:
       - name: Configure Cluster CA Cert for TLS-Enabled Tests
         shell: bash
         id: configure-cluster-ca-cert
-        if: ${{ matrix.pod_to_pod_tls_enabled == "true"}}
+        if: ${{ matrix.pod_to_pod_tls_enabled == 'true'}}
         run: |
-          kubectl get secret kfp-api-tls-cert -n kubeflow -o jsonpath='{.data.ca\.crt}' | base64 -d > "${{ env.API_TESTS_DIR }}"/ca.crt
+          CA_CERT_FILE="${GITHUB_WORKSPACE}/ca.crt"
+          kubectl get secret kfp-api-tls-cert -n kubeflow -o jsonpath='{.data.ca\.crt}' | base64 -d > "$CA_CERT_FILE"
+          echo "TLS certificate retrieved and saved to $CA_CERT_FILE"
```
nsingla: can we rather just set env variable CA_CERT_FILE here?

.github/workflows/api-server-tests.yml
```diff
@@ -102,9 +102,11 @@ jobs:
...
 
       - name: Configure Input Variables
```
nsingla, line 109: The intent of this was to configure test variables based on input from manual trigger/dispatch

.github/workflows/api-server-tests.yml
```diff
@@ -125,8 +127,9 @@ jobs:
             PROXY=false
           fi
           
+          CA_CERT_PATH=""
```
nsingla: This can become redundant based on the above comment

.github/workflows/e2e-test.yml
```diff
@@ -132,8 +131,9 @@ jobs:
             NAMESPACE=${{ inputs.namespace }}
           fi
           
+          CA_CERT_PATH=""
           if [ "${{ matrix.pod_to_pod_tls_enabled }}" == "true" ]; then
-            CA_CERT_PATH="ca.crt"
+            CA_CERT_PATH="${GITHUB_WORKSPACE}/ca.crt"
```
nsingla: same case here

.github/workflows/legacy-v2-api-integration-tests.yml
```diff
@@ -91,10 +100,7 @@ jobs:
...
         if: ${{ steps.forward-mlmd-port.outcome == 'success' }}
         working-directory: ./backend/test/v2/integration
-        run: go test -v ./... -args -runIntegrationTests=true -namespace=kubeflow
-        with:
-          pod_to_pod_tls_enabled: ${{ matrix.pod_to_pod_tls_enabled }}
-          ca_cert_path: ${{ steps.configure.outputs.CA_CERT_PATH}}
+        run: go test -v ./... -args -runIntegrationTests=true -namespace=kubeflow -tlsEnabled=${{ steps.configure.outputs.TLS_ENABLED }} -caCertPath=${{ steps.configure.outputs.CA_CERT_PATH }}
         env:
           PULL_NUMBER: ${{ github.event.pull_request.number }}
           PIPELINE_STORE: ${{ matrix.pipeline_store }}
```
nsingla, line 89: https://github.com/kubeflow/pipelines/pull/12357/files#diff-44d2593ed49526a5d420236f65bf910f37cd7eded84b58ba71cc105db6746986R119, needs to include pod2pod tls in the name, otherwise uploading artifact fails due to conflicting name

.github/actions/deploy/action.yml
mprahl: @nsingla it was continuously trying to connect to MLMD but not able to connect over TLS with the 2.14.3 image so it was just causing a lot of TLS errors on the MLMD pod. I figured it was better to do this to make diagnosing test failures easier.

backend/src/apiserver/common/const.go
```diff
@@ -68,6 +68,6 @@ const (
 )
 
 const (
-	TLSCertCAPath = "/etc/pki/tls/cert/ca.crt"
-	CABundleDir   = "/etc/pki/tls/cert"
+	TLSCertCAPath = "/kfp/certs/ca.crt"
```
alyssacgoins, line 71: What is the reasoning behind changing the filepath?

.github/workflows/e2e-test.yml
```diff
@@ -132,8 +131,9 @@ jobs:
             NAMESPACE=${{ inputs.namespace }}
           fi
           
+          CA_CERT_PATH=""
```
alyssacgoins: `CA_CERT_PATH`: "" is declared under `env` on line 9. What's the reasoning behind using both `CA_CERT_PATH` and `CA_CERT_FILE`?

.github/workflows/legacy-v2-api-integration-tests.yml
```diff
@@ -110,5 +103,5 @@ jobs:
         if: always()
         uses: actions/upload-artifact@v4
         with:
-          name: kfp-api-integration-tests-v2-artifacts-k8s-${{ matrix.k8s_version }}-${{ matrix.pipeline_store }}
+          name: kfp-api-integration-tests-v2-artifacts-k8s-${{ matrix.k8s_version }}-${{ matrix.pipeline_store }}-tls-${{ matrix.pod_to_pod_tls_enabled }}
           path: /tmp/tmp*/*
```
alyssacgoins, line 103: EOF

.github/workflows/e2e-test.yml
mprahl: @alyssacgoins I think this is addressed in the last push.

backend/src/apiserver/common/const.go
mprahl: It was mostly because the volume mount could shadow the real /etc/pki/tls and AI kept getting confused that it was `/etc/pki/tls/cert` and not `/etc/pki/tls/certs` lol.

.github/workflows/legacy-v2-api-integration-tests.yml
mprahl: It seems to be an existing issue and a problem in other files. I suggest another PR to fix those. :smile:

.github/workflows/api-server-tests.yml
```diff
@@ -149,7 +148,7 @@ jobs:
           python_version: ${{ env.PYTHON_VERSION }}
           report_name: "Standalone_k8sVersion=${{ matrix.k8s_version }}_argoVersion=${{ matrix.argo_version }}_cacheEnabled=${{ matrix.cache_enabled }}_proxyEnabled=${{ matrix.proxy }}"
           tls_enabled: ${{ matrix.pod_to_pod_tls_enabled }}
-          ca_cert_path: ${{ steps.configure-cluster-ca-cert.outputs.CA_CERT_PATH}}
+          ca_cert_path: ${{ steps.configure-cluster-ca-cert.outputs.CA_CERT_PATH }}
```
nsingla: :thinking: i don't think this is a valid output anymore, CA_CERT_PATH is an env variable, so it should be accessible directly

.github/workflows/api-server-tests.yml
```diff
@@ -102,9 +102,13 @@ jobs:
       - name: Configure Cluster CA Cert for TLS-Enabled Tests
         shell: bash
         id: configure-cluster-ca-cert
-        if: ${{ matrix.pod_to_pod_tls_enabled == "true"}}
+        if: ${{ matrix.pod_to_pod_tls_enabled == 'true'}}
+        env:
+          CA_CERT_FILE: "${{ github.workspace }}/ca.crt"
```
nsingla: CA_CERT_FILE is already an env variable, i think you can just set the value here, rather than making it an env variable again

.github/workflows/e2e-test.yml
```diff
@@ -112,12 +112,13 @@ jobs:
       - name: Configure Cluster CA Cert for TLS-Enabled Tests
         shell: bash
         id: configure-cluster-ca-cert
-        if: ${{ matrix.pod_to_pod_tls_enabled == "true"}}
+        if: ${{ matrix.pod_to_pod_tls_enabled == 'true'}}
         env:
-          SCHEME: "https"
+          CA_CERT_FILE: "${{ github.workspace }}/ca.crt"
```
nsingla: same case here

.github/workflows/legacy-v2-api-integration-tests.yml
```diff
@@ -66,20 +70,12 @@ jobs:
         shell: bash
         id: configure-cluster-ca-cert
         if: ${{ matrix.pod_to_pod_tls_enabled }}
+        env:
+          CA_CERT_FILE: "${{ github.workspace }}/ca.crt"
```
nsingla: same

.github/workflows/api-server-tests.yml
mprahl: It pulls it from the `configure-cluster-ca-cert` so it's valid output. It's just the way it was before but I added the missing space.

.github/workflows/api-server-tests.yml
mprahl: Yup, I misunderstood your previous comments. Let me know if this is better now.

PR #12368, opened 2025-10-15
feat(sdk): auto-detect and install Kubeflow SDK for component functions #12027

### LINKED ISSUE #12027: [feature] Include Kubeflow SDK in KFP SDK

Prateekbala:
## Description
This PR implements automatic Kubeflow SDK integration for KFP components as requested in #12027. It adds the ability to automatically detect and install the Kubeflow SDK when components use kubeflow imports, while also providing explicit extras for opt-in installation.

### 1. Added Kubeflow Extras to setup.py
- Added `kubeflow = ['kubeflow']` extras option
- Included kubeflow in the `'all'` extras bundle
- Allows users to install with `pip install kfp[kubeflow]` or `pip install kfp[all]`

### 2. Implemented AST-based Auto-detection
- Added `_detect_kubeflow_imports_in_function()` in `component_factory.py`
- Uses Abstract Syntax Tree parsing (`ast` + `inspect` + `textwrap.dedent`) to detect kubeflow imports in component functions
- Supports multiple import patterns:
  - `import kubeflow`
  - `import kubeflow.<submodule>`
  - `from kubeflow import <symbol>`

### 3. Automatic Package Installation
- Modified `_get_packages_to_install_command(...)` to auto-detect kubeflow usage
- Automatically adds `'kubeflow'` to `packages_to_install` when detected and not already specified
- Respects explicit user-provided packages and version pins (no duplication)
- Recognizes kubeflow when specified via VCS URLs
- Fails closed: if source cannot be inspected or parsed, no auto-add occurs

### 4. Opt-out Control per Component
- Extended `@dsl.component` with `install_kubeflow_package: bool = True`
  - `True` (default): auto-add kubeflow if user code imports it
  - `False`: never auto-add kubeflow for that component

### 5. Comprehensive Test Coverage
- Added test cases covering:
  - All supported kubeflow import patterns and negative cases
  - Behavior when source inspection fails or syntax is invalid
  - Package parsing (versions, extras, VCS URLs) and duplicate avoidance
  - Decorator integration ensuring kubeflow is only installed when needed

### Install via extras:
```bash
pip install kfp[kubeflow]
pip install kfp[all]
```

### Default auto-detection (no user change needed):
```python
@dsl.component
def my_comp(...):
    import kubeflow
    ...
# Kubeflow SDK is added to packages_to_install automatically
```

### Opt-out:
```python
@dsl.component(install_kubeflow_package=False)
def my_comp(...):
    import kubeflow
    ...
# Kubeflow SDK will not be auto-added
```
## Related Issues
Fixes #12027


mprahl (commented):
/ok-to-test

mprahl:
@Prateekbala could you please add a sign off to your commit?

Prateekbala:
> @Prateekbala could you please add a sign off to your commit?
I’ve added the sign-off to my commit. Thank you for pointing that out.

sdk/python/kfp/dsl/component_decorator.py
```diff
@@ -30,6 +30,7 @@ def component(
     pip_index_urls: Optional[List[str]] = None,
     output_component_file: Optional[str] = None,
     install_kfp_package: bool = True,
+    install_kubeflow_package: bool = True,
```
mprahl: Could we make this an enum value instead? I'm thinking of the three options or similar:
1. auto
2. install
3. skip

Then we can default to `auto` but allow the user to override (force an installation) in cases where `auto` doesn't work. I suppose the user could always force it with adding a value to `packages_to_install` but this makes the behavior clearer.

sdk/python/kfp/dsl/component_factory.py
```diff
@@ -152,15 +244,32 @@ def make_pip_install_command(
...
     packages_to_install = packages_to_install or []
+
+    # Auto-detect and add kubeflow if needed
+    if install_kubeflow_package and func is not None:
+        detected_kubeflow = _detect_kubeflow_imports_in_function(func)
+
+        if detected_kubeflow:
+            # Parse existing packages to check for kubeflow
+            existing_package_names = [
+                _parse_package_name(pkg) for pkg in packages_to_install
```
mprahl: My preference is to use a regex to specifically detect the presence of kubeflow rather than having generic package name detection.

sdk/python/kfp/dsl/component_factory_test.py
```diff
@@ -338,6 +339,272 @@ def comp(
         self.assertFalse(comp.component_spec.inputs['i'].is_artifact_list)
 
 
+class TestKubeflowAutoDetection(unittest.TestCase):
+    """Tests for AST-based kubeflow import detection and auto-installation."""
+
+    def test_detect_kubeflow_imports_simple_import(self):
+        """Test detection of 'import kubeflow'."""
+
+        def component_with_kubeflow_import():
```
mprahl: These test components don't seem to actually import kubeflow.

Prateekbala:
I've implemented the requested changes

mprahl:
Please rebase so that there aren't any merge commits.

Prateekbala:
> Please rebase so that there aren't any merge commits.

Done

chensun:
>  It adds the ability to automatically detect and install the Kubeflow SDK when components use kubeflow imports, while also providing explicit extras for opt-in installation.

@Prateekbala @mprahl 
I generally dislike over-smart behaviors like this. Kubeflow imports inside users' component code are no different from any other library imports, and it should be users' responsibilities specifying the dependencies they want to install. 

We should stick to standard Python dependency management. Inferring dependencies from imports is problematic and rarely the right approach.

In practice, as the sdk evolves, there could be breaking changes, the version you install may not actually work with the user code, and there could be transient dependencies incompatible with user-specified dependencies--one of the reason we opted to install kfp with no dependencies.

Prateekbala:
> > It adds the ability to automatically detect and install the Kubeflow SDK when components use kubeflow imports, while also providing explicit extras for opt-in installation.
> 
> @Prateekbala @mprahl I generally dislike over-smart behaviors like this. Kubeflow imports inside users' component code are no different from any other library imports, and it should be users' responsibilities specifying the dependencies they want to install.

Hi @chensun, I understand the concern around inferring dependencies. In the earlier discussion on #12027 and in the Kubeflow SDK design thread, @mprahl proposed using an AST-based check to detect kubeflow imports by providing a kubeflow extras option. The direction in that conversation was to implement this AST-based detection together with the extras entry as an initial solution, with the expectation that it could be revised when the broader Kubeflow SDK integration work evolves.

andreyvelich (commented):
Thanks for this great contribution @Prateekbala!
/assign @kubeflow/kubeflow-sdk-team @franciscojavierarceo @Fiona-Waters @abhijeet-dhumal it would be great if you could help with review to include Kubeflow SDK into KFP!

sdk/python/kfp/dsl/component_decorator.py
```diff
@@ -21,6 +22,19 @@
...
+    Attributes:
+        AUTO: Automatically detect and install if kubeflow is imported in the component function.
+        INSTALL: Always install the kubeflow package regardless of usage.
+        SKIP: Never install the kubeflow package, even if detected in the component function.
+    """
+    AUTO = 'auto'
+    INSTALL = 'install'
+    SKIP = 'skip'
+
+
```
kramaranya, line 37: Why do we duplicate this definition from https://github.com/Prateekbala/pipelines/blob/feature/kubeflow-sdk-integrate/sdk/python/kfp/dsl/component_factory.py#L52-L62? We can just import it

sdk/python/kfp/dsl/component_decorator.py
```diff
@@ -30,6 +44,8 @@ def component(
     pip_index_urls: Optional[List[str]] = None,
     output_component_file: Optional[str] = None,
     install_kfp_package: bool = True,
+    install_kubeflow_package:
+    KubeflowPackageInstallMode = KubeflowPackageInstallMode.AUTO,
```
kramaranya, line 48:
```suggestion
    install_kubeflow_package: KubeflowPackageInstallMode = KubeflowPackageInstallMode.AUTO,
```

sdk/python/kfp/dsl/component_factory.py
```diff
@@ -152,15 +230,36 @@ def make_pip_install_command(
...
+    install_kubeflow_package:
+    KubeflowPackageInstallMode = KubeflowPackageInstallMode.AUTO,
     target_image: Optional[str] = None,
     pip_trusted_hosts: Optional[List[str]] = None,
     use_venv: bool = False,
 ) -> List[str]:
     packages_to_install = packages_to_install or []
+
+    # Handle kubeflow installation based on the mode
+    if install_kubeflow_package == KubeflowPackageInstallMode.INSTALL:
```
kramaranya, line 247: This won't work if user uses KubeflowPackageInstallMode from other definition, we should keep enum in one place only so it's always the same type

sdk/python/kfp/dsl/component_factory.py
```diff
@@ -47,6 +49,82 @@
...
+
+    Args:
+        func: The function to analyze for kubeflow imports.
+
+    Returns:
+        bool: True if any kubeflow import is found, False otherwise.
+
+    Detects these import patterns:
+        - import kubeflow
+        - import kubeflow.training (any submodule)
```
kramaranya, line 76: There is no training submodule in sdk https://github.com/kubeflow/sdk/tree/main/kubeflow/trainer
```suggestion
        - import kubeflow.trainer (any submodule)
```

sdk/python/kfp/dsl/component_decorator.py
```diff
@@ -88,6 +104,13 @@ def component(
             This flag is ignored when ``target_image`` is specified, which implies
             a choice to build a containerized component. Containerized components
             will always install KFP as part of the build process.
+        install_kubeflow_package: Controls whether the Kubeflow SDK is installed.
+            Can be KubeflowPackageInstallMode.AUTO (default), KubeflowPackageInstallMode.INSTALL,
+            or KubeflowPackageInstallMode.SKIP.
+            - AUTO: Detects kubeflow imports in the component function via AST parsing
+              and automatically adds 'kubeflow' to packages_to_install if detected.
+            - INSTALL: Always installs the kubeflow package regardless of whether it's used.
+            - SKIP: Never installs kubeflow, even if detected (useful if pre-installed in base_image).
```
kramaranya, line 113: `install_kubeflow_package` reads like a boolean rather than a mode. 
Would `kubeflow_sdk_install_mode` be clearer? cc @kubeflow/kubeflow-sdk-team @kubeflow/wg-pipeline-leads

kramaranya (commented):
Thank you @Prateekbala! 
I left a few initial comments

kramaranya:
I'm wondering if we should revisit this given @chensun's concerns? Any other concerns from @kubeflow/wg-pipeline-leads about this approach or are we happy to move forward?

Fiona-Waters (commented):
I think that there could be some complications with this implementation around version clashes etc. If we do go ahead we would need clear user documentation around the install mode.

mprahl:
> > It adds the ability to automatically detect and install the Kubeflow SDK when components use kubeflow imports, while also providing explicit extras for opt-in installation.
> 
> @Prateekbala @mprahl I generally dislike over-smart behaviors like this. Kubeflow imports inside users' component code are no different from any other library imports, and it should be users' responsibilities specifying the dependencies they want to install.
> 
> We should stick to standard Python dependency management. Inferring dependencies from imports is problematic and rarely the right approach.
> 
> In practice, as the sdk evolves, there could be breaking changes, the version you install may not actually work with the user code, and there could be transient dependencies incompatible with user-specified dependencies--one of the reason we opted to install kfp with no dependencies.

@chensun Thanks for the review. You raise valid points regarding the risks of "magic" behavior and dependency management. I definitely agree that we want to avoid inferring dependencies for general libraries.

However, I believe the Kubeflow SDK warrants a special exception separate from standard third-party packages. Our goal is to encourage seamless usage of the wider Kubeflow ecosystem within components. Conceptually, this somewhat aligns with how we already pre-install the `kfp` SDK into the container at runtime; we are simply extending that convenience to the broader Kubeflow namespace.

To address your concerns about version drift and user control, I propose the following safeguards for this PR:

* Respecting Opt-Outs: If the user explicitly disables `kfp` package installation in the component decorator, we should also disable the automatic detection/installation of the Kubeflow SDK.
* Strict Version Pinning: We could pin the Kubeflow SDK version to an x.y version in setup of the KFP SDK. This ensures the injected SDK is always aligned with the runtime environment, mitigating the risk of breaking changes or incompatibility.
* Best-Effort Detection: We could keep the autodetect logic for kubeflow imports, but it will only trigger the installation of the pinned, compatible version defined above.
* CI Verification: Let's add a pipeline run in CI to explicitly test this autodetection and installation path, ensuring the pinned version installs successfully without conflict.
* Visibility: We should log in the executor that Kubeflow SDK usage was detected and it will be automatically installed. We can also provide instructions to disable it.

Prateekbala:
> To address your concerns about version drift and user control, I propose the following safeguards for this PR:
> 
> * Respecting Opt-Outs: If the user explicitly disables `kfp` package installation in the component decorator, we should also disable the automatic detection/installation of the Kubeflow SDK.
> * Strict Version Pinning: We could pin the Kubeflow SDK version to an x.y version in setup of the KFP SDK. This ensures the injected SDK is always aligned with the runtime environment, mitigating the risk of breaking changes or incompatibility.
> * Best-Effort Detection: We could keep the autodetect logic for kubeflow imports, but it will only trigger the installation of the pinned, compatible version defined above.
> * CI Verification: Let's add a pipeline run in CI to explicitly test this autodetection and installation path, ensuring the pinned version installs successfully without conflict.
> * Visibility: We should log in the executor that Kubeflow SDK usage was detected and it will be automatically installed. We can also provide instructions to disable it.

Hey @mprahl , I've addressed the review feedback and added some safeguards based on the concerns. A few things I need clarification on:

1. Kubeflow SDK version - What version should we actually pin to is it 1.9.0 ? Should it match a specific KFP version or use a range like >=1.9.0,<2.0?

2. Logging - Where should we log the auto-detection? At compile time when building the spec, or at runtime in the executor? 

3. Default mode - Currently AUTO is the default. Given the concerns , should we start with SKIP as default instead and let users opt-in?

mprahl:
> > To address your concerns about version drift and user control, I propose the following safeguards for this PR:
> > 
> > * Respecting Opt-Outs: If the user explicitly disables `kfp` package installation in the component decorator, we should also disable the automatic detection/installation of the Kubeflow SDK.
> > * Strict Version Pinning: We could pin the Kubeflow SDK version to an x.y version in setup of the KFP SDK. This ensures the injected SDK is always aligned with the runtime environment, mitigating the risk of breaking changes or incompatibility.
> > * Best-Effort Detection: We could keep the autodetect logic for kubeflow imports, but it will only trigger the installation of the pinned, compatible version defined above.
> > * CI Verification: Let's add a pipeline run in CI to explicitly test this autodetection and installation path, ensuring the pinned version installs successfully without conflict.
> > * Visibility: We should log in the executor that Kubeflow SDK usage was detected and it will be automatically installed. We can also provide instructions to disable it.
> 
> Hey @mprahl , I've addressed the review feedback and added some safeguards based on the concerns. A few things I need clarification on:
> 
> 1. Kubeflow SDK version - What version should we actually pin to is it 1.9.0 ? Should it match a specific KFP version or use a range like >=1.9.0,<2.0?

I think we can keep the range `>=0.2.0,<0.3.0` for now and bump it every KFP release. @andreyvelich @kramaranya do you agree with that?

> 2. Logging - Where should we log the auto-detection? At compile time when building the spec, or at runtime in the executor?

I was thinking the executor so the user can have debug information about which Kubeflow SDK was used at runtime.

> 3. Default mode - Currently AUTO is the default. Given the concerns , should we start with SKIP as default instead and let users opt-in?

I think if the `install_kfp_package` argument is `True` default to auto. Otherwise let the user explicitly specify the behavior.

kramaranya:
> I think we can keep the range >=0.2.0,<0.3.0 for now and bump it every KFP release. @andreyvelich @kramaranya do you agree with that?

I think we can just keep >=0.2.0 and it should install the latest version, so we don't need to bump it every time

andreyvelich:
>I think we can just keep >=0.2.0 and it should install the latest version, so we don't need to bump it every time

@kramaranya If we are planning to introduce breaking changes between minor releases, it might be better to stitch with `<0.3` for now to not break users' pipelines after upgrade.

sdk/python/kfp/dsl/component_decorator.py
andreyvelich: Shall we just follow the same signature as `install_kfp_package` ?
If the default value is AUTO, why would user change it to INSTALL ?

mprahl:
Thank you very much for your efforts @Prateekbala!

We discussed this PR during today's KFP community call. The maintainers have decided to put a pause on this effort due to a versioning concern:

* **The Concern:** The KFP SDK version used to compile a pipeline does not necessarily match the active KFP deployment.
* **The Risk:** If we pin the Kubeflow SDK at the KFP SDK level, we risk installing a Kubeflow SDK that is incompatible with the actual Kubeflow deployment.

**Alternative Proposal**

Instead, we believe this should be handled as an **API Server concern**, not a KFP SDK concern. The system should ideally resolve the associated Kubeflow SDK version (using major.minor/x.y alignment) based on the *deployed* KFP version. This ensures we always use the SDKs compatible with what is actually installed.

Since this architectural change has significant repercussions, we need a member of the KFP community to propose a KEP (Kubeflow Enhancement Proposal) to define this path forward.

Prateekbala:
> Thank you very much for your efforts @Prateekbala!
> 
> We discussed this PR during today's KFP community call. The maintainers have decided to put a pause on this effort due to a versioning concern:

Thanks for the update @mprahl! I’ll pause the PR for now. Happy to help and contribute more.

PR #12369, opened 2025-10-15
fix(backend): Remove the default workspace configuration

mprahl:

This was added as a mistake as this is environment specific and not a safe default.

**Checklist:**

### DIFF
--- a/backend/src/v2/compiler/argocompiler/argo.go
+++ b/backend/src/v2/compiler/argocompiler/argo.go
@@ -542,6 +542,17 @@ func GetWorkspacePVC(
 		}
 	}
 
+	// Access modes define the read/write semantics for the underlying volume and
+	// gate whether multiple pods can mount it concurrently. It is required when creating the
+	// underlying PVC for the workspace.
+	if len(pvcSpec.AccessModes) == 0 {
+		return k8score.PersistentVolumeClaim{}, fmt.Errorf("workspace PVC spec must specify accessModes")
+	}
+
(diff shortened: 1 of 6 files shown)


backend/src/v2/compiler/argocompiler/argo_workspace_test.go
```diff
@@ -221,50 +269,6 @@ func TestGetWorkspacePVC(t *testing.T) {
 			},
 			expectError: false,
 		},
-		{
-			name: "workspace with nil Kubernetes config",
```
alyssacgoins, line 225: What is the reasoning behind completely removing the test cases for workspace with nil K8s config?

backend/src/v2/compiler/argocompiler/argo.go
```diff
@@ -534,6 +534,13 @@ func GetWorkspacePVC(
 		}
 	}
 
+	if len(pvcSpec.AccessModes) == 0 {
```
alyssacgoins, line 548: It would be helpful to have a comment here explaining pvcSpec.AccessModes

backend/src/v2/compiler/argocompiler/argo_workspace_test.go
mprahl: It's because the test case `workspace with size specified and default workspace` now covers this scenario.

gmfrasca (commented):
/lgtm

alyssacgoins (approved):

PR #12383, opened 2025-10-21
chore: update py image refs to 3.11

zazulam:
Modified all python image references to now point to python 3.11. 

As discussed offline, we will keep 3.9 as the lowest working python installation for KFP and keep it the lower end of the python testing matrix since some users may be slow to upgrade their setups.

Updated the warning to say:
`KFP will drop support for Python 3.9 on October 1, 2026. To use new versions of the KFP SDK after that date, you will need to upgrade to Python >= 3.10. See https://devguide.python.org/versions/ for more details.`

Pushed it out a year, I'm open to updating the date to whatever others think would be best.

**Checklist:**


droctothorpe:
Thanks for tackling this, @zazulam ! LGTM. Will approve once rebased and tests are passing! 👍

nsingla:
Should we have a couple of test pipeline IR with older python version until we completely migrate to 3.11?

zazulam:
> Should we have a couple of test pipeline IR with older python version until we completely migrate to 3.11?

@nsingla My thought is having the test matrix for the `sdk_tests` set for py3.9 & py3.13 should cover what we want for ensuring the sdk works for those versions and assuming the versions in between. As for the images that the components run on, I think it's safe to change to py3.11. There is 1 test where I commented out the image assert statement because it checks against a [yaml from master](https://github.com/kubeflow/pipelines/pull/12383/files#diff-8ce61eb744c413f9e3374febaea7a9b56f6b244ffadea1e0b3a6763f651b9900R122-R125).

.github/actions/kfp-k8s/action.yml
```diff
@@ -45,3 +45,12 @@ runs:
...
         pip install -e ./kubernetes_platform/python[dev] --find-links=sdk/python/dist
+
+    # testing reinstalling kfp package from source with no deps
+    - name: Reinstall kfp from source with no deps
+      id: reinstall-kfp
+      shell: bash
+      if: ${{ steps.install-kfp-kubernetes.outcome == 'success' }}
+      working-directory: ./sdk/python
+      run: |
+        pip install . -I --no-deps
```
zazulam, line 56: Added this since the previous section was not actually using the source install for the kfp sdk, see [here](https://github.com/kubeflow/pipelines/actions/runs/18757420269/job/53512878485?pr=12383) for the error prior to this add.

sdk/python/kfp/components/load_yaml_utilities_test.py
```diff
@@ -119,9 +119,10 @@ def test_load_component_from_url(self):
                          {'Output': structures.OutputSpec(type='String')})
         self.assertEqual(component._component_inputs, {'value'})
         self.assertEqual(component.name, 'identity')
-        self.assertEqual(
-            component.component_spec.implementation.container.image,
-            'python:3.9')
+        # TODO: uncomment once PR #12383 is merged since this is checking against a version on master
+        # self.assertEqual(
+        #     component.component_spec.implementation.container.image,
+        #     'python:3.11')
```
zazulam, line 125: commenting for reference for others

zazulam (commented):
Since updating the sdk to now have the base image set to py3.11, all compiled workflows needed to be regenerated, hence the magnitude of tiny changes. 

I also had to regen some of the proto files for the backend proto_tests.

PR #12394, opened 2025-10-27
Implemented HTTP Artifact Streaming to Prevent OOM Errors with Large Files

hbelmiro:
This PR implements HTTP streaming for KFP artifacts, preventing out-of-memory (OOM) errors when downloading large files through the API server. The implementation streams artifacts directly from blob storage to the HTTP response without loading them into memory.

### 1. Migrated to Cloud-Native Blob Storage (`gocloud.dev/blob`)

- Replaced MinIO-specific implementation with provider-agnostic `gocloud.dev/blob`
- Supports S3, GCS, Azure Blob, and MinIO through a unified interface
- Enables true streaming with `io.Copy` pattern

### 2. Implemented Streaming Throughout the Stack

- **Storage Layer**: `BlobObjectStore.GetFileReader()` returns an `io.ReadCloser` for streaming
- **Resource Manager**: Uses `io.Copy(httpWriter, reader)` to stream in chunks
- **HTTP Endpoint**: Streams response without buffering entire content

### 3. Refactored Client Management

- Consolidated blob storage initialization in `client_manager.go`
- Improved credential handling for both environment variables and Kubernetes secrets
- Added support for MinIO path-style URLs with `use_path_style=true`
- Reused existing Kubernetes client to avoid redundant connections

### 4. Improved Code Organization

- Extracted `TokenRefresher` into dedicated package
- Introduced `artifact.Client` for better abstraction
- Removed deprecated `ReadArtifact` functions
- Cleaned up redundant MinIO implementation

### Core Implementation

- `backend/src/apiserver/storage/blob_object_store.go`: New streaming-capable storage implementation
- `backend/src/apiserver/resource/resource_manager.go`: Updated `StreamArtifact` to use streaming reader
- `backend/src/apiserver/server/run_artifact_server.go`: HTTP endpoint now streams responses
- `backend/src/apiserver/client_manager/client_manager.go`: Refactored blob storage initialization

### Testing

- **Added streaming tests**:
  - `TestBlobObjectStore_GetFileReader_ChunkedStreaming`: Validates chunked reading mechanics
  - `TestBlobObjectStore_Streaming_1GB_File`: Proves 1GB files stream without OOM
  - `TestStreamArtifactV1_ChunkedResponse`: Validates HTTP endpoint streams in chunks

### Run Unit Tests

```bash
# Storage layer streaming tests
go test -v -run "TestBlobObjectStore.*Streaming" ./backend/src/apiserver/storage/

# HTTP endpoint streaming test  
go test -v -run TestStreamArtifactV1_ChunkedResponse ./backend/src/apiserver/server/
```

**Checklist:**


backend/src/apiserver/server/run_artifact_server.go
```diff
@@ -0,0 +1,148 @@
...
+)
+
+type RunArtifactServer struct {
+	resourceManager *resource.ResourceManager
+}
+
+// StreamArtifactV1 is an artifact streaming endpoint that streams artifacts directly from object storage
+// to the HTTP response without buffering the entire content in memory.
+// No size limits are imposed - the streaming approach itself provides the security benefit.
+func (s *RunArtifactServer) StreamArtifactV1(response http.ResponseWriter, r *http.Request) {
```
alyssacgoins: It could be helpful here to return an error (and nil if no error) - that way it's clear an error has occurred, wherever this function is called

backend/src/apiserver/client_manager/client_manager.go
```diff
@@ -929,23 +947,193 @@ func addDisplayNameColumn(db *gorm.DB, mdl interface{}, dialect SQLDialect) erro
...
+	bucket, err := openBucketWithRetry(ctx, config, blobConfig.useDirectBucket, initConnectionTimeout, k8sClient)
+	if err != nil {
+		return nil, fmt.Errorf("failed to open blob storage bucket: %w", err)
+	}
+
+	// For MinIO, ensure the bucket exists (create if it doesn't)
+	// This is needed because MinIO doesn't auto-create buckets like SeaweedFS does
+	if bucketName != "" && config.SessionInfo != nil {
+		switch config.SessionInfo.Provider {
+		case "minio", "s3":
```
alyssacgoins: Since there is only one switch case here, the logic reads a little confusingly

backend/src/apiserver/client_manager/client_manager.go
```diff
@@ -929,23 +947,193 @@ func addDisplayNameColumn(db *gorm.DB, mdl interface{}, dialect SQLDialect) erro
...
+	const minioArtifactSecretKeyKey = "secretkey"
+
+	// Set AWS environment variables
+	if accessKey != "" {
+		os.Setenv("AWS_ACCESS_KEY_ID", accessKey)
+	}
+	if secretKey != "" {
+		os.Setenv("AWS_SECRET_ACCESS_KEY", secretKey)
+	}
+	// Set region - use default for MinIO if not specified
```
alyssacgoins: As someone who has much less context on MinIO - it would be helpful to have a little context on a default MinIO region - which region it is, why, etc.

backend/src/apiserver/server/run_artifact_server_test.go
```diff
@@ -0,0 +1,362 @@
...
+
+	runArtifactServer := NewRunArtifactServer(resourceManager)
+
+	testCases := []struct {
+		name        string
+		vars        map[string]string
+		expectedMsg string
+	}{
+		{
+			name:        "Missing run_id",
```
alyssacgoins: Would it be worth including a case in which all three parameters are missing?

backend/src/agent/persistence/client/artifactclient/client.go
```diff
@@ -0,0 +1,187 @@
...
+}
+
+// ErrorCode represents different types of errors
+type ErrorCode int
+
+const (
+	// ErrorCodePermanent indicates errors that should not be retried
+	ErrorCodePermanent ErrorCode = iota
+	// ErrorCodeTransient indicates errors that may be retried
+	ErrorCodeTransient
```
alyssacgoins, line 55: Does ErrorCodeTransient have a type, or is this done purposefully?

backend/src/agent/persistence/client/tokenrefresher/token_refresher.go
```diff
@@ -8,18 +10,13 @@ import (
...
-}
-
 const SaTokenFile = "/var/run/secrets/kubeflow/tokens/persistenceagent-sa-token"
 
 type FileReader interface {
 	ReadFile(filename string) ([]byte, error)
 }
 
-type tokenRefresher struct {
+type TokenRefresher struct {
```
alyssacgoins, line 19: Is this capitalization purely a linting fix, or is it a more structural change?

backend/src/apiserver/server/run_artifact_server.go
hbelmiro: This is an HTTP endpoint and it's used as an argument for `func (r *Router) HandleFunc(path string, f func(http.ResponseWriter, *http.Request)) *Route` [here](https://github.com/hbelmiro/data-science-pipelines/blob/41597e0239572aad1ba9628f4f4387446c12afa0/backend/src/apiserver/main.go#L373) . So, besides having to match the expected signature, it's not supposed to be directly called from Go code and returning an error would be useless.

backend/src/agent/persistence/client/artifactclient/client.go
hbelmiro: Yes, `ErrorCodeTransient` does have a type - it's `ErrorCode`, just like `ErrorCodePermanent`.

In Go, when you use `iota` in a `const` block, both the type and the `iota` expression are automatically carried over to subsequent lines. So this:

```go
const (
    ErrorCodePermanent ErrorCode = iota  // 0
    ErrorCodeTransient                   // 1
)
```

Is equivalent to:

```go
const (
    ErrorCodePermanent ErrorCode = 0
    ErrorCodeTransient ErrorCode = 1
)
```

Both constants have type `ErrorCode`, with values 0 and 1, respectively.

The reason we use `iota` instead of explicit values is:

1. It's the idiomatic Go way to define enumerated types
2. It's less error-prone when adding new constants later (they auto-increment)
3. It signals to readers that these are sequential enumeration values

This is a common Go pattern you'll see throughout the codebase for defining type-safe enums!

backend/src/agent/persistence/client/tokenrefresher/token_refresher.go
hbelmiro: It's a structural change. This is how we change the visibility of a type, variable or method/function from private to public in Go.
Lowercase is private. Uppercase is public.

backend/src/apiserver/client_manager/client_manager.go
hbelmiro: Good catch! Refactored.

backend/src/apiserver/client_manager/client_manager.go
hbelmiro: Added a comment.

backend/src/apiserver/server/run_artifact_server_test.go
hbelmiro: Done. And you actually made me notice that the `expectedMsg` was not being used.

backend/src/apiserver/main.go
```diff
@@ -368,6 +368,11 @@ func startHTTPProxy(resourceManager *resource.ResourceManager, usePipelinesKuber
 	runLogServer := server.NewRunLogServer(resourceManager)
 	topMux.HandleFunc("/apis/v1alpha1/runs/{run_id}/nodes/{node_id}/log", runLogServer.ReadRunLogV1)
 
+	// Artifact streaming endpoints
+	runArtifactServer := server.NewRunArtifactServer(resourceManager)
+	topMux.HandleFunc("/apis/v1beta1/runs/{run_id}/nodes/{node_id}/artifacts/{artifact_name}:stream", runArtifactServer.StreamArtifactV1)
+	topMux.HandleFunc("/apis/v2beta1/runs/{run_id}/nodes/{node_id}/artifacts/{artifact_name}:stream", runArtifactServer.StreamArtifact)
```
HumairAK: we have to keep `:read` available - otherwise this is an api breaking change

backend/src/agent/persistence/client/artifact/client.go
```diff
@@ -0,0 +1,187 @@
...
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+// Package artifact provides client functionality for reading pipeline artifacts
+// using HTTP streaming.
+package artifact
```
HumairAK: nit: can you rename this package to `artifact_client`, `artifact` feels ambiguous

backend/src/apiserver/storage/minio_client.go
HumairAK, line 1: It does not seem like minio_client.go is being used any more, let's remove it

backend/src/apiserver/client_manager/client_manager.go
```diff
@@ -929,23 +945,204 @@ func addDisplayNameColumn(db *gorm.DB, mdl interface{}, dialect SQLDialect) erro
...
+	port := common.GetStringConfigWithDefault("ObjectStoreConfig.Port", "")
+	secure := common.GetBoolConfigWithDefault("ObjectStoreConfig.Secure", false)
+	region := common.GetStringConfigWithDefault("ObjectStoreConfig.Region", "")
+	accessKey := common.GetStringConfigWithDefault("ObjectStoreConfig.AccessKey", "")
+	secretKey := common.GetStringConfigWithDefault("ObjectStoreConfig.SecretAccessKey", "")
+
+	// Constants for MinIO secret (consistent with v2 config)
+	const minioArtifactSecretName = "mlpipeline-minio-artifact"
+	const minioArtifactAccessKeyKey = "accesskey"
+	const minioArtifactSecretKeyKey = "secretkey"
```
HumairAK: these should be at package level

backend/src/apiserver/client_manager/client_manager.go
```diff
@@ -929,23 +945,204 @@ func addDisplayNameColumn(db *gorm.DB, mdl interface{}, dialect SQLDialect) erro
...
+
+	// Set AWS environment variables
+	if accessKey != "" {
+		os.Setenv("AWS_ACCESS_KEY_ID", accessKey)
```
HumairAK: can you handle the errors for all the os.Setenv() calls

backend/src/apiserver/client_manager/client_manager.go
```diff
@@ -929,23 +945,204 @@ func addDisplayNameColumn(db *gorm.DB, mdl interface{}, dialect SQLDialect) erro
...
+	}
+	if secretKey != "" {
+		os.Setenv("AWS_SECRET_ACCESS_KEY", secretKey)
+	}
+	// Set region - use default for MinIO if not specified
+	isMinIO := host != ""
```
HumairAK: minio is no longer the default and the host needs to always be provided, in fact let's run validation on these fields: BucketName / Host / AccessKey / SecretAccessKey, these are always required.

backend/src/apiserver/client_manager/client_manager.go
```diff
@@ -929,23 +945,204 @@ func addDisplayNameColumn(db *gorm.DB, mdl interface{}, dialect SQLDialect) erro
...
+	params := map[string]string{}
+
+	// Configure credential handling based on endpoint type.
+	// MinIO requires special handling as fromEnv=true causes fallback to AWS endpoints.
+	queryString := ""
+	hasCredentials := accessKey != "" && secretKey != ""
+	useDirectBucket := false
+
+	// Set base parameters
+	if isMinIO {
```
HumairAK: only this case is allowed, let's get rid of this if statement, and if we are missing these configuration options from, throw an error 

so `useDirectBucket` is always true, and we can remove it

backend/src/apiserver/client_manager/client_manager.go
```diff
@@ -929,23 +945,204 @@ func addDisplayNameColumn(db *gorm.DB, mdl interface{}, dialect SQLDialect) erro
...
+	namespace := common.GetPodNamespace()
+
+	operation := func() error {
+		if useDirectBucket {
+			// MinIO with environment credentials - use direct blob.OpenBucket
+			bucketURL := config.Scheme + config.BucketName
+			if config.QueryString != "" {
+				bucketURL += "?" + config.QueryString
+			}
+			bucket, err = blob.OpenBucket(ctx, bucketURL)
```
HumairAK: always use this case, let's avoid using `v2/objectstore`, that's intended for launcher to use and it's pretty convoluted, let's keep api server simple here

backend/src/apiserver/client_manager/client_manager.go
```diff
@@ -929,23 +945,204 @@ func addDisplayNameColumn(db *gorm.DB, mdl interface{}, dialect SQLDialect) erro
...
+		Params:   params,
+	}
 
-	return storage.NewMinioObjectStore(&storage.MinioClient{Client: minioClient}, bucketName, pipelinePath, disableMultipart)
+	return &blobStorageConfig{
+		config: &objectstore.Config{
+			Scheme:      "s3://",
+			BucketName:  bucketName,
+			QueryString: queryString,
+			SessionInfo: sessionInfo,
```
HumairAK: SessionInfo is suppose to be used for parsing the kfp-launcher, which is intended for the launcher, I don't think we should mix concerns here by using it in the api server. Let's remove this and just provide fields if needed straight to `blobStorageConfig`, though as mentioned elsewhere, we shouldn't really be using v2/objectstore at all in apiserver so it's usefulness is further diminished here.

backend/src/apiserver/client_manager/client_manager.go
```diff
@@ -929,23 +945,204 @@ func addDisplayNameColumn(db *gorm.DB, mdl interface{}, dialect SQLDialect) erro
...
+	bucket, err := openBucketWithRetry(ctx, config, blobConfig.useDirectBucket, initConnectionTimeout, k8sClient)
+	if err != nil {
+		return nil, fmt.Errorf("failed to open blob storage bucket: %w", err)
+	}
+
+	if shouldEnsureObjectBucket(bucketName, config) {
+		if err := ensureMinioBucketExists(ctx, config, bucketName, blobConfig.accessKey, blobConfig.secretKey, k8sClient); err != nil {
+			glog.Warningf("Failed to ensure MinIO bucket exists (may already exist): %v", err)
+		}
+	}
```
HumairAK: most gocloud blob drivers (s3, gcs, azure) will throw an error at `blob.OpenBucket()` stage, so I think this check is unnecessary

backend/src/apiserver/server/run_artifact_server.go
```diff
@@ -0,0 +1,148 @@
...
+			glog.Warningf("Failed to close artifact reader: %v", closeErr)
+		}
+	}()
+	return true, nil
+}
+
+// StreamArtifact handles v2beta1 artifact streaming (same implementation as v1)
+func (s *RunArtifactServer) StreamArtifact(w http.ResponseWriter, r *http.Request) {
+	glog.Infof("Stream artifact v2 called")
+	s.StreamArtifactV1(w, r)
```
HumairAK: nit: it should be the other way around, `StreamArtifactV1` should use the V2 code, since we will remove v1 api endpoints soon

backend/src/apiserver/main.go
hbelmiro: Done. Renamed `:stream` to `: read`.

backend/src/agent/persistence/client/artifact/client.go
hbelmiro: Done.

backend/src/apiserver/storage/minio_client.go
hbelmiro: Done.

backend/src/apiserver/client_manager/client_manager.go
hbelmiro: Done.

backend/src/apiserver/client_manager/client_manager.go
hbelmiro: Done.

backend/src/apiserver/client_manager/client_manager.go
hbelmiro: Done.

backend/src/apiserver/client_manager/client_manager.go
hbelmiro: Done.

backend/src/apiserver/client_manager/client_manager.go
hbelmiro: Done. 

Also, created this issue:

- https://github.com/kubeflow/pipelines/issues/12439.

backend/src/apiserver/client_manager/client_manager.go
hbelmiro: Done.

backend/src/apiserver/client_manager/client_manager.go
hbelmiro: Removed.

backend/src/apiserver/server/run_artifact_server.go
hbelmiro: Done.

backend/src/apiserver/client_manager/client_manager.go
HumairAK: ensureBucketExists still exists, I think we can remove this, again OpenBucket should hit an error if the bucket exists on most common drivers

backend/src/apiserver/client_manager/client_manager.go
```diff
@@ -929,23 +942,137 @@ func addDisplayNameColumn(db *gorm.DB, mdl interface{}, dialect SQLDialect) erro
...
 
-	return storage.NewMinioObjectStore(&storage.MinioClient{Client: minioClient}, bucketName, pipelinePath, disableMultipart)
+	// Set AWS environment variables
+	if err := os.Setenv("AWS_ACCESS_KEY_ID", accessKey); err != nil {
+		return nil, fmt.Errorf("failed to set AWS_ACCESS_KEY_ID: %w", err)
+	}
+	if err := os.Setenv("AWS_SECRET_ACCESS_KEY", secretKey); err != nil {
+		return nil, fmt.Errorf("failed to set AWS_SECRET_ACCESS_KEY: %w", err)
+	}
```
HumairAK: This is a bit too passive, can we use a more declarative approach instead of passing credentials in the environment and passively consuming them? 

For example, you can do something like the following (AI generated, untested): 

```golang
// Create AWS config with explicit credentials (no environment variables)
		cfg, err := awsv2cfg.LoadDefaultConfig(ctx,
			awsv2cfg.WithRegion(config.region),
			awsv2cfg.WithCredentialsProvider(credentials.NewStaticCredentialsProvider(
				config.accessKey,
				config.secretKey,
				"", // session token (empty for static credentials)
			)),
		)
		if err != nil {
			return fmt.Errorf("failed to create AWS config: %w", err)
		}

		// Create S3 client with custom endpoint
		endpointWithProtocol := ensureProtocol(config.endpoint, config.secure)
		s3Client := s3.NewFromConfig(cfg, func(o *s3.Options) {
			o.BaseEndpoint = awsv2.String(endpointWithProtocol)
			o.UsePathStyle = true // Important for MinIO and other S3-compatible storage
		})

		// Open bucket using s3blob with explicit client
		bucket, err = s3blob.OpenBucketV2(ctx, s3Client, config.bucketName, nil)
		return err

```

we already import the aws sdk in the go mod, so it's available to use.

backend/src/apiserver/main.go
```diff
@@ -368,6 +368,11 @@ func startHTTPProxy(resourceManager *resource.ResourceManager, usePipelinesKuber
 	runLogServer := server.NewRunLogServer(resourceManager)
 	topMux.HandleFunc("/apis/v1alpha1/runs/{run_id}/nodes/{node_id}/log", runLogServer.ReadRunLogV1)
 
+	// Artifact reading endpoints (implemented with streaming for memory efficiency)
+	runArtifactServer := server.NewRunArtifactServer(resourceManager)
+	topMux.HandleFunc("/apis/v1beta1/runs/{run_id}/nodes/{node_id}/artifacts/{artifact_name}:read", runArtifactServer.ReadArtifactV1)
+	topMux.HandleFunc("/apis/v2beta1/runs/{run_id}/nodes/{node_id}/artifacts/{artifact_name}:read", runArtifactServer.ReadArtifact)
```
HumairAK: the original GRPC service had only support for `GET` method, but right now this endpoint looks like it's served on all methods, can you explicitly add only support for GET? 

```golang

	// Artifact reading endpoints (implemented with streaming for memory efficiency)
	runArtifactServer := server.NewRunArtifactServer(resourceManager)
	topMux.HandleFunc("/apis/v1beta1/runs/{run_id}/nodes/{node_id}/artifacts/{artifact_name}:read", runArtifactServer.ReadArtifactV1).Methods("GET")
	topMux.HandleFunc("/apis/v2beta1/runs/{run_id}/nodes/{node_id}/artifacts/{artifact_name}:read", runArtifactServer.ReadArtifact).Methods("GET")
```

HumairAK: can you also add an integration test to ensure this endpoint continues to work as intended, I think a new root context for "Verifying Read Artifact" would be good, this will also be useful in helping us maintain this endpoint for the long term support of 2.5 - wdyt? 

cc @nsingla would this be the right place for it?

backend/src/apiserver/client_manager/client_manager.go
hbelmiro: Oh, what I removed was actually `shouldEnsureObjectBucket`.
But anyway, `blob.OpenBucket()` succeeds when the bucket doesn't exist. The error only appears during the writing.

backend/src/apiserver/main.go
nsingla: I think i need more context, what is this endpoint used for? and where does it get used?
Any idea if UI is using it? 

And our e2e test are at a higher level, so we won;t be able to test the endpoint directly, for direct testing, we will have to add a new test to api tests. If no front end test exist, then we should atleast verify that manually (and if it has impact on Dashboard UI in ODH/RHOAI, then we need to create a ticket for the dashboard team to address with a timeline coinciding with the rebase of upstream to midstream)

hbelmiro: @nsingla the endpoint is not used by the KFP frontend. It's only used by Persistence Agent.
I don't know if it's used by the ODH UI.

But anyways, I've sent another PR with a test that we can use here (I've added the same test here). See https://github.com/kubeflow/pipelines/pull/12445#issuecomment-3534595127.

backend/src/apiserver/server/run_server.go
```diff
@@ -460,27 +460,6 @@ func (s *RunServerV1) ReportRunMetricsV1(ctx context.Context, request *apiv1beta
...
 }
 
-// Reads an artifact.
-// Supports v1beta1 behavior.
-func (s *RunServerV1) ReadArtifactV1(ctx context.Context, request *apiv1beta1.ReadArtifactRequest) (*apiv1beta1.ReadArtifactResponse, error) {
-	if s.options.CollectMetrics {
-		readArtifactRequests.Inc()
-	}
-
-	err := s.canAccessRun(ctx, request.RunId, &authorizationv1.ResourceAttributes{Verb: common.RbacResourceVerbReadArtifact})
```
HumairAK, line 470: can you re-add this check, this line  enforces authorization on this endpoint

backend/src/apiserver/server/run_server_test.go
```diff
@@ -1481,218 +1479,6 @@ func TestCanAccessRun_Unauthenticated(t *testing.T) {
...
-	response, err := runServer.ReadArtifactV1(ctx, artifact)
-	assert.Nil(t, err)
-
-	expectedResponse := &apiv1beta1.ReadArtifactResponse{
-		Data: []byte(expectedContent),
-	}
-	assert.Equal(t, expectedResponse, response)
-}
-
-func TestReadArtifactsV1_Unauthorized(t *testing.T) {
```
HumairAK, line 1550: it looks like this test coverage was not re-added in `run_artifact_server_test.go`

HumairAK: can you also do another look through the tests that were removed here and ensure we continue to test the equivalent functionality (where applicable)

backend/src/apiserver/server/run_server.go
hbelmiro: Good catch! Done.

backend/src/apiserver/server/run_server_test.go
hbelmiro: Done. That was the only one missing.

HumairAK:
/lgtm 
@hbelmiro can you squash your commits and rebase

PR #12377, opened 2025-10-17
Optional Runtime Parameters Fix

sduvvuri1603:
**Description of changes**
- Treat `ErrResolvedParameterNull` as skip for optional Kubernetes inputs and ensure the driver’s pod spec patch includes certs.
- Add `missing_kubernetes_optional_inputs` pipeline and regenerated compiler goldens covering secrets/configmaps, node selector, node affinity, and tolerations missing-input scenarios.

**Testing**
- Manual: ran node selector, affinity, secrets/configmaps pipelines in Kind (no launcher crashes).
- `ginkgo -v ./backend/test/compiler -- -createGoldenFiles=true -updateCompiledFiles=true` (pass)

Resolves: https://github.com/kubeflow/pipelines/issues/11899

PR #12399, opened 2025-10-29
Refine optional Kubernetes runtime test pipeline

sduvvuri1603:
**Description of changes**

- Follow-up to [#12377](https://github.com/kubeflow/pipelines/pull/12377) to ensure the optional Kubernetes runtime test actually exercises the driver logic.
- Collapse `missing_kubernetes_optional_inputs_pipeline` to a single `log_message` task and apply every K8s helper directly on that task.
- Add `optional_image_pull_secret_name` so image pull secrets participate in the skip logic.
- Remove the `dsl.If` guards; the helpers now see either real values or `null`, letting the driver hit both the “patch applied” path and the `ErrResolvedParameterNull` skip path.


VaniHaripriya:
@sduvvuri1603 Please add the TestData entry for new Pipeline here- https://github.com/kubeflow/pipelines/blob/master/sdk/python/test/compilation/pipeline_compilation_test.py#L271.

test_data/sdk_compiled_pipelines/valid/critical/missing_kubernetes_optional_inputs.py
```diff
@@ -11,58 +11,60 @@ def log_message(message: str) -> None:
...
 ):
-    base_task = log_message(message="baseline task")
-    base_task.set_caching_options(enable_caching=False)
+    task = log_message(message="baseline task")
+    task.set_caching_options(enable_caching=False)
 
-    with dsl.If(enable_affinity == True):
-        affinity_task = log_message(message="node affinity applied")
-        affinity_task.set_caching_options(enable_caching=False)
+    if optional_affinity is not None:
```
mprahl: You can't use conditionals like this in a `dsl.pipeline` because this code executes at compile time, not runtime.

Just remove the if statements. We also want to always apply the unset Kubernetes configuration to test your new code.

sduvvuri1603: My bad. Removed the if statements and pushed the updated code changes.

PR #12401, opened 2025-10-30
Remove unused Semaphore_key and mutex_name fields 

sduvvuri1603:

- Removes the unused semaphore_key and mutex_name options from dsl.PipelineConfig and related compiler wiring.
- Deletes the legacy sample pipeline plus golden YAMLs that referenced those fields.
- Updates the v2 proto schema (and regenerated Go/Python bindings, backend fixtures, and docs) to drop the fields entirely.
- Leaves the new pipeline_run_parallelism work for a follow-up PR (as tracked in [RHOAIENG-13108](https://issues.redhat.com/browse/RHOAIENG-13108)).

**-Next Steps**
- Follow-up PR will introduce pipeline_run_parallelism, backend ConfigMap management, and tests per RHOAIENG-13108.


api/v2alpha1/pipeline_spec.proto
```diff
@@ -1166,11 +1166,8 @@ message KubernetesWorkspaceConfig {
 
 // Spec for pipeline-level config options. See PipelineConfig DSL class.
 message PipelineConfig {
-  // Name of the semaphore key to control pipeline concurrency
-  string semaphore_key = 1;
-
-  // Name of the mutex to ensure mutual exclusion
-  string mutex_name = 2;
+  reserved 1, 2;
+  reserved "semaphore_key", "mutex_name";
```
mprahl: Could you check if this is needed since I don't think the Proto representation of it is ever sent on the wire? We might be able to just delete these.

sduvvuri1603: You’re right that we never ship this proto over the wire, but I hit the "validate-backwards-compatibility" job failing  when I tried deleting the fields—the check only passed once I kept resource_ttl workspace at 3 and 4 and marked 1/2 reserved so those numbers stay off-limits. 

I can drop the reserved lines if you’d prefer, but we’d still need to leave the remaining fields numbered 3 and 4 to keep that check. Let me know what you’d like me to do and I’ll update the PR.

HumairAK: I think it's fine to remove them since like Matt said these are configs that are never sent over the wire in a grpc call. @sduvvuri1603 you can regenerate the files for this check by running it locally with the env var `UPDATE_EXPECTED=true`  ([ref](https://github.com/kubeflow/pipelines/blob/348a6f5311ace3057fae4656dea8bc073ab48dc9/backend/test/proto_tests/util.go#L58)), this will pass the checks.

sduvvuri1603:
### Manual verification

Environment prep:
- Built a local image `local/python-kfp:3.11-kfp2146` (python:3.11 + kfp==2.14.6) and loaded it into the Kind cluster.
- Updated `pipeline_with_workspace.py` temporarily to reference that image.
- Rebuilt and reloaded `kfp-driver` and `kfp-launcher`.

Baseline (before removing `semaphore_key` / `mutex_name`):
- Pipeline version: `temp pipeline`
- Run: `Run of temp pipeline (a2413)` – succeeded 

After removal and rebuilt components:
- Pipeline version: `temp pipeline-version-at-2025-11-05t19-19-55.684z`
- Run: `Run of temp pipeline-version-at-2025-11-05t19-19-55.684z` – succeeded 

Both runs use the same pipeline definition (identical workspace configuration); the only difference is the backend binaries, demonstrating that removing the unused fields does not affect pipeline behavior.

api/v2alpha1/pipeline_spec.proto
```diff
@@ -1166,12 +1166,6 @@ message KubernetesWorkspaceConfig {
...
 message PipelineConfig {
-  // Name of the semaphore key to control pipeline concurrency
-  string semaphore_key = 1;
-
-  // Name of the mutex to ensure mutual exclusion
-  string mutex_name = 2;
-
   // Time to live configuration after the pipeline run is completed for
   // ephemeral resources created by the pipeline run.
   int32 resource_ttl = 3;
```
mprahl: Based on your manual verification, it sounds like we can reorder these to 1 and 2.

sduvvuri1603: yes, yet to push the updated code. Currently working on some final code changes

mprahl:
@sduvvuri1603 please rebase and I think the CI will pass.

mprahl:
@sduvvuri1603 it looks like there are several unrelated changes (e.g. Argo Workflow version changes in CI). Could you please keep this PR tightly scoped to just the Proto and SDK changes?

backend/test/proto_tests/testdata/generated-1791485/experiment.json
```diff
@@ -1,9 +1,9 @@
 {
-  "experiment_id":  "exp-456",
-  "display_name":  "Production Data Processing Experiment",
-  "description":  "Experiment for testing production data processing pipeline",
-  "created_at":  "2024-01-01T12:00:00Z",
-  "namespace":  "namespace1",
-  "storage_state":  "AVAILABLE",
-  "last_run_created_at":  "2024-01-01T12:00:00Z"
-}
+  "experiment_id": "exp-456",
```
mprahl, line 2: @sduvvuri1603 what caused all these formatting changes? Was it a script you ran?

sduvvuri1603: Yes, they are rewritten automatically when I ran UPDATE_EXPECTED=true go test ./backend/test/proto_tests. The current Go  emits a single space instead of two, so the regenerated file dropped that extra space.

PR #12248, opened 2025-09-16
feat(backend, sdk): Add custom_path field to RuntimeArtifact

alyssacgoins:
Resolves #12386 

**Description of your changes:**
Add a `custom_path` field to the `RuntimeArtifact` object that overrides the object's URI field during artifact upload, allowing artifacts to be uploaded from non-local storage.

Changes validated via unit test in `launcher_v2.Test_retrieve_artifact_path() ` as well as the sample pipeline `component_with_runtime_artifact_custom_path.py` added to `test_data/sdk_compiled_pipelines/valid/critical`, which sets and validates a custom artifact path.

**Checklist:**
- [X] You have [signed off your commits](https://www.kubeflow.org/docs/about/contributing/#sign-off-your-commits)
- [X] The title for your pull request (PR) should follow our title convention. [Learn more about the pull request title convention used in this repository](https://github.com/kubeflow/pipelines/blob/master/CONTRIBUTING.md#pull-request-title-convention).

PR #12402, opened 2025-10-31
Correct Runtime Artifact custom path logic/testing.

alyssacgoins:
Updates the logic added in #12248 and also adds spec files to compiled workflow and pipeline files in order to correct the `pipeline_with_artifact_custom_path` test case. 

**Checklist:**

### DIFF
--- a/sdk/python/kfp/dsl/types/artifact_types.py
+++ b/sdk/python/kfp/dsl/types/artifact_types.py
@@ -91,7 +91,7 @@ class Artifact:
 
     @path.setter
     def path(self, path: str) -> None:
-        self._set_custom_path(path)
+        self._set_path(path)
 
     def _get_path(self) -> Optional[str]:
         if self.custom_path:
@@ -114,7 +114,7 @@ class Artifact:
 
     @property
     def custom_path(self) -> str:
-        return self._custom_path
+        return self._get_custom_path()
 
     def _get_custom_path(self) -> str:
         return self._custom_path
@@ -128,6 +128,10 @@ class Artifact:
     @custom_path.setter
     def custom_path(self, value: str):
         self._custom_path = value
+
+    def set_path(self, path: str) -> None:
+        # If user specified a custom path, use it instead of the default path.
+        self._custom_path = path
 
 
 def convert_local_path_to_remote_path(path: str) -> str:
--- a/test_data/sdk_compiled_pipelines/valid/critical/pipeline_with_artifact_custom_path.py
+++ b/test_data/sdk_compiled_pipelines/valid/critical/pipeline_with_artifact_custom_path.py
@@ -1,22 +1,40 @@
-from kfp.dsl import Output
-from kfp.dsl.types.artifact_types import Artifact
-from kfp.v2 import dsl
-
-
-@dsl.component
-def generate_artifact() -> list:
-    return [1, 2, 3, 4]
+from kfp.dsl import Output, Input
+from kfp.dsl.types.artifact_types import Dataset
+from kfp.v2 import dsl, compiler
 
 @dsl.component
-def validate_artifact_path(exp_path: str, input_list: Output[Artifact]) -> bool:
-    if input_list.path is not exp_path:
-        raise ValueError(f"File uri is {input_list.path} but should be {exp_path}.")
+def component_output_artifact(out_dataset: Output[Dataset]):
+    out_dataset.set_path('/tmp/out_dataset')
 
-@dsl.pipeline
+    with open(out_dataset.path, 'w') as f:
+        f.write('Hello, World!')
+
+    if out_dataset.path != '/tmp/out_dataset':
+        raise ValueError(f"File path is {out_dataset.path} but should be '/tmp/out_dataset'.")
+    if out_dataset.custom_path != '/tmp/out_dataset':
+        raise ValueError(f"File uri is {out_dataset.custom_path} but should be '/tmp/out_dataset'.")
+
+@dsl.component
+def validate_input_artifact(in_dataset: Input[Dataset]) -> bool:
+    if in_dataset is None:
+        raise ValueError("Input artifact is None.")
+
+    with open(in_dataset.path, 'r') as f:
+        for line in f:
+            if line != 'Hello, World!':
+                raise ValueError(f"File content is {line} but should be 'Hello, World!'.")
+    return True
+
+@dsl.pipeline(description='pipeline with artifact with custom path')
 def pipeline_with_custom_path_artifact():
-    # Generate artifact, and set its custom path.
-    output_artifact_task = generate_artifact()
-    output_artifact_task.output.set_path('/etc/test/file/path')
+    # Generate an artifact, set a custom path and validate the path.
+    task_output_artifact = component_output_artifact().set_caching_options(False)
 
-    # Validate generated artifact's path.
-    validate_artifact_task = validate_artifact_path(path='/etc/test/file/path', input_list=output_artifact_task.output)
+    # Take the output artifact from above as input, and validate as non-nil.
+    task_validate_input_artifact = validate_input_artifact(in_dataset=task_output_artifact.outputs['out_dataset'])
+
+if __name__ == '__main__':
+    compiler.Compiler().compile(
+        pipeline_func=pipeline_with_custom_path_artifact,
+        package_path=__file__.replace('.py', '.yaml'))
+
--- a/backend/src/v2/component/launcher_v2.go
+++ b/backend/src/v2/component/launcher_v2.go
@@ -945,6 +945,10 @@ func mergeRuntimeArtifacts(src, dst *pipelinespec.RuntimeArtifact) {
 			}
 		}
 	}
+
+	if src.CustomPath != nil && *src.CustomPath != "" {
+		dst.CustomPath = src.CustomPath
+	}
 }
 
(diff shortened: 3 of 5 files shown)


test_data/sdk_compiled_pipelines/valid/critical/pipeline_with_artifact_custom_path.py
```diff
@@ -1,22 +1,28 @@
...
-    return [1, 2, 3, 4]
 
 @dsl.component
-def validate_artifact_path(exp_path: str, input_list: Output[Artifact]) -> bool:
-    if input_list.path is not exp_path:
-        raise ValueError(f"File uri is {input_list.path} but should be {exp_path}.")
+def validate_custom_artifact_path(num: int, out_dataset: Output[Dataset]):
+    with open(out_dataset.path, 'w') as f:
+        f.write(str(2 * num))
+    out_dataset._set_path('/etc/test/file/path')
```
mprahl: I thought the user was supposed to call it via `out_dataset.set_path('/etc/test/file/path')` or `out_dataset.custom_path = '/etc/test/file/path'`?

alyssacgoins: The Jira for this task specifies `I can call set_path on the output artifact` - I took that to mean that when a user calls `set_path`, it sets the custom path. But I also see both interpretations

alyssacgoins: @mprahl let me know if you think it makes more sense to update to your comment

mprahl: @alyssacgoins it was more that it should be `set_path` and not `_set_path` to test the way the user would leverage the feature.

test_data/sdk_compiled_pipelines/valid/critical/pipeline_with_artifact_custom_path.py
```diff
@@ -1,22 +1,28 @@
...
+        raise ValueError(f"File uri is {out_dataset.custom_path} but should be '/etc/test/file/path'.")
 
-@dsl.pipeline
+@dsl.pipeline(description='pipeline with artifact with custom path')
 def pipeline_with_custom_path_artifact():
-    # Generate artifact, and set its custom path.
-    output_artifact_task = generate_artifact()
-    output_artifact_task.output.set_path('/etc/test/file/path')
+    # Generate an artifact, set a custom path and validate the path.
+    task = validate_custom_artifact_path(num=1).set_caching_options(False)
```
mprahl: Please add another task that imports the artifact to ensure the correct thing got uploaded and then redownloaded.

test_data/sdk_compiled_pipelines/valid/critical/pipeline_with_artifact_custom_path.py
```diff
@@ -1,22 +1,28 @@
...
-    return [1, 2, 3, 4]
 
 @dsl.component
-def validate_artifact_path(exp_path: str, input_list: Output[Artifact]) -> bool:
-    if input_list.path is not exp_path:
-        raise ValueError(f"File uri is {input_list.path} but should be {exp_path}.")
+def validate_custom_artifact_path(num: int, out_dataset: Output[Dataset]):
+    with open(out_dataset.path, 'w') as f:
+        f.write(str(2 * num))
+    out_dataset.set_path('/etc/test/file/path')
```
mprahl: Could you use a writable path for non-root users like something in `/tmp`?

alyssacgoins: Updated

test_data/sdk_compiled_pipelines/valid/critical/pipeline_with_artifact_custom_path.py
alyssacgoins: Updated

test_data/sdk_compiled_pipelines/valid/critical/pipeline_with_artifact_custom_path.py
```diff
@@ -1,22 +1,35 @@
...
+from kfp.dsl.types.artifact_types import Dataset
+from kfp.v2 import dsl, compiler
 
 @dsl.component
-def generate_artifact() -> list:
-    return [1, 2, 3, 4]
+def component_output_artifact(out_dataset: Output[Dataset]):
+    with open(out_dataset.path, 'w') as f:
+        f.write('Hello, World!')
+    out_dataset.set_path('/tmp/output/dataset')
```
mprahl: Call this before writing to `out_dataset.path`. Then `validate_input_artifact` should assert the contents are as expected.

alyssacgoins: Not sure I follow - to clarify, which line should be executed before writing to `out_dataset.path`?

alyssacgoins: Updated

backend/src/v2/metadata/converter.go
```diff
@@ -116,6 +116,10 @@ func toMLMDArtifact(runtimeArtifact *pipelinespec.RuntimeArtifact) (*pb.Artifact
 		Properties:       make(map[string]*pb.Value),
 		CustomProperties: make(map[string]*pb.Value),
 	}
+	// If runtime artifact is set with a custom path, add this path to the custom properties map.
+	if runtimeArtifact.CustomPath != nil {
+		artifact.CustomProperties["custom_path"] = stringMLMDValue(*runtimeArtifact.CustomPath)
```
mprahl: I don't think we should be storing this as a custom property. It's not relevant to the artifact once the artifact is uploaded.

backend/src/v2/metadata/converter.go
```diff
@@ -177,13 +181,20 @@ func toRuntimeArtifact(artifact *pb.Artifact) (*pipelinespec.RuntimeArtifact, er
 	errorF := func(err error) (*pipelinespec.RuntimeArtifact, error) {
 		return nil, fmt.Errorf("failed to convert MLMD artifact to RuntimeArtifact: %w", err)
 	}
+	// Retrieve custom_path value from artifact.CustomProperties, if present.
+	var customPathStr string
+	if artifact.CustomProperties != nil {
+		customPath := artifact.CustomProperties["custom_path"]
```
mprahl: The custom path flow should be:
1. Executor sets the custom path property on the output artifact in the executor output
2. The `uploadOutputArtifacts` function checks if the custom path is present and uses that to upload instead of the predefined path
3. The custom path is no longer needed and be ignored and not set on the MLMD artifact.

backend/src/v2/metadata/converter.go
alyssacgoins: The issue is that `metadata_store.Artifact`, which I believe is a third-party resource, does not have a `custom_path` field. If I'm following the logic correctly, type `metadata_store.Artifact` is uploaded to MLMD - I added custom_path as a property in order to preserve this value upon retrieval from MLMD

backend/src/v2/metadata/converter.go
alyssacgoins: The issue I found is that, while I modified `pipeline_spec.RuntimeArtifact` to include the `custom_path` field, and uploadOutputArtifacts uploads to that custom path, when the MLMD artifact is retrieved with no custom path set on the object, the executor does not know what path to retrieve the file from when `launcher_v2.retrieveArtifactPath()` executes

alyssacgoins: Resolved

backend/src/v2/metadata/converter.go
alyssacgoins: Resolved

test_data/sdk_compiled_pipelines/valid/critical/pipeline_with_artifact_custom_path.py
```diff
@@ -1,22 +1,35 @@
...
+    if out_dataset.custom_path != '/tmp/out_dataset':
+        raise ValueError(f"File uri is {out_dataset.custom_path} but should be '/tmp/out_dataset'.")
 
 @dsl.component
-def validate_artifact_path(exp_path: str, input_list: Output[Artifact]) -> bool:
-    if input_list.path is not exp_path:
-        raise ValueError(f"File uri is {input_list.path} but should be {exp_path}.")
+def validate_input_artifact(in_dataset: Input[Dataset]) -> bool:
+    if in_dataset is None:
+        raise ValueError("Input artifact is None.")
```
mprahl, line 20: Could you please have the test read the input dataset at `in_dataset.path` and assert it equals `Hello, World!`?

sdk/python/kfp/dsl/types/artifact_types.py
```diff
@@ -112,23 +112,28 @@ def _get_path(self) -> Optional[str]:
...
+        return self._get_custom_path()
+
+    @custom_path.getter
+    def custom_path(self):
         return self._custom_path
 
     def _get_custom_path(self) -> str:
         return self._custom_path
 
-    def _set_path(self, path: str) -> None:
```
mprahl, line 122: Optional: It'd be nice if this was kept at its original location to not shadow the original commit history.

mprahl (approved):
/approve
/lgtm

PR #12406, opened 2025-11-03
chore: update SeaweedFS to 4.00 and make it more robust

akagami-harsh:
- Update seaweedfs image to 3.98

### DIFF
--- a/manifests/kustomize/third-party/seaweedfs/base/seaweedfs/seaweedfs-deployment.yaml
+++ b/manifests/kustomize/third-party/seaweedfs/base/seaweedfs/seaweedfs-deployment.yaml
@@ -24,9 +24,17 @@ spec:
           type: RuntimeDefault
       containers:
       - name: seaweedfs
-        envFrom:
-        - secretRef:
-            name: mlpipeline-minio-artifact
+        env:
+        - name: AWS_ACCESS_KEY_ID
+          valueFrom:
+            secretKeyRef:
+              name: mlpipeline-minio-artifact
+              key: accesskey
+        - name: AWS_SECRET_ACCESS_KEY
+          valueFrom:
+            secretKeyRef:
+              name: mlpipeline-minio-artifact
+              key: secretkey
         image: "chrislusf/seaweedfs:4.00"
         args:
         - "server"
@@ -35,24 +43,6 @@ spec:
         - "-iam"
         - "-filer"
         - "-master.volumePreallocate=false"
-        lifecycle:
-          postStart:
-            exec:
-              command:
-              - /bin/sh
-              - -ec
-              - |
-                # wait until seaweedfs master is ready
-                for i in $(seq 1 120); do
-                  if wget -q --spider http://127.0.0.1:8333/status; then
-                    break
-                  fi
-                  sleep 2
-                done
-                # create bucket if not exists (ignore error if exists)
-                echo "s3.bucket.create --name mlpipeline" | /usr/bin/weed shell || true
-                # configure admin user using keys from secret
-                echo "s3.configure -user kubeflow-admin -access_key $accesskey -secret_key $secretkey -actions Admin -apply" | /usr/bin/weed shell
         ports:
         - containerPort: 8333
         - containerPort: 8111


juliusvonkohout:
There was a recent fix here https://github.com/kubeflow/pipelines/commit/fc69198d86e91241f3ddfac710edb279f602851f please make sure to include it. You can also start with 3.93 and see how high you can get.

juliusvonkohout:
Or even better 4.00 https://github.com/seaweedfs/seaweedfs/releases/tag/4.00 https://hub.docker.com/layers/chrislusf/seaweedfs/4.00/images/sha256-7f3f541233fad057f1e37678eafb618afeab0f4338a0dc9197fa23554189f615

manifests/kustomize/third-party/seaweedfs/base/seaweedfs/seaweedfs-deployment.yaml
pschoen-itsc, line 1: L. 76: The new image uses a non-root user by default

manifests/kustomize/third-party/seaweedfs/base/seaweedfs/seaweedfs-deployment.yaml
```diff
@@ -27,7 +27,7 @@ spec:
         envFrom:
         - secretRef:
             name: mlpipeline-minio-artifact
-        image: "chrislusf/seaweedfs:3.92"
+        image: "chrislusf/seaweedfs:4.00"
```
pschoen-itsc, line 38: Don't know if this would be within the scope of this change, but with 3.96 [this feature](https://github.com/seaweedfs/seaweedfs/pull/7030/files) was released, which would enable us to create an admin user directly from the env vars.

akagami-harsh: hmm this looks amazing, i'll try it out, if it works i'll open a new PR for it

manifests/kustomize/third-party/seaweedfs/base/seaweedfs/seaweedfs-deployment.yaml
juliusvonkohout: Then lets switch to UID 1000 GID 1000. GID 0 instead of 1000 is actually best practice, but sicne this is really only used by seaweedfs lets stick to the upstream 1000:1000 for now.

manifests/kustomize/third-party/seaweedfs/base/seaweedfs/seaweedfs-deployment.yaml
juliusvonkohout: we can also do so in the same PR here.

juliusvonkohout:
@droctothorpe for approval
@akagami-harsh @pschoen-itsc for final review
/lgtm from my side so far.

juliusvonkohout:
actually since i am in the owners file for seaweedfs https://github.com/kubeflow/pipelines/blob/17c7d751b468095f7914b40b27cd7eaf546052a0/manifests/kustomize/third-party/seaweedfs/OWNERS and the tests are green i should do
/approve
and @pschoen-itsc  and @droctothorpe can /lgtm if they think that it is good enough.

pschoen-itsc (approved):
/lgtm

juliusvonkohout:
@pschoen-itsc has found out that we might have to revert his due to https://github.com/seaweedfs/seaweedfs/wiki/S3-Credentials#priority-system @akagami-harsh. Can you verify that the administrator password is not gone after sync.py has added more users and after restarting the pod ? Because if sync.py add users on the filer level and there is no user merging, then the administrator password is gone after a restart. We should of cousrse keep seaweedfs 4.00

juliusvonkohout:
> @pschoen-itsc has found out that we might have to revert his due to https://github.com/seaweedfs/seaweedfs/wiki/S3-Credentials#priority-system @akagami-harsh. Can you verify that the administrator password is not gone after sync.py has added more users and after restarting the pod ? Because if sync.py add users on the filer level and there is no user merging, then the administrator password is gone after a restart. We should of cousrse keep seaweedfs 4.00

CC @droctothorpe

droctothorpe:
Thanks for the heads up, @juliusvonkohout . Much appreciated. @akagami-harsh do you want to submit the revert PR? Tag me in it and I will expedite.

pschoen-itsc:
@juliusvonkohout @droctothorpe @akagami-harsh Did a quick revert here https://github.com/kubeflow/pipelines/pull/12460

PR #12306, opened 2025-10-01
# Fix: Add driver pod labels/annotations configuration support

thc1006:
## Description

This PR addresses [issue #12015](https://github.com/kubeflow/pipelines/issues/12015) by adding the ability to configure driver pod labels and annotations at the KFP API server level. This enables driver pods to work properly with Istio service mesh in STRICT mTLS mode and other infrastructure requirements.

## Problem

When Istio STRICT mTLS is enabled across the mesh, KFP driver pods cannot connect to in-mesh services (MinIO, MLMD) because they lack the necessary sidecar injection labels. This causes "connection reset by peer" errors and prevents pipeline execution.

## Solution

Implement admin-level configuration for driver pod metadata through:
- Environment variables (`DRIVER_POD_LABELS`, `DRIVER_POD_ANNOTATIONS`)
- ConfigMap-based configuration
- Pre-configured Istio overlay for production deployments

### Backend
- ✅ Added `driver_config.go` with configuration reader supporting JSON and k=v formats
- ✅ Added comprehensive unit tests (37 test cases) in `driver_config_test.go`
- ✅ Modified Argo compiler to apply labels/annotations to driver pod templates
- ✅ Implemented graceful degradation and reserved label protection

### Deployment
- ✅ Updated `ml-pipeline-apiserver-deployment.yaml` with optional ConfigMap references
- ✅ Added ConfigMap generator in base kustomization
- ✅ Created Istio STRICT mTLS overlay with optimized configuration
- ✅ Added PeerAuthentication and AuthorizationPolicy for service mesh security

### Testing
- ✅ Unit tests for configuration parsing and validation
- ✅ E2E test suite for Istio STRICT mTLS scenarios
- ✅ Test coverage for edge cases and error handling

### Documentation
- ✅ Comprehensive configuration guide with examples
- ✅ Istio integration documentation
- ✅ Troubleshooting guide and migration instructions

### Basic Configuration
```bash
# Set driver pod labels for Istio sidecar injection
export DRIVER_POD_LABELS='{"sidecar.istio.io/inject":"true"}'
export DRIVER_POD_ANNOTATIONS='{"proxy.istio.io/config":"{\"holdApplicationUntilProxyStarts\":true}"}'
```

### Production Deployment (Istio)
```bash
# Deploy with Istio overlay
kubectl apply -k manifests/kustomize/overlays/istio-strict-mtls
```

### Run Unit Tests
```bash
cd backend/src/apiserver/config
go test -v ./...
```

### Run E2E Tests
[...]

PR #12417, opened 2025-11-07
feat: Support driver pod labels and annotations configuration at API server startup

### LINKED ISSUE #12015: [backend] Add the ability to specify driver labels/annotations as a KFP api server configuration
Edited by: @HumairAK 

Based on the discussion below, this issue is now allowing admin users to configure what labels/annotations are added to driver pods at the KFP API Server level.

----
### Environment

Kubeflow manifests (kustomize)
*  KFP version: 
2.5.0
*  KFP SDK version: 
2.13.0

### Steps to reproduce

Enable STRICT mTLS on the entire mesh, then run a pipeline.

```
apiVersion: security.istio.io/v1beta1
kind: PeerAuthentication
metadata:
  name: mesh-traffic
  namespace: istio-system
spec:
  mtls:
    mode: STRICT
```

### Expected result

Pipeline runs successfully

### Materials and Reference

KFP driver pods run without Istio sidecars, therefore they can't connect to services within the mesh (if mTLS STRICT mode is enabled).

In particular, the default Kubeflow installation is shipped with MinIO object store, the driver pod fails when trying to connect to it:

```log
level=info msg="Saving file to s3" bucket=mlpipeline endpoint="minio-service.kubeflow:9000" key=artifacts/mnist-pipeline-mdtkv/2025/06/25/mnist-pipeline-mdtkv-system-dag-driver-938794099/main.log path=/tmp/argo/outputs/logs/main.log
level=info msg="Transient error: Get \"http://minio-service.kubeflow:9000/mlpipeline/?location=\": read tcp 10.244.2.191:60656->10.108.206.88:9000: read: connection reset by peer"
```

Notice also the MinIO pod Istio sidecar that reports a mismatch of TLS configuration between
the downstream client (driver pod) and the upstream server (MinIO Istio sidecar):

```log
"- - -" 0 NR filter_chain_not_found - "-" 0 0 0 - "-" "-" "-" "-" "-" - - 10.244.2.152:9000 10.244.2.191:60656 - -
```

Another service which is impacted is the MLDM gRPC service, the driver pod fails with error:

```log
KFP driver: driver.RootDAG(pipelineName=mnist-pipeline, runID=xyz, runtimeConfig, componentSpec) failed: Failed GetContextByTypeAndName(type="system.Pipeline", name="mnist-pipeline"): rpc error: code = Unavailable desc = connection error: desc = "error reading server preface: read tcp 10.244.2.191:51708->10.105.73.97:8080: read: connection reset by peer"
```

I could fix the above errors errors by selectively disabling mTLS:

```
apiVersion: security.istio.io/v1beta1
kind: PeerAuthentication
metadata:
  name: minio-traffic
  namespace: kubeflow
spec:
  selector:
    matchLabels:
      app: minio
  portLevelMtls:
    9000:
      mode: PERMISSIVE
---
apiVersion: security.istio.io/v1beta1
kind: PeerAuthentication
metadata:
  name: metadata-grpc-server-traffic
  namespace: kubeflow
spec:
  selector:
    matchLabels:
      component: metadata-grpc-server
  portLevelMtls:
    8080:
      mode: PERMISSIVE 
```

I wonder whether this is the right approach or there's a better solution to address mTLS communication.
Would it be an option to run driver pods within the mesh as well?
Thank you

---

thc1006:
## Summary

This PR implements admin-level configuration for driver pod labels and annotations, addressing issue #12015. Configuration is loaded once at API server startup and cached for performance.

### Key Changes

- **Startup-time configuration loading**: Driver pod config (labels and annotations) is initialized once when the API server starts via `common.InitDriverPodConfig()` in backend/src/apiserver/main.go
- **Thread-safe caching**: Uses `sync.RWMutex` for safe concurrent access to cached configuration
- **Reserved prefix filtering**: Automatically filters out labels and annotations with the `pipelines.kubeflow.org/` prefix to prevent overriding system-managed metadata
- **JSON configuration format**: Supports JSON format for configuration (as per reviewer feedback)
- **Comprehensive testing**: Unit tests covering all configuration scenarios including JSON string input from ConfigMap

### All Reviewer Requirements Addressed (7/7)

✅ **Removed .gitignore changes** - Only relevant backend and manifest changes included  
✅ **Using Viper configuration** - Integrated with existing Viper config system  
✅ **JSON format only** - Removed YAML support, JSON format only  
✅ **Kubernetes-compatible format** - Standard Kubernetes label/annotation format  
✅ **Startup-time loading with caching** - Config loaded once at startup, cached for performance  
✅ **Documentation in backend/README.md** - Concise documentation added  
✅ **Removed unimplemented options** - Removed environment variable and serviceaccount options

### Testing

Local tests passed:
- ✅ All driver config unit tests (backend/src/apiserver/common/driver_config_test.go)
- ✅ All argocompiler unit tests  
- ✅ go mod tidy
- ✅ go vet
- ✅ gofmt

### Configuration Example

```json
{
  "DriverPodLabels": {
    "custom-label": "value",
    "team": "ml-platform"
  },
  "DriverPodAnnotations": {
    "sidecar.istio.io/inject": "true",
    "custom-annotation": "value"
  }
}
```

Resolves: #12015

---

**Note**: This PR supersedes #12306 which was closed due to commit history issues. All code changes and reviewer feedback remain fully addressed in this clean PR.

### DIFF
--- a/backend/README.md
+++ b/backend/README.md
@@ -62,6 +62,35 @@ Run
 docker build . -f backend/Dockerfile.conformance -t <tag>
 ```
 
+## API Server Configuration
+
+### Driver Pod Labels and Annotations
+
+The API server supports configuring custom labels and annotations for driver pods through the configuration file or the Kubernetes ConfigMap. This is useful for integration with service mesh (Istio), monitoring systems, or other infrastructure requirements.
+
+**Configuration via config.json:**
+```json
+{
+  "DRIVER_POD_LABELS": {
+    "sidecar.istio.io/inject": "true",
+    "app": "ml-pipeline-driver"
+  },
+  "DRIVER_POD_ANNOTATIONS": {
+    "proxy.istio.io/config": "{\"holdApplicationUntilProxyStarts\":true}"
+  }
+}
+```
+
+**Configuration via Kubernetes ConfigMap (`pipeline-install-config`):**
+```yaml
+DRIVER_POD_LABELS: '{"sidecar.istio.io/inject":"true"}'
+DRIVER_POD_ANNOTATIONS: '{"proxy.istio.io/config":"{\"holdApplicationUntilProxyStarts\":true}"}'
+```
+
+When configured via the ConfigMap, every value must be a JSON string. Write `"true"` rather than the bare boolean `true`, and never `null`. The configuration is validated at API server startup: malformed JSON, any value that is not a JSON string, and label keys, label values, or annotation keys that Kubernetes would reject all cause startup to fail with a descriptive error rather than being silently ignored. Annotation values are free form, so a value such as an inline JSON document is accepted.
+
+Note: Labels and annotations with the prefix `pipelines.kubeflow.org/` are reserved and will be filtered out to prevent overriding system metadata. Changes to driver pod configuration require an API server restart.
+
 ## API Server Development
 
 ### Run the KFP Backend Locally With a Kind Cluster
--- a/backend/src/apiserver/common/driver_config.go
+++ b/backend/src/apiserver/common/driver_config.go
@@ -1,0 +1,253 @@
+// Copyright 2025 The Kubeflow Authors
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+// https://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+// Package common provides shared utilities and configuration for the KFP API server.
+package common
+
+import (
+	"encoding/json"
+	"fmt"
+	"strings"
+	"sync"
+
+	"github.com/golang/glog"
+	"github.com/spf13/viper"
+	"k8s.io/apimachinery/pkg/util/validation"
+)
+
+const (
+	// Config keys. Viper derives the environment variable name from the key, so these
+	// are also the names the API server Deployment must use when it wires the values
+	// in from the pipeline-install-config ConfigMap.
+	DriverPodLabels      = "DRIVER_POD_LABELS"
+	DriverPodAnnotations = "DRIVER_POD_ANNOTATIONS"
+
+	// Reserved label prefix that will be filtered out to prevent
+	// overriding system labels that control workflow behavior
+	ReservedLabelPrefix = "pipelines.kubeflow.org/"
+)
+
+// DriverPodConfig holds the labels and annotations applied to driver pods.
+type DriverPodConfig struct {
+	Labels      map[string]string
+	Annotations map[string]string
+}
+
+var (
+	// cachedDriverPodConfig holds the driver pod configuration loaded at startup.
+	// It is swapped atomically by InitDriverPodConfig so readers never observe a
+	// partly updated state. A nil pointer means the config has not been loaded yet.
+	cachedDriverPodConfig *DriverPodConfig
+	// driverConfigMutex protects the cached config from concurrent access.
+	driverConfigMutex sync.RWMutex
+)
+
+// InitDriverPodConfig loads, validates, and caches driver pod configuration at API
+// server startup. This should be called once after Viper configuration is loaded.
+// It returns an error when the configuration is present but cannot be parsed, so the
+// API server fails fast instead of starting with a silently empty configuration.
+func InitDriverPodConfig() error {
+	// Build and validate the new config before taking the lock for the swap.
+	cfg, err := newDriverPodConfig()
+	if err != nil {
+		return err
+	}
+
+	driverConfigMutex.Lock()
+	defer driverConfigMutex.Unlock()
+
+	if cachedDriverPodConfig != nil {
+		glog.Info("Re-initializing driver pod configuration")
+	}
+	cachedDriverPodConfig = cfg
+
+	glog.Infof("Driver pod configuration initialized: %d labels, %d annotations",
+		len(cfg.Labels), len(cfg.Annotations))
+	if len(cfg.Labels) > 0 {
+		glog.V(1).Infof("Driver pod labels: %v", cfg.Labels)
+	}
+	if len(cfg.Annotations) > 0 {
+		glog.V(1).Infof("Driver pod annotations: %v", cfg.Annotations)
+	}
+	return nil
+}
+
+// newDriverPodConfig reads the driver pod labels and annotations from Viper,
+// validates them, and returns a fully populated config or an error. Keeping
+// construction separate from the cache swap guarantees the update is applied
+// completely or not at all.
+func newDriverPodConfig() (*DriverPodConfig, error) {
+	if err := validateMapConfig(DriverPodLabels); err != nil {
+		return nil, err
+	}
+	if err := validateMapConfig(DriverPodAnnotations); err != nil {
+		return nil, err
+	}
+
+	labels := filterReservedEntries(GetMapConfig(DriverPodLabels))
+	if err := validateLabels(labels); err != nil {
+		return nil, err
+	}
+
+	annotations := filterReservedEntries(GetMapConfig(DriverPodAnnotations))
+	if err := validateAnnotationKeys(annotations); err != nil {
+		return nil, err
+	}
+
+	return &DriverPodConfig{
+		Labels:      labels,
+		Annotations: annotations,
+	}, nil
+}
+
+// validateLabels rejects label keys or values that Kubernetes would not accept, so a
+// bad entry fails at API server startup rather than later when the driver pod is
+// created and the pod is rejected by the API.
+func validateLabels(labels map[string]string) error {
+	for k, v := range labels {
+		if errs := validation.IsQualifiedName(k); len(errs) > 0 {
+			return fmt.Errorf("%s has an invalid label key %q: %s", DriverPodLabels, k, strings.Join(errs, "; "))
+		}
+		if errs := validation.IsValidLabelValue(v); len(errs) > 0 {
+			return fmt.Errorf("%s has an invalid label value %q for key %q: %s", DriverPodLabels, v, k, strings.Join(errs, "; "))
+		}
+	}
+	return nil
+}
+
+// validateAnnotationKeys rejects annotation keys that Kubernetes would not accept.
+// Annotation values are free form, so only the keys are checked.
+func validateAnnotationKeys(annotations map[string]string) error {
+	for k := range annotations {
+		if errs := validation.IsQualifiedName(k); len(errs) > 0 {
+			return fmt.Errorf("%s has an invalid annotation key %q: %s", DriverPodAnnotations, k, strings.Join(errs, "; "))
+		}
+	}
+	return nil
+}
+
+// validateMapConfig rejects a ConfigMap value that Viper would not turn into the map the
+// operator meant. On the ConfigMap string path viper.GetStringMapString swallows problems
+// in two different ways, and neither produces an error: a boolean or a number makes it
+// drop the whole map, while a JSON null is quietly coerced to an empty string. Both leave
+// the operator with a configuration that looks accepted but silently does not do what
+// they asked for, so the raw value is probed here and every entry must be a JSON string.
+func validateMapConfig(name string) error {
+	if !viper.IsSet(name) {
+		return nil
+	}
+
+	// A native map (config.json) reports an empty string here, and a blank or whitespace
+	// only value means no configuration. Neither needs probing.
+	raw := strings.TrimSpace(viper.GetString(name))
+	if raw == "" {
+		return nil
+	}
+
+	var probe map[string]interface{}
+	if err := json.Unmarshal([]byte(raw), &probe); err != nil {
+		return fmt.Errorf("%s could not be parsed as a JSON object: %w (raw value: %q)", name, err, raw)
+	}
+	for k, v := range probe {
+		if _, ok := v.(string); !ok {
+			return fmt.Errorf("%s has a value that is not a JSON string for key %q; every value must be quoted, for example \"true\" rather than true or null (raw value: %q)", name, k, raw)
+		}
+	}
+	return nil
+}
+
+// GetDriverPodConfig returns a copy of the cached driver pod configuration, or nil when
+// no labels or annotations are configured. The returned value is safe to mutate.
+//
+// Both maps are copied under a single read lock. Calling the two getters in turn would
+// take the lock twice, which would let a concurrent InitDriverPodConfig slip in between
+// and hand back labels from one version of the config and annotations from another. That
+// would defeat the point of swapping the cache as one atomic snapshot.
+func GetDriverPodConfig() *DriverPodConfig {
+	driverConfigMutex.RLock()
+	defer driverConfigMutex.RUnlock()
+
+	if cachedDriverPodConfig == nil {
+		return nil
+	}
+
+	labels := copyMap(cachedDriverPodConfig.Labels)
+	annotations := copyMap(cachedDriverPodConfig.Annotations)
+	if len(labels) == 0 && len(annotations) == 0 {
+		return nil
+	}
+	return &DriverPodConfig{
+		Labels:      labels,
+		Annotations: annotations,
+	}
+}
+
+// GetDriverPodLabels returns a copy of cached driver pod labels from configuration.
+// Labels with pipelines.kubeflow.org/ prefix are filtered out during initialization.
+// Returns nil if InitDriverPodConfig has not been called or no labels are configured.
+func GetDriverPodLabels() map[string]string {
+	driverConfigMutex.RLock()
+	defer driverConfigMutex.RUnlock()
+
+	if cachedDriverPodConfig == nil {
+		return nil
+	}
+	return copyMap(cachedDriverPodConfig.Labels)
+}
+
+// GetDriverPodAnnotations returns a copy of cached driver pod annotations from configuration.
+// Returns nil if InitDriverPodConfig has not been called or no annotations are configured.
+func GetDriverPodAnnotations() map[string]string {
+	driverConfigMutex.RLock()
+	defer driverConfigMutex.RUnlock()
+
+	if cachedDriverPodConfig == nil {
+		return nil
+	}
+	return copyMap(cachedDriverPodConfig.Annotations)
+}
+
+// copyMap returns a shallow copy of the given map, or nil if the input is nil.
+func copyMap(m map[string]string) map[string]string {
+	if m == nil {
+		return nil
+	}
+	cp := make(map[string]string, len(m))
+	for k, v := range m {
+		cp[k] = v
+	}
+	return cp
+}
+
+// filterReservedEntries removes entries whose key starts with the reserved prefix.
+// Used for both labels and annotations to prevent overriding metadata managed by the system.
+// Returns nil if the input is nil or the result is empty after filtering.
+func filterReservedEntries(entries map[string]string) map[string]string {
+	if entries == nil {
+		return nil
+	}
+
+	filtered := make(map[string]string, len(entries))
+	for k, v := range entries {
+		if strings.HasPrefix(k, ReservedLabelPrefix) {
+			glog.Warningf("Ignoring reserved key %s (prefix %s is reserved)", k, ReservedLabelPrefix)
+			continue
+		}
+		filtered[k] = v
+	}
+	if len(filtered) == 0 {
+		return nil
+	}
+	return filtered
+}
--- a/backend/src/apiserver/common/driver_config_test.go
+++ b/backend/src/apiserver/common/driver_config_test.go
@@ -1,0 +1,600 @@
+// Copyright 2025 The Kubeflow Authors
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+// https://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+// Do not add t.Parallel() to these tests, because they share global Viper and package level state.
+package common
+
+import (
+	"strings"
+	"testing"
+
+	"github.com/spf13/viper"
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+)
+
+// resetDriverConfig resets the driver config state for testing
+func resetDriverConfig() {
+	driverConfigMutex.Lock()
+	defer driverConfigMutex.Unlock()
+	cachedDriverPodConfig = nil
+}
+
+func TestInitDriverPodConfig(t *testing.T) {
+	tests := []struct {
+		name                string
+		labels              map[string]string
+		annotations         map[string]string
+		expectedLabels      map[string]string
+		expectedAnnotations map[string]string
+	}{
+		{
+			name:                "empty config",
+			labels:              nil,
+			annotations:         nil,
+			expectedLabels:      nil,
+			expectedAnnotations: nil,
+		},
+		{
+			name: "valid config with filtering",
+			labels: map[string]string{
+				"sidecar.istio.io/inject":       "true",
+				"pipelines.kubeflow.org/system": "reserved",
+				"app":                           "test",
+			},
+			annotations: map[string]string{
+				"proxy.istio.io/config": "{\"holdApplicationUntilProxyStarts\":true}",
+			},
+			expectedLabels: map[string]string{
+				"sidecar.istio.io/inject": "true",
+				"app":                     "test",
+			},
+			expectedAnnotations: map[string]string{
+				"proxy.istio.io/config": "{\"holdApplicationUntilProxyStarts\":true}",
+			},
+		},
+		{
+			name: "filters reserved annotations too",
+			labels: map[string]string{
+				"app": "test",
+			},
+			annotations: map[string]string{
+				"custom":                              "value",
+				"pipelines.kubeflow.org/v2_component": "true",
+			},
+			expectedLabels: map[string]string{
+				"app": "test",
+			},
+			expectedAnnotations: map[string]string{
+				"custom": "value",
+			},
+		},
+	}
+
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			// Reset viper and driver config state
+			viper.Reset()
+			resetDriverConfig()
+
+			if tt.labels != nil {
+				viper.Set(DriverPodLabels, tt.labels)
+			}
+			if tt.annotations != nil {
+				viper.Set(DriverPodAnnotations, tt.annotations)
+			}
+
+			// Initialize driver config
+			require.NoError(t, InitDriverPodConfig())
+
+			// Verify labels
+			labels := GetDriverPodLabels()
+			assert.Equal(t, tt.expectedLabels, labels)
+
+			// Verify annotations
+			annotations := GetDriverPodAnnotations()
+			assert.Equal(t, tt.expectedAnnotations, annotations)
+		})
+	}
+}
+
+func TestGetDriverPodLabels(t *testing.T) {
+	tests := []struct {
+		name     string
+		config   map[string]string
+		expected map[string]string
+	}{
+		{
+			name:     "empty config",
+			config:   nil,
+			expected: nil,
+		},
+		{
+			name: "valid labels",
+			config: map[string]string{
+				"sidecar.istio.io/inject": "true",
+				"app":                     "test",
+			},
+			expected: map[string]string{
+				"sidecar.istio.io/inject": "true",
+				"app":                     "test",
+			},
+		},
+		{
+			name: "filters reserved labels",
+			config: map[string]string{
+				"sidecar.istio.io/inject":       "true",
+				"pipelines.kubeflow.org/system": "reserved",
+				"app":                           "test",
+			},
+			expected: map[string]string{
+				"sidecar.istio.io/inject": "true",
+				"app":                     "test",
+			},
+		},
+		{
+			name: "all reserved labels returns nil",
+			config: map[string]string{
+				"pipelines.kubeflow.org/system": "reserved",
+				"pipelines.kubeflow.org/task":   "reserved",
+			},
+			expected: nil,
+		},
+	}
+
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			// Reset viper and driver config state
+			viper.Reset()
+			resetDriverConfig()
+
+			if tt.config != nil {
+				viper.Set(DriverPodLabels, tt.config)
+			}
+
+			// Initialize driver config to load from Viper
+			require.NoError(t, InitDriverPodConfig())
+
+			result := GetDriverPodLabels()
+			assert.Equal(t, tt.expected, result)
+		})
+	}
+}
+
+func TestGetDriverPodAnnotations(t *testing.T) {
+	tests := []struct {
+		name     string
+		config   map[string]string
+		expected map[string]string
+	}{
+		{
+			name:     "empty config",
+			config:   nil,
+			expected: nil,
+		},
+		{
+			name: "valid annotations",
+			config: map[string]string{
+				"proxy.istio.io/config": "{\"holdApplicationUntilProxyStarts\":true}",
+				"custom":                "annotation",
+			},
+			expected: map[string]string{
+				"proxy.istio.io/config": "{\"holdApplicationUntilProxyStarts\":true}",
+				"custom":                "annotation",
+			},
+		},
+		{
+			name: "filters reserved annotation prefix",
+			config: map[string]string{
+				"proxy.istio.io/config":               "value",
+				"pipelines.kubeflow.org/v2_component": "true",
+			},
+			expected: map[string]string{
+				"proxy.istio.io/config": "value",
+			},
+		},
+	}
+
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			// Reset viper and driver config state
+			viper.Reset()
+			resetDriverConfig()
+
+			if tt.config != nil {
+				viper.Set(DriverPodAnnotations, tt.config)
+			}
+
+			// Initialize driver config to load from Viper
+			require.NoError(t, InitDriverPodConfig())
+
+			result := GetDriverPodAnnotations()
+			assert.Equal(t, tt.expected, result)
+		})
+	}
+}
+
+func TestInitDriverPodConfigFromJSONString(t *testing.T) {
+	viper.Reset()
+	resetDriverConfig()
+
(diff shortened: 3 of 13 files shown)


thc1006:
Dear @hbelmiro,
  Here's fixed pre-commit staticcheck ST1000 warning by adding package comment to driver_config.go (commit ada110bc7).

  Regarding E2E test failures: these are unrelated to this PR. The failures are caused by metadata-grpc-deployment CrashLoopBackOff, which is an intermittent upstream issue. Evidence:

  - Master branch E2E tests on 2025-11-05 (commit 36c8437dd, our base) also failed with identical metadata-grpc errors
  - Latest master (2025-11-07 20:30) E2E tests passed successfully
  - Our changes only modify driver pod labels/annotations configuration and do not touch metadata services

  Local verification completed:
  - Go unit tests: 16/16 passed
  - go vet: passed
  - go fmt: passed
  - Package builds: successful

  Now ready for review.

thc1006:
## TL;DR
  E2E failures are caused by GCR shutdown (unrelated to this PR). Code is ready for review. Fix in progress: [PR #12426](https://github.com/kubeflow/pipelines/pull/12426).

  ---

  Dear @hbelmiro and @mprahl,

  Quick update on the CI failures:

  **Root Cause**: GCR shutdown (2025-07-17) - minio image no longer accessible
  **Impact**: Only 4 minio-dependent E2E tests failed
  **Evidence**: All other tests passed ✓ (seaweedfs tests, builds, pre-commit)

  **Our PR Scope**: Driver pod labels/annotations only - no minio changes
  **Local Tests**: 16/16 unit tests ✓, linters ✓

  **Upstream Fix**: [PR #12426](https://github.com/kubeflow/pipelines/pull/12426) migrates to Docker Hub
  **Next Step**: Will rebase once #12426 merges

  The driver config implementation is complete and ready for code review. Thank you! 🙏

thc1006:
## TL;DR
CI failures are infrastructure-related (flaky test + API rate limit), not code issues. 156/158 checks passed. Kindly requesting a re-run of failed jobs.

---

Dear @hbelmiro and @mprahl,

Thank you very much for your continued support and for running `/ok-to-test`. I sincerely appreciate your patience with this PR.

I have conducted a thorough analysis of the current CI failures and would like to share my findings:

### CI Status Summary
| Result | Count |
|--------|-------|
| ✅ Passed | 156 |
| ❌ Failed | 2 |

### Analysis of Failed Jobs

**1. `check_ci_status`** - GitHub API Rate Limit
- Error: `API rate limit exceeded for installation`
- This is a transient infrastructure issue, not related to the code changes.

**2. `artifact-proxy` E2E Test** - Workflow Timeout (Flaky Test)
- Error: `timed out waiting for the condition on workflows/artifact-proxy-e2e-test-mq79d`
- **Cross-validation**: The exact same test **passed** in recent PRs:
  - `feature/pipeline-run-parallelism` (run 21088172051) ✅
  - `feature/postgres-integration` (run 21087852548) ✅
  - `dependabot/pip/backend/pyasn1-0.6.2` (run 21078854729) ✅

**Why this PR cannot cause this failure:**
The `artifact-proxy` test creates Argo Workflows directly via `kubectl`, completely bypassing the KFP API compilation path. Our changes only affect the `argocompiler` module, which is not invoked in this test flow.

### Humble Request
Would it be possible to re-run the failed jobs at your convenience? I believe they will pass on retry, as this appears to be a known flaky test pattern.

I deeply apologize for any inconvenience this may have caused. Thank you very much for your time and consideration. 🙏

Warm regards

Zach-Fong:
Hi @thc1006.
I'm a new contributor who picked up issue #12015 and found your PR already implements the solution. Rather than duplicate the effort I'd like to help get it merged.

It looks like the two checks are failing again. Would it be possible to get a /retest run?

QUICK_REFERENCE.md
```diff
@@ -0,0 +1,155 @@
...
+✅ Supports Istio sidecar injection
+✅ Applies to ALL driver pods (DAG + Container)
+
+### Files Modified:
+```
+7 files changed
++460 lines added
+-3 lines deleted
+```
+
```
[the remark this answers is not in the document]
thc1006: Resolved: QUICK_REFERENCE.md and IMPLEMENTATION_SUMMARY.md have been removed from the PR in a previous rebase.

backend/src/v2/compiler/argocompiler/dag.go
```diff
@@ -629,6 +629,25 @@ func (c *workflowCompiler) addDAGDriverTemplate() string {
...
+			if t.Metadata.Annotations == nil {
+				t.Metadata.Annotations = make(map[string]string)
+			}
+			for k, v := range c.driverPodConfig.Labels {
+				t.Metadata.Labels[k] = v
+			}
+			for k, v := range c.driverPodConfig.Annotations {
+				t.Metadata.Annotations[k] = v
+			}
+		}
```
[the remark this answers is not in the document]
thc1006: Resolved: The inline code has been refactored into a shared \ method on the \ struct. The \ check is intentionally kept to avoid unnecessary map allocation when config is empty (the common case).

backend/src/v2/compiler/argocompiler/container.go
```diff
@@ -264,6 +271,25 @@ func (c *workflowCompiler) addContainerDriverTemplate() string {
...
+			}
+			if t.Metadata.Annotations == nil {
+				t.Metadata.Annotations = make(map[string]string)
+			}
+			for k, v := range c.driverPodConfig.Labels {
+				t.Metadata.Labels[k] = v
+			}
+			for k, v := range c.driverPodConfig.Annotations {
+				t.Metadata.Annotations[k] = v
+			}
```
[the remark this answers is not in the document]
thc1006: Resolved: The inline code has been refactored into a shared \ method. The suggested code also had a bug (nil check moved inside the loop). The \ guard is kept intentionally to avoid unnecessary map allocation.

backend/src/apiserver/common/driver_config.go
```diff
@@ -0,0 +1,115 @@
...
+	return cachedDriverPodAnnotations
+}
+
+// filterReservedLabels removes labels with reserved prefix
+func filterReservedLabels(labels map[string]string) map[string]string {
+	if labels == nil {
+		return nil
+	}
+
+	filtered := make(map[string]string)
```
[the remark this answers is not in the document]
thc1006: Resolved: Map preallocation with \ was implemented in a previous commit.

backend/src/apiserver/common/driver_config.go
```diff
@@ -0,0 +1,133 @@
...
+// Labels with pipelines.kubeflow.org/ prefix are filtered out during initialization.
+// Returns nil if InitDriverPodConfig has not been called or no labels are configured.
+func GetDriverPodLabels() map[string]string {
+	driverConfigMutex.RLock()
+	defer driverConfigMutex.RUnlock()
+
+	if !driverConfigInitialized {
+		glog.Warning("Driver pod config not initialized, returning nil labels")
+		return nil
+	}
```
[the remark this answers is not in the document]
thc1006: Fixed: Removed the glog.Warning call. Now returns nil silently when uninitialized, avoiding log spam in the standalone v2 compiler.

backend/src/apiserver/common/driver_config.go
```diff
@@ -0,0 +1,133 @@
...
+// GetDriverPodAnnotations returns a copy of cached driver pod annotations from configuration.
+// Returns nil if InitDriverPodConfig has not been called or no annotations are configured.
+func GetDriverPodAnnotations() map[string]string {
+	driverConfigMutex.RLock()
+	defer driverConfigMutex.RUnlock()
+
+	if !driverConfigInitialized {
+		glog.Warning("Driver pod config not initialized, returning nil annotations")
+		return nil
+	}
```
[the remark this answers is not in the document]
thc1006: Fixed: Same as above, removed the glog.Warning call for annotations too.

backend/src/apiserver/common/driver_config.go
```diff
@@ -0,0 +1,133 @@
...
+func GetDriverPodLabels() map[string]string {
+	driverConfigMutex.RLock()
+	defer driverConfigMutex.RUnlock()
+
+	if !driverConfigInitialized {
+		glog.Warning("Driver pod config not initialized, returning nil labels")
+		return nil
+	}
+	return copyMap(cachedDriverPodLabels)
+}
```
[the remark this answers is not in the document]
thc1006: Fixed: filterReservedEntries now returns nil when the result is empty after filtering. Added docstring clarification. Behavior is now consistent with the documented contract.

backend/src/apiserver/common/driver_config_test.go
```diff
@@ -0,0 +1,348 @@
...
+			if tt.labels != nil {
+				viper.Set(DriverPodLabels, tt.labels)
+			}
+			if tt.annotations != nil {
+				viper.Set(DriverPodAnnotations, tt.annotations)
+			}
+
+			// Initialize driver config
+			InitDriverPodConfig()
+
```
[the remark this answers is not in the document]
thc1006: Fixed: Added TestInitDriverPodConfigFromJSONString test that sets Viper values as JSON strings to verify the ConfigMap integration path.

backend/src/apiserver/common/driver_config.go
```diff
@@ -0,0 +1,133 @@
...
+)
+
+const (
+	DriverPodLabels      = "DriverPodLabels"
+	DriverPodAnnotations = "DriverPodAnnotations"
+
+	// Reserved label prefix that will be filtered out to prevent
+	// overriding system labels that control workflow behavior
+	ReservedLabelPrefix = "pipelines.kubeflow.org/"
+)
```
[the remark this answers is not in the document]
thc1006: Fixed: Updated the PR description. We intentionally only filter the pipelines.kubeflow.org/ prefix (KFP system metadata). The old description incorrectly mentioned kubernetes.io/k8s.io filtering.

backend/src/v2/compiler/argocompiler/container.go
```diff
@@ -148,6 +147,20 @@ func GetPipelineRunAsUser() *int64 {
 	return &runAsUser
 }
 
+// getDriverPodConfig loads driver pod labels and annotations from Viper configuration.
+// Returns nil when no labels or annotations are configured.
```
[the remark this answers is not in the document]
thc1006: Fixed: Updated the comment to say 'returns driver pod labels and annotations from the cached configuration in the apiserver/common package'.

thc1006:
Dear @hbelmiro Can help review again! thx

thc1006:
Dear @droctothorpe, thank you for your continued contributions to the project. I hope I'm not taking too much of your time.

I've just rebased this PR onto the latest master so there are no more conflicts and all checks are passing. This implements approach (1) from the discussion on #12015, adding configuration for driver pod labels and annotations loaded at API server startup to support Istio sidecar injection on driver pods.

The feature itself is fairly contained. Config loading and caching lives in `driver_config.go` with 16 test cases, there's one init call in `main.go`, and a few lines in the argo compiler files to apply the config to driver templates. Most of the 700 line diff is tests, so the review should hopefully not be too heavy.

I would be very grateful if you could take a look whenever your schedule allows. Please do not hesitate to point out anything that needs to be changed, I will address feedback as quickly as I can.

Thank you very much, and a sincere thanks to @hbelmiro as well for generously helping keep CI running on this PR throughout.

backend/src/apiserver/common/driver_config.go
```diff
@@ -0,0 +1,135 @@
...
+func InitDriverPodConfig() {
+	driverConfigMutex.Lock()
+	defer driverConfigMutex.Unlock()
+
+	if driverConfigInitialized {
+		glog.Info("Re-initializing driver pod configuration")
+	}
+
+	// Load labels from Viper configuration
+	rawLabels := GetMapConfig(DriverPodLabels)
```
hbelmiro: `InitDriverPodConfig()` claims to "catch configuration errors early" (line 649 in `main.go`), but `GetMapConfig` delegates to `viper.GetStringMapString`, which silently returns an empty map on the **ConfigMap string path** when it cannot parse the JSON string into `map[string]string`. Note: config.json native map values are **not** affected — Viper coerces non-string values (booleans, numbers) to strings when the source is a `map[string]interface{}`. The silent failure is specific to the ConfigMap path where values arrive as JSON strings, which is the documented deployment method in both the PR's README and manifest. Verified against Viper:

| Source / value                                    | Result                                           |
|---------------------------------------------------|--------------------------------------------------|
| ConfigMap: `'{"sidecar.istio.io/inject":"true"}'` | `map[sidecar.istio.io/inject:true]`              |
| ConfigMap: `'{"sidecar.istio.io/inject":true}'`   | `map[]` — boolean instead of string              |
| ConfigMap: `'{"replicas":3}'`                     | `map[]` — number instead of string               |
| ConfigMap: `'{bad json}'`                         | `map[]` — syntax error                           |
| config.json: `{"sidecar.istio.io/inject": true}`  | `map[sidecar.istio.io/inject:true]` — coerced OK |

An operator who writes `true` (JSON boolean) instead of `"true"` (JSON string) for the Istio sidecar injection value in the ConfigMap — a natural mistake — sees `"Driver pod configuration initialized: 0 labels, 0 annotations"` at INFO level and believes nothing is configured, when the configuration is silently broken.

**Why it matters:** The entire rationale for startup-time loading (stated in code comments and the PR description) is to surface configuration problems early. Silent swallowing defeats that purpose and makes misconfiguration hard to diagnose in production.

**Suggested fix:** Change `InitDriverPodConfig` to return `error` and fail startup on bad config. `initConfig()` in `main.go` already returns `error`, so this propagates naturally. After calling `GetMapConfig`, probe the raw string with `json.Unmarshal` when the parsed map is empty. `GetMapConfig` returns an empty non-nil map (not `nil`) for unparseable input, so a nil check is insufficient. A simple `len() == 0` heuristic would false-positive on valid `"{}"`, so an explicit parse check is more precise:

```go
func InitDriverPodConfig() error {
	driverConfigMutex.Lock()
	defer driverConfigMutex.Unlock()

	// ... existing re-init log ...

	rawLabels := GetMapConfig(DriverPodLabels)
	if err := validateMapConfig(DriverPodLabels, rawLabels); err != nil {
		return err
	}
	cachedDriverPodLabels = filterReservedEntries(rawLabels)

	rawAnnotations := GetMapConfig(DriverPodAnnotations)
	if err := validateMapConfig(DriverPodAnnotations, rawAnnotations); err != nil {
		return err
	}
	cachedDriverPodAnnotations = filterReservedEntries(rawAnnotations)

	driverConfigInitialized = true
	// ... existing logging ...
	return nil
}

func validateMapConfig(name string, parsed map[string]string) error {
	if len(parsed) > 0 || !viper.IsSet(name) {
		return nil
	}
	raw := viper.GetString(name)
	if raw == "" {
		return nil
	}
	var probe map[string]interface{}
	if err := json.Unmarshal([]byte(raw), &probe); err != nil {
		return fmt.Errorf("%s could not be parsed as JSON — check syntax: %w (raw value: %q)", name, err, raw)
	}
	if len(probe) > 0 {
		return fmt.Errorf("%s parsed as JSON but all values were dropped — ensure every value is a string, not a boolean or number (raw value: %q)", name, raw)
	}
	return nil
}
```

Update the call site in `main.go`:

```go
if err := common.InitDriverPodConfig(); err != nil {
	return fmt.Errorf("driver pod config: %w", err)
}
```

backend/src/apiserver/common/driver_config_test.go
hbelmiro, line 1: No test covers the ConfigMap failure paths identified in the previous comment. The test suite validates empty config, valid maps, valid JSON strings, reserved prefix filtering, and copy isolation — but never asserts that `InitDriverPodConfig` returns an error when Viper receives a non-string JSON value (e.g., boolean `true` instead of string `"true"`) or syntactically invalid JSON via the ConfigMap string path.

**Suggested fix:** Add test cases for both failure modes, asserting that startup-time validation surfaces the error:

```go
func TestInitDriverPodConfig_NonStringJSONValues(t *testing.T) {
	viper.Reset()
	resetDriverConfig()

	// Boolean instead of string — the most realistic ConfigMap typo
	viper.Set(DriverPodLabels, `{"sidecar.istio.io/inject":true}`)

	err := InitDriverPodConfig()

	require.Error(t, err, "non-string JSON values via ConfigMap should fail at startup")
	assert.Contains(t, err.Error(), "DriverPodLabels")
}

func TestInitDriverPodConfig_MalformedJSON(t *testing.T) {
	viper.Reset()
	resetDriverConfig()

	viper.Set(DriverPodLabels, "{bad json}")

	err := InitDriverPodConfig()

	require.Error(t, err, "malformed JSON should fail at startup")
	assert.Contains(t, err.Error(), "DriverPodLabels")
}
```

These verify that operators see a clear startup failure instead of a silent no-op.

backend/src/v2/compiler/argocompiler/container.go
```diff
@@ -275,6 +289,9 @@ func (c *workflowCompiler) addContainerDriverTemplate() string {
 		},
 	}
 	applySecurityContextToTemplate(template)
+
+	c.driverPodConfig.applyToTemplate(template)
```
hbelmiro: Tests cover `applyToTemplate` and `getDriverPodConfig` individually, but no test verifies the full compilation path: `Compile()` → `addContainerDriverTemplate()` / `addDAGDriverTemplate()` producing a workflow with the expected labels and annotations on driver templates. The existing `TestContainerDriverTemplate_IncludesKFPPodNameEnv` test (container_test.go:73) demonstrates exactly how to test this pattern — create a `workflowCompiler` with a `driverPodConfig`, call `addContainerDriverTemplate()`, and assert the resulting template's metadata.

**Why it matters:** Unit tests for `applyToTemplate` prove the method works in isolation, but don't guarantee the integration point in `addContainerDriverTemplate` and `addDAGDriverTemplate` actually calls it at the right time (after template creation, before template storage). A refactor that reorders those calls could silently break the feature.

**Suggested fix:** Add a test similar to `TestContainerDriverTemplate_IncludesKFPPodNameEnv` that sets `driverPodConfig` on the compiler and verifies labels appear in the generated template:

```go
func TestContainerDriverTemplate_IncludesDriverPodConfig(t *testing.T) {
	proxy.InitializeConfigWithEmptyForTests()
	c := &workflowCompiler{
		templates: make(map[string]*wfapi.Template),
		wf:        &wfapi.Workflow{Spec: wfapi.WorkflowSpec{Templates: []wfapi.Template{}}},
		spec:      &pipelinespec.PipelineSpec{PipelineInfo: &pipelinespec.PipelineInfo{Name: "test"}},
		job:       &pipelinespec.PipelineJob{},
		driverPodConfig: &driverPodConfig{
			Labels:      map[string]string{"sidecar.istio.io/inject": "true"},
			Annotations: map[string]string{"proxy.istio.io/config": "hold"},
		},
	}
	name := c.addContainerDriverTemplate()
	tmpl := c.templates[name]
	require.NotNil(t, tmpl)
	assert.Equal(t, "true", tmpl.Metadata.Labels["sidecar.istio.io/inject"])
	assert.Equal(t, "hold", tmpl.Metadata.Annotations["proxy.istio.io/config"])
}
```

backend/src/apiserver/common/driver_config.go
```diff
@@ -0,0 +1,135 @@
...
+const (
+	DriverPodLabels      = "DriverPodLabels"
+	DriverPodAnnotations = "DriverPodAnnotations"
+
+	// Reserved label prefix that will be filtered out to prevent
+	// overriding system labels that control workflow behavior
+	ReservedLabelPrefix = "pipelines.kubeflow.org/"
+)
+
+var (
```
hbelmiro, line 47: The cached driver pod configuration is spread across three independent package-level variables (`cachedDriverPodLabels`, `cachedDriverPodAnnotations`, `driverConfigInitialized`) that must stay in sync but are mutated one at a time. This makes it easy to leave the state half-updated — e.g., labels written but annotations not yet, or `driverConfigInitialized` set before both maps are ready. The `InitDriverPodConfig` function currently doesn't return errors, but if error handling is added (see Comment 1), early returns would leave these variables inconsistent.

**Suggested fix:** Group the cached state into a struct with a constructor that either returns a fully valid object or an error. The package-level state becomes a single pointer swapped atomically:

```go
type driverPodConfigCache struct {
	labels      map[string]string
	annotations map[string]string
}

var (
	cachedDriverPodConfig *driverPodConfigCache
	driverConfigMutex     sync.RWMutex
)

func InitDriverPodConfig() error {
	// Build and validate before taking the lock for the swap.
	cfg, err := newDriverPodConfigCache()
	if err != nil {
		return err
	}

	driverConfigMutex.Lock()
	defer driverConfigMutex.Unlock()
	cachedDriverPodConfig = cfg
	return nil
}

func newDriverPodConfigCache() (*driverPodConfigCache, error) {
	rawLabels := GetMapConfig(DriverPodLabels)
	if err := validateMapConfig(DriverPodLabels, rawLabels); err != nil {
		return nil, err
	}
	rawAnnotations := GetMapConfig(DriverPodAnnotations)
	if err := validateMapConfig(DriverPodAnnotations, rawAnnotations); err != nil {
		return nil, err
	}
	return &driverPodConfigCache{
		labels:      filterReservedEntries(rawLabels),
		annotations: filterReservedEntries(rawAnnotations),
	}, nil
}

func GetDriverPodLabels() map[string]string {
	driverConfigMutex.RLock()
	defer driverConfigMutex.RUnlock()
	if cachedDriverPodConfig == nil {
		return nil
	}
	return copyMap(cachedDriverPodConfig.labels)
}
```

This eliminates the `driverConfigInitialized` flag (nil check on the struct replaces it), guarantees all-or-nothing state transitions, and keeps validation/construction separate from the cache swap.

backend/src/v2/compiler/argocompiler/argo.go
```diff
@@ -318,6 +319,41 @@ type workflowCompiler struct {
 	defaultRunAsUser     *int64
 	defaultRunAsGroup    *int64
 	defaultRunAsNonRoot  *bool
+	driverPodConfig      *driverPodConfig
+}
+
+// driverPodConfig holds labels and annotations to be applied to driver pods
+type driverPodConfig struct {
```
hbelmiro: Once you create a struct as suggested in the previous comment, this one becomes duplicated.

backend/src/apiserver/common/driver_config.go
thc1006: Thank you so much for catching this, and for the clear table showing how Viper treats each value. You are right. I changed `InitDriverPodConfig` to return an error, and `initConfig` in `main.go` now passes it up, so the server refuses to start on a broken configuration instead of coming up silently empty. I added a helper `validateMapConfig` that inspects the raw value again with `json.Unmarshal` whenever `GetStringMapString` returns an empty map, which is exactly the approach you proposed. I also reproduced the behaviour locally to be sure: a boolean value, a numeric value, and invalid JSON all collapse to an empty map, while a valid empty object is treated as a genuine empty configuration rather than an error. I really appreciate the care you put into writing this up.

backend/src/apiserver/common/driver_config_test.go
thc1006: Good call, thank you. I added two tests for the failure paths you described. `TestInitDriverPodConfig_NonStringJSONValues` covers a boolean where a string is expected, and `TestInitDriverPodConfig_MalformedJSON` covers invalid syntax. Both assert that the returned error names the offending key so an operator can see what to correct. I also added `TestInitDriverPodConfig_EmptyJSONObjectIsValid` to keep a guard proving that a valid empty object does not fail startup.

backend/src/v2/compiler/argocompiler/container.go
thc1006: Thank you for pointing me to `TestContainerDriverTemplate_IncludesKFPPodNameEnv` as a model, that made it straightforward. I added `TestContainerDriverTemplate_IncludesDriverPodConfig`, and a matching `TestDAGDriverTemplate_IncludesDriverPodConfig` for the DAG driver, since the same reasoning applies to both templates. Each test builds a `workflowCompiler` with a configuration set, calls the real `addContainerDriverTemplate` or `addDAGDriverTemplate`, and asserts that the labels and annotations end up on the generated template metadata. A future change to the order of those calls would now be caught.

backend/src/apiserver/common/driver_config.go
thc1006: Agreed, and thank you for the suggestion. This is much cleaner. I grouped the cached state into a single `DriverPodConfig` value held behind one pointer guarded by the mutex. A new helper `newDriverPodConfig` reads and validates everything first, and `InitDriverPodConfig` then swaps the pointer in one step under the lock, so a reader never sees a configuration that is only partly populated. The `driverConfigInitialized` flag is gone, since a nil pointer already tells us the configuration has not been loaded yet.

backend/src/v2/compiler/argocompiler/argo.go
thc1006: Thank you, you are right that it would otherwise be duplicated. I removed the `driverPodConfig` struct from the argo compiler. The compiler now uses the exported `common.DriverPodConfig` directly, and the old `applyToTemplate` method became a small standalone function `applyDriverPodConfig`, so only one type describes this configuration now.

thc1006:
Thank you very much for the thorough and thoughtful review. I have addressed all five of your comments and rebased the branch onto the latest master, which also clears the earlier merge conflict. Locally the `common` and `argocompiler` packages pass build, vet and their tests including the new cases, and I confirmed that the `apiserver` package still compiles with CGO enabled. Please let me know if you would like any further changes, and thank you again for your time.

frontend/server/package-lock.json
hbelmiro, line 1: These changes seem unrelated.

test/frontend-integration-test/package-lock.json
hbelmiro, line 1: These changes seem unrelated.

frontend/server/package-lock.json
thc1006: Thank you, you are right. This was an unrelated npm lockfile change that slipped into the commit by accident. I have removed it and rebased onto the latest master, so this file is no longer part of the PR.

test/frontend-integration-test/package-lock.json
thc1006: Agreed, thank you. Same here, this lockfile change was unrelated and got committed by mistake. It is now reverted, so the diff only contains the driver pod configuration changes again.

thc1006:
The only failing job is the MLflow scenario with cacheEnabled=false and artifactStorage=file. The other three MLflow matrix variants passed on this exact commit, so this looks like a flaky issue in the MLflow operator deploy and port forward rather than anything in this change. This change only adds labels and annotations to driver pods when an admin configures them, and these tests leave that configuration empty, so the generated workflows are unchanged. Kindly requesting a retest when convenient, and thank you.

/retest

hbelmiro (approved):
/lgtm
/approve

cc @juliusvonkohout for approval

thc1006:
Thank you so much @hbelmiro for the review, the lgtm, the approve, and for cc'ing @juliusvonkohout. I really appreciate you helping move this along.

The only remaining step is an approval from the manifests/kustomize OWNERS, since the PR adds two ConfigMap keys there. @juliusvonkohout or @chensun, whenever you have a moment, I would be very grateful for a look. CI is green and the branch is currently free of conflicts, so I am hoping to land it before master moves again. Thank you all very much.

chensun (approved):

hbelmiro:
@chensun the GitHub approval no longer adds the `approved` label.
You need to comment `/approve` now.

thc1006:
Thank you so much @chensun for reviewing and approving, and @hbelmiro for helping route this. I understand the Prow approved label now needs a /approve comment rather than the review button, so whenever you have a moment @chensun I would be grateful for that. CI is green and the branch is still free of conflicts. Thank you both very much.

thc1006:
Before this merged I wanted to be sure the ConfigMap path actually worked end to end, and I am glad I checked, because it did not. I would rather hold the PR up now than let operators discover it afterwards.

@hbelmiro, it was your earlier point about the ConfigMap string path being the documented way to deploy this that sent me back to look at how those keys reach the API server. They never reached it. The API server only receives configuration through the environment variables listed in its Deployment, and DriverPodLabels and DriverPodAnnotations were not among them, so anyone following the README would have set the keys and seen nothing happen.

This revision wires both keys into the API server Deployment, and into the PostgreSQL install as well, whose apiserver patch replaces the container environment list and therefore needs its own entries. I also added validation so that a label or annotation Kubernetes would reject now fails at startup with a clear message rather than surfacing later as a rejected driver pod. Annotation values stay free form, so the inline Istio proxy config keeps working.

I checked this on a real Kubernetes cluster rather than trusting the manifests alone. With the rendered Deployment applied and the ConfigMap holding the Istio values, the values came through into the API server container environment intact, escaped inline JSON and all. I also rendered every overlay to confirm the wiring is present in each, including multi user and PostgreSQL.

I am sorry that this costs the lgtm and another round of CI after you had both already signed off. @hbelmiro, whenever your schedule allows, I would be grateful for another look. @chensun, the manifests will still need your /approve when you have a moment. Thank you both for your patience.

backend/README.md
```diff
@@ -62,6 +62,35 @@ Run
 docker build . -f backend/Dockerfile.conformance -t <tag>
 ```
 
+## API Server Configuration
+
+### Driver Pod Labels and Annotations
+
+The API server supports configuring custom labels and annotations for driver pods through the config file or the Kubernetes ConfigMap. This is useful for integration with service mesh (Istio), monitoring systems, or other infrastructure requirements.
```
juliusvonkohout:
```suggestion
The API server supports configuring custom labels and annotations for driver pods through the configuration file or the Kubernetes ConfigMap. This is useful for integration with service mesh (Istio), monitoring systems, or other infrastructure requirements.
```

manifests/kustomize/base/installs/generic/postgres/pipeline-install-config.yaml
juliusvonkohout, line 1: We really need to get rid of this duplicate file in a follow up PR @kaikaila . manifests/kustomize/base/installs/generic/pipeline-install-config.yaml shoud be the single source of truth

manifests/kustomize/base/postgresql/pipeline/ml-pipeline-apiserver-deployment-patch.yaml
```diff
@@ -65,6 +65,20 @@ spec:
               name: pipeline-install-config
               key: V1_ALLOWED_NAMESPACES
               optional: true
+        # Driver pod labels and annotations. This patch replaces the container env list,
+        # so these must be repeated here to reach the API server on PostgreSQL installs.
+        - name: DRIVERPODLABELS
```
juliusvonkohout:
```suggestion
        - name: DRIVER_POD_LABELS
```

manifests/kustomize/base/postgresql/pipeline/ml-pipeline-apiserver-deployment-patch.yaml
```diff
@@ -65,6 +65,20 @@ spec:
...
+          valueFrom:
+            configMapKeyRef:
+              name: pipeline-install-config
+              key: DriverPodLabels
```
juliusvonkohout:
```suggestion
              key: DRIVER_POD_LABELS
```

manifests/kustomize/base/postgresql/pipeline/ml-pipeline-apiserver-deployment-patch.yaml
```diff
@@ -65,6 +65,20 @@ spec:
...
+              optional: true
+        - name: DRIVERPODANNOTATIONS
+          valueFrom:
+            configMapKeyRef:
+              name: pipeline-install-config
+              key: DriverPodAnnotations
```
juliusvonkohout:
```suggestion
              key: DRIVER_POD_ANNOTATIONS
```

manifests/kustomize/base/postgresql/pipeline/ml-pipeline-apiserver-deployment-patch.yaml
```diff
@@ -65,6 +65,20 @@ spec:
...
               optional: true
+        # Driver pod labels and annotations. This patch replaces the container env list,
+        # so these must be repeated here to reach the API server on PostgreSQL installs.
+        - name: DRIVERPODLABELS
+          valueFrom:
+            configMapKeyRef:
+              name: pipeline-install-config
+              key: DriverPodLabels
+              optional: true
+        - name: DRIVERPODANNOTATIONS
```
juliusvonkohout:
```suggestion
        - name: DRIVER_POD_ANNOTATIONS
```

manifests/kustomize/base/installs/generic/postgres/pipeline-install-config.yaml
```diff
@@ -125,3 +125,9 @@ data:
   ## USERID_PREFIX is stripped from the header value before it is used as the user id.
   USERID_HEADER: "kubeflow-userid"
   USERID_PREFIX: ""
+  ## Custom labels and annotations for driver pods (JSON format).
+  ## Useful for enabling Istio sidecar injection or other infrastructure integration.
+  ## Keys with the prefix "pipelines.kubeflow.org/" are reserved and will be filtered out.
+  ## Example: '{"sidecar.istio.io/inject":"true"}'
+  DriverPodLabels: ""
+  DriverPodAnnotations: ""
```
juliusvonkohout:
```suggestion
  DRIVER_POD_LABELS: ""
  DRIVER_POD_ANNOTATIONS: ""
```

juliusvonkohout (changes requested):

thc1006:
TLDR: every suggestion from @juliusvonkohout is applied, the keys are now DRIVER_POD_LABELS and DRIVER_POD_ANNOTATIONS everywhere, and two real defects that Copilot found are fixed, one of which would have silently disabled Istio injection.

@juliusvonkohout, thank you for taking the time to go through this, and for the concrete suggestions. I have applied all of them. The rename is now consistent across every place the keys appear, not only in the files you marked: the default install config, the PostgreSQL install config, the API server Deployment, the PostgreSQL apiserver patch, the Go configuration keys, and both README examples. The ConfigMap key, the environment variable, the Viper key and the config.json key are now one single name, which I believe is what you were after. Your wording fix in the README is in as well.

On the duplicate install config file, I have left it untouched and kept your note for the follow up PR you mentioned to @kaikaila, since it reaches well beyond this change.

Copilot raised three points and two of them were real, so I am glad it looked. The first one matters to operators. Viper does not reject a JSON null the way it rejects a boolean or a number. It quietly turns null into an empty string, so someone writing null for the Istio injection flag would have ended up with an empty label and no warning at all, believing injection was on. I reproduced that in a test before changing anything. Every value must now be a JSON string, and a null, a boolean, a number or a nested object all fail startup with a message naming the offending key. The second point was that GetDriverPodConfig took the read lock twice, which could hand back labels from one version of the cache and annotations from another, defeating the atomic snapshot that @hbelmiro asked for in his earlier review. Both maps are now copied under one lock.

Copilot's third point was that the compiler reaches into API server state. To be precise, Compile already read API server configuration before this PR, for the service account name, the CA bundle and the metadata TLS setting, so the coupling is not new. The suggestion is still the better shape, so the driver pod configuration now travels through argocompiler.Options in the same way as DefaultRunAsUser and DefaultHostUsers, and the standalone compiler receives nil explicitly rather than by accident.

@juliusvonkohout, whenever your schedule allows, I would be grateful if you could take another look. @hbelmiro, the lgtm was dropped by these commits, so it would need your /lgtm again when you have a moment. Thank you both.

thc1006:
The two MLflow checks that are red are both artifactStorage=file variants, and the same ones are currently failing on master as well (four of them on 89d2530e8) with the identical SeaweedFS socket errors and read timeouts, so they do not appear to be caused by this change. I am of course happy to look again if you see it differently.

hbelmiro (approved):
/lgtm
/approve

PR #12420, opened 2025-11-07
fix(CI/SDK): Fix tests for pipelines with optional input parameters and an SDK syntax error

mprahl:

It would provide random values that would cause test_data/sdk_compiled_pipelines/valid/critical/missing_kubernetes_optional_inputs.py to fail because they were invalid tolerations.

**Checklist:**

### DIFF
--- a/sdk/python/kfp/dsl/types/artifact_types.py
+++ b/sdk/python/kfp/dsl/types/artifact_types.py
@@ -94,7 +94,7 @@ class Artifact:
         self._set_custom_path(path)
 
     def _get_path(self) -> Optional[str]:
-        if self.custom_path is not '':
+        if self.custom_path:
             return self._get_custom_path()
         if self.uri.startswith(RemotePrefix.GCS.value):
             return _GCS_LOCAL_MOUNT_PREFIX + self.uri[len(RemotePrefix.GCS.value


sdk/python/kfp/dsl/types/artifact_types.py
```diff
@@ -94,7 +94,7 @@ def path(self, path: str) -> None:
         self._set_custom_path(path)
 
     def _get_path(self) -> Optional[str]:
-        if self.custom_path is not '':
+        if self.custom_path:
```
nsingla, line 97: this will only check if custom_path is None or not, is that intentional to not check for empty string value?

mprahl: An empty string in Python is falsy:
```python
>>> if "":
...     print("true")
...     
>>> if not "":
...     print("true")
...     
true
```

alyssacgoins: More out of curiosity - is this correction related to the backend/test/testutil/pipeline_run_utils.go correction above, or are they included in the same PR for efficiency?

mprahl: @alyssacgoins for efficiency. I couldn't compile a pipeline without this change locally.

backend/test/testutil/pipeline_run_utils.go
```diff
@@ -125,7 +125,7 @@ func GetPipelineRunTimeInputs(pipelineSpecFile string) map[string]interface{} {
 				if parameterSpec.DefaultValue != nil {
 					defaultValExists = true
 				}
-				if !defaultValExists || !parameterSpec.IsOptional {
+				if !defaultValExists && !parameterSpec.IsOptional {
```
nsingla, line 128: i think this makes sense, 1 should not exist without the other, but I believe I have a case where default value is None, and the parameter is Optional, in which case this condition will be False, and no value will be set for the parameter. And fwir, I saw an error while trying to create a pipeline run in this case, which may have gotten resolved now :thinking:

mprahl: Yes, the intention here is only assign an input parameter value if it's not optional.

nsingla: Just out of curiosity, what was the toleration that it was setting? was it a random string value? and its expected to be a V1Toleration object?

mprahl: It was setting something like `["asdasdasd"]`.

nsingla: hmmm... so a random string list, because that's the type defined in the pipeline spec

nsingla (approved):
lgtm

PR #12428, opened 2025-11-10
fix(backend): Allow pod namespace configuration on env.go metadata-grpc service address

alyssacgoins:
Update `backend/src/v2/metadata/env.go` `DefaultConfig()` to return a metadata service config containing the pod-specific namespace. This also includes a change to `backend/src/apiserver/common/config.go` `GetPodNamespace()` that sets the default pod namespace to `"kubeflow"`

**Checklist:**

### DIFF
--- a/backend/metadata_writer/src/metadata_helpers.py
+++ b/backend/metadata_writer/src/metadata_helpers.py
@@ -32,9 +32,8 @@ def value_to_mlmd_value(value) -> metadata_store_pb2.Value:
 
 
 def connect_to_mlmd() -> metadata_store.MetadataStore:
-    metadata_service_host = os.environ.get("METADATA_GRPC_SERVICE_SERVICE_HOST", "metadata-grpc-service.kubeflow")
-    metadata_service_port = int(os.environ.get("METADATA_GRPC_SERVICE_SERVICE_PORT", 8080))
-    metadata_service_host = "metadata-grpc-service.kubeflow"
+    pod_namespace = os.environ.get("POD_NAMESPACE", "kubeflow")
+    metadata_service_host = "metadata-grpc-service." + pod_namespace + ".svc.cluster.local"
     metadata_service_port = 8080
 
     mlmd_connection_config = metadata_store_pb2.MetadataStoreClientConfig(
--- a/backend/src/apiserver/common/config.go
+++ b/backend/src/apiserver/common/config.go
@@ -109,7 +109,7 @@ func IsMultiUserSharedReadMode() bool {
 }
 
 func GetPodNamespace() string {
-	return GetStringConfig(PodNamespace)
+	return GetStringConfigWithDefault(PodNamespace, DefaultPodNamespace)
 }
 
 func GetBoolFromStringWithDefault(value string, defaultValue bool) bool {
--- a/backend/src/v2/driver/driver.go
+++ b/backend/src/v2/driver/driver.go
@@ -235,6 +235,8 @@ func initPodSpecPatch(
 	mlPipelineTLSEnabled bool,
 	metadataTLSEnabled bool,
 	caCertPath string,
+	mlmdServerAddress string,
+	mlmdServerPort string,
 ) (*k8score.PodSpec, error) {
 	executorInputJSON, err := protojson.Marshal(executorInput)
 	if err != nil {
@@ -275,10 +277,8 @@ func initPodSpecPatch(
 		fmt.Sprintf("$(%s)", component.EnvPodName),
 		"--pod_uid",
 		fmt.Sprintf("$(%s)", component.EnvPodUID),
-		"--mlmd_server_address",
-		fmt.Sprintf("$(%s)", component.EnvMetadataHost),
-		"--mlmd_server_port",
-		fmt.Sprintf("$(%s)", component.EnvMetadataPort),
+		"--mlmd_server_address", mlmdServerAddress,
+		"--mlmd_server_port", mlmdServerPort,
 		"--publish_logs", publishLogs,
 	}
 	if mlPipelineTLSEnabled {
--- a/backend/src/v2/metadata/env.go
+++ b/backend/src/v2/metadata/env.go
@@ -1,8 +1,10 @@
 package metadata
 
+import "github.com/kubeflow/pipelines/backend/src/apiserver/common"
+
 const (
-	metadataGrpcServiceAddress = "metadata-grpc-service.kubeflow"
-	metadataGrpcServicePort    = "8080"
+	metadataGrpcServiceName = "metadata-grpc-service"
+	metadataGrpcServicePort = "8080"
 )
 
 type ServerConfig struct {
@@ -12,7 +14,7 @@ type ServerConfig struct {
 
 func DefaultConfig() *ServerConfig {
 	return &ServerConfig{
-		Address: metadataGrpcServiceAddress,
+		Address: metadataGrpcServiceName + "." + common.GetPodNamespace() + ".svc.cluster.local",
 		Port:    metadataGrpcServicePort,
 	}
 }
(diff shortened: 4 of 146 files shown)


backend/src/apiserver/common/config.go
```diff
@@ -36,6 +36,7 @@ const (
 	CaBundleSecretName                      string = "CABUNDLE_SECRET_NAME"
 	RequireNamespaceForPipelines            string = "REQUIRE_NAMESPACE_FOR_PIPELINES"
 	CompiledPipelineSpecPatch               string = "COMPILED_PIPELINE_SPEC_PATCH"
+	DefaultPodNamespace                     string = "kubeflow"
```
gmfrasca: Just for consistency, I think this should go in `backend/src/apiserver/common/const.go`

gmfrasca (commented):
/lgtm

backend/metadata_writer/src/metadata_helpers.py
```diff
@@ -431,4 +430,6 @@ def isIPv6(ip: str) -> bool:
         print('Error: Exception:{}'.format(str(e)), file=sys.stderr)
         sys.stderr.flush()
 
-
+# Returns the default_value if the environment variable is not found.
+def get_env_var_with_default(config_name: str, default_value: str) -> str:
```
gmfrasca: this is essentially just renaming os.getenv, so we should probably just use that directly rather than create a new function

mprahl: Agreed

backend/src/v2/metadata/env.go
```diff
@@ -12,7 +14,7 @@ type ServerConfig struct {
 
 func DefaultConfig() *ServerConfig {
 	return &ServerConfig{
-		Address: metadataGrpcServiceAddress,
+		Address: metadataGrpcServiceName + "." + common.GetPodNamespace() + ".svc.cluster.local",
```
mprahl, line 17: Please update the certificate subject names such as manifests/kustomize/env/cert-manager/base-tls-certs/kfp-api-cert.yaml to include the FQDN.

mprahl, line 17: This is called by the driver right? In that case, the pod namespace would be incorrect in multi-user mode as pipelines run in namespaces outside of `kubeflow`.

I would opt for the API server to derive the value and pass it as an argument to the driver and launcher with `--mlmd_server_address` and `--mlmd_server_port` since the pod namespace of the API server is where MLMD is deployed.

backend/metadata_writer/src/metadata_helpers.py
```diff
@@ -32,9 +32,8 @@ def value_to_mlmd_value(value) -> metadata_store_pb2.Value:
 
 
 def connect_to_mlmd() -> metadata_store.MetadataStore:
-    metadata_service_host = os.environ.get("METADATA_GRPC_SERVICE_SERVICE_HOST", "metadata-grpc-service.kubeflow")
-    metadata_service_port = int(os.environ.get("METADATA_GRPC_SERVICE_SERVICE_PORT", 8080))
-    metadata_service_host = "metadata-grpc-service.kubeflow"
+    pod_namespace = get_env_var_with_default("POD_NAMESPACE", "kubeflow")
```
mprahl: Could you check to see if metadawriter runs in the user's namespace in the multi-user mode or in the kubeflow namespace? If it's the former, then the POD_NAMESPACE will be incorrect similar to my comment on driver.

backend/metadata_writer/src/metadata_helpers.py
alyssacgoins: Updated

backend/src/v2/metadata/env.go
alyssacgoins: Updated

alyssacgoins: Updated - refactored so that only the argo compiler calls this method. No longer called by driver - `mlmd_server_address` and `mlmd_server_port` are passed into the driver and launcher now

backend/metadata_writer/src/metadata_helpers.py
```diff
@@ -32,10 +32,10 @@ def value_to_mlmd_value(value) -> metadata_store_pb2.Value:
 
 
 def connect_to_mlmd() -> metadata_store.MetadataStore:
+    pod_namespace = os.environ.get("POD_NAMESPACE", "kubeflow")
     metadata_service_host = os.environ.get("METADATA_GRPC_SERVICE_SERVICE_HOST", "metadata-grpc-service.kubeflow")
     metadata_service_port = int(os.environ.get("METADATA_GRPC_SERVICE_SERVICE_PORT", 8080))
-    metadata_service_host = "metadata-grpc-service.kubeflow"
-    metadata_service_port = 8080
+    metadata_service_host = "metadata-grpc-service." + pod_namespace + ".svc.cluster.local"
```
mprahl, line 36: This seems to ignore metadata_service_host's previous assignment. Should we delete that previous assignment?

backend/src/v2/driver/driver.go
```diff
@@ -275,10 +277,8 @@ func initPodSpecPatch(
 		fmt.Sprintf("$(%s)", component.EnvPodName),
 		"--pod_uid",
 		fmt.Sprintf("$(%s)", component.EnvPodUID),
-		"--mlmd_server_address",
-		fmt.Sprintf("$(%s)", component.EnvMetadataHost),
-		"--mlmd_server_port",
-		fmt.Sprintf("$(%s)", component.EnvMetadataPort),
+		"--mlmd_server_address", mlmdServerAddress,
```
mprahl, line 280: We also need this change in `addImporterTemplate`.

backend/metadata_writer/src/metadata_helpers.py
alyssacgoins: Pre-upstream TLS implementation, metadata-writer hard-coded the service host as `"metadata-grpc-service.kubeflow"`, (https://github.com/kubeflow/pipelines/blob/e31038da5f124e309f5e71193bce0c0959acb7e0/backend/metadata_writer/src/metadata_helpers.py) and this env var is only set in the TLS manifests - so it looks like metadata-writer runs in the kubeflow namespace only

backend/src/v2/driver/driver.go
mprahl, line 280: This will also need the equivalent change in `addContainerExecutorTemplate`. Please scan for other areas.

backend/metadata_writer/src/metadata_helpers.py
alyssacgoins: Yep - updated now to revert to https://github.com/kubeflow/pipelines/blob/e31038da5f124e309f5e71193bce0c0959acb7e0/backend/metadata_writer/src/metadata_helpers.py implementation (pre-TLS implementation) but with a configurable pod namespace

backend/src/v2/driver/driver.go
alyssacgoins: Because I added the mlmd address/host in `driver.initPodSpecpatch()`, those arguments are applied to the container executor pod - but does it make more sense to add those arguments via `addContainerExecutorTemplate()` instead?

alyssacgoins: Updated

mprahl: Yeah, let's just let the API server own that to make it consistent.

alyssacgoins: Sounds good - updated. Removed these two flags from `initPodSpecPatch()`, and added them to `addContainerExecutorTemplate()`

mprahl (approved):
/approve
/lgtm

mprahl (approved):
/approve
/lgtm

PR #12445, opened 2025-11-14
test(backend): Add integration tests for artifact reading and large artifacts handling

hbelmiro:
This PR adds a new integration test that uploads and runs a pipeline which generates a large binary artifact, then exercises the artifact read endpoint to ensure it correctly returns the artifact data. The test validates both the response metadata (HTTP headers, status codes) and payload format (JSON with base64‑encoded, gzip‑compressed data) to catch regressions in large artifact handling.

**Checklist:**

### DIFF
--- a/backend/test/integration/artifact_api_test.go
+++ b/backend/test/integration/artifact_api_test.go
@@ -1,0 +1,300 @@
+// Copyright 2025 The Kubeflow Authors
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+// https://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package integration
+
+import (
+	"encoding/base64"
+	"encoding/json"
+	"fmt"
+	"io"
+	"net/http"
+	"regexp"
+	"testing"
+	"time"
+
+	"github.com/golang/glog"
+	params "github.com/kubeflow/pipelines/backend/api/v1beta1/go_http_client/experiment_client/experiment_service"
+	"github.com/kubeflow/pipelines/backend/api/v1beta1/go_http_client/experiment_model"
+	pipelineParams "github.com/kubeflow/pipelines/backend/api/v1beta1/go_http_client/pipeline_client/pipeline_service"
+	uploadParams "github.com/kubeflow/pipelines/backend/api/v1beta1/go_http_client/pipeline_upload_client/pipeline_upload_service"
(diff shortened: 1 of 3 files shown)


hbelmiro:
@HumairAK we can use this test to ensure that [Open
Implemented HTTP Artifact Streaming to Prevent OOM Errors with Large Files #12394](https://github.com/kubeflow/pipelines/pull/12394) won't break compatibility.
We can merge this PR prior to https://github.com/kubeflow/pipelines/pull/12394 or just ignore it, since https://github.com/kubeflow/pipelines/pull/12394 comprises the same test.

/assign @HumairAK

backend/test/integration/artifact_api_test.go
```diff
@@ -0,0 +1,300 @@
...
+		time.Sleep(2 * time.Second)
+	}
+
+	require.Fail(t, "Run did not complete within %v", maxWaitTime)
+}
+
+func (s *ArtifactAPITest) extractWorkflowNodeID(runID string) string {
+	t := s.T()
+
+	resp, err := http.Get(fmt.Sprintf("http://localhost:8888/apis/v1beta1/runs/%s", runID))
```
HumairAK, line 216: can you re-use the api server host info  (as well as the scheme http/https) from `backend/test/config/flags.go` instead of hardcoding it? This applies to everywhere, this applies to any place you are hardcoding localhost:8888

backend/test/integration/artifact_api_test.go
```diff
@@ -0,0 +1,300 @@
...
+	require.NoError(t, err)
```
HumairAK, line 217: handle error (same with other resp.Body.Close() calls)

backend/test/integration/artifact_api_test.go
```diff
@@ -0,0 +1,300 @@
...
+func (s *ArtifactAPITest) TestV1PipelineArtifactRead() {
+	runID, experimentID, pipelineID := s.runPipeline()
+
+	defer func() {
+		s.deleteRun(runID)
+		s.deleteExperiment(experimentID)
+		s.deleteAllPipelineVersions(pipelineID)
+	}()
+
+	s.waitForRunCompletion(runID)
```
HumairAK, line 116: Sorry I should have been more clear, what I was suggesting we test is the endpoint allows us to continue testing that we can download artifacts - and not that we can download large files. My concern around download large files is that it can choke the CI and cause flaky behavior. 

My preference is to just write a simple pipeline that will output an artifact (using @component instead of `ContainerOp`). The artifact shouldn't be large, we just need to confirm we can continue to download the artifact via this endpoint, and then verify the data is what we expect. This helps us ensure your PR doesn't break the current api behavior. 

Verifying large downloads in master doesn't help us verify the issue once it's fixed, since to do that you would need to emulate the DOS for api server, and this is more of a perf test for which we don't have CI infrastructure for. So I think we should just rely on the fact blob handles this gracefully since that's in its spec, and don't test large file downloads for this endpoint. 

Also, you can make this part of the original PR to keep things simple.

HumairAK:
closed in favor of : https://github.com/kubeflow/pipelines/pull/12394

PR #12467, opened 2025-11-21
fix(backend): Parse JSON and decode base64 in artifact client

hbelmiro:

Fixes a bug in the artifact client where it returned raw JSON response bytes instead of parsing JSON and decoding base64 content. This caused the persistence agent's metrics reporting to fail with "gzip: invalid header" errors.

**Checklist:**


nsingla:
how do we catch these issues going forward? how can we add validations for persistent agent? or may be we don't it because its going away with mlmd removal? @HumairAK :thinking:

PR #12473, opened 2025-11-22
feat(launcher): Add configurable CA bundle environment variables to launcher v2

### LINKED ISSUE #12329: [feature] Configure custom CA cert environment variables
When a custom CA path is added to the launcher (added in #12082 ) the variables  `REQUESTS_CA_BUNDLE`, `AWS_CA_BUNDLE` and `SSL_CERT_FILE` are set by the KFP launcher for the executor: https://github.com/kubeflow/pipelines/blob/master/backend/src/v2/component/launcher_v2.go#L527-L538 . It would be useful to configure the launcher to take input variable names and only set those specified variables.

It would be useful to be able to set specific, necessary CA cert environment variables and avoid setting variables not necessary for the use case.

Currently `REQUESTS_CA_BUNDLE`, `AWS_CA_BUNDLE` and `SSL_CERT_FILE` are set by the KFP launcher.

---

Love this idea? Give it a 👍.

rahul810050:
### What this PR does / why we need it

This PR introduces support for configuring which CA-related environment variables the v2 launcher sets when a user provides a custom CA bundle path.

Currently, launcher_v2 unconditionally sets:

- REQUESTS_CA_BUNDLE
- AWS_CA_BUNDLE
- SSL_CERT_FILE

whenever CaCertPath is provided.
This behavior is too strict for some environments that want finer control—e.g., cloud-specific setups, distroless containers, or Python environments that expect only a subset of these variables.

This feature allows users to configure exactly which environment variables should be set.

### Which issue(s) this PR fixes

Fixes #12329

### Description of changes
1. Added new field to LauncherV2Options:
```bash
CACertEnvVars []string
```
Users may now specify which CA-related environment variables should be set.

2. Added full plumbing of the new option through the launcher stack

Including:

- NewLauncherV2
- LauncherV2.Execute
- executeV2
- execute

3. Updated default behavior

If CACertEnvVars is nil or empty, the launcher preserves legacy behavior and sets the original 3 environment variables:

```bash
REQUESTS_CA_BUNDLE
AWS_CA_BUNDLE
SSL_CERT_FILE
```

This ensures full backward compatibility.

4. Updated unit tests

All calls to executeV2 now include the new argument, and tests verify that launcher execution behavior continues to work as expected.

No changes were made to end-to-end or integration tests.

### Testing done

- Added/updated unit tests in `launcher_v2_test.go`
- Verified all backend unit tests pass:

```bash
cd backend
go test ./...
```

- Verified the component module tests pass:

```bash
ok   github.com/kubeflow/pipelines/backend/src/v2/component
```

End-to-end and API integration tests expectedly fail locally due to requiring a real Kubernetes cluster; this is normal and identical to upstream contributor experience.

### Backward compatibility

Fully backward compatible.

- If users do not specify `CACertEnvVars`, the launcher behaves exactly as before.
- No existing workflows break.
- No user APIs changed.

### Documentation

A short follow-up PR will update the docs in `docs/backend/tls` explaining how to configure CA env vars via `LauncherV2Options`.


rahul810050:
Hii @VaniHaripriya @droctothorpe could you please review it whenever you get a chance??

rahul810050:
@HumairAK could you please review it whenever you get a chance ??

nsingla:
I think this PR needs to be tested fully end to end. I am curious why it didn't work for you locally, did you try it on a kind/mini line cluster?

Also, if you are changing the cert on launcher, then we need the same cert on workflow controller and mlmd server as these 2 services are talking to the launcher. And with the upcoming Mlmd removal, we need the cert on all pods because all pods will talk to each other via api server.

nsingla:
Can you please bring this up in the upcoming community meeting and we can discuss this further. We may already have a solution in place to what you are trying to achieve, if not, we can discuss possible solutions.

PR #12474, opened 2025-11-23
Add fsGroup and seaweedfs

juliusvonkohout:
@droctothorpe @pschoen-itsc @akagami-harsh 
this might help on some clusters with special filesystems.
Let us wait for successful tests in https://github.com/kubeflow/manifests/pull/3279 too


juliusvonkohout:
We also have the same for the minio deployment, so now it is feature parity.

PR #12477, opened 2025-11-24
ci: Git artifact check

nsingla:
Replace the download artifact step to a new action to check if an artifact exists with a specific name as part of all the artifacts for this run or not. This is should clean up the unnecessary annotations that gets added to the run summary

**Checklist:**

Ran following runs for validation:
https://github.com/kubeflow/pipelines/actions/runs/19643843397/job/56254056732?pr=12477
https://github.com/kubeflow/pipelines/actions/runs/19644003435/job/56254604080?pr=12477

### DIFF
--- a/.github/actions/check-artifact-exists/check_artifact.py
+++ b/.github/actions/check-artifact-exists/check_artifact.py
@@ -1,0 +1,147 @@
+#!/usr/bin/env python3
+"""
+GitHub Action script to check if an artifact exists in the repository for a specific commit.
+Uses the GitHub REST API to list artifacts by name and filters by head SHA.
+Returns true if the artifact exists, false otherwise.
+"""
+
+import os
+import sys
+import requests
+import json
+
+def get_github_token() -> str:
+    """Get GitHub token from environment variables."""
+    token = os.getenv('GITHUB_TOKEN')
+    if not token:
+        print("::error::GITHUB_TOKEN environment variable is required")
+        sys.exit(1)
+    return token
+
+def get_head_sha() -> str:
+    """Get the head SHA from environment variables."""
+    head_sha = os.getenv('GITHUB_SHA')
+    if not head_sha:
+        print("::error::GITHUB_SHA environment variable is required")
+        sys.exit(1)
+    return head_sha
+
+def get_repository() -> str:
+    """Get the repository name from environment variables."""
+    repo = os.getenv('GITHUB_REPOSITORY')
+    if not repo:
+        print("::error::GITHUB_REPOSITORY environment variable is required")
+        sys.exit(1)
+    return repo
+
+def set_output(name: str, value: str) -> None:
+    """Set GitHub Actions output."""
+    github_output = os.getenv('GITHUB_OUTPUT')
+    if github_output:
+        with open(github_output, 'a') as f:
+            f.write(f"{name}={value}\n")
+    else:
+        # Fallback for older GitHub Actions runner versions
+        print(f"::set-output name={name}::{value}")
+
+def check_artifact_exists(artifact_name: str, token: str, repo: str, head_sha: str) -> bool:
+    """
+    Check if an artifact exists in the repository for a specific commit.
+
+    Args:
+        artifact_name: Name of the artifact to check
+        token: GitHub token for authentication
+        repo: Repository in format 'owner/repo'
+        head_sha: Commit SHA to filter artifacts by
+
+    Returns:
+        True if artifact exists, False otherwise
+    """
+    headers = {
+        'Authorization': f'token {token}',
+        'Accept': 'application/vnd.github+json',
+        'X-GitHub-Api-Version': '2022-11-28'
+    }
+
+    # GitHub API endpoint to list artifacts for a repository
+    url = f"https://api.github.com/repos/{repo}/actions/artifacts"
+
+    # Add query parameters - API will filter by artifact name
+    params = {
+        'per_page': 100,
+        'name': artifact_name
+    }
+
+    max_iterations = 100
+
+    try:
+        page = 1
+        while max_iterations > 0:
+            params['page'] = page
+            response = requests.get(url, headers=headers, params=params)
+            response.raise_for_status()
+
+            artifacts_data = response.json()
+            artifacts = artifacts_data.get('artifacts', [])
+            total_count = artifacts_data.get('total_count', 0)
+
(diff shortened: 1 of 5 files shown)


.github/actions/check-artifact-exists/check_artifact.py
```diff
@@ -0,0 +1,138 @@
+#!/usr/bin/env python3
```
gmfrasca, line 1: This should probably go in the `.github/scripts/` directory

.github/actions/check-artifact-exists/check_artifact.py
```diff
@@ -0,0 +1,138 @@
...
+import os
+import sys
+import requests
+import json
+
+def get_github_token() -> str:
+    """Get GitHub token from environment variables."""
+    token = os.getenv('GITHUB_TOKEN')
+    if not token:
+        print("::error::GITHUB_TOKEN environment variable is required")
```
gmfrasca, line 17: optional/minor: use `logging` instead so that log levels are appropriate and pythonic

.github/actions/check-artifact-exists/check_artifact.py
```diff
@@ -0,0 +1,138 @@
...
+        return False
+
+    except requests.exceptions.RequestException as e:
+        print(f"::error::Failed to check artifact existence: {e}")
+        sys.exit(1)
+    except json.JSONDecodeError as e:
+        print(f"::error::Failed to parse GitHub API response: {e}")
+        sys.exit(1)
+
+def set_output(name: str, value: str) -> None:
```
gmfrasca: code organization/trivial: move this up with the other getter/setter functions

.github/actions/check-artifact-exists/check_artifact.py
```diff
@@ -0,0 +1,138 @@
...
+
+    # Add query parameters - API will filter by artifact name
+    params = {
+        'per_page': 100,
+        'name': artifact_name
+    }
+
+    try:
+        page = 1
+        while True:
```
gmfrasca: maybe add an iteration counter/max here, so we're not at risk of looping infinitely

.github/actions/check-artifact-exists/check_artifact.py
```diff
@@ -0,0 +1,138 @@
...
+            params['page'] = page
+            response = requests.get(url, headers=headers, params=params)
+            response.raise_for_status()
+
+            artifacts_data = response.json()
+            artifacts = artifacts_data.get('artifacts', [])
+
+            if not artifacts:
```
gmfrasca, line 89: add a log message here so we know when we've completed iterating through the artifact pages

.github/actions/check-artifact-exists/check_artifact.py
```diff
@@ -0,0 +1,138 @@
...
+                artifact_head_sha = workflow_run.get('head_sha')
+
+                if artifact_head_sha == head_sha:
+                    print(f"::debug::Artifact '{artifact_name}' found with ID {artifact.get('id')}, matching head_sha {head_sha}")
+                    return True
+                else:
+                    print(f"::debug::Artifact '{artifact_name}' found but head_sha doesn't match (expected: {head_sha}, found: {artifact_head_sha})")
+
+            # Check if there are more pages
+            if len(artifacts) < params['per_page']:
```
gmfrasca: what if there were _exactly_ `per_page` number of artifacts?  Does this loop complete gracefully if we check the next page and there are no artifacts (handled on line 75), or will we hit a page not found error that would falsely break the script?

.github/actions/check-artifact-exists/check_artifact.py
nsingla: I thought about it as well, infact initially I had it there, but then I thought its better to keep it, so that its easy to manage and know where the corresponding script for this action is. 
I am working on few other actions, so it will organize things cleanly if we have the action and the script isolated into its own directory.
Let me know if the thought process works for you.

.github/actions/check-artifact-exists/check_artifact.py
nsingla: this is actually a github notice style print statement, not sure if logging to console using python logger would trigger the github notice :thinking:

.github/actions/check-artifact-exists/check_artifact.py
nsingla: good point, I will add it

.github/actions/check-artifact-exists/check_artifact.py
nsingla: in that case it will request the next page and will get 0 artifacts and break then

.github/actions/check-artifact-exists/check_artifact.py
nsingla: Done

.github/actions/check-artifact-exists/check_artifact.py
nsingla: Added a hard coded iteration count of 100

.github/actions/check-artifact-exists/check_artifact.py
nsingla: Done

.github/workflows/image-builds-with-cache.yml
```diff
@@ -68,25 +70,15 @@ jobs:
...
+          artifact_name: ${{ env.ARTIFACT_NAME }}
+        id: artifact-check
 
       - name: Set up Docker Buildx
         uses: docker/setup-buildx-action@v3
-        if: ${{ steps.artifact-download.outcome == 'failure' }}
+        if: ${{ steps.artifact-check.outputs.exists == 'false' }}
         id: setup-buildx
 
       - name: Build and save Docker image
```
VaniHaripriya, line 97: @nsingla  I think we also need to change the condition on this step- https://github.com/kubeflow/pipelines/blob/ff630c3ad9d9d03e50384d2165bfe6a0ec5f528b/.github/workflows/image-builds-with-cache.yml#L97
Since  steps.save-image.outcome != 'success' evaluate to true ,when` Check artifact exists` returns true and `save-image` gets skipped, so its outcome becomes skipped. But in that case, `Rebuild Images in case of failure` step should get skipped since the artifact exists.

nsingla: Nice catch, fixed

PR #12478, opened 2025-11-24
ci: adding a github action to generate test summary with html report link attached

nsingla:
Enhancing test report on the github step summary page, now we control what we print on the step summary page, and it now includes the link to the HTML report for easier debugging as well

**Checklist:**

Example of summary: https://github.com/kubeflow/pipelines/actions/runs/19865312242?pr=12478

### DIFF
--- a/.github/actions/junit-summary/action.yml
+++ b/.github/actions/junit-summary/action.yml
@@ -1,0 +1,76 @@
+name: "JUnit Test Summary"
+description: "Parse JUnit XML reports and publish to GitHub step summary with custom data support"
+
+inputs:
+  xml_files:
+    description: "Path(s) to JUnit XML file(s) or directories (space-separated, supports glob patterns)"
+    required: true
+  custom_data:
+    description: "JSON string with custom data to include in the summary (e.g., '{\"Build\": \"1234\", \"Branch\": \"main\"}')"
+    required: false
+    default: ""
+  custom_data_file:
+    description: "Path to JSON file with custom data to include in the summary"
+    required: false
+    default: ""
+  fail_on_test_failures:
+    description: "Whether to fail the step if tests fail (default: true)"
+    required: false
+    default: "true"
+  python_version:
+    description: "Python version to use"
+    required: false
+    default: "3.11"
+
+outputs:
+  total_tests:
+    description: "Total number of tests"
+    value: ${{ steps.parse.outputs.total_tests }}
+  total_passed:
+    description: "Number of passed tests"
+    value: ${{ steps.parse.outputs.total_passed }}
+  total_failed:
+    description: "Number of failed tests"
+    value: ${{ steps.parse.outputs.total_failed }}
+  total_errors:
+    description: "Number of tests with errors"
+    value: ${{ steps.parse.outputs.total_errors }}
+  total_skipped:
+    description: "Number of skipped tests"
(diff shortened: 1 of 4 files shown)


.github/actions/test-and-report/action.yml
```diff
@@ -142,7 +133,24 @@ runs:
       with:
         name: ${{ steps.name_gen.outputs.REPORT_NAME }}
         path: ${{ inputs.test_directory }}/reports/test-report.html
-        retention-days: 30
+        retention-days: 7
+      continue-on-error: true
+
+    - name: Publish Test Summary
```
gmfrasca: I think this and the following will both show up on the test run, just with skipped status if the conditional is met, so perhaps lets change the names of these steps to differentiate between them, so the reader isn't confused why one ran and what appears like a duplicate of same step was also skipped

nsingla: These are basically the same steps, 1 as a backup in case HTML report generation fails for any reason. But I will change the name to clear the confusion

nsingla: Done

.github/actions/junit-summary/action.yml
```diff
@@ -0,0 +1,154 @@
...
+    - name: Set up Python
+      uses: actions/setup-python@v4
+      with:
+        python-version: ${{ inputs.python_version }}
+
+    - name: Parse JUnit XML and Generate Summary
+      id: parse
+      shell: bash
+      run: |
+        # Expand glob patterns and collect all XML files
```
mprahl: If we are using Python anyway, can we just have all this logic in .github/actions/junit-summary/junit_to_summary.py as well?

nsingla: good point, I will move it

nsingla: Done

.github/workflows/api-server-tests.yml
```diff
@@ -36,6 +36,7 @@ on:
     paths:
       - '.github/workflows/api-server-tests.yml'
       - '.github/actions/create-cluster/**'
+      - '.github/actions/test-and-report/**'
```
mprahl, line 39: Can we also add .github/actions/junit-summary here too?

.github/actions/junit-summary/junit_to_summary.py
```diff
@@ -0,0 +1,430 @@
...
+        failure_type=failure_type,
+        error_message=error_message,
+        system_out=system_out,
+        system_err=system_err
+    )
+
+
+def aggregate_reports(xml_files: List[Path]) -> TestReport:
+    """Aggregate multiple JUnit XML files into a single report."""
+    all_suites = []
```
mprahl, line 165: Is it valid if this is empty after processing? Should we return an error instead?

nsingla: Yeah, we will just see 0 tests in the report if the xml report files have 0 tests, but there are other safety nets:
1. empty file list cannot be passed - The `expand_file_patterns()` function which runs before `aggregate_reports`in `main` , will exit with code 1 if no XML files are found
2. If xml files are invalid, then `parse_junit_xml` function will throw exit code 1

.github/workflows/api-server-tests.yml
nsingla: Done

Issue #12508, opened 2025-12-02
[backend] KFP 2.15.0 breaks IRSA on AWS: AccessKey is now required


rvalkenaers:
### Environment

*  How did you deploy Kubeflow Pipelines (KFP)?
  Standalone deployment using the official KFP Kustomize manifests.

*  KFP version:
  2.15.0

*  KFP SDK version:
  Not relevant for this issue (problem occurs in api-server startup, independent of SDK).

### Steps to reproduce

1. Deploy Kubeflow Pipelines 2.15.0 using the official standalone Kustomize manifests.
2. Configure object storage to use AWS S3 with IRSA authentication.
3. Set the following environment variables to empty values (as was valid in KFP 2.14.x):
   - `OBJECTSTORECONFIG_ACCESSKEY`
   - `OBJECTSTORECONFIG_SECRETACCESSKEY`
4. Start the KFP API server.
Observed error in the api-server logs:
```
Failed to initialize ClientManager: failed to initialize object store: failed to build config from environment variables: ObjectStoreConfig.AccessKey is required
```

### Expected result

In KFP 2.14.x it was valid to leave `OBJECTSTORECONFIG_ACCESSKEY` and
`OBJECTSTORECONFIG_SECRETACCESSKEY` empty when using IRSA authentication on AWS.
The API server should accept empty values and fall back to the underlying IRSA
credentials, as it did in previous versions.

### Materials and Reference

- [Use IRSA with the AWS SDK](https://docs.aws.amazon.com/eks/latest/userguide/iam-roles-for-service-accounts-minimum-sdk.html)
- Regression confirmed between 2.14.x → 2.15.0
- Likely introduced by the following PR: https://github.com/kubeflow/pipelines/pull/12394 

---

Impacted by this bug? Give it a 👍.

juliusvonkohout:
I strongly recommend to go via seaweedfs so https://github.com/kubeflow/pipelines/tree/master/manifests/kustomize/third-party/seaweedfs#gateway-to-remote-object-storage gateway to aws. our tests are done with seaweedfs

rvalkenaers:
Thanks for the quick response.

I looked at the SeaweedFS gateway documentation you linked, but the examples only show configurations using explicit access keys and secret keys. I don’t see any guidance on how to use the gateway with IRSA, and our setup relies on IRSA-based authentication instead of static credentials.

[...]

PR #12512, opened 2025-12-03
fix(backend): Refactor S3 bucket management by replacing MinIO client code with AWS SDK v2 and improving configuration validation

hbelmiro:
### Problem

The API server required `OBJECTSTORECONFIG_ACCESSKEY` and `OBJECTSTORECONFIG_SECRETACCESSKEY` to be non-empty, which broke AWS IRSA (IAM Roles for Service Accounts) authentication where credentials are obtained from the environment.

### Solution

- Make `AccessKey` and `SecretAccessKey` optional in validation
- When credentials are empty, use the AWS SDK's default credential chain (supports IRSA, environment variables, instance profiles)
- Replace minio-go with AWS SDK v2 for bucket operations (consistent SDK usage)
- Only suppress `BucketAlreadyOwnedByYou` errors; `BucketAlreadyExists` now surfaces as a warning to help detect misconfiguration

### Changes

- `validateRequiredConfig`: No longer requires credentials; only validates both are set or both empty
- `loadAWSConfig`: Uses default credential chain when credentials are empty
- `ensureBucketExists`: Refactored to use AWS SDK v2 instead of minio-go
- `isBucketAlreadyOwnedByUs`: Only suppresses `BucketAlreadyOwnedByYou` (concurrent creation race), surfaces `BucketAlreadyExists` as warning

Fixes:

- https://github.com/kubeflow/pipelines/issues/12508

**Checklist:**

### DIFF
--- a/backend/src/apiserver/client_manager/client_manager.go
+++ b/backend/src/apiserver/client_manager/client_manager.go
@@ -1072,7 +1072,7 @@ func validateRequiredConfig(bucketName, host, accessKey, secretKey string) error
 	if host == "" {
 		return fmt.Errorf("ObjectStoreConfig.Host is required")
 	}
-	if (accessKey != "") != (secretKey != "") {
+	if (accessKey == "" && secretKey != "") || (accessKey != "" && secretKey == "") {
 		return fmt.Errorf("ObjectStoreConfig.AccessKey and ObjectStoreConfig.SecretAccessKey must both be set or both be empty")
 	}
 	return nil


backend/src/apiserver/client_manager/client_manager.go
```diff
@@ -952,15 +953,15 @@ func initBlobObjectStore(ctx context.Context, initConnectionTimeout time.Duratio
 		return nil, fmt.Errorf("failed to build config from environment variables: %w", err)
 	}
 
+	if err := ensureBucketExists(ctx, blobConfig); err != nil {
+		glog.Warningf("Failed to ensure bucket exists (may already exist): %v", err)
```
gmfrasca: Since we are now providing and boiling up context here on what the error is, this warning could be a bit confusing/misleading, especially the `(may already exist)` part.  For example, a warning shown could be:

`Failed to ensure bucket exists (may already exist): failed to create object store bucket: Error occurred for some other reason coming from s3 client (network unavailable, etc).`  

here we're given multiple possible causes for the error even though the last one is the true issue, and "(may already exist)" is a red herring.

hbelmiro: Done.

backend/src/apiserver/client_manager/client_manager.go
```diff
@@ -952,15 +953,15 @@ func initBlobObjectStore(ctx context.Context, initConnectionTimeout time.Duratio
 		return nil, fmt.Errorf("failed to build config from environment variables: %w", err)
 	}
 
+	if err := ensureBucketExists(ctx, blobConfig); err != nil {
+		glog.Warningf("Failed to ensure bucket exists: %v", err)
```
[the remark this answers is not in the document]
hbelmiro: This is intentional. Some S3 backends (e.g. MinIO/SeaweedFS) can return bucket create/existence errors even when the bucket is usable, and we don’t want apiserver startup to fail on those false negatives.
I’ve added a comment explaining why we keep it as a warning instead of returning an error.

gmfrasca (commented):
/lgtm

backend/src/apiserver/client_manager/client_manager.go
```diff
@@ -952,15 +953,17 @@ func initBlobObjectStore(ctx context.Context, initConnectionTimeout time.Duratio
 		return nil, fmt.Errorf("failed to build config from environment variables: %w", err)
 	}
 
+	if err := ensureBucketExists(ctx, blobConfig); err != nil {
+		// Best-effort: some S3-compatible stores (e.g., MinIO/SeaweedFS) can return non-AWS-ish errors for bucket
+		// creation/existence checks even when the bucket is usable; don't fail apiserver init on those false negatives.
```
HumairAK, line 958: I'm a little concerned about this, but not enough to block this PR. It feels like it could result in valid errors getting through. However eventually I would like KFP to stop creating buckets all together, and we can just validate credentials as part of a health check.

PR #12391, opened 2025-10-24
feature: add ha overlay for seaweedfs

pschoen-itsc:
**Description of your changes:**

**Checklist:**
- [x] You have [signed off your commits](https://www.kubeflow.org/docs/about/contributing/#sign-off-your-commits)
- [x] The title for your pull request (PR) should follow our title convention. [Learn more about the pull request title convention used in this repository](https://github.com/kubeflow/pipelines/blob/master/CONTRIBUTING.md#pull-request-title-convention).

PR #12515, opened 2025-12-03
Add KEP-12513: Introduce PVC-based artifact storage for Kubeflow Pipelines

### LINKED ISSUE #12513: KEP-12513: Filesystem-Based Artifact Storage for Kubeflow Pipelines
## Description

This KEP proposes adding PVC-based artifact storage as an alternative to S3/GCS/MinIO.

**Problem:** KFP requires object storage (S3/GCS/MinIO), which adds complexity for local/air-gapped deployments.

**Solution:** New `kfp-artifacts://` URI scheme that routes artifacts through an artifact server API, with storage on Kubernetes PVCs.

hbelmiro:
Resolves: https://github.com/kubeflow/pipelines/issues/12513

### DIFF
--- a/proposals/12513-pvc-artifact-storage/README.md
+++ b/proposals/12513-pvc-artifact-storage/README.md
@@ -155,7 +155,7 @@ By providing PVC-based storage as an alternative, we:
 **When to use filesystem storage:**
 
 - Deployments where eliminating object storage dependency is preferred
-- Environments where Kubeflow distributions or platform providers prefer not to support additional storage systems
+- Environments where Kubeflow distributions or platform providers do not offer a fully supported object store solution
 - Multi-tenant deployments requiring per-namespace storage isolation, scaling, and quotas
 - Development and experimentation scenarios
 
--- a/proposals/12513-pvc-artifact-storage/README.md
+++ b/proposals/12513-pvc-artifact-storage/README.md
@@ -194,7 +194,7 @@ This KEP proposes adding a new artifact storage backend that uses filesystem sto
 
 1. Create one PVC per namespace for artifact storage
 2. Use configurable access mode with sensible defaults (RWO)
-3. Organize artifacts in a filesystem hierarchy within the PVC
+3. Organize artifacts in a filesystem hierarchy within the PVC that is namespace aware
 4. Provide transparent access through the existing KFP artifact APIs with new `kfp-artifacts://` URI scheme
 5. Maintain compatibility with existing pipeline definitions that don't have hardcoded storage paths
 6. Support separate scaling of artifact serving through artifacts-only instances
--- a/proposals/12513-pvc-artifact-storage/README.md
+++ b/proposals/12513-pvc-artifact-storage/README.md
@@ -353,7 +353,7 @@ Storage backends in KFP are determined by the `defaultPipelineRoot` URL scheme, 
 For filesystem storage, this KEP introduces a new URL scheme:
 
 ```text
-kfp-artifacts://<namespace>/<pipeline>/<run-id>/<node-id>/<artifact-name>
+kfp-artifacts://<namespace>/<pipeline-name>/<run-id>/<node-id>/<artifact-name>
 ```
 
 This follows the same pattern as existing storage schemes (s3://, gs://, minio://), where the namespace maps to the bucket name and the rest maps to the prefix.
@@ -370,11 +370,11 @@ defaultPipelineRoot: "kfp-artifacts://{namespace}"
 
 There are three related but distinct formats:
 
-| Concept                  | Format                                                                       | Example                                                       |
-|--------------------------|------------------------------------------------------------------------------|---------------------------------------------------------------|
-| **URI** (stored in MLMD) | `kfp-artifacts://<namespace>/<pipeline>/<run-id>/<node-id>/<artifact-name>`  | `kfp-artifacts://team-a/my-pipeline/abc123/step1/output`      |
-| **API Endpoint**         | `/apis/v2beta1/runs/{run_id}/nodes/{node_id}/artifacts/{artifact_name}:read` | `/apis/v2beta1/runs/abc123/nodes/step1/artifacts/output:read` |
-| **Filesystem Path**      | `<mount_path>/<namespace>/<pipeline>/<run-id>/<node-id>/<artifact-name>`     | `/artifacts/team-a/my-pipeline/abc123/step1/output`           |
+| Concept                  | Format                                                                           | Example                                                       |
+|--------------------------|----------------------------------------------------------------------------------|---------------------------------------------------------------|
+| **URI** (stored in MLMD) | `kfp-artifacts://<namespace>/<pipeline-name>/<run-id>/<node-id>/<artifact-name>` | `kfp-artifacts://team-a/my-pipeline/abc123/step1/output`      |
+| **API Endpoint**         | `/apis/v2beta1/runs/{run_id}/nodes/{node_id}/artifacts/{artifact_name}:read`     | `/apis/v2beta1/runs/abc123/nodes/step1/artifacts/output:read` |
+| **Filesystem Path**      | `<mount_path>/<namespace>/<pipeline>/<run-id>/<node-id>/<artifact-name>`         | `/artifacts/team-a/my-pipeline/abc123/step1/output`           |
 
 **Why the API endpoint doesn't include namespace/pipeline:**
 
@@ -815,7 +815,7 @@ Note: This is a new endpoint that doesn't exist in the current KFP artifact serv
 
 ```json
 {
-  "uri": "kfp-artifacts://<namespace>/<pipeline>/<run_id>/<node_id>/<artifact_name>"
+  "uri": "kfp-artifacts://<namespace>/<pipeline-name>/<run_id>/<node_id>/<artifact_name>"
 }
 ```
 
@@ -1438,7 +1438,7 @@ This design allows the control plane to serve artifacts from any namespace while
 
 In namespace-local mode, workflow pods connect directly to their namespace's artifact server:
 
-- Artifact URI: `kfp-artifacts://<namespace>/<pipeline>/<run-id>/...`
+- Artifact URI: `kfp-artifacts://<namespace>/<pipeline-name>/<run-id>/...`
 - Resolved to: `http://ml-pipeline-artifact-server.<namespace>.svc.cluster.local:8080/...`
 - UI handles the resolution based on deployment mode
 
--- a/proposals/12513-pvc-artifact-storage/README.md
+++ b/proposals/12513-pvc-artifact-storage/README.md
@@ -116,7 +116,7 @@ While KFP currently ships with S3-compatible storage by default, this still requ
 
 ### Reduced External Dependencies
 
-While KFP's default manifests ship with S3-compatible storage pre-configured (no manual credential setup required), filesystem storage replaces the object storage system with a KFP-native artifact server. This means:
+While KFP's default manifests ship with S3-compatible storage pre-configured (with credentials set up in the manifests, so no manual credential setup is required during deployment), filesystem storage replaces the object storage system with a KFP-native artifact server. This means:
 
 - No external object storage project to track for updates and security patches
 - Artifact issues debugged within KFP ecosystem (no separate S3 API layer)
--- a/proposals/12513-pvc-artifact-storage/README.md
+++ b/proposals/12513-pvc-artifact-storage/README.md
@@ -511,7 +511,7 @@ GET /apis/v2beta1/runs/{run_id}/nodes/{node_id}/artifacts/{artifact_name}:read
 GET /apis/v1beta1/runs/{run_id}/nodes/{node_id}/artifacts/{artifact_name}:read
 ```
 
-Note: Both v1beta1 and v2beta1 endpoints are supported because both V1 and V2 pipelines will need to read artifacts via this API when using filesystem storage. Both endpoints use the same implementation. There is an open question about whether this API should be considered public or internal. See [Open Questions: External Client Support](#open-questions).
+Note: Both v1beta1 and v2beta1 endpoints are supported because both V1 and V2 pipelines will need to read artifacts via this API when using filesystem storage. Both endpoints use the same implementation. There is an open question about whether this API should be considered public or internal. See [Open Questions: External Client Support](#external-client-support).
 
 **Request Parameters:**
 
--- a/proposals/12513-pvc-artifact-storage/README.md
+++ b/proposals/12513-pvc-artifact-storage/README.md
@@ -257,7 +257,7 @@ Example configurations:
 
 ```yaml
 # Both modes use the same path structure for consistency
-# {namespace} is substituted at runtime (e.g., "kubeflow" in single-user mode)
+# {namespace} is substituted at runtime (single-user mode uses the KFP install namespace)
 defaultPipelineRoot: "kfp-artifacts://{namespace}"
 ```
 
@@ -370,7 +370,7 @@ Routing rules:
                    ▼
 ┌──────────────────────────────────────┐
 │      Central Artifact Server         │
-│        (namespace: kubeflow)         │
+│   (namespace: KFP install namespace) │
 │                                      │
 │  • SubjectAccessReview               │
 │  • Mounts shared PVC                 │
@@ -655,8 +655,11 @@ The KFP API server manages artifact infrastructure lifecycle for the shared arti
 
 ##### Namespace Annotation-Based Provisioning
 
-The KFP API server watches for namespaces with a specific annotation to provision artifact infrastructure. When Kubeflow Profiles are used, the Profile controller creates namespaces with the `pipelines.kubeflow.org/enabled: "true"` annotation. For standalone KFP, namespaces can be annotated directly.
-For annotated namespaces, the API server provisions the required artifact infrastructure and reads the namespace `kfp-launcher` ConfigMap to determine whether `artifactServer.dedicated: true` is set.
+The shared artifact server (default) is managed in the KFP install namespace and does not depend on namespace annotations (including in single-user mode).
+
+Namespace annotations are only used to drive optional per-namespace provisioning in multi-tenant setups. In multi-user Kubeflow, the Profile controller creates user namespaces (Profiles) and can apply the `pipelines.kubeflow.org/enabled: "true"` annotation to those namespaces. The `kubeflow` namespace is the control-plane namespace, not a Profile. For standalone KFP, namespaces can be annotated directly.
+
+For annotated namespaces, the API server reads the namespace `kfp-launcher` ConfigMap to determine whether `artifactServer.dedicated: true` is set; if so, it provisions dedicated per-namespace infrastructure.
 
 ##### API Server Implementation
 
@@ -836,10 +839,9 @@ For example, components that directly call an object-store SDK (e.g., `boto3`) s
 
 #### `ArtifactServerManager` (API Server)
 
-- Watches namespaces with `pipelines.kubeflow.org/enabled=true` annotation
-- Creates PVC, Deployment, Service based on the namespace's dedicated artifact server opt-in
-- Reads `artifactServer.dedicated` from namespace's `kfp-launcher` ConfigMap, falls back to the shared default (`false`) when omitted
-- No action for namespaces without the annotation
+- Ensures the shared artifact server (default) exists in the KFP install namespace when filesystem storage is enabled
+- Watches KFP-enabled namespaces (e.g., `pipelines.kubeflow.org/enabled: "true"`) for optional per-namespace provisioning
+- Creates per-namespace PVC and artifact server workload only when the namespace opts in (`artifactServer.dedicated: true` in `kfp-launcher`)
 - Skips namespaces where infrastructure already exists
 - Idempotent: Re-processing same namespace doesn't create duplicate resources
 - Different namespaces can use different settings (shared default or dedicated per-namespace)
--- a/proposals/12513-pvc-artifact-storage/README.md
+++ b/proposals/12513-pvc-artifact-storage/README.md
@@ -166,7 +166,7 @@ Based on the user story "As a user, I want to provision Kubeflow Pipelines with 
 2. **Enable zero-configuration storage** for experimentation use cases - a KFP server can be installed with just a PVC for artifact storage
 3. **Provide namespace-isolated artifact storage** with proper subject access review guards in multi-user mode
 4. **Allow any Kubernetes access mode to be configured** - KFP passes through the configuration to Kubernetes (RWO default)
-5. **Support existing pipelines** that use KFP's standard artifact types (Dataset, Model, etc.) - pipelines work unchanged with the new filesystem backend
+5. **Pipeline compatibility (no pipeline code changes)** - pipelines that use KFP-managed artifact passing (e.g., `Input/Output[Dataset|Model]` and `.path`) work unchanged with the filesystem backend (unless they contain hardcoded `s3://`/`gs://`/`minio://` URIs or call object-store SDKs directly)
 6. **Match existing artifact persistence behavior** - artifacts persist indefinitely until explicitly deleted (no automatic cleanup)
 7. **Enable separate scaling of artifact serving** through an artifacts-only KFP instance with `--artifacts-only` flag
 8. **Ensure UI can properly handle artifact downloads** with the new `kfp-artifacts://` URI scheme
@@ -176,7 +176,7 @@ Based on the user story "As a user, I want to provision Kubeflow Pipelines with 
 1. **Replace object storage** - filesystem storage is an additional option, not a replacement
 2. **Cross-namespace artifact sharing** - Initial implementation focuses on namespace isolation
 3. **Performance optimization for large-scale workloads** - filesystem storage may not match object storage performance for massive datasets or high concurrency scenarios
-4. **Storage migration tools** - Migrating from S3 to PVC/filesystem is not supported
+4. **Migration of existing installations / artifacts** - switching an existing KFP installation from S3-compatible storage to filesystem storage does not migrate previously produced artifacts or rewrite stored artifact URIs (no migration tooling in this KEP)
 5. **Advanced storage features** - No versioning, replication, or geo-distribution initially
 
 ## Proposal
@@ -309,7 +309,7 @@ As an operator, I want to deploy KFP in central mode by default, but configure s
 3. **Scalability**: filesystem storage is not designed for massive scale like object storage
 4. **Cost**: Cost depends on the storage class and provider - could be more or less expensive than object storage
 5. **Portability**: Pipelines using filesystem storage are tied to Kubernetes infrastructure
-6. **No Migration Support**: Migrating existing artifacts from S3/GCS to filesystem storage is not supported. Users must choose their storage backend at deployment time and stick with it
+6. **No Migration Support**: Migrating existing artifacts from S3-compatible storage to filesystem storage is not supported in this KEP (no artifact copy or URI rewrite tooling)
 
 ### Risks and Mitigations
 
@@ -1791,14 +1791,14 @@ OBJECTSTORECONFIG_ARTIFACTSERVER_DEPLOYMENTMODE=central
 
 ### Migration Path
 
-To migrate from object storage to filesystem storage:
+To switch an installation from S3-compatible storage to filesystem storage for **new runs**:
 
 1. Update `defaultPipelineRoot` to use `kfp-artifacts://` scheme
 2. Add `ObjectStoreConfig.Filesystem` configuration
 3. Configure `ObjectStoreConfig.ArtifactServer` based on deployment needs
 4. Restart KFP components to apply changes
 
-**Note**: Migrating existing artifacts from S3/GCS to filesystem storage is not supported. Users must choose their storage backend at deployment time and stick with it.
+**Note**: Existing runs/artifacts are not migrated. Artifact URIs already recorded in MLMD remain `s3://`/`minio://`/`gs://` (etc.) and continue to refer to the original object store location. This KEP does not include tooling to copy historical artifacts or rewrite stored URIs.
 
 ### Backward Compatibility
 
--- a/proposals/12513-pvc-artifact-storage/README.md
+++ b/proposals/12513-pvc-artifact-storage/README.md
@@ -1,0 +1,1650 @@
+# KEP-12513: Filesystem-Based Artifact Storage for Kubeflow Pipelines
+
+## Table of Contents
+
+- [Summary](#summary)
+- [Architecture at a Glance](#architecture-at-a-glance)
+- [Motivation](#motivation)
+  - [Goals](#goals)
+  - [Non-Goals](#non-goals)
+- [Proposal](#proposal)
+  - [Overview](#overview)
+  - [User Stories](#user-stories)
+  - [Notes/Constraints/Caveats](#notesconstraintscaveats)
+  - [Risks and Mitigations](#risks-and-mitigations)
+- [Design Details](#design-details)
+  - [Configuration Approach](#configuration-approach)
+  - [Storage Backend Selection](#storage-backend-selection)
+    - [URI, API Endpoint, and Filesystem Path Relationship](#uri-api-endpoint-and-filesystem-path-relationship)
+  - [Artifact Server Architecture](#artifact-server-architecture)
+    - [Backend Responsibilities](#backend-responsibilities)
+    - [Request Routing](#request-routing)
+    - [UI Integration](#ui-integration)
+  - [PVC Management](#pvc-management)
+    - [PVC Creation and Configuration](#pvc-creation-and-configuration)
+    - [Storage Quota Enforcement](#storage-quota-enforcement)
+  - [Component Modifications](#component-modifications)
+  - [Multi-User Isolation and Authorization](#multi-user-isolation-and-authorization)
+    - [Namespace Isolation](#namespace-isolation)
+    - [Multi-User Isolation Strategy](#multi-user-isolation-strategy)
+    - [Namespace-Local Artifact Pods Architecture](#namespace-local-artifact-pods-architecture)
+    - [Subject Access Review Integration](#subject-access-review-integration)
+    - [Single-User vs Multi-User Deployments](#single-user-vs-multi-user-deployments)
+  - [Artifact Lifecycle Management](#artifact-lifecycle-management)
+    - [Artifact Persistence](#artifact-persistence)
+    - [Cleanup Mechanisms](#cleanup-mechanisms)
+    - [Caching Support](#caching-support)
+    - [Features Not Replicated](#features-not-replicated)
+  - [Pipeline Compatibility](#pipeline-compatibility)
+- [Test Plan](#test-plan)
+  - [Unit Tests](#unit-tests)
+  - [Integration Tests](#integration-tests)
+- [Configuration Reference](#configuration-reference)
+  - [Complete Configuration Example](#complete-configuration-example)
+  - [Environment Variable Overrides](#environment-variable-overrides)
+- [Migration and Compatibility](#migration-and-compatibility)
+  - [Migration Path](#migration-path)
+  - [Backward Compatibility](#backward-compatibility)
+- [Open Questions](#open-questions)
+- [Drawbacks](#drawbacks)
+- [Infrastructure Needed](#infrastructure-needed)
+
+## Summary
+
+This KEP proposes adding filesystem-based storage as an alternative artifact storage backend for Kubeflow Pipelines v2. Currently, KFP requires object storage (S3-compatible storage or Google Cloud Storage) for storing pipeline artifacts, which can be a barrier for users who want to experiment with KFP in development environments. This proposal introduces filesystem storage as an additional option that eliminates the need for object storage configuration, making it easier for users to get started with KFP.
+
+The filesystem backend will primarily use `PersistentVolumeClaim` (PVC) based storage in Kubernetes environments, providing namespace-isolated storage using Kubernetes native `PersistentVolumes`. However, the design is flexible enough to support other filesystem backends (e.g., local filesystem for development). Users can specify any access mode value which KFP will pass through to Kubernetes without validation (e.g., `ReadWriteMany` for parallel task execution across nodes, `ReadWriteOnce` for single node access, etc.), with RWO as the default if not specified. The actual behavior depends on what the underlying storage class supports. Existing pipelines will work without modification unless they contain hardcoded S3/object storage paths.
+
+To support scalability, KFP will introduce a new artifact URI scheme (`kfp-artifacts://`) that routes all artifact requests through KFP's artifact server. Pipeline pods never mount PVCs directly - they upload and download artifacts via the artifact server API, which handles the actual filesystem operations. The artifact server can be scaled independently as an artifacts-only instance.
+
+## Architecture at a Glance
+
+### Quick Comparison
+
+| Aspect               | Current (Object Storage)                   | Proposed (Filesystem Storage)         |
+|----------------------|--------------------------------------------|---------------------------------------|
+| **Storage Backend**  | S3, GCS, MinIO                             | Kubernetes PVC                        |
+| **URI Scheme**       | `s3://`, `gs://`, `minio://`               | `kfp-artifacts://`                    |
+| **Configuration**    | Credentials, endpoints, buckets (required) | Optional (uses K8s defaults)          |
+| **Multi-tenancy**    | Bucket per namespace                       | PVC per namespace                     |
+| **Setup Complexity** | High (external dependencies)               | Low (Kubernetes native with defaults) |
+
+### High-Level Flow
+
+```text
+Pipeline Submission
+        │
+        ▼
+Driver (detects "kfp-artifacts://")
+        │
+        ▼
+Artifact Server (mounts PVC)
+        │
+        ├──────────────────────────────────┐
+        │                                  │
+        ▼                                  ▼
+Pipeline Pods                         UI/API Clients
+(upload/download via API)             (view artifacts)
+        │                                  │
+        └──────────► Artifact Server ◄─────┘
+                           │
+                           ▼
+                    Read/Write to PVC
+```
+
+### Key Components
+
+- **Storage**: Kubernetes `PersistentVolumeClaims` (one per namespace)
+- **URI Format**: `kfp-artifacts://<namespace>/<pipeline>/<run-id>/<node-id>/<artifact-name>`
+- **API Changes**: Extended `/healthz` endpoint for UI configuration discovery
+- **Deployment Modes**:
+  - **Central**: Single artifact server for all namespaces (default)
+  - **Namespace-local**: One artifact server per namespace (better isolation)
+
+## Motivation
+
+Setting up object storage is often a significant hurdle for users who want to experiment with Kubeflow Pipelines. Users must:
+
+- Configure S3-compatible storage with proper credentials
+- Manage access keys and secrets
+- Understand object storage concepts
+- Set up and maintain additional infrastructure
+
+For development, testing, and experimentation use cases, this overhead is unnecessary and discourages adoption. Many users simply want to try KFP without dealing with external dependencies.
+
+Additionally, in some environments:
+
+- Object storage may not be available or allowed
+- Network policies may restrict external storage access
+- Organizations may prefer to use existing Kubernetes storage infrastructure
+- Development clusters may have limited resources
+
+By providing PVC-based storage as an alternative, we can:
+
+- Lower the barrier to entry for new users
+- Enable quick experimentation and prototyping
+- Leverage existing Kubernetes storage infrastructure
+- Simplify development environment setup
+- Enable portability between different storage backends for pipelines using KFP's artifact APIs
+
+### Goals
+
+Based on the user story "As a user, I want to provision Kubeflow Pipelines with just a PVC for artifact storage so that I can quickly get started", this KEP aims to:
+
+1. **Add filesystem storage as an additional backend option** alongside S3-compatible and Google Cloud Storage, primarily using PVC but not limited to it
+2. **Enable zero-configuration storage** for experimentation use cases - a KFP server can be installed with just a PVC for artifact storage
+3. **Provide namespace-isolated artifact storage** with proper subject access review guards in multi-user mode
+4. **Allow any Kubernetes access mode to be configured** - KFP passes through the configuration to Kubernetes (RWO default)
+5. **Support existing pipelines** that use KFP's standard artifact types (Dataset, Model, etc.) - pipelines work unchanged with the new filesystem backend
+6. **Match existing artifact persistence behavior** - artifacts persist indefinitely until explicitly deleted (no automatic cleanup)
+7. **Enable separate scaling of artifact serving** through an artifacts-only KFP instance with `--artifacts-only` flag
+8. **Ensure UI can properly handle artifact downloads** with the new `kfp-artifacts://` URI scheme
+
+### Non-Goals
+
+1. **Replace object storage** - filesystem storage is an additional option, not a replacement
+2. **Cross-namespace artifact sharing** - Initial implementation focuses on namespace isolation
+3. **Performance optimization for large-scale workloads** - filesystem storage may not match object storage performance for massive datasets or high concurrency scenarios
+4. **Storage migration tools** - Migrating from S3 to PVC/filesystem is not supported
+5. **Advanced storage features** - No versioning, replication, or geo-distribution initially
+
+## Proposal
+
+### Overview
+
+This KEP proposes adding a new artifact storage backend that uses filesystem storage (primarily Kubernetes `PersistentVolumeClaims`) instead of object storage. The implementation will:
+
+1. Create one PVC per namespace for artifact storage
+2. Use configurable access mode with sensible defaults (RWO)
+3. Organize artifacts in a filesystem hierarchy within the PVC
+4. Provide transparent access through the existing KFP artifact APIs with new `kfp-artifacts://` URI scheme
+5. Maintain compatibility with existing pipeline definitions that don't have hardcoded storage paths
+6. Support separate scaling of artifact serving through artifacts-only instances
+7. Update the UI to seamlessly handle artifact downloads from filesystem storage
+
+### User Stories
+
+#### Story 1: User Running Pipelines on Local Kubernetes
+
+As a user with KFP on kind/minikube/k3s, I want my pipeline artifacts to automatically use the local cluster's default `StorageClass` via the `kfp-artifacts://` scheme, so that I can develop pipelines offline without any storage configuration.
+
+**Acceptance Criteria:**
+
+- KFP works out-of-the-box with filesystem storage on local clusters
+- Artifacts are stored using the `kfp-artifacts://` URI scheme
+- No S3/GCS credentials required
+- Artifact viewing in UI works seamlessly
+
+#### Story 2: Operator Using Default Filesystem Configuration
+
+As an operator, I want to deploy KFP with minimal configuration using filesystem storage defaults, so that I can get KFP running quickly without cloud storage dependencies.
+
+**Acceptance Criteria:**
+
+- Single configuration option to enable filesystem storage
+- No need to create or manage S3/GCS buckets
+- No cloud IAM policies to configure
+- Storage automatically provisioned via PVCs
+- Backup/restore follows standard Kubernetes PVC procedures
+
+#### Story 3: Operator Configuring Storage Class and Size
+
+As an operator, I want to configure KFP to use a specific StorageClass and PVC size instead of defaults, so that I can match storage performance and capacity to my workload requirements.
+
+**Acceptance Criteria:**
+
+- Can specify `StorageClass` in KFP configuration
+- Can set PVC size limits (global configuration)
+- Storage quotas enforced via Kubernetes `ResourceQuotas`
+- Clear error messages when storage limits are reached
+- Can choose between RWO and RWX access modes based on needs
+
+#### Story 4: User Migrating from S3 to Filesystem Storage
+
+As a user with existing pipelines containing components that call `boto3.upload_file()` directly, I want KFP system artifacts to use `kfp-artifacts://` with PVC storage while my custom components continue accessing S3, so that I can migrate incrementally without rewriting all components at once.
+
+**Acceptance Criteria:**
+
+- KFP artifacts use filesystem storage via `kfp-artifacts://` URIs
+- Components using boto3/gsutil for S3 continue to work
+- Can mix storage backends (system artifacts on PVC, custom data on S3)
+- Clear documentation on what uses filesystem vs object storage
+
+#### Story 5: Operator Deploying Multi-Tenant KFP with Namespace Isolation
+
+As an operator, I want to deploy KFP in namespace-local mode where each namespace annotated with `pipelines.kubeflow.org/enabled=true` gets its own artifact server pod and dedicated PVC, so that Team A's artifacts in namespace `team-a` are physically isolated from Team B's artifacts in namespace `team-b`.
+
+**Acceptance Criteria:**
+
+- Each namespace with `pipelines.kubeflow.org/enabled=true` annotation gets its own artifact server deployment
+- Each namespace gets its own dedicated PVC (no shared storage)
+- Artifact server in `team-a` namespace cannot access PVC in `team-b` namespace
+- Users can only access artifacts in namespaces they have RBAC permissions for (via `SubjectAccessReview`)
+- Physical isolation verified: deleting `team-a` namespace doesn't affect `team-b`'s artifacts
+
+#### Story 6: Operator Implementing Air-Gapped ML Platform
+
+As an operator in a regulated environment (e.g., healthcare, finance), I want to deploy KFP with filesystem storage using an encrypted StorageClass (e.g., `encrypted-gp3`) and no external network access, so that PHI/PII in artifacts never leaves our on-premises Kubernetes cluster.
+
+**Acceptance Criteria:**
+
+- All artifacts stored on PVCs within the cluster (no S3/GCS/external storage)
+- KFP configuration uses `Filesystem.Type: "pvc"` with encrypted StorageClass
+- Artifact server pods have no external network routes (verified by network policies)
+- `SubjectAccessReview` validates all artifact access requests
+- Encryption at rest provided by the configured StorageClass (e.g., `encrypted-gp3`)
+
+#### Story 7: Operator Running KFP on Storage-Constrained Infrastructure
+
+As an operator with limited storage budget (only 1TB total available), I want to deploy KFP in central mode with a single 500GB PVC shared across 10 team namespaces, so that all teams can run pipelines without each needing their own 100GB PVC (which would require 1TB total).
+
+**Acceptance Criteria:**
+
+- Central mode configured with: `DeploymentMode: "central"`, `Size: "500Gi"`
+- Single PVC created in `kubeflow` namespace mounted by one artifact server
+- All 10 teams' artifacts stored in `/artifacts/<namespace>/` directories on same PVC
+- Teams can run pipelines concurrently without storage allocation failures
+- No per-namespace storage limits (trade-off of central mode - shared PVC means no per-team quotas)
+
+#### Story 8: Operator Scaling High-Throughput Model Training Platform
+
+As an operator supporting 100+ concurrent pipeline runs with multi-GB model checkpoints, I want to deploy KFP in central mode with RWX storage (e.g., NFS/CephFS) and multiple artifact server replicas behind a load balancer, so that artifact upload/download operations can scale horizontally without bottlenecks.
+
+**Acceptance Criteria:**
+
+- Can deploy artifact servers in both central and namespace-local modes
+- Artifact servers stream large files without loading into memory
+- Can use high-performance `StorageClasses` (e.g., SSD-backed)
+- Horizontal scaling possible in central mode
+- Direct pod-to-pod communication in namespace-local mode reduces latency
+
+### Notes/Constraints/Caveats
+
+1. **Performance**: filesystem storage may have different performance characteristics than object storage, especially for large files or high concurrency
+2. **Storage Classes**: Functionality depends on available `StorageClasses` - RWX preferred for parallel execution
+3. **Scalability**: filesystem storage is not designed for massive scale like object storage
+4. **Cost**: Cost depends on the storage class and provider - could be more or less expensive than object storage
+5. **Portability**: Pipelines using filesystem storage are tied to Kubernetes infrastructure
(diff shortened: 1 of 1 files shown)


juliusvonkohout:
Here are some thoughts, but I still need to read the full document.

That is very far away from V1 artifact passing via PVC and would violate zero overhead namespaces. Just imagine a cluster with 1000 namespaces for scalability. Then you add 1000 permanent pods, so massive overhead. In V1 we just accepted that the artifacts will not be in the UI instead. PVC for all namespace also sounds scary. Why should we offer such a security nightmare in the first place? That would break the namespace isolation contract.

Maybe I am missing something and I very much admire that you spend the time for the KEP, but there might be some fundamental problems as mentioned above regarding security and scalability that we did not have in the KFP V1 Implementation.

But as I said some of my statements could be wrong and that is just my initial assessment without checking it thoroughly.

What I also do not understand is that by default we ship seaweedfs and you do not have to configure object storage yourself. So where is this making it easier for beginners, i think it is actually more difficult?
I see the main benefit in enterprise environments if you want to have purely per namespace storage/PVCs for faster performance than S3 and per namespace storage quotas and stticter isolation. For everyone else i would recommend to stick to seaweedfs.

hbelmiro:
@juliusvonkohout 

> Here are some thoughts, but I still need to read the full document.
> 
> That is very far away from V1 artifact passing via PVC and would violate zero overhead namespaces. Just imagine a cluster with 1000 namespaces for scalability. Then you add 1000 permanent pods, so massive overhead.

There will be two modes: [central and namespace-local](https://github.com/hbelmiro/data-science-pipelines/tree/pvc-kep/proposals/12513-pvc-artifact-storage#multi-user-isolation-strategy).

Users should choose which mode fits to their scenarios or not even use the filesystem storage if it doesn't make sense to them. The idea is not to replace the existing storage solutions, but add a new alternative.

> In V1 we just accepted that the artifacts will not be in the UI instead. PVC for all namespace also sounds scary. Why should we offer such a security nightmare in the first place? That would break the namespace isolation contract.
>
> Maybe I am missing something and I very much admire that you spend the time for the KEP, but there might be some fundamental problems as mentioned above regarding security and scalability that we did not have in the KFP V1 Implementation.
>
> But as I said some of my statements could be wrong and that is just my initial assessment without checking it thoroughly.

I'm not sure I'm following. Please correct me if I'm wrong, but the existing s3 solutions already break the namespace isolation contract once we have one instance for all namespaces. With the namespace-local mode proposed here we can achieve complete namespace isolation, which we don't have today.

> What I also do not understand is that by default we ship seaweedfs and you do not have to configure object storage yourself. So where is this making it easier for beginners, i think it is actually more difficult? I see the main benefit in enterprise environments if you want to have purely per namespace storage/PVCs for faster performance than S3 and per namespace storage quotas and stticter isolation. For everyone else i would recommend to stick to seaweedfs.

Thanks for clarifying regarding SeaweedFS. I updated the motivation with that in mind. You can see the specific commit [here](https://github.com/kubeflow/pipelines/pull/12515/commits/cc1ccedba5debf870e66ef956910b84174875205).

Maybe taking a look on [Goals](https://github.com/hbelmiro/data-science-pipelines/tree/pvc-kep/proposals/12513-pvc-artifact-storage#goals) and [Non-Goals](https://github.com/hbelmiro/data-science-pipelines/tree/pvc-kep/proposals/12513-pvc-artifact-storage#non-goals) may clarify some things before going deep into the proposal.

juliusvonkohout:
> I'm not sure I'm following. Please correct me if I'm wrong, but the existing s3 solutions already break the namespace isolation contract once we have one instance for all namespaces. With the namespace-local mode proposed here we can achieve complete namespace isolation, which we don't have today.

That is not the case anymore. Seaweedfs is now multi-tenant with ACLS and credentials per namespace as of KFP release 2.15. We even have tests that verify the namespace isolation of seaweedfs. Therefore i doubt that the central mode is needed at all. It just adds complexity and decreases security.

proposals/12513-pvc-artifact-storage/README.md
```diff
@@ -0,0 +1,1867 @@
...
+#### Story 1: User Running Pipelines on Local Kubernetes
+
+As a user with KFP on kind/minikube/k3s, I want my pipeline artifacts to automatically use the local cluster's default `StorageClass` via the `kfp-artifacts://` scheme, so that I can develop pipelines offline without any storage configuration.
+
+**Acceptance Criteria:**
+
+- KFP works out-of-the-box with filesystem storage on local clusters
+- Artifacts are stored using the `kfp-artifacts://` URI scheme
+- No S3/GCS credentials required
+- Artifact viewing in UI works seamlessly
```
juliusvonkohout:
```suggestion
```

That is already covered by the default seaweedfs with zero effort. No storage configuration needed. seaweedfs just uses the default `StorageClass`. It already satisfies
- KFP works out-of-the-box with filesystem storage on local clusters
- No S3/GCS credentials required
- Artifact viewing in UI works seamlessly

So i recommend to remove this story

proposals/12513-pvc-artifact-storage/README.md
```diff
@@ -0,0 +1,1867 @@
...
+#### Story 2: Operator Deploying KFP Without External Object Storage
+
+As an operator for a Kubeflow distribution, I want to deploy KFP with filesystem storage so that I don't need to productize and support a separate object storage system.
+
+**Acceptance Criteria:**
+
+- Single configuration option to enable filesystem storage
+- Artifact handling is part of KFP (no separate object storage component)
+- Storage automatically provisioned via PVCs
+- Backup/restore follows standard Kubernetes PVC procedures
```
juliusvonkohout:
```suggestion
```

That is already covered by the default seaweedfs with zero effort. No storage configuration needed. seaweedfs just uses the default `StorageClass`. It already satisfies
- Artifact handling is part of KFP (no separate object storage component)
- Storage automatically provisioned via PVCs
- Backup/restore follows standard Kubernetes PVC procedures

So i recommend to remove this story

proposals/12513-pvc-artifact-storage/README.md
```diff
@@ -0,0 +1,1867 @@
...
+
+As an operator, I want to configure KFP to use a specific StorageClass and PVC size instead of defaults, so that I can match storage performance and capacity to my workload requirements.
+
+**Acceptance Criteria:**
+
+- Can specify `StorageClass` in KFP configuration
+- Can set PVC size limits (global configuration)
+- Storage quotas enforced via Kubernetes `ResourceQuotas`
+- Clear error messages when storage limits are reached
+- Can choose between RWO and RWX access modes based on needs
```
juliusvonkohout:
```suggestion
```

That is already covered by the default seaweedfs with zero effort. You can already decide which storageclass is used for Seaweedfs. It already satisfies
- Can specify `StorageClass` in KFP configuration
- Can set PVC size limits (global configuration)
- Clear error messages when storage limits are reached
- Can choose between RWO and RWX access modes based on needs

So i recommend to remove this story

The only thing remaining is - Storage quotas enforced via Kubernetes `ResourceQuotas`. We can set it per bucket already, but i am not sure about per folder. So it is partially solved and fully solved if you create on bucket per namespace. This item can be moved to another story.

proposals/12513-pvc-artifact-storage/README.md
```diff
@@ -0,0 +1,1867 @@
...
+
+As an operator, I want to deploy KFP in namespace-local mode where each namespace annotated with `pipelines.kubeflow.org/enabled=true` gets its own artifact server pod and dedicated PVC, so that Team A's artifacts in namespace `team-a` are physically isolated from Team B's artifacts in namespace `team-b`.
+
+**Acceptance Criteria:**
+
+- Each namespace with `pipelines.kubeflow.org/enabled=true` annotation gets its own artifact server deployment
+- Each namespace gets its own dedicated PVC (no shared storage)
+- Artifact server in `team-a` namespace cannot access PVC in `team-b` namespace
+- Users can only access artifacts in namespaces they have RBAC permissions for (via `SubjectAccessReview`)
+- Physical isolation verified: deleting `team-a` namespace doesn't affect `team-b`'s artifacts
```
juliusvonkohout:
```suggestion
```

This storay has been made completely obsolete by the multi-tenant default seaweedfs. It would just be more complicated and break zero overhead namespaces. Its all done already in a secure manner and without massive overhead from additinal artifact servers per namespace. Just imagine 1000 namespaces and 1000 extra pods when idle.

proposals/12513-pvc-artifact-storage/README.md
```diff
@@ -0,0 +1,1867 @@
...
+
+As an operator in a regulated environment (e.g., healthcare, finance), I want to deploy KFP with filesystem storage using an encrypted `StorageClass` (e.g., `encrypted-gp3`), so that artifact handling stays within the KFP codebase and I don't need to include a separate object storage system in my security audits.
+
+**Acceptance Criteria:**
+
+- All artifacts stored on PVCs within the cluster
+- KFP configuration uses `Filesystem.Type: "pvc"` with encrypted `StorageClass`
+- `SubjectAccessReview` validates all artifact access requests
+- Encryption at rest provided by the configured `StorageClass` (e.g., `encrypted-gp3`)
+- No separate object storage component to audit
```
juliusvonkohout:
```suggestion
```

This story is also obsolete because you can just make the seaweedfs PVC encrypted.

proposals/12513-pvc-artifact-storage/README.md
```diff
@@ -0,0 +1,1867 @@
...
+
+As an operator with limited storage budget (only 1TB total available), I want to deploy KFP in central mode with a single 500GB PVC shared across 10 team namespaces, so that all teams can run pipelines without each needing their own 100GB PVC (which would require 1TB total).
+
+**Acceptance Criteria:**
+
+- Central mode configured with: `DeploymentMode: "central"`, `Size: "500Gi"`
+- Single PVC created in `kubeflow` namespace mounted by one artifact server
+- All 10 teams' artifacts stored in `/artifacts/<namespace>/` directories on same PVC
+- Teams can run pipelines concurrently without storage allocation failures
+- No per-namespace storage limits (trade-off of central mode - shared PVC means no per-team quotas)
```
juliusvonkohout:
```suggestion
```

That is also just fully covered by the default multi-tenant seaweedfs. Just one single pvc in the kubeflow namespace.

proposals/12513-pvc-artifact-storage/README.md
```diff
@@ -0,0 +1,1867 @@
...
+
+As an operator supporting 100+ concurrent pipeline runs with multi-GB model checkpoints, I want to deploy KFP in central mode with RWX storage (e.g., NFS/CephFS) and multiple artifact server replicas behind a load balancer, so that artifact upload/download operations can scale horizontally without bottlenecks.
+
+**Acceptance Criteria:**
+
+- Can deploy artifact servers in both central and namespace-local modes
+- Artifact servers stream large files without loading into memory
+- Can use high-performance `StorageClasses` (e.g., SSD-backed)
+- Horizontal scaling possible in central mode
+- Direct pod-to-pod communication in namespace-local mode reduces latency
```
juliusvonkohout:
```suggestion
```

That is also obsolete by merging https://github.com/kubeflow/pipelines/pull/12391/, so a distributed seaweedfs.

proposals/12513-pvc-artifact-storage/README.md
```diff
@@ -0,0 +1,1867 @@
...
+As an operator, I want to deploy KFP in central mode by default, but configure specific namespaces (e.g., `team-finance`) to use namespace-local mode for stricter isolation, so that most teams share the simple central server while sensitive teams get dedicated resources.
+
+**Acceptance Criteria:**
+
+- Global deployment mode set to `central` (default)
+- Specific namespaces can override to `namespaced` via their `kfp-launcher` ConfigMap
+- Teams using central mode share the central artifact server
+- Teams with namespace-local override get their own artifact server and PVC
+- UI correctly routes artifact requests based on each namespace's deployment mode
+- No cluster-wide restart needed to change a namespace's mode
```
juliusvonkohout:
```suggestion
#### Story 9: Operator with enterprise isolation requirements and fast local storage for giant models

I want to have as in kfp v1 data_passing_method() a way to have a per pipeline or per namespace PVC that is very fast local storage such that i do not need to up and download large models for each step from and to S3.

```

That is a security nightmare i would not want to support or recommend to any enterprise. It is just worse than the default multi-tenant seaweedfs.  I added a typical usecase i have seen so far.

proposals/12513-pvc-artifact-storage/README.md
```diff
@@ -0,0 +1,1867 @@
...
+
+A single artifact server in the main KFP namespace serves all namespaces, configured via `ObjectStoreConfig.ArtifactServer.DeploymentMode: "central"`.
+
+**Central Mode Characteristics:**
+
+- **Single PVC** with directory structure: `/artifacts/<namespace>/<pipeline>/<run-id>/<node-id>/<artifact-name>`
+- **Authorization**: Uses `SubjectAccessReview` to verify namespace access
+- **Best for**: Simple deployments, single-user setups, small teams
+- **Advantages**: Simple setup, single storage location, easy backup
+- **Limitations**: All namespaces share same PVC and storage quota
```
juliusvonkohout:
```suggestion
```

For the reasons mentioned above and below this mode has been made fully obsolete by the default multi-tenant seaweedfs. Furthermore it is a security nightmare and adds unnecessary complexity. There is zero benefit implementing this.

proposals/12513-pvc-artifact-storage/README.md
```diff
@@ -0,0 +1,1867 @@
...
+Each namespace runs its own artifact server, configured via `ObjectStoreConfig.ArtifactServer.DeploymentMode: "namespaced"`.
+
+**Namespace-Local Mode Characteristics:**
+
+- **PVC per namespace**: Complete storage isolation
+- **Direct access**: Clients connect directly to namespace servers (no proxying)
+- **Authorization**: Natural isolation (each server only accesses its namespace's PVC)
+- **Best for**: Large multi-tenant deployments, strict isolation requirements
+- **Advantages**: True multi-tenancy, per-namespace scaling, independent quotas
+- **Deployment**: Proactive initialization when namespace has `pipelines.kubeflow.org/enabled=true` annotation
```
juliusvonkohout: We already have multi tenancy from the default seaweedfs. The question is how can we avoid the per namespace overhead and not break zero overhead namespaces when idle. This was working well in KFP v1 already. Just copying the kfp v1 data_passing_method() to kfp v2 would be enough.

proposals/12513-pvc-artifact-storage/README.md
```diff
@@ -0,0 +1,1867 @@
...
+```text
+Client
+  │
+  │ GET kfp-artifacts://<namespace>/...
+  ▼
+KFP API Server (central)
+  │
+  │ Authorization check (SubjectAccessReview)
+  ▼
+Serve from /artifacts/<namespace>/...
```
juliusvonkohout:
```suggestion
```

As explained above the central mode is dangerous and worse than the default multi-tenant seaweedfs.

proposals/12513-pvc-artifact-storage/README.md
```diff
@@ -0,0 +1,1867 @@
...
+└──────────────────┬───────────────────┘
+                   │ mounts
+                   ▼
+┌──────────────────────────────────────┐
+│  Central PVC (kfp-artifacts-central) │
+│  /artifacts/                         │
+│    ├── ns1/                          │
+│    ├── ns2/                          │
+│    └── ns3/                          │
+└──────────────────────────────────────┘
```
juliusvonkohout: As explained above the central mode is dangerous and worse than the default multi-tenant seaweedfs.
```suggestion
```

proposals/12513-pvc-artifact-storage/README.md
```diff
@@ -0,0 +1,1867 @@
...
+│              Driver                  │
+│   (validates artifact server exists) │
+└──────────────────┬───────────────────┘
+                   │
+                   ▼
+┌──────────────────────────────────────┐
+│             Launcher                 │
+│   (uploads/downloads via namespace   │
+│    artifact server API)              │
+└──────────────────┬───────────────────┘
```
juliusvonkohout, line 421: This artifact server overhead per namespace (just imagine 1000 namespaces) should be disabled by default.

proposals/12513-pvc-artifact-storage/README.md
```diff
@@ -0,0 +1,1867 @@
...
+
+2. **Reverse scenario is rare**: If an organization chooses `namespaced` as the default, they likely have strong reasons (compliance, security, multi-tenancy). It's unlikely they'd want some namespaces to break that pattern and use shared central storage.
+
+3. **Avoids infrastructure complexity**: If global is `namespaced`, there's no central server deployed. Allowing overrides to `central` raises the question: "Who deploys the central server when the default is namespaced?" This adds unnecessary complexity.
+
+4. **Simple mental model**:
+   - `central` default → shared storage, teams can opt into isolation
+   - `namespaced` default → everyone isolated, no shared option
+
+This keeps the configuration model simple while covering realistic use cases.
```
juliusvonkohout, line 481:
```suggestion
```

removed based on explanations in above comments

proposals/12513-pvc-artifact-storage/README.md
```diff
@@ -0,0 +1,1867 @@
...
+    return `ml-pipeline-artifact-server.${namespace}.svc:8080`;
+  }
+  
+  // Default is central - check if namespace has overridden to namespaced
+  if (config.namespaced_servers?.includes(namespace)) {
+    return `ml-pipeline-artifact-server.${namespace}.svc:8080`;
+  }
+  
+  // Use central server
+  return 'ml-pipeline-artifact-server.kubeflow.svc:8080';
```
juliusvonkohout:
```suggestion
```

removed based on explanations in above comments

proposals/12513-pvc-artifact-storage/README.md
```diff
@@ -0,0 +1,1867 @@
...
+        service := m.buildArtifactServerService(namespace, deploymentName)
+        _, err = m.k8sClient.CoreV1().Services(namespace).
+            Create(context.Background(), service, metav1.CreateOptions{})
+        if err != nil && !errors.IsAlreadyExists(err) {
+            return fmt.Errorf("failed to create service: %w", err)
+        }
+    }
+    
+    return nil
+}
```
juliusvonkohout: That belongs into sync.py where it already is. CC @droctothorpe

proposals/12513-pvc-artifact-storage/README.md
```diff
@@ -0,0 +1,1867 @@
...
+
+// Called during API server startup to initialize artifact management
+func (m *ArtifactServerManager) Start(ctx context.Context) error {
+    // For central mode (default), ensure the central server exists on startup
+    if m.config.DeploymentMode == "central" {
+        if err := m.ensureCentralArtifactServer(); err != nil {
+            return fmt.Errorf("failed to ensure central artifact server: %w", err)
+        }
+        log.Info("Central artifact server verified/created")
+    }
```
juliusvonkohout:
```suggestion
```

proposals/12513-pvc-artifact-storage/README.md
```diff
@@ -0,0 +1,1867 @@
...
+```go
+func (d *Driver) validateArtifactInfrastructure(ctx context.Context, pipelineRoot string) error {
+    if !strings.HasPrefix(pipelineRoot, "kfp-artifacts://") {
+        return nil // Not using filesystem storage
+    }
+    
+    namespace := d.options.Namespace
+    
+    // Get deployment mode from kfp-launcher ConfigMap (or global default)
+    deploymentMode := d.getNamespaceDeploymentMode(ctx, namespace)
```
juliusvonkohout:
```suggestion
```

proposals/12513-pvc-artifact-storage/README.md
```diff
@@ -0,0 +1,1867 @@
...
+        }
+        
+        // Verify service exists
+        _, err = d.k8sClient.CoreV1().Services(namespace).
+            Get(ctx, deploymentName, metav1.GetOptions{})
+        if err != nil {
+            return fmt.Errorf("artifact service not found or inaccessible: %w", err)
+        }
+    }
+    // For central mode, artifact server is pre-deployed in kubeflow namespace
```
juliusvonkohout:
```suggestion
```

proposals/12513-pvc-artifact-storage/README.md
```diff
@@ -0,0 +1,1867 @@
...
+
+#### Multi-User Isolation Strategy
+
+The approach to multi-user isolation depends on the deployment mode:
+
+##### Central Mode Isolation
+
+- Directory-based separation: Artifacts stored in `/artifacts/<namespace>/...` structure
+- Authorization checks: Every request validated using `SubjectAccessReview`
+- Single PVC: Shared storage with logical separation
```
juliusvonkohout:
```suggestion
```

proposals/12513-pvc-artifact-storage/README.md
```diff
@@ -0,0 +1,1867 @@
...
+- UI handles the resolution based on deployment mode
+
+##### Deployment Strategy
+
+1. **Proactive Provisioning**: The KFP API server watches for namespaces with `pipelines.kubeflow.org/enabled=true` annotation and creates:
+   - The artifact server Deployment
+   - The artifact server Service
+   - The namespace PVC (which only the artifact server will mount)
+2. **Lifecycle Management**: Artifact servers persist as long as the namespace exists
+3. **Resource Efficiency**: Minimal resource usage when idle (can scale to zero if configured)
```
juliusvonkohout: That is already in sync.py and should probably stay there. Why do you want to move it to the main api server ? Sync.py is for the pipelines profile controller to manage kfp enabled namespaces.

proposals/12513-pvc-artifact-storage/README.md
```diff
@@ -0,0 +1,1867 @@
...
+
+#### `ArtifactServerManager` (API Server)
+
+- Watches namespaces with `pipelines.kubeflow.org/enabled=true` annotation
+- Creates PVC, Deployment, Service based on resolved deployment mode
+- Reads `artifactServer.deploymentMode` from namespace's `kfp-launcher` ConfigMap, falls back to global default
+- No action for namespaces without the annotation
+- Skips namespaces where infrastructure already exists
+- Idempotent: Re-processing same namespace doesn't create duplicate resources
+- Mixed mode: Different namespaces can use different deployment modes
```
juliusvonkohout: That is already in sync.py and should probably stay there. Why do you want to move it to the main api server ? Sync.py is for the pipelines profile controller to manage kfp enabled namespaces.

proposals/12513-pvc-artifact-storage/README.md
```diff
@@ -0,0 +1,1867 @@
...
+- Verify artifacts stored at correct filesystem paths
+- Verify artifact server API returns correct content
+
+#### Central Mode
+
+- Deploy with `DeploymentMode: "central"`
+- Run pipelines in multiple namespaces
+- Verify single PVC in `kubeflow` namespace contains all artifacts
+- Verify path structure: `/artifacts/<namespace>/<pipeline>/<run-id>/...`
+- Verify `SubjectAccessReview` prevents cross-namespace access
```
juliusvonkohout:
```suggestion
```

proposals/12513-pvc-artifact-storage/README.md
```diff
@@ -0,0 +1,1867 @@
...
+
+#### Mixed Mode
+
+- Deploy with global `DeploymentMode: "central"`
+- Create `team-a` namespace with annotation (no `artifactServer` override in ConfigMap - uses central)
+- Create `team-b` namespace with annotation and `kfp-launcher` ConfigMap containing `artifactServer.deploymentMode: "namespaced"`
+- Verify `team-a` uses central artifact server
+- Verify `team-b` gets its own artifact server deployment and PVC
+- Run pipelines in both namespaces, verify correct routing
+- Update `team-a` ConfigMap to add `artifactServer.deploymentMode: "namespaced"`, verify infrastructure created
```
juliusvonkohout:
```suggestion
```

proposals/12513-pvc-artifact-storage/README.md
```diff
@@ -0,0 +1,1867 @@
...
+| `ObjectStoreConfig.Filesystem.Type`                  | Storage backend type              | -                 | `"pvc"`, `"local"` (testing only)       |
+| `ObjectStoreConfig.Filesystem.MountPath`             | Path where PVC is mounted         | `"/artifacts"`    | Any valid path                          |
+| `ObjectStoreConfig.Filesystem.PVC.StorageClassName`  | K8s StorageClass to use           | Cluster default   | Any available StorageClass              |
+| `ObjectStoreConfig.Filesystem.PVC.Size`              | Size of PVC to create             | `"10Gi"`          | K8s quantity (e.g., `"100Gi"`, `"1Ti"`) |
+| `ObjectStoreConfig.Filesystem.PVC.AccessMode`        | PVC access mode                   | `"ReadWriteOnce"` | `"ReadWriteOnce"`, `"ReadWriteMany"`    |
+| `ObjectStoreConfig.Filesystem.PVC.CreateIfNotExists` | Auto-create PVC if missing        | `true`            | `true`, `false`                         |
+| `ObjectStoreConfig.ArtifactServer.DeploymentMode`    | Default deployment mode           | `"central"`       | `"central"`, `"namespaced"`             |
```
juliusvonkohout:
```suggestion
```

proposals/12513-pvc-artifact-storage/README.md
```diff
@@ -0,0 +1,1867 @@
...
+
+**Notes:**
+
+- `DeploymentMode: "central"`: Single shared artifact server in kubeflow namespace
+- `DeploymentMode: "namespaced"`: Per-namespace artifact servers for isolation
+
+#### Per-Namespace Configuration (kfp-launcher ConfigMap)
+
+When global default is `central`, individual namespaces can override to `namespaced` via their `kfp-launcher` ConfigMap. **Note: Overrides are only supported when global default is `central`.** When global is `namespaced`, all namespaces use their own artifact server (no overrides allowed).
```
juliusvonkohout:
```suggestion
```

proposals/12513-pvc-artifact-storage/README.md
```diff
@@ -0,0 +1,1867 @@
...
+
+| Key                             | Description                                            | Default                  | Valid Values                  |
+|---------------------------------|--------------------------------------------------------|--------------------------|-------------------------------|
+| `defaultPipelineRoot`           | Artifact storage root URI                              | `minio://mlpipeline/...` | `kfp-artifacts://<namespace>` |
+| `artifactServer.deploymentMode` | Override to namespaced (only when global is `central`) | (global default)         | `"namespaced"`                |
```
juliusvonkohout:
```suggestion
```

proposals/12513-pvc-artifact-storage/README.md
```diff
@@ -0,0 +1,1867 @@
...
+  defaultPipelineRoot: "kfp-artifacts://team-a"
+  artifactServer: |
+    deploymentMode: namespaced
+```
+
+**Deployment Mode Scenarios:**
+
+| Global Default | Per-Namespace Override | Result                                   |
+|----------------|------------------------|------------------------------------------|
+| `central`      | (none)                 | All namespaces use central server        |
```
juliusvonkohout:
```suggestion
```

proposals/12513-pvc-artifact-storage/README.md
```diff
@@ -0,0 +1,1867 @@
...
+OBJECTSTORECONFIG_FILESYSTEM_MOUNTPATH=/artifacts
+
+# PVC configuration
+OBJECTSTORECONFIG_FILESYSTEM_PVC_STORAGECLASSNAME=standard
+OBJECTSTORECONFIG_FILESYSTEM_PVC_SIZE=100Gi
+OBJECTSTORECONFIG_FILESYSTEM_PVC_ACCESSMODE=ReadWriteMany
+OBJECTSTORECONFIG_FILESYSTEM_PVC_CREATEIFNOTEXISTS=true
+
+# Artifact server configuration
+OBJECTSTORECONFIG_ARTIFACTSERVER_DEPLOYMENTMODE=central
```
juliusvonkohout:
```suggestion
```

juliusvonkohout:
Hello, i hope to have added enough clarification and what the status quo is since the KFP 2.15.0 release. What i still see left as usecase is "I want to have as in kfp v1 data_passing_method() a way to have a per pipeline or per namespace PVC that is very fast local storage such that i do not need to up and download large models for each step from and to S3."

```
    @data_passing_method.setter
    def data_passing_method(self, value):
        """Sets the object representing the method used for intermediate data
        passing.

        Example:
          ::

            from kfp.dsl import PipelineConf, data_passing_methods
            from kubernetes.client.models import V1Volume, V1PersistentVolumeClaimVolumeSource
            pipeline_conf = PipelineConf()
            pipeline_conf.data_passing_method =
            data_passing_methods.KubernetesVolume(
                volume=V1Volume(
                    name='data',
                    persistent_volume_claim=V1PersistentVolumeClaimVolumeSource('data-volume'),
                ),
                path_prefix='artifact_data/',
            )
        """
        self._data_passing_method = value
```

See https://kubeflow-pipelines.readthedocs.io/en/1.8.22/_modules/kfp/dsl/_pipeline.html it works very well in KFP v1.

hbelmiro:
@juliusvonkohout 

> Seaweedfs is now multi-tenant with ACLS and credentials per namespace as of KFP release 2.15. We even have tests that verify the namespace isolation of seaweedfs. 

> Just imagine 1000 namespaces and 1000 extra pods when idle.

Are you saying that SeaweedFS uses one PVC per namespace (like the namespaced-local mode in the proposal)?
I'm curious how it workarounds the physical limitation between the api server and the PVCs on different namespaces without an additional pod between them.

juliusvonkohout:
> @juliusvonkohout
> 
> > Seaweedfs is now multi-tenant with ACLS and credentials per namespace as of KFP release 2.15. We even have tests that verify the namespace isolation of seaweedfs.
> 
> > Just imagine 1000 namespaces and 1000 extra pods when idle.
> 
> Are you saying that SeaweedFS uses one PVC per namespace (like the namespaced-local mode in the proposal)? I'm curious how it workarounds the physical limitation between the api server and the PVCs on different namespaces without an additional pod between them.

No, it uses a hard multi-tenant separated S3 storage that is directly accessed by the ml-pipeline-ui by default. no proxy needed. See also https://github.com/kubeflow/manifests#architecture for my diagram

grafik


This is also in the new release and blogposts
1. https://github.com/kubeflow/manifests/releases/tag/v1.11.0-rc.1 check the highlights
2. https://blog.kubeflow.org/gsoc/community/kubeflow/2025/09/06/kubeflow-and-gsoc2025.html#project-1-kubeflow-platform-enhancements
3. https://medium.com/@hpotpose26/kubeflow-pipelines-embraces-seaweedfs-9a7e022d5571 here are technical details, each namespace has its own S3 user and ml-pipeline-ui has administrative access. Read access would also be enough.

proposals/12513-pvc-artifact-storage/README.md
```diff
@@ -0,0 +1,1867 @@
...
+This KEP proposes adding filesystem-based storage as an alternative artifact storage backend for Kubeflow Pipelines v2. While KFP currently ships with S3-compatible storage by default, some deployments prefer not to depend on a separate object storage system. This proposal introduces filesystem storage as an additional option where artifact handling is integrated into KFP itself, eliminating the need for an external object storage component.
+
+The filesystem backend will primarily use `PersistentVolumeClaim` (PVC) based storage in Kubernetes environments, providing namespace-isolated storage using Kubernetes native `PersistentVolumes`. However, the design is flexible enough to support other filesystem backends (e.g., local filesystem for development). Users can specify any access mode value which KFP will pass through to Kubernetes without validation (e.g., `ReadWriteMany` for parallel task execution across nodes, `ReadWriteOnce` for single node access, etc.), with RWO as the default if not specified. The actual behavior depends on what the underlying storage class supports. Existing pipelines will work without modification unless they contain hardcoded S3/object storage paths.
```
mprahl: This paragraph implies the PVCs are mounted to the pods "which KFP will pass through to Kubernetes without validation". The following paragraph contradicts that.

proposals/12513-pvc-artifact-storage/README.md
```diff
@@ -0,0 +1,1867 @@
...
+
+### Quick Comparison
+
+| Aspect                 | Current (S3-compatible Storage)                 | Proposed (Filesystem Storage)                     |
+|------------------------|-------------------------------------------------|---------------------------------------------------|
+| **Storage Backend**    | S3-compatible (SeaweedFS, MinIO, S3, GCS, etc.) | Kubernetes PVC                                    |
+| **URI Scheme**         | `s3://`, `gs://`, `minio://`                    | `kfp-artifacts://`                                |
+| **Architecture**       | Separate object storage service                 | KFP-native artifact server                        |
+| **Required Knowledge** | S3 concepts (buckets, endpoints, regions)       | Kubernetes concepts (PVCs, StorageClasses)        |
+| **Multi-tenancy**      | Shared storage (single instance)                | Per-namespace PVCs (in namespace-local mode)      |
```
mprahl: Per-namespace PVCs should be optional. In other words, default to the central artifact server, respecting multi-tenancy, but allow individual namespaces to use a different solution (e.g. S3 or another artifact server).

proposals/12513-pvc-artifact-storage/README.md
```diff
@@ -0,0 +1,1867 @@
...
+### High-Level Flow
+
+```text
+Pipeline Submission
+        │
+        ▼
+Driver (detects "kfp-artifacts://")
+        │
+        ▼
+Artifact Server (mounts PVC)
```
mprahl: Wouldn't the PVC always be mounted?

proposals/12513-pvc-artifact-storage/README.md
```diff
@@ -0,0 +1,1867 @@
...
+        └──────────► Artifact Server ◄─────┘
+                           │
+                           ▼
+                    Read/Write to PVC
+```
+
+### Key Components
+
+- **Storage**: Kubernetes `PersistentVolumeClaims` (one per namespace)
+- **URI Format**: `kfp-artifacts://<namespace>/<pipeline>/<run-id>/<node-id>/<artifact-name>`
```
mprahl: Could you be more specific on `<pipeline>`? Is it a name or ID? Is it a pipeline or pipeline version?

Also, what is <node-id>?

proposals/12513-pvc-artifact-storage/README.md
```diff
@@ -0,0 +1,1867 @@
...
+#### UI Integration
+
+The UI discovers artifact server routing through the KFP API server, which acts as a registry. The API server already watches namespaces and reads their `kfp-launcher` ConfigMaps to determine deployment mode, so it can derive the routing information without additional infrastructure.
+
+##### Filesystem Storage Configuration Endpoint
+
+A dedicated endpoint provides filesystem storage configuration and routing:
+
+```http
+GET /apis/v2beta1/filesystem-storage/config
```
mprahl, line 450: I don't think we need this. The artifact URI should contain the hostname of the artifact server that was used and it can be parsed directly to know how to route the request.

proposals/12513-pvc-artifact-storage/README.md
```diff
@@ -0,0 +1,1867 @@
...
+- **Deployment Modes**:
+  - **Central**: Single artifact server for all namespaces (default)
+  - **Namespace-local**: One artifact server per namespace (better isolation)
+  - **Mixed**: Per-namespace mode override via `kfp-launcher` ConfigMap
+
+## Motivation
+
+**This proposal does not aim to replace existing object storage solutions.** S3-compatible storage remains fully supported and recommended for most production workloads. Instead, this KEP provides an additional option for deployments where a simpler, KFP-native artifact storage solution is preferred.
+
+While KFP currently ships with S3-compatible storage by default, this still requires deploying and maintaining a separate object storage service. For some deployment scenarios, this additional component may not be desired.
```
mprahl, line 105: This also helps with running KFP locally, outside of Kubernetes.

proposals/12513-pvc-artifact-storage/README.md
```diff
@@ -0,0 +1,1867 @@
...
+
+### Reduced External Dependencies
```
mprahl: I don't think this section adds a lot and could be consolidated in the Motivation section.

proposals/12513-pvc-artifact-storage/README.md
```diff
@@ -0,0 +1,1867 @@
...
+- No external object storage project to track for updates and security patches
+- Artifact issues debugged within KFP ecosystem (no separate S3 API layer)
+- Artifact handling uses the same codebase as the rest of KFP
+- No S3-specific knowledge required (buckets, endpoints, regions). Only familiar Kubernetes concepts (PVCs, `StorageClasses`)
+
+### Enterprise Considerations
+
+Many enterprises and Kubeflow distributions prefer not to have additional external dependencies. With filesystem storage:
+
+- No separate object storage project to productize and support
```
mprahl, line 113: Also include that this will directly leverage Kubernetes RBAC, aligned with existing permission mechanisms used by other parts of KFP. This makes onboarding and provisioning new namespaces simpler.

proposals/12513-pvc-artifact-storage/README.md
```diff
@@ -0,0 +1,1867 @@
...
+While KFP's default manifests ship with S3-compatible storage pre-configured (no manual credential setup required), filesystem storage replaces the object storage system with a KFP-native artifact server. This means:
+
+- No external object storage project to track for updates and security patches
+- Artifact issues debugged within KFP ecosystem (no separate S3 API layer)
+- Artifact handling uses the same codebase as the rest of KFP
+- No S3-specific knowledge required (buckets, endpoints, regions). Only familiar Kubernetes concepts (PVCs, `StorageClasses`)
+
+### Enterprise Considerations
+
+Many enterprises and Kubeflow distributions prefer not to have additional external dependencies. With filesystem storage:
```
mprahl, line 111: Another aspect is you'll automatically get namespace isolation of artifacts on the central artifact server through namespace aware paths and Kubernetes RBAC.

proposals/12513-pvc-artifact-storage/README.md
```diff
@@ -0,0 +1,1867 @@
...
+- No separate object storage project to productize and support
+- Artifact handling is part of the KFP codebase (same team, same release cycle)
+- Troubleshooting stays within the KFP domain
+
+### Per-Namespace Isolation and Scaling
+
+In namespace-local mode, each namespace gets its own dedicated artifact server and PVC. This provides:
+
+- **Storage isolation**: Each team's artifacts are physically separated in their own PVC
+- **Independent scaling**: Teams can scale their artifact server horizontally (with RWX storage) and size their PVC based on workload requirements
```
mprahl:
```suggestion
- **Independent scaling**: Teams can scale their artifact server horizontally and size their PVC based on workload requirements
```

proposals/12513-pvc-artifact-storage/README.md
```diff
@@ -0,0 +1,1867 @@
...
+
+Many enterprises and Kubeflow distributions prefer not to have additional external dependencies. With filesystem storage:
+
+- No separate object storage project to productize and support
+- Artifact handling is part of the KFP codebase (same team, same release cycle)
+- Troubleshooting stays within the KFP domain
+
+### Per-Namespace Isolation and Scaling
+
+In namespace-local mode, each namespace gets its own dedicated artifact server and PVC. This provides:
```
mprahl: I don't think we should have a specific mode.

Essentially, the default behavior when using the KFP artifact server would be to use the central/default instance, but every namespace can override this configuration using the `kfp-launcher` ConfigMap or equivalent.

proposals/12513-pvc-artifact-storage/README.md
```diff
@@ -0,0 +1,1867 @@
...
+
+By providing PVC-based storage as an alternative, we:
+
+- Offer a KFP-native artifact storage option (no separate storage system)
+- Enable portability between different storage backends for pipelines using KFP's artifact APIs
+
+**When to use filesystem storage:**
+
+- Deployments where eliminating object storage dependency is preferred
+- Environments where Kubeflow distributions or platform providers prefer not to support additional storage systems
```
mprahl:
```suggestion
- Environments where Kubeflow distributions or platform providers do not offer a fully supported object store solution
```

proposals/12513-pvc-artifact-storage/README.md
```diff
@@ -0,0 +1,1867 @@
...
+- Multi-tenant deployments requiring per-namespace storage isolation, scaling, and quotas
```
mprahl: I think this can all be achieved through S3.

proposals/12513-pvc-artifact-storage/README.md
```diff
@@ -0,0 +1,1867 @@
...
+
+- High-throughput production workloads with large artifacts
+- Scenarios requiring object storage features (versioning, lifecycle policies, geo-replication)
+- When existing object storage infrastructure is already available and preferred
+
+### Goals
+
+Based on the user story "As a user, I want to provision Kubeflow Pipelines with just a PVC for artifact storage so that I can quickly get started", this KEP aims to:
+
+1. **Add filesystem storage as an additional backend option** alongside S3-compatible and Google Cloud Storage, primarily using PVC but not limited to it
```
mprahl, line 153: What do you mean by "but not limited to it"?

proposals/12513-pvc-artifact-storage/README.md
```diff
@@ -0,0 +1,1867 @@
...
+2. **Enable zero-configuration storage** for experimentation use cases - a KFP server can be installed with just a PVC for artifact storage
+3. **Provide namespace-isolated artifact storage** with proper subject access review guards in multi-user mode
+4. **Allow any Kubernetes access mode to be configured** - KFP passes through the configuration to Kubernetes (RWO default)
```
mprahl: What does this mean?

proposals/12513-pvc-artifact-storage/README.md
```diff
@@ -0,0 +1,1867 @@
...
+5. **Support existing pipelines** that use KFP's standard artifact types (Dataset, Model, etc.) - pipelines work unchanged with the new filesystem backend
```
mprahl: I think we can add a Non-Goal of migrating existing installations for now. This is primarily geared towards users that don't already have an object store solution.

proposals/12513-pvc-artifact-storage/README.md
```diff
@@ -0,0 +1,1867 @@
...
+6. **Match existing artifact persistence behavior** - artifacts persist indefinitely until explicitly deleted (no automatic cleanup)
+7. **Enable separate scaling of artifact serving** through an artifacts-only KFP instance with `--artifacts-only` flag
```
mprahl, line 159: We should also support deploying the artifact server as a DaemonSet (ensuring a pod is on every Kubernetes node) and set the Service to `internalTrafficPolicy: Local` to keep all artifact traffic local to the Kubernetes node.

proposals/12513-pvc-artifact-storage/README.md
```diff
@@ -0,0 +1,1867 @@
...
+4. **Storage migration tools** - Migrating from S3 to PVC/filesystem is not supported
+5. **Advanced storage features** - No versioning, replication, or geo-distribution initially
+
+## Proposal
+
+### Overview
+
+This KEP proposes adding a new artifact storage backend that uses filesystem storage (primarily Kubernetes `PersistentVolumeClaims`) instead of object storage. The implementation will:
+
+1. Create one PVC per namespace for artifact storage
```
mprahl: We should not require this as the central artifact server should be namespace aware (namespace in the path and subject access review based on the namespace in the path).

proposals/12513-pvc-artifact-storage/README.md
```diff
@@ -0,0 +1,1867 @@
...
+2. Use configurable access mode with sensible defaults (RWO)
```
mprahl: What does this mean?

proposals/12513-pvc-artifact-storage/README.md
```diff
@@ -0,0 +1,1867 @@
...
+3. Organize artifacts in a filesystem hierarchy within the PVC
```
mprahl:
```suggestion
3. Organize artifacts in a filesystem hierarchy within the PVC that is namespace aware
```

proposals/12513-pvc-artifact-storage/README.md
```diff
@@ -0,0 +1,1867 @@
...
+4. Provide transparent access through the existing KFP artifact APIs with new `kfp-artifacts://` URI scheme
```
mprahl: Do we have "KFP artifact APIs"? Perhaps this is referencing to @HumairAK 's MLMD removal PR.

proposals/12513-pvc-artifact-storage/README.md
```diff
@@ -0,0 +1,1867 @@
...
+5. Maintain compatibility with existing pipeline definitions that don't have hardcoded storage paths
```
mprahl: When is this ever the case?

proposals/12513-pvc-artifact-storage/README.md
mprahl: I agree. This user story is already covered. The KEP just proposes a new way to do this.

proposals/12513-pvc-artifact-storage/README.md
mprahl: I think this user story still applies as it's about the admin not wanting to maintain an object storage service, which is valid for use cases where the organization wants on-premise storage (no AWS access) but doesn't have an enterprise license/support for SeaweedFS.

proposals/12513-pvc-artifact-storage/README.md
```diff
@@ -0,0 +1,1867 @@
...
+**Acceptance Criteria:**
+
+- Single configuration option to enable filesystem storage
+- Artifact handling is part of KFP (no separate object storage component)
+- Storage automatically provisioned via PVCs
+- Backup/restore follows standard Kubernetes PVC procedures
+
+#### Story 3: Operator Configuring Storage Class and Size
+
+As an operator, I want to configure KFP to use a specific StorageClass and PVC size instead of defaults, so that I can match storage performance and capacity to my workload requirements.
```
mprahl: This is implied whenever you use a PVC, so if the user story is about making artifact storage configurable per namespace (e.g. use a dedicated artifact server or use S3 in this namespace), then I think this can be removed.

proposals/12513-pvc-artifact-storage/README.md
```diff
@@ -0,0 +1,1867 @@
...
+
+- Can specify `StorageClass` in KFP configuration
+- Can set PVC size limits (global configuration)
+- Storage quotas enforced via Kubernetes `ResourceQuotas`
+- Clear error messages when storage limits are reached
+- Can choose between RWO and RWX access modes based on needs
+
+#### Story 4: User Migrating from S3 to Filesystem Storage
+
+As a user with existing pipelines containing components that call `boto3.upload_file()` directly, I want KFP system artifacts to use `kfp-artifacts://` with PVC storage while my custom components continue accessing S3, so that I can migrate incrementally without rewriting all components at once.
```
mprahl: I don't think this user story is relevant. That's just a user's custom Python code.

proposals/12513-pvc-artifact-storage/README.md
```diff
@@ -0,0 +1,1867 @@
...
+**Acceptance Criteria:**
+
+- KFP artifacts use filesystem storage via `kfp-artifacts://` URIs
+- Components using boto3/gsutil for S3 continue to work
+- Can mix storage backends (system artifacts on PVC, custom data on S3)
+- Clear documentation on what uses filesystem vs object storage
+
+#### Story 5: Operator Deploying Multi-Tenant KFP with Namespace Isolation
+
+As an operator, I want to deploy KFP in namespace-local mode where each namespace annotated with `pipelines.kubeflow.org/enabled=true` gets its own artifact server pod and dedicated PVC, so that Team A's artifacts in namespace `team-a` are physically isolated from Team B's artifacts in namespace `team-b`.
```
mprahl: Like I said, we don't need the concept of modes. We just need a central artifact server and allow each namespace to override to use a different solution (e.g. dedicated artifact server or S3) through the `kfp-launcher` ConfigMap.

An admin can still opt in to this by configuring each provisioned namespace, but this largely defeats the spirit of the KEP of making administration easier (and potentially cheaper if enterprise licensing is required) with less components involved.

proposals/12513-pvc-artifact-storage/README.md
mprahl: I agree. I think we can remove this.

proposals/12513-pvc-artifact-storage/README.md
```diff
@@ -0,0 +1,1867 @@
...
+
+1. Create one PVC per namespace for artifact storage
+2. Use configurable access mode with sensible defaults (RWO)
+3. Organize artifacts in a filesystem hierarchy within the PVC
+4. Provide transparent access through the existing KFP artifact APIs with new `kfp-artifacts://` URI scheme
+5. Maintain compatibility with existing pipeline definitions that don't have hardcoded storage paths
+6. Support separate scaling of artifact serving through artifacts-only instances
+7. Update the UI to seamlessly handle artifact downloads from filesystem storage
+
+### User Stories
```
mprahl, line 184: I think we only need 2 user stories:

1. As an admin with a requirement to have storage on-premise, I want a simple artifact storage solution for KFP artifacts without having to maintain a separate service for artifacts due to administrative overhead or enterprise licensing costs.
2. As an admin using the KFP artifact server, I want the ability to override a namespace's artifact configuration to use alternative storage such as S3 or a dedicated artifact server.

hbelmiro:
> No, it uses a hard multi-tenant separated S3 storage that is directly accessed by the ml-pipeline-ui by default. no proxy needed. See also https://github.com/kubeflow/manifests#architecture for my diagram
> grafik
> 
> This is also in the new release and blogposts
> 
>     1. https://github.com/kubeflow/manifests/releases/tag/v1.11.0-rc.1 check the highlights
> 
>     2. https://blog.kubeflow.org/gsoc/community/kubeflow/2025/09/06/kubeflow-and-gsoc2025.html#project-1-kubeflow-platform-enhancements
> 
>     3. https://medium.com/@hpotpose26/kubeflow-pipelines-embraces-seaweedfs-9a7e022d5571 here are technical details, each namespace has its own S3 user and ml-pipeline-ui has administrative access. Read access would also be enough.

Thank you for clarifying that @juliusvonkohout.

Yes, it’s hard-enforced multi-tenancy, but it’s still a shared underlying volume/bucket; not per-namespace PVC physical isolation (like the opt-in namespace-local mode).

> Therefore i doubt that the central mode is needed at all. It just adds complexity and decreases security.

@juliusvonkohout I don't see how it decreases security. The idea is to have logical isolation (in central mode - default), just like how we're doing with SeaweedFS.

proposals/12513-pvc-artifact-storage/README.md
hbelmiro: Yes, it would. Why do you think it wouldn't?

proposals/12513-pvc-artifact-storage/README.md
hbelmiro: It's the pipeline name. I'll fix it.

> Also, what is ?

I think there's a word missing in your sentence.

proposals/12513-pvc-artifact-storage/README.md
hbelmiro: The URI being tied to a specific hostname would reduce portability. If the hostname changes for any reason it will be impossible to reach that artifact. I'm not sure how often (if ever) this happens though.

proposals/12513-pvc-artifact-storage/README.md
hbelmiro: That we can extend to local development (outside of Kubernetes) in the future.

proposals/12513-pvc-artifact-storage/README.md
hbelmiro: It means KFP won’t interpret or restrict the access mode value. It will forward the exact string the operator configured into the PVC spec.

proposals/12513-pvc-artifact-storage/README.md
hbelmiro: There are two different kinds of "support" here and I agree we should be explicit.

1) Pipeline compatibility (what Goal # 5 means): pipelines/components that use KFP-managed artifact passing (e.g. `Input/Output[Dataset|Model]` and `.path`) should run unchanged when the pipeline root uses `kfp-artifacts://....` There’s no pipeline-code migration needed as long as users aren’t hardcoding `s3://...`/`gs://...` URIs or calling object-store SDKs directly.

2) Installation/data migration (what you’re suggesting as a Non-Goal): switching an existing KFP installation from S3-compatible storage to filesystem storage does not migrate previously produced artifacts or rewrite stored artifact URIs. Old runs keep pointing to the original object store (`s3://...`, `minio://...`, etc.). Providing migration tooling is out of scope for this KEP.

I’ll update the KEP to distinguish these clearly (pipeline compatibility vs migrating existing installations/artifacts).

proposals/12513-pvc-artifact-storage/README.md
hbelmiro: It means the artifact server’s PVC access mode is configurable, but defaults to `ReadWriteOnce` if you don’t set it.
I'll update this bullet to make it clearer.

proposals/12513-pvc-artifact-storage/README.md
hbelmiro: Yes. KFP has artifact access endpoints in the API server today.

```text
GET /apis/v2beta1/runs/{run_id}/nodes/{node_id}/artifacts/{artifact_name}:read
GET /apis/v1beta1/runs/{run_id}/nodes/{node_id}/artifacts/{artifact_name}:read
```

These are already used by KFP components. The KEP proposes adding the corresponding :write endpoint for filesystem storage.
I'll rephrase the item to make it clearer.

proposals/12513-pvc-artifact-storage/README.md
hbelmiro: This is about pipeline compatibility, not migrating an existing installation’s artifacts.
Pipelines that use KFP-managed artifact passing in the Python DSL (e.g. `Input/Output[Dataset|Model]`, `OutputPath`, and reading/writing via `.path`) don’t hardcode any storage URIs, so they can run unchanged when the pipeline root switches to `kfp-artifacts://...`.
Pipelines only break if they embed storage-specific URIs/paths (s3://, gs://, minio://) or call object-store SDKs directly.
This does not mean old runs/artifacts are migrated or their stored URIs rewritten.

I'll rephrase that item.

juliusvonkohout:
I think we first have to reduce the size of this KEP by at least 50 % by removing now obsolete usecases. At this size it is difficult to review. So cutting it down to the bare minimum first and then expanding if needed.

proposals/12513-pvc-artifact-storage/README.md
hbelmiro: Done.

proposals/12513-pvc-artifact-storage/README.md
hbelmiro: Done.

proposals/12513-pvc-artifact-storage/README.md
hbelmiro: Done.

proposals/12513-pvc-artifact-storage/README.md
hbelmiro: Done.

proposals/12513-pvc-artifact-storage/README.md
hbelmiro: Done.

proposals/12513-pvc-artifact-storage/README.md
hbelmiro: Done.

proposals/12513-pvc-artifact-storage/README.md
hbelmiro: Done.

proposals/12513-pvc-artifact-storage/README.md
hbelmiro: Done.

proposals/12513-pvc-artifact-storage/README.md
hbelmiro: Done.

proposals/12513-pvc-artifact-storage/README.md
hbelmiro: Done.

proposals/12513-pvc-artifact-storage/README.md
hbelmiro: Done.

proposals/12513-pvc-artifact-storage/README.md
hbelmiro: Done.

proposals/12513-pvc-artifact-storage/README.md
hbelmiro: Done.

proposals/12513-pvc-artifact-storage/README.md
hbelmiro: Done.

proposals/12513-pvc-artifact-storage/README.md
hbelmiro: Done.

proposals/12513-pvc-artifact-storage/README.md
hbelmiro: Done.

proposals/12513-pvc-artifact-storage/README.md
hbelmiro: Done.

proposals/12513-pvc-artifact-storage/README.md
hbelmiro: Done.

proposals/12513-pvc-artifact-storage/README.md
hbelmiro: Done.

proposals/12513-pvc-artifact-storage/README.md
hbelmiro: Done.

proposals/12513-pvc-artifact-storage/README.md
```diff
@@ -0,0 +1,1028 @@
...
+
+The driver validates that required artifact infrastructure exists before pipeline execution:
+
+- If not using `kfp-artifacts://`, no filesystem checks are needed.
+- If the namespace opted into a dedicated artifact server, verify the namespace-local Service/workload exist.
+- Otherwise, the shared artifact server is used.
+
+### Security and RBAC
+
+#### Permission Model
```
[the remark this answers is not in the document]
hbelmiro: Duplicate comment.

proposals/12513-pvc-artifact-storage/README.md
```diff
@@ -0,0 +1,1028 @@
...
+**Driver/Workflow Pod Permissions:**
+
+- Read-only access to verify the artifact server workload/Service exists when validating filesystem storage prerequisites.
+
+**Artifact Server Pod Permissions:**
+
+- Create `SubjectAccessReview` resources for authorization checks.
+
+This keeps infrastructure management in the control plane, while workflow pods only perform read-only verification.
+
```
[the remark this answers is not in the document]
hbelmiro: Done.

proposals/12513-pvc-artifact-storage/README.md
```diff
@@ -0,0 +1,1028 @@
...
+- Deploy KFP with filesystem storage enabled (shared default)
+- Configure namespaces to exercise overrides:
+  - `team-a`: no override (uses shared artifact server)
+  - `team-b`: `artifactServer.dedicated: true` (dedicated artifact server + PVC)
+  - `team-c`: override `defaultPipelineRoot` to an S3-compatible scheme (e.g., `s3://...` or `minio://...`)
+- Verify:
+  - KFP routes artifact operations correctly per namespace (shared vs dedicated vs S3-compatible)
+  - enabling/disabling a namespace override does not require a cluster-wide restart
+  - when `WorkloadKind: "daemonset"` is configured, the expected Kubernetes resources are deployed (DaemonSet and Service with `internalTrafficPolicy: Local`)
```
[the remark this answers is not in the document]
hbelmiro: DOne.

proposals/12513-pvc-artifact-storage/README.md
```diff
@@ -0,0 +1,1028 @@
...
+- **Deployment**: Proactive initialization when namespace has `pipelines.kubeflow.org/enabled=true` annotation
+
+##### Per-Namespace Configuration Support
+
+For multi-tenant deployments with varying isolation requirements, administrators can configure namespace-local artifact servers per namespace using the `kfp-launcher` ConfigMap:
+
+Configure the namespace `kfp-launcher` ConfigMap to:
+
+- set `defaultPipelineRoot` to `kfp-artifacts://<namespace>`, and
+- set `artifactServer.dedicated: true` (within the `artifactServer` YAML value) to opt the namespace into a dedicated artifact server + PVC.
```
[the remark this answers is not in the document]
hbelmiro: Done.

proposals/12513-pvc-artifact-storage/README.md
```diff
@@ -0,0 +1,1028 @@
...
+{
+  "ObjectStoreConfig": {
+    "Filesystem": {
+      "Type": "pvc",
+      "MountPath": "/artifacts",
+      "PVC": {
+        "StorageClassName": "standard",
+        "Size": "100Gi",
+        "AccessMode": "ReadWriteMany",
+        "CreateIfNotExists": true
```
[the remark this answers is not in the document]
hbelmiro: Done.

proposals/12513-pvc-artifact-storage/README.md
```diff
@@ -0,0 +1,1028 @@
...
+}
+```
+
+### Configuration Field Reference
+
+| Field                                                | Description                       | Default           | Valid Values                            |
+|------------------------------------------------------|-----------------------------------|-------------------|-----------------------------------------|
+| `ObjectStoreConfig.Filesystem.Type`                  | Storage backend type              | -                 | `"pvc"`, `"local"` (testing only)       |
+| `ObjectStoreConfig.Filesystem.MountPath`             | Path where PVC is mounted         | `"/artifacts"`    | Any valid path                          |
+| `ObjectStoreConfig.Filesystem.PVC.StorageClassName`  | K8s StorageClass to use           | Cluster default   | Any available StorageClass              |
```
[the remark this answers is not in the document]
hbelmiro: That config was not supposed to exist.

juliusvonkohout:
I will try to review it again after my vacation. In the mean time https://github.com/kubeflow/pipelines/pull/12550 is also a very interesting related read.

juliusvonkohout:
I still think that the KEP is way to complex (significant maintenance burden) for almost zero gain. Also the daemonset topic would rather be solved by something like https://github.com/kubeflow/pipelines/pull/12391 in a way cleaner way. So for me only the userstory of adding back kfp v1 data_passing_method() in a minimal non-intrusive, no proxy pod overhead way makes sense. For example with node local storage PVC for very fast performance. Maybe with a KEP of 250 instead of 1000 lines. But we could also have a dedicated architectural call on this.

PR #12519, opened 2025-12-03
feat(backend): double input size limit. Fixes #12510

### LINKED ISSUE #2286: Increase pipeline input argument length/size
There is a limitation on backend server, which triggers when input argument reaches certain length:

`The pipeline spec is invalid.: Invalid input error: The input parameter length exceed maximum size of 10000."`

In some cases it might be necessary to provide large input arguments, when, for example, you got a highly configurable model or a very diversified dataset that requires detailed description.

Are there any practical limitations on input argument length and what's your opinion on increasing input argument length?

### LINKED ISSUE #4828: Increase input parameter length limitation
Is there any plan to change/increase the artificial limitation set on parameter length for the pipeline spec? I find myself frequently running into the following error:

`"The pipeline spec is invalid.: Invalid input error: The input parameter length exceed maximum size of 10000."`

We have limited the individual arguments to the components as either strings, small lists or dicts as suggested in [Data passing tutorial](https://github.com/kubeflow/pipelines/blob/6a70faab12f91ad3eaf2398b9de2c5970d2aff92/samples/tutorials/Data%20passing%20in%20python%20components.ipynb) but the size of all arguments passed to `arguments` in `create_run_from_pipeline_func` exceeds the size limitation, regardless.

Passing files instead of arguments would defeat the purpose of the Experiment UI `Input Parameters` tracking. 

_Originally posted by @WalidMustapha in https://github.com/kubeflow/pipelines/issues/2286#issuecomment-639331095_

### LINKED ISSUE #12510: [feature] Increase input parameter size limitation
### Feature Area

backend
sdk

KFP currently enforces a **maximum input parameter size limit of 10,000 characters**, returning the error:

```
The pipeline spec is invalid: Invalid input error:
The input parameter length exceed maximum size of 10000.
```

I would like to request **allowing higher configurable overrides** of this 10k-character limit in KFP v2.  We already had a patch that works internally and would like to contribute back to the community.
According to the v2 spec and recent compiler/IR changes, pipeline inputs should support significantly larger payloads when encoded as `Parameter` values, especially for JSON-structured metadata.

Many real-world KFP pipelines pass **large JSON configuration payloads**, experiment variant sets, or input look-up tables that can easily exceed 10k characters.  
In our production usage:

- A pipeline receives a generated JSON configuration describing multiple experiment variants (control + experimental arms).
- The strict 10k cutoff blocks valid pipelines that compiled and ran correctly under KFP v1 semantics and should also be supported under KFP v2 IR.

The current limit makes it difficult to use KFP v2 as a general-purpose orchestration layer for ML workflows where parameters can be moderately large but not large enough to justify artifacts.

We actually already had a patch that works for us internally and would like to contribute back to upstream. We are happy to open a PR/doc for further discussion

A native increase of the parameter size limit would significantly improve developer experience and pipeline flexibility.

Originally mentioned by @timsolovev in #2286 and @Leewaldoe in #4828. 
---

Love this idea? Give it a 👍.

neo99zhang:
  This PR increases the maximum parameter size limit from 10KB to 20KB by doubling `MaxParameterBytes` from
   10000 to 20000.

  ## Motivation
  Fixes #2286, fixes #4828, fixes #12510

  Users need to pass larger parameter values to their pipelines. The current 10KB limit is too restrictive 
  for complex parameter configurations.

  ## Test Results:
  ✅ Parameter validation tests - PASSED (2/2)
  - TestValidatePipelineSpecAndResourceReferences_ParameterTooLongWithPipelineId
  - TestValidatePipelineSpecAndResourceReferences_ParameterTooLongWithWorkflowManifest

  ✅ Util package tests - PASSED (all 59 tests)

  ✅ Apiserver package tests - PASSED (all 385 tests)

  ✅ Backend build - SUCCESS

**Checklist:**

@HumairAK @chensun Thanks for your time and attention

### DIFF
--- a/backend/src/common/util/consts.go
+++ b/backend/src/common/util/consts.go
@@ -14,7 +14,12 @@
 
 package util
 
-import constants "github.com/kubeflow/pipelines/backend/src/crd/pkg/apis/scheduledworkflow"
+import (
+	"os"
+	"strconv"
+
+	constants "github.com/kubeflow/pipelines/backend/src/crd/pkg/apis/scheduledworkflow"
+)
 
 const (
 	// LabelKeyScheduledWorkflowEnabled is a label on a ScheduledWorkflow.
@@ -24,8 +29,8 @@ const (
 	// It captures the status of the scheduled workflow.
 	LabelKeyScheduledWorkflowStatus = constants.FullName + "/status"
 
-	// The maximum byte sizes of the parameter column in package/pipeline DB.
-	MaxParameterBytes = 10000
+	// MaxParameterBytesEnvVar is the environment variable name for configuring max parameter bytes.
+	MaxParameterBytesEnvVar = "MAX_PARAMETER_BYTES"
 
 	// LabelKeyWorkflowEpoch is a label on a Workflow.
 	// It captures the epoch at which the workflow was scheduled.
@@ -56,3 +61,18 @@ const (
 	// To disable/enable cache for a single run, this label needs to be added in every step under a run.
 	LabelKeyCacheEnabled = "pipelines.kubeflow.org/cache_enabled"
 )
+
+// GetMaxParameterBytes returns the maximum byte size of parameters.
+// It reads from the MAX_PARAMETER_BYTES environment variable, defaulting to 10000.
+func GetMaxParameterBytes() int {
+	defaultMaxParameterBytes := 10000
+	if envVal := os.Getenv(MaxParameterBytesEnvVar); envVal != "" {
+		if val, err := strconv.Atoi(envVal); err == nil && val > 0 {
+			return val
+		}
+	}
+	return defaultMaxParameterBytes
+}
+
+// MaxParameterBytes is the maximum byte size of the parameter column in package/pipeline DB.
+var MaxParameterBytes = GetMaxParameterBytes()
--- a/backend/src/common/util/consts_test.go
+++ b/backend/src/common/util/consts_test.go
@@ -1,0 +1,131 @@
+// Copyright 2018 The Kubeflow Authors
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+// https://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package util
+
+import (
+	"strconv"
+	"strings"
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+)
+
+func TestMaxParameterBytes_Default(t *testing.T) {
+	defaultMaxParameterBytes := 10000
+	assert.Equal(t, defaultMaxParameterBytes, GetMaxParameterBytes())
+}
+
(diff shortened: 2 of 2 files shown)


backend/src/common/util/consts.go
```diff
@@ -25,7 +25,7 @@ const (
 	LabelKeyScheduledWorkflowStatus = constants.FullName + "/status"
 
 	// The maximum byte sizes of the parameter column in package/pipeline DB.
-	MaxParameterBytes = 10000
+	MaxParameterBytes = 20000
```
chensun: While this change may work in the most cases, I just want to call out there are other system limits exists. 
On top of my head, the input parameters are passed into a container via its command line arguments, which maxes out at 128k on a linux based system, and much smaller (like 8k or 4k) on a windows based system. 
For those using lightweight Python-based component, the entire user code is also included in the command line arguments and counts towards the limit.

neo99zhang: Thanks for pointing this point! That consideration does make sense. My argument is that the current 10k max length is already above their 4/8k limit, so this change would help linux users with double limit without affecting windows users who are already constrained by 4/8k system limit.

alyssacgoins (changes requested):
Can you add some test cases that validate successful runs with parameters size 20KB?

droctothorpe:
Thanks for the contribution, @neo99zhang! An alternative to Alyssa's suggestion would also be acceptable is making it configurable through an environment variable and defaulting it to 10,000.

droctothorpe:
Following up on this @neo99zhang.

neo99zhang:
Thanks @alyssacgoins and @droctothorpe for the review. Hope you had a good holiday.
I just updated the PR to make MaxParameterBytes configurable with default value being 10 KB, as well as testcase to verify it works with parameter size 20 KB.

droctothorpe:
Apologies for the delay! Thanks for addressing the comments! Just rebased. Let's see if CI passes.

neo99zhang:
> Apologies for the delay! Thanks for addressing the comments! Just rebased. Let's see if CI passes.

@droctothorpe Hi Alex, no worries! It looks like the CI exceed github API rate limit, is it possible to trigger again? Thanks!

backend/src/common/util/consts.go
```diff
@@ -56,3 +65,15 @@ const (
 	// To disable/enable cache for a single run, this label needs to be added in every step under a run.
 	LabelKeyCacheEnabled = "pipelines.kubeflow.org/cache_enabled"
 )
+
+// MaxParameterBytes is the maximum byte size of the parameter column in package/pipeline DB.
+// It is initialized from the MAX_PARAMETER_BYTES environment variable, defaulting to DefaultMaxParameterBytes.
+var MaxParameterBytes = DefaultMaxParameterBytes
+
+func init() {
```
droctothorpe: Rather than handling this implicitly in an `init` function, I suggest defining a `GetMaxParameterBytes` func and setting  `MaxParameterBytes` to the output of said func. You can specify the default there as well instead of as a const.

neo99zhang: Thanks for reviewing.  Refactored to use `GetMaxParameterBytes`() with the default defined inside the function. Also updated testcases accordingly.

droctothorpe:
/lgtm
/approve

Thank you for the contribution and thank you for addressing the feedback, @neo99zhang! Much appreciated. Should merge once CI completes.

neo99zhang:
> /lgtm /approve
> 
> Thank you for the contribution and thank you for addressing the feedback, @neo99zhang! Much appreciated. Should merge once CI completes.

Thank you all for reviewing the MR. Appreciate the attention and feedbacks!

PR #12540, opened 2025-12-08
chore(CI): Embed Argo Workflow logs into HTML test reports

hbelmiro:
Embed Argo Workflow archived logs into failing testcases by augmenting `junit.xml` before `junit2html`, so the HTML report contains AWF logs inline.

Example of a failing run: https://github.com/kubeflow/pipelines/actions/runs/20209815433?pr=12540 

Screenshot 2025-12-15 at 09 42 19

## Changes

- **Go tests**: record **test → workflow name(s)** on failure (via `pipeline/runid`).
- **CI action**: fetch archived logs, append to failing testcases in `junit.xml`, then generate/upload HTML.

**Checklist:**

### DIFF
--- a/backend/test/testutil/archived_workflow_logs.go
+++ b/backend/test/testutil/archived_workflow_logs.go
@@ -18,12 +18,6 @@ import (
 )
 
 const (
-	minioSecretName   = "mlpipeline-minio-artifact"
-	minioBucket       = "mlpipeline"
-	logsPrefixBaseFmt = "private-artifacts/%s"
-
-	minioLocalEndpoint = "localhost:9000"
-
 	pollTimeout  = 90 * time.Second
 	pollInterval = 2 * time.Second
 
@@ -34,20 +28,20 @@ const (
 )
 
 func getMinioCreds(k8Client *kubernetes.Clientset, namespace string) (string, string, error) {
-	secret, err := k8Client.CoreV1().Secrets(namespace).Get(context.Background(), minioSecretName, metav1.GetOptions{})
+	secret, err := k8Client.CoreV1().Secrets(namespace).Get(context.Background(), *config.MinioSecretName, metav1.GetOptions{})
 	if err != nil {
 		return "", "", err
 	}
 	access := strings.TrimSpace(string(secret.Data["accesskey"]))
 	secretKey := strings.TrimSpace(string(secret.Data["secretkey"]))
 	if access == "" || secretKey == "" {
-		return "", "", fmt.Errorf("missing accesskey/secretkey in secret %s/%s", namespace, minioSecretName)
+		return "", "", fmt.Errorf("missing accesskey/secretkey in secret %s/%s", namespace, *config.MinioSecretName)
 	}
 	return access, secretKey, nil
 }
 
 func newMinioClient(accessKey, secretKey string) (*minio.Client, error) {
-	return minio.New(minioLocalEndpoint, &minio.Options{
+	return minio.New(*config.MinioEndpoint, &minio.Options{
 		Creds:  credentials.NewStaticV4(accessKey, secretKey, ""),
 		Secure: false,
 	})
@@ -132,15 +126,16 @@ func tryAppendWorkflowKeysOrMessage(
 	workflowNamespace string,
 	workflow string,
 ) (keys []string, ok bool) {
-	prefix := fmt.Sprintf(logsPrefixBaseFmt, workflowNamespace)
+	prefix := fmt.Sprintf(*config.MinioLogsPrefixFmt, workflowNamespace)
 	wfPrefix := fmt.Sprintf("%s/%s/", prefix, workflow)
-	keys, waited, err := pollForLogKeys(ctx, client, minioBucket, wfPrefix)
+	bucket := *config.MinioBucket
+	keys, waited, err := pollForLogKeys(ctx, client, bucket, wfPrefix)
 	if err != nil {
-		*lines = append(*lines, fmt.Sprintf("[error waiting for *.log under s3://%s/%s after %s] %v", minioBucket, wfPrefix, waited, err))
+		*lines = append(*lines, fmt.Sprintf("[error waiting for *.log under s3://%s/%s after %s] %v", bucket, wfPrefix, waited, err))
 		return nil, false
 	}
 	if len(keys) == 0 {
-		*lines = append(*lines, fmt.Sprintf("[no *.log files found under s3://%s/%s after waiting %s]", minioBucket, wfPrefix, waited))
+		*lines = append(*lines, fmt.Sprintf("[no *.log files found under s3://%s/%s after waiting %s]", bucket, wfPrefix, waited))
 		return nil, false
 	}
 	return keys, true
@@ -148,6 +143,7 @@ func tryAppendWorkflowKeysOrMessage(
 
 func buildWorkflowLogsText(ctx context.Context, client *minio.Client, workflowNamespace string, workflows []string) string {
 	lines := []string{"===== Argo Workflows archived logs (tailed) ====="}
+	bucket := *config.MinioBucket
 	for _, wf := range workflows {
 		lines = append(lines, fmt.Sprintf("--- Workflow: %s ---", wf))
 		keys, ok := tryAppendWorkflowKeysOrMessage(&lines, ctx, client, workflowNamespace, wf)
@@ -164,7 +160,7 @@ func buildWorkflowLogsText(ctx context.Context, client *minio.Client, workflowNa
 			step := path.Base(path.Dir(key))
 			lines = append(lines, fmt.Sprintf("[step: %s]", step))
 			maxBytes := min(maxBytesPerStep, bytesBudget)
-			content, err := tailObject(ctx, client, minioBucket, key, maxBytes)
+			content, err := tailObject(ctx, client, bucket, key, maxBytes)
 			if err != nil {
 				lines = append(lines, fmt.Sprintf("[failed to fetch %s] %v", key, err))
 				lines = append(lines, "")
@@ -194,7 +190,7 @@ func BuildArchivedWorkflowLogsReport(k8Client *kubernetes.Clientset, runIDs []st
 	deployNamespace := *config.Namespace
 	accessKey, secretKey, err := getMinioCreds(k8Client, deployNamespace)
 	if err != nil {
-		return fmt.Sprintf("WARNING: failed to read %s secret (%s/%s): %v\n", minioSecretName, deployNamespace, minioSecretName, err), err
+		return fmt.Sprintf("WARNING: failed to read %s secret (%s/%s): %v\n", *config.MinioSecretName, deployNamespace, *config.MinioSecretName, err), err
 	}
 	client, err := newMinioClient(accessKey, secretKey)
 	if err != nil {
--- a/backend/test/config/flags.go
+++ b/backend/test/config/flags.go
@@ -51,3 +51,10 @@ var (
 	KubeflowMode  = flag.Bool("kubeflowMode", false, "Runs tests in full Kubeflow mode")
 	UserNamespace = flag.String("userNamespace", "kubeflow-user-example-com", "The namespace that will store the test resources in Kubeflow mode")
 )
+
+var (
+	MinioSecretName    = flag.String("minioSecretName", "mlpipeline-minio-artifact", "Name of the secret containing MinIO credentials")
+	MinioBucket        = flag.String("minioBucket", "mlpipeline", "S3 bucket name for archived workflow logs")
+	MinioEndpoint      = flag.String("minioEndpoint", "localhost:9000", "MinIO endpoint (host:port)")
+	MinioLogsPrefixFmt = flag.String("minioLogsPrefixFmt", "private-artifacts/%s", "Format string for logs prefix (use %s for workflow namespace)")
+)


HumairAK:
@github Review this PR for performance issues

.github/actions/test-and-report/action.yml
```diff
@@ -101,19 +101,69 @@ runs:
...
+        sleep 2
+        if ! kill -0 "$PF_PID" 2>/dev/null; then
+          echo "ERROR: minio port-forward exited early"
+          cat /tmp/minio-port-forward.log || true
+          exit 1
+        fi
+
+    - name: Augment junit.xml with workflow logs
+      id: augment-junit-xml
+      if: ${{ steps.run-tests.outcome != 'success' }}
```
nsingla: i think this should also check if `install-mc` is a success

.github/actions/test-and-report/action.yml
```diff
@@ -129,36 +179,21 @@ runs:
     - name: Upload HTML Report
       id: upload
       uses: actions/upload-artifact@v4
-      if: (!cancelled())
+      if: (!cancelled()) && steps.generate-html-report.outcome == 'success'
       with:
         name: ${{ steps.name_gen.outputs.REPORT_NAME }}
         path: ${{ inputs.test_directory }}/reports/test-report.html
         retention-days: 7
-      continue-on-error: true
```
nsingla, line 137: so we will fail the workflow without generating junit summary if html report upload fails? I think I would like to see a test report if 1 type of test report generation/upload fails

.github/actions/test-and-report/action.yml
```diff
@@ -129,36 +179,21 @@ runs:
...
 
     - name: Publish Test Summary With HTML Report
       id: publish
       uses: ./.github/actions/junit-summary
-      if: (!cancelled()) && steps.upload.outcome != 'failure'
+      if: always() && !cancelled()
```
nsingla, line 158: similar to above, I will still keep 2 summary steps, for just in case?

.github/actions/test-and-report/action.yml
```diff
@@ -129,36 +179,21 @@ runs:
...
-      with:
-        xml_files: '${{ inputs.test_directory }}/reports'
-      continue-on-error: true
 
     - name: Mark Workflow failure if test step failed
       if: steps.run-tests.outcome != 'success' && !cancelled()
       shell: bash
       run: exit 1
-
-    - name: Mark Workflow failure if test reporting failed
```
nsingla, line 161: this is to catch failures, just in case

.github/actions/test-and-report/augment-junit-xml-with-workflow-logs.py
```diff
@@ -0,0 +1,323 @@
...
+import os
+import re
+import subprocess
+import sys
+import tempfile
+from xml.etree import ElementTree
+from typing import Dict, List, NoReturn, Optional, Tuple
+
+
+REPORTS_DIRNAME = "reports"
```
nsingla: can these not be hardcoded and instead be optional arguments to the script?

.github/actions/test-and-report/augment-junit-xml-with-workflow-logs.py
```diff
@@ -0,0 +1,323 @@
...
+    )
+
+
+def _get_artifact_repo(namespace: str) -> str:
+    cp = _run(
+        [
+            "kubectl",
+            "get",
+            "configmap",
+            "workflow-controller-configmap",
```
nsingla: can this configmap name not change?

.github/actions/test-and-report/augment-junit-xml-with-workflow-logs.py
```diff
@@ -0,0 +1,323 @@
...
+    return cp.stdout or ""
+
+
+def _get_minio_creds(namespace: str) -> tuple[str, str]:
+    access = _run(
+        [
+            "kubectl",
+            "get",
+            "secret",
+            "mlpipeline-minio-artifact",
```
nsingla: hardcoding it seems like inflexibility and brittle in case the names change in the future

.github/actions/test-and-report/augment-junit-xml-with-workflow-logs.py
```diff
@@ -0,0 +1,323 @@
...
+    secret_key = base64.b64decode(secret.encode("utf-8")).decode("utf-8").strip()
+    return access_key, secret_key
+
+
+def _mc_alias_set(mc: str, endpoint: str, access_key: str, secret_key: str) -> None:
+    _run([mc, "alias", "set", "kfp-minio", endpoint, access_key, secret_key, "--api", "S3v4"])
+
+
+def _mc_find_logs(mc: str, bucket: str, prefix: str) -> list[str]:
+    cp = _run([mc, "find", f"kfp-minio/{bucket}/{prefix}", "--name", "*.log"], check=False)
```
nsingla: where is this `kfp-minio` prefix defined? is this configurable somewhere?

.github/actions/test-and-report/action.yml
hbelmiro: If `install-mc` fails, the job fails immediately.

.github/actions/test-and-report/action.yml
hbelmiro: IMO, we should be strict here. If anything in the HTML report fails, the report generation should be fixed. Unless it's a network issue, which is rare. Otherwise we may never notice that there's something wrong with the report generation.
It also keeps the workflow simpler and easier to maintain.

.github/actions/test-and-report/augment-junit-xml-with-workflow-logs.py
hbelmiro: That name is pinned in our manifests (see [manifests/kustomize/third-party/argo/base/workflow-controller-configmap-patch.yaml](https://github.com/kubeflow/pipelines/blob/b050f5c90650488edce15ec1350340c0eb86c3e8/manifests/kustomize/third-party/argo/base/workflow-controller-configmap-patch.yaml)) and is the expected Argo Workflows controller configmap name in our deployment, so it’s safe to reference directly.

.github/actions/test-and-report/augment-junit-xml-with-workflow-logs.py
hbelmiro: Same as above. It's defined in the manifests.

.github/actions/test-and-report/augment-junit-xml-with-workflow-logs.py
hbelmiro: It’s the MinIO client alias name we create at runtime in this script via:
`mc alias set kfp-minio http://localhost:9000 <access> <secret> --api S3v4` (see the `_mc_alias_set` function)
After that, all mc commands can reference the endpoint as `kfp-minio/<bucket>/....`
So it’s only a local label within the script.

.github/actions/test-and-report/action.yml
nsingla: i agree, but i think we can run into other situations like network issue, or other issues with junit2html plugin, which will block generation of regular junit reports as well.

.github/actions/test-and-report/augment-junit-xml-with-workflow-logs.py
nsingla: hmmm..... can we then add a comment here referencing the manifest and a comment in the manifest referencing this script, so that if in future we are changing it, we know where else the impact would be

.github/actions/test-and-report/augment-junit-xml-with-workflow-logs.py
nsingla: can we then add comments in both places for reference then?

.github/actions/test-and-report/augment-junit-xml-with-workflow-logs.py
nsingla: can this be extracted out and used as a variable then? so that we have a single place

backend/test/end2end/pipeline_e2e_test.go
```diff
@@ -94,10 +95,21 @@ var _ = Describe("Upload and Verify Pipeline Run >", Label(FullRegression), func
...
 
+		// Capture workflow mapping for failed tests before cleanup deletes the workflows
+		if CurrentSpecReport().Failed() && len(testContext.PipelineRun.CreatedRunIds) > 0 {
+			currentDir, err := os.Getwd()
+			if err == nil {
+				testutil.WriteTestWorkflowMapping(
+					GinkgoT().Name(),
+					testContext.PipelineRun.CreatedRunIds,
+					testutil.GetNamespace(),
+					filepath.Join(currentDir, testReportDirectory, "test-workflow-mapping.txt"),
```
nsingla: Can we convert `test-workflow-mapping.txt` to an environment variable, so that both script and go tests can read the file name from the env variable rather than hard coding it

backend/test/end2end/pipeline_e2e_test.go
```diff
@@ -94,10 +95,21 @@ var _ = Describe("Upload and Verify Pipeline Run >", Label(FullRegression), func
 	})
 
 	AfterEach(func() {
-
-		// Delete pipelines created during the test
 		logger.Log("################### Global Cleanup after each test #####################")
 
+		// Capture workflow mapping for failed tests before cleanup deletes the workflows
+		if CurrentSpecReport().Failed() && len(testContext.PipelineRun.CreatedRunIds) > 0 {
```
nsingla: This step is already available in the suite file, you can just enhance that

hbelmiro: `ReportAfterEach` is already used in `e2e_suite_test.go`, but it runs after the spec’s `AfterEach` nodes. In our E2E tests the `AfterEach` in `pipeline_e2e_test.go` performs the cleanup that can delete the Argo Workflows.

hbelmiro:
@nsingla 

> Rather than augmenting the junit xml, is there no way, we can just fetch the workflow logs directly from s3/minio and append the logs in the AfterEach step?

I tried that, but Argo archives logs asynchronously. We don't find them immediately after workflow completion.

.github/actions/test-and-report/action.yml
hbelmiro: Done.

.github/actions/test-and-report/action.yml
hbelmiro: I brought that behavior back.

.github/actions/test-and-report/action.yml
hbelmiro: Done.

.github/actions/test-and-report/action.yml
hbelmiro: Brought back.

.github/actions/test-and-report/augment-junit-xml-with-workflow-logs.py
hbelmiro: Done.

.github/actions/test-and-report/augment-junit-xml-with-workflow-logs.py
hbelmiro: Done.

.github/actions/test-and-report/augment-junit-xml-with-workflow-logs.py
hbelmiro: Done.

.github/actions/test-and-report/augment-junit-xml-with-workflow-logs.py
hbelmiro: Done.

backend/test/end2end/pipeline_e2e_test.go
hbelmiro: Done.

.github/actions/test-and-report/augment-junit-xml-with-workflow-logs.py
```diff
@@ -0,0 +1,358 @@
...
+        _append_system_out(tc, out)
+        modified += 1
+
+    if modified == 0:
+        return 0
+
+    with tempfile.NamedTemporaryFile("wb", delete=False, dir=os.path.dirname(junit_xml_path) or None) as tmp:
+        tmp_path = tmp.name
+        tree.write(tmp, encoding="utf-8", xml_declaration=True)
+    os.replace(tmp_path, junit_xml_path)
```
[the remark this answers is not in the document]
hbelmiro: We don't care if the file is not cleaned up in the CI.

.github/actions/test-and-report/augment-junit-xml-with-workflow-logs.py
```diff
@@ -0,0 +1,358 @@
...
+            "-n",
+            namespace,
+            "-o",
+            "jsonpath={.data.artifactRepository}",
+        ]
+    )
+    return cp.stdout or ""
+
+
+def _get_minio_creds(namespace: str) -> tuple[str, str]:
```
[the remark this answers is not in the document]
hbelmiro: Done.

.github/actions/test-and-report/action.yml
```diff
@@ -101,17 +102,70 @@ runs:
...
+        echo "MC_PATH=$MC_PATH" >> "$GITHUB_ENV"
+      continue-on-error: true
+
+    - name: Port-forward MinIO service for workflow log access
+      id: port-forward-minio
+      if: ${{ steps.run-tests.outcome != 'success' }}
+      shell: bash
+      run: |
+        NAMESPACE="${{ inputs.default_namespace }}"
+        kubectl port-forward -n "$NAMESPACE" svc/minio-service 9000:9000 >/tmp/minio-port-forward.log 2>&1 &
```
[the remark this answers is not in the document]
hbelmiro: Done.

.github/actions/test-and-report/action.yml
```diff
@@ -101,17 +102,70 @@ runs:
...
+        PF_PID=$!
+        sleep 2
+        if ! kill -0 "$PF_PID" 2>/dev/null; then
+          echo "ERROR: minio port-forward exited early"
+          cat /tmp/minio-port-forward.log || true
+          exit 1
+        fi
```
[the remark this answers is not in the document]
hbelmiro: Not needed.

.github/actions/test-and-report/augment-junit-xml-with-workflow-logs.py
```diff
@@ -0,0 +1,358 @@
...
+        _warn("Argo Workflows log archiving is not enabled (archiveLogs: true not found); skipping.")
+        return
+
+    try:
+        access_key, secret_key = _get_minio_creds(namespace)
+    except subprocess.CalledProcessError as e:
+        _warn(f"Failed to read artifact credentials: {e}; skipping.")
+        return
+    try:
+        _mc_alias_set(mc, "http://localhost:9000", access_key, secret_key)
```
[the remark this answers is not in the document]
hbelmiro: This is a test environment. Exposing credentials is not an issue.

backend/test/testutil/test_utils.go
```diff
@@ -88,6 +89,80 @@ func WriteLogFile(specReport types.SpecReport, testName, logDirectory string) {
...
+		logger.Log("No workflows found for run IDs %v, skipping mapping", runIDs)
+		return
+	}
+
+	mappingDir := filepath.Dir(mappingFilePath)
+	if err := os.MkdirAll(mappingDir, 0755); err != nil {
+		logger.Log("Failed to create mapping directory %q: %v", mappingDir, err)
+		return
+	}
+	file, err := os.OpenFile(mappingFilePath, os.O_APPEND|os.O_CREATE|os.O_WRONLY, 0644)
```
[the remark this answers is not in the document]
hbelmiro: Done.

nsingla:
> @nsingla
> 
> > Rather than augmenting the junit xml, is there no way, we can just fetch the workflow logs directly from s3/minio and append the logs in the AfterEach step?
> 
> I tried that, but Argo archives logs asynchronously. We don't find them immediately after workflow completion.

can we wait for it then? if we know its supposed to be there, we can try to poll for it with a timeout (just like we poll for run status in the e2e tests :) )
I think it will much simpler and more efficient way than augmenting the test report

hbelmiro:
> > @nsingla
> > > Rather than augmenting the junit xml, is there no way, we can just fetch the workflow logs directly from s3/minio and append the logs in the AfterEach step?
> > 
> > 
> > I tried that, but Argo archives logs asynchronously. We don't find them immediately after workflow completion.
> 
> can we wait for it then? if we know its supposed to be there, we can try to poll for it with a timeout (just like we poll for run status in the e2e tests :) ) I think it will much simpler and more efficient way than augmenting the test report

> much simpler

Not really. It actually moves the complexity to another place. Besides having to wait, every test suite would need minio setup, credentials... Also, we would have to find a way to reuse those. Currently the CI step does this once and easily.

I'd be happy to get rid of the Python script though :)

> more efficient

It's actually less efficient. The amount of fetches is the same, but we would have to wait (due to the asynchronicity) in each failed test. I can't tell regarding the different complexities.

nsingla:
> > > @nsingla
> > > > Rather than augmenting the junit xml, is there no way, we can just fetch the workflow logs directly from s3/minio and append the logs in the AfterEach step?
> > > 
> > > 
> > > I tried that, but Argo archives logs asynchronously. We don't find them immediately after workflow completion.
> > 
> > 
> > can we wait for it then? if we know its supposed to be there, we can try to poll for it with a timeout (just like we poll for run status in the e2e tests :) ) I think it will much simpler and more efficient way than augmenting the test report
> 
> > much simpler
> 
> Not really. It actually moves the complexity to another place. Besides having to wait, every test suite would need minio setup, credentials... Also, we would have to find a way to reuse those. Currently the CI step does this once and easily.
> 
> I'd be happy to get rid of the Python script though :)
> 
> > more efficient
> 
> It's actually less efficient. The amount of fetches is the same, but we would have to wait (due to the asynchronicity) in each failed test. I can't tell regarding the different complexities.

I think its still makes sense to fetch it in the aftereach step (as its only gonna happen for failed tests only).

>besides having to wait, every test suite would need minio setup, credentials.
That's just one time config, and we only need it in e2e tests suite - we can create a s3 client once in the `beforesuite` and use it in the tests - or can we move it to test_utils as well, and have the tests suite call that to avoid code duplicacy

Or We can also use argo REST APIs to fetch logs, which may not be preferred actually :thinking: 

Also, we need to think about how these tests would fetch logs when running into non local mode i.e. not on a kind cluster

hbelmiro:
> > > > @nsingla
> > > > > Rather than augmenting the junit xml, is there no way, we can just fetch the workflow logs directly from s3/minio and append the logs in the AfterEach step?
> > > > 
> > > > 
> > > > I tried that, but Argo archives logs asynchronously. We don't find them immediately after workflow completion.
> > > 
> > > 
> > > can we wait for it then? if we know its supposed to be there, we can try to poll for it with a timeout (just like we poll for run status in the e2e tests :) ) I think it will much simpler and more efficient way than augmenting the test report
> > 
> > 
> > > much simpler
> > 
> > 
> > Not really. It actually moves the complexity to another place. Besides having to wait, every test suite would need minio setup, credentials... Also, we would have to find a way to reuse those. Currently the CI step does this once and easily.
> > I'd be happy to get rid of the Python script though :)
> > > more efficient
> > 
> > 
> > It's actually less efficient. The amount of fetches is the same, but we would have to wait (due to the asynchronicity) in each failed test. I can't tell regarding the different complexities.
> 
> I think its still makes sense to fetch it in the aftereach step (as its only gonna happen for failed tests only).
> 
> > besides having to wait, every test suite would need minio setup, credentials.
> > That's just one time config, and we only need it in e2e tests suite - we can create a s3 client once in the `beforesuite` and use it in the tests - or can we move it to test_utils as well, and have the tests suite call that to avoid code duplicacy
> 
> Or We can also use argo REST APIs to fetch logs, which may not be preferred actually 🤔
> 
> Also, we need to think about how these tests would fetch logs when running into non local mode i.e. not on a kind cluster

Done. Can you please take another look?

manifests/kustomize/third-party/minio/base/mlpipeline-minio-artifact-secret.yaml
juliusvonkohout, line 1: given the release with seaweedfs and the dead minio project we should actually just remove minio entirely. CC @pschoen-itsc @akagami-harsh

nsingla: I think we should, but not in this PR and plan for it in a separate release

backend/test/v2/api/integration_suite_test.go
```diff
@@ -163,10 +163,31 @@ var _ = BeforeEach(func() {
...
+
+	if specReport.Failed() {
+		if len(testContext.PipelineRun.CreatedRunIds) > 0 {
+			report, _ := testutil.BuildArchivedWorkflowLogsReport(k8Client, testContext.PipelineRun.CreatedRunIds)
+			AddReportEntry(testutil.ArchivedWorkflowLogsReportTitle, report)
+		}
+
+		logger.Log("Test failed... Capturing pod logs from %v to %v", testContext.TestStartTimeUTC, time.Now().UTC())
+		podLogs := testutil.ReadContainerLogs(k8Client, *config.Namespace, "pipeline-api-server", nil, &testContext.TestStartTimeUTC, config.PodLogLimit)
+		AddReportEntry("Pod Log", podLogs)
```
nsingla, line 182: If we have the argo workflow logs, do we still need Pod Logs, can we actually try to fetch pod logs only as a fallback?

nsingla: nvmd, this is only fetching api server logs, so its fine

backend/test/testutil/archived_workflow_logs.go
```diff
@@ -0,0 +1,223 @@
...
+	"github.com/kubeflow/pipelines/backend/test/config"
+	"github.com/kubeflow/pipelines/backend/test/logger"
+	"github.com/minio/minio-go/v7"
+	"github.com/minio/minio-go/v7/pkg/credentials"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/client-go/kubernetes"
+)
+
+const (
+	minioSecretName   = "mlpipeline-minio-artifact"
```
nsingla: I think these (L21-l25) are better under [flags.go](https://github.com/kubeflow/pipelines/blob/master/backend/test/config/flags.go), so that if needed (for e.g. in a Openshift cluster installation where tests run w/o creating a tunnel and minio may have different bucket and secret), its easy to overrride the values

manifests/kustomize/third-party/minio/base/mlpipeline-minio-artifact-secret.yaml
pschoen-itsc: > given the release with seaweedfs and the dead minio project we should actually just remove minio entirely. CC @pschoen-itsc @akagami-harsh

+1 from me on this.

akagami-harsh: I was thinking the same since @HumairAK suggested we should [keep it for a release](https://github.com/kubeflow/pipelines/pull/11965#discussion_r2226164854) to let users migrate. i'll create a PR to remove minio entirely and will also try to address [minio naming concerns](https://github.com/kubeflow/pipelines/pull/11965#discussion_r2260436117)

backend/test/testutil/archived_workflow_logs.go
hbelmiro: Done.

Issue #12536, opened 2025-12-08
[bug] Default parameters not displayed in Compare Runs


ishida-tomomi:
### Environment

*   Kubeflow: v1.10.2
*   Kubeflow Pipelines: v2.5.0

### **Summary**

In these versions, we observed the following issue with the Compare Runs feature:

*   **Parameters may not be displayed**
    *   When creating a Run from the GUI and leaving all parameters at their default values, the Compare Runs screen does not show any parameter information.
    *   If the user interacts with a parameter (e.g., deletes a character and re-enters it), the parameters are displayed.

### Steps to reproduce

1.  Create a new Run from the GUI with all parameters left at their default values.
2.  Compare the target Run in the Compare Runs screen.
3.  Confirm that the Parameters section is empty.

### Expected result

*   The Compare Runs screen should display parameters, including default values.

### **Investigation Results**

*   In the GUI code (`NewRunV2.tsx`), parameter information is updated only via the `onChange` event, so if there is no user interaction, it is not sent to the API.
*   The latest version (v2.14.3) still does not include a fix.
*   Related code:
    *   NewRunV2.tsx#L673: <https://github.com/kubeflow/pipelines/blob/2.5.0/frontend/src/pages/NewRunV2.tsx#L673>

### **Possible Cause**

*   Due to an API-side specification change, default values are not set in `executorInput` when there is no user input.
*   The GUI does not implement logic to display default values in Compare Runs.

### **Proposal**

*   On the GUI side, if the API does not provide parameter values, display the default values from the pipeline definition.
*   If this is by design, update the UI text to clarify (e.g., “Parameters for Executor Input”).

### **Impact**

*   This limitation prevents accurate parameter comparison.
*   Users must perform unnecessary GUI interactions as a workaround.

### Labels

---

Impacted by this bug? Give it a 👍.

PR #12541, opened 2025-12-09
Fix /default parameters not displayed in Compare Runs 

rishabh998186:
This PR fixes an issue where default pipeline parameters are not shown in the Compare Runs view when a run is created from the UI without changing any parameter values. Previously, parameter values were only propagated when the user interacted with the inputs, so untouched defaults were missing from runtime_config and therefore from Compare Runs.

## changes ##

NewRunParametersV2.tsx
``` 

Extend the existing parameter initialization effect to:

1. Build runtimeParametersInRealType from specParameters using convertInput for each parameter type.

2. Call handleParameterChange(runtimeParametersInRealType) once on mount/when spec changes so that:

 * Default parameter values are propagated to the parent (setRuntimeParameters).

 * These defaults are included in runtime_config even if the user does not modify any parameter fields.

3. Preserve the existing validation behavior with setIsValidInput.
```

NewRunParametersV2.test.tsx

```
1. Add a test: calls handleParameterChange with default values on mount:

Mount NewRunParametersV2 with a set of parameters of different types: STRING, NUMBER_INTEGER, BOOLEAN, LIST, and STRUCT, each with a defaultValue.

2.Assert that handleParameterChange is:

* Called exactly once.

*Called with the correctly converted runtime values.

3. Assert that these default values are rendered in the UI inputs.
```

Fixes: #12536

### DIFF
--- a/frontend/src/components/NewRunParametersV2.test.tsx
+++ b/frontend/src/components/NewRunParametersV2.test.tsx
@@ -947,140 +947,139 @@ describe('NewRunParametersV2', () => {
   });
 });
 
-  describe('Bug Fix: Default Parameters in Compare Runs (#12536)', () => {
-    
-    it('SCENARIO 1: User creates run with ALL default parameters (no changes)', () => {
-      const handleParameterChangeSpy = jest.fn();
-      
-      const props = {
-        titleMessage: 'Specify parameters required by the pipeline',
-        specParameters: {
-          string_param: {
-            parameterType: ParameterType_ParameterTypeEnum.STRING,
-            defaultValue: 'default_string_value',
-          },
-          integer_param: {
-            parameterType: ParameterType_ParameterTypeEnum.NUMBER_INTEGER,
-            defaultValue: 42,
-          },
-          boolean_param: {
-            parameterType: ParameterType_ParameterTypeEnum.BOOLEAN,
-            defaultValue: true,
-          },
-          float_param: {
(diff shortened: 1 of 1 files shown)


rishabh998186:
Hey @droctothorpe, thanks for looking at my PR!

I've made some new changes to address the Compare Runs issue:

NewRunParametersV2.tsx: Now calls `handleParameterChange` on mount with all default parameter values, ensuring they're included in runtime_config even without user interaction

NewRunParametersV2.test.tsx: Updated existing tests and added 3 new test cases covering default-only runs, partial changes, and falsy default values

All tests are passing and the frontend build succeeds. Could you please review these changes and let me know if you need any modifications?

Thanks!

rishabh998186:
@droctothorpe I think all required checks are passed the failing check is just an syntex error . If anything is still missing  let me know , I fix it.

thinks for looking at my pr .

droctothorpe:
Np! Tag me once you fix the syntax error.

rishabh998186:
@droctothorpe done ,i fixed the syntex error.

droctothorpe:
/approve

Thanks for the contribution, @rishabh998186  ! 🎉

rishabh998186:
@droctothorpe  I think tke LGTM is required to merge this pr. Bot removed the label due to the new commit of syntex  error fix .
And  Thanks for reviewing my pr.

PR #12543, opened 2025-12-09
fix(manifests): Allow same-namespace access to SeaweedFS for standalone deployments

hbelmiro:

The SeaweedFS NetworkPolicy only allows traffic from namespaces labeled `app.kubernetes.io/part-of: kubeflow-profile` or from `istio-system`. This blocks the API server (running in the `kubeflow` namespace) from accessing object storage on clusters with NetworkPolicy enforcement, **breaking standalone KFP deployments**.

**Fix:** Add an ingress rule allowing same-namespace pod traffic to SeaweedFS on port 8333.

**Note:** This change aligns with the upstream Kubeflow manifests repository, which deploys a [`default-allow-same-namespace.yaml`](https://github.com/kubeflow/manifests/blob/master/common/networkpolicies/base/default-allow-same-namespace.yaml) NetworkPolicy for the kubeflow namespace. For standalone KFP, we achieve the same functional result by adding the rule directly to the SeaweedFS policy rather than introducing a separate blanket policy. This keeps the standalone deployment simpler while solving the connectivity issue.

## Changes

- Added same-namespace ingress rule to SeaweedFS NetworkPolicy
- Fixes standalone deployments by allowing API server and UI to access object storage

**Checklist:**


manifests/kustomize/cluster-scoped-resources/namespace.yaml
```diff
@@ -2,3 +2,5 @@ apiVersion: v1
 kind: Namespace
 metadata:
   name: '$(kfp-cluster-scoped-namespace)'
+  labels:
+    app.kubernetes.io/part-of: kubeflow-profile
```
HumairAK: this doesn't seem like the right solution, the "kubeflow" namespace is not a "profile" namespace, those are user namespaces, whereas the "kubeflow" namespace is where kubeflow platform pods are deployed

.github/resources/manifests/base/allow-same-namespace-networkpolicy.yaml
```diff
@@ -1,12 +0,0 @@
-apiVersion: networking.k8s.io/v1
```
HumairAK, line 1: we shouldn't delete these, this is just mimicking what network policy exists in Kubeflow platform deployment

hbelmiro:
@juliusvonkohout can you PTAL on this?

juliusvonkohout:
Thank you @hbelmiro it looks good and and more robust than before, but may you raise the same PR against kubeflow/manifests to also test the CI/CD there? If it is green and you see a kfp tests, then both are good to merge.

juliusvonkohout:
https://github.com/kubeflow/manifests/pull/3306 is merged so
/lgtm

and thank you for making it more robust

juliusvonkohout:
Just be careful if more than this single port is needed in the future for the same namespace. But that can also be adjusted in the future if needed.

hbelmiro:
@HumairAK can you please take a final look?

HumairAK:
/lgtm
@hbelmiro can you rebase and ping me

PR #12442, opened 2025-11-12
feat: Add pipeline run parallelism config

sduvvuri1603:
## Summary
- Follow-up to the cleanup of unused mutex_name/semaphore_key: adds pipeline_run_parallelism as the official workflow-level concurrency knob so pipeline versions can declare how many runs may execute concurrently.
- The feature Pipeline_run_parallelism lets a pipeline version declare how many runs are allowed to execute at the same time, using Argo semaphores backed by a ConfigMap for workflow-level concurrency 
- Threads pipeline_run_parallelism through SDK → IR → compiler → backend; compiler validates (>0) without setting spec.parallelism, and backend CreateRun upserts the kfp-pipeline-config ConfigMap keyed by PipelineVersionId and wires Argo spec.synchronization.semaphores via configMapKeyRef.

## Validation
- SDK and backend goldens now include the updated sample, showing consistent IR and Argo outputs with the parallelism limit.
- Built custom API server and driver images from this branch, loaded them into a kind cluster, ran the sample, and confirmed that the number of simultaneously running component pods never exceeded the configured limit.
- Added the parallelism validation helper to the e2e suite (`e2e_utils.go` + invocation in `pipeline_e2e_test.go`), rebuilt the test cluster with the fresh backend images, exercised the focused pipeline_run_parallelism scenario, and then ran the end-to-end suite to confirm the new check passes with the concurrency cap enforced.

Note:
Hardened .github/actions/test-and-report/action.yml: the workflow now waits for the ml-pipeline deployment, port-forwards svc/ml-pipeline to localhost:8888, and curls /apis/v2beta1/healthz (TLS-aware). This eliminates the dial tcp [::1]:8888: connect: connection refused failures that intermittently broke the parallelism/recurring-run E2E suite.

Follow up to [PR](https://github.com/kubeflow/pipelines/pull/12401) - remove unused semaphore_key and mutex_name fields

PR #12549, opened 2025-12-11
Cleanup kind workflows and Harden release downloads(incl. artifact paging)

sduvvuri1603:
### Description of your changes:

- add a shared cleanup step to integration-tests-v1, kfp-webhooks, and e2e-test-frontend so we free disk and skip log collection when the kubeflow namespace never comes up; this unblocks the disk-exhaustion regressions seen in kubeflow/pipelines#12442

- harden .github/actions/check-artifact-exists so GitHub’s 403 pagination limit is treated as “artifact not found” rather than crashing the job

- fix our Docker builds (backend/Dockerfile, test/release/Dockerfile.release) to fetch Argo CLI and git-cliff with Accept: application/octet-stream and then extract, avoiding HTML responses that were breaking the unpack step

### DIFF
--- a/.github/workflows/e2e-test-frontend.yml
+++ b/.github/workflows/e2e-test-frontend.yml
@@ -28,9 +28,6 @@ jobs:
       - name: Checkout code
         uses: actions/checkout@v5
 
-      - name: Free up disk space
-        run: ./.github/resources/scripts/free-disk-space.sh
-
       - name: Set up Python
         uses: actions/setup-python@v6
         with:
@@ -73,11 +70,7 @@ jobs:
       - name: Collect failed logs
         if: ${{ steps.deploy.outcome != 'success' || steps.forward-frontend-port.outcome != 'success' || steps.tests.outcome != 'success' }}
         run: |
-          if kubectl get namespace kubeflow >/dev/null 2>&1; then
-            ./.github/resources/scripts/collect-logs.sh --ns kubeflow --output /tmp/tmp_pod_log.txt
-          else
-            echo "kubeflow namespace not found; skipping log collection."
-          fi
+          ./.github/resources/scripts/collect-logs.sh --ns kubeflow --output /tmp/tmp_pod_log.txt
           exit 1
 
       - name: Collect test results
--- a/.github/actions/create-cluster/action.yml
+++ b/.github/actions/create-cluster/action.yml
@@ -13,6 +13,10 @@ inputs:
 runs:
   using: "composite"
   steps:
+    - name: Free up disk space on runner
+      shell: bash
+      run: ./.github/resources/scripts/free-disk-space.sh
+
     - name: Create k8s Kind Cluster
       uses: helm/kind-action@v1.12.0
       id: cluster-create
--- a/.github/resources/scripts/collect-logs.sh
+++ b/.github/resources/scripts/collect-logs.sh
@@ -23,9 +23,10 @@ fi
 
 function check_namespace {
     if ! kubectl get namespace "$1" &>/dev/null; then
-        echo "Namespace '$1' does not exist."
-        exit 1
+        echo "Namespace '$1' does not exist. Skipping log collection."
+        return 1
     fi
+    return 0
 }
 
 function display_pod_info {
@@ -61,5 +62,8 @@ function display_pod_info {
     echo "Pod information stored in $OUTPUT_FILE"
 }
 
-check_namespace "$NS"
-display_pod_info "$NS"
+if check_namespace "$NS"; then
+    display_pod_info "$NS"
+else
+    exit 0
+fi
--- a/.github/workflows/integration-tests-v1.yml
+++ b/.github/workflows/integration-tests-v1.yml
@@ -37,9 +37,6 @@ jobs:
       - name: Checkout code
         uses: actions/checkout@v5
 
-      - name: Free up disk space
-        run: ./.github/resources/scripts/free-disk-space.sh
-
       - name: Set up Python
         uses: actions/setup-python@v6
         with:
@@ -88,11 +85,7 @@ jobs:
       - name: Collect failed logs
         if: ${{ steps.forward-mysql-port.outcome != 'success' || steps.integration-tests.outcome != 'success' || steps.initialization-tests.outcome != 'success' }}
         run: |
-          if kubectl get namespace kubeflow >/dev/null 2>&1; then
-            ./.github/resources/scripts/collect-logs.sh --ns kubeflow --output /tmp/tmp_pod_log.txt
-          else
-            echo "kubeflow namespace not found; skipping log collection."
-          fi
+          ./.github/resources/scripts/collect-logs.sh --ns kubeflow --output /tmp/tmp_pod_log.txt
           exit 1
 
       - name: Collect test results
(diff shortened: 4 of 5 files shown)


.github/actions/check-artifact-exists/check_artifact.py
```diff
@@ -79,6 +79,12 @@ def check_artifact_exists(artifact_name: str, token: str, repo: str, head_sha: s
         while max_iterations > 0:
             params['page'] = page
             response = requests.get(url, headers=headers, params=params)
+
+            if response.status_code == requests.codes.forbidden:
```
nsingla, line 83: Nice catch! thanks for fixing it
I am assuming 403 is when we hit some type of api rate limit?

.github/workflows/e2e-test-frontend.yml
```diff
@@ -70,7 +73,11 @@ jobs:
       - name: Collect failed logs
         if: ${{ steps.deploy.outcome != 'success' || steps.forward-frontend-port.outcome != 'success' || steps.tests.outcome != 'success' }}
         run: |
-          ./.github/resources/scripts/collect-logs.sh --ns kubeflow --output /tmp/tmp_pod_log.txt
+          if kubectl get namespace kubeflow >/dev/null 2>&1; then
```
nsingla: wondering, if we should move this check to the collect-logs.sh script itself, create a function that checks if the namespace exist and if it does not then skips collection of the logs, just like here. Because this step is used at multiple places, so will be good to avoid code duplicacy

.github/workflows/e2e-test-frontend.yml
```diff
@@ -28,6 +28,9 @@ jobs:
       - name: Checkout code
         uses: actions/checkout@v5
 
+      - name: Free up disk space
```
nsingla: I think all workflows that create/deploy cluster needs this step, so may be the better solution is to move this step to create_cluster action and remove this step from all the workflows that explicitly define it. (FYI: AI Assistant can perform this step for you very easily)

.github/actions/check-artifact-exists/check_artifact.py
sduvvuri1603: Thanks! 403 is what GitHub returns once our page is past 1 000 artifacts (?per_page=100&page>10). After that point, the endpoint just stops serving data. 

I added the warning/early exit so we treat that case as “artifact not found” instead of failing the job.

.github/workflows/e2e-test-frontend.yml
sduvvuri1603: Sure, That's actually a better way to do. Should I add that in a follow up PR after merging in this quick fix? Pls lmk!

.github/workflows/e2e-test-frontend.yml
sduvvuri1603: Same with this too. Should I do a seperate PR for moving to collect-logs.sh and the above to move disk space name to create_cluster action?

.github/actions/check-artifact-exists/check_artifact.py
nsingla: got it, your changes makes more sense now

.github/workflows/e2e-test-frontend.yml
nsingla: its up to you, if you think that's a lot of work to do in this PR, then I am fine with a follow up PR, but if you think its trivial to do it in this PR, then thats fine by me too.

sduvvuri1603: Cool, I will add that as a follow up since this has been a blocker for a while in several PR's

.github/workflows/e2e-test-frontend.yml
sduvvuri1603: I will include this as well in the follow up PR

nsingla (approved):
lgtm, follow up is required to resolve some of the ask in this PR

backend/Dockerfile
```diff
@@ -39,7 +39,9 @@ RUN echo "Building for architecture: ${TARGETARCH}"
 
 # Downloading Argo CLI so that the samples are validated
 ENV ARGO_VERSION=v3.7.3
-RUN curl -sLO https://github.com/argoproj/argo-workflows/releases/download/${ARGO_VERSION}/argo-linux-$TARGETARCH.gz && \
+RUN curl -sL -H "Accept: application/octet-stream" \
+        https://github.com/argoproj/argo-workflows/releases/download/${ARGO_VERSION}/argo-linux-$TARGETARCH.gz \
+    -o argo-linux-$TARGETARCH.gz && \
```
hbelmiro, line 44: How does this change relates to the cleanup?

test/release/Dockerfile.release
```diff
@@ -49,5 +49,6 @@ RUN chmod -R 777 $NVM_DIR && \
 # Configure npm cache location
 RUN npm config set cache /tmp/.npm --global
 
-RUN curl -L https://github.com/orhun/git-cliff/releases/download/v${GIT_CLIFF_VERSION}/git-cliff-${GIT_CLIFF_VERSION}-x86_64-unknown-linux-gnu.tar.gz \
+RUN curl -L -H "Accept: application/octet-stream" \
+        https://github.com/orhun/git-cliff/releases/download/v${GIT_CLIFF_VERSION}/git-cliff-${GIT_CLIFF_VERSION}-x86_64-unknown-linux-gnu.tar.gz \
```
hbelmiro, line 53: How does this change relates to the cleanup?

.github/actions/check-artifact-exists/check_artifact.py
hbelmiro: How does this change relates to the cleanup?

backend/Dockerfile
sduvvuri1603: While chasing the cleanup flakes, we were also blocked by CI builds failing because GitHub started serving HTML for those release artifacts (Argo CLI and git-cliff) and our downloader script’s 403s. 

The quick tweaks here—adding the Accept: application/octet-stream header and skipping 403s in check-artifact-exists—were needed to keep the workflows passing so the cleanup change could land.

 I’ve already called this out in the PR description(3rd bullet point)

test/release/Dockerfile.release
sduvvuri1603: same as above

backend/Dockerfile
sduvvuri1603: Updating Title

.github/workflows/e2e-test-frontend.yml
sduvvuri1603: Nvm, changes were small enough to fold into this PR—latest push now moves the cleanup into create-cluster and updates the workflows accordingly. Thanks again for the suggestion!

.github/workflows/e2e-test-frontend.yml
sduvvuri1603: Nvm, changes were small enough to fold into this PR—latest push now moves the cleanup into create-cluster and updates the workflows accordingly. Thanks again for the suggestion!

.github/actions/check-artifact-exists/check_artifact.py
sduvvuri1603: @hbelmiro Downloaded artifacts were blocking the cleanup tests too, so I rolled that safeguard into this PR with the description/title updated accordingly( Lmk if you’d like a different phrasing.). 

Happy to split if you prefer, but this keeps the fix self-contained.

PR #12572, opened 2025-12-22
Update CA Cert Conditional

alyssacgoins:
Update the CA cert logic on driver/launcher so that if `CABUNDLE_SECRET_NAME` is set, even if pod-to-pod TLS is not enabled, the cert will be passed into the launcher.

`backend/test/compiler/argo_ginkgo_test.go` is updated to fix env var testing, in order to validate the change above.

**Checklist:**

### DIFF
--- a/backend/src/apiserver/common/config.go
+++ b/backend/src/apiserver/common/config.go
@@ -34,6 +34,8 @@ const (
 	TokenReviewAudience                     string = "TOKEN_REVIEW_AUDIENCE"
 	MetadataTLSEnabled                      string = "METADATA_TLS_ENABLED"
 	CaBundleSecretName                      string = "CABUNDLE_SECRET_NAME"
+	CaBundleConfigMapName                   string = "CABUNDLE_CONFIGMAP_NAME"
+	CaBundleKeyName                         string = "CABUNDLE_KEY_NAME"
 	RequireNamespaceForPipelines            string = "REQUIRE_NAMESPACE_FOR_PIPELINES"
 	CompiledPipelineSpecPatch               string = "COMPILED_PIPELINE_SPEC_PATCH"
 	MLPipelineServiceName                   string = "ML_PIPELINE_SERVICE_NAME"
@@ -152,6 +154,14 @@ func GetMetadataTLSEnabled() bool {
 
 func GetCaBundleSecretName() string {
 	return GetStringConfigWithDefault(CaBundleSecretName, "")
+}
+
+func GetCABundleKey() string {
+	return GetStringConfigWithDefault(CaBundleKeyName, "")
+}
+
+func GetCaBundleConfigMapName() string {
+	return GetStringConfigWithDefault(CaBundleConfigMapName, "")
 }
 
 func GetCompiledPipelineSpecPatch() string {
--- a/backend/src/v2/compiler/argocompiler/common.go
+++ b/backend/src/v2/compiler/argocompiler/common.go
@@ -55,28 +55,47 @@ var commonEnvs = []k8score.EnvVar{{
 
 // ConfigureCustomCABundle adds CABundle environment variables and volume mounts if CABUNDLE_SECRET_NAME is set.
 func ConfigureCustomCABundle(tmpl *wfapi.Template) {
-	caBundleDir := common.CABundleDir
 	caBundleSecretName := common.GetCaBundleSecretName()
-	if caBundleSecretName == "" {
-		glog.Error("Env var CABUNDLE_SECRET_NAME is blank or empty. Failed to configure custom CA bundle.")
+	caBundleConfigMapName := common.GetCaBundleConfigMapName()
+	// If CABUNDLE_KEY_NAME is not set, use "ca.crt".
+	caBundleKeyName := common.GetCABundleKey()
+	if caBundleKeyName == "" {
+		caBundleKeyName = "ca.crt"
+	}
+	volumeSource := k8score.VolumeSource{}
+
+	// CABUNDLE_SECRET_NAME is prioritized above CABUNDLE_CONFIGMAP_NAME.
+	if caBundleSecretName != "" { // nolint:gocritic // ifElseChain is preferred here for clarity over a switch
+		volumeSource.Secret = &k8score.SecretVolumeSource{
+			SecretName: caBundleSecretName,
+			Items: []k8score.KeyToPath{
+				{
+					Key:  caBundleKeyName,
+					Path: "ca.crt",
+				},
+			},
+		}
+	} else if caBundleConfigMapName != "" {
+		volumeSource.ConfigMap = &k8score.ConfigMapVolumeSource{
+			LocalObjectReference: k8score.LocalObjectReference{Name: caBundleConfigMapName},
+			Items: []k8score.KeyToPath{
+				{
+					Key:  caBundleKeyName,
+					Path: "ca.crt",
+				},
+			},
+		}
+	} else {
+		glog.Error("Neither CABUNDLE_SECRET_NAME nor CABUNDLE_CONFIGMAP_NAME is set. Failed to configure custom CA bundle.")
 		return
 	}
 	volume := k8score.Volume{
-		Name: "ca-secret",
-		VolumeSource: k8score.VolumeSource{
-			Secret: &k8score.SecretVolumeSource{
-				SecretName: caBundleSecretName,
-			},
-		},
+		Name:         "custom-ca",
+		VolumeSource: volumeSource,
 	}
-
 	tmpl.Volumes = append(tmpl.Volumes, volume)
 
-	volumeMount := k8score.VolumeMount{
-		Name:      "ca-secret",
-		MountPath: caBundleDir,
-	}
-
+	volumeMount := k8score.VolumeMount{Name: "custom-ca", MountPath: common.CABundleDir}
 	tmpl.Container.VolumeMounts = append(tmpl.Container.VolumeMounts, volumeMount)
 
 }
--- a/backend/src/v2/compiler/argocompiler/container.go
+++ b/backend/src/v2/compiler/argocompiler/container.go
@@ -224,8 +224,9 @@ func (c *workflowCompiler) addContainerDriverTemplate() string {
 	}
 
 	setCABundle := false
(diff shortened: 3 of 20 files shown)


backend/src/v2/compiler/argocompiler/container.go
```diff
@@ -224,7 +224,7 @@ func (c *workflowCompiler) addContainerDriverTemplate() string {
 	}
 
 	setCABundle := false
-	if common.GetCaBundleSecretName() != "" && (c.mlPipelineTLSEnabled || common.GetMetadataTLSEnabled()) {
+	if common.GetCaBundleSecretName() != "" || c.mlPipelineTLSEnabled || common.GetMetadataTLSEnabled() {
```
mprahl: I think we still need to guarantee that the secret name is set.
```suggestion
	if common.GetCaBundleSecretName() != "" {
```

backend/src/v2/compiler/argocompiler/container.go
```diff
@@ -551,7 +551,7 @@ func (c *workflowCompiler) addContainerExecutorTemplate(task *pipelinespec.Pipel
 		},
 	}
 	// If the apiserver is TLS-enabled, add the custom CA bundle to the executor.
-	if common.GetCaBundleSecretName() != "" && (c.mlPipelineTLSEnabled || common.GetMetadataTLSEnabled()) {
+	if common.GetCaBundleSecretName() != "" || c.mlPipelineTLSEnabled || common.GetMetadataTLSEnabled() {
```
mprahl: I think we still need to guarantee that the secret name is set.
```suggestion
	if common.GetCaBundleSecretName() != "" {
```

hbelmiro (commented):
We should do the same in `backend/src/v2/compiler/argocompiler/dag.go` and `backend/src/v2/compiler/argocompiler/importer.go`.

backend/src/v2/compiler/argocompiler/common.go
```diff
@@ -55,28 +55,33 @@ var commonEnvs = []k8score.EnvVar{{
...
+	// CABUNDLE_SECRET_NAME is prioritized above CABUNDLE_CONFIGMAP_NAME.
 	caBundleSecretName := common.GetCaBundleSecretName()
-	if caBundleSecretName == "" {
-		glog.Error("Env var CABUNDLE_SECRET_NAME is blank or empty. Failed to configure custom CA bundle.")
+	caBundleConfigMapName := common.GetCaBundleConfigMapName()
+	if caBundleSecretName != "" {
+		volumeSource.Secret = &k8score.SecretVolumeSource{SecretName: caBundleSecretName}
+	} else if caBundleConfigMapName != "" {
+		caBundleConfigMapKey := common.GetCaBundleConfigMapKey()
+		if caBundleConfigMapKey == "" {
```
mprahl: I think we should align the behavior between Secret and ConfigMap such that:
1. Prefer Secret over ConfigMap (like your code)
2. We have one configuration option of `CABUNDLE_KEY_NAME` that applies to both Secret and ConfigMap.
3. If `CABUNDLE_KEY_NAME` is not set, then don't set `SubPath` on either the Secret or ConfigMap (matches existing Secret behavior).

alyssacgoins: Updated

backend/src/apiserver/common/config.go
```diff
@@ -154,6 +156,14 @@ func GetCaBundleSecretName() string {
 	return GetStringConfigWithDefault(CaBundleSecretName, "")
 }
 
+func GetCaBundleConfigMapKey() string {
```
mprahl:
```suggestion
func GetCABundleKey() string {
```

backend/src/v2/compiler/argocompiler/common.go
```diff
@@ -55,28 +55,31 @@ var commonEnvs = []k8score.EnvVar{{
...
 	tmpl.Volumes = append(tmpl.Volumes, volume)
 
-	volumeMount := k8score.VolumeMount{
-		Name:      "ca-secret",
-		MountPath: caBundleDir,
+	// If CABUNDLE_KEY_NAME is set, set its value to the VolumeMount subpath.
+	caBundleKeyName := common.GetCABundleKey()
+	volumeMount := k8score.VolumeMount{Name: "custom-ca", MountPath: common.CABundleDir}
+	if caBundleKeyName != "" {
+		volumeMount.SubPath = caBundleKeyName
```
mprahl: I don't think this logic is quite right. I think we want to do something like this instead where the key can change but the mounted path is always the same. Consider using a constant variable for `ca.crt`.

```diff
diff --git a/backend/src/v2/compiler/argocompiler/common.go b/backend/src/v2/compiler/argocompiler/common.go
index 8e4c8e081..6112679e3 100644
--- a/backend/src/v2/compiler/argocompiler/common.go
+++ b/backend/src/v2/compiler/argocompiler/common.go
@@ -58,12 +58,32 @@ func ConfigureCustomCABundle(tmpl *wfapi.Template) {
 	caBundleSecretName := common.GetCaBundleSecretName()
 	caBundleConfigMapName := common.GetCaBundleConfigMapName()
 	volumeSource := k8score.VolumeSource{}
+	caBundleKeyName := common.GetCABundleKey()
+	if caBundleKeyName == "" {
+		caBundleKeyName = "ca.crt"
+	}
 
 	// CABUNDLE_SECRET_NAME is prioritized above CABUNDLE_CONFIGMAP_NAME.
 	if caBundleSecretName != "" { // nolint:gocritic // ifElseChain is preferred here for clarity over a switch
-		volumeSource.Secret = &k8score.SecretVolumeSource{SecretName: caBundleSecretName}
+		volumeSource.Secret = &k8score.SecretVolumeSource{
+			SecretName: caBundleSecretName,
+			Items: []k8score.KeyToPath{
+				{
+					Key:  caBundleKeyName,
+					Path: "ca.crt",
+				},
+			},
+		}
 	} else if caBundleConfigMapName != "" {
-		volumeSource.ConfigMap = &k8score.ConfigMapVolumeSource{LocalObjectReference: k8score.LocalObjectReference{Name: caBundleConfigMapName}}
+		volumeSource.ConfigMap = &k8score.ConfigMapVolumeSource{
+			LocalObjectReference: k8score.LocalObjectReference{Name: caBundleConfigMapName},
+			Items: []k8score.KeyToPath{
+				{
+					Key:  caBundleKeyName,
+					Path: "ca.crt",
+				},
+			},
+		}
 	} else {
 		glog.Error("Neither CABUNDLE_SECRET_NAME nor CABUNDLE_CONFIGMAP_NAME is set. Failed to configure custom CA bundle.")
 		return
@@ -75,13 +95,8 @@ func ConfigureCustomCABundle(tmpl *wfapi.Template) {
 	tmpl.Volumes = append(tmpl.Volumes, volume)
 
 	// If CABUNDLE_KEY_NAME is set, set its value to the VolumeMount subpath.
-	caBundleKeyName := common.GetCABundleKey()
 	volumeMount := k8score.VolumeMount{Name: "custom-ca", MountPath: common.CABundleDir}
-	if caBundleKeyName != "" {
-		volumeMount.SubPath = caBundleKeyName
-	}
 	tmpl.Container.VolumeMounts = append(tmpl.Container.VolumeMounts, volumeMount)
-
 }
 
 // addExitTask adds an exit lifecycle hook to a task if exitTemplate is not empty.
```

alyssacgoins: Updated

mprahl (approved):
/approve
/lgtm

backend/test/compiler/argo_ginkgo_test.go
```diff
@@ -39,31 +39,78 @@ var _ = BeforeEach(func() {
...
 	}{
 		{
-			compilerOptions: argocompiler.Options{CacheDisabled: true},
+			compilerOptions:   argocompiler.Options{CacheDisabled: true},
+			pipelineFilePaths: allPipelineFiles,
+		},
+		{
+			compilerOptions:   argocompiler.Options{CacheDisabled: true},
+			envVars:           map[string]string{"PIPELINE_RUN_AS_USER": "1001", "PIPELINE_LOG_LEVEL": "3"},
+			pipelineFilePaths: []string{filepath.Join(pipelineFilesRootDir, pipelineDirectory, "run_as_user_cache_disabled.yaml")},
```
nsingla, line 56: I am confused about this, why do you need a separate yaml? can the existing pipelines not run as a user and with a different log level? if so, we can always set the required values in the expected compiled workflow object

backend/test/compiler/argo_ginkgo_test.go
```diff
@@ -39,31 +39,78 @@ var _ = BeforeEach(func() {
...
+		{
+			compilerOptions:   argocompiler.Options{CacheDisabled: false},
+			pipelineFilePaths: allPipelineFiles,
 		},
 		{
-			compilerOptions: argocompiler.Options{CacheDisabled: true},
-			envVars:         map[string]string{"PIPELINE_RUN_AS_USER": "1001", "PIPELINE_LOG_LEVEL": "3"},
+			compilerOptions:   argocompiler.Options{CacheDisabled: false},
+			envVars:           map[string]string{"PIPELINE_RUN_AS_USER": "1001", "PIPELINE_LOG_LEVEL": "3"},
+			pipelineFilePaths: []string{filepath.Join(pipelineFilesRootDir, pipelineDirectory, "run_as_user_cache_enabled.yaml")},
```
nsingla, line 65: same here

backend/test/compiler/argo_ginkgo_test.go
alyssacgoins: A separate YAML is necessary because the environment variables update the contents of the YAML - for example, the variable `PIPELINE_RUN_AS_USER` adds  `  securityContext: runAsUser: 1001` to the compiled workflow. But we can't update the general YAML with this specialized case, because then the test cases with no env variables set will fail.

nsingla: in that case, we can check if we need to update or not based on the environment variable

backend/test/compiler/argo_ginkgo_test.go
alyssacgoins: Addressed above ^

PR #12574, opened 2025-12-22
feat(backend): add propagation policy option to DeleteRecurringRun. Fixes #7421, fixes #12573

### LINKED ISSUE #7421: [backend] "Updating" a recurring run, whilst a recurring run is in progress causes it to terminate
### Environment

Atop AWS using the `kubeflow/manifests` official distribution with some overlays specific to Zillow.

*  KFP version: 

We use KF v1.3.1 currently (plan to update to v1.5 next month), looks like that includes KFP v1.5.1 ([link](https://github.com/kubeflow/manifests/blob/v1.3.1/apps/pipeline/upstream/base/pipeline/kustomization.yaml#L43)).

*  KFP SDK version: 

We run a forked version ([see here](https://github.com/zillow/pipelines)), I think we last pulled in at ~v1.8.

### Steps to reproduce

The problem is that via CICD we want our scientists to be able to define a `recurring-run` and have it updated on subsequent CICD executions if they change their settings. Given the limitations of the "Job" API in KFP we delete and re-add to "update" a recurring run, (see the [API reference](https://www.kubeflow.org/docs/components/pipelines/reference/api/kubeflow-pipeline-api-spec/#tag-JobService) for more info).

What happens in this case if the first recurring run has triggered a workflow then the scheduledworkflow controller deletes the `ScheduledWorkflow` and then via Kubernetes garbage collection it cascades and deletes any in-progress `Workflow` resources. Then when the "new" updated recurring run is created it needs to wait for the new trigger point to schedule the next run, it won't try to resurrect the now terminated run (and it definitely shouldn't thinking about it).

### Expected result

Basically some way of keeping the run that was in progress before the "update" occurred.

Potentially when deleting a recurring run if there was a flag available in the API/SDK/CLI to say to [orphan on deletion](https://kubernetes.io/docs/tasks/administer-cluster/use-cascading-deletion/#set-orphan-deletion-policy) that'd solve this issue and you'd not need a complicated update API.

### Materials and Reference

Apologies, maybe this is more of a "feature" request thinking about it now. Apologies for the mischaracterization!

---

Impacted by this bug? Give it a 👍. We prioritise the issues with the most 👍.

### LINKED ISSUE #12573: [backend] Deleting Recurring Runs Forces Running Workflows to be Terminated
### Environment

KFP version: 2.15.0
KFP SDK version: 2.15.0

### Steps to reproduce

When deleting a recurring run via the API, the ScheduledWorkflow CRD is deleted with Kubernetes default cascade behavior. This causes any in-progress workflows (which have an ownerReference to the ScheduledWorkflow) to be terminated immediately

This is problematic for CI/CD workflows that need to update recurring runs by deleting and recreating them since the delete kills any currently running pipeline

This is also been brought up in this stale issue #7421 but does not seem to ever be addressed

### Expected result

There should be a way to set the deletion policy when calling delete recurring runs, this way users can select to orphan current workflows to keep them running

---

JerT33:

Add `propagation_policy` option to the `DeleteRecurringRun` API endpoint.

 This allows users to control Kubernetes deletion behavior when deleting a recurring run:
  - `FOREGROUND` - Delete child workflows first, then parent
  - `BACKGROUND` - Delete parent immediately, garbage collect children async
  - `ORPHAN` - Delete parent, leave child workflows running

**If the propagation_policy is not specified in the api call, then no propagation policy will be set in the delete options when deleting the scheduled workflow. This is the current behavior which results in background deletion (Kubernetes default)**

Live Cluster Testing:
`apis/v2beta1/recurringruns/<run_id>?propagation_policy=ORPHAN`
(Workflow is still running, Recurring run + SWF deleted)
Screenshot 2025-12-22 at 8 41 44 PM

Notes: 
- The majority of these file changes are auto-generated. The files with manual changes are the following:
  - `backend/src/apiserver/resource/resource_manager.go`
    - Added a `propagationPolicyMap` variable that maps api enum values to Kubernetes `DeletionPropagation` values
    - Updated `DeleteJob` to accept a `propagationPolicy` parameter and apply it to the Kubernetes delete options when deleting the scheduled workflow (if the `propagationPolicy` is valid)
  - `backend/src/apiserver/resource/resource_manager_test.go`
    - Added tests for each propagation policy option
    - Updated existing tests to pass the new required parameter 
  - `backend/src/apiserver/server/job_server.go`
    - Updated `deleteJob` and `DeleteRecurringRun` handlers to pass the propagation policy through to the resource manager (defaulted v1 api to UNSPECIFIED)
  - `backend/api/v2beta1/recurring_run.proto`
     - Added `DeletePropagationPolicy` enum
     - Added optional `propagation_policy` field to `DeleteRecurringRunRequest`

- My editor automatically removes ending whitespace upon saving, this has caused some additional diffs, these can be reverted if requested 

**Checklist:**


droctothorpe:
/lgtm

Thank you for the contribution, @JerT33! This is great. Would appreciate having another set of eyes on it. @zazulam @HumairAK @mprahl whenever you have a chance.

droctothorpe:
Bump @zazulam @mprahl

backend/src/apiserver/resource/resource_manager.go
```diff
@@ -1243,30 +1251,36 @@ func (r *ResourceManager) ChangeJobMode(ctx context.Context, jobId string, enabl
 }
 
 // Deletes a recurring run with given id.
-func (r *ResourceManager) DeleteJob(ctx context.Context, jobId string) error {
-	job, err := r.GetJob(jobId)
+func (r *ResourceManager) DeleteJob(ctx context.Context, jobID string, propagationPolicy apiv2beta1.DeletePropagationPolicy) error {
+	job, err := r.GetJob(jobID)
 	if err != nil {
```
zazulam, line 1256: nit: I see that other functions/methods utilize `JobId` and here it's updated. I _would prefer_ for it to be reverted.

zazulam: I won't let this be a blocker.

Issue #11760, opened 2025-03-19
chore(backend/frontend): Remove MLMD as a depedency of KFP


HumairAK:
## Description

[MLMD](https://github.com/google/ml-metadata) is effectively no longer supported. As it is such a critical part of KFP, this is not great for KFP outlook and we need to prioritize dropping this dependency. The two options I see are: 

1. Find an alternative (which?) 
2. Directly bake into KFP and replace the functionality required to support the MLMD dependent code
3. ??

Not sure how feasible (2) is yet, this will require some investigation.

HumairAK:
fyi @chensun

andreyvelich:
Thank you for starting this @HumairAK!

I recall that during our discussions about the Model Registry proposal, one of the key objectives was to replace KFP MLMD with it: https://docs.google.com/document/d/1T3KfOqIfJohp0s1koQ2XrJJQQhj7TECO-m2xPsW59_c/edit?tab=t.0.
Do we know where we are with that, or I misunderstood something  ?
cc @kubeflow/wg-data-leads 

On the side note, I think we should discuss a reusable solution that can serve multiple Kubeflow projects. Throughout the entire AI/ML lifecycle, a vast amount of metadata is generated, and we might want to have a standardized, portable, and extensible way to manage it.
For example:
- Pre-trained models, datasets, metrics, and logs from the Kubeflow TrainJobs
- Models from Trials, hyperparameters, and neural architectures from the Kubeflow OptimizeJob (e.g. Katib/Optimizer/AutoML)

It would be great to explore how we can consolidate efforts around MLMD across different Kubeflow projects.

cc @kubeflow/wg-training-leads @shravan-achar @akshaychitneni @bigsur0

Would love to hear your thoughts!

franciscojavierarceo:
Coincidentally, I discussed this with @crazyglue and @syntaxsdev and we also talked about the Feast and KFP registry so I agree @andreyvelich!

rareddy:
@andreyvelich @franciscojavierarceo We recently initiated a similar proposal in Model Registry: https://github.com/kubeflow/model-registry/issues/865.

Originally, the goal was to adopt MLMD from KFP into Model Registry, but that approach lost traction due to missing features like multi-tenancy. We considered contributing to MLMD to extend its capabilities, but that wasn’t feasible. Next, we explored rewriting the MLMD backend in Golang, but given Model Registry’s usage patterns, that didn’t seem like the right approach.

[...]

PR #12596, opened 2025-12-31
feat(#12169): Show retry attempts for pipeline tasks in UI

### LINKED ISSUE #12169: [feature] Show retry attempts for pipeline tasks in the UI
### Feature Area

/area frontend
/area backend

When a pipeline task is configured with .set_retry(...), the Kubeflow Pipelines UI should display retry attempts clearly in the Run Details view.

For example:
- Each retry attempt (Attempt 1, Attempt 2, …) should be listed.
- Users should be able to view logs and execution details for each attempt.
- Optionally, in the DAG view, annotate the task node with the number of retries executed (e.g., train [3 attempts]).

This behavior would be similar to Airflow’s UI, where retries are visible and each attempt’s logs can be inspected separately.

Currently, the UI shows only the final attempt’s status and logs. Users cannot easily tell if a task succeeded on the first attempt or only after multiple retries.
This makes it difficult to debug transient failures or understand pipeline reliability. Developers and operators have to dig into Kubernetes pods/events or backend logs to find retry information.

The only workaround is to manually inspect Kubernetes pods?

---

Love this idea? Give it a 👍.

Sapthagiri777:

Added "Retry Attempts" field to Task Details table in `RuntimeNodeDetailsV2.tsx`.

Displays `retry_count/max_retries` from execution custom properties for pipeline tasks with `.set_retry()`. 

Addresses core #12169 request: makes retry information visible in Run Details UI.

**Before:** Only final attempt shown  
**After:** `Retry Attempts: 2/3` in Task Details table[1]

**Checklist:**
- [x] The title for your pull request (PR) should follow our title convention. [Learn more about the pull request title convention used in this repository](https://[github](https://github.com/kubeflow/pipelines/issues/12169).com/kubeflow/pipelines/blob/master/CONTRIBUTING.md#pull-request-title-convention).

/area frontend

Closes #12169

### DIFF
--- a/frontend/src/components/tabs/RuntimeNodeDetailsV2.tsx
+++ b/frontend/src/components/tabs/RuntimeNodeDetailsV2.tsx
@@ -242,9 +242,17 @@ function getTaskDetailsFields(
 
       const createdAt = new Date(execution.getCreateTimeSinceEpoch()).toString();
       details.push(['Created At', createdAt]);
-      
-      const retryCount = execution?.getCustomPropertiesMap().get('retry_count')?.getStringValue() || '0';
-      const maxRetries = execution?.getCustomPropertiesMap().get('max_retries')?.getStringValue() || '∞';
+
+      const retryCount =
+        execution
+          ?.getCustomPropertiesMap()
+          .get('retry_count')
+          ?.getStringValue() || '0';
+      const maxRetries =
+        execution
+          ?.getCustomPropertiesMap()
+          .get('max_retries')
+          ?.getStringValue() || '∞';
       details.push(['Retry Attempts', `${retryCount}/${maxRetries}`]);
 
       const lastUpdatedTime = execution.getLastUpdateTimeSinceEpoch();


droctothorpe:
/ok-to-test

This is awesome, @Sapthagiri777 ! Thanks for the contribution! Can you post a screenshot of what it looks like?

droctothorpe:
Make sure to run `npm run format`: https://github.com/kubeflow/pipelines/actions/runs/20616308128/job/59226994150?pr=12596

Sapthagiri777:
@droctothorpe, thanks for the positive feedback!
I ran into some local environment issues while trying to spin up the frontend, so I couldn't grab a screenshot just yet. The change basically adds a "Retry Attempts: 2/3" style row to the Task Details table whenever a task has retry info available.
I'll try to get a screenshot up once I sort out my setup, but feel free to check it out when you're reviewing if that works better for you. Let me know if you need anything else!

jeffspahr:
## Code Review

### Overview

This PR adds a "Retry Attempts" field to the Task Details table in `RuntimeNodeDetailsV2.tsx`, reading `retry_count` and `max_retries` from MLMD execution custom properties and displaying them as `retryCount/maxRetries`.

### Blocking Issue: Backend Does Not Populate These Properties

The custom properties `retry_count` and `max_retries` are **never written to MLMD executions** by the backend. Tracing through the code:

- The SDK's `set_retry()` stores `max_retry_count` in the pipeline spec's `RetryPolicy` protobuf.
- The Argo compiler translates this into Argo workflow retry parameters (`retry-max-count`, `retry-backoff-duration`, etc.) in `backend/src/v2/compiler/argocompiler/container.go`.
- These parameters end up inside the execution's `"inputs"` custom property as a nested struct — **not** as top-level custom properties named `retry_count` or `max_retries`.
- There is **no property tracking the current retry attempt number** on the execution. The driver/launcher don't know what attempt they're on — Argo's `{{retries}}` template variable is never used by KFP.

**Result:** Both `.get('retry_count')` and `.get('max_retries')` will always return `undefined`, so the UI will always display `0/∞` for every task. This needs backend work to populate these properties before the frontend can read them.

### Additional Issues

**1. Unconditional display for all tasks**
The retry field renders for every execution, including tasks with no retry policy. Showing `0/∞` on tasks without `.set_retry()` is confusing. It should only display when retry metadata is present.

**2. Potential type mismatch: `getStringValue()` vs `getIntValue()`**
Retry counts are logically integers. If/when the backend stores them, they'd likely use `setIntValue()`. The PR uses `getStringValue()`, which returns empty string for int-typed values. The existing code in this file uses `getStringValue()` because it reads string properties (`display_name`, `pod_name`, `namespace`), but numeric properties would need `getIntValue()`.

**3. Unnecessary optional chaining**
```typescript
execution
  ?.getCustomPropertiesMap()
  .get('retry_count')
```
The `?.` on `execution` is unnecessary — this code is inside an `if (execution)` guard (line 230). All other property reads in this block use `execution.getCustomPropertiesMap()` without `?.`. Minor, but breaks the existing code style.

**4. No test coverage**
The existing test file (`RuntimeNodeDetailsV2.test.tsx`) has no retry-related tests. A new UI field should have tests for:
- Field renders when properties are present
- Behavior when properties are absent
- Expected display format

**5. Fallback value `'∞'` for max_retries**
Using the infinity symbol when no max is configured implies unlimited retries, which isn't how the system works. If max retries isn't set, the task has no retry policy and the field shouldn't render.

### Summary

| Area | Assessment |
|------|-----------|
| Correctness | Properties don't exist in backend — always shows `0/∞` |
| Code style | Minor inconsistency (unnecessary `?.`) |
| Tests | None added |
| UX | Shows unconditionally; misleading fallback values |

### What's Needed

This feature requires a small backend change first: the driver or launcher needs to write retry metadata as MLMD custom properties on the execution. Specifically, using Argo's `{{retries}}` template variable to get the current attempt number, and reading `max_retry_count` from the pipeline spec's `RetryPolicy`. Once those properties exist, the frontend change here (with the fixes above) would work.

jeffspahr:
## Context: MLMD Removal and KFP 3.0

For awareness — MLMD is being removed from KFP as part of the **KFP 3.0 milestone** (target: May 2026). The relevant tracking:

- **#11760** — Remove MLMD as a dependency of KFP (on the [KFP 3.0 milestone](https://github.com/kubeflow/pipelines/milestone/21))
- **#12430** — Replace MLMD with KFP Server APIs (the implementation PR, ~550 files)
- **KEP:** `proposals/12147-mlmd-removal/`

PR #12430 explicitly calls out `RuntimeNodeDetailsV2.tsx` as a file that will be rewritten to stop using MLMD. The entire `execution.getCustomPropertiesMap()` pattern this PR builds on will be replaced — post-MLMD, task metadata comes from structured fields on the KFP API's `PipelineTaskDetail` response, not from MLMD execution objects.

### What This Means for This PR

This feature can still land in a 2.x release if the backend work is done soon. The MLMD removal targets 3.0, so any 2.x release between now and then would include it. But the window is small.

The work breaks down into two small changes:

1. **Backend:** Have the driver or launcher write `retry_count` (from Argo's `{{retries}}` template variable) and `max_retry_count` (from the pipeline spec `RetryPolicy`) as MLMD custom properties on the execution. This is a small change in `backend/src/v2/driver/container.go` or `backend/src/v2/component/launcher_v2.go`.

2. **Frontend (this PR, with fixes):** Read those properties, display conditionally only when present, use correct value types, add tests.

### For 3.0

When the MLMD removal lands, the retry feature would need to be re-implemented against the new architecture:
- Retry fields added to the `Task` proto / `PipelineTaskDetail` response
- Driver/launcher report retry info via the new KFP API (instead of MLMD custom properties)
- Frontend reads from the task API response

This is straightforward and would be part of the broader `RuntimeNodeDetailsV2.tsx` rewrite that #12430 requires anyway. The feature request in #12169 is valid for both architectures — the question is just which data source the frontend reads from.

jeffspahr:
@Sapthagiri777 FYI I submitted a PR for the backend changes to complement this.  If you want to land this, it will probably need to be fixed and merged in the next week. Please let me know if you're still working on this.

jeffspahr:
@Sapthagiri777 I'm going to close both of our PRs since we missed the 2.16.0 release and mlmd is about to be removed.  If you want to attempt addressing this issue after the mlmd changes are merged, that'd be amazing.  The issue will remain open.

/close

PR #12606, opened 2026-01-04
feat(sdk): add support for Literal input parameters in compiler

### LINKED ISSUE #12601: [feature] KEP-11385: Extend the SDK compiler to support Literal input parameters for both pipeline and component input
### Feature Area

/area sdk

Part of [KEP-11385](https://github.com/kubeflow/pipelines/blob/master/proposals/11385-literal-input-parameters/README.md).

### Acceptance Criteria

 * `InputSpec` class definition contains a `literals` field of type `Optional[List[str] | List[int] | List[float]] = None`
 * Compiler throws an error for Literal parameters that contain elements of multiple types (ie. `Literal[1, 2, "a"]`). While this behavior is supported by the `typing.Literal` class, we want to limit KFP to Literals of a single type.
 * Compiler validates hard-coded input values against their Literal parameters. Compiler throws an error if input is invalid
 * Validate these changes by adding the [valid](https://github.com/kubeflow/pipelines/blob/master/proposals/11385-literal-input-parameters/README.md?plain=1#L178) and [invalid](https://github.com/kubeflow/pipelines/blob/master/proposals/11385-literal-input-parameters/README.md?plain=1#L178) cases outlined in the KEP to the SDK compiler unit test cases.  Only include behavior outlined under the *SDK Compiler* column.

Hint: changes should be made primarily to  `kfp/dsl/structures.py` and `kfp/dsl/types/type_utils.py`

shreeharshshinde:
### What this PR does
Adds SDK compiler support for `typing.Literal` input parameters, enabling validation of allowed constant values and enforcing single-type Literal definitions.

### Key changes
- Extend `InputSpec` with `literals` field
- Parse `typing.Literal` annotations into `(type, literals)` during compilation
- Reject mixed-type Literal definitions at component definition time
- Validate constant inputs against allowed Literal values during type compatibility checks
- Add compiler unit tests covering valid and invalid Literal usage

### Scope
- SDK compiler only
- No runtime or backend changes

### Related
- Part of KEP-11385

Fixes #12601

### DIFF
--- a/sdk/python/kfp/compiler/compiler_test.py
+++ b/sdk/python/kfp/compiler/compiler_test.py
@@ -19,7 +19,7 @@ import re
 import subprocess
 import tempfile
 import textwrap
-from typing import Any, Dict, List, NamedTuple, Optional
+from typing import Any, Dict, List, Literal, NamedTuple, Optional
 import unittest
 
 from absl.testing import parameterized
@@ -186,6 +186,53 @@ def print_artifact(a: Input[Artifact]):
 
 
 class TestCompilePipeline(parameterized.TestCase):
+
+    def test_compile_pipeline_with_literal_input_valid(self):
+
+        @dsl.component
+        def literal_component(literal_param: Literal['a', 'b']):
+            print_op(message=literal_param)
+
+        @dsl.pipeline(name='test-literal-pipeline')
+        def literal_pipeline():
+            literal_component(literal_param='a')
+
+        with tempfile.TemporaryDirectory() as tmpdir:
+            target_file = os.path.join(tmpdir, 'result.yaml')
+            compiler.Compiler().compile(
+                pipeline_func=literal_pipeline,
+                package_path=target_file,
+            )
+            self.assertTrue(os.path.exists(target_file))
+            with open(target_file) as f:
+                data = yaml.safe_load(f)
+            param = data['components']['comp-literal-component'][
+                'inputDefinitions']['parameters']['literal_param']
+            self.assertIn('literals', param)
+            self.assertEqual(param['literals'], ['a', 'b'])
+
+    def test_compile_pipeline_with_literal_input_mixed_types_invalid(self):
+        # Mixed-type Literal should be rejected at component definition time
+        with self.assertRaises(TypeError):
+
+            @dsl.component
+            def literal_component_with_mixed_types(literal_param: Literal[1, 2,
+                                                                          'a']):
+                print_op(message=literal_param)
+
+    def test_compile_pipeline_with_literal_input_invalid_value(self):
+        # Compilation should fail when a constant argument is not in the allowed Literal set
+        @dsl.component
+        def literal_component(literal_param: Literal['a', 'b']):
+            print_op(message=literal_param)
+
+        # The validation happens when the pipeline is defined (not during compile)
+        # because the invalid value is detected when calling the component
+        with self.assertRaises(type_utils.InconsistentTypeException):
+
+            @dsl.pipeline(name='test-literal-pipeline-invalid')
+            def literal_pipeline():
+                literal_component(literal_param='c')
 
     def test_can_use_dsl_attribute_on_kfp(self):
 
--- a/sdk/python/kfp/compiler/compiler_test.py
+++ b/sdk/python/kfp/compiler/compiler_test.py
@@ -234,6 +234,19 @@ class TestCompilePipeline(parameterized.TestCase):
             def literal_pipeline():
                 literal_component(literal_param='c')
 
+    def test_compile_pipeline_with_disjoint_literal_channel_raises(self):
+        # Fix 3: pipeline input Literal['c'] wired into component Literal['a','b'] must raise
+
+        @dsl.component(base_image='python:3.11')
+        def consume(x: Literal['a', 'b']):
+            pass
+
+        with self.assertRaises(type_utils.InconsistentTypeException):
+
+            @dsl.pipeline(name='test-disjoint-literal-pipeline')
+            def p(x: Literal['c']):
+                consume(x=x)
+
     def test_can_use_dsl_attribute_on_kfp(self):
 
         @kfp.dsl.pipeline
@@ -470,7 +483,7 @@ class TestCompilePipeline(parameterized.TestCase):
 
         @dsl.pipeline(description='Prefer me.')
         def my_pipeline():
-            """Don't prefer me"""
+            """Don't prefer me."""
             VALID_PRODUCER_COMPONENT_SAMPLE(input_param='input')
 
         self.assertEqual(my_pipeline.pipeline_spec.pipeline_info.description,
@@ -492,7 +505,8 @@ class TestCompilePipeline(parameterized.TestCase):
         def my_pipeline():
             """Docstring-specified description.
 
-            More information about this pipeline."""
+            More information about this pipeline.
+            """
             VALID_PRODUCER_COMPONENT_SAMPLE(input_param='input')
 
         self.assertEqual(
@@ -2887,6 +2901,7 @@ class TestYamlComments(unittest.TestCase):
                 sample_input1: bool = True,
                 sample_input2: str = 'string') -> str:
             """docstring short description.
+
             docstring long description. docstring long description.
             """
             op1 = my_comp(string=sample_input2, model=sample_input1)
@@ -2913,10 +2928,9 @@ class TestYamlComments(unittest.TestCase):
         def pipeline_with_multiline_definition(
                 sample_input1: bool = True,
                 sample_input2: str = 'string') -> str:
-            """
-            docstring long description.
-            docstring long description.
-            docstring long description.
+            """docstring long description.
+
+            docstring long description. docstring long description.
             """
             op1 = my_comp(string=sample_input2, model=sample_input1)
             result = op1.output
@@ -2945,8 +2959,8 @@ class TestYamlComments(unittest.TestCase):
         def my_pipeline(sample_input1: bool = True,
                         sample_input2: str = 'string') -> str:
             """docstring short description.
-            docstring long description.
-            docstring long description.
+
+            docstring long description. docstring long description.
             """
             op1 = my_comp(string=sample_input2, model=sample_input1)
             result = op1.output
@@ -4512,7 +4526,8 @@ class TestPlatformConfig(unittest.TestCase):
 
     def test_compile_fails_when_workspace_placeholder_used_without_workspace_config(
             self):
-        """Tests that compilation fails if placeholder is used and no workspace configured."""
+        """Tests that compilation fails if placeholder is used and no workspace
+        configured."""
 
         @dsl.component
         def uses_workspace(workspace_path: str) -> str:
@@ -4540,7 +4555,8 @@ class TestPlatformConfig(unittest.TestCase):
 
     def test_compile_fails_when_workspace_placeholder_used_in_nested_groups_without_workspace_config(
             self):
-        """Tests that compilation fails if placeholder is used within nested groups and no workspace configured."""
+        """Tests that compilation fails if placeholder is used within nested
+        groups and no workspace configured."""
 
         import os
         import tempfile
@@ -4580,7 +4596,8 @@ class TestPlatformConfig(unittest.TestCase):
 
     def test_compile_fails_when_importer_download_to_workspace_without_workspace_config(
             self):
-        """Tests that compilation fails if importer uses download_to_workspace without workspace config."""
+        """Tests that compilation fails if importer uses download_to_workspace
+        without workspace config."""
 
         import os
         import tempfile
@@ -4609,7 +4626,8 @@ class TestPlatformConfig(unittest.TestCase):
 
     def test_compile_succeeds_when_importer_download_to_workspace_with_workspace_config(
             self):
-        """Tests that compilation succeeds with both download_to_workspace and workspace config."""
+        """Tests that compilation succeeds with both download_to_workspace and
+        workspace config."""
 
         import os
         import tempfile
@@ -4635,7 +4653,8 @@ class TestPlatformConfig(unittest.TestCase):
             self.assertTrue(os.path.exists(output_yaml))
 
     def test_pipeline_with_ttl_seconds_after_finished(self):
-        """ttl_seconds_after_finished → resource_ttl_on_completion (SecondsAfterCompletion)."""
+        """ttl_seconds_after_finished → resource_ttl_on_completion
+        (SecondsAfterCompletion)."""
 
         @dsl.pipeline(
             pipeline_config=dsl.PipelineConfig(ttl_seconds_after_finished=300))
@@ -4654,7 +4673,8 @@ class TestPlatformConfig(unittest.TestCase):
         self.assertEqual(loaded_pipeline.platform_spec, expected)
 
     def test_pipeline_with_ttl_seconds_after_success(self):
-        """ttl_seconds_after_success → resource_ttl_on_success (SecondsAfterSuccess)."""
+        """ttl_seconds_after_success → resource_ttl_on_success
+        (SecondsAfterSuccess)."""
 
         @dsl.pipeline(
             pipeline_config=dsl.PipelineConfig(ttl_seconds_after_success=600))
@@ -4672,7 +4692,8 @@ class TestPlatformConfig(unittest.TestCase):
         self.assertEqual(loaded_pipeline.platform_spec, expected)
 
     def test_pipeline_with_ttl_seconds_after_failure(self):
-        """ttl_seconds_after_failure → resource_ttl_on_failure (SecondsAfterFailure)."""
+        """ttl_seconds_after_failure → resource_ttl_on_failure
+        (SecondsAfterFailure)."""
 
         @dsl.pipeline(
             pipeline_config=dsl.PipelineConfig(ttl_seconds_after_failure=120))
@@ -4716,9 +4737,10 @@ class TestPlatformConfig(unittest.TestCase):
     def test_pipeline_with_ttl_zero(self):
         """TTL of 0 is accepted but treated as "not set" and not serialized.
 
-        A value of 0 is indistinguishable from the proto3 int32 default, so the
-        serializer skips it.  The resulting platform spec contains no
-        pipelineConfig entry at all, identical to not setting a TTL.
+        A value of 0 is indistinguishable from the proto3 int32 default,
+        so the serializer skips it.  The resulting platform spec
+        contains no pipelineConfig entry at all, identical to not
+        setting a TTL.
         """
 
         @dsl.pipeline(
@@ -4804,7 +4826,8 @@ class TestPlatformConfig(unittest.TestCase):
         self.assertEqual(loaded_pipeline.platform_spec, expected)
 
     def test_pipeline_with_active_deadline_none_omits_field(self):
-        """active_deadline_seconds=None (default) produces no platform spec entry."""
+        """active_deadline_seconds=None (default) produces no platform spec
+        entry."""
 
         @dsl.pipeline(
             pipeline_config=dsl.PipelineConfig(active_deadline_seconds=None))
@@ -5074,7 +5097,7 @@ class ExtractInputOutputDescription(unittest.TestCase):
             string: str,
             in_artifact: Input[Artifact],
         ) -> Outputs:
-            """Pipeline description. Returns
+            """Pipeline description. Returns.
 
             Args:
                 string: Return Pipeline input string. Returns
@@ -5537,7 +5560,9 @@ class TestDslOneOf(unittest.TestCase):
     # To help narrow the tests further (we already test lots of aspects in the following cases), we choose focus on the dsl.OneOf behavior, not the conditional logic if If/Elif/Else. This is more verbose, but more maintainable and the behavior under test is clearer.
 
     def test_if_else_returned(self):
-        """Uses If and Else branches, parameters passed to dsl.OneOf, dsl.OneOf returned from a pipeline, and different output keys on dsl.OneOf channels."""
+        """Uses If and Else branches, parameters passed to dsl.OneOf, dsl.OneOf
+        returned from a pipeline, and different output keys on dsl.OneOf
+        channels."""
 
         @dsl.pipeline
         def roll_die_pipeline() -> str:
@@ -5598,7 +5623,9 @@ class TestDslOneOf(unittest.TestCase):
         )
 
     def test_if_elif_else_returned(self):
-        """Uses If, Elif, and Else branches, parameters passed to dsl.OneOf, dsl.OneOf returned from a pipeline, and different output keys on dsl.OneOf channels."""
+        """Uses If, Elif, and Else branches, parameters passed to dsl.OneOf,
+        dsl.OneOf returned from a pipeline, and different output keys on
+        dsl.OneOf channels."""
 
         @dsl.pipeline
         def roll_die_pipeline() -> str:
@@ -5673,7 +5700,9 @@ class TestDslOneOf(unittest.TestCase):
         )
 
     def test_if_elif_else_consumed(self):
-        """Uses If, Elif, and Else branches, parameters passed to dsl.OneOf, dsl.OneOf passed to a consumer task, and different output keys on dsl.OneOf channels."""
+        """Uses If, Elif, and Else branches, parameters passed to dsl.OneOf,
+        dsl.OneOf passed to a consumer task, and different output keys on
+        dsl.OneOf channels."""
 
         @dsl.pipeline
         def roll_die_pipeline():
@@ -5750,7 +5779,9 @@ class TestDslOneOf(unittest.TestCase):
         )
 
     def test_if_else_consumed_and_returned(self):
-        """Uses If, Elif, and Else branches, parameters passed to dsl.OneOf, and dsl.OneOf passed to a consumer task and returned from the pipeline."""
+        """Uses If, Elif, and Else branches, parameters passed to dsl.OneOf,
+        and dsl.OneOf passed to a consumer task and returned from the
+        pipeline."""
 
         @dsl.pipeline
         def flip_coin_pipeline() -> str:
@@ -5823,7 +5854,8 @@ class TestDslOneOf(unittest.TestCase):
         )
 
     def test_if_else_consumed_and_returned_artifacts(self):
-        """Uses If, Elif, and Else branches, artifacts passed to dsl.OneOf, and dsl.OneOf passed to a consumer task and returned from the pipeline."""
+        """Uses If, Elif, and Else branches, artifacts passed to dsl.OneOf, and
+        dsl.OneOf passed to a consumer task and returned from the pipeline."""
 
         @dsl.pipeline
         def flip_coin_pipeline() -> Artifact:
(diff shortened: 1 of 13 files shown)


shreeharshshinde:
This PR implements the SDK compiler portion of Literal input support as described in KEP-11385. Happy to adjust based on feedback.

droctothorpe:
/ok-to-test

@shreeharshshinde please make sure your commits are signed. The DCO gate is failing. 

@alyssacgoins can you review this one too so it's the backend side of the excellent literals KEP you submitted? Thank you!

shreeharshshinde:
> /ok-to-test
> 
> @shreeharshshinde please make sure your commits are signed. The DCO gate is failing.
> 
> @alyssacgoins can you review this one too so it's the backend side of the excellent literals KEP you submitted? Thank you!

Thanks @droctothorpe!
I’ve **fixed the DCO issue** by signing my commits.
I also ran the required SDK formatters (`isort`, `yapf`, `docformatter`) under `sdk/python/kfp` to address the **failing CI checks**.
The additional changes are formatting-only and were needed to unblock the pipeline.
Please let me know if you’d prefer the formatting to be split into a separate commit.

sdk/python/kfp/dsl/structures.py
```diff
@@ -45,12 +45,19 @@ class InputSpec:
         type: The type of the input.
         default (optional): the default value for the input.
         optional: Wether the input is optional. An input is optional when it has an explicit default value.
+        literals: Optional list of allowed Literal values for parameter inputs.
```
alyssacgoins:
```suggestion
literals (optional): list of allowed input values.
```
This, or something similar in this format, would be a more clear description

alyssacgoins (changes requested):
@shreeharshshinde Nice work! Can you add a test case to `sdk/python/test/compilation/pipeline_compilation_test.py` that validates successful compilation of a pipeline with Literal inputs? (You will need to add this pipeline as both a python file as well as a compiled IR yaml to `test_data/sdk_compiled_pipelines/valid`).  Once you've added this test I can do a more thorough review of your changes.

In addition - it looks like `KFP SDK Docformatter Test`, `KFP SDK Isort Test` and `KFP SDK YAPF Tests` are failing - can you please fix those?

shreeharshshinde:
> @shreeharshshinde Nice work! Can you add a test case to `sdk/python/test/compilation/pipeline_compilation_test.py` that validates successful compilation of a pipeline with Literal inputs? (You will need to add this pipeline as both a python file as well as a compiled IR yaml to `test_data/sdk_compiled_pipelines/valid`). Once you've added this test I can do a more thorough review of your changes.
> 
> In addition - it looks like `KFP SDK Docformatter Test`, `KFP SDK Isort Test` and `KFP SDK YAPF Tests` are failing - can you please fix those?

Yeah sure, I'll commit changes immediately

shreeharshshinde:
@alyssacgoins I did make changes but I am not aware of these 62 files changed, I just ran scripts for docformatter, isort, yapf and add tests as per your suggestion

alyssacgoins:
> @alyssacgoins I did make changes but I am not aware of these 62 files changed, I just ran scripts for docformatter, isort, yapf and add tests as per your suggestion

@shreeharshshinde can you update your PR to remove the files that were unintentionally changed?

shreeharshshinde:
Hi @alyssacgoins, thanks for the review and patience.

I’ve updated the PR to clean up the changes and limit the scope strictly to the intended SDK compiler + test updates for Literal support. The goal was to remove any unrelated or accidental changes that made the earlier diff noisy.

Regarding the current status:
- The `ok-to-test` and workflow approvals are expected since I don’t have maintainer permissions.
- DCO is currently failing — I’m happy to amend/rebase commits again if you’d like it handled a specific way.
- CI failures appear to be related to workflow approval rather than code issues on my side.

At this point, could you please let me know:
- If the current diff scope looks correct now, and
- Whether you’d prefer me to further rebase/amend anything before re-running CI?

test_data/sdk_compiled_pipelines/valid/pipeline_with_literal_inputs.py
```diff
@@ -0,0 +1,26 @@
+"""Pipeline used to validate successful compilation of Literal input parameters.
```
alyssacgoins, line 1: Can you rename this file and its corresponding yaml file to `pipeline_with_literal_inputs`? This will provide some more context for the term `literal` here

test_data/sdk_compiled_pipelines/valid/pipeline_with_literal_inputs.py
```diff
@@ -0,0 +1,26 @@
...
+This test ensures that pipelines using typing.Literal inputs compile correctly
+and produce stable IR output.
+"""
+
+import typing
+from kfp import dsl
+
+
+@dsl.component
+def literal_component(x: typing.Literal['a', 'b']) -> str:
```
alyssacgoins: It would be helpful to have a more descriptive parameter than `x` here - `literal_param` or `input` are some options. You can go with another option too - just make it a bit longer.

test_data/sdk_compiled_pipelines/valid/pipeline_with_literal_inputs.py
```diff
@@ -0,0 +1,26 @@
...
+
+This test ensures that pipelines using typing.Literal inputs compile correctly
+and produce stable IR output.
+"""
+
+import typing
+from kfp import dsl
+
+
+@dsl.component
```
alyssacgoins, line 11: Instead of returning the input - one commonly used component pattern across a lot of these test pipelines is `print_op()` - prints the input:
```suggestion
@dsl.component
def print_op(input: typing.Literal['a', 'b']):
    """Prints a message."""
    print(input)
```

sdk/python/kfp/dsl/structures.py
```diff
@@ -45,12 +45,19 @@ class InputSpec:
         type: The type of the input.
         default (optional): the default value for the input.
         optional: Wether the input is optional. An input is optional when it has an explicit default value.
+literals (optional): list of allowed input values.
```
alyssacgoins: should be same indentation level as the rest of the `Attributes` block:
```suggestion
        literals (optional): list of allowed input values.
```

sdk/python/kfp/dsl/structures.py
```diff
@@ -1016,6 +1023,7 @@ def get_inputs(task_group: tasks_group.TasksGroup,
 
 def normalize_time_string(duration: str) -> str:
     """Normalizes a time string.
+
```
alyssacgoins: Can you revert this?

sdk/python/kfp/compiler/compiler_test.py
```diff
@@ -187,6 +188,32 @@ def print_artifact(a: Input[Artifact]):
 
 class TestCompilePipeline(parameterized.TestCase):
 
+    def test_compile_pipeline_with_literal_single_type_valid(self):
```
alyssacgoins:
```suggestion
    def test_compile_pipeline_with_literal_input_valid(self):
```
This will clarify to the user that this pipeline utilizes literal inputs. Since literal inputs are only valid when a single type is present, adding `single-type` isn't necessary here.

sdk/python/kfp/compiler/compiler_test.py
```diff
@@ -187,6 +188,32 @@ def print_artifact(a: Input[Artifact]):
...
+
+        @dsl.component
```
alyssacgoins, line 192: Can you apply the comments I made on `test_data/sdk_compiled_pipelines/valid/literal_pipeline.py` (function and variable naming) to this sample pipeline as well? For clarity, as well as to fit in with common formatting practices. You can call `print_op()` from line 104, which is a common function utilized throughout the test.

sdk/python/kfp/dsl/types/type_utils.py
```diff
@@ -618,6 +636,60 @@ def _annotation_to_type_struct(annotation):
     return type_struct or schema_title
 
 
+def annotation_to_type_struct_and_literals(annotation):
```
alyssacgoins: You don't need to create a separate function with a different name here - you can just update `annotation_to_type_struct()` with the additional logic you've added here to process literal inputs.

sdk/python/kfp/compiler/compiler_test.py
```diff
@@ -187,6 +188,32 @@ def print_artifact(a: Input[Artifact]):
...
+                package_path=target_file,
+            )
+            self.assertTrue(os.path.exists(target_file))
+
+    def test_literal_mixed_elements_should_raise_type_error(self):
+        # Mixed-type Literal should be rejected at component definition time
+        with self.assertRaises(TypeError):
+
+            @dsl.component
+            def bad_literal(x: typing.Literal[1, 2, 'a']):
```
alyssacgoins: Same as above - can you update the name from `bad_literal` to match the formatting of other test cases?

sdk/python/kfp/compiler/compiler_test.py
```diff
@@ -187,6 +188,32 @@ def print_artifact(a: Input[Artifact]):
...
+
+        with tempfile.TemporaryDirectory() as tmpdir:
+            target_file = os.path.join(tmpdir, 'result.yaml')
+            compiler.Compiler().compile(
+                pipeline_func=literal_pipeline,
+                package_path=target_file,
+            )
+            self.assertTrue(os.path.exists(target_file))
+
+    def test_literal_mixed_elements_should_raise_type_error(self):
```
alyssacgoins:
```suggestion
    def test_compile_pipeline_with_literal_input_invalid(self):
```
This or something similar would be a clearer way to describe the test

alyssacgoins (changes requested):

alyssacgoins:
> Hi @alyssacgoins, thanks for the review and patience.
> 
> I’ve updated the PR to clean up the changes and limit the scope strictly to the intended SDK compiler + test updates for Literal support. The goal was to remove any unrelated or accidental changes that made the earlier diff noisy.
> 
> Regarding the current status:
> 
> * The `ok-to-test` and workflow approvals are expected since I don’t have maintainer permissions.
> * DCO is currently failing — I’m happy to amend/rebase commits again if you’d like it handled a specific way.
> * CI failures appear to be related to workflow approval rather than code issues on my side.
> 
> At this point, could you please let me know:
> 
> * If the current diff scope looks correct now, and
> * Whether you’d prefer me to further rebase/amend anything before re-running CI?

@shreeharshshinde thanks for your updates - a couple things before we run CI again: 
- can you rebase your branch so that it's up-to-date with the base branch?
- DCO is failing because one of your commits is incorrectly signed off - you can correct this with `git commit --amend -s -m "<your commit message>" while rebasing.`
- I've added some more review comments as well

sdk/python/kfp/compiler/compiler_test.py
shreeharshshinde: Done

test_data/sdk_compiled_pipelines/valid/pipeline_with_literal_inputs.py
shreeharshshinde: Done, I will keep in mind about naming onwards

test_data/sdk_compiled_pipelines/valid/pipeline_with_literal_inputs.py
shreeharshshinde: I replaced with 'literal_param' which makes sense to me

shreeharshshinde:
@alyssacgoins Done with some modifications, Thank you very much for your guidance !!!

sdk/python/kfp/compiler/compiler_test.py
```diff
@@ -187,6 +188,34 @@ def print_artifact(a: Input[Artifact]):
 
 class TestCompilePipeline(parameterized.TestCase):
 
+    def test_compile_pipeline_with_literal_input_valid(self):
+
+        @dsl.component
+        def literal_component(literal_param: typing.Literal['a', 'b']):
```
alyssacgoins: Can you reference [ `print_op()`](https://github.com/kubeflow/pipelines/blob/master/sdk/python/kfp/compiler/compiler_test.py#L104-L106) here instead of declaring your own print function? `print_op()` is used throughout this test, so it would be best to follow that practice

sdk/python/kfp/compiler/compiler_test.py
```diff
@@ -187,6 +188,34 @@ def print_artifact(a: Input[Artifact]):
...
+            self.assertTrue(os.path.exists(target_file))
+
+
+    def test_compile_pipeline_with_literal_input_mixed_types_invalid(self):
+        # Mixed-type Literal should be rejected at component definition time
+        with self.assertRaises(TypeError):
+
+            @dsl.component
+            def literal_component_with_mixed_types(literal_param: typing.Literal[1, 2, 'a']):
+                print(literal_param)
```
alyssacgoins: Same as line 194 - can you reference [`print_op()`](https://github.com/kubeflow/pipelines/blob/master/sdk/python/kfp/compiler/compiler_test.py#L104-L106) here instead of declaring your own print function? print_op() is used throughout this test, so it would be best to follow that practice

alyssacgoins:
@shreeharshshinde your changes look good - I've added one small suggestion on two lines, but other than that - can you update the branch and also fix the failing tests?  It looks like `compiler_test.test_compile_pipeline_with_literal_input_valid() `is failing, as well as some compiler tests. I would update your branch before you take a look at the compiler tests.

sdk/python/kfp/compiler/compiler_test.py
shreeharshshinde: Done

sdk/python/kfp/compiler/compiler_test.py
shreeharshshinde: Done

shreeharshshinde:
@alyssacgoins I resolved workflow compiler test as it had easy solution cause yaml file was missing from compiled-workflows while I am not able to understand other than the formatting tests why sdk and unit sdk tests are failing

alyssacgoins:
> @alyssacgoins I resolved workflow compiler test as it had easy solution cause yaml file was missing from compiled-workflows while I am not able to understand other than the formatting tests why sdk and unit sdk tests are failing

@shreeharshshinde The SDK workflows here are failing for a couple reasons:

- Your test case `test_compile_pipeline_with_literal_input_valid` in `sdk/python/kfp/compiler/compiler_test.py` is failing - this is visible in the workflow logs. 
- `KFP SDK YAPF Tests` fails due to the formatting of your changes to `sdk/python/kfp/dsl/types/type_utils.py` - the workflow logs above provide the fixed formatting.
- `KFP SDK Isort Test` fails because of the formatting of the import you added to `pipelines/sdk/python/test/compilation/pipeline_compilation_test.py` - check the workflow logs for more info (you can find the workflow logs by clicking the failed workflow above). 

In addition - make sure you update your branch to be up-to-date with the base branch.

shreeharshshinde:
@alyssacgoins finally, all checks are passed

sdk/python/kfp/dsl/types/type_utils.py
```diff
@@ -586,6 +604,57 @@ def _annotation_to_type_struct(annotation):
...
+
+        if not is_literal:
+            try:
+                import typing
+                if hasattr(typing, 'Literal') and origin is typing.Literal:
+                    is_literal = True
+            except ImportError:
+                pass
+
+        if not is_literal:
```
alyssacgoins: It looks like you repeat this block `if not is_literal`: - can you fix this?

sdk/python/kfp/dsl/component_factory.py
```diff
@@ -385,17 +391,30 @@ def make_input_output_spec_args(annotation: Any) -> Dict[str, Any]:
...
+        'type': typ,
+        'is_artifact_list': is_artifact_list,
+        'literals': literals,
+    }
 
 
 def make_output_spec(annotation: Any) -> structures.OutputSpec:
     annotation = canonicalize_annotation(annotation)
     args = make_input_output_spec_args(annotation)
+    args.pop('literals', None)
```
alyssacgoins, line 418: why are you removing `literals` here?

sdk/python/test/compilation/pipeline_compilation_test.py
```diff
@@ -1,3 +1,4 @@
+# isort: skip_file
```
alyssacgoins: Can you remove this line?

alyssacgoins (changes requested):

sdk/python/kfp/dsl/component_factory.py
shreeharshshinde: `make_input_output_spec_args()` is shared for both inputs and outputs and returns `{type, is_artifact_list, literals}`.
However, OutputSpec in KFP does not support a literals field, while InputSpec does.
So `pop('literals', None)` removes it before constructing OutputSpec to avoid passing an unsupported argument.

I think this keeps the shared helper while ensuring the output spec stays compatible with the IR schema.

sdk/python/kfp/compiler/pipeline_spec_builder.py
```diff
@@ -417,6 +417,10 @@ def _build_component_spec_from_component_spec_structure(
             if input_spec.description:
                 component_spec.input_definitions.parameters[
                     input_name].description = input_spec.description
+            if input_spec.literals:
+                for lit_val in input_spec.literals:
```
alyssacgoins: Can you update `lit_val` to a clearer term? Perhaps `literal_val`, `val` or `value`.

sdk/python/kfp/dsl/component_factory.py
```diff
@@ -311,10 +311,9 @@ def get_name_to_specs(
 
         # parameter type
         else:
-            type_string = type_utils._annotation_to_type_struct(annotation)
             name_to_input_specs[maybe_make_unique(
                 name, list(name_to_input_specs))] = make_input_spec(
-                    type_string, func_param)
+                    annotation, func_param)
```
alyssacgoins, line 316: This change looks potentially extraneous - is this directly necessary for supporting literals, or can you change it back?

sdk/python/kfp/dsl/component_factory.py
```diff
@@ -385,17 +392,30 @@ def make_input_output_spec_args(annotation: Any) -> Dict[str, Any]:
     if is_artifact_list:
         annotation = type_annotations.get_inner_type(annotation)
 
+    literals = None
+
```
alyssacgoins, line 396: can remove this extra line

alyssacgoins (changes requested):

sdk/python/kfp/compiler/pipeline_spec_builder.py
shreeharshshinde: Done to `literal_value`

sdk/python/kfp/dsl/component_factory.py
shreeharshshinde: this is actually required for literal support. `make_input_spec()` needs the original annotation object for type introspection (e.g. detecting `Literal`, distinguishing artifacts vs primitives, etc.) passing `type_string` breaks that since it loses type information, which ends up affecting all input type resolution not just literals so I thought this change is intentional and necessary

sdk/python/kfp/dsl/component_factory.py
shreeharshshinde: Simillarly needed to avoid a **NameError** in non-Literal cases `_annotation_to_type_struct()` only sets literals when a Literal is detected, so without this initialization, most standard parameters would hit an undefined variable so this ensures safe handling for both literal and non-literal types

alyssacgoins:
@zazulam @droctothorpe I've been working with @shreeharshshinde on this feature and everything looks good to me. Recalling that you reviewed the API server/driver component for this work (#12607), would you mind taking a look here?
Here is the original [KEP for Literal Inputs](https://github.com/kubeflow/pipelines/blob/master/proposals/11385-literal-input-parameters/README.md), for reference.
Looking forward to getting this merged!

droctothorpe (commented):
Thanks for adding Literal support. I found a few correctness gaps that should be addressed before this lands:

1. Literal defaults are not validated (`sdk/python/kfp/dsl/component_factory.py`)

A component can declare a Literal constraint with a default that is outside the allowed set, and it still compiles:

```python
@dsl.component
def literal_component(x: Literal['a', 'b'] = 'c'):
    ...

@dsl.pipeline
def p():
    literal_component()
```

This produces an internally inconsistent component contract: `defaultValue` is `c`, while `literals` is `['a', 'b']`.

- Why this matters: callers can omit the argument and get a value that violates the component's own Literal constraint.
- Suggested fix: when building the `InputSpec`, validate any explicit default against `literals` and reject defaults outside the allowed set.

2. Literal constraints are skipped for pipeline-channel arguments (`sdk/python/kfp/dsl/types/type_utils.py`)

This compiles even though the root input and component input have disjoint Literal sets:

```python
@dsl.component
def consume(x: Literal['a', 'b']):
    ...

@dsl.pipeline
def p(x: Literal['c']):
    consume(x=x)
```

The compiled IR wires the root parameter into the component, but the root parameter can only be `c` while the component only accepts `a` or `b`.

- Why this matters: the compiler can emit a pipeline where a parameter can never satisfy the downstream component contract.
- Suggested fix: preserve Literal metadata on parameter channels, or otherwise validate source and destination Literal sets when both sides are known.

3. Warn-only compatibility mode returns `True` for Literal mismatches (`sdk/python/kfp/dsl/types/type_utils.py`)

`verify_type_compatibility(..., raise_on_error=False)` warns for an invalid Literal value but still returns `True`.

- Why this matters: the function contract says it returns `False` when values are incompatible, and callers using the boolean result will treat invalid Literal values as compatible.
- Suggested fix: mark the compatibility result as `False` before warning/returning for Literal mismatches.

4. `@dsl.container_component` still rejects Literal inputs (`sdk/python/kfp/dsl/component_factory.py`)

A container component with a Literal input fails at definition time:

```python
@dsl.container_component
def cc(x: Literal['a', 'b']):
    return dsl.ContainerSpec(image='alpine', command=['echo', x])
```

The placeholder path passes the raw `Literal[...]` annotation to `get_parameter_type_name()`, which raises because it is not normalized first.

- Why this matters: the PR adds Literal support through shared component interface extraction, but container components still cannot use it.
- Suggested fix: normalize the annotation before setting `placeholder._ir_type`, or derive the placeholder IR type from `component_spec.inputs[parameter.name].type`.

shreeharshshinde:
The bulk of the `compiler_test.py` diff is docformatter reformatting triggered by running the required formatters on all modified files. No test logic was changed. 

The only new test in that file is `test_compile_pipeline_with_disjoint_literal_channel_raises`.

shreeharshshinde:
> Thanks for adding Literal support. I found a few correctness gaps that should be addressed before this lands:
> 
> 1. Literal defaults are not validated (`sdk/python/kfp/dsl/component_factory.py`)
> 
> A component can declare a Literal constraint with a default that is outside the allowed set, and it still compiles:
> 
> ```python
> @dsl.component
> def literal_component(x: Literal['a', 'b'] = 'c'):
>     ...
> 
> @dsl.pipeline
> def p():
>     literal_component()
> ```
> 
> This produces an internally inconsistent component contract: `defaultValue` is `c`, while `literals` is `['a', 'b']`.
> 
> * Why this matters: callers can omit the argument and get a value that violates the component's own Literal constraint.
> * Suggested fix: when building the `InputSpec`, validate any explicit default against `literals` and reject defaults outside the allowed set.
> 
> 2. Literal constraints are skipped for pipeline-channel arguments (`sdk/python/kfp/dsl/types/type_utils.py`)
> 
> This compiles even though the root input and component input have disjoint Literal sets:
> 
> ```python
> @dsl.component
> def consume(x: Literal['a', 'b']):
>     ...
> 
> @dsl.pipeline
> def p(x: Literal['c']):
>     consume(x=x)
> ```
> 
> The compiled IR wires the root parameter into the component, but the root parameter can only be `c` while the component only accepts `a` or `b`.
> 
> * Why this matters: the compiler can emit a pipeline where a parameter can never satisfy the downstream component contract.
> * Suggested fix: preserve Literal metadata on parameter channels, or otherwise validate source and destination Literal sets when both sides are known.
> 
> 3. Warn-only compatibility mode returns `True` for Literal mismatches (`sdk/python/kfp/dsl/types/type_utils.py`)
> 
> `verify_type_compatibility(..., raise_on_error=False)` warns for an invalid Literal value but still returns `True`.
> 
> * Why this matters: the function contract says it returns `False` when values are incompatible, and callers using the boolean result will treat invalid Literal values as compatible.
> * Suggested fix: mark the compatibility result as `False` before warning/returning for Literal mismatches.
> 
> 4. `@dsl.container_component` still rejects Literal inputs (`sdk/python/kfp/dsl/component_factory.py`)
> 
> A container component with a Literal input fails at definition time:
> 
> ```python
> @dsl.container_component
> def cc(x: Literal['a', 'b']):
>     return dsl.ContainerSpec(image='alpine', command=['echo', x])
> ```
> 
> The placeholder path passes the raw `Literal[...]` annotation to `get_parameter_type_name()`, which raises because it is not normalized first.
> 
> * Why this matters: the PR adds Literal support through shared component interface extraction, but container components still cannot use it.
> * Suggested fix: normalize the annotation before setting `placeholder._ir_type`, or derive the placeholder IR type from `component_spec.inputs[parameter.name].type`.

Thanks for the thorough review @droctothorpe ! I've addressed all four points:

1. **Literal defaults are now validated** when building the `InputSpec`. 
Components with defaults outside the allowed `Literal` set now raise a `ValueError` instead of producing an inconsistent spec.

```python
literals = input_output_spec_args.get('literals')
if literals is not None and default is not None and default not in literals:
    raise ValueError(...)
```

2. **`verify_type_compatibility(..., raise_on_error=False)` now returns `False`** for `Literal` mismatches. 
The compatibility flag is updated before emitting the warning so the return value matches the documented behavior.

3. **Literal metadata is now preserved on pipeline parameter channels.** 
I added optional `literals` metadata to `PipelineParameterChannel`, threaded it through pipeline input channel creation, and updated `verify_type_compatibility` to validate source/destination `Literal` sets when both are known (using a subset check). Channels without `Literal` metadata (e.g. task outputs) continue to be left unchecked.

4. **`@dsl.container_component` now accepts `Literal` inputs.** 
The placeholder path now normalizes `Literal[...]` annotations before calling `get_parameter_type_name()`.

```python
result = type_utils._annotation_to_type_struct(annotation)
type_struct = result[0] if isinstance(result, tuple) else result
placeholder._ir_type = type_utils.get_parameter_type_name(type_struct)
```

I've also added regression tests covering each of these fixes, including an end-to-end compiler test for incompatible `Literal` pipeline channels.

Please let me know if there's anything else you'd like me to update.

prakharsingh-74:
@mprahl, @droctothorpe  and @alyssacgoins, Since @shreeharshshinde is not active for a long time can i continue the work from here if you guys allow me to do that?

shreeharshshinde:
@prakharsingh-74 I believe work is already done, just some E2E tests are failing which are not on my side so, It's not the case I am not been active, since maintainers have thier own workload so it will take time to get merged 

Thanks !!!

droctothorpe:
Hi @shreeharshshinde, thank you for your PR! The linked issue (#12601) must be labeled `ready` before your contribution can be reviewed. Please check our [CONTRIBUTING.md](https://github.com/kubeflow/pipelines/blob/master/CONTRIBUTING.md) for details.

PR #12608, opened 2026-01-05
feat(backend): Add HuggingFace Hub artifact import support. Fixes #12501

### LINKED ISSUE #12501: [feature] Allow models/datasets import option to dsl.importer
### Feature Area
/area sdk 

I would like to use dsl.importer to allow imports from hugging face that may or may not be behind authentication (just to keep things consistent)

Pain point is consistency, we have dsl.importer for oci, s3, gcp but not for other storages like hugging face.

Direct call to hugging face api with tokens defined in the pipeline code as enviornment variable (via kubesecret)

---

Love this idea? Give it a 👍.

Siva-Sainath:
Addresses the issue #12501, 

Implements secure HuggingFace model artifact import via the standard DSL importer decorator pattern (artifact_uri='huggingface://repo-id'), consistent with S3/GCS/OCI implementations.

Extracts HuggingFace authentication token from Kubernetes Secret metadata (through HuggingFaceParams in pipeline spec), injects it at runtime as HF_TOKEN environment variable in pod context and downloads models using func huggingface_hub.snapshot_download() to location /huggingface/<repo>/<revision>.

SDK Layer (artifact_types.py): Registers huggingface:// URI scheme for importer decorator

Backend Config (config.go): HuggingFaceParams struct with func StructuredHuggingFaceParams() for parsing secret metadata

Token Injection (object_store.go): getHuggingFaceToken() retrieves token from Secret and sets HF_TOKEN env variable

Import Workflow (importer_launcher.go): handleHuggingFaceImport() takes care of artifact download with injected credentials

Token remains in Kubernetes Secret and is injected at pod execution time—never embedded in pipeline code.

**Checklist:**

### DIFF
--- a/backend/Dockerfile.launcher
+++ b/backend/Dockerfile.launcher
@@ -29,7 +29,11 @@ RUN GO111MODULE=on CGO_ENABLED=0 GOOS=linux go build -tags netgo -ldflags '-extl
 
 FROM alpine:3.19
 
-RUN adduser -S -u 65532 appuser
+# Install Python and pip for HuggingFace support
+RUN apk add --no-cache python3 py3-pip && \
+    pip3 install --no-cache-dir --break-system-packages huggingface-hub==1.2.4 && \
+    adduser -S -u 65532 appuser
+
 USER 65532
 
 WORKDIR /bin
--- a/backend/src/v2/component/importer_launcher.go
+++ b/backend/src/v2/component/importer_launcher.go
@@ -4,6 +4,9 @@ import (
 	"context"
 	"encoding/json"
 	"fmt"
+	"net/url"
+	"os"
+	"os/exec"
 	"strings"
 
 	"github.com/kubeflow/pipelines/backend/src/common/util"
@@ -274,6 +277,14 @@ func (l *ImportLauncher) ImportSpecToMLMDArtifact(ctx context.Context) (artifact
 		return artifact, nil
 	}
 
+	if strings.HasPrefix(artifactUri, "huggingface://") {
+		// HuggingFace artifacts
+		if err := l.handleHuggingFaceImport(ctx, artifactUri, artifact); err != nil {
+			return nil, err
+		}
+		return artifact, nil
+	}
+
 	provider, err := objectstore.ParseProviderFromPath(artifactUri)
 	if err != nil {
 		return nil, fmt.Errorf("no provider scheme found in artifact URI: %s", artifactUri)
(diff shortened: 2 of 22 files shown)


backend/src/v2/component/importer_launcher.go
```diff
@@ -320,6 +330,97 @@ func (l *ImportLauncher) ImportSpecToMLMDArtifact(ctx context.Context) (artifact
 	return artifact, nil
 }
 
+func (l *ImportLauncher) handleHuggingFaceImport(ctx context.Context, artifactURI string, artifact *pb.Artifact) error {
+	parts := strings.TrimPrefix(artifactURI, "huggingface://")
+	pathParts := strings.Split(parts, "/")
+
+	if len(pathParts) < 2 {
+		return fmt.Errorf("invalid HuggingFace URI format: %q, expected huggingface://repo_id/revision", artifactURI)
```
nsingla: what if I want to download a specific file, can I not specify `huggingface://repo_id/file_name` or if I want to download data set, then repo_type is dataset in that case

Can we add support for that?

Also, not sure if `huggingface://` is a standard protocol prefix i.e not sure if its obvious to users that they need to specify uri with this prefix

backend/src/v2/component/importer_launcher.go
```diff
@@ -320,6 +330,97 @@ func (l *ImportLauncher) ImportSpecToMLMDArtifact(ctx context.Context) (artifact
...
+	if l.importer.GetDownloadToWorkspace() {
+		if err := l.downloadHuggingFaceModel(ctx, repoID, revision, artifactURI); err != nil {
+			return fmt.Errorf("failed to download HuggingFace model: %w", err)
+		}
+	}
+
+	return nil
+}
+
+func (l *ImportLauncher) downloadHuggingFaceModel(ctx context.Context, repoID, revision, artifactURI string) error {
```
nsingla: what if I have defined Workspace config with a PVC, will it download to that PVC or will it first upload to S3 and download to PVC from s3?

backend/src/v2/component/importer_launcher.go
```diff
@@ -320,6 +330,97 @@ func (l *ImportLauncher) ImportSpecToMLMDArtifact(ctx context.Context) (artifact
...
+		hfToken, _ = objectstore.GetHuggingFaceTokenFromSessionInfo(ctx, l.k8sClient, l.launcherV2Options.Namespace, bucketConfig.SessionInfo)
+	}
+
+	glog.Infof("Downloading HuggingFace model %q (revision=%q) to workspace path %q", repoID, revision, localPath)
+
+	if err := os.MkdirAll(localPath, 0755); err != nil {
+		return fmt.Errorf("failed to create directory %q: %w", localPath, err)
+	}
+
+	cmd := exec.CommandContext(ctx, "python3", "-c", fmt.Sprintf(`
```
nsingla: where is `huggingface_hub` sdk getting pip installed, are we expecting that to come as part of `packages_to_install`. If so, then in the future there may be version conflicts and method signature in those versions could be different, hence `snapshot_download` can fail

nsingla (changes requested):
I think we need some tests to verify that this works:
1. Python Pipeline code
2. Compiled version
3. Argo Compiled Version
4. Local Test Runner showing this works
5. Add this pipeline to e2e tests

backend/src/v2/component/importer_launcher.go
Siva-Sainath: No S3 intermediary, models download directly to the PVC when workspace is configured with it

backend/src/v2/component/importer_launcher.go
Siva-Sainath: The huggingface-hub library is installed in the launcher container image at build time, pinned to version 1.2.4, present in the Dockerfile.launcher file. Tried to ensure version compatibility through dynamic parameter detection using inspect.signature(), which adapts the code to work with any future version of huggingface_hub without requiring code changes

backend/src/v2/component/importer_launcher.go
Siva-Sainath: In the latest commit, added support for specific datasets or files using huggingface://repo_id/file_name this 
or a pattern huggingface://repo_id?allow_patterns=*.safetensors

To let users know that they have to use the URI huggingface://added a error message that details:
return fmt.Errorf("Invalid HuggingFace URI format: %q, expected huggingface://repo_id[/revision]\n"+
    "Examples:\n"+
    "  huggingface://gpt2\n"+
    "  huggingface://meta-llama/Llama-2-7b\n"+
    "  huggingface://bert-base-uncased/config.json\n"+
    "  huggingface://wikitext?repo_type=dataset",
    parts[0])

huggingface:// is NOT a standard protocol prefix it was written to match  gs://, s3:// URI scheme that was present in the code

Siva-Sainath:
@nsingla will work on writing the tests. I would like to clarify, by adding this pipeline to e2e tests, do you mean adding the HF artifact import tests to the existing e2e test runner, or is there a specific test suite/location where I would integrate this? should I commit any of these tests to the PR or simply run and verify if they work or not?

backend/Dockerfile.launcher
```diff
@@ -27,9 +27,11 @@ COPY . .
 
 RUN GO111MODULE=on CGO_ENABLED=0 GOOS=linux go build -tags netgo -ldflags '-extldflags "-static"' -o /bin/launcher-v2 ./backend/src/v2/cmd/launcher-v2/*.go
 
-FROM alpine:3.19
+FROM python:3.12-slim
```
nsingla: I agree with copilot
Also 3.12 is not the officially supported python version even though it works. And we still need to support python 3.9 for backwards compatibility

nsingla:
> @nsingla will work on writing the tests. I would like to clarify, by adding this pipeline to e2e tests, do you mean adding the HF artifact import tests to the existing e2e test runner, or is there a specific test suite/location where I would integrate this? should I commit any of these tests to the PR or simply run and verify if they work or not?

To add to the e2e suite, you need to:
1. Create a pipeline (in your case a pipeline that downloads from hugging face) - both python & compiled yaml under "test_data/sdk_compiled_pipelines/valid/essential"
2. cd to `backend/test/compiler`, and run `ginkgo -- -createGoldenFiles=True`

These steps do 2 things:
1. Add your new pipeline to the e2e suite automatically as the e2e suite is watching this directory
2. Add a argo compiler test, which compiles the yaml file and compare it against a argo compiled workflow (which you generated in step 2)

Hope this helps

PR #12610, opened 2026-01-05
Recurring Runs Queue Throughput Optimization

ntny:

Optimized the execution of recurring runs.
Previously, in the ScheduledWorkflow controller, every event-driven/full reconciliation unconditionally updated the corresponding ScheduledWorkflow. 
- Each update triggered another reconciliation of the same resource, which led to workqueue flooding when thousands of ScheduledWorkflows were present in the cluster.
- Additionally, mass updates of ScheduledWorkflows became a bottleneck due to excessive Kubernetes API server requests: update calls had 0.2–0.5s p95 latency under load. This bottleneck was further amplified by client-side QPS/Burst limits.

The fix [introduces](https://github.com/kubeflow/pipelines/pull/12610/files#diff-d773c400e6cc4d6bab184a77e5bc9a5801030a787a2c5704c4e6400f4a4355b3R657) a pre-update state comparison and skips no-op updates to the Kubernetes API, eliminating redundant writes.

According to the informer configuration, a full resync still runs every 30 seconds. During this resync, all ScheduledWorkflows are iterated to check whether it is time to trigger a new run.
This change does not alter or remove that behavior. Instead, it just eliminates unnecessary updates reconciliation, so the periodic full resync remains the single place where time-based scheduling logic is applied.

Additionally
-  added metrics to scheduledworkflow controller 
- made QPS/burst configurable parameters for the controller.

@juliusvonkohout

### DIFF
--- a/backend/Dockerfile.scheduledworkflow
+++ b/backend/Dockerfile.scheduledworkflow
@@ -44,4 +44,4 @@ COPY --from=builder /bin/controller /bin/controller
 ENV NAMESPACE=""
 ENV LOG_LEVEL=info
 
-CMD ["/bin/sh", "-c", "/bin/controller --logtostderr=true --namespace=${NAMESPACE} --logLevel=${LOG_LEVEL} --clientQPS=${CLIENT_QPC:-5} --clientBurst=${CLIENT_BURST:-10} --metricsPort=${METRICS_PORT:-9090}"]
+CMD ["/bin/sh", "-c", "/bin/controller --logtostderr=true --namespace=${NAMESPACE} --logLevel=${LOG_LEVEL} --clientQPS=${CLIENT_QPS:-5} --clientBurst=${CLIENT_BURST:-10} --recurringRunResyncIntervalSeconds=${RESYNC_INTERVAL_SECONDS:-30} --metricsPort=${METRICS_PORT:-9090}"]
--- a/backend/src/crd/controller/scheduledworkflow/README.md
+++ b/backend/src/crd/controller/scheduledworkflow/README.md
@@ -1,0 +1,27 @@
+## Controller Flags
+
+The controller supports the following command-line flags:
+
+| Flag                                  | Description | Default |
+|---------------------------------------|-------------|---------|
+| `--logtostderr`                       | Log output to stderr instead of files. | `true` |
+| `--logLevel`                          | Defines the log level for the application. | `""` |
+| `--clientQPS`                         | Maximum queries per second to the Kubernetes API server. | `5` |
+| `--clientBurst`                       | Maximum burst for throttle from this client. | `10` |
+| `--recurringRunResyncIntervalSeconds` | Full resync interval in seconds for recurring run reconciliations. | `30` |
+| `--metricsPort`                       | The port for the metrics endpoint. | `9090` |
+
+## Prometheus Metrics
+
+The controller exposes the following Prometheus metrics:
+
+| Metric | Type | Description | Labels |
+|--------|------|-------------|--------|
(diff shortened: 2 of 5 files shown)


juliusvonkohout:
Thank you
/ok-to-test

juliusvonkohout:
@hbelmiro @HumairAK @droctothorpe that would be a very good scalability improvement for large clusters. May you take a look?

@ntny regarding "According to the informer configuration, a full resync still runs every 30 seconds. During this resync, all ScheduledWorkflows are iterated to check whether it is time to trigger a new run." Is there a particular reason to not run the resnynchronization every 60 seconds? That should also help a lot to decrease the synchronization effort and Kubernetes API server requests.

droctothorpe:
1. @aniketpati1121 it looks like the first time contributor workflow might be getting false positives. @ntny has multiple PRs prior to this one.

2. This is fantastic, @ntny. Can I trouble you to add some explanatory comments to the core logic itself? Thanks! 

/ok-to-test

ntny:
@droctothorpe, sure! Will add inline comments to the code later, but here's the logic of the change:
Before this MR, the recurring run reconciliation flow worked as follows:
There are Kubernetes CRs called _ScheduledWorkflow_. They contain the cron schedule, enable/disable state, and run history.

- The _Scheduled Workflow Controller_ runs a full reconciliation every 30 seconds. For each _ScheduledWorkflow_, it checks whether a new Run should be created at this moment.

- After that, the controller unconditionally updates the corresponding _ScheduledWorkflow_ via the Kubernetes API.

In most cases, the ScheduledWorkflow object is not actually changed, and the update just rewrites the same data. This causes two problems:

1. Due to the QPS limits (made configurable in this PR), the number of updates per second is capped (e.g. at 10), which becomes a significant bottleneck.

2. Each update triggers an event and a driver reconciliation for the ScheduledWorkflow, even when nothing has changed. As a result, the reconciliation flow becomes flooded and congested.

**Main change:**
- This MR [introduces](https://github.com/kubeflow/pipelines/pull/12610/files#diff-d773c400e6cc4d6bab184a77e5bc9a5801030a787a2c5704c4e6400f4a4355b3R657) a pre-update check. The controller updates the ScheduledWorkflow object only if it has actually changed (most commonly, the run history).

**Additional changes:**
- Introduce Prometheus metrics for the Scheduled Workflow Controller
- Make QPS and burst parameters configurable via Kubernetes manifests

@juliusvonkohout Hi,I don’t think there is a strict reason for choosing exactly 30 seconds. It looks like a reasonable default that balances responsiveness and load, and it has been working well so far.
I’m planning to make this interval configurable via manifests later today. This is likely a case-by-case choice, and with the new metrics in place it should be easier to tune it more precisely.

ntny:
Last update:
- _fullResyncInterval_ is now a configurable parameter
- Added explanatory commentary and a README describing all command-line arguments and exposed metrics of the scheduled workflow controller

aniketpati1121:
> 1. @aniketpati1121 it looks like the first time contributor workflow might be getting false positives. @ntny has multiple PRs prior to this one.
> 2. This is fantastic, @ntny. Can I trouble you to add some explanatory comments to the core logic itself? Thanks!
> 
> /ok-to-test

Hi @droctothorpe ,

I noticed the first-time contributor workflow sometimes gives false positives, but I checked the overall workflow and tested by creating a new PR. Currently, there doesn’t seem to be any issue with the first-time contributor check.

Just wanted to confirm — everything should be fine from my side.

Thanks!

droctothorpe:
Apologies for the delay on this. This is a phenomenal contribution. Thank you so much, @ntny! 🎉 

/lgtm
/approve

PR #12639, opened 2026-01-12
fix(manifests): remove unused kustomize vars causing build warnings

### LINKED ISSUE #12615: [bug] Kustomize build is displaying warnings for unsubstituted vars
### Environment
MacOS running kind

As part of https://github.com/kubeflow/manifests/issues/2991 I am trying to fix up the kustomize warnings and come across this warning 

Image

The warning occurs in `manifests/kustomize/base/installs/generic` directory while running ```kustomize build  .```

### Steps to reproduce

- navigate to `manifests/kustomize/base/installs/generic` dir
- execute `kustomize build .`

### Expected result

The warning shouldnt have occurred in first place as all vars should have been substituted.
If I were to remove these vars with replacements, is it safe for me to ignore these vars and their replacements?
PS: unit tests in the CI peipline check for these vars so they cannot be safely ignored unless CI tests are updated. But the broader question is if its safe to remove these vars alltogather

Please note, warnings are not a problem, the ask is to 
- confirm if the vars are no longer requred
-  update the check for param.yaml so that it can be safely removed

### Materials and reference

see screenshot attached 

### Labels

/area manifests

---

shreeharshshinde:
## Walkthrough - Fix Kustomize Warnings
I have removed the unused variables from the Kustomize manifests to resolve the "well-defined vars that were never replaced" warnings.

### Changes
generic
[MODIFY]  `kustomization.yaml`
Removed the following unused vars:

```
kfp-namespace
kfp-app-name
kfp-app-version
kfp-artifact-bucket-name
```
Kept `kfp-default-pipeline-root` as it is used in 
`base/pipeline/kfp-launcher-configmap.yaml`
.

[MODIFY]  postgres/kustomization.yaml
Removed the same unused `vars`.

### Verification Results
Automated Tests
Ran `kustomize build .` in `manifests/kustomize/base/installs/generic` and `manifests/kustomize/base/installs/generic/postgres`.

- Result: The "well-defined vars that were never replaced" warnings are gone.
- Note: A warning `# Warning: 'vars' is deprecated. Please use 'replacements' instead.` remains, which is expected since we are still using one legacy var `kfp-default-pipeline-root`.

Manual Verification
Verified that `kfp-launcher` ConfigMap is generated with` defaultPipelineRoot: "" (empty string)`, confirming that the variable `$(kfp-default-pipeline-root)` was correctly substituted.

```
apiVersion: v1
data:
  defaultPipelineRoot: ""
kind: ConfigMap
metadata:
  name: kfp-launcher
  namespace: kubeflow
```

Fixes : #12615

### DIFF
--- a/manifests/kustomize/base/installs/generic/kustomization.yaml
+++ b/manifests/kustomize/base/installs/generic/kustomization.yaml
@@ -9,34 +9,6 @@ resources:
 - mysql-secret.yaml
 vars:
 - fieldref:
-    fieldPath: metadata.namespace
-  name: kfp-namespace
-  objref:
-    apiVersion: apps/v1
-    kind: Deployment
-    name: ml-pipeline
-- fieldref:
-    fieldPath: data.appName
-  name: kfp-app-name
-  objref:
-    apiVersion: v1
-    kind: ConfigMap
-    name: pipeline-install-config
-- fieldref:
-    fieldPath: data.appVersion
-  name: kfp-app-version
-  objref:
-    apiVersion: v1
-    kind: ConfigMap
-    name: pipeline-install-config
-- fieldref:
-    fieldPath: data.bucketName
-  name: kfp-artifact-bucket-name
-  objref:
-    apiVersion: v1
-    kind: ConfigMap
-    name: pipeline-install-config
-- fieldref:
     fieldPath: data.defaultPipelineRoot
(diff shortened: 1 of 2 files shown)


shreeharshshinde:
@droctothorpe @gmfrasca I made some removals according to the issue, please have a look whenever you consider !!

PR #12648, opened 2026-01-14
Fix the reconciliation bug:

ntny:

Fix the reconciliation bug

Previously, the reconciliation code filtered jobs by pod.namespace. At the time, this might have made sense, as reconciliation could occur within the users' namespaces.

However, in the current implementation, reconciliation happens inside the API server, where the pod.namespace is always kubeflow. As a result, recurring runs were effectively not updated and continued to operate with the driver/launcher versions they were originally created with. In single-user mode, filtering by namespace seems entirely unnecessary.

Consider moving the reconciliation logic to a separate service in the future. Currently, the API server could be deployed with a replicaset size >1, which may result in reconciliation being executed multiple times unnecessarily.

сс: @HumairAK @droctothorpe @juliusvonkohout

### DIFF
--- a/backend/src/apiserver/model/resource_reference.go
+++ b/backend/src/apiserver/model/resource_reference.go
@@ -181,6 +181,10 @@ type FilterContext struct {
 	*ReferenceKey
 }
 
+func EmptyFilterContext() *FilterContext {
+	return &FilterContext{}
+}
+
 // Checks whether the resource-reference relationship combination is valid.
 func ValidateResourceReferenceRelationship(resType ResourceType, refType ResourceType, relType Relationship) bool {
 	check, err := validResourceReferenceRelationship[ResourceReferenceRelationshipTriplet{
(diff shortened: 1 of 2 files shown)


droctothorpe:
> However, in the current implementation, reconciliation happens inside the API server, where the pod.namespace is always kubeflow. As a result, recurring runs were effectively not updated and continued to operate with the driver/launcher versions they were originally created with. In single-user mode, filtering by namespace seems entirely unnecessary.

I'm confused by this. In multi-user mode, aren't the CRs in end-user namespaces?

ntny:
> I'm confused by this. In multi-user mode, aren't the CRs in end-user namespaces?

Yes, the CRs are in the user namespaces.
However, in the current implementation, reconciliation happens inside the api-server, which itself runs in the kubeflow namespace.
As a result, pod.namespace is always kubeflow during reconciliation.
This is why recurring runs in multi-user mode were not updated during the reconcillation

droctothorpe:
Thanks for clarifying and for tackling this, @ntny!

/retest
/lgtm
/approve

PR #12657, opened 2026-01-16
fix(ci): port-forward minio-service instead of pod

kaikaila:
- **Problem**: Port-forward script searched for pods with `app=minio` label, which fails when using seaweedfs backend (pods have `app=seaweedfs` label)
- **Solution**: Port-forward directly to `service/minio-service`, which works with both minio and seaweedfs backends
- **Impact**: E2E tests can now access archived workflow logs regardless of storage backend


hbelmiro (approved):
/lgtm
/approve

PR #12664, opened 2026-01-16
fix(compiler): Support None/null values in pipeline parameters (#12240)

### LINKED ISSUE #12240: [bug] Compiling pipeline parameters with NoneType raises protobuf error
### Environment

Deployed on Vertex AI by compiling to YAML and then calling `PipelineJob.submit()`. Currently failing at compilation step, however.

*  KFP version:

*  KFP SDK version: 2.5, 2.7, 2.13.0 (Python SDK)

### Steps to reproduce

We have the following compilation function, which we use to convert a YAML file to a dataclass for parameter sanitisation, which is then turned into a dictionary for compilation.

```python
from pathlib import Path
from dataclasses import asdict
from kfp.components import BaseComponent

def compile_pipeline(
    path_to_yaml_config: Path | str, 
    pipeline_config_dataclass: type[PipelineConfig], 
    pipeline_func: BaseComponent, 
    pipeline_package_path: Path | str | None = None
):
    [... sanity checks etc]

    with path_to_yaml_config.open() as fp:
          yaml_dict = yaml.safe_load(fp)
    pipeline_config = pipeline_config_dataclass(**yaml_dict)

     [...]

     Compiler().compile(
        pipeline_func=pipeline_func,
        package_path=...,
        pipeline_parameters=asdict(pipeline_config)
     )
```

When the YAML config file contains `null` values (usually nested inside a mapping or sequence, but also top-level), these get converted to Python `None` values when loading in the YAML file. These values have previously compiled and submitted without error. 

This now leads to a protobuf error within the recursive call to `kfp.compiler.pipeline_spec_builder.to_protobuf_value()` :
> ... line 79 in to_protobuf_value
>   struct_value=struct_pb2.Struct(
>TypeError: bad argument type for built-in operation

Our version of protobuf hasn't changed and I can see that there is a `_NullValue` type in the protobuf types file in `google.protobuf.struct_pb2`. Removing all the Python `None`s and YAML `null`s causes the error to change but is still occurring in the recursive calls to `to_protobuf_value()`. 

I think the `if-else isinstance` checks in `to_protobuf_value()` are missing the `NoneType` case, or alternatively the compiler has changed in some way that causes `NoneType` to raise an exception.

Depending on what nested structures are in the YAML file, sometimes those nested structures are unpacked and lead to:
> ValueError('Value must be one of the following types...')
which is from the `else`-block of the `isinstance()` checks in `to_protobuf_value()`.

I would like to know what has changed between these minor versions of KFP that could have caused this change in behaviour.

### Expected result

The pipeline compiles. We have successfully compiled and submitted pipelines in exactly the same fashion on KFP 2.5, 2.7 and 2.13. The behaviour appears to have changed recently but I can't see any recent changes to the compiler or protobuf that would affect this.

### Labels

/area sdk

---

BhoomiAgrawal12:
This PR fixes a compilation failure that occurred when pipeline parameters contained `None` values, which commonly arise from YAML `null` values.

**The Problem:**
When users loaded pipeline configurations from YAML files containing `null` values, the compiler would fail with either a `TypeError: bad argument type for built-in operation` or `ValueError: Value must be one of the following types...`. This happened because the `to_protobuf_value()` function didn't handle Python's `None` type, even though:
- Google Protocol Buffers natively support null values via `struct_pb2.NULL_VALUE`
- YAML's `null` is a standard scalar value that converts to Python's `None`
- Users frequently use `null` for optional or unset parameters

**The Solution:**
This PR adds proper `None` handling by:
1. Adding `None` to the `PARAMETER_TYPES` type hint to accurately reflect supported types
2. Checking for `None` values first in `to_protobuf_value()` and converting them to `struct_pb2.Value(null_value=struct_pb2.NULL_VALUE)`
3. Updating error messages to reflect that `None` is now a valid type
4. Adding 13 comprehensive unit tests covering None at all nesting levels

**What Now Works:**
```python
# YAML with null values now compiles successfully
config = yaml.safe_load("""
  text_param: "hello"
  optional_param: null
  nested:
    key: null
  list_param: [1, null, 3]
""")
```
**Related Issue:**
Fixes #12240

### DIFF
--- a/sdk/python/kfp/compiler/pipeline_spec_builder.py
+++ b/sdk/python/kfp/compiler/pipeline_spec_builder.py
@@ -63,8 +63,11 @@ def to_protobuf_value(value: type_utils.PARAMETER_TYPES) -> struct_pb2.Value:
     Raises:
         ValueError if the given value is not one of the parameter types.
     """
+    # None check must come first to handle null values from YAML
+    if value is None:
+        return struct_pb2.Value(null_value=struct_pb2.NULL_VALUE)
     # bool check must be above (int, float) check because bool is a subclass of int so isinstance(True, int) == True
-    if isinstance(value, bool):
+    elif isinstance(value, bool):
         return struct_pb2.Value(bool_value=value)
     elif isinstance(value, str):
         return struct_pb2.Value(string_value=value)
@@ -81,7 +84,7 @@ def to_protobuf_value(value: type_utils.PARAMETER_TYPES) -> struct_pb2.Value:
                 values=[to_protobuf_value(v) for v in value]))
     else:
         raise ValueError('Value must be one of the following types: '
-                         'str, int, float, bool, dict, and list. Got: '
+                         'str, int, float, bool, dict, list, and None. Got: '
                          f'"{value}" of type "{type(value)}".')
 
 
--- a/sdk/python/kfp/compiler/pipeline_spec_builder_test.py
+++ b/sdk/python/kfp/compiler/pipeline_spec_builder_test.py
@@ -561,5 +561,111 @@ def pipeline_spec_from_file(filepath: str) -> str:
     return json_format.ParseDict(dictionary, pipeline_spec_pb2.PipelineSpec())
 
 
+class ToProtobufValueTest(unittest.TestCase):
+
+    def test_none_value(self):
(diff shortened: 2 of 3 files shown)


BhoomiAgrawal12:
Hey @zazulam @nsingla , Can you please review my PR linked to this issue.

Thank You.

PR #12666, opened 2026-01-16
Fix pipeline parameter propagation for PVC mounts inside conditionals

### LINKED ISSUE #12617: [bug] kfp.kubernetes.mount_pvc with pvc_name passed as param fails inside conditional
Mounting a pvc to a task inside a conditional (kfp.dsl.If()) fails when the pvc name comes from a parameter. It appears the conditional statement is preventing the passing along of top-level pipeline arguments, as the example below illustrates.

### Environment

*  KFP version:
    Backend tested was using Openshift AI 3.0 (dsp 2.5)

*  KFP SDK version:
    kfp version = 2.15.2

### Steps to reproduce

The following pipeline produces the issue:
``` python
from kfp.dsl import component, pipeline, If

import kfp.kubernetes as k8s

pvc_name = "gemma-3n-e4b-it"
namespace = 'pipeline-test'

# dummy component 1
@component(base_image="python:3.12-slim-bullseye")
def dummy_component1():
    print("dummy component")

# dummy component 2
@component(base_image="python:3.12-slim-bullseye")
def dummy_component2():
    print("dummy component 2")

# pipeline
@pipeline(display_name="debug-pvc-name-issue")
def debug_pvc_name_issue(
  pvc_name: str = pvc_name,
  condition: bool = True
):
    task1 = dummy_component1()
    k8s.mount_pvc(
        task1,
        pvc_name=pvc_name,
        mount_path="/models"
    )

    with If(condition == True):
        task2 = dummy_component2().after(task1)
        k8s.mount_pvc(
            task2,
            pvc_name=pvc_name,
            mount_path="/models"
        )
```

Run the pipeline with the following code:

```python
from kfp.client import Client

# submit pipeline run
client = Client(
  host=<pipeline_host>,
  namespace=namespace,
  existing_token=<token>
)

client.create_run_from_pipeline_func(
    debug_pvc_name_issue,
    namespace=namespace,
    experiment_name="default",
    enable_caching=False,
    arguments={
        "pvc_name": pvc_name,
        "condition": True
    }
)
```

You may also compile the pipeline to yaml, and then submit through the GUI.

The pipeline fails on the second task with this in logs: 
```
...
I0107 17:51:32.864075      22 resolve.go:118] parent DAG input parameters: map[pipelinechannel--condition:bool_value:true], artifacts: map[]
F0107 17:51:32.864170      22 main.go:108] KFP driver: driver.Container(pipelineName=debug-pvc-name-issue, runID=cc6929ce-ae18-4dca-964f-c2400eb59909, taskDisplayName="dummy-component2", taskName="dummy-component2", component="comp-dummy-component2", dagExecutionID=35, componentSpec, KubernetesExecutorConfig) failed: failed to extract volume mount info while generating fingerprint: failed to resolve pvc name: resolving input parameter with spec component_input_parameter:"pvc_name": parent DAG does not have input parameter pvc_name
time="2026-01-07T17:51:33.574Z" level=info msg="sub-process exited" argo=true error="<nil>"
time="2026-01-07T17:51:33.574Z" level=error msg="cannot save parameter /tmp/outputs/pod-spec-patch" argo=true error="open /tmp/outputs/pod-spec-patch: no such file or directory"
time="2026-01-07T17:51:33.574Z" level=error msg="cannot save parameter /tmp/outputs/cached-decision" argo=true error="open /tmp/outputs/cached-decision: no such file or directory"
time="2026-01-07T17:51:33.574Z" level=error msg="cannot save parameter /tmp/outputs/condition" argo=true error="open /tmp/outputs/condition: no such file or directory"
Error: exit status 1
```

Note that removing the conditional statement makes the pipeline run successfully. Also, hard-coding the pvc_name as a string inside the conditional (instead of passing it as a variable) is a workaround. 

### Expected result

The `pvc_name` variable (pipeline parameter) should be able to be passed through to the mount_pvc function inside a conditional statement (just like it is possible outside the conditional), and the pipeline should run without error.
 
### Materials and reference

This may be related to:
https://github.com/kubeflow/pipelines/issues/10918

### Labels

---

Goku2099:
Fixes #12617
### What this PR does
Fixes an issue where pipeline parameters used by k8s.mount_pvc were not
propagated into conditional (dsl.If) DAGs.

### Problem
Pipeline parameters referenced only through task.platform_config were not
included when building conditional DAG inputs. As a result, pvc_name worked
outside dsl.If but failed inside it.

### Solution
The compiler now collects pipeline channels referenced in platform_config
and propagates them into conditional DAG inputs.

### Tests
- Added a compiler test that verifies pvc_name is available when mount_pvc
  is used inside a dsl.If block.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -104,3 +104,11 @@ logs/
 
 # Project-local tools
 bin/
+# Local developer virtualenv (custom name)
+k/
+
+# Local repro / debug scripts
+repro_pvc_conditional.py
+
+# Local scratch directory (not repo kubernetes code)
+kubernetes/
--- a/sdk/python/kfp/compiler/compiler_utils.py
+++ b/sdk/python/kfp/compiler/compiler_utils.py
@@ -173,6 +173,18 @@ def get_condition_channels_for_tasks(
 
     _get_condition_channels_for_tasks_helper(root_group, [])
     return conditions
+def _get_channels_from_platform_config(task):
+    """Extract PipelineChannels from task.platform_config (e.g. pvcMount)."""
+    channels = []
+    if not task.platform_config:
+        return channels
+
+    for cfg in task.platform_config.values():
+        channels.extend(
+            pipeline_channel.extract_pipeline_channels_from_any(cfg)
+        )
+
+    return channels
 
 
 def get_inputs_for_all_groups(
@@ -209,8 +221,10 @@ def get_inputs_for_all_groups(
         # task's inputs and all channels used in conditions for that task are
         # considered.
         task_condition_inputs = list(condition_channels[task.name])
+        platform_channels = _get_channels_from_platform_config(task)
 
-        for channel in task.channel_inputs + task_condition_inputs:
+        for channel in task.channel_inputs + platform_channels + task_condition_inputs:
+
 
             # If the value is already provided (immediate value), then no
             # need to expose it as input for its parent groups.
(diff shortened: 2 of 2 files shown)


Goku2099:
@HumairAK @nsingla
Hi! This PR is ready for review now. Tests are passing. Would appreciate your feedback when you have time. Thanks!

PR #12676, opened 2026-01-19
fix(compiler): apply retryPolicy per item for tasks in dsl.ParallelFor

### LINKED ISSUE #12389: [sdk] set_retry is ignored for tasks in dsl.ParallelFor
### Environment

*  KFP version:
*  KFP SDK version: 2.12.1

### Steps to reproduce

When used as a single task, .set_retry(N) works as expected. When the same component is used inside dsl.ParallelFor, retries do not trigger per item: a failing item fails once and the pipeline marks that branch failed without retrying.

```
from kfp import dsl, compiler
from typing import List

@dsl.component
def random_failure_op(exit_codes: str, worker_id: str):
    """Randomly exits with one of the provided exit codes.

    Args:
        exit_codes: Comma-separated list of integer exit codes, e.g. "0,1,2".
        worker_id: Label to identify the parallel worker in logs.
    """
    import random
    import sys
    import time

    code = int(random.choice(exit_codes.split(",")))
    print(f"[{worker_id}] picked exit code: {code}")
    time.sleep(1.0)
    sys.exit(code)


@dsl.pipeline(name="parallel-retry-demo")
def parallel_retry_pipeline(
    workers: List[str] = ["A", "B", "C", "D", "E"],
    exit_codes: str = "0,1,2,3",
):
    with dsl.ParallelFor(workers) as wid:
        t = random_failure_op(exit_codes=exit_codes, worker_id=wid)
        t.set_retry(20)
        t.set_caching_options(False)


if __name__ == "__main__":
    compiler.Compiler().compile(parallel_retry_pipeline, __file__ + '.yaml')
```

### Expected result

Each dsl.ParallelFor item should be its own tasks and respect .set_retry(...) independently, i.e., on failure it should retry up to the configured number of times per item (similar to behavior outside ParallelFor).

---

shreeharshshinde:
### What this PR does

Fixes an issue where `Task.set_retry()` is ignored for tasks inside `dsl.ParallelFor`.

When a component with a retryPolicy is used inside `ParallelFor`, the compiler currently attaches the retryStrategy at a level that Argo does not retry per iteration item. As a result, a failing item is not retried and the branch fails immediately.

This PR lifts the retryPolicy from the single task in the loop body and applies it to the iteration template, enabling retries to be executed independently per item.

### Why this is needed

- Standalone tasks correctly respect `set_retry()`, but tasks inside `dsl.ParallelFor` do not
- This behavior is inconsistent and surprising for users
- Argo retries are applied per template execution, so the retryStrategy must live on the iteration template to enable per-item retries

### Scope of change

- Affects only `dsl.ParallelFor`
- Applies only when the loop body contains exactly one task with a retryPolicy
- No changes to SDK, DSL surface, or runtime behavior
- Preserves existing semantics for non-ParallelFor components and multi-task DAGs

### How it works

- Detects a retryPolicy on the single task inside a ParallelFor loop body
- Lifts the retryPolicy to the iteration template generated by the Argo compiler
- This allows Argo to retry each iteration item independently

### Testing

- Added a compiler unit test that verifies retryStrategy is correctly applied to the iteration template for ParallelFor tasks

### Related issue

Fixes #12389

### DIFF
--- a/backend/src/v2/compiler/argocompiler/dag.go
+++ b/backend/src/v2/compiler/argocompiler/dag.go
@@ -17,6 +17,7 @@ import (
 	"fmt"
 	"os"
 	"sort"
+	"strconv"
 	"strings"
 
 	"github.com/kubeflow/pipelines/backend/src/apiserver/config/proxy"
@@ -441,6 +442,7 @@ func (c *workflowCompiler) iterationItemTask(name string, task *pipelinespec.Pip
 		DAG: &wfapi.DAGTemplate{
 			Tasks: iterationTasks,
 		},
+		RetryStrategy: c.getParallelForItemRetryStrategy(componentName),
 	}
 	iterationsTmplName, err := c.addTemplate(iterationsTmpl, componentName+"-"+name)
 	if err != nil {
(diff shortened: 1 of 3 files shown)


shreeharshshinde:
@gmfrasca @zazulam
I request you to please review my authored PR for the linked issue. I would be happy to make any changes or corrections if required.

mr-morj:
Hey! Just checking in—how's this MR going? Any updates?

PR #12679, opened 2026-01-20
test(sdk): Replace deprecated dsl.Condition with dsl.If in compiler tests

Vedant2005goyal:
This PR replaces the use of old/depreciated dsl.Condition with the modern dsl.If .
No functional changes.

### DIFF
--- a/.github/actions/test-and-report/action.yml
+++ b/.github/actions/test-and-report/action.yml
@@ -102,7 +102,7 @@ runs:
         fi
         PULL_NUMBER="${{ github.event.inputs.pull_number || github.event.pull_request.number }}"
         REPO_NAME="${{ github.repository }}"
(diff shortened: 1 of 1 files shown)


sdk/python/kfp/compiler/compiler_test.py
```diff
@@ -626,12 +626,11 @@ def test_invalid_data_dependency_condition(self):
 
         with self.assertRaisesRegex(
                 compiler_utils.InvalidTopologyException,
-                r'Illegal task dependency across DSL context managers\. A downstream task cannot depend on an upstream task within a dsl\.Condition context unless the downstream is within that context too\. Found task dummy-op which depends on upstream task producer-op within an uncommon dsl\.Condition context\.'
-        ):
+                r'Illegal task dependency .* within a dsl\.If context .*'):
```
zazulam, line 629: qq: What's the reason for modifying the exception message?

Vedant2005goyal: Hi @zazulam,
Since I updated the test code to use dsl.If, asserting for dsl.If in the error message caused a failure because the underlying compiler error message still references the old dsl.Condition. I updated the regex to use wildcards (.*) so the test verifies the logic correctly.

Issue #12546, opened 2025-12-10
[sdk] Executor crashes with TypeError when using Input[List[Dataset]] with dsl.Collected


wassimbensalem:
### Environment

* KFP version: N/A (SDK bug, not server-specific)
* KFP SDK version: 2.15.2 (and likely earlier versions)
* All dependencies version:
- kfp==2.15.2
- kfp-pipeline-spec==2.15.2
- kfp-server-api==2.15.2

### Steps to reproduce

When a component uses `Input[List[Dataset]]` with `dsl.Collected`, the executor crashes at runtime.

### Code:
```

from kfp import dsl
from kfp.dsl import Input
from typing import List
from kfp.dsl.types.artifact_types import Dataset

@dsl.component
def aggregate_predictions(input_predictions_list: Input[List[Dataset]]) -> str:
    return f"Processed {len(input_predictions_list)} datasets"

@dsl.pipeline
def my_pipeline():
    predict_op = some_component_that_outputs_predictions()
    collected = dsl.Collected(predict_op.outputs["output_predictions"])
    aggregate_predictions(input_predictions_list=collected)  # ← Crashes here
```
### Error:

TypeError: list() takes no keyword arguments

###Root Cause:

The executor tries to instantiate List[Dataset] by calling list(**kwargs), but Python's list() doesn't accept keyword arguments. The code extracts List[Dataset] as the artifact class instead of extracting the inner type (Dataset).

### Expected result
The component should receive a Python list of Dataset instances, where each item is a properly instantiated Dataset artifact object.

---

Impacted by this bug? Give it a 👍.

Mateusz-Switala:
Any updates on that issue? It also does not work with kfp==2.14.4. It seems that the example from Kubeflow Pipelines [documentation](https://www.kubeflow.org/docs/components/pipelines/user-guides/core-functions/control-flow/#dslcollected) is now outdated due to this issue...

zazulam:
Resolved with #12707

PR #12707, opened 2026-01-26
Fix executor handling for collected list artifacts

sduvvuri1603:
### Bug

- Components that declare Input[List[Dataset]] and receive the value via dsl.Collected(...) were crashing at runtime with TypeError: list() takes no keyword arguments. 

- The executor was reading the type annotation straight from the component signature, but instead of peeling off the inner Dataset class it treated the entire List[Dataset] annotation as the “class” to instantiate. 

- Since Python’s list() constructor doesn’t accept keyword arguments, the executor failed before the component ever ran.

### Fix

- Inside Executor.assign_input_and_output_artifacts and Executor.make_artifact we now normalize list annotations by repeatedly unwrapping their inner type until only an artifact class remains. 

- If we encounter a union (List[Dataset | Model], etc.) or something that isn’t an artifact, we raise a descriptive TypeError so the issue is surfaced explicitly. 

- With the inner annotation unwrapped to Dataset, the executor instantiates each element correctly and assembles the expected Python list. 

- We cemented the behavior with a regression test that feeds the executor two system.Dataset artifacts, confirming the component receives a list of Dataset instances without errors.

### DIFF
--- a/sdk/python/kfp/dsl/executor.py
+++ b/sdk/python/kfp/dsl/executor.py
@@ -104,7 +104,8 @@ class Executor:
 
     def _resolve_list_artifact_inner_type(self, input_name: str,
                                           list_annotation: Any) -> Any:
-        """Unwraps nested list annotations and validates the inner artifact type."""
+        """Unwraps nested list annotations and validates the inner artifact
+        type."""
         inner_annotation = type_annotations.get_inner_type(list_annotation)
 
         if inner_annotation is None:


sduvvuri1603:
> Also can you share an example pipeline in a comment of the issue you were encountering?

Thanks for the review @zazulam!
The pipeline that triggered this is captured in issue [#12546](https://github.com/kubeflow/pipelines/issues/12546). It compiles but fails at runtime when a component with Input[List[Dataset]] receives a dsl.Collected(...) output

Additionally updated code changes to use a helper function and call it where necessary.

PR #12717, opened 2026-01-27
chore: expose Kubeflow Pipelines version via public ConfigMap

sameerdattav:
This change exposes the Kubeflow Pipelines control-plane version via a public,
namespaced, read-only ConfigMap, following the same intent as the PR: https://github.com/kubeflow/trainer/pull/3083

Unlike Trainer, Pipelines uses Kustomize as the canonical install path, so this
is implemented using a Kustomize `configMapGenerator` in the shared
namespace-scoped pipeline base. A minimal Role and RoleBinding grant
`get/list/watch` access on the single ConfigMap to `system:authenticated`.

The version defaults to `latest` for non-release installs and is updated to the
release tag as part of the release process (documented in `RELEASE.md`).

Fixes https://github.com/kubeflow/sdk/issues/221

### DIFF
--- a/manifests/kustomize/base/pipeline/kustomization.yaml
+++ b/manifests/kustomize/base/pipeline/kustomization.yaml
@@ -39,7 +39,7 @@ resources:
 configMapGenerator:
   - name: kubeflow-pipelines-public
     literals:
-      - kubeflow_pipelines_version=latest
+      - kubeflow_pipelines_version=dev
     options:
       disableNameSuffixHash: true
 images:


manifests/kustomize/base/pipeline/kustomization.yaml
```diff
@@ -34,6 +34,14 @@ resources:
   - container-builder-sa.yaml
   - viewer-sa.yaml
   - kfp-launcher-configmap.yaml
+  - public_configmap_role.yaml
+  - public_configmap_role_binding.yaml
+configMapGenerator:
+  - name: kubeflow-pipelines-public
+    literals:
+      - kubeflow_pipelines_version=latest
```
andreyvelich: It should be `dev` by default: https://github.com/kubeflow/trainer/blob/master/charts/kubeflow-trainer/templates/_helpers.tpl#L87
Can you also submit PR to fix it in Trainer manifests please: https://github.com/kubeflow/trainer/blob/master/manifests/overlays/manager/kustomization.yaml#L31

sameerdattav: @andreyvelich , I changed latest to dev, and
I also opened the PR to fix it in trainer manifests: 
https://github.com/kubeflow/trainer/pull/3146

Kindly review again.

droctothorpe:
/lgtm
/approve

Thanks for tackling this, @sameerdattav!

sameerdattav:
@droctothorpe 
Thank you!

PR #12718, opened 2026-01-28
Update ML Pipeline server address to facilitate TLS-enabled gRPC connection

alyssacgoins:
**Description of your changes:**

This PR updates the TLS config (used only when specifically enabled) on the API server with field `ServerName` in order for the TLS-enabled gRPC connection to successfully serve requests for certs without the `localhost` dns name.  `GetMLPipelineServerConfig()` in `backend/src/v2/config/env.go` is also updated to append `".svc.cluster.local"` to the end of the configured address.

Note that `localhost` is retained on `manifests/kustomize/env/cert-manager/base-tls-certs/kfp-api-cert.yaml`, due to this manifest's use in KFP integration testing.  

Resolves #12720 

**Checklist:**
- [ ] You have [signed off your commits](https://www.kubeflow.org/docs/about/contributing/#sign-off-your-commits)
- [ ] The title for your pull request (PR) should follow our title convention. [Learn more about the pull request title convention used in this repository](https://github.com/kubeflow/pipelines/blob/master/CONTRIBUTING.md#pull-request-title-convention).

PR #12729, opened 2026-01-29
chore(manifests): remove "localhost" from kfp-api-cert

### LINKED ISSUE #12720: chore(manifests): Remove "localhost" from kfp-api-cert
## Chore description
`manifests/kustomize/env/cert-manager/base-tls-certs/kfp-api-cert.yaml` currently contains dnsName `"localhost"`, which is not good practice. This should be removed.

### Labels

---

Love this idea? Give it a 👍.

Sayan4496:
Fixes #12720

Removed "localhost" from kfp-api-cert dnsNames list to avoid insecure/non-standard certificate SAN configuration.


Sayan4496:
It looks like this issue has already been addressed in #12718.
Closing this PR to avoid duplicate work. Thanks!

PR #12749, opened 2026-01-31
feat: Add use_local_pip_config parameter with all 7 security and corr…

### LINKED ISSUE #12731: [feature] Use the user's local pip conf
### Feature Area

/area sdk

/area components

Currently, when making components, users can specify the pip index url:
```
@dsl.component(
    pip_index_urls=["<>"]
)
```
This works, but there are a number of other pip configuration commands that may be in a user's local pip configuration. Specifying that the components should be built by inheriting the user's local pip conf file would allow more custom configurations. 

Specifying the local pip would allow users to have more complex pip configuration and avoid re-specifying the pip configuration for each component.

Specify config manually for each component.

---

Love this idea? Give it a 👍.

Shivam140802:
This PR adds the use_local_pip_config parameter to KFP components. It allows components to automatically inherit safe pip configuration settings from the user's local pip.conf file.

What it does

Reads pip configuration from ~/.config/pip/pip.conf or ~/.pip/pip.ini
Safely filters out credentials and passwords
Extracts only safe options: index-url, extra-index-url, trusted-host, timeout, retries, no-cache-dir, disable-pip-version-check
Explicit parameters always take precedence over local config
-Works with both @dsl.component and @dsl.notebook_component decorators
-CLI component builder correctly handles multiple components
Changes

component_factory.py - Core pip config parsing logic
component_decorator.py - Added use_local_pip_config parameter
notebook_component_decorator.py - Added use_local_pip_config parameter
component.py - CLI builder support
Added 3 new tests (all passing)
Usage

`from kfp import dsl

@dsl.component(use_local_pip_config=True)
def my_component(x: int) -> int:
return x * 2`

Testing

All 54 existing tests pass + 3 new tests added +  resolve previous 7 issues 

Fixes https://github.com/kubeflow/pipelines/issues/12731


zazulam:
@Shivam140802 please run the pre-commit hooks associated with the repo

Shivam140802:
> @Shivam140802 please run the pre-commit hooks associated with the repo

All CI checks are now passing.
Kindly requesting review. Thanks!

zazulam:
@Shivam140802 sorry for the delay, can you please rebase and address the remaining copilot suggestions? Feel free to either tag me or ping me in the cncf slack after.

Shivam140802:
@zazulam looking forward for review

zazulam (changes requested):
## PR Overview

Thanks for working on local pip configuration support. This needs another revision before it is ready for maintainer review.

[0m

Please address the following issues:

- **Shell injection:** Values read from the local pip config are interpolated unquoted into `sh -c` and Dockerfile `RUN` commands. A URL or trusted host containing shell metacharacters can execute arbitrary commands and expose runtime or build secrets. Please avoid shell-string construction where possible, or correctly quote every imported value.
- **Malformed pip commands:** Calling `.strip()` when combining the generated options removes the separator expected by both command templates. This produces tokens such as `--timeout 60'package1'` and `--timeout 60--no-cache-dir`. Please construct arguments with explicit token boundaries and add tests that validate the complete generated commands rather than only checking substrings.
- **Incomplete config discovery:** The hard-coded paths omit standard macOS, Windows, and legacy Unix locations; `~/.pip/pip.ini` is also not the standard legacy Unix filename. Please reuse pip's configuration discovery/loading semantics rather than maintaining a partial path list.
- **Incorrect `no-cache-dir` semantics:** pip treats the presence of any valid configured value, including `false`, as disabling caching for backward compatibility. The current `getboolean()` handling drops false values and therefore does not preserve pip behavior.
- **Config parsing incompatibility:** `ConfigParser` enables interpolation, while valid pip values may contain percent-encoded characters. Use `RawConfigParser` or disable interpolation so accepted pip URLs do not raise `InterpolationSyntaxError`.

The malformed-command and interpolation issues also have unresolved inline review threads. Please address the outstanding review comments and **rebase this PR onto the latest `master`** before requesting another review.

droctothorpe:
Hi @Shivam140802, thank you for your PR! The linked issue (#12731) must be labeled `ready` before your contribution can be reviewed. Please check our [CONTRIBUTING.md](https://github.com/kubeflow/pipelines/blob/master/CONTRIBUTING.md) for details.

PR #12750, opened 2026-01-31
feat(backend): Remediate ADA-KUBEFL-01: Server-Side Request Forgery

JerT33:

## Background

Fixes SSRF vulnerability (ADA-KUBEFL-01) in CreatePipelineV1/V2 APIs where user-provided URLs are fetched without destination validation

```
Recommended remediation
Restrict where the backend will fetch from and validate the final destination:
• Replace arbitrary URLs with file uploads or server-generated pre-signed URLs for controlled object storage.
• If URLs must remain, enforce an allow-list of trusted domains; block private/loopback/link-local/metadata IP ranges
after DNS resolution and on every redirect; restrict schemes/ports (prefer https:// on standard ports); and use tight
timeouts.
```

To avoid disruption of any current users who rely on URLs for pipeline upload functionality, the following remediations have been completed:

- [X] Domain allow-list
- [X] Block private/loopback/link-local/metadata IP ranges
- [X] DNS resolution before request
- [X] Validation on every redirect
- [X] Restrict schemes
- [X] Restrict ports
- [X] Request timeout

### Validate Pipeline URL:
1. Scheme check 
   - https required by default
   - user controlled http allowance
2.  Port check
    - https restricted to port 443
    - http restricted to port 80 
3.  DNS allow-list check
    - validate DNS against a preset list of common DNS 
        - github
        - aws s3
        - google cloud storage
    - user controlled additional allowed DNS
4.  DNS resolution
5. IP check
   - validate against a preset non-configurable blacklist of CIDR ranges
       - private
       - loopback
       - link-local
       - metadata
       - invalid 

### SafePipelineHTTPClient:
- Implement timeouts
   - defaults to 30 seconds
   - user controlled override
- Restrict redirects
   - maximum of 10
   - validate pipeline url on every redirect

### General:
- Prevent XSS attacks by returning generic error messages, but log full error report 

### User-configurable options:

* `PIPELINE_URL_ALLOWED_DOMAINS`: extend the default domain allow-list
* `PIPELINE_URL_ALLOW_HTTP`: allow HTTP URLs (default false)
* `PIPELINE_URL_TIMEOUT`: configure request timeout in seconds (default 30)
* `PIPELINE_URL_VALIDATION_ENABLED`: toggle pipeline url validation - for testing purposes (default true)

### Before:
Screenshot 2026-02-04 at 10 42 49 PM
Screenshot 2026-02-04 at 10 48 56 PM
Screenshot 2026-02-04 at 10 43 58 PM

### After
Screenshot 2026-02-04 at 11 16 55 PM
Screenshot 2026-02-04 at 11 16 47 PM
Screenshot 2026-02-04 at 11 17 10 PM
(no additional URL fetch from latest test)

**Checklist:**

### DIFF
--- a/backend/src/apiserver/server/pipeline_server_test.go
+++ b/backend/src/apiserver/server/pipeline_server_test.go
@@ -41,6 +41,12 @@ import (
 	authorizationv1 "k8s.io/api/authorization/v1"
 )
 
+func TestMain(m *testing.M) {
+	// Disable URL validation for tests (allows mock server URLs)
+	viper.Set(common.PipelineURLValidationEnabled, "false")
+	os.Exit(m.Run())
+}
+
 func createPipelineServerV1(resourceManager *resource.ResourceManager, httpClient *http.Client) *PipelineServerV1 {
 	return &PipelineServerV1{
 		BasePipelineServer: &BasePipelineServer{
@@ -1234,4 +1240,94 @@ func TestCanAccessPipeline_SharedPipeline_ReadAllowed_EvenWhenUnauthorized(t *te
 	// LIST on shared pipeline should still be allowed even for unauthorized user
 	err = pipelineServer.canAccessPipeline(ctx, "", &authorizationv1.ResourceAttributes{Verb: common.RbacResourceVerbList})
 	assert.Nil(t, err)
+}
+
+func TestCreatePipelineV1_URLValidation_BlocksDisallowedDomain(t *testing.T) {
+	// Enable URL validation for this test (overrides TestMain's global disable)
+	viper.Set(common.PipelineURLValidationEnabled, "true")
+	defer viper.Set(common.PipelineURLValidationEnabled, "false")
+
+	clientManager := resource.NewFakeClientManagerOrFatal(util.NewFakeTimeForEpoch())
+	resourceManager := resource.NewResourceManager(clientManager, &resource.ResourceManagerOptions{CollectMetrics: false})
+
+	pipelineServer := createPipelineServerV1(resourceManager, http.DefaultClient)
+	_, err := pipelineServer.CreatePipelineV1(
+		context.Background(), &api.CreatePipelineRequest{
+			Pipeline: &api.Pipeline{
+				Url:  &api.Url{PipelineUrl: "https://evil.com/malicious.yaml"},
+				Name: "test-blocked-domain",
+			},
+		},
+	)
+	assert.NotNil(t, err)
+	assert.Contains(t, err.Error(), "Pipeline URL validation failed")
+}
+
+func TestCreatePipelineV1_URLValidation_BlocksHTTPByDefault(t *testing.T) {
+	viper.Set(common.PipelineURLValidationEnabled, "true")
+	defer viper.Set(common.PipelineURLValidationEnabled, "false")
+
+	clientManager := resource.NewFakeClientManagerOrFatal(util.NewFakeTimeForEpoch())
+	resourceManager := resource.NewResourceManager(clientManager, &resource.ResourceManagerOptions{CollectMetrics: false})
+
+	pipelineServer := createPipelineServerV1(resourceManager, http.DefaultClient)
+	_, err := pipelineServer.CreatePipelineV1(
+		context.Background(), &api.CreatePipelineRequest{
+			Pipeline: &api.Pipeline{
+				Url:  &api.Url{PipelineUrl: "http://storage.googleapis.com/bucket/file.yaml"},
+				Name: "test-http-blocked",
+			},
+		},
+	)
+	assert.NotNil(t, err)
+	assert.Contains(t, err.Error(), "Pipeline URL validation failed")
+}
+
+func TestCreatePipelineV1_URLValidation_BlocksFileScheme(t *testing.T) {
+	viper.Set(common.PipelineURLValidationEnabled, "true")
+	defer viper.Set(common.PipelineURLValidationEnabled, "false")
+
+	clientManager := resource.NewFakeClientManagerOrFatal(util.NewFakeTimeForEpoch())
+	resourceManager := resource.NewResourceManager(clientManager, &resource.ResourceManagerOptions{CollectMetrics: false})
+
+	pipelineServer := createPipelineServerV1(resourceManager, http.DefaultClient)
+	_, err := pipelineServer.CreatePipelineV1(
+		context.Background(), &api.CreatePipelineRequest{
+			Pipeline: &api.Pipeline{
+				Url:  &api.Url{PipelineUrl: "file:///etc/passwd"},
+				Name: "test-file-blocked",
+			},
+		},
+	)
+	assert.NotNil(t, err)
+	assert.Contains(t, err.Error(), "Pipeline URL validation failed")
+}
+
+type errorRoundTripper struct {
+	err error
+}
+
+func (rt errorRoundTripper) RoundTrip(*http.Request) (*http.Response, error) {
+	return nil, rt.err
+}
+
+func TestCreatePipelineV1_URLValidation_MapsTransportValidationErrorToInvalidInput(t *testing.T) {
+	viper.Set(common.PipelineURLValidationEnabled, "false")
+
+	clientManager := resource.NewFakeClientManagerOrFatal(util.NewFakeTimeForEpoch())
+	resourceManager := resource.NewResourceManager(clientManager, &resource.ResourceManagerOptions{CollectMetrics: false})
+
+	pipelineServer := createPipelineServerV1(resourceManager, &http.Client{
+		Transport: errorRoundTripper{err: util.NewInvalidInputError("connection to blocked IP range denied")},
+	})
+	_, err := pipelineServer.CreatePipelineV1(
+		context.Background(), &api.CreatePipelineRequest{
+			Pipeline: &api.Pipeline{
+				Url:  &api.Url{PipelineUrl: "https://storage.googleapis.com/bucket/file.yaml"},
+				Name: "test-transport-validation-error",
+			},
+		},
+	)
+	assert.NotNil(t, err)
+	assert.Equal(t, codes.InvalidArgument, err.(*util.UserError).ExternalStatusCode())
 }
--- a/backend/src/apiserver/validation/pipeline_url.go
+++ b/backend/src/apiserver/validation/pipeline_url.go
@@ -1,0 +1,255 @@
+// Copyright 2025 The Kubeflow Authors
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+// https://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package validation
+
+import (
+	"context"
+	"fmt"
+	"net"
+	"net/http"
+	"net/url"
+	"regexp"
+	"strings"
+	"sync"
+	"time"
+
+	"github.com/golang/glog"
+	"github.com/kubeflow/pipelines/backend/src/apiserver/common"
+	"github.com/kubeflow/pipelines/backend/src/common/util"
+)
+
+// Default allowed domains for pipeline URLs
+var defaultAllowedDomains = []string{
+	"storage.googleapis.com",
+	"s3.amazonaws.com",
+	"github.com",
+	"raw.githubusercontent.com",
+}
+
+// Blocked CIDRs - private/loopback/link-local/metadata
+var blockedCIDRs = []string{
+	// Loopback
+	"127.0.0.0/8",
+	"::1/128",
+
+	// Private
+	"10.0.0.0/8",
+	"172.16.0.0/12",
+	"192.168.0.0/16",
+	"fc00::/7",
+
+	// Link-local
+	"169.254.0.0/16",
+	"fe80::/10",
+
+	// Metadata
+	"169.254.169.254/32",
+	"fd00:ec2::254/128",
+
+	// Invalid
+	"0.0.0.0/8",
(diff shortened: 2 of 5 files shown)


zazulam:
Bummed I missed this PR, @JerT33 if you can resolve the conflict with master I can merge this

JerT33:
> Bummed I missed this PR, @JerT33 if you can resolve the conflict with master I can merge this

@zazulam I lost track of this PR, thanks for the comment here. I had reached out to a member of the security team to look this over, but never head anything back. I was hoping to get a review from them to ensure it meets their criteria before reaching out to maintainers

nsingla:
> > Bummed I missed this PR, @JerT33 if you can resolve the conflict with master I can merge this
> 
> @zazulam I had actually forgot about this PR, thanks for the comment here. I had reached out to a member of the security team to look this over, but never head anything back. I was hoping to get a review from them to ensure it meets their criteria before reaching out to maintainers

@JerT33 Any updates on this PR?

JerT33:
> @JerT33 Any updates on this PR?

@nsingla I reached out to the security team again for them to take a look, but still never heard back. PR is ready for review/approvals, let me know if there is any feedback. Thanks!

jeffspahr (commented):
I found a few issues that should be addressed before this fully closes the SSRF gap:

- P1 `backend/src/apiserver/validation/pipeline_url.go:168` / `:204`: DNS/IP validation is done before the request, but the HTTP client still uses the default transport, so the actual dial performs a separate DNS lookup. An allowlisted hostname can pass validation and then resolve to a blocked/private address for the real connection. The mitigation should validate in the transport dial path, or resolve once and force the request to use the validated address.

- P2 `backend/src/apiserver/server/pipeline_server.go:210` and `:794`: Redirect validation failures from `CheckRedirect` are currently returned by `http.Client.Get` as errors and wrapped as internal server errors. A disallowed redirect or redirect loop should be surfaced as invalid input, not a 500.

- P3 `backend/src/apiserver/server/pipeline_server_test.go:44`: The server tests disable URL validation globally, so the API-level wiring has no regression coverage for unsafe URL rejection or redirect validation error mapping.

Verification I ran against head `1dc8b390591cfef19c97d8a6371895f869d18066`:

- `go test ./backend/src/apiserver/validation ./backend/src/apiserver/server`
- `git diff --check origin/master...origin/pr/12750`

JerT33:
@jeffspahr thanks for the feedback! PR has been updated for re-review when you have a chance. 
1. moved IP validation into the transport's DialContext so it happens at dial time, instead of a separate step before the request
2. redirect validation failures now return as invalid input errors instead of getting wrapped as 500s
3. added server-level tests that enable validation and cover the reject paths, plus unit tests for the dialer IP blocking

backend/src/apiserver/validation/pipeline_url.go
```diff
@@ -0,0 +1,253 @@
...
+	if timeoutSeconds < minimumHTTPTimeout {
+		timeoutSeconds = 30
+	}
+	timeout := time.Duration(timeoutSeconds) * time.Second
+
+	return &http.Client{
+		Timeout: timeout,
+		Transport: &http.Transport{
+			DialContext: safeDialContext,
+		},
```
jeffspahr: This custom transport drops `http.DefaultTransport` behavior, including `ProxyFromEnvironment`. Proxy-enabled API server deployments set `HTTP_PROXY` / `HTTPS_PROXY`, so start from a clone of the default transport and only override the dialer. If the configured proxy resolves to an in-cluster address, `safeDialContext` also needs to explicitly allow that configured proxy endpoint while continuing to validate direct target dials.

```suggestion
	transport := http.DefaultTransport.(*http.Transport).Clone()
	transport.DialContext = safeDialContext

	return &http.Client{
		Timeout:   timeout,
		Transport: transport,
```

backend/src/apiserver/validation/pipeline_url.go
```diff
@@ -0,0 +1,253 @@
...
+	initURLConfig()
+
+	addrs, err := net.DefaultResolver.LookupIPAddr(ctx, host)
+	if err != nil {
+		return nil, fmt.Errorf("DNS resolution failed for %q: %v", host, err)
+	}
+
+	for _, a := range addrs {
+		if isBlockedIP(a.IP) {
+			return nil, fmt.Errorf("connection to blocked IP %v denied", a.IP)
```
jeffspahr: This is a validation denial, but returning a plain `fmt.Errorf` means the server wrapper still maps the blocked fetch to a 500. Return a `*util.UserError` here so the existing `*url.Error` unwrapping path returns invalid input.

```suggestion
			return nil, util.NewInvalidInputError("connection to blocked IP range denied")
```

backend/src/apiserver/server/pipeline_server_test.go
```diff
@@ -1235,3 +1241,64 @@ func TestCanAccessPipeline_SharedPipeline_ReadAllowed_EvenWhenUnauthorized(t *te
...
+		context.Background(), &api.CreatePipelineRequest{
+			Pipeline: &api.Pipeline{
+				Url:  &api.Url{PipelineUrl: "file:///etc/passwd"},
+				Name: "test-file-blocked",
+			},
+		},
+	)
+	assert.NotNil(t, err)
+	assert.Contains(t, err.Error(), "Pipeline URL validation failed")
+}
```
jeffspahr, line 1304: These API-level tests cover pre-fetch validation. Please also cover the transport/dial validation mapping so a blocked dial cannot regress back to a 500.

```suggestion
}

type errorRoundTripper struct {
	err error
}

func (rt errorRoundTripper) RoundTrip(*http.Request) (*http.Response, error) {
	return nil, rt.err
}

func TestCreatePipelineV1_URLValidation_MapsTransportValidationErrorToInvalidInput(t *testing.T) {
	viper.Set(common.PipelineURLValidationEnabled, "false")

	clientManager := resource.NewFakeClientManagerOrFatal(util.NewFakeTimeForEpoch())
	resourceManager := resource.NewResourceManager(clientManager, &resource.ResourceManagerOptions{CollectMetrics: false})

	pipelineServer := createPipelineServerV1(resourceManager, &http.Client{
		Transport: errorRoundTripper{err: util.NewInvalidInputError("connection to blocked IP range denied")},
	})
	_, err := pipelineServer.CreatePipelineV1(
		context.Background(), &api.CreatePipelineRequest{
			Pipeline: &api.Pipeline{
				Url:  &api.Url{PipelineUrl: "https://storage.googleapis.com/bucket/file.yaml"},
				Name: "test-transport-validation-error",
			},
		},
	)
	assert.NotNil(t, err)
	assert.Equal(t, codes.InvalidArgument, err.(*util.UserError).ExternalStatusCode())
}
```

jeffspahr (commented):
Thanks for updating this. The original DNS rebinding concern is substantially addressed by moving validation into the dial path, and the redirect user-error path is improved. I left inline suggestions for the remaining pieces: preserving default transport/proxy behavior, returning invalid input for dial-time validation denials, and adding regression coverage for the transport-error mapping.

Verification I ran on head `b82cebd706c15370801d34595f455980ce9da202`:

- `go test ./backend/src/apiserver/validation ./backend/src/apiserver/server`
- `git diff --check origin/master...origin/pr/12750`

JerT33:
@jeffspahr all suggestions implemented! Thanks again for all the great feedback

backend/src/apiserver/validation/pipeline_url.go
```diff
@@ -0,0 +1,254 @@
...
+
+func SafePipelineHTTPClient() *http.Client {
+	timeoutSeconds := common.GetIntConfigWithDefault(common.PipelineURLTimeout, 30)
+	if timeoutSeconds < minimumHTTPTimeout {
+		timeoutSeconds = 30
+	}
+	timeout := time.Duration(timeoutSeconds) * time.Second
+
+	transport := http.DefaultTransport.(*http.Transport).Clone()
+	transport.DialContext = safeDialContext
```
jeffspahr, line 209: [P1] This now preserves `ProxyFromEnvironment`, but every dial still goes through `safeDialContext`. In proxy-enabled deployments the API server sets `HTTPS_PROXY=http://squid.squid.svc.cluster.local:3128`; `http.Transport` dials that proxy host, not the final pipeline URL host. Because the proxy service resolves to an in-cluster/private IP, `safeDialContext` rejects the configured proxy itself, so allowed pipeline downloads fail whenever the proxy is used. A focused fix is to bypass blocked-CIDR checks only for the configured proxy host:port, while keeping the safe dialer for non-proxy dials and retaining target URL allowlist/redirect validation.

```suggestion
	transport := http.DefaultTransport.(*http.Transport).Clone()
	transport.DialContext = safeDialContextAllowingConfiguredProxy
```

Then add the helper below this function. It should compare `addr` against the parsed `HTTP_PROXY` / `HTTPS_PROXY` endpoint(s), use a plain `net.Dialer` only for that configured proxy endpoint, and delegate all other dials to `safeDialContext`.

jeffspahr (commented):
Re-reviewed head `4d31449b4dee98eb347a6244d56424c52449999a`. The previous comments are mostly addressed: default transport behavior is now cloned, dial-time blocked-IP denials return invalid input, and an API-level transport-error mapping test was added.

I left one remaining inline concern around proxy-enabled API server deployments. The current failing CI job looks unrelated to this patch: it failed while downloading `github.com/onsi/ginkgo/v2` from `proxy.golang.org`, then report generation failed because no JUnit XML was produced.

Verification I ran:

- `go test ./backend/src/apiserver/validation ./backend/src/apiserver/server`
- `git diff --check origin/master...origin/pr/12750`

backend/src/apiserver/validation/pipeline_url.go
```diff
@@ -0,0 +1,295 @@
...
+
+package validation
+
+import (
+	"context"
+	"fmt"
+	"net"
+	"net/http"
+	"net/url"
+	"os"
```
jeffspahr: This import is only needed by the proxy-bypass helper. With the safer direct-dial fix below, remove it.

```suggestion
```

backend/src/apiserver/validation/pipeline_url.go
```diff
@@ -0,0 +1,295 @@
...
+
+func SafePipelineHTTPClient() *http.Client {
+	timeoutSeconds := common.GetIntConfigWithDefault(common.PipelineURLTimeout, 30)
+	if timeoutSeconds < minimumHTTPTimeout {
+		timeoutSeconds = 30
+	}
+	timeout := time.Duration(timeoutSeconds) * time.Second
+
+	transport := http.DefaultTransport.(*http.Transport).Clone()
+	transport.DialContext = safeDialContextAllowingConfiguredProxy()
```
jeffspahr: For this SSRF-protected fetch path, don't use `ProxyFromEnvironment`. When a proxy is active the apiserver dials only the proxy, so the final target's connection-time IP validation is skipped. Disable proxying for this client and keep direct validated dials.

```suggestion
	transport := http.DefaultTransport.(*http.Transport).Clone()
	transport.Proxy = nil
	transport.DialContext = safeDialContext
```

backend/src/apiserver/validation/pipeline_url.go
```diff
@@ -0,0 +1,295 @@
...
+		port := parsed.Port()
+		if port == "" {
+			port = defaultPorts[parsed.Scheme]
+		}
+		if host != "" && port != "" {
+			addrs[net.JoinHostPort(host, port)] = true
+		}
+	}
+	return addrs
+}
```
jeffspahr: This helper restores connectivity to the configured proxy, but it also means the apiserver no longer performs connection-time validation of the final target host. Remove the proxy bypass path for this security-sensitive client.

```suggestion
```

backend/src/apiserver/validation/pipeline_url_test.go
```diff
@@ -0,0 +1,358 @@
...
+	assert.Equal(t, 0, len(addrs))
+}
+
+func TestConfiguredProxyAddrs_DefaultsPort(t *testing.T) {
+	t.Setenv("HTTPS_PROXY", "http://proxy.local")
+	t.Setenv("HTTP_PROXY", "")
+
+	addrs := configuredProxyAddrs()
+	assert.True(t, addrs["proxy.local:80"])
+}
```
jeffspahr: Replace the proxy address parsing tests with coverage for the intended security behavior: this client should ignore proxy environment variables so the safe dialer still validates the pipeline URL target at connection time.

```suggestion
func TestSafePipelineHTTPClient_DisablesProxyFromEnvironment(t *testing.T) {
	t.Setenv("HTTPS_PROXY", "http://squid.squid.svc.cluster.local:3128")
	t.Setenv("HTTP_PROXY", "http://squid.squid.svc.cluster.local:3128")
	viper.Reset()
	resetURLConfig()

	client := SafePipelineHTTPClient()
	transport, ok := client.Transport.(*http.Transport)
	assert.True(t, ok)
	assert.Nil(t, transport.Proxy)
}
```

jeffspahr (commented):
Follow-up correction from my previous review: the current proxy exemption follows the earlier suggestion, but that suggestion was incomplete for the SSRF threat model. In proxy mode the apiserver only dials Squid, so connection-time validation no longer applies to the final pipeline URL target. The minimal safe fix is to make this SSRF-protected client ignore environment proxies and keep direct, validated dials.

I left inline suggestions for that path: disable proxy use on the cloned transport, remove the proxy-bypass helper, remove the now-unused import, and replace the proxy parsing tests with coverage that proves this client ignores proxy env vars.

If proxy support is required for pipeline URL downloads, that should be a separate design where the proxy enforces the same allowlist and blocked-CIDR policy, or where the apiserver pins the already-validated resolved target IP while preserving the original host semantics.

jeffspahr:
No remaining findings on latest head `9e82c219e902da7ccf049aaba26bb10341538e7d`.

The corrected direct-dial fix is in place: `SafePipelineHTTPClient` disables proxy use, sets `DialContext = safeDialContext`, removes the proxy bypass helper, and includes coverage that `HTTP_PROXY` / `HTTPS_PROXY` are ignored. That closes the proxy-mode gap where target host connection-time validation was skipped.

Verified:
- `go test ./backend/src/apiserver/validation ./backend/src/apiserver/server`
- `git diff --check origin/master...origin/pr/12750`

CI has no failed checks reported at review time; only `tide` is pending.

/lgtm

PR #12751, opened 2026-01-31
fix(cache): include artifact URIs in cache key to prevent incorrect r…

Aman-Cool:
# Fix Cache Key Collisions for Input Artifacts with Different URIs

## Summary
This change fixes an issue in KFP v2 cache key generation where only input artifact **logical names** were considered.
As a result, tasks could incorrectly reuse cached outputs even when they were executed with **different input data locations (URIs)**.

## Problem
Cache keys were generated using artifact names only, ignoring the artifact URI that identifies the actual data source.
If the same pipeline was run multiple times with the same artifact name but different URIs, the cache would treat them as identical and return previously cached results.

This caused silent and incorrect cache hits when:
- Re-running pipelines with new dataset versions
- Sharing pipeline templates across teams or tenants
- Running experiments or A/B tests on different input data

## Fix
The cache key now includes the artifact URI alongside the name (`name@uri`) when generating the cache identifier.
This ensures cache hits occur only when input artifacts truly refer to the same data.

Artifacts without a URI continue to fall back to name-only behavior for backward compatibility.

## Tests
Existing cache tests were updated, and new tests were added to validate that:
- Same artifact name + different URI → different cache fingerprints
- Same name + same URI → cache hit
- Artifacts without URIs remain supported

## Behavior Change
Previously cached executions will no longer match under the new cache key format, resulting in cache misses.
This is expected and correct, as it prevents reuse of outputs computed from different inputs.

## Result
- Correct cache behavior for dataset versioning and multi-tenant usage
- No more silent reuse of outputs from different data sources
- Deterministic and trustworthy caching semantics

### DIFF
--- a/api/v2alpha1/cache_key.proto
+++ b/api/v2alpha1/cache_key.proto
@@ -21,6 +21,10 @@ import "google/protobuf/struct.proto";
 import "pipeline_spec.proto";
 
 message CacheKey {
+  // Input artifact identifiers. Each string in ArtifactNameList now uses the format
+  // "name@uri" to include both the artifact name and URI, ensuring different data
+  // sources produce different cache keys. Previously only names were used, which
+  // caused incorrect cache reuse when artifacts had the same name but different URIs.
   map<string, ArtifactNameList> inputArtifactNames = 1;
   map<string, Value> inputParameters = 2 [deprecated = true];
   map<string, RuntimeArtifact> outputArtifactsSpec = 3;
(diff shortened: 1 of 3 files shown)


Aman-Cool:
@mprahl @droctothorpe ,Fixes incorrect cache reuse by including input artifact URIs in cache keys; existing cached runs may miss once, as expected.

zazulam (commented):
@Aman-Cool Thanks for this contribution, but can you remove the comments from the proto file? I think just documenting it in `cache.go` would be fine.

Aman-Cool:
@zazulam , Removed the comments from the proto file and kept the documentation in `cache.go` only, as suggested.

zazulam:
@Aman-Cool can you run the repo's pre-commit hooks?

Aman-Cool:
@zazulam ,Pre-commit hooks have been run locally and the resulting fixes are pushed.

zazulam:
@Aman-Cool can you rebase your PR with master

PR #12735, opened 2026-01-29
fix(testing) retention for image-builds-with-cache artifacts

alyssacgoins:
**Description of your changes:**
Add the current date to the cache key used in `.github/workflows/image-builds-with-cache.yml`. This forces the cache key to change daily, so that old image markers disappear after midnight, syncing cache expiration with single-day artifact retention.This change should fix the CI bug in which image artifacts fail to be located.

This PR is a follow-up to the cache lookups added in #12712. 
Resolves #12734 

**Checklist:**
- [X] You have [signed off your commits](https://www.kubeflow.org/docs/about/contributing/#sign-off-your-commits)
- [X] The title for your pull request (PR) should follow our title convention. [Learn more about the pull request title convention used in this repository](https://github.com/kubeflow/pipelines/blob/master/CONTRIBUTING.md#pull-request-title-convention).
- [] Make sure the changes work with DSPO. Run the tests in DSPO with DSPA images pointing to this PR's images

PR #12763, opened 2026-02-03
Fix: cache image tarballs for PR builds

sduvvuri1603:
### Summary

- Cache/restore image tarballs on PR runs so cache hits still produce artifacts.
- Keep deploy strict while preventing missing image tarball failures.

### Why this change is necessary

- Current master uses a build‑marker cache on PRs: when it hits, the build is skipped.
- The image tarball is created only during the build step, so on a cache hit no tarball exists.
- Deploy now requires every tarball (e.g., metadata-writer) and fails with:
     `open .../metadata-writer/metadata-writer.tar: no such file or directory`
- This is a reliability gap: cache hits can cause missing artifacts even when the image “exists” logically.

### Fix approach

- Cache the actual tarball instead of a marker.
- On cache hit, restore the tarball and still upload it as an artifact.
- This preserves CI speed while guaranteeing artifacts for deploy.

### DIFF
--- a/.github/workflows/image-builds-with-cache.yml
+++ b/.github/workflows/image-builds-with-cache.yml
@@ -86,14 +86,14 @@ jobs:
         id: date
         run: echo "date=$(date +'%Y-%m-%d')" >> "$GITHUB_OUTPUT"
 
-      # Check if the image was already built for this SHA using cache
-      - name: Check if already built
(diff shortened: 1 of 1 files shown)


sduvvuri1603:
Duplicate PR. Issue Resolved in another PR: https://github.com/kubeflow/pipelines/pull/12735

PR #12764, opened 2026-02-03
ci: add pr number to the cache key

nsingla:
Adding pr number to the cache key, hopefully this resolves the issue with cross pr contamination when master is merged in prs and commits are not squashed, then the sha of the last commit is the master sha, or when different prs are not rebased correctly and the sha matches. For e.g. this failure:
https://github.com/kubeflow/pipelines/actions/runs/21638591850/job/62376928389?pr=12442

**Checklist:**

### DIFF
--- a/.github/workflows/image-builds-with-cache.yml
+++ b/.github/workflows/image-builds-with-cache.yml
@@ -93,7 +93,7 @@ jobs:
         uses: actions/cache@v5
         with:
           path: .build-marker-${{ matrix.image }}
-          key: image-built-${{ matrix.image }}-${{ github.sha }}-${{ steps.date.outputs.date }}
+          key: image-built-${{ matrix.image }}-${{ github.sha }}-${{ github.event.pull_request.number || github.ref_name }}-${{ steps.date.outputs.date }}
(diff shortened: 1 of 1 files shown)


sduvvuri1603:
Thanks for the fix! @nsingla 

This is for my clarification: This does help with cross‑PR cache contamination but what happens if I do a rerun of the same PR on the same day?
The cache marker still hits, the build is skipped, and no artifacts are uploaded for that rerun. The deploy job only downloads artifacts from the current run, so it can still fail with `metadata-writer.tar` not found right?

nsingla:
> Thanks for the fix! @nsingla
> 
> This is for my clarification: This does help with cross‑PR cache contamination but what happens if I do a rerun of the same PR on the same day? The cache marker still hits, the build is skipped, and no artifacts are uploaded for that rerun. The deploy job only downloads artifacts from the current run, so it can still fail with `metadata-writer.tar` not found right?

I think for the rerun it should still work, if the artifact is found in the cache, then it should be available (based on the key which includes date, pr number and sha). I think the issue was only due to cross contamination

PR #12577, opened 2025-12-23
feat(kfp-kubernetes): Add securityContext support for container components (Fixes #12560)

AnupamSingh2004:
## What is the problem this PR solves?

This PR implements the ability to set `securityContext` on container component pods, as requested in #12560.

Currently, there is no way to configure security context settings (such as `privileged`, `runAsUser`, `capabilities`, etc.) for pipeline tasks. Users running workloads that require specific security configurations (e.g., GPU workloads, containers needing specific capabilities, or hardened security policies) cannot configure these settings through the KFP SDK.

## How does this PR solve the problem?

This PR adds a new `set_security_context()` function to the `kfp-kubernetes` SDK extension that allows users to configure container-level security context settings. The implementation includes:

1. **Proto schema changes**: Added `SecurityContext`, `Capabilities`, `SELinuxOptions`, and `SeccompProfile` messages to `kubernetes_executor_config.proto`
2. **Python SDK**: Added `set_security_context()` function with validation and comprehensive documentation
3. **Go backend**: Implemented security context handling in the driver to apply settings to pod specs
4. **Tests**: Added unit tests for both Python SDK (18 tests) and Go backend (15 tests)

### Supported Security Context Fields

| Field | Description |
|-------|-------------|
| `privileged` | Run container in privileged mode |
| `allow_privilege_escalation` | Control whether a process can gain more privileges |
| `run_as_user` | UID to run the container process |
| `run_as_group` | GID to run the container process |
| `run_as_non_root` | Require container to run as non-root |
| `read_only_root_filesystem` | Mount root filesystem as read-only |
| `capabilities_add` | Linux capabilities to add |
| `capabilities_drop` | Linux capabilities to drop |
| `se_linux_options_*` | SELinux context labels |
| `seccomp_profile_type` | Seccomp profile type (RuntimeDefault/Unconfined/Localhost) |
| `seccomp_profile_localhost_profile` | Path to localhost seccomp profile |

## Example Usage

```python
from kfp import dsl
from kfp import kubernetes

@dsl.component
def my_component():
    pass

@dsl.pipeline
def privileged_pipeline():
    """Pipeline with privileged container."""
    task = my_component()
    kubernetes.set_security_context(
        task,
        privileged=True,
    )

@dsl.pipeline
def hardened_pipeline():
    """Pipeline with security hardening."""
    task = my_component()
    kubernetes.set_security_context(
        task,
        run_as_non_root=True,
        run_as_user=1000,
        run_as_group=1000,
[...]

Issue #12205, opened 2025-09-02
[backend] Cannot set the explicit securityContext capabilities like Runasuser/Runasgroup


vivekrsintc:
### Environment

*  How did you deploy Kubeflow Pipelines (KFP)?
Im using SDK it generates the spec where I could not set the security context privlige

https://www.kubeflow.org/docs/pipelines/installation/overview/. -->
*  KFP version: 
kfp                      2.14.3
kfp-pipeline-spec        2.14.0
kfp-server-api           2.14.0
*  KFP SDK version: 
kfp                      2.14.3
kfp-pipeline-spec        2.14.0
kfp-server-api           2.14.0

### Steps to reproduce

Try to add securityContext options
### Expected result

User should give same option as configure like k8s capabilities. at the end its going to deploy on kubernetes resource

### Materials and Reference

---

Impacted by this bug? Give it a 👍.

Prateekbala:
Hi! I'd like to work on this issue. Could you assign it to me?

juliusvonkohout:
> Hi! I'd like to work on this issue. Could you assign it to me?

Done, but you might need help from @droctothorpe and @zazulam

juliusvonkohout:
@droctothorpe @HumairAK @hbelmiro do you have any timeline for this feature? I think it is critical to sunset KFP v1.

We need a tight securitycontext such as 

```
    securityContext:
      runAsNonRoot: true
      allowPrivilegeEscalation: false
      capabilities:
        drop:
        - ALL
      seccompProfile:
        type: RuntimeDefault
```

Prateekbala:
Hey @juliusvonkohout @droctothorpe 

Before I start implementing, I wanted to clarify one thing - should the security context be applied at the container level, pod level, or should users be able to choose both?

Thanks!

juliusvonkohout:
> Hey [@juliusvonkohout](https://github.com/juliusvonkohout) [@droctothorpe](https://github.com/droctothorpe)
> 
> Before I start implementing, I wanted to clarify one thing - should the security context be applied at the container level, pod level, or should users be able to choose both?
> 
> Thanks!

Users should be able to set both. Some things can only be set at the pod level and some at the container level. so something like set_container_security_context() and set_pod_security_context(). set_container_security_context()  is more important and the first one you should tackle.

The default for the container should be 
```
    securityContext:
      runAsNonRoot: true
      allowPrivilegeEscalation: false
      capabilities:
        drop:
        - ALL
[...]

PR #12782, opened 2026-02-07
feat(security): enforce pod and container security contexts

jsonmp-k8:
## Summary
- Enforce default pod- and container-level security contexts across KFP v2 compiler templates (driver, executor, importer, DAG driver) and workflow spec.
- Harden control-plane manifests and GCP Marketplace chart with runAsNonRoot + RuntimeDefault seccomp, and container allowPrivilegeEscalation=false with dropped ALL capabilities.
- Update test workflows and compiled workflow goldens to include the new defaults.

## Testing
- go test ./backend/src/v2/compiler/argocompiler/...
- go test ./backend/test/compiler/...

### DIFF
--- a/backend/src/v2/compiler/argocompiler/security_context_test.go
+++ b/backend/src/v2/compiler/argocompiler/security_context_test.go
@@ -126,6 +126,7 @@ func TestApplySecurityContextToTemplate_WithContainer(t *testing.T) {
 	assert.NotNil(t, template.Container.SecurityContext)
 	assert.False(t, *template.Container.SecurityContext.AllowPrivilegeEscalation)
 	assert.True(t, *template.Container.SecurityContext.RunAsNonRoot)
+	assert.Equal(t, []k8score.Capability{"ALL"}, template.Container.SecurityContext.Capabilities.Drop)
 }
 
 func TestApplySecurityContextToTemplate_WithInitContainersAndSidecars(t *testing.T) {
@@ -146,6 +147,7 @@ func TestApplySecurityContextToTemplate_WithInitContainersAndSidecars(t *testing
 		assert.NotNil(t, initContainer.SecurityContext, "init container %s should have security context", initContainer.Name)
 		assert.False(t, *initContainer.SecurityContext.AllowPrivilegeEscalation)
 		assert.True(t, *initContainer.SecurityContext.RunAsNonRoot)
+		assert.Equal(t, []k8score.Capability{"ALL"}, initContainer.SecurityContext.Capabilities.Drop)
 	}
 
 	// All sidecars should have security context
@@ -153,6 +155,7 @@ func TestApplySecurityContextToTemplate_WithInitContainersAndSidecars(t *testing
 		assert.NotNil(t, sidecar.SecurityContext, "sidecar %s should have security context", sidecar.Name)
 		assert.False(t, *sidecar.SecurityContext.AllowPrivilegeEscalation)
 		assert.True(t, *sidecar.SecurityContext.RunAsNonRoot)
+		assert.Equal(t, []k8score.Capability{"ALL"}, sidecar.SecurityContext.Capabilities.Drop)
 	}
 }
 
@@ -283,6 +286,7 @@ func TestApplySecurityContextToExecutorTemplate_WithContainerAndInitContainers(t
 		assert.NotNil(t, initContainer.SecurityContext)
 		assert.True(t, *initContainer.SecurityContext.RunAsNonRoot)
 		assert.False(t, *initContainer.SecurityContext.AllowPrivilegeEscalation)
+		assert.Equal(t, []k8score.Capability{"ALL"}, initContainer.SecurityContext.Capabilities.Drop)
 	}
 
 	// Sidecars: no RunAsNonRoot, but capabilities drop ALL


juliusvonkohout:
Thank you for the PR, How does this align with https://github.com/kubeflow/pipelines/pull/12577 and the corresponding Issue https://github.com/kubeflow/pipelines/issues/12205 ?

juliusvonkohout:
Please always use rebase instead of merge from master. Merge from master destroys the commit history.

droctothorpe:
## PR Review: feat(security): enforce pod and container security contexts

### Overall Assessment

Well-structured security hardening PR. The approach of differentiating between system templates (driver/DAG driver/importer) and executor templates (user containers) is the right design - it avoids breaking user pipelines that use root-based images like `python:3.12`. The unit tests are thorough and cover the key edge cases.

### Design

- The split between `applySecurityContextToTemplate` (system containers - full hardening including `runAsNonRoot`) and `applySecurityContextToExecutorTemplate` (user containers - no `runAsNonRoot`, no capability drop) is well-reasoned and clearly documented in the code comments.
- The Argo wait sidecar consideration (setting `seccompProfile` at pod level but `runAsNonRoot` only at container level) shows good awareness of the Argo Workflows architecture.
- All functions are nil-safe and fill-only (never overwrite user-specified values). This is the correct behavior for security defaults.

### Issues

1. **`metadata-grpc-deployment.yaml` inconsistency**: This file already has `seccompProfile: RuntimeDefault` at pod level and full container-level hardening, but is **missing `runAsNonRoot: true` at the pod level** - unlike every other manifest in this PR. The PR only removes a trailing blank line from this file. This should be fixed for consistency.

2. **`metadata-envoy-deployment.yaml` already has container-level security context**: The existing file already has a complete container-level `securityContext` (lines 27-36 with `allowPrivilegeEscalation: false`, `seccompProfile`, `runAsNonRoot`, `runAsUser`, `runAsGroup`, caps drop). The PR correctly adds the pod-level security context, but it's worth double-checking this was intentional and not an oversight where you meant to also add something at container level.

3. **`metacontroller` StatefulSet**: The PR adds pod-level `runAsNonRoot` + `seccompProfile`, but the container already has comprehensive security context including `readOnlyRootFilesystem: true` and `privileged: false`. Good that the PR doesn't touch the container level here. However, there's no `allowPrivilegeEscalation: false` being added at container level, but this is already present in the existing manifest so this is fine.

4. **Duplicate `seccompProfile` at pod and container levels**: In `argo.go`, `seccompProfile` is set at the workflow spec level (pod), but `applySecurityContextToTemplate` also sets `seccompProfile` at both pod and container levels for each template. The pod-level seccomp in `argo.go` would apply to all pods, so the per-template pod-level seccomp is redundant (but not harmful). Consider whether the workflow-level setting in `argo.go` is sufficient, or add a comment explaining why both are needed for defense-in-depth.

5. **Workflow-level security context in `argo.go` is inconsistent with template-level**: In `argo.go` lines 153-158, the workflow-level `PodSecurityContext` only sets `SeccompProfile` (and optionally `RunAsUser`), while `defaultPodSecurityContext()` in `security_context.go` includes both `RunAsNonRoot` and `SeccompProfile`. The comment in `applySecurityContextToTemplate` explains why `RunAsNonRoot` is omitted at pod level (Argo wait sidecar), but the workflow-level context should also follow this reasoning. Currently it's correct behavior but undocumented at the `argo.go` callsite.

### Minor / Nits

- **`test/build_image.yaml` line 1158**: The `securityContext` block is placed at the same indent level as `container` properties like `args`, but it should be a sibling of `image`, `imagePullPolicy`, etc. It currently appears at the `container` child level which is correct YAML semantically, but the placement before `args` looks like it could be a readability issue (other test files in this PR place it right after `container:` or `image:`).

- **Compiled workflow goldens**: Only `run_as_user_cache_disabled.yaml` and `run_as_user_cache_enabled.yaml` are updated. If other golden files exist that don't include security contexts, the compiler tests may need `-- -updateCompiledFiles=true` to regenerate them. Verify that `ginkgo -v ./backend/test/compiler` passes without the update flag.

- **Dockerfile UID 65532**: Good choice - this is the well-known `nonroot` UID from distroless images, making it a recognized convention. The change from `USER appuser` (name) to `USER 65532` (numeric) is also better for Kubernetes since `runAsNonRoot` checks the numeric UID, not the username.

### Test Coverage

The unit tests in `security_context_test.go` are comprehensive:
- Nil safety for all entry points
- Partial context preservation (existing values not overwritten)
- Explicit `false` values for `RunAsNonRoot` are preserved
- Custom `SeccompProfile` types are preserved
- Executor template behavior correctly validated (no `runAsNonRoot` on user container)

The test matcher changes in `workflow_matcher.go` are a pragmatic approach - switching from exact equality to field-by-field matching for security contexts lets the golden files stay focused on pipeline-specific fields without having to enumerate every security default.

droctothorpe:
Experimenting with https://github.com/opencode-ai/opencode for the first time and passed this PR through it for review. Take it with a grain of salt.

jsonmp-k8:
Thanks for the thorough review! All issues have been addressed in 27feb4399:

1. **metadata-grpc-deployment.yaml**: Added missing `runAsNonRoot: true` at pod level.
2. **metadata-envoy-deployment.yaml**: Confirmed intentional — pod-level context was the only missing piece. No change needed.
3. **metacontroller StatefulSet**: Confirmed correct as-is. No change needed.
4. **Duplicate seccompProfile**: Added comment in `argo.go` explaining defense-in-depth rationale.
5. **Workflow-level RunAsNonRoot omission**: Added comment in `argo.go` documenting the Argo wait sidecar constraint. Also removed the unused `defaultPodSecurityContext`/`applyPodSecurityContext` functions and their tests.

backend/test/compiler/matchers/workflow_matcher.go
```diff
@@ -266,3 +266,62 @@ func AreStringsSameWithoutOrder(s1, s2 string) bool {
 	// Compare the sorted slices
 	return reflect.DeepEqual(r1, r2)
 }
+
+func matchPodSecurityContext(actual *v1.PodSecurityContext, expected *v1.PodSecurityContext, msg string) {
+	if expected == nil {
+		return
+	}
```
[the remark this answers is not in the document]
jsonmp-k8: The relaxed matching is intentional. The compiler now injects security context defaults into every template, but ~120+ golden files don't include security contexts. Asserting `actual == nil` when `expected == nil` would break all of them, requiring a massive golden regeneration that would obscure the actual security changes.

The design separates concerns: goldens only specify fields they care about (e.g. `runAsUser`), while security context defaults are validated by dedicated unit tests in `security_context_test.go`. This keeps goldens focused and maintainable.

test_data/compiled-workflows/run_as_user_cache_enabled.yaml
```diff
@@ -20,6 +20,8 @@ spec:
       pipelines.kubeflow.org/v2_component: "true"
   securityContext:
     runAsUser: 1001
+    seccompProfile:
+      type: RuntimeDefault
```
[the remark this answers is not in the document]
jsonmp-k8: These goldens exist to test `runAsUser` behavior, not to validate security context defaults. The field-by-field matcher intentionally skips fields not present in the golden — security context defaults are covered by unit tests in `security_context_test.go`. Adding full security contexts here would make the goldens brittle and require updating every time a security default changes.

test_data/compiled-workflows/run_as_user_cache_disabled.yaml
```diff
@@ -20,6 +20,8 @@ spec:
       pipelines.kubeflow.org/v2_component: "true"
   securityContext:
     runAsUser: 1001
+    seccompProfile:
+      type: RuntimeDefault
```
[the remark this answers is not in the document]
jsonmp-k8: Same as above — this golden tests `runAsUser` + cache disabled behavior. Security context defaults are validated separately in `security_context_test.go`.

backend/src/v2/compiler/argocompiler/security_context.go
```diff
@@ -0,0 +1,145 @@
...
+package argocompiler
+
+import (
+	wfapi "github.com/argoproj/argo-workflows/v3/pkg/apis/workflow/v1alpha1"
+	k8score "k8s.io/api/core/v1"
+)
+
+func defaultContainerSecurityContext() *k8score.SecurityContext {
+	allowPrivilegeEscalation := false
+	runAsNonRoot := true
```
droctothorpe, line 24: Could you make these two values overridable through environment variables? I want to make sure that if a platform engineer wants to override them for any reason, they don't have to update the image.

jsonmp-k8: Done in a64e6142a. Added two environment variables:

- `PIPELINE_SECURITY_ALLOW_PRIVILEGE_ESCALATION` (default: `"false"`)
- `PIPELINE_SECURITY_RUN_AS_NON_ROOT` (default: `"true"`)

These are read via `getBoolEnvWithDefault()` and used in both `defaultContainerSecurityContext()` and the fill-only path in `applySecurityContextToContainer()`. Invalid values log an error and fall back to the defaults. Added tests for both the override and invalid-value cases.

juliusvonkohout:
The pod level sceccompprofile was needed, because of shortcomings in istio sidecars. That is now fixed in istio upstream

backend/src/v2/compiler/argocompiler/argo.go
```diff
@@ -150,9 +150,20 @@ func Compile(jobArg *pipelinespec.PipelineJob, kubernetesSpecArg *pipelinespec.S
 		},
 	}
 
+	// Set SeccompProfile at the workflow level as a baseline default for all pods.
+	// RunAsNonRoot is intentionally omitted here because the Argo Workflows
+	// "wait" sidecar (argoexec) runs as root; pod-level RunAsNonRoot would
```
juliusvonkohout: `Argo Workflows "wait" sidecar (argoexec) runs as root` that should not be the case and must be changed.

juliusvonkohout: to verify /guarantee this you have to add the PSS restricted label to the namespace of KFP

backend/src/v2/compiler/argocompiler/security_context.go
```diff
@@ -0,0 +1,172 @@
...
+	"github.com/golang/glog"
+
+	wfapi "github.com/argoproj/argo-workflows/v3/pkg/apis/workflow/v1alpha1"
+	k8score "k8s.io/api/core/v1"
+)
+
+const (
+	// AllowPrivilegeEscalationEnvVar overrides the default allowPrivilegeEscalation
+	// value for system container security contexts. Default: "false".
+	AllowPrivilegeEscalationEnvVar = "PIPELINE_SECURITY_ALLOW_PRIVILEGE_ESCALATION"
```
juliusvonkohout: Privilege escalation should never be allowed under no circumstances. That not just breaks PSS restricted, but even PSS baseline.

backend/src/v2/compiler/argocompiler/security_context.go
```diff
@@ -0,0 +1,172 @@
...
+	// RunAsNonRootEnvVar overrides the default runAsNonRoot value for system
+	// container security contexts. Default: "true".
+	RunAsNonRootEnvVar = "PIPELINE_SECURITY_RUN_AS_NON_ROOT"
```
juliusvonkohout: All system containers must always runasnonroot. There must be no exception. For customer workloads we could allow deviations.

backend/src/v2/compiler/argocompiler/security_context.go
```diff
@@ -0,0 +1,172 @@
...
+		},
+	}
+}
+
+func applySecurityContextToTemplate(t *wfapi.Template) {
+	if t == nil {
+		return
+	}
+	// Use SeccompProfile only at the pod level (not RunAsNonRoot) because
+	// Argo Workflows injects a "wait" sidecar (argoexec) that runs as root.
```
juliusvonkohout: IT must not run as root. Argo is explicitly designed to run rootless. https://argo-workflows.readthedocs.io/en/latest/workflow-pod-security-context/ https://argo-workflows.readthedocs.io/en/latest/workflow-pod-security-context/#non-root-executor-image

Image

juliusvonkohout (changes requested):
THhank you for the PR. I added several comments, but most important is that system level stuff never runs as root 

Image

https://argo-workflows.readthedocs.io/en/latest/workflow-pod-security-context/ https://argo-workflows.readthedocs.io/en/latest/workflow-pod-security-context/#non-root-executor-image We should allow exceptions for runasnonroot, runasuser and runasgroup for  the customer workload container, but not Argo. You could also drop the runasnonroot and seccompprofile at the pod level, but there also is no harm in having it. The pod level seccomprofile was needed as workaround for a bug in istio that should now be fixed, but to be safe we can leave that seccomp stuff in that i added a long time ago. Essential is that you set the proper PSS baseline/restricted annotation to the KFP namespace to actually enforce and guarantee the standard that you implement. See also https://kubernetes.io/docs/concepts/security/pod-security-standards/ and https://kubernetes.io/docs/concepts/security/pod-security-admission/ for the namespace level annotation. You can start with PSS baseline for this PR.

jsonmp-k8:
> THhank you for the PR. I added several comments, but most important is that system level stuff never runs as root
> 
> Image
> https://argo-workflows.readthedocs.io/en/latest/workflow-pod-security-context/ https://argo-workflows.readthedocs.io/en/latest/workflow-pod-security-context/#non-root-executor-image We should allow exceptions for runasnonroot, runasuser and runasgroup for the customer workload container, but not Argo. You could also drop the runasnonroot and seccompprofile at the pod level, but there also is no harm in having it. The pod level seccomprofile was needed as workaround for a bug in istio that should now be fixed, but to be safe we can leave that seccomp stuff in that i added a long time ago. Essential is that you set the proper PSS baseline/restricted annotation to the KFP namespace to actually enforce and guarantee the standard that you implement. See also https://kubernetes.io/docs/concepts/security/pod-security-standards/ and https://kubernetes.io/docs/concepts/security/pod-security-admission/ for the namespace level annotation. You can start with PSS baseline for this PR.

@juliusvonkohout Thank you for the thorough review! I've pushed a commit addressing all your feedback. Please let me know if you have further comments. Happy to address the same.

droctothorpe:
@jsonmp-k8 CI is failing FYI.

@juliusvonkohout will standby for an LGTM from you on this since you're our resident manifests expert. Really appreciate your input on this PR.

jsonmp-k8:
> @jsonmp-k8 CI is failing FYI.
> 
> @juliusvonkohout will standby for an LGTM from you on this since you're our resident manifests expert. Really appreciate your input on this PR.

Messed up some of the changes, will update the PR soon

juliusvonkohout:
@jsonmp-k8 please fix the tests.
For example `metadata-grpc-deployment-d8f8fd4dc-2s265          0/1     Init:CreateContainerConfigError   0             12m` is responsible for multiple test failures.

backend/src/v2/compiler/argocompiler/security_context.go
juliusvonkohout: Please remove PIPELINE_SECURITY_ALLOW_PRIVILEGE_ESCALATION that should never ever be allowed since it violates even PSS baseline as the minimal standard, not just PSS restricted.

backend/src/v2/compiler/argocompiler/security_context.go
```diff
@@ -0,0 +1,168 @@
...
+			Drop: []k8score.Capability{"ALL"},
+		},
+		SeccompProfile: &k8score.SeccompProfile{
+			Type: k8score.SeccompProfileTypeRuntimeDefault,
+		},
+	}
+}
+
+func applySecurityContextToTemplate(t *wfapi.Template) {
+	if t == nil {
```
juliusvonkohout:
```suggestion
	if template == nil {
```

please proper variable naming everywhere. Long, verbose, pronounceable and expressive.

juliusvonkohout: In all files

backend/src/v2/compiler/argocompiler/security_context.go
```diff
@@ -0,0 +1,168 @@
...
+	// Apply SeccompProfile and RunAsNonRoot at the pod level.
+	// Argo Workflows is designed to run rootless:
+	// https://argo-workflows.readthedocs.io/en/latest/workflow-pod-security-context/
+	applyPodSecurityDefaults(&t.SecurityContext)
+	applySecurityContextToContainer(t.Container)
+	applySecurityContextToUserContainers(t.InitContainers)
+	applySecurityContextToUserContainers(t.Sidecars)
+}
+
+func applySecurityContextToContainer(c *k8score.Container) {
```
juliusvonkohout:
```suggestion
func applySecurityContextToContainer(container *k8score.Container) {
```

please proper variable naming everywhere. Long, verbose, pronounceable and expressive in all files.

backend/src/v2/compiler/argocompiler/security_context.go
```diff
@@ -0,0 +1,168 @@
...
+	for i := range containers {
+		applySecurityContextToContainer(&containers[i].Container)
+	}
+}
+
+// applySecurityContextToExecutorTemplate applies security defaults to executor
+// templates without enforcing runAsNonRoot on the pod or main container, since
+// user-specified images (e.g. python:3.12) may run as root. Init containers
+// (KFP system images like the launcher) still get the full security context.
+func applySecurityContextToExecutorTemplate(t *wfapi.Template) {
```
juliusvonkohout: please proper variable naming everywhere. Long, verbose, pronounceable and expressive in all files.
```suggestion
func applySecurityContextToExecutorTemplate(template *wfapi.Template) {
```

backend/src/v2/compiler/argocompiler/security_context.go
```diff
@@ -0,0 +1,168 @@
...
+	applyPodSeccompProfileOnly(&t.SecurityContext)
+	applyUserContainerSecurityContext(t.Container)
+	// Init containers use KFP system images (e.g. launcher) that support non-root.
+	applySecurityContextToUserContainers(t.InitContainers)
+	for i := range t.Sidecars {
+		applyUserContainerSecurityContext(&t.Sidecars[i].Container)
+	}
+}
+
+func applyPodSecurityDefaults(sc **k8score.PodSecurityContext) {
```
juliusvonkohout:
```suggestion
func applyPodSecurityDefaults(security_context **k8score.PodSecurityContext) {
```
please proper variable naming everywhere. Long, verbose, pronounceable and expressive in all files.

backend/src/v2/compiler/argocompiler/security_context.go
```diff
@@ -0,0 +1,168 @@
...
+		(*sc).SeccompProfile = &k8score.SeccompProfile{
+			Type: k8score.SeccompProfileTypeRuntimeDefault,
+		}
+	}
+}
+
+// applyPodSeccompProfileOnly sets only the SeccompProfile at the pod level,
+// without RunAsNonRoot. Used for executor templates where user-specified
+// images may run as root.
+func applyPodSeccompProfileOnly(sc **k8score.PodSecurityContext) {
```
juliusvonkohout: please proper variable naming everywhere. Long, verbose, pronounceable and expressive in all files.

backend/src/v2/compiler/argocompiler/security_context_test.go
```diff
@@ -0,0 +1,297 @@
...
+
+	wfapi "github.com/argoproj/argo-workflows/v3/pkg/apis/workflow/v1alpha1"
+	"github.com/stretchr/testify/assert"
+	k8score "k8s.io/api/core/v1"
+)
+
+func boolPtr(b bool) *bool    { return &b }
+func int64Ptr(i int64) *int64 { return &i }
+
+func TestDefaultContainerSecurityContext(t *testing.T) {
```
juliusvonkohout, line 28: please proper variable naming everywhere. Long, verbose, pronounceable and expressive in all files.

backend/src/v2/compiler/argocompiler/security_context_test.go
```diff
@@ -0,0 +1,297 @@
...
+	assert.NotNil(t, sc.SeccompProfile)
+	assert.Equal(t, k8score.SeccompProfileTypeRuntimeDefault, sc.SeccompProfile.Type)
+	// Should not set fields that are left to the user
+	assert.Nil(t, sc.RunAsUser)
+	assert.Nil(t, sc.RunAsGroup)
+	assert.Nil(t, sc.Privileged)
+	assert.Nil(t, sc.ReadOnlyRootFilesystem)
+}
+
+func TestApplySecurityContextToContainer_NilContainer(t *testing.T) {
```
juliusvonkohout, line 47: please proper variable naming everywhere. Long, verbose, pronounceable and expressive in all files.

backend/src/v2/compiler/argocompiler/security_context_test.go
```diff
@@ -0,0 +1,297 @@
...
+	// Should not panic
+	applySecurityContextToContainer(nil)
+}
+
+func TestApplySecurityContextToContainer_NilSecurityContext(t *testing.T) {
```
juliusvonkohout, line 52: please proper variable naming everywhere. Long, verbose, pronounceable and expressive in all files.

backend/src/v2/compiler/argocompiler/security_context_test.go
```diff
@@ -0,0 +1,297 @@
...
+	// Should preserve existing fields
+	assert.Equal(t, int64(1000), *c.SecurityContext.RunAsUser)
+	// Should fill in missing defaults
+	assert.False(t, *c.SecurityContext.AllowPrivilegeEscalation)
+	assert.True(t, *c.SecurityContext.RunAsNonRoot)
+	assert.Equal(t, []k8score.Capability{"ALL"}, c.SecurityContext.Capabilities.Drop)
+	assert.Equal(t, k8score.SeccompProfileTypeRuntimeDefault, c.SecurityContext.SeccompProfile.Type)
+}
+
+func TestApplySecurityContextToContainer_FullSecurityContext(t *testing.T) {
```
juliusvonkohout, line 81: please proper variable naming everywhere. Long, verbose, pronounceable and expressive in all files.

juliusvonkohout (changes requested):
please proper variable naming everywhere. Long, verbose, pronounceable and expressive in all files. t -> template, sc -> securitycontext, c -> container etc.

please fix the tests.
For example metadata-grpc-deployment-d8f8fd4dc-2s265          0/1     Init:CreateContainerConfigError   0             12m is responsible for multiple test failures.

jsonmp-k8:
@juliusvonkohout Thanks for the review feedback! Addressed in the latest commits:

1. **`CreateContainerConfigError` in TLS tests** (5b7014b): The `metadata-grpc-deployment` TLS overlay had an init container (`generate-config`) without a security context. With pod-level `runAsNonRoot: true`, it failed validation. Added a proper security context matching the base deployment (`runAsUser: 1000`).

2. **Variable naming** (144eac1): Renamed all short variable names to long, pronounceable and expressive names across `security_context.go` and `security_context_test.go`:
   - `t` → `template`
   - `c` → `container`
   - `sc` → `podSecurityContext` / `securityContext` / `sidecar` (context-dependent)
   - `ic` → `initContainer`
   - `tmpl` → `template`

backend/src/v2/compiler/argocompiler/container.go
```diff
@@ -265,6 +265,7 @@ func (c *workflowCompiler) addContainerDriverTemplate() string {
 			Env:       append(proxy.GetConfig().GetEnvVars(), commonEnvs...),
 		},
 	}
+	applySecurityContextToTemplate(t)
```
juliusvonkohout:
```suggestion
	applySecurityContextToTemplate(template)
```

backend/src/v2/compiler/argocompiler/dag.go
```diff
@@ -630,6 +630,7 @@ func (c *workflowCompiler) addDAGDriverTemplate() string {
 			Env:       proxy.GetConfig().GetEnvVars(),
 		},
 	}
+	applySecurityContextToTemplate(t)
```
juliusvonkohout:
```suggestion
	applySecurityContextToTemplate(template)
```

backend/src/v2/compiler/argocompiler/security_context.go
juliusvonkohout: This was not addressed yet

juliusvonkohout:
Please check again whether you really addressed and marked as resolved all comments and fixed the naming. See for example https://github.com/kubeflow/pipelines/pull/12782#discussion_r2804764164 and https://github.com/kubeflow/pipelines/pull/12782#discussion_r2798192034. All comments must be resolved. The tests look good so far.

backend/src/v2/compiler/argocompiler/container.go
jsonmp-k8: Done in 4214088 — renamed `t` to `template` in `addContainerDriverTemplate()`.

backend/src/v2/compiler/argocompiler/dag.go
jsonmp-k8: Done in 4214088 — renamed `t` to `template` in `addDAGDriverTemplate()`.

backend/src/v2/compiler/argocompiler/security_context.go
jsonmp-k8: .

backend/src/v2/compiler/argocompiler/security_context.go
```diff
@@ -0,0 +1,168 @@
...
+
+func defaultUserContainerSecurityContext() *k8score.SecurityContext {
+	allowPrivilegeEscalation := false
+	return &k8score.SecurityContext{
+		AllowPrivilegeEscalation: &allowPrivilegeEscalation,
+		SeccompProfile: &k8score.SeccompProfile{
+			Type: k8score.SeccompProfileTypeRuntimeDefault,
+		},
+	}
+}
```
juliusvonkohout, line 152: We also need to drop all capabilities for all containers. Whether it is a user controlled container or argo executor.

juliusvonkohout: Capabilities: &k8score.Capabilities{
			Drop: []k8score.Capability{"ALL"},
		},  must be there as well

backend/src/v2/compiler/argocompiler/security_context.go
```diff
@@ -0,0 +1,168 @@
...
+		return
+	}
+	if container.SecurityContext.AllowPrivilegeEscalation == nil {
+		allowPrivilegeEscalation := false
+		container.SecurityContext.AllowPrivilegeEscalation = &allowPrivilegeEscalation
+	}
+	if container.SecurityContext.SeccompProfile == nil {
+		container.SecurityContext.SeccompProfile = &k8score.SeccompProfile{
+			Type: k8score.SeccompProfileTypeRuntimeDefault,
+		}
```
juliusvonkohout, line 174: The same for drop all capabilities

backend/src/v2/compiler/argocompiler/security_context_test.go
```diff
@@ -0,0 +1,297 @@
...
+	applySecurityContextToContainer(container)
+
+	assert.NotNil(t, container.SecurityContext)
+	assert.False(t, *container.SecurityContext.AllowPrivilegeEscalation)
+	assert.True(t, *container.SecurityContext.RunAsNonRoot)
+	assert.Equal(t, []k8score.Capability{"ALL"}, container.SecurityContext.Capabilities.Drop)
+	assert.Equal(t, k8score.SeccompProfileTypeRuntimeDefault, container.SecurityContext.SeccompProfile.Type)
+}
+
+func TestApplySecurityContextToContainer_PartialSecurityContext(t *testing.T) {
```
juliusvonkohout, line 63: here as well

juliusvonkohout:
Also you have to check the namespace manifests and make sure that there is the enforce PSS baseline annotation and warn on restricted annotation on the namespace. Only then the stuff you change here is at least partially enforced/tested.

jsonmp-k8:
> Also you have to check the namespace manifests and make sure that there is the enforce PSS baseline annotation and warn on restricted annotation on the namespace. Only then the stuff you change here is at least partially enforced/tested.

Can you please check now and let me know if I have missed any ?

backend/src/v2/compiler/argocompiler/argo.go
```diff
@@ -150,9 +150,20 @@ func Compile(jobArg *pipelinespec.PipelineJob, kubernetesSpecArg *pipelinespec.S
 		},
 	}
 
+	// Set security defaults at the workflow level as a baseline for all pods.
+	// Argo Workflows is designed to run rootless:
+	// https://argo-workflows.readthedocs.io/en/latest/workflow-pod-security-context/
+	// Per-template security context is also applied for defense-in-depth.
+	runAsNonRoot := true
+	wf.Spec.SecurityContext = &k8score.PodSecurityContext{
+		RunAsNonRoot: &runAsNonRoot,
```
juliusvonkohout, line 159: please also disable all capabilities and privilege escalation here.

backend/src/v2/compiler/argocompiler/security_context_test.go
```diff
@@ -0,0 +1,295 @@
...
+		},
+	}
+	applySecurityContextToTemplate(template)
+
+	// Pod security context: SeccompProfile and RunAsNonRoot
+	assert.NotNil(t, template.SecurityContext)
+	assert.NotNil(t, template.SecurityContext.RunAsNonRoot)
+	assert.True(t, *template.SecurityContext.RunAsNonRoot)
+	assert.Equal(t, k8score.SeccompProfileTypeRuntimeDefault, template.SecurityContext.SeccompProfile.Type)
+
```
juliusvonkohout, line 124: is is missing the drop all capabilities test ?

juliusvonkohout: Maybe even in multiple tests. Please check all tests.

backend/src/v2/compiler/argocompiler/argo.go
jsonmp-k8: .

backend/src/v2/compiler/argocompiler/security_context.go
jsonmp-k8: Done — `defaultUserContainerSecurityContext()` and `applyUserContainerSecurityContext()` now both include `Capabilities.Drop: ALL`. The modelcar driver containers also get matching security contexts. See commits b1927a32d and 92de7b6a7.

jsonmp-k8: Added in commit b1927a32d.

backend/src/v2/compiler/argocompiler/security_context.go
jsonmp-k8: Added in commit b1927a32d.

backend/src/v2/compiler/argocompiler/security_context_test.go
jsonmp-k8: Updated in commit b1927a32d — tests now assert `Capabilities.Drop: ALL` for all container types.

backend/src/v2/compiler/argocompiler/spec_patch_test.go
juliusvonkohout, line 1: is this missing drop all capabilities ?

manifests/gcp_marketplace/chart/kubeflow-pipelines/templates/argo.yaml
juliusvonkohout, line 873: that is orthogonal but should be the path provided by the actual argo-workfflow-controller configmap. You can copy it from  here https://github.com/kubeflow/pipelines/blob/1db659c02a758d2c7a3d233179e5640ab50fc1c9/manifests/kustomize/third-party/argo/base/workflow-controller-configmap-patch.yaml#L32

juliusvonkohout, line 874: we now have seaweedfs instead of minio

juliusvonkohout:
@jsonmp-k8 just a few small changes in the new comments and we should be good to merge.
CC @droctothorpe @HumairAK @zazulam @hbelmiro for final review.

/approve
already from my side since it improves the default a lot to roughly PSS baseline, without affecting the average user, since we mostly focus on KFP system level components and PSS baseline which still allows arbitrary users such as root/0 for customer workloads. But it blocks the typical easy Kubernetes cluster exploits.

jsonmp-k8:
Re: [argo.go:159 comment](https://github.com/kubeflow/pipelines/pull/12782#discussion_r2805577417) — `AllowPrivilegeEscalation` and `Capabilities` are container-level `SecurityContext` fields. They are not available on `PodSecurityContext` (which only supports RunAsUser, RunAsGroup, RunAsNonRoot, FSGroup, SeccompProfile, Sysctls, etc.).

We already enforce these on **every container** via the per-template functions:
- `applySecurityContextToTemplate()` sets `AllowPrivilegeEscalation=false` and `Capabilities.Drop=ALL` on all containers in system templates (driver, importer)
- `applySecurityContextToExecutorTemplate()` sets the same on all containers in executor templates (user containers, init containers, sidecars)
- Modelcar containers (added at driver runtime) also get the same security context (commit 92de7b6a7)

ref: https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.31/#podsecuritycontext-v1-core

droctothorpe:
Thank you for the contribution, @jsonmp-k8, and thank you for the thorough review, @juliusvonkohout!

/lgtm
/approve

juliusvonkohout:
Thank you @jsonmp-k8 , a follow up PR is described in https://github.com/kubeflow/pipelines/issues/12205#issuecomment-3901743839 if you are interested

jsonmp-k8:
> Thank you @jsonmp-k8 , a follow up PR is described in [#12205 (comment)](https://github.com/kubeflow/pipelines/issues/12205#issuecomment-3901743839) if you are interested

Happy to take that

PR #12783, opened 2026-02-07
fix(multi-user): sync frontend images on namespace deployments

jsonmp-k8:
## Summary
- Add `kubeflow-pipelines/image-spec-hash` annotation to pod templates in `sync.py` profile controller files
- When frontend or visualization server image/tag changes, the hash annotation changes, causing metacontroller to patch the Deployment and Kubernetes to trigger a rolling update
- Previously, namespace deployments had to be manually deleted and recreated to pick up image changes, as metacontroller's `InPlace` update strategy did not reliably detect container image changes

## Test plan
- [ ] Deploy updated profile controller with a known frontend image tag
- [ ] Change the `FRONTEND_TAG` (or `KFP_VERSION`) environment variable on the controller
- [ ] Verify that namespace `ml-pipeline-ui-artifact` deployments are automatically updated with the new image without manual deletion
- [ ] Verify that `ml-pipeline-visualizationserver` deployments (minio variant) are also updated

Signed-off-by: Jaison Paul <paul.jaison@gmail.com>

### DIFF
--- a/backend/src/v2/cmd/driver/main.go
+++ b/backend/src/v2/cmd/driver/main.go
@@ -29,6 +29,7 @@ import (
 	"os"
 	"path/filepath"
 	"strconv"
+	"strings"
 
 	"github.com/golang/glog"
 	"github.com/kubeflow/pipelines/api/v2alpha1/go/pipelinespec"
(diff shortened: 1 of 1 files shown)


jeffspahr:
Opus 4.6 assisted review:

  Looks Good

  - Correct pattern: Hashing the image spec into a pod annotation is the standard way to force rolling updates when non-spec fields change. This is the same
  approach used by helm with checksum/config annotations.
  - Clean refactor of the istio annotation: The old disable_istio_sidecar and {...} or {} pattern is a Python anti-pattern (it can silently produce wrong results
   if the truthy dict value is empty). The new {**({...} if condition else {})} is more explicit and correct.
  - Consistent across all pod templates: All three annotation blocks (1 in base sync.py, 2 in minio sync.py) are updated.
  - Minimal blast radius: Only the annotation changes, no other structural modifications.
  - No test infrastructure exists for these sync.py files, so the lack of tests is consistent with the project's existing pattern.

  ---
  Summary

  This is a small, well-scoped fix. The approach is sound, the implementation is consistent, and the code is actually slightly improved by the istio annotation
  refactor. No blockers — looks good to merge after fixing the "three files" mention in the description.

droctothorpe:
/approve
/lgtm

Thanks for the contribution, @jsonmp-k8 ! 🔥

PR #12701, opened 2026-01-24
fix(backend): support non-default cluster domain for metadata service

Priyanshu-u07:
**Description of your changes:**
This PR removes the hard-coded cluster.local value and makes the cluster domain configurable using a global environment variable.
If the variable is not set, it defaults to .svc.cluster.local to maintain backward compatibility.

This allows Kubeflow Pipelines to work correctly on clusters that use a non-default DNS domain.

Fixes kubeflow/pipelines#12649

**Checklist:**
- [x] You have [signed off your commits](https://www.kubeflow.org/docs/about/contributing/#sign-off-your-commits)
- [x] The title for your pull request (PR) should follow our title convention. [Learn more about the pull request title convention used in this repository](https://github.com/kubeflow/pipelines/blob/master/CONTRIBUTING.md#pull-request-title-convention).

PR #12786, opened 2026-02-07
fix(frontend); make cluster domain configurable for TensorBoard

### LINKED ISSUE #12785: [frontend] Hard-coded cluster.local breaks TensorBoard in custom DNS environments
### Environment

*  Installed via official docs: https://www.kubeflow.org/docs/components/pipelines/operator-guides/installation/
   Tested on: Kubernetes clusters with custom cluster domain (e.g., `cluster.corp`)
* KFP version: 2.14.3
* KFP SDK version: N/A (frontend server issue)

### Steps to reproduce
1. Deploy Kubeflow Pipelines on a Kubernetes cluster configured with a **custom cluster domain** (not the default `cluster.local`)
2. Run a pipeline that generates TensorBoard logs
3. Click "Open TensorBoard" from the Pipelines UI
4. TensorBoard fails to connect
The frontend server constructs the TensorBoard URL with a hard-coded `cluster.local`:

  **File**: [frontend/server/k8s-helper.ts (line 170)](https://github.com/kubeflow/pipelines/blob/master/frontend/server/k8s-helper.ts#L170)

 
`const address = http://${viewer.metadata.name}-service.${namespace}.svc.cluster.local:80/tensorboard/${viewer.metadata.name}/`;

In clusters using custom domains, this causes DNS resolution to fail.
### Expected result
The cluster domain should be configurable via the CLUSTER_DOMAIN environment variable. TensorBoard should work in any Kubernetes cluster regardless of its DNS configuration.

### Materials and Reference

- Affected file: [frontend/server/k8s-helper.ts (line 170)](https://github.com/kubeflow/pipelines/blob/master/frontend/server/k8s-helper.ts#L170)
- I will try to raise a PR to fix this issue.

---

Priyanshu-u07:

This PR removes the hard-coded `cluster.local` value for TensorBoard service addresses and makes the cluster domain configurable using the `CLUSTER_DOMAIN` environment variable. It also makes the object store service name configurable via the `OBJECT_STORE_HOST` environment variable, resolving CI DNS failures by providing a flexible way to point to the correct artifact storage host (defaulting to `seaweedfs`).

**Key improvements:**

- Configurable Domain: Adds `CLUSTER_DOMAIN` support to frontend and service layers.
- Configurable Service Host: Adds `OBJECT_STORE_HOST` support to the profile controller (sync.py) and deployment manifests to allow custom naming for artifact storage.
- Auto-Fallback:  Defaults to `cluster.local` and `seaweedfs` respectively to prevent breaking existing setups.
- Manifest Integration: Automatically pulls configuration from the shared ConfigMap.

This allows Kubeflow Pipelines to work correctly on Kubernetes clusters that use a non-default DNS domain (e.g., `cluster.corp`, `k8s.internal`).

**Related Fix**: This PR adds the frontend support for the custom cluster domain. Fix already implemented in the backend (#12701).

- Resolves CI DNS failures in multi-user E2E tests.

Fixes #12785 

**Checklist:**

### DIFF
--- a/.github/actions/create-cluster/action.yml
+++ b/.github/actions/create-cluster/action.yml
@@ -36,21 +36,21 @@ runs:
         docker save kindest/node:${{ inputs.k8s_version }} -o /tmp/kind-node-${{ inputs.k8s_version }}.tar
 
     - name: Create k8s Kind Cluster
-      uses: helm/kind-action@v1.12.0
+      uses: helm/kind-action@v1.13.0
       id: cluster-create
       with:
         cluster_name: ${{ inputs.cluster_name }}
         kubectl_version: ${{ inputs.k8s_version }}
-        version: v0.29.0
+        version: v0.31.0
         node_image: kindest/node:${{ inputs.k8s_version }}
         registry: true
 
     - name: Retry Create k8s Kind Cluster
-      uses: helm/kind-action@v1.12.0
+      uses: helm/kind-action@v1.13.0
       if: ${{ steps.cluster-create.outcome == 'failure' }}
       with:
         cluster_name: ${{ inputs.cluster_name }}
         kubectl_version: ${{ inputs.k8s_version }}
-        version: v0.29.0
+        version: v0.31.0
         node_image: kindest/node:${{ inputs.k8s_version }}
         registry: true
--- a/.github/workflows/pre-commit.yml
+++ b/.github/workflows/pre-commit.yml
@@ -22,7 +22,7 @@ jobs:
         with:
           go-version-file: go.mod
       - name: golangci-lint
-        uses: golangci/golangci-lint-action@v8
+        uses: golangci/golangci-lint-action@v9
         with:
           version: v2.3
           args: --new --new-from-merge-base=origin/${{ github.base_ref || github.ref_name }}
--- a/.github/workflows/trivy.yml
+++ b/.github/workflows/trivy.yml
@@ -14,7 +14,7 @@ jobs:
         uses: actions/checkout@v4
 
       - name: Run Trivy vulnerability scanner in repo mode
-        uses: aquasecurity/trivy-action@0.33.1
+        uses: aquasecurity/trivy-action@0.34.0
         with:
           scan-type: 'fs'
           ignore-unfixed: true
--- a/frontend/server/app.test.ts
+++ b/frontend/server/app.test.ts
@@ -38,10 +38,54 @@ describe('UIServer apis', () => {
   const commitHash = 'abcdefg';
   const { argv, buildDate, indexHtmlContent } = commonSetup({ tagName, commitHash });
 
-  afterEach(() => {
-    if (app) {
-      app.close();
+  async function waitForListening(server: Server): Promise<void> {
+    if (server.listening) {
+      return;
     }
+    await new Promise<void>((resolve, reject) => {
+      const onListening = (): void => {
+        cleanup();
+        resolve();
+      };
+      const onError = (err: Error): void => {
+        cleanup();
+        reject(err);
+      };
+      const cleanup = (): void => {
+        server.off('listening', onListening);
+        server.off('error', onError);
+      };
+
+      server.on('listening', onListening);
+      server.on('error', onError);
+    });
+  }
+
+  afterEach(async () => {
+    if (app) {
+      await app.close();
+    }
+  });
+
+  it('allows restarting on the same port after close resolves', async () => {
+    app = new UIServer(loadConfigs(argv, {}));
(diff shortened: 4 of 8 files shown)


frontend/server/k8s-helper.ts
```diff
@@ -146,6 +149,7 @@ export async function newTensorboardInstance(
 export async function getTensorboardInstance(
   logdir: string,
   namespace: string,
```
jeffspahr, line 152: **Bug:** The backend PR #12701 (`metadata_helpers.py:36`) defaults `CLUSTER_DOMAIN` to `.svc.cluster.local` (with the `.svc.` prefix), but here you hardcode `.svc.` in the template and default `CLUSTER_DOMAIN` to `cluster.local` (without the prefix).

This means the same env var has different expected formats in the backend vs frontend. An operator cannot set a single value that works for both.

Either:
- Change this PR to match the backend convention (default to `.svc.cluster.local`, drop the hardcoded `.svc.` from the template), or
- Change the backend to match this PR's convention

The former is probably safer since #12701 is already merged.

frontend/server/handlers/tensorboard.ts
```diff
@@ -57,7 +57,13 @@ export const getTensorboardHandlers = (
...
+      res.send(
+        await k8sHelper.getTensorboardInstance(
+          logdir as string,
+          namespace as string,
+          tensorboardConfig.clusterDomain,
+        ),
+      );
     } catch (err) {
       const details = await parseError(err);
       console.error(`Failed to list Tensorboard pods: ${details.message}`, details.additionalInfo);
```
jeffspahr, line 69: **Bug:** `clusterDomain` is passed to `getTensorboardInstance` here (the `get` handler), but `waitForTensorboardInstance` (called in the `create` handler at line 127) and `deleteTensorboardInstance` (called in the `delete` handler at line 168) both call `getTensorboardInstance` internally **without** passing `clusterDomain`. They will always use the default `cluster.local`.

Both `waitForTensorboardInstance` and `deleteTensorboardInstance` need the same `clusterDomain` parameter threaded through, and the `create`/`delete` handlers need to pass it.

Priyanshu-u07:
Hi @jeffspahr,
I have updated the PR based on your comment and all CI checks are passing now. Would appreciate another review when you get a chance. Thanks

frontend/server/k8s-helper.ts
```diff
@@ -165,12 +170,17 @@ export async function getTensorboardInstance(
       // but if there is a hash collision, not check logdir will return error tensorboard link
       // if check logdir and then create Viewer CRD with same name will break anyway.
       // TODO fix hash collision
-      (viewer: any) => {
+      (response: any) => {
+        const viewer = response.body || response;
```
jeffspahr: **Behavioral change unrelated to feature:** The original code accesses `viewer.body.spec` and `viewer.body.metadata.name` directly. This PR changes the unpacking to `response.body || response` then accesses `viewer.spec` / `viewer.metadata.name`.

This silently changes how the K8s client response object is parsed. The existing code on `master` (line 169) uses `viewer.body.spec` and it works, so the old shape was correct for the current client library version.

If this fixes a real issue (e.g., a client library version change), it should be called out in the PR description and ideally in its own commit. If not, it risks breaking the existing flow. Please justify this change or revert it.

manifests/kustomize/base/installs/multi-user/pipelines-profile-controller/sync.py
```diff
@@ -91,9 +96,10 @@ def server_factory(frontend_image,
                    frontend_tag,
                    disable_istio_sidecar,
                    artifacts_proxy_enabled,
-                   artifact_retention_days,
-                   url="",
-                   controller_port=8080):
+                    artifact_retention_days,
```
jeffspahr: `cluster_domain` is a required positional parameter with no default, unlike every other parameter in this function. If any caller invokes `server_factory` without it, they will get a `TypeError`. Add a default for safety and consistency:

```python
cluster_domain=".svc.cluster.local",
```

Priyanshu-u07:
Hi @jeffspahr 
All CI checks are passing now. Please review when you get a chance.

jeffspahr:
lgtm @Priyanshu-u07! 

@droctothorpe this is good to merge when you get a chance.

Priyanshu-u07:
updated the PR description and style nits.
@droctothorpe could you approve and merge when get a chance?

droctothorpe:
/lgtm
/approve

Thank you for the contribution, @Priyanshu-u07, and thank you for the reviews, @jeffspahr! 🙏

PR #12790, opened 2026-02-09
feat(security): run all container images as non-root user

jsonmp-k8:
## Summary
- Add non-root user (UID 65532) to Dockerfiles that were running as root: apiserver, viewercontroller, visualization, metadata-writer, cache-deployer, frontend, and metadata-envoy.
- Pin UID for images that already had a non-root user but without a stable UID: cacheserver, persistenceagent, scheduledworkflow.
- Tighten cache-deployer file permissions from 777 to 755.

## Details

All deployment manifests already enforce `runAsNonRoot: true` and `runAsUser: 1000` via securityContext. These Dockerfile changes ensure the images themselves default to a non-root user (UID 65532) when run outside of Kubernetes (e.g. standalone `docker run`), and provide a stable UID for `/etc/passwd` resolution.

### Skipped images
- `proxy/Dockerfile` — GCP-specific, uses a pinned base image digest with `hostNetwork: true`
- `third_party/minio/Dockerfile` — minio manages its own user internally; manifest already enforces non-root

## Test plan
- [ ] Verify all modified Dockerfiles build successfully
- [ ] Verify containers start correctly with the non-root user
- [ ] Confirm no file permission issues at runtime for each image
- [ ] Validate cache-deployer scripts execute correctly with 755 permissions

### DIFF
--- a/backend/Dockerfile
+++ b/backend/Dockerfile
@@ -82,6 +82,9 @@ RUN chmod +x /bin/apiserver
 RUN sed -E "s#/(blob|tree)/master/#/\1/${COMMIT_SHA}/#g" -i /config/sample_config.json && \
     sed -E "s/%252Fmaster/%252F${COMMIT_SHA}/#g" -i /config/sample_config.json
 
+RUN adduser --system --uid 65532 --no-create-home appuser
+USER 65532
+
 # Expose apiserver port
 EXPOSE 8888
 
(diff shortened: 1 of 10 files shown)


juliusvonkohout:
Hello, did you raise a follow up PR?

jsonmp-k8:
> Hello, did you raise a follow up PR?

Yes, trying to run containers are non-root. There is more work needed on this. Will let you know once its ready.

PR #12792, opened 2026-02-09
Fix SeaweedFS IAM endpoint in profile controller

sduvvuri1603:
### Summary
The profile controller creates the IAM client without using AWS_ENDPOINT_URL, so in SeaweedFS mode it can call AWS IAM instead of SeaweedFS. This prevents creation of the per‑profile mlpipeline-minio-artifact secret and causes the artifact‑proxy service to be unreachable in the user namespace. This PR makes the IAM client respect AWS_ENDPOINT_URL and adds a small test to ensure the endpoint is used.

### Bug
In pipelines-profile-controller/sync.py, iam = session.create_client('iam', region_name='foobar') ignores AWS_ENDPOINT_URL even though the deployment sets it to http://seaweedfs.kubeflow:8111.

### Change
- Create the IAM client with endpoint_url=AWS_ENDPOINT_URL when provided.
- Add a unit-style test that asserts the endpoint is passed to the IAM client.

### Expected impact
SeaweedFS multi‑user profile reconciliation will create mlpipeline-minio-artifact in user namespaces, unblocking the artifact‑proxy health check.

### DIFF
--- a/manifests/kustomize/base/installs/multi-user/pipelines-profile-controller/sync.py
+++ b/manifests/kustomize/base/installs/multi-user/pipelines-profile-controller/sync.py
@@ -24,11 +24,20 @@ import botocore.session
 S3_BUCKET_NAME = 'mlpipeline'
 
 session = botocore.session.get_session()
-# To interact with seaweedfs user management. Region does not matter.
-iam = session.create_client('iam', region_name='foobar')
 # S3 client for lifecycle policy management
 s3_endpoint_url = os.environ.get("S3_ENDPOINT_URL", "http://seaweedfs.kubeflow:8333")
 s3 = session.create_client('s3', region_name='foobar', endpoint_url=s3_endpoint_url)
+
+
+def create_iam_client():
+    # To interact with SeaweedFS user management. Region does not matter.
+    endpoint_url = os.environ.get("AWS_ENDPOINT_URL")
+    if endpoint_url:
+        return session.create_client('iam', region_name='foobar', endpoint_url=endpoint_url)
+    return session.create_client('iam', region_name='foobar')
+
+
+iam = create_iam_client()
 
 
 def main():
--- a/manifests/kustomize/base/installs/multi-user/pipelines-profile-controller/test_sync.py
+++ b/manifests/kustomize/base/installs/multi-user/pipelines-profile-controller/test_sync.py
@@ -1,6 +1,7 @@
 import os
 from unittest import mock
 import threading
+import sync
 from sync import get_settings_from_env, server_factory
 import json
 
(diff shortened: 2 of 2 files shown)


manifests/kustomize/base/installs/multi-user/pipelines-profile-controller/sync.py
```diff
@@ -251,10 +259,9 @@ def sync(self, parent, attachments):
                                     "labels": {
                                         "app": "ml-pipeline-ui-artifact"
                                     },
-                                    "annotations": {
-                                        **({"sidecar.istio.io/inject": "false"} if disable_istio_sidecar else {}),
-                                        "kubeflow-pipelines/image-spec-hash": hashlib.sha256(f"{frontend_image}:{frontend_tag}".encode()).hexdigest()[:16],
```
nsingla, line 256: why was this removed?

sduvvuri1603: That was removed by accident. Reverting back the change
Thanks for that!

alyssacgoins: @nsingla @sduvvuri1603 it looks like there is still a diff here. Is that intentional?

manifests/kustomize/base/installs/multi-user/pipelines-profile-controller/sync.py
```diff
@@ -24,13 +24,22 @@
...
 
 session = botocore.session.get_session()
-# To interact with seaweedfs user management. Region does not matter.
-iam = session.create_client('iam', region_name='foobar')
 # S3 client for lifecycle policy management
 s3_endpoint_url = os.environ.get("S3_ENDPOINT_URL", "http://seaweedfs.kubeflow:8333")
 s3 = session.create_client('s3', region_name='foobar', endpoint_url=s3_endpoint_url)
 
 
+def create_iam_client():
```
alyssacgoins, line 32: more of a nitpick, but is there a specific reason why you moved `iam` initialization below `s3_endpoint_url` and `s3`, when it was previously declared above?

sduvvuri1603: Thanks for the nit. I moved IAM initialization because it now depends on AWS_ENDPOINT_URL; I wanted to keep the env-derived values in place before creating the client. Happy to adjust if that is recommended.

manifests/kustomize/base/installs/multi-user/pipelines-profile-controller/sync.py
nsingla: is there a diff other than formatting 🤔

sduvvuri1603: @alyssacgoins I pushed the commit earlier and it doesn't show up on my end anymore. Pls lmk if the issue still persists.

nsingla: is there a diff other than formatting 🤔

manifests/kustomize/base/installs/multi-user/pipelines-profile-controller/sync.py
nsingla: i think that's to account for condition when `AWS_ENDPOINT_URL` is not specified.

manifests/kustomize/base/installs/multi-user/pipelines-profile-controller/sync.py
alyssacgoins: Diff is still there - you should be able to see it on the Github diff here. One option would be to copy your changes, revert the file changes, and copy only the necessary changes back.

manifests/kustomize/base/installs/multi-user/pipelines-profile-controller/sync.py
alyssacgoins: Makes sense :+1:

nsingla (approved):
Lgtm

nsingla (approved):
/approve
/lgtm

PR #12754, opened 2026-02-01
feat(frontend): migrate CRA to Vite + Vitest

jeffspahr:
## What
- Replace CRA with Vite 7 for build/dev (ESM-first) and align build output + proxy behavior.
- Migrate UI tests from Jest/Enzyme to Vitest + Testing Library; keep server tests on Jest for now.
- Restore TypeScript typecheck parity and align TS/tooling versions for Node 22.
- Version-risk callout: `react-router`/`react-router-dom` upgraded from `^4.3.1` to `^5.3.4` (routing behavior and API differences can affect runtime flows).
- Version-risk callout: `typescript` upgraded from `^3.8.3` to `^4.9.5` (stricter checking and generated-client compatibility adjustments via `scripts/fix_swagger_codegen_delete.js`).
- Update Storybook config, docs, and CI commands; add local visual diff helper + runner.

## Why
- CRA blocks the React upgrade; this migration removes that blocker and aligns Node/TS/tooling for the next React bump.
- We cannot safely upgrade React while CRA + legacy test/tooling dependencies remain in place.
- This work directly addresses https://github.com/kubeflow/pipelines/issues/10098.
- Vitest is Vite-native and stabilizes flaky UI suites without skipping coverage.
- The docs and helper script make the migration reviewable and repeatable.

## Why this is a good commit
- Scope is documented and bounded (no React upgrade); runtime behavior preserved.
- Test parity maintained while removing flake sources; no skipped tests or reduced coverage.
- Breaking change (modern-only browsers) is explicit and tracked in docs.
- Follow-up work is clearly deferred to post-effort TODOs.

## Scope Add (Intentional Non-Migration Behavior Changes)
- `frontend/src/lib/StatusUtils.ts`: treat `NodePhase.OMITTED` as `notStarted` in status background mapping.
- `frontend/src/pages/StatusV2.tsx`: add explicit `PAUSED` runtime-state handling and suppress unknown-state logs when state is missing.
- `frontend/src/pages/NewRun.tsx`: normalize non-`Error` exceptions into `Error` objects in experiment fetch and validation paths (while keeping v1->v2 experiment fallback behavior).
- `frontend/src/pages/RunList.tsx`: add unmount-safe `setStateSafe` guards for async updates and pass runtime state directly to status rendering (`run.state` instead of `run.state || '-'`).
- These were intentional scope-adds to improve edge-case UX/reliability and reduce runtime/test noise surfaced during migration.

## Docs / Plans
- frontend/docs/vite-migration-plan.md
[...]

PR #12796, opened 2026-02-10
fix(frontend): update server integration tests to support Vite build output

kaikaila:
## Summary
- Update server integration tests to handle both CRA (Create React App) and Vite build output formats
- Fix JS/CSS bundle path matching regex to support Vite's `static/index-[hash].js` pattern alongside CRA's `static/js/main.[hash].js`
- Add case-insensitive check for HTML doctype to handle Vite's `<!DOCTYPE html>` vs CRA's `<!doctype html>`
- Handle Vite's relative path prefix (`./`) in asset URLs

## Test plan
- [ ] Run server integration tests: `node test/server-integration-test/server-test.js`
- [ ] Verify tests pass with both CRA and Vite frontend builds

### DIFF
--- a/test/server-integration-test/server-test.js
+++ b/test/server-integration-test/server-test.js
@@ -154,33 +154,35 @@ async function testStaticServing() {
   await test('GET / returns index.html', async () => {
     const res = await request('GET', '/');
     assertEqual(res.status, 200, 'Status code');
-    assertContains(res.body, '<!doctype html>', 'HTML doctype');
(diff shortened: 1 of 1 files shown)


jeffspahr:
This looks good to me as a quick fix.  My [follow up PR](https://github.com/kubeflow/pipelines/pull/12793) includes a change in CI to prevent the regression introduced in #12754, but let's merge this first to unblock others.

PR #12803, opened 2026-02-12
fix(backend): increase gRPC message size limit for MLMD. Fixes #11949

### LINKED ISSUE #11949: [backend] Received message larger than max when creating MLMD artifact
### Environment

Standalone installation on AWS EKS cluster, using AWS Aurora MySQL database and AWS S3 for storage.
*  KFP version: 

2.5.0
*  KFP SDK version: 

2.13.0

### Steps to reproduce

When trying to write classification metrics larger than ~4MB, the following error is raised:
```
main.go:51] failed to execute component: unable to produce MLMD artifact for output "classification_metrics": rpc error: code = ResourceExhausted desc = Received message larger than max (91000474 vs. 4194304)
```

To reproduce, you can create the following pipeline (V2):
```python
import kfp
from kfp.dsl import ClassificationMetrics, Output, component


@component(base_image="python:3.9", packages_to_install=["numpy"])
def generate_large_classification_metrics(
    classification_metrics: Output[ClassificationMetrics],
) -> None:
    """Generate a classification metrics artifact larger than 4MB."""
    import json
    import numpy as np

    # Generate a large number of ROC curve points to create a large metrics file
    # This will create approximately 10,000,000 data points
    # Each point is (fpr, tpr, threshold) which will exceed 4MB when serialized
    num_points = 1_000_000

    # Create evenly spaced values between 0 and 1
    fpr = np.linspace(0, 1, num_points).tolist()
    tpr = np.linspace(0, 1, num_points).tolist()
    thresholds = np.linspace(1, 0, num_points).tolist()

    # Log the ROC curve data
    classification_metrics.log_roc_curve(fpr, tpr, thresholds)

    # Save the metrics to disk
    with open(classification_metrics.path, "w") as file_handle:
        json.dump(classification_metrics.metadata, file_handle)

    # Print the size of the metrics file
    import os

    file_size_bytes = os.path.getsize(classification_metrics.path)
    file_size_mb = file_size_bytes / (1024 * 1024)
    print(f"Generated classification metrics file size: {file_size_mb:.2f} MB")


# Define the pipeline
@kfp.dsl.pipeline(
    name="large-metrics-test-pipeline",
    description="Pipeline to test large classification metrics artifacts",
)
def large_metrics_pipeline():
    generate_large_classification_metrics()


if __name__ == "__main__":
    kfp.compiler.Compiler().compile(large_metrics_pipeline, "large_metrics_pipeline.yaml")
```

### Expected result

MLMD artifact is created successfully, no error is raised.

### Materials and Reference

We've tried to:
- Increase `GroupConcatMaxLen` in the `apiserver` [config.json](https://github.com/kubeflow/pipelines/blob/f342504655c9edfff95fbfa1eba2f1e2f491cbb6/backend/src/apiserver/config/config.json#L6) by mounting a custom config map to `ml-pipeline` container. The file was successfully replaced, deployment restarted, but it didn't help
- Patching the cache server - setting `db_group_concat_max_len` argument => no change:
```yaml
apiVersion: apps/v1
kind: Deployment
metadata:
  name: cache-server
spec:
  template:
    spec:
      containers:
        - name: server
          args:
            - "--tls_cert_filename=tls.crt"
            - "--tls_key_filename=tls.key"
            - "--db_host=$(DBCONFIG_HOST_NAME)"
            - "--db_user=$(DBCONFIG_USER)"
            - "--db_password=$(DBCONFIG_PASSWORD)"
            - "--db_group_concat_max_len=18446744073709551615" # https://dev.mysql.com/doc/refman/8.4/en/server-system-variables.html#sysvar_group_concat_max_len
```
---

jsonmp-k8:
## Summary
- Increase the gRPC max message size from the default 4MB to 100MB for the metadata (MLMD) client and server, fixing `ResourceExhausted` errors when writing large artifacts (e.g. classification metrics with many data points).
- Add `MaxCallRecvMsgSize`/`MaxCallSendMsgSize` (100MB) to the metadata gRPC client in `client.go`, matching the pattern already used by the cache client in `cacheutils/cache.go`.
- Add `grpc.max_receive_message_length` and `grpc.max_send_message_length` (100MB) to the MLMD server deployment args (base and TLS variants).

Fixes #11949

## Test plan
- [ ] Deploy updated manifests and verify MLMD server starts with the new gRPC channel arguments
- [ ] Run a pipeline that produces a classification metrics artifact larger than 4MB and verify it completes without `ResourceExhausted` error
- [ ] Verify existing pipelines with small artifacts continue to work

### DIFF
--- a/backend/src/v2/metadata/client.go
+++ b/backend/src/v2/metadata/client.go
@@ -22,6 +22,7 @@ import (
 	"encoding/json"
 	"errors"
 	"fmt"
+	"os"
 	"path"
 	"strconv"
 	"strings"
@@ -55,11 +56,29 @@ const (
 	ImporterExecutionTypeName          = "system.ImporterExecution"
 	ImporterWorkspaceExecutionTypeName = "system.ImporterWorkspaceExecution"
 	mlmdClientSideMaxRetries           = 3
-	// MaxGRPCMessageSize is the max gRPC message size for the metadata client.
-	// The default gRPC limit is 4MB which is too small for large artifacts
-	// (e.g. classification metrics with many data points).
-	MaxGRPCMessageSize = 100 * 1024 * 1024 // 100MB
+	defaultMaxGRPCMessageSize          = 100 * 1024 * 1024 // 100MB
+	maxGRPCMessageSizeEnv              = "METADATA_GRPC_MESSAGE_SIZE"
 )
+
+// MaxGRPCMessageSize is the max gRPC message size for the metadata client.
+// The default gRPC limit is 4MB which is too small for large artifacts
+// (e.g. classification metrics with many data points).
+// Configurable via the METADATA_GRPC_MESSAGE_SIZE environment variable (in bytes).
+var MaxGRPCMessageSize = defaultMaxGRPCMessageSize
+
+func init() {
+	if v := os.Getenv(maxGRPCMessageSizeEnv); v != "" {
+		size, err := strconv.Atoi(v)
+		if err != nil {
+			glog.Fatalf("%s environment variable must be a valid integer: %v", maxGRPCMessageSizeEnv, err)
+		}
+		if size <= 0 {
+			glog.Fatalf("%s environment variable must be a positive integer, got %d", maxGRPCMessageSizeEnv, size)
+		}
+		MaxGRPCMessageSize = size
+		glog.Infof("MaxGRPCMessageSize set to %d bytes from %s", size, maxGRPCMessageSizeEnv)
+	}
+}
 
 type ExecutionType string
 


backend/src/v2/metadata/client.go
```diff
@@ -55,6 +55,10 @@ const (
 	ImporterExecutionTypeName          = "system.ImporterExecution"
 	ImporterWorkspaceExecutionTypeName = "system.ImporterWorkspaceExecution"
 	mlmdClientSideMaxRetries           = 3
+	// MaxGRPCMessageSize is the max gRPC message size for the metadata client.
+	// The default gRPC limit is 4MB which is too small for large artifacts
+	// (e.g. classification metrics with many data points).
+	MaxGRPCMessageSize = 100 * 1024 * 1024 // 100MB
```
droctothorpe: I don't mind merging this but FYI we are going to deprecate MLMD in v3. Also, this should probably be configurable through an env var as well.

zazulam: +1. I think it would still be a nice addition for teams who stay on 2.x for a while

zazulam (commented):
/lgtm
/approve

backend/src/v2/metadata/client.go
jsonmp-k8: Done — `MaxGRPCMessageSize` is now configurable via the `METADATA_GRPC_MESSAGE_SIZE` environment variable (in bytes), defaulting to 100MB if unset.

PR #12812, opened 2026-02-13
docs(agent): Add code-tree knowledge graph for AI-assisted development and PR reviews

nsingla:
### Summary                                                   

  - Restructure monolithic AGENTS.md (600 lines) into a slim entry point (65 lines) with 7 topic-specific guides under docs/agents/, enabling agents to
  load only the context relevant to their current task
  - Add code-tree knowledge graph tooling (tools/code-tree/) with pre-built artifacts (docs/code-tree/) -- 24,766 nodes, 86,988 edges covering 2,736
  files across 7 languages -- for symbol lookup, dependency tracing, impact analysis, and test mapping without reading source files
  - Restructure .github/copilot-instructions.md with batch-based review strategy, context pruning rules, and code-tree integration for PR reviews
  - Add 5 focused review guides under .github/review-guides/ (architecture-context, impact-analysis, language-checklists, security-review, test-quality)
  - Add CLAUDE.md as a task-routing operational playbook for Claude Code sessions with development strategy, validation checklists, and on-demand guide loading

  ### New files

  | Path | Purpose |                                                                                                                                    
  |------|---------|                                                                                                                                    
  | `tools/code-tree/code_tree.py` | Builds knowledge graph from repo (parser + call graph + test mapping) |                                            
  | `tools/code-tree/query_graph.py` | Queries the graph (symbol lookup, deps, impact, callers, hierarchy) |
  | `docs/code-tree/{graph,modules,tags}.json` | Pre-built graph artifacts |                                                                            
  | `docs/code-tree/summary.md` | Architecture overview generated from the graph |                                                                      
  | `docs/agents/*.md` (7 files) | Topic-specific agent guides split from AGENTS.md |
  | `.github/review-guides/*.md` (5 files) | PR review checklists by concern area |
  | `CLAUDE.md` | Claude Code session playbook |

  ### Motivation

  AI agents (Copilot, Claude Code, etc.) working in this monorepo previously loaded the entire 600-line AGENTS.md into context regardless of task. This
  wastes context window and reduces response quality. The new approach:

  1. Load on demand -- agents load only the guide matching their task type
  2. Code-tree first -- query the knowledge graph to locate symbols and assess blast radius before reading source files, reducing unnecessary file reads
  3. Structured PR reviews -- batch files by subsystem, prune context between batches, use impact analysis to flag untested blast radius

### How Knowledge graph is generated:

1. **Discover**

  Walks the repo with os.walk(), finds all source files (.py, .go, .js, .ts, .java, .rs, etc.), skips binary/generated/vendored directories, and detects
   build files (go.mod, package.json, pyproject.toml) as module boundaries.

2. **Parse**

  Extracts symbols (classes, functions, methods, constants) and edges (imports, inheritance, calls) from each file using three parser tiers:
  - Python AST (ast module) — for .py files, most accurate
  - Tree-sitter — for Go/JS/TS/Java/Rust if installed, good accuracy
  - Regex patterns — fallback for everything else

  It also extracts raw call references (e.g., "function A calls something named foo" at line 42).

3. **Resolve**

  - Import resolution: maps import X to the actual file path in the repo
  - Call resolution: matches raw call names (foo, self.bar, pkg.func) to known symbol IDs using same-file lookup, import tracking, and name matching
  - Test mapping: links test files/functions to the source they test via path mirroring (test_foo.py → foo.py), imports, and naming conventions (test_X
  → X)

4. **Output**

  Writes 4 files to docs/code-tree/:
  - graph.json — full knowledge graph (nodes + edges + metadata)
  - tags.json — flat symbol index for quick lookup
  - modules.json — directory-level dependency map
  - summary.md — human-readable overview with hub files, inheritance trees, stats

  **Key data model**

  Symbol  = { id, type, name, file, line_start, line_end, params, bases, ... }
  Edge    = { source, target, type }  # types: contains, imports, inherits, calls, tests
  CallRef = { caller_id, raw_name, line, file }  # unresolved, gets resolved in step 3

  It also supports incremental mode (--incremental): hashes files, compares to previous run, only reparses changed files and merges with cached results.

**Checklist:**

### DIFF
--- a/.github/copilot-instructions.md
+++ b/.github/copilot-instructions.md
@@ -1,44 +1,114 @@
 # Copilot Instructions
 
+Tier 1 PR reviewer for the Kubeflow Pipelines monorepo. Maintainer time is limited; your vigilance helps ensure project continuity against high-volume LLM-generated PRs.
+
 ## Review style
 
-- The volume of PRs has increased significantly as a result of LLMs. Your job is to help maintain the codebase by being a tier 1 reviewer for incoming PRs. Maintainer time is limited. Your vigilance will help ensure the continuity of the project.
-- Explain the "why" behind recommendations using bullet points.
-- Don't get bogged down in small details that will be caught by the hooks listed in `.pre-commit-config.yaml` anyway.
-- Make sure that functions have clear docstrings and are well-documented.
-- Make sure that type hints are used consistently for Python code.
-- Make sure that logic is clear and easy to follow.
-- Make sure that functions are reasonably sized and properly decomposed.
-- Make sure complex logic is broken down into smaller functions.
-- Make sure complex logic has sufficient comments, within reason, to explain the why behind the logic.
-- Make sure that variable names are descriptive and meaningful.
-- Make sure that new logic is covered by unit tests.
-- Make sure that large changes have adequate large tests, not just small tests.
-- Make sure that the PR description is clear.
-- Identify design flaws, uneccessary redundancy, and security issues introduced by the PR.
-- Provide an estimate about whether or not the PR is LLM-generated. Include a confidence level for the estimate. Do this as an HTML comment in the "PR Overview" section.
-- Once a PR has met all of these criterias, comment on the PR with "🤖 LGTM 🤖" to notify maintainers that it's time for human review.
-- Adhere to the following software engineering principles:
+- Explain the "why" behind recommendations.
+- Skip details caught by `.pre-commit-config.yaml` hooks.
+- Verify: clear docstrings, type hints, descriptive names, proper decomposition, sufficient comments.
+- Verify: new logic has tests proportional to change size; PR description explains the "why".
+- Identify design flaws, redundancy, and security issues.
+- Estimate LLM-generation likelihood with confidence level as an HTML comment in "PR Overview".
+- Comment "All Copilot criteria are met." when all criteria are met.
 
-FUNDAMENTAL SOFTWARE ENGINEERING PRINCIPLES
+## Review strategy
 
---- CORE DESIGN PRINCIPLES (SOLID) ---
-S - Single Responsibility Principle (SRP): A class or module should have one, and only one, reason to change.
-O - Open/Closed Principle (OCP): Software entities should be open for extension, but closed for modification.
-L - Liskov Substitution Principle (LSP): Subtypes must be substitutable for their base types without altering program correctness.
-I - Interface Segregation Principle (ISP): Clients should not be forced to depend on methods they do not use; prefer small, specific interfaces.
-D - Dependency Inversion Principle (DIP): Depend upon abstractions (interfaces), not concretes (classes).
+This file is the canonical source for PR review procedure details. Keep `AGENTS.md` aligned as
+high-level guardrails and avoid duplicating fine-grained workflow logic in multiple places.
 
---- ESSENTIAL DEVELOPMENT PRINCIPLES ---
-DRY (Don't Repeat Yourself): Every piece of knowledge must have a single, unambiguous representation within a system.
-KISS (Keep It Simple, Stupid): Avoid unnecessary complexity; simple code is easier to maintain and debug.
-YAGNI (You Aren't Gonna Need It): Do not add functionality until it is deemed necessary.
-Composition Over Inheritance: Prefer achieving polymorphic behavior and code reuse by composing objects rather than inheriting from a base class.
-Law of Demeter (Principle of Least Knowledge): A module should only talk to its immediate neighbors, not to strangers.
+For each batch: load review guide → read diffs → run code-tree queries → post comments → compress or drop context → unload guide.
 
---- SOFTWARE ARCHITECTURE & QUALITY PRINCIPLES ---
-Separation of Concerns: Divide code into distinct sections, each addressing a separate concern or functionality.
-Encapsulation: Hide the internal state and behavior of an object, exposing only necessary functionality.
-High Cohesion: Elements within a module should belong together, focusing on a single, well-defined task.
-Low Coupling: Minimize dependencies between modules to reduce the impact of changes.
-Boy Scout Rule: Always leave the code cleaner than you found it.
+### Small-batch file review
+
+Review files in batches of 3-5, in this order:
+
+1. **Proto/API** -- `.proto` files and generated code diffs
+2. **Backend Go** -- `backend/` changes (error handling, interface compliance)
+3. **SDK Python** -- `sdk/python/` changes (type annotations, docstrings)
+4. **Kubernetes Platform** -- `kubernetes_platform/` changes (proto, Go, Python reviewed together as a self-contained module)
+5. **Tests and goldens** -- test files against the code they test
+6. **Manifests and CI** -- `manifests/` and `.github/` changes
+
+### Context pruning between batches
+
+Skip for PRs with ≤5 changed files.
+
+Decide whether to compress or drop based on downstream consumers:
+
+| Completed batch | Downstream consumers |
+|-----------------|----------------------|
+| Proto/API | Backend Go, SDK Python, Kubernetes Platform |
+| Backend Go | Tests |
+| SDK Python | Tests |
+| Kubernetes Platform | Backend Go, Tests |
+| Tests | _(none)_ |
+| Manifests/CI | _(none)_ |
+
+- **Has consumers:** collapse diffs into a ≤10-line summary -- new/changed public APIs, error contracts, invariants, and issues found (file:line, severity). Drop all line-level detail.
+- **No consumers:** drop batch context entirely after posting review comments. Do not compress.
+
+If a later batch uncovers a dependency on dropped context, use code-tree queries (next section) to retrieve only the specific symbol -- never reload the full batch.
+
+### Code-tree impact analysis
+
+After reading changed files, regenerate the knowledge graph and run queries before writing comments:
+
+```bash
+# Regenerate graph to capture PR changes
+python3 tools/code-tree/code_tree.py --repo-root . --incremental -q
+
+python3 tools/code-tree/query_graph.py --rdeps <changed-file>      # blast radius
+python3 tools/code-tree/query_graph.py --impact <changed-file> --depth 5  # transitive
+python3 tools/code-tree/query_graph.py --test-impact <changed-file> # affected tests
+python3 tools/code-tree/query_graph.py --callers <function-name>    # all call sites
+python3 tools/code-tree/query_graph.py --hierarchy <class-name>     # inheritance
+```
+
+Use results to flag untested blast radius, missing caller updates, and cardinality assumption breaks.
+
+### Backend endpoint/query reference
+
+Whenever a new backend API endpoint and/or SQL/query logic is added or edited, use
+[PR #12999](https://github.com/kubeflow/pipelines/pull/12999) comments and test coverage level as a
+reference for review depth:
+
+- Check endpoint behavior consistency across create/upload/get/list/update flows and call out intentional differences.
+- Require positive + negative tests for each new or edited endpoint, including validation/error paths.
+- For `pipeline_store.go` changes, verify filter/query-building correctness and safe parameterization.
+- For `pipeline_store_kubernetes.go` changes, evaluate cache correctness vs performance tradeoffs and ensure race-sensitive tests cover update/delete then immediate read paths.
+
+### Security scan
+
+Run the [security-code-reviewer](https://github.com/anthropics/claude-code-action/blob/main/.claude/agents/security-code-reviewer.md) as a separate, independent pass on every PR. It does not require the knowledge graph or saved context from other review batches. It scans for:
+
+- **OWASP Top 10** -- injection, broken auth, sensitive data exposure, XXE, broken access control, security misconfiguration, XSS, insecure deserialization, known-vulnerable components, insufficient logging
+- **Input validation** -- unsanitized user input, path traversal, missing type/format/range checks
+- **Auth/authz** -- session management, privilege escalation, IDOR, RBAC gaps
+
+Findings are reported by severity (Critical > High > Medium > Low > Informational) with CWE references.
+
+## Review guide files
+
+| File | Load when |
+|------|-----------|
+| [architecture-context.md](review-guides/architecture-context.md) | PR crosses subsystems, modifies controllers, changes class hierarchies, or needs backward compat analysis |
+| [impact-analysis.md](review-guides/impact-analysis.md) | Assessing blast radius, proto changes, XXL PRs, cardinality changes |
+| [language-checklists.md](review-guides/language-checklists.md) | Any code review (pick the relevant language section) |
+| [security-review.md](review-guides/security-review.md) | PRs touching manifests, RBAC, security contexts, Dockerfiles, credentials |
+| [security-code-reviewer.md](https://github.com/anthropics/claude-code-action/blob/main/.claude/agents/security-code-reviewer.md) | Independent security scan on every PR -- OWASP Top 10, input validation, auth/authz |
+| [test-quality.md](review-guides/test-quality.md) | PRs that add/modify tests, or should include tests |
+
+## Quick checklist
+
+1. PR description explains the "why"
+2. Generated files not hand-edited
+3. Hub file changes get extra scrutiny
+4. Cross-boundary changes justified
+5. Tests proportional to change size
+6. Security addressed for manifest/RBAC/auth changes
+7. Proto changes: all language targets regenerated, test files added
+8. Code-tree impact analysis run on changed files
+9. Backward compatibility verified for API/proto/data-format changes
+10. Every new Go `err` variable checked or returned; every validation path tested
+11. For backend API/storage PRs that add/edit endpoints or SQL/query logic, use `#12999` comments and test coverage depth as a review reference
--- a/.github/review-guides/architecture-context.md
+++ b/.github/review-guides/architecture-context.md
@@ -1,0 +1,87 @@
+# Architecture Context for Reviews
+
+Load when reviewing PRs that touch multiple subsystems, modify controllers, or change class hierarchies.
+
+**See also:** [impact-analysis.md](impact-analysis.md), [language-checklists.md](language-checklists.md)
+
+---
+
+## High-level architecture
+
+```
+SDK (Python) -> compile -> IR YAML (PipelineSpec proto)
+  -> API Server (Go) -> compile -> Argo Workflow YAML
+    -> Driver (Go init container) -> resolve inputs, check cache
+      -> Launcher (Go) -> download artifacts -> invoke user container
+        -> Executor (Python) -> component function -> output artifacts
+          -> Launcher -> upload artifacts -> record in MLMD
+```
+
+## Key subsystem boundaries
+
+| Subsystem | Language | Root path |
+|-----------|----------|-----------|
+| SDK DSL/Compiler/Client/Local | Python | `sdk/python/kfp/` |
+| API Server | Go | `backend/src/apiserver/` |
+| Backend Compiler | Go | `backend/src/v2/compiler/argocompiler/` |
+| Driver | Go | `backend/src/v2/driver/` |
+| Launcher | Go | `backend/src/v2/component/` |
+| Cache Server | Go | `backend/src/cache/` |
+| Object Store | Go | `backend/src/v2/objectstore/` |
+| MLMD Client | Go | `backend/src/v2/metadata/` |
+| CRD Controllers | Go | `backend/src/crd/controller/` |
+| K8s Platform | Python+Proto | `kubernetes_platform/` |
+| Pipeline Spec | Proto | `api/v2alpha1/` |
+| Frontend | TypeScript | `frontend/` |
+
+## Critical dependency directions
+
+Flag violations of these dependency rules:
+
+```
+# Allowed direction (arrow = "may depend on"). Reverse is forbidden.
+sdk/python/kfp/compiler  -> sdk/python/kfp/dsl       (compiler may use dsl, NOT the reverse)
+sdk/python/kfp/client    -> sdk/python/kfp/dsl       (client may use dsl, NOT the reverse)
+
+# Forbidden imports
+backend/src/v2/*         must NOT import backend/src/apiserver/
+backend/src/crd/controller/* may only import backend/src/crd/pkg, NOT apiserver/
+```
+
+Full dependency map: see [code-tree.md](../../docs/agents/code-tree.md#key-module-dependencies).
+
+## Controller patterns
+
+- **No synchronous API callbacks**: Controllers must not call back to the API server. Pass data via CRD spec/annotations/labels.
+- **Idempotent reconciliation**: Each loop iteration must be safe to retry.
+- **Informer-based reads**: Read from caches, not direct API calls.
+- **New gRPC dependencies**: Require architectural justification.
+
+## Cross-language struct alignment
+
+When Go structs mirror Python dataclasses or Proto messages (e.g., `TaskKubernetesConfig`):
+
+- Add cross-reference comments in both languages
+- Proto is the source of truth
+- Flag changes to one side without corresponding changes to the other
+
+## Backward compatibility
+
+For API/proto/data-format changes:
+
+- What happens when an old backend receives new SDK output? New fields must have safe defaults.
+- What happens when a new backend reads old persisted data? Migration path must be explicit.
+- Key format changes (maps, ConfigMaps, databases) must handle old-format keys during transition.
+- Cached executions may be invalidated -- document the upgrade path.
+
+## Cardinality changes
+
+When a value changes from single to list (or vice versa):
+
+1. Use `--callers` and `--rdeps` to find all consumers
+2. Check for `[0]` indexing, `.length == 1` assumptions, nil checks
+3. Verify downstream systems (MLMD, cache keys, Argo specs)
+
+## Inheritance hierarchies
+
+Changes to base classes affect all subclasses. Verify LSP compliance. Key hierarchies in [sdk-guide.md](../../docs/agents/sdk-guide.md#artifact-types) and [code-tree.md](../../docs/agents/code-tree.md#key-inheritance-hierarchies).
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -1,0 +1,48 @@
+# Kubeflow Pipelines
+
+ML pipeline orchestration platform. Users define pipelines in Python (SDK), compile to YAML, and submit to a Kubernetes-based backend that schedules, caches, and tracks runs via MLMD.
+
+## Tech stack
+
+- **Backend**: Go (API server, persistence, scheduling) — `backend/`
+- **SDK**: Python (DSL, compiler, client) — `sdk/python/`
+- **Frontend**: React 18, MUI v5 — `frontend/`
+- **Protos**: gRPC/protobuf API definitions — `api/`, `kubernetes_platform/`
+
+## Key directories
+
+| Directory | Purpose |
+|-----------|---------|
+| `backend/src/apiserver/` | Go API server (REST + gRPC) |
+| `sdk/python/kfp/` | Python SDK: compiler, components, client |
+| `frontend/src/` | React UI |
+| `api/v2alpha1/` | Proto definitions for pipeline IR |
+| `kubernetes_platform/` | K8s-specific platform proto + Python package |
+| `test_data/` | Golden test files and sample pipelines |
+| `tools/code-tree/` | Codebase analysis tool for impact/dependency tracing |
+
+## Essential commands
+
+| Task | Command |
+|------|---------|
+| Compile pipeline | `kfp dsl compile --py pipeline.py --output pipeline.yaml` |
+| SDK tests | `pytest -v sdk/python/kfp` |
+| K8s platform tests | `pytest -v kubernetes_platform/python/test` |
+| Backend unit tests | `go test -v $(go list ./backend/... \| grep -v backend/test)` |
+| Compiler tests | `ginkgo -v ./backend/test/compiler` |
+| Frontend dev | `cd frontend && npm ci && npm start` |
+| Frontend tests | `cd frontend && npm run test:ci` |
+| Generate protos | `make -C api python && make -C api golang` |
+| Generate backend API | `make -C backend/api generate` |
+| Python lint | `yapf --recursive --diff sdk/python/ && pycln --check sdk/python && isort --check --profile google sdk/python` |
+| Go lint | `golangci-lint run` |
+| Ginkgo install | `make ginkgo && export PATH="$PWD/bin:$PATH"` |
+
+## Commit policy
+
+- Sign off all commits: `git commit -s`
+- Do not include AI agents as co-authors
+
+## Agent workflows
+
+See @AGENTS.md for task routing, PR review workflow, reference docs, and validation checklist.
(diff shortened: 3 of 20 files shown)


docs/code-tree/graph.json
hbelmiro, line 1: Can we not commit these JSON files? They are relatively big.
I'm also concerned about how to keep them up to date. We'd have to enforce that in the CI.
Could we make the generation of these files part of the review process? I mean, instead committing them, generate them locally before each code review.

tools/code-tree/code_tree.py
hbelmiro, line 1: It would be nice to have tests for this. At minimum for:

- `parse_python()` — the AST-based parser with call graph extraction
- `resolve_calls()` — the multi-strategy call resolution
- `resolve_imports()` — the import-to-file resolution (including Go module prefix stripping)
- `detect_test_relationships()` — the test-to-source mapping heuristics
- `_match_test_to_source_path()` — path-based test matching

These are the most complex functions and the ones most likely to regress.

tools/code-tree/query_graph.py
```diff
@@ -0,0 +1,1038 @@
...
+        need to be re-run or reviewed when code changes.
+        """
+        # Reuse the BFS directly — affected already contains all node IDs
+        affected, start_ids = self._compute_impact_bfs(target)
+
+        if not start_ids:
+            return {"error": f"No file or symbol found matching '{target}'"}
+
+        # Build the impact report for the summary
+        impact = self.get_impact(target)
```
hbelmiro: `get_test_impact()` calls `_compute_impact_bfs()` on line 568, then `self.get_impact()` on line 574 which calls `_compute_impact_bfs()` a second time. The entire BFS is run twice. Please either reuse the `affected` map already computed, or refactor `get_impact` to accept a pre-computed `affected` dict.

CLAUDE.md
```diff
@@ -0,0 +1,185 @@
...
+| File | Dependents |
+|------|------------|
+| `backend/src/common/util/json.go` | 659 |
+| `backend/src/common/util/time.go` | 470 |
+| `backend/src/common/util/string.go` | 303 |
+| `backend/src/common/util/error.go` | 284 |
+| `backend/src/common/util/uuid.go` | 270 |
+| `backend/src/common/util/consts.go` | 259 |
+| `backend/src/apiserver/common/utils.go` | 178 |
+| `backend/api/v*beta*/python_http_client/kfp_server_api/rest.py` | 95 each |
```
hbelmiro: The hub file counts here (e.g., `json.go`=659, `time.go`=470) are significantly different from the same files in `docs/agents/code-tree.md` (`json.go`=319, `time.go`=369) and `.github/review-guides/impact-analysis.md` (`json.go`=319, `time.go`=369). The ordering also differs.

If "Dependents" intentionally means transitive dependents (vs direct "Imported by"), please add a note clarifying the distinction. Otherwise, unify the numbers across all three files to a single source of truth.

docs/agents/frontend-guide.md
```diff
@@ -0,0 +1,160 @@
...
+   export VITE_NAMESPACE=kubeflow-user-example-com
+   npm run build
+   # Install mod-header Chrome extension for auth headers
+   npm run start:proxy-and-server
+   ```
+
+## Key technologies and architecture
+
+- **React 17** with TypeScript
+- **Material-UI v3** for components
```
hbelmiro: The guide says "Material-UI v3 for components" but `package.json` has `@mui/material: ^5.18.0`.

tools/code-tree/code_tree.py
```diff
@@ -0,0 +1,2251 @@
...
+    lines.append("|----------|-------|")
+    for lang, count in sorted(lang_counts.items(), key=lambda x: -x[1]):
+        lines.append(f"| {lang} | {count} |")
+    lines.append("")
+
+    # Directory structure
+    lines.append("## Directory Structure\n")
+    lines.append("```")
+    for d in sorted(dir_stats.keys()):
+        stats = dir_stats[d]
```
hbelmiro: `stats` is also the name of the function parameter. Here it's reassigned to `dir_stats[d]`, shadowing the parameter for the rest of the function. Can we rename this loop variable?

tools/code-tree/code_tree.py
```diff
@@ -0,0 +1,2251 @@
...
+            # Check for build files
+            if fname in BUILD_FILES:
+                rel_dir = os.path.relpath(dirpath, root)
+                if rel_dir == ".":
+                    rel_dir = ""
+                boundary = ModuleBoundary(root_dir=rel_dir, build_file=fname)
+                try:
+                    _enrich_module_boundary(abs_path, boundary)
+                except Exception:
+                    pass
```
hbelmiro: Shouldn't we handle the exception?

tools/code-tree/code_tree.py
```diff
@@ -0,0 +1,2251 @@
...
+                    continue
+            resolved.append(edge)
+        else:
+            resolved.append(edge)
+
+    # Build class/struct/interface name -> symbol for O(1) inheritance resolution
+    class_by_name = {}
+    for sym in symbols:
+        if sym.type in ("class", "struct", "interface"):
+            class_by_name[sym.name] = sym
```
hbelmiro: `class_by_name` uses a plain dict keyed by class name, so if multiple classes share the same name across different files, only the last one wins. Inheritance resolution may link to the wrong class.

tools/code-tree/code_tree.py
```diff
@@ -0,0 +1,2251 @@
...
+    return sorted(files, key=lambda x: x[1]), boundaries
+
+
+def _enrich_module_boundary(abs_path: str, boundary: ModuleBoundary):
+    """Extract module name and language from a build file."""
+    fname = boundary.build_file
+
+    if fname == "go.mod":
+        boundary.language = "go"
+        with open(abs_path, "r") as f:
```
hbelmiro: Multiple `open()` calls omit the `encoding` parameter (lines 280, 288, 297, 304, 311 for reads; 2067, 2205, 2212, 2219, 2229 for writes). Python defaults to the system locale encoding, which is not UTF-8 on all platforms. Since `read_file()` already handles encoding explicitly, these other paths should too.

```python
open(..., encoding="utf-8")
```

CLAUDE.md
hbelmiro, line 1: Several sections are duplicated between `AGENTS.md` and `CLAUDE.md`, and they already diverge in this PR (e.g. hub file numbers differ). Can we keep a single source of truth in `AGENTS.md` (the vendor-neutral entry point) and have `CLAUDE.md` link back to it for shared content?

tools/code-tree/test_query_graph.py
```diff
@@ -0,0 +1,315 @@
...
+    ]
+
+
+class TestGraphQuery(unittest.TestCase):
+    """Tests for GraphQuery using an in-memory test graph."""
+
+    @classmethod
+    def setUpClass(cls):
+        """Write test graph artifacts to a temp directory."""
+        cls.tmpdir = tempfile.mkdtemp()
```
hbelmiro: `tempfile.mkdtemp()` is used without a corresponding `tearDownClass` to clean up.

tools/code-tree/code_tree.py
```diff
@@ -0,0 +1,2251 @@
...
+    if not available:
+        try:
+            from tree_sitter_languages import get_parser, get_language
+            for lang in ["python", "go", "javascript", "typescript", "java", "rust", "ruby", "c", "cpp"]:
+                try:
+                    available[lang] = (get_parser(lang), get_language(lang))
+                except Exception:
+                    pass
+        except ImportError:
+            pass
```
hbelmiro: Are we ignoring exceptions because tree-sitter is optional or because some languages won't be available?
If so, I'd check availability first to avoid ignoring exceptions, or at least narrow the expected failure type.

tools/code-tree/code_tree.py
```diff
@@ -0,0 +1,2251 @@
...
+    type: str  # file, class, function, method, interface, struct, enum, constant, variable
+    name: str
+    file: str
+    language: str
+    line_start: int
+    line_end: int
+    scope: Optional[str] = None
+    params: list = field(default_factory=list)
+    bases: list = field(default_factory=list)
+    decorators: list = field(default_factory=list)
```
hbelmiro:
```suggestion
    params: list[str] = field(default_factory=list)
    bases: list[str] = field(default_factory=list)
    decorators: list[str] = field(default_factory=list)
```

.github/review-guides/architecture-context.md
```diff
@@ -0,0 +1,84 @@
...
+| Object Store | Go | `backend/src/v2/objectstore/` |
+| MLMD Client | Go | `backend/src/v2/metadata/` |
+| CRD Controllers | Go | `backend/src/crd/controller/` |
+| K8s Platform | Python+Proto | `kubernetes_platform/` |
+| Pipeline Spec | Proto | `api/v2alpha1/` |
+| Frontend | TypeScript | `frontend/` |
+
+## Critical dependency directions
+
+Flag violations of these rules:
```
hbelmiro:
```suggestion
Flag violations of these dependency rules:
```

.github/review-guides/architecture-context.md
```diff
@@ -0,0 +1,84 @@
...
+
+```
+sdk/python/kfp/compiler  -> sdk/python/kfp/dsl (never reverse)
+sdk/python/kfp/client    -> sdk/python/kfp/dsl (never reverse)
+backend/src/v2/*         -> never imports backend/src/apiserver/
+backend/src/crd/controller/* -> backend/src/crd/pkg only, never apiserver/
```
hbelmiro: This is confusing. It's not clear if it's the correct dependency or what should not be done (or both).

.github/copilot-instructions.md
```diff
@@ -1,44 +1,96 @@
...
-High Cohesion: Elements within a module should belong together, focusing on a single, well-defined task.
-Low Coupling: Minimize dependencies between modules to reduce the impact of changes.
-Boy Scout Rule: Always leave the code cleaner than you found it.
+- Explain the "why" behind recommendations.
+- Skip details caught by `.pre-commit-config.yaml` hooks.
+- Verify: clear docstrings, type hints, descriptive names, proper decomposition, sufficient comments.
+- Verify: new logic has tests proportional to change size; PR description explains the "why".
+- Identify design flaws, redundancy, and security issues.
+- Estimate LLM-generation likelihood with confidence level as an HTML comment in "PR Overview".
+- Comment "LGTM" when all criteria are met.
```
hbelmiro: I think we shouldn't `lgtm` PRs automatically. We could make it leave a comment that doesn't add the `lgtm` label instead.


```suggestion
- Comment "All Copilot criteria are met." when all criteria are met.
```

docs/code-tree/graph.json
droctothorpe: Agreed.

jsonmp-k8:
## PR #12812 Review: `feat: Add code-tree knowledge graph for AI-assisted development and PR reviews`

**+1,136,404 / -645** | 22 files | CI passing

### High-level take

The intent is sound — splitting a 600-line monolithic AGENTS.md into focused guides and adding a code-tree knowledge graph for smarter PR reviews. But the execution has significant problems. I agree with hbelmiro and Copilot's existing feedback. Here's what stands out:

---

### Blockers

**1. Committing ~1.13M lines of generated JSON artifacts is a non-starter**

`graph.json` (870K lines), `tags.json` (252K lines), and `modules.json` (8.4K lines) are committed as checked-in files. These will bloat the repo permanently, go stale immediately, and require a CI job to keep them in sync (which doesn't exist). These should be generated on demand, not committed. Add them to `.gitignore` and document the regeneration command.

**2. No tests for `code_tree.py` (the main 2,251-line file)**

`test_query_graph.py` covers the query side, but the actual graph builder — parsing, call resolution, import resolution, incremental mode — has zero test coverage. hbelmiro listed the exact functions that need tests. For a tool that's supposed to inform PR review decisions, this is high-risk.

**3. Duplicate BFS in `get_test_impact()`**

`query_graph.py:574` — `get_test_impact()` calls `_compute_impact_bfs()`, then calls `self.get_impact()` which runs the same BFS again. This is an O(N) BFS over the entire graph run twice for no reason. Refactor `get_impact()` to accept a pre-computed affected map, or extract the shared logic.

---

### Significant issues

**4. Incremental mode is broken**

The incremental merge logic loads previous nodes/edges for unchanged files, but the merge is fragile: it drops cross-file edges where only one endpoint changed, and the previous `call_refs` are never preserved. Running `--incremental` will produce an incomplete graph.

**5. Variable shadowing: `stats` parameter vs loop variable**

`generate_summary_md` takes a `stats` parameter, but the loop `stats = dir_stats[d]` shadows it for the rest of the iteration. This doesn't break currently, but it's a trap.

**6. `class_by_name` dict collision**

Multiple classes with the same name across files: only the last one wins. Inheritance resolution can silently link to the wrong class. Use a `defaultdict(list)` and handle ambiguity.

**7. Missing `encoding` parameter on many `open()` calls**

Many `open()` calls in reads and writes omit `encoding`. Python defaults to the system locale, which isn't UTF-8 on all platforms. The `read_file()` helper already handles this correctly — the rest should too.

**8. `docs/agents/frontend-guide.md:69` says "Material-UI v3"**

`package.json` has `@mui/material: ^5.18.0`. This is MUI v5, not v3.

**9. Hub file numbers are inconsistent across docs**

`CLAUDE.md`, `docs/agents/code-tree.md`, and `.github/review-guides/impact-analysis.md` all cite different import counts for the same files (e.g., `json.go`=659 vs 319). If the data comes from generated artifacts, there should be one source of truth, not hardcoded numbers in multiple markdown files.

---

### Moderate issues

**10. Bare `except Exception: pass` in tree-sitter loading**

Every language grammar load is wrapped in a bare `except Exception: pass`. This swallows real errors (corrupted installs, ABI mismatches). At minimum, narrow to `ImportError` or log a warning.

**11. Tempdir cleanup missing in tests**

`tempfile.mkdtemp()` in `setUpClass` with no corresponding `tearDownClass`. Use `tempfile.TemporaryDirectory` context manager or add cleanup.

**12. Regex end-line estimation is off-by-one**

The `range(lineno, min(lineno + 500, len(lines) + 1))` scan never reaches the actual last line in some edge cases, causing symbols near EOF to have truncated ranges.

**13. CLAUDE.md adds no value**

It's 5 lines that just say "go read AGENTS.md." If that's all it does, either don't add it or make AGENTS.md a symlink. hbelmiro's suggestion to keep a single source of truth in AGENTS.md and just link from CLAUDE.md is the right call — which is already what it does, so the question is whether it needs to exist at all.

**14. Copilot instructions should not auto-LGTM**

`Comment "LGTM" when all criteria are met` risks auto-adding labels that signal human approval. hbelmiro's suggestion to change this to "All Copilot criteria are met." is correct.

---

### What's good

- The split from monolithic AGENTS.md to 7 focused guides is a clear improvement for context window efficiency.
- The `code_tree.py` parser architecture (AST → tree-sitter → regex fallback) is well-designed.
- `query_graph.py` has a clean API with useful impact/blast-radius queries.
- Test file for `query_graph.py` exists with reasonable coverage of BFS, edge filtering, and determinism.
- The copilot review strategy (batch-based, context-pruning) is thoughtful.

### Verdict

Not ready to merge. The generated JSON artifacts need to be removed from the commit, `code_tree.py` needs tests, the duplicate BFS and variable shadowing bugs need fixing, and the inconsistent hardcoded numbers need a single source of truth. The docs restructuring is good but needs the factual errors fixed (MUI version, hub file counts).

tools/code-tree/query_graph.py
jsonmp-k8: **Bug: Duplicate BFS.** `_compute_impact_bfs()` is called on line 574, then `self.get_impact()` on line 580 runs the exact same BFS again. The entire graph traversal executes twice for every `get_test_impact()` call.

Refactor `get_impact()` to accept a pre-computed `affected` map, or have `get_test_impact()` build its summary directly from the first BFS result.

tools/code-tree/code_tree.py
jsonmp-k8: **Bug: Variable shadowing.** `stats` is a function parameter (the pre-computed graph stats), but this line reassigns `stats = dir_stats[d]`, shadowing the parameter for the rest of the loop body. Rename this loop variable to e.g. `directoryStats` or `directorySummary`.

tools/code-tree/code_tree.py
jsonmp-k8: **Bug: Dict collision.** `class_by_name` is a plain dict keyed by class name. If multiple classes share the same name across different files (common in a monorepo), only the last one seen wins. Inheritance resolution may silently link to the wrong class.

Use `defaultdict(list)` and disambiguate by file path or scope.

tools/code-tree/code_tree.py
```diff
@@ -0,0 +1,2251 @@
...
+                lang_obj = Language(language_func())
+                parser = Parser(lang_obj)
+                available[lang] = (parser, lang_obj)
+        except Exception:
+            pass
+
+    if not available:
+        try:
+            from tree_sitter_languages import get_parser, get_language
+            for lang in ["python", "go", "javascript", "typescript", "java", "rust", "ruby", "c", "cpp"]:
```
jsonmp-k8, line 864: **Code quality: Overly broad exception handling.** Every grammar load is wrapped in `except Exception: pass`. This swallows real errors like ABI mismatches, segfaults from incompatible tree-sitter versions, or corrupted installs.

Narrow to `(ImportError, AttributeError)` or at minimum log a warning with the exception type so debugging isn't a black hole.

tools/code-tree/code_tree.py
```diff
@@ -0,0 +1,2251 @@
...
+
+    if fname == "go.mod":
+        boundary.language = "go"
+        with open(abs_path, "r") as f:
+            for line in f:
+                if line.startswith("module "):
+                    boundary.name = line.split()[1].strip()
+                    break
+
+    elif fname == "package.json":
```
jsonmp-k8, line 287: **Bug: Missing `encoding` parameter.** This `open()` call (and several others at lines 294, 303, 310, 317 for reads; 2217, 2224, 2232, 2241 for writes) omit the `encoding` parameter. Python defaults to the system locale, which is not UTF-8 on all platforms (e.g., Windows).

The `read_file()` helper already handles encoding correctly — these other paths should use `encoding='utf-8'` consistently.

tools/code-tree/code_tree.py
```diff
@@ -0,0 +1,2251 @@
...
+                try:
+                    _enrich_module_boundary(abs_path, boundary)
+                except Exception:
+                    pass
+                boundaries.append(boundary)
+
+            # Check for source files
+            ext = Path(fname).suffix.lower()
+            lang = LANG_EXTENSIONS.get(ext)
+            if not lang:
```
jsonmp-k8, line 249: **Code quality:** Bare `except Exception: pass` here silently swallows all errors during module boundary enrichment. At minimum, narrow the exception type or log a debug message so malformed build files don't cause silent data loss in the graph.

docs/agents/frontend-guide.md
```diff
@@ -0,0 +1,162 @@
...
```
jsonmp-k8: **Factual error:** This says "Material-UI v3" but `package.json` has `@mui/material: ^5.18.0`. Should be "Material-UI v5 (MUI)".

.github/copilot-instructions.md
```diff
@@ -1,44 +1,99 @@
...
+
 ## Review style
 
-- The volume of PRs has increased significantly as a result of LLMs. Your job is to help maintain the codebase by being a tier 1 reviewer for incoming PRs. Maintainer time is limited. Your vigilance will help ensure the continuity of the project.
-- Explain the "why" behind recommendations using bullet points.
-- Don't get bogged down in small details that will be caught by the hooks listed in `.pre-commit-config.yaml` anyway.
-- Make sure that functions have clear docstrings and are well-documented.
-- Make sure that type hints are used consistently for Python code.
-- Make sure that logic is clear and easy to follow.
-- Make sure that functions are reasonably sized and properly decomposed.
```
jsonmp-k8, line 11: **Process risk:** `Comment "LGTM"` could be misread as human approval or trigger automation that adds the `lgtm` label. Agree with hbelmiro — change to something like "All Copilot criteria are met" to make it clear this is a bot assessment, not a maintainer sign-off.

tools/code-tree/test_query_graph.py
jsonmp-k8: **Test hygiene:** `tempfile.mkdtemp()` creates a temporary directory but there's no `tearDownClass` to clean it up. Use `tempfile.TemporaryDirectory()` as a context manager or add cleanup in `tearDownClass`.

CLAUDE.md
```diff
@@ -0,0 +1,5 @@
+# CLAUDE.md - Kubeflow Pipelines (KFP) Monorepo
+
+All agent and developer instructions are maintained in [AGENTS.md](AGENTS.md).
+
+Refer to AGENTS.md for: task routing, development strategy, essential commands, code-tree queries, generated files, hub files, dependency rules, validation checklist, and troubleshooting.
```
jsonmp-k8: **Questionable addition:** This file is 5 lines that say "go read AGENTS.md." It adds a maintenance surface (already diverging — hub file numbers differ between this and other docs) for no incremental value. Either remove it entirely or have it be literally just a one-liner redirect. Agree with hbelmiro that AGENTS.md should be the single source of truth.

docs/agents/frontend-guide.md
jsonmp-k8: **Factual error:** This says "Material-UI v3" but `package.json` has `@mui/material: ^5.18.0`. Should be "Material-UI v5 (MUI)".

.github/copilot-instructions.md
jsonmp-k8, line 11: **Process risk:** `Comment "LGTM"` could be misread as human approval or trigger automation that adds the `lgtm` label. Agree with hbelmiro — change to something like "All Copilot criteria are met" to make it clear this is a bot assessment, not a maintainer sign-off.

tools/code-tree/test_query_graph.py
jsonmp-k8: **Test hygiene:** `tempfile.mkdtemp()` creates a temporary directory but there's no `tearDownClass` to clean it up. Use `tempfile.TemporaryDirectory()` as a context manager or add cleanup in `tearDownClass`.

CLAUDE.md
jsonmp-k8: **Questionable addition:** This file is 5 lines that say "go read AGENTS.md." It adds a maintenance surface (already diverging — hub file numbers differ between this and other docs) for no incremental value. Either remove it entirely or keep it as a literal one-liner redirect. Agree with hbelmiro that AGENTS.md should be the single source of truth.

jeffspahr:
I'm fairly ignorant on the performance differences for AI agents using code-tree knowledge graphs vs agents files, but I don't think token savings should be the argument for this change.   Opus 4.6 just moved to a default 1 million context window, and GPT-5.4 seems to be around ~250k (At least in my environment.  I haven't found good docs on this).  All the state of the art models seem to be trending upwards.  

Our AGENTS.md is around 6,000 tokens so I'd take simplicity over optimization in this instance.  If folks feel strongly on the code-tree benefits then some more info and data on those benefits would be helpful to push that argument. 

> AGENTS.md is ~24KB / ~3,100 words. Using a standard estimate of ~0.75 tokens per word for English prose (or ~4 characters per token), that's roughly 5,500–6,000 tokens.

nsingla:
> I'm fairly ignorant on the performance differences for AI agents using code-tree knowledge graphs vs agents files, but I don't think token savings should be the argument for this change. Opus 4.6 just moved to a default 1 million context window, and GPT-5.4 seems to be around ~250k (At least in my environment. I haven't found good docs on this). All the state of the art models seem to be trending upwards.
> 
> Our AGENTS.md is around 6,000 tokens so I'd take simplicity over optimization in this instance. If folks feel strongly on the code-tree benefits then some more info and data on those benefits would be helpful to push that argument.
> 
> > AGENTS.md is ~24KB / ~3,100 words. Using a standard estimate of ~0.75 tokens per word for English prose (or ~4 characters per token), that's roughly 5,500–6,000 tokens.

The benefit of the changes are not just token reduction, these changes will also provide better context to coding agents. **Infact, with these changes, the instruction set size will increase, its just that not everything will load at once**, but when and where required. Agents will be access accurate code info, dependencies, workflows, and be able to perform much better impact analysis and take meaningful decisions. Agents will also become more performant, because they won't have to load unncessarily large context all the time, which adds more complexity and reduces the accuracy (I know 1M context models are available now, **but even so, more permutations means more chances of error**). This change is also adding windowed context, which should help with efficient subagent runs and will also improve our chances of getter PR reviews with free tier agents like copilot (which is only 64k input tokens).

If we talk about our AGENTS.md, then it may be just 6k tokens, but when its loading agents.md, system prompts, architectural diagrams, full repo code, token output (when interacting with the agent), it goes way beyond that, so the accuracy decreases as the context size increases. So this PR is also aiming at loading whatever is necessary (and that too is accurate because its based on the knowledge graph) and adding proper instructions to agents.md. This PR is also adding more instructions to test coverage, development/security standards (prompt injection will be caught) and also asking agent to always self review the changes before presenting to the use. In short, many enhancements are being added to the agents.md in this PR.

FYI: I have been using the changes in this PR and comparing it with vanilla AGENTS.md, and I am seeing better results (at least in the PR reviews). Will try it for development as well.

P.S. This does reduce cost per token and in fact improving value per token

CLAUDE.md
nsingla: Made the CLAUDE.md just a redirect to AGENT.md

.github/copilot-instructions.md
nsingla: Replaced with the suggestion

docs/code-tree/graph.json
nsingla: Removed

tools/code-tree/code_tree.py
nsingla: Done

tools/code-tree/test_query_graph.py
nsingla: Dupe

tools/code-tree/code_tree.py
```diff
@@ -0,0 +1,2259 @@
...
+    for edge in edges:
+        if edge.type != "imports":
+            resolved.append(edge)
+            continue
+
+        target = edge.target
+        if target.startswith("module:"):
+            module_name = target[7:]
+
+            # Try direct file mapping
```
jsonmp-k8, line 1567: Relative JS/TS imports are not being resolved against the importing file here. This lookup only tries the raw module string globally, so imports like `./foo` or `../providers/utils` remain unresolved `module:` edges instead of becoming local file edges. I reproduced this with a small TypeScript repo: `import { x } from "../providers/utils"` stayed as `module:../providers/utils` in the graph. That leaves a large share of frontend reverse-dependency and impact analysis incomplete.

tools/code-tree/code_tree.py
```diff
@@ -0,0 +1,2259 @@
...
+        if target_name in symbol_by_name:
+            matches.extend(symbol_by_name[target_name])
+
+    # test_ClassName_method -> ClassName.method
+    if name.startswith("test_") and "_" in name[5:]:
+        parts = name[5:].split("_", 1)
+        if len(parts) == 2:
+            class_name, method_hint = parts
+            qualified = f"{class_name}.{method_hint}"
+            if qualified in symbol_by_name:
```
jsonmp-k8, line 1468: This method-level test matcher cannot succeed with the current index. `detect_test_relationships()` populates `symbol_by_name` with bare `sym.name` keys only, but this branch looks up a qualified key like `ClassName.method`. As written, tests named in the common `test_Class_method` form will never create symbol-level `tests` edges to the target method, so `--test-impact` underreports class-based tests and falls back to coarse file-level relationships only.

tools/code-tree/code_tree.py
```diff
@@ -0,0 +1,2259 @@
...
+                        params=node.get("params", []),
+                        bases=node.get("bases", []),
+                        decorators=node.get("decorators", []),
+                        docstring=node.get("docstring"),
+                        exported=node.get("exported", True),
+                        signature=node.get("signature"),
+                        is_test=node.get("is_test", False),
+                    ))
+
+        for edge_data in prev_edges:
```
jsonmp-k8, line 2197: Incremental mode still drops live cross-file edges here. This merge keeps previous edges only when both endpoints are unchanged, so if `a.py` changes and unchanged `b.py` still imports or calls it, the old `b.py -> a.py` edges are discarded and `b.py` is never reparsed. I reproduced this with a two-file Python repo: after an incremental rerun, the prior `imports` and `calls` edges disappeared from `graph.json`. The incremental merge needs to preserve unchanged->changed and changed->unchanged edges, or recompute them from the combined graph state before writing output.

tools/code-tree/code_tree.py
```diff
@@ -0,0 +1,2259 @@
...
+        symbols, file_edges, call_refs, lc = parse_file(abs_path, rel_path, lang, ts_parsers, is_test)
+        line_counts[rel_path] = lc
+        all_symbols.extend(symbols)
+        all_edges.extend(file_edges)
+        if not args.no_call_graph:
+            all_call_refs.extend(call_refs)
+
+    log(f"Extracted {len(all_symbols)} symbols, {len(all_edges)} edges, {len(all_call_refs)} call references")
+
+    # Resolve imports
```
jsonmp-k8: Call resolution is happening after import normalization, which loses the alias information `resolve_calls()` depends on. `resolve_calls()` builds `file_imports` only from raw `symbol:` and `module:` import edges, but by the time it runs here those edges have already been rewritten to `file:` or symbol-id targets. I reproduced this with two different `foo()` definitions: `from pkg1.a import foo; foo()` produced the correct import edge, but the `calls` edge disappeared because the imported alias was no longer recoverable. This needs to resolve calls before collapsing imports, or preserve enough import metadata for call resolution.

tools/code-tree/code_tree.py
nsingla: Good catch, I will fix it

tools/code-tree/code_tree.py
nsingla: Fixing it

tools/code-tree/code_tree.py
nsingla: Thats a really good catch, let me fix it

tools/code-tree/code_tree.py
nsingla: Fixed

tools/code-tree/code_tree.py
nsingla: Done

tools/code-tree/code_tree.py
nsingla: Done

tools/code-tree/code_tree.py
nsingla: Changed the order and added unit tests as well

mprahl:
@nsingla after discussing this with other KFP maintainers, we’re concerned about the complexity and long-term maintenance cost of this approach without clear evidence that it materially improves development or PR review outcomes.

We’re also concerned about staleness: if the generated graph or derived docs fall out of date, agents may trust those artifacts instead of treating the code itself as the source of truth.

This also seems directionally consistent with recent work on repository-level context files for coding agents, which found that these kinds of context files often reduced task success rates and increased inference cost, and concluded that any such guidance should stay minimal: [Evaluating AGENTS.md: Are Repository-Level Context Files Helpful for Coding Agents?](https://arxiv.org/abs/2602.11988).

What we are in favor of is keeping `AGENTS.md` lean and using it as an entry point that links to more focused guides by area of the monorepo, for example SDK, frontend, backend, testing/CI, etc.

If you find the code-tree approach helpful in your own development workflow, a Kubeflow blog post or external write-up on how to set it up and use it effectively would be a great way to share that without making it part of the repo’s standard workflow!

PR #12817, opened 2026-02-14
feat(frontend): migrate API codegen to OpenAPI Generator

### LINKED ISSUE #12775: Migrate frontend API codegen from swagger-codegen to openapi-generator and automate in CI
## What

The frontend TypeScript API clients (`src/apis/`, `src/apisv2beta1/`) are generated using [swagger-codegen-cli 2.4.7](https://repo1.maven.org/maven2/io/swagger/swagger-codegen-cli/2.4.7/), which is effectively unmaintained. The generated code requires a post-processing fix script (`frontend/scripts/fix_swagger_codegen_delete.js`) to produce valid TypeScript 4.0+ output. Additionally, code generation is a manual developer-only step with no CI enforcement, meaning committed generated code can drift from the protobuf/swagger specs.

## Why

1. **swagger-codegen 2.x is unmaintained.** The actively maintained successor is [openapi-generator](https://openapi-generator.tech/), which was forked from swagger-codegen and has progressive improvements including fixes for TypeScript compatibility issues.

2. **The `delete` operator bug requires a workaround.** swagger-codegen 2.x generates `delete localVarUrlObj.search;`, which is a TypeScript 4.0+ compile error (`TS2790: The operand of a 'delete' operator must be optional`). This was fixed in openapi-generator ([PR #6960](https://github.com/OpenAPITools/openapi-generator/pull/6960)), but not in swagger-codegen 2.x. We currently work around this with a post-processing script that rewrites `delete` to `= null`. See also upstream issues: [#7282](https://github.com/OpenAPITools/openapi-generator/issues/7282), [#7865](https://github.com/OpenAPITools/openapi-generator/issues/7865).

3. **No CI validation of generated code.** The jar is manually downloaded by developers per the CONTRIBUTING.md instructions. `npm run apis` is run locally and the output is committed. There is no CI step to verify that committed generated code matches what the current specs would produce, so drift is possible.

4. **Prior art.** Issue #3871 migrated the Python SDK client from swagger-codegen to openapi-generator. The frontend TypeScript clients were not included in that effort.

### Phase 1: Migrate to openapi-generator
- Replace `swagger-codegen-cli.jar` with `openapi-generator-cli` (available via npm: `@openapitools/openapi-generator-cli`, or as a jar)
- Update `package.json` scripts from `swagger-codegen-cli.jar generate -l typescript-fetch` to `openapi-generator-cli generate -g typescript-fetch`
- Verify generated output compiles cleanly under TypeScript 4.9+ with `strict: true` without the fix script
- Remove `frontend/scripts/fix_swagger_codegen_delete.js` if no longer needed
- Validate that all consuming code still works with the new generated types/functions (may require import adjustments)
- Reference the [migration guide](https://openapi-generator.tech/docs/swagger-codegen-migration/)

### Phase 2: Automate codegen in CI
- Add a CI step that regenerates the API clients from the current swagger specs
- Compare the regenerated output against the committed files
- Fail the build if they differ, ensuring specs and generated code stay in sync
- This removes the dependency on developers remembering to run `npm run apis` after spec changes

### Risks
- openapi-generator's `typescript-fetch` template may produce different function signatures or types than swagger-codegen, requiring updates to consuming code
- Should be done as a standalone effort, not bundled with other migrations

jeffspahr:
## Summary
- migrate frontend API client generation from Swagger Codegen to OpenAPI Generator for issue #12775
- add pinned Docker-based generator script (`openapitools/openapi-generator-cli:v7.19.0`) and wire all `npm run apis*` scripts to it
- regenerate v1/v2beta1 frontend + server auth clients in `typescript-fetch` layout and remove legacy swagger artifacts/postprocess script
- update frontend/server integrations for new generated API surface (constructors, enum members, method names)
- add frontend CI drift check to regenerate APIs and fail if generated clients are out of date
- update frontend/docs contributor guidance for the new codegen workflow

## Why
- removes dependency on legacy `swagger-codegen-cli.jar` flow
- eliminates the brittle `localVarUrlObj.search` postprocess patch path
- makes generation deterministic and CI-enforced

## Verification
- `npm ci` (frontend)
- `npm run typecheck` (frontend)
- `npm run lint` (frontend)
- `node frontend/scripts/generate_openapi_typescript_fetch.js v1 v2beta1`
- determinism check: reran generation and confirmed generated-path status hash unchanged

Closes #12775

### DIFF
--- a/frontend/server/package-lock.json
+++ b/frontend/server/package-lock.json
@@ -29,7 +29,7 @@
         "@types/tar": "^4.0.3",
         "@types/tar-stream": "^1.6.1",
         "@vitest/coverage-v8": "^4.0.18",
-        "prettier": "1.19.1",
+        "prettier": "^3.8.1",
         "supertest": "^4.0.2",
         "tslint": "^5.20.1",
         "typescript": "^4.9.5",
@@ -5677,16 +5677,19 @@
       }
     },
     "node_modules/prettier": {
-      "version": "1.19.1",
-      "resolved": "https://registry.npmjs.org/prettier/-/prettier-1.19.1.tgz",
-      "integrity": "sha512-s7PoyDv/II1ObgQunCbB9PdLmUcBZcnWOcxDh7O0N/UwDEsHyqkW+Qh28jW+mVuCdx7gLB0BotYI1Y6uI9iyew==",
+      "version": "3.8.1",
+      "resolved": "https://registry.npmjs.org/prettier/-/prettier-3.8.1.tgz",
+      "integrity": "sha512-UOnG6LftzbdaHZcKoPFtOcCKztrQ57WkHDeRD9t/PTQtmT0NHSeWWepj6pS0z/N7+08BHFDQVUrfmfMRcZwbMg==",
       "dev": true,
       "license": "MIT",
       "bin": {
-        "prettier": "bin-prettier.js"
+        "prettier": "bin/prettier.cjs"
       },
       "engines": {
-        "node": ">=4"
+        "node": ">=14"
+      },
+      "funding": {
+        "url": "https://github.com/prettier/prettier?sponsor=1"
       }
     },
     "node_modules/process-nextick-args": {
--- a/frontend/server/package.json
+++ b/frontend/server/package.json
@@ -27,7 +27,7 @@
     "@types/tar": "^4.0.3",
     "@types/tar-stream": "^1.6.1",
     "@vitest/coverage-v8": "^4.0.18",
-    "prettier": "1.19.1",
+    "prettier": "^3.8.1",
     "supertest": "^4.0.2",
     "tslint": "^5.20.1",
     "typescript": "^4.9.5",


frontend/server/app.ts
```diff
@@ -83,11 +87,34 @@ export class UIServer {
...
+    const server = this.httpServer;
+    if (!server) {
+      return this;
     }
-    this.httpServer = undefined;
+
+    this.closePromise = new Promise<void>((resolve, reject) => {
+      server.close(err => {
+        const errorCode = (err as NodeJS.ErrnoException | undefined)?.code;
+        if (err && errorCode !== 'ERR_SERVER_NOT_RUNNING') {
```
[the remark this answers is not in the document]
jeffspahr: Given this repo baseline (`frontend/.nvmrc`) is Node `v22.19.0`, `ERR_SERVER_NOT_RUNNING` is available in supported versions. Also, broadening the guard to ignore undefined error codes could swallow non-node errors and reduce observability. Keeping the explicit `ERR_SERVER_NOT_RUNNING` check as implemented.

frontend/scripts/generate_openapi_typescript_fetch.js
```diff
@@ -0,0 +1,246 @@
+#!/usr/bin/env node
+'use strict';
+
+const fs = require('fs');
+const path = require('path');
+const { spawnSync } = require('child_process');
+
+// Keep this pinned to the latest stable non-snapshot release.
```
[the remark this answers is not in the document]
jeffspahr: Good suggestion. We currently keep the inline comment short and document the update flow in `frontend/CONTRIBUTING.md` and `AGENTS.md` (pin to stable non-snapshot, regenerate, verify). I am leaving this as-is for now to keep the migration diff focused, but we can expand the inline policy text in a follow-up if maintainers prefer.

frontend/server/app.ts
```diff
@@ -70,10 +71,13 @@ export class UIServer {
...
    */
   start(port?: number | string) {
+    if (this.closePromise) {
+      throw new Error('UIServer is closing.');
+    }
     if (this.httpServer) {
       throw new Error('UIServer already started.');
     }
-    port = port || this.options.server.port;
+    port = port ?? this.options.server.port;
```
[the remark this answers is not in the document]
jeffspahr: Thanks. We intentionally use nullish coalescing here so `app.start(0)` remains valid for dynamic port allocation. This behavior is covered by `frontend/server/app.test.ts` (`allows restarting on the same port after close resolves`), which explicitly starts with port `0`. Treating `0` as fallback would regress that behavior.

frontend/scripts/generate_openapi_typescript_fetch.js
```diff
@@ -0,0 +1,324 @@
...
+  },
+  'v2beta1:visualization': {
+    spec: 'backend/api/v2beta1/swagger/visualization.swagger.json',
+    output: 'frontend/src/apisv2beta1/visualization',
+  },
+  'v2beta1:auth': {
+    spec: 'backend/api/v2beta1/swagger/auth.swagger.json',
+    output: 'frontend/server/src/generated/apisv2beta1/auth',
+  },
+};
```
manaswinidas, line 81: **DRY**: All 14 entries follow one of two templates (`v1beta1/swagger/{name}` → `src/apis/{name}`, `v2beta1/swagger/{name}` → `src/apisv2beta1/{name}`) with only `auth` and `recurringrun` as exceptions. Consider generating this map programmatically from version + name arrays to avoid maintaining 80 near-identical lines.

frontend/scripts/generate_openapi_typescript_fetch.js
```diff
@@ -0,0 +1,324 @@
...
+    if (!entry.isFile() || !entry.name.endsWith('.ts')) {
+      continue;
+    }
+    const filePath = path.join(apisDir, entry.name);
+    const original = fs.readFileSync(filePath, 'utf8');
+    let updated = original;
+
+    updated = updated.replace(/\b([A-Z][A-Za-z0-9_]*)V1Request\b/g, '$1Request');
+    updated = updated.replace(/\b([a-z][A-Za-z0-9_]*)V1Raw\b/g, '$1Raw');
+    updated = updated.replace(/\b([a-z][A-Za-z0-9_]*)V1\b/g, '$1');
```
manaswinidas: **Fragile regex**: `\b([a-z][A-Za-z0-9_]*)V1\b` matches *any* lowercase-starting identifier ending in `V1`, not just method names. A generated variable like `apiVersionV1` or `endpointV1` would be silently renamed to `apiVersion` / `endpoint`. Safer to target only known OpenAPI-generated method/request patterns, or use the generator's `--name-mappings` option instead.

frontend/scripts/generate_openapi_typescript_fetch.js
```diff
@@ -0,0 +1,324 @@
...
+      updated = updated.replace(/WindowOrWorkerGlobalScope\['fetch'\]/g, 'typeof fetch');
+      updated = updated.replace(/\bRequestCredentials\b/g, "RequestInit['credentials']");
+      // Keep the generated runtime compatible with Node-only type libs.
+      updated = updated.replace(
+        /export type FetchAPI = typeof fetch;/g,
+        'export type FetchAPI = (input: string, init: RequestInit) => Promise<Response>;',
+      );
+      updated = updated.replace(
+        /private fetchApi = async \(url: string, init: RequestInit\) => {/g,
+        'private fetchApi = async (url: string, init: RequestInit): Promise<Response> => {',
```
manaswinidas: **Bundle/tree-shaking concern**: Stripping `import type` → `import` defeats TypeScript's type-only import elision — types that should be erased at compile time become runtime imports, increasing bundle size and risking circular-dependency issues.

Similarly, stripping `override` (line below) removes a safety mechanism that catches accidental method signature mismatches on TS upgrades.

Both of these, plus the `WindowOrWorkerGlobalScope`/`RequestCredentials` rewrites below, are fragile string patches over generated output. If the generator template changes whitespace or phrasing, they silently stop matching. Prefer configuring the OpenAPI Generator template or `tsconfig` settings (e.g. `verbatimModuleSyntax: false`) to avoid post-hoc regex surgery.

frontend/scripts/generate_openapi_typescript_fetch.js
```diff
@@ -0,0 +1,324 @@
...
+}
+
+function main() {
+  const scriptDir = __dirname;
+  const repoRoot = path.resolve(scriptDir, '..', '..');
+  const args = process.argv.slice(2);
+  const targets = resolveTargets(args);
+
+  ensureDockerAvailable();
+  for (const targetKey of targets) {
```
manaswinidas, line 339: **Performance**: Each of the 14 targets spawns a separate `docker run` (JVM startup + container creation). On a cold cache this can be very slow. Consider:
- Using OpenAPI Generator's batch mode to process multiple specs in one container invocation, or
- Running targets in parallel with async `spawn` instead of `spawnSync`, or
- At minimum, adding a `console.time`/`timeEnd` so developers can see the cost.

.github/workflows/frontend.yml
```diff
@@ -33,6 +35,16 @@ jobs:
     - name: Install dependencies
       run: cd ./frontend && npm ci
 
+    - name: Verify generated API clients are up to date
+      run: |
+        cd ./frontend
+        npm run apis
+        npm run apis:v2beta1
```
manaswinidas: **Perf / DRY**: These two commands together spawn 14 Docker containers. The script already accepts multiple groups in one invocation. Simplify to:
```yaml
node scripts/generate_openapi_typescript_fetch.js all
```
This also avoids the subtle issue where `npm run apis` alone regenerates v1 + formats, then `npm run apis:v2beta1` regenerates v2beta1 + formats again (Prettier runs twice).

frontend/package.json
```diff
@@ -47,23 +47,22 @@
...
+    "apis:visualization": "node scripts/generate_openapi_typescript_fetch.js v1:visualization",
+    "apis:auth": "node scripts/generate_openapi_typescript_fetch.js v1:auth",
+    "apis:v2beta1": "node scripts/generate_openapi_typescript_fetch.js v2beta1",
+    "apis:v2beta1:experiment": "node scripts/generate_openapi_typescript_fetch.js v2beta1:experiment",
+    "apis:v2beta1:recurringrun": "node scripts/generate_openapi_typescript_fetch.js v2beta1:recurringrun",
+    "apis:v2beta1:pipeline": "node scripts/generate_openapi_typescript_fetch.js v2beta1:pipeline",
+    "apis:v2beta1:run": "node scripts/generate_openapi_typescript_fetch.js v2beta1:run",
+    "apis:v2beta1:filter": "node scripts/generate_openapi_typescript_fetch.js v2beta1:filter",
+    "apis:v2beta1:visualization": "node scripts/generate_openapi_typescript_fetch.js v2beta1:visualization",
+    "apis:v2beta1:auth": "node scripts/generate_openapi_typescript_fetch.js v2beta1:auth",
```
manaswinidas, line 67: **DRY**: 16 script entries that each just call `node scripts/generate_openapi_typescript_fetch.js <target>`. The script already accepts arbitrary target args directly. Consider keeping only `apis`, `apis:v2beta1`, and `apis:all`, and documenting that specific targets can be passed via `node scripts/... v1:experiment` directly. The per-target aliases add maintenance burden without adding discoverability.

frontend/scripts/generate_openapi_typescript_fetch.js
```diff
@@ -0,0 +1,324 @@
...
+    GLOBAL_PROPERTIES,
+  ];
+
+  console.log(`Generating ${targetKey} -> ${target.output}`);
+  runCommand('docker', dockerArgs);
+  removeGeneratorMetadata(outputPath);
+  if (targetKey.startsWith('v1:')) {
+    normalizeV1ApiMethodNames(outputPath);
+  }
+  normalizeGeneratedTypeScriptForLegacyTooling(outputPath, {
```
manaswinidas: **Double I/O pass**: For v1 targets, `normalizeV1ApiMethodNames` reads + writes every `.ts` file in `apis/`, then `normalizeGeneratedTypeScriptForLegacyTooling` immediately reads + writes the same files again. Merge both normalization passes into a single read-transform-write to halve the file I/O.

frontend/scripts/generate_openapi_typescript_fetch.js
jeffspahr: Keeping `SPEC_TARGETS` explicit here. The `auth` server targets and the `recurring_run` naming exception reduce the value of generating the map programmatically, and the literal table stays easier to audit when adding or moving surfaces.

frontend/scripts/generate_openapi_typescript_fetch.js
jeffspahr: Fixed in the latest push. The rename now only targets generated request/raw symbols plus method-call sites, and there is a regression test that keeps `apiVersionV1` unchanged.

frontend/scripts/generate_openapi_typescript_fetch.js
jeffspahr: Kept the fetch-type rewrites because `frontend/server` compiles without DOM libs, and kept the `import type`/`override` compatibility rewrite only because the repo is still pinned to Prettier 1.19.1 today. The normalization is now isolated in one place and covered by targeted tests; once the Prettier 3 upgrade lands, the Prettier-specific rewrite can be removed cleanly.

frontend/scripts/generate_openapi_typescript_fetch.js
jeffspahr: Agreed that batching/parallelism is the next obvious performance lever, but I am not taking that on in this PR. The latest push does remove the duplicate workflow invocation/formatting pass; I would rather treat container batching as a separate follow-up than expand this migration diff further.

.github/workflows/frontend.yml
jeffspahr: Fixed in the latest push. The workflow now uses the single `npm run apis:all` entrypoint, so CI no longer regenerates in two passes or runs Prettier twice.

frontend/package.json
jeffspahr: Keeping the per-target npm aliases for contributor ergonomics. The direct `node frontend/scripts/generate_openapi_typescript_fetch.js <target>` path is now documented as well, but the aliases are still useful for quick targeted regeneration and change rarely.

frontend/scripts/generate_openapi_typescript_fetch.js
jeffspahr: Fixed in the latest push. The normalization logic now runs as a single read/transform/write pass, so v1 renames and legacy-tooling compatibility updates no longer hit the same files twice.

jeffspahr:
Addressed review summary:

Fixed in code:
- Direct repo-root invocation now resolves the frontend-local Prettier binary instead of relying on `PATH`.
- CI now uses the single `apis:all` entrypoint, so generation/formatting no longer runs in two passes.
- The normalization flow is now a single read/transform/write pass.
- The V1 rename was tightened to generated request/raw symbols plus method-call sites, with regression coverage to keep unrelated identifiers such as `apiVersionV1` unchanged.
- Contributor docs now document direct repo-root invocation.

Verified / addressed operationally:
- The branch was rebased onto current `master`.
- Generated-client constructor/call-site compatibility was rechecked with successful regeneration and targeted tests.

Intentionally deferred / kept as-is:
- `SPEC_TARGETS` remains explicit rather than generated programmatically.
- Per-target npm aliases remain for contributor ergonomics.
- The sequential `docker run` performance concern is deferred to a follow-up rather than expanding this migration further.
- The `import type` / `override` compatibility rewrite is kept temporarily because the repo is still pinned to Prettier 1.19.1 today; once the Prettier 3 upgrade lands, that rewrite can be removed cleanly.

So the final state is: all review threads are addressed/closed, but not every suggestion turned into a code change.

manaswinidas (approved):
/lgtm - nice migration. One follow-up suggestion: track removing the `import type` / `override` stripping in `normalizeGeneratedTypeScriptSource()` once the Prettier 3 upgrade lands, and consider batching/parallelizing the 14 sequential Docker invocations in a separate PR to improve generation speed.

droctothorpe:
/approve
/lgtm

Thanks, Jeff! The LOCs on these frontend PRs have been absolutely bananas.

PR #12818, opened 2026-02-14
fix: Handle literal dict items in ParallelFor (Issue #12191)

### LINKED ISSUE #12191: Error “Failed to parseJson from string” while running sample code from KFP official doc[1]
Issue Description : When using dsl.ParallelFor in the KFP pipeline mentioned in the sample[3], specifically when iterating over a list of dictionaries (the same example as is), encountering error [2] 'Failed to parseJson from string', indicating an issue with how the input data is being handled.

Workaround: The suggested workaround is to either use json.dumps() to convert the dictionary to a JSON string or pass the list of dictionaries as a pipeline parameter. Below examples work (only as workaround)

Example:

```python
@dsl.pipeline
def dummy_pipeline2():
    with dsl.ParallelFor(
        items=[json.dumps({'a': 1, 'b': 10}), json.dumps({'a': 2, 'b': 20})],
        parallelism=1
    ) as item:
        task1 = my_component(number=item.a)
        task2 = my_component(number=item.b)
```

Or 
```python
@dsl.pipeline
def dummy_pipeline2(loop_items: list = [{'a': 1, 'b': 10}, {'a': 2, 'b': 20}]):
    with dsl.ParallelFor(
        items=loop_items,
        parallelism=1
    ) as item:
        task1 = my_component(number=item.a)
        task2 = my_component(number=item.b)
```

Recommendation: The issue is a known limitation, and this specific issue has been raised again to fix the same. The sample code is failing when using the code as per the public docs. The issue may be a regression within the open-source community's code. 

Reference git issues: 
https://github.com/kubeflow/pipelines/issues/9386#issuecomment-1974849876
https://github.com/kubeflow/pipelines/issues/9366#issue-1702436413

[1] - https://kubeflow-pipelines.readthedocs.io/en/stable/source/dsl.html#kfp.dsl.ParallelFor

[2] - 'Failed to evaluate the expression with error: INVALID_ARGUMENT: Failed to parseJson from string.; Failed to evaluate the parameter_expression_selector.'

[3] - 
```python
#! /usr/bin/env python


from datetime import datetime

import google.cloud.aiplatform as aip
from kfp import dsl, compiler

@dsl.component(base_image="python:3.12")
def my_component(number: int):
    print(number)

@dsl.pipeline
def dummy_pipeline():
    with dsl.ParallelFor(
        items=[{'a': 1, 'b': 10}, {'a': 2, 'b': 20}],
        parallelism=1
    ) as item:
        task1 = my_component(number=item.a)
        task2 = my_component(number=item.b)

if __name__ == "__main__":
    compiler.Compiler().compile(pipeline_func=dummy_pipeline, package_path="dummy-pipeline.yaml")

    ts = datetime.now().strftime("%m%d-%H%M%S")
    job = aip.PipelineJob(
        display_name=f"dummy-pipeline--{ts}",
        template_path="dummy-pipeline.yaml",
        job_id=f"dummy-pipeline--{ts}",
        pipeline_root="gs://<any-bucket>",  # give a specific existing bucket
        project="<PROJECT_ID>",  # give a valid GCP project
        location="<ANY_REGION>",  # give a valid region to execute this pipeline
    )
    job.submit()
```

Shivam140802:
```markdown
## Fix: Handle literal dict items in ParallelFor (Issue #12191)

### Problem
ParallelFor with literal dict items was failing at runtime. This is a common use case, but the compiler wasn't handling it correctly:

```python
with dsl.ParallelFor(items=[{'a': 1, 'b': 10}, {'a': 2, 'b': 20}]) as item:
    process_item(item_a=item.a, item_b=item.b)
```

### What Was Happening
After investigating the compiler code, I found that:
- Type inference was working correctly - it identified the parameter type as `STRUCT`
- But the items were being serialized as a JSON object array: `[{...}, {...}]`
- The runtime expression engine uses `parseJson(string_value)["field"]` to access nested fields
- This expects string values, not objects - which is where the mismatch occurs

### The Fix
I added logic to detect when items are dicts and convert each one to a JSON string before the outer serialization. This ensures the runtime gets the data in the format it expects.

The change is minimal (7 lines) and only affects this specific case:
```python
if isinstance(raw_values, list) and raw_values and isinstance(raw_values[0], dict):
    raw_values = [json.dumps(item, sort_keys=True) for item in raw_values]
```

### Testing & Validation
- Added a regression test to catch this issue if it comes back
- Verified all existing ParallelFor tests still pass
- Tested that the json.dumps() workaround still produces identical results (backward compatibility)
- Manually verified the generated YAML is correct

### Changes
- pipeline_spec_builder.py (7 lines added)
- compiler_test.py (28 lines added for regression test)

Fixes #12191


Shivam140802:
@zazulam looking forward for review

droctothorpe:
Hi @Shivam140802, thank you for your PR! The linked issue (#12191) must be labeled `ready` before your contribution can be reviewed. Please check our [CONTRIBUTING.md](https://github.com/kubeflow/pipelines/blob/master/CONTRIBUTING.md) for details.

PR #12829, opened 2026-02-15
refactor(frontend): replace react-vis with recharts

jeffspahr:
## Summary
- replace the ROC curve viewer from react-vis to recharts
- pin recharts transitive react-redux to 8.1.3 for React 17 compatibility
- improve ROC tooltip behavior by selecting nearest display point and add regression coverage
- sync the React 17 checklist and remove stale vite migration docs

## Validation
- npm run test:ui -- src/components/viewers/ROCCurve.test.tsx
- npm run lint
- npm run typecheck
- npm run test:server:coverage
- npm run test:ui:coverage (116 files, 1646 tests passed)

### DIFF
--- a/frontend/src/components/viewers/ROCCurve.tsx
+++ b/frontend/src/components/viewers/ROCCurve.tsx
@@ -66,9 +66,6 @@ const css = stylesheet({
     marginRight: 6,
     width: 10,
   },
-  root: {
-    margin: 'auto',
-  },
 });
 
 // Used the following color palette, with some slight brightness modifications, as reference:
@@ -192,7 +189,6 @@ class ROCCurve extends Viewer<ROCCurveProps, ROCCurveState> {
           width={width}
           height={height}
           data={chartData}
-          className={css.root}
           margin={{ top: 5, right: 5, bottom: 20, left: 32 }}
           onMouseDown={
             !isSmall
--- a/frontend/src/components/viewers/__snapshots__/ROCCurve.test.tsx.snap
+++ b/frontend/src/components/viewers/__snapshots__/ROCCurve.test.tsx.snap
@@ -4,7 +4,7 @@ exports[`ROCCurve > does not break on empty data 1`] = `
 <DocumentFragment>
   <div>
     <div
-      class="recharts-wrapper root_ftu9a9a"
+      class="recharts-wrapper"
       height="520"
       style="position: relative; cursor: default; width: 800px; height: 520px;"
       width="800"
@@ -229,7 +229,7 @@ exports[`ROCCurve > does not break on no config 1`] = `
 <DocumentFragment>
   <div>
     <div
-      class="recharts-wrapper root_ftu9a9a"
+      class="recharts-wrapper"
       height="520"
       style="position: relative; cursor: default; width: 800px; height: 520px;"
       width="800"
@@ -454,7 +454,7 @@ exports[`ROCCurve > renders a simple ROC curve given one config 1`] = `
 <DocumentFragment>
   <div>
     <div
-      class="recharts-wrapper root_ftu9a9a"
+      class="recharts-wrapper"
       height="520"
       style="position: relative; cursor: default; width: 800px; height: 520px;"
       width="800"
@@ -2355,7 +2355,7 @@ exports[`ROCCurve > renders an ROC curve using three configs 1`] = `
 <DocumentFragment>
   <div>
     <div
-      class="recharts-wrapper root_ftu9a9a"
+      class="recharts-wrapper"
       height="520"
       style="position: relative; cursor: default; width: 800px; height: 520px;"
       width="800"
--- a/frontend/src/components/viewers/__snapshots__/ViewerContainer.test.tsx.snap
+++ b/frontend/src/components/viewers/__snapshots__/ViewerContainer.test.tsx.snap
@@ -139,7 +139,7 @@ exports[`ViewerContainer > renders a viewer of type ROC 1`] = `
 <DocumentFragment>
   <div>
     <div
-      class="recharts-wrapper root_ftu9a9a"
+      class="recharts-wrapper"
       height="520"
       style="position: relative; cursor: default; width: 800px; height: 520px;"
       width="800"
--- a/frontend/src/pages/__snapshots__/CompareV1.test.tsx.snap
+++ b/frontend/src/pages/__snapshots__/CompareV1.test.tsx.snap
@@ -2078,7 +2078,7 @@ exports[`CompareV1 > creates an extra aggregation plot for compatible viewers 1`
             </div>
             <div>
               <div
-                class="recharts-wrapper root_ftu9a9a"
+                class="recharts-wrapper"
                 height="260"
                 style="position: relative; cursor: default; width: 400px; height: 260px;"
                 width="400"
@@ -2366,7 +2366,7 @@ exports[`CompareV1 > creates an extra aggregation plot for compatible viewers 1`
             </div>
             <div>
               <div
-                class="recharts-wrapper root_ftu9a9a"
+                class="recharts-wrapper"
                 height="260"
                 style="position: relative; cursor: default; width: 400px; height: 260px;"
                 width="400"
@@ -2623,7 +2623,7 @@ exports[`CompareV1 > creates an extra aggregation plot for compatible viewers 1`
             </div>
             <div>
               <div
-                class="recharts-wrapper root_ftu9a9a"
+                class="recharts-wrapper"
                 height="260"
                 style="position: relative; cursor: default; width: 400px; height: 260px;"
                 width="400"


jeffspahr:
Follow-up update from review feedback:

- Added an identity-based cache for ROC chart preprocessing in ROCCurve so buildChartData is not recomputed on every zoom/hover rerender.
- This specifically reduces repeated Map/Set/sort work during drag interactions, while keeping the migration scope small.
- Cache key is configs reference identity (documented in frontend/docs/react-17-upgrade-checklist.md).

Also included in this push:
- semantic reset zoom button (button type="button")
- safer tooltip fallback parsing
- removed unused label{index} chart row fields
- edge-case tests for findNearestDisplayPoint (empty/single point)

Validation rerun:
- npm run test:ui -- src/components/viewers/ROCCurve.test.tsx (14 passed)
- npm run lint -- src/components/viewers/ROCCurve.tsx

frontend/src/components/viewers/ROCCurve.tsx
```diff
@@ -116,130 +146,199 @@ class ROCCurve extends Viewer<ROCCurveProps, ROCCurveState> {
...
-    const baseLineData = Array.from(Array(100).keys()).map(x => ({ x: x / 100, y: x / 100 }));
-
-    const { hoveredValues, lastDrawLocation, highlightIndex } = this.state;
+    const { lastDrawLocation, highlightIndex, refAreaLeft, refAreaRight } = this.state;
+    const chartData = this.buildChartData(datasets);
+    const xDomain = lastDrawLocation
+      ? ([lastDrawLocation.left, lastDrawLocation.right] as [number, number])
+      : ([0, 1] as [number, number]);
+    const colors = this.props.colors || lineColors;
+    const showLegend = this.props.forceLegend || datasets.length > 1;
```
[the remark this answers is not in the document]
jeffspahr: Fixed in commit 294caef0d. Added identity-based memoization for ROC chart preprocessing via `cachedConfigs` and `cachedChartData`, and render now calls `getChartData(configs, datasets)` so zoom/hover rerenders do not rebuild chart data each time.

frontend/src/components/viewers/ROCCurve.tsx
```diff
@@ -116,130 +146,199 @@ class ROCCurve extends Viewer<ROCCurveProps, ROCCurveState> {
...
 
-          {lastDrawLocation && <span>Click to reset zoom</span>}
+          {lastDrawLocation && (
+            <span
+              role='button'
+              onClick={() => this.setState({ lastDrawLocation: null })}
+              style={{ cursor: 'pointer' }}
+            >
+              Click to reset zoom
+            </span>
```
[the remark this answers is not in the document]
jeffspahr: Fixed in commit 294caef0d. The reset control now uses a semantic button element (`<button type="button">`) instead of `span role="button"`, which addresses keyboard/accessibility behavior.

jeffspahr:
CI follow-up: fixed frontend lockfile drift under CI toolchain.

Root cause was npm version skew (`npm ci` passes locally on npm 11 but failed in CI on npm 10.9.3 with `Missing: babel-plugin-macros@3.1.0 from lock file`).

Fix:
- regenerated `frontend/package-lock.json` under Node 22.19.0 / npm 10.9.3
- verified `npm ci` succeeds locally under that same version pair

Commit: 1072ae21a

jeffspahr:
compare-seeded-roc-1280x800
artifact-lineage-from-list-1280x800

Attaching compare screenshots from the 2 screens that use charts.  Note that there are some changes in the line graph, but I got it as close as reasonably possible.

srtanish1992:
@jeffspahr I did some manual testing and found few issues. I have attached screenshots of master vs branch.

**ROC curve under metrics section on Compare runs page.**

1. The grid lines are spaced out due to the intervals on y axis.
2. The text inside the tooltip is not visible.
3. When I hover over the series legend color blue, I see the red color as well. Same happens for blue.
4. The series legend icon is different.
5. The y axis title is not completely visible.

**Lineage explorer graph.**

1. The edges don't connect correctly.


Screenshot 2026-02-17 at 9 32 30 PM
Screenshot 2026-02-17 at 9 35 19 PM
Screenshot 2026-02-17 at 9 37 40 PM
Screenshot 2026-02-17 at 9 37 20 PM

jeffspahr:
Thanks for the review @srtanish1992!  I forgot to push the latest commit that my screenshots came from.  Please take another look, and I'll start looking into the issues you called out as well.

srtanish1992:
> Thanks for the review @srtanish1992! I forgot to push the latest commit that my screenshots came from. Please take another look, and I'll start looking into the issues you called out as well.

It's looking much better now 🎉 . The only minor issue is that the last data point/tick is missing on the x and y axis i.e `1` in this case.

Screenshot 2026-02-17 at 10 59 57 PM

frontend/src/components/viewers/ROCCurve.tsx
srtanish1992: I don't see `css.axis` used anywhere. Should we remove this class?

frontend/src/components/viewers/ROCCurve.tsx
```diff
@@ -115,131 +150,234 @@ class ROCCurve extends Viewer<ROCCurveProps, ROCCurveState> {
...
-            color={color.disabledBg}
+          <CartesianGrid />
+          <XAxis
+            type='number'
+            dataKey='x'
+            domain={xDomain}
+            ticks={this.xTicks}
+            interval={0}
+            tickFormatter={(value: number) => value.toFixed(2)}
+            tick={{ fill: '#666', fontSize: 12, fontWeight: 400 }}
```
srtanish1992: Should we use tokens from `Css.tsx` file instead of hardcoded values?

frontend/src/components/viewers/ROCCurve.tsx
```diff
@@ -87,19 +115,26 @@ interface ROCCurveProps {
...
   lastDrawLocation: { left: number; right: number } | null;
   highlightIndex: number;
+  refAreaLeft: number | null;
+  refAreaRight: number | null;
 }
 
 class ROCCurve extends Viewer<ROCCurveProps, ROCCurveState> {
+  private cachedConfigs: ROCCurveConfig[] | null = null;
+  private cachedChartData: Array<Record<string, number | null>> = [];
+  private readonly xTicks = Array.from({ length: 20 }, (_, i) => Number((i * 0.05).toFixed(2)));
```
srtanish1992, line 149: Changing length to 21 will fix the issue of including `1`.

frontend/src/components/viewers/ROCCurve.tsx
```diff
@@ -87,19 +115,26 @@ interface ROCCurveProps {
...
+  private readonly yTicks = Array.from({ length: 10 }, (_, i) => Number((i * 0.1).toFixed(1)));
```
srtanish1992, line 150: Changing length to 11 will fix the issue of including `1`.

frontend/src/components/viewers/ROCCurve.tsx
```diff
@@ -115,131 +150,234 @@ class ROCCurve extends Viewer<ROCCurveProps, ROCCurveState> {
...
-              <div className={css.crosshair}>
-                {hoveredValues.map((value, i) => (
-                  <div key={i} className={css.crosshairLabel}>{`${labels[i]}: ${value.label}`}</div>
-                ))}
-              </div>
-            </Crosshair>
+            <Tooltip
+              cursor={{ stroke: '#9aa0a6' }}
+              content={({ active, payload, label }) => {
+                if (!active || !payload?.length) {
```
srtanish1992: Just for readability, would recommend extracting the logic for tooltip content into a separate function. And when we upgrade to react 18, we can memoize it so that a new function is not created on each render.

frontend/package.json
```diff
@@ -174,6 +173,9 @@
   },
   "overrides": {
     "nanoid": "3.3.8",
+    "recharts": {
+      "react-redux": "8.1.3"
```
srtanish1992, line 177: Do we need to remove this override for react 18 upgrade?

srtanish1992:
@jeffspahr I have completed my code review.

frontend/src/components/viewers/ROCCurve.tsx
jeffspahr: Fixed in 9878875d3. Removed the unused `css.axis` style block and dropped the now-unused `fontsize` import from `ROCCurve.tsx`.

frontend/src/components/viewers/ROCCurve.tsx
jeffspahr: Addressed in 373440e09.

Regression summary: during the `react-vis` -> `recharts` migration, ROC tooltip rendering moved from `Crosshair` (styled by `react-vis/dist/style.css`) to a custom Recharts tooltip container. We kept the dark tooltip background but did not set an explicit text color, so text inherited a dark page color and became low-contrast.

Fixes in this commit:
- Tokenized ROC axis and cursor styling to use shared `Css.tsx` values (`color.grey`, `color.strong`, `color.weak`, `fontsize.small`) instead of local hardcoded literals.\n- Added shared tooltip tokens in `Css.tsx` (`color.tooltipBg`, `color.tooltipShadow`) and switched ROC tooltip styling to use them.
- Explicitly set tooltip text color to `color.background` for both the tooltip container and labels to restore readable contrast.

frontend/src/components/viewers/ROCCurve.tsx
jeffspahr: Addressed in commit 510b01519.

For readability, I extracted tooltip rendering logic out of the inline `Tooltip.content` callback into dedicated helpers:
- `resolveHoveredX(...)`
- `buildTooltipRows(...)`
- `renderTooltipRows(...)`
- `renderTooltipContent(...)`

The `Tooltip` now uses a small adapter (`content={tooltipProps => this.renderTooltipContent(...)}`), so `render()` stays easier to scan while preserving existing behavior.

frontend/package.json
jeffspahr: Yes. We should keep this override while frontend is on React 17.

Without it, `recharts@3.7.0` resolves `react-redux@9.x`, whose peer deps are `react ^18 || ^19`, so npm raises `ERESOLVE overriding peer dependency` and we end up in an unsupported React 17 combination.

Plan: remove this override in the React 18 upgrade PR, regenerate lockfile, and verify `npm ls react-redux` resolves recharts to `react-redux@9.x` cleanly under React 18.

jeffspahr:
### ROC curve under metrics section on Compare runs page.

> 1. The grid lines are spaced out due to the intervals on y axis.

Fixed

> 2. The text inside the tooltip is not visible.

Fixed

> 3. When I hover over the series legend color blue, I see the red color as well. Same happens for blue.

Fixed. I couldn't reproduce this in the latest commit so please confirm on your end as well.

> 4. The series legend icon is different.

Good catch. I think the newer legend looks better.  I'll keep it unless you have a strong opinion on it.

> 6. The y axis title is not completely visible.

Fixed. Please confirm on your end.

###  Lineage explorer graph.

> 7. The edges don't connect correctly.

Fixed.  Decided to make it better than the original rather than just match it.
image

frontend/src/components/viewers/ROCCurve.tsx
jeffspahr: My first pass included the 1, but existing behavior actually ends at .9 and .95.  I've matched that for now.  I started a thread in the pipelines slack channel to see if there are any strong opinions on this.  If folks want 1, then I'll follow up with a small fix.

frontend/src/components/viewers/ROCCurve.tsx
jeffspahr: My first pass included the 1, but existing behavior actually ends at .9 and .95.  I've matched that for now.  I started a thread in the pipelines slack channel to see if there are any strong opinions on this.  If folks want 1, then I'll follow up with a small fix.

droctothorpe:
/approve
/lgtm

Thank you both! 🙏

Issue #12603, opened 2026-01-02
[feature] KEP-11385: Implement UI drop-down menu for Literal input parameters


mprahl:
### Feature Area

/area frontend

Part of [KEP-11385](https://github.com/kubeflow/pipelines/blob/master/proposals/11385-literal-input-parameters/README.md).

### Acceptance Criteria

The UI should display a drop-down menu for Literal input parameters to indicate to a user that the input is limited to the specific set of values.

 * UI displays drop-down menu containing parameter value options for Literal runtime parameters. The user cannot type a custom input into the box - options are selection-only.
 * A value must be selected from the drop-down menu before the user can initiate a pipeline run.

khushiiagrawal:
hii @mprahl , I would like to work on this issue.

mprahl:
@khushiiagrawal thanks for volunteering! I'd suggest implementing the SDK portion first.

khushiiagrawal:
@mprahl  sure, i have mentioned that in the /sdk issue.
thanks!

Issue #12602, opened 2026-01-02
[feature] KEP-11385: Add validation for Literal runtime parameters in API Server and Driver


mprahl:
### Feature Area

/area backend

Part of [KEP-11385](https://github.com/kubeflow/pipelines/blob/master/proposals/11385-literal-input-parameters/README.md).

### Acceptance Criteria

The `ComponentInputsSpec_ParameterSpec` struct should be updated with a `literals` field:

```go
type ComponentInputsSpec_ParameterSpec struct {
    state protoimpl.MessageState `protogen:"open.v1"`
    Type PrimitiveType_PrimitiveTypeEnum `protobuf:"varint,1,opt,name=type,proto3,enum=ml_pipelines.PrimitiveType_PrimitiveTypeEnum" json:"type,omitempty"`
    ParameterType ParameterType_ParameterTypeEnum `protobuf:"varint,2,opt,name=parameter_type,json=parameterType,proto3,enum=ml_pipelines.ParameterType_ParameterTypeEnum" json:"parameter_type,omitempty"`
    DefaultValue *structpb.Value `protobuf:"bytes,3,opt,name=default_value,json=defaultValue,proto3" json:"default_value,omitempty"`
    IsOptional bool `protobuf:"varint,4,opt,name=is_optional,json=isOptional,proto3" json:"is_optional,omitempty"`
    Description   string `protobuf:"bytes,5,opt,name=description,proto3" json:"description,omitempty"`
    Literals []*structpb.Value `protobuf:"bytes,6,rep,name=literals,proto3" json:"literals,omitempty"`
    unknownFields protoimpl.UnknownFields
    sizeCache     protoimpl.SizeCache
}

```

### Acceptance Criteria

* **ComponentInputsSpec_ParameterSpec** proto contains a `literals` field.
* **Pipeline Run Server** checks each runtime parameter’s `ComponentInputsSpec_ParameterSpec` literals field for a non-empty value. Runtime parameter is validated against this field if non-empty. (*Hint:* `v2_template.validatePipelineJobInputs()`)
* Extend `backend/src/v2/driver.resolveInputParameter()` to validate resolved upstream inputs against Literal parameters.
[...]

PR #12839, opened 2026-02-17
feat(frontend): add Literal parameter dropdown for pipeline run inputs  #12603

khushiiagrawal:

Implements the UI portion of KEP-11385 (#12603): adds a drop-down `<Select>` menu for pipeline parameters that have a non-empty `literals` constraint, replacing the free-text `<TextField>`.

### Changes

- **`api/v2alpha1/pipeline_spec.proto`**: Adds `repeated google.protobuf.Value literals = 6` field to `ComponentInputsSpec.ParameterSpec` (shared with backend PR #12602).
- **`frontend/src/generated/pipeline_spec/pipeline_spec.ts`**: Regenerated TypeScript from updated proto, adding `literals: any[]` to `ComponentInputsSpec_ParameterSpec`.
- **`frontend/src/components/NewRunParametersV2.tsx`**:
  - When `param.literals` is non-empty, renders a Material-UI `<Select>` dropdown instead of a `<TextField>`.
  - The dropdown is selection-only — users cannot type custom values.
  - Shows a disabled "Select a value" placeholder when no default is set.
  - Uses `??` (nullish coalescing) to correctly handle falsy values like `0`.
  - Uses `String()` coercion for consistent type handling across string, integer, and float literals.
- **`frontend/src/components/NewRunParametersV2.test.tsx`**: Adds 8 new test cases in a dedicated `Literal Parameter Dropdown (#12603)` describe block:
  - String, integer, and float literal dropdown rendering
  - Empty literals gracefully falls back to TextField
  - `handleParameterChange` fires correctly on dropdown selection
  - No-default literal parameter marks form invalid (blocks run initiation)
  - Selecting a value from no-default dropdown marks form valid
  - Cloned RuntimeConfig pre-selects the correct literal value

### Acceptance Criteria

- [x] UI displays drop-down menu containing parameter value options for Literal runtime parameters. The user cannot type a custom input into the box — options are selection-only.
- [x] A value must be selected from the drop-down menu before the user can initiate a pipeline run.

## Checklist:
- [x] The title for your pull request (PR) follows the [pull request title convention](https://github.com/kubeflow/pipelines/blob/master/CONTRIBUTING.md#pull-request-title-convention).


khushiiagrawal:
@alyssacgoins   Please take a look.
thanks!

alyssacgoins:
> @alyssacgoins Please take a look. thanks!

Hey Khushi, thanks for your PR - can you take a look at the Copilot comments? Not all of them will be applicable, but some may be.

frontend/src/components/NewRunParametersV2.tsx
```diff
@@ -418,6 +430,33 @@ class ParamEditor extends React.Component<ParamEditorProps, ParamEditorState> {
...
+          >
+            {!param.value && param.value !== 0 && (
+              <MenuItem value='' disabled>
+                Select a value
+              </MenuItem>
+            )}
+            {param.literals.map(literal => (
+              <MenuItem key={String(literal)} value={String(literal)}>
+                {String(literal)}
+              </MenuItem>
```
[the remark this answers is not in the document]
alyssacgoins: @khushiiagrawal can you add an edge case verifying that this logic is covered?

alyssacgoins:
@khushiiagrawal can you fix the `frontend tests` CI failure?

khushiiagrawal:
@alyssacgoins  thanks , i have addressed all the copilot suggestions and fixed the failing CI as well. PTAL.
thanks for your time.

alyssacgoins:
@jeffspahr would you mind taking a look at this PR? Here is the full KEP for this feature, for reference: https://github.com/kubeflow/pipelines/blob/master/proposals/11385-literal-input-parameters/README.md

jeffspahr:
**Correction on item #4 (proto scope concern):**

I was wrong to flag the `pipeline_spec.ts` regeneration as "unrelated changes bundled in." Since `pipeline_spec.ts` is generated from `pipeline_spec.proto`, regenerating it to pick up the new `literals` field necessarily pulls in every other proto change that accumulated since the last TypeScript regeneration (`TaskConfigPassthrough`, `WorkspaceConfig`, resource deprecations, etc.). That's expected behavior — not scope creep. Please disregard that concern.

The same applies to item #6 (`WorkflowUtils.test.ts`) — the `platform: ''` addition is a downstream consequence of the regenerated proto adding a field to `SinglePlatformSpec`, which is the right fix to keep the test aligned with the updated type.

jeffspahr:
**Follow-up: are the regenerated proto changes safe to merge?**

Yes — I reviewed the frontend codebase for usage of every type/field that changed in the regeneration. Summary:

**No risk (most changes):** The new types (`WorkspaceConfig`, `TaskConfigPassthrough`, `PipelineConfig`, `KubernetesWorkspaceConfig`, etc.) are not imported or referenced anywhere in the frontend outside the generated file. New fields on existing types like `ResourceSpec`, `CachingOptions`, `ImporterSpec`, `RuntimeArtifact`, and `AcceleratorConfig` are also unused in frontend code. These are all additive/optional — merging them is inert.

**One minor logic note:** The new `TASK_CONFIG = 8` value on `ParameterType_ParameterTypeEnum`. `NewRunParametersV2.tsx` has several switch statements over this enum, but they all have `default` cases so it won't crash. `TASK_CONFIG` parameters are backend-injected (not user-facing), so they shouldn't appear in the run parameters UI anyway.

No coordination with other work needed — these are safe as-is.

jeffspahr:
**Revised summary — actionable requests:**

After further discussion, here's what we're actually asking for in this PR:

1. **Add a "Select a value" placeholder for no-default literal params.** The PR description mentions a "disabled 'Select a value' placeholder when no default is set," but the code doesn't implement it — the dropdown just renders blank. Something like `displayEmpty` + `renderValue` on the `<Select>` would make it clear the user needs to pick a value.

2. **Narrow the `literals` type on the `Param` interface** from `any[]` to `(string | number | boolean)[]` to match the KEP's supported types and prevent accidentally passing objects/arrays.

Everything else from the initial review is either withdrawn (proto scope concern) or informational only. The core dropdown implementation and test coverage look good.

khushiiagrawal:
@jeffspahr  i have addressed the revised summary changes that you mentioned. thank you so much for the review and the feedback. PTAL.

frontend/src/components/NewRunParametersV2.tsx
```diff
@@ -418,6 +430,43 @@ class ParamEditor extends React.Component<ParamEditorProps, ParamEditorState> {
...
+            aria-label={param.key}
+            displayEmpty
+            renderValue={selected =>
+              selected === '' ? (
+                <span style={{ color: '#aaa' }}>Select a value</span>
+              ) : (
+                String(selected)
+              )
+            }
+          >
```
jeffspahr, line 459: nit (optional): Inline `style={{ color: '#aaa' }}` works fine, but using `theme.palette.text.disabled` (via MUI's `useTheme` or `sx` prop) would be more consistent with the rest of the MUI theming. Not blocking.

frontend/src/components/NewRunParametersV2.test.tsx
```diff
@@ -1083,3 +1083,285 @@ describe('Bug Fix: Default Parameters in Compare Runs (#12536)', () => {
...
+        },
+      },
+      clonedRuntimeConfig: { parameters: { env: 'prod' } },
+      handleParameterChange: handleParameterChangeSpy,
+      setIsValidInput: vi.fn(),
+    };
+    render(<NewRunParametersV2 {...props} />);
+
+    // The cloned value 'prod' should be displayed, not the default 'dev'
+    screen.getByText('prod');
```
jeffspahr, line 1267: nit (optional): The `'does not show placeholder when literal default value is 0'` test validates the negative case (placeholder absent when value is selected). Consider adding `expect(screen.getByText('Select a value')).toBeInTheDocument()` to the existing `'marks input as invalid when literal parameter has no default value'` test to also verify the positive case — that the placeholder **is** rendered when nothing is selected.

jeffspahr:
@khushiiagrawal The `frontend-tests` CI failure is a Prettier formatting issue in `src/components/NewRunParametersV2.tsx` (not a test failure). To fix:

```sh
cd frontend && npm run format
```

Then commit and push. The `check_ci_status` failure is just a meta-check that gates on all other checks, so it will resolve once `frontend-tests` passes.

khushiiagrawal:
@jeffspahr  I have ran the formatting. all the test cases passed now. could you please take a look. 
Thanks a lot!

jeffspahr:
@khushiiagrawal CI is passing now, but there's an issue that will cause problems at merge time.

**What happened**: Your branch has `package-lock.json` pinning **Prettier 1.19.1**, while master has **Prettier 3.8.1**. The Prettier 1.x default for `arrowParens` is `"avoid"` (`e =>`), but Prettier 3.x defaults to `"always"` (`(e) =>`). Since `.prettierrc.yaml` doesn't set this explicitly, running `npm run format` with your older Prettier reformatted 171 files that were already correctly formatted on master — just in a different style.

CI passes because it installs from your lockfile (1.19.1) and the formatting is self-consistent. But when this merges to master, the 171 reformatted files will conflict with master's Prettier 3.8.1 style.

**To fix:**

```sh
# Rebase onto current master to pick up the correct lockfile
git fetch origin master
git rebase origin/master

# Install dependencies from the updated lockfile (gets Prettier 3.8.1)
cd frontend
npm ci

# Re-format only your file
npx prettier --write src/components/NewRunParametersV2.tsx src/components/NewRunParametersV2.test.tsx

# Commit and force-push
git add -A
git commit -m "style: reformat with Prettier 3.8.1"
git push --force-with-lease
```

After the rebase, master's files will already be formatted correctly for Prettier 3.8.1, so only your new code will need formatting. The 171 extra file changes will disappear.

(Filed #13020 to add a CI check that catches this kind of lockfile drift in the future.)

khushiiagrawal:
@jeffspahr  thanks for the review and telling the fix, i have ran all the commands and fixed it. 
thankyou.

jeffspahr:
Thanks for the quick fix @khushiiagrawal!

PR #12862, opened 2026-02-19
doc: Add a KEP for native MLflow integration in KFP

mprahl:

Kubeflow Pipelines has built-in experiment tracking with run comparison in the UI, but it is limited in capability and
exists in isolation from the rest of the Kubeflow ecosystem. Users of Kubeflow Notebooks, Kubeflow Trainer, and Kubeflow Pipelines have no unified view of experiment metrics across these components. MLflow is one option for providing that unified experience. However, users who want MLflow tracking today must manually configure it in each pipeline component, leading to fragmented setup, no automatic correlation between KFP pipeline runs and MLflow experiments/runs, and no centralized view across the platform.

This proposal adds a plugin-style metadata interface to the `Run` proto (`plugins_input` / `plugins_output`) and
implements MLflow as the first built-in integration across the API server, driver, and launcher. The metadata interface is based on the generic plugin architecture proposed in [KEP #12700](https://github.com/kubeflow/pipelines/pull/12700). KEP #12700 will align with the interface defined here and provide the full generic plugin architecture (plugin server, lifecycle hooks, entry points). However, KEP #12700 is blocked on MLMD removal, which is a prerequisite for it. This proposal implements MLflow as a built-in special case so that when KEP #12700 lands, the transition to the generic plugin architecture is seamless and clients experience no breaking changes. Long-term, the intention is to migrate this built-in MLflow functionality to the generic plugin architecture without breaking changes.

**Checklist:**

### DIFF
--- a/proposals/12862-mlflow-integration/README.md
+++ b/proposals/12862-mlflow-integration/README.md
@@ -1,0 +1,812 @@
+# KEP-12862: MLflow Integration for Kubeflow Pipelines
+
+## Summary
+
+Kubeflow Pipelines has built-in experiment tracking with run comparison in the UI, but it is limited in capability and
+exists in isolation from the rest of the Kubeflow ecosystem. Users of Kubeflow Notebooks, Kubeflow Trainer, and Kubeflow
+Pipelines have no unified view of experiment metrics across these components. MLflow is one option for providing that
+unified experience. However, users who want MLflow tracking today must manually configure it in each pipeline component,
+leading to fragmented setup, no automatic correlation between KFP pipeline runs and MLflow experiments/runs, and no
+centralized view across the platform.
+
+This proposal adds a plugin-style metadata interface to the `Run` proto (`plugins_input` / `plugins_output`) and
+implements MLflow as the first built-in integration across the API server, driver, and launcher. The metadata interface
+is based on the generic plugin architecture proposed in [KEP #12700](https://github.com/kubeflow/pipelines/pull/12700).
+KEP #12700 will align with the interface defined here and provide the full generic plugin architecture (plugin server,
+lifecycle hooks, entry points). However, KEP #12700 is blocked on MLMD removal, which is a prerequisite for it. This
+proposal implements MLflow as a built-in special case so that when KEP #12700 lands, the transition to the generic
+plugin architecture is seamless and clients experience no breaking changes. Long-term, the intention is to migrate this
+built-in MLflow functionality to the generic plugin architecture without breaking changes.
+
+## Motivation
+
+KFP's built-in run comparison allows users to view metrics across pipeline runs within the KFP UI. This works for simple
+cases but does not scale to workflows that span multiple Kubeflow components. A data scientist may train a model with
+Kubeflow Trainer, preprocess data in a Kubeflow Notebook, and orchestrate the full workflow with Kubeflow Pipelines.
+Today, metrics from each of these components are siloed in their respective UIs with no way to correlate them.
+
+MLflow is a widely adopted experiment tracking framework. Version 3.10 introduced
+[workspaces](https://mlflow.org/docs/latest/self-hosting/workspaces/) for multi-tenancy, which aligns well with KFP's
+multi-user model where each user operates within a Kubernetes namespace. A Kubernetes-native deployment of MLflow can
+map namespaces to MLflow workspaces and accept Kubernetes tokens for authentication, making it a natural fit for the
+Kubeflow ecosystem. A broader proposal for integrating MLflow into the Kubeflow ecosystem will be pursued separately.
+
+The primary workaround today is to add MLflow client calls directly in each pipeline component. This requires every
+component author to configure `MLFLOW_TRACKING_URI`, manage credentials, create or look up experiments, and manually
+correlate MLflow runs with KFP runs. The result is challenging, in particular for users not familiar with Kubernetes.
+There is also no standard way to see which MLflow run corresponds to a given KFP pipeline run, or to automatically mark
+an MLflow run as complete when the pipeline finishes.
+
+Providing a built-in integration where the API server, driver, and launcher handle MLflow interactions transparently
+would eliminate this manual setup and give users automatic, consistent experiment tracking across all pipeline
+components. Crucially, pipelines are not required to change and will continue to work on deployments without MLflow.
+When MLflow is configured, the same pipelines gain an enhanced experience with automatic experiment tracking, parameter
+and metric logging, and cross-referencing between KFP and MLflow runs without any modification to the pipeline code
+itself.
+
+### Goals
+
+1. Automatically create an MLflow parent run when a KFP pipeline run is created, and mark it as complete or failed when
+   the pipeline finishes.
+1. Track each task as a nested MLflow run, including individual iterations of parallel-for loops.
+1. Automatically log scalar metrics and input/output parameters from each task to MLflow without requiring changes to
+   pipeline components.
+1. Allow the user to optionally specify an MLflow experiment name when creating a run. Default to the `"Default"` MLflow
+   experiment when omitted.
+1. Support the same MLflow configuration on recurring runs so that every triggered run is automatically tracked.
+1. Expose `MLFLOW_TRACKING_URI`, `MLFLOW_WORKSPACE`, and `MLFLOW_RUN_ID` as environment variables on user containers so
+   that component code can interact with MLflow directly if needed. `MLFLOW_RUN_ID` points to the task's nested run,
+   allowing `mlflow.log_metric()` and similar calls to work without explicit run management.
+1. Pre-configure MLflow authentication on user containers by setting the appropriate environment variables
+   (`MLFLOW_TRACKING_AUTH`, `MLFLOW_TRACKING_TOKEN`, or `MLFLOW_TRACKING_USERNAME` / `MLFLOW_TRACKING_PASSWORD`) so that
+   the MLflow Python SDK authenticates automatically without manual setup in component code.
+1. When MLflow is deployed with Kubernetes-native multi-tenancy, govern access to MLflow experiments through Kubernetes
+   RBAC.
+1. The plugin metadata interface should align with the future generic plugin architecture in
+   [KEP #12700](https://github.com/kubeflow/pipelines/pull/12700) to ensure no breaking changes when that architecture
+   is adopted.
+
+### Non-Goals
+
+1. Artifact synchronization between KFP and MLflow is out of scope. KFP artifacts will remain in their configured object
+   storage (S3, GCS, MinIO). Only scalar metrics and parameters will be synced to MLflow.
+1. Classification and structured metrics (e.g., confusion matrices, ROC curves) will not be logged to MLflow initially.
+   Only scalar metrics are supported in this proposal.
+1. The generic plugin architecture defined in [KEP #12700](https://github.com/kubeflow/pipelines/pull/12700) is out of
+   scope. That KEP is blocked on MLMD removal. This proposal simulates the plugin interface for forward compatibility.
+1. SDK and frontend changes are out of scope for the initial backend proposal. Placeholder sections are included below;
+   detailed design will follow pending approval of this proposal.
+1. AWS SageMaker managed MLflow is not supported initially. SageMaker uses AWS SigV4 request signing, which requires the
+   AWS SDK for Go in the driver/launcher and IAM Roles for Service Accounts (IRSA) for pod-level credentials. This can
+   be added as a future `authType` (e.g., `"aws-sigv4"`) without breaking changes.
+
+## Proposal
+
+The integration touches three backend components: the API server, the driver, and the launcher. New `plugins_input` and
+`plugins_output` fields are added to the `Run` proto along with `PluginOutput` and `MetadataValue` messages. MLflow is
+implemented as built-in logic keyed on the `"mlflow"` entry in the plugin maps.
+
+### Proto Changes
+
+Two new messages (`PluginOutput`, `MetadataValue`) are introduced and `plugins_input` / `plugins_output` fields are
+added to both `Run` and `RecurringRun`:
+
+```proto
+message MetadataValue {
+  enum RenderType {
+    UNSPECIFIED = 0;
+    URL = 1;
+  }
+  google.protobuf.Value value = 1;
+  optional RenderType render_type = 2;  // Hint for UI rendering
+}
+
+message PluginOutput {
+  map<string, MetadataValue> entries = 1;
+  RuntimeState state = 2;   // Reuses the existing RuntimeState enum (e.g., SUCCEEDED, FAILED)
+  string state_message = 3; // Human-readable detail, especially on failure
+}
+
+```
+
+The `plugins_input` and `plugins_output` fields are placed directly on the `Run` message (no wrapper). The
+`plugins_input` map uses `google.protobuf.Struct` for each plugin key, allowing arbitrary JSON input per plugin without
+schema coupling. The `plugins_output` map uses `PluginOutput` which wraps a map of `MetadataValue` entries, each with an
+optional `render_type` hint for UI rendering (e.g., rendering a value as a hyperlink when `render_type` is `URL`).
+
+A realistic example of `plugins_input` and `plugins_output` on a run:
+
+```json
+{
+  "plugins_input": {
+    "mlflow": {
+      "experiment_name": "sentiment-classifier-tuning"
+    }
+  },
+  "plugins_output": {
+    "mlflow": {
+      "entries": {
+        "experiment_name": { "value": "sentiment-classifier-tuning" },
+        "experiment_id": { "value": "42" },
+        "root_run_id": { "value": "a1b2c3d4e5f6" },
+        "run_url": {
+          "value": "https://mlflow.example.com/experiments/42/runs/a1b2c3d4e5f6?workspace=my-namespace",
+          "render_type": "URL"
+        }
+      }
+    }
+  }
+}
+```
+
+### End-to-End Flow
+
+The following diagram shows how the MLflow integration spans the pipeline run lifecycle:
+
+```mermaid
+sequenceDiagram
+    participant User
+    participant APIServer as API Server
+    participant MLflow
+    participant MLMD
+    participant Driver
+    participant Launcher
+
+    User->>APIServer: CreateRun (optionally with plugins_input.mlflow overrides)
+    APIServer->>APIServer: Authorization check (authType=kubernetes)
+    APIServer->>MLflow: Create experiment (if needed)
+    APIServer->>MLflow: Create parent run (tag: KFP run URL)
+    APIServer-->>User: Return Run with plugins_output.mlflow
+
+    loop For each task
+        Driver->>Driver: Read parent MLflow run ID from env var
+        Driver->>MLflow: Create nested run
+        Driver->>MLMD: Store nested run ID as execution property
+        Driver->>Driver: Set MLFLOW_TRACKING_URI + MLFLOW_WORKSPACE + MLFLOW_RUN_ID env vars
+        Driver->>Launcher: Launch user container
+
+        Launcher->>Launcher: Execute user code
+        Launcher->>MLflow: Log input parameters
+        Launcher->>MLflow: Log scalar output metrics
+        Launcher->>MLflow: Mark nested run complete/failed
+    end
+
+    APIServer->>APIServer: ReportWorkflowResource (terminal state)
+    APIServer->>MLflow: Mark parent run complete/failed
+    APIServer->>MLflow: Close any open nested runs
+```
+
+### Risks and Mitigations
+
+**MLflow API availability.** If the MLflow server is unreachable when a run is created, the MLflow API calls within the
+KFP `CreateRun` handler could fail. All MLflow REST API calls will use retries with exponential backoff and a
+configurable timeout (default 30 seconds total per operation) to handle transient failures. If the MLflow call still
+fails after retries, the KFP run is still created and `plugins_output.mlflow.state` is set to `FAILED` with a
+`state_message` describing the failure. A similar approach applies to driver and launcher MLflow calls; failures are
+retried with the same policy, and if they ultimately fail, the errors are logged but do not block task execution. If the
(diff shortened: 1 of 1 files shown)


proposals/12862-mlflow-integration/README.md
```diff
@@ -0,0 +1,611 @@
...
+All MLflow REST API calls will use retries with exponential backoff and a configurable timeout (default 30 seconds total
+per operation) to handle transient failures. If the MLflow call still fails after retries, the KFP run is still created
+but `plugins_output.mlflow` is not populated. The error is logged and surfaced in the run's error field. A similar
+approach applies to driver and launcher MLflow calls; failures are retried with the same policy, and if they ultimately
+fail, the errors are logged but do not block task execution. If the API server successfully creates an MLflow parent run
+but subsequently fails to persist the KFP run in the database, the orphan MLflow run is left as-is. Orphan runs in
+MLflow are harmless (they appear as "running" indefinitely) and can be cleaned up manually or via MLflow's garbage
+collection. The narrow failure window makes compensating deletes not worth the added complexity.
+
+**Performance overhead.** Each task execution incurs additional HTTP calls to the MLflow REST API (create nested run,
```
nsingla, line 192: Is there a limitation to number of total number parameters mlflow can handle in a single call? Doesn't look like it based on their documentation 🤔

proposals/12862-mlflow-integration/README.md
```diff
@@ -0,0 +1,611 @@
...
+log parameters, log metrics, update run status). For pipelines with hundreds of tasks, this adds latency. This is
```
nsingla, line 193: How should we handle failures i.e. when mlflow api call fails but the component code execution itself succeeds? I think, we should mark the component run successful but somehow surface that the mlflow calls failed. 
We can have a parallel routine/subroutine that runs in parallel with the executor routine. This way we have the mlflow logic independent of the executor, but log mlflow calls in the same logger so that in case of mlflow api call failures, it shows up in the executor logs?

proposals/12862-mlflow-integration/README.md
```diff
@@ -0,0 +1,611 @@
...
+mitigated by keeping MLflow calls in the launcher's post-execution path (non-blocking to the user's code) and by using
+the MLflow REST API's batch endpoints where available.
+
+**Configuration divergence.** The API server resolves the MLflow configuration (tracking URI, workspace, parent run ID)
```
nsingla, line 199: How do we handle hard coded configuration with in the pipeline code? which can override the mlflow environment variables at run time. In this case, we may not receive any mlflow event to the mlflow instance we expect or fail silently.

proposals/12862-mlflow-integration/README.md
```diff
@@ -0,0 +1,611 @@
...
+```diff
+ message Run {
+   // ... existing fields 1-17 ...
+
+   // Output. A sequence of run statuses. This field keeps a record
+   // of state transitions.
+   repeated RuntimeStatus state_history = 17;
++
++  // Optional input. Plugin inputs provided by the user at run creation.
++  map<string, google.protobuf.Struct> plugins_input = 19;
```
nsingla, line 246: I think its much safer to make this strict than an open ended Struct. That will make some of the validations logic in the api server (when parsing input) much simpler. And very trivial, but this will also allow us to declare default value for experiment_name here in the message

proposals/12862-mlflow-integration/README.md
nsingla: I guess the failure case is handled by the API server at the end of the workflow execution

proposals/12862-mlflow-integration/README.md
mprahl: I think it should be somewhat light due to it being an API call per pipeline task rather than one for the whole run.

proposals/12862-mlflow-integration/README.md
mprahl: Yes, I opted for a plugin status field to allow the MLflow plugin to error but not affect the pipeline run. The UI would eventually render such plugin statuses.

proposals/12862-mlflow-integration/README.md
mprahl: That doesn't impact the integration but rather only impacts the convenience of preconfiguring the MLflow SDK at runtime. KFP itself is not reliant on that.

proposals/12862-mlflow-integration/README.md
mprahl: @HumairAK , what are your thoughts? I personally don't see another way without removing the dynamic nature of the plugin architecture proposed in the KEP #12700 .

proposals/12862-mlflow-integration/README.md
nsingla: yeah, but some of the components such as training can have really long list of input parameters

proposals/12862-mlflow-integration/README.md
nsingla: yeah, but we have 2 problems with this:
1. API server will create experiment and start parent run w/o realizing that executor will behave differently
2. Create confusion with user about where to go
Don't think its an easy problem to solve and we can't just blindly override user provided env variables (that are available only at run time)

proposals/12862-mlflow-integration/README.md
```diff
@@ -0,0 +1,611 @@
...
+
+### API Server
+
+MLflow integration requires an explicit opt-in at the API server level via a `plugins.mlflow` (JSON notation here) entry
+in the API server config. In multi-user mode, the per-namespace `kfp-launcher` ConfigMap (`plugins.mlflow` key) provides
+namespace-specific overrides but is only effective when the API server has MLflow enabled. If neither the API server nor
+the namespace ConfigMap provides MLflow configuration, MLflow integration is disabled for that namespace.
+
+When MLflow is enabled for a namespace (via either configuration level), it applies to **all** pipeline runs in that
+namespace automatically. There is no per-run toggle to enable or disable MLflow tracking. The `plugins_input.mlflow`
```
droctothorpe: Is there any value to providing optional opt-out for an individual run?

proposals/12862-mlflow-integration/README.md
```diff
@@ -0,0 +1,611 @@
...
+1. Add `MLFLOW_TRACKING_URI`, `MLFLOW_WORKSPACE`, and `MLFLOW_RUN_ID` (set to the nested run ID) to the user
+   container's environment variables via the pod spec patch. The driver already sets environment variables on the pod
+   spec in `backend/src/v2/driver/driver.go`. Setting `MLFLOW_RUN_ID` allows user code that calls the MLflow Python
+   API (e.g., `mlflow.log_metric()`) to automatically log to the correct nested run without needing to call
+   `mlflow.start_run()` or pass a run ID explicitly.
+
+#### Loop Iterations (`dag.go`)
+
+In a parallel-for loop, each iteration is executed by an iteration DAG driver followed by a container driver. Without
+coordination, both would attempt to create a nested MLflow run for the same iteration. To avoid this, the loop's parent
```
droctothorpe: Why have the parent DAG driver handle the MLFlow run creation rather than the container driver? For large fan outs this could inadvertently DDOS the MLFlow API and result in rate-limiting.

proposals/12862-mlflow-integration/README.md
```diff
@@ -0,0 +1,611 @@
+# KEP-12862: MLflow Integration for Kubeflow Pipelines
+
+## Summary
```
droctothorpe, line 3: What happens to the MLFlow parent run when a KFP run is cloned?

proposals/12862-mlflow-integration/README.md
```diff
@@ -0,0 +1,611 @@
...
+    end
+
+    APIServer->>APIServer: ReportWorkflowResource (terminal state)
+    APIServer->>MLflow: Mark parent run complete/failed
+    APIServer->>MLflow: Close any open nested runs
+```
+
+### Risks and Mitigations
+
+**MLflow API availability.** If the MLflow server is unreachable when a run is created, the `CreateRun` call could fail.
```
HumairAK: > the `CreateRun` call could fail.

Sounds like you mean mlflow `CreateRun` and not KFP run, can we make that more explicit?

proposals/12862-mlflow-integration/README.md
```diff
@@ -0,0 +1,611 @@
...
+        Launcher->>MLflow: Log scalar output metrics
+        Launcher->>MLflow: Mark nested run complete/failed
+    end
+
+    APIServer->>APIServer: ReportWorkflowResource (terminal state)
+    APIServer->>MLflow: Mark parent run complete/failed
+    APIServer->>MLflow: Close any open nested runs
+```
+
+### Risks and Mitigations
```
HumairAK, line 179: I guess not really a risk but worth pointing out that there is a difference between mlflow.log_metric and kfp.log_metric, if you use the latter then kfp logs metrics to mlflow post code execution, where as mlflow.log_metric in user code would be instant

proposals/12862-mlflow-integration/README.md
```diff
@@ -0,0 +1,611 @@
...
+the current token.
+
+The provider's behavior depends on the configuration mode:
+
+- **Kubernetes-native mode**: The provider reads the pod's projected service account token from
+  `/var/run/secrets/kubernetes.io/serviceaccount/token` on each request and passes it as a `Bearer` token in the
+  `Authorization` header on all MLflow API requests.
+- **Secret-based mode**: When `tokenSecretRef` is set in the `plugins.mlflow` configuration, the driver mounts the
+  referenced Secret into the user container. The provider reads the token from the mounted path on each request and
+  passes it as a `Bearer` token in the `Authorization` header on all MLflow API requests.
```
HumairAK: is this for mlflow vanilla auth?

proposals/12862-mlflow-integration/README.md
mprahl: Good point. I've expanded the "Configuration divergence" risk section to address this explicitly.

The key distinction is that KFP's built-in MLflow logging is performed by the Go launcher using its own resolved environment variables -- it does not go through the user's Python process. So if user code calls `mlflow.set_tracking_uri()` or hard-codes `MLFLOW_TRACKING_URI`, this only affects the user's own `mlflow.log_metric()` calls, not the KFP-managed parent/nested run hierarchy. The KFP-managed runs remain consistent on the KFP-configured MLflow instance regardless of what the user's code does at runtime.

proposals/12862-mlflow-integration/README.md
mprahl: So the whole integration will be opt-in but I also just added an opt-out in the plugin input.

proposals/12862-mlflow-integration/README.md
mprahl: I've updated the wording in the doc to be clearer.

When there is a loop task, the task hierarchy is essentially:
DAG driver -> Loop driver -> Iteration driver -> container driver

The point I was trying to make is we should skip creating the MLflow nested run for "Iteration driver" and instead, the container driver just uses the loop driver as its "parent run" in MLflow. The additional iteration driver is not intuitive.

proposals/12862-mlflow-integration/README.md
mprahl: Thanks for catching that. I updated the doc with clarification.

The same MLflow experiment would be used, but it'd be a new MLflow run, just like it is in KFP.

proposals/12862-mlflow-integration/README.md
```diff
@@ -0,0 +1,666 @@
...
+
+#### MLflow Request Auth Provider
+
+When the `mlflow` Python package is available in the user's container image, the KFP Python executor
+(`sdk/python/kfp/dsl/executor_main.py`) registers an MLflow
+[request auth provider](https://mlflow.org/docs/latest/ml/plugins/#authentication-plugins) before invoking the user's
+component function. The executor already runs inside the user container as the entrypoint for component execution, so
+registration happens in-process with no separate package needed. The Go launcher sets the necessary environment
+variables (e.g., `KFP_MLFLOW_AUTH_MODE`, `KFP_MLFLOW_TOKEN_PATH`) on the executor's environment; the executor checks for
+these and conditionally imports and registers the provider when `mlflow` is importable.
```
HumairAK: Are you suggesting we package an mlflow specific auth provider in kfp that we register in executor prior to component code execution?

proposals/12862-mlflow-integration/README.md
```diff
@@ -0,0 +1,666 @@
...
+
+When a run is created and MLflow is enabled for the namespace:
+
+1. If the deployment uses Kubernetes-native mode, perform the authorization check (see [Authorization](#authorization)).
+1. Validate `plugins_input.mlflow` against the `MLflowPluginInput` schema (see below). Reject the request with a
+   descriptive error if the struct does not conform to the expected fields. If `plugins_input.mlflow.disabled` is
+   `true`, skip all subsequent MLflow steps for this run.
+1. Read `plugins_input.mlflow.experiment_name` from the request if provided. If absent or if `plugins_input.mlflow` is
+   not set at all, default to `"Default"`.
+1. Call the MLflow REST API to create the experiment if it does not already exist
```
HumairAK, line 309: anything we want to add to experiment description? "Created by KFP" or some generic message.

proposals/12862-mlflow-integration/README.md
```diff
@@ -0,0 +1,666 @@
...
+    "mlflow": {
+      "experiment_name": "sentiment-classifier-tuning"
+    }
+  },
+  "plugins_output": {
+    "mlflow": {
+      "entries": {
+        "experiment_name": { "value": "sentiment-classifier-tuning" },
+        "experiment_id": { "value": "42" },
+        "run_id": { "value": "a1b2c3d4e5f6" },
```
HumairAK: how about root_run_id? since there will be various runs assiciated with this kfp run

proposals/12862-mlflow-integration/README.md
```diff
@@ -0,0 +1,666 @@
...
+affect the MLflow endpoint or workspace.
+
+User code may also override MLflow environment variables at runtime (e.g., calling `mlflow.set_tracking_uri()` or
+setting `MLFLOW_TRACKING_URI` programmatically). This does not affect KFP's built-in MLflow logging, which is performed
+by the Go launcher using its own resolved environment variables. The divergence only impacts the user's own MLflow SDK
+calls (e.g., `mlflow.log_metric()`) if the user explicitly redirects them to a different MLflow instance. In this
+scenario, the KFP-managed parent/nested run hierarchy remains consistent on the KFP-configured MLflow instance. Users
+are recommended to rely on the pre-configured environment variables rather than hard-coding MLflow connection details.
+
+## Design Details
```
HumairAK, line 212: If we could make a general implementation note that we should implement this in an agnostic way so all the logic is isolated to some "mlflow" module, and in KFP we call this package somewhat generically. So that it is easier to pinpoint later on what we need to maintain backwards compatibility with when we move to plugin architecture.

proposals/12862-mlflow-integration/README.md
```diff
@@ -0,0 +1,666 @@
...
+	Name string `json:"name"` // K8s Secret name in the namespace
+	Key  string `json:"key"`  // Key within the Secret
+}
+
+type MLflowPluginConfig struct {
+	TrackingURI          string     `json:"trackingURI"`
+	WorkspacesEnabled    *bool      `json:"workspacesEnabled,omitempty"`    // defaults to true
+	TokenSecretRef       *SecretRef `json:"tokenSecretRef,omitempty"`       // Reference to a Secret containing an auth token
+	AuthorizationEnabled *bool      `json:"authorizationEnabled,omitempty"` // defaults to false in standalone mode
+}
```
HumairAK: I think we might be hardcoding too many plugin specific details her as this is part of kfp backend code. Can we make this design more generic and then add passthrough configs that we parse in code? In the plugin proposal I had done: 

```json
{
  "PluginServers": [
    {
      "Name": "mlflow-plugin",
      "Endpoint": "http://mlflow-plugin-server:8080",
      "Timeout": "30s",
      "TLS": {
        "Enabled": false,
        "CABundlePath": "",
        "SkipVerify": false
      }
    },
    {
      "Name": "notifications-plugin",
      "Endpoint": "http://notifications-plugin-server:8080",
      "Timeout": "10s"
    }
  ]
}
```

but now I realize we need some "plugin" specific configurations, like `WorkspacesEnabled`, maybe we can do some generic object struct that we parse in code? 

the same would apply to kfp launcher

proposals/12862-mlflow-integration/README.md
HumairAK, line 309: we do the authz check for mlflow in step (1) if it's k8s mode, but api server also needs to ensure it's using `tokenSecretRef` if it's provided in non k8s mode, or ensuring it's using its SA bearer in k8s mode for authn/authz yes?

proposals/12862-mlflow-integration/README.md
```diff
@@ -0,0 +1,666 @@
...
+The provider's behavior depends on the configuration mode:
+
+- **Kubernetes-native mode**: The provider reads the pod's projected service account token from
+  `/var/run/secrets/kubernetes.io/serviceaccount/token` on each request and passes it as a `Bearer` token in the
+  `Authorization` header on all MLflow API requests.
+- **Secret-based mode**: When `tokenSecretRef` is set in the `plugins.mlflow` configuration, the driver mounts the
+  referenced Secret into the user container. The provider reads the token from the mounted path on each request and
+  passes it as a `Bearer` token in the `Authorization` header on all MLflow API requests.
+
+### Authorization
```
HumairAK: Unless I missed it, I think we should note at which layers which entities need permissions. 

For example in k8s mode, API Server SA and Pipeline Runner SA will need permissions. For the latter, if the user specifies a different SA for their pipeline run, they will need to ensure they need the appropriate permissions otherwise calls to mlflow can fail.

proposals/12862-mlflow-integration/README.md
```diff
@@ -0,0 +1,666 @@
...
+
+#### Container Tasks (`container.go`)
+
+Before launching a user container:
+
+1. Read the MLflow tracking URI, workspace, and parent run ID from the environment variables injected by the Argo
+   compiler at run creation time. For `tokenSecretRef` (which Secret to mount on the user container), read the
+   `kfp-launcher` ConfigMap (`plugins.mlflow` key) via the existing `Config` struct in `backend/src/v2/config/env.go`.
+1. Create a nested MLflow run under the parent run (`POST /api/2.0/mlflow/runs/create`).
+1. Store the nested MLflow run ID as the `mlflow_run_id` custom property on the MLMD execution, consistent with existing
```
HumairAK: we should also store the parent ID so we can recreate the tree if we needed to for migration

proposals/12862-mlflow-integration/README.md
```diff
@@ -0,0 +1,666 @@
...
+
+New `plugins_input` and `plugins_output` columns of type `LargeText` will be added to the `run_details` table,
+consistent with existing JSON blob columns such as `StateHistoryString` and `PipelineRuntimeManifest` in
+`backend/src/apiserver/model/run.go`. A database migration will be added following the existing migration pattern. The
+GORM ORM handles schema migration automatically; existing runs will have `NULL` for these columns, which the API server
+treats as empty maps.
+
+### Driver
+
+The driver (`backend/src/v2/driver/`) is responsible for creating nested MLflow runs and passing MLflow context to the
```
HumairAK, line 403: there are edge cases where only driver runs besides cache hit, I think it's just the pvc creation/deletion tasks, those would need to also be handled by driver similar to launch, i.e. driver needs to mark mlflow run complete

proposals/12862-mlflow-integration/README.md
mprahl: @HumairAK yes, though this part changes if we can contribute the client auth provider directly into MLflow. Since we seem to be headed that way, I'll mention this as the path forward as we can just set `MLFLOW_TRACKING_AUTH=kubernetes`. We can log a warning if the MLflow SDK is less than 3.11 since I assume that's when this will land.

proposals/12862-mlflow-integration/README.md
```diff
@@ -0,0 +1,801 @@
...
+  each namespace to define its own MLflow configuration via the `kfp-launcher` ConfigMap. If neither the API server nor
+  the namespace ConfigMap provides MLflow configuration, MLflow integration is disabled for that namespace.
+
+The Go types for the configuration use a generic `PluginConfig` wrapper with plugin-specific settings in a
+`json.RawMessage` field. This structure aligns with [KEP #12700](https://github.com/kubeflow/pipelines/pull/12700)'s
+plugin server config pattern and keeps the KFP core code decoupled from plugin-specific fields:
+
+```go
+type PluginConfig struct {
+	Endpoint string          `json:"endpoint"`
```
HumairAK, line 645: can we also include tls configuration options?

proposals/12862-mlflow-integration/README.md
nsingla: On second thoughts, I think this is fine, if we make it strict, it won;t be useful for other plugins

proposals/12862-mlflow-integration/README.md
```diff
@@ -0,0 +1,812 @@
...
+added to both `Run` and `RecurringRun`:
+
+```proto
+message MetadataValue {
+  enum ContentType {
+    UNSPECIFIED = 0;
+    URL = 1;
+  }
+  google.protobuf.Value value = 1;
+  optional ContentType content_type = 2;  // Hint for UI rendering
```
hbelmiro: `ContentType` [can confuse readers](https://github.com/kubeflow/pipelines/pull/12919#discussion_r2859440711). How about renaming it to `UIHint` or something to avoid confusion?

proposals/12862-mlflow-integration/README.md
HumairAK: yes it has to be consistent with runtime inputs

proposals/12862-mlflow-integration/README.md
mprahl: Updated to render_type.

mprahl:
@hbelmiro @HumairAK @droctothorpe thanks for the reviews! Could you take a look at the name change on the proto?
https://github.com/kubeflow/pipelines/pull/12862#discussion_r2874321105

PR #12863, opened 2026-02-20
Fix: Prevent Silent Duplicate Runs in V2 Recurring Pipelines

atharrva01:
## Summary

So I was working with V2 recurring pipelines and noticed something weird – retries were silently creating duplicate runs for the same scheduled slot.

Here's what's going on:

The controller has this idempotency guard that checks if an Argo Workflow CR named `swf.NextResourceName()` already exists:

```go
workflowName := swf.NextResourceName()
_, isNotFoundError, err := c.workflowClient.Get(swf.Namespace, workflowName)
```

This works perfectly fine for **V1**, because we directly create an Argo Workflow with that deterministic name.

But in **V2**, we're doing things differently – we delegate run creation to the API server:

```go
run, err := c.runClient.CreateRun(ctx, &api.CreateRunRequest{
    Run: &api.Run{
        DisplayName:    swf.NextResourceName(),
        RecurringRunId: string(swf.UID),
    },
})
```

The problem? The API server generates its own UUID-based workflow name internally, not the deterministic slot name we're checking for.

So the controller's guard always returns "not found" for V2 runs.

If `CreateRun` succeeds but then `updateStatus` fails with a `409 Conflict` (which happens normally sometimes), the controller retries and calls `CreateRun` again. This creates two runs for the same scheduled slot – and there's no visible error or anything.

---

## Impact

From what I can see, this might be causing:

Duplicate pipeline executions
Wasted compute (think duplicate GPU training runs)
Duplicate experiment artifacts and metrics
Silent data inconsistency

This affects V2 recurring pipelines, which are the default and recommended path for users.

---

## The Fix

Instead of relying on that controller-side Argo CR check, I've added **server-side idempotency** directly inside `CreateRun`.

Now we check if a run already exists for the same scheduled slot using:
```
(RecurringRunId, DisplayName)
```

At the top of `CreateRun`:

```go
if run.RecurringRunId != "" && run.DisplayName != "" {
    existing, err := r.runStore.GetRunByScheduledSlot(
        run.RecurringRunId,
        run.DisplayName,
    )
    if existing != nil {
        return existing, nil // idempotent return
    }
}
```

So now:
* First execution → we create the run normally
* Retry → we return the existing run
* No duplicate Argo Workflow
* No duplicate DB record

And non-scheduled runs aren't affected at all.

---

## Result

After this change:
* Each scheduled slot creates at most one run
* Controller retries are safe
* Those routine 409 conflicts don't cause duplication anymore
* We preserve experiment integrity and compute efficiency

This basically brings V2 recurring behavior in line with the idempotency guarantees that V1 already had.

### DIFF
--- a/backend/src/apiserver/resource/resource_manager.go
+++ b/backend/src/apiserver/resource/resource_manager.go
@@ -542,6 +542,23 @@ func (r *ResourceManager) CreateRun(ctx context.Context, run *model.Run) (*model
 		return nil, util.NewInternalServerError(err, "Failed to create a run due to error fetching manifest")
 	}
 
+	// Idempotency guard for V2 recurring runs: the SWF controller may invoke CreateRun
+	// a second time if the subsequent SWF status update fails with a 409 Conflict (a
+	// normal occurrence under load). Because the controller never creates an Argo
+	// Workflow CR with the deterministic slot name for V2 runs, its existing
+	// workflowClient.Get guard is always a no-op for this path. Instead, we check the
+	// DB here before any side-effectful work begins so that a retry returns the run
+	// that was already created for this (recurringRunId, displayName) slot.
+	if run.RecurringRunId != "" && run.DisplayName != "" {
+		existing, err := r.runStore.GetRunByScheduledSlot(run.RecurringRunId, run.DisplayName)
+		if err != nil {
+			return nil, util.Wrap(err, "Failed to check for existing run (idempotency check)")
+		}
+		if existing != nil {
+			return existing, nil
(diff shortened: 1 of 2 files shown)


atharrva01:
hi @VaniHaripriya @HumairAK This PR fixes a silent duplication issue in V2 recurring pipelines where retries (e.g., after a 409 on status update) could create multiple runs for the same scheduled slot. We add server-side idempotency in CreateRun keyed on (RecurringRunId, DisplayName) to ensure each slot executes at most once.

PR #12866, opened 2026-02-20
# fix(sdk): remove overly restrictive accelerator limit validation Fixes #12865

### LINKED ISSUE #12865: [sdk] [bug] [set_accelerator_limit] rejects valid accelerator counts (only allows 0, 1, 2, 4, 8, 16)
### Environment

- **KFP version:** 2.15.2
- **KFP SDK version:** 2.15.2
- **All dependencies version (`pip list | grep kfp`):**

```
kfp                  2.15.2
kfp-pipeline-spec    0.6.0
```

### Steps to reproduce

```python
from kfp import dsl

@dsl.component
def my_component():
    pass

@dsl.pipeline
def my_pipeline():
    # Any value not in {0, 1, 2, 4, 8, 16} raises ValueError
    my_component().set_accelerator_limit(3)   # ValueError
    my_component().set_accelerator_limit(32)  # ValueError
    my_component().set_accelerator_limit(128) # ValueError
```

Running the above raises:

```
ValueError: 'limit' must be one of 0, 1, 2, 4, 8, 16.
```

The validation is implemented in `sdk/python/kfp/dsl/pipeline_task.py` using a hardcoded allowlist regex:

```python
if isinstance(limit, str) and re.match(r'^0$|^1$|^2$|^4$|^8$|^16$', limit) is None:
    raise ValueError(f'{"limit"!r} must be one of 0, 1, 2, 4, 8, 16.')
```

### Expected result

Any non-negative integer should be accepted. Kubernetes imposes no restriction on accelerator resource counts — values like `3`, `5`, `32`, or `128` are valid and commonly needed in multi-GPU training workloads.

The method should behave consistently with `set_cpu_request` and `set_memory_limit`, which use pattern-based validation rather than an arbitrary allowlist.

There is also a minor bug in the error message: the f-string `f'{"limit"!r} must be one of 0, 1, 2, 4, 8, 16.'` produces the literal text `'limit' must be one of 0, 1, 2, 4, 8, 16.` rather than including the actual value passed.

### Materials and Reference

- Relevant source: [`sdk/python/kfp/dsl/pipeline_task.py`, `set_accelerator_limit`](https://github.com/kubeflow/pipelines/blob/master/sdk/python/kfp/dsl/pipeline_task.py)
- Kubernetes docs confirm no restriction on resource count values: https://kubernetes.io/docs/tasks/manage-gpus/scheduling-gpus/

---

kevo-1:
## Summary

`set_accelerator_limit()` validated its `limit` argument against a hardcoded allowlist `{0, 1, 2, 4, 8, 16}` using a regex, silently rejecting valid Kubernetes accelerator counts like 3, 32, or 128. This PR removes the restriction and replaces it with a simple non-negative integer check.

Fixes #12865

### `sdk/python/kfp/dsl/pipeline_task.py`

- On the **`int` path**: explicitly checks `limit < 0` and raises before converting to string — intent is clear without any regex.
- On the **`str` path**: replaces the hardcoded allowlist regex `^0$|^1$|^2$|^4$|^8$|^16$` with `^\d+$`, accepting any non-negative integer string.
- **Error message**: unified to `'limit must be a non-negative integer.'` on both paths (also fixes a minor bug where the original f-string produced `'limit'` as a literal rather than the variable value).
- **Docstring**: updated to reflect the actual constraint.

### `sdk/python/kfp/dsl/pipeline_task_test.py`

- Extended `test_set_accelerator_limit` with previously-rejected but valid values: `0`, `3`, `32`, `128`.
- Added new parameterized test `test_set_accelerator_limit_invalid` covering: `-1`, `-128`, `'abc'`, `'1.5'`, `'-1'`.

## Testing

```
pytest -v sdk/python/kfp/dsl/pipeline_task_test.py
```

```
47 tests passed
```

All 12 new test cases pass (7 valid, 5 invalid). No regressions in the existing test suite.

### DIFF
--- a/sdk/python/kfp/dsl/pipeline_task.py
+++ b/sdk/python/kfp/dsl/pipeline_task.py
@@ -430,12 +430,10 @@ class PipelineTask:
         else:
             if isinstance(limit, int):
                 if limit < 0:
-                    raise ValueError(
-                        'limit must be a non-negative integer.')
+                    raise ValueError('limit must be a non-negative integer.')
                 limit = str(limit)
             if isinstance(limit, str) and re.match(r'^\d+$', limit) is None:
-                raise ValueError(
-                    'limit must be a non-negative integer.')
+                raise ValueError('limit must be a non-negative integer.')
 
         if self.container_spec.resources is not None:
             self.container_spec.resources.accelerator_count = limit


zazulam:
@kevo-1 just [sign off](https://github.com/kubeflow/pipelines/pull/12866/checks?check_run_id=64348588069#:~:text=DCO%20to%20pass-,DCO,-There%20is%20one) your latest commit when you get the chance

kevo-1:
Signed off the commit ✔️

PR #12878, opened 2026-02-23
fix: validate GitHub API response in upgrade-test.yml

### LINKED ISSUE #12875: [bug] upgrade-test.yml has no error handling when GitHub API fails
### Environment

N/A — this is a CI workflow issue

### Steps to reproduce

1. The `upgrade-test.yml` workflow runs the "Get last release tag" step
2. The GitHub API call fails (due to rate limiting, network timeout, or outage)
3.  `lastRelease` becomes empty or `"null"`
4. The "Deploy Last Release" step runs `kubectl apply -k ...?ref=` with a bad ref
5. CI fails with a confusing Kubernetes error that hides the actual root cause

### Expected result

The workflow should validate the API response and fail with a clear error message when the GitHub API is unavailable, instead of proceeding with an invalid ref.

### Materials and reference

Affected file: [.github/workflows/upgrade-test.yml](https://github.com/kubeflow/pipelines/blob/master/.github/workflows/upgrade-test.yml#L57-L63)

### Labels

 /area testing 

---

gaurav712-gujjar:
## Problem
If the GitHub API fails (rate limit, timeout, or outage), `lastRelease` 
becomes empty or "null". The workflow then runs kubectl apply with an 
invalid ref like `?ref=null`, causing a confusing failure that hides 
the real cause.

## Fix
- `curl -sSLf` now fails immediately on HTTP errors (4xx/5xx)
- Validation checks that lastRelease is not empty or "null"
- Workflow exits with a clear error message before kubectl ever runs

Fixes #12875

### DIFF
--- a/.github/workflows/upgrade-test.yml
+++ b/.github/workflows/upgrade-test.yml
@@ -58,7 +58,17 @@ jobs:
         shell: bash
         id: get-tag
         run: |
-          lastRelease=$(curl -sSL -H "Accept: application/vnd.github+json" "https://api.github.com/repos/kubeflow/pipelines/releases/latest" | jq -r .tag_name)
+          lastRelease=$(curl -sSLf \
+            -H "Accept: application/vnd.github+json" \
+            "https://api.github.com/repos/kubeflow/pipelines/releases/latest" \
+            | jq -r .tag_name)
+
+          if [[ -z "$lastRelease" || "$lastRelease" == "null" ]]; then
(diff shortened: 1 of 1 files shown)


gaurav712-gujjar:
Hi maintainers! 👋 

This is my first contribution to kubeflow/pipelines. 
This PR adds error handling to the GitHub API call in upgrade-test.yml to fix #12875.

Could someone please add the `ok-to-test` label so CI can run? Thank you!

PR #12915, opened 2026-02-25
test(e2e): migrate frontend integration tests to WDIO v9 and expand coverage

jeffspahr:
## Summary
This PR modernizes frontend integration test coverage and upgrades the WebDriverIO stack used by `test/frontend-integration-test`.

## Why
- `tensorboard-example.spec.js` was not being executed, and it used legacy sync WDIO APIs.
- PR changes under `test/frontend-integration-test/**` did not trigger the frontend e2e workflow.
- The WDIO v8 dependency graph contained vulnerable transitive packages (notably minimatch paths).

## What changed
- Add `test/frontend-integration-test/**` to PR path filters in `.github/workflows/e2e-test-frontend.yml`.
- Run both frontend integration specs in `test/frontend-integration-test/wdio.conf.js`:
  - `helloworld.spec.js`
  - `tensorboard-example.spec.js`
- Rewrite `tensorboard-example.spec.js` to async/await WDIO APIs and align selectors/flow with current UI patterns.
- Upgrade WDIO ecosystem dependencies in `test/frontend-integration-test/package.json` + lockfile:
  - `@wdio/cli`, `@wdio/junit-reporter`, `@wdio/local-runner`, `@wdio/mocha-framework`, `@wdio/spec-reporter`, `webdriverio` to `^9.24.0`
- Remove `@wdio/selenium-standalone-service` (only used in DEBUG/local path; no v9 line).
- Update WDIO config for v9 (`host` -> `hostname`, remove deprecated fields).
- Re-enable pipeline cleanup in specs now that pipeline names are timestamp-unique.

## Verification
- `node --check test/frontend-integration-test/wdio.conf.js`
- `node --check test/frontend-integration-test/tensorboard-example.spec.js`
- `node --check test/frontend-integration-test/helloworld.spec.js`

## Notes
- Full frontend integration e2e run was not executed locally here because it requires a live KFP cluster deployment and port-forwarded frontend endpoint.

### DIFF
--- a/test/frontend-integration-test/helloworld.spec.js
+++ b/test/frontend-integration-test/helloworld.spec.js
@@ -18,12 +18,11 @@ const {
   buildTableRowSelector,
   clearDefaultInput,
   getValueFromDetailsTable,
-  isSelectorDisplayed,
   saveDebugScreenshot,
   waitForCondition,
   waitForGraphNodeCount,
   waitForHashPrefix,
-  waitForSelectedPipelineVersion,
+  waitForRunPageReady,
 } = require('./test-helpers');
 
 const experimentName = 'helloworld-experiment-' + Date.now();
@@ -38,16 +37,6 @@ const uiTimeout = 5000;
 const runStartTimeout = 30000;
 const runCompletionTimeout = 60000;
 const outputParameterValue = 'Hello world in test';
-const v1RunFormSelectors = {
-  description: '#descriptionInput',
-  message: 'input#newRunPipelineParam0',
-  runName: '#runNameInput',
-};
-const v2RunFormSelectors = {
-  description: '//label[normalize-space()="Description"]/following::*[self::textarea or self::input][1]',
-  message: '#message',
-  runName: '//label[starts-with(normalize-space(), "Run name")]/following::input[1]',
-};
 
 async function selectPipelineForRun() {
   await $('#choosePipelineBtn').waitForDisplayed({ timeout: uiTimeout });
@@ -84,41 +73,6 @@ async function selectPipelineForRun() {
   await $('#pipelineSelectorDialog').waitForDisplayed({ timeout: uiTimeout, reverse: true });
 }
 
-async function waitForRunFormSelectors() {
-  let selectors;
-
-  try {
-    await waitForCondition(
-      async () => {
-        if (
-          (await isSelectorDisplayed(v1RunFormSelectors.runName)) &&
-          (await isSelectorDisplayed(v1RunFormSelectors.description))
-        ) {
-          selectors = v1RunFormSelectors;
-          return true;
-        }
-        if (
-          (await isSelectorDisplayed(v2RunFormSelectors.runName)) &&
-          (await isSelectorDisplayed(v2RunFormSelectors.description))
-        ) {
-          selectors = v2RunFormSelectors;
-          return true;
-        }
-        return false;
-      },
-      {
-        timeout: uiTimeout,
-        timeoutMsg: 'expected a run creation form to load',
-      },
-    );
-  } catch (error) {
-    await saveDebugScreenshot('run-creation-form');
-    throw error;
-  }
-
-  return selectors;
-}
-
 async function waitForRunParameterField(selector) {
   try {
     await $(selector).waitForDisplayed({ timeout: runStartTimeout });
@@ -129,7 +83,11 @@ async function waitForRunParameterField(selector) {
 }
 
 async function fillRunForm({ runName, description, message }) {
-  const selectors = await waitForRunFormSelectors();
+  const runFormVariant = await waitForRunPageReady({
+    timeout: runStartTimeout,
+    timeoutMsg: 'expected a run creation form to load',
+  });
+  const selectors = runFormVariant.selectors;
 
   await $(selectors.runName).click();
   await clearDefaultInput();
@@ -202,7 +160,6 @@ describe('deploy helloworld sample run', () => {
 
   it('creates a new run in the experiment', async () => {
     await selectPipelineForRun();
-    await waitForSelectedPipelineVersion({ timeout: uiTimeout });
 
     await fillRunForm({
       description: runDescription,
@@ -303,7 +260,6 @@ describe('deploy helloworld sample run', () => {
     await $('#createNewRunBtn').click();
 
     await selectPipelineForRun();
-    await waitForSelectedPipelineVersion({ timeout: uiTimeout });
 
     await fillRunForm({
       description: runWithoutExperimentDescription,
--- a/test/frontend-integration-test/tensorboard-example.spec.js
+++ b/test/frontend-integration-test/tensorboard-example.spec.js
@@ -23,7 +23,7 @@ const {
   waitForCondition,
   waitForGraphNodeCount,
   waitForHashPrefix,
-  waitForSelectedPipelineVersion,
+  waitForRunPageReady,
 } = require('./test-helpers');
 
 const experimentName = 'tensorboard-example-experiment-' + Date.now();
@@ -315,10 +315,12 @@ describe('deploy tensorboard example run', () => {
     await runPhase('create run', async () => {
       await $('#choosePipelineBtn').waitForDisplayed({ timeout: uiTimeout });
       await selectPipelineForRun();
-      await waitForSelectedPipelineVersion({ timeout: uiTimeout });
+      const runFormVariant = await waitForRunPageReady({
+        timeout: runStartTimeout,
+        timeoutMsg: 'expected a run creation form to load',
+      });
 
-      await $('#runNameInput').waitForDisplayed({ timeout: uiTimeout });
-      await $('#runNameInput').click();
+      await $(runFormVariant.selectors.runName).click();
       await clearDefaultInput();
       await browser.keys(runName);
       await $('#startNewRunBtn').click();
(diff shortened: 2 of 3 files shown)


test/frontend-integration-test/helloworld.spec.js
```diff
@@ -382,15 +382,30 @@ describe('deploy helloworld sample run', () => {
...
+    await browser.keys(pipelineName);
+
+    const pipelineRowSelector =
+      `//*[@data-testid="table-row"][.//a[normalize-space()="${pipelineName}"]]`;
+
+    await browser.waitUntil(
+      async () => (await $(pipelineRowSelector).isExisting()),
+      waitTimeout,
+      `expected pipeline row for ${pipelineName} after filtering`,
+    );
```
manaswinidas: WDIO v9 breaking change: `waitUntil(condition, timeout, message)` is the legacy v8 positional-argument form. v9 expects `waitUntil(condition, { timeout, timeoutMsg })`. The `timeoutMsg` string on line 402 will be silently ignored under v9.

Suggestion:
```js
await browser.waitUntil(
  async () => (await $(pipelineRowSelector).isExisting()),
  {
    timeout: waitTimeout,
    timeoutMsg: `expected pipeline row for ${pipelineName} after filtering`,
  },
);
```

The same legacy form appears throughout the rest of `helloworld.spec.js` (lines 49, 63, 80, etc.) — those predate this PR but should also be migrated in a follow-up since v8 is gone.

test/frontend-integration-test/tensorboard-example.spec.js
```diff
@@ -13,174 +13,410 @@
...
-function getValueFromDetailsTable(key) {
+async function getValueFromDetailsTable(key) {
   // Find the span that shows the key, get its parent div (the row), then
-  // get that row's inner text, and remove the key
-  const row = $(`span=${key}`).$('..');
-  return row.getText().substr(`${key}\n`.length);
+  // get that row's inner text, and remove the key.
+  const rowText = await $(`span=${key}`).$('..').getText();
+  return rowText.substr(`${key}\n`.length);
 }
```
manaswinidas: DRY: `getValueFromDetailsTable` and `clearDefaultInput` (line 53) are copy-pasted from `helloworld.spec.js`. The pipeline selection/version selection/cleanup flows are also near-identical between the two files.

Consider extracting shared helpers into a `test-helpers.js` module to avoid drift and reduce duplication.

test/frontend-integration-test/tensorboard-example.spec.js
```diff
@@ -13,174 +13,410 @@
...
+}
 
-  it('opens pipeline details', () => {
-    $('.tableRow a').waitForVisible(waitTimeout);
-    browser.execute('document.querySelector(".tableRow a").click()');
-  });
+async function isDisplayed(selector) {
+  const element = await $(selector);
+  return (await element.isExisting()) && (await element.isDisplayed());
+}
```
manaswinidas: Nit: `isDisplayed` shadows WebDriverIO's built-in `element.isDisplayed()`, which could confuse future contributors reading code like `await isDisplayed('button=Start Tensorboard')` vs `await element.isDisplayed()`. A more specific name like `isElementVisible` or `isSelectorDisplayed` would reduce ambiguity.

test/frontend-integration-test/tensorboard-example.spec.js
```diff
@@ -13,174 +13,410 @@
...
+    await $('#pipelinesBtn').click();
+    await waitForHashPrefix('#/pipelines', uiTimeout);
+
+    await $('#tableFilterBox').waitForDisplayed({ timeout: uiTimeout });
+    await $('#tableFilterBox').click();
+    await clearDefaultInput();
+    await browser.keys(pipelineName);
+
+    const pipelineRowSelector = buildTableRowSelector('', pipelineName);
+    await waitForCondition(
```
manaswinidas, line 247: `buildTableRowSelector('', pipelineName)` — empty string as `containerXPath` produces a selector starting with `//*[@data-testid=...]` which works but is a code smell. Consider making the prefix optional:

```js
function buildTableRowSelector(rowLabel, containerXPath = '') {
```

Or at minimum handle the empty case explicitly to avoid the dangling empty prefix.

test/frontend-integration-test/tensorboard-example.spec.js
```diff
@@ -13,174 +13,410 @@
...
+    await deleteUploadedPipeline();
   });
 
-  it('has a 1-node graph', () => {
-    const nodeSelector = '.graphNode';
-    const nodes = $$(nodeSelector).length;
-    assert(nodes === 1, 'should have a 1-node graph, instead has: ' + nodes);
-  });
+  it('deploys the tensorboard example and opens tensorboard', async () => {
+    await $('#createPipelineVersionBtn').waitForDisplayed({ timeout: uiTimeout });
```
manaswinidas: Consolidating ~12 discrete `it()` blocks into a single ~80-line `it()` trades failure diagnostics for sequential-dependency correctness. When this fails in CI, the step that broke is only identifiable from the error message, not the test name.

Consider keeping at least the major phases as separate `it()` blocks (upload, run, tensorboard) — Mocha's `--bail` flag or `this.skip()` can handle sequential dependencies while preserving granularity.

test/frontend-integration-test/wdio.conf.js
```diff
@@ -61,8 +59,7 @@ exports.config = {
       },
     ],
   ],
-  services: debug ? [['selenium-standalone', { drivers: { chrome: 'latest' } }]] : [],
-  specs: ['./helloworld.spec.js'],
-  sync: true,
+  services: [],
+  specs: ['./helloworld.spec.js', './tensorboard-example.spec.js'],
```
manaswinidas: This removes the `DEBUG=1` local development path (`selenium-standalone-service`) with no replacement. Developers who used `DEBUG=1 npm test` to auto-start a local Selenium server will hit a connection-refused error.

Consider either documenting the new local workflow (e.g., manually run `chromedriver`) or adding `@wdio/devtools-service` which has a v9 release and supports local debugging.

manaswinidas (commented):
Nice work modernizing the WDIO stack and resurrecting the tensorboard spec. The async migration, debug screenshots, and `after()` cleanup are solid improvements.

Key items to address:

1. **Breaking**: New `helloworld.spec.js` cleanup uses legacy `waitUntil(cond, timeout, msg)` positional args — v9 expects an options object. The `timeoutMsg` will be silently dropped.
2. **DRY**: `getValueFromDetailsTable`, `clearDefaultInput`, pipeline selection, and cleanup logic are duplicated across both spec files. Extract to a shared module.
3. **Diagnostics**: Single mega-`it()` in tensorboard spec reduces CI failure granularity — consider splitting into 2-3 phase-level tests.
4. **Local dev**: `DEBUG=1` path is broken with no documented alternative after removing `selenium-standalone-service`.

manaswinidas:
This PR needs a rebase and all integration tests should pass

test/frontend-integration-test/helloworld.spec.js
jeffspahr: Addressed. I migrated the frontend integration waits to WDIO v9-safe helpers and removed the legacy positional `waitUntil(condition, timeout, message)` usage from the active specs. `helloworld.spec.js` and `tensorboard-example.spec.js` now go through `waitForCondition(...)` in `test/frontend-integration-test/test-helpers.js`, so the timeout and timeout message are passed with the v9 options-object form. I kept the raw `browser.waitUntil` call isolated inside that helper instead of duplicating it across specs.

test/frontend-integration-test/tensorboard-example.spec.js
jeffspahr: Addressed. I pulled the shared low-level test utilities into `test/frontend-integration-test/test-helpers.js` to stop the spec drift you called out. The extracted helpers now cover the duplicated waits and selector utilities (`waitForCondition`, `clearDefaultInput`, `getValueFromDetailsTable`, `buildTableRowSelector`, screenshots, etc.). I intentionally did not force the full happy-path flows into one abstraction because the helloworld and tensorboard scenarios have already diverged enough that a shared high-level flow would add coupling without reducing risk.

test/frontend-integration-test/tensorboard-example.spec.js
jeffspahr: Addressed. I renamed the selector helper to `isSelectorDisplayed`, which avoids shadowing WDIO's `element.isDisplayed()` and makes the call sites read unambiguously.

test/frontend-integration-test/tensorboard-example.spec.js
jeffspahr: I looked at this one and chose a different shape deliberately. I kept tensorboard as a single scenario-level `it(...)` to avoid going back to the earlier stateful chain of dependent tests, which was producing large cascades and obscuring the first real failure. To offset the diagnostics cost, I added phase-level annotations and screenshot capture (`runPhase(...)`) so CI failures still identify the broken phase directly. I think that keeps the e2e semantics correct while containing tech debt better than reintroducing interdependent `it()` blocks.

test/frontend-integration-test/wdio.conf.js
jeffspahr: Addressed, but via the documentation/runtime path rather than a new service dependency. I added a supported local debug workflow for the frontend integration tests, including the `HEADLESS`/`WDIO_SPECS` controls in `wdio.conf.js` and the documented local flow in `test/frontend-integration-test/README.md`. I chose not to add `@wdio/devtools-service` during this migration because it would introduce a second execution mode while we were still stabilizing the v9 harness itself.

manaswinidas:
`selectPipelineForRun()` is still near-identical in both spec files — the only difference is extra diagnostic logging in the helloworld catch block. This is a reusable mid-level helper, not a full happy-path flow, so it belongs in `test-helpers.js` alongside the other extracted utilities. Could you move it there (with an optional diagnostics flag or callback for the helloworld case)? Alos, is the duplication intentional ?

PR #12919, opened 2026-02-26
feat(backend): Add plugin input/output fields to RecurringRun and Run protobufs

hbelmiro:
Adds a generic plugin metadata interface to `Run` and `RecurringRun` as specified in [KEP-12862](https://github.com/kubeflow/pipelines/pull/12862). This is the schema-only change -- no MLflow or plugin business logic is introduced.

- Add `MetadataValue`, `PluginOutput` messages and `plugins_input` (field 19) / `plugins_output` (field 20) to `Run`
- Add `plugins_input` (field 19) to `RecurringRun`
- Propagate through Go models (`*LargeText`), storage layer (SQL NULL for absent), and API converters
- Regenerate Go gRPC, swagger, and HTTP client code

## Details

- Model fields use `*LargeText` with nil = absent (SQL NULL), never empty string
- Storage write paths use `largeTextToNullableSQL`; read paths use pointer assignment
- `GetRun`/`ListRuns`/`GetRecurringRun` return plugins fields as stored, no external calls
- Existing function signatures (`toApiRun`, `toApiRecurringRun`, etc.) are unchanged

## Test plan

- Unit tests for JSON serialization helpers (nil, empty, valid, malformed)
- Converter tests for `toModelRun`, `toApiRun`, `toModelJob`, `toApiRecurringRun` (with data + nil)
- Storage round-trip tests (create, get, update, list) for both runs and jobs
- DB-level NULL assertions verifying absent fields produce SQL NULL, not empty string
- Proto golden file tests updated

**Checklist:**

### DIFF
--- a/backend/src/apiserver/common/config.go
+++ b/backend/src/apiserver/common/config.go
@@ -15,7 +15,9 @@
 package common
 
 import (
+	"fmt"
 	"strconv"
+	"strings"
 	"time"
 
 	"github.com/golang/glog"
@@ -44,7 +46,18 @@ const (
 	DefaultSecurityContextRunAsUser         string = "DEFAULT_SECURITY_CONTEXT_RUN_AS_USER"
 	DefaultSecurityContextRunAsGroup        string = "DEFAULT_SECURITY_CONTEXT_RUN_AS_GROUP"
 	DefaultSecurityContextRunAsNonRoot      string = "DEFAULT_SECURITY_CONTEXT_RUN_AS_NON_ROOT"
+	PluginMaxKeys                           string = "PLUGIN_MAX_KEYS"
+	PluginMaxPayloadBytes                   string = "PLUGIN_MAX_PAYLOAD_BYTES"
+	PluginMaxTotalPayloadBytes              string = "PLUGIN_MAX_TOTAL_PAYLOAD_BYTES"
+	PluginMaxNestingDepth                   string = "PLUGIN_MAX_NESTING_DEPTH"
 )
+
+type PluginLimitsConfig struct {
+	MaxKeys              int
+	MaxPayloadBytes      int
+	MaxTotalPayloadBytes int
+	MaxNestingDepth      int
+}
 
 func IsPipelineVersionUpdatedByDefault() bool {
 	return GetBoolConfigWithDefault(UpdatePipelineVersionByDefault, true)
@@ -99,6 +112,27 @@ func GetIntConfigWithDefault(configName string, value int) int {
 		return value
 	}
 	return viper.GetInt(configName)
+}
+
+func getPositiveIntConfigWithDefault(configName string, value int) (int, error) {
+	if !viper.IsSet(configName) {
+		return value, nil
+	}
+
+	raw := strings.TrimSpace(viper.GetString(configName))
+	if raw == "" {
+		return 0, fmt.Errorf("invalid value for %s: must be a positive integer", configName)
+	}
+
+	parsed, err := strconv.Atoi(raw)
+	if err != nil {
+		return 0, fmt.Errorf("invalid value for %s: %w", configName, err)
+	}
+	if parsed <= 0 {
+		return 0, fmt.Errorf("invalid value for %s: must be > 0", configName)
+	}
+
+	return parsed, nil
 }
 
 func GetDurationConfig(configName string) time.Duration {
@@ -186,4 +220,40 @@ func GetDefaultSecurityContextRunAsGroup() string {
 
 func GetDefaultSecurityContextRunAsNonRoot() string {
 	return GetStringConfigWithDefault(DefaultSecurityContextRunAsNonRoot, "")
+}
+
+func GetPluginLimitsConfig() (PluginLimitsConfig, error) {
+	maxKeys, err := getPositiveIntConfigWithDefault(PluginMaxKeys, DefaultPluginMaxKeys)
+	if err != nil {
+		return PluginLimitsConfig{}, err
+	}
+	maxPayloadBytes, err := getPositiveIntConfigWithDefault(PluginMaxPayloadBytes, DefaultPluginMaxPayloadBytes)
+	if err != nil {
+		return PluginLimitsConfig{}, err
+	}
+	maxTotalPayloadBytes, err := getPositiveIntConfigWithDefault(PluginMaxTotalPayloadBytes, DefaultPluginMaxTotalPayloadBytes)
+	if err != nil {
+		return PluginLimitsConfig{}, err
+	}
+	maxNestingDepth, err := getPositiveIntConfigWithDefault(PluginMaxNestingDepth, DefaultPluginMaxNestingDepth)
+	if err != nil {
+		return PluginLimitsConfig{}, err
+	}
+
+	if maxTotalPayloadBytes < maxPayloadBytes {
+		return PluginLimitsConfig{}, fmt.Errorf(
+			"invalid plugin limits: %s (%d) must be >= %s (%d)",
+			PluginMaxTotalPayloadBytes,
+			maxTotalPayloadBytes,
+			PluginMaxPayloadBytes,
+			maxPayloadBytes,
+		)
+	}
+
+	return PluginLimitsConfig{
+		MaxKeys:              maxKeys,
+		MaxPayloadBytes:      maxPayloadBytes,
+		MaxTotalPayloadBytes: maxTotalPayloadBytes,
+		MaxNestingDepth:      maxNestingDepth,
+	}, nil
 }
--- a/backend/src/apiserver/common/config_test.go
+++ b/backend/src/apiserver/common/config_test.go
@@ -19,6 +19,7 @@ import (
 
 	"github.com/spf13/viper"
 	"github.com/stretchr/testify/assert"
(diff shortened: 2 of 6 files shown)


backend/api/v2beta1/run.proto
```diff
@@ -258,6 +266,26 @@ message RuntimeStatus {
...
 }
 
+// A typed metadata value with an optional rendering hint for the UI.
+message MetadataValue {
+  // Hint for UI rendering of the value.
+  enum ContentType {
+    // Default. No special rendering.
+    UNSPECIFIED = 0;
+    // Render the value as a hyperlink.
+    URL = 1;
```
nsingla, line 276: Don't we need number, text types here?

backend/src/apiserver/model/job.go
```diff
@@ -109,7 +109,8 @@ type Job struct {
 	ResourceReferences []*ResourceReference `gorm:"-"`
 	Trigger
 	PipelineSpec
-	Conditions string `gorm:"column:Conditions; not null;"`
+	Conditions         string     `gorm:"column:Conditions; not null;"`
+	PluginsInputString *LargeText `gorm:"column:PluginsInput; default:null;"`
```
nsingla, line 113: This might be my lack of knowledge, why are we adding pluginsinput to Job?

hbelmiro: `Job` is the Go model for `RecurringRun`.

backend/api/v2beta1/run.proto
hbelmiro: `ContentType` is a UI rendering hint, not a data type. The value itself is `google.protobuf.Value` which already carries type information (string, number, etc.). `UNSPECIFIED` means "render as-is" which covers plain text and numbers. `URL` is the only case that needs special rendering (hyperlink).
Maybe we should renamen it from `ContentType` to `UIHint` or something to avoid confusing. But that should be done in the KEP first.

backend/api/v2beta1/run.proto
```diff
@@ -200,6 +200,14 @@ message Run {
   // Output. A sequence of run statuses. This field keeps a record
   // of state transitions.
   repeated RuntimeStatus state_history = 17;
+
+  // Optional input. Plugin-specific inputs provided by the user at run creation.
+  // Each key is a plugin name (e.g., "mlflow") and the value is arbitrary JSON config.
+  map<string, google.protobuf.Struct> plugins_input = 19;
```
alyssacgoins, line 206: Same as above

backend/api/v2beta1/recurring_run.proto
```diff
@@ -154,6 +154,10 @@ message RecurringRun {
 
   // ID of the parent experiment this recurring run belongs to.
   string experiment_id = 17;
+
+  // Optional input. Plugin inputs to propagate to each triggered run.
+  // Each triggered run will inherit these values in its plugins_input field.
+  map<string, google.protobuf.Struct> plugins_input = 19;
```
alyssacgoins, line 160: Is there a particular reason why the number skips to **19** instead of **18**?

hbelmiro: Field 18 is `pipeline_version_reference`, defined inside the `oneof pipeline_source` block. The same is true in `Run` where field 18 is also `pipeline_version_reference`. So field 19 is the next available number in both messages.

backend/api/v2beta1/run.proto
hbelmiro: Field 18 is `pipeline_version_reference`, defined inside the `oneof pipeline_source` block. The same is true in `Run` where field 18 is also `pipeline_version_reference`. So field 19 is the next available number in both messages.

VaniHaripriya:
@hbelmiro Is AutoMigrate sufficient for adding plugins_input and plugins_output to run_details table, or should we include an explicit DB migration script to avoid schema discrepancies ?

hbelmiro:
> @hbelmiro Is AutoMigrate sufficient for adding plugins_input and plugins_output to run_details table, or should we include an explicit DB migration script to avoid schema discrepancies ?

@VaniHaripriya AutoMigrate is sufficient. But anyway, I've added a `TestAutoMigrateSucceeds` test that verifies the schema is created correctly.

backend/api/v2beta1/run.proto
nsingla: Got it, can you please comment on the KEP to make the change? I think its confusing from purely code pov and when we start to use these protos for generic plugin development in the future, it will be more confusing

backend/src/apiserver/model/job.go
nsingla: Ah! :+1:

nsingla:
Few questions:
1. Don't we need to update UpdateRun to include plugins_input and plugins_output metadata?
2. I know the same methodology is used elsewhere, but I think we need to add some kind of upper bound to the open google.protobuf.Struct values
3. We need to validate/parse URL, in plugins_output in case a custom plugin sets it to an arbitrary url with javascript injected in it

backend/api/v2beta1/go_client/run.pb.go
```diff
@@ -544,6 +613,120 @@ func (x *RuntimeStatus) GetError() *status.Status {
...
+		return *x.ContentType
+	}
+	return MetadataValue_UNSPECIFIED
+}
+
+// Output from a single plugin, containing keyed metadata entries and overall state.
+type PluginOutput struct {
+	state         protoimpl.MessageState    `protogen:"open.v1"`
+	Entries       map[string]*MetadataValue `protobuf:"bytes,1,rep,name=entries,proto3" json:"entries,omitempty" protobuf_key:"bytes,1,opt,name=key" protobuf_val:"bytes,2,opt,name=value"`
+	State         RuntimeState              `protobuf:"varint,2,opt,name=state,proto3,enum=kubeflow.pipelines.backend.api.v2beta1.RuntimeState" json:"state,omitempty"`
```
nsingla: Should we be reusing runtime state? mostly concerned about `canceling`, if a pipeline run is cancelled mid way, what does that mean for a plugin? 🤔 should it be `skipped` for a plugin in this case?

backend/api/v2beta1/run.proto
hbelmiro: Done.
https://github.com/kubeflow/pipelines/pull/12862/changes#r2874321105

hbelmiro:
/hold
Until I address review comments.

hbelmiro:
> I know the same methodology is used elsewhere, but I think we need to add some kind of upper bound to the open google.protobuf.Struct values

@nsingla I'm not sure I understand. Can you please elaborate?

backend/api/v2beta1/go_client/run.pb.go
hbelmiro: `plugins_output.mlflow.state` tracks whether the plugin's own operations succeeded or failed, not the pipeline run state. When a pipeline run is canceled, `ReportWorkflowResource` closes out MLflow runs and sets the plugin state to SUCCEEDED (cleanup worked) or FAILED (cleanup failed). CANCELING/CANCELED/PAUSED/SKIPPED are never written by the MLflow integration. Reusing `RuntimeState` avoids a near-duplicate enum while the KEP only uses the SUCCEEDED/FAILED/RUNNING subset.

hbelmiro:
> Few questions:
> 
>     1. Don't we need to update UpdateRun to include plugins_input and plugins_output metadata?
> 
>     2. I know the same methodology is used elsewhere, but I think we need to add some kind of upper bound to the open google.protobuf.Struct values
> 
>     3. We need to validate/parse URL, in plugins_output in case a custom plugin sets it to an arbitrary url with javascript injected in it

@nsingla 1 and 2 addressed. [3 needs clarification](https://github.com/kubeflow/pipelines/pull/12919#issuecomment-3986757672).

backend/api/v2beta1/run.proto
hbelmiro: Renamed to `RenderType` in the KEP.

backend/api/v2beta1/go_client/run.pb.go
nsingla: I assumed that would be case, but just reading the code (not the intent), it's confusing and I unintentionally tied it with Run's state

I think its cleaner to have a separate Enum, imo

nsingla:
> > I know the same methodology is used elsewhere, but I think we need to add some kind of upper bound to the open google.protobuf.Struct values
> 
> @nsingla I'm not sure I understand. Can you please elaborate?

What I meant was upper bound in terms of size & depth, otherwise, somecan create 100 deep (1000s of key/value pairs) blob of json in multiple mbs of size

backend/api/v2beta1/go_client/run.pb.go
hbelmiro: Yeah, makes total sense given that.
Done.

hbelmiro:
> > > I know the same methodology is used elsewhere, but I think we need to add some kind of upper bound to the open google.protobuf.Struct values
> > 
> > 
> > @nsingla I'm not sure I understand. Can you please elaborate?
> 
> What I meant was upper bound in terms of size & depth, otherwise, somecan create 100 deep (1000s of key/value pairs) blob of json in multiple mbs of size

@nsingla done.

PR #12920, opened 2026-02-26
docs(migration): Design document for the MLMD database migration

VaniHaripriya:

Design document for MLMD database migrations as described in [this](https://github.com/kubeflow/pipelines/blob/master/proposals/12147-mlmd-removal/README.md) proposal

**Checklist:**

### DIFF
--- a/proposals/12147-mlmd-removal/database-migration-design.md
+++ b/proposals/12147-mlmd-removal/database-migration-design.md
@@ -1,0 +1,458 @@
+# MLMD Database Migration Design
+
+## Overview
+
+This document describes the design for migrating historical pipeline metadata from ML Metadata (MLMD) to native KFP database tables as part of the MLMD removal effort.
+
+## Motivation
+
+When upgrading KFP to a version without MLMD, existing deployments have historical data (executions, artifacts, events) stored in MLMD that must be preserved. This migration design addresses the challenge of moving this data to the new KFP native schema with the following goals:
+- Minimizing operational disruption (zero-downtime migration)
+- Providing flexibility for different operational constraints (manual vs automatic execution)
+- Ensuring data integrity and safety (idempotent, resumable on failure)
+
+## Goals
+
+1. **Idempotent Migration**: Can run multiple times, migrating only new data since last run
+2. **Zero-Downtime**: Migration runs while API server serves traffic
+3. **Dual Execution Modes**: Support both standalone CLI tool and API server integration
+4. **Safe for Multi-Replica**: Database locks prevent concurrent migration
+5. **Resumable**: Can resume after interruption (pod crash, SIGTERM)
+
+## Design Details
+
+### Architecture
+
+The migration system consists of:
+
+1. **Migration Engine** (`backend/src/apiserver/migration/`)
+   - Core migration logic used by both execution modes   
+
+2. **Standalone CLI Tool** (`backend/cmd/mlmd-migrate/`)
+   - Compiled Go binary for manual execution
+   - Wrapper that parses CLI flags, calls migration engine, exits
+
+3. **API Server Integration** (`backend/src/apiserver/main.go`)
+   - Wrapper for automatic execution
+   - Checks env vars, spawns background goroutine
+
+**Note:** Both modes use the **exact same migration logic**. Only the entry point differs.
+
+### File Structure
+
+```
+backend/src/apiserver/
+├── migration/                          # NEW: Shared migration engine
+│   ├── manager.go                      # Migration orchestration & state machine
+│   ├── transformers.go                 # MLMD entity → KFP model transformations
+│   ├── checkpoint.go                   # Idempotency tracking & resume logic
+│   ├── validation.go                   # Pre/post migration validation
+│   └── models.go                       # Migration tracking database models
+│
+├── model/
+│   ├── migration_status.go             # NEW: Migration status tracking
+│   ├── migration_checkpoint.go         # NEW: Resume checkpoint tracking
+│   └── mlmd_id_map.go                  # NEW: MLMD ID → KFP UUID mapping
+│
+└── main.go                             # MODIFIED: Add migration check on startup
+
+backend/cmd/
+└── mlmd-migrate/                       # NEW: Standalone CLI tool
+    └── main.go                         # CLI wrapper around migration engine
+```
+
+### Migration Data Flow
+
+```
+MLMD Database                 KFP Database
+━━━━━━━━━━━━━                 ━━━━━━━━━━━━━━━━━━━━
+Contexts (Runs)      ──►  Validate against run_details (no insert)
+
+Executions           ──►  tasks table
+                          - Generate deterministic UUID
+                          - Map state, timestamps, parameters
+                          - Detect TaskType (ROOT, RUNTIME, LOOP, etc.)
+
+Artifacts            ──►  artifacts table
+                          - Split metrics into separate rows
+                          - Map to ArtifactType enum
+                          - Store metrics with NumberValue field
+
+Events               ──►  artifact_tasks table  (LAST — has FKs to both tasks and artifacts)
+                          - Link artifacts to tasks
+                          - Preserve IOType and producer info
+```
+
+> **Dependency order is critical.** `artifact_tasks` has foreign keys to both `tasks` and `artifacts`, so it must be migrated last. Migrating it before its dependencies would cause FK constraint violations.
+
+### Database Schema Changes
+
+#### New Tables for Migration Tracking
+
+**`migration_status` table:**
+```sql
+CREATE TABLE migration_status (
+    ID               int AUTO_INCREMENT PRIMARY KEY,
+    MigrationName    varchar(128) NOT NULL UNIQUE,
+    Status           varchar(32) NOT NULL,  -- NOT_STARTED, IN_PROGRESS, COMPLETED, FAILED
+    StartedAt        bigint,
+    CompletedAt      bigint,
+    LastCheckpoint   text,                   -- JSON checkpoint data
+    ErrorMessage     text,
+    TotalEntities    bigint DEFAULT 0,
+    MigratedEntities bigint DEFAULT 0,
+    CurrentStage     varchar(64)
+);
+```
+
+**Purpose:** Track migration progress and status across restarts.
+
+**`mlmd_id_map` table:**
+```sql
+CREATE TABLE mlmd_id_map (
+    mlmd_entity_type varchar(20) NOT NULL,  -- "execution" or "artifact"
+    mlmd_id          bigint NOT NULL,
+    kfp_uuid         varchar(191) NOT NULL,
+    migrated_at      bigint NOT NULL,
+    PRIMARY KEY (mlmd_entity_type, mlmd_id)
+);
+```
+
+**Purpose:** Map MLMD integer IDs to KFP UUIDs for:
+- Resolving parent task references
+- Rollback support
+- Audit trail
+
+**Lifecycle:** Temporary table, can be dropped after 30-90 day retention period.
+
+### Idempotency Strategy
+
+Migration can be run multiple times safely without creating duplicate records. This is achieved through three mechanisms:
+
+#### 1. Deterministic UUID Generation
+
+**Problem:** MLMD uses integer IDs (e.g., `execution_id = 12345`), but KFP uses UUIDs (e.g., `task_uuid = "abc-def-123"`). We need to convert IDs consistently.
+
+**Solution:** Use UUID v5 (SHA1-based) to generate the same UUID every time for the same MLMD ID.
+
+**How it works:**
+```
+Input:  MLMD Execution ID = 12345
+Step 1: Create string = "execution:12345"
+Step 2: Hash with SHA1 using DNS namespace = a1b2c3d4...
+Step 3: Format as UUID v5 = "a1b2c3d4-e5f6-5789-abcd-ef0123456789"
+
+Key property: Same input (12345) ALWAYS produces same UUID
+```
+
+**Example:**
+```
+First run:  Execution 12345 → UUID "a1b2c3d4-e5f6-5789-abcd-ef0123456789"
+Second run: Execution 12345 → UUID "a1b2c3d4-e5f6-5789-abcd-ef0123456789" (identical!)
+```
+
+**Why this matters:** If migration crashes and reruns, the same MLMD execution generates the same UUID, so `INSERT IGNORE` will skip it (no duplicate).
+
+**In-flight data safety:** The new API server writes new runs using random UUID v4, while migrated records use deterministic UUID v5. These UUID spaces never collide. Combined with `INSERT IGNORE`, this means the migration can never overwrite a record created by the live API server,even if both are writing to the `tasks` table simultaneously.
+
+#### 2. Safe Insert Pattern
+
+Use MySQL's `INSERT IGNORE` to skip records that already exist:
+
+```sql
+INSERT IGNORE INTO tasks (UUID, Name, State, ...) VALUES (...);
+```
+
+**Behavior:**
+- If UUID already exists: Skip insert (RowsAffected = 0), no error
+- If UUID is new: Insert record (RowsAffected = 1)
+
+**Example scenario:**
+```
+Run 1: Migrates executions 1-1000 (inserts 1000 tasks)
+       Crashes at execution 500
+Run 2: Tries to migrate executions 1-1000 again
+       - Executions 1-500: Same UUIDs, INSERT IGNORE skips them
+       - Executions 501-1000: New inserts succeed
+Result: No duplicates, migration completes!
+```
+
+#### 3. Checkpoint-Based Resume
+
+Track migration progress to avoid re-processing large batches:
+
+**Checkpoint structure:**
+```json
+{
+  "last_execution_id": 1000,
+  "last_artifact_id": 5000,
+  "last_migration_timestamp": 1709000000
+}
+```
+
+**Resume logic:**
+```sql
+-- Query MLMD for only unprocessed entities
+SELECT * FROM Execution
+WHERE id > 1000  -- Resume from checkpoint
+ORDER BY id ASC
+```
+
+**Combined with deterministic UUIDs:**
+- Checkpoint reduces redundant work (skip querying already-migrated entities)
+- Deterministic UUIDs ensure safety if checkpoint is lost (INSERT IGNORE prevents duplicates)
+
+### MLMD Entity Transformations
+
+#### Executions → Tasks
+
+**Task Type Detection:**
+
+| MLMD Execution Type | Detection Logic | KFP TaskType |
+|---------------------|----------------|--------------|
+| `system.DAGExecution` (root) | No `parent_dag_id` | ROOT (0) |
+| `system.ContainerExecution` | Type check | RUNTIME (1) |
+| `system.DAGExecution` (condition-branches-*) | Name prefix | CONDITION_BRANCH (2) |
+| `system.DAGExecution` (condition-*) | Name prefix, child of CONDITION_BRANCH | CONDITION (3) |
+| `system.DAGExecution` + `iteration_count` | Custom property | LOOP (4) |
+| `system.DAGExecution` (exit-handler-*) | Name prefix + fallbacks (see below) | EXIT_HANDLER (5) |
+| `system.ImporterExecution` | Type check | IMPORTER (6) |
+| `system.DAGExecution` (with parent) | Has `parent_dag_id` | DAG (7) |
+
+> **Detection order matters.** Check specific patterns (exit-handler, condition, loop) *before* falling back to generic DAG. Exit handler detection uses name prefix `exit-handler-*` as the primary signal, with fallbacks: `on-exit-*` or `onexit-*` prefixes, and `exit_handler` custom property.
+
+**MLMD State → KFP TaskState Mapping:**
+
+| MLMD Execution State | KFP TaskState | Notes |
+|---------------------|---------------|-------|
+| `NEW` | RUNTIME_STATE_UNSPECIFIED (0) | Execution created but not started |
+| `RUNNING` | RUNNING (1) | |
+| `COMPLETE` | SUCCEEDED (2) | |
+| `FAILED` | FAILED (4) | |
+| `CACHED` | CACHED (5) | |
+| `CANCELED` | FAILED (4) | No CANCELED in KFP TaskState; maps to FAILED |
+
+> **Note:** `SKIPPED (3)` is not mapped from MLMD — it is only used by the new API server for conditional branches that were not taken.
+
+**Namespace Resolution:**
+
+MLMD sub-executions don't always carry the namespace. The migration resolves it in priority order:
+1. Execution's own `namespace` custom property (if present)
+2. Fallback: parent `run_details.Namespace` (always available)
+
+This ensures `Namespace NOT NULL` constraints are satisfied for all migrated tasks and artifacts.
+
+**Cache Fingerprint Handling:**
+- Only set `Fingerprint` field for executions in `COMPLETE` or `CACHED` state
+- Leave fingerprint empty for `FAILED` or `RUNNING` executions
+- Per upstream requirement: prevent caching from incomplete executions
+
+**Example Transformation:**
+```
+MLMD Execution:
+  id: 12345
+  type: "system.ContainerExecution"
+  name: "preprocess-data"
+  state: COMPLETE
+  cache_fingerprint: "abc123"
+
+→ KFP Task:
+  UUID: "deterministic-uuid-for-12345"
+  Type: RUNTIME (1)
+  Name: "preprocess-data"
+  State: SUCCEEDED (2)
+  Fingerprint: "abc123"  ← Only set because state is COMPLETE
+```
+
+#### Artifacts → Artifacts
+
+**Metrics Handling:**
+- MLMD stores metrics as single artifact with multiple key-value pairs
+- KFP stores each metric as **separate artifact row** with `Type=Metric` and `NumberValue`
+
+**Example:**
+```
+MLMD Artifact (id: 100):
+  type: "system.Metrics"
+  custom_properties: {
+    "accuracy": 0.95,
+    "loss": 0.05
+  }
+
+→ KFP Artifacts:
+  [
+    {
+      UUID: "deterministic-uuid-100-accuracy",
+      Type: Metric (6),
+      NumberValue: 0.95
+    },
+    {
+      UUID: "deterministic-uuid-100-loss",
+      Type: Metric (6),
+      NumberValue: 0.05
+    }
+  ]
+```
+
+**Why separate rows?** Enables efficient querying and indexing of individual metrics.
+
+**ID Mapping for split metrics:** When a single MLMD artifact produces multiple KFP rows, only the *first* metric's UUID is registered in `mlmd_id_map`. This is the UUID that `artifact_tasks` will reference when linking Events to the original MLMD artifact.
+
+### Execution Modes
+
+#### Mode 1: Standalone CLI Tool (Manual)
+
+**Usage:**
+```bash
+mlmd-migrate \
+  --mlmd-address=metadata-grpc-service:8080 \
+  --mysql-host=mysql:3306 \
+  --batch-size=5000 \
+  --workers=8
+```
+
+**Behavior:**
+- Runs in foreground (blocks until complete)
+- Exits with status code (0=success, non-zero=failure)
+- Admin has full control over timing and resources
+
+**Configuration:** CLI flags
+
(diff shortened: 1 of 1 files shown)


alyssacgoins:
@HumairAK @mprahl - 

@VaniHaripriya  and I went over the pros/cons of blocking apiserver startup during migration vs cancelling pipeline runs midflight to be retried after migration completes. The latter has zero-downtime, but also risks data loss, given that we would need a mechanism/logic to delay retry until migration completes successfully. What are your thoughts?

mprahl:
> @HumairAK @mprahl -
> 
> @VaniHaripriya and I went over the pros/cons of blocking apiserver startup during migration vs cancelling pipeline runs midflight to be retried after migration completes. The latter has zero-downtime, but also risks data loss, given that we would need a mechanism/logic to delay retry until migration completes successfully. What are your thoughts?

Perhaps a better way than cancelling the pipeline runs outright is to [suspend](https://argo-workflows.readthedocs.io/en/latest/walk-through/suspending/) the Argo Workflow objects so the current inflight steps finish, then cancel the pipeline runs once they are all suspended, then upgrade, then retry the pipeline runs. The cancel and retry is needed to recompile the Argo Workflow objects to use the latest driver and launcher.

During the migration, I'd expect the KFP API server to be down. Perhaps we can add a "maintenance mode" to KFP which has all of its API endpoints return "This Kubeflow Pipelines is down for maintenance. Try again later." or similar.

The API server should refuse start up only if it can't migrate (e.g. no MLMD connection info).

What do you think @HumairAK?

proposals/12147-mlmd-removal/database-migration-design.md
```diff
@@ -0,0 +1,463 @@
...
+
+## Proposal
+
+### User Stories
+
+**Story 1: Admin upgrading an existing KFP deployment**
+
+> As a KFP platform admin, I want to upgrade from a MLMD-backed version of KFP to a MLMD-free version without losing any historical pipeline run data or artifact lineage, so that my team can continue to audit and debug past runs after the upgrade.
+
+**Story 2: Admin upgrading during a maintenance window**
```
mprahl: I think this is not a user story but rather a limitation of the upgrade experience. I think we can remove this user story.

proposals/12147-mlmd-removal/database-migration-design.md
```diff
@@ -0,0 +1,463 @@
...
+
+---
+
+## Goals
+
+1. **Data preservation** — All executions, artifacts, and events recorded in MLMD are migrated into the KFP native schema with full fidelity.
+2. **Safe upgrade** — During migration the API runs in maintenance mode (see below). Normal traffic resumes only after migration completes successfully. Startup fails only when migration is required but cannot run (for example, missing MLMD connection info).
+3. **Idempotency** — The migration can be run multiple times. Repeated runs produce no duplicates and leave the database in a consistent state.
+4. **Resumability** — A migration interrupted by a pod crash, SIGTERM, or node failure can resume from a checkpoint without replaying previously migrated entities.
+5. **Concurrency safety** — In HA deployments, and when the standalone CLI and API server run concurrently, exactly one process performs the migration at a time.
```
mprahl: I don't think we need to define concurrency safety. I doubt that the Gorm automigrate handles this well today. If it does, then we should consider that.

proposals/12147-mlmd-removal/database-migration-design.md
```diff
@@ -0,0 +1,463 @@
...
+
+#### Architecture
+
+The migration system has three components that share a single migration engine:
+
+```
+┌─────────────────────────────────────────────────────────────┐
+│                    Migration Engine                          │
+│            backend/src/apiserver/migration/                 │
+│  manager.go · transformers.go · checkpoint.go · validation.go│
```
mprahl: Let's not include file names in the KEP. That is an implementation detail.

proposals/12147-mlmd-removal/database-migration-design.md
```diff
@@ -0,0 +1,463 @@
...
+│  backend/cmd/       │       │  backend/src/apiserver│
+│  mlmd-migrate/      │       │  main.go              │
+│  main.go            │       │  (migrate under       │
+│  (manual, blocking) │       │   maintenance mode)   │
+└────────────────────┘       └──────────────────────┘
+```
+
+Both execution modes use identical migration logic. Only the entry point and configuration source differ.
+
+#### File Structure
```
mprahl: We don't need this level of detail in the KEP. What goes into which file can be decided during the implementation.

proposals/12147-mlmd-removal/database-migration-design.md
```diff
@@ -0,0 +1,463 @@
...
+
+- Verify MLMD gRPC connectivity.
+- Verify MySQL connectivity and sufficient disk space (estimated ~2× MLMD database size).
+- Confirm all MLMD contexts (runs) have a corresponding row in `run_details`. Orphaned MLMD records (contexts without a `run_details` counterpart) are logged as warnings and skipped rather than failing the entire migration. This handles common partial-failure states from past runs.
+
+**Phase 2 — Schema migration**
+
+Schema changes are handled differently depending on whether a table is new or pre-existing, to avoid GORM `AutoMigrate` silently reverting or corrupting manual modifications to existing tables:
+
+- **New tables** (`artifacts`, `artifact_tasks`, `migration_status`, `mlmd_id_map`): created via manually written SQL (see [schema_changes.sql](./schema_changes.sql)), verified against GORM `AutoMigrate` in CI. No pre-existing data or schema to protect.
```
mprahl: For "verified against GORM `AutoMigrate` in CI", do you mean we are ensuring GORM does not detect a schema diff?

proposals/12147-mlmd-removal/database-migration-design.md
```diff
@@ -0,0 +1,463 @@
...
+- **Modified existing table** (`tasks` only): migrated via a table-copy-and-rename workflow rather than in-place `AutoMigrate`. The current `tasks` table stores cache fingerprints only; the post-migration schema is incompatible (see [schema_changes.sql](./schema_changes.sql)):
+  1. Create a temporary copy of the table (`tasks_migration_tmp`) with the target schema, using manually written SQL (not `AutoMigrate`) as the source of truth for the DDL.
+  2. Backfill `tasks_migration_tmp` from the existing `tasks` table (cache-fingerprint rows), mapping old columns into the new schema.
+  3. Within a single transaction: rename the existing table to a backup name (`tasks_pre_migration`), rename the temporary table to `tasks`.
+  4. Retain `tasks_pre_migration` for the same 30–90 day rollback window as `mlmd_id_map` (see Rollback).
```
mprahl: I think we don't need to support rollbacks as we don't for other schema changes today. We should suggest the admin backup the DB prior to migration.

proposals/12147-mlmd-removal/database-migration-design.md
```diff
@@ -0,0 +1,463 @@
...
+MLMD Artifacts   ──►  artifacts table
+                       (metrics split into separate rows, register in mlmd_id_map)
+
+MLMD Events      ──►  artifact_tasks table       ← LAST: FK deps on both above
+                       (link tasks to artifacts, preserve IOType)
+
+Second pass      ──►  Resolve parent_task_uuid references via mlmd_id_map
+```
+
+Each entity type is processed in serial batches (controlled by `--batch-size` / `MIGRATION_BATCH_SIZE`) in ascending MLMD ID order. Parallel workers are intentionally not used, so a single `last_*_id` checkpoint cursor accurately represents resume progress.
```
mprahl: Could you add that each batch is its own transaction if it's not a destructive operation? This allows an interrupted migration to pick up where it left off on the next run.

proposals/12147-mlmd-removal/database-migration-design.md
```diff
@@ -0,0 +1,463 @@
...
+
+In API Server mode, Phase 3 runs while the server remains in maintenance mode. Normal API traffic resumes only after Phase 3 (and Phase 4 validation) complete successfully.
+
+**Phase 4 — Post-migration validation**
```
mprahl: I don't think we need this step. We should rely on CI/tests to verify the migration works properly. This will unnecessarily add to the downtime of KFP IMO.

proposals/12147-mlmd-removal/database-migration-design.md
```diff
@@ -0,0 +1,463 @@
...
+    CompletedAt      BIGINT,
+    LastCheckpoint   TEXT,                   -- JSON checkpoint (last processed IDs)
+    ErrorMessage     TEXT,
+    TotalEntities    BIGINT DEFAULT 0,
+    MigratedEntities BIGINT DEFAULT 0,
+    CurrentStage     VARCHAR(64)
+);
+```
+
+**`mlmd_id_map`** — maps MLMD integer IDs to KFP UUIDs for parent-task resolution and rollback:
```
mprahl: As mentioned in another comment, no need for rollbacks. You can drop the table right after the migration is complete.

proposals/12147-mlmd-removal/database-migration-design.md
```diff
@@ -0,0 +1,463 @@
...
+);
+```
+
+`mlmd_id_map` is a temporary table and may be dropped 30–90 days after migration completes.
+
+#### Idempotency
+
+Three overlapping mechanisms ensure the migration is safe to run multiple times:
+
+**1. Deterministic UUID generation (UUID v5)**
```
mprahl, line 173: Do we need the migration status table if the UUIDs are deterministic? Perhaps on startup, we can perform a binary search in MLMD to find where the migration left off rather than storing that additional state.

proposals/12147-mlmd-removal/database-migration-design.md
```diff
@@ -0,0 +1,463 @@
...
+
+The same MLMD ID always produces the same UUID. Live API server writes use random UUID v4; these two namespaces never collide.
+
+**2. Safe insert semantics**
+
+```sql
+INSERT IGNORE INTO tasks (UUID, ...) VALUES (...);
+```
+
+Duplicate rows (same UUID) are silently skipped. A crashed-and-resumed run encounters `RowsAffected = 0` for already-migrated rows and `RowsAffected = 1` for new ones — no errors, no duplicates.
```
mprahl, line 192: This is fine, but do we actually need that if we are using transactions for each batch and we are able to find where the migration left off?

proposals/12147-mlmd-removal/database-migration-design.md
```diff
@@ -0,0 +1,463 @@
...
+```bash
+mlmd-migrate \
+  --mlmd-address=metadata-grpc-service:8080 \
+  --mysql-host=mysql:3306 \
+  --batch-size=5000
+```
+
+- Runs in foreground, blocks until complete.
+- Exits 0 on success, non-zero on failure.
+- Supports `--resume` flag to restart from last checkpoint.
```
mprahl: Do we need a resume flag? I think the script should just be idempotent and detect if this new or resuming.

proposals/12147-mlmd-removal/database-migration-design.md
```diff
@@ -0,0 +1,463 @@
...
+- Configuration via CLI flags.
+
+**Mode 2 — API Server integration (automatic)**
+
+```yaml
+env:
```
mprahl, line 312: I don't think we need to opt-in to the migration or to opt-in to retry a failed migration.

proposals/12147-mlmd-removal/database-migration-design.md
```diff
@@ -0,0 +1,463 @@
...
+  - name: MLMD_MIGRATE
+    value: "true"
+  - name: RETRY_FAILED_MIGRATION
+    value: "false"
+  - name: MIGRATION_BATCH_SIZE
```
mprahl: This is nice but probably not meaningful to the admin. I think we can just pick a batch size and not make it configurable.

proposals/12147-mlmd-removal/database-migration-design.md
```diff
@@ -0,0 +1,463 @@
...
+Health/readiness behavior should allow the pod to be up for migration work while still signaling that the service is not ready for user traffic.
+
+#### Handling In-Flight Runs
+
+Pipelines that are already running before the upgrade cannot safely keep going once the system switches to the new metadata path. Prefer finishing the current step, then canceling and retrying after migration.
+
+Recommended upgrade flow:
+
+1. Enter maintenance mode (no new run submissions through the API).
+2. Suspend in-flight Argo Workflow objects so the **current** step can finish.
```
mprahl, line 342: I think we may need a small amount of APIs operational during this maintenance mode. I think just what the persistence agent needs and driver and launcher.

proposals/12147-mlmd-removal/database-migration-design.md
```diff
@@ -0,0 +1,463 @@
...
+3. Once workflows are suspended, cancel those pipeline runs. Cancel + later retry is required so Argo Workflows are recompiled against the new driver and launcher images.
+4. Run schema and data migration (Phases 2–4). Refuse startup only if migration cannot proceed.
+5. Exit maintenance mode after migration succeeds.
+6. Retry the canceled pipeline runs after the upgrade.
+
+Wait until migration finishes successfully before retrying canceled runs.
+
+#### Rollback
```
mprahl: As mentioned in another comment, rollbacks should be from an admin backing up the KFP database, not by us rolling back for them.

proposals/12147-mlmd-removal/database-migration-design.md
```diff
@@ -0,0 +1,463 @@
...
+  --mysql-host=mysql:3306 \
+  --batch-size=5000
+```
+
+- Runs in foreground, blocks until complete.
+- Exits 0 on success, non-zero on failure.
+- Supports `--resume` flag to restart from last checkpoint.
+- Configuration via CLI flags.
+
+**Mode 2 — API Server integration (automatic)**
```
mprahl, line 309: What are the configuration options to configure the MLMD connection from the API server?

proposals/12147-mlmd-removal/database-migration-design.md
```diff
@@ -0,0 +1,463 @@
...
+After deletion: revert to the previous API server image and restart with MLMD enabled.
+
+#### Error Handling
+
+| Error Type | Behavior |
+|---|---|
+| Individual entity transformation failure | Log warning, skip entity, continue |
+| Missing optional MLMD properties | Log warning, use default, continue |
+| Migration required but MLMD connection info missing / unreachable at startup | Fatal: refuse to start (cannot migrate) |
+| MLMD gRPC connection loss mid-migration | Fatal: set `Status=FAILED`, remain in maintenance mode, surface error |
```
mprahl: Could we add retries around this and then fail after retries are exhausted?

proposals/12147-mlmd-removal/database-migration-design.md
```diff
@@ -0,0 +1,463 @@
...
+| MySQL connection loss | Fatal: set `Status=FAILED`, surface error |
+| Disk space exhaustion | Fatal: set `Status=FAILED`, surface error |
+| Schema mismatch | Fatal: set `Status=FAILED`, surface error |
+
+Recovery: set `RETRY_FAILED_MIGRATION=true` and restart, or re-run the CLI with `--resume`. Do not exit maintenance mode or retry canceled pipeline runs until migration completes successfully.
+
+---
+
+## Risks and Mitigations
```
mprahl, line 374: I think some of the risks are not really risks. Try to keep this list concise to think we need to be worried about and not implementation details.

proposals/12147-mlmd-removal/database-migration-design.md
```diff
@@ -0,0 +1,463 @@
...
+### Integration Tests
+
+- **Full migration:** Seed a test MLMD database with executions, artifacts, and events across all TaskTypes. Run migration. Assert KFP tables contain expected records with correct state, UUID, type, and relationships.
+- **Idempotent re-run:** Run migration to completion, then run it again. Assert no duplicate rows and no errors (`INSERT IGNORE` + deterministic UUIDs).
+- **Resume after crash:** Inject a failure mid-way through Phase 3. Restart migration. Assert the final state is identical to a clean run.
+- **Checkpoint loss:** Run migration, delete `LastCheckpoint`, restart. Assert idempotency: no duplicate rows, no errors.
+- **Large metrics artifacts:** Seed MLMD with multi-metric artifacts. Assert correct splitting and `mlmd_id_map` registration.
+- **Orphaned MLMD contexts:** Include MLMD contexts without matching `run_details` rows. Assert they are skipped with a warning and remaining data is migrated correctly.
+
+### Multi-Replica Tests
```
mprahl: We can skip this test I think. I'm not sure we really support multi-replica API server.

mprahl (changes requested):

proposals/12147-mlmd-removal/database-migration-design.md
```diff
@@ -0,0 +1,463 @@
...
+Pipelines that are already running before the upgrade cannot safely keep going once the system switches to the new metadata path. Prefer finishing the current step, then canceling and retrying after migration.
+
+Recommended upgrade flow:
+
+1. Enter maintenance mode (no new run submissions through the API).
+2. Suspend in-flight Argo Workflow objects so the **current** step can finish.
+3. Once workflows are suspended, cancel those pipeline runs. Cancel + later retry is required so Argo Workflows are recompiled against the new driver and launcher images.
+4. Run schema and data migration (Phases 2–4). Refuse startup only if migration cannot proceed.
+5. Exit maintenance mode after migration succeeds.
+6. Retry the canceled pipeline runs after the upgrade.
```
droctothorpe, line 346: Will retry resume the suspended AWF? Or will it start them from scratch? What if e.g. caching is not enabled.

droctothorpe:
Does this plan include a mechanism to back out of a partial migration? We need a way to reverse the migration if significant errors surface.

proposals/12147-mlmd-removal/database-migration-design.md
VaniHaripriya: Yes, that's the intent. A non-empty diff fails the job.

droctothorpe:
I would add more detail about precisely how we're going to automate the testing of this script. Specifically, seeding a database, executing logic to validate that data is what you expect it to be when accessed through the KFP API, running the migration script, and validating that the same requests still succeed. I also recommend taking into account some scale testing. Specifically, how long migration takes, especially for databases with a large volume of runs.

proposals/12147-mlmd-removal/database-migration-design.md
VaniHaripriya: Agree that migration status table is not required, on restart we can binary search each table (tasks, artifacts, artifact_tasks) for the first missing UUID and continue, then run an idempotent parent-resolution fixup until there are zero unresolved ParentTaskUUIDs.

proposals/12147-mlmd-removal/database-migration-design.md
VaniHaripriya: Agree, we don't need checkpoint based resume when we are using transactions for each batch and can perform binary search in MLMD.

proposals/12147-mlmd-removal/database-migration-design.md
VaniHaripriya: Retry starts the canceled runs from scratch as new pipeline runs, it won't resume the suspended Argo Workflows.Suspend only lets the current step finish cleanly. After that we cancel so the workflows can be recompiled with the new (post-MLMD) driver and launcher. Resuming the old workflow objects would leave us on the old images and metadata path.
With caching enabled, migrated fingerprints may let some steps cache-hit on retry. Without caching, the full pipeline runs again,

proposals/12147-mlmd-removal/database-migration-design.md
VaniHaripriya: Agreed, updated the doc to add it as a limitation.

proposals/12147-mlmd-removal/database-migration-design.md
VaniHaripriya: Done.

proposals/12147-mlmd-removal/database-migration-design.md
VaniHaripriya: Removed the implementation file structure details.

proposals/12147-mlmd-removal/database-migration-design.md
VaniHaripriya: Got it, remove the roll back option and suggested the admin backup prior to the migration.

proposals/12147-mlmd-removal/database-migration-design.md
VaniHaripriya: Good point, enhanced the doc using this approach.

proposals/12147-mlmd-removal/database-migration-design.md
VaniHaripriya: Removed the resume flag since it should be detected automatically.

proposals/12147-mlmd-removal/database-migration-design.md
VaniHaripriya: Updated accordingly.

proposals/12147-mlmd-removal/database-migration-design.md
VaniHaripriya: Makes sense, updated to have a default batch size.

proposals/12147-mlmd-removal/database-migration-design.md
VaniHaripriya: Updated the required config options.

proposals/12147-mlmd-removal/database-migration-design.md
VaniHaripriya: Done.

VaniHaripriya:
> Does this plan include a mechanism to back out of a partial migration? We need a way to reverse the migration if significant errors surface.

Yes, before upgrade, the admin backs up the KFP database. If significant errors show up during or after migration, they need to restore that backup and roll back to the previous version.

proposals/12147-mlmd-removal/database-migration-design.md
```diff
@@ -0,0 +1,424 @@
...
+```
+
+Both entry points use the same migration logic. Only how migration is triggered and configured differs.
+
+#### Migration Phases
+
+**Phase 1 — Pre-migration validation**
+
+- Verify MLMD gRPC connectivity.
+- Verify MySQL connectivity and sufficient disk space (estimated ~2× MLMD database size).
```
mprahl, line 112: Now that Postgres support is merged
```suggestion
- Verify KFP database connectivity and sufficient disk space (estimated ~2× MLMD database size).
```

proposals/12147-mlmd-removal/database-migration-design.md
```diff
@@ -0,0 +1,424 @@
...
+  1. Create a temporary copy of the table (`tasks_migration_tmp`) with the target schema, using manually written SQL (not `AutoMigrate`) as the source of truth for the DDL.
+  2. Backfill `tasks_migration_tmp` from the existing `tasks` table (cache-fingerprint rows), mapping old columns into the new schema.
+  3. Within a single transaction: rename the existing table to a backup name (`tasks_pre_migration`), rename the temporary table to `tasks`.
+  4. After migration completes successfully, drop `tasks_pre_migration` (see [Backup and Recovery](#backup-and-recovery)).
+
+**Schema verification (CI safeguard):** To confirm the manual SQL scripts and `AutoMigrate` never diverge for the *new*-table schemas, CI generates a schema dump after running the manual migration SQL, generates a second dump after running `GORM.AutoMigrate()` against a clean database, and diffs the two. A non-empty diff fails CI. `AutoMigrate` is used here as a schema check only, not to migrate production databases. This does not apply to `tasks`, which never goes through `AutoMigrate` directly.
```
mprahl, line 126: The last sentence is a bit misleading as we will rely on AutoMigration for new installs to create the tasks table, right?

proposals/12147-mlmd-removal/database-migration-design.md
```diff
@@ -0,0 +1,424 @@
...
+```
+
+Each entity type is processed in serial batches of a fixed size (define a default batch size) in ascending MLMD ID order. Each batch runs in its own transaction using `INSERT IGNORE`. After a crash, already-committed batches stay in the database; the next run binary-searches for the first MLMD ID whose deterministic UUID is missing and continues from there.
+
+In API Server mode, Phase 3 runs while the server remains in maintenance mode. Normal API traffic resumes only after the [completion checklist](#migration-completion-checklist) passes.
+
+#### New Database Tables
+
+
+**`mlmd_id_map`** — temporary map of MLMD integer IDs to KFP UUIDs for parent-task resolution during migration:
```
mprahl, line 155: Do we actually need this table? Is there a reason not to use the MLMD ID as is or do something deterministic like hash the MLMD ID and output as UUID v4 format?

proposals/12147-mlmd-removal/database-migration-design.md
```diff
@@ -0,0 +1,424 @@
...
+
+```sql
+INSERT IGNORE INTO tasks (UUID, ...) VALUES (...);
+```
+
+Duplicate rows (same UUID) are silently skipped. A crashed-and-resumed run encounters `RowsAffected = 0` for already-migrated rows and `RowsAffected = 1` for new ones — no errors, no duplicates.
+
+**3. Resume by binary search**
+
+After a crash, for each insert stream (`tasks`, `artifacts`, `artifact_tasks`) the engine binary searches over ascending MLMD IDs and checks the KFP DB for the deterministic UUID. The first missing ID is the resume point, migration continues with per batch transactions from there.
```
mprahl, line 196:
```suggestion
After a crash or running after a partially migrated DB through the utility script, for each insert stream (`tasks`, `artifacts`, `artifact_tasks`) the engine binary searches over ascending MLMD IDs and checks the KFP DB for the deterministic UUID. The first missing ID is the resume point, migration continues with per batch transactions from there.
```

proposals/12147-mlmd-removal/database-migration-design.md
```diff
@@ -0,0 +1,424 @@
...
+
+### User Stories
+
+**Story 1: Admin upgrading an existing KFP deployment**
+
+> As a KFP platform admin, I want to upgrade from a MLMD-backed version of KFP to a MLMD-free version without losing any historical pipeline run data or artifact lineage, so that my team can continue to audit and debug past runs after the upgrade.
+
+**Story 2: Admin preferring manual control**
+
+> As a KFP admin who manages migration windows carefully, I want to run the migration as a standalone job on my own schedule, before or after the API server upgrade, so I can control when the extra database load occurs and verify results before proceeding.
```
mprahl, line 74: Could you make it more explicit that it's a primary user story for an admin to run the utility script ahead of time so that the API server migration is just an idempotent "true up" since the script last ran to minimize downtime?

proposals/12147-mlmd-removal/database-migration-design.md
```diff
@@ -0,0 +1,424 @@
...
+   - Cleanup done: `mlmd_id_map` and `tasks_pre_migration` are absent.
+
+#### MLMD connection configuration
+
+Migration needs MLMD gRPC connection settings on the migrator entry points:
+
+| Setting | Purpose |
+|---|---|
+| MLMD gRPC address (host + port) | Connect to `metadata-grpc-service` (or equivalent) to read executions, artifacts, and events |
+| Optional TLS / credentials | Only if the deployment’s MLMD endpoint requires them |
```
mprahl, line 219: Rather than credentials, maybe allow accepting arbitrary headers so it'd allow an admin to specify an authorization header if needed?

proposals/12147-mlmd-removal/database-migration-design.md
```diff
@@ -0,0 +1,424 @@
...
+- Locks are session-scoped and auto-release on connection loss or crash — no manual cleanup required.
+
+#### Execution Modes
+
+**Mode 1 — Standalone CLI (manual)**
+
+```bash
+mlmd-migrate \
+  --mlmd-address=metadata-grpc-service:8080 \
+  --mysql-host=mysql:3306
```
mprahl, line 301: Maybe something like this to not be limited to just MySQL?
```suggestion
  --db-uri=mysql:3306
```

mprahl (changes requested):
Just a few minor comments but this looks good!

proposals/12147-mlmd-removal/database-migration-design.md
VaniHaripriya: Good catch, updated to clarify that new installs still create tasks via AutoMigrate from the GORM model.

proposals/12147-mlmd-removal/database-migration-design.md
VaniHaripriya: Agreed, Deterministic UUID v5 is enough for resume, parent fixup, and artifact/event links, so a side lookup table isn’t required. Updated the KEP to drop mlmd_id_map and rely on UUID v5.

mprahl (approved):
/lgtm
/approve

PR #12927, opened 2026-02-27
Switch to slim base images for components and frontend

jsonmp-k8:
## Summary

Use slim variants of base images in component definitions and the frontend Dockerfile to reduce container image sizes. This targets components that do not require system tools like git or gcc at runtime.

- **frontend**: `node:22.19` -> `node:22.19-slim`
- **Google Cloud starry_net** (8 components): `python:3.9` -> `python:3.9-slim`
- **Google Cloud forecasting**: `python:3.8` -> `python:3.8-slim`
- **Snowflake** (2 files): `python:3.11` -> `python:3.11-slim`
- **PyTorch launcher sample**: `python:3.11` -> `python:3.11-slim`

The SDK default base image (`python:3.11`) is intentionally unchanged, as lightweight Python components rely on system tools (e.g. `git`) that are not available in slim images.

## Test plan

- [ ] Verify all pre-commit checks pass
- [ ] Verify E2E tests pass (no impact expected since only explicit component images are changed, not the SDK default)
- [ ] Verify frontend builds successfully with `node:22.19-slim`

### DIFF
--- a/frontend/Dockerfile
+++ b/frontend/Dockerfile
@@ -1,5 +1,5 @@
 ARG BASE_IMAGE=alpine
-FROM node:22.19-slim as build
+FROM node:22.19-slim AS build
 
 ARG COMMIT_HASH
 ENV COMMIT_HASH=${COMMIT_HASH}


jiayuzhao05 (commented):
Most of these components appear to be runtime-only and do not rely on
system-level tools or native compilation, so switching
to python:*-slim / node:*-slim is safe.

LGTM.

components/google-cloud/google_cloud_pipeline_components/_implementation/starry_net/get_training_artifacts/component.py
```diff
@@ -19,7 +19,7 @@
 
 
 @dsl.component(
-    base_image='python:3.9', packages_to_install=['tensorflow==2.16.1']
+    base_image='python:3.9-slim', packages_to_install=['tensorflow==2.16.1']
```
jeffspahr, line 22: This component installs `tensorflow==2.16.1` at runtime via `packages_to_install`. While TF 2.16.1 ships manylinux wheels (so `pip install` should succeed without gcc), some TF operations depend on system libraries like `libgomp1` that aren't in slim images.

Worth verifying with:

If it fails, you may need `apt-get install -y libgomp1` or keep the full image for this component.

frontend/Dockerfile
```diff
@@ -1,5 +1,5 @@
 ARG BASE_IMAGE=alpine
-FROM node:22.19 as build
+FROM node:22.19-slim as build
```
jeffspahr: This change only affects the build stage — the runtime stage already uses `node:22.19-alpine` (via `BASE_IMAGE` ARG). Since the build only runs `npm ci` + `npm run build` and there are no git-based or native addon dependencies in `package.json`, this should be safe.

Minor note: the `as build` syntax is deprecated in favor of `AS build` (case-sensitive) in newer BuildKit versions. Pre-existing issue, not introduced here.

jeffspahr (commented):
## Code Review: Switch to slim base images for components and frontend

### Overview

This PR changes 13 files to switch from full base images to their `-slim` variants across component definitions and the frontend Dockerfile. The goal is to reduce container image sizes without affecting runtime behavior.

### Risk Assessment

**Low-to-medium risk.** Most changes are safe, but two components install packages with native extensions that deserve verification.

### Analysis by Component

**Safe (pure Python / prebuilt wheels):**
- starry_net components without `packages_to_install` (6 files) — no native deps, trivially safe
- forecasting `prepare_data_for_train` — no `packages_to_install`, safe
- pytorch-launcher — `kubernetes`, `kubeflow-training`, `retrying` are pure Python, safe

**Needs verification:**
- **`get_training_artifacts` and `set_test_set`** — install `tensorflow==2.16.1` on `python:3.9-slim`. TensorFlow 2.16.1 ships manylinux wheels that bundle native libs, so `pip install` should succeed. However, some TF operations depend on system libraries (e.g. `libgomp1` for parallelism) that may not be present in slim. This should be smoke-tested.
- **`upload_decomposition_plots`** — installs `google-cloud-aiplatform[tensorboard]==1.87.0` + `protobuf==3.20.*`. grpcio and protobuf have prebuilt wheels for common platforms, but protobuf 3.20.x is old enough that it may need compilation on some architectures. Likely fine on linux/amd64.
- **Snowflake components** — `snowflake-connector-python` ships manylinux wheels, should work on slim. (Pre-existing note: the `packages_to_install` uses a colon instead of `==` — `snowflake-connector-python:3.12.3` — which would fail with pip. But that's a pre-existing bug, not introduced here.)

**Frontend (`node:22.19-slim`):**
- No git-based npm dependencies found in `package.json`
- No known native addon packages (node-sass, sharp, etc.) detected
- The slim image lacks git, python3, make, and gcc — but the build only runs `npm ci` + `npm run build`, which should work with prebuilt binaries
- Note: this only affects the **build stage**; the runtime stage already uses `node:22.19-alpine`

### Suggestions

1. **Smoke-test the TensorFlow components** — run a quick `docker build` or `pip install tensorflow==2.16.1` on `python:3.9-slim` to confirm no missing shared libraries at import time (`import tensorflow` should not segfault or error on missing `libgomp`).

2. The PR summary is well-written and the scope exclusion (SDK default intentionally unchanged) is a good call.

### Verdict

Looks good overall. The changes follow a consistent pattern, the scope is well-bounded, and the PR summary clearly documents what's in and out. The main action item is verifying TensorFlow imports work on the slim base.

components/google-cloud/google_cloud_pipeline_components/_implementation/starry_net/get_training_artifacts/component.py
jsonmp-k8: Verified — `tensorflow==2.16.1` installs and imports successfully on `python:3.9-slim`:

```
$ docker run --rm python:3.9-slim sh -c "pip install tensorflow==2.16.1 && python -c 'import tensorflow as tf; print(tf.__version__)'"
2.16.1
```

Also confirmed `tf.io.gfile` (which both `get_training_artifacts` and `set_test_set` use) works correctly. No missing shared libraries — TF 2.16.1 manylinux wheels bundle all native deps including OpenBLAS/libgomp.

frontend/Dockerfile
jsonmp-k8: Good call — fixed the `as` → `AS` deprecation while touching this file.

jsonmp-k8:
Thanks for the thorough review @jeffspahr!

Addressing the remaining points:

**`upload_decomposition_plots` (`protobuf==3.20.*`)**: Verified — `protobuf` 3.20.x ships prebuilt manylinux wheels for linux/amd64 (and aarch64), so `pip install` succeeds on `python:3.9-slim` without needing a compiler. `grpcio` (pulled in by `google-cloud-aiplatform`) also ships manylinux wheels. No issues expected here.

**Snowflake colon bug**: Good catch on the pre-existing `snowflake-connector-python:3.12.3` (colon instead of `==`). That's outside the scope of this PR, but worth a separate fix — I'll file a follow-up PR for it.

The TF verification and `AS build` fix were addressed in the inline replies.

PR #12935, opened 2026-02-27
test(v2/driver): add test coverage for cache, validation, and helpers

jsonmp-k8:
## Summary

- Increase test coverage for the `backend/src/v2/driver` package from **29.8% to 36.7%** by adding **58 new test cases** targeting previously untested functions
- Add tests for cache subsystem (`getFingerPrint`, `getFingerPrintsAndID`, `createCache`) with a mock `cacheutils.Client`
- Add validation tests for `validateContainer`, `validateDAG`, `validateRootDAG`, `validateNonRoot`
- Add tests for helper functions (`inputParamConstant`, `inputParamComponent`, `inputParamTaskOutput`, `getItems`)
- Add tests for `provisionOutputs` and `InferIndexedTaskName`

## Test plan

- [x] All 58 new test cases pass with `go test -v -cover ./backend/src/v2/driver/...`
- [x] No existing tests broken
- [x] Coverage verified at 36.7% (up from 29.8%)

### DIFF
--- a/backend/src/v2/driver/cache_test.go
+++ b/backend/src/v2/driver/cache_test.go
@@ -428,6 +428,20 @@ func Test_createCache(t *testing.T) {
 			wantErr: false,
 		},
 		{
+			// nil execution is handled gracefully by Execution.GetID() returning 0,
+			// which triggers the "execution ID is 0" error path.
+			name:      "nil execution returns error",
+			execution: nil,
+			opts: &Options{
+				PipelineName: "my-pipeline",
+				Namespace:    "default",
+				RunID:        "run-001",
+			},
+			fingerPrint: "fingerprint-nil",
+			mockClient:  &mockCacheClient{},
+			wantErr:     true,
+		},
+		{
 			name:      "error when execution ID is 0",
 			execution: metadata.NewExecution(&pb.Execution{Id: proto.Int64(0)}),
 			opts: &Options{
--- a/backend/src/v2/driver/driver_test.go
+++ b/backend/src/v2/driver/driver_test.go
@@ -1783,6 +1783,7 @@ func Test_provisionOutputs(t *testing.T) {
 		wantArtifacts    []string
 		wantParameters   []string
 		wantLogsArtifact bool
+		wantOutputFile   string
 	}{
 		{
 			name:         "provisions output artifacts with URIs",
@@ -1803,6 +1804,7 @@ func Test_provisionOutputs(t *testing.T) {
 			publishOutput:    "false",
 			wantArtifacts:    []string{"model"},
 			wantLogsArtifact: false,
+			wantOutputFile:   "/gcs/my-bucket/pipeline-root/my-task/salt-123/output_metadata.json",
 		},
 		{
 			name:         "provisions output parameters",
@@ -1871,6 +1873,9 @@ func Test_provisionOutputs(t *testing.T) {
 			assert.NotNil(t, outputs.Artifacts)
 			assert.NotNil(t, outputs.Parameters)
 			assert.NotEmpty(t, outputs.OutputFile)
+			if test.wantOutputFile != "" {
+				assert.Equal(t, test.wantOutputFile, outputs.OutputFile)
+			}
 
 			for _, artifactName := range test.wantArtifacts {
 				artifactList, ok := outputs.Artifacts[artifactName]
--- a/backend/src/v2/driver/util_test.go
+++ b/backend/src/v2/driver/util_test.go
@@ -242,9 +242,9 @@ func Test_getItems(t *testing.T) {
 		t.Run(test.name, func(t *testing.T) {
 			items, err := getItems(test.value)
 			if test.wantErr {
-				assert.NotNil(t, err)
+				assert.Error(t, err)
 			} else {
-				assert.Nil(t, err)
+				assert.NoError(t, err)
 				assert.Equal(t, test.wantCount, len(items))
 			}
 		})


jsonmp-k8 (commented):
.

backend/src/v2/driver/util_test.go
```diff
@@ -101,6 +101,156 @@ func Test_extractInputParameterFromChannel(t *testing.T) {
...
+			wantErr: true,
+		},
+	}
+
+	for _, test := range tests {
+		t.Run(test.name, func(t *testing.T) {
+			items, err := getItems(test.value)
+			if test.wantErr {
+				assert.NotNil(t, err)
+			} else {
```
jeffspahr, line 246: Nit: This uses `assert.NotNil(t, err)` / `assert.Nil(t, err)` while every other test in this PR uses `assert.Error(t, err)` / `assert.NoError(t, err)`. The latter provides better failure messages (prints the actual error vs. just `expected not nil`). Consider switching for consistency:
```suggestion
				assert.Error(t, err)
```

backend/src/v2/driver/util_test.go
```diff
@@ -101,6 +101,156 @@ func Test_extractInputParameterFromChannel(t *testing.T) {
...
+				assert.Nil(t, err)
+				assert.Equal(t, test.wantCount, len(items))
```
jeffspahr, line 248: Same here:
```suggestion
				assert.NoError(t, err)
```

backend/src/v2/driver/driver_test.go
```diff
@@ -1675,3 +1675,232 @@ func Test_initPodSpecPatch_mlPipelineServerConfig(t *testing.T) {
...
+				test.taskName,
+				test.outputsSpec,
+				test.outputURISalt,
+				test.publishOutput,
+			)
+			assert.NotNil(t, outputs)
+			assert.NotNil(t, outputs.Artifacts)
+			assert.NotNil(t, outputs.Parameters)
+			assert.NotEmpty(t, outputs.OutputFile)
+
```
jeffspahr, line 1879: Suggestion: `provisionOutputs` has branching logic for local vs. remote pipeline roots that affects the output file path. Consider asserting the expected path value (not just non-empty) for at least one test case to guard against regressions in that logic. For example:
```go
assert.Equal(t, "/tmp/kfp/outputs/executor_output.json", outputs.OutputFile)
```
or the GCS-derived path, depending on which branch is expected.

backend/src/v2/driver/cache_test.go
```diff
@@ -0,0 +1,477 @@
...
+					assert.Equal(t, "123", task.MlmdExecutionID)
+					assert.Equal(t, "fingerprint-abc", task.Fingerprint)
+					return nil
+				},
+			},
+			wantErr: false,
+		},
+		{
+			name:      "error when execution ID is 0",
+			execution: metadata.NewExecution(&pb.Execution{Id: proto.Int64(0)}),
```
jeffspahr, line 446: Optional: Consider adding a test case for nil `execution` input. `createCache` calls `execution.GetID()` which would panic on nil. If the function is expected to handle this gracefully, a test would document that contract. If it's intentionally left as a caller responsibility, a brief comment in the test would clarify.

backend/src/v2/driver/util_test.go
jeffspahr, line 246: Nit: This uses `assert.NotNil(t, err)` / `assert.Nil(t, err)` while every other test in this PR uses `assert.Error(t, err)` / `assert.NoError(t, err)`. The latter provides better failure messages (prints the actual error vs. just `expected not nil`). Consider switching for consistency:
```suggestion
				assert.Error(t, err)
```

backend/src/v2/driver/util_test.go
jeffspahr, line 248: Same here:
```suggestion
				assert.NoError(t, err)
```

backend/src/v2/driver/driver_test.go
jeffspahr, line 1879: Suggestion: `provisionOutputs` has branching logic for local vs. remote pipeline roots that affects the output file path. Consider asserting the expected path value (not just non-empty) for at least one test case to guard against regressions in that logic. For example:
```go
assert.Equal(t, "/tmp/kfp/outputs/executor_output.json", outputs.OutputFile)
```
or the GCS-derived path, depending on which branch is expected.

backend/src/v2/driver/cache_test.go
jeffspahr, line 446: Optional: Consider adding a test case for nil `execution` input. `createCache` calls `execution.GetID()` which would panic on nil. If the function is expected to handle this gracefully, a test would document that contract. If it's intentionally left as a caller responsibility, a brief comment in the test would clarify.

backend/src/v2/driver/util_test.go
jsonmp-k8: Good catch — switched both to `assert.Error`/`assert.NoError` for consistency. Fixed in d17b0c25f.

backend/src/v2/driver/util_test.go
jsonmp-k8: Done.

backend/src/v2/driver/driver_test.go
jsonmp-k8: Great suggestion. Added a `wantOutputFile` field to the test struct and asserted the exact GCS-derived path (`/gcs/my-bucket/pipeline-root/my-task/salt-123/output_metadata.json`) on the first test case. This covers the `LocalPathForURI` → `filepath.Join` branch. Fixed in d17b0c25f.

backend/src/v2/driver/cache_test.go
jsonmp-k8: Added a nil execution test case. `Execution.GetID()` already handles nil gracefully by returning 0, which triggers the existing `id == 0` error path — so no panic. Added a comment in the test documenting this contract. Fixed in d17b0c25f.

backend/src/v2/driver/util_test.go
jsonmp-k8: Addressed — switched to `assert.Error`/`assert.NoError` in d17b0c25f.

droctothorpe:
/approve
/lgtm

🎉

PR #12936, opened 2026-02-28
test(backend): add test coverage for apiserver/common package

jsonmp-k8:
## Summary
- Add comprehensive unit tests for the `backend/src/apiserver/common` package, increasing test coverage from **33.3% to 95.7%**
- Test all config core helpers (`GetStringConfigWithDefault`, `GetBoolConfigWithDefault`, `GetFloat64ConfigWithDefault`, `GetIntConfigWithDefault`, `GetMapConfig`, `GetBoolFromStringWithDefault`, `GetStringConfig`, `GetDurationConfig`)
- Test all 17 config wrapper functions for both default values and custom env var overrides
- Test utility functions: `CreateArtifactPath`, `ParseLogLevel`, `FileExists`, `CustomMarshaler`, `PatchPipelineDefaultParameter`

## Test plan
- [x] All new tests pass locally with `CGO_ENABLED=0 go test -v -cover ./backend/src/apiserver/common/...`
- [x] No existing tests broken
- [x] `gofmt -w` applied to all modified files
- [x] Coverage verified at 95.7% (up from 33.3%)

### DIFF
--- a/backend/src/v2/driver/cache_test.go
+++ b/backend/src/v2/driver/cache_test.go
@@ -1,0 +1,491 @@
+// Copyright 2025 The Kubeflow Authors
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//	https://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package driver
+
+import (
+	"context"
+	"fmt"
+	"testing"
+
+	"github.com/kubeflow/pipelines/api/v2alpha1/go/cachekey"
+	"github.com/kubeflow/pipelines/api/v2alpha1/go/pipelinespec"
+	api "github.com/kubeflow/pipelines/backend/api/v1beta1/go_client"
+	"github.com/kubeflow/pipelines/backend/src/v2/cacheutils"
+	"github.com/kubeflow/pipelines/backend/src/v2/metadata"
+	pb "github.com/kubeflow/pipelines/third_party/ml-metadata/go/ml_metadata"
+	"github.com/stretchr/testify/assert"
+	"google.golang.org/protobuf/proto"
+	"google.golang.org/protobuf/types/known/structpb"
+)
+
+// mockCacheClient implements cacheutils.Client for testing.
+type mockCacheClient struct {
+	getExecutionCacheFunc    func(fingerPrint, pipelineName, namespace string) (string, error)
+	createExecutionCacheFunc func(ctx context.Context, task *api.Task) error
+	generateCacheKeyFunc     func(inputs *pipelinespec.ExecutorInput_Inputs, outputs *pipelinespec.ExecutorInput_Outputs, outputParametersTypeMap map[string]string, cmdArgs []string, image string, pvcNames []string) (*cachekey.CacheKey, error)
+	generateFingerPrintFunc  func(cacheKey *cachekey.CacheKey) (string, error)
+}
+
+var _ cacheutils.Client = &mockCacheClient{}
+
+func (m *mockCacheClient) GetExecutionCache(fingerPrint, pipelineName, namespace string) (string, error) {
+	if m.getExecutionCacheFunc != nil {
+		return m.getExecutionCacheFunc(fingerPrint, pipelineName, namespace)
+	}
+	return "", nil
+}
+
+func (m *mockCacheClient) CreateExecutionCache(ctx context.Context, task *api.Task) error {
+	if m.createExecutionCacheFunc != nil {
+		return m.createExecutionCacheFunc(ctx, task)
+	}
+	return nil
+}
+
+func (m *mockCacheClient) GenerateCacheKey(
+	inputs *pipelinespec.ExecutorInput_Inputs,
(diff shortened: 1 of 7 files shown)


backend/src/apiserver/common/config_test.go
```diff
@@ -16,11 +16,535 @@ package common
...
+		expected float64
+	}{
+		{
+			name:     "returns default when not set",
+			setEnv:   false,
+			expected: 3.14,
+		},
+		{
+			name:     "returns float value when set",
+			setEnv:   true,
```
jeffspahr, line 127: **Suggestion: add an invalid-input test case.** What happens when the env var is set to a non-numeric value like `"not-a-number"`? Same applies to `TestGetIntConfigWithDefault` below.

If viper silently falls back to the default, that's worth documenting here. If it panics or returns zero, that's also worth knowing.

```go
{
    name:     "returns default when env var is not a valid float",
    setEnv:   true,
    envValue: "not-a-number",
    expected: 3.14,
},
```

backend/src/apiserver/common/config_test.go
```diff
@@ -16,11 +16,535 @@ package common
...
+			viper.Reset()
+
+			if testCase.setEnv {
+				os.Setenv("TEST_STRING_CONFIG", testCase.envValue)
+				defer os.Unsetenv("TEST_STRING_CONFIG")
+				viper.AutomaticEnv()
+			} else {
+				os.Unsetenv("TEST_STRING_CONFIG")
+			}
+
```
jeffspahr, line 55: **Nit: consider using `t.Setenv()` (Go 1.17+)** instead of `os.Setenv` / `defer os.Unsetenv` throughout this file. `t.Setenv()` automatically restores the original value on test cleanup, and will call `t.Fatalf` if `t.Parallel()` is ever enabled on the test — preventing subtle global-state flakiness.

```go
// Before:
os.Setenv("TEST_STRING_CONFIG", testCase.envValue)
defer os.Unsetenv("TEST_STRING_CONFIG")

// After:
t.Setenv("TEST_STRING_CONFIG", testCase.envValue)
```

This applies to all `os.Setenv`/`os.Unsetenv` pairs in both test files.

backend/src/apiserver/common/utils_test.go
```diff
@@ -15,12 +15,209 @@
...
+	})
+
+	t.Run("non-existent path returns false", func(t *testing.T) {
+		assert.False(t, FileExists("/tmp/non-existent-file-that-does-not-exist-12345"))
+	})
+
+	t.Run("existing directory returns true", func(t *testing.T) {
+		tempDir, err := os.MkdirTemp("", "test-dir-exists-*")
+		assert.NoError(t, err)
+		defer os.RemoveAll(tempDir)
```
jeffspahr, line 154: **Suggestion: document that `FileExists` returns `true` for directories.** The function name implies file-only, but this test correctly shows it also matches directories. A brief comment would help future readers understand this is intentional behavior rather than a bug:

```go
t.Run("existing directory returns true (FileExists checks path existence, not file type)", func(t *testing.T) {
```

Alternatively, if this is unintended, it could be worth filing a follow-up to rename to `PathExists`.

backend/src/apiserver/common/utils_test.go
```diff
@@ -15,12 +15,209 @@
...
+			input:    "gs://{{kfp-default-bucket}}/{{kfp-project-id}}/output",
+			bucket:   "my-bucket",
+			project:  "my-project",
+			expected: "gs://my-bucket/my-project/output",
+		},
+		{
+			name:     "text with no placeholders is unchanged",
+			input:    "gs://some-bucket/some-project/output",
+			bucket:   "my-bucket",
+			project:  "my-project",
```
jeffspahr, line 188: **Suggestion: add an error-path test case.** The function signature is `PatchPipelineDefaultParameter(string) (string, error)`, but all test cases set both env vars successfully, so `err` is never non-nil.

Consider adding a case where the env vars are unset to exercise the error return (or confirm the function handles missing env vars gracefully):

```go
{
    name:  "returns error when bucket env var is not set",
    input: "gs://{{kfp-default-bucket}}/data",
},
```

backend/src/apiserver/common/config_test.go
jsonmp-k8: Done — added test cases for both `TestGetFloat64ConfigWithDefault` and `TestGetIntConfigWithDefault` with `"not-a-number"` input. Confirmed that viper silently returns `0` (the zero value) for unparseable values rather than falling back to the default.

backend/src/apiserver/common/config_test.go
jsonmp-k8: Done — removed all redundant `os.Unsetenv` calls from else branches across both test files (7 total). Since `t.Setenv` auto-restores on subtest cleanup and these tests don't use `t.Parallel()`, the explicit unsets were unnecessary. Also removed the now-unused `os` import from `config_test.go`.

backend/src/apiserver/common/utils_test.go
jsonmp-k8: Done — updated the test name to `"existing directory returns true (FileExists checks path existence, not file type)"` to clarify this is intentional behavior.

backend/src/apiserver/common/utils_test.go
jsonmp-k8: Done — added `TestPatchPipelineDefaultParameterMissingEnv` using the subprocess test pattern. Worth noting: `PatchPipelineDefaultParameter` never actually returns a non-nil error — when required env vars (`BUCKET_NAME`, `PROJECT_ID`) are missing, `GetStringConfig` calls `glog.Fatalf` which exits the process. The new test verifies this fatal-exit behavior by running the function in a subprocess with the env vars stripped and asserting a non-zero exit code.

PR #12937, opened 2026-02-28
test(backend): add test coverage for apiserver/server package

jsonmp-k8:
## Summary

- Add 45 new test functions covering 53 previously uncovered functions in the `backend/src/apiserver/server` package
- Create new test files for `task_server.go` and `run_log_server.go` which had zero test coverage
- Add tests for archive, unarchive, delete, terminate, and retry operations across v1beta1 and v2beta1 APIs
- Add table-driven tests for storage state converters, pipeline version batch converters, and runtime config converter
- Add scheduled workflow reporting tests for invalid manifest and missing field cases

## Test plan

- [ ] Verify all new tests pass in CI via `go test -v -cover ./backend/src/apiserver/server/...`
- [ ] Verify no existing tests are broken
- [ ] Confirm coverage increase from 65.2% baseline

### DIFF
--- a/backend/src/apiserver/common/config_test.go
+++ b/backend/src/apiserver/common/config_test.go
@@ -14,7 +14,6 @@
 package common
 
 import (
-	"os"
 	"testing"
 	"time"
 
@@ -52,8 +51,6 @@ func TestGetStringConfigWithDefault(t *testing.T) {
 			if testCase.setEnv {
 				t.Setenv("TEST_STRING_CONFIG", testCase.envValue)
 				viper.AutomaticEnv()
-			} else {
-				os.Unsetenv("TEST_STRING_CONFIG")
 			}
 
 			result := GetStringConfigWithDefault("TEST_STRING_CONFIG", "my-default")
@@ -105,8 +102,6 @@ func TestGetBoolConfigWithDefault(t *testing.T) {
 			if testCase.setEnv {
 				t.Setenv("TEST_BOOL_CONFIG", testCase.envValue)
 				viper.AutomaticEnv()
-			} else {
-				os.Unsetenv("TEST_BOOL_CONFIG")
 			}
 
 			result := GetBoolConfigWithDefault("TEST_BOOL_CONFIG", testCase.defaultValue)
@@ -133,6 +128,12 @@ func TestGetFloat64ConfigWithDefault(t *testing.T) {
 			envValue: "2.718",
 			expected: 2.718,
 		},
+		{
+			name:     "returns zero when env var is not a valid float",
+			setEnv:   true,
+			envValue: "not-a-number",
+			expected: 0,
+		},
 	}
 
 	for _, testCase := range tests {
@@ -142,8 +143,6 @@ func TestGetFloat64ConfigWithDefault(t *testing.T) {
 			if testCase.setEnv {
 				t.Setenv("TEST_FLOAT_CONFIG", testCase.envValue)
 				viper.AutomaticEnv()
-			} else {
-				os.Unsetenv("TEST_FLOAT_CONFIG")
 			}
 
 			result := GetFloat64ConfigWithDefault("TEST_FLOAT_CONFIG", 3.14)
@@ -170,6 +169,12 @@ func TestGetIntConfigWithDefault(t *testing.T) {
 			envValue: "100",
 			expected: 100,
 		},
+		{
+			name:     "returns zero when env var is not a valid int",
+			setEnv:   true,
+			envValue: "not-a-number",
+			expected: 0,
+		},
 	}
 
 	for _, testCase := range tests {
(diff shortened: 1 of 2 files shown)


backend/src/apiserver/common/config_test.go
```diff
@@ -16,11 +16,535 @@ package common
...
+	}{
+		{
+			name:     "returns default when config not set",
+			setEnv:   false,
+			expected: "my-default",
+		},
+		{
+			name:     "returns config value when set via env var",
+			setEnv:   true,
+			envValue: "custom-value",
```
jeffspahr, line 42: Nit: Consider using `t.Setenv("TEST_STRING_CONFIG", testCase.envValue)` instead of `os.Setenv` + `defer os.Unsetenv`. It auto-restores the value on test cleanup and will panic if called from a parallel test (acting as a safety guard against the global viper singleton race). Available since Go 1.17.

Same pattern applies to all the other `os.Setenv`/`defer os.Unsetenv` pairs in this file and `utils_test.go`.

backend/src/apiserver/server/run_server_test.go
```diff
@@ -1512,8 +1512,143 @@ func TestRetryRun(t *testing.T) {
 			PipelineRoot: "model-pipeline-root",
 		},
 	}
-	run, err := server.CreateRun(nil, &apiv2beta1.CreateRunRequest{Run: run})
+	run, err := server.CreateRun(context.Background(), &apiv2beta1.CreateRunRequest{Run: run})
```
jeffspahr, line 1515: Good catch fixing this from `nil` context to `context.Background()` and updating the assertion. The original test was wrong on two counts:
1. `nil` context could panic in middleware
2. A freshly created run is not in Failed/Error state, so `RetryRun` should return an error

Worth calling this out in the PR description since it's the one existing test behavioral change.

backend/src/apiserver/server/task_server_test.go
```diff
@@ -0,0 +1,241 @@
...
+	assert.NotNil(t, err)
+	assert.Contains(t, err.Error(), "Id should not be set")
+}
+
+func TestCreateTaskV1_MissingPipelineName(t *testing.T) {
+	clients, manager, _ := initWithExperiment(t)
+	defer clients.Close()
+	server := createTaskServer(manager)
+	_, err := server.CreateTaskV1(context.Background(), &api.CreateTaskRequest{
+		Task: &api.Task{
```
jeffspahr: Nit: The validation tests (`TestCreateTaskV1_MissingPipelineName`, `_MissingRunId`, `_MissingMlmdExecutionID`, `_MissingFingerprint`, `_MissingCreatedAt`) all follow the same pattern — create a task with one field omitted and assert a specific error. These would be a natural fit for a single table-driven test, which would cut ~60 lines of near-identical boilerplate and match the style used elsewhere in this PR.

backend/src/apiserver/server/api_converter_test.go
```diff
@@ -4723,3 +4723,182 @@ func Test_toApiRun(t *testing.T) {
...
+	runtimeConfig := &model.RuntimeConfig{
+		Parameters:   model.LargeText(serializedParameters),
+		PipelineRoot: "gs://my-bucket/pipeline-root",
+	}
+	result := toPipelineSpecRuntimeConfig(runtimeConfig)
+	assert.NotNil(t, result)
+	assert.Equal(t, "gs://my-bucket/pipeline-root", result.GcsOutputDirectory)
+	assert.NotNil(t, result.ParameterValues)
+	assert.Equal(t, 2, len(result.ParameterValues))
+}
```
jeffspahr, line 4904: Consider adding a test case for `toPipelineSpecRuntimeConfig` where `model.RuntimeConfig.Parameters` contains invalid JSON (e.g., `"not valid json"`). This would document whether the converter silently returns empty params or propagates an error — useful for anyone debugging runtime config issues later.

backend/src/apiserver/server/run_log_server_test.go
```diff
@@ -0,0 +1,83 @@
...
+}
+
+func TestRunLogServer_writeErrorToResponse(t *testing.T) {
+	clients, manager, _ := initWithExperiment(t)
+	defer clients.Close()
+	server := NewRunLogServer(manager)
+
+	recorder := httptest.NewRecorder()
+	server.writeErrorToResponse(recorder, http.StatusBadRequest, assert.AnError)
+
```
jeffspahr, line 44: Nit: The URL path `/apis/v1beta1/runs/logs` is misleading since `mux.SetURLVars` overrides variable extraction anyway. Consider using a comment or a more obviously-placeholder URL like `/test` to make it clear the URL itself doesn't matter here — only the vars map does.

droctothorpe:
/approve
/lgtm

This is fantastic, @jsonmp-k8. Thank you so much for the contribution!

PR #12945, opened 2026-03-02
chore(frontend): migrate react-flow-renderer to @xyflow/react

### LINKED ISSUE #12893: chore(frontend): Migrate react-flow-renderer to @xyflow/react
## Chore description

Replace deprecated `react-flow-renderer` v9 with `@xyflow/react`. Update DAG visualization components, node/edge type definitions, event handlers, and related stories/tests. Remove peer gate allowlist entry.

Depends on https://github.com/kubeflow/pipelines/pull/12881 and here's the roadmap/checklist: https://github.com/kubeflow/pipelines/blob/master/frontend/docs/react-18-19-upgrade-checklist.md

**Acceptance Criteria**:
- [ ] Pipeline graph renders correctly with drag, zoom, and pan interactions
- [ ] `npm run check:react-peers` passes without `react-flow-renderer` in allowlist
- [ ] `npm run test:ci` passes

### Labels

 /area frontend 

---

Love this idea? Give it a 👍.

PR3MM:
**Summary**

Replace deprecated `react-flow-renderer` v9 with `@xyflow/react` v12 for the DAG pipeline visualization. This is step 5 of the [React 18/19 upgrade checklist](https://github.com/kubeflow/pipelines/blob/master/frontend/docs/react-18-19-upgrade-checklist.md). Depends on #12881.

Closes #12893

### Why

`react-flow-renderer` is deprecated and unmaintained. The authors restructured it as `@xyflow/react`, which supports React 17/18/19 and is actively maintained. Migrating now unblocks the React 18 upgrade path and removes the last entry from the peer gate allowlist.

### Migration strategy

To minimize the diff surface, we preserve the existing mixed `elements` array pattern throughout the application. A union type `PipelineFlowElement = Node<FlowElementDataBase> | Edge` replaces the old `FlowElement<FlowElementDataBase>`. The split into separate `nodes[]` and `edges[]` (required by the new API) happens only at the `DagCanvas` rendering boundary.

### Key API changes applied

| Old (`react-flow-renderer` v9) | New (`@xyflow/react` v12) |
|---|---|
| `import ReactFlow` (default export) | `import { ReactFlow }` (named export) |
| `<ReactFlow elements={...}>` | `<ReactFlow nodes={...} edges={...}>` |
| `onLoad` callback | `onInit` callback |
| `onElementClick` | `onNodeClick` + `onEdgeClick` |
| `arrowHeadType: ArrowHeadType.ArrowClosed` | `markerEnd: { type: MarkerType.ArrowClosed }` |
| Implicit drag (mutable elements) | Controlled drag via `onNodesChange` + `applyNodeChanges` |
| CSS bundled automatically | Explicit `import '@xyflow/react/dist/style.css'` |

### Files changed (19)

| Category | Files | What changed |
|---|---|---|
| Dependency | `package.json`, `package-lock.json` | Swap `react-flow-renderer` → `@xyflow/react` |
| Core types | `StaticFlow.ts`, `DynamicFlow.ts` | New `PipelineFlowElement` type, `isNode` guard, `MarkerType` edges |
| Graph renderer | `DagCanvas.tsx` | Split elements, controlled drag, new callbacks |
| Custom nodes | `ArtifactNode.tsx`, `ExecutionNode.tsx`, `SubDagNode.tsx` | Import path swap only |
| Pages | `RunDetailsV2.tsx`, `PipelineDetailsV2.tsx` | Type updates, safe `getNodeName` guard |
| Tabs | `RuntimeNodeDetailsV2.tsx`, `StaticNodeDetailsV2.tsx` | Type updates |
| Stories | 4 story files | New API (`nodes`/`edges`, `onInit`) |
| Test infra | `vitest.setup.ts`, `vitest.config.mts` | jsdom polyfills for `DOMMatrixReadOnly` and d3-drag |
| Test | `DynamicFlow.test.ts` | Import path swap only |

### Verification

- [x] `npm run check:react-peers` passes with empty allowlist
- [x] `npm run test:ci` passes — 1654 UI tests + 163 server tests, 0 errors
- [x] Pipeline graph renders correctly with drag, zoom, and pan (tested via Storybook)
- [x] Zero remaining `react-flow-renderer` references in source or lockfile

**Checklist:**


frontend/src/vitest.setup.ts
```diff
@@ -4,6 +4,31 @@ import { cleanup } from '@testing-library/react';
...
+if (typeof window !== 'undefined') {
+  window.addEventListener('error', (event: ErrorEvent) => {
+    if (
+      event.error instanceof TypeError &&
+      event.error.message.includes("Cannot read properties of null (reading 'document')")
+    ) {
+      event.preventDefault();
+    }
+  });
+}
```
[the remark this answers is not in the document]
PR3MM: The error message is specific to the d3-drag/nodrag.js codepath ,it only fires when event.view is null

frontend/src/pages/v2/DagCanvas.tsx
```diff
@@ -45,19 +58,25 @@ export default function DagCanvas({
...
 
-  elements.forEach(elem => {
-    // For each SubDag node, provide a callback function if expand button is clicked.
-    if (elem && elem.type === NodeTypeNames.SUB_DAG && elem.data) {
-      elem.data.expand = subDagExpand;
+  nodes.forEach(node => {
+    if (node && node.type === NodeTypeNames.SUB_DAG && node.data) {
+      node.data.expand = subDagExpand;
     }
   });
```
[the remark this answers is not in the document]
PR3MM: it is a pre-existing patterns that existed before migration

frontend/src/pages/PipelineDetailsV2.tsx
```diff
@@ -74,9 +78,7 @@ function PipelineDetailsV2({
             layers={layers}
             onLayersUpdate={layerChange}
             elements={pipelineFlowElements}
-            onElementClick={(event: ReactMouseEvent, element: Node | Edge) =>
-              setSelectedNode(element)
-            }
+            onElementClick={(_event, element) => setSelectedNode(element)}
             setFlowElements={() => {}}
           ></DagCanvas>
```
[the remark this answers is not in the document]
PR3MM: it is a pre-existing patterns that existed before migration

frontend/src/pages/PipelineDetailsV2.tsx
```diff
@@ -74,9 +78,7 @@ function PipelineDetailsV2({
             layers={layers}
             onLayersUpdate={layerChange}
             elements={pipelineFlowElements}
-            onElementClick={(event: ReactMouseEvent, element: Node | Edge) =>
-              setSelectedNode(element)
-            }
+            onElementClick={(_event, element) => setSelectedNode(element)}
```
manaswinidas, line 68: Can you confirm whether the dragging works fine here? The old react-flow-renderer v9 used mutable internal state for drag, so the no-op happened to work. Under the new controlled model, it won't. Either pass a real state setter or explicitly disable dragging with nodesDraggable={false} on the PipelineDetailsV2 usage path.

frontend/src/pages/v2/DagCanvas.tsx
```diff
@@ -45,19 +58,25 @@ export default function DagCanvas({
...
+  const nodes = useMemo(() => elements.filter(isNode), [elements]);
+  const edges = useMemo(() => elements.filter((el): el is Edge => !isNode(el)), [elements]);
+
+  const onNodesChange = useCallback(
+    (changes: NodeChange[]) => {
+      const updatedNodes = applyNodeChanges(changes, nodes) as PipelineNode[];
+      setFlowElements([...updatedNodes, ...edges]);
+    },
+    [nodes, edges, setFlowElements],
+  );
```
manaswinidas: Every node drag movement fires `onNodesChange`, which spreads the full nodes and edges arrays into a new array, triggers `setFlowElements`, which re-renders the parent, which re-filters into nodes/edges via `useMemo`. For pipelines with 100+ nodes, this could introduce noticeable drag jank. The old `onNodeDragStop` callback only updated on drag end, which was much more efficient.
Recommendation: Consider using `onNodeDragStop` instead of `onNodesChange` if only drag-position persistence is needed, or debounce the change handler.

frontend/src/vitest.setup.ts
```diff
@@ -4,6 +4,31 @@ import { cleanup } from '@testing-library/react';
...
+// @xyflow/react's d3-drag dependency accesses event.view.document on mousedown,
+// but jsdom leaves event.view null. Suppress the resulting uncaught TypeError so
+// Vitest does not treat it as a test failure.
+if (typeof window !== 'undefined') {
+  window.addEventListener('error', (event: ErrorEvent) => {
+    if (
+      event.error instanceof TypeError &&
+      event.error.message.includes("Cannot read properties of null (reading 'document')")
+    ) {
+      event.preventDefault();
```
manaswinidas: This runs for the entire test suite, not just `@xyflow/react` tests. While the error message check is specific, a beforeAll/afterAll scoped to relevant test files would be cleaner.

PR3MM:
### Visual verification

1. Static DAG (PipelineDetailsV2)
SCR-20260302-ozcy


2. Runtime DAG (RunDetailsV2)

SCR-20260302-ozof


Runtime states require MLMD data; the DAG canvas renders correctly. Full execution state coloring depends on a healthy cluster pipeline run.

3. Edge arrows (MarkerType.ArrowClosed)
Visible in the screenshots/videos, arrowheads render correctly on all edges.

4. Drag, zoom, and pan

https://github.com/user-attachments/assets/f7c739d0-a110-4a2d-8bbd-df003a354b21


5. SubDag expand/collapse

https://github.com/user-attachments/assets/62ed496c-c7eb-47d4-ad13-30783507e30a

frontend/src/pages/PipelineDetailsV2.tsx
PR3MM: Fixed. Added a nodesDraggable prop to DagCanvas and set nodesDraggable={false} in PipelineDetailsV2

frontend/src/pages/v2/DagCanvas.tsx
PR3MM: Replaced onNodesChange + applyNodeChanges with onNodeDragStop, which only fires once when the user releases the node.

frontend/src/vitest.setup.ts
PR3MM: I tried moving it to a beforeAll/afterAll helper registered only in the 2 affected test files (PipelineDetailsV2.test.tsx, RunDetailsV2.test.tsx). It passes when running those files individually, but fails in the full parallel suite . Vitest's process level uncaught-exception monitor catches the error before the per-file window.addEventListener handler can intercept it.

jeffspahr:
@PR3MM can you also rebase?

PR3MM:
Rebased onto master and addressed all review feedback:
- Fixed memoized node data mutation — subDagExpand is now integrated during useMemo computation instead of mutating the memoized array
- Deduplicated isNode — exported from StaticFlow.ts and removed the duplicate in DagCanvas.tsx
- Documented controlled flow — added comment explaining intentional omission of onNodesChange/onEdgesChange
- Documented test setup workaround — added comment explaining the global error suppression for the jsdom/d3-drag incompatibility
- Simplified story onInit — changed to onInit={instance => instance.fitView()}
- DCO sign-off — all commits signed

All 1654 UI tests + 163 server tests pass locally. Screenshots and recordings from the previous round still apply (no visual changes).

frontend/src/pages/RunDetailsV2.tsx
```diff
@@ -84,13 +82,18 @@ export function RunDetailsV2(props: RunDetailsV2Props) {
...
   const [selectedTab, setSelectedTab] = useState(0);
-  const [selectedNode, setSelectedNode] = useState<FlowElement<FlowElementDataBase> | null>(null);
+  const [selectedNode, setSelectedNode] = useState<PipelineFlowElement | null>(null);
   const [selectedNodeMlmdInfo, setSelectedNodeMlmdInfo] = useState<NodeMlmdInfo | null>(null);
   const [, forceUpdate] = useState();
   const [runFinished, setRunFinished] = useState(false);
 
-  const getNodeName = function (element: FlowElement<FlowElementDataBase> | null): string {
-    if (element && element.data && element.data.label) {
+  const getNodeName = function (element: PipelineFlowElement | null): string {
```
manaswinidas: This `getNodeName` function is identical to the one in `PipelineDetailsV2.tsx`. Consider extracting it into a shared utility (e.g., in `StaticFlow.ts` alongside `isNode` and `PipelineFlowElement`) and importing it in both pages.

frontend/src/lib/v2/StaticFlow.ts
```diff
@@ -13,22 +13,20 @@
...
 import { ArtifactFlowElementData, FlowElementDataBase } from 'src/components/graph/Constants';
 import ExecutionNode from 'src/components/graph/ExecutionNode';
 import SubDagNode from 'src/components/graph/SubDagNode';
 import { ComponentSpec, PipelineSpec, PipelineTaskSpec } from 'src/generated/pipeline_spec';
 import { ComponentInputsSpec_ArtifactSpec } from 'src/generated/pipeline_spec/pipeline_spec';
 
+export type PipelineFlowElement = Node<FlowElementDataBase> | Edge;
+
+export function isNode(el: PipelineFlowElement): el is Node<FlowElementDataBase> {
+  return 'position' in el;
```
manaswinidas: Nit: `'position' in el` works today since `Edge` has no `position` property, but it's a structural check that could break silently if `@xyflow/react` adds one in a future version. A more defensive guard:

```ts
return 'position' in el && !('source' in el);
```

`source` is always present on `Edge` and never on `Node`, making this robust in both directions.

frontend/src/pages/v2/DagCanvas.tsx
```diff
@@ -44,46 +53,59 @@ export default function DagCanvas({
...
-                  if (value.id === node.id) {
-                    return node;
-                  }
-                  return value;
-                }),
-              );
-            }}
+            onNodeClick={(event, node) => onElementClick(event, node)}
+            onEdgeClick={(event, edge) => onElementClick(event, edge)}
+            onNodeDragStop={nodesDraggable ? onNodeDragStop : undefined}
```
manaswinidas, line 117: Nit: These wrappers create new function references on every render, which bypasses React Flow's internal memoization for click handlers. Since the signatures align, you could wrap these with `useCallback`:

```ts
const handleNodeClick = useCallback(
  (event: ReactMouseEvent, node: PipelineNode) => onElementClick(event, node),
  [onElementClick],
);
```

Or pass `onElementClick` directly if the generic types align. Minor for typical DAG sizes.

frontend/src/pages/v2/DagCanvas.tsx
```diff
@@ -14,28 +14,37 @@
...
   Controls,
   Edge,
-  Elements,
   MiniMap,
   Node,
-  OnLoadParams,
-  ReactFlowProvider,
-} from 'react-flow-renderer';
+} from '@xyflow/react';
+import '@xyflow/react/dist/style.css';
```
manaswinidas: Nit: This CSS import also appears in all 4 story files. For the app bundle, a single import at the app entry point would avoid duplicate style injection. The story-level imports are fine for Storybook isolation.

frontend/src/pages/RunDetailsV2.tsx
PR3MM: Extracted shared getNodeName — moved to StaticFlow.ts, imported in both PipelineDetailsV2.tsx and RunDetailsV2.tsx

frontend/src/lib/v2/StaticFlow.ts
PR3MM: Defensive isNode guard — added && !('source' in el) for future-proofing

frontend/src/pages/v2/DagCanvas.tsx
PR3MM: Wrapped click handlers with useCallback — handleNodeClick and handleEdgeClick now have stable references

frontend/src/pages/v2/DagCanvas.tsx
PR3MM: Moved CSS import to index.tsx — single app-level import, story files keep their local imports for Storybook isolation

jeffspahr:
## Code Review (Opus 4.6)

### Overview

This PR replaces the deprecated `react-flow-renderer` v9 with `@xyflow/react` v12 as part of the React 18/19 upgrade path. The migration is well-structured — it preserves the existing mixed `elements` array pattern internally and only splits into `nodes[]`/`edges[]` at the rendering boundary (`DagCanvas`). This is a smart strategy that keeps the diff focused.

**+355 / -334 across 20 files** — clean swap with minimal scope creep.

---

### Positives

- **Well-documented PR description** — the migration table, file-by-file breakdown, and verification checklist are excellent.
- **Minimal blast radius** — the `PipelineFlowElement` union type preserves the existing internal API, limiting changes to the rendering boundary.
- **Proper memoization** — `DagCanvas` correctly uses `useMemo`/`useCallback` for the split + handlers, avoiding unnecessary re-renders.
- **`getNodeName` deduplication** — extracting the shared helper from two pages into `StaticFlow.ts` is a clean improvement.
- **Dependencies cleaned up** — `react-redux`, `redux`, `react-draggable`, and `classnames` removed as transitive deps of the old library. Good hygiene.

---

### Issues & Suggestions

#### 1. `isNode` guard may produce false positives (Medium)

**`StaticFlow.ts`:**
```ts
export function isNode(el: PipelineFlowElement): el is Node<FlowElementDataBase> {
  return 'position' in el && !('source' in el);
}
```

This duck-typing check is fragile. If an edge ever gains a `position` property (e.g., from a library update or custom data), it breaks. `@xyflow/react` doesn't export an `isNode` utility anymore, but a safer guard would check for the presence of node-specific fields more defensively:

```ts
export function isNode(el: PipelineFlowElement): el is Node<FlowElementDataBase> {
  return !('source' in el) && !('target' in el);
}
```

Edges are defined by having both `source` and `target`. Checking for the absence of both is more semantically correct than relying on `position`.

#### 2. Error suppression in `vitest.setup.ts` is broad (Medium)

```ts
window.addEventListener('error', (event: ErrorEvent) => {
  if (
    event.error instanceof TypeError &&
    event.error.message.includes("Cannot read properties of null (reading 'document')")
  ) {
    event.preventDefault();
  }
});
```

This swallows **all** `TypeError`s matching that substring globally, not just ones from d3-drag. If a real bug produces the same message, tests won't catch it. Consider:
- Adding a stack trace check (e.g., `event.error.stack?.includes('d3-drag')`) to narrow the scope.
- Filing a tracking issue for removal (the TODO comment is good, but a linked issue is better for visibility).

#### 3. `onNodeDragStop` rebuilds the full elements array on every drag (Low)

```ts
const onNodeDragStop = useCallback(
  (_event: ReactMouseEvent, draggedNode: PipelineNode) => {
    const updatedElements = elements.map((el) =>
      isNode(el) && el.id === draggedNode.id ? { ...el, position: draggedNode.position } : el,
    );
    setFlowElements(updatedElements);
  },
  [elements, setFlowElements],
);
```

This is fine for typical DAG sizes, but the `useCallback` dependency on `elements` means a new callback is created on every elements change, which defeats the memoization benefit. For a DAG viewer this is unlikely to matter, but worth noting.

#### 4. `nodesDraggable` prop defaults to `true` but `PipelineDetailsV2` passes `false` (Nit)

In `DagCanvas.tsx`, the default is `nodesDraggable = true`, and `PipelineDetailsV2` passes `nodesDraggable={false}`. This is correct, but the conditional `onNodeDragStop={nodesDraggable ? onNodeDragStop : undefined}` creates the handler even when it won't be used. Minor — not worth changing, just noting the pattern.

#### 5. Type assertion in `RunDetailsV2.tsx` (Low)

```ts
const runId = (props.match.params as Record<string, string>)[RouteParams.runId];
```

This `as Record<string, string>` cast was presumably needed because a type changed. It would be cleaner to have the route params typed properly (e.g., `RouteComponentProps<{ [RouteParams.runId]: string }>`), but that's a pre-existing issue, not introduced by this PR.

#### 6. CSS import location (Nit)

```ts
// frontend/src/index.tsx
import '@xyflow/react/dist/style.css';
```

Global import at the app entry point is fine and matches the migration guide. Just verify this doesn't conflict with any existing CSS overrides for the flow renderer (e.g., custom `.react-flow` class styles).

---

### Test Infrastructure

The `vitest.config.mts` change to inline `@xyflow/react` and `@xyflow/system` is correct — these ESM packages need inlining for Vitest's jsdom environment. The `DOMMatrixReadOnly` polyfill is a known jsdom gap and the mock is minimal. Both are appropriate.

---

### Security / Performance

- No security concerns — this is a pure UI library swap with no API or data changes.
- Performance should be neutral or slightly better (`@xyflow/react` v12 uses zustand internally instead of redux, which is lighter).

---

### Verdict

**This is a well-executed library migration.** The strategy of preserving the internal `elements` array and only splitting at the render boundary is pragmatic and keeps the diff manageable. The main suggestions are around hardening the `isNode` guard and narrowing the error suppression scope in tests — neither is a blocker.

frontend/src/pages/RunDetailsV2.tsx
```diff
@@ -75,7 +73,7 @@ interface RunDetailsV2Info {
 export type RunDetailsV2Props = RunDetailsV2Info & RunDetailsProps;
 
 export function RunDetailsV2(props: RunDetailsV2Props) {
-  const runId = props.match.params[RouteParams.runId];
+  const runId = (props.match.params as Record<string, string>)[RouteParams.runId];
```
manaswinidas: This `as Record<string, string>` cast is unrelated to the react-flow migration and masks a type error rather than fixing it. If `match.params` does not type-check for index access, the proper fix is to type the route params at the component level (e.g., via `RouteComponentProps<{ [RouteParams.runId]: string }>`) so TypeScript can verify the param actually exists. With the cast, accessing a param that is missing at runtime silently returns `undefined` instead of producing a compile-time error.

This appears in two places in this file (lines 76 and 158). Consider fixing the underlying typing or splitting this into a separate commit/PR so the migration stays scoped.

PR3MM: Fixed in commit 8f999e9. Replaced the as Record<string, string> cast with proper route param typing via RouteComponentProps<RunDetailsV2Params>. Defined RunDetailsV2Params interface with [RouteParams.runId]: string, applied it to both RunDetailsV2 and RunDetailsRouter so TypeScript verifies the param exists at compile time.

jeffspahr:
## Suggestion: Replace error suppression with MouseEvent polyfill

The global `window.addEventListener('error', ...)` handler in `vitest.setup.ts` suppresses the d3-drag `TypeError` rather than fixing the root cause. The TODO comment says to remove it "once jsdom natively supports event.view on synthetic mouse events," but jsdom has explicitly declined to fix this — [jsdom#3935](https://github.com/jsdom/jsdom/issues/3935) was closed as "not planned."

The root issue is that jsdom doesn't default `MouseEvent.view` to `window` like real browsers do. d3-drag then crashes accessing `event.view.document` on a `null` view. Instead of catching and swallowing the error, we can polyfill the gap directly:

```ts
// jsdom does not default MouseEvent.view to window (jsdom/jsdom#3935, closed "not planned").
// d3-drag accesses event.view.document on mousedown, which throws when view is null.
// Polyfill the constructor to match real browser behavior.
const OrigMouseEvent = globalThis.MouseEvent;
globalThis.MouseEvent = class extends OrigMouseEvent {
  constructor(type: string, init?: MouseEventInit) {
    super(type, { view: window, ...init });
  }
} as typeof MouseEvent;
```

This fixes the problem at the source — d3-drag gets a real `view.document` — and the error suppression handler + its TODO can be removed entirely.

jeffspahr:
## Two minor nits

### 1. Missing newline at EOF in `vitest.config.mts`

The diff shows `\ No newline at end of file` was introduced. Trivial fix.

### 2. `DOMMatrixReadOnly` mock is minimal

The mock in `vitest.setup.ts` only implements `m22`. If xyflow starts using other matrix properties (e.g., `m11`, `m41`, `m42` for transforms), tests will silently get `undefined` instead of clear errors. Consider filling in the identity matrix defaults:

```ts
class DOMMatrixReadOnlyMock {
  m11 = 1; m12 = 0; m13 = 0; m14 = 0;
  m21 = 0; m22 = 1; m23 = 0; m24 = 0;
  m31 = 0; m32 = 0; m33 = 1; m34 = 0;
  m41 = 0; m42 = 0; m43 = 0; m44 = 1;
  constructor(_init?: string) {}
}
```

PR3MM:
Addressed all feedback from @jeffspahr's latest review:
1. MouseEvent.view polyfill (replaces error suppression)
Tried the suggested constructor polyfill (class extends OrigMouseEvent with { view: window, ...init }), but jsdom's internal UIEvent-impl.js rejects the view property during construction , crashes with "member view is not of type Window". Also tried a Proxy wrapper and a UIEvent.prototype.view getter override; both break jsdom's WebIDL validation in different ways.
The working approach: capture-phase listeners on window for mousedown and mouseup that patch event.view from null to window via Object.defineProperty after jsdom finishes construction but before d3-drag reads it. Same end result — d3-drag gets a real view.document, no error suppression needed — just a different mechanism to work around jsdom's construction constraints. Reference: jsdom#3935 (closed "not planned").
2. Missing newline at EOF in vitest.config.mts — Fixed.
3. DOMMatrixReadOnly mock — Expanded to full 4×4 identity matrix (m11–m44).
All CI checks pass locally: format, lint, typecheck, 1736 UI tests + 165 server tests, 0 errors.

jeffspahr:
Heads up — this branch now has merge conflicts with master (likely from #12946 merging). Three files need resolution:

- **`frontend/package.json`** — dependency entries overlap with the react-query → @tanstack/react-query migration
- **`frontend/package-lock.json`** — regenerate after resolving `package.json`
- **`frontend/src/pages/RunDetailsV2.tsx`** — both PRs touched this file (#12946 changed query hooks, this PR changed flow/node handling). You'll need to keep both sets of changes.

A rebase onto latest master + `npm install` should resolve it cleanly.

PR #12951, opened 2026-03-02
test(backend): add test coverage for common/util package

jsonmp-k8:
## Summary
- Add 7 new test files and expand `error_test.go` to increase coverage for the `common/util` package from **36.3% to 53.6%**
- New test files cover `pointer.go`, `time.go`, `json.go`, `uuid.go`, `metrics.go`, `tls_config.go`, and `service.go` (ExtractMasterIPAndPort)
- Expand `error_test.go` from 1 test to 15+ test functions covering CustomError, all 14 UserError constructors, UserError methods, Wrapf/Wrap, IsUserErrorCodeMatch, ToRpcStatus, ToError, ToGRPCError, ToGRPCStatus, and GRPCStatus

## Bug fix
- **`tls_config.go`**: Fixed a bug where `GetTLSConfig` silently returned `(nil, nil)` when `AppendCertsFromPEM` failed on invalid PEM content. The original code returned the stale `err` from `os.ReadFile` (which was `nil` since the read succeeded), so callers received no TLS config and no error. Now returns a descriptive error: `"failed to parse certificate from <path>"`.

## Test plan
- [x] `CGO_ENABLED=0 go test -v -cover ./backend/src/common/util/...` — all tests pass, 53.6% coverage
- [x] `gofmt -w` — all test files formatted
- [x] `golangci-lint run --new --new-from-merge-base=origin/master` — 0 issues

### DIFF
--- a/backend/src/common/util/error_test.go
+++ b/backend/src/common/util/error_test.go
@@ -21,6 +21,7 @@ import (
 	"github.com/go-openapi/runtime"
 	"github.com/stretchr/testify/assert"
 	"google.golang.org/grpc/codes"
+	grpcStatus "google.golang.org/grpc/status"
 	k8errors "k8s.io/apimachinery/pkg/api/errors"
 	"k8s.io/apimachinery/pkg/runtime/schema"
 )
@@ -35,14 +36,14 @@ func TestIsNotFound(t *testing.T) {
 func TestNewCustomError(t *testing.T) {
 	cause := fmt.Errorf("root cause")
 	customError := NewCustomError(cause, CUSTOM_CODE_TRANSIENT, "something went %s", "wrong")
-	assert.Contains(t, customError.Error(), "CustomError (code: 0)")
+	assert.Contains(t, customError.Error(), fmt.Sprintf("CustomError (code: %d)", CUSTOM_CODE_TRANSIENT))
 	assert.Contains(t, customError.Error(), "something went wrong")
 	assert.Contains(t, customError.Error(), "root cause")
 }
 
 func TestNewCustomErrorf(t *testing.T) {
 	customError := NewCustomErrorf(CUSTOM_CODE_PERMANENT, "item %d not found", 42)
-	assert.Contains(t, customError.Error(), "CustomError (code: 1)")
+	assert.Contains(t, customError.Error(), fmt.Sprintf("CustomError (code: %d)", CUSTOM_CODE_PERMANENT))
 	assert.Contains(t, customError.Error(), "item 42 not found")
 }
 
@@ -359,11 +360,12 @@ func TestToError(t *testing.T) {
 		assert.Nil(t, ToError(nil))
 	})
 
-	t.Run("with status returns error", func(t *testing.T) {
+	t.Run("with status returns error preserving code", func(t *testing.T) {
 		userError := NewNotFoundError(fmt.Errorf("cause"), "not found")
 		rpcStatus := ToRpcStatus(userError)
 		err := ToError(rpcStatus)
 		assert.NotNil(t, err)
+		assert.Equal(t, codes.NotFound, grpcStatus.Code(err))
 	})
 }
 
--- a/backend/src/common/util/metrics_test.go
+++ b/backend/src/common/util/metrics_test.go
@@ -21,24 +21,26 @@ import (
 	"github.com/stretchr/testify/assert"
 )
 
-func TestGetMetricValueWithGauge(t *testing.T) {
-	gauge := prometheus.NewGauge(prometheus.GaugeOpts{
-		Name: "test_gauge",
-		Help: "A test gauge",
+func TestGetMetricValue(t *testing.T) {
+	t.Run("gauge", func(t *testing.T) {
+		gauge := prometheus.NewGauge(prometheus.GaugeOpts{
+			Name: "test_gauge",
+			Help: "A test gauge",
+		})
+		gauge.Set(42.5)
+
+		value := GetMetricValue(gauge)
+		assert.Equal(t, 42.5, value)
 	})
-	gauge.Set(42.5)
 
-	value := GetMetricValue(gauge)
-	assert.Equal(t, 42.5, value)
-}
+	t.Run("counter", func(t *testing.T) {
+		counter := prometheus.NewCounter(prometheus.CounterOpts{
+			Name: "test_counter",
+			Help: "A test counter",
+		})
+		counter.Add(10)
 
-func TestGetMetricValueWithCounter(t *testing.T) {
-	counter := prometheus.NewCounter(prometheus.CounterOpts{
-		Name: "test_counter",
-		Help: "A test counter",
+		value := GetMetricValue(counter)
+		assert.Equal(t, float64(10), value)
 	})
-	counter.Add(10)
-
-	value := GetMetricValue(counter)
-	assert.Equal(t, float64(10), value)
 }
--- a/backend/src/common/util/tls_config_test.go
+++ b/backend/src/common/util/tls_config_test.go
@@ -15,9 +15,17 @@
 package util
 
 import (
+	"crypto/ecdsa"
+	"crypto/elliptic"
+	"crypto/rand"
+	"crypto/x509"
+	"crypto/x509/pkix"
+	"encoding/pem"
+	"math/big"
 	"os"
 	"path/filepath"
 	"testing"
+	"time"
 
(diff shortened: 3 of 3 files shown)


backend/src/common/util/tls_config_test.go
```diff
@@ -0,0 +1,33 @@
...
+	config, err := GetTLSConfig("")
+	assert.Nil(t, config)
+	assert.NoError(t, err)
+}
+
+func TestGetTLSConfigNonexistentPath(t *testing.T) {
+	config, err := GetTLSConfig("/nonexistent/path/to/cert.pem")
+	assert.Nil(t, config)
+	assert.Error(t, err)
+}
```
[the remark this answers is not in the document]
jsonmp-k8: Fixed. Now using `filepath.Join(t.TempDir(), "nonexistent-cert.pem")` to build a deterministic nonexistent path.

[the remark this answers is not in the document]
jsonmp-k8: Good catch. Added `TestGetTLSConfigInvalidPEM` that writes invalid PEM content to a temp file and verifies the `AppendCertsFromPEM` false branch is covered.

backend/src/common/util/error_test.go
```diff
@@ -15,14 +15,382 @@
...
+}
+
+// GRPCStatus method test
+
+func TestGRPCStatus(t *testing.T) {
+	userError := NewNotFoundError(fmt.Errorf("cause"), "not found")
+	grpcStatus := userError.GRPCStatus()
+	assert.NotNil(t, grpcStatus)
+	assert.Equal(t, codes.NotFound, grpcStatus.Code())
+	assert.NotEmpty(t, grpcStatus.Details())
```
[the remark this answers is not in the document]
jsonmp-k8: Fixed. Now capturing the return value into a `details` variable and asserting `len(details) > 0`.

backend/src/common/util/tls_config.go
```diff
@@ -33,7 +34,7 @@ func GetTLSConfig(caCertPath string) (*tls.Config, error) {
 			return nil, err
 		}
 		if ok := caCertPool.AppendCertsFromPEM(caCert); !ok {
```
jeffspahr, line 36: Good catch: `AppendCertsFromPEM` returns `false` on failure, but `err` at this point is `nil` (the `os.ReadFile` above succeeded). The original `return nil, err` was silently returning `(nil, nil)` to callers — no TLS config and no error propagated. This is an important correctness fix and is worth calling out explicitly in the PR description rather than leaving it as an implicit side-effect of adding tests.

backend/src/common/util/error_test.go
```diff
@@ -15,14 +15,400 @@
...
+	assert.Equal(t, true, IsNotFound(k8errors.NewNotFound(schema.GroupResource{}, "NAME")))
+	assert.Equal(t, false, IsNotFound(k8errors.NewAlreadyExists(schema.GroupResource{}, "NAME")))
+}
+
+// CustomError tests
+
+func TestNewCustomError(t *testing.T) {
+	cause := fmt.Errorf("root cause")
+	customError := NewCustomError(cause, CUSTOM_CODE_TRANSIENT, "something went %s", "wrong")
+	assert.Contains(t, customError.Error(), "CustomError (code: 0)")
```
jeffspahr: This hardcodes the numeric value of `CUSTOM_CODE_TRANSIENT` (0). If the iota order ever changes this test will pass while asserting the wrong constant. Prefer:
```go
assert.Contains(t, customError.Error(), fmt.Sprintf("CustomError (code: %d)", CUSTOM_CODE_TRANSIENT))
```

backend/src/common/util/error_test.go
```diff
@@ -15,14 +15,400 @@
...
+	assert.Contains(t, customError.Error(), "something went wrong")
+	assert.Contains(t, customError.Error(), "root cause")
+}
+
+func TestNewCustomErrorf(t *testing.T) {
+	customError := NewCustomErrorf(CUSTOM_CODE_PERMANENT, "item %d not found", 42)
+	assert.Contains(t, customError.Error(), "CustomError (code: 1)")
```
jeffspahr: Same issue — hardcodes `CUSTOM_CODE_PERMANENT == 1`. Use the constant name in the assertion string to make the intent explicit and guard against reordering.

backend/src/common/util/error_test.go
```diff
@@ -15,14 +15,400 @@
...
+func TestToError(t *testing.T) {
+	t.Run("nil returns nil", func(t *testing.T) {
+		assert.Nil(t, ToError(nil))
+	})
+
+	t.Run("with status returns error", func(t *testing.T) {
+		userError := NewNotFoundError(fmt.Errorf("cause"), "not found")
+		rpcStatus := ToRpcStatus(userError)
+		err := ToError(rpcStatus)
+		assert.NotNil(t, err)
```
jeffspahr, line 367: `assert.NotNil(t, err)` verifies the call does not return nil but nothing about the error content. Since this is a round-trip through `ToRpcStatus` → `ToError`, consider also asserting the status code is preserved — e.g. check that `status.Code(err) == codes.NotFound`.

backend/src/common/util/tls_config_test.go
```diff
@@ -0,0 +1,47 @@
...
+func TestGetTLSConfigInvalidPEM(t *testing.T) {
+	invalidPEMFile := filepath.Join(t.TempDir(), "invalid-cert.pem")
+	err := os.WriteFile(invalidPEMFile, []byte("this is not valid PEM content"), 0644)
+	assert.NoError(t, err)
+
+	config, err := GetTLSConfig(invalidPEMFile)
+	assert.Nil(t, config)
+	assert.Error(t, err)
+	assert.Contains(t, err.Error(), "failed to parse certificate")
+}
```
jeffspahr, line 83: The three tests cover empty path, missing file, and invalid PEM — but the happy path (valid PEM → non-nil config, nil error) goes untested. Adding a test with a self-signed cert written to `t.TempDir()` would complete coverage and would have caught the original silent-nil bug had tests existed before the fix.

backend/src/common/util/metrics_test.go
```diff
@@ -0,0 +1,44 @@
...
+package util
+
+import (
+	"testing"
+
+	"github.com/prometheus/client_golang/prometheus"
+	"github.com/stretchr/testify/assert"
+)
+
+func TestGetMetricValueWithGauge(t *testing.T) {
```
jeffspahr: Minor consistency nit: the rest of this PR uses `t.Run(...)` subtests inside a single `TestFoo` function. These two are separate top-level functions instead. Not blocking, but for consistency:
```go
func TestGetMetricValue(t *testing.T) {
    t.Run("gauge", func(t *testing.T) { ... })
    t.Run("counter", func(t *testing.T) { ... })
}
```

backend/src/common/util/tls_config.go
jsonmp-k8: Good call. Updated the PR description with a dedicated **Bug fix** section calling this out explicitly.

backend/src/common/util/error_test.go
jsonmp-k8: Fixed. Now using `fmt.Sprintf("CustomError (code: %d)", CUSTOM_CODE_TRANSIENT)` instead of the hardcoded `"0"`.

backend/src/common/util/error_test.go
jsonmp-k8: Fixed. Same change applied — using `CUSTOM_CODE_PERMANENT` constant instead of hardcoded `"1"`.

backend/src/common/util/error_test.go
jsonmp-k8: Fixed. Now asserting `grpcStatus.Code(err) == codes.NotFound` to verify the status code survives the round-trip.

backend/src/common/util/tls_config_test.go
jsonmp-k8: Added `TestGetTLSConfigValidCert` that generates a self-signed ECDSA certificate, writes it to `t.TempDir()`, and asserts the returned config is non-nil with populated RootCAs.

backend/src/common/util/metrics_test.go
jsonmp-k8: Fixed. Consolidated into a single `TestGetMetricValue` with `t.Run("gauge", ...)` and `t.Run("counter", ...)` subtests.

PR #12963, opened 2026-03-05
fix(frontend): add npm cache retry to prevent EEXIST race condition in Docker build

droctothorpe:
## Summary

- `frontend/Dockerfile` intermittently fails during `npm ci` with an `EEXIST`/`ENOENT` race condition: npm's parallel download workers race to rename a temp file to the same cache destination; one worker succeeds, and the other's rename fails because the tmp source is gone (`ENOENT`) while the target already exists (`EEXIST`)
- Adds a retry with `npm cache clean --force` on failure in both the build stage (full frontend deps) and the final stage (server-only deps)
- Fixes the `KFP upgrade tests` failure seen at https://github.com/kubeflow/pipelines/actions/runs/22699025197

Signed-off-by: droctothorpe <mythicalsunlight@gmail.com>

### DIFF
--- a/frontend/Dockerfile
+++ b/frontend/Dockerfile
@@ -14,7 +14,9 @@ COPY . .
 
 WORKDIR ./frontend
 
-RUN npm ci && npm run postinstall
+# Retry with cache clean as a workaround for an intermittent npm EEXIST/ENOENT
+# race condition where parallel download workers collide on the same cache entry.
+RUN npm ci && npm run postinstall || (npm cache clean --force && npm ci && npm run postinstall)
 RUN npm run build
 
 RUN mkdir -p ./server/dist && \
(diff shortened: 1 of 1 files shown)


jeffspahr:
Nice targeted fix for the npm cache race.

One non-blocking suggestion in the build stage:

Current:
```Dockerfile
RUN npm ci && npm run postinstall || (npm cache clean --force && npm ci && npm run postinstall)
```

Suggested:
```Dockerfile
RUN (npm ci || (npm cache clean --force && npm ci)) && npm run postinstall
```

Reason: this scopes retry to `npm ci` only. With the current line, the fallback also runs when `npm run postinstall` fails, which likely won’t be fixed by cleaning npm cache.

PR #12976, opened 2026-03-06
fix(frontend): remove CVE-2026-3449 from ui server

jeffspahr:

Removes the `frontend/server` dependency path that pulls in `@tootallnate/once` via `@google-cloud/storage -> @google-cloud/common -> teeny-request -> http-proxy-agent`.

This keeps the existing GCS artifact behavior by replacing the narrow `@google-cloud/storage` usage with a small `google-auth-library` based helper that:
- lists matching objects through the GCS JSON API
- downloads object streams for preview and full artifact reads
- preserves provider-info based credentials from Kubernetes secrets
- reuses one auth client per GCS artifact request instead of constructing a fresh auth client for each list/download helper call

It also adds direct `gcs-helper` unit coverage for auth setup, pagination, and encoded URL construction, and extends the artifact integration tests to cover wildcard multi-file GCS reads.

Additionally, this fixes the pre-existing `integration-tests/artifact-proxy.test.ts` failures by removing the hard-coded local port used by the fake namespaced artifact service. The test now binds to an ephemeral loopback port, which avoids collisions with local `kubectl` port-forwards and other dev services that commonly use `3002`.

Two correctness fixes are also now called out explicitly:
- the GCS provider-info path now returns immediately after sending the missing-namespace `500`, avoiding follow-on handler execution after a response is already sent
- the multi-file GCS read path is now sequential and deterministic; the old `forEach(... on('end'))` approach could send the response when the last-indexed file finished even if earlier files had not

**Verification:**
- `cd frontend/server && npx vitest run gcs-helper.test.ts --reporter=dot`
- `cd frontend/server && npx vitest run integration-tests/artifact-get.test.ts integration-tests/artifact-proxy.test.ts --reporter=dot`
- `cd frontend/server && npm run format:check`
- `cd frontend/server && npm audit --package-lock-only --omit=dev --json`
  - `@tootallnate/once` / `http-proxy-agent` / `teeny-request` / `@google-cloud/storage` are no longer present
  - remaining prod audit findings are unrelated `fast-xml-parser` / `@aws-sdk/xml-builder`

**Checklist:**
- [ ] Do you want this pull request (PR) cherry-picked into the current release branch?

### DIFF
--- a/frontend/server/gcs-helper.ts
+++ b/frontend/server/gcs-helper.ts
@@ -18,12 +18,14 @@ import { Readable } from 'stream';
 const GCS_SCOPE = 'https://www.googleapis.com/auth/devstorage.read_write';
 const GCS_API_BASE = 'https://storage.googleapis.com/storage/v1';
 
+export type GCSClient = Awaited<ReturnType<GoogleAuth['getClient']>>;
+
 interface GCSListResponse {
   items?: Array<{ name?: string }>;
   nextPageToken?: string;
 }
 
-async function getGCSClient(credentials?: CredentialBody) {
+export async function getGCSClient(credentials?: CredentialBody): Promise<GCSClient> {
   const auth = new GoogleAuth({
     credentials,
     scopes: GCS_SCOPE,
@@ -52,14 +54,15 @@ export async function listGCSObjectNames(options: {
   bucket: string;
   prefix: string;
   credentials?: CredentialBody;
+  client?: GCSClient;
 }): Promise<string[]> {
-  const { bucket, prefix, credentials } = options;
-  const client = await getGCSClient(credentials);
+  const { bucket, prefix, credentials, client } = options;
+  const resolvedClient = client ?? (await getGCSClient(credentials));
   const objectNames: string[] = [];
 
   let pageToken: string | undefined;
   do {
-    const response = await client.request<GCSListResponse>({
+    const response = await resolvedClient.request<GCSListResponse>({
       url: getListObjectsUrl(bucket, prefix, pageToken),
     });
     objectNames.push(
@@ -77,10 +80,11 @@ export async function downloadGCSObjectStream(options: {
   bucket: string;
   objectName: string;
   credentials?: CredentialBody;
+  client?: GCSClient;
 }): Promise<Readable> {
-  const { bucket, objectName, credentials } = options;
-  const client = await getGCSClient(credentials);
-  const response = await client.request<Readable>({
+  const { bucket, objectName, credentials, client } = options;
+  const resolvedClient = client ?? (await getGCSClient(credentials));
+  const response = await resolvedClient.request<Readable>({
     responseType: 'stream',
     url: getDownloadObjectUrl(bucket, objectName),
   });
--- a/frontend/server/handlers/artifacts.ts
+++ b/frontend/server/handlers/artifacts.ts
@@ -25,7 +25,8 @@ import { Handler, Request, Response } from 'express';
 import { createProxyMiddleware } from 'http-proxy-middleware';
 import { HACK_FIX_HPM_PARTIAL_RESPONSE_HEADERS } from '../consts.js';
 import { URL } from 'url';
-import { listGCSObjectNames, downloadGCSObjectStream } from '../gcs-helper.js';
+import { getGCSClient, listGCSObjectNames, downloadGCSObjectStream } from '../gcs-helper.js';
+import type { GCSClient } from '../gcs-helper.js';
 
 import * as fs from 'fs';
 import { isAllowedDomain } from './domain-checker.js';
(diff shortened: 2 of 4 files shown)


frontend/server/gcs-helper.ts
```diff
@@ -0,0 +1,88 @@
...
+}
+
+async function getGCSClient(credentials?: CredentialBody) {
+  const auth = new GoogleAuth({
+    credentials,
+    scopes: GCS_SCOPE,
+  });
+  return auth.getClient();
+}
+
```
manaswinidas, line 35: `getGCSClient` creates a new `GoogleAuth` + `getClient()` on every call. In the non-peek handler path, a request with N matching files triggers N+1 auth constructions (1 list + N downloads). The old `Storage` class cached its auth client internally, so this is a performance regression. Consider caching the client or accepting a pre-built one.

frontend/server/gcs-helper.ts
```diff
@@ -0,0 +1,88 @@
...
+        .filter((name): name is string => typeof name === 'string' && name.length > 0),
+    );
+    pageToken = response.data.nextPageToken;
+  } while (pageToken);
+
+  return objectNames;
+}
+
+export async function downloadGCSObjectStream(options: {
+  bucket: string;
```
manaswinidas, line 80: The pagination loop, URL construction, and name-filtering logic here have zero test coverage — the integration tests mock this entire module via `vi.mock('../gcs-helper.js')`. A unit test that stubs `client.request` would cover the `nextPageToken` loop, `encodeURIComponent` edge cases, and empty-name filtering without needing real GCS credentials.

frontend/server/handlers/artifacts.ts
```diff
@@ -309,47 +322,41 @@ function getGCSArtifactHandler(
...
-      const files = await storage.bucket(bucket).getFiles({ prefix });
-      const matchingFiles = files[0].filter((f) => {
-        // Escape regex characters
-        const escapeRegexChars = (s: string) => s.replace(/[|\\{}()[\]^$+*?.]/g, '\\$&');
-        // Build a RegExp object that only recognizes asterisks ('*'), and
-        // escapes everything else.
-        const regex = new RegExp('^' + key.split(/\*+/).map(escapeRegexChars).join('.*') + '$');
-        return regex.test(f.name);
-      });
+      const matchingFiles = (await listGCSObjectNames({ bucket, prefix, credentials })).filter(
```
manaswinidas: No test exercises the wildcard/multi-file path — all three GCS tests use an exact single filename. The regex construction (`escapeRegexChars`, `key.split(/\*+/)`) and the sequential concatenation loop were rewritten in this PR but remain untested. At least one test with a `key` containing `*` and multiple mock files returned from `listGCSObjectNames` would cover this.

frontend/server/handlers/artifacts.ts
```diff
@@ -293,14 +305,15 @@ function getGCSArtifactHandler(
...
+      let credentials: CredentialBody | undefined;
       if (providerInfoString) {
         const providerInfo = parseJSONString<GCSProviderInfo>(providerInfoString);
         if (providerInfo && providerInfo.Params.fromEnv === 'false') {
           if (!namespace) {
             res.status(500).send('Failed to parse provider info. Reason: No namespace provided');
+            return;
           } else {
-            storageOptions = await parseGCSProviderInfo(providerInfo, namespace);
+            credentials = await parseGCSProviderInfo(providerInfo, namespace);
```
manaswinidas, line 318: Good catch — the old code was missing this `return`, so it continued executing after sending the 500 response, which could cause "headers already sent" crashes. Worth calling out as an explicit bug fix in the PR description.

frontend/server/handlers/artifacts.ts
```diff
@@ -309,47 +322,41 @@ function getGCSArtifactHandler(
...
       if (peek) {
-        matchingFiles[0].createReadStream().pipe(new PreviewStream({ peek })).pipe(res);
+        const stream = await downloadGCSObjectStream({
+          bucket,
+          credentials,
+          objectName: matchingFiles[0],
+        });
+        stream.pipe(new PreviewStream({ peek })).pipe(res);
         return;
       }
```
manaswinidas, line 363: Worth noting: the old `forEach` + `.on('end')` code had a race condition — it sent `res.send(contents)` when the last-indexed file finished, regardless of whether earlier files had completed. The new sequential `for...of` with `await` fixes this. Consider calling it out in the PR description as a correctness improvement.

jeffspahr:
Addressed the open review items in `15f3c0a74`:

- reuse a single GCS auth client per artifact request instead of constructing `GoogleAuth`/`getClient()` for each helper call
- add direct `gcs-helper` unit coverage for auth setup, pagination, filtering, and encoded URL construction
- add wildcard / multi-file GCS integration coverage in `artifact-get.test.ts`
- update the PR description to call out the missing-`return` fix and the old multi-file race fix

Verification run locally:
- `cd frontend/server && npx vitest run gcs-helper.test.ts --reporter=dot`
- `cd frontend/server && npx vitest run integration-tests/artifact-get.test.ts integration-tests/artifact-proxy.test.ts --reporter=dot`
- `cd frontend/server && npm run format:check`

One local-only note: `cd frontend/server && npm run lint` still fails in this checkout because `frontend/server/tslint.json` extends `tslint-config-prettier`, but that package is only declared in the root `frontend` package, not `frontend/server`. I did not change that repo-level lint-tooling mismatch in this PR.

jeffspahr:
Opened the lint follow-up here: #12997. This migrates \ off TSLint, makes root frontend ESLint the only supported lint path, and removes the local/CI mismatch we called out in review.

jeffspahr:
Opened the lint follow-up here: #12997. This migrates frontend/server off TSLint, makes root frontend ESLint the only supported lint path, and removes the local/CI mismatch we called out in review.

droctothorpe:
/approve
/lgtm

Love me a negative line PR.

PR #12987, opened 2026-03-08
feat: fail pipeline runs on image pull failures

### LINKED ISSUE #12973: [feature] Fail pipeline run when component image cannot be pulled
### Feature Area

/area backend

/area components

When a pipeline component cannot pull its container image (for example, the image does not exist in the registry). Currently, the pipeline run does not fail immediately in this situation and may remain in a running state for a long time.

It would be helpful if Kubeflow Pipelines detected this condition and marked the component (and pipeline run) as **failed** when the pod enters a terminal image pull error state.

In practice this situation usually happens due to:

- incorrect image tag
- image not pushed to registry yet
- incorrect repository path
- permission issues to pull the image

When this happens, the pipeline run appears to be running but is actually stuck because the pod cannot start. This can be confusing for users and requires manual inspection of the Kubernetes pod status to understand the root cause.

Failing the step and pipeline earlier would make the failure mode clearer and improve debugging.

There is no built-in mechanism in Kubeflow Pipelines to automatically fail the pipeline in this situation.

---

Love this idea? Give it a 👍.

jsonmp-k8:
## Summary

- Detect `ImagePullBackOff`/`ErrImagePull` on workflow pods and terminate the parent Argo workflow after a configurable grace period, preventing pipeline runs from staying in "Running" state indefinitely when a container image cannot be pulled
- Integrated into the existing `WorkflowSaver.Save()` flow in the persistence agent — no separate watcher goroutine needed
- Feature is opt-in via `IMAGE_PULL_FAILURE_HANDLING_ENABLED` env var (default: `false`, grace period default: 5 minutes)
- Updated RBAC for the persistence agent to allow pod listing and workflow patching

Closes #12973

## Test plan

- [x] Unit tests for `ImagePullFailureChecker` covering: no failure, `ImagePullBackOff`, `ErrImagePull`, init container failures, grace period behavior, mixed healthy/failing pods, workflow-scoped pod listing
- [x] All existing `WorkflowSaver` tests pass unchanged
- [x] All 37 persistence worker tests pass
- [ ] Manual testing: deploy with `IMAGE_PULL_FAILURE_HANDLING_ENABLED=true`, run a pipeline with a nonexistent image, verify the run fails after the grace period


jeffspahr (commented):
Review findings from local inspection of PR #12987.

manifests/kustomize/base/pipeline/ml-pipeline-persistenceagent-role.yaml
```diff
@@ -11,6 +11,14 @@ rules:
   - get
   - list
   - watch
+  - patch
+- apiGroups:
```
jeffspahr, line 15: [P1] Add pod watch RBAC

The new pod SharedInformerFactory starts a watch, but this Role grants only get/list on pods; the multi-user ClusterRole mirrors the same verbs. Once enabled, the reflector will hit RBAC forbidden on its watch and the cache will fall back to retry/relist behavior instead of normal informer updates. Add watch to pods in both RBAC manifests.

backend/src/agent/persistence/main.go
```diff
@@ -139,18 +146,52 @@ func main() {
...
+			util.CurrentExecutionType(), cfg)
+		if err != nil {
+			log.Fatalf("Error building execution client for image pull failure handling: %v", err)
+		}
+
+		// Create a shared informer for pods so that image pull failure checks
+		// read from a local cache instead of making API calls on every check.
+		var podInformerFactory informers.SharedInformerFactory
+		if namespace == "" {
+			podInformerFactory = informers.NewSharedInformerFactory(kubeClient, time.Second*30)
```
jeffspahr: [P2] Filter the pod informer

When namespace is empty in multi-user mode, this creates a cluster-wide informer for every pod even though the checker only needs Argo workflow pods. That can be a large cache and broad watch for an opt-in feature. Use a filtered informer with a label selector for workflows.argoproj.io/workflow in both namespace-scoped and cluster-scoped cases.

backend/src/agent/persistence/main.go
```diff
@@ -21,56 +21,63 @@ import (
...
-	saTokenRefreshIntervalFlagName        = "saTokenRefreshIntervalInSecs"
-	caCertPathFlagName                    = "caCertPath"
+	logLevelFlagName                         = "logLevel"
+	kubeconfigFlagName                       = "kubeconfig"
+	masterFlagName                           = "master"
+	initializationTimeoutFlagName            = "initializeTimeout"
+	timeoutFlagName                          = "timeout"
+	mlPipelineAPIServerBasePathFlagName      = "mlPipelineAPIServerBasePath"
+	mlPipelineAPIServerNameFlagName          = "mlPipelineAPIServerName"
+	mlPipelineAPIServerHTTPPortFlagName      = "mlPipelineServiceHTTPPort"
```
jeffspahr: [P2] Keep the existing HTTP port flag

The registered flag changed from mlPipelineServiceHttpPort to mlPipelineServiceHTTPPort. Go flags are case-sensitive, so existing deployments or scripts that pass the old flag will fail during startup with flag provided but not defined. Rename the Go variable if desired, but keep the flag string stable.

jsonmp-k8:
@hbelmiro can you please review this ?

backend/src/agent/persistence/worker/image_pull_failure_checker.go
```diff
@@ -0,0 +1,157 @@
...
+
+const (
+	// ArgoWorkflowLabelKey is the label Argo sets on pods to identify the parent workflow.
+	ArgoWorkflowLabelKey = "workflows.argoproj.io/workflow"
+)
+
+// ImagePullFailureCheckerInterface checks workflow pods for image pull failures
+// and terminates the workflow if the grace period has elapsed.
+type ImagePullFailureCheckerInterface interface {
+	CheckAndTerminate(namespace string, workflowName string) error
```
hbelmiro: `CheckAndTerminate` does not accept a `context.Context` parameter. The implementation creates `context.TODO()` internally (line 99) for the Patch API call, which means callers cannot propagate cancellation or deadlines through this I/O call chain.

backend/src/agent/persistence/worker/image_pull_failure_checker.go
```diff
@@ -0,0 +1,157 @@
...
+		if podAge < c.gracePeriod {
+			log.Debugf("Pod %s/%s has image pull failure for %q (age: %v), waiting for grace period (%v)",
+				pod.Namespace, pod.Name, failedImage, podAge.Round(time.Second), c.gracePeriod)
+			continue
+		}
+
+		log.Infof("Terminating workflow %s/%s: pod %s has image pull failure for %q (age: %v exceeds grace period %v)",
+			namespace, workflowName, pod.Name, failedImage, podAge.Round(time.Second), c.gracePeriod)
+		return c.terminateWorkflow(context.TODO(), namespace, workflowName)
+	}
```
hbelmiro, line 100: When a workflow is terminated due to image pull failure, the only user-visible failure reason is "exceeded activeDeadlineSeconds" (the Argo workflow condition message). There is no event, annotation, or condition update on the workflow or run that indicates the root cause was an image pull failure. Users looking at the KFP UI will see "Failed" with an `activeDeadlineSeconds` message and may not understand why.

The persistence agent logs the reason (`"Terminating workflow ... image pull failure for ..."`), but those logs are not surfaced in the KFP UI.

backend/src/agent/persistence/worker/image_pull_failure_checker_test.go
```diff
@@ -0,0 +1,378 @@
...
+			},
+		},
+	}
+
+	podLister := newTestPodLister(pod)
+	// Grace period is 5 min, pod is 10 min old -- should attempt to terminate.
+	// executionClient is nil so the termination will return an error.
+	checker := NewImagePullFailureChecker(podLister, nil, 5*time.Minute)
+	err := checker.CheckAndTerminate("default", "my-workflow")
+	assert.Error(t, err)
```
hbelmiro:
```suggestion
	require.Error(t, err)
```

backend/src/agent/persistence/worker/image_pull_failure_checker_test.go
```diff
@@ -0,0 +1,378 @@
...
+	}
+
+	podLister := newTestPodLister(pod)
+	fakeExecInterface := &fakeExecutionInterface{}
+	fakeExecClient := &fakeExecutionClient{executionInterface: fakeExecInterface}
+
+	checker := NewImagePullFailureChecker(podLister, fakeExecClient, 5*time.Minute)
+	err := checker.CheckAndTerminate("default", "my-workflow")
+
+	assert.NoError(t, err)
```
hbelmiro:
```suggestion
	require.NoError(t, err)
```

backend/src/agent/persistence/worker/workflow_saver_test.go
```diff
@@ -218,3 +233,171 @@ func TestWorkflow_Save_SkippedDDueToMissingRunID(t *testing.T) {
...
+	})
+
+	workflowFake.Put("MY_NAMESPACE", "MY_NAME", workflow)
+
+	saver := NewWorkflowSaver(workflowFake, pipelineFake, 100)
+	saver.SetImagePullFailureChecker(checker)
+
+	err := saver.Save("MY_KEY", "MY_NAMESPACE", "MY_NAME", 20)
+
+	assert.NoError(t, err)
```
hbelmiro:
```suggestion
	require.NoError(t, err)
```

backend/src/agent/persistence/worker/workflow_saver_test.go
```diff
@@ -218,3 +233,171 @@ func TestWorkflow_Save_SkippedDDueToMissingRunID(t *testing.T) {
...
```
hbelmiro:
```suggestion
	require.NoError(t, err)
```

backend/src/agent/persistence/worker/workflow_saver_test.go
```diff
@@ -218,3 +233,171 @@ func TestWorkflow_Save_SkippedDDueToMissingRunID(t *testing.T) {
...
```
hbelmiro:
```suggestion
	require.NoError(t, err)
```

backend/src/agent/persistence/worker/workflow_saver_test.go
```diff
@@ -218,3 +233,171 @@ func TestWorkflow_Save_SkippedDDueToMissingRunID(t *testing.T) {
...
```
hbelmiro:
```suggestion
	require.NoError(t, err)
```

backend/src/agent/persistence/worker/workflow_saver_test.go
```diff
@@ -218,3 +233,171 @@ func TestWorkflow_Save_SkippedDDueToMissingRunID(t *testing.T) {
...
+
+	workflowFake.Put("MY_NAMESPACE", "MY_NAME", workflow)
+
+	saver := NewWorkflowSaver(workflowFake, pipelineFake, 100)
+	saver.SetImagePullFailureChecker(checker)
+
+	err := saver.Save("MY_KEY", "MY_NAMESPACE", "MY_NAME", 20)
+
+	// Checker error should be logged but Save should still succeed.
+	assert.NoError(t, err)
```
hbelmiro:
```suggestion
	require.NoError(t, err)
```

jsonmp-k8:
@hbelmiro can you please check? all review comments are addressed.

backend/src/agent/persistence/worker/image_pull_failure_checker.go
```diff
@@ -0,0 +1,174 @@
...
+)
+
+const (
+	// ArgoWorkflowLabelKey is the label Argo sets on pods to identify the parent workflow.
+	ArgoWorkflowLabelKey = "workflows.argoproj.io/workflow"
+)
+
+// ImagePullFailureCheckerInterface checks workflow pods for image pull failures
+// and terminates the workflow if the grace period has elapsed.
+type ImagePullFailureCheckerInterface interface {
```
hbelmiro: The idiomatic Go pattern is to name the interface after the behavior (`ImagePullFailureChecker`) and make the concrete implementation unexported (`imagePullFailureChecker`). Consumers (`WorkflowSaver`) depend on the interface, not the struct, so the struct doesn't need to be exported.

**Suggested fix:** Rename `ImagePullFailureCheckerInterface` to `ImagePullFailureChecker`, rename the current `ImagePullFailureChecker` struct to `imagePullFailureChecker`, and have `NewImagePullFailureChecker` return the interface type:

```go
type ImagePullFailureChecker interface {
    CheckAndTerminate(ctx context.Context, namespace string, workflowName string) error
}

type imagePullFailureChecker struct {
    podLister       corelisters.PodLister
    executionClient util.ExecutionClient
    gracePeriod     time.Duration
}

func NewImagePullFailureChecker(...) ImagePullFailureChecker {
    return &imagePullFailureChecker{...}
}
```

backend/src/common/util/execution_client.go
```diff
@@ -78,6 +80,21 @@ type ExecutionInterface interface {
...
+	switch execType {
+	case ArgoWorkflow:
+		argoProjClient, err := argoclient.NewForConfig(cfg)
+		if err != nil {
+			return nil, fmt.Errorf("failed to create Argo client: %w", err)
+		}
+		return &WorkflowClient{client: argoProjClient}, nil
+	default:
+		return nil, fmt.Errorf("not supported type of Execution: %s", execType)
+	}
```
hbelmiro: We can simplify this with a regular `if` statement.

hbelmiro (commented):
I have more two aesthetic comments that I haven't noticed before. Since the images for the CI already expired and we'll have to run the entire CI again, I think it is worth addressing them.

jsonmp-k8:
@hbelmiro can you please check ?

hbelmiro (commented):
/lgtm

hbelmiro:
The images have expired.
Let's reopen the PR to re-trigger the build.
/close

mr-morj:
Hi @hbelmiro, @chensun,

Just wanted to kindly check the status of this PR.

It looks like the requested changes were addressed, CI is passing, and the PR already has lgtm, but it still seems to be waiting for the remaining approval.

Could you please let us know if there are any blockers or additional changes required? Thanks in advance!

hbelmiro:
> Hi @hbelmiro, @chensun,
> 
> Just wanted to kindly check the status of this PR.
> 
> It looks like the requested changes were addressed, CI is passing, and the PR already has lgtm, but it still seems to be waiting for the remaining approval.
> 
> Could you please let us know if there are any blockers or additional changes required? Thanks in advance!

I can approve only for the backend directory. @juliusvonkohout can you take a look?

Also, @jsonmp-k8 can you please rebase since it's pending for almost a month?

jsonmp-k8:
@hbelmiro I have rebase this PR

jeffspahr:
Review finding from local inspection: [P2] Keep the GCP Marketplace chart in sync for image pull failure handling. The Kustomize manifests now grant the persistence agent workflow patch and pod get/list/watch, and expose IMAGE_PULL_FAILURE_HANDLING_ENABLED / IMAGE_PULL_FAILURE_GRACE_PERIOD_SEC. The GCP Marketplace Helm template still only grants workflow get/list/watch and does not grant pod permissions or expose those envs on the persistence-agent deployment. If Marketplace is still a supported install path, enabling this feature there would either require chart-local edits or fail/hang on RBAC when the pod informer/termination patch starts. Please update manifests/gcp_marketplace/chart/kubeflow-pipelines/templates/pipeline.yaml with the same workflow patch, pod get/list/watch, and feature env wiring/defaults as the Kustomize base, or explicitly confirm that this chart is no longer a supported path for the feature.

juliusvonkohout:
I actually would like to delete gcp-marketplace unless someone else is willing to maintain it. The regular standalone installation is maintained and works on GCP. For everything else there is kubeflow community distribution.

juliusvonkohout:
Proposal do delete it https://github.com/kubeflow/pipelines/pull/13647

PR #12988, opened 2026-03-08
fix(backend): fail run on image pull errors. Fixes #12973

### LINKED ISSUE #12973: [feature] Fail pipeline run when component image cannot be pulled

ahirraoparth1-ops:
Fail run when underlying Argo workflow nodes report terminal image pull errors (ImagePullBackOff / ErrImagePull), so runs do not remain in Running state when component images cannot be pulled.

### DIFF
--- a/backend/src/common/util/workflow.go
+++ b/backend/src/common/util/workflow.go
@@ -459,11 +459,17 @@ func (w *Workflow) FinishedAt() int64 {
 	return w.Status.FinishedAt.Unix()
 }
 
+// imagePullErrorGracePeriod is the duration to wait before treating an
+// image pull error as terminal. ImagePullBackOff is often transient and
+// Kubernetes will retry for up to ~5 minutes before giving up.
+const imagePullErrorGracePeriod = 5 * time.Minute
+
+// Condition returns the execution phase of the workflow. If the workflow is
+// still in a non-final phase but one of its nodes reports a terminal image
+// pull error that has persisted beyond the grace period, surface this as a
+// failed condition to Kubeflow Pipelines so that the corresponding run can
+// be marked as failed instead of remaining in a misleading running state.
 func (w *Workflow) Condition() exec.ExecutionPhase {
-	// If the workflow is still in a non-final phase but one of its nodes
-	// reports a terminal image pull error, surface this as a failed condition
-	// to Kubeflow Pipelines so that the corresponding run can be marked as
-	// failed instead of remaining in a misleading running state.
 	if !w.IsInFinalState() && w.hasImagePullError() {
 		return exec.ExecutionFailed
 	}
@@ -472,12 +478,43 @@ func (w *Workflow) Condition() exec.ExecutionPhase {
 
 // hasImagePullError returns true if any node in the workflow reports a message
 // indicating a terminal container image pull error (for example,
-// ImagePullBackOff or ErrImagePull). This is based on the node status message
-// surfaced by Argo Workflows, which in turn reflects the underlying Pod's
-// container waiting reason.
+// ImagePullBackOff or ErrImagePull) that has persisted beyond the grace period.
+// This is based on the node status message surfaced by Argo Workflows, which
+// in turn reflects the underlying Pod's container waiting reason.
 func (w *Workflow) hasImagePullError() bool {
 	if w.Status.Nodes == nil {
 		return false
+	}
+	now := time.Now()
+	for _, node := range w.Status.Nodes {
+		if node.Message == "" {
+			continue
+		}
+		if !strings.Contains(node.Message, "ImagePullBackOff") &&
+			!strings.Contains(node.Message, "ErrImagePull") {
+			continue
+		}
+		// Check if the error has persisted beyond the grace period.
+		// Use the node's start time to determine how long it has been in this state.
+		if node.StartedAt.IsZero() {
+			// If no start time, treat as immediate failure (shouldn't happen in practice)
+			return true
+		}
+		elapsed := now.Sub(node.StartedAt.Time)
+		if elapsed >= imagePullErrorGracePeriod {
+			return true
+		}
+	}
+	return false
+}
+
+// ImagePullErrorMessage returns the image pull error message from any node
+// that reports ImagePullBackOff or ErrImagePull, or empty string if none found.
+// This is used to surface the error to users when a run fails due to image
+// pull issues.
+func (w *Workflow) ImagePullErrorMessage() string {
+	if w.Status.Nodes == nil {
+		return ""
 	}
 	for _, node := range w.Status.Nodes {
 		if node.Message == "" {
@@ -485,14 +522,20 @@ func (w *Workflow) hasImagePullError() bool {
 		}
 		if strings.Contains(node.Message, "ImagePullBackOff") ||
 			strings.Contains(node.Message, "ErrImagePull") {
-			return true
+			return fmt.Sprintf("Node %q failed: %s", node.DisplayName, node.Message)
 		}
 	}
-	return false
+	return ""
 }
 
+// Message returns the workflow status message. If the workflow has an image
+// pull error but no top-level message, returns the image pull error message
+// to surface the error to users.
 func (w *Workflow) Message() string {
-	return w.Status.Message
+	if w.Status.Message != "" {
+		return w.Status.Message
+	}
+	return w.ImagePullErrorMessage()
 }
 
 func (w *Workflow) FinishedAtTime() metav1.Time {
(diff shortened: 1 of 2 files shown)


ahirraoparth1-ops:
Hi, thanks for the CI setup here!

My PR currently shows `CI Check / check_ci_status` as failing with `Some checks have failed or timed out, please check the workflow run summary to get the details`, but from the UI I only see this aggregator job and I’m not fully sure which underlying check is red or gated.

This change only touches `backend/src/common/util/workflow.go` and its tests to mark runs as failed when Argo nodes hit image pull errors (ImagePullBackOff / ErrImagePull). If there’s a specific failing job I should look at, or if this PR needs an `ok-to-test` label or a CI re-run, could someone please point me in the right direction? I’m happy to fix anything needed.

hbelmiro:
This duplicates https://github.com/kubeflow/pipelines/pull/12987.

jeffspahr (commented):
## PR #12988 Review: Fail run on image pull errors

*Review performed using Claude Opus 4.6*

**Author:** ahirraoparth1-ops | **Files changed:** 2 | **+69 / -10**

### Overview

This PR addresses issue #12973 where runs remain stuck in `Running` state when a component image can't be pulled. It modifies `Workflow.Condition()` to return `ExecutionFailed` when the workflow is still running but a node reports `ImagePullBackOff` or `ErrImagePull`.

### Issues and Concerns

#### 1. Race condition / premature failure (High)

`ImagePullBackOff` is often **transient**. Kubernetes uses exponential backoff and will retry pulling the image for up to ~5 minutes before the pod is truly stuck. This PR treats the *first observation* of `ImagePullBackOff` as a terminal failure, which could fail runs that would otherwise recover.

At minimum, the PR should either:
- Only trigger after the node has been in this state for a configurable duration (e.g., check `node.StartedAt` age), or
- Only match on `ErrImagePull` with a specific substring indicating a permanent error (e.g., `"not found"`, `"unauthorized"`), treating `ImagePullBackOff` as retryable.

#### 2. Inconsistency between `Condition()` and `IsInFinalState()` (High)

After this change, `Condition()` can return `Failed` while `IsInFinalState()` returns `false` (since `IsInFinalState()` checks `w.Status.Phase`, not the overridden condition). This creates a split-brain situation in callers.

In `resource_manager.go:1349-1358`, the run state is set to `Failed` but `FinishedAt` remains `0` (since the workflow hasn't actually finished). This means:
- The run is marked failed but has no finish timestamp
- The workflow is still running in Argo and consuming resources — nobody terminates it
- The persistence agent may keep re-reporting this workflow since it's not in a final state

**The PR should also terminate the underlying Argo workflow** when an image pull error is detected, or the detection should happen at the controller level where the workflow can be actively stopped.

#### 3. `Message()` not updated (Medium)

When `Condition()` returns `Failed` due to image pull errors, `Message()` still returns the workflow-level message (which may be empty). The user won't know *why* the run failed. Consider overriding `Message()` to surface the node-level error message, e.g., `"Node 'step-foo' failed: ImagePullBackOff: ..."`.

#### 4. Incomplete string matching (Low)

The check uses `strings.Contains` on `node.Message`. If Argo changes its message format or the error string appears in a different context (e.g., a log message parameter), this could produce false positives or miss errors. This is a minor concern but worth noting.

#### 5. Missing test: already-completed workflow (Low)

There's no test verifying that a workflow already in `Failed` or `Succeeded` state with an image pull message in a node doesn't get double-processed. The `!w.IsInFinalState()` guard should handle this, but a test would confirm.

### Code Quality

- **Tests are well-structured** — good use of `t.Run()` subtests with descriptive names, and realistic error messages in test data.
- **Comments are clear** and explain the rationale well.
- **The refactoring of existing tests** from flat to subtests is a nice improvement.

### Suggestions

1. **Add a workflow termination step**: When an image pull error is detected, actively stop the Argo workflow to avoid resource leaks.
2. **Add a grace period or retry threshold**: Don't fail on the first `ImagePullBackOff` observation — it's a transient state by design.
3. **Surface the error message**: Override `Message()` to propagate the node-level error to the user.
4. **Add a test for final-state workflows**: Verify that workflows already in `Succeeded`/`Failed` phase are unaffected by image pull messages in their nodes.
5. **Consider `FinishedAt` handling**: When synthetically marking a run as failed, set `FinishedAt` to `time.Now()` so the run has a valid completion timestamp.

### Verdict

The intent is good — stuck runs are a real UX problem. However, the implementation has a meaningful risk of **prematurely failing runs** due to transient `ImagePullBackOff` states, and it creates an **inconsistency between the reported condition and the actual workflow state** that could confuse downstream logic (missing `FinishedAt`, leaked running workflows). These issues should be addressed before merging.

jeffspahr (changes requested):
Two issues from the previous review are still unresolved on the current head:

1. `Workflow.Condition()` can now report `Failed` while `w.Status.Phase` is still `Running` in `backend/src/common/util/workflow.go`. That synthetic state is persisted into the run record in `backend/src/apiserver/resource/resource_manager.go:1349-1358`, and the frontend treats `Failed` as terminal by stopping auto-refresh and enabling Retry in `frontend/src/pages/RunDetails.tsx:771-881`. But the backend retry path still rejects retries unless the real Argo phase is `Failed`/`Error` in `backend/src/common/util/workflow.go:197-203` and `backend/src/apiserver/resource/resource_manager.go:940-944`. This leaves a run looking finished and retryable in the UI while the server still considers it running.

2. The same synthetic-failure path leaves `FinishedAtInSec` at `0` in `backend/src/apiserver/resource/resource_manager.go:1358`, while both API converters serialize that as `time.Unix(0, 0)` in `backend/src/apiserver/server/api_converter.go:1478` and `backend/src/apiserver/server/api_converter.go:1533`. The frontend computes duration for failed runs whenever `finished_at` is non-empty in `frontend/src/lib/Utils.tsx:131-145`, so this can render a bogus 1970 completion time or negative duration for an image-pull failure that Argo has not actually finished yet.

The grace period, error-message surfacing, and added tests address several points from the earlier review, but these two state-model mismatches are still blocking.

Amr-Shams:
> Two issues from the previous review are still unresolved on the current head:
> 
>     1. `Workflow.Condition()` can now report `Failed` while `w.Status.Phase` is still `Running` in `backend/src/common/util/workflow.go`. That synthetic state is persisted into the run record in `backend/src/apiserver/resource/resource_manager.go:1349-1358`, and the frontend treats `Failed` as terminal by stopping auto-refresh and enabling Retry in `frontend/src/pages/RunDetails.tsx:771-881`. But the backend retry path still rejects retries unless the real Argo phase is `Failed`/`Error` in `backend/src/common/util/workflow.go:197-203` and `backend/src/apiserver/resource/resource_manager.go:940-944`. This leaves a run looking finished and retryable in the UI while the server still considers it running.
> 
>     2. The same synthetic-failure path leaves `FinishedAtInSec` at `0` in `backend/src/apiserver/resource/resource_manager.go:1358`, while both API converters serialize that as `time.Unix(0, 0)` in `backend/src/apiserver/server/api_converter.go:1478` and `backend/src/apiserver/server/api_converter.go:1533`. The frontend computes duration for failed runs whenever `finished_at` is non-empty in `frontend/src/lib/Utils.tsx:131-145`, so this can render a bogus 1970 completion time or negative duration for an image-pull failure that Argo has not actually finished yet.
> 
> 
> The grace period, error-message surfacing, and added tests address several points from the earlier review, but these two state-model mismatches are still blocking.

Not the original author here, but since this has been blocked for a while, I took a look.

Maybe one way to handle it: when an image pull failure happens, we proactively call `TerminateWorkflow` on the Argo workflow to set the phase to `Failed` (fixing the `IsInFinalState` mismatch), and set `FinishedAtInSec = time.Now().Unix()` at the same time.

Not 100% sure if this fits the intended approach, but could this help address the remaining block?

droctothorpe:
Closing in favor of https://github.com/kubeflow/pipelines/pull/12987.

Issue #12843, opened 2026-02-17
[feature] Enhance support and visualization for pod lifecycle failure in Kubeflow Pipelines


alyssacgoins:
### Feature Area

/area frontend
/area backend

### What feature would you like to see?

A pipeline in Kubeflow Pipelines (KFP) fails for one of two reasons: the user-defined pipeline script returns an error, or a pod lifecycle failure occurs. The latter failure can be classified according to lifecycle stage; a failure is either at the provisioning level (e.g., ImagePullBackOff or Unschedulable), runtime level (e.g., CrashLoopBackOff or OOMKilled), or node level (e.g., NodeLost or Preempted).

KFP currently provides limited support when a pipeline run hits a pod lifecycle failure. While the Kubernetes CLI allows users to view the status of their pipeline pods in real time on the terminal, the KFP UI currently provides only minimal visualization. When a pod lifecycle failure occurs, the KFP UI displays a pipeline frozen at the current pod – not progressing, finishing or failing. The cause of failure is not displayed. This creates a confusing and frustrating user experience.  KFP intends to function as a Kubernetes abstraction for data scientists and AI engineers, meaning that not all KFP users are experienced with Kubernetes. Requiring use of Kubernetes tools and skills leaks that abstraction. 

The solution here proposes error timeouts and enhanced visualization. It would be helpful to parameterize time limits on the three categories of pod lifecycle failure outlined above, as environment variables on the API server. A default time limit of one hour would be provided, and users could optionally specify time limits per pod status (e.g. ImagePullBackOff). In addition to the time management component, it would enhance user experience to visually log these failures in the UI. 
Pod lifecycle failures should be logged and displayed in the UI in the same way pipeline script failures are displayed, in order to provide users with real-time pod lifecycle updates.

### What is the use case or pain point?

Users executing pipeline runs with pod lifecycle failures are unable to view pipeline failure in the UI, and unable to manage failure timeout.

### Is there a workaround currently?
The current workaround to lifecycle failure timeout is to manually delete a pipeline run. The current workaround to the limited UI visualization is `kubectl get pods -w` using the Kubernetes CLI in the terminal. There is no workaround to failure visualization directly within the UI.

---

Love this idea? Give it a 👍.

CREVIOS:
Hi @alyssacgoins, I'm interested in this as a potential GSoC project.

[...]

PR #12989, opened 2026-03-08
feat: add environment variables for pod lifecycle timeouts (#12843)

Ashish241:
Part of #12843

## Summary
Adds the configuration layer for pod lifecycle failure timeouts on the
API server. This is the first small PR in the series — it defines the
environment variables and default values without any pod-killing logic.

## New environment variables (default: `3600s` / 1 hour)

| Variable | Failure stage | Example causes |
|---|---|---|
| `DEFAULT_POD_PROVISIONING_TIMEOUT` | Provisioning | `ImagePullBackOff`, `Unschedulable` |
| `DEFAULT_POD_RUNTIME_TIMEOUT` | Runtime | `CrashLoopBackOff`, `OOMKilled` |
| `DEFAULT_POD_NODE_FAILURE_TIMEOUT` | Node-level | `NodeLost`, `Preempted` |

Operators can override any of these by setting the env var on the
API server deployment.

## Files changed
- [backend/src/apiserver/common/const.go](cci:7://file:///c:/My%20Projects/kubeflow-pipelines%20-%20Copy1/backend/src/apiserver/common/const.go:0:0-0:0) — default value constants
- [backend/src/apiserver/common/config.go](cci:7://file:///c:/My%20Projects/kubeflow-pipelines%20-%20Copy1/backend/src/apiserver/common/config.go:0:0-0:0) — env var name constants + getter functions
- [backend/src/apiserver/config/config.json](cci:7://file:///c:/My%20Projects/kubeflow-pipelines%20-%20Copy1/backend/src/apiserver/config/config.json:0:0-0:0) — config file defaults
- [manifests/kustomize/base/pipeline/ml-pipeline-apiserver-deployment.yaml](cci:7://file:///c:/My%20Projects/kubeflow-pipelines%20-%20Copy1/manifests/kustomize/base/pipeline/ml-pipeline-apiserver-deployment.yaml:0:0-0:0) — container env vars
- [backend/src/apiserver/common/config_test.go](cci:7://file:///c:/My%20Projects/kubeflow-pipelines%20-%20Copy1/backend/src/apiserver/common/config_test.go:0:0-0:0) — unit tests (default + override for each getter)

## Testing

go  test ./backend/src/apiserver/common/... -v -run "TestConfigWrapper"


backend/src/apiserver/common/config.go
```diff
@@ -187,3 +190,22 @@ func GetDefaultSecurityContextRunAsGroup() string {
 func GetDefaultSecurityContextRunAsNonRoot() string {
 	return GetStringConfigWithDefault(DefaultSecurityContextRunAsNonRoot, "")
 }
+
+func GetDurationConfigWithDefault(configName string, value time.Duration) time.Duration {
+	if !viper.IsSet(configName) {
+		return value
+	}
+	return viper.GetDuration(configName)
```
jeffspahr: `viper.AllowEmptyEnv(true)` makes blank env vars count as set, and `viper.GetDuration("")` resolves to `0s`. That means `DEFAULT_POD_*_TIMEOUT=""` bypasses the documented one-hour fallback instead of using it. Consider treating an empty value as unset here:

```suggestion
func GetDurationConfigWithDefault(configName string, value time.Duration) time.Duration {
	if !viper.IsSet(configName) {
		return value
	}
	if viper.GetString(configName) == "" {
		return value
	}
	return viper.GetDuration(configName)
}
```

manifests/kustomize/base/pipeline/ml-pipeline-apiserver-deployment.yaml
```diff
@@ -150,6 +150,24 @@ spec:
...
+              valueFrom:
+                configMapKeyRef:
+                  name: pipeline-install-config
+                  key: defaultPodRuntimeTimeout
+                  optional: true
+            - name: DEFAULT_POD_NODE_FAILURE_TIMEOUT
+              valueFrom:
+                configMapKeyRef:
+                  name: pipeline-install-config
+                  key: defaultPodNodeFailureTimeout
```
jeffspahr, line 169: This wires the timeout env vars into the Deployment, but the matching `pipeline-install-config` keys still need to be added to the shipped install ConfigMaps so stock installs can actually discover and override them. Please add the same defaults/comments to:
- `manifests/kustomize/base/installs/generic/pipeline-install-config.yaml`
- `manifests/kustomize/base/installs/generic/postgres/pipeline-install-config.yaml`

Suggested snippet:

```yaml
  ## Default timeout for pods stuck in provisioning-related failure states.
  defaultPodProvisioningTimeout: "3600s"
  ## Default timeout for pods stuck in runtime-related failure states.
  defaultPodRuntimeTimeout: "3600s"
  ## Default timeout for pods stuck in node-failure-related states.
  defaultPodNodeFailureTimeout: "3600s"
```

jeffspahr (changes requested):
I found two issues to address before this lands:

- Empty timeout values currently resolve to `0s` instead of falling back to the documented `3600s` default.
- The new timeout env vars are wired into the Deployment, but the matching keys are not added to the shipped install ConfigMaps yet, so stock installs do not expose the new knobs.

jeffspahr:
## Code Review: Pod Lifecycle Timeout Configuration

### Overview

Adds three new environment variables (`DEFAULT_POD_PROVISIONING_TIMEOUT`, `DEFAULT_POD_RUNTIME_TIMEOUT`, `DEFAULT_POD_NODE_FAILURE_TIMEOUT`) with 1-hour defaults. Config-only — no pod-killing logic yet. Part of #12843. Tests are well-structured (table-driven, covering both default and override paths) and the two generic `pipeline-install-config.yaml` files are kept in sync.

---

### Blocking Issues

#### P1: PostgreSQL installs never receive the new timeout env vars

`manifests/kustomize/base/postgresql/pipeline/ml-pipeline-apiserver-deployment-patch.yaml` replaces the entire `env` array via strategic merge patch, so the three timeout variables added to the base `ml-pipeline-apiserver-deployment.yaml` are absent from Postgres-based installs. Verified with `kubectl kustomize`: the generic install includes all three `DEFAULT_POD_*_TIMEOUT` entries, postgres includes none. The feature is silently unconfigurable for the Postgres install path.

**Fix:** Add the same three env var entries to the Postgres deployment patch.

#### P2: Invalid timeout strings silently become 0s

`GetDurationConfigWithDefault` only handles unset and empty values before delegating to `viper.GetDuration(...)`. Viper returns `0s` for unparseable strings (e.g., `"bogus"`, `"1hour"`, `"3600"` without a unit) instead of an error. A typo in these operator-facing settings would silently result in a zero-duration timeout, which in the follow-up enforcement PR would mean immediately failing pods.

Verified locally: both `"bogus"` and `"1hour"` resolve to `0s`.

**Fix:** Validate the parsed result — if the raw string is non-empty but the parsed duration is `<= 0`, log a warning and return the default. This is consistent with the project's existing stance in `GetDurationConfig` which calls `glog.Fatalf` on missing values rather than accepting silent misconfiguration.

#### P3: No test for invalid duration input

The `GetDurationConfigWithDefault` test table covers unset, empty, and valid (`"30m"`) but doesn't cover a malformed string. This is the most important edge case given P2.

**Fix:** Add a test case with an invalid duration string (e.g., `"notaduration"`) that asserts the default is returned (once P2's validation is added).

---

### Non-Blocking Issues

#### `const.go` — Noisy whitespace-only diff

The entire `const.go` shows as deleted and re-added (+99 / -90). The real change is adding `import "time"` and three constants at the bottom (~5 lines). This appears to be a line-ending or trailing-whitespace normalization. It makes the PR hard to review and pollutes `git blame`. Please clean this up so only the actual changes appear in the diff.

#### `const.go` — Missing trailing newline

The file ends with `\No newline at end of file`.

#### Viper key normalization

The ConfigMap keys use camelCase (`defaultPodProvisioningTimeout`) while the env var constants use SCREAMING_SNAKE_CASE (`DEFAULT_POD_PROVISIONING_TIMEOUT`). This follows the existing pattern, but it's worth confirming that viper's `AutomaticEnv()` key normalization correctly maps between the two naming conventions. If not, the ConfigMap value could be silently ignored even when set.

---

### Minor Observations

- The three timeout defaults are all 1 hour. The failure modes described (provisioning vs runtime vs node failure) have very different characteristics — node failures are typically detected quickly while provisioning issues can legitimately take longer. Consider whether differentiated defaults would be more appropriate.
- Good use of `t.Setenv` + `viper.Reset()` in tests.

Ashish241:
Thanks for the detailed review! I've updated the PR to address all the feedback:

### Blocking Issues Resolved:

**PostgreSQL deployment missing env vars (P1):** Added the three `DEFAULT_POD_*_TIMEOUT` entries to the PostgreSQL deployment patch (`ml-pipeline-apiserver-deployment-patch.yaml`), mirroring the base deployment.

**Invalid duration silently becoming 0s (P2):** Updated `GetDurationConfigWithDefault` in `config.go` to explicitly validate the parsed duration. If Viper fails to parse a non-empty string and returns `0s`, we now emit a glog.Warningf and gracefully fall back to the provided default duration instead.

**Missing test coverage (P3):** Added a new test case (returns default when env var is an invalid duration string) to `TestGetDurationConfigWithDefault` with a malformed input "notaduration" to explicitly verify the new validation logic falls back correctly to the 1-hour default. 

### Non-Blocking Issues Addressed:

**Noisy `const.go` diff:** The noisy +99/-90 whitespace diff was caused by a mismatch in line-ending formats between the branch's blob (CRLF) and upstream (LF). I have re-normalized `const.go` to LF line endings and ensured the missing trailing newline at EOF is present. The diff against upstream is now perfectly clean (showing just the 8 new lines added).

**Viper Key Normalization Query:** Regarding the naming conventions: Kubernetes naturally handles the projection from the ConfigMap's camelCase key into the container's `SCREAMING_SNAKE_CASE` environment variable at the deployment level. By the time Viper runs in the binary, it just interacts directly with the `SCREAMING_SNAKE_CASE `environment variable using an exact string match. This exactly mirrors existing patterns like `AUTO_UPDATE_PIPELINE_DEFAULT_VERSION` and `DBCONFIG_CONMAXLIFETIME.`

backend/src/apiserver/common/config.go
```diff
@@ -187,3 +190,31 @@ func GetDefaultSecurityContextRunAsGroup() string {
...
+	raw := viper.GetString(configName)
+	if raw == "" {
+		return value
+	}
+	parsedDuration := viper.GetDuration(configName)
+	if parsedDuration <= 0 {
+		glog.Warningf("Config %s has invalid duration value %q, using default %v", configName, raw, value)
+		return value
+	}
+	return parsedDuration
```
jeffspahr: This still lets unitless numeric strings through because `viper.GetDuration("3600")` becomes `3.6µs`. Parsing the raw string explicitly here would reject missing or unknown units and preserve the documented 1h fallback.

```suggestion
	raw := viper.GetString(configName)
	if raw == "" {
		return value
	}
	parsedDuration, err := time.ParseDuration(raw)
	if err != nil || parsedDuration <= 0 {
		glog.Warningf("Config %s has invalid duration value %q, using default %v", configName, raw, value)
		return value
	}
	return parsedDuration
```

jeffspahr (commented):
Re-reviewing the latest diff: the previous PostgreSQL manifest gap is fixed, and both rendered generic and PostgreSQL installs now carry the three DEFAULT_POD_*_TIMEOUT env vars. One correctness issue remains: GetDurationConfigWithDefault still accepts unitless numeric strings as tiny positive durations (for example, "3600" -> 3.6µs), so a missing unit bypasses the intended 1h fallback. I re-ran `go test ./backend/src/apiserver/common/... -run 'Test(GetDurationConfigWithDefault|ConfigWrapperDefaults|ConfigWrapperCustomValues)' -count=1`, and it passed.

backend/src/apiserver/common/config.go
Ashish241: Good catch, `viper.GetDuration("3600")` silently becomes `3.6µs` instead of falling back to the default. Switching to `time.ParseDuration` is the right fix. I'll apply your suggestion and add a test case for a bare numeric string to lock this in.

Ashish241:
Replaced `viper.GetDuration` with `time.ParseDuration` in GetDurationConfigWithDefault so that unitless numeric strings (e.g., '3600') are rejected instead of silently being interpreted as nanoseconds (`3.6µs`). This preserves the documented default fallback for misconfigured values.

jeffspahr (commented):
Current head looks behaviorally clean after the follow-up fixes. The remaining item I found is a targeted test gap for the exact unitless-duration regression that motivated the `time.ParseDuration` change.

backend/src/apiserver/common/config_test.go
```diff
@@ -295,6 +295,59 @@ func TestGetDurationConfig_WhenSet(t *testing.T) {
...
+			defaultValue:  time.Hour,
+			expectedValue: 30 * time.Minute,
+		},
+		{
+			name:          "returns default when env var is an invalid duration string",
+			setEnv:        true,
+			envValue:      "notaduration",
+			defaultValue:  time.Hour,
+			expectedValue: time.Hour,
+		},
```
jeffspahr, line 339: The parser fix in `config.go` now correctly rejects unitless numeric strings, but this regression table still only covers a generic invalid token (`notaduration`). The exact bug here was bare numeric input like `"3600"`, which `viper.GetDuration(...)` previously interpreted as `3.6µs` instead of falling back to the default. Adding one explicit unitless case will lock in the path this PR just fixed.
```suggestion
		{
			name:          "returns default when env var is a unitless numeric string",
			setEnv:        true,
			envValue:      "3600",
			defaultValue:  time.Hour,
			expectedValue: time.Hour,
		},
		{
			name:          "returns default when env var is an invalid duration string",
			setEnv:        true,
			envValue:      "notaduration",
			defaultValue:  time.Hour,
			expectedValue: time.Hour,
		},
```

jeffspahr:
Re-review on the current PR head `1b60e3a70d8a0efdb9f377204385166fa3fcd89e` from March 27, 2026: I’m not seeing any remaining findings.

The only delta since the last reviewed head `3429515e8091245d7802ca2be3315758d15f529c` is the expected seven-line addition in `backend/src/apiserver/common/config_test.go`, adding the explicit `"3600"` unitless-duration regression case that was previously missing. I diffed the new head against that prior reviewed commit, reran `go test ./backend/src/apiserver/common/...` in an isolated worktree, and it passed.

PR checks still show `DCO`, `check_ci_status`, and `ok-to-test` passing; only `tide` is still pending approval/labels. Residual risk is low and limited to the broader timeout-enforcement logic, which this PR still intentionally does not implement.

Ashish241:
@jeffspahr, thanks for the initial approval! I pushed a few minor changes to improve [tests for zero and negative duration config]. Could you take another quick look and re-apply the /lgtm when you have a chance?

jeffspahr:
Looks good @Ashish241.  Could you resolve all the inline comments?

/lgtm

Ashish241:
> Looks good @Ashish241. Could you resolve all the inline comments?
> 
> /lgtm

Yes! I've resolved all the inline comments

droctothorpe:
@Ashish241: Hey, thank you for your PR! As you are not a member of Kubeflow organization, your PR must link to an issue with 'ready' label. Please read our [CONTRIBUTING.md](https://github.com/kubeflow/pipelines/blob/master/CONTRIBUTING.md), closing this PR

PR #12991, opened 2026-03-09
fix(sdk): replace lstrip/rstrip with removeprefix/removesuffix for type name parsing

### LINKED ISSUE #12990: [sdk] Wrong string method used to remove type prefixes, lstrip instead of removeprefix
### Environment

*  KFP version: All versions

*  KFP SDK version: All versions

### Steps to reproduce

In Python, `lstrip("abc")` does not remove the prefix `"abc"`. It removes any combination of the individual characters `a`, `b`, `c` from the start. The correct method is `removeprefix("abc")`.

Two places in the SDK use `lstrip` when they should use `removeprefix`:

1. sdk/python/kfp/dsl/types/type_annotations.py, line 225:
           `return annotation.lstrip('Optional[').rstrip(']')`

      This breaks for common types:

   `'Optional[int]'.lstrip('Optional[').rstrip(']')` -   returns empty string, expected 'int'
   `'Optional[int]'.removeprefix('Optional[').removesuffix(']')` -   returns 'int' 
            
2. sdk/python/kfp/dsl/types/type_utils.py, line 529:
               `type_name = type_name.lstrip('typing.')`

This works by luck for current inputs (List, Dict) but is incorrect in principle and could break for other type names.

### Expected result

Use `removeprefix()` / `removesuffix()` instead (available since Python 3.9, which is the project's minimum version):
 

- type_annotations.py:225

`return annotation.removeprefix('Optional[').removesuffix(']')`

- type_utils.py:529

`type_name = type_name.removeprefix('typing.')`

### Materials and Reference

- Python docs explaining the difference:
https://docs.python.org/3/library/stdtypes.html#str.removeprefix

---

Priyanshu-u07:
Two functions in the SDK incorrectly use `lstrip`/`rstrip` to remove prefixes/suffixes. `lstrip("abc")` strips individual characters from the set {a, b, c}, not the substring "abc". This causes incorrect results for certain type names (e.g., `Optional[int]` returns empty string instead of `int`).
Replaced with `removeprefix`/`removesuffix` (available since Python 3.9, which is the project's minimum version).

Fixes #12990

### DIFF
--- a/frontend/package-lock.json
+++ b/frontend/package-lock.json
@@ -1,7 +1,7 @@
 {
   "name": "pipelines-frontend",
   "version": "0.1.0",
-  "lockfileVersion": 3,
+  "lockfileVersion": 2,
   "requires": true,
   "packages": {
     "": {
@@ -14,6 +14,7 @@
         "@emotion/styled": "^11.14.1",
         "@mui/icons-material": "^5.18.0",
         "@mui/material": "^5.18.0",
+        "@tanstack/react-query": "^4.36.0",
         "@types/lodash.groupby": "^4.6.6",
         "@types/pako": "^1.0.3",
         "ace-builds": "^1.4.14",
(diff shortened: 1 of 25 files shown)


droctothorpe:
/approve
/lgtm

Thank you for the contribution, @Priyanshu-u07! 🎉

PR #12995, opened 2026-03-09
fix(frontend): remove tautological and duplicate UI tests

### LINKED ISSUE #12994: chore(frontend): fix tautological tests and reduce DRY violations in UI test suite
## Summary

Several frontend test files contain tautological tests (tests that always pass regardless of component behavior) and DRY violations (identical test logic repeated without parameterization).

## Problems

1. **Tautological spy-on-self test** (`Css.test.tsx`, `mlmd/Css.test.tsx`): spies on `padding()`, calls it, then asserts the spy was called — proves nothing about the function's behavior.
2. **Assertion-less test** (`MD2Tabs.test.tsx`): "does not try to call the onSwitch handler" has zero `expect()` calls.
3. **Duplicate snapshot tests** (`Status.test.tsx`): 10 individual phase tests (`handles ERROR phase`, etc.) are exact duplicates of the `Object.keys(NodePhase).forEach(...)` loop in the same file.
4. **Repetitive test cases** (`MinioArtifactPreview.test.tsx`, `ExecutionNode.test.tsx`, `ROCCurveHelper.test.tsx`): identical render+assert patterns repeated 5-7 times where `it.each` would suffice.

## Proposed fix

- Replace tautological tests with meaningful assertions
- Add explicit assertion to the no-op test
- Remove the 10 duplicate snapshot tests (and their orphaned snapshots)
- Consolidate repetitive tests using `it.each`

Related: #12959

manaswinidas:
Replace spy-on-self tests in Css.test.tsx and mlmd/Css.test.tsx with meaningful assertions on the padding() return value. Add an explicit assertion to the no-op MD2Tabs click test. Remove 10 duplicate statusToIcon snapshot tests already covered by the forEach loop. Consolidate repetitive test cases in MinioArtifactPreview, ExecutionNode, and ROCCurveHelper using it.each.

Fixes #12994 

**Checklist:**

### DIFF
--- a/frontend/src/Css.test.tsx
+++ b/frontend/src/Css.test.tsx
@@ -53,7 +53,7 @@ describe('Css', () => {
       });
     });
 
-    it('delegates to _paddingInternal and returns a CSS class name', () => {
+    it('returns a non-empty CSS class name', () => {
       const result = Css.padding(100, 'lr');
       expect(typeof result).toBe('string');
       expect(result.length).toBeGreaterThan(0);
--- a/frontend/src/components/MinioArtifactPreview.test.tsx
+++ b/frontend/src/components/MinioArtifactPreview.test.tsx
@@ -53,7 +53,7 @@ describe('MinioArtifactPreview', () => {
     ],
   ])('renders nothing for invalid artifact: %s', (_label, value) => {
     const { container } = render(<MinioArtifactPreview value={value} />);
-    expect(container.innerHTML).toBe('');
+    expect(container).toBeEmptyDOMElement();
   });
 
   it('handles string value', () => {
--- a/frontend/src/components/viewers/ROCCurveHelper.test.tsx
+++ b/frontend/src/components/viewers/ROCCurveHelper.test.tsx
@@ -38,11 +38,7 @@ describe('validateConfidenceMetrics', () => {
     ['array containing null', [null], true],
   ] as const)('validates %s', (_label, input, expectError) => {
     const result = validateConfidenceMetrics(input);
-    if (expectError) {
-      expect(result.error).toBeDefined();
-    } else {
-      expect(result.error).toBeUndefined();
-    }
+    expect(result.error).toEqual(expectError ? expect.anything() : undefined);
   });
 });
 
--- a/frontend/src/mlmd/Css.test.tsx
+++ b/frontend/src/mlmd/Css.test.tsx
@@ -69,7 +69,7 @@ describe('Css', () => {
       });
     });
 
-    it('delegates to _paddingInternal and returns a CSS class name', () => {
+    it('returns a non-empty CSS class name', () => {
       const result = Css.padding(100, 'lr');
       expect(typeof result).toBe('string');
       expect(result.length).toBeGreaterThan(0);


jeffspahr:
## Code Review

**+76 / -465** | 8 files | Test-only changes

### Overview

Cleans up frontend tests by:
1. Replacing spy-on-self tests in `Css.test.tsx` / `mlmd/Css.test.tsx` with actual return-value assertions
2. Adding an explicit `not.toThrow()` assertion to the MD2Tabs no-op click test
3. Removing 10 duplicate `statusToIcon` snapshot tests already covered by the `Object.keys(NodePhase).forEach` loop
4. Consolidating repetitive test cases in `MinioArtifactPreview`, `ExecutionNode`, and `ROCCurveHelper` using `it.each`

### Positives

- The spy-on-self tests were genuinely tautological (spying on `Css.padding` then calling it and asserting it was called — always true). Good removal.
- The 10 individual `statusToIcon` phase tests are exact duplicates of what the `forEach` loop already covers. Clean removal with 293 lines of dead snapshots deleted.
- The `it.each` consolidations are well-structured and preserve the same assertions.
- The MD2Tabs `expect(() => ...).not.toThrow()` is a genuine improvement over the implicit "didn't throw" pattern.
- Net -389 lines with no loss of meaningful coverage.

### Suggestions

1. **Css.test.tsx replacement is weak** — The new test asserts `typeof result === 'string'` and `result.length > 0`. This is marginally better than the tautological spy but still doesn't test much. A more meaningful assertion would check that the returned class name contains the expected padding value (e.g., `expect(result).toContain('100')`). Minor — not blocking.

2. **MinioArtifactPreview assertion change** — The original tests used `toMatchInlineSnapshot('<div />')` but the replacement uses `expect(container.innerHTML).toBe('')`. These are semantically different: `<div />` (the container div with nothing inside) vs empty string. Worth verifying this actually passes — `container.innerHTML` for an empty React render is typically `""` but it depends on test setup. If the component renders a wrapper div, this could be a false green or false red.

3. **ROCCurveHelper `if/else` in `it.each`** — The conditional pattern inside `it.each` is acceptable but could be slightly cleaner as a single assertion: `expect(result.error).toEqual(expectError ? expect.anything() : undefined)`. Stylistic preference, not blocking.

### Risk Assessment

**Low risk.** All changes are test-only. No production code modified. The removed tests were provably redundant or tautological. The `it.each` consolidations preserve equivalent assertions.

### Verdict

Looks good overall. Main thing to double-check is the `MinioArtifactPreview` assertion change (`container.innerHTML === ''` vs the old inline snapshot `<div />`) to ensure it accurately reflects what the component renders for invalid inputs.

jeffspahr:
**Review follow-up — checking prior feedback:**

1. **Css.test.tsx replacement is weak** — suggested `expect(result).toContain('100')`. Still uses the weaker `typeof` + `length > 0` check. Non-blocking per original review.

2. **MinioArtifactPreview assertion change** — flagged `container.innerHTML === ''` vs old inline snapshot. Now uses `toBeEmptyDOMElement()` which is the correct jest-dom matcher. **Addressed.**

3. **ROCCurveHelper `if/else` in `it.each`** — suggested `expect(result.error).toEqual(expectError ? expect.anything() : undefined)`. Current diff uses exactly that. **Addressed.**

2/3 suggestions incorporated. Remaining item is non-blocking. Looks good to approve once CI passes.

PR #12997, opened 2026-03-09
chore(frontend): migrate server lint to eslint and fix artifact domain matching

jeffspahr:
## Why
`frontend/server` was still carrying a standalone TSLint path even though TSLint has been officially deprecated since February 23, 2019 in `palantir/tslint#4534`, and the upstream TSLint repository was archived on March 25, 2021.

That left us with two problems:
- we were still depending on deprecated, unmaintained lint tooling in one part of the frontend
- local `frontend/server` linting and CI were checking different things, because CI only enforced the root frontend ESLint path

This PR removes that drift. It makes ESLint the single supported lint path for frontend code, including `frontend/server`, so contributors and CI use the same maintained tooling and get the same result.

While doing that migration, it also fixes a small artifact-domain matching bug in `frontend/server/handlers/domain-checker.ts`: `domain_from_url` was returning the full URL prefix instead of the extracted hostname. That meant host-only `ALLOWED_ARTIFACT_DOMAIN_REGEX` patterns such as `^example\.com$` could reject otherwise valid artifact URLs.

## Scope
This PR is intentionally limited to:
- migrating `frontend/server` off TSLint onto the shared frontend ESLint path
- correcting artifact allowlist host extraction so `ALLOWED_ARTIFACT_DOMAIN_REGEX` is matched against the hostname rather than the full URL prefix

It is not stacked on top of [#12976](https://github.com/kubeflow/pipelines/pull/12976), and it avoids touching the artifact-handler source changes from that PR.

The only shared files are `frontend/server/package.json` and `frontend/server/package-lock.json`, because removing TSLint from the standalone server package necessarily updates the server package manifests.

## Summary
- migrate `frontend/server` off TSLint and remove `frontend/server/tslint.json`
- make root frontend linting the single source of truth by adding `lint:ui` and `lint:server`
- have `frontend/server` delegate its `lint` script to the root frontend lint path
- add TypeScript import resolution for server ESM `.js` specifiers and ignore generated server code
- fix the small set of existing server lint violations exposed by the shared ESLint config
- fix artifact hostname extraction so host-based `ALLOWED_ARTIFACT_DOMAIN_REGEX` patterns match the extracted domain instead of the full URL prefix
- add regression coverage for host-only allowlists with `http://`, ports, malformed inputs, subdomains, and the production default allowlist pattern

## Verification
- `cd frontend && npm run lint:server`
- `cd frontend/server && npm run lint`
- `cd frontend && npm run format:check`
- `cd frontend/server && npm test -- domain-checker.test.ts`
- `cd frontend && npm run test:server:coverage`
- `cd frontend && npm run build`
- `cd frontend && npm run test:ci`
  - this still hits two pre-existing UI test timeouts unrelated to this diff:
    - `src/pages/ExperimentDetails.test.tsx > ExperimentDetails > switches to another tab will change Archive/Restore button`
    - `src/pages/RunDetailsV2.test.tsx > RunDetailsV2 > topbar tabs > shows Execution Sidepanel`

### DIFF
--- a/frontend/package.json
+++ b/frontend/package.json
@@ -159,7 +159,6 @@
     "ts-node": "^10.9.2",
     "ts-node-dev": "^2.0.0",
     "tsconfig-paths": "^3.10.1",
-    "tslint-config-prettier": "^1.18.0",
     "typescript": "^4.9.5",
     "vite": "^7.3.1",
     "vitest": "^4.0.17",
--- a/frontend/mock-backend/tslint.json
+++ b/frontend/mock-backend/tslint.json
@@ -1,8 +1,0 @@
-{
-  "extends": ["../tslint.json"],
-  "rules": {
-    "no-console": false,
-    "no-default-export": false,
-    "no-var-requires": false
-  }
-}
--- a/frontend/package-lock.json
+++ b/frontend/package-lock.json
@@ -102,7 +102,6 @@
         "ts-node": "^10.9.2",
         "ts-node-dev": "^2.0.0",
         "tsconfig-paths": "^3.10.1",
-        "tslint-config-prettier": "^1.18.0",
         "typescript": "^4.9.5",
         "vite": "^7.3.1",
         "vitest": "^4.0.17",
@@ -11511,17 +11510,6 @@
       "version": "1.14.1",
       "dev": true,
       "license": "0BSD"
-    },
-    "node_modules/tslint-config-prettier": {
-      "version": "1.18.0",
-      "dev": true,
-      "license": "MIT",
-      "bin": {
-        "tslint-config-prettier-check": "bin/check.js"
-      },
-      "engines": {
-        "node": ">=4.0.0"
-      }
     },
     "node_modules/tsutils": {
       "version": "3.21.0",


frontend/server/handlers/domain-checker.ts
```diff
@@ -23,10 +23,6 @@ export function isAllowedDomain(urlStr: string, allowedDomain: string): boolean
 }
 
 function domain_from_url(url: string): string {
-  let result: string = '';
-  let match = url.match(/^(?:https?:\/\/)?(?:[^@\/\n]+@)?([^:\/?\n]+)/);
-  if (match) {
-    result = match[0];
-  }
-  return result;
+  const match = url.match(/^(?:https?:\/\/)?(?:[^@/\n]+@)?([^:/?\n]+)/);
```
manaswinidas, line 26: This changes `domain_from_url` from returning the full match (including protocol prefix, e.g. `https://example.com`) to just the captured hostname (`example.com`). That's the correct behavior, but it's a **behavioral fix**, not just a lint cleanup.

Any deployment that customized `ALLOWED_ARTIFACT_DOMAIN_REGEX` to match the protocol prefix would silently start rejecting valid requests. Worth calling out explicitly in the PR description (and ideally as a separate commit) so reviewers and operators understand the risk.

frontend/server/handlers/domain-checker.test.ts
```diff
@@ -0,0 +1,30 @@
...
+  it('matches a host-only allowlist when the URL contains user info', () => {
+    expect(isAllowedDomain('https://user@example.com/path/to/artifact', '^example\\.com$')).toBe(
+      true,
+    );
+  });
+
+  it('rejects a URL whose host does not match the allowlist', () => {
+    expect(isAllowedDomain('https://evil.example/path/to/artifact', '^example\\.com$')).toBe(false);
+  });
+});
```
manaswinidas, line 50: Coverage is thin for a security-sensitive function that gates artifact access. The three tests all use the same allowlist pattern and miss important production shapes:

- URL with port (`https://example.com:8080/path`) -- `getArtifactServiceGetter` produces `http://service.ns:port`
- HTTP URLs (the proxy path uses `http://`)
- Empty/malformed inputs (`domain_from_url` returns `''`)
- Subdomain matching (`sub.example.com` vs `^example\.com$`)
- The actual production default pattern (`^.*$`)

frontend/package.json
```diff
@@ -137,6 +139,7 @@
     "autoprefixer": "^10.4.1",
     "browserslist": "4.16.5",
     "eslint": "^8.57.1",
+    "eslint-import-resolver-typescript": "3.5.5",
```
manaswinidas: nit: Exact pin `3.5.5` while all other eslint plugins use caret ranges (`^2.32.0`, etc.). Intentional (known breakage in newer versions) or should this be `^3.5.5`?

frontend/.eslintrc.yaml
```diff
@@ -69,3 +76,14 @@ rules:
     react/react-in-jsx-scope: off
     react/prop-types: off
     react/jsx-no-target-blank: ['error', { "allowReferrer": true }]
+overrides:
+    - files:
+          - server/handlers/artifacts.ts
```
manaswinidas: nit: Permanent `import/no-duplicates: off` suppression -- is there a plan to fix the duplicate imports in `artifacts.ts`, or is this intentionally permanent? A brief comment or linked issue would help future contributors.

frontend/.eslintrc.yaml
```diff
@@ -69,3 +76,14 @@ rules:
...
+      rules:
+          import/no-duplicates: off
+          '@typescript-eslint/no-unused-vars':
```
manaswinidas: nit: `^req$` is scoped only to `artifacts.ts`, but unused `req` params are a common Express handler pattern. If other server handlers hit the same issue, consider broadening this override to `server/**/*.ts`.

frontend/server/package.json
```diff
@@ -53,7 +52,7 @@
     "build": "tsc --project .",
     "format": "prettier --write './**/*.{ts,tsx}'",
     "format:check": "prettier --check './**/*.{ts,tsx}' || node ../scripts/check-format-error-info.js",
-    "lint": "tslint -c tslint.json -p tsconfig.json",
+    "lint": "cd .. && npm run lint:server",
```
manaswinidas: nit: `cd ..` assumes server is always one directory below the frontend root. `npm --prefix .. run lint:server` would be slightly more resilient to layout changes.

frontend/.eslintrc.yaml
jeffspahr: Instead of broadening the `^req$` exception, I removed the need for it in `artifacts.ts` by renaming the intentionally unused handler args to `_req` / `_res`. The file-specific override is now gone.

frontend/server/package.json
jeffspahr: Switched the server package lint delegation to `npm --prefix .. run lint:server`.

frontend/package.json
jeffspahr: Changed this to `^3.5.5` in `frontend/package.json` and the corresponding root lockfile entry for consistency with the other ESLint packages.

frontend/server/handlers/domain-checker.test.ts
jeffspahr: Expanded `frontend/server/handlers/domain-checker.test.ts` to cover `http://` URLs, port-bearing URLs, malformed input, exact-host vs subdomain matching, and the default `^.*$` allowlist.

frontend/.eslintrc.yaml
jeffspahr: Removed the suppression entirely. `artifacts.ts` now consolidates the duplicate `@google-cloud/storage` imports, so the `import/no-duplicates` override is gone from `.eslintrc.yaml`.

frontend/server/handlers/domain-checker.ts
jeffspahr: Called this out explicitly in the PR metadata. The title/body now describe the `domain_from_url` hostname-extraction change as a behavioral artifact-domain fix, not just lint cleanup.

frontend/package.json
manaswinidas, line 1: nit: `tslint-config-prettier` (line 162) is still listed as a devDependency but nothing references it now that `frontend/server/tslint.json` has been deleted and the root `frontend` never had its own TSLint config. Should be removed to complete the TSLint migration.

jeffspahr: Confirmed it was no longer used and removed it from `frontend/package.json` and `frontend/package-lock.json`. I also deleted the stale `frontend/mock-backend/tslint.json`, which was already broken because it extended a missing `../tslint.json` and is not referenced by any active lint script.

Issue #1235, opened 2019-04-25
flexible pipeline service (host) path in client SDK 


yaronha:
when creating an SDK `Client()` the path to `ml-pipeline` API service is loaded from a hard coded value (`ml-pipeline.kubeflow.svc.cluster.local:8888`) which indicate a specific k8s namespace. it can be valuable to load that default value from an env variable, i.e. changing the line in `_client.py` from:

`config.host = host if host else Client.IN_CLUSTER_DNS_NAME`

to:

`config.host = host or os.environ.get('ML_PIPELINE_DNS_NAME',Client.IN_CLUSTER_DNS_NAME)`

also note that when a user provide the `host` parameter, the ipython output points to the API server and not to the UI service (see the logic in `_get_url_prefix()`), it seems like a potential bug

if its acceptable i can submit a PR for the line change above

Ark-kun:
>it can be valuable to load that default value from an env variable

I'm not sure about the user scenarios, but I guess it might be useful for some cases.

>ML_PIPELINE_DNS_NAME

Maybe, `KF_PIPELINES_ENDPOINT`?

>it seems like a potential bug

It probably is. The Client class can definitely use some improvements, especially for out-of-cluster usage.

> if its acceptable i can submit a PR

We always appreciate PRs.

gaoning777:
Fixed in https://github.com/kubeflow/pipelines/pull/1293

Issue #1234, opened 2019-04-25
allow output artifact store configuration (vs hard coded)


yaronha:
it seems like the output artifacts are always stored in a specific minio service, port, namespace, bucket, secrets, etc (`minio-service.kubeflow:9000`). 

see: https://github.com/kubeflow/pipelines/blob/f40a22a3f4a8e06d20cf3e3f425b5058d5c87e0b/sdk/python/kfp/compiler/_op_to_template.py#L148

it would be great to make it flexible, e.g. allow using S3, or change namespace or bucket names.
i suggest making it configurable, i can do such PR if we agree its needed.

Ark-kun:
There is already a PR for this. But we'd like to solve this in a different way - by having an in-cluster configMap with the storage configuration. Argo already supports that, but we're hesitating to use that mechanism since KF Pipelines might not be the exclusive user of the Argo controller.

Fortunately, the problem will be solved in Argo 2.3 (per-pipeline configMap reference)

Ark-kun:
Also https://github.com/kubeflow/pipelines/issues/363 https://github.com/kubeflow/pipelines/issues/1195 https://github.com/kubeflow/pipelines/issues/362

yaronha:
@Ark-kun a mechanism like config map sounds better, it will also make the YAMLs cleaner, currently u get the same 27 lines added to every container, see below, would be nice to hide it:

```yaml
    outputs:
      artifacts:
      - name: mlpipeline-ui-metadata
        path: /mlpipeline-ui-metadata.json
        s3:
          accessKeySecret:
            key: accesskey
            name: mlpipeline-minio-artifact
          bucket: mlpipeline
          endpoint: minio-service.kubeflow:9000
          insecure: true
          key: runs/{{workflow.uid}}/{{pod.name}}/mlpipeline-ui-metadata.tgz
          secretKeySecret:
            key: secretkey
            name: mlpipeline-minio-artifact
      - name: mlpipeline-metrics
        path: /mlpipeline-metrics.json
        s3:
          accessKeySecret:
            key: accesskey
            name: mlpipeline-minio-artifact
          bucket: mlpipeline
          endpoint: minio-service.kubeflow:9000
          insecure: true
          key: runs/{{workflow.uid}}/{{pod.name}}/mlpipeline-metrics.tgz
          secretKeySecret:
            key: secretkey
            name: mlpipeline-minio-artifact
```

Ark-kun:
[...]

PR #12379, opened 2025-10-19
feat(backend): postgres integration

kaikaila:
# Summary
This PR adds full PostgreSQL (pgx driver) support to Kubeflow Pipelines backend, enabling users to choose between MySQL and PostgreSQL as the metadata database. The implementation introduces a clean dialect abstraction layer and includes a major query optimization that benefits both database backends.

Fixes #7512
Fixes #9813

## Key achievements:
- Complete PostgreSQL integration for API Server and Cache Server
- All CI tests passing (MySQL + PostgreSQL)
- Significant performance improvement for ListRuns queries. This PR is expected to address
  the root causes behind #10778, #10230, #9780, #9701
- Backward compatible with existing MySQL deployments (see Migration Guide for one metric-sort ordering change)

### 1. Storage Layer Refactoring - Dialect Abstraction
- Introduced a `DBDialect` interface encapsulating database-specific identifier quoting, placeholders, and aggregation.
- Dialect-aware filter builders (`backend/src/apiserver/storage/list_filters.go`).

### 2. ListRuns Query Performance Optimization
- Optimizes adding metrics, resource references, and tasks by performing LEFT JOINs solely on PrimaryKey UUIDs + aggregated columns, and applying the final INNER JOIN to fetch LONGTEXT spec columns.

### 3. Deployment & CI Configurations
- Standalone / Multi-user Kustomize manifests under `platform-agnostic-postgresql`.
- Adapted developer targets (`make DATABASE=postgres dev-kind-cluster`).
- Added CI workflow variants utilizing GitHub actions matrix `db_type: ["mysql", "pgx"]`.

### 4. Consistency & Inconsitency across backend databases
- casefolding, see [here](https://github.com/kubeflow/pipelines/pull/12379#discussion_r3108227948)
- null in sorting: Deterministic `NULL` ordering for sorts across MySQL/PostgreSQL (e.g. runs missing the sorted metric always sort last).
- collation details (-vs _), see [here](https://github.com/kubeflow/pipelines/pull/12379#issuecomment-4211319012)

## Testing
- Extended `integration-tests-v1` and V2 api tests to utilize both databases (with cache matrices). Unit coverage expanded.

## Migration Guide
- New PostgreSQL deployments must use an operator overlay that explicitly sets `pipeline-install-config.data.postgresExtraParams` with an `sslmode` before applying `platform-agnostic-postgresql`. The base overlay intentionally fails closed when this is omitted; see its README for an example.
[...]

Issue #7512, opened 2022-04-02
feat(backend): Add support of postgresql


yiyuanyu17:
### What feature would you like to see?

kubeflow pipeline add support of postgresql

### What is the use case or pain point?

for some case , we can not use mysql for kubeflow pipeline , hope kubeflow pipeline can add the suppoort of postgresql

Love this idea? Give it a 👍. We prioritize fulfilling features with the most 👍.

zijianjoy:
Hello @yiyuanyu17 , can you help us understand what is the reason of not being able to use mysql? And do you want to use postgresql within cluster or outside the cluster? Or are you looking for a way to configure postgresql as an alternative of cloudsql?

yiyuanyu17:
> Hello @yiyuanyu17 , can you help us understand what is the reason of not being able to use mysql? And do you want to use postgresql within cluster or outside the cluster? Or are you looking for a way to configure postgresql as an alternative of cloudsql?

hello, we use kubeflow pipeline for AI model training in our platform. In the process of privatization delivery, some customers explicitly require that self built MySQL is not allowed, and the PostgreSQL provided by the customer side must be used. Therefore, our applications are modified into ORM framework to adapt to different database types. However, it is noted that the kubeflow pipeline server has not increased its support for PostgreSQL. Therefore, it proposes this issue and hopes to get the help of the community.

zijianjoy:
Thank you for the info, @yiyuanyu17 . I will keep this issue open so people can upvote if they are also interested in this postgresql support. People can create overlay which connects to postgresql but such support is not available in this repo yet.

imiller445:
We would also be interested in this feature. We do a lot of on prem and disconnected/airgapped deployments. As such, Cloud Vendor hosted databases are not an option. In most scenarios it is easiest to run our own database clusters colocated on the same k8s environment as we run Kubeflow. The Crunchy Postgres experience on k8s is the best experience we've found to operate RDBMS clusters on k8s and we leverage it for other tooling. Would be nice to leverage it from Kubeflow as well, as operating MySql clusters on k8s is not as seamless an experience.

vasireddyvkl:
In our case we only have Postgres as an option for managed on prem DB. So looking for out fo the box Postgres support. @zijianjoy  Can you please elaborate what creating overlay means, If that helps connect Kubeflow to postgres, I am interested to give it a try. Thanks!

RoyerRamirez:
Hi @zijianjoy,

[...]

PR #12999, opened 2026-03-09
feat(backend): Add option to apply tags to create, upload pipelines and create pipeline version. Also allow filtering on the tag when fetching resources

nsingla:
### Summary                                                   
                                                                                                                                                        
Add a tags feature (map<string, string>) to pipelines and pipeline versions across the full stack:

  1. Proto: Added tags field to Pipeline and PipelineVersion messages; added UpdatePipeline and UpdatePipelineVersion RPCs; tags parameter on upload
  endpoints
  2. Model: Added Tags field to Pipeline and PipelineVersion Go models with PipelineCustomMetadata for JSON storage
  3. Storage: Full tag CRUD - CreateOrUpdatePipelineTags, CreateOrUpdatePipelineVersionTags, GetPipelineTags, GetPipelineVersionTags, tag cleanup on
  delete, list filter support (tags.<key> predicates), field updates
  4. Resource Manager: Tag validation (key/value max 20 chars), tag operations integrated into create/update/upload flows
  5. Server: Tags in create, get, list, update, upload pipeline/version endpoints
  6. Client: All tag operations exposed through the HTTP client
  7. Tests: Comprehensive positive and negative test suites covering tag CRUD, validation, filtering, and edge cases
  8. Frontend: TypeScript clients regenerated with tag support
  9. Webhook validation rewrite (pipelineversion_webhook.go:121-128): Switches from Generation-based immutability check to reflect.DeepEqual spec
  comparison.

**Checklist:**
<!--
   PR titles examples:
    * `fix(frontend): fixes empty page. Fixes #1234`
       Use `fix` to indicate that this PR fixes a bug.
    * `feat(backend): configurable service account. Fixes #1234, fixes #1235`
       Use `feat` to indicate that this PR adds a new feature. 
    * `chore: set up changelog generation tools`
       Use `chore` to indicate that this PR makes some changes that users don't need to know.
    * `test: fix CI failure. Part of #1234`
        Use `part of` to indicate that a PR is working on an issue, but shouldn't close the issue when merged.
-->


backend/src/apiserver/main.go
```diff
@@ -378,6 +378,7 @@ func startHTTPProxy(resourceManager *resource.ResourceManager, usePipelinesKuber
 	// API v2beta1
 	topMux.HandleFunc("/apis/v2beta1/pipelines/upload", sharedPipelineUploadServer.UploadPipeline)
 	topMux.HandleFunc("/apis/v2beta1/pipelines/upload_version", sharedPipelineUploadServer.UploadPipelineVersion)
+	topMux.HandleFunc("/apis/v2beta1/pipelines/{pipeline_id}/tags", sharedPipelineUploadServer.UpdatePipelineTags)
 	topMux.HandleFunc("/apis/v2beta1/healthz", func(w http.ResponseWriter, r *http.Request) {
```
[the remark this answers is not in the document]
nsingla: This is no longer valid now

backend/src/apiserver/server/pipeline_upload_server.go
```diff
@@ -396,6 +422,92 @@ func (s *PipelineUploadServer) writeErrorToResponse(w http.ResponseWriter, code
...
+	}
+
+	// Parse request body
+	var requestBody struct {
+		Tags map[string]string `json:"tags"`
+	}
+	if err := json.NewDecoder(r.Body).Decode(&requestBody); err != nil {
+		glog.Errorf("Failed to parse request body for updating tags: %v", err)
+		s.writeErrorToResponse(w, http.StatusBadRequest, errors.New("Failed to parse request body. Expected JSON with 'tags' field"))
+		return
```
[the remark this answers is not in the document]
nsingla: Shouldn't be valid anymore

backend/src/apiserver/storage/pipeline_store_kubernetes.go
```diff
@@ -599,3 +599,38 @@ func (k *PipelineStoreKubernetes) createPipelineVersionWithPipeline(ctx context.
...
+	return nil, nil
+}
+
+func (k *PipelineStoreKubernetes) GetPipelineVersionTagsForVersions(pipelineVersionIds []string) (map[string]map[string]string, error) {
+	return make(map[string]map[string]string), nil
+}
+
+func (k *PipelineStoreKubernetes) DeletePipelineVersionTags(pipelineVersionId string) error {
+	return nil
+}
```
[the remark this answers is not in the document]
nsingla: shouldn't be valid anymore

.github/actions/test-and-report/action.yml
```diff
@@ -101,9 +101,13 @@ runs:
         if [ -z $CA_CERT_PATH ]; then
           CA_CERT_PATH=''
         fi
+        pipelineStoreKubernetes='false'
+         if [ "$PIPELINE_STORE" == "kubernetes" ]; then
+          pipelineStoreKubernetes='true'
+        fi
```
hbelmiro:
```suggestion
        pipelineStoreKubernetes='false'
        if [ "$PIPELINE_STORE" == "kubernetes" ]; then
          pipelineStoreKubernetes='true'
        fi
```

.github/actions/test-and-report/action.yml
```diff
@@ -101,9 +101,13 @@ runs:
         if [ -z $CA_CERT_PATH ]; then
           CA_CERT_PATH=''
         fi
+        pipelineStoreKubernetes='false'
+         if [ "$PIPELINE_STORE" == "kubernetes" ]; then
```
hbelmiro:
```suggestion
        if [ "$PIPELINE_STORE" == "kubernetes" ]; then
```

backend/src/apiserver/storage/pipeline_store.go
```diff
@@ -1041,3 +1259,153 @@ func (s *PipelineStore) ExecuteSQL(sql string, args []interface{}, op string, ob
 	}
 	return nil
 }
+
+// createOrUpdateTags replaces all tags for a given entity in the specified table.
+// tableName is the tag table (e.g., "pipeline_tags" or "pipeline_version_tags").
+// idColumn is the foreign key column name (e.g., "PipelineId" or "PipelineVersionId").
+func (s *PipelineStore) createOrUpdateTags(tableName, idColumn, entityID string, tags map[string]string) error {
```
hbelmiro: `createOrUpdateTags` duplicates the delete+insert+gap-lock workaround pattern that already exists in `updateEntityFields`. It should delegate to `updateEntityFields` or share the same helper instead of reimplementing the logic.

Alternatively, KFP already has a `db.Upsert()` abstraction in `storage/db.go` (used by `task_store.go`) that handles both MySQL (`ON DUPLICATE KEY UPDATE`) and SQLite (`ON CONFLICT ... DO UPDATE SET`). Instead of delete-all-then-insert, upsert all new tags and then delete only stale keys (`DELETE WHERE PipelineId = ? AND TagKey NOT IN (new_keys)`). This avoids the empty-range gap lock (the DELETE targets specific keys, not the whole range), keeps everything in a single transaction, and eliminates the commit-and-reopen workaround.

backend/src/apiserver/resource/resource_manager.go
```diff
@@ -382,6 +384,85 @@ func (r *ResourceManager) UpdatePipelineDefaultVersion(pipelineId string, versio
...
+		if utf8.RuneCountInString(value) > MaxTagValueLength {
+			return util.NewInvalidInputError("tag value %q for key %q exceeds maximum length of %d characters", value, key, MaxTagValueLength)
+		}
+	}
+	return nil
+}
+
+// UpdatePipeline updates mutable fields of a pipeline (display_name, tags).
+// Both fields are updated in a single transaction via UpdatePipelineFields.
+func (r *ResourceManager) UpdatePipeline(pipelineID string, displayName string, tags map[string]string) (*model.Pipeline, error) {
```
hbelmiro, line 426: `UpdatePipeline` calls `GetPipeline` to verify existence, then `UpdatePipelineFields`, then `GetPipeline` again to return the result. This is three separate database round-trips (plus the internal transaction in `UpdatePipelineFields`). The first `GetPipeline` is a TOCTOU check -- the pipeline could be deleted between the existence check and the update.

The same pattern exists in `UpdatePipelineVersion`.

**Why it matters:** The existence check provides no atomicity guarantee. If the pipeline is deleted concurrently, the update will fail with a confusing error instead of a clean "not found." The extra round-trip also adds latency.

**Suggested fix:** Remove the pre-check `GetPipeline` call. Let `UpdatePipelineFields` fail naturally if the pipeline doesn't exist (the UPDATE will affect 0 rows), and check the affected row count to return a proper "not found" error. This eliminates one round-trip and the TOCTOU window.

backend/src/apiserver/storage/pipeline_store.go
```diff
@@ -700,6 +786,88 @@ func (s *PipelineStore) UpdatePipelineStatus(id string, status model.PipelineSta
...
+			return util.NewInternalServerError(err, "Failed to insert tag (%v=%v) for %v %v", key, value, tableName, entityID)
+		}
+	}
+	return nil
+}
+
+// updateEntityFields updates mutable fields (DisplayName) and tags for a pipeline
+// or pipeline version in a single transaction to prevent deadlocks from concurrent
+// lock acquisition.
+func (s *PipelineStore) updateEntityFields(entityTable, tagTable, idColumn, id, displayName string, tags map[string]string) error {
```
hbelmiro, line 838: When `tags` is `nil` (not provided in the request), `updateEntityFields` still executes `DELETE FROM <tagTable> WHERE <idColumn> = ?` and inserts nothing. This means calling `UpdatePipeline` with only a `display_name` change will **silently clear all tags**.

The same issue exists for `UpdatePipelineVersion`.

**Why it matters:** Users updating only the display name will unexpectedly lose all their tags. This is a data loss bug. The API should distinguish between "tags not provided" (no change) and "tags explicitly set to empty map" (clear tags). The proto `map` field distinguishes between absent (nil in Go) and empty (non-nil, zero-length map).

**Suggested fix:** In `updateEntityFields`, skip the tag delete+insert logic when `tags == nil`. Only perform tag replacement when tags is explicitly provided (non-nil). Also update the proto comments on `UpdatePipelineRequest` and `UpdatePipelineVersionRequest` to document the semantics: "If tags is omitted, existing tags are preserved. If tags is set to an empty map, all tags are cleared."

backend/src/apiserver/storage/pipeline_store_kubernetes.go
```diff
@@ -32,6 +33,12 @@ var (
 type PipelineStoreKubernetes struct {
 	client        ctrlclient.Client
 	clientNoCache ctrlclient.Client
+	// freshPipelines stores pipeline models from update responses to avoid
+	// stale reads from the informer cache after writes. Entries are consumed
+	// (deleted) on the next GetPipeline call for the same ID.
+	freshPipelines sync.Map
+	// freshPipelineVersions stores pipeline version models similarly.
+	freshPipelineVersions sync.Map
```
hbelmiro: The `freshPipelines` and `freshPipelineVersions` `sync.Map` caches are consumed via `LoadAndDelete` on the next `GetPipeline` call. This creates several problems:

1. **Memory leak:** If `UpdatePipelineFields` is called but `GetPipeline` is never called for that ID (e.g., the update response is sufficient), the entry stays in the map forever.
2. **Single-consumer:** Only the first `GetPipeline` call after an update gets fresh data; subsequent calls before the informer syncs still get stale data.
3. **No expiration:** There's no TTL or cleanup mechanism.
4. **Concurrency concern:** Multiple goroutines calling `UpdatePipelineFields` for the same ID will overwrite each other's entries in the `sync.Map` without coordination.

**Why it matters:** In a long-running API server, this can accumulate stale entries over time, and the single-consume pattern means concurrent readers may still get stale data. The `clientNoCache` field already exists on the struct but isn't used here.

**Suggested fix:** Use `clientNoCache` for the read-after-write in the update path instead of caching. This is simpler, correct, and the field already exists for this purpose.

backend/src/apiserver/webhook/pipelineversion_webhook.go
```diff
@@ -120,8 +121,15 @@ func (p *PipelineVersionsWebhook) ValidateUpdate(_ context.Context, oldObj, newO
...
+	// Allow mutable field changes (DisplayName, Tags) but reject changes to
+	// immutable pipeline spec fields.
+	if !reflect.DeepEqual(oldPipelineVersion.Spec.PipelineSpec, newPipelineVersion.Spec.PipelineSpec) ||
+		!reflect.DeepEqual(oldPipelineVersion.Spec.PlatformSpec, newPipelineVersion.Spec.PlatformSpec) ||
+		oldPipelineVersion.Spec.PipelineName != newPipelineVersion.Spec.PipelineName ||
+		oldPipelineVersion.Spec.Description != newPipelineVersion.Spec.Description ||
+		oldPipelineVersion.Spec.CodeSourceURL != newPipelineVersion.Spec.CodeSourceURL ||
+		oldPipelineVersion.Spec.PipelineSpecURI != newPipelineVersion.Spec.PipelineSpecURI {
+		return nil, newBadRequestError("Pipeline spec is immutable; only mutable fields (display_name, tags) can be updated")
 	}
```
hbelmiro: Optional: The `ValidateUpdate` webhook uses field-by-field comparison with `reflect.DeepEqual` to check immutability. The approach enumerates every immutable field explicitly, which is fragile. If a new immutable field is added to `PipelineVersionSpec` in the future, it won't be caught here unless this code is also updated.

**Why it matters:** Missing a new immutable field in this check would silently allow mutation of data that should be immutable.

**Suggested fix:** Invert the check: instead of listing all immutable fields, copy the old spec and overwrite only the mutable fields (`DisplayName`, `Tags`) from the new spec, then compare the entire spec. This way, new fields are immutable by default and don't require updating this code:

```go
expected := oldPipelineVersion.Spec.DeepCopy()
expected.DisplayName = newPipelineVersion.Spec.DisplayName
expected.Tags = newPipelineVersion.Spec.Tags
if !reflect.DeepEqual(*expected, newPipelineVersion.Spec) {
    return nil, newBadRequestError("Pipeline spec is immutable; only mutable fields (display_name, tags) can be updated")
}
```

backend/test/v2/api/pipeline_api_test.go
hbelmiro, line 1: Can you please add tests for the following scenarios?

1. **Update with nil tags** -- calling `UpdatePipeline` with only `display_name` and no tags field to verify tags are preserved. Note: tag clearing with an explicit empty map `{}` **is** covered by the "Remove all tags by passing empty map" test case.
2. **Unit tests for `validateTags`** -- the validation function has no dedicated unit tests; it's only tested indirectly through integration tests.
3. **Unit tests for `extractTagFiltersFromFilterSpec`** -- complex parsing logic with no dedicated unit tests.
4. **Unit tests for `updateEntityFields`** -- new store method with no unit tests.

backend/api/v2beta1/pipeline.proto
```diff
@@ -136,6 +136,22 @@ service PipelineService {
       delete: "/apis/v2beta1/pipelines/{pipeline_id}/versions/{pipeline_version_id}"
     };
   }
+
+  // Updates a pipeline's mutable fields (display_name, tags).
+  rpc UpdatePipeline(UpdatePipelineRequest) returns (Pipeline) {
+    option (google.api.http) = {
+      put: "/apis/v2beta1/pipelines/{pipeline.pipeline_id}"
```
hbelmiro: How about using `patch` to make partial update semantics explicit?

backend/api/v2beta1/pipeline.proto
```diff
@@ -136,6 +136,22 @@ service PipelineService {
...
+      body: "pipeline"
+    };
+  }
+
+  // Updates a pipeline version's mutable fields (display_name, tags).
+  rpc UpdatePipelineVersion(UpdatePipelineVersionRequest) returns (PipelineVersion) {
+    option (google.api.http) = {
+      put: "/apis/v2beta1/pipelines/{pipeline_version.pipeline_id}/versions/{pipeline_version.pipeline_version_id}"
```
hbelmiro: How about using `patch` to make partial update semantics explicit?

backend/api/v2beta1/python_http_client/test/test_pipeline_service_api.py
```diff
@@ -98,6 +98,20 @@ def test_pipeline_service_list_pipelines(self):
...
+        Updates a pipeline's mutable fields (display_name, tags).  # noqa: E501
+        """
+        pass
+
+    def test_pipeline_service_update_pipeline_version(self):
+        """Test case for pipeline_service_update_pipeline_version
+
+        Updates a pipeline version's mutable fields (display_name, tags).  # noqa: E501
+        """
+        pass
```
hbelmiro, line 113: Not implemented.

backend/src/apiserver/model/pipeline_custom_metadata.go
```diff
@@ -0,0 +1,55 @@
+// Copyright 2025 The Kubeflow Authors
```
hbelmiro:
```suggestion
// Copyright 2026 The Kubeflow Authors
```

backend/src/apiserver/model/pipeline_custom_metadata.go
```diff
@@ -0,0 +1,55 @@
...
+// PipelineTag stores a single user-defined tag (key-value pair) associated with a pipeline.
+// The composite primary key (PipelineId, TagKey) ensures uniqueness per pipeline.
+// Both TagKey and TagValue are limited to 20 characters.
+type PipelineTag struct {
+	// PipelineID references the parent pipeline.
+	PipelineID string `gorm:"column:PipelineId; not null; primaryKey; type:varchar(64);"`
+	// TagKey is the tag key, limited to 20 characters.
+	TagKey string `gorm:"column:TagKey; not null; primaryKey; type:varchar(20);"`
+	// TagValue is the tag value, limited to 20 characters.
+	TagValue string `gorm:"column:TagValue; type:varchar(20);"`
```
hbelmiro:
```suggestion
	TagValue string `gorm:"column:TagValue; not null; type:varchar(20);"`
```

backend/src/apiserver/model/pipeline_custom_metadata.go
```diff
@@ -0,0 +1,55 @@
...
+// PipelineVersionTag stores a single user-defined tag (key-value pair) associated with a pipeline version.
+// The composite primary key (PipelineVersionId, TagKey) ensures uniqueness per pipeline version.
+// Both TagKey and TagValue are limited to 20 characters.
+type PipelineVersionTag struct {
+	// PipelineVersionID references the parent pipeline version.
+	PipelineVersionID string `gorm:"column:PipelineVersionId; not null; primaryKey; type:varchar(191);"`
+	// TagKey is the tag key, limited to 20 characters.
+	TagKey string `gorm:"column:TagKey; not null; primaryKey; type:varchar(20);"`
+	// TagValue is the tag value, limited to 20 characters.
+	TagValue string `gorm:"column:TagValue; type:varchar(20);"`
```
hbelmiro:
```suggestion
	TagValue string `gorm:"column:TagValue; not null; type:varchar(20);"`
```

backend/src/apiserver/storage/pipeline_store.go
```diff
@@ -700,6 +786,88 @@ func (s *PipelineStore) UpdatePipelineStatus(id string, status model.PipelineSta
 	return s.ExecuteSQL(sql, args, "update", "pipeline status")
 }
 
+// insertTagsInTx inserts tags into the specified tag table within the given transaction.
+// Returns an error if any tag insert fails; the caller is responsible for rolling back.
+func insertTagsInTx(tx *sql.Tx, tableName, idColumn, entityID string, tags map[string]string) error {
```
hbelmiro, line 793: `insertTagsInTx` inserts each tag with a separate `INSERT` statement inside a loop (up to 10 iterations, per the max tags limit). A single batch `INSERT INTO ... VALUES (...), (...), (...)` would reduce round-trips from N to 1. The squirrel builder supports this via chained `.Values()` calls on a single `Insert`.

backend/src/apiserver/storage/pipeline_store.go
```diff
@@ -373,6 +413,23 @@ func (s *PipelineStore) ListPipelines(filterContext *model.FilterContext, opts *
...
+		tagsMap, err := s.GetPipelineTagsForPipelines(pipelineIds)
+		if err != nil {
+			return nil, 0, "", util.NewInternalServerError(err, "Failed to load tags for listed pipelines")
+		}
+		for _, p := range pipelines {
+			if tags, ok := tagsMap[p.UUID]; ok {
+				p.Tags = tags
+			}
+		}
+	}
```
hbelmiro, line 425: Tag loading for listed pipelines/versions happens *outside* the transaction. The transaction is committed, then tags are loaded in separate queries. If tags are modified concurrently between the commit and the tag load, the returned tags could be inconsistent with the pipeline/version data.

**Suggested fix:** Move the tag loading inside the transaction, before the commit. This ensures the tag data is consistent with the pipeline/version data returned in the same response.

backend/src/apiserver/storage/pipeline_store.go
```diff
@@ -998,6 +1199,23 @@ func (s *PipelineStore) ListPipelineVersions(pipelineId string, opts *list.Optio
...
+		allTags, err := s.GetPipelineVersionTagsForVersions(versionIds)
+		if err != nil {
+			return nil, 0, "", util.NewInternalServerError(err, "Failed to load tags for listed pipeline versions")
+		}
+		for _, pv := range pipelineVersions {
+			if tags, ok := allTags[pv.UUID]; ok {
+				pv.Tags = tags
+			}
+		}
+	}
```
hbelmiro, line 1258: Tag loading for listed pipelines/versions happens *outside* the transaction. The transaction is committed, then tags are loaded in separate queries. If tags are modified concurrently between the commit and the tag load, the returned tags could be inconsistent with the pipeline/version data.

**Suggested fix:** Move the tag loading inside the transaction, before the commit. This ensures the tag data is consistent with the pipeline/version data returned in the same response.

backend/src/apiserver/storage/pipeline_store_kubernetes.go
```diff
@@ -599,3 +629,125 @@ func (k *PipelineStoreKubernetes) createPipelineVersionWithPipeline(ctx context.
 
 	return k8sPipelineVersion.ToModel()
 }
+
+func (k *PipelineStoreKubernetes) UpdatePipelineFields(pipelineID string, displayName string, tags map[string]string) error {
```
hbelmiro, line 673: Same nil-clears-tags bug as the SQL store. It unconditionally sets `k8sPipeline.Spec.Tags = tags`. When `tags` is `nil`, this clears all tags on the CRD.

**Suggested fix:** skip the tag assignment when `tags == nil`.

backend/src/apiserver/storage/pipeline_store_kubernetes.go
```diff
@@ -599,3 +629,125 @@ func (k *PipelineStoreKubernetes) createPipelineVersionWithPipeline(ctx context.
...
+	if err := k.client.Update(context.TODO(), k8sPipeline); err != nil {
+		return util.NewInternalServerError(err, "Failed to update pipeline %v", pipelineID)
+	}
+	// Store the fresh pipeline model so the next GetPipeline call returns
+	// up-to-date data instead of potentially stale informer cache data.
+	k.freshPipelines.Store(pipelineID, k8sPipeline.ToModel())
+	return nil
+}
+
+func (k *PipelineStoreKubernetes) UpdatePipelineVersionFields(pipelineVersionID string, displayName string, tags map[string]string) error {
```
hbelmiro, line 692: Same nil-clears-tags bug as the SQL store. It unconditionally sets `k8sPipeline.Spec.Tags = tags`. When `tags` is `nil`, this clears all tags on the CRD.

**Suggested fix:** skip the tag assignment when `tags == nil`.

backend/src/apiserver/storage/pipeline_store_kubernetes.go
```diff
@@ -599,3 +629,125 @@ func (k *PipelineStoreKubernetes) createPipelineVersionWithPipeline(ctx context.
...
+	if displayName != "" {
+		k8sPipelineVersion.Spec.DisplayName = displayName
+	}
+	k8sPipelineVersion.Spec.Tags = tags
+	if err := k.client.Update(context.TODO(), k8sPipelineVersion); err != nil {
+		return util.NewInternalServerError(err, "Failed to update pipeline version %v", pipelineVersionID)
+	}
+	// Store the fresh pipeline version model so the next GetPipelineVersion call
+	// returns up-to-date data instead of potentially stale informer cache data.
+	if freshModel, err := k8sPipelineVersion.ToModel(); err == nil {
```
hbelmiro: The error from `k8sPipelineVersion.ToModel()` is silently swallowed.

backend/src/apiserver/server/pipeline_upload_server.go
hbelmiro, line 172: When uploading a pipeline with `?tags={"team":"ml-ops"}`, tags are parsed and set on the pipeline, but the first pipeline version created alongside it does not inherit the tags:

```go
pipelineVersion := &model.PipelineVersion{
    Name:         pipeline.Name,
    DisplayName:  pipeline.DisplayName,
    Description:  pipeline.Description,
    PipelineSpec: model.LargeText(pipelineFile),
}
```

`pipelineVersion.Tags` is never set, so the initial version has no tags even though the pipeline does. This is inconsistent with `CreatePipelineAndVersion` (in `pipeline_server.go`) which accepts separate `versionTags`.

Is this intentional?

backend/test/testutil/test_utils.go
```diff
@@ -48,6 +49,20 @@ func ParsePointersToString(s *string) string {
...
 
+// TagsMapToJSONStringPtr serializes a map[string]string to a JSON string pointer.
+// Returns nil if the map is nil.
+func TagsMapToJSONStringPtr(m map[string]string) *string {
+	if m == nil {
+		return nil
+	}
+	data, err := json.Marshal(m)
+	if err != nil {
+		return nil
```
hbelmiro: I think we should return `err`.

backend/src/apiserver/server/pipeline_upload_server.go
nsingla: yeah, this was intentional as the tags on version are different. But i see your point, if we inherit tags to pipeline version from pipeline. But then that means it has to happen for update as well

nsingla: what do you think?

hbelmiro: The problem isn't really about inheritance. It's about API completeness.

Here's the situation:
  1. `CreatePipelineAndVersion` (gRPC/REST in `pipeline_server.go`) — accepts separate versionTags via request.GetPipelineVersion().GetTags(), so the caller can explicitly set tags on both the pipeline and its first version independently. This is clean and explicit.
  2. `uploadPipeline` (multipart upload in `pipeline_upload_server.go`) — accepts a single ?tags=... query param, sets it on the pipeline, but the first version created alongside it gets no tags at all. There's no way for the caller to specify version tags through this endpoint.

I suggest adding a separate version_tags query param to the upload endpoint, mirroring how `CreatePipelineAndVersion` handles it.
Optionally, we could also add inheritance to both endpoints as a fallback (when the pipeline has tags, but the version doesn't). But that may not make sense, since you said that the tags on version are different.

mprahl: I think `CreatePipelineAndVersion` can just set tags on the pipeline. Version tags can be added separately to keep the experience consistent for the first pipeline version and subsequent pipeline versions.

backend/api/v2beta1/python_http_client/test/test_pipeline_service_api.py
nsingla: I believe these are self generated, but these scenarios are tested in the api tests as well as unit tests

backend/src/apiserver/storage/pipeline_store.go
```diff
@@ -700,6 +787,109 @@ func (s *PipelineStore) UpdatePipelineStatus(id string, status model.PipelineSta
...
+
+// upsertTagsInTx upserts tags using REPLACE INTO, which atomically deletes
+// any existing row with the same primary key and inserts the new one.
+// This avoids gap locks that cause deadlocks with the DELETE + INSERT pattern.
+// The tag tables have a composite primary key on (idColumn, TagKey), which
+// makes this safe. REPLACE INTO is supported by both MySQL and SQLite.
+func upsertTagsInTx(tx *sql.Tx, tableName, idColumn, entityID string, tags map[string]string) error {
+	for key, value := range tags {
+		replaceSQL := fmt.Sprintf(
+			"REPLACE INTO %s (%s, TagKey, TagValue) VALUES (?, ?, ?)",
```
hbelmiro: `REPLACE INTO` is not a true upsert. On both MySQL and SQLite it performs a **DELETE + INSERT** when a row with the matching primary key already exists. The function comment says this "avoids gap locks that cause deadlocks with the DELETE + INSERT pattern," but that concern applies to the **blanket DELETE** pattern (`DELETE FROM pipeline_tags WHERE PipelineId = ?`) which acquires gap locks on the index range. Both `REPLACE INTO` and `ON DUPLICATE KEY UPDATE` target specific primary keys (the composite `(PipelineId, TagKey)`), not ranges, so neither causes gap locks. In fact, the `updateEntityFields` comment on line 831 already describes the intent as "INSERT ... ON DUPLICATE KEY UPDATE" -- the implementation in `upsertTagsInTx` is inconsistent with this documented intent.

The practical downsides of `REPLACE INTO` compared to `ON DUPLICATE KEY UPDATE`:

- The row gets a new internal row ID (InnoDB's hidden `DB_ROW_ID` changes on every "update"). This generates extra row-change events in row-based replication binlogs and inflates monitoring metrics that track row churn.
- Any `BEFORE DELETE` / `AFTER DELETE` triggers fire unnecessarily on every tag value change, even though the logical intent is an update.
- If any future table adds a FK referencing `pipeline_tags` (making it a parent), the `REPLACE INTO` delete step would cascade-delete those child rows on every tag value change. The tag tables are leaf tables today, so this is not currently exploitable, but it's a latent risk.

The codebase already has a proper upsert abstraction in `storage/db.go`:

```go
// MySQL
func (d MySQLDialect) Upsert(query, key string, overwrite bool, columns ...string) string {
    // appends: ON DUPLICATE KEY UPDATE col=VALUES(col), ...
}
// SQLite
func (d SQLiteDialect) Upsert(query, key string, overwrite bool, columns ...string) string {
    // appends: ON CONFLICT(key) DO UPDATE SET col=excluded.col, ...
}
```

This is used by `task_store.go` and generates the standard `ON DUPLICATE KEY UPDATE` (MySQL) / `ON CONFLICT ... DO UPDATE SET` (SQLite) clauses, which update in-place without deleting.

hbelmiro: `tableName` and `idColumn` are interpolated directly into the SQL string via `fmt.Sprintf`. The values are hardcoded constants from within the codebase (`"pipeline_tags"`, `"PipelineId"`, etc.), so this is not exploitable today. However, every other SQL query in `pipeline_store.go` uses the squirrel query builder, which parameterizes all values. Mixing raw `fmt.Sprintf` SQL makes the code harder to audit and sets a precedent that could lead to injection if a future caller passes user-derived values.

backend/src/apiserver/storage/pipeline_store.go
nsingla: This is already addressed

backend/src/apiserver/server/pipeline_server.go
```diff
@@ -937,15 +1019,41 @@ func (s *PipelineServer) ListPipelineVersions(ctx context.Context, request *apiv
...
+	if err != nil {
+		return nil, util.Wrapf(err, "Failed to list pipeline versions due invalid list options: pageToken: %v, pageSize: %v, sortBy: %v, filter: %v", pageToken, int(pageSize), sortBy, cleanedFilterSpec)
+	}
+
+	// Fail fast if pipeline id is missing
+	if pipelineId == "" {
+		return nil, util.NewInvalidInputError("Failed to list pipeline versions. Pipeline id cannot be empty")
+	}
+
+	// Check authorization
```
hbelmiro: The v2beta1 `ListPipelineVersions` implementation was refactored to add tag filtering, but it also added inline authorization and pipeline ID validation that duplicates logic already in `listPipelineVersions` (the shared base method). The base method `listPipelineVersions` already performs authorization and pipeline ID validation. The v2beta1 handler now bypasses the base method entirely and calls `s.resourceManager.ListPipelineVersions` directly.

**Why it matters:** This creates two code paths for listing pipeline versions: the v1beta1 path goes through the shared base method (with its authorization logic), while v2beta1 has its own inline authorization. If authorization logic changes in the base method, the v2beta1 path won't pick it up.

**Suggested fix:** Extend the shared `listPipelineVersions` base method to accept `tagFilters` (passing `nil` from v1beta1), rather than duplicating the authorization and validation logic inline.

backend/src/apiserver/storage/pipeline_store.go
nsingla: Done, thank you for this

backend/src/apiserver/server/pipeline_server.go
nsingla: Done

backend/src/apiserver/storage/pipeline_store_kubernetes.go
```diff
@@ -479,26 +507,13 @@ func (k *PipelineStoreKubernetes) deleteWithTimeout(namespace string, name strin
...
-
-	for _, k8sPipeline := range pipelines.Items {
-		if string(k8sPipeline.UID) == pipelineId {
-			return &k8sPipeline, nil
-		}
-	}
-
-	// Fallback to not using the cache
-	err = k.clientNoCache.List(context.TODO(), &pipelines, listOptions...)
+	err := k.clientNoCache.List(context.TODO(), &pipelines, listOptions...)
```
hbelmiro: The `getK8sPipeline` and `getK8sPipelineVersions` functions were changed to:
1. Remove `ctrlclient.UnsafeDisableDeepCopy`
2. Skip the cached client entirely, always using `clientNoCache`
3. Remove the cache-first-then-fallback pattern

**Context:** The old code on master had a latent bug: `getK8sPipeline` used `UnsafeDisableDeepCopy` with the cached client, which returns pointers into the cache's internal storage. Callers like `DeletePipeline` then passed these objects to `k.client.Delete()`, which mutates the object (sets deletion timestamp, etc.), corrupting the cache. The new `UpdatePipelineFields` and `CreateOrUpdatePipelineTags` callers also mutate the object before calling `k.client.Update()`. Removing `UnsafeDisableDeepCopy` and switching to `clientNoCache` fixes this correctness issue.

**Remaining concern:** The change is broader than necessary. `getK8sPipelineVersions` is also called by `ListPipelineVersions` (a read-heavy list operation) and `GetLatestPipelineVersion`, which don't mutate the result. These read-only paths now also bypass the cache, increasing K8s API server load. `ListPipelines` still uses the cached client with `UnsafeDisableDeepCopy`, creating an inconsistency.

**Suggested fix:** Consider splitting `getK8sPipeline` into two variants: one that returns a mutable copy (for update/delete callers, using `clientNoCache` without `UnsafeDisableDeepCopy`) and one that returns a read-only view (for `GetPipeline`/`GetPipelineTags`, using the cached client with `UnsafeDisableDeepCopy`). The same split could apply to `getK8sPipelineVersions` for the `ListPipelineVersions` path. This would preserve the correctness fix while avoiding the performance regression on read paths.

nsingla:
> Just one more comment. Also, there's still Matt's comment here: [#12999 (comment)](https://github.com/kubeflow/pipelines/pull/12999#discussion_r2955451189)

I think Matt is agreeing to not add the pipeline version tags options to the upload endpoint for now

backend/src/apiserver/storage/pipeline_store_kubernetes.go
nsingla: The reason is because of race conditions that were happening in the tests, that after an update/delete tests were trying to do a get or list and were still fetching old object from cache, because of latency in updating the objects (k8s api delays). The suggested fix still has high probability of running into the same race condition, if the call to update and get are right after each other.

backend/src/apiserver/model/pipeline_custom_metadata.go
```diff
@@ -0,0 +1,55 @@
...
+}
+
+// PipelineVersionTag stores a single user-defined tag (key-value pair) associated with a pipeline version.
+// The composite primary key (PipelineVersionId, TagKey) ensures uniqueness per pipeline version.
+// Both TagKey and TagValue are limited to 20 characters.
+type PipelineVersionTag struct {
+	// PipelineVersionID references the parent pipeline version.
+	PipelineVersionID string `gorm:"column:PipelineVersionId; not null; primaryKey; type:varchar(191);"`
+	// TagKey is the tag key, limited to 20 characters.
+	TagKey string `gorm:"column:TagKey; not null; primaryKey; type:varchar(20);"`
```
mprahl: Can we keep the length similar to Kubernetes labels? I think it's 63 characters.

backend/src/apiserver/model/pipeline_custom_metadata.go
```diff
@@ -0,0 +1,55 @@
...
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package model
+
+// PipelineTag stores a single user-defined tag (key-value pair) associated with a pipeline.
+// The composite primary key (PipelineId, TagKey) ensures uniqueness per pipeline.
+// Both TagKey and TagValue are limited to 20 characters.
+type PipelineTag struct {
```
mprahl, line 20: Can we add an index on (TagKey, TagValue, PipelineId)? The ordering is important so that existing queries can benefit from simple tag key and tag value filtering.

Same applies to pipeline version.

backend/src/apiserver/model/pipeline_custom_metadata.go
```diff
@@ -0,0 +1,55 @@
...
+	// PipelineID references the parent pipeline.
+	PipelineID string `gorm:"column:PipelineId; not null; primaryKey; type:varchar(64);"`
+	// TagKey is the tag key, limited to 20 characters.
+	TagKey string `gorm:"column:TagKey; not null; primaryKey; type:varchar(20);"`
```
mprahl: I'm debating if `tag` is the right term or if we should use the term `label` as this is more closely aligned with Kubernetes.

backend/src/apiserver/resource/resource_manager.go
```diff
@@ -382,6 +384,75 @@ func (r *ResourceManager) UpdatePipelineDefaultVersion(pipelineId string, versio
...
 }
 
+// MaxTagKeyLength is the maximum allowed length (in characters) for a tag key.
+const MaxTagKeyLength = 20
+
+// MaxTagValueLength is the maximum allowed length (in characters) for a tag value.
+const MaxTagValueLength = 20
+
+// MaxTagsPerEntity is the maximum number of tags allowed on a single pipeline or pipeline version.
+const MaxTagsPerEntity = 10
```
mprahl: Why a maximum of 10?

backend/src/apiserver/storage/pipeline_store_kubernetes.go
```diff
@@ -107,7 +123,7 @@ func (k *PipelineStoreKubernetes) ListPipelines(filterContext *model.FilterConte
 				case float64:
 					return elementA.(float64) > elementB.(float64)
 				case string:
-					return elementA.(string) > elementB.(string)
+					return strings.ToLower(elementA.(string)) > strings.ToLower(elementB.(string))
```
mprahl, line 126: Why this change?

backend/src/apiserver/storage/pipeline_store_kubernetes.go
mprahl: I prefer `UnsafeDisableDeepCopy`. The fix in my opinion, is to do a deep copy prior to any update or delete. Otherwise, we use a lot more memory for every list operation. The pipeline specs are not small in size.

The Kubernetes native API code already tries to handle some of these cache inconsistencies. On updates and deletes, before returning, the code should polling the cache until the delete or update is seen by the cache and use a timeout. There's a similar pattern somewhere for this.

backend/src/apiserver/webhook/pipelineversion_webhook.go
```diff
@@ -120,8 +121,15 @@ func (p *PipelineVersionsWebhook) ValidateUpdate(_ context.Context, oldObj, newO
 		return nil, newBadRequestError(fmt.Sprintf("Expected a PipelineVersion but got %T", newObj))
 	}
 
-	if oldPipelineVersion.Generation != newPipelineVersion.Generation {
```
mprahl, line 123: Only do the follow up checks if the generation changes. We can skip this for metadata and status only changes.

backend/src/apiserver/storage/pipeline_store_kubernetes.go
nsingla: because this was incosistent, SQL sorting was case insensitive and K8s was case sensitive, producing 2 different sorting results.

backend/src/apiserver/resource/resource_manager.go
nsingla: Just a number I thought was more than enough, we shouldn't allow too many tags per pipeline/pipeline version as well. Happy to change if you have something else in mind

backend/src/apiserver/model/pipeline_custom_metadata.go
nsingla: I also debated, and then I asked claude and codex, both recommended tag. I believe MLflow also uses `tag` for custom metadata. It could also cause confusion when using both k8s labels and pipeline tags

backend/src/apiserver/model/pipeline_custom_metadata.go
nsingla: sure

nsingla: Done

backend/src/apiserver/model/pipeline_custom_metadata.go
nsingla: Done

backend/src/apiserver/webhook/pipelineversion_webhook.go
nsingla: Done

backend/src/apiserver/storage/pipeline_store_kubernetes.go
mprahl: @nsingla this seems to be an inconsistent default per database. Is the query specifically ordering by `LOWER(column)`, if not, we may need to adjust the query get this consistent.

backend/src/apiserver/resource/resource_manager.go
mprahl: Let's maybe bump this to 20 or 25. 10 seems too low IMO.

nsingla: ok, bumping it to 20

backend/src/apiserver/storage/pipeline_store_kubernetes.go
nsingla: I think its mysql default:
```
In MySQL, sorting on character columns is case-insensitive by default, provided the column uses a case-insensitive collation (which most default to). You can force a specific case-insensitive sort for a query using the COLLATE clause or the LOWER() function.
```
So i can force the query to use lower(), which should work for postgres as well (in the future)

nsingla: or can we address this in a separate PR? this change is bringing consistency, and will only break once we start to support more than mysql database.

backend/src/apiserver/resource/resource_manager.go
nsingla: Done

backend/src/apiserver/storage/pipeline_store_kubernetes.go
nsingla: Fixed, update, delete are are now using polling cache with a  timeout, and the rest is left as it is, i.e. with UnsafeDisableDeepCopy

backend/src/apiserver/storage/pipeline_store_kubernetes.go
mprahl: @nsingla that's fine just tag the PR author for this #7512 about this

mprahl:
/approve based on looking at the new API changes, database schema, and Kubernetes native API changes.

I'll defer to @hbelmiro to provide another lgtm.

backend/src/apiserver/storage/pipeline_store_kubernetes.go
kaikaila: Thanks for flagging this issue. My response is [here](https://github.com/kubeflow/pipelines/pull/12379#issuecomment-4115471405)

PR #13002, opened 2026-03-10
docs: add testing policy to AGENTS.md

jsonmp-k8:
## Summary
- Add a testing policy section to AGENTS.md requiring:
  - Unit tests for every new function, method, or exported API
  - All existing tests must pass locally before pushing changes
  - New test cases when modifying existing behavior
  - Explicit PR notes for pre-existing test failures

## Test plan
- [x] Documentation-only change, no code affected

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -28,11 +28,8 @@
 
 ### Testing policy (agents and contributors)
 
-- Every new function, method, or exported API must have accompanying unit tests before merging.
-- All existing tests must pass locally before pushing changes. Run the relevant test suites:
-  - Go backend: `go test -v $(go list ./backend/... | grep -v backend/test/)`
-  - Python SDK: `pytest -v sdk/python/kfp`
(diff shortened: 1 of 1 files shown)


jsonmp-k8:
> A few observations:
> 
> 1. **Duplicates existing content.** The test commands are already documented in the "Local testing" section (lines 163–230) and the "Quick reference" section (lines 584–593). Consider referencing those sections instead of inlining the commands again — keeps one source of truth and avoids drift.
> 2. **Missing `kfp-kubernetes` tests.** The existing "Local testing" section includes `pytest -v kubernetes_platform/python/test`, but the new policy omits it. If the intent is to list what contributors should run, this gap could mislead.
> 3. **"Every new function" may be too broad.** Requiring unit tests for _every_ new function/method is a high bar for trivial helpers, glue code, or generated-code adapters. Consider scoping to "non-trivial functions and public APIs" to avoid policy that gets routinely ignored.

@manaswinidas i have addressed the issues

PR #13007, opened 2026-03-11
fix(components): use == instead of : for snowflake package version spec

jsonmp-k8:
## Summary
- Fix invalid pip version specifier in Snowflake component: `snowflake-connector-python:3.12.3` → `snowflake-connector-python==3.12.3`
- pip requires `==` for version pinning — the colon syntax causes `pip install` to fail at runtime
- Fixed in both `snowflake_unload_data.py` (source) and `component.yaml` (compiled spec)

## Test plan
- [ ] Verify `pip install snowflake-connector-python==3.12.3` succeeds (the `:` variant would fail with `Invalid requirement`)

### DIFF
--- a/components/snowflake/snowflake_unload_data.py
+++ b/components/snowflake/snowflake_unload_data.py
@@ -7,7 +7,7 @@ from kfp.dsl import component
 
 @component(
     base_image="python:3.11-slim",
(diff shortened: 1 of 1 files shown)


jsonmp-k8:
Superseded by #13014 which includes this fix along with the Go dependency upgrades.

PR #13009, opened 2026-03-11
feat(frontend,backend): show actual retry start time for pipeline runs. Fixes #12845

### LINKED ISSUE #12845: [feature] Show actual retry start time for pipeline runs
### Feature Area

/area frontend
/area backend

When a pipeline run is retried (either manually or via recurring/scheduled runs), the Start time displayed in the Kubeflow UI reflects the original scheduled start time rather than the actual retry execution time.

It would be helpful if the UI displayed:

- the actual retry start time, or
- both values: Scheduled start time, Actual execution (retry) start time

This would make run timelines more accurate and easier to interpret.

When investigating failures or delays, the current behavior can be misleading.

For recurring runs in particular:

- a run scheduled at 02:00 may retry at 04:30
- the UI still shows Start time = 02:00
- this creates confusion when debugging latency, SLA compliance, or resource availability issues

Currently, users must:
- inspect pod start times in Kubernetes,
- check logs,
- or compare timestamps manually.

This is time-consuming and not intuitive.

---

Love this idea? Give it a 👍.

Sridhar1030:
## Description

When a pipeline run is retried (manually or via recurring/scheduled runs),
the **Start time** shown in the UI reflected the original scheduled time
rather than the actual retry execution time. This was misleading when
debugging failures, SLA compliance, or resource availability issues.
Before : 
image
Notice that after retry the started at doesnot change 
After
Screenshot 2026-03-11 at 1 33 30 PM
The started at reflects the retried time

## Changes

**Backend (`resource_manager.go`)**
- Refactored `RetryRun` to update fields directly on the fetched `run`
  object before persisting, rather than constructing a sparse struct.
  This preserves all existing run fields (e.g. namespace, service account)
  correctly on retry.

**Frontend (`RunDetailsV2.tsx`)**
- Added `getActualStartTime()` that reads the last `RUNNING` entry from
  `state_history` to determine the actual execution start time, with a
  safe guard for missing `update_time`, falling back to `scheduled_at`.
- Updated run details to display `Started at` as the actual execution time.
- When the actual start time differs from the scheduled time, a
  `Scheduled at` row is conditionally shown alongside it.

## Testing

- Deployed locally against a Kind cluster and verified:
  - `Started at` shows the retry execution time after a manual retry
  - `Scheduled at` row appears only when the two times differ
  - `Started at` correctly falls back to `scheduled_at` when no
    `state_history` is present or `update_time` is undefined

### DIFF
--- a/backend/src/apiserver/resource/resource_manager_test.go
+++ b/backend/src/apiserver/resource/resource_manager_test.go
@@ -2588,14 +2588,14 @@ func TestRetryRun_PreservesRunFields(t *testing.T) {
 	defer store.Close()
 
 	originalRun, err := manager.GetRun(runDetail.UUID)
-	assert.Nil(t, err)
+	require.Nil(t, err)
 	assert.Equal(t, string(v1alpha1.WorkflowFailed), string(originalRun.Conditions))
 
 	err = manager.RetryRun(context.Background(), runDetail.UUID)
-	assert.Nil(t, err)
+	require.Nil(t, err)
 
 	retriedRun, err := manager.GetRun(runDetail.UUID)
-	assert.Nil(t, err)
+	require.Nil(t, err)
 
 	// Core identifying fields must be preserved after retry
 	assert.Equal(t, originalRun.UUID, retriedRun.UUID)
@@ -2611,7 +2611,7 @@ func TestRetryRun_PreservesRunFields(t *testing.T) {
 	// to a single entry; passing the full run object preserves history.
 	originalHistoryLen := len(originalRun.StateHistory)
 	assert.Greater(t, originalHistoryLen, 0)
-	assert.Greater(t, len(retriedRun.StateHistory), originalHistoryLen)
+	require.Greater(t, len(retriedRun.StateHistory), originalHistoryLen)
 	lastEntry := retriedRun.StateHistory[len(retriedRun.StateHistory)-1]
 	assert.Equal(t, model.RuntimeStateRunning, lastEntry.State)
 }
--- a/backend/src/apiserver/resource/resource_manager_test.go
+++ b/backend/src/apiserver/resource/resource_manager_test.go
@@ -2583,6 +2583,30 @@ func TestRetryRun(t *testing.T) {
 	assert.Equal(t, actualRunDetail.RunDetails.State, model.RuntimeStateRunning)
 }
 
+func TestRetryRun_PreservesRunFields(t *testing.T) {
+	store, manager, runDetail := initWithOneTimeFailedRun(t)
+	defer store.Close()
+
+	originalRun, err := manager.GetRun(runDetail.UUID)
+	assert.Nil(t, err)
+	assert.Equal(t, string(v1alpha1.WorkflowFailed), string(originalRun.Conditions))
+
+	err = manager.RetryRun(context.Background(), runDetail.UUID)
+	assert.Nil(t, err)
+
+	retriedRun, err := manager.GetRun(runDetail.UUID)
+	assert.Nil(t, err)
+
+	// Core identifying fields must be preserved after retry
+	assert.Equal(t, originalRun.UUID, retriedRun.UUID)
+	assert.Equal(t, originalRun.DisplayName, retriedRun.DisplayName)
+	assert.Equal(t, originalRun.ExperimentId, retriedRun.ExperimentId)
+	// FinishedAtInSec must be reset to 0 on retry
+	assert.Equal(t, int64(0), retriedRun.RunDetails.FinishedAtInSec)
+	// State must be updated to Running
+	assert.Equal(t, model.RuntimeStateRunning, retriedRun.RunDetails.State)
+}
+
 func TestRetryRun_RunNotExist(t *testing.T) {
 	store := NewFakeClientManagerOrFatal(util.NewFakeTimeForEpoch())
 	defer store.Close()
--- a/backend/src/apiserver/resource/resource_manager_test.go
+++ b/backend/src/apiserver/resource/resource_manager_test.go
@@ -2605,6 +2605,15 @@ func TestRetryRun_PreservesRunFields(t *testing.T) {
 	assert.Equal(t, int64(0), retriedRun.FinishedAtInSec)
 	// State must be updated to Running
 	assert.Equal(t, model.RuntimeStateRunning, retriedRun.State)
+
+	// StateHistory must preserve pre-retry entries and append a new RUNNING entry.
+	// With the old sparse-struct UpdateRun call, StateHistory would be overwritten
+	// to a single entry; passing the full run object preserves history.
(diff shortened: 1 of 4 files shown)


frontend/src/pages/RunDetailsV2.test.tsx
```diff
@@ -403,6 +403,78 @@ describe('RunDetailsV2', () => {
...
+        <CommonTestWrapper>
+          <RunDetailsV2
+            pipeline_job={v2YamlTemplateString}
+            run={runWithHistory}
+            {...generateProps()}
+          ></RunDetailsV2>
+        </CommonTestWrapper>,
+      );
+
+      userEvent.click(screen.getByText('Detail'));
```
jeffspahr: These new cases call async `userEvent.click(...)` without `await`, so the assertions can run before the tab switches to Detail. Locally `npm run test:ui -- src/pages/RunDetailsV2.test.tsx` fails in the first two new tests for that reason, and the third passes vacuously before the tab change. Please `await` these clicks so the coverage actually exercises the Detail tab.

backend/src/apiserver/resource/resource_manager_test.go
```diff
@@ -2378,6 +2378,30 @@ func TestRetryRun(t *testing.T) {
...
+	assert.Nil(t, err)
+	assert.Equal(t, string(v1alpha1.WorkflowFailed), string(originalRun.Conditions))
+
+	err = manager.RetryRun(context.Background(), runDetail.UUID)
+	assert.Nil(t, err)
+
+	retriedRun, err := manager.GetRun(runDetail.UUID)
+	assert.Nil(t, err)
+
+	// Core identifying fields must be preserved after retry
```
jeffspahr, line 2600: This test still doesn't lock in the backend regression this patch fixes. `RunStore.UpdateRun` only writes `run_details` columns, so asserting `UUID`, `DisplayName`, and `ExperimentId` would pass even with the old sparse update. The behavior to protect here is preserved/appended `StateHistory`; please assert that the pre-retry history remains and a new `RUNNING` entry is appended.

jeffspahr (changes requested):
Two issues need follow-up before this is ready:

- the new frontend tests race the Detail tab because the async `userEvent.click(...)` calls are not awaited, which leaves the added coverage failing or vacuous under the current Vitest/React setup
- the new backend retry test does not assert the behavior this patch actually fixes; it should verify that `state_history` is preserved and a new `RUNNING` transition is appended

Verification I ran:
- `go test ./backend/src/apiserver/resource -run 'TestRetryRun|TestRetryRun_PreservesRunFields' -count=1`\n- `go test ./backend/src/apiserver/storage -run 'TestGetRun|TestUpdateRun' -count=1`\n- `cd frontend && fnm exec --using .nvmrc -- npm run test:ui -- src/pages/RunDetailsV2.test.tsx`

jeffspahr (approved):
Re-reviewing the current head `dc489b6124de1b9bba8b094e8db89a1102c5b931`: the previously reported blockers are resolved.

- the new `RunDetailsV2` cases now await the Detail-tab clicks, so the retry-start coverage executes the intended UI path
- `TestRetryRun_PreservesRunFields` now checks that `StateHistory` is preserved and a new `RUNNING` entry is appended

Focused verification on this head:
- `go test ./backend/src/apiserver/resource -run 'TestRetryRun|TestRetryRun_PreservesRunFields' -count=1`\n- `cd frontend && fnm exec --using .nvmrc -- npm ci --ignore-scripts`\n- `cd frontend && fnm exec --using .nvmrc -- npm run test:ui -- src/pages/RunDetailsV2.test.tsx`

backend/src/apiserver/resource/resource_manager_test.go
```diff
@@ -2583,6 +2583,39 @@ func TestRetryRun(t *testing.T) {
 	assert.Equal(t, actualRunDetail.RunDetails.State, model.RuntimeStateRunning)
 }
 
+func TestRetryRun_PreservesRunFields(t *testing.T) {
+	store, manager, runDetail := initWithOneTimeFailedRun(t)
+	defer store.Close()
+
+	originalRun, err := manager.GetRun(runDetail.UUID)
+	assert.Nil(t, err)
```
hbelmiro:
```suggestion
	require.Nil(t, err)
```

backend/src/apiserver/resource/resource_manager_test.go
```diff
@@ -2583,6 +2583,39 @@ func TestRetryRun(t *testing.T) {
...
+	assert.Equal(t, string(v1alpha1.WorkflowFailed), string(originalRun.Conditions))
+
+	err = manager.RetryRun(context.Background(), runDetail.UUID)
+	assert.Nil(t, err)
```
hbelmiro:
```suggestion
	require.Nil(t, err)
```

backend/src/apiserver/resource/resource_manager_test.go
```diff
@@ -2583,6 +2583,39 @@ func TestRetryRun(t *testing.T) {
...
+
+	retriedRun, err := manager.GetRun(runDetail.UUID)
+	assert.Nil(t, err)
```
hbelmiro:
```suggestion
	require.Nil(t, err)
```

backend/src/apiserver/resource/resource_manager_test.go
```diff
@@ -2583,6 +2583,39 @@ func TestRetryRun(t *testing.T) {
...
+	assert.Equal(t, int64(0), retriedRun.FinishedAtInSec)
+	// State must be updated to Running
+	assert.Equal(t, model.RuntimeStateRunning, retriedRun.State)
+
+	// StateHistory must preserve pre-retry entries and append a new RUNNING entry.
+	// With the old sparse-struct UpdateRun call, StateHistory would be overwritten
+	// to a single entry; passing the full run object preserves history.
+	originalHistoryLen := len(originalRun.StateHistory)
+	assert.Greater(t, originalHistoryLen, 0)
+	assert.Greater(t, len(retriedRun.StateHistory), originalHistoryLen)
```
hbelmiro:
```suggestion
	require.Greater(t, len(retriedRun.StateHistory), originalHistoryLen)
```

Sridhar1030:
Hey @hbelmiro sorry for the late reply , pushed the review changes

hbelmiro (approved):
/ok-to-test
/lgtm
/approve

Sridhar1030:
Hi @hbelmiro, applied your review suggestions (switched to `require.Nil`/`require.Greater` for error guards). Had to rebase to pick up the Go 1.26.2 bump from master which removed the `lgtm` label  could you re-add `/lgtm` when you get a chance? Thanks!

hbelmiro (commented):
/lgtm
/approve

hbelmiro:
@jeffspahr can you please approve?

Issue #12895, opened 2026-02-25
chore(frontend): Modernize JSX runtime and test utilities


manaswinidas:
## Chore description

Switch to modern JSX transform (`react-jsx` in tsconfig). Upgrade `@testing-library/react` from v11 to v14 and `@testing-library/user-event` from v13 to v14. Remove `react-test-renderer` and `@types/react-test-renderer`.

Depends on #12893 #12891 #12892 #12894 

Migrate 5 files from `react-test-renderer` to `render()` + `asFragment()`:
- `src/atoms/BusyButton.test.tsx`
- `src/atoms/Hr.test.tsx`
- `src/atoms/IconWithTooltip.test.tsx`
- `src/components/viewers/ROCCurve.test.tsx`
- `src/components/viewers/VisualizationCreator.test.tsx`

Migrate remaining `act` imports from `react-dom/test-utils` to `@testing-library/react` in 5 files:
- `src/mlmd/LineageActionBar.test.tsx`
- `src/components/viewers/VisualizationCreator.test.tsx`
- `src/components/viewers/Tensorboard.test.tsx`
- `src/components/UploadPipelineDialog.test.tsx`
- `src/components/SideNav.test.tsx`

Capture coverage baseline before and compare after -- coverage must not decrease.

**Acceptance Criteria**:
- [ ] `npm run test:ci` passes with zero deprecation warnings
- [ ] No file imports `React` solely for JSX (unused import)
- [ ] Line/branch coverage not decreased from baseline

### Labels

 /area frontend

---

Love this idea? Give it a 👍.

kanishka-commits:
I want to work on this issue.

PR #13025, opened 2026-03-13
chore(frontend): upgrade markdown-to-jsx, react-dropzone, react-texta…

### LINKED ISSUE #12896: chore(frontend): Update remaining ecosystem dependencies
## Chore description

Upgrade remaining deps with React 17-limited peer ranges: `markdown-to-jsx` v6 to v7, `react-dropzone` v5 to v14, any `re-resizable` residuals, `snapshot-diff` update. Review and update `package.json` `overrides` and `resolutions` entries -- remove stale ones, add new ones as needed. Drive peer gate allowlist to empty.

Depends on #12895 and here's the checklist/roadmap: https://github.com/kubeflow/pipelines/blob/master/frontend/docs/react-18-19-upgrade-checklist.md

Known direct peer-gate blockers to clear in this issue (React 18 target):**
- `react-textarea-autosize@8.3.3` (`react=^16.8.0 || ^17.0.0`)
- Note: `react-dom` / `react-test-renderer` major bumps are handled in #12897 and #12895 

**Acceptance Criteria**:
- [ ] `npm run check:react-peers` passes with empty allowlist
- [ ] `npm run check:react-peers:18` direct blockers are reduced to only planned React core bump items
- [ ] `npm run test:ci && npm run build` pass
- [ ] Markdown rendering and file upload work identically to pre-migration

### Labels

 /area frontend

---

Love this idea? Give it a 👍.

PR3MM:
Upgrades three frontend dependencies to widen React peer ranges as part of the React 18/19 migration track (checklist: `frontend/docs/react-18-19-upgrade-checklist.md`).

### Changes

* **`markdown-to-jsx`:** `v6` → `v7`
  * Removes `@types/markdown-to-jsx` (v7 ships its own types).
* **`react-dropzone`:** `v5` → `v14`
  * Adds `src/atoms/DropzoneArea.tsx` — a thin functional wrapper using `forwardRef` + `useImperativeHandle` to bridge the hook-based v14 API with existing class components without rewriting them.
* **`react-textarea-autosize`:** `8.3.3` → `8.5.9`
  * Peer range extended to include React 18.

### Fixes included

* **`Description.tsx`:** Restores v6 `forceInline` behavior. Overrides `em`/`strong` so `*text*` renders as plain text instead of `<em>` (`markdown-to-jsx` v7 regression).
* **`_onDrop`:** Guards against empty `acceptedFiles` to prevent a crash and stuck overlay when a dropped file is rejected by the `accept` filter.
* **`NewRun.test.tsx`:** Adds `types: ['Files']` to `fireEvent.drop` mock required by `react-dropzone` v14 in JSDOM.

### Peer gate

* **React 17 (must pass with empty allowlist):** `check:react-peers` → **PASS**
* **React 18 (only planned core bumps remain):** `check:react-peers:18`

### Manual Testing / Screenshots

* [x] **Getting Started page:** `markdown-to-jsx` v7 renders correctly.
  
SCR-20260313-lhwb

* [x] **New Pipeline:** File upload working .
  
  
SCR-20260313-lfvk

SCR-20260313-lfyr

Test results: npm run test:ci — 131 UI test files (1777 tests) + 12 server test files (177 tests), all passing.

Closes #12896 Depends on #12895

### DIFF
--- a/frontend/package-lock.json
+++ b/frontend/package-lock.json
@@ -103,6 +103,9 @@
         "vite": "^7.3.1",
         "vitest": "^4.0.17",
         "yaml": "^2.2.2"
+      },
+      "engines": {
+        "node": "24.14.0"
       }
     },
     "node_modules/@adobe/css-tools": {
@@ -151,7 +154,6 @@
       "version": "7.28.5",
       "dev": true,
       "license": "MIT",
-      "peer": true,
       "dependencies": {
         "@babel/code-frame": "^7.27.1",
         "@babel/generator": "^7.28.5",
@@ -648,7 +650,6 @@
         }
       ],
       "license": "MIT",
-      "peer": true,
       "engines": {
         "node": ">=18"
       },
@@ -670,7 +671,6 @@
         }
       ],
       "license": "MIT",
-      "peer": true,
       "engines": {
         "node": ">=18"
       }
@@ -1793,7 +1793,6 @@
     "node_modules/@mui/material": {
       "version": "5.18.0",
       "license": "MIT",
-      "peer": true,
       "dependencies": {
         "@babel/runtime": "^7.23.9",
         "@mui/core-downloads-tracker": "^5.18.0",
@@ -2759,7 +2758,6 @@
       "version": "8.7.0",
       "dev": true,
       "license": "MIT",
-      "peer": true,
       "dependencies": {
         "@babel/code-frame": "^7.10.4",
         "@babel/runtime": "^7.12.5",
@@ -3178,7 +3176,6 @@
       "version": "4.17.21",
       "devOptional": true,
       "license": "MIT",
-      "peer": true,
       "dependencies": {
         "@types/body-parser": "*",
         "@types/express-serve-static-core": "^4.17.33",
@@ -3270,7 +3267,6 @@
     "node_modules/@types/node": {
       "version": "20.19.28",
       "license": "MIT",
-      "peer": true,
       "dependencies": {
         "undici-types": "~6.21.0"
       }
@@ -3308,7 +3304,6 @@
     "node_modules/@types/react": {
       "version": "17.0.91",
       "license": "MIT",
-      "peer": true,
       "dependencies": {
         "@types/prop-types": "*",
         "@types/scheduler": "^0.16",
@@ -3319,7 +3314,6 @@
       "version": "17.0.26",
       "devOptional": true,
       "license": "MIT",
-      "peer": true,
       "peerDependencies": {
         "@types/react": "^17.0.0"
       }
@@ -3473,7 +3467,6 @@
       "version": "5.62.0",
       "dev": true,
       "license": "BSD-2-Clause",
-      "peer": true,
       "dependencies": {
         "@typescript-eslint/scope-manager": "5.62.0",
         "@typescript-eslint/types": "5.62.0",
@@ -3910,7 +3903,6 @@
     "node_modules/@xyflow/system/node_modules/d3-selection": {
       "version": "3.0.0",
       "license": "ISC",
-      "peer": true,
       "engines": {
         "node": ">=12"
       }
@@ -3965,7 +3957,6 @@
       "version": "8.15.0",
       "dev": true,
       "license": "MIT",
-      "peer": true,
       "bin": {
         "acorn": "bin/acorn"
       },
@@ -4475,7 +4466,6 @@
       "version": "4.16.5",
       "dev": true,
       "license": "MIT",
-      "peer": true,
       "dependencies": {
         "caniuse-lite": "^1.0.30001214",
         "colorette": "^1.2.2",
@@ -5652,7 +5642,6 @@
       "dev": true,
       "hasInstallScript": true,
       "license": "MIT",
-      "peer": true,
       "bin": {
         "esbuild": "bin/esbuild"
       },
@@ -5700,7 +5689,6 @@
       "version": "8.57.1",
       "dev": true,
       "license": "MIT",
-      "peer": true,
       "dependencies": {
         "@eslint-community/eslint-utils": "^4.2.0",
         "@eslint-community/regexpp": "^4.6.1",
@@ -5892,7 +5880,6 @@
       "version": "2.32.0",
       "dev": true,
       "license": "MIT",
-      "peer": true,
       "dependencies": {
         "@rtsao/scc": "^1.1.0",
         "array-includes": "^3.1.9",
@@ -7329,7 +7316,6 @@
     "node_modules/immer": {
       "version": "9.0.6",
       "license": "MIT",
-      "peer": true,
       "funding": {
         "type": "opencollective",
         "url": "https://opencollective.com/immer"
@@ -8886,7 +8872,6 @@
         }
       ],
       "license": "MIT",
-      "peer": true,
       "dependencies": {
         "nanoid": "^3.3.11",
         "picocolors": "^1.1.1",
@@ -8993,7 +8978,6 @@
       "version": "3.8.1",
       "dev": true,
       "license": "MIT",
-      "peer": true,
       "bin": {
         "prettier": "bin/prettier.cjs"
       },
@@ -9158,7 +9142,6 @@
     "node_modules/react": {
       "version": "17.0.2",
       "license": "MIT",
-      "peer": true,
       "dependencies": {
         "loose-envify": "^1.1.0",
         "object-assign": "^4.1.1"
@@ -9224,7 +9207,6 @@
     "node_modules/react-dom": {
       "version": "17.0.2",
       "license": "MIT",
-      "peer": true,
       "dependencies": {
         "loose-envify": "^1.1.0",
         "object-assign": "^4.1.1",
@@ -9253,8 +9235,7 @@
     },
     "node_modules/react-is": {
       "version": "16.13.1",
-      "license": "MIT",
-      "peer": true
+      "license": "MIT"
     },
     "node_modules/react-lifecycles-compat": {
       "version": "3.0.4",
@@ -9472,7 +9453,6 @@
     "node_modules/recharts/node_modules/react-redux": {
       "version": "8.1.3",
       "license": "MIT",
-      "peer": true,
       "dependencies": {
         "@babel/runtime": "^7.12.1",
         "@types/hoist-non-react-statics": "^3.3.1",
@@ -9513,8 +9493,7 @@
     },
     "node_modules/recharts/node_modules/redux": {
       "version": "5.0.1",
-      "license": "MIT",
-      "peer": true
+      "license": "MIT"
     },
     "node_modules/recharts/node_modules/redux-thunk": {
       "version": "3.1.0",
@@ -10162,7 +10141,6 @@
       "version": "10.2.13",
       "dev": true,
       "license": "MIT",
-      "peer": true,
       "dependencies": {
         "@storybook/global": "^5.0.0",
         "@storybook/icons": "^2.0.1",
@@ -10732,7 +10710,6 @@
       "version": "4.0.3",
       "dev": true,
       "license": "MIT",
-      "peer": true,
       "engines": {
         "node": ">=12"
       },
@@ -10797,7 +10774,6 @@
       "version": "10.9.2",
       "dev": true,
       "license": "MIT",
-      "peer": true,
       "dependencies": {
         "@cspotcode/source-map-support": "^0.8.0",
         "@tsconfig/node10": "^1.0.7",
@@ -11079,7 +11055,6 @@
       "version": "4.9.5",
       "dev": true,
       "license": "Apache-2.0",
-      "peer": true,
       "bin": {
         "tsc": "bin/tsc",
         "tsserver": "bin/tsserver"
@@ -11405,7 +11380,6 @@
       "version": "7.3.1",
       "dev": true,
       "license": "MIT",
-      "peer": true,
       "dependencies": {
         "esbuild": "^0.27.0",
         "fdir": "^6.5.0",
@@ -11495,7 +11469,6 @@
       "version": "4.0.3",
       "dev": true,
       "license": "MIT",
-      "peer": true,
       "engines": {
         "node": ">=12"
       },
@@ -11550,7 +11523,6 @@
       "version": "4.0.17",
       "dev": true,
       "license": "MIT",
-      "peer": true,
       "dependencies": {
         "@vitest/expect": "4.0.17",
         "@vitest/mocker": "4.0.17",
@@ -11910,7 +11882,6 @@
       "version": "2.8.2",
       "dev": true,
       "license": "ISC",
-      "peer": true,
       "bin": {
         "yaml": "bin.mjs"
       },


frontend/src/atoms/DropzoneArea.tsx
```diff
@@ -0,0 +1,85 @@
...
+    disabled,
+    onDragEnter: (_e) => onDragEnter?.(),
+    onDragLeave: (_e) => onDragLeave?.(),
+  });
+
+  React.useImperativeHandle(ref, () => ({ open }), [open]);
+
+  return (
+    <div {...getRootProps()} id={id} style={style} data-testid={dataTestId} aria-label={ariaLabel}>
+      <input {...getInputProps()} {...inputProps} />
```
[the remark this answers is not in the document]
PR3MM: Fixed. Switched to getInputProps(inputProps) and getRootProps({ id, style, 'data-testid': dataTestId }) so react-dropzone merges the props internally rather than relying on spread order.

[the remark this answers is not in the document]
PR3MM: Fixed. Moved aria-label from the root <div> (which has role="presentation" and is excluded from the accessibility tree) to the <input> element, where it will be correctly announced by screen readers.

frontend/src/components/Description.tsx
```diff
@@ -34,6 +34,13 @@ const options = {
 const optionsForceInline = {
   ...options,
   forceInline: true,
+  overrides: {
+    ...options.overrides,
+    // markdown-to-jsx v7 parses `*...*` as <em> even in forceInline mode.
+    // Strip the tags so list-marker asterisks render as plain text, matching v6 behavior.
```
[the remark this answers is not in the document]
PR3MM: Fixed. Updated the comment to accurately say "inline emphasis renders as plain text (no italic/bold)" rather than "list-marker asterisks", which better reflects what the em/strong overrides actually do.

frontend/src/pages/NewPipelineVersion.tsx
```diff
@@ -399,16 +399,19 @@ export class NewPipelineVersion extends Page<NewPipelineVersionProps, NewPipelin
...
               onDragEnter={this._onDropzoneDragEnter.bind(this)}
               onDragLeave={this._onDropzoneDragLeave.bind(this)}
               style={{ position: 'relative' }}
               ref={this._dropzoneRef}
               inputProps={{ tabIndex: -1 }}
-              accept='.yaml,.yml,.zip,.tar.gz'
+              accept={{
+                'application/yaml': ['.yaml', '.yml'],
+                'application/zip': ['.zip'],
+                'application/gzip': ['.tar.gz'],
```
manaswinidas: `attr-accept` v2 (used by react-dropzone v14) matches against the **last** extension segment. A file named `pipeline.tar.gz` has extension `.gz`, not `.tar.gz`. This mapping will silently reject valid `.tar.gz` uploads — a functional regression from the v5 string-based `accept`.

```suggestion
                'application/gzip': ['.gz', '.tar.gz'],
```

frontend/src/atoms/DropzoneArea.tsx
```diff
@@ -0,0 +1,85 @@
...
+    disabled,
+    onDragEnter: (_e) => onDragEnter?.(),
+    onDragLeave: (_e) => onDragLeave?.(),
+  });
+
+  React.useImperativeHandle(ref, () => ({ open }), [open]);
+
+  return (
+    <div {...getRootProps({ id, style, 'data-testid': dataTestId })}>
+      <input aria-label={ariaLabel} {...getInputProps(inputProps)} />
```
manaswinidas: `aria-label` is set *before* the `getInputProps(inputProps)` spread. If `getInputProps` ever returns an `aria-label`, or `inputProps` contains one, it silently overrides the explicit value. Move it after the spread, or pass it through `getInputProps`:

```suggestion
      <input {...getInputProps({ ...inputProps, 'aria-label': ariaLabel })} />
```

frontend/src/atoms/DropzoneArea.tsx
```diff
@@ -0,0 +1,85 @@
...
+ * directly hold a ref of this type and call open() to trigger the file picker.
+ */
+export interface DropzoneAreaHandle {
+  open: () => void;
+}
+
+interface DropzoneAreaProps {
+  onDrop: (files: File[]) => void;
+  onDragEnter?: () => void;
+  onDragLeave?: () => void;
```
manaswinidas, line 32: The `onDrop` prop signature only accepts accepted files. In v14, `onDrop` also receives `FileRejection[]` as the second argument. Both consumers silently swallow rejected drops (empty-array guard, no feedback). Consider either:
- Exposing `onDropRejected` as a prop, or
- Accepting the second `rejected` parameter here so consumers can show an error when a file is filtered out by `accept`.

frontend/src/components/UploadPipelineDialog.tsx
```diff
@@ -253,6 +252,10 @@ class UploadPipelineDialog extends React.Component<
   }
 
   private _onDrop(files: File[]): void {
+    if (!files.length) {
```
manaswinidas: DRY: This empty-array guard is duplicated identically in `NewPipelineVersion._onDrop`. Since `DropzoneArea` is the new shared wrapper, it should handle this internally — e.g. by using `onDropAccepted` instead of `onDrop` from the hook, or by filtering before calling the prop callback. That way every consumer gets the guard for free.

frontend/src/pages/__snapshots__/GettingStarted.test.tsx.snap
```diff
@@ -46,20 +46,15 @@ exports[`GettingStarted page > initially renders documentation 1`] = `
...
-            </li>
-          </ul>
+           - Shows how to pass data between Python components. 
+          <a
+            class="link_f1fk43bf"
+            href="https://github.com/kubeflow/pipelines/tree/master/samples/tutorials/Data%20passing%20in%20python%20components"
+            rel="noopener"
+            target="_blank"
+          >
+            source code
```
manaswinidas: The nested markdown list (`- Shows how to pass data...`) no longer renders as `<ul><li>` HTML — it collapses to literal text with a dash. This is a visible rendering regression on the Getting Started page. Was this intentional? The v7 parser handles nested lists differently; you may need a `wrapper` or `overrides` config similar to the `forceInline` fix in `Description.tsx`.

frontend/src/atoms/DropzoneArea.tsx
manaswinidas, line 1: nit: New component with no unit tests. At minimum, testing the imperative `open()` handle via ref and verifying `onDrop`/`onDragEnter`/`onDragLeave` fire would be valuable — especially since this is a shared wrapper that two consumers depend on.

frontend/src/pages/NewPipelineVersion.tsx
PR3MM: Added .gz alongside .tar.gz

frontend/src/atoms/DropzoneArea.tsx
PR3MM: Moved aria-label inside the getInputProps spread as suggested

frontend/src/atoms/DropzoneArea.tsx
PR3MM: Exposed onDropRejected as a prop on DropzoneArea and wired it to the hook. Both consumers now handle rejections with user-facing feedback ,UploadPipelineDialog shows an inline error, NewPipelineVersion shows a snackbar.

frontend/src/components/UploadPipelineDialog.tsx
PR3MM: Switched the hook from onDrop to onDropAccepted, which guarantees non-empty arrays. Removed the guards from both consumers.

frontend/src/atoms/DropzoneArea.tsx
PR3MM: Added in 9258e6e. 7 unit tests covering: rendering, imperative open() ref handle, onDrop via onDropAccepted, onDropRejected, aria-label forwarding, and inputProps passthrough.

frontend/src/pages/__snapshots__/GettingStarted.test.tsx.snap
PR3MM: Fixed in d6e47cb.

frontend/src/components/UploadPipelineDialog.tsx
```diff
@@ -83,12 +89,21 @@ interface UploadPipelineDialogState {
...
 }
 
+export const PIPELINE_PACKAGE_ACCEPT = {
+  'application/yaml': ['.yaml', '.yml'],
+  'application/zip': ['.zip'],
+  'application/gzip': ['.gz', '.tar.gz'],
+};
+
+export const PIPELINE_PACKAGE_REJECT_MESSAGE =
+  'Invalid file type. Supported formats: .yaml, .yml, .zip, .gz, .tar.gz';
```
jeffspahr: Allowing plain `.gz` here creates a frontend/backend contract mismatch. The UI now accepts and advertises `.gz`, but the backend upload path still only supports `.tar.gz`, `.zip`, `.json`, or YAML (`backend/src/apiserver/server/util.go`). That means a user can select `foo.gz`, pass frontend validation, and then fail server-side with `Unexpected pipeline file format`. Please keep the `.tar.gz` workaround without claiming generic `.gz` support, and add a regression test for the accepted extensions.

PR3MM: Fixed — removed .gz from the accept list and reject message. The config now uses application/gzip: ['.tar.gz'] (no plain .gz), with a useDropzone validator that rejects .gz and .tgz files that sneak through the shared application/gzip MIME check. Also added regression tests pinning the accepted extensions. See 
0b2b214.

jeffspahr:
I found one remaining regression in the `Description` workaround for `markdown-to-jsx` v7.

`forceInline` now strips legitimate emphasis because `optionsForceInline` overrides both `em` and `strong` to render plain children. On `master` with `markdown-to-jsx` v6, `Description` in inline mode still renders `*abc*` as emphasis and `**abc**` as strong text. On this PR head, both become plain text.

That matters because `forceInline` is used in description cells and forms, so bold/italic markdown disappears in places like pipeline/run descriptions.

I would fix this with the smallest possible change:

1. Remove the `em` and `strong` overrides from `optionsForceInline` in `frontend/src/components/Description.tsx`.
2. Keep `preserveListMarkers()` and continue applying it only when `forceInline` is true.
3. Add regression tests in `frontend/src/components/Description.test.tsx` for:
   - `*abc*` in `forceInline` renders emphasis
   - `**abc**` in `forceInline` renders strong text
   - list markers still render as visible bullets/text in `forceInline`

Reasoning: the list-marker issue is caused by inline parsing of lines like `* item`, and `preserveListMarkers()` already fixes that by converting only leading list markers to `•`. The `em`/`strong` overrides are what flatten valid markdown formatting.

I verified this locally against the PR branch: with `preserveListMarkers()` alone, `*abc*` renders as `<em>`, `**abc**` renders as `<strong>`, and `* abc` / `- abc` / `+ abc` still render as visible bullets in inline mode.

frontend/src/atoms/DropzoneArea.tsx
```diff
@@ -0,0 +1,95 @@
...
+  const { getRootProps, getInputProps, open } = useDropzone({
+    onDropAccepted: onDrop,
+    onDropRejected,
+    noClick: true,
+    accept,
+    disabled,
+    validator,
+    onDragEnter: (_e) => onDragEnter?.(),
+    onDragLeave: (_e) => onDragLeave?.(),
+  });
```
manaswinidas, line 85: `useDropzone` defaults to `multiple: true`, so the native file picker allows selecting several files. But both consumers (`UploadPipelineDialog._onDrop` and `NewPipelineVersion._onDrop`) only use `files[0]` — extra files are silently discarded.

Consider exposing a `multiple` prop (defaulting to `false`) and passing it to the hook here, or at least setting `multiple: false` since neither consumer supports multi-file upload.

frontend/src/pages/NewPipelineVersion.tsx
```diff
@@ -748,6 +755,17 @@ export class NewPipelineVersion extends Page<NewPipelineVersionProps, NewPipelin
       },
     );
   }
+
+  private _onDropRejected(): void {
+    this.setStateSafe({ dropzoneActive: false, file: null, fileName: '' }, () => {
```
manaswinidas, line 760: This clears `file` and `fileName` but doesn't reset `pipelineVersionName`. If a user drops a valid file (populating the version name), then drops an invalid one, the form shows the old version name with no file attached — an inconsistent state.

`UploadPipelineDialog._onDropRejected` already clears `uploadPipelineName` in the same scenario; this should clear `pipelineVersionName` too for consistency.

frontend/src/atoms/DropzoneArea.tsx
PR3MM: added a multiple prop to DropzoneAreaProps (defaults to false) and passed it through to useDropzone

frontend/src/pages/NewPipelineVersion.tsx
PR3MM: Fixed — pipelineVersionName: '' is now included in the setStateSafe call inside _onDropRejected, matching the pattern already used in UploadPipelineDialog

jeffspahr:
I found one remaining regression in the latest follow-up commit.

`NewPipelineVersion._onDropRejected()` now clears `pipelineVersionName` along with the file fields. That version name is not derived from the dropped file: for existing-pipeline version creation it is prefilled from the selected pipeline on load, and users can also edit it independently. If they drag or pick one invalid file, the required version name is wiped and they have to re-enter it before retrying, even though only the file input was invalid.

I would fix this with the smallest possible change:

1. In `frontend/src/pages/NewPipelineVersion.tsx`, remove `pipelineVersionName: ''` from `_onDropRejected()`.
2. Keep the reject handler limited to file-related state (`dropzoneActive`, `file`, `fileName`) and the snackbar.
3. Add a regression test in `frontend/src/pages/NewPipelineVersion.test.tsx` that sets a version name, triggers the reject path, and verifies the version name is preserved.

Reasoning: rejecting an invalid upload should reset only the invalid upload state. Clearing unrelated form fields makes the retry flow worse and is especially noticeable on the existing-pipeline version path where `pipelineVersionName` is required.

manaswinidas:
@PR3MM - can you fix the failing tests ?

jeffspahr:
Re-reviewed the current PR head `fa573a30a` and I do not have any remaining findings.

The two issues I called out earlier are now fixed:
- `Description.forceInline` preserves real emphasis/strong rendering again while keeping visible bullet markers
- `NewPipelineVersion` no longer clears `pipelineVersionName` on rejected file drops, and there is regression coverage for that path

Verification I ran locally on the rebased head:
- `fnm use 24.14.0 && npm ci`
- `fnm use 24.14.0 && npm run test:ui -- src/atoms/DropzoneArea.test.tsx src/components/Description.test.tsx src/components/UploadPipelineDialog.test.tsx src/pages/NewPipelineVersion.test.tsx src/pages/NewRun.test.tsx`
- `fnm use 24.14.0 && npm run build`

Results:
- targeted UI suite passed (`149/149`)
- production build passed
- `npm ci` completed cleanly under the repo's pinned Node `24.14.0`

PR #13026, opened 2026-03-13
ci(frontend): add lockfile drift check to prevent Prettier version mismatch

### LINKED ISSUE #13020: ci(frontend): add lockfile drift check to prevent Prettier version mismatch
## Problem

The `frontend-tests` CI workflow runs `npm ci` against the PR's own `package-lock.json`, then runs `prettier --check`. This means a PR can pass CI while using a completely different Prettier version than master, as long as the code is self-consistent with that version.

This was discovered in #12839, where the PR's lockfile pinned Prettier **1.19.1** while master has **3.8.1**. The default for `arrowParens` changed between these versions (`"avoid"` → `"always"`), so `npm run format` reformatted 171 unrelated files. CI passed because it installed 1.19.1 from the PR's lockfile and the formatting was self-consistent — but merging would introduce formatting that immediately fails against master's Prettier version.

## Suggested fix

Add a step to `.github/workflows/frontend.yml` that verifies the PR's `frontend/package-lock.json` hasn't drifted from master's. For example:

```yaml
- name: Check lockfile matches base branch
  run: |
    git fetch origin ${{ github.base_ref }} --depth=1
    if ! git diff --quiet origin/${{ github.base_ref }} -- frontend/package-lock.json; then
      echo "::error::frontend/package-lock.json has drifted from ${{ github.base_ref }}. Please rebase onto ${{ github.base_ref }} and run 'npm ci' to sync."
      exit 1
    fi
```

This would catch lockfile drift from any source (stale branch, accidental `npm install` that resolves different versions, etc.) and prevent formatting mismatches from slipping through CI.

## Context

- Master lockfile: Prettier 3.8.1
- PR #12839 lockfile: Prettier 1.19.1
- `.prettierrc.yaml` does not set `arrowParens`, so behavior depends entirely on Prettier version
- CI has no check that the lockfile matches the base branch

manaswinidas:
Adds a CI step that fails when package-lock.json drifts from the base branch without a corresponding package.json change. Also makes `arrowParens` explicit in `.prettierrc.yaml `so formatting behavior is version-independent.

Fixes #13020

**Checklist:**

### DIFF
--- a/.github/workflows/frontend.yml
+++ b/.github/workflows/frontend.yml
@@ -33,6 +33,7 @@ jobs:
       if: github.event_name == 'pull_request'
       run: |
         git fetch origin ${{ github.base_ref }} --depth=1
+        git fetch --deepen=100 2>/dev/null || true
         cd frontend && node scripts/check-lockfile-drift.mjs --base-ref "origin/${{ github.base_ref }}"
 
     - name: Clean npm cache
--- a/frontend/scripts/check-lockfile-drift.mjs
+++ b/frontend/scripts/check-lockfile-drift.mjs
@@ -15,6 +15,7 @@ const DEP_FIELDS = [
 
 const TOOLCHAIN_PACKAGES = ['prettier', 'eslint', 'typescript', 'vitest'];
 const TOOLCHAIN_PREFIXES = ['@typescript-eslint/'];
+const SILENT_STDIO = ['pipe', 'pipe', 'pipe'];
 
 function parseArgs(argv) {
   let baseRef = null;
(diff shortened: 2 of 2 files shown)


jeffspahr:
Thanks for addressing the earlier feedback. The current version fixes the main issues from the original review:

- the check no longer treats any edit anywhere in `frontend/package.json` as a dependency change
- the policy moved out of inline workflow shell into `frontend/scripts/check-lockfile-drift.mjs`
- lockfile-only PRs are no longer rejected by default; the script now focuses on CI-sensitive toolchain packages
- `arrowParens` is now pinned explicitly in `.prettierrc.yaml`

What still looks unresolved is the baseline used for `depsChanged`. The script currently compares dependency-bearing fields in `HEAD` against the current base branch, not against the merge base / actual PR diff. That means a stale branch can still be classified as an intentional dependency update just because `master` changed after the branch was cut.

So I think the remaining step is to key the intent check off the merge base (or otherwise determine whether this PR itself changed dependency-bearing fields), rather than only comparing `HEAD` to current `origin/${{ github.base_ref }}` semantics. That would preserve the improvements above while closing the stale-branch gap this check is meant to catch.

jeffspahr:
Re-reviewed the latest revision.

The earlier concern about treating any edit anywhere in `frontend/package.json` as a dependency change is addressed in the current implementation: the check now keys off dependency-bearing fields and uses the merge base for intent detection.

I do not have any remaining findings on the current diff.

/lgtm

PR #13037, opened 2026-03-14
chore(frontend): upgrade Node.js from 22.19.0 to 24.14.0

jsonmp-k8:
## Summary

- Upgrades Node.js from 22.19.0 (Maintenance LTS, EOL April 2026) to 24.14.0 (Active LTS, EOL May 2027)
- Fixes stale Node 18.x pin in frontend integration test Dockerfile
- Pins all Node version references to a single source of truth (`frontend/.nvmrc`)
- Updated files:
  - `frontend/.nvmrc`: v22.19.0 → v24.14.0
  - `frontend/Dockerfile`: node:22.19 → node:24.14.0 (both build and runtime stages, full patch pin)
  - `.github/workflows/frontend.yml`: `node-version: '22'` → `node-version-file: frontend/.nvmrc`
  - `frontend/CONTRIBUTING.md`: version references updated
  - `AGENTS.md`: version references updated
  - `test/frontend-integration-test/Dockerfile`: replaced NodeSource `setup_18.x` with direct nodejs.org binary install via `ARG NODE_VERSION`, defaulting to 24.14.0 and derivable from `.nvmrc` at build time
  - `.github/workflows/e2e-test-frontend.yml`: passes `--build-arg NODE_VERSION` derived from `frontend/.nvmrc` when building the integration test image
  - `test/frontend-integration-test/README.md`: updated local build instructions to show `--build-arg` derivation from `.nvmrc`

## Test plan

- [ ] `npm ci` succeeds with Node 24
- [ ] `npm run build` produces a valid production build
- [ ] `npm run test:ci` passes (format + lint + typecheck + React peer check + Vitest + Jest)
- [ ] `npm run check:react-peers` confirms React 17 peer compatibility
- [ ] Frontend Docker image builds and starts successfully
- [ ] Frontend integration test Docker image builds successfully
- [ ] Frontend UI loads and basic navigation works

PR #13038, opened 2026-03-14
chore(frontend): add engines field to package.json

jsonmp-k8:
## Summary

- Adds `"engines": { "node": ">=24.0.0 <25.0.0" }` to `frontend/package.json` to constrain to the Node 24 LTS line
- Adds `frontend/.npmrc` with `engine-strict=true` so `npm ci`/`npm install` fails (not just warns) on version mismatch
- Un-ignores `.npmrc` in `frontend/.gitignore` so the project config is tracked (personal overrides belong in `~/.npmrc`)
- Surfaces Node.js version mismatches early during `npm install` rather than at runtime

Depends on #13037 (Node 24 upgrade) being merged first.

## Test plan

- [ ] `npm ci` succeeds with Node 24
- [ ] `npm ci` fails with Node 22 (engines check enforced via `.npmrc`)
- [ ] `npm ci` fails with Node 25+ (upper bound constraint)
- [ ] CI frontend tests pass

### DIFF
--- a/frontend/package.json
+++ b/frontend/package.json
@@ -4,7 +4,7 @@
   "license": "Apache-2.0",
   "private": true,
   "engines": {
-    "node": ">=24.0.0 <25.0.0"
+    "node": "24.14.0"
   },
   "dependencies": {
     "@emotion/react": "^11.14.0",


frontend/package.json
```diff
@@ -3,6 +3,9 @@
   "version": "0.1.0",
   "license": "Apache-2.0",
   "private": true,
+  "engines": {
+    "node": ">=24.0.0 <25.0.0"
```
jeffspahr: The repo already standardizes frontend tooling on exact `v24.14.0` in `.nvmrc`, the frontend workflow, and the frontend Docker image, but this range still accepts any `24.x` runtime. With `engine-strict=true`, installs on `24.0.0` or a later `24.x` patch will still pass even though they do not match the pinned toolchain, so this does not actually catch the local/CI drift the PR says it is preventing. Consider matching the exact pin, or at least the pinned minor/patch line, instead of the whole major line.

jsonmp-k8: Good call — pinned `engines.node` to exactly `24.14.0` to match `.nvmrc`, Dockerfile, and CI. Also added `engine-strict=true` via `.npmrc` so `npm install` will hard-fail on a mismatch instead of just warning.

PR #13040, opened 2026-03-15
refactor(apiserver): make initConfig and registerHTTPHandlerFromEndpoint testable

jsonmp-k8:
## Summary

The `backend/src/apiserver` main package has 18.9% test coverage. The primary blocker is that most functions use `glog.Fatalf` for error handling, which calls `os.Exit(1)` and makes them untestable — error paths kill the test process.

This is the first step in a staged plan to raise coverage to ~80%. It refactors two helper functions to return errors instead of fataling:

- `initConfig()` now returns `error`; `main()` handles it with `glog.Fatalf`
- `registerHTTPHandlerFromEndpoint()` now returns `error`; `startHTTPProxy()` handles it via a `register` closure that also cleans up the 13 repeated `ctx`/`runtimeMux`/`tlsCfg` arguments
- Moves `ctx` to first parameter per Go convention
- Replaces deprecated `grpc.WithInsecure()` → `insecure.NewCredentials()`
- Logs config reload errors in `OnConfigChange` instead of silently swallowing them

New tests:
- `TestInitConfig`: success with temp config file, error with missing path (uses `errors.As` for robust error type checking)
- `TestRegisterHTTPHandlerFromEndpoint`: success without TLS, success with TLS, handler error propagation, endpoint uses configured RPC port

## Test plan

- [ ] `go test -v -cover ./backend/src/apiserver/` passes in CI (Linux)
- [ ] `go vet ./backend/src/apiserver/` passes (verified locally)
- [ ] `golangci-lint run --new --new-from-merge-base=origin/master` reports 0 issues (verified locally)
- [ ] No behavior change — callers still `glog.Fatalf` on errors


droctothorpe:
This and all the other test coverage work you've been doing is so impactful, @jsonmp-k8. Ty!

Issue #13042, opened 2026-03-16
chore(frontend): Add unit tests for MLMD detail pages (0% coverage on ArtifactDetails and ExecutionDetails)


PR3MM:
## Chore description

`ArtifactDetails.tsx` and `ExecutionDetails.tsx` are the MLMD (ML Metadata) detail pages that display artifact and execution information, including lineage navigation and cross-references to pipeline runs. Neither file has any test coverage. Error states (invalid IDs, missing data, service failures), loading states, and tab navigation are all untested. A regression in these pages would break the artifact/execution inspection experience without any test catching it.

`ExecutionDetails.tsx` contains 5 distinct testable pieces: a pure function (`parseEventsByType`), two sub-components (`SectionIO`, `ArtifactRow`), the main content component (`ExecutionDetailsContent`), and a reference section (`ExecutionReference`).

### Scope

| File | Status | Work |
|------|--------|------|
| `ArtifactDetails.tsx` | Zero tests | Write from scratch |
| `ExecutionDetails.tsx` | Zero tests | Write from scratch (5 testable pieces) |

#### New files

- `src/pages/ArtifactDetails.test.tsx`
- `src/pages/ExecutionDetails.test.tsx`

### Test cases (~25 total)

**ArtifactDetails (~9)**

- [ ] shows `CircularProgress` spinner while artifact is loading
- [ ] renders Overview tab with `ResourceInfo` after artifact loads
- [ ] updates toolbar title with artifact name after load
- [ ] shows page error banner when no artifact found for the given ID
- [ ] shows page error banner when multiple artifacts found for the given ID
- [ ] shows page error banner on service error
- [ ] navigates to lineage URL when Lineage Explorer tab is clicked
- [ ] navigates back to overview URL when Overview tab is clicked
- [ ] renders a new `ArtifactDetails` instance when artifact ID in URL changes

**ExecutionDetails (~16)**

`parseEventsByType` (pure function, zero mocking needed)

- [ ] returns all empty arrays when response is null
- [ ] groups events correctly into INPUT/OUTPUT/DECLARED_INPUT/DECLARED_OUTPUT buckets
- [ ] ignores events where type or artifactId is null

`SectionIO`

- [ ] renders null when events array is empty
- [ ] renders a table row for each event when events are present

`ArtifactRow`

- [ ] renders artifact name as a link when artifact ID exists
- [ ] renders plain name when artifact ID is zero/missing

`ExecutionDetailsContent`

[...]

PR #13045, opened 2026-03-16
chore(frontend): Add unit tests for MLMD detail pages (0% coverage on ArtifactDetails and ExecutionDetails) #13042

kanishka-commits:
## Description

FIXES:  https://github.com/kubeflow/pipelines/issues/13042

Add unit tests for the `ArtifactDetails.tsx` and `ExecutionDetails.tsx` MLMD (ML Metadata) detail pages. These pages display artifact and execution information including lineage navigation and cross-references to pipeline runs. Neither file had any test coverage — error states, loading states, and tab navigation were all untested. A regression in these pages would break the artifact/execution inspection experience without any test catching it.

`ExecutionDetails.tsx` contains 5 distinct testable pieces: a pure function (`parseEventsByType`), two sub-components (`SectionIO`, `ArtifactRow`), the main content component (`ExecutionDetailsContent`), and a reference section (`ExecutionReference`).

### New files

- `src/pages/ArtifactDetails.test.tsx`
- `src/pages/ExecutionDetails.test.tsx`

### `ArtifactDetails.test.tsx` (10 tests)

- **Loading state:** Verifies `CircularProgress` spinner renders while artifact data is being fetched.
- **Successful render:** Confirms Overview and Lineage Explorer tabs appear, `ResourceInfo` renders with artifact type, and toolbar title updates with artifact name (including version suffix when the artifact has a version property).
- **Error states:** Covers three failure paths — no artifact found for the given ID, multiple artifacts found for the same ID, and service/network error. Each verifies the correct error banner is shown.
- **Tab navigation:** Clicking "Lineage Explorer" pushes the `/lineage` URL suffix; clicking "Overview" navigates back to the base artifact URL.
- **Remount on ID change:** Uses a controlled `createMemoryHistory` to navigate from `/artifacts/1` to `/artifacts/2` within the same router, verifying the `key` prop forces a full component remount and fresh data load.

### `ExecutionDetails.test.tsx` (19 tests)

- **`parseEventsByType`** (3 tests) — Pure function, zero mocking. Tests null response returns empty buckets, events are grouped correctly into `INPUT`/`OUTPUT`/`DECLARED_INPUT`/`DECLARED_OUTPUT` buckets, and events with default protobuf type (`UNKNOWN`/`0`) land in the `UNKNOWN` bucket.
- **`ExecutionDetailsContent`** (8 tests) — Loading spinner while data is fetched. Validation errors for invalid IDs (`NaN`, negative). Error banners when no execution found, multiple executions found, or execution type cannot be resolved. Successful render verifies `ResourceInfo` displays the execution type name and IO section headers appear when events are present.
- **`SectionIO`** (2 tests) — Returns `null` when the events array is empty (no section header rendered). When events are present, verifies the correct number of table rows per section using `within()` scoped queries.
- **`ArtifactRow`** (2 tests) — Artifact name renders as a `<Link>` to the artifact details page when artifact ID is non-zero. Renders plain text (no link) when artifact ID is zero, verified via `queryAllByRole('link')`.
- **`ExecutionReference`** (3 tests) — Renders a pipeline run link when `getContextByExecution` returns a context. Renders an original execution link when the execution has a `cached_execution_id` custom property. Renders neither link when no context is found and no cached ID exists.
- **Page wrapper** (1 test) — Verifies `ExecutionDetails` renders with initial toolbar state using `TestUtils.generatePageProps`.

### Overall Coverage (All files)

| Metric     | Coverage |
|------------|----------|
| Statements | 73.11%   |
| Branches   | 65.38%   |
| Functions  | 70.86%   |
| Lines      | 73.4%    |

Screenshot 2026-03-16 at 7 14 39 PM

### Source change

Export `parseEventsByType` from `ExecutionDetails.tsx` (single keyword addition: `function` → `export function`) so tests can import and test the real implementation instead of duplicating the logic.

**Checklist:**

### DIFF
--- a/frontend/src/pages/ExecutionDetails.test.tsx
+++ b/frontend/src/pages/ExecutionDetails.test.tsx
@@ -161,7 +161,7 @@ describe('ExecutionDetailsContent', () => {
 
     getArtifactTypesSpy.mockResolvedValue(new GetArtifactTypesResponse());
     getArtifactsByIDSpy.mockResolvedValue(new GetArtifactsByIDResponse());
-    getContextByExecutionSpy.mockResolvedValue(null);
+    getContextByExecutionSpy.mockReturnValue(new Promise(() => {}));
   });
 
   function renderContent(id = 1) {
@@ -512,6 +512,7 @@ describe('ExecutionDetailsContent', () => {
     });
 
     it('renders neither link when no context and no cached ID', async () => {
+      getContextByExecutionSpy.mockResolvedValue(null);
       mockSuccessfulLoad();
       renderContent();
 
@@ -521,6 +522,7 @@ describe('ExecutionDetailsContent', () => {
 
       expect(screen.queryByText('Pipeline Run')).not.toBeInTheDocument();
       expect(screen.queryByText('Original Execution')).not.toBeInTheDocument();
+      await TestUtils.flushPromises();
     });
   });
 });
@@ -576,35 +578,16 @@ describe('ExecutionDetails (page wrapper)', () => {
     );
   });
 
-  it('updates toolbar title via onTitleUpdate after successful load', async () => {
-    vi.spyOn(Api.getInstance().metadataStoreService, 'getExecutionsByID').mockResolvedValue(
-      buildExecutionsByIDResponse([buildExecution()]),
-    );
-    vi.spyOn(Api.getInstance().metadataStoreService, 'getEventsByExecutionIDs').mockResolvedValue(
-      buildEventsResponse([]),
-    );
-    vi.spyOn(Api.getInstance().metadataStoreService, 'getExecutionTypesByID').mockResolvedValue(
-      buildExecutionTypesByIDResponse([buildExecutionType()]),
-    );
-    vi.spyOn(Api.getInstance().metadataStoreService, 'getArtifactTypes').mockResolvedValue(
-      new GetArtifactTypesResponse(),
-    );
-    vi.spyOn(MlmdUtils, 'getContextByExecution').mockResolvedValue(null);
-
+  it('updates toolbar title via onTitleUpdate after successful load', () => {
     const { props, updateToolbarSpy } = buildPageProps();
+    const wrapperElement = new ExecutionDetails(props).render() as any;
+    const childElement = wrapperElement.props.children;
+    childElement.props.onTitleUpdate('test-execution');
 
-    render(
-      <CommonTestWrapper>
-        <ExecutionDetails {...props} />
-      </CommonTestWrapper>,
+    expect(updateToolbarSpy).toHaveBeenCalledWith(
+      expect.objectContaining({
+        pageTitle: 'test-execution',
+      }),
     );
-
-    await waitFor(() => {
-      expect(updateToolbarSpy).toHaveBeenCalledWith(
-        expect.objectContaining({
-          pageTitle: 'test-execution',
-        }),
-      );
-    });
   });
 });


frontend/src/pages/ExecutionDetails.test.tsx
```diff
@@ -0,0 +1,467 @@
...
+    renderContent(-1);
+
+    await waitFor(() => {
+      expect(onErrorSpy).toHaveBeenCalledWith(
+        expect.stringContaining('Invalid execution id'),
+        expect.any(Error),
+        'error',
+        expect.any(Function),
+      );
+    });
```
manaswinidas: Repetitive: both this test and the NaN test above hit the same `if (isNaN(numberId) || numberId < 0)` branch and assert the identical error string. Collapse into `it.each([NaN, -1])` or drop one — two tests for one branch with the same assertion is pure duplication.

frontend/src/pages/ArtifactDetails.test.tsx
```diff
@@ -0,0 +1,316 @@
...
+
+  it('shows page error banner on service error', async () => {
+    getArtifactsByIDSpy.mockRejectedValue({ message: 'Service unavailable' });
+
+    renderWithRouter(generateProps());
+
+    await waitFor(() => {
+      expect(updateBannerSpy).toHaveBeenCalledWith(
+        expect.objectContaining({
+          message: expect.stringContaining('Error'),
```
manaswinidas: Weak assertion: `stringContaining('Error')` matches virtually any error message. The source produces `'Error: Service unavailable'` for this mock — assert against that (or at least `stringContaining('Service unavailable')`) so the test actually catches wrong messages.

frontend/src/pages/ExecutionDetails.test.tsx
```diff
@@ -0,0 +1,467 @@
...
+
+  it('places events with default type (UNKNOWN) into the UNKNOWN bucket', () => {
+    const eventDefaultType = new Event();
+    eventDefaultType.setArtifactId(100);
+
+    const response = buildEventsResponse([eventDefaultType]);
+    const result = parseEventsByType(response);
+
+    expect(result[Event.Type.UNKNOWN]).toHaveLength(1);
+    expect(result[Event.Type.INPUT]).toHaveLength(0);
```
manaswinidas: Unnecessary: the previous test already proves events only land in their correct bucket, so asserting INPUT is empty when only an UNKNOWN event was added adds no signal. Drop this line.

frontend/src/pages/ExecutionDetails.test.tsx
```diff
@@ -0,0 +1,467 @@
...
+      mockSuccessfulLoad();
+      renderContent();
+
+      await waitFor(() => {
+        expect(screen.getByText('Pipeline Run')).toBeInTheDocument();
+      });
+
+      expect(screen.getByRole('link', { name: 'runs/details/run-abc-123' })).toHaveAttribute(
+        'href',
+        expect.stringContaining('runs/details/run-abc-123'),
```
manaswinidas: Near-tautological: you find the link by accessible name `'runs/details/run-abc-123'` then assert its href *contains* that same string. The actual href includes additional segments (e.g. the execution ID). Assert the exact href value instead — that way this test catches routing construction bugs.

Same issue applies to the "original execution link" test below.

frontend/src/pages/ExecutionDetails.test.tsx
```diff
@@ -0,0 +1,467 @@
...
+    renderContent();
+
+    await waitFor(() => {
+      expect(screen.getByText('Inputs')).toBeInTheDocument();
+    });
+    expect(screen.getByText('Outputs')).toBeInTheDocument();
+  });
+
+  describe('SectionIO', () => {
+    it('renders null when events array is empty (no section headers)', async () => {
```
manaswinidas: Misleading scope: these tests render the full `ExecutionDetailsContent` (with all its data-fetching lifecycle) rather than rendering `SectionIO` in isolation with props. The test names suggest focused unit tests of the sub-component, but it's actually integration testing through the parent. Either render `<SectionIO title="Inputs" events={[]} />` directly, or rename the describe block to reflect what's actually being tested.

Also, "renders null when events array is empty" is just the successful-load test (line ~283) with extra negative assertions — the absence of IO sections when `mockSuccessfulLoad()` provides zero events is self-evident from the setup.

frontend/src/pages/ExecutionDetails.test.tsx
```diff
@@ -0,0 +1,467 @@
...
+    );
+
+    render(
+      <CommonTestWrapper>
+        <ExecutionDetails {...props} />
+      </CommonTestWrapper>,
+    );
+
+    expect(screen.getByRole('progressbar')).toBeInTheDocument();
+  });
```
manaswinidas: Redundant: this test just checks `getByRole('progressbar')`, which is identical to the `ExecutionDetailsContent` loading spinner test at line 195. The title says "initial toolbar state" but nothing about the toolbar is actually asserted. Either remove this or make it assert something the other test doesn't (e.g. verify `getInitialToolbarState` output).

kanishka-commits:
Thanks for the thorough review @manaswinidas.

## Coverage Improvement for MLMD Detail Pages

Increased coverage and improved test quality for `ArtifactDetails.tsx` and `ExecutionDetails.tsx`.

| File | Lines | Branches |
|------|-------|----------|
| `ArtifactDetails.tsx` | 86.2% → **89.65%** | 64% → **84%** |
| `ExecutionDetails.tsx` | 87.62% → **95.87%** | 73.77% → **93.44%** |

**Files changed:**
- `frontend/src/pages/ArtifactDetails.test.tsx` (+3 tests, -1 removed, 1 strengthened)
- `frontend/src/pages/ExecutionDetails.test.tsx` (+7 tests, -2 removed, 2 strengthened)

Screenshot 2026-03-17 at 2 06 30 PM

---

### Changes applied from your review

1. **Collapsed NaN/-1 tests** — Merged into `it.each([NaN, -1])`. Single test, same branch.
2. **Strengthened service error assertion** — Now asserts `stringContaining('Service unavailable')` instead of the overly broad `stringContaining('Error')`.
3. **Removed redundant INPUT assertion** — The grouping test already proves events land in correct buckets.
4. **Exact href assertions** — Pipeline run and original execution link tests now assert exact href values (`'/runs/details/run-abc-123/execution/1'`, `'/executions/original-exec-42'`) instead of `stringContaining`.
5. **Renamed misleading describe blocks** — Changed to `'IO sections (via ExecutionDetailsContent)'` and `'artifact linking (via ExecutionDetailsContent)'` to reflect they're integration tests through the parent. Removed the redundant "renders null when events array is empty" test.
6. **Fixed page wrapper test** — Now asserts actual `getInitialToolbarState` output (`pageTitle: 'Execution #1'`, `breadcrumbs`) instead of duplicating the progressbar check. Added a second test verifying the `onTitleUpdate` callback flows through to `updateToolbar`.

---

### Additional improvements

**Added 10 tests** covering error handling (service/non-service errors, artifact types failure, multiple execution types), edge cases (empty type list, empty type name, empty URI), and the page wrapper `onTitleUpdate` flow.

**Removed 3 weak tests**: redundant "no version" assertion, and two mechanical `stopPropagation` tests for trivial one-liner lambdas.

**Strengthened 3 tests** that had tautological assertions — each now verifies behavior specific to its scenario instead of repeating assertions from other tests.

Remaining uncovered lines are justified skips: private static methods only reachable from mocked components, trivial `refresh()` delegation, internal methods not called by React lifecycle, and one-liner arrow functions.

jeffspahr:
I ran the two touched test files locally:

```bash
npm run test:ui -- src/pages/ArtifactDetails.test.tsx src/pages/ExecutionDetails.test.tsx
```

They both pass, but `ExecutionDetails.test.tsx` emits repeated React Query stderr:

```text
Query data cannot be undefined. Please make sure to return a value other than undefined from your query function.
```

I think the root cause is in `ExecutionReference` in `frontend/src/pages/ExecutionDetails.tsx`: `getContextByExecution()` legitimately returns `undefined` for "no matching context", but that value is passed straight through `useQuery`, and TanStack Query treats `undefined` as an invalid successful result.

Suggested fix: normalize the query result to `null` inside the component rather than only changing the test mock.

```tsx
const { isSuccess, data: context } = useQuery<Context | null, Error>({
  queryKey: queryKeys.contextByExecution(execution.getId(), execution.getLastKnownState()),
  queryFn: async () =>
    (await getContextByExecution(execution, KFP_V2_RUN_CONTEXT_TYPE)) ?? null,
  staleTime: Infinity,
});
```

That keeps the UI behavior the same:
- context found: render the Pipeline Run link
- no context: render nothing
- no React Query invariant noise

I would avoid fixing this only in the tests by mocking `null`, because that hides the runtime mismatch instead of addressing it.

kanishka-commits:
Thanks @jeffspahr. Applied your suggestion — queryFn now normalizes undefined → null via ?? null and the type parameter is Context | null. The "Query data cannot be undefined" warning is gone. Updated the test mocks to return null accordingly.

jeffspahr:
Re-review on the latest PR head:

The previous issue is fixed. I reran the targeted frontend checks and I no longer see the earlier React Query error:

```text
Query data cannot be undefined
```

The `Context | null` + `?? null` change in `ExecutionReference` addressed that correctly.

The issue that needs attention now is different: the latest targeted Vitest run still emits this warning during the new `ExecutionDetails` coverage:

```text
Warning: Can't perform a React state update on an unmounted component.
```

So the current state is:
- old issue: fixed
- new issue: the tests still leave an async update running past cleanup, which keeps stderr noisy even though the assertions pass

My best read is that this did not become visible until the React Query warning was removed. In other words, the last fix seems to have exposed the next remaining issue rather than causing a regression.

The likely cause is that some of the new success-path tests in `frontend/src/pages/ExecutionDetails.test.tsx` finish as soon as they see `Reference` / the toolbar title update, but `ExecutionDetailsContent` still has async child work in flight (`ExecutionReference` query and possibly `SectionIO` artifact lookups). Cleanup then unmounts the tree while one of those updates is still resolving.

Suggested fix:
- make the affected `ExecutionDetails` tests wait for the async child work to settle before the test exits
- in practice, that likely means explicitly waiting for the `getContextByExecution` / linked-artifact async path to complete and then flushing queued microtasks before cleanup
- if those tests are not meant to exercise that child async behavior, another clean option is to stub/mock that async work out for those cases so the test only covers the behavior it is actually asserting
- if the warning still remains after tightening the tests, then the underlying component likely needs an unmount guard around async `setState`, especially in `SectionIO.componentDidMount`

For reference, this came from rerunning:

```bash
npm run test:ui -- src/pages/ArtifactDetails.test.tsx src/pages/ExecutionDetails.test.tsx
npm run typecheck
```

Both passed on the latest head, but the test run still emitted the unmounted-component warning in stderr.

frontend/src/pages/ExecutionDetails.test.tsx
```diff
@@ -0,0 +1,610 @@
...
+      Api.getInstance().metadataStoreService,
+      'getExecutionTypesByID',
+    );
+    getArtifactTypesSpy = vi.spyOn(Api.getInstance().metadataStoreService, 'getArtifactTypes');
+    getArtifactsByIDSpy = vi.spyOn(Api.getInstance().metadataStoreService, 'getArtifactsByID');
+    getContextByExecutionSpy = vi.spyOn(MlmdUtils, 'getContextByExecution');
+
+    getArtifactTypesSpy.mockResolvedValue(new GetArtifactTypesResponse());
+    getArtifactsByIDSpy.mockResolvedValue(new GetArtifactsByIDResponse());
+    getContextByExecutionSpy.mockResolvedValue(null);
```
jeffspahr: Tested locally: making the shared default a never-resolving promise cleaned up the noisy success-path tests, because most of them are not asserting anything about `ExecutionReference` and don't need that query to complete.

```suggestion
    getContextByExecutionSpy.mockReturnValue(new Promise(() => {}));
```

frontend/src/pages/ExecutionDetails.test.tsx
```diff
@@ -0,0 +1,610 @@
...
+      mockSuccessfulLoad();
+      renderContent();
+
+      await waitFor(() => {
+        expect(screen.getByText('Reference')).toBeInTheDocument();
+      });
+
+      expect(screen.queryByText('Pipeline Run')).not.toBeInTheDocument();
+      expect(screen.queryByText('Original Execution')).not.toBeInTheDocument();
+    });
```
jeffspahr, line 526: Tested locally: once the shared default stays pending, this test can still exercise the explicit `null` case, but it needs to drain that resolved query before teardown or the run still emits the unmounted-component warning.

```suggestion
    it('renders neither link when no context and no cached ID', async () => {
      getContextByExecutionSpy.mockResolvedValue(null);
      mockSuccessfulLoad();
      renderContent();

      await waitFor(() => {
        expect(screen.getByText('Reference')).toBeInTheDocument();
      });

      expect(screen.queryByText('Pipeline Run')).not.toBeInTheDocument();
      expect(screen.queryByText('Original Execution')).not.toBeInTheDocument();
      await TestUtils.flushPromises();
    });
```

frontend/src/pages/ExecutionDetails.test.tsx
```diff
@@ -0,0 +1,610 @@
...
+    );
+
+    await waitFor(() => {
+      expect(updateToolbarSpy).toHaveBeenCalledWith(
+        expect.objectContaining({
+          pageTitle: 'test-execution',
+        }),
+      );
+    });
+  });
```
jeffspahr: Tested locally: rewriting this wrapper test to invoke the child `onTitleUpdate` callback directly keeps the intent of the assertion, but avoids mounting the full async `ExecutionDetailsContent` subtree just to verify one prop passthrough. With that change, the touched test pair passed cleanly with no stderr warning.

```suggestion
  it('updates toolbar title via onTitleUpdate after successful load', () => {
    const { props, updateToolbarSpy } = buildPageProps();
    const wrapperElement = new ExecutionDetails(props).render() as any;
    const childElement = wrapperElement.props.children;

    childElement.props.onTitleUpdate('test-execution');

    expect(updateToolbarSpy).toHaveBeenCalledWith(
      expect.objectContaining({
        pageTitle: 'test-execution',
      }),
    );
  });
```

kanishka-commits:
Thanks @jeffspahr — all three suggestions have been addressed:

- Default mock now uses a never-resolving promise to avoid triggering unused queries.
- Updated test to use `mockResolvedValue(null)` with `flushPromises()` before teardown.
- Simplified `onTitleUpdate` test by invoking it directly instead of mounting the async subtree.

Both test files now pass with zero stderr warnings.

Issue #13044, opened 2026-03-16
chore(frontend): Add unit tests for MLMD canvas and column layout (0% coverage on 3 files)


PR3MM:
## Chore description

`EdgeCanvas.tsx`, `ControlledEdgeCanvas.tsx`, and `LineageCardColumn.tsx` handle the edge-drawing and column layout for the Lineage Explorer. None of these files has any test coverage. These components contain non-trivial pixel offset math (`CARD_OFFSET`, `CARD_ROW_HEIGHT`, same-card vs next-card calculations) that is easy to regress silently.

**Note — latent bug:** `LineageCardColumn.tsx` line 202 has `artifactCardIndex === null` but the variable is declared as `number | undefined` and initialized to `undefined`. This condition is dead code — it will never be true. The intended check was likely `=== undefined`. This means the first artifact in each execution's artifact list always takes the `else` branch and adds `NEXT_ITEM_NEXT_CARD_OFFSET` even when it shouldn't. Writing tests for this path will surface the bug.

### Scope

| File | Status | Description |
|------|--------|-------------|
| `EdgeCanvas.tsx` | Zero tests | Draws connecting SVG lines between card columns |
| `ControlledEdgeCanvas.tsx` | Zero tests | Draws lines for execution-to-artifact connections with offset logic |
| `LineageCardColumn.tsx` | Zero tests | Assembles cards + canvases into a column with complex offset math |

#### New files

- `src/mlmd/EdgeCanvas.test.tsx`
- `src/mlmd/ControlledEdgeCanvas.test.tsx`
- `src/mlmd/LineageCardColumn.test.tsx`

### Test cases (~22 total)

**EdgeCanvas (~6)**

- [ ] renders no `EdgeLine`s when no elements have a `next` connection
- [ ] renders one `EdgeLine` per connected element
- [ ] applies `reverseEdges` CSS class when `reverseEdges` is true
- [ ] positions canvas with correct left offset (`cardWidth`) and top offset
- [ ] increments height by `CARD_ROW_HEIGHT` per connected element
- [ ] increments height by `CARD_OFFSET` at each card boundary

**ControlledEdgeCanvas (~6)**

- [ ] renders one `EdgeLine` per `artifactId`
- [ ] applies initial offset to y-coordinate when `offset` prop is non-zero
- [ ] adds only `CARD_OFFSET` when consecutive artifacts are on the same card
[...]

PR #13049, opened 2026-03-17
chore(frontend): add unit tests for MLMD canvas and column layout

pnaik1:
Fixes https://github.com/kubeflow/pipelines/issues/13044

Bug fix in LineageCardColumn.tsx: corrects an off-by-one offset miscalculation in the ControlledEdgeCanvas positioning logic. The initial check for artifactCardIndex was using === null, but the variable is initialized as undefined, causing the first artifact to incorrectly advance the artifact offset instead of being treated as the initialization step. Fixed to === undefined.

New test file `EdgeCanvas.test.tsx` : Vitest unit tests covering the EdgeCanvas component, including:
Edge rendering counts (zero edges, multiple connected elements)
reverseEdges CSS class toggling
viewHeight pixel offset math (same card, cross-card, CARD_OFFSET accumulation)

New test file `ControlledEdgeCanvas.test.tsx `: Vitest unit tests covering the ControlledEdgeCanvas component, including:
Per-artifact EdgeLine rendering counts
reverseEdges CSS class toggling
viewHeight math for same-card vs. different-card artifacts, and non-zero offset prop
Canvas CSS top positioning

New test file `LineageCardColumn.test.tsx` : Vitest unit tests covering LineageCardColumn, including:
Structural rendering (column header, per-card headings)
EdgeCanvas vs. ControlledEdgeCanvas branch selection
skipEdgeCanvas prop
ControlledEdgeCanvas prop math: offset, top, edgeWidth, cardWidth

**Checklist:**

### DIFF
--- a/frontend/src/mlmd/ControlledEdgeCanvas.test.tsx
+++ b/frontend/src/mlmd/ControlledEdgeCanvas.test.tsx
@@ -1,0 +1,230 @@
+/*
+ * Copyright 2021 The Kubeflow Authors
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import * as React from 'react';
+import { render } from '@testing-library/react';
+import { describe, expect, it } from 'vitest';
+import { CARD_OFFSET, ControlledEdgeCanvas } from './ControlledEdgeCanvas';
+import { edgeCanvasCss } from './EdgeCanvas';
+import { CARD_ROW_HEIGHT, CARD_TITLE_HEIGHT } from './LineageCss';
+
+const defaultProps = {
+  cardWidth: 200,
+  edgeWidth: 40,
+  reverseEdges: false,
+  offset: 0,
+  top: 0,
+  outputExecutionToOutputArtifactMap: new Map<number, number[]>(),
+};
+
+function svgHeights(container: HTMLElement): number[] {
+  return Array.from(container.querySelectorAll('svg')).map((svg) =>
+    Number(svg.getAttribute('height')),
+  );
+}
+
+function pathDs(container: HTMLElement): (string | null)[] {
+  return Array.from(container.querySelectorAll('path')).map((p) => p.getAttribute('d'));
+}
+
+describe('ControlledEdgeCanvas', () => {
+  // --- rendering counts ---
+
+  it('renders one EdgeLine per artifactId', () => {
+    const { container } = render(
+      <ControlledEdgeCanvas
+        {...defaultProps}
+        artifactIds={[10, 11, 12]}
+        artifactToCardMap={
+          new Map([
+            [10, 0],
+            [11, 0],
+            [12, 0],
+          ])
+        }
+      />,
+    );
+    expect(container.querySelectorAll('svg').length).toBe(3);
+  });
+
+  it('renders no EdgeLines when artifactIds is empty', () => {
+    const { container } = render(
+      <ControlledEdgeCanvas {...defaultProps} artifactIds={[]} artifactToCardMap={new Map()} />,
+    );
+    expect(container.querySelectorAll('svg').length).toBe(0);
+  });
+
+  it('applies edgeCanvasReverse class when reverseEdges is true', () => {
+    const { container } = render(
+      <ControlledEdgeCanvas
+        {...defaultProps}
+        artifactIds={[10]}
+        artifactToCardMap={new Map([[10, 0]])}
+        reverseEdges={true}
+      />,
+    );
+    expect((container.firstChild as HTMLElement).classList.contains('edgeCanvasReverse')).toBe(
+      true,
+    );
+  });
+
+  it('does not apply edgeCanvasReverse class when reverseEdges is false', () => {
+    const { container } = render(
+      <ControlledEdgeCanvas
+        {...defaultProps}
+        artifactIds={[10]}
+        artifactToCardMap={new Map([[10, 0]])}
+      />,
+    );
+    expect((container.firstChild as HTMLElement).classList.contains('edgeCanvasReverse')).toBe(
+      false,
+    );
+  });
+
+  // --- viewHeight / pixel offset math ---
+
+  it('single artifact: first EdgeLine starts at viewHeight=1 (canvas base)', () => {
+    const { container } = render(
+      <ControlledEdgeCanvas
+        {...defaultProps}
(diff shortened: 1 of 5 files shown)


frontend/src/mlmd/ControlledEdgeCanvas.test.tsx
```diff
@@ -0,0 +1,221 @@
...
+        artifactToCardMap={
+          new Map([
+            [10, 0],
+            [20, 1],
+          ])
+        }
+      />,
+    );
+    expect(svgHeights(container)).toEqual([1, 1 + CARD_OFFSET + CARD_ROW_HEIGHT]);
+  });
```
jeffspahr, line 140: These cases still follow the implementation's `artifactId - 1` lookup, so they won't catch the real spacing bug when MLMD IDs are non-consecutive. `[10, 42]` on the same card should still stay on the same-card path; today it gets treated as a cross-card jump because `artifactToCardMap.get(41)` is `undefined`. Please add a regression like this and then compare against the previous entry in `artifactIds`, not `artifactId - 1`.

```suggestion
    expect(svgHeights(container)).toEqual([1, 1 + CARD_OFFSET + CARD_ROW_HEIGHT]);
  });

  it('uses the previous artifact in the sequence when ids are non-consecutive', () => {
    const { container } = render(
      <ControlledEdgeCanvas
        {...defaultProps}
        artifactIds={[10, 42]}
        artifactToCardMap={
          new Map([
            [10, 0],
            [42, 0],
          ])
        }
      />,
    );
    expect(svgHeights(container)).toEqual([1, 1 + CARD_OFFSET]);
  });
```

frontend/src/mlmd/ControlledEdgeCanvas.test.tsx
```diff
@@ -0,0 +1,221 @@
...
+    const { container } = render(
+      <ControlledEdgeCanvas
+        {...defaultProps}
+        artifactIds={[10]}
+        artifactToCardMap={new Map([[10, 0]])}
+        offset={10}
+      />,
+    );
+    expect(svgHeights(container)[0]).toBe(11);
+  });
```
jeffspahr, line 154: This suite currently exercises only non-negative starting offsets, but the `=== undefined` change in `LineageCardColumn` makes the second reverse-edge canvas receive `-54` in the two-execution case. Right now that renders `<svg height="-53">` because `ControlledEdgeCanvas` adds the negative offset directly into `viewHeight`. Please cover that path before merging.

```suggestion
    expect(svgHeights(container)[0]).toBe(11);
  });

  it('keeps rendered SVG heights positive when the starting offset is negative', () => {
    const { container } = render(
      <ControlledEdgeCanvas
        {...defaultProps}
        artifactIds={[2]}
        artifactToCardMap={new Map([[2, 1]])}
        offset={-54}
        reverseEdges={true}
      />,
    );
    expect(svgHeights(container)[0]).toBeGreaterThan(0);
  });
```

manaswinidas:
Fixes #13044 - @pnaik1 can you add this to your PR description ? This PR doesn't link to the issue currently

pnaik1:
@jeffspahr thanks for the review here, Update the PR accordingly

frontend/src/mlmd/ControlledEdgeCanvas.tsx
```diff
@@ -64,18 +64,20 @@ export class ControlledEdgeCanvas extends React.Component<ControlledEdgeCanvasPr
...
     this.props.artifactIds.forEach((artifactId, index) => {
       if (index !== 0) {
         let offset = CARD_OFFSET;
 
+        const prevArtifactId = this.props.artifactIds[index - 1];
         if (
           this.props.artifactToCardMap.get(artifactId) !==
-          this.props.artifactToCardMap.get(artifactId - 1)
+          this.props.artifactToCardMap.get(prevArtifactId)
         ) {
```
pnaik1, line 81: made the changes here @jeffspahr

frontend/src/mlmd/LineageCardColumn.test.tsx
```diff
@@ -0,0 +1,264 @@
...
+      connectedCards: executionCards,
+      outputExecutionToOutputArtifactMap: new Map([
+        [10, [1]],
+        [11, [2]],
+      ]),
+    });
+    const calls = (ControlledEdgeCanvas as unknown as Mock).mock.calls;
+    expect(calls.length).toBe(2);
+    expect(calls[0][0].offset).toBe(0);
+    expect(calls[1][0].offset).toBe(-NEXT_ITEM_SAME_CARD_OFFSET);
```
jeffspahr: This expectation currently locks in the regressed `-54` offset. Once the column keeps the existing cross-card spacing, the second canvas should still start `67px` below the first execution row (`121 - 54`).

```suggestion
    expect(calls[1][0].offset).toBe(
      NEXT_ITEM_NEXT_CARD_OFFSET - NEXT_ITEM_SAME_CARD_OFFSET,
    );
```

frontend/src/mlmd/LineageCardColumn.tsx
```diff
@@ -199,7 +199,7 @@ export class LineageCardColumn extends React.Component<LineageCardColumnProps> {
 
       // Advance starting artifact offset.
       artifactIds.forEach((artifactId) => {
-        if (artifactCardIndex === null) {
+        if (artifactCardIndex === undefined) {
```
jeffspahr: This changes the running artifact offset for single-artifact executions. In the existing `Map([[10, [1]], [11, [2]]])` layout, this now leaves `artifactOffset` at `0` after the first execution, so the second canvas gets `offset=-54` instead of the current `67`, and `ControlledEdgeCanvas` flattens that reverse edge to `height=1`. The old behavior advanced the running offset on the first artifact as well.

I think the minimal fix is to make that first-artifact advance explicit instead of returning early here, e.g. derive `newArtifactIndex` first and then apply `NEXT_ITEM_NEXT_CARD_OFFSET` whenever `artifactCardIndex` is unset or changes cards, otherwise apply `NEXT_ITEM_SAME_CARD_OFFSET`.

jeffspahr (changes requested):
One blocking issue remains on the refreshed head. The new `artifactCardIndex === undefined` path changes the existing reverse-edge layout for the two-executions / two-artifact-cards case: `LineageCardColumn` now computes offsets `[0, -54]` where `origin/master` computes `[0, 67]`, and `ControlledEdgeCanvas` flattens the second connection to `height=1` instead of the current `68px` curve.

I re-ran `npm run test:ui -- src/mlmd/EdgeCanvas.test.tsx src/mlmd/ControlledEdgeCanvas.test.tsx src/mlmd/LineageCardColumn.test.tsx` on the live PR head and reproduced the offset/path mismatch directly against `origin/master`. The inline comments call out the minimal fix and the test expectation that should move with it.

pnaik1:
@jeffspahr addressed the fix as suggested

jeffspahr:
Re-reviewed the latest head `ee7945af5321679f4c6c8a1b4e386dbf8d2f8706` on March 23, 2026.

No blocking or meaningful findings on the current diff.

The earlier reverse-edge regression is addressed by the latest `LineageCardColumn.tsx` update. I reran the targeted MLMD Vitest suite (`35` tests passed) and re-rendered the previously failing two-execution / two-artifact scenario plus a same-card variant against fresh `origin/master`; both branches now produced the same SVG heights and paths.

PR #12925, opened 2026-02-27
[frontend] Migrate Material UI v4 to MUI v5

jeffspahr:
## Summary
- migrate frontend dependencies from `@material-ui/core` / `@material-ui/icons` to MUI v5 (`@mui/material`, `@mui/icons-material`) with Emotion
- apply codemod-driven import/API updates across the frontend app
- migrate theme override structure to MUI v5 (`components.styleOverrides`) in the main and MLMD themes
- fix MUI v5 transition compatibility in `SidePanel` and update related tests/selectors/snapshots for MUI v5 behavior
- add migration checklist artifacts and lessons learned under `tasks/`

## Verification
- `cd frontend && npm run test:ci`
  - UI: 118 files / 1654 tests passed
  - server: 10 files / 163 tests passed
- `cd frontend && npm run build`
- guard checks
  - `git diff -- frontend/src/apis frontend/src/apisv2beta1 frontend/src/third_party` (empty)
  - `rg -n "@material-ui/" frontend/src frontend/package.json` (no matches)

## Notes
- this PR intentionally keeps React at 17 and focuses only on MUI v5 migration scope (issue #12894)

PR #13062, opened 2026-03-18
fix(frontend): restore sidebar nav button styling

jeffspahr:
## Summary
- restore the master sidebar nav buttons to inherit their color instead of picking up the shared primary button variant
- keep the MUI v5 sidebar buttons out of the global `MuiButton.textPrimary` border override that introduced the boxed nav regression
- add a focused SideNav regression test and refresh the affected snapshots

## Regression Context
- the regression was introduced by `c3f9f67c2b4d9c682907aabd332ca268b5be350b` on March 2, 2026: [frontend] Migrate Material UI v4 to MUI v5 ([#12925](https://github.com/kubeflow/pipelines/pull/12925))
- before that change, the `2.16.0` sidebar rendered nav items as plain text buttons without the primary button variant classes
- after the MUI v5 migration, the sidebar `Button` instances started rendering with `MuiButton-textPrimary` / `MuiButton-colorPrimary`
- the shared MUI theme in `frontend/src/Css.tsx` already adds `border: 1px solid #ddd` to `MuiButton.textPrimary`, so the migrated sidebar unexpectedly picked up bordered, boxed nav items
- this fix scopes the sidebar back to `color='inherit'`, which restores the pre-regression behavior without changing the shared primary button styling used elsewhere

## Testing
- `fnm exec --using v24.14.0 npx vite build`
- `fnm exec --using v24.14.0 npm run test:ui -- src/components/SideNav.test.tsx -u`
- browser compare of `http://localhost:3001/#/pipelines` vs `http://localhost:3011/#/pipelines` after the fix
  - sidebar crop diff before fix: `6035` pixels (`2.2859%`)
  - sidebar crop diff after fix: `0` pixels

### DIFF
--- a/frontend/src/components/SideNav.tsx
+++ b/frontend/src/components/SideNav.tsx
@@ -202,14 +202,22 @@ interface SideNavState {
   manualCollapseState: boolean;
 }
 
-interface SideNavButtonProps {
+type SideNavButtonProps = Omit<
+  React.ComponentPropsWithoutRef<typeof Button>,
+  'children' | 'color'
+> & {
   children: React.ReactNode;
   collapsed: boolean;
-  className?: false | string;
-}
+};
 
-const SideNavButton: React.FC<SideNavButtonProps> = ({ children, collapsed, className }) => (
+const SideNavButton: React.FC<SideNavButtonProps> = ({
+  children,
+  collapsed,
+  className,
+  ...buttonProps
+}) => (
   <Button
+    {...buttonProps}
     color='inherit'
     className={classes(css.button, className, collapsed && css.collapsedButton)}
   >
@@ -221,6 +229,7 @@ export class SideNav extends React.Component<SideNavInternalProps, SideNavState>
   private _isMounted = true;
   private readonly _AUTO_COLLAPSE_WIDTH = 800;
   private readonly _HUB_ADDRESS = '/hub/';
+  private readonly _boundMaybeResize = this._maybeResize.bind(this);
 
   constructor(props: any) {
     super(props);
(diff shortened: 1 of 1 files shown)


frontend/src/components/SideNav.tsx
```diff
@@ -202,6 +202,21 @@ interface SideNavState {
   manualCollapseState: boolean;
 }
 
+interface SideNavButtonProps {
+  children: React.ReactNode;
+  collapsed: boolean;
+  className?: false | string;
```
manaswinidas: Nit: `false | string` is unconventional — the callers' `&&` short-circuit pattern is leaking into the component API. `string | undefined` is more idiomatic and `classes()` handles both fine.

frontend/src/components/SideNav.tsx
```diff
@@ -202,6 +202,21 @@ interface SideNavState {
...
+}
+
+const SideNavButton: React.FC<SideNavButtonProps> = ({ children, collapsed, className }) => (
```
manaswinidas: Minor: rest props aren't forwarded to the inner `<Button>`. Fine today, but extending with `...rest` (or basing on `ComponentPropsWithoutRef<typeof Button>`) would future-proof for `aria-*` / `data-*` needs.

manaswinidas:
The frontend integration tests in CI are failing

jeffspahr:
Addressed the SideNav review feedback in 90f30ebe7: the resize listener now unregisters on unmount, SideNavButton uses standard Button props instead of false | string className typing, and the wrapper forwards rest props so aria/data attributes can pass through later if needed. The previous red CI state was down to the frontend integration v1.34 tls=false lane on the earlier head; fresh checks are rerunning on this commit.

frontend/src/components/SideNav.tsx
jeffspahr: Addressed in 90f30ebe7. `SideNavButton` now forwards the remaining `Button` props to the inner MUI button, so future `aria-*` / `data-*` usage can flow through without changing the wrapper again.

frontend/src/components/SideNav.tsx
jeffspahr: Addressed in 90f30ebe7. `SideNavButton` now takes normal `Button` props, and the `className` callers were normalized to pass `string | undefined` instead of `false | string`.

jeffspahr:
The current frontend integration failures do not look related to the SideNav diff. In the failed lane logs, the tensorboard-example pod is hitting Docker Hub unauthenticated pull limits when pulling python:alpine3.9 (429 Too Many Requests), which leaves the sample run stuck in Running and causes the downstream WDIO graph assertions to fail.

/retest

PR #13065, opened 2026-03-18
feat(frontend): surface pod lifecycle failure reason in run details UI. Part of #12843

Harshdhall01:
When a pod hits a lifecycle failure (e.g. OOMKilled, ImagePullBackOff),
the KFP UI currently shows only a generic "Resource failed to execute" 
message with no indication of the actual failure reason. Users are 
forced to use kubectl to find out why their pipeline failed, which 
breaks KFP's goal of being a Kubernetes abstraction for data scientists.

Changes made:
- StatusV2.tsx: When nodeMessage is present on a FAILED node, display 
  the specific failure reason in the status tooltip instead of the 
  generic "Resource failed to execute" message
- RunDetailsV2.tsx: Add inline red failure reason badge next to the 
  status icon in the run header so users can immediately see why 
  their pod failed

Evidence of gap found during codebase exploration:
- grep -r "ImagePullBackOff" frontend/src/ → zero results
- grep -r "OOMKilled" frontend/src/ → zero results  
- grep -r "pod_lifecycle" frontend/src/ → zero results

This confirms pod lifecycle failures are completely unhandled 
in the frontend today.

This PR is the frontend visualization piece of #12843. 
Remaining work tracked in the issue:
- Backend (Go): extract pod lifecycle reasons from 
  pod.status.containerStatuses[].state.waiting.reason and 
  state.terminated.reason
- Frontend: extend side panel with full pod events timeline
- Timeout configuration per failure category via env variables

### DIFF
--- a/frontend/src/components/tabs/RuntimeNodeDetailsV2.tsx
+++ b/frontend/src/components/tabs/RuntimeNodeDetailsV2.tsx
@@ -202,6 +202,16 @@ function TaskNodeDetail({
         {/* Logs tab */}
         {selectedTab === 2 && (
           <div className={commonCss.page}>
+            {execution?.getLastKnownState() === Execution.State.FAILED && (
+              <div style={{ margin: "16px", padding: "16px", backgroundColor: "#ffebee", borderRadius: "8px", borderLeft: "4px solid #c62828" }}>
+                <div style={{ fontWeight: 600, color: "#c62828", fontSize: "14px", marginBottom: "8px" }}>
+                  Pod Lifecycle Failure Detected
+                </div>
+                <div style={{ fontSize: "13px", color: "#555555" }}>
+                  This node failed due to a pod lifecycle error. Possible causes include OOMKilled, ImagePullBackOff, or CrashLoopBackOff. Check logs below for details.
+                </div>
+              </div>
+            )}
             {logsBannerMessage && (
               <React.Fragment>
                 <Banner message={logsBannerMessage} additionalInfo={logsBannerAdditionalInfo} />
--- a/frontend/src/pages/StatusV2.tsx
+++ b/frontend/src/pages/StatusV2.tsx
@@ -45,7 +45,9 @@ export function statusToIcon(
     case V2beta1RuntimeState.FAILED:
       IconComponent = ErrorIcon;
       iconColor = color.errorText;
-      title = 'Resource failed to execute';
+      title = nodeMessage
+        ? `Failed: ${nodeMessage}`
+        : 'Resource failed to execute';
       break;
     case V2beta1RuntimeState.PENDING:
       IconComponent = PendingIcon;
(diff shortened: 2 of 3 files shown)


jiayuzhao05 (commented):
This improves the frontend surfacing of failure information, but it does not fully solve the end-to-end pod lifecycle failure visibility problem yet. The UI now displays nodeMessage / runMetadata.error.message when available, but the backend extraction of concrete pod lifecycle reasons still appears to be tracked separately in the issue. So this looks like a partial fix rather than the complete solution.

Harshdhall01:
> This improves the frontend surfacing of failure information, but it does not fully solve the end-to-end pod lifecycle failure visibility problem yet. The UI now displays nodeMessage / runMetadata.error.message when available, but the backend extraction of concrete pod lifecycle reasons still appears to be tracked separately in the issue. So this looks like a partial fix rather than the complete solution.

Thank you for the review @jiayuzhao05 You are correct
this PR is intentionally the frontend visualization piece 
as a starting point. The backend extraction from 
containerStatuses is tracked in my GSoC proposal as the 
core summer work.

I am aware of PR #13097 which takes a different approach 
through the Persistence Agent. My proposal focuses on the 
pod watcher + timeout configuration angle described in the 
original issue by @alyssacgoins.

Would love guidance on which approach the maintainers 
prefer before I extend the backend work.

frontend/src/components/tabs/RuntimeNodeDetailsV2.tsx
```diff
@@ -202,6 +202,16 @@ function TaskNodeDetail({
         {/* Logs tab */}
         {selectedTab === 2 && (
           <div className={commonCss.page}>
+            {execution?.getLastKnownState() === Execution.State.FAILED && (
```
jiayuzhao05, line 205: This condition is too broad. Execution.State.FAILED does not imply a pod lifecycle failure. As written, the UI will label any failed execution as “Pod Lifecycle Failure Detected”, which seems misleading for user-code errors, dependency failures, backend errors, etc.

frontend/src/pages/StatusV2.tsx
```diff
@@ -45,7 +45,9 @@ export function statusToIcon(
     case V2beta1RuntimeState.FAILED:
       IconComponent = ErrorIcon;
       iconColor = color.errorText;
-      title = 'Resource failed to execute';
+      title = nodeMessage
```
jiayuzhao05, line 48: I agree with the change, but I’d recommend keeping the phrasing consistent between the generic and detailed cases. Something like Execution failed vs Execution failed: ${nodeMessage} may read more coherently than mixing Failed: with Resource failed to execute.

droctothorpe:
@Harshdhall01: Hey, thank you for your PR! As you are not a member of Kubeflow organization, your PR must link to an issue with 'ready' label. Please read our [CONTRIBUTING.md](https://github.com/kubeflow/pipelines/blob/master/CONTRIBUTING.md), closing this PR

PR #13069, opened 2026-03-18
Feat(workflow) Cherry-pick PR#12804 "Feat(workflow) Add multi-arch docker buildx support to the image building workflows for building arm64 and amd64 container images" to release-2.16

jtu-ampere:
Cherry-pick PR#12804 "Feat(workflow) Add multi-arch docker buildx support to the image building workflows for building arm64 and amd64 container images" to release-2.16

---------

Description of your changes:
---
Feat(workflow) Add multi-arch docker buildx with 4 workflows - build-and-push.yml, image-builds-master.yml, create-manifest.yml and image-builds-release.yml for building arm64 and amd64 container images on GitHub native runners. The method is building the container image on the architecture matched GitHub native runners, pushing the images by digest, and then merge two image digests with a manifest. Finally, push the single multi-arch container image to container registry.

1. Update the workflows build-and-push.yml and add create-manifest.yml with GitHub native runners :
 * Changing runs-on from ubuntu-latest only to ubuntu-latest for amd64 platform and ubuntu-24.04-arm for arm64 platform
 * Add Docker Buildx action
 * Add the new parameters, platforms and arch_tag for linux/amd64 and linux/arm64 for building container image on the architecture matched GitHub native runners

2. Update the workflows image-builds-master.yml and image-builds-release.yml to add platform parameter for handling kfp-metadata-writer and kfp-inverse-proxy-agent for amd64 only while building the others in Multi-Arch container images

3. Add two parameters for docker/setup-buildx-action@v3 : builder and use for multi-arch container image

**Checklist:**

### DIFF
--- a/.github/workflows/build-and-push.yml
+++ b/.github/workflows/build-and-push.yml
@@ -12,6 +12,10 @@ on:
         type: string
         description: 'Target Image Tag'
         required: true
+      arch_tag:
+        type: string
+        description: 'Architecture tag suffix e.g. amd64 or arm64'
+        required: true
       overwrite_imgs:
         type: string
         default: 'true'
@@ -39,6 +43,14 @@ on:
         type: string
         description: 'Provide the docker file name'
         required: true
+      platforms:
+        type: string
+        required: true
+        description: e.g, linux/amd64 or linux/arm64
+      runs_on:
+        type: string
+        required: true
+        description: 'Runner label to execute on'
       push:
         type: boolean
         default: false
@@ -55,6 +67,11 @@ on:
         type: string
         default: 'X.Y.Z'
         description: 'Target Image Tag'
+        required: true
+      arch_tag:
+        type: string
+        default: 'amd64'
+        description: 'Architecture tag suffix e.g. amd64 or arm64'
         required: true
       fail_fast:
         type: string
@@ -91,6 +108,15 @@ on:
         default: ''
         description: 'Provide the docker file name'
         required: true
+      platforms:
+        type: string
+        default: ''
+        required: true
+        description: 'Provide the target architecture e.g, linux/amd64, linux/arm64'
+      runs_on:
+        type: string
+        required: true
+        description: 'Runner label to execute on'
       push:
         type: boolean
         default: false
@@ -109,7 +135,7 @@ env:
 jobs:
   build-and-push-images:
     continue-on-error: false
-    runs-on: ubuntu-latest
+    runs-on: ${{ inputs.runs_on }}
     permissions:
       contents: read
       packages: write
@@ -131,6 +157,12 @@ jobs:
           registry: ${{ env.IMAGE_REGISTRY }}
           username: ${{ github.actor }}
           password: ${{ secrets.GITHUB_TOKEN }}
+
+      - name: Set up Docker Buildx
+        id: buildx
+        uses: docker/setup-buildx-action@v3
+        with:
+          driver: docker-container
 
       - name: Check if image tag already exists
         id: check_tag
@@ -173,21 +205,44 @@ jobs:
         uses: docker/build-push-action@v6
         if: steps.check_tag.outcome == 'success'
         with:
+          platforms: ${{ inputs.platforms }}
           context: ${{ inputs.image_context }}
           file: ${{ inputs.docker_file }}
-          push: ${{ inputs.push }}
           tags: ${{ steps.meta.outputs.tags }}
           labels: ${{ steps.meta.outputs.labels }}
+          # Conditionally switch between standard push and digest-only push
+          outputs: > 
+            ${{ github.event_name == 'workflow_dispatch' && 
+                format('type=image,name={0}/{1}/{2}:{3},push={4}', env.IMAGE_REGISTRY, env.IMAGE_ORG, inputs.app_to_build, inputs.target_tag, inputs.push) || 
+                format('type=image,name={0}/{1}/{2},push-by-digest=true,name-canonical=true,push={3}', env.IMAGE_REGISTRY, env.IMAGE_ORG, inputs.app_to_build, inputs.push) }}
           build-args: |
             COMMIT_SHA=${{ steps.get_commit.outputs.sha }}
             COMMIT_HASH=${{ steps.get_commit.outputs.sha }}
             TAG_NAME=${{ inputs.target_tag }}
+          builder: ${{ steps.buildx.outputs.name }}
+          cache-from: type=gha,scope=${{ inputs.app_to_build }}-${{ inputs.arch_tag }}
+          cache-to: type=gha,mode=max,scope=${{ inputs.app_to_build }}-${{ inputs.arch_tag }}
 
-      # This step generates an artifact attestation for the image,
-      # which is an unforgeable statement about where and how it was built.
-      # It increases supply chain security for people who consume the
-      # image.
-      # Ref: https://docs.github.com/en/actions/security-for-github-actions/using-artifact-attestations/using-artifact-attestations-to-establish-provenance-for-builds
+
+      # Export the digest
+      - name: Export digest
+        if: steps.check_tag.outcome == 'success' && inputs.push
+        run: |
+          mkdir -p /tmp/digests
+          digest="${{ steps.push.outputs.digest }}"
+          touch "/tmp/digests/${digest#sha256:}"
+
+      # Upload the digest
+      - name: Upload digest
+        if: steps.check_tag.outcome == 'success' && inputs.push
+        uses: actions/upload-artifact@v4
+        with:
+          name: digests-${{ inputs.app_to_build }}-${{ inputs.arch_tag }}
+          path: /tmp/digests/*
+          if-no-files-found: error
+          retention-days: 1
+
+      # This step generates an artifact attestation for the image
       - name: Generate artifact attestation
         uses: actions/attest-build-provenance@v1
         if: steps.check_tag.outcome == 'success' && inputs.push
--- a/.github/workflows/build-and-push.yml
+++ b/.github/workflows/build-and-push.yml
@@ -3,7 +3,7 @@ run-name: Build images
 on:
   workflow_call:
     secrets:
-      GITHUB_TOKEN:
+      GITHUBTOKEN:
         required: true
     inputs:
       src_branch:
@@ -159,7 +159,7 @@ jobs:
         with:
           registry: ${{ env.IMAGE_REGISTRY }}
           username: ${{ github.actor }}
-          password: ${{ secrets.GITHUB_TOKEN }}
+          password: ${{ secrets.GITHUBTOKEN }}
 
       - name: Set up Docker Buildx
         id: buildx
--- a/.github/workflows/build-and-push.yml
+++ b/.github/workflows/build-and-push.yml
@@ -2,6 +2,9 @@ name: Build and Push images
 run-name: Build images
 on:
   workflow_call:
+    secrets:
+      GITHUB_TOKEN:
+        required: true
     inputs:
       src_branch:
         type: string
--- a/.github/workflows/create-manifest.yml
+++ b/.github/workflows/create-manifest.yml
@@ -1,6 +1,9 @@
 name: Create Multi-Arch Manifest
 on:
   workflow_call:
+    secrets:
+      GITHUB_TOKEN:
+        required: true    
     inputs:
       target_tag:
         type: string
--- a/.github/workflows/create-manifest.yml
+++ b/.github/workflows/create-manifest.yml
@@ -1,0 +1,150 @@
+name: Create Multi-Arch Manifest
+on:
+  workflow_call:
+    inputs:
+      target_tag:
+        type: string
+        description: 'Target Image Tag'
+        required: true
+      app_to_build:
+        type: string
+        description: 'Provide the app name'
+        required: true
+      set_latest:
+        type: string
+        default: 'true'
+        description: 'Set latest tag'
+        required: false
+      add_sha_tag:
+        type: string
+        default: 'true'
+        required: false
+      sha_tag:
+        type: string
+        required: false
+
+env:
+  TARGET_IMAGE_TAG: ${{ inputs.target_tag }}
+  IMAGE_REGISTRY: ghcr.io
+  IMAGE_ORG: ${{ github.repository_owner }}
+  APP_NAME: ${{ inputs.app_to_build }}
+  SET_LATEST: ${{ inputs.set_latest }}
+  ADD_SHA_TAG: ${{ inputs.add_sha_tag }}
+  SHA_TAG: ${{ inputs.sha_tag }}
+
+jobs:
+  create-manifest:
+    runs-on: ubuntu-latest
+    timeout-minutes: 60
+    permissions:
+      contents: read
+      packages: write
+      attestations: write
+      id-token: write
+    steps:
+      - name: Download digests
+        uses: actions/download-artifact@v4
+        with:
+          path: /tmp/digests
+          pattern: digests-${{ env.APP_NAME }}-*
+          merge-multiple: true
+          
+
+      - name: Validate inputs and artifacts
+        run: |
+          # Validate target_tag format
+          if ! [[ "$TARGET_IMAGE_TAG" =~ ^[a-zA-Z0-9._-]+$ ]]; then
+            echo "ERROR: Invalid target_tag format: $TARGET_IMAGE_TAG"
+            exit 1
+          fi
+          
+          # Validate digest artifacts exist
+          if [ ! -d /tmp/digests ] || [ -z "$(ls -A /tmp/digests)" ]; then
+            echo "ERROR: No digest artifacts found for ${APP_NAME}"
+            echo "This workflow must run after build-and-push.yml uploads digests"
+            exit 1
+          fi
+
+      - name: Log in to the Container registry
+        uses: docker/login-action@v3
+        with:
+          registry: ${{ env.IMAGE_REGISTRY }}
+          username: ${{ github.actor }}
+          password: ${{ secrets.GITHUB_TOKEN }}
+
+      - name: Set up Docker Buildx
+        id: set_buildx
+        uses: docker/setup-buildx-action@v3
+
+      - name: Create and Push Manifest
+        id: create_manifest
+        working-directory: /tmp/digests
+        run: |
+          set -eo pipefail
+
+          IMAGE="${IMAGE_REGISTRY}/${IMAGE_ORG}/${APP_NAME}"
+          
+          echo "Building sources from digests..."
+          SOURCES=""
+          for digest in *; do
+            # Only accept filenames that look like a 64-character hexadecimal SHA-256 digest
+            if [[ "$digest" =~ ^[a-fA-F0-9]{64}$ ]]; then
+              SOURCES+="${IMAGE}@sha256:${digest} "
+            else
+              echo "Skipping invalid digest file: ${digest}" >&2
+            fi
+          done
+
+          # Validate we have sources
+          if [ -z "$SOURCES" ]; then
+            echo "ERROR: No valid digests found in /tmp/digests"
+            ls -la /tmp/digests
+            exit 1
+          fi
+          
+          # Prepare tags to create manifest for the image.
+          # Always include the target tag, and conditionally include latest and SHA tag based on inputs.
+          TAGS=("${TARGET_IMAGE_TAG}")
+          if [ "${SET_LATEST}" = "true" ]; then
+            TAGS+=("latest")
+          fi
+          if [ "${ADD_SHA_TAG}" = "true" ] && [ -n "${SHA_TAG}" ]; then
+            TAGS+=("${SHA_TAG}")
+          fi
+          
+          # Collecting manifest digest after create and push manifest for each tag
+          MANIFEST_DIGEST=""
+          for tag in "${TAGS[@]}"; do
+            metadata_file="/tmp/${APP_NAME}-${tag}-metadata.json"
+            echo "Creating manifest for ${IMAGE}:${tag} with sources: ${SOURCES}"
+            docker buildx imagetools create \
+              -t "${IMAGE}:${tag}" \
+              --metadata-file "${metadata_file}" \
+              ${SOURCES}
+            digest="$(jq -r '."containerimage.descriptor".digest // empty' "${metadata_file}")"
+            if [ -z "${digest}" ]; then
+              echo "ERROR: No manifest digest found in ${metadata_file} for ${IMAGE}:${tag}"
+              cat "${metadata_file}"
+              exit 1
+            fi
+            if [ -z "${MANIFEST_DIGEST}" ]; then
+              MANIFEST_DIGEST="${digest}"
+              echo "manifest_digest=${MANIFEST_DIGEST}" >> "$GITHUB_OUTPUT"
+            elif [ "${MANIFEST_DIGEST}" != "${digest}" ]; then
+              echo "ERROR: Manifest digest mismatch across tags"
+              echo "First digest: ${MANIFEST_DIGEST}"
+              echo "${tag} digest: ${digest}"
+              exit 1
+            fi
+          done
+          # Log summary
+          echo "## Manifest Build Summary" >> $GITHUB_STEP_SUMMARY
+          echo "- Image: ${IMAGE}" >> $GITHUB_STEP_SUMMARY
+          echo "- Tags: ${TAGS[*]}" >> $GITHUB_STEP_SUMMARY
+          echo "- Manifest digest: ${MANIFEST_DIGEST}" >> $GITHUB_STEP_SUMMARY
+          echo "- Sources: ${SOURCES}" >> $GITHUB_STEP_SUMMARY
+      - name: Generate manifest artifact attestation
+        uses: actions/attest-build-provenance@v1
+        with:
+          subject-name: ${{ env.IMAGE_REGISTRY }}/${{ env.IMAGE_ORG }}/${{ env.APP_NAME }}
+          subject-digest: ${{ steps.create_manifest.outputs.manifest_digest }}
--- a/.github/workflows/create-manifest.yml
+++ b/.github/workflows/create-manifest.yml
@@ -2,7 +2,7 @@ name: Create Multi-Arch Manifest
 on:
   workflow_call:
     secrets:
-      GITHUB_TOKEN:
+      GITHUBTOKEN:
         required: true    
     inputs:
       target_tag:
@@ -73,7 +73,7 @@ jobs:
         with:
           registry: ${{ env.IMAGE_REGISTRY }}
           username: ${{ github.actor }}
-          password: ${{ secrets.GITHUB_TOKEN }}
+          password: ${{ secrets.GITHUBTOKEN }}
 
       - name: Set up Docker Buildx
         id: set_buildx
--- a/.github/workflows/image-builds-master.yml
+++ b/.github/workflows/image-builds-master.yml
@@ -66,7 +66,7 @@ jobs:
 
     uses: ./.github/workflows/build-and-push.yml
     secrets: 
-      GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+      GITHUBTOKEN: ${{ secrets.GITHUB_TOKEN }}
     with:
       src_branch: master
       target_tag: master
@@ -109,7 +109,7 @@ jobs:
       id-token: write
     uses: ./.github/workflows/create-manifest.yml
     secrets:
-      GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+      GITHUBTOKEN: ${{ secrets.GITHUB_TOKEN }}
     with:
       target_tag: master
       app_to_build: ${{ matrix.component.image }}
--- a/.github/workflows/image-builds-master.yml
+++ b/.github/workflows/image-builds-master.yml
@@ -8,54 +8,111 @@ jobs:
     strategy:
       fail-fast: true
       matrix:
-        include:
-          - image: kfp-api-server
+        image:
+          - name: kfp-api-server
             dockerfile: backend/Dockerfile
             context: .
-          - image: kfp-frontend
+          - name: kfp-frontend
             dockerfile: frontend/Dockerfile
             context: .
-          - image: kfp-persistence-agent
+          - name: kfp-persistence-agent
             dockerfile: backend/Dockerfile.persistenceagent
             context: .
-          - image: kfp-scheduled-workflow-controller
+          - name: kfp-scheduled-workflow-controller
             dockerfile: backend/Dockerfile.scheduledworkflow
             context: .
-          - image: kfp-viewer-crd-controller
+          - name: kfp-viewer-crd-controller
             dockerfile: backend/Dockerfile.viewercontroller
             context: .
-          - image: kfp-visualization-server
+          - name: kfp-visualization-server
             dockerfile: backend/Dockerfile.visualization
             context: .
-          - image: kfp-launcher
+          - name: kfp-launcher
             dockerfile: backend/Dockerfile.launcher
             context: .
-          - image: kfp-driver
+          - name: kfp-driver
             dockerfile: backend/Dockerfile.driver
             context: .
-          - image: kfp-cache-deployer
+          - name: kfp-cache-deployer
             dockerfile: backend/src/cache/deployer/Dockerfile
             context: .
-          - image: kfp-cache-server
+          - name: kfp-cache-server
             dockerfile: backend/Dockerfile.cacheserver
             context: .
-          - image: kfp-metadata-writer
-            dockerfile: backend/metadata_writer/Dockerfile
-            context: .
-          - image: kfp-metadata-envoy
+          - name: kfp-metadata-envoy
             dockerfile: third_party/metadata_envoy/Dockerfile
             context: .
-          - image: kfp-inverse-proxy-agent
+          - name: kfp-metadata-writer
+            dockerfile: backend/metadata_writer/Dockerfile
+            context: .
+          - name: kfp-inverse-proxy-agent
             dockerfile: proxy/Dockerfile
             context: ./proxy
+
+        arch:
+          - arch_name: amd64
+            runs_on: ubuntu-latest
+            platform: linux/amd64
+          - arch_name: arm64
+            runs_on: ubuntu-24.04-arm
+            platform: linux/arm64
+
+        exclude:
+          - image: { name: kfp-metadata-writer }
+            arch: { arch_name: arm64 }
+
+          - image: { name: kfp-inverse-proxy-agent }
+            arch: { arch_name: arm64 }
+
     uses: ./.github/workflows/build-and-push.yml
+    secrets: 
+      GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
     with:
       src_branch: master
       target_tag: master
+      arch_tag: ${{ matrix.arch.arch_name }}
       overwrite_imgs: True
       set_latest: False
       add_sha_tag: False
-      app_to_build: ${{ matrix.image }}
-      image_context: ${{ matrix.context }}
-      docker_file: ${{ matrix.dockerfile }}
+      app_to_build: ${{ matrix.image.name }}
+      image_context: ${{ matrix.image.context }}
+      docker_file: ${{ matrix.image.dockerfile }}
+      platforms: ${{ matrix.arch.platform }}
+      runs_on: ${{ matrix.arch.runs_on }}
       push: true
+
+
+
+  create-manifests:
+    needs: build-images-for-master
+    strategy:
+      fail-fast: true
+      matrix:
+        component:
+          - image: kfp-api-server
+          - image: kfp-frontend
+          - image: kfp-persistence-agent
+          - image: kfp-scheduled-workflow-controller
+          - image: kfp-viewer-crd-controller
+          - image: kfp-visualization-server
+          - image: kfp-launcher
+          - image: kfp-driver
+          - image: kfp-cache-deployer
+          - image: kfp-cache-server
+          - image: kfp-metadata-writer
+          - image: kfp-metadata-envoy
+          - image: kfp-inverse-proxy-agent
+    permissions:
+      contents: read
+      packages: write
+      attestations: write
+      id-token: write
+    uses: ./.github/workflows/create-manifest.yml
+    secrets:
+      GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+    with:
+      target_tag: master
+      app_to_build: ${{ matrix.component.image }}
+      set_latest: False
+      add_sha_tag: False
+      
--- a/.github/workflows/image-builds-release.yml
+++ b/.github/workflows/image-builds-release.yml
@@ -95,7 +95,7 @@ jobs:
 
     uses: ./.github/workflows/build-and-push.yml
     secrets: 
-      GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+      GITHUBTOKEN: ${{ secrets.GITHUB_TOKEN }}
     with:
       src_branch: ${{ inputs.src_branch }}
       target_tag: ${{ inputs.target_tag }}
@@ -149,7 +149,7 @@ jobs:
       id-token: write
     uses: ./.github/workflows/create-manifest.yml
     secrets:
-      GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+      GITHUBTOKEN: ${{ secrets.GITHUB_TOKEN }}
     with:
       target_tag: ${{ inputs.target_tag }}
       app_to_build: ${{ matrix.component.image }}
--- a/.github/workflows/image-builds-release.yml
+++ b/.github/workflows/image-builds-release.yml
@@ -94,7 +94,8 @@ jobs:
             arch: { arch_name: arm64 }
 
     uses: ./.github/workflows/build-and-push.yml
-    secrets: inherit
+    secrets: 
+      GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
     with:
       src_branch: ${{ inputs.src_branch }}
       target_tag: ${{ inputs.target_tag }}
--- a/.github/workflows/image-builds-release.yml
+++ b/.github/workflows/image-builds-release.yml
@@ -37,54 +37,121 @@ jobs:
     strategy:
       fail-fast: true
       matrix:
-        include:
-          - image: kfp-api-server
+        image:
+          - name: kfp-api-server
             dockerfile: backend/Dockerfile
             context: .
-          - image: kfp-frontend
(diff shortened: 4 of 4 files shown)


jtu-ampere:
Hello 
I have run [a workflow](https://github.com/jtu-ampere/kubeflow-pipelines/actions/runs/23261623975) by the action of "Build and Push for Release" with the branch `backport-workflow-to-relase-2.16` under my folked repo and the workflow is completed. 
The snapshots below shows the 13 jobs of create-manifests completed and the multi-arch KFP container images are ready


Fig. 1 
image

Fig. 2
image

Fig.3 kfp-api-server:2.16.0
image

.github/workflows/build-and-push.yml
```diff
@@ -132,10 +158,15 @@ jobs:
...
+      - name: Set up Docker Buildx
+        uses: docker/setup-buildx-action@v3
+        with:
+          driver: docker-container
+
       - name: Check if image tag already exists
         id: check_tag
         env:
-          IMAGE: ${{ env.IMAGE_REGISTRY }}/${{ env.IMAGE_ORG }}/${{ inputs.app_to_build }}:${{env.TARGET_IMAGE_TAG}}
+          IMAGE: ${{ env.IMAGE_REGISTRY }}/${{ env.IMAGE_ORG }}/${{ inputs.app_to_build }}:${{env.TARGET_IMAGE_TAG}}-${{inputs.arch_tag}}
```
jeffspahr: This guard now checks only the synthetic target_tag-arch tag, but the build step no longer publishes any arch-suffixed tag because it pushes digest-addressed images and leaves the real target_tag/latest tagging to create-manifest.yml. As a result, rerunning a release with overwrite_imgs=false can still overwrite an existing published manifest instead of failing as the previous workflow did.

.github/workflows/create-manifest.yml
```diff
@@ -0,0 +1,73 @@
...
+            # Only accept filenames that look like a 64-character hexadecimal SHA-256 digest
+            if [[ "$digest" =~ ^[a-fA-F0-9]{64}$ ]]; then
+              SOURCES+="${IMAGE}@sha256:${digest} "
+            else
+              echo "Skipping invalid digest file: ${digest}" >&2
+            fi
+          done
+          
+          echo "Creating manifest for ${IMAGE}:${TARGET_IMAGE_TAG} with sources: ${SOURCES}"
+          docker buildx imagetools create -t "${IMAGE}:${TARGET_IMAGE_TAG}" ${SOURCES}
```
jeffspahr: The release workflow still exposes and forwards add_sha_tag, but the new manifest assembly path only creates the target tag and optional latest tag. That silently drops the SHA-tagged image that the previous release workflow produced when add_sha_tag=true.

jeffspahr (changes requested):
Found two workflow regressions in the multi-arch backport: the overwrite guard no longer protects the published release tags, and add_sha_tag no longer produces a SHA-tagged image in the release flow.

.github/workflows/build-and-push.yml
jtu-ampere: Thanks for your inputs. I've tested the case and fixed this issue. Will remove the "-${{inputs.arch_tag}}" for the overwrite protection. This part is updated.

.github/workflows/create-manifest.yml
jtu-ampere: @jeffspahr I found the issue after several tests. Thanks for pointing it out. I would like to use the solution below:
* In `create-manifest.yml` - override it with `add_sha_tag: false` in the workflow even when add_sha_tag=true.

How do you think?

jtu-ampere:
@jeffspahr I have run several tests for checking the new fixes handling those two issues. Both seems working.

1. The overwriting images is stopped. Find the test workflow via the [link](https://github.com/jtu-ampere/kubeflow-pipelines/actions/runs/23369941209)

Fig. 1
image

Fig. 2
image

2. The `add_sha_tag: false` is added to the local to override the one from the inputs. Find the test workflow via the [link](https://github.com/jtu-ampere/kubeflow-pipelines/actions/runs/23357704299)

Fig. 3
image

.github/workflows/create-manifest.yml
jiayuzhao05: Yes, I think there is a clear fix here. Since the new flow already builds per-arch images and then assembles a multi-arch manifest, the missing piece is to preserve the old add_sha_tag=true behavior in the manifest creation step as well.
Create-manifest.yml should accept add_sha_tag plus the SHA-derived tag value, and when enabled, create an additional manifest tag in the same way it already creates ${target_tag} and optional latest.If SHA tags are no longer intended to be supported, then the workflow contract should be cleaned up explicitly by removing or deprecating that parameter instead of silently changing behavior.

jiayuzhao05 (commented):
I think the multi-arch build/manifest flow is mostly in place now, but I still don’t see the previous add_sha_tag behavior preserved. create-manifest.yml only takes target_tag, app_to_build, and set_latest, and appears to hardcode add_sha_tag:false, so SHA tags from the old release flow may still be dropped silently. Could we thread add_sha_tag through the new manifest workflow as well?

.github/workflows/build-and-push.yml
```diff
@@ -173,24 +200,39 @@ jobs:
         uses: docker/build-push-action@v6
         if: steps.check_tag.outcome == 'success'
         with:
+          platforms: ${{ inputs.platforms }}
           context: ${{ inputs.image_context }}
           file: ${{ inputs.docker_file }}
           push: ${{ inputs.push }}
-          tags: ${{ steps.meta.outputs.tags }}
           labels: ${{ steps.meta.outputs.labels }}
+          outputs: type=image,name=${{ env.IMAGE_REGISTRY }}/${{ env.IMAGE_ORG }}/${{ inputs.app_to_build }},push-by-digest=true,name-canonical=true,push=${{ inputs.push }}
```
jeffspahr: This now builds with push-by-digest/name-canonical output, so this workflow no longer creates the requested tags on its own. Because build-and-push.yml still has a standalone workflow_dispatch path, a manual run will succeed without ever publishing :target_tag or :latest, and the release workflow's add_sha_tag flag is now a silent no-op. To address it, either restore explicit tag creation for the standalone path (for example by keeping metadata-action tags when this workflow is dispatched directly, or by adding a follow-up manifest/tagging job here), or remove workflow_dispatch plus the dead set_latest/add_sha_tag inputs if this file is only meant to be called from the multi-arch wrapper workflows.

jeffspahr (changes requested):
Found one blocking regression in the current head.\n\nThe multi-arch split is fine, but build-and-push.yml still exposes a standalone workflow_dispatch entrypoint while the build step now only pushes anonymous digest references. That means manual runs no longer publish the requested target/latest tags, and add_sha_tag is currently a no-op.\n\nRequesting changes until the standalone/manual contract is either restored or removed.

.github/workflows/build-and-push.yml
jtu-ampere: The `tags: ${{ steps.meta.outputs.tags }}` is making a strange tags with the branch name like `backport-workflow-to-release-2.16` and add `tags: ghcr.io/jtu-ampere/kfp-api-server:backport-workflow-to-release-2.16` when I tested `build-and-push.yml` under my forked repo.  It will change the `image.name` to `ghcr.io/jtu-ampere/kfp-api-server:backport-workflow-to-release-2.16`

Fig. 1
image

Fig. 2
image

FIg. 3  It shows a new tag `backport-workflow-to-release-2.16` 
image

The new fixes are handling the case with the condition highlighted in Fig. 4:

Fig. 4
image

jtu-ampere: The tests are done for 2 categories:
1. 2x direct `workflow_dispatch` to `build-and-push.yml` with `2.16.1-arm64` and `2.16.1-amd64`
2. 1x `workflow_call` from `image-build-release.yml`

image

.github/workflows/build-and-push.yml
```diff
@@ -173,24 +200,46 @@ jobs:
...
         if: steps.check_tag.outcome == 'success'
         with:
+          platforms: ${{ inputs.platforms }}
           context: ${{ inputs.image_context }}
           file: ${{ inputs.docker_file }}
-          push: ${{ inputs.push }}
-          tags: ${{ steps.meta.outputs.tags }}
           labels: ${{ steps.meta.outputs.labels }}
+          # Conditionally switch between standard push and digest-only push
+          outputs: >
```
jeffspahr: This looks like the right split point to restore the standalone/manual contract. Right now the direct `workflow_dispatch` path only publishes `:${target_tag}` here, while `set_latest` and `add_sha_tag` are still advertised inputs. The smallest safe fix would be to make the publish mode explicit with an input like `push_by_digest`/`multi_arch_mode`: keep the current digest-only `outputs:` branch when that is true for the wrapper workflows, but for direct/manual runs fall back to the old tagged push (`push:` plus `tags: ${{ steps.meta.outputs.tags }}`). That preserves the old `target_tag`/`latest`/SHA-tag behavior instead of leaving those inputs as silent no-ops.

.github/workflows/image-builds-release.yml
```diff
@@ -37,54 +37,101 @@ jobs:
...
+          - image: kfp-cache-server
+          - image: kfp-metadata-writer
+          - image: kfp-metadata-envoy
+          - image: kfp-inverse-proxy-agent
+    uses: ./.github/workflows/create-manifest.yml
+    secrets: inherit
+    with:
+      target_tag: ${{ inputs.target_tag }}
+      app_to_build: ${{ matrix.component.image }}
+      set_latest: ${{ inputs.set_latest }}
```
jeffspahr, line 156: Once `create-manifest.yml` accepts the SHA-tag inputs, this caller should thread them through here as well. The smallest pattern would be a tiny prep job that checks out `inputs.src_branch`, computes the source commit SHA (or the exact metadata-action SHA tag string), exposes it as a job output, and then passes `add_sha_tag` plus that derived tag into `create-manifest.yml`. That keeps the old release behavior without trying to rediscover the SHA separately in every matrix leg.

.github/workflows/create-manifest.yml
```diff
@@ -0,0 +1,74 @@
...
+        description: 'Set latest tag'
+        required: false
+
+env:
+  TARGET_IMAGE_TAG: ${{ inputs.target_tag }}
+  IMAGE_REGISTRY: ghcr.io
+  IMAGE_ORG: ${{ github.repository_owner }}
+  APP_NAME: ${{ inputs.app_to_build }}
+  SET_LATEST: ${{ inputs.set_latest }}
+  add_sha_tag: false
```
jeffspahr: Hardcoding `add_sha_tag: false` here does not actually fix the regression because this workflow never reads that variable. I would remove the dead env override and make the manifest job accept either `(add_sha_tag, sha_tag)` or, even cleaner, a generic `tags` input. Then build `SOURCES` once and loop over the final manifest tags (`target_tag`, optional `latest`, optional SHA tag) with `docker buildx imagetools create -t "${IMAGE}:${tag}" ${SOURCES}`. That preserves the old `add_sha_tag=true` contract instead of silently dropping it.

jeffspahr (commented):
Inline suggestions for the remaining workflow contract regressions.

.github/workflows/build-and-push.yml
jtu-ampere: I have tried to include `steps.meta.outputs.tags` as the option for direct/manual runs, but it turned out to use the branch `backport-workflow-to-release-2.16` as a new tag instead of using the `target_tag` (in the test, it's `2.16.10-arm64`) or the others (see the snapshot below).

image

jtu-ampere: I submitted the fixes for `build-and-push.yml`

1. Add the tags back for the step of "Extract metadata (tags, labels) for Build"
2. Add the tags back with ${{ steps.meta.outputs.tags }} for the step of "Build and push Image"

ayushsgit: Hi @jtu-ampere  and @jeffspahr . I've been studying this CI/CD architecture for my GSoC proposal and ran into the same manual tag issues locally.

@jtu-ampere, I saw your recent commit to restore the metadata tags. To help catch the remaining items from the review, I drafted a patch that safely toggles those tags off during digest runs to prevent registry overwrites. It also adds the prep job and dynamically threads the SHA tags through the manifest assembly as Jeff requested. I've fully tested both the standalone and multi-arch paths on my fork.

I've opened a PR against your branch here with the remaining fixes if you want to pull them in: https://github.com/jtu-ampere/kubeflow-pipelines/pull/3

.github/workflows/create-manifest.yml
```diff
@@ -0,0 +1,74 @@
...
+    inputs:
+      target_tag:
+        type: string
+        description: 'Target Image Tag'
+        required: true
+      app_to_build:
+        type: string
+        description: 'Provide the app name'
+        required: true
+      set_latest:
```
jeffspahr, line 16: To preserve the old `add_sha_tag=true` behavior, please add real inputs here instead of hardcoding an env override later. Suggested shape:
```yaml
      add_sha_tag:
        type: string
        default: 'true'
        required: false
      sha_tag:
        type: string
        required: false
```
If you want a cleaner contract, an even better option is a generic `tags` input and letting the caller decide the full final manifest tag list.

.github/workflows/create-manifest.yml
```diff
@@ -0,0 +1,74 @@
...
+          for digest in *; do
+            # Only accept filenames that look like a 64-character hexadecimal SHA-256 digest
+            if [[ "$digest" =~ ^[a-fA-F0-9]{64}$ ]]; then
+              SOURCES+="${IMAGE}@sha256:${digest} "
+            else
+              echo "Skipping invalid digest file: ${digest}" >&2
+            fi
+          done
+          
+          echo "Creating manifest for ${IMAGE}:${TARGET_IMAGE_TAG} with sources: ${SOURCES}"
```
jeffspahr: This is the point where the old SHA-tag behavior needs to be restored. Please remove the dead `add_sha_tag: false` env entry and build the final manifest tag list explicitly here, then loop over it. For example:
```bash
TAGS=("${TARGET_IMAGE_TAG}")
if [ "${SET_LATEST}" = "true" ]; then
  TAGS+=("latest")
fi
if [ "${ADD_SHA_TAG}" = "true" ] && [ -n "${SHA_TAG}" ]; then
  TAGS+=("${SHA_TAG}")
fi
for tag in "${TAGS[@]}"; do
  echo "Creating manifest for ${IMAGE}:${tag} with sources: ${SOURCES}"
  docker buildx imagetools create -t "${IMAGE}:${tag}" ${SOURCES}
done
```
That keeps the multi-arch merge logic in one place and preserves the old release contract.

.github/workflows/image-builds-release.yml
jeffspahr, line 156: Please add a tiny prep job before `create-manifests` to resolve the source SHA once from `inputs.src_branch`, then pass it through here along with `add_sha_tag`. Something like:
```yaml
  resolve-release-sha:
    runs-on: ubuntu-latest
    outputs:
      sha_tag: ${{ steps.sha.outputs.sha_tag }}
    steps:
      - uses: actions/checkout@v5
        with:
          ref: ${{ inputs.src_branch }}
          fetch-depth: 1
      - id: sha
        run: echo "sha_tag=sha-$(git rev-parse --short HEAD)" >> "$GITHUB_OUTPUT"

  create-manifests:
    needs: [build-images-for-release, resolve-release-sha]
    ...
    with:
      target_tag: ${{ inputs.target_tag }}
      app_to_build: ${{ matrix.component.image }}
      set_latest: ${{ inputs.set_latest }}
      add_sha_tag: ${{ inputs.add_sha_tag }}
      sha_tag: ${{ needs.resolve-release-sha.outputs.sha_tag }}
```
If you want exact parity with `docker/metadata-action` SHA formatting, compute that exact tag string in this prep job and pass it through instead of hardcoding it in `create-manifest.yml`.

jeffspahr (commented):
Inline fix suggestions for the remaining SHA-tag gap in the release flow.

.github/workflows/image-builds-release.yml
jtu-ampere: @jeffspahr 
The fixes are applied. 
The tests are done on my forked repo.
The tags are showed on each container image entry on the package page. For limited space, three container images are listed as the test results. 

Fig. 1 [kfp-api-server](https://github.com/jtu-ampere/kubeflow-pipelines/pkgs/container/kfp-api-server)
image

Fig. 2 [kfp-driver](https://github.com/jtu-ampere/kubeflow-pipelines/pkgs/container/kfp-driver)
image

Fig. 3 [kfp-persistence-agent](https://github.com/jtu-ampere/kubeflow-pipelines/pkgs/container/kfp-persistence-agent)
image

.github/workflows/image-builds-master.yml
```diff
@@ -8,54 +8,114 @@ jobs:
...
+          - image: kfp-metadata-writer
+          - image: kfp-metadata-envoy
+          - image: kfp-inverse-proxy-agent
+    uses: ./.github/workflows/create-manifest.yml
+    secrets: inherit
+    with:
+      target_tag: master
+      app_to_build: ${{ matrix.component.image }}
+      set_latest: 'false'
+      add_sha_tag: ${{ inputs.add_sha_tag }}
```
jeffspahr: This workflow runs on `push`, so there is no `inputs` context here. If the intent for this backport is to keep master builds without a SHA manifest tag, hardcode `'false'` here instead of reading a nonexistent input.

```suggestion
      add_sha_tag: 'false'
```

jeffspahr:
One remaining issue in the current head:

`image-builds-master.yml` now passes `${{ inputs.add_sha_tag }}` into `create-manifest.yml`, but this workflow is triggered by `push`, so there is no `inputs` context in scope here. `actionlint` flags the reference at `.github/workflows/image-builds-master.yml:120`, and in practice it means the new SHA-manifest path in the master workflow is broken or silently disabled.

The smallest safe fix in this backport is to hardcode `add_sha_tag: 'false'` for the master workflow (or remove the now-dead `resolve-release-sha` / `sha_tag` plumbing from this path entirely). The release workflow can keep threading the real `inputs.add_sha_tag` value through because that one is `workflow_dispatch`-based.

Verification: reviewed PR head `a9ad7602b05315773a407ab77acc527a8500102a` against `origin/release-2.16` and ran `actionlint` on the isolated worktree; the changed-file findings were this undefined `inputs` reference and the stale `steps.buildx.outputs.name` reference in `build-and-push.yml`.

.github/workflows/image-builds-master.yml
jtu-ampere: Should we do the both for the `image-builds-master.yml` workflow:
1. Hardcode `add_sha_tag: 'false'` 
2. Remove the step resolve-release-sha from this path entirely

jeffspahr: Yes, I would do both here. Hardcoding `add_sha_tag: 'false'` is the actual fix for the undefined `inputs` reference, and once that is in place the `resolve-release-sha` job becomes dead code on the master path because nothing should consume `sha_tag` anymore. Removing it in the same patch makes the intent explicit and avoids an unnecessary checkout job.

.github/workflows/build-and-push.yml
```diff
@@ -173,24 +205,47 @@ jobs:
...
+          # Conditionally switch between standard push and digest-only push
+          outputs: > 
+            ${{ github.event_name == 'workflow_dispatch' && 
+                format('type=image,name={0}/{1}/{2}:{3},push={4}', env.IMAGE_REGISTRY, env.IMAGE_ORG, inputs.app_to_build, inputs.target_tag, inputs.push) || 
+                format('type=image,name={0}/{1}/{2},push-by-digest=true,name-canonical=true,push={3}', env.IMAGE_REGISTRY, env.IMAGE_ORG, inputs.app_to_build, inputs.push) }}
           build-args: |
             COMMIT_SHA=${{ steps.get_commit.outputs.sha }}
             COMMIT_HASH=${{ steps.get_commit.outputs.sha }}
             TAG_NAME=${{ inputs.target_tag }}
+          builder: ${{ steps.buildx.outputs.name }}
```
jeffspahr, line 225: One more issue on the current head: this still references `steps.buildx`, but the setup step above is `id: set_buildx`, so `actionlint` fails on the current branch. The minimal fix is:
```suggestion
          builder: ${{ steps.set_buildx.outputs.name }}
```

jtu-ampere: Yes, I found that issue as well. It will be fixed.

.github/workflows/image-builds-master.yml
jtu-ampere: Thanks for your confirmation. The new fixes are applied. Please review. Thanks.

.github/workflows/image-builds-master.yml
```diff
@@ -8,54 +8,104 @@ jobs:
...
+          - image: kfp-cache-server
+          - image: kfp-metadata-writer
+          - image: kfp-metadata-envoy
+          - image: kfp-inverse-proxy-agent
+    uses: ./.github/workflows/create-manifest.yml
+    secrets: inherit
+    with:
+      target_tag: master
+      app_to_build: ${{ matrix.component.image }}
+      set_latest: 'false'
```
jeffspahr: `actionlint` still flags this caller because the reusable workflow declares `set_latest` as a string input. Using the double-quoted string form here keeps the intent the same and clears the lint failure.
```suggestion
      set_latest: "false"
```

.github/workflows/image-builds-master.yml
```diff
@@ -8,54 +8,104 @@ jobs:
...
+      add_sha_tag: 'false'
```
jeffspahr: Same issue here for `add_sha_tag`: the current single-quoted literal is linted as a bool in this reusable-workflow call.
```suggestion
      add_sha_tag: "false"
```

jeffspahr (commented):
Current head `e8582cd4` resolves the earlier undefined-`inputs` and stale buildx-id issues from the previous revision. One small issue remains in the current lint pass: `actionlint` still flags the master manifest caller because the single-quoted `'false'` literals at `.github/workflows/image-builds-master.yml:109-110` are treated as bools when passed into string-typed inputs on the reusable workflow. Switching those two values to `"false"` (or `${{ '''false''' }}`) makes the changed-workflow `actionlint` run pass cleanly.

.github/workflows/build-and-push.yml
jeffspahr: Current head e8582cd4 keeps the overwrite guard on the published target tag again, so this one is addressed.

.github/workflows/create-manifest.yml
jeffspahr: Current head threads `add_sha_tag` and `sha_tag` through `create-manifest.yml` and loops the final manifest tag list, so the release SHA-tag behavior is restored.

.github/workflows/build-and-push.yml
jeffspahr: Current head restores a separate non-digest export path for direct `workflow_dispatch` runs while keeping digest-only output for the wrapper workflows, so this contract gap is addressed.

.github/workflows/build-and-push.yml
jeffspahr: Current head addresses this with the event-based output split instead of a new input: manual runs keep tagged pushes, while the wrapper workflows stay on digest-only output.

.github/workflows/image-builds-release.yml
jeffspahr: Current head adds `resolve-release-sha` on the release path and threads both `add_sha_tag` and `sha_tag` into `create-manifest.yml`, so this caller-side gap is fixed.

.github/workflows/create-manifest.yml
jeffspahr: Current head now accepts `add_sha_tag` and `sha_tag` as real manifest inputs and builds the final tag list in one place, so the old dead override is gone.

.github/workflows/create-manifest.yml
jeffspahr: Current head adds these inputs to `create-manifest.yml`, so this part is addressed.

.github/workflows/create-manifest.yml
jeffspahr: Current head now builds the manifest `TAGS` list and loops over target/latest/SHA tags here, which covers this suggestion.

.github/workflows/image-builds-master.yml
jeffspahr: Current head hardcodes master SHA tagging off and removes the dead master `resolve-release-sha` path, so this is fixed.

.github/workflows/build-and-push.yml
jeffspahr: Current head renames the setup step to `id: buildx`, so this reference is now valid.

.github/workflows/image-builds-master.yml
jtu-ampere: The suggestion is applied.

.github/workflows/image-builds-master.yml
jtu-ampere: The suggestion is applied.

.github/workflows/image-builds-master.yml
```diff
@@ -8,54 +8,104 @@ jobs:
...
+          - image: kfp-metadata-writer
+          - image: kfp-metadata-envoy
+          - image: kfp-inverse-proxy-agent
+    uses: ./.github/workflows/create-manifest.yml
+    secrets: inherit
+    with:
+      target_tag: master
+      app_to_build: ${{ matrix.component.image }}
+      set_latest: "false"
+      add_sha_tag: "false"
```
jeffspahr: I got the earlier `"false"` suggestion wrong. `actionlint` still treats that form as bool-like at this reusable-workflow call site, so it does not clear the remaining failure.

To keep the file consistent, I would use bare `False` in both reusable-workflow caller blocks in this file, including these two lines:

```suggestion
      set_latest: False
      add_sha_tag: False
```

I rechecked this form locally against the touched workflows, and it clears the remaining lint failure. `${{ 'false' }}` also lint-passes here, but I do not want to recommend mixing two different spellings in the same file.

jtu-ampere: @jeffspahr 

Thanks for your inputs. It's good to evaluate the cases to make sure the workflows delivering good outcomes.
I have a question for the two inputs.

Are the same inputs to the step on `build-and-push.yml` with `add_sha_tag: False` and `set_latest: False` (see Fig. 1)? 
If yes, should we just use `False` without " or '?

Fig. 1
image

jeffspahr: Yes, same idea here. `build-and-push.yml` and `create-manifest.yml` both declare these as string inputs, and I rechecked this locally.

For this caller, bare `False` passes `actionlint`, so the smallest fix is to make these two lines match the existing `build-and-push.yml` call above:

```suggestion
      set_latest: False
      add_sha_tag: False
```

`${{ 'false' }}` also works, but since the earlier call site already uses `False`, matching that existing pattern keeps the patch smaller and more consistent.

jeffspahr: Good call out.  I updated the inline comment to use False for consistency.

jeffspahr (approved):
No remaining blocking or meaningful review findings on the current head `1b6b8f559045b1f799a92dbedc3340c9ed882cfd`.

The only delta since the last reviewed head was in `.github/workflows/image-builds-master.yml`, where the `create-manifest.yml` caller now uses `False` for both `set_latest` and `add_sha_tag`. That clears the prior `actionlint` type complaint on the reusable-workflow string inputs.

Verification:
- diffed the current head against both `origin/release-2.16` and the prior reviewed head `62a6beec964400010a00647e2337420831df71ff`
- ran `go run github.com/rhysd/actionlint/cmd/actionlint@latest .github/workflows/build-and-push.yml .github/workflows/create-manifest.yml .github/workflows/image-builds-master.yml .github/workflows/image-builds-release.yml`

`actionlint` passes on the touched workflows. I did not run the workflows on GitHub, so runtime behavior is still assessed by static review plus lint only.

/lgtm

.github/workflows/image-builds-master.yml
```diff
@@ -8,54 +8,104 @@ jobs:
...
+
+        exclude:
+          - image: { name: kfp-metadata-writer }
+            arch: { arch_name: arm64 }
+
+          - image: { name: kfp-inverse-proxy-agent }
+            arch: { arch_name: arm64 }
+
     uses: ./.github/workflows/build-and-push.yml
+    secrets: inherit
```
mprahl, line 68: ** 🤖 AI Review **

`secrets: inherit` widens the credential boundary for this reusable workflow chain even though the callee appears to need only the automatic `GITHUB_TOKEN`. That makes every caller secret available to the downstream workflow and its third-party actions, and the new `create-manifest.yml` also requests `id-token: write` without obviously using OIDC or attestation there. Please prefer passing only the specific secret(s) actually needed and drop the unused `id-token: write` permission.

.github/workflows/create-manifest.yml
```diff
@@ -0,0 +1,126 @@
...
+          if [ "${SET_LATEST}" = "true" ]; then
+            TAGS+=("latest")
+          fi          
+          if [ "${ADD_SHA_TAG}" = "true" ] && [ -n "${SHA_TAG}" ]; then
+            TAGS+=("${SHA_TAG}")
+          fi
+          # Create and push manifest for each tag
+          for tag in "${TAGS[@]}"; do
+            echo "Creating manifest for ${IMAGE}:${tag} with sources: ${SOURCES}"
+            docker buildx imagetools create -t "${IMAGE}:${tag}" ${SOURCES}
```
mprahl, line 115: ** 🤖 AI Review **

This publishes the user-facing multi-arch tags, but the workflow never captures or attests the resulting manifest-list digest. `build-and-push.yml` still attests the per-arch image digests, so after this change consumers pulling `master`/release/`latest` verify a different digest than the one covered by provenance. Please capture the digest produced here and run `actions/attest-build-provenance` for the final manifest digest as well.

.github/workflows/image-builds-master.yml
jtu-ampere: Just applied the fixes to the two workflows (image-builds-master.yml and create-manifest.yml) for the review.

.github/workflows/create-manifest.yml
```diff
@@ -0,0 +1,125 @@
...
+  ADD_SHA_TAG: ${{ inputs.add_sha_tag }}
+  SHA_TAG: ${{ inputs.sha_tag }}
+
+jobs:
+  create-manifest:
+    runs-on: ubuntu-latest
+    timeout-minutes: 60
+    permissions:
+      contents: read
+      packages: write
```
jeffspahr: Since this workflow would now call `actions/attest-build-provenance`, the OIDC and attestation scopes are no longer unused here. Keep them scoped to this manifest job, alongside the existing package write permission used for GHCR.

```suggestion
    permissions:
      contents: read
      packages: write
      attestations: write
      id-token: write
```

jeffspahr (commented):
Added inline suggestions for the remaining manifest attestation path. The suggestions keep `secrets: inherit` out of the create-manifest call, add the OIDC/attestation permissions only where the attestation action runs, and capture the final manifest-list digest before calling `actions/attest-build-provenance`.

.github/workflows/create-manifest.yml
```diff
@@ -0,0 +1,125 @@
...
+          
+          # Log summary
+          echo "## Manifest Build Summary" >> $GITHUB_STEP_SUMMARY
+          echo "- Image: ${IMAGE}" >> $GITHUB_STEP_SUMMARY
+          echo "- Tags: ${TARGET_IMAGE_TAG}" >> $GITHUB_STEP_SUMMARY
+          if [ "${SET_LATEST}" == "true" ]; then
+            echo "- Latest: true" >> $GITHUB_STEP_SUMMARY
+          fi
+          echo "- Sources: ${SOURCES}" >> $GITHUB_STEP_SUMMARY
+
```
jeffspahr: Capture the manifest-list digest from `imagetools create --metadata-file` and attest that digest after the tags are pushed. This covers the digest consumers get when pulling the final multi-arch tag instead of only the per-arch build digests.

```suggestion
      - name: Create and Push Manifest
        id: create_manifest
        working-directory: /tmp/digests
        run: |
          set -eo pipefail

          IMAGE="${IMAGE_REGISTRY}/${IMAGE_ORG}/${APP_NAME}"

          echo "Building sources from digests..."
          SOURCES=""
          for digest in *; do
            # Only accept filenames that look like a 64-character hexadecimal SHA-256 digest
            if [[ "$digest" =~ ^[a-fA-F0-9]{64}$ ]]; then
              SOURCES+="${IMAGE}@sha256:${digest} "
            else
              echo "Skipping invalid digest file: ${digest}" >&2
            fi
          done

          # Validate we have sources
          if [ -z "$SOURCES" ]; then
            echo "ERROR: No valid digests found in /tmp/digests"
            ls -la /tmp/digests
            exit 1
          fi

          # Prepare tags to create manifest for the image.
          # Always include the target tag, and conditionally include latest and SHA tag based on inputs.
          TAGS=("${TARGET_IMAGE_TAG}")
          if [ "${SET_LATEST}" = "true" ]; then
            TAGS+=("latest")
          fi
          if [ "${ADD_SHA_TAG}" = "true" ] && [ -n "${SHA_TAG}" ]; then
            TAGS+=("${SHA_TAG}")
          fi

          MANIFEST_DIGEST=""
          for tag in "${TAGS[@]}"; do
            metadata_file="/tmp/${APP_NAME}-${tag}-metadata.json"
            echo "Creating manifest for ${IMAGE}:${tag} with sources: ${SOURCES}"
            docker buildx imagetools create \
              -t "${IMAGE}:${tag}" \
              --metadata-file "${metadata_file}" \
              ${SOURCES}

            digest="$(jq -r '."containerimage.descriptor".digest // empty' "${metadata_file}")"
            if [ -z "${digest}" ]; then
              echo "ERROR: No manifest digest found in ${metadata_file}"
              cat "${metadata_file}"
              exit 1
            fi

            if [ -z "${MANIFEST_DIGEST}" ]; then
              MANIFEST_DIGEST="${digest}"
              echo "manifest_digest=${MANIFEST_DIGEST}" >> "$GITHUB_OUTPUT"
            elif [ "${MANIFEST_DIGEST}" != "${digest}" ]; then
              echo "ERROR: Manifest digest mismatch across tags"
              echo "First digest: ${MANIFEST_DIGEST}"
              echo "${tag} digest: ${digest}"
              exit 1
            fi
          done

          # Log summary
          echo "## Manifest Build Summary" >> $GITHUB_STEP_SUMMARY
          echo "- Image: ${IMAGE}" >> $GITHUB_STEP_SUMMARY
          echo "- Tags: ${TAGS[*]}" >> $GITHUB_STEP_SUMMARY
          echo "- Manifest digest: ${MANIFEST_DIGEST}" >> $GITHUB_STEP_SUMMARY
          echo "- Sources: ${SOURCES}" >> $GITHUB_STEP_SUMMARY

      - name: Generate manifest artifact attestation
        uses: actions/attest-build-provenance@v1
        with:
          subject-name: ${{ env.IMAGE_REGISTRY }}/${{ env.IMAGE_ORG }}/${{ env.APP_NAME }}
          subject-digest: ${{ steps.create_manifest.outputs.manifest_digest }}
```

.github/workflows/image-builds-master.yml
```diff
@@ -8,54 +8,106 @@ jobs:
...
+          - image: kfp-launcher
+          - image: kfp-driver
+          - image: kfp-cache-deployer
+          - image: kfp-cache-server
+          - image: kfp-metadata-writer
+          - image: kfp-metadata-envoy
+          - image: kfp-inverse-proxy-agent
+    uses: ./.github/workflows/create-manifest.yml
+    secrets: 
+      GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
```
jeffspahr: Keep the secret boundary narrow, but explicitly pass the token permissions needed by the called workflow once it performs the manifest attestation. This avoids returning to `secrets: inherit` while allowing the reusable workflow to request the OIDC token.

```suggestion
    permissions:
      contents: read
      packages: write
      attestations: write
      id-token: write
    uses: ./.github/workflows/create-manifest.yml
    secrets:
      GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
```

.github/workflows/image-builds-release.yml
```diff
@@ -37,54 +37,115 @@ jobs:
...
+          - image: kfp-visualization-server
+          - image: kfp-launcher
+          - image: kfp-driver
+          - image: kfp-cache-deployer
+          - image: kfp-cache-server
+          - image: kfp-metadata-writer
+          - image: kfp-metadata-envoy
+          - image: kfp-inverse-proxy-agent
+    uses: ./.github/workflows/create-manifest.yml
+    secrets: inherit
```
jeffspahr: The release caller needs the same manifest-attestation permissions, and this can also avoid passing every caller secret into `create-manifest.yml`.

```suggestion
    permissions:
      contents: read
      packages: write
      attestations: write
      id-token: write
    uses: ./.github/workflows/create-manifest.yml
    secrets:
      GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
```

.github/workflows/image-builds-master.yml
jtu-ampere: Thanks for your insight. The suggestion is added

.github/workflows/create-manifest.yml
jtu-ampere: Thanks for your insight. The suggestion is added

.github/workflows/create-manifest.yml
jtu-ampere: Thanks for your insight. The suggestion is added

.github/workflows/image-builds-release.yml
jtu-ampere: Thanks for your insight. The suggestion is added

.github/workflows/create-manifest.yml
jtu-ampere: Thanks for your insight. The fixes are applied.

jeffspahr:
The caller-side secret narrowing was applied, but the reusable workflows still need to declare the named secret they accept. `actionlint` currently reports the explicit `GITHUB_TOKEN` mappings as invalid because neither callee defines `on.workflow_call.secrets.GITHUB_TOKEN`.

GitHub will not let me attach an applyable inline suggestion to the unchanged `workflow_call` line in `build-and-push.yml`, so here are the exact snippets to apply.

In `.github/workflows/build-and-push.yml`, change the start of the reusable workflow block to:

```yaml
on:
  workflow_call:
    secrets:
      GITHUB_TOKEN:
        required: true
    inputs:
      src_branch:
```

In `.github/workflows/create-manifest.yml`, change the start of the reusable workflow block to:

```yaml
on:
  workflow_call:
    secrets:
      GITHUB_TOKEN:
        required: true
    inputs:
      target_tag:
```

After that, rerun:

```bash
go run github.com/rhysd/actionlint/cmd/actionlint@latest \
  .github/workflows/build-and-push.yml \
  .github/workflows/create-manifest.yml \
  .github/workflows/image-builds-master.yml \
  .github/workflows/image-builds-release.yml
```

jeffspahr (commented):
One remaining workflow regression after the secret declaration update.

.github/workflows/build-and-push.yml
```diff
@@ -130,7 +159,13 @@ jobs:
         with:
           registry: ${{ env.IMAGE_REGISTRY }}
           username: ${{ github.actor }}
-          password: ${{ secrets.GITHUB_TOKEN }}
+          password: ${{ secrets.GITHUBTOKEN }}
```
jeffspahr, line 162: [P1] Preserve the automatic token name for manual runs

The reusable callers now pass `GITHUBTOKEN`, so the workflow-call path validates, but this file is still directly runnable via `workflow_dispatch`. Direct dispatches do not get a mapped `GITHUBTOKEN` secret; GitHub exposes the automatic token as `secrets.GITHUB_TOKEN`. A manual run with `push=true` will reach Docker login with an empty/missing password. Keep the declared reusable secret named `GITHUB_TOKEN` and use `secrets.GITHUB_TOKEN` here, with callers mapping `GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}`.

.github/workflows/build-and-push.yml
```diff
@@ -91,6 +111,15 @@ on:
...
         description: 'Provide the docker file name'
         required: true
+      platforms:
+        type: string
+        default: ''
+        required: true
+        description: 'Provide the target architecture e.g, linux/amd64, linux/arm64'
+      runs_on:
+        type: string
+        required: true
```
mprahl, line 121: ** 🤖 AI Review **

Severity: high

`workflow_dispatch` is no longer self-contained here: `runs_on` is required with no default, so a manual dispatch can fail before the job even starts unless the user knows to provide a runner label explicitly. This same direct path also depends later on a mapped `GITHUBTOKEN`, so the old one-click manual contract is now much easier to break. Please restore safe manual defaults such as `ubuntu-latest` and `linux/amd64`, and keep the direct-dispatch credential path compatible with the built-in GitHub token.

.github/workflows/build-and-push.yml
```diff
@@ -173,24 +208,47 @@ jobs:
...
+          cache-from: type=gha,scope=${{ inputs.app_to_build }}-${{ inputs.arch_tag }}
+          cache-to: type=gha,mode=max,scope=${{ inputs.app_to_build }}-${{ inputs.arch_tag }}
+
+
+      # Export the digest
+      - name: Export digest
+        if: steps.check_tag.outcome == 'success' && inputs.push
+        run: |
+          mkdir -p /tmp/digests
+          digest="${{ steps.push.outputs.digest }}"
```
mprahl, line 235: ** 🤖 AI Review **

Severity: medium

This assumes `steps.push.outputs.digest` is always present. If `docker/build-push-action` ever returns an empty or unexpected digest here, the workflow will create a bogus artifact filename and fail later in `create-manifest.yml` with a much less actionable "No valid digests found" error. Please add an explicit non-empty digest check before exporting, uploading, or attesting it so this fails fast at the real source of the problem.

.github/workflows/image-builds-release.yml
```diff
@@ -37,54 +37,122 @@ jobs:
...
+    runs-on: ubuntu-latest
+    outputs:
+      sha_tag: ${{ steps.sha.outputs.sha_tag }}
+    steps:
+      - uses: actions/checkout@v5
+        with:
+          ref: ${{ inputs.src_branch }}
+          fetch-depth: 1
+      - id: sha
+        run: echo "sha_tag=sha-$(git rev-parse --short HEAD)" >> "$GITHUB_OUTPUT"
```
mprahl, line 123: ** 🤖 AI Review **

Severity: medium

This derives `sha_tag` from a fresh checkout of `inputs.src_branch`, but the build matrix above still passes the same mutable branch name into each `build-and-push.yml` invocation. If `src_branch` moves while this workflow is running, the published `sha-...` tag can point at a different commit than the images that were actually built and assembled into the manifest. Please resolve the exact commit once and thread that immutable SHA through the build jobs as well.

droctothorpe:
No longer necessary since 2.17 was released.

PR #13083, opened 2026-03-19
fix: replace bare except with except Exception in KFP client

harshadkhetpal:
## Summary

Replace two bare `except:` clauses with `except Exception:` in `sdk/python/kfp/client/client.py`.

**Changes (lines 305 and 316):**
```python
# Before
except:
    in_cluster = False

# After
except Exception:
    in_cluster = False
```

Bare `except:` catches BaseException including KeyboardInterrupt and SystemExit. These two blocks are config-loading fallbacks that should only handle normal exceptions. Using `except Exception:` is the correct scope.

## Testing
No behavior change for normal exception paths. Ctrl-C signals now propagate correctly instead of being silently swallowed.

### DIFF
--- a/sdk/python/kfp/client/client.py
+++ b/sdk/python/kfp/client/client.py
@@ -302,7 +302,7 @@ class Client:
         in_cluster = True
         try:
             k8s.config.load_incluster_config()
-        except:
+        except Exception:
             in_cluster = False
 
         if in_cluster:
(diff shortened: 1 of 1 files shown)


jiayuzhao05 (commented):
Replacing both bare except: blocks with except Exception: addresses the E722 issue while preserving the intended fallback behavior in config loading. LGTM.

PR #13084, opened 2026-03-19
fix: replace bare `except:` with `except Exception:` in SDK

harshadkhetpal:
## Summary

Replace bare `except:` clauses with `except Exception:` across three SDK files to comply with PEP 8 E722.

Bare `except` clauses catch everything including `KeyboardInterrupt` and `SystemExit`, which is rarely intended for fallback/default handling.

**Files changed:**
- `sdk/python/kfp/dsl/v1_modelbase.py` (line 135): type conversion fallback
- `sdk/python/kfp/client/client.py` (lines 305, 316): k8s config loading fallback
- `sdk/python/kfp/dsl/types/type_utils.py` (line 113): JSON parsing fallback

## Testing
No behavior change for normal exceptions — `except Exception:` covers all standard exceptions while allowing `KeyboardInterrupt` and `SystemExit` to propagate correctly.

### DIFF
--- a/sdk/python/kfp/client/client.py
+++ b/sdk/python/kfp/client/client.py
@@ -302,7 +302,7 @@ class Client:
         in_cluster = True
         try:
             k8s.config.load_incluster_config()
-        except:
+        except Exception:
             in_cluster = False
 
         if in_cluster:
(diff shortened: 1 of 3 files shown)


jiayuzhao05 (commented):
Replaced the bare except: blocks with except Exception: in all targeted SDK locations. This addresses the E722 concern without changing the intended fallback behavior. LGTM.

PR #13094, opened 2026-03-20
refactor(apiserver): extract buildHTTPRouter for testability

jsonmp-k8:
## Summary
- Extract route registration from `startHTTPProxy()` into a standalone `buildHTTPRouter()` function that returns `*mux.Router` without starting a listener
- Define `HTTPRouterDeps` struct so tests can supply lightweight handler stubs without constructing real `ResourceManager`/server instances
- Make `/metrics` registration conditional on `collectMetrics` flag (was always-on before)
- Add 5 test functions covering all routes, healthz responses, handler dispatch, metrics toggle, and gRPC gateway catch-all

## Test plan
- [ ] CI `backend-tests` pass (all existing + new `TestBuildHTTPRouter_*` tests)
- [ ] `go vet ./backend/src/apiserver/...` passes
- [ ] `golangci-lint` reports no new issues


jsonmp-k8:
**Fix: Register /metrics unconditionally (2d78b5de)**

The `collectMetricsFlag` only controls per-handler counters in the apiserver, but `grpc_prometheus.UnaryServerInterceptor` is wired unconditionally in `startRPCServer` — so Prometheus and Go runtime metrics are always produced regardless of the flag. Wrapping `/metrics` in `if collectMetrics` was a behavior regression that would 404 existing scrape targets while the interceptor still incurs the overhead of producing unscrapable metrics.

Removed the `collectMetrics` parameter from `buildHTTPRouter` and always register `/metrics`, matching the pre-refactor behavior. Also removed the `TestBuildHTTPRouter_MetricsConditional` test that asserted the incorrect conditional behavior.

jeffspahr:
Re-reviewed the current head `2d78b5de40d48b97f1e4ef1479b94319250303ac` against refreshed `origin/master`, and I don't see blocking issues in the current diff.

The latest follow-up commit fixes the one contract concern from the prior revision by restoring unconditional `/metrics` registration while keeping the router extraction and the JSON-safe healthz handlers. The extracted router helper still preserves the explicit route ordering relative to the `/apis/` gateway catch-all, and the tests directly cover route registration, healthz payloads, handler dispatch, and unmatched gateway routing.

Verification I ran locally in an isolated worktree:
- `go test ./backend/src/apiserver/...`
- `go vet ./backend/src/apiserver/...`

CI note as of 2026-03-28: `backend-tests` and the API-server matrices are green. The remaining red jobs are two E2E critical shards (`K8s v1.29.2 / Argo v3.5.14 / cacheEnabled=false` and `K8s v1.34.0 / Argo v3.6.7 / cacheEnabled=false`). Spot-checking those logs shows unrelated pipeline-runtime failures and Ginkgo panics in existing E2E scenarios, not router, healthz, or `/metrics` regressions.

droctothorpe:
/approve
/lgtm

Nice refactor! 🎉

PR #13095, opened 2026-03-20
fix: prevent stale ci-passed labels

jsonmp-k8:
## Summary
- remove stale `ci-passed` labels immediately on `synchronize` and `reopened`
- re-add `ci-passed` only after a successful `CI Check` on a PR that is still label-eligible
- remove the artifact handoff from `ci-checks.yml` and derive PR metadata directly from `workflow_run`

## Validation
- parsed both updated workflow files with Ruby YAML loading

### DIFF
--- a/.github/workflows/ci-checks.yml
+++ b/.github/workflows/ci-checks.yml
@@ -9,12 +9,14 @@ jobs:
   check_ci_status:
     runs-on: ubuntu-latest
     # Only run for eligible PRs: must have 'ok-to-test' and not 'needs-ok-to-test'.
-    # For 'unlabeled' events, only react when 'needs-ok-to-test' is removed (the only
-    # label removal that can change eligibility), to avoid restarting the 50-min polling
-    # loop on unrelated label changes like 'size/L' or 'lgtm'.
+    # For 'labeled' events, only react when 'ok-to-test' is added.
+    # For 'unlabeled' events, only react when 'needs-ok-to-test' is removed.
+    # These are the only label changes that can change eligibility, avoiding
+    # restarting the 50-min polling loop on unrelated labels like 'size/L' or 'lgtm'.
     # Using a job-level condition so ineligible PRs are 'skipped' rather than 'success',
     # which prevents a wasted downstream workflow_run trigger.
     if: >
+      (github.event.action != 'labeled' || github.event.label.name == 'ok-to-test') &&
       (github.event.action != 'unlabeled' || github.event.label.name == 'needs-ok-to-test') &&
       contains(github.event.pull_request.labels.*.name, 'ok-to-test') &&
       !contains(github.event.pull_request.labels.*.name, 'needs-ok-to-test')


.github/workflows/ci-checks.yml
```diff
@@ -1,21 +1,17 @@
-# This workflow checks if all CI checks have passed by polling every 5 minutes for a total of 8 attempts.
+# This workflow checks if all CI checks have passed by polling every 5 minutes.
 name: CI Check
 
 on:
   pull_request_target:
-    types: [opened, synchronize, reopened, labeled]
+    types: [opened, synchronize, reopened, labeled, unlabeled]
```
manaswinidas, line 6: Adding `unlabeled` here means removing *any* label (e.g. `size/L`, `lgtm`) re-triggers this workflow, which then polls checks for up to 50 minutes. The only scenario where `unlabeled` is useful is removing `needs-ok-to-test` to make a PR eligible.

Consider guarding the polling step with an additional check that the removed label is actually `needs-ok-to-test` (via `github.event.label.name`), or skip `unlabeled` entirely and handle the `needs-ok-to-test` removal case a different way -- otherwise this will burn unnecessary runner minutes on every unrelated label removal.

.github/workflows/ci-checks.yml
```diff
@@ -1,21 +1,17 @@
...
 jobs:
   check_ci_status:
     runs-on: ubuntu-latest
     permissions:
       checks: read
     steps:
-      - name: Check out the repository
-        uses: actions/checkout@v6
-
       - name: Check for 'needs-ok-to-test' and 'ok-to-test' labels
```
manaswinidas: Nit: this label check exits 0 when `needs-ok-to-test` is found or `ok-to-test` is absent, so the overall workflow "succeeds" and triggers a downstream `workflow_run` with `conclusion: 'success'`. That fires the `fetch_data` job in `add-ci-passed-label.yml` (which correctly marks it ineligible, so no label is added), but it's still a wasted runner invocation.

Consider promoting this to a job-level `if` condition so the job is `skipped` rather than `success`, which would prevent the `workflow_run` from firing at all.

.github/workflows/add-ci-passed-label.yml
```diff
@@ -1,88 +1,120 @@
...
+# or its eligibility labels change, then adds it back only after the
+# 'CI Check' workflow completes successfully for a label-eligible PR.
 name: Add CI Passed Label
 
 on:
+  pull_request_target:
+    types:
+      - synchronize
+      - reopened
+      - labeled
```
manaswinidas, line 11: These `labeled`/`unlabeled` types fire on *every* label change across all PRs (e.g. `size/L`, `lgtm`), not just eligibility-related ones. The shell logic below correctly exits early for irrelevant changes, but each event still spins up a runner.

Same concern applies to `unlabeled` added in `ci-checks.yml` -- there it's worse because a stray label removal restarts the full 50-minute CI polling loop.

For this workflow the overhead is minimal (quick shell exit), but you could restrict `labeled`/`unlabeled` handling to the `fetch_data` path if you add `github.event.label.name` guards.

.github/workflows/add-ci-passed-label.yml
```diff
@@ -1,88 +1,81 @@
...
-            const download = await github.rest.actions.downloadArtifact({
+            const { data: pullRequest } = await github.rest.pulls.get({
               owner: context.repo.owner,
               repo: context.repo.repo,
-              artifact_id: matchArtifact.id,
-              archive_format: 'zip',
+              pull_number: pr.number,
             });
-            const fs = require('fs');
-            fs.writeFileSync('${{github.workspace}}/pr.zip', Buffer.from(download.data));
```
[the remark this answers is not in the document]
jsonmp-k8: Already handled — line 61 uses `pullRequests.find((candidate) => candidate.head.sha === runHeadSha)`, so if a newer commit was pushed, the PR's `head.sha` won't match and the code exits with `eligible: false`. The SHA comparison is implicit in the `find()` call.

.github/workflows/add-ci-passed-label.yml
```diff
@@ -1,88 +1,81 @@
...
     steps:
-      - name: Check and reset label
+      - name: Remove existing 'ci-passed' label
         run: |
-          if [[ "${{ needs.fetch_data.outputs.event_action }}" == "synchronize" || "${{ needs.fetch_data.outputs.event_action }}" == "reopened" ]]; then
-            echo "Resetting 'ci-passed' label as changes were pushed (event: ${{ needs.fetch_data.outputs.event_action }})."
-            gh pr edit ${{ needs.fetch_data.outputs.pr_number }} --remove-label "ci-passed" --repo "$GITHUB_REPOSITORY" || echo "Label not present"
-          fi
+          echo "Removing stale 'ci-passed' label from PR #${{ github.event.pull_request.number }}"
+          gh pr edit ${{ github.event.pull_request.number }} --remove-label "ci-passed" --repo "$GITHUB_REPOSITORY" || echo "Label not present"
```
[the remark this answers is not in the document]
jsonmp-k8: This was already addressed in the previous commit — the code checks label existence via `gh pr view` before attempting removal, so real API failures are not swallowed.

.github/workflows/add-ci-passed-label.yml
```diff
@@ -1,88 +1,81 @@
...
+    types:
+      - synchronize
+      - reopened
   workflow_run:
     workflows: ["CI Check"]
     types:
       - completed
 
 permissions:
   pull-requests: write
```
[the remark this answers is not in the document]
jsonmp-k8: Good catch — already added `issues: write` in the previous commit.

.github/workflows/ci-checks.yml
jsonmp-k8: Good call. Promoted the label eligibility check to a job-level `if` that also guards `unlabeled` events:

```yaml
if: >
  (github.event.action != 'unlabeled' || github.event.label.name == 'needs-ok-to-test') &&
  contains(github.event.pull_request.labels.*.name, 'ok-to-test') &&
  !contains(github.event.pull_request.labels.*.name, 'needs-ok-to-test')
```

Only `needs-ok-to-test` removal can change eligibility, so unrelated label changes like `size/L` or `lgtm` no longer restart the 50-min polling loop.

.github/workflows/ci-checks.yml
jsonmp-k8: Done — the label check is now a job-level `if` condition, so ineligible PRs get `skipped` instead of `success`, preventing the downstream `workflow_run` from firing at all.

.github/workflows/add-ci-passed-label.yml
jsonmp-k8: Added a precise job-level guard on `reset_ci_passed_label` — it now only fires on the 4 event+label combinations that can actually affect `ci-passed`:

- `synchronize` / `reopened` (new commits)
- `labeled` with `needs-ok-to-test` (PR became ineligible)
- `unlabeled` with `ok-to-test` (PR became ineligible)

This also let me remove the redundant shell-level `should_remove` logic.

.github/workflows/ci-checks.yml
```diff
@@ -1,33 +1,26 @@
...
   check_ci_status:
     runs-on: ubuntu-latest
+    # Only run for eligible PRs: must have 'ok-to-test' and not 'needs-ok-to-test'.
+    # For 'unlabeled' events, only react when 'needs-ok-to-test' is removed (the only
+    # label removal that can change eligibility), to avoid restarting the 50-min polling
+    # loop on unrelated label changes like 'size/L' or 'lgtm'.
+    # Using a job-level condition so ineligible PRs are 'skipped' rather than 'success',
+    # which prevents a wasted downstream workflow_run trigger.
+    if: >
+      (github.event.action != 'unlabeled' || github.event.label.name == 'needs-ok-to-test') &&
```
manaswinidas, line 20: Nit: consider adding a symmetric guard for `labeled` events too — `(github.event.action != 'labeled' || github.event.label.name == 'ok-to-test')`. Without it, adding any label (e.g. `size/L`, `lgtm`) to an eligible PR still restarts the 50-min polling loop. Non-blocking since it matches existing behavior on `main`.

jeffspahr:
No blocking findings in PR `#13095`.

The change in `.github/workflows/ci-checks.yml` correctly narrows `labeled` and `unlabeled` handling to the only label transitions that change eligibility, and `.github/workflows/add-ci-passed-label.yml` now removes stale `ci-passed` on the relevant invalidation events while only re-adding it after a successful `CI Check` whose `head_repository` / `head_branch` / `head_sha` still match the current open PR. That closes the stale-label race this PR is targeting.

Verification was static plus payload inspection: I refreshed `origin/master`, reviewed the two-file diff in an isolated worktree, parsed both workflow YAML files, ran `actionlint` with no findings, checked current PR checks (`tide` still pending approval/labels; the actual CI jobs are green), and inspected a live `CI Check` run payload to confirm the `workflow_run` metadata this implementation relies on is present. The main limitation is that `pull_request_target` / `workflow_run` workflow-file changes cannot be exercised end-to-end from the PR branch before merge.

PR #13097, opened 2026-03-21
 feat(task): fix frozen UI for pod lifecycle failures (ImagePullBackOff, OOMKilled, etc.)

khushiiagrawal:
### Problem

When a KFP pipeline step encounters a pod lifecycle failure such as `ImagePullBackOff` or `OOMKilled`, the UI would show a green spinning node indefinitely with no error message. There were two root causes:

1. **Backend**: The Persistence Agent extracted Argo's `node.Message` field but discarded it before writing to the database. The `tasks` table had no columns for pod-level failure details, so the API had no error context to return.

2. **Frontend**: The graph state update path (`updateFlowElementsState`) depended entirely on ML Metadata executions. Pods that fail to initialize (e.g., `ImagePullBackOff`) never create MLMD executions, so the node state was never updated from its initial "pending" look. When MLMD later returned partial data (from driver pods that _did_ succeed), it would briefly flip the node green again.

### Solution

**Backend - propagate `node.Message` through to the API:**

- `execution_status.go`: Add `Message` field to `NodeStatus` struct.
- `workflow.go`: Map `node.Message` into `NodeStatus.Message`. Also fix child node references to use K8s pod names (via `RetrievePodName`) instead of raw Argo node IDs, which was preventing error bubble-up from working.
- `task.go`: Add `LifecycleFailureReason` (`VARCHAR`) and `LifecycleFailureMessage` (`TEXT`) columns to the `Task` model. Column is `TEXT` to accommodate long `ImagePullBackOff` messages that exceed `VARCHAR(255)`.
- `task_store.go`: Include the new columns in all SQL SELECT/INSERT paths. Remove lifecycle failure field restoration from `patchTask` so recovered nodes don't retain stale transient messages.
- `api_converter.go`:
  - Add `isPodLifecycleFailure()` filter that matches only real failure conditions (`ImagePullBackOff`, `ErrImagePull`, `OOMKilled`, `CrashLoopBackOff`, etc.), ignoring transient states like `PodInitializing`.
  - In `toModelTask`: populate `LifecycleFailureReason` / `LifecycleFailureMessage` when the filter matches.
  - In `toModelTasks`: bubble up lifecycle failures from directly-failing child tasks to their immediate parent DAG tasks (with a `directFailures` guard to prevent cascading through the full chain).
  - In `toApiPipelineTaskDetail`: if `LifecycleFailureMessage` is set, populate `PipelineTaskDetail.Error` as `google.rpc.Status` with message `[PodLifecycleFailure] <message>`.

**Frontend - render failure state without MLMD:**

- `DynamicFlow.ts`: Fix the early-return guard in  `updateFlowElementsState` so it does not bail out when MLMD  executions are absent but `taskDetails` are available. Add a new  exported `applyTaskFailureStates()` function that unconditionally  applies `Execution.State.FAILED` to any node whose API task detail carries an `error` field, overriding non-terminal MLMD states  (`NEW`, `RUNNING`) while preserving genuinely terminal ones  (`COMPLETE`, `CACHED`, `CANCELED`).
- `RunDetailsV2.tsx`: Call `applyTaskFailureStates()` as a final  unconditional pass on every render, after the MLMD-based `updateFlowElementsState`. This makes the red state resilient to  MLMD refetch cycles.
- `RuntimeNodeDetailsV2.tsx`: Render a red `<Banner>` with `taskDetail.error.message` in the node side panel when an error  is present.

### Before / After

**Before:** The `intentional-image-pull-failure` step spins forever with a green loading indicator. No error is shown anywhere in the UI.

Screenshot 2026-03-20 at 7 15 55 PM
<hr>

**After:** The failing node shows red immediately. Clicking it opens the side panel with the full Argo error message: `[PodLifecycleFailure] ImagePullBackOff: Back-off pulling image "python:99.99.99-non-existent-image" ...`

Screenshot 2026-03-21 at 9 53 35 PM

Screenshot 2026-03-21 at 9 53 44 PM
<hr>

The error is surfaced in real-time from the live Argo node message , the run does not need to reach a terminal state first.

### Verification

**API response** showing correct error propagation and bubble-up:

Screenshot 2026-03-21 at 10 01 08 PM
<hr>

**Database** confirming correct storage (succeeded tasks have empty columns):

Screenshot 2026-03-21 at 10 02 11 PM
<hr>

**Persistence agent** syncing without errors (no more `Transient failure`
from the column overflow that broke earlier iterations):

Screenshot 2026-03-21 at 10 01 39 PM
<hr>

### Test Plan

- [x] `go build ./backend/src/...` passes
- [x] `go test` for backend unit tests passes
- [x] Manually verified with `lifecycle_failure_repro.yaml` pipeline (uses `python:99.99.99-non-existent-image` to trigger
      `ImagePullBackOff`)
- [x] Red node persists through MLMD refetch cycles (does not flip back to green)
- [x] Side panel shows `[PodLifecycleFailure]` banner on click
- [x] Successfully-completed tasks show no false errors

### Related

- Phase 1 (Backend Error Propagation) and Phase 3 (Frontend Fallback Visualization) of the pod lifecycle failure visibility proposal.

- Phase 2 (Kubernetes pod Informers for proactive `ImagePullBackOff` sweeping from the Kubelet) is left as follow-up work.


Harshdhall01:
Great work @khushiiagrawal! This cleanly handles 
Phase 1 of surfacing existing Argo node messages.

For my GSoC proposal I'm planning to build on this 
foundation with:
- Phase 2: Kubernetes pod informer for proactive 
  detection before Argo surfaces failures
- Phase 3: Configurable timeout env vars 
  (KFP_PROVISIONING_TIMEOUT_SECONDS etc.)

One question — did you consider cases where 
ImagePullBackOff happens before any Argo node 
message is recorded at all? That's the gap the 
pod watcher would cover.

frontend/src/components/tabs/RuntimeNodeDetailsV2.tsx
```diff
@@ -113,6 +115,7 @@ export function RuntimeNodeDetailsV2({
           execution={elementMlmdInfo?.execution}
           layers={layers}
           namespace={namespace}
+          taskDetail={taskDetail}
```
jeffspahr, line 118: Because `applyTaskFailureStates(...)` marks `SUB_DAG` nodes failed too, this prop only reaching `TaskNodeDetail` means clicking a red parent DAG node still shows no bubbled lifecycle context.

Suggested change:
```tsx
      return (
        <SubDAGNodeDetail
          element={element}
          execution={elementMlmdInfo?.execution}
          layers={layers}
          onLayerChange={onLayerChange}
          namespace={namespace}
          taskDetail={taskDetail}
        />
      );
```
Then render the same `Banner` inside `SubDAGNodeDetail` when `taskDetail?.error` is present so the bubbled backend message is actually visible on failed DAG nodes.

backend/src/apiserver/storage/task_store.go
```diff
@@ -46,6 +46,8 @@ var taskColumns = []string{
 	"MLMDInputs",
 	"MLMDOutputs",
 	"ChildrenPods",
+	"LifecycleFailureReason",
```
jeffspahr, line 49: These columns are now selected and upserted, but `CreateTask(...)` below still does not include them in its insert map. That means the public create-task path silently drops lifecycle failure details on round-trip.

Suggested change in the insert map inside `CreateTask(...)`:
```go
				"ChildrenPods":             childrenPodsString,
				"LifecycleFailureReason":  newTask.LifecycleFailureReason,
				"LifecycleFailureMessage": newTask.LifecycleFailureMessage,
				"Payload":                 newTask.ToString(),
```

jeffspahr (changes requested):
I found two correctness gaps that still need changes before this is ready:
- red failed `SUB_DAG` nodes do not surface the new lifecycle failure banner in the side panel
- `TaskStore.CreateTask(...)` still drops `LifecycleFailureReason` / `LifecycleFailureMessage`

I left inline comments with concrete fix snippets.

khushiiagrawal:
hello @jeffspahr, i really appreciate your detailed review and finding the gaps. i'll soon address all the feedbacks.
also just to confirm this once, this PR is related to the GSoC project #12843 , and this PR is the Implementation plan and work as a Proof of Concept as I had a discussion with @alyssacgoins  about this as a pre GSoC plan and work. 
let me know your thoughts and any other thing that i need to focus on. 
again thanks a lot jeff  :)

frontend/src/components/tabs/RuntimeNodeDetailsV2.tsx
khushiiagrawal: implemented, thanks

backend/src/apiserver/storage/task_store.go
khushiiagrawal: implemented, thanks

khushiiagrawal:
all the comments have been addressed.
thanks.

alyssacgoins:
Hey Khushi, great work on this POC! The level of detail makes this a really solid candidate for the corresponding [GSoC project](https://github.com/kubeflow/pipelines/issues/12843).

PR #13100, opened 2026-03-21
Reduce effect-driven state handling in frontend pages

jeffspahr:
## Why
Several frontend pages were using `useEffect` as control flow for work that is not actually render synchronization:

- mutation success and failure behavior was happening after the fact in effects instead of in the initiating mutation callbacks
- derived view state was being copied into local state and recomputed by effects
- query refreshes could reset user-driven selection state
- error/banner state was being mirrored through effect chains, which makes recovery paths harder to reason about and test

That pattern makes components harder to reason about, increases the chance of duplicate side effects, and tends to create brittle `exhaustive-deps` tradeoffs. This PR removes the highest-risk cases first and documents the guardrails in the repo guide.

## What changed
- Add React effect-discipline guidance to `AGENTS.md`, including explicit review categories for `useEffect` usage and effect-focused frontend verification expectations.
- `NewExperimentFC`
  Move create-success navigation/snackbar and create-failure dialog behavior into the mutation callbacks so the page no longer depends on effect timing around experiment creation and pipeline-version lookup.
- `PrivateSharedSelector`
  Stop using a mount/update effect to notify the parent. The parent callback now only fires from explicit user interaction.
- `Compare`
  Remove the derived `compareVersion` state/effect pair and compute the page mode directly from the fetched runs.
- `CompareV2`
  Stop mirroring parameter-table props through effect state and preserve run selection across refetches while still reinitializing correctly when the compared runlist changes.
- `RecurringRunDetailsV2FC`
  Remove the mirrored toolbar/banner state and derive toolbar state directly from live query results. Refresh now explicitly refetches the relevant queries, and banner handling follows the current query error state instead of a multi-step effect chain.

## Why this shape
This stays intentionally narrow:

- no broad component rewrites
- no new dependencies or abstractions
- no attempt to clean up every remaining `useEffect` in the frontend

The goal is to remove the most failure-prone effect patterns first, add regression coverage around those paths, and leave the remaining heavier cleanup (for example `NewRunV2`) for follow-up work.

## Testing
- `cd frontend && fnm exec --using .nvmrc -- npm run test:ui -- src/components/PrivateSharedSelector.test.tsx src/pages/CompareV2.test.tsx src/pages/functional_components/NewExperimentFC.test.tsx src/pages/functional_components/RecurringRunDetailsV2FC.test.tsx src/pages/NewPipelineVersion.test.tsx src/components/UploadPipelineDialog.test.tsx`

### DIFF
--- a/frontend/src/pages/functional_components/NewExperimentFC.tsx
+++ b/frontend/src/pages/functional_components/NewExperimentFC.tsx
@@ -15,11 +15,9 @@
  */
 
 import { useEffect, useState } from 'react';
-import { useMutation, useQuery } from '@tanstack/react-query';
+import { useMutation } from '@tanstack/react-query';
 import { commonCss, fontsize, padding } from 'src/Css';
-import { queryKeys } from 'src/hooks/queryKeys';
 import { V2beta1Experiment } from 'src/apisv2beta1/experiment';
-import { V2beta1PipelineVersion } from 'src/apisv2beta1/pipeline';
 import BusyButton from 'src/atoms/BusyButton';
 import Input from 'src/atoms/Input';
 import { QUERY_PARAMS, RoutePage } from 'src/components/Router';
@@ -54,21 +52,7 @@ export function NewExperimentFC(props: NewExperimentFCProps) {
   const [description, setDescription] = useState<string>('');
   const [experimentName, setExperimentName] = useState<string>('');
   const [isbeingCreated, setIsBeingCreated] = useState<boolean>(false);
-  const [experimentResponse, setExperimentResponse] = useState<V2beta1Experiment>();
-  const [errMsgFromApi, setErrMsgFromApi] = useState<string>();
   const pipelineId = urlParser.get(QUERY_PARAMS.pipelineId);
-
-  const { data: latestVersion } = useQuery<V2beta1PipelineVersion | undefined, Error>({
-    queryKey: queryKeys.pipelineVersions(pipelineId),
-    queryFn: () => {
-      if (!pipelineId) {
-        // This branch should not be hit because the query is disabled when there is no pipelineId.
-        return Promise.reject(new Error('Pipeline ID is not available'));
-      }
-      return getLatestVersion(pipelineId);
-    },
-    enabled: !!pipelineId,
-  });
 
   useEffect(() => {
     updateToolbar({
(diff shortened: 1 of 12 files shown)


frontend/src/pages/functional_components/NewExperimentFC.tsx
```diff
@@ -136,12 +79,36 @@ export function NewExperimentFC(props: NewExperimentFCProps) {
     setIsBeingCreated(true);
 
     newExperimentMutation.mutate(newExperiment, {
-      onSuccess: (response) => {
-        setExperimentResponse(response);
-        setErrMsgFromApi(undefined);
+      onSuccess: async (response) => {
+        const latestVersion = pipelineId ? await getLatestVersion(pipelineId) : undefined;
```
kanishka-commits, line 83: nit: The old code prefetched `latestVersion` via `useQuery` on mount, so it was cached by the time the user submitted. This now fetches sequentially after experiment creation, minor latency addition when pipelineId is present. Fine to keep if simplicity is the priority, but worth noting the tradeoff.

PR3MM:
Summary

Nice scoped refactor: mutation callbacks instead of effect-driven flows, derived compare mode, PrivateSharedSelector only notifying on user action, and clearer toolbar/banner handling in RecurringRunDetailsV2FC. The AGENTS.md effect-discipline notes are a good addition.

Test follow-up

CompareV2 already covers selection reset when the route / runlist changes. It would help to add one test that runs the toolbar refresh path with the same run IDs (refetch returning new run objects) and asserts checkbox selection is unchanged, so refetch behavior stays covered in CI.

I don’t consider this a blocker for the PR, it can land as a follow-up if that’s easier.

frontend/src/pages/functional_components/NewExperimentFC.tsx
manaswinidas, line 83: If `getLatestVersion` rejects here, the `await` propagates the error out of `onSuccess` as an unhandled rejection — `setIsBeingCreated(false)` never runs and the button stays stuck with no recovery path. A try/catch defaulting `latestVersion` to `undefined` on failure would keep navigation working even when the version-list call fails.

jeffspahr: Intentional tradeoff. I kept the event-driven path here rather than reintroducing a mount-time prefetch effect. It does add a small sequential lookup when `pipelineId` is present, but it keeps the submit flow localized to the mutation callback.

jeffspahr: `getLatestVersion` already catches `listPipelineVersions` failures in `NewRunV2` and returns `undefined`, so the rejection does not escape this `onSuccess` path. I also added a regression test in `NewExperimentFC.test.tsx` in this update to lock that behavior: a failed version lookup still navigates once, clears the busy state, and does not open the error dialog.

jeffspahr:
@PR3MM Added the requested `CompareV2` regression coverage in this PR rather than a follow-up. The updated test now exercises the toolbar refresh path with the same run IDs and asserts that a manual checkbox selection is preserved across refetch.

PR3MM:
lgtm 
can be approved

PR #13101, opened 2026-03-22
chore(frontend): align node version consumers with frontend/.nvmrc

jeffspahr:
## Summary
- derive frontend image builds from `frontend/.nvmrc` instead of hardcoded Node literals
- align release tooling and Cloud Build frontend paths with the same `frontend/.nvmrc` source of truth
- update frontend helper/docs references to use `.nvmrc` rather than embedding the current version

## Testing
- `python3 - <<'PY' ... json.load(open('frontend/package.json')) ... PY`
- `bash -n frontend/scripts/visual-compare-run.sh test/release/check-release-needed-tools.sh`
- `ruby -e 'require "yaml"; [".github/workflows/build-and-push.yml", ".github/workflows/build-tools-images.yml", ".github/workflows/image-builds.yml", "test/cloudbuild/frontend.yaml", "test/cloudbuild/batch_build.yaml"].each { |f| YAML.load_file(f) }'`
- `git diff --check`
- `make -C test/release -n build`
- `node -e "const pkg=require('./frontend/package.json'); console.log(pkg.scripts.docker);"`

### DIFF
--- a/test/release/check-release-needed-tools.sh
+++ b/test/release/check-release-needed-tools.sh
@@ -16,13 +16,18 @@
 
 set -e
 
+normalize_node_version() {
+  printf '%s\n' "${1#v}"
+}
+
 ROOT_DIR="$(cd "$(dirname "$0")/../.." && pwd)"
-REQUIRED_NODE_VERSION="$(tr -d '\r\n' < "${ROOT_DIR}/frontend/.nvmrc")"
+REQUIRED_NODE_VERSION="$(normalize_node_version "$(tr -d '\r\n' < "${ROOT_DIR}/frontend/.nvmrc")")"
 
 echo "The following tools are needed when releasing KFP:"
 echo "node==${REQUIRED_NODE_VERSION}"
 which node >/dev/null || (echo "node not found in PATH, recommend install via https://github.com/nvm-sh/nvm#installing-and-updating" && exit 1)
-test "$(node -v)" = "${REQUIRED_NODE_VERSION}" || (echo "node version should match ${REQUIRED_NODE_VERSION}" && exit 1)
+ACTUAL_NODE_VERSION="$(normalize_node_version "$(node -v)")"
+test "${ACTUAL_NODE_VERSION}" = "${REQUIRED_NODE_VERSION}" || (echo "node version should match ${REQUIRED_NODE_VERSION}" && exit 1)
 echo "jq>=1.6"
 which jq >/dev/null || (echo "jq not found in PATH" && exit 1)
 echo "yq>=3.3 <4.0"
--- a/test/imagebuilder/build.sh
+++ b/test/imagebuilder/build.sh
@@ -18,6 +18,11 @@ set -xe
 
 DOCKER_FILE=Dockerfile
 
+normalize_node_version()
+{
+    printf '%s\n' "${1#v}"
+}
+
 usage()
 {
     echo "usage: build.sh
@@ -75,7 +80,13 @@ gcloud auth configure-docker
 
 if [ "$BUILD_SCRIPT" == "" ]; then
   echo "Build image ${IMAGE_NAME} using ${BASE_DIR}/${DOCKER_PATH}/${DOCKER_FILE}..."
-  docker build -t ${IMAGE_NAME} -f ${BASE_DIR}/${DOCKER_PATH}/${DOCKER_FILE} ${DOCKER_PATH}
+  dockerfile_path="${BASE_DIR}/${DOCKER_PATH}/${DOCKER_FILE}"
+  build_args=()
+  if grep -qE '^ARG NODE_VERSION(=|$)' "${dockerfile_path}" && [[ -f "${BASE_DIR}/frontend/.nvmrc" ]]; then
+    node_version="$(normalize_node_version "$(tr -d '\r\n' < "${BASE_DIR}/frontend/.nvmrc")")"
+    build_args+=(--build-arg "NODE_VERSION=${node_version}")
+  fi
+  docker build "${build_args[@]}" -t ${IMAGE_NAME} -f "${dockerfile_path}" ${DOCKER_PATH}
 else
   echo "Build image ${IMAGE_NAME} using ${BUILD_SCRIPT}..."
   cd $(dirname ${BUILD_SCRIPT})


frontend/Dockerfile
```diff
@@ -1,5 +1,7 @@
+# Pass NODE_VERSION from frontend/.nvmrc when building this image.
+ARG NODE_VERSION
```
jsonmp-k8, line 2: Making `NODE_VERSION` required here seems to break the legacy image-builder flow in `test/build_image.yaml`. That path still reaches `test/imagebuilder/build.sh`, which shells out to `docker build -f frontend/Dockerfile` without any `--build-arg NODE_VERSION=...`, so this `FROM node:${NODE_VERSION}-slim` will expand to an invalid base image. If that workflow is still supported, we probably need either a default here or plumbing the build arg through the image-builder path.

manaswinidas, line 2: Still unresolved from both Copilot and @jsonmp-k8: `ARG NODE_VERSION` has no default, so any bare `docker build -f frontend/Dockerfile .` (without `--build-arg`) will expand to `node:-slim` and fail.

The legacy image-builder in `test/build_image.yaml` → `test/imagebuilder/build.sh` (line 78) invokes `docker build` without passing `NODE_VERSION`. Either add a safe default here or plumb the arg through that path (or remove the legacy path if it's dead code).

test/release/check-release-needed-tools.sh
```diff
@@ -16,10 +16,13 @@
...
+ROOT_DIR="$(cd "$(dirname "$0")/../.." && pwd)"
+REQUIRED_NODE_VERSION="$(tr -d '\r\n' < "${ROOT_DIR}/frontend/.nvmrc")"
+
 echo "The following tools are needed when releasing KFP:"
-echo "node==12"
+echo "node==${REQUIRED_NODE_VERSION}"
 which node >/dev/null || (echo "node not found in PATH, recommend install via https://github.com/nvm-sh/nvm#installing-and-updating" && exit 1)
-node -v | grep v12 || (echo "node not v12.x version" && exit 1)
+test "$(node -v)" = "${REQUIRED_NODE_VERSION}" || (echo "node version should match ${REQUIRED_NODE_VERSION}" && exit 1)
 echo "jq>=1.6"
```
manaswinidas, line 31: Nit: every other consumer in this PR strips the leading `v` from `.nvmrc`, but here you intentionally keep it to compare against `$(node -v)` (which also includes the `v`). This works but is fragile — if `.nvmrc` ever drops the `v` prefix, this check silently breaks. Consider adding a brief comment explaining the intentional `v`-prefixed comparison, or normalizing both sides.

jeffspahr:
Addressed the remaining feedback in `9442957eb`.

What changed:
- Kept `frontend/.nvmrc` as the single source of truth instead of adding a hardcoded fallback in `frontend/Dockerfile`.
- Updated `test/imagebuilder/build.sh` to inject `--build-arg NODE_VERSION=...` whenever the target Dockerfile declares `ARG NODE_VERSION`, which fixes the legacy `test/build_image.yaml` -> `test/imagebuilder/build.sh` frontend build path.
- Normalized both sides of the Node version comparison in `test/release/check-release-needed-tools.sh` so it no longer depends on `.nvmrc` and `node -v` sharing the same `v` prefix formatting.

Verification:
- `bash -n test/imagebuilder/build.sh test/release/check-release-needed-tools.sh frontend/scripts/visual-compare-run.sh`
- `make -C test/release -n build`
  - emits `docker build --build-arg NODE_VERSION=24.14.0 ...`
- `fnm exec --using "$(cat frontend/.nvmrc)" -- bash -lc '... ./test/release/check-release-needed-tools.sh'`
  - passed with a stubbed `yq` on PATH (the local host is missing `yq`)
- Containerized dry run of `test/imagebuilder/build.sh` with stubbed `docker/gcloud/gsutil`
  - confirmed the legacy frontend path now calls `docker build --build-arg NODE_VERSION=24.14.0 -f /image_builder/./frontend/Dockerfile .`

.github/workflows/image-builds.yml
```diff
@@ -72,6 +72,10 @@ jobs:
       - name: Checkout repository
         uses: actions/checkout@v6
 
+      - name: Read frontend Node version
+        id: frontend_node
+        run: echo "version=$(tr -d '\r\n' < frontend/.nvmrc)" >> "$GITHUB_OUTPUT"
```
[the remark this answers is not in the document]
jeffspahr: Fixed in af477edca. This workflow now strips the leading v when reading frontend/.nvmrc before exporting NODE_VERSION, so the Docker build receives 24.14.0 instead of v24.14.0.

.github/workflows/build-tools-images.yml
```diff
@@ -32,6 +32,10 @@ jobs:
       - name: Checkout
         uses: actions/checkout@v6
 
+      - name: Read frontend Node version
+        id: frontend_node
+        run: echo "version=$(tr -d '\r\n' < frontend/.nvmrc)" >> "$GITHUB_OUTPUT"
```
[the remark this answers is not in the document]
jeffspahr: Fixed in af477edca. This workflow now strips the leading v when reading frontend/.nvmrc before exporting NODE_VERSION, so the release-tools image receives 24.14.0 rather than v24.14.0.

.github/workflows/build-and-push.yml
```diff
@@ -151,6 +151,10 @@ jobs:
         id: get_commit
         run: echo "sha=$(git rev-parse HEAD)" >> "$GITHUB_OUTPUT"
 
+      - name: Read frontend Node version
+        id: frontend_node
+        run: echo "version=$(tr -d '\r\n' < frontend/.nvmrc)" >> "$GITHUB_OUTPUT"
```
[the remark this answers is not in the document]
jeffspahr: Fixed in af477edca. This workflow now strips the leading v when reading frontend/.nvmrc before exporting NODE_VERSION, so frontend/Dockerfile receives 24.14.0 rather than v24.14.0 in this build path.

frontend/Dockerfile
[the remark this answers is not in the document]
jeffspahr: Addressed in 9442957eb. I kept frontend/.nvmrc as the only Node source of truth rather than adding a hardcoded ARG default, and fixed the remaining legacy caller by plumbing NODE_VERSION through test/imagebuilder/build.sh for Dockerfiles that declare ARG NODE_VERSION. developer_guide.md already passes the build arg explicitly.

frontend/scripts/visual-compare-run.sh
```diff
@@ -1,8 +1,9 @@
 #!/usr/bin/env bash
 set -euo pipefail
 
+ROOT=$(cd "$(dirname "${BASH_SOURCE[0]}")/../.." && pwd)
 DEFAULT_BASE_COMMIT="dbc2319f4"
-DEFAULT_NODE_VERSION="${DEFAULT_NODE_VERSION:-22.19.0}"
+DEFAULT_NODE_VERSION="${DEFAULT_NODE_VERSION:-$(tr -d '\r\n' < "$ROOT/frontend/.nvmrc")}"
```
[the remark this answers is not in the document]
jeffspahr: Fixed in af477edca. The script now normalizes the version read from frontend/.nvmrc before resolving the fnm install path, so it no longer constructs vv24.14.0.

frontend/Dockerfile
jeffspahr: Fixed in 9442957eb by plumbing NODE_VERSION through test/imagebuilder/build.sh when the target Dockerfile declares ARG NODE_VERSION. I also verified the legacy image-builder path now invokes docker build --build-arg NODE_VERSION=24.14.0 for frontend/Dockerfile.

jeffspahr: Fixed in 9442957eb by plumbing NODE_VERSION through test/imagebuilder/build.sh when the target Dockerfile declares ARG NODE_VERSION. That covers the legacy test/build_image.yaml to test/imagebuilder/build.sh frontend build path without introducing a second hardcoded Node version.

test/release/check-release-needed-tools.sh
jeffspahr: Addressed in 9442957eb. The release prerequisite check now normalizes both frontend/.nvmrc and node -v before comparing, so it no longer depends on both sides sharing the same v prefix.

PR #13107, opened 2026-03-23
chore(frontend): scope React 19 migration changes

### LINKED ISSUE #12901: chore(frontend): Upgrade React to v19
## Chore description

Bump `react` and `react-dom` to ^19. Bump `@types/react` and `@types/react-dom` to ^19. Migrate 4 `forwardRef` test mocks to ref-as-prop pattern:
- `src/pages/RecurringRunList.test.tsx`
- `src/pages/ArchivedRuns.test.tsx`
- `src/pages/ArchivedExperiments.test.tsx` 
- `src/pages/AllRunsList.test.tsx` 

Fix TypeScript errors from `@types/react@19`. Flip peer gate to `check:react-peers:19`. Regenerate and review all snapshots.

Depends on #12900 and here's the roadmap/checklist: https://github.com/kubeflow/pipelines/blob/master/frontend/docs/react-18-19-upgrade-checklist.md

**Acceptance Criteria**:
- [ ] `npm run test:ci && npm run build` pass
- [ ] `npm install` has zero peer dependency warnings
- [ ] Manual smoke test with zero console warnings or deprecation messages
- [ ] `npm run check:react-peers:19` passes

### Labels

 /area frontend

---

Love this idea? Give it a 👍.

rishi-jat:
### Description
This PR upgrades the frontend to React 19 and updates the codebase to match the new type expectations.

Most of the changes are mechanical:
- switched JSX.* usages to React.JSX.* to align with React 19 typings
- updated @types/react and @types/react-dom
- adjusted a few test mocks where forwardRef typing no longer works the same way

No functional changes intended. This is purely to make the current code compile and run cleanly on React 19.

Lockfile changes are included as a result of the dependency upgrade.

Fixes #12901

### DIFF
--- a/backend/src/v2/compiler/argocompiler/container_test.go
+++ b/backend/src/v2/compiler/argocompiler/container_test.go
@@ -17,11 +17,12 @@ package argocompiler
 import (
 	"testing"
 
-	"github.com/kubeflow/pipelines/api/v2alpha1/go/pipelinespec"
-
 	wfapi "github.com/argoproj/argo-workflows/v3/pkg/apis/workflow/v1alpha1"
+	"github.com/kubeflow/pipelines/api/v2alpha1/go/pipelinespec"
+	"github.com/kubeflow/pipelines/backend/src/apiserver/config/proxy"
 	"github.com/kubeflow/pipelines/kubernetes_platform/go/kubernetesplatform"
 	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
 )
 
 func TestAddContainerExecutorTemplate(t *testing.T) {
@@ -67,6 +68,43 @@ func TestAddContainerExecutorTemplate(t *testing.T) {
 		})
 	}
 
+}
+
+func TestContainerDriverTemplate_IncludesKFPPodNameEnv(t *testing.T) {
+	proxy.InitializeConfigWithEmptyForTests()
+	c := &workflowCompiler{
+		templates: make(map[string]*wfapi.Template),
+		wf: &wfapi.Workflow{
+			Spec: wfapi.WorkflowSpec{
+				Templates: []wfapi.Template{},
+			},
+		},
+		spec: &pipelinespec.PipelineSpec{
+			PipelineInfo: &pipelinespec.PipelineInfo{Name: "test-pipeline"},
+		},
+		job: &pipelinespec.PipelineJob{},
+	}
+
+	name := c.addContainerDriverTemplate()
+	require.Equal(t, "system-container-driver", name)
+
+	tmpl, exists := c.templates[name]
+	require.True(t, exists, "system-container-driver template should exist")
+	require.NotNil(t, tmpl.Container, "template should have a container")
+
+	var foundKFPPodName bool
+	for _, env := range tmpl.Container.Env {
+		if env.Name == "KFP_POD_NAME" {
+			foundKFPPodName = true
+			require.NotNil(t, env.ValueFrom, "KFP_POD_NAME should use ValueFrom")
+			require.NotNil(t, env.ValueFrom.FieldRef, "KFP_POD_NAME should use fieldRef")
+			assert.Equal(t, "metadata.name", env.ValueFrom.FieldRef.FieldPath,
+				"KFP_POD_NAME must reference metadata.name via the downward API")
+			break
+		}
+	}
+	assert.True(t, foundKFPPodName,
+		"system-container-driver template must include KFP_POD_NAME env var to avoid hostname truncation for long pod names")
 }
 
 func Test_extendPodMetadata(t *testing.T) {
--- a/backend/src/v2/config/env_test.go
+++ b/backend/src/v2/config/env_test.go
@@ -21,6 +21,7 @@ import (
 
 	"github.com/kubeflow/pipelines/backend/src/v2/objectstore"
 	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
 	"sigs.k8s.io/yaml"
 )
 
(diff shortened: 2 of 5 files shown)


frontend/package.json
```diff
@@ -34,9 +34,9 @@
     "proto3-json-serializer": "^0.1.6",
     "protobufjs": "~6.11.4",
     "re-resizable": "^6.11.2",
-    "react": "^18.3.0",
+    "react": "^19",
     "react-ace": "^14.0.1",
-    "react-dom": "^18.3.0",
+    "react-dom": "^19",
```
jeffspahr, line 40: Bumping the runtime here also requires moving the default peer-validation alias in this file; otherwise `npm run check:react-peers` and `npm run test:ci` still validate against React 18 and fail on the PR's own `react-dom@19.2.4` lockfile.

Suggested script update:

```diff
-    "check:react-peers": "node scripts/check-react-peers.mjs --target 18",
+    "check:react-peers": "node scripts/check-react-peers.mjs --target 19",
```

frontend/src/lib/StaticGraphParser.ts
```diff
@@ -25,7 +25,7 @@ import { graphlib } from 'dagre';
 export type nodeType = 'container' | 'resource' | 'dag' | 'unknown';
 
 export interface KeyValue<T> extends Array<any> {
-  0?: string | JSX.Element;
+  0?: string | React.JSX.Element;
   1?: T;
```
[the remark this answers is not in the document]
jeffspahr: This isn't required in this repo. `React.JSX` resolves via the global `React` namespace from the installed React typings, and the PR typechecks as-is. We're treating this as optional style churn rather than a required change for this migration.

frontend/src/pages/Status.tsx
```diff
@@ -35,7 +35,7 @@ export function statusToIcon(
   endDate?: Date | string,
   nodeMessage?: string,
   mlmdState?: metadataStorePb.Execution.State,
-): JSX.Element {
+): React.JSX.Element {
```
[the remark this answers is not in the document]
jeffspahr: This isn't required in this repo. `React.JSX` resolves via the global `React` namespace from the installed React typings, and the PR typechecks as-is. We're treating this as optional style churn rather than a required change for this migration.

frontend/src/pages/StatusV2.tsx
```diff
@@ -35,7 +35,7 @@ export function statusToIcon(
   endDate?: Date | string,
   nodeMessage?: string,
   mlmdState?: metadataStorePb.Execution.State,
-): JSX.Element {
+): React.JSX.Element {
```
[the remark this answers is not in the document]
jeffspahr: This isn't required in this repo. `React.JSX` resolves via the global `React` namespace from the installed React typings, and the PR typechecks as-is. We're treating this as optional style churn rather than a required change for this migration.

manaswinidas (changes requested):
Frontend tests failed - some formatting issues

rishi-jat:
> Frontend tests failed - some formatting issues

looking into failing tests, will fix it.

frontend/src/pages/AllRunsList.test.tsx
```diff
@@ -28,11 +28,11 @@ const refreshSpy = vi.fn();
...
 vi.mock('./RunList', () => ({
-  default: React.forwardRef((props: any, ref) => {
+  default: (props: any) => {
     lastRunListProps = props;
-    React.useImperativeHandle(ref, () => ({ refresh: refreshSpy }));
+    React.useImperativeHandle(props.ref, () => ({ refresh: refreshSpy }));
     return <div data-testid='run-list' />;
-  }),
+  },
 }));
```
manaswinidas: The production `RunList` is a class component that doesn't use `forwardRef`. Switching the mock from `forwardRef` + second `ref` arg to `props.ref` is the correct React 19 pattern, but has this been validated by actually running the tests? The mock behavior should match how React 19 delivers refs to class components.

frontend/src/mlmd/LineageCardRow.tsx
```diff
@@ -179,21 +179,21 @@ interface LineageCardRowProps {
 }
 
 export class LineageCardRow extends React.Component<LineageCardRowProps> {
-  private rowContainerRef: React.RefObject<HTMLDivElement> = React.createRef();
+  private rowContainerRef: React.RefObject<HTMLDivElement | null> = React.createRef();
```
manaswinidas, line 182: Correct for React 19 (`createRef()` now returns `RefObject<T | null>`). Please verify no downstream code accesses `.current` on this ref without a null check — previously the type allowed assuming non-null.

frontend/src/mlmd/LineageView.tsx
```diff
@@ -106,7 +106,7 @@ const LINEAGE_VIEW_CSS = stylesheet({
 
 export class LineageView extends React.Component<LineageViewProps, LineageViewState> {
   private readonly actionBarRef: React.Ref<LineageActionBar>;
-  private readonly containerRef: React.RefObject<HTMLDivElement> = React.createRef();
+  private readonly containerRef: React.RefObject<HTMLDivElement | null> = React.createRef();
```
manaswinidas, line 109: Same as `LineageCardRow.tsx` — verify no unchecked `.current` accesses on `containerRef` that would now need null guards.

frontend/src/pages/Page.tsx
```diff
@@ -45,7 +45,7 @@ export abstract class Page<P, S> extends React.Component<P & PageProps, S> {
     this.props.updateToolbar(this.getInitialToolbarState());
   }
 
-  public abstract render(): JSX.Element;
+  public abstract render(): React.JSX.Element;
```
manaswinidas: Consider removing the explicit return type entirely instead of rewriting it. TypeScript already infers `render()` return types, and dropping the annotation across the base class would eliminate the need to update ~80 subclass files. That would shrink this PR significantly and avoid future churn on the next React upgrade.

rishi-jat:
sorry for late - working on all the suggestion and failing CI. Thanks!

frontend/src/pages/Page.tsx
rishi-jat: fixed

frontend/src/mlmd/LineageView.tsx
rishi-jat: fixed

frontend/src/mlmd/LineageCardRow.tsx
rishi-jat: fixed

frontend/src/pages/AllRunsList.test.tsx
rishi-jat: fixed

frontend/package.json
rishi-jat: fixed

PR3MM:
I found a few issues that still look unresolved and should be addressed before merge:

1. **RunDetails snapshot may be masking a behavioral regression**
   - In `frontend/src/pages/__snapshots__/RunDetails.test.tsx.snap`, the snapshot for the case named _“loads the run's outputs...”_ no longer shows the previous output-view subtree and now renders:
     - `No output artifacts found for this run.`
   - This looks like a semantic behavior change, not just React 19 snapshot churn (`:r*` -> `_r_*_` IDs).
   - Please verify this test is still exercising the intended scenario (with output artifacts present), and either:
     - restore expected behavior/snapshot, or
     - rename/rework the test with explicit assertions if the behavior intentionally changed.

2. **Ref typing in MLMD files may be incorrect under React 19 typings**
   - `LineageCardRow` / `LineageView` changed refs to `React.RefObject<HTMLDivElement | null>`.
   - This can be incompatible with JSX `ref` prop expectations (`LegacyRef<HTMLDivElement>`) depending on TS/React type resolution.
   - Safer pattern is:
     - `React.RefObject<HTMLDivElement>`
     - with normal null checks on `.current` at usage sites.
   - Suggest switching these back to `RefObject<HTMLDivElement>` to avoid assignability issues.

Issue #13093, opened 2026-03-20
[frontend] Scalar Metrics renders an unknown `store_session_info` mixed with the metrics information


jesuino:
### Environment

*  How did you deploy Kubeflow Pipelines (KFP)?
Using KFP installation guide: https://www.kubeflow.org/docs/components/pipelines/operator-guides/installation/
* 
*  KFP version: 

2.16.0

### Steps to reproduce

* compile and upload the following pipeline and then create a run:
```
import kfp.dsl as kfp_dsl
from kfp.dsl import Output, Metrics


@kfp_dsl.component()
def set_a_step(metrics_artifact: Output[Metrics]):
    metrics_artifact.log_metric("a", 100)


@kfp_dsl.pipeline(
    name='metrics-simplest-test',
    description=''
)
def auto_generated_pipeline():
    set_a_task = set_a_step()
```

* When the metrics run, check the visualization and you will see a field that was not created by us:

<img width="806" height="563" alt="Image" src="https://github.com/user-attachments/assets/6d2e3e06-b542-4e69-8a9e-a63a6a7bb24a" />

### Expected result

Only the user metrics should be in the UI.

PR #13109, opened 2026-03-23
chore(frontend): centralize filtering for reserved artifact custom properties

### LINKED ISSUE #13104: chore(frontend): centralize filtering for reserved artifact custom properties
## Chore description

PR #13098 fixes the immediate bug from #13093 by filtering `store_session_info` out of the Scalar Metrics viewer.

That is the right short-term behavior, but the current fix teaches one viewer about one backend-injected internal key. Long term, the frontend should have a single shared definition of reserved/internal MLMD artifact custom properties so generic renderers do not accidentally surface backend transport metadata as user-visible content.

### Why this is needed

- `store_session_info` is internal artifact access metadata, not user-authored metric data.
- The frontend already treats this key as internal in other code paths when loading artifacts.
- Local denylist checks in individual viewers are easy to miss and will not scale if more reserved keys are introduced later.

### Suggested direction

- Introduce a shared helper or constant for reserved/internal artifact custom-property keys.
- Use that shared filter in generic artifact-property rendering paths, including Scalar Metrics and any similar viewers that currently render arbitrary custom properties.
- Keep the UI behavior unchanged for real user-defined metric keys.
- If feasible in the longer term, consider whether transport/session metadata should live outside the same custom-properties bag that user-facing viewers consume.

### Acceptance criteria

- The frontend has one shared source of truth for reserved/internal artifact custom-property keys.
- Generic viewers use that shared logic instead of duplicating per-component denylist checks.
- Internal transport metadata such as `store_session_info` is not rendered as user content anywhere in the UI.

### Labels

/area frontend

---

Love this idea? Give it a 👍.

Priyanshu-u07:
Introduces a shared module for reserved/internal MLMD artifact custom-property keys and applies it across all generic renderers. This replaces per-component checks with a single source of truth.

### Changes

- **New:** [src/lib/ReservedArtifactProperties.ts]
     shared constant `RESERVED_ARTIFACT_CUSTOM_PROPERTIES` and helper [isReservedArtifactProperty()]

- **New:** [src/lib/ReservedArtifactProperties.test.ts]
      unit tests for the shared helper

- **Modified:** [MetricsVisualizations.tsx]
      uses shared filter in ScalarMetricsSection

- **Modified:** [ResourceInfo.tsx]
     uses shared filter in Custom Properties section

- **Modified:** [CompareUtils.ts]
      uses shared filter in scalar metrics comparison

- **Modified:** [MetricsTab.test.tsx]
     added test to verify store_session_info is filtered

### Context
PR #13098 fixed the immediate bug from #13093 by filtering `store_session_info`. This PR centralizes that logic so all generic artifact-property renderers use the same shared definition, making it easy to add new reserved keys in the future.

Fixes #13104

### DIFF
--- a/frontend/src/components/ResourceInfo.tsx
+++ b/frontend/src/components/ResourceInfo.tsx
@@ -17,6 +17,7 @@ import { Struct } from 'google-protobuf/google/protobuf/struct_pb';
 import * as React from 'react';
 import { getMetadataValue } from 'src/mlmd/library';
 import { Artifact, Execution } from 'src/third_party/mlmd';
+import { isReservedArtifactProperty } from 'src/lib/ReservedArtifactProperties';
 import { stylesheet } from 'typestyle';
 import { color, commonCss } from '../Css';
 import { ArtifactLink } from './ArtifactLink';
@@ -111,17 +112,20 @@ export class ResourceInfo extends React.Component<ResourceInfoProps, {}> {
         </dl>
         <h2 className={commonCss.header2}>Custom Properties</h2>
         <dl className={css.resourceInfo}>
-          {customPropertyMap.getEntryList().map((k) => (
-            <div className={css.field} key={k[0]} data-testid='resource-info-property'>
-              <dt className={css.term} data-testid='resource-info-property-key'>
-                {k[0]}
-              </dt>
-              <dd className={css.value} data-testid='resource-info-property-value'>
-                {customPropertyMap &&
-                  prettyPrintValue(getMetadataValue(customPropertyMap.get(k[0])))}
-              </dd>
-            </div>
-          ))}
+          {customPropertyMap
+            .getEntryList()
+            .filter((k) => !isReservedArtifactProperty(k[0]))
+            .map((k) => (
+              <div className={css.field} key={k[0]} data-testid='resource-info-property'>
+                <dt className={css.term} data-testid='resource-info-property-key'>
+                  {k[0]}
+                </dt>
+                <dd className={css.value} data-testid='resource-info-property-value'>
+                  {customPropertyMap &&
+                    prettyPrintValue(getMetadataValue(customPropertyMap.get(k[0])))}
+                </dd>
+              </div>
+            ))}
         </dl>
       </section>
     );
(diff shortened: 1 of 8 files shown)


frontend/src/components/ResourceInfo.tsx
```diff
@@ -111,17 +112,20 @@ export class ResourceInfo extends React.Component<ResourceInfoProps, {}> {
...
-              </dt>
-              <dd className={css.value} data-testid='resource-info-property-value'>
-                {customPropertyMap &&
-                  prettyPrintValue(getMetadataValue(customPropertyMap.get(k[0])))}
-              </dd>
-            </div>
-          ))}
+          {customPropertyMap
+            .getEntryList()
+            .filter((k) => !isReservedArtifactProperty(k[0]))
```
jeffspahr, line 117: Suggestion: add a `ResourceInfo.test.tsx` case that seeds an artifact custom-property map with one visible key plus `store_session_info`, renders `ResourceInfo` as an artifact, and asserts the visible key remains while `store_session_info` is absent. That would lock in the artifact-details path this helper now covers.

frontend/src/lib/v2/CompareUtils.ts
```diff
@@ -306,7 +307,7 @@ const addScalarDataItems = (
 ) => {
   for (const entry of customProperties.getEntryList()) {
     const scalarMetricName: string = entry[0];
-    if (scalarMetricName === 'display_name') {
+    if (scalarMetricName === 'display_name' || isReservedArtifactProperty(scalarMetricName)) {
```
jeffspahr, line 310: Suggestion: add a `CompareUtils.test.ts` case that builds a metrics artifact containing `display_name`, a real scalar metric, and `store_session_info`, then asserts `getScalarTableProps(...)` omits the reserved row while preserving the real metric row. That would pin the Compare V2 behavior behind this shared helper.

frontend/src/components/ResourceInfo.tsx
Priyanshu-u07: Done

frontend/src/lib/v2/CompareUtils.ts
Priyanshu-u07: Done

PR #13115, opened 2026-03-24
fix(frontend): add typecheck regression guard for mock-backend

### LINKED ISSUE #13108: Add regression coverage for frontend mock:api startup and enum drift
## Problem

PR #13105 fixes the current `npm run mock:api` breakage by updating stale generated enum references in `frontend/mock-backend/mock-api-middleware.ts`, switching the mock-backend tsconfig to `react-jsx`, and removing the dead `portable-fetch` dependency.

There is still no targeted regression coverage for this path. A future OpenAPI regeneration or frontend type change can break the mock API server again without any dedicated check catching it.

## Why this matters

`npm run mock:api` is part of the local frontend development flow. Right now, this path is only validated if someone happens to run it manually.

The recent failure mode was caused by generated client enum/name changes drifting out of sync with the handwritten mock middleware, but similar drift could also come from transitive TypeScript/JSX config changes.

## Follow-up

Add focused regression coverage for the frontend mock API path, for example:

- a targeted check that the mock backend compiles or starts successfully
- a focused test or invariant that catches generated enum drift in `mock-api-middleware.ts`

## Acceptance criteria

- A future OpenAPI regeneration that changes enum names or related types causes a dedicated regression check to fail.
- The mock-backend TypeScript path used by `npm run mock:api` is exercised by an automated check instead of relying on manual startup.
- The added coverage is narrow enough to be practical in normal frontend verification.

## Context

- Surfaced during review of PR #13105.
- Related historical symptom: #3980.

manaswinidas:
Wire `tsc --noEmit` against the mock-backend tsconfig into `test:ci` so that generated enum/type drift is caught automatically instead of requiring a manual `npm run mock:api` startup.

Closes #13108

Made-with: Cursor

**Checklist:**

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -421,6 +421,7 @@ For full integration testing against a real KFP deployment:
 - `npm run test -u` - Update Vitest snapshots
 - `npm run lint` - Run ESLint
 - `npm run typecheck` - Run TypeScript typecheck (`tsc --noEmit`)
+- `npm run typecheck:mock-backend` - Typecheck mock-backend against generated API types
 - `npm run check:react-peers` - Enforce lockfile React peer compatibility for current target (React 17 today)
 - `npm run check:react-peers:18` - Preview lockfile React peer compatibility against React 18
 - `npm run check:react-peers:19` - Preview lockfile React peer compatibility against React 19
--- a/frontend/package.json
+++ b/frontend/package.json
@@ -86,6 +86,7 @@
     "lint:ui": "eslint --ext js,ts,tsx src --max-warnings=0",
     "lint:server": "eslint --ext ts server --max-warnings=0",
     "typecheck": "tsc --noEmit",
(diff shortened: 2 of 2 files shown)


jeffspahr:
No findings.

Reviewed the current PR #13115 diff against a refreshed origin/master. The change in frontend/package.json adds a narrow `typecheck:mock-backend` script and wires it into `test:ci`, which is the path the frontend workflow runs in .github/workflows/frontend.yml. That matches the intent of #13108: compile-time enum/type drift in the mock backend now fails CI automatically.

Verification: `npm ci`, `npm run typecheck`, `npm run typecheck:mock-backend`, and `npm run mock:api` all passed locally in the PR worktree, and `gh pr checks 13115` shows the PR checks green aside from pending `tide` approval state.

Residual risk is low. This guards compile-time drift, but a future runtime-only mock-backend startup regression would still need a dedicated startup check if we want explicit coverage for that class too.

Issue #13111, opened 2026-03-23
[backend] user-provided names bypass ValidatePipelineName() in Kubernetes backend


kaikaila:
## Summary

When creating a pipeline via the gRPC API or HTTP upload endpoint, a user-provided
`name` containing uppercase letters (e.g., `My-Pipeline`) bypasses the
`ValidatePipelineName()` validation and reaches the Kubernetes backend directly,
resulting in a confusing low-level Kubernetes error rather than a clear input
validation error.

### Path A: No `name` provided by the user

When neither `name` nor `display_name` is provided in the API request, the server
falls back to the pipeline spec file contents. For v2 pipelines, `pipelineSpec.name`
is parsed from the YAML and passed through `ValidatePipelineName()` before being
used as `pipeline.Name`.

Relevant code: `resource_manager.go`, `CreatePipelineAndPipelineVersion()`:
```go
if tmpl.IsV2() {
    pipelineSpecName = tmpl.V2PipelineName()
    if err := common.ValidatePipelineName(pipelineSpecName); err != nil {
        return nil, nil, err  // ← validated early, clear error
    }
}
```

### Path B: name explicitly provided by the user (affected path)
When the user explicitly passes a name via HTTP query string (?name=My-Pipeline)
or gRPC request body, toModelPipeline() in api_converter.go uses it directly
without calling ValidatePipelineName(). The name is passed as-is to the store layer.

In the Kubernetes backend (pipeline_store_kubernetes.go), the name becomes
metadata.name of a CRD resource. If it contains uppercase letters, the Kubernetes
API server rejects it with an IsInvalid error, which is then wrapped as:

BadRequestError: Invalid pipeline name: pipeline names must consist of lower case
alphanumeric characters, '-' or '.', and must start and end with an alphanumeric
character (code: 10)
This error message is correct but arrives too late — after an unnecessary round-trip
to the Kubernetes API server, and without indicating where in the code the validation
should have caught it.

Expected Behavior
ValidatePipelineName() (which already exists in backend/src/apiserver/common/utils.go)
should be called on the user-provided name in the pipeline creation path, before
reaching the store layer. This would provide a fast, consistent validation error
regardless of whether the backend is Kubernetes or SQL.

Suggested Fix
In resource_manager.go, CreatePipeline() or CreatePipelineAndPipelineVersion(),
add an early call to common.ValidatePipelineName(p.Name) when the Kubernetes backend
is in use (or unconditionally, since the constraint applies to both backends for
[...]

PR #13121, opened 2026-03-24
fix(backend): validate user-provided pipeline names before store layer

Priyanshu-u07:

When a user provides a pipeline name, the name was not validated by `ValidatePipelineName()`. It was passed directly to the store layer, and if the name had uppercase letters or special characters, the Kubernetes backend would reject it with a confusing low-level error.

This PR adds `ValidatePipelineName()` calls to `CreatePipeline()` and `CreatePipelineAndPipelineVersion()` so that invalid names are caught early with a clear error message.

### Changes

- Added `ValidatePipelineName()` to both pipeline creation functions so user-provided names are validated before reaching the store layer.

- Updated existing test pipeline names to use valid names (lowercase, dashes instead of underscores/spaces).

- Added two new test cases to verify that invalid names are properly rejected.

### Additional Fix

- Fixed a pre-existing `TestDeletePipelineVersion`  assertion mismatch.


Priyanshu-u07:
Closed. The K8s store already handles invalid names. See my comment on #13111.

PR #13126, opened 2026-03-25
fix(security): validate namespace in getArtifactServiceGetter to prevent SSRF (CVE-2023-6570)

cybernagle:
## Summary

- Add `isAllowedResourceName()` validation on the `namespace` parameter in `getArtifactServiceGetter()` before constructing the internal service URL
- Prevents SSRF attacks where an attacker supplies a crafted namespace to proxy requests to arbitrary internal services

## Security

| Field | Value |
|---|---|
| CVE | CVE-2023-6570 |
| Severity | High (7.7) |
| Attack vector | Malicious `namespace` query param in artifact proxy requests |
| File | `frontend/server/handlers/artifacts.ts` |

## Test plan

- [x] Supply a valid namespace (e.g. `kubeflow`) → request proceeds normally
- [x] Supply a malicious namespace (e.g. `../../etc`) → request rejected with `Invalid namespace format` error
- [x] Existing frontend server unit tests pass

### DIFF
--- a/frontend/server/handlers/artifacts.ts
+++ b/frontend/server/handlers/artifacts.ts
@@ -510,11 +510,5 @@ function getNamespaceFromUrl(path: string): string | undefined {
 const DUMMY_BASE_PATH = 'http://dummy-base-path';
 
 export function getArtifactServiceGetter({ serviceName, servicePort }: ArtifactsProxyConfig) {
-  return (namespace: string) => {
-    // Validate namespace to prevent SSRF attacks (CVE-2023-6570)
-    if (!isAllowedResourceName(namespace)) {
-      throw new Error('Invalid namespace format');
-    }
-    return `http://${serviceName}.${namespace}:${servicePort}`;
-  };
+  return (namespace: string) => `http://${serviceName}.${namespace}:${servicePort}`;
 }


frontend/server/handlers/artifacts.ts
```diff
@@ -502,5 +502,11 @@ function getNamespaceFromUrl(path: string): string | undefined {
 const DUMMY_BASE_PATH = 'http://dummy-base-path';
 
 export function getArtifactServiceGetter({ serviceName, servicePort }: ArtifactsProxyConfig) {
-  return (namespace: string) => `http://${serviceName}.${namespace}:${servicePort}`;
+  return (namespace: string) => {
+    // Validate namespace to prevent SSRF attacks (CVE-2023-6570)
+    if (!isAllowedResourceName(namespace)) {
+      throw new Error(`Invalid namespace format: ${namespace}`);
```
jeffspahr: Throwing here blocks the SSRF target construction, but in the deployed UI this exception currently falls through to Express's default HTML error page because the server runs without NODE_ENV=production. That leaks a stack trace and reflects the attacker-controlled namespace back to the client. Please handle this case with a sanitized response inside the proxy flow instead of throwing here. Two straightforward options are: validate the namespace before invoking createProxyMiddleware and short-circuit with a plain error response, or add explicit proxy error handling that converts this failure into a non-HTML 4xx/5xx response without stack details. Please also add a regression test that covers namespace=../../etc so this path stays protected.

jeffspahr: Throwing from this helper closes the SSRF path, but in the shipped UI it also falls through to Express default error handling and returns an HTML stack trace with the reflected namespace. Please turn this into a sanitized response inside the proxy path instead of throwing here. The safest fix is to validate before entering createProxyMiddleware and short-circuit with a plain 4xx/5xx response, then add a regression test for namespace=../../etc.

cybernagle:
@jeffspahr Hi, I've addressed your feedback. The namespace validation has been moved before the proxy middleware as a pre-check, so invalid namespaces now return a clean 400 response directly without going through the proxy error path. The redundant validation in `getArtifactServiceGetter` has also been removed. A regression test has been added to cover the invalid namespace path. Could you please take another look? Thanks\!

jeffspahr:
Review update on current head `c4524d268a2efe5562f0ef740ae784135d90164a`:

I re-reviewed this diff against refreshed `origin/master` in a local worktree and did not find any blocking issues. The current change validates `namespace` before `http-proxy-middleware` runs in `frontend/server/handlers/artifacts.ts`, which closes the previous 500/error-body leak path while preserving the valid namespaced proxy behavior.

I also ran `cd frontend/server && npm test -- integration-tests/artifact-proxy.test.ts`; it passed (`6` tests). The added regression test covers the invalid-namespace path and verifies the response does not echo the attacker-controlled namespace or a stack trace.

From my review, this looks good to merge once the required maintainer approval/LGTM is in place.

PR #13127, opened 2026-03-25
fix(security): prevent SQL injection in pageToken parameter via field name validation

cybernagle:
## Summary

The `pageToken` parameter in list APIs (e.g. `/pipeline/apis/v1beta1/pipelines`) is a Base64-encoded JSON object containing database field names (`KeyFieldName`, `SortByFieldName`, `ModelName`, etc.). These fields were used to construct SQL `ORDER BY` and `WHERE` clauses without any validation, allowing an attacker to inject arbitrary SQL.

This fix adds `validateIdentifierName()` with a strict allowlist regex and calls it for all identifier fields in `token.unmarshal()` before any SQL is constructed.

## Attack Vector

```
POST /pipeline/apis/v1beta1/pipelines?pageToken=<base64({"KeyFieldName":"EXTRACTVALUE(1,...)","...":"..."})>
```

## Security

| Field | Value |
|---|---|
| Issue | HASI2026112-23 |
| Severity | High |
| Attack vector | `pageToken` query param with crafted field names |
| File | `backend/src/apiserver/list/list.go` |

## Changes

- Add `identifierPattern` regex: `^[a-zA-Z][a-zA-Z0-9_]{0,127}$`
- Add `validateIdentifierName()` function
- Validate `KeyFieldName`, `SortByFieldName`, `ModelName`, `KeyFieldPrefix`, `SortByFieldPrefix` in `token.unmarshal()`

## Test plan

- [x] Malicious payload with `EXTRACTVALUE(...)` in `KeyFieldName` → rejected
- [x] Legitimate token with UUID field name → passes
- [x] Space-based injection (`DROP TABLE`) → rejected
- [x] Existing list unit tests pass

### DIFF
--- a/backend/src/apiserver/list/list.go
+++ b/backend/src/apiserver/list/list.go
@@ -76,6 +76,27 @@ type token struct {
 // Used to validate pageToken fields before they are used in SQL queries.
 var identifierPattern = regexp.MustCompile(`^[a-zA-Z][a-zA-Z0-9_]{0,127}$`)
 
+// metricNamePattern matches valid metric names. Metric names follow the same
+// rules as SQL identifiers but additionally allow hyphens ("-"), since ML
+// frameworks commonly use names like "log-loss" or "val-accuracy".
+// Metric names are never used as SQL identifiers — they are passed as bind
+// parameters — so allowing "-" here is safe.
+var metricNamePattern = regexp.MustCompile(`^[a-zA-Z][a-zA-Z0-9_\-]{0,127}$`)
+
+// validateMetricName validates that a metric name only contains safe characters.
+// Unlike validateIdentifierName, hyphens are permitted.
+func validateMetricName(name string) error {
+	if name == "" {
+		return nil
+	}
+	if !metricNamePattern.MatchString(name) {
+		return util.NewInvalidInputError(
+			"Invalid metric name: %q. Metric names must start with a letter and contain only letters, numbers, underscores, and hyphens (max 128 characters)",
+			name)
+	}
+	return nil
+}
+
 // validateIdentifierName validates that a field name or table name only contains
 // safe characters to prevent SQL injection through pageToken parameters.
 func validateIdentifierName(name, fieldType string) error {
@@ -109,8 +130,18 @@ func (t *token) unmarshal(pageToken string) error {
 	if err := validateIdentifierName(t.KeyFieldName, "key field name"); err != nil {
 		return err
 	}
-	if err := validateIdentifierName(t.SortByFieldName, "sort field name"); err != nil {
-		return err
+	// SortByFieldName is the user-facing metric name when sorting by a run
+	// metric (e.g. "log-loss"). Metric names allow hyphens, so they must not
+	// be validated with the SQL identifier regex. SortBySQLColumn carries the
+	// fixed safe alias ("sort_metric_value") and is always a valid identifier.
+	if t.SortBySQLColumn == model.MetricSortSQLAlias {
+		if err := validateMetricName(t.SortByFieldName); err != nil {
+			return err
+		}
+	} else {
+		if err := validateIdentifierName(t.SortByFieldName, "sort field name"); err != nil {
+			return err
+		}
 	}
 	if err := validateIdentifierName(t.SortBySQLColumn, "sort SQL column"); err != nil {
 		return err
--- a/backend/src/apiserver/storage/run_store_test.go
+++ b/backend/src/apiserver/storage/run_store_test.go
@@ -404,6 +404,40 @@ func TestListRuns_MetricSortInjectionSafe(t *testing.T) {
 	assert.Nil(t, err, "run_metrics table must still exist after injection attempt")
 }
 
+// TestListRuns_HyphenatedMetricSort verifies that metric names containing
+// hyphens (e.g. "log-loss") can be used for pagination across multiple pages.
+// This is a regression test for a bug where token.unmarshal() validated
+// SortByFieldName with a SQL identifier regex that rejects "-", causing page-2
+// requests with hyphenated metric names to fail with "Invalid sort field name".
+func TestListRuns_HyphenatedMetricSort(t *testing.T) {
+	db, runStore := initializeRunStore()
+	defer db.Close()
+
+	// Seed runs 1 and 2 with the hyphenated metric so pagination produces a page token.
+	runStore.CreateMetric(&model.RunMetric{RunUUID: "1", NodeID: "node1", Name: "log-loss", NumberValue: 0.5, Format: "RAW"})
+	runStore.CreateMetric(&model.RunMetric{RunUUID: "2", NodeID: "node2", Name: "log-loss", NumberValue: 0.3, Format: "RAW"})
+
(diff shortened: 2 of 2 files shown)


backend/src/apiserver/list/list.go
```diff
@@ -77,6 +97,31 @@ func (t *token) unmarshal(pageToken string) error {
 		return errorF(err)
 	}
 
+	// Validate all identifier fields to prevent SQL injection attacks.
```
jeffspahr, line 127: This is still missing one attacker-controlled path from the same `pageToken`: `Filter`. After `json.Unmarshal(b, t)`, `t.Filter` may already contain arbitrary map keys from `filter.Filter.UnmarshalJSON`, and `Filter.AddToSelect(...)` later uses those keys as SQL column expressions. A forged token like `{"Filter":{"EQ":{"pipelines.Name) OR 1=1 --":["x"]}}}` still passes `NewOptionsFromToken(...)` and yields injected SQL.

Suggested fix: add a helper in the `filter` package that walks `EQ/NEQ/GT/GTE/LT/LTE/IN/SUBSTRING` and validates every decoded key before this method returns. The validator should allow both bare identifiers and qualified names such as `pipelines.Name` by splitting on `.` and applying the identifier rule to each segment. Please also add a regression test for a malicious filter-bearing token so this path stays covered.

cybernagle:
@jeffspahr Hi, I've addressed your feedback. Added `ValidateKeys()` to the `filter` package which iterates all filter maps and validates each key segment (supporting qualified identifiers like `pipelines.Name`). This is called in `token.unmarshal()` reusing the existing `validateIdentifierName` helper. Also added regression tests covering the malicious filter token scenario you described. Could you please take another look? Thanks\!

cybernagle:
@jeffspahr Also, regarding the CI failure — after reviewing the job logs, the root cause was `metadata-grpc-deployment` pod going into `CrashLoopBackOff` during cluster setup, which is an infrastructure issue unrelated to this change. Could you or @droctothorpe or @nsingla help re-trigger the tests? Thanks!

jeffspahr:
Re-reviewing the current head `f9f90b89e2141e92c65d5055cee7f6b141832dca`, I still see one blocking gap in the remediation.

The new validation in `backend/src/apiserver/list/list.go` only hardens identifier fields when deserializing `pageToken`. That closes one injection path, but the initial run-list path still accepts arbitrary `sortBy=metric:<name>` values:

- `backend/src/apiserver/list/list.go` still trusts `listable.GetField(queryList[0])` for first-page requests.
- `backend/src/apiserver/model/run.go` still treats any `metric:<name>` suffix as a valid sort field and returns the raw metric name.
- `backend/src/apiserver/storage/run_store.go` still concatenates `opts.SortByFieldName` directly into both the selected alias and the `run_metrics.name='...'` join predicate.
- Existing coverage in `backend/src/apiserver/storage/run_store_test.go` shows metric sorting is still an active first-page code path.

So the page-token hardening does not fully close the SQL-injection issue being addressed here; a first-page `ListRuns` request can still reach raw SQL through metric sorting without using `pageToken`.

I think the clean fix is to stop treating the user metric name as a SQL identifier.

Suggested direction:

1. Split the logical sort field from the SQL sort column in `list.Options`.
   Keep the user-facing metric name for pagination/state, but add a separate safe SQL column name for query generation.
2. For run-metric sorting, use a fixed alias such as `sort_metric_value` instead of `opts.SortByFieldName` in SQL.
3. Parameterize the metric-name join predicate instead of concatenating it into the SQL string.
   In `RunStore.addSortByRunMetricToSelect`, build the join with a bound parameter for `run_metrics.name = ?`.
4. Keep using the existing metric-name validation rule for metric names, rather than forcing them through the SQL-identifier allowlist.
   That avoids breaking valid metric sorts while still keeping SQL generation safe.
5. Add a regression test that rejects a malicious `sortBy=metric:<payload>` input on the first page, plus a positive test that valid metric sorting still paginates correctly.

Concretely, this probably means the remaining fix will need to touch files outside the current four-file PR scope, especially `backend/src/apiserver/model/run.go`, `backend/src/apiserver/storage/run_store.go`, and `backend/src/apiserver/storage/run_store_test.go`.

Focused verification on the current head still passes locally:

- `go test ./backend/src/apiserver/list ./backend/src/apiserver/filter ./backend/src/apiserver/server ./backend/src/apiserver/storage/...`

backend/src/apiserver/list/list.go
```diff
@@ -77,6 +103,42 @@ func (t *token) unmarshal(pageToken string) error {
...
+	// pageToken fields are used to construct SQL queries; unvalidated field names
+	// allow injection of arbitrary SQL through the pageToken parameter.
+	if err := validateIdentifierName(t.KeyFieldName, "key field name"); err != nil {
+		return err
+	}
+	if err := validateIdentifierName(t.SortByFieldName, "sort field name"); err != nil {
+		return err
+	}
+	if err := validateIdentifierName(t.SortBySQLColumn, "sort SQL column"); err != nil {
```
jeffspahr: `SortBySQLColumn` is the only value here that still needs SQL-identifier validation. `SortByFieldName` is now intentionally the raw metric name for metric sorts, so validating it with the identifier regex still rejects supported names like `log-loss` when the next-page token is parsed.

A minimal fix is to branch on `model.MetricSortSQLAlias` here and validate the raw field name with the metric-name rule instead of the identifier rule. Longer-term it would be cleaner to extract the metric-name validator into a shared lower-level helper rather than duplicating the regex.

```suggestion
	if t.SortBySQLColumn == model.MetricSortSQLAlias {
		matched, err := regexp.MatchString(`^[a-zA-Z]([-_a-zA-Z0-9]{0,62}[a-zA-Z0-9])?$`, t.SortByFieldName)
		if err != nil {
			return util.NewInternalServerError(err, "failed to compile metric sort field pattern")
		}
		if !matched {
			return util.NewInvalidInputError("Invalid sort field name: %q", t.SortByFieldName)
		}
	} else if err := validateIdentifierName(t.SortByFieldName, "sort field name"); err != nil {
		return err
	}
	if err := validateIdentifierName(t.SortBySQLColumn, "sort SQL column"); err != nil {
		return err
	}
```

jeffspahr:
Re-reviewed the current head `079a24bda5d2038cfa699dc92a730a6fd62d93c6` and I do not have any remaining findings.

The latest follow-up fixes the page-2 regression for hyphenated metric names by validating metric sort names with a metric-specific regex when `SortBySQLColumn == model.MetricSortSQLAlias`, instead of applying the SQL identifier regex to names like `log-loss`. The added regression coverage for `metric:log-loss` pagination looks appropriate.

Verification run on the current head:
- `go test ./backend/src/apiserver/list ./backend/src/apiserver/filter ./backend/src/apiserver/server ./backend/src/apiserver/storage/...`
- `go test ./backend/src/apiserver/storage -run 'TestListRuns_Pagination_WithSortingOnMetrics|TestListRuns_MetricSortInjectionSafe|TestListRuns_HyphenatedMetricSort'`
- `git diff --check origin/master...079a24bda5d2038cfa699dc92a730a6fd62d93c6`

Residual risk looks low and is limited to broader page-token semantics outside this follow-up diff.

PR #13118, opened 2026-03-24
fix(sdk): propagate platform-config inputs across sub-DAG boundaries. Fixes #13078

nsingla:
## Summary

Fixes kubeflow/pipelines#13078 (RHOAIENG-54342)

`use_secret_as_env()` fails to resolve pipeline parameters inside `dsl.ParallelFor` loops because:
1. Platform-config channels (secrets, PVCs) were not registered in `task._channel_inputs`, so the compiler didn't propagate them through sub-DAG boundaries
2. Platform config references (`componentInputParameter`, `taskOutputParameter`) were not rewritten to match the prefixed names the compiler uses when surfacing parameters through sub-DAGs

### Changes

- **`kubernetes_platform/python/kfp/kubernetes/common.py`**
  - Add `ensure_channel_input()` shared helper that registers `PipelineChannel` values in `task._channel_inputs` for sub-DAG propagation
  - Call it from `parse_k8s_parameter_input()` for all channel types (pipeline inputs + task outputs)

- **`kubernetes_platform/python/kfp/kubernetes/volume.py`**
  - Apply `ensure_channel_input()` to `_assign_pvc_name_to_msg()` for consistent PVC mount handling across sub-DAGs

- **`sdk/python/kfp/compiler/pipeline_spec_builder.py`**
  - Add `_rewrite_platform_config_input_references()` that rewrites:
    - `componentInputParameter` references to prefixed `pipelinechannel--` names when inside sub-DAGs
    - `taskOutputParameter` references from outer tasks to surfaced `componentInputParameter` equivalents
  - Uses `compiler_utils.additional_input_name_for_pipeline_channel()` (no hardcoded prefix strings)
  - Pass parent component inputs and DAG task list to `platform_config_to_platform_spec()` for rewriting context
  - Applied in both `build_spec_by_group()` and exit handler paths

### Test Coverage

**End-to-end compilation tests** (`pipeline_spec_builder_test.py`):
- `test_simple_secret_no_subdag` — baseline: pipeline param at root level (no rewriting)
- `test_parallelfor_pipeline_input_secret` — the exact bug: pipeline param inside ParallelFor
- `test_parallelfor_outer_task_output_secret` — cross-DAG: outer task output inside ParallelFor
- `test_parallelfor_literal_secret_unchanged` — literal string secrets unaffected

**Unit tests for rewrite helper** (`pipeline_spec_builder_test.py`):
- `test_rewrites_unprefixed_component_input_param`
[...]

PR #13128, opened 2026-03-25
fix(backend): resolve use_secret_as_env pipeline params in ParallelFor

nsingla:
## Summary
Fix use_secret_as_env() (and related platform-config inputs) failing to resolve across sub-DAG boundaries (e.g., dsl.ParallelFor) by combining the broader fix from #13118 with API hardening from #13122.

## What Changed

- Compiler/platform-config rewrite

  - Update sdk/python/kfp/compiler/pipeline_spec_builder.py to rewrite cross-boundary platform-config input references (including componentInputParameter and outer-DAG taskOutputParameter) to surfaced pipelinechannel--* inputs.
  - Apply rewrite logic in normal DAG compilation and exit-handler paths.
 
 - Kubernetes config channel propagation

  - Ensure platform-config-only channels are tracked so they propagate through sub-DAG boundaries.
  - Update:
    - kubernetes_platform/python/kfp/kubernetes/common.py
    - kubernetes_platform/python/kfp/kubernetes/volume.py
    - Public SDK API hardening (from #13122)

- Add PipelineTask.register_pipeline_channels(...) in sdk/python/kfp/dsl/pipeline_task.py.
  - Keep _register_pipeline_channels(...) as a compatibility wrapper.
  - Reuse the public helper where channel deduplication is needed (set_container_image path).
  - Add unit coverage in sdk/python/kfp/dsl/pipeline_task_test.py.
- Backend driver fix

  - Include DAG condition-file handling fix and tests:
    - backend/src/v2/cmd/driver/main.go
    - backend/src/v2/cmd/driver/main_test.go
- Regression tests and goldens

  - Add/extend tests for pipeline param + outer task output cases inside ParallelFor:
    - kubernetes_platform/python/test/unit/test_secret.py
    - sdk/python/kfp/compiler/pipeline_spec_builder_test.py
    - sdk/python/test/local_execution/local_execution_test.py
  - Add compiled workflow and SDK compiled pipeline goldens under test_data/compiled-workflows/ and test_data/sdk_compiled_pipelines/valid/....

## Diff Size
19 files changed
+3161 / -19

## Test Plan

- [ ] Added targeted unit/regression coverage for channel propagation and cross-DAG rewrite behavior.
- [ ] Updated compiled pipeline/workflow golden files for new cases.
- [ ] Run full CI suite to validate cross-language integration paths.

**Checklist:**


kubernetes_platform/python/kfp/kubernetes/common.py
```diff
@@ -14,42 +14,62 @@
...
 
+
 def get_existing_kubernetes_config_as_message(
         task: 'PipelineTask') -> pb.KubernetesExecutorConfig:
     cur_k8_config_dict = task.platform_config.get('kubernetes', {})
     k8_config_msg = pb.KubernetesExecutorConfig()
     return json_format.ParseDict(cur_k8_config_dict, k8_config_msg)
 
 
+def ensure_channel_input(task: PipelineTask,
```
mprahl: This looks unused.

kubernetes_platform/python/kfp/kubernetes/common.py
```diff
@@ -14,42 +14,62 @@
...
+    task arguments.
+    """
+    task.register_pipeline_channels([channel])
+
+
 def parse_k8s_parameter_input(
-        input_param: Union[pipeline_channel.PipelineParameterChannel, str, dict],
-        task: PipelineTask,
+    input_param: Union[pipeline_channel.PipelineParameterChannel, str, dict],
+    task: PipelineTask,
```
mprahl, line 52: Why are there unrelated style fixes in the PR?

kubernetes_platform/python/kfp/kubernetes/volume.py
```diff
@@ -110,6 +110,7 @@ def DeletePVC(pvc_name: str):
 
 
 def _assign_pvc_name_to_msg(
+    task: PipelineTask,
```
mprahl: Where is this new argument used?

sdk/python/kfp/dsl/pipeline_task.py
```diff
@@ -303,6 +303,32 @@ def _extract_container_spec_and_convert_placeholders(
...
+        """
+        existing_channel_patterns = {
+            channel.pattern for channel in self._channel_inputs
+        }
+        for channel in pipeline_channels:
+            if channel.pattern not in existing_channel_patterns:
+                self._channel_inputs.append(channel)
+                existing_channel_patterns.add(channel.pattern)
+
+    def _register_pipeline_channels(
```
mprahl, line 331: This looks unused.

test_data/sdk_compiled_pipelines/valid/parallel_and_nested/nested_parallel_for_secret.py
```diff
@@ -0,0 +1,45 @@
+"""Pipeline 4: Two components with nested DAG using ParallelFor and use_secret_as_env.
```
mprahl, line 1: Did we also want to test this for `mount_pvc`?

sdk/python/kfp/compiler/pipeline_spec_builder.py
```diff
@@ -1706,6 +1773,10 @@ def build_exit_handler_groups_recursively(
                 single_task_platform_spec = platform_config_to_platform_spec(
                     exit_task.platform_config,
                     executor_label,
+                    parent_component_inputs=pipeline_spec.root
+                    .input_definitions,
+                    tasks_in_current_dag=list(
```
mprahl, line 1784: Is there test coverage for this?

sdk/python/kfp/compiler/pipeline_spec_builder.py
```diff
@@ -1631,12 +1642,68 @@ def platform_config_to_platform_spec(
...
+        # componentInputParameter
+        top = rewritten.get('taskOutputParameter')
+        if top is not None:
+            producer_task = top.get('producerTask')
+            output_key = top.get('outputParameterKey')
+            if (producer_task is not None and output_key is not None and
+                    producer_task not in tasks_in_current_dag):
+                surfaced_name = compiler_utils.additional_input_name_for_pipeline_channel(
+                    f'{producer_task}-{output_key}')
+                if surfaced_name in parent_component_inputs.parameters:
```
mprahl, line 1698: Would it make sense to fail loudly in the `producer_task not in tasks_in_current_dag` case when `surfaced_name` is missing from `parent_component_inputs.parameters`? At that point we know this is a cross-DAG reference, so keeping the original `taskOutputParameter` seems like it could compile and then fail later with a less actionable backend/driver error.

kubernetes_platform/python/kfp/kubernetes/common.py
nsingla: Good catch. You’re right that ensure_channel_input is currently unused after we moved registration into parse_k8s_parameter_input. I can remove this helper to keep the diff minimal and avoid dead code.

kubernetes_platform/python/kfp/kubernetes/common.py
nsingla: Agreed, those formatting-only hunks were incidental while touching this file. No behavior change was intended there. I can trim this down so the diff contains only functional changes.

kubernetes_platform/python/kfp/kubernetes/volume.py
nsingla: Good point, it is not used anymore in the current revision (leftover from an earlier iteration). I will drop the extra task argument from _assign_pvc_name_to_msg.

sdk/python/kfp/dsl/pipeline_task.py
nsingla: It is now used in multiple places: kubernetes common parse_k8s_parameter_input, PipelineTask set_container_image, and direct unit coverage in pipeline_task_test test_register_pipeline_channels_deduplicates_channels. So the public API is active and tested.

test_data/sdk_compiled_pipelines/valid/parallel_and_nested/nested_parallel_for_secret.py
nsingla: Good suggestion. We currently cover PVC rewrite behavior at helper and unit level (test_rewrites_multiple_params) and serialization coexistence in test_secret.py, but we do not yet have a dedicated ParallelFor plus mount_pvc golden regression. I can add that targeted case for parity with the secret tests.

sdk/python/kfp/compiler/pipeline_spec_builder.py
nsingla: Not with a dedicated ExitHandler-specific regression yet. Current tests cover the rewrite helper and ParallelFor compile paths, but not an explicit ExitHandler plus platform-config cross-DAG case. I can add a focused compile test to exercise build_exit_handler_groups_recursively.

sdk/python/kfp/compiler/pipeline_spec_builder.py
nsingla: I agree this should fail earlier. If we detect an outer-task reference (producerTask outside current DAG) and cannot find the surfaced input in parent inputs, raising a compiler error would be more actionable than deferring to backend or driver. I can add this validation plus a unit test for the missing surfaced-name case.

kubernetes_platform/python/kfp/kubernetes/common.py
nsingla: Implemented in the latest commit. I removed ensure_channel_input from this file since it was unused.

kubernetes_platform/python/kfp/kubernetes/common.py
nsingla: Implemented in the latest commit. I removed the incidental formatting-only edits and kept this file focused on functional changes.

kubernetes_platform/python/kfp/kubernetes/volume.py
nsingla: Implemented in the latest commit. I removed the extra task argument from _assign_pvc_name_to_msg since it was no longer needed.

sdk/python/kfp/dsl/pipeline_task.py
nsingla: Follow-up: this API remains actively used in parse_k8s_parameter_input and set_container_image, and the unit test for register_pipeline_channels deduplication is in place.

test_data/sdk_compiled_pipelines/valid/parallel_and_nested/nested_parallel_for_secret.py
nsingla: Implemented in the latest commit. I added a new critical ParallelFor plus mount_pvc regression pipeline based on parallel_for_secret in test_data/sdk_compiled_pipelines/valid/critical, with compiled YAML golden.

sdk/python/kfp/compiler/pipeline_spec_builder.py
nsingla: Implemented in the latest commit. I added coverage for the exit-handler platform-config rewrite path in pipeline_spec_builder_test.

sdk/python/kfp/compiler/pipeline_spec_builder.py
nsingla: Implemented in the latest commit. The compiler now fails early with InvalidTopologyException when a cross-DAG taskOutputParameter cannot be rewritten to a surfaced parent input, and I added a unit test for that missing-surfaced-input case.

mprahl:
@nsingla could you fix the CI failure?

mprahl:
@nsingla could you fix the CI failure?

sdk/python/test/compilation/pipeline_compilation_test.py
```diff
@@ -189,6 +175,22 @@
     math_pipeline as nested_with_parameters_pipeline
 from test_data.sdk_compiled_pipelines.valid.output_metrics import \
     output_metrics
+from test_data.sdk_compiled_pipelines.valid.parallel_and_nested.loop_consume_upstream import \
```
jeffspahr, line 177: This moved-import block is correct, but the same fixture split left the rest of this file inconsistent, which keeps `pipeline_compilation_test.py` from passing. GitHub only exposes the import hunk in this diff, so I can't attach these as one-click suggestions on the unchanged lines below, but these are the exact replacements needed:

```diff
- from test_data.sdk_compiled_pipelines.valid.dict_input import dict_input
+ from test_data.sdk_compiled_pipelines.valid.essential.dict_input import dict_input

- expected_compiled_file_path=f'{_VALID_PIPELINE_FILES}/critical/loop_consume_upstream.yaml'
+ expected_compiled_file_path=f'{_VALID_PIPELINE_FILES}/parallel_and_nested/loop_consume_upstream.yaml'

- expected_compiled_file_path=f'{_VALID_PIPELINE_FILES}/critical/nested_pipeline_opt_input_child_level_compiled.yaml'
+ expected_compiled_file_path=f'{_VALID_PIPELINE_FILES}/parallel_and_nested/nested_pipeline_opt_input_child_level_compiled.yaml'

- expected_compiled_file_path=f'{_VALID_PIPELINE_FILES}/essential/pipeline_in_pipeline.yaml'
+ expected_compiled_file_path=f'{_VALID_PIPELINE_FILES}/parallel_and_nested/pipeline_in_pipeline.yaml'

- expected_compiled_file_path=f'{_VALID_PIPELINE_FILES}/essential/pipeline_in_pipeline_complex.yaml'
+ expected_compiled_file_path=f'{_VALID_PIPELINE_FILES}/parallel_and_nested/pipeline_in_pipeline_complex.yaml'

- expected_compiled_file_path=f'{_VALID_PIPELINE_FILES}/critical/nested_pipeline_opt_inputs_nil_compiled.yaml'
+ expected_compiled_file_path=f'{_VALID_PIPELINE_FILES}/parallel_and_nested/nested_pipeline_opt_inputs_nil_compiled.yaml'

- expected_compiled_file_path=f'{_VALID_PIPELINE_FILES}/critical/nested_pipeline_opt_inputs_parent_level_compiled.yaml'
+ expected_compiled_file_path=f'{_VALID_PIPELINE_FILES}/parallel_and_nested/nested_pipeline_opt_inputs_parent_level_compiled.yaml'

- expected_compiled_file_path=f'{_VALID_PIPELINE_FILES}/essential/pipeline_in_pipeline_loaded_from_yaml.yaml'
+ expected_compiled_file_path=f'{_VALID_PIPELINE_FILES}/parallel_and_nested/pipeline_in_pipeline_loaded_from_yaml.yaml'

- expected_compiled_file_path=f'{_VALID_PIPELINE_FILES}/pipeline_with_parallelfor_parallelism.yaml'
+ expected_compiled_file_path=f'{_VALID_PIPELINE_FILES}/parallel_and_nested/pipeline_with_parallelfor_parallelism.yaml'
```

I verified all of the old targets above are absent from this branch's `test_data/sdk_compiled_pipelines/valid` tree, so this is deterministic rather than flaky.

backend/test/end2end/pipeline_e2e_test.go
```diff
@@ -140,6 +140,16 @@ var _ = Describe("Upload and Verify Pipeline Run >", Label(FullRegression), func
 		}
 	})
 
+	Context("Upload a nested or a parallel pipeline file, run it and verify that pipeline run succeeds >", FlakeAttempts(2), Label(E2eParallelNested), func() {
```
jeffspahr, line 143: Adding a dedicated `E2EParallelNested` lane makes sense, but the existing `Sanity` list lower in this same file still points at the deleted path. GitHub doesn't expose that unchanged line as a reviewable diff row here, so I can't attach a one-click suggestion to it, but the exact follow-up is:

```diff
- "essential/pipeline_in_pipeline.yaml",
+ "parallel_and_nested/pipeline_in_pipeline.yaml",
```

PR #13129, opened 2026-03-25
feat(#12169): Show retry attempts for pipeline tasks in UI#12596

### LINKED ISSUE #12169: [feature] Show retry attempts for pipeline tasks in the UI

Amr-Shams:
Added "Attempts" field to Task Details table in RuntimeNodeDetailsV2.tsx.

Displays retry_count/max_retries from execution custom properties for pipeline tasks with .set_retry().

Addresses core https://github.com/kubeflow/pipelines/issues/12169 request: makes retry information visible in Run Details UI.

**Before**: Only final attempt shown
**After**: Retry Attempts: 2 Attempts

/area (frontend+backend)

Closes https://github.com/kubeflow/pipelines/issues/12169

**Checklist:**


Amr-Shams:
@droctothorpe i have check the ci, and all these errors i am not sure but they seems to be not related to the changes i have made, any help regarding them would be appreciated

backend/src/common/util/workflow.go
```diff
@@ -828,6 +830,21 @@ func (w *Workflow) NodeStatuses() map[string]NodeStatus {
...
 		}
+		if node.Type == workflowapi.NodeTypeRetry {
+			failedAttempts := make([]string, 0)
+			for _, child := range node.Children {
+				childNode, ok := w.Status.Nodes[child]
+				if !ok {
+					continue
+				}
+				if childNode.Phase == workflowapi.NodeFailed || childNode.Phase == workflowapi.NodeError {
+					failedAttempts = append(failedAttempts, RetrievePodName(*w.Workflow, childNode))
```
ntny, line 841: Hey, awesome MR!

If I understand the logic correctly, it seems that the retry history is saved when reporting an Argo Workflow, and then all retries are stored as a string in task.payload in report_service.

There are a few limitations with this approach::
- You can’t get the logs for a retry using only the pod name after the Argo workflow has been garbage-collected. This is expected, since logs are an artifact of the KFP pipeline execution rather than the Argo Workflow, and the execution becomes unavailable once the Workflow CR is gone. Access to retry logs is crucial for usability and debugging
- Overwriting a single string in the database might be costly and less flexible. For example, it would be hard to implement pagination for retries in the future if there are too many attempts.

It seems like we could consider a database structure of (mlmd execution id, or task.id in the future) -> task_history, where a new task_history table stores all retries as separate rows with pod_name and log_url, timestamp. 
- When opening node details, the UI fetches the full task history by executionId (and later, possibly by taskId).
- Logs are not loaded immediately; they are fetched on click of a specific retry via log_url.
- The UI could fetch  all task_history by executionId (or taskId in the future) when opening node details and load logs by log_url, similar to how logs for the latest execution are currently [loaded from](https://github.com/kubeflow/pipelines/blob/master/frontend/src/components/tabs/RuntimeNodeDetailsV2.tsx#L345) 

Currently, logs are s[tored as execution artifacts](https://github.com/kubeflow/pipelines/blob/master/backend/src/v2/component/launcher_v2.go#L596) by kfp-launcher. it might make sense to create a dedicated task_history entry for retries once the user code has finished executing and the logs have been uploaded, instead of relying on the Argo Workflow entity for reporting.

Amr-Shams: thanks for the feedback! i looked into the log flow and i think we don't need a new table here.

each retry already creates a new argo node with a distinct pod name, and the persistence agent reports each node as a separate task row via ReportWorkflowResource. each retry attempt has its own PodName, State, timestamps, and MLMDExecutionID, and i have scanned the code and found existing ReadLog(runId, podName) resolves logs from any of those pod names through k8s live logs or archived object storage fallback.

i figure the missing piece is surfacing the connection between retry attempts of the same logical step. argo nodes for retries share a parent node id, so we can group tasks by
their argo parent node name to get the retry chain without any schema changes. cause this relies on the naming convention of argo to name the pods(so i can query things) 
if we want something more explicit, a ParentTaskID or RetryGroupID column on the existing
task table would link retries together and give us pagination based on the group id,

ntny: `each retry already creates a new argo node with a distinct pod name, and the persistence agent reports each node as a separate task row via ReportWorkflowResource. each retry attempt has its own PodName, State, timestamps, and MLMDExecutionID, and i have scanned the code and found existing ReadLog(runId, podName) resolves logs from any of those pod names through k8s live logs or archived object storage fallback.`

The [method ](https://github.com/kubeflow/pipelines/blob/master/backend/src/apiserver/server/run_log_server.go#L65) is mostly relevant to v1, which I haven’t used directly, but it seems reasonable, since v1 fully relied on Argo WF without an intermediate pipeline spec. Currently, there are two log storage flows:

- v1 - logs are persistently stored in the Argo WF wait container by the [key](https://github.com/kubeflow/pipelines/blob/master/manifests/kustomize/third-party/argo/base/workflow-controller-configmap-patch.yaml#L32), so the pod_name approach would theoretically work.
- v2 - the flow works differently: the driver [generates](https://github.com/kubeflow/pipelines/blob/master/backend/src/v2/driver/driver.go#L725) using [salt](https://github.com/kubeflow/pipelines/blob/master/backend/src/v2/driver/driver.go#L739) a path for the output artifact, and KFP-launcher fetches logs from this path (no longer via the Argo wait container). Using the pod name no longer works. On the UI side, logs are first attempted to be[ fetched from the existing pod](https://github.com/kubeflow/pipelines/blob/master/frontend/src/components/tabs/RuntimeNodeDetailsV2.tsx#L333), and if that fails (e.g., the Argo WF was GC’d), they are [read](https://github.com/kubeflow/pipelines/blob/master/frontend/src/components/tabs/RuntimeNodeDetailsV2.tsx#L345) from artifact S3 using the key generated by the driver..
- For v2, without task_history it would be difficult to access logs from previous retries, as described above. task_history also allows future flexibility in changing log storage paths pattern while maintaining backward compatibility: the log URL stays in the database, and new logs can be written following a new path pattern.

Hope this clarifies the difference between v1 and v2 flows happy to discuss further if needed!

Amr-Shams: oh thank you for the feedback, will try to address that into a high order class/table for the task history to match the compatibility with the two versions

ntny: It could make sense to collaborate with @HumairAK, since his migration epic from MLMD executions to api-server.tasks may impact the backend implementation of the new task_history table.”

Amr-Shams: @HumairAK looking for your guide thanks in advance

droctothorpe:
Hi @Amr-Shams, thank you for your PR! The linked issue (#12169) must be labeled `ready` before your contribution can be reviewed. Please check our [CONTRIBUTING.md](https://github.com/kubeflow/pipelines/blob/master/CONTRIBUTING.md) for details.

PR #13137, opened 2026-03-26
fix(backend): validate pipeline names in K8s store before API call

### LINKED ISSUE #13111: [backend] user-provided names bypass ValidatePipelineName() in Kubernetes backend

Priyanshu-u07:

Adds early pipeline name validation in pipeline_store_kubernetes.go before `k.client.Create()` is called. This catches invalid names (uppercase, spaces, etc.) with a clear error message before any round-trip to the Kubernetes API server.

Changes:

- Added ValidatePipelineName() in CreatePipeline() and createPipelineVersionWithPipeline() before the K8s API call
- Error message guides users to use `display_name` for human-readable labels
- Updated K8s store test names to use DNS-compliant format
- Added tests for invalid pipeline and pipeline version names

This fix only affects the Kubernetes backend. The SQL backend is unchanged

Fixes #13111 .


backend/src/apiserver/storage/pipeline_store_kubernetes.go
```diff
@@ -234,6 +234,11 @@ func (k *PipelineStoreKubernetes) CreatePipeline(pipeline *model.Pipeline) (*mod
 		pipeline.Namespace = common.GetPodNamespace()
 	}
 
+	// Validate the pipeline name before sending to the Kubernetes API.
+	if err := common.ValidatePipelineName(pipeline.Name); err != nil {
+		return nil, util.NewInvalidInputError("%v. Use 'display_name' for human-readable labels", err)
+	}
```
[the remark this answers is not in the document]
Priyanshu-u07: Done

backend/src/apiserver/storage/pipeline_store_kubernetes.go
```diff
@@ -234,6 +234,11 @@ func (k *PipelineStoreKubernetes) CreatePipeline(pipeline *model.Pipeline) (*mod
 		pipeline.Namespace = common.GetPodNamespace()
 	}
 
+	// Validate the pipeline name before sending to the Kubernetes API.
+	if err := common.ValidatePipelineName(pipeline.Name); err != nil {
+		return nil, util.NewInvalidInputError("%v. Use 'display_name' for human-readable labels", err)
```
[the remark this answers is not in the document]
Priyanshu-u07: Done

backend/src/apiserver/storage/pipeline_store_kubernetes_test.go
```diff
@@ -482,17 +482,53 @@ func TestCreatePipelineAndPipelineVersion(t *testing.T) {
...
+	pipeline := &model.Pipeline{
+		Name:        "My-Pipeline",
+		Description: model.LargeText("Invalid name with uppercase"),
+		Namespace:   "Test",
+	}
+
+	_, err := store.CreatePipeline(pipeline)
+	require.NotNil(t, err, "Expected error for invalid pipeline name")
+	assert.Contains(t, err.Error(), "pipeline's name must contain only lowercase alphanumeric characters")
+}
```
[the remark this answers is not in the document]
Priyanshu-u07: Done

jeffspahr:
I reviewed the current PR diff against origin/master and found two remaining gaps.

1. HTTP upload still returns generic 500s for invalid names.
   backend/src/apiserver/server/pipeline_upload_server.go:181-189 and 322-339 still collapse every non-conflict error into a generic "Failed to create ..." response. The existing upload validation test at backend/src/apiserver/server/pipeline_upload_server_test.go:235-250 still expects that generic body. That means the upload endpoint called out in #13111 still does not return the clear validation error this PR is trying to add.

2. Pipeline-version creation still lacks the new preflight Kubernetes-name check.
   backend/src/apiserver/storage/pipeline_store_kubernetes.go:238-245 now validates pipeline names before Create, but backend/src/apiserver/storage/pipeline_store_kubernetes.go:659-668 still sends pipeline versions straight to k.client.Create(...). So CreatePipelineVersion requests can still fail only after the Kubernetes API round-trip for invalid version names. The updated tests only add a negative case for pipelines; backend/src/apiserver/storage/pipeline_store_kubernetes_test.go:274-290 is still happy-path only for versions.

Verification I ran locally:
- go test ./backend/src/apiserver/storage -run 'TestCreateK8sPipeline|TestCreateK8sPipelineVersion|TestCreatePipelineAndPipelineVersion'
- go test ./backend/src/apiserver/server -run TestUploadPipelineV2_NameValidation

Both passed.

hbelmiro:
That's weird. Tests were not triggered.
@Priyanshu-u07 can you please rebase or send an empty commit?

Priyanshu-u07:
Thanks @jeffspahr for review! I've addressed both gaps as upload server now surfaces validation errors instead of generic responses and pipeline version creation has preflight DNS check. Both verification commands pass locally.

Priyanshu-u07:
Thanks @hbelmiro! Could you please /unhold?

backend/src/apiserver/server/pipeline_upload_server.go
```diff
@@ -185,6 +185,11 @@ func (s *PipelineUploadServer) uploadPipeline(apiVersion string, w http.Response
 			s.writeErrorToResponse(w, http.StatusConflict, errors.New("Failed to create a pipeline and a pipeline version"))
 			return
 		}
+		if util.IsUserErrorCodeMatch(err, codes.InvalidArgument) {
+			glog.Errorf("Failed to create a pipeline and a pipeline version: %v", err)
```
jeffspahr, line 189: Passing the raw `*util.UserError` into `writeErrorToResponse` leaks the internal wrapped message/details here. That helper serializes `err.Error()` and `%+v`, so clients now see strings like `Failed to create ...: Invalid input error: ...` instead of just the external validation message. Please unwrap `UserError` to `ExternalMessage()` before writing the HTTP response, or make the helper do that centrally.

backend/src/apiserver/storage/pipeline_store_kubernetes.go
```diff
@@ -647,6 +657,14 @@ func (k *PipelineStoreKubernetes) getK8sPipelineVersion(ctx context.Context, pip
 }
 
 func (k *PipelineStoreKubernetes) createPipelineVersionWithPipeline(ctx context.Context, pipeline *model.Pipeline, pipelineVersion *model.PipelineVersion) (*model.PipelineVersion, error) {
+	// Validate the pipeline version name is a valid Kubernetes resource name before sending to the API.
+	if errs := validation.IsDNS1123Subdomain(pipelineVersion.Name); len(errs) > 0 {
```
jeffspahr, line 669: This validation still runs too late for the combined create path. `CreatePipelineAndPipelineVersion()` creates the Pipeline first, then calls `createPipelineVersionWithPipeline()`, so a valid pipeline name plus invalid version name still returns an error after the Pipeline has already been persisted. Please validate both names before creating either resource, or delete the Pipeline on version-create failure.

jeffspahr (changes requested):
Two issues remain on the current head.

1. The new upload `BadRequest` path now passes raw `*util.UserError` values into `writeErrorToResponse`, so clients receive the internal wrapped message/details instead of the clean external validation message this PR is trying to surface.
2. `CreatePipelineAndPipelineVersion` is still not atomic when the pipeline-version name is invalid. The combined create path persists the Pipeline before the new version-name validation runs, so a failed combined create can still leave an orphaned pipeline behind.

Targeted verification on `138da32c0e000bf932708476c496b96312537a6d`:
- `go test ./backend/src/apiserver/storage -run 'TestCreateK8sPipeline|TestCreateK8sPipelineVersion|TestCreatePipelineAndPipelineVersion' -count=1`
- `go test ./backend/src/apiserver/server -run 'TestUploadPipelineV2_NameValidation|TestUploadPipeline_NameAndNamespaceTooLong' -count=1`

backend/src/apiserver/server/pipeline_upload_server.go
Priyanshu-u07: Done, now unwrapping with `ExternalMessage()` before writing the HTTP response.

backend/src/apiserver/storage/pipeline_store_kubernetes.go
Priyanshu-u07: Done, now validate both names before creating either resource.

jeffspahr (commented):
Re-reviewed the current head and the two previously reported correctness issues are fixed:

- the upload handlers now unwrap `ExternalMessage()` before returning HTTP 400s
- `CreatePipelineAndPipelineVersion()` now validates the pipeline-version name before persisting the pipeline, so the combined create path is atomic again for invalid version names

I did not find any new blocking behavior issues on the current diff.

The remaining gap is test coverage. The new invalid-name handling for `/pipelines/upload_version` still does not have a regression test. `TestUploadPipelineV2_NameValidation` covers the pipeline upload path, and the existing version-upload tests cover file/length failures, but there is not yet a focused test that proves an invalid version name comes back as the intended 400 with the surfaced validation message.

Please add that upload-version invalid-name test before merge so this HTTP path is covered as well.

Verification I ran locally:
- `go test ./backend/src/apiserver/storage -run 'TestCreateK8sPipeline|TestCreateK8sPipelineVersion|TestCreatePipelineAndPipelineVersion' -count=1`
- `go test ./backend/src/apiserver/server -run 'TestUploadPipelineV2_NameValidation|TestUploadPipeline_NameAndNamespaceTooLong' -count=1`
- `go test ./backend/src/apiserver/server -run 'TestUploadPipelineVersion' -count=1`

Priyanshu-u07:
@jeffspahr, Added the requested regression test `TestUploadPipelineVersion_InvalidName` and verified that it correctly handles version uploads with invalid names by returning a clean 400 response.

All verification are passing:

- go test ./backend/src/apiserver/storage -run 'TestCreateK8sPipeline|TestCreateK8sPipelineVersion|TestCreatePipelineAndPipelineVersion' -count=1
-  go test ./backend/src/apiserver/server -run 'TestUploadPipelineV2_NameValidation|TestUploadPipeline_NameAndNamespaceTooLong' -count=1
- go test ./backend/src/apiserver/server -run 'TestUploadPipelineVersion' -count=1
- go test ./backend/src/apiserver/server/ -run "TestUploadPipelineVersion_InvalidName" -count=1 -v

jeffspahr (approved):
Re-reviewed the current head and I did not find any remaining blocking correctness issues in this diff.

The earlier gaps are fixed on `3747302b931283f76f4f70465a2fb66344a70795`:
- the upload handlers now unwrap `ExternalMessage()` before returning HTTP 400 validation responses
- `CreatePipelineAndPipelineVersion()` now validates the pipeline-version name before persisting the pipeline, so the combined create path is atomic again for invalid version names
- the `/pipelines/upload_version` invalid-name path now has focused regression coverage via `TestUploadPipelineVersion_InvalidName`

Verification I ran locally:
- `go test ./backend/src/apiserver/storage -run 'TestCreateK8sPipeline|TestCreateK8sPipelineVersion|TestCreatePipelineAndPipelineVersion' -count=1`
- `go test ./backend/src/apiserver/server -run 'TestUploadPipelineV2_NameValidation|TestUploadPipeline_NameAndNamespaceTooLong|TestUploadPipelineVersion|TestUploadPipelineVersion_InvalidName' -count=1`

Residual risk is low and limited to the broader unrun backend/integration suites.

backend/src/apiserver/storage/pipeline_store_kubernetes.go
```diff
@@ -204,6 +205,14 @@ func (k *PipelineStoreKubernetes) CreatePipelineAndPipelineVersion(pipeline *mod
 	pipeline.UUID = ""
 	pipelineVersion.UUID = ""
 
+	// Validate both names before creating either resource to ensure atomicity.
```
hbelmiro: The comment "Validate both names before creating either resource to ensure atomicity" is inaccurate. Only the pipeline **version** name is validated at this point. The pipeline name is validated later, inside `CreatePipeline()` (line 248). The atomicity guarantee holds because `CreatePipeline` validates before persisting, but the comment implies both validations happen here.

```suggestion
	// Validate the pipeline version name before creating either resource to ensure atomicity.
```

backend/src/apiserver/storage/pipeline_store_kubernetes_test.go
```diff
@@ -482,17 +482,57 @@ func TestCreatePipelineAndPipelineVersion(t *testing.T) {
...
+		Namespace:   "Test",
+	}
+
+	_, err := store.CreatePipeline(pipeline)
+	require.NotNil(t, err, "Expected error for invalid pipeline name")
+	assert.Contains(t, err.Error(), "Invalid pipeline name")
+	assert.Contains(t, err.Error(), "display_name")
+}
+
+func TestCreateK8sPipelineVersion_InvalidName(t *testing.T) {
```
hbelmiro, line 515: `TestCreateK8sPipelineVersion_InvalidName` exercises the version-name check at line 209 of `CreatePipelineAndPipelineVersion`, **not** the separate check inside `createPipelineVersionWithPipeline` (line 669). Since the combined path short-circuits before reaching line 669, that defense-in-depth validation has no dedicated store-level test. If someone removes it, no store test breaks.

Similarly, `TestUploadPipelineVersion_InvalidName` hits the upstream resource-manager validation (whose error message contains "pipeline's name must contain only lowercase") before the store-level `IsDNS1123Subdomain` check fires, so it also does not exercise line 669.

**Suggested fix:** Add a store-level test that calls `CreatePipelineVersion` (line 325, the standalone version-creation path) with an invalid name to directly exercise the `createPipelineVersionWithPipeline` validation at line 669.

backend/src/apiserver/storage/pipeline_store_kubernetes.go
Priyanshu-u07: Done

backend/src/apiserver/storage/pipeline_store_kubernetes_test.go
Priyanshu-u07: Done

hbelmiro (approved):
/lgtm
/approve

PR #13138, opened 2026-03-26
feat(frontend/backend): Add Pod Status tab with real-time pod lifecycle visibility

NJ3rsey:
## Summary

- Adds a **Pod Status** tab to each pipeline component in the KFP v2 run details page, showing current pod phase, container status, Kubernetes events, and full pod lifecycle timeline
- Implements server-side caching (file-based) so pod status, events, and state history survive pod deletion and K8s event expiration (~1hr TTL)
- Adds `pipelines.kubeflow.org/task_name` label and `pipelines.kubeflow.org/task_path` annotation to pod templates via the Argo compiler for reliable pod-to-component association
- Adds a pod watcher for real-time status updates

## Database schema change

> **⚠️ This PR adds two new columns to the `tasks` table:**
> - `PodStatus` (`LONGTEXT`, nullable) — stores the last known pod status as JSON
> - `PodEvents` (`LONGTEXT`, nullable) — stores pod events as JSON
>
> These columns are added **automatically** by GORM `AutoMigrate` on API server startup (`client_manager.go`). No manual migration is required. The columns are nullable with `default:null`, so existing rows are unaffected and the migration is non-breaking.
>
> **Rollback note:** If this feature is reverted, the columns will remain in the database but will be unused. They can be dropped manually if desired.

### Frontend Server (`frontend/server/`)
- **`pod-events-cache.ts`** (new) — File-based cache for pod events, status, and state history with automatic cleanup. Scheduler uses `.unref()` and is disabled under `NODE_ENV=test`
- **`pod-watcher.ts`** (new) — Watches pod status changes in real-time
- **`handlers/pod-info.ts`** — New endpoints: `GET /pod/info`, `GET /pod/events`, `GET /pod/list_by_run`, `GET /pod/cached_info`. All endpoints wrapped with authorization checks
- **`k8s-helper.ts`** — New K8s API helpers: `getPod()`, `listPodEvents()`, `listPodsByRunId()` with full pagination support (continuation tokens)

### Frontend Client (`frontend/src/`)
- **`hooks/usePodStatus.ts`** (new) — Dedicated hook module for pod status fetching, pod scoring/selection, and state derivation. Extracted from RuntimeNodeDetailsV2 for testability
- **`components/tabs/RuntimeNodeDetailsV2.tsx`** — Pod Status tab with current status, events, and lifecycle timeline sections. Uses full task path (from `layers`) for nested DAG disambiguation
- **`lib/Apis.ts`** — New API client methods for pod status endpoints

### Backend (`backend/src/v2/compiler/argocompiler/`)
- **`container.go`** — Adds `pipelines.kubeflow.org/task_name` label and `pipelines.kubeflow.org/task_path` annotation to container pod templates
- **`dag.go`** — Adds task path annotations and `task_name` label to DAG driver templates

### Backend (`backend/src/apiserver/`)
- **`model/task.go`** — Adds `PodStatus` and `PodEvents` fields to the Task model (with `PodStatus`, `PodEvent`, and `ContainerStatus` structs)
- **`storage/task_store.go`** — Persists/loads pod status and events; pod events are merged/deduplicated (by type+reason+message+timestamp) instead of replaced, so events accumulate over time

### First round
- Removed unused imports in `app.ts`
- Added `.trim()` to namespace file read in `pod-watcher.ts`
- Removed all debug `console.log` statements from frontend code
- Added `.unref()` and `NODE_ENV=test` guard to cleanup scheduler in `pod-events-cache.ts`
- Wrapped `/k8s/pods/byRunId` and `/k8s/pod/cached` endpoints with authorization
- Merged/deduplicated pod events in `task_store.go` `patchTask` instead of replacing
- Removed internal tracking docs (`PERFORMANCE_OPTIMIZATION_ROLLBACK.md`, `POD_STATUS_CHANGES.md`)

### Second round
- Added pagination (continuation tokens) to `listPodsByRunId` for large ParallelFor runs exceeding 100 pods
- Used full task path (`layers` + leaf key → `root.sub-dag.task-name`) for nested DAG disambiguation via `task_path` annotation
- Dropped `run.proto` schema additions (PodStatus/PodEvent messages) since generated API clients were not regenerated
- Extracted `getPodStatusAndEvents` into dedicated `usePodStatus` hook module with smaller decomposed helpers (`findBestPod`, `buildTaskPath`, `parseContainerStatus`)
- Added tests for `/k8s/pods/byRunId` and `/k8s/pod/cached` endpoints (param validation + auth rejection)
- Fixed K8s client API calls to use object parameter style (matching master branch)
- Fixed ESM import extensions in `pod-watcher.ts`

### Third round
- Fixed `buildTaskPath` to resolve the parent DAG's **componentRef name** from the pipeline spec (matching the compiler's `task_path` annotation format from `dag.go:247-249`)
- Added `layers` to the React Query key to prevent stale cache hits for same-named tasks in different sub-DAGs

## Test plan
- [ ] Verify Pod Status tab renders correctly for running and completed pipeline components
- [ ] Verify pod events are displayed and cached beyond K8s event TTL
- [ ] Verify pod lifecycle timeline shows state transitions including transient error states
- [ ] Verify cache survives frontend pod restart (with PVC)
- [ ] Verify no regressions in existing tabs (Input/Output, Logs, Visualizations)
- [ ] Verify new endpoints reject unauthorized requests (authz tests added)
- [ ] Verify nested DAG tasks show correct pod status (task_path disambiguation)
- [ ] Verify large ParallelFor runs (>100 pods) correctly paginate pod listing
- [ ] Verify API server starts successfully with existing database (AutoMigrate adds new columns)

---

**Checklist:**

### DIFF
--- a/frontend/server/handlers/pod-info.ts
+++ b/frontend/server/handlers/pod-info.ts
@@ -14,6 +14,7 @@
 
 import { Handler } from 'express';
 import * as k8sHelper from '../k8s-helper.js';
+import * as podEventsCache from '../pod-events-cache.js';
 import {
   AuthorizeRequestResources,
   AuthorizeRequestVerb,
@@ -21,55 +22,10 @@ import {
 import { AuthorizeFn } from '../helpers/auth.js';
 
 /**
- * Get pod info handlers.
+ * Get pod info handlers with authorization.
  */
 export function getPodInfoHandlers(authorizeFn: AuthorizeFn) {
-  const podInfoHandler: Handler = async (req, res) => {
-    const { podname, podnamespace } = req.query;
-    if (!podname) {
-      // 422 status code "Unprocessable entity", refer to https://stackoverflow.com/a/42171674
-      res.status(422).send('podname argument is required');
-      return;
-    }
-    if (!podnamespace) {
-      res.status(422).send('podnamespace argument is required');
-      return;
-    }
-
-    // Check access to namespace
-    try {
-      const authError = await authorizeFn(
-        {
-          verb: AuthorizeRequestVerb.GET,
-          resources: AuthorizeRequestResources.VIEWERS,
-          namespace: podnamespace as string,
-        },
-        req,
-      );
-      if (authError) {
-        res.status(403).send('Access denied to namespace');
-        return;
-      }
-    } catch (error) {
-      console.error('Authorization error:', error);
-      res.status(500).send('Authorization check failed');
-      return;
-    }
-
-    const podName = decodeURIComponent(podname as string);
-    const podNamespace = decodeURIComponent(podnamespace as string);
-
-    const [pod, err] = await k8sHelper.getPod(podName, podNamespace);
-    if (err) {
-      const { message, additionalInfo } = err;
-      console.error(message, additionalInfo);
-      res.status(500).send(message);
-      return;
-    }
-    res.status(200).send(JSON.stringify(pod));
-  };
-
-  const podEventsHandler: Handler = async (req, res) => {
+  const authedPodInfoHandler: Handler = async (req, res) => {
     const { podname, podnamespace } = req.query;
     if (!podname) {
       res.status(422).send('podname argument is required');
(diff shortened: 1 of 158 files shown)


frontend/server/k8s-helper.ts
```diff
@@ -347,6 +347,214 @@ export async function listPodEvents(podName: string, podNamespace: string): Prom
...
+      `pipelines.kubeflow.org/run_id=${runId}`,       // Alternative KFP label
+      `workflows.argoproj.io/workflow=${runId}`,      // Argo workflow label (workflow name = run ID)
+    ];
+    console.log('[listPodsByRunId] Will try label selectors:', labelSelectors);
+
+    let allPods: V1Pod[] = [];
+
+    // Performance optimization: limit the number of pods fetched per query
+    // For large for-loop pipelines, this prevents fetching hundreds of pods at once
+    const POD_LIMIT = 100;
```
jeffspahr: `listPodsByRunId` caps each selector query at 100 items and never follows the Kubernetes continuation token. For the large `ParallelFor` runs this feature is explicitly targeting, the requested task pod can easily land beyond the first page, so `/k8s/pods/byRunId` will return an incomplete set and the Pod Status tab can miss the task entirely or choose the wrong pod. This needs pagination, or an early-exit loop that keeps paging until the matching task is found.

frontend/src/components/tabs/RuntimeNodeDetailsV2.tsx
```diff
@@ -301,6 +376,664 @@ function getNodeVolumeMounts(
...
+    const customPropertiesMap = execution.getCustomPropertiesMap();
+    podName = customPropertiesMap.get(KfpExecutionProperties.POD_NAME)?.getStringValue() || '';
+    podNameSpace = customPropertiesMap.get('namespace')?.getStringValue() || namespace || '';
+  }
+
+  // If pod name is not in MLMD, try to find it by run ID label
+  // This handles early failures like ImagePullBackOff where the driver never ran
+  if (!podName && runId && podNameSpace) {
+    try {
+      const taskName = element?.id ? getTaskKeyFromNodeKey(element.id) : undefined;
```
jeffspahr: This lookup only passes `getTaskKeyFromNodeKey(element.id)`, which is the task key local to the current DAG, not the full nested path. That means the new `pipelines.kubeflow.org/task_path` annotation is never used to disambiguate sibling nested tasks that share the same child name, so early-failure lookup can still attach the wrong pod/events to the selected node. This needs to pass a full path derived from `layers` and match against `task_path`, not just the leaf task key.

backend/api/v2beta1/run.proto
```diff
@@ -337,6 +337,84 @@ message PipelineTaskDetail {
 
   // Sequence of dependen tasks.
   repeated ChildTask child_tasks = 16;
+
+  // Pod status information - persisted for UX visibility even after pod termination.
+  PodStatus pod_status = 17;
```
jeffspahr: `run.proto` adds `pod_status` and `pod_events`, but the generated API artifacts in this repo were not regenerated and the converter still only populates fields through `child_tasks`. As a result, the committed Go and frontend client types remain out of sync with the schema and callers of the v2beta1 run API still cannot consume these new fields. Either regenerate/update the generated clients plus `toApiPipelineTaskDetail`, or drop the schema change from this PR.

jeffspahr (changes requested):
Blocking issues:

1. `/k8s/pods/byRunId` only fetches the first 100 pods and never follows pagination, so large runs can miss the selected task pod entirely.
2. The new pod lookup still matches nested tasks by their DAG-local leaf task key, so `task_path` does not actually disambiguate same-named nested tasks.
3. `run.proto` adds `pod_status` / `pod_events`, but the committed generated clients and API converter are still on the old shape, so those fields are not exposed through the checked-in v2beta1 API artifacts.

Verification: `go test ./backend/src/v2/compiler/argocompiler ./backend/src/apiserver/storage ./backend/src/apiserver/server`, `cd frontend && fnm exec --using .nvmrc -- npm ci`, `npm run typecheck`, `npm run test:ui -- src/components/tabs/RuntimeNodeDetailsV2.test.tsx`.

frontend/src/hooks/usePodStatus.ts
```diff
@@ -0,0 +1,363 @@
...
+
+// Build the full task path from layers and element ID for nested DAG disambiguation.
+function buildTaskPath(
+  elementId: string | undefined,
+  layers: string[] | undefined,
+): string | undefined {
+  const leafTaskKey = elementId ? getTaskKeyFromNodeKey(elementId) : undefined;
+  if (!leafTaskKey) return undefined;
+  if (layers && layers.length > 0) {
+    return 'root.' + [...layers, leafTaskKey].join('.');
```
jeffspahr: This still does not produce the same identifier that the compiler writes into `pipelines.kubeflow.org/task_path`. `layers` starts with `root`, so root tasks become `root.root.<task>`, and nested tasks are built from DAG task keys, while the compiler emits root tasks as bare names and nested paths using component-name segments (for example `root.comp-inner-pipeline.print-op1`). Because of that mismatch, the new exact-path lookup never actually disambiguates nested tasks and still falls back to leaf-name matching. The fix needs to derive the lookup path from the same source the compiler uses, or at minimum strip the leading `root` and translate each sub-DAG layer through its `componentRef` name before appending the leaf task.

frontend/src/hooks/usePodStatus.ts
```diff
@@ -0,0 +1,363 @@
...
+// React hook wrapping getPodStatusAndEvents with useQuery.
+export function usePodStatus(
+  execution: Execution | undefined,
+  namespace: string | undefined,
+  runId: string | undefined,
+  element: PipelineFlowElement | null | undefined,
+  layers: string[],
+) {
+  return useQuery<PodStatusResult, Error>({
+    queryKey: ['podStatus', execution?.getId(), runId, element?.id, namespace],
```
jeffspahr: `layers` affects `getPodStatusAndEvents()`, but it is missing from the query key. When `execution` is undefined, different nested nodes can share the same key because `element.id` is only DAG-local like `task.print-op1`. React Query can then reuse pod status/events from another layer without refetching. Include a stable `layers` token in the key, such as `layers.join('/')`.

jeffspahr (changes requested):
Findings:

1. `frontend/src/hooks/usePodStatus.ts:138` still builds a UI-layer path instead of the compiler-emitted `task_path`. `layers` includes `root` and task keys, but the compiler annotates root tasks as bare task names and nested tasks with component-name segments such as `root.comp-inner-pipeline.print-op1`. Exact `task_path` matching therefore still never works for nested tasks and falls back to ambiguous leaf-name matching.

2. `frontend/src/hooks/usePodStatus.ts:358` uses `layers` in the query function but not in the React Query key. In the early-failure case this hook is meant to cover, `execution` is absent, so same-named tasks under different sub-DAG layers can reuse stale pod status/events instead of refetching.

Rechecked fixes on the current head `14c1b19a3`: pagination in `listPodsByRunId` is present now, and the `run.proto` additions were removed. Local verification passed for `go test ./backend/src/v2/compiler/argocompiler`, frontend `npm ci`, `npm run typecheck`, `npm run test:ui -- src/components/tabs/RuntimeNodeDetailsV2.test.tsx`, and server `npm test -- app.test.ts`. CI still has one failing job as of March 28, 2026: `End to End E2ECritical Tests - K8s v1.34.0 cacheEnabled=false argoVersion=v3.7.3 proxy=false pod_to_pod_tls_enabled=false`.

frontend/src/hooks/usePodStatus.ts
NJ3rsey: Fixed. `buildTaskPath` now matches the compiler's `task_path` annotation format from `dag.go:247-249`:
- Root-level tasks (`layers = ["root"]`): returns bare task name (e.g., `print-op1`)
- Nested tasks (`layers = ["root", "inner-pipeline"]`): returns `root.<immediateParentDagName>.<task>` (e.g., `root.inner-pipeline.print-op1`)

The fix uses `layers[layers.length - 1]` as the parent DAG name (skipping the always-present `"root"` first element), which mirrors how the compiler sets `parentDagName` to the immediate parent DAG's task name.

frontend/src/hooks/usePodStatus.ts
NJ3rsey: Fixed. Added `layers.join("/")` to the React Query key so that same-named tasks under different sub-DAG layers each get their own cache entry:

```ts
queryKey: ["podStatus", execution?.getId(), runId, element?.id, namespace, layers.join("/")],
```

frontend/server/k8s-helper.ts
NJ3rsey: Fixed in a previous commit. `listPodsByRunId` now follows K8s continuation tokens in a do/while loop with a `MAX_PAGES=20` safety cap (2000 pods max). See `k8s-helper.ts:383-398`.

frontend/src/components/tabs/RuntimeNodeDetailsV2.tsx
NJ3rsey: Fixed. `buildTaskPath` now resolves the parent DAG's **componentRef name** from the pipeline spec (via `resolveParentDagComponentName`), matching the compiler's `task_path` annotation format exactly:
- Root tasks (`layers=["root"]`): bare task name (e.g., `print-op1`)
- Nested tasks (`layers=["root","inner-pipeline"]`): `root.<componentRefName>.<task>` (e.g., `root.comp-inner-pipeline.print-op1`)

The function walks the pipeline spec's component tree using the layer task keys to find each layer's `componentRef.name`, which is exactly what the compiler uses as `parentDagName` in `dag.go:248-249`.

backend/api/v2beta1/run.proto
NJ3rsey: Fixed in a previous commit. The `pod_status` and `pod_events` fields were removed from `run.proto`. Confirmed via grep — no matches for `pod_status`, `pod_events`, `PodStatus`, or `PodEvent` in `backend/api/v2beta1/run.proto`. Pod status data is fetched directly from the K8s API via the frontend server endpoints.

frontend/src/hooks/usePodStatus.ts
NJ3rsey: Updated fix: `buildTaskPath` now resolves the parent DAG's **componentRef name** from the pipeline spec via a new `resolveParentDagComponentName` helper. This walks the pipeline spec's component tree using the layer task keys to find the correct `componentRef.name`, matching the compiler's format exactly (e.g., task key `inner-pipeline` → componentRef `comp-inner-pipeline` → `root.comp-inner-pipeline.print-op1`).

The `usePodStatus` hook now accepts `pipelineJobString` to perform this resolution. TypeScript, server tests, and Go compiler tests all pass.

frontend/src/hooks/usePodStatus.ts
jeffspahr: [P2] React Query key ignores sub-DAG layers. In the early-failure case this hook is meant to cover, `execution` is absent, so the key collapses to `runId + element.id + namespace`. `element.id` is only the DAG-local `task.<leafKey>`, so selecting same-named tasks in different sub-DAGs can reuse stale pod status/events instead of refetching. Including the active layer path in the key closes that hole.

```suggestion
    queryKey: ['podStatus', execution?.getId(), runId, element?.id, layers.join('.'), namespace],
```

jeffspahr (commented):
Current-head rereview on `733378c5b`: one remaining stale-data issue in `usePodStatus`; suggested fix is inline.

NJ3rsey:
## Recent changes (commit `74b78dc`)

### Fix: task_path matching for nested DAGs

`buildTaskPath` now resolves the parent DAG's **componentRef name** from the pipeline spec (via `resolveParentDagComponentName`), matching the compiler's `task_path` annotation format exactly (`dag.go:247-249`):

- **Root tasks** (`layers=["root"]`): bare task name (e.g., `print-op1`)
- **Nested tasks** (`layers=["root","inner-pipeline"]`): `root.<componentRefName>.<task>` (e.g., `root.comp-inner-pipeline.print-op1`)

The function walks the pipeline spec's component tree using the layer task keys to find each layer's `componentRef.name`, which is exactly what the compiler uses as `parentDagName`.

### Fix: React Query cache key for pod status

Added `layers.join('/')` to the React Query key so that same-named tasks under different sub-DAG layers each get their own cache entry instead of reusing stale data.

### Verification

| Check | Result |
|-------|--------|
| TypeScript (`tsc --noEmit`) | ✅ pass |
| Prettier | ✅ pass |
| Server tests (`app.test.ts`, 40 tests) | ✅ pass |
| Go compiler tests (`argocompiler`) | ✅ pass |
| DCO (Signed-off-by) | ✅ pass |
| E2E Critical (modelcar tests) | ⚠️ [fails on master too](https://github.com/kubeflow/pipelines/actions/runs/23854791021) — infrastructure issue, not related to this PR |

### Files changed

- `frontend/src/hooks/usePodStatus.ts` — new `resolveParentDagComponentName` helper, fixed `buildTaskPath`, added `layers` to query key, accepts `pipelineJobString`
- `frontend/src/components/tabs/RuntimeNodeDetailsV2.tsx` — passes `pipelineJobString` to `usePodStatus`

NJ3rsey:
## Update (Apr 2): rebased and CI status

Branch rebased onto latest `master` (`e7d6aa73d`).

### CI results for `787c7bf`

| Category | Status |
|----------|--------|
| frontend-tests | ✅ pass |
| backend-tests | ✅ pass |
| Workflow Compiler Tests | ✅ pass |
| pre-commit | ✅ pass |
| Build tools images | ✅ pass |
| All image builds | ✅ pass |
| Frontend Integration Tests (all 4 variants) | ✅ pass |
| API Integration v2 (all 3 variants) | ✅ pass |
| API Server tests (all Standalone + K8s Native) | ✅ pass |
| Initialization & Integration v1 | ✅ pass |
| E2E Failure Tests | ✅ pass |
| E2E Proxy Tests | ✅ pass |
| E2E Essential Tests | ✅ pass |
| E2E Critical Tests (3 of ~15 jobs) | ❌ flaky — see below |
| check_ci_status | ❌ gates on E2E Critical |

### E2E Critical failures (all flaky, same on master)

Failed tests on our PR:
- `component_with_optional_inputs.yaml`
- `nested_pipeline_opt_input_child_level_compiled.yaml`
- `flip_coin.yaml`
- `nested_pipeline_opt_inputs_parent_level_compiled.yaml`

[Latest master E2E run](https://github.com/kubeflow/pipelines/actions/runs/23886268285) also fails with:
- `pipeline_with_pod_metadata.yaml` (3×)
- `loop_consume_upstream.yaml` (2×)
- `collected_parameters.yaml` (2×)
- `nested_pipeline_opt_input_child_level_compiled.yaml`
- `nested_pipeline_opt_inputs_parent_level_compiled.yaml`

None of these are related to our changes — they are pre-existing infrastructure/timing flakes in the E2E test suite.

frontend/server/pod-watcher.ts
```diff
@@ -0,0 +1,367 @@
...
+ * Get taskName from pod labels/annotations
+ */
+function getTaskNameFromPod(pod: V1Pod): string | undefined {
+  const labels = pod.metadata?.labels || {};
+  const annotations = pod.metadata?.annotations || {};
+  return (
+    labels['pipelines.kubeflow.org/task_name'] ||
+    labels['component'] ||
+    annotations['pipelines.kubeflow.org/task_path']
+  );
```
jeffspahr, line 146: The cache lookup side now exact-matches the full dotted `task_path`, but the watcher still prefers the short `task_name` label when both are present. That means nested DAG pods are cached under the leaf name here and won't be found after the live pod/events disappear. Flipping the precedence keeps the watcher aligned with the new lookup key.

```suggestion
function getTaskNameFromPod(pod: V1Pod): string | undefined {
  const labels = pod.metadata?.labels || {};
  const annotations = pod.metadata?.annotations || {};
  return (
    annotations['pipelines.kubeflow.org/task_path'] ||
    labels['pipelines.kubeflow.org/task_name'] ||
    labels['component']
  );
}
```

frontend/server/handlers/pod-info.ts
```diff
@@ -100,18 +91,470 @@ export function getPodInfoHandlers(authorizeFn: AuthorizeFn) {
...
+    const cachePromises = pods.map(async (pod) => {
+      const podName = (pod as any).metadata?.name;
+      if (!podName) return;
+
+      const status = extractPodStatus(pod);
+      if (!status) return;
+
+      const labels = (pod as any).metadata?.labels || {};
+      const podTaskName =
+        labels['pipelines.kubeflow.org/task_name'] || labels['component'] || taskName;
```
jeffspahr: The request-time cache path has the same mismatch: when both the short label and full annotation exist, this still stores nested pods under the leaf task name, but `pod-events-cache` later exact-matches the full dotted path. Could we prefer the full annotation here too so the fallback cache path stays consistent with the watcher?

```suggestion
      const labels = (pod as any).metadata?.labels || {};
      const annotations = (pod as any).metadata?.annotations || {};
      const podTaskName =
        annotations['pipelines.kubeflow.org/task_path'] ||
        labels['pipelines.kubeflow.org/task_name'] ||
        labels['component'] ||
        taskName;
```

frontend/server/k8s-helper.ts
```diff
@@ -347,6 +347,217 @@ export async function listPodEvents(podName: string, podNamespace: string): Prom
...
+      const name = pod.metadata?.name || '';
+      return (
+        (name.includes('-impl-') || name.includes('system-container-impl')) &&
+        !name.includes('driver')
+      );
+    };
+
+    // If task name provided, try to filter (but don't require it)
+    let filteredPods = uniquePods;
+    if (taskName && isAllowedResourceName(taskName) && uniquePods.length > 1) {
```
jeffspahr: Now that `taskName` can be the dotted `task_path`, gating the filter on `isAllowedResourceName(taskName)` disables nested-task filtering entirely. `taskName` is only used for local string comparisons in this block, so I would drop that validator and let the exact/`endsWith` checks below do the matching.

```suggestion
    if (taskName && uniquePods.length > 1) {
```

frontend/server/pod-watcher.ts
```diff
@@ -0,0 +1,367 @@
...
+            allPods.push(pod);
+          }
+        }
+      }
+    } catch (fallbackErr) {
+      console.error(`[PodWatcher] Error listing pods in ${namespace}:`, fallbackErr);
+    }
+  }
+
+  return allPods;
```
jeffspahr, line 302: The on-demand `/k8s/pods/byRunId` path already paginates, but the watcher still only warms the first page here. Mirroring the same `_continue` handling keeps large ParallelFor runs from silently dropping cached status/history for pods after the first 100.

```suggestion
async function listRunningPipelinePods(namespace: string): Promise<V1Pod[]> {
  const allPods: V1Pod[] = [];
  const PAGE_SIZE = 100;
  const MAX_PAGES = 20;

  const collectPods = async (fieldSelector?: string): Promise<void> => {
    let continueToken: string | undefined;
    let pages = 0;

    do {
      const result = await k8sV1Client.listNamespacedPod({
        namespace,
        labelSelector: 'pipeline/runid',
        fieldSelector,
        limit: PAGE_SIZE,
        _continue: continueToken,
      });

      if (result.items) {
        for (const pod of result.items) {
          const phase = pod.status?.phase;
          if (phase === 'Running' || (!fieldSelector && phase === 'Pending')) {
            allPods.push(pod);
          }
        }
      }

      continueToken = result.metadata?._continue || undefined;
      pages++;
    } while (continueToken && pages < MAX_PAGES);
  };

  try {
    await collectPods('status.phase=Running');
  } catch (err) {
    try {
      await collectPods();
    } catch (fallbackErr) {
      console.error(`[PodWatcher] Error listing pods in ${namespace}:`, fallbackErr);
    }
  }

  return allPods;
}
```

jeffspahr (changes requested):
Two current-head issues still block this for me:

1. Nested-task pod caches are still written/read under different keys, so the new Pod Status tab can still lose cached history for nested DAG tasks once the live pod or K8s events disappear.
2. The background watcher still stops after the first 100 running pods, so large ParallelFor runs are only partially prewarmed into the cache.

I left inline suggestions below with the minimal changes I think would close both gaps.

frontend/server/pod-watcher.ts
NJ3rsey: Fixed. `getTaskNameFromPod` now prefers `task_path` annotation over `task_name` label, so nested DAG pods are cached under the full dotted path that the lookup side exact-matches.

frontend/server/handlers/pod-info.ts
NJ3rsey: Fixed. The request-time cache path now also prefers `task_path` annotation over `task_name` label, consistent with the watcher.

frontend/server/k8s-helper.ts
NJ3rsey: Fixed. Dropped the `isAllowedResourceName(taskName)` gate — dotted `task_path` values are only used for local string comparisons in the filter block, not as K8s API inputs.

frontend/server/pod-watcher.ts
NJ3rsey: Fixed. `listRunningPipelinePods` now follows continuation tokens with a `MAX_PAGES=20` safety cap, matching the pagination pattern in `listPodsByRunId`.

NJ3rsey:
## Update (Apr 6): fourth round review fixes

Addressed all 4 comments from @jeffspahr in commit `a371d5a`:

1. **task_path precedence in watcher** (`pod-watcher.ts:143`) — `getTaskNameFromPod` now prefers `task_path` annotation over `task_name` label so nested DAG pods are cached under the full dotted path
2. **task_path precedence in handler** (`pod-info.ts:454`) — request-time cache path also prefers `task_path` annotation, consistent with the watcher
3. **Drop `isAllowedResourceName` gate** (`k8s-helper.ts:458`) — dotted `task_path` values are only used for local string comparisons, not K8s API inputs
4. **Watcher pagination** (`pod-watcher.ts:279`) — `listRunningPipelinePods` now follows continuation tokens with `MAX_PAGES=20` safety cap, matching `listPodsByRunId`

Also rebased onto latest `master` (`6250ac6`) and regenerated 5 compiled workflow fixtures that had merge conflicts.

jeffspahr (changes requested):
Current-head rereview on `30b2dc0a9d815172cfb31d04c5933c62e1114dcd`: the earlier `task_path` and watcher pagination issues are fixed, but I still see one blocking regression from the new pod watcher.

The watcher now does namespace-wide pod listing from the `ml-pipeline-ui` server, while the shipped UI RBAC still only grants `get` on `pods` in the standalone and multi-user roles. Current CI already shows the result: `system:serviceaccount:kubeflow:ml-pipeline-ui` gets `403 Forbidden` when it tries to list pods in `kubeflow`.

Please either add the required pod `list` permission everywhere `ml-pipeline-ui` is installed, including the marketplace/chart manifests, or gate/disable the watcher until that permission exists. I left an inline comment on the new list callsite.

frontend/server/pod-watcher.ts
```diff
@@ -0,0 +1,373 @@
...
+  const allPods: V1Pod[] = [];
+  const PAGE_SIZE = 100;
+  const MAX_PAGES = 20;
+
+  const collectPods = async (fieldSelector?: string): Promise<void> => {
+    let continueToken: string | undefined;
+    let pages = 0;
+
+    do {
+      const result = await k8sV1Client.listNamespacedPod({
```
jeffspahr: This watcher now issues namespace-wide `listNamespacedPod` calls from the `ml-pipeline-ui` server, but the shipped UI RBAC still only grants `get` on `pods` in `manifests/kustomize/base/pipeline/ml-pipeline-ui-role.yaml` and `manifests/kustomize/base/installs/multi-user/pipelines-ui/cluster-role.yaml`. Current E2E logs already show the result: `system:serviceaccount:kubeflow:ml-pipeline-ui` cannot list `pods` in `kubeflow` (`403 Forbidden`). Without the matching manifest/chart updates, the background watcher will fail in deployed environments.

jeffspahr (commented):
Non-blocking follow-up on current head `30b2dc0a9d815172cfb31d04c5933c62e1114dcd`: I’d still like direct regression coverage for the two server-side paths that were just fixed here.

Right now the suite exercises them only indirectly. Please add focused tests for:
- watcher pagination in `frontend/server/pod-watcher.ts`, so a multi-page `listNamespacedPod` response proves we keep following `_continue`
- dotted `task_path` matching/cache-key behavior across `frontend/server/pod-watcher.ts`, `frontend/server/handlers/pod-info.ts`, and `frontend/server/k8s-helper.ts`

I’m not treating this as a blocker now that the functional bugs were fixed, but it would materially reduce the risk of these regressions coming back.

NJ3rsey:
@jeffspahr Fixed the RBAC blocker and added the regression tests you requested. Changes in `b2caf8be0`:

**Blocking: pod `list` permission (RBAC)**
- Added `list` verb to pods/pods-log resources in both the standalone Role (`manifests/kustomize/base/pipeline/ml-pipeline-ui-role.yaml`) and the multi-user ClusterRole (`manifests/kustomize/base/installs/multi-user/pipelines-ui/cluster-role.yaml`). The GCP marketplace Helm chart already had this permission.

**Non-blocking: regression tests**
- **`pod-watcher.test.ts`** (10 tests): covers `task_path` annotation precedence over `task_name` label, watcher pagination with continuation tokens, MAX_PAGES safety cap, field selector fallback, and error handling.
- **`k8s-helper.test.ts`** (7 new tests in `listPodsByRunId` suite): covers pagination with continuation tokens, MAX_PAGES cap, dotted `task_path` exact/suffix matching, deduplication across label selectors, and invalid resource name rejection.

Also rebased on upstream/master (3 new commits since last push).

jeffspahr:
Current rereview status on head `57226d0503d563e67d7c5b8eb2dedb3cd2387309`:

No new blocking or medium-severity findings.

The previously raised issues are addressed on this head:
- watcher pagination now follows `_continue`
- dotted `task_path` lookup/cache behavior now has direct regression coverage
- the pod `list` RBAC needed by `ml-pipeline-ui` is now present in the standalone and multi-user kustomize roles, and the marketplace chart already had that permission

Focused verification I ran in an isolated rereview worktree:
- `git diff --check 30b2dc0a9d815172cfb31d04c5933c62e1114dcd..HEAD`
- `go test ./backend/src/v2/compiler/argocompiler`
- `cd frontend/server && fnm exec --using ../.nvmrc -- npm test -- pod-watcher.test.ts k8s-helper.test.ts app.test.ts`

As of April 7, 2026, the PR checks are green aside from the normal pending `tide` status.

jeffspahr:
/lgtm

@NJ3rsey can you resolve conflicts

NJ3rsey:
@jeffspahr Conflicts resolved and rebased on latest master (14 new upstream commits).

frontend/server/pod-watcher.ts
```diff
@@ -0,0 +1,382 @@
...
+/**
+ * Process a single pod: extract status, fetch events, and cache (all async)
+ */
+async function processPod(pod: V1Pod): Promise<boolean> {
+  const podName = pod.metadata?.name;
+  const namespace = pod.metadata?.namespace;
+
+  if (!podName || !namespace) return false;
+
+  if (!hasPodChanged(pod)) return false;
```
jeffspahr: `hasPodChanged(...)` only tracks pod phase/resourceVersion, so this early return also suppresses `fetchPodEvents(...)`. Kubernetes events are separate resources and can keep updating while the pod object stays unchanged, which means repeated `BackOff` / `Pulling` / `Pulled` updates never make it into the cache. To preserve the event history this PR is adding, the watcher needs an event refresh path that is not gated solely on pod object changes.

jeffspahr (commented):
Current-head review on `fafb0f3d66fa05dbb0aff55960c4bdc916a08890`:

I found one medium-severity regression in the new watcher path.

`processPod` bails out before `fetchPodEvents(...)` whenever the pod object's phase/resourceVersion are unchanged, but Kubernetes events are separate resources and can keep evolving while the pod object stays stable. That means repeated `BackOff` / `Pulling` / `Pulled` updates can be missing from the persisted event history that this feature is explicitly trying to preserve past the normal event TTL.

NJ3rsey:
@jeffspahr Addressed the watcher event-refresh regression in commit `90a7b8ebb`.

**Fix in [`frontend/server/pod-watcher.ts`](https://github.com/kubeflow/pipelines/blob/feat/pod-status-tab-master/frontend/server/pod-watcher.ts):** removed the early return on `!hasPodChanged(pod)` before `fetchPodEvents`. `processPod` now always fetches events, diffs them against the cached set, and only skips `savePodInfo` when neither the pod object nor the event set changed. That keeps the disk-I/O bounded while still persisting evolving K8s events (BackOff / Pulling / Pulled) past the cluster's event TTL.

**Regression coverage in [`frontend/server/pod-watcher.test.ts`](https://github.com/kubeflow/pipelines/blob/feat/pod-status-tab-master/frontend/server/pod-watcher.test.ts):**

1. Events are refetched on every tick even when the pod's phase/resourceVersion are unchanged.
2. New events arriving on a stable pod are persisted (with existing events preserved via merge).
3. Save is skipped when both pod and events are unchanged.

**Verification:**

| Check | Result |
|-------|--------|
| `tsc --noEmit` (frontend/server) | pass |
| prettier | pass |
| `pod-watcher.test.ts` | 13/13 pass |
| `app.test.ts` | 40/40 pass |
| `k8s-helper.test.ts` | 40/40 pass |

jeffspahr:
Current rereview status on head `90a7b8ebb0405864b8c89b44804d63961ab450a4`:

The earlier event-refresh gating issue is fixed on this head: `processPod(...)` now refetches pod events even when the pod object's phase/resourceVersion are unchanged.

I still see one medium-severity issue remaining in the updated logic: `hasNewEvents` only keys events by `type/reason/message`, so Kubernetes Event objects that are updated in place by bumping `count` or advancing timestamps still hit the `!podChanged && !hasNewEvents` fast path and never get persisted. That means repeated `BackOff` / `Pulling` / `Pulled` updates can still undercount or lag in the cached history this feature is trying to preserve.

NJ3rsey:
@jeffspahr Fixed the coalesced-event gap in commit `b1ede6b22`.

**Fix:** the change-detection key in `processPod` now includes `count` and `lastTimestamp` in addition to `type/reason/message`, so in-place K8s Event updates (repeated occurrences that bump `count` and advance `lastTimestamp` on the same Event object) are recognized as new and trigger `savePodInfo`. The pod-events-cache merge already keeps the most recent version when keys collide on `type/reason/message`, so the cached entry gets updated with the bumped count/timestamp.

**Test added to [`pod-watcher.test.ts`](https://github.com/kubeflow/pipelines/blob/feat/pod-status-tab-master/frontend/server/pod-watcher.test.ts):** "persists in-place event updates (bumped count / lastTimestamp)" — verifies that when the same event arrives with `count: 3 → 7` and advanced `lastTimestamp`, `savePodInfo` is called and the cached event carries the new values.

**Verification:**

| Check | Result |
|-------|--------|
| `tsc --noEmit` (frontend/server) | pass |
| prettier | pass |
| `pod-watcher.test.ts` | 14/14 pass |

jeffspahr (commented):
No new blocking or medium-severity findings on current head `b1ede6b22f482de0a74a1a3c4065ae44b8a031da`.

This rereview narrowed to the two new watcher follow-ups on top of the previously cleared head:
- `frontend/server/pod-watcher.ts` now refreshes pod events even when the pod object is stable
- the latest watcher follow-up correctly treats in-place K8s event updates as changes by including `count` and `lastTimestamp` in the watcher-side change key
- `frontend/server/pod-watcher.test.ts` adds focused coverage for both stable-pod event refresh and coalesced event updates

Focused verification passed in an isolated rereview worktree:
- `cd frontend/server && fnm exec --using ../.nvmrc -- npm ci --ignore-scripts`
- `cd frontend/server && fnm exec --using ../.nvmrc -- npm test -- pod-watcher.test.ts app.test.ts`
- `cd frontend/server && fnm exec --using ../.nvmrc -- npm run build`

As of April 15, 2026, the PR checks are green aside from the normal pending `tide` status.

jeffspahr:
I only have approval permissions on /frontend so this will need review/approval from someone else like @mprahl or @droctothorpe.

ntny:
This is already a significant improvement over what was there before.

There are still edge cases where a pod is never created due to various issues. For example, if Kyverno is in place, it might block pod creation, or if resource quotas are configured, the requested resources might exceed those limits. In such cases, the pod itself never appears, which means its events are not available either, making these failures harder to observe at the pod level.

At the same time, this information is available at the level of the Argo Workflows resource, specifically in status.nodes.

Because of this, it seems more robust to treat the Argo Workflow status as the source of truth for failures that affect observability, especially in cases where pods are never created (e.g. due to Kyverno policies or resource quota violations). In such scenarios, pod-level events are not available, but the failure is still reflected in status.nodes.
This suggests that argo workflow.status should be the primary source of truth for these types of errors, and this information could be propagated to MLMD executions (or their replacement) and tasks. Extending the persistence agent to capture this would likely be sufficient.

That said, this is a larger architectural change and is out of scope for this MR - just sharing this as a thought for future improvements, not something that should block merging.

HumairAK:
Hi everyone,

Thank you for the significant effort and investment put into this PR so far. We truly value your contribution to Kubeflow Pipelines.

I’m writing to provide some necessary clarification on the path forward for this feature, as there has been some miscommunication regarding the process:

* Process Requirements: Given the size and complexity of these changes, this work needs to undergo the [KEP (Kubeflow Enhancement Proposal) process](https://github.com/kubeflow/pipelines/tree/master/proposals) before implementation to ensure alignment with the project's long-term architecture.

* GSoC Alignment: This effort has significant overlap with a [Google Summer of Code (GSoC) project proposed in February](https://github.com/kubeflow/pipelines/issues/12843). We currently have prospective students volunteered for this specific scope, pending board approval.

We want to make sure your subject matter expertise isn't lost. Since you’ve already done deep dives into this logic, we would love to have you involved in a reviewer/guidance capacity for the GSoC project. This could include:

* Providing technical guidance to the selected student.
* Acting as a core reviewer for the KEP and subsequent PRs.

If you’re interested in collaborating with the student in this way, please reach out to @Alyssa Goins on the CNCF Slack ([#kubeflow-pipelines](https://cloud-native.slack.com/archives/C073N7BMLB1)).

We apologize for the oversight in coordination and any frustration this may cause. We’d love to keep you involved in the design of this feature as it moves through the formal proposal stage

PR #13143, opened 2026-03-26
fix(backend, frontend): Infer FAILED state for pods stuck in CreateContainerConfigError

hbelmiro:
### Problem

When a pipeline task pod hits a fatal startup error like `CreateContainerConfigError` (e.g. missing Secret referenced by `secretAsEnv`), the pod stays in `Pending` phase indefinitely. Argo never transitions the node to `Failed` — it only reports the error in `NodeStatus.Message`. The API server persists the task as `Pending`/`Running`, so the frontend shows the task as still running while the pipeline is stuck.

### Fix

**Backend:** `NodeStatuses()` now infers `Failed` when a Pod-type node is `Pending`/`Running` and its message contains a known fatal substring (`CreateContainerConfigError`, `ImagePullBackOff`, `CrashLoopBackOff`, etc.). `FinishedAt` is normalized to `StartedAt` when unset for inferred failures, and to `0` (instead of `-62135596800`) for unset times generally.

**Frontend:** `updateFlowElementsState` accepts `task_details` from the API and overlays `FAILED` onto graph nodes, matching by MLMD `execution_id` (preferred) or `display_name`. This covers the case where MLMD still reports `RUNNING` but the API already knows the task failed.

**Checklist:**

### DIFF
--- a/backend/src/apiserver/server/api_converter_test.go
+++ b/backend/src/apiserver/server/api_converter_test.go
@@ -28,6 +28,7 @@ import (
 	"github.com/kubeflow/pipelines/backend/src/common/util"
 	"github.com/pkg/errors"
 	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
 	"google.golang.org/genproto/googleapis/rpc/status"
 	"google.golang.org/protobuf/types/known/structpb"
 	"google.golang.org/protobuf/types/known/timestamppb"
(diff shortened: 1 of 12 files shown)


dobbydobap:
This is a nice approach, and inferring `Failed` from the node message covers the startup errors well. One case I wanted to ask about, because I don't think message-matching can reach it.

For `Unschedulable` (say a task with `set_cpu_limit("128")` on a cluster that can't satisfy it), there's no container yet, so there's no `containerStatuses` entry, and as far as I can tell Argo leaves the node `Pending` without writing a message. The signal exists only as a Kubernetes `FailedScheduling` **Event** on the pod. That would put it outside the substring set here, since there's no message to match against.

That seems to line up with #13401, where the reporter says the run sits in `RUNNING` indefinitely and the cause is only visible via `kubectl describe pod`.

Checking `master`, nothing in the backend reads pod events today:

```
$ grep -rn "FailedScheduling\|Unschedulable" backend/src/
(no matches)
```

The only core/v1 Events usage I found is `backend/src/crd/controller/scheduledworkflow/controller.go`, which records its own events rather than watching pods.

So my question is whether `Unschedulable` is intentionally out of scope for this PR, or whether Argo does surface something in `node.Message` for it that I've missed. If it's out of scope that seems entirely reasonable, and it might be worth a line in the description so the gap is explicit, the provisioning category in #12843 would then be half covered here and half needing an Event watch.

Happy to help with the Event-watch side if that would be useful.

PR #13166, opened 2026-03-28
refactor(frontend): reduce NewRun effect syncing

jeffspahr:
## Summary
- derive `specParameters` and the initial pipeline root in `NewRunV2` instead of copying them into local state from a parent effect
- key `NewRunParametersV2` off template plus cloned-runtime-config identity so parameter and pipeline-root state resets with the right defaults when those inputs change
- initialize cloned pipeline-root UI state from props in `NewRunParametersV2` and add rerender/remount regressions for the new reset path

## Why
This is the next `useEffect` cleanup slice after #13140. The remaining hotspot in the New Run flow was the template/runtime-config handoff between `NewRunV2` and `NewRunParametersV2`:

- `NewRunV2` parsed the template and then copied `specParameters` and `pipelineRoot` into local state in an effect
- `NewRunParametersV2` had another effect just to mirror `clonedRuntimeConfig.pipeline_root` into local checkbox/input state

That parent/child sync made the flow harder to reason about and kept the reset behavior spread across effects. This change keeps the existing mount-time parameter propagation, but moves template-derived values to render-time derivation and uses keyed remounts for the child-owned editable state.

I kept this PR narrower than the remaining selector-label sync in `NewRunV2`; that part can still be a follow-up.

## Verification
- `cd frontend && fnm exec --using .nvmrc -- npx prettier@3.8.1 --check src/pages/NewRunV2.tsx src/components/NewRunParametersV2.tsx src/pages/NewRunV2.test.tsx src/components/NewRunParametersV2.test.tsx`
- `cd frontend && fnm exec --using .nvmrc -- npm run test:ui -- src/pages/NewRunV2.test.tsx src/components/NewRunParametersV2.test.tsx`

### DIFF
--- a/frontend/src/pages/NewRunV2.tsx
+++ b/frontend/src/pages/NewRunV2.tsx
@@ -97,13 +97,55 @@ export type SpecParameters = { [key: string]: ComponentInputsSpec_ParameterSpec 
 export type RuntimeParameters = { [key: string]: any };
 type KeyedState<T> = { key: string; value: T };
 
-const hashString = (value: string): string => {
-  let hash = 0;
+const hashString64 = (value: string): string => {
+  let first = 0x9e3779b1;
+  let second = 0x85ebca77;
+
   for (let index = 0; index < value.length; index++) {
-    hash = (hash * 31 + value.charCodeAt(index)) | 0;
+    const charCode = value.charCodeAt(index);
+    first = Math.imul(first ^ charCode, 2654435761);
+    second = Math.imul(second ^ charCode, 1597334677);
   }
-  return (hash >>> 0).toString(36);
+
+  first =
+    Math.imul(first ^ (first >>> 16), 2246822507) ^ Math.imul(second ^ (second >>> 13), 3266489909);
+  second =
+    Math.imul(second ^ (second >>> 16), 2246822507) ^ Math.imul(first ^ (first >>> 13), 3266489909);
+
+  return `${(first >>> 0).toString(16).padStart(8, '0')}${(second >>> 0)
+    .toString(16)
+    .padStart(8, '0')}`;
 };
+
+const getEmptyTemplateData = () => ({
+  defaultPipelineRoot: undefined as string | undefined,
+  specParameters: {} as SpecParameters,
+});
+
+const getTemplateData = (templateString?: string) => {
+  if (!templateString) {
+    return getEmptyTemplateData();
+  }
+
+  try {
+    const spec = convertYamlToV2PipelineSpec(templateString);
+    return {
+      defaultPipelineRoot: spec.defaultPipelineRoot,
+      specParameters: spec.root?.inputDefinitions?.parameters ?? {},
+    };
+  } catch (error) {
+    logger.error('Cannot parse pipeline template.', error);
+    return getEmptyTemplateData();
+  }
+};
+
+function useKeyedState<T>(key: string, initialValue: T) {
+  const [state, setState] = useState<KeyedState<T>>({ key: '', value: initialValue });
+  const value = state.key === key ? state.value : initialValue;
+  const setValue = useCallback((value: T) => setState({ key, value }), [key]);
+
+  return [value, setValue] as const;
+}
 
 type CloneOrigin = {
   isClone: boolean;
@@ -223,18 +265,7 @@ function NewRunV2(props: NewRunV2Props) {
   const [experiment, setExperiment] = useState(chosenExperiment);
   const [experimentName, setExperimentName] = useState('');
   const [serviceAccount, setServiceAccount] = useState('');
-  const [runtimeParametersState, setRuntimeParametersState] = useState<
-    KeyedState<RuntimeParameters>
-  >({ key: '', value: {} });
-  const [pipelineRootState, setPipelineRootState] = useState<KeyedState<string | undefined>>({
-    key: '',
-    value: undefined,
-  });
   const [isStartingNewRun, setIsStartingNewRun] = useState(false);
-  const [isParameterValidState, setIsParameterValidState] = useState<KeyedState<boolean>>({
-    key: '',
-    value: false,
-  });
   const [openNewExperiment, setOpenNewExperiment] = useState(false);
   const [isRecurringRun, setIsRecurringRun] = useState(
     urlParser.get(QUERY_PARAMS.isRecurring) === '1' || cloneOrigin.isRecurring,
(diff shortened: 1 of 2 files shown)


frontend/src/pages/NewRunV2.tsx
```diff
@@ -248,6 +256,49 @@ function NewRunV2(props: NewRunV2Props) {
...
+      defaultPipelineRoot: spec.defaultPipelineRoot,
+      specParameters: spec.root?.inputDefinitions?.parameters ?? {},
+    };
+  }, [templateString]);
+  const parameterStateKey = useMemo(
+    () =>
+      JSON.stringify({
+        clonedRuntimeConfig: clonedRuntimeConfig ?? null,
+        templateString: templateString ?? '',
+      }),
```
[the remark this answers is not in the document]
jeffspahr: Addressed in 050c4cbf3. I replaced the raw `JSON.stringify(...)` key with a compact deterministic hash of the template string and cloned runtime config, so the reset semantics stay stable without carrying the full YAML/config strings through the React `key` and state path. I did not use the random-key suggestion because it would break deterministic rerender/reset behavior. Verification: `npm run test:ui -- src/pages/NewRunV2.test.tsx src/components/NewRunParametersV2.test.tsx` and `npm run typecheck` on the PR branch.

frontend/src/pages/NewRunV2.tsx
```diff
@@ -95,6 +95,15 @@ type NewRunV2Props = RunV2Props & PageProps;
...
 export type RuntimeParameters = { [key: string]: any };
+type KeyedState<T> = { key: string; value: T };
+
+const hashString = (value: string): string => {
+  let hash = 0;
+  for (let index = 0; index < value.length; index++) {
+    hash = (hash * 31 + value.charCodeAt(index)) | 0;
+  }
+  return (hash >>> 0).toString(36);
+};
```
manaswinidas, line 118: **suggestion:** 32-bit hash has realistic collision risk for structurally similar templates (e.g. two versions differing only in a default value). A collision would silently show stale parameters. Since `parameterStateKey` is only compared for equality, raw strings work and are collision-free:

```suggestion
const parameterStateKey = useMemo(
  () => `${templateString ?? ''}::${JSON.stringify(clonedRuntimeConfig ?? null)}`,
  [clonedRuntimeConfig, templateString],
);
```

If size is a concern, `hashString` could be kept but should use a wider window (64-bit or crypto digest).

frontend/src/pages/NewRunV2.tsx
```diff
@@ -248,6 +264,48 @@ function NewRunV2(props: NewRunV2Props) {
...
+        specParameters: {} as SpecParameters,
+      };
+    }
+
+    const spec = convertYamlToV2PipelineSpec(templateString);
+    return {
+      defaultPipelineRoot: spec.defaultPipelineRoot,
+      specParameters: spec.root?.inputDefinitions?.parameters ?? {},
+    };
+  }, [templateString]);
```
manaswinidas: **suggestion:** `convertYamlToV2PipelineSpec` throws on incomplete specs. Inside `useMemo` that crashes the component tree. Pre-existing gap, but since this is being touched — a `try/catch` with empty-state fallback would be resilient:

```typescript
const { defaultPipelineRoot, specParameters } = useMemo(() => {
  if (!templateString) {
    return { defaultPipelineRoot: undefined, specParameters: {} as SpecParameters };
  }
  try {
    const spec = convertYamlToV2PipelineSpec(templateString);
    return {
      defaultPipelineRoot: spec.defaultPipelineRoot,
      specParameters: spec.root?.inputDefinitions?.parameters ?? {},
    };
  } catch {
    return { defaultPipelineRoot: undefined, specParameters: {} as SpecParameters };
  }
}, [templateString]);
```

frontend/src/pages/NewRunV2.tsx
```diff
@@ -248,6 +264,48 @@ function NewRunV2(props: NewRunV2Props) {
...
+      )}`,
+    [clonedRuntimeConfig, templateString],
+  );
+  const initialPipelineRoot = clonedRuntimeConfig?.pipeline_root ?? defaultPipelineRoot;
+  const pipelineRoot =
+    pipelineRootState.key === parameterStateKey ? pipelineRootState.value : initialPipelineRoot;
+  const runtimeParameters =
+    runtimeParametersState.key === parameterStateKey ? runtimeParametersState.value : {};
+  const isParameterValid =
+    isParameterValidState.key === parameterStateKey ? isParameterValidState.value : false;
```
manaswinidas: **nit:** The `KeyedState<T>` + key-check + `useCallback` setter is repeated 3×. A small hook would make the pattern self-documenting:

```typescript
function useKeyedState<T>(key: string, initial: T) {
  const [s, setS] = useState<KeyedState<T>>({ key: '', value: initial });
  const value = s.key === key ? s.value : initial;
  const setValue = useCallback((v: T) => setS({ key, value: v }), [key]);
  return [value, setValue] as const;
}
```

Then these three blocks collapse to:
```typescript
const [pipelineRoot, handlePipelineRootChange] = useKeyedState(parameterStateKey, initialPipelineRoot);
const [runtimeParameters, handleParameterChange] = useKeyedState(parameterStateKey, {} as RuntimeParameters);
const [isParameterValid, handleParameterValidityChange] = useKeyedState(parameterStateKey, false);
```

frontend/src/pages/NewRunV2.tsx
jeffspahr: Addressed in 32aadf3ba. I moved the template parsing into a small helper with a `try/catch` fallback, log the parse failure, and return empty template-derived state instead of crashing the page. I also added a regression test that renders the page with an invalid template string and verifies it falls back to the empty-parameters state.

frontend/src/pages/NewRunV2.tsx
jeffspahr: Addressed in 32aadf3ba. I replaced the repeated keyed-state blocks with a local `useKeyedState` helper, which removes the repeated key check and setter wiring while keeping the same reset semantics.

frontend/src/pages/NewRunV2.tsx
jeffspahr: Addressed in 32aadf3ba. I kept the compact key, but widened it from a single 32-bit hash to a 64-bit-style pair of mixed 32-bit lanes and included the template/config string lengths in the final key. That keeps the reset key small while making silent collisions far less realistic than the earlier version.

PR #13167, opened 2026-03-28
test(frontend): add literal input frontend integration test

### LINKED ISSUE #12603: [feature] KEP-11385: Implement UI drop-down menu for Literal input parameters

jeffspahr:
## Summary
- add a frontend integration fixture for a v2 pipeline with literal runtime inputs
- add a WDIO test that verifies the literal parameter renders as a dropdown, blocks start until a value is selected, and carries the selected value into the created run
- include the new spec in the default frontend integration test suite

This closes the last remaining task called out in #12603.
Closes #12603.

## Verification
- `node --check test/frontend-integration-test/literal-input.spec.js`
- `node --check test/frontend-integration-test/wdio.conf.js`
- `ruby -e "require 'yaml'; YAML.load_file('test/frontend-integration-test/literal-input.yaml')"`

## Notes
- I attempted to run the WDIO flow locally, but the available browser target in this environment served a stale deployed UI image rather than the branch code, so I did not count that as trustworthy branch verification.

### DIFF
--- a/test/frontend-integration-test/literal-input.spec.js
+++ b/test/frontend-integration-test/literal-input.spec.js
@@ -17,6 +17,8 @@ const URL = require('url').URL;
 const {
   buildTableRowSelector,
   clearDefaultInput,
+  saveDebugScreenshot,
+  selectPipelineForRun,
   waitForCondition,
   waitForHashPrefix,
   waitForRunPageReady,
@@ -28,28 +30,8 @@ const selectedLiteral = 'staging';
 const uiTimeout = 10000;
 const runStartTimeout = 30000;
 
-async function selectPipelineForRun() {
-  await $('#choosePipelineBtn').waitForDisplayed({ timeout: uiTimeout });
-  await $('#choosePipelineBtn').click();
-
-  await $('#pipelineSelectorDialog').waitForDisplayed({ timeout: uiTimeout });
-  const pipelineRowSelector = buildTableRowSelector(pipelineName, {
-    containerXPath: '//*[@id="pipelineSelectorDialog"]',
-  });
-
-  await waitForCondition(
-    async () => (await $(pipelineRowSelector).isExisting()),
-    {
-      timeout: uiTimeout,
-      timeoutMsg: `expected pipeline row for ${pipelineName} to appear`,
-    },
-  );
-
-  await $(pipelineRowSelector).click();
-  await $('#usePipelineBtn').waitForEnabled({ timeout: uiTimeout });
-  await $('#usePipelineBtn').click();
-  await $('#pipelineSelectorDialog').waitForDisplayed({ timeout: uiTimeout, reverse: true });
-}
+let createdRunId = '';
+let pipelineUploaded = false;
 
 async function waitForCreatedRunId() {
   const currentHash = new URL(await browser.getUrl()).hash;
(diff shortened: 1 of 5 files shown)


Sridhar1030 (commented):
The added integration test covers the core enum-input flow well. Would it also make sense to add a case for a parameter with a default enum value, or for multiple literal parameters, to verify the UI behavior beyond the single-parameter scenario?

jeffspahr:
@Sridhar1030 I'm going to keep this one tightly scoped, but feel free to open an issue/pr for other improved testing.

test/frontend-integration-test/literal-input.spec.js
```diff
@@ -0,0 +1,234 @@
...
+      timeout: uiTimeout,
+      timeoutMsg: `expected pipeline row for ${pipelineName} to appear`,
+    },
+  );
+
+  await $(pipelineRowSelector).click();
+  await $('#usePipelineBtn').waitForEnabled({ timeout: uiTimeout });
+  await $('#usePipelineBtn').click();
+  await $('#pipelineSelectorDialog').waitForDisplayed({ timeout: uiTimeout, reverse: true });
+}
```
manaswinidas: DRY: `selectPipelineForRun()` is copy-pasted near-identically in `helloworld.spec.js` and `tensorboard-example.spec.js`. Extract to `test-helpers.js`, parameterized by `pipelineName`. The sibling copies also include diagnostic `try/catch` + `saveDebugScreenshot` that this copy omits — a shared version should include those.

test/frontend-integration-test/literal-input.spec.js
```diff
@@ -0,0 +1,234 @@
...
+
+const assert = require('assert');
+const URL = require('url').URL;
+const {
+  buildTableRowSelector,
+  clearDefaultInput,
+  waitForCondition,
+  waitForHashPrefix,
+  waitForRunPageReady,
+} = require('./test-helpers');
```
manaswinidas, line 25: `saveDebugScreenshot` is not imported. Both sibling specs capture screenshots on failure for CI debugging — without it, failures here will produce no diagnostic artifacts.

test/frontend-integration-test/literal-input.spec.js
```diff
@@ -0,0 +1,234 @@
...
+    return {
+      ok: response.ok,
+      run: responseBody,
+      status: response.status,
+    };
+  }, runId);
+}
+
+describe('literal input parameter integration', () => {
+  before(async () => {
```
manaswinidas, line 159: No `after` hook for cleanup. Both sibling specs clean up (helloworld deletes the pipeline; tensorboard uses `after`). This leaves the uploaded pipeline and created run behind, risking environment pollution over time.

test/frontend-integration-test/literal-input.spec.js
```diff
@@ -0,0 +1,234 @@
...
+
+    await $('#createNewRunBtn').waitForDisplayed({ timeout: uiTimeout });
+    await $('#createNewRunBtn').click();
+    await waitForHashPrefix('#/runs/new', { timeout: uiTimeout });
+
+    await selectPipelineForRun();
+    await waitForRunPageReady({ timeout: runStartTimeout });
+  });
+
+  it('renders the literal parameter as a dropdown and requires a selection before start', async () => {
```
manaswinidas, line 196: Redundant call: `waitForRunPageReady` was already called at line 131 in the previous `it` block. Remove one.

test/frontend-integration-test/literal-input.spec.js
```diff
@@ -0,0 +1,234 @@
...
+
+    await $(selectors.runName).click();
+    await clearDefaultInput();
+    await browser.keys(runName);
+
+    const literalSelect = await $('//*[@role="combobox" and @id="environment"]');
+    await literalSelect.waitForDisplayed({ timeout: uiTimeout });
+    const literalNativeInput = await literalSelect.$(
+      './following-sibling::input[contains(@class, "MuiSelect-nativeInput")]',
+    );
```
manaswinidas: Nit: `MuiSelect-nativeInput` is an MUI internal class. If MUI changes its Select rendering this breaks silently. Prefer a `data-testid` on the component side.

test/frontend-integration-test/literal-input.spec.js
```diff
@@ -0,0 +1,234 @@
...
+      async () => {
+        fetchedRun = await fetchRunById(runId);
+        if (!fetchedRun.ok) {
+          return false;
+        }
+        return fetchedRun.run.runtime_config?.parameters?.environment === selectedLiteral;
+      },
+      {
+        timeout: runStartTimeout,
+        interval: 1000,
```
manaswinidas, line 279: Tautological: if `waitForCondition` resolved without throwing, the equality `=== selectedLiteral` is already guaranteed. This `assert.equal` can never fail. Either drop it, or restructure so `waitForCondition` waits only for `fetchedRun.ok` and the assert checks the parameter value separately.

test/frontend-integration-test/literal-input.yaml
```diff
@@ -0,0 +1,105 @@
+# Copyright 2018 The Kubeflow Authors
```
manaswinidas: Nit: copyright year should be `2026` for a newly created file.

test/frontend-integration-test/literal-input.spec.js
jeffspahr: Moved selectPipelineForRun into test-helpers.js and switched helloworld.spec.js, tensorboard-example.spec.js, and literal-input.spec.js over to the shared helper. The shared path now keeps the screenshot capture plus selector diagnostics together.

test/frontend-integration-test/literal-input.spec.js
jeffspahr: Added cleanup diagnostics in literal-input.spec.js, and the pipeline-selector screenshot path now comes from the shared selectPipelineForRun helper in test-helpers.js, so the missing screenshot coverage is closed.

test/frontend-integration-test/literal-input.spec.js
jeffspahr: Added an after hook that best-effort deletes both the created run via the v2beta1 API and the uploaded pipeline via the UI. The cleanup path also captures screenshots if either delete step fails.

test/frontend-integration-test/literal-input.spec.js
jeffspahr: Removed the earlier waitForRunPageReady call from the route-navigation step. The form-ready wait now happens only in the dropdown assertion test.

test/frontend-integration-test/literal-input.spec.js
jeffspahr: Removed the MuiSelect-nativeInput dependency entirely. The test now asserts the visible placeholder text and the selected option text instead of reaching into an MUI private class.

test/frontend-integration-test/literal-input.spec.js
jeffspahr: Restructured the run polling so waitForCondition stops once the run fetch succeeds and exposes runtime parameters, then the selected literal is asserted separately afterward.

test/frontend-integration-test/literal-input.yaml
jeffspahr: Updated the new literal-input.yaml fixture header to use the 2026 copyright year.

jeffspahr:
@manaswinidas pushed 077f9fe25 to address the current review threads.

This follow-up:
- moves selectPipelineForRun into test-helpers.js and switches the helloworld, tensorboard, and literal-input specs to the shared helper,
- adds literal-input cleanup for both the created run and uploaded pipeline,
- removes the redundant run-page wait,
- replaces the MUI internal-class lookup with visible placeholder/selection assertions, and
- reworks the run API polling so the final literal-value assert is no longer tautological.

Local verification:
- node --check test/frontend-integration-test/test-helpers.js
- node --check test/frontend-integration-test/helloworld.spec.js
- node --check test/frontend-integration-test/tensorboard-example.spec.js
- node --check test/frontend-integration-test/literal-input.spec.js
- ruby -e "require 'yaml'; YAML.load_file('test/frontend-integration-test/literal-input.yaml')"

manaswinidas (approved):
/lgtm

PR #13171, opened 2026-03-29
Central driver implementation

### LINKED ISSUE #12642: feat(backend): Reduce KFP Driver pod footprint
Driver Pod foot print is way too high. Let's use Argo Executor Plugin feature to reduce this down to one pod per pipeline. 

Proposal: https://github.com/kubeflow/pipelines/pull/12269
Implementation: https://github.com/kubeflow/pipelines/pull/12269

ntny:
**backup**: save [central-driver-poc](https://github.com/kubeflow/pipelines/pull/12269) branch before merge
This branch is a backup of [central-driver-poc ](https://github.com/kubeflow/pipelines/pull/12269) before merging master.
The merge caused too many conflicts for a clean rebase, so the diffs
were extracted into this backup branch to preserve all work safely.
All discussions and comments remain in the original [branch](https://github.com/kubeflow/pipelines/pull/12269).

POC for https://github.com/kubeflow/pipelines/pull/12023

Resolves: https://github.com/kubeflow/pipelines/issues/12642

feat(backend): introduce centralized driver via Argo executor plugin and eliminate per-component driver pods

Previously, each driver component was executed as a separate Kubernetes pod,
leading to significant overhead in pod scheduling, resource consumption,
and control plane load, especially for pipelines with many small components.

This change replaces the pod-per-driver model with a centralized driver invoked via an Argo Workflow executor plugin using a sidecar-based execution model.

A new driver service is introduced and exposed via an HTTP API, allowing Argo to delegate driver execution without spawning dedicated pods.

Key changes:
- Remove pod-per-driver execution model
- Introduce centralized driver implemented as an Argo executor plugin
- Add driver-plugin sidecar configuration via Kubernetes ConfigMaps
- Add new driver implementation (backend/src/driver) with HTTP API
- Add required Kubernetes manifests, including RBAC roles and bindings, for the plugin
- Integrate plugin configuration into all kustomize deployment variants
- Update CI, TLS handling, and debugging scripts to support plugin-based execution
- Improve logging with context-aware driver logger
- Persist driver logs to S3 and expose them in the UI

Benefits:
- Eliminates driver pod churn and reduces scheduling overhead
- Reduces pressure on Kubernetes control plane (API server, scheduler, etcd)
- Improves execution latency
- Enables more efficient and scalable pipeline execution
- Provides better observability: driver logs are now available in the UI and persistently stored in S3

This change introduces a new execution model for driver components and lays the groundwork for further optimizations in pipeline runtime.

**Checklist:**


manifests/kustomize/base/installs/multi-user/pipelines-profile-controller/pipelines-profile-controller-admin.yaml
```diff
@@ -0,0 +1,42 @@
...
+    verbs: ["create", "get", "list", "watch", "update", "delete"]
+  - apiGroups: ["rbac.authorization.k8s.io"]
+    resources: ["roles", "rolebindings"]
+    verbs: ["create", "get", "list", "watch", "update", "delete"]
+  - apiGroups: [ "" ]
+    resources: [ "configmaps" ]
+    verbs: [ "get", "list", "watch" ]
+  - apiGroups: [ "" ]
+    resources: [ "secrets" ]
+    verbs: [ "get", "list", "watch" ]
```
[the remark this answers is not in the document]
ntny: The profile-controller creates RBAC permissions to read Secrets in all namespaces for the ServiceAccount under which the executor plugin runs, and this is required for its operation.  
The driver uses S3 credentials from the artifact S3 Secret to upload driver logs.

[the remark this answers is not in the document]
droctothorpe: Does the secret used by the driver to upload logs have a consistent name? If so, can you add a `resourceNames` restriction like copilot suggests?

manifests/kustomize/base/installs/multi-user/pipelines-profile-controller/deployment.yaml
```diff
@@ -27,7 +28,14 @@ spec:
...
         image: docker.io/alpine/k8s:1.32.3
-        command: ["python", "/hooks/sync.py"]
+        command:
+          - sh
+          - -c
+          - |
+            python3 -m venv /tmp/venv && \
+            . /tmp/venv/bin/activate && \
+            pip install --no-cache-dir kubernetes boto3 && \
+            python /hooks/sync.py
```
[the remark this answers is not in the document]
droctothorpe: Hopefully addressed by https://github.com/ntny/pipelines/pull/2

manifests/kustomize/base/installs/multi-user/pipelines-profile-controller/sync.py
```diff
@@ -173,6 +175,81 @@ def upsert_lifecycle_policy(self, bucket_name, artifact_retention_days):
...
+                ),
+                rules=[
+                    client.V1PolicyRule(
+                        api_groups=[""],
+                        resources=resources,
+                        resource_names=resource_names,
+                        verbs=verbs
+                    )
+                ]
+            )
```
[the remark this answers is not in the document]
droctothorpe: Should be addressed by https://github.com/ntny/pipelines/pull/2.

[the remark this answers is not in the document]
ntny: upsert_executor_plugin_sa is declared as a method of the Controller class, which itself is nested inside server_factory.

So the nesting is:

server_factory -> Controller -> upsert_executor_plugin_sa

The sync method is declared at the same class level:

server_factory -> Controller -> sync

Because both sync and upsert_executor_plugin_sa are methods of the same Controller class, the call

self.upsert_executor_plugin_sa(namespace)

works correctly.

If upsert_executor_plugin_sa were actually nested inside another method, then self would indeed not have that attribute and we would get an AttributeError, but that is not the case here.

I suspect this may not be obvious from the diff view, especially since Controller itself is defined inside server_factory.

I also tested profile_controller locally, and all multi-user tests fail if this method does not execute, so the current implementation is definitely being called as expected.

manifests/kustomize/base/installs/multi-user/pipelines-profile-controller/pipelines-profile-controller-admin.yaml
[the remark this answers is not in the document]
ntny: done

manifests/kustomize/base/installs/multi-user/pipelines-profile-controller/sync.py
juliusvonkohout, line 1: Hello, I am really concerned about the sync.py changes. Why do you want to create new serviceaccounts? Pipelines usually run as default-editor in Kubeflow. Usually such roles should be aggregated to kubeflow-edit. Adding attidional service accounts i really want to avoid. That will probably conflict with the average multi-user Kubeflow and create technical debt.

juliusvonkohout (changes requested):
Hello, I am really concerned about the sync.py changes. Why do you want to create new serviceaccounts? Pipelines usually run as default-editor in Kubeflow. Usually such roles should be aggregated to kubeflow-edit. Adding attidional service accounts i really want to avoid. That will probably conflict with the average multi-user Kubeflow and create technical debt.

manifests/kustomize/base/installs/multi-user/pipelines-profile-controller/sync.py
ntny: Hello

That is a reasonable concern.

The additional service account is mandatory for the [Argo Workflow Executor Plugin](https://argo-workflows.readthedocs.io/en/latest/executor_plugins/).

The central driver implementation is based on the Argo Workflow Executor Plugin. For safety reasons, the executor plugin uses a separate service account with the [static name](https://github.com/argoproj/argo-workflows/blob/main/workflow/controller/agent.go#L284) that must be created specifically for the plugin, and it does not use the Argo Workflow service account.

Details are [here](https://github.com/kubeflow/pipelines/blob/master/proposals/separate-standalone-driver/plugin.md); please check prerequisite point 6.

It is compatible with the multi-user environment. I've checked it manually, and existed multi-user E2E tests cover this case.

test_data/kubernetes/seaweedfs/kubeflow-edit-clusterrole.yaml
juliusvonkohout, line 1: i saw this patch in multiple places. Can we consolidate ?

manifests/kustomize/base/installs/multi-user/pipelines-profile-controller/deployment.yaml
```diff
@@ -9,6 +9,7 @@ spec:
       labels:
         sidecar.istio.io/inject: "false"
     spec:
+      serviceAccountName: pipelines-profile-controller-admin
```
juliusvonkohout:
```suggestion

```
It does not work like this. Metacontroller is getting the results and works on it. Please do not do create any Kubernetes reources in sync.py.

manifests/kustomize/base/installs/multi-user/pipelines-profile-controller/pipelines-profile-controller-admin.yaml
juliusvonkohout, line 1: That looks very dangerous. Please only whitelist specific resource names and i am wondering why you are adding this file. You need to modify the metacontroller serviceaccount instead.

manifests/kustomize/base/installs/multi-user/pipelines-profile-controller/deployment.yaml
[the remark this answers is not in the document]
juliusvonkohout: The metacontroller and pipelines profile controller is different. PPC only provides a json to metacontroller and metacontroller handles the rest.

manifests/kustomize/base/installs/multi-user/pipelines-profile-controller/kustomization.yaml
```diff
@@ -5,6 +5,7 @@ resources:
 - service.yaml
 - deployment.yaml
 - decorator-controller.yaml
+- pipelines-profile-controller-admin.yaml
```
juliusvonkohout:
```suggestion
```

juliusvonkohout: Please do not add this file

juliusvonkohout: See https://github.com/kubeflow/pipelines/blob/master/manifests/kustomize/third-party/metacontroller/base/cluster-role.yaml

manifests/kustomize/base/installs/multi-user/pipelines-profile-controller/pipelines-profile-controller-admin.yaml
juliusvonkohout: see https://github.com/kubeflow/pipelines/blob/master/manifests/kustomize/third-party/metacontroller/base/cluster-role.yaml

manifests/kustomize/base/installs/multi-user/pipelines-profile-controller/deployment.yaml
juliusvonkohout: see see https://github.com/kubeflow/pipelines/blob/master/manifests/kustomize/third-party/metacontroller/base/cluster-role.yaml

manifests/kustomize/base/installs/multi-user/pipelines-profile-controller/deployment.yaml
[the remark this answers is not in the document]
juliusvonkohout: see see https://github.com/kubeflow/pipelines/blob/master/manifests/kustomize/third-party/metacontroller/base/cluster-role.yaml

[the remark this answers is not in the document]
juliusvonkohout: and https://github.com/kubeflow/pipelines/blob/master/manifests/kustomize/base/installs/multi-user/pipelines-profile-controller/decorator-controller.yaml

juliusvonkohout (changes requested):
I added a few comments. The PipelineProfileController+Metacontroller architecture is different than what you expect your code does not match it.

See https://github.com/kubeflow/pipelines/blob/master/manifests/kustomize/base/installs/multi-user/pipelines-profile-controller/decorator-controller.yaml and https://github.com/kubeflow/pipelines/blob/master/manifests/kustomize/third-party/metacontroller/base/cluster-role.yaml. The pipelines profile controller just provides the input for metacontroller.

You also patch the permissions for "workflowtasksets/status" in many files which adds a lot of code duplication.

The PSS restricted stuff looks good.

manifests/kustomize/base/installs/multi-user/pipelines-profile-controller/kustomization.yaml
ntny: Hi @juliusvonkohout , thanks for the clarification  this really helps me wrap the namespace reconciliation flow in my head.

I [updated](https://github.com/kubeflow/pipelines/pull/13171/commits/e3c655cc13934c14953864b0c9e4e9f2bb565637) the change accordingly:

- move the additional RBAC roles/grants to the metacontroller
- drop it from the profile controller
- remove direct Kubernetes resource creation from sync.py
- add the additional resources to watch in the decorator controller
- remove the redundant workflowtasksets/status patch, as it is not actually used
Thanks again for pointing me to the right place.

One follow-up question: I noticed that sync.py currently also updates the S3 bucket lifecycle policy via upsert_lifecycle_policy(), which creates/deletes state outside Kubernetes
I assume this is acceptable here since it creates state outside Kubernetes rather than Kubernetes resources, so I guess this is also somewhat of a workaround
In fact, this method was what initially led me to the idea of adding the resource creation logic here as well.

manifests/kustomize/base/installs/multi-user/pipelines-profile-controller/sync.py
```diff
@@ -173,6 +173,185 @@ def upsert_lifecycle_policy(self, bucket_name, artifact_retention_days):
                 else:
                     print(f"ERROR: Failed to configure lifecycle policy: {exception}")
 
+        def executor_plugin_resources(self, namespace):
+            # Argo Workflow Executor Plugin Necessary Resources
+            agent_sa_name = 'ml-pipeline-driver-agent-executor-plugin'
```
juliusvonkohout:
```suggestion
            agent_service_account_name = 'ml-pipeline-driver-agent-executor-plugin'
```

manifests/kustomize/base/installs/multi-user/pipelines-profile-controller/sync.py
```diff
@@ -173,6 +173,185 @@ def upsert_lifecycle_policy(self, bucket_name, artifact_retention_days):
...
+                        'labels': {
+                            'application-crd-id': 'kubeflow-pipelines'
+                        }
+                    },
+                    'secrets': [{
+                        'name': secret_name
+                    }]
+                },
+                {
+                    'apiVersion':
```
juliusvonkohout: Please create this as cluster role and bind via rolebinding. That is how we also do it for the Default-Editor etc. And prevent duplication.

manifests/kustomize/base/installs/multi-user/pipelines-profile-controller/sync.py
```diff
@@ -173,6 +173,185 @@ def upsert_lifecycle_policy(self, bucket_name, artifact_retention_days):
...
+                        'apiGroup': 'rbac.authorization.k8s.io',
+                    }
+                },
+                {
+                    'apiVersion':
+                        'rbac.authorization.k8s.io/v1',
+                    'kind':
+                        'Role',
+                    'metadata': {
+                        'name': 'ml-pipeline-driver-pods-reader',
```
juliusvonkohout: Please use a cluster role to avoid duplication

manifests/kustomize/base/installs/multi-user/pipelines-profile-controller/sync.py
```diff
@@ -173,6 +173,185 @@ def upsert_lifecycle_policy(self, bucket_name, artifact_retention_days):
...
+                    },
+                    'rules': [{
+                        'apiGroups': [''],
+                        'resources': ['pods'],
+                        'verbs': ['get', 'list', 'watch'],
+                    }]
+                },
+                {
+                    'apiVersion': 'rbac.authorization.k8s.io/v1',
+                    'kind': 'RoleBinding',
```
juliusvonkohout, line 278: Why do you need so many different rolebindings? One should be enough. Use an aggregated cluster role instead.

manifests/kustomize/base/installs/multi-user/pipelines-profile-controller/sync.py
```diff
@@ -173,6 +173,185 @@ def upsert_lifecycle_policy(self, bucket_name, artifact_retention_days):
...
+                        'apiGroups': [''],
+                        'resources': ['persistentvolumeclaims'],
+                        'verbs': ['create', 'get', 'list'],
+                    }]
+                },
+                {
+                    'apiVersion': 'rbac.authorization.k8s.io/v1',
+                    'kind': 'RoleBinding',
+                    'metadata': {
+                        'name': 'ml-pipeline-driver-pvc-editor-binding',
```
juliusvonkohout: /hold the PR, that is not well maintainable

juliusvonkohout:
/hold

For the changes in sync.py. the architecture does not look right and uses multiple rolebindings for a single service account and clutters the namespaces.

manifests/kustomize/base/installs/multi-user/pipelines-profile-controller/sync.py
```diff
@@ -173,6 +173,185 @@ def upsert_lifecycle_policy(self, bucket_name, artifact_retention_days):
...
+                        'namespace': namespace,
+                    }],
+                    'roleRef': {
+                        'kind': 'Role',
+                        'name': 'ml-pipeline-driver-pvc-editor',
+                        'apiGroup': 'rbac.authorization.k8s.io',
+                    }
+                },
+                {
+                    'apiVersion':
```
juliusvonkohout: Please use a single clusterrole

droctothorpe:
@juliusvonkohout thank you so much for the feedback! We're just waiting on a final lgtm from you and we will merge. It's hard to overstate the impact of this PR. The number of pods being halved has massive performance, scaling, and execution time implications.

juliusvonkohout:
> @juliusvonkohout thank you so much for the feedback! We're just waiting on a final lgtm from you and we will merge. It's hard to overstate the impact of this PR. The number of pods being halved has massive performance, scaling, and execution time implications.

Yes, I will try to have a call next week, because it's important to merge but we also must be careful regarding the added complexity and maintenance.

backend/src/driver/rpc_handler.go
```diff
@@ -0,0 +1,469 @@
...
+		CaCertPath:              args.CACertPath,
+	}
+
+	var driverErr error
+	switch args.Type {
+	case RootDag:
+		options.RuntimeConfig = runtimeConfig
+		execution, pipeline, driverErr = driver.RootDAG(ctx, options, client)
+		if driverErr != nil {
+			return nil, err
```
nsingla: **[CRITICAL]** Bug: `err` is nil at this point (all prior error checks passed), so this silently swallows RootDAG failures — the caller proceeds as if execution succeeded, leading to silent pipeline failures.

Fix: Change `return nil, err` to `return nil, driverErr`

backend/src/driver/rpc_handler.go
```diff
@@ -0,0 +1,469 @@
...
+			}
+		}
+	}()
+
+	log := util.GetLoggerFrom(ctx)
+
+	log.Infof("driver plugin arguments: %v", args)
+	// Support reading component spec from a file if value starts with @
+	// This bypasses exec() argument size limits for large workflows
+	if strings.HasPrefix(args.Component, "@") {
```
nsingla: **[SUGGESTION]** No path validation on the `@` prefix file read. Consider allowlisting paths (e.g., only `/tmp/` or a known component spec directory) to mitigate path traversal risk if plugin args are ever user-influenced.

backend/src/v2/compiler/argocompiler/dag.go
```diff
@@ -635,22 +638,12 @@ func (c *workflowCompiler) addDAGDriverTemplate() string {
...
 			},
 		},
-		Container: &k8score.Container{
-			Image:     c.driverImage,
-			Command:   c.driverCommand,
-			Args:      args,
-			Resources: driverResources,
-			Env:       proxy.GetConfig().GetEnvVars(),
-		},
+		Plugin: dagPlugin,
```
nsingla, line 573: **[IMPORTANT]** The DAG driver template now uses `Plugin` instead of `Container`, but the outputs still use file-based `Path` (`/tmp/outputs/execution-id`, etc.) instead of `JSONPath` like the container driver template (`$.pod-spec-patch`, `$.cached-decision`). Can you confirm Argo supports `Path`-based output extraction from executor plugins? If not, DAG execution IDs won't propagate correctly.

manifests/kustomize/base/pipeline/pipeline-runner-role.yaml
```diff
@@ -85,3 +89,54 @@ rules:
...
+  - apiGroups: [""]
+    resources: ["secrets"]
+    resourceNames: ["mlpipeline-minio-artifact"]
+    verbs: ["get"]
+---
+apiVersion: rbac.authorization.k8s.io/v1
+kind: RoleBinding
+metadata:
+  name: ml-pipeline-driver-agent-executor-plugin-binding
+  namespace: kubeflow
```
nsingla: **[IMPORTANT]** `namespace: kubeflow` is hardcoded in both the RoleBinding metadata and subjects. This will break in non-`kubeflow` namespace deployments and multi-user setups. Consider using kustomize `namespace` transformer or removing the hardcoded value.

manifests/kustomize/base/installs/multi-user/pipelines-profile-controller/sync.py
```diff
@@ -370,6 +444,37 @@ def sync(self, parent, attachments):
                     },
                 ])
 
+            print('Creating executor-plugin service accounts')
+            # Argo Workflow Executor Plugin Necessary Resources
```
nsingla: **[IMPORTANT]** These two Secrets (`default-editor.service-account-token` and `ml-pipeline-driver-agent-executor-plugin.service-account-token`) are already created by `executor_plugin_resources()` called at line 329. This results in duplicate Secret definitions in the desired state, which may cause Metacontroller reconciliation issues.

manifests/kustomize/base/installs/multi-user/pipelines-profile-controller/sync.py
```diff
@@ -173,6 +173,70 @@ def upsert_lifecycle_policy(self, bucket_name, artifact_retention_days):
                 else:
                     print(f"ERROR: Failed to configure lifecycle policy: {exception}")
 
+        def executor_plugin_resources(self, namespace):
```
nsingla, line 256: **[IMPORTANT]** +1 to @juliusvonkohout's feedback — creating a separate `ml-pipeline-driver-agent-executor-plugin` ServiceAccount per namespace deviates from the Kubeflow convention where workloads run as `default-editor` with permissions aggregated via `kubeflow-edit`. The executor plugin permissions could instead be added to the existing `aggregate-to-kubeflow-pipelines-edit` ClusterRole (in `view-edit-cluster-roles.yaml`) with the label `rbac.authorization.kubeflow.org/aggregate-to-kubeflow-pipelines-edit: "true"`, keeping the existing RBAC model intact and avoiding per-namespace SA proliferation.

nsingla (commented):
### PR Review Summary

**Scope**: Key files reviewed across 5 functional groups (driver plugin, compiler, RBAC/manifests, driver logic, frontend/CI)

#### Impact Assessment
- **Overall Risk**: HIGH
- **Breaking Changes**: Yes — new execution model, RBAC changes, sync.py SA creation
- **Cross-Module Impact**: Compiler, driver, manifests, frontend, CI all affected

#### Code Quality
- **Critical Issues**: 1 found (RootDAG error swallowed)
- **Important Issues**: 4 found (DAG output mechanism, hardcoded namespace, duplicate secrets, SA pattern)
- **Suggestions**: 1 (path traversal mitigation)

#### Inline Comments
6 inline comments posted across 4 files.

#### Recommendation
REQUEST_CHANGES — The RootDAG error-swallowing bug is a correctness blocker. The DAG/Container output mechanism inconsistency needs verification. The sync.py concerns align with the existing reviewer holds.

manifests/kustomize/base/installs/multi-user/pipelines-profile-controller/sync.py
ntny: Hello @nsingla,
Commented here: https://github.com/kubeflow/pipelines/pull/13171#discussion_r3014802581

These are unfortunately the requirements for using the Argo Workflow agent with RBAC roles necessary for the driver (e.g., to create a PVC). 
I had also considered allowing the use of a shared single  Service Account in Argo Workflows, but the community raised security concerns, and I agreed with them.
https://github.com/argoproj/argo-workflows/pull/14711#issuecomment-3155502204

backend/src/driver/rpc_handler.go
ntny: thanks! Done

manifests/kustomize/base/pipeline/pipeline-runner-role.yaml
ntny: The pipeline-runner is a special account used only for standalone mode, though I agree with the point about removing the hardcoding.
done

manifests/kustomize/base/installs/multi-user/pipelines-profile-controller/sync.py
juliusvonkohout: given that hardcoded stuff here https://github.com/argoproj/argo-workflows/blob/main/workflow/controller/agent.go#L284 it is really ugly but necessary.

juliusvonkohout: ml-pipeline-driver-agent is the plugin name and the service account in argo is hardcoded to add the suffix -executor-plugin

manifests/kustomize/base/installs/multi-user/pipelines-profile-controller/sync.py
```diff
@@ -369,6 +429,20 @@ def sync(self, parent, attachments):
...
+                    "metadata": {
+                        "name": "default-editor.service-account-token",
+                        "namespace": namespace,
+                        "annotations": {
+                            "kubernetes.io/service-account.name": "default-editor"
+                        }
+                    },
+                    "type": "kubernetes.io/service-account-token"
+                },
+            ])
```
juliusvonkohout, line 445: Please remove this. Kubernetes does not create secrets for serviceaccount anymore for security reasons. Anyone who can read secrets can then easily escalate from default-viewer to default-editor

manifests/kustomize/base/installs/multi-user/pipelines-profile-controller/sync.py
```diff
@@ -186,12 +238,18 @@ def sync(self, parent, attachments):
...
+                    len(attachments['ConfigMap.v1']) == 3 and
+                    len(attachments['Deployment.apps/v1']) ==
+                    (1 if artifacts_proxy_enabled.lower() == 'true' else 0) and
+                    len(attachments['Service.v1']) ==
+                    (1 if artifacts_proxy_enabled.lower() == 'true' else 0) and
+                    len(attachments['ServiceAccount.v1']) == 2 and
+                    len(attachments['Role.rbac.authorization.k8s.io/v1']) == 4
+                    and
+                    len(attachments['RoleBinding.rbac.authorization.k8s.io/v1'])
+                    == 4 and 'True' or 'False'
```
juliusvonkohout, line 252: Please restore the formatting, this way it is difficult to read.

manifests/kustomize/base/installs/multi-user/pipelines-profile-controller/sync.py
```diff
@@ -173,6 +173,58 @@ def upsert_lifecycle_policy(self, bucket_name, artifact_retention_days):
...
+                    'kind': 'Secret',
+                    'metadata': {
+                        'name': agent_secret_name,
+                        'namespace': namespace,
+                        'annotations': {
+                            'kubernetes.io/service-account.name': agent_service_account_name,
+                        },
+                    },
+                    'type': 'kubernetes.io/service-account-token',
+                },
```
juliusvonkohout, line 225: Please remove this. Kubernetes does not create secrets for serviceaccount anymore for security reasons. Anyone who can read secrets can then easily escalate from default-viewer to default-editor

manifests/kustomize/base/installs/multi-user/pipelines-profile-controller/sync.py
```diff
@@ -173,6 +173,58 @@ def upsert_lifecycle_policy(self, bucket_name, artifact_retention_days):
...
+                    'metadata': {
+                        'name': agent_service_account_name,
+                        'namespace': namespace,
+                        'labels': {
+                            'application-crd-id': 'kubeflow-pipelines'
+                        }
+                    },
+                    'secrets': [{
+                        'name': agent_secret_name
+                    }]
```
juliusvonkohout, line 194: Please remove this. Kubernetes does not create secrets for serviceaccount anymore for security reasons. Anyone who can read secrets can then easily escalate from default-viewer to default-editor

manifests/kustomize/base/installs/multi-user/pipelines-profile-controller/decorator-controller.yaml
```diff
@@ -27,6 +27,18 @@ spec:
...
     updateStrategy:
       method: InPlace
+  - apiVersion: v1
+    resource: serviceaccounts
+    updateStrategy:
+      method: InPlace
+  - apiVersion: rbac.authorization.k8s.io/v1
+    resource: roles
+    updateStrategy:
+      method: InPlace
```
juliusvonkohout, line 37: Please remove this

manifests/kustomize/base/installs/multi-user/view-edit-cluster-roles.yaml
```diff
@@ -98,7 +98,8 @@ rules:
   - workfloweventbindings
   - workflowtemplates
   - workflowtaskresults
-
+  - workflowtasksets/status
+  - workflowtasksets
```
juliusvonkohout, line 102: added for https://argo-workflows.readthedocs.io/en/latest/http-template/#argo-agent-rbac

manifests/kustomize/base/installs/multi-user/ml-pipeline-driver-agent-executor-plugin-cluster-role.yaml
juliusvonkohout, line 1: May you please explain in written here as inline comments for what each permission is needed and that the permission set is minimal

manifests/kustomize/third-party/metacontroller/base/cluster-role.yaml
```diff
@@ -18,9 +18,21 @@ rules:
 - apiGroups: [""]
   resources: ["services"]
   verbs: ["get", "list", "watch", "create", "update", "patch", "delete"]
+- apiGroups: [""]
+  resources: ["serviceaccounts"]
+  verbs: ["get", "list", "watch", "create", "update", "patch", "delete"]
+- apiGroups: ["rbac.authorization.k8s.io"]
+  resources: ["roles", "rolebindings"]
```
juliusvonkohout, line 25:
```suggestion
  resources: ["rolebindings"]
```

manifests/kustomize/third-party/metacontroller/base/cluster-role.yaml
```diff
@@ -18,9 +18,21 @@ rules:
...
+  verbs: ["get", "list", "watch", "create", "update", "patch", "delete"]
 - apiGroups: ["networking.istio.io"]
   resources: ["destinationrules"]
   verbs: ["get", "list", "watch", "create", "update", "patch", "delete"]
+- apiGroups: [""]
+  resources: ["pods"]
+  verbs: ["get", "list", "watch"]
+- apiGroups: [""]
+  resources: ["persistentvolumeclaims"]
+  verbs: ["create", "get", "list"]
```
juliusvonkohout, line 35: May you please explain in written here as inline comments for what each permission is needed and that the permission set is minimal. For example since kubernetes blocks creating rolebindings to roles with higher permiisons than metacontroller itself...

manifests/kustomize/third-party/argo/base/workflow-controller-argo-taskset-clusterrole.yaml
juliusvonkohout, line 1: Does argo upstream really not include this already ?

manifests/kustomize/env/dev/driver-plugin-cm-path.yaml
```diff
@@ -0,0 +1,41 @@
...
+            key: secretkey
+    ports:
+      - containerPort: 8080
+    resources:
+      limits:
+        cpu: "1"
+        memory: 1Gi
+      requests:
+        cpu: 250m
+        memory: 512Mi
```
juliusvonkohout, line 28:
```suggestion
        cpu: 10m
        memory: 64Mi
```

250m is defintely overkill and will affect customers negatively. regarding the memory we should also optimize based on empirical measurements.

manifests/kustomize/env/cert-manager/platform-agnostic-standalone-tls/patches/ml-pipeline-driver-plugin-cm.yaml
```diff
@@ -0,0 +1,45 @@
...
+            key: secretkey
+    ports:
+      - containerPort: 8080
+    resources:
+      limits:
+        cpu: "1"
+        memory: 1Gi
+      requests:
+        cpu: 250m
+        memory: 512Mi
```
juliusvonkohout, line 29:
```suggestion
        cpu: 10m
        memory: 64Mi
```

juliusvonkohout: Same as https://github.com/kubeflow/pipelines/pull/13171/changes#r3059648327

manifests/kustomize/base/pipeline/pipeline-runner-sa.yaml
```diff
@@ -2,3 +2,12 @@ apiVersion: v1
...
   name: pipeline-runner
+
+---
+apiVersion: v1
+kind: Secret
+metadata:
+  name: pipeline-runner.service-account-token
+  annotations:
+    kubernetes.io/service-account.name: pipeline-runner
+type: kubernetes.io/service-account-token
```
juliusvonkohout, line 13: Same problem as in https://github.com/kubeflow/pipelines/pull/13171/changes#r3059538918, secrets for serviceaccounts are dangerous

manifests/kustomize/base/pipeline/pipeline-runner-role.yaml
```diff
@@ -85,3 +89,52 @@ rules:
...
+secrets:
+  - name: ml-pipeline-driver-agent-executor-plugin.service-account-token
+---
+apiVersion: v1
+kind: Secret
+metadata:
+  name: ml-pipeline-driver-agent-executor-plugin.service-account-token
+  annotations:
+    kubernetes.io/service-account.name: ml-pipeline-driver-agent-executor-plugin
+type: kubernetes.io/service-account-token
```
juliusvonkohout, line 109: same problem as https://github.com/kubeflow/pipelines/pull/13171/changes#r3059538918

manifests/kustomize/base/pipeline/ml-pipeline-driver-plugin-cm.yaml
```diff
@@ -0,0 +1,45 @@
...
+            key: secretkey
+    ports:
+      - containerPort: 8080
+    resources:
+      limits:
+        cpu: "1"
+        memory: 1Gi
+      requests:
+        cpu: 250m
+        memory: 512Mi
```
juliusvonkohout, line 32:
```suggestion
        cpu: 10m
        memory: 64Mi
```

same problem as in other comments

juliusvonkohout: I see a lot of code duplications in general in this PR. Is it not possible to use kustomize components?

juliusvonkohout: You can create a kustomize component within the argo folder and use it in 3+ places. This might apply also to other duplications.

manifests/kustomize/base/installs/multi-user/pipelines-profile-controller/sync.py
juliusvonkohout: That is really the hard blocker for this PR. We need to find a way around the secret or we create a big security problem.

manifests/kustomize/base/installs/multi-user/pipelines-profile-controller/sync.py
juliusvonkohout: That is really the hard blocker for this PR. We need to find a way around the secret or we create a big security problem.

.github/resources/manifests/standalone/tls-enabled/driver-plugin-cm-path.yaml
```diff
@@ -0,0 +1,45 @@
...
+          secretKeyRef:
+            name: mlpipeline-minio-artifact
+            key: secretkey
+    resources:
+      requests:
+        cpu: 100m
+        memory: 256Mi
+      limits:
+        cpu: 500m
+        memory: 512Mi
```
juliusvonkohout, line 29: Lets use a kustomize component, that is the fourth copy of this file

manifests/kustomize/base/installs/multi-user/pipelines-profile-controller/sync.py
juliusvonkohout: https://argo-workflows.readthedocs.io/en/latest/service-account-secrets/ is a hard blocker than and we need an upstream fix

.github/resources/manifests/base/driver-plugin-cm-path.yaml
juliusvonkohout, line 1: THis is the fith instance of this file, please use a kustomize component, for example in the argo folder.

juliusvonkohout (changes requested):
I did a thorough review and many thinks we can solve, but there is a hard blocker. The executor plugin  does not work with securely mounted secrets and would open up the escalation path from many serviceaccounts for example default-viewer to default-editor. And the executor plugin require hardcoded service account names. Bot thing must be fixed in upstream argo and then we should be good to go. But without that we create a big security problem. @ntny  is even willing to raise the PRs, we just have to put a bit of political pressure behind it as KFP and KSC members.

manifests/kustomize/base/installs/multi-user/pipelines-profile-controller/sync.py
droctothorpe: I lean towards not making this a blocker for merge. @ntny is taking steps to address this in AWF but I don't think that this should be a blocker for merging. We shouldn't make the perfect the enemy of the good. An executor plugin SA in each namespace is a negligible offense vis a vis doubling the number of pods in every single run. We can optimize iteratively.

manifests/kustomize/base/installs/multi-user/ml-pipeline-driver-agent-executor-plugin-cluster-role.yaml
```diff
@@ -0,0 +1,18 @@
...
+kind: ClusterRole
+metadata:
+  name: ml-pipeline-driver-agent-executor-plugin-cluster-role
+rules:
+  - apiGroups: [""]
+    resources: ["pods"]
+    verbs: ["get", "list", "watch"]
+  - apiGroups: [""]
+    resources: ["persistentvolumeclaims"]
+    verbs: ["create", "get", "list"]
```
juliusvonkohout: For what do we need the PVC access ?

juliusvonkohout: Please add a comment inline for what each permission is actually neeed.

juliusvonkohout: If i could read my own comments https://github.com/kubeflow/pipelines/pull/13171/changes#r3059571791 ....

ntny:
I’ll take a look at the failed tests and see what’s going on

manifests/kustomize/base/installs/multi-user/ml-pipeline-driver-agent-executor-plugin-cluster-role.yaml
ntny: @juliusvonkohout  Thanks for the reminder  added comments explaining the need for the additional RBAC [here](https://github.com/kubeflow/pipelines/pull/13171/changes/a1b40ff50ba193d816a71624afb541244a16173a#diff-d7e8ec8ecff0fc26527f6c51a55972d9bb44b8a077246d63bbb5effda0d7284cR1)

manifests/kustomize/base/installs/multi-user/ml-pipeline-driver-agent-executor-plugin-cluster-role.yaml
```diff
@@ -1,17 +1,31 @@
...
   - apiGroups: [""]
     resources: ["pods"]
     verbs: ["get", "list", "watch"]
+    # the driver may need permissions to create and delete PVCs
+    # see test_data/sdk_compiled_pipelines/valid/pipeline_with_volume.py for example
   - apiGroups: [""]
     resources: ["persistentvolumeclaims"]
-    verbs: ["create", "get", "list"]
+    verbs: ["create", "get", "list", "delete"]
+    # The root DAG driver reads the runtime configuration from the kfp-launcher cm
```
juliusvonkohout, line 21: do you mean configmap instead of cm ?

ntny: Yes, I meant ConfigMap. I usually use the shorthand cm as it’s commonly used in kubectl (e.g. kubectl -n kubeflow get cm)

manifests/kustomize/base/installs/multi-user/ml-pipeline-driver-agent-executor-plugin-cluster-role.yaml
```diff
@@ -0,0 +1,32 @@
...
+    # Kubernetes API to list pods during publishing. see publishDriverExecution method
+  - apiGroups: [""]
+    resources: ["pods"]
+    verbs: ["get", "list", "watch"]
+    # the driver may need permissions to create and delete PVCs
+    # see test_data/sdk_compiled_pipelines/valid/pipeline_with_volume.py for example
+  - apiGroups: [""]
+    resources: ["persistentvolumeclaims"]
+    verbs: ["create", "get", "list", "delete"]
+    # The root DAG driver reads the runtime configuration from the kfp-launcher cm
```
juliusvonkohout, line 21:
```suggestion
    # The root DAG driver reads the runtime configuration from the kfp-launcher configmap
```

.github/resources/manifests/kubernetes-native/default/kustomization.yaml
juliusvonkohout, line 1: should this patch not live as patch in the argo folder or so? The goal is to not patch it everywhere but have it somewhere in the common base.

.github/resources/manifests/multiuser/artifact-proxy/kustomization.yaml
juliusvonkohout, line 1: should this patch not live as patch in the argo folder or so? The goal is to not patch it everywhere but have it somewhere in the common base.

.github/resources/manifests/multiuser/cache-disabled/kustomization.yaml
juliusvonkohout, line 1: should this patch not live as patch in the argo folder or so? The goal is to not patch it everywhere but have it somewhere in the common base.

.github/resources/manifests/multiuser/default/kustomization.yaml
juliusvonkohout, line 1: should this patch not live as patch in the argo folder or so? The goal is to not patch it everywhere but have it somewhere in the common base.

juliusvonkohout, line 1: should this patch not live as patch in the argo folder or so? The goal is to not patch it everywhere but have it somewhere in the common base.

.github/resources/manifests/standalone/cache-disabled-proxy/kustomization.yaml
juliusvonkohout, line 1: should this patch not live as patch in the argo folder or so? The goal is to not patch it everywhere but have it somewhere in the common base.

.github/resources/manifests/base/driver-plugin-cm-path.yaml
juliusvonkohout: should this patch not live as patch in the argo folder or so? The goal is to not patch it everywhere 10 times but have it somewhere in the common base.

.github/resources/manifests/standalone/cache-disabled/kustomization.yaml
juliusvonkohout, line 1: should this patch not live as patch in the argo folder or so? The goal is to not patch it everywhere but have it somewhere in the common base.

.github/resources/manifests/standalone/default/kustomization.yaml
juliusvonkohout, line 1: should this patch not live as patch in the argo folder or so? The goal is to not patch it everywhere but have it somewhere in the common base.

.github/resources/manifests/standalone/proxy/kustomization.yaml
juliusvonkohout, line 1: should this patch not live as patch in the argo folder or so? The goal is to not patch it everywhere but have it somewhere in the common base.

.github/resources/manifests/standalone/tls-enabled/kustomization.yaml
juliusvonkohout, line 1: should this patch not live as patch in the argo folder or so? The goal is to not patch it everywhere but have it somewhere in the common base.

.github/resources/scripts/collect-logs.sh
```diff
@@ -29,6 +29,36 @@ function check_namespace {
     return 0
 }
 
+function describe_argo_workflows {
+    local NAMESPACE=$1
+    echo "===== Argo Workflows Inspection ====="
+    for wf in $(kubectl get wf -n "$NAMESPACE" -o json | jq -r '.items[] | select(.status.phase=="Failed" or .status.phase=="Running") | .metadata.name'); do
```
juliusvonkohout: i recommend proper long expressive human readable and pronounceable variable names

backend/src/common/util/context_logger.go
```diff
@@ -0,0 +1,73 @@
...
+)
+
+type CtxKey string
+
+const (
+	contextLoggerKey CtxKey = "driver_log_key"
+)
+
+func newFileLogger(logFile string) (*logrus.Logger, io.Closer, error) {
+	f, err := os.Create(logFile)
```
juliusvonkohout, line 19: i recommend proper long expressive human readable and pronounceable variable names

backend/src/common/util/context_logger.go
```diff
@@ -0,0 +1,73 @@
...
+	return logger, f, nil
+}
+
+// WithExistingLogger For testing only
+func WithExistingLogger(ctx context.Context, logger *logrus.Logger) context.Context {
+	return context.WithValue(ctx, contextLoggerKey, logger)
+}
+
+func WithLogger(ctx context.Context, logFile string) (context.Context, io.Closer, error) {
+	if ctx == nil {
```
juliusvonkohout, line 36: i recommend proper long expressive human readable and pronounceable variable names

backend/src/common/util/context_logger.go
```diff
@@ -0,0 +1,73 @@
...
+	if GetLoggerFrom(ctx) != nil {
+		return nil, nil, fmt.Errorf("logger already exists in context")
+	}
+
+	logger, f, err := newFileLogger(logFile)
+	if err != nil {
+		return nil, nil, fmt.Errorf(
+			"error during creation of the logger for logId: %v details: %w",
+			logFile,
+			err,
```
juliusvonkohout, line 52: i recommend proper long expressive human readable and pronounceable variable names

backend/src/driver/main.go
```diff
@@ -0,0 +1,90 @@
...
+var (
+	logLevel   = flag.String("log_level", "1", "The verbosity level to log.")
+	serverPort = flag.String("server_port", ":8080", "Server port")
+)
+
+func main() {
+	flag.Parse()
+
+	glog.Infof("Setting log level to: '%s'", *logLevel)
+	err := flag.Set("v", *logLevel)
```
juliusvonkohout, line 48: i recommend proper long expressive human readable and pronounceable variable names

backend/src/driver/main.go
```diff
@@ -0,0 +1,90 @@
...
+
+// Use WARNING default logging level to facilitate troubleshooting.
+func init() {
+	flag.Set("logtostderr", "true")
+	// Change the WARNING to INFO level for debugging.
+	flag.Set("stderrthreshold", "WARNING")
+}
+
+func parseExecConfigJSON(k8sExecConfigJSON *string) (*kubernetesplatform.KubernetesExecutorConfig, error) {
+	var k8sExecCfg *kubernetesplatform.KubernetesExecutorConfig
```
juliusvonkohout, line 69: i recommend proper long expressive human readable and pronounceable variable names

backend/src/driver/main_test.go
```diff
@@ -0,0 +1,106 @@
...
+					{SecretName: "value1"},
+				},
+			},
+			false,
+		},
+	}
+
+	for _, tc := range tt {
+		t.Logf("Running test case: %s", tc.name)
+		cfg, err := parseExecConfigJSON(tc.input)
```
juliusvonkohout, line 81: i recommend proper long expressive human readable and pronounceable variable names
```suggestion
		configuration, error := parseExecConfigJSON(tc.input)
```

backend/src/driver/rpc_handler.go
```diff
@@ -0,0 +1,475 @@
...
+	OutputPathPrefix string
+	Namespace        string
+	PipelineRoot     string
+	StoreSessionInfo string
+	LogID            string
+}
+
+func ExecutePlugin(w http.ResponseWriter, r *http.Request) {
+	defer func(Body io.ReadCloser) {
+		err := Body.Close()
```
juliusvonkohout, line 69: i recommend proper long expressive human readable and pronounceable variable names

backend/src/driver/rpc_handler.go
```diff
@@ -0,0 +1,475 @@
...
+					},
+					Message: fmt.Sprintf("unable to drive execution: failed to marshal ExecutorInput to JSON: %v", err),
+				},
+			})
+			return
+		}
+		executorInputJSON := string(executorInputBytes)
+		glog.Infof("output ExecutorInput:%s\n", prettyPrint(executorInputJSON))
+	}
+	resp := driverapi.DriverResponse{
```
juliusvonkohout, line 124: i recommend proper long expressive human readable and pronounceable variable names
```suggestion
	response := driverapi.DriverResponse{
```

backend/src/driver/rpc_handler.go
```diff
@@ -0,0 +1,475 @@
...
+		pipelineRoot     string
+		storeSessionInfo string
+		namespace        string
+		outputPathPrefix string
+	)
+	var pipeline *metadata.Pipeline
+	logID := fmt.Sprintf("%d-%v-%v-%v", time.Now().UnixMilli(), args.IterationIndex, args.Type, args.TaskName)
+	logDir := "/kfp/log"
+	logFile := fmt.Sprintf("%s/%s.log", logDir, logID)
+	ctx, f, err := util.WithLogger(context.Background(), logFile)
```
juliusvonkohout, line 325:
```suggestion
	context, file, error := util.WithLogger(context.Background(), logFile)
```

backend/src/v2/compiler/argocompiler/container.go
juliusvonkohout, line 1: Can we specify the new executor plugin image location somewhere?

backend/src/v2/compiler/argocompiler/container.go
```diff
@@ -261,27 +243,17 @@ func (c *workflowCompiler) addContainerDriverTemplate() string {
...
 	}
 	applySecurityContextToTemplate(template)
-	// If TLS is enabled (apiserver or metadata), add the custom CA bundle to the container driver template.
-	if setCABundle {
-		ConfigureCustomCABundle(template)
-	}
 	c.templates[name] = template
 	c.wf.Spec.Templates = append(c.wf.Spec.Templates, *template)
-	return name
+	return name, err
```
juliusvonkohout, line 272: i recommend proper long expressive human readable and pronounceable variable names
```suggestion
	return name, error
```

backend/src/v2/compiler/argocompiler/container.go
```diff
@@ -179,71 +162,70 @@ func (c *workflowCompiler) containerDriverTask(name string, inputs containerDriv
...
 		condition:    taskOutputParameter(name, paramCondition),
 	}
-	return dagTask, outputs
+	return dagTask, outputs, nil
 }
 
-func (c *workflowCompiler) addContainerDriverTemplate() string {
+// Create the Argo Workflow executor plugin template for the container driver.
+// See https://argo-workflows.readthedocs.io/en/latest/executor_plugins/
+func (c *workflowCompiler) addContainerDriverTemplate() (string, error) {
```
juliusvonkohout, line 189:
```suggestion
func (compiler *workflowCompiler) addContainerDriverTemplate() (string, error) {
```

i recommend proper long expressive human readable and pronounceable variable names

backend/src/v2/compiler/argocompiler/container_test.go
```diff
@@ -85,7 +90,8 @@ func TestContainerDriverTemplate_IncludesKFPPodNameEnv(t *testing.T) {
 		job: &pipelinespec.PipelineJob{},
 	}
 
-	name := c.addContainerDriverTemplate()
+	name, err := c.addContainerDriverTemplate()
```
juliusvonkohout, line 93: i recommend proper long expressive human readable and pronounceable variable names
```suggestion
	name, error := c.addContainerDriverTemplate()
```

backend/src/v2/compiler/argocompiler/dag.go
```diff
@@ -561,58 +566,56 @@ func (c *workflowCompiler) dagDriverTask(name string, inputs dagDriverInputs) (*
 	}, nil
 }
 
-func (c *workflowCompiler) addDAGDriverTemplate() string {
+// Create the Argo Workflow executor plugin template for the dag driver.
+// See https://argo-workflows.readthedocs.io/en/latest/executor_plugins/
+func (c *workflowCompiler) addDAGDriverTemplate() (string, error) {
```
juliusvonkohout, line 570:
```suggestion
func (compiler *workflowCompiler) addDAGDriverTemplate() (string, error) {
```

i recommend proper long expressive human readable and pronounceable variable names

backend/src/v2/compiler/argocompiler/plugin.go
```diff
@@ -0,0 +1,39 @@
...
+
+// Create the Argo Workflow executor plugin template with parameters.
+// // See https://argo-workflows.readthedocs.io/en/latest/executor_plugins/
+func driverPlugin(params map[string]interface{}) (*wfapi.Plugin, error) {
+	pluginConfig := map[string]interface{}{
+		"driver-plugin": map[string]interface{}{
+			"args": params,
+		},
+	}
+	jsonConfig, err := json.Marshal(pluginConfig)
```
juliusvonkohout, line 32:
```suggestion
	jsonConfig, error := json.Marshal(pluginConfig)
```

i recommend proper long expressive human readable and pronounceable variable names

backend/src/v2/driver/container_test.go
```diff
@@ -288,28 +293,31 @@ func TestContainer_CreateExecutionSuccess(t *testing.T) {
...
-		RunID:          "run-1",
-		TaskName:       "task-1",
-		Component: &pipelinespec.ComponentSpec{
-			Implementation:   &pipelinespec.ComponentSpec_ExecutorLabel{ExecutorLabel: "executor"},
-			InputDefinitions: &pipelinespec.ComponentInputsSpec{Parameters: map[string]*pipelinespec.ComponentInputsSpec_ParameterSpec{}},
-			OutputDefinitions: &pipelinespec.ComponentOutputsSpec{
-				Parameters: map[string]*pipelinespec.ComponentOutputsSpec_ParameterSpec{"output": {ParameterType: pipelinespec.ParameterType_STRING}},
+	pipeline := &metadata.Pipeline{}
+
+	execution, err := Container(util.WithExistingLogger(context.Background(), logrus.New()),
```
juliusvonkohout:
```suggestion
	execution, error := Container(util.WithExistingLogger(context.Background(), logrus.New()),
```

i recommend proper long expressive human readable and pronounceable variable names

backend/src/v2/driver/dag.go
```diff
@@ -38,14 +38,18 @@ func validateDAG(opts Options) (err error) {
 	return validateNonRoot(opts)
 }
 
-func DAG(ctx context.Context, opts Options, mlmd *metadata.Client) (execution *Execution, err error) {
+func DAG(ctx context.Context, pipeline *metadata.Pipeline, opts Options, mlmd *metadata.Client) (execution *Execution, err error) {
```
juliusvonkohout, line 43:
```suggestion
func DAG(context context.Context, pipeline *metadata.Pipeline, opts Options, mlmd *metadata.Client) (execution *Execution, error error) {
```

i recommend proper long expressive human readable and pronounceable variable names

backend/src/v2/metadata/client.go
juliusvonkohout, line 640:
```suggestion
func (client *Client) CreateExecution(ctx context.Context, pipeline *Pipeline, config *ExecutionConfig) (*Execution, error) {
```

i recommend proper long expressive human readable and pronounceable variable names

manifests/kustomize/base/pipeline/ml-pipeline-driver-plugin-cm.yaml
juliusvonkohout, line 1: This looks like a duplicate of other files

manifests/kustomize/env/cert-manager/platform-agnostic-standalone-tls/patches/ml-pipeline-driver-plugin-cm.yaml
juliusvonkohout, line 1: This looks like a duplicate

manifests/kustomize/env/dev/driver-plugin-cm-path.yaml
juliusvonkohout, line 1: This looks like a duplicate

juliusvonkohout:
Hello @ntny , I resolved many threads and added new comments. I think the main thing is the excessive manifests duplication and patch duplication (10-15 duplicates) which should be in one of the bases and then the unreadable code that deviates a lot from proper long expressive, contextless, readable, pronounceable variable naming.

ntny:
> Hello @ntny , I resolved many threads and added new comments. I think the main thing is the excessive manifests duplication and patch duplication (10-15 duplicates) which should be in one of the bases and then the unreadable code that deviates a lot from proper long expressive, contextless, readable, pronounceable variable naming.

 Hi @juliusvonkohout , thanks for the review.

  I looked into the kustomize component approach you suggested. The limitation is that Argo executor
  plugins are currently configured through a `ConfigMap`, not a structured CRD. Kubernetes parses
  `ConfigMap.data` as `map[string]string`, and Argo then unmarshals `data["sidecar.container"]` from that
  string into the container spec. So the final manifest must keep `sidecar.container` as a YAML string.

  Because of that, a component can help reuse/apply the ConfigMap patch, but it still cannot patch only
  `image` or `volumeMounts` inside `sidecar.container` as structured fields. I also tried
  `configMapGenerator` with a separate `driver-plugin-sidecar.yaml` it reduces some boilerplate, but
  image-only variants still need to replace the whole string, so the main duplication remains while
  kustomize gets more complex. 

  A cleaner way to remove this duplication is to move executor plugin settings into the Argo Workflow spec,
  where the sidecar is structured YAML. I have an upstream Argo PR for that here:
  https://github.com/argoproj/argo-workflows/pull/15440

  Small update: during testing, @droctothorpe  found failures on very large parallel pipelines. I investiga
  ted and found this is an issue in Argo’s executor plugin implementation. I also opened fixes upstream, c
  urrently waiting for review: https://github.com/argoproj/argo-workflows/pull/16075

  As far as I know, the central driver work is currently postponed until the remove-MLMD release, since
  that work should significantly reduce workflow size. @droctothorpe and @HumairAK  can share more details on that.. 
If Imanage to finish and merge the workflow-level executor plugin configuration in Argo before that, we may be able to remove this duplication before merging this work.

juliusvonkohout:
please rebase to master

ntny:
Well, all tests are finally passing 

On Monday I'll post an update with the current status and a summary of what changed during the rebase.

ntny:
Hi @mprahl, could you please take a look at the MLflow runtime config [changes](https://github.com/kubeflow/pipelines/pull/13171/changes/5dda2a699c0bdbc02c9226b9959407796ab92d01) for the driver plugin?

Since the driver is now an Argo executor plugin, its env vars and pod annotations are only available for global plugin configuration. Because of that, I no longer pass `KFP_MLFLOW_CONFIG` to the driver through env vars.

Previously the flow was:
  - add the driver role annotation to the pod in the compiler
  - in `UpsertRuntimeEnvVars`, add `KFP_MLFLOW_CONFIG` to the driver env vars

Now I detect the driver plugin via `tmpl.Plugin != nil` instead of the driver's pod annotation role. If the driver role is requested, I [update](https://github.com/ntny/pipelines/blob/central-driver-poc-backup/backend/src/common/util/workflow.go#L678-L748) `driver-plugin.args.runtime_args` instead of env vars and put the runtime config 

The driver plugin then parses `runtime_args` [here](https://github.com/ntny/pipelines/blob/central-driver-poc-backup/backend/src/driver/driverapi/request.go#L23-L44)

I also added MLflow RBAC (`mlflow.kubeflow.org/experiments` `get/list/create/update`) for the separate driver plugin service account. Without this, the API server can create the root run successfully, but task-level nested runs from the driver plugin get `403`, because MLflow Kubernetes auth checks the bearer token from the pod service account.
- [Standalone RBAC/SA](https://github.com/ntny/pipelines/blob/central-driver-poc-backup/manifests/kustomize/base/pipeline/pipeline-runner-role.yaml#L102-L162)
- [Multi-user ClusterRole](https://github.com/ntny/pipelines/blob/central-driver-poc-backup/manifests/kustomize/base/installs/multi-user/ml-pipeline-driver-agent-executor-plugin-cluster-role.yaml#L1-L35)

ntny:
> Copilot wasn't able to review this pull request because it exceeds the maximum number of lines (20,000). Try reducing the number of changed lines and requesting a review from Copilot again.

This is probably the biggest PR I've ever worked on :)

juliusvonkohout:
I think you have to rebase

ntny:
I'll leave a detailed comment a bit later about what changed when merging the MLMD removal branch; the main change concerns how the driver obtains the token for creating tasks. I'll also add that I'll fix any possible issues in E2E

@droctothorpe  Hi, could you please help reopen the PR? It seems this happened because the linked issue was closed automatically.

ntny:
@HumairAK @droctothorpe @juliusvonkohout 
hi! The most important change resulting from rebasing this branch onto the latest master is how the centralized driver obtains the run-scoped KFP token:


Previously, each driver ran in its own workflow Pod and inherited the workflow’s runtime ServiceAccount, which is configured at the Workflow level
  https://github.com/kubeflow/pipelines/blob/8cdc86bdd4c417fe5dc1f2eb6e93f1655570391e/backend/src/v2/compiler/argocompiler/argo.go#L188-L205 The compiler mounted a projected, automatically rotated token with a run-scoped KFP audience into both the container driver
  https://github.com/kubeflow/pipelines/blob/8cdc86bdd4c417fe5dc1f2eb6e93f1655570391e/backend/src/v2/compiler/argocompiler/container.go#L263-L293 and the DAG driver https://github.com/kubeflow/pipelines/blob/8cdc86bdd4c417fe5dc1f2eb6e93f1655570391e/backend/src/v2/compiler/argocompiler/dag.go#L599-L630 
The driver read that token and automatically picked up kubelet rotations through its token source https://github.com/kubeflow/pipelines/blob/8cdc86bdd4c417fe5dc1f2eb6e93f1655570391e/backend/src/v2/apiclient/auth.go#L30-L66

With the centralized driver, the driver runs as an executor plugin sidecar in the Argo agent Pod. In Argo v4.1.2, the agent Pod uses the workflow’s runtime ServiceAccount but disables the normal ServiceAccount token automount
  https://github.com/argoproj/argo-workflows/blob/v4.1.2/workflow/controller/agent.go#L155-L167 The executor plugin sidecar is instead given a separate <plugin>-executor-plugin ServiceAccount identity plugin-sidecar implementation https://github.com/argoproj/argo-workflows/blob/v4.1.2/workflow/controller/agent.go#L340-L358 Argo mounts this identity using a legacy ServiceAccount token Secret (token-volume implementation https://github.com/argoproj/argo-workflows/blob/v4.1.2/workflow/controller/serviceaccount.go#L13-L30

The driver uses the plugin identity to access Kubernetes. It fetches the agent Pod, verifies that the request namespace and run ID match the Pod, checks its
Workflow owner reference, and reads the runtime ServiceAccount from pod.spec.serviceAccountName implementation https://github.com/ntny/pipelines/blob/central-driver-poc-backup/backend/src/driver/plugin_auth.go#L86-L151

The driver then uses the plugin identity to request a short-lived token for the runtime ServiceAccount. The token has a run-scoped KFP audience, is bound to the agent Pod, and is refreshed by the driver before it expires (TokenRequest implementation https://github.com/ntny/pipelines/blob/central-driver-poc-backup/backend/src/v2/apiclient/token_request.go#L52-L146  RBAC permits token creation only for explicitly listed runtime ServiceAccounts: pipeline-runner in standalone mode https://github.com/ntny/pipelines/blob/central-driver-poc-backup/manifests/kustomize/base/pipeline/pipeline-runner-role.yaml#L147-L152 and default-editor in the standard multi-user installation https://github.com/ntny/pipelines/blob/central-driver-poc-backup/manifests/kustomize/base/installs/multi-user/ml-pipeline-driver-agent-executor-plugin-cluster-role.yaml#L16-L21

  This workaround has several limitations:
  - The plugin’s Kubernetes credential is a persistent legacy token Secret shared by runs in the namespace.
  - RBAC restricts which ServiceAccounts the plugin may request tokens for, but it cannot enforce the requested audience or Pod binding. If the plugin credential
    is compromised, it could be used to request a differently scoped token for an allowlisted ServiceAccount. This is the main security tradeoff.
- Custom runtime ServiceAccounts must be explicitly added to the token-creation RBAC allowlist. Otherwise, their runs fail during driver initialization.

We should remove this workaround once the Argo version used by KFP can provide executor plugins with a projected, automatically rotated runtime-ServiceAccount token with the required KFP audience, and KFP upgrades to that version.

droctothorpe:
image

Hm, I'm unable to reopen the PR since you rebased. Weird GitHub policy. You can open a new PR and link to this one to preserve history. Apologies for the inconvenience. This should not have been closed since you're a kubeflow org member. If it happens again (and was not a one off), I'll investigate.

PR #13175, opened 2026-03-30
fix(backend): isolate executor-logs per retry pod

droctothorpe:
## Summary

Fixes executor log artifact collisions across Argo retries. Each retry attempt now writes logs to a distinct, sequentially-numbered object-store key (`executor-logs-0`, `executor-logs-1`, …) and registers a corresponding distinct MLMD artifact URI. Previously all retry pods shared the single URI provisioned by the driver before any attempt ran, causing every retry to overwrite the same S3 key.

## Root Cause

The driver runs once per task and calls `provisionOutputs()` with a single `outputURISalt`, baking that URI into the `ExecutorInput` proto that all Argo retry pods receive unchanged. The launcher previously passed this URI through without modification, so every attempt overwrote `…/executor-logs` with its logs, and MLMD ended up with N artifact rows all pointing to the same (last-write-wins) file.

## What Changed

**`backend/src/v2/component/launcher_v2.go`**
- Added `qualifyExecutorLogsURI(artifacts, retryIndex)`: appends `-{retryIndex}` to the `executor-logs` artifact URI before `prepareOutputFolders` runs, so each attempt creates its own path.
- Added `retryIndexFromPodAnnotation(ctx, k8sClient, namespace, podName)`: reads the `workflows.argoproj.io/node-name` pod annotation (e.g. `…executor(3)`) and parses the 0-based retry index from it. This is the fallback path.
- In `execute()`: prefers `KFP_RETRY_INDEX` env var (set by the Argo compiler via `{{retries}}` — fast path, no k8s API call). Falls back to the pod annotation read. Defaults to `"0"` if neither is available.
- Removed `podName` parameter from `execute()` and `executeV2()` — no longer needed.

**`backend/src/v2/compiler/argocompiler/common.go`**
- Added `KFP_RETRY_INDEX: "{{retries}}"` to `commonEnvs`. Argo substitutes `{{retries}}` at pod creation time with the 0-based attempt count, making the env-var path zero-cost (no k8s API call) once a cluster is running the updated API server.

**`backend/src/v2/component/constants.go`**
- Added `EnvRetryIndex = "KFP_RETRY_INDEX"`.

## Tests

- Added `Test_retryIndexFromPodAnnotation` (5 cases: happy path ×2, missing annotation, no parenthesised suffix, non-integer index).
- Updated `Test_qualifyExecutorLogsURI`: renamed `podName` → `retryIndex`, updated expected URIs to use `-N` suffix.
- Updated `Test_executeV2_publishLogs`: replaced `podName` field with `retryIndex`, sets `KFP_RETRY_INDEX` env var via `t.Setenv`, asserts sequential `executor-logs-N` keys.
- Removed `podName` arg from `Test_executeV2_Parameters` `executeV2` call.
- `go test ./backend/src/v2/...` — all packages pass.

## E2E Verification (local KFP standalone on OrbStack)

Submitted a `retry=4` pipeline (5 total attempts, always-fail component):

**Before fix** — all 5 MLMD artifacts pointed to the same URI; 1 S3 file (last-write-wins):
```
…/always-fail/{salt}/executor-logs   ×5  (identical)
```

**After fix** — 5 distinct URIs, 5 distinct S3 files, timestamps confirm each is a separate attempt:
```
…/always-fail/{salt}/executor-logs-0   18:21:22
…/always-fail/{salt}/executor-logs-1   18:21:32
…/always-fail/{salt}/executor-logs-2   18:21:42
…/always-fail/{salt}/executor-logs-3   18:21:52
…/always-fail/{salt}/executor-logs-4   18:22:02
```

## Risk / Compatibility

- **Low risk**: behavior only changes in the `publishLogs=true` path.
- **Backward compatible**: if `KFP_RETRY_INDEX` is unset (older API server) and the pod annotation is absent, the index defaults to `"0"` — same as the old single-file behavior, but no regression beyond that.
- The annotation fallback means the launcher fix is self-contained and works independently of the API server version.

### DIFF
--- a/test_data/compiled-workflows/add_numbers.yaml
+++ b/test_data/compiled-workflows/add_numbers.yaml
@@ -1,6 +1,7 @@
 apiVersion: argoproj.io/v1alpha1
 kind: Workflow
 metadata:
+  creationTimestamp: null
   generateName: add-numbers-
 spec:
   arguments:
@@ -26,10 +27,6 @@ spec:
       pipelines.kubeflow.org/v2_component: "true"
     labels:
       pipelines.kubeflow.org/v2_component: "true"
-  securityContext:
-    runAsNonRoot: true
-    seccompProfile:
-      type: RuntimeDefault
   serviceAccountName: pipeline-runner
   templates:
   - container:
@@ -98,14 +95,6 @@ spec:
         requests:
           cpu: 100m
           memory: 64Mi
-      securityContext:
-        allowPrivilegeEscalation: false
-        capabilities:
-          drop:
-          - ALL
-        runAsNonRoot: true
-        seccompProfile:
-          type: RuntimeDefault
     inputs:
       parameters:
       - name: component
@@ -134,10 +123,6 @@ spec:
         valueFrom:
           default: "true"
           path: /tmp/outputs/condition
-    securityContext:
-      runAsNonRoot: true
-      seccompProfile:
-        type: RuntimeDefault
   - dag:
       tasks:
       - arguments:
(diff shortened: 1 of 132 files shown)


droctothorpe:
Here's a screenshot of what a failed task with 4 retries looks like in the UI:
image

Issue #1426, opened 2019-06-03
cloudpickle import error


hougangliu:
after #1372, cloudpickle must be installed in _default_base_image='tensorflow/tensorflow:1.11.0-py3', otherwise comp.func_to_container_op will fail owning to importing cloudpickle error

Ark-kun:
Thanks for discovering the bug! I'll fix it ASAP.

PR #13176, opened 2026-03-30
fix(backend): make runtime status errors JSON-safe

yashasvi290906:
## Summary

Fixes the `RuntimeStatus.Error` serialization issue that prevents runtime errors from being safely persisted in `state_history`.

## Problem

`RuntimeStatus.Error` was typed as Go's `error` interface, which does not serialize/deserialize properly through JSON. This caused:

- `"Error": {}` during marshaling
- unmarshal failures for `state_history`
- inability to persist runtime errors (blocking TODO(#1426))

## Changes

- Replaced `error` with a concrete `*model.RuntimeError`
- Added JSON-safe error struct with exported fields
- Added helper constructors for runtime error/status
- Updated `run_store.go` to use new constructor
- Fixed model ↔ API conversion in `api_converter.go`
- Updated tests to use new runtime error structure

## Why no migration is needed

`StateHistory` is already stored as JSON and existing rows do not rely on serialized Go error types.

## Testing

```bash
go test ./backend/src/apiserver/server/... ./backend/src/apiserver/storage/...

### DIFF
--- a/backend/src/apiserver/model/run.go
+++ b/backend/src/apiserver/model/run.go
@@ -15,7 +15,10 @@
 package model
 
 import (
+	"fmt"
 	"strings"
+
+	"github.com/kubeflow/pipelines/backend/src/common/util"
 )
 
 type (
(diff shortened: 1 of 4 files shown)


yashasvi290906:
Hi maintainers, this PR addresses the serialization/model blocker for persisting runtime errors in `state_history`.

Could someone please take a look and run `/ok-to-test` if it looks good?

yashasvi290906:
Thanks for the detailed review!

You're right — the current implementation forces errors into InvalidArgument and does not preserve the original gRPC status information.

I'll update the implementation to:
- preserve the original error message without rewriting
- avoid forcing a specific error type/code
- improve RuntimeError.Type to reflect actual error type
- add additional test cases for different error codes

Will push an update shortly.

yashasvi290906:
Thanks for the review — I updated the implementation to preserve runtime status error code/message, removed the duplicate doc comment, and added coverage for a non-InvalidArgument error case.

I also updated the API conversion to keep compatibility with existing runtime status expectations and tests.

droctothorpe:
@yashasvi290906: Hey, thank you for your PR! As you are not a member of Kubeflow organization, your PR must link to an issue with 'ready' label. Please read our [CONTRIBUTING.md](https://github.com/kubeflow/pipelines/blob/master/CONTRIBUTING.md), closing this PR

PR #13179, opened 2026-03-31
fix(backend): increase MLMD client retry budget for MySQL deadlocks

droctothorpe:
## Summary

When many pipeline tasks run in parallel, their driver and launcher pods simultaneously call `PutExecution` on the MLMD gRPC server. The server translates these into concurrent MySQL transactions that deadlock each other, returning `codes.Aborted` to the client. The previous retry configuration (3 retries, 300 ms base backoff) was exhausted quickly under sustained contention, causing pods to exit non-zero and Argo to mark steps as failed.

## Changes

**`backend/src/v2/metadata/client.go`**
- Raise `mlmdClientSideMaxRetries` from 3 → 10
- Replace `BackoffExponentialWithJitter(300ms, 0.20)` with `BackoffExponentialWithJitterBounded(1s, 0.25, 30s)` — bounded exponential so retries spread out under contention without stalling a pod indefinitely (max 30 s per attempt)
- Also retry on `codes.Unavailable` in addition to `codes.Aborted` to cover transient connectivity losses to the MLMD server
- Extract backoff parameters into named constants for readability

## Testing

Validated on a local OrbStack cluster with custom-built `kfp-driver:mlmd-retry-fix` and `kfp-launcher:mlmd-retry-fix` images:
- Deployed KFP standalone (`manifests/kustomize/env/platform-agnostic`)
- Submitted a 6-task fan-out/fan-in pipeline with `enable_caching=False` to force concurrent `PutExecution` calls
- Run completed with state `SUCCEEDED` in ~80 s

## Retry budget analysis

| | Before | After |
|---|---|---|
| Max retries | 3 | 10 |
| Base backoff | 300 ms | 1 s |
| Max per-attempt wait | unbounded | 30 s |
| Retry codes | `Aborted` | `Aborted`, `Unavailable` |
| Worst-case total wait | ~2.4 s | ~195 s (capped) |

### DIFF
--- a/backend/test/testutil/kubernetes_utils.go
+++ b/backend/test/testutil/kubernetes_utils.go
@@ -78,6 +78,7 @@ func ReadPodLogs(client *kubernetes.Clientset, namespace string, podName string,
 			podLogs, err := podLogsRequest.Stream(context.Background()) // Pass a context for cancellation
 			if err != nil {
 				logger.Log("Failed to stream pod logs due to %v", err)
+				continue
 			}
 			defer func(podLogs io.ReadCloser) {
 				err = podLogs.Close()


droctothorpe:
## CI Status Note

The 6 failing checks are all `End to End E2ECritical/E2EEssential Tests - K8s v1.34.0` jobs. These same jobs failed on the most recent master merge (run [23776177814](https://github.com/kubeflow/pipelines/actions/runs/23776177814), commit `eab7c115`, PR #13175) before this PR was opened — confirming they are **pre-existing flakes on K8s v1.34.0**, not regressions introduced here.

Specifically failing test cases (`nested_pipeline_opt_input_child_level_compiled.yaml`, `modelcar.yaml`) exhibit the same `componentInput* should be X but is None` pattern seen in the baseline master run.

All other checks (127 passing) are green, including `backend-tests`, `KFP SDK Client Tests`, `KFP API Server Tests`, `KFP upgrade tests`, `KFP Webhooks`, and all image builds.

droctothorpe:
## E2E flakiness analysis

Filed #13180 with a full root cause analysis of the intermittent K8s v1.34.0 E2E failures observed on master and on this PR.

**TL;DR**: The `kindest/node:v1.34.0` Docker image is a mutable floating tag with no SHA digest pinned in the cache key. When the cache expired around 2026-03-25, CI pulled an updated image that subtly changed cluster behavior, making the pre-existing flaky tests fail more often. No code change in this repo caused the regression.

This commit also adds a fix for a nil pointer panic in `ReadPodLogs` (secondary failure mode) where `podLogs.Close()`/`io.Copy` were called on a nil reader when `Stream()` returned an error.

PR #13186, opened 2026-04-01
fix(frontend): persist rows per page selection in localStorage

### LINKED ISSUE #13185: [feature] Cookie to store selected rows per page
### Feature Area

/area frontend

I would like for the value of the runs per page to be sticky when navigating around the KFP UI.

Image

As an end user navigating through the UI multiple times I would like for the value of rows per page to not reset every time I navigate away to another view.

---

Love this idea? Give it a 👍.

Priyanshu-u07:

The `CustomTable` component resets the "Rows per page" selection to `10` whenever the user navigates away and returns. This is because `pageSize` is stored only in React component state, which is destroyed on unmount.

This PR persists the rows-per-page preference in `localStorage`, with each page (Runs, Pipelines, Experiments, etc.) storing its value independently using the URL hash as a page identifier.

**Changes:**

- `LocalStorage.ts`:  Added `getTablePageSize()` and `saveTablePageSize()` with optional `pageId` for per-page storage.
- `CustomTable.tsx`: Read saved page size on mount, save on change, derive page key from URL hash.
- `LocalStorage.test.ts`: Added 6 unit tests covering default values, validation and per-page independence.

https://github.com/user-attachments/assets/98fc68e6-a712-40c8-9afa-d913fc150a3a

Fixes #13185


frontend/src/components/CustomTable.tsx
```diff
@@ -619,10 +620,16 @@ export default class CustomTable extends React.Component<CustomTableProps, Custo
...
-
+    LocalStorage.saveTablePageSize(pageSize, this._getPageId());
     this._resetToFirstPage(await this.reload({ pageSize, pageToken: '' }));
   }
 
+  private _getPageId(): string | undefined {
+    const hash = window.location.hash;
+    const match = hash.match(/^#\/([a-zA-Z_/]+)/);
+    return match ? match[1] : undefined;
+  }
```
jeffspahr, line 647: This key needs to be stable for the page type, not derived from whatever alpha prefix happens to be present in the resource id. Right now `#/runs/details/run-abc123` becomes `runs/details/run`, which means two different run-details pages do not reliably share the same preference. Normalizing the hash to known static route segments fixes that without changing the public `LocalStorage` API.

```suggestion
  private _getPageId(): string | undefined {
    const pathSegments = window.location.hash
      .split('?')[0]
      .replace(/^#\//, '')
      .split('/')
      .filter(Boolean);

    if (pathSegments.length === 0) {
      return undefined;
    }

    const [root, second] = pathSegments;

    if (root === 'shared' && second === 'pipelines') {
      return 'shared/pipelines';
    }

    if (root === 'archive' && second) {
      return `${root}/${second}`;
    }

    if (second && ['details', 'new', 'lineage'].includes(second)) {
      return `${root}/${second}`;
    }

    return root;
  }
```

frontend/src/components/CustomTable.test.tsx
```diff
@@ -737,4 +738,28 @@ describe('CustomTable', () => {
...
+  });
+
+  it('saves page size to localStorage when changed', async () => {
+    const wrapper = renderTable({ rows: [], columns, reload: vi.fn(async () => '') });
+    fireEvent.mouseDown(getRowsPerPageCombobox());
+    fireEvent.click(await screen.findByText('20'));
+    await TestUtils.flushPromises();
+    expect(localStorage.getItem('tablePageSize')).toBe('20');
+    wrapper.unmount();
+  });
```
jeffspahr: The new tests only cover the global fallback key, so they would still pass if a details route persisted to something resource-specific like `tablePageSize_runs/details/run`. A parameterized-route regression makes the behavior this PR is trying to add explicit.

```suggestion
  it('persists page size across resources on the same details page', async () => {
    window.location.hash = '#/runs/details/run-abc123';

    const firstWrapper = renderTable({ rows: [], columns, reload: vi.fn(async () => '') });
    fireEvent.mouseDown(getRowsPerPageCombobox());
    fireEvent.click(await screen.findByText('20'));
    await TestUtils.flushPromises();
    firstWrapper.unmount();

    const reload = vi.fn(async () => '');
    window.location.hash = '#/runs/details/run-def456';

    const secondWrapper = renderTable({ rows, columns, reload });
    await waitFor(() =>
      expect(reload).toHaveBeenLastCalledWith({
        filter: '',
        orderAscending: false,
        pageSize: 20,
        pageToken: '',
        sortBy: '',
      }),
    );
    secondWrapper.unmount();
    window.location.hash = '';
  });
```

Priyanshu-u07: Done.

frontend/src/components/CustomTable.tsx
Priyanshu-u07: Done

jeffspahr: Verified on current head `efeca9629af2565ad149445dadb2885aa933320b`: `_getPageId()` now normalizes parameterized routes to stable page segments, so this issue is fixed.

frontend/src/components/CustomTable.test.tsx
jeffspahr: Verified on current head `efeca9629af2565ad149445dadb2885aa933320b`: the parameterized-route regression test is in place and passes in the targeted `CustomTable` suite, so this thread is resolved.

Issue #12008, opened 2025-06-24
kfp metadata-writer crashes with "Too large resource version" on on-prem Kubernetes 1.32.1 / etcd 3.5.9 cluster


keshabamahapatra:
### Environment

*  How did you deploy Kubeflow Pipelines (KFP)?
    Installed using manifests Link (https://github.com/kubeflow/manifests/tree/v1.10.1)

*  KFP version: 
    pipeline:2.5.0 

To find the version number, See version number shows on bottom of KFP UI left sidenav. -->
*  KFP SDK version: 
    2.11.0

### Steps to reproduce

I recently installed Kubeflow (Version kubeflow-v-1-10-1) on On-Prem Kubernetes Cluster, getting error pipeline metadata-writer pod.
It keep crashing with the below error messages.

kubectl -n kubeflow get pods metadata-writer-69fb57bdc-v5tzl
NAME                              READY   STATUS             RESTARTS       AGE
metadata-writer-69fb57bdc-v5tzl   1/2     CrashLoopBackOff   23 (93s ago)   105m

Connected to the metadata store
Start watching Kubernetes Pods created by Argo
Traceback (most recent call last):
  File "/kfp/metadata_writer/metadata_writer.py", line 168, in <module>
    for event in pod_stream:
  File "/usr/local/lib/python3.9/site-packages/kubernetes/watch/watch.py", line 195, in stream
    raise client.rest.ApiException(
kubernetes.client.exceptions.ApiException: (504)
Reason: Timeout: Timeout: Too large resource version: 944613377, current: 944613361

kubectl version
Client Version: v1.32.1
Server Version: v1.32.1

etcdctl version
etcdctl version: 3.5.9
API version: 3.5

### Materials and Reference

---

Impacted by this bug? Give it a 👍.

keshabamahapatra:
Hi @OutSorcerer , Do you know any solutions to this problem? could you help me.
Thank you in advance!

yashasvi290906:
Hi @keshabamahapatra , I’d like to work on this issue.

I’ll begin by reproducing the metadata-writer crash and investigating how resourceVersion is handled in the watch loop, especially during timeouts and reconnections.

yashasvi290906:
Hi @keshabamahapatra, I’ve opened a PR to address this issue:

https://github.com/kubeflow/pipelines/pull/13193

The fix handles the stale `resourceVersion` case in the metadata-writer watch loop by catching the error and restarting the watch instead of crashing.

[...]

PR #13193, opened 2026-04-01
fix(metadata-writer): handle stale resourceVersion in watch stream to prevent crash

yashasvi290906:
### Problem

The metadata-writer crashes with:
"Too large resource version"

This happens when the Kubernetes watch stream uses an invalid or stale resourceVersion, typically due to etcd compaction.

### Solution

* Move watch initialization inside try block
* Catch ApiException for stale resourceVersion
* Restart watch from latest state instead of crashing
* Add retry delay to prevent tight loop

### Impact

* Prevents metadata-writer CrashLoopBackOff
* Improves robustness of watch handling
* Aligns with Kubernetes watch best practices

### Notes

This leverages the existing behavior that new watchers receive the latest resource state.

### DIFF
--- a/backend/metadata_writer/src/metadata_writer.py
+++ b/backend/metadata_writer/src/metadata_writer.py
@@ -17,13 +17,14 @@ import hashlib
 import os
 import re
 import collections
+import traceback
 import kubernetes
 import yaml
 from time import sleep
 import lru
 
 from metadata_helpers import *
-
+from kubernetes.client.exceptions import ApiException
 
 namespace_to_watch = os.environ.get('NAMESPACE_TO_WATCH', 'default')
 pod_name_to_execution_id_size = os.environ.get('POD_NAME_TO_EXECUTION_ID_SIZE', 5000)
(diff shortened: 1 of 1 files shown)


yashasvi290906:
Hi reviewers, this PR handles the metadata-writer watch failure caused by stale resourceVersion errors by restarting the watch instead of crashing the process.

Since this is my first PR to Kubeflow Pipelines, could a maintainer please take a look and, if it looks reasonable, help with `/ok-to-test`?

yashasvi290906:
Hi @hbelmiro , @HumairAK ,

I’ve addressed the review comments, cleaned up the exception handling, and updated the PR with DCO sign-off.

Since this is my first PR to Kubeflow Pipelines, could a maintainer please take a look and, if it looks reasonable, help with `/ok-to-test`?

Thanks!

jeffspahr:
Thanks for investigating the metadata-writer watch failure. Closing this PR after reviewing it against release-2.18.

That branch already catches exceptions raised while iterating the watch and restarts the stream. A focused comparison using the Kubernetes watcher with mocked API responses showed that both the baseline and this patch recover from 504, 410, and 403 responses and issue the same subsequent watch request. The main behavioral change is a one-second delay for resource-version errors, rather than an additional recovery path. We are not taking this as a 2.18 crash fix.

The reporter of #12008 also confirmed that upgrading etcd resolved their problem. On master / KFP 3.0, #13986 removed metadata-writer entirely. A future retry-throttling change for an MLMD-based release would need a focused rationale, retained error diagnostics, and regression coverage.

PR #13195, opened 2026-04-01
test(backend): Add MLFlow Integrations tests with KFP

VaniHaripriya:

**Summary**
Adds automated end-to-end tests that run KFP against a real MLflow tracking server, and wires a CI job so those tests run on Kind with MLflow deployed and the API server configured for the plugin.

**What’s changed**

- New E2E tests — Ginkgo specs that create pipeline runs with MLflow-related plugins_input, verify plugins_output.mlflow, and call the MLflow REST API to check parent/nested runs, tags, and status (including multi-task, parallel-for, failure, retry, opt-out, and default-plugin behavior).
- Test helpers — Shared utilities for MLflow HTTP calls (tracking URI, TLS, bearer token, workspace header), run creation with plugins_input, and assertions on plugin output and nested runs.
- CI workflow — New E2E job that deploys MLflow, runs the configure script (API server plugin + port forwards + env for tests), and executes the suite with the MLflow label and -mlflowEnabled=true.
- Configure script — Script used by CI to patch the API server with MLflow settings and grant the pipeline service account access to MLflow.

**Checklist:**

### DIFF
--- a/.github/resources/scripts/configure-mlflow.sh
+++ b/.github/resources/scripts/configure-mlflow.sh
@@ -1,0 +1,119 @@
+#!/bin/bash
+# Copyright 2026 The Kubeflow Authors
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#      https://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+# Purpose:
+# This script configures KFP to use an already-deployed MLflow instance for
+# MLflow E2E tests.
+#
+# CI helper: patch the KFP API server with plugins.mlflow, roll it out, and
+# port-forward the API server and MLflow so E2E tests can reach both.
+# It also exports workspace/auth variables used by MLflow test helpers.
+#
+# Usage: configure-mlflow.sh <KFP_NAMESPACE> <MLFLOW_NAMESPACE> <CONFIG_JSON_PATH>
+
+set -e
+
+KFP_NAMESPACE="${1:?KFP namespace required}"
+MLFLOW_NAMESPACE="${2:?MLflow namespace required}"
+CONFIG_JSON_PATH="${3:?Path to source config.json required}"
+
+echo "Services in ${MLFLOW_NAMESPACE} namespace:"
+kubectl get svc -n "$MLFLOW_NAMESPACE" --no-headers
+MLFLOW_SVC=$(kubectl get svc -n "$MLFLOW_NAMESPACE" --no-headers -o custom-columns=":metadata.name" | grep -i mlflow | head -1)
+if [ -z "$MLFLOW_SVC" ]; then
+  echo "ERROR: No service matching 'mlflow' found in namespace $MLFLOW_NAMESPACE"
+  exit 1
+fi
+MLFLOW_PORT=$(kubectl get svc -n "$MLFLOW_NAMESPACE" "$MLFLOW_SVC" -o jsonpath='{.spec.ports[0].port}')
+MLFLOW_HOST="${MLFLOW_SVC}.${MLFLOW_NAMESPACE}.svc.cluster.local"
+MLFLOW_ENDPOINT="https://${MLFLOW_HOST}:${MLFLOW_PORT}"
+echo "MLflow service: $MLFLOW_SVC port=$MLFLOW_PORT endpoint=$MLFLOW_ENDPOINT"
+
+MLFLOW_PATCH=$(jq -n --arg endpoint "$MLFLOW_ENDPOINT" '{
+  endpoint: $endpoint,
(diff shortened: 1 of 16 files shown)


.github/resources/scripts/configure-mlflow.sh
```diff
@@ -0,0 +1,109 @@
+#!/bin/bash
```
nsingla, line 1: Just curious, why do we need this file? deploy mlflow is already doing this and returning only after mlflow instance is found healthy

backend/src/common/client/api_server/v2/run_client.go
```diff
@@ -36,6 +36,7 @@ type RunInterface interface {
 	Get(params *params.RunServiceGetRunParams) (*model.V2beta1Run, error)
 	List(params *params.RunServiceListRunsParams) ([]*model.V2beta1Run, int, string, error)
 	ListAll(params *params.RunServiceListRunsParams, maxResultSize int) ([]*model.V2beta1Run, error)
+	Retry(params *params.RunServiceRetryRunParams) error
```
nsingla, line 39: why is non test code part of this PR? shouldn't this be a separate PR against master branch?

.github/resources/scripts/configure-mlflow.sh
VaniHaripriya: The script contains the patches KFP plugin config, forwards ports, and exports MLflow test env vars. Updated the header comments with the purpose of the script.

manifests/kustomize/base/installs/multi-user/api-service/cluster-role.yaml
```diff
@@ -61,3 +61,11 @@ rules:
   - tokenreviews
   verbs:
   - create
+- apiGroups:
+  - ""
```
nsingla, line 65: We discussed potentially using pipelines.kubeflow.org as the api group here

manifests/kustomize/base/pipeline/ml-pipeline-apiserver-role.yaml
```diff
@@ -63,3 +63,20 @@ rules:
...
+  - mlflow.kubeflow.org
+  resources:
+  - experiments
+  verbs:
+  - get
+  - list
+  - create
+  - update
+- apiGroups:
+  - ""
```
nsingla, line 76: same here

backend/test/end2end/utils/mlflow_utils.go
```diff
@@ -0,0 +1,450 @@
...
+		return nil
+	}
+	_, ok := run.PluginsOutput[mlflowPluginKey]
+	if ok {
+		return fmt.Errorf("plugins_output should not contain %q key when MLflow is disabled", mlflowPluginKey)
+	}
+	return nil
+}
+
+type MLflowExperiment struct {
```
nsingla: we talked about reusing client code from mlflow package rather than redefining it here

manifests/kustomize/base/installs/multi-user/api-service/cluster-role.yaml
VaniHaripriya: Since ConfigMap is a core API object, its group is empty string.

manifests/kustomize/base/pipeline/ml-pipeline-apiserver-role.yaml
VaniHaripriya: Since ConfigMap is a core API object, its group is empty string.

backend/test/end2end/utils/mlflow_utils.go
VaniHaripriya: Update to use the client code.

.github/workflows/e2e-test.yml
```diff
@@ -256,3 +261,118 @@ jobs:
...
+            NAMESPACE=${{ inputs.namespace }}
+          fi
+
+          {
+            echo "NUMBER_OF_NODES=$NUMBER_OF_NODES"
+            echo "TEST_LABEL=$TEST_LABEL"
+            echo "NAMESPACE=$NAMESPACE"
+          } >> "$GITHUB_OUTPUT"
+
+      - name: Build and upload the sample Modelcar image to Kind
```
nsingla, line 356: are you using model car test in this suite?

nsingla (approved):
/lgtm

.github/workflows/e2e-test.yml
```diff
@@ -256,3 +261,118 @@ jobs:
...
+        proxy: [ "false" ]
+        test_label: [ "MLflow" ]
+        pod_to_pod_tls_enabled: [ "false" ]
+        multi_user: [ "false" ]
+        artifact_proxy: [ "false" ]
+        artifact_storage: [ "file", "s3" ]
+        backend_store: [ "postgres" ]
+        registry_store: [ "postgres" ]
+      fail-fast: false
+    name: End to End Critical Scenario MLflow Tests - K8s ${{ matrix.k8s_version }} cacheEnabled=${{ matrix.cache_enabled }} multiUser=${{ matrix.multi_user }} artifactProxy=${{ matrix.artifact_proxy }}
```
nsingla: can you replace the variables in the job name here to match what's actually changing so that we can know which job is testing what

VaniHaripriya: Updated the job name.

nsingla (commented):
/lgtm

alyssacgoins (approved):

PR #13196, opened 2026-04-02
chore(frontend): clean up NewRunV2 test boilerplate, validation contract, memo dependencies

### LINKED ISSUE #13146: frontend: NewRunV2 cleanup — test boilerplate, validation UX, loading state, and memo dependencies
## Summary

While reviewing #13140, several pre-existing issues in `NewRunV2` were identified that are worth addressing independently. None are correctness bugs, but they increase maintenance cost and leave minor UX gaps.

---

### 1. Test boilerplate repetition

**File:** `frontend/src/pages/NewRunV2.test.tsx`

Every test renders `<NewRunV2 />` with a near-identical ~15-line props block. There are 30+ copies of this pattern, and each new test adds another. A shared helper would cut hundreds of lines and lower the barrier to adding coverage.

**Suggested fix:** Extract a helper like:

```tsx
function renderNewRunV2(overrides?: Partial<NewRunV2Props>) {
  const props = {
    ...generatePropsNewRun(),
    existingRunId: null,
    existingRun: undefined,
    existingRecurringRunId: null,
    existingRecurringRun: undefined,
    existingPipeline: ORIGINAL_TEST_PIPELINE,
    handlePipelineIdChange: vi.fn(),
    existingPipelineVersion: ORIGINAL_TEST_PIPELINE_VERSION,
    handlePipelineVersionIdChange: vi.fn(),
    templateString: v2XGYamlTemplateString,
    chosenExperiment: undefined,
    ...overrides,
  };
  return render(
    <CommonTestWrapper>
      <NewRunV2 {...props} />
    </CommonTestWrapper>,
  );
}
```

Then each test becomes a one-liner:

```tsx
renderNewRunV2({ existingPipeline: undefined, existingPipelineVersion: undefined });
```

**Impact:** No functional change; purely maintenance / developer experience.

---

### 2. No user-facing message when template is loading

**File:** `frontend/src/pages/NewRunV2.tsx`

When `templateString` is not yet available but `existingPipeline` and `existingPipelineVersion` are both selected, the Start button is correctly disabled, but the validation error area shows no message. The user sees a disabled button with no explanation.

**Suggested fix:** Return a message like `'Loading pipeline template...'` from the validation logic when the template hasn't arrived yet but pipeline/version are selected.

---

### 3. No loading indicator between pipeline version selection and template arrival

**File:** `frontend/src/pages/NewRunV2.tsx`

After selecting a pipeline version, there is no visual feedback (spinner, skeleton, etc.) while the parent component fetches `templateString`. The form appears static with a disabled Start button.

**Suggested fix:** Add a lightweight loading indicator (e.g., a `CircularProgress` near the Start button or in the parameters section) when a pipeline version is selected but `templateString` is still undefined.

---

### 4. `useMemo` / `useEffect` dependencies use full object references

**File:** `frontend/src/pages/NewRunV2.tsx`

Several hooks depend on entire objects (`existingRun`, `existingRecurringRun`, `existingPipelineVersion`) rather than the specific fields they read (e.g., `.display_name`). If the parent recreates these objects on each render (common with API response mapping), hooks re-fire unnecessarily. For `useMemo` with `generateRandomString` inside, this also regenerates the random suffix in the default run name.

**Suggested fix:** Depend on the specific fields:

```tsx
const defaultRunName = useMemo(
  () => getDefaultRunName(existingRun, existingRecurringRun, existingPipelineVersion),
  [existingRun?.display_name, existingRecurringRun?.display_name, existingPipelineVersion?.display_name],
);
```

---

### Priority suggestion

| Item | Effort | Value |
|------|--------|-------|
| 1. Test helper | Small | High (compounds over time) |
| 2. Validation message | Small | Medium (UX clarity) |
| 3. Loading indicator | Small–Medium | Medium (UX polish) |
| 4. Memo dependencies | Small | Low (micro-optimization) |

Item 1 is a good candidate for a first contribution. Items 2–3 could be addressed together in a small UX follow-up. Item 4 is only worth doing if there's evidence of unnecessary re-renders in practice.

manaswinidas:

Addresses items from #13146 with refinements based on review feedback:

1. **Extract test helpers** — `buildNewRunV2Element`/`renderNewRunV2` eliminate ~30 repeated 15-line render blocks in `NewRunV2.test.tsx`.
2. **Add clone-mode test helpers** — `buildClonedRunV2Element`/`renderClonedRunV2` use `generatePropsClonedRun()` as the PageProps base directly, making the override chain explicit instead of silently clobbering the `generatePropsNewRun()` base from `buildNewRunV2Element`. All 5 existing clone tests migrated.
3. **Clean up validation contract** — `getRunValidationErrorMessage` returns `''` (not a loading message) when the template is unavailable; the Start button stays disabled via the existing `!!templateString` guard in `isStartButtonEnabled`. This keeps the validation function's contract clean: empty = valid, non-empty = actual validation error.
4. **Remove misleading loading UX** — Removed `isTemplateLoading`, `CircularProgress` spinner, and "Loading pipeline template..." message. `NewRunSwitcher` already blocks rendering with "Currently loading pipeline information" while fetching, so `NewRunV2` never sees a true template-loading state — showing a spinner for what is actually a post-fetch-failure state would mislead users indefinitely.
5. **Narrow memo/effect dependencies** — `useMemo`/`useEffect` dependency arrays use specific fields (e.g. `?.display_name`) instead of full object references to avoid unnecessary re-fires when the parent recreates objects.
6. **Add clone-mode spinner test** — Verifies no spinner appears when cloning a run with `templateString: undefined`.

Closes #13146

**Checklist:**

### DIFF
--- a/frontend/src/pages/NewRunV2.test.tsx
+++ b/frontend/src/pages/NewRunV2.test.tsx
@@ -248,6 +248,35 @@ describe('NewRunV2', () => {
     return render(buildNewRunV2Element(overrides));
   }
 
+  // Clone-mode helpers: use generatePropsClonedRun() as the PageProps base and
+  // default existingPipeline/existingPipelineVersion to undefined, matching how
+  // clone runs work (pipeline info comes from the original run, not the URL).
+  function buildClonedRunV2Element(overrides: Partial<ComponentProps<typeof NewRunV2>> = {}) {
+    const props: ComponentProps<typeof NewRunV2> = {
+      ...generatePropsClonedRun(),
+      existingRunId: null,
+      existingRun: undefined,
+      existingRecurringRunId: null,
+      existingRecurringRun: undefined,
+      existingPipeline: undefined,
+      handlePipelineIdChange: vi.fn(),
+      existingPipelineVersion: undefined,
+      handlePipelineVersionIdChange: vi.fn(),
+      templateString: v2XGYamlTemplateString,
+      chosenExperiment: undefined,
+      ...overrides,
+    };
+    return (
+      <CommonTestWrapper>
+        <NewRunV2 {...props} />
+      </CommonTestWrapper>
+    );
+  }
+
+  function renderClonedRunV2(overrides: Partial<ComponentProps<typeof NewRunV2>> = {}) {
+    return render(buildClonedRunV2Element(overrides));
+  }
+
   // For creating new run with no pipeline is selected (enter from run list)
   function generatePropsNoPipelineDef(eid: string | null): PageProps {
     return {
@@ -421,26 +450,22 @@ describe('NewRunV2', () => {
     expect(await screen.findByText('This pipeline has no parameters')).toBeInTheDocument();
   });
 
-  it('shows a loading message when pipeline version is selected but template has not arrived', async () => {
+  it('disables start button when pipeline version is selected but template is unavailable', async () => {
     renderNewRunV2({ templateString: undefined });
-
-    expect(await screen.findByText('Loading pipeline template...')).toBeInTheDocument();
 
     const startButton = screen.getByText('Start');
     expect(startButton.closest('button')?.disabled).toEqual(true);
   });
 
-  it('shows a loading spinner when template is loading for a new run', async () => {
-    renderNewRunV2({ templateString: undefined });
-
-    expect(screen.getByRole('progressbar')).toBeInTheDocument();
-  });
-
-  it('does not show the loading spinner once the template has arrived', async () => {
-    renderNewRunV2();
+  it('does not show a loading spinner when cloning a run with an undefined template', async () => {
+    renderClonedRunV2({
+      existingRunId: TEST_RUN_ID,
+      existingRun: API_UI_CREATED_NEW_RUN_DETAILS,
+      existingPipelineVersion: ORIGINAL_TEST_PIPELINE_VERSION,
+      templateString: undefined,
+    });
 
     expect(screen.queryByRole('progressbar')).toBeNull();
-    expect(screen.queryByText('Loading pipeline template...')).toBeNull();
   });
 
   it('shows the cloned pipeline root after rerendering into clone mode', async () => {
@@ -786,12 +811,9 @@ describe('NewRunV2', () => {
 
   describe('cloning an existing run', () => {
     it('only shows clone run name from original run', () => {
-      renderNewRunV2({
-        ...generatePropsClonedRun(),
+      renderClonedRunV2({
         existingRunId: TEST_RUN_ID,
         existingRun: API_UI_CREATED_NEW_RUN_DETAILS,
-        existingPipeline: undefined,
-        existingPipelineVersion: undefined,
       });
       screen.findByDisplayValue(`Clone of ${API_UI_CREATED_NEW_RUN_DETAILS.display_name}`);
     });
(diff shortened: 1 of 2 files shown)


jeffspahr:
I found one medium-severity frontend issue.

The new loading UI in `frontend/src/pages/NewRunV2.tsx` (`isTemplateLoading`, the spinner, and `Loading pipeline template...`) does not actually cover the normal "template is still loading" path. `frontend/src/pages/NewRunSwitcher.tsx` keeps the whole route on `Currently loading pipeline information` until `pipelineVersionIsFetching` / `v1TemplateStrIsFetching` finish, so `NewRunV2` only sees `templateString` missing after a failed or otherwise inconsistent lookup.

In practice that means this PR now shows `Loading pipeline template...` and a spinner indefinitely in cases that are no longer loading. The new tests in `frontend/src/pages/NewRunV2.test.tsx` only exercise that direct-prop state (`templateString: undefined` with a selected version), which the routed page does not expose during normal loading.

I think this UX needs to be driven from the real query loading state in `NewRunSwitcher`, or `NewRunV2` needs an explicit loading/error prop so it can distinguish between "still fetching" and "template lookup failed".

Local verification in `/tmp/pipelines-pr13196-review-ETmCHQ/frontend`:
- `fnm exec --using .nvmrc -- npm ci --ignore-scripts`
- `fnm exec --using .nvmrc -- npm run test:ui -- src/pages/NewRunV2.test.tsx src/pages/NewRunSwitcher.test.tsx` (47 tests passed)

jsonmp-k8 (commented):
## Review

Nice cleanup — the test helper extraction is well done and the net -483 lines is a big readability win.

A few things worth addressing:

### 1. Validation function conflates loading state with validation errors

```tsx
// NewRunV2.tsx:224 — previously returned ''
return 'Loading pipeline template...';
```

`getRunValidationErrorMessage` is a validation function, but now it returns a user-facing loading message. The caller treats any non-empty string as "invalid" (disables Start button), so this works mechanically, but it conflates two concerns. The `isTemplateLoading` boolean already exists separately — consider returning `''` from the validation function and using `isTemplateLoading` to independently control the button disabled state and show the loading message. Keeps the validation function's contract clean: empty = valid, non-empty = actual validation error.

Not blocking since it works, but worth considering.

### 2. Missing test for clone mode + undefined template

The PR says the spinner is "new-run flow only, not clone", and the code checks `!cloneOrigin.isClone`. But there's no test verifying the spinner is absent during clone mode with `templateString: undefined`. Something like:

```tsx
it('does not show the loading spinner when cloning a run with an undefined template', async () => {
  renderNewRunV2({
    ...generatePropsClonedRun(),
    existingRunId: TEST_RUN_ID,
    existingRun: API_UI_CREATED_NEW_RUN_DETAILS,
    existingPipelineVersion: ORIGINAL_TEST_PIPELINE_VERSION,
    templateString: undefined,
  });

  expect(screen.queryByRole('progressbar')).toBeNull();
  expect(screen.queryByText('Loading pipeline template...')).toBeNull();
});
```

### 3. `generatePropsClonedRun()` spread inside `renderNewRunV2` overrides silently

In clone tests:
```tsx
renderNewRunV2({
  ...generatePropsClonedRun(),
  existingRunId: TEST_RUN_ID,
  ...
});
```

`buildNewRunV2Element` already spreads `generatePropsNewRun()` as the base. Then `generatePropsClonedRun()` overrides some of those defaults when spread into the overrides object. This works today, but could cause confusion if someone adds a prop to `generatePropsNewRun()` that gets unexpectedly clobbered. Consider adding a dedicated `renderClonedRunV2` helper or at least a comment noting the override chain.

manaswinidas:
Addressed all the review comments @jsonmp-k8 @jeffspahr

PR #13201, opened 2026-04-03
test(backend): increase apiserver/client test coverage to 62.3%

jsonmp-k8:
## Summary

- Add unit tests for all untested fake client methods in `backend/src/apiserver/client/`, covering pod, scheduled workflow, and workflow fake implementations
- Fix a signature bug in `FakeSubjectAccessReviewClientError.Create` — it was missing `context.Context` and `v1.CreateOptions` parameters, making it incompatible with `SubjectAccessReviewInterface`
- Add compile-time interface compliance checks (`var _ Interface = Type{}`) for all three SAR fake types to prevent future signature drift

Coverage: **44.9% → 62.3%** (+17.4%)

## Test plan

- [x] `go test -v -cover ./backend/src/apiserver/client/` — all tests pass, 62.3% coverage
- [x] `golangci-lint run --new --new-from-merge-base=origin/master` — 0 issues
- [x] `gofmt` applied to all changed files

### DIFF
--- a/backend/src/apiserver/client/swf_fake_test.go
+++ b/backend/src/apiserver/client/swf_fake_test.go
@@ -127,8 +127,17 @@ func TestFakeSwfClient_Update(t *testing.T) {
 		t.Fatalf("setup: Create() unexpected error: %v", err)
 	}
 
-	created.Spec.MaxConcurrency = intPtr(5)
-	updated, err := swfClient.Update(ctx, created)
+	fetchedBeforeUpdate, err := swfClient.Get(ctx, "update-me", v1.GetOptions{})
+	if err != nil {
+		t.Fatalf("Get() before Update() unexpected error: %v", err)
+	}
+	if fetchedBeforeUpdate.Spec.MaxConcurrency != nil {
+		t.Fatal("Create() unexpectedly persisted MaxConcurrency before Update")
+	}
+
+	updatedInput := created.DeepCopy()
+	updatedInput.Spec.MaxConcurrency = intPtr(5)
+	updated, err := swfClient.Update(ctx, updatedInput)
 	if err != nil {
 		t.Fatalf("Update() unexpected error: %v", err)
 	}


backend/src/apiserver/client/swf_fake_test.go
```diff
@@ -114,6 +114,87 @@ func TestFakeSwfClientWithBadWorkflow_ScheduledWorkflow(t *testing.T) {
...
+	if updated.Name != "update-me" {
+		t.Errorf("Update() Name = %q, want %q", updated.Name, "update-me")
+	}
+
+	fetched, err := swfClient.Get(ctx, "update-me", v1.GetOptions{})
+	if err != nil {
+		t.Fatalf("Get() after Update() unexpected error: %v", err)
+	}
+	if fetched.Spec.MaxConcurrency == nil || *fetched.Spec.MaxConcurrency != 5 {
+		t.Errorf("Get() after Update() MaxConcurrency not persisted")
```
jeffspahr, line 153: Because `Create` stores the same `*ScheduledWorkflow` pointer in the fake map, mutating `created.Spec.MaxConcurrency` here changes the stored object before `Update` runs. That makes the final `Get` assertion pass even if `Update` is a no-op. Using a fresh copy and checking the pre-update state makes this test actually cover the update path.

```suggestion
	created, err := swfClient.Create(ctx, swf)
	if err != nil {
		t.Fatalf("setup: Create() unexpected error: %v", err)
	}

	fetchedBeforeUpdate, err := swfClient.Get(ctx, "update-me", v1.GetOptions{})
	if err != nil {
		t.Fatalf("Get() before Update() unexpected error: %v", err)
	}
	if fetchedBeforeUpdate.Spec.MaxConcurrency != nil {
		t.Fatal("Create() unexpectedly persisted MaxConcurrency before Update")
	}

	updatedInput := created.DeepCopy()
	updatedInput.Spec.MaxConcurrency = intPtr(5)
	updated, err := swfClient.Update(ctx, updatedInput)
	if err != nil {
		t.Fatalf("Update() unexpected error: %v", err)
	}
	if updated.Name != "update-me" {
		t.Errorf("Update() Name = %q, want %q", updated.Name, "update-me")
	}

	fetched, err := swfClient.Get(ctx, "update-me", v1.GetOptions{})
	if err != nil {
		t.Fatalf("Get() after Update() unexpected error: %v", err)
	}
	if fetched.Spec.MaxConcurrency == nil || *fetched.Spec.MaxConcurrency != 5 {
		t.Errorf("Get() after Update() MaxConcurrency not persisted")
	}
```

jeffspahr (changes requested):
The new fake-client coverage looks good overall, but the scheduled workflow update test still misses the behavior it is supposed to protect. The inline suggestion below makes the test verify that `Update` actually persists the change instead of mutating the same object pointer that `Create` already stored.

PR #13207, opened 2026-04-03
fix(ci): gate ci-passed on real CI poll completion

jeffspahr:
## Summary
- make `CI Check` publish an explicit `ci-check-result` artifact for every run, including no-op success cases
- make `Add CI Passed Label` consume that artifact so it only adds `ci-passed` when the triggering run actually completed the CI polling gate successfully
- preserve the existing stale-label guards by re-checking the current PR head SHA and eligibility labels before any label add

## Root cause
- #13095 changed `Add CI Passed Label` to infer add eligibility from the PR's current labels during `workflow_run`
- #13172 changed irrelevant label events from skipped jobs to no-op successful jobs so Tide would keep the helper contexts green
- together, that let older no-op-success `CI Check` runs re-evaluate as eligible after `ok-to-test` was added and incorrectly add `ci-passed` before real CI finished

## Validation
- `ruby -e 'require "yaml"; [".github/workflows/ci-checks.yml", ".github/workflows/add-ci-passed-label.yml"].each { |path| YAML.load_file(path); puts "YAML OK: #{path}" }'`
- `go run github.com/rhysd/actionlint/cmd/actionlint@latest .github/workflows/ci-checks.yml .github/workflows/add-ci-passed-label.yml`
- `git diff --check`

## Note
- This PR itself will still show the current helper-workflow behavior until the fix lands, because `pull_request_target` and downstream `workflow_run` executions use the workflow files from the base branch, not from this PR branch.

### DIFF
--- a/.github/workflows/add-ci-passed-label.yml
+++ b/.github/workflows/add-ci-passed-label.yml
@@ -35,8 +35,7 @@ jobs:
         uses: actions/github-script@v8
         with:
           script: |
-            core.setOutput("eligible", "false");
-            core.setOutput("should_add_label", "false");
+            core.setOutput("artifact_found", "false");
 
             if (context.eventName !== "workflow_run") {
               core.info(`No ci-passed eligibility lookup needed for ${context.eventName}.`);
@@ -60,9 +59,10 @@ jobs:
             });
 
             if (!resultArtifact) {
-              throw new Error(
-                `Required artifact "ci-check-result" was not found for workflow run ID ${context.payload.workflow_run.id}.`
+              core.warning(
+                `Required artifact "ci-check-result" was not found for workflow run ID ${context.payload.workflow_run.id}; skipping ci-passed update.`
               );
+              return;
             }
 
             const download = await github.rest.actions.downloadArtifact({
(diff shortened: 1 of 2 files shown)


jsonmp-k8:
Two issues worth addressing:

**1. `should_poll` could produce invalid JSON if empty**

In the `Record CI check result` step:
```bash
"should_poll": ${should_poll},
```

`should_poll` is unquoted to produce a JSON boolean — that's correct when it's `true` or `false`. But if `steps.eligibility.outputs.should_poll` is ever empty (e.g., the eligibility step is skipped or errors), `${should_poll}` expands to nothing, yielding `"should_poll": ,` — invalid JSON that will break the consumer's `JSON.parse()`.

There's already a defensive default for `poll_outcome`:
```bash
if [[ -z "${poll_outcome}" ]]; then
  poll_outcome=skipped
fi
```

Add the same for `should_poll`:
```bash
if [[ -z "${should_poll}" ]]; then
  should_poll=false
fi
```

**2. Heredoc indentation becomes literal content in result.json**

```bash
          cat > ci-check-result/result.json <<EOF
          {
            "pr_number": "...",
```

The ~10 spaces of indentation inside the heredoc are literal whitespace in the output file. `JSON.parse()` handles it fine so this isn't broken, but it produces a messy artifact. Consider either dedenting the heredoc content or using `jq -n` to build the JSON.

jeffspahr:
Addressed the review feedback in `d60ba09f3`.

Changes made:
- In [`.github/workflows/ci-checks.yml`](https://github.com/kubeflow/pipelines/blob/d60ba09f32fcab208f4d4db598102b208755e9d7/.github/workflows/ci-checks.yml), `should_poll` now defaults to `false` before serialization, and the result file is generated with `jq -n` instead of an interpolated heredoc.
- In [`.github/workflows/add-ci-passed-label.yml`](https://github.com/kubeflow/pipelines/blob/d60ba09f32fcab208f4d4db598102b208755e9d7/.github/workflows/add-ci-passed-label.yml), the dead default outputs were removed from `download_result` and moved to the `ci_result` step, where they now act as the actual safe defaults.
- Missing `ci-check-result` artifacts now produce a warning and a no-op instead of a hard failure, and the unzip/parse steps are gated on `artifact_found == 'true'` so pre-merge in-flight runs do not create noisy red helper checks.

Verification rerun on the updated branch:
- `ruby -e 'require "yaml"; [".github/workflows/ci-checks.yml", ".github/workflows/add-ci-passed-label.yml"].each { |path| YAML.load_file(path); puts "YAML OK: #{path}" }'`
- `go run github.com/rhysd/actionlint/cmd/actionlint@latest .github/workflows/ci-checks.yml .github/workflows/add-ci-passed-label.yml`
- `git diff --check`

PR #13199, opened 2026-04-03
fix(frontend): prevent spinner during template refetch

RATNAPRADEEP:
Fixes unnecessary loading spinner during template string refetch in RunDetailsRouter.

The spinner now only depends on run data loading (runIsLoading), not templateStrIsLoading.

This ensures the UI does not show a loading state during background template refetch operations.

No unrelated changes included.

PR #13212, opened 2026-04-05
fix(frontend): prevent run details spinner on template refetch

jeffspahr:
Supersedes #13199.

This carries forward the original intent of the router fix and applies the minimal validated follow-through:
- scope the Run Details loading state to initial loads instead of background refetches
- restore the requested spinner loading UI in `RunDetails`
- add regression coverage for the cached-template refetch path

Credit to @RATNAPRADEEP for the original investigation and first pass in #13199.

Verification:
- `cd frontend && fnm exec --using .nvmrc -- npm run test:ui -- src/pages/RunDetailsRouter.test.tsx`
- `cd frontend && fnm exec --using .nvmrc -- npx prettier --check src/pages/RunDetailsRouter.tsx src/pages/RunDetails.tsx src/pages/RunDetailsRouter.test.tsx`
- `cd frontend && fnm exec --using .nvmrc -- npx eslint src/pages/RunDetailsRouter.tsx src/pages/RunDetails.tsx --no-ignore`

### DIFF
--- a/frontend/src/pages/RunDetailsRouter.test.tsx
+++ b/frontend/src/pages/RunDetailsRouter.test.tsx
@@ -15,7 +15,6 @@
  */
 
 import { act, render, screen, waitFor } from '@testing-library/react';
-import '@testing-library/jest-dom';
 import { QueryClientProvider } from '@tanstack/react-query';
 import * as JsYaml from 'js-yaml';
 import * as features from 'src/features';
--- a/frontend/src/pages/RunDetails.tsx
+++ b/frontend/src/pages/RunDetails.tsx
@@ -172,7 +172,7 @@ export const css = stylesheet({
     paddingLeft: 20,
   },
   loadingMessage: {
-    textAlign: 'center' as const,
+    textAlign: 'center',
     paddingTop: 40,
   },
 });


RATNAPRADEEP:
Thanks for carrying this forward! Glad the initial investigation helped. 
Happy to contribute further if needed.

frontend/src/pages/RunDetailsRouter.tsx
```diff
@@ -36,7 +36,7 @@ export default function RunDetailsRouter(
   // Retrieves v2 run detail.
   const {
     isSuccess: getV2RunSuccess,
-    isFetching: runIsFetching,
+    isLoading: runIsLoading,
```
manaswinidas, line 39: `RecurringRunDetailsRouter` shares the same `usePipelineVersionTemplate` hook and still uses `isFetching` on both its own query (line 41) and the template query (line 64). It will have the same spinner-on-refetch bug. Could you apply the same fix there (or file a follow-up issue)?

frontend/src/pages/RunDetailsRouter.test.tsx
```diff
@@ -165,6 +170,63 @@ describe('RunDetailsRouter', () => {
...
+    };
+    const pipelineVersion: V2beta1PipelineVersion = {
+      pipeline_id: TEST_PIPELINE_ID,
+      pipeline_version_id: TEST_PIPELINE_VERSION_ID,
+      pipeline_spec: argoWorkflow,
+    };
+    const wrapper = (props: { children: React.ReactElement }) => (
+      <QueryClientProvider client={queryClientTest}>{props.children}</QueryClientProvider>
+    );
+
```
manaswinidas, line 199: This test uses a raw `QueryClientProvider` + `queryClientTest` while every other test in the file uses `CommonTestWrapper`. The inconsistency means this test runs against a different query client config than the rest of the suite. Consider adding a brief comment explaining why the separate wrapper is needed (you need a handle to `invalidateQueries`), or unifying all tests to use `queryClientTest`.

frontend/src/pages/RunDetailsRouter.test.tsx
```diff
@@ -77,6 +81,7 @@ describe('RunDetailsRouter', () => {
   });
 
   afterEach(() => {
+    queryClientTest.clear();
```
manaswinidas: nit: `queryClientTest.clear()` is only meaningful for the one new test that uses `queryClientTest` — the other tests render via `CommonTestWrapper` with its own internal client. Consider scoping this cleanup to a nested `describe` to keep the intent clear.

frontend/src/pages/RunDetailsRouter.test.tsx
```diff
@@ -14,17 +14,21 @@
...
 import * as features from 'src/features';
 import { CommonTestWrapper } from 'src/TestWrapper';
 import { RouteParams } from 'src/components/Router';
+import { queryKeys } from 'src/hooks/queryKeys';
 import { Apis } from 'src/lib/Apis';
+import { queryClientTest } from 'src/TestUtils';
 import { V2beta1Run } from 'src/apisv2beta1/run';
 import { V2beta1PipelineVersion } from 'src/apisv2beta1/pipeline';
 import RunDetailsRouter from './RunDetailsRouter';
 import v2YamlTemplateString from 'src/data/test/lightweight_python_functions_v2_pipeline_rev.yaml?raw';
```
manaswinidas, line 29: nit: `expect` is a Vitest global (already used without import in the existing tests). The explicit import is unnecessary.

manaswinidas (commented):
Nice fix — the `isFetching` → `isLoading` swap is the right call for immutable cached data (`staleTime: Infinity`). A few comments below.

jeffspahr:
Followed up on the review comments.

Addressed in the latest commits:
- cleaned up the `RunDetailsRouter.test.tsx` nits around the explicit `expect` import, `queryClientTest.clear()` scope, and the dedicated query-client wrapper comment
- applied the same initial-load-only fix to `RecurringRunDetailsRouter`
- added matching cached-template refetch regression coverage for recurring run details

Verification rerun after the follow-up:
- `cd frontend && fnm exec --using .nvmrc -- npm run test:ui -- src/pages/RunDetailsRouter.test.tsx src/pages/RecurringRunDetailsRouter.test.tsx src/pages/RecurringRunDetailsV2.test.tsx src/pages/functional_components/RecurringRunDetailsV2FC.test.tsx`
- `cd frontend && fnm exec --using .nvmrc -- npx prettier --check src/pages/RunDetailsRouter.test.tsx src/pages/RecurringRunDetailsRouter.tsx src/pages/RecurringRunDetailsRouter.test.tsx`
- `cd frontend && fnm exec --using .nvmrc -- npx eslint src/pages/RunDetailsRouter.tsx src/pages/RunDetails.tsx src/pages/RecurringRunDetailsRouter.tsx --no-ignore`

frontend/src/pages/RunDetailsRouter.tsx
jeffspahr: Addressed in d527c6352. RecurringRunDetailsRouter now uses isLoading for both the recurring-run query and the shared template query, and the branch includes matching cached-template refetch coverage in RecurringRunDetailsRouter.test.tsx.

frontend/src/pages/RunDetailsRouter.test.tsx
jeffspahr: Addressed in f1f05ef4e. The regression case now includes a short comment explaining why it uses a dedicated QueryClientProvider/queryClientTest wrapper, and the queryClientTest cleanup is scoped to the nested describe that owns the invalidation test.

frontend/src/pages/RunDetailsRouter.test.tsx
jeffspahr: Addressed in f1f05ef4e by moving queryClientTest.clear() into the nested describe that owns the invalidateQueries regression case.

frontend/src/pages/RunDetailsRouter.test.tsx
jeffspahr: Addressed in f1f05ef4e by removing the explicit expect import.

Issue #13191, opened 2026-04-01
[bug] Investigate and address E2E CI saturation causing metadata-grpc crash loops and Init:OOMKilled failures


jeffspahr:
## Summary

`master` E2E failures are not explained by a single product regression. The dominant failure family looks like CI cluster resource saturation / startup contention in the single-node Kind environment used by the GitHub Actions E2E jobs.

PR [#13179](https://github.com/kubeflow/pipelines/pull/13179) fixed two real issues:

- it increased MLMD client retry budget for `codes.Aborted` / `codes.Unavailable`
- it fixed the nil log-stream panic in `ReadPodLogs`

Those changes are useful, but they do not address the broader E2E instability. The same failure family is still present after `#13179` merged.

## Problem We Need To Solve

We need the E2E workflows to run reliably on `master` and PRs without depending on flaky retries or secondary harness failures.

The problem is broader than MLMD deadlocks:

- `metadata-grpc-deployment` repeatedly enters `CrashLoopBackOff`
- many workflow task pods end in `Init:OOMKilled`
- the scheduler reports `0/1 nodes are available: 1 Insufficient cpu`
- workflow-controller later reports `Pod failed before main container starts due to PodInitializing`
- launcher/wait containers exit with `137`

MLMD deadlocks are also real and show up in the same failing runs, but they appear to be one downstream symptom of an overloaded cluster rather than the sole root cause.

## Evidence

Relevant runs:

- Last green `master` E2E run before the bad period: [23638070877](https://github.com/kubeflow/pipelines/actions/runs/23638070877) on March 27, 2026
- First bad `master` E2E run after that: [23651967509](https://github.com/kubeflow/pipelines/actions/runs/23651967509) on March 27, 2026
- `master` E2E run for the merge commit of `#13179`: [23780118750](https://github.com/kubeflow/pipelines/actions/runs/23780118750) on March 31, 2026
- Another post-merge failing `master` run: [23808198402](https://github.com/kubeflow/pipelines/actions/runs/23808198402) on March 31, 2026

Observed failure signals across those runs:

- `metadata-grpc-deployment` `CrashLoopBackOff`
- task pods in `Init:OOMKilled`
- `0/1 nodes are available: 1 Insufficient cpu`
- `Pod failed before main container starts due to PodInitializing`
- `command terminated with exit code 137`
- `PutExecution failed: mysql_query aborted: Deadlock found when trying to get lock`

The failure family spans:

- `E2ECritical`
[...]

PR #13213, opened 2026-04-05
test(e2e): reduce pipeline_with_loops fanout

jeffspahr:
## Summary
- address the remaining genuine saturation case in `E2EParallelNested`
- reduce fanout inside `parallel_and_nested/pipeline_with_loops` instead of changing global E2E parallelism again
- regenerate the matching compiled IR and Argo golden files

## Root Cause
Follow-up from #13191.

On current `master`, `E2EParallelNested` is already forced to a single Ginkgo node in `.github/workflows/e2e-test.yml`, so the remaining flake is not caused by multiple specs running at once.

The latest failing `master` runs still showed `metadata-grpc` `CrashLoopBackOff`, `Init:OOMKilled`, `PodInitializing`, and launcher `exit 137` in the `E2EParallelNested` lane. Parsing those logs narrowed the persistent offender to `test_data/sdk_compiled_pipelines/valid/parallel_and_nested/pipeline_with_loops.yaml`:
- in run `23957264553` (`argo=v3.5.14`), `pipeline_with_loops.yaml` was the only failing fixture in the lane
- in run `23962218937` (`argo=v3.7.3`), `loop_consume_upstream.yaml` failed once, but `pipeline_with_loops.yaml` still failed repeatedly afterward

That fixture creates three independent `ParallelFor` sections, and two of them also fan out multiple sibling tasks per iteration. The resulting workflow can burst into roughly a dozen-plus component pods even though the test only needs to validate that loop handling works end to end.

## Fix
- add `parallelism=1` to each `ParallelFor` in `pipeline_with_loops.py`
- serialize the sibling tasks in the second and third loops with `.after(...)`
- regenerate:
  - `test_data/sdk_compiled_pipelines/valid/parallel_and_nested/pipeline_with_loops.yaml`
  - `test_data/compiled-workflows/pipeline_with_loops.yaml`

This keeps the nested/loop coverage in the lane, but removes the unnecessary burstiness that was saturating the single-node Kind cluster.

## Verification
- `python3 -m py_compile test_data/sdk_compiled_pipelines/valid/parallel_and_nested/pipeline_with_loops.py`
- `go test ./backend/test/compiler -run TestCompilation -args -ginkgo.focus='pipeline_with_loops.yaml' -updateCompiledFiles=true`
- `go test ./backend/test/compiler -run TestCompilation -args -ginkgo.focus='pipeline_with_loops.yaml'`
- `git diff --check`

I did not rerun the full GitHub `E2EParallelNested` workflow locally.

### DIFF
--- a/test_data/compiled-workflows/pipeline_with_loops.yaml
+++ b/test_data/compiled-workflows/pipeline_with_loops.yaml
@@ -48,11 +48,11 @@ spec:
         kfp\nfrom kfp import dsl\nfrom kfp.dsl import *\nfrom typing import *\n\ndef
         print_struct(struct: Dict):\n    print(struct)\n\n"],"image":"python:3.11"}'
     - name: components-comp-for-loop-2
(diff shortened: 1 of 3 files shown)


jeffspahr:
Closing as test flakiness was resolved with:
https://github.com/kubeflow/pipelines/pull/13221
https://github.com/kubeflow/pipelines/pull/13223

PR #13227, opened 2026-04-07
ci: preload Kind images instead of pushing via local registry

jeffspahr:
## Summary
- preload CI images directly into the Kind node instead of pushing them through the local registry
- retag `driver` and `launcher` to `:ci` and point the API server at those refs so runtime-created pods also use preloaded images
- remove the now-unused local registry startup from the shared `create-cluster` action

## Why
The current deploy path spends a large chunk of `Deploy KFP` moving already-built tarball artifacts through a temporary local registry:
- `docker load`
- retag to `localhost:5000/...`
- `docker push`
- remove the temporary tags

The image-build workflow already produces tarballs tagged as `kind-registry:5000/<image>:latest`, so the control-plane images can be loaded directly into the Kind node with `kind load image-archive`.

The only extra compatibility work needed is for runtime-created `driver` and `launcher` pods. Those refs came from the API server env vars and previously used untagged `kind-registry:5000/driver` / `launcher`, which fall back to `latest` pull semantics. This PR changes them to `:ci`, loads those images directly into Kind, and adds CI-only `IfNotPresent` patches for the remaining control-plane deployments that previously relied on `Always`.

This keeps the image names used by CI manifests and runtime pods consistent while removing the registry push path.

## Expected impact
From recent green E2E timing analysis, the first safe slice of direct Kind loading removed about 91 seconds of registry-push / cleanup time per affected lane. This PR removes the remaining push path as well, so the main expected win is a faster `Deploy KFP` step across the Kind-based CI workflows.

## What changed
- `.github/actions/deploy/action.yml`
  - replace the shared `docker load -> tag -> push localhost:5000` path with:
    - `kind load image-archive` for control-plane images
    - `docker load` + retag to `:ci` + `kind load docker-image` for `driver` and `launcher`
  - keep preloading the runtime base images used by E2E tasks (`python:3.11`, `registry.access.redhat.com/ubi9/python-311:latest`)
- `.github/actions/create-cluster/action.yml`
  - stop starting the local registry in the shared Kind cluster setup
- `.github/resources/manifests/base/apiserver-env.yaml`
  - point `V2_DRIVER_IMAGE` and `V2_LAUNCHER_IMAGE` at `kind-registry:5000/<name>:ci`
- `.github/resources/manifests/.../kustomization.yaml`
  - add CI-only `IfNotPresent` pull-policy patches for `metadata-writer`, `viewer-crd-controller`, `cache-deployer`, `cache-server`, and `metadata-envoy`

## Verification
- `ruby -e "require 'yaml'; ..."` on the edited actions and manifests
- `go run github.com/rhysd/actionlint/cmd/actionlint@latest` across the workflows that use `create-cluster` / `deploy`
- `kubectl kustomize --load-restrictor LoadRestrictionsNone .github/resources/manifests/standalone/default`
- `kubectl kustomize --load-restrictor LoadRestrictionsNone .github/resources/manifests/multiuser/artifact-proxy`
- `git diff --check`

I did not run a live Docker-backed Kind deploy locally because Docker is not available in this environment right now.

### DIFF
--- a/.github/actions/create-cluster/action.yml
+++ b/.github/actions/create-cluster/action.yml
@@ -43,7 +43,6 @@ runs:
         kubectl_version: ${{ inputs.k8s_version }}
         version: v0.31.0
         node_image: kindest/node:${{ inputs.k8s_version }}
-        registry: true
 
     - name: Retry Create k8s Kind Cluster
       uses: helm/kind-action@v1.13.0
@@ -53,4 +52,3 @@ runs:
         kubectl_version: ${{ inputs.k8s_version }}
         version: v0.31.0
         node_image: kindest/node:${{ inputs.k8s_version }}
-        registry: true
--- a/.github/actions/deploy/action.yml
+++ b/.github/actions/deploy/action.yml
@@ -62,18 +62,37 @@ runs:
         path: "images_${{ github.run_id }}"
         pattern: "!*.dockerbuild"
 
-    - name: Load Docker Images
+    - name: Load Control Plane Images Into Kind
       shell: bash
       run: |
-        APPS=("apiserver" "driver" "launcher" "scheduledworkflow" "persistenceagent" "frontend" "metadata-writer" "viewer-crd-controller" "visualization-server" "cache-deployer" "cache-server" "metadata-envoy"  )
+        # These images are deployed directly by the CI manifests and use
+        # IfNotPresent in the CI install path, so loading the image archive
+        # into the Kind node is sufficient.
+        APPS=("apiserver" "scheduledworkflow" "persistenceagent" "frontend" "metadata-writer" "viewer-crd-controller" "visualization-server" "cache-deployer" "cache-server" "metadata-envoy")
         for app in "${APPS[@]}"; do
-          docker image load -i ${{ inputs.image_path }}/$app/$app.tar
-          docker tag ${{ inputs.image_registry }}/$app:${{ inputs.image_tag }} localhost:5000/$app:${{ inputs.image_tag }}
-          docker push localhost:5000/$app:${{ inputs.image_tag }}
+          kind --name "${{ inputs.cluster_name }}" load image-archive "${{ inputs.image_path }}/$app/$app.tar"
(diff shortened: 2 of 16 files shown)


manaswinidas (approved):
/lgtm

jeffspahr:
Measured runtime impact for this change, using the PR E2E run [24068550357](https://github.com/kubeflow/pipelines/actions/runs/24068550357) on April 7, 2026 vs the last green `master` baseline run [24048515801](https://github.com/kubeflow/pipelines/actions/runs/24048515801) on April 6, 2026.

At the workflow level, `KFP E2E Pipeline tests` improved from `28m37s` to `25m49s`, which is about `2m48s` faster (`~10%`).

The speedup is coming from `Deploy KFP`, not from the test bodies themselves:
- All `18/18` comparable E2E lanes were faster overall.
- `Deploy KFP` was faster in every comparable lane by `2m40s` to `4m58s`, averaging `3m36s` faster.
- `Run Tests` was essentially flat, averaging `+11s` across lanes.
- `Create cluster` was also essentially flat, averaging `+11s`.

Representative examples:
- `E2ECritical v1.34 cache=false argo=v3.5.14`: `21m34s -> 16m43s`
  - `Deploy KFP`: `8m49s -> 5m11s`
  - `Run Tests`: `10m33s -> 9m53s`
- `E2EEssential v1.34 cache=false argo=v3.7.3`: `21m22s -> 16m49s`
  - `Deploy KFP`: `9m46s -> 4m48s`
  - `Run Tests`: `8m16s -> 8m37s`
- multi-user artifact-proxy lane: `24m52s -> 21m55s`
  - `Deploy KFP`: `10m21s -> 6m48s`
  - `Run Tests`: `11m29s -> 11m39s`

So the measured result matches the intent of the change: the PR reduces deploy/setup overhead materially, while test execution time stays basically unchanged.

droctothorpe:
Awesome!

/lgtm

hbelmiro (commented):
/lgtm
/approve

PR #13229, opened 2026-04-07
chore(frontend): enable React StrictMode in dev and tests. Fixes #12902

### LINKED ISSUE #12902: chore(frontend): Enable StrictMode in dev/test

manaswinidas:

Enable `<React.StrictMode>` in development builds and the Vitest test harness (`reactStrictMode: true`). Fix the production-side class-component lifecycle bugs that StrictMode's double-mount diagnostic surfaced, and update tests to work correctly with the new behavior.

Supersedes #13159 — addresses all review feedback from @jeffspahr (specifically: fix real effects / cleanup / deduplication, don't just loosen spy counts).

### Production fixes (cleanup, deduplication, idempotency)

**Core bug — `_isMounted` not reset on remount** (15 components):
`Page._isMounted` was set `true` once in the class-field initializer and set `false` in `componentWillUnmount`. Under StrictMode's unmount→remount cycle, it stayed `false` permanently — making every `setStateSafe` call a no-op. The component became a zombie that could never update its own state. Fixed by resetting `_isMounted = true` in `componentDidMount` across `Page` and all subclasses (`RunDetails`, `NewRun`, `ExperimentDetails`, `PipelineDetails`, `CompareV1`, `RecurringRunDetails`, `RecurringRunDetailsV2`, `ArtifactDetails`, `GettingStarted`, `NewPipelineVersion`, `NewExperiment`), plus non-Page components (`CustomTable`, `SideNav`, `RunList`, `ResourceSelector`).

**Reload deduplication in `CustomTable`**:
Added `_activeReloadGeneration` counter so only the latest `reload()` clears `isBusy`. Prevents overlapping reloads from leaving the table in a stale busy state.

**Raw `setState` → `setStateSafe`**:
`RecurringRunDetails.load()`, `NewPipelineVersion.componentDidMount`, and `NewExperiment.componentDidMount` used raw `this.setState` after `await`, bypassing the `_isMounted` guard. Changed to `this.setStateSafe`.

**`index.tsx`**: Wrap app in `<StrictMode>` only when `import.meta.env.DEV` is true. Production builds are unaffected.

**`CustomTable` React 19 types**: Converted `HeaderRowSelectionSection` / `BodyRowSelectionSection` from `React.FC` to explicit function components with `JSX.Element | null` returns for React 19 compatibility.

### Test infrastructure

- `vitest.setup.ts`: `configure({ reactStrictMode: true })` — RTL renders now run under StrictMode globally
- `vitest.setup.ts`: `/// <reference types="node" />` for `process.env.TZ` typing
- `vitest.setup.ts`: Pin default locale to `en-US` for `toLocaleString`/`toLocaleDateString`/`toLocaleTimeString` — only when caller omits locale; explicit locales are preserved (addresses @jeffspahr's locale-masking concern from #13159)
- `muiSnapshot.ts`: Added `normalizeReactUseIdAttrs` to stabilize MUI `aria-controls`/`id` values under double-render

### Test adjustments (addressing @jeffspahr's review)

Per the [review guidance](https://github.com/kubeflow/pipelines/pull/13159#discussion_r3020150699), test changes focus on making effects resilient rather than loosening spy counts:

- **`mockClear()` after mount settles** — interaction-driven assertions are isolated from mount behavior, keeping exact counts for user actions
- **`toHaveBeenCalledWith` / `toHaveBeenLastCalledWith`** — replaces fragile mount-time `toHaveBeenCalledTimes(1)` with argument-correctness checks
- **`mockImplementation` instead of `mockImplementationOnce`** — both StrictMode mounts get consistent mock data rather than one mount hitting a fallback
- **No `N→2N` count changes** — zero instances of doubling an expected count to accept the double-fire

### Results

- **1956 / 1956 tests pass** with StrictMode enabled
- **Lint** ✅, **TypeScript typecheck** ✅, **Prettier** ✅
- **55 files changed**: 17 production files + 29 test files + 5 snapshot files + 4 infra/utility files

Fixes #12902

**Checklist:**

### DIFF
--- a/frontend/src/pages/ArtifactDetails.tsx
+++ b/frontend/src/pages/ArtifactDetails.tsx
@@ -196,10 +196,14 @@ class ArtifactDetails extends Page<{}, ArtifactDetailsState> {
       if (version) {
         title += ` (version: ${version})`;
       }
+      if (!this._isMounted) {
+        return;
+      }
+
       this.props.updateToolbar({
         pageTitle: title,
       });
-      this.setState({ artifact, artifactType });
+      this.setStateSafe({ artifact, artifactType });
     } catch (err) {
       if (isServiceError(err)) {
         this.showPageError(serviceErrorToString(err));
--- a/frontend/src/pages/RecurringRunDetails.tsx
+++ b/frontend/src/pages/RecurringRunDetails.tsx
@@ -186,6 +186,10 @@ class RecurringRunDetails extends Page<{}, RecurringRunConfigState> {
     toolbarActions[ButtonKeys.ENABLE_RECURRING_RUN].disabled = !!run.enabled;
     toolbarActions[ButtonKeys.DISABLE_RECURRING_RUN].disabled = !run.enabled;
 
+    if (!this._isMounted) {
+      return;
+    }
+
     this.props.updateToolbar({ actions: toolbarActions, breadcrumbs, pageTitle });
 
     this.setStateSafe({ run });


PR3MM:
Looks good to me — no blockers. StrictMode in dev/tests is wired correctly, production stays off StrictMode, and the lifecycle fixes match what React 19 + double-mount need.

Ran locally: cd frontend && npm run test:ci — all UI tests (1956) and server tests passed; npm run build also succeeded (lint + typecheck + Vite build).

LGTM

frontend/src/pages/RecurringRunDetails.tsx
```diff
@@ -127,6 +127,7 @@ class RecurringRunDetails extends Page<{}, RecurringRunConfigState> {
   }
 
   public componentDidMount(): Promise<void> {
+    this._isMounted = true;
     return this.load();
```
jeffspahr, line 131: Unlike `RecurringRunDetailsV2`, this V1 path still updates the toolbar unconditionally after the awaited `getJob` / `getExperiment` work. If the user leaves the page before those requests resolve, the stale response can still overwrite the next page's toolbar. Mirror the V2 mounted guard before touching toolbar or state, for example:

```ts
const toolbarActions = this.props.toolbarProps.actions;
toolbarActions[ButtonKeys.ENABLE_RECURRING_RUN].disabled = !!run.enabled;
toolbarActions[ButtonKeys.DISABLE_RECURRING_RUN].disabled = !run.enabled;

if (!this._isMounted) {
  return;
}

this.props.updateToolbar({ actions: toolbarActions, breadcrumbs, pageTitle });

this.setStateSafe({ run });
```

frontend/src/pages/ArtifactDetails.tsx
```diff
@@ -102,6 +102,7 @@ class ArtifactDetails extends Page<{}, ArtifactDetailsState> {
   private api = Api.getInstance();
 
   public async componentDidMount(): Promise<void> {
+    this._isMounted = true;
```
jeffspahr, line 105: Anchoring here because this is the line that opts `ArtifactDetails` back into the StrictMode remount flow. `load()` still does an unconditional `updateToolbar(...)` and raw `this.setState(...)` after the awaited MLMD requests at lines 199-202, so navigating away before those promises resolve can still clobber the next page's toolbar and update an unmounted instance. Please add the same mounted guard / safe state update there, for example:

```ts
if (!this._isMounted) {
  return;
}

this.props.updateToolbar({
  pageTitle: title,
});
this.setStateSafe({ artifact, artifactType });
```

frontend/src/pages/RecurringRunDetails.tsx
manaswinidas: Done in 7c78a8e — added the `_isMounted` guard before `updateToolbar` and `setStateSafe` in `RecurringRunDetails.load()`, mirroring the V2 pattern. Full test suite (1956/1956) passes.

frontend/src/pages/ArtifactDetails.tsx
manaswinidas: Done in 7c78a8e — added the `_isMounted` guard before `updateToolbar` and switched `this.setState` to `this.setStateSafe` in `ArtifactDetails.load()`. Full test suite (1956/1956) passes.

PR #13231, opened 2026-04-07
refactor(frontend): narrow NewRun parameter init effect

jeffspahr:
## Summary
- move `NewRun` parameter/default initialization into a shared pure helper
- seed `NewRunV2` parent state from that helper instead of waiting for child mount-time effect propagation
- narrow `NewRunParametersV2`'s remaining effect to the callback-only mount contract
- add a page-level regression proving default root parameters still submit correctly

## Why
`NewRunParametersV2` still had the broadest remaining effect in the `NewRun` flow. It was doing several jobs at once:
- deriving editable parameter strings from `specParameters` / `clonedRuntimeConfig`
- deriving initial validation state and missing-parameter errors
- pushing the real runtime parameter payload back up to `NewRunV2`

That made `NewRunV2` depend on a child mount effect for its initial `runtime_config.parameters` and `isParameterValid` state.

This PR moves the pure derivation into shared initialization logic so the page owns its initial runtime payload and validity directly. The child still keeps a small mount-time effect for `handleParameterChange` / `setIsValidInput`, because that callback contract is still covered by the standalone component tests and used by the current `NewRunV2` integration.

## What changed
- added `frontend/src/lib/NewRunParametersUtils.ts` for:
  - `convertInput`
  - `convertNonUserInputParamToString`
  - `generateInputValidationErrMsg`
  - `getInitialParameterState`
- `NewRunV2` now seeds keyed `runtimeParameters` and `isParameterValid` from `getInitialParameterState(...)`
- `NewRunParametersV2` now initializes `updatedParameters` and `errorMessages` from the same helper instead of setting them inside its effect
- the remaining effect in `NewRunParametersV2` is reduced to mount-time propagation of the already-derived initial `runtimeParameters` / validity
- parameter/error updates in `NewRunParametersV2` now use immutable object updates instead of mutating the previous map in place
- added a `NewRunV2` regression showing root default parameters from the lightweight pipeline template are still submitted when the user only fills the required `message` field

## Verification
- `cd frontend && fnm exec --using .nvmrc -- npx prettier@3.8.1 --check src/lib/NewRunParametersUtils.ts src/components/NewRunParametersV2.tsx src/pages/NewRunV2.tsx src/pages/NewRunV2.test.tsx`
- `cd frontend && fnm exec --using .nvmrc -- npm run test:ui -- src/components/NewRunParametersV2.test.tsx src/pages/NewRunV2.test.tsx`

Result: formatter passed, focused UI verification passed with `86` tests, and typecheck passed.

### DIFF
--- a/frontend/src/components/NewRunParametersV2.tsx
+++ b/frontend/src/components/NewRunParametersV2.tsx
@@ -34,6 +34,8 @@ import {
   convertInput,
   generateInputValidationErrMsg,
   getInitialParameterState,
+  protoMap,
+  type InitialParameterState,
   type ParameterErrorMessages,
   type RuntimeParameters,
   type SpecParameters,
@@ -72,24 +74,17 @@ interface NewRunParametersProps {
   // ComponentInputsSpec_ParameterSpec
   specParameters: SpecParameters;
   clonedRuntimeConfig?: PipelineSpecRuntimeConfig;
+  initialParameterState?: InitialParameterState;
   handlePipelineRootChange?: (pipelineRoot?: string) => void;
   handleParameterChange?: (parameters: RuntimeParameters) => void;
   setIsValidInput?: (isValid: boolean) => void;
 }
 
-const protoMap = new Map<string, string>([
-  ['NUMBER_DOUBLE', 'double'],
-  ['NUMBER_INTEGER', 'integer'],
-  ['STRING', 'string'],
-  ['BOOLEAN', 'boolean'],
-  ['LIST', 'list'],
-  ['STRUCT', 'dict'],
-]);
-
 function NewRunParametersV2(props: NewRunParametersProps) {
   const {
     specParameters,
     clonedRuntimeConfig,
+    initialParameterState: providedInitialParameterState,
     handlePipelineRootChange,
     handleParameterChange,
     setIsValidInput,
@@ -100,8 +95,10 @@ function NewRunParametersV2(props: NewRunParametersProps) {
     clonedPipelineRoot ?? props.pipelineRoot,
   );
   const initialParameterState = React.useMemo(
-    () => getInitialParameterState(specParameters, clonedRuntimeConfig),
-    [clonedRuntimeConfig, specParameters],
+    () =>
+      providedInitialParameterState ??
+      getInitialParameterState(specParameters, clonedRuntimeConfig),
+    [clonedRuntimeConfig, providedInitialParameterState, specParameters],
   );
   const [errorMessages, setErrorMessages] = useState<ParameterErrorMessages>(
     initialParameterState.errorMessages,
@@ -111,6 +108,11 @@ function NewRunParametersV2(props: NewRunParametersProps) {
   );
 
   useEffect(() => {
+    // NewRunV2 already seeds parent state before render, so only preserve the
+    // standalone mount-time callback behavior for callers that do not provide it.
+    if (providedInitialParameterState) {
+      return;
+    }
     if (setIsValidInput) {
       setIsValidInput(initialParameterState.isValid);
     }
@@ -122,6 +124,7 @@ function NewRunParametersV2(props: NewRunParametersProps) {
     handleParameterChange,
     initialParameterState.isValid,
     initialParameterState.runtimeParameters,
+    providedInitialParameterState,
     setIsValidInput,
   ]);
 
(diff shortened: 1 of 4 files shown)


frontend/src/lib/NewRunParametersUtils.ts
```diff
@@ -0,0 +1,187 @@
...
+  runtimeParameters: RuntimeParameters;
+  updatedParameters: RuntimeParameters;
+};
+
+type ClonedRuntimeConfig = {
+  parameters?: RuntimeParameters;
+  pipeline_root?: string;
+};
+
+const protoMap = new Map<string, string>([
```
manaswinidas: `protoMap` is now defined in two places — here and at `NewRunParametersV2.tsx:73`. Export it from this module and import it in the component to keep a single source of truth.

```suggestion
export const protoMap = new Map<string, string>([
```

frontend/src/components/NewRunParametersV2.tsx
```diff
@@ -192,67 +99,31 @@ function NewRunParametersV2(props: NewRunParametersProps) {
   const [customPipelineRoot, setCustomPipelineRoot] = useState(
     clonedPipelineRoot ?? props.pipelineRoot,
   );
-  const [errorMessages, setErrorMessages] = useState<string[]>([]);
+  const initialParameterState = React.useMemo(
+    () => getInitialParameterState(specParameters, clonedRuntimeConfig),
+    [clonedRuntimeConfig, specParameters],
+  );
```
manaswinidas, line 102: This re-derives the same initial state that `NewRunV2` already computes via an identical `useMemo` (line 336). Both parent and child run `getInitialParameterState` on every `specParameters`/`clonedRuntimeConfig` change.

Consider receiving the computed state as a prop to avoid the duplicate work — that would also let you drop the `protoMap` duplicate from this file.

frontend/src/components/NewRunParametersV2.tsx
```diff
@@ -192,67 +99,31 @@ function NewRunParametersV2(props: NewRunParametersProps) {
...
-      handleParameterChange(runtimeParametersInRealType);
+      handleParameterChange(initialParameterState.runtimeParameters);
     }
-  }, [clonedRuntimeConfig, specParameters, handleParameterChange, setIsValidInput]);
+  }, [
+    handleParameterChange,
+    initialParameterState.isValid,
+    initialParameterState.runtimeParameters,
+    setIsValidInput,
+  ]);
```
manaswinidas, line 129: With the parent now seeding `runtimeParameters` and `isParameterValid` from `getInitialParameterState`, this effect overwrites them with the same values on mount, triggering a wasted render cycle.

If the child receives initial state via props (removing the `useMemo` above), this entire effect can be dropped.

frontend/src/pages/NewRunV2.test.tsx
```diff
@@ -565,6 +565,43 @@ describe('NewRunV2', () => {
         );
       });
     }, 20000);
+
+    it('submits root default parameters without waiting for child effect initialization', async () => {
```
manaswinidas, line 569: The child's mount effect still calls `handleParameterChange(initialParameterState.runtimeParameters)` before the click is processed, so this test passes regardless of whether the parent seeds the state. It can't distinguish "parent seeded synchronously" from "child effect propagated before click."

To make this meaningful, either remove the child's redundant propagation effect (which would also fix the duplicate computation above) or mock it away in this test.

frontend/src/components/NewRunParametersV2.tsx
jeffspahr: Addressed in `19f37ff1a`. `NewRunV2` now passes its computed `initialParameterState` down to `NewRunParametersV2`, and the child reuses that provided state instead of recomputing it for the page path.

frontend/src/lib/NewRunParametersUtils.ts
jeffspahr: Addressed in `19f37ff1a`. `protoMap` now lives in `frontend/src/lib/NewRunParametersUtils.ts` and `NewRunParametersV2` imports it from there, so the page and component path use a single source of truth.

frontend/src/pages/NewRunV2.test.tsx
jeffspahr: Addressed in `19f37ff1a`. The page path no longer depends on the child mount effect, and I added a focused regression in `NewRunParametersV2.test.tsx` proving that provided initial state renders without firing the mount callbacks. That makes the existing `NewRunV2` regression meaningful for the parent-seeded path.

frontend/src/components/NewRunParametersV2.tsx
jeffspahr: Addressed in `19f37ff1a`. When `initialParameterState` is provided by `NewRunV2`, the child skips the mount-time propagation effect entirely, so it no longer overwrites already-seeded parent state on mount. The fallback mount behavior remains only for standalone callers/tests that do not provide initial state.

PR #13245, opened 2026-04-09
fix(frontend): add root ErrorBoundary and loading indicators

manaswinidas:

Users can hit a white screen when a render-time error occurs because there is no root-level error boundary. Additionally, two components (`RecurringRunDetailsV2FC` and `InputOutputTab`) show blank content while their data queries are in flight, with no loading feedback.

### Changes

**`index.tsx`**
- Wrap `<Router />` with `<ErrorBoundary>` so unhandled render-time throws display a styled error Banner instead of a blank white page.

**`ErrorBoundary.tsx`**
- Upgrade the fallback from raw `<div>/<h2>/<details>` HTML to the existing `<Banner>` component with `mode='error'`, matching the visual style used throughout the application.

**`RecurringRunDetailsV2FC.tsx`**
- Destructure `isLoading` from the recurring run `useQuery`.
- Return `<CircularProgress />` while the query is in flight, consistent with the pattern used in `ExecutionDetails` and other pages.

**`InputOutputTab.tsx`**
- Destructure `isLoading` from the artifact `useQuery`.
- Render `<CircularProgress />` below the execution title while artifacts are being fetched.

### Tests added

- **`ErrorBoundary.test.tsx`** (new): verifies children render normally, and verifies the Banner fallback appears with error details when a child throws.
- **`RecurringRunDetailsV2FC.test.tsx`**: added test for loading spinner while recurring run is being fetched.
- **`InputOutputTab.test.tsx`**: added test for loading spinner while artifacts are being fetched.

**Checklist:**

- [x] The title for your pull request (PR) follows the title convention

### DIFF
--- a/frontend/src/index.tsx
+++ b/frontend/src/index.tsx
@@ -17,11 +17,11 @@
 // import './CSSReset';
 import 'src/build/tailwind.output.css';
 import '@xyflow/react/dist/style.css';
-import { StrictMode } from 'react';
+import React, { StrictMode } from 'react';
 import { ThemeProvider, StyledEngineProvider } from '@mui/material/styles';
 import { createRoot } from 'react-dom/client';
 import { QueryClient, QueryClientProvider } from '@tanstack/react-query';
-import { HashRouter } from 'react-router-dom';
+import { HashRouter, useLocation } from 'react-router-dom';
 import { cssRule } from 'typestyle';
 import { ErrorBoundary } from './atoms/ErrorBoundary';
 import Router from './components/Router';
@@ -36,6 +36,11 @@ import {
 } from './lib/KubeflowClient';
 import { BuildInfoProvider } from './lib/BuildInfo';
 // TODO: license headers
+
+function LocationKeyedErrorBoundary({ children }: React.PropsWithChildren) {
+  const location = useLocation();
+  return <ErrorBoundary key={location.key}>{children}</ErrorBoundary>;
+}
 
 if (KFP_FLAGS.DEPLOYMENT === Deployments.KUBEFLOW) {
   initKfClient();
@@ -61,9 +66,9 @@ const app = (
         <BuildInfoProvider>
           <GkeMetadataProvider>
             <HashRouter>
-              <ErrorBoundary>
+              <LocationKeyedErrorBoundary>
                 <Router />
-              </ErrorBoundary>
+              </LocationKeyedErrorBoundary>
             </HashRouter>
           </GkeMetadataProvider>
         </BuildInfoProvider>
--- a/frontend/src/index.tsx
+++ b/frontend/src/index.tsx
@@ -23,6 +23,7 @@ import { createRoot } from 'react-dom/client';
 import { QueryClient, QueryClientProvider } from '@tanstack/react-query';
 import { HashRouter } from 'react-router-dom';
 import { cssRule } from 'typestyle';
+import { ErrorBoundary } from './atoms/ErrorBoundary';
 import Router from './components/Router';
 import { fonts, theme } from './Css';
 import { initFeatures } from './features';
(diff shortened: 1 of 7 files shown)


frontend/src/index.tsx
```diff
@@ -60,7 +61,9 @@ const app = (
         <BuildInfoProvider>
           <GkeMetadataProvider>
             <HashRouter>
-              <Router />
+              <ErrorBoundary>
```
jeffspahr: This new app-level wrapper makes route crashes unrecoverable. `ErrorBoundary` latches `errorInfo` and never clears it, so once any routed page throws, the fallback replaces the entire router shell and the user cannot navigate away to recover. Because `RoutedPage` is keyed by `location.key`, route changes would normally remount only the page subtree, but that remount never happens once the root boundary has tripped.

Please either key this boundary by location so navigation remounts it, or move the boundary lower so route changes can recover without a full browser reload.

jeffspahr (changes requested):
One blocking issue remains on the current head.

Wrapping the whole router with the existing `ErrorBoundary` changes the failure mode from a white screen to a permanent app-level fallback, because this boundary never resets its latched error state. After any routed page throws once, the entire SPA stays on the banner until the browser is fully reloaded. The inline comment points at the new root wrapper.

I did not find other concrete behavior regressions in the loading-state changes.

frontend/src/pages/functional_components/RecurringRunDetailsV2FC.test.tsx
```diff
@@ -112,6 +112,24 @@ describe('RecurringRunDetailsV2FC', () => {
     });
   });
 
+  it('shows a loading spinner while the recurring run is being fetched', async () => {
```
jeffspahr, line 118: This new test no longer distinguishes the router's existing loading spinner from the inner `RecurringRunDetailsV2FC` spinner it intends to cover. If `waitFor(() => screen.getByRole('progressbar'))` resolves while `RecurringRunDetailsRouter` is still on its own loading branch, the second `mockReturnValueOnce(new Promise(() => {}))` is left queued on `getRecurringRunSpy`. Because `beforeEach` only calls `vi.clearAllMocks()`, that leftover one-time mock leaks into the next test and the file fails with the router stuck on `Currently loading recurring run information`.

I reproduced that with `fnm exec --using .nvmrc -- npm run test:ui -- src/pages/functional_components/RecurringRunDetailsV2FC.test.tsx`.

This should wait for the router loading state to clear before asserting the spinner, or fully reset the mock between tests.

jeffspahr (changes requested):
One blocking issue remains on the current head.

The earlier runtime blocker around the root `ErrorBoundary` is resolved: `index.tsx` now keys the boundary by `location.key`, and `ErrorBoundary.test.tsx` covers navigation recovery.

A new test regression remains in `RecurringRunDetailsV2FC.test.tsx`. The added loading-state test can pass while `RecurringRunDetailsRouter` is still showing its own spinner, which leaves the queued one-time mock on `getRecurringRunSpy` unconsumed. That leaked mock then breaks the next test in the file when the suite runs in order.

I reproduced this with:

`fnm exec --using .nvmrc -- npm run test:ui -- src/pages/functional_components/RecurringRunDetailsV2FC.test.tsx`

The inline comment points at the failing test.

jeffspahr (approved):
No findings on the current head.

The previous blocker in `RecurringRunDetailsV2FC.test.tsx` is resolved: the loading-state test now waits for `RecurringRunDetailsRouter`'s loading text to clear before asserting the inner spinner, and `beforeEach` now calls `getRecurringRunSpy.mockReset()` so leftover `mockOnce` state cannot leak into the next test.

Focused verification on `11b262b8afafadbb96dccd7742f80f8f49a8efb5` passed:
- `fnm exec --using .nvmrc -- npm ci`
- `fnm exec --using .nvmrc -- npm run test:ui -- src/pages/functional_components/RecurringRunDetailsV2FC.test.tsx`
- `fnm exec --using .nvmrc -- npm run test:ui -- src/atoms/ErrorBoundary.test.tsx src/components/tabs/InputOutputTab.test.tsx`

I did not rerun broader frontend CI beyond the changed-file checks above.

PR #13253, opened 2026-04-10
fix(test): Add `Eventually` for informer cache sync in pipeline version tests (#13312)

### LINKED ISSUE #13252: fix(test): Flaky pipeline version sorting tests due to K8s store informer cache lag
## Problem

The pipeline version sorting tests in `backend/test/v2/api/pipeline_api_test.go` are flaky when running against the Kubernetes-native pipeline store.

**Failing tests:**
- "Sort by name in ascending order" (line 673)
- "Sort by name in descending order" (line 702)

**Assertion that fails:**

```go
Expect(len(versions)).To(BeNumerically(">=", 2))
```

Returns `1` instead of `>= 2`.

## Root Cause

Both tests upload a second pipeline version and then immediately call `ListPipelineVersions`. With the Kubernetes-native store, the informer cache may not have synced yet, so the listing returns only the original version.

This is the same informer cache sync issue already addressed elsewhere in the same file:
- Line 79: list pipelines uses `Eventually` with a 30s timeout
- `uploadPipelineAndVerify` (in `pipeline_upload_api_test.go`, line 496) also uses `Eventually` when waiting for versions to appear

## Suggested Fix

Wrap the `ListPipelineVersions` call and the `>= 2` count assertion in an `Eventually` block, consistent with the existing pattern:

```go
var versions []*pipeline_model.V2beta1PipelineVersion
Eventually(func() int {
    versions, _, _, err = pipelineClient.ListPipelineVersions(...)
    if err != nil {
        return 0
    }
    return len(versions)
}, 30*time.Second, 2*time.Second).Should(BeNumerically(">=", 2),
    "Expected at least 2 pipeline versions after uploading a second version")
```

The SQL store returns immediately on the first attempt, so this adds no latency outside the K8s-native store.

## Evidence

CI failure: https://github.com/opendatahub-io/data-science-pipelines/actions/runs/24246276464/job/70822180608

The same job passes consistently on master.

/area test
/kind bug

hbelmiro:
Resolves https://github.com/kubeflow/pipelines/issues/13252.

- Updated test cases in `pipeline_api_test.go` to use `Eventually` for handling informer cache synchronization.
- Improved reliability of pipeline version listing and filtering assertions.

**Checklist:**

### DIFF
--- a/backend/test/v2/api/pipeline_api_test.go
+++ b/backend/test/v2/api/pipeline_api_test.go
@@ -59,6 +59,38 @@ func newListPipelinesParams() *pipeline_params.PipelineServiceListPipelinesParam
 	}
 }
 
+const (
+	informerSyncTimeout  = 30 * time.Second
+	informerSyncInterval = 2 * time.Second
+)
+
+type pipelineVersionListResult struct {
+	Versions      []*pipeline_model.V2beta1PipelineVersion
+	TotalSize     int
+	NextPageToken string
+}
+
+// eventuallyListPipelineVersions returns a poll function for Gomega's Eventually
+// that lists pipeline versions and yields totalSize. The last successful response
+// is stored in the returned result pointer so callers can inspect Versions,
+// TotalSize, and NextPageToken after Eventually succeeds.
+func eventuallyListPipelineVersions(
+	params *pipeline_params.PipelineServiceListPipelineVersionsParams,
+) (*pipelineVersionListResult, func() int) {
+	result := &pipelineVersionListResult{}
+	pollFn := func() int {
+		versions, totalSize, nextPageToken, err := pipelineClient.ListPipelineVersions(params)
+		if err != nil {
+			return 0
+		}
+		result.Versions = versions
+		result.TotalSize = totalSize
+		result.NextPageToken = nextPageToken
+		return totalSize
+	}
+	return result, pollFn
+}
+
 // ###########################################
 // ################## TESTS ##################
 // ###########################################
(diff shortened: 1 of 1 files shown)


backend/test/v2/api/pipeline_api_test.go
```diff
@@ -568,11 +568,20 @@ var _ = Describe("List Pipelines Versions API Tests >", Label(constants.POSITIVE
...
 			Expect(err).NotTo(HaveOccurred())
 
-			versions, totalSize, _, err := pipelineClient.ListPipelineVersions(&pipeline_params.PipelineServiceListPipelineVersionsParams{
-				PipelineID: createdPipeline.PipelineID,
-			})
+			// Use Eventually to allow the informer cache to sync (applies to K8s store;
+			// SQL store returns immediately on first attempt).
+			var versions []*pipeline_model.V2beta1PipelineVersion
+			var totalSize int
+			Eventually(func() int {
```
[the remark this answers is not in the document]
hbelmiro: Done.

backend/test/v2/api/pipeline_api_test.go
```diff
@@ -59,6 +59,38 @@ func newListPipelinesParams() *pipeline_params.PipelineServiceListPipelinesParam
...
+// TotalSize, and NextPageToken after Eventually succeeds.
+func eventuallyListPipelineVersions(
+	params *pipeline_params.PipelineServiceListPipelineVersionsParams,
+) (*pipelineVersionListResult, func() int) {
+	result := &pipelineVersionListResult{}
+	pollFn := func() int {
+		versions, totalSize, nextPageToken, err := pipelineClient.ListPipelineVersions(params)
+		if err != nil {
+			return 0
+		}
```
sduvvuri1603, line 88: Nit / test clarity: In a few places Eventually keeps retrying when ListPipelineVersions fails because the poll returns 0 (e.g. pipeline_api_test.go 83–85, 688–690), and when using GetSortedPipelineVersionsByCreatedAt failures become nil so we only see len == 0 until timeout (1606–1609, 1722–1725). 

That can look like “informer still catching up” when the real problem is a broken API call. Could we assert err inside Eventually (or stop with the real error) so true failures fail fast with a clear message?

hbelmiro: Done.

PR #13254, opened 2026-04-10
backend(proposal): adding proposal to extend and unify storage layer

nsingla:

**Checklist:**

### DIFF
--- a/proposals/storage-backend-abstraction/README.md
+++ b/proposals/storage-backend-abstraction/README.md
@@ -1,0 +1,592 @@
+# KEP: Extensible Multi-Backend Object Storage with Unified Storage Layer
+
+## Summary
+
+Kubeflow Pipelines needs to support additional storage backends (notably Azure Blob)
+without introducing repeated refactors across API server, driver, and launcher.
+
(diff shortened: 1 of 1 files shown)


Amr-Shams (commented):
i read the proposal, admired the idea and i have a genuine question,

go CDK already has a URL-based mux that dispatches by scheme, so why would adding another custom layer to register? while can be done with the custom url by go CDK?

nsingla:
> i read the proposal, admired the idea and i have a genuine question,
> 
> go CDK already has a URL-based mux that dispatches by scheme, so why would adding another custom layer to register? while can be done with the custom url by go CDK?

That's a good question that I probably should address in the proposal as well, but to answer your question:

The custom registry exists because Go CDK's URL mux is designed for stateless, environment-driven credential resolution, while KFP's credential model is request-scoped and depends on Kubernetes API calls that require injected runtime objects. The URL mux is still used as the fallback for the simpler FromEnv case.
Also, the intent of the change is to make the change with the least possible code change with no regression, and Go CDK provides that layer that is required to support this, but in order to do that, we need to bypass its mux to continue to use our own.

Issue #13228, opened 2026-04-07
chore(frontend): Modernize test patterns — remove Enzyme, querySelector, and waitFor anti-patterns


manaswinidas:
## Summary

The frontend test suite has accumulated legacy patterns that increase maintenance burden, reduce test reliability, and make it harder to write good new tests. This issue tracks incremental cleanup across three categories.

### 1. Enzyme wrappers (23 files)

These files still use `shallow()` or `mount()` from Enzyme. All new tests use Testing Library (`@testing-library/react`), and Enzyme is not compatible with React 19 without community adapters.

<details>
<summary>Files</summary>

- `src/components/Trigger.test.tsx`
- `src/pages/AllRunsList.test.tsx`
- `src/mlmd/LineageActionBar.test.tsx`
- `src/pages/PipelineVersionList.test.tsx`
- `src/pages/PipelineDetails.test.tsx`
- `src/components/CustomTable.test.tsx`
- `src/pages/NewExperiment.test.tsx`
- `src/components/SidePanel.interaction.test.tsx`
- `src/pages/ExperimentList.test.tsx`
- `src/pages/NewPipelineVersion.test.tsx`
- `src/pages/RecurringRunList.test.tsx`
- `src/pages/NewRun.test.tsx`
- `src/pages/ArchivedExperiments.test.tsx`
- `src/pages/ArchivedRuns.test.tsx`
- `src/pages/PipelineList.test.tsx`
- `src/pages/ResourceSelector.test.tsx`
- `src/pages/RunList.test.tsx`
- `src/pages/AllRecurringRunsList.test.tsx`
- `src/components/ExperimentList.test.tsx`
- `src/pages/CompareV1.test.tsx`
- `src/pages/RecurringRunDetails.test.tsx`
- `src/pages/RunDetails.test.tsx`
- `src/components/UploadPipelineDialog.test.tsx`
</details>

### 2. `querySelector` usage in tests (25 files)

Tests that use `container.querySelector` or `document.querySelector` instead of Testing Library queries (`getByRole`, `getByText`, `getByTestId`, etc.). These bypass accessibility semantics and are more brittle.

<details>
<summary>Files</summary>

- `src/mlmd/EdgeCanvas.test.tsx`
- `src/components/CustomTable.test.tsx`
- `src/pages/ExperimentDetails.test.tsx`
- `src/components/CustomTableNameColumn.test.tsx`
- `src/mlmd/ControlledEdgeCanvas.test.tsx`
[...]

PR #13255, opened 2026-04-10
chore(frontend): modernize ExperimentList, UploadPipelineDialog, LineageActionBar, ArchivedRuns tests

Priyanshu-u07:
Modernize four tests file from Enzyme-style patterns to Testing Library patterns.

**Changes across all files:**
- Removed Wrapper classes, `instance()`, and `state()` access
- Replaced `fireEvent` with `userEvent`, used semantic `getByRole` queries
- Replaced `waitFor(spy)` with `screen.findByText` where UI reflects data

**Snapshot files deleted (2):**
- `ExperimentList.test.tsx.snap`: Repeated MUI table markup; behavioral assertions already cover all rendered content
- `UploadPipelineDialog.test.tsx.snap`: MUI Dialog renders via portal, so `asFragment()` captured only an empty `<DocumentFragment />`
 

**Snapshot files kept (2):**
- `LineageActionBar.test.tsx.snap`: Small structure (3 snapshots)
- `ArchivedRuns.test.tsx.snap`: Minimal page container (14 lines) 

**Tests:**
 43 tests passing across 4 files.

Part of #13228


frontend/src/components/ExperimentList.test.tsx
```diff
@@ -192,69 +136,40 @@ describe('ExperimentList', () => {
...
+    const decoded = JSON.parse(decodeURIComponent(filterArg));
+    expect(decoded.predicates).toEqual(
+      expect.arrayContaining([
+        expect.objectContaining({
+          key: 'storage_state',
+          operation: V2beta1PredicateOperation.EQUALS,
+          string_value: V2beta1ExperimentStorageState.ARCHIVED.toString(),
+        }),
+      ]),
     );
```
jeffspahr: This rewrite no longer exercises the case where `_loadExperiments()` receives an existing `request.filter`. Right now the assertion only inspects the initial mount request, where the filter is empty, so the test would still pass if the implementation started overwriting caller-supplied predicates instead of appending the `storage_state` predicate. A small ref-based escape hatch keeps the Testing Library style while restoring the old regression coverage.

```suggestion
    const ref = React.createRef<ExperimentList>();
    render(
      <MemoryRouter>
        <ExperimentList ref={ref} {...props} />
      </MemoryRouter>,
    );
    await screen.findByText('experiment with id: testexperiment1');
    listExperimentsSpy.mockClear();
    await (ref.current as any)._loadExperiments({
      filter: encodeURIComponent(
        JSON.stringify({
          predicates: [{ key: 'k', op: 'op', string_value: 'val' }],
        }),
      ),
    });
    expect(listExperimentsSpy).toHaveBeenLastCalledWith(
      '',
      10,
      ExperimentSortKeys.CREATED_AT + ' desc',
      encodeURIComponent(
        JSON.stringify({
          predicates: [
            {
              key: 'k',
              op: 'op',
              string_value: 'val',
            },
            {
              key: 'storage_state',
              operation: V2beta1PredicateOperation.EQUALS,
              string_value: V2beta1ExperimentStorageState.ARCHIVED.toString(),
            },
          ],
        } as V2beta1Filter),
      ),
      undefined,
    );
```

frontend/src/components/ExperimentList.test.tsx
```diff
@@ -192,69 +136,40 @@ describe('ExperimentList', () => {
...
+        <ExperimentList ref={ref} {...props} />
+      </MemoryRouter>,
+    );
+    await waitFor(() => expect(listExperimentsSpy).toHaveBeenCalled());
     listExperimentsSpy.mockClear();
-    await act(async () => {
-      await wrapper.instance().refresh();
-    });
+    await ref.current!.refresh();
     await waitFor(() => expect(listExperimentsSpy).toHaveBeenCalledTimes(1));
```
jeffspahr: This imperative `refresh()` call now triggers React `act(...)` warnings because it drives state updates inside both `CustomTable` and `ExperimentList`. Please wrap it again so the test stays synchronized with the rendered updates. This also needs `act` added back to the Testing Library import above.

```suggestion
    await act(async () => {
      await ref.current!.refresh();
    });
```

jeffspahr (changes requested):
Two test issues remain on the current head.

1. `ExperimentList.test.tsx` no longer covers the important case where `_loadExperiments()` augments an existing caller-supplied filter instead of starting from an empty mount-time request.
2. The ref-driven `refresh()` test now emits React `act(...)` warnings because the imperative refresh call is no longer wrapped.

Focused verification on `df29763b3733288dafeb478e11016400b8c702dd`:
- `fnm exec --using .nvmrc -- npm ci`
- `fnm exec --using .nvmrc -- npm run test:ui -- src/components/ExperimentList.test.tsx src/components/UploadPipelineDialog.test.tsx src/mlmd/LineageActionBar.test.tsx src/pages/ArchivedRuns.test.tsx`

The touched suites passed, but the `refresh()` test printed the React warnings called out inline.

frontend/src/components/ExperimentList.test.tsx
Priyanshu-u07: Done

frontend/src/components/ExperimentList.test.tsx
Priyanshu-u07: Done

jeffspahr (approved):
Re-reviewed the current head `d21240c526542cb6eb69ea1fb2b252ffa96bb09a`.

The previously reported `ExperimentList.test.tsx` issues are fixed:
- the existing-filter augmentation coverage is restored via the ref-driven `_loadExperiments({ filter: ... })` path
- the imperative `refresh()` path is wrapped in `act(...)`, and the targeted rerun no longer emits the prior React warnings

Focused verification passed:
- `fnm exec --using .nvmrc -- npm ci`
- `fnm exec --using .nvmrc -- npm run test:ui -- src/components/ExperimentList.test.tsx`

I did not find new correctness or coverage regressions in the updated diff.

PR #13256, opened 2026-04-10
refactor(frontend): simplify CompareV2 selection state

jeffspahr:
## Summary
- make `MetricsDropdown` fully controlled by `selectedArtifacts` props instead of mirroring those props into local component state
- key `CompareV2` run selection off the route `runlist` so the page no longer repairs `selectedIds` from fetched `runs`
- add a controlled-rerender regression in `MetricsDropdown.test.tsx` and keep the existing compare selection tests green

## Why
This is the next useEffect cleanup slice after the recent `NewRun` work.

The remaining compare selection flow still had two derived-state patterns:
- `MetricsDropdown` copied `selectedArtifacts` props into local state with an effect
- `CompareV2` used an effect to reset `selectedIds` after render based on fetched run data

The new version keeps selection ownership in the parent/page path and limits the remaining effects in this area to actual external sync.

## Verification
- `cd frontend && fnm exec --using .nvmrc -- npx prettier@3.8.1 --check src/components/viewers/MetricsDropdown.tsx src/components/viewers/MetricsDropdown.test.tsx src/pages/CompareV2.tsx src/pages/CompareV2.test.tsx`
- `cd frontend && fnm exec --using .nvmrc -- npm run test:ui -- src/components/viewers/MetricsDropdown.test.tsx src/pages/CompareV2.test.tsx`
- `cd frontend && fnm exec --using .nvmrc -- npm run lint`

## Notes
The focused test run still emits pre-existing React `act(...)` warnings in the compare/visualization tests, but the suites pass.

### DIFF
--- a/frontend/src/pages/CompareV2.tsx
+++ b/frontend/src/pages/CompareV2.tsx
@@ -251,6 +251,18 @@ interface RocCurveSelectionState {
   lineColorsStack: string[];
 }
 
+function useKeyedState<T>(key: string, initialValue: T): [T, (value: T) => void] {
+  const [stateKey, setStateKey] = useState(key);
+  const [value, setValue] = useState(initialValue);
+
+  if (stateKey !== key) {
+    setStateKey(key);
+    setValue(initialValue);
+  }
+
+  return [value, setValue];
+}
+
 const createSelectedArtifactArray = (count: number): SelectedArtifact[] => {
   const array: SelectedArtifact[] = [];
   for (let i = 0; i < count; i++) {
@@ -434,12 +446,14 @@ function CompareV2(props: CompareV2Props) {
   const { updateBanner, updateToolbar, namespace } = props;
 
   const runlistRef = useRef<RunList>(null);
-  const [selectedIds, setSelectedIds] = useState<string[]>([]);
+  const queryParamRunIds = new URLParser(props).get(QUERY_PARAMS.runlist);
+  const runIds = (queryParamRunIds && queryParamRunIds.split(',')) || [];
+  const runIdsKey = runIds.join(',');
+  const [selectedIds, setSelectedIds] = useKeyedState<string[]>(runIdsKey, runIds);
   const [metricsTab, setMetricsTab] = useState(MetricsType.SCALAR_METRICS);
   const [isOverviewCollapsed, setIsOverviewCollapsed] = useState(false);
   const [isParamsCollapsed, setIsParamsCollapsed] = useState(false);
   const [isMetricsCollapsed, setIsMetricsCollapsed] = useState(false);
-  const selectionRunIdsKeyRef = useRef<string>('');
   const [rocCurveSelection, setRocCurveSelection] = useState<RocCurveSelectionState>(
     createInitialRocCurveSelectionState,
   );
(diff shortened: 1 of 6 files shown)


frontend/src/pages/CompareV2.tsx
```diff
@@ -251,6 +251,18 @@ interface RocCurveSelectionState {
   lineColorsStack: string[];
 }
 
+function useKeyedState<T>(key: string, initialValue: T): [T, (value: T) => void] {
```
manaswinidas: `useKeyedState` already exists in `NewRunV2.tsx` with a different implementation (single state atom + `useCallback`). Extract to a shared hook (e.g. `src/hooks/useKeyedState.ts`) and import in both pages to avoid divergent copies — per the repo's code reuse policy.

frontend/src/pages/CompareV2.tsx
```diff
@@ -434,12 +446,14 @@ function CompareV2(props: CompareV2Props) {
   const { updateBanner, updateToolbar, namespace } = props;
 
   const runlistRef = useRef<RunList>(null);
-  const [selectedIds, setSelectedIds] = useState<string[]>([]);
+  const queryParamRunIds = new URLParser(props).get(QUERY_PARAMS.runlist);
+  const runIds = (queryParamRunIds && queryParamRunIds.split(',')) || [];
+  const runIdsKey = runIds.join(',');
+  const [selectedIds, setSelectedIds] = useKeyedState<string[]>(runIdsKey, runIds);
```
manaswinidas: The removed `useEffect` had a stale-run filtering path: when the route stayed the same but `runs` refetched (e.g. via the refresh button), it pruned `selectedIds` to only IDs still present in the fetched data. `useKeyedState` only resets on key (route) change, so a server-side run deletion between refreshes leaves a phantom ID in `selectedIds`.

jeffspahr: Addressed in c91a2d01a8e7f516982b1a1ee2f2f1836d199479. CompareV2 now derives the visible selectedIds from the current fetched runs, so same-route refreshes drop missing run ids while route changes still reset through useKeyedState. I also added a regression in CompareV2.test.tsx that exercises the stale-selection refresh case.

frontend/src/pages/CompareV2.tsx
jeffspahr: Addressed in c91a2d01a8e7f516982b1a1ee2f2f1836d199479. I extracted the shared keyed-state helper to frontend/src/hooks/useKeyedState.ts and switched both NewRunV2 and CompareV2 to import the single implementation so the pages no longer carry divergent copies.

manaswinidas (approved):
/lgtm
/approve

jeffspahr:
Added the missing page-level regression for the stale-artifact route-change path.

What changed:
- [CompareV2.test.tsx](https://github.com/kubeflow/pipelines/pull/13256/files) now covers the case where the compare route changes, the visible run/execution/artifact labels stay the same, but the underlying HTML artifact ID changes.
- The test opens the HTML picker on the original route, selects the artifact, rerenders with the new route, and asserts that `getHtmlViewerConfig` is subsequently called with the new linked artifact ID rather than the old one.

This closes the gap from the earlier review comment: we now have both the focused `MetricsDropdown` regression and the page-level `CompareV2` regression for the same-label/different-artifact route-change scenario.

Verification on commit `b8a343eac`:
- `fnm exec --using .nvmrc -- npx prettier@3.8.1 --check src/pages/CompareV2.test.tsx`
- `fnm exec --using .nvmrc -- npm run test:ui -- src/pages/CompareV2.test.tsx`
- `fnm exec --using .nvmrc -- npm run test:ui -- src/components/viewers/MetricsDropdown.test.tsx src/pages/CompareV2.test.tsx`
- `fnm exec --using .nvmrc -- npm run lint`

Result: all passed. The compare/visualization area still emits the same pre-existing React `act(...)` warnings and the same post-success Vitest worker-timeout warning in this area, but the runs exit green.

PR #13258, opened 2026-04-10
[DRAFT] feat(backend): move driver into init container

droctothorpe:
## Summary

This PR collapses the two-pod driver+executor pattern into a **single executor pod** for regular container tasks by running `kfp-driver` as an Argo init container instead of as a standalone DAG task. Dummy-image tasks (e.g. `createPVC`/`deletePVC`) retain the legacy standalone driver pod path.

---

## Motivation

Every regular container task previously compiled into **two** Argo DAG tasks:

```
dag-task-A-driver  ──►  dag-task-A
```

Each DAG task maps to one Kubernetes pod, so a single user step required two pods and two rounds of scheduling latency. The driver pod produced a `PodSpecPatch` JSON blob, Argo applied it to the executor pod template at runtime, and the executor then started. This introduced several problems:

- **Double scheduling overhead** — two pods, two scheduler interactions, two image pulls per user step.
- **Fragile runtime mutation** — Argo's `podSpecPatch` mechanism mutated the compiled template opaquely at runtime, making compiled workflows hard to reason about and debug.
- **Invisible platform config** — static Kubernetes settings (PVC mounts, tolerations, node selectors, accelerators, etc.) were applied at runtime through `extendPodSpecPatch` rather than being visible in the compiled workflow YAML.
- **Large, hard-to-test code paths** — `initPodSpecPatch` + `extendPodSpecPatch` together spanned ~850 lines and were notoriously difficult to maintain.
- **Cache handling spread across two tasks** — on a cache hit the driver writes `cachedDecision=true` as an Argo output parameter; the executor DAG task still starts and its inner `system-container-impl` step is skipped via an Argo `when` condition. This is implicit, spread across two tasks, and relies on Argo output parameter wiring rather than a single clear ownership point.

---

## Old driver flow

```
┌─────────────────────────────────────────────────────────────────┐
│  Argo DAG                                                       │
│                                                                 │
│  ┌──────────────────────┐     ┌──────────────────────────────┐  │
│  │   driver pod         │────►│   executor pod               │  │
│  │                      │     │                              │  │
│  │  kfp-driver binary   │     │  init: copy-launcher         │  │
│  │  • resolve inputs    │     │  main: user image            │  │
│  │  • write podSpecPatch│     │        (kfp-launcher wraps   │  │
│  │  • output cached-    │     │         user entrypoint)     │  │
│  │    decision param    │     │                              │  │
│  └──────────────────────┘     └──────────────────────────────┘  │
│                                                                 │
│  Argo applies podSpecPatch at pod-creation time (runtime)       │
└─────────────────────────────────────────────────────────────────┘
```

**Data handoff:** driver → `podSpecPatch` JSON → Argo applies patch at executor pod creation → executor reads env vars injected by patch.

**Cache hit path:** driver outputs `cachedDecision=true` as an Argo task output parameter → executor DAG task starts, but its inner `system-container-impl` step is guarded by `When: cachedDecision != true` → the step is **skipped** by Argo → no user pod is ever created → the outer executor DAG task completes as succeeded because all steps either completed or were skipped.

---

## New driver flow

```
┌─────────────────────────────────────────────────────────────────┐
│  Argo DAG                                                       │
│                                                                 │
│  ┌───────────────────────────────────────────────────────────┐  │
│  │  executor pod  (one pod per task)                         │  │
│  │                                                           │  │
│  │  init[0]: kfp-driver        init[1]: copy-launcher        │  │
│  │  • resolve inputs           • copies launcher binary      │  │
│  │  • write outputs to           to shared emptyDir          │  │
│  │    /tmp/kfp-driver-outputs                                │  │
│  │  • write cached-decision                                  │  │
│  │                                                           │  │
│  │  main: user image                                         │  │
│  │    └─ kfp-launcher reads driver outputs from emptyDir     │  │
│  │       • cache hit  → exit 0, skip user code               │  │
│  │       • cache miss → execute user entrypoint              │  │
│  └───────────────────────────────────────────────────────────┘  │
│                                                                 │
│  Static K8s config embedded in compiled template (compile time) │
└─────────────────────────────────────────────────────────────────┘
```

**Data handoff:** driver writes six files to a shared `emptyDir` volume at `/tmp/kfp-driver-outputs`; launcher reads them directly — no Argo API involvement.

| File | Content |
|---|---|
| `execution-id` | Resolved MLMD execution ID |
| `executor-input` | JSON-serialised `ExecutorInput` proto |
| `user-command` | JSON array of resolved container command tokens |
| `user-args` | JSON array of resolved container arg tokens |
| `cached-decision` | `true` / `false` |
| `condition` | `true` / `false` / `nil` (nil = unconditional) |

**Cache hit path:** driver writes `cached-decision=true` → all init containers complete normally → launcher reads the flag and exits 0 without running user code → pod finishes `Succeeded`. No Argo API calls, no pod self-deletion, no race conditions.

---

### ✅ Pros

| Benefit | Detail |
|---|---|
| **Fewer pods** | One pod per user step instead of two; halves scheduling overhead, node capacity pressure, and Kubernetes API churn. |
| **No runtime template mutation** | `podSpecPatch` is gone. The compiled workflow YAML is what runs — static settings are readable in the YAML at review time. |
| **Simpler cache-hit path** | Cache decision is communicated via a plain file in a shared volume, colocated with all other driver outputs. No Argo output parameter wiring, no cross-task dependency on an output parameter, and no user pod scheduled at all on cache hit regardless of image presence on the node. |
| **~850 lines of complex code removed** | `initPodSpecPatch` + `extendPodSpecPatch` are deleted; replaced by a straightforward file-write/read protocol. |
| **Compile-time K8s config visibility** | PVC mounts, node selectors, tolerations, and accelerator settings are embedded in the compiled template and visible without running the pipeline. |
| **Cleaner MLMD lifecycle** | Execution state transitions happen fully inside the init container before the user's code starts; no partial states from the patch application window. |

### ⚠️ Cons / trade-offs

| Trade-off | Detail |
|---|---|
| **User image pulled even on cache hit** | Kubernetes pulls the user container image at pod creation, before init containers complete. For large images on cold nodes this adds latency even when the result is cached. In the old model the executor pod was never scheduled on cache hit. |
| **Init container failures are less prominent** | A crash in the `kfp-driver` init container surfaces as an init container error rather than a named failed DAG task, which is slightly less discoverable in the Argo UI. |
| **Dynamic pod-level config not supported** | Any pod-level spec derived from a runtime parameter (e.g. a volume name that is itself a pipeline parameter) must be pre-computed at compile time or encoded as an Argo template parameter reference. Fully dynamic volume names are not supported. |
| **Slightly higher per-pod resource request** | The driver init container (128 Mi) and launcher binary share the pod with the user container, increasing the pod's total resource footprint. |
| **Dummy-image tasks unchanged** | `createPVC`/`deletePVC` still use the two-task pattern; the migration is not yet complete for system tasks. |

---

### PVC volumes embedded at compile time

The original implementation attempted to resolve PVC volume names (which can be pipeline parameters) at runtime via a pre-pod hook. This was removed in favour of **compile-time embedding**: the compiler inspects the task's `kubernetes_platform` config and adds all PVC volume/volumeMount specs to the Argo template unconditionally. Dynamic PVC name parameters are kept as Argo template parameter references (`{{inputs.parameters.*}}`) so they remain resolvable at runtime without requiring pod-mutation hooks.

### OCI artifact path injection (`modelcar`)

The Modelcar sidecar copies `/models` from the OCI image into an emptyDir volume at `/oci-artifact-{i}/models` (index based on alphabetical sort of OCI input names). Previously the Python SDK computed the artifact path from the URI (`/oci/{normalized-uri}/models`), which no longer matched the mount point. The launcher now calls `injectOCIArtifactMountPaths()` before serialising `ExecutorInput`, setting `customPath = "/oci-artifact-{i}"` on each OCI artifact. The Python executor reads `customPath` and applies it so that `input_model.path` resolves to the correct path.

### CA bundle propagation

The `kfp-driver` init container makes TLS calls to MLMD and the API server. The custom CA bundle (secret or ConfigMap) is now mounted on the init container in addition to the main container.

### `configMapAsEnv` empty-name guard

A nil/empty-string check was added for `configMapAsEnv` entries to prevent the compiler from emitting invalid `envFrom` specs that caused E2E failures in the `missing_kubernetes_optional_inputs` test.

### Driver init container memory request

Set to `128Mi` (up from `64Mi`) to prevent OOM kills observed during E2E runs, particularly in pipelines with large numbers of input parameters or artifacts.

---

### Compiler (`backend/src/v2/compiler/argocompiler/`)

- **`container.go`**: Replaces the two-template (driver + executor) pattern with `addContainerExecutorTemplate()` that embeds `kfp-driver` as init container 0. Static K8s settings (PVCs, node selectors, tolerations, configMaps, secrets, accelerators, image pull policy) are applied at compile time. CA bundle is propagated to the driver init container. `configMapAsEnv` skips entries with empty names.
- **`dag.go`**: Regular container tasks now emit a single `[]wfapi.DAGTask` instead of two.

### Driver (`backend/src/v2/driver/`)

- **`driver.go`**: Removes `initPodSpecPatch()`. `Execution.PodSpecPatch` replaced by `Execution.UserCommand` / `Execution.UserArgs`.
- **`container.go`**: Strips pod-spec-patch construction; resolves container command/arg placeholders only.
- **`k8s.go`**: Removes `extendPodSpecPatch()` and `addModelcarsToPodSpec()` (~650 lines).

### Driver CLI (`backend/src/v2/cmd/driver/`)

- **`main.go`**: Adds `--driver_outputs_dir` flag; `writeDriverOutputsDir()` writes all container execution outputs to the shared emptyDir.
- **`execution_paths.go`**: Adds `DriverOutputsDir` field; legacy fields annotated accordingly.

### Launcher (`backend/src/v2/component/`, `backend/src/v2/cmd/launcher-v2/`)

- **`launcher_v2.go`**: Adds `injectOCIArtifactMountPaths()` to set `customPath` on OCI input artifacts before serialising `ExecutorInput`, fixing `input_model.path` resolution for Modelcar.
- **`main.go`**: Adds `--driver_outputs_dir` flag and `readDriverOutputs()`. On cache hit, exits 0 without running user code.

### SDK executor (`sdk/python/kfp/dsl/`)

- **`executor.py`**: `create_artifact_instance()` reads `customPath` (camelCase, protojson) or `custom_path` (snake_case, Python SDK) and applies it to the artifact after construction.

### Golden files (`test_data/compiled-workflows/`)

All ~150 compiled-workflow golden files updated to reflect the new single-task, single-pod structure.

### Dev overlay (`manifests/kustomize/env/dev-orbstack/`)

New Kustomize overlay for local development with OrbStack.

---

## Testing

- All compiler golden-file tests pass (`ginkgo -v ./backend/test/compiler`).
- `TestAddContainerExecutorTemplate` asserts init-container order, parameterised image, and presence of the driver-outputs volume.
- New unit tests for `injectOCIArtifactMountPaths` in `launcher_v2_test.go`.
- New unit tests for `create_artifact_instance` custom-path handling in `executor_test.py`.
- Full CI green on commit `d0cceedbf`: E2E Pipeline tests, API Server Tests, SDK tests, Compiler tests, and all Integration tests passing.


droctothorpe:
Here's some manual test evidence using a 5 component DAG:
image
image

  All 5 steps completed successfully with 0 restarts.

  The pipeline validated:

   - generate_data → produced integers 1–10 as an artifact
   - compute_stats → computed mean=5.5, count=10
   - filter_values → kept 5 values above mean: [6,7,8,9,10]
   - transform → squared them: [36, 49, 64, 81, 100]
   - summarize → printed the final report correctly

  Key observations:

   - Notice the pod names: system-container-impl-* (single pod per step) rather than the old system-container-driver-* + system-container-impl-* pairs — the init-driver pattern is 
  working
   - 5 user pods + 1 DAG driver pod (for the root DAG)
   - All pods Completed with exit code 0, no restarts
   - Artifact passing, parameter passing, and fan-in (step 5 reads from steps 3 and 4) all worked correctly

droctothorpe:
Here's some more manual validation evidence using conditionals and nested DAGs:
image
image
image
image

PR #13262, opened 2026-04-13
test: Add MLflow Integration test plan

VaniHaripriya:
This PR adds MLflow Integration Test plan

**Checklist:**

### DIFF
--- a/proposals/12862-mlflow-integration/TestPlan.md
+++ b/proposals/12862-mlflow-integration/TestPlan.md
@@ -1,0 +1,126 @@
+# MLflow Integration — Test Plan
+
+**Feature**: MLflow Integration for Kubeflow Pipelines  
+**Design Doc**: [KEP-12862](https://github.com/kubeflow/pipelines/blob/master/proposals/12862-mlflow-integration/README.md)
+
+---
+
+## 1\. Test Plan
+
+### 1.1 CreateRun with MLflow
+
+| Priority | Scenario | Steps | Expected |
+| :---- | :---- | :---- | :---- |
+| **P0** | Default experiment | 1\. Configure global `plugins.mlflow` 2\. Create a run without `plugins_input` | Experiment **`KFP-Default`** (unless the admin default name overrides it); MLflow parent run exists; `plugins_output.mlflow` includes `experiment_id`, `root_run_id`, `run_url`; plugin `state` \= `SUCCEEDED` |
+| **P0** | Custom experiment name | 1\. Create a run with `plugins_input.mlflow.experiment_name = "my-exp"` | Experiment `my-exp` exists; parent run lives under it |
+| **P0** | Experiment id wins over name | 1\. Send both `experiment_name` and `experiment_id` | The id is used; no lookup by name |
+| **P0** | Runtime config on driver and launcher | 1\. Create a run with MLflow enabled 2\. Inspect the compiled workflow | A JSON MLflow config env var appears on **driver** and **launcher** templates only, with endpoint, workspace (when workspaces are enabled), parent run id, experiment id, auth type, timeout, and TLS skip flag |
+| **P0** | Failed save after MLflow parent exists | 1\. MLflow parent is created, then workflow create or DB write fails | KFP run creation fails; orphan MLflow parent run remains in `RUNNING` state in MLflow indefinitely. No compensating update or delete is performed on the MLflow run, which is acceptable per design. |
+| **P0** | Clients cannot set plugin output | 1\. Create a run with client-supplied `plugins_output` | Request is rejected |
+| **P1** | KFP metadata in MLflow tags | 1\. Run linked to `pipeline_id` / `pipeline_version_id` | MLflow tags include pipeline run id, pipeline run URL, and pipeline identifiers when present |
+| **P1** | Deep links are usable | 1\. Complete a successful run with MLflow | Tag and `plugins_output` URLs open the correct **KFP run** page and **MLflow run** page (including workspace query when workspaces are on) |
+| **P1** | Experiment description variants | 1\. Omit description, send empty string, or send a custom description | Behavior matches configured defaults (including optional empty description) |
+| **P0** | Same pipeline, different experiments | 1\. Two runs with different `experiment_name` values | Two MLflow experiments and two separate parent runs |
+| **P2** | Concurrent experiment creation | 1\. Two creates race; MLflow reports the experiment already exists | Integration recovers (e.g. lookup by name) and both runs succeed |
+
+### 1.2 Run finishes (terminal pipeline state)
+
+| Priority | Scenario | Steps | Expected |
+| :---- | :---- | :---- | :---- |
+| **P0** | Success | 1\. Pipeline ends in **Succeeded** | MLflow parent run ends `FINISHED`; plugin output reflects success |
+| **P0** | Failure / cancel | 1\. Pipeline ends in **Failed** or **Canceled** | KFP **Canceled** (or canceling) maps to MLflow `KILLED`; KFP **Failed** and other non-success terminals (per mapping) map to `FAILED` |
+| **P1** | Straggler nested runs | 1\. Some nested runs still active when the pipeline completes | All relevant nested runs are found and updated (pagination respected for large fan-out) |
+| **P1** | Persisted output | 1\. After terminal handling | Latest plugin output is stored on the KFP run record |
+| **P1** | Sync errors | 1\. MLflow returns errors while syncing terminal state | Errors are logged; the KFP run’s terminal state is still recorded as usual |
+
+### 1.3 RetryRun
+
+| Priority | Scenario | Steps | Expected |
+| :---- | :---- | :---- | :---- |
+| **P0** | Retry after failure | 1\. Failed run that had MLflow output 2\. Retry | Parent MLflow run and nested runs in `FAILED`/`KILLED` go back to `RUNNING`; nested runs that were `FINISHED` stay `FINISHED`; when retried tasks re-execute, existing reopened nested runs are reused (matched by task name tag) rather than duplicated, so nested run count under the parent remains unchanged and reused runs contain updated retry metrics/parameters |
+| **P1** | Retry without MLflow output | 1\. Failed run with no MLflow plugin output | No MLflow side effects; no errors |
+
+### 1.4 CloneRun
+
+| Priority | Scenario | Steps | Expected |
+| :---- | :---- | :---- | :---- |
+| **P0** | Clone keeps plugin input | 1\. Clone a run that had `plugins_input.mlflow` | New run inherits input; a **new** MLflow parent is created for the clone |
+| **P1** | Clone vs original | 1\. Both runs complete | May share an experiment; must have distinct MLflow run ids |
+
+### 1.5 Recurring runs
+
+| Priority | Scenario | Steps | Expected |
+| :---- | :---- | :---- | :---- |
+| **P0** | Job remembers plugin input | 1\. Create a recurring job with `plugins_input` | Value is stored and appears on the scheduled workflow |
+| **P0** | Each trigger inherits input | 1\. Schedule fires → new run | Triggered runs carry the same `plugins_input` |
+| **P0** | Recurring job tied to a pipeline version | 1\. Job references a pipeline version and uses plugins | Job is accepted; controller path creates runs via the API without requiring an inline workflow spec |
+| **P1** | One MLflow parent per trigger | 1\. Several scheduled executions | Each trigger gets its own MLflow parent under the configured experiment |
+
+### 1.6 Negative & edge cases
+
+| Priority | Scenario | Steps | Expected |
+| :---- | :---- | :---- | :---- |
+| **P0** | MLflow unreachable at create | 1\. Bad or unreachable endpoint | KFP run is created; plugin output shows failure; runtime env is absent since setup never completed |
+| **P0** | Plugin explicitly disabled | 1\. `CreateRun` with `plugins_input.mlflow.disabled: true` | No MLflow side effects; KFP run is created and proceeds normally |
+| **P0** | Invalid `plugins_input.mlflow` | 1\. Invalid JSON, unknown fields, or structurally invalid values under `plugins_input.mlflow` | Request rejected with a clear validation error |
+| **P1** | Terminal handling without stored MLflow ids | 1\. Corrupt or incomplete stored plugin output | Failure path updates plugin state appropriately |
+| **P1** | MLflow HTTP behavior | 1\. Client vs server errors from MLflow | Retries only where appropriate; no infinite retry on permanent errors |
+
+### 1.7 Compiled workflow injection
+
+| Priority | Scenario | Steps | Expected |
+| :---- | :---- | :---- | :---- |
+| **P0** | Env set only on driver and launcher | 1\. Inspect templates after compile | MLflow runtime env appears only on driver/launcher (not on arbitrary DAG or executor-only templates unless specified by design) |
+| **P0** | User container gets runtime MLflow env vars | 1\. Run a pipeline task with MLflow enabled 2\. Inspect the launched user container env | User container receives `MLFLOW_TRACKING_URI`, `MLFLOW_EXPERIMENT_ID`, `MLFLOW_RUN_ID`, and `MLFLOW_TRACKING_AUTH` from the driver (plus `MLFLOW_WORKSPACE` when workspaces are enabled); `MLFLOW_RUN_ID` points to the task nested run (not the parent run) |
+| **P1** | Duplicate env keys | 1\. Template already defines the same env name | Injected value replaces consistently; no duplicate keys |
+| **P2** | MLflow globally disabled | 1\. No `plugins.mlflow` configuration | No MLflow env on workflows |
+| **P1** | `KFP_MLFLOW_CONFIG` not exposed to user code | 1\. Run a pipeline task with MLflow enabled 2\. Inspect the user container environment | `KFP_MLFLOW_CONFIG` is not present in the user process environment; only standard `MLFLOW_*` env vars are available |
+
+### 1.8 Cluster & E2E
+
+| Priority | Scenario | Steps | Expected |
+| :---- | :---- | :---- | :---- |
+| **P0** | End-to-end success | 1\. Cluster with KFP and MLflow 2\. Run a pipeline | Parent and nested MLflow runs; `plugins_output` populated |
+| **P0** | User `mlflow.log_metric()` without explicit setup | 1\. Run a pipeline component that calls `mlflow.log_metric()` directly without `mlflow.start_run()` or explicit run id 2\. Complete the run | Metric is recorded on the task's nested MLflow run identified by `MLFLOW_RUN_ID`; no duplicate nested run is created for this logging path |
+| **P0** | Retry and schedules | 1\. Failed retry; recurring job | Behavior matches sections 1.3 and 1.5 |
+| **P0** | Upgrade without MLflow | 1\. Deploy without `plugins.mlflow` | Existing behavior unchanged; no MLflow env |
+| **P1** | Enable MLflow on an existing deployment | 1\. Start from a deployment with existing runs created before MLflow enablement 2\. Enable `plugins.mlflow` and roll out 3\. Verify `GetRun`/`ListRuns` on pre-existing runs and observe an in-flight run started before enablement 4\. Create a new run after enablement | Pre-existing runs remain readable with plugin fields handled safely (empty when absent); in-flight runs complete normally; new runs created after enablement receive MLflow tracking (`plugins_output` and parent/nested MLflow runs) |
+| **P1** | Multi-tenant namespace config merge | 1\. Per-namespace `kfp-launcher` overrides selected fields 2\. Run in multiple namespaces | Namespace overrides apply only to that namespace; unset fields inherit global config; workspace behavior matches effective settings |
+| **P0** | Delete KFP run | 1\. After a successful run, delete the KFP run | MLflow runs remain |
+| **P2** | TLS networks | Deploy MLflow behind **HTTPS** | MLflow run creation must be successful. |
+
+### 1.9 Driver-level MLflow behavior
+
+| Priority | Scenario | Steps | Expected |
+| :---- | :---- | :---- | :---- |
+| **P0** | Parallel-for hierarchy without extra iteration DAG run | 1\. Run a pipeline with a parallel-for loop 2\. Inspect MLflow run tree and task tags | Hierarchy is parent run -> loop nested run -> iteration nested runs; no extra intermediate nested run is created by the iteration DAG driver |
+| **P0** | Cache hit still appears in MLflow | 1\. Run a cacheable pipeline task once 2\. Re-run to trigger cache hit 3\. Inspect MLflow nested runs for second run | Cached task has a nested MLflow run on the second run; cached outputs are logged and that nested run is marked complete |
+| **P0** | Driver-only task creates nested run | 1\. Execute a workflow that includes a driver-only task (for example PVC create/delete) 2\. Inspect MLflow runs | Driver-only task gets its own nested MLflow run and it is marked complete/failed according to task outcome, even without launcher execution |
+
+### 1.10 Launcher MLflow logging
+
+| Priority | Scenario | Steps | Expected |
+| :---- | :---- | :---- | :---- |
+| **P0** | Logs parameters and scalar metrics to nested run | 1\. Run a task with known input parameters and scalar outputs 2\. Inspect the task nested MLflow run | Nested run contains the expected parameters and scalar metrics logged by launcher `log-batch` calls |
+| **P1** | Chunks large parameter sets correctly | 1\. Run a task with more than 100 input parameters 2\. Inspect MLflow nested run contents | All parameters are present on the nested run even when launcher must split them across multiple `log-batch` requests |
+| **P0** | Nested run terminal status mirrors task result | 1\. Run one successful task and one failing task 2\. Inspect nested MLflow run statuses | Successful task nested run is marked `FINISHED`; failing task nested run is marked `FAILED` (or `KILLED` for canceled task, per mapping) |
+| **P0** | Launcher MLflow failure does not block task outcome | 1\. Make MLflow unreachable during launcher post-execution logging 2\. Run a task that otherwise succeeds | Task completion result in KFP is unchanged; MLflow logging/status-update errors are logged and do not fail the user task |
+
+### 1.11 API readback compatibility
+
+| Priority | Scenario | Steps | Expected |
+| :---- | :---- | :---- | :---- |
+| **P1** | `ListRuns` returns plugin fields for all returned runs | 1\. Create multiple runs (MLflow-enabled and non-MLflow) 2\. Call `ListRuns` for the namespace | Each returned run includes `plugins_input` / `plugins_output` as stored for that run (not only the currently inspected run) |
+| **P1** | Pre-upgrade runs with `NULL` plugin columns are handled safely | 1\. Query runs created before MLflow columns existed (or seed rows with `NULL` plugin columns) 2\. Call `GetRun` and `ListRuns` | API returns empty maps for plugin fields and does not error on `NULL` database values |
+
+### 1.12 Authentication modes and credential paths
+
+| Priority | Scenario | Steps | Expected |
+| :---- | :---- | :---- | :---- |
+| **P0** | Kubernetes auth mode | 1\. Configure `authType: "kubernetes"` 2\. Create and run a pipeline | API server uses Kubernetes SA token for MLflow calls; user container gets `MLFLOW_TRACKING_AUTH=kubernetes`; no secret credential env vars are required |
+| **P0** | Bearer auth mode | 1\. Configure `authType: "bearer"` with namespace `credentialSecretRef` 2\. Create and run a pipeline | Secret is mounted with `valueFrom.secretKeyRef`; `MLFLOW_TRACKING_TOKEN` is present on driver/launcher/user containers; MLflow operations succeed with token auth |
+| **P0** | Basic-auth mode | 1\. Configure `authType: "basic-auth"` with namespace `credentialSecretRef` 2\. Create and run a pipeline | Secret is mounted with `valueFrom.secretKeyRef`; `MLFLOW_TRACKING_USERNAME` and `MLFLOW_TRACKING_PASSWORD` are present on driver/launcher/user containers; MLflow operations succeed with basic auth |
+| **P0** | Missing namespace `credentialSecretRef` disables secret-based auth | 1\. In multi-user mode set `authType: "bearer"` or `"basic-auth"` without namespace `credentialSecretRef` 2\. Create a run in that namespace | MLflow integration is disabled for that namespace/run (no MLflow side effects); run creation and execution still proceed |
+| **P1** | Invalid credentials at API-server MLflow call path | 1\. Provide invalid/expired credentials affecting CreateRun or terminal sync calls 2\. Create/complete a run | Run proceeds; `plugins_output.mlflow.state` is set to failed with an auth-related `state_message` for API-server-side failures |
+| **P1** | Invalid credentials at driver/launcher MLflow call path | 1\. Provide credentials that fail only during driver/launcher MLflow calls (nested run/logging) 2\. Run a task that otherwise succeeds | Task outcome is unchanged (not blocked by MLflow auth failure); errors are logged; do not require per-task auth failures to be reflected in `plugins_output.mlflow.state` |
+


proposals/12862-mlflow-integration/TestPlan.md
```diff
@@ -0,0 +1,86 @@
+# MLflow Integration — Test Plan
+
+**Feature**: MLflow Integration for Kubeflow Pipelines  
+**Design Doc**: [KEP-12862](https://github.com/mprahl/pipelines/blob/065dcbe24ae664201777c95144346dfbc25be110/proposals/12862-mlflow-integration/README.md)
```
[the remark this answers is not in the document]
VaniHaripriya: Done

proposals/12862-mlflow-integration/TestPlan.md
```diff
@@ -0,0 +1,87 @@
+# MLflow Integration — Test Plan
+
+**Feature**: MLflow Integration for Kubeflow Pipelines  
+**Design Doc**: [KEP-12862](https://github.com/kubeflow/pipelines/blob/master/proposals/12862-mlflow-integration/README.md)
+
+---
+
+## 1\. Test Plan
+
+### 1.1 CreateRun with MLflow (happy path)
```
alyssacgoins: The `happy path` reference here is a little confusing because it's only applied to heading **1.1** - I believe that sections 1.2 (Run Finishes); 1.3 (Retry Run - although the initial run fails, we want to see RetryRun execute without failure); 1.4 (CloneRun); 1.5 (Recurring Runs); 1.7 (Compiled workflow injection) should each contain happy path scenarios. It might make more sense not to include `happy path` in the heading, but rather label it on specific scenarios within each heading.

proposals/12862-mlflow-integration/TestPlan.md
```diff
@@ -0,0 +1,87 @@
...
+
+| Pri | Scenario | Steps | Expected |
```
alyssacgoins:
```suggestion
| Priority | Scenario | Steps | Expected |
```

proposals/12862-mlflow-integration/TestPlan.md
```diff
@@ -0,0 +1,87 @@
...
+| **P1** | Terminal handling without stored MLflow ids | 1\. Corrupt or incomplete stored plugin output | Failure path updates plugin state appropriately |
+| **P1** | MLflow HTTP behavior | 1\. Client vs server errors from MLflow | Retries only where appropriate; no infinite retry on permanent errors |
+
+### 1.7 Compiled workflow injection
+
+| Pri | Scenario | Steps | Expected |
+| :---- | :---- | :---- | :---- |
+| **P0** | Env only on driver and launcher | 1\. Inspect templates after compile | MLflow runtime env appears only on driver/launcher (not on arbitrary DAG or executor-only templates unless specified by design) |
+| **P1** | Duplicate env keys | 1\. Template already defines the same env name | Injected value replaces consistently; no duplicate keys |
+| **P2** | MLflow globally off | 1\. No `plugins.mlflow` configuration | No MLflow env on workflows |
```
alyssacgoins:
```suggestion
| **P2** | MLflow globally disabled | 1\. No `plugins.mlflow` configuration | No MLflow env on workflows |
```

proposals/12862-mlflow-integration/TestPlan.md
```diff
@@ -0,0 +1,87 @@
...
+| **P0** | Plugin disabled or empty input | 1\. `plugins_input.mlflow.disabled: true` or invalid handling | No MLflow effects or clear validation error |
+| **P0** | Malformed plugin JSON | 1\. Invalid JSON or unknown fields under `plugins_input.mlflow` | Request rejected with a validation error |
+| **P1** | Terminal handling without stored MLflow ids | 1\. Corrupt or incomplete stored plugin output | Failure path updates plugin state appropriately |
+| **P1** | MLflow HTTP behavior | 1\. Client vs server errors from MLflow | Retries only where appropriate; no infinite retry on permanent errors |
+
+### 1.7 Compiled workflow injection
+
+| Pri | Scenario | Steps | Expected |
+| :---- | :---- | :---- | :---- |
+| **P0** | Env only on driver and launcher | 1\. Inspect templates after compile | MLflow runtime env appears only on driver/launcher (not on arbitrary DAG or executor-only templates unless specified by design) |
```
alyssacgoins:
```suggestion
| **P0** | Env set only on driver and launcher | 1\. Inspect templates after compile | MLflow runtime env appears only on driver/launcher (not on arbitrary DAG or executor-only templates unless specified by design) |
```

alyssacgoins (changes requested):
Overall, I think you could divide the headings a little differently - sections `1.1-1.5` as well as `1.7` all refer to specific behavior to test, whereas section `1.6` refers to edge cases within those headings and section `1.8` covers complete E2E behavior.
I think you could divide the plan into three major headings - **happy path**, **negative and edge cases** and **E2E**, with subheadings covering each behavioral component (ie `CreateRun`)

proposals/12862-mlflow-integration/TestPlan.md
VaniHaripriya: Updated to remove happy path to avoid confusion.

proposals/12862-mlflow-integration/TestPlan.md
```diff
@@ -0,0 +1,87 @@
...
+
+### 1.7 Compiled workflow injection
+
+| Pri | Scenario | Steps | Expected |
+| :---- | :---- | :---- | :---- |
+| **P0** | Env set only on driver and launcher | 1\. Inspect templates after compile | MLflow runtime env appears only on driver/launcher (not on arbitrary DAG or executor-only templates unless specified by design) |
+| **P1** | Duplicate env keys | 1\. Template already defines the same env name | Injected value replaces consistently; no duplicate keys |
+| **P2** | MLflow globally disabled | 1\. No `plugins.mlflow` configuration | No MLflow env on workflows |
+
+### 1.8 Cluster & E2E
```
hbelmiro, line 77: The E2E scenario in section 1.8 ("Parent and nested MLflow runs; `plugins_output` populated") implicitly validates that the driver creates nested runs for a basic pipeline. However, the KEP design doc (sections "Driver - Container Tasks", "Loop Iterations", "Cache Hits", "Driver-Only Tasks") specifies three additional driver behaviors that have no corresponding test scenarios:

1. **Parallel-for loop hierarchy** -- The design specifies a clean two-level hierarchy (parent run -> loop nested run -> iteration nested runs) where the iteration DAG driver deliberately skips MLflow run creation. No scenario verifies this hierarchy or checks for extraneous intermediate runs.
2. **Cache hit behavior** -- When a cache hit is detected, the driver must create a nested MLflow run, log cached outputs, and mark it complete immediately. No scenario covers this; a pipeline with a mix of cached and non-cached tasks could silently drop MLflow tracking for cached steps.
3. **Driver-only tasks** (e.g., PVC creation/deletion) -- These run only in the driver without a launcher. No scenario verifies they produce nested MLflow runs.

**Why it matters:** These are core feature behaviors from the KEP. The E2E covers the basic nested-run happy path, but parallel-for, cache hits, and driver-only tasks exercise distinct code paths that would go untested.

**Suggested fix:** Add a new section (e.g., "1.X Driver-level MLflow behavior") with P0 scenarios for a parallel-for pipeline verifying the two-level hierarchy, a cache-hit pipeline verifying the cached task appears in MLflow, and a driver-only task scenario.

hbelmiro, line 77: The KEP (section "Launcher") describes post-execution MLflow logging as a primary user-facing goal (KEP Goal #3: "Automatically log scalar metrics and input/output parameters from each task to MLflow without requiring changes to pipeline components"). The E2E scenario in section 1.8 verifies `plugins_output` is populated on the KFP side, but no scenario -- including the E2E -- verifies that metrics and parameters actually appear in the MLflow nested run. Specifically:

1. **Parameter and metric logging** -- After user code completes, the launcher logs input parameters and scalar output metrics via `POST /api/2.0/mlflow/runs/log-batch`. No scenario checks that parameters and metrics appear correctly in the nested MLflow run.
2. **Large parameter set chunking** -- The design specifies chunking into multiple `log-batch` calls when parameters exceed 100 per call. No boundary-condition scenario exists.
3. **Nested run completion** -- The launcher marks the nested run as complete or failed. No scenario explicitly verifies the nested run status after a successful or failed task (distinct from the parent run status tested in section 1.2).
4. **Launcher failure isolation** -- MLflow logging errors in the launcher should not block task completion. Section 1.6 "MLflow HTTP behavior" partially touches this ("Retries only where appropriate; no infinite retry on permanent errors"), but it targets MLflow API behavior in general, not the launcher's specific guarantee that user-task completion is not blocked.

**Why it matters:** Automatic metric/parameter logging without pipeline code changes is a primary user-facing goal of the KEP. The E2E verifies KFP-side plugin output but not MLflow-side data (metrics, parameters on the nested run).

**Suggested fix:** Add a new section (e.g., "1.X Launcher MLflow logging") with scenarios for: a task with known parameters/metrics verifying they appear in the MLflow nested run, a task with >100 parameters verifying chunking works, a task where the MLflow server is unreachable during post-execution logging verifying the task still succeeds.

proposals/12862-mlflow-integration/TestPlan.md
```diff
@@ -0,0 +1,87 @@
...
+
+| Pri | Scenario | Steps | Expected |
+| :---- | :---- | :---- | :---- |
+| **P0** | End-to-end success | 1\. Cluster with KFP and MLflow 2\. Run a pipeline | Parent and nested MLflow runs; `plugins_output` populated |
+| **P0** | Retry and schedules | 1\. Failed retry; recurring job | Behavior matches sections 1.3 and 1.5 |
+| **P0** | Upgrade without MLflow | 1\. Deploy without `plugins.mlflow` | Existing behavior unchanged; no MLflow env |
+| **P0** | Multi-tenant namespace config | 1\. Per-namespace `kfp-launcher` overrides | Merged settings and workspace behavior; bearer/basic need namespace `credentialSecretRef` |
```
hbelmiro, line 84: The KEP defines three authentication modes (`"kubernetes"`, `"bearer"`, `"basic-auth"`), each exercising a distinct code path. Section 1.8 "Multi-tenant namespace config" acknowledges this ("bearer/basic need namespace `credentialSecretRef`"), but packs all modes into a single scenario. The three auth modes should each have their own test scenario because they involve different credential resolution, Secret mounting, and env var injection:

- **Kubernetes auth** -- API server uses its SA token; user containers get `MLFLOW_TRACKING_AUTH=kubernetes`. No Secret involved.
- **Bearer token** -- `credentialSecretRef` Secret is mounted via `valueFrom.secretKeyRef`; `MLFLOW_TRACKING_TOKEN` is set on driver/launcher/user containers.
- **Basic auth** -- Same mechanism but with `MLFLOW_TRACKING_USERNAME` / `MLFLOW_TRACKING_PASSWORD`.
- **Missing credentialSecretRef** -- In multi-user mode with `"bearer"` or `"basic-auth"`, a namespace without `credentialSecretRef` should disable MLflow for runs in that namespace. This is a security boundary worth testing explicitly.
- **Invalid or expired credentials** -- Run should proceed; MLflow plugin output should reflect the auth failure.

**Why it matters:** Authentication is security-critical. Each auth mode exercises a different code path (SA token vs. Secret mount vs. env var injection). The current single catch-all scenario cannot verify that all three paths work correctly.

**Suggested fix:** Expand section 1.8 "Multi-tenant namespace config" into per-auth-mode P0 scenarios, or add a dedicated auth section covering each mode and the missing-credential fallback.

proposals/12862-mlflow-integration/TestPlan.md
hbelmiro, line 77: The E2E scenario in section 1.8 implicitly requires `GetRun` to return `plugins_output` (to verify it is "populated"). However, the KEP also specifies `ListRuns` returns `plugins_input` and `plugins_output` as stored in the database, and there is no scenario verifying:

- `ListRuns` includes plugin fields for all returned runs (not just the run currently being tested).
- Runs created before the MLflow feature (NULL columns after DB migration) return empty maps rather than errors.

**Suggested fix:** Add P1 scenarios verifying ListRuns returns plugin data and that pre-upgrade runs with NULL plugin columns are handled gracefully.

proposals/12862-mlflow-integration/TestPlan.md
```diff
@@ -0,0 +1,87 @@
...
+
+| Pri | Scenario | Steps | Expected |
+| :---- | :---- | :---- | :---- |
+| **P0** | Success | 1\. Pipeline ends in **Succeeded** | MLflow parent run ends `FINISHED`; plugin output reflects success |
+| **P0** | Failure / cancel | 1\. Pipeline ends in **Failed** or **Canceled** | KFP **Canceled** (or canceling) maps to MLflow `KILLED`; KFP **Failed** and other non-success terminals (per mapping) map to `FAILED` |
+| **P1** | Straggler nested runs | 1\. Some nested runs still active when the pipeline completes | All relevant nested runs are found and updated (pagination respected for large fan-out) |
+| **P1** | Persisted output | 1\. After terminal handling | Latest plugin output is stored on the KFP run record |
+| **P1** | Sync errors | 1\. MLflow returns errors while syncing terminal state | Errors are logged; the KFP run’s terminal state is still recorded as usual |
+
+### 1.3 RetryRun
```
hbelmiro, line 36: The KEP specifies that when a retried task re-executes, the driver "reuses the existing reopened nested MLflow run (matched by task name tag) rather than creating a new one." The current "Retry after failure" scenario (line 41) only checks that parent and nested runs go back to `RUNNING`, but does not verify that the retried task reuses the existing nested run instead of creating a duplicate.

**Suggested fix:** Add an expected outcome to the "Retry after failure" scenario: after retry completes, the nested run count under the parent should remain the same (no new duplicates), and the reused nested run should have updated metrics/parameters from the retry.

proposals/12862-mlflow-integration/TestPlan.md
```diff
@@ -0,0 +1,87 @@
...
+
+| Pri | Scenario | Steps | Expected |
+| :---- | :---- | :---- | :---- |
+| **P0** | MLflow unreachable at create | 1\. Bad or unreachable endpoint | KFP run may still be created; plugin output shows failure; runtime env may be absent if setup never completed |
+| **P0** | Plugin disabled or empty input | 1\. `plugins_input.mlflow.disabled: true` or invalid handling | No MLflow effects or clear validation error |
+| **P0** | Malformed plugin JSON | 1\. Invalid JSON or unknown fields under `plugins_input.mlflow` | Request rejected with a validation error |
+| **P1** | Terminal handling without stored MLflow ids | 1\. Corrupt or incomplete stored plugin output | Failure path updates plugin state appropriately |
+| **P1** | MLflow HTTP behavior | 1\. Client vs server errors from MLflow | Retries only where appropriate; no infinite retry on permanent errors |
+
+### 1.7 Compiled workflow injection
```
hbelmiro, line 69: Section 1.7 ("Compiled workflow injection") verifies that the `KFP_MLFLOW_CONFIG` env var appears only on driver/launcher templates. However, the KEP also specifies that the *driver* sets `MLFLOW_TRACKING_URI`, `MLFLOW_WORKSPACE`, `MLFLOW_EXPERIMENT_ID`, and `MLFLOW_RUN_ID` on *user containers* via the pod spec patch. These are distinct from the compile-time env vars and are critical for enabling user code to call `mlflow.log_metric()` directly.

No scenario verifies:
- User container receives `MLFLOW_RUN_ID` pointing to the nested run (not the parent).
- User container receives `MLFLOW_EXPERIMENT_ID` and `MLFLOW_TRACKING_URI`.
- User code calling `mlflow.log_metric()` without explicit run management logs to the correct nested run.

**Suggested fix:** Add a P0 scenario verifying user-container env vars and an E2E scenario where a pipeline component calls `mlflow.log_metric()` without explicit setup and the metric appears on the correct nested MLflow run.

proposals/12862-mlflow-integration/TestPlan.md
```diff
@@ -0,0 +1,87 @@
...
+| **P1** | Duplicate env keys | 1\. Template already defines the same env name | Injected value replaces consistently; no duplicate keys |
+| **P2** | MLflow globally disabled | 1\. No `plugins.mlflow` configuration | No MLflow env on workflows |
+
+### 1.8 Cluster & E2E
+
+| Pri | Scenario | Steps | Expected |
+| :---- | :---- | :---- | :---- |
+| **P0** | End-to-end success | 1\. Cluster with KFP and MLflow 2\. Run a pipeline | Parent and nested MLflow runs; `plugins_output` populated |
+| **P0** | Retry and schedules | 1\. Failed retry; recurring job | Behavior matches sections 1.3 and 1.5 |
+| **P0** | Upgrade without MLflow | 1\. Deploy without `plugins.mlflow` | Existing behavior unchanged; no MLflow env |
```
hbelmiro, line 83: Section 1.8 has "Upgrade without MLflow" which verifies that deploying without `plugins.mlflow` leaves existing behavior unchanged. However, there is no scenario for the more common real-world path: enabling MLflow on a deployment that already has existing runs in the database. The KEP specifies new `plugins_input` and `plugins_output` columns of type `LargeText` with GORM auto-migration; existing runs will have NULL values. No scenario verifies:

- Enabling `plugins.mlflow` on a deployment with pre-existing runs does not break those runs (NULL plugin columns handled gracefully by GetRun/ListRuns).
- Pipelines running at the time of the upgrade and MLflow enablement complete normally.

**Suggested fix:** Add a P1 scenario in section 1.8 for enable-MLflow-on-existing-deployment, verifying pre-existing runs remain intact and new runs get MLflow tracking.

proposals/12862-mlflow-integration/TestPlan.md
```diff
@@ -0,0 +1,87 @@
...
+| **P0** | Clients cannot set plugin output | 1\. Create a run with client-supplied `plugins_output` | Request is rejected |
+| **P1** | KFP metadata in MLflow tags | 1\. Run linked to `pipeline_id` / `pipeline_version_id` | MLflow tags include pipeline run id, pipeline run URL, and pipeline identifiers when present |
+| **P1** | Deep links are usable | 1\. Complete a successful run with MLflow | Tag and `plugins_output` URLs open the correct **KFP run** page and **MLflow run** page (including workspace query when workspaces are on) |
+| **P1** | Experiment description variants | 1\. Omit description, send empty string, or send a custom description | Behavior matches configured defaults (including optional empty description) |
+| **P0** | Same pipeline, different experiments | 1\. Two runs with different `experiment_name` values | Two MLflow experiments and two separate parent runs |
+| **P2** | Concurrent experiment creation | 1\. Two creates race; MLflow reports the experiment already exists | Integration recovers (e.g. lookup by name) and both runs succeed |
+
+### 1.2 Run finishes (terminal pipeline state)
+
+| Pri | Scenario | Steps | Expected |
```
hbelmiro, line 28:
```suggestion
| Priority | Scenario | Steps | Expected |
```

proposals/12862-mlflow-integration/TestPlan.md
```diff
@@ -0,0 +1,87 @@
...
+| :---- | :---- | :---- | :---- |
+| **P0** | Success | 1\. Pipeline ends in **Succeeded** | MLflow parent run ends `FINISHED`; plugin output reflects success |
+| **P0** | Failure / cancel | 1\. Pipeline ends in **Failed** or **Canceled** | KFP **Canceled** (or canceling) maps to MLflow `KILLED`; KFP **Failed** and other non-success terminals (per mapping) map to `FAILED` |
+| **P1** | Straggler nested runs | 1\. Some nested runs still active when the pipeline completes | All relevant nested runs are found and updated (pagination respected for large fan-out) |
+| **P1** | Persisted output | 1\. After terminal handling | Latest plugin output is stored on the KFP run record |
+| **P1** | Sync errors | 1\. MLflow returns errors while syncing terminal state | Errors are logged; the KFP run’s terminal state is still recorded as usual |
+
+### 1.3 RetryRun
+
+| Pri | Scenario | Steps | Expected |
```
hbelmiro, line 38:
```suggestion
| Priority | Scenario | Steps | Expected |
```

proposals/12862-mlflow-integration/TestPlan.md
```diff
@@ -0,0 +1,87 @@
...
+| :---- | :---- | :---- | :---- |
+| **P0** | Retry after failure | 1\. Failed run that had MLflow output 2\. Retry | Parent MLflow run and nested runs in `FAILED`/`KILLED` go back to `RUNNING`; nested runs that were `FINISHED` stay `FINISHED` |
+| **P1** | Retry without MLflow output | 1\. Failed run with no MLflow plugin output | No MLflow side effects; no errors |
+
+### 1.4 CloneRun
+
+| Pri | Scenario | Steps | Expected |
```
hbelmiro, line 45:
```suggestion
| Priority | Scenario | Steps | Expected |
```

proposals/12862-mlflow-integration/TestPlan.md
```diff
@@ -0,0 +1,87 @@
...
+| :---- | :---- | :---- | :---- |
+| **P0** | Clone keeps plugin input | 1\. Clone a run that had `plugins_input.mlflow` | New run inherits input; a **new** MLflow parent is created for the clone |
+| **P1** | Clone vs original | 1\. Both runs complete | May share an experiment; must have distinct MLflow run ids |
+
+### 1.5 Recurring runs
+
+| Pri | Scenario | Steps | Expected |
```
hbelmiro, line 52:
```suggestion
| Priority | Scenario | Steps | Expected |
```

proposals/12862-mlflow-integration/TestPlan.md
```diff
@@ -0,0 +1,87 @@
...
+| :---- | :---- | :---- | :---- |
+| **P0** | Job remembers plugin input | 1\. Create a recurring job with `plugins_input` | Value is stored and appears on the scheduled workflow |
+| **P0** | Each trigger inherits input | 1\. Schedule fires → new run | Triggered runs carry the same `plugins_input` |
+| **P0** | Recurring job tied to a pipeline version | 1\. Job references a pipeline version and uses plugins | Job is accepted; controller path creates runs via the API without requiring an inline workflow spec |
+| **P1** | One MLflow parent per trigger | 1\. Several scheduled executions | Each trigger gets its own MLflow parent under the configured experiment |
+
+### 1.6 Negative & edge cases
+
+| Pri | Scenario | Steps | Expected |
```
hbelmiro, line 61:
```suggestion
| Priority | Scenario | Steps | Expected |
```

proposals/12862-mlflow-integration/TestPlan.md
```diff
@@ -0,0 +1,87 @@
...
+| :---- | :---- | :---- | :---- |
+| **P0** | MLflow unreachable at create | 1\. Bad or unreachable endpoint | KFP run may still be created; plugin output shows failure; runtime env may be absent if setup never completed |
+| **P0** | Plugin disabled or empty input | 1\. `plugins_input.mlflow.disabled: true` or invalid handling | No MLflow effects or clear validation error |
+| **P0** | Malformed plugin JSON | 1\. Invalid JSON or unknown fields under `plugins_input.mlflow` | Request rejected with a validation error |
+| **P1** | Terminal handling without stored MLflow ids | 1\. Corrupt or incomplete stored plugin output | Failure path updates plugin state appropriately |
+| **P1** | MLflow HTTP behavior | 1\. Client vs server errors from MLflow | Retries only where appropriate; no infinite retry on permanent errors |
+
+### 1.7 Compiled workflow injection
+
+| Pri | Scenario | Steps | Expected |
```
hbelmiro, line 71:
```suggestion
| Priority | Scenario | Steps | Expected |
```

proposals/12862-mlflow-integration/TestPlan.md
```diff
@@ -0,0 +1,87 @@
...
+| :---- | :---- | :---- | :---- |
+| **P0** | Env set only on driver and launcher | 1\. Inspect templates after compile | MLflow runtime env appears only on driver/launcher (not on arbitrary DAG or executor-only templates unless specified by design) |
+| **P1** | Duplicate env keys | 1\. Template already defines the same env name | Injected value replaces consistently; no duplicate keys |
+| **P2** | MLflow globally disabled | 1\. No `plugins.mlflow` configuration | No MLflow env on workflows |
+
+### 1.8 Cluster & E2E
+
+| Pri | Scenario | Steps | Expected |
```
hbelmiro, line 79:
```suggestion
| Priority | Scenario | Steps | Expected |
```

proposals/12862-mlflow-integration/TestPlan.md
VaniHaripriya: Updated 1.8 section wiht new tests.

proposals/12862-mlflow-integration/TestPlan.md
VaniHaripriya: Done.

VaniHaripriya: Done, added a new section.

VaniHaripriya: Done, added a new section.

proposals/12862-mlflow-integration/TestPlan.md
VaniHaripriya: Updated the 1.7 section with new tests

proposals/12862-mlflow-integration/TestPlan.md
VaniHaripriya: Done, added a new section.

proposals/12862-mlflow-integration/TestPlan.md
VaniHaripriya: Added the test.

proposals/12862-mlflow-integration/TestPlan.md
```diff
@@ -0,0 +1,125 @@
...
+| **P0** | Job remembers plugin input | 1\. Create a recurring job with `plugins_input` | Value is stored and appears on the scheduled workflow |
+| **P0** | Each trigger inherits input | 1\. Schedule fires → new run | Triggered runs carry the same `plugins_input` |
+| **P0** | Recurring job tied to a pipeline version | 1\. Job references a pipeline version and uses plugins | Job is accepted; controller path creates runs via the API without requiring an inline workflow spec |
+| **P1** | One MLflow parent per trigger | 1\. Several scheduled executions | Each trigger gets its own MLflow parent under the configured experiment |
+
+### 1.6 Negative & edge cases
+
+| Priority | Scenario | Steps | Expected |
+| :---- | :---- | :---- | :---- |
+| **P0** | MLflow unreachable at create | 1\. Bad or unreachable endpoint | KFP run may still be created; plugin output shows failure; runtime env may be absent if setup never completed |
```
hbelmiro: Test expectations should be deterministic.

```suggestion
| **P0** | MLflow unreachable at create | 1\. Bad or unreachable endpoint | KFP run is created; plugin output shows failure; runtime env is absent since setup never completed |
```

proposals/12862-mlflow-integration/TestPlan.md
```diff
@@ -0,0 +1,125 @@
...
+| **P0** | Retry and schedules | 1\. Failed retry; recurring job | Behavior matches sections 1.3 and 1.5 |
+| **P0** | Upgrade without MLflow | 1\. Deploy without `plugins.mlflow` | Existing behavior unchanged; no MLflow env |
+| **P1** | Enable MLflow on an existing deployment | 1\. Start from a deployment with existing runs created before MLflow enablement 2\. Enable `plugins.mlflow` and roll out 3\. Verify `GetRun`/`ListRuns` on pre-existing runs and observe an in-flight run started before enablement 4\. Create a new run after enablement | Pre-existing runs remain readable with plugin fields handled safely (empty when absent); in-flight runs complete normally; new runs created after enablement receive MLflow tracking (`plugins_output` and parent/nested MLflow runs) |
+| **P1** | Multi-tenant namespace config merge | 1\. Per-namespace `kfp-launcher` overrides selected fields 2\. Run in multiple namespaces | Namespace overrides apply only to that namespace; unset fields inherit global config; workspace behavior matches effective settings |
+| **P0** | Delete KFP run | 1\. After a successful run, delete the KFP run | MLflow runs may remain (by design) |
```
hbelmiro: Test expectations should be deterministic.

```suggestion
| **P0** | Delete KFP run | 1\. After a successful run, delete the KFP run | MLflow runs remain |
```

proposals/12862-mlflow-integration/TestPlan.md
```diff
@@ -0,0 +1,125 @@
...
+| **P0** | Each trigger inherits input | 1\. Schedule fires → new run | Triggered runs carry the same `plugins_input` |
+| **P0** | Recurring job tied to a pipeline version | 1\. Job references a pipeline version and uses plugins | Job is accepted; controller path creates runs via the API without requiring an inline workflow spec |
+| **P1** | One MLflow parent per trigger | 1\. Several scheduled executions | Each trigger gets its own MLflow parent under the configured experiment |
+
+### 1.6 Negative & edge cases
+
+| Priority | Scenario | Steps | Expected |
+| :---- | :---- | :---- | :---- |
+| **P0** | MLflow unreachable at create | 1\. Bad or unreachable endpoint | KFP run may still be created; plugin output shows failure; runtime env may be absent if setup never completed |
+| **P0** | Plugin disabled or empty input | 1\. `plugins_input.mlflow.disabled: true` or invalid handling | No MLflow effects or clear validation error |
```
hbelmiro: The "Plugin disabled or empty input" scenario combines two distinct behaviors in one row: `plugins_input.mlflow.disabled: true` (explicit opt-out) and "invalid handling" (error path). These exercise different code paths -- one is a valid no-op configuration, the other is a validation failure. The expected outcome _"No MLflow effects or clear validation error"_ uses "or" to paper over the ambiguity.

**Suggested fix:** Split into two rows:

| **P0** | Plugin explicitly disabled | `plugins_input.mlflow.disabled: true` | No MLflow side effects; run proceeds normally |
| **P0** | Invalid plugin input structure | Send structurally invalid `plugins_input.mlflow` | Request rejected with a validation error |

VaniHaripriya: Updated the test plan.

proposals/12862-mlflow-integration/TestPlan.md
```diff
@@ -0,0 +1,125 @@
...
+| **P0** | Default experiment | 1\. Configure global `plugins.mlflow` 2\. Create a run without `plugins_input` | Experiment **`KFP-Default`** (unless the admin default name overrides it); MLflow parent run exists; `plugins_output.mlflow` includes `experiment_id`, `root_run_id`, `run_url`; plugin `state` \= `PLUGIN_SUCCEEDED` |
+| **P0** | Custom experiment name | 1\. Create a run with `plugins_input.mlflow.experiment_name = "my-exp"` | Experiment `my-exp` exists; parent run lives under it |
+| **P0** | Experiment id wins over name | 1\. Send both `experiment_name` and `experiment_id` | The id is used; no lookup by name |
+| **P0** | Runtime config on driver and launcher | 1\. Create a run with MLflow enabled 2\. Inspect the compiled workflow | A JSON MLflow config env var appears on **driver** and **launcher** templates only, with endpoint, parent run id, experiment id, auth mode, timeout, TLS skip flag, and user-env injection flag when configured |
+| **P0** | Failed save after MLflow parent exists | 1\. MLflow parent is created, then workflow create or DB write fails | MLflow parent run status updated to `FAILED`. The MLflow run is orphaned, which is acceptable as per design. |
```
hbelmiro: The "Failed save after MLflow parent exists" scenario expects "MLflow parent run status updated to `FAILED`." This contradicts the KEP design doc, which explicitly states (Risks and Mitigations section):

> "If the API server successfully creates an MLflow parent run but subsequently fails to persist the KFP run in the database, the orphan MLflow run is left as-is. Orphan runs in MLflow are harmless (they appear as 'running' indefinitely) and can be cleaned up manually or via MLflow's garbage collection. The narrow failure window makes compensating deletes not worth the added complexity."

The KEP says no compensating action is taken on the orphan MLflow run -- it stays in `RUNNING` state. The test plan says the status is updated to `FAILED`, which implies the API server makes an additional MLflow API call after a DB failure, which is the exact complexity the KEP rejects.


```suggestion
| **P0** | Failed save after MLflow parent exists | 1\. MLflow parent is created, then workflow create or DB write fails | KFP run creation fails; orphan MLflow parent run remains in `RUNNING` state in MLflow indefinitely. No compensating update or delete is performed on the MLflow run, which is acceptable per design. |
```

proposals/12862-mlflow-integration/TestPlan.md
```diff
@@ -0,0 +1,125 @@
...
+## 1\. Test Plan
+
+### 1.1 CreateRun with MLflow
+
+| Priority | Scenario | Steps | Expected |
+| :---- | :---- | :---- | :---- |
+| **P0** | Default experiment | 1\. Configure global `plugins.mlflow` 2\. Create a run without `plugins_input` | Experiment **`KFP-Default`** (unless the admin default name overrides it); MLflow parent run exists; `plugins_output.mlflow` includes `experiment_id`, `root_run_id`, `run_url`; plugin `state` \= `PLUGIN_SUCCEEDED` |
+| **P0** | Custom experiment name | 1\. Create a run with `plugins_input.mlflow.experiment_name = "my-exp"` | Experiment `my-exp` exists; parent run lives under it |
+| **P0** | Experiment id wins over name | 1\. Send both `experiment_name` and `experiment_id` | The id is used; no lookup by name |
+| **P0** | Runtime config on driver and launcher | 1\. Create a run with MLflow enabled 2\. Inspect the compiled workflow | A JSON MLflow config env var appears on **driver** and **launcher** templates only, with endpoint, parent run id, experiment id, auth mode, timeout, TLS skip flag, and user-env injection flag when configured |
```
hbelmiro: The "Runtime config on driver and launcher" scenario lists the `KFP_MLFLOW_CONFIG` fields as: "endpoint, parent run id, experiment id, auth mode, timeout, TLS skip flag, and user-env injection flag when configured." Two issues:

1. **"user-env injection flag" does not exist** in the KEP's `MLflowRuntimeConfig` struct. The struct fields are: `Endpoint`, `Workspace`, `ParentRunID`, `ExperimentID`, `AuthType`, `Timeout`, `InsecureSkipVerify`. There is no user-env injection flag.

2. **`Workspace` is missing** from the list. The KEP specifies `Workspace string` which is set "only when workspacesEnabled is true." This is a conditional field that should be verified.

**Suggested fix:** Replace the expected with:

> A JSON MLflow config env var appears on **driver** and **launcher** templates only, with endpoint, workspace (when workspaces are enabled), parent run id, experiment id, auth type, timeout, and TLS skip flag

proposals/12862-mlflow-integration/TestPlan.md
```diff
@@ -0,0 +1,125 @@
...
+| **P0** | Invalid `plugins_input.mlflow` | 1\. Invalid JSON, unknown fields, or structurally invalid values under `plugins_input.mlflow` | Request rejected with a clear validation error |
+| **P1** | Terminal handling without stored MLflow ids | 1\. Corrupt or incomplete stored plugin output | Failure path updates plugin state appropriately |
+| **P1** | MLflow HTTP behavior | 1\. Client vs server errors from MLflow | Retries only where appropriate; no infinite retry on permanent errors |
+
+### 1.7 Compiled workflow injection
+
+| Priority | Scenario | Steps | Expected |
+| :---- | :---- | :---- | :---- |
+| **P0** | Env set only on driver and launcher | 1\. Inspect templates after compile | MLflow runtime env appears only on driver/launcher (not on arbitrary DAG or executor-only templates unless specified by design) |
+| **P0** | User container gets runtime MLflow env vars | 1\. Run a pipeline task with MLflow enabled 2\. Inspect the launched user container env | User container receives `MLFLOW_TRACKING_URI`, `MLFLOW_EXPERIMENT_ID`, and `MLFLOW_RUN_ID` from the driver (plus `MLFLOW_WORKSPACE` when enabled); `MLFLOW_RUN_ID` points to the task nested run (not the parent run) |
```
hbelmiro: The "User container gets runtime MLflow env vars" scenario lists `MLFLOW_TRACKING_URI`, `MLFLOW_EXPERIMENT_ID`, `MLFLOW_RUN_ID`, and `MLFLOW_WORKSPACE` but omits `MLFLOW_TRACKING_AUTH`. The KEP's environment variable table shows `MLFLOW_TRACKING_AUTH` is set by the driver on user containers (set to `kubernetes` when `authType` is `"kubernetes"`). While section 1.12 covers this for specific auth modes, section 1.7's general compiled-workflow-injection scenario should list the complete set of driver-injected env vars.

```suggestion
| **P0** | User container gets runtime MLflow env vars | 1\. Run a pipeline task with MLflow enabled 2\. Inspect the launched user container env | User container receives `MLFLOW_TRACKING_URI`, `MLFLOW_EXPERIMENT_ID`, `MLFLOW_RUN_ID`, and `MLFLOW_TRACKING_AUTH` from the driver (plus `MLFLOW_WORKSPACE` when workspaces are enabled); `MLFLOW_RUN_ID` points to the task nested run (not the parent run) |
```

proposals/12862-mlflow-integration/TestPlan.md
```diff
@@ -0,0 +1,125 @@
...
+
+---
+
+## 1\. Test Plan
+
+### 1.1 CreateRun with MLflow
+
+| Priority | Scenario | Steps | Expected |
+| :---- | :---- | :---- | :---- |
+| **P0** | Default experiment | 1\. Configure global `plugins.mlflow` 2\. Create a run without `plugins_input` | Experiment **`KFP-Default`** (unless the admin default name overrides it); MLflow parent run exists; `plugins_output.mlflow` includes `experiment_id`, `root_run_id`, `run_url`; plugin `state` \= `PLUGIN_SUCCEEDED` |
```
hbelmiro: The proto at `backend/api/v2beta1/run.proto:215-243` defines `RuntimeState` with values: `RUNTIME_STATE_UNSPECIFIED`, `PENDING`, `RUNNING`, `SUCCEEDED`, `SKIPPED`, `FAILED`, `CANCELING`, `CANCELED`, `PAUSED`. The KEP also confirms: "On success, `state` is set to `SUCCEEDED`."

```suggestion
| **P0** | Default experiment | 1\. Configure global `plugins.mlflow` 2\. Create a run without `plugins_input` | Experiment **`KFP-Default`** (unless the admin default name overrides it); MLflow parent run exists; `plugins_output.mlflow` includes `experiment_id`, `root_run_id`, `run_url`; plugin `state` \= `SUCCEEDED` |
```

proposals/12862-mlflow-integration/TestPlan.md
```diff
@@ -0,0 +1,125 @@
...
+
+### 1.7 Compiled workflow injection
+
+| Priority | Scenario | Steps | Expected |
+| :---- | :---- | :---- | :---- |
+| **P0** | Env set only on driver and launcher | 1\. Inspect templates after compile | MLflow runtime env appears only on driver/launcher (not on arbitrary DAG or executor-only templates unless specified by design) |
+| **P0** | User container gets runtime MLflow env vars | 1\. Run a pipeline task with MLflow enabled 2\. Inspect the launched user container env | User container receives `MLFLOW_TRACKING_URI`, `MLFLOW_EXPERIMENT_ID`, and `MLFLOW_RUN_ID` from the driver (plus `MLFLOW_WORKSPACE` when enabled); `MLFLOW_RUN_ID` points to the task nested run (not the parent run) |
+| **P1** | Duplicate env keys | 1\. Template already defines the same env name | Injected value replaces consistently; no duplicate keys |
+| **P2** | MLflow globally disabled | 1\. No `plugins.mlflow` configuration | No MLflow env on workflows |
+
```
hbelmiro, line 78: The KEP explicitly states that `KFP_MLFLOW_CONFIG` is "an internal implementation detail and should not be exposed to user code" and that "the launcher will be updated to explicitly filter `KFP_MLFLOW_CONFIG` from the subprocess environment before launching the user command." No test scenario verifies that `KFP_MLFLOW_CONFIG` is absent from the user container's environment. Given that the KEP calls this out as an explicit implementation requirement, a negative assertion would prevent accidental leakage of internal config to user code.


```suggestion
| **P1** | `KFP_MLFLOW_CONFIG` not exposed to user code | 1\. Run a pipeline task with MLflow enabled 2\. Inspect the user container environment | `KFP_MLFLOW_CONFIG` is not present in the user process environment; only standard `MLFLOW_*` env vars are available |

```

hbelmiro (commented):
/lgtm

nsingla (approved):

PR #13264, opened 2026-04-13
fix(frontend): remove React 19 compatibility shim

jeffspahr:
## Summary
- replace remaining frontend JSX.Element annotations with React.JSX.Element and add type-only React namespace imports where needed
- remove the temporary React 19 compatibility shim at frontend/src/react19-compat.d.ts
- normalize mock-backend query param parsing so Express req.query values are narrowed before sorting, filtering, paging, or decodeURIComponent calls
- drop the mock-backend tsconfig include for the shim and update the React 18/19 checklist so only the remaining CustomTable.test.tsx act-warning cleanup is still listed as open

## Why
The React 19 upgrade still depended on a global JSX namespace shim to keep older JSX.Element annotations compiling. This change finishes that cleanup so the codebase uses React 19-native types directly and no longer needs the temporary compatibility file.

The branch also fixes the current mock-backend typecheck failure by making query handling explicit about Express query unions instead of treating every value as a bare string.

## Validation
- cd frontend && fnm exec --using .nvmrc -- npm run typecheck
- cd frontend && fnm exec --using .nvmrc -- npm run typecheck:mock-backend
- cd frontend && fnm exec --using .nvmrc -- npm run lint:ui
- cd frontend && fnm exec --using .nvmrc -- npm run check:react-peers
- cd frontend && fnm exec --using .nvmrc -- npm run test:ui -- src/components/Graph.test.tsx src/pages/ArtifactDetails.test.tsx src/pages/PipelineDetails.test.tsx


frontend/mock-backend/mock-api-middleware.ts
```diff
@@ -62,6 +62,29 @@ interface BaseResource {
...
+  const queryString = getQueryString(queryParam);
+  if (queryString === undefined || queryString === '') {
+    return undefined;
+  }
+  const queryNumber = Number(queryString);
+  return Number.isNaN(queryNumber) ? undefined : queryNumber;
+}
+
+function getDecodedQueryString(queryParam: unknown): string {
+  return decodeURIComponent(getQueryString(queryParam) || '');
```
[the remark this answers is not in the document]
jeffspahr: Addressed in `1f7a80754`.

`getDecodedQueryString` now returns `undefined` for missing or invalid decoded values, the required decode paths for `name`, `key`, and `podname` now return `400` instead of silently coercing bad input, and `frontend/server/integration-tests/mock-backend-query-validation.test.ts` covers the missing and invalid percent-encoding cases for those routes.

manaswinidas (changes requested):
I think we should change the PR title from [codex] to fix(frontend)

frontend/mock-backend/mock-api-middleware.ts
```diff
@@ -62,6 +62,62 @@ interface BaseResource {
...
+    return decodeURIComponent(queryString);
+  } catch {
+    return undefined;
+  }
+}
+
+function getRequiredDecodedQueryString(
+  res: Response,
+  queryParam: unknown,
+  queryParamName: string,
```
manaswinidas, line 99: `getRequiredDecodedQueryString` calls `getQueryString(queryParam)` on line 70, then calls `getDecodedQueryString(queryParam)` on line 76 which internally calls `getQueryString` again on the same input. Minor DRY nit — consider decoding the already-extracted string directly:

```suggestion
function getRequiredDecodedQueryString(
  res: Response,
  queryParam: unknown,
  queryParamName: string,
): string | undefined {
  const queryString = getQueryString(queryParam);
  if (queryString === undefined || queryString === '') {
    res.status(400).send(`${queryParamName} argument is required`);
    return undefined;
  }

  let decodedQueryString: string;
  try {
    decodedQueryString = decodeURIComponent(queryString);
  } catch {
    res.status(400).send(`${queryParamName} argument is invalid`);
    return undefined;
  }

  if (decodedQueryString === '') {
    res.status(400).send(`${queryParamName} argument is required`);
    return undefined;
  }

  return decodedQueryString;
}
```

Not blocking — just avoids the redundant re-extraction.

jeffspahr: Addressed in b47e2fdf2d3189f8a4620f12a88d9824347a96e6. `getRequiredDecodedQueryString(...)` now decodes the already-extracted `queryString` directly, so the required decoded path no longer re-calls `getQueryString(...)` through `getDecodedQueryString(...)`, while preserving the same required/invalid response behavior.

manaswinidas (approved):
/lgtm
/approve

PR #13265, opened 2026-04-13
fix(frontend): add mock backend route regression coverage

jeffspahr:
## Summary
- add a dedicated `frontend/server` route-level regression suite for `frontend/mock-backend/mock-api-middleware.ts`
- cover the current health, list/detail, mutation, file, and pod-log routes with direct request/response assertions
- factor shared mock-backend integration test setup into `frontend/server/integration-tests/test-helper.ts` so the route suite can reuse the same server bootstrapping and request helpers cleanly
- leave two current anomalous paths as `todo` coverage gaps rather than changing mock-backend behavior in this test-focused PR

## Why
The mock backend had no direct route-level regression tests. Recent work in this area was validated mostly through typecheck and indirect UI coverage, which made it easy to miss route-specific behavior changes.

This branch has also been rebased onto current `master` after [#13264](https://github.com/kubeflow/pipelines/pull/13264) merged on April 16, 2026. The React 18/19 checklist is already in its correct final state on `master`, so this PR intentionally no longer changes that file.

## Validation
- `cd frontend/server && npm test -- integration-tests/mock-backend-routes.test.ts`

### DIFF
--- a/frontend/server/integration-tests/mock-backend-routes.test.ts
+++ b/frontend/server/integration-tests/mock-backend-routes.test.ts
@@ -1,0 +1,478 @@
+// Copyright 2026 The Kubeflow Authors
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//      http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+import { afterAll, beforeAll, beforeEach, describe, expect, it, vi } from 'vitest';
+import express from 'express';
+import path from 'path';
+import { fileURLToPath } from 'url';
+import requests from 'supertest';
+
+const __dirname = path.dirname(fileURLToPath(import.meta.url));
+const frontendRoot = path.resolve(__dirname, '..', '..');
+const originalCwd = process.cwd();
+
+function buildQuery(query: Record<string, string | number | undefined>): string {
+  const searchParams = new URLSearchParams();
+  Object.entries(query).forEach(([key, value]) => {
+    if (value !== undefined) {
+      searchParams.set(key, String(value));
+    }
+  });
+  const queryString = searchParams.toString();
+  return queryString ? `?${queryString}` : '';
+}
+
+async function createRequest(): Promise<ReturnType<typeof requests>> {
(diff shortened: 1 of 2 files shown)


frontend/server/integration-tests/mock-backend-routes.test.ts
```diff
@@ -0,0 +1,478 @@
...
+
+function buildQuery(query: Record<string, string | number | undefined>): string {
+  const searchParams = new URLSearchParams();
+  Object.entries(query).forEach(([key, value]) => {
+    if (value !== undefined) {
+      searchParams.set(key, String(value));
+    }
+  });
+  const queryString = searchParams.toString();
+  return queryString ? `?${queryString}` : '';
```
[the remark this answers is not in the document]
jeffspahr: Addressed in 6c8ffb5051b3b535bc8d40443e834c204a310632. The test now imports `buildQuery` from `./test-helper.js`, and the shared helper was widened to accept `string | number | undefined`, so the duplication is removed without changing the existing server-test import convention.

manaswinidas (changes requested):
I think we should change the PR title from `[codex]` to `fix(frontend)`

frontend/server/integration-tests/mock-backend-routes.test.ts
```diff
@@ -0,0 +1,468 @@
...
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+import { afterAll, beforeAll, beforeEach, describe, expect, it, vi } from 'vitest';
+import express from 'express';
+import path from 'path';
+import { fileURLToPath } from 'url';
```
manaswinidas: Nit: `createRequest()`, the cwd setup (`beforeAll`/`afterAll`), and the `beforeEach` console suppression are duplicated verbatim with `mock-backend-query-validation.test.ts` from #13264. Consider extracting these into `test-helper.ts` (which already exists for shared test utilities) so both suites reuse a single `mockBackendSetup()` helper. Same for `asText`.

Not blocking — just consolidation to avoid divergence as more mock-backend test files land.

jeffspahr: Addressed in 1d2b1ac69ae34056bf0e1c52e2b25da12d96b748. The shared mock-backend harness now lives in `frontend/server/integration-tests/test-helper.ts` via `setupMockBackendTest()`, `createMockBackendRequest()`, and `asText()`, and `mock-backend-routes.test.ts` was switched over to those helpers to avoid divergence with the sibling suite.

manaswinidas (approved):
/lgtm
/approve

PR #13271, opened 2026-04-14
fix: replace bare except: with except Exception in metadata_helpers

harshadkhetpal:
## Summary

Bare `except:` clauses catch **everything**, including `KeyboardInterrupt` and `SystemExit`, which makes Ctrl-C unresponsive and swallows errors that should propagate. PEP 8 and flake8 rule E722 recommend catching `Exception` instead, which covers all runtime/user errors but leaves control-flow exceptions alone.

**Change:**
```python
# Before
try:
    ...
except:
    handle()

# After
try:
    ...
except Exception:
    handle()
```

## Testing
No behavior change for normal error paths — `Exception` is the parent of every non-system-exiting error. `KeyboardInterrupt` and `SystemExit` (which inherit from `BaseException`) now propagate correctly, which is the intended behavior.

### DIFF
--- a/backend/metadata_writer/src/metadata_helpers.py
+++ b/backend/metadata_writer/src/metadata_helpers.py
@@ -83,7 +83,7 @@ def get_or_create_artifact_type(store, type_name, properties: dict = None) -> me
     try:
         artifact_type = store.get_artifact_type(type_name=type_name)
         return artifact_type
-    except:
+    except Exception:
         artifact_type = metadata_store_pb2.ArtifactType(
             name=type_name,
             properties=properties,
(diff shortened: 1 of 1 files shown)


jeffspahr:
Thanks for improving the exception handling in metadata_helpers. Closing as obsolete on master: #13986 removed metadata-writer and the helper code changed by this PR from the KFP 3.0 architecture.

Release-2.18 retains metadata-writer. Context lookup and error handling for that line are being followed through #13713 and #14411, with changes requested and a request to target release-2.18. This closure does not claim that this patch was merged or backported.

PR #13276, opened 2026-04-14
fix: Fix duplicate entry in MLMD with Kubeflow driver

cbartram:

This PR fixes an issues where the driver will fail because of a duplicate key in MLMD for the `run/<runID>` when a pipeline component is retried by the workflow engine. 

If there is an error creating the execution in MLMD this fix will check if the error is a duplicate entry and try to lookup and return the existing execution entry instead.

**Checklist:**


ntny:
Hi, great catch  thanks for working on this!

I believe a similar issue might also exist in the container driver. It also creates an MLMD execution here:
https://github.com/kubeflow/pipelines/blob/master/backend/src/v2/driver/container.go#L181

It might make sense to reuse the existing there as well to avoid failure.

cbartram:
> Hi, great catch thanks for working on this!
> 
> I believe a similar issue might also exist in the container driver. It also creates an MLMD execution here: https://github.com/kubeflow/pipelines/blob/master/backend/src/v2/driver/container.go#L181
> 
> It might make sense to reuse the existing there as well to avoid failure.

Great point, I added the same check to the execution creation in `container.go` as well to catch any potential issues there.

ntny:
@zazulam @gmfrasca  
Could you please take a look?  
This handles duplicate errors during driver retries.
Note: the branch history includes multiple merges from master.

backend/src/v2/driver/container.go
```diff
@@ -180,6 +180,19 @@ func Container(ctx context.Context, opts Options, mlmd *metadata.Client, cacheCl
...
+				return nil, fmt.Errorf("failed to lookup existing execution: %w", lookupErr)
+			}
+
+			if existing == nil {
+				return nil, fmt.Errorf("execution already exists but lookup returned nil: %w", err)
+			}
+			glog.Infof("Found existing execution: %s", existing)
+			return &Execution{ID: existing.GetID()}, nil
+		}
 		return execution, err
```
jeffspahr: This handles the duplicate case as though it were a DAG execution, but this code just set `ecfg.ExecutionType = metadata.ContainerExecutionTypeName`. A retried container driver will not find the existing container execution. Also, returning a new `Execution` immediately skips the normal cache decision and pod spec patch generation below, so Argo can pass the default empty patch to the executor template. If the container driver handles duplicates, it needs to reuse the existing container execution and then continue through the rest of this function.

```suggestion
		if isAlreadyExistsErr(err) {
			glog.Infof("Execution %q already exists, looking up existing execution", ecfg.Name)
			existing, lookupErr := mlmd.GetExecutionByTypeAndName(ctx, string(metadata.ContainerExecutionTypeName), ecfg.Name)
			if lookupErr != nil {
				return execution, fmt.Errorf("failed to lookup existing execution: %w", lookupErr)
			}

			if existing == nil {
				return execution, fmt.Errorf("execution already exists but lookup returned nil: %w", err)
			}
			glog.Infof("Found existing execution: %s", existing)
			createdExecution = existing
		} else {
			return execution, err
		}
```

backend/src/v2/driver/util_test.go
```diff
@@ -58,6 +61,46 @@ func Test_isInputParameterChannel(t *testing.T) {
...
+			err:  status.Error(codes.Internal, "some other failure"),
+			want: false,
+		},
+	}
+
+	for _, test := range tests {
+		t.Run(test.name, func(t *testing.T) {
+			assert.Equal(t, test.want, isAlreadyExistsErr(test.err))
+		})
+	}
```
jeffspahr, line 101: This proves only the classifier helper. The behavior changed in `RootDAG` and `Container` is still untested, and this PR also removes the existing `Test_validateRootDAG` table. Please add branch-level coverage that drives `CreateExecution` returning `AlreadyExists`/duplicate and asserts the lookup behavior. Suggested targets: `RootDAG` uses `DagExecutionTypeName` and returns the existing ID; `Container` uses `ContainerExecutionTypeName` and still produces the cached decision / pod spec patch; lookup failure returns an actionable error.

jeffspahr (changes requested):
Found one blocker in the container retry path and one meaningful coverage gap. Focused local verification still passes (`go test ./backend/src/v2/metadata` and `go test ./backend/src/v2/driver`), so the blocker is semantic rather than a compile/test failure.

jeffspahr:
Re-reviewed the latest head (`bd2e43ee1e40fb02d1c3422f266d1fd49971457f`).

The container retry recovery issue is resolved: `Container` now looks up `metadata.ContainerExecutionTypeName` and continues through the normal cache / pod spec patch generation path.

The test coverage issue is still outstanding: the restored `Test_validateRootDAG` covers validation, but the duplicate-create recovery branches in `RootDAG` and `Container` still need focused tests. `Test_isAlreadyExistsErr` only covers the classifier helper and would not catch a wrong execution type lookup or an early return before container pod spec patch generation.

backend/src/v2/driver/util_test.go
cbartram: Addressed in: https://github.com/kubeflow/pipelines/pull/13276/commits/10567ac5ebc2f9eddd6ac7a2e92cc49ee3657647

backend/src/v2/driver/root_dag_test.go
```diff
@@ -166,3 +171,83 @@ func Test_validateRootDAG(t *testing.T) {
 		})
 	}
 }
+
+func TestRootDAG_CreateExecutionAlreadyExistsReturnsExistingID(t *testing.T) {
```
jeffspahr, line 179: These new cases call `RootDAG(...)` directly, but `RootDAG` still unconditionally loads the `kfp-launcher` ConfigMap before it reaches the MLMD duplicate-recovery branch. As written, the tests are not hermetic: `go test ./backend/src/v2/driver` now fails here with `Get "https://127.0.0.1:51785/api/v1/namespaces/kubeflow/configmaps/kfp-launcher": dial tcp 127.0.0.1:51785: connect: connection refused`. Please stub or inject the config dependency so the tests actually exercise the intended retry path.

backend/src/v2/driver/container_test.go
```diff
@@ -100,3 +107,269 @@ func Test_validateContainer(t *testing.T) {
...
+			return &pb.GetExecutionsByIDResponse{Executions: []*pb.Execution{{Id: func() *int64 { i := int64(55); return &i }()}}}, nil
+		},
+
+		// Trigger the AlreadyExists path
+		PutExecutionFunc: func(ctx context.Context, in *pb.PutExecutionRequest, opts ...grpc.CallOption) (*pb.PutExecutionResponse, error) {
+			return nil, status.Error(codes.AlreadyExists, "execution already exists")
+		},
+
+		// Return a successful lookup, but with NO executions (translates to nil existing execution)
+		GetExecutionsByTypeAndNameFunc: func(ctx context.Context, in *pb.GetExecutionByTypeAndNameRequest, opts ...grpc.CallOption) (*pb.GetExecutionByTypeAndNameResponse, error) {
```
jeffspahr: This still misses the container-side branch I called out earlier. `NewTestClient` calls `GetExecutionByTypeAndName`, but this test stubs `GetExecutionsByTypeAndNameFunc`, so it never exercises the intended lookup failure path. There is also still no success-path test proving that an `AlreadyExists` recovery continues through normal container output generation instead of stopping after the lookup.

jeffspahr (changes requested):
Re-reviewed at `10567ac5ebc2f9eddd6ac7a2e92cc49ee3657647`. The earlier container duplicate-recovery logic issue is fixed, but two issues remain:

- the new `RootDAG` duplicate-recovery tests are not isolated and currently break `go test ./backend/src/v2/driver`
- the container duplicate-recovery coverage still does not exercise the success path after an `AlreadyExists` recovery

Verification:
- `go test ./backend/src/v2/metadata`
- `go test ./backend/src/v2/driver`
- `go test ./backend/src/v2/driver -run TestContainer_CreateExecution`

backend/src/v2/driver/container_test.go
```diff
@@ -100,3 +108,322 @@ func Test_validateContainer(t *testing.T) {
...
+			CachingOptions: &pipelinespec.PipelineTaskSpec_CachingOptions{EnableCache: true},
+		},
+		Container: &pipelinespec.PipelineDeploymentConfig_PipelineContainerSpec{
+			Image:   "python:3.11",
+			Command: []string{"python", "main.py"},
+		},
+	}, mlmdClient, &mockCacheClient{})
+
+	require.NotNil(t, execution)
+	require.NoError(t, err)
```
jeffspahr, line 315: This now drives the `AlreadyExists` lookup-success path, but it still only asserts `NoError` and a non-nil execution. The earlier broken implementation would also satisfy those checks while still returning before cache / pod-spec generation. Please pin some post-lookup output from the normal tail of `Container`, for example:
```suggestion
	require.NotNil(t, execution)
	require.NoError(t, err)
	assert.Equal(t, int64(1234), execution.ID)
	require.NotNil(t, execution.Cached)
	assert.False(t, *execution.Cached)
	assert.NotEmpty(t, execution.PodSpecPatch)
```

jeffspahr (changes requested):
Re-reviewed at `28e6373589ba31aa64d4afa9ae46a2cd45e9da5b`. The earlier RootDAG test-isolation issue is fixed and `go test ./backend/src/v2/driver` is green again. One coverage gap remains in the new container duplicate-recovery success test.

Verification:
- `go test ./backend/src/v2/metadata`
- `go test ./backend/src/v2/driver`
- `go test ./backend/src/v2/driver -run 'TestRootDAG_CreateExecutionAlreadyExists|TestContainer_CreateExecution'`

jeffspahr (approved):
Re-reviewed at `541471ebd4c32ae07a2ec40c95d50d4bc98b81a4`. The earlier container success-path coverage gap is now addressed: the test asserts the recovered execution ID, cache state, and generated pod spec patch after the `AlreadyExists` lookup path.

No blocking issues found.

Verification:
- `go test ./backend/src/v2/metadata ./backend/src/v2/driver`
- `go test ./backend/src/v2/component ./backend/src/v2/cmd/driver`
- `go test ./backend/src/v2/...`

droctothorpe:
We should aim to have this merged by Wednesday or so if we want to include this in the 2.16.1 release.

jeffspahr:
Quick DCO check: all commits do have `Signed-off-by:` trailers. The failure is that three trailers do not exactly match the commit author identity.

DCO expects `Christian Bartram <cbartram3@gmail.com>` on these commits:

- `0376f33`
- `d30e56d`
- `bc45243`

Those commits currently have `Signed-off-by: cbartram <cbartram3@gmail.com>`.

The remaining commits have author `cbartram <cbartram3@gmail.com>`, so their trailers match.

cbartram:
> Quick DCO check: all commits do have `Signed-off-by:` trailers. The failure is that three trailers do not exactly match the commit author identity.
> 
> DCO expects `Christian Bartram <cbartram3@gmail.com>` on these commits:
> 
> * `0376f33`
> * `d30e56d`
> * `bc45243`
> 
> Those commits currently have `Signed-off-by: cbartram <cbartram3@gmail.com>`.
> 
> The remaining commits have author `cbartram <cbartram3@gmail.com>`, so their trailers match.

I just rebased and updated the trailer on those 3 commits.

PR #13281, opened 2026-04-16
refactor(frontend): remove NewRunParameters mount effect

jeffspahr:
## Summary
- remove the remaining mount-time `useEffect` from `NewRunParametersV2`
- keep `NewRunV2` as the owner of seeded initial parameter state on current master
- replace the old mount-callback assertions with pure `getInitialParameterState(...)` coverage and interaction-driven component assertions

## Why
`NewRunV2` already seeds `runtimeParameters`, `pipelineRoot`, and parameter validity before rendering `NewRunParametersV2`. The remaining child effect was only preserving standalone mount-time callback behavior, which kept a non-external-sync `useEffect` in the form component and duplicated state ownership.

This PR removes that last effect and tightens the tests around the current contract: initial parameter/default logic is verified as pure initialization, and parent callbacks are verified when the user edits values.

## Verification
- `cd frontend && fnm exec --using .nvmrc -- npx prettier@3.8.1 --check src/components/NewRunParametersV2.tsx src/components/NewRunParametersV2.test.tsx`
- `cd frontend && fnm exec --using .nvmrc -- npm run test:ui -- src/components/NewRunParametersV2.test.tsx`
- `cd frontend && fnm exec --using .nvmrc -- npm run test:ui -- src/pages/NewRunV2.test.tsx`
- `cd frontend && fnm exec --using .nvmrc -- npm run lint`

### DIFF
--- a/frontend/src/components/NewRunParametersV2.test.tsx
+++ b/frontend/src/components/NewRunParametersV2.test.tsx
@@ -118,6 +118,21 @@ describe('NewRunParametersV2', () => {
     expect(initialParameterState.errorMessages).toEqual({ intParam: 'Missing parameter.' });
   });
 
+  it('computes invalid initial state for literal parameter with no default', () => {
+    const initialParameterState = getInitialParameterState({
+      env: {
+        parameterType: ParameterType_ParameterTypeEnum.STRING,
+        literals: ['dev', 'staging', 'prod'],
+        isOptional: false,
+      },
+    });
+
+    expect(initialParameterState.isValid).toBe(false);
+    expect(initialParameterState.runtimeParameters).toEqual({});
+    expect(initialParameterState.updatedParameters).toEqual({});
+    expect(initialParameterState.errorMessages).toEqual({ env: 'Missing parameter.' });
+  });
+
   it('call convertInput function for string type with default value', () => {
     const handleParameterChangeSpy = vi.fn();
     const props = {
--- a/frontend/src/components/NewRunParametersV2.test.tsx
+++ b/frontend/src/components/NewRunParametersV2.test.tsx
@@ -19,6 +19,7 @@ import { CommonTestWrapper } from 'src/TestWrapper';
 import { fireEvent, render, screen, waitFor } from '@testing-library/react';
 import { ParameterType_ParameterTypeEnum } from 'src/generated/pipeline_spec/pipeline_spec';
 import NewRunParametersV2 from 'src/components/NewRunParametersV2';
+import { getInitialParameterState } from 'src/lib/NewRunParametersUtils';
 
 testBestPractices();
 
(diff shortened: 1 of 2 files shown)


frontend/src/components/NewRunParametersV2.test.tsx
```diff
@@ -19,6 +19,7 @@ import { CommonTestWrapper } from 'src/TestWrapper';
 import { fireEvent, render, screen, waitFor } from '@testing-library/react';
```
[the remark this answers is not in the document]
jeffspahr: Fixed in `fbfa4500a`. The unused `waitFor` import has been removed from `NewRunParametersV2.test.tsx`. Verification: `prettier --check`, `npm run test:ui -- src/components/NewRunParametersV2.test.tsx`, and `npm run lint:ui` all passed.

pnaik1:
looks good

frontend/src/components/NewRunParametersV2.test.tsx
```diff
@@ -90,6 +91,33 @@ describe('NewRunParametersV2', () => {
...
+        parameterType: ParameterType_ParameterTypeEnum.NUMBER_INTEGER,
+      },
+    });
+
+    expect(initialParameterState.isValid).toBe(false);
+    expect(initialParameterState.runtimeParameters).toEqual({});
+    expect(initialParameterState.updatedParameters).toEqual({});
+    expect(initialParameterState.errorMessages).toEqual({ intParam: 'Missing parameter.' });
+  });
+
```
manaswinidas, line 120: Nit: consider adding a matching test for a literal parameter with no default — that covers the scenario from the removed `marks input as invalid when literal parameter has no default value` test at the pure-function level:

```typescript
it('computes invalid initial state for literal param with no default', () => {
  const initialParameterState = getInitialParameterState({
    env: {
      parameterType: ParameterType_ParameterTypeEnum.STRING,
      literals: ['dev', 'staging', 'prod'],
      isOptional: false,
    },
  });

  expect(initialParameterState.isValid).toBe(false);
  expect(initialParameterState.errorMessages).toEqual({ env: 'Missing parameter.' });
});
```

`getInitialParameterState` doesn't look at `literals` today, but this would guard against regressions if literal-aware logic is ever added there.

jeffspahr: Added in `c33a11acf`. `NewRunParametersV2.test.tsx` now covers a required literal parameter with no default through `getInitialParameterState(...)`, including invalid state, empty runtime/updated parameters, and the `Missing parameter.` error. Verification: `prettier --check`, `npm run test:ui -- src/components/NewRunParametersV2.test.tsx`, and `npm run lint:ui` all passed.

manaswinidas (approved):
LGTM — the effect was already dead code for the only consumer (`NewRunV2` always passes `initialParameterState`), and the pure-function tests on `getInitialParameterState` are a cleaner way to cover the initialization logic.

Minor optional follow-up: consider making the `initialParameterState` prop required (or adding a JSDoc note) so future callers don't silently lose mount-time propagation. Not blocking.

manaswinidas (approved):
/approve

PR #13217, opened 2026-04-06
fix(frontend): surface silent useQuery errors to users

manaswinidas:
**Description of your changes:**

Several components using `@tanstack/react-query` silently ignore query failures, leaving users with no indication that data failed to load. This PR wires `isError`/`error` into the affected queries and surfaces failures through existing UI mechanisms (banners, inline messages).

### Changes

**`NewRunSwitcher.tsx`**
- Destructure `isError`/`error` from all six `useQuery` calls (run, recurring run, pipeline, pipeline version, v1 template, experiment).
- Add a `useEffect` that calls `updateBanner` on the first query failure, giving users a clear error message with details.
- Remove dead code: the `else` branch in the v1 template `queryFn` was unreachable after the guard that ensures `pipelineVersionId` is truthy.

**`RuntimeNodeDetailsV2.tsx` (TaskNodeDetail)**
- Destructure `isError`/`error` from the logs `useQuery`.
- Propagate the query-level error into `logsBannerMessage`/`logsBannerAdditionalInfo` so the Logs tab shows a failure banner instead of blank output.
- Remove unnecessary `React.Fragment` wrapper around `<Banner>`.
- Rename duplicate `ArtifactNodeDetailProps` interface to `ArtifactInfoProps` to resolve the TypeScript shadowing smell.

**`RunDetailsV2.tsx`**
- Destructure `isError`/`error` from the experiment `useQuery`.
- Add a `useEffect` that shows a warning banner on experiment fetch failure (gated behind `!isError` for the MLMD query to avoid overwriting a more critical error).
- Fix typo: `QUERY_REFETCH_INTERNAL` → `QUERY_REFETCH_INTERVAL`.

**`ExecutionDetails.tsx` (ExecutionReference)**
- Destructure `isError` from the context `useQuery`.
- Render an inline error row in the Reference table when the context query fails, instead of silently omitting the Pipeline Run link.

### Code quality fixes included
- **Dead code removal**: unreachable `else` branch in `NewRunSwitcher` v1 template query
- **DRY / naming**: renamed duplicate `ArtifactNodeDetailProps` interface in `RuntimeNodeDetailsV2`
- **Typo fix**: `QUERY_REFETCH_INTERNAL` → `QUERY_REFETCH_INTERVAL` in `RunDetailsV2`
- **Unnecessary wrapper**: removed redundant `React.Fragment` around single `<Banner>` element

**Checklist:**
[...]

PR #13291, opened 2026-04-21
fix(frontend): surface usePipelineVersionTemplate errors in run detail routers

manaswinidas:

`RunDetailsRouter` and `RecurringRunDetailsRouter` both call `usePipelineVersionTemplate()` but only destructure `isLoading` and `data`, silently dropping query failures. When the pipeline version fetch fails, the user sees no error feedback and falls through to a legacy or broken view.

### Changes

**`RunDetailsRouter.tsx`**
- Destructure `isError`/`error` from `usePipelineVersionTemplate`.
- Add a `useEffect` that calls `updateBanner` on template fetch failure with an error message and details.

**`RecurringRunDetailsRouter.tsx`**
- Destructure `isError`/`error` from `usePipelineVersionTemplate`.
- Add a `useEffect` that calls `updateBanner` on template fetch failure, gated behind `!getRecurringRunError` to avoid overwriting a more critical recurring run error.

### Tests added

- **`RunDetailsRouter.test.tsx`**: verifies `updateBanner` is called with the template error message when `getPipelineVersion` rejects.
- **`RecurringRunDetailsRouter.test.tsx`**: verifies `updateBanner` is called with the template error message when `getPipelineVersion` rejects.

This follows the same error-surfacing pattern established in PR #13217 for other `useQuery` consumers (`NewRunSwitcher`, `RuntimeNodeDetailsV2`, `RunDetailsV2`, `ExecutionReference`).

**Checklist:**

### DIFF
--- a/frontend/src/pages/RecurringRunDetailsRouter.tsx
+++ b/frontend/src/pages/RecurringRunDetailsRouter.tsx
@@ -62,8 +62,15 @@ export default function RecurringRunDetailsRouter(props: PageProps) {
   const pipelineId = v2RecurringRun?.pipeline_version_reference?.pipeline_id;
   const pipelineVersionId = v2RecurringRun?.pipeline_version_reference?.pipeline_version_id;
 
-  const { isLoading: templateStrIsLoading, data: templateStrFromPipelineVersion } =
-    usePipelineVersionTemplate(pipelineId, pipelineVersionId);
+  const {
+    isLoading: templateStrIsLoading,
+    isError: templateStrIsError,
+    error: templateStrError,
+    data: templateStrFromPipelineVersion,
+  } = usePipelineVersionTemplate(
+    pipelineManifest ? undefined : pipelineId,
+    pipelineManifest ? undefined : pipelineVersionId,
+  );
 
   const templateString = pipelineManifest ?? templateStrFromPipelineVersion;
 
@@ -85,6 +92,26 @@ export default function RecurringRunDetailsRouter(props: PageProps) {
     }
     return undefined;
   }, [getRecurringRunError, recurringRunError, recurringRunId, updateBanner]);
+
+  useEffect(() => {
+    if (templateStrIsError && templateStrError && !getRecurringRunError) {
+      let cancelled = false;
+      errorToMessage(templateStrError).then((msg) => {
+        if (!cancelled) {
+          updateBanner({
+            message:
+              'Error: failed to retrieve pipeline version template. Click Details for more information.',
+            mode: 'error',
+            additionalInfo: msg,
+          });
+        }
+      });
+      return () => {
+        cancelled = true;
+      };
+    }
+    return undefined;
+  }, [templateStrIsError, templateStrError, getRecurringRunError, updateBanner]);
 
   if (getRecurringRunSuccess && v2RecurringRun && templateString) {
     const isV2Pipeline = WorkflowUtils.isPipelineSpec(templateString);
--- a/frontend/src/pages/RunDetailsRouter.tsx
+++ b/frontend/src/pages/RunDetailsRouter.tsx
@@ -14,11 +14,13 @@
  * limitations under the License.
  */
 
+import { useEffect } from 'react';
 import * as JsYaml from 'js-yaml';
 import { useQuery } from '@tanstack/react-query';
 import { V2beta1Run } from 'src/apisv2beta1/run';
 import { RouteParams } from 'src/components/Router';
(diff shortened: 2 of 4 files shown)


frontend/src/pages/RunDetailsRouter.tsx
```diff
@@ -50,8 +53,32 @@ export default function RunDetailsRouter(
...
+            additionalInfo: msg,
+          });
+        }
+      });
+      return () => {
+        cancelled = true;
+      };
+    }
+    return undefined;
+  }, [templateStrIsError, templateStrError, updateBanner]);
```
jeffspahr, line 81: This path should not surface a template-fetch error when `v2Run.pipeline_spec` is already available, because the page renders from that inline spec and the template response is not used. Skipping the fallback query in that case avoids the misleading banner and the unnecessary request.

```suggestion
  const {
    isLoading: templateStrIsLoading,
    isError: templateStrIsError,
    error: templateStrError,
    data: templateStrFromPipelineVersion,
  } = usePipelineVersionTemplate(
    pipelineManifest ? undefined : pipelineId,
    pipelineManifest ? undefined : pipelineVersionId,
  );

  useEffect(() => {
    if (templateStrIsError && templateStrError) {
      let cancelled = false;
      errorToMessage(templateStrError).then((msg) => {
        if (!cancelled) {
          updateBanner({
            message:
              'Error: failed to retrieve pipeline version template. Click Details for more information.',
            mode: 'error',
            additionalInfo: msg,
          });
        }
      });
      return () => {
        cancelled = true;
      };
    }
    return undefined;
  }, [templateStrIsError, templateStrError, updateBanner]);
```

frontend/src/pages/RecurringRunDetailsRouter.tsx
```diff
@@ -86,6 +90,26 @@ export default function RecurringRunDetailsRouter(props: PageProps) {
...
+            additionalInfo: msg,
+          });
+        }
+      });
+      return () => {
+        cancelled = true;
+      };
+    }
+    return undefined;
+  }, [templateStrIsError, templateStrError, getRecurringRunError, updateBanner]);
```
jeffspahr, line 111: Same fallback issue here: when `v2RecurringRun.pipeline_spec` exists, the router uses that inline template, so a rejected pipeline-version fetch creates a false error banner. Skipping the fallback query while the inline spec is available keeps the banner tied to data the page actually needs.

```suggestion
  const {
    isLoading: templateStrIsLoading,
    isError: templateStrIsError,
    error: templateStrError,
    data: templateStrFromPipelineVersion,
  } = usePipelineVersionTemplate(
    pipelineManifest ? undefined : pipelineId,
    pipelineManifest ? undefined : pipelineVersionId,
  );

  const templateString = pipelineManifest ?? templateStrFromPipelineVersion;

  useEffect(() => {
    if (getRecurringRunError && recurringRunError) {
      let cancelled = false;
      errorToMessage(recurringRunError).then((msg) => {
        if (!cancelled) {
          updateBanner({
            message: `Error: failed to retrieve recurring run: ${recurringRunId}. Click Details for more information.`,
            mode: 'error',
            additionalInfo: msg,
          });
        }
      });
      return () => {
        cancelled = true;
      };
    }
    return undefined;
  }, [getRecurringRunError, recurringRunError, recurringRunId, updateBanner]);

  useEffect(() => {
    if (templateStrIsError && templateStrError && !getRecurringRunError) {
      let cancelled = false;
      errorToMessage(templateStrError).then((msg) => {
        if (!cancelled) {
          updateBanner({
            message:
              'Error: failed to retrieve pipeline version template. Click Details for more information.',
            mode: 'error',
            additionalInfo: msg,
          });
        }
      });
      return () => {
        cancelled = true;
      };
    }
    return undefined;
  }, [templateStrIsError, templateStrError, getRecurringRunError, updateBanner]);
```

jeffspahr (changes requested):
Thanks for the focused fix. I found one regression pattern in both routers: when an inline `pipeline_spec` is already available, the pipeline-version template fetch is only fallback data, so a failure there should not surface an error banner. Suggested changes below skip the unused fetch path and should be paired with regression tests for inline spec + rejected `getPipelineVersion`.

jeffspahr (approved):
Re-review complete. The inline-spec regression is addressed in both routers: the pipeline-version template query is now skipped when an inline pipeline_spec is already available, and the new regression tests cover the no-banner path when getPipelineVersion rejects in that inline-spec case. Focused verification passed for RunDetailsRouter/RecurringRunDetailsRouter tests and ESLint on the implementation files.

PR #13295, opened 2026-04-21
fix(ci): Add `proxy` parameter to e2e-test workflow deployment and test steps - Fixes #13294

### LINKED ISSUE #13294: chore(test): HTTP proxy must match between deploy and E2E Ginkgo (`-useProxy`)
## Chore description

For the **E2EProxy** matrix row (`proxy: "true"`), the E2E suite label **`E2EProxy`** runs `Create a pipeline run with HTTP proxy` (`env-var.yaml`). That spec branches on **`RunProxyTests`**, which is set from the Ginkgo flag **`-useProxy`** (see `backend/test/config/flags.go` — flag name `useProxy`).

```225:233:backend/test/end2end/pipeline_e2e_test.go
			if *config.RunProxyTests {
				// … validate compiled workflow + component statuses (success path)
			} else {
				runState := testutil.GetPipelineRun(runClient, &createdRunID).State
				expectedRunState := run_model.V2beta1RuntimeStateFAILED
				Expect(runState).To(Equal(&expectedRunState), fmt.Sprintf("Expected run with id=%s to fail with proxy=false", createdRunID))
			}
```

- **`-useProxy=true`:** test assumes HTTP proxy is **enabled** for the scenario and checks **success** plus workflow validation.
- **`-useProxy=false`:** test assumes **no** HTTP proxy and expects the run to end in **FAILED** (pipeline exercises `http_proxy`-style env; without the proxy-enabled deployment the run is expected to fail).

If Ginkgo runs with **`-useProxy=false`** but the cluster was deployed **with** HTTP proxy, the run can **succeed** while the test still executes the **`else`** branch and expects **FAILED**. Ginkgo then reports something like:

`Expected run with id=… to fail with proxy=false`

That is a **flag mismatch**, not an application regression in the PR under test.

## Root cause (workflow wiring)

1. **`.github/actions/test-and-report`** defaults **`inputs.proxy`** to **`'false'`** and passes **`USE_PROXY`** through to Ginkgo as **`-useProxy=$USE_PROXY`**.
2. Unless **`proxy: ${{ matrix.proxy }}`** is set on the **Run Tests** step, **E2EProxy** jobs always get **`-useProxy=false`**, regardless of matrix.
3. **`.github/actions/deploy`** also defaults **`proxy`** to **`'false'`**. The **Deploy KFP** step must pass **`proxy: ${{ matrix.proxy }}`** whenever the matrix includes **`proxy: "true"`**, or the job name says `proxy=true` but the cluster is still deployed without HTTP proxy — another form of inconsistency.

So: **matrix `proxy`, deploy `inputs.proxy`, and test `inputs.proxy` must agree** for the E2E proxy scenario.

**Note:** **`artifact_proxy`** on deploy (multi-user jobs) is **not** the same as HTTP **`proxy`**; it does not set **`-useProxy`**.

## Proposed fix

In **`.github/workflows/e2e-test.yml`**, job **`end-to-end-scenario-tests`**:

1. **Deploy KFP** — ensure **`proxy`** is passed when the matrix defines it, for example:

   ```yaml
   proxy: ${{ matrix.proxy }}
   ```

2. **Run Tests** (`test-and-report`) — same:

   ```yaml
   proxy: ${{ matrix.proxy }}
   ```

Confirm both steps use the same expression so **`proxy: "false"`** / **`proxy: "true"`** from the matrix reaches deploy and Ginkgo.

## Other workflows

- **`api-server-tests.yml`** (`api-test-standalone`): already passes **`proxy: ${{ matrix.proxy }}`** to deploy and **`test-and-report`**.

### Labels

/area testing

---

Love this idea? Give it a 👍.

hbelmiro:
Fixes https://github.com/kubeflow/pipelines/issues/13294.

**Checklist:**

### DIFF
--- a/.github/workflows/e2e-test.yml
+++ b/.github/workflows/e2e-test.yml
@@ -101,6 +101,7 @@ jobs:
         if: ${{ steps.create-cluster.outcome == 'success' }}
         id: deploy
         with:
+          proxy: ${{ matrix.proxy }}
           cache_enabled: ${{ matrix.cache_enabled }}
           argo_version: ${{ matrix.argo_version }}
           pod_to_pod_tls_enabled: ${{ matrix.pod_to_pod_tls_enabled }}
@@ -151,6 +152,7 @@ jobs:
         id: test-run
         if: ${{ steps.configure.outcome == 'success' }}
         with:
+          proxy: ${{ matrix.proxy }}
           cache_enabled: ${{ matrix.cache_enabled }}
           test_directory: ${{ env.E2E_TESTS_DIR }}
           test_label: ${{ steps.configure.outputs.TEST_LABEL }}


nsingla (approved):
lgtm

PR #13306, opened 2026-04-23
fix(backend): stop requiring workspace PVC storageClassName

mprahl:
## Summary

Fixes a regression from #12369.

- remove the compiler validation that required `workspace.pvcSpec.storageClassName` in `GetWorkspacePVC`
- keep `accessModes` as the only required PVC spec field in this code path
- allow clusters to rely on their default storage class when none is specified

## Test plan
- [x] `go test ./backend/src/v2/compiler/argocompiler -run TestGetWorkspacePVC -count=1`


nsingla (approved):
lgtm

PR #13307, opened 2026-04-23
chore(backend): Fix MLflow tags,nested run status update

VaniHaripriya:
This PR addressed the following issues identified in MLflow KFP integration:
- KFP run URL tags in MLflow are not accessible.
- plugins_output.mlflow.run_url used internal MLflow service URLs instead of externally accessible UI URLs.
- KFP run creation could still be impacted when MLflow was down.
- MLflow client retry loop continued retrying after context cancellation/deadline.
- Nested MLflow runs could remain open due to narrow terminal-sync status filtering.

**Checklist:**

### DIFF
--- a/backend/src/apiserver/plugins/mlflow/config_test.go
+++ b/backend/src/apiserver/plugins/mlflow/config_test.go
@@ -246,11 +246,22 @@ func TestMergePluginConfigAndSettingsDefaults(t *testing.T) {
 	global := commonmlflow.PluginConfig{
 		Endpoint: "https://global-mlflow.example.com",
 		Timeout:  "30s",
-		Settings: &commonmlflow.MLflowPluginSettings{ExperimentDescription: &globalDesc},
+		Settings: &commonmlflow.MLflowPluginSettings{
+			ExperimentDescription: &globalDesc,
+			KFPBaseURL:            "https://global-kfp.example.com",
+			KFPRunURLPathTemplate: "/global/{run_id}",
+			MLflowBaseURL:         "https://global-mlflow-ui.example.com",
+			MLflowUIPathPrefix:    "/global-mlflow",
+		},
 	}
 	namespace := &commonmlflow.PluginConfig{
 		Endpoint: "https://ns-mlflow.example.com",
-		Settings: &commonmlflow.MLflowPluginSettings{WorkspacesEnabled: &wsDisabled},
+		Settings: &commonmlflow.MLflowPluginSettings{
+			WorkspacesEnabled:     &wsDisabled,
+			KFPRunURLPathTemplate: "/ns/{namespace}/runs/{run_id}",
+			MLflowBaseURL:         "https://ns-mlflow-ui.example.com",
+			MLflowUIPathPrefix:    "/ns-mlflow",
+		},
 	}
 
 	merged := commonmlflow.MergePluginConfig(global, namespace)
@@ -263,6 +274,10 @@ func TestMergePluginConfigAndSettingsDefaults(t *testing.T) {
 	assert.False(t, *settings.WorkspacesEnabled)
 	require.NotNil(t, settings.ExperimentDescription)
 	assert.Equal(t, "Global desc", *settings.ExperimentDescription)
+	assert.Equal(t, "https://global-kfp.example.com", settings.KFPBaseURL)
+	assert.Equal(t, "/ns/{namespace}/runs/{run_id}", settings.KFPRunURLPathTemplate)
+	assert.Equal(t, "https://ns-mlflow-ui.example.com", settings.MLflowBaseURL)
+	assert.Equal(t, "/ns-mlflow", settings.MLflowUIPathPrefix)
 }
 
 func TestBuildMLflowRequestContextKubernetesAuth(t *testing.T) {
@@ -344,26 +359,47 @@ func TestEnsureExperimentExists(t *testing.T) {
 func TestBuildKFPTags(t *testing.T) {
 	run := &apiserverPlugins.PendingRun{
 		RunID:             "kfp-run-1",
+		Namespace:         "ns-1",
 		PipelineID:        "pipeline-1",
 		PipelineVersionID: "pipeline-version-1",
 	}
-	tags := BuildKFPTags(run, "")
+	tags := BuildKFPTags(run, "", "")
 	require.Len(t, tags, 4)
 	assert.Contains(t, tags, commonmlflow.Tag{Key: TagKFPRunID, Value: "kfp-run-1"})
-	assert.Contains(t, tags, commonmlflow.Tag{Key: TagKFPRunURL, Value: "/#/runs/details/kfp-run-1"})
+	assert.Contains(t, tags, commonmlflow.Tag{Key: TagKFPRunURL, Value: ""})
 	assert.Contains(t, tags, commonmlflow.Tag{Key: TagKFPPipelineID, Value: "pipeline-1"})
 	assert.Contains(t, tags, commonmlflow.Tag{Key: TagKFPPipelineVersionID, Value: "pipeline-version-1"})
 }
 
 func TestBuildKFPTags_WithBaseURL(t *testing.T) {
-	run := &apiserverPlugins.PendingRun{RunID: "run-1"}
-	tags := BuildKFPTags(run, "https://kfp.example.com")
+	run := &apiserverPlugins.PendingRun{
+		RunID:     "run-1",
+		Namespace: "ns-1",
+	}
+	tags := BuildKFPTags(run, "https://kfp.example.com", "")
 	require.Len(t, tags, 2)
-	assert.Contains(t, tags, commonmlflow.Tag{Key: TagKFPRunURL, Value: "https://kfp.example.com/#/runs/details/run-1"})
+	assert.Contains(t, tags, commonmlflow.Tag{
+		Key:   TagKFPRunURL,
+		Value: "https://kfp.example.com/#/runs/details/run-1",
+	})
+}
+
+func TestBuildKFPTags_WithCustomPathTemplate(t *testing.T) {
+	run := &apiserverPlugins.PendingRun{
+		RunID:     "run-b",
+		Namespace: "proj-1",
+	}
+	tmpl := "/demo/console/pipelines/{namespace}/runs/{run_id}"
+	tags := BuildKFPTags(run, "https://console.example.com", tmpl)
+	require.Len(t, tags, 2)
+	assert.Contains(t, tags, commonmlflow.Tag{
+		Key:   TagKFPRunURL,
+		Value: "https://console.example.com/demo/console/pipelines/proj-1/runs/run-b",
+	})
 }
 
 func TestBuildKFPTags_NilRun(t *testing.T) {
-	assert.Nil(t, BuildKFPTags(nil, ""))
+	assert.Nil(t, BuildKFPTags(nil, "", ""))
 }
 
 func TestCreateRunWithKFPTags(t *testing.T) {
@@ -390,7 +426,7 @@ func TestCreateRunWithKFPTags(t *testing.T) {
 		PipelineID:        "pipeline-1",
 		PipelineVersionID: "pipeline-version-1",
 	}
-	tags := BuildKFPTags(run, "")
+	tags := BuildKFPTags(run, "", "")
 	mlflowRunID, err := mlflowCtx.Client.CreateRun(context.Background(), "exp-1", "sample-run", tags)
 	require.NoError(t, err)
 	assert.Equal(t, "mlflow-parent-run-1", mlflowRunID)
--- a/backend/src/apiserver/plugins/mlflow/dispatcher.go
+++ b/backend/src/apiserver/plugins/mlflow/dispatcher.go
@@ -16,6 +16,7 @@ package mlflow
 
 import (
 	"context"
+	"time"
 
 	"github.com/golang/glog"
 	apiv2beta1 "github.com/kubeflow/pipelines/backend/api/v2beta1/go_client"
@@ -25,6 +26,8 @@ import (
 )
 
 var _ apiserverPlugins.PluginDispatcher = (*Dispatcher)(nil)
+
+const preRunMLflowTimeout = 3 * time.Second
 
 // Dispatcher implements PluginDispatcher for MLflow.
 type Dispatcher struct {
@@ -62,14 +65,24 @@ func (d *Dispatcher) OnBeforeRunCreation(ctx context.Context, run *apiserverPlug
 
 	resolvedCfg, err := ResolveMLflowRequestConfig(ctx, d.kubeClients, run.Namespace)
 	if err != nil {
-		return util.NewInternalServerError(err, "MLflow config resolution failed for run %q", run.RunID)
+		message := "MLflow config resolution failed; run creation will continue: " + err.Error()
+		glog.Warningf("MLflow OnBeforeRunCreation failed for run %q (%s)", run.RunID, message)
+		pluginOutput := FailedPluginOutput(selectedID, selectedName, "", "", "", message)
+		if outputErr := SetPendingRunPluginOutput(run, PluginName, pluginOutput); outputErr != nil {
+			glog.Warningf("Failed to persist MLflow plugin output for run %q: %v", run.RunID, outputErr)
+		}
+		return nil
 	}
 	if resolvedCfg == nil {
 		return nil
 	}
 
 	handler := NewHandler(mlflowInput, run.Namespace)
-	pluginOutput, pluginErr := handler.OnBeforeRunCreation(ctx, run, resolvedCfg.Config)
+	// Limit MLflow pre-run calls to a short timeout budget while still
+	// honoring parent request cancellation.
+	mlflowCtx, cancel := context.WithTimeout(ctx, preRunMLflowTimeout)
+	defer cancel()
+	pluginOutput, pluginErr := handler.OnBeforeRunCreation(mlflowCtx, run, resolvedCfg.Config)
 	if pluginErr != nil {
 		glog.Warningf("MLflow OnBeforeRunCreation failed for run %q (run creation will continue): %v", run.RunID, pluginErr)
 	}
--- a/backend/src/apiserver/plugins/mlflow/handler_test.go
+++ b/backend/src/apiserver/plugins/mlflow/handler_test.go
@@ -20,6 +20,7 @@ import (
 	"io"
 	"net/http"
 	"net/http/httptest"
+	"net/url"
 	"strings"
 	"testing"
 
@@ -345,10 +346,12 @@ func TestHandleRetry_Success(t *testing.T) {
 
 func TestBuildKFPRunURL(t *testing.T) {
 	tests := []struct {
-		name       string
-		runID      string
-		kfpBaseURL string
-		wantURL    string
+		name         string
+		runID        string
+		namespace    string
+		kfpBaseURL   string
+		pathTemplate string
+		wantURL      string
 	}{
 		{
 			name:    "empty runID returns empty",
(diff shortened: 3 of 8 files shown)


backend/src/common/mlflow/client.go
```diff
@@ -420,10 +385,8 @@ type retryRoundTripper struct {
 
 func (rt *retryRoundTripper) RoundTrip(req *http.Request) (*http.Response, error) {
 	// Skip retries for non-idempotent endpoints to avoid duplicates.
-	for _, path := range noRetryPaths {
-		if strings.HasSuffix(req.URL.Path, path) {
-			return rt.next.RoundTrip(req)
-		}
+	if noRetryPaths[req.URL.Path] {
```
hbelmiro: The `retryRoundTripper` decides whether to skip retries by checking `noRetryPaths[req.URL.Path]`, but the path stored in `noRetryPaths` is the API-relative path (e.g. `/api/2.0/mlflow/runs/create`). When `Client.buildURL()` constructs the full URL, it prepends the endpoint's own base path. If the MLflow endpoint is configured as `https://mlflow.example.com/some-proxy-prefix`, the resulting request path becomes `/some-proxy-prefix/api/2.0/mlflow/runs/create`, which does **not** match any key in `noRetryPaths`. This causes non-idempotent endpoints (runs/create, experiments/create) to be retried on transient errors, potentially creating duplicate MLflow experiments and runs.

**Why it matters:** Duplicate experiments and runs in MLflow are difficult to clean up and cause data integrity issues. Any deployment behind a reverse proxy or API gateway with a path prefix will hit this.

**Suggested fix:** Match on a suffix instead of an exact path:

```go
func isNoRetryPath(requestPath string) bool {
    for p := range noRetryPaths {
        if strings.HasSuffix(requestPath, p) {
            return true
        }
    }
    return false
}
```

Or, tag the request with a context value in `postJSON` before non-idempotent calls and check that in the round tripper instead of relying on path matching.

backend/src/apiserver/server/api_converter.go
hbelmiro, line 1622: Optional: `toApiRun` calls `common.GetPluginLimitsConfig()` on every invocation. This function is in the hot read path — it's called for every `GetRun` and every element in `ListRuns`. Each call does four `viper.GetString` lookups, `strconv.Atoi` conversions, and a cross-field comparison. In practice the config is immutable at runtime (baked into the pod via ConfigMap), so the repeated parsing is pure waste — but the cost is negligible (in-memory viper map lookups, nanosecond-scale), so I'll leave it up to you to decide if addressing it is worthy.

hbelmiro, line 1607: In `toApiRun`, when `validatePluginsOutputWithLimits` fails on data read *from the database*, the function returns a minimal error-only `Run` object containing only `RunId`, `ExperimentId`, and `Error` — all other run data (display name, state, pipeline details, timestamps, run details) is discarded. This means if an admin lowers the plugin limits config after runs were already stored with larger payloads, those runs become **permanently unreadable** via the API.

**Why it matters:** Size-limit validation on the read path retroactively rejects data that was accepted and stored under previous limits. This can make runs silently disappear from the UI. Validation of size limits should be enforced on the write path only.

**Suggested fix:** Remove `validatePluginsOutputWithLimits` from `toApiRun` (the read/response path). Keep it only in `toModelRun` (the write/request path) where it prevents oversized data from being persisted. The URL-safety validation for `RenderType == URL` entries could remain on the read path if desired.

backend/src/apiserver/plugins/mlflow/dispatcher.go
```diff
@@ -62,14 +65,24 @@ func (d *Dispatcher) OnBeforeRunCreation(ctx context.Context, run *apiserverPlug
...
 	}
 	if resolvedCfg == nil {
 		return nil
 	}
 
 	handler := NewHandler(mlflowInput, run.Namespace)
-	pluginOutput, pluginErr := handler.OnBeforeRunCreation(ctx, run, resolvedCfg.Config)
+	// Limit MLflow pre-run calls to a short timeout budget so MLflow outages
+	// do not exhaust the CreateRun request deadline.
+	mlflowCtx, cancel := context.WithTimeout(context.Background(), preRunMLflowTimeout)
```
hbelmiro: `OnBeforeRunCreation` creates a new context with `context.WithTimeout(context.Background(), preRunMLflowTimeout)`. While the short timeout (3s) is intentionally decoupled from the parent request context to prevent MLflow outages from blocking run creation, this also means the MLflow call **cannot be canceled** by the client. If the parent gRPC request is canceled (e.g. client disconnects), the MLflow experiment/run creation continues for up to 3 seconds doing work that will be discarded.

Consider deriving from the parent context instead of `context.Background()`:

```go
mlflowCtx, cancel := context.WithTimeout(ctx, preRunMLflowTimeout)
```

This caps MLflow calls at 3s while also respecting parent cancellation. The tradeoff is that if the parent has less than 3s remaining, the effective MLflow timeout is shorter than 3s — but that is appropriate since the result would be discarded anyway.

backend/src/apiserver/mlflow/dispatcher.go
hbelmiro, line 131: `executePostAction` returns `bool` but can fail for three distinct reasons: config resolution (line 137), MLflow API sync (line 148 via handler), or DB persist (line 155). All three are logged internally and collapsed into a single `false`. Returning `error` instead would let callers (`OnRunEnd`, `OnRunRetry`) see *what* failed without digging through logs, and follows Go conventions for functions that can fail.

**Suggested fix:** Change `executePostAction` to return `error`. `OnRunEnd` converts it to the retry-decision bool at the call site:

```go
if err := d.executePostAction(ctx, run, "OnRunEnd", ...); err != nil {
    glog.Warningf("MLflow OnRunEnd failed for run %q: %v", run.RunID, err)
    if hasParentRun { return false }
}
return true
```

backend/api/v2beta1/run.proto
hbelmiro, line 286: **KEP-12862 deviation:** The KEP specifies `RuntimeState state = 2` on `PluginOutput`, reusing the existing 9-value `RuntimeState` enum. This PR introduces a new `PluginState` enum with only 4 values (`PLUGIN_STATE_UNSPECIFIED`, `PLUGIN_RUNNING`, `PLUGIN_SUCCEEDED`, `PLUGIN_FAILED`).

The narrower enum is a reasonable design choice — plugins don't need `CANCELING`/`PAUSED`/`SKIPPED` states — but it changes the API surface from the approved KEP. Clients and the frontend that were designed against the KEP's `RuntimeState` contract will encounter a different type.

**Why it matters:** This is a public protobuf API change that affects forward compatibility with KEP 12700 (generic plugin architecture), which references the same `PluginOutput` message. If `PluginState` is the right choice, the KEP should be amended before this PR merges so the approved design stays in sync with the implementation.

**Suggested fix:** Either amend KEP-12862 to adopt `PluginState` (preferred if the narrower enum is intentional), or switch the proto to `RuntimeState` to match the approved KEP.

backend/src/common/mlflow/config.go
hbelmiro, line 82: **KEP-12862 gap:** The KEP defines three authentication modes (`"kubernetes"`, `"bearer"`, `"basic-auth"`) with an `AuthType` field on `MLflowPluginSettings` and a `CredentialSecretRef` struct for Secret-based credentials. This PR implements only `"kubernetes"` auth. The `AuthType` config field, the `CredentialSecretRef` struct, and all Secret-reading logic are absent.

This means the PR cannot support:
- MLflow deployments using token-based auth (including Databricks managed MLflow)
- MLflow deployments using built-in basic-auth
- Any deployment where the MLflow server is not behind a Kubernetes-native workspace provider

**Why it matters:** The KEP explicitly lists `"bearer"` and `"basic-auth"` as first-class auth modes, not future work. Shipping without them limits adoption to a single deployment topology.

**Suggested fix:** If `"bearer"` and `"basic-auth"` are planned for a follow-up PR on `mlflow-integration`, document that explicitly (e.g., in the PR description or a tracking issue). If they're intended for this PR, add the `AuthType` field, `CredentialSecretRef` struct, Secret-reading logic in `ResolveMLflowCredentials`, and the Argo compiler changes to mount credential Secrets via `valueFrom.secretKeyRef`.

backend/src/apiserver/mlflow/config.go
hbelmiro, line 37: **KEP-12862 deviation:** The KEP specifies defaulting to the `"Default"` MLflow experiment when the user doesn't provide an experiment name. The PR uses `"KFP-Default"` instead.

While `"KFP-Default"` is arguably more descriptive (avoids collision with MLflow's own built-in `"Default"` experiment), the KEP was approved with `"Default"` as the documented behavior. Users reading the KEP will expect `"Default"`.

**Suggested fix:** Either update the KEP to document `"KFP-Default"` as the new default, or change the constant to `"Default"` to match the approved KEP.

backend/src/common/mlflow/client_test.go
hbelmiro, line 450: The existing tests for `retryRoundTripper` (`TestDoWithRetry_SkipsRetryForRunsCreate`, `TestDoWithRetry_SkipsRetryForExperimentsCreate`) only use `httptest.NewServer` URLs, which have no base path prefix. This means the `noRetryPaths` bug is not caught by any test. A test with an endpoint like `https://mlflow.example.com/proxy-prefix` would fail today and would prevent regressions after the fix.

**Suggested fix:** Add test cases mirroring the existing ones but with a base-path endpoint. These fail today and would guard against the bug in the other comment:

```go
func TestDoWithRetry_SkipsRetryForRunsCreate_WithBasePath(t *testing.T) {
	var callCount int32
	server := httptest.NewServer(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
		atomic.AddInt32(&callCount, 1)
		w.WriteHeader(http.StatusServiceUnavailable)
		_, _ = w.Write([]byte(`{"error_code":"SERVICE_UNAVAILABLE","message":"retry later"}`))
	}))
	defer server.Close()

	// Endpoint with a base path, simulating a reverse proxy / API gateway.
	c := newTestClientWithFastRetry(t, server.URL+"/gateway/mlflow")
	_, err := c.CreateRun(context.Background(), "exp-1", "test-run", nil)
	require.Error(t, err)
	// Non-idempotent endpoint: should be called exactly once, no retries.
	assert.Equal(t, int32(1), atomic.LoadInt32(&callCount))
}

func TestDoWithRetry_SkipsRetryForExperimentsCreate_WithBasePath(t *testing.T) {
	var callCount int32
	server := httptest.NewServer(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
		atomic.AddInt32(&callCount, 1)
		w.WriteHeader(http.StatusServiceUnavailable)
		_, _ = w.Write([]byte(`{"error_code":"SERVICE_UNAVAILABLE","message":"retry later"}`))
	}))
	defer server.Close()

	c := newTestClientWithFastRetry(t, server.URL+"/gateway/mlflow")
	_, err := c.CreateExperiment(context.Background(), "test-exp", nil)
	require.Error(t, err)
	assert.Equal(t, int32(1), atomic.LoadInt32(&callCount))
}
```

backend/src/apiserver/plugins/mlflow/run_start.go
```diff
@@ -139,15 +139,17 @@ func CreateExperiment(ctx context.Context, requestCtx *commonmlflow.RequestConte
 }
 
 // BuildKFPRunURL constructs a KFP pipeline run URL.
-func BuildKFPRunURL(runID, kfpBaseURL string) string {
-	if runID == "" {
+func BuildKFPRunURL(runID, namespace, kfpBaseURL string) string {
+	if runID == "" || namespace == "" || kfpBaseURL == "" {
```
alyssacgoins: Nit: potentially helpful to add a log explaining why returned URL is empty

backend/src/apiserver/plugins/mlflow/run_start.go
```diff
@@ -168,8 +170,24 @@ func BuildKFPTags(run *apiserverPlugins.PendingRun, kfpBaseURL string) []commonm
 	return tags
 }
 
-func BuildRunURL(requestCtx *commonmlflow.RequestContext, experimentID, runID string) string {
-	if requestCtx == nil || requestCtx.BaseURL == nil || experimentID == "" || runID == "" {
+// BuildRunURL constructs an MLflow run URL for plugin output metadata.
+func BuildRunURL(requestCtx *commonmlflow.RequestContext, experimentID, runID, kfpBaseURL string) string {
+	if experimentID == "" || runID == "" {
```
alyssacgoins, line 214: Same as above ^ potentially helpful here to add a log explaining why returned URL is empty

backend/src/apiserver/plugins/mlflow/run_start.go
VaniHaripriya: Added a log statement here.

backend/src/apiserver/plugins/mlflow/run_start.go
VaniHaripriya: Added a log statement here.

backend/src/apiserver/plugins/mlflow/dispatcher.go
VaniHaripriya: Great point.
I switched the pre-run MLflow context from `context.Background() `to the parent request context: `context.WithTimeout(ctx, preRunMLflowTimeout)`.

backend/src/apiserver/plugins/mlflow/run_start.go
```diff
@@ -139,15 +140,23 @@ func CreateExperiment(ctx context.Context, requestCtx *commonmlflow.RequestConte
...
 		return ""
 	}
-	path := fmt.Sprintf("/#/runs/details/%s", runID)
-	if kfpBaseURL != "" {
-		return strings.TrimRight(kfpBaseURL, "/") + path
-	}
-	return path
+	trimmedBaseURL := strings.TrimRight(kfpBaseURL, "/")
+	return fmt.Sprintf(
+		"%s/develop-train/pipelines/runs/%s/runs/%s",
```
hbelmiro: `BuildKFPRunURL` hardcodes the path segment `/develop-train/pipelines/runs/{namespace}/runs/{runID}`. This appears to be specific to OpenShift AI / Red Hat's dashboard UI and is not the standard KFP UI path. Upstream KFP's React UI uses hash routing (`/#/runs/details/{runID}`). Any non-RHOAI deployment that configures `KFPBaseURL` will get URLs that produce 404s in their KFP frontend.

backend/src/apiserver/plugins/mlflow/run_start.go
```diff
@@ -168,8 +177,34 @@ func BuildKFPTags(run *apiserverPlugins.PendingRun, kfpBaseURL string) []commonm
...
+		base := strings.TrimRight(kfpBaseURL, "/")
+		fragmentRoute := fmt.Sprintf(
+			"/experiments/%s/runs/%s",
+			url.PathEscape(experimentID),
+			url.PathEscape(runID),
+		)
+		if requestCtx != nil && requestCtx.WorkspacesEnabled && requestCtx.Workspace != "" {
+			fragmentRoute = fmt.Sprintf("%s?workspace=%s", fragmentRoute, url.QueryEscape(requestCtx.Workspace))
+		}
+		return fmt.Sprintf("%s/mlflow/#%s", base, fragmentRoute)
```
hbelmiro: `BuildRunURL` hardcodes `/mlflow/#` as the path to the MLflow UI when `kfpBaseURL` is provided. This assumes the MLflow UI is hosted at `{kfpBaseURL}/mlflow/` — an assumption that only holds for specific platform deployments (e.g., RHOAI). For standalone MLflow deployments or deployments behind different reverse proxy paths, this URL will be wrong.

The existing fallback path that uses `requestCtx.BaseURL` is the correct, deployment-neutral approach since it derives the MLflow URL from the actual MLflow endpoint configuration.

backend/src/apiserver/plugins/mlflow/handler_test.go
```diff
@@ -358,29 +359,96 @@ func TestBuildKFPRunURL(t *testing.T) {
...
 	for _, tt := range tests {
 		t.Run(tt.name, func(t *testing.T) {
-			got := BuildKFPRunURL(tt.runID, tt.kfpBaseURL)
+			got := BuildKFPRunURL(tt.runID, tt.namespace, tt.kfpBaseURL)
 			assert.Equal(t, tt.wantURL, got)
 		})
 	}
 }
 
+func TestBuildRunURL(t *testing.T) {
```
hbelmiro, line 413: `TestBuildRunURL` covers the `kfpBaseURL` branch and the "empty everything" branch, but no test case exercises the `requestCtx.BaseURL` fallback path. This is the original code path that uses the MLflow endpoint URL to construct the run link, and it's now the secondary path when `kfpBaseURL` takes precedence. There is no test that passes `kfpBaseURL: ""` with a non-nil `requestCtx.BaseURL` to verify the fallback behavior.

backend/src/apiserver/plugins/mlflow/dispatcher.go
```diff
@@ -26,6 +27,8 @@ import (
 
 var _ apiserverPlugins.PluginDispatcher = (*Dispatcher)(nil)
 
+const preRunMLflowTimeout = 3 * time.Second
```
nsingla, line 30: should this be here, or somewhere we can define common plugin constants/configs?

VaniHaripriya: Keeping this here is reasonable since it’s only used for MLflow pre–CreateRun timeout.

backend/src/apiserver/plugins/mlflow/run_start.go
VaniHaripriya: Good catch,updated BuildRunURL no longer uses kfpBaseURL. run_url is built from plugins.mlflow.endpoint (via requestCtx.BaseURL), optionally overridden by settings.mlflowBaseURL and settings.mlflowUIPathPrefix. kfpBaseURL is only used for KFP links in BuildKFPRunURL.

backend/src/apiserver/plugins/mlflow/handler_test.go
VaniHaripriya: Updated the test.

backend/src/apiserver/plugins/mlflow/run_start.go
VaniHaripriya: Good catch, fixed this by adding a new field `kfpRunURLPathTemplate` which specifies platform specific run url path in the plugin config settings.

backend/src/common/plugins/mlflow/config.go
```diff
@@ -81,6 +84,9 @@ type MLflowPluginSettings struct {
 	ExperimentDescription *string `json:"experimentDescription,omitempty"`
 	DefaultExperimentName string  `json:"defaultExperimentName,omitempty"`
 	KFPBaseURL            string  `json:"kfpBaseURL,omitempty"`
+	KFPRunURLPathTemplate string  `json:"kfpRunURLPathTemplate,omitempty"`
+	MLflowBaseURL         string  `json:"mlflowBaseURL,omitempty"`
+	MlflowUIPathPrefix    string  `json:"mlflowUIPathPrefix,omitempty"`
```
hbelmiro:
```suggestion
	MLflowUIPathPrefix    string  `json:"mlflowUIPathPrefix,omitempty"`
```

backend/src/apiserver/plugins/mlflow/config_test.go
hbelmiro, line 243: This can be done as a follow up PR if you like. Please create an issue in case you want to address it later.

`TestMergePluginConfigAndSettingsDefaults` does not cover the three new settings fields added in this PR: `KFPRunURLPathTemplate`, `MLflowBaseURL`, and `MlflowUIPathPrefix`. The `mergeSettings` function now handles these fields, but no test verifies that namespace-level values override global values for them.

**Suggested fix:** Extend the existing test (or add a new one) to set these fields on both global and namespace configs and assert the namespace values win:

```go
func TestMergeSettingsNewURLFields(t *testing.T) {
    global := commonmlflow.PluginConfig{
        Endpoint: "https://global.example.com",
        Timeout:  "30s",
        Settings: &commonmlflow.MLflowPluginSettings{
            KFPBaseURL:            "https://global-kfp.example.com",
            KFPRunURLPathTemplate: "/global/{run_id}",
            MLflowBaseURL:         "https://global-mlflow.example.com",
            MlflowUIPathPrefix:    "/global-mlflow",
        },
    }
    namespace := &commonmlflow.PluginConfig{
        Settings: &commonmlflow.MLflowPluginSettings{
            KFPRunURLPathTemplate: "/ns/{namespace}/runs/{run_id}",
            MLflowBaseURL:         "https://ns-mlflow.example.com",
            MlflowUIPathPrefix:    "/ns-mlflow",
        },
    }
    merged := commonmlflow.MergePluginConfig(global, namespace)
    require.NotNil(t, merged.Settings)
    assert.Equal(t, "https://global-kfp.example.com", merged.Settings.KFPBaseURL)
    assert.Equal(t, "/ns/{namespace}/runs/{run_id}", merged.Settings.KFPRunURLPathTemplate)
    assert.Equal(t, "https://ns-mlflow.example.com", merged.Settings.MLflowBaseURL)
    assert.Equal(t, "/ns-mlflow", merged.Settings.MlflowUIPathPrefix)
}
```

VaniHaripriya: Extended the test to include new fields.

backend/src/common/plugins/mlflow/config.go
VaniHaripriya: Updated to use `MLflowUIPathPrefix`.

hbelmiro (approved):
/lgtm
/approve

PR #13320, opened 2026-04-27
fix(frontend): download v2 directory artifacts as tar.gz. Fixes #7809

### LINKED ISSUE #7809: [frontend] Error when downloading Artifact that is Folder via UI
### Environment

AWS EKS k8s cluster + vanilla installation with multi-user auth
*  KFP version: 
1.7

### Steps to reproduce

In Kubeflow pipelines V2 when an artifact is a folder it is not possible to download the Artifact via UI. Data is passed between components in a pipeline fine. When we are using V1 syntax folder (and non folders) are packed in tar archive and everything works in the same environment.

### Expected result

Artifact is packed automatically in archive or just downloadable as a folder when it is a folder originally

### Materials and Reference

There is closed issues on github, but I think it was a fix only for pipelines V1:
https://github.com/kubeflow/pipelines/issues/3667
https://github.com/kubeflow/pipelines/issues/4799

Here is ho it looks on UI when we are trying to download an artifact that was a folder:
image

---

Impacted by this bug? Give it a 👍. We prioritise the issues with the most 👍.

zswf-mle:

KFP v2 output artifacts can be directories (S3 prefixes) rather than single
objects. The artifact download handler currently fails with `500 NoSuchKey`
when a user clicks download on a directory artifact.

This adds a fallback: when `getObject` returns NoSuchKey, the handler lists
the prefix and streams the contents back as `<artifact>.tar.gz`.

Implementation notes:

- New `isNoSuchKeyError` helper normalizes the error code across providers
  (`NoSuchKey` / `NotFound` / message-based).
- New `listObjectsUnderPrefix` helper pages via the lower-level
  `listObjectsV2Query` at 300 keys/page. The public `listObjectsV2` API
  hard-codes `maxKeys=1000`, but minio's bundled fast-xml-parser caps
  entity expansions at 1000 — each Contents entry has ~2 `&quot;` entities
  in its ETag, so 1000-key pages fail with "Entity expansion limit
  exceeded" once a directory holds more than ~500 objects.
- `tar-stream` is already a dependency; `zlib` is a Node built-in. No new
  deps.

Fixes #7809

**Checklist:**
- [x] The title for your pull request (PR) follows our title convention

### DIFF
--- a/frontend/server/handlers/artifacts.ts
+++ b/frontend/server/handlers/artifacts.ts
@@ -24,6 +24,7 @@ import {
   getObjectStream,
   isNoSuchKeyError,
   listObjectsUnderPrefix,
+  summarizeDirectoryUnderPrefix,
 } from '../minio-helper.js';
 import * as tar from 'tar-stream';
 import * as zlib from 'zlib';
@@ -380,6 +381,24 @@ function getMinioArtifactHandler(
       // a .tar.gz so users can still download them. See
       // https://github.com/kubeflow/pipelines/issues/7809
       if (isNoSuchKeyError(err)) {
+        if (peek > 0) {
+          // Preview request (e.g. the run details panel calls
+          // Apis.readFile with a small peek size). We must not stream a
+          // full directory archive here — that would list every object
+          // under the prefix and gzip the whole tree just to render a few
+          // KB of inline text. Instead, answer with a small text summary
+          // backed by one capped ListObjectsV2 call: cost stays bounded
+          // (one round trip, <1KB body) and the user sees that the
+          // artifact is a directory with N files.
+          try {
+            await previewDirectorySummary(options, res);
+            return;
+          } catch (summaryErr) {
+            console.error(summaryErr);
+            res.status(500).send(`Failed to summarize directory: ${summaryErr}`);
+            return;
+          }
+        }
         try {
           await streamDirectoryAsTarGz(options, res);
           return;
@@ -397,6 +416,25 @@ function getMinioArtifactHandler(
       res.status(500).send(`Failed to get object in bucket: ${err}`);
     }
   };
+}
+
+async function previewDirectorySummary(
+  options: { bucket: string; key: string; client: MinioClient },
+  res: Response,
+) {
+  const { bucket, key, client } = options;
+  // Trailing slash so prefix "foo" doesn't also match sibling key "foobar".
+  const prefix = key.endsWith('/') ? key : `${key}/`;
+  const summary = await summarizeDirectoryUnderPrefix(client, bucket, prefix);
+  if (!summary) {
+    res.status(404).send(`No objects found at ${bucket}/${key}`);
+    return;
+  }
+  const baseName = key.replace(/\/+$/, '').split('/').pop() || 'artifact';
+  const countLabel = `${summary.count}${summary.truncated ? '+' : ''}`;
+  res
+    .type('text/plain')
+    .send(`Directory artifact "${baseName}" — ${countLabel} file(s). Download to view contents.\n`);
 }
 
 async function streamDirectoryAsTarGz(
--- a/frontend/server/handlers/artifacts.ts
+++ b/frontend/server/handlers/artifacts.ts
@@ -19,7 +19,14 @@ import {
   parseJSONString,
   isAllowedResourceName,
 } from '../utils.js';
-import { createMinioClient, getObjectStream } from '../minio-helper.js';
+import {
+  createMinioClient,
+  getObjectStream,
+  isNoSuchKeyError,
+  listObjectsUnderPrefix,
+} from '../minio-helper.js';
+import * as tar from 'tar-stream';
+import * as zlib from 'zlib';
 import * as serverInfo from '../helpers/server-info.js';
 import { Handler, Request, Response, NextFunction } from 'express';
 import { createProxyMiddleware } from 'http-proxy-middleware';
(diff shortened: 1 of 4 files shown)


zswf-mle:
Addressed review feedback in 5cc9c3bf4. Summary:

- **Tar-slip protection** — tar entry names are now sanitized: leading slashes stripped, `.` / `..` segments removed before being passed to `pack.entry`.
- **Runtime guard for `listObjectsV2Query`** — the internal-API access is now wrapped in a typed accessor with a runtime check that throws a clear error if the minio bundle ever drops the method, instead of failing with `undefined is not a function` deep in a stream pipeline.
- **Streaming pagination** — `listObjectsUnderPrefix` is now an `async function*` and `streamDirectoryAsTarGz` peeks the first object before sending headers (so empty prefixes still 404), then writes entries as they're yielded. No more buffering every key in memory before the first byte.
- **Safe Content-Disposition** — added a small RFC 5987 helper that emits an ASCII-only `filename="…"` fallback alongside `filename*=UTF-8''…`. No new dep; `content-disposition` is only present transitively via express.
- **Tests** —
  - Unit: `isNoSuchKeyError` (lowercase `code`, `Code`, message-substring fallback, non-object inputs); `listObjectsUnderPrefix` (single page, multi-page continuation, default-size + missing-name handling, runtime guard).
  - Integration: directory fallback happy path producing a valid `.tar.gz`, empty prefix → 404, tar-slip stripping, safe Content-Disposition for keys with spaces / quotes / non-ASCII.

All 51 tests pass locally; project typecheck and prettier clean.

mprahl:
@jeffspahr could you please take a look?

frontend/server/minio-helper.ts
```diff
@@ -315,3 +315,90 @@ export async function getObjectStream({
...
+}
+
+type ListObjectsV2Query = (
+  bucket: string,
+  prefix: string,
+  continuationToken: string,
+  delimiter: string,
+  maxKeys: number,
+  startAfter: string,
+) => NodeJS.EventEmitter;
```
jeffspahr: `minio@8.0.7` declares and implements `listObjectsV2Query` as returning a Promise, not an EventEmitter. Typing the internal helper that way lets TypeScript catch the runtime shape and avoids building the Promise wrapper below around a non-stream.

```suggestion
type ListObjectsV2QueryResult = {
  objects: Array<{ name?: string; size?: number }>;
  isTruncated: boolean;
  nextContinuationToken: string;
};

type ListObjectsV2Query = (
  bucket: string,
  prefix: string,
  continuationToken: string,
  delimiter: string,
  maxKeys: number,
  startAfter: string,
) => Promise<ListObjectsV2QueryResult>;
```

jeffspahr (commented):
I found one blocking runtime issue in the directory-artifact fallback. The inline suggestions update the helper to match the real `minio@8.0.7` return shape; the tests should also mock `listObjectsV2Query` as a Promise-returning helper so this does not regress.

frontend/server/minio-helper.ts
```diff
@@ -315,3 +315,90 @@ export async function getObjectStream({
...
+  while (isTruncated) {
+    const page = await new Promise<{
+      objects: Array<{ name?: string; size?: number }>;
+      isTruncated: boolean;
+      nextContinuationToken: string;
+    }>((resolve, reject) => {
+      const stream = listObjectsV2Query(bucket, prefix, continuationToken, '', PAGE_SIZE, '');
+      stream.on('error', reject);
+      stream.on('data', resolve);
+    });
```
jeffspahr: The locked `minio@8.0.7` implementation returns `Promise<ListObjectV2Res>` from `listObjectsV2Query`, not an EventEmitter. This code assigns that Promise to `stream` and calls `stream.on(...)`, so the new directory-artifact fallback throws `TypeError: stream.on is not a function` before it can list or tar any objects. The tests currently mask this by mocking `listObjectsV2Query` as an EventEmitter; please update those mocks to return resolved page Promises too.

```suggestion
    const page = await listObjectsV2Query(bucket, prefix, continuationToken, '', PAGE_SIZE, '');
```

zswf-mle:
Good catch @jeffspahr. Confirmed against the locked minio@8.0.7 source. `listObjectsV2Query` is now declared `async` and returns `Promise<ListObjectV2Res>`.

Fixed in e45070fa1:
- `ListObjectsV2Query` is now typed as `Promise<ListObjectsV2QueryResult>`.
- The page fetch is just `const page = await listObjectsV2Query(...)` — no more EventEmitter wrapper.
- Both unit and integration mocks now resolve a page record instead of emitting one, so the test shape matches the real client.

Re-ran the suites against the real `minio@8.0.7` from `package-lock.json` — all 51 tests still pass.

jeffspahr (changes requested):
I found one blocking issue in the directory-artifact fallback: preview requests now package and stream the entire directory archive instead of honoring `peek`, which can make opening artifact details unexpectedly download large outputs. The targeted server tests and build pass on the PR head, but this preview path needs a regression test before merge.

`frontend/server/handlers/artifacts.ts:384`: This fallback also runs for `/artifacts/get` preview requests. `MinioArtifactPreview` calls `Apis.readFile(..., peek=maxbytes + 1)`, and `Apis._fetch` buffers the response as text, but this branch ignores the `peek` value and calls `streamDirectoryAsTarGz` with no limit. For a large directory artifact, simply opening the run details page can list/read every object and send the full gzip archive to the browser just to render the small preview. Please either skip the tar fallback for `peek` requests or pass `peek` through and cap the tar/gzip stream, with a regression test for `/artifacts/get?...&peek=...`.

frontend/server/handlers/artifacts.ts
```diff
@@ -368,12 +375,115 @@ function getMinioArtifactHandler(
         .pipe(new PreviewStream({ peek }))
         .pipe(res);
     } catch (err) {
+      // In KFP v2, output artifacts may be directories (prefixes) rather than
+      // single objects. Fall back to packaging the contents of the prefix as
+      // a .tar.gz so users can still download them. See
+      // https://github.com/kubeflow/pipelines/issues/7809
+      if (isNoSuchKeyError(err)) {
+        try {
+          await streamDirectoryAsTarGz(options, res);
```
jeffspahr, line 403: This fallback also runs for `/artifacts/get` preview requests. `MinioArtifactPreview` calls `Apis.readFile(..., peek=maxbytes + 1)`, and `Apis._fetch` buffers the response as text, but this branch ignores the `peek` value and calls `streamDirectoryAsTarGz` with no limit. For a large directory artifact, simply opening the run details page can list/read every object and send the full gzip archive to the browser just to render the small preview.

One safe way to resolve this is to keep the directory tar fallback only on download requests (`tryExtract === false`) and return a bounded/404-style preview response for `/artifacts/get?...&peek=...`; alternatively, pass `peek` through and cap the bytes emitted by the gzip stream. Please add a regression test for `/artifacts/get?source=minio&bucket=ml-pipeline&key=directory&peek=...` so preview requests cannot stream the full archive again.

zswf-mle:
Hey @jeffspahr 

Addressed in aa890edd7.

The preview path (`/artifacts/get?...&peek=N`) no longer streams the directory archive. Instead it answers with one capped `listObjectsV2Query` call (no pagination) and returns a small text body, e.g.:

> `Directory artifact "model" — 5 file(s). Download to view contents.`

For directories larger than the cap (currently 50), the count reads `50+`. Empty prefixes still return 404. Downloads (`peek === 0`) keep the full `.tar.gz` fallback.

**Cost guarantees, verified by the new tests:**
- Preview of any directory: 1 `listObjectsV2Query` call (max 50 keys) + the original failing `getObject`. Body is ~100 bytes regardless of directory size.
- No per-child `getObject` fetches on the preview path.
- The truncated-count test confirms the helper does not loop on `nextContinuationToken` even when minio reports more pages exist.

**Tests added:**
- `minio-helper.test.ts`: 4 unit tests for the new `summarizeDirectoryUnderPrefix` (complete page, truncated, empty prefix, maxKeys cap).
- `artifact-get.test.ts`: 3 integration tests (200 + text body + bounded calls; "50+" suffix for truncated; 404 for empty prefix).

58 tests passing locally against the real `minio@8.0.7`.

jeffspahr (approved):
Re-reviewed aa890edd7. The previous blocker is resolved: `/artifacts/get?...&peek=N` now returns a bounded text summary instead of streaming the directory tarball, and the new tests cover the capped listing/no child getObject behavior.

Local verification with Node v24.14.0:
- `npm test -- minio-helper.test.ts integration-tests/artifact-get.test.ts` passed: 58 tests.
- `npm run build` passed.
- `npm run format:check` passed.
- Full `frontend/server` `npm test` passed: 253 passed, 2 todo.

Issue #11597, opened 2025-02-07
[backend] recurring runs fail in Kubeflow 1.10 (kfp 2.4.0)


orfeas-k:
### Environment

* How did you deploy Kubeflow Pipelines (KFP)?
I installed Kubeflow from manifests 1.10-rc.0 tag following https://github.com/kubeflow/manifests?tab=readme-ov-file#install-with-a-single-command but on Microk8s 1.31. FWIW, I hit this first in Charmed Kubeflow when trying upgrading manifests for KFP components.

*  KFP version: 
2.4.0
*  KFP SDK version: 
Not used

### Steps to reproduce
First,
```bash
sudo snap install microk8s --channel 1.31/stable --classic
sudo microk8s enable dns rbac hostpath-storage ingress metallb:"10.64.140.43-10.64.140.49,192.168.0.105-192.168.0.111"
git clone https://github.com/kubeflow/manifests
cd manifests
export KUBECONFIG=~/.kube/config
sudo snap install kustomize
sudo snap alias microk8s.kubectl kubectl
git checkout tags/v1.10.0-rc.0
while ! kustomize build example | kubectl apply --server-side --force-conflicts -f -; do echo "Retrying to apply resources"; sleep 20; done
```
Then,
1. Login to kubeflow following https://github.com/kubeflow/manifests?tab=readme-ov-file#connect-to-your-kubeflow-cluster
1. Pipelines > Experiments: create an experiment.
1. Pipelines > Recurring Runs: create a recurring run. FWIW, I used shared pipeline Data passing and left all other fields empty. I also set frequency to every 1 minute.
2. Pipelines > Runs: Look at recurring runs failing.

### Expected result

Runs succeed. In contrast, all recurring runs were failing:
![Image](https://github.com/user-attachments/assets/3aa3ac45-d465-4bbe-b39f-abf0b64423b7)

The only successful run is the one created from `Pipelines > Runs` page which seem to be spinning up runs without issues.

### Materials and Reference
Did quite some debugging. Neither logs from `api-server`  (from pod ml-pipeline-78bf858557-gzbbj) nor from `scheduledworkflow` deployment had something useful. What I found was the following events
```bash
[...]

PR #11578, opened 2025-01-31
fix(backend): Replaced hardcoded ServiceAccount with default config

hbelmiro:
**Description of your changes:**
Fixes: https://github.com/kubeflow/pipelines/issues/11574

**Checklist:**
- [x] You have [signed off your commits](https://www.kubeflow.org/docs/about/contributing/#sign-off-your-commits)
- [x] The title for your pull request (PR) should follow our title convention. [Learn more about the pull request title convention used in this repository](https://github.com/kubeflow/pipelines/blob/master/CONTRIBUTING.md#pull-request-title-convention).

PR #11481, opened 2024-12-20
chore(backend): Fixed ServiceAccount in job creation

hbelmiro:
When creating recurring runs, the ScheduledWorkflow CR is created with no ServiceAccount set.
This PR fixes that.

**Checklist:**
- [x] You have [signed off your commits](https://www.kubeflow.org/docs/about/contributing/#sign-off-your-commits)
- [x] The title for your pull request (PR) should follow our title convention. [Learn more about the pull request title convention used in this repository](https://github.com/kubeflow/pipelines/blob/master/CONTRIBUTING.md#pull-request-title-convention).

PR #13325, opened 2026-04-27
fix(backend): apply recurring run service account unconditionally

### LINKED ISSUE #13311: [api-server]scheduled workflow service account issue
### Environment

*  KFP version: 

tested on both 2.15.0 and 2.16.0(latest version) ,meet the same issue .
*  KFP SDK version: 

### Steps to reproduce
go to pipelines--recuring runs menu 
create a new recuring run and input a service account ***.
then get the scheduled workflow and check it's service account using `kubectl get scheduledworkflow from $namespace -oyaml |grep serviceaccount.`
whatever user input for the ksa , the scheduledworkflow always use the default ksa --default-editor .if I delete the defaut-editor configuration from https://github.com/kubeflow/pipelines/blob/f3d8e1fe7903c1dc0de956db70d0b09262cb4292/manifests/kustomize/base/installs/multi-user/api-service/params.env#L2 ,then it will always take pipeline-runner as the scheduledworkflow service account.

### Expected result
if the user input any service account on UI ,then the scheduled workflow should use it as service account instead of always using the current default service account default-editor.

### Materials and Reference

we can chang the logic to below for https://github.com/kubeflow/pipelines/blob/f3d8e1fe7903c1dc0de956db70d0b09262cb4292/backend/src/apiserver/template/v2_template.go#L149-L151
I tested the change is working.
```go
-       if executionSpec.ServiceAccount() == "" {
-               setDefaultServiceAccount(executionSpec, modelJob.ServiceAccount)
+       if modelJob.ServiceAccount != "" {
+               executionSpec.SetServiceAccount(modelJob.ServiceAccount)
+       } else if executionSpec.ServiceAccount() == "" {
+               setDefaultServiceAccount(executionSpec, "")
        }
``` 

---

neo99zhang:
## Summary

When creating a recurring run with a user-specified service account, the service account is silently dropped and the workflow always runs with the compiler default instead.

**Root cause:** In `ScheduledWorkflow()`, `setDefaultServiceAccount` was guarded by `if executionSpec.ServiceAccount() == ""`. Since the Argo compiler (via `argocompiler.Compile`) always sets a `ServiceAccountName` on the workflow spec (changed from hardcoded `"pipeline-runner"` to a configurable default in #11578), this condition is never true for V2 pipelines. As a result, `modelJob.ServiceAccount` is never applied.

**Fix:** Remove the guard and call `setDefaultServiceAccount` unconditionally. This matches what `RunWorkflow()` already does (line 388 in the same file), and `setDefaultServiceAccount` already handles both cases internally: it applies the user-specified SA when non-empty, and falls back to the configured default otherwise.

## Related issues

- Fixes #13311 — scheduled workflow service account ignored
- Regressed in #11481 — added the `if executionSpec.ServiceAccount() == ""` guard that is always false because the Argo compiler populates the field
- Related to #11597 — recurring runs failing because the compiler-injected default SA (`pipeline-runner`) didn't exist in the target namespace, while the user-configured SA was silently dropped
- Related to #11578 — replaced the hardcoded `pipeline-runner` with a configurable default in the Argo compiler, which made the guard condition in `ScheduledWorkflow` permanently false for any non-default configuration

## Test plan

- [ ] Existing unit tests pass (`go test ./backend/src/apiserver/template/...`)
- [ ] Create a recurring run with a custom service account; verify the resulting ScheduledWorkflow uses it
- [ ] Create a recurring run without specifying a service account; verify it falls back to the configured default
- [ ] Verify one-off runs are unaffected (the `RunWorkflow` path was already unconditional)

### DIFF
--- a/backend/src/apiserver/template/template_test.go
+++ b/backend/src/apiserver/template/template_test.go
@@ -31,6 +31,7 @@ import (
 	commonutil "github.com/kubeflow/pipelines/backend/src/common/util"
 	scheduledworkflow "github.com/kubeflow/pipelines/backend/src/crd/pkg/apis/scheduledworkflow/v1beta1"
 	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
 	"google.golang.org/grpc/codes"
 	"google.golang.org/protobuf/encoding/protojson"
 	structpb "google.golang.org/protobuf/types/known/structpb"
@@ -245,6 +246,40 @@ func TestScheduledWorkflow(t *testing.T) {
 	// It is also tested in compiler.
 	actualScheduledWorkflow.Spec.Workflow.Spec = ""
 	assert.Equal(t, &expectedScheduledWorkflow, actualScheduledWorkflow)
+}
+
+func TestScheduledWorkflow_CustomServiceAccount(t *testing.T) {
+	proxy.InitializeConfigWithEmptyForTests()
+
+	v2SpecHelloWorldYAML := loadYaml(t, "testdata/hello_world.yaml")
+	v2Template, _ := New([]byte(v2SpecHelloWorldYAML), TemplateOptions{CacheDisabled: true, DefaultWorkspace: defaultPVC})
+
+	modelJob := &model.Job{
+		K8SName:        "name1",
+		Enabled:        true,
+		MaxConcurrency: 1,
+		NoCatchup:      true,
+		ServiceAccount: "custom-sa",
+		Trigger: model.Trigger{
+			CronSchedule: model.CronSchedule{
+				CronScheduleStartTimeInSec: util.Int64Pointer(1),
+				CronScheduleEndTimeInSec:   util.Int64Pointer(10),
+				Cron:                       util.StringPointer("1 * * * *"),
+			},
+		},
+		PipelineSpec: model.PipelineSpec{
+			PipelineId:           "1",
+			PipelineName:         "pipeline name",
+			PipelineSpecManifest: model.LargeText(v2SpecHelloWorldYAML),
+			RuntimeConfig: model.RuntimeConfig{
+				Parameters: "{\"y\":\"world\"}",
+			},
+		},
+	}
+
+	actualScheduledWorkflow, err := v2Template.ScheduledWorkflow(modelJob)
+	require.Nil(t, err)
+	assert.Equal(t, "custom-sa", actualScheduledWorkflow.Spec.ServiceAccount)
 }
 
 func TestModelToCRDTrigger_Cron(t *testing.T) {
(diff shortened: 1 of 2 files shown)


hbelmiro (commented):
The fix is correct and aligns `ScheduledWorkflow()` with `RunWorkflow()`, which already calls `setDefaultServiceAccount` unconditionally.

However, the existing `TestScheduledWorkflow` only covers the case where no user SA is specified (expects the compiler default `"pipeline-runner"`). Could you add a test for the scenario this PR fixes — where `modelJob.ServiceAccount` is set to a custom value and the resulting `scheduledWorkflow.Spec.ServiceAccount` reflects it?

Without that test, someone could reintroduce the same guard and all tests would still pass.

neo99zhang:
@hbelmiro Thanks for the prompt review! 
Just added a unit test `TestScheduledWorkflow_CustomServiceAccount` which sets `modelJob.ServiceAccount` == "custom-sa"  and asserts the resulting `scheduledWorkflow.Spec.ServiceAccount`. This would fail if the guard were reintroduced. Please let me know how you feels about it.
BTW, I noticed this are two CIs failing, but it looks like they are unrelated with this PR

neo99zhang:
@hbelmiro Hi it looks like the CI failed again due to some github action infra issue, could you help with the retest again? Thanks!

backend/src/apiserver/template/template_test.go
```diff
@@ -247,6 +247,40 @@ func TestScheduledWorkflow(t *testing.T) {
...
+			PipelineName:         "pipeline name",
+			PipelineSpecManifest: model.LargeText(v2SpecHelloWorldYAML),
+			RuntimeConfig: model.RuntimeConfig{
+				Parameters: "{\"y\":\"world\"}",
+			},
+		},
+	}
+
+	actualScheduledWorkflow, err := v2Template.ScheduledWorkflow(modelJob)
+	assert.Nil(t, err)
```
hbelmiro: Nit

```suggestion
	require.Nil(t, err)
```

hbelmiro (commented):
The CI is failing probably because the images expired.
Closing and reopening the PR will restart the CI and the images will be rebuilt.
Since you'll have to do that, can you please fix a nit?

backend/src/apiserver/template/template_test.go
neo99zhang: That makes sense. Thanks for the review, just fixed it

neo99zhang: Hi @hbelmiro I addressed the nit and closed/reopened the PR, but the CI failure is still there. Could you /retest when you get a chance? Thanks!

hbelmiro (approved):
/lgtm
/approve
/ok-to-test

hbelmiro:
Closing and reopening to rerun the tests.

/close

hbelmiro:
@neo99zhang there's a [bug](https://github.com/wechuli/allcheckspassed/issues/104) in the action used by the CI Check workflow that's preventing this PR from being merged. Can you please rebase to see if it passes? All tests are passing now.

neo99zhang:
@hbelmiro Hi sorry for the delay! Just rebased and the LGTM label was removed, would appreciate if you could re-approve once the CI pass

neo99zhang:
@hbelmiro It looks like there is still one CI failing after rebasing, wondering if there is anything else we should do given the CI infrastructure know issue.

hbelmiro (commented):
/lgtm

PR #13345, opened 2026-05-04
fix(backend): prevent premature workflow GC due to DB write race condition (#13342)

### LINKED ISSUE #13342: [backend] Race condition in resource_manager.go causes pipeline run workflows to be garbage collected during high kfp-db write latency
### Environment

KFP was installed as part of the Canonical Kubeflow Distribution, with the command:
```
juju deploy kubeflow --channel=1.8/stable --trust
```
*  KFP version: 
2.0.3
*  KFP SDK version: 
2.4.0

### Steps to reproduce

1. Throttle disk I/O on the storage backing kfp-db to approximately 5 IOPS. Either programmatically or via misconfiguration, whoops.
2. Submit one or more pipeline runs via the KFP UI or SDK while the database is under this degraded I/O constraint.
3. Observe the Kubernetes resources, created to back the run, namely Argo Workflows.

### Expected result

The pipeline run should start and complete successfully. The garbage collector in resource_manager.go should not treat a workflow as orphaned unless it has confirmed that no corresponding database entry exists, accounting for in-progress writes.

### Actual result

The Argo Workflow backing the pipeline run is deleted within a few seconds of creation. This is caused by the garbage collection logic introduced in #6189, which deletes workflows that have no corresponding entry in the KFP database. Because the `kfp-db` write has not completed before the GC check runs, the workflow is incorrectly identified as orphaned and deleted. The pipeline run never executes.
The race condition exists in `backend/src/apiserver/resource/resource_manager.go`: the workflow is created and the GC check can fire before the database write for that run has committed, particularly when database write latency is high.

### Materials and Reference

This workflow deletion will happen to any Kubeflow pipeline run that is created while the `kfp-db` is experiencing heavily degraded performance. I suspect this has not been reported before due to it not being a common occurrence that the database will have such poor performance. In this case, a combination of misconfigurations, low disk bandwidth, and high concurrent load caused the storage backing the `kfp-db` to average around 5 IOPS, causing database writes to take extremely long. This triggers the garbage collection logic introduced in #6189, which deletes Argo Workflows that have no corresponding database entry, incorrectly treating in-flight writes as missing entries. There exists a race condition in `backend/src/apiserver/resource/resource_manager.go` where Argo Workflows for Kubeflow pipeline runs will be garbage collected if the database write takes too long to complete. The fix should ensure the GC cannot delete a workflow within a grace period of its creation, or that the GC confirms no write is in progress before acting.

---

shreeharshshinde:
### Summary

This PR fixes a race condition in `ReportWorkflowResource()` where newly created workflows could be incorrectly garbage collected if the corresponding run record had not yet been persisted to the database.

---

### Problem

For one-time runs, the API server creates the workflow CR and then writes the run record to the database. Under high DB latency, the persistence agent may observe the workflow before the DB write completes.

When `GetRun(runId)` returns `NotFound`, the current GC logic assumes the workflow is orphaned and deletes it immediately. This results in valid pipeline runs being terminated before execution.

---

### Solution

Introduce a configurable grace period before garbage collecting workflows whose run record is not found in the DB.

* Compute workflow age using Kubernetes `CreationTimestamp`
* If workflow age < grace period:

  * Skip GC
  * Return `codes.Unavailable` (retryable)
* Otherwise:

  * Proceed with existing GC logic

---

### Key Changes

* Add `WORKFLOW_GC_GRACE_PERIOD_SECONDS` config (default: 120s)
* Update `ReportWorkflowResource()` to:

  * Check workflow age before GC
  * Return retryable error for young workflows
* Preserve existing GC behavior for workflows beyond grace period
* Add deterministic tests using fake time:

  * Within grace period → no deletion, retry
  * Beyond grace period → GC proceeds

---

### Why This Approach

* Minimal and safe change (no schema or API changes)
* Handles eventual consistency between Kubernetes and DB
* Avoids premature deletion without disabling GC
* Fully configurable for different cluster conditions

---

### Trade-offs

* Real orphan workflows may be GC’d with a delay (up to grace period)
* Relies on persistence agent retry behavior (already in place)

---

### Testing

* Added unit tests for both within and beyond grace period scenarios
* Existing tests updated to remain deterministic
* All backend tests pass

---

### Related Issue

Fixes #13342

### DIFF
--- a/backend/src/apiserver/common/config.go
+++ b/backend/src/apiserver/common/config.go
@@ -57,6 +57,7 @@ const (
 	PluginMaxPayloadBytes                   string = "PLUGIN_MAX_PAYLOAD_BYTES"
 	PluginMaxTotalPayloadBytes              string = "PLUGIN_MAX_TOTAL_PAYLOAD_BYTES"
 	PluginMaxNestingDepth                   string = "PLUGIN_MAX_NESTING_DEPTH"
+	WorkflowGCGracePeriodSeconds            string = "WORKFLOW_GC_GRACE_PERIOD_SECONDS"
 )
 
 type PluginLimitsConfig struct {
@@ -64,6 +65,15 @@ type PluginLimitsConfig struct {
 	MaxPayloadBytes      int
 	MaxTotalPayloadBytes int
 	MaxNestingDepth      int
+}
+
+// GetWorkflowGCGracePeriodSeconds returns the grace period in seconds before
+// a workflow without a corresponding DB entry is eligible for garbage collection.
+// This prevents race conditions where the persistence agent reports a workflow
+// before the API server has finished writing the run record to the database.
+// See https://github.com/kubeflow/pipelines/issues/13342.
+func GetWorkflowGCGracePeriodSeconds() int {
+	return GetIntConfigWithDefault(WorkflowGCGracePeriodSeconds, 120)
 }
 
 func IsPipelineVersionUpdatedByDefault() bool {
--- a/backend/src/apiserver/common/config.go
+++ b/backend/src/apiserver/common/config.go
@@ -71,7 +71,6 @@ type PluginLimitsConfig struct {
 // a workflow without a corresponding DB entry is eligible for garbage collection.
 // This prevents race conditions where the persistence agent reports a workflow
 // before the API server has finished writing the run record to the database.
-// See https://github.com/kubeflow/pipelines/issues/13342.
 func GetWorkflowGCGracePeriodSeconds() int {
 	return GetIntConfigWithDefault(WorkflowGCGracePeriodSeconds, 120)
 }
--- a/backend/src/apiserver/resource/resource_manager.go
+++ b/backend/src/apiserver/resource/resource_manager.go
@@ -945,6 +945,14 @@ func (r *ResourceManager) UnarchiveRun(runId string) error {
 	return nil
 }
 
+// newStandardBackoffPolicy returns a configured backoff policy for retrying operations.
+func newStandardBackoffPolicy() backoff.BackOff {
+	exponentialBackoff := backoff.NewExponentialBackOff()
+	exponentialBackoff.InitialInterval = 100 * time.Millisecond
+	exponentialBackoff.MaxInterval = 5 * time.Second
+	return backoff.WithMaxRetries(exponentialBackoff, 10)
+}
+
 // Deletes a run entry with a given id.
 func (r *ResourceManager) DeleteRun(ctx context.Context, runId string) error {
 	run, err := r.GetRun(runId)
@@ -1050,11 +1058,7 @@ func TerminateWorkflow(ctx context.Context, wfClient util.ExecutionInterface, na
 		_, err = wfClient.Patch(ctx, name, types.MergePatchType, patch, v1.PatchOptions{})
 		return util.Wrapf(err, "Failed to terminate workflow %s due to patching error", name)
 	}
-	exponentialBackoff := backoff.NewExponentialBackOff()
-	exponentialBackoff.InitialInterval = 100 * time.Millisecond
-	exponentialBackoff.MaxInterval = 5 * time.Second
-	backoffPolicy := backoff.WithMaxRetries(exponentialBackoff, 10)
-	err = backoff.Retry(operation, backoffPolicy)
+	err = backoff.Retry(operation, newStandardBackoffPolicy())
 	if err != nil {
 		return util.Wrapf(err, "Failed to terminate workflow %s due to patching error after multiple retries", name)
 	}
@@ -1614,11 +1618,7 @@ func (r *ResourceManager) ReportWorkflowResource(ctx context.Context, execSpec u
 				}
 				return nil
 			}
-			exponentialBackoff := backoff.NewExponentialBackOff()
-			exponentialBackoff.InitialInterval = 100 * time.Millisecond
-			exponentialBackoff.MaxInterval = 5 * time.Second
-			backoffPolicy := backoff.WithMaxRetries(exponentialBackoff, 10)
-			if err := backoff.Retry(deleteOperation, backoffPolicy); err != nil {
+			if err := backoff.Retry(deleteOperation, newStandardBackoffPolicy()); err != nil {
 				return nil, util.NewInternalServerError(err, "Failed to delete the obsolete workflow for run %s after multiple retries", runId)
 			}
 
@@ -1771,11 +1771,7 @@ func addWorkflowLabel(ctx context.Context, wfClient util.ExecutionInterface, nam
 		_, err = wfClient.Patch(ctx, name, types.MergePatchType, patch, v1.PatchOptions{})
 		return err
 	}
-	exponentialBackoff := backoff.NewExponentialBackOff()
-	exponentialBackoff.InitialInterval = 100 * time.Millisecond
-	exponentialBackoff.MaxInterval = 5 * time.Second
-	backoffPolicy := backoff.WithMaxRetries(exponentialBackoff, 10)
-	err = backoff.Retry(operation, backoffPolicy)
+	err = backoff.Retry(operation, newStandardBackoffPolicy())
 	return err
 }
 
--- a/backend/src/apiserver/resource/resource_manager.go
+++ b/backend/src/apiserver/resource/resource_manager.go
@@ -1585,7 +1585,7 @@ func (r *ResourceManager) ReportWorkflowResource(ctx context.Context, execSpec u
 			// See https://github.com/kubeflow/pipelines/issues/13342.
 			gracePeriodSeconds := common.GetWorkflowGCGracePeriodSeconds()
 			gracePeriod := time.Duration(gracePeriodSeconds) * time.Second
-			workflowAge := r.time.Now().Sub(objMeta.CreationTimestamp.Time)
+			workflowAge := time.Since(objMeta.CreationTimestamp.Time)
 			if workflowAge < gracePeriod {
 				glog.Warningf(
 					"Workflow name=%q namespace=%q runId=%q not found in run store, "+
--- a/backend/src/apiserver/resource/resource_manager.go
+++ b/backend/src/apiserver/resource/resource_manager.go
@@ -1577,7 +1577,28 @@ func (r *ResourceManager) ReportWorkflowResource(ctx context.Context, execSpec u
 				return nil, util.Wrap(updateError, "Failed to update the run")
 			}
 			// Handle run not found in run store error.
-			// To avoid letting the workflow leak for ever, we need to GC it when its record does not exist in KFP DB.
+			// Before GC, apply a grace period to avoid deleting workflows whose
+			// DB writes are still in-flight.
+			// See https://github.com/kubeflow/pipelines/issues/13342.
+			gracePeriodSeconds := common.GetWorkflowGCGracePeriodSeconds()
+			gracePeriod := time.Duration(gracePeriodSeconds) * time.Second
+			workflowAge := r.time.Now().Sub(objMeta.CreationTimestamp.Time)
+			if workflowAge < gracePeriod {
+				glog.Warningf(
+					"Workflow name=%q namespace=%q runId=%q not found in run store, "+
+						"but workflow is only %v old (grace period: %v). "+
+						"Skipping GC to allow in-flight DB write to complete. "+
+						"See https://github.com/kubeflow/pipelines/issues/13342",
+					execSpec.ExecutionName(), execSpec.ExecutionNamespace(), runId,
+					workflowAge.Round(time.Second), gracePeriod)
+				return nil, util.NewUnavailableServerError(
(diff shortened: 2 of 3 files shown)


shreeharshshinde:
## 🤖 InspectAI Code Review

🟡 **Severity**: `minor` (90% confidence)

### Review Comment
I don't think it's a good idea.

<details>
<summary>📚 3 similar historical review(s) used as context</summary>

1. _it's not immediately obvious what this controls... godoc? same for sharedSecretsPerNamespace..._
2. _should we populate all the `*PerPod` dimensions here? could extract the pod-specific bits of `genera..._
3. _We should also add service accounts to this at some point...._
</details>

---
_Generated by [InspectAI](https://github.com/shreeharsh157/InspectAI) • Processing time: 4678ms_

backend/src/apiserver/resource/resource_manager.go
```diff
@@ -1482,7 +1482,28 @@ func (r *ResourceManager) ReportWorkflowResource(ctx context.Context, execSpec u
 				return nil, util.Wrap(updateError, "Failed to update the run")
 			}
 			// Handle run not found in run store error.
-			// To avoid letting the workflow leak for ever, we need to GC it when its record does not exist in KFP DB.
+			// Before GC, apply a grace period to avoid deleting workflows whose
+			// DB writes are still in-flight.
+			// See https://github.com/kubeflow/pipelines/issues/13342.
+			gracePeriodSeconds := common.GetWorkflowGCGracePeriodSeconds()
+			gracePeriod := time.Duration(gracePeriodSeconds) * time.Second
+			workflowAge := r.time.Now().Sub(objMeta.CreationTimestamp.Time)
```
Amr-Shams, line 1591: you can use time.since much clenaer and even better for nano seconds formatting

Amr-Shams (commented):
i think you might consider using retery with cap instead of timing cause it so fragile, and backoff with exponintial would be something worth trying

shreeharshshinde:
> i think you might consider using retery with cap instead of timing cause it so fragile, and backoff with exponintial would be something worth trying

agreed @Amr-Shams , retry with capped exponential backoff feels more robust here than depending on a fixed timing window, so i updated the implementation in that direction

backend/src/apiserver/resource/resource_manager.go
shreeharshshinde: we can use `time.Since(objMeta.CreationTimestamp.Time)` which uses the monotonic clock
  directly for more accurate nanosecond precision as you said. Updated tests to use real relative
  timestamps (`time.Now().Add(...)`) instead of the fake clock accordingly

backend/src/apiserver/resource/resource_manager.go
```diff
@@ -1482,7 +1485,28 @@ func (r *ResourceManager) ReportWorkflowResource(ctx context.Context, execSpec u
 				return nil, util.Wrap(updateError, "Failed to update the run")
 			}
 			// Handle run not found in run store error.
-			// To avoid letting the workflow leak for ever, we need to GC it when its record does not exist in KFP DB.
+			// Before GC, apply a grace period to avoid deleting workflows whose
```
Amr-Shams, line 1587: i mean here we could have made use of hte backoff and wrapp the operation in anyomus func and then try with the backoff policy

shreeharshshinde: Wrapped it, is it okay now ? @Amr-Shams

shreeharshshinde:
Any updates @Amr-Shams ?

Amr-Shams:
> Any updates @Amr-Shams ?

nothing more from my end, looks lgtm

shreeharshshinde:
@Amr-Shams can you tag ok-to-test and approve further to merge ?

shreeharshshinde:
@Amr-Shams @hbelmiro solved the merge conflict, good to merge ?

Amr-Shams:
LGTM from my side

backend/src/apiserver/resource/resource_manager.go
```diff
@@ -1482,17 +1485,46 @@ func (r *ResourceManager) ReportWorkflowResource(ctx context.Context, execSpec u
...
+			// DB writes are still in-flight.
+			// See https://github.com/kubeflow/pipelines/issues/13342.
+			gracePeriodSeconds := common.GetWorkflowGCGracePeriodSeconds()
+			gracePeriod := time.Duration(gracePeriodSeconds) * time.Second
+			workflowAge := time.Since(objMeta.CreationTimestamp.Time)
```
hbelmiro: `time.Since()` bypasses the injectable `r.time` clock that `ResourceManager` uses everywhere else. The struct has a `time util.TimeInterface` field (line 147) for fake clocks in tests, and other code in the same struct uses it (e.g., line 662: `r.time.Now().Unix()`). This new grace period check goes directly to the wall clock instead.

I know this was suggested in the earlier review thread — `time.Since` is more idiomatic Go in general, so the suggestion makes sense. However, `ResourceManager` has its own clock abstraction, so using `r.time` would keep the code consistent with the rest of the struct and allow future tests to control time deterministically. Not a practical correctness issue — the tests work because their values (10s and 10min) are far from the 120s boundary.

backend/src/apiserver/resource/resource_manager_test.go
```diff
@@ -3418,6 +3420,65 @@ func TestReportWorkflowResource_RunNotFound(t *testing.T) {
...
+		"Expected Unavailable error for young workflow within grace period, got: %v", err)
+	assert.Contains(t, err.Error(), "GC grace period")
+
+	// Verify the workflow was NOT deleted by checking it's still accessible.
+	wf, getErr := store.ExecClient().Execution("kubeflow").Get(ctx, "young-workflow", v1.GetOptions{})
+	assert.Nil(t, getErr, "Workflow should still exist after grace period skip")
+	assert.NotNil(t, wf)
+}
+
+func TestReportWorkflowResource_RunNotFound_BeyondGracePeriod(t *testing.T) {
```
hbelmiro: `TestReportWorkflowResource_RunNotFound_BeyondGracePeriod` is functionally identical to the updated `TestReportWorkflowResource_RunNotFound` (line 3402). Both use a `CreationTimestamp` of `-10 * time.Minute`, both assert `codes.NotFound`, and both check for `"Run run-id-not-exist not found"`. The only difference is the workflow name string ("obsolete" vs "old-orphan-workflow").

**Suggested fix:** Remove `TestReportWorkflowResource_RunNotFound_BeyondGracePeriod` since the existing `TestReportWorkflowResource_RunNotFound` already covers this path after the CreationTimestamp update.

backend/src/apiserver/resource/resource_manager.go
```diff
@@ -984,7 +984,10 @@ func TerminateWorkflow(ctx context.Context, wfClient util.ExecutionInterface, na
 		_, err = wfClient.Patch(ctx, name, types.MergePatchType, patch, v1.PatchOptions{})
 		return util.Wrapf(err, "Failed to terminate workflow %s due to patching error", name)
 	}
-	backoffPolicy := backoff.WithMaxRetries(backoff.NewConstantBackOff(100), 10)
+	exponentialBackoff := backoff.NewExponentialBackOff()
```
hbelmiro: The exponential backoff configuration block is duplicated three times with identical parameters (lines 987-990, 1522-1525, 1666-1669):

```go
exponentialBackoff := backoff.NewExponentialBackOff()
exponentialBackoff.InitialInterval = 100 * time.Millisecond
exponentialBackoff.MaxInterval = 5 * time.Second
backoffPolicy := backoff.WithMaxRetries(exponentialBackoff, 10)
```

Per the repo's code reuse policy, this should be extracted to a helper function.

**Suggested fix:** Create a helper that returns the configured backoff policy:

```go
func newStandardBackoffPolicy() backoff.BackOff {
	exponentialBackoff := backoff.NewExponentialBackOff()
	exponentialBackoff.InitialInterval = 100 * time.Millisecond
	exponentialBackoff.MaxInterval = 5 * time.Second
	return backoff.WithMaxRetries(exponentialBackoff, 10)
}
```

backend/src/apiserver/common/config.go
```diff
@@ -46,12 +46,22 @@ const (
...
 	PipelineURLAllowHTTP                    string = "PIPELINE_URL_ALLOW_HTTP"
 	PipelineURLTimeout                      string = "PIPELINE_URL_TIMEOUT"
 	PipelineURLValidationEnabled            string = "PIPELINE_URL_VALIDATION_ENABLED"
 )
 
+// GetWorkflowGCGracePeriodSeconds returns the grace period in seconds before
+// a workflow without a corresponding DB entry is eligible for garbage collection.
+// This prevents race conditions where the persistence agent reports a workflow
+// before the API server has finished writing the run record to the database.
+// See https://github.com/kubeflow/pipelines/issues/13342.
```
hbelmiro: This is only noise.
```suggestion
```

backend/src/apiserver/resource/resource_manager.go
```diff
@@ -1482,17 +1485,46 @@ func (r *ResourceManager) ReportWorkflowResource(ctx context.Context, execSpec u
...
+			// See https://github.com/kubeflow/pipelines/issues/13342.
+			gracePeriodSeconds := common.GetWorkflowGCGracePeriodSeconds()
+			gracePeriod := time.Duration(gracePeriodSeconds) * time.Second
+			workflowAge := time.Since(objMeta.CreationTimestamp.Time)
+			if workflowAge < gracePeriod {
+				glog.Warningf(
+					"Workflow name=%q namespace=%q runId=%q not found in run store, "+
+						"but workflow is only %v old (grace period: %v). "+
+						"Skipping GC to allow in-flight DB write to complete. "+
+						"See https://github.com/kubeflow/pipelines/issues/13342",
```
hbelmiro: Please remove the reference to the issue.

shreeharshshinde:
@hbelmiro, some E2E tests are failing, but they don't seem to be related to my changes. Could you please let me know if these are known issues or if I should rebase and rerun the tests? I'd like to get this PR merged.

hbelmiro:
@shreeharshshinde please rebase. We've fixed a few issues in the CI.

shreeharshshinde:
@hbelmiro successfully rebased, looking forward to get merged ...

backend/src/apiserver/resource/resource_manager.go
```diff
@@ -1577,17 +1584,42 @@ func (r *ResourceManager) ReportWorkflowResource(ctx context.Context, execSpec u
 				return nil, util.Wrap(updateError, "Failed to update the run")
 			}
 			// Handle run not found in run store error.
-			// To avoid letting the workflow leak for ever, we need to GC it when its record does not exist in KFP DB.
+			// Before GC, apply a grace period to avoid deleting workflows whose
+			// DB writes are still in-flight.
+			// See https://github.com/kubeflow/pipelines/issues/13342.
```
hbelmiro: Please remove this link.

backend/src/apiserver/resource/resource_manager.go
hbelmiro: I still see the link.

backend/src/apiserver/resource/resource_manager.go
```diff
@@ -1577,17 +1584,42 @@ func (r *ResourceManager) ReportWorkflowResource(ctx context.Context, execSpec u
...
+						execSpec.ExecutionName(), workflowAge.Round(time.Second), gracePeriod),
+					"Skipping GC for workflow %s - will retry",
+					execSpec.ExecutionName())
+			}
+			// Workflow is beyond the grace period. To avoid letting the workflow
+			// leak forever, GC it since its record does not exist in KFP DB.
 			glog.Errorf("Cannot find reported workflow name=%q namespace=%q runId=%q in run store. "+
 				"Deleting the workflow to avoid resource leaking. "+
 				"This can be caused by installing two KFP instances that try to manage the same workflows "+
 				"or an unknown bug. If you encounter this, recommend reporting more details in https://github.com/kubeflow/pipelines/issues/6189",
```
hbelmiro, line 1610: Please remove this.

backend/src/apiserver/resource/resource_manager_test.go
```diff
@@ -3774,6 +3777,40 @@ func TestReportWorkflowResource_RunNotFound(t *testing.T) {
 	assert.Contains(t, err.Error(), "Run run-id-not-exist not found")
 }
 
+func TestReportWorkflowResource_RunNotFound_WithinGracePeriod(t *testing.T) {
+	// When a workflow is young (within the GC grace period) and its run is not
+	// found in the DB, it should NOT be deleted. Instead, a retryable
+	// UNAVAILABLE error should be returned so the persistence agent retries.
+	// This prevents the race condition described in issue #13342.
```
hbelmiro: Please don't reference issues.

```suggestion
```

shreeharshshinde:
@hbelmiro removed the issue based comments

hbelmiro (approved):
/lgtm
/approve

Issue #13349, opened 2026-05-05
frontend: clean up React act warnings in Compare and Router tests


jeffspahr:
## Summary

The frontend test suite currently emits a large number of React `act(...)` warnings when running Compare and Router tests. This came up while reviewing #13343, but it is pre-existing test debt rather than a regression from that PR.

These warnings do not currently fail the tests, but they make the test output noisy and indicate that some tests are not fully synchronized with React's update cycle. Cleaning this up should make failures easier to spot, reduce async leakage between tests, and lower risk during future React/MUI/testing-library upgrades.

## Evidence / Repro

Using Node `v24.14.0` from `frontend/.nvmrc`:

```bash
cd frontend
npm run test:ui -- CompareV1.test.tsx CompareV2.test.tsx Router.test.tsx
```

During review of #13343, I compared the warning-heavy subset on both base and PR head:

- Base commit `53e083561`: 6 files passed / 69 tests passed, with `1478` `act(...)` warning instances
- PR head `1f624077e`: 6 files passed / 69 tests passed, with `1455` `act(...)` warning instances
- The same 31 warning contexts appeared on both base and PR head; there were no PR-only warning contexts

Most frequent warning sources observed on PR head:

- `ForwardRef(Tooltip)`: 348
- `ForwardRef(ButtonBase)`: 222
- `TransitionGroup`: 180
- `TestCompare`: 136
- `ForwardRef(FormControl)`: 82
- `CustomTable`: 57
- `TensorboardViewer`: 52
- `RunList`: 38
- `CompareV2`: 22
- `CompareV1`: 16
- `Router`: 1

## Why this matters

React `act(...)` warnings mean state updates are happening after the test's current act scope has ended. That can be benign, but it also means tests may assert before the UI has fully settled. The risks are:

- important assertions can run too early and miss user-visible states
- async updates can leak into later tests
- test output is noisy enough to hide new warnings
- future React, MUI, Testing Library, or Vitest upgrades may turn these into harder failures
- developers may normalize warning-heavy output and miss real regressions

This does not mean the Compare or Router tests are currently invalid. It means the test harnesses and async waits should be tightened so the tests prove the intended UI behavior without relying on incidental timing.

### Slice 1: Router test warning

Small, low-risk starter task.

- File: `frontend/src/components/Router.test.tsx`
- Symptom: `An update to Router inside a test was not wrapped in act(...)`
[...]

PR #13358, opened 2026-05-07
test(frontend): wrap router navigation in act

jeffspahr:
## Summary

- Wrap the imperative `history.push('/pear')` in `Router.test.tsx` with Testing Library `act`
- Remove the isolated Router `act(...)` warning tracked as the first slice of #13349
- Keep the change scoped to the test harness; no production behavior changes

## Why this helps

The test manually changes routes with `history.push('/pear')`. That call updates React Router state and causes `Router` to re-render, but because the update is triggered outside Testing Library's normal render/user-event helpers, React reports:

```text
An update to Router inside a test was not wrapped in act(...)
```

Wrapping the navigation in `act` tells React that this operation is expected to cause updates and those updates should be flushed before the test continues. This makes the test harness more deterministic, removes a real synchronization warning, and reduces suite noise so future warnings are easier to spot.

This does not change application behavior. It only makes the test explicitly wait for the React update caused by manual history navigation.

## Verification

- `npm run test:ui -- Router.test.tsx`
  - before: passed with 1 `act(...)` warning
  - after: passed with 0 `act(...)` warnings
- `npm run lint:ui`
- `npm run typecheck`

Part of #13349.

/area frontend
/kind cleanup

### DIFF
--- a/frontend/src/components/Router.test.tsx
+++ b/frontend/src/components/Router.test.tsx
@@ -14,7 +14,7 @@
  * limitations under the License.
  */
 
-import { render, screen, waitFor } from '@testing-library/react';
+import { act, render, screen, waitFor } from '@testing-library/react';
 import { Router as ReactRouter } from 'react-router';
 import { MemoryRouter } from 'react-router-dom';
 import { createMemoryHistory } from 'history';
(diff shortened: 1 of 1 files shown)


manaswinidas (approved):
/lgtm
/approve

PR #13364, opened 2026-05-08
test(frontend): flush NewExperiment setup in act

jeffspahr:
## Summary

Replace the remaining raw promise flushes in `NewExperiment.test.tsx` with `flushPromisesInAct()`.

This PR is preventive helper debt cleanup, not a fix for an active warning in this file. The focused test already passed with zero `act(...)` warning lines before this change. The value here is that the existing flushes sit after render or user actions that can resolve promises and update React state, so using the repo's act-aware helper better matches the intent of those waits and keeps future async React updates from escaping the test boundary.

Part of #13349.

## Verification

- `npm run test:ui -- src/pages/NewExperiment.test.tsx` before: 15 tests passed with zero `act(...)` warning lines
- `npm run test:ui -- src/pages/NewExperiment.test.tsx` after: 15 tests passed with zero `act(...)` warning lines
- `npm run lint:ui`
- `npm run typecheck`
- `git diff --check`

### DIFF
--- a/frontend/src/pages/NewExperiment.test.tsx
+++ b/frontend/src/pages/NewExperiment.test.tsx
@@ -19,7 +19,7 @@ import { fireEvent, render, screen, waitFor } from '@testing-library/react';
 import userEvent from '@testing-library/user-event';
 import { vi } from 'vitest';
 import { NewExperiment } from './NewExperiment';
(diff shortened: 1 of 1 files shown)


manaswinidas (approved):
/lgtm
/approve

PR #13365, opened 2026-05-08
test(frontend): flush Tensorboard updates in act

jeffspahr:
## Summary

Wrap Tensorboard viewer test promise drains and async click follow-up work in act-aware helpers.

Most tests only needed the existing `flushPromisesInAct()` helper after render or API-triggered updates. The polling helper also invokes an interval callback, so this PR wraps the interval callback and its promise drain in the same `act()` boundary. Three start-button actions now use `invokeAndFlush()` because they synchronously click and then schedule async React updates.

Part of #13349.

## Verification

- `npm run test:ui -- src/components/viewers/Tensorboard.test.tsx` before: 15 tests passed with 121 `act(...)` warning lines
- `npm run test:ui -- src/components/viewers/Tensorboard.test.tsx` after: 15 tests passed with zero `act(...)` warning lines
- `npm run lint:ui`
- `npm run typecheck`
- `git diff --check`


manaswinidas (approved):
/lgtm
/approve

Issue #13314, opened 2026-04-24
[backend] S3 operations fail with non-AWS object stores after AWS SDK v2 checksum defaults change


hbelmiro:
AWS SDK for Go v2 changed the default checksum behavior (see [aws/aws-sdk-go-v2#2960](https://github.com/aws/aws-sdk-go-v2/issues/2960)):

- **Request checksum calculation** now defaults to computing checksums on all S3 PUT requests.
- **Response checksum validation** now defaults to validating checksums on all S3 GET responses.

Non-AWS S3-compatible stores (MinIO, Ceph, etc.) may not support these checksums or may return malformed/missing checksum headers, causing both uploads and downloads to fail.

The standard fix is to set both options to `WhenRequired`:

```go
config.WithRequestChecksumCalculation(aws.RequestChecksumCalculationWhenRequired)
config.WithResponseChecksumValidation(aws.ResponseChecksumValidationWhenRequired)
```

Neither option is currently set in KFP.

---

Impacted by this bug? Give it a 👍.

PR #13293, opened 2026-04-21
KEP-13293: Gatekeep CI workflows for triaged issues

alyssacgoins:
**Description of your changes:**
This KEP proposes a two-part approach to managing the recent influx of drive-by AI-generated PRs in the KFP project, and the strain they've placed on the KFP community. 

**Open questions:**
- What level of Copilot AI triage do we want to utilize in this design? Do we want to use this tool to completely mark issues `ready`, or should this tool only assist human reviewers?

**Checklist:**
- [X] You have [signed off your commits](https://www.kubeflow.org/docs/about/contributing/#sign-off-your-commits)
- [X] The title for your pull request (PR) should follow our title convention. [Learn more about the pull request title convention used in this repository](https://github.com/kubeflow/pipelines/blob/master/CONTRIBUTING.md#pull-request-title-convention).

Issue #13180, opened 2026-03-31
fix: E2E test flakiness on K8s v1.34.0 — root cause analysis


droctothorpe:
## Summary

Intermittent E2E test failures have been observed exclusively on the K8s `v1.34.0` CI matrix leg since approximately 2026-03-25. This issue documents the root cause analysis and proposes fixes.

## Affected tests (primary pattern)

- `test_data/sdk_compiled_pipelines/valid/critical/nested_pipeline_opt_input_child_level_compiled.yaml`
- `test_data/sdk_compiled_pipelines/valid/critical/modelcar.yaml`

Both files have `FlakeAttempts(2)` in the E2E test suite, indicating known pre-existing flakiness.

## Affected tests (secondary pattern — test code bug)

- Any test that calls `CapturePodLogsForUnsuccessfulTasks` when a pod's log stream is unavailable will **panic** due to a nil pointer dereference in `ReadPodLogs` ([kubernetes_utils.go:78–88](https://github.com/kubeflow/pipelines/blob/master/backend/test/testutil/kubernetes_utils.go#L78-L88)). This is fixed in #13179.

## Timeline

| Date (UTC) | Event |
|---|---|
| 2026-03-24 19:15 | Last fully-passing master run (commit `28fe2830`, run #7448) |
| 2026-03-24 21:38 | Commit `5f4501ef` merged (3 frontend-only PRs: #13101, #13115, #13100) |
| 2026-03-25 01:38 | **First failure** on master (run #7456) — same K8s `v1.34.0` jobs fail |
| 2026-03-25 10:20 | PR #13124 merged — adds `pipeline_with_volume_long_name.yaml` to `critical/` |
| 2026-03-25+ | Failures continue intermittently |

Key observation: **the first failure (run #7456) contains only frontend-only code changes relative to the last success**. No backend, compiler, driver, launcher, or test data was changed. This rules out any specific PR as the direct cause of the regression.

### Primary: `kindest/node:v1.34.0` is a mutable floating tag

The [create-cluster action](https://github.com/kubeflow/pipelines/blob/master/.github/actions/create-cluster/action.yml) pulls `kindest/node:v1.34.0` (a mutable tag, no SHA digest) and caches it with key `kind-node-${{ runner.os }}-${{ inputs.k8s_version }}` — **no image digest in the cache key**.

[...]

PR #13372, opened 2026-05-08
automation: add AI-powered issue quality analysis workflow

modichika:
An AI-Powered issue analyzer that analyzes GitHub issues as soon as it opens. It uses 'gpt-4o-mini' via GitHub Models which is free, fast, and provides clear reasoning devided under: scope, context and guidance, complexity, overall issue quality verdict.

**Addressing the problem.**
Repositories that are high-usage to the tech community, its maintainers often are under the gun to analyse issues. That's a pain point I felt while finding the right issue that i need to work on for kubeflow/pipelines. 

**Why this is needed?**
Analysing issues is one kind of job in itself when maintainers are busy and too under the gun.

**The solutions:**
- A GitHub Action that triggers immediately upon issue creation.
- Wrote a ai_analyzer.yml file in .github/workflows/ai_analyzer.yml, utilize the built-in GITHUB_TOKEN for security and simplicity.
- Included a script folder suggested by @alyssacgoins so that the main .yml file remains clean and we have a strict prompt for llm.
- I included the comment one where ai leaves a comment after labelling of why it selected those labels, this provides transparency and allows maintainers to verify the AI's logic

**Checklist:**
- [x] You have [signed off your commits]

### DIFF
--- a/.github/scripts/ai_triage.sh
+++ b/.github/scripts/ai_triage.sh
@@ -1,0 +1,69 @@
+set -e
+gh extension install github/gh-models --force || true
+
+echo "Using model: $TARGET_MODEL"
+
+read -r -d '' SYSTEM_INSTRUCTIONS << 'EOF'  || true
+You are an expert open-source maintainer for Kubeflow Pipelines.
+
+Analyze the quality of the incoming issue based on Scope, Context, Guidance, and Complexity.
+             
+Calibrate your evaluation against these compressed reference standards:
+- BACKEND (#13314): [backend] S3 operations fail with non-AWS object stores after AWS SDK v2 checksum defaults change. Clear isolated scope.
+- BUG TIER (#13180): [bug] fix: E2E test flakiness on K8s v1.34.0 — root cause analysis. High-quality root-cause analysis and environment data.
+- FRONTEND (#13108): [frontend] Adds coverage for frontend mock:api startup and enum drift. Explicit file paths and definitions of done.
+- SDK (#12865): [sdk] [bug] [set_accelerator_limit] rejects valid accelerator counts (only allows 0, 1, 2, 4, 8, 16). Elite precision with failing parameters.
+
+Compare the incoming issue detail density directly against the relevant blueprint standard above.
+
+Respond strictly following this format structure without other markdown wraps:
+
+- Each section MUST contain exactly 2 to 3 short, bullet fragments. 
+- Do NOT write full-length paragraphs or introductory text. Keep it highly concise for quick scanning.
+- Do NOT include any time frame or implementation window estimations.
+
+
+### 📊 Scope
+- <If the technical task boundaries are clear or ambiguous>
+- <If the issue isolates specific components, files, or packages correctly>
+
+
+### 📝 Context & Guidance
+<Evaluate if steps, expected behavior, or links are provided against repo standards>
+
+### ⚡ Complexity
+- <State difficulty tier: Low, Medium, or High, calibrated against this exact rubric:
+  * LOW: Task is isolated to single-file fixes, shallow tweaks, or documentation updates.
+  * MEDIUM: Task has moderate architectural depth, affecting internal logic patterns or specific layer wrappers.
+  * HIGH: Task has deep architectural depth or high breadth, spanning multiple system components simultaneously (e.g., changes across the SDK, backend, frontend, or argo compiler engines).>
+- <Break down the breadth (cross-layer impact) and depth (internal system complexity) of the proposed change>
+
+
+### 🎯 Overall Issue Quality Verdict
+- <State definitively if this is ready for immediate developer pickup>
+- <Outline the single most impactful recommendation to improve the issue quality>
+EOF
+
+
+USER_PROMPT="Title: ${TITLE} | Body: ${RAW_BODY}"
+
+if RESULT=$(gh models run "$TARGET_MODEL" --system-prompt "$SYSTEM_INSTRUCTIONS" "$USER_PROMPT" 2>&1); then
+    echo "✅ AI model executed successfully."
+    ANALYSIS_REPORT="$RESULT"
+else
+    echo "⚠️ CRITICAL: AI Model execution failed or preview tier limit hit."
+    echo "Error details: $RESULT"
+            
+    read -r -d '' ANALYSIS_REPORT << 'EOF' || true
+    ### ⚠️ Automated Triage Skipped
+    The issue body text or environment logs exceeded processing size boundaries for this triage pass.
+EOF
+fi
+
+echo "analysis<<EOF" >> $GITHUB_OUTPUT
+echo "$ANALYSIS_REPORT" >> $GITHUB_OUTPUT
+echo "EOF" >> $GITHUB_OUTPUT
+
+          
+echo "DEBUG: The raw analysis sent to output was:"
+echo "$ANALYSIS_REPORT"
--- a/.github/workflows/ai_analyzer.yml
+++ b/.github/workflows/ai_analyzer.yml
@@ -1,0 +1,37 @@
+name: AI Issue Quality Analyzer
+on:
+  issues:
+    types: [opened]
+
+env:
+  TARGET_MODEL: "openai/gpt-4o-mini"
+
+jobs:
+  analyze-issue:
+    runs-on: ubuntu-latest
+    permissions:
+      issues: write
+      contents: read
+      models: read
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v4
+
+      - name: Execute AI Issue Triage
+        id: ai_logic
+        env:
+          GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+          TITLE: ${{ github.event.issue.title }}
+          RAW_BODY: ${{ github.event.issue.body }}
+        run: |
+          ./.github/scripts/ai_triage.sh
+
+      - name: Apply Quality Analysis Comment
+        env:
+          GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+          ISSUE_NUMBER: ${{ github.event.issue.number }}
+          REPORT: ${{ steps.ai_logic.outputs.analysis }}
+        run: |
+          gh issue comment "$ISSUE_NUMBER" --body "## 🤖 AI Issue Quality Review
+          $REPORT"
+


droctothorpe:
@alyssacgoins Hey does this align with your grooming KEP? I want to make sure that we're not duplicating efforts.

modichika:
Hey @droctothorpe the already labeled good first issue in official kubeflow/pipelines i ran my updated version it labeled only one as good first issue and others as not as they required knowledge of codebase, can you test on them as well?

alyssacgoins:
> @alyssacgoins Hey does this align with your grooming KEP? I want to make sure that we're not duplicating efforts.

hey @droctothorpe this does overlap with my [KEP](https://github.com/kubeflow/pipelines/pull/13293) - but I really like the work here. I see you've added some feedback on my KEP (thank you! will address next week when I return from PTO) - @modichika would you mind taking a look at my KEP as well? 
My thinking is that we collaborate together on the KEP, get that merged, and I'll create a secondary PR that addresses anything out of scope for this PR.

modichika:
Hey @alyssacgoins thank you for the review. 
Yeah I would like to collaborate with you on the KEP.
I have looked into the KEP but I'll take a deeper look.

.github/workflows/ai_labeler.yml
```diff
@@ -0,0 +1,70 @@
...
+          GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+          TITLE: ${{ github.event.issue.title }}
+          BODY: ${{ github.event.issue.body }}
+        run: |
+          # 1. Install and run in the same step
+          gh extension install github/gh-models
+
+
+          # 2. Strict prompt to get ONLY good first issue label
+          RESULT=$(gh models run openai/gpt-4o-mini "You are a maintainer of the repo.
```
alyssacgoins: Do we need to worry about the long-term availability of this model for our use case? It's free for now - what is the plan for if this model becomes paid?

.github/workflows/ai_labeler.yml
```diff
@@ -0,0 +1,70 @@
...
+          TITLE: ${{ github.event.issue.title }}
+          BODY: ${{ github.event.issue.body }}
+        run: |
+          # 1. Install and run in the same step
+          gh extension install github/gh-models
+
+
+          # 2. Strict prompt to get ONLY good first issue label
+          RESULT=$(gh models run openai/gpt-4o-mini "You are a maintainer of the repo. 
+          Analyze if this issue is a good first issue based on these rules:
```
alyssacgoins: Did you compare `openai/gpt-4o-mini` against any other similar options?

.github/workflows/ai_labeler.yml
```diff
@@ -0,0 +1,70 @@
+name: AI Issue Labeler
+on:
+  issues:
+    types: [opened]
```
alyssacgoins: I think this should also run when an existing issue is modified

alyssacgoins (changes requested):
Have you tested this on your KFP fork? It would be useful to see some examples of this workflow in action

.github/workflows/ai_labeler.yml
modichika: Hey so i tested it and once it did fail with the rate limit error but the thing is before the issue reaches the llm we can make the issue template mandatory?

.github/workflows/ai_labeler.yml
modichika: updated this @alyssacgoins to address the modified ones

modichika:
yeah i did ran tests for it, it passed for the good first issues here's it : https://github.com/modichika/pipelines/actions/runs/25742713067/job/75597340390
@alyssacgoins you can look it labelled good first issue

But needs to address with the modified ones as well and with the fallbacks you mentioned.

.github/workflows/ai_labeler.yml
modichika: I haven't until now but i will compare it as well

alyssacgoins:
Hey @modichika,
@mprahl and I were thinking it would be useful to introduce a human component to your workflow, something like the following - 
1. As in your current workflow, when an issue is created or modified, the workflow kicks off and sends a request to `gpt-4o-mini` through `GH models`
2. Instead of prompting the model to identify if the issue is a good first issue, prompt the model to analyze the issue for general scope, context guidance and complexity (the same parameters you currently query) and respond with a general analysis for issue quality. You can also give the model several pre-approved issues for context on what a good issue looks like.
3. Prompt the model to use both the issue header and body to identify the section of the codebase targeted by the issue (issue header will often contain the section name outright). 
4. The script would then tag the owners and approvers for that section of the codebase (found in the OWNERS.md file for a subdirectory), paste in the feedback from `gpt-4o-mini` and ask those users to use their judgement and the AI analysis to decide if the issue should be labeled `/ready`.

What are your thoughts?

CC @droctothorpe

modichika:
Hey @alyssacgoins it's amazing let's implement this.
I'll implement pt. 2.
I need a little bit more context towards pt. 3 and 4, it would be great if it's clear exactly what's needed.
what do you think?

modichika:
@alyssacgoins made a new commit in the code to address pt. 2 here's a test output: https://github.com/modichika/pipelines/issues/58

Also lmk about the pre-approved issues which ones to give exactly.

I noticed one thing tho without the issue form that users fills before opening the issue, the llm might hallucinate and point out small things that might have been covered through the form for context and do not look at the complexity of the issue.

So maybe including the form will be good, what do you think?

.github/workflows/ai_labeler.yml
```diff
@@ -0,0 +1,80 @@
...
+    runs-on: ubuntu-latest
+    permissions:
+      issues: write
+      contents: read
+      models: read
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v4
+
+      - name: Get AI Label
```
alyssacgoins:
```suggestion
      - name: Execute AI Issue Triage
```
(or something similar)

.github/workflows/ai_analyzer.yml
```diff
@@ -0,0 +1,80 @@
...
+      - name: Checkout
+        uses: actions/checkout@v4
+
+      - name: Get AI Label
+        id: ai_logic
+        env:
+          GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+          TITLE: ${{ github.event.issue.title }}
+          BODY: ${{ github.event.issue.body }}
+        run: |
```
alyssacgoins, line 26: Consider placing this script in an external .sh file under `.github/scripts` rather than inline

alyssacgoins:
@modichika Great work! I added some minor comments, but this is exactly what we were thinking. I agree that it would be good context to include the issue template in the prompt. 
Here are some good examples of issues for each section of the codebase:

- **backend**: https://github.com/kubeflow/pipelines/issues/13314 
- **bug**: https://github.com/kubeflow/pipelines/issues/13180
- **frontend**: https://github.com/kubeflow/pipelines/issues/13108
- **sdk**: https://github.com/kubeflow/pipelines/issues/12865

modichika:
Hey @alyssacgoins added the pre-approved issues in the prompt including some tests and my observations here:

1. https://github.com/modichika/pipelines/issues/59  the difference between 1st llm's output and 2nd one is that in the 2nd one the template was filled accurately but the 1st one lacked two sections and you can see the llm's output improved.

2. https://github.com/modichika/pipelines/issues/60  in this issue testing, the token was so much that the llm couldn't handle so we need to think around this edge case.

Lmk your thoughts!

alyssacgoins:
> Hey @alyssacgoins added the pre-approved issues in the prompt including some tests and my observations here:
> 
> 1. [[sdk]  [bug] [set_accelerator_limit] rejects valid accelerator counts (only allows 0, 1, 2, 4, 8, 16) modichika/pipelines#59](https://github.com/modichika/pipelines/issues/59)  the difference between 1st llm's output and 2nd one is that in the 2nd one the template was filled accurately but the 1st one lacked two sections and you can see the llm's output improved.
> 2. [fix: E2E test flakiness on K8s v1.34.0 — root cause analysis modichika/pipelines#60](https://github.com/modichika/pipelines/issues/60)  in this issue testing, the token was so much that the llm couldn't handle so we need to think around this edge case.
> 
> Lmk your thoughts!

@modichika Glad to see such a noticeable change from adding the issue context - this is exactly what I was hoping for. Can you do some research and identify a couple slightly more powerful models offered through `GH models ` to test out on the second larger issue? We don't want people to be able deterred from creating detailed issues, so finding a powerful enough model is key here. 
Some supported models you can try:
- [OpenAI o4-mini](https://github.com/marketplace/models/azure-openai/o4-mini/playground) - while the name is quite similar to gpt-4o-mini, `o4` is specifically a reasoning model, and contains an expanded token window
- [Mistral Codestral 25.01](https://github.com/marketplace/models/azureml-mistral/Codestral-2501/playground) - great for rigid pattern/structural matching, which may work well for issue analysis

modichika:
Hey @alyssacgoins the problem with gh models are that github has set 4000 or 8000 tokens on every model by default so even if I changed the model and researched on loads of them this issue https://github.com/modichika/pipelines/issues/72
gave same limit error, I have solved the crashing error it won't crash even if the rate limit is there.

One concern is there: this pre-approved issue - https://github.com/kubeflow/pipelines/issues/13180  was created by @droctothorpe(a maintainer) do we need llm to analyse these kinds of issues?

Please look at the above test maybe we can come up with some solutions what are your thought?

.github/scripts/ai_issue_triage.sh
```diff
@@ -0,0 +1,56 @@
+set -e
+gh extension install github/gh-models --force || true
+
+echo "Using model: $TARGET_MODEL"
+
+read -r -d '' SYSTEM_INSTRUCTIONS << 'EOF'  || true
+You are an expert open-source maintainer for Kubeflow.
```
zazulam: Let's be specific and say `Kubeflow Pipelines`

.github/scripts/ai_issue_triage.sh
```diff
@@ -0,0 +1,56 @@
...
+- BUG TIER (#13180): [bug] fix: E2E test flakiness on K8s v1.34.0 — root cause analysis. High-quality root-cause analysis and environment data.
+- FRONTEND (#13108): [frontend] Adds coverage for frontend mock:api startup and enum drift. Explicit file paths and definitions of done.
+- SDK (#12865): [sdk] [bug] [set_accelerator_limit] rejects valid accelerator counts (only allows 0, 1, 2, 4, 8, 16). Elite precision with failing parameters.
+
+Compare the incoming issue detail density directly against the relevant blueprint standard above.
+
+Respond strictly following this format structure without other markdown wraps:
+
+### 📊 Scope
+<Evaluate task boundaries and estimate the time window required to complete it>
```
zazulam: Let's remove time window as it does not aid much in communicating the bug/feat scope.

.github/scripts/ai_issue_triage.sh
```diff
@@ -0,0 +1,56 @@
...
+
+### 📝 Context & Guidance
+<Evaluate if steps, expected behavior, or links are provided against repo standards>
+
+### ⚡ Complexity
+<Rate difficulty: Low, Medium, High, detailing architectural deepness vs shallow tweaks>
```
zazulam: If we don't define the the difficulty levels then the llm will determine it. There should be a mapping of some kind that where the breadth is taken into consideration along with the depth i.e. a bug fix that spans the sdk and argo compiler vs a feature that touches the sdk, frontend, backend that adds new modules in each.

.github/scripts/ai_triage.sh
```diff
@@ -0,0 +1,56 @@
...
+EOF
+fi
+
+echo "analysis<<EOF" >> $GITHUB_OUTPUT
+echo "$ANALYSIS_REPORT" >> $GITHUB_OUTPUT
+echo "EOF" >> $GITHUB_OUTPUT
+
+          
+echo "DEBUG: The raw analysis sent to output was:"
+echo "$ANALYSIS_REPORT"
```
alyssacgoins: EOF

.github/workflows/ai_labeler.yml
```diff
@@ -0,0 +1,37 @@
...
+          ./.github/scripts/ai_triage.sh
+
+      - name: Apply Quality Analysis Comment
+        env:
+          GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+          ISSUE_NUMBER: ${{ github.event.issue.number }}
+          REPORT: ${{ steps.ai_logic.outputs.analysis }}
+        run: |
+          gh issue comment "$ISSUE_NUMBER" --body "## 🤖 AI Issue Quality Review
+          $REPORT"
```
alyssacgoins: EOF

.github/scripts/ai_triage.sh
```diff
@@ -0,0 +1,56 @@
...
+             
+Calibrate your evaluation against these compressed reference standards:
+- BACKEND (#13314): [backend] S3 operations fail with non-AWS object stores after AWS SDK v2 checksum defaults change. Clear isolated scope.
+- BUG TIER (#13180): [bug] fix: E2E test flakiness on K8s v1.34.0 — root cause analysis. High-quality root-cause analysis and environment data.
+- FRONTEND (#13108): [frontend] Adds coverage for frontend mock:api startup and enum drift. Explicit file paths and definitions of done.
+- SDK (#12865): [sdk] [bug] [set_accelerator_limit] rejects valid accelerator counts (only allows 0, 1, 2, 4, 8, 16). Elite precision with failing parameters.
+
+Compare the incoming issue detail density directly against the relevant blueprint standard above.
+
+Respond strictly following this format structure without other markdown wraps:
```
alyssacgoins, line 19: I like that the the AI review is divided into sections - makes it very readable. I think we could potentially shorten the length of the review, however - I think you could cut down the size or switch the paragraphs to bullet points, for maintainers/approvers looking to quickly understand the issue analysis

.github/scripts/ai_triage.sh
modichika: hey @alyssacgoins I didn't understand this can you please elaborate?

alyssacgoins: @modichika EOF means end of file - the red circle above appears when there is a newline missing at the end of the file - can you add it?

.github/workflows/ai_labeler.yml
alyssacgoins: @modichika same as above

alyssacgoins (changes requested):

.github/scripts/ai_triage.sh
modichika: yeah added it!

.github/scripts/ai_triage.sh
```diff
@@ -0,0 +1,69 @@
+set -e
+gh extension install github/gh-models --force || true
+
+echo "Using model: $TARGET_MODEL"
+
+read -r -d '' SYSTEM_INSTRUCTIONS << 'EOF'  || true
+You are an expert open-source maintainer for Kubeflow pipelines.
```
mprahl:
```suggestion
You are an expert open-source maintainer for Kubeflow Pipelines.
```

.github/workflows/ai_labeler.yml
```diff
@@ -0,0 +1,38 @@
...
+        uses: actions/checkout@v4
+
+      - name: Execute AI Issue Triage
+        id: ai_logic
+        env:
+          GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+          TITLE: ${{ github.event.issue.title }}
+          RAW_BODY: ${{ github.event.issue.body }}
+        run: |
+          chmod +x ./.github/scripts/ai_triage.sh
```
mprahl: Nit: I think we can just commit the script with `chmod +x` already applied.

.github/workflows/ai_analyzer.yml
```diff
@@ -0,0 +1,37 @@
...
+          ./.github/scripts/ai_triage.sh
+
+      - name: Apply Quality Analysis Comment
+        env:
+          GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+          ISSUE_NUMBER: ${{ github.event.issue.number }}
+          REPORT: ${{ steps.ai_logic.outputs.analysis }}
+        run: |
+          gh issue comment "$ISSUE_NUMBER" --body "## 🤖 AI Issue Quality Review
+          $REPORT"
```
mprahl, line 36: The workflow never applies any labels. This step only creates a comment, and the earlier model step only emits prose, so the `AI Issue Labeler` job will not perform its advertised labeling behavior. Please add a label-selection output and an API/gh step that applies allowed repository labels, or rename/scope the workflow to issue-quality commenting.

.github/scripts/ai_triage.sh
```diff
@@ -0,0 +1,69 @@
...
+Compare the incoming issue detail density directly against the relevant blueprint standard above.
+
+Respond strictly following this format structure without other markdown wraps:
+
+- Each section MUST contain exactly 2 to 3 short, bullet fragments. 
+- Do NOT write full-length paragraphs or introductory text. Keep it highly concise for quick scanning.
+- Do NOT include any time frame or implementation window estimations.
+
+
+### 📊 Scope
```
mprahl, line 26: The system prompt does not ask the model to select labels or constrain it to this repo's label taxonomy. Before using model output for labeling, please include an explicit allowlist, selection rules, and a strict machine-readable schema such as JSON with `labels` and `comment`; otherwise the workflow will be forced to parse free-form prose and may request labels that do not exist.

.github/workflows/ai_labeler.yml
```diff
@@ -0,0 +1,37 @@
+name: AI Issue Labeler
+on:
+  issues:
+    types: [opened, edited]
```
mprahl: Because this also runs on every issue edit and always calls `gh issue comment`, routine edits can create repeated AI comments and model calls on the same issue. If edited-trigger retriage is needed, please update an existing bot comment or add gating/rate limiting so the workflow does not spam issue threads.

mprahl (commented):
PR Overview

This adds an AI triage workflow and shell script, but it currently comments on issues rather than labeling them.

.github/workflows/ai_analyzer.yml
modichika: yeah renamed the workflow!

.github/scripts/ai_triage.sh
modichika: now it's to add comment for maintainers and not to add tags.

modichika:
Thanks for the review @mprahl 
updated the workflow name, inner definitions from 'Labeler' to 'Analyzer' inside the code to  its behavior (commenting analysis prose). 
the trigger responds to opened to avoid many comments on edits

mprahl (approved):
/approve
/lgtm

PR #13378, opened 2026-05-10
feat(sdk): support namespace configuration via environment variable. Fixes #13338

### LINKED ISSUE #13338: [feature] Namespace as global constant and env variable in SDK

I would like to see the possibility of parameterizing the namespace parameter in the Client sdk during initialization. One way, it seems to me, is the environment variable

We have deployed kubeflow pipelines in cluster mode, but not in `namespace=kubeflow`, but in another `kubeflow-pipelines`. Therefore, we needed to make two edits locally to make the sdk compatible.

https://github.com/kubeflow/pipelines/blob/390ec4588aab82bfc3ba628672c3312c552cc75e/sdk/python/kfp/client/client.py#L146
https://github.com/kubeflow/pipelines/blob/390ec4588aab82bfc3ba628672c3312c552cc75e/sdk/python/kfp/client/client.py#L183

At the moment, I don't see anything except a local patch.

---

Love this idea? Give it a 👍.

lh1564803535-code:
## Summary

Allow the KFP SDK client namespace to be configured via the `KF_PIPELINES_NAMESPACE` environment variable, following the same pattern used for `host`, `client_id`, and other parameters.

## Motivation

The `Client.__init__` namespace parameter defaults to `'kubeflow'`, which is hardcoded in two places. Users who deploy Kubeflow Pipelines in a different namespace (e.g. `kubeflow-pipelines`) must pass `namespace=` explicitly on every `Client()` call — there is no way to set it globally via environment variable like other configuration options.

This forces users to either:
- Pass `namespace=` everywhere in their code
- Patch the SDK locally

See: https://github.com/kubeflow/pipelines/issues/13338

## Changes

- `sdk/python/kfp/client/client.py`:
  - Added `KF_PIPELINES_NAMESPACE_ENV = 'KF_PIPELINES_NAMESPACE'` module-level constant
  - Added `_DEFAULT_NAMESPACE = 'kubeflow'` class constant to eliminate magic string
  - Changed `namespace` parameter type from `str` (default `'kubeflow'`) to `Optional[str]` (default `None`)
  - Added resolution chain: explicit parameter → environment variable → default (`'kubeflow'`)
  - Replaced hardcoded `'kubeflow'` comparison with `self._DEFAULT_NAMESPACE`

- `sdk/python/kfp/client/client_test.py`:
  - Added `TestNamespaceResolution` class with 3 tests:
    - `test_namespace_from_env_variable` — env var is used when no explicit param
    - `test_namespace_explicit_overrides_env` — explicit param takes priority over env var
    - `test_namespace_default_when_nothing_set` — falls back to `'kubeflow'` when nothing is configured

## Backward Compatibility

This change is fully backward compatible:
- Users who don't set the env var and don't pass `namespace=` get the same behavior as before (`'kubeflow'`)
- Users who already pass `namespace=` explicitly are unaffected (explicit param takes highest priority)
- The only new behavior is that `KF_PIPELINES_NAMESPACE` env var is now respected as a middle-priority option

## Verification

$ python -m pytest sdk/python/kfp/client/client_test.py -v ======================== 53 passed in 1.93s ========================

All existing tests continue to pass. No regressions.

> If maintainers request additional coverage, I can add tests for empty string and explicit `None` edge cases.

Fixes #13338

/area sdk

### DIFF
--- a/sdk/python/kfp/client/client.py
+++ b/sdk/python/kfp/client/client.py
@@ -58,6 +58,7 @@ KF_PIPELINES_OVERRIDE_EXPERIMENT_NAME = 'KF_PIPELINES_OVERRIDE_EXPERIMENT_NAME'
 KF_PIPELINES_IAP_OAUTH2_CLIENT_ID_ENV = 'KF_PIPELINES_IAP_OAUTH2_CLIENT_ID'
 KF_PIPELINES_APP_OAUTH2_CLIENT_ID_ENV = 'KF_PIPELINES_APP_OAUTH2_CLIENT_ID'
 KF_PIPELINES_APP_OAUTH2_CLIENT_SECRET_ENV = 'KF_PIPELINES_APP_OAUTH2_CLIENT_SECRET'
+KF_PIPELINES_NAMESPACE_ENV = 'KF_PIPELINES_NAMESPACE'
 
 
 @dataclasses.dataclass
@@ -138,12 +139,14 @@ class Client:
 
     _LOCAL_KFP_CONTEXT = os.path.expanduser('~/.config/kfp/context.json')
 
+    _DEFAULT_NAMESPACE = 'kubeflow'
+
     # TODO: Wrap the configurations for different authentication methods.
     def __init__(
         self,
         host: Optional[str] = None,
         client_id: Optional[str] = None,
-        namespace: str = 'kubeflow',
+        namespace: Optional[str] = None,
         other_client_id: Optional[str] = None,
         other_client_secret: Optional[str] = None,
         existing_token: Optional[str] = None,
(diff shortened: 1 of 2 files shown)


lh1564803535-code:
Thanks for the thorough review.

Re namespace coupling: you're right -- I'll scope \KF_PIPELINES_NAMESPACE\ to only apply when the caller doesn't explicitly pass \
amespace=...\, so it doesn't override the multi-user namespace detection. Will update.

Re in-cluster test: agreed, will add a test that simulates in-cluster config with the env var set and asserts the constructed host uses the env-provided namespace.

Re YAPF formatting: noted, will reformat to match the existing multi-line style.

PR #11673, opened 2025-02-25
fix(backend) fix execution-level retry on the Argo Workflows backend

ntny:
**Description of your changes:**

**Checklist:**
- [ ] You have [signed off your commits](https://www.kubeflow.org/docs/about/contributing/#sign-off-your-commits)
- [ ] The title for your pull request (PR) should follow our title convention. [Learn more about the pull request title convention used in this repository](https://github.com/kubeflow/pipelines/blob/master/CONTRIBUTING.md#pull-request-title-convention). 

close https://github.com/kubeflow/pipelines/issues/9950 

In this [PR](https://github.com/kubeflow/pipelines/pull/11585), the manual retry via the Pipeline Run UI Page button was fixed.
However [set_retry]( https://kubeflow-pipelines.readthedocs.io/en/latest/source/dsl.html#kfp.dsl.PipelineTask.set_retry) at the execution level still does not work (for a different reason).

**Description of changes**: 
- `retryPolicy` is present in the generated KFP Pipeline Spec, but it is not translated into the Argo Workflow spec. Argo Workflows uses the [retryStrategy](https://argo-workflows.readthedocs.io/en/latest/retries/) instead.
- `retryPolicy` from the KFP Pipeline spec is mapped into the Argo Workflow spec under dag.task.input.retryPolicy, but Argo Workflows ignores it. I would assume that input.retryPolicy should be handled somewhere in the launcher; however, the only usage of this field I found is in the UI, and it doesn’t look like retry handling.

This PR adds the conversion of retryPolicy to the Argo Workflow retryStrategy for tasks.

in attached the Argo yaml spec which was generated in master branch from [this](https://github.com/kubeflow/pipelines/pull/11673/files#diff-b8ad378407da3d23983f50ce8c20238c8851530f9c836c1139ae73440b1afdfb) pipeline spec

[hello-world-with-retry-master.txt](https://github.com/user-attachments/files/19027510/hello-world-with-retry-master.txt)

PR #13386, opened 2026-05-12
feat(sdk/backed): Add retryPolicy Option to set_retry

JerT33:

Currently `set_retry` only retries when the container exits non-zero (Argo's default `OnFailure`). Pods that disappear due to eviction, node loss, etc end up in Argo's error phase and are never retried. The only existing workaround is a cluster-wide `retryStrategy` default in the Argo workflow controller cm

This PR adds an optional `policy` parameter to `set_retry()` that maps to Argo Workflow `retryStrategy.retryPolicy`. Individual components can now opt into retry on infra level failures without any cluster level config

```python
hello_world().set_retry(num_retries=3, policy='OnError')
```

Valid values: `'Always'`, `'OnFailure'`, `'OnError'`, `'OnTransientError'`. Defaults to `None`, which compiles to Argo's default `OnFailure`  (existing pipelines are unchanged)

Implementation notes:

- Proto: new `PipelineTaskSpec.RetryPolicy.Policy` enum (`POLICY_UNSPECIFIED`, `POLICY_ALWAYS`, `POLICY_ON_FAILURE`, `POLICY_ON_ERROR`, `POLICY_ON_TRANSIENT_ERROR`).
- SDK: validates the string value at `set_retry()` call time and again at proto serialization.
- Backend: the policy can't be parameterized — Argo strict-validates `retryStrategy.retryPolicy` against the enum at workflow submission, so any placeholder string fails. Instead, the policy is baked into the impl template name (`retry-system-container-impl-onerror` etc.) and the literal policy value is emitted into the `retryStrategy`. Tasks with the same policy still share a template.

## Live Cluster Testing

**Before** (no `policy`):

```python
@dsl.pipeline(name='retry-policy-on-error-demo')
def retry_policy_demo():
    hello_world().set_retry(num_retries=3)
```

Delete the running pod → Argo marks the node Error → no retry → workflow fails.

Screenshot 2026-05-31 at 2 15 30 PM
Screenshot 2026-05-31 at 2 16 05 PM
Screenshot 2026-05-31 at 2 16 14 PM

**After** (`policy='OnError'`):

```python
@dsl.pipeline(name='retry-policy-on-error-demo')
def retry_policy_demo():
    hello_world().set_retry(num_retries=3, policy='OnError')
```

Delete the running pod → Argo marks the node Error → retry scheduled → workflow succeeds on attempt 2.

Screenshot 2026-05-31 at 2 43 08 PM
Screenshot 2026-05-31 at 2 43 25 PM
Screenshot 2026-05-31 at 2 44 03 PM

**Checklist:**

### DIFF
--- a/sdk/python/kfp/dsl/pipeline_task.py
+++ b/sdk/python/kfp/dsl/pipeline_task.py
@@ -605,12 +605,15 @@ class PipelineTask:
 
         return self
 
+    _VALID_RETRY_POLICIES = frozenset(structures._RETRY_POLICY_MAP)
+
     @block_if_final()
     def set_retry(self,
                   num_retries: int,
                   backoff_duration: Optional[str] = None,
                   backoff_factor: Optional[float] = None,
-                  backoff_max_duration: Optional[str] = None) -> 'PipelineTask':
+                  backoff_max_duration: Optional[str] = None,
+                  policy: Optional[str] = None) -> 'PipelineTask':
         """Sets task retry parameters.
 
         Args:
@@ -618,15 +621,26 @@ class PipelineTask:
             backoff_duration: Number of seconds to wait before triggering a retry. Defaults to ``'0s'`` (immediate retry).
             backoff_factor: Exponential backoff factor applied to ``backoff_duration``. For example, if ``backoff_duration="60"`` (60 seconds) and ``backoff_factor=2``, the first retry will happen after 60 seconds, then again after 120, 240, and so on. Defaults to ``2.0``.
             backoff_max_duration: Maximum duration during which the task will be retried. Defaults to ``'3600s'``.
+            policy: Controls which failure types trigger a retry. Only supported
+                when using Argo Workflows as the pipeline execution engine. One of
+                ``'Always'``, ``'OnFailure'`` (default), ``'OnError'``, or
+                ``'OnTransientError'``. Use ``'OnError'`` or ``'Always'`` to retry on
+                node-level failures such as eviction or SIGTERM (exit code 143).
+                Defaults to ``None`` (equivalent to ``'OnFailure'``).
 
         Returns:
             Self return to allow chained setting calls.
         """
+        if policy is not None and policy not in self._VALID_RETRY_POLICIES:
+            raise ValueError(
+                f'Invalid retry policy {policy!r}. '
+                f'Must be one of: {sorted(self._VALID_RETRY_POLICIES)}')
         self._task_spec.retry_policy = structures.RetryPolicy(
             max_retry_count=num_retries,
             backoff_duration=backoff_duration,
             backoff_factor=backoff_factor,
             backoff_max_duration=backoff_max_duration,
+            policy=policy,
         )
         return self
 
--- a/sdk/python/kfp/dsl/pipeline_task.py
+++ b/sdk/python/kfp/dsl/pipeline_task.py
@@ -610,12 +610,15 @@ class PipelineTask:
 
         return self
 
+    _VALID_RETRY_POLICIES = frozenset(structures._RETRY_POLICY_MAP)
+
     @block_if_final()
     def set_retry(self,
                   num_retries: int,
                   backoff_duration: Optional[str] = None,
                   backoff_factor: Optional[float] = None,
-                  backoff_max_duration: Optional[str] = None) -> 'PipelineTask':
+                  backoff_max_duration: Optional[str] = None,
+                  policy: Optional[str] = None) -> 'PipelineTask':
         """Sets task retry parameters.
 
         Args:
@@ -623,15 +626,32 @@ class PipelineTask:
             backoff_duration: Number of seconds to wait before triggering a retry. Defaults to ``'0s'`` (immediate retry).
             backoff_factor: Exponential backoff factor applied to ``backoff_duration``. For example, if ``backoff_duration="60"`` (60 seconds) and ``backoff_factor=2``, the first retry will happen after 60 seconds, then again after 120, 240, and so on. Defaults to ``2.0``.
             backoff_max_duration: Maximum duration during which the task will be retried. Defaults to ``'3600s'``.
+            policy: Controls which failure types trigger a retry. Only supported
+                when using Argo Workflows as the pipeline execution engine. One of
+                ``'Always'``, ``'OnFailure'``, ``'OnError'``, or
+                ``'OnTransientError'``. ``'OnFailure'`` retries container exits,
+                including SIGTERM (exit code 143); ``'OnError'`` retries pods that
+                report no exit code, such as deletion or node loss;
+                ``'OnTransientError'`` retries either when the failure message
+                matches the controller's transient-error pattern; ``'Always'``
+                retries both.
+                Defaults to ``None``, which omits the policy so the
+                deployment's Argo configuration applies; KFP's bundled
+                manifests set ``'OnError'``.
 
         Returns:
             Self return to allow chained setting calls.
         """
+        if policy is not None and policy not in self._VALID_RETRY_POLICIES:
+            raise ValueError(
+                f'Invalid retry policy {policy!r}. '
+                f'Must be one of: {sorted(self._VALID_RETRY_POLICIES)}')
         self._task_spec.retry_policy = structures.RetryPolicy(
             max_retry_count=num_retries,
             backoff_duration=backoff_duration,
             backoff_factor=backoff_factor,
             backoff_max_duration=backoff_max_duration,
+            policy=policy,
         )
         return self
 
--- a/sdk/python/kfp/dsl/structures.py
+++ b/sdk/python/kfp/dsl/structures.py
@@ -358,6 +358,14 @@ class ContainerSpecImplementation:
             resources=None)  # can only be set on tasks
 
 
+_RETRY_POLICY_MAP = {
+    'Always': 'POLICY_ALWAYS',
+    'OnFailure': 'POLICY_ON_FAILURE',
+    'OnError': 'POLICY_ON_ERROR',
+    'OnTransientError': 'POLICY_ON_TRANSIENT_ERROR',
+}
+
+
 @dataclasses.dataclass
 class RetryPolicy:
     """The retry policy of a container execution.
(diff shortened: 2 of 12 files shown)


sdk/python/kfp/dsl/pipeline_task.py
```diff
@@ -605,28 +605,41 @@ def set_memory_limit(
...
+                  backoff_max_duration: Optional[str] = None,
+                  policy: Optional[str] = None) -> 'PipelineTask':
         """Sets task retry parameters.
 
         Args:
             num_retries : Number of times to retry on failure.
             backoff_duration: Number of seconds to wait before triggering a retry. Defaults to ``'0s'`` (immediate retry).
             backoff_factor: Exponential backoff factor applied to ``backoff_duration``. For example, if ``backoff_duration="60"`` (60 seconds) and ``backoff_factor=2``, the first retry will happen after 60 seconds, then again after 120, 240, and so on. Defaults to ``2.0``.
             backoff_max_duration: Maximum duration during which the task will be retried. Maximum duration is 1 hour (3600s). Defaults to ``'3600s'``.
+            policy: Controls which failure types trigger a retry. One of ``'Always'``,
```
demarna1: Need to be clear in the docstring that this parameter will only work if using ArgoWorkflows

JerT33:
@ntny could I get a review/opinions on this change when you are able
(tagging due to review on previous retry pr)

ntny:
It's great that you found the same issue :)

I recently opened a PR: https://github.com/kubeflow/pipelines/pull/13456

It enables retries for infrastructure-related `Error` failures (except OOM) without introducing any additional configuration.

What do you think about this approach? It seems to me that, from a user's perspective, `set_retry()` should retry all reasonable failures. The distinction between `Failed` and `Error` is an Argo compiler implementation detail, so I tend to view the current behavior as a bug rather than an intentional feature.

Curious to hear your thoughts.

AWF execution-level retry implementation PR: https://github.com/kubeflow/pipelines/pull/11673

@juliusvonkohout, what do you think about this? I remember you were also involved in the discussions around fixing retries.

juliusvonkohout:
I will try to provide feedback tomorrow. I did provide a proper expression for transient errors (espescially cluster autoscaler) here https://github.com/kubeflow/pipelines/blob/63dac612a3d8ce4f45b7113dba41701ebe1df811/manifests/kustomize/third-party/argo/base/workflow-controller-deployment-patch.yaml#L20-L25 and have to switch soon the Argo default workflow template configuration to OnError which includes transient. We also need to expose all the retry policy options in the sdk. Default OnError is good via argo, but we must be able to opt out via the SDK (always, onfailure, transient)

ntny:
Agreed. It makes sense to change the defaults and still allow users to override the retry policy when needed.
@juliusvonkohout, if I understand correctly, you're suggesting changing the default retry policy in the workflow controller deployment.

I think it would be better to have a single place that defines retry behavior and override the defaults there, as in https://github.com/kubeflow/pipelines/pull/13456.

juliusvonkohout:
Please lgtm https://github.com/kubeflow/pipelines/pull/13485

JerT33:
CI tests failing unrelated MLflow tests. Fix for CI implemented in #14260. This PR is temporality stacked to utilize CI fix

jeffspahr (changes requested):
I found two runtime correctness issues and two documentation issues that should be addressed before merge.

1. **[P1] KFP's global retry default overrides and multiplies task-level policies.**

   The selected policy is emitted only on the [inner implementation template](https://github.com/kubeflow/pipelines/blob/578da220d57bf2918e5fef4a57421796eeb2f1c6/backend/src/v2/compiler/argocompiler/container.go#L593-L599); the [outer DAG wrapper](https://github.com/kubeflow/pipelines/blob/578da220d57bf2918e5fef4a57421796eeb2f1c6/backend/src/v2/compiler/argocompiler/container.go#L415-L456) has no strategy. Stock KFP applies `limit: 2, retryPolicy: OnError` to every template. Consequently, `OnFailure`/`OnTransientError` can still retry through the wrapper, while `OnError`/`Always` can exceed `num_retries`--for example, three inner retries can become twelve total attempts. Please explicitly neutralize retries on the wrapper or ensure exactly one layer owns retry behavior.

2. **[P1] Reusing a nested pipeline with different policies silently keeps only the first policy.**

   The new [policy-specific template naming](https://github.com/kubeflow/pipelines/blob/578da220d57bf2918e5fef4a57421796eeb2f1c6/backend/src/v2/compiler/argocompiler/container.go#L387-L393) does not cover nested pipeline invocations. The visitor [deduplicates components by name](https://github.com/kubeflow/pipelines/blob/578da220d57bf2918e5fef4a57421796eeb2f1c6/backend/src/v2/compiler/visitor.go#L89-L95) after [mutating shared child tasks](https://github.com/kubeflow/pipelines/blob/578da220d57bf2918e5fef4a57421796eeb2f1c6/backend/src/v2/compiler/visitor.go#L163-L166). Thus two calls such as `inner().set_retry(..., 'OnError')` and `inner().set_retry(..., 'OnFailure')` share one compiled descendant policy. Please compile inherited retry configurations per invocation instead of mutating the canonical component.

3. **[P2] `None` is not equivalent to `OnFailure` in the bundled deployment.**

   The [SDK documentation](https://github.com/kubeflow/pipelines/blob/578da220d57bf2918e5fef4a57421796eeb2f1c6/sdk/python/kfp/dsl/pipeline_task.py#L629-L634) and [proto comment](https://github.com/kubeflow/pipelines/blob/578da220d57bf2918e5fef4a57421796eeb2f1c6/api/v2alpha1/pipeline_spec.proto#L602-L612) claim `OnFailure`, but the compiler [omits the policy](https://github.com/kubeflow/pipelines/blob/578da220d57bf2918e5fef4a57421796eeb2f1c6/backend/src/v2/compiler/argocompiler/container.go#L703-L717), allowing KFP's `OnError` template default to fill it. Please either emit `OnFailure` or document that unset defers to deployment/controller configuration.

4. **[P2] The SIGTERM guidance recommends the wrong policy.**

   A main container exiting 143 is a failed node, so `OnError` will not retry it; `OnFailure` or `Always` will. This conflicts with [Argo's retry-policy semantics](https://github.com/argoproj/argo-workflows/blob/main/docs/retries.md). Please correct the guidance at [pipeline_task.py:632](https://github.com/kubeflow/pipelines/blob/578da220d57bf2918e5fef4a57421796eeb2f1c6/sdk/python/kfp/dsl/pipeline_task.py#L632-L634).

Verification: `go test ./backend/src/v2/compiler/argocompiler` and Python syntax checks passed; the current GitHub CI suite is green.

sdk/python/kfp/dsl/structures.py
```diff
@@ -383,13 +398,19 @@ def to_proto(self) -> pipeline_spec_pb2.PipelineTaskSpec.RetryPolicy:
         backoff_duration_seconds = f'{convert_duration_to_seconds(backoff_duration)}s'
         backoff_max_duration_seconds = f'{convert_duration_to_seconds(backoff_max_duration)}s'
 
+        d = {
```
juliusvonkohout, line 401: what is d ? please use professional long expressive variable names.

jeffspahr (changes requested):
Re-reviewed commit 2733b881860bd1b1e91c15ac86cdc05845a8ce7a. One runtime issue from my earlier review remains incompletely addressed.

**[P1] Enclosing DAG retries can still bypass task retry settings.**

The new zero-limit strategy at [container.go:457–464](https://github.com/kubeflow/pipelines/blob/2733b881860bd1b1e91c15ac86cdc05845a8ce7a/backend/src/v2/compiler/argocompiler/container.go#L457-L464) disables retries only on the immediate container-executor wrapper. The generated `root` and `entrypoint` DAG templates still have no explicit strategy, so the bundled controller configuration supplies `limit: 2, retryPolicy: OnError`. Argo propagates child Error phases to the enclosing DAG and applies its retry strategy, allowing an enclosing DAG retry despite the task's selected policy or exhausted retry count. Please ensure ancestor retries cannot bypass task settings and add coverage for the effective strategies under the bundled template defaults.

Verification: `go test ./backend/src/v2/compiler/argocompiler` passed. A targeted check applying Argo's strategic default merge to the new workflow golden confirmed limit `0` on the immediate wrapper but limit `2` with `OnError` on both `root` and `entrypoint`. Runtime propagation was checked against Argo v4.1.2 controller source; I did not perform a live cluster reproduction.

The SDK/proto documentation findings are addressed. I also need to correct my earlier nested-pipeline finding: ordinary SDK invocations deep-copy and rename their component definitions, so the SDK example I gave does not demonstrate shared policies. I withdraw that finding for ordinary SDK usage. The shared-component limitation in directly supplied IR predates this PR.

Separately, [pre-commit is failing](https://github.com/kubeflow/pipelines/actions/runs/34170647685/job/101890153264): docformatter wraps the existing `TestTaskInFinalState` docstring in the touched `sdk/python/kfp/dsl/pipeline_task_test.py`. Please apply the hook's formatting change. Other substantive checks passed at the time of this review.

JerT33:
> **[P1] Enclosing DAG retries can still bypass task retry settings.**
> 
> The new zero-limit strategy at [container.go:457–464](https://github.com/kubeflow/pipelines/blob/2733b881860bd1b1e91c15ac86cdc05845a8ce7a/backend/src/v2/compiler/argocompiler/container.go#L457-L464) disables retries only on the immediate container-executor wrapper. The generated `root` and `entrypoint` DAG templates still have no explicit strategy, so the bundled controller configuration supplies `limit: 2, retryPolicy: OnError`. Argo propagates child Error phases to the enclosing DAG and applies its retry strategy, allowing an enclosing DAG retry despite the task's selected policy or exhausted retry count. Please ensure ancestor retries cannot bypass task settings and add coverage for the effective strategies under the bundled template defaults.

@juliusvonkohout @jeffspahr

I would like some open discussion on the desired behavior in this case. It is my understanding that #13485 introduces this main concern found here. If a user uses the stock per-cluster retry default, which is `limit: 2`, since there are 4 levels of templates this applies to, this will result in up to 3^4 = 81 pod executions

```
 entrypoint   [DAG]  limit 2 -> 3 attempts
   root       [DAG]  limit 2 -> 3 attempts
     executor [DAG]  limit 2 -> 3 attempts
       impl   [POD]  limit 2 -> 3 attempts
```

The current state of this PR should only override the pod level retry to maintain backwards compatibility with the current approach. This is currently being flagged. However I feel this is a larger behavioral change I would like alignment on first before moving ahead with any implementation

The way I see it we have 3 options
1. Remove the multiple layers of retries that is currently set. That would align the default retry policy with the expected behavior (default retry of 2 will result in pod running max 3 times). Then follow this up by setting the current pod level retry override on user provided configuration on the task
2. Preserve all existing default retry behavior, just override the per task retries if set by user (current PR implementation)
3. A middle solution where if `set_retry` is configured, then we remove the default retry policy on all layers above (fixes issue called out by Jeff, maintains most backwards compatibility with existing behavior)

juliusvonkohout:
I am in favor of "A middle solution where if set_retry is configured, then we remove the default retry policy on all layers above (fixes issue called out by Jeff, maintains most backwards compatibility with existing behavior)"

juliusvonkohout:
there seems to be a bug in the github actions. CC @droctothorpe

PR #13396, opened 2026-05-14
feat(sdk): Add set_debug_pause method to PipelineTask 

### LINKED ISSUE #13316: [feature] Implement debug-pause type capabilities into SDK
### Feature Area

/area backend
/area sdk

Argo workflows has a feature `debug-pause` that allows users to pause a workflow step for debugging purposes. This pause can occur before or after the step runs. Additionally, the after version has an option to only trigger on failure. The user can then interact with the pod with kubectl exec. Currently, KFP does not allow users to implement this feature on their components via the SDK. Adding this feature would be a powerful new tool for developers and end users as it would give additional flexibility for investigating their pipelines. We have tested adding the env vars manually into IRs and the feature does work. However, we should make this native to KFP SDK.

This proposal adds the ability to include a new function into the SDK `kfp.kubernetes.enable_debug_pause()`. This function would be applicable to tasks defined in a pipeline. The function accepts before, after, and on_error_only parameters to configure which pause mode is applied. Once applied, the function adds one of the following env vars to the task based on the user's configuration: `ARGO_DEBUG_PAUSE_BEFORE`,
 `ARGO_DEBUG_PAUSE_AFTER`, or `ARGO_DEBUG_PAUSE_ON_ERROR`. This will then be detected by Emissary and the respective behavior will be followed. Once debugging is complete, the user releases the pod by running `kubectl exec <pod> -c main -- sh -c "touch /var/run/argo/ctr/main/proceed"`.
 
 ### Alternate Design
 An alternate consideration should be made about implementing this outside of argo workflows and the native debug pause implementation. The question being, could we pause our component in a way such that we aren't restricted by the structure and format of the argo workflows architecture. One option to do this would be to implement this via a pause in the `launcher` directly that occurs once the python code is done running. This would allow us to pause within the `launcher` itself, as opposed to `emissary` which is where `debug-pause` gets checked and triggered. While there is a theoretical benefit of shelling into the environment while the `launcher` process is still alive, since both `emissary` and `launcher` run on the same container the filesystem and environment are identical either way, providing no meaningful additional debugging capability. There is added complexity of making changes to the backend as well to implement this vision.

Feature would allow developers and end users to shell into and interact with a component after it is done running. This would allow for greatly increased debugging, component understand and investigation, and accessibility for our developers and end users. 

Without this feature users cannot systematically pause their component. 

@zazulam @droctothorpe 
---

Love this idea? Give it a 👍.

carlosdanielarroyo:
# Summary
Argo workflows has a feature [`debug-pause`](https://argo-workflows.readthedocs.io/en/latest/debug-pause/) that allows users to pause a workflow step for debugging purposes. This pause can occur before or after the step runs. Additionally, the after version has an option to only trigger on failure. The user can then interact with the pod with `kubectl exec`. Currently, KF does not allow users to implement this feature on their components via the SDK. Adding this feature would be a powerful new tool for developers and end users as it would give additional flexibility for investigating their pipelines.

This proposal adds a new mehod to `PipelineTask` int he core SDK: `task.set_debug_pause()`. This method is applicable to tasks degined in a pipeline. 
It accepts `before`, `after`, and `on_error` parameters to configure which pause mode is applied. Example below:

```python
@dsl.pipeline
def my_pipeline():
    task = my_component()

    # Pause after execution (default)
    task.set_debug_pause()
    
    # Pause before execution
    task.set_debug_pause(before=True, after=False)
    
    # Pause on error
    task.set_debug_pause(on_error=True)
```

Once applied, the method adds one of the following env vars to the task:
- `ARGO_DEBUG_PAUSE_AFTER=true`: Pause after execution (default)
- `ARGO_DEBUG_PAUSE_BEFORE=true`: Pause before execution
- `ARGO_DEBUG_PAUSE_ON_ERROR=true`: Pause on error

Once the component is running, the `debug-pause` feature will then be activated. It should be noted that we have tested adding these env vars manually on runs kicked off with KFP and the feature does work once the env vars are added. The ultimate purpose of this PR, and the remaining steps to make this native to KFP, is the ability to implement this via the SDK. Once debugging is complete, the users releases the pod by running `kubectl exec <pod> c main -- sh -b "touch /var/run/argo/ctr/main/proceed`

### Frontend changes
The only effect to the frontend would be that components that are paused "after" would still have a green check. This is because once the task code is done running the `launcher` then updates the `MLMD` stating the task is completed. This then cascades to the KFP UI. Therefore visually the components looks "done" but in reality `emissary` has paused this pod from completing. The user could verify the state pod's actual state via `kubectl get pods` or the Argo UI. This can cause confusion. Therefore, we'd need to consider a future to update the backend and frontend to make the pause state clearer in the UI. However, as of now, there will be no impact in the UI. 

### Local Experience
No changes to the KFP local experience

### Testing

- Added a new suite of unit tests in `sdk/python/kfp/dsl/pipeline_task_test.py` to ensure the new function behaves as expected. 

Note: Confirmed in manual testing that setting ARGO_DEBUG_PAUSE_* env var was enough to trigger pause behavior. No annotation required. Additionally, and the reason for testing only with env var, is that the annotation is not referred to in the argo [debug-pause](https://argo-workflows.readthedocs.io/en/latest/debug-pause/) page.

**Checklist:**


kubernetes_platform/python/kfp/kubernetes/debug.py
```diff
@@ -0,0 +1,92 @@
...
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+from google.protobuf import json_format
+from kfp.dsl import PipelineTask
+from kfp.kubernetes import common
```
droctothorpe: Is this an unused import?

kubernetes_platform/python/kfp/kubernetes/debug.py
```diff
@@ -0,0 +1,92 @@
+# Copyright 2026 The Kubeflow Authors
```
droctothorpe: Why put it in the kubernetes-platform and not the kfp package?

carlosdanielarroyo: `enable_debug_pause()` is leveraging the Argo Workflows debug-pause feature. Argo Workflows is a "Kubernetes workflow engine", and therefore is Kubernetes specific, meaning it wouldn't apply to other backends. Because of this, I placed it in kfp-kubernetes rather than the core KFP SDK, which I assumed was intended for backend agnostic functions/methods/etc.

kubernetes_platform/python/kfp/kubernetes/debug.py
carlosdanielarroyo: Oh yes, will remove now, thanks for catching

kubernetes_platform/python/kfp/kubernetes/debug.py
droctothorpe: I think from an end user's perspective, that distinction is irrelevant. @HumairAK curious to hear your thoughts on this. The backend agnosticism req came from Google and is no longer a requirement.

droctothorpe:
FYI, @carlosdanielarroyo, you pushed up an unsigned commit.

kubernetes_platform/python/kfp/kubernetes/debug.py
zazulam: @carlosdanielarroyo I would prefer to place the method within the kfp package itself given argo is the only backend that is officially supported.

carlosdanielarroyo: Thank you both for your feedback. Just made the change and moved it to the SDK as a method to PipelineTask. Working on some CICD errors, but hoping to have ready soon.

carlosdanielarroyo:
Hi @droctothorpe @zazulam , moved the code to the SDK and CICD is now passing. I believe this is ready for re-review

zazulam (approved):
/lgtm
/approve

PR #13400, opened 2026-05-18
 sdk: add namespace fallback to list_pipelines, list_recurring_runs,   upload_pipeline

### LINKED ISSUE #13397: [feature] Add namespace fallback to `upload_pipeline`, `list_pipelines`, and `list_recurring_runs` in KFP Python SDK
### Feature Area

/area sdk

Add the same namespace resolution used by other `kfp.Client` methods to `upload_pipeline`, `list_pipelines`, and `list_recurring_runs`:
```python
namespace = namespace or self.get_user_namespace()
```
at the start of each method body, before calling the underlying API.

Today, several client methods already resolve namespace this way (for example `create_experiment`, `list_experiments`, `get_experiment`, and `list_runs`). The three methods above accept an optional namespace argument but pass it through unchanged when callers omit it, even when the client was configured with a user namespace via `Client(namespace=...)` or the active kubeconfig context.

After the change, callers can omit namespace and still operate in their configured user namespace, consistent with the rest of the client API.

Affected code: sdk/python/kfp/client/client.py — methods `upload_pipeline`, `list_pipelines`, and `list_recurring_runs`.

On multi-user Kubeflow deployments, users typically set their namespace once (kubeconfig context or `kfp.Client(namespace="kubeflow-user-example-com")`). Tools and integrations (Kale, CI scripts, notebooks) then call client methods without repeating the namespace on every call.

When `namespace` is omitted:
- upload_pipeline may upload to the wrong scope or fail because `namespace=None` is sent to the API.
- `list_pipelines` and `list_recurring_runs` may list resources across all namespaces or behave differently from `list_runs` / `list_experiments`, which already honor the configured namespace.

This inconsistency forces every caller to thread `namespace` through manually and is easy to get wrong in automation that only configures the client once.

Yes:
1. Pass `namespace` explicitly on every call, e.g. `client.upload_pipeline(..., namespace=client.get_user_namespace())`.
2. Or always pass the same string you used when constructing `Client(namespace=...)`.
Both work but duplicate logic that other client methods already encapsulate and are easy to forget when adding new code paths.
---

Love this idea? Give it a 👍.

shreeharshshinde:
Fixes #13397
  
  ### What
  
  Three client methods accepted `namespace` but didn't fall back to the
  client-configured namespace when it was omitted. In multi-user mode
  this
  caused `upload_pipeline` to fail with HTTP 400 and `list_*` methods
  to
  query across all namespaces instead of the configured one.
  
  ### Change
  
  Applied `namespace = namespace or self.get_user_namespace()` to:
  - `list_pipelines`
  - `list_recurring_runs`
  - `upload_pipeline`
  
  This is the same pattern already used by `create_experiment`,
  `list_experiments`, `get_experiment`, and `list_runs`.
  
  Also updated the `upload_pipeline` docstring to reflect the new
  default
  behaviour.
  
  ### Tests
  
  7 new unit tests in `client_test.py`:
  - fallback fires when namespace is omitted
  - explicit namespace is not overridden
  - `list_recurring_runs` with `experiment_id` does not forward
  namespace
  
  ### Notes
  
  - No behaviour change for single-user deployments
  (`get_user_namespace()`
    returns `''`, backend blanks it via `ReplaceNamespace` anyway).

### DIFF
--- a/sdk/python/kfp/client/client.py
+++ b/sdk/python/kfp/client/client.py
@@ -1422,10 +1422,11 @@ class Client:
             pipeline_name: Name of the pipeline to be shown in the UI.
             description: Description of the pipeline to be shown in the UI.
             namespace: Optional. Kubernetes namespace where the pipeline should
-                be uploaded. Defaults to the namespace configured on the client
-                (e.g. via ``Client(namespace=...)`` or
-                ``set_user_namespace()``). For single-user deployments, leave
-                it as ``None``.
+                be uploaded. If omitted or set to ``None``, the client uses
+                its configured namespace (for example via
+                ``Client(namespace=...)`` or ``set_user_namespace()``). In
+                single-user deployments, this effectively uses the empty
+                namespace.
 
         Returns:
             ``V2beta1Pipeline`` object.
--- a/sdk/python/kfp/client/client.py
+++ b/sdk/python/kfp/client/client.py
@@ -668,6 +668,7 @@ class Client:
         Returns:
             ``V2beta1ListPipelinesResponse`` object.
         """
+        namespace = namespace or self.get_user_namespace()
         return self._pipelines_api.pipeline_service_list_pipelines(
             namespace=namespace,
             page_token=page_token,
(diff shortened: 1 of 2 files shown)


droctothorpe:
Hi @shreeharshshinde, thank you for your PR! The linked issue (#13397) must be labeled `ready` before your contribution can be reviewed. Please check our [CONTRIBUTING.md](https://github.com/kubeflow/pipelines/blob/master/CONTRIBUTING.md) for details.

PR #13405, opened 2026-05-19
feat(sdk): add kubeflow_client PipelinesClient (phase 1). Fixes #13380

### LINKED ISSUE #13380: [SDK] Implement PipelinesClient (KEP-125) - Phase 1
### Feature Area

/area sdk

Implement PipelinesClient in the KFP SDK as specified in [KEP-125](https://github.com/kubeflow/sdk/tree/main/docs/proposals/125-kfp-client).

PipelinesClient is a new, simplified, name-first client for Kubeflow Pipelines that will be implemented in the KFP repository.

Today, Kubeflow users who want to orchestrate ML pipelines must install and use kfp separately from the Kubeflow SDK. This creates friction:

- Two SDKs, two import styles — users mix from kfp import ... with from kubeflow.trainer import ...
- Inconsistent constructor — kfp.Client(host=..., existing_token=...) vs the backend_config pattern used by TrainerClient, SparkClient, and OptimizerClient
- ID-centric API — kfp.Client requires pipeline IDs and experiment IDs before triggering a run. Other SDK clients are name-first
- Too many methods for simple tasks — uploading a pipeline requires choosing between 4 methods depending on source type and whether it's new or a version. Running a pipeline requires choosing between 3 methods
- Scattered wait semantics — kfp.Client.wait_for_run_completion hardcodes terminal states and cannot wait for intermediate states like Running

Users work directly with kfp.Client, which requires:

- Separate pip install kfp and different import patterns from the rest of the Kubeflow SDK
- Looking up pipeline/experiment IDs before running pipelines (ID-centric API)
- Choosing between multiple upload/run methods depending on source type
- Limited wait functionality (only terminal states, no callbacks)

PipelinesClient addresses all of these while keeping kfp.Client fully supported as an escape hatch for advanced use cases.

---

Love this idea? Give it a 👍.

MStokluska:

## Summary

Implements Phase 1 of the `PipelinesClient` ([KEP-125](https://github.com/kubeflow/sdk/tree/main/docs/proposals/125-kfp-client)) — a simplified, name-first Python client for Kubeflow Pipelines built directly on `kfp_server_api`.

- Adds `PipelinesClient` class with `PipelinesBackendConfig` dataclass for connection configuration
- Pipeline CRUD: `upload_pipeline`, `get_pipeline`, `get_pipeline_version`, `list_pipelines`, `list_pipeline_versions`, `delete_pipeline`
- Run management: `run` (callable, YAML path, name, Pipeline/PipelineVersion objects), `get_run`, `list_runs`, `wait_for_run_status` (flexible target states, callbacks, token refresh)
- Experiment management: `create_experiment` (idempotent), `get_experiment`, `list_experiments`, `delete_experiment`
- Host auto-discovery (env var, in-cluster DNS, kubeconfig proxy)
- Bearer token auth with refresh and retry
- Escape hatch to `kfp.Client` via `kfp_client` property
- Unit tests covering all public and critical private methods

## Test plan

- [x] Unit tests pass locally: `pytest -v sdk/python/kfp/kubeflow_client/pipelines_client_test.py`
- [x] Manual verification against a live DSPA endpoint (in-cluster pod)

Issue: #13380

**Checklist:**

### DIFF
--- a/sdk/python/kfp/kubeflow_client/api/pipelines_client.py
+++ b/sdk/python/kfp/kubeflow_client/api/pipelines_client.py
@@ -20,17 +20,11 @@ designed for re-export by the Kubeflow SDK at ``kubeflow.pipelines``.
 from __future__ import annotations
 
 from collections.abc import Callable
-import datetime
-import logging
-import os
-import shutil
-import tempfile
 from typing import Any, TYPE_CHECKING
 
 if TYPE_CHECKING:
     from kfp.client import Client
 
-from kfp import compiler
 from kfp.kubeflow_client.backends.kubernetes import KubernetesBackend
 from kfp.kubeflow_client.backends.kubernetes import KubernetesBackendConfig
 from kfp.kubeflow_client.types import Experiment
@@ -41,13 +35,8 @@ from kfp.kubeflow_client.types import ListRunsResponse
 from kfp.kubeflow_client.types import Pipeline
 from kfp.kubeflow_client.types import PipelineVersion
 from kfp.kubeflow_client.types import Run
-import yaml
-
-logger = logging.getLogger(__name__)
 
 __all__ = ['PipelinesClient']
-
-_VALID_UPLOAD_EXTENSIONS = ('.yaml', '.yml', '.tar.gz', '.tgz', '.zip')
 
 
 class PipelinesClient:
@@ -110,22 +99,8 @@ class PipelinesClient:
         Returns:
             A ``PipelineVersion`` object representing the uploaded version.
         """
-        package_path, temp_dir = self._resolve_pipeline_to_file(pipeline)
-
-        try:
-            if name is None:
-                name = self._infer_pipeline_name(pipeline, package_path)
-            self._validate_pipeline_name(name)
-
-            return self._backend.upload_pipeline(
-                package_path,
-                name=name,
-                version_name=version,
-                description=description,
-            )
-        finally:
-            if temp_dir is not None:
-                shutil.rmtree(temp_dir, ignore_errors=True)
+        return self._backend.upload_pipeline(
+            pipeline, name=name, version_name=version, description=description)
 
     def get_pipeline(self, name: str) -> Pipeline:
         """Get a pipeline by name.
@@ -266,70 +241,13 @@ class PipelinesClient:
         Returns:
             A ``Run`` object.
         """
-        run_name = name or self._generate_run_name(pipeline)
-
-        version_is_usable = (
-            isinstance(pipeline, Pipeline) or
-            (isinstance(pipeline, str) and not self._is_yaml_path(pipeline) and
-             not self._is_archive_path(pipeline)))
-        if version is not None and not version_is_usable:
-            logger.warning(
-                'The version parameter is ignored when pipeline is not a '
-                'name string (got %s).',
-                type(pipeline).__name__
-                if not isinstance(pipeline, str) else repr(pipeline))
-
-        if isinstance(pipeline, PipelineVersion):
-            return self._backend.run_from_version(
-                pipeline_id=pipeline.pipeline_id,
-                version_id=pipeline.pipeline_version_id,
-                params=params,
-                run_name=run_name,
-                experiment=experiment,
-            )
-
-        if isinstance(pipeline, Pipeline):
-            return self._backend.run_pipeline(
-                pipeline=pipeline,
-                version=version,
-                params=params,
-                run_name=run_name,
-                experiment=experiment,
-            )
-
-        if callable(pipeline):
-            return self._run_inline(
-                pipeline_callable=pipeline,
-                params=params,
-                run_name=run_name,
-                experiment=experiment,
-            )
-
-        if isinstance(pipeline, str) and self._is_yaml_path(pipeline):
-            return self._backend.run_from_file(
-                file_path=pipeline,
-                params=params,
-                run_name=run_name,
-                experiment=experiment,
-            )
-
-        if isinstance(pipeline, str) and self._is_archive_path(pipeline):
-            raise ValueError(
-                f'Archive files ({pipeline!r}) cannot be used for inline '
-                'runs. Use upload_pipeline() first, then run by name.')
-
-        if isinstance(pipeline, str):
-            return self._backend.run_by_name(
-                pipeline_name=pipeline,
-                version_name=version,
-                params=params,
-                run_name=run_name,
-                experiment=experiment,
-            )
-
-        raise ValueError(f'Unsupported pipeline type: {type(pipeline)!r}. '
-                         'Expected a pipeline name, file path, callable, '
-                         'Pipeline, or PipelineVersion.')
+        return self._backend.run(
+            pipeline,
+            params=params,
+            name=name,
+            experiment=experiment,
+            version=version,
+        )
 
     def get_run(self, run_id: str) -> Run:
         """Get a run by ID.
@@ -511,170 +429,3 @@ class PipelinesClient:
                 kwargs['verify_ssl'] = config.is_secure
             self._kfp_client_instance = Client(**kwargs)
         return self._kfp_client_instance
-
-    # ------------------------------------------------------------------
-    # Private helpers — SDK-level preprocessing
-    # ------------------------------------------------------------------
-
-    def _resolve_pipeline_to_file(
-        self,
-        pipeline: Callable | str,
-    ) -> tuple[str, str | None]:
-        """Resolve a pipeline source to a compiled YAML file path.
-
-        If ``pipeline`` is callable, compile it to a temporary directory.
-        If it's a string, treat it as a file path and validate existence.
-
-        Returns:
-            A tuple of (package_path, temp_dir). ``temp_dir`` is ``None``
-            when the input is a user-provided file path (no cleanup needed).
-        """
-        if callable(pipeline):
-            temp_dir = tempfile.mkdtemp()
-            package_path = os.path.join(temp_dir, 'pipeline.yaml')
-            try:
-                compiler.Compiler().compile(
-                    pipeline_func=pipeline,
-                    package_path=package_path,
-                )
-            except Exception as error:
-                shutil.rmtree(temp_dir, ignore_errors=True)
-                raise ValueError(
-                    f'Failed to compile pipeline: {error}') from error
-            return package_path, temp_dir
-        if isinstance(pipeline, str):
-            if not os.path.isfile(pipeline):
-                raise ValueError(f'Pipeline file not found: {pipeline!r}.')
-            if not any(
-                    pipeline.endswith(ext) for ext in _VALID_UPLOAD_EXTENSIONS):
-                raise ValueError(f'Unsupported file type: {pipeline!r}. '
-                                 f'Expected one of: '
-                                 f'{", ".join(_VALID_UPLOAD_EXTENSIONS)}')
-            return pipeline, None
-        raise ValueError(
-            f'Expected a callable or file path, got {type(pipeline)!r}.')
-
-    def _infer_pipeline_name(
-        self,
-        pipeline: Callable | str,
-        package_path: str,
-    ) -> str:
-        """Infer a pipeline name from the source.
-
-        Resolution order: callable's .name attribute, callable's
-        __name__, pipeline name from compiled YAML, filename stem, or
-        'pipeline'.
-        """
-        if callable(pipeline) and hasattr(pipeline, 'name') and pipeline.name:
-            return pipeline.name
-        if callable(pipeline) and hasattr(pipeline,
-                                          '__name__') and pipeline.__name__:
-            return pipeline.__name__.replace('_', '-')
-
-        name_from_spec = self._read_pipeline_name_from_yaml(package_path)
-        if name_from_spec:
-            return name_from_spec
-
-        if isinstance(pipeline, str):
-            basename = os.path.basename(pipeline)
-            return self._strip_pipeline_extension(basename)
-        return 'pipeline'
-
-    @staticmethod
-    def _validate_pipeline_name(name: str) -> None:
-        """Validate that a pipeline name is non-empty."""
-        if not name or name.isspace():
-            raise ValueError(
-                'Invalid pipeline name. Pipeline name cannot be empty '
-                'or contain only whitespace.')
-
-    @staticmethod
-    def _read_pipeline_name_from_yaml(package_path: str,) -> str | None:
-        """Try to extract the pipeline name from a compiled YAML file."""
-        try:
-            if not package_path.endswith(('.yaml', '.yml')):
-                return None
-            with open(package_path, 'r') as f:
-                doc = yaml.safe_load(f)
-            if isinstance(doc, dict):
-                pipeline_info = doc.get('pipelineInfo', {})
-                name = pipeline_info.get('name')
-                if name and isinstance(name, str) and not name.isspace():
-                    return name
-        except (OSError, yaml.YAMLError):
-            logger.debug(
-                'Could not read pipeline name from %s.',
-                package_path,
-                exc_info=True)
-        return None
-
-    # ------------------------------------------------------------------
-    # Private helpers — run creation
-    # ------------------------------------------------------------------
-
-    def _run_inline(
-        self,
-        pipeline_callable: Callable,
-        params: dict[str, Any] | None,
-        run_name: str,
-        experiment: str | None,
-    ) -> Run:
-        """Compile a callable and submit inline (no upload)."""
-        package_path, temp_dir = self._resolve_pipeline_to_file(
-            pipeline_callable)
-        try:
-            return self._backend.run_from_file(
-                file_path=package_path,
-                params=params,
-                run_name=run_name,
-                experiment=experiment,
-            )
-        finally:
-            if temp_dir is not None:
-                shutil.rmtree(temp_dir, ignore_errors=True)
-
-    # ------------------------------------------------------------------
-    # Private helpers — utilities
-    # ------------------------------------------------------------------
-
-    @staticmethod
-    def _is_yaml_path(value: str) -> bool:
-        """Check if a string looks like a YAML pipeline file path."""
-        return value.endswith('.yaml') or value.endswith('.yml')
-
-    @staticmethod
-    def _is_archive_path(value: str) -> bool:
-        """Check if a string looks like an archive pipeline file path."""
-        return (value.endswith('.tar.gz') or value.endswith('.tgz') or
-                value.endswith('.zip'))
-
-    @staticmethod
-    def _strip_pipeline_extension(filename: str) -> str:
-        """Strip known pipeline file extensions including compound ones."""
-        for suffix in ('.tar.gz', '.tgz', '.zip', '.yaml', '.yml'):
-            if filename.endswith(suffix):
-                return filename[:-len(suffix)]
-        return filename
-
-    @staticmethod
-    def _generate_run_name(
-        pipeline: str | Callable | Pipeline | PipelineVersion,) -> str:
-        """Generate a default run display name."""
-        timestamp = (
-            datetime.datetime.now().astimezone().strftime('%Y-%m-%d %H-%M-%S'))
-        if callable(pipeline):
-            base_name = getattr(pipeline, 'name', None)
-            if base_name is None:
-                base_name = getattr(pipeline, '__name__', 'pipeline')
-            return f'{base_name} {timestamp}'
-        if isinstance(pipeline, str):
-            if (PipelinesClient._is_yaml_path(pipeline) or
-                    PipelinesClient._is_archive_path(pipeline)):
-                basename = os.path.basename(pipeline)
-                name = PipelinesClient._strip_pipeline_extension(basename)
-                return f'{name} {timestamp}'
-            return f'{pipeline} {timestamp}'
-        if isinstance(pipeline, (Pipeline, PipelineVersion)):
-            display_name = getattr(pipeline, 'display_name', 'pipeline')
-            return f'{display_name} {timestamp}'
-        return f'pipeline {timestamp}'
--- a/sdk/python/kfp/kubeflow_client/api/pipelines_client.py
+++ b/sdk/python/kfp/kubeflow_client/api/pipelines_client.py
@@ -27,12 +27,10 @@ import shutil
 import tempfile
 import time
 from typing import Any, Callable, TYPE_CHECKING
-import warnings
 
 if TYPE_CHECKING:
     from kfp.client import Client
 
-from google.protobuf import json_format
 from kfp import compiler
 from kfp.kubeflow_client import constants
 from kfp.kubeflow_client.backends.kubernetes import KubernetesBackend
@@ -45,7 +43,6 @@ from kfp.kubeflow_client.types import ListRunsResponse
 from kfp.kubeflow_client.types import Pipeline
 from kfp.kubeflow_client.types import PipelineVersion
 from kfp.kubeflow_client.types import Run
-from kfp.pipeline_spec import pipeline_spec_pb2
 import kfp_server_api
 import yaml
 
@@ -123,17 +120,17 @@ class PipelinesClient:
                 name = self._infer_pipeline_name(pipeline, package_path)
             self._validate_pipeline_name(name)
 
-            existing_pipeline_id = self._get_pipeline_id_by_name(name)
+            existing_pipeline_id = self._backend._get_pipeline_id_by_name(name)
 
             if existing_pipeline_id is not None:
-                return self._upload_version(
+                return self._backend._upload_version(
                     package_path,
                     pipeline_id=existing_pipeline_id,
                     version_name=version,
                     description=description,
                 )
             else:
-                return self._upload_new_pipeline(
+                return self._backend._upload_new_pipeline(
                     package_path,
                     name=name,
                     version_name=version,
@@ -155,7 +152,7 @@ class PipelinesClient:
         Raises:
             ValueError: If no pipeline matches or multiple pipelines match.
         """
-        pipeline_id = self._get_pipeline_id_by_name(name)
+        pipeline_id = self._backend._get_pipeline_id_by_name(name)
         if pipeline_id is None:
             raise ValueError(f'Pipeline not found: {name!r}. '
                              'Use list_pipelines() to see available pipelines.')
@@ -180,7 +177,8 @@ class PipelinesClient:
             ValueError: If the pipeline or version is not found.
         """
         pipeline = self.get_pipeline(name)
-        version_id = self._get_version_id_by_name(pipeline.pipeline_id, version)
+        version_id = self._backend._get_version_id_by_name(
+            pipeline.pipeline_id, version)
         if version_id is None:
             raise ValueError(f'Pipeline version not found: {version!r} '
                              f'for pipeline {name!r}.')
@@ -229,7 +227,7 @@ class PipelinesClient:
             A ``ListPipelineVersionsResponse`` with ``.pipeline_versions``
             and ``.next_page_token``.
         """
-        pipeline_id = self._get_pipeline_id_by_name(name)
+        pipeline_id = self._backend._get_pipeline_id_by_name(name)
         if pipeline_id is None:
             raise ValueError(f'Pipeline not found: {name!r}. '
                              'Use list_pipelines() to see available pipelines.')
@@ -259,13 +257,14 @@ class PipelinesClient:
             ValueError: If the pipeline has multiple versions and
                 ``force=False``.
         """
-        pipeline_id = self._get_pipeline_id_by_name(name)
+        pipeline_id = self._backend._get_pipeline_id_by_name(name)
         if pipeline_id is None:
             raise ValueError(f'Pipeline not found: {name!r}. '
                              'Use list_pipelines() to see available pipelines.')
 
         if version is not None:
-            version_id = self._get_version_id_by_name(pipeline_id, version)
+            version_id = self._backend._get_version_id_by_name(
+                pipeline_id, version)
             if version_id is None:
                 raise ValueError(f'Pipeline version not found: {version!r} '
                                  f'for pipeline {name!r}.')
@@ -336,7 +335,7 @@ class PipelinesClient:
         Returns:
             A ``Run`` object.
         """
-        experiment_id = self._resolve_experiment_id(experiment)
+        experiment_id = self._backend._resolve_experiment_id(experiment)
         run_name = name or self._generate_run_name(pipeline)
 
         version_is_usable = (
@@ -351,7 +350,7 @@ class PipelinesClient:
                 if not isinstance(pipeline, str) else repr(pipeline))
 
         if isinstance(pipeline, PipelineVersion):
-            return self._run_from_version_reference(
+            return self._backend._run_from_version_reference(
                 pipeline_id=pipeline.pipeline_id,
                 version_id=pipeline.pipeline_version_id,
                 params=params,
@@ -361,14 +360,15 @@ class PipelinesClient:
 
         if isinstance(pipeline, Pipeline):
             if version:
-                version_id = self._get_version_id_by_name(
+                version_id = self._backend._get_version_id_by_name(
                     pipeline.pipeline_id, version)
                 if version_id is None:
                     raise ValueError(f'Pipeline version not found: {version!r} '
                                      f'for pipeline {pipeline.display_name!r}.')
             else:
-                version_id = self._get_latest_version_id(pipeline.pipeline_id)
-            return self._run_from_version_reference(
+                version_id = self._backend._get_latest_version_id(
+                    pipeline.pipeline_id)
+            return self._backend._run_from_version_reference(
                 pipeline_id=pipeline.pipeline_id,
                 version_id=version_id,
                 params=params,
@@ -387,7 +387,7 @@ class PipelinesClient:
         # Dispatch YAML and archive separately because YAML triggers inline
         # run while archives must be rejected with guidance.
         if isinstance(pipeline, str) and self._is_yaml_path(pipeline):
-            return self._run_from_file(
+            return self._backend._run_from_file(
                 file_path=pipeline,
                 params=params,
                 run_name=run_name,
@@ -400,7 +400,7 @@ class PipelinesClient:
                 'runs. Use upload_pipeline() first, then run by name.')
 
         if isinstance(pipeline, str):
-            return self._run_by_name(
+            return self._backend._run_by_name(
                 pipeline_name=pipeline,
                 version_name=version,
                 params=params,
@@ -455,13 +455,13 @@ class PipelinesClient:
         """
         pipeline_id = None
         if pipeline is not None:
-            pipeline_id = self._get_pipeline_id_by_name(pipeline)
+            pipeline_id = self._backend._get_pipeline_id_by_name(pipeline)
             if pipeline_id is None:
                 raise ValueError(f'Pipeline not found: {pipeline!r}.')
 
         experiment_id = None
         if experiment is not None:
-            experiment_id = self._get_experiment_id_by_name(experiment)
+            experiment_id = self._backend._get_experiment_id_by_name(experiment)
             if experiment_id is None:
                 raise ValueError(f'Experiment not found: {experiment!r}. '
                                  'Use create_experiment() first.')
@@ -507,7 +507,7 @@ class PipelinesClient:
         run: str | Run,
         *,
         status: set[str] | None = None,
-        timeout: int | None = None,
+        timeout: int = 600,
         polling_interval: int = 5,
         callbacks: list[Callable[[Run], None]] | None = None,
     ) -> Run:
@@ -519,7 +519,7 @@ class PipelinesClient:
                 ``{constants.RUN_COMPLETE}`` (``"succeeded"``). The wait
                 always exits immediately on any terminal state regardless
                 of this parameter.
-            timeout: Maximum seconds to wait. If ``None``, wait indefinitely.
+            timeout: Maximum seconds to wait. Defaults to 600 (10 minutes).
             polling_interval: Seconds between status checks.
             callbacks: Called with the final ``Run`` object when the wait
                 ends (on any stop condition).
@@ -603,7 +603,7 @@ class PipelinesClient:
         Returns:
             An ``Experiment`` object.
         """
-        existing_id = self._get_experiment_id_by_name(name)
+        existing_id = self._backend._get_experiment_id_by_name(name)
         if existing_id is not None:
             if description:
                 logger.warning(
@@ -632,7 +632,7 @@ class PipelinesClient:
         Raises:
             ValueError: If no experiment with that name is found.
         """
-        experiment_id = self._get_experiment_id_by_name(name)
+        experiment_id = self._backend._get_experiment_id_by_name(name)
         if experiment_id is None:
             raise ValueError(f'Experiment not found: {name!r}. '
                              'Use create_experiment() to create one.')
@@ -670,7 +670,7 @@ class PipelinesClient:
         Raises:
             ValueError: If no experiment with that name is found.
         """
-        experiment_id = self._get_experiment_id_by_name(name)
+        experiment_id = self._backend._get_experiment_id_by_name(name)
         if experiment_id is None:
             raise ValueError(f'Experiment not found: {name!r}.')
         self._backend.experiment_api.experiment_service_delete_experiment(
@@ -703,82 +703,6 @@ class PipelinesClient:
                 kwargs['verify_ssl'] = config.is_secure
             self._kfp_client_instance = Client(**kwargs)
         return self._kfp_client_instance
-
-    # ------------------------------------------------------------------
-    # Private helpers — name resolution
-    # ------------------------------------------------------------------
-
-    @staticmethod
-    def _equals_filter(key: str, value: str) -> str:
-        """Build a JSON filter string for an equality predicate."""
-        return json.dumps({
-            'predicates': [{
-                'operation': 'EQUALS',
-                'key': key,
-                'stringValue': value,
-            }]
-        })
-
-    def _get_pipeline_id_by_name(self, name: str) -> str | None:
-        """Resolve a pipeline display name to its ID."""
-        result = self._backend.pipelines_api.pipeline_service_list_pipelines(
-            namespace=self._backend.namespace,
-            filter=self._equals_filter('display_name', name))
-        pipelines = result.pipelines or []
-        if len(pipelines) == 0:
-            return None
-        if len(pipelines) == 1:
-            return pipelines[0].pipeline_id
-        pipeline_ids = [p.pipeline_id for p in pipelines]
-        raise ValueError(
-            f'Multiple pipelines found with name {name!r}: {pipeline_ids}. '
-            'Use kfp.Client directly to operate by pipeline ID.')
-
-    def _get_version_id_by_name(
-        self,
-        pipeline_id: str,
-        version_name: str,
-    ) -> str | None:
-        """Resolve a version display name to its ID within a pipeline."""
-        result = self._backend.pipelines_api.pipeline_service_list_pipeline_versions(
-            pipeline_id=pipeline_id,
-            filter=self._equals_filter('display_name', version_name),
-        )
-        versions = result.pipeline_versions or []
-        if len(versions) == 0:
-            return None
-        if len(versions) == 1:
-            return versions[0].pipeline_version_id
-        version_ids = [v.pipeline_version_id for v in versions]
-        raise ValueError(f'Multiple versions found with name {version_name!r}: '
-                         f'{version_ids}.')
-
-    def _get_latest_version_id(self, pipeline_id: str) -> str:
-        """Get the latest (most recently created) version ID for a pipeline."""
-        result = self._backend.pipelines_api.pipeline_service_list_pipeline_versions(
-            pipeline_id=pipeline_id,
-            page_size=1,
-            sort_by='created_at desc',
-        )
-        versions = result.pipeline_versions or []
-        if not versions:
-            raise ValueError(f'Pipeline {pipeline_id!r} has no versions.')
-        return versions[0].pipeline_version_id
-
-    def _get_experiment_id_by_name(self, name: str) -> str | None:
-        """Resolve an experiment display name to its ID."""
-        result = self._backend.experiment_api.experiment_service_list_experiments(
-            namespace=self._backend.namespace,
-            filter=self._equals_filter('display_name', name),
-        )
-        experiments = result.experiments or []
-        if len(experiments) == 0:
-            return None
-        if len(experiments) == 1:
-            return experiments[0].experiment_id
-        experiment_ids = [e.experiment_id for e in experiments]
-        raise ValueError(f'Multiple experiments found with name {name!r}: '
-                         f'{experiment_ids}.')
 
     # ------------------------------------------------------------------
     # Private helpers — pipeline compilation and upload
@@ -876,162 +800,9 @@ class PipelinesClient:
                 exc_info=True)
         return None
 
-    def _upload_new_pipeline(
-        self,
-        package_path: str,
-        *,
-        name: str,
-        version_name: str | None,
-        description: str | None,
-    ) -> PipelineVersion:
-        """Upload a brand-new pipeline and return its first version."""
-        upload_kwargs: dict[str, Any] = {
-            'name': name,
-            'namespace': self._backend.namespace,
-        }
-        if description:
-            upload_kwargs['description'] = description
-
-        pipeline_response = self._backend.upload_api.upload_pipeline(
-            package_path, **upload_kwargs)
-
-        versions_response = (
-            self._backend.pipelines_api.pipeline_service_list_pipeline_versions(
-                pipeline_id=pipeline_response.pipeline_id,
-                page_size=1,
-                sort_by='created_at desc',
-            ))
-        versions = versions_response.pipeline_versions or []
-        if not versions:
-            raise RuntimeError(
-                f'Pipeline {name!r} was uploaded but no version was '
-                'created by the server. This is unexpected.')
-
-        first_version = versions[0]
-
-        if version_name and first_version.display_name != version_name:
-            try:
-                update_body = kfp_server_api.V2beta1PipelineVersion(
-                    pipeline_id=first_version.pipeline_id,
-                    pipeline_version_id=first_version.pipeline_version_id,
-                    display_name=version_name,
-                )
-                self._backend.pipelines_api.pipeline_service_update_pipeline_version(
-                    pipeline_version_pipeline_id=first_version.pipeline_id,
-                    pipeline_version_pipeline_version_id=(
-                        first_version.pipeline_version_id),
-                    pipeline_version=update_body,
-                )
-                first_version.display_name = version_name
-            except kfp_server_api.ApiException:
-                warnings.warn(
-                    f'Could not rename the first pipeline version to '
-                    f'{version_name!r}. The upload API does not accept a '
-                    f'version name for the initial version; subsequent '
-                    f'versions will use the provided name.',
-                    stacklevel=2,
-                )
-
-        return first_version
-
-    def _upload_version(
-        self,
-        package_path: str,
-        *,
-        pipeline_id: str,
-        version_name: str | None,
-        description: str | None,
-    ) -> PipelineVersion:
-        """Upload a new version to an existing pipeline."""
-        if not version_name:
-            version_name = (
-                datetime.datetime.now().astimezone().strftime(
-                    '%Y-%m-%d %H-%M-%S'))
-        upload_kwargs: dict[str, Any] = {
-            'pipelineid': pipeline_id,
-            'name': version_name,
-        }
-        if description:
-            upload_kwargs['description'] = description
-
-        return self._backend.upload_api.upload_pipeline_version(
-            package_path, **upload_kwargs)
-
     # ------------------------------------------------------------------
     # Private helpers — run creation
     # ------------------------------------------------------------------
-
-    def _resolve_experiment_id(
-        self,
-        experiment: str | None,
-    ) -> str | None:
-        """Resolve an experiment name to its ID.
-
-        Returns None when experiment is None, letting the server use its
-        default experiment.
-        """
-        if experiment is None:
-            return None
-        experiment_id = self._get_experiment_id_by_name(experiment)
-        if experiment_id is None:
-            raise ValueError(f'Experiment not found: {experiment!r}. '
-                             'Use create_experiment() first.')
-        return experiment_id
-
-    def _run_by_name(
-        self,
-        pipeline_name: str,
-        version_name: str | None,
-        params: dict[str, Any] | None,
-        run_name: str,
-        experiment_id: str | None,
-    ) -> Run:
-        """Run an uploaded pipeline by name."""
-        pipeline_id = self._get_pipeline_id_by_name(pipeline_name)
-        if pipeline_id is None:
-            raise ValueError(f'Pipeline not found: {pipeline_name!r}. '
-                             'Upload it first with upload_pipeline().')
-
-        if version_name:
-            version_id = self._get_version_id_by_name(pipeline_id, version_name)
-            if version_id is None:
-                raise ValueError(
-                    f'Pipeline version not found: {version_name!r} '
-                    f'for pipeline {pipeline_name!r}.')
-        else:
-            version_id = self._get_latest_version_id(pipeline_id)
-
-        return self._run_from_version_reference(
-            pipeline_id=pipeline_id,
-            version_id=version_id,
-            params=params,
-            run_name=run_name,
-            experiment_id=experiment_id,
-        )
-
-    def _run_from_version_reference(
-        self,
-        pipeline_id: str,
-        version_id: str,
-        params: dict[str, Any] | None,
-        run_name: str,
-        experiment_id: str | None,
-    ) -> Run:
-        """Create a run from a pipeline version reference (ID-based)."""
-        runtime_config = kfp_server_api.V2beta1RuntimeConfig(
-            parameters=params or {},)
-        pipeline_version_reference = (
-            kfp_server_api.V2beta1PipelineVersionReference(
-                pipeline_id=pipeline_id,
-                pipeline_version_id=version_id,
-            ))
-        run_body = kfp_server_api.V2beta1Run(
-            display_name=run_name,
-            experiment_id=experiment_id,
-            pipeline_version_reference=pipeline_version_reference,
-            runtime_config=runtime_config,
-        )
-        return self._backend.run_api.run_service_create_run(run=run_body)
 
     def _run_inline(
         self,
@@ -1044,7 +815,7 @@ class PipelinesClient:
         package_path, temp_dir = self._resolve_pipeline_to_file(
             pipeline_callable)
         try:
-            return self._run_from_file(
+            return self._backend._run_from_file(
                 file_path=package_path,
                 params=params,
                 run_name=run_name,
@@ -1053,65 +824,6 @@ class PipelinesClient:
         finally:
             if temp_dir is not None:
                 shutil.rmtree(temp_dir, ignore_errors=True)
-
-    def _run_from_file(
-        self,
-        file_path: str,
-        params: dict[str, Any] | None,
-        run_name: str,
-        experiment_id: str | None,
-    ) -> Run:
-        """Submit a run from a compiled pipeline YAML file."""
-        if not os.path.isfile(file_path):
-            raise ValueError(f'Pipeline file not found: {file_path}')
-        pipeline_spec_dict = self._load_pipeline_spec(file_path)
-        runtime_config = kfp_server_api.V2beta1RuntimeConfig(
-            parameters=params or {},)
-        run_body = kfp_server_api.V2beta1Run(
-            display_name=run_name,
-            experiment_id=experiment_id,
-            pipeline_spec=pipeline_spec_dict,
-            runtime_config=runtime_config,
-        )
-        return self._backend.run_api.run_service_create_run(run=run_body)
-
-    def _load_pipeline_spec(self, file_path: str) -> dict:
-        """Load a pipeline spec from a YAML file and return as dict."""
-        with open(file_path, 'r') as f:
-            try:
-                docs = list(yaml.safe_load_all(f))
-            except yaml.YAMLError as e:
-                raise ValueError(
-                    f'Failed to parse pipeline YAML at {file_path!r}: {e}'
-                ) from e
-
-        if not docs or not docs[0]:
-            raise ValueError(
-                f'Pipeline file is empty or contains no valid YAML '
-                f'documents: {file_path}')
-        pipeline_spec_dict = docs[0]
-        platform_spec_dict = {}
-        for doc in docs[1:]:
-            if doc:
-                platform_spec_dict.update(doc)
-
-        pipeline_spec = json_format.ParseDict(
-            pipeline_spec_dict,
-            pipeline_spec_pb2.PipelineSpec(),
-            ignore_unknown_fields=True)
-        platform_spec = json_format.ParseDict(
-            platform_spec_dict,
-            pipeline_spec_pb2.PlatformSpec(),
-            ignore_unknown_fields=True)
-
-        # The API expects different shapes depending on whether a platform spec
-        # is present: bare pipeline spec dict vs. wrapped {pipeline_spec, platform_spec}.
-        if platform_spec == pipeline_spec_pb2.PlatformSpec():
-            return json_format.MessageToDict(pipeline_spec)
-        return {
-            'pipeline_spec': json_format.MessageToDict(pipeline_spec),
-            'platform_spec': json_format.MessageToDict(platform_spec),
-        }
 
     # ------------------------------------------------------------------
     # Private helpers — utilities
--- a/sdk/python/kfp/kubeflow_client/api/pipelines_client.py
+++ b/sdk/python/kfp/kubeflow_client/api/pipelines_client.py
@@ -1,4 +1,4 @@
-# Copyright 2026 The Kubeflow Authors
+# Copyright The Kubeflow Authors
 #
 # Licensed under the Apache License, Version 2.0 (the "License");
 # you may not use this file except in compliance with the License.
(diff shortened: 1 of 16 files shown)


sdk/python/kfp/kubeflow_client/api/pipelines_client.py
```diff
@@ -0,0 +1,1408 @@
...
+
+        Returns:
+            A ``Run`` object.
+        """
+        return self._run_api.run_service_get_run(run_id=run_id)
+
+    def list_runs(
+        self,
+        *,
+        pipeline: str | None = None,
```
MStokluska, line 266: Regarding the pipeline filter in `list_runs`:

The pipeline parameter is included because KEP-125 specifies it as part of the Phase 1 API surface. However, it's worth noting that this filter is applied client-side - the KFP v2beta1 ListRuns API does not currently support server-side filtering by pipeline ID.

Why it's client-side: The backend stores PipelineId as a column in the run_details table (via the embedded PipelineSpec struct in backend/src/apiserver/model/pipeline_spec.go), but this field is not exposed in the runAPIToModelFieldMap (backend/src/apiserver/model/run.go), which defines the set of valid filter keys accepted by the filter query parameter.

Caveats of the current approach:

The returned page may contain fewer items than page_size (including zero items with a non-empty next_page_token) because filtering happens after the server returns a page.
Callers must continue paginating until next_page_token is empty to get complete results which IMO isn't great UX.
How this can be fixed upstream: enable server-side filtering via the existing filter protocol buffer mechanism, eliminating the pagination inconsistency. Once that backend change lands, this client method can switch to server-side filtering transparently (no public API change needed).

sdk/python/kfp/kubeflow_client/pipelines_client.py
```diff
@@ -0,0 +1,1408 @@
...
+        if isinstance(pipeline, str):
+            if not os.path.isfile(pipeline):
+                raise ValueError(
+                    f'Pipeline file not found: {pipeline!r}.')
+            if not any(pipeline.endswith(ext)
+                       for ext in _VALID_UPLOAD_EXTENSIONS):
+                raise ValueError(
+                    f'Unsupported file type: {pipeline!r}. '
+                    f'Expected one of: '
+                    f'{", ".join(_VALID_UPLOAD_EXTENSIONS)}')
```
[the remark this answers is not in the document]
MStokluska: Good point - I'll add targeted tests for file-not-found, unsupported extension, and compilation failure + cleanup in a follow-up commit.

sdk/python/kfp/kubeflow_client/pipelines_client.py
```diff
@@ -0,0 +1,1408 @@
...
+        if not token_path:
+            if os.path.exists(_kfp_sa_token_path):
+                token_path = _kfp_sa_token_path
+            elif os.path.exists(_k8s_sa_token_path):
+                token_path = _k8s_sa_token_path
+                os.environ[_token_path_env] = token_path
+                logger.debug(
+                    'Kubeflow pipelines token path missing; using Kubernetes '
+                    'service account token at %s for API auth.',
+                    token_path)
```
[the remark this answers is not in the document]
MStokluska: I'll refactor to pass path=token_path directly - this keeps the same behavior without the process-wide side effect in follow up commit

andreyvelich:
Thanks for this work @MStokluska! Will check this out soon

/assign @andreyvelich

MStokluska:
hey @andreyvelich @droctothorpe @mprahl - I would appreciate a review when time allows, thanks!

sdk/python/kfp/kubeflow_client/constants.py
```diff
@@ -0,0 +1,48 @@
+# Copyright 2026 The Kubeflow Authors
```
andreyvelich: Can we make directory consistent?
That will help us to extend for other backends in the future easier. cc @kramaranya @Fiona-Waters @szaher 
```
kubeflow_client/api/pipelines_client.py
kubeflow_client/backends/kubernetes/backend.py
kubeflow_client/backends/kubernetes/types.py 
kubeflow_client/backends/kubernetes/constants.py 
```

sdk/python/kfp/kubeflow_client/pipelines_client.py
```diff
@@ -0,0 +1,1377 @@
+# Copyright 2026 The Kubeflow Authors
```
andreyvelich: @tariq-hasan recently introduce change to remove year from the licence to follow k8s best practice, it would be nice if we follow the same in KFP: https://github.com/kubeflow/trainer/pull/3359
cc @droctothorpe @HumairAK @mprahl



@nsingla We can even move this script to a common repo, so all subprojects are aligned.

sdk/python/kfp/kubeflow_client/pipelines_client.py
```diff
@@ -0,0 +1,1377 @@
...
+        namespace: Kubernetes namespace. If omitted, auto-detected.
+    """
+
+    base_url: str | None = None
+    user_token: str | None = None
+    is_secure: bool | None = None
+    custom_ca: str | None = None
+    namespace: str | None = None
+
+    def __repr__(self) -> str:
```
andreyvelich: Why do you need __repr__ ?

sdk/python/kfp/kubeflow_client/pipelines_client.py
```diff
@@ -0,0 +1,1377 @@
...
+
+    Args:
+        backend_config: Connection parameters for the KFP API server.
+            When ``None``, uses ``PipelinesBackendConfig()`` (zero-arg
+            construction with auto-discovery).
+    """
+
+    def __init__(
+        self,
+        backend_config: PipelinesBackendConfig | None = None,
```
andreyvelich: It should be

```suggestion
        backend_config: KubernetesBackendConfig | None = None,
```

sdk/python/kfp/kubeflow_client/pipelines_client.py
```diff
@@ -0,0 +1,1377 @@
...
+        self._namespace = backend_config.namespace
+
+        self._api_config = self._build_api_configuration(backend_config)
+        api_client = kfp_server_api.ApiClient(self._api_config)
+
+        self._pipelines_api = kfp_server_api.PipelineServiceApi(api_client)
+        self._run_api = kfp_server_api.RunServiceApi(api_client)
+        self._experiment_api = kfp_server_api.ExperimentServiceApi(api_client)
+        self._upload_api = kfp_server_api.PipelineUploadServiceApi(api_client)
+        self._kfp_client_instance = None
```
andreyvelich: Why do we need those vars exist in PipelineClient itself? Can we just extract it from:
```
PipelinesClient.backend.namespace
```

sdk/python/kfp/kubeflow_client/pipelines_client.py
```diff
@@ -0,0 +1,1377 @@
...
+        self,
+        backend_config: PipelinesBackendConfig | None = None,
+    ) -> None:
+        if backend_config is None:
+            backend_config = PipelinesBackendConfig()
+        self._config = backend_config
+        self._namespace = backend_config.namespace
+
+        self._api_config = self._build_api_configuration(backend_config)
+        api_client = kfp_server_api.ApiClient(self._api_config)
```
andreyvelich, line 70: Is `api_client` applicable only to k8s backend?

andreyvelich (commented):
Thank you for this work @MStokluska!
Left a few thoughts.

sdk/python/kfp/kubeflow_client/pipelines_client.py
```diff
@@ -0,0 +1,1377 @@
...
+        if self._namespace:
+            return self._namespace
+        try:
+            with open(_NAMESPACE_PATH, 'r') as f:
+                self._namespace = f.read().strip()
+                return self._namespace
+        except FileNotFoundError:
+            pass
+        try:
+            import kubernetes as k8s
```
andreyvelich: **Critical bug**: if the `kubernetes` package isn't installed, `import kubernetes as k8s` raises `ImportError`, but the `except (k8s.config.ConfigException, FileNotFoundError)` clause below dereferences the unbound `k8s` and raises `NameError`. The documented step-4 fallback to `_DEFAULT_NAMESPACE` is unreachable. Mirror the pattern in `_discover_host()` (lines 794–798): wrap the import in its own `try/except ImportError` first, then use `k8s` in a nested try.

sdk/python/kfp/kubeflow_client/pipelines_client.py
```diff
@@ -0,0 +1,1377 @@
...
+            return self._run_from_version_reference(
+                pipeline_id=pipeline.pipeline_id,
+                version_id=pipeline.pipeline_version_id,
+                params=params,
+                run_name=run_name,
+                experiment_id=experiment_id,
+            )
+
+        if isinstance(pipeline, Pipeline):
+            latest_version_id = self._get_latest_version_id(
```
andreyvelich: **Proposal mismatch**: KEP-125 (line 640) lists `version` as ignored only for *callable, file path, or `PipelineVersion`*. A `Pipeline` object should respect `version`, but this branch always calls `_get_latest_version_id()` and the `version_is_usable` warning at L390–395 fires incorrectly here too. Suggest: when `version` is set, resolve via `_get_version_id_by_name(pipeline.pipeline_id, version)`; include `Pipeline` in `version_is_usable`.

sdk/python/kfp/kubeflow_client/pipelines_client.py
```diff
@@ -0,0 +1,1377 @@
...
+
+    def _resolve_experiment_id(
+        self,
+        experiment: str | None,
+    ) -> str:
+        """Resolve an experiment name to its ID, using default if None."""
+        if experiment is None:
+            default_id = self._get_experiment_id_by_name('Default')
+            if default_id is None:
+                default_experiment = self.create_experiment('Default')
```
andreyvelich: **Proposal mismatch + side effect**: KEP-125 (line 639) says "the server's default experiment is used" when `experiment=None`, but here we *create* `'Default'` if missing — a write the user didn't ask for. The docstring on `run()` (L374–377) also disagrees with the proposal. Suggest either omitting `experiment_id` and letting the server resolve, or aligning the proposal + docstring with this auto-create behavior.

sdk/python/kfp/kubeflow_client/pipelines_client.py
```diff
@@ -0,0 +1,1377 @@
...
+                    page_size=2,
+                ))
+            versions = versions_response.pipeline_versions or []
+            if len(versions) > 1:
+                raise ValueError(
+                    f'Pipeline {name!r} has multiple versions. '
+                    'Use force=True to delete the pipeline and all versions, '
+                    'or specify version= to delete a single version.')
+
+        self._pipelines_api.pipeline_service_delete_pipeline(
```
andreyvelich, line 233: `cascade=True` is passed unconditionally — `force` only gates the count check, not the server cascade. A single-version `force=False` delete still cascades. Suggest `cascade=force` (or omit when `force=False`) so `force` actually controls cascade behavior.

sdk/python/kfp/kubeflow_client/pipelines_client.py
```diff
@@ -0,0 +1,1377 @@
...
+        except ImportError:
+            logger.debug(
+                'In-cluster credential module not available.', exc_info=True)
+            return
+        try:
+            credentials = ServiceAccountTokenVolumeCredentials(path=token_path)
+            credentials.refresh_api_key_hook(api_config)
+            api_config.api_key_prefix['authorization'] = 'Bearer'
+            api_config.refresh_api_key_hook = credentials.refresh_api_key_hook
+        except Exception:
```
andreyvelich: Bare `except Exception` swallows real bugs (`AttributeError`, `TypeError`) along with token failures, leaving the client unauthenticated and surfacing as an opaque 401 later. Narrow to the specific exceptions `ServiceAccountTokenVolumeCredentials.refresh_api_key_hook` actually raises (e.g. `OSError`, `kfp_server_api.ApiException`) and let programming errors propagate.

sdk/python/kfp/kubeflow_client/pipelines_client.py
```diff
@@ -0,0 +1,1377 @@
...
+            if not package_path.endswith(('.yaml', '.yml')):
+                return None
+            with open(package_path, 'r') as f:
+                doc = yaml.safe_load(f)
+            if isinstance(doc, dict):
+                pipeline_info = doc.get('pipelineInfo', {})
+                name = pipeline_info.get('name')
+                if name and isinstance(name, str) and not name.isspace():
+                    return name
+        except Exception:
```
andreyvelich: Bare `except Exception` hides `yaml.YAMLError` for malformed compiled YAML — caller silently falls back to the filename stem (or literal `'pipeline'`). Narrow to `(OSError, yaml.YAMLError)` so unexpected exceptions still propagate.

sdk/python/kfp/kubeflow_client/pipelines_client.py
```diff
@@ -0,0 +1,1377 @@
...
+                    display_name=version_name,
+                )
+                self._pipelines_api.pipeline_service_update_pipeline_version(
+                    pipeline_version_pipeline_id=first_version.pipeline_id,
+                    pipeline_version_pipeline_version_id=(
+                        first_version.pipeline_version_id),
+                    pipeline_version=update_body,
+                )
+                first_version.display_name = version_name
+            except (kfp_server_api.ApiException, AttributeError):
```
andreyvelich: When the rename fails, the returned `first_version.display_name` differs from the user's `version_name` and the caller has no signal beyond a log line. Suggest catching only `kfp_server_api.ApiException` (let `AttributeError` propagate as a real bug) and either raising or surfacing the failure via `warnings.warn` so it isn't log-only.

sdk/python/kfp/kubeflow_client/pipelines_client.py
```diff
@@ -0,0 +1,1377 @@
...
+                'Proceeding without authentication.',
+                exc_info=True)
+
+    def _refresh_credentials(self) -> None:
+        """Refresh the API token using the configured refresh hook."""
+        if hasattr(self._api_config, 'refresh_api_key_hook') and callable(
+                self._api_config.refresh_api_key_hook):
+            self._api_config.refresh_api_key_hook(self._api_config)
+        else:
+            logger.warning('Token expired but no refresh hook is configured. '
```
andreyvelich: When no `refresh_api_key_hook` is configured this just logs a warning. The `wait_for_run_status` retry loop then burns 2 retries against the same expired token before the original 401 is raised — producing two confusing "Access token expired, refreshing…" log lines. Suggest raising here so the first 401 surfaces a clear "no refresh hook configured" error immediately.

sdk/python/kfp/kubeflow_client/pipelines_client.py
```diff
@@ -0,0 +1,1377 @@
...
+
+        Args:
+            pipeline: Pipeline to run (see above).
+            params: Pipeline parameters as a dict.
+            name: Run display name. Auto-generated if omitted.
+            experiment: Experiment name. If ``None``, a ``Default`` experiment
+                is used (created automatically if it does not already exist).
+                If provided and the experiment does not exist, raises
+                ``ValueError``.
+            version: Pipeline version name (used only when ``pipeline`` is a
```
andreyvelich: Doc says `version` is "used only when pipeline is a name string" but the impl also drops it for `Pipeline` objects (always uses latest — see comment on L407). Either fix the impl to match KEP-125 or expand the doc: *"Used only when `pipeline` is a pipeline name string. Ignored for callables, file paths, `Pipeline`, and `PipelineVersion` objects."*

sdk/python/kfp/kubeflow_client/api/pipelines_client.py
```diff
@@ -0,0 +1,1377 @@
...
+            original_count = len(response.runs)
+            response.runs = [
+                run for run in response.runs
+                if (run.pipeline_version_reference and
+                    run.pipeline_version_reference.pipeline_id == pipeline_id)
+            ]
+            filtered_count = original_count - len(response.runs)
+            if filtered_count > 0:
+                logger.debug(
+                    'Client-side pipeline filter removed %d of %d runs.',
```
andreyvelich, line 409: Filtering all runs out of a page only emits `logger.debug`. From the caller's view, an empty `.runs` with a non-empty `next_page_token` looks identical to "no matches". Promote to `logger.info` so the cause is visible without enabling debug logging.

andreyvelich (commented):
Thanks for driving Phase 1 of KEP-125, @MStokluska — solid foundation. Inline comments below were generated with AI tooling assistance and reviewed before posting; please push back on anything that looks off-target.

sdk/python/kfp/kubeflow_client/api/pipelines_client.py
```diff
@@ -0,0 +1,1377 @@
...
+        )
+        return self._run_api.run_service_create_run(run=run_body)
+
+    def _load_pipeline_spec(self, file_path: str) -> dict:
+        """Load a pipeline spec from a YAML file and return as dict."""
+        with open(file_path, 'r') as f:
+            docs = list(yaml.safe_load_all(f))
+
+        if not docs or not docs[0]:
+            raise ValueError(
```
andreyvelich, line 867: Suggestion: wrap `yaml.safe_load_all` in `try/except yaml.YAMLError` and re-raise with the path: `raise ValueError(f'Failed to parse pipeline YAML at {file_path!r}: {e}') from e`. Keeps the error loud while adding the file path to the traceback.

sdk/python/kfp/kubeflow_client/pipelines_client_test.py
```diff
@@ -0,0 +1,1235 @@
...
+                    experiment_id='exp-1',
+                )
+                self.assertEqual(result.run_id, 'r-6')
+
+    def test_run_unsupported_type_raises(self):
+        with patch.object(
+                self.client, '_get_experiment_id_by_name',
+                return_value='exp-1'):
+            with self.assertRaisesRegex(ValueError, 'Unsupported pipeline'):
+                self.client.run(12345)
```
andreyvelich: Missing tests for the `version_is_usable` warning branch (pipelines_client.py L387–395). Please add: `run(Pipeline_obj, version='v')`, `run(callable, version='v')`, `run('/x.yaml', version='v')`, `run(PipelineVersion_obj, version='v')` — each should log a warning and not pass `version` downstream. Also add a positive case for `Pipeline_obj` with `version='v'` once the L407 issue is resolved.

sdk/python/kfp/kubeflow_client/pipelines_client_test.py
```diff
@@ -0,0 +1,1235 @@
...
+                    kfp_server_api.ApiException(status=401),
+                ]):
+            with patch.object(self.client, '_refresh_credentials'):
+                with self.assertRaises(kfp_server_api.ApiException) as ctx:
+                    self.client.wait_for_run_status(
+                        'r-1',
+                        status={constants.RUN_COMPLETE},
+                        polling_interval=0,
+                    )
+                self.assertEqual(ctx.exception.status, 401)
```
andreyvelich: Three contract-relevant cases for `wait_for_run_status` are not asserted: (a) first-poll `ApiException(status=401)` re-raises and does NOT call `_refresh_credentials` (the `first_poll_succeeded` security guard at L588); (b) non-401 `ApiException` propagates immediately; (c) callbacks fire on `TimeoutError` and on terminal-but-not-target state. Existing tests cover the happy paths but not these guarantees promised by the docstring.

sdk/python/kfp/kubeflow_client/api/pipelines_client_test.py
```diff
@@ -0,0 +1,1235 @@
...
+# limitations under the License.
+"""Unit tests for PipelinesClient."""
+
+import os
+import unittest
+from unittest.mock import MagicMock
+from unittest.mock import Mock
+from unittest.mock import patch
+
+from absl.testing import parameterized
```
andreyvelich: Suggestion: consider aligning the test style with the Kubeflow SDK convention (since this client is re-exported by `kubeflow.pipelines`). The SDK uses `pytest` + `pytest.mark.parametrize` over a `TestCase` dataclass, with **one `test_<public_function>` per public method** rather than nested `unittest.TestCase` classes. Reference: [`kubeflow/trainer/backends/kubernetes/backend_test.py`](https://github.com/kubeflow/sdk/blob/main/kubeflow/trainer/backends/kubernetes/backend_test.py) and the shared `TestCase` in [`kubeflow/trainer/test/common.py`](https://github.com/kubeflow/sdk/blob/main/kubeflow/trainer/test/common.py). This keeps each public method's behavior in a single parametrized test, making the contract surface easier to read and diff. Non-blocking — leaving final call to maintainers given the file lives in the kfp repo, but worth aligning before the SDK re-export ships.

sdk/python/kfp/kubeflow_client/api/pipelines_client.py
```diff
@@ -0,0 +1,1377 @@
...
+            if self._config.user_token:
+                kwargs['existing_token'] = self._config.user_token
+            if self._config.namespace:
+                kwargs['namespace'] = self._config.namespace
+            if self._config.custom_ca:
+                kwargs['ssl_ca_cert'] = self._config.custom_ca
+            if self._config.is_secure is not None:
+                kwargs['verify_ssl'] = self._config.is_secure
+            self._kfp_client_instance = Client(**kwargs)
+        return self._kfp_client_instance
```
andreyvelich, line 432: Do we find this useful?
Given that `kfp` will be installed users can always directly import client, and construct it if needed.

```
from kfp.client import Client
```

sdk/python/kfp/kubeflow_client/pipelines_client.py
```diff
@@ -0,0 +1,1377 @@
...
+            if self._config.is_secure is not None:
+                kwargs['verify_ssl'] = self._config.is_secure
+            self._kfp_client_instance = Client(**kwargs)
+        return self._kfp_client_instance
+
+    # ------------------------------------------------------------------
+    # Private helpers — connection
+    # ------------------------------------------------------------------
+
+    def _build_api_configuration(
```
andreyvelich: Why do we need this?

andreyvelich: Shall we try to remove additional helpers from this PR, so we can add them later it if users requested?

sdk/python/kfp/kubeflow_client/constants.py
MStokluska: Thanks @andreyvelich 

I'll restructure to align. One question: in the Trainer SDK `backend.py` contains the execution logic. In our case, the "backend" is the KFP API server - should `backends/kubernetes/backend.py` hold the connection/auth logic (host discovery, token management, credential setup), while `api/pipelines_client.py` stays as the user-facing interface that delegates to the backend?

sdk/python/kfp/kubeflow_client/pipelines_client.py
MStokluska: Good call, will update.

sdk/python/kfp/kubeflow_client/pipelines_client.py
MStokluska: To mask user_token in logs and tracebacks. Without it, the default dataclass __repr__ would print the full bearer token in plain text whenever the config object appears in a log line, exception, or notebook repr() output. The custom repr replaces the token value with '***'.

sdk/python/kfp/kubeflow_client/constants.py
MStokluska: Based on further comments my understanding is:

We want to align with the TrainerClient/OptimizerClient pattern:

- backends/kubernetes/backend.py: KubernetesBackend with connection setup, host discovery, auth/token management, namespace resolution, and API service instances
- api/pipelines_client.py: thin user-facing PipelinesClient that delegates to self.backend

Will move types and constants accordingly. Should types/constants live under backends/kubernetes/ or at the top level (like trainer/types/ in the SDK)?

sdk/python/kfp/kubeflow_client/pipelines_client.py
MStokluska: Agreed, this will be addressed in the restructure. I'll move namespace resolution and API instances into KubernetesBackend, and the client will access them via `self.backend.namespace`, `self.backend.pipelines_api`, etc.

sdk/python/kfp/kubeflow_client/pipelines_client.py
MStokluska: Yes - this will move into KubernetesBackend as part of the restructure.

sdk/python/kfp/kubeflow_client/pipelines_client.py
MStokluska: Good catch. Will fix

sdk/python/kfp/kubeflow_client/pipelines_client.py
MStokluska: `cascade=True` is intentionally unconditional here due to a server-side constraint.

The backend's DeletePipeline requires `cascade=true` whenever the pipeline has any versions:

- If cascade is false, fails if the pipeline has existing pipeline versions.
- If cascade is true, deletes all pipeline versions first, then deletes the pipeline.

With cascade=false, the server returns:

"Failed to delete pipeline ... as it has existing pipeline versions. Set cascade=true to delete all versions"

Since every pipeline has at least one version (created at upload time), `cascade=false` would always fail in practice. So:

force = client-side safety gate (prevents accidental multi-version deletion without explicit intent)
cascade=True = server-side requirement (always needed because versions always exist)
Setting cascade=force would break the force=False + single-version case, the server would reject the delete.

sdk/python/kfp/kubeflow_client/api/pipelines_client_test.py
MStokluska: The existing `kfp.Client` tests (`sdk/python/kfp/client/client_test.py`) use the same `unittest.TestCase` + `absl.testing.parameterized` style. I followed the in-repo convention. Happy to align with Kubeflow SDK style (`pytest.mark.parametrize`) if this is preferred

sdk/python/kfp/kubeflow_client/api/pipelines_client.py
MStokluska: Fair point. The main value is as a migration aid during the transition, it forwards connection config so users don't have to re-specify host, token, namespace, ssl_ca_cert when they need an operation not yet covered by `PipelinesClient` (e.g., recurring runs, artifacts).

Once Phase 2+ covers the full surface, this can be deprecated and removed. Happy to drop it now if you'd prefer users explicitly construct `kfp.Client` when needed, the cost is just a few lines of config duplication on the user side.

sdk/python/kfp/kubeflow_client/pipelines_client.py
MStokluska: Thanks @andreyvelich 

Could you clarify what you mean here? `_build_api_configuration` is called from `__init__ `  it translates `KubernetesBackendConfig` into the `kfp_server_api.Configuration` (host discovery, SSL, auth setup). It's required for the client to connect.

Are you asking whether:

- This logic should live in a separate backends/kubernetes/backend.py rather than in PipelinesClient directly (per the restructure discussion)?
- The auto-discovery complexity (env var, in-cluster DNS, kubeconfig proxy) is too much for Phase 1 and we should require an explicit base_url instead?
- Something else?

sdk/python/kfp/kubeflow_client/constants.py
andreyvelich: >backends/kubernetes/backend.py: KubernetesBackend with connection setup, host discovery, auth/token management, namespace resolution, and API service instances
api/pipelines_client.py: thin user-facing PipelinesClient that delegates to self.backend

Yeah, that sounds good to me! Let's keep the auth logic in the `backends/kubernetes/` directory for now.

>Should types/constants live under backends/kubernetes/ or at the top level (like trainer/types/ in the SDK)?

If types or constants are specific to k8s backend, we can move them to `backends/kubernetes/types.py and backends/kubernetes/constants.py`.
Later, we can decouple it, once we support other backends.

sdk/python/kfp/kubeflow_client/pipelines_client.py
andreyvelich: I see, that makes sense.

sdk/python/kfp/kubeflow_client/pipelines_client.py
andreyvelich: Oh, sorry, I missed it.
Yeah, maybe moving those helper functions to a separate files make sense:
```
backends/kubernetes/utils.py
backends/kubernetes/auth.py
```

So, `backends/kubernetes/backend.py` mostly focus on APIs that exposed by `KubernetesBackend` class, like in Trainer: https://github.com/kubeflow/sdk/blob/main/kubeflow/trainer/backends/kubernetes/backend.py


WDYT @MStokluska ?

sdk/python/kfp/kubeflow_client/constants.py
andreyvelich: Clarified my points about auth here: https://github.com/kubeflow/pipelines/pull/13405#discussion_r3374588342

sdk/python/kfp/kubeflow_client/api/pipelines_client_test.py
andreyvelich: We found `pytest.mark.parametrize` is very easy to maintain and track test cases we verify. LLMs are pretty good to understand it too.
So if @HumairAK @droctothorpe and @mprahl don't have any concerns, I suggest we do the same for PipelinesClient().

MStokluska:
hey @andreyvelich 

I've addressed feedback in https://github.com/kubeflow/pipelines/pull/13405/commits/4ec9c6901154f1b65f791bc0c0f70f004f98da07

What changed:

Split the monolithic pipelines_client.py into:
- api/pipelines_client.py: core client logic
- backends/kubernetes/{backend,auth,utils,types,constants}.py: connection layer

Other fixes: 
- fix resolve_namespace NameError bug, 
- narrow bare except clauses, 
- drop experiment auto-create in favor of server default, 
- respect version param for Pipeline objects, 
- raise on missing refresh hook, 
- surface version rename failure via warnings.warn.

I would appreciate another round of review when time allows. Thanks!

sdk/python/kfp/kubeflow_client/pipelines_client.py
nsingla: @andreyvelich I am not clear about your question. Are you suggesting to move pipelines_client to a common repo or the license to a common place and component repos can then inject during build time?

nsingla (commented):
### PR Review Summary

**Scope**: 14 new files (+3091 lines) reviewed across client API, Kubernetes backend, auth, utils, types, constants, and tests.

#### Impact Assessment
- **Overall Risk**: Medium — well-structured Phase 1 with good test coverage; a few correctness and safety items
- **Breaking Changes**: None (entirely new package)

#### Key Findings
- `_load_pipeline_spec()` silently merges >2 YAML documents via `dict.update()` — data loss risk
- Broad `OSError` catch in auth masks `PermissionError` on token files
- `wait_for_run_status()` defaults to no timeout — can block indefinitely
- `str | None` union syntax is inconsistent with the rest of the SDK (which uses `typing.Optional`)

#### Positive Observations
- Clean separation of concerns ready for multi-backend support
- Token masking in `__repr__`, `first_poll_succeeded` guard in wait loop
- Thorough auth retry test coverage
- `force=True` guard on `delete_pipeline()` with efficient `page_size=2` check
- Comprehensive docstrings with `Note:` blocks for edge cases

sdk/python/kfp/kubeflow_client/api/pipelines_client.py
```diff
@@ -0,0 +1,1176 @@
...
+            upload. A best-effort rename is attempted afterward; if the rename
+            fails (e.g. due to permissions), a warning is logged and the
+            returned version retains the server-generated name.
+
+        Args:
+            pipeline: A ``@dsl.pipeline``-decorated function or a path to a
+                compiled pipeline YAML file.
+            name: Display name for the pipeline. If omitted, auto-generated
+                from the function's ``@dsl.pipeline(name=...)`` value or
+                the filename without extension.
```
nsingla, line 93: **[IMPORTANT]** Style inconsistency: the existing `kfp.Client` (and the rest of the SDK codebase) uses `typing.Optional[str]` for optional type hints, not the `str | None` union syntax.

While `from __future__ import annotations` makes `str | None` parse correctly on Python 3.9 (the SDK's minimum), it creates two issues:

1. **Runtime breakage in edge cases**: `dataclasses.fields()` and `typing.get_type_hints()` evaluate annotations at runtime. On Python 3.9, `str | None` raises `TypeError` during evaluation. This directly affects `KubernetesBackendConfig` (a dataclass with `str | None` fields in `types.py`).

2. **Codebase inconsistency**: Every other module in `sdk/python/kfp/` uses `Optional[str]` from `typing`. Mixing styles makes the codebase harder to grep and maintain.

`Optional[X]` is equivalent to `X | None` — there's no functional advantage to either. The only reason to prefer `Optional` here is consistency with the existing codebase and safe runtime behavior on Python 3.9.

Suggestion: Switch to `Optional[str]` throughout this new package to match the rest of the SDK.

sdk/python/kfp/kubeflow_client/api/pipelines_client.py
```diff
@@ -0,0 +1,1176 @@
...
+
+        if not docs or not docs[0]:
+            raise ValueError(
+                f'Pipeline file is empty or contains no valid YAML '
+                f'documents: {file_path}')
+        pipeline_spec_dict = docs[0]
+        platform_spec_dict = {}
+        for doc in docs[1:]:
+            if doc:
+                platform_spec_dict.update(doc)
```
nsingla: **[CRITICAL]** `dict.update()` on arbitrary extra YAML documents silently overwrites keys if there are >2 documents or overlapping keys between documents. This is a data loss risk for malformed pipeline files.

The KFP pipeline spec format expects at most two YAML documents: the pipeline spec and an optional platform spec. Enforcing this would be safer:

```python
if len(docs) > 2:
    raise ValueError(
        f'Expected at most 2 YAML documents (pipeline spec + platform spec), '
        f'got {len(docs)} in {file_path!r}.')
platform_spec_dict = docs[1] if len(docs) > 1 and docs[1] else {}
```

sdk/python/kfp/kubeflow_client/backends/kubernetes/auth.py
```diff
@@ -0,0 +1,70 @@
...
+        logger.debug(
+            'No in-cluster token file found; skipping credential setup.')
+        return
+
+    try:
+        from kfp.client.set_volume_credentials import ServiceAccountTokenVolumeCredentials
+    except ImportError:
+        logger.debug(
+            'In-cluster credential module not available.', exc_info=True)
+        return
```
nsingla, line 51: **[IMPORTANT]** `OSError` is too broad here — it's the parent of `PermissionError`, `IsADirectoryError`, etc. If the token file exists but has wrong permissions, this silently proceeds without authentication, which is a security-relevant misconfiguration that should not be swallowed.

Suggestion: Catch `FileNotFoundError` specifically:
```python
except FileNotFoundError:
    logger.warning('Token file not found; proceeding without auth.', exc_info=True)
```
Let `PermissionError` and other `OSError` subclasses propagate so the user knows their credentials are misconfigured.

sdk/python/kfp/kubeflow_client/api/pipelines_client.py
```diff
@@ -0,0 +1,1176 @@
...
+                    filtered_count, original_count)
+
+        return response
+
+    def wait_for_run_status(
+        self,
+        run: str | Run,
+        *,
+        status: set[str] | None = None,
+        timeout: int | None = None,
```
nsingla: **[IMPORTANT]** `timeout=None` (the default) means this method polls forever. For a "simplified client" aimed at new users, an indefinite block is a footgun — especially in notebooks where there's no obvious way to interrupt.

Consider either:
1. Setting a sensible default timeout (e.g., `timeout=3600`)
2. Adding a prominent note in the docstring: "**Warning**: With `timeout=None`, this method blocks indefinitely."

sdk/python/kfp/kubeflow_client/backends/kubernetes/backend.py
```diff
@@ -0,0 +1,126 @@
...
+
+    @property
+    def experiment_api(self) -> kfp_server_api.ExperimentServiceApi:
+        """Experiment service API instance."""
+        return self._experiment_api
+
+    @property
+    def upload_api(self) -> kfp_server_api.PipelineUploadServiceApi:
+        """Pipeline upload service API instance."""
+        return self._upload_api
```
nsingla, line 108: **[IMPORTANT]** Double namespace resolution: `_build_api_configuration()` calls `utils.resolve_namespace(config.namespace)` and assigns to `self._namespace` (line 78). This makes the lazy resolution in the `namespace` property (line 62-64) dead code — `self._namespace` is never `None` after `__init__`.

This also means namespace is resolved eagerly during construction, even if it's never used. Consider removing the assignment here and letting the property handle it lazily, or removing the lazy path from the property.

sdk/python/kfp/kubeflow_client/backends/kubernetes/auth.py
```diff
@@ -0,0 +1,70 @@
...
+    try:
+        credentials = ServiceAccountTokenVolumeCredentials(path=token_path)
+        credentials.refresh_api_key_hook(api_config)
+        api_config.api_key_prefix['authorization'] = 'Bearer'
+        api_config.refresh_api_key_hook = credentials.refresh_api_key_hook
+    except OSError:
+        logger.warning(
+            'Could not set up in-cluster credentials. '
+            'Proceeding without authentication.',
+            exc_info=True)
```
nsingla, line 60: **[SUGGESTION]** The `hasattr` + `callable` check is unnecessarily complex. `kfp_server_api.Configuration` always defines `refresh_api_key_hook` (initialized to `None`), so `hasattr` always returns `True`. Simplify:

```python
if api_config.refresh_api_key_hook is not None:
    api_config.refresh_api_key_hook(api_config)
else:
    raise RuntimeError('Token expired but no refresh hook is configured. '
                       'Re-create the client with a fresh token.')
```

sdk/python/kfp/kubeflow_client/api/pipelines_client_test.py
```diff
@@ -0,0 +1,1134 @@
...
+
+        def bad_pipeline():
+            pass
+
+        with patch('tempfile.mkdtemp', return_value='/tmp/fake-tmpdir'):
+            with patch(
+                    'kfp.compiler.Compiler.compile',
+                    side_effect=RuntimeError('compile failed')):
+                with patch('shutil.rmtree') as mock_rm:
+                    with self.assertRaisesRegex(ValueError,
```
nsingla: **[SUGGESTION]** `os.unlink(f.name)` in the test body leaks the temp file if the assertion on line 998 fails before reaching cleanup. Use `self.addCleanup(os.unlink, f.name)` right after creating the file, or use `NamedTemporaryFile(delete=True)` with a context manager.

sdk/python/kfp/kubeflow_client/api/pipelines_client.py
alyssacgoins: @nsingla @MStokluska what is the benefit of `timeout=None` here? I think option **#1** (default timeout) makes more sense.

sdk/python/kfp/kubeflow_client/api/pipelines_client.py
```diff
@@ -0,0 +1,1176 @@
...
+            construction with auto-discovery).
+    """
+
+    def __init__(
+        self,
+        backend_config: KubernetesBackendConfig | None = None,
+    ) -> None:
+        if backend_config is None:
+            backend_config = KubernetesBackendConfig()
+        self._backend = KubernetesBackend(backend_config)
```
alyssacgoins, line 61: What is the  expected behavior for incorrect KubernetesBackendConfig values (ie incorrect `base_url` or `custom_ca` formatting or value)? From my understanding, it looks like a user would not realize their PipelinesClient wasn't properly configured until they tried to execute a request.

sdk/python/kfp/kubeflow_client/api/pipelines_client.py
MStokluska: Thanks @nsingla 

I'm happy to change this for consistency with the existing `kfp.client` however, although this is just a semantic change it diverges from what was agreed in the KEP. If I understand correctly, with `from _future_ import annotations` in every module, annotations are lazy strings meaning there will be no runtime TypeErrors on 3.9.

Any thoughts @andreyvelich - would you be happy with such change?

sdk/python/kfp/kubeflow_client/api/pipelines_client.py
MStokluska: Thanks @nsingla @alyssacgoins 

The KEP explicitly defines timeout=None as "poll until a stop condition", and the signature was agreed upon as timeout: int | None = None. While I agree that an indefinite block is a footgun for notebook users, changing the default to a fixed value would diverge from the approved contract.

I don't see a strong technical objection to setting a default timeout (e.g. 3600s), but I'd like to get input from the KEP approvers as well before changing the agreed-upon signature. Happy to make the change once we get agreement.

In the meantime, every usage example in the KEP explicitly passes timeout= - so the documentation already guides users toward setting one.

@andreyvelich any opinion on this?

sdk/python/kfp/kubeflow_client/api/pipelines_client.py
MStokluska: Good point @alyssacgoins! Currently, invalid config (e.g. unreachable base_url, nonexistent custom_ca) only surfaces at the first API call.

There are two patterns in the ecosystem:

- kfp.Client: calls healthz in the constructor with 5 retries (blocks up to 25s, raises TimeoutError)
- TrainerClient: calls verify_backend() in the constructor (single K8s API call, logs a warning but never raises)

I can add a lightweight _verify_backend() matching the Trainer pattern (single healthz call, warning-only on failure). Would that be sufficient, or would you prefer the hard-fail approach like kfp.Client?

Any preference? (cc @andreyvelich)

MStokluska:
@nsingla @alyssacgoins thank you for your review - I left 3 opened questions on: 

1. https://github.com/kubeflow/pipelines/pull/13405#discussion_r3411652074
2. https://github.com/kubeflow/pipelines/pull/13405#discussion_r3411696649
3. https://github.com/kubeflow/pipelines/pull/13405#discussion_r3411755075

Once we agree I'll update the PR.

sdk/python/kfp/kubeflow_client/api/pipelines_client.py
alyssacgoins: @MStokluska I think matching the TrainerClient pattern with `_verify_backend()` is the right call here. Thanks!

sdk/python/kfp/kubeflow_client/api/pipelines_client.py
andreyvelich: Yeah, I would go ahead and use `|` if we don't have any blockers from the KFP side.
Since Python 3.9 is deprecated, we fully migrated to Py 3.10 as part of: https://github.com/kubeflow/sdk/pull/288
cc @Fiona-Waters @kramaranya @jonburdo

sdk/python/kfp/kubeflow_client/api/pipelines_client.py
andreyvelich: Looking at Trainer APIs, we set the default timeout to 600 seconds: https://github.com/kubeflow/sdk/blob/main/kubeflow/trainer/api/trainer_client.py#L237
Maybe it makes sense to do the same for KFP for now?
We can update KEP with that change, if you agree @MStokluska @HumairAK @mprahl @kramaranya

sdk/python/kfp/kubeflow_client/api/pipelines_client.py
andreyvelich: Yeah, adding verify_backend() makes sense, since in the future we also should check the version of installed KFP control plane via ConfigMap: https://github.com/kubeflow/sdk/issues/221

sdk/python/kfp/kubeflow_client/api/pipelines_client.py
```diff
@@ -0,0 +1,1176 @@
...
+        }
+        if description:
+            upload_kwargs['description'] = description
+
+        return self._backend.upload_api.upload_pipeline_version(
+            package_path, **upload_kwargs)
+
+    # ------------------------------------------------------------------
+    # Private helpers — run creation
+    # ------------------------------------------------------------------
```
andreyvelich: @MStokluska Would it make sense to move all of the helper function and API implementations to the `backends/kubernetes/backend.py` ?
I feel like those functions are specific to Kubernetes only.
If you take a look at [trainer_client.py](https://github.com/kubeflow/sdk/blob/main/kubeflow/trainer/api/trainer_client.py), `TrainerClient()` is a simple wrapper on top of backends (Kubernetes, Container, Local) APIs.
E.g.:
```py
    def list_runtimes(self) -> list[types.Runtime]:
        return self.backend.list_runtimes()
```

So all implementation which is specific to the backend is delegated to the `backend.py`.
SparkClient has the [similar model](https://github.com/kubeflow/sdk/blob/main/kubeflow/spark/api/spark_client.py#L129-L143).

andreyvelich (commented):
/ok-to-test

sdk/python/kfp/kubeflow_client/api/pipelines_client_test.py
andreyvelich: @MStokluska Any thoughts on this?

sdk/python/kfp/kubeflow_client/api/pipelines_client.py
MStokluska: Thanks @andreyvelich 

Addressed in https://github.com/kubeflow/pipelines/pull/13405/commits/d8b8c0b399a9f3aa52943fac0b97a60fa820d47a

sdk/python/kfp/kubeflow_client/api/pipelines_client.py
MStokluska: Thank you, addressed in https://github.com/kubeflow/pipelines/pull/13405/commits/d8b8c0b399a9f3aa52943fac0b97a60fa820d47a

sdk/python/kfp/kubeflow_client/api/pipelines_client.py
MStokluska: Thanks, I updated it to 600s in https://github.com/kubeflow/pipelines/pull/13405/commits/d8b8c0b399a9f3aa52943fac0b97a60fa820d47a

Let me know please if you see this as an issue. Thanks

sdk/python/kfp/kubeflow_client/api/pipelines_client.py
MStokluska: Updated in https://github.com/kubeflow/pipelines/pull/13405/commits/d8b8c0b399a9f3aa52943fac0b97a60fa820d47a

Let me know if you see any issues with this approach. Thanks

sdk/python/kfp/kubeflow_client/api/pipelines_client.py
andreyvelich: Thanks @MStokluska!
Please can you also update KEP, so we aligned with the implementation.

sdk/python/kfp/kubeflow_client/api/pipelines_client.py
```diff
@@ -0,0 +1,888 @@
...
+
+        Raises:
+            ValueError: If no pipeline matches or multiple pipelines match.
+        """
+        pipeline_id = self._backend._get_pipeline_id_by_name(name)
+        if pipeline_id is None:
+            raise ValueError(f'Pipeline not found: {name!r}. '
+                             'Use list_pipelines() to see available pipelines.')
+        return self._backend.pipelines_api.pipeline_service_get_pipeline(
+            pipeline_id=pipeline_id)
```
andreyvelich: I believe, `pipelines_api` also makes sense only for k8s backend, not for container or local.

@MStokluska Would it make sense to move everything to `kubernetes/backend.py` which is specific to that backend, and PipelinesClient() just simple invoke those functions like in [TrainerClient](https://github.com/kubeflow/sdk/blob/main/kubeflow/trainer/api/trainer_client.py) or SparkClient:
```py
PipelinesClient:

    def upload_pipeline(...):
      self.backend.upload_pipeline(...)
    
    def run(...):
      self.backend.run(..)

...
```

sdk/python/kfp/kubeflow_client/api/pipelines_client_test.py
MStokluska: hey @andreyvelich (sorry for slow responses) - I'm happy to change it. Will update it in next commit and if required we can always roll back.

sdk/python/kfp/kubeflow_client/api/pipelines_client.py
MStokluska: Hey @andreyvelich 

I'll restructure so that PipelinesClient becomes a thin wrapper matching the TrainerClient pattern.

The only logic that stays in PipelinesClient would be the pre-processing that's genuinely SDK-level (compiling callables to YAML, inferring names from @dsl.pipeline decorators), which the client can pass as resolved inputs to the backend.

MStokluska: Hey, changes pushed in https://github.com/kubeflow/pipelines/pull/13405/commits/da7de56bcf39ac84995575f0f3fcef8dc8e75648 - let me know what you think! Thanks!

sdk/python/kfp/kubeflow_client/api/pipelines_client_test.py
MStokluska: Hey, updated in https://github.com/kubeflow/pipelines/pull/13405/commits/10f4441c35f4003111aa6c4e6189f9f634e54da6

sdk/python/kfp/kubeflow_client/api/pipelines_client.py
MStokluska: Thanks, PR opened: https://github.com/kubeflow/sdk/pull/535

andreyvelich (commented):
Thanks for this work @MStokluska!
/lgtm
/assign @nsingla @droctothorpe @HumairAK @Fiona-Waters @kramaranya @alyssacgoins @mprahl

sdk/python/kfp/kubeflow_client/backends/kubernetes/backend.py
```diff
@@ -0,0 +1,1080 @@
...
+                return run_response
+
+            if current_state in constants.TERMINAL_STATES:
+                logger.info(
+                    'Run %s reached terminal state %r before target %s.',
+                    run_id, current_state, target_states)
+                self._invoke_callbacks(callbacks, run_response)
+                return run_response
+
+            if timeout is not None:
```
kramaranya, line 472: Is it still the case considering https://github.com/kubeflow/sdk/pull/535/changes?

kramaranya (commented):
Thanks @MStokluska for this great work!
Left a tiny comment, other than that lgtm

sdk/python/kfp/kubeflow_client/backends/kubernetes/backend.py
andreyvelich: Yeah, @MStokluska let's remove this check to wait without the timeout.

MStokluska: Thanks @kramaranya 
Done in https://github.com/kubeflow/pipelines/pull/13405/commits/de8cc0149ec3c712f6358bb0d802eebf7b62e075

kramaranya:
Thanks!
/lgtm

sdk/python/kfp/kubeflow_client/backends/kubernetes/auth.py
```diff
@@ -0,0 +1,69 @@
+# Copyright The Kubeflow Authors
```
droctothorpe, line 1: Curious what the rationale is behind redefining this as opposed to just importing the relevant module from kfp.

sdk/python/kfp/kubeflow_client/backends/kubernetes/backend.py
```diff
@@ -0,0 +1,1079 @@
...
+                'operation': 'EQUALS',
+                'key': 'state',
+                'stringValue': status.upper(),
+            })
+
+        filter_str = None
+        if filter_predicates:
+            filter_str = json.dumps({'predicates': filter_predicates})
+
+        response = self._run_api.run_service_list_runs(
```
mprahl, line 391: This seems like a limitation on the KFP backend for not being able to filter runs by pipeline ID. Could you please create a GitHub issue to add support for this? 😄

sdk/python/kfp/kubeflow_client/api/pipelines_client.py
```diff
@@ -0,0 +1,431 @@
...
+
+    Args:
+        backend_config: Connection parameters for the KFP API server.
+            When ``None``, uses ``KubernetesBackendConfig()`` (zero-arg
+            construction with auto-discovery).
+    """
+
+    def __init__(
+        self,
+        backend_config: KubernetesBackendConfig | None = None,
```
mprahl, line 57: I think the KEP said the constructor would be backend agnostic even if now we only support the Kubernetes backend. Could you define a base class that `KubernetesBackendConfig` inherits from and change the constructor to use the base class?

sdk/python/kfp/kubeflow_client/backends/kubernetes/utils.py
```diff
@@ -0,0 +1,111 @@
...
+def resolve_namespace(configured_namespace: str | None) -> str:
+    """Return the configured namespace, auto-detecting if needed.
+
+    Resolution order:
+        1. Explicitly configured via ``KubernetesBackendConfig.namespace``.
+        2. In-cluster: ``/var/run/secrets/kubernetes.io/serviceaccount/namespace``.
+        3. Out-of-cluster: namespace from the current kubeconfig context.
+        4. Fallback: ``"kubeflow"``.
+
+    Note: Does not read ~/.config/kfp/context.json (used by kfp.Client's
```
mprahl, line 128: Should this honor the existing `kfp.Client` context file as well? The KEP says namespace auto-detection should follow `kfp.Client` conventions, and today `kfp.Client` persists namespace in `~/.config/kfp/context.json` via `set_user_namespace()`. Ignoring that here seems like a user-visible behavior change for local multi-user setups.

sdk/python/kfp/kubeflow_client/api/pipelines_client.py
andreyvelich: @mprahl Could we do this as a followup?
We might need to define base class for other clients as well in the future: https://github.com/kubeflow/sdk/blob/main/kubeflow/trainer/api/trainer_client.py#L35-L37

sdk/python/kfp/kubeflow_client/backends/kubernetes/backend.py
```diff
@@ -0,0 +1,1079 @@
...
+                    filtered_count, original_count)
+
+        return response
+
+    def wait_for_run_status(
+        self,
+        run: str | Run,
+        *,
+        status: set[str] | None = None,
+        timeout: int = 600,
```
mprahl: Could we support no timeout (wait forever) by specifying `timeout=None`?

sdk/python/kfp/kubeflow_client/backends/kubernetes/backend.py
```diff
@@ -0,0 +1,1079 @@
...
+        polling_interval: int = 5,
+        callbacks: list[Callable[[Run], None]] | None = None,
+    ) -> Run:
+        """Wait for a run to reach a target state."""
```
mprahl: Nit: A docstring would be nice to indicate the time units for timeout and polling_interval.

sdk/python/kfp/kubeflow_client/constants.py
```diff
@@ -0,0 +1,48 @@
...
+    'RUN_PENDING',
+    'RUN_PAUSED',
+    'RUN_COMPLETE',
+    'TERMINAL_STATES',
+]
+
+RUN_SUCCEEDED = 'succeeded'
+RUN_FAILED = 'failed'
+RUN_SKIPPED = 'skipped'
+RUN_ERROR = 'error'
```
mprahl: I think this exposes a run state that the v2beta1 API does not actually return. The generated `RuntimeState` enum only includes `PENDING`, `RUNNING`, `SUCCEEDED`, `SKIPPED`, `FAILED`, `CANCELING`, `CANCELED`, and `PAUSED` — not `ERROR`. As written, `wait_for_run_status(status={constants.RUN_ERROR})` looks supported but can never be satisfied by a real run state.

sdk/python/kfp/kubeflow_client/backends/kubernetes/backend.py
```diff
@@ -0,0 +1,1079 @@
...
+                    display_name=version_name,
+                )
+                self._pipelines_api.pipeline_service_update_pipeline_version(
+                    pipeline_version_pipeline_id=first_version.pipeline_id,
+                    pipeline_version_pipeline_version_id=(
+                        first_version.pipeline_version_id),
+                    pipeline_version=update_body,
+                )
+                first_version.display_name = version_name
+            except kfp_server_api.ApiException:
```
mprahl: Could we avoid treating every `ApiException` here as "the upload API does not accept an initial version name"? The PATCH endpoint is a real supported API, so a failure here could also be RBAC, validation, or a transient server error. Swallowing all of those into a warning seems like it would hide genuine update failures behind a misleading fallback story.

sdk/python/kfp/kubeflow_client/backends/kubernetes/utils.py
```diff
@@ -0,0 +1,111 @@
...
+        if not (host.startswith('http://') or host.startswith('https://')):
+            logger.warning(
+                'No scheme in KF_PIPELINES_ENDPOINT %r, defaulting '
+                'to https.', endpoint_from_env)
+            host = 'https://' + host
+        return host
+
+    try:
+        import kubernetes as k8s
+    except ImportError:
```
mprahl, line 83: Should we fail loudly here instead of assuming the in-cluster DNS name? `ImportError` only tells us the Python Kubernetes client is missing, not that we are actually running in cluster. Falling back to `ml-pipeline.<ns>.svc.cluster.local` seems likely to turn a local misconfiguration into a much less obvious connection failure later. It feels safer to require explicit `base_url` / `KF_PIPELINES_ENDPOINT`, or only use the in-cluster DNS fallback when we have stronger evidence that we are really in cluster.

sdk/python/kfp/kubeflow_client/backends/kubernetes/utils.py
```diff
@@ -0,0 +1,111 @@
...
+
+import logging
+import os
+
+from kfp.kubeflow_client.backends.kubernetes import constants
+
+logger = logging.getLogger(__name__)
+
+
+def discover_host(namespace: str) -> str:
```
mprahl, line 69: I think the `namespace` flowing into `discover_host()` may be the wrong one for multi-user installs. This code uses it to build the `ml-pipeline.<namespace>.svc.cluster.local` / kube-proxy service path, but in our manifests the `ml-pipeline` service is deployed in the control-plane namespace (`kubeflow`), while this resolver is otherwise trying to discover the user/resource namespace. That seems like it would point at the wrong service in profile namespaces such as `kubeflow-user-example-com`.

sdk/python/kfp/kubeflow_client/backends/kubernetes/backend.py
```diff
@@ -0,0 +1,1079 @@
...
+
+    def _build_api_configuration(
+        self,
+        config: KubernetesBackendConfig,
+    ) -> kfp_server_api.Configuration:
+        """Build a kfp_server_api.Configuration from
+        KubernetesBackendConfig."""
+        api_config = kfp_server_api.Configuration()
+
+        if config.custom_ca:
```
mprahl, line 1076: Could we prefer a detected system CA bundle here before falling back to the generated client's certifi bundle? In a lot of enterprise setups the trusted internal roots are installed at the OS level rather than passed explicitly per client. Even a small best-effort probe for common bundle locations, for example Debian/Ubuntu and RHEL/Fedora paths, would make the default behavior line up better with the host trust configuration, while still letting custom_ca override when the user wants something explicit.

mprahl (changes requested):
I provided a high level review. I can approve it once these findings are addressed since @alyssacgoins already has an lgtm. :)

sdk/python/kfp/kubeflow_client/backends/kubernetes/auth.py
MStokluska: Hey @droctothorpe 

We import and use ServiceAccountTokenVolumeCredentials from kfp.client.set_volume_credentials directly. What auth.py adds is the discovery/fallback orchestration around it:

env var -> KFP token path -> K8s SA token path -> skip gracefully

The upstream class has no multi-path fallback, and the upstream method that does similar orchestration (_get_config_with_default_credentials) is a private instance method on kfp.Client that can't be called standalone from KubernetesBackend. The only genuinely new constant is K8S_SA_TOKEN_PATH - the other two could be imported from kfp.client to reduce duplication; happy to do that if preferred.

sdk/python/kfp/kubeflow_client/backends/kubernetes/backend.py
MStokluska: Thanks - https://github.com/kubeflow/pipelines/issues/13578

sdk/python/kfp/kubeflow_client/backends/kubernetes/utils.py
MStokluska: Hey @mprahl 

This was intentionally skipped - reading `context.json` would implicitly require supporting `set_user_namespace()` (shared mutable state between clients), which is scope beyond Phase 1. For multi-user setups, in-cluster namespace detection (SA mount) already returns the correct user namespace. Out-of-cluster users can pass it explicitly via `KubernetesBackendConfig(namespace="...")`. Happy to update it if you believe this is a strong requirement.

sdk/python/kfp/kubeflow_client/backends/kubernetes/backend.py
MStokluska: This is what we had initially and removed it after a discussion here: https://github.com/kubeflow/pipelines/pull/13405#discussion_r3389139253

I see pros and cons for both approaches but leaning towards leaving it as currently is.

sdk/python/kfp/kubeflow_client/backends/kubernetes/backend.py
MStokluska: Makes sense! Ty

sdk/python/kfp/kubeflow_client/constants.py
MStokluska: You're right - ERROR is not in the v2beta1 `RuntimeState` enum and the API will never return it. I included it because `kfp.Client.wait_for_run_completion` lists 'error' in its terminal states, but that's v1 legacy. Will remove `RUN_ERROR` from the constants and `TERMINAL_STATES`.

sdk/python/kfp/kubeflow_client/backends/kubernetes/backend.py
MStokluska: Good point - we are being too broad here. Will narrow the exception handling

sdk/python/kfp/kubeflow_client/backends/kubernetes/utils.py
MStokluska: Good catch! We will now only fall back to in-cluster DNS when the ServiceAccount namespace file (/var/run/secrets/kubernetes.io/serviceaccount/namespace) exists. Without that evidence, we will raise a ValueError directing the user to set base_url in KubernetesBackendConfig or the KF_PIPELINES_ENDPOINT env var.

sdk/python/kfp/kubeflow_client/backends/kubernetes/utils.py
MStokluska: Thanks, good point! Will follow the existing kfp client pattern to fix

sdk/python/kfp/kubeflow_client/backends/kubernetes/backend.py
MStokluska: Good idea - we are currently aligned with `kfp.Client` which handles it the same way - the generated API client (`kfp_server_api`) falls back to certifi when no explicit CA is set. System CA probing is a good enhancement idea but I'm not sure if it should be done as part of this PR.

andreyvelich:
@MStokluska @mprahl @alyssacgoins Do we have any other open questions before move this forward?

MStokluska:
> @MStokluska @mprahl @alyssacgoins Do we have any other open questions before move this forward?

Hey @andreyvelich - from me, just need a second review from Matthew based on his comments.

sdk/python/kfp/kubeflow_client/api/pipelines_client.py
mprahl: Sure!

sdk/python/kfp/kubeflow_client/backends/kubernetes/backend.py
mprahl: My preference is to not restrict the user if they want to wait forever.

sdk/python/kfp/kubeflow_client/backends/kubernetes/backend.py
mprahl: My preference is the system CA in this PR or a quick follow up PR because I think it will be a very common usecase.

mprahl:
@MStokluska , I responded to the two PR comments that still need addressing/responses (system CA bundles and be able to not timeout). After that, I can approve. :)

sdk/python/kfp/kubeflow_client/backends/kubernetes/backend.py
MStokluska: Thanks, will add `timeout=None` with default 600s.

MStokluska: @andreyvelich will update the KEP shortly too to reflect.

andreyvelich: Sure, thanks!

MStokluska:
Thanks @mprahl 

Added the system CA bundles and ability to run without timeout in [75a1e1a218d0259c70f5cc98c5bc4998f9302380](https://github.com/kubeflow/pipelines/pull/13405/changes/75a1e1a218d0259c70f5cc98c5bc4998f9302380) and rebased.

mprahl (approved):

PR #13415, opened 2026-05-19
test(frontend): use act helpers in small viewer tests

jeffspahr:
## Summary
- Replace raw `TestUtils.flushPromises()` calls in small viewer/component tests with act-aware helpers.
- Use `flushPromisesInAct()` for render follow-up drains in `HTMLViewer.test.tsx` and `CustomTableRow.test.tsx`.
- Use `invokeAndFlush()` for `PagedTable.test.tsx` sort clicks so each user action and its async follow-up are flushed together.

## Why
This is helper cleanup rather than warning reduction: the focused test batch already passed with zero `act(...)` warning lines before the change. The benefit is keeping these tests on the same React 19-safe helper pattern as the rest of the cleanup work, so future async updates introduced in these tests are less likely to escape Testing Library act boundaries.

The ascending sort test intentionally keeps separate helper calls for the two clicks. Batching both clicks into one `act` changes the interaction timing because the second click can observe the pre-flush sort state, so separate action/flush boundaries preserve the original test semantics.

## Validation
- `npm run test:ui -- src/components/viewers/HTMLViewer.test.tsx src/components/viewers/PagedTable.test.tsx src/components/CustomTableRow.test.tsx`
  - Before: 16 tests passed, 0 `act(...)` warning lines.
  - After: 16 tests passed, 0 `act(...)` warning lines.
- `npm run lint:ui`
- `npm run typecheck`
- `git diff --check`

### DIFF
--- a/frontend/src/components/CustomTableRow.test.tsx
+++ b/frontend/src/components/CustomTableRow.test.tsx
@@ -16,7 +16,7 @@
 
 import { render } from '@testing-library/react';
 import { Column, Row } from './CustomTable';
-import TestUtils from '../TestUtils';
+import { flushPromisesInAct } from '../TestUtils';
 import { CustomTableRow } from './CustomTableRow';
 
 describe('CustomTable', () => {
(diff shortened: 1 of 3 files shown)


Priyanshu-u07:
LGTM
Verified locally: All tests pass with zero act warnings. 
Helper changes look correct:
`flushPromisesInAct()` for post-render drains and `invokeAndFlush()` for sort clicks with separate boundaries preserving observation order.
No behavioral changes in tests.

manaswinidas (approved):
/lgtm
/approve

PR #13433, opened 2026-05-26
feat(frontend): fix artifact cross namespace IDOR vuln

JerT33:

Currently there is a minor issue in regarding cross namespace artifact IDOR: a user with access to namespace `B` can swap the `?namespace=` query param to view an artifact actually owned by namespace `A` when both share object storage credentials

This is a [follow up](https://github.com/kubeflow/pipelines/pull/12860#issuecomment-4189003892) from PR #12860 

This PR adds a new `mlmd-validator` helper that confirms the artifact actually belongs to the requested namespace via an MLMD lookup. The check runs in the artifact auth middleware after the existing `SubjectAccessReview` check passes, so the user still has to have access to the claimed namespace. The middleware returns `403 Artifact does not belong to the requested namespace` when the artifact is not associated with that namespace

Note: Default to a current behavior in situations when MLMD is unreachable, the artifact isn't tracked in MLMD, or when no PipelineRun namespace evidence exists

### Before

NOTE: This testing is built from an image that has the existing logic from #12860. Currently this change is not released in any KFP versions, so the following before evidence will not reflect any environments on any currently released images

Artifact viewable with namespace access
Screenshot 2026-06-11 at 3 58 23 PM

Artifact viewable with namespace access

Artifact is blocked when namespace access is removed
Screenshot 2026-06-11 at 8 04 24 PM
Switching namespace in URL allows for artifact access again
Screenshot 2026-06-11 at 8 03 50 PM

### After

Artifact is blocked when namespace access is removed
Screenshot 2026-06-11 at 7 17 01 PM
Switching namespace in URL blocks access as well
Screenshot 2026-06-11 at 7 17 30 PM

**Checklist:**

### DIFF
--- a/frontend/server/handlers/artifacts.test.ts
+++ b/frontend/server/handlers/artifacts.test.ts
@@ -1,0 +1,195 @@
+// Copyright 2025 The Kubeflow Authors
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//      http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+import { describe, it, expect } from 'vitest';
+import type { Request } from 'express';
+import { resolveArtifactCoordinates } from '../helpers/artifact-coordinates.js';
+
+function makeRequest(path: string, query: Record<string, unknown> = {}): Request {
+  return { path, query } as unknown as Request;
+}
+
+describe('resolveArtifactCoordinates', () => {
+  describe('path-based routes', () => {
+    it('extracts coordinates from /artifacts/:source/:bucket/* style URLs', () => {
+      const req = makeRequest('/artifacts/minio/ml-pipeline/hello/world.txt');
+      expect(resolveArtifactCoordinates(req)).toEqual({
+        source: 'minio',
+        bucket: 'ml-pipeline',
+        key: 'hello/world.txt',
+      });
+    });
+
+    it('extracts coordinates from /pipeline-prefixed routes', () => {
+      const req = makeRequest('/pipeline/artifacts/s3/my-bucket/path/to/file.csv');
+      expect(resolveArtifactCoordinates(req)).toEqual({
+        source: 's3',
+        bucket: 'my-bucket',
+        key: 'path/to/file.csv',
+      });
+    });
+
+    it('extracts coordinates from a non-/pipeline base path', () => {
+      const req = makeRequest('/foo/bar/artifacts/minio/my-bucket/some/key');
+      expect(resolveArtifactCoordinates(req)).toEqual({
+        source: 'minio',
+        bucket: 'my-bucket',
+        key: 'some/key',
+      });
+    });
+
+    it('decodes percent-encoded path segments once (matching Express req.params semantics)', () => {
+      const req = makeRequest('/artifacts/minio/ml-pipeline/hello%2Fworld.txt');
+      expect(resolveArtifactCoordinates(req)).toEqual({
+        source: 'minio',
+        bucket: 'ml-pipeline',
+        key: 'hello/world.txt',
+      });
+    });
+
+    it('preserves a literal %2F when the URL is double-encoded (%252F)', () => {
+      const req = makeRequest('/artifacts/minio/ml-pipeline/hello%252Fworld.txt');
+      expect(resolveArtifactCoordinates(req)).toEqual({
+        source: 'minio',
+        bucket: 'ml-pipeline',
+        key: 'hello%2Fworld.txt',
+      });
+    });
+
+    it('returns null on malformed percent-encoding (fail-closed)', () => {
+      const req = makeRequest('/artifacts/minio/ml-pipeline/bad%ZZkey');
+      expect(resolveArtifactCoordinates(req)).toBeNull();
+    });
+
+    it('handles keys that contain multiple slashes', () => {
+      const req = makeRequest('/artifacts/gcs/my-bucket/a/b/c/d.json');
+      expect(resolveArtifactCoordinates(req)).toEqual({
+        source: 'gcs',
+        bucket: 'my-bucket',
+        key: 'a/b/c/d.json',
+      });
+    });
+  });
+
+  describe('query-based fallback (/artifacts/get and unrecognized paths)', () => {
+    it('falls back to query when path is /artifacts/get', () => {
+      const req = makeRequest('/artifacts/get', {
+        source: 'minio',
+        bucket: 'ml-pipeline',
+        key: 'hello/world.txt',
+      });
+      expect(resolveArtifactCoordinates(req)).toEqual({
+        source: 'minio',
+        bucket: 'ml-pipeline',
+        key: 'hello/world.txt',
+      });
+    });
+
+    it('falls back to query when path is /pipeline/artifacts/get', () => {
+      const req = makeRequest('/pipeline/artifacts/get', {
+        source: 's3',
+        bucket: 'my-bucket',
+        key: 'data.csv',
+      });
+      expect(resolveArtifactCoordinates(req)).toEqual({
+        source: 's3',
+        bucket: 'my-bucket',
+        key: 'data.csv',
+      });
+    });
+
+    it('falls back to query when path does not match the artifact patterns', () => {
+      const req = makeRequest('/foo/bar', {
+        source: 'minio',
+        bucket: 'ml-pipeline',
+        key: 'k',
+      });
+      expect(resolveArtifactCoordinates(req)).toEqual({
+        source: 'minio',
+        bucket: 'ml-pipeline',
+        key: 'k',
+      });
+    });
+  });
+
+  describe('missing or non-string query values', () => {
+    it('returns empty strings when query params are absent', () => {
+      const req = makeRequest('/artifacts/get');
+      expect(resolveArtifactCoordinates(req)).toEqual({
+        source: '',
+        bucket: '',
+        key: '',
+      });
+    });
+
+    it('rejects array-valued query params (treats them as missing)', () => {
+      const req = makeRequest('/artifacts/get', {
+        source: ['minio', 'sneaky'],
+        bucket: 'ml-pipeline',
+        key: 'k',
+      });
+      expect(resolveArtifactCoordinates(req)).toEqual({
+        source: '',
+        bucket: 'ml-pipeline',
+        key: 'k',
+      });
+    });
+
+    it('rejects object-valued query params (treats them as missing)', () => {
+      const req = makeRequest('/artifacts/get', {
+        source: { nested: 'minio' },
+        bucket: 'ml-pipeline',
+        key: 'k',
+      });
+      expect(resolveArtifactCoordinates(req)).toEqual({
+        source: '',
+        bucket: 'ml-pipeline',
+        key: 'k',
+      });
+    });
+  });
+
+  describe('coordinate-source spoofing defense', () => {
+    it('uses path coordinates when both path and query are present', () => {
+      // Attacker plants benign values in query, real values in path.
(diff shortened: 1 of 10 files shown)


frontend/server/helpers/mlmd-validator.ts
```diff
@@ -0,0 +1,293 @@
...
+  let artifacts: ArtifactInfo[];
+  try {
+    artifacts = await getArtifactsByUri(envoyAddress, [artifactUri]);
+  } catch (error) {
+    console.warn(
+      `[SECURITY] MLMD artifact lookup failed for URI "${artifactUri}", ` +
+        `allowing access (fail-open). Error: ${error}`,
+    );
+    return { valid: true, reason: 'mlmd-unavailable' };
+  }
```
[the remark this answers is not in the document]
JerT33: This would be an optional follow-up PR to have an variable for this logic

frontend/server/handlers/artifacts.ts
```diff
@@ -187,6 +191,46 @@ export function getArtifactsAuthMiddleware(
...
+              `User: ${userId}, ` +
+              `Claimed namespace: ${namespace}, ` +
+              `Actual namespace: ${validation.actualNamespace}, ` +
+              `URI: ${artifactUri}, ` +
+              `Path: ${request.originalUrl}`,
+          );
+          response.status(403).send('Artifact does not belong to the requested namespace');
+          return;
+        }
+      }
```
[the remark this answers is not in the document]
JerT33: This could be implemented, however i feel that this would not have a significant impact on normal use. Would be open to implementing a caching functionality for this, but would like additional push for this by reviewers

frontend/server/helpers/mlmd-validator.ts
```diff
@@ -0,0 +1,334 @@
...
+
+  if (artifacts.length === 0) {
+    console.warn(
+      `[SECURITY] Artifact not found in MLMD for URI "${artifactUri}", ` +
+        `allowing access (fail-open for legacy/external artifacts).`,
+    );
+    return { valid: true, reason: 'artifact-not-found' };
```
jeffspahr: This fail-open path leaves the new protection bypassable for tracked storage sources whenever the request URI is a prefix or wildcard rather than an exact MLMD artifact URI. For example, the MinIO/S3 handler can serve a directory prefix after `getObject` misses, and the GCS handler expands `*` patterns, so `artifact-not-found` here can still return objects under that unowned prefix/pattern. Fail closed when ownership evidence is absent for these validated sources, or make legacy fail-open an explicit opt-in/allowlist.

```suggestion
  if (artifacts.length === 0) {
    console.warn(
      `[SECURITY] Artifact not found in MLMD for URI "${artifactUri}", ` +
        `denying access because namespace ownership could not be established.`,
    );
    return { valid: false, reason: 'artifact-not-found' };
  }
```

frontend/server/handlers/artifacts.ts
```diff
@@ -86,6 +87,40 @@ export interface GCSProviderInfo {
...
+ */
+export function resolveArtifactCoordinates(
+  request: Request,
+): { source: string; bucket: string; key: string } | null {
+  const isExactGetEndpoint =
+    request.path === '/artifacts/get' || request.path === '/pipeline/artifacts/get';
+  if (!isExactGetEndpoint) {
+    const downloadPathMatch = request.path.match(
+      /^(?:\/pipeline)?\/artifacts\/([^/]+)\/([^/]+)\/(.+)$/,
+    );
```
jeffspahr: This resolver only recognizes bare `/artifacts/...` and `/pipeline/artifacts/...`, but `app.ts` mounts artifact routes under the configurable `basePath`. With any other base path, the middleware falls back to query coordinates while the handler later serves path params. Parse from the `/artifacts/` segment so middleware and handler stay aligned for all mounted base paths, and add a regression test for a non-`/pipeline` prefix.

```suggestion
  const artifactPathStart = request.path.indexOf('/artifacts/');
  const artifactPath =
    artifactPathStart >= 0 ? request.path.slice(artifactPathStart) : request.path;
  const isExactGetEndpoint = artifactPath === '/artifacts/get';
  if (!isExactGetEndpoint) {
    const downloadPathMatch = artifactPath.match(/^\/artifacts\/([^/]+)\/([^/]+)\/(.+)$/);
```

jeffspahr (changes requested):
I found two issues that still leave gaps in the intended artifact ownership check. The first is a security bypass for tracked storage sources when the served object set is broader than the exact URI checked in MLMD; the second is parser drift for non-default base paths. I included inline suggestions for the smallest viable fixes.

JerT33:
Feedback implemented, thanks for the review! @jeffspahr

frontend/server/handlers/artifacts.ts
```diff
@@ -187,6 +225,36 @@ export function getArtifactsAuthMiddleware(
...
+      const coords = resolveArtifactCoordinates(request);
+      if (coords === null) {
+        console.warn(
+          `[SECURITY] Malformed percent-encoding in artifact path. ` +
+            `User: ${userId}, Path: ${request.path}`,
+        );
+        response.status(400).send('Malformed URL encoding in artifact path');
+        return;
+      }
+      const mlmdTrackedSources = new Set(['minio', 's3', 'gcs']);
```
jeffspahr: **[P1] Validate tracked HTTP(S) artifacts too.** The ownership gate only runs for MinIO/S3/GCS, but this same endpoint supports `http` and `https`, the frontend turns MLMD HTTP(S) artifact URIs into these coordinates, and `getHttpArtifactsHandler` can fetch them with a server-wide default/inherited authorization header. A user authorized for namespace B can therefore request a victim HTTP(S) URI with `namespace=B`; this set check skips MLMD and the handler dereferences it with shared credentials. Include both schemes here (and add a mismatch integration test for one of them); `buildArtifactUri` already preserves their URI schemes.

```suggestion
      const mlmdTrackedSources = new Set(['minio', 's3', 'gcs', 'http', 'https']);
```

jeffspahr (changes requested):
The two findings from my previous review are fixed: MLMD misses now fail closed, and path coordinates are parsed from the artifact route with regression coverage. One cross-namespace path remains, however: HTTP(S) artifacts bypass the new ownership check even though the server can retrieve them with shared credentials. I included the minimal inline change and recommend adding an HTTP(S) namespace-mismatch integration test.

JerT33:
Review 2 feedback addressed. Thanks again! @jeffspahr

zazulam (commented):
/lgtm

frontend/server/helpers/mlmd-validator.ts
```diff
@@ -0,0 +1,334 @@
...
+let storePb: any = null;
+let protoLoadAttempted = false;
+
+function loadProtos(): boolean {
+  if (servicePb !== null && storePb !== null) {
+    return true;
+  }
+  try {
+    // eslint-disable-next-line @typescript-eslint/no-var-requires
+    servicePb = require('../../src/third_party/mlmd/generated/ml_metadata/proto/metadata_store_service_pb.js');
```
jeffspahr: **[P1] Keep MLMD proto loading working from the built server.** This relative require works from `server/helpers/...` and in the Docker image because the Dockerfile copies the bundle into `/server/src/...`, but `npm run start:proxy-and-server` builds and runs `server/dist/server.js`. From `dist/helpers/mlmd-validator.js`, `../../src/...` resolves to `frontend/server/src/...`, which does not exist, so `loadProtos()` returns `protos-unavailable` and fail-opens the IDOR check. Please resolve both source/dist locations or copy the bundle as part of the server build/start path.

frontend/server/handlers/artifacts.test.ts
```diff
@@ -0,0 +1,195 @@
...
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+import { describe, it, expect } from 'vitest';
+import type { Request } from 'express';
+import { resolveArtifactCoordinates } from './artifacts.js';
```
jeffspahr: **[P2] Avoid importing the full artifact handler for this pure resolver test.** Importing `./artifacts.js` initializes `minio-helper -> k8s-helper` at module load, and `k8s-helper` calls `kc.makeApiClient(...)`; on machines without an active kube context this suite fails before executing any tests with `Error: No active cluster!`. Move `resolveArtifactCoordinates` into a side-effect-free module, or mock the k8s/minio path before this import, so the test is hermetic.

jeffspahr (changes requested):
I found two remaining issues: one keeps the namespace ownership check from running in the built local server path, and one makes the new artifact resolver test non-hermetic on machines without an active kube context. Inline comments have the concrete fixes.

jeffspahr (changes requested):
Re-reviewed `a90b7174122e20a20e408e897dbe2a00164de204`. The two existing inline comments still apply and remain unresolved:

- `frontend/server/helpers/mlmd-validator.ts:30`: the built server still cannot load the MLMD proto bundle from `dist/helpers/mlmd-validator.js`; direct import returns `{ valid: true, reason: 'protos-unavailable' }` after `Cannot find module '../../src/third_party/mlmd/generated/ml_metadata/proto/metadata_store_service_pb.js`.
- `frontend/server/handlers/artifacts.test.ts:17`: the resolver unit test still imports the full artifact handler module, and `npm test -- handlers/artifacts.test.ts` fails before executing tests with `Error: No active cluster!` via `minio-helper -> k8s-helper`.

Verification on the latest head: `npm ci`, `npm run build`, and `npm test -- helpers/mlmd-validator.test.ts` pass; `npm test -- handlers/artifacts.test.ts` fails as above. I’m leaving the existing inline threads as the blockers rather than duplicating them.

JerT33:
> Re-reviewed `a90b7174122e20a20e408e897dbe2a00164de204`. The two existing inline comments still apply and remain unresolved:
> 
> * `frontend/server/helpers/mlmd-validator.ts:30`: the built server still cannot load the MLMD proto bundle from `dist/helpers/mlmd-validator.js`; direct import returns `{ valid: true, reason: 'protos-unavailable' }` after `Cannot find module '../../src/third_party/mlmd/generated/ml_metadata/proto/metadata_store_service_pb.js`.
> * `frontend/server/handlers/artifacts.test.ts:17`: the resolver unit test still imports the full artifact handler module, and `npm test -- handlers/artifacts.test.ts` fails before executing tests with `Error: No active cluster!` via `minio-helper -> k8s-helper`.
> 
> Verification on the latest head: `npm ci`, `npm run build`, and `npm test -- helpers/mlmd-validator.test.ts` pass; `npm test -- handlers/artifacts.test.ts` fails as above. I’m leaving the existing inline threads as the blockers rather than duplicating them.

Latest commit was just rebasing with master branch, updating requested changes now. Thank you!

jeffspahr (approved):
Re-reviewed `903e376dd441a6eb43578f488f674b00554279eb`. The two prior blockers are addressed:

- `frontend/server/handlers/artifacts.test.ts` now imports `resolveArtifactCoordinates` from the side-effect-free helper, and `npm test -- handlers/artifacts.test.ts` passes locally.
- The built `dist/helpers/mlmd-validator.js` path now loads the generated MLMD bundle when using the documented frontend-root install path; the direct import check now reaches the MLMD network call and returns `mlmd-unavailable` for the intentionally unreachable test URL instead of `protos-unavailable`.

Verification run locally on the latest head: `npm ci` in `frontend`, `npm run build` in `frontend/server`, `npm test -- handlers/artifacts.test.ts`, and `npm test -- helpers/mlmd-validator.test.ts` all pass. `npm test -- integration-tests/artifact-auth.test.ts` still cannot run in my no-cluster local environment because the existing integration test suite initializes Kubernetes at import time; a neighboring existing test (`integration-tests/artifact-get.test.ts`) fails the same way. GitHub `frontend-tests` and frontend integration checks are passing on this head.

juliusvonkohout:
This change breaks KFP v1 @droctothorpe  @jeffspahr @JerT33 and is a strong regression I found when testing this for Kubeflow community distribution and KFP v1 "Artifact does not belong to the requested namespace". I will try to come up with a fix

juliusvonkohout:
i plan to raise PR in the next hour

juliusvonkohout:
https://github.com/kubeflow/pipelines/pull/13683 is supposed to fix this. Otherwise the 2.17 prerelease looks very good

PR #13437, opened 2026-05-27
fix(backend): resolve artifact OneOf outputs at runtime (#13436)

### LINKED ISSUE #13436: [backend] dsl.OneOf artifact outputs from mutually exclusive branches fail at runtime
### Environment

* KFP version:
Observed with Kubeflow Pipelines `2.16.0`.

* KFP SDK version:
Observed with Kubeflow Pipelines SDK `2.16.0`.

### Steps to reproduce

Run a pipeline that uses `dsl.OneOf` to collect an artifact output from mutually exclusive `dsl.If` / `dsl.Else` branches, then pass that artifact to a downstream task.

```python
from kfp import dsl
from typing import NamedTuple

@dsl.component()
def first_component(text: str) -> str:
    return text

@dsl.component()
def print_and_return(text: str, result: dsl.Output[dsl.Artifact]) -> NamedTuple('outputs', text=str):
    with open(result.path, 'w') as f:
        f.write(text)

    from collections import namedtuple
    outputs = namedtuple('outputs', ['text'])
    return outputs(text=text)

@dsl.component()
def announce_result(result: dsl.Input[dsl.Artifact]) -> str:
    with open(result.path, 'r') as f:
        text = f.read()
    print(f'The result is: {text}')
    return text

@dsl.pipeline(name='my-pipeline')
def my_pipeline(text: str) -> dsl.Artifact:
    coin_flip_task = first_component(text=text)
    with dsl.If(coin_flip_task.output == 'heads'):
        t1 = print_and_return(text='Got heads')
    with dsl.Else():
        t2 = print_and_return(text='Draw')

    oneof = dsl.OneOf(t1.outputs['result'], t2.outputs['result'])
    announce_result(result=oneof)
    return oneof
```

The pipeline fails before the downstream task runs. The driver logs contain an error like:

```text
failed to resolve input artifact result with spec task_output_artifact:{producer_task:"condition-branches-1" output_artifact_key:"pipelinechannel--condition-branches-1-oneof-1"}: subTaskName, print-and-return-2_518, not in tasks
```

### Expected result

`dsl.OneOf` should resolve the artifact from whichever mutually exclusive branch actually executed, using the compiled task name consistently rather than looking up a runtime-specific task name that is not present in the task map.

For comparison, `dsl.OneOf` over parameter outputs from the same branches works as expected.

### Materials and Reference

This appears to be a backend/runtime resolution issue rather than a compilation failure, because the pipeline compiles successfully and the downstream task fails when the driver resolves the selected artifact input.

Observed behavior from an internal downstream report: `RHOAIENG-61956`.

mprahl:
Fixes #13436.

## Summary
- Resolve artifact `dsl.OneOf` DAG outputs by selecting the branch whose producer task actually completed or was cached instead of always taking the last selector.
- Propagate `GetProducerTask()` resolution failures through DAG collection and add focused driver coverage for complete, cached, missing-task, and no-success cases.
- Add a compiled `if_else_with_oneof_artifacts` sample under `test_data/sdk_compiled_pipelines/valid/critical` so the existing `E2ECritical` CI sweep exercises the runtime path.

## Test plan
- [x] `GOTMPDIR="/home/mprahl/git/pipelines2/.tmp/go-build" GOCACHE="/home/mprahl/git/pipelines2/.tmp/gocache" go test ./backend/src/v2/driver -run 'TestGetProducerTask_SelectsSuccessfulArtifactOneOfBranch|TestGetProducerTask_ArtifactOneOfReturnsErrorWhenNoBranchSucceeds|TestGetProducerTask_SelectsCachedArtifactOneOfBranch|TestGetProducerTask_ArtifactOneOfReturnsErrorWhenSubtasksAreMissing|Test_InferIndexedTaskName'`
- [x] `PYTHONPATH="/home/mprahl/git/pipelines2/sdk/python:/home/mprahl/git/pipelines2/api/v2alpha1/python:/home/mprahl/git/pipelines2/kubernetes_platform/python" "/home/mprahl/git/pipelines2/.venv/bin/python" "/home/mprahl/git/pipelines2/test_data/sdk_compiled_pipelines/valid/critical/if_else_with_oneof_artifacts.py"`
- [ ] Full CI


nsingla (commented):
This pr is missing compiled workflow file for the newly added pipeline

backend/src/v2/driver/resolve_test.go
hbelmiro, line 1: All success-path tests (`TestGetProducerTask_SelectsSuccessfulArtifactOneOfBranch` and `TestGetProducerTask_SelectsCachedArtifactOneOfBranch`) place the successful task as the **first** selector in the list. While these tests do correctly catch the old "pick last selector" bug (they assert `condition-2` is selected, which the old code would not return), they do not exercise the scenario where the loop must skip a non-successful first selector and select a later one.

A test where `condition-2` is CANCELED/FAILED and `condition-3` is COMPLETE would verify the loop iterates correctly past failed branches and selects the right later entry.

hbelmiro (approved):
/lgtm
/approve

PR #13452, opened 2026-05-30
fix(sdk): remove max duration limit on retries

JerT33:

Currently there is a hard cap of 1 hour on retries, regardless of user configuration. Importantly, this 1hr limit is cumulative across all retry attempts of the same component (Argo measures it from the first attempt's start). Due to the interaction with Argo Workflows, once this limit is reached the active pod will fail or no pod will be retried (if pod ran over 1hr before failing).

This PR is intended to be a quick fix. The default behavior is all still the same, but now users can override it to get around this issue for long-running components.

Further discussion/changes can be proposed to fix the issue where this time is taken from the total component time, but that will require much larger configuration changes.

#### Compiled retryPolicy with `backoff_max_duration='4h'`
```yaml
retryPolicy:
  backoffMaxDuration: 3600s   # ← reduced from '4h'
  maxRetryCount: 1
```

#### 50 minute wait example:
Screenshot 2026-05-29 at 11 58 57 PM

Initial pod start time: `2026-05-30T02:51:40Z`:
```yaml
    state:
      terminated:
        containerID: containerd://62169eb1743a4d8e19fcc9663dfd5253396ff62714f698d64f476de6cfca1919
        exitCode: 1
        finishedAt: "2026-05-30T03:47:29Z"
        reason: Error
        startedAt: "2026-05-30T02:51:40Z"
```

Pod is retried, but fails prematurely
Retry pod deadline:
4 minutes after initial pod failure (56 mins in)
```yaml
    state:
      terminated:
        containerID: containerd://75afccb4eeb1ff6b2e3752dfa88398743048c00ac56e92d0beec4e8df9912b12
        exitCode: 143
        finishedAt: "2026-05-30T03:51:12Z"
        reason: Error
        startedAt: "2026-05-30T03:47:45Z"
```

Failure forced due to argo deadline:
```yaml
- name: ARGO_DEADLINE
  value: "2026-05-30T03:51:10Z"
```

#### >1hr wait example:
Screenshot 2026-05-30 at 12 13 03 AM

Pod fails after 1 of executing, no pods are retried, the workflow fails immediately 

#### Compiled retryPolicy with `backoff_max_duration='4h'`
```yaml
retryPolicy:
  backoffMaxDuration: 14400s   # ← '4h' preserved
  maxRetryCount: 1
```

#### 50 minute wait example:
Screenshot 2026-05-31 at 10 34 19 AM

Initial pod fails after 1hr, retry pod is spun up as expected
argo deadline on retry pod (4hrs after initial execution time):
```yaml
    - name: ARGO_DEADLINE
      value: "2026-05-31T17:20:36Z"
```

#### 61 minute wait example:
Screenshot 2026-05-31 at 10 45 14 AM
(initial pod ran past 1hr threshold, failed, pod was still retried due to the 4hr limit increase)
argo deadline on retry pod (4hrs after initial execution time):
```yaml
- name: ARGO_DEADLINE
  value: "2026-05-31T17:20:59Z"
```

**Checklist:**

argoWF when my pod attempts to retry after an hour:
Screenshot 2026-05-31 at 11 00 31 AM

### DIFF
--- a/api/v2alpha1/go/pipelinespec/pipeline_spec.pb.go
+++ b/api/v2alpha1/go/pipelinespec/pipeline_spec.pb.go
@@ -4321,8 +4321,7 @@ type PipelineTaskSpec_RetryPolicy struct {
 	// unspecified, will default to 2.
 	BackoffFactor float64 `protobuf:"fixed64,3,opt,name=backoff_factor,json=backoffFactor,proto3" json:"backoff_factor,omitempty"`
 	// The maximum duration during which the task will be retried according to
-	// the backoff strategy. Max allowed is 1 hour - higher value will be capped
-	// to this limit. If unspecified, will set to 1 hour.
+	// the backoff strategy. If unspecified, will set to 1 hour.
 	BackoffMaxDuration *durationpb.Duration `protobuf:"bytes,4,opt,name=backoff_max_duration,json=backoffMaxDuration,proto3" json:"backoff_max_duration,omitempty"`
 	unknownFields      protoimpl.UnknownFields
 	sizeCache          protoimpl.SizeCache
--- a/api/v2alpha1/pipeline_spec.proto
+++ b/api/v2alpha1/pipeline_spec.proto
@@ -592,8 +592,7 @@ message PipelineTaskSpec {
     double backoff_factor = 3;
 
     // The maximum duration during which the task will be retried according to
-    // the backoff strategy. Max allowed is 1 hour - higher value will be capped
-    // to this limit. If unspecified, will set to 1 hour.
+    // the backoff strategy. If unspecified, will set to 1 hour.
     google.protobuf.Duration backoff_max_duration = 4;
   }
 
--- a/sdk/python/kfp/compiler/compiler_test.py
+++ b/sdk/python/kfp/compiler/compiler_test.py
@@ -1816,7 +1816,7 @@ class TestSetRetryCompilation(unittest.TestCase):
         self.assertEqual(retry_policy.max_retry_count, 3)
         self.assertEqual(retry_policy.backoff_duration.seconds, 30)
         self.assertEqual(retry_policy.backoff_factor, 1.0)
-        self.assertEqual(retry_policy.backoff_max_duration.seconds, 3600)
+        self.assertEqual(retry_policy.backoff_max_duration.seconds, 10800)
 
 
 from google.protobuf import json_format
--- a/sdk/python/kfp/dsl/pipeline_task.py
+++ b/sdk/python/kfp/dsl/pipeline_task.py
@@ -617,7 +617,7 @@ class PipelineTask:
             num_retries : Number of times to retry on failure.
             backoff_duration: Number of seconds to wait before triggering a retry. Defaults to ``'0s'`` (immediate retry).
             backoff_factor: Exponential backoff factor applied to ``backoff_duration``. For example, if ``backoff_duration="60"`` (60 seconds) and ``backoff_factor=2``, the first retry will happen after 60 seconds, then again after 120, 240, and so on. Defaults to ``2.0``.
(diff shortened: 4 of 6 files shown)


ntny:
@JerT33  Good point, this is a real issue.
I think changing current backoff_max_duration is tricky because SDK already writes 3600s into IR by default. So [argocompiler](https://github.com/kubeflow/pipelines/blob/master/backend/src/v2/compiler/argocompiler/container.go#L667) cannot distinguish if user really set 1h or if it is just the old default.
Maybe we should deprecate this parameter and add a new optional one. Then Argo compiler can just skip setting maxDuration when the new field is not set, and set only if user explicitly asks for it.

@zazulam @droctothorpe  hi! could you also take a look? This may be controversial, so I’d like to get a second opinion

JerT33:
> @JerT33 Good point, this is a real issue. I think changing current backoff_max_duration is tricky because SDK already writes 3600s into IR by default. So [argocompiler](https://github.com/kubeflow/pipelines/blob/master/backend/src/v2/compiler/argocompiler/container.go#L667) cannot distinguish if user really set 1h or if it is just the old default. Maybe we should deprecate this parameter and add a new optional one. Then Argo compiler can just skip setting maxDuration when the new field is not set, and set only if user explicitly asks for it.
> 
> @zazulam @droctothorpe hi! could you also take a look? This may be controversial, so I’d like to get a second opinion

@ntny thanks for the feedback!
Yeah this was an option I considered as well. The thought here is that this implementation won't change the current default behavior, but it would allow users to override the 1hr max duration if needed for a longer running component. It seems like this issue should have been present since 2022 when retries were first implemented (#7867), so I wasn't sure how much we would want to change the existing default behavior. Agreed, would love some extra opinions here!

ntny:
> > @JerT33 Good point, this is a real issue. I think changing current backoff_max_duration is tricky because SDK already writes 3600s into IR by default. So [argocompiler](https://github.com/kubeflow/pipelines/blob/master/backend/src/v2/compiler/argocompiler/container.go#L667) cannot distinguish if user really set 1h or if it is just the old default. Maybe we should deprecate this parameter and add a new optional one. Then Argo compiler can just skip setting maxDuration when the new field is not set, and set only if user explicitly asks for it.
> > @zazulam @droctothorpe hi! could you also take a look? This may be controversial, so I’d like to get a second opinion
> 
> @ntny thanks for the feedback! Yeah this was an option I considered as well. The thought here is that this implementation won't change the current default behavior, but it would allow users to override the 1hr max duration if needed for a longer running component. It seems like this issue should have been present since 2022 when retries were first implemented (#7867), so I wasn't sure how much we would want to change the existing default behavior. Agreed, would love some extra opinions here!

@JerT33  Thanks, that makes sense. I agree that this is the most practical approach if we want to keep the change small and maintain backward compatibility.
The API semantics could probably be revisited separately in the future, but that's a much larger discussion than this PR. I'm happy with this direction.

For a possible follow-up, I could imagine something along these lines:

* Deprecate the current backoff_max_duration parameter. The current name is somewhat misleading because it defines the maximum duration of all retries for a component combined rather than the duration of an individual retry. In addition, because the SDK serializes a default, downstream components cannot distinguish between an explicit user choice and a historical default.
* Introduce a new optional parameter with a name that more clearly describes the total retry window semantics.
* Allow platform administrators to configure a global default retry window at the ML Pipeline API Server deployment level instead of having it effectively defined by the SDK. This would allow different deployments to choose defaults/or just disable default deadline that make sense for their environment.
* Use the platform default whenever the component-level value is not explicitly specified.
* If a user explicitly configures a retry window for a component, that value should override the platform default.

The main motivation would be to separate explicit user intent from historical SDK defaults and make platform-wide policy easier to evolve over time.

ntny:
@lgtm

droctothorpe:
Thanks for tackling this, @JerT33! And thanks for the review, @ntny!

/approve
/lgtm

PR #13451, opened 2026-05-29
chore(deps): driver and launcher Go builder to 1.26.3

jeffspahr:
## Summary

- Bump the `kfp-driver` and `kfp-launcher` builder images from `golang:1.26.2-alpine` to `golang:1.26.3-alpine`.
- Keep the final runtime image unchanged at `alpine:3.21`.
- Pin the updated builder tag to Docker manifest digest `sha256:91eda9776261207ea25fd06b5b7fed8d397dd2c0a283e77f2ab6e91bfa71079d`.

## Context

This is part of kubeflow/pipelines#13449. The 2.16.1 `kfp-driver` and `kfp-launcher` scans reported Go stdlib vulnerabilities from binaries built with an older Go toolchain. Go 1.26.3 is the current Go 1.26 patch release and includes the relevant security fixes.

This PR only addresses the Go toolchain side. The remaining third-party module findings are covered by the relevant Dependabot PRs.

## Verification

- `git diff --check -- backend/Dockerfile.driver backend/Dockerfile.launcher`
- `docker build -f backend/Dockerfile.driver -t kfp-driver:go1.26.3-test .`
- `docker build -f backend/Dockerfile.launcher -t kfp-launcher:go1.26.3-test .`
- `trivy image --scanners vuln --format json kfp-driver:go1.26.3-test`
- `trivy image --scanners vuln --format json kfp-launcher:go1.26.3-test`

Local Trivy scans on the rebuilt images no longer report `stdlib`; remaining findings are third-party Go modules covered by the reviewed Dependabot PRs.

PR #13454, opened 2026-05-30
ci: stabilize SeaweedFS resource requests

jeffspahr:
## Summary
- add a shared CI-only SeaweedFS resource request patch
- apply it to standalone and multi-user CI overlays that deploy SeaweedFS
- keep Ginkgo parallelism unchanged

## Context
PR #13451 exposed repeated Argo wait-container artifact/log upload failures to SeaweedFS, for example `failed to put file ... seaweedfs ... i/o timeout`.

Validated on PR #13451 before splitting this into a separate PR:
- `E2EParallelNested cacheEnabled=false argoVersion=v4.0.4` passed with normal 3-worker parallelism
- `E2ECritical pod_to_pod_tls_enabled=true` passed
- `Critical Scenario Multi User cacheEnabled=false` passed
- full test matrix was green except Tide pending

## Local verification
- `git diff --check`
- rendered standalone overlays: default, cache-disabled, cache-disabled-proxy, proxy, tls-enabled
- rendered multi-user overlays: default, cache-disabled, artifact-proxy
- verified rendered SeaweedFS requests are `cpu: 250m` and `memory: 256Mi`

### DIFF
--- a/.github/resources/manifests/base/seaweedfs-resources.yaml
+++ b/.github/resources/manifests/base/seaweedfs-resources.yaml
@@ -1,0 +1,13 @@
+apiVersion: apps/v1
+kind: Deployment
+metadata:
+  name: seaweedfs
+spec:
+  template:
+    spec:
+      containers:
+        - name: seaweedfs
+          resources:
+            requests:
+              cpu: 250m
+              memory: 256Mi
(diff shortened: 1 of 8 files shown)


juliusvonkohout:
Thank you @jeffspahr , I think we can make it a bit easier and consistent.
@danish9039 please revert the CI only changes and instead double it here https://github.com/kubeflow/pipelines/blob/a831e01c1e055134e9b77116728f98dba9de560e/manifests/kustomize/third-party/seaweedfs/base/seaweedfs/seaweedfs-deployment.yaml#L145 from 32m/128mb to 64m and 256 Mb. That way CI and real life applications do not deviate.

PR #13456, opened 2026-05-30
Improve component retry behavior

ntny:

Previously, the Argo compiler did not set a retry expression when compiling retryStrategy. According to Argo retry semantics, when neither retryPolicy nor expression is set, Argo defaults to OnFailure, which retries only main-container failures: https://argo-workflows.readthedocs.io/en/latest/retries

This means KFP tasks were not retried for Argo/Kubernetes node-level failures (for example, when a pod was deleted after a node failure) and therefore were not automatically rescheduled onto another node.

This change makes the retry condition explicit, retries both Failed and Error attempts, and continues to skip OOMKilled failures.

**Checklist:**


juliusvonkohout:
see also https://github.com/kubeflow/pipelines/pull/13467

ntny:
Duplicates https://github.com/kubeflow/pipelines/pull/13485

PR #13472, opened 2026-06-01
fix(backend): inject user identity header in scheduled-workflow-controller for multi-user auth

jaewak:

*Problem*

In multi-user mode, the scheduled-workflow-controller calls `CreateRun` via gRPC to the API server when "Always use latest pipeline version" is enabled on a recurring run. The controller attaches an `Authorization: Bearer <token>` header (when an SA token volume is mounted), but it never sets the user identity header (e.g. `kubeflow-userid`).

The API server's `HTTPHeaderAuthenticator` requires this header and rejects the request with:

> Unauthenticated: Request header error: there is no user identity header

This was masked in clusters that mount a service-account token volume (where `TokenReviewAuthenticator` succeeds), but customers without that volume see recurring runs silently fail to create child runs.

*Solution*

Add two new opt-in CLI flags to the scheduled-workflow-controller:

| Flag | Description | Default |
| ---- | ----------- | ------- |
| `--userIdentityHeader` | User identity gRPC metadata key (e.g. `kubeflow-userid`) | `""` (disabled) |
| `--userIdentityValue` | Value for the metadata key (e.g. the controller's SA identity) | `""` (disabled) |

When both flags are set, the controller injects the key/value into the outgoing gRPC metadata alongside the existing Bearer token. When either is empty, nothing is injected — preserving backward compatibility for single-user deployments.

The configured key is normalized to lowercase and validated against gRPC metadata key rules at controller startup, so a malformed `--userIdentityHeader` fails fast with a clear error instead of causing failed requests during reconciliation.

The multi-user kustomize deployment patch is updated to pass these flags using the controller's own service account identity, derived from the pod's namespace via the Kubernetes downward API.

*Changes*

- `backend/src/crd/controller/scheduledworkflow/controller.go` — Added `userIdentityHeader`/`userIdentityValue` fields to the `Controller` struct; wired into `NewController` (which now normalizes and validates the key, failing fast on an invalid one); inject the key/value into outgoing gRPC metadata in `submitNewWorkflowIfNotAlreadySubmitted` after the existing Bearer token block.
- `backend/src/crd/controller/scheduledworkflow/main.go` — Added `flag.StringVar` registrations for the two new flags (help text clarifies the value is injected as outgoing gRPC metadata); passed to `NewController`.
- `manifests/kustomize/base/installs/multi-user/scheduled-workflow/deployment-patch.yaml` — Added `NAMESPACE_SELF` env var (downward API) and `args:` with `--userIdentityHeader=kubeflow-userid` and `--userIdentityValue=system:serviceaccount:$(NAMESPACE_SELF):ml-pipeline-scheduledworkflow`.
- `backend/src/crd/controller/scheduledworkflow/controller_auth_header_test.go` (new) — Unit tests covering:
  - Injection when both fields are set (key/value present in outgoing metadata).
  - No injection when both fields are empty.
  - No injection when only one field is set.
  - Coexistence with the existing `Authorization: Bearer` token.
  - Metadata key normalization/validation (lowercasing, accepted and rejected characters).
  - `NewController` rejecting a malformed key.

**Checklist:**

### DIFF
--- a/backend/src/crd/controller/scheduledworkflow/controller.go
+++ b/backend/src/crd/controller/scheduledworkflow/controller.go
@@ -18,6 +18,7 @@ import (
 	"context"
 	"encoding/json"
 	"fmt"
+	"regexp"
 	"strconv"
 	"strings"
 	"time"
@@ -114,10 +115,30 @@ type Controller struct {
 	// This will be nil when authentication is not enabled.
 	tokenSrc transport.ResettableTokenSource
 
-	// userIdentityHeader is the header name for user identity in multi-user mode (e.g. "kubeflow-userid").
+	// userIdentityHeader is the outgoing gRPC metadata key for user identity in multi-user mode
+	// (e.g. "kubeflow-userid"). It is normalized to lowercase and validated when the controller is created.
 	userIdentityHeader string
-	// userIdentityValue is the value to set for the user identity header.
+	// userIdentityValue is the value to set for the user identity gRPC metadata key.
 	userIdentityValue string
+}
+
+// gRPC metadata keys are limited to lowercase ASCII letters, digits, and the
+// characters '-', '_', and '.'. See google.golang.org/grpc/metadata and the
+// HTTP/2 header field rules. Keys outside this set can cause outgoing requests
+// to fail at the transport layer, so we validate the configured key up front.
+var validGRPCMetadataKey = regexp.MustCompile(`^[a-z0-9._-]+$`)
+
+// normalizeAndValidateMetadataKey lowercases the given gRPC metadata key (matching
+// grpc-go's own normalization) and verifies it only contains characters allowed in
+// a gRPC metadata key. It returns the normalized key or an error describing why the
+// key is invalid, allowing callers to fail fast at startup instead of risking failed
+// requests during reconciliation.
+func normalizeAndValidateMetadataKey(key string) (string, error) {
+	normalized := strings.ToLower(key)
+	if !validGRPCMetadataKey.MatchString(normalized) {
+		return "", fmt.Errorf("must match %s (lowercase letters, digits, '-', '_', '.')", validGRPCMetadataKey.String())
+	}
+	return normalized, nil
 }
 
 // NewController returns a new sample controller
@@ -134,6 +155,17 @@ func NewController(
 	userIdentityHeader string,
 	userIdentityValue string,
 ) (*Controller, error) {
+	// Normalize and validate the user identity metadata key up front so the
+	// controller fails fast at startup rather than risking failed requests
+	// during reconciliation when an invalid key reaches the gRPC transport.
+	if userIdentityHeader != "" {
+		normalizedHeader, err := normalizeAndValidateMetadataKey(userIdentityHeader)
+		if err != nil {
+			return nil, fmt.Errorf("invalid userIdentityHeader %q: %w", userIdentityHeader, err)
+		}
+		userIdentityHeader = normalizedHeader
+	}
+
 	// obtain references to shared informers
 	swfInformer := swfInformerFactory.Scheduledworkflow().V1beta1().ScheduledWorkflows()
 
--- a/backend/src/crd/controller/scheduledworkflow/controller.go
+++ b/backend/src/crd/controller/scheduledworkflow/controller.go
@@ -113,6 +113,11 @@ type Controller struct {
 	// tokenSrc provides a way to get the latest refreshed token when authentication to the REST API server is enabled.
 	// This will be nil when authentication is not enabled.
 	tokenSrc transport.ResettableTokenSource
+
+	// userIdentityHeader is the header name for user identity in multi-user mode (e.g. "kubeflow-userid").
+	userIdentityHeader string
+	// userIdentityValue is the value to set for the user identity header.
+	userIdentityValue string
 }
 
 // NewController returns a new sample controller
@@ -126,6 +131,8 @@ func NewController(
 	time commonutil.TimeInterface,
 	location *time.Location,
 	tokenSrc transport.ResettableTokenSource,
+	userIdentityHeader string,
+	userIdentityValue string,
 ) (*Controller, error) {
 	// obtain references to shared informers
 	swfInformer := swfInformerFactory.Scheduledworkflow().V1beta1().ScheduledWorkflows()
@@ -148,9 +155,11 @@ func NewController(
 		workflowClient: client.NewWorkflowClient(workflowClientSet, executionInformer),
 		workqueue: workqueue.NewNamedRateLimitingQueue(
 			workqueue.NewItemExponentialFailureRateLimiter(DefaultJobBackOff, MaxJobBackOff), swfregister.Kind),
-		time:     time,
-		location: location,
-		tokenSrc: tokenSrc,
+		time:               time,
+		location:           location,
+		tokenSrc:           tokenSrc,
+		userIdentityHeader: userIdentityHeader,
+		userIdentityValue:  userIdentityValue,
 	}
 
 	log.Info("Setting up event handlers")
@@ -608,6 +617,11 @@ func (c *Controller) submitNewWorkflowIfNotAlreadySubmitted(
 		}
 
 		ctx = metadata.AppendToOutgoingContext(ctx, "Authorization", "Bearer "+token.AccessToken)
+	}
+
+	// Inject user identity header for multi-user mode authorization.
+	if c.userIdentityHeader != "" && c.userIdentityValue != "" {
+		ctx = metadata.AppendToOutgoingContext(ctx, c.userIdentityHeader, c.userIdentityValue)
 	}
 
 	var runtimeConfig *api.RuntimeConfig
--- a/backend/src/crd/controller/scheduledworkflow/controller.go
+++ b/backend/src/crd/controller/scheduledworkflow/controller.go
@@ -138,6 +138,9 @@ func normalizeAndValidateMetadataKey(key string) (string, error) {
 	if !validGRPCMetadataKey.MatchString(normalized) {
 		return "", fmt.Errorf("must match %s (lowercase letters, digits, '-', '_', '.')", validGRPCMetadataKey.String())
 	}
+	if strings.HasPrefix(normalized, "grpc-") {
+		return "", fmt.Errorf("keys beginning with \"grpc-\" are reserved by the gRPC runtime and will be silently dropped")
+	}
 	return normalized, nil
 }
 
--- a/backend/src/crd/controller/scheduledworkflow/controller_auth_header_test.go
+++ b/backend/src/crd/controller/scheduledworkflow/controller_auth_header_test.go
@@ -228,3 +228,56 @@ func TestUserIdentityHeader_CoexistsWithBearerToken(t *testing.T) {
 	assert.Equal(t, []string{"system:serviceaccount:kubeflow:ml-pipeline-scheduledworkflow"},
 		md.Get("kubeflow-userid"))
 }
+
+// TestNormalizeAndValidateMetadataKey verifies that gRPC metadata keys are
+// lowercased and validated, so the controller can fail fast at startup on a
+// malformed --userIdentityHeader instead of failing requests at runtime.
+func TestNormalizeAndValidateMetadataKey(t *testing.T) {
+	tests := []struct {
+		name      string
+		input     string
+		want      string
+		expectErr bool
+	}{
+		{name: "already valid", input: "kubeflow-userid", want: "kubeflow-userid"},
+		{name: "uppercase normalized to lowercase", input: "Kubeflow-UserID", want: "kubeflow-userid"},
+		{name: "digits, dot, underscore allowed", input: "x-user_id.v2", want: "x-user_id.v2"},
+		{name: "space is invalid", input: "kubeflow userid", expectErr: true},
+		{name: "colon is invalid", input: "kubeflow:userid", expectErr: true},
+		{name: "empty is invalid", input: "", expectErr: true},
+	}
+
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			got, err := normalizeAndValidateMetadataKey(tt.input)
+			if tt.expectErr {
+				require.Error(t, err)
+				assert.Empty(t, got)
+				return
+			}
+			require.NoError(t, err)
+			assert.Equal(t, tt.want, got)
+		})
+	}
+}
+
+// TestNewController_InvalidUserIdentityHeader verifies that NewController fails
+// fast when the configured user identity metadata key is malformed. Validation
+// happens before any client or informer is used, so nil dependencies are fine.
+func TestNewController_InvalidUserIdentityHeader(t *testing.T) {
+	_, err := NewController(
+		nil, // kubeClientSet
+		nil, // swfClientSet
+		nil, // workflowClientSet
+		nil, // runClient
+		nil, // swfInformerFactory
+		nil, // executionInformer
+		nil, // time
+		nil, // location
+		nil, // tokenSrc
+		"invalid header",
+		"some-value",
+	)
+	require.Error(t, err)
+	assert.Contains(t, err.Error(), "invalid userIdentityHeader")
+}
--- a/backend/src/crd/controller/scheduledworkflow/controller_auth_header_test.go
+++ b/backend/src/crd/controller/scheduledworkflow/controller_auth_header_test.go
@@ -1,0 +1,230 @@
+// Copyright 2024 The Kubeflow Authors
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+// https://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package main
+
+import (
+	"context"
+	"testing"
+	"time"
+
+	api "github.com/kubeflow/pipelines/backend/api/v2beta1/go_client"
+	"github.com/kubeflow/pipelines/backend/src/crd/controller/scheduledworkflow/client"
+	util "github.com/kubeflow/pipelines/backend/src/crd/controller/scheduledworkflow/util"
+	swfapi "github.com/kubeflow/pipelines/backend/src/crd/pkg/apis/scheduledworkflow/v1beta1"
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+	"golang.org/x/oauth2"
+	"google.golang.org/grpc"
+	"google.golang.org/grpc/metadata"
+	"google.golang.org/protobuf/types/known/emptypb"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+// fakeRunServiceClient captures the context passed to CreateRun so tests can
+// inspect outgoing gRPC metadata.
+type fakeRunServiceClient struct {
+	capturedCtx context.Context
+}
+
+func (f *fakeRunServiceClient) CreateRun(ctx context.Context, in *api.CreateRunRequest, opts ...grpc.CallOption) (*api.Run, error) {
+	f.capturedCtx = ctx
+	return &api.Run{DisplayName: "fake-run"}, nil
+}
+
+func (f *fakeRunServiceClient) GetRun(ctx context.Context, in *api.GetRunRequest, opts ...grpc.CallOption) (*api.Run, error) {
+	return nil, nil
+}
+
+func (f *fakeRunServiceClient) ListRuns(ctx context.Context, in *api.ListRunsRequest, opts ...grpc.CallOption) (*api.ListRunsResponse, error) {
+	return nil, nil
+}
+
+func (f *fakeRunServiceClient) ArchiveRun(ctx context.Context, in *api.ArchiveRunRequest, opts ...grpc.CallOption) (*emptypb.Empty, error) {
+	return nil, nil
+}
+
+func (f *fakeRunServiceClient) UnarchiveRun(ctx context.Context, in *api.UnarchiveRunRequest, opts ...grpc.CallOption) (*emptypb.Empty, error) {
+	return nil, nil
+}
+
+func (f *fakeRunServiceClient) DeleteRun(ctx context.Context, in *api.DeleteRunRequest, opts ...grpc.CallOption) (*emptypb.Empty, error) {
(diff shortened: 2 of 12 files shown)


manifests/kustomize/base/installs/multi-user/scheduled-workflow/deployment-patch.yaml
```diff
@@ -11,3 +11,10 @@ spec:
         - name: NAMESPACE
           value: '' # Empty namespace let viewer controller watch all namespaces
           valueFrom: null # HACK: https://github.com/kubernetes-sigs/kustomize/issues/2606
+        - name: NAMESPACE_SELF
+          valueFrom:
+            fieldRef:
+              fieldPath: metadata.namespace
+        args:
+        - --userIdentityHeader=kubeflow-userid
+        - --userIdentityValue=system:serviceaccount:$(NAMESPACE_SELF):ml-pipeline-scheduledworkflow
```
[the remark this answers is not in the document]
jaewak: The patch parses correctly. args lands at the container level
  as a sibling of env (container keys = ["name", "env", "args"]), and env contains only NAMESPACE +
  NAMESPACE_SELF.

backend/src/crd/controller/scheduledworkflow/controller.go
```diff
@@ -610,6 +619,11 @@ func (c *Controller) submitNewWorkflowIfNotAlreadySubmitted(
 		ctx = metadata.AppendToOutgoingContext(ctx, "Authorization", "Bearer "+token.AccessToken)
 	}
 
+	// Inject user identity header for multi-user mode authorization.
+	if c.userIdentityHeader != "" && c.userIdentityValue != "" {
+		ctx = metadata.AppendToOutgoingContext(ctx, c.userIdentityHeader, c.userIdentityValue)
+	}
```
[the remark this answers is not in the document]
jaewak: addressed comment.

droctothorpe:
I found two backward-compatibility concerns in the new multi-user manifest:

- `manifests/kustomize/base/installs/multi-user/scheduled-workflow/deployment-patch.yaml` now hardcodes `USER_IDENTITY_HEADER=kubeflow-userid` and an unprefixed identity value, but the API server already allows both the user-id header name and prefix to be customized via `KUBEFLOW_USERID_HEADER` / `KUBEFLOW_USERID_PREFIX`.
  - **Why this matters:** existing multi-user installs that customized either setting will not upgrade coherently, so recurring runs can still fail after this change because the controller and API server are no longer using the same auth contract.
  - Related refs: `backend/src/apiserver/common/config.go`, `manifests/kustomize/base/installs/multi-user/api-service/deployment-patch.yaml`.

- The same manifest also hardcodes `USER_IDENTITY_VALUE=system:serviceaccount:$(NAMESPACE_SELF):ml-pipeline-scheduledworkflow`.
  - **Why this matters:** this only matches the repo’s default service account name. Deployments that already renamed the scheduled-workflow controller service account can start sending the wrong identity header after upgrade. Since the API server tries header auth before token review, that can cause authorization failures even when the bearer token path would otherwise work.
  - Related refs: `manifests/kustomize/base/pipeline/ml-pipeline-scheduledworkflow-deployment.yaml`, `backend/src/apiserver/auth/auth.go`, `backend/src/apiserver/resource/resource_manager.go`.

The controller-side flag defaults look safe, but I don't think the shipped multi-user overlay is backward-compatible until these manifest values are sourced from the same config as the API server, or at least made explicitly overridable for customized installs.

jaewak:
> I found two backward-compatibility concerns in the new multi-user manifest:
> 
> * `manifests/kustomize/base/installs/multi-user/scheduled-workflow/deployment-patch.yaml` now hardcodes `USER_IDENTITY_HEADER=kubeflow-userid` and an unprefixed identity value, but the API server already allows both the user-id header name and prefix to be customized via `KUBEFLOW_USERID_HEADER` / `KUBEFLOW_USERID_PREFIX`.
>   
>   * **Why this matters:** existing multi-user installs that customized either setting will not upgrade coherently, so recurring runs can still fail after this change because the controller and API server are no longer using the same auth contract.
>   * Related refs: `backend/src/apiserver/common/config.go`, `manifests/kustomize/base/installs/multi-user/api-service/deployment-patch.yaml`.
> * The same manifest also hardcodes `USER_IDENTITY_VALUE=system:serviceaccount:$(NAMESPACE_SELF):ml-pipeline-scheduledworkflow`.
>   
>   * **Why this matters:** this only matches the repo’s default service account name. Deployments that already renamed the scheduled-workflow controller service account can start sending the wrong identity header after upgrade. Since the API server tries header auth before token review, that can cause authorization failures even when the bearer token path would otherwise work.
>   * Related refs: `manifests/kustomize/base/pipeline/ml-pipeline-scheduledworkflow-deployment.yaml`, `backend/src/apiserver/auth/auth.go`, `backend/src/apiserver/resource/resource_manager.go`.
> 
> The controller-side flag defaults look safe, but I don't think the shipped multi-user overlay is backward-compatible until these manifest values are sourced from the same config as the API server, or at least made explicitly overridable for customized installs.

Good catches, both addressed in the latest commit.

Header name: Added a pipeline-multi-user-params ConfigMap at the multi-user overlay level that holds KUBEFLOW_USERID_HEADER and KUBEFLOW_USERID_PREFIX as the single source of truth. The api-service, pipelines-ui, and scheduled-workflow deployment patches now all read from it via configMapKeyRef instead of hardcoding the value. Operators who customize the header name patch one ConfigMap and all three components stay in sync automatically.

Service account name: Replaced the hardcoded ml-pipeline-scheduledworkflow in USER_IDENTITY_VALUE with a downward API field ref on spec.serviceAccountName. The pod now resolves its own actual service account name at runtime, so renamed service accounts are handled automatically without any manifest changes.

droctothorpe:
@juliusvonkohout any thoughts on this since it modifies the charts?

manifests/kustomize/base/installs/multi-user/multi-user-params-configmap.yaml
```diff
@@ -0,0 +1,7 @@
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  name: pipeline-multi-user-params
```
juliusvonkohout:
```suggestion
  name: pipeline-multi-user-parameters
```

manifests/kustomize/base/installs/multi-user/kustomization.yaml
```diff
@@ -6,6 +6,7 @@ resources:
 - ../../cache-deployer/cluster-scoped
 - ../generic
 - view-edit-cluster-roles.yaml
+- multi-user-params-configmap.yaml
```
juliusvonkohout:
```suggestion
- multi-user-parameters-configmap.yaml
```

manifests/kustomize/base/installs/multi-user/scheduled-workflow/deployment-patch.yaml
```diff
@@ -11,3 +11,27 @@ spec:
         - name: NAMESPACE
           value: '' # Empty namespace let viewer controller watch all namespaces
```
juliusvonkohout:
```suggestion
          value: '' # Empty namespace lets viewer controller watch all namespaces
```

manifests/kustomize/base/installs/multi-user/scheduled-workflow/deployment-patch.yaml
```diff
@@ -11,3 +11,27 @@ spec:
...
           valueFrom: null # HACK: https://github.com/kubernetes-sigs/kustomize/issues/2606
+        - name: NAMESPACE_SELF
+          valueFrom:
+            fieldRef:
+              fieldPath: metadata.namespace
+        # Inject the controller's own service-account identity as the user identity
+        # header so CreateRun calls pass the API server's multi-user authorization.
+        # The controller maps these env vars to its --userIdentityHeader/--userIdentityValue
```
juliusvonkohout:
```suggestion
        # The controller maps these environment variables to its --userIdentityHeader/--userIdentityValue
```

manifests/kustomize/base/installs/multi-user/scheduled-workflow/deployment-patch.yaml
```diff
@@ -11,3 +11,27 @@ spec:
...
+        # flags (see backend/Dockerfile.scheduledworkflow).
+        # USER_IDENTITY_HEADER is sourced from the shared ConfigMap so it stays in
+        # sync with the API server's KUBEFLOW_USERID_HEADER; operators patch the
```
juliusvonkohout:
```suggestion
        # synchronzation with the API server's KUBEFLOW_USERID_HEADER; operators patch the
```

manifests/kustomize/base/installs/multi-user/pipelines-ui/deployment-patch.yaml
```diff
@@ -29,9 +29,15 @@ spec:
...
-          value: kubeflow-userid
+          valueFrom:
+            configMapKeyRef:
+              name: pipeline-multi-user-params
+              key: KUBEFLOW_USERID_HEADER
         - name: KUBEFLOW_USERID_PREFIX
-          value: ""
+          valueFrom:
+            configMapKeyRef:
+              name: pipeline-multi-user-params
```
juliusvonkohout:
```suggestion
              name: pipeline-multi-user-parameters
```

manifests/kustomize/base/installs/multi-user/pipelines-ui/deployment-patch.yaml
```diff
@@ -29,9 +29,15 @@ spec:
         - name: ENABLE_AUTHZ
           value: 'true'
         - name: KUBEFLOW_USERID_HEADER
-          value: kubeflow-userid
+          valueFrom:
+            configMapKeyRef:
+              name: pipeline-multi-user-params
```
juliusvonkohout:
```suggestion
              name: pipeline-multi-user-parameters
```

manifests/kustomize/base/installs/multi-user/api-service/deployment-patch.yaml
```diff
@@ -12,6 +12,12 @@ spec:
...
-          value: kubeflow-userid
+          valueFrom:
+            configMapKeyRef:
+              name: pipeline-multi-user-params
+              key: KUBEFLOW_USERID_HEADER
         - name: KUBEFLOW_USERID_PREFIX
-          value: ""
+          valueFrom:
+            configMapKeyRef:
+              name: pipeline-multi-user-params
```
juliusvonkohout:
```suggestion
              name: pipeline-multi-user-parameters
```

manifests/kustomize/base/installs/multi-user/api-service/deployment-patch.yaml
```diff
@@ -12,6 +12,12 @@ spec:
             name: pipeline-api-server-config
         env:
         - name: KUBEFLOW_USERID_HEADER
-          value: kubeflow-userid
+          valueFrom:
+            configMapKeyRef:
+              name: pipeline-multi-user-params
```
juliusvonkohout:
```suggestion
              name: pipeline-multi-user-parameters
```

backend/src/crd/controller/scheduledworkflow/controller_auth_header_test.go
```diff
@@ -0,0 +1,283 @@
...
+		nil, // executionInformer
+		nil, // time
+		nil, // location
+		nil, // tokenSrc
+		"invalid header",
+		"some-value",
+	)
+	require.Error(t, err)
+	assert.Contains(t, err.Error(), "invalid userIdentityHeader")
+}
```
juliusvonkohout:
```suggestion
func TestNewController_InvalidUserIdentityHeader(test *testing.T) {
	_, error := NewController(
		nil, // kubeClientSet
		nil, // swfClientSet
		nil, // workflowClientSet
		nil, // runClient
		nil, // swfInformerFactory
		nil, // executionInformer
		nil, // time
		nil, // location
		nil, // tokenSrc
		"invalid header",
		"some-value",
	)
	require.Error(test, error)
	assert.Contains(t, error.Error(), "invalid userIdentityHeader")
}
```

backend/src/crd/controller/scheduledworkflow/controller_auth_header_test.go
```diff
@@ -0,0 +1,283 @@
...
+	controller := &Controller{
+		workflowClient:     client.NewWorkflowClient(&fakeExecutionClient{}, &fakeExecutionInformer{}),
+		runClient:          fakeRun,
+		tokenSrc:           &fakeTokenSource{token: "my-sa-token"},
+		userIdentityHeader: "kubeflow-userid",
+		userIdentityValue:  "system:serviceaccount:kubeflow:ml-pipeline-scheduledworkflow",
+	}
+
+	swf := newTestSWFForAPIPath()
+	submitted, _, err := controller.submitNewWorkflowIfNotAlreadySubmitted(
```
juliusvonkohout, line 215:
```suggestion
	submitted, _, error := controller.submitNewWorkflowIfNotAlreadySubmitted(
```

backend/src/crd/controller/scheduledworkflow/controller_auth_header_test.go
```diff
@@ -0,0 +1,283 @@
...
+		context.Background(), swf, 100, 200)
+
+	require.NoError(t, err)
```
juliusvonkohout:
```suggestion
	require.NoError(t, error)
```

backend/src/crd/controller/scheduledworkflow/controller.go
```diff
@@ -126,7 +152,20 @@ func NewController(
...
+	// controller fails fast at startup rather than risking failed requests
+	// during reconciliation when an invalid key reaches the gRPC transport.
+	if userIdentityHeader != "" {
+		normalizedHeader, err := normalizeAndValidateMetadataKey(userIdentityHeader)
+		if err != nil {
+			return nil, fmt.Errorf("invalid userIdentityHeader %q: %w", userIdentityHeader, err)
+		}
+		userIdentityHeader = normalizedHeader
+	}
+
```
juliusvonkohout, line 171:
```suggestion
	if userIdentityHeader != "" {
		normalizedHeader, error := normalizeAndValidateMetadataKey(userIdentityHeader)
		if err != nil {
			return nil, fmt.Errorf("invalid userIdentityHeader %q: %w", userIdentityHeader, error)
		}
		userIdentityHeader = normalizedHeader
	}

```

juliusvonkohout (changes requested):
Please use proper English spelling and proper long, expressive and pronounceable variable names.

juliusvonkohout (changes requested):
What is the architectural requirement for "TokenReviewAuthenticator". In which case is it not enough ?On which kind of clusters is it not working ?

jaewak:
> Please use proper English spelling and proper long, expressive and pronounceable variable names.

Good call, I revised the abbreviated variable names/comments and fixed spellings in my latest commit. I left the variables`err` as it is, since renaming it to `error` would shadow the interface type with the same name.

backend/src/crd/controller/scheduledworkflow/controller_auth_header_test.go
```diff
@@ -0,0 +1,283 @@
...
+		input     string
+		want      string
+		expectErr bool
+	}{
+		{name: "already valid", input: "kubeflow-userid", want: "kubeflow-userid"},
+		{name: "uppercase normalized to lowercase", input: "Kubeflow-UserID", want: "kubeflow-userid"},
+		{name: "digits, dot, underscore allowed", input: "x-user_id.v2", want: "x-user_id.v2"},
+		{name: "space is invalid", input: "kubeflow userid", expectErr: true},
+		{name: "colon is invalid", input: "kubeflow:userid", expectErr: true},
+		{name: "empty is invalid", input: "", expectErr: true},
```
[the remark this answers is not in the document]
jaewak: Applied suggested change

jaewak:
> What is the architectural requirement for "TokenReviewAuthenticator". In which case is it not enough ?On which kind of clusters is it not working ?

In our case (Capital One), this happens on a multi-user EKS cluster where a customer hit an authorization failure on a biweekly scheduled workflow with "always use latest version" enabled (This is relevant because the authorization fails when the workflow tries to fetch the latest pipeline version from the KFP API). 

This authentication fails because the expected `kubeflow-userid` is missing when the `HTTPHeaderAuthenticator` is run. `TokenReviewAuthenticator` is the alternative that gets used when the `HTTPHeaderAuthenticator` fails to authenticate.

juliusvonkohout (changes requested):
Please do a proper git rebase master and do not use merge from master

manifests/kustomize/base/installs/multi-user/multi-user-parameters-configmap.yaml
```diff
@@ -0,0 +1,7 @@
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  name: pipeline-multi-user-parameters
+data:
+  KUBEFLOW_USERID_HEADER: kubeflow-userid
```
juliusvonkohout: is this in line with what centraldashboard uses ?

jaewak:
> Please do a proper git rebase master and do not use merge from master

Done! I'll git rebase master instead of merging from the UI from now

manifests/kustomize/base/installs/multi-user/multi-user-parameters-configmap.yaml
jaewak: Nope, the central dashboard actually doesn't have the `KUBEFLOW` prefix, and it just uses `USERID_HEADER` (I'm not too familiar with the central dashboard to be honest, but I found [this line](https://github.com/kubeflow/dashboard/blob/main/components/centraldashboard/app/server.ts#L32)). 

The `KUBEFLOW` prefix, however, is expected by the [KFP API server](https://github.com/kubeflow/pipelines/blob/master/backend/src/apiserver/common/config.go#L33).

If you want, I can drop the `KUBEFLOW` prefix, then expose it to the API server with a `configMapKeyRef`

juliusvonkohout:
Sorry for the delay, I will try to get back to the review over the weekend.

manifests/kustomize/base/installs/multi-user/multi-user-parameters-configmap.yaml
juliusvonkohout, line 1: Please move this into the  global "pipeline-install-config" configuration map.

manifests/kustomize/base/installs/multi-user/api-service/deployment-patch.yaml
juliusvonkohout, line 1: Please make these environment variables optional

manifests/kustomize/base/installs/multi-user/pipelines-ui/deployment-patch.yaml
juliusvonkohout, line 1: Please make these environment variables optional

juliusvonkohout (changes requested):
Added comments. In addition, is it possible to reduce the duplications?

jaewak:
> Added comments. In addition, is it possible to reduce the duplications?

Thanks! I did a pass on this. The duplication that originally caused the multi-user auth drift is now gone. `USERID_HEADER` and `USERID_PREFIX` live in a single place (pipeline-install-config), and the api-service, UI, and scheduled-workflow deployments all read from there via `configMapKeyRef` instead of carrying their own copies. Did you have any other suggestions?

juliusvonkohout:
Lets wait for the copilot review and @droctothorpe

manifests/kustomize/base/installs/multi-user/scheduled-workflow/deployment-patch.yaml
```diff
@@ -9,5 +9,32 @@ spec:
...
+        # ConfigMap so it stays in synchronization with the API server's
+        # USERID_HEADER; operators patch the ConfigMap once to update both
+        # deployments. The key is optional so the controller falls back to its own
+        # default header when it is unset.
+        - name: USER_IDENTITY_HEADER
+          valueFrom:
+            configMapKeyRef:
+              name: pipeline-install-config
+              key: USERID_HEADER
+              optional: true
```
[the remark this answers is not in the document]
juliusvonkohout: I want to keep it opt in for now

manifests/kustomize/base/installs/multi-user/api-service/deployment-patch.yaml
```diff
@@ -12,6 +12,14 @@ spec:
...
+              name: pipeline-install-config
+              key: USERID_HEADER
+              optional: true
         - name: KUBEFLOW_USERID_PREFIX
-          value: ""
+          valueFrom:
+            configMapKeyRef:
+              name: pipeline-install-config
+              key: USERID_PREFIX
+              optional: true
```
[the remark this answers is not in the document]
juliusvonkohout: i want to keep it opt in for now

manifests/kustomize/base/installs/multi-user/pipelines-ui/deployment-patch.yaml
```diff
@@ -29,9 +29,17 @@ spec:
         - name: ENABLE_AUTHZ
           value: 'true'
         - name: KUBEFLOW_USERID_HEADER
-          value: kubeflow-userid
+          valueFrom:
+            configMapKeyRef:
+              name: pipeline-install-config
+              key: USERID_HEADER
+              optional: true
```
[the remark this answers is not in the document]
juliusvonkohout: i want to keep it opt in for now

manifests/kustomize/base/installs/multi-user/pipelines-ui/deployment-patch.yaml
```diff
@@ -29,9 +29,17 @@ spec:
...
         - name: KUBEFLOW_USERID_PREFIX
-          value: ""
+          valueFrom:
+            configMapKeyRef:
+              name: pipeline-install-config
+              key: USERID_PREFIX
+              optional: true
```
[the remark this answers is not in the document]
juliusvonkohout: i want to keep it opt in for now

manifests/kustomize/base/installs/multi-user/api-service/deployment-patch.yaml
```diff
@@ -12,6 +12,14 @@ spec:
             name: pipeline-api-server-config
         env:
         - name: KUBEFLOW_USERID_HEADER
-          value: kubeflow-userid
+          valueFrom:
+            configMapKeyRef:
+              name: pipeline-install-config
+              key: USERID_HEADER
+              optional: true
```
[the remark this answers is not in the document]
juliusvonkohout: it should stay opt in for now

droctothorpe:
/approve
/lgtm

Great work, @jaewak! If this is your second (or greater) merged PR, feel free to file a PR to add yourself as a Kubeflow member. Here's an example for reference: https://github.com/kubeflow/internal-acls/pull/924.

You can list @jeffspahr and I as sponsors.

Thanks for reviewing the manifest changes, @juliusvonkohout!

jaewak:
Awesome, thank you. Big thanks to @juliusvonkohout for the review as well!

PR #13473, opened 2026-06-02
feat(backend): enable create and schedule time placeholders

### LINKED ISSUE #13424: [sdk] kfp.dsl placeholders not getting replaced at runtime (v2)
### Environment

*  KFP version: 2.16.1

*  KFP SDK version: 2.13.0

*  All dependencies version: N/A

### Steps to reproduce

Here is a short code snippet of how our team is trying to use the [kfp.dsl](https://kubeflow-pipelines.readthedocs.io/en/stable/source/dsl.html) placeholders. 
```
with dsl.ExitHandler(exit_handler_op):
    entry_handler_op = entry_handler_comp(
        fpst=fp_params_str,
        esp=explicit_string_params,
        pipeline_job_name=dsl.PIPELINE_JOB_NAME_PLACEHOLDER,
        pipeline_job_resource_name=dsl.PIPELINE_JOB_RESOURCE_NAME_PLACEHOLDER,
        pipeline_job_id=dsl.PIPELINE_JOB_ID_PLACEHOLDER,
        pipeline_task_name=dsl.PIPELINE_TASK_NAME_PLACEHOLDER,
        pipeline_task_id=dsl.PIPELINE_TASK_ID_PLACEHOLDER,
        pipeline_task_executor_output_path=dsl.PIPELINE_TASK_EXECUTOR_OUTPUT_PATH_PLACEHOLDER,
        pipeline_task_executor_input=dsl.PIPELINE_TASK_EXECUTOR_INPUT_PLACEHOLDER,
        pipeline_root=dsl.PIPELINE_ROOT_PLACEHOLDER,
        pipeline_job_create_time_utc=dsl.PIPELINE_JOB_CREATE_TIME_UTC_PLACEHOLDER,
        pipeline_job_schedule_time_utc=dsl.PIPELINE_JOB_SCHEDULE_TIME_UTC_PLACEHOLDER,
```
But at runtime some of the placeholders are not populated correct, while some are.
Image

### Expected result

The placeholders should be populated with correct, expected values.

### Materials and Reference

---

droctothorpe:
## Summary

This PR fixes [#13424](https://github.com/kubeflow/pipelines/issues/13424) for the **Kubernetes-backed v2 runtime**.

`dsl.PIPELINE_JOB_CREATE_TIME_UTC_PLACEHOLDER` and `dsl.PIPELINE_JOB_SCHEDULE_TIME_UTC_PLACEHOLDER` were not being populated correctly during Kubernetes-backed execution, even though the backend placeholder resolver already knew how to substitute those values once they were present in `driver.Options`.

This change is intentionally scoped to the Kubernetes-backed path. It does **not** modify the local execution mode.

## Problem

The issue report showed that several v2 placeholders were substituted correctly at runtime, but the pipeline job time placeholders were missing.

In the Kubernetes-backed path, the gap was not in placeholder resolution itself. The resolver already supported:

- `PIPELINE_JOB_CREATE_TIME_UTC_PLACEHOLDER`
- `PIPELINE_JOB_SCHEDULE_TIME_UTC_PLACEHOLDER`

The problem was that the v2 driver never received the data needed to populate those fields.

## Root cause

There were two separate issues:

1. **Missing compiler/runtime wiring for create time**
   - The Argo compiler did not pass workflow creation time into the generated driver templates.
   - As a result, the driver had no value to place into `PipelineJobCreateTimeUTC`.

2. **Schedule time cannot be safely injected from the template in all cases**
   - The exact recurring-run schedule time is stored on the Argo Workflow as the `scheduledworkflows.kubeflow.org/workflowEpoch` label.
   - That label exists for recurring runs, but is absent for ordinary manual runs.
   - A compiler-side fix that directly interpolates that label in the workflow template causes Argo validation to fail for manual runs before the driver ever starts.

### Compiler changes

- Passes `{{workflow.creationTimestamp}}` into both container and DAG driver templates as `--pipeline_job_create_time_utc`.
- Avoids compiler-side interpolation of `workflowEpoch`, because that is not safe for manual runs.

### Driver changes

- Adds runtime resolution for pipeline job times in the v2 driver entrypoint.
- Uses the existing `--run_name` value (`{{workflow.name}}`) to fetch the owning Argo Workflow directly when runtime metadata is needed, instead of first looking up the current Pod.
- Uses the Workflow's `scheduledworkflows.kubeflow.org/workflowEpoch` label when it is present.
- Falls back to the Workflow creation timestamp for manual runs that do not have a schedule label.
- Treats schedule-time metadata lookup as **best effort** when create time is already available from the compiler, so missing Workflow-read RBAC or transient metadata lookup failures fall back to create time instead of failing the task.
- Normalizes the resolved values into the existing UTC fields consumed by placeholder resolution.

### Tests, docs, and generated outputs

- Adds focused unit coverage for:
  - compiler wiring of create-time arguments,
  - driver normalization of raw time inputs,
  - workflow-metadata-based schedule-time fallback behavior,
  - best-effort metadata lookup when only schedule time still needs runtime resolution.
- Adds doc comments in the driver code to explain why runtime workflow metadata lookup is necessary for schedule time.
- Refreshes the workflow compiler goldens under `test_data/compiled-workflows/` to reflect the new create-time driver arg emitted in generated workflows.

## Why the runtime workflow metadata lookup is necessary

`pipeline_job_create_time_utc` is straightforward: Argo exposes workflow creation time directly through `{{workflow.creationTimestamp}}`, so the compiler can safely wire that value into every generated driver template.

`pipeline_job_schedule_time_utc` is different. The exact scheduled time for recurring runs is stored as a Workflow label (`scheduledworkflows.kubeflow.org/workflowEpoch`) rather than as a template-safe timestamp field. That label is not present on manual runs.

If the compiler directly references the label, manual runs fail workflow validation because Argo treats the missing label as an unresolved template expression. Resolving the value at runtime from the Workflow object avoids that failure mode and preserves the correct recurring-run semantics:

- recurring runs get the true scheduled time;
- manual runs fall back to workflow creation time instead of failing.

Because `--run_name` is already the Argo Workflow name, the driver can read the Workflow directly. That avoids an unnecessary Pod lookup while still allowing exact recurring-run schedule resolution when the metadata is available.

### Focused Go test coverage

Ran:

```bash
go test ./backend/src/v2/compiler/argocompiler ./backend/src/v2/cmd/driver ./backend/src/v2/driver
```

### Compiler golden verification

Ran:

```bash
ginkgo -v ./backend/test/compiler
```

The workflow compiler suite initially exposed stale expected compiled workflow YAMLs after the new driver arg was added. The generated goldens under `test_data/compiled-workflows/` were refreshed and the suite then passed locally.

### End-to-end Kubernetes validation

Validated this on a local `orbstack` Kubernetes cluster that initially did **not** have KFP installed.

#### Environment/setup

- Installed standalone KFP onto the local cluster from the repo manifests.
- Port-forwarded the local API server.
- Used a dedicated repro pipeline that prints and hard-fails if any of the following remain unresolved:
  - `pipeline_job_create_time_utc`
  - `pipeline_job_schedule_time_utc`
  - `pipeline_job_resource_name`

#### Reproduced the existing bug before the fix

With stock images, the repro failed on-cluster because:

- `pipeline_job_create_time_utc` was empty
- `pipeline_job_schedule_time_utc` was empty

That confirmed the bug in the Kubernetes-backed runtime path.

#### Deployed and validated the fix

Because the local cluster nodes are `arm64`, I built and deployed local `arm64` images for:

- `kfp-driver`
- `kfp-api-server`

After redeploying the patched images, the same repro pipeline succeeded on-cluster.

Successful validation run:

- run id: `4b54d2e5-d822-4525-ada2-8408c4e0c2ce`
- workflow: `timestamp-placeholder-repro-6jj6t`

Observed task output included concrete resolved values:

- `pipeline_job_create_time_utc=2026-06-01T20:49:51Z`
- `pipeline_job_schedule_time_utc=2026-06-01T20:49:51Z`
- `pipeline_job_resource_name=timestamp-placeholder-repro-6jj6t`

That local validation specifically covered the manual-run case where schedule time must fall back safely instead of depending on a missing Workflow label.

## Behavior notes

- This PR preserves exact schedule-time behavior for recurring runs when `workflowEpoch` is present and readable.
- For manual runs, schedule time safely falls back to workflow creation time.
- If schedule-time metadata lookup fails after create time is already available, the driver falls back to create time instead of failing the task.
- The change is intentionally limited to the Kubernetes-backed v2 runtime and does not alter the local execution path.

## Tradeoffs / considerations

This approach is more correct than always equating schedule time with creation time, but it still introduces some runtime coupling:

- the driver may perform a Workflow metadata lookup when schedule time still needs runtime resolution;
- exact recurring-run schedule time still depends on Workflow metadata availability;
- recurring-run schedule time can degrade to create time when Workflow metadata is unavailable, rather than failing the task outright.

That tradeoff is intentional to preserve recurring-run semantics when possible while keeping manual runs and reduced-RBAC environments from regressing.

## Fixes

Fixes #13424.

### DIFF
--- a/backend/src/v2/cmd/driver/main.go
+++ b/backend/src/v2/cmd/driver/main.go
@@ -23,11 +23,9 @@ import (
 	"time"
 
 	argoclient "github.com/argoproj/argo-workflows/v3/pkg/client/clientset/versioned"
-	workflowcommon "github.com/argoproj/argo-workflows/v3/workflow/common"
 	"github.com/spf13/viper"
 	"google.golang.org/protobuf/encoding/protojson"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
-	"k8s.io/client-go/kubernetes"
 
 	"github.com/kubeflow/pipelines/backend/src/apiserver/config/proxy"
 	"github.com/kubeflow/pipelines/backend/src/common/util"
@@ -151,35 +149,24 @@ func validate() error {
 	return nil
 }
 
-// getCurrentWorkflowMetadata returns the owning Argo Workflow metadata for the
-// current driver pod.
+// getCurrentWorkflowMetadata returns metadata for the Argo Workflow backing the
+// current run.
 //
 // The compiler can safely pass workflow creation time directly via
 // {{workflow.creationTimestamp}}, but recurring-run schedule time is stored in
 // the workflowEpoch label and that label is absent for ad hoc runs. Referencing
 // the label directly from the compiled template causes Argo to reject manual
 // runs before the driver starts, so the driver resolves the label at runtime
-// from the Workflow object instead.
-func getCurrentWorkflowMetadata(ctx context.Context, namespace string) (*metav1.ObjectMeta, error) {
+// from the Workflow object instead. The driver already receives --run_name as
+// {{workflow.name}}, so it can read the Workflow directly without first looking
+// up its own Pod.
+func getCurrentWorkflowMetadata(ctx context.Context, namespace string, workflowName string) (*metav1.ObjectMeta, error) {
+	if workflowName == "" {
+		return nil, fmt.Errorf("workflow name is empty")
+	}
 	restConfig, err := util.GetKubernetesConfig()
 	if err != nil {
 		return nil, fmt.Errorf("failed to initialize kubernetes config for workflow metadata: %w", err)
-	}
-	k8sClient, err := kubernetes.NewForConfig(restConfig)
-	if err != nil {
-		return nil, fmt.Errorf("failed to initialize kubernetes client for workflow metadata: %w", err)
-	}
-	podName, err := config.InPodName()
-	if err != nil {
-		return nil, fmt.Errorf("failed to determine driver pod name: %w", err)
-	}
-	pod, err := k8sClient.CoreV1().Pods(namespace).Get(ctx, podName, metav1.GetOptions{})
-	if err != nil {
-		return nil, fmt.Errorf("failed to retrieve driver pod %q: %w", podName, err)
-	}
-	workflowName := pod.Labels[workflowcommon.LabelKeyWorkflow]
-	if workflowName == "" {
-		return nil, nil
 	}
 	argoClient, err := argoclient.NewForConfig(restConfig)
 	if err != nil {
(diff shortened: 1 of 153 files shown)


droctothorpe:
Addressed the final Copilot concern in `backend/src/v2/cmd/driver/main.go`:
- dropped the Pod lookup entirely and use `--run_name` (`{{workflow.name}}`) to fetch the Workflow directly
- made schedule-time metadata lookup best-effort when create time is already present, so clusters without Workflow read RBAC fall back to create time instead of failing the task
- added unit coverage for the new lookup/fallback behavior

Also reran the focused backend Go tests for `cmd/driver`, `argocompiler`, and `driver`.

droctothorpe:
Closed in favor of https://github.com/kubeflow/pipelines/pull/13474

PR #13496, opened 2026-06-05
feat: add Start Time, End Time and Schedule Status columns to recurring runs table

### LINKED ISSUE #13406: [feature] Show start and end time in the recurring runs table
### Feature Area

/area frontend

I would like to see `start_time` and `end_time` displayed directly in the recurring runs table.

Currently, the recurring runs list shows whether a job is enabled or disabled, but it is not immediately clear whether an enabled recurring run is still effectively active if an `end_time` was configured.

A possible improvement could be to add columns such as:

- Start time
- End time
- Schedule status / Effective status

The schedule status could be derived from the current time and the configured schedule window, for example:

- `Scheduled` — enabled, but `start_time` is in the future
- `Active` — enabled and current time is within the schedule window
- `Expired` — enabled, but `end_time` is set and already passed
- `Disabled` — manually disabled

Alternatively, even just displaying `start_time` and `end_time` in the table would already improve the user experience, because users would not need to open each recurring run individually to understand its actual schedule window.

When managing multiple recurring runs, it is hard to quickly understand which ones are still expected to trigger new runs and which ones have already passed their configured `end_time`.

For example, a recurring run can still be shown as `Enabled` even after its `end_time` has passed. From a scheduler perspective this may be correct, because enabled/disabled is a separate mode from the schedule window. However, from a user and operations perspective this can be confusing.

Users may see an enabled recurring run and assume it is still active, even though it will no longer create new runs because the configured `end_time` is already in the past.

Showing the start and end time directly in the recurring runs table, or adding a derived schedule status, would make it much easier to:

- distinguish active recurring runs from expired ones;
- review temporary recurring runs;
- clean up old recurring jobs;
- avoid confusion during operational checks;
- understand the effective lifecycle of a recurring run without opening each job details page.

Currently, the workaround is to open each recurring run details page or inspect the recurring run through the API/SDK and manually check its `start_time` and `end_time`.

This works, but it is inconvenient when there are many recurring runs. It also makes it easy to miss expired recurring runs that still appear as `Enabled` in the list.

Adding these fields directly to the recurring runs table, or showing a derived schedule status, would make this information available at a glance.

---

Love this idea? Give it a 👍.

Priyanshu-u07:
Adds Start Time, End Time, and Schedule Status columns to the recurring runs table. Currently, the recurring runs list shows only whether a job is enabled or disabled, but it's not immediately clear whether an enabled recurring run is still effectively active if an end_time was configured. Users must click into each run's details to check the schedule window.

**Changes**

**`RecurringRunList.tsx`**
- Added `getScheduleTimes()` helper to extract `start_time / end_time` from cron or periodic schedule triggers.
- Added `getScheduleStatus()` helper to derive a human-readable status from the run's enabled state and trigger time window:

          

    **Active (green)** — enabled, current time is within the schedule window
    **Scheduled (blue)** — enabled, but start_time is in the future
    **Expired (orange)** — enabled, but end_time has passed
    **Disabled (grey)** — manually disabled

- Added `_scheduleStatusCustomRenderer` with color-coded output.
- Added three new columns: `Start Time`, `End Time`, `Schedule Status`.
- Updated `Experiment` column splice index from 3 to 6 to account for new columns.

**`RecurringRunList.test.tsx`**

- Updated existing assertions for new `otherFields` indices and array length.
- Added 14 new tests covering all helpers and renderers.

**Before**
pipelines3

**After**
pipelines4

Closes #13406

### DIFF
--- a/frontend/src/pages/RecurringRunList.test.tsx
+++ b/frontend/src/pages/RecurringRunList.test.tsx
@@ -23,7 +23,11 @@ import { range } from 'lodash';
 import * as Utils from 'src/lib/Utils';
 import TestUtils from 'src/TestUtils';
 import { Apis, JobSortKeys, ListRequest } from 'src/lib/Apis';
-import RecurringRunList, { RecurringRunListProps } from './RecurringRunList';
+import RecurringRunList, {
+  RecurringRunListProps,
+  getScheduleTimes,
+  getScheduleStatus,
+} from './RecurringRunList';
 import { V2beta1RecurringRun, V2beta1RecurringRunStatus } from 'src/apisv2beta1/recurringrun';
 import { color } from 'src/Css';
 
@@ -166,6 +170,9 @@ describe('RecurringRunList', () => {
           'recurring run with id: testrecurringrun1',
           undefined,
           undefined,
+          '-',
+          '-',
+          '-',
           undefined,
           '-',
         ],
@@ -322,7 +329,7 @@ describe('RecurringRunList', () => {
     await loadRecurringRuns({});
 
     expect(props.onError).not.toHaveBeenCalled();
-    expect(lastCustomTableProps.rows[0].otherFields[3]).toEqual({
+    expect(lastCustomTableProps.rows[0].otherFields[6]).toEqual({
       displayName: 'test experiment',
       id: 'test-experiment-id',
     });
@@ -349,7 +356,7 @@ describe('RecurringRunList', () => {
     expect(lastCustomTableProps.columns.map((column: any) => column.label)).not.toContain(
       'Experiment',
     );
-    expect(lastCustomTableProps.rows[0].otherFields).toHaveLength(4);
+    expect(lastCustomTableProps.rows[0].otherFields).toHaveLength(7);
   });
 
   it('renders recurring run trigger in seconds', () => {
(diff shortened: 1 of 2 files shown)


frontend/src/pages/RecurringRunList.tsx
```diff
@@ -30,6 +30,31 @@ import {
...
+  const { startTime, endTime } = getScheduleTimes(recurringRun.trigger);
+  const now = new Date();
+  if (startTime && now < startTime) {
+    return 'Scheduled';
+  }
+  if (endTime && now > endTime) {
+    return 'Expired';
+  }
+  return 'Active';
+}
```
jeffspahr, line 60: Rows can reach this component without a usable status/trigger: the v2 API documents partial-error recurring-run rows with only ID + error, and the masked-list path has the same shape when `getRecurringRun` fails. Because this only checks for `DISABLED`, those rows get a green `Active` schedule status. Please only derive window status for explicitly enabled runs and show `-` for missing/unknown state, with a regression test for that path.

```suggestion
export function getScheduleStatus(recurringRun: V2beta1RecurringRun): string {
  if (recurringRun.status === V2beta1RecurringRunStatus.DISABLED) {
    return 'Disabled';
  }
  if (recurringRun.status !== V2beta1RecurringRunStatus.ENABLED) {
    return '-';
  }

  const { startTime, endTime } = getScheduleTimes(recurringRun.trigger);
  const now = new Date();
  if (startTime && now < startTime) {
    return 'Scheduled';
  }
  if (endTime && now > endTime) {
    return 'Expired';
  }
  return 'Active';
}
```

frontend/src/pages/RecurringRunList.test.tsx
```diff
@@ -439,4 +446,177 @@ describe('RecurringRunList', () => {
...
+          start_time: startTime,
+          end_time: endTime,
+        },
+      },
+    });
+    renderRecurringRunList();
+    await loadRecurringRuns({});
+
+    expect(lastCustomTableProps.rows[0].otherFields[3]).toBe('1/2/2019, 12:34:56 PM');
+    expect(lastCustomTableProps.rows[0].otherFields[4]).toBe('1/2/2019, 12:34:56 PM');
```
jeffspahr, line 467: These assertions still pass if `getScheduleTimes` reads the wrong fields, because the date formatter mock returns the same string for every defined date. Please assert the formatter receives the configured start/end values, or return distinct formatted strings per input. This should be mirrored in the cron test below too.

```suggestion
    expect(formatDateStringSpy.mock.calls.slice(0, 2)).toEqual([[startTime], [endTime]]);
    expect(lastCustomTableProps.rows[0].otherFields[3]).toBe('1/2/2019, 12:34:56 PM');
    expect(lastCustomTableProps.rows[0].otherFields[4]).toBe('1/2/2019, 12:34:56 PM');
```

jeffspahr (changes requested):
Thanks for adding schedule visibility to the recurring runs table. I found one correctness issue where unavailable schedule data can be displayed as active, plus a test assertion gap around the new start/end time columns.

Verification run locally against this head:
- `git diff --check refs/review/pr-13496/base...refs/review/pr-13496/head`
- `npm run test -- RecurringRunList.test.tsx`
- `npm run typecheck`
- `npx prettier --check src/pages/RecurringRunList.tsx src/pages/RecurringRunList.test.tsx`

frontend/src/pages/RecurringRunList.tsx
Priyanshu-u07: Done

frontend/src/pages/RecurringRunList.test.tsx
Priyanshu-u07: Done

Priyanshu-u07:
Thanks for the review @jeffspahr! I've addressed the review comments.PTAL.

PR #13508, opened 2026-06-09
fix(backend): fix RetryRun concurrency conflict handling. Fixes #13507

### LINKED ISSUE #13507: [backend] RetryRun fails with concurrent Argo workflow modification (409 Conflict + already exists)
### Description

When calling `RetryRun` on a recently-failed pipeline run, the API server returns an `InternalServerError` because the `RetryRun` implementation does not handle Kubernetes optimistic concurrency conflicts (`409 Conflict`). The failure is consistently reproducible in E2E tests that immediately retry a failed run.

### Error

```
Failed to retry a run: InternalServerError: Failed to retry run <run-id> due to error updating and creating a workflow.
Update error: Operation cannot be fulfilled on workflows.argoproj.io "<wf-name>": the object has been modified;
please apply your changes to the latest version and try again:
workflows.argoproj.io "<wf-name>" already exists, the server was not able to generate a unique name for the object (code: 13)
```

### Root Cause

`ResourceManager.RetryRun()` in `backend/src/apiserver/resource/resource_manager.go` follows a GET → modify → UPDATE pattern on the Argo `Workflow` object. When the Argo workflow-controller concurrently updates the same object (e.g. finalizing status fields right after a pipeline failure), the `resourceVersion` mismatches and the UPDATE returns a `409 Conflict`.

The fallback CREATE also fails because the original workflow object still exists in Kubernetes under the same name.

Unlike `ReconcileSwfCrs()` in the same file, which explicitly handles `apierrors.IsConflict(err)` with a `continue`/retry loop, `RetryRun()` has no such handling.

### Steps to Reproduce

Reliably triggered by the E2E test:

```
[It] Should reopen MLflow runs on retry and then reflect the retried status
[MLflow, FullRegression, MLflowFailure]
backend/test/end2end/mlflow_e2e_test.go:478
```

The test uses `fail_v2.yaml` (a pipeline designed to fail), waits for `FAILED` state, then immediately calls `RetryRun`. At that point the Argo workflow-controller is still processing the workflow, creating the race window.

### Observed in

- CI run: https://github.com/kubeflow/pipelines/actions/runs/27157942311/job/80166203385
- Job: `End to End Critical Scenario MLflow Tests - K8s v1.34.0 cacheEnabled=false artifactStorage=s3`

### Suggested Fix

Add `apierrors.IsConflict` handling in `RetryRun()`, following the existing pattern in `ReconcileSwfCrs()`:

```go
// Current (no conflict handling)
err = r.getWorkflowClient(namespace).Update(ctx, workflow)
if err != nil {
    return util.NewInternalServerError(err, "Failed to retry run ...")
}

// Suggested: retry on 409 Conflict
for {
    err = r.getWorkflowClient(namespace).Update(ctx, workflow)
    if apierrors.IsConflict(err) {
        // re-fetch and reapply
        workflow, err = r.getWorkflowClient(namespace).Get(ctx, workflow.Name, ...)
        if err != nil { ... }
        // reapply retry mutations
        continue
    }
    break
}
```

Alternatively, consider a Delete + Create approach for stronger idempotency guarantees.

**Labels:** `area/backend`, `kind/bug`

Ayush-kathil:
**Description of the change:**

This PR resolves a concurrency issue in `ResourceManager.RetryRun()` where the API server would return an `InternalServerError` (409 Conflict) when attempting to retry a recently failed pipeline run.

**Root Cause:**
`ResourceManager.RetryRun()` followed a `Get` -> modify -> `Update` pattern on the Argo Workflow object. If the Argo workflow-controller modified the object concurrently (e.g., finalizing status fields just after the failure), the resource version mismatched and the `Update` attempt returned a 409 Conflict. Unlike `ReconcileSwfCrs()` which natively handled `apierrors.IsConflict()`, `RetryRun` lacked this loop.

**Changes:**
- Wrapped the workflow `Get` -> `Update` -> fallback `Create` logic in `backend/src/apiserver/resource/resource_manager.go` in a retry loop.
- Added `apierrors.IsConflict` checks. If an optimistic concurrency conflict occurs, the loop continues, fetching the newest workflow state and re-applying the retry execution spec, mirroring existing idempotency patterns in the codebase.

**Issues fixed:**
Fixes #13507

**Testing:**
- Relying on the E2E test: `MLflow Integration > Failed pipeline + RetryRun with MLflow` which consistently triggered this race condition.


backend/src/apiserver/resource/resource_manager.go
```diff
@@ -1127,23 +1127,46 @@ func (r *ResourceManager) RetryRun(ctx context.Context, runId string) error {
...
-			if createError, ok := createError.(net.Error); ok && createError.Timeout() {
-				return util.NewUnavailableServerError(createError, "Failed to retry run %s due to error creating and updating a workflow - try again later. Update error: %s", runId, updateError.Error())
+	maxRetries := 10
+	var finalErr error
+	for i := 0; i < maxRetries; i++ {
+		if ctx.Err() != nil {
+			return util.NewInternalServerError(ctx.Err(), "Failed to retry run %s due to context cancellation", runId)
+		}
+
+		latestWorkflow, updateError := r.getWorkflowClient(namespace).Get(ctx, newExecSpec.ExecutionName(), v1.GetOptions{})
```
ntny, line 1142: Since this adds conflict handling, should we also verify that the refetched workflow is still in the expecte retryable state (failed or error) before updating it? A rare case this could catch is two retry requests racing: one already moves the workflow to `Running`, and the other then applies `newExecSpec` again after refetching.

This may be out of scope for this PR, but the same idea might be useful for [terminate](https://github.com/kubeflow/pipelines/blob/master/backend/src/apiserver/resource/resource_manager.go#L1043) as well. Since termination is often triggered manually, a slow API server response could result in a second terminate request after the workflow has already reached a terminal state. Maybe this could be considered in a follow-up PR

Ayush-kathil: Thanks @ntny for the review and catching that edge case! I agree that we should check the state after refetching to prevent a race condition where the workflow is already Running. Doing the same state check for terminate makes a lot of sense too.

Since this PR is already approved, I'll create a follow-up PR to handle these state validation checks for both retry and terminate to keep the scope of this one focused. Thanks again for the LGTM!

ntny: > ###
> **[Ayush-kathil](/Ayush-kathil) ** [Jun 16, 2026](#discussion_r3418277821)

Small clarification: this PR has LGTM, but it is not approved yet.
I’m fine with handling the terminate case separately, but the retry state re-check seems closely related to this PR.

Ayush-kathil: @ntny 
Makes total sense! I've gone ahead and pushed the CanRetry() state check directly to this PR. I also took the opportunity to replace the flat time.Sleep() with a context-aware exponential backoff so the conflict loop respects context cancellations properly. I'll leave the terminate logic for a separate follow-up. Let me know if everything looks good for approval!

droctothorpe:
Hi @Ayush-kathil, thank you for your PR! The linked issue (#13507) must be labeled `ready` before your contribution can be reviewed. Please check our [CONTRIBUTING.md](https://github.com/kubeflow/pipelines/blob/master/CONTRIBUTING.md) for details.

PR #13510, opened 2026-06-09
fix(frontend): harden TensorBoard proxy access

droctothorpe:
## Summary

This change hardens TensorBoard access in the frontend by replacing the generic URL-based proxy flow with a dedicated TensorBoard proxy route.

- stop returning internal TensorBoard service URLs to the client
- issue scoped TensorBoard proxy paths and validate them on every proxied request
- use a shared signing secret so scoped proxy paths remain valid across UI-server replicas and restarts
- reject unsafe TensorBoard proxy paths and rate-limit the dedicated proxy route
- deprecate the legacy generic `/_proxy/` endpoint with an explicit gone response
- enforce namespace authorization on proxied TensorBoard traffic
- remove the legacy generic proxy middleware and update the mock backend to match

## Testing

- `cd frontend && npm run typecheck:mock-backend`
- `cd frontend/server && npm test -- app.test.ts configs.test.ts handlers/tensorboard-proxy.test.ts integration-tests/tensorboard.test.ts integration-tests/mock-backend-routes.test.ts`


droctothorpe:
Closing in favor of #13511, which continues this TensorBoard hardening change from the fork branch.

PR #13512, opened 2026-06-09
fix(filter): make k8s CRD backend filter matching case-insensitive

kaikaila:
## Problem

The `matchesFilter()` function used by the k8s CRD pipeline store backend performs case-sensitive string comparisons for `EQ`, `NEQ`, `IN`, and `IS_SUBSTRING` predicates. This is inconsistent with the SQL database backends (MySQL/PostgreSQL), which already apply `LOWER()` wrapping and are therefore case-insensitive.

As a result, a filter like `name = "My-Pipeline"` would match in the database backend but not in the k8s CRD backend.

## Fix

Replace direct string equality and `strings.Contains` calls in `matchesFilter()` with case-insensitive equivalents:

- `EQ` / `NEQ`: `strings.EqualFold`
- `IN`: `strings.EqualFold`
- `IS_SUBSTRING`: `strings.ToLower` on both sides before `strings.Contains`

## Tests

Added case-insensitive test cases for all four affected operators in both `FilterK8sPipelines` (k8s CRD path) and `AddToSelect` / `AddToSelectV1` (SQL path).

## Notes

This fix addresses application-layer behavioral inconsistency — it is distinct from the cross-backend collation ordering differences (e.g. `-` vs `_` sort order between MySQL and PostgreSQL), which are a known database-level limitation and are documented separately.

This commit was originally authored on `feature/postgres-integration` and is being submitted as a standalone fix targeting master.

### DIFF
--- a/backend/src/apiserver/filter/filter.go
+++ b/backend/src/apiserver/filter/filter.go
@@ -241,7 +241,7 @@ func (f *Filter) matchesFilter(getField func(string) interface{}) (bool, error) 
 	for k := range f.eq {
 		fieldVal := fmt.Sprint(getField(k))
 		for _, v := range f.eq[k] {
-			if fieldVal != fmt.Sprint(v) {
+			if !strings.EqualFold(fieldVal, fmt.Sprint(v)) {
 				return false, nil
 			}
 		}
@@ -251,7 +251,7 @@ func (f *Filter) matchesFilter(getField func(string) interface{}) (bool, error) 
 	for k := range f.neq {
 		fieldVal := fmt.Sprint(getField(k))
 		for _, v := range f.neq[k] {
-			if fieldVal == fmt.Sprint(v) {
+			if strings.EqualFold(fieldVal, fmt.Sprint(v)) {
 				return false, nil
 			}
 		}
@@ -283,7 +283,7 @@ func (f *Filter) matchesFilter(getField func(string) interface{}) (bool, error) 
 				return false, nil
 			}
 			for i := 0; i < rv.Len(); i++ {
-				if fieldVal == fmt.Sprint(rv.Index(i).Interface()) {
+				if strings.EqualFold(fieldVal, fmt.Sprint(rv.Index(i).Interface())) {
 					inOne = true
 					break
 				}
@@ -296,9 +296,9 @@ func (f *Filter) matchesFilter(getField func(string) interface{}) (bool, error) 
 
 	// SUBSTRING: all specified substrings must be present.
 	for k := range f.substring {
-		fieldVal := fmt.Sprint(getField(k))
+		lowerFieldVal := strings.ToLower(fmt.Sprint(getField(k)))
 		for _, v := range f.substring[k] {
-			if !strings.Contains(fieldVal, fmt.Sprint(v)) {
+			if !strings.Contains(lowerFieldVal, strings.ToLower(fmt.Sprint(v))) {
 				return false, nil
 			}
 		}
--- a/backend/src/apiserver/filter/filter_test.go
+++ b/backend/src/apiserver/filter/filter_test.go
@@ -722,6 +722,26 @@ func TestFilterK8sPipelines_EQ_NEQ(t *testing.T) {
 	if found {
 		t.Fatalf("expected AND filter not to match when one predicate fails")
 	}
+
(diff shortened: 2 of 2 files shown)


backend/src/apiserver/filter/filter.go
```diff
@@ -298,7 +298,7 @@ func (f *Filter) matchesFilter(getField func(string) interface{}) (bool, error)
 	for k := range f.substring {
 		fieldVal := fmt.Sprint(getField(k))
 		for _, v := range f.substring[k] {
-			if !strings.Contains(fieldVal, fmt.Sprint(v)) {
+			if !strings.Contains(strings.ToLower(fieldVal), strings.ToLower(fmt.Sprint(v))) {
```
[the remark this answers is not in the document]
kaikaila: `IS_SUBSTRING` uses `strings.ToLower` while `EQ/NEQ/IN` use `strings.EqualFold`, which means Unicode case folding edge cases (e.g. ß vs ss) are handled inconsistently between operators. This is a known limitation — MLflow takes the same approach ([re.IGNORECASE for substring](https://github.com/mlflow/mlflow/blob/master/mlflow/utils/search_utils.py#L64), no Unicode fold normalization) and accepts it without further abstraction. Fixing this properly would require a Unicode-aware fold for substring matching, which is out of scope for this PR.

backend/src/apiserver/filter/filter.go
```diff
@@ -241,7 +241,7 @@ func (f *Filter) matchesFilter(getField func(string) interface{}) (bool, error)
 	for k := range f.eq {
 		fieldVal := fmt.Sprint(getField(k))
 		for _, v := range f.eq[k] {
-			if fieldVal != fmt.Sprint(v) {
+			if !strings.EqualFold(fieldVal, fmt.Sprint(v)) {
```
mprahl, line 244: Hey @kaikaila , I'm not seeing this case insensitive behavior in the code. Is it possibly a MySQL behavioral quirk you're experiencing?

```go
func (f *Filter) AddToSelect(sb squirrel.SelectBuilder) squirrel.SelectBuilder {
	for k := range f.eq {
		for _, v := range f.eq[k] {
			m := map[string]interface{}{k: v}
			sb = sb.Where(squirrel.Eq(m)) // plain "="; no LOWER(...)
		}
	}

	for k := range f.neq {
		for _, v := range f.neq[k] {
			m := map[string]interface{}{k: v}
			sb = sb.Where(squirrel.NotEq(m)) // plain "<>"; no LOWER(...)
		}
	}

	// ...

	for k := range f.substring {
		for _, v := range f.substring[k] {
			like := make(squirrel.Like)
			like[k] = fmt.Sprintf("%%%s%%", v)
			sb = sb.Where(like) // plain LIKE; no ILIKE / LOWER(...)
		}
	}

	return sb
}
```

kaikaila: Hey @mprahl, thanks for the feedback! 

To clarify my intent here: I believe KFP's filter API should behave consistently regardless of which pipeline store backend is used. Right now,  the K8s pipeline store case-sensitive collation causes filter behavior to silently differ from MySQL's— and I don't think that's intentional.

So I'd love to get your or the community's input on what the right policy should be:

A. Case-insensitive everywhere (my current fix. align to MySQL behavior, fix K8s pipeline store to match, and postgres in #12379)
B. Case-sensitive everywhere (align to K8s pipeline store / postgres behavior, breaking change for MySQL users)
C. Leave behavior backend-dependent (K8s pipeline store differs from MySQL — users get different filter semantics depending on their deployment)

Happy to implement whichever direction is decided.

For reference, MLflow faced the same problem and they force case-sensitive behavior across all backends by using `BINARY` in MySQL queries. [See here](https://github.com/mlflow/mlflow/blob/4953f67599d183af2436cb9c5ed9c2e403fd65e1/mlflow/utils/search_utils.py#L257).

mprahl: @kaikaila I'm torn because either approach we take it's technically a breaking change for one of the backends. I tend to prefer case insensitive as the least breaking option.

What do you think @droctothorpe @zazulam ?

droctothorpe: I agree with you, Matt.

mprahl (approved):
/lgtm
/approve

PR #13514, opened 2026-06-10
fix: use recursive child traversal for artifact lookup with retryStrategy

Raakshass:
## Description

Fixes the Integration v1 `TestArtifactAPI` 404 failure on `proper-argo-retries`.

### Root Cause

With `templateDefaults.retryStrategy`, the Argo workflow node hierarchy becomes:

The Integration v1 test regex captures `matches[0]` from `status.nodes`, which may be a **StepGroup** node (not a Retry node). The `NodeTypeRetry` guard added in `08fc835` blocks child traversal for non-Retry nodes, causing `FindObjectStoreArtifactKeyOrEmpty` to return empty → 404.

### Fix

- Replace the `NodeTypeRetry` type guard with **bounded recursive traversal** (`maxDepth=3`)
- Resolves artifacts regardless of which node in the hierarchy the test captures
- Covers: direct Pod, Retry → Pod, StepGroup → Retry → Pod

### Testing

All existing unit tests pass. New test added for the `StepGroup → Retry → Pod` path:

Signed-off-by: Siddhant Jain <siddhantjainofficial26@gmail.com>

### DIFF
--- a/backend/src/common/util/workflow.go
+++ b/backend/src/common/util/workflow.go
@@ -776,8 +776,10 @@ func (w *Workflow) SetCannonicalLabels(name string, nextScheduledEpoch int64, in
 }
 
 // FindObjectStoreArtifactKeyOrEmpty looks up the first S3 artifact with the specified artifactName
-// on the specified nodeName. If the node has no outputs (e.g., a retry parent node), it also checks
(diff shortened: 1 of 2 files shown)


Raakshass:
@juliusvonkohout

PR #13515, opened 2026-06-10
ci: add timeout-minutes: 60 to image-build job

kaikaila:
## Summary

- The `image-build` job has no explicit timeout, so GitHub Actions falls back to the 360-minute default.
- This causes stuck builds to hold runners for up to 6 hours before being killed.
- Set `timeout-minutes: 60`, which is ~2× the expected build time for the heaviest images (`apiserver`, `frontend`) and leaves room for the built-in rebuild-on-failure retry step.

Example of a job that ran without bound:
https://github.com/kubeflow/pipelines/actions/runs/27294585775/job/80623821224

## Tests

- [x] Verify the `image-build` job completes within 60 minutes on a normal run
- [x] Confirm no legitimate builds are cut short by the new limit

### DIFF
--- a/.github/workflows/image-builds.yml
+++ b/.github/workflows/image-builds.yml
@@ -3,13 +3,13 @@ on:
   workflow_call:
     outputs:
       IMAGE_PATH:
-        description: 'A description of my output'
+        description: "A description of my output"
         value: ${{ jobs.image-build.outputs.image_path }}
       IMAGE_TAG:
-        description: 'A description of my output'
+        description: "A description of my output"
         value: ${{ jobs.image-build.outputs.image_tag }}
       IMAGE_REGISTRY:
-        description: 'A description of my output'
+        description: "A description of my output"
         value: ${{ jobs.image-build.outputs.registry }}
 
 env:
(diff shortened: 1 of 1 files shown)


mprahl:
@nsingla what are your thoughts on this one?

nsingla:
> @nsingla what are your thoughts on this one?

I think 60min is more than sufficient and if a build is taking longer than 60min, then it deserves to be killed. But I think 60min at the job level mean retry will also be killed if total time exceeds 60min

kaikaila:
> > @nsingla what are your thoughts on this one?
> 
> I think 60min is more than sufficient and if a build is taking longer than 60min, then it deserves to be killed. But I think 60min at the job level mean retry will also be killed if total time exceeds 60min

Hi @nsingla cc @mprahl 
Good point! To be more precise, I moved timeout-minutes: 60 from the job level to the individual build steps (save-image and rebuild), so each attempt gets its own 60-minute budget without the retry being killed by a shared job-level clock. Would that address your concern?

mprahl:
@kaikaila I'm fine with it but a shorter timeout would likely be more useful to fail jobs that are stuck quickly to allow us to retry.

PR #13516, opened 2026-06-11
feat(backend,frontend): surface pod lifecycle failures in the KFP UI. Fixes #12843

khushiiagrawal:
**Description:**

When a pipeline task hits a pod lifecycle failure (e.g. `ImagePullBackOff`, `OOMKilled`, `Unschedulable`), the KFP UI previously showed the task node frozen in a spinning/running state with no error message. Users had to run `kubectl get pods` to understand what was happening, which breaks the Kubernetes abstraction KFP is meant to provide.

This PR fixes the visualization part of #12843. The failure message that Kubernetes reports is now captured, stored, and displayed in the UI exactly the way pipeline script failures are — with a red node and an error banner in the side panel.

**What changed:**

*Backend:*
- `util.NodeStatus` gains a `Message` field; `NodeStatuses()` populates it from Argo's `node.Message`
- Two new columns on the `tasks` table: `LifecycleFailureMessage` (persisted via GORM `AutoMigrate`, no manual migration needed)
- `toModelTasks()` calls `resolveNodeLifecycleMessages()` — a recursive helper that bubbles the pod-level failure message up to the parent task node the UI renders (Argo records the message on the executor child pod, not the task node itself)
- Transient startup messages (`PodInitializing`, `ContainerCreating`) and non-failure node states (`Succeeded`, `Skipped`, `Omitted`) are explicitly filtered so healthy runs produce no false positives
- `toApiPipelineTaskDetail()` exposes `LifecycleFailureMessage` via the existing `error` field on `PipelineTaskDetail`

*Frontend:*
- `ExecutionFlowElementData` gains a `lifecycleError` field
- `updateFlowElementsState()` reads `task_details[].error.message` from the KFP API and sets `lifecycleError` on the matching graph node; if set, the node state is forced to `FAILED` (overriding the MLMD `RUNNING` state that persists because the driver registers the execution before the user container fails)
- `RunDetailsV2` passes `task_details` into the graph even before MLMD data loads
- `RuntimeNodeDetailsV2` shows the "Pod lifecycle failure" banner whenever `lifecycleError` is present

---

**How to verify:**

1. Deploy the API server from this branch and run `npm run build` in `frontend/`

2. **Failure case** — create a pipeline with a non-existent image:
```python
from kfp import dsl, compiler

@dsl.component(base_image="python:99.99.99-nonexistent")
def will_fail(x: int) -> int:
    return x

@dsl.pipeline
def bad_pipeline():
    will_fail(x=1)

[...]

PR #13517, opened 2026-06-12
feat(proposals): add KEP-12843 pod lifecycle failure support and visualization

### LINKED ISSUE #12843: [feature] Enhance support and visualization for pod lifecycle failure in Kubeflow Pipelines

khushiiagrawal:

Adds the proposal for [#12843](https://github.com/kubeflow/pipelines/issues/12843) - surfacing pod lifecycle failures (`ImagePullBackOff`, `OOMKilled`, `Unschedulable`, etc.) in the KFP run details UI.

Currently, when a task pod fails at the Kubernetes level, the task node sits in a green running state forever with no error shown. The Argo node message that describes the failure is dropped during conversion and never reaches the UI.

This KEP proposes:
- Persisting the Argo node failure message on the `Task` model via a new nullable `LifecycleFailureMessage` column (additive schema change, no migration required)
- Returning it through the existing `task_details[].error` field in the v2beta1 `GetRun` response (no new API surface)
- Overriding the node visual state to Failed in the frontend and showing the failure reason in the side panel banner, even when MLMD still reports `RUNNING`

The initial implementation draft is open in [#13516](https://github.com/kubeflow/pipelines/pull/13516).

**Related:** Fixes [#12843](https://github.com/kubeflow/pipelines/issues/12843)


proposals/12843-pod-lifecycle-failure/README.md
```diff
@@ -0,0 +1,301 @@
...
+I submit a healthy pipeline. Every task transitions through normal pod startup states (`PodInitializing`, `ContainerCreating`) before running. None of the task nodes ever shows a lifecycle failure banner. The run completes green.
+
+### Risks and Mitigations
+
+| Risk | Mitigation |
+|---|---|
+| Argo emits `PodInitializing` and `ContainerCreating` on healthy pods, which would create a false-positive banner on every run | Filter these known transient strings in `normalizeLifecycleFailureMessage` before persisting |
+| Argo also emits messages on `Succeeded`, `Skipped`, and `Omitted` nodes (cache hits, conditional skips) | Suppress messages whose node state is non-failure in `nodeLifecycleFailureMessage` |
+| The Argo node graph could in theory contain a cycle | Cycle-safe traversal in `resolveNodeLifecycleMessages` using a recursion-stack guard |
+| A pod that recovers on retry could leave a stale failure message on the row | `patchTask` always overwrites the field with the latest filtered value from Argo, with no preservation logic |
```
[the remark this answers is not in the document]
khushiiagrawal: it was in the earlier version, there is no issue in this one. thanks.

ntny:
I really like the idea of using the message from `AWF status.nodes[].message` instead of simply surfacing pod events.

This also works in cases where the pod was never created at all (for example, when resource quotas are exceeded or Kyverno blocks pod creation due to security policies).

Good catch on filtering transient messages such as PodInitializing and ContainerCreating.

One design question that may be worth adding to the KEP: can Argo node messages be parsed into a more meaningful task lifecycle status in addition to surfacing he raw text message? Most of the useful messages here correspond to standard Kubernetes problems such as `OOMKilled`, image pull errors, resource/scheduling issues, or configuration/admission failures. For unknown or newly introduced failure patterns, the structured status could stay `Unknown` while the raw message would still give users the original diagnostic details.


### Concerns
Reporting latency is one concern. Fast reporting is critical for recurring (cron) runs. Unlike regular runs, cron runs are created in MySQL [through persistenceagent reporting](https://github.com/kubeflow/pipelines/blob/master/backend/src/apiserver/resource/resource_manager.go#L1662).

There is already at least one path that depends on the run row existing before task execution finishes. After a task completes:
- the driver writes a cache entry through `CreateExecutionCache`:
  https://github.com/kubeflow/pipelines/blob/master/backend/src/v2/cacheutils/cache.go#L140-L145
- That goes through `TaskServer.CreateTaskV1`:
  https://github.com/kubeflow/pipelines/blob/master/backend/src/apiserver/server/task_server.go#L36-L50
- And `ResourceManager.CreateTask` first calls `GetRun(t.RunID)`:
  https://github.com/kubeflow/pipelines/blob/master/backend/src/apiserver/resource/resource_manager.go#L991-L995
If a driver starts executing before the persistence agent has reported the cron run (`ScheduledWorkflow`), this path can fail because the run row does not exist yet.
Because of that, any additional processing of lifecycle messages should not significantly slow down the existing persistence agent reporting loop. It would be good to either ensure that this work has negligible impact on reporting latency or keep heavier processing on a separate path.
Could you also add a small performance measurement to the KEP for the persistence-agent reporting path before and after this change?

### Possible future improvements

1. One open question is whether task updates should be deduplicated in the persistence-agent reporting path. Today the agent reports task status snapshots from the workflow, and this proposal extends that existing path with lifecycle failure messages. This is probably fine for the initial implementation,  but it may make sense in the future to avoid sending/upserting updates for tasks whose effective state has not changed, to reduce unnecessary database writes.
Lifecycle messages add one extra nuance here: even when the task state is unchanged, Kubernetes/Argo messages may change in ways that do not provide meaningful new information. For example, an invalid image can alternate between ErrImagePull and ImagePullBackOff. If this becomes noisy, we may want to suppress repeated or semantically equivalent message updates as well.

2. Another possible future improvement would be to distinguish between transient and terminal failures.
For example, OOMKilled is definitely terminal, and it may make sense to immediately transition the task into a Failed state in MySQL. On the other hand, ErrImagePull, ImagePullBackOff, or scheduling-related issues such as FailedScheduling are often transient conditions. In such cases, it may be worth introducing a separate Warning task state in MySQL rather than treating them as either Running or Failed.

Not suggesting this for the current KEP, but it may be worth keeping in mind as a future enhancement.

khushiiagrawal:
@ntny Thanks for the thorough review! I've updated the KEP to address each point:
- Added a structured classification section with the proposed category taxonomy and rationale for deferring it to a follow up
- Added a latency section covering the resolveNodeLifecycleMessages traversal and the cron run race condition, the traversal is in-memory only and doesn't affect the existing DB write path; a before/after measurement will be in the implementation PR
- Added a Future Work section capturing the transient vs terminal distinction, message deduplication, and task update deduplication points

Let me know if anything needs more detail!

khushiiagrawal:
Hii @alyssacgoins  @HumairAK  @mprahl  @james-jwu ,  Would really appreciate your thoughts and review on this. 
Thanks!

proposals/12843-pod-lifecycle-failure/README.md
```diff
@@ -0,0 +1,340 @@
...
+### Goals
+
+1. Capture the Argo node failure message and persist it on the KFP `Task` row.
+2. Expose the message through the existing `task_details[].error` field of the v2beta1 `GetRun` response. No new API surface.
+3. In the UI, when a lifecycle failure is present on a task, override the node visual state to Failed (red), even if MLMD still reports `RUNNING`, and show the failure reason in a banner inside the node detail side panel.
+4. Avoid false positives. Argo emits transient messages like `PodInitializing` during normal pod startup. Healthy runs must not show a failure banner.
+
+### Non-Goals
+
+1. Configurable per-class timeouts (for example, fail after one hour of `ImagePullBackOff`). The original issue mentions this. It is a useful follow-up but it is a separate piece of work and is not covered here.
```
alyssacgoins: Just curious - is there a specific reason why configurable timeouts are out of scope for this feature?

proposals/12843-pod-lifecycle-failure/README.md
```diff
@@ -0,0 +1,340 @@
...
+
+1. `TestNormalizeLifecycleFailureMessage` covers that real failure strings (`Back-off pulling image ...`, `OOMKilled`) pass through unchanged, `PodInitializing` and `ContainerCreating` return empty, and empty input returns empty.
+2. `TestNodeLifecycleFailureMessage` covers that a Failed or Pending node with a real message returns the message, and Succeeded, Skipped, and Omitted nodes return empty even when the message is non-empty.
+3. `TestResolveNodeLifecycleMessages` covers a direct message on a task node, propagation from a child executor node, parent precedence over a child, an empty result when no descendant has a message, a graph with a cycle, and two task nodes that share a descendant.
+
+Backend, in `backend/src/apiserver/storage/task_store_test.go`: existing CRUD tests cover the new column once it is added to the column list. No new test files are required.
+
+Frontend: existing `RunDetailsV2.test.tsx` continues to pass without changes, which is the regression bar. New behavior is exercised through the test cases that already cover `updateFlowElementsState`.
+
+### Manual Verification (E2E)
```
alyssacgoins: The manual verification you've included here is quite useful for verifying changes locally. Can you add some info on how you plan to validate/verify this feature (particularly frontend-wise) through integration testing in CI?

proposals/12843-pod-lifecycle-failure/README.md
```diff
@@ -0,0 +1,340 @@
...
+
+## Migration Strategy
+
+The only schema change is one new nullable column on the `tasks` table, `LifecycleFailureMessage` (stored as `longtext` in MySQL via GORM's `LargeText` type). It is added on API server startup by GORM's `AutoMigrate`, which performs additive-only changes. No manual migration script is required and no downtime is needed.
+
+Existing rows have a `null` value, which the API converter treats as "no lifecycle failure". This is indistinguishable from "no failure" by design.
+
+Rollback is straightforward. Reverting the API server image and either dropping the column or leaving it in place unused restores the previous behavior. The column is informational only, so there is no data loss.
+
+## Frontend Considerations
```
alyssacgoins, line 244: Can you emphasize here that the frontend will clearly desiplay pod-level failures with this feature?

proposals/12843-pod-lifecycle-failure/README.md
```diff
@@ -0,0 +1,340 @@
...
+
+When a pod lifecycle failure happens, Argo Workflows records a human-readable message on the failed node, for example `Back-off pulling image "ghcr.io/example/does-not-exist:v1"`. KFP does not store or expose this message anywhere. The persistence agent reads the Argo workflow, converts node statuses into the internal `util.NodeStatus` struct, and drops the message field. The `Task` row in the database has no column for it, and the v2beta1 `GetRun` API never returns it.
+
+On the UI side, the run details page renders the pipeline graph from MLMD execution records. The driver pod writes the MLMD execution row in `RUNNING` state before the user container starts. If the user container then fails to start at all, the MLMD record stays at `RUNNING` forever, and the graph node renders green and spins indefinitely.
+
+KFP positions itself as a Kubernetes abstraction for data scientists and ML engineers. Many of its users are not Kubernetes operators. When the UI shows a task that never finishes and never fails, with no message, the only path forward is to ask someone else to run `kubectl get pods`. That breaks the abstraction the product is built on.
```
alyssacgoins, line 41: This is a great paragraph. This expresses the exact core of the problem, and what we're trying to solve with this feature. Great work :slightly_smiling_face:

proposals/12843-pod-lifecycle-failure/README.md
```diff
@@ -0,0 +1,340 @@
...
+
+In `toModelTask`, write the resolved message into `task.LifecycleFailureMessage`.
+
+In `toApiPipelineTaskDetail`, when the field is non-empty, build the response error as a `*rpcstatus.Status` with `Code: codes.Internal` and `Message: "Pod lifecycle failure: <argo message>"`. The status is constructed directly instead of using `util.NewInternalServerError`, because the latter prefixes the message with `InternalServerError:`, which would be misleading. The API server itself did not fail.
+
+#### `backend/src/apiserver/storage/task_store.go`
+
+Add `LifecycleFailureMessage` to `taskColumns`, to `scanRows`, and to the `CreateTask` and `CreateOrUpdateTasks` insert paths.
+
+In `patchTask`, do not preserve the previous value of this column. The fresh filtered value computed from the latest workflow state always wins. This is what makes the field self-clearing if a pod recovers on retry.
```
alyssacgoins: The overwrite logic you propose here is a departure from the current `backend/src/apiserver/storage/task_store.go/patch_task()` logic, which fills only empty fields. Your logic makes sense in the context of this feature, but make sure it does not break any existing `patch_task()` behavior - you may need to create a separate method.

proposals/12843-pod-lifecycle-failure/README.md
```diff
@@ -0,0 +1,340 @@
...
+Backend, added in `backend/src/apiserver/server/api_converter_test.go`:
+
+1. `TestNormalizeLifecycleFailureMessage` covers that real failure strings (`Back-off pulling image ...`, `OOMKilled`) pass through unchanged, `PodInitializing` and `ContainerCreating` return empty, and empty input returns empty.
+2. `TestNodeLifecycleFailureMessage` covers that a Failed or Pending node with a real message returns the message, and Succeeded, Skipped, and Omitted nodes return empty even when the message is non-empty.
+3. `TestResolveNodeLifecycleMessages` covers a direct message on a task node, propagation from a child executor node, parent precedence over a child, an empty result when no descendant has a message, a graph with a cycle, and two task nodes that share a descendant.
+
+Backend, in `backend/src/apiserver/storage/task_store_test.go`: existing CRUD tests cover the new column once it is added to the column list. No new test files are required.
+
+Frontend: existing `RunDetailsV2.test.tsx` continues to pass without changes, which is the regression bar. New behavior is exercised through the test cases that already cover `updateFlowElementsState`.
+
```
alyssacgoins: Can you add some information on additional frontend unit test coverage for this feature?

alyssacgoins (changes requested):

alyssacgoins:
@jeffspahr can we get your thoughts here from the frontend perspective?

alyssacgoins:
@droctothorpe @zazulam can we get your thoughts?

ntny:
I'm all for this proposal. The only risk I see is hypothetical: if KFP wants to add another backend in addition to Argo in the future, would this KEP  block that?
What might makes this less concerning to me is that most of the work seems to be on the persistence agent and UI side.

alyssacgoins:
> I'm all for this proposal. The only risk I see is hypothetical: if KFP wants to add another backend in addition to Argo in the future, would this KEP block that? What might makes this less concerning to me is that most of the work seems to be on the persistence agent and UI side.

@ntny If KFP was to add an additional backend in the future, I think the design would just need to propagate pod failure events by querying the K8s API directly - @khushiiagrawal includes this as an alternative in her doc [here](https://github.com/kubeflow/pipelines/pull/13517/changes#diff-5cd9dfcbcc93f401ca1a6e1fe864f1f56a4577850d02dd3d82b8018123e1ee23R328). This alternative is unnecessary in this case because the Argo message already contains the information, but it's an option for a non-Argo backend.

proposals/12843-pod-lifecycle-failure/README.md
khushiiagrawal: Yes. Timeouts need per-class timestamp tracking or a sweep mechanism in the persistence agent, which felt like a separate chunk of work from the core visibility fix. I've updated the Non-Goals section with a short rationale explaining this.

proposals/12843-pod-lifecycle-failure/README.md
khushiiagrawal: Good point, I've added a CI Validation section to the Test Plan covering a bad image API test and a frontend integration test for the banner. Both will ship with the implementation PR.

proposals/12843-pod-lifecycle-failure/README.md
khushiiagrawal: Thanks for flagging this Alyssa. There's actually no special overwrite logic added. LifecycleFailureMessage is just left out of the preserve-if-empty loop entirely, so the fresh value from the workflow sync always wins and the existing behavior for all other fields stays the same. I've updated the KEP wording to make this clearer.

proposals/12843-pod-lifecycle-failure/README.md
khushiiagrawal: Sure, updated the Test Plan with specific test cases covering the state override and banner rendering in DynamicFlow and RuntimeNodeDetailsV2.

proposals/12843-pod-lifecycle-failure/README.md
khushiiagrawal: @alyssacgoins  let me know what you think.

khushiiagrawal:
@alyssacgoins  I have addressed the changes, PTAL! 
Thanks.

proposals/12843-pod-lifecycle-failure/README.md
```diff
@@ -0,0 +1,360 @@
...
+- [Future Work](#future-work)
+- [Implementation History](#implementation-history)
+- [Drawbacks](#drawbacks)
+- [Alternatives](#alternatives)
+
+## Summary
+
+A KFP pipeline task can fail in two ways. The first is a user-script failure, where the Python code inside the container raises an exception. The second is a pod lifecycle failure, where the Kubernetes pod backing the task never reaches a healthy running state, or is killed by the system. Common examples of the second category are `ImagePullBackOff`, `Unschedulable`, `OOMKilled`, `CrashLoopBackOff`, and `NodeLost`.
+
+KFP handles the first case well today. The second case is currently invisible in the UI. The task node stays in a running state forever, no error is shown, and the user has no way to find out what happened without using `kubectl`. This proposal threads the pod lifecycle failure message that Argo Workflows already records all the way through the API server and into the run details page, so the task node turns red and the failure reason is shown in the side panel, just like a user-script failure is shown today.
```
ntny: I agree terminal vs transient classification can be deferred, but I’d handle `FailedScheduling` / `Unschedulable` carefully in the initial design.

On a well-utilized cluster, a pod may wait for resources, then get scheduled once other workloads finish. Showing that task as red could be misleading because it may recover without user action. Frequent red -> Running transitions would look strange in the UI because the task may recover without user action.
This feels more like a warning

Also, as far as I understand, the current V2 UI maps node status from MLMD executions rather than tasks, while there is ongoing MLMD-removal work to move it into tasks

- https://github.com/kubeflow/pipelines/pull/13478
- https://github.com/kubeflow/pipelines/pull/12430

alyssacgoins: @ntny @khushiiagrawal can we add a color (such as yellow) that indicates warning rather than failure for these cases? That way we're still communicating status.

proposals/12843-pod-lifecycle-failure/README.md
alyssacgoins: That make sense to me - once you've completed/merged this feature, be sure to add a follow-up issue for this

proposals/12843-pod-lifecycle-failure/README.md
```diff
@@ -0,0 +1,360 @@
...
+- A test confirms no `Banner` is rendered when `lifecycleError` is absent.
+
+### CI Validation
+
+The existing KFP end-to-end test suite in `.github/workflows/e2e-test.yml` runs full pipeline runs against a live cluster and verifies final task states. Once this change lands, one or more of the following will be added to cover the lifecycle failure path in CI:
+
+- A dedicated test pipeline with a deliberately bad container image that verifies the failing task's `error.message` contains the expected pod lifecycle failure string via the v2beta1 `GetRun` API.
+- A frontend integration test (in `test/frontend-integration-test/`) that submits the bad-image pipeline, waits for the task node to turn red, and asserts the banner text in the side panel.
+
+These CI additions will be part of the implementation PR rather than this KEP.
```
alyssacgoins:
```suggestion
```
You can remove this line - the KEP is a feature outline, so any work here will always be a part of the implementation PR

proposals/12843-pod-lifecycle-failure/README.md
```diff
@@ -0,0 +1,360 @@
...
+Frontend component tests, in `frontend/src/components/tabs/RuntimeNodeDetailsV2.test.tsx`:
+
+- A test confirms the `Banner` is rendered when `lifecycleError` is set on the flow element data.
+- A test confirms no `Banner` is rendered when `lifecycleError` is absent.
+
+### CI Validation
+
+The existing KFP end-to-end test suite in `.github/workflows/e2e-test.yml` runs full pipeline runs against a live cluster and verifies final task states. Once this change lands, one or more of the following will be added to cover the lifecycle failure path in CI:
+
+- A dedicated test pipeline with a deliberately bad container image that verifies the failing task's `error.message` contains the expected pod lifecycle failure string via the v2beta1 `GetRun` API.
```
alyssacgoins: Can you create a matrix here outlining scenario, the resulting pod status, and expected behavior?
You can use [this matrix](https://github.com/kubeflow/pipelines/tree/master/proposals/11385-literal-input-parameters#the-following-test-cases-are-expected-to-pass) from my Literal Inputs KEP as reference.

alyssacgoins (changes requested):

proposals/12843-pod-lifecycle-failure/README.md
ntny: Yes, that makes sense to me.

khushiiagrawal: @alyssacgoins  @ntny  That makes a lot of sense. I'll update the KEP to split the failure classification into terminal (red, e.g. OOMKilled, image pull after Argo gives up) and transient/warning (yellow, e.g. Unschedulable, FailedScheduling). For the frontend, does adding a yellow/amber node color require a new state in the existing Execution.State enum, or would it make more sense to handle it as a separate visual property on the flow element? Happy to go whichever direction you prefer before I update the design.

proposals/12843-pod-lifecycle-failure/README.md
khushiiagrawal: Done, updated the CI Validation section with a matrix covering the key scenarios, pod status, and expected behavior at each layer (persistence agent, API, and frontend). Used the Literal Inputs KEP format as reference.

khushiiagrawal:
@mprahl @alyssacgoins , really appreciate if you could take a look once when you get a chance. I would be happy to finalize the KEP asap!

proposals/12843-pod-lifecycle-failure/README.md
Snehadas2005: Reproduced the Unschedulable case locally to check how this plays out in practice (kind cluster, forced `Insufficient cpu/memory` via `set_cpu_limit`). Confirms what's being discussed here — Argo already puts a clean `Unschedulable: 0/1 nodes are available: ...` message on the node, so @ntny's concern is real and observable: a task can sit in that state for a while and later actually get scheduled once resources free up, so red/FAILED would misrepresent it.

Noticed the implementation in #13516 hasn't been updated since before this thread; it still overrides to `Execution.State.FAILED` unconditionally for any `lifecycleError`, scheduling included. On @khushiiagrawal's open question: a new `Execution.State.WARNING` enum value seems cleaner than a separate visual property, since `state` is already the single thing everything downstream (filters, tests, other views) switches on; a second property risks drifting out of sync with it. Happy to pick up that piece in #13516 if it's still open, using the taxonomy table above to split terminal vs transient.

alyssacgoins: Hey @Snehadas2005 thanks for your interest here. This feature has been selected for the [LFX term 3 mentorship](https://mentorship.lfx.linuxfoundation.org/project/aba3293b-593c-4afd-a22d-27d70fc50ff8), so we're not looking for anyone to pick up this work until we've selected an applicant. We encourage you to apply!

proposals/12843-pod-lifecycle-failure/README.md
```diff
@@ -0,0 +1,366 @@
...
+
+KFP handles the first case well today. The second case is currently invisible in the UI. The task node stays in a running state forever, no error is shown, and the user has no way to find out what happened without using `kubectl`. This proposal threads the pod lifecycle failure message that Argo Workflows already records all the way through the API server and into the run details page, so the task node turns red and the failure reason is shown in the side panel, just like a user-script failure is shown today.
+
+## Motivation
+
+This work started as a GSoC 2026 project idea opened by [@alyssacgoins](https://github.com/alyssacgoins) in [#12843](https://github.com/kubeflow/pipelines/issues/12843). [@khushiiagrawal](https://github.com/khushiiagrawal) wrote a GSoC proposal for it and did an initial draft of the implementation. The project was not selected for GSoC this year, but after discussing with the maintainer we agreed that it is still worth landing in KFP, so the work is continuing outside of GSoC and being submitted through the regular KEP process.
```
mprahl: I think we can drop this paragraph.

proposals/12843-pod-lifecycle-failure/README.md
```diff
@@ -0,0 +1,366 @@
...
+
+When a pod lifecycle failure happens, Argo Workflows records a human-readable message on the failed node, for example `Back-off pulling image "ghcr.io/example/does-not-exist:v1"`. KFP does not store or expose this message anywhere. The persistence agent reads the Argo workflow, converts node statuses into the internal `util.NodeStatus` struct, and drops the message field. The `Task` row in the database has no column for it, and the v2beta1 `GetRun` API never returns it.
```
mprahl: Could you change the proposal to be based on MLMD removal?

Here is the WIP feature rebase PR:
https://github.com/kubeflow/pipelines/pull/13986

Here is the KEP:
https://github.com/kubeflow/pipelines/tree/master/proposals/12147-mlmd-removal

proposals/12843-pod-lifecycle-failure/README.md
```diff
@@ -0,0 +1,366 @@
...
+
+On the UI side, the run details page renders the pipeline graph from MLMD execution records. The driver pod writes the MLMD execution row in `RUNNING` state before the user container starts. If the user container then fails to start at all, the MLMD record stays at `RUNNING` forever, and the graph node renders green and spins indefinitely.
+
+KFP positions itself as a Kubernetes abstraction for data scientists and ML engineers. Many of its users are not Kubernetes operators. When the UI shows a task that never finishes and never fails, with no message, the only path forward is to ask someone else to run `kubectl get pods`. That breaks the abstraction the product is built on.
+
+### Goals
+
+1. Capture the Argo node failure message and persist it on the KFP `Task` row.
```
mprahl: We should avoid being Argo specific. The underlying implementation may end up using Argo but that should be under the existing pipeline engine abstractions.

proposals/12843-pod-lifecycle-failure/README.md
```diff
@@ -0,0 +1,366 @@
...
+2. Expose the message through the existing `task_details[].error` field of the v2beta1 `GetRun` response. No new API surface.
+3. In the UI, when a lifecycle failure is present on a task, override the node visual state to Failed (red), even if MLMD still reports `RUNNING`, and show the failure reason in a banner inside the node detail side panel.
+4. Avoid false positives. Argo emits transient messages like `PodInitializing` during normal pod startup. Healthy runs must not show a failure banner.
+
+### Non-Goals
+
+1. Configurable per-class timeouts (for example, fail after one hour of `ImagePullBackOff`). Timeouts require tracking how long a pod has been in a given failure state, which needs either a new timestamp column on the `Task` model or a separate sweep mechanism in the persistence agent. That is a distinct feature that can be built on top of the `LifecycleFailureMessage` foundation this proposal lays. Bundling it here would significantly increase scope and risk without changing the core visibility fix.
```
mprahl: I think it's still worth adding a single timeout value for all infra errors, or else the pipeline run will run forever during `ImagePullBackoff` from user error.

proposals/12843-pod-lifecycle-failure/README.md
```diff
@@ -0,0 +1,366 @@
...
+
+I push a pipeline that references a misspelled container image. The run starts, and the affected task pod hits `ImagePullBackOff`. Within a few seconds the task node in the UI turns red. I click on it and the side panel shows `Pod lifecycle failure: Back-off pulling image "..."`. I fix the image name and re-run.
+
+#### Story 2: Out of memory
+
+A task pod runs out of memory and gets `OOMKilled` by the kernel. The task node turns red. The side panel says `Pod lifecycle failure: OOMKilled`. I bump the memory request on that component and re-run.
+
+#### Story 3: Healthy pipeline
+
+I submit a healthy pipeline. Every task transitions through normal pod startup states (`PodInitializing`, `ContainerCreating`) before running. None of the task nodes ever shows a lifecycle failure banner. The run completes green.
```
mprahl: This user story can be dropped.

proposals/12843-pod-lifecycle-failure/README.md
```diff
@@ -0,0 +1,366 @@
...
+
+```
+parent task node (rendered by the UI)
+    phase: Pending
+    message: ""
+    children: [executor pod node]
+
+executor pod node (one per task pod)
+    phase: Pending
+    message: 'Back-off pulling image "ghcr.io/example/does-not-exist:v1"'
```
mprahl, line 98: Rather than relying on this Argo specific behavior, could we consider having the controller check Kubernetes directly? I think depending on this behavior is fragile as it doesn't really seem like an API but rather just information for their UI.

proposals/12843-pod-lifecycle-failure/README.md
```diff
@@ -0,0 +1,366 @@
...
+
+```go
+type Task struct {
+    // ... existing fields ...
+
+    LifecycleFailureMessage LargeText `gorm:"column:LifecycleFailureMessage; type:longtext;"`
+}
+```
+
+The new field is kept on its own line, separated from `ParentTaskId` by a blank line, so that `gofmt` does not pull the surrounding fields into a single aligned block and trigger `golangci-lint --new` on the pre-existing `ParentTaskId` name.
```
mprahl: This is far too much detail for a KEP. Please keep the KEP focused on architecture, behavior, database schema, and public API changes.

proposals/12843-pod-lifecycle-failure/README.md
```diff
@@ -0,0 +1,366 @@
...
+
+#### `backend/src/apiserver/server/api_converter.go`
+
+Three small helpers, then one place that uses them.
+
+1. `normalizeLifecycleFailureMessage(message string) string`. Returns an empty string for transient startup messages (`PodInitializing`, `ContainerCreating`). Otherwise returns the input unchanged.
+2. `nodeLifecycleFailureMessage(node util.NodeStatus) string`. Calls `normalizeLifecycleFailureMessage`, then additionally returns empty for nodes whose state is `Succeeded`, `Skipped`, or `Omitted`.
+3. `resolveNodeLifecycleMessages(nodes map[string]util.NodeStatus) map[string]string`. Walks the node graph starting from each task node. For a given node, the resolved message is its own filtered message if non-empty, otherwise the first non-empty resolved message from any of its descendants. Cycle detection uses an `onStack` map keyed on node id, with a deferred delete on return so that shared descendants (the same node reached through two different parents) are not incorrectly treated as cycles.
+
+In `toModelTask`, write the resolved message into `task.LifecycleFailureMessage`.
```
mprahl: **🤖 AI Review**

Found by: gpt-5.4
Severity: high | Confidence: high

Verified against `backend/api/v2beta1/run.proto`, which documents `PipelineTaskDetail.error` as "Only populated when the task is in FAILED or CANCELED state," and against the current `backend/src/apiserver/server/api_converter.go`, which would continue to return the stored task `State` unchanged. The proposal says to populate `error` whenever `LifecycleFailureMessage` is non-empty, while its own design examples depend on lifecycle failures surfacing before Argo/MLMD mark the task terminal. That would turn a documented terminal-only field into a normal non-terminal signal for `GetRun` clients. Use a separate lifecycle-diagnostic field, or define the backend behavior so `task_details.state` is also transitioned to a terminal failure state and cover that with API tests.

proposals/12843-pod-lifecycle-failure/README.md
```diff
@@ -0,0 +1,366 @@
...
+}
+```
+
+#### `backend/src/common/util/workflow.go`
+
+In `NodeStatuses()`, populate the new field from `node.Message` when iterating over the Argo workflow nodes. This is the only place the message enters the KFP code path.
+
+#### `backend/src/apiserver/model/task.go`
+
+Add a nullable `LifecycleFailureMessage LargeText` column to the `Task` model. It is stored as a dedicated field rather than embedded inside the existing payload column so that it can be queried, indexed, and updated independently.
```
mprahl: **🤖 AI Review**

Found by: gpt-5.4
Severity: high | Confidence: high

Verified against `backend/src/apiserver/model/common.go`, where `model.LargeText` already maps to `LONGTEXT` for MySQL and `TEXT` for other dialects, and against `backend/src/apiserver/client_manager/client_manager.go` plus `backend/src/apiserver/client_manager/dialect_test.go`, which confirm the API server still supports `pgx`/PostgreSQL. Hard-coding `gorm:"...; type:longtext;"` on a `LargeText` field would bypass that cross-database abstraction and is likely to break `AutoMigrate` or schema creation on PostgreSQL even though sqlite-based tests would not catch it. Drop the explicit `type:longtext` tag and let `LargeText` supply the dialect-specific column type; add migration coverage for the `pgx` path if this field is implemented.

proposals/12843-pod-lifecycle-failure/README.md
```diff
@@ -0,0 +1,366 @@
...
+```ts
+export type ExecutionFlowElementData = FlowElementDataBase & {
+  state?: Execution.State;
+  lifecycleError?: string;
+};
+```
+
+#### `frontend/src/lib/v2/DynamicFlow.ts`
+
+`updateFlowElementsState` now takes the run's `task_details` array as an extra argument. From it, build a map keyed on the pipeline task display name, valued by the error message string. While iterating execution nodes, if the node's task name has an entry in the map, write the message onto `data.lifecycleError`. If `data.lifecycleError` is non-empty, override `data.state` to `Execution.State.FAILED`. This last step is the critical one: it makes the node render red even though MLMD still reports `RUNNING`.
```
mprahl: **🤖 AI Review**

Found by: gpt-5.4
Severity: medium | Confidence: high

Verified against `frontend/src/lib/v2/StaticFlow.ts` and `frontend/src/lib/v2/DynamicFlow.ts`: execution nodes are identified by `task.<taskKey>`, and runtime lookups use `getTaskKeyFromNodeKey(elem.id)`, not the rendered label. I also checked `test_data/sdk_compiled_pipelines/valid/two_step_pipeline.yaml`, which has task keys such as `write-to-gcs` but `taskInfo.name` values such as `Producer`, proving those identifiers can diverge in this repo today. A map built from `task_details[].display_name` and then consulted with the node's task key will miss lifecycle errors for pipelines with custom names and can alias duplicate display names. Key the map by a stable identifier already present in the API, such as task key or `task_id`, and add a test where the task key and display name differ.

mprahl (changes requested):

proposals/12843-pod-lifecycle-failure/README.md
ntny: Hi @mprahl!
Argo Workflows is currently the execution backend for Kubeflow Pipelines, and the Argo controller already detects and surfaces most pod/runtime failure conditions we care about.

Could you clarify what gaps you see in relying on Argo for this? It seems that Argo already provides much of this detection logic, so I’m curious whether we could reuse that rather than implement similar logic in KFP.

mprahl: I’m okay with using Argo node status as the backend implementation source, but I’d like us to keep this behind the execution abstraction and treat it as best-effort diagnostic data rather than a stable API contract. I would avoid guaranteeing exact Argo message strings in KFP’s public API and prefer normalizing them into KFP-owned semantics where possible.

ntny: If I understood correctly, we reuse Argo on the backend while keeping the KFP API Argo-agnostic. Totally agree.

proposals/12843-pod-lifecycle-failure/README.md
khushiiagrawal: sure, thanks for letting me know this. helps a lot

khushiiagrawal:
@mprahl  @ntny  @alyssacgoins 
Thanks for the review! I've addressed all the points:

- Reframed the proposal around the MLMD removal direction (KEP-12147, #13986). The lifecycle data lives on the Task row and the UI reads from task_details, so it fits naturally into the post-MLMD architecture.
- Added a single configurable infra-error timeout (LIFECYCLE_FAILURE_TIMEOUT, default 1h). Per-category granularity deferred to a follow-up.
- Switched from reusing the terminal-only error field to a new dedicated lifecycle diagnostic field on PipelineTaskDetail, since error is documented as "FAILED or CANCELED only" in run.proto.
- Frontend map now uses task key via getTaskKeyFromNodeKey instead of display_name.

Let me know if anything needs more detail!

proposals/12843-pod-lifecycle-failure/README.md
```diff
@@ -0,0 +1,391 @@
...
+
+On the UI side, the run details page currently renders the pipeline graph from MLMD execution records. The driver pod writes the MLMD execution row in `RUNNING` state before the user container starts. If the user container then fails to start at all, the MLMD record stays at `RUNNING` forever, and the graph node renders green and spins indefinitely.
+
+This proposal aligns with and builds on the ongoing MLMD removal effort ([KEP-12147](../12147-mlmd-removal/README.md), [#13986](https://github.com/kubeflow/pipelines/pull/13986)). As MLMD is removed, the frontend will read execution state from task APIs rather than from MLMD. The lifecycle failure information proposed here is stored on the `Task` row and exposed through the task API, so it fits naturally into the post-MLMD architecture. The UI state override described below converges with the MLMD removal: once node status comes from tasks, the override simply becomes the value.
```
mprahl: We can shorten this section and just keep it as simple as this depends on MLMD removal.

proposals/12843-pod-lifecycle-failure/README.md
```diff
@@ -0,0 +1,391 @@
...
+
+KFP positions itself as a Kubernetes abstraction for data scientists and ML engineers. Many of its users are not Kubernetes operators. When the UI shows a task that never finishes and never fails, with no message, the only path forward is to ask someone else to run `kubectl get pods`. That breaks the abstraction the product is built on.
+
+### Goals
+
+1. Capture the pod lifecycle failure message from the execution engine and persist it on the KFP `Task` row. The underlying implementation uses the execution engine's node status, but the KFP API remains engine-agnostic.
+2. Expose the lifecycle failure through a dedicated field on `PipelineTaskDetail` in the v2beta1 `GetRun` response, separate from the existing terminal-only `error` field (see [Backend Changes](#backend-changes) for rationale).
+3. In the UI, when a lifecycle failure is present on a task, override the node visual state to reflect the failure, even if MLMD still reports `RUNNING` (or, post-MLMD removal, if the task state has not yet been updated), and show the failure reason in a banner inside the node detail side panel.
```
mprahl: Please remove all references to MLMD and assume MLMD is removed.

proposals/12843-pod-lifecycle-failure/README.md
```diff
@@ -0,0 +1,391 @@
...
+### Non-Goals
+
+1. Per-category configurable timeouts with distinct thresholds per failure class. A single infra-error timeout is in scope; per-class granularity can be built on top of it later.
+2. Any changes to the SDK, the pipeline spec, or the compiler.
+3. Any changes to how user-script failures are captured. Those already work.
+4. Surfacing pod logs in the UI. Out of scope.
+
+## Proposal
+
+The execution engine's node status already contains exactly the information the user needs. The fix is to stop dropping it. Persist it on the `Task` model behind the engine abstraction, return it in a dedicated lifecycle diagnostic field in the run response, and have the frontend render it. The work is split across six backend files and four frontend files. There are no proto changes to existing fields, no new API endpoints, and no breaking changes to existing responses.
```
mprahl: Please remove unnecessary details such as "The work is split across six backend files and four frontend files." that don't help KFP maintainers make decisions on the technical design and proposed UX.

proposals/12843-pod-lifecycle-failure/README.md
```diff
@@ -0,0 +1,391 @@
...
+
+A task pod runs out of memory and gets `OOMKilled` by the kernel. The task node indicates a failure. The side panel shows the reason. I bump the memory request on that component and re-run.
+
+#### Story 3: Stuck run auto-terminates
+
+I push a pipeline with a misspelled image. After the configured infra-error timeout elapses (default: 1 hour), the run is automatically terminated with the failure reason recorded. I do not need to manually cancel the run or discover via `kubectl` that it will never succeed.
+
+### Risks and Mitigations
+
+| Risk | Mitigation |
```
mprahl, line 78: Please keep this down to high-level risks and not implementation details best discussed at the code PR level (e.g. `patchTask` always overwrites the field with the latest filtered value, with no preservation logic).

proposals/12843-pod-lifecycle-failure/README.md
```diff
@@ -0,0 +1,391 @@
...
+
+The current execution engine (Argo Workflows) records these messages via its `assessNodeStatus` function, which writes onto `node.Message` on every pod sync. This covers all three failure levels: provisioning (`ImagePullBackOff`, `Unschedulable`), runtime (`OOMKilled`, `CrashLoopBackOff`), and node-level (`NodeLost`, eviction). This behavior is treated as best-effort diagnostic data rather than a stable API contract. KFP normalizes these messages into its own semantics so that the public API does not expose engine-specific message formats.
+
+An alternative approach would be to query Kubernetes pod status or events directly. This was rejected because it would require additional RBAC grants on the persistence agent role, would duplicate work the engine already does, and would miss cases where the pod is never created at all (such as quota or admission blocks), which the engine still records.
+
+### Backend Changes
```
mprahl: This section is too implementation detail heavy.

Please keep this scoped to just public HTTP API changes, public SDK changes, database schema changes, and interface changes if it's part of an API contract (e.g. pipeline engine agnosticism, plugin interface, etc.)

Please focus on high-level behavior, not the actual code changes.

proposals/12843-pod-lifecycle-failure/README.md
```diff
@@ -0,0 +1,391 @@
...
+
+#### `backend/src/apiserver/model/task.go`
+
+Add a nullable `LifecycleFailureMessage` column to the `Task` model using the `LargeText` type, which maps to `LONGTEXT` on MySQL and `TEXT` on other dialects (such as PostgreSQL via pgx). No explicit `type:` GORM tag is used, allowing `LargeText` to handle dialect-specific column types through the existing cross-database abstraction in `model/common.go`.
+
+```go
+type Task struct {
+    // ... existing fields ...
+
+    LifecycleFailureMessage LargeText `gorm:"column:LifecycleFailureMessage;"`
```
mprahl: I'm wondering if it makes sense to rename this to `LifecycleMessage` so that we can capture in progress lifecycle changes for display in the UI.

proposals/12843-pod-lifecycle-failure/README.md
```diff
@@ -0,0 +1,391 @@
...
+
+| Category | Example messages |
+|---|---|
+| Image pull | `Back-off pulling image`, `ErrImagePull`, `ImagePullBackOff` |
+| Resource / scheduling | `Unschedulable`, `FailedScheduling`, `Insufficient cpu` |
+| Runtime / OOM | `OOMKilled`, `Error`, `CrashLoopBackOff` |
+| Admission / policy | `pods "..." is forbidden` (Kyverno, PodSecurityPolicy) |
+| Unknown | Anything not matching the above |
+
+Structured classification is intentionally deferred from this proposal. The raw message already gives the user the information needed to diagnose the failure, and the engine's node message works even in cases where the pod was never created at all (for example, resource quota exceeded or an admission controller blocking pod creation). A structured category field can be added later as a separate column on the `Task` model without any breaking changes to the existing `LifecycleFailureMessage` field or the API response. For unknown or newly introduced failure patterns, the structured status would stay `Unknown` while the raw message would still provide the original diagnostic details.
```
mprahl: I don't think this can be deferred. KFP is supposed to be pipeline engine agnostic. I have a PoC that will turn into a KEP after MLMD is removed to allow the KFP API server to run pipelines locally (subprocess or Docker), so having a way to categorize infra failures is worthwhile IMO.

proposals/12843-pod-lifecycle-failure/README.md
```diff
@@ -0,0 +1,391 @@
...
+
+### Reporting Latency
+
+The message resolution function runs inside the persistence agent reporting path, during task model construction, which is called for each node during `ReportWorkflowResource`. The traversal is a DFS over the node graph with a memoization cache, so each node is visited at most once. The time complexity is O(n) where n is the number of nodes in the workflow.
```
mprahl: Again, this is too much detail for a KEP. This is best for code comments or PR comments.

proposals/12843-pod-lifecycle-failure/README.md
```diff
@@ -0,0 +1,391 @@
...
+
+For typical KFP pipelines this adds negligible overhead. A pipeline with 100 tasks produces roughly 200-300 nodes (one task node and one executor pod node per task). The traversal is entirely in-memory map lookups with no I/O.
+
+For cron runs, the persistence agent reporting path is more sensitive because the run row may not yet exist in MySQL when the first workflow state update arrives. The lifecycle message processing does not add any new database reads or writes beyond the existing `patchTask` upsert. The graph traversal happens entirely in memory before the database write, so it does not affect the likelihood or window of the existing race condition on the run row.
+
+A before-and-after performance measurement of the persistence agent reporting loop will be included in the implementation PR.
```
mprahl: This is not needed IMO.

proposals/12843-pod-lifecycle-failure/README.md
```diff
@@ -0,0 +1,391 @@
...
+
+Without a timeout, a pipeline run stuck in a pod lifecycle failure such as `ImagePullBackOff` will run indefinitely, consuming cluster resources. This proposal includes a single configurable infra-error timeout that applies to all pod lifecycle failures.
+
+The persistence agent gains a sweep worker on its own `time.Ticker`, separate from the reporting loop. It maintains a map tracking when each task first received a non-empty `LifecycleFailureMessage`. If the elapsed time exceeds the configured threshold, the run is terminated with the failure reason recorded as the termination message. The timestamp resets if the message changes or clears (recovery).
+
+The timeout is configured via a single environment variable:
+
+| Variable | Default | Description |
+|---|---|---|
+| `LIFECYCLE_FAILURE_TIMEOUT` | `1h` | Duration after which a run stuck in a pod lifecycle failure state is terminated. Set to `0` to disable. |
```
mprahl: @droctothorpe @zazulam , should this be configurable per pipeline or just a single global setting? Is 1 hour sufficient do you think?

In other words, if a task is stuck in ImagePullBackoff for 1 hour, is it safe to fail the run at that point or do we want longer?

proposals/12843-pod-lifecycle-failure/README.md
```diff
@@ -0,0 +1,391 @@
...
+The message resolution function runs inside the persistence agent reporting path, during task model construction, which is called for each node during `ReportWorkflowResource`. The traversal is a DFS over the node graph with a memoization cache, so each node is visited at most once. The time complexity is O(n) where n is the number of nodes in the workflow.
+
+For typical KFP pipelines this adds negligible overhead. A pipeline with 100 tasks produces roughly 200-300 nodes (one task node and one executor pod node per task). The traversal is entirely in-memory map lookups with no I/O.
+
+For cron runs, the persistence agent reporting path is more sensitive because the run row may not yet exist in MySQL when the first workflow state update arrives. The lifecycle message processing does not add any new database reads or writes beyond the existing `patchTask` upsert. The graph traversal happens entirely in memory before the database write, so it does not affect the likelihood or window of the existing race condition on the run row.
+
+A before-and-after performance measurement of the persistence agent reporting loop will be included in the implementation PR.
+
+### Infra-Error Timeout
```
mprahl: Which infra issues would cause a timeout? I don't think resource scheduling delays should do that since that's not necessarily user error (e.g. waiting for a GPU to become available). Pending volume mounts for an invalid storage class in a PVC would be a good candidate to fail early as that is most likely user error.

proposals/12843-pod-lifecycle-failure/README.md
```diff
@@ -0,0 +1,391 @@
...
+    --> C["api_converter.go: normalize / filter / resolve lifecycle messages"]
+    --> D["task_store.go: Task.LifecycleFailureMessage stored in DB"]
+    --> E["api_converter.go: PipelineTaskDetail.lifecycle_failure returned in API"]
+    --> F["DynamicFlow.ts: lifecycleError set, node state overridden"]
+    --> G["RuntimeNodeDetailsV2.tsx: Banner shown in side panel"]
+```
+
+## Test Plan
+
+### Unit Tests
```
mprahl: I think you can delete the "Unit Tests" section. This is too detailed for a KEP.

proposals/12843-pod-lifecycle-failure/README.md
mprahl: What process is responsible for detecting a timeout? How do you handle statuses that change but are due to infra error?

proposals/12843-pod-lifecycle-failure/README.md
```diff
@@ -0,0 +1,391 @@
...
+|---|---|---|---|---|
+| Bad image (`ImagePullBackOff`) | yes | yes | yes | yes |
+| Runtime OOM (`OOMKilled`) | yes | yes | yes | yes |
+| Pod never created (resource quota exceeded) | yes | yes | yes | yes |
+| Healthy pipeline (regression check) | no message stored | no field | stays green | no banner |
+| Retry: first attempt fails, retry succeeds | yes | cleared after recovery | turns green after recovery | banner clears after recovery |
+| Transient startup (`PodInitializing`) | filtered, not stored | no field | stays green during startup | no banner during startup |
+| Infra-error timeout | yes | yes | run terminated | reason recorded |
+
+### Manual Verification (E2E)
```
mprahl: Could you please delete the manual testing section? I think just high-level integration test scenarios is sufficient for the testing section.

proposals/12843-pod-lifecycle-failure/README.md
```diff
@@ -0,0 +1,391 @@
...
+
+The only schema change is one new nullable column on the `tasks` table, `LifecycleFailureMessage`, stored using the `LargeText` type which maps to `LONGTEXT` on MySQL and `TEXT` on PostgreSQL. It is added on API server startup by GORM's `AutoMigrate`, which performs additive-only changes. No manual migration script is required and no downtime is needed.
+
+Existing rows have a `null` value, which the API converter treats as "no lifecycle failure". This is indistinguishable from "no failure" by design.
+
+Rollback is straightforward. Reverting the API server image and either dropping the column or leaving it in place unused restores the previous behavior. The column is informational only, so there is no data loss.
+
+This migration is compatible with the MLMD removal effort. The `Task` table changes proposed in [KEP-12147](../12147-mlmd-removal/README.md) drop and recreate the table; the `LifecycleFailureMessage` column can be included in the new schema definition.
+
+## Frontend Considerations
```
mprahl, line 244: Please keep this section to describing the UX unless there is an actual technical decision to be made.

proposals/12843-pod-lifecycle-failure/README.md
```diff
@@ -0,0 +1,391 @@
...
+
+## Future Work
+
+The following improvements are out of scope for this proposal but are worth keeping in mind for follow-up work.
+
+**Structured message classification.** See the [Structured Message Classification](#structured-message-classification) section for the proposed category taxonomy. Once the raw message is stable, a structured classifier can be added as a new `LifecycleFailureCategory` column without breaking changes to the existing field or API response.
+
+**Per-category configurable timeouts.** The single infra-error timeout in this proposal covers the common case. Distinct thresholds per failure class (for example, a shorter timeout for admission blocks than for scheduling failures) can be added later once structured classification is in place.
+
+**Transient vs terminal failure distinction.** `OOMKilled` is a terminal failure and it may make sense to immediately transition the task to a `Failed` state. `ImagePullBackOff` and `FailedScheduling` are transient and may resolve on retry. A separate `Warning` task state could be introduced for transient conditions, but this requires additional schema changes and UI work.
```
mprahl: Could we just detect that if it crashed more than say three times, we fail?

proposals/12843-pod-lifecycle-failure/README.md
```diff
@@ -0,0 +1,391 @@
...
+## Implementation History
+
+- 2026-02-18: Original feature request opened by [@alyssacgoins](https://github.com/alyssacgoins) as [#12843](https://github.com/kubeflow/pipelines/issues/12843).
+- 2026-06-11: Initial implementation drafted by [@khushiiagrawal](https://github.com/khushiiagrawal) as [#13516](https://github.com/kubeflow/pipelines/pull/13516) and verified end to end on a local Kind cluster against both failure and success pipelines.
+- 2026-06-12: KEP submitted (this document).
+
+## Drawbacks
+
+1. The displayed message is the raw engine node message, normalized into KFP-owned semantics. It is human-readable but not structured, so it cannot be programmatically classified into provisioning, runtime, or node-level failures from the API alone. A structured classification could be added later without breaking this change.
+2. One additional nullable column is added to the `tasks` table. For most operators this is a non-issue, but it is worth being explicit about.
```
mprahl: This isn't a drawback IMO.

proposals/12843-pod-lifecycle-failure/README.md
```diff
@@ -0,0 +1,391 @@
...
+
+Rollback is straightforward. Reverting the API server image and either dropping the column or leaving it in place unused restores the previous behavior. The column is informational only, so there is no data loss.
+
+This migration is compatible with the MLMD removal effort. The `Task` table changes proposed in [KEP-12147](../12147-mlmd-removal/README.md) drop and recreate the table; the `LifecycleFailureMessage` column can be included in the new schema definition.
+
+## Frontend Considerations
+
+This proposal directly improves what users see in the run details UI. When a pod lifecycle failure occurs, the affected task node changes state to indicate a failure, and the side panel shows a banner with the failure reason. This happens even when the task has not yet been marked as failed by the execution engine, because the frontend overrides the node state based on the lifecycle failure field in the API response rather than relying solely on MLMD execution records (or, post-MLMD removal, the task state alone). Users get a clear, actionable error message without ever leaving the KFP UI or running `kubectl`.
```
mprahl: Does it actually change state unless there is a timeout? From the proposed backend changes, I thought it was an information only message until timeout?

I think the proposed UX requires more thought.

mprahl (changes requested):

proposals/12843-pod-lifecycle-failure/README.md
khushiiagrawal: yeah it makes sense, i was taking both the possibilities. made it to remove MLMD.

proposals/12843-pod-lifecycle-failure/README.md
khushiiagrawal: I was used to write detailed proposal so wrote the KEP also like that 😅.  Thanks for letting me know.

proposals/12843-pod-lifecycle-failure/README.md
khushiiagrawal: yea makes sense, i have renamed it. thanks!

proposals/12843-pod-lifecycle-failure/README.md
khushiiagrawal: Good point. I updated the section to split categories into timeout-eligible and non-timeout. Image pull, admission/policy, and volume mount failures trigger the timeout since they are almost always user error and will not self-resolve. Resource/scheduling failures like waiting for a GPU do not trigger it, since those can resolve on their own. OOMKilled and CrashLoopBackOff are already handled by the engine's retry policy, so those are excluded too.

khushiiagrawal: The persistence agent runs a separate sweep worker on its own ticker, independent of the reporting loop. For statuses that change but stay in the same failure class (like Kubernetes alternating between ErrImagePull and ImagePullBackOff), the timer is not reset since the underlying problem hasn't changed. The timer only resets when the message clears (recovery) or changes to a non-timeout category. Updated the section with these details.

proposals/12843-pod-lifecycle-failure/README.md
khushiiagrawal: its done, classification is now in scope. The LifecycleCategory column is part of the initial implementation and ClassifyPodFailure is in a shared package that both the persistence agent and the timeout sweep import. This also fits naturally with the local runner use case you mentioned.

proposals/12843-pod-lifecycle-failure/README.md
khushiiagrawal: Updated. The section now describes a warning-first UX: when a lifecycle message is present, the node shows a warning indicator (amber) and a banner in the side panel, but does not change state to failed. The node only turns red when the timeout terminates the run or the engine marks the task as terminal. This way a pod waiting for a GPU or recovering on retry shows a warning without falsely indicating failure.

Priyanshu-u07:
I reproduced these failure modes on a cluster a while back and two of the results look relevant to the open questions here, sharing in case they're useful.
Both from Argo Workflows 4.0.4 / Kubernetes 1.35.0, on bare Argo rather than a full KFP install.

**On the volume-mount row in the timeout table:**  A pod blocked on a nonexistent storage class doesn't report a volume-specific reason. It reports `Unschedulable`, same as a pod waiting for CPU:
```
Unschedulable: 0/1 nodes are available: 1 Insufficient cpu…
Unschedulable: 0/1 nodes are available: pod has unbound immediate PersistentVolumeClaims. not found
```
Same reason, same pod condition (`PodScheduled` / `False` / `Unschedulable`), same event reason (`FailedScheduling`), and no container statuses in either. They appear as adjacent events in one capture, on one cluster at one moment. Under substring matching against the classification table, both land in Resource / scheduling, which is the non-timeout row, so the storage-class case wouldn't time out. The text after the colon is the only thing separating them.

**On CrashLoopBackOff:** Task pods run `restartPolicy: Never`, so the kubelet never restarts a container. `CrashLoopBackOff` isn't reachable for a task container and `restartCount` stays at zero, which affects both the Runtime / OOM row and the crash-count threshold in Future Work. Argo's `retryStrategy` does create a new pod per attempt, so a task-level count exists but a stuck pod never reaches a failed state, so the retry trigger may not fire for this class anyway.

There also isn't a common retry counter across these failures. In my runs the kubelet emitted `BackOff` ×151 for a mistyped tag, `InspectFailed` ×26 for a malformed reference, and `Failed` ×26 for a missing Secret. Different event types per failure mode, so there's no single number to threshold on.

Captures are at https://github.com/Priyanshu-u07/kfp-pod-lifecycle-evidence if useful. 
Thank you!

proposals/12843-pod-lifecycle-failure/README.md
```diff
@@ -0,0 +1,293 @@
...
+
+It is stored as a dedicated field rather than embedded inside the existing payload column so that it can be queried, indexed, and updated independently.
+
+#### Message processing
+
+Before persisting, the message is normalized (transient startup strings like `PodInitializing` and `ContainerCreating` are dropped), non-failure nodes (`Succeeded`, `Skipped`, `Omitted`) are suppressed, and the message is propagated from child executor nodes up to the parent task node the UI renders. The result is written to the task via `UpdateTask`.
+
+#### API exposure
+
+The lifecycle message is exposed through a **new dedicated field** on `PipelineTaskDetail` rather than through the existing `error` field. The `error` field is documented in `run.proto` as "Only populated when the task is in FAILED or CANCELED state." Lifecycle failures need to surface while the task is still in a non-terminal state (the engine may still be retrying), so reusing `error` would violate the API contract. The new field is populated whenever the lifecycle message is non-empty, regardless of task state, and clears when the pod recovers. This requires a proto addition to `PipelineTaskDetail`.
```
alyssacgoins: `PipelineTaskDetail` has been removed and `RunDetails.task_details` is reserved; the replacement is `PipelineTask`, exposed through task APIs and Run.tasks. Specify the lifecycle diagnostic on PipelineTask. Update the terminal-error rationale to reference the current PipelineTask.StatusMetadata.message contract.

proposals/12843-pod-lifecycle-failure/README.md
```diff
@@ -0,0 +1,293 @@
...
+
+#### Storage behavior
+
+The `LifecycleMessage` field always reflects the latest value from the engine, clearing on recovery. It does not participate in preserve-if-empty semantics.
+
+### Frontend Changes
+
+The run details graph reads `task_details` from the API response and maps lifecycle messages to nodes using the **task key** (not `display_name`, since the two can diverge). When a lifecycle message is present, the node shows a warning indicator and a banner is rendered in the side panel with the message. The node state is not overridden to failed; it transitions to failed only when the timeout terminates the run or the engine marks the task as terminal. Nodes without a lifecycle message are unaffected.
```
alyssacgoins: Update here how the diagnostic is consumed from `PipelineTask` objects returned by the task-list endpoint, rather than `task_details`.

proposals/12843-pod-lifecycle-failure/README.md
```diff
@@ -0,0 +1,293 @@
...
+
+### End-to-End Data Flow
+
+```mermaid
+flowchart TD
+    A["Execution engine: node message"]
+    --> B["workflow.go: util.NodeStatus.Message"]
+    --> C["api_converter.go: normalize / filter / resolve lifecycle messages"]
+    --> D["task_store.go: Task.LifecycleMessage"]
+    --> E["api_converter.go: PipelineTaskDetail.lifecycle_failure returned in API"]
```
alyssacgoins: replace `PipelineTaskDetail.lifecycle_failure` with the corresponding field on `PipelineTask`.

proposals/12843-pod-lifecycle-failure/README.md
```diff
@@ -0,0 +1,293 @@
...
+
+### Alternative 3: Reuse `RuntimeStatus` or existing error fields without a new column
+
+The `Task` model already has a `Payload` text column carrying serialized state. Stuffing the lifecycle message into the payload was considered but rejected because the payload is a serialized blob and not a queryable field, filtering or reporting on lifecycle failures across runs would require parsing the blob, and a dedicated column makes the database semantics obvious and the `patchTask` overwrite rule (fresh value wins) easy to express. The cost is one nullable column.
+
+### Alternative 4: Reuse the existing `error` field on `PipelineTaskDetail`
+
+The existing `error` field on `PipelineTaskDetail` is documented as "Only populated when the task is in FAILED or CANCELED state." Lifecycle failures need to surface while the task is still in a non-terminal state (the engine may still be retrying). Reusing `error` would violate the API contract and could break existing clients that use it to determine terminal state. A dedicated lifecycle diagnostic field avoids this issue.
```
alyssacgoins: Nit: can you discuss the current status metadata field rather than the removed `PipelineTaskDetail.error`?

alyssacgoins (changes requested):

khushiiagrawal:
@alyssacgoins  I have addressed all the comments, could you please take a look. 
thanks!

alyssacgoins (approved):
/lgtm

alyssacgoins (approved):
/lgtm

PR #13531, opened 2026-06-16
Fix/workflow state on conflict 2

### LINKED ISSUE #13507: [backend] RetryRun fails with concurrent Argo workflow modification (409 Conflict + already exists)

Ayush-kathil:
### What this PR does

This is a follow-up to PR #13508 that adds extra state validation checks during `RetryRun` and `TerminateRun`. 

When the API server encounters a conflict during a retry or terminate operation, it refetches the latest workflow state and tries again. However, in rare cases (like two manual requests racing each other), the first request might successfully move the workflow into a `Running` or terminal state, while the second request refetches it and blindly applies the operation again.

To fix this:
- **RetryRun**: We now verify that the refetched workflow is still in a retryable state (e.g., `Failed` or `Error`) using `.CanRetry()` before trying to update it again.
- **TerminateRun**: We check if the workflow has already reached a final/terminal state (`IsInFinalState()`) before patching the `activeDeadlineSeconds`.

This ensures we don't accidentally retry a running workflow or redundantly terminate an already completed one.

### Related Issues
Follow-up to #13508 
Fixes #13507


droctothorpe:
Hi @Ayush-kathil, thank you for your PR! The linked issue (#13507) must be labeled `ready` before your contribution can be reviewed. Please check our [CONTRIBUTING.md](https://github.com/kubeflow/pipelines/blob/master/CONTRIBUTING.md) for details.

PR #13546, opened 2026-06-17
(fix backend/frontend): Fix debug_pause UI updates and display

carlosdanielarroyo:

**Checklist:**

### DIFF
--- a/backend/src/common/util/workflow.go
+++ b/backend/src/common/util/workflow.go
@@ -908,7 +908,7 @@ func (w *Workflow) PatchTemplateOutputArtifacts() {
 type podSpecPatch struct {
 	Containers []struct {
 		Env []struct {
-			Name string `json:"name"`
+			Name  string `json:"name"`
 			Value string `json:"value"`
 		} `json:"env"`
 	} `json:"containers"`
--- a/backend/src/common/util/workflow.go
+++ b/backend/src/common/util/workflow.go
@@ -903,13 +903,53 @@ func (w *Workflow) PatchTemplateOutputArtifacts() {
 	}
 }
 
+//podSpecPatch is a minimal struct for unmarshaling the pod-spec-patch parameter
+// to detect debug pause env vars injected by set_debug_pause().
+type podSpecPatch struct {
+	Containers []struct {
+		Env []struct {
+			Name string `json:"name"`
+			Value string `json:"value"`
+		} `json:"env"`
+	} `json:"containers"`
+}
+
+//nodeHasDebugPauseEnv returns true when the node's pod-spec-patch input parameter 
+// contains ARGO_DEBUG_PAUSE_AFTER=true or ARGO_DEBUG_PAUSE_BEFORE=true.
+func nodeHasDebugPauseEnv(node workflowapi.NodeStatus) bool {
+	if node.Inputs == nil {
+		return false
+	}
+	for _, p := range node.Inputs.Parameters {
+		if p.Name != "pod-spec-patch" || p.Value == nil {
+			continue
+		}
+		var patch podSpecPatch
+		if err:= json.Unmarshal([]bytte(p.Value.String()), &patch); err != nil {
+			return false
+		}
+		for _, c := range patch.Containers {
+			for _, e := range c.Env { 
+				if e.Value == "true" &&  (e.Name == "ARGO_DEBUG_PAUSE_AFTER" || e.Name == "ARGO_DEBUG_PAUSE_BEFORE") {
+					return true
+				}
+			}
+		}
+	}
+	return false
+}
+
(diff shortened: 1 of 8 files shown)


zazulam:
@carlosdanielarroyo can you build the frontend image with these changes and share how the UI would capture this debug/pause element?

carlosdanielarroyo:
> @carlosdanielarroyo can you build the frontend image with these changes and share how the UI would capture this debug/pause element?

Hi Mike, here is a screenshot showing the new display. And yes will provide an architectural overview of how this works soon. 

Screenshot 2026-06-26 at 3 27 01 PM

zazulam:
cc: @jeffspahr

backend/src/common/util/workflow.go
```diff
@@ -903,13 +903,53 @@ func (w *Workflow) PatchTemplateOutputArtifacts() {
...
+		}
+	}
+	return false
+}
+
 func (w *Workflow) NodeStatuses() map[string]NodeStatus {
 	rev := make(map[string]NodeStatus, len(w.Status.Nodes))
 	for id, node := range w.Status.Nodes {
+		state := string(node.Phase)
+		if node.Phase == workflowapi.NodeRunning && nodeHasDebugPauseEnv(node) {
```
jeffspahr, line 946: This changes every Running Argo node with `ARGO_DEBUG_PAUSE_AFTER` or `ARGO_DEBUG_PAUSE_BEFORE` configured into `Paused`, but those env vars are present for the entire pod lifetime. With the default `set_debug_pause()` path (`after=True`), a normal task will report `Paused` while it is still executing, before it reaches the post-execution pause point. This needs to be driven by a runtime signal that the Argo wait container is actually paused, or else remain `Running` until such a signal exists.

frontend/src/lib/v2/DynamicFlow.ts
```diff
@@ -265,13 +270,14 @@ export function updateFlowElementsState(
...
-        getTaskLabelByPipelineFlowElement(elem),
-        executionLayers,
-      );
+      const taskLabel = getTaskLabelByPipelineFlowElement(elem);
+      const executions = getExecutionsUnderDAG(taskNameToExecution, taskLabel, executionLayers);
       if (executions) {
-        (updatedElem.data as ExecutionFlowElementData).state = executions[0]?.getLastKnownState();
+        const state = executions[0]?.getLastKnownState();
+        (updatedElem.data as ExecutionFlowElementData).state = state;
+        if (state === Execution.State.COMPLETE && debugPauseTasks.has(taskLabel)) {
```
jeffspahr: This has the stale-state version of the same problem in the V2 graph. After an `ARGO_DEBUG_PAUSE_AFTER` task is resumed, MLMD remains `COMPLETE` while downstream tasks can continue and `runIsActive` stays true, so this node keeps rendering as `debugPaused` until the whole run finishes. The paused badge needs to come from a current pause signal, not just `COMPLETE` plus a task that was configured for debug pause.

frontend/src/lib/v2/DynamicFlow.ts
```diff
@@ -478,6 +484,53 @@ function getArtifactNodeKeyToArtifact(
...
+  ][]) {
+    const componentName = taskSpec.componentRef?.name;
+    if (!componentName) continue;
+    const component = spec.components[componentName];
+    if (!component?.executorLabel) continue;
+    const executor = executors[component.executorLabel];
+    const envList = executor?.container?.env;
+    if (!Array.isArray(envList)) continue;
+
+    const hasDebugPause = envList.some(
```
jeffspahr: This includes `ARGO_DEBUG_PAUSE_ON_ERROR` in the configured pause tasks, but the caller only sets `debugPaused` when `state === Execution.State.COMPLETE`. The V2 launcher publishes failed components as MLMD `FAILED`, so `set_debug_pause(on_error=True)` will never show the paused display. The backend and `WorkflowParser` helpers also skip `ON_ERROR`, so the three paths disagree on which debug-pause modes are supported.

jeffspahr (changes requested):
Thanks for working on this. I found a few blockers around the core state model: the current implementation treats debug-pause configuration as if it were the current pause state, which can create false or stale paused UI/API states and also misses the on-error pause case.

jeffspahr:
@carlosdanielarroyo please address and resolve copilot's reviews too.

droctothorpe:
@carlosdanielarroyo: Hey, thank you for your PR! As you are not a member of Kubeflow organization, your PR must link to an issue with 'ready' label. Please read our [CONTRIBUTING.md](https://github.com/kubeflow/pipelines/blob/master/CONTRIBUTING.md), closing this PR

PR #13547, opened 2026-06-17
fix(backend): decouple metric name from SQL column in list page token

kaikaila:
## What this PR does

Decouples the **raw metric name** from the **SQL column name** in the list/pagination page token, giving each its own field with a single, well-defined meaning. This fixes two defects in the prior encoding and removes a field that was doing double duty. It also replaces the page token's shape-based validation (regexes guessing whether a string "looks like" a SQL identifier) with precise, type-aware checks against each model's real field set.

The current-state design reference (token fields, validators, migration behavior) lives in `backend/src/apiserver/list/PAGE_TOKEN_SORT.md`. This PR description is the canonical record of version history and the reasoning behind the current design.

## Background: how the token's sort-field encoding evolved

List APIs paginate with a base64-encoded `token` struct that records both *where to continue* and *how to order*. The sort-field encoding has gone through 3 versions:

| Version | Commit | What changed |
|---|---|---|
| **A** (oldest) | `9685df8ca` and earlier | Baseline — spliced the raw metric name into SQL |
| **B** | `ac3a4c656` (PR #13127) | Closed SQL injection; introduced `SortBySQLColumn` |
| **C** (this PR) | — | Dedicated `SortByMetricName` field; `SortBySQLColumn` removed |

### Version A — the SQL injection vulnerability

A's `addSortByRunMetricToSelect` spliced the metric name (raw user input, also re-derived from the pageToken on later pages) directly into SQL:

```go
Select("selected_runs.*, run_metrics.numbervalue as " + opts.SortByFieldName).        // ① alias
LeftJoin("run_metrics ON ... AND run_metrics.name='" + opts.SortByFieldName + "'")    // ② string literal
```

Both spots are raw string concatenation, no parameterization. PR #13127 fixed this.

### Version B — fixed injection, but overloaded a field

B parameterized the metric name and pinned the alias to a constant:

```go
Select("selected_runs.*, run_metrics.numbervalue as " + model.MetricSortSQLAlias).   // fixed alias
LeftJoin("run_metrics ON ... AND run_metrics.name=?", opts.SortByFieldName)           // bind param
```

Injection closed, but `SortByFieldName` became overloaded — it meant two different things:

- regular sort → a real SQL column name (must pass identifier check)
- metric sort → a raw metric name like `log-loss` (fails identifier check)

### Field layout comparison (B vs C)

| Field | B: sort by CreatedAt | B: sort by metric:accuracy | C: sort by CreatedAt | C: sort by metric:accuracy |
|---|---|---|---|---|
| `SortByFieldName` | `CreatedAtInSec` (column) | `accuracy` (**raw metric**) | `CreatedAtInSec` (column) | `sort_metric_value` (**alias**) |
| `SortBySQLColumn` | `CreatedAtInSec` | `sort_metric_value` (alias) | — (removed) | — (removed) |
| `SortByMetricName` | — (n/a) | — (n/a) | `""` (empty) | `accuracy` (raw metric) |

In short — who holds what:

| | holds "safe column for ORDER BY" | holds "raw metric name" |
|---|---|---|
| **B** | `SortBySQLColumn` | `SortByFieldName` (double duty — the defect) |
| **C** | `SortByFieldName` | `SortByMetricName` (dedicated, clean) |

Migration is a horizontal "shift back into place": B's `SortByFieldName` → C's `SortByMetricName`; B's `SortBySQLColumn` → C's `SortByFieldName`.

## The two defects in version B

1. **Validation conflict.** `SortByFieldName` held a real SQL column for regular sorts but a raw metric name (e.g. `log-loss`, with hyphens) for metric sorts. One field, two incompatible validation rules — tightening the identifier check rejects in-flight metric tokens during a rolling upgrade. ([review](https://github.com/kubeflow/pipelines/pull/12379#discussion_r3073316630))
2. **Token mismatch.** Every metric sort shared the same alias in `SortBySQLColumn`, and `Options.Matches()` compared the wrong field, so a token minted for `metric:accuracy` could be reused for `metric:loss`. ([review](https://github.com/kubeflow/pipelines/pull/12379#discussion_r3059641043))

## The change (version C)

- `SortByFieldName` — **always** the SQL-safe column name (the fixed pivot alias `sort_metric_value` for metric sorts).
- `SortByMetricName` — **dedicated** field for the raw metric name; flows only as a bind parameter. Empty for regular sorts.
- `SortBySQLColumn` — **removed** from the written layout; read only transiently during `unmarshal` to migrate legacy B tokens, then cleared.
- `Options.Matches()` now compares **both** `SortByFieldName` and `SortByMetricName`, fixing defect 2.

## Validation: two independent checks, not one regex doing both jobs

The original implementation validated every page-token field with one of two character-shape regexes (`identifierPattern`, `metricNamePattern`). That conflated two unrelated questions:

- **"Is this string a real field/table name on this model?"** — answered by a new `Listable.IsRegularField(name string) bool` method (added to the interface; each of the 6 models implements it as a one-line delegation to its existing `APIToModelFieldMap()`, via a shared `isRegularFieldIn` helper in `model/common.go`). Used for `KeyFieldName`, `SortByFieldName` (non-metric-alias case), `ModelName` (compared via `GetModelName()` directly), and filter-key column segments. This is strictly more precise than shape-matching: a forged-but-identifier-shaped non-column is correctly rejected, where the old regex would have let it through.
- **"Is this string safe to use as a SQL bind-parameter value?"** — answered by `validateBindValue` (renamed from `validateMetricName`), used only for `SortByMetricName`. This used to be an **allowlist** regex (`metricNamePattern`) restricting which characters a metric name could contain. That was solving the wrong problem: `SortByMetricName` only ever reaches SQL as a `?` bind parameter, so parameterization — not character shape — is what closes injection. The allowlist regex rejected real historical metric names containing `/` or `.` (e.g. `val/loss`, `foo.bar`), which would have been a functional regression with zero security benefit. `validateBindValue` is now a **denylist**-style hygiene check instead: a length bound and rejection of non-printable/control characters, nothing more.

Two more sites got more precise handling as part of this:

- `KeyFieldPrefix`/`SortByFieldPrefix` are table aliases (e.g. `"runs."`), not field names or bind values. Rather than validating an attacker-supplied prefix string, `unmarshal()` now **ignores whatever the token claims and recomputes both unconditionally** from the listable (`GetKeyFieldPrefix()` / `GetSortByFieldPrefix()`) — strictly safer than validation, since the claimed value never reaches SQL.
- Filter keys (e.g. `"pipelines.Name"`) are now validated per-segment: the final (column) segment via `IsRegularField`, any earlier (table-qualifier) segment via `GetModelName()` — matching the same field-vs-table distinction used everywhere else in this PR. (`filter.Filter.ValidateKeys`'s callback signature changed from `func(segment string) error` to `func(segment string, isLast bool) error` to support this.)

## Backward compatibility (rolling upgrades)

`token.unmarshal()` migrates older tokens into the C layout in memory (nothing is re-persisted). The two/three-version classification is now an explicit `tokenVersion` enum with a `detectTokenVersion()` classifier, rather than undifferentiated sequential `if`s.

### Migration decision table

<table>
<tr>
  <th>Detection marker</th>
  <th>Version</th>
  <th>Validators used</th>
  <th>Case</th>
  <th>Migration action</th>
</tr>
<tr>
  <td><code>SortByMetricName != ""</code></td>
  <td><b>C</b></td>
  <td>—</td>
  <td>Any</td>
  <td>None (already current layout)</td>
</tr>
<tr>
  <td rowspan="2"><code>SortBySQLColumn == MetricSortSQLAlias</code></td>
  <td rowspan="2"><b>B</b></td>
  <td><b>Metric validator</b> confirms</td>
  <td>Metric sort</td>
  <td><code>SortByFieldName</code> → <code>SortByMetricName</code>; <code>SortBySQLColumn</code> → <code>SortByFieldName</code></td>
</tr>
<tr>
  <td><b>Metric validator</b> rejects</td>
  <td>—</td>
  <td>Token rejected (<code>InvalidInputError</code>)</td>
</tr>
<tr>
  <td rowspan="3">Neither marker (fallback)</td>
  <td rowspan="3"><b>A</b></td>
  <td><b>Field validator</b> returns <code>true</code></td>
  <td>Real column</td>
  <td>None (already a valid regular sort)</td>
</tr>
<tr>
  <td><b>Field validator</b> returns <code>false</code>, then <b>metric validator</b> confirms</td>
  <td>Metric sort</td>
  <td>Same field shift as B</td>
</tr>
<tr>
  <td>Both validators reject</td>
  <td>Neither field nor metric</td>
  <td>Token rejected (<code>InvalidInputError</code>)</td>
</tr>
</table>

`SortBySQLColumn` is consumed and cleared in all cases so it never propagates.

Note: version B merged only ~2.5 months ago, so most outstanding user-held tokens are still **version A** — the A→C path is the one most tokens will actually exercise.

## Resolved: the version-A identifier-safe metric name limitation

This PR's earlier draft documented a "known limitation, intentional and unfixable in principle": version A tokens carry no field marking them as metric sorts, so the original A→C migration could only guess from whether `SortByFieldName` *looks like* a SQL identifier (`identifierPattern`). That guess worked for obviously-non-identifier metric names (`log-loss`, `val/loss`, which contain `-`/`/`) but **failed** for identifier-safe metric names (`accuracy`, `f1_score` — pure letters/digits/underscore): they passed the regex, so the heuristic couldn't tell them apart from a real column. The token was left as a regular sort, and page 2 errored with `ORDER BY accuracy` (no such column).

This is now fixed, not just narrowed. The full reasoning:

**Why a better regex couldn't have fixed this.** The information needed to disambiguate "the metric named `accuracy`" from "a column literally named `accuracy`" is not derivable from character shape — both are identifier-shaped strings. No regex over `SortByFieldName` alone can resolve that ambiguity, no matter how it's tuned.

**The actual fix: ask the model, not the string.** `unmarshal()`'s A→C branch now calls `listable.IsRegularField(SortByFieldName)` — checking membership in the real, concrete field set of the `Listable` the token belongs to, rather than guessing from shape. `accuracy` is migrated correctly if it is not actually one of Run's columns, full stop.

**Why this didn't require fixing `Run.GetModelName()` first.** `GetModelName()` currently returns `""` for Run (a separate, still-open TODO, unrelated to this fix) and it's tempting to think that TODO was the blocker — it seems like you'd need to know *which* `Listable` type a generic token belongs to before you could ask anything about its fields. But `IsRegularField` doesn't need that: `unmarshal()`'s two real call sites — `NewOptionsFromToken` in `server/list_request_util.go`, and the job-pagination loop in `resource/resource_manager.go` — already have a concrete `Listable` instance in scope when they call into the `list` package. The actual unlock was a plumbing change (threading that instance through to `unmarshal(listable Listable, pageToken string)`), not a `GetModelName()` fix.

**Why a hardcoded Run-column allowlist was tried and rejected.** An earlier attempt at this fix hardcoded a Run-column allowlist directly inside the generic `token.unmarshal()`. That broke because `unmarshal()` is shared by **all** `Listable` types (Pipeline, Experiment, Job, Task, PipelineVersion) — a Run-specific allowlist wrongly migrated other types' legitimate sort columns, and broke the existing `TestNewOptions_FromValidSerializedToken`/`TestTokenSerialization` tests, which intentionally use synthetic field names like `SortField` that a hardcoded allowlist falsely treated as metric names. Asking the listable itself (`listable.IsRegularField`) avoids this entirely: each type answers correctly for its own fields, and the synthetic-`Listable` tests are untouched because they're never asked to satisfy a real model's field map.

**Follow-up (anti-promotion guard).** `IsRegularField` returning `false` only means "not a real column" — it was never a positive signal that the value is a legacy metric name, and using it that way let a forged or replayed A/B token be migrated into a metric-sort token even against a listable (Job, Experiment, Pipeline, PipelineVersion, Task) that has no metric-sort support at all, which would then fail downstream when SQL building tried to reference `MetricSortSQLAlias`. Both the A and B migration branches now additionally confirm via `listable.GetField("metric:" + SortByFieldName)` before promoting, and reject the token otherwise.

## Testing

- `backend/src/apiserver/list/list_test.go`:
  - `TestListSortBehaviorMatrix` — a new table-driven test covering 4 dimensions: page-1 parsing (`NewOptions`), page-2 legacy migration (`unmarshal`), token round-trip fidelity, and `Options.Matches()`. Exercises real model types (Run, Pipeline, Job, Experiment, PipelineVersion, Task), including the previously-failing identifier-safe metric name case and regression guards for `val/loss`/`foo.bar`-style metric names that the old allowlist regex used to reject.
  - `TestValidateBindValue` — direct unit tests for the new denylist-style hygiene check.
  - Existing `TestUnmarshalLegacyTokenMigration_BtoC`/`_AtoC`, `TestNewOptions_FromValidSerializedToken`, `TestTokenSerialization` updated for the new `unmarshal(listable, ...)` signature; the latter two intentionally keep using synthetic (non-real-column) field names.
- `backend/src/apiserver/model/is_regular_field_test.go` — new file, one `IsRegularField` test per model type (known field, unknown field, the metric alias constant).
- `backend/src/apiserver/storage/run_store_test.go`:
  - `TestListRuns_VersionATokenIdentifierSafeMetricResolved` — regression guard confirming `ListRuns` never hits an unknown-column database error for a version-A token with an identifier-safe metric name.
  - `TestListRuns_MetricSortInjectionSafe` — updated to assert the correct safety property: a SQL-injection-shaped metric name is *not* rejected by `validateBindValue` (it's not supposed to be — that's not its job), and is instead rendered harmless by parameterization.
  - `TestListRuns_MetricBindValueHygieneRejected` — new test for `validateBindValue`'s actual job: rejecting control characters and oversized values.

## References

- PR #13127 / commit `ac3a4c656` — version B, the SQL injection fix (security issue HASI2026112-23).
- Review thread on PR #12379 — the chronological record of C's evolution:
  - [Token mismatch](https://github.com/kubeflow/pipelines/pull/12379#discussion_r3059641043) — `Options.Matches()` compared the wrong field.
  - [Backward-compat migration](https://github.com/kubeflow/pipelines/pull/12379#discussion_r3073316630) — token validation not backward compatible with pre-refactor tokens.
  - [Metric-name regex](https://github.com/kubeflow/pipelines/pull/12379#discussion_r3126610194) — new regex stricter than historical metric-name behavior.
  - [Identifier-safe legacy migration](https://github.com/kubeflow/pipelines/pull/12379#discussion_r3397995563) — the A-token ambiguity and its resolution via `listable.IsRegularField`.


backend/src/apiserver/list/list.go
```diff
@@ -124,47 +150,67 @@ func (t *token) unmarshal(pageToken string) error {
...
+	if t.KeyFieldName != "" && !listable.IsRegularField(t.KeyFieldName) {
+		return util.NewInvalidInputError("Invalid key field name: %q", t.KeyFieldName)
 	}
-	if err := validateIdentifierName(t.ModelName, "model name"); err != nil {
-		return err
+	// SortByFieldName is always either a real column or the fixed metric alias.
+	if t.SortByFieldName != "" && t.SortByFieldName != model.MetricSortSQLAlias &&
+		!listable.IsRegularField(t.SortByFieldName) {
+		return util.NewInvalidInputError("Invalid sort field name: %q", t.SortByFieldName)
 	}
```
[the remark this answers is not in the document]
kaikaila: Good catch — confirmed this was a real gap. `IsRegularField` returning `false` only means "not a real column on this model," which we'd already documented elsewhere (see PAGE_TOKEN_SORT_DESIGN.md's "Two independent checks") as *not* a positive signal for "this is a legacy metric name." The version-A and version-B migration branches were using it that way regardless, so a forged/replayed legacy token decoded against a listable with no metric-sort support (Job, Experiment, Pipeline, PipelineVersion, Task — only Run supports `"metric:"` sorting) could be silently promoted into a metric-sort token and fail downstream when SQL building referenced `MetricSortSQLAlias` for a model that has no such column.

Fixed by reusing `Listable.GetField("metric:" + SortByFieldName)` — the same capability check `NewOptions` already applies on page 1 — to gate both the version-A and version-B migration branches before promoting `SortByFieldName`/`SortByMetricName`. If the listable doesn't recognize the name as a metric, `unmarshal()` now returns `util.NewInvalidInputError` instead of migrating. No new interface method was needed since `GetField` already encodes per-model metric support.

Added `DimensionB2_Unmarshal_Page2Migration_RejectsNonMetricListable` in list_test.go covering forged version-A/B tokens against Job, Experiment, and Pipeline (all now rejected), and updated PAGE_TOKEN_SORT_DESIGN.md to document the new check. All existing tests still pass for Run.

backend/src/apiserver/list/list.go
```diff
@@ -467,7 +558,7 @@ func (o *Options) nextPageToken(listable Listable) (*token, error) {
 
 	return &token{
 		SortByFieldName:   o.SortByFieldName,
-		SortBySQLColumn:   o.SortBySQLColumn,
+		SortByMetricName:  o.SortByMetricName,
```
nsingla, line 577: **[CRITICAL]** Rolling upgrade backward compatibility is broken.

New tokens write `SortByMetricName` but no longer write `SortBySQLColumn` (tagged `json:",omitempty"` and never populated). During a rolling deployment, when a **new server** writes a token and an **old server** (pre-this-PR) reads it:

- **Regular sorts:** Old code gates ORDER BY on `SortBySQLColumn != ""`. With `SortBySQLColumn=""`, the ORDER BY clause is **skipped entirely**. Page-2+ results arrive in database-default order — a silent data correctness bug.
- **Metric sorts:** Old code binds `SortByFieldName` as the metric name. With `SortByFieldName` now set to `"sort_metric_value"` (the alias), it queries for a nonexistent metric, producing NULLs.

The reverse direction (old→new) is handled correctly by the migration logic.

**Recommended fix:** Continue writing `SortBySQLColumn` for one release cycle, then remove it in a follow-up PR:

```suggestion
		SortByMetricName:  o.SortByMetricName,
		SortBySQLColumn:   o.SortByFieldName, // backward compat: remove after one release cycle
```

backend/src/apiserver/list/list.go
```diff
@@ -248,12 +317,22 @@ func NewOptions(listable Listable, pageSize int, sortBy string, filter *filter.F
...
 	if len(queryList) > 0 {
 		n, sqlCol, ok := listable.GetField(queryList[0])
 		if ok {
-			token.SortByFieldName = n
-			token.SortBySQLColumn = sqlCol
+			// sqlCol is the SQL-safe column name (for metric sorts it is the
+			// fixed alias MetricSortSQLAlias). n is the original field/metric
+			// name used for value lookups.
+			token.SortByFieldName = sqlCol
+			if n != sqlCol {
```
nsingla, line 335: **[IMPORTANT]** Edge case: metric named `"sort_metric_value"` causes silent misclassification.

If a user creates a metric literally named `"sort_metric_value"` (the value of `MetricSortSQLAlias`), then `Run.GetField("metric:sort_metric_value")` returns `n="sort_metric_value", sqlCol="sort_metric_value"`, making `n != sqlCol` false. The metric branch is never entered, and the token is treated as a regular column sort on a nonexistent column, causing SQL errors downstream.

The same issue affects the B→C and A→C migration branches in `unmarshal()` (lines 164 and 184).

**Practical impact:** Extremely low — naming a metric `"sort_metric_value"` is vanishingly unlikely. But there's no guard.

**Suggested fix:** Reject the reserved name in `Run.GetField` (in `model/run.go` around line 410):
```go
if strings.HasPrefix(name, "metric:") {
    metricName := name[7:]
    if metricName == MetricSortSQLAlias {
        return "", "", false // reserved name
    }
    return metricName, MetricSortSQLAlias, true
}
```

nsingla (commented):
### PR Review Summary

**Scope**: 22 files reviewed across 4 functional groups (core list/pagination, models, filter, storage/server)

#### Impact Assessment
- **Overall Risk**: Medium
- **Breaking Changes**: `NewOptionsFromToken` signature, `Listable` interface addition, `ValidateKeys` callback — all callers updated, no missed callers found
- **Cross-Module Impact**: Fully contained within `backend/src/apiserver/`
- **Security**: Improved — model-aware `IsRegularField` is strictly stronger than regex pattern matching; `SortByMetricName` data flow verified to only reach SQL as a bind parameter

#### Code Quality
- **Critical Issues**: 1 found (rolling upgrade backward compat)
- **Important Issues**: 1 found (MetricSortSQLAlias collision edge case)
- **Suggestions**: 3

#### Inline Comments
5 inline comments posted across 2 files.

#### Positive Observations
- Well-designed separation of SQL-structure validation (`IsRegularField`) from bind-parameter validation (`validateBindValue`)
- `detectTokenVersion()` with explicit enum is much cleaner than sequential if-else
- Comprehensive 4-dimension test matrix across all 6 model types
- Prefix recomputation from the listable (ignoring token claims) is strictly safer
- Excellent PR description with version history table and migration decision table

#### Recommendation
REQUEST_CHANGES — Finding #1 (rolling upgrade backward compat) is a silent data correctness bug during deployments. The fix is a one-line addition with zero risk.

Generated by automated PR review workflow

backend/src/apiserver/list/list.go
```diff
@@ -71,47 +79,65 @@ type token struct {
...
+// characters such as "/" or "." are allowed, since real historical metric
+// names (e.g. "val/loss") use them and were never a SQL-structure risk.
+// Values that are meant to reach SQL structure (column/table names) must
+// use IsRegularField / GetModelName instead, never this function.
+func validateBindValue(name string) error {
 	if name == "" {
 		return nil
 	}
-	if !metricNamePattern.MatchString(name) {
+	if len(name) > maxBindValueLength {
```
nsingla, line 103: **[SUGGESTION]** `len(name)` counts bytes, not runes. A 86-character CJK metric name (3 bytes/char = 258 bytes) would be rejected despite being well under 256 *characters*.

The constant name and error message say "bytes" so the intent is clear, but consider whether rune count (`utf8.RuneCountInString`) would be more user-friendly for non-ASCII metric names. If bytes is intentional (to bound payload size), the current code is correct — just noting the distinction.

backend/src/apiserver/list/list_test.go
```diff
@@ -828,14 +899,239 @@ func TestTokenSerialization(t *testing.T) {
...
+				`{"SortByFieldName":"val/loss","KeyFieldName":"UUID"}`,
+				model.MetricSortSQLAlias, "val/loss",
+			},
+			{
+				"A identifier-safe, now resolved (was the known limitation)",
+				`{"SortByFieldName":"accuracy","KeyFieldName":"UUID"}`,
+				model.MetricSortSQLAlias, "accuracy",
+			},
+			{
+				"Plain C regular field",
```
nsingla, line 1022: **[SUGGESTION]** Consider adding a test case for a crafted version-C token that includes a stale `SortBySQLColumn`. While the code handles this correctly (classified as C, `SortBySQLColumn` cleared at line 195), an explicit test would guard against regressions:

```go
			{
				"C token with stale SortBySQLColumn is cleaned up",
				`{"SortByFieldName":"sort_metric_value","SortByMetricName":"accuracy","SortBySQLColumn":"garbage","KeyFieldName":"UUID"}`,
				model.MetricSortSQLAlias, "accuracy",
			},
```

backend/src/apiserver/list/list_test.go
```diff
@@ -828,14 +899,239 @@ func TestTokenSerialization(t *testing.T) {
...
+					t.Errorf("SortByMetricName = %q, want %q", got.SortByMetricName, tc.wantSortByMetric)
+				}
+				if got.SortBySQLColumn != "" {
+					t.Errorf("SortBySQLColumn = %q, want empty after migration", got.SortBySQLColumn)
+				}
+			})
+		}
+	})
+
+	t.Run("DimensionB2_Unmarshal_Page2Migration_RejectsNonMetricListable", func(t *testing.T) {
```
nsingla, line 1047: **[SUGGESTION]** Consider adding a companion test section for filter key validation with qualified keys (e.g., `"experiments.Name"`) that exercises the `isLast` flag distinction. The existing `TestNewOptionsFromToken_MaliciousFilterKey` covers injection payloads but doesn't verify that the table-qualifier segment uses `GetModelName()` while the column segment uses `IsRegularField()`. A direct test of `ValidateKeys` with a qualified key token would strengthen coverage of the new `isLast` callback parameter introduced in `filter.go`.

backend/src/apiserver/list/list.go
kaikaila: I believe rolling upgrade backward compatibility is necessary for this page token schema change.

When a version B pod deserializes a version C token, SortBySQLColumn defaults to  "". The version B code gates ORDER BY on `SortBySQLColumn != ""` — with an empty string, the ORDER BY clause is silently skipped. Page-2+ results return in arbitrary database order with no error. This is a silent data correctness bug, not an explicit failure that users can detect and retry.

If we're going with rolling upgrade,  I propose the following roadmap:

Release | Action | Rationale
-- | -- | --
baseline+1 (expand) | nextPageToken() writes SortBySQLColumn; unmarshal() adds restore logic | Backward compat with baseline version B pods during rolling upgrade
baseline+2 (contract) | nextPageToken() stops writing SortBySQLColumn; restore logic retained | During rolling upgrade, (baseline+1) pods may still produce compat tokens
baseline+3 (cleanup) | Remove restore logic, remove B→C migration, optionally remove SortBySQLColumn field | No (baseline+2)-era tokens remain in circulation

Given the latest release KFP 2.17.0 as baseline

PR #11672, opened 2025-02-25
feat(backend): Add types for KFP Kubernete Native API

rimolive:
**Description of your changes:**

**Checklist:**
- [ ] You have [signed off your commits](https://www.kubeflow.org/docs/about/contributing/#sign-off-your-commits)
- [ ] The title for your pull request (PR) should follow our title convention. [Learn more about the pull request title convention used in this repository](https://github.com/kubeflow/pipelines/blob/master/CONTRIBUTING.md#pull-request-title-convention).

PR #13552, opened 2026-06-18
fix: allow same-named pipeline versions under different pipelines (#13534, #13535, #13556)

### LINKED ISSUE #13534: Kubernetes store cannot create same-named pipeline versions under different pipelines
### Description

In the Kubernetes pipeline store, `PipelineVersion` custom resources use `pipelineVersion.Name` directly as `metadata.name` (in `FromPipelineVersionModel` in `pipelineversion_types.go`):

```go
ObjectMeta: metav1.ObjectMeta{
    Name:      pipelineVersion.Name,
    Namespace: pipeline.Namespace,
    ...
}
```

Since Kubernetes resource names must be unique per namespace, this enforces **namespace-wide uniqueness** on pipeline version names. However, the data model's unique constraint is `(PipelineId, Name)` — different pipelines in the same namespace should be allowed to have versions with the same name.

If two pipelines try to create a version with the same name (e.g., `"3.5.0"`), the second creation fails because a Kubernetes resource with that `metadata.name` already exists.

### Impact

This prevents any use case where multiple pipelines need identically-named versions — for example, using a release version string as the version name across all managed pipelines.

### Proposed fix

Use a composite `metadata.name` that includes the pipeline name: `{pipelineName}-{versionName}` (e.g., `my-pipeline-3.5.0-ea.2`). The model's `Name` field stays as the bare version name.

For backward compatibility with existing clusters:
- New PipelineVersion CRs are created with composite names
- Lookups try the composite name first, then fall back to the bare name (for CRs created before the fix)
- When found via bare name, verify it belongs to the correct pipeline using the existing `pipelines.kubeflow.org/pipeline-id` label or owner reference

**Files affected:**
- `backend/src/apiserver/storage/pipeline_store_kubernetes.go` — `CreatePipelineVersion`, `GetPipelineVersionByName`
- `api/v2beta1/go/pipelineversion_types.go` — `FromPipelineVersionModel`

### LINKED ISSUE #13535: GetPipelineVersionByName returns wrong version when multiple pipelines have same-named versions
> **Blocked by:** https://github.com/kubeflow/pipelines/issues/13534

### Description

`GetPipelineVersionByName` in `backend/src/apiserver/storage/pipeline_store.go` queries the `pipeline_versions` table by `Name` only, with `LIMIT 1` and no `PipelineId` filter:

```go
func (s *PipelineStore) GetPipelineVersionByName(name string) (*model.PipelineVersion, error) {
    version, err := s.getPipelineVersionByCol("Name", name, model.PipelineVersionReady)
    ...
}
```

The underlying query is:
```sql
SELECT * FROM pipeline_versions WHERE Name = ? AND Status = ? LIMIT 1
```

However, the unique constraint on `pipeline_versions` is a **composite index** on `(PipelineId, Name)` (`uniqueIndex:idx_pipelineid_name`), which means different pipelines are allowed to have versions with the same name. When this happens, `GetPipelineVersionByName` returns whichever version the database happens to pick first, regardless of which pipeline owns it.

### Impact

The only production caller is `LoadSamples` in `backend/src/apiserver/config/config.go`, which uses this method for dedup:

```go
_, fetchErr = resourceManager.GetPipelineVersionByName(pvName)
```

If two pipelines have versions with the same name, the check finds the first one and skips creating the version for the second pipeline.

This is currently hidden because version names default to pipeline names (unique within a namespace), but it becomes a real bug when version names are explicitly set to the same value across different pipelines.

### Comparison

The pipeline equivalent, `GetPipelineByNameAndNamespace`, correctly scopes by both `Name` and `Namespace`.

### Proposed fix

Add `pipelineId string` to the method signature across the interface (`PipelineStoreInterface`), SQL store, Kubernetes store, resource manager, and the single production caller (`LoadSamples`, which already has the pipeline ID in scope as `p.UUID`).

### LINKED ISSUE #13556: LoadSamples skips pipeline version creation when pipeline exists with empty namespace
### What happened?

`LoadSamples` fails to create new pipeline versions on API server restart when a pipeline already exists with `Namespace=""`.

### Root cause

`LoadSamples` looks up pipelines using `common.GetPodNamespace()` directly (`config.go:172`), but creates them without setting `Namespace` on the model — so they are stored with `Namespace=""`. The API server path uses `ReplaceNamespace()` to normalize the namespace before storage, but `LoadSamples` does not.

On restart, the namespace-scoped lookup (using the non-empty `POD_NAMESPACE`) returns NotFound, `CreatePipeline` fails with AlreadyExists (name uniqueness is namespace-agnostic), and the code hits `continue` — skipping version creation entirely.

This affects both standalone and multi-user deployments whenever `POD_NAMESPACE` is set.

### Expected behavior

When `LoadSamples` encounters an existing pipeline (regardless of its stored namespace), it should proceed to create any missing pipeline versions.

### Steps to reproduce

1. Deploy KFP with `POD_NAMESPACE` set to a non-empty value
2. Let `LoadSamples` run on first boot — pipelines are created with `Namespace=""`
3. Restart the API server (or upgrade)
4. `LoadSamples` looks up pipelines with the non-empty `POD_NAMESPACE` — doesn't find them
5. `CreatePipeline` fails with AlreadyExists
6. New pipeline versions are silently skipped

### Proposed fix

1. Use `ReplaceNamespace(GetPodNamespace())` for the lookup, matching the API server convention
2. Handle `AlreadyExists` on `CreatePipeline` by retrying the lookup with empty namespace (backwards compatibility for pre-existing data)

hbelmiro:

Fixes #13534, fixes #13535, fixes #13556 — pipeline versions with the same name under different pipelines were rejected by the Kubernetes store (unique `metadata.name` constraint) and misrouted by the SQL store (`GetPipelineVersionByName` queried by name only, ignoring pipeline ownership). Additionally, `LoadSamples` skipped pipeline version creation when a pre-existing pipeline had an empty namespace and the pod namespace later changed.

### How

- Use composite `metadata.name` = `{pipelineName}-{versionName}` for PipelineVersion CRs
- Add `spec.versionName` field to store the bare version name (required for new CRs)
- Update `GetPipelineVersionByName` to require `pipelineID` across the interface, both store implementations, resource manager, and all callers
- Kubernetes store: try composite name first, fall back to bare name with ownership verification for CRs created before composite naming or managed via GitOps
- SQL store: filter by both `Name` and `PipelineId`
- `LoadSamples`: use `ReplaceNamespace` instead of raw `GetPodNamespace`, and fall back to empty-namespace lookup on `AlreadyExists` to find pre-existing pipelines after a namespace change

### Backward compatibility

- CRs with bare `metadata.name` (created before composite naming, or managed via GitOps) are handled via fallback in `ToModel()`, `GetField()`, and `GetPipelineVersionByName()`
- Ownership verified via OwnerReferences

### Known limitation

Hyphen collision: pipeline "foo" + version "bar-baz" and pipeline "foo-bar" + version "baz" produce the same composite name `foo-bar-baz`. The second creation will fail with "already exists".

**Checklist:**


backend/src/apiserver/storage/pipeline_store.go
```diff
@@ -1055,11 +1055,36 @@ func (s *PipelineStore) GetPipelineVersion(versionId string) (*model.PipelineVer
...
+func (s *PipelineStore) GetPipelineVersionByName(pipelineID, name string) (*model.PipelineVersion, error) {
+	query, args, err := sq.
+		Select(pipelineVersionColumns...).
+		From("pipeline_versions").
+		Where(sq.And{
+			sq.Eq{"pipeline_versions.Name": name},
+			sq.Eq{"pipeline_versions.PipelineId": pipelineID},
+			sq.Eq{"pipeline_versions.Status": model.PipelineVersionReady},
+		}).
+		Limit(1).
```
nsingla: do we need to limit? we don't expect more than 1 in this case, right?

backend/src/crd/kubernetes/v2beta1/pipelineversion_types.go
```diff
@@ -210,15 +216,20 @@ func (p *PipelineVersion) ToModel() (*model.PipelineVersion, error) {
 		}
 	}
 
+	versionName := p.Spec.VersionName
+	if versionName == "" {
+		versionName = p.Name
```
nsingla, line 233: if pipeline versionName isn't speficied, then it uses pipeline.Name, so doesn't that mean for pipelines with existing pipeline version, pipeline versioName is kind of required (if its someone like me who does not like to specify name everytime I upload a pipeline) or my existing scripts will start to break if i am not setting a unique versionName?

backend/src/apiserver/storage/pipeline_store.go
hbelmiro: You're right — removed the LIMIT.

backend/src/crd/kubernetes/v2beta1/pipelineversion_types.go
hbelmiro: Existing scripts won't break. When no name is specified, the API server populates it from the pipeline spec's `pipelineInfo.name` (`resource_manager.go:2000-2004`) before it reaches the store. For legacy CRs without `spec.versionName`, `ToModel()` and `GetField()` fall back to `metadata.name`. The CRD field is intentionally optional.

nsingla: but `pipelineInfo.name` is not unique, which is what I was referring to, that if a user tries to upload a new pipeline version to an existing pipeline with a version with name `pipelineInfo.name`, then it will fail

backend/src/crd/kubernetes/v2beta1/pipelineversion_types.go
```diff
@@ -137,7 +142,7 @@ func FromPipelineVersionModel(pipeline model.Pipeline, pipelineVersion model.Pip
 
 	return &PipelineVersion{
 		ObjectMeta: metav1.ObjectMeta{
-			Name:      pipelineVersion.Name,
+			Name:      pipeline.Name + "-" + pipelineVersion.Name,
```
nsingla: **[IMPORTANT]** Composite name construction is duplicated here without using the `compositeVersionName()` helper from `pipeline_store_kubernetes.go`. This inline concatenation bypasses both the bare-name DNS-1123 validation and the composite-name DNS-1123 validation that the helper provides.

This works today only because callers happen to call `compositeVersionName()` separately before calling `FromPipelineVersionModel`. But any future caller that forgets to pre-validate will produce invalid Kubernetes resource names silently.

This also conflicts with the repo's code reuse policy in CLAUDE.md: *"Always reuse existing functions, helpers, and utilities before writing new code."*

**Suggestion**: Have `FromPipelineVersionModel` call `compositeVersionName()` (may need to move the helper to a shared package or to this file):
```go
compositeName, err := compositeVersionName(pipeline.Name, pipelineVersion.Name)
if err != nil {
    return nil, err
}
// use compositeName in ObjectMeta.Name
```

nsingla (commented):
### PR Review Summary

**Scope**: 10 files reviewed across backend storage, CRD types, config, and manifests.

**Overall Risk**: Medium — the core fix is correct and well-tested, but a few issues are worth addressing.

**Note**: `backend/src/common/client/api_server/v2/pipeline_upload_client_kubernetes.go` calls `FromPipelineVersionModel` (lines 165 and 258) but is not in this PR's changed files. Since `FromPipelineVersionModel` now produces composite `metadata.name` (`{pipelineName}-{versionName}` instead of `{versionName}`), the `UploadPipeline` path — where pipeline and initial version often share the same name — will produce CRs named `{name}-{name}`. Please verify this is the intended behavior.

**Positives**:
- Thorough backward compatibility for legacy CRs via fallback in `ToModel()`, `GetField()`, and `GetPipelineVersionByName()`
- Excellent test coverage: 12+ new test cases covering composite lookup, legacy compat, hyphen collision, cross-pipeline isolation, orphan prevention, and namespace switching
- Atomicity guard in `CreatePipelineAndPipelineVersion` validates composite name before creating Pipeline CR
- Known hyphen collision limitation is honestly documented and tested

backend/src/apiserver/storage/pipeline_store_kubernetes.go
```diff
@@ -367,15 +383,45 @@ func (k *PipelineStoreKubernetes) GetPipelineVersion(pipelineVersionId string) (
 	return pipelineVersion.ToModel()
 }
 
-func (k *PipelineStoreKubernetes) GetPipelineVersionByName(name string) (*model.PipelineVersion, error) {
-	pipelineVersion := v2beta1.PipelineVersion{}
-
+func (k *PipelineStoreKubernetes) GetPipelineVersionByName(pipelineID, name string) (*model.PipelineVersion, error) {
 	if common.GetPodNamespace() == "" {
 		return nil, fmt.Errorf("Error returning the pod namespace. Ensure you have POD_NAMESPACE environment variable set in the API Server pod.")
```
nsingla: **[SUGGESTION]** Every call to `GetPipelineVersionByName` now fetches the full Pipeline CR from Kubernetes just to get `pipeline.Name` for constructing the composite name. This doubles the K8s API calls for this operation.

Since the Kubernetes client uses an informer cache, the practical impact may be low. But worth noting — if this is called in a hot loop (e.g., `LoadSamples` with many pipelines), consider passing the pipeline name alongside `pipelineID` or caching the result.

backend/src/apiserver/storage/pipeline_store_kubernetes.go
```diff
@@ -385,6 +431,12 @@ func (k *PipelineStoreKubernetes) GetPipelineVersionByName(name string) (*model.
 		return nil, err
 	}
 
+	// Legacy CRs may carry ownership via OwnerReference, label, or both.
+	if !pipelineVersion.IsOwnedByPipeline(pipelineID) &&
```
nsingla: **[IMPORTANT]** The ownership check uses OR semantics — a legacy CR passes if *either* `IsOwnedByPipeline` (OwnerReference UID match) *or* the label matches. These are independent checks, so a CR where the OwnerReference points to pipeline A but the label was manually edited to point to pipeline B would incorrectly pass when queried with B's ID.

**Suggestion**: Require the OwnerReference check to pass when OwnerReferences exist, falling back to label-only when they don't:
```go
if !pipelineVersion.IsOwnedByPipeline(pipelineID) {
    if len(pipelineVersion.OwnerReferences) > 0 ||
        pipelineVersion.Labels["pipelines.kubeflow.org/pipeline-id"] != pipelineID {
        return nil, util.NewResourceNotFoundError("PipelineVersion", name)
    }
}
```

backend/src/apiserver/resource/resource_manager.go
```diff
@@ -2046,9 +2046,9 @@ func (r *ResourceManager) GetPipelineVersion(pipelineVersionId string) (*model.P
...
 }
 
-// GetPipelineVersionByName returns a pipeline version by Name. Tags are loaded at the store level.
-func (r *ResourceManager) GetPipelineVersionByName(name string) (*model.PipelineVersion, error) {
-	pipelineVersion, err := r.pipelineStore.GetPipelineVersionByName(name)
+// GetPipelineVersionByName returns a pipeline version by pipeline ID and version name. Tags are loaded at the store level.
+func (r *ResourceManager) GetPipelineVersionByName(pipelineID, name string) (*model.PipelineVersion, error) {
+	pipelineVersion, err := r.pipelineStore.GetPipelineVersionByName(pipelineID, name)
 	if err != nil {
 		return nil, util.Wrapf(err, "Failed to get a pipeline version with name %v", name)
```
nsingla: **[SUGGESTION]** The error message still only includes `name` but not the new `pipelineID` parameter, making production debugging harder.

```go
return nil, util.Wrapf(err, "Failed to get a pipeline version with pipelineID=%v and name=%v", pipelineID, name)
```

backend/src/apiserver/storage/pipeline_store_kubernetes.go
```diff
@@ -367,15 +383,45 @@ func (k *PipelineStoreKubernetes) GetPipelineVersion(pipelineVersionId string) (
...
+
+	namespace := common.GetPodNamespace()
+	pipelineVersion := v2beta1.PipelineVersion{}
+
+	// Try composite name first ({pipelineName}-{versionName})
+	compositeName := pipeline.Name + "-" + name
+	if errs := validation.IsDNS1123Subdomain(compositeName); len(errs) == 0 {
+		err = k.client.Get(context.TODO(), ctrlclient.ObjectKey{
+			Namespace: namespace,
+			Name:      compositeName,
```
nsingla, line 391: **[SUGGESTION]** When the composite name fails DNS-1123 validation, this manufactures a "not found" error to fall through to the bare-name lookup. While the control flow is correct, it masks the actual reason (invalid composite name). A debug-level log here would help troubleshooting:
```go
glog.V(4).Infof("Composite name %q is not a valid DNS-1123 subdomain, skipping to bare-name lookup", compositeName)
```

backend/src/apiserver/config/config.go
```diff
@@ -169,22 +169,33 @@ func LoadSamples(resourceManager *resource.ResourceManager, sampleConfigPath str
...
-					// of when API server starts.
-					glog.Warningf(fmt.Sprintf(
-						"Failed to create pipeline for %s. Error: %v", cfg.Name, configErr))
-					continue
+					if util.IsUserErrorCodeMatch(configErr, codes.AlreadyExists) {
+						// Retry without namespace filter. In standalone mode this
+						// finds the empty-namespace pipeline; safe because sample
+						// names are deterministic and collisions across namespaces
+						// are not expected for managed pipelines.
+						p, fetchErr = resourceManager.GetPipelineByNameAndNamespace(cfg.Name, "")
```
alyssacgoins, line 188: this additional `GetPipelineByNameAndNamespace` call doubles the API server calls for a common operation - thoughts on caching options?

backend/src/apiserver/storage/pipeline_store_kubernetes.go
```diff
@@ -30,6 +31,23 @@ var (
 	ErrUnsupportedField = errors.New("the field is unsupported")
 )
 
+func compositeVersionName(pipelineName, versionName string) (string, error) {
```
alyssacgoins: Would be useful to have a docstring here explaining logic

alyssacgoins (changes requested):

backend/src/apiserver/config/config.go
hbelmiro: This second call only fires on the error recovery path: first fetch (with namespace) → NotFound → create → AlreadyExists → retry fetch (without namespace). It handles the namespace-switching scenario (e.g., standalone mode transitioning from empty namespace to a real one, or concurrent pod restarts).

Since this runs once at startup over a small, deterministic set of managed pipelines — and only on the AlreadyExists branch — caching would add complexity without meaningful benefit here.

backend/src/apiserver/storage/pipeline_store_kubernetes.go
```diff
@@ -367,24 +363,58 @@ func (k *PipelineStoreKubernetes) GetPipelineVersion(pipelineVersionId string) (
...
 	if err != nil {
 		if k8serrors.IsNotFound(err) {
-			return nil, util.NewResourceNotFoundError("PipelineVersion", name)
+			return nil, util.NewResourceNotFoundError("PipelineVersion", versionName)
 		}
 		return nil, err
 	}
 
+	// Legacy CRs may carry ownership via OwnerReference, label, or both.
+	if !pipelineVersion.IsOwnedByPipeline(pipelineID) {
```
mprahl, line 416: When would there not be an OwnerReference?

hbelmiro: Never. Removed.

backend/src/apiserver/storage/pipeline_store_kubernetes.go
```diff
@@ -205,12 +205,8 @@ func (k *PipelineStoreKubernetes) CreatePipelineAndPipelineVersion(pipeline *mod
 	pipeline.UUID = ""
 	pipelineVersion.UUID = ""
 
-	// Validate the pipeline version name before creating either resource to ensure atomicity.
-	if errs := validation.IsDNS1123Subdomain(pipelineVersion.Name); len(errs) > 0 {
-		return nil, nil, util.NewInvalidInputError(
-			"Invalid pipeline version name %q: %s. Use 'display_name' for human-readable labels",
-			pipelineVersion.Name, strings.Join(errs, "; "),
-		)
+	if _, err := v2beta1.NewPipelineVersionName(pipeline.Name, pipelineVersion.Name); err != nil {
```
mprahl: ** 🤖 AI Review **

This create path now relies on the composite Kubernetes object name for uniqueness. That lets a post-upgrade legacy CR with bare metadata.name `v1` coexist with a new CR named `{pipeline}-v1` under the same pipeline, so the same pipeline can end up with two versions named `v1`. Please add an explicit pre-create lookup by `(pipelineID, pipelineName, versionName)` before create, or otherwise reject the legacy/new duplicate case, and cover it with a regression test.

backend/src/crd/kubernetes/v2beta1/pipelineversion_types.go
```diff
@@ -94,7 +100,47 @@ type PipelineVersionList struct {
...
+				compositeName, strings.Join(errs, "; "),
+			)
+		}
+	}
+	return PipelineVersionName{pipelineName: pipelineName, versionName: versionName}, nil
+}
+
+func (n PipelineVersionName) Name() string {
+	if n.pipelineName != "" {
+		return n.pipelineName + "-" + n.versionName
```
mprahl: ** 🤖 AI Review **

This raw `{pipelineName}-{versionName}` encoding is ambiguous when both parts may contain hyphens: `(foo, bar-baz)` and `(foo-bar, baz)` both become `foo-bar-baz`. That turns previously valid cross-pipeline version creations into `AlreadyExists` failures. Please switch `metadata.name` to an unambiguous stable encoding, for example a bounded hash or suffix, instead of plain concatenation.

manifests/kustomize/base/crds/pipelines.kubeflow.org_pipelineversions.yaml
```diff
@@ -57,6 +57,10 @@ spec:
                 additionalProperties:
                   type: string
                 type: object
+              versionName:
+                maxLength: 253
```
mprahl, line 61: ** 🤖 AI Review **

`spec.versionName` is no longer independently valid up to 253 characters once `metadata.name` is derived from `{pipelineName}-{versionName}`. A long existing pipeline name can now make an otherwise CRD-valid version fail later in the API server because the combined DNS-1123 name exceeds 253 chars. Please either stop using raw concatenation for the durable Kubernetes name, or tighten and document the CRD and API validation so users get one consistent limit with a migration path.

backend/src/crd/kubernetes/v2beta1/pipelineversion_types.go
hbelmiro: Same-named versions under the same pipeline still fail with "already exists," but that's expected and was already the case before this PR

backend/src/crd/kubernetes/v2beta1/pipelineversion_types.go
hbelmiro: Addressed — `compositeVersionName()` replaced by `NewPipelineVersionName` type in the `v2beta1` package. `FromPipelineVersionModel` now calls it directly. No duplication remains.

backend/src/apiserver/storage/pipeline_store_kubernetes.go
hbelmiro: Addressed — `GetPipelineVersionByName` now accepts `(pipelineID, pipelineName, versionName)`. Pipeline name is passed through from callers, no extra K8s API call.

backend/src/apiserver/storage/pipeline_store_kubernetes.go
hbelmiro: Addressed — removed the label fallback entirely. Ownership is now determined solely by OwnerReferences (present since #11672).

backend/src/apiserver/resource/resource_manager.go
hbelmiro: Addressed — error now includes both parameters: `"Failed to get a pipeline version with pipelineID=%v and name=%v"`.

backend/src/apiserver/storage/pipeline_store_kubernetes.go
hbelmiro: Addressed — added `glog.V(4).Infof(...)` when the composite name fails DNS-1123 validation.

backend/src/apiserver/storage/pipeline_store_kubernetes.go
hbelmiro: `compositeVersionName` was replaced by the `PipelineVersionName` type with `NewPipelineVersionName` constructor — self-documenting via constructor validation and `Name()` method.

backend/src/crd/kubernetes/v2beta1/pipelineversion_types.go
hbelmiro: This is a known limitation documented in the PR description. The collision is extremely rare — it requires two pipelines whose names differ only at a hyphen boundary (e.g. `foo` and `foo-bar`) with complementary version names (`bar-baz` and `baz`). When it does happen, K8s returns a clear `AlreadyExists` error; the case is covered by `TestCreatePipelineVersion_HyphenCollision`.

A hash or random suffix would actually make the system less robust: two identical creates (`pipeline1` + `version1`) would produce different `metadata.name` values, bypassing the natural K8s `AlreadyExists` guard and requiring full reliance on the pre-create lookup for dedup.

manifests/kustomize/base/crds/pipelines.kubeflow.org_pipelineversions.yaml
hbelmiro: Existing versions are unaffected. They use legacy bare-name `metadata.name`.

backend/src/apiserver/storage/pipeline_store_kubernetes.go
```diff
@@ -665,14 +691,6 @@ func (k *PipelineStoreKubernetes) getK8sPipelineVersion(ctx context.Context, pip
...
-	if errs := validation.IsDNS1123Subdomain(pipelineVersion.Name); len(errs) > 0 {
-		return nil, util.NewInvalidInputError(
-			"Invalid pipeline version name %q: %s. Use 'display_name' for human-readable labels",
-			pipelineVersion.Name, strings.Join(errs, "; "),
-		)
-	}
-
 	k8sPipelineVersion, err := v2beta1.FromPipelineVersionModel(*pipeline, *pipelineVersion)
 	if err != nil {
 		return nil, util.NewBadRequestError(err, "Invalid pipeline spec")
```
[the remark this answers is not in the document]
hbelmiro: Addressed — `FromPipelineVersionModel` errors that are already `*UserError` (e.g. `InvalidArgument` from naming validation) now pass through instead of being re-wrapped as `BadRequestError`. `AlreadyExists` error now includes the composite K8s resource name.

backend/src/apiserver/storage/pipeline_store_kubernetes.go
```diff
@@ -367,24 +363,54 @@ func (k *PipelineStoreKubernetes) GetPipelineVersion(pipelineVersionId string) (
...
-func (k *PipelineStoreKubernetes) GetPipelineVersionByName(name string) (*model.PipelineVersion, error) {
-	pipelineVersion := v2beta1.PipelineVersion{}
-
+func (k *PipelineStoreKubernetes) GetPipelineVersionByName(pipelineID, pipelineName, versionName string) (*model.PipelineVersion, error) {
 	if common.GetPodNamespace() == "" {
 		return nil, fmt.Errorf("Error returning the pod namespace. Ensure you have POD_NAMESPACE environment variable set in the API Server pod.")
 	}
 
+	namespace := common.GetPodNamespace()
+	pipelineVersion := v2beta1.PipelineVersion{}
```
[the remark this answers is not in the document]
hbelmiro: Addressed — `GetPipelineVersionByName` now derives the namespace from the pipeline (via `getK8sPipeline`) instead of hardcoding `POD_NAMESPACE`.

backend/src/apiserver/storage/pipeline_store_kubernetes.go
hbelmiro: Addressed — added a pre-create lookup via `GetPipelineVersionByName` (checks both composite and bare names) before `k.client.Create`, so a legacy bare-name CR and a new composite CR with the same logical version name are correctly rejected. Covered by `TestCreatePipelineVersion_DuplicateLegacyBareName`.

backend/src/apiserver/storage/pipeline_store_kubernetes.go
```diff
@@ -665,27 +693,33 @@ func (k *PipelineStoreKubernetes) getK8sPipelineVersion(ctx context.Context, pip
...
-	}
-
 	k8sPipelineVersion, err := v2beta1.FromPipelineVersionModel(*pipeline, *pipelineVersion)
 	if err != nil {
+		var userError *util.UserError
+		if errors.As(err, &userError) {
+			return nil, err
+		}
 		return nil, util.NewBadRequestError(err, "Invalid pipeline spec")
 	}
```
nsingla, line 708: **[IMPORTANT]** Redundant k8s List call on every version creation.

`GetPipelineVersionByName` calls `getK8sPipeline(pipelineID)` internally, which does a List-all-pipelines-then-filter-by-UID. But `createPipelineVersionWithPipeline` already receives the `pipeline *model.Pipeline` parameter, and its caller (`CreatePipelineVersion`) already resolved the pipeline via `GetPipeline`.

Consider an internal variant that accepts the already-known namespace to avoid the extra List round-trip on every version creation:

```go
func (k *PipelineStoreKubernetes) getPipelineVersionByNameInNamespace(
    namespace, pipelineID, pipelineName, versionName string,
) (*model.PipelineVersion, error) { ... }
```

backend/src/crd/kubernetes/v2beta1/pipelineversion_types.go
```diff
@@ -95,6 +98,15 @@ type PipelineVersionList struct {
 }
 
 func FromPipelineVersionModel(pipeline model.Pipeline, pipelineVersion model.PipelineVersion) (*PipelineVersion, error) {
+	if pipelineVersion.Name == "" {
```
nsingla, line 102: **[SUGGESTION]** This returns a raw `fmt.Errorf`, but the `NewPipelineVersionName` call two lines below returns a `util.UserError` (via `NewInvalidInputError`). The caller in `createPipelineVersionWithPipeline` checks `errors.As(err, &userError)` to decide whether to wrap as `BadRequestError` — a raw error takes the wrapping path, which works but produces a less specific error code.

Consider `util.NewInvalidInputError(...)` for consistency so both validation failures return the same error type.

backend/src/apiserver/storage/pipeline_store.go
```diff
@@ -1055,11 +1055,35 @@ func (s *PipelineStore) GetPipelineVersion(versionId string) (*model.PipelineVer
 	return version, nil
 }
```
nsingla, line 1057: **[SUGGESTION]** Minor: a one-line comment explaining why `pipelineName` is unused here would help future readers understand the interface design:

```go
// pipelineName is unused by the SQL store; it exists in the interface for the
// Kubernetes store, which needs it to construct composite CR names.
func (s *PipelineStore) GetPipelineVersionByName(pipelineID, _, versionName string) (*model.PipelineVersion, error) {
```

nsingla:
**[IMPORTANT]** `FromPipelineVersionModel` caller not updated in this PR.

`backend/src/common/client/api_server/v2/pipeline_upload_client_kubernetes.go` — at lines 165–191 (`Upload` method), both `pipelineModel.Name` and `pipelineVersionModel.Name` are set to the same `name` variable:

```go
pipelineModel := apimodel.Pipeline{
    Name: name,  // line 166
    ...
}
pipelineVersionModel := apimodel.PipelineVersion{
    Name: name,  // line 185
    ...
}
pipelineVersion, err := k8sapi.FromPipelineVersionModel(pipelineModel, pipelineVersionModel)
```

After this PR, `FromPipelineVersionModel` produces composite `metadata.name = {pipelineName}-{versionName}`. Since both are `name`, the resulting CR will be named `{name}-{name}` (e.g., `my-pipeline-my-pipeline`).

This is the common `UploadPipeline` path where the initial version shares the pipeline's name. The CR name is cosmetically redundant but also means:
- `ToModel()` will recover the bare name from `spec.versionName` correctly, so the model layer is fine
- But any code that reads `metadata.name` directly (kubectl, controller watches, logging) will see the doubled name

The `UploadPipelineVersion` path (line 288–297) is fine because the version name is independently derived.

**Suggestion**: Either update this file to set `pipelineVersionModel.Name` to a distinct initial version name, or document that `{name}-{name}` is the expected composite CR name for the upload path.

backend/src/apiserver/storage/pipeline_store_kubernetes.go
```diff
@@ -367,24 +364,55 @@ func (k *PipelineStoreKubernetes) GetPipelineVersion(pipelineVersionId string) (
 	return pipelineVersion.ToModel()
 }
 
-func (k *PipelineStoreKubernetes) GetPipelineVersionByName(name string) (*model.PipelineVersion, error) {
+func (k *PipelineStoreKubernetes) GetPipelineVersionByName(pipelineID, pipelineName, versionName string) (*model.PipelineVersion, error) {
```
mprahl: I don't think we need `pipelineName` in the function signature because the database one doesn't use it and the Kubernetes one fetches the pipeline by ID anyways and can get the name.

backend/src/crd/kubernetes/v2beta1/pipeline_version_name.go
```diff
@@ -0,0 +1,57 @@
...
+		return PipelineVersionName{}, util.NewInvalidInputError(
+			"Invalid pipeline version name %q: %s. Use 'display_name' for human-readable labels",
+			versionName, strings.Join(errs, "; "),
+		)
+	}
+	if pipelineName != "" {
+		compositeName := pipelineName + "-" + versionName
+		if errs := validation.IsDNS1123Subdomain(compositeName); len(errs) > 0 {
+			return PipelineVersionName{}, util.NewInvalidInputError(
+				"The combined pipeline and version name %q exceeds Kubernetes naming limits: %s",
```
mprahl: Can you tell the user what the limit is?

backend/src/apiserver/storage/pipeline_store_kubernetes.go
```diff
@@ -367,24 +364,55 @@ func (k *PipelineStoreKubernetes) GetPipelineVersion(pipelineVersionId string) (
...
+			// collision). Fall through to bare-name lookup for backward compat.
+		} else if !k8serrors.IsNotFound(err) {
+			return nil, err
+		}
 	}
 
-	err := k.client.Get(context.TODO(), ctrlclient.ObjectKey{
-		Namespace: common.GetPodNamespace(),
-		Name:      name,
+	// Fallback: try bare name (legacy-style CRs created before composite naming)
```
mprahl: Legacy is not quite accurate. When deploying via GitOps, this path is also used.

backend/src/apiserver/storage/pipeline_store_kubernetes.go
```diff
@@ -367,24 +364,55 @@ func (k *PipelineStoreKubernetes) GetPipelineVersion(pipelineVersionId string) (
...
+
+	namespace := k8sPipeline.Namespace
 	pipelineVersion := v2beta1.PipelineVersion{}
 
-	if common.GetPodNamespace() == "" {
-		return nil, fmt.Errorf("Error returning the pod namespace. Ensure you have POD_NAMESPACE environment variable set in the API Server pod.")
+	// Try composite name first ({pipelineName}-{versionName})
+	pvName, pvNameErr := v2beta1.NewPipelineVersionName(pipelineName, versionName)
+	if pvNameErr != nil {
+		glog.V(4).Infof("Composite name for pipeline %q version %q is not a valid DNS-1123 subdomain, skipping to bare-name lookup", pipelineName, versionName)
```
mprahl: Nit: Is this if branch useful? It seems like we could skip validation on get. We assume it's created in a valid way either through GitOps or the API.

mprahl (changes requested):

mprahl (approved):
/approve
/hold for other reviews

backend/src/apiserver/storage/pipeline_store_kubernetes.go
hbelmiro: Agreed — `createPipelineVersionWithPipeline` already calls `getPipelineVersionByNameInNamespace` directly, no redundant List.

backend/src/crd/kubernetes/v2beta1/pipelineversion_types.go
hbelmiro: Addressed — now uses `util.NewInvalidInputError` for consistency.

backend/src/apiserver/storage/pipeline_store.go
hbelmiro: Addressed differently — removed `pipelineName` from the interface entirely. The K8s store derives it via `getK8sPipeline`.

backend/src/apiserver/storage/pipeline_store_kubernetes.go
hbelmiro: Agreed — removed `pipelineName` from the interface, both implementations, and all callers. The internal `getPipelineVersionByNameInNamespace` still accepts it so `createPipelineVersionWithPipeline` can pass it without re-fetching.

backend/src/crd/kubernetes/v2beta1/pipeline_version_name.go
hbelmiro: Done — error now reads "exceeds the Kubernetes 253-character naming limit".

backend/src/apiserver/storage/pipeline_store_kubernetes.go
hbelmiro: Updated — comments now say "CRs created before composite naming, or managed via GitOps".

backend/src/apiserver/storage/pipeline_store_kubernetes.go
hbelmiro: Agreed — replaced validation with direct string concatenation on the get path.

hbelmiro:
> **[IMPORTANT]** `FromPipelineVersionModel` caller not updated in this PR.
> 
> `backend/src/common/client/api_server/v2/pipeline_upload_client_kubernetes.go` — at lines 165–191 (`Upload` method), both `pipelineModel.Name` and `pipelineVersionModel.Name` are set to the same `name` variable:
> 
> ```go
> pipelineModel := apimodel.Pipeline{
>     Name: name,  // line 166
>     ...
> }
> pipelineVersionModel := apimodel.PipelineVersion{
>     Name: name,  // line 185
>     ...
> }
> pipelineVersion, err := k8sapi.FromPipelineVersionModel(pipelineModel, pipelineVersionModel)
> ```
> 
> After this PR, `FromPipelineVersionModel` produces composite `metadata.name = {pipelineName}-{versionName}`. Since both are `name`, the resulting CR will be named `{name}-{name}` (e.g., `my-pipeline-my-pipeline`).
> 
> This is the common `UploadPipeline` path where the initial version shares the pipeline's name. The CR name is cosmetically redundant but also means:
> 
>     * `ToModel()` will recover the bare name from `spec.versionName` correctly, so the model layer is fine
> 
>     * But any code that reads `metadata.name` directly (kubectl, controller watches, logging) will see the doubled name
> 
> 
> The `UploadPipelineVersion` path (line 288–297) is fine because the version name is independently derived.
> 
> **Suggestion**: Either update this file to set `pipelineVersionModel.Name` to a distinct initial version name, or document that `{name}-{name}` is the expected composite CR name for the upload path.

@nsingla Done. `FromPipelineVersionModel` now uses bare `metadata.name` when pipeline and version share the same name, avoiding the redundant `{name}-{name}` pattern. The `CreatePipelineAndPipelineVersion` pre-validation skips composite checks in the same case.

nsingla (approved):
/lgtm

PR #13554, opened 2026-06-18
fix(sdk): resolve symlinks when archiving embedded artifacts

### LINKED ISSUE #13533: [backend] Embedded Artifact does not resolve symlinks
### Environment

OpenshiftAI cluster installation
*  KFP version: **latest RHOAI**

*  KFP SDK version: **2.16.1**

```
kfp                                       2.16.1
kfp-components                            1.11.0      <editable_location>
kfp-kubernetes                            2.16.1
kfp-pipeline-spec                         2.16.0
kfp-server-api                            2.16.0
```

### Steps to reproduce

Minimal reproducible example

1) create files to be included as embedded artifacts
    ```sh
    # Setup: real file + symlinked directory
    mkdir -p /tmp/very_local_path
    echo "template content" > /tmp/very_local_path/template.txt

    mkdir -p /tmp/assets
    ln -s /tmp/very_local_path/template.txt /tmp/assets/template.txt
    ```

2) create the minimal component and pipeline using embedded artifact
    ```python
    from kfp import dsl, compiler
    from kfp.client import Client

    @dsl.component(
        base_image="python:3.12",
        embedded_artifact_path="/tmp/assets/template.txt",
    )
    def my_component(embedded_artifact: dsl.EmbeddedInput[dsl.Dataset]):
        print(embedded_artifact.path)  # this will still be a symlink making the component fail on cluster


    @dsl.pipeline
    def my_pipeline():
        my_component()

    compiler.Compiler().compile(my_pipeline, "pipeline.yaml")

    client = Client(host="https://<kfp-endpoint>")
    client.create_run_from_pipeline_package("pipeline.yaml")
    # RuntimeError: Failed to extract embedded archive
    ```

The compiled pipeline.yaml now contains a tar.gz archive where _template.txt_ is a symlink to _/tmp/very_local_path/template.txt_ instead of the actual file content.
At runtime this fails because _/tmp/very_local_path/_ does not exist inside the container.

### Expected result

An embedded artifact should contain the expected file's content.

### Materials and Reference

---

Priyanshu-u07:

Resolves symlinks when creating tar archives for embedded artifacts. Previously, `tarfile.open()` preserved symlinks in the archive, causing `RuntimeError: Failed to extract embedded archive` at runtime when the symlink target didn't exist inside the container.

Fixes #13533 
Changes:

- Added _safe_add_to_tar() helper that walks the artifact tree with controlled symlink resolution: Resolves the top-level `embedded_artifact_path`, Dereferences in-tree file symlinks and Rejects dangling, escaping, and cyclic symlinks (ancestor-based (`st_dev, st_ino`) cycle detection)

- Added `_get_archive()` test helper to extract and decompress the embedded archive from a compiled component
- Added regression tests: symlinked file, directory with in-tree symlink, cyclic symlink, nested escaping symlink, dangling symlink, and valid sibling symlink
- This fix only affects the SDK compile-time logic. The backend is unchanged

Note: The issue is labeled [backend] but the bug is in the Python SDK's archiving logic, not the backend API server.

### DIFF
--- a/sdk/python/kfp/dsl/component_factory.py
+++ b/sdk/python/kfp/dsl/component_factory.py
@@ -13,6 +13,7 @@
 # limitations under the License.
 import base64
 import dataclasses
+import errno
 import gzip
 import inspect
 import io
@@ -503,6 +504,73 @@ CONTAINERIZED_PYTHON_COMPONENT_COMMAND = [
 ]
 
 
+def _safe_add_to_tar(
+    tar: tarfile.TarFile,
+    root: pathlib.Path,
+    arcname: str = '.',
+) -> None:
+    """Add a file or directory to a tar archive with controlled symlink
+    resolution.
+
+    Resolves the top-level *root* path, then walks its contents:
+    - Regular files are added directly.
+    - In-tree file symlinks are dereferenced (archived as regular files).
+    - Dangling symlinks raise ``ValueError``.
+    - Files whose resolved path escapes the artifact tree raise
+      ``ValueError``.
+    - Directory symlinks are rejected unconditionally.
+
+    This helper operates on trusted local input and does not attempt to
+    enforce source-tree immutability against concurrent modification.
+    """
+    resolved_root = root.resolve()
+
+    if resolved_root.is_file():
+        tar.add(str(resolved_root), arcname=arcname)
+        return
+
+    if not resolved_root.is_dir():
+        raise ValueError(f'Embedded artifact path does not exist or is not a '
+                         f'file/directory: {root}')
+
+    def _walk(
+        current: pathlib.Path,
+        arc_prefix: str,
+    ) -> None:
+        try:
+            resolved = current.resolve(strict=True)
+        except RuntimeError:
+            raise ValueError(f'Symlink cycle detected at: {current}')
+        except OSError as e:
+            if e.errno == errno.ELOOP:
+                raise ValueError(f'Symlink cycle detected at: {current}')
+            if e.errno in (errno.ENOENT, errno.ENOTDIR):
+                raise ValueError(f'Dangling symlink: {current}')
+            raise
+
+        if resolved.is_dir():
+            if current != resolved:
+                raise ValueError(
+                    f'Directory symlinks are not supported: {current}')
+
+            tar.add(str(resolved), arcname=arc_prefix, recursive=False)
+            for child in sorted(resolved.iterdir()):
+                child_arc = (f'{arc_prefix}/{child.name}'
+                             if arc_prefix != '.' else child.name)
+                _walk(child, child_arc)
+        elif resolved.is_file():
+            try:
+                resolved.relative_to(resolved_root)
+            except ValueError:
+                raise ValueError(
+                    f'Path escapes artifact tree: {current} -> {resolved}')
+            tar.add(str(resolved), arcname=arc_prefix)
+        else:
+            raise ValueError(f'Unsupported file type: {current}')
+
+    _walk(resolved_root, arcname)
+
+
 def _get_command_and_args_for_lightweight_component(
     func: Callable,
     additional_funcs: Optional[List[Callable]] = None,
@@ -534,9 +602,9 @@ def _get_command_and_args_for_lightweight_component(
         buf = io.BytesIO()
         with tarfile.open(fileobj=buf, mode='w') as tar:
             if asset_path.is_dir():
-                tar.add(asset_path, arcname='.')
+                _safe_add_to_tar(tar, asset_path, arcname='.')
             else:
-                tar.add(asset_path, arcname=asset_path.name)
+                _safe_add_to_tar(tar, asset_path, arcname=asset_path.name)
 
         archive_bytes = buf.getvalue()
 
--- a/sdk/python/kfp/dsl/component_factory_test.py
+++ b/sdk/python/kfp/dsl/component_factory_test.py
@@ -411,5 +411,259 @@ class TestPythonEOLWarning(unittest.TestCase):
                 pass
 
 
+def _get_archive(component_fn):
+    """Extract the raw tar bytes from a compiled component's embedded
+    archive."""
+    import base64
+    import gzip
+
+    source = component_fn.component_spec.implementation.container.command[-1]
+    match = re.search(r"__KFP_EMBEDDED_ARCHIVE_B64\s*=\s*'([^']+)'", source)
+    if not match:
+        raise ValueError(
+            'Could not locate __KFP_EMBEDDED_ARCHIVE_B64 in compiled '
+            'component source.')
+    archive_b64 = match.group(1)
+
+    compressed = base64.b64decode(archive_b64.encode('ascii'))
+    return gzip.decompress(compressed)
+
+
+class TestEmbeddedArtifactSymlinkResolution(unittest.TestCase):
+    """Tests for #13533: embedded artifacts should resolve symlinks."""
+
+    def test_symlinked_file_is_dereferenced_in_archive(self):
+        """A symlinked file should be archived as a regular file, not a
+        symlink."""
+        import io
+        import os
+        import tarfile
+        import tempfile
+
+        with tempfile.TemporaryDirectory() as tmpdir:
+            real_dir = os.path.join(tmpdir, 'real')
+            os.makedirs(real_dir)
+            real_file = os.path.join(real_dir, 'data.txt')
+            with open(real_file, 'w') as f:
+                f.write('real content')
+
+            link_file = os.path.join(tmpdir, 'link.txt')
+            try:
+                os.symlink(real_file, link_file)
+            except OSError:
+                self.skipTest('Cannot create symlinks on this platform')
+
+            @component(
+                base_image='python:3.11',
+                embedded_artifact_path=link_file,
+                install_kfp_package=False,
+            )
+            def my_comp(cfg: dsl.EmbeddedInput[dsl.Dataset]):
+                pass
+
+            raw = _get_archive(my_comp)
+            with tarfile.open(fileobj=io.BytesIO(raw), mode='r') as tar:
+                members = tar.getmembers()
+                self.assertEqual(len(members), 1)
+                member = members[0]
+                self.assertFalse(member.issym(),
+                                 'Symlink was not dereferenced in archive')
+                self.assertFalse(member.islnk(),
+                                 'Hardlink was not dereferenced in archive')
+                self.assertTrue(member.isfile())
+                with tar.extractfile(member) as extracted:
+                    self.assertIsNotNone(extracted)
+                    self.assertEqual(extracted.read().decode(), 'real content')
+
+    def test_directory_with_symlinked_file_is_dereferenced(self):
+        """When a directory is embedded and it contains a symlinked file whose
+        target is inside the artifact tree, the archive should contain the real
+        file content."""
+        import io
+        import os
+        import tarfile
+        import tempfile
+
+        with tempfile.TemporaryDirectory() as tmpdir:
+            embed_dir = os.path.join(tmpdir, 'assets')
+            os.makedirs(embed_dir)
+
+            real_file = os.path.join(embed_dir, 'original.txt')
+            with open(real_file, 'w') as f:
+                f.write('original content')
+
+            link_in_dir = os.path.join(embed_dir, 'linked.txt')
+            try:
+                os.symlink(real_file, link_in_dir)
+            except OSError:
+                self.skipTest('Cannot create symlinks on this platform')
+
+            @component(
+                base_image='python:3.11',
+                embedded_artifact_path=embed_dir,
+                install_kfp_package=False,
+            )
+            def my_comp(cfg: dsl.EmbeddedInput[dsl.Dataset]):
+                pass
+
+            raw = _get_archive(my_comp)
+            with tarfile.open(fileobj=io.BytesIO(raw), mode='r') as tar:
+                members = tar.getmembers()
+                for member in members:
+                    self.assertFalse(
+                        member.issym(),
+                        f'Member {member.name} is still a symlink')
+                    self.assertFalse(
+                        member.islnk(),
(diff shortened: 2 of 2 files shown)


sdk/python/kfp/dsl/component_factory.py
```diff
@@ -532,7 +532,7 @@ def _get_command_and_args_for_lightweight_component(
             )
 
         buf = io.BytesIO()
-        with tarfile.open(fileobj=buf, mode='w') as tar:
+        with tarfile.open(fileobj=buf, mode='w', dereference=True) as tar:
```
jeffspahr: `dereference=True` applies to the entire recursive archive walk, not only to the explicitly selected file symlink. If an embedded directory contains a symlink to itself or an ancestor, compilation follows it repeatedly and fails with `ELOOP`/`ENAMETOOLONG`; a nested directory symlink can also silently pull files from outside the selected artifact tree into the compiled pipeline. Please use controlled symlink handling: resolve the top-level `embedded_artifact_path`; while walking a directory, dereference in-tree file symlinks, reject dangling or escaping links, and either reject directory symlinks or traverse them with inode-based cycle detection. Please add regression coverage for a cycle, a nested link escaping the root, and a dangling link. The implementation and tests belong in the two files this PR already changes, but this needs a helper plus tests, so it is not a safe one-line GitHub suggestion.

jeffspahr (changes requested):
The direct symlink case is worth supporting, but blanket recursive dereferencing needs safety boundaries before merge. Please resolve the inline finding and add regression coverage for cyclic and escaping nested symlinks.

sdk/python/kfp/dsl/component_factory_test.py
```diff
@@ -411,5 +411,237 @@ def foo():
...
+    """Extract the raw tar bytes from a compiled component's embedded
+    archive."""
+    import base64
+    import gzip
+
+    source = component_fn.component_spec.implementation.container.command[-1]
+    marker = "__KFP_EMBEDDED_ARCHIVE_B64 = '"
+    start = source.index(marker) + len(marker)
+    end = source.index("'", start)
+    archive_b64 = source[start:end]
```
jeffspahr: This force-push reintroduces the fixed-marker parsing that was already changed in response to the earlier resolved review thread. Please restore the resilient parsing and explicit failure; this suggestion is the previously reviewed form.

```suggestion
    match = re.search(r"__KFP_EMBEDDED_ARCHIVE_B64\s*=\s*'([^']+)'", source)
    if not match:
        raise ValueError(
            'Could not locate __KFP_EMBEDDED_ARCHIVE_B64 in compiled '
            'component source.')
    archive_b64 = match.group(1)
```

sdk/python/kfp/dsl/component_factory_test.py
```diff
@@ -411,5 +411,237 @@ def foo():
...
+                self.assertEqual(len(members), 1)
+                member = members[0]
+                self.assertFalse(member.issym(),
+                                 'Symlink was not dereferenced in archive')
+                self.assertFalse(member.islnk(),
+                                 'Hardlink was not dereferenced in archive')
+                self.assertTrue(member.isfile())
+                extracted = tar.extractfile(member)
+                self.assertIsNotNone(extracted)
+                self.assertEqual(extracted.read().decode(), 'real content')
```
jeffspahr: This also reintroduces the previously resolved unclosed extraction stream. Please keep the returned file object scoped to a context manager.

```suggestion
                with tar.extractfile(member) as extracted:
                    self.assertIsNotNone(extracted)
                    self.assertEqual(extracted.read().decode(), 'real content')
```

sdk/python/kfp/dsl/component_factory_test.py
```diff
@@ -411,5 +411,237 @@ def foo():
...
+                linked_member = next(
+                    (m for m in members
+                     if m.name.endswith('linked.txt')), None)
+                self.assertIsNotNone(
+                    linked_member,
+                    'linked.txt not found in archive')
+                extracted = tar.extractfile(linked_member)
+                self.assertIsNotNone(extracted)
+                self.assertEqual(extracted.read().decode(),
+                                 'original content')
```
jeffspahr: Please restore the same context-managed extraction here as well.

```suggestion
                with tar.extractfile(linked_member) as extracted:
                    self.assertIsNotNone(extracted)
                    self.assertEqual(extracted.read().decode(),
                                     'original content')
```

jeffspahr (changes requested):
The controlled archive walker addresses the prior symlink-safety feedback, and all 29 focused tests pass. I found no remaining production correctness issue. Before approval, please fix the deterministic YAPF and docformatter failures on this head and restore the previously resolved test-helper cleanups below. Once those checks are green and the inline comments are addressed, this looks ready.

sdk/python/kfp/dsl/component_factory.py
Priyanshu-u07: Done

sdk/python/kfp/dsl/component_factory_test.py
Priyanshu-u07: Done

sdk/python/kfp/dsl/component_factory_test.py
Priyanshu-u07: Done

sdk/python/kfp/dsl/component_factory_test.py
Priyanshu-u07: Done

Priyanshu-u07:
Thanks for the review @jeffspahr! I've addressed all your review comments, updated PR description and fixed the CI failures. PTAL.

jeffspahr (approved):
Re-reviewed at b5b2221307f6da728c27788930b2ac8574a20810. No actionable findings. The controlled symlink traversal, cycle/escape/dangling coverage, resilient archive parsing, and context-managed extraction streams are all in place. All 29 focused tests and pinned local formatting checks pass, and GitHub CI is green.

jeffspahr:
Needs an /approve from @mprahl @droctothorpe @zazulam

sdk/python/kfp/dsl/component_factory.py
```diff
@@ -503,6 +503,75 @@ def parse_docstring_with_return_as_args(
...
+    if not resolved_root.is_dir():
+        raise ValueError(f'Embedded artifact path does not exist or is not a '
+                         f'file/directory: {root}')
+
+    def _walk(
+        current: pathlib.Path,
+        arc_prefix: str,
+        ancestors: frozenset,
+    ) -> None:
+        resolved = current.resolve()
```
jeffspahr: **[P2] Normalize direct symlink-loop failures across supported Python versions**

`current.resolve()` runs before the inode-cycle check. On Python 3.9–3.12, a self-referential or mutual symlink loop such as `a -> b -> a` raises an uncaught `RuntimeError` here, so the helper does not uphold its documented behavior that cycles raise `ValueError`. On Python 3.13, the same graph resolves to an unresolved path and is instead classified as dangling. The existing `loop -> ancestor directory` test resolves successfully and does not cover this case.

Please add a direct/self-referential symlink-loop regression test and normalize the resolution failure so cycle handling is consistent across supported Python versions.

jeffspahr (commented):
One additional symlink-cycle edge case is not handled consistently across the SDK's supported Python versions.

sdk/python/kfp/dsl/component_factory.py
Priyanshu-u07: Done, wrapped `resolve()` in `try/except RuntimeError` to normalize cycle handling across Python 3.9–3.13 and added `test_self_referential_symlink_raises_error` regression test. PTAL.

sdk/python/kfp/dsl/component_factory.py
```diff
@@ -503,6 +503,78 @@ def parse_docstring_with_return_as_args(
...
+        raise ValueError(f'Embedded artifact path does not exist or is not a '
+                         f'file/directory: {root}')
+
+    def _walk(
+        current: pathlib.Path,
+        arc_prefix: str,
+        ancestors: frozenset,
+    ) -> None:
+        try:
+            resolved = current.resolve()
```
jeffspahr: **[P2] Keep cycle classification consistent on Python 3.13**

Non-strict `current.resolve()` only raises `RuntimeError` for this graph on older supported Python versions. On Python 3.13, a self-referential link resolves to the unresolved path, so the subsequent `exists()` check reports it as `Dangling symlink` rather than a cycle. The new test permits either `cycle` or `dangling` despite its docstring promising consistent cycle handling. Please use strict resolution and distinguish `ELOOP` from a genuinely missing target, or otherwise normalize this case so cycles are classified consistently across supported Python versions.

jeffspahr (commented):
The original uncaught-exception path is fixed and the focused/CI coverage is green. One cross-version cycle-classification issue remains inline.

sdk/python/kfp/dsl/component_factory.py
```diff
@@ -503,6 +503,78 @@ def parse_docstring_with_return_as_args(
...
+            st = resolved.stat()
+            key = (st.st_dev, st.st_ino)
+            if key in ancestors:
+                raise ValueError(f'Symlink cycle detected at: {current}')
+
+            if current.is_symlink():
+                try:
+                    resolved.relative_to(resolved_root)
+                except ValueError:
+                    raise ValueError(f'Symlink escapes artifact tree: '
```
jeffspahr: **[P2] Reject nested directory aliases instead of recursively following them**

`Path.is_symlink()` does not identify Windows directory junctions, so an out-of-tree junction can resolve to a directory and skip this containment branch entirely. Even for contained POSIX symlinks, ancestor-only inode tracking prevents cycles but permits acyclic link DAGs to multiply traversal and archive size exponentially. Since directory aliases are not necessary to fix the original file-symlink bug, the safer policy is to reject any nested directory whose canonical path differs from the path encountered during the walk.

```suggestion
            if current != resolved:
                raise ValueError(
                    f'Directory symlinks are not supported: {current}')
```

jeffspahr (commented):
Security conclusion: following contained file symlinks is reasonable for a trusted, stable local compile-time source tree. I do not think this helper should attempt to enforce source-tree immutability; race-proof enforcement would require a filesystem snapshot or descriptor/handle-relative traversal and is disproportionate here, especially because compilation already executes the user's Python code.

Nested directory-link traversal is different: it creates avoidable static risks. Windows junctions are not reported by `Path.is_symlink()`, so they can bypass the current containment branch, and ancestor-only cycle detection still permits acyclic link DAGs to amplify traversal combinatorially. I recommend rejecting nested directory aliases, applying containment independently of link-type detection for files, and adding entry/byte/depth limits as defense in depth. The helper should be documented as operating on trusted local input, not as a boundary against a hostile concurrent filesystem writer.

Separately, explicitly using tarfile's `data` extraction filter where the runtime supports it would be worthwhile defense in depth, but that can be handled outside this focused traversal fix.

sdk/python/kfp/dsl/component_factory.py
```diff
@@ -503,6 +503,78 @@ def parse_docstring_with_return_as_args(
...
+                child_arc = (f'{arc_prefix}/{child.name}'
+                             if arc_prefix != '.' else child.name)
+                _walk(child, child_arc, child_ancestors)
+        elif resolved.is_file():
+            if current.is_symlink():
+                try:
+                    resolved.relative_to(resolved_root)
+                except ValueError:
+                    raise ValueError(f'Symlink escapes artifact tree: '
+                                     f'{current} -> {resolved}')
```
jeffspahr: **[P2] Enforce file containment independently of link detection**

Containment is the security invariant; whether the platform classifies the encountered path as a symbolic link should not decide whether it is checked. Applying the resolved-path check to every file also covers additional reparse-point behavior without weakening regular-file handling.

```suggestion
            try:
                resolved.relative_to(resolved_root)
            except ValueError:
                raise ValueError(
                    f'Path escapes artifact tree: {current} -> {resolved}')
```

sdk/python/kfp/dsl/component_factory.py
Priyanshu-u07: Done, switched to `resolve(strict=True)` and added `errno.ELOOP` handling so cycles are consistently classified across Python 3.9–3.13.

sdk/python/kfp/dsl/component_factory.py
Priyanshu-u07: Done

sdk/python/kfp/dsl/component_factory.py
Priyanshu-u07: Done

jeffspahr (commented):
The prior cycle-normalization and security feedback is otherwise addressed, and the focused tests plus GitHub CI are green. One filesystem error-classification issue remains inline.

sdk/python/kfp/dsl/component_factory.py
```diff
@@ -503,6 +504,71 @@ def parse_docstring_with_return_as_args(
...
+        arc_prefix: str,
+    ) -> None:
+        try:
+            resolved = current.resolve(strict=True)
+        except RuntimeError:
+            raise ValueError(f'Symlink cycle detected at: {current}')
+        except OSError as e:
+            if e.errno == errno.ELOOP:
+                raise ValueError(f'Symlink cycle detected at: {current}')
+            raise ValueError(f'Dangling symlink: {current}')
```
jeffspahr: **[P2] Preserve non-dangling filesystem errors**

This converts every strict-resolution failure other than `ELOOP` into `ValueError("Dangling symlink")`, including permission, I/O, and path-length failures. I reproduced this with a symlink into an unreadable directory: compilation reported a dangling link instead of the underlying `PermissionError`. Please classify only missing/invalid targets as dangling and propagate other `OSError` values so users retain the actionable cause.

```suggestion
        except OSError as e:
            if e.errno == errno.ELOOP:
                raise ValueError(f'Symlink cycle detected at: {current}')
            if e.errno in (errno.ENOENT, errno.ENOTDIR):
                raise ValueError(f'Dangling symlink: {current}')
            raise
```

PR #13459, opened 2026-05-31
chore(ci): support argo workflows 4.x

jeffspahr:
## Summary

- Promote the bundled/default Argo Workflows dependency to v4.0.5 across manifests, backend defaults, marketplace templates, install scripts, and docs.
- Migrate backend Argo imports to `github.com/argoproj/argo-workflows/v4` and update v4 API call sites for workflow validation and compression handling.
- Keep Argo 4 CI coverage current by moving v4 lanes to v4.0.5 and widening the support-matrix freshness script to read both E2E and API server workflow matrices.
- Drop EOL Argo v3.5/v3.6 compatibility lanes from the declared support matrix and keep the legacy line on the latest supported v3.7 patch, v3.7.14.

## Testing

- `go test ./backend/src/common/util ./backend/src/apiserver/client ./backend/src/apiserver/template ./backend/src/v2/compiler/argocompiler ./backend/src/crd/controller/scheduledworkflow/... ./backend/src/agent/persistence/...`
- `go test $(go list ./backend/... | grep -v backend/test/v2/api | grep -v backend/test/integration | grep -v backend/test/v2/integration | grep -v backend/test/initialization | grep -v backend/test/v2/initialization | grep -v backend/test/compiler | grep -v backend/test/end2end)` passed on rerun after one transient `backend/src/v2/component` log-artifact read failure.
- `./bin/ginkgo -v ./backend/test/compiler` passed: 299 passed, 4 skipped, 0 failed. The locally installed CLI emitted a version mismatch warning because `make ginkgo` installed v2.29.0 while the repo imports v2.27.2.
- `PYTHONPATH=/tmp/kfp-argo-matrix-pkgs python3 .github/scripts/verify-argo-matrix.py`
- `ruby -e "require 'yaml'; YAML.load_file('.github/workflows/e2e-test.yml'); YAML.load_file('.github/workflows/api-server-tests.yml'); puts 'workflow YAML parsed'"`
- `kubectl kustomize` for `manifests/kustomize/third-party/argo/installs/{cluster,namespace,namespace/cluster-scoped}`
- `git diff --check`
- stale-reference scan for `v4.0.4`, Argo v3 imports/images, v3 `ARGO_VERSION` references, and removed `v3.5`/`v3.6` matrix references

Issue #13202, opened 2026-04-03
[feature] Add Argo Workflows 4.x compatibility


jeffspahr:
### Feature Area

/area backend
/area testing

### What feature would you like to see?

Add support for Argo Workflows 4.x in KFP while preserving compatibility with the current Argo 3.x support window during the migration period.

Today this repo documents support for Argo Workflows 3.5, 3.6, and 3.7, with Argo 3.7 as the default bundled/version-plumbed line. Argo Workflows 4.x is now released upstream, and KFP needs a tracked effort to validate and add 4.x support without regressing existing 3.x users.

Proposed scope:
- add at least one Argo 4.x CI lane while keeping a representative Argo 3.7 lane green
- update vendored Argo manifests, CRDs, RBAC, deploy/version plumbing, and support documentation
- verify compile, upload/run, retry, recurring runs, logs/artifacts, archived workflows, and metadata-writing flows against Argo 4.x
- only introduce narrow compatibility handling where Argo 3.x and 4.x behavior diverges, rather than a broad provider abstraction

### What is the use case or pain point?

Users who want to run KFP on clusters with current Argo Workflows releases do not have a supported path today. Supporting only Argo 3.x forces operators either to pin to older Argo versions or to run KFP against an untested Argo 4.x controller.

This is also a good forcing function to identify and harden the most version-sensitive integration seams in KFP, especially around workflow status interpretation, retry behavior, logs/artifacts, metadata labels, and recurring-run interactions.

### Is there a workaround currently?

The current workaround is to stay on an Argo 3.7 deployment or to try Argo 4.x without an official compatibility claim. Neither is ideal for operators who need newer Argo fixes and features.

---

Love this idea? Give it a 👍.

jeffspahr:
Local spike results from an isolated Argo 4.0.4 probe:

- Added local CI spike lanes for `v4.0.4` in `.github/workflows/e2e-test.yml` and `.github/workflows/api-server-tests.yml` as explicit `include` jobs, while leaving the default 3.x matrices unchanged.
- `third_party/argo` manifest/test/doc update targets rewrite cleanly to `v4.0.4`.
- `kubectl kustomize --load-restrictor LoadRestrictionsNone` succeeds for both the vendored Argo base and the standalone KFP overlay with Argo 4 refs.
[...]

PR #13555, opened 2026-06-18
test(backend): cover Argo 4 runtime compatibility

jeffspahr:
## Summary

- add focused API tests for disabled recurring-run creation and ScheduledWorkflow persistence
- exercise KFP run-level retry against Argo Workflows 4
- verify task execution metadata, artifact IDs, and archived log fallback after pod deletion
- run the new coverage only in the existing canonical standalone Argo 4.0.5 API matrix job

## Why

PR #13459 added the Argo 4 dependency and CI matrices, but issue #13202 also calls out runtime behaviors that were not asserted directly: recurring runs, run-level retry, logs/artifacts, archived workflows, and metadata writing. These tests cover those behaviors without adding another E2E job or matrix dimension.

Related to #13202.

## Impact

The existing `backend/test/end2end` jobs are unchanged. The tests execute only in the existing standalone Kubernetes 1.35.0 / Argo 4.0.5 / cache-enabled / no-proxy / no-TLS API job; other API jobs skip them before resource creation.

## Verification

- `go test ./backend/test/v2/api -run '^$'`
- backend unit suite excluding cluster, compiler, and E2E packages
- `go test ./backend/src/apiserver/resource -run '^TestRetryRun'`
- `go test ./backend/src/apiserver/archive`
- `go test ./backend/src/apiserver/server -run '^(TestReadRunLogV1|TestRetryRun|TestArchiveRun)'`
- parsed `.github/workflows/api-server-tests.yml` with Ruby YAML
- `git diff --check`

The new cluster-backed scenarios require the API Server Tests workflow because no KFP cluster is deployed in the local worktree.


zazulam:
@jeffspahr can you rebase with master, this recently [merged PR](https://github.com/kubeflow/pipelines/pull/13576) should help with the mlflow ci test

zazulam:
GPT 5.5 Inspected:

The test is failing because the **retried MLflow failure run never transitions back from `RUNNING` to `FAILED` in the KFP API**, even though the underlying `fail_v2` pipeline is expected to fail.

### Actual failed test

From the JUnit report, the failing spec is:

```text
Should reopen MLflow runs on retry and then reflect the retried status
```

Failure:

```text
Pipeline run with id=2abb2dcb-e102-455f-82fb-dd6709aaf60d did not reach one of [FAILED] within timeout, current state: RUNNING
In [It] at: backend/test/end2end/mlflow_e2e_test.go:498
```

The [relevant test code](https://github.com/kubeflow/pipelines/blob/41f6c3eda895ec9056a188da18aef4da2215797d/backend/test/end2end/mlflow_e2e_test.go#L494-L500) is:

```go
// Retry the run
e2e_utils.RetryPipelineRun(runClient, createdRun.RunID)

// Wait for the retried run to reach terminal state
testutil.WaitForRunToBeInState(runClient, &createdRun.RunID, []run_model.V2beta1RuntimeState{
	run_model.V2beta1RuntimeStateFAILED,
}, &timeout)
```

### Why the `fail-pipeline` logs are misleading

The `fail-pipeline-*` logs at the tail are expected. The test intentionally uses:

```text
test_data/sdk_compiled_pipelines/valid/failing/fail_v2.yaml
```

That pipeline calls `sys.exit(1)`, so both the original run and the retried run should end as `FAILED`.

The initial run does correctly reach `FAILED`:

```text
Pipeline run ... reached expected state FAILED
Retried Pipeline Run, runId=2abb2dcb-e102-455f-82fb-dd6709aaf60d
```

The failure happens **after retry**: the KFP run stays `RUNNING` for the full timeout.

### Likely root cause

This points to a retry-state synchronization bug, not a bad test pipeline:

1. `RetryRun` resets the run to `RUNNING`.
2. The retried Argo workflow/pod runs and fails again, as expected.
3. The KFP persistence path does not update the run back to `FAILED`.
4. By cleanup time, the report says no Workflow CR can be found for that run ID:

```text
WARNING: no workflows found for run IDs; skipping
```

So the API has a stale run state: **KFP DB says `RUNNING`, while the retried workflow is already gone or no longer discoverable.**

The [retry code](https://github.com/kubeflow/pipelines/blob/41f6c3eda895ec9056a188da18aef4da2215797d/backend/src/apiserver/resource/resource_manager.go#L1155-L1169) resets the workflow and run here:

```go
// Notify plugins of retry
if run.PluginsOutputString != nil && *run.PluginsOutputString != "" {
	if pr, prErr := apiservermlflow.ModelToPersistedRun(run, namespace); prErr == nil {
		r.pluginDispatcher.OnRunRetry(ctx, pr)
	}
}

condition := string(newExecSpec.ExecutionStatus().Condition())
run.Conditions = condition
run.FinishedAtInSec = 0
run.WorkflowRuntimeManifest = model.LargeText(newExecSpec.ToStringForStore())
run.State = model.RuntimeState(condition).ToV2()
// OnRunRetry persists plugin output independently; leave PluginsOutput unchanged here.
run.PluginsOutputString = nil
err = r.runStore.UpdateRun(run)
```

That correctly sets the run back to `RUNNING`, but the later terminal-state report is what appears to be missed.

### Recommended fix

Fix the retry/persistence path so a retried workflow cannot be deleted or skipped before the KFP run is updated to its retried terminal state.

A good defensive fix is to make `RetryRun` explicitly clear any final-state labels and completion labels on the live workflow before/while updating it, then verify the DB update happens after the retry workflow is accepted.

The [retry generation](https://github.com/kubeflow/pipelines/blob/41f6c3eda895ec9056a188da18aef4da2215797d/backend/src/common/util/workflow.go#L209-L216) already removes labels from the stored manifest:

```go
newWF := w.DeepCopy()
// Delete/reset fields which indicate workflow completed
delete(newWF.Labels, common.LabelKeyCompleted)
// Delete/reset fields which indicate workflow is finished being persisted to the database
delete(newWF.Labels, LabelKeyWorkflowPersistedFinalState)
newWF.Labels[common.LabelKeyPhase] = string(workflowapi.NodeRunning)
newWF.Status.Phase = workflowapi.WorkflowRunning
newWF.Status.Conditions.UpsertCondition(workflowapi.Condition{Status: metav1.ConditionFalse, Type: workflowapi.ConditionTypeCompleted})
```

But given the CI behavior, the safer server-side patch is to ensure the **live** workflow’s final-state labels are cleared before it is handed back to Argo:
`backend/src/apiserver/resource/resource_manager.go`
```go
// Before updating/creating the retry workflow, ensure labels that cause
// persistence-agent skip/delete behavior are cleared on the object being written.
labels := newExecSpec.ExecutionObjectMeta().GetLabels()
delete(labels, util.LabelKeyWorkflowPersistedFinalState)
delete(labels, common.LabelKeyCompleted)
labels[common.LabelKeyPhase] = string(workflowapi.NodeRunning)
newExecSpec.ExecutionObjectMeta().Labels = labels
```

Then keep the existing `UpdateRun` to mark the DB run as `RUNNING`.

### Test-side improvement

The test should also assert that retry actually created a new retried workflow/update before waiting for terminal state. Right now it only waits for the same run ID to become `FAILED` again, which gives poor diagnostics when persistence misses the final update.

Add a post-timeout diagnostic around line 498 that dumps:

- current run state
- state history
- `K8SName`
- workflow lookup by `pipeline/runid`
- workflow phase/labels if present

For example:

`backend/test/end2end/mlflow_e2e_test.go`

```go
timeout := time.Duration(maxPipelineWaitTime)
testutil.WaitForRunToBeInState(runClient, &createdRun.RunID, []run_model.V2beta1RuntimeState{
	run_model.V2beta1RuntimeStateFAILED,
}, &timeout)

retriedRun := testutil.GetPipelineRun(runClient, &createdRun.RunID)
Expect(retriedRun.State).NotTo(BeNil())
Expect(*retriedRun.State).To(Equal(run_model.V2beta1RuntimeStateFAILED),
	"Retried pipeline run should still be FAILED (fail_v2 always fails)")
```

If this continues to fail, the issue is not MLflow validation itself; it is that **retried run final-state reporting is not reliably persisted after `RetryRun`**.

Note: the code-search results I inspected may be incomplete due result limits; more matching code can be reviewed in GitHub’s code search for `RetryRun`, `GenerateRetryExecution`, and `LabelKeyWorkflowPersistedFinalState`.

zazulam:
Given that this test is failing on master as well, I'm fine with merging as is and addressing the RetryRun issue for the MLFlow e2e integration test in a separate PR.
/lgtm
/approve

PR #13562, opened 2026-06-19
fix(backend) stream metrics artifact extraction

jeffspahr:
## Summary

- require `mlpipeline-metrics` archives to contain exactly one regular tar entry
- stream that entry directly into incremental JSON decoding instead of materializing a file map and full decompressed string
- retain the first 50 metrics while continuing to validate the complete bounded input stream
- enforce the existing 64-character metric-name contract while decoding
- add configurable `MAX_METRICS_FILE_BYTES` protection, defaulting to 1 MiB

## Why

ADA-KUBEFL-06 identified that metrics tar extraction used an unbounded `io.ReadAll`, allowing a small compressed artifact to cause excessive memory consumption. Generic archive entry and aggregate quotas are unnecessary here because the metrics artifact already has a stricter one-file contract.

This change applies that semantic contract first, retains only bounded metric data, and keeps a configurable decompressed-byte ceiling for pathological JSON whitespace or encoding.

Valid metrics continue to accept both `numberValue` and `number_value`. Inputs with more than 50 metrics preserve the existing behavior of reporting the first 50. Invalid metric-name patterns continue through the existing API reporting path; only contract-breaking lengths are rejected during decoding.

## Validation

- `go test -count=1 ./backend/src/common/...`
- `go test -count=1 ./backend/src/agent/persistence/worker`
- `go vet ./backend/src/common/util`
- `git diff --check origin/master..HEAD`

## Residual scope

The artifact client still materializes the compressed response before tar processing. This change bounds decompressed metrics processing but does not add a compressed-object download limit.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -729,6 +729,7 @@ docformatter --check --recursive sdk/python/ --exclude "compiler_test.py"
 - `TENSORBOARD_PROXY_SIGNING_SECRET=...`: Optional shared frontend-server secret for scoped TensorBoard proxy URLs; defaults to `MINIO_SECRET_KEY` when unset
 - `FRONTEND_SERVER_NAMESPACE=...`: Sets the namespace used by the local frontend Node server for Kubernetes lookups when it is not running inside a cluster pod. `npm run start:proxy-and-server` derives this from `NAMESPACE`.
 - `MINIO_ENDPOINT_REWRITE=from=to[,from=to]`: Rewrites explicit object-store endpoints for local proxy mode, for example from `seaweedfs.kubeflow:9000` to `localhost:9000`.
+- `MAX_METRICS_FILE_BYTES=...`: Maximum uncompressed size of the single metrics JSON file read from an `mlpipeline-metrics` archive; defaults to 1 MiB
 
 ## Troubleshooting and pitfalls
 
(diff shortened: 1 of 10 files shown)


juliusvonkohout:
Will this also help with the performance ? I am in general sometimes worried about very large artifacts performance

jeffspahr:
@juliusvonkohout Yes, for `mlpipeline-metrics` artifacts specifically. This removes the decompressed file map and full JSON string, incrementally decodes the single entry, retains only the first 50 metrics, and caps uncompressed input (1 MiB by default). That reduces peak memory and allocations for oversized or pathological metrics files.

It does not change performance for arbitrary large artifacts, and the artifact client still buffers the compressed response before this code runs. Broader large-artifact improvements would need a separate client-side streaming/download-limit change.

I also pushed `5bede57fda`, which addresses the latest review feedback and restores the prior ProtoJSON compatibility for `"metrics": null`.

jeffspahr:
Rebased onto current `master` (`5935348057`) and resolved the sole conflict by keeping the current `AGENTS.md` date while preserving the `MAX_METRICS_FILE_BYTES` documentation. `git range-diff` confirms the reviewed code patches are unchanged.

Verification completed on the rebased head:
- focused `backend/src/common/util` and artifact-client tests
- full `./test/presubmit-backend-test.sh` suite (including module tidiness and all non-integration backend Go packages)
- `gofmt`, `go vet` on touched packages, and `git diff --check`

PR #13574, opened 2026-06-23
fix(frontend): fix MetricsDropdown table cell borders

### LINKED ISSUE #13370: frontend: fix MetricsDropdown table borders for future multi-panel support
## Description

`MetricsDropdown` has two CSS issues related to table cell borders that are currently not user-facing (since `COMPARE_PANEL_COUNT = 2`) but would cause visual defects if the panel count is ever increased:

1. **`middleCell` double-border:** The `middleCell` class (introduced in #13277) applies both `borderLeft` and `borderRight`, while adjacent `leftCell` / `rightCell` classes also apply borders on their shared edge. If a panel ever lands "in the middle," dividers would render as double-thick lines.

2. **`borderCollapse` on `<td>` instead of `<table>`:** `borderCollapse: 'collapse'` is set on the `cell` class (applied to `<td>` elements), but `border-collapse` is a table-level property and has no effect on individual cells.

## Suggested Fix

- Apply `borderLeft` only on all cells except the first (remove `rightCell` and `middleCell`), or move `borderCollapse: 'collapse'` to the `<table>` element and simplify the per-cell border rules.
- Either approach produces a single consistent divider between panels regardless of count.

## Context

Identified during review of #13277. Not user-facing today since `COMPARE_PANEL_COUNT = 2` means no panel ever receives the `middleCell` class.

Priyanshu-u07:

Fix two table-border issues in `MetricsDropdown`:

- Replace the `leftCell` / `rightCell` / `middleCell` classes with a single divider class applied to every panel except the first, ensuring a single divider between panels regardless of panel count.
- Move `borderCollapse: 'collapse'` from the `<td>` element (where it has no effect) to the `<table>` element.

Notes:
-  User-visible effect: the divider between the two compare panels now renders as a single line instead of a doubled border.
- No changes to selection behavior or component logic.

Closes #13370

### DIFF
--- a/frontend/src/components/viewers/MetricsDropdown.tsx
+++ b/frontend/src/components/viewers/MetricsDropdown.tsx
@@ -46,18 +46,13 @@ import {
 import { CircularProgress } from '@mui/material';
 
 const css = stylesheet({
-  leftCell: {
-    borderRight: `3px solid ${color.divider}`,
+  table: {
+    borderCollapse: 'collapse',
   },
-  rightCell: {
+  divider: {
     borderLeft: `3px solid ${color.divider}`,
-  },
-  middleCell: {
-    borderLeft: `3px solid ${color.divider}`,
-    borderRight: `3px solid ${color.divider}`,
   },
   cell: {
-    borderCollapse: 'collapse',
     padding: '1rem',
     verticalAlign: 'top',
   },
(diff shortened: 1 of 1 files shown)


manaswinidas:
Thanks for the cleanup — the CSS fix is correct (`borderCollapse` belongs on `<table>`, not `<td>`) and collapsing three classes into one `divider` is a nice DRY improvement.

One request: the PR description mentions only the border/CSS changes, but the diff also converts the hardcoded two-panel JSX into a dynamic `selectedArtifacts.map()` loop. That's a behavioral shift (now renders N panels instead of exactly 2). Could you update the description to call that out? It helps reviewers and future `git blame` readers understand the full scope.

Optional but recommended: a quick test rendering with a single `selectedArtifact` entry would cover the new edge case the dynamic loop introduces.

LGTM otherwise — no performance, DRY, or breaking-change concerns.

manaswinidas (approved):
/lgtm
/approve

PR #13584, opened 2026-06-24
test(frontend): harden node log viewer e2e wait

mprahl:
## Summary
- wait for the helloworld logs pane to finish loading before asserting log output
- retry transient pod-log fetch failures through the logs banner refresh path
- capture better diagnostics when the log viewer still does not settle

## Test plan
- [x] `node --check test/frontend-integration-test/test-helpers.js`
- [x] `node --check test/frontend-integration-test/helloworld.spec.js`
- [ ] Full frontend integration tests not run locally

### DIFF
--- a/test/frontend-integration-test/test-helpers.js
+++ b/test/frontend-integration-test/test-helpers.js
@@ -221,6 +221,80 @@ async function waitForTableRows(
   return await $$(selector);
 }
 
+async function waitForLogViewerText(
+  expectedText,
+  { timeout = defaultTimeout, interval = 1000, screenshotName = 'log-viewer' } = {},
+) {
+  let lastLogsState = 'not-started';
+  let lastViewerText = '';
+  let lastRefreshTimestamp = 0;
+  let isRefreshingLogs = false;
+
+  try {
+    await waitForCondition(
+      async () => {
+        if (await isSelectorDisplayed('#logViewer')) {
+          const viewerText = await $('#logViewer').getText();
+          lastViewerText = viewerText;
+          if (!expectedText || viewerText.includes(expectedText)) {
+            lastLogsState = 'viewer-ready';
+            return true;
+          }
+          lastLogsState = viewerText.trim() ? 'viewer-missing-expected-text' : 'viewer-empty';
+        }
+
+        const logsPanelBusy = await isSelectorDisplayed('[role="progressbar"]');
+        if (isRefreshingLogs) {
+          if (logsPanelBusy) {
+            lastLogsState = 'waiting-for-log-refresh-to-settle';
+            return false;
+          }
+          isRefreshingLogs = false;
+        }
+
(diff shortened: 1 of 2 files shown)


test/frontend-integration-test/test-helpers.js
```diff
@@ -221,6 +221,65 @@ async function waitForTableRows(
...
+          if (
+            (await refreshButton.isExisting()) &&
+            (await refreshButton.isDisplayed()) &&
+            (await refreshButton.isEnabled())
+          ) {
+            lastLogsState = 'refreshing-log-fetch';
+            lastRefreshTimestamp = Date.now();
+            await refreshButton.click();
+            return false;
+          }
```
jeffspahr, line 272: This can click `Refresh` once per wait interval even if the previous `_loadSelectedNodeLogs()` call is still in flight. The side panel keeps children mounted while showing its busy spinner, so the banner button can remain visible/enabled and a later failed response can overwrite an earlier successful retry. Gating the refresh click until the current attempt settles should make the helper less flaky.

```suggestion
  let lastRefreshTimestamp = 0;
  let isRefreshingLogs = false;

  try {
    await waitForCondition(
      async () => {
        if (await isSelectorDisplayed('#logViewer')) {
          const viewerText = await $('#logViewer').getText();
          lastViewerText = viewerText;
          if (!expectedText || viewerText.includes(expectedText)) {
            lastLogsState = 'viewer-ready';
            return true;
          }
          lastLogsState = viewerText.trim() ? 'viewer-missing-expected-text' : 'viewer-empty';
        }

        const logsFetchFailed = await pageContainsText('Failed to retrieve pod logs.');
        if (
          logsFetchFailed &&
          !isRefreshingLogs &&
          Date.now() - lastRefreshTimestamp >= interval
        ) {
          const refreshButton = await $('button=Refresh');
          if (
            (await refreshButton.isExisting()) &&
            (await refreshButton.isDisplayed()) &&
            (await refreshButton.isEnabled())
          ) {
            lastLogsState = 'refreshing-log-fetch';
            lastRefreshTimestamp = Date.now();
            isRefreshingLogs = true;
            try {
              await refreshButton.click();
              await waitForCondition(
                async () =>
                  (await isSelectorDisplayed('#logViewer')) ||
                  !(await pageContainsText('Failed to retrieve pod logs.')),
                {
                  timeout: Math.min(timeout, 10000),
                  interval,
                  timeoutMsg: 'expected log refresh to settle',
                },
              );
            } catch {
              // Let the outer wait keep polling and preserve its timeout diagnostics.
            } finally {
              isRefreshingLogs = false;
            }
            return false;
          }
        }
```

jeffspahr (commented):
I left one inline suggestion for the retry loop in `waitForLogViewerText`. The main concern is avoiding overlapping log refresh attempts while a previous refresh is still settling.

mprahl:
@jeffspahr , I made another update to fix the flakes.

PR #13585, opened 2026-06-24
fix(test): reduce MLflow E2E terminal-state flakiness

mprahl:
## Summary
- wait for MLflow E2E assertions to poll for terminal run states
  instead of checking immediately after KFP run completion
- retry transient RunService get/retry errors that happen while
  the API server and Argo reconcile failed-run updates

## Test plan
- [x] `go test ./backend/test/testutil ./backend/test/end2end/utils \
  ./backend/test/end2end -run TestDoesNotExist`

### DIFF
--- a/backend/test/end2end/mlflow_e2e_test.go
+++ b/backend/test/end2end/mlflow_e2e_test.go
@@ -153,8 +153,8 @@ var _ = Describe("MLflow Integration >", Label(MLflow, FullRegression), func() {
 			Expect(err).NotTo(HaveOccurred())
 			Expect(mlflowExp.ID).To(Equal(mlflowExperimentID))
 
-			// Verify parent run status
-			err = e2e_utils.VerifyMLflowRunStatus(mlflowEndpoint, rootRunID, mlflowExperimentID, "FINISHED")
+			// MLflow terminal sync happens after the KFP run reaches a terminal state.
+			err = e2e_utils.WaitForMLflowRunStatus(mlflowEndpoint, rootRunID, mlflowExperimentID, "FINISHED", &timeout)
 			Expect(err).NotTo(HaveOccurred())
 
 			// Verify KFP tags on parent run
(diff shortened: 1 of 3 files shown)


backend/test/end2end/mlflow_e2e_test.go
```diff
@@ -454,19 +456,13 @@ var _ = Describe("MLflow Integration >", Label(MLflow, FullRegression), func() {
...
 			for _, run := range allRuns {
-				if run.Info.RunID != rootRunID {
-					// Nested runs for failed tasks should be FAILED
-					Expect(run.Info.Status).To(Equal("FAILED"),
-						fmt.Sprintf("Nested MLflow run %s should be FAILED", run.Info.RunID))
-				}
+				err = e2e_utils.WaitForMLflowRunStatus(mlflowEndpoint, run.Info.RunID, mlflowExperimentID, "FAILED", &timeout)
+				Expect(err).NotTo(HaveOccurred(),
+					fmt.Sprintf("MLflow run %s should be FAILED", run.Info.RunID))
 			}
```
jeffspahr: The old test explicitly verified the parent/root MLflow run reached `FAILED`. With this version, `QueryMLflowRuns` is snapshotted once and the loop only checks whatever was returned, so an empty result or a transient result missing `rootRunID` would pass without checking the parent at all. Please keep the direct parent wait before the aggregate loop.

```suggestion
			// MLflow terminal sync happens after the KFP run reaches a terminal state.
			err = e2e_utils.WaitForMLflowRunStatus(mlflowEndpoint, rootRunID, mlflowExperimentID, "FAILED", &timeout)
			Expect(err).NotTo(HaveOccurred())

			allRuns, err := e2e_utils.QueryMLflowRuns(mlflowEndpoint, mlflowExperimentID)
			Expect(err).NotTo(HaveOccurred())
			for _, run := range allRuns {
				err = e2e_utils.WaitForMLflowRunStatus(mlflowEndpoint, run.Info.RunID, mlflowExperimentID, "FAILED", &timeout)
				Expect(err).NotTo(HaveOccurred(),
					fmt.Sprintf("MLflow run %s should be FAILED", run.Info.RunID))
			}
```

jeffspahr (commented):
I found one test assertion regression in the failed-pipeline MLflow case. The PR otherwise looks coherent, and the lightweight Go verification passed locally.

PR #13590, opened 2026-06-25
fix(backend,frontend): wire code_source_url end-to-end through upload endpoints

demarna1:
## Summary

- **Backend (converter):** `toApiPipelineVersion` now sets `code_source_url` and `package_url` independently, so existing DB values are returned in API responses even when only one is populated.
- **Backend (upload handlers):** `uploadPipeline` and `uploadPipelineVersion` read the new `code_source_url` query parameter and persist it on the `PipelineVersion` model.
- **Frontend (Apis):** `uploadPipelineV2` and `uploadPipelineVersionV2` accept an optional `codeSourceUrl` and append it to the upload query string.
- **Frontend (NewPipelineVersion):** The `codeSourceUrl` form field value is now passed through all three creation paths (new pipeline from file, new version from file, new version from URL).

## Test plan

- [x] Backend unit tests pass (`go test ./backend/src/apiserver/server/...`)
- [x] New tests: `TestUploadPipeline_CodeSourceUrl`, `TestUploadPipelineVersion_CodeSourceUrl`
- [x] New frontend test: `uploadPipelineVersionV2 with codeSourceUrl` (Apis.test.ts)
- [x] New frontend test: `sends code_source_url when creating a version from URL` (NewPipelineVersion.test.tsx)
- [x] Manual: create pipeline with Code Source URL filled → verify DB `CodeSourceUrl` column is populated

## Test Evidence

Before:

Screenshot 2026-07-12 at 10 29 55 PM

After:

Screenshot 2026-07-12 at 11 26 37 PM

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -7,7 +7,7 @@
 
 ### Document metadata
 
-- Last updated: 2026-07-19
+- Last updated: 2026-07-20
 - Scope: KFP master branch (v2 engine), backend (Go), SDK (Python), frontend (React 19)
 
 ### Maintenance (agents and contributors)
@@ -328,6 +328,12 @@ The following files are generated; edit their sources and regenerate:
 - Frontend OpenAPI clients under `frontend/src/apis`, `frontend/src/apisv2beta1`, `frontend/server/src/generated/apis`, and `frontend/server/src/generated/apisv2beta1`, with shared runtime/model support under `frontend/src/generated/openapi` and `frontend/server/src/generated/openapi`
   - Sources: Swagger specs under `backend/api/**/swagger/*.json`
   - Generate: `cd frontend && npm run apis` / `npm run apis:v2beta1` / `npm run apis:all` (uses pinned Docker image `openapitools/openapi-generator-cli:v7.19.0`)
+- Backend Go HTTP client under `backend/api/v2beta1/go_http_client`
+  - Sources: `backend/api/v2beta1/swagger/*.swagger.json` (compiled from `backend/api/v2beta1/*.proto`; `pipeline.upload.swagger.json` is manually maintained)
+  - Generate: `cd backend/api && make generate`
+- Backend Python HTTP client under `backend/api/v2beta1/python_http_client`
+  - Sources: `backend/api/v2beta1/swagger/kfp_api_single_file.swagger.json`
+  - Generate: `cd backend/api && make generate-kfp-server-api-package`
 - Frontend MLMD proto outputs under `frontend/src/third_party/mlmd/generated`
   - Sources: `third_party/ml-metadata/*.proto`
   - Generate: `cd frontend && npm run build:protos`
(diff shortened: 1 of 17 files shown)


jeffspahr (approved):
I reviewed the latest head across the database and Kubernetes-store upload paths, the v2 API conversion, generated Go and Python clients, and all affected frontend creation flows. The existing verification is substantial: focused backend converter and upload-handler regression tests pass; frontend API and component tests cover new-pipeline upload, existing-version upload, and URL-based creation; generated-client validation and pre-commit pass; and the full 192-check CI matrix is green.

Blocking coverage gap: backend/src/common/client/api_server/v2/pipeline_upload_client_kubernetes.go now propagates CodeSourceURL in both Upload and UploadPipelineVersion, but this package has no tests. The database handlers and frontend flows are covered, yet neither new Kubernetes-store branch is directly exercised. Please add focused tests asserting that both Kubernetes upload paths persist CodeSourceURL before merge.

jeffspahr (changes requested):
Requesting changes for the blocking Kubernetes upload-client coverage gap detailed in my review: add focused tests proving that both Upload and UploadPipelineVersion persist CodeSourceURL before merge.

jeffspahr (changes requested):
The requested Kubernetes upload-client coverage is now solid. TestUpload_CodeSourceURL verifies the initial PipelineVersion CRD stores the supplied URL, and TestUploadPipelineVersion_CodeSourceURL verifies both the created CRD and returned model; nil behavior is covered for both paths. The full api_server_v2 package test suite passes.

One blocking repository-policy item remains: this PR changes generated backend Go and Python HTTP clients, but AGENTS.md still does not list those generated outputs. AGENTS.md explicitly requires generated-file changes to update the generated-files section with their sources and regeneration commands and to bump the last-updated date. Please document the v2beta1 go_http_client and python_http_client outputs, using pipeline.upload.swagger.json / the combined Swagger as the sources and the backend/api/README.md generation commands.

jeffspahr (changes requested):
The generated-files guide update has one blocking command error. The new Backend Python HTTP client entry says `cd backend/api && ./build_kfp_server_api_python_package.sh`, but that script reads `$API_VERSION` directly and provides no default. In a normal shell with API_VERSION unset, it resolves the output and Swagger input under `backend/api//...` instead of `backend/api/v2beta1/...`, so the documented command cannot regenerate the listed client.

Please either document `cd backend/api && API_VERSION=v2beta1 ./build_kfp_server_api_python_package.sh`, or use `cd backend/api && make generate-kfp-server-api-package`; the Make target defaults API_VERSION to v2beta1 and passes it into the container. The Go generation command, source descriptions, updated date, and the requested Kubernetes upload-client tests otherwise check out.

jeffspahr (approved):
Re-reviewed the latest head.

The previous generated-files documentation blocker is fixed: the backend Python HTTP client entry now points at `cd backend/api && make generate-kfp-server-api-package`, which supplies the default `API_VERSION=v2beta1` instead of relying on an unset shell variable.

The requested Kubernetes upload-client coverage is also in place and should stay part of the merge bar. The tests cover `CodeSourceURL` set and nil for both initial pipeline upload and pipeline-version upload, and they assert the value on the persisted PipelineVersion CRD, not just on a local model.

I did not find any remaining code-review blockers in the diff. CI still needs to be green before merge.

PR #13595, opened 2026-06-25
fix(backend): narrow MLflow namespace overrides

### LINKED ISSUE #13582: [backend] Namespace MLflow endpoint override can reuse API server credentials
### Environment

* How did you deploy Kubeflow Pipelines (KFP)? Multi-user deployment with MLflow integration enabled
* KFP version: master / unreleased
* KFP SDK version: N/A

### Steps to reproduce

1. Configure the API server with MLflow enabled and `authType: kubernetes`.
2. In a user namespace, create or update the `kfp-launcher` ConfigMap with a `plugins.mlflow` block that overrides the MLflow `endpoint`.
3. Start a pipeline run with MLflow enabled.
4. Observe that the API server resolves the namespace-provided endpoint and issues MLflow requests using the API server's Kubernetes credentials.

### Expected result

A namespace-provided `kfp-launcher` MLflow configuration should not be able to redirect API-server MLflow requests to a different endpoint while still reusing API-server credentials.

### Materials and Reference

The bug is that the per-namespace `kfp-launcher` ConfigMap allows overriding the MLflow endpoint, and the API server's credentials are then used for requests to that resolved endpoint. This creates a trust-boundary problem because a namespace-scoped configuration can steer control-plane traffic while the API server still supplies its own credentials.

This is especially concerning in multi-user deployments where the namespace-level config is intended to customize MLflow behavior without gaining control over how API-server credentials are used.

mprahl:

Fixes #13582 by keeping the MLflow endpoint server-owned and limiting namespace kfp-launcher overrides to non-routing settings. This prevents a namespace ConfigMap from steering API-server MLflow requests while still allowing explicit multi-user override layering.

Use the currently resolved MLflow endpoint for terminal sync and retries instead of reusing the value stored in plugin output. This is less surprising because the configured endpoint remains the source of truth, even if DNS names or service endpoints change while the logical MLflow server stays the same.

**Checklist:**

### DIFF
--- a/backend/src/apiserver/plugins/mlflow/config.go
+++ b/backend/src/apiserver/plugins/mlflow/config.go
@@ -21,6 +21,7 @@ import (
 	"fmt"
 	"io"
 
+	"github.com/kubeflow/pipelines/backend/src/apiserver/common"
 	apiserverPlugins "github.com/kubeflow/pipelines/backend/src/apiserver/plugins"
 	commonmlflow "github.com/kubeflow/pipelines/backend/src/common/plugins/mlflow"
 	"github.com/kubeflow/pipelines/backend/src/common/util"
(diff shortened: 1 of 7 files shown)


VaniHaripriya:
The changes look good to me. Could you add a test to verify that terminal sync and retries use the freshly resolved config rather than reusing an endpoint stored in plugins_output.

mprahl:
> The changes look good to me. Could you add a test to verify that terminal sync and retries use the freshly resolved config rather than reusing an endpoint stored in plugins_output.

Good point, updated!

PR #13577, opened 2026-06-24
fix(backend): stream bounded object store reads

jeffspahr:
## Summary
- stream archived log reads from object storage through the log archive parser
- bound stored pipeline spec reads with the existing common.MaxFileLength contract
- remove the unbounded GetFile API and add an explicit ReadFileLimited helper
- add regression coverage to prevent re-exposing BlobObjectStore.GetFile

## Why
This closes the remaining ADA-KUBEFL-04 object-store buffering gap after artifact downloads were moved to streaming in #12394. Artifact downloads already use GetFileReader; this change moves the remaining production object-store reads to streaming or explicit bounded reads.

## Verification
- GOCACHE=/tmp/kfp-go-cache go test -count=1 ./backend/src/apiserver/storage ./backend/src/apiserver/resource ./backend/src/apiserver/archive
- GOCACHE=/tmp/kfp-go-cache go test -run '^$' ./backend/src/apiserver/...
- rg -n "\.GetFile\(|GetFile\(" backend/src/apiserver backend/src/common backend/test
- git diff --check

PR #13598, opened 2026-06-25
test(backend): avoid retrying permanent archived log errors

jeffspahr:
## Summary
- Treat permanent S3/MinIO auth failures during archived workflow log diagnostics as non-retryable.
- Keep polling for successful, empty, and transient list responses.
- Add focused tests for permanent-error classification and poll behavior.

## Root cause
The archived log collector retried every ListObjects error until the 90s timeout. In the failing multi-user artifact-proxy E2E lane, MinIO returned AccessDenied for private-artifacts paths; retrying a permanent auth error delayed diagnostics and compounded flake time.

## Verification
- go test -count=1 ./backend/test/testutil

Related context: observed while triaging kubeflow/pipelines#13577 CI.

### DIFF
--- a/backend/test/testutil/archived_workflow_logs.go
+++ b/backend/test/testutil/archived_workflow_logs.go
@@ -2,8 +2,10 @@ package testutil
 
 import (
 	"context"
+	"errors"
 	"fmt"
 	"io"
+	"net/http"
 	"path"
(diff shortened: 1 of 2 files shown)


mprahl (approved):
/approve
/lgtm

PR #13600, opened 2026-06-26
fix(test): wire matrix axes to deploy and test actions. Fixes #13594

### LINKED ISSUE #13594: [CI] api-test-k8s-native: 2 matrix axes are never wired through to deploy/test actions
In `.github/workflows/api-server-tests.yml`, the `api-test-k8s-native` job defines two matrix axes that are never passed to the deploy or test-and-report composite actions:

1. **`uploadPipelinesWithKubernetesClient`** — appears in the matrix, job name, and `report_name`, but is never passed as an input to either `.github/actions/deploy` or `.github/actions/test-and-report`. In fact, neither action even defines this input. The axis doubles the number of CI jobs without changing what is actually tested.

2. **`pod_to_pod_tls_enabled`** — the deploy action accepts this input, but the k8s-native job never passes it. The TLS `include` row also flips `pipeline_store` back to `"database"`, effectively running a standalone-like configuration inside the k8s-native job. The comment in the workflow already notes that "Pod to Pod TLS manifests are not yet implemented for Kubernetes Native API mode."

For comparison, the `api-test-standalone` job correctly wires both `pod_to_pod_tls_enabled` (to deploy and test-and-report) and `db_type` (to deploy) through to the actions.

### Impact

- CI reports coverage of `uploadPipelinesWithKubernetesClient=true/false` combinations that are never actually exercised, giving false confidence.
- The TLS include row runs a configuration that doesn't match the k8s-native job's stated purpose.
- Unnecessary CI cost from duplicate jobs that test the same configuration.

### Suggested fix

- Remove the `uploadPipelinesWithKubernetesClient` axis from the matrix, job name, and report name (or wire it end-to-end if the deploy/test actions gain support for it).
- Remove the TLS `include` row (or wire `pod_to_pod_tls_enabled` through once TLS manifests are implemented for k8s-native mode).

First raised [here](https://github.com/kubeflow/pipelines/pull/12379#discussion_r3398001449).

BROGOTSOME:
**Description of changes:**
fixes: #13594

Removed unsupported TLS configuration: Deleted the TLS matrix include from the api-test-k8s-native workflow, as Pod-to-Pod TLS manifests are not yet implemented for Kubernetes Native API mode.

Fixed matrix wiring: Passed the uploadPipelinesWithKubernetesClient matrix axis into the test-and-report action so the feature is actually tested rather than defaulting to false.

Corrected test integrity: Eliminated a contradictory workaround that forced pipeline_store: "database", ensuring the job strictly tests the Kubernetes-native environment.

Optimized CI execution: Reduced the total number of jobs per run from 9 to 8 while simultaneously increasing valid test coverage.


jeffspahr:
Superseded by #13607, which already removed the unsupported k8s-native TLS/database matrix include and wired upload_pipelines_with_kubernetes_client through to test-and-report. The current head also has duplicate upload_pipelines_with_kubernetes_client entries after merging master, so this no longer needs review.

/close

Issue #4484, opened 2020-09-11
Persistence Agent keep trying to delete workflows that do not exist


Bobgy:
In #4482, a different problem is that

There are many error messages like
> Transient failure while syncing resource (default/train-pipeline-ltd84): CustomError (code: 0): Syncing Workflow (train-pipeline-ltd84): transient failure: CustomError (code: 0): Error while reporting workflow resource (code: Internal, message: Report workflow failed.: InternalServerError: Failed to delete the completed workflow for run 3e649710-8b97-4c64-9f39-a0536d8abf1b: workflows.argoproj.io \"train-pipeline-ltd84\" not found): 

I suspect the workqueue is filled with workflows that have already been deleted, so it cannot move forward with others.

## The Fix

KFP API server should mark delete workflow not found errors as permanent, so persistence agent won't need to keep retrying them.

PR #13627, opened 2026-07-02
fix(backend): avoid stale final-state labels on retry

jeffspahr:
## Summary

Follow-up to #13555 for the MLflow E2E RetryRun flake where a retried intentionally failing run could remain RUNNING until timeout.

ReportWorkflowResource writes terminal run state before plugin terminal sync and before adding pipeline/persistedFinalState. If RetryRun happens before or during terminal reporting, a stale terminal report can either terminalize plugin state again or fall through to task/metric reporting for an already-retried workflow.

This updates terminal reporting to:

- Check the reported workflow resourceVersion against the live workflow before terminal DB/plugin handling when a resourceVersion is available.
- Return a retryable Unavailable signal when terminal handling is stale or deferred, so reportWorkflow and the persistence worker skip downstream task/metric reporting and retry with the current workflow state.
- Keep the resourceVersion-guarded persistedFinalState label patch so a workflow updated by RetryRun is not labeled as persisted.
- Fix FakeWorkflowClient.Update so unit tests assert live workflow updates.

## Tests

- go test -count=1 ./backend/src/apiserver/resource
- go test -count=1 ./backend/src/apiserver/server
- go test -count=1 ./backend/src/agent/persistence/worker
- git diff --check

### DIFF
--- a/backend/src/apiserver/client/workflow_fake_test.go
+++ b/backend/src/apiserver/client/workflow_fake_test.go
@@ -135,8 +135,9 @@ func TestFakeWorkflowClient_Update(t *testing.T) {
 	if err != nil {
 		t.Fatalf("Update() unexpected error: %v", err)
 	}
-	workflow.SetLabels("updated-label", "true")
-	_, err = client.Update(ctx, workflow, v1.UpdateOptions{})
+	workflowWithUpdatedLabels := util.NewWorkflow(workflow.Workflow.DeepCopy())
+	workflowWithUpdatedLabels.SetLabels("updated-label", "true")
+	_, err = client.Update(ctx, workflowWithUpdatedLabels, v1.UpdateOptions{})
 	if err != nil {
 		t.Fatalf("Update() unexpected error: %v", err)
 	}
--- a/backend/src/apiserver/client/workflow_fake_test.go
+++ b/backend/src/apiserver/client/workflow_fake_test.go
@@ -135,7 +135,7 @@ func TestFakeWorkflowClient_Update(t *testing.T) {
 	if err != nil {
 		t.Fatalf("Update() unexpected error: %v", err)
 	}
-	workflowWithUpdatedLabels := util.NewWorkflow(workflow.Workflow.DeepCopy())
+	workflowWithUpdatedLabels := util.NewWorkflow(workflow.DeepCopy())
 	workflowWithUpdatedLabels.SetLabels("updated-label", "true")
 	_, err = client.Update(ctx, workflowWithUpdatedLabels, v1.UpdateOptions{})
 	if err != nil {
--- a/backend/src/apiserver/client/workflow_fake_test.go
+++ b/backend/src/apiserver/client/workflow_fake_test.go
@@ -135,6 +135,18 @@ func TestFakeWorkflowClient_Update(t *testing.T) {
 	if err != nil {
 		t.Fatalf("Update() unexpected error: %v", err)
 	}
+	workflow.SetLabels("updated-label", "true")
+	_, err = client.Update(ctx, workflow, v1.UpdateOptions{})
+	if err != nil {
+		t.Fatalf("Update() unexpected error: %v", err)
+	}
+	updated, err := client.Get(ctx, "update-me", v1.GetOptions{})
+	if err != nil {
+		t.Fatalf("Get() unexpected error after Update(): %v", err)
+	}
+	if updated.ExecutionObjectMeta().Labels["updated-label"] != "true" {
+		t.Errorf("Update() did not persist workflow changes, got labels: %v", updated.ExecutionObjectMeta().Labels)
+	}
 }
 
 func TestFakeWorkflowClient_UpdateNotFound(t *testing.T) {
--- a/backend/src/apiserver/resource/resource_manager.go
+++ b/backend/src/apiserver/resource/resource_manager.go
@@ -1538,6 +1538,19 @@ func (r *ResourceManager) ReportWorkflowResource(ctx context.Context, execSpec u
 	if execSpec.IsTerminating() {
 		state = model.RuntimeState(string(exec.ExecutionPhase(model.RunTerminatingConditionsV1))).ToV2()
 	}
+	if execStatus.IsInFinalState() {
+		workflowStillMatchesReport, err := r.workflowStillMatchesReportedVersion(ctx, execSpec)
+		if err != nil {
+			return nil, err
+		}
+		if !workflowStillMatchesReport {
+			return nil, terminalWorkflowReportDeferredError(
+				runId,
+				execSpec,
+				"workflow resource version changed before terminal report was persisted",
+			)
+		}
+	}
 	// If run already exists, simply update it
 	run, updateError := r.GetRun(runId)
 	if updateError == nil {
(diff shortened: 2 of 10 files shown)


backend/src/apiserver/resource/resource_manager.go
```diff
@@ -1719,7 +1719,15 @@ func (r *ResourceManager) ReportWorkflowResource(ctx context.Context, execSpec u
 			}
```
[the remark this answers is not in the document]
jeffspahr: Addressed in 77dbdd4: plugin sync failure now skips only the persistedFinalState label, then falls through to SetLabels and returns the updated ExecutionSpec.

backend/src/apiserver/resource/resource_manager.go
```diff
@@ -1719,7 +1719,15 @@ func (r *ResourceManager) ReportWorkflowResource(ctx context.Context, execSpec u
...
 		}
 
-		err := addWorkflowLabel(ctx, r.getWorkflowClient(execSpec.ExecutionNamespace()), execSpec.ExecutionName(), util.LabelKeyWorkflowPersistedFinalState, "true")
+		shouldPersistFinalState, err := r.runStillMatchesReportedFinalState(runId, state, execStatus.FinishedAt())
+		if err != nil {
+			return nil, err
+		}
+		if !shouldPersistFinalState {
+			return nil, nil
+		}
```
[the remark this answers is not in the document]
jeffspahr: Addressed in 77dbdd4: the stale retry guard now withholds persistedFinalState without returning nil, so BaseReportServer still receives a non-nil updated ExecutionSpec.

backend/src/apiserver/client/workflow_fake_test.go
```diff
@@ -135,6 +135,18 @@ func TestFakeWorkflowClient_Update(t *testing.T) {
...
+	if err != nil {
+		t.Fatalf("Update() unexpected error: %v", err)
+	}
+	updated, err := client.Get(ctx, "update-me", v1.GetOptions{})
+	if err != nil {
+		t.Fatalf("Get() unexpected error after Update(): %v", err)
+	}
+	if updated.ExecutionObjectMeta().Labels["updated-label"] != "true" {
+		t.Errorf("Update() did not persist workflow changes, got labels: %v", updated.ExecutionObjectMeta().Labels)
+	}
```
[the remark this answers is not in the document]
jeffspahr: Addressed in 77dbdd4: the fake client Update test now mutates a deep-copied workflow, so the assertion fails if Update is a no-op.

backend/src/apiserver/resource/resource_manager.go
```diff
@@ -1682,68 +1682,143 @@ func (r *ResourceManager) ReportWorkflowResource(ctx context.Context, execSpec u
 		// Notify plugins of terminal state. If a plugin sync fails and
 		// needs retry, defer the persistedFinalState label so the
 		// persistence agent re-reports the workflow on its next cycle.
+		shouldPersistFinalState := true
 		if run != nil && run.PluginsOutputString != nil && *run.PluginsOutputString != "" {
 			pr, prErr := apiservermlflow.ModelToPersistedRun(run, execSpec.ExecutionNamespace())
 			if prErr != nil {
 				glog.Warningf("Failed to build PersistedRun for plugin sync on run %q: %v", run.UUID, prErr)
 			} else if !r.pluginDispatcher.OnRunEnd(ctx, pr) {
```
mprahl, line 1702: **🤖 AI Review**

This race guard runs after `OnRunEnd()`, so a stale terminal report can still execute terminal plugin side effects before we discover that `RetryRun()` already moved the run back to `Running`. In the MLflow dispatcher, `OnRunRetry()` reopens the parent/nested runs, but `OnRunEnd()` immediately terminalizes them again. That means the retry-during-plugin-sync case this PR is fixing can still leave MLflow closed while the KFP run is running again.

Can we check whether the report is still current before invoking `OnRunEnd()` (or re-check immediately before the hook) so the stale terminal report becomes a true no-op for plugin state too?

backend/src/apiserver/resource/resource_manager.go
```diff
@@ -1682,68 +1682,143 @@ func (r *ResourceManager) ReportWorkflowResource(ctx context.Context, execSpec u
...
-			if util.IsNotFound(err) {
-				return nil, util.NewNotFoundError(err, "%s", message)
-			} else {
-				return nil, util.Wrapf(err, "%s", message)
+		if shouldPersistFinalState {
+			stillMatchesReportedFinalState, err := r.runStillMatchesReportedFinalState(runId, state, execStatus.FinishedAt())
+			if err != nil {
+				return nil, err
 			}
+			shouldPersistFinalState = stillMatchesReportedFinalState
```
mprahl: **🤖 AI Review**

When `shouldPersistFinalState` flips to `false` here, `ReportWorkflowResource()` still returns `execSpec, nil`, so the caller treats the stale terminal report as success. `report_server.go` will still call `reportTasksFromExecution()`, and the persistence worker will still `ReportMetrics(wf)` on success. For retried runs that is risky because `GenerateRetryExecution()` drops failed pod nodes, so a stale failed report can upsert old failed tasks that later reports never clear.

I think this branch needs an explicit no-op/retry signal that suppresses downstream task/metric reporting, plus a regression test through `reportWorkflow()` or `WorkflowSaver.Save()` rather than only `ReportWorkflowResource()`.

mprahl (commented):
**🤖 AI Review**

backend/src/apiserver/resource/resource_manager.go
jeffspahr: Addressed in `4d2e56b32`: terminal workflow reporting now checks the reported workflow `resourceVersion` against the live workflow before updating the run row or invoking `OnRunEnd()` when a version is available. If the workflow has already moved on, `ReportWorkflowResource` returns a retryable `Unavailable` result and skips terminal plugin sync.

Added `TestReportWorkflowResource_SkipsTerminalPluginSyncWhenReportedWorkflowIsStale` to cover this ordering.

backend/src/apiserver/resource/resource_manager.go
jeffspahr: Addressed in `4d2e56b32`: stale or deferred terminal handling now returns retryable `Unavailable` instead of `execSpec, nil`. That makes `reportWorkflow()` exit before `reportTasksFromExecution()`, and the persistence worker treats the report as transient and returns before `ReportMetrics(wf)`.

Updated the retry-during-plugin test to expect the retryable signal and added `TestWorkflow_Save_RetryableReportSkipsMetrics` to cover the persistence-worker caller path.

backend/src/apiserver/resource/resource_manager.go
```diff
@@ -1679,20 +1692,43 @@ func (r *ResourceManager) ReportWorkflowResource(ctx context.Context, execSpec u
...
+		// completed, return a retryable signal before callers report tasks or
+		// workflow metrics from a stale terminal report.
 		if run != nil && run.PluginsOutputString != nil && *run.PluginsOutputString != "" {
 			pr, prErr := apiservermlflow.ModelToPersistedRun(run, execSpec.ExecutionNamespace())
 			if prErr != nil {
 				glog.Warningf("Failed to build PersistedRun for plugin sync on run %q: %v", run.UUID, prErr)
 			} else if !r.pluginDispatcher.OnRunEnd(ctx, pr) {
 				glog.Warningf("Plugin sync failed for run %q; deferring persistedFinalState label so persistence agent retries", run.UUID)
-				return nil, nil
+				return nil, terminalWorkflowReportDeferredError(
```
mprahl, line 1704: **🤖 AI Review**

This makes every `OnRunEnd` failure block final-state persistence by returning `Unavailable`, but MLflow currently marks permanent config problems as `PLUGIN_FAILED` (`config unavailable` in the handler) and `Dispatcher.OnRunEnd()` still returns `false` for anything except `PLUGIN_SUCCEEDED`. In that state the run stays in terminal-report retry forever, never gets `pipeline/persistedFinalState`, and the workflow cannot be GC'd. Please distinguish transient plugin retry requests from permanent plugin failures so broken MLflow config does not strand completed runs.

backend/src/apiserver/resource/resource_manager.go
```diff
@@ -1538,6 +1538,19 @@ func (r *ResourceManager) ReportWorkflowResource(ctx context.Context, execSpec u
 	if execSpec.IsTerminating() {
 		state = model.RuntimeState(string(exec.ExecutionPhase(model.RunTerminatingConditionsV1))).ToV2()
 	}
+	if execStatus.IsInFinalState() {
+		workflowStillMatchesReport, err := r.workflowStillMatchesReportedVersion(ctx, execSpec)
```
mprahl, line 1542: **🤖 AI Review**

This preflight check now runs before the existing `GetRun()/UpdateRun()` path. If the persistence agent already read a terminal workflow but the CR is deleted before this `Get()` runs, `workflowStillMatchesReportedVersion()` returns `NotFound` and `ReportWorkflow` treats that as permanent, so the run row never gets updated to its terminal state. Please either update the run first for the `NotFound` case or treat this miss as deferred/retryable instead of returning a permanent not-found before the DB write.

mprahl (commented):
**🤖 AI Review**

backend/src/apiserver/resource/resource_manager.go
jeffspahr: Addressed in 6458d8a09. The retry/permanent split now happens at the source: `syncMLflowRuns` only requests a retry when the MLflow API calls themselves failed (network/server errors a later attempt can succeed at); config-class failures — config unavailable, missing settings, unresolvable credentials, missing parent run id — are recorded in `plugins_output` as `PLUGIN_FAILED` and reported as nothing-to-retry, so the run still gets `persistedFinalState` and the workflow can be GC'd. `RunPluginHandler.OnRunEnd` returns the retry signal explicitly, and the dispatcher still retries independently when the plugin output itself could not be persisted.

New dispatcher-level tests cover both classifications: `TestDispatcherOnRunEnd_PermanentConfigFailureDoesNotRequestRetry` (returns true, FAILED output persisted) and `TestDispatcherOnRunEnd_TransientSyncFailureRequestsRetry` (500-ing MLflow server → returns false).

backend/src/apiserver/resource/resource_manager.go
jeffspahr: Addressed in b8bb7bc90. When the preflight `Get` returns NotFound, `workflowStillMatchesReportedVersion` now proceeds with the reported terminal state instead of failing before the DB write: the run row is finalized (and plugin terminal sync runs) first, and the `persistedFinalState` label step then surfaces the NotFound signal to the caller afterwards — preserving the #4484 stop-retrying contract while never losing the terminal state.

Added `TestReportWorkflowResource_FinalizesRunWhenWorkflowDeletedBeforeTerminalReport`, which asserts the run reaches FAILED with the reported finish time and the caller still receives NotFound.

jeffspahr:
Fixed the `backend-tests` failure in e2ff1bc77: `TestReportWorkflow_WithMLflowOnRunEnd` still asserted the earlier revision's behavior where every plugin sync failure deferred the terminal report with `Unavailable`. Under the permanent-failure semantics from the last review round, an unavailable MLflow config records `PLUGIN_FAILED` and the terminal report succeeds — the test now asserts no error, keeps the `PLUGIN_FAILED`/`config unavailable` output assertions, and additionally asserts the run reaches its terminal state. Full `resource` package passes locally.

PR #13609, opened 2026-06-29
feat: add TLS MinVersion and cipher suite CLI flags to apiserver

ugiordan:
## Summary

Add `--tlsMinVersion` and `--tlsCipherSuites` CLI flags to the apiserver binary so TLS configuration can be controlled externally by the deploying operator or platform.

## Changes

- `backend/src/apiserver/main.go`: Add `--tlsMinVersion` flag (default `VersionTLS12`), `--tlsCipherSuites` flag (comma-separated Go cipher names), and `buildTLSConfig()` helper
- All 3 TLS surfaces (gRPC :8887, HTTP :8888, webhook :8443) now honor the configured MinVersion, CipherSuites, and NextProtos (ALPN `["h2", "http/1.1"]`)

## Context

Platform operators (e.g., OpenShift, Kubernetes distributions with TLS policy enforcement) need to pass cluster-wide TLS security settings to operands. This change enables that without adding platform-specific dependencies to the apiserver itself. The operator reads the cluster TLS policy and passes the values via these flags in the deployment template.

## Test plan

- [x] Build passes
- [x] Integration tests pass
- [x] Backward compatible: flags have sensible defaults (TLS 1.2, Go default ciphers)

PR #13629, opened 2026-07-02
fix(ci): prevent ci-passed label when workflows await approval

hbelmiro:
## Summary

- Add `reopened` trigger to the approve workflow so it fires when a PR is closed and reopened
- Add a retry loop (60s) to the approve workflow to wait for `pull_request` runs to be created before approving, fixing the race condition where the approve script ran before GitHub finished creating the workflow runs
- Add an approval gate step in CI Check that blocks `allcheckspassed` polling until all `action_required` workflow runs are approved (5-min timeout), preventing `ci-passed` from being added while workflows are invisible to the Check Runs API
- Bump `wechuli/allcheckspassed` to v2.4.0 (`204ff63e89eabdc8086f0c50b91f675bcf37c8f6`)

## Root cause

Workflows awaiting approval (`action_required`) do not create GitHub Check Runs. The `allcheckspassed` action polls Check Runs, so unapproved workflows are invisible to it. This caused `ci-passed` to be added while 11 workflows were still stuck awaiting approval (observed on #13609).

## Test plan

- [ ] Verify on an external-contributor PR: add `ok-to-test` → approve workflow fires with retry, CI Check gate waits for approval to clear before polling
- [ ] Close and reopen a PR → verify approve workflow fires on `reopened` event
- [ ] Verify org member PRs are unaffected (0 `action_required` runs → gate passes immediately)
- [ ] Confirm `ci-passed` is only added after all workflows actually complete

### DIFF
--- a/.github/workflows/ci-checks.yml
+++ b/.github/workflows/ci-checks.yml
@@ -16,6 +16,7 @@ jobs:
   check_ci_status:
     runs-on: ubuntu-latest
     permissions:
+      actions: read
       checks: read
     steps:
       - name: Determine whether CI polling is needed
@@ -42,10 +43,50 @@ jobs:
           echo "reason=${reason}" >> "$GITHUB_OUTPUT"
           echo "${reason}"
 
+      - name: Wait for action_required workflow runs to be approved
+        id: approval_gate
+        if: steps.eligibility.outputs.should_poll == 'true'
+        uses: actions/github-script@v8
+        with:
+          script: |
+            const headSha = context.payload.pull_request.head.sha;
+            const maxAttempts = 31;
+            const intervalMs = 10_000;
+
+            const listRequest = {
+              owner: context.repo.owner,
+              repo: context.repo.repo,
+              event: "pull_request",
+              status: "action_required",
+              head_sha: headSha,
+            };
+
+            for (let attempt = 1; attempt <= maxAttempts; attempt++) {
+              core.info(`[Attempt ${attempt}/${maxAttempts}] Checking for action_required runs for ${headSha}`);
+              const runs = await github.paginate(github.rest.actions.listWorkflowRunsForRepo, listRequest);
+
+              if (runs.length === 0) {
+                core.info("No action_required workflow runs found. All runs have been approved.");
+                return;
+              }
+
+              core.info(`Found ${runs.length} action_required run(s): ${runs.map(r => r.name).join(', ')}`);
+
+              if (attempt < maxAttempts) {
+                core.info(`Waiting ${intervalMs / 1000}s for approval...`);
+                await new Promise(r => setTimeout(r, intervalMs));
+              }
+            }
+
+            core.setFailed(
+              `${headSha} still has action_required workflow runs after ${(maxAttempts - 1) * intervalMs / 1000}s. ` +
+              `CI cannot pass while workflows are awaiting approval.`
+            );
+
       - name: Check if all CI checks passed
         id: poll
-        if: steps.eligibility.outputs.should_poll == 'true'
-        uses: wechuli/allcheckspassed@0b68b3b7d92e595bcbdea0c860d05605720cf479
+        if: steps.eligibility.outputs.should_poll == 'true' && steps.approval_gate.outcome == 'success'
+        uses: wechuli/allcheckspassed@204ff63e89eabdc8086f0c50b91f675bcf37c8f6
         with:
           delay: '5'
           retries: '10'
--- a/.github/workflows/ci-checks.yml
+++ b/.github/workflows/ci-checks.yml
@@ -48,10 +48,13 @@ jobs:
         if: steps.eligibility.outputs.should_poll == 'true'
         uses: actions/github-script@v8
         with:
+          retries: 3
           script: |
             const headSha = context.payload.pull_request.head.sha;
+            // Timing: approve window (60s) < this gate (300s) < allcheckspassed delay (5min)
             const maxAttempts = 31;
             const intervalMs = 10_000;
+            const initialDelayMs = 15_000;
 
             const listRequest = {
               owner: context.repo.owner,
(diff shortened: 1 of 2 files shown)


.github/workflows/gh-workflow-approve.yml
```diff
@@ -47,26 +48,47 @@ jobs:
         if: steps.membership-check.outputs.is_member == 'true' || contains(github.event.pull_request.labels.*.name, 'ok-to-test')
         uses: actions/github-script@v8
         with:
-          retries: 3
+          retries: 3 # github-script level: retries the entire script on transient API errors
           script: |
-            const request = {
+            // Poll for runs that may not exist yet due to race with workflow creation
+            const maxAttempts = 6;
```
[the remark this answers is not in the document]
hbelmiro: Fixed — bumped `maxAttempts` to 7 (6 sleeps × 10s = 60s).

.github/workflows/ci-checks.yml
```diff
@@ -42,10 +43,50 @@ jobs:
...
           echo "${reason}"
 
+      - name: Wait for action_required workflow runs to be approved
+        id: approval_gate
+        if: steps.eligibility.outputs.should_poll == 'true'
+        uses: actions/github-script@v8
+        with:
+          script: |
+            const headSha = context.payload.pull_request.head.sha;
+            const maxAttempts = 30;
```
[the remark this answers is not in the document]
hbelmiro: Fixed — bumped `maxAttempts` to 31 (30 sleeps × 10s = 300s).

.github/workflows/ci-checks.yml
```diff
@@ -42,10 +43,50 @@ jobs:
...
+              core.info(`Found ${runs.length} action_required run(s): ${runs.map(r => r.name).join(', ')}`);
+
+              if (attempt < maxAttempts) {
+                core.info(`Waiting ${intervalMs / 1000}s for approval...`);
+                await new Promise(r => setTimeout(r, intervalMs));
+              }
+            }
+
+            core.setFailed(
+              `${headSha} still has action_required workflow runs after ${maxAttempts * intervalMs / 1000}s. ` +
```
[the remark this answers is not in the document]
hbelmiro: Fixed — changed formula to `(maxAttempts - 1) * intervalMs / 1000`.

.github/workflows/gh-workflow-approve.yml
```diff
@@ -47,26 +48,47 @@ jobs:
...
+            };
 
-            core.info(`Getting workflow runs that need approval for commit ${request.head_sha}`)
-            const runs = await github.paginate(github.rest.actions.listWorkflowRunsForRepo, request)
+            let runs = [];
+            for (let attempt = 1; attempt <= maxAttempts; attempt++) {
+              core.info(`[Attempt ${attempt}/${maxAttempts}] Querying action_required runs for ${headSha}`);
+              runs = await github.paginate(github.rest.actions.listWorkflowRunsForRepo, listRequest);
+
+              if (runs.length > 0) {
```
nsingla, line 71: **[SUGGESTION]** Nit: the loop breaks on the first non-empty result, so if GitHub has only created a subset of workflow runs by this point, the rest won't be approved by this invocation. Not a correctness issue since the `approval_gate` in `ci-checks.yml` will wait for all `action_required` runs to clear — but worth being aware of as it may extend the gate's wait time.

.github/workflows/ci-checks.yml
```diff
@@ -42,10 +43,50 @@ jobs:
...
+              event: "pull_request",
+              status: "action_required",
+              head_sha: headSha,
+            };
+
+            for (let attempt = 1; attempt <= maxAttempts; attempt++) {
+              core.info(`[Attempt ${attempt}/${maxAttempts}] Checking for action_required runs for ${headSha}`);
+              const runs = await github.paginate(github.rest.actions.listWorkflowRunsForRepo, listRequest);
+
+              if (runs.length === 0) {
```
nsingla, line 75: **[IMPORTANT]** This returns success on the very first check if 0 `action_required` runs are found. For org-member PRs, this is correct. But for external contributor PRs, there's a race: if CI Check fires before GitHub has finished creating the `pull_request`-triggered workflow runs, the gate sees 0 runs on attempt 1 and passes immediately — the same logical gap this PR aims to fix.

In practice, `allcheckspassed` has a 5-minute initial delay before its first poll, so the approve workflow usually catches up in time. But this is a probabilistic mitigation, not a deterministic one.

Consider adding a minimum initial wait before the first check (e.g., 15s) to let GitHub finish creating runs:
```javascript
core.info("Waiting 15s for workflow runs to be registered...");
await new Promise(r => setTimeout(r, 15_000));

for (let attempt = 1; attempt <= maxAttempts; attempt++) {
```

.github/workflows/ci-checks.yml
```diff
@@ -42,10 +43,50 @@ jobs:
           echo "reason=${reason}" >> "$GITHUB_OUTPUT"
           echo "${reason}"
 
+      - name: Wait for action_required workflow runs to be approved
+        id: approval_gate
+        if: steps.eligibility.outputs.should_poll == 'true'
+        uses: actions/github-script@v8
```
nsingla, line 49: **[SUGGESTION]** Consider adding `retries: 3` here for parity with the approve workflow. Without it, a single transient API failure on the `github.paginate` call will fail the approval_gate step, skip the poll, and block CI with no automatic recovery.
```yaml
        uses: actions/github-script@v8
        with:
          retries: 3
          script: |
```

.github/workflows/ci-checks.yml
```diff
@@ -42,10 +43,50 @@ jobs:
...
+        with:
+          script: |
+            const headSha = context.payload.pull_request.head.sha;
+            const maxAttempts = 31;
```
nsingla, line 55: **[SUGGESTION]** Minor: consider documenting the timing relationship between the two workflows inline. The correctness of this system depends on: approve window (60s) < approval gate (300s) < allcheckspassed delay (5min). A one-line comment noting this ordering would help future maintainers avoid accidentally breaking it.

.github/workflows/ci-checks.yml
hbelmiro: Good catch — added a 15s initial delay before the first check.

.github/workflows/ci-checks.yml
hbelmiro: Added `retries: 3`.

.github/workflows/ci-checks.yml
hbelmiro: Added a comment documenting the timing ordering.

.github/workflows/gh-workflow-approve.yml
hbelmiro: I don't think it's worthy to add more complexity to handle that, since it's not a correctness issue.

nsingla (approved):
/lgtm

PR #13637, opened 2026-07-04
fix(backend): prevent API server crash when MLflow sync defers a run's final state

Sanchit2662:
## **Description of your changes:**

I found a bug where the API server could crash. When a run finishes and MLflow sync hasn't completed yet, the code was returning (nil, nil) instead of an error, and the gRPC handler wasn't expecting that so it crashed on the nil value.

Fixed it by returning an Unavailable error instead, so the persistence agent just retries later with backoff. Same pattern we already use for the GC grace period case, so nothing new, just consistent.

### **Checklist:**

### DIFF
--- a/backend/src/apiserver/resource/resource_manager.go
+++ b/backend/src/apiserver/resource/resource_manager.go
@@ -1687,8 +1687,15 @@ func (r *ResourceManager) ReportWorkflowResource(ctx context.Context, execSpec u
 			if prErr != nil {
 				glog.Warningf("Failed to build PersistedRun for plugin sync on run %q: %v", run.UUID, prErr)
 			} else if !r.pluginDispatcher.OnRunEnd(ctx, pr) {
+				// Return a retryable error (Unavailable) rather than success:
+				// the persistence agent treats it as transient and re-reports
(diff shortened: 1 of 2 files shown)


jeffspahr (approved):
No blocking findings.

Reviewed head dec0ed9c89f9a3d683e4565ded6fd960571f1811. I checked the ReportWorkflowResource nil-success path, UserError/gRPC wrapping, and persistence-agent retry mapping.

Verification:
- git diff --check origin/master...refs/review/pr-13637/head
- go test ./backend/src/apiserver/resource -run TestReportWorkflow_WithMLflowOnRunEnd -count=1
- go test ./backend/src/apiserver/resource -count=1

Sanchit2662:
Closing this in favor of #13627, which landed on master with a more complete fix for the same underlying nil-pointer panic in `ReportWorkflowResource` (via the shared `terminalWorkflowReportDeferredError()` helper), plus additional handling for the related stale final-state-label retry race. After rebasing this branch onto current master, the diff is empty, so there's nothing left for this PR to contribute. Thanks for the review, @jeffspahr!

PR #13680, opened 2026-07-08
feat(backend): add server-side pipeline_id filter to ListRuns API

### LINKED ISSUE #13578: [feature] Add server-side pipeline_id filter to ListRuns API
### Feature Area

/area backend

Server-side filtering of runs by `pipeline_id` in the `ListRuns` API (`/apis/v2beta1/runs`).

Currently, `ListRuns` supports server-side filtering by `experiment_id` (dedicated parameter) and `state` (via filter predicates), but not by pipeline_id.

Proposed: add `pipeline_id` as either a dedicated query parameter (like `experiment_id`) or as a supported key in the `filter` JSON predicates.

When listing runs for a specific pipeline via the SDK, there is no server-side filter available. Clients must fetch all runs page-by-page and filter client-side by checking `pipeline_version_reference.pipeline_id` on each run. This is inefficient when the server has thousands of runs but only a few belong to the target pipeline — pages may return fewer results than `page_size` (or zero results with a non-empty `next_page_token`), and callers must paginate exhaustively.

Yes — after calling `ListRuns`, iterate through all pages and manually check `run.pipeline_version_reference.pipeline_id` against the target pipeline ID. This works but requires exhaustive pagination and scales poorly with the total number of runs on the server.

---

Love this idea? Give it a 👍.

Priyanshu-u07:
## Description

`ListRuns` supports server-side filtering by `experiment_id` and `state` but not `pipeline_id`, so clients must paginate all runs and filter client-side. The `PipelineId` column already exists on `run_details` via the embedded `PipelineSpec`; it just wasn't exposed in `runAPIToModelFieldMap`.
## Change

**Expose the filter key** — `"pipeline_id": "PipelineId"` in `runAPIToModelFieldMap` plus the matching `GetFieldValue` case. The existing predicate framework does the rest; no proto or API schema changes. This also
makes `sort_by=pipeline_id` valid, since the map backs both filtering and sorting.

**Backfill legacy rows.** `scanRowsToRuns` reconstructs these columns from resource references *after* the query returns, so a filter or sort reading the column directly would disagree with an unfiltered read. `backfillPipelineRefsToRunTable` fills them in three statements: `PipelineId` from `'pipeline'` refs, `PipelineVersionId` from `'PipelineVersion'` refs, and then — for runs whose only pipeline link is their version — the parent resolved through `pipeline_versions.PipelineId`. Direct references take precedence.

It runs on both migration paths, not just `runLegacyUpgradeFlow`: a deployment that already took the legacy upgrade is no longer detected as legacy but can still hold reference-only rows. The work is recorded in a new `migration_statuses` table so it runs once per database rather than on every replica startup, and the claim shares a transaction with the UPDATEs so replicas starting together during a rolling upgrade do not all run the scan.

## Testing
Filter, sort-with-pagination, and backfill coverage (reference-only / no-reference / stale-reference rows), plus per-dialect quoting. The backfill was executed against real MySQL 8 and PostgreSQL 16, not just SQLite.
Also covers version-only and column-only rows, the migration gate, and claim contention.

Part of #13578


backend/src/apiserver/model/run.go
```diff
@@ -400,6 +400,7 @@ var runAPIToModelFieldMap = map[string]string{
 	"state_history":    "StateHistory",            // v2beta1 API
 	"runtime_details":  "PipelineRuntimeManifest", // v2beta1 API
 	"recurring_run_id": "JobUUID",                 // v2beta1 API
+	"pipeline_id":      "PipelineId",              // v2beta1 API
```
jeffspahr, line 403: [P1] Backfill legacy pipeline IDs before exposing this filter/sort key

`ListRuns` applies this mapping directly to `run_details.PipelineId`, but `scanRowsToRuns` intentionally reconstructs an empty stored `PipelineId` from legacy pipeline resource references after the SQL query. That makes the filter contract inconsistent: an unfiltered read can return a run with `PipelineId == X`, while `pipeline_id EQUALS X` omits it (and `NOT_EQUALS X` can return it). This mapping also enables `sort_by=pipeline_id`; SQL then orders on the empty stored value while the page token records the reconstructed value, which can skip or duplicate rows.

This is solvable in code. Please backfill `run_details.PipelineId` for legacy pipeline and pipeline-version references (or use one legacy-aware SQL expression consistently for filtering, sorting, and token values), and add regression coverage for a reference-only row.

Priyanshu-u07: Thanks! Fixed by backfilling.

`backfillPipelineRefsToRunTable` copies the values from resource references into the columns: `PipelineId` from `'pipeline'` refs and `PipelineVersionId` from `'PipelineVersion'` refs, mirroring both branches in `scanRowsToRuns`.
Filtering, sorting and page tokens then read the same value.

It runs on both migration paths, not just `runLegacyUpgradeFlow`, since a deployment that already took the legacy upgrade can still hold reference-only rows. It's idempotent, happy to add an existence guard if you'd rather avoid
the scan on every startup.

Added a reference-only row test, plus coverage for `sort_by=pipeline_id` across a page boundary. Verified on MySQL 8 and PostgreSQL 16.

backend/src/apiserver/client_manager/client_manager.go
```diff
@@ -1395,6 +1400,63 @@ func initPipelineVersionsFromPipelines(db *gorm.DB) {
...
+	if err != nil {
+		return err
+	}
+	// One statement per column, mirroring the two reconstruction branches in
+	// scanRowsToRuns.
+	for _, c := range []struct {
+		column  string
+		refType model.ResourceType
+	}{
+		{"PipelineId", model.PipelineResourceType},
```
jeffspahr: [P1] Derive `PipelineId` for version-only legacy references

This only fills `PipelineId` from a direct run→`pipeline` reference. Older supported KFP creation paths also persisted runs with `PipelineId=''` and only a run→`PipelineVersion` reference; the existing `validReferencesOfExperimentAndPipelineVersion` fixture still represents that shape. For those rows this loop fills `PipelineVersionId` but leaves `PipelineId` empty, even though the parent is available from `pipeline_versions.PipelineId`. The new filter therefore makes `pipeline_id EQUALS parent` omit the run and `NOT_EQUALS parent` include it, and sorting places it under the empty key.

Please derive the parent ID through the version relation when the direct pipeline reference is absent, and add a regression using a version-only reference—the current helper always inserts both references and masks this case.

backend/src/apiserver/client_manager/client_manager.go
```diff
@@ -507,6 +507,11 @@ func InitDBClient(initConnectionTimeout time.Duration) (*storage.DB, func() bool
 		util.TerminateIfError(autoMigrate(db))
 	}
 
+	// Runs on both paths: a deployment that took the legacy upgrade before this
+	// backfill existed is no longer detected as legacy, but can still hold
+	// reference-only rows.
+	util.TerminateIfError(backfillPipelineRefsToRunTable(db, dialect))
```
jeffspahr, line 516: [P1] Do not run both backfill `UPDATE`s on every startup

`InitDBClient` executes this unconditionally in every replica, even after all legacy rows are migrated. With MySQL 8.4 and matching GORM indexes, the exact statement on 100k runs / 10k references took 1.286s; a zero-row rerun still locked referenced run rows and the scanned resource-reference range, causing concurrent run updates and pipeline-reference inserts to hit `ER_LOCK_WAIT_TIMEOUT`. PostgreSQL's no-op rerun also sequentially scanned all 100k runs.

This makes rollout/restart latency and write blocking scale with retained history, and a timeout reaches `TerminateIfError`. Please gate it with persistent one-time migration state, or at minimum a nonlocking existence guard plus coordinated/batched execution for the actual backfill.

backend/src/apiserver/model/run.go
jeffspahr, line 403: [P2] Wire the public SDK to this predicate

`KubernetesBackend.list_runs` still resolves the requested pipeline ID but sends only the `state` predicate, then filters one returned server page client-side (`sdk/python/kfp/kubeflow_client/backends/kubernetes/backend.py:379-405`). The public `PipelinesClient` docs still warn about short or empty pages with a non-empty token. Consequently the SDK use case in #13578 remains unchanged unless callers manually build raw filter JSON.

Please include `pipeline_id` in the server filter, remove the one-page prune, and update the SDK docs/tests—or keep an explicit follow-up open rather than closing #13578.

jeffspahr (commented):
Re-review of head `1698477` after the legacy backfill update. I found two blocking backend issues and one SDK integration gap; details inline.

backend/src/apiserver/client_manager/client_manager.go
Priyanshu-u07: Done, resolves the parent through `pipeline_versions.PipelineId`, with `insertVersionOnlyRun` replacing the helper that masked the case.

It reads `run_details.PipelineVersionId` rather than the reference, so it also covers runs that only ever stored the version id, not just legacy reference rows. This is wider than you asked for. I can limit it to legacy rows if you prefer.

backend/src/apiserver/client_manager/client_manager.go
Priyanshu-u07: Done, completion is recorded in a new `migration_statuses` table, so later startups skip the backfill entirely: no scan, no locks. Replicas coordinate through the same table, only the first to insert the marker runs the UPDATEs,
and since the insert shares their transaction, a crash rolls everything back.

The one-time run does hold locks across all three statements. I didn't batch, since that needs a completion flag separate from the claim so a crashed partial run isn't skipped. Happy to add both if you'd rather bound that window.

backend/src/apiserver/model/run.go
Priyanshu-u07: Agreed, switched to Part of #13578 so it stays open, and I'll do the client switch as a follow-up.

backend/src/apiserver/client_manager/client_manager.go
jeffspahr: [P2] Handle duplicate migration claims with MySQL's connection settings

The migration marker addresses repeated scans on later startups, but the concurrent-replica claim still has a MySQL-specific gap. `initDBDriver` enables `mysqlConfig.ClientFoundRows = true`, and the pinned GORM MySQL driver implements `OnConflict{DoNothing: true}` as `ON DUPLICATE KEY UPDATE Name=Name`. With that connection setting, a duplicate no-op returns `RowsAffected = 1`, so `claimMigration` returns true for the losing caller as well.

If multiple replicas pass `migrationApplied` before the first transaction commits, each consequently executes all three backfill statements in sequence, repeating the history-wide scans and prolonging write blocking during rollout.

Please distinguish duplicate claims independently of affected-row counts under this setting, and add MySQL coverage with `ClientFoundRows=true`; the existing claim test only exercises SQLite.

Verified against the pinned driver implementation and the [MySQL affected-row semantics](https://dev.mysql.com/doc/refman/8.0/en/insert-on-duplicate.html). The full client_manager, storage, and model test suites pass; this finding was not tested against a live MySQL instance.

Priyanshu-u07: Reproduced against MySQL 8 with `clientFoundRows=true`: the second caller did get `true`.

`claimMigration` now writes a UUID claim token with the insert and reads the row back: the conflict path leaves the token untouched, so only the caller whose insert created the row reads its own token. That avoids affected-row counts, and avoids a duplicate-key error, which would abort the transaction on PostgreSQL.

Added `TestClaimMigration_MySQLClientFoundRows`, which asserts the DSN carries the setting so it can't pass without reproducing the case.

jeffspahr:
Re-review of head `aa4bd7b`: the earlier backfill and MySQL claim findings are addressed, and the SDK change is explicitly deferred. One remaining finding:

**[P2] Resolve pipeline IDs for future runs from legacy recurring jobs**

`backfillPipelineRefsToRunTable` repairs only existing `run_details` rows. Legacy recurring jobs containing only a pipeline-version ID can still produce new runs with an empty `PipelineId` after this one-time migration completes. `JobStore.scanRows` reconstructs the version ID without resolving its parent, and `ReportWorkflowResource` copies `recurringJob.PipelineSpec` directly into the new run (`backend/src/apiserver/resource/resource_manager.go:2221–2245`). Consequently, `pipeline_id EQUALS parent` omits those new runs, and subsequent startups skip the completed migration.

Please backfill the source jobs as well, or resolve the parent pipeline when persisting recurring runs, with regression coverage for a version-only legacy job producing a run after migration.

Verified with a focused resource-manager regression: a job with an empty pipeline ID and a valid pipeline-version ID produced a new recurring run whose pipeline ID remained empty, despite the version having a valid parent. The full `client_manager`, `storage`, and `model` test suites pass, and changed Go files are formatted. Live MySQL/PostgreSQL checks were not rerun.

Priyanshu-u07:
Fixed on both sides. The migration now repairs `jobs` as well as `run_details`, and `resolveRecurringWorkflowReport` resolves the parent when a job carries only a version, so the run path is correct whether or not the migration has run. Repairing the job also fixes the other two consumers of `job.PipelineId`: the recurring-run API response and the scheduled-workflow CRD.

`TestReportWorkflowResource_LegacyVersionOnlyJob_ResolvesParentPipeline` covers reproduction; I checked it fails without the fix. Worth noting the fixture has to delete the pipeline resource reference as well as blank the column,
otherwise `JobStore.scanRows` reconstructs the id and the test passes for the wrong reason.

Job backfill verified on MySQL 8 and PostgreSQL 16.

jeffspahr:
Re-review of head `ccfeea06b`: no actionable findings. The previous recurring-run issue is addressed by job backfilling and parent-pipeline resolution.

Verification passed:
- Full `client_manager`, `storage`, `model`, and `resource` Go test suites.
- Formatting and whitespace checks.
- GitHub CI checks.

Live MySQL/PostgreSQL tests were not rerun. The SDK integration remains an agreed follow-up.

Issue #759, opened 2019-01-31
Add Support for Docker-in-Docker in DSL (See https://github.com/argoproj/argo/tree/master/examples#docker-in-docker-using-sidecars)


cliveseldon:
We would like to have steps in a pipeline that allow image builders. This might require docker-in-docker.
How can this be easily facilitated?

jlewi:
I think one option would be to have your pipeline step run a separate K8s job. Then you could set whatever options (e.g. DockerInDocker, Security context are needed).

paveldournov:
@cliveseldon - Please take a look at the KFP SDK, and the methods that allow building containers from Python functions. The SDK is creating a K8S job for building containers images, you can either use the SDK within a pipeline step or use the same technique for creating an image. Please let me know if that works.

vicaire:
@cliveseldon, Argo supports building containers inside containers. 

See https://github.com/argoproj/argo/tree/master/examples#docker-in-docker-using-sidecars

Would making this functionality available from the DSL work for you?

cliveseldon:
Sorry, missed this. Yes that would be very useful if there was easy access to docker-in-docker.

vicaire:
/cc @hongye-sun (Adding Hongye who may have some input on this). 

@cliveseldon, would you be interested in contributing so that it is easy to specify the following in the DSL? 

https://github.com/argoproj/argo/tree/master/examples#docker-in-docker-using-sidecars

cliveseldon:
@vicaire Happy to help. Just a bit new to the code base so not sure where to start.

hongye-sun:
@cliveseldon, the docker in docker support depends on sidecar feature, which is currently working in the PR (https://github.com/kubeflow/pipelines/pull/879). Once the PR is in, you should be able to build similar pipeline as argo example does.

gaoning777:
This #879 is merged. Feel free to add the docker-in-docker feature.

hongye-sun:
I don't think there is more work need to be done to support dind other than the sidecar support. Closing this bug.

PR #758, opened 2019-01-31
Add UI actions to Buttons module

yebrahim:
This PR refactors button action methods into the `Buttons` module, which fixes a few bugs, and allows for sharing common functionality.

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/pipelines/758)

Issue #666, opened 2019-01-10
python components don't work for images with python 2


alan-krumholz:
Same simple python code runs fine in an python 3 image but it won't run on a python 2 image:

```
import kfp
from kfp import compiler
import kfp.dsl as dsl
import kfp.notebook
import kfp.gcp as gcp
import datetime
OUTPUT_DIR = 'gs://pipelinestest/out'
PROJECT_NAME = 'energy-forecasting'
@dsl.python_component(
    name='Python test')
def python_Func() -> str:
    import tensorflow
    return("done")
P2Op = compiler.build_python_component(
    base_image = 'tensorflow/tensorflow:1.10.0',
    component_func = python_Func,
    staging_gcs_path = OUTPUT_DIR,
    target_image = 'gcr.io/' + PROJECT_NAME + '/component-p2:latest')
P3Op = compiler.build_python_component(
    base_image = 'tensorflow/tensorflow:1.10.0-py3',
    component_func = python_Func,
    staging_gcs_path = OUTPUT_DIR,
    target_image = 'gcr.io/' + PROJECT_NAME + '/component-p3:latest')
```

In the python 2 image, I get the following error:

> ImportError: No module named 'tensorflow'

In the python3 image, the component runs fine.

gaoning777:
By default, the build_python_component configures the entrypoint as python3. That is why in python2 images, the tensorflow package is not found when you run python3 program. 
To support this, we need to change the component builder to make the entrypoint configurable.

gaoning777:
fixing it in https://github.com/kubeflow/pipelines/pull/730

gaoning777:
Now the component build supports python2. Fixed

Ark-kun:
> def python_Func() -> str:

We only support Python 3.5+ officially.
Note that your function definition is only valid in Python 3.5+. It's a limitation of python, not Pipelines.

Issue #665, opened 2019-01-10
Document DSL functionality to specify execution order and execution dependencies


swiftdiaries:
Is there a way to order steps other than passing file_outputs to enforce dependencies? 
I'd like to be able to do: 
dsl.ContainerOp(step_name='first')
dsl.ContainerOp(step_name='second')
dsl.ContainerOp(step_name='third', dependencies=['first', 'second'])

Something similar to https://github.com/argoproj/argo/blob/master/examples/dag-coinflip.yaml.

swiftdiaries:
TIL
You can add dependencies by doing:
op_name_second.after(op_name_first)

https://github.com/kubeflow/pipelines/blob/master/sdk/python/kfp/dsl/_container_op.py#L104

vicaire:
Does it solve your problem? Can we close this issue?

swiftdiaries:
Yes it does solve the problem, I left it open for visibility in case other people were looking for something similar. We should probably add instructions to the website

vicaire:
Good point. Re-opening and changing the title to keep track of adding the documentation.

vicaire:
Note: instead of documentation on the website, it seems that having good samples would be a good option (assuming small samples, each documenting a particular feature).

gaoning777:
Keeping track of the DSL samples in https://github.com/kubeflow/pipelines/issues/1034

Ark-kun:
An easier way to add dependencies is to have data dependencies.
When you pass some output of task A to task B, the dependency is added, so you do not need  to write anything.

gaoning777:
We do have samples illustrating the feature: https://github.com/kubeflow/pipelines/blob/a39ae8eb779e69e4e85e424abba8ecac3f60435f/samples/basic/recursion.py#L65. However it is not specifically for this feature. Agree that it is better to have a sample for each feature in the basic folder.

SinaChavoshi:
submitted PR  #1668 for this issue pending addition of automated test for the sample.

gaoning777:
Sample tests are tracked here: https://github.com/kubeflow/pipelines/issues/1813

PR #13681, opened 2026-07-08
fix(backend): avoid logging secret references

jeffspahr:
## Summary

- Stop logging raw driver pod spec patches and Kubernetes executor configs; log byte counts instead.
- Stop returning configured S3 bucket and MLflow credential secret key names in errors that are logged by callers.
- Add focused regression tests for the sensitive log-message and error-string contracts.

## Why

CodeQL flagged `go/clear-text-logging` flows where Kubernetes secret reference fields could reach backend logs through driver, objectstore/launcher, and MLflow plugin paths.

This is expected to address code scanning alerts #759, #758, #757, #666, #665, #664, #663, #198, and #56.

## Tests

- `go test ./backend/src/v2/cmd/driver ./backend/src/v2/objectstore ./backend/src/apiserver/plugins/mlflow`
- `git diff --check`

### DIFF
--- a/backend/src/apiserver/plugins/mlflow/config.go
+++ b/backend/src/apiserver/plugins/mlflow/config.go
@@ -430,7 +430,7 @@ func resolveBearerSecretCredentials(
 	if err != nil {
 		return commonmlflow.MLflowCredentials{}, err
 	}
-	token, err := readRequiredSecretKey(secret, namespace, ref.TokenKey)
+	token, err := readRequiredSecretKey(secret, namespace, ref.TokenKey, "bearer token")
 	if err != nil {
 		return commonmlflow.MLflowCredentials{}, err
 	}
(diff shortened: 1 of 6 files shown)


jeffspahr:
Addressed in the latest commit: the error now names which credential role is missing — `bucket credential secret has no value for: access key, secret key` (or just the absent one) — while still never echoing the configured secret data key names. Combined with the existing wrap (`failed to get Bucket credentials from secret name=%q namespace=%q`), operators get "which secret, which role" with no sensitive strings.

Test updated to assert the specific message and that the configured key names (`does_not_exist_access_key`/`..._secret_key`) still never appear; `go test ./backend/src/v2/objectstore/...` passes.

zazulam (commented):
/lgtm
/approve

PR #13683, opened 2026-07-08
fix(frontend): object storage Insecure Direct Object Reference (IDOR) prefix fallback

juliusvonkohout:

Broken since https://github.com/kubeflow/pipelines/pull/13433
for KFPv1 almost completely and kfp v2 to some degree
The UI just blocks artifact downloads even though it is the same namespace

and there is another now exposed security bug where ml-pipeline-ui tries to read customer secrets and is luckily blocked by rbac 

image

@droctothorpe for review

**Checklist:**

### DIFF
--- a/.github/scripts/ai_triage.sh
+++ b/.github/scripts/ai_triage.sh
@@ -1,0 +1,69 @@
+set -e
+gh extension install github/gh-models --force || true
+
+echo "Using model: $TARGET_MODEL"
+
+read -r -d '' SYSTEM_INSTRUCTIONS << 'EOF'  || true
+You are an expert open-source maintainer for Kubeflow Pipelines.
+
+Analyze the quality of the incoming issue based on Scope, Context, Guidance, and Complexity.
+             
+Calibrate your evaluation against these compressed reference standards:
+- BACKEND (#13314): [backend] S3 operations fail with non-AWS object stores after AWS SDK v2 checksum defaults change. Clear isolated scope.
+- BUG TIER (#13180): [bug] fix: E2E test flakiness on K8s v1.34.0 — root cause analysis. High-quality root-cause analysis and environment data.
+- FRONTEND (#13108): [frontend] Adds coverage for frontend mock:api startup and enum drift. Explicit file paths and definitions of done.
+- SDK (#12865): [sdk] [bug] [set_accelerator_limit] rejects valid accelerator counts (only allows 0, 1, 2, 4, 8, 16). Elite precision with failing parameters.
+
+Compare the incoming issue detail density directly against the relevant blueprint standard above.
+
+Respond strictly following this format structure without other markdown wraps:
+
+- Each section MUST contain exactly 2 to 3 short, bullet fragments. 
+- Do NOT write full-length paragraphs or introductory text. Keep it highly concise for quick scanning.
+- Do NOT include any time frame or implementation window estimations.
+
+
+### 📊 Scope
+- <If the technical task boundaries are clear or ambiguous>
+- <If the issue isolates specific components, files, or packages correctly>
+
+
+### 📝 Context & Guidance
+<Evaluate if steps, expected behavior, or links are provided against repo standards>
+
+### ⚡ Complexity
+- <State difficulty tier: Low, Medium, or High, calibrated against this exact rubric:
+  * LOW: Task is isolated to single-file fixes, shallow tweaks, or documentation updates.
+  * MEDIUM: Task has moderate architectural depth, affecting internal logic patterns or specific layer wrappers.
+  * HIGH: Task has deep architectural depth or high breadth, spanning multiple system components simultaneously (e.g., changes across the SDK, backend, frontend, or argo compiler engines).>
+- <Break down the breadth (cross-layer impact) and depth (internal system complexity) of the proposed change>
+
+
+### 🎯 Overall Issue Quality Verdict
+- <State definitively if this is ready for immediate developer pickup>
+- <Outline the single most impactful recommendation to improve the issue quality>
+EOF
+
+
+USER_PROMPT="Title: ${TITLE} | Body: ${RAW_BODY}"
+
+if RESULT=$(gh models run "$TARGET_MODEL" --system-prompt "$SYSTEM_INSTRUCTIONS" "$USER_PROMPT" 2>&1); then
+    echo "✅ AI model executed successfully."
+    ANALYSIS_REPORT="$RESULT"
+else
+    echo "⚠️ CRITICAL: AI Model execution failed or preview tier limit hit."
+    echo "Error details: $RESULT"
+            
+    read -r -d '' ANALYSIS_REPORT << 'EOF' || true
+    ### ⚠️ Automated Triage Skipped
+    The issue body text or environment logs exceeded processing size boundaries for this triage pass.
+EOF
+fi
+
+echo "analysis<<EOF" >> $GITHUB_OUTPUT
+echo "$ANALYSIS_REPORT" >> $GITHUB_OUTPUT
+echo "EOF" >> $GITHUB_OUTPUT
+
+          
+echo "DEBUG: The raw analysis sent to output was:"
+echo "$ANALYSIS_REPORT"
--- a/frontend/server/handlers/artifacts.ts
+++ b/frontend/server/handlers/artifacts.ts
@@ -109,12 +109,14 @@ export interface GCSProviderInfo {
  *    scalability, and is prone to many CVEs in the artifact proxy
  *    deployment.
  *
- * Note: Secret-backed provider mode (fromEnv === 'false') is unsupported
- * in multi-user deployments. The ml-pipeline-ui ClusterRole no longer
- * grants secrets:get/list permissions, so getK8sSecret() calls will be
- * denied by RBAC at the cluster level. This mode may still work in
- * standalone (single-tenant) deployments where the service account has
- * direct secret access. See: https://github.com/kubeflow/pipelines/pull/12860
+ * Note: Secret-backed provider mode (fromEnv === 'false') names a Kubernetes
+ * Secret to source object-store credentials from. The frontend server only
+ * honors it when the requested namespace is the server's own namespace, so it
+ * never reads Secrets from a customer namespace. In multi-user deployments the
+ * provider info is dropped for user namespaces and artifact retrieval falls
+ * back to the server's own environment credentials (SeaweedFS in the kubeflow
+ * namespace) or the per-namespace artifact proxy.
+ * See: https://github.com/kubeflow/pipelines/pull/12860
  *
  * Security: This addresses the vulnerability where the namespace parameter
  * could be manipulated to access artifacts from other namespaces.
@@ -286,14 +288,30 @@ export function getArtifactsHandler({
     }
     console.log(`Getting storage artifact at: ${source}: ${bucket}/${key}`);
 
+    // Security: The ml-pipeline-ui service account is only permitted to read
+    // Secrets from its own (server) namespace. Secret-backed provider info
+    // (fromEnv === 'false') names a Secret to read for object-store
+    // credentials; honoring it for a customer/user namespace would read
+    // Secrets cross-namespace, which is forbidden. When the requested
+    // namespace is not the server's own namespace we drop the provider info so
+    // credential resolution falls back to the server's own environment
+    // credentials (SeaweedFS in the kubeflow namespace) or, when enabled, the
+    // per-namespace artifact proxy. See:
+    // https://github.com/kubeflow/pipelines/pull/12860
+    const allowProviderSecrets = !!namespace && namespace === options.server.serverNamespace;
+    const effectiveProviderInfo = allowProviderSecrets ? providerInfo : '';
+
     let client: MinioClient;
     switch (source) {
       case 'gcs':
-        await getGCSArtifactHandler({ bucket, key }, peek, providerInfo, namespace)(req, res);
+        await getGCSArtifactHandler({ bucket, key }, peek, effectiveProviderInfo, namespace)(
+          req,
+          res,
+        );
         break;
       case 'minio':
         try {
-          client = await createMinioClient(minio, 'minio', providerInfo, namespace);
+          client = await createMinioClient(minio, 'minio', effectiveProviderInfo, namespace);
         } catch (e) {
           res.status(500).send(`Failed to initialize Minio Client for Minio Provider: ${e}`);
           return;
@@ -310,7 +328,7 @@ export function getArtifactsHandler({
         break;
       case 's3':
         try {
-          client = await createMinioClient(aws, 's3', providerInfo, namespace);
+          client = await createMinioClient(aws, 's3', effectiveProviderInfo, namespace);
         } catch (e) {
           res.status(500).send(`Failed to initialize Minio Client for S3 Provider: ${e}`);
           return;
@@ -595,12 +613,14 @@ function sanitizeTarEntryName(name: string): string | null {
 }
 
 /**
- * Parses GCS provider info and retrieves credentials from a Kubernetes secret.
+ * Parses GCS provider info and retrieves credentials from a Kubernetes Secret.
  *
- * WARNING: This function is unsupported in multi-user deployments.
- * The ml-pipeline-ui ClusterRole no longer grants secrets:get/list
- * permissions, so getK8sSecret() calls will be denied by RBAC.
- * See: https://github.com/kubeflow/pipelines/pull/12860
+ * Security: The artifact handler only forwards provider info when the
+ * requested namespace is the frontend server's own namespace, so this function
+ * never reads Secrets from a customer namespace. In multi-user deployments the
+ * provider info is dropped for user namespaces and credentials fall back to
+ * the server's own environment credentials or the per-namespace artifact
+ * proxy. See: https://github.com/kubeflow/pipelines/pull/12860
  */
 async function parseGCSProviderInfo(
   providerInfo: GCSProviderInfo,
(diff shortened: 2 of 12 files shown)


droctothorpe (commented):
I think this needs one more guard before merge: the prefix fallback validates the raw URI string, but HTTP(S) fetches normalize dot-segments.

- Example: `https://host/bucket/private-artifacts/team-a/../team-b/secret` matches `team-a` in `namespaceFromArtifactUri()`, so `decideFromPrefixFallback()` allows it for a `team-a` request.
- The actual HTTP request is normalized by URL/fetch to `.../private-artifacts/team-b/secret`, so an artifact absent from MLMD can pass the fallback for one namespace and fetch another namespace object.
- Lazy fix: keep prefix fallback only for object-store sources (`minio`/`s3`/`gcs`) and preserve `artifact-not-found` denial for `http`/`https`. Only add canonical path normalization if HTTP(S) fallback is really needed.

droctothorpe:
One more compatibility concern: this should preserve v1 deployments where namespace isolation is done with one S3 bucket per namespace rather than a `private-artifacts/<namespace>/...` key prefix. That matters because this is not a major release, so v1 compatibility should not regress.

Rationale:

- The current fallback only derives ownership from the object key prefix.
- In bucket-per-namespace deployments, a valid v1 artifact may look like `s3://team-a-bucket/workflow/pod/main.log`, with no namespace in the key.
- If MLMD has no artifact record, this PR would deny that as `prefix-absent`, even though the bucket itself is the namespace boundary.
- Bucket fallback should be opt-in, not unconditional, because shared-bucket installs would be unsafe if `bucket === mlpipeline` alone allowed access.

Could we add a third ownership mode, for example `mlmd-then-prefix-or-bucket`, that checks the namespace's configured artifact repository bucket only when explicitly enabled?

Sketch:

```ts
export async function decideFromPrefixOrBucketFallback(
  artifactUri: string,
  claimedNamespace: string,
  requestedBucket: string,
): Promise<ValidationResult> {
  const prefixDecision = decideFromPrefixFallback(
    artifactUri,
    claimedNamespace,
    'mlmd-then-prefix',
  );
  if (prefixDecision.valid || prefixDecision.reason === 'prefix-namespace-mismatch') {
    return prefixDecision;
  }

  const namespaceBucket = await getArtifactRepositoryBucket(claimedNamespace);
  if (namespaceBucket && requestedBucket === namespaceBucket) {
    return { valid: true, reason: 'bucket-match' };
  }

  return { valid: false, reason: 'bucket-namespace-mismatch' };
}
```

The bucket lookup should come from the namespace's `artifact-repositories` ConfigMap / existing artifact-repository helper path, not from bucket naming conventions. That keeps shared-bucket installs on the current prefix-only behavior while letting bucket-per-namespace v1 installs opt in without breaking.

jeffspahr:
Opened #13687 (stacked on this branch) with integration tests that run the real MLMD validator through the artifacts auth middleware — the archived-log test fails on master and passes with this PR's prefix fallback, so it guards against this regression recurring. It should merge after this PR.

frontend/server/helpers/mlmd-validator.ts
```diff
@@ -271,6 +314,48 @@ export function decideFromContexts(
...
+  ownershipMode: string = NAMESPACE_OWNERSHIP_MODE,
+): ValidationResult {
+  if (ownershipMode === 'mlmd-only') {
+    console.warn(
+      `[SECURITY] Artifact not found in MLMD for URI "${artifactUri}", ` +
+        `denying access (no namespace ownership evidence).`,
+    );
+    return { valid: false, reason: 'artifact-not-found' };
+  }
+  const prefixNamespace = namespaceFromArtifactUri(artifactUri);
```
jeffspahr, line 355: The object key is fully caller-controlled (query param or decoded path segment) and is later passed verbatim to the object store, so a key containing dot-segments — e.g. `private-artifacts/team-a/../victim-ns/obj` — carries the claimant's prefix while potentially addressing another namespace's object. Whether that traverses depends on the backend: MinIO rejects dot-segment object names, but AWS S3 behavior varies by path style and SeaweedFS maps S3 keys onto filer paths. Since this validator is the only guard (ml-pipeline-ui uses full-bucket credentials), it shouldn't delegate normalization to the store. Before this PR the strict denial incidentally blocked this; the fallback should deny non-normalized keys explicitly:

```suggestion
  // The object key is caller-controlled and is later passed verbatim to the object
  // store, so a key containing empty or dot segments (for example
  // "private-artifacts/team-a/../victim-ns/obj") would carry the claimant's prefix
  // while addressing another namespace's object on stores that normalize paths.
  // Deny non-normalized keys outright rather than trusting every backend to reject them.
  const objectKey = artifactUri.replace(/^[a-zA-Z][a-zA-Z0-9+.-]*:\/\/[^/]+\//, '');
  const hasUnsafeSegment = objectKey
    .split('/')
    .some((segment) => segment === '' || segment === '.' || segment === '..');
  if (hasUnsafeSegment) {
    console.warn(
      `[SECURITY] Insecure direct object reference blocked: artifact "${artifactUri}" ` +
        `contains empty or dot path segments, so its object key cannot be trusted to ` +
        `stay inside its owning-namespace prefix; denying access.`,
    );
    return { valid: false, reason: 'key-not-normalized' };
  }
  const prefixNamespace = namespaceFromArtifactUri(artifactUri);
```

And a matching unit test in the new `decideFromPrefixFallback` describe block:

```ts
it('denies a key that hides a dot-segment traversal behind a matching prefix', () => {
  expect(
    decideFromPrefixFallback(
      'minio://mlpipeline/private-artifacts/team-a/../victim-ns/obj',
      'team-a',
      'mlmd-then-prefix',
    ),
  ).toEqual({ valid: false, reason: 'key-not-normalized' });
});
```

frontend/server/helpers/mlmd-validator.ts
```diff
@@ -271,6 +314,48 @@ export function decideFromContexts(
...
+// factored out of `validateArtifactNamespace` so that every branch (strict `mlmd-only`
+// denial, absent prefix denial, prefix/namespace mismatch denial, and prefix match) is
+// unit-testable without a live metadata store. The ownership mode is a parameter, again
+// defaulting to the module constant, so tests can exercise the strict mode directly.
+export function decideFromPrefixFallback(
+  artifactUri: string,
+  claimedNamespace: string,
+  ownershipMode: string = NAMESPACE_OWNERSHIP_MODE,
+): ValidationResult {
+  if (ownershipMode === 'mlmd-only') {
```
jeffspahr: Any value other than exactly `mlmd-only` — a typo like `mlmd_only` or `strict` — currently falls through to the weaker prefix fallback with no signal. An operator who intends the strict mode and mistypes silently gets the weaker guard. Inverting the check makes unrecognized values fail closed:

```suggestion
  // Only the documented opt-in value enables the prefix fallback; any other value
  // (including typos) fails closed to the strict denial, so a misconfigured mode can
  // never silently weaken the guard.
  if (ownershipMode !== 'mlmd-then-prefix') {
```

manifests/kustomize/base/pipeline/ml-pipeline-ui-deployment.yaml
```diff
@@ -70,6 +70,16 @@ spec:
...
           - name: DISABLE_GKE_METADATA
             value: "true"
+          # Namespace-ownership resolution for artifacts the metadata store does not track
+          # (for example pod logs and other objects legitimately absent from the metadata
+          # store). The default "mlmd-then-prefix" derives the owning namespace from the
+          # "private-artifacts/<namespace>/" object-key prefix, restoring v1 and v2 artifact
+          # retrieval while still blocking cross-namespace access; an object that carries no
+          # such prefix has no derivable owner and is denied, because nothing is ever written
+          # to the bucket root. Set to "mlmd-only" to enforce the strict guard that denies
+          # every artifact absent from the metadata store.
```
jeffspahr: `ARTIFACT_NAMESPACE_KEY_PREFIX` exists in the code but isn't discoverable from the manifest; a one-line mention helps operators running a customized `keyFormat`:

```suggestion
          # every artifact absent from the metadata store. Deployments with a customized
          # Argo keyFormat / pipeline root can override the expected leading key segment
          # via ARTIFACT_NAMESPACE_KEY_PREFIX (default "private-artifacts").
```

jeffspahr (commented):
AI review using Fable:

Thanks for the fix @juliusvonkohout — the approach is sound. I verified the layout assumption it rests on: both the base Argo `keyFormat` (`manifests/kustomize/third-party/argo/base/workflow-controller-configmap-patch.yaml`) and the multi-user profile controller (`sync.py` keyFormat, `defaultPipelineRoot`, and bucket policies) put everything under `private-artifacts/{namespace}/`, so deriving ownership from that prefix is legitimate. The regex anchoring to the first key segment is correct and well tested, and preserving the strict behavior behind `mlmd-only` is a good escape hatch.

I've left three inline suggestions. The important one is the dot-segment check: the prefix fallback deliberately opens a path the strict guard previously closed, and a caller-controlled key like `private-artifacts/team-a/../victim-ns/obj` carries the claimant's prefix while addressing another namespace's object on stores that normalize paths. Since ml-pipeline-ui holds full-bucket credentials (the per-namespace policies in `sync.py` only constrain in-cluster workload credentials), this validator is the only guard.

Two more things that don't map to diff lines:

1. **`http`/`https` sources**: the middleware validates sources `minio, s3, gcs, http, https`, so the fallback now also allows `http://<any-allowed-host>/private-artifacts/team-a/...`. The prefix ownership guarantee only exists for the object-store bucket; consider limiting the fallback to `minio`/`s3`/`gcs` and keeping the strict denial for `http`/`https`.
2. **Test coverage for the wiring**: none of the CI lanes exercise this middleware with auth enabled (the frontend e2e lane deploys standalone where the middleware short-circuits, and the multi-user lanes only hit the API server), which is how the original regression shipped. I opened #13687 — stacked on this branch — with integration tests that run the real validator end to end through the Express app, mocking only the gRPC-web transport. The archived-log test fails on master and passes with this PR. You're welcome to cherry-pick that commit into this PR if you'd rather land it together; otherwise it can merge right after.

Nit: PR title typo — "object sotrage" → "storage".

frontend/server/helpers/mlmd-validator.ts
jeffspahr: Amending this suggestion so it doesn't collide with @droctothorpe's bucket-per-namespace proposal: as written above, `ownershipMode !== 'mlmd-then-prefix'` would send a new `mlmd-then-prefix-or-bucket` mode into the strict-deny branch — silently reintroducing the deny-all regression for exactly the installs that mode is meant to fix. An explicit allowlist keeps the fail-closed property while leaving room for additional recognized modes:

```suggestion
  // Only recognized fallback modes enable prefix-based ownership; any other value
  // (including typos) fails closed to the strict denial, so a misconfigured mode can
  // never silently weaken the guard. If the opt-in bucket-per-namespace mode from the
  // review discussion is added, include it here.
  const prefixFallbackModes = ['mlmd-then-prefix'];
  if (!prefixFallbackModes.includes(ownershipMode)) {
```

On the bucket mode itself: the design looks sound to me — the lookup keys off the *claimed* namespace, which RBAC has already verified, and the resolution path already exists (`workflow-helper.ts` reads the per-namespace `artifact-repositories` ConfigMap). Two implementation notes:

1. **Shared-bucket footgun**: in shared-bucket installs the profile controller writes every namespace's `artifact-repositories` ConfigMap pointing at the same bucket, so if an operator enables the bucket mode there, `requestedBucket === namespaceBucket` holds for every namespace and the guard collapses. Since opt-in alone doesn't prevent that misconfiguration, consider only honoring the bucket match when the resolved repository's `keyFormat` does not scope by namespace (i.e. contains no `{{workflow.namespace}}` / namespace-prefixed segment), or at minimum documenting loudly that the mode is only safe when every namespace has a dedicated bucket.
2. **Lookup cost**: the ConfigMap read is a per-request Kubernetes API call on the artifact download path; it should be cached with a short TTL.

.github/scripts/ai_triage.sh
```diff
@@ -1,69 +0,0 @@
-set -e
```
droctothorpe, line 1: Just out of curiosity why are these two files being deleted in this PR?

juliusvonkohout: by accident, let me restore

frontend/server/handlers/artifacts.ts
```diff
@@ -286,14 +288,30 @@ export function getArtifactsHandler({
...
+    // (fromEnv === 'false') names a Secret to read for object-store
+    // credentials; honoring it for a customer/user namespace would read
+    // Secrets cross-namespace, which is forbidden. When the requested
+    // namespace is not the server's own namespace we drop the provider info so
+    // credential resolution falls back to the server's own environment
+    // credentials (SeaweedFS in the kubeflow namespace) or, when enabled, the
+    // per-namespace artifact proxy. See:
+    // https://github.com/kubeflow/pipelines/pull/12860
+    const allowProviderSecrets = !!namespace && namespace === options.server.serverNamespace;
+    const effectiveProviderInfo = allowProviderSecrets ? providerInfo : '';
```
jeffspahr: Two details on the gate:

1. **Standalone regression**: when auth is disabled, requests legitimately omit `namespace` (the auth middleware only enforces it when auth is enabled), and `!!namespace && ...` then drops `fromEnv === 'false'` provider info that previously worked in standalone single-tenant installs. Treating a missing namespace as server-local preserves that behavior, and it can't be attacker-controlled in multi-user because the middleware 400s namespace-less requests before this handler runs.
2. **Silent drop**: when provider info is stripped, retrieval falls back to the server's default store and a missing object surfaces as `NoSuchKey` with no hint why. A warn here makes that diagnosable.

```suggestion
    // A missing namespace only occurs when auth is disabled (single-tenant): the
    // auth middleware rejects namespace-less requests whenever auth is enabled, so
    // treating it as server-local cannot be triggered by a multi-user caller.
    const allowProviderSecrets = !namespace || namespace === options.server.serverNamespace;
    if (!allowProviderSecrets && providerInfo) {
      console.warn(
        `Ignoring secret-backed provider info for namespace "${namespace}": Secrets may ` +
          `only be read from the server namespace; falling back to environment credentials.`,
      );
    }
    const effectiveProviderInfo = allowProviderSecrets ? providerInfo : '';
```

jeffspahr (commented):
Re-ran the review on the new head. Summary:

**Verified from the previous round:** both hardening suggestions are applied (dot-segment/empty-segment keys denied as `key-not-normalized`; unrecognized ownership modes fail closed to the strict denial), the key-prefix override is documented in the manifest, and sourcing `ARTIFACT_NAMESPACE_OWNERSHIP_MODE` from `pipeline-install-config` (optional, code default preserved) is a nice operator surface. My earlier note about turning the `!== 'mlmd-then-prefix'` check into an allowlist still applies only if the opt-in bucket-ownership mode from the earlier discussion lands — nothing to do now.

**New Secrets fix:** the direction is right and I verified the load-bearing pieces: `FRONTEND_SERVER_NAMESPACE` is injected via the downward API in the base deployment and defaults to `kubeflow` in `configs.ts`, `getK8sSecret` resolves to the server namespace, and reaching the handler with `namespace=<server ns>` in multi-user still requires passing the SubjectAccessReview for that namespace. So cross-namespace Secret reads are now prevented at the application layer in addition to RBAC, with regression tests for both the s3 and gcs drop paths. Two inline comments on the details.

**Tests:** as requested, I've added my changes to this PR the closest way permissions allow — #13687 is now retargeted at this branch (`fix-frontend-artifact-idor-prefix-fallback`), so merging it pulls three test commits straight into this PR: the integration suite that runs the real MLMD validator through the artifacts auth middleware (including a middleware-level traversal case and the bucket-per-namespace default-mode pin), plus unit tests for `key-not-normalized` and the fail-closed mode. Full frontend server suite is green on your current head with them applied (356 passed, 2 todo).

frontend/server/workflow-helper.ts
```diff
@@ -328,7 +328,17 @@ export async function getPodLogsMinioRequestConfigfromWorkflow(
...
+  // namespace is the server's own namespace. In multi-user deployments the
+  // namespace is a customer/user namespace, and the ml-pipeline-ui service
+  // account may not read Secrets there. In that case we fall back to the
+  // server's own environment credentials (SeaweedFS in the kubeflow
+  // namespace). See: https://github.com/kubeflow/pipelines/pull/12860
+  const serverNamespace = getServerNamespace();
+  const { accessKey, secretKey } =
+    namespace && namespace === serverNamespace
+      ? await getMinioClientSecrets(s3Artifact, namespace)
+      : {};
```
jeffspahr: The comment says this falls back to the server's own environment credentials, but `: {}` yields `accessKey`/`secretKey` of `undefined`, and `createMinioClient` only consults a credential chain for AWS S3 endpoints — for a non-AWS endpoint like SeaweedFS it logs `Encountered S3-compatible provider type with no provided credentials...` and builds an **anonymous** client, so this attempt fails and log retrieval proceeds to the third fallback in `getPodLogsHandler` (the configured archive settings, which do carry server credentials). Not a security issue, and no worse than the RBAC-denied behavior it replaces, but worth either (a) wiring the server's minio credentials here so the workflow-status path actually works for user namespaces against the shared store — which is this PR's headline use case — or (b) correcting the comment and skipping the doomed anonymous attempt so the chain falls through without the wasted round trip.

frontend/server/workflow-helper.ts
```diff
@@ -328,7 +334,45 @@ export async function getPodLogsMinioRequestConfigfromWorkflow(
...
+  // workflow-referenced Secret so custom object-store credentials are honored,
+  // while still refusing to read Secrets from any user namespace.
+  const serverNamespace = getServerNamespace();
+  let accessKey: string | undefined;
+  let secretKey: string | undefined;
+  if (namespace && namespace === serverNamespace) {
+    // Explicit server-namespace run (including multi-user runs whose namespace is
+    // the server namespace): read the Secret and use whatever it yields, exactly
+    // as before.
+    ({ accessKey, secretKey } = await getMinioClientSecrets(s3Artifact, namespace));
```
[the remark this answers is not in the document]
juliusvonkohout: That might be optional and the same behavior as on the master branch. Maybe something for a follow up PR.

juliusvonkohout:
/approve

@jeffspahr yes please follow up with https://github.com/kubeflow/pipelines/pull/13687 after this is merged

PR #13684, opened 2026-07-08
fix(backend): authorize ReadRunLogV1 before streaming run logs

naruto-lgtm:

The run log endpoint `/apis/v1alpha1/runs/{run_id}/nodes/{node_id}/log` streams pod logs through `ReadRunLogV1`, but unlike the sibling artifact endpoints (`ReadArtifact` / `ReadArtifactV1`) it never calls `canAccessRun`. In multi-user mode any authenticated user can read the logs of a run in a namespace they have no access to by supplying that run id, since the handler goes straight to `resourceManager.ReadLog`.

Add the same `canAccessRun` gate the artifact reader already uses, with the `get` verb, before any response is written. Keeping the check in the handler mirrors `ReadArtifact` and the gRPC run methods, so authorization sits next to the resource it guards rather than being pushed into `ReadLog`, which other read paths share. The added test drives the handler in multi-user mode with an unauthorized caller and asserts a 403.

**Checklist:**

### DIFF
--- a/backend/src/apiserver/server/run_log_server.go
+++ b/backend/src/apiserver/server/run_log_server.go
@@ -23,7 +23,10 @@ import (
 	"github.com/golang/glog"
 	"github.com/gorilla/mux"
 	api "github.com/kubeflow/pipelines/backend/api/v1beta1/go_client"
+	"github.com/kubeflow/pipelines/backend/src/apiserver/common"
 	"github.com/kubeflow/pipelines/backend/src/apiserver/resource"
+	"github.com/kubeflow/pipelines/backend/src/common/util"
+	authorizationv1 "k8s.io/api/authorization/v1"
 )
 
 const (
@@ -33,8 +36,8 @@ const (
 )
 
 type RunLogServer struct {
-	resourceManager *resource.ResourceManager
-	httpClient      *http.Client
+	*BaseRunServer
+	httpClient *http.Client
 }
 
 // Log streaming endpoint
@@ -58,6 +61,11 @@ func (s *RunLogServer) ReadRunLogV1(w http.ResponseWriter, r *http.Request) {
 
 	follow := vars[Follow] == "true" // defaults to false
 
+	if err := s.canAccessRun(r.Context(), runId, &authorizationv1.ResourceAttributes{Verb: common.RbacResourceVerbGet}); err != nil {
+		s.writeErrorToResponse(w, http.StatusForbidden, util.Wrap(err, "Failed to authorize the request"))
+		return
+	}
+
 	w.WriteHeader(http.StatusOK)
 	w.Header().Set("Content-Type", "text/plain")
 	w.Header().Set("Cache-Control", "no-cache, private")
@@ -80,5 +88,11 @@ func (s *RunLogServer) writeErrorToResponse(w http.ResponseWriter, code int, err
 }
 
 func NewRunLogServer(resourceManager *resource.ResourceManager) *RunLogServer {
-	return &RunLogServer{resourceManager: resourceManager, httpClient: http.DefaultClient}
+	return &RunLogServer{
+		BaseRunServer: &BaseRunServer{
+			resourceManager: resourceManager,
+			options:         &RunServerOptions{CollectMetrics: false},
+		},
+		httpClient: http.DefaultClient,
+	}
 }
--- a/backend/src/apiserver/server/run_log_server.go
+++ b/backend/src/apiserver/server/run_log_server.go
@@ -81,13 +81,15 @@ func (s *RunLogServer) ReadRunLogV1(w http.ResponseWriter, r *http.Request) {
 // so the caller identity arrives in the request headers, not in gRPC metadata.
 // Copy the headers into metadata so the authenticators can read them, as
 // canUploadVersionedPipeline does.
+// The dedicated readLog verb keeps log access behind a SubjectAccessReview even
+// in shared read mode, which auto-approves the shared get/list verbs.
 func (s *RunLogServer) authorize(r *http.Request, runID string) error {
 	md := metadata.MD{}
 	for key, values := range r.Header {
 		md.Set(key, values...)
 	}
 	ctx := metadata.NewIncomingContext(r.Context(), md)
-	return s.canAccessRun(ctx, runID, &authorizationv1.ResourceAttributes{Verb: common.RbacResourceVerbGet})
+	return s.canAccessRun(ctx, runID, &authorizationv1.ResourceAttributes{Verb: common.RbacResourceVerbReadLog})
 }
 
 func (s *RunLogServer) writeErrorToResponse(w http.ResponseWriter, code int, err error) {
--- a/backend/src/apiserver/server/run_log_server.go
+++ b/backend/src/apiserver/server/run_log_server.go
@@ -67,7 +67,9 @@ func (s *RunLogServer) ReadRunLogV1(w http.ResponseWriter, r *http.Request) {
 		return
 	}
 
-	w.WriteHeader(http.StatusOK)
+	// Set the success headers without committing a status code: the first
+	// streamed log write sends 200 implicitly, while a validation failure
+	// inside ReadLog can still produce a non-2xx JSON response.
 	w.Header().Set("Content-Type", "text/plain")
 	w.Header().Set("Cache-Control", "no-cache, private")
 
--- a/backend/src/apiserver/server/run_log_server.go
+++ b/backend/src/apiserver/server/run_log_server.go
@@ -78,6 +78,7 @@ func (s *RunLogServer) ReadRunLogV1(w http.ResponseWriter, r *http.Request) {
 
 func (s *RunLogServer) writeErrorToResponse(w http.ResponseWriter, code int, err error) {
 	glog.Errorf("Failed to read run log. Error: %+v", err)
+	w.Header().Set("Content-Type", "application/json")
 	w.WriteHeader(code)
 	errorResponse := &api.Error{ErrorMessage: err.Error(), ErrorDetails: fmt.Sprintf("%+v", err)}
 	errBytes, err := json.Marshal(errorResponse)
--- a/backend/src/apiserver/server/run_log_server.go
+++ b/backend/src/apiserver/server/run_log_server.go
@@ -81,13 +81,13 @@ func (s *RunLogServer) ReadRunLogV1(w http.ResponseWriter, r *http.Request) {
 // so the caller identity arrives in the request headers, not in gRPC metadata.
 // Copy the headers into metadata so the authenticators can read them, as
 // canUploadVersionedPipeline does.
-func (s *RunLogServer) authorize(r *http.Request, runId string) error {
+func (s *RunLogServer) authorize(r *http.Request, runID string) error {
 	md := metadata.MD{}
 	for key, values := range r.Header {
 		md.Set(key, values...)
 	}
 	ctx := metadata.NewIncomingContext(r.Context(), md)
-	return s.canAccessRun(ctx, runId, &authorizationv1.ResourceAttributes{Verb: common.RbacResourceVerbGet})
+	return s.canAccessRun(ctx, runID, &authorizationv1.ResourceAttributes{Verb: common.RbacResourceVerbGet})
 }
 
 func (s *RunLogServer) writeErrorToResponse(w http.ResponseWriter, code int, err error) {
--- a/backend/src/apiserver/server/run_log_server.go
+++ b/backend/src/apiserver/server/run_log_server.go
@@ -26,6 +26,7 @@ import (
 	"github.com/kubeflow/pipelines/backend/src/apiserver/common"
 	"github.com/kubeflow/pipelines/backend/src/apiserver/resource"
 	"github.com/kubeflow/pipelines/backend/src/common/util"
+	"google.golang.org/grpc/metadata"
 	authorizationv1 "k8s.io/api/authorization/v1"
 )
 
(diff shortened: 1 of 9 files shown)


backend/src/apiserver/server/run_log_server.go
```diff
@@ -58,6 +61,11 @@ func (s *RunLogServer) ReadRunLogV1(w http.ResponseWriter, r *http.Request) {
 
 	follow := vars[Follow] == "true" // defaults to false
 
+	if err := s.canAccessRun(r.Context(), runId, &authorizationv1.ResourceAttributes{Verb: common.RbacResourceVerbGet}); err != nil {
+		s.writeErrorToResponse(w, http.StatusForbidden, util.Wrap(err, "Failed to authorize the request"))
+		return
```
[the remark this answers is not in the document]
naruto-lgtm: Done, writeErrorToResponse now sets application/json before writing the status, and the unauthorized test asserts the header. Didn't touch the artifact server's ordering here to keep this scoped to the log endpoint.

naruto-lgtm:
any update?

backend/src/apiserver/server/run_log_server.go
```diff
@@ -58,6 +61,11 @@ func (s *RunLogServer) ReadRunLogV1(w http.ResponseWriter, r *http.Request) {
 
 	follow := vars[Follow] == "true" // defaults to false
 
+	if err := s.canAccessRun(r.Context(), runId, &authorizationv1.ResourceAttributes{Verb: common.RbacResourceVerbGet}); err != nil {
```
Priyanshu-u07: This route is registered directly on the mux, not through grpc-gateway, so the handler gets a plain `*http.Request` with no gRPC metadata. The authenticators only read identity from incoming metadata, so `canAccessRun` fails with `Unauthenticated` for every caller in multi-user mode, authorized users included. I ran locally with an authorized user and got `403: User identity is empty in the request header`.

`canUploadVersionedPipeline` handles this by copying `r.Header` into `metadata.MD` and wrapping the context before authorizing ([pipeline_upload_server.go#L412-L417](https://github.com/kubeflow/pipelines/blob/7b70c7b95f6832772efd6fd0029a4024491a0f44/backend/src/apiserver/server/pipeline_upload_server.go#L412-L417)). Same is needed here.
A test for the authorized path would catch it. The current one injects the identity as metadata, which this route never receives.

backend/src/apiserver/server/run_log_server_test.go
```diff
@@ -83,3 +89,32 @@ func TestReadRunLogV1_MissingNodeId(t *testing.T) {
...
+
+	// Deny access for the log read while keeping the seeded run intact, so the
+	// request reaches the authorization check rather than failing earlier.
+	clients.SubjectAccessReviewClientFake = client.NewFakeSubjectAccessReviewClientUnauthorized()
+	manager := resource.NewResourceManager(clients, &resource.ResourceManagerOptions{CollectMetrics: false})
+	server := NewRunLogServer(manager)
+
+	md := metadata.New(map[string]string{common.GoogleIAPUserIdentityHeader: common.GoogleIAPUserIdentityPrefix + "user@google.com"})
+	ctx := metadata.NewIncomingContext(context.Background(), md)
+	req := httptest.NewRequest("GET", "/test", nil).WithContext(ctx)
```
Priyanshu-u07: The router hands this handler a plain `*http.Request`. The identity arrives in `r.Header`, never as gRPC metadata, so building the context by hand means this test passes whether or not the handler can read the identity at all.
Set the header on the request instead:
```go
req := httptest.NewRequest("GET", "/apis/v1alpha1/runs/"+run.UUID+"/nodes/node-1/log", nil)
req.Header.Set(common.GoogleIAPUserIdentityHeader, common.GoogleIAPUserIdentityPrefix+"user@google.com")
```
Please also add the authorized case, since that is the one that catches this. It needs to assert `authorize()` rather than the full handler, because `FakePodClient.GetLogs` returns nil (see [pod_fake.go](https://github.com/kubeflow/pipelines/blob/7b70c7b95f6832772efd6fd0029a4024491a0f44/backend/src/apiserver/client/pod_fake.go#L88-L91)) and a successful run panics in `readRunLogFromPod`:
```go
func TestReadRunLogV1_AuthorizedOverPlainHTTP(t *testing.T) {
	viper.Set(common.MultiUserMode, "true")
	defer viper.Set(common.MultiUserMode, "false")

	clients, manager, run := initWithOneTimeRun(t) // default SAR fake authorizes
	defer clients.Close()
	server := NewRunLogServer(manager)

	req := httptest.NewRequest("GET", "/apis/v1alpha1/runs/"+run.UUID+"/nodes/node-1/log", nil)
	req.Header.Set(common.GoogleIAPUserIdentityHeader, common.GoogleIAPUserIdentityPrefix+"user@google.com")

	assert.NoError(t, server.authorize(req, run.UUID))
}
```

backend/src/apiserver/server/run_log_server_test.go
```diff
@@ -83,3 +89,32 @@ func TestReadRunLogV1_MissingNodeId(t *testing.T) {
...
+		RunKey:  run.UUID,
+		NodeKey: "node-1",
+	})
+
+	recorder := httptest.NewRecorder()
+	server.ReadRunLogV1(recorder, req)
+
+	assert.Equal(t, http.StatusForbidden, recorder.Code)
+	assert.Equal(t, "application/json", recorder.Header().Get("Content-Type"))
+	assert.Contains(t, recorder.Body.String(), "Failed to authorize")
```
Priyanshu-u07: `httptest.ResponseRecorder.Header()` returns the live header map, while `Result().Header` is the snapshot taken at `WriteHeader`, so this passes with or without the ordering fix:
```
wrong order:  rec.Header()="application/json"   rec.Result().Header=""
right order:  rec.Header()="application/json"   rec.Result().Header="application/json"
```
The body assertion has a similar gap: the handler wraps every failure path with "Failed to authorize the request", so it also matches a run that does not exist or one with an empty namespace. Asserting on the denial message pins the path you care about.
```suggestion
	assert.Equal(t, "application/json", recorder.Result().Header.Get("Content-Type"))
	assert.Contains(t, recorder.Body.String(), "Check if you have access to namespace")
```

Priyanshu-u07 (changes requested):
Thanks for working on this! `ReadLog` streams pod logs with no authorization at all, so the endpoint does need the check. One thing is blocking though: as written the check denies every caller in multi user mode, not just unauthorized ones. Left three comments.

backend/src/apiserver/server/run_log_server.go
naruto-lgtm: Good catch, the check never saw the identity. Added an authorize helper that copies r.Header into metadata.MD and wraps the request context before calling canAccessRun, same shape as canUploadVersionedPipeline.

backend/src/apiserver/server/run_log_server_test.go
naruto-lgtm: Done. The unauthorized test now sets the identity as a request header on the real route path, and I added TestReadRunLogV1_AuthorizedOverPlainHTTP asserting authorize() directly as you suggested, since the fake pod client makes the full success path panic.

backend/src/apiserver/server/run_log_server_test.go
naruto-lgtm: Switched to Result().Header for the snapshot at WriteHeader, and the body assertion now pins the namespace denial message instead of the generic wrapper.

backend/src/apiserver/server/run_log_server.go
```diff
@@ -68,8 +77,22 @@ func (s *RunLogServer) ReadRunLogV1(w http.ResponseWriter, r *http.Request) {
...
+// Copy the headers into metadata so the authenticators can read them, as
+// canUploadVersionedPipeline does.
+func (s *RunLogServer) authorize(r *http.Request, runId string) error {
+	md := metadata.MD{}
+	for key, values := range r.Header {
+		md.Set(key, values...)
+	}
+	ctx := metadata.NewIncomingContext(r.Context(), md)
+	return s.canAccessRun(ctx, runId, &authorizationv1.ResourceAttributes{Verb: common.RbacResourceVerbGet})
+}
```
Priyanshu-u07: Only lines this PR adds get linted, so no need to rename `runId` elsewhere in the file, just inside this new function.
```suggestion
func (s *RunLogServer) authorize(r *http.Request, runID string) error {
	md := metadata.MD{}
	for key, values := range r.Header {
		md.Set(key, values...)
	}
	ctx := metadata.NewIncomingContext(r.Context(), md)
	return s.canAccessRun(ctx, runID, &authorizationv1.ResourceAttributes{Verb: common.RbacResourceVerbGet})
}
```

naruto-lgtm: Done, renamed to runID in the helper. Left the pre-existing runId in the handler alone since only added lines get linted.

Priyanshu-u07 (approved):
Thank you @naruto-lgtm, everything addressed and CI is green. Looks good to me.

Priyanshu-u07:
@jeffspahr @alyssacgoins @hbelmiro this looks good to merge when you get the chance. Thanks!

jeffspahr (changes requested):
Requesting changes for one conditional security gap:

**[P1] Prevent shared-read mode from becoming an arbitrary pod-log proxy**

[`authorize`](https://github.com/kubeflow/pipelines/blob/afb266a2894724589eb4b907527953ce2cab1a1d/backend/src/apiserver/server/run_log_server.go#L84-L90) uses the ordinary `get` verb. When `MULTIUSER_SHARED_READ=true`, [`ResourceManager.IsAuthorized`](https://github.com/kubeflow/pipelines/blob/afb266a2894724589eb4b907527953ce2cab1a1d/backend/src/apiserver/resource/resource_manager.go#L2314-L2319) returns success for `get`/`list` before checking identity or issuing a SubjectAccessReview.

The handler then accepts a caller-controlled `node_id`. [`ReadLog`](https://github.com/kubeflow/pipelines/blob/afb266a2894724589eb4b907527953ce2cab1a1d/backend/src/apiserver/resource/resource_manager.go#L1152-L1184) derives the namespace from the selected run and calls `PodClient(namespace).GetLogs(nodeID, ...)` without confirming that the pod belongs to that run. A caller can therefore use any known foreign run as a namespace selector and read the `main` container logs of any known pod there using the API server's credentials. Even if shared-read mode is intended to expose run metadata or those runs' own logs, it should not expose unrelated workload logs in the namespace.

Please use a non-shared authorization verb such as a dedicated `readLog` permission, and bind the live pod to the requested run (for example, by validating its `pipeline/runid` label or stored workflow membership) before streaming. A regression test should cover `MULTIUSER_SHARED_READ=true` with a foreign run and a pod that does not belong to it.

naruto-lgtm:
Agreed, with shared read on the get verb was a no-op here and node_id was effectively a free pod selector for the run's namespace. Pushed a commit that switches authorize() to a dedicated readLog verb (granted in the multi-user view/edit cluster roles next to readArtifact so authorized users keep access), and makes readRunLogFromPod fetch the pod and check its pipeline/runid label against the requested run before streaming. All run creation paths (v1 argo template, v2 template, scheduled workflows) set that label on pod metadata, and the archive fallback was already scoped to the run's own workflow manifest so it didn't need the check. Added the two regression tests: MULTIUSER_SHARED_READ=true with an unauthorized caller still gets a 403, and a pod labeled with a different run id is rejected before any logs are read.

jeffspahr (changes requested):
Re-review of af944c89ce4d373d2f65fc045627c92f1d11619d: the earlier shared-read and cross-pod disclosure path is addressed. Two items remain.

**[P1] Rename the new runId parameter to runID**

The new readRunLogFromPod parameter at backend/src/apiserver/resource/resource_manager.go:1176 violates ST1003. The pre-commit check currently fails on this exact issue. Please rename the parameter and its references to runID.

**[P2] Do not commit 200 OK before pod/run validation succeeds**

ReadRunLogV1 calls WriteHeader(200) at backend/src/apiserver/server/run_log_server.go:70 before ResourceManager.ReadLog performs the new label check. A mismatched pod writes no logs and returns an error, but the later attempted 500 and JSON content type are ignored because the response is already committed. Clients therefore receive 200 with a JSON error presented as log content.

Please set the success headers before ReadLog without explicitly writing 200; the first streamed log write will commit 200 automatically, while a pre-stream validation failure can still return a non-2xx JSON response. Add a handler-level mismatched-label regression test that asserts the non-2xx status.

The dedicated readLog SAR, multi-user view/edit grants, pod label binding, and archive fallback otherwise look sound.

naruto-lgtm:
Both done in 5ac5d11. Renamed the parameter to runID, and the handler no longer writes 200 up front: it only sets the success headers, so the first streamed write commits 200 while a pod/run mismatch inside ReadLog now reaches the client as a non-2xx JSON error. The new handler-level mismatched-label test asserts the 500, and I checked it fails against the old ordering. To inject the mismatched pod at the handler level I exposed the fake client manager's core client field the same way the SAR fake already is.

jeffspahr (changes requested):
Re-review of `5ac5d11ffce2f0a87262686038ffcc66e43e5f68`: the earlier authorization and premature response-status findings are resolved. One blocker remains.

**[P1] Rename the adjacent `nodeId` parameter to `nodeID`**

The `runId` to `runID` update fixes the originally reported identifier, but the same rewritten `readRunLogFromPod` signature still declares `nodeId` at [resource_manager.go:1176](https://github.com/kubeflow/pipelines/blob/5ac5d11ffce2f0a87262686038ffcc66e43e5f68/backend/src/apiserver/resource/resource_manager.go#L1176). Required [pre-commit](https://github.com/kubeflow/pipelines/actions/runs/33402533992/job/99522178922) now fails ST1003: `method parameter nodeId should be nodeID`.

Please rename the parameter and its references to `nodeID`.

The migration and frontend failures appear unrelated to this change: all migration cases passed before cert-manager timed out, while the frontend job hit an external registry 403. `check_ci_status` is derivative.

naruto-lgtm:
Done in 65e8cf7, renamed nodeId to nodeID across readRunLogFromPod. Should have caught that one alongside runID. gofmt/vet clean and the resource and server package tests pass locally.

jeffspahr (approved):
Re-review of 65e8cf7cf6adf0d6aa89ef04be0d0f9e0abed959: no actionable findings. The earlier authorization, shared-read bypass, pod ownership, premature response status, and naming issues are resolved. Targeted run-log tests and the PR CI checks pass.

PR #13686, opened 2026-07-08
fix(manifests): raise SeaweedFS CPU request and add failure diagnostics

jeffspahr:
## Summary

- Raise the product SeaweedFS CPU request from `64m` to `128m`.
- Keep memory unchanged at `256Mi`.
- Make the change in the product manifest so CI standalone manifests and product manifests stay aligned.

## Context

This came from investigating an E2E flake seen while reviewing #13588. The failed lane passed on rerun, but the original failure had cluster/storage symptoms rather than a PR-specific code regression:

- The failing `E2EEssential` job ran with 10 Ginkgo nodes.
- It accumulated multiple flakes before `iris_pipeline_compiled.yaml` timed out.
- Failed Argo wait-container logs showed S3 artifact/log upload failures to `seaweedfs.kubeflow.svc:9000`, including TCP I/O timeouts.
- The PR under review changed `RetryRun`, while the stuck path was an ordinary `CreateRun` pipeline execution.

The standalone CI overlays include the product `platform-agnostic` manifests, so product manifest changes also apply to the CI lane that showed the flake.

## Why 128m

CPU requests are not CPU limits: SeaweedFS can burst above the request when node CPU is available. That means the flake data does not prove SeaweedFS is intrinsically CPU-bound, and a large product request such as `500m` would be too strong for the evidence.

This PR uses `128m` as a conservative request bump that gives SeaweedFS a little more reserved share under node contention while keeping product resource expectations close to the existing `64m` request. We should continue to monitor whether storage/DNS timeout flakes recur in CI before making any larger resource change.

## Local pressure test notes

I ran exploratory local Kind tests with the same 10-node `E2EEssential` Ginkgo shape at several SeaweedFS CPU request values:

| SeaweedFS CPU request | Local result |
| --- | --- |
| `200m` | Passed eventually, but had flakes and SeaweedFS S3 timeout output. |
| `250m` | Passed eventually, but had flakes and slow completion. |
| `300m` | One clean run. |
| `350m` | One clean run. |
| `400m` | Passed eventually, but had a flake and SeaweedFS S3 timeout output. |
| `450m` | Passed eventually, but had a flake with SeaweedFS S3 timeout and metadata DNS timeout output. |
| `500m` | Best local result: two runs, with the second run passing 20/20 with 0 flakes and no SeaweedFS timeout matches. |

These local results are useful for stress context, but they are not conclusive root-cause evidence. The explicit DNS timeout output in the local runs suggests broader Kind node/CoreDNS/network pressure may have contributed.

## Verification

- `git diff --check`
- Local 10-node `E2EEssential` exploratory runs during the flake investigation
- PR CI after opening this PR passed the relevant `E2EEssential` lanes with the original `500m` version; this reduced `128m` version still needs normal PR CI signal

## Diagnostics: validate the lever on the next occurrence

The CPU-request bump is a conservative, evidence-thin mitigation (as the table above shows, local stress could not conclusively attribute the flake to SeaweedFS CPU). To stop guessing, this PR also adds `collect-seaweedfs-diagnostics.sh`, invoked from the existing failure log-collection step in `test-and-report`, which captures on every E2E test failure the exact evidence that discriminates the three candidate causes of the `dial tcp <seaweedfs>:9000: i/o timeout` signature:

1. **SeaweedFS crashed/restarted** — pod restart count and last terminated state (OOMKilled, etc.). If non-zero, dial timeouts are the pod being down and CPU request is not the lever.
2. **Node CPU/memory contention** — the node's allocated-requests-vs-allocatable and pressure conditions. The deployment sets a CPU *request* with no limit, so scheduling shares only matter when the node is contended; if the node is idle at failure time, a request bump will not help. This is the evidence that directly validates or refutes this PR's change.
3. **Cluster dataplane failure** — SeaweedFS Running with no restarts and a healthy node, yet the ClusterIP path times out (kube-proxy / CNI). Ruled in by eliminating 1 and 2.

`metrics-server` is not deployed in these Kind clusters, so runtime CPU utilization is unavailable; node allocated-requests stand in for scheduling contention. The script is best-effort (a missing object never fails the job) and its output goes to both the step log and the uploaded pod-log artifact.

So the two halves of this PR are complementary: the request bump is a low-risk attempt at mitigation, and the diagnostics make the *next* dial-timeout failure produce the data to confirm whether it worked or point at the real cause.

### DIFF
--- a/.github/resources/scripts/collect-seaweedfs-diagnostics.sh
+++ b/.github/resources/scripts/collect-seaweedfs-diagnostics.sh
@@ -32,9 +32,9 @@ OUTPUT_FILE=""
 
 while [[ "$#" -gt 0 ]]; do
     case "$1" in
-        --ns) NS="$2"; shift ;;
-        --selector) SELECTOR="$2"; shift ;;
-        --output) OUTPUT_FILE="$2"; shift ;;
+        --ns) NS="${2:?--ns requires a value}"; shift ;;
+        --selector) SELECTOR="${2:?--selector requires a value}"; shift ;;
+        --output) OUTPUT_FILE="${2:?--output requires a value}"; shift ;;
         *) echo "Unknown parameter passed: $1"; exit 1 ;;
     esac
     shift
@@ -52,7 +52,10 @@ emit() {
 {
     echo "================ SeaweedFS failure diagnostics ================"
 
-    POD=$(kubectl get pod -n "$NS" -l "$SELECTOR" -o jsonpath='{.items[0].metadata.name}' 2>/dev/null || true)
+    # Newest matching pod: with strategy Recreate there is normally one, but a
+    # rollout can briefly leave a Terminating pod alongside the new one, and the
+    # new pod is the one whose state matters.
+    POD=$(kubectl get pod -n "$NS" -l "$SELECTOR" --sort-by=.metadata.creationTimestamp -o jsonpath='{.items[-1].metadata.name}' 2>/dev/null || true)
     if [[ -z "$POD" ]]; then
         echo "No SeaweedFS pod (selector '$SELECTOR') found in namespace '$NS'; skipping."
         echo "=============================================================="
@@ -87,8 +90,10 @@ emit() {
     # If SeaweedFS is Running with no restarts and the node is healthy but the
     # ClusterIP still times out, the failure is in the service dataplane.
     kubectl get pod -n kube-system -l k8s-app=kube-proxy -o wide 2>/dev/null || true
-    kubectl get endpoints -n "$NS" -l "$SELECTOR" -o wide 2>/dev/null \
-        || kubectl get endpoints seaweedfs -n "$NS" -o wide 2>/dev/null || true
+    # Query Endpoints by service name: the SeaweedFS Service has no labels, so a
+    # label selector matches nothing (and would still exit 0, masking the miss).
+    # Empty ENDPOINTS here means the Service has no ready backends.
+    kubectl get endpoints seaweedfs -n "$NS" -o wide 2>/dev/null || true
 
     echo
     echo "----- full describe (events, resource config) -----"
--- a/.github/resources/scripts/collect-seaweedfs-diagnostics.sh
+++ b/.github/resources/scripts/collect-seaweedfs-diagnostics.sh
@@ -11,7 +11,7 @@
 #
 #   1. SeaweedFS crashed/restarted   -> pod restart count and last terminated
 #                                        state (OOMKilled, etc.)
-#   2. Node CPU/mem contention       -> node allocated requests vs allocatable
+#   2. Node CPU/memory contention    -> node allocated requests vs allocatable
 #                                        and node pressure conditions. The
 #                                        SeaweedFS deployment sets a CPU request
 #                                        (no limit), so scheduling shares only
@@ -26,13 +26,13 @@
 
 set -u
 
-NS="kubeflow"
+NAMESPACE="kubeflow"
 SELECTOR="app=seaweedfs"
 OUTPUT_FILE=""
 
 while [[ "$#" -gt 0 ]]; do
     case "$1" in
-        --ns) NS="${2:?--ns requires a value}"; shift ;;
+        --namespace) NAMESPACE="${2:?--namespace requires a value}"; shift ;;
         --selector) SELECTOR="${2:?--selector requires a value}"; shift ;;
         --output) OUTPUT_FILE="${2:?--output requires a value}"; shift ;;
         *) echo "Unknown parameter passed: $1"; exit 1 ;;
@@ -55,9 +55,9 @@ emit() {
     # Newest matching pod: with strategy Recreate there is normally one, but a
     # rollout can briefly leave a Terminating pod alongside the new one, and the
     # new pod is the one whose state matters.
-    POD=$(kubectl get pod -n "$NS" -l "$SELECTOR" --sort-by=.metadata.creationTimestamp -o jsonpath='{.items[-1].metadata.name}' 2>/dev/null || true)
+    POD=$(kubectl get pod -n "$NAMESPACE" -l "$SELECTOR" --sort-by=.metadata.creationTimestamp -o jsonpath='{.items[-1].metadata.name}' 2>/dev/null || true)
     if [[ -z "$POD" ]]; then
-        echo "No SeaweedFS pod (selector '$SELECTOR') found in namespace '$NS'; skipping."
+        echo "No SeaweedFS pod (selector '$SELECTOR') found in namespace '$NAMESPACE'; skipping."
         echo "=============================================================="
         exit 0
     fi
@@ -67,10 +67,10 @@ emit() {
     echo "----- (1) restarts / last terminated state -----"
     # A non-zero restart count or an OOMKilled/Error last state means the dial
     # timeouts are the pod being down, and the CPU request is not the lever.
-    kubectl get pod "$POD" -n "$NS" -o wide 2>/dev/null || true
-    kubectl get pod "$POD" -n "$NS" -o jsonpath='restartCount={.status.containerStatuses[0].restartCount}{"\n"}lastState={.status.containerStatuses[0].lastState}{"\n"}qosClass={.status.qosClass}{"\n"}' 2>/dev/null || true
+    kubectl get pod "$POD" -n "$NAMESPACE" -o wide 2>/dev/null || true
+    kubectl get pod "$POD" -n "$NAMESPACE" -o jsonpath='restartCount={.status.containerStatuses[0].restartCount}{"\n"}lastState={.status.containerStatuses[0].lastState}{"\n"}qosClass={.status.qosClass}{"\n"}' 2>/dev/null || true
 
-    NODE=$(kubectl get pod "$POD" -n "$NS" -o jsonpath='{.spec.nodeName}' 2>/dev/null || true)
+    NODE=$(kubectl get pod "$POD" -n "$NAMESPACE" -o jsonpath='{.spec.nodeName}' 2>/dev/null || true)
 
     echo
     echo "----- (2) node contention on $NODE -----"
@@ -93,11 +93,11 @@ emit() {
     # Query Endpoints by service name: the SeaweedFS Service has no labels, so a
     # label selector matches nothing (and would still exit 0, masking the miss).
     # Empty ENDPOINTS here means the Service has no ready backends.
-    kubectl get endpoints seaweedfs -n "$NS" -o wide 2>/dev/null || true
+    kubectl get endpoints seaweedfs -n "$NAMESPACE" -o wide 2>/dev/null || true
 
     echo
     echo "----- full describe (events, resource config) -----"
-    kubectl describe pod "$POD" -n "$NS" 2>/dev/null || true
+    kubectl describe pod "$POD" -n "$NAMESPACE" 2>/dev/null || true
 
     echo "=============================================================="
 } | emit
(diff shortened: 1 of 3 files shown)


juliusvonkohout:
It works in large clusters with the current settings.

juliusvonkohout:
So going to 128m might be fine, but i would not go higher

juliusvonkohout:
Requests != limits so seaweedfs can burst to what is available on the node also 2048m for example. That works even for very large clusters in practice and idle is usually way less than 64 m.

jeffspahr:
Added SeaweedFS failure diagnostics (84871d7d3) alongside the CPU request bump. Findings from digging into the last two days of E2E lane failures that motivated this:

- The `dial tcp <seaweedfs>:9000: i/o timeout` signature (100-150 occurrences per failing lane) is connection *establishment* failing, not slow uploads.
- Of ~23 raw "E2E core lane" failures in the window, most were not this class — 16 were a release-bump PR's executor pods failing deterministically (unpublished version), and a few were feature-PR bugs. The genuine SeaweedFS dial-timeout flake was ~3 lanes, all on CI-only PRs that can't affect runtime.
- Because the deployment sets a CPU *request* with no limit, CFS throttling isn't the mechanism; the request only affects proportional shares under node contention. The diagnostics capture node allocated-requests so we can see whether that contention is actually present when the timeouts happen.

The bump stays as a low-risk mitigation; the diagnostics are what let us confirm or refute it on the next occurrence instead of arguing from failure shape.

jeffspahr:
Thanks @juliusvonkohout.  I had better results with a higher request when testing locally, but it's pretty inconclusive.  I'm changing this to 128m like you suggested, and adding diagnostic info so we can get more info if it keeps flaking.

Please approve/merge if you're good with this.

.github/resources/scripts/collect-seaweedfs-diagnostics.sh
```diff
@@ -0,0 +1,98 @@
...
+
+while [[ "$#" -gt 0 ]]; do
+    case "$1" in
+        --ns) NS="$2"; shift ;;
+        --selector) SELECTOR="$2"; shift ;;
+        --output) OUTPUT_FILE="$2"; shift ;;
+        *) echo "Unknown parameter passed: $1"; exit 1 ;;
+    esac
+    shift
+done
```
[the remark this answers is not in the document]
jeffspahr: Fixed in 67f246303 — valued flags now use `${2:?--<flag> requires a value}`, so a missing value fails with a clear message instead of a `set -u` unbound-variable abort.

.github/resources/scripts/collect-seaweedfs-diagnostics.sh
```diff
@@ -0,0 +1,98 @@
...
+        tee -a "$OUTPUT_FILE"
+    else
+        cat
+    fi
+}
+
+{
+    echo "================ SeaweedFS failure diagnostics ================"
+
+    POD=$(kubectl get pod -n "$NS" -l "$SELECTOR" -o jsonpath='{.items[0].metadata.name}' 2>/dev/null || true)
```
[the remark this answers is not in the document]
jeffspahr: Fixed in 67f246303 — now selects the newest matching pod via `--sort-by=.metadata.creationTimestamp` and `{.items[-1].metadata.name}`. The deployment uses strategy Recreate so there is normally one pod, but a rollout can briefly leave a Terminating pod alongside the new one, and the new pod is the one whose state matters.

.github/resources/scripts/collect-seaweedfs-diagnostics.sh
```diff
@@ -0,0 +1,98 @@
...
+        echo "Could not resolve SeaweedFS node."
+    fi
+
+    echo
+    echo "----- (3) dataplane: kube-proxy / CNI -----"
+    # If SeaweedFS is Running with no restarts and the node is healthy but the
+    # ClusterIP still times out, the failure is in the service dataplane.
+    kubectl get pod -n kube-system -l k8s-app=kube-proxy -o wide 2>/dev/null || true
+    kubectl get endpoints -n "$NS" -l "$SELECTOR" -o wide 2>/dev/null \
+        || kubectl get endpoints seaweedfs -n "$NS" -o wide 2>/dev/null || true
```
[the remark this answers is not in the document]
jeffspahr: Good catch — fixed in 67f246303. Confirmed the SeaweedFS Service (seaweedfs-service.yaml) has no metadata.labels, so the selector matched nothing while still exiting 0, leaving the dataplane check silent. Now queries `kubectl get endpoints seaweedfs` by name directly; verified it returns the real backend endpoints against a live cluster.

.github/resources/scripts/collect-seaweedfs-diagnostics.sh
```diff
@@ -0,0 +1,103 @@
...
+# E2E lanes intermittently fail with many
+#   dial tcp <seaweedfs-clusterip>:9000: i/o timeout
+# errors while pipelines upload artifacts and logs. That signature (the TCP
+# connection is never established) has three candidate causes, and this script
+# captures the evidence that distinguishes them without needing metrics-server
+# (which these Kind clusters do not run):
+#
+#   1. SeaweedFS crashed/restarted   -> pod restart count and last terminated
+#                                        state (OOMKilled, etc.)
+#   2. Node CPU/mem contention       -> node allocated requests vs allocatable
```
juliusvonkohout:
```suggestion
#   2. Node CPU/memory contention       -> node allocated requests vs allocatable
```

.github/resources/scripts/collect-seaweedfs-diagnostics.sh
```diff
@@ -0,0 +1,103 @@
...
+
+while [[ "$#" -gt 0 ]]; do
+    case "$1" in
+        --ns) NS="${2:?--ns requires a value}"; shift ;;
+        --selector) SELECTOR="${2:?--selector requires a value}"; shift ;;
+        --output) OUTPUT_FILE="${2:?--output requires a value}"; shift ;;
+        *) echo "Unknown parameter passed: $1"; exit 1 ;;
+    esac
+    shift
+done
```
juliusvonkohout:
```suggestion
NAMESPACE="kubeflow"
SELECTOR="app=seaweedfs"
OUTPUT_FILE=""

while [[ "$#" -gt 0 ]]; do
    case "$1" in
        --namespace) NAMESPACE="${2:?--namespace requires a value}"; shift ;;
        --selector) SELECTOR="${2:?--selector requires a value}"; shift ;;
        --output) OUTPUT_FILE="${2:?--output requires a value}"; shift ;;
        *) echo "Unknown parameter passed: $1"; exit 1 ;;
    esac
    shift
done
```

.github/resources/scripts/collect-seaweedfs-diagnostics.sh
```diff
@@ -0,0 +1,103 @@
...
+    fi
+}
+
+{
+    echo "================ SeaweedFS failure diagnostics ================"
+
+    # Newest matching pod: with strategy Recreate there is normally one, but a
+    # rollout can briefly leave a Terminating pod alongside the new one, and the
+    # new pod is the one whose state matters.
+    POD=$(kubectl get pod -n "$NS" -l "$SELECTOR" --sort-by=.metadata.creationTimestamp -o jsonpath='{.items[-1].metadata.name}' 2>/dev/null || true)
```
juliusvonkohout:
```suggestion
    POD=$(kubectl get pod -n "$NAMESPACE" -l "$SELECTOR" --sort-by=.metadata.creationTimestamp -o jsonpath='{.items[-1].metadata.name}' 2>/dev/null || true)
```

.github/resources/scripts/collect-seaweedfs-diagnostics.sh
```diff
@@ -0,0 +1,103 @@
...
+    if [[ -z "$POD" ]]; then
+        echo "No SeaweedFS pod (selector '$SELECTOR') found in namespace '$NS'; skipping."
```
juliusvonkohout:
```suggestion
        echo "No SeaweedFS pod (selector '$SELECTOR') found in namespace '$NAMESPACE'; skipping."
```

.github/resources/scripts/collect-seaweedfs-diagnostics.sh
```diff
@@ -0,0 +1,103 @@
...
+    echo "SeaweedFS pod: $POD"
+
+    echo
+    echo "----- (1) restarts / last terminated state -----"
+    # A non-zero restart count or an OOMKilled/Error last state means the dial
+    # timeouts are the pod being down, and the CPU request is not the lever.
+    kubectl get pod "$POD" -n "$NS" -o wide 2>/dev/null || true
+    kubectl get pod "$POD" -n "$NS" -o jsonpath='restartCount={.status.containerStatuses[0].restartCount}{"\n"}lastState={.status.containerStatuses[0].lastState}{"\n"}qosClass={.status.qosClass}{"\n"}' 2>/dev/null || true
+
+    NODE=$(kubectl get pod "$POD" -n "$NS" -o jsonpath='{.spec.nodeName}' 2>/dev/null || true)
```
juliusvonkohout:
```suggestion
    kubectl get pod "$POD" -n "$NAMESPACE" -o wide 2>/dev/null || true
    kubectl get pod "$POD" -n "$NAMESPACE" -o jsonpath='restartCount={.status.containerStatuses[0].restartCount}{"\n"}lastState={.status.containerStatuses[0].lastState}{"\n"}qosClass={.status.qosClass}{"\n"}' 2>/dev/null || true

    NODE=$(kubectl get pod "$POD" -n "$NAMESPACE" -o jsonpath='{.spec.nodeName}' 2>/dev/null || true)
```

.github/resources/scripts/collect-seaweedfs-diagnostics.sh
jeffspahr: Applied in 5c7697f6c — renamed to `--namespace`/`NAMESPACE` throughout (and updated the caller in test-and-report), plus spelled out "CPU/memory". Verified no residual `NS`/`--ns` and the script runs against a live cluster.

.github/resources/scripts/collect-seaweedfs-diagnostics.sh
jeffspahr: Applied in 5c7697f6c — renamed to `--namespace`/`NAMESPACE` throughout (and updated the caller in test-and-report), plus spelled out "CPU/memory". Verified no residual `NS`/`--ns` and the script runs against a live cluster.

.github/resources/scripts/collect-seaweedfs-diagnostics.sh
jeffspahr: Applied in 5c7697f6c — renamed to `--namespace`/`NAMESPACE` throughout (and updated the caller in test-and-report), plus spelled out "CPU/memory". Verified no residual `NS`/`--ns` and the script runs against a live cluster.

.github/resources/scripts/collect-seaweedfs-diagnostics.sh
jeffspahr: Applied in 5c7697f6c — renamed to `--namespace`/`NAMESPACE` throughout (and updated the caller in test-and-report), plus spelled out "CPU/memory". Verified no residual `NS`/`--ns` and the script runs against a live cluster.

.github/resources/scripts/collect-seaweedfs-diagnostics.sh
jeffspahr: Applied in 5c7697f6c — renamed to `--namespace`/`NAMESPACE` throughout (and updated the caller in test-and-report), plus spelled out "CPU/memory". Verified no residual `NS`/`--ns` and the script runs against a live cluster.

juliusvonkohout:
/lgtm 
Let's try it out then. If doubling the CPU request does not help it is probably something else.

jeffspahr:
Yeah the extra diagnostic info will help here even if the resource change doesn't.  Thanks. Could you /approve too?

zazulam (commented):
/approve

PR #13687, opened 2026-07-08
test(frontend): exercise the real MLMD validator through the artifacts auth middleware. Part of #13683

jeffspahr:

Targets the #13683 branch (`fix-frontend-artifact-idor-prefix-fallback`) directly — merging this PR adds the test commits straight into #13683, per the author's request. The diff contains only test files.

#13433 introduced a regression (all artifacts absent from MLMD denied, breaking v1 log/artifact retrieval) that no CI lane could catch:

- The frontend e2e lane deploys standalone, where `authEnabled` is false and the artifacts auth middleware short-circuits before the MLMD validator runs.
- The multi-user e2e lanes exercise the API server, never ml-pipeline-ui's `/artifacts/*` routes.
- The existing `artifact-auth.test.ts` integration tests mock `validateArtifactNamespace` entirely, so the middleware ↔ validator wiring for MLMD-untracked objects had no coverage at any level.

This adds an integration suite (`artifact-idor-fallback.test.ts`) that runs the **real** MLMD validator end to end through the Express app, mocking only the gRPC-web transport with framed MLMD proto responses (and the MinIO client):

- serves an archived pod log absent from MLMD when its `private-artifacts/<namespace>/` key prefix matches the claimed namespace — this is the regression guard: it **fails against current master** and passes with #13683's prefix fallback
- denies untracked objects whose prefix names another namespace, or that carry no owning-namespace prefix, and asserts the object store is never contacted
- denies a dot-segment key (`private-artifacts/team-a/../victim-ns/...`) that spoofs the claimed namespace prefix, covering the traversal hardening added from review
- prefers MLMD namespace evidence over a matching object-key prefix (fallback only applies when MLMD has no record)
- serves an MLMD-tracked artifact whose PipelineRun context matches the claimed namespace
- pins the default-mode denial of a bucket-per-namespace v1 artifact (`s3://team-a-bucket/...`, no owning-namespace key prefix) — documents the compatibility gap raised in review; if the proposed opt-in bucket-ownership mode is added, it needs its own allow/deny cases while this default-mode denial must keep holding for shared-bucket installs

It also extends `mlmd-validator.test.ts` with unit cases for the review hardening: dot-segment / single-dot / empty key segments denied as `key-not-normalized`, and an unrecognized `ARTIFACT_NAMESPACE_OWNERSHIP_MODE` failing closed to the strict denial.

Verification: on top of the current #13683 head the full frontend server suite is green (356 passed, 2 todo); with master's validator swapped in, exactly the archived-log test fails, confirming the suite would have caught the #13433 regression.

Follow-up (not in this PR): a multi-user variant of the frontend integration CI lane would close the remaining end-to-end gap; that needs auth-header injection and namespace-scoped assertions in the webdriverio specs and is worth tracking separately.

**Checklist:**

### DIFF
--- a/frontend/server/helpers/mlmd-validator.test.ts
+++ b/frontend/server/helpers/mlmd-validator.test.ts
@@ -327,4 +327,38 @@ describe('decideFromPrefixFallback', () => {
       reason: 'prefix-match',
     });
   });
+
+  it('denies a key that hides a dot-segment traversal behind a matching prefix', () => {
+    expect(
+      decideFromPrefixFallback(
+        'minio://mlpipeline/private-artifacts/team-a/../victim-ns/obj',
+        'team-a',
+        'mlmd-then-prefix',
+      ),
+    ).toEqual({ valid: false, reason: 'key-not-normalized' });
+  });
+
+  it('denies a key containing a single-dot or empty path segment', () => {
+    expect(
+      decideFromPrefixFallback(
+        'minio://mlpipeline/private-artifacts/team-a/./obj',
+        'team-a',
+        'mlmd-then-prefix',
+      ),
+    ).toEqual({ valid: false, reason: 'key-not-normalized' });
+    expect(
+      decideFromPrefixFallback(
+        'minio://mlpipeline/private-artifacts/team-a//obj',
+        'team-a',
+        'mlmd-then-prefix',
+      ),
+    ).toEqual({ valid: false, reason: 'key-not-normalized' });
+  });
+
+  it('fails closed to the strict denial for an unrecognized ownership mode', () => {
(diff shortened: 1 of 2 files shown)


jeffspahr:
Added a sixth case covering the bucket-per-namespace scenario @droctothorpe raised on #13683: a v1 artifact in a dedicated namespace bucket (`s3://team-a-bucket/...`, no owning-namespace key prefix) is denied under the default `mlmd-then-prefix` mode. This pins today's behavior and marks the gap the proposed opt-in bucket-ownership mode would fill — if that mode lands in #13683, this suite should gain namespace-bucket-allow and foreign-bucket-deny cases for it (the `k8s-helper` mock already covers the `artifact-repositories` ConfigMap lookup path).

juliusvonkohout:
Please rebase against master

PR #12916, opened 2026-02-26
chore(backend): Increase timeout for PVC cache test to reduce flakiness

aniketpati1121:
Test `TestCacheSingleRunWithPVC_SameName_Caches` intermittently fails in CI because PVC provisioning can be slow on Kind clusters. Increased the timeout from 2 minutes to 4 minutes to reduce flakiness. No other logic changes; this is a safe change for CI stability.  fixes #12913

PR #13693, opened 2026-07-08
ci: give legacy v2 integration tests an intentional time budget

jeffspahr:
## The failure

Ten lane failures across the last two days of PR CI share one signature in the `KFP API Integration V2 (Legacy)` workflow:

```
panic: test timed out after 10m0s
    running tests:
        TestPipelineVersionAPI (0s)
```

The `go test` invocation passes no `-timeout`, so the binary runs under Go's default **10-minute** test timeout. When it fires, Go does not fail the current test — it kills the entire binary with a panic and a goroutine dump, discarding every already-passed result and every test not yet run.

## Why the suite trips it

Measured on green runs, the test step takes **10.5–12.5 minutes**, of which ~1.5 min is compilation (the timeout clock starts when the compiled binary starts) — so the binary's normal runtime is **~9–9.5 minutes against a 10-minute fuse: under a minute of headroom**.

The runtime is legitimate, not waste. `TestCache` alone is ~7 of those minutes (74% of the suite), and its three subtests each execute **sequential full pipeline runs by definition** — a cache test must complete a run, then run the same pipeline again to observe the cache hit:

| Test | Time |
|---|---|
| `TestCache/TestCacheSingleRunWithPVC_SameName_Caches` | 191s |
| `TestCache/TestCacheRecurringRun` | 151s |
| `TestCache/TestCacheSingleRun` | 81s |

With <5% headroom, ordinary lane slowness — a busy runner, slower cluster operations — tips the cumulative runtime past 10:00. The panic evidence confirms cumulative slowness rather than a hang: the interrupted test shows **`(0s)`** — it had just started when the alarm fired; the budget was consumed a few seconds at a time by everything before it. This also explains why the failure looks random: which test gets caught depends only on where the clock runs out. Historical note: this suite ran at the same ~11-minute step time on K8s v1.35.0 before the recent version bump, so this is a chronic budget problem, not a regression.

## The changes

1. **`-timeout 20m`** on the `go test` invocation — roughly 2× normal runtime. This matches the headroom ratio this suite's own past flake fixes settled on (#11916 and #12916 both resolved cache-test flakiness by raising wait ceilings to ~2× nominal). A genuinely hung test still dies, with a budget that matches what the suite actually is.
2. **`timeout-minutes: 40`** on the job — it previously inherited GitHub's **6-hour** default, meaning the go-test timeout was the only effective hang-stop. A green job takes ~20–23 minutes end to end; 40 gives the enlarged test budget plus setup and log collection real room while capping a true hang at a sane cost.

## Making future overruns self-explanatory

When a `-timeout` still fires, the raw signal is a panic and goroutine dump buried mid-log, which reads like a crash. The test step now tees its output, and a follow-up step (which runs only when the failure is a budget overrun — ordinary test failures are untouched) converts it into:

- a **GitHub error annotation** on the job and PR checks UI stating the budget that was exceeded and that a near-zero elapsed time on the interrupted test means cumulative slowness, not a hang
- a **job summary** with the interrupted-test block from the panic and explicit guidance: if this recurs on otherwise healthy PRs, the suite has outgrown its budget — raise `-timeout` or speed up the suite instead of rerunning repeatedly

Both paths are functionally tested against a captured panic block (budget string is extracted dynamically, so the message stays correct if the budget changes).

## What this deliberately does not do

Alternatives considered and set aside, with reasons:

- **Shrinking `TestCacheRecurringRun`'s 60-second schedule interval**: the interval is load-bearing, not arbitrary. The test's wait predicate requires *every* run in the namespace to be `SUCCEEDED` before asserting; the hello-world run takes ~35s, so any interval below run duration (10s, even 30s) spawns overlapping runs faster than they finish and the predicate can never hold — a deterministic failure, not a speedup.
- **Parallelizing the three cache subtests** (~4 min potential): blocked by the same predicate — `ListAll` sees sibling tests' runs and indexes `allRuns[1]` from the shared namespace list. Parallelizing safely requires scoping the run listing to each test's own experiment first. Worth doing as a follow-up test refactor; not a workflow budget fix.

## Verification

- YAML parse of the modified workflow
- This PR's own `KFP API Integration V2 (Legacy)` lanes run the suite under the new budget

### DIFF
--- a/.github/workflows/legacy-v2-api-integration-tests.yml
+++ b/.github/workflows/legacy-v2-api-integration-tests.yml
@@ -97,11 +97,42 @@ jobs:
         # by cache tests that execute sequential pipeline runs. Go's default
         # 10-minute test binary timeout leaves under a minute of headroom, so
         # ordinary lane slowness kills the whole binary mid-run with a panic.
-        run: go test -v -timeout 20m ./... -args -runIntegrationTests=true -namespace=kubeflow -tlsEnabled=${{ matrix.pod_to_pod_tls_enabled }} -caCertPath=${{ env.CA_CERT_PATH }} -repoName=${{ github.repository }} -branchName=${{ github.ref_name }}
+        # Output is teed to a file so the step below can turn a budget
+        # overrun into an explicit annotation instead of a buried panic.
+        shell: bash
+        run: go test -v -timeout 20m ./... -args -runIntegrationTests=true -namespace=kubeflow -tlsEnabled=${{ matrix.pod_to_pod_tls_enabled }} -caCertPath=${{ env.CA_CERT_PATH }} -repoName=${{ github.repository }} -branchName=${{ github.ref_name }} 2>&1 | tee "$RUNNER_TEMP/go-test-output.log"
         env:
           PULL_NUMBER: ${{ github.event.pull_request.number }}
           PIPELINE_STORE: ${{ matrix.pipeline_store }}
         continue-on-error: true
+
+      - name: Explain suite time budget overrun
+        if: ${{ steps.tests.outcome != 'success' }}
+        shell: bash
+        run: |
+          LOG="$RUNNER_TEMP/go-test-output.log"
+          if [ ! -f "$LOG" ] || ! grep -q 'panic: test timed out after' "$LOG"; then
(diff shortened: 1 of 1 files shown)


hbelmiro (approved):
/lgtm
/approve

PR #13701, opened 2026-07-09
ci: dump pod state when multi-user readiness waits time out

jeffspahr:
## Problem

The multi-user E2E lane intermittently fails in `Deploy KFP` with:

```
error: timed out waiting for the condition on pods/profiles-deployment-...
```

The profile-controller pod does not reach `condition=Ready` within the 180s wait in `deploy-kfp.sh`. It is **intermittent** (multi-user lanes pass ~80% of completed runs), so it is not a systemic break of the `kubeflow/manifests` `ref=master` sources it is deployed from — it is a readiness-lag flake on the loaded v1.36.1 lane. The pod is Istio-sidecar-injected, so "not Ready" is often the `istio-proxy` sidecar, but **we cannot tell**: the wait is a bare `kubectl wait` under `set -e`, so a timeout exits with that single line and **no pod state** — every occurrence is undiagnosable.

## Change

Add `wait_for_pods_ready`, which waits for `condition=Ready` and, on timeout, dumps before failing:
- pod list (`-o wide`, restart counts / READY column),
- `describe` (per-container readiness and pod events — surfaces sidecar-not-ready and image-pull),
- recent namespace events,
- all-container logs including the `istio-proxy` sidecar.

Applied to the two readiness waits in the multi-user block (Istio pods and the profile-controller pod). The success path is unchanged.

This does not attempt to *fix* the flake — deliberately. Bumping the timeout or pinning manifests without knowing whether the slow component is the sidecar, an image pull, or controller init would be guessing. This makes the next occurrence produce the evidence to fix it precisely.

## Verification

- `bash -n` clean
- Functionally exercised both paths against a live cluster: the ready path returns 0 for existing pods; the timeout path prints all diagnostic sections and then fails under `set -e`

### DIFF
--- a/.github/resources/scripts/deploy-kfp.sh
+++ b/.github/resources/scripts/deploy-kfp.sh
@@ -140,9 +140,19 @@ wait_for_pods_ready() {
   echo "----- describe (container readiness / events) -----"
   kubectl "${scope_args[@]}" describe pods || true
   echo "----- recent namespace events -----"
-  kubectl -n "$namespace" get events --sort-by=.lastTimestamp 2>/dev/null | tail -40 || true
+  # Sort by creationTimestamp: lastTimestamp is deprecated/absent for
+  # events.k8s.io/v1 objects, which leaves the output unsorted on new clusters.
+  kubectl -n "$namespace" get events --sort-by=.metadata.creationTimestamp 2>/dev/null | tail -40 || true
   echo "----- logs (all containers, incl. istio-proxy sidecar) -----"
-  kubectl "${scope_args[@]}" logs --all-containers=true --tail=100 --prefix=true || true
+  if [ -n "$selector" ]; then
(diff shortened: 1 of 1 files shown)


jeffspahr:
Both findings addressed in the latest commit:

1. **Selector-less log dump fixed** — `kubectl logs` needs a resource or selector, so the Istio (namespace-wide) timeout path could never produce logs. With no selector the helper now dumps each pod individually (bounded to 20); the label-selector path is unchanged (and no longer risks a duplicated `-l` from the shared scope args).
2. **Event sorting** — switched to `.metadata.creationTimestamp`; `.lastTimestamp` is deprecated/absent for `events.k8s.io/v1` objects and left output unsorted on newer clusters.

Verified with a stubbed `kubectl` exercising both paths: empty selector → per-pod `logs` calls + creationTimestamp sort; non-empty selector → single `-l` logs call.

PR #13706, opened 2026-07-11
ci(e2e): capture ClusterIP dataplane and accept-queue diagnostics

jeffspahr:
## Problem

E2E lanes intermittently fail with bursts of:

```
dial tcp <clusterip>:<port>: i/o timeout
```

while pipelines communicate with in-cluster services. This fingerprint means the TCP connection to the **service ClusterIP** did not establish, while the backing pod's `httpGet /status` liveness probe—run by kubelet against the **pod IP**—can continue passing. The pod therefore remains `Ready` and is not restarted.

#13686 added SeaweedFS restart, pod, and node-contention diagnostics, but it did not inspect the ClusterIP datapath directly.

**This is not necessarily SeaweedFS-specific.** SeaweedFS (`:9000`) and MLflow (`:8443`) service VIPs have both timed out on the same node in the same E2E job, which makes a shared node-level datapath failure one candidate alongside service-specific backend saturation.

## Change

Extend `collect-seaweedfs-diagnostics.sh` with failure-only node and pod datapath probes:

- New `--dial-timeout-log <file>` support scans the already-collected pod logs for every `dial tcp <ip>:<port>: i/o timeout` VIP. Section (4) probes the SeaweedFS ClusterIP plus every discovered VIP, with no additional service names hardcoded.
- The script inspects **every Kind node**, because ClusterIP DNAT and conntrack processing occur on the client pod's node. Per node it captures:
  - the current conntrack count/max;
  - durable `drop` / `insert_failed` counters via `conntrack -S`, falling back to `/proc/net/stat/nf_conntrack`;
  - any `nf_conntrack: table full` kernel messages;
  - the service program for each VIP, followed end to end through KUBE-SERVICES → KUBE-SVC → KUBE-SEP → the DNAT target.
- Section (5) captures the SeaweedFS pod's listen sockets and cumulative `ListenOverflows` / `ListenDrops` counters. Nonzero counters would support accept-queue saturation as the cause of dropped SYNs while the pod remained healthy.

Each probe distinguishes an unavailable or forbidden query from a successful empty result. The collector remains best-effort: missing tools, non-Docker nodes, or failed diagnostic commands do not fail the caller. Passing runs are unchanged because these probes execute only in the existing failure-diagnostics path.

## Live result and remaining evidence gap

An earlier version of section (4) ran on a real SeaweedFS `:9000` dial-timeout failure ([job 86648705812](https://github.com/kubeflow/pipelines/actions/runs/29152867712/job/86648705812), E2EParallelNested v1.36.1, 64 dial timeouts). It found:

```
conntrack in-use / max: 1726/262144
kernel 'nf_conntrack: table full' events: (no table-full event logged)
service program for 10.96.25.77:9000: top-level KUBE-SERVICES rule present
```

That evidence made sustained conntrack exhaustion unlikely and confirmed that the top-level VIP rule existed, but it did **not** conclusively rule out either category:

- the point-in-time conntrack gauge may have drained, and the intended durable `/proc/net/stat/nf_conntrack` source was unavailable;
- the probe did not yet traverse KUBE-SVC/KUBE-SEP or verify the endpoint DNAT.

The current revision adds the missing durable `conntrack -S` counters, complete endpoint-chain traversal, and SeaweedFS accept-queue counters. Those refinements have been exercised with stubs but have not yet captured another real dial-timeout occurrence. This PR is therefore diagnostic instrumentation, not a root-cause fix.

## Verification

- `bash -n` clean.
- Stubbed runs exercised:
  - multi-VIP discovery and deduplication, plus SeaweedFS-only compatibility;
  - the multi-node loop;
  - KUBE-SERVICES → KUBE-SVC → KUBE-SEP → DNAT traversal;
  - `conntrack -S` present, absent, and error paths;
  - SeaweedFS listen-queue counters;
  - unavailable `dmesg` and missing rule-tool distinctions.

Follow-up to #13686.

### DIFF
--- a/.github/actions/test-and-report/action.yml
+++ b/.github/actions/test-and-report/action.yml
@@ -363,12 +363,15 @@ runs:
         if [ "${{ inputs.proxy }}" = "true" ]; then
           ./.github/resources/scripts/collect-logs.sh --ns tinyproxy --output /tmp/tmp_pod_log.txt
         fi
-        # Targeted SeaweedFS diagnostics: E2E lanes intermittently fail with
-        # dial-timeout errors to the SeaweedFS ClusterIP. This captures the
-        # restart/contention/dataplane evidence needed to tell whether a CPU
-        # request bump (which this workflow's manifests carry) is the right
-        # lever or the failure is elsewhere.
-        ./.github/resources/scripts/collect-seaweedfs-diagnostics.sh --namespace "$NAMESPACE" --output /tmp/tmp_pod_log.txt
+        # Dataplane diagnostics: E2E lanes intermittently fail with dial-timeout
+        # errors to in-cluster ClusterIPs (SeaweedFS :9000, the MLflow service
+        # VIP, ...). This captures SeaweedFS restart/contention evidence plus,
+        # per Kind node, the conntrack table state and the iptables/ipvs program
+        # for every ClusterIP the just-collected pod logs recorded a dial
+        # timeout against -- distinguishing conntrack exhaustion from a
+        # missing/stale service program. --dial-timeout-log points at the pod
+        # logs gathered above so the probe self-targets whichever VIP failed.
+        ./.github/resources/scripts/collect-seaweedfs-diagnostics.sh --namespace "$NAMESPACE" --dial-timeout-log /tmp/tmp_pod_log.txt --output /tmp/tmp_pod_log.txt
       continue-on-error: true
 
     - name: Set up Python for reporting
--- a/.github/resources/scripts/collect-seaweedfs-diagnostics.sh
+++ b/.github/resources/scripts/collect-seaweedfs-diagnostics.sh
@@ -19,8 +19,20 @@
 #   3. Cluster dataplane failure     -> SeaweedFS is Running with no restarts
 #                                        and the node is healthy, yet the
 #                                        ClusterIP path times out (kube-proxy /
-#                                        CNI). Captured by ruling out 1 and 2
-#                                        plus kube-proxy pod state.
+#                                        CNI). Ruled in by 1 and 2 being clean
+#                                        plus kube-proxy pod state, and probed
+#                                        directly in section (4).
+#
+# Section (4) inspects the Kind node's netfilter state for the failing
+# SeaweedFS ClusterIP. The leading dataplane candidate is conntrack-table
+# saturation: the nested-parallel lanes open a burst of simultaneous
+# connections, and once the node conntrack table fills, new SYNs are dropped
+# and every dial reports 'i/o timeout' while the pod's own liveness probe
+# (kubelet -> pod IP) keeps passing, so the pod is never restarted. The
+# durable evidence is the cumulative insert_failed/drop counters in
+# /proc/net/stat/nf_conntrack and any 'nf_conntrack: table full' dmesg line;
+# the presence or absence of iptables/ipvs rules for the ClusterIP separates
+# conntrack exhaustion from a missing/stale service program.
 #
 # All commands are best-effort; a missing object never fails the caller.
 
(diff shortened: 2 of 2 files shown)


jeffspahr:
Thanks — all three addressed in 391c23d1d:

1. **Multi-node coverage.** Rather than skip multi-node clusters, section (4) now iterates every node from `kubectl get nodes` and probes each. A ClusterIP dial is DNAT'd and conntrack-tracked on the *client* pod's node, so this now captures the node that actually matters even when it isn't the SeaweedFS node. Nodes that aren't Docker containers (non-Kind / remote runtime) skip individually.

2. **`dmesg` blank vs failed.** The query now captures `dmesg`'s exit status: a forbidden/failed `dmesg` reports `(dmesg unavailable — cannot confirm table-full events)`, and only a genuinely successful empty query reports `(no table-full event logged)`, so a blank line can no longer falsely rule out the signal.

3. **Rule absence vs tool absence.** The service-rule probe tracks whether any ruleset was read successfully. If neither `iptables-save` nor `ipvsadm` is available it reports `(iptables-save and ipvsadm both unavailable — cannot confirm rule presence)` instead of claiming no rule references the ClusterIP.

Verified with `bash -n` plus stubbed runs exercising the multi-node loop and each new distinction (dmesg forbidden, dmesg empty-success, both rule tools missing, rule present-but-absent).

jeffspahr:
Broadened scope in fa99544e0 (title/body updated): the node-dataplane probe now covers **every** timed-out ClusterIP, not just SeaweedFS.

Motivation: the same `dial tcp <vip>: i/o timeout` fingerprint was observed on two different services on the same node in the same E2E job — the SeaweedFS VIP (`:9000`) and the MLflow service VIP (`:8443`) — which points at a shared node-level dataplane failure rather than a SeaweedFS-specific one.

- New `--dial-timeout-log` scans the already-collected pod logs for each `dial tcp <ip>:<port>: i/o timeout` VIP; section (4) probes the SeaweedFS ClusterIP plus each of those (no service names hardcoded).
- Per node: node-global conntrack state (count/max, cumulative `insert_failed`/`drop`, `table full` dmesg) is dumped **once**; the iptables/ipvs program is dumped **per VIP**.
- `test-and-report` passes the pod-log file it already gathers, so the probe self-targets whichever service timed out (SeaweedFS, MLflow, …).

Verified with `bash -n` plus stubbed runs for the multi-VIP, back-compat (SeaweedFS-only without the flag), and multi-node paths.

jeffspahr:
### Live finding + refinements

An earlier version of section (4) ran on a real SeaweedFS `:9000` dial-timeout failure ([job 86648705812](https://github.com/kubeflow/pipelines/actions/runs/29152867712/job/86648705812), E2EParallelNested v1.36.1, 64 dial timeouts). It found:

```
conntrack in-use / max: 1726/262144
kernel 'nf_conntrack: table full' events: (no table-full event logged)
service program for 10.96.25.77:9000: top-level KUBE-SERVICES rule present
```

This **narrowed, but did not conclusively eliminate, the datapath hypotheses**:

- The low point-in-time conntrack gauge and clean `dmesg` made sustained table exhaustion unlikely, but the gauge may have drained after the burst and the intended durable `/proc/net/stat/nf_conntrack` source was unavailable.
- The top-level KUBE-SERVICES rule existed, but the probe did not yet follow KUBE-SVC → KUBE-SEP or verify the endpoint DNAT.

The current revision adds the evidence that was missing from that live run:

- **Durable conntrack counters** — prefer `conntrack -S` cumulative `drop` / `insert_failed` counters, falling back to `/proc/net/stat/nf_conntrack`.
- **Complete endpoint-chain traversal** — resolve the port-specific KUBE-SVC chain and dump its KUBE-SEP endpoint chain and DNAT target. This distinguishes a live backend from an empty or incomplete service program.
- **SeaweedFS accept-queue state** — capture `ss` / `netstat` backlog plus cumulative `ListenOverflows` / `ListenDrops` from the pod network namespace. Nonzero counters would support accept-queue saturation as the source of dropped SYNs while the pod remained healthy.

These refinements pass `bash -n` and stubbed coverage, but they have not yet captured another real dial-timeout occurrence. The PR remains diagnostic instrumentation rather than a root-cause fix.

The CI failures observed on this PR are the pre-existing dial-timeout class and a separate v1.36.1 multi-user flake, not failures introduced by the diagnostics.

Issue #13239, opened 2026-04-08
[sdk] dsl.ParallelFor name= breaks dsl.Collected at runtime


tom-pavz:
### Environment

* **KFP SDK version:** `kfp==2.14.3`, `kfp-kubernetes==2.14.3`
* **KFP backend:** Open-source Kubeflow Manifests on AWS EKS (Argo Workflows executor)
* **All dependencies version:**
```
kfp                                2.14.3
kfp-kubernetes                     2.14.3
kfp-pipeline-spec                  2.14.0
kfp-server-api                     2.16.0
```

### Summary

Passing `name=` to `dsl.ParallelFor` causes a runtime failure when downstream tasks consume `dsl.Collected` outputs from the loop. The KFP Argo driver cannot resolve the producer task because the compiled pipeline IR has inconsistent identifiers for the `ParallelFor` group.

### Steps to reproduce

1. Create a pipeline that uses `dsl.ParallelFor` with a custom `name=` parameter and `dsl.Collected` to fan-in outputs:

```python
from kfp import compiler, dsl

@dsl.component
def echo(item: str) -> str:
    return item

@dsl.component
def collect(items: list) -> str:
    return str(items)

@dsl.pipeline(name="repro-parallelfor-name-bug")
def my_pipeline():
    with dsl.ParallelFor(items=["a", "b", "c"], name="My Custom Loop") as item:
        work = echo(item=item)
    collect(items=dsl.Collected(work.output))

compiler.Compiler().compile(my_pipeline, "pipeline.yaml")
```

2. Upload and run the pipeline on a KFP deployment using the Argo Workflows executor.

3. The `collect` step fails at the driver level with:

```
F0408 16:10:52.114093  16 main.go:104] KFP driver: failed to resolve inputs:
  failed to resolve input artifact input_paths with spec
  task_output_artifact:{producer_task:"for-loop-2" output_artifact_key:"..."}: 
  cannot find producer task "for-loop-2_19426"
```

4. **Removing `name="My Custom Loop"` from the `dsl.ParallelFor` call makes the pipeline succeed.**

### Root cause

The compiler (`pipeline_spec_builder.build_task_spec_for_group`) sets `taskInfo.name` from the user-provided display name:

```python
# kfp/compiler/pipeline_spec_builder.py
pipeline_task_spec.task_info.name = group.display_name or group.name
```

[...]

PR #12808, opened 2026-02-13
fix(backend): Fix mismatched task name in ParallelFor - Fixes #11964

cbartram:
**Description of your changes:**

Fixes a mismatched task name in ParallelFor sub tasks when the `name` parameter is provided to the `ParallelFor`. 
More details about the error can be seen in #11964.

In the UI this manifested itself as not being able to see the individual tasks within a subdag (see before and after screenshots). I used the pipeline in #11964 as a test to verify the fix.

## Before
kubeflow-before

## After
kf-after

**Checklist:**
- [x] You have [signed off your commits](https://www.kubeflow.org/docs/about/contributing/#sign-off-your-commits)
- [x] The title for your pull request (PR) should follow our title convention. [Learn more about the pull request title convention used in this repository](https://github.com/kubeflow/pipelines/blob/master/CONTRIBUTING.md#pull-request-title-convention).

PR #13707, opened 2026-07-11
fix(sdk): align parallelfor taskInfo name with system key

modichika:

this PR fixes the issue https://github.com/kubeflow/pipelines/issues/13239 -  runtime bug where dsl.ParallelFor with a custom name brreaks downstream dsl.Collected output resolution at runtime.

to ensure output resolution works, taskInfo.name is now aligned to the internal system ID (e.g., for-loop-2), addressed the sibling issue for same loops names.

### open discussion: UI display name

by aligning taskInfo.name to the system key, the KFP dashboard will visually fall back to  showing for-loop-2 as the node label and in the MLMD execution panel.
if we want to reflect the user's custom display_name in the UI, we have a few options:

1. Modify the frontend graph rendering (StaticFlow.ts) to parse the custom display name from the component reference name ( componentRef.name ), keeping the status tracking mapped to the system ID.
2. Modify the frontend's MLMD lookup logic in the run graph to map execution statuses using the task's compiled display name rather than the system ID.

please let me know how the maintainers will prefer to proceed with the UI display name aspect!

**Checklist:**


zazulam:
@modichika can you please add some test cases to verify the changes and share a sample of a generated IR showcasing the changes the compiler would have.

modichika:
Hey @zazulam added a test case and it's output where the taskinfo.name is same as the internal system ID - "for-loop-N"

zazulam (commented):
I think we should still address two gaps before merge:
1. `parallelfor_naming_test.py` isn’t an actual test yet, the file has no  test_  functions/assertions, and all logic is under  `if __name__ == "__main__"`:  (which pytest won’t run). So this currently provides no regression protection. 
2.` pipeline_named.yaml` appears to be generated output, it hardcodes compiler output ( sdk version, executor command content) and can drift over time; the script also writes/overwrites it at runtime.

* Convert this into proper pytest tests that compile to an in-memory target (e.g.,  io.StringIO) and assert the name. You can refer to other compiler tests or even just add your tests into one of the existing files.

Also, I think we should ensure that the `display_name` the user passes is preserved for the UI but not used for the backend i/o resolution.

modichika:
So for backend I/O resolution we need the "for-loop-N" right?

zazulam:
Thanks for iterating on this. The runtime bug should be fixed without dropping the user-provided `ParallelFor` display name.

The existing component-task implementation is the model to follow:

- The SDK stores the user-facing label in [`task_info.name`](https://github.com/kubeflow/pipelines/blob/master/sdk/python/kfp/compiler/pipeline_spec_builder.py#L119-L126).
- The Argo compiler passes the stable task key separately through `taskName` / `Options.TaskName` ([DAG task path](https://github.com/kubeflow/pipelines/blob/master/backend/src/v2/compiler/argocompiler/dag.go#L248-L262), [container driver](https://github.com/kubeflow/pipelines/blob/master/backend/src/v2/driver/container.go#L115-L130)).
- The driver publishes the stable key as `ecfg.TaskName` and the user-facing label as `ecfg.DisplayName`.

Please make `ParallelFor` use that same identity/display-name separation rather than replacing `task_info.name` with `for-loop-N`.

Recommended approach:

1. **Reproduce against current `master` first.** The issue reports backend 2.14.3, while #12808 already added `ParallelFor` task-key propagation. Current code passes `parallelForTaskKey` into both the loop DAG driver and iteration tasks in [`iterationItemTask`](https://github.com/kubeflow/pipelines/blob/master/backend/src/v2/compiler/argocompiler/dag.go#L410-L443).
2. **If current `master` works**, this SDK change is unnecessary; resolve the affected release by documenting or backporting #12808.
3. **If current `master` still fails**, trace where the outer DAG-map key (`for-loop-N`) is lost between `iteratorTask`/`iterationItemTask`, `dagDriverInputs.taskName`, the generated `--task_name` argument, and `Options.TaskName`. Fix that propagation point. Keep the existing driver behavior in [`backend/src/v2/driver/dag.go`](https://github.com/kubeflow/pipelines/blob/master/backend/src/v2/driver/dag.go#L97-L106): stable identity from `opts.TaskName`, display label from `task_info.name`.
4. **Preserve the SDK contract**:
   ```python
   pipeline_task_spec.task_info.name = group.display_name or group.name
   ```
5. **Add cross-layer regression coverage**:
   - SDK: DAG key and collected `producer_task` are `for-loop-2`; `task_info.name` is `My custom Loop`.
   - Argo compiler: generated `task-name` is `for-loop-2`; serialized task spec retains `My custom Loop`.
   - Driver: MLMD `TaskName` is `for-loop-2`; `DisplayName` is `My custom Loop`.
6. **Remove `group_name_to_allocated_component_name` unless independently required.** If it fixes a separate component-name collision, isolate it with its own test/PR.

The key invariant is: internal references and MLMD lookup use the stable DAG task key, while UI and execution metadata retain the user-provided display name.

modichika:
Hey @zazulam I tested the reproduction case against current master and it compiles successfully. The collect task correctly references producerTask: for-loop-2, and no runtime errors occur. 

1. ✅ The pipeline compiles successfully (no immediate crash)
2. ✅ The collect task correctly references `producerTask: for-loop-2`
3. ❌ **The user's custom name `"My Custom Loop"` is missing from the compiled IR**
   - Expected: `taskInfo.name: My Custom Loop`
   - Actual: `taskInfo.name: for-loop-2`

So the PR prevents the runtime crash, but aligns with your concern about "fixing without dropping the display name."

Should we change the backend to see if we can:
- Keep `taskInfo.name = "My Custom Loop"`
- Pass the stable `for-loop-2` key separately so the driver can do MLMD lookup?

Or is there a reason the current approach (replacing display name with system ID) is necessary?
Attaching a ss to address this.

image

zazulam:
> Should we change the backend to see if we can:
> Keep taskInfo.name = "My Custom Loop"
> Pass the stable for-loop-2 key separately so the driver can do MLMD lookup?

@modichika yes, you can refer to the argocompiler for the container.go and dag.go files to see how the taskName is being passed from the IR and set on the Argo Workflow.

You can see from #12061 how it was handled:

-  [argocompiler/dag.go](https://github.com/kubeflow/pipelines/pull/12061/changes#diff-25b4791570bdc6ec4b4ca6c22a3eeba5ca4bd224870973a64dddb707f5aed993)
- [argocompiler/container.go](https://github.com/kubeflow/pipelines/pull/12061/changes#diff-c787345d412d375985c3fec92bc324fbcdfbcdcc793d0043bb90865bde86db19)
- [driver/container.go](https://github.com/kubeflow/pipelines/pull/12061/changes#diff-4d44a8b061b551e8a31db2c80e7033daf133e7ba9f806b20b2130f7ecdbddac1)

modichika:
Hey @zazulam  I think the PR: https://github.com/kubeflow/pipelines/pull/12808  

1. has fixed the issue of name mismatch as you can see in the images below the DAG key (`for-loop-2`) separates from display name (`My Custom Loop`) in the UI.

2. Though i added a regression test that tests the taskInfo.name to be display name and the downstream `dsl.Collected` to "for-loop-2"


Screenshot 2026-08-27 191634
Screenshot 2026-08-27 191704

zazulam (commented):
/lgtm
/approve

PR #13474, opened 2026-06-02
feat(backend): enable create and schedule time placeholders

droctothorpe:
## Summary

As currently opened, this PR contains two sets of changes:

1. **Backend fix for #13424**
   - Populates `dsl.PIPELINE_JOB_CREATE_TIME_UTC_PLACEHOLDER` and `dsl.PIPELINE_JOB_SCHEDULE_TIME_UTC_PLACEHOLDER` in the Kubernetes-backed v2 runtime.
   - Passes `{{workflow.creationTimestamp}}` into the DAG and container drivers as `--pipeline_job_create_time_utc`.
   - Resolves schedule time from Argo Workflow metadata at runtime when needed, using `scheduledworkflows.kubeflow.org/workflowEpoch` when present and falling back to workflow creation time for manual runs.
   - Limits Workflow metadata reads to DAG/container tasks that actually use these placeholders. Root DAG drivers and tasks that do not use the placeholders skip the lookup entirely.
   - Keeps schedule-time lookup best-effort when create time is already available, so reduced-RBAC clusters fall back to create time instead of failing.
   - Avoids regressing older compiled workflows that do not include the new create-time arg unless they actually rely on these placeholders.

2. **Release process docs / automation updates already in the branch**
   - Refreshes `RELEASE.md`.
   - Adds `gh` CLI examples for running release workflows and cutting releases.
   - Updates `test/release/release.sh` to bump the public version ConfigMap during release prep.

## Validation

- `go test ./backend/src/v2/compiler/argocompiler ./backend/src/v2/cmd/driver ./backend/src/v2/driver`
- Compiler goldens refreshed under `test_data/compiled-workflows/`

## Notes

- The backend placeholder change is scoped to the Kubernetes-backed v2 runtime. It does not change local execution.
- Exact recurring-run schedule time still depends on Workflow metadata availability. When the metadata cannot be read after create time is already known, the driver falls back to create time.

## Fixes

Fixes #13424.

PR #13711, opened 2026-07-13
chore(backend): validate required driver and launcher flags

### LINKED ISSUE #12725: chore(<backend>): Refactor Driver and Launcher: Make All Parameters Required
## Chore description

Currently, the driver and launcher receive several parameters from the API server (and the launcher receives from the driver too). Most of these parameters are optional from the driver/launcher perspective when they shouldn't be. This is:

- **Error-prone**: Silent failures when the API server doesn't pass required values
- **Hard to catch in tests**: Tests may pass when they shouldn't because default values mask developer mistakes where a value should have been explicitly provided
- **Difficult to develop**: New developers can't easily tell what's truly required vs. optional
- **Hard to debug**: When something fails, it's unclear if a value was intentionally empty or missing
- **Hard to make changes**: Implicit contracts between components make refactoring risky since dependencies on defaults are not visible

We should not rely on defaults for driver and launcher parameters.

### Driver Flags

The driver defines ~30 flags, most with default values (`""`, `0`, `false`, `"{}"`):

```go
// backend/src/v2/cmd/driver/main.go (lines 55-94)
driverType        = flag.String(driverTypeArg, "", "task driver type, one of ROOT_DAG, DAG, CONTAINER")
pipelineName      = flag.String("pipeline_name", "", "pipeline context name")
runID             = flag.String("run_id", "", "pipeline run uid")
runName           = flag.String("run_name", "", "pipeline run name (Kubernetes object name)")
componentSpecJson = flag.String("component", "{}", "component spec")
taskSpecJson      = flag.String("task", "", "task spec")
// ... and many more
```

### Driver Validation

The current `validate()` function only checks 4 flags out of ~30:

```go
// backend/src/v2/cmd/driver/main.go (lines 121-136)
func validate() error {
    if *driverType == "" {
        return fmt.Errorf("argument --%s must be specified", driverTypeArg)
    }
    if *httpProxy == unsetProxyArgValue {
        return fmt.Errorf("argument --%s is required but can be an empty value", httpProxyArg)
    }
    // ... only proxy flags validated
    // validation responsibility lives in driver itself, so we do not validate all other args
    return nil
}
```

### Monolithic Options Struct

The `driver.Options` struct has comments saying "required" or "optional, required only by..." but these are documentation-only with no enforcement:

```go
// backend/src/v2/driver/driver.go (lines 36-94)
type Options struct {
    // required, pipeline context name
    PipelineName string
    // required, KFP run ID
    RunID string
    // required, Component spec
    Component *pipelinespec.ComponentSpec
    // optional, iteration index. -1 means not an iteration.
    IterationIndex int

    // optional, required only by root DAG driver
    RuntimeConfig *pipelinespec.PipelineJob_RuntimeConfig
    Namespace     string

    // optional, required by non-root drivers
    Task           *pipelinespec.PipelineTaskSpec
    DAGExecutionID int64

    // optional, required only by container driver
    Container *pipelinespec.PipelineDeploymentConfig_PipelineContainerSpec
    // ... more fields
}
```

### Launcher Flags

The launcher has similar issues with ~20 flags and no validation:

```go
// backend/src/v2/cmd/launcher-v2/main.go (lines 31-52)
pipelineName            = flag.String("pipeline_name", "", "pipeline context name")
runID                   = flag.String("run_id", "", "pipeline run uid")
parentDagID             = flag.Int64("parent_dag_id", 0, "parent DAG execution ID")
executorType            = flag.String("executor_type", "container", "The type of the ExecutorSpec")
executionID             = flag.Int64("execution_id", 0, "Execution ID of this task.")
// ... no validation function exists
```

### 1. Use `flag.Visit()` to Detect Missing Flags

Go's `flag.Visit()` function tells us which flags were actually provided on the command line, regardless of their value. This distinguishes between "not provided" (uses default) and "provided as zero/empty" (explicit value).

```go
flag.Parse()

// Collect which flags were actually passed
providedFlags := make(map[string]bool)
flag.Visit(func(f *flag.Flag) {
    providedFlags[f.Name] = true
})
```

### 2. Add Validation for Required Flags

Validate that all required flags for the driver type were provided:

```go
func validateRequiredFlags(provided map[string]bool, driverType string) error {
    // Common flags required by all driver types
    common := []string{"type", "pipeline_name", "run_id", "component", "namespace",
        "ml_pipeline_server_address", "ml_pipeline_server_port",
        "mlmd_server_address", "mlmd_server_port",
        "http_proxy", "https_proxy", "no_proxy",
        "publish_logs", "cache_disabled", "log_level"}
    
    // Driver-type specific required flags
    var specific []string
    switch driverType {
    case "ROOT_DAG":
        specific = []string{"runtime_config", "run_name", "run_display_name"}
    case "DAG":
        specific = []string{"task", "dag_execution_id", "iteration_index", "task_name"}
    case "CONTAINER":
        specific = []string{"task", "dag_execution_id", "container", "kubernetes_config",
            "iteration_index", "task_name", "run_name", "run_display_name"}
    }
    
    required := append(common, specific...)
    for _, name := range required {
        if !provided[name] {
            return fmt.Errorf("--%s is required for %s but was not provided", name, driverType)
        }
    }
    return nil
}
```

### 3. Compiler Always Passes All Flags

The compiler (argocompiler) must always include ALL flags for the driver type being invoked. No conditional inclusion - every flag for that driver type is always passed.

```go
// backend/src/v2/compiler/argocompiler/container.go
// Always include ALL flags for CONTAINER driver - no conditionals
args := []string{
    "--type", "CONTAINER",
    "--pipeline_name", c.spec.GetPipelineInfo().GetName(),
    "--run_id", runID(),
    "--run_name", runResourceName(),
    "--run_display_name", c.job.DisplayName,
    "--dag_execution_id", inputValue(paramParentDagID),
    "--task", inputValue(paramTask),
    "--component", inputValue(paramComponent),
    "--container", inputValue(paramContainer),
    "--kubernetes_config", inputValue(paramKubernetesConfig),
    "--iteration_index", inputValue(paramIterationIndex),
    "--task_name", inputValue(paramTaskName),
    // ... ALL other required flags
}
```

### 4. Apply Same Pattern to Launcher

The launcher validates its required flags using the same `flag.Visit()` approach:

```go
func validateLauncherFlags(provided map[string]bool, executorType string) error {
    common := []string{"pipeline_name", "run_id", "pod_name", "pod_uid",
        "ml_pipeline_server_address", "ml_pipeline_server_port",
        "mlmd_server_address", "mlmd_server_port", "executor_type"}
    
    var specific []string
    switch executorType {
    case "container":
        specific = []string{"execution_id", "executor_input", "component_spec"}
    case "importer":
        specific = []string{"component_spec", "importer_spec", "task_spec", "parent_dag_id"}
    }
    
    required := append(common, specific...)
    for _, name := range required {
        if !provided[name] {
            return fmt.Errorf("--%s is required for %s executor but was not provided", name, executorType)
        }
    }
    return nil
}
```

## Benefits

- **Fail-fast**: Missing parameters are caught immediately with clear error messages
- **Explicit contracts**: The list of required flags per driver/executor type is explicit in code
- **Testability**: Validation logic can be unit tested independently
- **Developer experience**: Clear errors tell you exactly which flag is missing
- **Minimal structural changes**: Keep existing structs and functions, just add validation

## Risks

- **Existing failures may be uncovered**: Making all parameters required may reveal places where the code was silently relying on defaults when a value should have been explicitly provided. This could expose latent bugs in the compiler or API server that were previously masked by default values.
- **Requires fixes in upstream code**: If validation failures are discovered, the compiler (`argocompiler`) may need to be updated to explicitly pass values that were previously omitted.

### Driver

- `backend/src/v2/cmd/driver/main.go` - Add `flag.Visit()` tracking, add `validateRequiredFlags()`

### Launcher

- `backend/src/v2/cmd/launcher-v2/main.go` - Add `flag.Visit()` tracking, add `validateLauncherFlags()`

### Compiler (must pass all flags)

- `backend/src/v2/compiler/argocompiler/dag.go` - Ensure all ROOT_DAG and DAG flags are passed
- `backend/src/v2/compiler/argocompiler/container.go` - Ensure all CONTAINER flags are passed
- `backend/src/v2/compiler/argocompiler/importer.go` - Ensure all importer launcher flags are passed

## Future Improvements

Once validation is in place, further improvements could include:

- Split the monolithic `Options` struct into driver-specific config structs (adds compile-time type safety)
- Use constructor functions with private fields to make invalid states unrepresentable

## Acceptance Criteria

- [ ] Driver validates all required flags were provided using `flag.Visit()`
- [ ] Launcher validates all required flags were provided using `flag.Visit()`
- [ ] Missing required parameters result in clear, actionable error messages (e.g., "--dag_execution_id is required for CONTAINER but was not provided")
- [ ] Compiler always passes ALL flags for each driver type (no conditional inclusion)
- [ ] Required flags are documented per driver type (ROOT_DAG, DAG, CONTAINER)
- [ ] Required flags are documented per executor type (container, importer)
- [ ] Unit tests cover validation for each driver and executor type
- [ ] Existing integration and E2E tests pass

### Labels

/area backend

---

Love this idea? Give it a 👍.

Priyanshu-u07:
 Today the driver and launcher mostly rely on default values, so a missing value from the compiler silently becomes a wrong default and fails confusingly later. This PR validates that every flag required for a given driver/executor type was actually provided, and errors immediately with a clear message otherwise.

Fixes #12725 

## Approach
Type-specific flags stay optional at the flag level and are validated per-type at runtime once the driver/executor type is known,  so the compiler never emits zero/empty placeholders. Uses `flag.Visit()` which reports only the flags actually set on the command line, distinguishing "not provided" (falls back to default) from "provided but empty". Required-flag lists are explicit and per-type.

## Change

- **Driver** (`cmd/driver/main.go`): `flag.Visit()` tracking and `validateRequiredFlags()` with required-flag lists per driver type    (`ROOT_DAG`, `DAG`, `CONTAINER`).
- **Launcher** (`cmd/launcher-v2/main.go`): same pattern via `validateLauncherFlags()` per executor type (`container`, `importer`).
- **Compiler** (`dag.go`, `container.go`, `importer.go`): always emit the previously-conditional flags `cache_disabled`, `ml_pipeline_tls_enabled`, `metadata_tls_enabled`, `ca_cert_path`, `log_level`, `publish_logs` as explicit values, so each can be required.
- Removed `pipeline_job_schedule_time_epoch_seconds`: never passed on the command line (resolved at runtime from the `workflowEpoch` label / create-time fallback). Dropped the flag registration, the `scheduleTimeEpochSeconds` params in `resolvePipelineJobTimes`/`getWorkflowMetadataForPipelineJobTimes`, the dead `ParseInt` branch and its tests.
- **Driver to launcher** (`driver/driver.go`):  pass `--executor_type container` explicitly; always emit `log_level`/`cache_disabled`/TLS/`ca_cert_path`; fixed a duplicate `--publish_logs`.
- **Test helper:** `ConfigureCacheSettings` now rewrites the always-present `--cache_disabled=<bool>` instead of adding/removing a bare flag. 
- **Goldens:** regenerated `test_data/compiled-workflows/` (146 files) to reflect the new driver args 
- Unit tests for validation of every driver and executor type.

## Note

- `default_run_as_user`/`group`/`non_root` and `default_host_users` kept optional as admin defaults are emitted only when configured, and a missing one is indistinguishable from "not set", so requiring them adds no safety and would mean emitting `-1`/`""` placeholders everywhere.
- Proxy flags are validated via their existing `"unset"` sentinel (required, may be empty).
- `namespace` excluded, not a flag; resolved at runtime via `config.InPodNamespace()`.


hbelmiro (changes requested):
The PR's stated goal is to make the apiserver→driver and driver→launcher contracts fail-fast instead of silently falling back to defaults. But a large number of flags remain optional — unvalidated and relying on exactly the default-fallback behavior this PR is meant to prevent. If one of these flags is accidentally dropped from the compiler or driver, the binary will silently use a wrong default and fail confusingly later — the scenario described in the PR description and in issue #12725.

**Driver — optional flags (not in any required list, not proxy-validated):**

| Flag                                       | Default | Emitted by compiler                                                                                                                 |
|--------------------------------------------|---------|-------------------------------------------------------------------------------------------------------------------------------------|
| `cache_disabled`                           | `false` | conditionally (`if c.cacheDisabled`)                                                                                                |
| `ml_pipeline_tls_enabled`                  | `false` | conditionally                                                                                                                       |
| `metadata_tls_enabled`                     | `false` | conditionally                                                                                                                       |
| `ca_cert_path`                             | `""`    | conditionally                                                                                                                       |
| `default_run_as_user`                      | `-1`    | conditionally                                                                                                                       |
| `default_run_as_group`                     | `-1`    | conditionally                                                                                                                       |
| `default_run_as_non_root`                  | `""`    | conditionally                                                                                                                       |
| `default_host_users`                       | `""`    | conditionally                                                                                                                       |
| `pipeline_job_create_time_utc`             | `""`    | always — compiler emits `{{workflow.creationTimestamp}}` (Argo resolves it); could be required with zero behavior change            |
| `pipeline_job_schedule_time_epoch_seconds` | `""`    | never — resolved at runtime from Workflow `workflowEpoch` label or fallback to create time; should not be a flag at all (see below) |

**Launcher — optional flags (not in any required list):**

| Flag                      | Default  | Emitted by driver/compiler                                                                                  |
|---------------------------|----------|-------------------------------------------------------------------------------------------------------------|
| `log_level`               | `"1"`    | driver: only when `!= "1"` ; importer template: conditionally via `os.LookupEnv`                            |
| `publish_logs`            | `"true"` | driver: always + duplicated conditionally (see below) ; importer template: conditionally via `os.LookupEnv` |
| `cache_disabled`          | `false`  | conditionally                                                                                               |
| `ca_cert_path`            | `""`     | conditionally                                                                                               |
| `ml_pipeline_tls_enabled` | `false`  | conditionally                                                                                               |
| `metadata_tls_enabled`    | `false`  | conditionally                                                                                               |

The PR note says these are excluded because "a boolean whose `false` default is a meaningful state, not a missing value." But that argument applies equally to `log_level` (default `"1"`) and `publish_logs` (default `"true"`), which the driver *does* require. The distinction is arbitrary — the real value of validation is catching accidental omissions by the compiler/driver, not whether the default happens to be a no-op.

For all boolean/toggle flags (`cache_disabled`, TLS, `default_run_as_*`), the compiler always knows the intended value — it just conditionally emits the flag. Having the compiler always emit these flags explicitly (e.g., `--cache_disabled=false`) and requiring them in validation would make the contract complete. The compiler changes are mechanical: replace conditional `append` with unconditional `append` using the value.

Of the two time flags, `pipeline_job_create_time_utc` is already always emitted by the compiler (as `{{workflow.creationTimestamp}}`, which Argo resolves at runtime) — making it required would have zero behavior change. `pipeline_job_schedule_time_epoch_seconds` is never passed on the command line — no caller emits it. The driver resolves it entirely at runtime from the Workflow's `workflowEpoch` label (recurring runs) or falls back to create time (manual runs). It shouldn't be a `flag.String` at all; it's internal driver state, not a CLI contract. Defining it as a flag is misleading — it suggests the compiler or some caller is expected to provide it.

Instead of maintaining per-type required-flag lists, consider a single function that validates every registered flag was explicitly set:

```go
// validateAllFlagsSet ensures every registered flag was explicitly passed on
// the command line. No flag may rely on its default — the caller (compiler or
// driver) must always emit it.
func validateAllFlagsSet(provided map[string]bool) error {
    var missing []string
    flag.VisitAll(func(f *flag.Flag) {
        if !provided[f.Name] {
            missing = append(missing, f.Name)
        }
    })
    if len(missing) > 0 {
        return fmt.Errorf("missing required flags: %v", missing)
    }
    return nil
}
```

At the call site, add a comment reinforcing the policy:

```go
// Every flag this binary registers must be explicitly provided by the caller
// (compiler or driver). No flag may silently fall back to its default.
if err := validateAllFlagsSet(providedFlags); err != nil {
    return err
}
```

This eliminates the manually curated required-flag lists — any new flag is automatically validated. Apply the same pattern to both driver and launcher.

**Pre-existing bug — duplicate `--publish_logs` in driver→launcher command (`driver.go`):**

The driver constructs the launcher command with `--publish_logs` always included in the base args (line 338), then appends it again conditionally when `publishLogs == "true"` (lines 356-358). When the value is `"true"`, the launcher receives `--publish_logs true --publish_logs true`. Go's `flag` package takes the last value so it works, but it's a bug that should be cleaned up — especially since this PR already touches the launcher command to add `--executor_type`.

Priyanshu-u07:
Thanks for the feedback @hbelmiro! I agree on the main point. The goal was "no flag silently falls back to a default," and I only enforced that for a subset so the exact failure mode from the PR is still possible for the flags I left optional. And the `cache_disabled` vs `log_level/publish_logs` split is inconsistent as the "meaningful default" isn't the right criterion; catching accidental omissions is.

I'd like to go with the `validateAllFlagsSet` (`flag.VisitAll`) approach in both driver and launcher and make the compiler always emit the currently-conditional flags (`cache_disabled`, the TLS toggles, `ca_cert_path`, `default_run_as_*`) so they can be required, as you described. 

**One thing I'd like to flag here:**
`pipeline_job_schedule_time_epoch_seconds` is never passed and it's resolved at runtime from the `workflowEpoch` label and it's documented in #13474 as an intentional optional input ([resolvePipelineJobTimes](https://github.com/kubeflow/pipelines/blob/3460971aa5df4213b02482e1e3b4c6120529e473/backend/src/v2/cmd/driver/main.go#L287-L290)): "Schedule time may come from the compiled flag when explicitly provided, or from workflow metadata...". Under a strict "every flag required" check it would fail on every run.

I'm thinking of adding a small, documented `optionalFlags` allowlist, so the default stays "required" for everything else (new flags auto-validated) but genuinely-optional ones like this are exempted with a written reason. 
Does that sound right to you or would you rather handle `schedule_time` a different way (dropping the flag)?

hbelmiro:
No allowlist — that reintroduces the "some flags are optional" pattern we're trying to eliminate.

`pipeline_job_schedule_time_epoch_seconds` is never passed on the command line — no caller emits it, the compiler tests assert `NotContains` for it, and no golden files reference it. It shouldn't be a flag. Remove the `flag.String` registration, drop the `scheduleTimeEpochSeconds` parameter from `resolvePipelineJobTimes` and `getWorkflowMetadataForPipelineJobTimes` (it's always `""`), and clean up the dead `ParseInt` branch and its test cases. Then `validateAllFlagsSet` works with zero exceptions.

Priyanshu-u07:
Got it. One check before the compiler work: removing `schedule_time` alone won't reach zero exceptions because the type-specific flags (`container`/`kubernetes_config` are CONTAINER-only, `runtime_config` and the DAG output paths are DAG-only, the container-vs-importer launcher flags differ) and `--copy` are also never passed on every run. 
To make `validateAllFlagsSet` pass, the compiler has to emit all of them as placeholders everywhere e.g. the DAG driver emitting `--container "{}"` `--runtime_config ""`. Happy to build it that way, just wanted to confirm first.

hbelmiro:
> Got it. One check before the compiler work: removing `schedule_time` alone won't reach zero exceptions because the type-specific flags (`container`/`kubernetes_config` are CONTAINER-only, `runtime_config` and the DAG output paths are DAG-only, the container-vs-importer launcher flags differ) and `--copy` are also never passed on every run. To make `validateAllFlagsSet` pass, the compiler has to emit all of them as placeholders everywhere e.g. the DAG driver emitting `--container "{}"` `--runtime_config ""`. Happy to build it that way, just wanted to confirm first.

@Priyanshu-u07 In that case, they should be optional from Viper's point of view. We should validate them during runtime immediately once we know the type (CONTAINER/DAG). That way the compiler doesn't need to set zero/empty values.

Priyanshu-u07:
@hbelmiro, I've reworked it to validate per-type at runtime (no zero/empty placeholders), removed the `schedule_time` flag, fixed the duplicate `--publish_logs` and regenerated the goldens. Details in the updated PR description.
One thing to confirm: I kept `default_run_as_*` optional (a missing admin default is indistinguishable from "not set", so requiring it would just mean `-1`/`""` placeholders). Open to changing it if you'd prefer. PTAL,thanks!

hbelmiro:
Keeping `default_run_as_*` optional is fine (please add a comment saying that it's intentionally optional), but you can drop the `-1`/`""` sentinels. You already have `flag.Visit()` tracking which flags were provided — use that instead:

```go
if providedFlags["default_run_as_user"] {
    options.DefaultRunAsUser = defaultRunAsUser
}
```

The `Options` struct already uses pointers (`*int64`, `*bool`) where nil means not set. Let `flag.Visit()` be the authority, not the sentinel value.

backend/src/v2/compiler/argocompiler/importer.go
```diff
@@ -86,29 +87,30 @@ func (c *workflowCompiler) addImporterTemplate(downloadToWorkspace bool) string
...
 	setCABundle := false
-	// If CABUNDLE_SECRET_NAME or CABUNDLE_CONFIGMAP_NAME is set, add the custom CA bundle to the importer.
 	if common.GetCaBundleSecretName() != "" || common.GetCaBundleConfigMapName() != "" {
-		args = append(args, "--ca_cert_path", common.CustomCaCertPath)
+		caCertPath = common.CustomCaCertPath
 		setCABundle = true
 	}
+	args = append(args, "--ca_cert_path", caCertPath)
 
+	logLevel := "1"
```
hbelmiro: `importer.go` inlines the env-lookup logic for `log_level` and `publish_logs` (lines 103-113) instead of calling the `pipelineLogLevelArg()` and `publishLogsArg()` helpers already defined in `container.go`. Both `dag.go` and `container.go` use these helpers; `importer.go` is the only file that reimplements the same pattern.

**Suggested fix:** Replace the inline lookups with the existing helpers:

```go
args = append(args, "--log_level", pipelineLogLevelArg(), "--publish_logs", publishLogsArg())
```

This removes the local `logLevel` and `publishLogs` variables and the two `os.LookupEnv` blocks entirely (lines 103-113).

backend/src/v2/driver/driver.go
hbelmiro: `commonRequiredDriverFlags` includes `"run_name"` and `"run_display_name"`, but the `Options` struct still documents them as optional (lines 62-65). The compiler always emits them, so the validation is correct — update these comments to remove the "optional" qualifier.

backend/src/v2/cmd/driver/main.go
hbelmiro: `defaultRunAsNonRoot` silently swallows `strconv.ParseBool` errors (line 528: `if err == nil`), while `defaultHostUsers` correctly returns the error (line 535: `return fmt.Errorf(...)`). If the admin sets `--default_run_as_non_root=notabool`, it is silently ignored rather than failing fast — the opposite of the PR's stated goal.

This is pre-existing (not introduced by this PR), but it directly contradicts the fail-fast principle being established here. Consider fixing it in this PR since the surrounding code is already being touched.

**Suggested fix:**

```go
if *defaultRunAsNonRoot != "" {
    v, err := strconv.ParseBool(*defaultRunAsNonRoot)
    if err != nil {
        return fmt.Errorf("invalid --default_run_as_non_root value %q: %w", *defaultRunAsNonRoot, err)
    }
    options.DefaultRunAsNonRoot = &v
}
```

hbelmiro: The proxy flags (`http_proxy`, `https_proxy`, `no_proxy`) use an `"unset"` sentinel to distinguish "not provided" from "provided as empty." Now that `flag.Visit()` is in place, the sentinel is unnecessary — `flag.Visit()` already makes that distinction.

**Suggested fix:** Add `"http_proxy"`, `"https_proxy"`, `"no_proxy"` to `commonRequiredDriverFlags`, change the default from `"unset"` to `""`, and remove the three sentinel checks from `validate()`.

hbelmiro: `iteration_index` defaults to `-1` (meaning "not an iteration") and is required for DAG and CONTAINER but not ROOT_DAG. The compiler already passes `-1` explicitly for non-iteration tasks in DAG/CONTAINER types, so the default is never reached for those. Move `iteration_index` from `nonRootRequiredDriverFlags` to `commonRequiredDriverFlags` and have the compiler pass `--iteration_index -1` for ROOT_DAG too. That makes `-1` purely a domain value passed by the caller, with no reliance on the default.

backend/src/v2/compiler/argocompiler/importer.go
Priyanshu-u07: Done

backend/src/v2/driver/driver.go
Priyanshu-u07: Done

backend/src/v2/cmd/driver/main.go
Priyanshu-u07: Done

Priyanshu-u07: Done. Also switched the proxy entries to string literals (`"http_proxy"`, etc.) to keep the required-flags list consistent.

Priyanshu-u07: Done. Moved `iteration_index` to `commonRequiredDriverFlags`. No compiler change was needed for ROOT_DAG: the shared `system-dag-driver` template (which the ROOT_DAG driver uses) already emits `--iteration_index` unconditionally, resolving to -1 via the input default. So it's always passed on the command line, with no reliance on the flag's own default.

Priyanshu-u07:
Thanks for the review @hbelmiro. I've addressed all the review comments. PTAL when you get a chance.
Also, one extra change: I sorted `dependentTasks` in `addImplicitDependencies`. It was built from Go map iteration, so the compiled `depends:` order was nondeterministic and caused a flaky golden. It's a set, so order is semantically irrelevant.

backend/src/v2/compiler/argocompiler/container.go
hbelmiro: `addContainerExecutorTemplate()` passes `--cache_disabled`, `--log_level`, and `--publish_logs` to the init container (lines 459-465), but `--copy` causes the launcher to exit immediately without reading any other flags. These are dead args — the init container only copies the binary.

This is pre-existing (master already had these flags conditionally added), not introduced by this PR. But since the surrounding code is being touched and the PR is establishing explicit contracts, this is a good opportunity to clean it up.

**Suggested fix:**

```go
args := []string{
    "--copy", component.KFPLauncherPath,
}
```

This makes it obvious the init container only copies the binary and removes the inline `os.LookupEnv` calls that are inconsistent with the `pipelineLogLevelArg()`/`publishLogsArg()` helpers used everywhere else in the file.

backend/src/v2/cmd/driver/main.go
```diff
@@ -135,20 +134,103 @@ func init() {
...
-		return fmt.Errorf("argument --%s is required but can be an empty value", httpsProxyArg)
-	}
-	if *noProxy == unsetProxyArgValue {
-		return fmt.Errorf("argument --%s is required but can be an empty value", noProxyArg)
+	if err := validateRequiredFlags(providedFlags, *driverType); err != nil {
+		return err
 	}
-	// validation responsibility lives in driver itself, so we do not validate all other args
 	return nil
 }
```
hbelmiro: `validate()` (line 227) is a wrapper that adds a redundant `if *driverType == ""` check before calling `validateRequiredFlags`. The check is redundant: `driverTypeArg` (`"type"`) is already in `commonRequiredDriverFlags`, and even without that, `requiredDriverFlags("")` returns an error from the `default` branch. Since `validate()` has a single call site (line 388) and adds no logic beyond what `validateRequiredFlags` already provides, remove `validate()` entirely and call `validateRequiredFlags` directly.

backend/src/v2/compiler/argocompiler/container.go
Priyanshu-u07: Done, applied your suggested fix. Also cleaned up the matching dead code in the test helper `ConfigureCacheSettings` (`backend/test/compiler/utils/workflow_utils.go`). Since the copy init container is now just `--copy`, there are no cache args on any init container, so I dropped the init-container branch. The helper only rewrites `template.Container.Args`.

backend/src/v2/cmd/driver/main.go
Priyanshu-u07: Done

hbelmiro (approved):
Thank you very much for working on this @Priyanshu-u07.

/lgtm
/approve

PR #13715, opened 2026-07-13
ci: add runner telemetry, failure-signature summary, and Kind log export

jeffspahr:
## Problem

CI has no metrics-server and no external observability service, so intermittent lane failures (MLflow client timeouts, SeaweedFS ClusterIP dial timeouts — see #13686 / #13706) are triaged from raw job logs with zero system-level measurements. The current diagnosis of the MLflow lane is stalled on a question the logs cannot answer: **is the bottleneck the application (MLflow worker queueing) or the shared runner CPUs being saturated?** Both produce the identical client-side signature.

## Change

Three self-contained observability layers added to the `test-and-report` composite action, so every lane using it benefits. Everything renders into GitHub job summaries / artifacts — no external service, no tokens, no new infrastructure.

1. **Runner telemetry** — a self-contained `runner-telemetry.sh`: a background sampler reads host CPU/memory/load from `/proc` every 5s, and a render step writes a stats table (avg/peak) plus **mermaid charts directly into the job summary**. Deliberately hand-rolled: marketplace telemetry actions render charts by POSTing the collected samples to external services (verified in-source for the initially proposed action), which this repo must not depend on — nothing here leaves the runner. This is the discriminating measurement for "app-level bottleneck vs runner pegged" — for the MLflow lane, the SeaweedFS accept-queue hypothesis, and any future saturation question at once. Rendered for passing lanes too, so green-run baselines make saturated failing runs recognizable.

2. **Failure-signature summary** (on test failure) — a new `failure-signature-summary.sh` greps the already-collected pod logs and JUnit reports for the known failure classes and writes a classification table to the job summary:

   | Signature | Points at |
   |---|---|
   | ClusterIP dial `i/o timeout` (with per-VIP breakdown) | service dataplane / backend accept saturation |
   | client deadline / `Client.Timeout` | slow server (queueing) |
   | connection refused/reset | process down |
   | HTTP 429 | upstream rate limiting |
   | OOMKilled / CrashLoopBackOff | pod lifecycle |
   | `nf_conntrack: table full` | node conntrack exhaustion |

   plus the failed JUnit test names — so triage starts from a classified signal instead of log archaeology.

3. **Kind cluster log export** (on test failure) — `kind export logs` for every cluster, uploaded as a `kind-logs - <report>` artifact (5-day retention): kubelet, containerd, and kernel/dmesg logs the custom collectors don't capture.

Additionally, raw **JUnit XML from failed runs** is uploaded as a `junit-xml - <report>` artifact (30-day retention). The existing HTML report isn't machine-parseable; the XML enables per-test flake-rate analysis by a scheduled CI-health report (separate follow-up PR).

All additions are `continue-on-error` and change nothing for passing runs beyond the telemetry sampler.

## Verification

- Action YAML parses cleanly; script `bash -n` clean.
- `failure-signature-summary.sh` functionally tested against fixture pod logs + JUnit XML: renders the signature table with correct counts, per-VIP dial-timeout breakdown (e.g. `10.96.25.77:9000` vs `10.96.241.155:8443`), and failed-test list; exits 0 with missing inputs.
- Telemetry/export behavior is exercised by this PR's own CI lanes: every lane's job summary should now include telemetry charts, and any failing lane should include the signature table and a kind-logs artifact.


jeffspahr:
Good catch — confirmed in the pinned source: `statCollector.ts` POSTs the collected samples to `https://api.globadge.com/v1/chartgen/...` and embeds externally hosted chart URLs. That's an outbound data dependency this PR explicitly promised not to take.

Replaced in cae2e9b5d: the third-party action is gone entirely. Runner telemetry is now a self-contained `runner-telemetry.sh` — a background sampler reading host CPU/memory/load from `/proc` every 5s, plus a render step that writes an avg/peak stats table and **mermaid xychart charts natively into the job summary**. No third-party action, no outbound data, fully auditable in-repo. It also renders on passing lanes, so green-run baselines make saturated failing runs recognizable.

Verified: renderer tested against a synthetic sample CSV (stats table + both charts render); sampler smoke-tested against a real Linux `/proc` in a container (produces correct `epoch,cpu_pct,mem_used_mb,load1` samples); `bash -n` and action YAML parse clean.

jeffspahr:
CI triage for this PR's failing lanes: all are the known pre-existing v1.36.1 flake families (SeaweedFS `:9000` dial-timeouts — e.g. the E2EParallelNested failure logged 100× `dial tcp <vip>:9000: i/o timeout` — plus the MLflow-latency and multi-user classes; the Frontend Integration lane doesn't use `test-and-report` at all). Nothing here is introduced by this change.

Meanwhile this run **production-validated the PR**: the telemetry sampler ran for the full job on every lane, the render and failure-signature steps executed cleanly on the failed lanes, and all 5 failing E2E lanes uploaded both `kind-logs - <report>` and `junit-xml - <report>` artifacts as designed.

jeffspahr:
Three further review findings addressed in 1c7118617:

1. **[P1] Classifier now scans JUnit failure/error bodies** — client-side failures (MLflow timeouts, `connection refused` raised in the test process) live only in the JUnit XML, so pod-log-only scanning reported zero for exactly the client-observed classes. The scan corpus is now pod logs **+** extracted JUnit `<failure>`/`<error>` messages and text; the summary names which sources were scanned (`_Scanned: pod logs, JUnit failure bodies._`), and the table renders when either source exists. Repro from the review — empty pod log paired with an MLflow JUnit artifact containing `connection refused` — now reports `Connection refused / reset | 2` and `Client deadline / timeout | 1` instead of zeros.

2. **[P2] Bucket-max downsampling** — the every-Nth downsampler could drop a spike that the stats table still reported as the peak. Charts now take the max within each bucket, so a short saturation spike stays visible. Repro from the review — 51 samples, 50 zeros plus a final 100% CPU spike — now keeps the `100` in the chart line.

3. **[P2] Render under `always()`** — with the implicit `success()` condition, a setup failure (e.g. API configuration) skipped the render, left the sampler running until teardown, and dropped the telemetry for exactly the runs that need diagnosing. The render step now carries `if: ${{ always() }}` (still `continue-on-error`).

All three verified against the review's own repro scenarios; `bash -n` and action YAML parse clean.

jeffspahr:
**[P2] JUnit double-counting fixed** in the latest commit: Ginkgo duplicates the failure text into both the `<failure>` `message` attribute and the element body, and the extractor emitted both — doubling every JUnit-sourced signature count. It now prefers the body and falls back to `message` only when the body is empty.

Verified against the review's repros: the one-test timeout fixture counts **1** (was 2); a six-line connection-refused corpus with the same text duplicated into the attribute counts **6** (was 12); and a message-only failure (empty body) still counts via the fallback.

PR #13725, opened 2026-07-15
fix(sdk): relax exact requests pin to a compatible range. Fixes #13723

### LINKED ISSUE #13723: [sdk] Relax exact pin on requests (==2.33.0) to a compatible range
### Environment

*  KFP SDK version: 2.17.0
*  Python version: 3.10+ (any)

### Steps to reproduce

```
uv pip install kfp==2.17.0 requests==2.34.2
  × No solution found when resolving dependencies:
  ╰─▶ Because kfp==2.17.0 depends on requests==2.33.0 and you require kfp==2.17.0, we can conclude that you require requests==2.33.0.
      And because you require requests==2.34.2, we can conclude that your requirements are unsatisfiable.
```

### Expected result

Relax the exact pin to a compatible range that still respects the Python-version boundary that motivated #13661.

### Materials and Reference

requests is one of the most widely used libraries in the Python ecosystem, so an exact pin on it is especially disruptive, forcing every downstream consumer to hold requests at precisely the version kfp chose, and blocks routine security/patch updates to requests.

This is also a recurrence of the general concern raised long ago in #1134 (use ranges rather than exact pins), which is worth revisiting for requests specifically.

---

harshhh817:

`kfp` pins `requests` exactly (`==2.32.5` on Python 3.9, `==2.33.0` on Python 3.10+), which makes it uninstallable next to any other `requests` version and blocks downstream patch/security updates:

```
uv pip install kfp==2.17.0 requests==2.34.2
  × No solution found when resolving dependencies:
  ╰─▶ Because kfp==2.17.0 depends on requests==2.33.0 …
```

This PR keeps the Python-version boundary introduced in #13661 (requests 2.33.0 drops Python 3.9) but expresses both branches as ranges in `requirements.in` for the SDK and the kubernetes_platform package:

- `python_version < "3.10"`: `requests>=2.32.5,<2.33`
- `python_version >= "3.10"`: `requests>=2.33.0,<3`

The generated `requirements.txt` lock files are intentionally unchanged — the locked versions already satisfy the new ranges, so CI and reproducible installs keep resolving exactly as before. Only the wheel metadata (`install_requires`, read from `requirements.in` by `setup.py`) becomes range-based.

**Verified:**
- On master: `pip install --dry-run ./sdk/python "requests==2.34.2"` → `ResolutionImpossible` (reproduces the issue)
- With this change: same command resolves and would install `requests-2.34.2` alongside `kfp`

Fixes #13723

**Checklist:**

### DIFF
--- a/kubernetes_platform/python/requirements.in
+++ b/kubernetes_platform/python/requirements.in
@@ -7,9 +7,11 @@
 protobuf>=6.33.5,<7.0
 kfp>=2.14.5,<3
 # requests 2.33.0 drops Python 3.9 support. Keep Python 3.9 on the
-# compatible line and use the patched line where supported.
-requests==2.32.5; python_version < "3.10"
-requests==2.33.0; python_version >= "3.10"
+# compatible 2.32.x line and use the patched line where supported.
+# Ranges (not exact pins) so downstream consumers can take patch/security
+# updates to requests without conflicting with kfp (see #13723).
+requests>=2.32.5,<2.33; python_version < "3.10"
+requests>=2.33.0,<3; python_version >= "3.10"
 # urllib3 2.7.0 drops Python 3.9 support. Keep Python 3.9 on the
 # compatible line and use the patched line where supported.
 urllib3==2.6.3; python_version < "3.10"
--- a/sdk/python/requirements.in
+++ b/sdk/python/requirements.in
@@ -27,9 +27,11 @@ kubernetes>=8.0.0,<31
 protobuf>=6.31.1,<7.0
 PyYAML>=5.3,<7
 # requests 2.33.0 drops Python 3.9 support. Keep Python 3.9 on the
-# compatible line and use the patched line where supported.
-requests==2.32.5; python_version < "3.10"
-requests==2.33.0; python_version >= "3.10"
+# compatible 2.32.x line and use the patched line where supported.
+# Ranges (not exact pins) so downstream consumers can take patch/security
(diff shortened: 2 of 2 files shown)


harshhh817:
Friendly ping — this is still on `needs-ok-to-test`, so CI hasn't run. Could a maintainer trigger it when convenient?

It's a small SDK change: relaxes the exact `requests` pin to a compatible range while keeping the Python-version boundary introduced in #13661, with the generated lock files intentionally unchanged. Happy to rebase if it's drifted.

harshhh817:
@mprahl @chensun — this one is still sitting on `needs-ok-to-test`, so CI has never run on it. Could one of you trigger it when you have a moment?

Re-checked against master today and the exact pin is still there in `sdk/python/requirements.in`:

```
requests==2.32.5; python_version < "3.10"
requests==2.33.0; python_version >= "3.10"
```

So `kfp` still cannot be installed alongside any other `requests` version, and downstream patch/security updates to `requests` are still blocked. #13723 is open and has no comments.

The change is two lines in two `requirements.in` files — it keeps the Python-version boundary from #13661 and just expresses each branch as a range instead of an equality. Happy to rebase or adjust the bounds if you would rather they were tighter.

droctothorpe:
Hi @harshhh817, thank you for your PR! The linked issue (#13723) must be labeled `ready` before your contribution can be reviewed. Please check our [CONTRIBUTING.md](https://github.com/kubeflow/pipelines/blob/master/CONTRIBUTING.md) for details.

Issue #13724, opened 2026-07-15
ci: configure MLflow workers at initial deploy and re-evaluate E2E concurrency


jeffspahr:
## Goal

Allow the operator-managed Kubernetes-auth MLflow deployment to set its worker
count in the initial custom resource, then determine how much E2E concurrency
can be restored without reintroducing shared-cluster capacity failures.

## Current state — 2026-07-24

#13720 reduced the Kubernetes-auth MLflow lanes to two Ginkgo nodes after
higher concurrency produced MLflow, KFP API, MLMD, and SeaweedFS failures on a
single-node Kind cluster.

#13739 subsequently removed the kindnet CPU bottleneck and restored higher
concurrency across the broader E2E matrix. The MLflow lanes intentionally
remain at two Ginkgo nodes while their separate deployment constraint is
resolved.

The pinned
`opendatahub-io/mlflow-operator/.github/actions/deploy` action still does not
expose a `workers` input:

- file-backed Kubernetes-auth lanes patch `spec.workers: 2` after creation and
  wait for the rollout
- S3-backed Kubernetes-auth lanes retain the action's one-worker default
  because a post-deploy rolling update previously required a transient extra
  pod and stalled on the constrained runner
- direct basic-auth/none lanes are independent and remain unchanged

The post-deploy patch is a useful targeted mitigation, but it does not satisfy
the initial-CR configuration requirement.

### Upstream action

- Add a backward-compatible `workers` input to
  `opendatahub-io/mlflow-operator/.github/actions/deploy`.
- Keep the existing one-worker default.
- Apply the requested value when the initial MLflow custom resource is
  created, avoiding a second rollout.
- Add focused action/deploy-script coverage for the default and configured
  values.

### Pipelines integration

- Update KFP to a merged, SHA-pinned upstream action commit.
- Configure two workers on the initial Kubernetes-auth MLflow custom resource
  where appropriate.
- Remove the file-only post-deploy worker patch.
- Increase MLflow Ginkgo concurrency incrementally and retain the existing
  service, pod, throttling, and latency diagnostics.

## Acceptance criteria

- The upstream deploy action supports an optional worker count without changing
  its default behavior.
- KFP consumes a merged, SHA-pinned action revision and no longer needs a
  post-deploy worker patch.
- File- and S3-backed Kubernetes-auth lanes create the intended worker count
  without rollout scheduling failures.
- All Kubernetes-auth MLflow matrix lanes pass at the proposed higher
[...]

PR #13726, opened 2026-07-15
ci: capture service state on connection failures

jeffspahr:
## Summary

- correlate ClusterIP connection refusals and resets in addition to dial timeouts
- snapshot the matched Service, EndpointSlices, backing pod readiness, restart state, and events
- keep node conntrack and iptables/IPVS evidence while filtering historical Pod-IP probe noise
- distinguish Kubernetes API lookup failures from genuinely absent Services or endpoints
- add stdlib regression tests and run them in the CI Scripts Tests workflow

## Why

The failed Kubernetes-auth MLflow lane in #13720 logged `connection refused` for the MLflow ClusterIP. The existing collector only recognized `i/o timeout`, so it inspected SeaweedFS but missed the Service and backend that actually refused the connection.

This change makes a future occurrence show whether the matched Service had ready endpoints, whether its backing pod restarted or became unready, and whether the node service program remained valid. It supports the capacity investigation tracked in #13724.

## Verification

- `bash -n .github/resources/scripts/collect-seaweedfs-diagnostics.sh`
- `cd .github/resources/scripts && python3 -m unittest -v collect_seaweedfs_diagnostics_test ci_health_report_test` (24 tests)
- parsed the changed workflow and composite action as YAML
- `git diff --check`

## Also folded in: runner-kernel softirq backlog capture (from #13729)

Telemetry from the first fully-instrumented SeaweedFS dial-timeout failure showed the 4-CPU runner at **load 17–28** while every instrumented drop point was clean (conntrack ~0.7% with no table-full, live KUBE-SEP DNAT, pod `ListenOverflows=0`). The remaining drop point upstream of all of those is the **per-CPU softirq backlog** — veth traffic queues there, and when `ksoftirqd` is starved the queue overflows and packets drop silently.

A new host-level section reads `/proc/net/softnet_stat` **once on the runner** (it's kernel-global state shared by every Kind node), decoding `processed / dropped / time_squeeze` per CPU using the kernel's CPU-id field (13-column format; row-ordinal fallback for 11-column kernels), plus the current `netdev_max_backlog` / `netdev_budget` / `netdev_budget_usecs` limits so each failure records what the queue limits were.

Deliberately diagnostics-only (no sysctl bump), so the counter gives one unconfounded read: **`dropped > 0`** on the next failure confirms the softirq-backlog mechanism and justifies a follow-up bump; **zero** refutes it and redirects the investigation. Both changes rewrite this script, so they ship together; #13729 is closed in favor of this PR. The test harness asserts the section renders in the fake-cluster run.

### DIFF
--- a/.github/actions/test-and-report/action.yml
+++ b/.github/actions/test-and-report/action.yml
@@ -391,15 +391,14 @@ runs:
         if [ "${{ inputs.proxy }}" = "true" ]; then
           ./.github/resources/scripts/collect-logs.sh --ns tinyproxy --output /tmp/tmp_pod_log.txt
         fi
-        # Dataplane diagnostics: E2E lanes intermittently fail with dial-timeout
+        # Dataplane diagnostics: E2E lanes intermittently fail with connection
         # errors to in-cluster ClusterIPs (SeaweedFS :9000, the MLflow service
         # VIP, ...). This captures SeaweedFS restart/contention evidence plus,
         # per Kind node, the conntrack table state and the iptables/ipvs program
-        # for every ClusterIP the just-collected pod logs recorded a dial
-        # timeout against -- distinguishing conntrack exhaustion from a
-        # missing/stale service program. --dial-timeout-log points at the pod
-        # logs gathered above so the probe self-targets whichever VIP failed.
-        ./.github/resources/scripts/collect-seaweedfs-diagnostics.sh --namespace "$NAMESPACE" --dial-timeout-log /tmp/tmp_pod_log.txt --output /tmp/tmp_pod_log.txt
+        # for every ClusterIP the just-collected pod logs recorded a timeout,
+        # refusal, or reset against -- distinguishing backend lifecycle from
+        # conntrack exhaustion or a missing/stale service program.
+        ./.github/resources/scripts/collect-seaweedfs-diagnostics.sh --namespace "$NAMESPACE" --connection-failure-log /tmp/tmp_pod_log.txt --output /tmp/tmp_pod_log.txt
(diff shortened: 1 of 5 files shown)


jeffspahr:
Addressed in the latest commit: the no-pod early exit is gone. Only the pod-scoped sections (1 restarts, 2 node contention, 6 socket state, final describe) are guarded on pod availability — the Service inventory and sections 4–5 (Service/EndpointSlice snapshots, node netfilter/service-program probes) now always run, so an MLflow or other ClusterIP failure gets its full snapshot even when SeaweedFS is absent. A note names the skipped sections.

Covered by a new regression test (`NO_SEAWEEDFS_POD` toggle in the fake kubectl): with the pod absent, the MLflow VIP still receives its `Service: opendatahub/mlflow (failed VIP …)` backend snapshot and both `ClusterIPs correlated` lines, while sections 1/2/6 report clean skips. Full scripts suite (25 tests) passes.

jeffspahr:
**[P2] Unowned logged VIPs now keep their node-level probe** (latest commit): a Service deleted/recreated/re-IP'd after the failure leaves its old VIP absent from the current inventory — the missing/stale-Service case the node probe exists for — so those targets are no longer dropped. Backend-object lookups (section 4) stay limited to currently-owned VIPs; the node netfilter probe (section 5) now covers owned **plus** unowned logged targets, and the section-4 note says "backend lookup skipped; node program probed in section 5" instead of "ignored".

Test updated accordingly: the unowned target is asserted absent from section 4's correlated list, present in section 5's, and still receives no backend-object lookup. Suite green.

PR #13727, opened 2026-07-15
ci: mirror observability output to logs and fix conntrack self-match

jeffspahr:
## Problem

Two defects surfaced by the first fully-instrumented failure on master (run [29385440051](https://github.com/kubeflow/pipelines/actions/runs/29385440051), follow-ups to #13715/#13706):

1. **The observability output is invisible to automation.** The failure-signature table and runner-telemetry stats render only into `GITHUB_STEP_SUMMARY`, which has no API — the telemetry that finally evidenced runner CPU oversubscription (load avg 16.9 / peak **28.1 on a 4-CPU runner** during the SeaweedFS dial-timeout window) could only be read by a human opening the job-summary page. Log-based consumers (tooling, the CI health report, anyone triaging via `gh api`) get nothing.

2. **A signature self-match.** The table reported `conntrack table full | 1` on a run whose dmesg read said "no table-full event logged": `collect-seaweedfs-diagnostics.sh` tees its own header line — `kernel 'nf_conntrack: table full' events:` — into the collected pod log, and the signature regex matched the header text.

## Change

- Both `failure-signature-summary.sh` and `runner-telemetry.sh` now **mirror their rendered output to stdout** (tee) in addition to the step summary, so the job log is a machine-readable record of the same data. Without `GITHUB_STEP_SUMMARY` set they write to stdout only, as before.
- The conntrack signature matches the kernel's actual message form, `nf_conntrack: table full, dropping packet`, so the diagnostics' own headers can never count as hits.

## Verification

- A fixture pod log containing the diagnostics header **plus one real kernel table-full line** now counts **1** (was 2).
- With `GITHUB_STEP_SUMMARY` set, both scripts emit identical content to the summary file and stdout; without it, stdout only. `bash -n` clean on both.

Context: the mirrored telemetry is what evidenced the current leading root cause for the `dial tcp <vip>: i/o timeout` family — sustained 4–7× CPU oversubscription of the shared runner — complementing #13720's parallelism reduction.

### DIFF
--- a/.github/resources/scripts/failure-signature-summary.sh
+++ b/.github/resources/scripts/failure-signature-summary.sh
@@ -30,7 +30,15 @@ while [[ "$#" -gt 0 ]]; do
     shift
 done
 
-OUT="${GITHUB_STEP_SUMMARY:-/dev/stdout}"
+# Job summaries are not retrievable through the logs API, so mirror the
+# output to stdout: the job log stays the machine-readable record.
+publish() {
+    if [[ -n "${GITHUB_STEP_SUMMARY:-}" ]]; then
+        tee -a "$GITHUB_STEP_SUMMARY"
+    else
+        cat
+    fi
+}
 
 # Signature scan corpus: the collected pod logs PLUS the JUnit failure/error
 # bodies. Client-side failures (MLflow timeouts, connection refused from the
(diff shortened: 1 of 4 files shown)


jeffspahr:
**[P2] Regression coverage added** in the latest commit: new `observability_output_test.py` (stdlib-only, drives both shell scripts via subprocess) covering exactly the changed behaviors:

- **Conntrack classification** — the collect-seaweedfs diagnostics header line (`kernel 'nf_conntrack: table full' events:`) counts **zero**, while real kernel messages (`nf_conntrack: table full, dropping packet`) count correctly (fixture with two events asserts count 2).
- **Output mirroring** — for both `failure-signature-summary.sh` and `runner-telemetry.sh render`, stdout is asserted **byte-identical** to what gets appended to `GITHUB_STEP_SUMMARY`, and rendering still works with the summary unset (stdout-only).

The module is wired into the `ci-scripts-tests.yml` presubmit, and that gate now also triggers on `.github/resources/scripts/*.sh` changes since the Python tests cover the shell scripts. All 5 tests pass locally and run in this PR's CI.

PR #13728, opened 2026-07-15
fix(ci): wait for delayed test readiness

jeffspahr:
## Summary

- poll the artifact proxy health endpoint until the API server is ready
- retain the final health response and artifact-proxy logs when readiness times out
- increase the frontend hello-world run completion budget from one to three minutes

## Root cause

The artifact proxy can return a successful health response with
`apiServerReady: false` while its API server dependency is still starting. The
test treated that first response as final. Separately, the frontend integration
test allowed only 60 seconds for a hello-world workflow that can complete
successfully shortly after that deadline under CI load.

## Verification

- `bash -n test/artifact-proxy/test-artifact-proxy.sh`
- `node --check test/frontend-integration-test/helloworld.spec.js`
- `git diff --check`
- fake-kubectl execution covering recovery on a later health response
- fake-kubectl execution covering bounded exhaustion and diagnostic output

### DIFF
--- a/.github/workflows/e2e-test.yml
+++ b/.github/workflows/e2e-test.yml
@@ -31,6 +31,9 @@ on:
       - 'backend/**'
       - 'proxy/**'
       - 'manifests/kustomize/**'
+      # The artifact-proxy E2E script is exercised only by this workflow, so
+      # changes to it must trigger these lanes.
+      - 'test/artifact-proxy/**'
       - 'test_data/sdk_compiled_pipelines/**'
       - '!**/*.md'
       - '!**/OWNERS'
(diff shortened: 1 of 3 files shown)


jeffspahr:
**[P2] Coverage gap fixed** in the latest push (also rebased onto current master): `e2e-test.yml` is the only caller of `test/artifact-proxy/test-artifact-proxy.sh`, but its `pull_request` path filters omitted `test/artifact-proxy/**` — so this PR's readiness fix would have merged without ever running the workflow that consumes it. Added the path, so the artifact-proxy lanes now trigger on this PR (and any future change to the script).

The PR's other file, `test/frontend-integration-test/helloworld.spec.js`, was already covered — `e2e-test-frontend.yml` includes `test/frontend-integration-test/**` in its filters.

PR #13739, opened 2026-07-16
ci: stabilize and accelerate loaded Kind test lanes

jeffspahr:
## Summary

This PR stabilizes the loaded Kind-based test lanes, removes the kindnet CPU bottleneck identified by the new diagnostics, and shortens the slowest E2E/API paths without reducing executed test coverage.

- keep kindnet's `100m` CPU request but remove its CPU limit in every CI Kind cluster, allowing the NFQUEUE forwarding path to drain bursts without a `100m` quota
- return automatic Critical, Essential, and multi-user E2E lanes to the historical ten Ginkgo nodes; retain lower concurrency for MLflow, E2EFailure, and nested-pipeline lanes while preserving `workflow_dispatch` overrides
- split standard automatic and multi-user Critical suites into duration-balanced 17/16 pipeline shards on independent Kind runners while keeping low-Kubernetes, TLS, and manual-dispatch coverage whole; cache-enabled multi-user shards retain artifact-proxy coverage
- return multi-user API integration to two Ginkgo nodes and partition the suite across the existing cache-enabled/cache-disabled jobs with complementary filters
- harden SeaweedFS IAM/Profile setup, Argo compatibility diagnostics, v1 termination-state handling, and the loaded frontend/TensorBoard waits that produced deterministic failures while this PR was iterating
- add test-window Service-path, kindnet/NFQUEUE, conntrack, TCP, CPU-pressure, and bounded packet-capture diagnostics to attribute future failures without changing the production-equivalent DNS path used by the tests
- skip expensive hosted-runner cleanup when at least 60 GiB is already free and import control-plane archives two at a time
- overlap profile-controller startup with the main multi-user KFP rollout and overlap upgrade setup with current-branch image builds
- split the legacy V2 PVC-cache long tail into an independent job while preserving every store/TLS variant

This supports #13724 and the remaining Service-path investigation in #13748.

## Root cause and mitigation

At head `d79ec92637`, three otherwise different lanes failed while kindnetd reported `failed to set verdict with label`, `netlink send: i/o timeout`, and repeated NFQUEUE receive errors:

- Critical E2E lost the SeaweedFS Endpoint path in 630/727 probe cycles and its Service VIP in 629/727 while CoreDNS and the Kubernetes API remained healthy.
- Direct/no-auth MLflow lost both SeaweedFS paths in 194/284 cycles while both control paths remained healthy.
- Multi-user API lost the VIP and Endpoint paths for SeaweedFS, CoreDNS, and the Kubernetes API together; the API server then timed out resolving MySQL through CoreDNS.

Packet capture showed SYNs entering the Kind node without timely forwarding to the destination veth. SeaweedFS stayed ready with no restarts or listen-queue drops, and conntrack remained far below capacity. This localized the outage to Kind's kindnetd/NFQUEUE forwarding layer rather than any individual backend Service.

The Kind manifest requested `100m` CPU for kindnet and also capped it at `100m`. This PR keeps the request for scheduling/accounting but removes the limit so forwarding can burst when the test runner is loaded.

For multi-user setup, Profile creation now waits from the profile controller's own network namespace until SeaweedFS IAM `:8111` is reachable. The ingress NetworkPolicy permits only the profile controller to make that connection. Readiness still uses the same DNS target as production; concurrent VIP/Endpoint checks provide diagnostics only. Failures collect the relevant Service, EndpointSlice, pod, controller, Metacontroller, and event state.

## Test scheduling changes

- Critical, Essential, and multi-user E2E: 10 nodes
- standard automatic and multi-user Critical E2E: duration-balanced shard A/B jobs; low-Kubernetes, TLS, and manual dispatch remain unsharded, while cache-enabled multi-user shards retain artifact-proxy coverage
- MLflow and E2EFailure: 2 nodes
- nested-pipeline E2E: 3 nodes
- multi-user API integration: 2 nodes within each of two complementary cache shards
- focused Argo runtime compatibility specs: serial

The API selectors statically partition 626 eligible specs into 327/299 groups, including an even 151/151 split of the 302 cache-sensitive run cases. In the settled run, Ginkgo executed 325 and 291 specs respectively: 616 total, exactly matching the previous unsharded lane's executed coverage. The remaining selector-eligible specs are runtime-skipped in both arrangements.

The workflow-dispatch label expression is passed through environment-backed configuration and a quoted `--label-filter`, so arbitrary expressions continue to work.

## Diagnostics added

The test-and-report path now:

- compares SeaweedFS S3, Kubernetes API, and CoreDNS Service VIPs with ready direct Endpoints from one stable pod/network namespace
- establishes baselines before the first active probe and samples runner CPU PSI/cgroup state plus per-node conntrack, TCP, socket, NFQUEUE, nftables, and kindnet scheduler/throttling state every five seconds
- correlates the first failed probe with NFQUEUE verdict errors, slow nftables syncs, and the first following kernel sample while retaining stronger later spikes
- snapshots the exact ready SeaweedFS pod's network namespace, TCP state, and `:8333` listen queue
- validates bounded SYN/RST packet capture and retains raw JSONL/pcap evidence only for failed jobs
- quiesces probes before stopping observers and verifies process identity before signalling collector PIDs
- applies the same network window to the legacy V2 integration workflow
- captures one bounded pod describe/events snapshot when an init or regular container remains in `ContainerCreating` for more than 60 seconds, even if deployment later succeeds

UDP DNS timeouts retain their protocol during Service-rule correlation, and optional kindnet collection cannot hide the core node metrics if an optional probe fails.

## Settled baseline

Baseline head `8857c8effa` passed all 193 GitHub Actions jobs. The only remaining pending external check is Tide waiting for review labels.

| Measurement | Previous green head (`c0ef9a4`) | Baseline head (`8857c8e`) | Change |
| --- | ---: | ---: | ---: |
| Longest Actions workflow | 30:57 | 25:59 | -4:58 (-16%) |
| E2E workflow | 30:54 | 24:07 | -6:47 (-22%) |
| Critical E2E test-step mean | 12:15 | 9:47 | -20% |
| Multi-user E2E test-step mean | 13:16 | 10:28 | -21% |

API sharding reduced the longest Ginkgo execution from 17:24 to 8:21 while preserving the same 616 executed specs. The remaining API critical-path variance is cluster creation and KFP deployment rather than test execution.

Across representative 8-node Critical, 8-node multi-user, and API-shard lanes, the collector made 1,974 Endpoint/VIP connection attempts with no DNS, Kubernetes API, or SeaweedFS failure or timeout. Those lanes also recorded zero NFQUEUE send/receive errors, conntrack drops, TCP listen drops/overflows, SYN retransmits, TCP timeouts, and runner cgroup throttling.

The first settled eight-node ramp materially loaded the runner—representative Critical and multi-user lanes averaged 80-83% CPU and peaked at 99%—but remained network-clean. That evidence supports the next measured pass at the historical ten-node setting, with Critical sharding reducing per-cluster pipeline pressure.

## Settled wall-clock experiment

Head `3c4b5481fd` passed all 196 completed checks. Compared with green baseline `8857c8effa`, the longest workflow improved from 25:59 to 23:50; API Server, legacy V2, and upgrade improved by 6:43, 5:03, and 7:22 respectively. The conditional cleanup, bounded image imports, startup overlap, upgrade overlap, and legacy sharding paths all executed as intended. Representative diagnostics recorded 2,088 Service-path attempts with zero failures/timeouts and zero NFQUEUE, conntrack, TCP, or cgroup-throttling signals.

Legacy sharding cut workflow wall clock by 21% while increasing aggregate test-job runner time by 39%; this is an accepted tradeoff on the base hosted-runner pool. The remaining gate was one 33-spec Critical E2E suite at 667.8 seconds, motivating the next pass.

## Next wall-clock pass

Head `7699c1acbe` combines the next three measured experiments:

- all 12 previously serialized image-loading job definitions now create Kind while the reusable image build runs; the shared deploy action joins the complete 13-artifact inventory immediately before download
- the matrix-wide artifact poll uses a 30-second cadence and ten-minute bound to avoid a thundering herd against the shared repository API limit
- automatic Critical, Essential, and multi-user E2E return from eight to the historical ten Ginkgo nodes
- standard Critical variants run two independent Kind jobs selected by pipeline-level shard labels; both multi-user cache variants are sharded, with the cache-enabled pair retaining artifact-proxy compatibility, while low-Kubernetes and TLS lanes stay whole
- manual dispatch retains the original unsharded labels and requested concurrency

All 33 Critical specs appeared in each of 12 recent green lane logs. Greedy balancing by their per-spec mean duration produces 17/16 groups with 2,090.2/2,036.4 seconds of cumulative work, a 1.3% imbalance. New Critical pipelines default to shard B so they remain covered, while the unit test requires the partition to be rebalanced when the directory changes.

This pass deliberately adds hosted runners to reduce feedback latency; it does not remove variants, relax readiness, shorten product timeouts, or alter production behavior.

## Artifact-proxy sharding and deployment-tail diagnostics

Head `3197ba5935` shards the cache-enabled multi-user/artifact-proxy Critical lane across the existing duration-balanced A/B labels. Automatic runs now use four multi-user jobs (cache-disabled A/B and artifact-proxy A/B); manual dispatch remains unsharded for both cache modes.

The KFP readiness loop now tracks pod name/UID across init and regular containers. After 60 seconds of continuous `ContainerCreating`, it emits one bounded, best-effort `kubectl describe pod` plus UID-filtered pod events. The episode resets when the condition clears or the pod is recreated, and diagnostic failures cannot change deployment readiness semantics.

## Additional flake fixes

- wait for the asynchronous v1 run termination transition instead of immediately asserting while status can still be `Unknown`
- allow loaded Kind lanes up to 30 seconds for the frontend pod-log smoke request and 120 seconds for the TensorBoard example run
- retry transient Service/setup operations and emit focused orchestration state when Argo compatibility cases time out

## Verification

- all 122 stdlib CI-script tests, including the deploy-overlap inventory guard, artifact-proxy shard guard, and slow-pod diagnostic state coverage
- `go test ./backend/test/constants` plus Ginkgo dry-runs selecting exactly 17 shard-A and 16 shard-B Critical specs
- `go test -p=1 ./backend/test/v2/api -run 'Test(FindArgoCompatibility|RuntimeState)' -count=1`
- `go test ./backend/test/integration -run '^$' -count=1`
- selector audit covering all 626 eligible multi-user API specs with no shard intersection
- repository-pinned actionlint and Ruby YAML parsing for the changed workflows/action
- focused Go-filter proof that the legacy V2 main/PVC-cache selectors are complementary
- Argo matrix and static concurrency/dispatch/selector assertions
- Go formatting and constants-package compilation for the new labels
- frontend integration JavaScript syntax check and frontend workflow YAML parse
- rendered all three multi-user overlays and verified the IAM source selector matches the profile-controller pod labels
- `bash -n` for the changed shell scripts
- Python compilation for the changed diagnostics and test harnesses


jeffspahr:
## Post-green 8-node and API-sharding results

The settled `8857c8effa` matrix is fully green: all 193 GitHub Actions checks passed. `tide` is the only pending external check, and it is waiting for review labels rather than CI.

### Wall-clock results

| Measurement | Previous green head (`c0ef9a4`) | Current head (`8857c8e`) | Change |
| --- | ---: | ---: | ---: |
| Longest Actions workflow | 30:57 | 25:59 | -4:58 (-16%) |
| E2E workflow | 30:54 | 24:07 | -6:47 (-22%) |
| Critical E2E test-step mean | 12:15 | 9:47 | -20% |
| Multi-user E2E test-step mean | 13:16 | 10:28 | -21% |

Runs: [current E2E](https://github.com/kubeflow/pipelines/actions/runs/29625756957), [previous E2E](https://github.com/kubeflow/pipelines/actions/runs/29623211291), [current API](https://github.com/kubeflow/pipelines/actions/runs/29625756886), [previous API](https://github.com/kubeflow/pipelines/actions/runs/29623211306).

### API sharding

The previous cache-disabled API lane ran 616 specs in 17:24. The new shards ran:

- cache enabled: 325 specs in 8:21
- cache disabled: 291 specs in 7:59

The two shards still execute exactly 616 specs, preserving the prior coverage while cutting the longest Ginkgo execution roughly in half. The cache-disabled job remains the overall critical path because this run spent 4:56 creating the cluster and 7:57 deploying KFP; its test step was only 9:05. Setup/deployment variance is therefore the next wall-clock opportunity.

### Dataplane and host signals

I inspected representative 8-node Critical, 8-node multi-user, and API-shard lanes: 329 sampling cycles covering 1,974 endpoint/VIP connection attempts. All completed without a DNS, Kubernetes API, or SeaweedFS connection failure or timeout.

The same lanes recorded:

- zero NFQUEUE send/receive errors
- zero conntrack insert failures or drops
- zero TCP listen drops, overflows, SYN retransmits, or TCP timeouts
- zero runner cgroup throttling

The higher E2E concurrency is materially loading the runner: the representative Critical and multi-user lanes averaged 80-83% CPU, peaked at 99%, and recorded CPU PSI `some` at 58-61%. Connection p95 reached about 48 ms. For comparison, the prior 5-node Critical TLS lane averaged 46% CPU, peaked at 86%, recorded 22% CPU PSI `some`, and kept SeaweedFS connection p95 below 7 ms—but it was also slower.

### Conclusion

Eight Ginkgo nodes produced a real 20-22% E2E speedup and remained healthy with the uncapped Kind networking daemon. I recommend holding at 8 rather than moving immediately to 10 based on a single green 8-node run. The next optimization target should be cluster creation and KFP deployment variance in the API jobs.

jeffspahr:
## Settled combined wall-clock experiment

Head `3c4b5481fd` completed with all 196 GitHub Actions checks passing. The combined scheduling/setup changes reduced the full-matrix gate from 25:59 at green baseline `8857c8effa` to 23:50.

| Workflow | Baseline | Current | Change |
|---|---:|---:|---:|
| Overall gate | 25:59 | 23:50 | **-2:09 (-8%)** |
| API Server | 25:59 | 19:16 | **-6:43 (-26%)** |
| Legacy V2 | 24:03 | 19:00 | **-5:03 (-21%)** |
| Upgrade | 23:53 | 16:31 | **-7:22 (-31%)** |
| E2E | 24:07 | 23:50 | -0:17 |

Runs: [current E2E](https://github.com/kubeflow/pipelines/actions/runs/29629385146), [API Server](https://github.com/kubeflow/pipelines/actions/runs/29629385143), [Legacy V2](https://github.com/kubeflow/pipelines/actions/runs/29629385139), [upgrade](https://github.com/kubeflow/pipelines/actions/runs/29629385196); [baseline E2E](https://github.com/kubeflow/pipelines/actions/runs/29625756957), [API Server](https://github.com/kubeflow/pipelines/actions/runs/29625756886), [Legacy V2](https://github.com/kubeflow/pipelines/actions/runs/29625756873), [upgrade](https://github.com/kubeflow/pipelines/actions/runs/29625756955).

### What the experiment established

- Conditional cleanup skipped runners with 85-88 GiB free, but correctly retained the fallback on the gating E2E runner, which started with only 15 GiB free.
- Upgrade setup ran concurrently with branch image builds. When preparation reached the new barrier, all 13 artifacts were available and the wait took only two seconds.
- Profile-controller readiness was hidden behind the main KFP rollout. In the sampled API multi-user job this removed about 52 seconds of serialized waiting while retaining the IAM and Profile-creation gates.
- Two-at-a-time control-plane imports completed 11-36 seconds faster than sequential imports in sampled current/baseline jobs.
- The Legacy V2 main/PVC-cache split reduced workflow wall clock by 5:03. Its six test jobs used 80:52 aggregate runner time versus 58:09 for the three baseline jobs (+22:43, +39%). These are base hosted runners with no marginal runner charge, so the compute tradeoff is acceptable for the 21% reduction in feedback latency.

### Remaining gate and network state

The current gate is one `E2ECritical`, Argo v4/cache-disabled job. It ran the same 33 specs in 667.8 seconds versus 552.7 seconds at baseline (+1:55). Across all six Critical variants, mean test time moved only from 9:47 to 10:09; this single outlier accounts for nearly all of the difference.

This was not a dataplane recurrence. Representative current Critical, multi-user, and Legacy lanes made 2,088 paired endpoint/VIP connection attempts with zero failures or timeouts. They also recorded zero NFQUEUE send/receive errors, conntrack insertion failures/drops, TCP listen drops/overflows, SYN retransmits, TCP timeouts, and runner cgroup throttling. The gating lane averaged 40% CPU, peaked at 74%, and had less CPU PSI `some` stall time than its faster baseline counterpart.

### Next wall-clock opportunities

1. Extend the upgrade workflow's artifact-barrier pattern to the loaded E2E/API/Legacy workflows: start checkout and Kind cluster creation while image builds run, then join artifacts immediately before deploy. In this run that can hide up to the gating lane's 2:39 cluster-creation step and should move the overall gate toward 21 minutes without changing test concurrency or coverage.
2. Then test the historical 10-node setting for Critical/Essential/multi-user E2E. Eight-node networking remains clean with the kindnet limit removed; the likely incremental gain is about 45-90 seconds, with the existing diagnostics guarding the experiment.
3. If the E2E tail still dominates, split the 33 Critical pipelines into two duration-balanced groups on independent Kind runners. Base-runner cost makes this a reasonable latency tradeoff, but the partition should use multi-run JUnit durations so the longest pipelines are balanced rather than hashed blindly.

The combined changes should remain. The first next step should be broadening build/setup overlap because the upgrade run has now validated that mechanism end to end.

jeffspahr:
## Settled 10-node / build-overlap / Critical-sharding pass

Head `7699c1acbe` completed with all 200 CI checks successful; Tide was the only pending external gate at the time of measurement. [E2E run](https://github.com/kubeflow/pipelines/actions/runs/29631968160).

### Wall clock

| Measurement | Previous green pass (`3c4b5481fd`) | Current | Change |
|---|---:|---:|---:|
| Longest workflow / E2E | 23:50 | 23:10 | **-0:40** |
| Loaded-workflow aggregate (9 workflows) | 140:22 | 135:32 | **-4:50** |
| Legacy V2 | 19:00 | 17:32 | **-1:28** |
| KFP Kubernetes Native Migration | 12:10 | 10:43 | **-1:27** |
| v1 integration | 14:31 | 13:41 | **-0:50** |

Relative to the original `8857c8effa` baseline, the overall gate is down from 25:59 to 23:10 (**-2:49**) and E2E is down from 24:07 to 23:10 (**-0:57**).

Build/deploy overlap is working, but the gate improvement was modest in this sample: image inventory became available about 3:42 after workflow start versus a previous build completion around 3:21. Cluster setup overlapped that work, but the slower build plus the 30-second poll cadence consumed part of the expected gain.

### Critical sharding and concurrency

The standard Critical test step improved from a previous mean of 10:24 to:

- Shard A: **6:32 mean**
- Shard B: **5:17 mean**
- Per-variant Critical gate: **6:32**, about **37% faster**

The historical split prediction was close in aggregate, but observed 10-node makespan still differs by about 75 seconds. A measured rebalance could recover roughly another minute, although this lane no longer gates the workflow.

Ten-node concurrency continues to look healthy:

- Essential mean test time improved from 6:51 to 6:31 (**4.7% faster**).
- Standard Critical CPU PSI some-stall averaged 44.7%, versus 47.0% at 8 nodes.
- Multi-user CPU PSI some-stall averaged 37.9%, versus 47.5% previously.
- No runner or kindnet cgroup throttling was observed.

### Network evidence

Across all eight standard Critical shards and three multi-user jobs there were no Service-path failures, NFQUEUE errors, conntrack insert/drop signals, softnet drops, TCP listen drops/overflows, SYN retransmit/timeout signals, or runner/kindnet throttling.

One unsharded low-Kubernetes Critical lane recorded a single simultaneous SeaweedFS VIP and endpoint timeout in 129 probe cycles. It coincided with a brief runner PSI / kindnet scheduler-wait spike, while every kernel/drop counter remained zero and the tests passed. That is consistent with a short scheduling stall, not evidence of the prior dataplane quota/drop failure.

### New critical path and next steps

The E2E gate has shifted to two multi-user jobs:

1. The cache-enabled artifact-proxy lane has about **11:44** of deterministic unsharded test work.
2. A cache-disabled shard spent about six minutes with `cache-server` in `ContainerCreating`; `cache-deployer` restarted once before readiness, despite its test step taking only **6:18**.

The next opportunities are therefore:

1. Shard the cache-enabled multi-user/artifact-proxy lane.
2. Capture focused describe/events diagnostics for pods that remain `ContainerCreating` longer than 60 seconds, even when deployment eventually succeeds.
3. Rebalance Critical A/B from observed 10-node makespan after the current gate work.

The tradeoff is higher runner use: E2E now has 31 jobs versus 26, with aggregate E2E job wall time up about 39.7%, while aggregate test-step time increased only 4.6%. Given these are base GitHub-hosted runners, the wall-clock reduction remains worthwhile.

jeffspahr:
## Final settled artifact-proxy sharding pass

Head `3197ba5935` completed with all **202 GitHub Actions checks successful**. Tide is the only pending external gate. [E2E run](https://github.com/kubeflow/pipelines/actions/runs/29633790588).

| Measurement | Previous green pass (`7699c1acbe`) | Current | Change |
|---|---:|---:|---:|
| Longest workflow / E2E | 23:10 | 21:56 | **-1:14** |
| Artifact-proxy Critical test phase | 11:44 | 6:40 max (6:05 / 6:40) | **-5:04** |
| Artifact-proxy job wall clock | 23:06 | 17:54 max | **-5:12** |

Relative to the original `8857c8effa` baseline, the longest workflow is down from 25:59 to 21:56 (**-4:03 / 16%**) and E2E is down from 24:07 to 21:56 (**-2:11 / 9%**).

Both artifact-proxy shards passed deployment, the artifact-proxy smoke test, namespace isolation, and their duration-balanced Critical test partition. The new slow-pod diagnostic did not fire in any of the four multi-user jobs, so none continuously remained in `ContainerCreating` for more than 60 seconds during this pass.

The E2E critical path has moved to the deliberately unsharded compatibility lanes:

- low-Kubernetes Critical: 21:50 job wall, 10:23 test phase
- pod-to-pod TLS Critical: 21:41 job wall, 10:07 test phase

Recommendation: merge the current PR. It is fully green, has no unresolved review threads or requested changes, and the new shard delivered the expected lane reduction. Sharding low-Kubernetes/TLS could save another few minutes, but it adds compatibility jobs and should be evaluated separately after this stabilization work lands. Keep #13748 open for post-merge observation and the remaining product-valid connection-handling work rather than expanding this already-large PR.

manifests/kustomize/third-party/seaweedfs/base/seaweedfs/seaweedfs-networkpolicy.yaml
```diff
@@ -24,6 +24,12 @@ spec:
     - podSelector: {}
     ports:
     - port: 8333
+  - from:
```
juliusvonkohout, line 29: That is already covetered from the previous podselector

juliusvonkohout:
Please reduce the seaweedfs CPU request then to the previous value of 64m. So 64m and 256Mi

jeffspahr:
@juliusvonkohout Restored in 42dfc81015. I checked the earlier changes before reverting: #13686 moved the product request from 64m/256Mi to 128m/256Mi, while #13454 was a separate CI-only 250m/256Mi override. This restores only the product default to 64m CPU and 256Mi memory; the diagnostics and independent CI stability patch remain unchanged. The standalone product overlay renders the requested values.

manifests/kustomize/third-party/seaweedfs/base/seaweedfs/seaweedfs-networkpolicy.yaml
jeffspahr: The preceding `podSelector: {}` ingress rule is restricted by its `ports` block to 8333, so it does not cover IAM port 8111. Adding 8111 there would expose IAM to every same-namespace pod. I kept the separate profile-controller selector for least privilege and added an adjacent comment in 42dfc81015 to make the distinction explicit; the rendered multi-user overlay still has exactly one 8111 rule scoped to the profile controller.

PR #13743, opened 2026-07-16
fix(backend): add persistedFinalState label only after task details and run metrics are persisted

Sanchit2662:
## **Description**

While digging through the terminal report path I found that the persistedFinalState label gets added to the workflow CR too early, and it can silently lose the final task states and v1 run metrics of a run.

The label basically means "this workflow is fully persisted, don't report it again". Once it's on the CR, the persistence agent skips the workflow for the whole TTL (7 days by default), and if a report does come in after that, the API server just deletes the CR. The problem is that the label was being added inside ReportWorkflowResource, which runs before the report server writes the task details to the DB, and also before the agent reports run metrics. So the ordering was: label first, then tasks, then metrics.

That ordering breaks in a pretty normal way. Say the run row update succeeds and the label gets patched onto the CR, but then the task details write fails because of a transient DB error. The agent retries the report, but now Save() sees PersistedFinalState() == true and skips it. Nothing ever retries the tasks. The run shows Succeeded in the UI but the task rows stay frozen at whatever non-terminal state they had last. Same story for metrics: if the agent pod restarts between ReportWorkflow and ReportMetrics, those metrics are just gone. And after the TTL passes, the re-report path deletes the CR entirely, so at that point there's nothing left to recover from.

#13627 hardened this area (resourceVersion c stale reports) but that covered the "stalereport overwrites newer state" direction. This is the other direction, a failure after the label lands. I didn't find
an existing issue or PR for it, closest one  symptom but a different root cause.

The fix is basically just reordering things :

- Moved the label patch (plus the run-state  counters) out of ReportWorkflowResource into a new FinalizeReportedWorkflow method on the resource manager. The report handler calls it only after
reportTasksFromExecution succeeds. If tasks  unlabeled and the agent just retries thewhole report, which is what you'd want.
- On the agent side, for terminal workflows efore the workflow report. Transient metricfailures now block the report and get retried. If the metrics call comes back NotFound (run row doesn't exist yet, which happens on the very first report), it  report since that's what creates the run,and metrics get retried after. Permanent metric failures don't block persisting the run itself.
- For that to work I had to classify gRPC No1 as CUSTOM_CODE_NOT_FOUND in the pipelineclient instead of lumping it in with transient errors.

A nice side effect: the post-TTL re-report path now persists the tasks before returning the NotFound that triggers CR deletion, so runs that were already affectedal when the TTL re-report fires.

Tests: updated the four ReportWorkflowResourreport alone doesn't label anymore, thefinalize step does), added a handler-level test that checks the label actually lands on the CR after a full ReportWorkflow call, and added four saver teing cases (success, transient blocks thereport, NotFound falls through, permanent falls through). go test is green for apiserver/resource, apiserver/server and agent/persistence/..., gofmt and vet are

**Checklist:**


Sanchit2662:
@zazulam @jeffspahr , please review the fix whenever possible.

backend/src/agent/persistence/worker/workflow_saver.go
```diff
@@ -68,6 +68,35 @@ func (s *WorkflowSaver) Save(key string, namespace string, name string, nowEpoch
...
+		case metricsError == nil:
+			metricsReported = true
+		case util.HasCustomCode(metricsError, util.CUSTOM_CODE_NOT_FOUND):
+			// The run row may not exist yet (this can be the first report of
+			// this workflow). The workflow report below creates the run row.
+			// The API server defers the persistedFinalState label when a
+			// terminal report creates the run, returning a retryable error, so
+			// the retried Save reports metrics against the now-existing run
+			// before the workflow is finalized.
+		case util.HasCustomCode(metricsError, util.CUSTOM_CODE_PERMANENT):
```
jeffspahr, line 90: [P1] Preserve transient artifact-read failures before finalization

`ReportMetrics` can reach this branch with `CUSTOM_CODE_PERMANENT` even when reading the metrics artifact failed transiently. `artifactclient.ReadArtifact` returns `*artifactclient.Error` (for example, HTTP 500 is `ErrorCodeTransient`), but `MetricsReporter.aggregateErrors` only checks `util.HasCustomCode`, so that transient artifact error is rewrapped as `CUSTOM_CODE_PERMANENT`. This branch then reports and finalizes the workflow; the next `Save()` skips it because of `persistedFinalState`, silently losing the metrics. Please translate or recognize `artifactclient.ErrorCodeTransient` before this switch and add a direct workflow-saver regression test for a transient artifact-read failure.

backend/src/apiserver/server/report_server.go
```diff
@@ -87,6 +87,14 @@ func (s *BaseReportServer) reportWorkflow(ctx context.Context, workflow string)
 	if err != nil {
 		return nil, util.Wrap(err, "Failed to report task details")
 	}
+	// Add the persistedFinalState label only after the task details above have
+	// been persisted. The label makes the persistence agent skip re-reporting
+	// this workflow and lets the API server garbage collect it, so adding it
+	// before the task write completes would silently lose the run's final task
+	// states if the write fails or the request is interrupted.
+	if err := s.resourceManager.FinalizeReportedWorkflow(ctx, newExecSpec); err != nil {
```
jeffspahr, line 95: [P2] Cover the task-write failure before finalization

The new handler test verifies only the successful path. Please add a handler-level regression test where `reportTasksFromExecution` / `CreateOrUpdateTasks` fails and assert that `FinalizeReportedWorkflow` does not leave `persistedFinalState` on the workflow. That is the exact failure ordering this PR is intended to fix, and without this test the central invariant is only covered indirectly.

jeffspahr (changes requested):
Thanks for tackling this persistence-ordering bug. I found one blocking path where transient metric failures can still become unretryable, plus a missing regression test for the task-write failure that motivated the change.

backend/src/agent/persistence/worker/workflow_saver.go
Sanchit2662: Fixed in b4b36f04c

backend/src/apiserver/server/report_server.go
Sanchit2662: Added this too in b4b36f04c

Sanchit2662:
@jeffspahr , the ci has passed now.

Sanchit2662:
I've narrowed this PR quite a bit, so a heads-up on what changed and why.

I originally opened this to fix two things: the persistedFinalState label landing before task details were written, and before run metrics were reported. Since then #13986 replaced MLMD tracking with the native task and artifact APIs, and that removed workflow-derived task persistence entirely , reportWorkflow no longer loads the run or calls reportTasksFromExecution, and those functions are gone. So the task half of this PR is fixing a race that no longer exists.

What's still live on master is the metrics half. The persistence agent reports run metrics after the workflow report, and a successful terminal report labels the workflow, after which Save() skips it for the whole TTL. An agent restart between those two calls loses the metrics for good.

So this is now just that: report metrics before the workflow report for terminal workflows, classify gRPC NotFound from ReportRunMetrics apart from transient failures so the agent can tell "run row doesn't exist yet" from a real outage, and defer the label when the terminal report is what creates the run row , without that, the first report finalizes the workflow before any metrics can attach to it.

Dropped FinalizeReportedWorkflow and its call site: with no task write left in that path, there's nothing to sequence between the report and the label.

droctothorpe:
Thanks for narrowing this down to the still-live metrics path. I found one blocking issue:

Pre-reporting metrics in `workflow_saver.Save()` now happens *before* `ReportWorkflow()`, but the stale-report / retry-generation / ownership validation lives in `ReportWorkflowResource`, not in `ReportRunMetricsV1`.

That means a stale terminal workflow snapshot can now persist metrics before the API server gets a chance to reject or defer the workflow report. A bad case looks like this:
1. an old terminal workflow snapshot is still in the persistence queue,
2. the run is retried or otherwise guarded by the `ReportWorkflowResource` fences,
3. `Save()` reports metrics first,
4. only afterwards does `ReportWorkflow()` return the retryable stale/defer error.

Because `ReportRunMetricsV1` only authorizes by run ID and "first reporting wins" for `(run_id, node_id, name)`, those stale metrics can stick and block the real workflow's later metrics as duplicates.

So this fixes one ordering bug, but it also moves metric persistence ahead of the only place that validates whether the workflow snapshot is still the right one. I think this needs a fix that keeps metrics behind the same workflow-identity/staleness checks as terminal workflow reporting, plus a regression test for a stale terminal snapshot that is rejected/deferred by `ReportWorkflow()` but must not write metrics.

Sanchit2662:
You're right @droctothorpe, I confirmed it. ReportRunMetricsV1 only checks that the run exists, metrics are first-write-wins on (run_id, node_id, name), and the agent ignores DUP_REPORTING, so stale metrics would stick and the real ones would disappear silently.

Plan: put the workflow report back first so it runs all the fences, but have it defer the persistedFinalState label rather than add it. The agent then reports metrics against the validated run, and the retried report adds the label. Metrics stay behind the fences, label stays behind the metrics.

One thing I'd like your read on before I build it that the agent has to tell "accepted but deferred" from "rejected, write nothing", and terminalWorkflowReportDeferredError currently returns Unavailable, same as ~14 other returns in that path. Distinct gRPC code for the deferral, or something explicit in the response? Happy either way.

Issue #12654, opened 2026-01-14
[bug] Address GitHub API rate-limiting errors in CI


droctothorpe:
GHA workflows are periodically failing due to GitHub API rate limits. We need to investigate what root cause and remediate this problem. In general, we should try to reduce pressure on the GH APIs from our CI. We can, for example, only upload test results in the event that a test fails. 

### Labels
/area ci

---

Impacted by this bug? Give it a 👍.

Goku2099:
@aniketpati1121 
Switching from always() to failure() should definitely help reduce unnecessary artifact uploads and GitHub API usage.

One thought — for long-running jobs, would it make sense to also upload artifacts when a workflow is manually cancelled? (failure() || cancelled()), just in case we need logs for debugging.

Otherwise, the change makes sense and aligns well with the goal of reducing CI pressure.

aniketpati1121:
> [@aniketpati1121](https://github.com/aniketpati1121) Switching from always() to failure() should definitely help reduce unnecessary artifact uploads and GitHub API usage.
> 
> One thought — for long-running jobs, would it make sense to also upload artifacts when a workflow is manually cancelled? (failure() || cancelled()), just in case we need logs for debugging.
> 
> Otherwise, the change makes sense and aligns well with the goal of reducing CI pressure.

Good point handling cancelled() as well could be useful for debugging long-running jobs. Thanks for the suggestion!

jeffspahr:
closed by https://github.com/kubeflow/pipelines/pull/13761
/close

PR #13761, opened 2026-07-19
fix(ci): address GitHub API rate-limiting by conditionally uploading test results

prakharsingh-74:
fixes - #12654 

**Description**
We periodically run into GitHub API rate-limiting failures during CI runs due to the high volume of artifact uploads. To address this, this PR optimizes artifact management across all integration and end-to-end test lanes:

1. **Test-and-Report Composite Action (`.github/actions/test-and-report/action.yml`)**:
   - Modified `Upload HTML Report` to run only when tests fail (`steps.run-tests.outcome != 'success'`).
   - Modified the validation step `Mark Workflow failure if test reporting failed` to prevent skipped HTML report uploads from falsely failing successful workflow runs.

2. **Integration & E2E Workflows**:
   - Updated the final artifact collection steps from `always()` to `failure()` in five major workflows. This ensures test logs, kind cluster traces, and diagnostics are only uploaded when a test fails.
     - `.github/workflows/kfp-sdk-client-tests.yml`
     - `.github/workflows/e2e-test-frontend.yml`
     - `.github/workflows/integration-tests-v1.yml`
     - `.github/workflows/kfp-kubernetes-native-migration-tests.yaml`
     - `.github/workflows/legacy-v2-api-integration-tests.yml`

This change significantly reduces API traffic and pressure on GitHub's artifact storage services on successful runs, while keeping full diagnostic info available when runs fail.


prakharsingh-74:
@hbelmiro can you please review this PR?

jeffspahr:
Before treating this as a fix for #12654, I think we need to separate two different rate-limit paths.

The confirmed CI 429s we have recently investigated were ordinary GitHub REST/raw-content requests:

- #12698 and #12712 removed REST calls and paginated artifact *lookups*.
- #13695 and #13697 authenticated and retried the observed `raw.githubusercontent.com` 429s.

`actions/upload-artifact` does not use that `GITHUB_TOKEN` REST bucket. The artifact client uses `ACTIONS_RUNTIME_TOKEN` and `ACTIONS_RESULTS_URL`, uploads through a signed blob URL, and has separate 429 retry handling. Therefore, the recent authentication fixes neither fix nor obsolete a distinct Actions Results throttling problem—but we need evidence that such a problem is actually occurring here.

Issue #12654 proposed failure-only uploads as one possible mitigation, but it does not contain a failing upload step, endpoint, or run. I also checked the retained failed workflow runs from July 12–19 and found no artifact-upload or reporting-sentinel failure.

There is still a reasonable cleanup optimization here. On one recent green PR head, the artifacts affected by this change represented 64 objects totaling about 5.26 MB. That meaningfully reduces object count and UI clutter, although the storage reduction is small.

Could you please provide a run showing an `upload-artifact` / Actions Results 429? If there is no such run, I suggest reframing the PR as artifact-volume cleanup and removing the claim that high artifact volume caused the observed GitHub API rate limits or that this definitively fixes #12654.

If the optimization proceeds, it should also:

- preserve cancelled-run diagnostics with `failure() || cancelled()`;
- cover the still-`always()` webhook integration artifact upload; and
- avoid publishing an empty `HTML Report:` field on successful runs.

prakharsingh-74:
@jeffspahr and @droctothorpe can you please review?

prakharsingh-74:
> /ok-to-test /lgtm

All checks have been passed @droctothorpe sir please check it

.github/actions/test-and-report/action.yml
```diff
@@ -579,27 +579,38 @@ runs:
...
       continue-on-error: true
 
+    - name: Prepare custom data
+      id: custom-data-prep
+      if: always() && !cancelled()
+      shell: bash
+      run: |
+        CUSTOM_DATA=""
+        if [ -n "${{ steps.upload.outputs.artifact-url }}" ]; then
+          CUSTOM_DATA="{\"HTML Report\": \"${{ steps.upload.outputs.artifact-url }}\"}"
```
Priyanshu-u07, line 596: This line drops the escaping the old code relied on. Bash consumes the `\"` here, so the output becomes plain `{"HTML Report": "..."}`. But `junit-summary/action.yml` substitutes `${{ inputs.custom_data }}` directly into its bash script with unescaped quotes, that script fails with `syntax error in conditional expression`.

Since the upload only happens on failed runs, this breaks the `publish` step (and trips the reporting-failure guard) on every failing run, while staying invisible on green runs. The test summary and HTML report link are lost exactly when they're needed.

Keeping the escaping fixes it:

```suggestion
          CUSTOM_DATA="{\\\"HTML Report\\\": \\\"${{ steps.upload.outputs.artifact-url }}\\\"}"
```

prakharsingh-74:
@droctothorpe reminder

.github/actions/test-and-report/action.yml
```diff
@@ -579,27 +579,38 @@ runs:
...
       continue-on-error: true
 
+    - name: Prepare custom data
+      id: custom-data-prep
+      if: always() && !cancelled()
+      shell: bash
+      run: |
+        CUSTOM_DATA=""
+        if [ -n "${{ steps.upload.outputs.artifact-url }}" ]; then
+          CUSTOM_DATA="{\"\\\"HTML Report\\\"\": \"\\\"${{ steps.upload.outputs.artifact-url }}\\\"\"}"
```
jeffspahr: This escaping still breaks the nonempty artifact-link path. The assignment writes `{"\"HTML Report\"": "\"https://…\""}` to `CUSTOM_DATA`; when `junit-summary/action.yml` interpolates that value directly into its double-quoted Bash expression, Bash sees source like `if [[ -n "{"\"HTML Report\"": ...}" ]]` and fails with `syntax error in conditional expression`. Direct JSON decoding also produces a key and value that contain literal quote characters.

Because this path only runs after failed tests and a successful HTML upload, the green checks do not exercise it. Please pass `custom_data` through an environment variable in `junit-summary`, emit ordinary JSON such as `{"HTML Report":"https://…"}`, and add regression coverage for a nonempty artifact URL.

jeffspahr (changes requested):
The `failure() || cancelled()` workflow changes look correct and complete. One failed-run regression still needs fixing before merge.

Non-blocking: the PR description says five workflows, but the current diff changes six, including `.github/workflows/kfp-webhooks.yml`.

.github/actions/test-and-report/action.yml
prakharsingh-74: image

prakharsingh-74:
@jeffspahr reminder

.github/actions/junit-summary/junit_to_summary_test.py
```diff
@@ -0,0 +1,60 @@
...
+import os
+from pathlib import Path
+import tempfile
+import unittest
+from junit_to_summary import TestReport, TestSuite, TestCase, generate_markdown_summary
+
+
+class JunitToSummaryTest(unittest.TestCase):
+
+    def test_generate_markdown_summary_with_nonempty_artifact_url(self):
```
jeffspahr, line 46: This test bypasses the code path that failed: it constructs already-valid JSON, decodes it into a Python dictionary, and calls only `generate_markdown_summary`. Reverting the action to the broken direct `${{ inputs.custom_data }}` Bash interpolation would leave this test green. The new file is also not invoked by any CI test command (and its name is not matched by default `unittest` discovery). Please add CI-wired coverage that passes nonempty JSON through the composite action input → environment variable → CLI argument path, or extract and test that argument-building boundary directly.

jeffspahr (changes requested):
The functional env-var/JSON fix is sound, but the added test does not exercise that boundary or run in CI, so the regression remains unguarded.

.github/actions/junit-summary/junit_to_summary_test.py
prakharsingh-74: image

.github/actions/junit-summary/junit_to_summary.py
```diff
@@ -387,6 +384,14 @@ def main():
 
     # Expand file patterns and collect XML files
     xml_files = expand_file_patterns(args.xml_files)
+    if not xml_files:
```
jeffspahr: `--no-fail-on-test-failures` should suppress failures reported inside valid test results, not missing or mistyped XML inputs. Returning zero here produces no summary, leaves outputs unset, and prevents callers from detecting the reporting failure through the step outcome. Keep missing XML as an unconditional error; only suppress the final nonzero exit for failures/errors found in a successfully parsed report.

.github/actions/junit-summary/junit_to_summary_test.py
```diff
@@ -0,0 +1,210 @@
...
+            self.skipTest("bash shell integration test runs on Linux CI environments")
+
+        bash_executable = shutil.which("bash")
+        if not bash_executable:
+            self.skipTest("bash is not available in environment")
+
+        artifact_url = "https://github.com/kubeflow/pipelines/actions/runs/12345/artifacts/67890"
+        custom_data_input = f'{{"HTML Report": "{artifact_url}"}}'
+
+        bash_script = """
```
jeffspahr: This test copies the safe Bash block into a string instead of exercising the implementation in `action.yml`. Reverting the actual action to unsafe direct `${{ inputs.custom_data }}` interpolation would leave every new test green. Please exercise the real composite action in a workflow fixture, or extract the argument-building logic into a script shared by the action and this test.

jeffspahr (changes requested):
The original JSON/Bash fix now works, but two follow-up issues still need addressing.

Non-blocking: `git diff --check` reports an extra blank line at the end of `.github/workflows/ci-scripts-tests.yml`.

prakharsingh-74:
@zazulam and @jeffspahr can you review this PR?

.github/workflows/gh-workflow-approve.yml
```diff
@@ -75,32 +75,46 @@ jobs:
...
+                    newRunsFound = true;
+                  } catch (err) {
+                    core.error(`Failed to approve run ${run.id}: ${err.message}`);
+                  }
+                }
+              }
 
-              if (runs.length > 0) {
-                core.info(`Found ${runs.length} workflow run(s) to approve.`);
+              // If we have approved some runs and found no new runs in this check, we can safely exit.
```
jeffspahr: This exits after the first poll that approves nothing once any prior approval succeeded. For example, workflow A can register and be approved at t=0, the t=10 poll can be empty and break the loop, and workflow B can register at t=15 and remain `action_required`. It also stops early after a partial API failure when an earlier run succeeded. Poll for the full bounded window or require a deliberate multi-poll quiescence threshold before exiting.

.github/workflows/ci-scripts-tests.yml
```diff
@@ -32,3 +32,6 @@ jobs:
       - name: Run Python script unit tests
         working-directory: .github/resources/scripts
         run: python3 -m unittest discover -v -p '*_test.py'
+      - name: Run JUnit summary action unit tests
```
jeffspahr, line 46: This adds the JUnit-summary test command, but `on.pull_request.paths` does not include `.github/actions/junit-summary/**`. A PR modifying only `action.yml`, `parse_junit_summary.sh`, `junit_to_summary.py`, or these tests will therefore skip its own regression suite. Add that action directory to the path filter.

.github/actions/junit-summary/parse_junit_summary.sh
```diff
@@ -0,0 +1,58 @@
...
+if [[ -n "${OUTPUT_PATH:-}" ]]; then
+  args+=(--output "$OUTPUT_PATH")
+fi
+
+# XML files / targets specified as positional args or via XML_FILES environment variable
+xml_targets=()
+if [[ $# -gt 0 ]]; then
+  xml_targets=("$@")
+elif [[ -n "${XML_FILES:-}" ]]; then
+  read -r -a xml_targets <<< "$XML_FILES"
```
jeffspahr, line 50: Reading `XML_FILES` into an array preserves glob strings for Python instead of expanding them in the shell. Relative globs work, but absolute patterns such as `/tmp/reports/*.xml` now reach `Path(".").glob()` and raise `NotImplementedError`, even when files match. The action promises glob support; preserve absolute-pattern expansion and add shared-script coverage.

jeffspahr (changes requested):
The earlier missing-XML and shared-script findings are resolved. Three follow-up issues remain in the latest head.

prakharsingh-74:
@jeffspahr sir, all your follow up issues has been resolved, kindly review it again

.github/workflows/gh-workflow-approve.yml
```diff
@@ -75,32 +75,40 @@ jobs:
...
+                  try {
+                    await github.rest.actions.approveWorkflowRun({
+                      owner: context.repo.owner,
+                      repo: context.repo.repo,
+                      run_id: run.id,
+                    });
+                    approvedRunIds.add(run.id);
+                    totalApproved++;
+                    newRunsFound = true;
+                  } catch (err) {
```
jeffspahr, line 98: Approval errors are caught on every poll but never propagated. After seven failed attempts, the step still exits successfully and may log `Successfully approved ...` while another run remains `action_required`; if none succeed, it misleadingly reports that no runs were found. Track unresolved run IDs and fail the step after the final poll, or allow the final approval error to escape. The job-level `continue-on-error` can still control the workflow rollup.

jeffspahr (changes requested):
The previous polling, trigger-path, and absolute-glob findings are resolved. One approval-error reporting issue remains.

prakharsingh-74:
hi @jeffspahr sir, current CI checks are failing on the current run are due to standard backend integration test flakes on the master branch, can you please look into that and tell me what should i do now?

prakharsingh-74:
@jeffspahr sir, can you please check this PR now

jeffspahr (approved):
Re-reviewed the current head. The prior workflow-approval error reporting issue is resolved, the focused JUnit-summary tests and syntax checks pass, and the latest CI run is green. No remaining actionable findings.

PR #12848, opened 2026-02-18
feat(backend): Add backend support for cache_key

mnehete32:
**Description of your changes:**
Add backend support for cache_key for cache key customization.

**Fixes:** #11328.

**Checklist:**
- [x] You have [signed off your commits](https://www.kubeflow.org/docs/about/contributing/#sign-off-your-commits)
- [x] The title for your pull request (PR) should follow our title convention. [Learn more about the pull request title convention used in this repository](https://github.com/kubeflow/pipelines/blob/master/CONTRIBUTING.md#pull-request-title-convention).

PR #11466, opened 2024-12-13
feat: Introduce cache_key to sdk

zzmao:
**Description of your changes:**
feat: Introduce cache_key to sdk

**Checklist:**
- [ ] You have [signed off your commits](https://www.kubeflow.org/docs/about/contributing/#sign-off-your-commits)
- [ ] The title for your pull request (PR) should follow our title convention. [Learn more about the pull request title convention used in this repository](https://github.com/kubeflow/pipelines/blob/master/CONTRIBUTING.md#pull-request-title-convention).

PR #13762, opened 2026-07-19
feat(backend): support task-level custom cache key customization

### LINKED ISSUE #11328: [feature] Cache Key Customization
### Feature Area

/area backend
/area sdk 

Setting custom cache keys at the task level

Currently, the [cache key](https://github.com/kubeflow/pipelines/blob/master/api/v2alpha1/cache_key.proto) in KFP is generated internally, and users are unable to customize or set it.
However, some users have expressed the need for more granular control over caching behavior by setting custom cache keys at the task level.

N/A

---

Love this idea? Give it a 👍.

prakharsingh-74:
fixes #11328 

**Description**
This PR introduces support for task-level custom cache keys in the Kubeflow Pipelines backend driver.

- **Custom Cache Key Resolution**: Updated the backend execution driver (`cache.go`) to prioritize a user-defined `cache_key` if specified in `PipelineTaskSpec.CachingOptions`. When provided, the driver hashes the key using SHA-256 and uses it as the cache fingerprint, bypassing the default auto-generation logic.
- **Unit Testing**: Added test cases in `cache_test.go` (`Test_getFingerPrint` and `Test_getFingerPrintsAndID`) to ensure custom cache keys are resolved, hashed, and validated correctly.
- **Windows compatibility fix**: Replaced `filepath.Join` with `path.Join` for the container-bound `outputs.OutputFile` in `driver.go`. This resolves a test failure on Windows due to incorrect path separators (`\`) inside the container spec.

**Checklist:**

image


backend/src/v2/driver/cache.go
```diff
@@ -80,7 +82,21 @@ func reuseCachedOutputs(ctx context.Context, executorInput *pipelinespec.Executo
 // getFingerPrint generates a fingerprint for caching. The PVC names are included in the fingerprint since it's assumed
 // PVCs have side effects (e.g. files written for tasks later on in the run) on the execution. If the PVC names are
 // different, the execution shouldn't be reused for the cache.
+// If a custom cache key is set, all automatic inputs, image, command, arguments, and PVC names are ignored.
+// Instead, the custom key is combined with the component name (to avoid cross-component cache collisions) and hashed.
 func getFingerPrint(opts Options, executorInput *pipelinespec.ExecutorInput, cacheClient cacheutils.Client, pvcNames []string) (string, error) {
+	if opts.Task.GetCachingOptions() != nil && opts.Task.GetCachingOptions().GetCacheKey() != "" {
+		hash := sha256.New()
+<<<<<<< HEAD
```
jeffspahr: [P0] This unresolved merge-conflict marker is checked into the PR head, so this package is invalid Go and cannot compile. Please resolve all conflict blocks in both `cache.go` and `cache_test.go`; keep the component-reference branch here because it matches the local runner and prevents identical custom keys on different components from sharing cached outputs.

jeffspahr (changes requested):
The linked issue is still relevant, but specifically for remote KFP execution. The API field and SDK/compiler support landed in #11466, and local execution now honors custom keys, while `master`'s remote driver still always generates the ordinary fingerprint and never reads `CachingOptions.cache_key`. So backend support is still needed.

There is one blocking problem in the current head: unresolved merge-conflict markers remain in both `cache.go` and `cache_test.go`, and the driver package cannot compile. Please resolve every conflict consistently in favor of the component-scoped algorithm and fixtures; `SHA256(component_name + NUL + custom_key)` matches the local runner and avoids cross-component cache collisions.

Please also coordinate this work with the already-open backend implementation in #12848, since the two PRs substantially overlap.

Before merging, I recommend:

- Updating the public proto/SDK contract to explain that the supplied key is combined with component identity and that automatic inputs, image, command, arguments, and PVC names are ignored.
- Strengthening the unit test with identical component/key values but differing non-empty inputs, image, command/arguments, and PVCs, so it actually proves the override semantics. A remote cache-hit integration regression would provide the strongest coverage.

The expected component-scoped SHA fixtures and the `path.Join` change are otherwise correct.

prakharsingh-74:
@jeffspahr and @droctothorpe can you guys please review this PR now?

prakharsingh-74:
@droctothorpe and @jeffspahr reminder

jeffspahr (approved):
Re-reviewed at `1ea85c6b27bc010d5488ff01d914004c991ce485`. The previous blockers are resolved: all conflict markers are gone, the component-reference-scoped SHA-256 behavior matches local execution, the proto and primary DSL documentation describe the override semantics, and the expanded unit test proves that non-empty inputs, output definitions, image, command/arguments, and PVC names are ignored. The component-collision fixtures and `path.Join` change are also correct.

Focused verification passed: `go test ./backend/src/v2/driver -count=1`, `go vet ./backend/src/v2/driver`, Go formatting checks for the changed driver files, and Python syntax validation for `pipeline_task.py`. CI should still verify that regenerated protobuf output is reproducible.

Two non-blocking follow-ups:

- Please align the remaining public `cache_key` docstrings in `sdk/python/kfp/client/client.py` (and ideally `dsl/structures.py`) with the new component-reference scoping and override semantics; they still say the supplied value is used directly as the task cache key.
- A remote two-run cache-hit integration regression would provide stronger end-to-end coverage than the unit tests alone.

PR #12848 still overlaps this work, so maintainers should consolidate on one implementation.

prakharsingh-74:
@jeffspahr I've merged the master branch now, so lgtm label has been removed. Requesting you to add again

jeffspahr (approved):
Re-reviewed at `c8afb096720f7f4613e2609ffc30aea18465ecb8`. Since the previous approval, the only PR-authored delta is `d006a58e`, which updates the remaining public client and `TaskSpec.cache_key` docstrings; the other new commits merge current `master`. I found no cache implementation regression or reintroduced conflict markers. Focused driver tests pass, and the changed Go/Python sources pass formatting and syntax checks. The wording could say "component reference identity" more precisely, and remote integration coverage would still be useful; both are non-blocking. Approved.

api/v2alpha1/pipeline_spec.proto
```diff
@@ -505,7 +505,11 @@ message PipelineTaskSpec {
     // Whether or not to enable cache for this task. Defaults to false.
     bool enable_cache = 1;
     // Customized cache key for this task. If set, the cache_key will be used
-    // as the key for the task's cache.
+    // as the key for the task's cache. The customized cache key is combined
+    // with component reference identity (e.g. name) to prevent cross-component
```
jeffspahr, line 509: This source comment now says `component reference identity`, but `api/v2alpha1/go/pipelinespec/pipeline_spec.pb.go` still has the generated field comment `the component's identity`. The generated-file validation workflow runs Go proto generation followed by `make check-diff`, so this leaves the branch generation-dirty. Please run `make -C api golang` and commit the regenerated binding (or revert this source wording change).

jeffspahr (changes requested):
Re-reviewed at `fa557610ecb7970cac1038b07700dc622ae9c1b4`. The cache implementation is unchanged since the prior approval, and the refined SDK/local wording is consistent with the component-reference-scoped behavior. Remote integration coverage remains useful defense-in-depth but non-blocking: compiler and driver tests cover the new serialization/fingerprint branches, while the remote lookup path itself is unchanged.

There is one blocking generated-file consistency issue in the latest authored commit: the proto source comment changed, but the committed Go protobuf binding still contains the previous wording. The repository's `validate-generated-files` workflow regenerates Go protos and runs `make check-diff`, so the current head is not generation-clean. Please regenerate with `make -C api golang` and commit the result (or revert the proto wording change). Once source and generated output agree, I have no remaining blocker.

zazulam:
@prakharsingh-74 please regen the proto files

zazulam (approved):
/lgtm

PR #13764, opened 2026-07-20
fix: avoid duplicate processReportMetricResult call for failed metrics

### LINKED ISSUE #13763: processReportMetricResults calls processReportMetricResult twice, discarding the first result
### Environment
Not deployment-specific — this is a pure Go logic bug in the persistence agent's metrics reporting worker, reproducible regardless of deployment method.
*  KFP version:
Present on master as of July 2026 (backend/src/agent/persistence/worker/metrics_reporter.go)
*  KFP SDK version:
N/A — this is a backend-only bug, not SDK-related.

### Steps to reproduce
In `backend/src/agent/persistence/worker/metrics_reporter.go`, the function`processReportMetricResults` (lines 71–81) is meant to collect one error per failed metric-report result:

```go
func processReportMetricResults(
	reportMetricsResponse *api.ReportRunMetricsResponse) []error {
	errors := []error{}
	for _, result := range reportMetricsResponse.GetResults() {
		err := processReportMetricResult(result)
		if err != nil {
			errors = append(errors, processReportMetricResult(result))
		}
	}
	return errors
}
```
On line 77, instead of appending the already-computed `err` from line 75, thecode calls `processReportMetricResult(result)` a second time and appendsthat return value instead. The first call's result is discarded — it's only used for the nil check.
This is currently masked because `processReportMetricResult` is a pure function with no side effects, so both calls return equivalent errors. But it's semantically wrong: it does double the work unnecessarily, and `processReportMetricResult` ever gains a side effect (logging, a metricscounter, anything stateful), that side effect would silently fire twice per failed result.

### Expected result
Line 77 should append the already-computed `err` variable:

```go
if err != nil {
    errors = append(errors, err)
}
```
No behavior change is expected today given the current pure implementation,but this removes the redundant call and the correctness risk it introduces for any future change to `processReportMetricResult`.
### Materials and Reference
- File: `backend/src/agent/persistence/worker/metrics_reporter.go`, lines 71–81
- `processReportMetricResult` (lines 83–100) confirmed to have no side effects as of this writing, which is why the bug is currently unobservable externally.
- Happy to submit a fix + a regression test covering `processReportMetricResults` directly (currently only exercised indirectly via `ReportMetrics()` in the existing test file).

jahnavigajjala-3:
Fixes #13763
`processReportMetricResults` in `backend/src/agent/persistence/worker/metrics_reporter.go`computed `err` from `processReportMetricResult(result)` on line 75, then discardedit and called `processReportMetricResult(result)` a second time on line 77 when appending to the errors slice.

This is currently unobservable externally since `processReportMetricResult` is a pure function with no side effects — both calls return equivalent errors — but it does unnecessary redundant work and is a correctness risk for any future change
to `processReportMetricResult` that introduces side effects (e.g. logging, a metrics counter).

This PR:
- Fixes the duplicate call by appending the already-computed `err` variable instead of calling the function again.
- Adds a direct unit test for `processReportMetricResults` (previously only exercised indirectly through `ReportMetrics()` in the existing test suite), covering the case of multiple results with a mix of OK/INVALID_ARGUMENT/ INTERNAL_ERROR statuses, asserting the returned error count and messages.

### DIFF
--- a/backend/src/agent/persistence/worker/metrics_reporter.go
+++ b/backend/src/agent/persistence/worker/metrics_reporter.go
@@ -74,7 +74,7 @@ func processReportMetricResults(
 	for _, result := range reportMetricsResponse.GetResults() {
 		err := processReportMetricResult(result)
 		if err != nil {
-			errors = append(errors, processReportMetricResult(result))
+			errors = append(errors,err)
 		}
 	}
 	return errors
--- a/backend/src/agent/persistence/worker/metrics_reporter_test.go
+++ b/backend/src/agent/persistence/worker/metrics_reporter_test.go
@@ -569,3 +569,57 @@ func TestReportMetrics_Unauthorized(t *testing.T) {
 	assert.NotNil(t, err1)
 	assert.Contains(t, err1.Error(), "failed to read artifacts")
 }
+func TestProcessReportMetricResults_MixedStatuses(t *testing.T) {
+	results := []*api.ReportRunMetricsResponse_ReportRunMetricResult{
+		{
+			MetricNodeId: "node-1",
+			MetricName:   "accuracy",
+			Status:       api.ReportRunMetricsResponse_ReportRunMetricResult_OK,
+		},
+		{
+			MetricNodeId: "node-1",
+			MetricName:   "logloss",
+			Status:       api.ReportRunMetricsResponse_ReportRunMetricResult_INVALID_ARGUMENT,
+			Message:      "invalid metric value",
+		},
+		{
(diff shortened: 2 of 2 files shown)


droctothorpe:
Hi @jahnavigajjala-3, thank you for your PR! The linked issue (#13763) must be labeled `ready` before your contribution can be reviewed. Please check our [CONTRIBUTING.md](https://github.com/kubeflow/pipelines/blob/master/CONTRIBUTING.md) for details.

PR #13782, opened 2026-07-22
fix(sdk): convert v1 if placeholder else branch

### LINKED ISSUE #13781: [sdk] v1 YAML if placeholder keeps unconverted single else placeholder
### Environment

* **KFP SDK version:** current master
* **Area:** Python SDK v1 component YAML loading

### Steps to reproduce

Load or convert a v1 component YAML placeholder shaped like this:

```yaml
if:
  cond:
    isPresent: fallback
  then: present
  else:
    inputValue: fallback
```

The same conversion path is used by `kfp.dsl.placeholders.maybe_convert_v1_yaml_placeholder_to_v2_placeholder()` when v1 YAML component command or argument entries are loaded.

### Current behavior

The `then` branch is converted, but when `else` is a single placeholder object rather than a list, the converted value is discarded. The resulting v2 `IfPresent` structure can still contain the v1-style dictionary:

```json
{inputValue: fallback}
```

### Expected behavior

Both `then` and `else` should be converted consistently. In this example, the generated `Else` value should be:

```text
{{$.inputs.parameters['fallback']}}
```

### Root cause

In `sdk/python/kfp/dsl/placeholders.py`, the non-list `else` branch calls `maybe_convert_v1_yaml_placeholder_to_v2_placeholder(...)` but does not assign the return value back to `else_`.

Impacted by this bug? Give it a thumbs up.

hcthakur2004:
## Summary

- convert the single-value `else` branch in v1 `if` placeholders instead of discarding the converted value
- add a regression test for a v1 `if` placeholder whose `else` branch contains an `inputValue` placeholder

Fixes #13781

## Tests

- `python -m py_compile sdk/python/kfp/dsl/placeholders.py sdk/python/kfp/dsl/placeholders_test.py`
- lightweight direct conversion check for `maybe_convert_v1_yaml_placeholder_to_v2_placeholder`

Note: `python -m pytest -q sdk/python/kfp/dsl/placeholders_test.py -k v1_if_placeholder` could not collect in this sparse checkout because the generated `kfp.pipeline_spec.pipeline_spec_pb2` module is unavailable locally.


dobbydobap:
This is the right fix, and it's now the canonical one for #13991 as well, which was closed as a duplicate of #13781.

Your PR note says the test couldn't be collected locally because the generated `pipeline_spec_pb2` module wasn't available. I ran it in a clean container against the SDK from current master, and it does what you intended.

Without the change:

```
- , "Else": "{{$.inputs.parameters['fallback']}}"}}
+ , "Else": {"inputValue": "fallback"}}}
1 failed
```

With the one-line change applied:

```
1 passed
```

Worth noting the raw v1 dict survives all the way into the serialized IfPresent placeholder, not just into validation, so the test is a real regression guard rather than just covering the error path.

What's keeping this stuck looks procedural rather than technical: it's behind master and still carries `needs-ok-to-test`. A rebase would probably help move it.

droctothorpe:
Hi @hcthakur2004, thank you for your PR! The linked issue (#13781) must be labeled `ready` before your contribution can be reviewed. Please check our [CONTRIBUTING.md](https://github.com/kubeflow/pipelines/blob/master/CONTRIBUTING.md) for details.

PR #13790, opened 2026-07-23
fix(frontend): scope artifact node resolution to the current DAG

lifo9:

The run-details graph (v2) surfaces the **wrong output artifact** on artifact nodes that live inside sibling sub-DAGs.

When the same component runs in two sibling sub-DAGs that differ only by an input parameter (e.g. a reporting task instantiated once per `target_column`), both executions share the same `task_name` and each emits an output artifact with the same name. `getArtifactNodeKeyToArtifact` keyed a `Map` by `(task_name, artifact_name)` and built it last-write-wins across the **entire run**, so the two producers collapsed to a single entry. Clicking the artifact child node of one sub-DAG then showed the other sub-DAG's artifact — wrong **Upstream Task Name** and wrong content/URI. The correct artifact was only reachable via the task node's *Output Artifacts* tab, which resolves through the (correctly scoped) execution.

Execution and sub-DAG nodes are unaffected because they are already disambiguated by `parent_dag_id` (via `getExecutionsUnderDAG` + the `mlmdId` stamped onto the node). Artifact nodes were the one path that resolved purely by name.

### Fix

`getArtifactNodeKeyToArtifacts` now collects **all** candidate `LinkedArtifact`s per node key instead of keeping only the last, and `selectLinkedArtifactUnderDag` keeps only a producer that belongs to the DAG being viewed:

- **Direct artifacts** resolve to a producer whose `parent_dag_id` is the current DAG execution.
- **Sub-DAG output artifacts** are collected from an inner subtask, so they resolve against that sub-DAG's own execution, found via `getKeysFromArtifactNodeKey(elem.id)` + `getExecutionsUnderDAG(...)`. Sibling instances of the same sub-DAG component no longer collide on the inner `(producerSubtask, artifact)` key.
- **ParallelFor outputs** are collected from the per-iteration executions, so a producer under an iteration of that loop counts as well. The lowest iteration index wins, which keeps the choice stable across renders.
- **Otherwise the node stays unresolved.** A namesake from elsewhere in the run is not evidence of ownership, and neither is being the only candidate: a node whose own producer is still pending would otherwise borrow a sibling DAG's artifact. An absent child sub-DAG execution likewise resolves to nothing rather than to a run-wide match.

`updateFlowElementsState` stamps both the resolved artifact id and its **producing execution id** onto the node; `getNodeMlmdInfo` resolves the side panel by that pair. Matching on the execution id — not the artifact id alone — keeps the right producer when an artifact carries more than one OUTPUT event in a run (e.g. a same-run cache hit republishes the cached artifact under a new execution).

One adjacent crash is fixed alongside it: `getExecutionsUnderDAG` returns an empty list when a task ran elsewhere in the run but not in this DAG, which the execution-node guard treated as a hit, so `ExecutionHelpers.getName(executions[0])` (added in 7f60100) threw. Sibling sub-DAGs sharing task names are exactly when that happens; one of the regression tests below hits it.

### Scope

This targets `release-2.18`, the last release branch carrying the MLMD-based graph. On master, #13986 replaced MLMD tracking with the native task/artifact APIs, and that graph resolves artifacts through `parent_task_id`-scoped task records and iteration context — the scenarios above already pass there, so no native adaptation is needed.

## Testing

`DynamicFlow` regression tests, each verified to fail on the pre-fix code and pass after:

- **sibling sub-DAGs, same-named artifact** — two executions sharing `task_name` + artifact name but different `parent_dag_id`, the sibling's OUTPUT event listed last; asserts the node is stamped with the current-DAG artifact id **and** producer execution id, and that `getNodeMlmdInfo` returns the current-DAG producer.
- **pending local producer** — the producer in this DAG is still RUNNING and has no output, while a sibling DAG has a namesake artifact; asserts the node stays unresolved and the side panel returns nothing.
- **only foreign candidates** — two producers of the same name, both under other DAGs; asserts the node stays unresolved.
- **sibling sub-DAG output artifacts** — two sibling sub-DAG output nodes of the same component; asserts each resolves to its own inner producer, node and side panel alike.
- **sub-DAG not started** — the sub-DAG behind an output node has no execution; asserts the node stays unresolved rather than borrowing the sibling's inner artifact.
- **ParallelFor output** — two iterations produce the same artifact key with the second listed first; asserts the node resolves to the lowest iteration's producer. A companion test asserts a producer outside the loop leaves the node unresolved.
- **same-run cache hit** — one artifact with two OUTPUT events (different executions); asserts `getNodeMlmdInfo` returns the stamped producer, not merely the first event referencing the artifact id.

Also updated the existing `artifact found` test to carry the resolved `mlmdId` + `producerExecutionId` (as the node does after `updateFlowElementsState`). Locally: `DynamicFlow` suite 16/16, `src/lib/v2` + `RunDetailsV2` 56/56, `npm run typecheck`, `npm run lint:ui`, and `prettier --check` all clean. Against the unpatched `release-2.18` baseline, 9 of the 16 fail.

No existing issue tracks this; happy to open one if preferred.

## Checklist

- [x] The title follows the [conventional commits](https://www.conventionalcommits.org/) format.
- [x] Commit is signed off (DCO).
- [x] Unit tests added/updated for the changed behaviour.

### DIFF
--- a/frontend/src/lib/v2/DynamicFlow.ts
+++ b/frontend/src/lib/v2/DynamicFlow.ts
@@ -25,6 +25,7 @@ import {
   buildDag,
   buildGraphLayout,
   getArtifactNodeKey,
+  getKeysFromArtifactNodeKey,
   getIterationIdFromNodeKey,
   getTaskKeyFromNodeKey,
   isNode,
@@ -231,11 +232,12 @@ export function updateFlowElementsState(
   const taskNameToExecution = getTaskNameToExecution(executions);
   const executionIdToExectuion = getExectuionIdToExecution(executions);
   const artifactIdToArtifact = getArtifactIdToArtifact(artifacts);
-  const artifactNodeKeyToArtifact = getArtifactNodeKeyToArtifact(
+  const artifactNodeKeyToArtifacts = getArtifactNodeKeyToArtifacts(
     events,
     executionIdToExectuion,
     artifactIdToArtifact,
   );
+  const currentDagExecutionId = executionLayers[executionLayers.length - 1]?.getId();
 
   let flowGraph: PipelineFlowElement[] = [];
 
@@ -279,21 +281,40 @@ export function updateFlowElementsState(
         );
       }
     } else if (NodeTypeNames.ARTIFACT === elem.type) {
-      let linkedArtifact = artifactNodeKeyToArtifact.get(elem.id);
+      let linkedArtifact = selectLinkedArtifactInDag(
+        artifactNodeKeyToArtifacts.get(elem.id),
+        executionIdToExectuion,
+        currentDagExecutionId,
+      );
 
       // Detect whether Artifact is an output of SubDAG, if so, search its source artifact.
       let artifactData = elem.data as ArtifactFlowElementData;
       if (artifactData && artifactData.outputArtifactKey && artifactData.producerSubtask) {
-        // SubDAG output artifact has reference to inner subtask and artifact.
+        // A SubDAG output artifact is produced by an inner subtask one layer deeper, so
+        // scope the lookup to that child sub-DAG's execution (the artifact node's own task)
+        // rather than the current DAG; otherwise sibling instances of the same sub-DAG
+        // component collide on the inner (producerSubtask, artifact) key.
+        const [subDagTaskName] = getKeysFromArtifactNodeKey(elem.id);
+        const subDagExecutionId = getExecutionsUnderDAG(
+          taskNameToExecution,
+          subDagTaskName,
+          executionLayers,
+        )?.[0]?.getId();
         const subArtifactKey = getArtifactNodeKey(
           artifactData.producerSubtask,
           artifactData.outputArtifactKey,
         );
-        linkedArtifact = artifactNodeKeyToArtifact.get(subArtifactKey);
+        linkedArtifact = selectLinkedArtifactInDag(
+          artifactNodeKeyToArtifacts.get(subArtifactKey),
+          executionIdToExectuion,
+          subDagExecutionId,
+        );
       }
 
       (updatedElem.data as ArtifactFlowElementData).state = linkedArtifact?.artifact?.getState();
       (updatedElem.data as ArtifactFlowElementData).mlmdId = linkedArtifact?.artifact?.getId();
+      (updatedElem.data as ArtifactFlowElementData).producerExecutionId =
+        linkedArtifact?.event.getExecutionId();
     } else if (NodeTypeNames.SUB_DAG === elem.type) {
       // TODO: Update sub-dag state based on future design.
       const executions = getExecutionsUnderDAG(
@@ -368,11 +389,6 @@ export function getNodeMlmdInfo(
   const taskNameToExecution = getTaskNameToExecution(executions);
   const executionIdToExectuion = getExectuionIdToExecution(executions);
   const artifactIdToArtifact = getArtifactIdToArtifact(artifacts);
-  const artifactNodeKeyToArtifact = getArtifactNodeKeyToArtifact(
-    events,
-    executionIdToExectuion,
-    artifactIdToArtifact,
-  );
 
   if (NodeTypeNames.EXECUTION === elem.type) {
     const taskLabel = getTaskLabelByPipelineFlowElement(elem);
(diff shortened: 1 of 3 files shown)


frontend/src/lib/v2/DynamicFlow.ts
```diff
@@ -473,9 +504,39 @@ function getArtifactNodeKeyToArtifact(
...
+function selectLinkedArtifactInDag(
+  candidateLinkedArtifacts: LinkedArtifact[] | undefined,
+  executionIdToExecution: Map<number, Execution>,
+  dagExecutionId: number | undefined,
+): LinkedArtifact | undefined {
+  if (!candidateLinkedArtifacts || candidateLinkedArtifacts.length === 0) {
+    return undefined;
+  }
+  if (candidateLinkedArtifacts.length === 1 || dagExecutionId === undefined) {
+    return candidateLinkedArtifacts[0];
```
jeffspahr: [P2] Validate DAG ownership before selecting even a single candidate

This shortcut still displays a sibling DAG's artifact when the local producer has not emitted output. Reproduced through `updateFlowElementsState`: active DAG ID 1 has a RUNNING `report` execution with no output; another `report` under DAG 99 emits artifact 100. The local `artifact.report.output` node gets `mlmdId=100`, although it should remain unresolved. Candidate count must not bypass the parent-DAG check. The `artifactInDag ?? candidateLinkedArtifacts[0]` fallback below also selects a foreign producer when multiple candidates exist but none match. Please resolve valid loop/child-DAG ancestry explicitly and return undefined when no scoped producer exists; cover the single- and multiple-foreign-candidate cases plus missing child-DAG context in tests.

jeffspahr (changes requested):
Retargeted this PR to `release-2.18`, where the MLMD-based graph code remains in use. Native master resolves artifacts through parent-task/iteration-scoped task records, and the original scenarios pass there; no native adaptation of this patch is needed.

This is a useful 2.18 fix, but the remaining fallback behavior needs correction before merge.

### Required changes

- In `selectLinkedArtifactInDag`, validate producer scope even when there is only one candidate. A sole artifact elsewhere in the run is not evidence that it belongs to this DAG.
- Remove the arbitrary first-candidate fallback when no producer matches. Resolve supported ParallelFor/nested-DAG cases through explicit iteration/ancestry context; if ownership cannot be established, leave the node unresolved. An absent child-DAG execution must not turn into a run-wide artifact selection.
- Add regressions for a pending local producer with no output and a sibling with output, multiple foreign candidates with no local match, an absent local child-DAG producer, and loop-iteration selection. Verify both node identity and side-panel resolution, retaining the existing same-run cache and sibling sub-DAG coverage.

### Verification supporting this review

At head `ddf184998608525ad0abe9a7d89ed0c4a947fcf5`, the patch applies cleanly to release-2.18 snapshot `75f3f4bbd9affbe6c3c80d6be3465ddeeb90face`. Its DynamicFlow suite passes 11/11; against the release baseline, four of those tests fail. An additional missing-output regression still fails with the patch: a local RUNNING `report` execution has no output, but the local artifact node receives artifact ID 100 from a sibling DAG instead of remaining unresolved. The inline comment identifies the shortcut responsible; the final fallback has the same scope problem for multiple candidates.

For comparison, native master `d8e0634a46b3b35fd4a0f1cf5399e9c4d8d418a0` passes 28 existing plus eight targeted graph tests, including missing-output and loop cases. Removing native producer scoping makes four targeted tests fail.

These are unit-level checks of production graph functions using synthetic task/MLMD fixtures, not a browser/cluster end-to-end test or full frontend build. The additional investigation fixtures are not committed to this PR.

lifo9:
> Retargeted this PR to `release-2.18`...

Thanks for the review @jeffspahr!

I rebased to `release-2.18` and addressed all points.

The fix is no longer needed on `master`, but `2.18` still resolves artifacts through MLMD, and we'd rather not wait for the release after it to upgrade.

frontend/src/lib/v2/DynamicFlow.ts
lifo9: Fixed, the parent DAG check now runs for every candidate, the first-candidate fallback was removed and the loop producers are matched explicitly one layer deep.

jeffspahr (commented):
Re-reviewed at `14f78189ac3112152cf4de97ca51bfdc4173973d`. No actionable findings. The latest revision addresses the previous review's ownership and fallback concerns, including pending or absent local producers, foreign candidates, loop iteration selection, and same-run cache-hit producer identity.

Verification:
- All 56 graph/run-details tests pass.
- Nine regression tests fail against the pre-fix implementation.
- Typecheck, UI lint, and changed-file formatting pass.
- DAG and loop ownership checks agree with the backend metadata hierarchy.

Verification was limited to targeted automated tests and code review; no browser/cluster end-to-end test or full frontend build was run.

PR #13791, opened 2026-07-23
fix(backend): validate driver pod configuration from one read of one source

thc1006:

## Summary

- **behavior change**: a JSON object written directly in `config.json` is now refused at startup, with an error naming the string form to write instead. The ConfigMap path, which every standard install uses, is not affected
- read and parse the configuration once, so the map that was checked is the map that reaches the cache
- start the configuration watcher after startup finishes rather than in the middle of it, register its callback before starting it, and stop the callback re-reading a file the watcher has already read. This orders startup only; reads that happen while serving still overlap with the watcher
- reject a top level JSON `null` on the string path, and record in a test that a bare `null` in the configuration file reads as an absent key
- validate annotations with the validator Kubernetes itself applies, which stops refusing keys Kubernetes accepts and brings the total size limit that was not checked at all
- say in the README that the configuration file supplies the built in default rather than the setting an operator changes, and pin that precedence in a test
- document the Argo kill command an injected Istio proxy needs, and the compiler annotation that sits beside the injection label
- correct the note about restarting the API server, which described the opposite of what happens to existing recurring runs
- list the two variables in `AGENTS.md` and refresh its date, as its maintenance section asks for when environment variables change

## Why

#12417 added driver pod labels and annotations. The review that followed turned up several corrections. Only the first changes what an existing configuration does, and it affects the configuration file alone.

That form cannot be checked, which is why it is refused rather than validated more carefully. The configuration loader lowers the keys of a JSON object while reading the file, before any of this code runs. Two keys differing only in case are merged into one and a value is dropped, so a value this configuration would refuse can vanish before it is ever seen. Where both keys need lowering, which one survives follows map iteration order, so the same file starts the API server on one restart and stops it on the next. Both outcomes are covered by tests. Refusing the shape and naming the string form is the only answer that keeps the promise that every value is checked, and the string form is what the ConfigMap already uses, so the two sources finally parse the same way. Anyone affected has built a custom image and can quote the object.

The value is also read once now instead of twice, once to check it and again to convert it, which had left a window for the watcher to reload in between and cache something no check had seen. For the same reason `viper.WatchConfig` now runs after startup finishes rather than in the middle of it. Viper is not safe for concurrent reads and writes, the watcher reloads on its own goroutine, and the race detector reports that overlap both for the driver configuration and for the cluster domain the proxy configuration reads, which sat after the watcher until now. Moving the watcher rather than each reader covers whatever is added to startup later.

Two smaller things in the same block go with it. The change callback is registered before the watcher rather than after: `WatchConfig` only returns once its watcher is live, so registering afterwards left a gap in which an event found no callback and skipped it, on a field written and read without synchronization. Viper's own example registers first. The callback also no longer re-reads the file, because the watcher reads it before calling in, so reading again replaced the configuration a second time for one event and could validate a different version than the one just loaded.

That fixes the ordering of startup and nothing more, and the code says so where the watcher is started. Configuration is also read while serving, when a run is created for instance, and those reads still overlap with the watcher. Making reloading itself safe needs either synchronization around every configuration read or no watcher at all, which is a decision for whoever owns that layer rather than something to settle here.

On the string path a bare `null` was getting through, since it unmarshals into a nil map without an error and left the loop over its entries with nothing to walk. A configuration file key whose value is a bare `null` still reads as an absent key, because Viper reports it that way and offers no means to tell the two apart, `InConfig` included, and the shipped default depends on exactly that. A test records the boundary rather than putting a second reader of the file beside Viper for one input.

Annotation validation was written by hand and refused keys Kubernetes accepts, since Kubernetes lowers an annotation key before checking its syntax. Calling `apivalidation.ValidateAnnotations` fixes that, drops the duplicated rule, and brings the total size limit that was missing, which had let an oversized value through until Argo asked Kubernetes to create the pod. Labels keep their own validation, since Kubernetes does not lower label keys. The startup check measures the configured annotations alone, and the test at the boundary says so, because the pod carries compiler added annotations too and a value near that size could not be delivered anyway: each setting arrives as one environment variable holding its whole JSON document, and Linux caps a single environment string at 32 memory pages, around 128 KiB on the usual 4 KiB page size. The README now mentions that, since a value over the cap stops the API server from starting at all.

The README also says plainly what the configuration file is for. The Deployment always passes both keys in from the ConfigMap, and an environment variable takes precedence over the file even when it is empty, so on a standard install the file supplies the built in default and the ConfigMap is what an operator controls. A test pins that precedence. Removing the two empty defaults from the shipped ConfigMap would let the file apply again, and I have left that decision to the manifest owners rather than changing what was just reviewed.

Two things about sidecar injection were missing. Argo shuts an injected sidecar down with `/bin/sh -c 'kill 1'`, which the Istio proxy ignores, so the proxy outlives the driver and the pod stays in `Running`; Argo's per container kill command handles it, and the default install already grants the `pods/exec` that needs. Separately the compiler annotates every template with `sidecar.istio.io/inject: "false"`, which lands beside the injection label without disabling it.

The note about restarting the API server was wrong in the other direction and is corrected, though not as simply as either version suggested, since each API server reads the configuration once as it starts and what a run gets depends on which one compiles it. A recurring run that did not pin its pipeline is compiled through the runs API when it triggers, by whichever API server takes that call, which is the new configuration once the restart has finished but not necessarily during it, because the default rollout brings a replacement up before retiring the one it replaces and both serve for a short window. A pinned one is recompiled by the reconciliation an API server runs once as it starts, and that reconciliation stops at the first recurring run it cannot process and is not retried, so the ones after a failure keep their previous specification until the next restart. The README says all of this rather than promising that every recurring run updates.

Both injection claims were checked on a cluster rather than read off the source. With Argo, a pod built from a template carrying the configured labels and annotations held all of them, the kill command included and its escaping intact. With Istio, a pod holding the label `true` and the annotation `false` was injected, and the reverse combination was not. The environment variable limit was reproduced separately. `golangci-lint` v2.10.1 and the race detector were run in a Linux container and report nothing in the changed files.

**Checklist:**

### DIFF
--- a/backend/README.md
+++ b/backend/README.md
@@ -71,15 +71,19 @@ The API server supports configuring custom labels and annotations for driver pod
 **Configuration via config.json:**
 ```json
 {
-  "DRIVER_POD_LABELS": {
-    "sidecar.istio.io/inject": "true",
-    "app": "ml-pipeline-driver"
-  },
-  "DRIVER_POD_ANNOTATIONS": {
-    "proxy.istio.io/config": "{\"holdApplicationUntilProxyStarts\":true}"
-  }
+  "DRIVER_POD_LABELS": "{\"sidecar.istio.io/inject\":\"true\",\"app\":\"ml-pipeline-driver\"}",
+  "DRIVER_POD_ANNOTATIONS": "{\"proxy.istio.io/config\":\"{\\\"holdApplicationUntilProxyStarts\\\":true}\"}"
 }
 ```
+
+Write each value as a JSON string, not as a JSON object. A JSON object is refused at startup,
+because the configuration loader folds its keys to lower case and merges keys that differ only
+in case.
+
+This file is baked into the image and supplies the built in default. On a standard install the
+Deployment always passes both keys in from the ConfigMap as environment variables, and those
+take precedence over this file even when empty, so use the ConfigMap below unless you build
+your own image.
 
 **Configuration via Kubernetes ConfigMap (`pipeline-install-config`):**
 ```yaml
@@ -87,9 +91,26 @@ DRIVER_POD_LABELS: '{"sidecar.istio.io/inject":"true"}'
 DRIVER_POD_ANNOTATIONS: '{"proxy.istio.io/config":"{\"holdApplicationUntilProxyStarts\":true}"}'
 ```
 
-When configured via the ConfigMap, every value must be a JSON string. Write `"true"` rather than the bare boolean `true`, and never `null`. The configuration is validated at API server startup: malformed JSON, any value that is not a JSON string, and label keys, label values, or annotation keys that Kubernetes would reject all cause startup to fail with a descriptive error rather than being silently ignored. Annotation values are free form, so a value such as an inline JSON document is accepted.
+Every entry must have a string value: write `"true"` rather than the bare boolean `true`, and never `null`. The configuration is validated at API server startup, so malformed JSON, a value that is not a string, and label keys, label values or annotation keys that Kubernetes would reject all fail startup with a descriptive error rather than being silently ignored. Annotation values are free form, so an inline JSON document is accepted. Annotations are also held to the 256 KiB Kubernetes allows per object, and each setting travels as one environment variable, which Linux caps at 32 memory pages (around 128 KiB with 4 KiB pages).
 
-Note: Labels and annotations with the prefix `pipelines.kubeflow.org/` are reserved and will be filtered out to prevent overriding system metadata. Changes to driver pod configuration require an API server restart.
+**Terminating an injected Istio sidecar:**
+
+Argo shuts an injected sidecar down with `/bin/sh -c 'kill 1'`, which the Istio proxy does not
+act on, so the proxy outlives the driver and the pod stays in `Running`. Argo's per container
+kill command handles it, and the default install already grants the `pods/exec` permission it
+needs through the `pipeline-runner` Role. Grant `pods/exec` too if the driver runs under a
+service account with narrower permissions.
+
+```yaml
+DRIVER_POD_LABELS: '{"sidecar.istio.io/inject":"true"}'
+DRIVER_POD_ANNOTATIONS: '{"proxy.istio.io/config":"{\"holdApplicationUntilProxyStarts\":true}","workflows.argoproj.io/kill-cmd-istio-proxy":"[\"pilot-agent\", \"request\", \"POST\", \"quitquitquit\"]"}'
+```
+
+The compiler annotates every template with `sidecar.istio.io/inject: "false"`, which does not disable the label above, since Istio prefers the label and treats the annotation form as deprecated. See the Argo Workflows [sidecar injection](https://argo-workflows.readthedocs.io/en/latest/sidecar-injection/) guide for the full description of the problem.
+
+Note: Labels and annotations with the prefix `pipelines.kubeflow.org/` are reserved and will be filtered out to prevent overriding system metadata.
+
+Changes require an API server restart, and each API server reads the configuration once as it starts. An unpinned recurring run picks the change up when it next triggers, once the restart has finished; during a rolling restart both the old and new API server serve for a short window. A pinned recurring run is recompiled by the reconciliation that runs at startup, which stops at the first recurring run it cannot process and is not retried, so the ones after a failure keep their previous specification until the next restart.
 
 ## API Server Development
 
--- a/backend/README.md
+++ b/backend/README.md
@@ -71,15 +71,12 @@ The API server supports configuring custom labels and annotations for driver pod
 **Configuration via config.json:**
 ```json
 {
-  "DRIVER_POD_LABELS": {
-    "sidecar.istio.io/inject": "true",
-    "app": "ml-pipeline-driver"
-  },
-  "DRIVER_POD_ANNOTATIONS": {
-    "proxy.istio.io/config": "{\"holdApplicationUntilProxyStarts\":true}"
-  }
+  "DRIVER_POD_LABELS": "{\"sidecar.istio.io/inject\":\"true\",\"app\":\"ml-pipeline-driver\"}",
+  "DRIVER_POD_ANNOTATIONS": "{\"proxy.istio.io/config\":\"{\\\"holdApplicationUntilProxyStarts\\\":true}\"}"
 }
 ```
+
+Write each value as a JSON string, not a JSON object (rejected at startup). The ConfigMap takes precedence over this file.
 
 **Configuration via Kubernetes ConfigMap (`pipeline-install-config`):**
 ```yaml
@@ -87,9 +84,24 @@ DRIVER_POD_LABELS: '{"sidecar.istio.io/inject":"true"}'
 DRIVER_POD_ANNOTATIONS: '{"proxy.istio.io/config":"{\"holdApplicationUntilProxyStarts\":true}"}'
 ```
 
-When configured via the ConfigMap, every value must be a JSON string. Write `"true"` rather than the bare boolean `true`, and never `null`. The configuration is validated at API server startup: malformed JSON, any value that is not a JSON string, and label keys, label values, or annotation keys that Kubernetes would reject all cause startup to fail with a descriptive error rather than being silently ignored. Annotation values are free form, so a value such as an inline JSON document is accepted.
+Every entry must have a string value: write `"true"` rather than the bare boolean `true`, and never `null`. The configuration is validated at API server startup, so malformed JSON, a value that is not a string, and label keys, label values or annotation keys that Kubernetes would reject all fail startup with a descriptive error rather than being silently ignored. Annotation values are free form, so an inline JSON document is accepted. Annotations are also held to the 256 KiB limit Kubernetes allows per object.
 
-Note: Labels and annotations with the prefix `pipelines.kubeflow.org/` are reserved and will be filtered out to prevent overriding system metadata. Changes to driver pod configuration require an API server restart.
+**Terminating an injected Istio sidecar:**
+
+Argo's default sidecar kill command leaves the Istio proxy running, so the pod stays in
+`Running`. Argo's per container kill command handles it and needs `pods/exec`, which the
+default `pipeline-runner` Role already grants.
+
+```yaml
+DRIVER_POD_LABELS: '{"sidecar.istio.io/inject":"true"}'
+DRIVER_POD_ANNOTATIONS: '{"proxy.istio.io/config":"{\"holdApplicationUntilProxyStarts\":true}","workflows.argoproj.io/kill-cmd-istio-proxy":"[\"pilot-agent\", \"request\", \"POST\", \"quitquitquit\"]"}'
+```
+
+The compiler's `sidecar.istio.io/inject: "false"` annotation does not disable the label above, since Istio prefers the label. See the Argo Workflows [sidecar injection](https://argo-workflows.readthedocs.io/en/latest/sidecar-injection/) guide.
+
+Note: Labels and annotations with the prefix `pipelines.kubeflow.org/` are reserved and will be filtered out to prevent overriding system metadata.
+
+Changes require an API server restart.
 
 ## API Server Development
 
--- a/backend/src/apiserver/common/driver_config.go
+++ b/backend/src/apiserver/common/driver_config.go
@@ -23,7 +23,9 @@ import (
 
 	"github.com/golang/glog"
 	"github.com/spf13/viper"
+	apivalidation "k8s.io/apimachinery/pkg/api/validation"
 	"k8s.io/apimachinery/pkg/util/validation"
+	"k8s.io/apimachinery/pkg/util/validation/field"
 )
 
 const (
@@ -88,20 +90,22 @@ func InitDriverPodConfig() error {
 // construction separate from the cache swap guarantees the update is applied
 // completely or not at all.
 func newDriverPodConfig() (*DriverPodConfig, error) {
-	if err := validateMapConfig(DriverPodLabels); err != nil {
+	rawLabels, err := parseMapConfig(DriverPodLabels)
+	if err != nil {
 		return nil, err
 	}
-	if err := validateMapConfig(DriverPodAnnotations); err != nil {
+	rawAnnotations, err := parseMapConfig(DriverPodAnnotations)
+	if err != nil {
 		return nil, err
 	}
 
-	labels := filterReservedEntries(GetMapConfig(DriverPodLabels))
+	labels := filterReservedEntries(rawLabels)
 	if err := validateLabels(labels); err != nil {
 		return nil, err
 	}
 
-	annotations := filterReservedEntries(GetMapConfig(DriverPodAnnotations))
-	if err := validateAnnotationKeys(annotations); err != nil {
+	annotations := filterReservedEntries(rawAnnotations)
+	if err := validateAnnotations(annotations); err != nil {
 		return nil, err
 	}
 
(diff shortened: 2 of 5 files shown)


thc1006:
@hbelmiro this is the follow up to the review comments left on #12417 after it merged, together with a few things I ran into while checking them.

One change is worth knowing before any of it runs. A JSON object written directly in config.json is now refused at startup instead of accepted, because the configuration loader lowers that object's keys while reading the file, so two keys differing only in case get merged and one of the values disappears before anything can look at it. The ConfigMap path that standard installs use is untouched, and the error names the string form to write instead. The rest is the null handling, the annotation validator, the order the config watcher starts in, and documentation that had the behavior of existing recurring runs backwards.

Whenever it suits you, an ok-to-test would let the suite run against this. You gave the original change its ok-to-test and its review, and I am grateful for both.

backend/src/apiserver/common/driver_config_test.go
```diff
@@ -452,6 +644,203 @@ func TestInitDriverPodConfig_IstioMetadataAccepted(t *testing.T) {
...
+// Where a single key needs lowering the outcome is fixed and the refused value simply vanishes;
+// where both need it the survivor follows map iteration order, so the same file would start the
+// API server on one restart and stop it on the next. Neither is visible to anything downstream,
+// which is why the shape is refused rather than validated.
+// A second shape, two keys both needing to be lowered, was measured while deciding this and
+// produced either survivor across repeated reads, since map iteration order picks the winner.
+// It is left out of the assertions here because a test that depends on that order can only be
+// flaky, and the case below already shows a refused value being merged away.
+func TestInitDriverPodConfig_ConfigFileObjectHidesValues(t *testing.T) {
+	const runs = 40
```
hbelmiro: 40 runs? Does it need to be that much?

backend/src/apiserver/common/driver_config.go
```diff
@@ -126,45 +130,96 @@ func validateLabels(labels map[string]string) error {
...
+// particular is not the one used for labels: an annotation key is lowered before its syntax is
+// checked, so Example.com/name is a valid annotation key even though the same key would be
+// refused as a label. Only the syntax check is case insensitive; the keys themselves keep the
+// case they were written with.
+//
+// The size is worth checking at startup because Kubernetes only applies it when the pod is
+// created, which is long after the API server has started and the workflow has compiled. What is
+// measured here is the configured annotations alone. The driver pod also carries annotations
+// added by the compiler and possibly by admission, so passing this check bounds the operator's
+// own input rather than guaranteeing the pod will be accepted.
```
hbelmiro: I think we can be more succint here:

```suggestion
// validateAnnotations uses the Kubernetes annotation validator, which lowers keys before
// checking syntax (so Example.com/Name is valid unlike for labels) and enforces the 256 KiB
// total size limit. Only the configured annotations are measured; compiler-added ones are not.
```

backend/src/apiserver/common/driver_config.go
```diff
@@ -126,45 +130,96 @@ func validateLabels(labels map[string]string) error {
...
+// something that was never checked.
+//
+// A value written as a JSON object in the configuration file is refused rather than parsed. The
+// configuration loader lowers the keys of such an object while reading the file, before any of
+// this code runs, and that is destructive in a way nothing downstream can repair. Two keys
+// differing only in case collapse into one, so a value this function would refuse can be
+// discarded before it is ever seen, and when both keys need lowering the survivor depends on map
+// iteration order, which makes the same file start the API server on one restart and fail on the
+// next. Writing the object as a JSON string, the way the ConfigMap does, keeps the keys exactly
+// as they were written and gives every value to the check below.
```
hbelmiro:
```suggestion
// parseMapConfig reads and parses the configured value in one step so the validated map is
// the one cached. A JSON object in the config file is refused because Viper lowercases its
// keys, silently merging case-differing entries; the string form keeps keys as written.
```

backend/src/apiserver/common/driver_config.go
```diff
@@ -126,45 +130,96 @@ func validateLabels(labels map[string]string) error {
...
+func parseMapConfig(name string) (map[string]string, error) {
+	// One read, deliberately. Asking IsSet first and fetching afterwards would be two, and the
+	// nil case below already covers a key that is not set, since IsSet is defined as the same
+	// lookup returning something.
```
hbelmiro: This comment is not needed.
```suggestion
```

backend/src/apiserver/common/driver_config.go
```diff
@@ -126,45 +130,96 @@ func validateLabels(labels map[string]string) error {
...
+	switch value := viper.Get(name).(type) {
+	case nil:
+		// Not configured. A configuration file value of null reports as nothing here too, and
+		// Viper offers no way to tell that apart from an absent key.
```
hbelmiro: Not needed.
```suggestion
```

backend/src/apiserver/common/driver_config.go
```diff
@@ -126,45 +130,96 @@ func validateLabels(labels map[string]string) error {
...
+		return nil, nil
+	case string:
+		return parseMapConfigString(name, value)
+	case map[string]string:
+		// Only reachable when the value is set programmatically, which a configuration file
+		// cannot do, so the keys are as written and the values are already strings.
```
hbelmiro: Not needed.
```suggestion
```

backend/src/apiserver/common/driver_config.go
```diff
@@ -126,45 +130,96 @@ func validateLabels(labels map[string]string) error {
...
+		return copyMap(value), nil
```
hbelmiro: `copyMap(value)` is redundant here. The caller `newDriverPodConfig` passes the result to `filterReservedEntries`, which already creates a new map.

```suggestion
		return value, nil
```

backend/src/apiserver/common/driver_config.go
```diff
@@ -126,45 +130,96 @@ func validateLabels(labels map[string]string) error {
...
+	case map[string]interface{}:
```
hbelmiro: Nit:
```suggestion
	case map[string]any:
```

backend/src/apiserver/common/driver_config.go
```diff
@@ -126,45 +130,96 @@ func validateLabels(labels map[string]string) error {
...
+			return nil, fmt.Errorf("%s has a value that is not a JSON string for key %q; every value must be quoted, for example \"true\" rather than true or null (raw value: %q)", name, k, raw)
 		}
+		parsed[k] = s
 	}
-	return nil
+	if len(parsed) == 0 {
+		return nil, nil
+	}
+	return parsed, nil
 }
```
hbelmiro: This is overly complex. This function just converts a JSON to a map but it's hard to figure out that from it's name, params and comment.
`name` is used only for the error message. We should return only an error saying that the value couldn't be parsed and the caller should be responsible for wrapping that error with an error that says that `name` could not be parsed.




```suggestion
func jsonToMap(value string) (map[string]string, error) {
	raw := strings.TrimSpace(value)
	if raw == "" {
		return nil, nil
	}

	var parsed map[string]string
	if err := json.Unmarshal([]byte(raw), &parsed); err != nil {
		return nil, fmt.Errorf("could not be parsed as a JSON object with string values: %w (raw value: %q)", err, raw)
	}
	if len(parsed) == 0 {
		return nil, nil
	}
	return parsed, nil
}
```

backend/src/apiserver/main.go
```diff
@@ -660,6 +649,31 @@ func initConfig() error {
...
+	// This orders startup only. Configuration is also read while serving, for example when a
+	// run is created, and those reads still overlap with the watcher. Making reloading itself
+	// safe needs either synchronization around every configuration read or no watcher at all,
+	// and belongs with whoever owns this configuration layer rather than here.
+	// The callback is registered first. WatchConfig only returns once its watcher is live, so
+	// registering afterwards leaves a gap in which an event finds no callback and skips it, and
+	// the field is written and read without synchronization. Viper's own example registers
+	// first for the same reason. The callback does not re-read the file either: the watcher
+	// reads it before calling here, so reading again would replace the configuration a second
+	// time for one event and could validate a different version than the one just loaded.
```
hbelmiro:
```suggestion
	// Start the watcher after all startup reads. Viper's watcher reloads on its own goroutine
	// and is not safe for concurrent reads, so starting it earlier races with the reads above.
	// Register the callback first — WatchConfig blocks until the watcher is live.
```

backend/src/apiserver/common/driver_config.go
```diff
@@ -126,45 +130,96 @@ func validateLabels(labels map[string]string) error {
...
+		// Viper offers no way to tell that apart from an absent key.
+		return nil, nil
+	case string:
+		return parseMapConfigString(name, value)
+	case map[string]string:
+		// Only reachable when the value is set programmatically, which a configuration file
+		// cannot do, so the keys are as written and the values are already strings.
+		return copyMap(value), nil
+	case map[string]interface{}:
+		return nil, fmt.Errorf("%s must be a JSON object written as a string, for example \"{\\\"app\\\":\\\"driver\\\"}\", because a JSON object written directly in the configuration file has its keys lowered before it can be checked and two keys differing only in case are merged", name)
```
hbelmiro: The error message is a full sentence explaining why JSON objects are refused. Error messages should say what's wrong and what to do, not justify the design decision.

```suggestion
		return nil, fmt.Errorf("%s must be a JSON string, not an object; write it as \"{\\\"app\\\":\\\"driver\\\"}\"", name)
```

backend/src/apiserver/common/driver_config_test.go
hbelmiro, line 1: The same verbose comment pattern from the production code applies to the test file. Test names are already descriptive. The worst examples:

- `AnnotationsAtLimitAccepted` (lines 589-598): 10 lines justifying why no margin is reserved for compiler annotations
- `ConfigFileObjectHidesValues` (lines 723-733): 11 lines re-explaining Viper's key lowering (already covered in the production code comment and the PR description)
- `ConfigFileNativeNullMeansUnset` (lines 791-798): 8 lines explaining Viper's null behavior
- `EmptyEnvVarShadowsConfigFile` (lines 760-766): 7 lines explaining Viper precedence

Please remove these comments. The test name and the assertions tell the reader what's being verified. The design reasoning belongs in the PR description.

backend/src/apiserver/common/driver_config_test.go
```diff
@@ -325,7 +440,7 @@ func TestInitDriverPodConfig_EmptyJSONObjectIsValid(t *testing.T) {
 }
 
 // TestInitDriverPodConfig_BlankValuesAreValid guards the manifest default, where the
-// ConfigMap ships DriverPodLabels and DriverPodAnnotations as empty strings. A blank or
+// ConfigMap ships DRIVER_POD_LABELS and DRIVER_POD_ANNOTATIONS as empty strings. A blank or
```
hbelmiro: Please revert this. Comments should reference the Go constant (`DriverPodLabels`), not the string value (`DRIVER_POD_LABELS`).

backend/src/apiserver/common/driver_config_test.go
```diff
@@ -341,7 +456,7 @@ func TestInitDriverPodConfig_BlankValuesAreValid(t *testing.T) {
 
 // TestInitDriverPodConfig_FromEnvVarWiring locks the contract between the API server
 // Deployment and Viper. The Deployment wires the pipeline-install-config keys into the
-// container as the environment variables DRIVERPODLABELS and DRIVERPODANNOTATIONS,
+// container as the environment variables DRIVER_POD_LABELS and DRIVER_POD_ANNOTATIONS,
```
hbelmiro: Please revert.

backend/src/apiserver/common/driver_config_test.go
```diff
@@ -368,7 +483,7 @@ func TestInitDriverPodConfig_FromEnvVarWiring(t *testing.T) {
 }
 
 // TestInitDriverPodConfig_EmptyEnvVarIsValid guards the default install. Every manifest
-// ships DriverPodLabels and DriverPodAnnotations as empty strings, and the Deployment
+// ships DRIVER_POD_LABELS and DRIVER_POD_ANNOTATIONS as empty strings, and the Deployment
```
hbelmiro: Please revert.

backend/README.md
hbelmiro, line 1: The driver pod configuration section grew from ~7 lines to ~50 lines. The same verbosity pattern from the code and test comments applies here.

thc1006:
All applied, thank you.

One line inside `jsonToMap` differs from your version. Unmarshalling straight into `map[string]string` turns a JSON null into an empty string with no error, so `{"sidecar.istio.io/inject": null}` would reach the pod with an empty value and a bare `null` would read as no configuration. The values are checked as `any` before conversion for that reason. Signature, name and error wrapping are yours.

Two things I noticed while applying it. The annotation error named the setting twice, once from my wrapper and once from the field path, so the wrapper is gone. And your watcher comment replaced a note that the ordering fixes startup only, since configuration is read while serving too; say the word if you want it back.

thc1006:
Correcting myself: you asked for those four test comments to be removed and I had shortened three of them. They are gone now.

backend/src/apiserver/common/driver_config_test.go
```diff
@@ -452,6 +631,168 @@ func TestInitDriverPodConfig_IstioMetadataAccepted(t *testing.T) {
...
+				assert.Contains(t, err.Error(), key)
+				assert.Contains(t, err.Error(), "must be a JSON string",
+					"the error should name the form the operator must use instead")
+				assert.Nil(t, GetDriverPodConfig(), "a refused value must not reach the cache")
+			}
+		})
+	}
+}
+
+func TestInitDriverPodConfig_ConfigFileObjectHidesValues(t *testing.T) {
```
hbelmiro, line 661: `TestInitDriverPodConfig_ConfigFileObjectHidesValues` asserts `map[string]any{"team": "ml"}` after Viper lowercases keys from `{"Team": "ml", "team": null}`. Both keys lowercase to `"team"`, so one overwrites the other in Viper's `insensitiviseMap`. Go map iteration order is randomized (Go 1.26), so which value survives is non-deterministic. The test passes when `"Team"` is iterated second (value `"ml"` overwrites `nil`), but fails when `"team"` is iterated second (value `nil` overwrites `"ml"`). This makes the test flaky.

**Suggested fix:** assert the merge happened (2 keys collapsed to 1) without asserting which value survived:

```go
func TestInitDriverPodConfig_ConfigFileObjectHidesValues(t *testing.T) {
	writeConfigFile(t, `{"DRIVER_POD_LABELS": {"Team": "ml", "team": null}}`)

	got, ok := viper.Get(DriverPodLabels).(map[string]any)
	require.True(t, ok, "Viper should return a map for a JSON object")
	assert.Len(t, got, 1, "case-differing keys merge, proving one value is hidden")
	_, hasTeam := got["team"]
	assert.True(t, hasTeam, "the surviving key should be lowercased")
}
```

This is deterministic and still demonstrates the key-folding problem that justifies refusing the object form.

backend/src/apiserver/common/driver_config.go
```diff
@@ -126,45 +130,65 @@ func validateLabels(labels map[string]string) error {
...
+			return nil, fmt.Errorf("has a value that is not a string for key %q; quote it, for example \"true\" rather than true or null (raw value: %q)", k, raw)
 		}
+		parsed[k] = s
 	}
-	return nil
+	if len(parsed) == 0 {
+		return nil, nil
+	}
+	return parsed, nil
 }
```
hbelmiro, line 190: This can be simplified.

```suggestion
func jsonToMap(value string) (map[string]string, error) {
	raw := strings.TrimSpace(value)
	if raw == "" {
		return nil, nil
	}
	var m map[string]any
	if err := json.Unmarshal([]byte(raw), &m); err != nil {
		return nil, fmt.Errorf("could not be parsed as a JSON object: %w (raw value: %q)", err, raw)
	}
	if m == nil {
		return nil, fmt.Errorf("must be a JSON object, not null (raw value: %q)", raw)
	}
	parsed := make(map[string]string, len(m))
	for k, v := range m {
		s, ok := v.(string)
		if !ok {
			return nil, fmt.Errorf("value for key %q must be a string (raw value: %q)", k, raw)
		}
		parsed[k] = s
	}
	if len(parsed) == 0 {
		return nil, nil
	}
	return parsed, nil
}
```

backend/README.md
```diff
@@ -71,25 +71,46 @@ The API server supports configuring custom labels and annotations for driver pod
...
 ```
 
+Write each value as a JSON string, not as a JSON object. A JSON object is refused at startup,
+because the configuration loader folds its keys to lower case and merges keys that differ only
+in case.
+
+This file is baked into the image and supplies the built in default. On a standard install the
+Deployment always passes both keys in from the ConfigMap as environment variables, and those
+take precedence over this file even when empty, so use the ConfigMap below unless you build
+your own image.
```
hbelmiro:
```suggestion
Write each value as a JSON string, not a JSON object (rejected at startup). The ConfigMap takes precedence over this file.
```

backend/README.md
```diff
@@ -71,25 +71,46 @@ The API server supports configuring custom labels and annotations for driver pod
...
+
+Note: Labels and annotations with the prefix `pipelines.kubeflow.org/` are reserved and will be filtered out to prevent overriding system metadata.
 
-Note: Labels and annotations with the prefix `pipelines.kubeflow.org/` are reserved and will be filtered out to prevent overriding system metadata. Changes to driver pod configuration require an API server restart.
+Changes require an API server restart, and each API server reads the configuration once as it starts. An unpinned recurring run picks the change up when it next triggers, once the restart has finished; during a rolling restart both the old and new API server serve for a short window. A pinned recurring run is recompiled by the reconciliation that runs at startup, which stops at the first recurring run it cannot process and is not retried, so the ones after a failure keep their previous specification until the next restart.
```
hbelmiro:
```suggestion
Changes require an API server restart.
```

backend/README.md
```diff
@@ -71,25 +71,46 @@ The API server supports configuring custom labels and annotations for driver pod
...
 ```
 
-When configured via the ConfigMap, every value must be a JSON string. Write `"true"` rather than the bare boolean `true`, and never `null`. The configuration is validated at API server startup: malformed JSON, any value that is not a JSON string, and label keys, label values, or annotation keys that Kubernetes would reject all cause startup to fail with a descriptive error rather than being silently ignored. Annotation values are free form, so a value such as an inline JSON document is accepted.
+Every entry must have a string value: write `"true"` rather than the bare boolean `true`, and never `null`. The configuration is validated at API server startup, so malformed JSON, a value that is not a string, and label keys, label values or annotation keys that Kubernetes would reject all fail startup with a descriptive error rather than being silently ignored. Annotation values are free form, so an inline JSON document is accepted. Annotations are also held to the 256 KiB Kubernetes allows per object, and each setting travels as one environment variable, which Linux caps at 32 memory pages (around 128 KiB with 4 KiB pages).
```
hbelmiro:
```suggestion
Every entry must have a string value: write `"true"` rather than the bare boolean `true`, and never `null`. The configuration is validated at API server startup, so malformed JSON, a value that is not a string, and label keys, label values or annotation keys that Kubernetes would reject all fail startup with a descriptive error rather than being silently ignored. Annotation values are free form, so an inline JSON document is accepted. Annotations are also held to the 256 KiB limit Kubernetes allows per object.
```

thc1006:
All applied.

The `AGENTS.md` change is gone as well. Master restructured that file into `docs/agents/*.md` while this was open, so the environment variable section my entry went into is no longer there, and the rule that asked for the entry went with it. Both variables are documented in `backend/README.md`, where the other API server settings live. The PR is backend only now.

thc1006:
You left 21 threads across four rounds and I answered none of them inline, only pushed. From your side this must have looked like the review went nowhere, and that's on me. Sorry, and thank you for staying with it.

TLDR: everything you raised is in, the last round is `d005eb189`, and CI is green.

The `driver_config.go` suggestions are applied as written, including the `jsonToMap` rewrite and the error message that no longer argues for the design. `main.go` carries your watcher comment verbatim. Test comments name `DriverPodLabels` rather than the string. `ConfigFileObjectHidesValues` is your version, asserting that the merge happened rather than which value survived. The three README wordings from the 11th are in as you wrote them.

On the verbose comments I had only removed the four you named, which was reading your point too narrowly. This round went through the whole test file: comment blocks of three lines or more went from 17 to 1. The Viper key-folding argument had been restated four times there and is gone, since `parseMapConfig` already carries it once. What's left is the few facts a test name can't give you, like Viper turning a JSON null into an empty string.

The README section is where I did something different from what you asked. Measured against the PR base it went from 27 lines to 39, and all of the growth is the Istio sidecar subsection. I cut the prose there and kept the YAML, because that example is the hardest case of the string form this PR introduces. Happy to drop the subsection entirely if you would rather it went.

Shall I resolve the threads, or would you prefer to walk them yourself?

droctothorpe:
@thc1006: Hey, thank you for your PR! As you are not a member of Kubeflow organization, your PR must link to an issue with 'ready' label. Please read our [CONTRIBUTING.md](https://github.com/kubeflow/pipelines/blob/master/CONTRIBUTING.md), closing this PR

PR #13789, opened 2026-07-23
fix(backend): record CANCELLING transition in TerminateRun state history

jahnavigajjala-3:
## Description

`RunStore.TerminateRun` transitions a run to the `CANCELLING` state using a direct SQL `UPDATE`, but unlike `CreateRun` and `UpdateRun`, it does not record this transition in `StateHistory`.

As a result, the run's state history skips the user-initiated `CANCELLING` state, making the recorded timeline inconsistent with other state transitions.

This PR updates `TerminateRun` to:

- Load the existing `StateHistory`
- Append a `CANCELLING` entry if it is not already the latest recorded state
- Persist the updated `StateHistory` together with the existing state transition

This keeps `TerminateRun` consistent with the other run state update paths and preserves a complete state transition history.

## Testing

- Updated `TestTerminateRun` to verify that `StateHistory` contains a `CANCELLING` entry after termination.
- Updated `TestTerminateRun_RunDoesNotExist`.
- Ran the `backend/src/apiserver/storage` test suite successfully.

## Checklist

- [x] I have signed off my commits.
- [x] The PR title follows the project's title convention.

PR #13792, opened 2026-07-23
fix(backend): append CANCELLING to StateHistory in TerminateRun

### LINKED ISSUE #13786: [backend] TerminateRun does not append CANCELLING state to StateHistory  Body:
### Steps to reproduce

RunStore.TerminateRun in backend/src/apiserver/storage/run_store.go transitions a run's State column to CANCELLING via a raw SQL UPDATE, but never appends this transition to the StateHistory column. Every other state-changing path in this file (CreateRun, UpdateRun) appends a new RuntimeStatus entry to StateHistory whenever the state changes; TerminateRun is the only exception.

The code has an existing TODO acknowledging this:

https://github.com/kubeflow/pipelines/blob/master/backend/src/apiserver/storage/run_store.go
```go
func (s *RunStore) TerminateRun(runId string) error {
	// TODO(gkcalat): append CANCELLING to StateHistory
	result, err := s.db.Exec(`
		UPDATE run_details
		SET Conditions = ?, State = ?
		WHERE UUID = ? AND (State = ? OR State = ? OR State = ? OR State = ?)`, ...)
```

### Expected result

StateHistory should contain an entry with State: CANCELLING recorded at the time of termination — consistent with how CreateRun and UpdateRun both append a new RuntimeStatus entry whenever the run's state changes.

Instead, no such entry is ever added; StateHistory is left completely unchanged by this call. This means StateHistory is missing the point at which cancellation was requested — the history jumps straight from Running to whatever the next writer (e.g. ReportWorkflowResource, once Argo reflects the termination) happens to append, with no record that termination was user-initiated.

### Materials and Reference

TerminateRun currently does a blind UPDATE with no read step first, unlike UpdateRun, which receives a fully-populated *model.Run (existing StateHistory already loaded by the caller) and appends to it in memory before writing:

```go
if len(run.RunDetails.StateHistory) == 0 || run.RunDetails.StateHistory[len(run.RunDetails.StateHistory)-1].State != run.RunDetails.State {
    run.RunDetails.StateHistory = append(run.RunDetails.StateHistory, &model.RuntimeStatus{
        UpdateTimeInSec: s.time.Now().Unix(),
        State:           run.RunDetails.State,
    })
}
```
A fix would need an equivalent read-append-write step in TerminateRun — fetching the run's current StateHistory, appending a Cancelling entry, and including the updated JSON in the same guarded UPDATE — while preserving the existing WHERE State IN (Paused, Pending, Running, Unspecified) guard, which is what makes the transition safe to apply.

SebTardif:
## Summary

`TerminateRun` set `State`/`Conditions` to CANCELLING but never updated
`StateHistory`, while `CreateRun`/`UpdateRun` always append a
`RuntimeStatus` entry on state change. This PR loads the run, appends
CANCELLING to history (same pattern as `UpdateRun`), and persists
`StateHistory` in the same UPDATE.

Fixes #13786

## Test plan

- [x] Matches CreateRun/UpdateRun history-append pattern
- [ ] CI unit tests for run store

### DIFF
--- a/backend/src/apiserver/storage/run_store.go
+++ b/backend/src/apiserver/storage/run_store.go
@@ -821,13 +821,33 @@ func NewRunStore(db *DB, time util.TimeInterface) *RunStore {
 }
 
 func (s *RunStore) TerminateRun(runId string) error {
-	// TODO(gkcalat): append CANCELLING to StateHistory
+	run, err := s.GetRun(runId)
+	if err != nil {
+		return err
+	}
+
+	cancelling := model.RuntimeStateCancelling
+	if len(run.RunDetails.StateHistory) == 0 ||
+		run.RunDetails.StateHistory[len(run.RunDetails.StateHistory)-1].State != cancelling {
+		run.RunDetails.StateHistory = append(run.RunDetails.StateHistory, &model.RuntimeStatus{
+			UpdateTimeInSec: s.time.Now().Unix(),
+			State:           cancelling,
+		})
+	}
+	stateHistoryString := ""
+	if historyString, err := json.Marshal(run.RunDetails.StateHistory); err == nil {
+		stateHistoryString = string(historyString)
+	} else {
+		return util.NewInternalServerError(err, "Failed to marshal state history while terminating run %s", runId)
+	}
+
 	result, err := s.db.Exec(`
 		UPDATE run_details
-		SET Conditions = ?, State = ?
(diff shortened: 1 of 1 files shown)


SebTardif:
Closing as a duplicate of #13789, which targets the same TerminateRun StateHistory gap and already includes unit tests.

Sorry for the overlapping PR.

PR #13794, opened 2026-07-23
ci: publish durable CI health and incident trends

jeffspahr:
## Summary

- publish an attempt-specific normalized result from successful and failed lanes in all seven tracked heavy workflows
- classify expected lanes that fail before publishing a normalized result as runner loss, job timeout, or missing-result infrastructure failures using GitHub check annotations
- record the final setup-gate outcome so pre-test deploy and artifact-barrier failures are classified as infrastructure failures
- preserve true per-test execution denominators, structured infrastructure signatures, matrix dimensions, and setup/test/report timing
- recognize external registry timeouts and missing branch image artifacts as structured infrastructure signatures
- group a failed image producer with every lane blocked on its artifact as one root-cause event while preserving every failed job in the headline rate
- correlate failed lane windows with relevant public GitHub Status incidents, including signature-backed evidence, strict overlap, and a separately labeled 15-minute pre-reporting window
- keep observed failures in the headline rate while showing the correlated count/rate and direct incident links in the issue and dashboard
- replace the capped rolling scan with a 90-day bootstrap and incremental 14-day overlap that preserves reruns and late incident updates
- persist a lightweight date manifest and exact per-day snapshots without time-based pruning on a generated `ci-metrics` branch
- keep the `ci-health` tracking issue as the current seven-day comparison, including the prior-period delta and Wilson interval
- deploy a dependency-free GitHub Pages dashboard with 7/14/28/90-day presets, all-history and custom inclusive date ranges, workflow/lane filtering, equal-length prior-period comparisons, duration phases, failure classes, network signatures, GitHub incident context, grouped infrastructure events, rerun rescues/time-to-green, per-test rates, and master change markers
- fetch and cache only the selected and comparison daily snapshots, while keeping the Pages deployment independent of historical data volume
- define every headline metric and analysis panel with compact, calculation-accurate tooltips available on mouse hover and keyboard focus
- define lane-board and flakiest-test table columns with the same accessible tooltip treatment
- balance the six headline metrics in a 3×2 desktop grid while retaining the existing tablet and mobile breakpoints
- separate the masthead eyebrow from the display title so their type never collides

## Why

The existing report answers which lanes and tests are failing in the current rolling window, but it overwrites its prior snapshot. That makes it difficult to identify the start of a regression, measure a CI improvement, separate product failures from infrastructure failures, or compare wall-clock changes over time. It also derives per-test counts only from failed-run JUnit artifacts, so it cannot calculate a true flake rate.

This change keeps artifacts as short-lived ingestion transport and writes their aggregates into durable, public, machine-readable history before they expire. Collected snapshots are retained without an arbitrary time limit; a retention policy can be introduced later if storage growth becomes material. Each date remains an exact snapshot rather than being aged into a lossy weekly or monthly rollup.

The dashboard initially downloads only a small manifest of available dates. It loads the selected range and immediately preceding equal-length comparison in bounded batches, caches those days for subsequent range changes, and reads production data directly from the public `ci-metrics` branch. The Pages artifact therefore stays small instead of re-uploading or eagerly downloading the complete historical corpus as it grows.

When a runner disappears or a job reaches its hard timeout, the post-job artifact cannot run; the report recognizes that missing result from the Actions job/check data instead of leaving the failure unclassified. Ordinary concurrency cancellations remain excluded. When setup reaches the publisher but tests never start, the recorded setup outcome preserves that failure as infrastructure rather than an unknown test result.

External registry failures can fan out from one failed image-build job into many lanes waiting for the missing artifact. The report still counts every affected job as a contributor-facing failure, but also groups the producer and blocked consumers into one root-cause event. Registry/artifact classification is independent of public status reporting, so an unreported outage remains explicitly marked as having no reported status correlation.

GitHub service incidents remain visible as real contributor-facing CI failures. The report fetches the public Statuspage feed without forwarding repository credentials and adds correlation as diagnostic context, not causation. A matching GitHub-service error in the failed job log strengthens the evidence; time-only matches remain explicitly labeled. The 14-day refresh repairs correlation after GitHub publishes or updates an incident late.

## Operational note

The repository does not currently have GitHub Pages enabled. Health collection, issue updates, and `ci-metrics` persistence continue normally while Pages is disabled. A repository administrator must make the one-time selection **Settings → Pages → Source: GitHub Actions**; the existing workflow will detect that setting and begin deploying the dashboard automatically.

## Verification

- `python3 -m unittest discover -v -p '*_test.py'`: 167 tests passed
- focused result, artifact-barrier, signature, and health-report tests: 65 tests passed
- live GitHub Status payload validation identified the expected July 23 and July 24 Actions incidents
- `actionlint` on every previously modified workflow; YAML parsing on the newly modified action and workflows
- Python compilation, shell syntax checks, dashboard JavaScript parsing, and `git diff --check`
- runner Go tool verified with `GOFLAGS=-json`
- dashboard rendered and exercised at desktop and 390px widths; workflow/lane filtering, responsive overflow, and tooltip hover/focus behavior passed


zazulam (changes requested):
## PR overview

Durable CI health history is valuable, but this is not ready to become a permanent operational dependency. The current 3,524-line change combines lane instrumentation, publishing, durable storage, incident enrichment, and a Pages UI, which makes correctness, rollout, and rollback difficult to review independently.



## Blocking findings

- **Failed tests can be reported as successful.** In `.github/workflows/integration-tests-v1.yml:100-116` and `.github/workflows/kfp-webhooks.yml:62-64`, `go test`/`make` is piped through `tee` without `pipefail`. The default Actions shell therefore returns the status from `tee`, so the step and normalized artifact can record a false pass. Please use `shell: bash` or explicitly propagate `PIPESTATUS[0]`, with regression coverage.

- **Incomplete refreshes can destroy complete history.** `.github/resources/scripts/ci_health_report.py:1001-1053` continues after rate limits, truncation, missing workflows, and job/artifact errors, while `merge_history` at lines 1345-1351 replaces the entire existing snapshot for each collected date. A partial refresh can permanently remove observations once that date leaves the overlap window. Please track per-date completeness and replace only complete days, or merge stable observation identities.

- **Manual dispatch can execute non-default-branch code with publishing privileges.** `.github/workflows/ci-health-report.yml:10-22,42,76-116` allows a dispatch ref to control checked-out code while granting `contents`, `issues`, `pages`, and OIDC write permissions. Please gate execution to `refs/heads/master`, explicitly check out the trusted ref, and scope permissions per job.

- **Publisher failures can be silent.** `.github/actions/publish-ci-result/action.yml:111-158` suppresses both result generation and upload failures, and `.github/actions/test-and-report/action.yml:623-646` suppresses the composite outcome. A systemic publisher regression can therefore produce no artifacts and no visible completeness failure. This needs a dedicated reporting/completeness signal; it does not need to turn product-test failures into telemetry failures.

There are also correctness issues in the reporting layer: the latest seven-day calculation at `.github/resources/scripts/ci_health_report.py:1408-1414` covers eight dates versus seven in the comparison period; parse errors emitted by `ci_result.py` are not incorporated into completeness; and workflow filters leave several dashboard panels globally scoped.

## Scope recommendation

Please split this into independently reviewable changes:

1. normalized lane instrumentation and artifact publishing;
2. durable ingestion, completeness semantics, and versioned storage;
3. incident correlation and dashboard presentation.

Each layer should have a defined contract and failure signal before the next depends on it. Please also document long-term ownership, retention/recovery behavior, and schema migration. The current branch conflict with `master` will need resolution separately.

jeffspahr:
Thanks — all four blocking findings and the three reporting-layer issues are addressed in 58e045a, kept as a single PR.

## Blocking

**Failed tests reported as passing.** Confirmed: the default `run:` shell is `bash -e {0}`, without `pipefail`, so the three lanes piping `go test`/`make` into `tee` exited with tee's status. The two `integration-tests-v1.yml` steps and the `kfp-webhooks.yml` step now declare `shell: bash`. `legacy-v2-api-integration-tests.yml` already did, which is the pattern I followed.

Regression coverage is `workflow_pipefail_test.py`, a repo-wide guard rather than three assertions: it scans every workflow and fails on any `tee` pipeline lacking `shell: bash` or explicit `PIPESTATUS` handling. Verified it reports exactly those three steps against the pre-fix tree. `ci-scripts-tests.yml` now also triggers on `.github/workflows/*.y*ml` so a newly added workflow cannot bypass it.

**Incomplete refreshes destroying complete history.** Collection now records which dates lost data — listing rate limits and truncation degrade the whole window, job-listing and artifact errors degrade only that run's date. Snapshots carry a `complete` flag and `merge_history` will not replace a complete stored day with a degraded one; a degraded day still fills a gap, and between two degraded versions the one observing more lane runs wins. Days written before this tracking are treated as complete, so the upgrade preserves existing history rather than discarding it (`schema_version` deliberately unchanged, since bumping it would empty the store).

Demonstrated directly: a degraded refresh reporting 3 lane runs against a stored complete day of 40 previously wrote 3; it now keeps 40.

**Dispatch privileges.** Both jobs are gated on `github.ref == 'refs/heads/master'`, checkout pins `ref: master`, and the workflow-level grant is replaced with `permissions: {}` plus per-job scopes. The report job no longer receives `id-token: write` — only `deploy` does. `ci_health_workflow_test.py` guards each of these properties.

**Silent publisher failures.** Two layers, neither of which fails a product lane for telemetry reasons:
- `publish-ci-result` emits an `::error::` annotation naming the lane when generate or upload fails, so the occurrence is visible on the run that produced it.
- The report computes publisher completeness and annotates its own run, escalating to `::error::` past a 25% gap. This also closes the parse-error finding: a result that arrived but yielded no tests counts toward the gap, so `test_parse_errors` is no longer ignored.

## Reporting layer

- **Window skew** confirmed: `window_totals` is half-open, so `[today-7, today+1)` was 8 dates against 7 for the comparison. Both are now `COMPARISON_WINDOW_DAYS` wide, with a test asserting equal date counts.
- **Globally scoped panels**: the infrastructure metric now aggregates per-lane `result_lanes` classes honoring the workflow and search filters, instead of day-level `failure_classes`. Reruns, rescues, and time-to-green exist only as day totals and genuinely cannot be narrowed, so rather than implying otherwise the tile is now labeled `all workflows · not filtered` while a filter is active. Snapshots predating per-lane classes report how many days could not be attributed instead of silently contributing repository-wide numbers.

## Scope

I kept this as one PR per the repository owner's direction. The layering concern is noted, and the completeness and publisher-health contracts added here give the ingestion boundary an explicit failure signal.

## Verification

- `.github/resources/scripts` suite: 264 tests pass
- All modified workflow and action YAML parses
- Python 3.9 syntax gate over 1,237 tracked files
- `git diff --check`
- Dashboard filter logic exercised headlessly: unfiltered 10, filtered-to-workflow 7 (previously 10), and legacy snapshots reported as unattributable rather than counted

PR #13798, opened 2026-07-23
fix(backend): append CANCELLING to run StateHistory on TerminateRun

### LINKED ISSUE #13786: [backend] TerminateRun does not append CANCELLING state to StateHistory  Body:

uditjainstjis:

`RunStore.TerminateRun` transitioned a run's `State` column to `CANCELLING` via a raw SQL `UPDATE` but never appended that transition to `StateHistory`, unlike every other state-changing path in `run_store.go` (`CreateRun`, `UpdateRun`), which append a `RuntimeStatus` entry whenever the state changes. This left the run history incomplete for terminated runs and was called out by an existing `// TODO(gkcalat): append CANCELLING to StateHistory`.

This PR makes `TerminateRun` load the current `StateHistory` inside a transaction, append a `CANCELLING` `RuntimeStatus` stamped with the current time, and persist it alongside the `State`/`Conditions` update under the same state guard — keeping the write atomic and race-safe. Behavior for non-terminable or already-finished/missing runs (the `Row not found` case) is unchanged.

**Testing:**
- Updated `TestTerminateRun` to assert the recorded `CANCELLING` transition in `StateHistory`.
- Added `TestTerminateRun_AppendsCancellingToStateHistory` as a focused regression test.
- `go test ./backend/src/apiserver/storage/` passes locally (full package, no regressions).

Fixes #13786

> AI disclosure (per the [Kubeflow AI policy](https://www.kubeflow.org/docs/about/ai_policy/)): parts of this change were drafted with AI assistance; see the `Assisted-by:` trailer in the commit. All logic was reviewed and is owned by the author.

**Checklist:**
- [x] The title for your pull request (PR) follows the [title convention](https://github.com/kubeflow/pipelines/blob/master/CONTRIBUTING.md#pull-request-title-convention).

### DIFF
--- a/backend/src/apiserver/storage/run_store.go
+++ b/backend/src/apiserver/storage/run_store.go
@@ -821,13 +821,62 @@ func NewRunStore(db *DB, time util.TimeInterface) *RunStore {
 }
 
 func (s *RunStore) TerminateRun(runId string) error {
-	// TODO(gkcalat): append CANCELLING to StateHistory
-	result, err := s.db.Exec(`
+	tx, err := s.db.DB.Begin()
+	if err != nil {
+		return util.NewInternalServerError(err, "transaction creation failed")
+	}
+
+	// Load the current state history inside the transaction so we can append the
+	// CANCELLING transition, mirroring CreateRun/UpdateRun which record every
+	// state change in StateHistory. The state guard matches the UPDATE below so a
+	// run that is not in a terminable state (or does not exist) yields no row.
+	var stateHistoryString sql.NullString
+	err = tx.QueryRow(`
+		SELECT StateHistory FROM run_details
+		WHERE UUID = ? AND (State = ? OR State = ? OR State = ? OR State = ?)`,
+		runId,
+		model.RuntimeStatePaused.ToString(),
+		model.RuntimeStatePending.ToString(),
+		model.RuntimeStateRunning.ToString(),
+		model.RuntimeStateUnspecified.ToString(),
+	).Scan(&stateHistoryString)
+	if err != nil {
+		tx.Rollback()
+		if err == sql.ErrNoRows {
+			return util.NewInvalidInputError("Failed to terminate a run %s. Row not found", runId)
+		}
+		return util.NewInternalServerError(err,
+			"Failed to terminate a run %s. Error: '%v'", runId, err.Error())
+	}
+
+	stateHistory := []*model.RuntimeStatus{}
+	if stateHistoryString.Valid && stateHistoryString.String != "" {
+		if err := json.Unmarshal([]byte(stateHistoryString.String), &stateHistory); err != nil {
+			tx.Rollback()
+			return util.NewInternalServerError(err,
+				"Failed to unmarshal state history while terminating run %s", runId)
+		}
+	}
+	if len(stateHistory) == 0 || stateHistory[len(stateHistory)-1].State != model.RuntimeStateCancelling {
+		stateHistory = append(stateHistory, &model.RuntimeStatus{
+			UpdateTimeInSec: s.time.Now().Unix(),
+			State:           model.RuntimeStateCancelling,
+		})
+	}
+	newStateHistoryString, err := json.Marshal(stateHistory)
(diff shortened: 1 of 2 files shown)


uditjainstjis:
@hbelmiro this one is still sitting on `needs-ok-to-test`, so no CI has run on it yet — would you mind dropping an `/ok-to-test` when you have a moment?

It's a small, self-contained backend change: `RunStore.TerminateRun` updated the `State` column to `CANCELLING` but never appended that transition to `StateHistory`, unlike `CreateRun` and `UpdateRun` in the same file. It closes the existing `// TODO(gkcalat): append CANCELLING to StateHistory` and comes with a `run_store_test.go` case. Two files, no API or schema change.

Happy to rebase if it has gone stale in the meantime.

uditjainstjis:
Thanks for kicking off the Copilot review @mprahl — it found a real one.

The read-modify-write wasn't row-locked, so the persistence path could commit a state change between the `SELECT` and the `UPDATE` and have its `StateHistory` entry overwritten by the stale JSON. Fixed with the dialect-aware `SelectForUpdate` helper it pointed at, which appends `FOR UPDATE` on MySQL and is a no-op on SQLite.

One caveat on coverage: because the SQLite dialect returns the query unchanged, the existing tests exercise the unlocked path either way — they confirm no regression, not that the lock works. Verifying that properly needs a MySQL-backed concurrency test, which I'm happy to add if you'd like it in this PR.

Also rebased onto master. @hbelmiro, a `/lgtm` when you have a moment would be very welcome — CI was green across all 100 checks before the rebase.

droctothorpe:
Hi @uditjainstjis, thank you for your PR! The linked issue (#13786) must be labeled `ready` before your contribution can be reviewed. Please check our [CONTRIBUTING.md](https://github.com/kubeflow/pipelines/blob/master/CONTRIBUTING.md) for details.

PR #13799, opened 2026-07-23
chore: apply gofmt formatting corrections in backend Go files

kaikaila:
## Description

Four backend Go files contain over-aligned struct field / const block formatting that does not match `gofmt` output. This PR runs `gofmt -w` on them so they conform to the standard formatting.

Files:
- \`backend/src/apiserver/server/pipeline_upload_server.go\`
- \`backend/src/apiserver/storage/job_store_test.go\`
- \`backend/src/common/util/scheduled_workflow_test.go\`
- \`backend/src/crd/controller/scheduledworkflow/util/scheduled_workflow_test.go\`

## Notes

Whitespace-only change with **no functional impact** — \`git diff --ignore-all-space\` is empty. \`golangci-lint\` (including \`golangci-lint fmt\`) passes.

### DIFF
--- a/backend/src/apiserver/server/pipeline_upload_server.go
+++ b/backend/src/apiserver/server/pipeline_upload_server.go
@@ -42,13 +42,13 @@ import (
 )
 
 const (
-	FormFileKey               = "uploadfile"
-	NameQueryStringKey        = "name"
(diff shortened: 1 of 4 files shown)


hbelmiro (approved):
/lgtm
/approve

PR #13803, opened 2026-07-24
fix(sdk): reject reserved oci-prepull-<N> init container names in add_init_container

### LINKED ISSUE #13802: [sdk] add_init_container does not reject reserved oci-prepull-<N> names, causing silent runtime failure with Modelcar artifacts
### Environment

*  KFP version: master (≥ v2.15.2, post-PR #13651)
*  KFP SDK version: kfp==2.15.2, kfp-kubernetes==2.15.2
*  All dependencies version:
```
kfp               2.15.2
kfp-kubernetes    2.15.2
kfp-pipeline-spec 0.5.0
```

---

### Steps to reproduce

1. Create a pipeline that uses a **Modelcar OCI artifact as an input**. This causes the backend driver to inject `oci-prepull-0`, `oci-prepull-1`, … init containers into the pod spec via `initPodSpecPatch`.

2. Call `add_init_container` with a name that matches one of those backend-injected names:

```python
from kfp import dsl
from kfp.kubernetes import add_init_container

@dsl.component
def my_task(model: dsl.Input[dsl.Model]):
    pass

@dsl.pipeline
def my_pipeline():
    task = my_task(model=some_oci_artifact)
    # SDK raises NO error here — but the backend will fail at runtime
    add_init_container(
        task,
        name='oci-prepull-0',   # conflicts with backend-injected init container
        image='busybox:1.36',
        command=['sh', '-c', 'echo pre-flight'],
    )
```

3. Compile and submit the pipeline. Observe that **no `ValueError` is raised** during SDK compile time.

4. The pipeline task pod fails at runtime inside the driver with:
```
init container name "oci-prepull-0" conflicts with an existing container in the pod
```

---

### Expected result

`add_init_container()` should raise a `ValueError` at **compile time** when the supplied `name` matches the KFP-reserved `oci-prepull-<N>` pattern — the same way it already rejects `kfp-launcher`.

The docstring for the `name` argument should also document all reserved names and patterns so users are not surprised.

---

### Actual result

The SDK-level guard only rejects the single hardcoded literal `kfp-launcher`:

```python
# kubernetes_platform/python/kfp/kubernetes/init_container.py  (lines 70–75)
if not name:
    raise ValueError('Argument for "name" must be a non-empty string.')
if name == 'kfp-launcher':                          # ← only guard that exists
    raise ValueError(
        'Init container name "kfp-launcher" is reserved by Kubeflow Pipelines.'
    )
```

The backend driver (`backend/src/v2/driver/k8s.go`, lines 789–804) seeds its
collision map from **all** existing pod containers — including `oci-prepull-*`
init containers that are injected by `initPodSpecPatch` **before**
`extendPodSpecPatch` processes the user-supplied init containers:

```go
// k8s.go – extendPodSpecPatch()
initContainerNames := make(map[string]bool)
for _, c := range podSpec.Containers {
    initContainerNames[c.Name] = true
}
for _, c := range podSpec.InitContainers {   // includes oci-prepull-0, oci-prepull-1, …
    initContainerNames[c.Name] = true
}
for _, initContainer := range kubernetesExecutorConfig.GetInitContainers() {
    if initContainerNames[initContainer.GetName()] {
        return fmt.Errorf(
            "init container name %q conflicts with an existing container in the pod",
            initContainer.GetName())
    }
    ...
}
```

`oci-prepull-N` containers are injected by `initPodSpecPatch`
(`backend/src/v2/driver/driver.go`, line 643):

```go
Name: fmt.Sprintf("oci-prepull-%d", i),
```

`initPodSpecPatch` is called **before** `extendPodSpecPatch`
(`backend/src/v2/driver/container.go`, lines 252–281):

```go
podSpec, err := initPodSpecPatch(...)    // injects oci-prepull-* first
...
if opts.KubernetesExecutorConfig != nil {
    err = extendPodSpecPatch(...)         // then processes user init containers
}
```

The collision is therefore only detected at runtime in the driver, never at
SDK compile time.

---

### Materials and Reference

**Affected files:**

| File | Role |
|---|---|
| `kubernetes_platform/python/kfp/kubernetes/init_container.py` | SDK validation — missing `oci-prepull-*` guard |
| `backend/src/v2/driver/k8s.go` | Backend collision check (Go) — already correct |
| `backend/src/v2/driver/driver.go` | Injects `oci-prepull-<N>` init containers |
| `backend/src/v2/driver/container.go` | Establishes `initPodSpecPatch → extendPodSpecPatch` call order |

**Root cause:**

PR #13651 (merged 2026-07-23) added `add_init_container` to the SDK and wired
the backend collision check. The PR's own commit message acknowledges:

> *"Container-name uniqueness now checks against all existing containers and
> init containers in the compiled template (not just user-provided ones),
> preventing collisions with kfp-launcher, oci-prepull-\*, or the main
> container."*

This protection was added **only** on the backend (Go) side. The SDK
(Python) still only rejects `kfp-launcher` and is unaware of the dynamically
generated `oci-prepull-<N>` names, leaving users with a confusing runtime
error instead of a clear compile-time message.

**Duplicate search summary:**

Searched GitHub issues, closed PRs, and discussions for:
- `oci-prepull reserved` — no results
- `add_init_container name conflict` — no results
- `init container kfp-launcher oci-prepull` — no results

No existing issue tracks this specific SDK-side validation gap.

**Proposed fix:**

In `kubernetes_platform/python/kfp/kubernetes/init_container.py`, extend
the name guard:

```python
import re

# After the existing kfp-launcher check:
if re.match(r'^oci-prepull-\d+$', name):
    raise ValueError(
        'Init container names matching "oci-prepull-<N>" are reserved by '
        'Kubeflow Pipelines for Modelcar OCI artifact pre-pull containers.'
    )
```

Update the `name` argument docstring to list all reserved names/patterns:

```
name: Name of the init container. Must be unique within the pod.
    The following names are reserved by Kubeflow Pipelines and must
    not be used: ``kfp-launcher``, ``oci-prepull-<N>`` (where N is
    a non-negative integer).
```

A corresponding unit test should be added to
`kubernetes_platform/python/test/unit/test_init_container.py`:

```python
def test_add_init_container_reserved_oci_prepull_name(self):
    with pytest.raises(
            ValueError,
            match=r'Init container names matching "oci-prepull-<N>" are reserved',
    ):
        @dsl.pipeline
        def my_pipeline():
            task = print_greeting()
            kubernetes.add_init_container(
                task,
                name='oci-prepull-0',
                image='busybox:1.36',
            )
```

---

---

vishalmore90:

When calling `add_init_container()`, the Python SDK previously only rejected the literal name `kfp-launcher`. However, Kubeflow Pipelines backend drivers also inject `oci-prepull-0`, `oci-prepull-1`, etc., init containers when Modelcar OCI artifacts are used as pipeline inputs. 

Because `add_init_container` did not check for `oci-prepull-<N>` at compile time, pipelines using these names passed SDK compilation silently but failed at runtime with:
`init container name "oci-prepull-0" conflicts with an existing container in the pod`

This PR updates `add_init_container()` in `kubernetes_platform/python/kfp/kubernetes/init_container.py` to:
1. Reject init container names matching the regex `^oci-prepull-\d+$` at compile time with a descriptive `ValueError`.
2. Update the docstring of `name` to document all reserved init container names (`kfp-launcher` and `oci-prepull-<N>`).
3. Add unit test coverage in `kubernetes_platform/python/test/unit/test_init_container.py`.

Fixes #13802

**Checklist:**

### DIFF
--- a/kubernetes_platform/python/kfp/kubernetes/init_container.py
+++ b/kubernetes_platform/python/kfp/kubernetes/init_container.py
@@ -12,6 +12,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+import re
 from typing import Dict, List, Optional
 
 from google.protobuf import json_format
@@ -43,7 +44,9 @@ def add_init_container(
 
     Args:
         task: Pipeline task.
-        name: Name of the init container. Must be unique within the pod.
+        name: Name of the init container. Must be unique within the pod. The
+            following names are reserved by Kubeflow Pipelines and must not be used:
+            ``kfp-launcher``, ``oci-prepull-<N>`` (where N is a non-negative integer).
         image: Container image of the init container.
         command: Entrypoint array. Corresponds to the ``command`` field.
         args: Arguments to the entrypoint. Corresponds to the ``args`` field.
@@ -72,6 +75,11 @@ def add_init_container(
     if name == 'kfp-launcher':
         raise ValueError(
             'Init container name "kfp-launcher" is reserved by Kubeflow Pipelines.'
+        )
+    if re.match(r'^oci-prepull-\d+$', name):
+        raise ValueError(
+            'Init container names matching "oci-prepull-<N>" are reserved by '
+            'Kubeflow Pipelines for Modelcar OCI artifact pre-pull containers.'
         )
     if not image:
         raise ValueError('Argument for "image" must be a non-empty string.')
--- a/kubernetes_platform/python/test/unit/test_init_container.py
+++ b/kubernetes_platform/python/test/unit/test_init_container.py
@@ -134,6 +134,21 @@ class TestAddInitContainer:
                     image='busybox:1.36',
                 )
 
+    def test_add_init_container_reserved_oci_prepull_name(self):
+        with pytest.raises(
+                ValueError,
+                match=r'Init container names matching "oci-prepull-<N>" are reserved',
+        ):
+
+            @dsl.pipeline
+            def my_pipeline():
+                task = print_greeting()
+                kubernetes.add_init_container(
+                    task,
+                    name='oci-prepull-0',
+                    image='busybox:1.36',
+                )
+
     def test_add_init_container_empty_name(self):
         with pytest.raises(
                 ValueError,


vishalmore90:
@hbelmiro @zazulam 
Could someone please take a look at this PR?

kubernetes_platform/python/kfp/kubernetes/init_container.py
```diff
@@ -73,6 +76,11 @@ def add_init_container(
         raise ValueError(
             'Init container name "kfp-launcher" is reserved by Kubeflow Pipelines.'
         )
+    if re.match(r'^oci-prepull-\d+$', name):
```
hbelmiro, line 79: Nit/Optional: You don't need a regex for it. You can remove the `import re` on line 15 and replace the regex with:

```suggestion
    if name.startswith('oci-prepull-') and name[len('oci-prepull-'):].isdigit():
```

kubernetes_platform/python/test/unit/test_init_container.py
```diff
@@ -134,6 +134,21 @@ def my_pipeline():
                     image='busybox:1.36',
                 )
 
+    def test_add_init_container_reserved_oci_prepull_name(self):
```
hbelmiro, line 137: The test only covers a single rejection case (`oci-prepull-0`). It does not verify that the validation works for other valid indices or that near-miss names are correctly accepted. A parametrized test covering both accepted and rejected names would strengthen confidence that the regex is neither too narrow nor too broad.

droctothorpe:
Hi @vishalmore90, thank you for your PR! The linked issue (#13802) must be labeled `ready` before your contribution can be reviewed. Please check our [CONTRIBUTING.md](https://github.com/kubeflow/pipelines/blob/master/CONTRIBUTING.md) for details.

PR #13780, opened 2026-07-22
fix(sdk): do not automatically add --trusted-host to pip commands by default. Fixes #11155

dobbydobap:
**Description of your changes:**

`make_index_url_options` auto-added `--trusted-host` for every pip index URL whenever `pip_trusted_hosts` was unset (`None`), silently disabling SSL certificate validation. This affects both `@dsl.component` pip installs and `kfp component build` Dockerfiles, since both route through this helper. The injected value was also the full index URL (`https://.../simple`) where pip expects a bare host, so the default was malformed as well as insecure.

#11151 already added the opt-in secure path (`pip_trusted_hosts=[]`). This makes the **default** secure: `--trusted-host` is emitted only when the user explicitly lists hosts via `pip_trusted_hosts`. `None` now behaves like `[]`.

**Note backward compatibility:** this is an intentional behavior change. Users relying on the implicit `--trusted-host` (e.g. a private mirror with a self-signed cert) must now set `pip_trusted_hosts` explicitly. Requested by maintainers in the issue thread (@droctothorpe, @HumairAK).

**Testing:** `make_index_url_options` with `None`/`[]` → no `--trusted-host`; explicit hosts still honored. Updated `component_factory_test.py` and `cli/component_test.py`; explicit/empty-host tests unchanged.

Fixes #11155

**Checklist:**
- [x] You have signed off your commits
- [x] The title for your pull request (PR) follows our title convention

PR #13807, opened 2026-07-24
fix(sdk): include trusted hosts when pip_index_urls is not set

vishalmore90:
Fixes a bug in `make_index_url_options` (`sdk/python/kfp/dsl/component_factory.py`) where `pip_trusted_hosts` was silently ignored if `pip_index_urls` was `None` or empty due to an early `return ''` check.

With this change:
- `pip_index_urls` and `pip_trusted_hosts` are evaluated independently so `--trusted-host` flags are generated whenever `pip_trusted_hosts` is specified.
- Unit tests added in `component_factory_test.py` covering all combinations of index URLs and trusted hosts.

**Checklist:**

### DIFF
--- a/sdk/python/kfp/dsl/component_factory.py
+++ b/sdk/python/kfp/dsl/component_factory.py
@@ -91,30 +91,32 @@ def make_index_url_options(pip_index_urls: Optional[List[str]],
 
     Returns:
         str:
-            - An empty string if pip_index_urls is empty or None.
+            - An empty string if pip_index_urls and pip_trusted_hosts are empty or None.
             - '--index-url url ' if pip_index_urls contains 1 URL.
             - The above followed by '--extra-index-url url ' for each additional URL in pip_index_urls
             if pip_index_urls contains more than 1 URL.
             - If pip_trusted_hosts is None or an empty List:
                 - No --trusted-host information will be added.
             - If pip_trusted_hosts contains any hosts:
-                - The above followed by '--trusted-host host ' for each host in pip_trusted_hosts.
+                - '--trusted-host host ' for each host in pip_trusted_hosts.
     Note:
-        In case pip_index_urls is not empty, the returned string will contain a space at the end.
+        In case pip_index_urls or pip_trusted_hosts is provided, the returned string will contain a space at the end.
     """
-    if not pip_index_urls:
-        return ''
+    options = []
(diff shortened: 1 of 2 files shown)


vishalmore90:
@nsingla @zazulam 
Could someone please take a look at this PR?

dobbydobap:
This looks correct to me. `options` accumulates index urls and trusted hosts independently and the `if not options` guard sits after both, so `pip_trusted_hosts` on its own now emits `--trusted-host host ` instead of being dropped by the old early return. I checked the ordering specifically at `34973c90f`, since getting that guard the wrong way round would silently not fix the bug.

The bug is still on master today, so nothing has landed in the meantime.

Worth knowing the same fix was opened again as #14026 on 12 August, which suggests people are still hitting this. Coverage there is equivalent to yours, the one structural difference is that it splits the cases into separate named test methods rather than assertions in a single one, which isolates failures better. Minor, but easy to adopt if you are touching this again.

On the surrounding behaviour, `--trusted-host` is still only emitted when a user passes `pip_trusted_hosts` explicitly, which is what #13780 established when it stopped adding it automatically, so this does not reintroduce that default.

The blocker looks procedural rather than technical: still `needs-ok-to-test` and behind master.

dobbydobap:
`yapf-sdk` is failing on the test you added. The job output asks for two reformats in `test_make_index_url_options`:

```
-                pip_trusted_hosts=None),
-            '--index-url https://pypi.org/simple ')
+                pip_trusted_hosts=None), '--index-url https://pypi.org/simple ')

-                pip_index_urls=['https://pypi.org/simple', 'https://extra.org/simple'],
+                pip_index_urls=[
+                    'https://pypi.org/simple', 'https://extra.org/simple'
+                ],
```

`yapf -i sdk/python/kfp/dsl/component_factory_test.py` with the pinned 0.43.0 should sort it.

`build / runtime-base-images` and the SDK client test are also red, but I would rebase before reading anything into those, since the branch is still behind master and two other PRs opened yesterday are fully green.

droctothorpe:
@vishalmore90: Hey, thank you for your PR! As you are not a member of Kubeflow organization, your PR must link to an issue with 'ready' label. Please read our [CONTRIBUTING.md](https://github.com/kubeflow/pipelines/blob/master/CONTRIBUTING.md), closing this PR

PR #13829, opened 2026-07-24
fix(backend): escape SQL LIKE wildcards in IS_SUBSTRING filters

### LINKED ISSUE #13828: [backend] IS_SUBSTRING list filters treat % and _ as SQL wildcards
### Environment

* How did you deploy Kubeflow Pipelines (KFP)? Any deployment that uses the database pipeline store and the v2beta1 list APIs (standalone, multi-user, local Kind, etc.). Does not apply to Kubernetes-native store-only filtering paths in the same way.
* KFP version: master branch (bug is in `backend/src/apiserver/filter/filter.go`)
* KFP SDK version: N/A. Reproducible via the list filter API or the Go unit test below

### Steps to reproduce

1. Call any list endpoint that supports filters, for example `ListRuns`, with an `IS_SUBSTRING` predicate on a string field like `display_name`.
2. Use a search value that contains a percent sign or underscore, for example `50%` or `a_b`.

Example filter (JSON shape used by the API):

```json
{
  "predicates": [
    {
      "key": "display_name",
      "operation": "IS_SUBSTRING",
      "string_value": "50%"
    }
  ]
}
```

3. Compare results against rows that literally contain `50%` in the name versus rows that merely start with `50`.

Alternatively, run the unit test `TestSubstringFilterEscapesLikeWildcardCharacters` in `backend/src/apiserver/filter/filter_test.go`. It **fails on current master** with:

```
=== RUN   TestSubstringFilterEscapesLikeWildcardCharacters/percent_sign_is_literal
    filter_test.go:1078: IS_SUBSTRING LIKE bind args = [%50%%], want [%50\%%] (unescaped args treat % and _ as SQL wildcards)
=== RUN   TestSubstringFilterEscapesLikeWildcardCharacters/underscore_is_literal
    filter_test.go:1078: IS_SUBSTRING LIKE bind args = [%a_b%], want [%a\_b%] (unescaped args treat % and _ as SQL wildcards)
```

### Expected result

A substring filter should match the literal text the user typed. Searching for `50%` should only match names that actually contain the characters `50%`. Searching for `a_b` should only match names that contain `a_b`, not `axb` or `a1b`.

### Actual result

The filter builder wraps the user value in a SQL LIKE pattern without escaping `%` or `_`. Those characters keep their wildcard meaning in SQL, so the query returns extra rows that do not contain the literal substring.

For `50%` the backend produces the pattern `%50%%`, which matches anything starting with `50`. For `a_b` it produces `%a_b%`, where `_` matches any single character.

The Kubernetes pipeline store applies substring filters with literal `strings.Contains`, so the same filter string can behave differently depending on which store backend is in use.

### Materials and Reference

- Filter builder: `backend/src/apiserver/filter/filter.go` (`AddToSelect`, substring handling around lines 364-372)
- Kubernetes store literal matching: `backend/src/apiserver/filter/filter.go` (`matchesFilter`, substring handling around lines 297-305)
- Existing test that documents unescaped behavior: `backend/src/apiserver/filter/filter_test.go` (substring case expecting `%label_substring%` without escaping `_`)
- Failing regression test for correct behavior: `TestSubstringFilterEscapesLikeWildcardCharacters` in the same file
- Suggested fix: escape `\`, `%`, and `_` in user values before building the LIKE pattern (or use an explicit ESCAPE clause consistently)

false200:
IS_SUBSTRING filter values were interpolated directly into LIKE patterns, so `%` and `_` kept their SQL wildcard meaning. A search for `50%` matched any name starting with `50`, and `a_b` matched `axb`.

Escape backslash, percent, and underscore in user values before building the pattern, and add a unit test for literal matching.

The database pipeline store builds `IS_SUBSTRING` filters with SQL `LIKE`, but user values were wrapped as `%value%` without escaping. Characters `%` and `_` therefore acted as SQL wildcards instead of literal text, returning extra rows. The Kubernetes pipeline store already uses literal `strings.Contains`, so the same filter could behave differently by backend.

This change adds `escapeLikePattern` to escape `\`, `%`, and `_` before constructing the `LIKE` bind argument in `AddToSelect`. Updated existing `AddToSelect` golden expectations for fixture values that contain underscores, and added `TestSubstringFilterEscapesLikeWildcardCharacters` to verify literal matching for `%` and `_`.

**Checklist:**

Fixes #13828


false200:
Hello maintainers, is this pr under review?

PR #13833, opened 2026-07-25
chore(frontend): migrate to ESLint 10 and adopt React Hooks 7 rules

### LINKED ISSUE #13827: frontend: migrate to ESLint 10 and adopt React Hooks 7 correctness rules
## Background

KFP used `frontend/.eslintrc.yaml`, which ESLint 10 no longer supports, and the
published `eslint-plugin-import` release did not declare ESLint 10
compatibility. #13825 landed the prerequisite frontend security updates while
retaining ESLint 9; #13833 now builds directly on that merged baseline.

## Resolution

#13833 implements the complete migration:

- replace `.eslintrc.yaml` with an ESLint 10 flat configuration
- replace `eslint-plugin-import` and its legacy TypeScript resolver with
  `eslint-plugin-import-x`
- upgrade TypeScript ESLint and React Hooks to ESLint-10-compatible releases
- remove `ESLINT_USE_FLAT_CONFIG=false`
- adopt ESLint 10's new core recommended rules and fix every resulting finding
- adopt React Hooks 7 correctness diagnostics without enabling React Compiler
- update frontend contributor and repository guidance

### Why `eslint-plugin-import-x`

`eslint-plugin-import-x` is actively maintained, officially supports ESLint 8,
9, and 10, provides a flat-config-native resolver, and handles package `exports`
directly. This removes the legacy resolver workaround while preserving KFP's
`no-unresolved` and `no-anonymous-default-export` policy.

### Why remove the React and accessibility plugins

This is based on KFP's effective configuration, not an assumption that core
ESLint replaces React or accessibility analysis:

- `eslint-plugin-jsx-a11y` was registered but no rules or presets were enabled.
  Its only repository footprint beyond the dependency was three stale disable
  comments, so removing it does not weaken active accessibility enforcement.
- `eslint-plugin-react` had two explicitly disabled rules and one active rule:
  `react/jsx-no-target-blank` with `allowReferrer: true`. That configuration only
  enforced `noopener`; for KFP's modern browser baseline, `_blank` already
  implies `noopener`.

### ESLint 10 and React Hooks policy

ESLint 10 adds `no-unassigned-vars`, `no-useless-assignment`, and
`preserve-caught-error` to its core recommended configuration. #13833 adopts all
three and fixes the existing findings instead of disabling them.

React Hooks 7 correctness rules are adopted independently of React Compiler:

- correctness rules with no remaining findings are errors
- `set-state-in-effect`, `incompatible-library`, `unsupported-syntax`, and
  `preserve-manual-memoization` are warnings
- the existing `refs`, `immutability`, and `set-state-in-effect` findings are
  fixed
- three compiler-only manual-memoization findings around mutable generated API
  values are narrowly documented and suppressed
- `config` and `gating` remain off

React Compiler itself is not enabled by this work.

## Acceptance criteria

- [x] no `.eslintrc` or `ESLINT_USE_FLAT_CONFIG=false` dependency remains
- [x] the selected import plugin officially supports ESLint 10
- [x] effective React/a11y rules were inventoried before removing dependencies
- [x] all new ESLint 10 core findings are fixed without disabling the rules
- [x] React Hooks 7 correctness diagnostics are enabled with an explicit
      compiler-rollout boundary
- [x] changed behavior has focused regression coverage
- [x] formatting, lint, typechecks, React peer validation, server coverage, and
      the frontend image build pass
- [x] decisions and rationale are documented in #13833

Implementation: #13833. This issue will close when that PR merges.

jeffspahr:
## Summary

- migrate the frontend from legacy `.eslintrc` configuration to ESLint 10 flat
  config
- replace `eslint-plugin-import` and its legacy TypeScript resolver with the
  actively maintained `eslint-plugin-import-x` flat-config resolver
- remove `eslint-plugin-jsx-a11y`, which had no enabled rules, and
  `eslint-plugin-react`, whose only active rule is redundant for KFP's supported
  browser baseline
- adopt ESLint 10's three new core recommended rules and fix every resulting
  finding
- adopt React Hooks 7's flat recommended preset, with explicit rollout
  severities, without enabling React Compiler
- document the new configuration entry point and remove
  `ESLINT_USE_FLAT_CONFIG=false`

### Import resolution

`eslint-plugin-import-x` officially supports ESLint 8, 9, and 10, while the
published `eslint-plugin-import` release currently declares support only through
ESLint 9. `import-x` is also more actively maintained, provides a
flat-config-native resolver, handles package `exports` directly, and removes
KFP's legacy resolver workaround.

The migration preserves the existing `no-unresolved` and
`no-anonymous-default-export` policy under the `import-x` namespace.

### React and accessibility plugins

Core ESLint does not generally replace React or accessibility linting. These
dependencies are removed because KFP's effective configuration received no
material coverage from them:

- `eslint-plugin-jsx-a11y` had no enabled rules or presets. The three associated
  disable comments were stale, so removing the plugin does not weaken active
  accessibility enforcement.
- `eslint-plugin-react` had two explicitly disabled rules and only
  `react/jsx-no-target-blank` enabled with `allowReferrer: true`. That setting
  only enforces `noopener`; modern browsers already imply `noopener` for
  `_blank`, and the upstream rule documentation considers the rule unnecessary
  when legacy browsers are unsupported.

### Rule changes

ESLint 10 adds `no-unassigned-vars`, `no-useless-assignment`, and
`preserve-caught-error` to its core recommended rules. This PR adopts all three
and fixes the existing findings rather than disabling the new checks.

React Hooks 7's flat recommended preset is adopted as a correctness gate
independently of React Compiler:

- `static-components`, `use-memo`, `immutability`, `globals`, `refs`,
  `error-boundaries`, `purity`, and `set-state-in-render` are errors
- `set-state-in-effect`, `incompatible-library`, `unsupported-syntax`, and
  `preserve-manual-memoization` are warnings
- existing `refs`, `immutability`, and `set-state-in-effect` findings are fixed
- three compiler-only manual-memoization findings in `NewRunV2` are narrowly
  documented and suppressed because generated API values may be mutated; #13948
  tracks removing these suppressions
- compiler `config` and `gating` diagnostics remain off

React Compiler itself is not enabled. That remains a separate rollout decision.

The fixes also remove a leaked global click listener, restore the namespace
change hook's one-render signal with a previous-committed-value ref, derive ROC
and run-finished state during render instead of synchronously setting state from
effects, and resolve a compiler-detected declaration-order issue.

## Rebase

Rebased onto master at `5935348057`. Master now contains #13825, so its two
prerequisite commits and the prior merge commit are no longer part of this PR's
history or diff.

Closes #13827.

## Verification

- clean `npm ci` — zero audit findings; master's `brace-expansion` compatibility
  patch applied successfully
- `npm run format:check`
- `npm run lint`
- `npm run typecheck`
- `npm run typecheck:mock-backend`
- `npm run check:react-peers`
- focused changed-area suite — 8 files, 243/244 passed initially; the one
  5-second timeout passed on isolated rerun
- review-fix regression suite — 5 files, 166/166 tests passed
- full UI coverage — 145 files and 2,043 tests passed; five unchanged tests hit
  the existing 5-second timeout under coverage, and the known legacy
  `RecurringRunDetailsV2` XHR-mock-pollution test failed only in the full suite
- affected timeout/XHR files passed separately — 3 files, 75/75 tests
- `npm run test:server:coverage` — 18 files, 354 tests passed, 2 todo
- `npm run build`
- `npm run docker`


frontend/package.json
```diff
@@ -114,6 +114,7 @@
     "build:storybook": "storybook build"
   },
   "devDependencies": {
+    "@eslint/js": "^9.39.5",
```
manaswinidas: Duplicate `@eslint/js` key — drop this `^9.39.5` entry (line 136 already pins `^10.0.1`). JSON keeps the last key, so this is dead/confusing.

frontend/src/lib/KubeflowClient.tsx
```diff
@@ -65,11 +65,16 @@ export class NamespaceContextProvider extends React.Component {
...
+    current: value,
   });
-  return ref.current;
+
+  if (value !== values.current) {
+    setValues({ current: value, previous: values.current });
+    return values.current;
+  }
+
+  return values.previous;
```
manaswinidas: After a change, later renders keep returning `values.previous`, so `useNamespaceChangeEvent()` stays `true` until unmount. Callers Redirect today, but this is a footgun. Please restore one-shot previous semantics (advance previous every render, or clear after the transition).

frontend/eslint.config.cjs
```diff
@@ -0,0 +1,125 @@
...
+      '**/*.test.tsx',
+      'src/generated/**',
+      'src/third_party/mlmd/generated/**',
+      'src/stories/**',
+    ],
+  },
+  {
+    files: sourceFiles,
+    linterOptions: {
+      reportUnusedDisableDirectives: false,
```
manaswinidas: Nit: prefer leaving unused-disable reporting on so stale `eslint-disable` comments stay visible.

frontend/eslint.config.cjs
```diff
@@ -0,0 +1,125 @@
...
+    },
+    rules: {
+      ...eslint.configs.recommended.rules,
+      ...importX.flatConfigs.recommended.rules,
+      ...importX.flatConfigs.typescript.rules,
+      // React Hooks 7 adds React Compiler diagnostics to its recommended preset.
+      // Enable correctness rules independently of compiler adoption.
+      'react-hooks/rules-of-hooks': 'error',
+      'react-hooks/exhaustive-deps': 'warn',
+      'react-hooks/static-components': 'error',
```
manaswinidas: Nit: spread Hooks recommended (or recommended-latest) and override severities/`config`/`gating`, instead of hand-maintaining this rule list.

frontend/src/pages/NewRunV2.tsx
```diff
@@ -315,6 +315,10 @@ function NewRunV2(props: NewRunV2Props) {
     () => getTemplateData(templateString),
     [templateString],
   );
+  /* eslint-disable react-hooks/preserve-manual-memoization --
```
manaswinidas, line 318: Nit: fine for now — please track a follow-up so these `preserve-manual-memoization` suppressions don’t become permanent.

manaswinidas (changes requested):
Thanks for the ESLint 10 / Hooks 7 migration — direction looks good. Requesting changes for the packaging mistake and the `usePrevious` sticky-true semantics; nits inline.

frontend/eslint.config.cjs
jeffspahr: The config now spreads `reactHooks.configs.flat.recommended.rules` and keeps only the intentional KFP rollout overrides for warning/off severities.

frontend/src/lib/KubeflowClient.tsx
jeffspahr: Restored previous-committed-render semantics with an effect-advanced ref. I also added a direct regression test that verifies the signal sequence is `false` on the initial namespace, `true` for the namespace transition, and `false` on the next render with the same namespace.

frontend/eslint.config.cjs
jeffspahr: Removed the unused-disable reporting opt-out. With reporting restored, lint exposed three stale directives in `Utils.tsx` and `mlmd-validator.ts`; those are removed as well, and both UI and server lint pass with zero warnings.

frontend/package.json
jeffspahr: Removed the duplicate `@eslint/js` 9.x entry. The manifest and lockfile now resolve only `@eslint/js` 10.x.

manaswinidas (approved):
Re-reviewed `2d52f70` — prior change requests are addressed.

- duplicate `@eslint/js` removed
- `usePrevious` restored to one-shot previous-commit semantics, with a regression test for `false → true → false`
- unused-disable reporting restored; stale disables cleaned up
- Hooks rules now spread `reactHooks.configs.flat.recommended` with intentional overrides

No remaining merge blockers from my side. PR is still behind `master` — rebase before merge. The `NewRunV2` `preserve-manual-memoization` suppressions remain acceptable debt for a Compiler follow-up.

manaswinidas (approved):
/lgtm
/approve

PR #13838, opened 2026-07-25
fix(backend): log the actual bucket close error, not the outer fetch error

Sharkyii:
## Summary

In `downloadArtifacts` (`backend/src/v2/component/launcher_v2.go`), the deferred `closeNonDefaultBuckets` closure logs `err.Error()` instead of `closeBucketErr.Error()`:

```go
closeNonDefaultBuckets := func(buckets map[string]*blob.Bucket) {
    for name, bucket := range nonDefaultBuckets {
        if closeBucketErr := bucket.Close(); closeBucketErr != nil {
            glog.Warningf("failed to close bucket %q: %q", name, err.Error())
        }
    }
}
```

`err` here is `fetchNonDefaultBuckets`'s return value from earlier in the function, not the bucket-close error. On the success path, `err` is `nil` by the time this deferred closure runs (the only later reassignment of a variable named `err` is inside a nested loop block, which shadows a new local `err` rather than reassigning this one). So if any non-default bucket's `Close()` fails after a successful download — a plausible transient GCS/S3 error — this line calls `.Error()` on a nil `error` interface, which panics. The cleanup code meant to just log a warning ends up crashing the launcher instead.

## Fix

Log `closeBucketErr.Error()`, the value that's actually known to be non-nil at that point.

## Test plan

- `go build ./backend/src/v2/component/...` passes.
- `go test ./backend/src/v2/component/...` passes (no behavior change on any currently passing path).
- No new test added: the existing test harness uses `gocloud.dev/blob/memblob`, an in-memory driver whose `Close()` doesn't fail, so there's no existing way to simulate a bucket-close error without a custom driver implementation. The fix itself is a one-line, unambiguous correction of a wrong variable reference.


hbelmiro (approved):
/ok-to-test
/lgtm
/approve

hbelmiro (commented):
/lgtm
/ok-to-test

PR #12590, opened 2025-12-25
feat(sdk): avoid importing kfp at module import time in dsl.executor

shreeharshshinde:
### Context
The dsl.executor module is executed inside user containers at runtime, but today it imports KFP SDK modules at module import time.
This causes import errors in environments where kfp is not installed and requires the _KFP_RUNTIME environment variable as a workaround.

### What this PR does
This PR reduces module-level coupling between `dsl.executor` and the KFP SDK by:
- Moving KFP SDK imports behind `TYPE_CHECKING` or into function scope
- Ensuring `dsl.executor` can be imported without triggering KFP SDK imports
- Keeping all existing behavior unchanged

### Changes
- Wrapped KFP SDK imports with TYPE_CHECKING to avoid runtime evaluation
- Added lazy imports inside runtime execution paths:
      - `__init__`
      - `assign_input_and_output_artifacts`
      - `make_artifact`
      - `handle_single_return_value`
      - `execute`
      - `create_artifact_instance`
      - `is_parameter`
      - `is_artifact`
- Removed SDK types from runtime-evaluated annotations to prevent import leaks

### Benefits
✅ dsl.executor can be imported without kfp installed
✅ Reduces reliance on _KFP_RUNTIME workaround
✅ Improves runtime robustness and modularity
✅ Backward compatible

### Files changed
`sdk/python/kfp/dsl/executor.py`

### Related issue
#12090

PR #12172, opened 2025-08-22
Remove kfp dependency executor

aniketpati1121:
This PR removes the KFP dependency executor and related checks from the Kubeflow Pipelines SDK. The _KFP_RUNTIME check in kfp/__init__.py has been removed, and necessary updates have been applied to ensure compatibility.

fixes #12090

PR #13840, opened 2026-07-25
feat(sdk): remove _KFP_RUNTIME environment variable and resolve Executor dependency on KFP

### LINKED ISSUE #12090: [feature] Removal of KFP dependency in the Executor
### Feature Area

 /area sdk

As a Kubeflow user I would like the removal of the KFP dependency within the dsl.executor so that the `_KFP_RUNTIME` environment variable can be removed as well, thus having any imports of kfp will not result in an error during the component execution since the dsl imports would not be locked to compile time only.

---

Love this idea? Give it a 👍.

prakharsingh-74:
fixes: #12090 

### Description:

This PR resolves the runtime dependency of the `dsl.executor` on the `kfp` compile-time package, allowing us to safely eliminate the `_KFP_RUNTIME` environment variable entirely. 

Prior to this change, when the component executor ran in a environment where `kfp` was installed with `--no-deps`, importing the SDK module or `kfp.dsl` eagerly resolved compile-time modules that pulled in heavy external dependencies (such as `kubernetes`, `google-cloud-storage`, `requests`, etc.). This caused runtime import failures. The `_KFP_RUNTIME=true` variable was used as a workaround to lock/disable these imports, but restricted importing the SDK during execution.

To fix this:
1. **Lazy Package Attribute Resolution (PEP 562)**: 
   - Replaced eager module-level imports in `kfp/__init__.py` and `kfp/dsl/__init__.py` with dynamic `__getattr__` and `__dir__` loaders. Compile-time modules (like component decorators, pipeline compilation builders, and client bindings) are now loaded lazily only when accessed.
2. **Executor Decoupling**:
   - Refactored `kfp/dsl/executor.py` to import the base `Artifact` directly from `kfp.dsl.types.artifact_types` instead of doing `from kfp import dsl` and using `dsl.Artifact`.
3. **Template Clean Up**:
   - Removed the command prefix `_KFP_RUNTIME=true` from the lightweight Python component command template in `component_factory.py`.
4. **Documentation**:
   - Cleaned up references to `_KFP_RUNTIME` from `AGENTS.md`.

### Checklist:


prakharsingh-74:
@jeffspahr can you review this PR too?

jeffspahr (changes requested):
I think the direction here is promising, but this currently introduces a blocking regression in the exact dependency-free runtime path it is intended to fix.

### Blocking runtime behavior

Lightweight components generate `from kfp.dsl import *` in `component_factory.py`. This change adds all compile-time symbols to `kfp.dsl.__all__` unconditionally, so the generated star import calls `__getattr__` for every lazy symbol. That immediately loads compile-only modules even though KFP is installed with `--no-deps`.

I reproduced this in a clean Python 3.13 environment with the PR source and no third-party dependencies:

- `import kfp` succeeds.
- `import kfp.dsl` succeeds.
- `import kfp.dsl.executor` succeeds.
- `from kfp.dsl import *` fails while resolving `component_decorator`, first with `ModuleNotFoundError: No module named 'docstring_parser'`.

This means a newly generated lightweight component will fail when `executor_main` imports its ephemeral component module. The existing runtime coverage does not catch this because it executes precompiled/static pipeline YAML that still contains `_KFP_RUNTIME=true`.

Please keep the generated runtime import set limited to runtime-safe symbols (or otherwise avoid forcing compile-time names through the star import) and add a regression test that compiles a fresh lightweight component and executes its generated command with KFP installed `--no-deps`.

### Parent issue relevance and scope

#12090 is still relevant: current master still installs KFP with `--no-deps`, prefixes executor execution with `_KFP_RUNTIME=true`, and conditions package-level imports on that variable.

However, I do not think this PR should close the issue without clarifying what “remove the KFP dependency from the executor” means. This patch removes the broad `from kfp import dsl` import, but `executor.py` still imports several `kfp.dsl.*` modules, the runtime still installs KFP, and execution remains `python -m kfp.dsl.executor_main`. It removes the environment-variable workaround through package-level lazy loading; it does not achieve the originally discussed long-term target of making the executor independent of the KFP distribution.

Suggested acceptance criteria for #12090:

- Newly generated lightweight components execute with KFP installed `--no-deps`.
- `_KFP_RUNTIME` is no longer required.
- Plain `import kfp` and `import kfp.dsl` remain runtime-safe.
- Compilation APIs remain available without eager loading.
- Explicitly decide whether executor extraction/embedding outside the KFP distribution is in scope.

### Context from earlier attempts

- #12081 added an opt-in `compile_time_imports` decorator flag. That was a workaround rather than decoupling. The maintainer discussion converged on keeping runtime containers lean and tracking complete executor decoupling as the preferred target. The PR later stalled and was self-closed.
- #12157 was incomplete: its final diff changed only `executor_test.py`, importing a `BaseArtifact` that it did not add. It received no human review and was self-closed.
- #12172 removed the root guard but did not isolate `dsl.executor` or address the `kfp.dsl` compile-time gate, and it mixed in unrelated formatting/version/CLI changes. The review explicitly noted that it missed the issue scope; it was self-closed.
- #12590 is still open but stale. It moves imports into function bodies, while retaining KFP imports and `_KFP_RUNTIME`; it has not received substantive maintainer review.
- #9886 is the important baseline: it deliberately introduced the dependency-free runtime install to reduce startup latency, dependency conflicts, transitive runtime bugs, and CVE exposure. Any replacement needs to preserve those properties.

There are also two minor `git diff --check` failures from new blank lines at EOF in the modified `kfp/__init__.py` and `kfp/dsl/__init__.py` files.

sdk/python/kfp/dsl/__init__.py
```diff
@@ -43,6 +43,28 @@
     'PIPELINE_JOB_SCHEDULE_TIME_UTC_PLACEHOLDER',
     'WORKSPACE_PATH_PLACEHOLDER',
     'run_notebook',
+    'component',
```
jeffspahr, line 46: **[P1] Preserve the dependency-free generated component import path**

These compile-time names are now part of `__all__` unconditionally, but lightweight components generate `from kfp.dsl import *`. That star import therefore calls `__getattr__` for every entry here and loads compile-only modules in the `--no-deps` runtime. In a clean environment, it fails first while importing `component_decorator` because `docstring_parser` is absent.

Please keep the runtime star-import surface limited to runtime-safe symbols (or replace the generated star import with explicit runtime imports), and cover a freshly compiled component executing without optional SDK dependencies.

jeffspahr:
@prakharsingh-74 this looks like a thorny issue with several past attempts.  I'd discuss it with @zazulam before going much further.  I'm happy to help review, but I'm not familiar with the original issue, and I don't have /sdk approval permission.

prakharsingh-74:
> @prakharsingh-74 this looks like a thorny issue with several past attempts. I'd discuss it with @zazulam before going much further. I'm happy to help review, but I'm not familiar with the original issue, and I don't have /sdk approval permission.

So should i left the PR open or what?
Please let me know after discussing with @zazulam that should i work on it or not

jeffspahr:
I'm suggesting _you_ work with @zazulam  on it.

prakharsingh-74:
> I'm suggesting _you_ work with @zazulam on it.

Ok sure, so can you just tell him to be active on this PR?

prakharsingh-74:
closing this PR due to protoc files mismangement

PR #13855, opened 2026-07-25
chore(test): remove obsolete image builder workflows

jeffspahr:
## Summary

- remove the unused `test/imagebuilder` image and build script
- remove the orphaned Argo workflow manifests and their uncalled presubmit launcher
- remove stale sample-test documentation and the obsolete requirements updater entry

## Why

PR #13850 exposed that the legacy image builder no longer works with the
current Cloud SDK image. Repository and CI history show that the builder should
be removed rather than modernized:

- no current GitHub Actions workflow or repository script invokes it
- the Prow E2E and upgrade jobs that used these manifests were removed after
  their GitHub Actions migrations in 2024
- the remaining manifests reference test Dockerfiles and scripts that have
  already been deleted
- the checked-in builder image tag dates to 2020

Removing the dormant path avoids adding new CI coverage and dependency updates
for infrastructure that no longer has an upstream KFP execution path.

## Verification

- `bash -n hack/update-all-requirements.sh`
- verified every directory remaining in `hack/update-all-requirements.sh` exists
- searched the repository for the deleted manifests, launcher, builder path,
  image name, and sample-test directory; no nonhistorical references remain


juliusvonkohout:
Thank you, removing all the dead weight is very important
/lgtm

Issue #13835, opened 2026-07-25
[feature] Unify pod lifecycle failure classification so timeout config and UI don't drift apart


Sharkyii:
### Feature Area

/area frontend
/area backend

  

### What feature would you like to see?

While going through [#12843](https://github.com/kubeflow/pipelines/issues/12843) (the pod lifecycle failure visibility issue), I noticed the existing open PRs are each quietly building their own version of the same thing, and they don't agree with each other.

[#12989](https://github.com/kubeflow/pipelines/pull/12989) defines the three timeout categories (Provisioning / Runtime / Node) as time.Duration constants. [#13102](https://github.com/kubeflow/pipelines/pull/13102) defines the same three categories again, but as env-var name strings instead different shape, same idea, done twice. Separately, [#13097](https://github.com/kubeflow/pipelines/pull/13097) adds actual failure detection, but it's a flat yes/no check - OOMKilled, ImagePullBackOff, CrashLoopBackOff, all of it, gets collapsed into one hardcoded label ("PodLifecycleFailure") with no category attached at all.

I'd like to propose pulling this into one shared place instead - a single function, something like ClassifyPodFailure(message string) (category, reason string), that takes a raw Kubernetes/Argo failure string and maps it to a category. The timeout logic and the UI-facing code would both call into this one function rather than each keeping their own copy of the same mapping.

### What is the use case or pain point?

This isn't hypothetical - it's already happened. Two different contributors built two different, incompatible timeout taxonomies on the same issue, in parallel, without either knowing about the other's PR. If both merge as-is, whichever one lands first becomes the accidental "correct" version, and the other has to be reworked around it. A shared classifier means a failure pattern only needs to be added or changed in one place, and the timeout side and the UI side can never quietly drift apart from each other.

### Is there a workaround currently?

Not really - this isn't a bug in any single PR, it's a coordination gap between them. Nothing's broken today because none of these have merged yet, but left alone, the drift just gets locked in once one of them does.

Related: [#13401](https://github.com/kubeflow/pipelines/issues/13401) (Unschedulable pods hanging forever - a real case this classification would feed into), [#12989](https://github.com/kubeflow/pipelines/pull/12989), [#13102](https://github.com/kubeflow/pipelines/pull/13102), [#13097](https://github.com/kubeflow/pipelines/pull/13097)

---

Love this idea? Give it a 👍.

Sharkyii:
@mprahl @alyssacgoins would love your view on this

[...]

PR #13863, opened 2026-07-26
feat(backend): add ClassifyPodFailure taxonomy for pod lifecycle failures

Sharkyii:

Adds ClassifyPodFailure(message string) (category, matchedReason string) in backend/src/common/util, mapping raw Kubernetes/Argo failure strings (ImagePullBackOff, OOMKilled, CrashLoopBackOff, NodeLost, etc.) into the Provisioning/Runtime/Node categories described in #12843.

Going through the existing PRs on that issue, #12989 and #13102 each independently defined their own timeout category constants (different names, different types, for the same three categories), and #13097's failure detection is a flat boolean with no category info at all. This is meant to be a single shared place both the timeout logic and the UI-facing code can use instead of drifting apart. It's a pure, standalone function with no dependency on NodeStatus or any of those PRs, so it doesn't block on any of them merging first.

More detail in #13835.

Also adds two functions that build on ClassifyPodFailure to resolve which message should actually be classified, since the raw Argo node.Message is not always safe to use as is.

ResolveNodeLifecycleMessage walks a task's Argo node tree with three rules a plain read of node.Message misses.

Only a Pod-type node's own message counts as a pod lifecycle failure. Control-flow nodes such as Retry or DAG carry their own orchestration messages, for example "No more retries left" once a retry group is exhausted, and that message is not a pod failure. Treating any non-empty message as a failure can hide the real reason, since a Retry node's own message would take priority over its last attempt's actual OOMKilled message.

```mermaid
flowchart LR
    subgraph before ["Before, raw node.Message read as is"]
        direction TB
        B1["Retry node<br/>Message: No more retries left"] --> BR["Shown to user:<br/>No more retries left"]
    end
    subgraph after ["After, ResolveNodeLifecycleMessage"]
        direction TB
        A1["Retry node<br/>Message: No more retries left, ignored, not Pod type"] --> A2["attempt(0), Pod<br/>Message: OOMKilled"]
        A2 --> AR["Shown to user:<br/>OOMKilled"]
    end
    style BR fill:#fce8e6,stroke:#c5221f
    style AR fill:#e6f4ea,stroke:#188038
```

For a retried task, the most recent attempt is also authoritative, not the first non-empty one found. Argo appends retry attempts oldest first, so a naive lookup can show a stale failure from an early attempt while a later attempt is already healthy and running.

```mermaid
flowchart LR
    subgraph before ["Before, first non-empty child wins"]
        direction TB
        B1["attempt(0), Failed, OOMKilled"] --> BR["Shown to user:<br/>OOMKilled, stale"]
    end
    subgraph after ["After, latest attempt is authoritative"]
        direction TB
        A1["attempt(1), Running, no message"] --> AR["Shown to user:<br/>nothing, correct"]
    end
    style BR fill:#fce8e6,stroke:#c5221f
    style AR fill:#e6f4ea,stroke:#188038
```

A Pod-type node's message is only trusted when its own phase is not NodeError. NodeFailed means the container itself failed, a real pod lifecycle failure. NodeError means Argo's own controller failed to manage the pod, for example a Kubernetes API error while creating it, not the pod itself failing, so its message should not be classified as a pod failure reason.

ResolveMissingTaskDriverFailure covers a case none of the current approaches on this issue handle. KFP's compiler emits two separate DAG tasks per component, a driver task that resolves inputs and builds the pod spec, and the executor task that runs the user's container, with the executor depending on the driver. If the driver itself fails, for example a kubernetes.mount_pvc() reference to a PVC that does not exist, Argo never creates the executor node, so there is no Pod-type node anywhere in the tree for anything to point a diagnostic at.

```mermaid
flowchart LR
    subgraph before ["Before, no fix"]
        direction TB
        B1["Executor node does not exist"] --> BR["Shown to user:<br/>nothing, task just sits"]
    end
    subgraph after ["After, ResolveMissingTaskDriverFailure"]
        direction TB
        A1["Driver Pod, Failed<br/>PVC typo-name not found"] --> AR["Shown to user:<br/>the driver's real message"]
    end
    style BR fill:#fce8e6,stroke:#c5221f
    style AR fill:#e6f4ea,stroke:#188038
```

Both functions are covered by table-driven tests, including a regression test for the retry-staleness case above and a case confirming two sibling tasks under the same parent are not confused with each other. Node shapes in the tests are modeled on a real workflow's live status.nodes, not just reconstructed from compiler source.

**Checklist:**

### DIFF
--- a/backend/src/common/util/pod_failure_classifier.go
+++ b/backend/src/common/util/pod_failure_classifier.go
@@ -36,6 +36,34 @@ const (
 	PodFailureCategoryNone PodFailureCategory = ""
 )
 
+// PodFailureSignalSource identifies where a PodFailureSignal's Reason came
+// from.
+//
+// ImagePullBackOff, CrashLoopBackOff, OOMKilled and the other reasons below
+// are all readable from the pod's own status (containerStatuses[].state).
+// Unschedulable is not: there is no container yet, so Argo leaves the pod
+// Pending with no message, and the only signal is a FailedScheduling Event
+// recorded against the pod. Carrying Source alongside Reason means a future
+// caller that watches pod Events can classify through this same function
+// without another signature change (see #12843).
+type PodFailureSignalSource string
+
+const (
+	// PodFailureSignalSourcePodStatus means Reason was read from the pod's
+	// own status (e.g. a container waiting/terminated reason).
+	PodFailureSignalSourcePodStatus PodFailureSignalSource = "PodStatus"
+	// PodFailureSignalSourcePodEvent means Reason was read from a
+	// Kubernetes Event recorded against the pod (e.g. FailedScheduling).
+	PodFailureSignalSourcePodEvent PodFailureSignalSource = "PodEvent"
+)
+
+// PodFailureSignal is the input to ClassifyPodFailure: a raw failure reason
+// or status message, and where it was read from.
+type PodFailureSignal struct {
+	Reason string
+	Source PodFailureSignalSource
+}
+
 type podFailurePattern struct {
 	substring string
 	category  PodFailureCategory
@@ -60,16 +88,22 @@ var podFailurePatterns = []podFailurePattern{
 	{"Evicted", PodFailureCategoryNode},
 }
 
-// ClassifyPodFailure inspects a raw Kubernetes/Argo failure reason or status
-// message and classifies it into a PodFailureCategory, along with the
-// specific substring that matched. It returns PodFailureCategoryNone and an
-// empty reason if the message does not match any known pattern.
-func ClassifyPodFailure(message string) (category PodFailureCategory, matchedReason string) {
-	if message == "" {
+// ClassifyPodFailure inspects a pod lifecycle failure signal and classifies
+// it into a PodFailureCategory, along with the specific substring that
+// matched. It returns PodFailureCategoryNone and an empty reason if the
+// signal does not match any known pattern.
+//
+// Matching is currently substring-based against signal.Reason regardless of
+// signal.Source; no pattern here is sourced from a pod Event yet, since
+// nothing in this codebase watches pod Events today. Source is carried so
+// that whoever adds that watch (see #12843) can pass a PodEvent-sourced
+// signal, such as a FailedScheduling reason, through this same function.
+func ClassifyPodFailure(signal PodFailureSignal) (category PodFailureCategory, matchedReason string) {
+	if signal.Reason == "" {
 		return PodFailureCategoryNone, ""
 	}
 	for _, pattern := range podFailurePatterns {
-		if strings.Contains(message, pattern.substring) {
+		if strings.Contains(signal.Reason, pattern.substring) {
 			return pattern.category, pattern.substring
 		}
 	}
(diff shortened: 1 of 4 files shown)


alyssacgoins:
Hey @Sharkyii thanks for your contribution! I've added a comment on the related issue here: https://github.com/kubeflow/pipelines/issues/13835#issuecomment-5105031436

dobbydobap:
Useful to have this as a standalone pure function, and factoring out the duplicate category constants from #12989 and #13102 is the right call.

One thing about the input contract, since it is hard to change once callers exist. `ClassifyPodFailure(message string)` keys on the raw message, which works for anything Argo writes a message for. `Unschedulable` is the exception: there is no container yet, so no `containerStatuses` entry, and Argo leaves the node `Pending` with no message. The signal exists only as a `FailedScheduling` Event on the pod, so no substring added to this function can reach it.

If the parameter were a small struct carrying the reason and its source (pod status vs pod event) rather than a bare message, the same taxonomy would cover all three categories in #12843, and whoever adds the Event watch later could plug into it without changing the signature or the existing callers.

Happy to be corrected if Argo does populate `node.Message` for unschedulable pods. I checked `backend/src/` and found no `FailedScheduling` or `Unschedulable` handling anywhere, and the only core/v1 Events usage is the scheduled-workflow controller.

Sharkyii:
You're right that unschedulable pods never get a message on the pod status, that only shows up as a FailedScheduling Event, so no amount of substring matching on the current signature gets there. Agreed on the struct instead of a bare string, something like:
```
type PodFailureSignal struct {
    Reason string
    Source SignalSource // PodStatus | PodEvent
}
```
I'll swap the signature and update the callers from #12989/#13102 in this PR. Since dobbydobap's taking the Event-watch piece separately, I won't touch that here, this PR just makes sure the classifier has a place for it to plug into later without another breaking change.

dobbydobap:
Rename to `reason` reads right, and dropping `Source` was the correct call since the signals all live on the pod.

One gap though. The doc comment says node failures are readable via the `DisruptionTarget` condition, but the reason values Kubernetes writes there don't match the table:

```go
// k8s.io/api v0.36.3, the version go.mod pins
PodReasonPreemptionByScheduler = "PreemptionByScheduler"
PodReasonTerminationByKubelet  = "TerminationByKubelet"
```

`strings.Contains("PreemptionByScheduler", "Preempted")` is false, since `Preempt` is followed by `ion`. Same for `TerminationByKubelet` against `Evicted`. Pass either condition reason in and you get `PodFailureCategoryNone`. `Unschedulable` is fine, matching `PodReasonUnschedulable` exactly, and `Evicted` works off `pod.Status.Reason` rather than the condition.

I was going to suggest swapping in the constants, but #13952 makes that the wrong fix. It mirrors this table into `StatusUtils.ts` and calls `classifyPodFailure(node.message)`, so there the input really is prose and `Preempted`/`Evicted` are the right things to match. The tests here feed prose too:

```go
reason: "pod was Preempted to make room for a higher priority pod",
```

So the table is being asked to serve two different inputs: an Argo message on the frontend, a condition reason per this doc comment. Adding the API constants alongside the prose substrings covers both, and noting which field each pattern is meant to match would keep the Go and TS copies from drifting.

Sharkyii:
Good catch, checked it against k8s.io/api v0.36.3: PreemptionByScheduler and TerminationByKubelet are the only DisruptionTarget reasons that exist, and neither contains Preempted/Evicted. Evicted actually comes off pod.Status.Reason, a different field.

Agreed a straight swap would break #13952 since it calls this table on Argo's message, not a condition reason.

simple fix is - keep the prose substrings, add the two constants, and fix the doc comment to say which pattern matches which input.
pushing it shortly.

Issue #13766, opened 2026-07-20
[feature] Migrate KFP documentation from kubeflow.org to existing KFP Read The Docs site


alyssacgoins:
### Feature Area

This issue covers documentation, which covers the following areas:
/area frontend
/area backend
/area sdk

### What feature would you like to see?
Migrate KFP documentation SOT from [kubeflow.org/docs/components/pipelines](https://www.kubeflow.org/docs/components/pipelines/) to [Pipelines.kubeflow.org](http://pipelines.kubeflow.org/). The readthedocs site allows for versioned documentation. 
* Note  that the KFP SDK documentation currently lives at [kubeflow-pipelines.readthedocs.io](https://kubeflow-pipelines.readthedocs.io/en/release-2.17/). However, Pipelines.kubeflow.org](http://pipelines.kubeflow.org/) would match the subdomain formatting used by the following KF subprojects:
    * https://trainer.kubeflow.org/en/latest/ 
    * https://sdk.kubeflow.org/en/latest/ 
    * https://spark.kubeflow.org/en/latest/ 
    * https://kale.kubeflow.org/en/latest/

The migration process is outlined as follows:
1. First, all KFP documentation will be added to versionthedocs. Note that v1 docs will not migrate.

* Note that this issue does not propose a major change to the documentation structure implemented on the kubeflow.org - the majority of pages will simply be copied over. Any major changes to page structure/information will be a result of the shift to version-based documentation - on pages for earlier versions, information included in more recent updates will be removed.  

	
2. Once all documentation is publicly available on the readthedocs site, [kubeflow.org](http://kubeflow.org/) KPF documentation is replaced with links to readthedocs
3. Add redirects such that when a user clicks a kubeflow.org KFP search result, they are rerouted to the readthedocs page.

### What is the use case or pain point?
1. Providing a single SOT for KFP documentation reduces friction for users trying to understand and run KFP end-to-end, beyond only working with the SDK. In addition, versioned documentation is easier to follow. 
2. This readthedocs move is happening across multiple Kubeflow components: [SDK](https://www.kubeflow.org/docs/components/sdk/), [Trainer](https://www.kubeflow.org/docs/components/trainer/) and [Kale](https://www.kubeflow.org/docs/components/kale/) each document via readthedocs, and providing the necessary links on the [kubeflow.org](http://kubeflow.org/) website.

### Is there a workaround currently?
[...]

PR #13864, opened 2026-07-26
docs: add unified Kubeflow Pipelines documentation site on Read the Docs

Priyanshu-u07:
## Description:
Migrates the Kubeflow Pipelines documentation from kubeflow.org into a unified Sphinx site (Furo + MyST Markdown) published via Read the Docs, alongside the existing SDK API reference.

Part of #13766 

## What's included

- All non-v1 pages migrated: Overview, Getting Started, Interfaces, Concepts, User Guides, Operator Guides, Reference.
- The existing SDK API reference is retained and nested as the Python SDK section, so one site now serves both guides and the
  API reference.
- Landing page, navigation, Contributor Guide and Community section.
- A docs CI workflow (`.github/workflows/docs.yml`) that builds with warnings treated as errors.

## Structure
The unified site lives at `docs/ ` (repo root), with the existing SDK reference nested inside it. `.readthedocs.yml` is repointed from `docs/sdk/conf.py` to `docs/conf.py`. Happy to restructure if maintainers prefer it elsewhere.

## SDK docstring fixes
Two SDK files have docstring-only changes (no logic change). The API reference is generated from these docstrings and both produced Sphinx warnings that fail the strict build:

- `sdk/python/kfp/dsl/placeholders.py`: `else_` was parsed by RST as a link reference; now an inline literal
- `sdk/python/kfp/dsl/pipeline_config.py`: the `size` attribute was documented on the class rather than the property; moved to the property, and a bare URL converted to a proper link

## Demo

https://github.com/user-attachments/assets/a7b85f25-9d92-493d-b604-eff23b3c19c3

## How to test

**From a clone of `kubeflow/pipelines`:**

```bash
git fetch origin pull/13864/head:pr-13864
git checkout pr-13864
pip install -r docs/requirements.txt   # from the repo root
cd docs
```
**Build and preview with live reload → http://localhost:8000**

```bash
make docs-serve
```
Other targets (from docs/):
```bash
make docs            # one-off build into docs/_build/html
make docs-test       # strict build (warnings = errors), same as CI
make docs-linkcheck  # check external links
```

**Read the Docs preview:** 
https://priyanshu-kfp-docs.readthedocs.io/en/latest/

## Note
 SDK API-reference links: I migrated the SDK API-reference links as-is, so they currently use version-pinned absolute URLs. These could instead be Python-domain cross-references. The pattern already used elsewhere in these docs, so they track the reader's version and don't break when the docs move. I kept it out to keep the diff focused. **Would you prefer I do it into this PR or do it as a fast follow-up?**


docs/reference/api/kubeflow-pipeline-api-spec.md
```diff
@@ -0,0 +1,38 @@
+# Pipelines API Reference (v2beta1)
```
alyssacgoins, line 1: This page does not load visuals the same way https://www.kubeflow.org/docs/components/pipelines/reference/api/kubeflow-pipeline-api-spec/ does - can you investigate why that may be?

docs/reference/sdk.md
```diff
@@ -0,0 +1,3 @@
+# Pipelines SDK Reference
+
+See the {doc}`KFP SDK v2 reference documentation </sdk/source/kfp>`.
```
alyssacgoins: I'm fine with this page being removed, since this site contains all SDK documentation. We can just update the kubeflow/website redirect to point towards `sdk/source/kfp` instead of here.
@mprahl what are your thoughts?

docs/reference/architecture.md
```diff
@@ -0,0 +1,8 @@
+# Kubeflow Pipelines Architecture
```
alyssacgoins, line 1: It looks like you've added this page - I don't see it in the original kubeflow/website docs. Do you have a particular reason?

docs/user-guides/migration.md
```diff
@@ -0,0 +1,701 @@
+# Migrate to Kubeflow Pipelines v2
```
alyssacgoins, line 1: Since we're not including the v1 docs here, I don't think we need to include the v2 migration docs. 
@mprahl @HumairAK @droctothorpe @zazulam what are your thoughts?

alyssacgoins (commented):
Thanks for your work here, @Priyanshu-u07 - added some comments. In addition - 
- Can you update the docs tree structure so that the **Operator Guides** section appears above **User Guides**?
- You've included a section on `GenAI Use Cases` - while I think this is a helpful section that can be added in a future PR, the goal of this work is just to migrate the existing work - I don't think this addition makes sense here

Priyanshu-u07:
Thanks @alyssacgoins for the feedback! 
> * You've included a section on `GenAI Use Cases` - while I think this is a helpful section that can be added in a future PR, the goal of this work is just to migrate the existing work - I don't think this addition makes sense here

The GenAI Use Cases page is actually pre-existing SDK content. It's already published on the docs site: https://kubeflow-pipelines.readthedocs.io/en/release-2.17/source/genai.html. I just surfaced it in the Python SDK navigation alongside the other SDK pages (Quickstart, Install, API Reference, CLI). Happy to remove it from the nav if you'd still prefer it not appear here.

docs/reference/api/kubeflow-pipeline-api-spec.md
Priyanshu-u07: Investigated this. The original kubeflow.org page embeds Swagger UI via a Hugo `swaggerui-inline` shortcode, which doesn't translate to MyST markdown, so only the surrounding text came across. I've reproduced it in Sphinx: an embedded Swagger UI pointing at the v2beta1 spec, so it now renders the same interactive explorer.
A few implementation notes:

- Swagger UI library: loaded from a CDN (jsDelivr), same as the current kubeflow.org page. I can host it locally in `_static/` instead if you'd prefer no external dependency.
- Spec file: served from a copy in `docs/_static/` so it loads from the same origin, which avoids CORS. That duplicates the one in `backend/`, so I can copy it during the build instead if you'd rather not keep a duplicate.
- Styling: rendered it as a light panel (readable in both light/dark site themes).

Let me know if this approach works, or if you'd like me to change any of these.
image

docs/reference/architecture.md
Priyanshu-u07: Yes, this page is new. The content isn't something I wrote. It's the existing Kubeflow pipelines architecture doc from the SDK docs (`docs/sdk/Architecture.md`), already published on the SDK site but excluded from this build. I recreated it under Reference to keep the diagram discoverable. Happy to remove it if you'd prefer to keep this to the migration.

docs/user-guides/migration.md
mprahl: Sounds good to me.

mprahl: Actually, on second thought, I think we should include it. Many users are still using v1 APIs on KFP v2. They'll need a migration guide on KFP v3.

docs/reference/api/kubeflow-pipeline-api-spec.md
alyssacgoins: Yep, the design in this screenshot looks good

alyssacgoins:
@Priyanshu-u07 thanks for pointing that out - yes, let's keep it in that case

docs/reference/architecture.md
alyssacgoins: Thanks :+1: let's keep it

docs/reference/sdk.md
alyssacgoins: @Priyanshu-u07 ^

Priyanshu-u07: Done. Removed the page and repointed the link to the Python SDK section.

alyssacgoins:
Looks good - one small thing - can you squash your commits to 1-2 commits? After that, I'll add `/lgtm`

alyssacgoins:
@mprahl @droctothorpe lgtm from my end, can I get a review from one/both of you here?

Priyanshu-u07:
@alyssacgoins Squashed to a single commit. Thanks!

droctothorpe:
This is fantastic, @Priyanshu-u07. Thanks for tackling!

Some notes:

- **`.github/workflows/docs.yml:33-39, 55-61`** — use the repository's `.github/actions/setup-python-pip-cache` action instead of direct `actions/setup-python` and `pip install`.
  - `docs/agents/ci.md` requires distinct cache scopes and hashes for installed requirements files.
  - Both new jobs reinstall identical dependencies without the required cache convention.

- **`docs/agents/ci.md`** — update this workflow guide with the new documentation CI lane.
  - The guide explicitly requires an update when workflows change.

- **`docs/_static/js/{topnav,sidebar-toggle,external-links}.js`** — add minimal browser-behavior coverage.
  - New DOM event behavior is not exercised by the Sphinx build.
  - Repository conventions require coverage for non-trivial changed behavior.

- **Missing MLflow migration.** Issue [#13766](https://github.com/kubeflow/pipelines/issues/13766) requires all non-v1 KFP docs to migrate, but the PR omits the existing MLflow operator guide and user-guide integration pages.
  - Source pages exist in `kubeflow/website` under `content/en/docs/components/pipelines/.../mlflow`.
  - The new operator and user-guide toctrees also do not link them.

Priyanshu-u07:
Thanks @droctothorpe for the review!
>docs/_static/js/{topnav,sidebar-toggle,external-links}.js — add minimal browser-behavior coverage.

Since `docs/` has no JS test setup, I'm thinking of adding coverage for the three files with a small Vitest suite under `docs/` along with a CI job to run it. Just wanted to confirm: does that approach sound good, or would you prefer a different path?

droctothorpe:
> **`docs/_static/js/{topnav,sidebar-toggle,external-links}.js`** — add minimal browser-behavior coverage.
> 
> * New DOM event behavior is not exercised by the Sphinx build.
> * Repository conventions require coverage for non-trivial changed behavior.

Disregard that suggestion. It's overkill.

Priyanshu-u07:
@droctothorpe Addressed all the review comments:

- Both docs jobs now use `setup-python-pip-cache` with a `docs` cache scope and the requirements hash.
- Documented the docs CI lane in `docs/agents/ci.md`.
- Migrated the MLflow pages. For context, these were merged recently, after this migration was authored, so they weren't in the source at the time.

PTAL when you get a chance. Thank you!

droctothorpe:
/approve
/lgtm

Thanks, @Priyanshu-u07!

PR #13865, opened 2026-07-26
fix(ci): constrain fastjsonschema on Python 3.9

jeffspahr:
## Summary

- constrain `fastjsonschema` to releases before 2.22 on Python 3.9
- apply the constraint to SDK development, default notebook component runtimes,
  and the published `kfp[notebooks]` extra
- leave dependency resolution unchanged on Python 3.10 and newer
- preserve explicit `packages_to_install` behavior

## Root cause

`nbformat==5.10.4` allows `fastjsonschema>=2.15`. Version 2.22.0 contains
Python 3.10 union syntax while its package metadata still permits Python 3.9.

The initial development-requirements constraint fixed the CI job environment,
but notebook components create an isolated runtime environment from their
default package list. Both Python 3.9 jobs therefore still installed the
unconstrained release inside notebook execution and failed while importing
`nbformat`.

Upstream has merged a minimum-Python metadata fix, but no corrected release has
been published and 2.22.0 has not yet been yanked. The conditional bound can be
removed after upstream resolution is reliable.

## Verification

- Python 3.9 notebook component decorator suite: 8 passed
- Python 3.9
  `TestRunLocalPipeline::test_notebook_component_local_exec`: passed
- Python 3.9
  `TestSubProcessRunner::test_execution[Notebook Component Simple]`: passed
- `kfp[notebooks]` dry-run resolves `fastjsonschema==2.21.2` on Python 3.9
- YAPF check on all changed Python files
- `git diff --check`

### DIFF
--- a/sdk/python/kfp/dsl/component_factory.py
+++ b/sdk/python/kfp/dsl/component_factory.py
@@ -661,7 +661,11 @@ def create_notebook_component_from_func(
     # Resolve default packages for notebooks when not specified
     if packages_to_install is None:
         packages_to_install = [
-            'nbclient>=0.10,<1', 'ipykernel>=6,<7', 'jupyter_client>=7,<9'
+            'nbclient>=0.10,<1',
+            'ipykernel>=6,<7',
+            'jupyter_client>=7,<9',
+            # 2.22.0 uses Python 3.10 union syntax but permits Python 3.9.
+            'fastjsonschema<2.22; python_version < "3.10"',
         ]
 
     # Validate notebook path and determine relpath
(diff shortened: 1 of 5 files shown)


zazulam:
I'm fine with this, but we should revisit when we are gonna remove support for python 3.9, maybe for 3.0 release 🤷‍♂️ 
/lgtm

jeffspahr:
Apparently [we'll support it until Oct 2026](https://github.com/kubeflow/pipelines/blob/2ccd057065e1e3cc848e601de0436ea2b2abe6f3/sdk/python/kfp/__init__.py#L29), but I agree 3.0 seems like the right place to drop it.

PR #13871, opened 2026-07-26
fix(sdk): surface the cause when loading kube config fails. Fixes #13870

### LINKED ISSUE #13870: [sdk] Failed kube config load discards the cause and swallows KeyboardInterrupt
### Environment

* KFP version: master (reproduced against `2.15.2`)
* KFP SDK version: `kfp==2.15.2`
* All dependencies version:

```
kfp                      2.15.2
kfp-pipeline-spec        2.17.0
kfp-server-api           2.17.0
kubernetes               30.1.0
```

### Steps to reproduce

When `kfp.Client()` is constructed without a `host` and outside a cluster, it falls back to
loading the local kube config. That fallback is wrapped in a bare `except:` that discards the
cause and reports every failure with the same fixed string
([`client.py#L313-L318`](https://github.com/kubeflow/pipelines/blob/master/sdk/python/kfp/client/client.py#L313-L318)):

```python
try:
    k8s.config.load_kube_config(
        client_configuration=config, context=kube_context)
except:
    print('Failed to load kube config.')
    return config
```

Any of the following produce byte-identical output:

```python
import kfp

# 1. no kubeconfig on the machine
# 2. kubeconfig present but malformed
# 3. kubeconfig fine, but the requested context does not exist
kfp.Client(kube_context='does-not-exist')
```

```
Failed to load kube config.
```

The underlying `kubernetes.config.ConfigException` carries the actionable detail in each case
(`Invalid kube-config file. No configuration found.`, `Context does-not-exist not found in
kube-config file`, and so on), but it is discarded.

Because the method then returns the partially populated `Configuration` rather than raising,
the first symptom the user actually sees is a later connection error against the default host,
detached from the real cause.

There are two further consequences of the same block:

1. A bare `except:` catches `BaseException`, so `KeyboardInterrupt` and `SystemExit` are
   swallowed during config resolution.
2. The message is written to stdout with `print`, so it cannot be suppressed, captured, or
   leveled by callers that configure `logging`. It is the only `print` in `client.py` used for
   an error; the other four are deliberate user-facing links.

The in-cluster probe immediately above (`client.py#L303-L306`) uses a bare `except:` as well.

### Expected result

The failure should identify its cause, and should be routed through `logging` so callers can
control it — consistent with `_get_config_with_default_credentials` in the same class, which
already handles the analogous case this way:

```python
except Exception:
    logging.warning('Failed to set up default credentials. Proceeding'
                    ' without credentials...')
```

Concretely:

* `except:` → `except Exception:`, so `KeyboardInterrupt` propagates.
* `print(...)` → `logging.warning(...)`, including the underlying exception.
* The in-cluster probe should log at debug level rather than warn, since running outside a
  cluster is the expected case on that path and a warning there would be a false alarm.

Loading the kube config should remain best effort — it should keep returning the configuration
rather than raising, so existing callers that recover downstream are unaffected.

### Materials and Reference

* `sdk/python/kfp/client/client.py`, `Client._load_config`
* Existing precedent in the same class: `Client._get_config_with_default_credentials`

---

tolgatuncoglu:

`Client._load_config` wrapped both `load_incluster_config()` and `load_kube_config()` in a bare
`except:`, and reported the second failure with `print('Failed to load kube config.')`.

That single line has three separate problems:

| | Before | After |
|---|---|---|
| Cause | discarded — missing, malformed, and unknown-context kubeconfigs were byte-identical | included in the message |
| Interrupts | `except:` catches `BaseException`, swallowing `KeyboardInterrupt` / `SystemExit` | `except Exception:` — interrupts propagate |
| Destination | `print` to stdout; not suppressible or capturable | `logging.warning`, controllable by the caller |

The third one matters most in practice: because the method returns the partially populated
`Configuration` instead of raising, the user's first visible symptom is a later connection
error against the default host, with nothing tying it back to the kubeconfig.

This mirrors `_get_config_with_default_credentials` in the same class, which already handles
the analogous failure with `except Exception:` + `logging.warning`.

The in-cluster probe is treated differently on purpose. `load_incluster_config()` is used as a
detection mechanism and is *expected* to fail whenever the SDK runs outside a cluster, so
warning there would emit a false alarm on every local invocation. It logs at debug level, which
keeps the detail available under `logging.DEBUG` without adding noise. A test asserts that this
path emits no `WARNING`, so the noise cannot regress.

**Deliberately not changed:** control flow. Loading the kube config stays best effort and still
returns the configuration rather than raising. Raising would arguably be more correct, but it
is a behavior change for callers that recover downstream, so it belongs in its own issue rather
than riding along with a diagnostics fix.

**Tests**

Five cases added in `TestLoadConfigKubeConfigFallback`, covering the fallback path that
previously had no coverage at all:

- the warning contains the underlying cause
- the config object is still returned (control flow preserved)
- the out-of-cluster path emits no `WARNING` (noise guard)
- `KeyboardInterrupt` propagates from the in-cluster probe
- `KeyboardInterrupt` propagates from the kube config load

All five fail against the pre-change code and pass after it:

```
$ git stash push -- sdk/python/kfp/client/client.py   # revert the fix, keep the tests
$ pytest sdk/python/kfp/client/client_test.py -k TestLoadConfigKubeConfigFallback
5 failed
        AssertionError: no logs of level WARNING or higher triggered on root
        Captured stdout call: Failed to load kube config.

$ git stash pop
$ pytest sdk/python/kfp/client/
68 passed
```

Formatting verified with `yapf`, `isort --profile google`, and `pycln` per
[sdk/CONTRIBUTING.md](https://github.com/kubeflow/pipelines/blob/master/sdk/CONTRIBUTING.md#code-style).

**Checklist:**

Fixes #13870


sdk/RELEASE.md
```diff
@@ -8,6 +8,8 @@
 
 ## Bug fixes and other changes
 
+* fix(sdk): surface the cause when loading kube config fails (#13871)
+
```
droctothorpe:
```suggestion
```

Contributors are not responsible for updating this file.

droctothorpe:
Nice improvement. Thanks, @tolgatuncoglu. Left a minor comment. Looks like pre-commit hooks need to be run as well.

PR #13877, opened 2026-07-26
fix(cache): avoid admission-webhook panic on malformed ARGO_TEMPLATE

Atishyy27:
## Description

`intersectStructureWithSkeleton` blindly asserted `value.(map[string]interface{})` when recursing into a non-nil skeleton key (e.g. `container`). That template comes from the user-controlled `ARGO_TEMPLATE` pod env via the `/mutate` cache webhook, so a pod whose `container` is not a JSON object (string / number / array / bool / null) panics the admission handler. With `failurePolicy: Ignore` the pod is still admitted but unmutated, so caching silently breaks for that step, plus stack-trace log spam on retries.

This type-checks before recursing and skips malformed keys instead of panicking.

## Testing

Added `TestGenerateCacheKeyFromTemplateMalformedContainer`, a table-driven test over the malformed `container` shapes (string / number / array / null / bool). Each previously panicked and now returns a valid cache key. `go build` and `go vet` are clean on `backend/src/cache/server/`.

## Documentation

None; this is an internal defensive fix with no user-facing surface change.

### DIFF
--- a/backend/src/cache/server/mutation.go
+++ b/backend/src/cache/server/mutation.go
@@ -242,7 +242,15 @@ func intersectStructureWithSkeleton(src map[string]interface{}, skeleton map[str
 			if skeletonValue == nil {
 				result[key] = value
 			} else {
-				result[key] = intersectStructureWithSkeleton(value.(map[string]interface{}), skeletonValue.(map[string]interface{}))
+				valueMap, valueIsMap := value.(map[string]interface{})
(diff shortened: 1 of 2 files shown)


droctothorpe:
@Atishyy27: Hey, thank you for your PR! As you are not a member of Kubeflow organization, your PR must link to an issue with 'ready' label. Please read our [CONTRIBUTING.md](https://github.com/kubeflow/pipelines/blob/master/CONTRIBUTING.md), closing this PR

PR #13878, opened 2026-07-27
fix(backend): prevent terminated run state regressions

jeffspahr:
## Summary

- prevent stale workflow reports from moving runs backward from `CANCELING` or `CANCELED`
- make workflow-driven run updates compare the persisted state atomically and retry after concurrent changes
- preserve workflow metadata updates while retaining the authoritative termination state

## Why

PR #13876 exposed a race in the legacy v2 Run API integration test: `TerminateRun` persisted `CANCELING`, but an older workflow observation could overwrite it with `RUNTIME_STATE_UNSPECIFIED`. The zero-valued protobuf enum was then omitted from JSON, so the client observed a nil `State`.

The storage compare-and-swap closes the read/update race; an in-memory transition check alone would still allow termination to commit between those operations.

## Testing

- `go test ./backend/src/apiserver/resource -count=1`
- `go test ./backend/src/apiserver/storage -count=1`
- deterministic regression coverage injects termination between the workflow report read and its first conditional update

Related to #13876.

### DIFF
--- a/backend/src/apiserver/resource/resource_manager.go
+++ b/backend/src/apiserver/resource/resource_manager.go
@@ -1547,6 +1547,25 @@ func (r *ResourceManager) CreateOrUpdateTasks(t []*model.Task, runID string) ([]
 	return tasks, nil
 }
 
+func shouldPreserveTerminatedRunState(currentState, reportedState model.RuntimeState) bool {
+	currentState = currentState.ToV2()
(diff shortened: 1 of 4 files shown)


jeffspahr:
Superseded by #13924. The workflow-specific conditional update there provides the stronger persistence boundary, and this PR's unique deterministic read/update race test has been adapted into #13924 so its coverage is retained.

PR #13886, opened 2026-07-27
fix(sdk): apply user namespace fallback to pipeline client methods

### LINKED ISSUE #13397: [feature] Add namespace fallback to `upload_pipeline`, `list_pipelines`, and `list_recurring_runs` in KFP Python SDK

Harsh63870:
### Summary

This PR applies the existing user namespace fallback pattern to `list_pipelines`.

Previously, when `namespace` was omitted, `list_pipelines` forwarded `None` directly to the API instead of resolving the configured user namespace through `get_user_namespace()`. This behavior was inconsistent with other SDK methods such as `create_experiment`, `list_experiments`, `get_experiment`, and `list_runs`.

This change aligns `list_pipelines` with the existing SDK behavior by resolving the namespace before making the API call:

```python
namespace = namespace or self.get_user_namespace()
```

This improves consistency for multi-user deployments where a client is configured with a default namespace.

Related issue: #13397

### Notes

* Behavior is unchanged when an explicit `namespace` is provided.
* The change is limited to the Python SDK client and does not affect backend APIs.
* I investigated the remaining methods mentioned in the issue (`list_recurring_runs` and `upload_pipeline`) as well. Since a similar implementation is already being discussed in #13400, I kept this PR focused on the change currently included here.

## Test plan

Due to a local dependency compatibility issue (`docformatter` with Python 3.14), I was unable to execute the SDK test suite in my current environment.

The change is limited to a single namespace fallback line and follows the existing implementation pattern already used by other SDK methods.

Fixes #13397


Harsh63870:
@VaniHaripriya @nsingla, whenever you get a chance, could one of you take a look at this? It's a small, scoped fix: `list_pipelines` didn't fall back to `self.get_user_namespace()` when `namespace` was omitted, unlike `create_experiment`, `list_experiments`, `get_experiment`, and `list_runs`, which already do. This aligns it with that existing pattern.
I kept this PR focused on just `list_pipelines` since #13400 already covers `list_recurring_runs` and `upload_pipeline`, happy to extend this PR or open a follow-up for those once there's a decision on which approach to take there.
Also flagging: I'll push a follow-up commit with unit tests for this change shortly. Would also appreciate a maintainer `/ok-to-test` when convenient so CI can run. Thanks!

Harsh63870:
could someone from the codeowners team review this pr?

droctothorpe:
Hi @Harsh63870, thank you for your PR! The linked issue (#13397) must be labeled `ready` before your contribution can be reviewed. Please check our [CONTRIBUTING.md](https://github.com/kubeflow/pipelines/blob/master/CONTRIBUTING.md) for details.

PR #13889, opened 2026-07-27
fix(backend): wait for all DAG tasks before marking a DAG FAILED

### LINKED ISSUE #13888: [backend] UpdateDAGExecutionsState marks a DAG FAILED before sibling tasks finish running
### Environment

Pure Go logic bug in `backend/src/v2/metadata/client.go`, `UpdateDAGExecutionsState` (~line 850). Not deployment-specific — reproducible on master regardless of how KFP is deployed.

### Steps to reproduce

1. Run a pipeline with two or more independent tasks under the same DAG (e.g. a `ParallelFor`, or two unrelated tasks with no dependency between them).
2. Have one task fail while at least one sibling task is still `RUNNING`.
3. `UpdateDAGExecutionsState` runs after the first task reports its terminal state (called from the launcher's deferred cleanup on every task completion, not just once at the end).

### Current behavior

```go
var newState pb.Execution_State
switch {
case completedTasks == int(expectedTaskCount):
    newState = pb.Execution_COMPLETE
case failedTasks > 0:
    newState = pb.Execution_FAILED
default:
    return nil
}
```

The `COMPLETE` branch correctly waits until *every* expected task reaches a terminal state. The `FAILED` branch does not — it fires the instant `failedTasks > 0`, regardless of how many sibling tasks are still `RUNNING` (a still-running task doesn't increment either counter, so it's silently excluded from the check). The DAG's execution record in MLMD gets marked `FAILED` while sibling tasks are still legitimately executing.

This is inconsistent with Argo's own documented fail-fast semantics: even Argo's built-in fail-fast DAG behavior explicitly "waits until all DAG nodes are completed before failing the DAG itself" (see the `FailFast` field doc in `argoproj/argo-workflows`). This code is more premature than the system it's tracking.

### Expected result

The DAG should only transition to `FAILED` once all `expectedTaskCount` tasks have reached a terminal state (matching the same "wait for all" discipline the `COMPLETE` branch already uses), i.e. `failedTasks > 0 && completedTasks+failedTasks == expectedTaskCount`.

### Materials and Reference

- `backend/src/v2/metadata/client_update_dag_state_test.go` has table-driven coverage for `[COMPLETE, FAILED]` (both terminal) and `[COMPLETE, RUNNING]` (no failures), but no case for `[FAILED, RUNNING]` — exactly the untested gap this bug lives in.
- Related to the `FailFast` discussion in #7330 — even once independent branches are allowed to keep running (Argo-side), this MLMD-side status would still misreport the DAG as failed prematurely unless also fixed.

---

Sharkyii:

UpdateDAGExecutionsState's FAILED branch fired as soon as any task failed, without checking if sibling tasks were still running. The COMPLETE branch correctly waits for every expected task to finish (completedTasks == expectedTaskCount), but FAILED had no equivalent check, so a DAG with 5 expected tasks and 1 failure got marked FAILED in MLMD immediately, even with 4 siblings still actively running. This is more premature than Argo's own documented fail-fast behavior, which explicitly waits for all DAG nodes to reach a terminal state before failing the workflow.

Fixed FAILED to also require completedTasks+failedTasks == expectedTaskCount, mirroring the same "wait for all" discipline the COMPLETE branch already uses.

Added a test case ("one failed while sibling still running") to the existing table-driven test, confirmed it fails against the old code before the fix and passes after.

Fixes #13888.

**Checklist:**

### DIFF
--- a/backend/src/v2/metadata/client.go
+++ b/backend/src/v2/metadata/client.go
@@ -850,7 +850,7 @@ func (c *Client) UpdateDAGExecutionsState(ctx context.Context, dag *DAG, pipelin
 	switch {
 	case completedTasks == int(expectedTaskCount):
 		newState = pb.Execution_COMPLETE
-	case failedTasks > 0:
+	case failedTasks > 0 && completedTasks+failedTasks == int(expectedTaskCount):
 		newState = pb.Execution_FAILED
 	default:
 		glog.V(4).Infof("DAG is still running")
--- a/backend/src/v2/metadata/client_update_dag_state_test.go
+++ b/backend/src/v2/metadata/client_update_dag_state_test.go
@@ -178,6 +178,12 @@ func TestUpdateDAGExecutionsState_StateResolution(t *testing.T) {
 			expectedState: pb.Execution_RUNNING,
 		},
 		{
+			name:          "one failed while sibling still running",
+			taskStates:    []pb.Execution_State{pb.Execution_FAILED, pb.Execution_RUNNING},
+			expectedTasks: 2,
+			expectedState: pb.Execution_RUNNING,
+		},
+		{
 			name:          "cached and canceled count as complete",
 			taskStates:    []pb.Execution_State{pb.Execution_COMPLETE, pb.Execution_CACHED, pb.Execution_CANCELED},
 			expectedTasks: 3,


Sharkyii:
@hbelmiro could you plz review this PR?

jeffspahr:
Thanks for investigating #13888. I recommend closing this MLMD-specific PR and implementing the issue against the native task APIs on `master`. #13986 removed the client code changed here, but the early-parent-failure behavior remains in `backend/src/v2/apiclient/kfpapi/api.go`: `anyFailed` is handled before `anyRunning`, with explicit coverage in `TestUpdateStatuses_PropagatesFailedDespiteRunningSibling`.

Please do not port the strict `completedTasks + failedTasks == expectedTaskCount` condition unchanged. A focused baseline/patch comparison exposed a regression: if task A fails before dependent task B starts, B may never create an execution record. With one FAILED execution and two expected tasks, this patch leaves the DAG RUNNING indefinitely; the baseline marks it FAILED.

For the master implementation, distinguish siblings that are still actively running from tasks that will never start because of failed dependencies/fail-fast behavior. Update #13888 to describe the native status semantics and cover both cases: defer final failure while an active sibling runs, and still reach a terminal failure when downstream tasks never start. Please also preserve nested-DAG propagation and successful retry recovery.

The issue remains relevant, but it needs a complete native-state solution rather than a 2.18 backport of this patch.

PR #13898, opened 2026-07-28
fix(backend): Propagate retry strategy for tasks in dsl.ParallelFor

### LINKED ISSUE #13897: [backend] set_retry is ignored for tasks in dsl.ParallelFor at compilation time
# Issue Description

**Title:** `[backend] set_retry is ignored for tasks in dsl.ParallelFor at compilation time`

**Description:**
### Environment
* KFP version: master
* KFP SDK version: 2.12.1+

### Steps to reproduce
When used as a single task, `.set_retry(N)` works as expected. When the same component is used inside `dsl.ParallelFor`, retries do not trigger per item: a failing item fails once and the pipeline marks that branch failed without retrying.

```python
from kfp import dsl, compiler
from typing import List

@dsl.component
def random_failure_op(exit_codes: str, worker_id: str):
    import random
    import sys
    import time
    code = int(random.choice(exit_codes.split(",")))
    sys.exit(code)

@dsl.pipeline(name="parallel-retry-demo")
def parallel_retry_pipeline(
    workers: List[str] = ["A", "B", "C"],
    exit_codes: str = "0,1,2,3",
):
    with dsl.ParallelFor(workers) as wid:
        t = random_failure_op(exit_codes=exit_codes, worker_id=wid)
        t.set_retry(20)

if __name__ == "__main__":
    compiler.Compiler().compile(parallel_retry_pipeline, 'parallel_retry_pipeline.yaml')
```

### Expected result
Each `dsl.ParallelFor` item should respect `.set_retry(...)` independently. On failure, it should retry up to the configured number of times per item (similar to behavior outside `ParallelFor`).

### Root Cause
The KFP Go backend (`backend/src/v2/compiler/argocompiler/dag.go`) fails to propagate the `RetryPolicy` from the iteration `PipelineTaskSpec` into the Argo `iterationsTmpl` template. 

---

Ayush-kathil:
**Description:**
### What does this PR do?
Fixes an issue where `.set_retry()` configuration on tasks inside a `dsl.ParallelFor` context is ignored by the backend compiler.

### Why is this needed?
Currently, if an iteration inside a parallel loop fails due to a transient error, the entire loop branch fails without any retries, regardless of whether `.set_retry()` was called on the inner task. This makes data-parallel ML jobs extremely brittle.

The Python SDK correctly attaches the `RetryPolicy` to the inner task's `PipelineTaskSpec`, but the Go backend Argo compiler (`dag.go`) was leaving the iteration template's `RetryStrategy` nil at compile time.

### How does this PR fix it?
In `backend/src/v2/compiler/argocompiler/dag.go` inside `iterationItemTask`:
1. We now check `task.GetRetryPolicy()`.
2. If present, we append the necessary retry parameter defaults to the `iterationsTmpl` inputs.
3. We set the `RetryStrategy` on the `iterationsTmpl` using the extracted parameters.
4. We propagate the retry parameter values through the `Arguments` of the DAGTask calling the template (`iteratorTasks`).

### Testing
- [x] Verified Go backend compilation.
- [x] Tested with `compiler_test.go` golden updates.

### Fixes Issue
Fixes #13897

### DIFF
--- a/backend/src/v2/compiler/argocompiler/dag.go
+++ b/backend/src/v2/compiler/argocompiler/dag.go
@@ -456,6 +456,16 @@ func (c *workflowCompiler) iterationItemTask(name string, task *pipelinespec.Pip
 			Tasks: iterationTasks,
 		},
 	}
+
+	taskRetrySpec := task.GetRetryPolicy()
+	if taskRetrySpec != nil {
+		iterationsTmpl.Inputs.Parameters = append(iterationsTmpl.Inputs.Parameters, c.addParameterDefault(c.getTaskRetryParameters(task), "0")...)
+		iterationsTmpl.RetryStrategy = c.getTaskRetryStrategyFromInput(inputParameter(paramRetryMaxCount),
+			inputParameter(paramRetryBackOffDuration),
+			inputParameter(paramRetryBackOffFactor),
+			inputParameter(paramRetryBackOffMaxDuration))
+	}
+
 	iterationsTmplName, err := c.addTemplate(iterationsTmpl, componentName+"-"+name)
 	if err != nil {
 		return nil, err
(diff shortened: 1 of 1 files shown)


droctothorpe:
Hi @Ayush-kathil, thank you for your PR! The linked issue (#13897) must be labeled `ready` before your contribution can be reviewed. Please check our [CONTRIBUTING.md](https://github.com/kubeflow/pipelines/blob/master/CONTRIBUTING.md) for details.

PR #13901, opened 2026-07-28
fix(backend): return all artifacts for a list-typed input, not just the last one

### LINKED ISSUE #13900: [backend] GetInputArtifactsByExecutionID drops all but one artifact for list-typed inputs
### Environment

Pure Go logic bug in \`backend/src/v2/metadata/client.go\`, \`GetInputArtifactsByExecutionID\` (~line 1150). Not deployment-specific.

### Steps to reproduce

1. Run a pipeline where a task receives a list-typed artifact input built from multiple producers (e.g. \`dsl.Collected()\` fan-in across a \`ParallelFor\`).
2. On the write side, \`PublishExecution\` stores this correctly: \`ExecutionConfig.InputArtifactIDs\` is \`map[string][]int64\`, and for a given input name, one \`INPUT\` event is written per artifact ID, all sharing the same \`eventPath(name)\`.
3. Later, anything that calls \`GetInputArtifactsByExecutionID\` for that execution (e.g. \`backend/src/v2/driver/resolve.go:134\` and \`:265\`, resolving a DAG's input artifacts) gets back only one artifact for that name instead of all of them.

### Current behavior

```go
inputs = make(map[string]*pipelinespec.ArtifactList)
for _, artifact := range artifacts {
    name, ok := nameByID[artifact.GetId()]
    ...
    inputs[name] = &pipelinespec.ArtifactList{
        Artifacts: []*pipelinespec.RuntimeArtifact{runtimeArtifact},
    }
}
```

Each iteration does a plain map assignment instead of appending, so when multiple artifacts share the same input name, every iteration overwrites the previous one — only the last artifact (in whatever order \`GetArtifacts\` returns them) survives. No error is raised; the data loss is silent.

### Expected result

All artifacts sharing an input name should be returned in that name's \`ArtifactList\`, matching what the write side (\`PublishExecution\`/\`GenerateExecutionConfig\`) and the consuming type (\`pipelinespec.ExecutorInput_Inputs.Artifacts map[string]*ArtifactList\`, itself a list) both already support.

### Materials and Reference

- No existing test coverage for this function at all.
- Contrast with the sibling function \`GetOutputArtifactsByExecutionId\`, which has an explicit \`// TODO: Support multiple artifacts someday\` comment acknowledging the same class of limitation on the output side — this function has no such comment, suggesting it's an oversight rather than a documented scope boundary.
- Confirmed via \`git log -L\` that this has been present since the original iterator/sub-DAG support was introduced (commit \`713558cf\`, 2021), not a regression.

---

Sharkyii:

GetInputArtifactsByExecutionID builds its result with a plain map assignment inside a loop keyed by artifact name, instead of appending. When multiple artifacts share the same input name, PublishExecution writes one MLMD event per artifact (InputArtifactIDs is map[string][]int64, one name can map to several IDs, this is exactly how a list-typed/dsl.Collected() input gets recorded), but reading it back only kept the last artifact per name and silently dropped the rest. Both the write side and the consuming type (ExecutorInput_Inputs.Artifacts, itself a list) already support multiple artifacts per name, so this was a read-back bug, not a design limitation.

Fixed by appending to the existing entry instead of overwriting it.

Added a test with two artifacts sharing one input name, confirmed it fails against the old code (only 1 artifact came back instead of 2) before the fix, and passes after.

Fixes #13900.

Follow-up commit: the fix above returns all artifacts for a shared input name, but relied on whatever order GetArtifactsByID happened to return them in, not the order the INPUT events were actually recorded. For a dsl.Collected() fan-in, that meant the returned list could come back in a different order on different runs even though the pipeline never changed. Now records each artifact ID against its input name in event order first, then rebuilds the list in that order regardless of what order GetArtifactsByID returns. Added a test where the mocked response is deliberately out of event order to confirm the fix does not depend on it lining up.

```mermaid
flowchart TD
    subgraph events ["INPUT events recorded, in order"]
        E1["event: artifact 101"] --> E2["event: artifact 102"]
    end
    subgraph backend ["GetArtifactsByID response, out of order"]
        F1["returns: artifact 102"] --> F2["returns: artifact 101"]
    end
    subgraph before ["Before, list built from backend response order"]
        B1["model-b, model-a"] --> BR["Wrong order,<br/>changes run to run"]
    end
    subgraph after ["After, list built from event order"]
        A1["model-a, model-b"] --> AR["Correct order,<br/>same every run"]
    end
    style BR fill:#fce8e6,stroke:#c5221f
    style AR fill:#e6f4ea,stroke:#188038
```

**Checklist:**

### DIFF
--- a/backend/src/v2/metadata/client.go
+++ b/backend/src/v2/metadata/client.go
@@ -1159,38 +1159,44 @@ func (c *Client) GetInputArtifactsByExecutionID(ctx context.Context, executionID
 		return nil, err
 	}
 	var artifactIDs []int64
-	nameByID := make(map[int64]string)
+	orderedIDsByName := make(map[string][]int64)
 	for _, event := range eventsRes.Events {
 		if *event.Type == pb.Event_INPUT {
-			artifactIDs = append(artifactIDs, event.GetArtifactId())
+			id := event.GetArtifactId()
+			artifactIDs = append(artifactIDs, id)
 			name, err := getArtifactName(event.Path)
 			if err != nil {
 				return nil, err
 			}
-			nameByID[event.GetArtifactId()] = name
+			orderedIDsByName[name] = append(orderedIDsByName[name], id)
 		}
 	}
 	artifacts, err := c.GetArtifacts(ctx, artifactIDs)
 	if err != nil {
 		return nil, err
 	}
-	inputs = make(map[string]*pipelinespec.ArtifactList)
+	runtimeArtifactByID := make(map[int64]*pipelinespec.RuntimeArtifact, len(artifacts))
 	for _, artifact := range artifacts {
-		name, ok := nameByID[artifact.GetId()]
-		if !ok {
-			return nil, fmt.Errorf("failed to get name of artifact with id %v", artifact.GetId())
-		}
 		runtimeArtifact, err := toRuntimeArtifact(artifact)
 		if err != nil {
 			return nil, err
 		}
-		if existing, ok := inputs[name]; ok {
-			existing.Artifacts = append(existing.Artifacts, runtimeArtifact)
-		} else {
-			inputs[name] = &pipelinespec.ArtifactList{
-				Artifacts: []*pipelinespec.RuntimeArtifact{runtimeArtifact},
+		runtimeArtifactByID[artifact.GetId()] = runtimeArtifact
+	}
+	// Rebuild each list in the order its INPUT events were recorded, rather than
+	// whatever order GetArtifacts returns them in, so a dsl.Collected() fan-in
+	// comes back in the same order every run.
+	inputs = make(map[string]*pipelinespec.ArtifactList)
+	for name, ids := range orderedIDsByName {
+		artifactList := &pipelinespec.ArtifactList{}
+		for _, id := range ids {
+			runtimeArtifact, ok := runtimeArtifactByID[id]
+			if !ok {
+				return nil, fmt.Errorf("failed to get artifact with id %v for input %q", id, name)
 			}
+			artifactList.Artifacts = append(artifactList.Artifacts, runtimeArtifact)
 		}
+		inputs[name] = artifactList
 	}
 	return inputs, nil
 }
--- a/backend/src/v2/metadata/client.go
+++ b/backend/src/v2/metadata/client.go
@@ -1184,8 +1184,12 @@ func (c *Client) GetInputArtifactsByExecutionID(ctx context.Context, executionID
 		if err != nil {
 			return nil, err
 		}
-		inputs[name] = &pipelinespec.ArtifactList{
-			Artifacts: []*pipelinespec.RuntimeArtifact{runtimeArtifact},
+		if existing, ok := inputs[name]; ok {
+			existing.Artifacts = append(existing.Artifacts, runtimeArtifact)
+		} else {
+			inputs[name] = &pipelinespec.ArtifactList{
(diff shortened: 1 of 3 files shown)


Sharkyii:
@hbelmiro PTAL

dobbydobap:
Fix reads correct to me. The write side already stores one event per artifact under the same name, and `ArtifactList` is a list, so keeping only the last one was purely a read-back bug. The test fails on master for the right reason too, two events sharing `eventPath("models")` coming back as one artifact.

One thing the issue raises but leaves open. You note there that only the last artifact survives "in whatever order `GetArtifacts` returns them", and the expected result asks for all of them without saying anything about order. After this fix that order becomes user-visible: `GetArtifacts` at `client.go:1102` returns `res.Artifacts` straight from `GetArtifactsByID` with no sorting, so a `dsl.Collected()` input comes back in whatever order MLMD happened to give. Worth deciding whether that needs pinning, by sorting on artifact ID or rebuilding in `artifactIDs` order, so two runs of the same pipeline agree. If order is not meant to be meaningful for a collected input, saying so somewhere would save the next reader the same question.

Sharkyii:
`GetInputArtifactsByExecutionID` now rebuilds each list based on the order the `INPUT` events were recorded, instead of relying on the order returned by `GetArtifactsByID`. This makes sure `dsl.Collected()` fan-in results stay in the same order across runs.

I went with the event order rather than sorting by artifact ID because artifact IDs are assigned based on the order parallel upstream tasks finish. That means sorting by ID wouldn't reliably match the actual logical fan-in order.

dobbydobap:
That reads right, and your reasoning for event order over artifact ID is better than what I suggested. Sorting by ID would track the completion order of the parallel tasks rather than the fan-in order, which is not what anyone wants.

One caveat on the new comment though. MLMD does not actually promise event order. `GetEventsByExecutionIDsResponse` is just `repeated Event events`, and the only documentation on the RPC is "Gets all events with matching execution ids", with no ordering clause. In practice you will almost certainly get insertion order back, and this is still a real improvement over relying on `GetArtifactsByID`, but "the same order every run" is a stronger claim than the API makes. Might be worth softening it to say the order follows the recorded events.

There is a native way to make it contractual, for later rather than here. `Event.Path.Step` is a `oneof` of `int64 index` and `string key`, so a path can carry a position as well as a name. KFP only ever writes a single `key` step in `eventPath`, reads it back with `Steps[0].GetKey()`, and nothing in `backend/src` references the index variant at all. Writing the fan-in position as a second step would pin order by construction rather than by row order. That is a write-side change though, so clearly not this PR.

Not a blocker either way, the change is an improvement as it stands.

Sharkyii:
it now says the fix ties list order to the recorded events rather than claiming a same-order guarantee, since GetEventsByExecutionIDs doesn't document any ordering either.

jeffspahr:
Thanks for this fix. Please retarget/rebase this PR to `release-2.18`. #13986 removed the MLMD client from master, but 2.18 still uses it, and this patch addresses real silent truncation of list-valued artifact inputs passed across DAG/subpipeline or loop boundaries.

I checked the patch against release-2.18: it applies cleanly, the existing implementation fails the two-artifact regression, and the patched implementation passes that test, five additional edge cases, and the existing mocked DAG tests. This is a useful, narrowly scoped read-path fix for 2.18. Please run the relevant release-branch checks after retargeting.

The ordering caveat in the updated code comment is appropriate: the result follows the returned MLMD event sequence, without claiming a guaranteed logical fan-in order. KFP 2.16 will not receive any further releases; `release-2.18` is the target for this work.

Issue #13712, opened 2026-07-13
[bug] generate_proto.py checks wrong output path before regenerating Python protobuf


HarshPopat23:
### Environment

*  How do you deploy Kubeflow Pipelines (KFP)?
Not deployment-specific. This issue is in the repository build/generation script.

*  KFP version:
Current `master` branch / source repository.

*  KFP SDK version:
Not SDK-runtime-specific. This affects protobuf generation for `api/v2alpha1/python`.

### Steps to reproduce

In `api/v2alpha1/python/generate_proto.py`, the script calculates the expected generated output file as:

```python
output = source.replace(".proto", "_pb2.py")
```
For this source file: `api/v2alpha1/pipeline_spec.proto`
that produces this expected output path: `api/v2alpha1/pipeline_spec_pb2.py`
However, the script actually invokes protoc with:
```python
protoc_command = [
    PROTOC,
    "-I%s" % PROTO_DIR,
    "--python_out=%s" % PKG_DIR,
    source
]
```
PKG_DIR points to: `api/v2alpha1/python/kfp/pipeline_spec`
So the real generated file should be:`api/v2alpha1/python/kfp/pipeline_spec/pipeline_spec_pb2.py`

_**Because of this mismatch, the script checks the modification time of the wrong output path.**_

**Expected result** 

**generate_proto.py** should check the actual generated Python protobuf file before deciding whether regeneration is needed.

The output path should be calculated relative to PKG_DIR, for example:

```python
output = os.path.join(
    PKG_DIR,
    os.path.basename(source).replace(".proto", "_pb2.py"),
)
```
or with pathlib:
```python
output = PKG_DIR / f"{source.stem}_pb2.py"
```

This would make the documented behavior correct:

```python
Does nothing if the output already exists and is newer than the input.
```
**Actual result**

The script currently checks: `api/v2alpha1/pipeline_spec_pb2.py`
but protoc writes to:`api/v2alpha1/python/kfp/pipeline_spec/pipeline_spec_pb2.py`

As a result, the timestamp optimization **may not work correctly**. In the common case, the script may regenerate the protobuf file every time because the checked output file does not exist.

It could also incorrectly skip generation if a stale or unrelated file exists at the incorrectly calculated path.

**Materials and reference**

Relevant file: `api/v2alpha1/python/generate_proto.py`

Incorrect current code:
```python
output = source.replace(".proto", "_pb2.py")
```

Suggested improvement:

```python
output = os.path.join(
[...]

PR #13909, opened 2026-07-28
fix(kubernetes_platform): check generated protobuf under PKG_DIR

Harsh63870:
## Summary
- Fix `kubernetes_platform/python/generate_proto.py` so the up-to-date check uses the real `protoc` output under `PKG_DIR`.
- Same class of bug as #13712 (which targets `api/v2alpha1/python/generate_proto.py`); this PR covers the kubernetes_platform copy only to avoid conflicting with #13719.

## Problem
The script used `source.replace(".proto", "_pb2.py")`, which looks next to the `.proto` under `proto/`. `protoc` is invoked with `--python_out=PKG_DIR`, so the real file is `python/kfp/kubernetes/kubernetes_executor_config_pb2.py`. The wrong path made the mtime optimization unreliable.

## Test plan
- [ ] Inspect `generate_proto.py` path logic
- [ ] Optional: with `protoc` available, run `python kubernetes_platform/python/generate_proto.py` twice and confirm the second run skips generation when output is newer

Related: #13712

### DIFF
--- a/kubernetes_platform/python/generate_proto.py
+++ b/kubernetes_platform/python/generate_proto.py
@@ -66,7 +66,11 @@ def generate_proto(source: str) -> None:
     Args:
       source: The source proto file that needs to be compiled.
     """
-    output = source.replace('.proto', '_pb2.py')
+    # protoc writes into PKG_DIR (--python_out), not next to the .proto source.
+    output = os.path.join(
+        PKG_DIR,
+        os.path.basename(source).replace('.proto', '_pb2.py'),
+    )
 
     if not os.path.exists(output) or (
             os.path.exists(source) and


Harsh63870:
this looks like it's waiting on /ok-to-test before ci can run, could @droctothorpe or @zazulam take a quick look and trigger that when they get a chance? small, scoped fix of 5 lines, so should be a quick pass.

droctothorpe:
@Harsh63870: Hey, thank you for your PR! As you are not a member of Kubeflow organization, your PR must link to an issue with 'ready' label. Please read our [CONTRIBUTING.md](https://github.com/kubeflow/pipelines/blob/master/CONTRIBUTING.md), closing this PR

PR #13913, opened 2026-07-28
fix(backend): propagate ctx instead of using context.TODO/Background in PVC operations

shivv23:
## Description

`createPVC`, `deletePVC`, and `publishDriverExecution` in the v2 driver use `context.Background()` / `context.TODO()` for Kubernetes API calls instead of propagating the `ctx` parameter already available in their function signatures. This means PVC create/delete operations and driver pod metadata lookups do not respect the caller's context cancellation or deadline.

### Root cause

Three Kubernetes API calls in `backend/src/v2/driver/k8s.go` ignore the passed context:

| Function | Line | Before (bug) | After (fix) |
|----------|------|---------------|-------------|
| `createPVC` | 1077 | `context.Background()` | `ctx` |
| `deletePVC` | 1197 | `context.TODO()` | `ctx` |
| `deletePVC` | 1203 | `context.TODO()` | `ctx` |
| `publishDriverExecution` | 1308 | `context.TODO()` | `ctx` |

### Fix

Replace each `context.Background()` and `context.TODO()` call with the existing `ctx` parameter that is already passed to the enclosing function.

### Changes

- `backend/src/v2/driver/k8s.go` — 4 lines changed

## Checklist

- [x] I have signed off my commits with `git commit -s`
- [x] All existing tests pass locally

### DIFF
--- a/backend/src/v2/driver/k8s.go
+++ b/backend/src/v2/driver/k8s.go
@@ -1074,7 +1074,7 @@ func createPVC(
 	}
 
 	// Create the PVC in the cluster
-	createdPVC, err := k8sClient.CoreV1().PersistentVolumeClaims(opts.Namespace).Create(context.Background(), pvc, metav1.CreateOptions{})
+	createdPVC, err := k8sClient.CoreV1().PersistentVolumeClaims(opts.Namespace).Create(ctx, pvc, metav1.CreateOptions{})
 	if err != nil {
 		return "", createdExecution, pb.Execution_FAILED, fmt.Errorf("failed to create pvc: %w", err)
 	}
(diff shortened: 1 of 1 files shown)


shivv23:
@hbelmiro Hi! Any changes to be done, if needed, then please tell.

hbelmiro:
> @hbelmiro Hi! Any changes to be done, if needed, then please tell.

@shivv23 The CI is failing. Can you PTAL?

shivv23:
@hbelmiro done. please have a look.

hbelmiro (approved):
/lgtm
/approve

PR #13924, opened 2026-07-29
fix(backend): fence workflow report, retry, and termination races

jeffspahr:
## Problem

A run's lifecycle is split between the KFP database and an Argo Workflow while the persistence agent reports Workflow snapshots asynchronously. `ReportWorkflowResource`, `TerminateRun`, and `RetryRun` can each read one snapshot and then mutate either system after another operation has advanced the run. Retries also reuse the Workflow name, so state-only checks do not identify the Workflow incarnation being changed.

The resulting races could:

- let a stale report overwrite `CANCELING`, a terminal state, or a newer retry's manifest;
- let a delayed retry recreate work after cancellation or after another retry took over;
- let an old terminal report label or delete the Workflow now used by a retry;
- roll back a retry claim after cancellation or a Workflow report had already won; or
- delete a replacement pod that reused a failed pod's name.

Legacy rows and Workflows add two compatibility cases: older run rows may keep state only in `Conditions`, and existing Workflows may have a missing or caller-supplied retry-generation annotation.

### Use one retry-generation fence end to end

- Reserve `pipelines.kubeflow.org/retry-generation` as an internal annotation and stamp generation `0` on initial one-time and recurring Workflows.
- Keep the database `RetryGeneration` as the monotonic claim token and stamp each retried Workflow with the claimed generation.
- Apply workflow-derived run updates only when both the observed effective state and retry generation still match in the same SQL `UPDATE`.
- Prevent active reports from regressing state or reopening terminal runs. A covered terminal report completes durable cancellation as `CANCELED`.
- Match legacy `State` and `Conditions` representations using the same normalization as `Run.ToV2`; successful writes migrate them to canonical state.
- Return a retryable `Unavailable` response for stale or ambiguous work before tasks, metrics, plugins, manifests, state history, labels, or deletion can be applied from the rejected report.

### Make retry reconciliation crash- and replay-safe

- Bind each database claim to the exact generation read by the caller, and roll it back only while that same generation is still `PENDING`.
- Reconcile expired claims against Kubernetes before takeover: adopt a live Workflow carrying the claimed generation, preserve the claim on ambiguous reads, and take over only when the prior mutation is definitively absent or older.
- Revalidate immediately before Workflow create or update, advance only an older generation, adopt an already-applied same-generation update, and never overwrite a newer generation.
- Re-read after ambiguous Kubernetes errors and preserve or adopt applied work rather than restoring a terminal, GC-eligible database row.
- Guard final retry persistence with the same state/generation compare-and-set and compensate when cancellation, rollback, or a newer retry won concurrently.
- Snapshot failed pod UIDs before claiming and delete with UID preconditions so a stale retry cannot remove replacement pods.

### Preserve cancellation and terminal cleanup

- Make termination idempotent, persist `CANCELING`, then re-read the generation that was actually canceled before patching Kubernetes.
- Patch only a matching Workflow resource version whose generation is no newer than the durable cancellation; reporting and retry recovery reapply that cancellation when delayed work appears.
- Require terminal reports to carry a resource version, verify the live Workflow before persisting terminal state, and add the persisted-final-state label with a resource-version-tested JSON patch.
- Delete a persisted-final-state Workflow with a resource-version precondition so a stale terminal snapshot cannot delete a retry that reused the same name.
- Normalize legacy retry annotations only after revalidating the database snapshot and Workflow resource version.

## Compatibility

This uses the existing retry-generation fields and annotation; it adds no schema migration or public API shape change. Initial user or plugin values for the reserved annotation are normalized to `0`, legacy database state representations remain accepted, and detected stale or ambiguous interleavings return a retryable `Unavailable` response.

This PR consolidates and supersedes #13878.

## Verification

- Deterministic regression coverage for report/terminate, report/retry, retry/retry, delayed retry writes, ambiguous Kubernetes responses, expired-claim adoption or takeover, rollback races, and stale terminal label/delete paths.
- Compatibility coverage for legacy run-state representations and legacy retry annotations.
- Pod UID and Workflow resource-version precondition tests.
- `go test ./backend/src/apiserver/storage`
- `go test ./backend/src/apiserver/resource`
- `go vet ./backend/src/apiserver/storage`
- At head `fa46e780`, DCO, pre-commit, backend tests, builds, and the repository integration/E2E matrix are green.


backend/src/apiserver/storage/run_store.go
```diff
@@ -640,40 +691,58 @@ func (s *RunStore) UpdateRun(run *model.Run) error {
...
+		}
+		currentRun, getErr := s.GetRun(run.UUID)
+		if getErr != nil {
+			return false, getErr
+		}
+		// MySQL may report zero affected rows when the matching row already has
+		// the requested values. Only treat it as applied when the compare state
+		// still matches; reaching the incoming state through another update must
+		// retry so this report cannot skip its manifest and history write.
+		if currentRun.State == expectedState.ToV2() {
```
zazulam: **Blocking:** Account for legacy rows whose `State` is `NULL` before treating zero affected rows as success.

- `RunDetails.State` is nullable, and `GetRun(...).ToV2()` reconstructs it from `Conditions` when the column is absent.
- For such a row, the new `WHERE State = ?` predicate matches nothing. This fallback then rereads the normalized state, sees it equal `expectedState`, and returns `true` even though `State`, `StateHistory`, and `WorkflowRuntimeManifest` were not persisted.
- `ReportWorkflowResource` consequently proceeds to tasks and metrics without retrying. A terminal report can leave an upgraded run permanently `RUNNING`.

I reproduced this by setting fixture run 1 to `State = NULL` and reporting `SUCCEEDED`: the method returned `true`, while a subsequent `GetRun` still returned `RUNNING` with the old manifest. Please include the legacy `Conditions` representation in the CAS (or backfill `State` before enabling it) and add a regression test.

zazulam (changes requested):
## PR Overview

The workflow-specific compare-and-set is well scoped and the concurrency test covers the intended live-row termination race. However, the new state comparison does not handle supported legacy rows whose `State` column is `NULL`, so upgraded runs can silently skip workflow updates.



**Requesting changes** for the legacy-state correctness issue in the inline comment. I reproduced it on this PR head by setting a fixture run`s `State` to `NULL`: `UpdateRunFromWorkflow` returned `true`, but the persisted state remained `RUNNING` and the workflow manifest remained unchanged.

backend/src/apiserver/storage/run_store.go
jeffspahr: Confirmed and fixed in 634e891. Your reproduction was exact — I reproduced it before changing anything:

```
after NULLing:  State="RUNNING" Conditions="Running" manifest="workflow1"
UpdateRunFromWorkflow -> updated=true err=<nil>
persisted:      State="RUNNING" Conditions="Running" manifest="workflow1"
```

The reconstruction is `model.Run.ToV2` (`model/run.go:301`), which sets `State` from `Conditions` when the column is empty. That makes both the reporter's `expectedState` and the zero-row fallback's re-read normalize to `RUNNING`, while `WHERE State = 'RUNNING'` matches nothing against a NULL column — so the fallback's equality check confirmed a write that never happened.

I went with including the legacy representation in the CAS rather than backfilling, since backfilling needs a migration and this keeps the fix self-contained. The predicate is now:

```sql
(State IN (<representations>)) OR (State IS NULL AND Conditions IN (<representations>))
```

where the representations are derived by inverting `ToV2` over the known state vocabulary rather than hand-listed, so the set can't drift from the mapping. This also covers a case adjacent to the one you found: a non-NULL row storing a non-canonical spelling (`State = 'Running'`) failed the predicate identically and hit the same false-success path. A matched legacy row is migrated off the legacy representation by the update itself.

Regression tests in `run_store_test.go`:
- `TestUpdateRunFromWorkflow_PersistsLegacyStateRepresentations` — your scenario plus the non-canonical-spelling variant; asserts state, conditions, manifest, and state history are actually persisted, not just that the call returned `true`. I verified it fails against the previous predicate.
- `TestUpdateRunFromWorkflow_RejectsStaleReportForLegacyStateRun` — a legacy row must still reject a stale nonterminal report, so the widened predicate doesn't weaken the guard this PR is for.
- `TestStoredRuntimeStates` — pins the derived representation sets.

jeffspahr:
Thanks — the legacy-state issue was real and is fixed in 634e891. Details in the [inline reply](https://github.com/kubeflow/pipelines/pull/13924#discussion_r3806666238); summary:

- The CAS predicate now matches every stored representation that normalizes to the observed state, including the legacy `Conditions` value when `State` is `NULL`. The representation sets are derived by inverting `ToV2` rather than hand-listed.
- This also covers non-canonical spellings in a non-NULL column (`State = 'Running'`), which failed the old predicate the same way and reached the same false-success fallback.
- Three regression tests added, including your exact reproduction. I verified it fails against the previous predicate.
- Rebased onto master; `go test ./backend/src/apiserver/storage ./backend/src/apiserver/resource` and `go vet` pass.

**Separate pre-existing issue, not addressed here.** While testing I found `RunStore.TerminateRun` has the same NULL-`State` blind spot on master, independent of this PR:

```go
WHERE UUID = ? AND (State = ? OR State = ? OR State = ? OR State = ?)
```

A legacy row with `State = NULL` matches nothing, so terminating such a run fails with `Failed to terminate a run %s. Row not found`. My legacy-row test originally called `TerminateRun` and hit this; it now applies the competing write directly so it stays focused on the stale-report guard.

I left it out of this PR because `TerminateRun` is on the explicit lifecycle path rather than the workflow-report path this change is scoped to, and widening it deserves its own review. Happy to open a follow-up issue, or fold it in here if you would rather see both fixed together.

jeffspahr:
Folded the `TerminateRun` fix into this PR as well (46aa5c6), since it is the same root cause and splitting it would have left the repository able to report a legacy run's state but not cancel it.

The compare-and-set predicate is now generalized over several states and reused in both places. It also covers a case neither call site handled: a row where neither `State` nor `Conditions` holds a value, which reads back as `RUNTIME_STATE_UNSPECIFIED` and is therefore cancelable.

Added coverage:

- `TestTerminateRun_LegacyStateRepresentations` — `NULL` state, empty state, non-canonical spelling, and neither column set. All four fail against the previous predicate.
- `TestTerminateRun_RejectsLegacyTerminalRun` — a legacy row whose `Conditions` is `Succeeded`/`Failed`/`Skipped` must stay unterminable, so matching `Conditions` does not widen which runs can be canceled.

The legacy stale-report test now drives the real `TerminateRun` path rather than the direct SQL write it needed previously.

The PR body has been updated to describe the widened scope. `go test ./backend/src/apiserver/storage ./backend/src/apiserver/resource` and `go vet` pass; rebased onto current master.

backend/src/apiserver/storage/run_store.go
```diff
@@ -611,18 +617,127 @@ func (s *RunStore) GetRunByRecurringRunIDAndDisplayName(recurringRunID, displayN
...
+		model.RuntimeState(model.LegacyStateDisabled), model.RuntimeState(model.LegacyStateError),
+		model.RuntimeState(model.LegacyStateReady), model.RuntimeState(model.LegacyStateRunning),
+		model.RuntimeState(model.LegacyStateSucceeded), model.RuntimeState(model.LegacyStateDone),
+		model.RuntimeState(model.RunTerminatingConditionsV1),
+	}
+	seen := make(map[string]bool, len(candidates))
+	stored := make([]string, 0, len(candidates))
+	for _, candidate := range candidates {
+		value := string(candidate)
+		if value == "" || seen[value] || candidate.ToV2() != canonical {
```
jeffspahr, line 790: [P1] This candidate list still does not invert `RuntimeState.ToV2()` on case-sensitive stores. `ToV2()` uppercases its input, so a legacy `State = "running"` (or absent State with `Conditions = "running"`) reads back as `RUNNING`, but this list only includes `ENABLED`, `RUNNING`, `Ready`, and `Running`. The guarded UPDATE then affects zero rows and line 799 re-reads the normalized `RUNNING` value and returns success even though the terminal state, history, and manifest were not written; `TerminateRun` similarly returns not found. Please make the SQL predicate mirror the case-insensitive normalization and cover lowercase/mixed-case State and Conditions rows.

jeffspahr:
One additional merge blocker from the current master conflict: please preserve the `RetryGeneration = run.RetryGeneration` fence added by #13532 when composing this PR’s effective-state CAS. The resolved workflow update needs UUID + retry generation + effective state. The zero-row fallback must also require the generation to still match—not just the normalized state—because an expired retry-claim takeover can increment the generation while leaving the state `PENDING`; otherwise a stale reporter can still be declared successful without writing its manifest/history. Please add a regression for a generation bump with unchanged effective state.

backend/src/apiserver/storage/run_store.go
jeffspahr: Fixed in 97d9bacd after rebasing onto current master. The predicate now compares uppercase normalized State/Conditions values, with lowercase and mixed-case regression coverage for both workflow CAS and termination. The full storage and resource packages plus go vet pass.

jeffspahr:
Resolved the retry-generation integration blocker in 97d9bacd. The branch is rebased onto eb8f4eec; workflow updates now compose UUID + retry generation + effective-state guards, and the zero-row fallback requires the generation snapshot to remain current. Added the unchanged-state generation-takeover regression. Full storage/resource unit packages and go vet pass.

backend/src/apiserver/storage/run_store.go
```diff
@@ -145,6 +147,11 @@ type RunStoreInterface interface {
 	// Note: only state, runtime manifest can be updated. Does not update dependent tasks.
 	UpdateRun(run *model.Run) (err error)
 
+	// Updates a run from a workflow report if its state has not changed since it
+	// was read and the report does not regress a canceling or terminal run.
+	// Returns false without modifying the run when the report is stale.
+	UpdateRunFromWorkflow(run *model.Run, expectedState model.RuntimeState) (bool, error)
```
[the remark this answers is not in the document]
jeffspahr: Fixed in d755075b. Added the missing `UpdateRunFromWorkflow` stub to `fakeRunStore`; the cancelling and blocking variants embed that fake, and the interface audit found no other direct implementations. `go test ./backend/src/apiserver/gc` and `go vet` for GC/storage/resource pass.

backend/src/apiserver/storage/run_store.go
```diff
@@ -699,18 +706,128 @@ func (s *RunStore) GetRunByRecurringRunIDAndDisplayName(recurringRunID, displayN
...
+}
+
+// storedRuntimeStates lists the uppercase forms of every known stored
+// representation that normalizes to the given runtime state. Rows predating
+// the State column carry a v1 condition instead, and older writers stored
+// non-canonical spellings, so a compare-and-set that only matched the canonical
+// v2 string would silently match nothing.
+func storedRuntimeStates(state model.RuntimeState) []string {
+	canonical := state.ToV2()
+	candidates := []model.RuntimeState{
```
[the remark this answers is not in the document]
jeffspahr: Fixed in d755075b. On zero affected rows, the workflow path now reuses the exact UUID + RetryGeneration + effective-state predicate in a same-transaction `SELECT FOR UPDATE`; only an actual predicate match is treated as a MySQL no-op success. Unknown raw states therefore fail closed instead of being certified through normalized `GetRun` state. Added `TestUpdateRunFromWorkflow_RejectsUnknownStoredState`, which asserts state, conditions, manifest, and history remain unchanged.

jeffspahr:
Closing in favor of #14322, which tracks the architectural concern: run lifecycle ownership and recovery are split across the database, asynchronous Workflow reports, and Kubernetes mutations. This PR kept expanding as each local guard exposed another coordination or recovery case; the follow-up should establish a shared contract and deliver it in small, independently reviewable changes.

The focused cancellation/report fix shipped in #14240. It does not resolve the broader lifecycle reconciliation problem. #14322 preserves the design questions and regression scenarios from this PR and links the related task-lifecycle work in #14032 and runtime-resilience work in #14020.

PR #13925, opened 2026-07-29
Feature/pagetoken rolling upgrade

### LINKED ISSUE #13858: [backend] Page tokens issued before `ac3a4c656` generate invalid SQL after upgrade
# List pagination breaks across upgrade: page tokens issued before `ac3a4c656` generate invalid SQL

## Summary

Page tokens minted by an API server built before `ac3a4c656` (PR #13127) lack the `SortBySQLColumn` field that commit introduced. When such a token is presented to a current API server, that field decodes to the empty string, the sort column disappears from the generated SQL, and the database rejects the statement — surfacing as HTTP 500 on the next-page request.

The blast radius is small and it fails closed, but it is not fully self-healing: retrying the same next-page action reproduces the error, because the client is still holding the same stale token.

- **Affects:** master (verified at `465403f96`), on any deployment upgraded from a release predating `ac3a4c656`.
- **Component:** `backend/src/apiserver/list`

## What happens

`token.unmarshal()` validates the decoded token but performs no version detection or normalization. A legacy token deserializes with `SortByFieldName` populated and `SortBySQLColumn` empty. `validateIdentifierName` returns early on empty input, so the token passes validation. SQL construction then reads `SortBySQLColumn` — not `SortByFieldName` — for both the cursor `WHERE` clause and `ORDER BY`, so the column name is simply absent from the emitted SQL.

The failure reaches the user as `codes.Internal` / HTTP 500 (`Failed to list runs: ...`), raised from `RunStore.ListRuns`.

## Reproduction

Self-contained test in `backend/src/apiserver/list`:

```go
func TestLegacyPageToken(t *testing.T) {
	raw := `{"SortByFieldName":"accuracy","SortByFieldValue":0.5,"KeyFieldName":"UUID","KeyFieldValue":"abc"}`
	opts, err := NewOptionsFromToken(base64.StdEncoding.EncodeToString([]byte(raw)), 10)
	if err != nil {
		t.Fatal(err)
	}
	sql, _, _ := opts.AddPaginationToSelect(sq.Select("*").From("T")).ToSql()
	t.Log(sql)
}
```

Output on master:

```
SELECT * FROM T WHERE ( > ? OR ( = ? AND UUID >= ?)) ORDER BY UUID ASC LIMIT 11
```

Note the empty left-hand operands (`( > ?`) and the missing sort key in `ORDER BY`.

## Scope: not limited to metric sorts

This is a property of the token format, not of metric handling. A legacy token for an ordinary column sort degrades identically:

| Legacy token (no `SortBySQLColumn`) | Emitted SQL fragment |
|---|---|
| `SortByFieldName: "DisplayName"` (regular column) | `WHERE ( > ? OR ( = ? AND UUID >= ?))` |
| `SortByFieldName: "accuracy"` (identifier-shaped metric) | `WHERE ( > ? OR ( = ? AND UUID >= ?))` |
| `SortByFieldName: "Name"`, `SortByFieldPrefix: "pipelines."` | `WHERE (pipelines. > ? OR (pipelines. = ? AND pipelines.UUID >= ?))` |

Legacy tokens whose `SortByFieldName` fails the SQL identifier regex (a hyphenated metric name such as `log-loss`) are rejected earlier in `unmarshal()` with `Invalid sort field name`. That is a cleaner error, but the page-2 request still fails. So the identifier regex determines only *how* the request fails, not *whether* it does.

## Severity and recovery

Low to moderate. The failure mode is benign in the ways that matter: the database rejects the malformed statement, so there is no silently misordered or under-filtered result set, and no injection surface (an empty string cannot alter query structure). Exposure is bounded by page-token lifetime — a single pagination session.

Recovery is worth stating precisely, because the intuitive retry does **not** work. In the web UI, `CustomTable` keeps one token per page index in `this.state.tokenList`. On a failed page change, [`_pageChanged`](https://github.com/kubeflow/pipelines/blob/master/frontend/src/components/CustomTable.tsx#L606-L626) leaves `tokenList` untouched, so clicking "next page" again re-sends the same stale token and reproduces the same 500. `tokenList` is only reset in [`_resetToFirstPage`](https://github.com/kubeflow/pipelines/blob/master/frontend/src/components/CustomTable.tsx#L658-L674):

| User action after the error | Token list reset? | Outcome |
|---|---|---|
| Click "next page" again | No | Same 500, repeatedly |
| Reload the page / re-enter the view | Yes (remount) | Recovers |
| Change sort order or page size | Yes | Recovers |

SDK and direct API consumers are affected the same way: `KFP Client.list_runs(page_token=...)` takes the token from the caller, so a script looping on `next_page_token` will fail at the same point and keep failing until it restarts pagination from an empty token.

## Why a regex-based fix does not work

The tempting fix is to detect legacy metric tokens by testing whether `SortByFieldName` looks like a SQL identifier, and promote the ones that do not. That only catches names containing characters an identifier cannot hold (`log-loss`, `val/loss`). It cannot resolve the ambiguous case: for an identifier-shaped name such as `accuracy` or `f1_score`, "the metric named `accuracy`" and "a column named `accuracy`" are the same string. The information needed to distinguish them is not present in the token.

Resolving it requires a model-aware lookup — asking the `Listable` whether the name is one of its real fields, and whether `metric:<name>` resolves — rather than pattern-matching the string.

## Why existing tests do not catch this

The tests added alongside `ac3a4c656` are not missing assertions; they are missing an input class. Every token in `list_test.go` is either constructed in the current layout or produced by `marshal()` and fed back into `unmarshal()`. A round-trip test is self-consistent by construction and green forever, while covering nothing about cross-version compatibility, because both ends of the trip use the current format.

More generally: the page token is base64-encoded and looks like an opaque internal detail, but it is persisted data held by external clients across server version boundaries. Under a "serialized struct" mental model, adding a field is harmless. Under a "wire format" model, it is a compatibility-breaking change. Any regression test for this class of bug has to use a hand-written legacy JSON literal as its input — never a token the current code minted.

## Proposed fix

Classify the token explicitly in `token.unmarshal()` and normalize legacy tokens before validation:

1. `SortBySQLColumn` present → current format, no action.
2. `SortBySQLColumn` absent → legacy. Ask the listable whether `SortByFieldName` is a real field; if so, set `SortBySQLColumn = SortByFieldName`. Otherwise confirm it resolves as a metric via `GetField("metric:" + SortByFieldName)` before promoting it to the metric-sort layout, and reject the token if neither holds.

Step 2 requires `unmarshal()` to have access to the `Listable`, which it currently does not. PR #13547 already implements this approach (adding `Listable.IsRegularField()` and threading the listable through `unmarshal`), so that PR likely resolves this issue; filing separately so the gap is tracked on its own and is not lost if #13547 stalls.

A defensive guard is worth having regardless of which fix lands: SQL construction should refuse to emit a comparison against an empty column name rather than silently producing `( > ?`.

## Suggested regression tests

Each of these must be driven by a hand-written legacy JSON token, not a marshalled one:

- Legacy token for an identifier-safe metric name (`accuracy`) paginates correctly instead of emitting empty-column SQL.
- Legacy token for a real column (`DisplayName`) paginates correctly.
- Legacy token whose `SortByFieldName` is neither a real column nor a resolvable metric is rejected with a clear `InvalidInputError`.
- Legacy token against a listable with no metric-sort support (e.g. `Pipeline`) is never promoted to a metric sort.

## References

- `ac3a4c656` / PR #13127 — introduced `SortBySQLColumn` (security fix, HASI2026112-23).
- PR #13547 — implements the model-aware classification described above.
- PR #12379 — Postgres integration; the review discussion there surfaced this gap.

palakbhati:
## What this PR does

Continues #13547, resolving the two outstanding review items from nsingla's
review before it can merge, and closes #13858 (the original bug report for
legacy page tokens breaking pagination after upgrade).

Built on top of #13547's model-aware token migration, co-authored with
@kaikaila (Yunkai Li), who did the core migration work.

## What's fixed here

1. **Rolling-upgrade backward compatibility** (nsingla, critical): already
   implemented in #13547 per the expand/contract/cleanup roadmap —
   `nextPageToken()` writes `SortBySQLColumn` for one release cycle so
   version-B (2.17) servers can still read version-C tokens correctly during
   a mixed-version rollout. Covered by `DimensionB_Unmarshal_Page2Migration`
   and `DimensionC2_RoundTrip_CompatToken` in `list_test.go`.

2. **`sort_metric_value` alias collision** (nsingla, important): a metric
   literally named `sort_metric_value` (the fixed SQL alias for metric
   sorts) could make `Run.GetField`'s `n != sqlCol` check false, causing the
   token to be silently misclassified as a regular column sort on a
   nonexistent column. Fixed with a reserved-name guard in `Run.GetField`
   (`model/run.go`) that rejects `metric:sort_metric_value` outright. New
   test: `TestRun_GetField_RejectsMetricNamedAfterSortAlias`.

## Testing
go test ./backend/src/apiserver/model/... ./backend/src/apiserver/list/...
ok github.com/kubeflow/pipelines/backend/src/apiserver/model
ok github.com/kubeflow/pipelines/backend/src/apiserver/list

## References
- Closes #13858
- Continues #13547
- cc @nsingla for review, per @kaikaila's note that she won't have review bandwidth going forward


hbelmiro:
@palakbhati I haven't done a proper review yet, but I can see that the comments are overly verbose. Can you please make them more succinct or even remove the ones that are obvious by just reading the code?

palakbhati:
@hbelmiro just pushed an update trimming the verbose comments down to what's actually load-bearing  dropped the ones that just restated the code, kept the ones explaining non-obvious reasoning (with a pointer to PAGE_TOKEN_SORT_DESIGN.md for the fuller design history where relevant). Let me know if anything's still too much.

droctothorpe:
Hi @palakbhati, thank you for your PR! The linked issue (#13858) must be labeled `ready` before your contribution can be reviewed. Please check our [CONTRIBUTING.md](https://github.com/kubeflow/pipelines/blob/master/CONTRIBUTING.md) for details.

PR #13930, opened 2026-07-29
test(e2e): add Kind GPU scheduling check with Fake GPU Operator

sduvvuri1603:
## Summary
- Adds a schedule-only E2E label/fixture (`gpu-scheduling-check`) that requests `nvidia.com/gpu` and verifies the run succeeds (no CUDA/`torch.cuda`).
- Adds `.github/actions/install-fgo-fake` to install Fake GPU Operator **fake** backend on Kind and wait for allocatable `nvidia.com/gpu`.
- Wires a dedicated `e2e-test.yml` matrix lane that installs FGO after deploy, then runs the new label.

## Test plan
- [x] Local Kind + FGO fake: `ginkgo -v --label-filter=gpu-scheduling-check ./backend/test/end2end -- -namespace=kubeflow -apiUrl=http://localhost:8888` → **PASSED**
- [x] CI: confirm the new `gpu-scheduling-check` E2E job installs FGO and passes
- [x] Confirm other E2E lanes are unchanged / still green

### DIFF
--- a/.github/actions/install-fgo-fake/action.yml
+++ b/.github/actions/install-fgo-fake/action.yml
@@ -1,0 +1,66 @@
+name: "Install Fake GPU Operator (fake backend)"
+description: >
+  Install Run:ai Fake GPU Operator with the default fake backend on Kind,
+  label nodes into the simulated GPU pool, and wait until nvidia.com/gpu is allocatable.
+
+inputs:
+  chart_version:
+    description: "fake-gpu-operator Helm chart version to pin"
+    required: false
+    default: "0.0.59"
+  namespace:
+    description: "Namespace for the Fake GPU Operator release"
+    required: false
+    default: "gpu-operator"
+  node_pool:
+    description: "Value for run.ai/simulated-gpu-node-pool node label"
+    required: false
+    default: "default"
+  gpu_count:
+    description: "Fake GPU count advertised per pool node"
+    required: false
+    default: "2"
+  helm_timeout:
+    description: "Helm --wait timeout"
+    required: false
+    default: "5m"
+
+runs:
+  using: "composite"
+  steps:
+    - name: Setup Helm
+      uses: azure/setup-helm@9bc31f4ebc9c6b171d7bfbaa5d006ae7abdb4310 
+      with:
+        version: v3.17.3
+
+    # Kind already removes control-plane NoSchedule on single-node clusters
+    # (create-cluster uses default one-node Kind), so no untaint step here.
+
+    - name: Label nodes for Fake GPU Operator pool
+      shell: bash
+      env:
+        NODE_POOL: ${{ inputs.node_pool }}
+      run: |
+        kubectl label node --all \
+          "run.ai/simulated-gpu-node-pool=${NODE_POOL}" \
+          --overwrite
+
+    - name: Install Fake GPU Operator (fake backend)
+      shell: bash
+      env:
+        CHART_VERSION: ${{ inputs.chart_version }}
+        NAMESPACE: ${{ inputs.namespace }}
+        NODE_POOL: ${{ inputs.node_pool }}
+        GPU_COUNT: ${{ inputs.gpu_count }}
+        HELM_TIMEOUT: ${{ inputs.helm_timeout }}
+        ACTION_PATH: ${{ github.action_path }}
+      run: |
+        "${ACTION_PATH}/install-fgo-fake.sh"
+
+    - name: Wait for allocatable nvidia.com/gpu
+      shell: bash
+      env:
+        NAMESPACE: ${{ inputs.namespace }}
+        ACTION_PATH: ${{ github.action_path }}
+      run: |
+        "${ACTION_PATH}/wait-for-allocatable-gpu.sh"
--- a/.github/actions/install-fgo-fake/wait-for-allocatable-gpu.sh
+++ b/.github/actions/install-fgo-fake/wait-for-allocatable-gpu.sh
@@ -1,0 +1,47 @@
+#!/usr/bin/env bash
+# Wait until a node advertises allocatable nvidia.com/gpu.
+# Env: NAMESPACE (FGO namespace, for diagnostics)
+
+set -euo pipefail
+
+# print output on success; warn on failure without aborting.
+run_diagnostic() {
+  local label=$1
+  shift
+  local out
+  if ! out=$("$@" 2>&1); then
+    echo "WARNING: ${label} failed: ${out}" >&2
+  else
+    printf '%s\n' "${out}"
+  fi
+}
+
+echo "Waiting for nodes to advertise nvidia.com/gpu..."
+deadline=$((SECONDS + 300))
+while (( SECONDS < deadline )); do
+  gpu_lines=()
+  if ! out=$(kubectl get nodes -o jsonpath='{range .items[*]}{.metadata.name}{"="}{.status.allocatable.nvidia\.com/gpu}{"\n"}{end}' 2>&1); then
+    echo "WARNING: kubectl get nodes (GPU poll) failed: ${out}" >&2
+  else
+    mapfile -t gpu_lines <<< "${out}"
+  fi
+  printf '%s\n' "${gpu_lines[@]}"
+  for line in "${gpu_lines[@]}"; do
+    gpu_qty="${line#*=}"
+    if [[ -n "${gpu_qty}" && "${gpu_qty}" != "0" && "${gpu_qty}" != "<none>" ]]; then
+      echo "Allocatable nvidia.com/gpu is ready (${line})."
+      run_diagnostic "kubectl get nodes" \
+        kubectl get nodes -o custom-columns='NAME:.metadata.name,GPU:.status.allocatable.nvidia\.com/gpu,POOL:.metadata.labels.run\.ai/simulated-gpu-node-pool'
+      run_diagnostic "kubectl get pods" \
+        kubectl get pods -n "${NAMESPACE}" -o wide
+      exit 0
+    fi
+  done
+  sleep 5
+done
+
+echo "Timed out waiting for nvidia.com/gpu allocatable."
+run_diagnostic "kubectl get nodes" kubectl get nodes -o yaml
+run_diagnostic "kubectl get pods" kubectl get pods -n "${NAMESPACE}" -o wide
+run_diagnostic "kubectl describe nodes" kubectl describe nodes
+exit 1
--- a/.github/workflows/e2e-test.yml
+++ b/.github/workflows/e2e-test.yml
@@ -23,6 +23,7 @@ on:
       - '.github/workflows/image-builds.yml'
       - '.github/actions/create-cluster/**'
       - '.github/actions/deploy/**'
+      - '.github/actions/install-fgo-fake/**'
       - '.github/actions/test-and-report/**'
       - '.github/resources/**'
       - 'api/**'
@@ -96,6 +97,13 @@ jobs:
             cache_enabled: "true"
             pod_to_pod_tls_enabled: "true"
             test_label: "E2ECritical"
+          # Schedule-only NVIDIA GPU lane (Kind + Fake GPU Operator fake backend).
+          - k8s_version: "v1.36.1"
+            cache_enabled: "true"
+            argo_version: "v3.7.17"
+            proxy: "false"
+            test_label: "gpu-scheduling-check"
+            pod_to_pod_tls_enabled: "false"
       fail-fast: false
     name: End to End ${{ matrix.test_label}} Tests - K8s ${{ matrix.k8s_version }} cacheEnabled=${{ matrix.cache_enabled }} argoVersion=${{ matrix.argo_version}} proxy=${{ matrix.proxy}} pod_to_pod_tls_enabled=${{ matrix.pod_to_pod_tls_enabled }}
     steps:
@@ -122,6 +130,11 @@ jobs:
           image_tag: latest
           image_registry: kind-registry:5000
 
+      - name: Install Fake GPU Operator (fake backend)
+        uses: ./.github/actions/install-fgo-fake
+        id: install-fgo-fake
+        if: ${{ matrix.test_label == 'gpu-scheduling-check' && steps.deploy.outcome == 'success' }}
+
       - name: Configure Cluster CA Cert for TLS-Enabled Tests
         shell: bash
         if: ${{ matrix.pod_to_pod_tls_enabled == 'true'}}
@@ -131,7 +144,7 @@ jobs:
       - name: Configure Input Variables
         shell: bash
         id: configure
-        if: ${{ steps.deploy.outcome == 'success' }}
+        if: ${{ steps.deploy.outcome == 'success' && steps.install-fgo-fake.outcome != 'failure' }}
         env:
           DISPATCH_NUMBER_OF_NODES: ${{ inputs.number_of_parallel_tests }}
           DISPATCH_TEST_LABEL: ${{ inputs.test_label }}
--- a/test_data/sdk_compiled_pipelines/valid/gpu-scheduling/nvidia_gpu_scheduling_check.py
+++ b/test_data/sdk_compiled_pipelines/valid/gpu-scheduling/nvidia_gpu_scheduling_check.py
@@ -1,0 +1,63 @@
+# Copyright 2026 The Kubeflow Authors
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#      https://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+"""Schedule-only NVIDIA GPU check for Kind + Fake GPU Operator (fake backend).
+
+Requests nvidia.com/gpu and a NoSchedule toleration. Succeeds when the task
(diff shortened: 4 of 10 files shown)


.github/actions/install-fgo-fake/action.yml
```diff
@@ -0,0 +1,92 @@
+name: "Install Fake GPU Operator (fake backend)"
+description: >
+  Install Run:ai Fake GPU Operator with the default fake backend on Kind,
+  label nodes into the simulated GPU pool, and wait until nvidia.com/gpu is allocatable.
+  Used by the gpu-scheduling-check E2E lane (schedule-only; no CUDA/mock).
```
hbelmiro: I think we can remove this line. It's harmless and actually true, but it tend to stale when people forget to update it when using in other workflows.

.github/actions/install-fgo-fake/action.yml
```diff
@@ -0,0 +1,92 @@
...
+  helm_timeout:
+    description: "Helm --wait timeout"
+    required: false
+    default: "5m"
+
+runs:
+  using: "composite"
+  steps:
+    - name: Setup Helm
+      uses: azure/setup-helm@v4
```
hbelmiro: Pinning to a particular SHA helps mitigate the risk of a bad actor adding a backdoor to the action's repository, as they would need to generate a SHA-1 collision for a valid Git object payload.

https://docs.github.com/en/actions/reference/security/secure-use#using-third-party-actions

```suggestion
      uses: azure/setup-helm@9bc31f4ebc9c6b171d7bfbaa5d006ae7abdb4310
```

.github/actions/install-fgo-fake/action.yml
```diff
@@ -0,0 +1,92 @@
...
+      with:
+        version: v3.17.3
+
+    # Kind already removes control-plane NoSchedule on single-node clusters
+    # (create-cluster uses default one-node Kind), so no untaint step here.
+
+    - name: Label nodes for Fake GPU Operator pool
+      shell: bash
+      run: |
+        set -euo pipefail
```
hbelmiro: This is not needed when `shell: bash`.
```suggestion
```

.github/actions/install-fgo-fake/action.yml
```diff
@@ -0,0 +1,92 @@
...
+        kubectl label node --all \
+          "run.ai/simulated-gpu-node-pool=${{ inputs.node_pool }}" \
+          --overwrite
+
+    - name: Install Fake GPU Operator (fake backend)
+      shell: bash
+      run: |
+        set -euo pipefail
```
hbelmiro:
```suggestion
```

.github/actions/install-fgo-fake/action.yml
```diff
@@ -0,0 +1,92 @@
...
+          --namespace "${{ inputs.namespace }}" \
+          --create-namespace \
+          --set "topology.nodePools.${{ inputs.node_pool }}.gpuCount=${{ inputs.gpu_count }}" \
+          --wait \
+          --timeout "${{ inputs.helm_timeout }}"
+
+    - name: Wait for allocatable nvidia.com/gpu
+      shell: bash
+      run: |
+        set -euo pipefail
```
hbelmiro:
```suggestion
```

.github/actions/install-fgo-fake/action.yml
```diff
@@ -0,0 +1,92 @@
...
+          --version "${{ inputs.chart_version }}" \
+          --namespace "${{ inputs.namespace }}" \
+          --create-namespace \
+          --set "topology.nodePools.${{ inputs.node_pool }}.gpuCount=${{ inputs.gpu_count }}" \
+          --wait \
+          --timeout "${{ inputs.helm_timeout }}"
+
+    - name: Wait for allocatable nvidia.com/gpu
+      shell: bash
+      run: |
```
hbelmiro, line 65: Can we extract this script to a `.sh` file?

.github/actions/install-fgo-fake/action.yml
```diff
@@ -0,0 +1,92 @@
...
+        set -euo pipefail
+        kubectl label node --all \
+          "run.ai/simulated-gpu-node-pool=${{ inputs.node_pool }}" \
+          --overwrite
+
+    - name: Install Fake GPU Operator (fake backend)
+      shell: bash
+      run: |
```
hbelmiro, line 57: Can we extract this script to a `.sh` file?

.github/workflows/e2e-test.yml
```diff
@@ -131,7 +144,7 @@ jobs:
       - name: Configure Input Variables
         shell: bash
         id: configure
-        if: ${{ steps.deploy.outcome == 'success' }}
+        if: ${{ steps.deploy.outcome == 'success' && (matrix.test_label != 'gpu-scheduling-check' || steps.install-fgo-fake.outcome == 'success') }}
```
nsingla: can this be simply:
`if: ${{ steps.deploy.outcome == 'success' && steps.install-fgo-fake.outcome != 'failure') }}` which means this step will run even if `install-fgo-fake` is skipped

.github/workflows/e2e-test.yml
```diff
@@ -152,6 +165,8 @@ jobs:
             # Each spec fans out into nested child pipelines, so retain its
             # historical lower cap instead of using the global default.
             NUMBER_OF_NODES=3
+          elif [ "$TEST_LABEL" == "gpu-scheduling-check" ]; then
+            NUMBER_OF_NODES=1
```
nsingla: Can gpu tests not run in parallel? does it have to hard coded to 1?

test_data/sdk_compiled_pipelines/valid/gpu-scheduling/nvidia_gpu_scheduling_check.py
```diff
@@ -0,0 +1,63 @@
...
+        operator="Exists",
+        effect="NoSchedule",
+    )
+
+
+@dsl.container_component
+def verify_gpu_scheduled():
+    """Lightweight success marker once the GPU-requesting pod runs."""
+    return dsl.ContainerSpec(
+        image='python:3.11-alpine',
```
hbelmiro: The container image `python:3.11-alpine` is not in the pre-loaded runtime base images list (`.github/resources/runtime-base-images.txt`). This means the GPU scheduling pod will pull the image from Docker Hub at runtime inside Kind, which is subject to Docker Hub rate limits and network flakiness that have historically caused CI failures in this project.

**Why it matters:** The runtime-base-images infrastructure exists specifically to avoid Docker Hub pulls during test execution. Other E2E pipelines use `python:3.11` which is pre-loaded.

```suggestion
        image='python:3.11',
```

.github/workflows/e2e-test.yml
sduvvuri1603: Hardcoded to 1 because there’s only one fixture today — more like “one test → one worker” convenience, not an FGO limitation. 
Dropping it  so this doesn't look like a special case;  GPU tests can be run in parallel.

.github/workflows/e2e-test.yml
sduvvuri1603: Thanks for the simplification. Will use this instead!

.github/actions/install-fgo-fake/action.yml
sduvvuri1603: I agree — removed the line so the description doesn’t go stale if other workflows reuse this

.github/actions/install-fgo-fake/action.yml
sduvvuri1603: Thanks for the reference and explanation. Replaced version with SHA

test_data/sdk_compiled_pipelines/valid/gpu-scheduling/nvidia_gpu_scheduling_check.py
sduvvuri1603: Thanks for pointing it out. Will keep in mind from next time!

.github/actions/install-fgo-fake/action.yml
sduvvuri1603: Extracted script into [‎.github/actions/install-fgo-fake/wait-for-allocatable-gpu.sh](https://github.com/kubeflow/pipelines/pull/13930/changes#diff-5bc912c60e3d39893b9885a2dc6b00f215e8ada5494a3e1aa5c1b9bb71246846)

.github/actions/install-fgo-fake/action.yml
sduvvuri1603: Extracted script into [‎.github/actions/install-fgo-fake/install-fgo-fake.sh](https://github.com/kubeflow/pipelines/pull/13930/changes#diff-afec7b07803f37384d600bbd55b8160d0bb81ebdaed9d7a36e1a1eff38779fb4)

.github/actions/install-fgo-fake/action.yml
```diff
@@ -0,0 +1,64 @@
...
+
+    # Kind already removes control-plane NoSchedule on single-node clusters
+    # (create-cluster uses default one-node Kind), so no untaint step here.
+
+    - name: Label nodes for Fake GPU Operator pool
+      shell: bash
+      run: |
+        kubectl label node --all \
+          "run.ai/simulated-gpu-node-pool=${{ inputs.node_pool }}" \
+          --overwrite
```
hbelmiro, line 46: The "Label nodes" step expands `${{ inputs.node_pool }}` directly in the `run:` block. Passing inputs through `env:` is always the safer pattern — today these inputs only have hardcoded defaults, but if callers start passing user-controlled values in the future, direct `${{ }}` expansion in `run:` becomes an injection vector. Using `env:` consistently now avoids that risk by default.


```suggestion
    - name: Label nodes for Fake GPU Operator pool
      shell: bash
      env:
        NODE_POOL: ${{ inputs.node_pool }}
      run: |
        kubectl label node --all \
          "run.ai/simulated-gpu-node-pool=${NODE_POOL}" \
          --overwrite
```

.github/actions/install-fgo-fake/action.yml
```diff
@@ -0,0 +1,64 @@
...
+      shell: bash
+      env:
+        CHART_VERSION: ${{ inputs.chart_version }}
+        NAMESPACE: ${{ inputs.namespace }}
+        NODE_POOL: ${{ inputs.node_pool }}
+        GPU_COUNT: ${{ inputs.gpu_count }}
+        HELM_TIMEOUT: ${{ inputs.helm_timeout }}
+      run: |
+        chmod +x "${{ github.action_path }}/install-fgo-fake.sh"
+        "${{ github.action_path }}/install-fgo-fake.sh"
```
hbelmiro: Same regarding using `env`.
```suggestion
    - name: Install Fake GPU Operator (fake backend)
      shell: bash
      env:
        CHART_VERSION: ${{ inputs.chart_version }}
        NAMESPACE: ${{ inputs.namespace }}
        NODE_POOL: ${{ inputs.node_pool }}
        GPU_COUNT: ${{ inputs.gpu_count }}
        HELM_TIMEOUT: ${{ inputs.helm_timeout }}
        ACTION_PATH: ${{ github.action_path }}
      run: |
        chmod +x "${ACTION_PATH}/install-fgo-fake.sh"
        "${ACTION_PATH}/install-fgo-fake.sh"
```

.github/actions/install-fgo-fake/action.yml
```diff
@@ -0,0 +1,64 @@
...
+
+    - name: Wait for allocatable nvidia.com/gpu
+      shell: bash
+      env:
+        NAMESPACE: ${{ inputs.namespace }}
+      run: |
+        chmod +x "${{ github.action_path }}/wait-for-allocatable-gpu.sh"
+        "${{ github.action_path }}/wait-for-allocatable-gpu.sh"
```
hbelmiro: Same regarding using `env`.

```suggestion
    - name: Wait for allocatable nvidia.com/gpu
      shell: bash
      env:
        NAMESPACE: ${{ inputs.namespace }}
        ACTION_PATH: ${{ github.action_path }}
      run: |
        chmod +x "${ACTION_PATH}/wait-for-allocatable-gpu.sh"
        "${ACTION_PATH}/wait-for-allocatable-gpu.sh"
```

.github/actions/install-fgo-fake/action.yml
nsingla: even in github ci runners? i remember reading something about it

test_data/sdk_compiled_pipelines/valid/gpu-scheduling/nvidia_gpu_scheduling_check.py
nsingla: can we also change this to `registry.access.redhat.com/ubi9/python-311:1-1779945715`, I believe this is what we have in few of the other pipelines as well (makes the pipeline run compatible on a fips cluster)

sduvvuri1603: @nsingla Makes sense. One doubt before I switch: .github/resources/runtime-base-images.txt currently preloads registry.access.redhat.com/ubi9/python-311:latest but not :1-1779945715 which means kind still pulls live at runtime (@hbelmiro original concern) even if RH registry is less rate-limit-y than Docker Hub. Is that ok?

hbelmiro: @nsingla this is upstream KFP? Why do you think we should use the Red Hat image?

hbelmiro:
/hold 
I have some more comments

.github/actions/install-fgo-fake/wait-for-allocatable-gpu.sh
hbelmiro, line 1: The diagnostic `kubectl` commands in this script either use `|| true` (lines 20, 28–30) which silently swallows failures, or have no guard at all (line 19) which under `set -euo pipefail` can turn a successful GPU detection into a script error. All diagnostic commands should log a warning on failure instead, so CI logs show what happened without silently aborting or hiding errors. For example:

```bash
if ! out=$(kubectl get nodes -o yaml 2>&1); then
  echo "WARNING: kubectl get nodes failed: ${out}" >&2
fi
```

.github/actions/install-fgo-fake/action.yml
```diff
@@ -0,0 +1,64 @@
...
+    - name: Install Fake GPU Operator (fake backend)
+      shell: bash
+      env:
+        CHART_VERSION: ${{ inputs.chart_version }}
+        NAMESPACE: ${{ inputs.namespace }}
+        NODE_POOL: ${{ inputs.node_pool }}
+        GPU_COUNT: ${{ inputs.gpu_count }}
+        HELM_TIMEOUT: ${{ inputs.helm_timeout }}
+      run: |
+        chmod +x "${{ github.action_path }}/install-fgo-fake.sh"
```
hbelmiro: The file is already committed with that permission.

```suggestion
```

.github/actions/install-fgo-fake/action.yml
```diff
@@ -0,0 +1,64 @@
...
+        "${{ github.action_path }}/install-fgo-fake.sh"
+
+    - name: Wait for allocatable nvidia.com/gpu
+      shell: bash
+      env:
+        NAMESPACE: ${{ inputs.namespace }}
+      run: |
+        chmod +x "${{ github.action_path }}/wait-for-allocatable-gpu.sh"
```
hbelmiro: The file is already committed with that permission.
```suggestion
```

sduvvuri1603: Got it. Removed the redundant chmod permisson!

.github/actions/install-fgo-fake/action.yml
sduvvuri1603: Got it. Removed the redundant chmod permisson!

.github/actions/install-fgo-fake/action.yml
sduvvuri1603: Thanks for pointing that out, changed code to pass through env

.github/actions/install-fgo-fake/action.yml
sduvvuri1603: Modified to pass through env

.github/actions/install-fgo-fake/action.yml
sduvvuri1603: Modified to pass through env

.github/actions/install-fgo-fake/wait-for-allocatable-gpu.sh
sduvvuri1603: Makes sense. Updated the diagnostic calls to your pattern (WARNING on failure, don’t abort). Also added print ${out} on success so the dumps still show in the log, and folded that into a small run_diagnostic helper to avoid repetitiveness. Pls lmk if this works!

.github/actions/install-fgo-fake/wait-for-allocatable-gpu.sh
```diff
@@ -0,0 +1,45 @@
...
+    printf '%s\n' "${out}"
+  fi
+}
+
+echo "Waiting for nodes to advertise nvidia.com/gpu..."
+deadline=$((SECONDS + 300))
+while (( SECONDS < deadline )); do
+  mapfile -t gpu_lines < <(
+    kubectl get nodes -o jsonpath='{range .items[*]}{.metadata.name}{"="}{.status.allocatable.nvidia\.com/gpu}{"\n"}{end}' \
+      2>/dev/null || true
```
hbelmiro: You've missed this one.  It swallows all kubectl failures.

sduvvuri1603: Sorry about that. Pushed the changes in!

hbelmiro (approved):
/lgtm
/approve
/unhold

hbelmiro (commented):
/lgtm

PR #13931, opened 2026-07-29
fix(backend): break infinite loop in isV2Spec on malformed YAML

### LINKED ISSUE #13926: Malformed YAML upload causes infinite loop in isV2Spec and permanent goroutine leak
## Environment

- KFP API server (any version using `gopkg.in/yaml.v3` in `template.go`)
- Reproducible with `gopkg.in/yaml.v3 v3.0.1`

## What happened

Uploading a `.yaml` pipeline file containing malformed YAML (e.g., unescaped single quotes inside a single-quoted scalar) causes the API server to hang permanently. The upload never returns a response, and the goroutine is leaked — it never unblocks, even after the client disconnects. Each such upload leaks one goroutine.

If a reverse proxy with a response timeout sits in front of the API server, the timeout fires first and the client sees a **502 Bad Gateway**.

## Workaround

Validate your pipeline YAML before uploading. The most common cause is unescaped single quotes (`'`) inside a single-quoted YAML scalar. For example, Python code injected into a YAML template may contain contractions like `doesn't` inside single-quoted strings:

```yaml
# Broken — bare ' inside a single-quoted scalar
command: 'echo "it doesn't work"'

# Fixed — double the inner quote to escape it
command: 'echo "it doesn''t work"'

# Or use a double-quoted scalar instead
command: "echo \"it doesn't work\""
```

You can validate your YAML before uploading with:

```bash
python3 -c "import yaml; yaml.safe_load(open('pipeline.yaml'))"
```

If validation fails, fix the quoting in the tool or script that generates the YAML.

## Root cause

`isV2Spec()` in `backend/src/apiserver/template/template.go` decodes the uploaded YAML in a loop that only breaks on `io.EOF`:

```go
func isV2Spec(template []byte) bool {
	decoder := goyaml.NewDecoder(bytes.NewReader(template))
	for {
		var value map[string]interface{}
		err := decoder.Decode(&value)
		// Break at end of file
		if errors.Is(err, io.EOF) {
			break
		}
		if value == nil {
			continue
		}
		if isPipelineSpec(value) {
			return true
		}
	}
	return false
}
```

When `gopkg.in/yaml.v3`'s `Decoder.Decode()` encounters a parse error, it returns a non-EOF error (e.g., `yaml: found unexpected end of stream`). On subsequent calls, it returns `yaml: expected stream start event but got none` — **indefinitely**. It never returns `io.EOF`, so the loop never breaks.

## Reproduction

Minimal reproduction — upload a file containing a single `'` character:

```bash
printf "'" > /tmp/malformed.yaml
curl -X POST "$API_SERVER/apis/v2beta1/pipelines/upload?name=test" \
  -F "uploadfile=@/tmp/malformed.yaml"
# Hangs forever
```

Go test confirming the infinite loop:

```go
decoder := goyaml.NewDecoder(bytes.NewReader([]byte("'")))
for i := 0; i < 5; i++ {
    var value map[string]interface{}
    err := decoder.Decode(&value)
    fmt.Printf("iter %d: err=%v, isEOF=%v\n", i, err, errors.Is(err, io.EOF))
}
// Output:
// iter 0: err=yaml: found unexpected end of stream, isEOF=false
// iter 1: err=yaml: expected stream start event but got none, isEOF=false
// iter 2: err=yaml: expected stream start event but got none, isEOF=false
// iter 3: err=yaml: expected stream start event but got none, isEOF=false
// iter 4: err=yaml: expected stream start event but got none, isEOF=false
```

Characters that trigger the bug (any of these as the sole file content): `'`, `"`, `:`, `[`, `]`, `{`, `}`, `%`, `@`, `&`, `*`

## Expected behavior

Uploading malformed YAML should return an error response (e.g., 400 Bad Request), not hang.

## Suggested fix

Break on any error, not just `io.EOF`:

```go
err := decoder.Decode(&value)
if err != nil {
    break
}
```

This makes `isV2Spec` return `false` for malformed YAML, which causes `inferTemplateFormat` to return `Unknown`, which causes `template.New` to return an "unknown template format" error — the correct behavior.

hbelmiro:
## Summary

- `isV2Spec()` entered an infinite loop when processing malformed YAML uploads because the decode loop only broke on `io.EOF`, but `gopkg.in/yaml.v3`'s decoder never returns `io.EOF` after a parse error — it returns the same error indefinitely. This caused a permanent goroutine leak per upload and a hung HTTP connection (502 via reverse proxy).
- Break on any decode error and propagate real YAML parse errors through `inferTemplateFormat` and `template.New` so users see `"failed to parse pipeline spec YAML: <details>"` instead of a hang. Valid YAML that is not a pipeline spec still returns `"unknown template format"` (`io.EOF` is filtered as normal end-of-stream).

Fixes https://github.com/kubeflow/pipelines/issues/13926

## Test plan

- [x] `TestParseSpecFormat` — verifies format detection for valid V1, V2, multi-document V2, empty, and malformed YAML inputs; asserts exact error for malformed cases and no error for valid inputs
- [x] `TestNewTemplateMalformedYAML` — verifies user-facing error message contains `"failed to parse pipeline spec YAML"` for malformed input
- [x] `TestNewTemplateValidNonPipelineYAML` — verifies valid non-pipeline YAML returns `"unknown template format"` (not a misleading parse error)

### DIFF
--- a/backend/src/apiserver/template/template_test.go
+++ b/backend/src/apiserver/template/template_test.go
@@ -69,6 +69,7 @@ func TestParseSpecFormat(t *testing.T) {
 	tt := []struct {
 		template     string
 		templateType TemplateType
+		wantErr      string
 	}{{
 		// standard match
 		template: `
@@ -108,14 +109,54 @@ kind: CronWorkflow`,
 	}, {
 		template:     loadYaml(t, "testdata/hello_world.yaml"),
 		templateType: V2,
+	}, {
+		// multi-document V2 pipeline (pipeline spec + platforms section)
+		template:     loadYaml(t, "testdata/pipeline_with_volume.yaml"),
+		templateType: V2,
+	}, {
+		// valid YAML scalar (not a mapping) — not a pipeline spec
+		template:     "not a valid spec",
+		templateType: Unknown,
+	}, {
+		// malformed YAML: unclosed single-quoted scalar
+		template:     "'",
+		templateType: Unknown,
+		wantErr:      "yaml: found unexpected end of stream",
+	}, {
+		// malformed YAML: unclosed flow sequence
+		template:     "[",
+		templateType: Unknown,
+		wantErr:      "yaml: line 1: did not find expected node content",
+	}, {
+		// malformed YAML: incomplete directive
+		template:     "%",
+		templateType: Unknown,
+		wantErr:      "yaml: could not find expected directive name",
 	}}
 
 	for _, test := range tt {
-		format := inferTemplateFormat([]byte(test.template))
+		format, err := inferTemplateFormat([]byte(test.template))
 		if format != test.templateType {
 			t.Errorf("InferTemplateFormat(%s)=%q, expect %q", test.template, format, test.templateType)
 		}
+		if test.wantErr != "" {
+			require.EqualError(t, err, test.wantErr)
+		} else {
+			require.NoError(t, err)
+		}
 	}
+}
+
+func TestNewTemplateMalformedYAML(t *testing.T) {
+	_, err := New([]byte("'"), TemplateOptions{})
+	require.Error(t, err)
+	assert.Contains(t, err.Error(), "failed to parse pipeline spec YAML")
+}
+
+func TestNewTemplateValidNonPipelineYAML(t *testing.T) {
+	_, err := New([]byte("key: value"), TemplateOptions{})
+	require.Error(t, err)
+	assert.Contains(t, err.Error(), "unknown template format")
 }
 
 func unmarshalWf(yamlStr string) *v1alpha1.Workflow {
--- a/backend/src/apiserver/template/template_test.go
+++ b/backend/src/apiserver/template/template_test.go
@@ -16,6 +16,8 @@ package template
 
 import (
 	"encoding/json"
+	"errors"
+	"io"
 	"os"
 	"strings"
 	"testing"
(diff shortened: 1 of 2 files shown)


backend/src/apiserver/template/template_test.go
```diff
@@ -108,16 +109,56 @@ kind: CronWorkflow`,
...
 	}}
 
 	for _, test := range tt {
-		format := inferTemplateFormat([]byte(test.template))
+		format, err := inferTemplateFormat([]byte(test.template))
 		if format != test.templateType {
 			t.Errorf("InferTemplateFormat(%s)=%q, expect %q", test.template, format, test.templateType)
 		}
+		if test.wantErr != "" {
+			require.EqualError(t, err, test.wantErr)
```
VaniHaripriya: Malformed-YAML cases use require.EqualError with exact gopkg.in/yaml.v3 message strings (see wantErr on lines 124, 129, 134). Those strings are library-internal and can change across yaml.v3 patch releases, which could cause CI failures unrelated to behavior regressions.

Consider substring matching instead:

```
require.Error(t, err)
assert.Contains(t, err.Error(), "unexpected end of stream")
```
Or assert the error is present and not io.EOF:
```
require.Error(t, err)
require.False(t, errors.Is(err, io.EOF))

```

alyssacgoins (changes requested):
Changes look good - one general comment -
It would be useful to have an integration test here verifying this behavior at the general API server level

hbelmiro:
> Changes look good - one general comment - It would be useful to have an integration test here verifying this behavior at the general API server level

@alyssacgoins from my point of view, the unit tests cover enough. The fix is in pure functions with no dependencies. Also, we already have integration tests for invalid workflow specs.
Do you see any specific scenarios uncovered?

alyssacgoins (approved):
/lgtm

PR #13934, opened 2026-07-30
fix(backend): use image digest instead image name in cache key. Fixes #13862

### LINKED ISSUE #13862: [backend]  caching fingerprint does not take into account the fact that image tags can be mutable
Currently, cache fingerprints include the container image reference string, such as "my-registry/app:latest". When a mutable tag is overwritten with new image content, the fingerprint remains the same, and steps may incorrectly reuse cached outputs when the image has actually changed.
To avoid this issue, it would be better to use the hash of the image as part of the caching key, rather than just the image name. This ensures that the correct image version is always served, and the cache for the previous version is not accidentally used when updating the image.

### Environment

latest
*  KFP version: 
latest

### Steps to reproduce

1) Send the pipeline version with the image "my-registry/app:latest".
2) Run the pipeline version.
3) The results of the launch will be cached.
4) Update the image "my-registry/app:latest" with new code.
5) Run the updated pipeline version.
6) The steps were skipped because the values of the run were taken from the cache. 

### Expected Result
6) The steps were not skipped, as the output data may vary if the image code has been changed.

---

CrashCumber:
Fixes #13862

**Checklist:**

### DIFF
--- a/backend/src/v2/cmd/driver/main.go
+++ b/backend/src/v2/cmd/driver/main.go
@@ -96,18 +96,22 @@ var (
 	logLevel           = flag.String("log_level", "1", "The verbosity level to log.")
 
 	// proxy
-	httpProxy            = flag.String(httpProxyArg, unsetProxyArgValue, "The proxy for HTTP connections.")
-	httpsProxy           = flag.String(httpsProxyArg, unsetProxyArgValue, "The proxy for HTTPS connections.")
-	noProxy              = flag.String(noProxyArg, unsetProxyArgValue, "Addresses that should ignore the proxy.")
-	publishLogs          = flag.String("publish_logs", "true", "Whether to publish component logs to the object store")
-	cacheDisabledFlag    = flag.Bool("cache_disabled", false, "Disable cache globally.")
-	mlPipelineTLSEnabled = flag.Bool("ml_pipeline_tls_enabled", false, "Set to true if mlpipeline API server serves over TLS.")
-	metadataTLSEnabled   = flag.Bool("metadata_tls_enabled", false, "Set to true if MLMD serves over TLS.")
-	caCertPath           = flag.String("ca_cert_path", "", "The path to the CA certificate to trust on connections to the ML pipeline API server and metadata server.")
-	defaultRunAsUser     = flag.Int64("default_run_as_user", -1, "Admin-configured default runAsUser for user containers. -1 means not set.")
-	defaultRunAsGroup    = flag.Int64("default_run_as_group", -1, "Admin-configured default runAsGroup for user containers. -1 means not set.")
-	defaultRunAsNonRoot  = flag.String("default_run_as_non_root", "", "Admin-configured default runAsNonRoot for user containers. Empty means not set.")
-	defaultHostUsers     = flag.String("default_host_users", "", "Administrator-configured default hostUsers for user workload pods. Empty means not set. Set to false to run pods in a dedicated Linux user namespace.")
+	httpProxy                          = flag.String(httpProxyArg, unsetProxyArgValue, "The proxy for HTTP connections.")
+	httpsProxy                         = flag.String(httpsProxyArg, unsetProxyArgValue, "The proxy for HTTPS connections.")
+	noProxy                            = flag.String(noProxyArg, unsetProxyArgValue, "Addresses that should ignore the proxy.")
+	publishLogs                        = flag.String("publish_logs", "true", "Whether to publish component logs to the object store")
+	cacheDisabledFlag                  = flag.Bool("cache_disabled", false, "Disable cache globally.")
+	cacheResolveImageDigestFlag        = flag.Bool("cache_resolve_image_digest", false, "Resolve container image tags to digests when computing cache fingerprints. Disabled by default.")
+	cacheImageDigestDockerConfig       = flag.String("cache_image_digest_docker_config", "", "Path to Docker config.json / .dockerconfigjson used for private registry auth when resolving image digests.")
+	cacheImageDigestDockerHubRegistry  = flag.String("cache_image_digest_dockerhub_registry_host", "", "Docker Hub / mirror registry API host (default: registry-1.docker.io).")
+	cacheImageDigestInsecureRegistries = flag.String("cache_image_digest_insecure_registries", "", "Comma-separated registry hosts that should use HTTP for digest resolution.")
+	mlPipelineTLSEnabled               = flag.Bool("ml_pipeline_tls_enabled", false, "Set to true if mlpipeline API server serves over TLS.")
+	metadataTLSEnabled                 = flag.Bool("metadata_tls_enabled", false, "Set to true if MLMD serves over TLS.")
+	caCertPath                         = flag.String("ca_cert_path", "", "The path to the CA certificate to trust on connections to the ML pipeline API server and metadata server.")
+	defaultRunAsUser                   = flag.Int64("default_run_as_user", -1, "Admin-configured default runAsUser for user containers. -1 means not set.")
+	defaultRunAsGroup                  = flag.Int64("default_run_as_group", -1, "Admin-configured default runAsGroup for user containers. -1 means not set.")
+	defaultRunAsNonRoot                = flag.String("default_run_as_non_root", "", "Admin-configured default runAsNonRoot for user containers. Empty means not set.")
+	defaultHostUsers                   = flag.String("default_host_users", "", "Administrator-configured default hostUsers for user workload pods. Empty means not set. Set to false to run pods in a dedicated Linux user namespace.")
 )
 
 // func RootDAG(pipelineName string, runID string, component *pipelinespec.ComponentSpec, task *pipelinespec.PipelineTaskSpec, mlmd *metadata.Client) (*Execution, error) {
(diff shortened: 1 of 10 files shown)


manifests/kustomize/base/installs/generic/pipeline-install-config.yaml
```diff
@@ -84,6 +84,28 @@ data:
...
+  ## CACHE_RESOLVE_IMAGE_DIGEST controls whether the API server asks the container
+  ## driver to resolve mutable image tags to content digests when computing v2
+  ## cache fingerprints. Default "false" preserves historical behavior (image
+  ## reference string as-is). Set to "true" to enable.
+  ## When enabled, resolution failures (network, auth, unsupported registry)
+  ## fall back to the original image reference so the run continues; mutable-tag
+  ## cache hits remain possible in that fail-open path. Enabling this flag also
+  ## normalizes resolved refs (e.g. ubuntu:22.04 -> docker.io/library/ubuntu@sha256:...),
+  ## which invalidates previously written cache entries that used the tag string.
+  CACHE_RESOLVE_IMAGE_DIGEST: "false"
```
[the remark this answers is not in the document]
CrashCumber: Keeping this opt-in is intentional, not an incomplete fix.
Enabling digest resolution by default would:
- add a registry network dependency on every cached container task;
- invalidate existing cache entries via ref normalization (ubuntu:22.04 → docker.io/library/ubuntu@sha256:…);
- require private-registry auth setup for many deployments.

With the flag off, behavior matches historical fingerprinting. Admins who hit [#13862](https://github.com/kubeflow/pipelines/issues/13862) can set CACHE_RESOLVE_IMAGE_DIGEST=true (and optionally mount a dockerconfig secret).
I’ll clarify in the PR description that this is an opt-in mitigation, not a default-on behavior change. Happy to revisit defaulting to true in a follow-up if maintainers want that trade-off.

backend/src/v2/driver/image_digest.go
```diff
@@ -0,0 +1,522 @@
...
+		resolver = func(img string) (string, error) {
+			return resolveImageDigestHTTP(img, client, cfg)
+		}
+	}
+
+	resolved, err := resolver(image)
+	if err != nil {
+		glog.Warningf("Failed to resolve image digest for %q; using image reference as-is for cache key: %v", image, err)
+		return image
+	}
```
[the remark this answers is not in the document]
CrashCumber: Agreed this is a real consistency gap for `IfNotPresent` / `Never`, but it’s out of scope for #13862. With `Always`, leaving the workload on the mutable tag can be expected behavior: the same pipeline version may be reused while the underlying image tag is updated, and the kubelet is expected to pull the current tag at pod start. Fingerprinting by digest still avoids stale cache reuse; pinning the pod to `@sha256:…` would freeze execution to the digest resolved at driver time and change that workflow.
[#13862](https://github.com/kubeflow/pipelines/issues/13862) asks that cache fingerprints not reuse stale outputs after a mutable tag is overwritten. Resolving the tag to a digest for the fingerprint addresses that. Changing what the user container actually runs (pull policy semantics, node-local image cache, failure modes when registry lookup fails) deserves a separate design/PR.

backend/src/v2/cmd/driver/main.go
```diff
@@ -96,18 +96,22 @@ var (
...
+	httpProxy                          = flag.String(httpProxyArg, unsetProxyArgValue, "The proxy for HTTP connections.")
+	httpsProxy                         = flag.String(httpsProxyArg, unsetProxyArgValue, "The proxy for HTTPS connections.")
+	noProxy                            = flag.String(noProxyArg, unsetProxyArgValue, "Addresses that should ignore the proxy.")
+	publishLogs                        = flag.String("publish_logs", "true", "Whether to publish component logs to the object store")
+	cacheDisabledFlag                  = flag.Bool("cache_disabled", false, "Disable cache globally.")
+	cacheResolveImageDigestFlag        = flag.Bool("cache_resolve_image_digest", false, "Resolve container image tags to digests when computing cache fingerprints. Disabled by default.")
+	cacheImageDigestDockerConfig       = flag.String("cache_image_digest_docker_config", "", "Path to Docker config.json / .dockerconfigjson used for private registry auth when resolving image digests.")
```
juliusvonkohout, line 105: We use OCI not Docker.

backend/src/v2/cmd/driver/main.go
```diff
@@ -96,18 +96,22 @@ var (
...
+	cacheImageDigestDockerHubRegistry  = flag.String("cache_image_digest_dockerhub_registry_host", "", "Docker Hub / mirror registry API host (default: registry-1.docker.io).")
```
juliusvonkohout, line 106: We use OCI not Docker

backend/src/v2/cmd/driver/main.go
```diff
@@ -420,16 +424,22 @@ func drive() (err error) {
...
-		MLMDServerAddress:          *mlmdServerAddress,
-		MLMDServerPort:             *mlmdServerPort,
-		MLPipelineTLSEnabled:       *mlPipelineTLSEnabled,
-		MLMDTLSEnabled:             *metadataTLSEnabled,
-		CaCertPath:                 *caCertPath,
-		PluginDispatcher:           pluginDispatcher,
+		CacheResolveImageDigest:    *cacheResolveImageDigestFlag,
+		CacheImageDigestConfig: driver.ImageDigestResolveConfig{
+			DockerConfigPath:      *cacheImageDigestDockerConfig,
+			DockerHubRegistryHost: *cacheImageDigestDockerHubRegistry,
```
juliusvonkohout, line 430: OCI not Docker

backend/src/v2/compiler/argocompiler/container.go
```diff
@@ -46,27 +46,36 @@ const (
...
+	PublishLogsEnvVar                        = "PUBLISH_LOGS"
+	CacheResolveImageDigestEnvVar            = "CACHE_RESOLVE_IMAGE_DIGEST"
+	CacheImageDigestDockerConfigSecretEnvVar = "CACHE_IMAGE_DIGEST_DOCKER_CONFIG_SECRET"
+	CacheImageDigestDockerHubRegistryEnvVar  = "CACHE_IMAGE_DIGEST_DOCKERHUB_REGISTRY_HOST"
+	CacheImageDigestInsecureRegistriesEnvVar = "CACHE_IMAGE_DIGEST_INSECURE_REGISTRIES"
+
+	cacheImageDigestDockerConfigVolume = "cache-image-digest-docker-config"
+	cacheImageDigestDockerConfigMount  = "/var/run/secrets/kfp-cache-image-digest"
+	cacheImageDigestDockerConfigPath   = cacheImageDigestDockerConfigMount + "/.dockerconfigjson"
+	cacheImageDigestDockerConfigKey    = ".dockerconfigjson"
```
juliusvonkohout, line 66: Docker is the wrong term, it is a company name, but we use OCI.

backend/src/v2/compiler/argocompiler/container_test.go
```diff
@@ -131,6 +131,96 @@ func TestContainerDriverTemplate_IncludesPipelineJobCreateTimeArg(t *testing.T)
...
+
+		var foundMount bool
+		for _, mount := range tmpl.Container.VolumeMounts {
+			if mount.Name == cacheImageDigestDockerConfigVolume {
+				foundMount = true
+				assert.Equal(t, cacheImageDigestDockerConfigMount, mount.MountPath)
+				assert.True(t, mount.ReadOnly)
+			}
+		}
+		assert.True(t, foundMount, "expected dockerconfig volume mount on container driver")
```
juliusvonkohout, line 205: We do not use Docker.

backend/src/v2/compiler/argocompiler/container.go
```diff
@@ -796,3 +817,40 @@ func extendMetadataMap(
...
+					Key:  cacheImageDigestDockerConfigKey,
+					Path: cacheImageDigestDockerConfigKey,
+				}},
+			},
+		},
+	})
+	tmpl.Container.VolumeMounts = append(tmpl.Container.VolumeMounts, k8score.VolumeMount{
+		Name:      cacheImageDigestDockerConfigVolume,
+		MountPath: cacheImageDigestDockerConfigMount,
+		ReadOnly:  true,
```
juliusvonkohout, line 854: We use OCI not Docker.

juliusvonkohout (changes requested):
Please check the spelling and naming. OCI image for example and we do not support the "Docker deamon" and are not directly affiliated with that company.

manifests/kustomize/base/pipeline/ml-pipeline-apiserver-deployment.yaml
```diff
@@ -218,6 +218,31 @@ spec:
...
+                configMapKeyRef:
+                  name: pipeline-install-config
+                  key: CACHE_IMAGE_DIGEST_DOCKERHUB_REGISTRY_HOST
+                  optional: true
+            - name: CACHE_IMAGE_DIGEST_INSECURE_REGISTRIES
+              valueFrom:
+                configMapKeyRef:
+                  name: pipeline-install-config
+                  key: CACHE_IMAGE_DIGEST_INSECURE_REGISTRIES
+                  optional: true
```
juliusvonkohout, line 245: OCI not Docker

droctothorpe:
Hi @CrashCumber, thank you for your PR! The linked issue (#13862) must be labeled `ready` before your contribution can be reviewed. Please check our [CONTRIBUTING.md](https://github.com/kubeflow/pipelines/blob/master/CONTRIBUTING.md) for details.

PR #13935, opened 2026-07-30
fix(backend): close SQL rows on all early return paths in storage layer

shivv23:
## Description

In the API server storage layer, `defer rows.Close()` is placed after the `scanRows` (or equivalent) call, not immediately after the `Query` succeeds. If `rows.Err()` or `scanRows` returns an error before the `defer` is reached, the result set leaks — the database connection is never returned to the pool. The same pattern affects `sizeRow` result sets.

Over time, accumulated leaked connections can exhaust the connection pool, causing the API server to become unavailable.

### Root cause

All five files follow the same pattern:

```go
rows, err := tx.Query(...)
if err != nil { ... }
if err := rows.Err(); err != nil { return ... }  // LEAK: rows not closed
result, err := s.scanRows(rows)
if err != nil { return ... }                      // LEAK: rows not closed
defer rows.Close()                                 // too late — early returns above skip this
```

### Fix

Move `defer rows.Close()` (and `defer sizeRow.Close()`) right after the error check for `Query`, before any early return path.

### Files changed

- `backend/src/apiserver/storage/experiment_store.go`
- `backend/src/apiserver/storage/task_store.go`
- `backend/src/apiserver/storage/pipeline_store.go`
- `backend/src/apiserver/storage/job_store.go`
- `backend/src/apiserver/storage/default_experiment_store.go`

## Checklist

- [x] I have signed off my commits with `git commit -s`

### DIFF
--- a/backend/src/apiserver/storage/default_experiment_store.go
+++ b/backend/src/apiserver/storage/default_experiment_store.go
@@ -46,12 +46,12 @@ func (s *DefaultExperimentStore) initializeDefaultExperimentTable() error {
 		tx.Rollback()
 		return util.NewInternalServerError(err, "Failed to get default experiment")
 	}
+	defer rows.Close()
 	if err := rows.Err(); err != nil {
 		tx.Rollback()
 		return util.NewInternalServerError(err, "Failed to get default experiment")
 	}
 	next := rows.Next()
-	defer rows.Close()
 
 	// If the table is not initialized, then set the default value.
 	if !next {
--- a/backend/src/apiserver/storage/experiment_store.go
+++ b/backend/src/apiserver/storage/experiment_store.go
@@ -96,6 +96,7 @@ func (s *ExperimentStore) ListExperiments(filterContext *model.FilterContext, op
 		tx.Rollback()
 		return errorF(err)
 	}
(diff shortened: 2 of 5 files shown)


shivv23:
@hbelmiro Hi! Could you please review this PR?

frontend/src/pages/NewRunSwitcher.tsx
hbelmiro, line 1: This seems unrelated.

frontend/src/pages/NewRunV2.tsx
hbelmiro, line 1: This seems unrelated.

hbelmiro (commented):
/ok-to-test

shivv23:
@hbelmiro done. please have a look.

frontend/src/pages/NewRunSwitcher.tsx
shivv23: fixed

frontend/src/pages/NewRunV2.tsx
shivv23: fixed

hbelmiro (approved):
/lgtm
/approve

PR #13938, opened 2026-07-30
fix(backend): tolerate task-level MLflow outages

mprahl:

Treat task-level MLflow integration as best-effort when a plugin cannot
start. The task plugin dispatcher now skips user container environment
variable retrieval for handlers whose `OnTaskStart` call failed. This lets
the user container run during an MLflow outage while preserving environment
variables from plugins that started successfully.

`injectUserEnvVars` controls whether MLflow integration context is exposed;
it does not make MLflow availability a pipeline requirement. Components that
use optional MLflow instrumentation can guard it with:

```python
if os.getenv("MLFLOW_RUN_ID"):
    mlflow.autolog()
```

Environment variable construction errors from successfully started plugins
continue to propagate.

Tests:

- Task plugin dispatcher tests: 20 passed
- MLflow task handler tests: 31 passed

**Checklist:**
- [x] The title for your pull request (PR) follows the repository title
  convention.

### DIFF
--- a/backend/src/v2/common/plugins/dispatcher.go
+++ b/backend/src/v2/common/plugins/dispatcher.go
@@ -125,6 +125,13 @@ func (t *TaskPluginDispatcherImpl) RetrieveUserContainerEnvVars(taskInfo *TaskIn
 
 	injectVarNameSet := map[string]bool{}
 	for _, handler := range t.handlers {
+		if t.startedHandlers != nil && !t.startedHandlers[handler.Name()] {
+			glog.Infof(
+				"Skipping user container env vars for handler %s (OnTaskStart did not succeed)",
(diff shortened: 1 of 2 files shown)


alyssacgoins (approved):

jeffspahr (approved):
No actionable findings. The dispatcher correctly skips environment retrieval only for handlers whose `OnTaskStart` failed, while preserving variables from successful handlers. The driver lifecycle and cached-task path remain consistent.

Verified at `748f99e881`:
- `go test -race ./backend/src/v2/common/plugins/...`
- `go vet ./backend/src/v2/common/plugins/...`
- `gofmt -d` and `git diff --check`

All passed/clean.

PR #13940, opened 2026-07-31
fix(backend): use pipeline version references for pinned recurring runs. Fixes #13933

### LINKED ISSUE #13933: Stop embedding compiled workflow manifests in ScheduledWorkflow for version-pinned recurring runs
## Summary

When a recurring run is created **without** pinning a pipeline version ("always use the latest version"), the `ScheduledWorkflow` controller creates runs by calling back into the API server's `RunService.CreateRun` with just `PipelineId`/`PipelineVersionId`, letting the API server resolve and compile the template fresh at trigger time. This avoids storing a compiled Argo manifest in the CRD.

However, when a recurring run **is** pinned to a specific pipeline version — including via the normal UI flow of selecting a version when creating a recurring run — the compiled workflow manifest is resolved once at creation time and baked into the `Job`/`ScheduledWorkflow` spec (`WorkflowSpecManifest`/`PipelineSpecManifest`). This is the same code path used for legacy, fully-freeform manifest submission.

This means the "reference at trigger time" architecture that already exists for the "always latest" case isn't actually available for the much more common production pattern of pinning to a specific, validated pipeline version.

## Problems this causes

1. **Duplication of state.** The compiled manifest stored in the `ScheduledWorkflow` CRD duplicates the pipeline version's spec, which is already stored durably (pipeline/pipeline-version tables + object store). The CRD becomes a second, driftable copy of the same logical artifact.
2. **Image lock-in / vulnerability exposure.** Because the manifest is compiled once at recurring-run-creation time, it bakes in whatever KFP launcher/driver/sidecar image references were current *then*. Upgrading the KFP platform (e.g. patching a CVE in the launcher image) has no effect on already-created recurring runs — they keep running the old images indefinitely unless the `ScheduledWorkflow`/recurring run is deleted and recreated. In practice this means fleets of recurring runs silently drift out of sync with the platform's current image versions and accumulate unpatched vulnerabilities.

## Where this happens (master branch)

- `backend/src/apiserver/resource/resource_manager.go`, recurring run creation path:
  ```go
  // If the pipeline version or pipeline spec is provided, this means the user wants to pin to a specific pipeline.
  // Otherwise, always let the ScheduledWorkflow controller pick the latest.
  if job.PipelineVersionId != "" || job.PipelineSpecManifest != "" || job.WorkflowSpecManifest != "" {
      // resolves template now and stores the compiled manifest on the Job / ScheduledWorkflow
  }
  ```
  A non-empty `PipelineVersionId` — even one referencing a normal, stored pipeline version rather than a raw uploaded manifest — takes the same "resolve and embed now" branch as legacy freeform manifest submission.

- `backend/src/apiserver/resource/resource_manager.go`, `ReconcileSwfCrs`:
  ```go
  if jobs[i].PipelineSpec.PipelineSpecManifest == "" && jobs[i].PipelineSpec.WorkflowSpecManifest == "" {
      continue // not pinned via manifest -> reference-based path used instead
  }
  ```
  Only jobs with **both** manifest fields empty skip embedding and rely on the reference-based path.

- `backend/src/crd/controller/scheduledworkflow/controller.go`, `submitNewWorkflowIfNotAlreadySubmitted`: already supports calling `c.runClient.CreateRun()` with a `PipelineVersionReference{PipelineId, PipelineVersionId}` where `PipelineVersionId` "can be empty, which causes the latest pipeline version to be selected" — this is the exact mechanism that should also be used for a *non-empty*, pinned `PipelineVersionId`.

## Reproduction

1. Deploy KFP 2.x with a V2 pipeline uploaded (multiple versions).
2. Create a recurring run, explicitly selecting a specific pipeline version in the UI (not "use latest").
3. Inspect the resulting `ScheduledWorkflow` CR: `kubectl get scheduledworkflow <name> -o yaml`.
4. Observe that `spec.workflow.spec` contains a fully compiled Argo `WorkflowSpec`, including specific launcher/driver image references, rather than just a `pipelineId`/`pipelineVersionId` reference.
5. Upgrade the KFP deployment (new launcher/driver images). Trigger the recurring run again and observe it still uses the old, pre-upgrade images.

Note: this reproduces even with `BLOCK_V1_PIPELINES=true`, since that flag only blocks V1-*templated* workflows from executing (`shouldEnforceV1Block` inspects the compiled workflow for V2 markers) — it does not change which jobs get their manifest embedded vs. referenced.

## Proposed fix

Extend the existing reference-based path to cover pinned versions, not just "latest":

1. In the recurring run creation path (`resource_manager.go`), only resolve-and-embed the manifest when a raw manifest is actually supplied (true legacy/freeform submission). When the request references a stored pipeline version by ID (pinned or not), skip resolution and persist `PipelineId`/`PipelineVersionId` on the `Job` only.
2. No `ScheduledWorkflow` CRD schema change is required — `spec.pipelineId`/`spec.pipelineVersionId` already exist and are already plumbed through to `CreateRun`'s `PipelineVersionReference` in the controller.
3. `ReconcileSwfCrs`'s existing skip-if-no-manifest check requires no change; it will now also correctly skip pinned-but-referenced jobs.
4. No controller change needed — `submitNewWorkflowIfNotAlreadySubmitted` already forwards a non-empty `PipelineVersionId` correctly; it just currently never receives one for pinned jobs because the manifest already exists earlier in the flow.

Net effect: "pin to a specific pipeline version" becomes "always run version X, compiled fresh from the stored version at trigger time" rather than "snapshot version X's compiled output — including images — forever." This removes the CRD/pipeline-version duplication and the image lock-in problem, while preserving the reproducibility guarantee (same pipeline logic) that pinning is meant to provide.

## Open questions / things to validate

- Per-trigger template compilation cost for high-frequency pinned schedules (the "latest" path already does this today, so likely acceptable, but worth confirming there's no caching assumption this would violate).
- Backward-compatibility/migration story for existing `ScheduledWorkflow` CRs that already have a manifest embedded.
- Whether this should be opt-in behind a flag initially, given it's a behavior change for a very common recurring-run pattern.

## Environment

- KFP backend: master branch (also reproduces on 2.x releases with `BLOCK_V1_PIPELINES=true`)
- Confirmed via direct testing: pinning a version through the recurring-run UI bakes the compiled workflow spec into the `ScheduledWorkflow` CR.

demarna1:

Fixes #13933.

When a recurring run is pinned to a specific pipeline version (the normal UI flow of selecting a version), `CreateJob` resolved the template at creation time and embedded the compiled Argo workflow into the `ScheduledWorkflow` CR (`spec.workflow.spec`) and the stored recurring run (`PipelineSpecManifest`/`WorkflowSpecManifest`). This duplicated the durably stored pipeline version and baked in whatever launcher/driver image references were current at creation time — platform upgrades (e.g. patching a CVE in the launcher image) never reached already-created recurring runs.

This PR routes pinned **V2** pipeline versions onto the reference-based path that already exists for the "always use latest" case (#13440):

- `CreateJob` now stores only `pipelineId`/`pipelineVersionId` on the `ScheduledWorkflow` (plus runtime parameters and pipeline root) and persists the recurring run without a manifest. Job inputs are still validated up front via `ValidateJobInputs`.
- At trigger time, the ScheduledWorkflow controller's existing reference branch calls `RunService.CreateRun` with a `PipelineVersionReference` carrying the pinned `PipelineVersionId`, so the run is compiled fresh from the stored pipeline version with the platform's current images. No controller changes were needed — this branch already forwards a non-empty pinned version id.
- The duplicated reference-SWF construction is extracted into a `referenceScheduledWorkflow` helper shared with the "latest" branch, and the manifest-persistence gate is now an explicit `embedManifest` flag instead of relying on `tmpl` nil-ness.

Behavior preserved:

- **Raw manifest submissions** (`PipelineSpecManifest`/`WorkflowSpecManifest` supplied) keep embedding, as before.
- **V1 (Argo) templates** keep embedding, including the v1beta1 API's pin-latest-at-creation shim.
- **Existing recurring runs** are unaffected: they keep their persisted manifests and continue to be reconciled by `ReconcileSwfCrs`; new pinned jobs are skipped by its existing empty-manifest check, same as latest-path jobs.
- Like the latest path, the recurring run's service account is no longer defaulted from the compiled spec at creation time; the default is applied when the run is compiled at trigger time.

Per-trigger compilation cost matches what the "latest" path already incurs today.

## Testing

- New `TestCreateJob_ThroughPipelineVersionV2`: pinned V2 version → stored job has no manifest, SWF carries `pipelineVersionId` with `spec.workflow.spec` absent and runtime parameters present.
- New `TestCreateJob_ThroughPipelineVersionV2_InvalidParams`: extra runtime parameter is rejected at creation time.
- New `TestSubmitNewWorkflowIfNotAlreadySubmitted_PipelineVersionReference`: closes an existing coverage gap on the controller's reference branch — asserts no Argo workflow is created directly and the `CreateRunRequest` carries the pinned `PipelineVersionReference`, recurring run id, and translated runtime config.
- Existing pinned-V1 and raw-manifest tests (`TestCreateJob_ThroughPipelineVersion`, `TestCreateJob_ThroughPipelineIdAndPipelineVersion`, `TestCreateJob_ThroughWorkflowSpec*`, `TestReconcileSwfCrs`, v1 `job_server` tests) pass unmodified.
- Full backend unit sweep (per repo guide, excluding integration/compiler/E2E suites) passes.

## Checklist

- [x] The title of your PR is conformant to the [contribution guidelines](https://github.com/kubeflow/pipelines/blob/master/CONTRIBUTING.md#pull-request-title-convention)

## Test Evidence

**Before**: ScheduledWorkflow spec with argo workflow baked in:

Screenshot 2026-08-20 at 9 22 07 AM

**After**: Now works the same as "always run latest" except it contains a pinned pipeline version like above:

Screenshot 2026-08-20 at 2 56 59 PM

Successful scheduled run:

Screenshot 2026-08-20 at 2 57 14 PM

### DIFF
--- a/backend/src/apiserver/resource/resource_manager.go
+++ b/backend/src/apiserver/resource/resource_manager.go
@@ -1348,11 +1348,21 @@ func (r *ResourceManager) CreateJob(ctx context.Context, job *model.Job) (*model
 	var manifest string
 	var scheduledWorkflow *scheduledworkflow.ScheduledWorkflow
 	var tmpl template.Template
+	// Only jobs that embed a compiled workflow into the ScheduledWorkflow persist their
+	// manifest in the recurring run; reference-based jobs are resolved from the stored
+	// pipeline version by the ScheduledWorkflow controller through the CreateRun API at
+	// trigger time.
+	embedManifest := false
 
 	// If the pipeline version or pipeline spec is provided, this means the user wants to pin to a specific pipeline.
 	// Otherwise, always let the ScheduledWorkflow controller pick the latest.
 	if job.PipelineVersionId != "" || job.PipelineSpecManifest != "" || job.WorkflowSpecManifest != "" {
 		var err error
+		// A raw manifest means a legacy/freeform submission that must keep embedding the
+		// compiled workflow. Capture this before fetchTemplateFromPipelineSpec, which
+		// back-fills pipeline reference fields on job.PipelineSpec.
+		hasRawManifest := job.PipelineSpecManifest != "" || job.WorkflowSpecManifest != ""
+
 		// Create a template based on the manifest of an existing pipeline version or used-provided manifest.
 		// Update the job.PipelineSpec if an existing pipeline version is used.
 		tmpl, manifest, err = r.fetchTemplateFromPipelineSpec(&job.PipelineSpec)
@@ -1360,19 +1370,34 @@ func (r *ResourceManager) CreateJob(ctx context.Context, job *model.Job) (*model
 			return nil, util.NewInternalServerError(err, "Failed to create a recurring run with an invalid pipeline spec manifest")
 		}
 
-		// When plugins are enabled, the SWF controller must call the CreateRun API
-		// so that per-run plugin logic executes.
-		if r.pluginDispatcher.PluginsRegistered() {
-			// Plugin-enabled: create a lightweight SWF without inline workflow spec
-			// so the SWF controller calls the CreateRun API for per-run plugin logic.
-			scheduledWorkflow, err = template.NewGenericScheduledWorkflow(job)
+		if v2Tmpl, ok := tmpl.(*template.V2Spec); ok && !hasRawManifest {
+			// Pinned to a specific V2 pipeline version by reference: validate the inputs
+			// up front, then store only the pipeline (version) reference in the
+			// ScheduledWorkflow so the controller creates runs through the CreateRun API
+			// at trigger time, compiling the stored pipeline version fresh on each run.
+			if err := v2Tmpl.ValidateJobInputs(job); err != nil {
+				return nil, util.Wrap(err, "Failed to validate the input parameters on the pinned pipeline version")
+			}
+			scheduledWorkflow, err = referenceScheduledWorkflow(job)
+			if err != nil {
+				return nil, err
+			}
 		} else {
-			// TODO(gkcalat): consider changing the flow. Other resource UUIDs are assigned by their respective stores (DB).
-			// Convert modelJob into scheduledWorkflow.
-			scheduledWorkflow, err = tmpl.ScheduledWorkflow(job)
-		}
-		if err != nil {
-			return nil, util.Wrap(err, "Failed to create a recurring run during scheduled workflow creation")
+			embedManifest = true
+			// When plugins are enabled, the SWF controller must call the CreateRun API
+			// so that per-run plugin logic executes.
+			if r.pluginDispatcher.PluginsRegistered() {
+				// Plugin-enabled: create a lightweight SWF without inline workflow spec
+				// so the SWF controller calls the CreateRun API for per-run plugin logic.
+				scheduledWorkflow, err = template.NewGenericScheduledWorkflow(job)
+			} else {
+				// TODO(gkcalat): consider changing the flow. Other resource UUIDs are assigned by their respective stores (DB).
+				// Convert modelJob into scheduledWorkflow.
+				scheduledWorkflow, err = tmpl.ScheduledWorkflow(job)
+			}
+			if err != nil {
+				return nil, util.Wrap(err, "Failed to create a recurring run during scheduled workflow creation")
+			}
 		}
 	} else if job.PipelineId == "" {
 		return nil, errors.New("Cannot create a job with an empty pipeline ID")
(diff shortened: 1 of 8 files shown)


droctothorpe:
Hi @demarna1, thank you for your PR! The linked issue (#13933) must be labeled `ready` before your contribution can be reviewed. Please check our [CONTRIBUTING.md](https://github.com/kubeflow/pipelines/blob/master/CONTRIBUTING.md) for details.

PR #13942, opened 2026-07-31
feat(backend): add kubeflow-pipelines tag to the S3 user agent

### LINKED ISSUE #13941: [feature] Append a kubeflow-pipelines tag to the API server S3 user agent

The API server's S3 client, built in `backend/src/apiserver/client_manager/client_manager.go` via `s3.NewFromConfig`, sends only the default `aws-sdk-go-v2/...` user agent. I would like it to append a `kubeflow-pipelines/<tag>` key so that object store operators can attribute the API server's S3 traffic to Kubeflow Pipelines.

The AWS SDK for Go v2 has a first-class hook for this, `middleware.AddUserAgentKeyValue`, which appends to the SDK user agent rather than replacing it. The version can come from the `TAG_NAME` config the API server already reads, with a fallback when it is unset.

When a cluster's artifact store is backed by an S3 API implementation and something goes wrong (unexpected `HeadBucket` or `CreateBucket` volume, throttling, a permissions question), the storage side cannot tell which workload is responsible. Every request looks like generic aws-sdk-go traffic. Attributing it requires correlating pod IPs against access logs, which does not survive NAT and is awkward in a multi-tenant cluster.

This is the same reasoning Kubeflow Pipelines already applies to its Kubernetes clients, which are tagged with a Pipelines-identifying user agent in several places. The S3 client is the one that is not.

Not from inside Pipelines. An operator can correlate source IPs against object store access logs, or put a proxy in front of the store that injects a header, but neither is practical in a normal cluster.

### Scope

I have a working branch that adds three lines to the client construction plus a test that asserts the tag is appended and the stock `aws-sdk-go-v2/` user agent is still present. It covers the API server client only; the driver/launcher client in `backend/src/v2/objectstore/object_store.go` would be a natural follow-up if the approach is accepted.

goanpeca:

Fixes #13941

The API server reaches its object store through an AWS SDK for Go v2 S3 client built in `newS3BucketClient` (`backend/src/apiserver/client_manager/client_manager.go`), so every request it sends carries only the SDK's default user agent, for example `aws-sdk-go-v2/1.43.1 os/macos lang/go#1.26.5 ...`. That endpoint is configurable through `OBJECTSTORECONFIG_HOST` and `OBJECTSTORECONFIG_PORT`, so the store behind it can be Amazon S3 or any S3-compatible implementation (Backblaze B2, Cloudflare R2, MinIO, SeaweedFS). Operators of those stores currently cannot separate KFP traffic from any other Go SDK client hitting the same bucket, which makes it awkward to attribute load, scope a rate limit, or line up an error spike with a KFP rollout.

This appends one key/value to the existing user agent instead of replacing it, so the outgoing header becomes `aws-sdk-go-v2/1.43.1 ... kubeflow-pipelines/2.17.0`.

The version comes from the `TAG_NAME` config that `backend/Dockerfile` already sets and that `/healthz` already reports, read through `common.GetStringConfigWithDefault` so an unset value degrades to `kubeflow-pipelines/unknown` rather than failing client construction. No new dependency is needed: `aws/middleware` ships inside `github.com/aws/aws-sdk-go-v2`, which is already a direct requirement in `go.mod`.

**Deliberately not changed:** the driver and launcher client in `backend/src/v2/objectstore`. Those images do not receive `TAG_NAME` today, so the tag would always read `unknown` there. Wiring a version into the driver and launcher images is a separate change and belongs in its own PR.

**Tests**

`TestNewS3BucketClient_AppendsUserAgent` drives a real request through the existing `fakeS3HTTPServer` and asserts the outgoing `User-Agent` still contains `aws-sdk-go-v2/`, proving nothing was replaced, and now also contains the `kubeflow-pipelines/` tag. The fake gained a `userAgents` recorder next to its existing request log.

```
$ go test ./backend/src/apiserver/client_manager/...
ok  	github.com/kubeflow/pipelines/backend/src/apiserver/client_manager	0.867s

$ golangci-lint run --new --new-from-rev=origin/master ./backend/src/apiserver/client_manager/...
0 issues.
```

Formatting verified with `gofmt -l` and `goimports -l` on the changed package, both clean.

**Checklist:**

### DIFF
--- a/backend/src/apiserver/client_manager/client_manager.go
+++ b/backend/src/apiserver/client_manager/client_manager.go
@@ -24,6 +24,7 @@ import (
 	"time"
 
 	awsv2 "github.com/aws/aws-sdk-go-v2/aws"
+	awsv2middleware "github.com/aws/aws-sdk-go-v2/aws/middleware"
 	awsv2cfg "github.com/aws/aws-sdk-go-v2/config"
 	awsv2creds "github.com/aws/aws-sdk-go-v2/credentials"
 	"github.com/aws/aws-sdk-go-v2/service/s3"
@@ -79,6 +80,9 @@ const (
 
 	clientQPS   = "ClientQPS"
 	clientBurst = "ClientBurst"
+
+	tagName      = "TAG_NAME"
+	userAgentKey = "kubeflow-pipelines"
 )
 
 var scheme *runtime.Scheme
@@ -1028,6 +1032,9 @@ func newS3BucketClient(ctx context.Context, config *blobStorageConfig) (*s3.Clie
 	s3Client := s3.NewFromConfig(awsCfg, func(o *s3.Options) {
 		o.BaseEndpoint = awsv2.String(endpointWithProtocol)
 		o.UsePathStyle = true
+		// Append to the SDK user agent so object store operators can attribute requests.
+		o.APIOptions = append(o.APIOptions, awsv2middleware.AddUserAgentKeyValue(
+			userAgentKey, common.GetStringConfigWithDefault(tagName, "unknown")))
 	})
 
 	return s3Client, nil
--- a/backend/src/apiserver/client_manager/client_manager_test.go
+++ b/backend/src/apiserver/client_manager/client_manager_test.go
@@ -302,10 +302,11 @@ func TestBuildCreateBucketInput_DefaultRegion(t *testing.T) {
 }
 
 type fakeS3HTTPServer struct {
-	t       *testing.T
-	server  *httptest.Server
-	buckets map[string]bool
-	log     []string
+	t          *testing.T
+	server     *httptest.Server
+	buckets    map[string]bool
+	log        []string
+	userAgents []string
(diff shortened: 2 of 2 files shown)


droctothorpe:
Thanks for the contribution — this is a useful observability improvement.

One request before we move forward: could you please put this behind a feature flag controlled by an environment variable, with the feature **off by default**?

My main concern is backwards compatibility for existing users with strict proxies, gateways, or S3-compatible systems that may be doing brittle matching on the current `User-Agent`. I think the change is probably fine in most environments, but making it opt-in initially would give operators a safe rollout path.

If you update the PR in that direction, I’m happy to take another look.

droctothorpe:
Hi @goanpeca, thank you for your PR! The linked issue (#13941) must be labeled `ready` before your contribution can be reviewed. Please check our [CONTRIBUTING.md](https://github.com/kubeflow/pipelines/blob/master/CONTRIBUTING.md) for details.

PR #13945, opened 2026-07-31
feat(backend): require explicit MySQL tls in extra params

### LINKED ISSUE #13796: [feature]  MySQL driver lacks symmetric secure-by-default TLS handling (parity with PostgreSQL sslmode)
## Background

PR #12379  made the PostgreSQL client secure-by-default: `CreatePostgreSQLConfig`
now requires `sslmode` to be set explicitly and fails fast (CrashLoop) if it is
missing, instead of silently defaulting to `sslmode=disable`. This prevents
production deployments from unknowingly sending database traffic unencrypted.
The base `platform-agnostic-postgresql` overlay intentionally leaves `sslmode`
unset; dev/CI overlays opt into `sslmode=disable` explicitly.

## Problem

The MySQL path was intentionally left out of that change to keep the PR scoped
and avoid breaking existing MySQL deployments. As a result the two drivers are
asymmetric:

- `CreateMySQLConfig` (backend/src/apiserver/client/sql.go and
  backend/src/cache/client/sql.go) sets no `tls` parameter at all — it relies on
  the go-sql-driver default, which is an unencrypted connection, with no explicit
  signal in code or manifests.
- There is no fail-fast: a MySQL deployment that intends TLS but misconfigures it
  still starts and silently runs in plaintext.

## Proposed follow-up

Bring MySQL to parity with PostgreSQL:
1. Decide the MySQL equivalent policy (the go-sql-driver `tls` parameter:
   `false` / `true` / `skip-verify` / a registered custom TLS config).
2. Consider requiring `tls` to be set explicitly (fail-fast) the same way
   `sslmode` is now required for PostgreSQL, OR at minimum emit an explicit
   startup warning when running unencrypted.
3. Make dev/CI overlays opt into the insecure value explicitly, keeping the base
   overlay secure-by-default.
4. Document the production TLS setup for MySQL, mirroring the PostgreSQL
   `verify-full` guidance in backend/README.md.

## Notes

This is deliberately a separate issue because a symmetric MySQL change is a
breaking change for existing MySQL deployments and needs its own migration
consideration.

sahilleth:
## Summary
- `CreateMySQLConfig` in apiserver and cache now returns an error unless `tls` is explicitly set in `ExtraParams`
- Callers propagate the error at startup instead of silently connecting without TLS
- CI Kind overlays set `tls=false` for local MySQL
- Document the requirement in `backend/README.md`

Fixes #13796

## Test plan
- [x] `go test ./backend/src/apiserver/client/... ./backend/src/cache/client/...`
- [ ] CI integration tests with updated manifests


hbelmiro:
This should be synced with @kaikaila.

kaikaila:
Hi @sahilleth , thanks for picking up #13796 — this is exactly the MySQL TLS parity that was intentionally left out of #12379 to keep scope manageable.

Heads-up on a merge-order concern: #12379 deletes both backend/src/apiserver/client/sql.go and backend/src/cache/client/sql.go entirely, consolidating the MySQL/PostgreSQL config logic into a shared backend/src/apiserver/common/sql/config.go. It also heavily refactors both client_manager.go files that this PR touches.

If this PR merges first, #12379 would need to carry forward your TLS validation into the new shared location — doable but adds friction. If #12379 merges first, the change you need here becomes very small: just add the tls presence check to the single CreateMySQLConfig in common/sql/config.go, plus the CI overlay and doc updates (which have no overlap).

Given that, I think the cleaner path is to land #12379 first and then rebase this PR on top. Would it work for you?

hbelmiro:
/hold

Due to https://github.com/kubeflow/pipelines/pull/13945#issuecomment-5198072640

droctothorpe:
Hi @sahilleth, thank you for your PR! The linked issue (#13796) must be labeled `ready` before your contribution can be reviewed. Please check our [CONTRIBUTING.md](https://github.com/kubeflow/pipelines/blob/master/CONTRIBUTING.md) for details.

PR #13947, opened 2026-07-31
fix(backend): close MLflow nested runs at task-level instead of a single apiserver-side sweep

VaniHaripriya:

Moves MLflow nested-run cleanup from a single recursive apiserver-side sweep at run-end to a distributed model: each task now closes its own open direct children in `OnTaskEnd`, while the apiserver keeps a bounded recursive sweep only as a safety net for orphaned runs.

**Changes:**

- Add `CloseOpenChildRuns` to `backend/src/common/plugins/mlflow`, a shared,  single-level "find and force-close direct open children of a run"  primitive usable by both the apiserver and the v2 driver/launcher.
- v2 driver/launcher (`OnTaskEnd`): after closing its own MLflow run, a task now also closes any of its own open direct children via `CloseOpenChildRuns`.
- Apiserver (`syncNestedRuns`): split into `closeNestedRuns` (terminal mode, now a thin recursive safety net built on the shared helper) and  `reopenNestedRuns` (retry mode, unchanged retry-specific logic for reopening FAILED/KILLED runs).

**Checklist:**

### DIFF
--- a/backend/src/apiserver/plugins/mlflow/util.go
+++ b/backend/src/apiserver/plugins/mlflow/util.go
@@ -175,11 +175,11 @@ func FailedPluginOutput(experimentID, experimentName, runID, runURL, stateMessag
 	return buildPluginOutput(experimentID, experimentName, runID, runURL, apiv2beta1.PluginState_PLUGIN_FAILED, stateMessage)
 }
 
-// maxSearchPages caps SearchRuns pagination to prevent infinite loops.
-const maxSearchPages = 100
+// maxSearchPages caps SearchRuns pagination in reopenNestedRuns (retry path).
+const maxSearchPages = 10
 
 // maxNestingDepth caps recursive nested run traversal.
-const maxNestingDepth = 4
+const maxNestingDepth = 6
 
 func SyncParentAndNestedRuns(ctx context.Context, requestCtx *commonmlflow.RequestContext, parentRunID, experimentID string, mode apiserverPlugins.RunSyncMode, terminalStatus string, endTimeMs *int64) []string {
 	if requestCtx == nil || requestCtx.Client == nil {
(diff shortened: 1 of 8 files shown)


backend/src/common/plugins/mlflow/util.go
```diff
@@ -0,0 +1,91 @@
...
+			}
+			childRunID := run.Info.RunID
+			if childRunID == "" {
+				childRunID = run.Info.RunUUID
+			}
+			if childRunID == "" || childRunID == parentRunID || !IsOpenRunStatus(run.Info.Status) {
+				continue
+			}
+			if beforeClose != nil {
+				beforeClose(childRunID)
```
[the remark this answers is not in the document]
VaniHaripriya: Addressed this.

backend/src/common/plugins/mlflow/util.go
```diff
@@ -0,0 +1,100 @@
...
+				continue
+			}
+			childRunID := run.Info.RunID
+			if childRunID == "" {
+				childRunID = run.Info.RunUUID
+			}
+			if childRunID == "" || childRunID == parentRunID {
+				continue
+			}
+			if beforeClose != nil {
```
mprahl, line 78: Should this be after `if !IsOpenRunStatus(run.Info.Status) {`?

backend/src/common/plugins/mlflow/util.go
```diff
@@ -0,0 +1,100 @@
...
+		return false
+	default:
+		return true
+	}
+}
+
+// CloseOpenChildRuns closes open direct children of parentRunID. beforeClose
+// runs for every valid child (including terminal ones) to reach orphaned
+// descendants; UpdateRun is called only for still-open children.
+func CloseOpenChildRuns(ctx context.Context, client *Client, experimentID, parentRunID, targetStatus string, endTimeMs *int64, beforeClose func(childRunID string)) []string {
```
mprahl, line 48: Instead of including `beforeClose` in the function signature, could `CloseOpenChildRuns` just return a slice of errors or a single error that has all of them wrapped? I think that'd be cleaner since I don't see another use for `beforeClose`.

mprahl (changes requested):
A couple small comments but otherwise LGTM!

backend/src/common/plugins/mlflow/util.go
VaniHaripriya: This was added before  `if !IsOpenRunStatus(run.Info.Status) {`  as a safety net, if we only called it for open children, we’d skip terminal direct children and miss open descendants left behind when a task’s own cleanup failed.

backend/src/apiserver/plugins/mlflow/util.go
```diff
@@ -175,11 +175,11 @@ func FailedPluginOutput(experimentID, experimentName, runID, runURL, stateMessag
...
 }
 
-// maxSearchPages caps SearchRuns pagination to prevent infinite loops.
-const maxSearchPages = 100
+// maxSearchPages caps SearchRuns pagination in reopenNestedRuns (retry path).
+const maxSearchPages = 10
 
 // maxNestingDepth caps recursive nested run traversal.
-const maxNestingDepth = 4
+const maxNestingDepth = 6
```
alyssacgoins, line 182: What is the significance of the change here form `4` to specifically `6`?

backend/src/common/plugins/mlflow/util.go
VaniHaripriya: Actually, `beforeClose` was added so that the API server can recursively traverse each direct child, process its entire subtree before closing that child, since CloseOpenChildRuns only handles one level of children.

backend/src/apiserver/plugins/mlflow/util.go
VaniHaripriya: Good question , there isn’t a hard requirement for exactly 6.

The original value of 4 was added to cover the expected MLflow hierarchy (parent → task/loop → iteration). I bumped it to 6 , to give the apiserver more headroom for deeper trees  e.g. nested pipelines where inner tasks inherit a new parentRunId, plus a loop/iteration level underneath.
Each task already closes its own MLflow run when it finishes. This constant only limits how deep the apiserver goes when it does a final pass to catch any runs that tasks missed.

alyssacgoins: Got it - the increase makes sense to me. (Adding an  /lgtm, so this is more of a general question) Is there a case where we would need to bump this number higher?

alyssacgoins (approved):
/lgtm

mprahl (approved):
/approve

PR #10797, opened 2024-05-07
fix(frontend): reduce list run latency

droctothorpe:
**Description of your changes:**
This PR updates the list runs page to reduce the number of API calls to the experiments API by retrieving all experiments with one request instead of making a discrete request for each run. It also includes a similar update for the list recurring runs page.

Standing by for design review before proceeding with test updates.

Additional information available in this issue: https://github.com/kubeflow/pipelines/issues/10778.

Shout out to @quinnovator and @tarat44 for collaborating on it. 

**Checklist:**
- [x] The title for your pull request (PR) should follow our title convention. [Learn more about the pull request title convention used in this repository](https://github.com/kubeflow/pipelines/blob/master/CONTRIBUTING.md#pull-request-title-convention).

PR #13951, opened 2026-08-02
fix(backend, frontend): skip the unused run count query on the runs list page

### LINKED ISSUE #11346: [frontend] Kubeflow does N queries for in the "Runs" page
### Environment

Kubeflow pipelines standalone, AWS setup

### Steps to reproduce
Open your network panel in dev tools.  Navigate to the Runs page and look at your network console. Try to increase page size in the UI, and see many requests.

When navigating to the "Runs" page, kubeflow will send an API call to fetch a list of runs. There are two problems here:
1) This query returns a `total_size` field, which does unneeded COUNT(*) query on the whole "run_details" table, the count is not even displayed in the UI.
2) More important, after the run list is fetched, the UI will do an **API call per run** (which is a DB query) to get its associated pipeline.  This is very slow. Instead, the runs API could simply do an SQL JOIN to get the pipeline info.
This is really slow, even for page size of 10, tested with my medium size RDS instance. When page size is 100, this page will do **over 100 SQL queries**.

image

### Expected result
To get the runs data with the pipeline info, one query should be enough. Page should load faster.
### Materials and Reference

---

Sharkyii:

Two independent causes were bundled into this one bug report. This PR resolves the first and partially improves the second.

Backend: `ListRuns` always ran a second, unconditional query just to compute `total_size`, even though the frontend never reads it. Adds an optional `skip_count` field to `ListRunsRequest` (default `false`, so every existing caller and the `total_size` integration tests keep their exact current behavior) that lets a caller skip that extra query. The KFP frontend is the one caller that actually sets it, since it never uses the value.

Frontend: `RunList.tsx` and `ExperimentList.tsx` pass `skip_count` on the run-list requests that never display a total. `RunList.tsx` already batched experiment lookups into one `listExperiments` call (from #10797); it now also deduplicates pipeline-version lookups, so for R displayed runs referencing U unique `(pipelineId, pipelineVersionId)` pairs it makes U requests instead of R. Failed lookups keep their original 403/404/network detail instead of being replaced with a generic message.

**What this PR does not accomplish:**

- It does not eliminate the general pipeline-version N+1 lookup pattern. If every displayed run references a different version then U = R, and the number of version requests is unchanged.
- It does not implement the batch/join solution requested in #11346, so that issue should remain open.

An earlier revision batched lookups per pipeline via `listPipelineVersions` with an `IN` filter. That was reverted: `GetPipelineVersion` authorizes `get` with the version ID as the resource name, while `ListPipelineVersions` authorizes `list` against the namespace, and Kubernetes `resourceNames` cannot scope a `list`. Users holding resource-scoped `get` would have seen 403s on every row. `ListPipelineVersions` also issues an unconditional `count(*)` alongside the rows query, making the worst case (N runs from N distinct pipelines) more expensive than the point gets it replaced.

Related: #12827 adds experiment/pipeline display names directly onto the Run proto via backend joins, a more complete fix for the pipeline lookup half of this, but it's stalled (never tested, stale since June) and doesn't touch the total_size half at all. This doesn't duplicate that work, and isn't blocked by it landing.

Refs #11346

**Checklist:**


Sharkyii:
@zazulam, please have a look when you’re free. You mentioned in the meeting that you’d take a look at this.

frontend/src/pages/RunList.tsx
```diff
@@ -498,16 +502,70 @@ class RunList extends React.PureComponent<RunListProps, RunListState> {
...
+    await Promise.all(
+      uniquePipelineIds.map(async (pipelineId) => {
+        try {
+          const response = await Apis.pipelineServiceApiV2.listPipelineVersions(
+            pipelineId,
+            undefined,
+            // Large enough to cover the vast majority of pipelines in one call. Runs whose
+            // version isn't found here (a pipeline with more versions than this covers) fall
+            // back to a direct per-run lookup in _setPipelineVersionName.
+            1000,
```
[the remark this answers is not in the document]
Sharkyii: Good catch, will fetch only the exact referenced version IDs instead of listing up to 1000.

frontend/src/components/ExperimentList.test.tsx
```diff
@@ -254,6 +254,7 @@ describe('ExperimentList', () => {
           ],
         } as V2beta1Filter),
       ),
+      true,
```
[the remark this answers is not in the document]
Sharkyii: This is a false positive. `ExperimentList.tsx` just renders `RunList`, and `RunList` handles the fetch. Since `_loadRuns` always passes `true` now, the spy gets `true` here as expected. No production change is needed.

frontend/src/pages/RunList.tsx
```diff
@@ -499,30 +507,72 @@ class RunList extends React.PureComponent<RunListProps, RunListState> {
...
+        });
+      }
+    });
+
+    const pipelineVersionsByKey = new Map<string, V2beta1PipelineVersion>();
+    await Promise.all(
+      Array.from(uniqueVersionRefs.entries()).map(
+        async ([key, { pipelineId, pipelineVersionId }]) => {
+          try {
+            const pipelineVersion = await Apis.pipelineServiceApiV2.getPipelineVersion(
```
jeffspahr, line 538: [P2] This still leaves the reported N+1 path for pages whose runs reference different versions. `Promise.all` issues one `getPipelineVersion` request per distinct `(pipelineId, pipelineVersionId)`, so 100 distinct references still produce 100 API calls and DB queries; the added test only covers two runs sharing the same pair. Please use a lightweight batch/join for display names, or narrow the PR/issue claim to duplicate-request deduplication and leave #11346 open.

frontend/src/pages/RunList.tsx
```diff
@@ -394,6 +399,7 @@ class RunList extends React.PureComponent<RunListProps, RunListState> {
           request.pageSize,
           request.sortBy,
           request.filter,
+          /* skip_count */ true, // this page never displays the total run count
```
jeffspahr, line 402: [P2] Please also pass this flag from `frontend/src/pages/ExperimentList.tsx`. Its `_reload` issues one `listRuns` call per displayed experiment at lines 222-239, consumes only `response.runs`, and still omits `skip_count`, so every experiment row retains the expensive count query this flag is intended to avoid. Update the matching page test as well.

frontend/src/pages/RunList.tsx
```diff
@@ -499,30 +507,72 @@ class RunList extends React.PureComponent<RunListProps, RunListState> {
...
           pipelineId: pipelineVersion.pipeline_id,
           usePlaceholder: false,
           versionId: pipelineVersion.pipeline_version_id,
         };
-      } catch (err) {
-        displayRun.error =
-          'Failed to get associated pipeline version: ' + (await errorToMessage(err));
-        return;
+      } else {
+        displayRun.error = 'Failed to get associated pipeline version';
```
jeffspahr: [P2] Preserve the original lookup error instead of replacing every failure with this generic message. The catch above currently only logs `err`, whereas the previous implementation appended `await errorToMessage(err)` to the row error; users therefore lose useful 403/404/network details from the warning. Store failures by key alongside successful versions and use the keyed error here, with a rejection-path test.

jeffspahr (changes requested):
Requesting changes for three remaining performance and error-handling issues. The backend/API and generated-client changes otherwise look sound.

frontend/src/pages/RunList.tsx
Sharkyii: The dedup only helped when runs shared a version, which isn't the reported case. Now batching instead: version IDs are grouped by pipeline and fetched with one listPipelineVersions call per pipeline using an IN filter.
Checked in the browser with 20 runs across 20 distinct versions in 5 pipelines: 20 requests before, 5 after. Also replaced the test so it actually covers distinct versions, not just the shared-pair case.
Worth flagging: this is a frontend batch, not the SQL join #11346 asks for - that also wants experiment data joined in.

frontend/src/pages/RunList.tsx
Sharkyii: Added skip_count to the per-experiment listRuns call in ExperimentList._reload and updated the assertion in ExperimentList.test.tsx.

frontend/src/pages/RunList.tsx
Sharkyii: Fixed. Failures now go into an errorsByKey map alongside the successful lookups, so the row shows the real 403/404/network message instead of the generic one. If a batched request fails, every version ID it asked for gets that error. Added a rejection-path test.

frontend/src/pages/RunList.tsx
```diff
@@ -499,30 +508,98 @@ class RunList extends React.PureComponent<RunListProps, RunListState> {
...
+        const versionIds = versionIdsByPipelineId.get(pipelineId) || new Set<string>();
+        versionIds.add(pipelineVersionId);
+        versionIdsByPipelineId.set(pipelineId, versionIds);
+      }
+    });
+
+    const pipelineVersionsByKey = new Map<string, V2beta1PipelineVersion>();
+    const errorsByKey = new Map<string, string>();
+    await Promise.all(
+      Array.from(versionIdsByPipelineId.entries()).map(async ([pipelineId, versionIds]) => {
```
jeffspahr: [P2] This still leaves the worst-case N+1 path. The Promise.all creates one listPipelineVersions call per distinct pipeline, so N runs from N pipelines still make N HTTP requests. On the SQL store each list also executes both the version rows query and an unconditional count query, making this case more expensive than the previous point gets. Please use a lightweight cross-pipeline batch/join; otherwise preserve the point-get path for singleton pipelines, narrow the PR claim, and leave #11346 open.

frontend/src/pages/RunList.tsx
```diff
@@ -499,30 +508,98 @@ class RunList extends React.PureComponent<RunListProps, RunListState> {
...
+              predicates: [
+                {
+                  key: 'pipeline_version_id',
+                  operation: V2beta1PredicateOperation.IN,
+                  string_values: { values: requestedIds },
+                },
+              ],
+            } as V2beta1Filter),
+          );
+          const response = await Apis.pipelineServiceApiV2.listPipelineVersions(
```
jeffspahr: [P2] This changes the permission required to render the pipeline name from get to list. getPipelineVersion authorizes get against the referenced pipeline, while listPipelineVersions authorizes list for the namespace. Those verbs are independent, and resource-name-scoped roles can grant get without a usable list permission, so valid least-privilege users will now see 403 row errors for every referenced pipeline. Please preserve get-compatible authorization via a batch-by-ID API or fall back to keyed gets when the batch is forbidden, and add multi-user coverage.

frontend/src/pages/RunList.tsx
```diff
@@ -499,30 +508,98 @@ class RunList extends React.PureComponent<RunListProps, RunListState> {
...
+      if (pipelineVersion) {
         displayRun.pipelineVersion = {
           displayName: pipelineVersion.display_name,
           pipelineId: pipelineVersion.pipeline_id,
           usePlaceholder: false,
           versionId: pipelineVersion.pipeline_version_id,
         };
-      } catch (err) {
+      } else {
+        const errorMessage = errorsByKey.get(key);
```
jeffspahr, line 579: [P2] A missing or deleted version still loses the actionable 404 detail. listPipelineVersions legitimately returns 200 with an empty or partial result for requested IDs that no longer exist, but errorsByKey is populated only when the entire batch rejects. Those omitted IDs therefore reach this branch with no errorMessage and get only the generic text, unlike the previous getPipelineVersion path. Please fall back only for omitted IDs, as the PR description currently promises, or provide deterministic not-found detail, and cover a partial-response case.

jeffspahr (changes requested):
The ExperimentList skip_count path and whole-batch error propagation are fixed. Requesting changes for the remaining worst-case fan-out, multi-user RBAC regression, and missing-version error regression.

frontend/src/pages/RunList.tsx
Sharkyii: Agreed. The count(*) makes the worst-case fan-out even worse. I’ve reverted the batching and kept this PR focused on deduplicating lookups.

frontend/src/pages/RunList.tsx
Sharkyii: Reverted the batching. `get` and `list` have different RBAC behavior, so resource-scoped `get` users could hit 403s on every row.

A fallback to keyed gets would just patch around the issue, so I went back to point gets until we have a proper batch-by-ID API.

frontend/src/pages/RunList.tsx
Sharkyii: Also fixed by the revert. Point gets give us the actual 404 for missing versions instead of the generic error.

Sharkyii:
Latest changes-
* Added `skip_count` and updated the test.
* Preserved the actual 403/404/network errors per version.
* Reverted batching due to RBAC issues and the extra count query.

So this PR now just deduplicates repeated version requests. #11346 stays open for the proper batch/join fix.

jeffspahr (changes requested):
Re-review of `3f298fa46991`: the revert resolves the RBAC regression and restores the original 403/404/network details for failed version lookups. One scope issue remains:

**[P2] Update the PR description and issue-closing claim.** The current implementation calls `getPipelineVersion` once per unique `(pipelineId, pipelineVersionId)` pair. That usefully deduplicates repeated references, but `N` distinct versions still produce `N` requests. The description still claims the reverted per-pipeline batching and fallback behavior, and `Fixes: #11346` would close the general N+1 issue without resolving it.

Please rewrite the description to match the current implementation and remove the closing keyword (for example, use `Refs #11346`). The backend/API changes and generated clients otherwise look sound.

jeffspahr:
### What this PR accomplishes

- Adds an opt-in `skip_count` parameter and uses it in the frontend paths that do not consume `total_size`, avoiding an unused count query on those run-list requests.
- Deduplicates identical pipeline-version point lookups. For `R` displayed runs referencing `U` unique `(pipelineId, pipelineVersionId)` pairs, the frontend now makes `U` version requests instead of `R`.
- Retains resource-scoped `get` authorization and preserves useful 403/404/network error details.

### What this PR does not accomplish

- It does not eliminate the general pipeline-version N+1 lookup pattern.
- If every displayed run references a different version, then `U = R`, so the number of version requests is unchanged.
- It does not implement the batch/join solution requested by #11346, so that issue should remain open.

Overall, this is a useful incremental optimization with narrower scope than the current PR description states.

Sharkyii:
**Where things stand:** Experiment lookups are already batched into a single `listExperiments` call (#10797). #13951 adds `skip_count`, which removes the unnecessary `COUNT(*)` from `ListRuns`, and also dedupes identical version lookups. The problem is that if R runs reference U unique versions, we're still making U requests, so the N+1 issue remains.

**One approach I tried:** I tested batching per pipeline using `listPipelineVersions` with a `pipeline_version_id IN (...)` filter. It reduces the number of requests quite a bit, but I reverted it because of the authorization model. `GetPipelineVersion` authorizes `get` using the version ID as the resource name, while `ListPipelineVersions` uses `list` at the namespace level. Since Kubernetes doesn't support `resourceNames` for `list`, callers with only resource-scoped `get` permissions would get 403s for every row. `ListPipelineVersions` also does an unconditional `count(*)`, so in the worst case (N runs across N pipelines) it can actually be slower than the individual gets.

So whatever we do needs to remain compatible with the existing `get` authorization model.

**Proposed idea:** a batch-by-ID endpoint like `BatchGetPipelineVersions(pipeline_version_ids[])` that checks `get` permission for each requested ID and only returns versions the caller is allowed to see. This keeps the current permission model, avoids the count query, and lets the frontend control what it fetches.

**Idea:** a batch-by-ID endpoint - `BatchGetPipelineVersions(pipeline_version_ids[])` - that uses the existing `get` authorization for each requested ID and only returns versions the caller can access. This keeps the current permission model intact, avoids the count query, and lets the frontend control what it fetches.

jeffspahr (changes requested):
Re-review of `3f298fa46991`: the updated PR title and description now accurately describe the narrower optimization, and I found no remaining implementation issue. One merge-metadata issue remains:

**[P2] Rewrite the stale source commit message before merge.** Commit [`d6bb2254c5fa`](https://github.com/kubeflow/pipelines/commit/d6bb2254c5fa166c92f502b6177731a878d531f3) still starts with `Fixes #11346` and describes the reverted per-pipeline batching/fallback behavior. This repository is configured to build squash messages from `COMMIT_MESSAGES`, so that text is carried into the commit landing on `master`. [GitHub applies closing keywords in commit messages](https://docs.github.com/en/issues/tracking-your-work-with-issues/using-issues/linking-a-pull-request-to-an-issue) when they reach the default branch, even though the containing PR is not shown as linked.

Please rewrite/squash the branch so the landing commit omits `Fixes #11346` and the obsolete batching description. The current `Refs #11346` scope in the PR body is appropriate, and #11346 should remain open for the general batch/join solution.

Sharkyii:
Squashed the branch into one commit so there's no stale message left - same code, Refs #11346, f1c5abb9c. Earlier inline comments may show as outdated since they were anchored to commits that no longer exist.

jeffspahr (commented):
Re-review of `e7a255c14f16`: no remaining findings.

The history rewrite resolves the final merge-metadata concern. The scoped commit `f1c5abb9cda0` has the exact same tree as the previously reviewed head, and the current base-to-head patch has the same stable patch ID and 21-file scope. Its title/body accurately describe the count-query optimization and duplicate pipeline-version lookup deduplication, use `Refs #11346`, and contain no closing keyword or stale current-behavior batching claim. The subsequent `master` merge introduces no custom conflict resolution.

#11346 remains open with no closing reference. Fresh CI on this head has 205 passing checks and no failures, and the branch is current with `master`.

jeffspahr:
/approve
/lgtm

Nice work @Sharkyii! Thanks for the contribution!

Issue #13956, opened 2026-08-03
platform-agnostic-postgresql does not produce a working deployment: postgres statement builder unused, identifier casing, and four manifest bugs


hsinhoyeh:
## Summary

The `platform-agnostic-postgresql` overlay does not produce a working deployment on 2.16.1. `ml-pipeline` (api-server) never reaches Ready. I hit six distinct problems getting through it, two of which are in the storage layer and cannot be fixed by configuration.

**None of this is architecture-specific.** I found it on arm64, but every issue below is a placeholder-dialect, identifier-casing, or manifest problem that applies equally on x86_64. I am filing separately from #10309 so that ARM is not conflated with it.

Version: `2.16.1` (`67a73cf`), `manifests/kustomize/env/platform-agnostic-postgresql`, single-node kind cluster.

## Reproduce

```
kubectl apply -k manifests/kustomize/cluster-scoped-resources
kubectl apply -k manifests/kustomize/env/platform-agnostic-postgresql
```

`ml-pipeline` stays `0/1` and its startup probe never passes.

---

## 1. The PostgreSQL statement builder is defined but never used

`backend/src/apiserver/client_manager/dialect.go:46` correctly sets up dollar placeholders:

```go
StatementBuilder: sq.StatementBuilder.PlaceholderFormat(sq.Dollar),
```

But nothing consumes it. In `backend/src/apiserver/storage/` (excluding tests):

| | count |
|---|---|
| bare `sq.Select` / `sq.Insert` / `sq.Update` / `sq.Delete` call sites | **26** across 8 files |
| call sites using the dialect's `StatementBuilder` | **0** |

`grep -rn 'StatementBuilder' backend/src/apiserver --include=*.go | grep -v _test` returns only the four lines in `dialect.go` that define it.

Every query is therefore built with squirrel's default `sq.Question` format and sent to PostgreSQL with `?` placeholders. Because `?` is a valid **jsonb operator** in PostgreSQL, the parser does not reject the `?` itself — it fails on the following token, which makes this look like a malformed WHERE clause:

```
Failed to get a pipeline with name <name> and namespace kubeflow:
  ERROR: syntax error at or near "AND" (SQLSTATE 42601)
```

That comes from `GetPipelineByNameAndNamespace` (`storage/pipeline_store.go:160`), whose SQL is structurally fine — only the placeholder dialect is wrong.

## 2. gorm creates case-preserved columns that the queries cannot see

[...]

PR #13958, opened 2026-08-03
Fix configuration errors in the platform-agnostic-postgresql overlay

hsinhoyeh:
Fixes the manifest half of #13956 — the configuration bugs in the `platform-agnostic-postgresql` overlay.

**This does not make the overlay work on its own.** The two storage-layer problems in #13956 (the postgres `StatementBuilder` being defined but never used, and the 106 CamelCase `gorm:"column:…"` tags) still block `ml-pipeline` from reaching Ready. Those need code changes and a design decision that is not mine to make. What this PR does is remove the configuration errors that sit *in front* of them, so anyone working on the storage layer is not also fighting four unrelated bugs.

None of these are architecture-specific. I found them on arm64, but every one applies identically on x86_64.

## 1. `pg-secret.yaml` uses `data:` with a plaintext value

`third-party/postgresql/base/pg-secret.yaml` supplies the literal string `password` under `data:`, which requires base64. Kubernetes therefore base64-*decodes* `password` — which happens to be valid base64 — producing six non-UTF-8 bytes:

```
$ printf 'password' | base64 -d | xxd
00000000: a5ab 2cc2 8add                           ..,...
```

That is byte-for-byte what ends up in the cluster. containerd cannot marshal it into a container spec, so the postgres pod never starts, with an error that gives no hint of the cause:

```
CreateContainerError: grpc: error while marshaling: string field contains invalid UTF-8
```

It also explains why `postgres-secret.root_password` and `postgres-secret-extended.password` never agree: the latter correctly uses `stringData:`, so its value really is the string `password`, while postgres is seeded with the six garbage bytes.

Changed to `stringData:`, matching `postgres-secret-extended.yaml`.

## 2. `postgresHost` names a Service that does not exist

`base/installs/generic/postgres/pipeline-install-config.yaml` sets `postgresHost: postgres`, but the Service created by `third-party/postgresql/base/pg-service.yaml` is named `postgres-service`. The name never resolves:

```
$ kubectl get svc -n kubeflow postgres
Error from server (NotFound): services "postgres" not found

psql: error: could not translate host name "postgres" to address: Try again
```

Changed to `postgres-service`. `dbHost` is updated for consistency, though it is marked in-file as a relic.

## 3. `dbType: postgres` is not a value the server accepts

`initDBDriver` (`backend/src/apiserver/client_manager/client_manager.go:401`) accepts only `mysql` or `pgx`, and `glog.Fatalf`s on anything else. Changed to `pgx`.

## 4. `DB_DRIVER_NAME` never binds

`base/postgresql/pipeline/ml-pipeline-apiserver-deployment-patch.yaml` sets env `DB_DRIVER_NAME`, but `initConfig()` in `backend/src/apiserver/main.go` uses:

```go
viper.SetEnvKeyReplacer(strings.NewReplacer(".", "_"))
viper.AutomaticEnv()
```

so `common.GetStringConfig("DBDriverName")` resolves to env **`DBDRIVERNAME`**. `DB_DRIVER_NAME` is silently ignored and `config.json`'s `"DBDriverName": "mysql"` default wins — the api-server attempts MySQL on a PostgreSQL install. The visible symptom is:

```
config.go:75] Config DBConfig.MySQLConfig.ExtraParams not specified, skipping
```

on an install with no MySQL anywhere.

Renamed the env var to `DBDRIVERNAME`. (Renaming is the smaller change; `viper.BindEnv("DBDriverName", "DB_DRIVER_NAME")` in `initConfig` would be the alternative if you would rather keep the existing env var name as the public interface.)

## 5. `$patch: replace` drops the object-store and runtime-image configuration

`base/postgresql/pipeline/ml-pipeline-apiserver-deployment-patch.yaml` begins its env list with:

```yaml
        env:
        - $patch: replace
```

That discards the base list rather than merging it by `name`, so `OBJECTSTORECONFIG_HOST` and `OBJECTSTORECONFIG_PORT` — set in `base/pipeline/ml-pipeline-apiserver-deployment.yaml:50` — never reach the rendered deployment:

```
F main.go:160] failed to initialize object store:
  failed to build config from environment variables: ObjectStoreConfig.Host is required
```

It also discards the release-pinned `V2_DRIVER_IMAGE` and `V2_LAUNCHER_IMAGE`, so the compiler falls back to `ghcr.io/kubeflow/kfp-{driver,launcher}:latest` and a released PostgreSQL install can run mismatched runtime images, along with `PUBLISH_LOGS`, `CLUSTER_DOMAIN`, `COMPILED_PIPELINE_SPEC_PATCH` and the `DEFAULT_SECURITY_CONTEXT_*` set.

Removed `$patch: replace` from both the init container and the api-server container, and instead `$patch: delete` only the MySQL-specific entries — the `mysql-secret` and `mysqlHost`/`mysqlPort` references that cannot resolve on a PostgreSQL install and would block Pod startup. Everything else merges from the base and stays in sync automatically. This is the same approach #12379 takes for this deployment.

## Verification

Applied these values to a live `platform-agnostic-postgresql` install and confirmed each one clears its failure:

- postgres pod starts instead of `CreateContainerError`
- `postgres-service` resolves, and TCP password auth succeeds from another pod
- the api-server stops reading MySQL config and initialises the postgres client:
  ```
  I client_manager.go:570] Running AutoMigrate.
  I client_manager.go:279] DB client initialized successfully
  ```

It then fails on the storage-layer issues in #13956, as expected — which is why this PR is explicitly only half the fix.

`kustomize build` of `env/platform-agnostic-postgresql` and `env/dev/postgresql` confirms the rendered api-server keeps the base env list, carries the pinned V2 images, and no longer references `mysql-secret` or `DB_DRIVER_NAME`.

I have not run the conformance or integration suites.

### DIFF
--- a/docs/agents/development.md
+++ b/docs/agents/development.md
@@ -41,5 +41,8 @@ Standalone mode is single-user and unauthenticated. Multi-user deployments requi
 | `FRONTEND_SERVER_NAMESPACE` | Namespace used by a local frontend server |
 | `MINIO_ENDPOINT_REWRITE` | Rewrites object-store endpoints in local proxy mode |
 | `MAX_METRICS_FILE_BYTES` | Maximum uncompressed metrics JSON size; defaults to 1 MiB |
+| `DBDRIVERNAME` | API-server database driver: `mysql` or `pgx` |
 
 `TENSORBOARD_PROXY_SIGNING_SECRET` is optional; it defaults to `MINIO_SECRET_KEY`.
+
+`initConfig()` in `backend/src/apiserver/main.go` binds config keys through `viper.AutomaticEnv()` with a `.` → `_` replacer, so the `DBDriverName` key reads from `DBDRIVERNAME`. A variable named `DB_DRIVER_NAME` does not bind, leaving the `mysql` default in `backend/src/apiserver/config/config.json` in force. The PostgreSQL overlay sets `DBDRIVERNAME` from the `dbType` key of the `pipeline-install-config` ConfigMap.
--- a/manifests/kustomize/base/installs/generic/postgres/pipeline-install-config.yaml
+++ b/manifests/kustomize/base/installs/generic/postgres/pipeline-install-config.yaml
@@ -12,8 +12,8 @@ data:
     namespace: `kubectl rollout restart deployment -n <your-namespace>`.
   appName: pipeline
   appVersion: 2.0.0
-  dbType: postgres
-  postgresHost: postgres
+  dbType: pgx
+  postgresHost: postgres-service
   postgresPort: "5432"
   mlmdDb: metadb
   cacheDb: cachedb
--- a/manifests/kustomize/base/postgresql/pipeline/ml-pipeline-apiserver-deployment-patch.yaml
+++ b/manifests/kustomize/base/postgresql/pipeline/ml-pipeline-apiserver-deployment-patch.yaml
@@ -8,7 +8,6 @@ spec:
       initContainers:
       - name: wait-for-database
         env:
-        - $patch: replace
         - name: WAIT_HOST
           valueFrom:
             configMapKeyRef:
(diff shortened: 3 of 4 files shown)


hsinhoyeh:
Apologies — the first push of this branch was based on the `2.16.1` tag rather than `master`, so the PR carried the entire 2.15.0 → 2.16.1 release diff (+13569/−1639 across 100 files) alongside the intended change. That was unintentional and clearly not reviewable.

Force-pushed a rebase onto `master`. The PR is now **+8/−4 across 3 files** — only the postgres overlay configuration fixes, no version bump and nothing else:

- `third-party/postgresql/base/pg-secret.yaml` — `data:` → `stringData:`
- `base/installs/generic/postgres/pipeline-install-config.yaml` — `postgresHost: postgres-service`, `dbType: pgx`
- `base/postgresql/pipeline/ml-pipeline-apiserver-deployment-patch.yaml` — `DB_DRIVER_NAME` → `DBDRIVERNAME`, restore `OBJECTSTORECONFIG_HOST`/`PORT`

I re-confirmed all five bugs are present on `master` at `8c219561`, not only on the 2.16.1 tag, so the fix applies where it is targeted.

juliusvonkohout:
Please align with @kaikaila and the postgres PR. https://github.com/kubeflow/pipelines/pull/12379

hsinhoyeh:
Pushed an update addressing both review comments and @juliusvonkohout's request to align with #12379. Also added the missing `Signed-off-by` so DCO passes.

## Alignment with #12379

I read through kaikaila's branch before changing anything. #12379 replaces this overlay wholesale: `base/installs/generic/postgres/` and `base/postgresql/` go away, and `env/platform-agnostic-postgresql/` gets self-contained `apiserver-patch.yaml` / `cache-server-patch.yaml` / `mlmd-patch.yaml`. Where it overlaps with this PR it reaches the same conclusions:

| | this PR | #12379 |
|---|---|---|
| driver env var | `DBDRIVERNAME` | `DBDRIVERNAME` |
| driver value | `pgx` | `DBDriverName=pgx` |
| postgres host | `postgres-service` | `postgresHost=postgres-service` |
| api-server env list | merge, `$patch: delete` the MySQL entries | merge, `$patch: delete` the MySQL entries |

So I changed the one place the two diverged — see below — and left the rest alone.

`third-party/postgresql/base/pg-secret.yaml` is untouched by #12379, so the `data:`/`stringData:` fix (item 1) is still needed there either way and does not conflict.

Nothing here blocks #12379: every file this PR touches is either deleted by that PR or unmodified by it, so it lands as a clean delete rather than a conflict. If maintainers would rather wait for the full integration and drop this one, that is fine by me — this was only meant to unblock the current overlay for people hitting #13956 today.

## `$patch: replace` dropping `V2_DRIVER_IMAGE` / `V2_LAUNCHER_IMAGE`

Copilot is right, and my original "conservative option" was the wrong call — restoring only `OBJECTSTORECONFIG_HOST`/`PORT` left the pinned runtime images (and `PUBLISH_LOGS`, `CLUSTER_DOMAIN`, `COMPILED_PIPELINE_SPEC_PATCH`, the `DEFAULT_SECURITY_CONTEXT_*` set) still discarded, which is exactly the drift that produced these bugs in the first place.

Removed `$patch: replace` from both the init container and the api-server container, and instead `$patch: delete` only the MySQL-specific entries — the `mysql-secret` and `mysqlHost`/`mysqlPort` references that cannot resolve on a PostgreSQL install and would block Pod startup. This is the same approach #12379 takes. The patch now shrinks to just the deletes plus `DBDRIVERNAME` and the `DBCONFIG_POSTGRESQLCONFIG_*` block; everything else merges from the base and stays in sync automatically.

`kustomize build manifests/kustomize/env/platform-agnostic-postgresql` now renders an api-server that carries `V2_DRIVER_IMAGE: ghcr.io/kubeflow/kfp-driver:2.17.0` and `V2_LAUNCHER_IMAGE: ghcr.io/kubeflow/kfp-launcher:2.17.0`, keeps `OBJECTSTORECONFIG_HOST`/`PORT` from the base, and no longer references `mysql-secret` or `DB_DRIVER_NAME`. `env/dev/postgresql` builds clean too.

For what it is worth, the `cache-server` patch also uses `$patch: replace`, but there it currently drops nothing — the base env list is a subset of what the patch redefines. Left it alone since #12379 rewrites that file anyway.

## `AGENTS.md` not updated

Fair. Added `DBDRIVERNAME` to "Key environment variables" — what it selects, why `DB_DRIVER_NAME` does not bind (`viper.AutomaticEnv()` with the `.`→`_` replacer, `backend/src/apiserver/main.go:702-706`), and where the overlay sets it — and bumped "Last updated" to 2026-08-05.

hsinhoyeh:
/cc @kaikaila @juliusvonkohout

@kaikaila — this touches the same overlay #12379 rewrites, so I aligned it with yours rather than diverging: the api-server patch now merges into the base env list and `$patch: delete`s only the MySQL entries, exactly as your `apiserver-patch.yaml` does, and it uses the same `DBDRIVERNAME` / `pgx` / `postgres-service` values. Every file here is either deleted by #12379 or untouched by it, so it should land as a clean delete rather than a conflict. If you would rather not carry a stopgap alongside the full integration, I am happy to close this — it was only meant to unblock the current overlay for people hitting #13956 today.

@juliusvonkohout — the rebase onto master reset the test gate, so the PR is back to `needs-ok-to-test` and the ten suites are sitting in `action_required`. A `/ok-to-test` would let them run; they were all green on the previous head (`7369356`) with the same set of files.

kaikaila:
Hi @hsinhoyeh  Thanks for the thorough analysis. All issues you identified are real and well-documented.

#12379 covers all five fixes here, but takes a different structural approach: the current `base/postgresql/pipeline/` patch and `base/installs/generic/postgres/pipeline-install-config.yaml` are removed, and the PostgreSQL configuration moves into `env/platform-agnostic-postgresql/` as an overlay on top of the MySQL base — so the files this PR modifies would be deleted either way.

If this lands first I can resolve the conflicts on my side easily (they're all modify/delete), so no objection from me if reviewers want to merge it as an incremental fix.

juliusvonkohout:
Please do a proper rebase. Merge from master destroys the commit history.

juliusvonkohout:
> Hi @hsinhoyeh Thanks for the thorough analysis. All issues you identified are real and well-documented.
> 
> #12379 covers all five fixes here, but takes a different structural approach: the current `base/postgresql/pipeline/` patch and `base/installs/generic/postgres/pipeline-install-config.yaml` are removed, and the PostgreSQL configuration moves into `env/platform-agnostic-postgresql/` as an overlay on top of the MySQL base — so the files this PR modifies would be deleted either way.
> 
> If this lands first I can resolve the conflicts on my side easily (they're all modify/delete), so no objection from me if reviewers want to merge it as an incremental fix.

I am more in favour of waiting for https://github.com/kubeflow/pipelines/pull/12379. How long do you think it will take ?

kaikaila:
@juliusvonkohout Thanks for the preference note. I plan to address all outstanding review comments on #12379 by this week. That said, the actual merge timeline depends on reviewer's feedback and approval, which is difficult to foresee. I'll keep the PR moving as fast as I can on my end.

hsinhoyeh:
Closing in favor of #12379, which has since merged. It moves the PostgreSQL configuration into the `env/platform-agnostic-postgresql` overlay and removes the manifest files this PR patched, so the fixes here are covered by that change.

PR #13961, opened 2026-08-04
fix(frontend): keep the reason pod logs could not be read

uditjainstjis:

`composePodLogsStreamHandler` tries the log sources in order — Kubernetes API, then the Argo workflow status, then the archive bucket. When a source failed, its error was discarded and the next one was tried silently.

That loses the diagnostic in exactly the case the fallbacks exist for. Once a pod is gone, the Kubernetes API is the only source that can say *why* — evicted, OOMKilled, image pull failure, or a namespace the caller cannot read. The archive lookup then fails too, and `getPodLogsHandler` answers with the last error alone:

```
Could not get main container logs: Error: Unable to find pod log archive information
```

The reason reaches neither the response nor the server log.

The `console.warn` already in the `catch` block does not cover this: it runs only when no fallback was supplied, and `getPodLogsHandler` always supplies one ([`pod-logs.ts#L69-L79`](https://github.com/kubeflow/pipelines/blob/master/frontend/server/handlers/pod-logs.ts#L69-L79)), so on the deployed path it never runs at all.

**What this changes**

- Each failure is recorded as a `cause` of the error that is finally thrown.
- Causes are appended at the **end** of the chain rather than written to the first free slot. The handlers nest when log archiving is enabled — `compose(fromK8s, compose(fromWorkflow, fromArchive))` — so the archive error already carries the workflow error by the time the outer handler attaches the Kubernetes one. Writing to the first free slot would silently drop the Kubernetes error on precisely that configuration. The walk is depth capped.
- `getPodLogsHandler` logs the error it catches, which prints the chain.
- Falling back is logged at `debug`, not `warn` — a completed run's pod is garbage collected, so failing over to the archive is the ordinary path for it and not worth reporting on every request.

**What this deliberately does not change**

- The thrown value's identity and message are untouched, so the `Unable to find pod log archive information` check in `getPodLogsHandler` and any other caller matching on the message keep working. The existing test asserting `rejects.toEqual('unknown error for fallback')` passes unmodified.
- The response body still reports only the last error. Widening what is returned to the browser is a separate call, given it is attacker-visible.
- A cause chain that is already complete is left alone, and rejections that are plain strings rather than `Error`s pass through unchanged.

## Checklist

- [x] Do you want this pull request (PR) cherry-picked into the current release branch? — No
- [x] Unit tests added.

**Tests**

Five cases added to `workflow-helper.test.ts`, covering the flat two-source shape, the nested three-source shape, the already-set cause, and the debug log. `npx vitest run workflow-helper.test.ts` → 15 passed.

Two of the new tests fail against `master` without the source change, and the nested-shape test fails against a first-free-slot implementation, which is what motivated the chain walk.

The `integration-tests/mock-backend-*` suites fail in my local checkout both with and without this change (54 failures either way); they are untouched by this PR.

### DIFF
--- a/frontend/server/handlers/pod-logs.ts
+++ b/frontend/server/handlers/pod-logs.ts
@@ -133,6 +133,11 @@ export function getPodLogsHandler(
       stream.on('end', () => res.end());
       stream.pipe(res);
     } catch (err) {
+      // Every log source failed. Logging the error prints the chain of causes
+      // built up by composePodLogsStreamHandler, which is the only place the
+      // earlier failures - notably why the pod itself could not be read - are
+      // still recorded. The response keeps reporting the last error alone.
+      console.error(`Could not get main container logs for pod, ${podName}.`, err);
       res.status(500).send('Could not get main container logs: ' + err);
     }
   };
--- a/frontend/server/workflow-helper.test.ts
+++ b/frontend/server/workflow-helper.test.ts
@@ -93,6 +93,91 @@ describe('workflow-helper', () => {
         ),
       ).rejects.toEqual('unknown error for fallback');
     });
+
+    it('records the default handler error as the cause when both fail.', async () => {
+      const defaultError = new Error('pods "podName" not found');
+      const fallbackError = new Error('Unable to find pod log archive information');
+      const defaultHandler = vi.fn((_podName: string, _createdAt: string, _namespace?: string) =>
+        Promise.reject(defaultError),
+      );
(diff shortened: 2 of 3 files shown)


droctothorpe:
@uditjainstjis: Hey, thank you for your PR! As you are not a member of Kubeflow organization, your PR must link to an issue with 'ready' label. Please read our [CONTRIBUTING.md](https://github.com/kubeflow/pipelines/blob/master/CONTRIBUTING.md), closing this PR

PR #13974, opened 2026-08-04
ci: wait for active image producers

jeffspahr:
## Summary

- extend the image-artifact barrier while a required producer is queued or running
- fail early when a producer completes unsuccessfully without publishing its artifact
- allow a short publication grace after successful producers finish
- bound the fallback when GitHub producer-state queries remain unavailable

## Why

Image-consuming test matrices currently start alongside the reusable image-build workflow. The shared deploy action gives producers a fixed 20-minute artifact window. A producer delayed by hosted-runner scheduling or a slow external package mirror can remain healthy beyond that window, causing every consumer lane to fail before tests begin.

This preserves the intentional cluster/build overlap while making the barrier depend on the state of the specific missing producer. One slow producer therefore no longer fans out into dozens of premature deployment failures, while genuine producer failures still terminate promptly with a useful diagnostic.

## Verification

- `python3 .github/resources/scripts/wait_for_image_artifacts_test.py -v`: 12 tests passed
- CI scripts unit suite: 157 tests passed
- `bash -n .github/resources/scripts/wait-for-image-artifacts.sh`
- `git diff --check`


zazulam (commented):
/lgtm
/approve

PR #13978, opened 2026-08-04
fix(ci): resume Trivy vulnerability scanning

jeffspahr:
## Summary

- allow the scheduled filesystem scan to download current Trivy vulnerability databases from the configured ECR mirrors
- pin reviewed Trivy action and binary releases, explicitly scope the scan to vulnerabilities, and report every severity including findings without an available fix
- add least-privilege permissions, manual dispatch, and regression coverage for the workflow's security and database-update invariants

## Root cause

PR #12563 moved database updates into a separate daily cache workflow and configured the scanner with `TRIVY_SKIP_DB_UPDATE` and `TRIVY_SKIP_JAVA_DB_UPDATE`. PR #12612 removed that cache workflow on January 6, 2026, but left both skip flags in the scanner. The January 9 scan could still use the remaining cache; after it expired, fresh runners failed because Trivy was forbidden from downloading its initial databases.

## Verification

- `python3 -m unittest -v trivy_workflow_test.py`
- `go run github.com/rhysd/actionlint/cmd/actionlint@v1.7.9 .github/workflows/trivy.yml .github/workflows/ci-scripts-tests.yml`
- `git diff --check`

### DIFF
--- a/.github/resources/scripts/trivy_workflow_test.py
+++ b/.github/resources/scripts/trivy_workflow_test.py
@@ -43,15 +43,18 @@ class TrivyWorkflowTest(unittest.TestCase):
         self.assertNotIn('TRIVY_SKIP_DB_UPDATE', self.workflow)
         self.assertNotIn('TRIVY_SKIP_JAVA_DB_UPDATE', self.workflow)
 
-    def test_scan_uses_reviewed_pins_and_vulnerability_scope(self):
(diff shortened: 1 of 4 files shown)


jeffspahr:
Superseded by #13979, which migrates the scheduled repository dependency scan to OSV-Scanner instead of restoring Trivy.

Issue #13713, opened 2026-07-13
[backend] kfp-metadata-writer crashes with MLMD AlreadyExistsError when workflow Context already exists


sagittariuslee:
### Environment

Python in kfp-metadata-writer:
3.9.21

ML Metadata / MLMD package version in kfp-metadata-writer:
1.14.0

Argo executor image used by affected workflow pods:
argoproj/argoexec:v3.4.17

The kfp-metadata-writer container does not have the kfp Python package installed, but it does have ml-metadata==1.14.0.

The failure appears server-side in kfp-metadata-writer while processing Workflow/Pod events and writing MLMD Context records, not inside a user SDK call.

### Steps to reproduce
* Run KFP workflows that create Argo Workflow CRs and MLMD run Context records.
* Allow a parent/orchestration workflow to create multiple short-lived child workflows.

### Expected result

```BASH
2026-07-13T14:23:05.547Z [kfp-metadata-writer] Traceback (most recent call last):
2026-07-13T14:23:05.547Z [kfp-metadata-writer]   File "/kfp/metadata_writer/metadata_writer.py", line 230, in <module>
2026-07-13T14:23:05.547Z [kfp-metadata-writer]     run_context = get_or_create_run_context(
2026-07-13T14:23:05.547Z [kfp-metadata-writer]   File "/kfp/metadata_writer/metadata_helpers.py", line 261, in get_or_create_run_context
2026-07-13T14:23:05.547Z [kfp-metadata-writer]     context = get_or_create_context_with_type(
2026-07-13T14:23:05.547Z [kfp-metadata-writer]   File "/kfp/metadata_writer/metadata_helpers.py", line 196, in get_or_create_context_with_type
2026-07-13T14:23:05.547Z [kfp-metadata-writer]     context = create_context_with_type(
2026-07-13T14:23:05.547Z [kfp-metadata-writer]   File "/kfp/metadata_writer/metadata_helpers.py", line 168, in create_context_with_type
2026-07-13T14:23:05.547Z [kfp-metadata-writer]     context.id = store.put_contexts([context])[0]
2026-07-13T14:23:05.547Z [kfp-metadata-writer]   File "/lib/python3.9/site-packages/ml_metadata/metadata_store/metadata_store.py", line 520, in put_contexts
2026-07-13T14:23:05.547Z [kfp-metadata-writer]     self._call('PutContexts', request, response)
[...]

PR #13981, opened 2026-08-05
fix: replace bare except with ValueError in metadata_helpers.py

praneshnikhar:
# Description

The `get_or_create_*` functions and `get_or_create_context_with_type` in `metadata_helpers.py` used bare `except:` clauses to catch "not found" errors from the ML Metadata store. The ml-metadata library raises `ValueError` when a type or context is not found, but bare `except:` catches everything — including `TypeError`, `KeyError`, network errors, memory errors — and silently falls through to the create path. This risks creating duplicate or invalid MLMD entities when unrelated errors occur.

# Change type

/kind bug

# Changelog

```release-note
fix: Replaced bare except: with except ValueError: in metadata_helpers.py get_or_create_* and get_or_create_context_with_type functions to prevent swallowing unrelated errors
```

### DIFF
--- a/backend/metadata_writer/src/metadata_helpers.py
+++ b/backend/metadata_writer/src/metadata_helpers.py
@@ -83,7 +83,7 @@ def get_or_create_artifact_type(store, type_name, properties: dict = None) -> me
     try:
         artifact_type = store.get_artifact_type(type_name=type_name)
         return artifact_type
-    except:
+    except ValueError:
         artifact_type = metadata_store_pb2.ArtifactType(
             name=type_name,
             properties=properties,
@@ -96,7 +96,7 @@ def get_or_create_execution_type(store, type_name, properties: dict = None) -> m
     try:
         execution_type = store.get_execution_type(type_name=type_name)
         return execution_type
-    except:
+    except ValueError:
         execution_type = metadata_store_pb2.ExecutionType(
             name=type_name,
             properties=properties,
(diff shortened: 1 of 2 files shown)


hbelmiro (commented):
Could you please add unit tests for the four get-or-create functions? The store is passed as a parameter, so they're easy to test with a mock. For each function, cover at least:

1. Get path: mock the get method to return an existing type/context — assert the put method is never called and the existing object is returned.
2. Create path: mock the get method to raise the caught exception — assert the put method is called and the new object is returned.
3. Unrelated exception propagates: mock the get method to raise a different exception (e.g., `RuntimeError`) — assert it propagates instead of silently falling through to the create path.

hbelmiro (commented):
The PR description claims ml-metadata raises `ValueError` on not-found, but inspecting the actual ml-metadata 1.21.0 source (`metadata_store.py:1104-1105`) shows `get_artifact_type` documents:

```
Raises:
    errors.NotFoundError: if no type exists.
    errors.InternalError: if query execution fails.
```

`errors.NotFoundError` inherits from `StatusError(Exception)` — it is **not** a subclass of `ValueError`. Catching `ValueError` here will never match a not-found response from the store, so the create path becomes unreachable. Every call will propagate `NotFoundError` as an unhandled exception instead of falling through to create the type.

jeffspahr:
Thanks for the exception-handling improvement. Closing because #13986 removed metadata-writer, including metadata_helpers.py, from master / KFP 3.0. These changes therefore no longer have an active code path on the target branch.

Release-2.18 retains metadata-writer. The context lookup/error-handling fix for that line is being followed through #13713 and #14411, with changes requested and a request to target release-2.18. This closure does not claim that this patch was merged or backported.

PR #13982, opened 2026-08-05
fix: log rows.Close() errors instead of silently discarding them

praneshnikhar:
The `dropLegacyIndexesMySQL` function in `client_manager.go` used `_ = rows.Close()` to suppress close errors from `sql.Rows`. This silently drops connection/IO errors that could indicate resource leaks or database issues.

Replaced `_ = rows.Close()` with a check that logs the error via `glog.Warningf` so close failures are visible in logs without altering control flow.

**Change type**
/kind bug

**Changelog**
fix: Log `sql.Rows.Close()` errors in `client_manager.go` instead of silently discarding them


Priyanshu-u07:
Thanks for the PR @praneshnikhar! Just sharing what I found when I looked into this.

I tested this against real MySQL with the driver this repo pins (`go-sql-driver/mysql v1.10.0`), going through `db.Raw(...).Rows()` like the code here does. None of the paths this PR touches can produce a log line:

- The first new log line sits inside the existing `if err := rows.Err(); err != nil` branch. That check runs before any `rows.Next()`, and `Err()` only reports iteration errors, so it's nil there and the branch never executes.
- After `Next()` returns false, `database/sql` has already auto-closed the rows, so `Close()` returns nil.
- With rows still open, `Close()` drains the unread rows, but only touches the network if they weren't already buffered. `SHOW INDEX` returns ~200 bytes, so it drains from memory. I measured `nil` even with the server killed mid-query.

So, I think the new `glog.Warningf` lines can't fire here. Nothing is broken, but it doesn't look like it changes behaviour.

Minor: the description refers to `migrateCleanupUniqueIndexes` but the function is `dropLegacyIndexesMySQL`.

droctothorpe:
@praneshnikhar: Hey, thank you for your PR! As you are not a member of Kubeflow organization, your PR must link to an issue with 'ready' label. Please read our [CONTRIBUTING.md](https://github.com/kubeflow/pipelines/blob/master/CONTRIBUTING.md), closing this PR

PR #13896, opened 2026-07-27
fix(sdk): remove _KFP_RUNTIME and decouple KFP dependency in Executor

vizansh:
### Description
Fixes #12090

This PR removes the `_KFP_RUNTIME` environment variable and decouples top-level `kfp` imports inside `dsl.executor`.

### Changes Made
- Refactored `sdk/python/kfp/dsl/executor.py` to use direct leaf imports (`from kfp.dsl.types.artifact_types import Artifact`, `from kfp.dsl.task_config import TaskConfig`, etc.) instead of `from kfp import dsl`.
- Updated `sdk/python/kfp/dsl/component_factory.py` to strip `_KFP_RUNTIME=true` from generated container component command scripts.
- Removed `_KFP_RUNTIME` conditional guards in `sdk/python/kfp/__init__.py` and `sdk/python/kfp/dsl/__init__.py`.

### Validation
- Verified `Executor` imports cleanly without triggering compile-time DSL dependencies.
- Ran unit test suites:
  - `python3 -m pytest sdk/python/kfp/dsl/executor_test.py` (47 passed)
  - `python3 -m pytest sdk/python/kfp/dsl/component_factory_test.py` (25 passed)

PR #12811, opened 2026-02-13
fix(backend): resolve input parameter placeholders in list/struct constants

puwun:
Fixes #12204

**Description of your changes:**
The backend driver only resolved  `{{$.inputs.parameters[...]}}`  placeholders in string constants. When pipeline parameters were wrapped in a list (e.g `recipients=[email]`), the placeholders inside list/struct constants were left unresolved at runtime. Added recursive placeholder resolution for list and struct constant values.

### Changes made:
- Add `resolveValuePlaceholders` function and a post-processing step in `resolveInputs` in `resolve.go` to recursively resolve parameter placeholders in list/struct constants
- Add 10 tests in `resolve_test.go` for the new function

**Checklist:**
- [x] You have [signed off your commits](https://www.kubeflow.org/docs/about/contributing/#sign-off-your-commits)
- [x] The title for your pull request (PR) should follow our title convention. [Learn more about the pull request title convention used in this repository](https://github.com/kubeflow/pipelines/blob/master/CONTRIBUTING.md#pull-request-title-convention).

PR #13984, opened 2026-08-05
fix(backend): iterate placeholder substitution to fixpoint in launcher_v2

### LINKED ISSUE #13977: [backend] Inconsistent Nested Placeholder Resolution
### Environment

  *  How did you deploy Kubeflow Pipelines (KFP)? Standalone / Dev Kind Deployment
  *  KFP version: v2
  *  KFP SDK version: 2.x

  ### Steps to reproduce

  A component whose container `args` reference a **parameter individually** (e.g. `{{$.inputs.parameters['dask_config']}}`, rather than the whole `{{$}}` executor input), where that parameter is a **constant Struct that contains a nested
  pipeline-channel placeholder**, is intermittently deployed with the nested placeholder left as a literal string.

  Compiled `deploymentSpec` (abridged):

  ```yaml
  exec-my-comp:
    container:
      args:
        - '-p'
        - dask_config
        - "{{$.inputs.parameters['dask_config']}}"
        - json
  ```

  with the task/param defined as a constant Struct containing a nested channel reference:

  ```yaml
  dask_config:
    runtimeValue:
      constant:
        n_workers: 2
        pvc_name: "{{$.inputs.parameters['pipelinechannel--volume_name']}}"
        # ...
  ```

  `pipelinechannel--volume_name` is a synthetic task-level input bound via `componentInputParameter: volume_name`; it is present in the task's `inputs.parameters` (and thus in `ExecutorInput.parameterValues`), but not in the component
  `inputDefinitions`.

  On roughly half of runs, the launched container receives:

  ```
  pvc_name: "{{$.inputs.parameters['pipelinechannel--volume_name']}}"   # literal, unresolved
  ```

  instead of the resolved value.

  **Root cause** — `backend/src/v2/component/launcher_v2.go`, `compileCmdAndArgs` (around L1063–1071):

  ```go
  for _, arg := range args {
      compiledArgTemplate := strings.ReplaceAll(arg, executorInputJSONKey, executorInputJSONString)
      for placeholder, replacement := range placeholders {   // <-- Go map: random iteration order
          compiledArgTemplate = strings.ReplaceAll(compiledArgTemplate, placeholder, replacement)
      }
      compiledArgs = append(compiledArgs, compiledArgTemplate)
  }
  ```

  This is a **single pass** over the `placeholders` map, whose iteration order is randomized by Go. Substituting `{{$.inputs.parameters['dask_config']}}` *injects* the nested `{{$.inputs.parameters['pipelinechannel--volume_name']}}` text
  into the arg. If `pipelinechannel--volume_name` was already iterated earlier in that same pass, it is never revisited, so the nested placeholder survives. If `dask_config` happens to be iterated first, the nested placeholder is then
  caught. → order-dependent, intermittent.

  The driver does not pre-resolve this: `driver/resolve.go` `RuntimeValue_Constant` handling (~L425–454) returns the constant Struct **verbatim** (`default: v = val`) for anything other than a fixed set of `{{$.pipeline_*}}` tokens, so the
  nested placeholder reaches the launcher unresolved by design.

  Note: components that consume the whole executor input via `{{$}}` are **not** affected, because `{{$}}` is expanded to the full `ExecutorInput` JSON *before* the map loop (L1067), so the nested placeholder text is present from the first
  iteration and gets substituted regardless of order.

  ### Expected result

  Placeholder substitution should be order-independent. A placeholder that only appears *after* another substitution (nested inside a constant Struct value) should always be resolved.

### Proposed solution
Iterate substitution to a fixpoint; repeat the placeholder pass until the string stops changing instead of a single pass.

```go
func compileCmdAndArgs(executorInput *pipelinespec.ExecutorInput, cmd string, args []string) (string, []string, error) {
        placeholders, err := getPlaceholders(executorInput)
        if err != nil {
                return "", nil, err
        }

        executorInputJSON, err := protojson.Marshal(executorInput)
        if err != nil {
                return "", nil, fmt.Errorf("failed to convert ExecutorInput into JSON: %w", err)
        }
        executorInputJSONKey := "{{$}}"
        executorInputJSONString := string(executorInputJSON)

        compiledCmd := substitutePlaceholders(cmd, executorInputJSONKey, executorInputJSONString, placeholders)

        compiledArgs := make([]string, 0, len(args))
        for _, arg := range args {
                compiledArgs = append(compiledArgs,
                        substitutePlaceholders(arg, executorInputJSONKey, executorInputJSONString, placeholders))
        }
        return compiledCmd, compiledArgs, nil
  }

// substitutePlaceholders expands {{$}} once, then repeatedly applies the
// placeholder map until no further substitutions occur (fixpoint). This makes
// resolution order-independent: a placeholder whose text is only introduced by
// another substitution (e.g. pipelinechannel--volume_name nested inside a
// resolved dask_config struct) is still resolved on a subsequent pass.
func substitutePlaceholders(s, execInputKey, execInputJSON string, placeholders map[string]string) string {
        s = strings.ReplaceAll(s, execInputKey, execInputJSON)
        for {
                prev := s
                for placeholder, replacement := range placeholders {
                        s = strings.ReplaceAll(s, placeholder, replacement)
                }
                if s == prev {
                        break
                }
        }
        return s
  }
```

vizansh:
Fixes #13977

Thanks to @ahn132 for the issue report and root cause breakdown.

### DIFF
--- a/backend/src/v2/component/launcher_v2.go
+++ b/backend/src/v2/component/launcher_v2.go
@@ -1047,30 +1047,42 @@ func fetchNonDefaultBuckets(
 	return nonDefaultBuckets, nil
 
 }
-
 func compileCmdAndArgs(executorInput *pipelinespec.ExecutorInput, cmd string, args []string) (string, []string, error) {
 	placeholders, err := getPlaceholders(executorInput)
+	if err != nil {
+		return "", nil, err
+	}
 
 	executorInputJSON, err := protojson.Marshal(executorInput)
 	if err != nil {
 		return "", nil, fmt.Errorf("failed to convert ExecutorInput into JSON: %w", err)
 	}
+
 	executorInputJSONKey := "{{$}}"
 	executorInputJSONString := string(executorInputJSON)
 
-	compiledCmd := strings.ReplaceAll(cmd, executorInputJSONKey, executorInputJSONString)
+	compiledCmd := substitutePlaceholders(cmd, executorInputJSONKey, executorInputJSONString, placeholders)
 	compiledArgs := make([]string, 0, len(args))
-	for placeholder, replacement := range placeholders {
-		cmd = strings.ReplaceAll(cmd, placeholder, replacement)
-	}
 	for _, arg := range args {
-		compiledArgTemplate := strings.ReplaceAll(arg, executorInputJSONKey, executorInputJSONString)
-		for placeholder, replacement := range placeholders {
-			compiledArgTemplate = strings.ReplaceAll(compiledArgTemplate, placeholder, replacement)
-		}
-		compiledArgs = append(compiledArgs, compiledArgTemplate)
+		compiledArgs = append(compiledArgs, substitutePlaceholders(arg, executorInputJSONKey, executorInputJSONString, placeholders))
 	}
+
 	return compiledCmd, compiledArgs, nil
+}
+
+func substitutePlaceholders(s, execInputKey, execInputJSON string, placeholders map[string]string) string {
+    s = strings.ReplaceAll(s, execInputKey, execInputJSON)
+    maxPasses := 10 // Prevents infinite loops on cyclic/self-referential placeholders
+    for pass := 0; pass < maxPasses; pass++ {
+        prev := s
+        for placeholder, replacement := range placeholders {
+            s = strings.ReplaceAll(s, placeholder, replacement)
+        }
+        if s == prev {
+            break
+        }
+    }
+    return s
 }
 
 // Add executor input placeholders to provided map.
--- a/backend/src/v2/component/launcher_v2_test.go
+++ b/backend/src/v2/component/launcher_v2_test.go
@@ -918,3 +918,28 @@ func Test_retrieve_artifact_path(t *testing.T) {
 		})
 	}
 }
+
+func TestCompileCmdAndArgs_NestedPlaceholders(t *testing.T) {
+	executorInput := &pipelinespec.ExecutorInput{
+		Inputs: &pipelinespec.ExecutorInput_Inputs{
+			ParameterValues: map[string]*structpb.Value{
+				"dask_config":                  structpb.NewStringValue(`{"pvc_name": "{{$.inputs.parameters['pipelinechannel--volume_name']}}"}`),
+				"pipelinechannel--volume_name": structpb.NewStringValue("my-pvc-volume"),
+			},
+		},
+	}
+
+	args := []string{
+		"--config={{$.inputs.parameters['dask_config']}}",
+	}
+
+	// Loop 100 times to guarantee map order randomization doesn't mask failures
+	for i := 0; i < 100; i++ {
+		_, compiledArgs, err := compileCmdAndArgs(executorInput, "", args)
+		assert.NoError(t, err)
+		assert.Len(t, compiledArgs, 1)
+
+		expectedArg := `--config={"pvc_name": "my-pvc-volume"}`
+		assert.Equal(t, expectedArg, compiledArgs[0], "Nested placeholder failed to resolve due to map order randomization!")
+	}
+}


dobbydobap:
Nice catch adding the `getPlaceholders` error check on top of the reported bug. That one was hiding the real cause behind a confusing `failed to convert ExecutorInput into JSON`.

The loop needs a cap though:

```go
for {
    prev := s
    for placeholder, replacement := range placeholders {
        s = strings.ReplaceAll(s, placeholder, replacement)
    }
    if s == prev {
        break
    }
}
```

It stops at a fixpoint, but a value containing its own placeholder never reaches one. Set parameter `a` to `x{{$.inputs.parameters['a']}}` and each pass re-injects the placeholder with one more `x` in front, so `s` never equals `prev`. Two parameters referencing each other do the same. Pipeline parameters are user supplied, so this hangs the launcher on ordinary input instead of erroring. A counted loop with a small cap keeps the nesting and drops the hang; anything unresolved after that passes through literally, same as today.

The SDK files look accidental. Removing the `_KFP_RUNTIME` guard is your #13896, not this fix, so the branch may have been cut from there rather than master. `launcher_v2_test.go` also lost its trailing newline.

On the test, it runs once over two placeholders, so map order gives it roughly a 50% pass rate against the unfixed code. Looping it makes the regression reproducible.

Worth a look at #12811 too, which resolves placeholders inside list and struct constants in the driver. If that lands, some of these nested cases would be resolved before the launcher sees them. Not a blocker, just worth settling which layer owns it.

vizansh:
Thanks @dobbydobap for the thorough review! These were excellent points:

1. **Infinite Loop Cap:** Added a `maxPasses` cap (10) to `substitutePlaceholders` to prevent infinite loops in the event of self-referential or cyclic user parameters.
2. **Flaky Test Fix:** Updated `TestCompileCmdAndArgs_NestedPlaceholders` to run 100 iterations per test run to guarantee map iteration order randomization doesn't hide regressions.

Updated and force-pushed!

dobbydobap:
Cap logic is right, and dropping the SDK files cleaned it up nicely.

One thing that will bite CI though: the new `substitutePlaceholders` body is indented with spaces where the rest of the file uses tabs. At head `6fa2aa19b` lines 1074-1085 are space-indented, while `getPlaceholders` right below at 1090 is tabs. `.golangci.yml` enables `gofmt` and `goimports`, and pre-commit runs golangci-lint with `--new-from-merge-base`, so it lints exactly the lines you changed. `gofmt -w backend/src/v2/component/launcher_v2.go` should clear it.

zazulam:
@vizansh please sign your commit

dobbydobap:
ok-to-test has landed, golangci-lint confirms the formatting point from earlier:

```
backend/src/v2/component/launcher_v2.go:1074:1: File is not properly formatted (gofmt)
backend/src/v2/component/launcher_v2_test.go:945:1: File is not properly formatted (gofmt)
```

1074 is the `substitutePlaceholders` body, which is space-indented where the rest of the file uses tabs. `gofmt -w` on both files clears it.DCO is the other blocker, which is what @zazulam was after: `6fa2aa19b` has no `Signed-off-by` trailer. If you reformat first, `git commit --amend -s` plus a force push covers both in one go.

vizansh:
Hi @zazulam @dobbydobap, sorry for the delay, I was away for a while but
I'm back on this.

- I've force-pushed an updated commit (3471966), rebased on current master:

- DCO: the commit is now signed off, and the DCO check is green.
- gofmt: ran gofmt -w on launcher_v2.go and launcher_v2_test.go, so the
  space-indented body flagged earlier is fixed.

Locally, TestCompileCmdAndArgs_NestedPlaceholders fails on the old code
on all 100 iterations and passes with the fix, and the full component
package tests pass.

On CI: the K8s Native API job failed on 8 specs, but they all die during
pipeline upload, before any run starts, with:

  Failed to create the pipeline version: etcdserver: request timed out

The launcher isn't involved at that point, so this looks like an infra
flake rather than something from this change. Could someone /retest?

Otherwise this should be ready for review. Thanks for the patience!

PR #13990, opened 2026-08-07
fix(sdk): load v1 component I/O names that YAML 1.1 resolves as booleans. Fixes #13756

### LINKED ISSUE #13756: [sdk] load_component_from_text fails for v1 components with YAML-boolean input names (on, off, yes, no)
### Environment

* **KFP version:** 2.17.0 (repo `VERSION` file; standalone Kind cluster)
* **KFP SDK version:** 2.15.2 (editable install from `master` at commit `1af7e02bf`)
* **All dependencies version:**
  ```
  kfp                       2.15.2
  kfp-kubernetes            2.15.2
  kfp-pipeline-spec         2.15.2
  kfp-server-api            2.17.0
  ```
* **Python:** 3.10.12
* **OS:** Linux

### Steps to reproduce

1. Create a legacy v1 component YAML using **flow-style** syntax with an input name that YAML 1.1 treats as a boolean:

```yaml
name: test
inputs:
- {name: off, type: String}
implementation:
  container:
    image: alpine
    command: [echo]
```

2. Load the component:

```python
from kfp.components import load_component_from_text

load_component_from_text("""
name: test
inputs:
- {name: off, type: String}
implementation:
  container:
    image: alpine
    command: [echo]
""")
```

3. The same failure occurs with **block-style** unquoted names:

```yaml
name: test
inputs:
- name: off
  type: String
implementation:
  container:
    image: alpine
    command: [echo]
```

4. Affected input names (verified locally):

| Input name | YAML parses `name` as | Result |
|------------|----------------------|--------|
| `on` | `True` | `TypeError` |
| `off` | `False` | `TypeError` |
| `yes` | `True` | `TypeError` |
| `no` | `False` | `TypeError` |

**Actual error (example for `off`):**
```
TypeError: Error: InputSpec.from_dict(struct={'name': False, 'type': 'String'}) failed with exception:
Error: Structure "False" is incompatible with type "<class 'str'>". Structure is not the instance of the type, the type does not have .from_dict method and is not generic.
Error: Structure "[{'name': False, 'type': 'String'}]" is not None.
Error: Structure "[{'name': False, 'type': 'String'}]" is incompatible with type "typing.Optional[typing.List[kfp.dsl.v1_structures.InputSpec]]" - none of the types in Union are compatible.
```

5. **Workaround** (quoted name loads correctly):

```yaml
inputs:
- {name: "off", type: String}
```

```python
comp = load_component_from_text(...)
list(comp.component_spec.inputs.keys())  # ['off'] - works
```

### Expected result

- Input names like `on`, `off`, `yes`, and `no` should load as **strings**, regardless of YAML flow-style vs block-style syntax.
- `load_component_from_text()` should not fail because PyYAML coerced the `name` field to a Python `bool`.
- If a name cannot be parsed safely, the SDK should raise a clear `ValueError` (e.g. “input name must be a string, got bool”), not a nested `TypeError` from `InputSpec.from_dict`.

### Materials and Reference

**Root cause:** In `sdk/python/kfp/dsl/structures.py`, `_load_component_spec_from_component_text()` uses `yaml.safe_load(text)` with no post-processing of input `name` fields. PyYAML (YAML 1.1) coerces unquoted scalars like `on`, `off`, `yes`, `no` to booleans before `ComponentSpec.from_dict()` runs.

```1085:1088:sdk/python/kfp/dsl/structures.py
def _load_component_spec_from_component_text(
        text) -> v1_structures.ComponentSpec:
    component_dict = yaml.safe_load(text)
    component_spec = v1_structures.ComponentSpec.from_dict(component_dict)
```

**Why this matters:** Legacy v1 component YAML often uses flow-style `{name: off, type: String}`. Names like `off` and `on` are valid parameter names but are common in real pipelines (e.g. feature flags, toggles).

---

Yigtwxx:

`load_component_from_text` and `load_component_from_file` failed on v1 component YAML whose input or output name is one of the YAML 1.1 boolean words `on`, `off`, `yes` or `no`, in both flow style and block style:

```yaml
name: test
inputs:
- {name: off, type: String}
implementation:
  container:
    image: alpine
    command: [echo]
```

PyYAML implements YAML 1.1, so `yaml.safe_load` resolved those unquoted scalars to Python booleans before `v1_structures.ComponentSpec.from_dict` ran, producing a nested `TypeError`:

```
TypeError: Error: InputSpec.from_dict(struct={'name': False, 'type': 'String'}) failed with exception:
Error: Structure "False" is incompatible with type "<class 'str'>". ...
```

The same coercion also affects placeholders: `{inputValue: off}` became `{'inputValue': False}`, which then fails in `utils.sanitize_input_name` with `AttributeError: 'bool' object has no attribute 'lower'`. Repairing only the `name` fields would therefore not be enough, and `bool -> str` cannot recover the original spelling (`on`, `On`, `ON`, ...).

Changes, all scoped to `_load_component_spec_from_component_text`, the single entry point for v1 component text:

* `_V1ComponentYamlLoader` is a `yaml.SafeLoader` subclass whose implicit resolvers keep the boolean tag only for scalars starting with `t`/`T`/`f`/`F`. This matches the YAML 1.2 core schema: `true`/`false` still resolve to booleans, `null`/`~` still resolve to null, and `on`/`off`/`yes`/`no` stay strings.
* `_normalize_v1_io_specs` restores the boolean meaning of the `optional` field when it is written as `yes`/`no`/`on`/`off`, so component YAML that loads today keeps loading. `optional` is the only strictly boolean-typed field in `v1_structures`.
* The same pass raises a `ValueError` naming the offending input or output when a name is not a string, replacing the nested `TypeError` from `InputSpec.from_dict`.

Boolean-typed defaults are unaffected: `default: yes` now reaches `type_utils.deserialize_v1_component_yaml_default` as the string `'yes'`, and `bool_cast_fn`/`strtobool` already map `yes`/`no`/`on`/`off` to booleans, so the resulting default is still `True`. A `String`-typed input with `default: yes` changes from `'True'` to `'yes'`, which is the value the YAML actually spells.

`load_documents_from_yaml` and the v2 IR path are deliberately unchanged. Compiled IR YAML is emitted by PyYAML, whose emitter quotes scalars that would otherwise resolve to a non-string type, so IR round-trips correctly today.

Fixes #13756

**Checklist:**

### DIFF
--- a/sdk/RELEASE.md
+++ b/sdk/RELEASE.md
@@ -8,6 +8,8 @@
 
 ## Bug fixes and other changes
 
+* Load v1 component input and output names that YAML 1.1 resolves as booleans (`on`, `off`, `yes`, `no`) as strings instead of failing with a `TypeError`. Fixes #13756
+
 # 2.15.2
 
 ## Bug fixes and other changes
--- a/sdk/python/kfp/dsl/structures.py
+++ b/sdk/python/kfp/dsl/structures.py
@@ -1084,9 +1084,71 @@ def load_documents_from_yaml(component_yaml: str) -> Tuple[dict, dict]:
     return pipeline_spec_dict, platform_spec_dict
 
 
+class _V1ComponentYamlLoader(yaml.SafeLoader):
+    """YAML loader that resolves booleans per the YAML 1.2 core schema.
+
+    PyYAML implements YAML 1.1, which resolves the unquoted scalars
+    ``on``, ``off``, ``yes`` and ``no`` to booleans. v1 component YAML
+    uses these words as ordinary strings, most commonly as input and
+    output names.
+    """
+
+
+def _drop_yaml_1_1_bool_resolvers() -> Dict[str, List[tuple]]:
+    """Builds implicit resolvers that only treat t/T/f/F scalars as bools.
+
+    This restricts implicit booleans to
+    true/True/TRUE/false/False/FALSE.
+    """
+    resolvers_by_first_char = {}
+    for first_char, resolvers in yaml.SafeLoader.yaml_implicit_resolvers.items(
+    ):
+        resolvers_by_first_char[first_char] = [
+            (tag, regexp)
+            for tag, regexp in resolvers
+            if tag != 'tag:yaml.org,2002:bool' or first_char in 'tTfF'
+        ]
+    return resolvers_by_first_char
+
+
+_V1ComponentYamlLoader.yaml_implicit_resolvers = _drop_yaml_1_1_bool_resolvers()
+
+_YAML_1_1_BOOL_STRINGS = {
+    'yes': True,
+    'no': False,
+    'on': True,
+    'off': False,
+}
+
+
+def _normalize_v1_io_specs(component_dict: Dict[str, Any]) -> None:
+    """Validates v1 I/O names and restores boolean ``optional`` values.
+
+    ``_V1ComponentYamlLoader`` keeps ``on``/``off``/``yes``/``no`` as
+    strings so that they can be used as I/O names. ``optional`` is
+    declared as a boolean, so the boolean meaning is restored for that
+    field only.
+    """
+    for field_name, io_kind in (('inputs', 'input'), ('outputs', 'output')):
+        for spec in component_dict.get(field_name) or []:
+            if not isinstance(spec, dict):
+                continue
+            name = spec.get('name')
+            if not isinstance(name, str):
+                raise ValueError(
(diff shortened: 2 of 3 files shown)


droctothorpe:
Hi @Yigtwxx, thank you for your PR! The linked issue (#13756) must be labeled `ready` before your contribution can be reviewed. Please check our [CONTRIBUTING.md](https://github.com/kubeflow/pipelines/blob/master/CONTRIBUTING.md) for details.

PR #13993, opened 2026-08-08
fix(backend): select the storage dialect by driver instead of always MySQL

hsinhoyeh:
Part of #13956. This is a bug I found while scoping items 1 and 2 that the issue does not mention.

`initDBClient` ends with:

```go
return storage.NewDB(newdb, storage.NewMySQLDialect())
```

The dialect is hardcoded, so **every install gets `MySQLDialect` regardless of `driverName`** — including PostgreSQL installs, which reach this line through the `pgx` branch a few lines above.

## What that breaks

`storage.SQLDialect` exists specifically to paper over engine differences, and only `MySQLDialect` and `SQLiteDialect` implement it — there is no PostgreSQL implementation at all. So on a PostgreSQL install the stores emit MySQL-only SQL:

| Method | Emitted (MySQL) | PostgreSQL needs |
| --- | --- | --- |
| `GroupConcat` | `GROUP_CONCAT(x SEPARATOR "y")` | `STRING_AGG(x::text, 'y')` |
| `Concat` | `CONCAT(a,"y",b)` | `CONCAT(a,'y',b)` — double quotes are an identifier in PG |
| `Upsert` | `ON DUPLICATE KEY UPDATE c=VALUES(c)` | `ON CONFLICT(k) DO UPDATE SET c=EXCLUDED.c` |
| `UpdateWithJointOrFrom` | `UPDATE t INNER JOIN u ...` | `UPDATE t SET ... FROM u ...` |
| `IsDuplicateError` | asserts `*mysql.MySQLError` | `*pgconn.PgError`, SQLSTATE `23505` |

`IsDuplicateError` is the quiet one. A pgx error can never satisfy `err.(*mysql.MySQLError)`, so it returns `false` for every conflict and callers take the "not a duplicate" branch — no error, just wrong behaviour.

## Change

- Add `PostgreSQLDialect` implementing `SQLDialect`, plus `NewPostgreSQLDialect()`.
- Replace the hardcoded call with `newStorageDialect(driverName)`, which rejects unknown drivers with the same message `initDBDriver` already uses rather than silently defaulting to one engine.

Implementation notes:

- `STRING_AGG` requires an explicit separator, unlike `GROUP_CONCAT` which defaults to `,`. An empty separator falls back to `,` so the SQL keeps parsing and keeps MySQL's behaviour. A single-argument `STRING_AGG` does not parse; there is a test for this.
- Separators are single-quoted, since PostgreSQL reads a double-quoted token as an identifier rather than a string literal.
- `IsDuplicateError` uses `errors.As` because the pgx stdlib driver wraps `*pgconn.PgError` before it surfaces through `database/sql`. Tested with wrapped, bare, unrelated-SQLSTATE, non-PG, and nil errors.
- The SQLSTATE is a local `pgUniqueViolation` constant rather than adding a dependency on `jackc/pgerrcode` for one value.

## What this does *not* fix

This does not make the PostgreSQL overlay work, and it is not intended to. Still outstanding from #13956:

- **Item 1** — the storage call sites still build with squirrel's default `?` placeholders instead of the `$1` form the dialect in `client_manager/dialect.go` already defines.
- **Item 2** — identifier casing.
- The `Concat` call sites in `job_store.go:220` and `run_store.go:270,283,298` pass double-quoted SQL string literals (`` `"["` ``, `` `"]"` ``) which PostgreSQL reads as identifiers. That is caller-side and left alone here deliberately, rather than writing MySQL-only syntax into a new PostgreSQL dialect.

Those need the design decision discussed in https://github.com/kubeflow/pipelines/issues/13956#issuecomment-5223303540. This PR removes one independent layer: the stores now receive a dialect matching the engine they are talking to.

## Testing

10 new tests in `db_test.go` covering every `PostgreSQLDialect` method, both separator branches, both `Upsert` overwrite branches, and five `IsDuplicateError` cases. Plus a regression test asserting `MySQLDialect.IsDuplicateError` returns `false` for a PostgreSQL error, which is the bug being fixed.

```
ok  github.com/kubeflow/pipelines/backend/src/apiserver/storage         4.360s
ok  github.com/kubeflow/pipelines/backend/src/apiserver/client_manager  4.804s
```

Note for reproducing locally: the `storage` package needs `CGO_ENABLED=1`, since `SQLiteDialect.IsDuplicateError` uses `mattn/go-sqlite3`, which only exposes `sqlite3.Error` and `sqlite3.ErrConstraint` under cgo.

I have not run this against a live PostgreSQL instance — the remaining items above still prevent `ml-pipeline` from reaching Ready, so there is nothing end-to-end to verify against yet.

### DIFF
--- a/backend/src/apiserver/client_manager/client_manager.go
+++ b/backend/src/apiserver/client_manager/client_manager.go
@@ -376,7 +376,24 @@ func InitDBClient(initConnectionTimeout time.Duration) *storage.DB {
 	if err != nil {
 		glog.Fatalf("Failed to retrieve *sql.DB from gorm.DB. Error: %v", err)
 	}
-	return storage.NewDB(newdb, storage.NewMySQLDialect())
+	return storage.NewDB(newdb, newStorageDialect(driverName))
+}
+
+// newStorageDialect returns the storage-layer dialect matching driverName. The
+// storage dialect is distinct from the SQLDialect used by the migration flow
+// above: it covers the query-building differences the stores rely on
+// (GROUP_CONCAT vs STRING_AGG, ON DUPLICATE KEY vs ON CONFLICT, UPDATE JOIN vs
+// UPDATE FROM, and driver-specific duplicate-key detection).
+func newStorageDialect(driverName string) storage.SQLDialect {
+	switch driverName {
+	case "mysql":
+		return storage.NewMySQLDialect()
+	case "pgx":
+		return storage.NewPostgreSQLDialect()
+	default:
+		glog.Fatalf("Driver %v is not supported, use \"mysql\" for MySQL, or \"pgx\" for PostgreSQL", driverName)
+		return nil
+	}
 }
 
 // Initializes Database driver. Use `driverName` to indicate which type of DB to use:
--- a/backend/src/apiserver/storage/db.go
+++ b/backend/src/apiserver/storage/db.go
@@ -17,13 +17,19 @@ package storage
 import (
 	"bytes"
 	"database/sql"
+	"errors"
 	"fmt"
 	"strings"
 
 	"github.com/VividCortex/mysqlerr"
 	"github.com/go-sql-driver/mysql"
+	"github.com/jackc/pgx/v5/pgconn"
 	sqlite3 "github.com/mattn/go-sqlite3"
 )
+
+// pgUniqueViolation is the PostgreSQL SQLSTATE for a unique constraint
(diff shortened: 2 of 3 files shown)


hsinhoyeh:
Closing in favor of #12379, which has since merged. It introduces `common/sql/dialect` and selects the dialect from the configured driver in the client manager, so the driver-based dialect selection this PR added is now handled upstream.

Issue #10690, opened 2024-04-11
[feature] Pydantic 


mikeedjones:
### What feature would you like to see?

I would like to be able to typehint containerized and lightweight python components, and pipelines, using Pydantic [Models](https://docs.pydantic.dev/latest/concepts/models/) - and have the validation of those models occur automatically, making the validated models available inside the component code.
```python
from pydantic import BaseModel

class MyModel(BaseModel):
    foo: str

@component(target_image="myimage")
def my_function_one(my_data: MyModel):
    print(my_data.foo)
```
### What is the use case or pain point?

At the moment I'm passing complex hierarchical dictionaries as parameters to kubeflow using Dict[str, Any] and then validating those dictionaries inside the function code, as below

```python
from pydantic import BaseModel

class MyModel(BaseModel):
    foo: str

@component(target_image="myimage")
def my_function_one(my_data: Dict[str, Any]):
    from . import MyModel
    my_data_validated = MyModel.model_validate(my_data)
    print(my_data_validated.foo)
```

Where `myimage` has the base model available.

### Is there a workaround currently?

[This gist](https://gist.github.com/mikeedjones/7a504b2aa9dcc4a5ad97c219972b7428) shows my current attempt at a reusable workaround.

---

Love this idea? Give it a 👍.

rimolive:
Please join our next [Pipelines WG meeting](https://www.kubeflow.org/docs/about/community/#kubeflow-community-calendars) to discuss about this feature.

mikeedjones:
Think mine is more of a feature request for future versions. I'm not sure that your request is covered by what I had in mind :(

Sharkyii:
I want to take this up, plan:

- Detect Pydantic BaseModel type annotations in component signatures during compilation (component_factory.py's existing type-hint handling), and represent them as a STRUCT parameter in the pipeline spec, since a model serializes cleanly to/from a dict via model_dump()/model_validate().
- At runtime, have the executor call model_validate() on the raw value before handing it to the component function, and model_dump() on the way out for Pydantic-typed outputs.
- Start with flat models and expand to nested ones once the basic path works

/assign

PR #13994, opened 2026-08-08
feat(sdk): recognize pydantic BaseModel as a STRUCT parameter type

Sharkyii:

First slice toward #10690. Component type hints that are a `pydantic.BaseModel` subclass are now recognized by the compiler and treated as a STRUCT parameter (the same bucket a plain `Dict` already goes into), instead of falling through to being misclassified as an artifact type. This is compile-time detection only, via a new `type_utils.is_pydantic_basemodel_subclass()` helper (lazy-imports pydantic so it stays an optional dependency) used in `_annotation_to_type_struct()`. Runtime validation/serialization in the executor is a separate follow-up PR. Covered by new unit tests, and the full `sdk/python/kfp/dsl` and `sdk/python/kfp/compiler` suites pass locally (902 passed).

**Checklist:**

### DIFF
--- a/sdk/python/kfp/dsl/types/type_utils.py
+++ b/sdk/python/kfp/dsl/types/type_utils.py
@@ -577,6 +577,21 @@ def validate_bundled_artifact_type(type_: str) -> None:
     validate_schema_version(schema_version)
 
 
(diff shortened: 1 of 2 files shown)


Sharkyii:
Redirected to #13995

PR #13995, opened 2026-08-08
feat(sdk): add pydantic BaseModel support for component I/O

Sharkyii:

First real slice of #10690: native pydantic BaseModel support for component I/O. The compiler now recognizes a BaseModel type hint and treats it as a STRUCT parameter (the same bucket a plain `Dict` already goes into), via a new `type_utils.is_pydantic_basemodel_subclass()` helper that lazy-imports pydantic so it stays an optional dependency. At runtime, the executor calls `model_validate()` on BaseModel-typed inputs and `model_dump()` on BaseModel-typed returns. The ephemeral component file also auto-imports any custom BaseModel class referenced in a component's signature, mirroring the existing mechanism KFP already has for custom Artifact subclasses — without it, a BaseModel type hint compiles fine but fails with a `NameError` at runtime, since the generated file only ever imported custom Artifact classes, never parameter types.

Verified with a real end-to-end `local.SubprocessRunner` run, not just unit tests: one component builds and returns a `Person` model, a second component consumes it, and the round trip produces the correctly validated object and correctly serialized output. Full `sdk/python/kfp/dsl` and `sdk/python/kfp/compiler` suites pass locally (912 passed).

**Checklist:**

### DIFF
--- a/sdk/python/kfp/dsl/types/custom_artifact_types.py
+++ b/sdk/python/kfp/dsl/types/custom_artifact_types.py
@@ -148,6 +148,66 @@ def validate_pydantic_basemodel_is_importable(basemodel_cls: type) -> None:
             '`base_image`.')
 
 
+def _maybe_unwrap_optional_subscript(
+        annotation: Union[ast.expr, None]) -> Union[ast.expr, None]:
+    """Unwraps an `Optional[X]`/`X | None`-shaped AST node to the inner `X`
+    node, mirroring the Optional-stripping already applied to the resolved type
+    in get_param_to_pydantic_basemodel_class.
+
+    Pydantic models are type-hinted directly (e.g. `person: Person`),
+    unlike custom artifact types, which are always wrapped (e.g.
+    `Input[VertexDataset]`), so this only needs to handle the
+    `Optional[...]` sugar form, not a generic subscript slice.
+    """
+    if isinstance(annotation, ast.Subscript) and isinstance(
+            annotation.slice, (ast.Name, ast.Attribute)):
+        return annotation.slice
+    return annotation
+
+
+def get_pydantic_basemodel_base_symbol_for_parameter(func: Callable,
+                                                     arg_name: str) -> str:
+    """Gets the symbol required for a pydantic.BaseModel parameter type
+    annotation to be referenced correctly in the generated component file."""
+    module_node = ast.parse(
+        component_factory._get_function_source_definition(func))
+    args = module_node.body[0].args.args
+    args = {arg.arg: arg for arg in args}
+    annotation = _maybe_unwrap_optional_subscript(args[arg_name].annotation)
+    return traverse_ast_node_values_to_get_id(annotation)
+
+
+def get_pydantic_basemodel_base_symbol_for_return(func: Callable,
+                                                  return_name: str) -> str:
+    """Gets the symbol required for a pydantic.BaseModel return type annotation
+    (including a typing.NamedTuple field) to be referenced correctly in the
+    generated component file."""
+    module_node = ast.parse(
+        component_factory._get_function_source_definition(func))
+    return_ann = module_node.body[0].returns
+
+    if return_name == RETURN_PREFIX:
+        annotation = _maybe_unwrap_optional_subscript(return_ann)
+        if isinstance(annotation, (ast.Name, ast.Attribute)):
+            return traverse_ast_node_values_to_get_id(annotation)
+    elif isinstance(return_ann, ast.Call):
+        func_node = return_ann.func
+        # handles NamedTuple and typing.NamedTuple
+        if (isinstance(func_node, ast.Attribute) and
+                func_node.value.id == 'typing' and
+                func_node.attr == 'NamedTuple') or (isinstance(
+                    func_node, ast.Name) and func_node.id == 'NamedTuple'):
+            nt_field_list = return_ann.args[1].elts
+            for el in nt_field_list:
+                if f'{RETURN_PREFIX}{el.elts[0].s}' == return_name:
+                    annotation = _maybe_unwrap_optional_subscript(el.elts[1])
+                    return traverse_ast_node_values_to_get_id(annotation)
+
+    raise TypeError(
+        f"Unexpected pydantic.BaseModel return type annotation '{ast.dump(return_ann)}' "
+        f'for function {func.__name__}.')
+
+
 def get_pydantic_basemodel_import_items_from_function(
         func: Callable) -> List[str]:
     """Gets the fully qualified name of the symbol that must be imported for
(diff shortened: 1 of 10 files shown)


HarshPopat23 (changes requested):
I reviewed the live diff, all seven changed files I would leave four blocking points :

### **Blocking 1 :** 
`is_pydantic_basemodel_subclass()` recognizes `BaseModel `subclasses from both Pydantic v1 and v2, but these calls are v2-only. With Pydantic 1.x, compilation succeeds and execution fails because `model_validate()` and `model_dump()` do not exist.

Could we either:

• explicitly require/gate Pydantic >=2 and return a clear error for older versions, or
• provide compatibility paths using `parse_obj()`/`dict() `for v1?

Please also add tests for the supported version contract. Since Pydantic remains optional, the expected compile-time and runtime dependency requirements should be documented.

### **Blocking 2:** 
this misses `Optional[BaseModel] `inputs. `get_name_to_specs()` and Executor.execute() both call `maybe_strip_optional_from_annotation(),` so an optional model is recognized as a parameter and validated at runtime. This import collector examines the raw `Optional[MyModel]`, does not recognize it as a `BaseModel `subclass, and omits the `MyModel `import from the generated source.

Could we normalize the annotation here using the same helper before checking it? Please add an execution-level test for:

```python
def component(data: Optional[MyModel] = None):
    ...
```

### **Blocking/design clarification 3:**
generating `from <module> import <Model>` only works when that module is also installed or otherwise available in the component runtime. It does not support the motivating example in #10690 where the model is defined alongside the pipeline/component code, because lightweight components do not automatically copy that source module into the container. Classes compiled from a script may also generate `from __main__ import MyModel.`

The `SubprocessRunner `verification shares the developer’s filesystem and installed packages, so it does not reproduce this isolation boundary.

Could we define the supported model-distribution contract and test it in an isolated environment? Possible contracts include:

• bundle the model definition and its required source;
• require the model to come from an installable package included through packages_to_install or the base image; or
• reject non-importable/local models with a clear compilation error.

The chosen behavior and the requirement to install Pydantic in the runtime should be documented.

### **Blocking 4:** 
the current serialization is not guaranteed to round-trip through `model_validate()` for the same model. For fields with aliases, `model_dump(mode='json')` emits field names by default, while validation commonly expects aliases.

For example:
```python
class Person(BaseModel):
    name: str = Field(alias='personName')
```
`Person.model_validate(person.model_dump(mode='json'))` fails under the default configuration, while dumping with by_alias=True produces the accepted wire key.

Could we define a stable wire-format policy-ikely involving `by_alias=True`, and `possibly round_trip=True`-and add a producer-to-consumer test using aliases? Custom serialization and validation aliases should also be considered when defining the contract.

Sharkyii:
@HarshPopat23 PTAL

HarshPopat23:
Tanks for addressing the previous feedback. I reviewed commit 8268ba0.

This PR is valuable, but not urgent or required for existing KFP behavior.

let's wait for maintainers point of view @andreyvelich @robert-bell .

zazulam:
@Sharkyii can you rebase?

zazulam (commented):
Thanks for the thorough implementation and tests. looks good, but I think a few things should be addressed before merge.

- `Optional[BaseModel]` outputs are not normalized during compilation or execution, so they can produce an unsupported output type or `Unknown return type`

- Generated imports do not preserve annotation bindings such as `models.Person` or `PersonAlias` causing `NameError` in lightweight component source

- `RootModel` subclasses are always declared as `STRUCT` although scalar and list root models serialize to incompatible values

- `model_dump(by_alias=True)` does not round-trip when `serialization_alias` and `validation_alias` differ

- Pydantic tests are conditionally skipped, but CI does not explicitly install Pydantic v2, so the new behavior is not guaranteed to be exercised

Could these cases be handled and covered by tests before merge?

sdk/python/kfp/dsl/types/custom_artifact_types.py
```diff
@@ -76,6 +76,181 @@ def get_param_to_custom_artifact_class(func: Callable) -> Dict[str, type]:
...
+
+    Pydantic models are type-hinted directly (e.g. `person: Person`),
+    unlike custom artifact types, which are always wrapped (e.g.
+    `Input[VertexDataset]`), so this only needs to handle the
+    `Optional[...]` sugar form, not a generic subscript slice.
+    """
+    if isinstance(annotation, ast.Subscript) and isinstance(
+            annotation.slice, (ast.Name, ast.Attribute)):
+        return annotation.slice
+    return annotation
```
zazulam: This helper's docstring claims `X | None` support, but a PEP 604 union is an `ast.BinOp`, so it currently falls through unchanged. Please unwrap both optional syntaxes:

```suggestion
    if isinstance(annotation, ast.Subscript):
        return annotation.slice
    if isinstance(annotation, ast.BinOp) and isinstance(annotation.op,
                                                         ast.BitOr):
        if isinstance(annotation.left,
                      ast.Constant) and annotation.left.value is None:
            return annotation.right
        if isinstance(annotation.right,
                      ast.Constant) and annotation.right.value is None:
            return annotation.left
    return annotation
```

This needs to be paired with the same runtime normalization in `maybe_strip_optional_from_annotation()`, using `typing.get_origin(annotation) in {Union, types.UnionType}` and selecting the non-`None` argument. Otherwise compilation and executor dispatch still reject `Person | None` before reaching this AST code.

sdk/python/kfp/dsl/types/custom_artifact_types.py
```diff
@@ -76,6 +76,181 @@ def get_param_to_custom_artifact_class(func: Callable) -> Dict[str, type]:
...
+
+
+def get_pydantic_basemodel_base_symbol_for_parameter(func: Callable,
+                                                     arg_name: str) -> str:
+    """Gets the symbol required for a pydantic.BaseModel parameter type
+    annotation to be referenced correctly in the generated component file."""
+    module_node = ast.parse(
+        component_factory._get_function_source_definition(func))
+    args = module_node.body[0].args.args
+    args = {arg.arg: arg for arg in args}
```
zazulam: `inspect.signature()` discovers keyword-only and positional-only model parameters, but this AST lookup only includes ordinary positional arguments. That makes `def component(*, person: Person)` fail with `KeyError`. Include every annotated parameter kind:

```suggestion
    function_args = module_node.body[0].args
    args = (function_args.posonlyargs + function_args.args +
            function_args.kwonlyargs)
    args = {arg.arg: arg for arg in args}
```

zazulam (commented):
I reproduced these cases against the current PR head with Pydantic v2. The inline suggestions below show possible fixes; please add focused tests for each case as well.

sdk/python/kfp/dsl/types/type_utils.py
```diff
@@ -577,6 +577,133 @@ def validate_bundled_artifact_type(type_: str) -> None:
...
+                f"('{serialization_key}') will not be accepted by "
+                f"model_validate() on the way back in, which expects "
+                f"'{validation_key}'{populate_by_name_hint}. KFP serializes "
+                "component outputs with model_dump(by_alias=True) and "
+                'deserializes inputs with model_validate(), so this field '
+                'would be silently dropped passing between components. Use '
+                'the same value for validation_alias and serialization_alias '
+                "(or a single `alias`), or set "
+                'model_config = ConfigDict(populate_by_name=True) and give '
+                "serialization_alias the field's own name.")
```
zazulam: I think KFP should avoid reimplementing Pydantic alias-resolution semantics here. `AliasChoices`, `AliasPath`, and validation config flags are Pydantic-owned behavior that may evolve independently, making this validator a long-term compatibility and maintenance burden.

Please remove this validator and its call from `_pydantic_basemodel_to_type_struct()`. KFP can define a minimal transport contract—serialize outputs with `model_dump(mode="json", by_alias=True)` and validate inputs with `model_validate()`—while documenting that model-specific alias compatibility remains the component author’s responsibility.

```suggestion
```

This keeps KFP responsible only for transporting the JSON-compatible value rather than wrapping or duplicating another package’s validation policy.

sdk/python/kfp/dsl/types/custom_artifact_types.py
Sharkyii: The runtime helper has the same gap as the AST one. `Person | None `gets rejected during compilation, so it never reaches this code, will fix both with your diff and add a test for each layer.

sdk/python/kfp/dsl/types/custom_artifact_types.py
Sharkyii: The AST lookup only checked `args.args` - willll apply your fix and add a keyword-only Pydantic test.

sdk/python/kfp/dsl/types/type_utils.py
Sharkyii: Fair point. Will remove the validator and call site.

zazulam (commented):
/lgtm
/approve

PR #14004, opened 2026-08-09
Fix (backend) : compileCmdAndArgs placeholder bug

Daniel-1600:
the `getplaceholders` error was never checked it was immediately overwritten by the error from the following `protojson.Marshal` call, so a failure in `getPlaceholders` would go unnoticed as long as Marshal succeeded.

the second change `compiledCmd` was computed from cmd before the placeholder-substitution loop ran. That loop then reassigned cmd (not compiledCmd), so the substituted result was discarded and the function returned a command string with only {{$}} replaced , any other placeholders (e.g. {{output_path}}) were left unsubstituted in cmd, while the same placeholders in args were correctly replaced. This meant unresolved template placeholders could leak into the actual executed command.

**Checklist:**

### DIFF
--- a/backend/src/v2/component/launcher_v2.go
+++ b/backend/src/v2/component/launcher_v2.go
@@ -1050,6 +1050,9 @@ func fetchNonDefaultBuckets(
 
 func compileCmdAndArgs(executorInput *pipelinespec.ExecutorInput, cmd string, args []string) (string, []string, error) {
 	placeholders, err := getPlaceholders(executorInput)
+	if err != nil {
+		return "", nil, fmt.Errorf("failed to get placeholders: %w", err)
+	}
 
 	executorInputJSON, err := protojson.Marshal(executorInput)
 	if err != nil {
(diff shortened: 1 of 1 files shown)


dobbydobap:
Both of these are already in #13984, which covers the same two defects plus the ordering bug #13977 actually reports.

That third one is the one worth looking at. `compileCmdAndArgs` substitutes in a single pass over `placeholders`, and Go randomises map iteration order, so when one parameter's value contains another placeholder, the nested one only resolves if its key happens to come later in that same pass. That's the intermittent unresolved `{{$.inputs.parameters[...]}}` the issue describes. Fixing the `cmd` assignment doesn't reach it, since the single-pass loop is still there.

#13984 is at review stage with a bounded repeat-until-stable loop, and a regression test that loops 100 times because a single run passes roughly half the time.

droctothorpe:
@Daniel-1600: Hey, thank you for your PR! As you are not a member of Kubeflow organization, your PR must link to an issue with 'ready' label. Please read our [CONTRIBUTING.md](https://github.com/kubeflow/pipelines/blob/master/CONTRIBUTING.md), closing this PR

PR #14006, opened 2026-08-09
ci: derive generator tool versions from modules

jeffspahr:
## Summary

- derive the grpc-gateway, protobuf, protoc-gen-go-grpc, and Argo CLI versions from their canonical Go module metadata
- track go-swagger in a dedicated Dependabot-managed tools module and remove dead or duplicated version pins
- run generated-file validation only for relevant generator changes while preserving the required check name

## Why

Generator versions could drift from the modules they serve because Dependabot updated `go.mod` without updating arbitrary Dockerfile environment variables. This makes the module manifests the source of truth and teaches CI which grouped Go updates actually require regeneration.

Standalone release downloads such as protoc and googleapis remain explicit pins because the configured Dependabot ecosystems cannot track them reliably.

## Verification

- built `backend/api/Dockerfile` successfully
- checked all installed generator versions in the resulting image
- regenerated the v2beta1 API and confirmed no generated-file diff
- ran all 158 CI script unit tests successfully
- ran `go mod verify` for the tools module
- parsed the modified workflows and checked Makefile dependency wiring
- ran `git diff --check`

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -17,6 +17,7 @@ updates:
     directories:
       - "/"
       - "/api"
+      - "/backend/api/tools"
       - "/kubernetes_platform"
       - "/third_party/ml-metadata"
       - "/test/tools/project-cleaner"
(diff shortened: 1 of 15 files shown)


zazulam (commented):
/lgtm
/approve

PR #14011, opened 2026-08-11
fix: preserve duplicate metrics in Compare view

### LINKED ISSUE #13988: [frontend] MultiRunMetricsCompareProps drops duplicate metric names without warning in Compare view
In CompareUtils.ts (multiRunMetricsCompareProps), the pipeline comparison view aggregates scalar metrics across runs. However, when looking up metrics for each run, it matches only the first metric matching a given metric name and silently ignores any subsequent metrics with the same name within the same run.

This behavior causes data loss in the UI when a run logs multiple metrics with duplicate names (e.g., intermediate epoch metrics or repeated evaluations).

An inline comment at CompareUtils.ts:72 explicitly acknowledges this limitation:

// TODO: handle duplicate metric names

Steps to Reproduce
  1. Execute a pipeline run that emits multiple metrics with identical names (e.g., accuracy logged twice across different steps/tasks).

  2. Select the run alongside another run in the Kubeflow Pipelines UI and navigate to the Compare Runs view.

  3. Observe the Metrics comparison table.

Expected Behavior
Either:

    1.Display all metric values associated with duplicate names (e.g., by qualifying them with task/node IDs or indexing them), OR

    2. Apply an explicit aggregation rule (e.g., last reported value, average, min/max) and clearly indicate in the UI that values were aggregated.

Actual Behavior
Only the first encountered metric value is displayed. Later values with the same metric name are silently dropped without any warning or indication to the user.

Affected Code
    Implementation: src/components/CompareUtils.ts (around line 65–72)

   Unit Tests: src/components/CompareUtils.test.ts (currently missing test cases for runs emitting duplicate metric names)

Sumitha007:
Fixes #13988

## What does this PR do?

Preserves duplicate metric values in the Compare Runs view instead of silently dropping subsequent metrics with the same name.

When a run contains multiple metrics with the same name, each occurrence is now displayed separately using labels such as `metric` and `metric (2)`.

## Changes

- Preserve all duplicate metric occurrences in `multiRunMetricsCompareProps`.
- Generate additional rows based on the maximum number of occurrences in any run.
- Add regression tests for duplicate metrics within a run and across multiple runs.

## Testing

- `npx vitest run src/lib/CompareUtils.test.ts`
- 25 tests passed.

### DIFF
--- a/frontend/src/lib/CompareUtils.test.ts
+++ b/frontend/src/lib/CompareUtils.test.ts
@@ -274,6 +274,49 @@ describe('CompareUtils', () => {
         yLabels: ['some-metric', 'another-metric'],
       });
     });
+
+    it('returns all values when a run has duplicate metric names', () => {
+      const runs = [
+        {
+          metrics: [
+            { name: 'some-metric', number_value: 0.33 },
+            { name: 'some-metric', number_value: 0.66 },
+          ],
+          name: 'run1',
+        } as ApiRun,
+      ];
+
+      expect(CompareUtils.multiRunMetricsCompareProps(runs)).toEqual({
+        rows: [['0.330'], ['0.660']],
+        xLabels: ['run1'],
+        yLabels: ['some-metric', 'some-metric (2)'],
+      });
+    });
+
+    it('returns duplicate metric values based on the maximum occurrences in a run', () => {
+      const runs = [
+        {
+          metrics: [
+            { name: 'some-metric', number_value: 0.33 },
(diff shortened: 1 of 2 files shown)


droctothorpe:
Hi @Sumitha007, thank you for your PR! The linked issue (#13988) must be labeled `ready` before your contribution can be reviewed. Please check our [CONTRIBUTING.md](https://github.com/kubeflow/pipelines/blob/master/CONTRIBUTING.md) for details.

PR #14017, opened 2026-08-11
ci: auto-regenerate Dependabot generated files

jeffspahr:
## Summary

- detect module-only Dependabot updates that change a tracked Go generator
- capture an allowlisted generated-file patch in the existing read-only validation workflow
- apply and push that patch from a separate default-branch `workflow_run` after independently validating the PR, artifact, paths, and current head

## Why

Generator versions now derive from Go module manifests, but a Dependabot generator bump still leaves checked-in outputs stale and requires a human companion commit. This completes the automation loop so those PRs can become self-contained.

Generation remains in the unprivileged `pull_request` workflow. The artifact consumer loads its policy and validation script from the default branch, checks out the Dependabot head with non-persisted read credentials, and exposes the SSH credential only inside the final push step. It validates the original PR file scope, head SHA, artifact metadata, claimed and applied file inventories, and the PR head again immediately before pushing a DCO-signed commit.

The generated-output allowlist is shared with repository metadata through the `dependabot-auto-generated` attributes in `.gitattributes`. Patch capture includes tracked modifications, deletions, and new untracked outputs. Workflow reruns overwrite the same-name artifact safely, and empty `workflow_run.pull_requests` payloads fall back to commit-associated PR lookup.

The push uses the repository's existing `SSH_PRIVATE_KEY` and `SSH_KNOWN_HOSTS` secrets rather than `GITHUB_TOKEN`, because pushes made with `GITHUB_TOKEN` do not retrigger the normal pull-request workflow matrix. The SSH identity must have permission to update same-repository Dependabot branches.

## Verification

- `python3 -m unittest discover -v -p '*_test.py'` — 165 tests passed
- `python3 -m unittest -v generated_files_dependency_change_test.py` — 14 tests passed
- Actionlint v1.7.7 on all modified workflows
- Ruby YAML parsing on all modified workflows
- `git diff --check`
- confirmed the live #14016 module diff is classified as eligible for auto-regeneration

Because `workflow_run` uses the workflow definition from the default branch, the privileged push path can only be exercised end-to-end after this PR merges. Synchronizing #14016 is the intended first production verification.


zazulam:
@jeffspahr can you rebase

PR #13481, opened 2026-06-03
components: Remove the components directory except google-cloud

mprahl:
**Description of your changes:**

The KFP components now live at:
https://github.com/kubeflow/pipelines-components

This deletion is in accordance to and several discussions on the KFP community calls:
https://github.com/kubeflow/pipelines/discussions/13051

This also removes the docs/code-tree directory that seems to have been accidentally committed and is now stale.

**Checklist:**
- [x] You have [signed off your commits](https://www.kubeflow.org/docs/about/contributing/#sign-off-your-commits)
- [x] The title for your pull request (PR) should follow our title convention. [Learn more about the pull request title convention used in this repository](https://github.com/kubeflow/pipelines/blob/master/CONTRIBUTING.md#pull-request-title-convention).

PR #14019, opened 2026-08-11
ci: add full Dependabot coverage

jeffspahr:
## Summary

- add weekly Dependabot coverage for every supported dependency ecosystem present in the repository
- recursively discover all current and future Go modules and npm projects
- cover fifteen maintained pip project directories
- cover GitHub workflow actions, reusable composite actions, and pre-commit hooks
- preserve the existing repository-wide Go module and Docker/Kubernetes image coverage
- group npm, pip, and GitHub Actions updates by dependency name across monorepo directories
- keep each ecosystem's version-update queue bounded at ten open pull requests

## Why

The existing configuration intentionally covered only Go modules and container images. As a result, npm and pip vulnerabilities detected by OSV Scanner did not have corresponding Dependabot version-update coverage. GitHub Actions and pre-commit dependencies were also unmanaged.

Generated v1beta1 and v2beta1 Python API-client requirements remain excluded because they express generated compatibility minimums and must be changed through API regeneration rather than direct dependency PRs.

## Coverage guard

The new stdlib-only test compares `.github/dependabot.yml` with the repository's actual Go, npm, pip, GitHub Actions, and pre-commit manifest inventory. It also verifies recursive `**/*` discovery for Go and npm. The CI scripts workflow now runs when the Dependabot configuration or any relevant manifest pattern changes, so a newly added project cannot silently fall outside update coverage.

## Validation

- `cd .github/resources/scripts && python3 -m unittest discover -p '*_test.py'`
- focused Dependabot and generated-file configuration tests
- parsed `.github/dependabot.yml` and `ci-scripts-tests.yml` as YAML
- Python bytecode compilation for the new coverage test
- `git diff --check`


.github/dependabot.yml
```diff
@@ -62,3 +56,67 @@ updates:
...
+    # Generated API clients are excluded: their requirements express generated
+    # compatibility minimums and must be changed through API regeneration.
+    directories:
+      - "/.github/resources/scripts/kfp-readiness"
+      - "/api/v2alpha1/python"
+      - "/backend"
+      - "/backend/metadata_writer"
+      - "/backend/src/apiserver/visualization"
+      - "/backend/src/v2/test"
+      - "/components/google-cloud"
```
zazulam, line 81: not sure if this directory is still being supported or if the owners ever migrated the code to other repo they manage.

zazulam (commented):
I'm fine with this, hopefully we don't get spammed with more dependabot PRs 😃

.github/dependabot.yml
jeffspahr: Still supported here — it was explicitly exempted from the `components/` migration. From #13051:

> **Important exception:** Google-maintained components are **not** part of this May 11 removal. They will migrate on a different timeline, and we will share more details separately.

#13481 removed everything under `components/` except `google-cloud`, and `kubeflow/pipelines-components` has no GCPC content today (code search returns no hits for `google_cloud_pipeline_components`). PyPI `google-cloud-pipeline-components` is still at 2.22.0, matching `google_cloud_pipeline_components/version.py` in this repo, with its `Source` URL pointing back here.

Two caveats worth flagging: there's no CI covering the directory anymore (#13481 also dropped `gcpc-modules-tests.yml`), and `setup.py` is the only manifest, with wide ranges (`kfp>=2.6.0,<3.0.0`, `google-cloud-aiplatform>=1.14.0,<2`) — so in practice this entry should rarely open a PR.

Either way the list is mechanically derived: `dependabot_config_test.py` asserts exact equality between the configured pip directories and every non-generated Python manifest in the tree, so once GCPC does migrate out the test fails until the entry is removed.

PR #14021, opened 2026-08-12
fix(ci): enforce repository Python syntax

jeffspahr:
## Summary

- fix the invalid `ddef` declaration and unbound `kfp` reference in the directory-output sample
- stop v1beta1 Python client generation from emitting unused API/model test stubs, remove the generated stubs, and remove their orphaned tox configuration
- add a dedicated repository-wide syntax workflow pinned to Python 3.9
- scan only tracked Python files using Git's NUL-delimited output, with filesystem-safe path decoding and handled parser errors
- make generated-file validation run when the Python client generator script changes
- replace implementation-coupled assertions with behavior-focused syntax discovery and error-handling coverage

## Why

CodeQL reported four Python extraction failures. One was a real sample typo. The other three were invalid OpenAPI-generated v1beta1 test stubs where schema prose was emitted as Python identifiers. Those stubs are excluded from the published package and are not executed by the repository.

Removing the unused generated test directory at its source keeps production client code in scan scope without hand-editing generated files or broadly excluding paths from CodeQL. The dedicated syntax gate prevents future committed Python parse errors from reaching the default branch without running the full CI-script test suite for every Python change.

## Verification

- Python 3.9 syntax checker: 1,233 tracked files passed
- Python 3.9 checker unit tests: 5 passed
- `.github/resources/scripts` suite: 178 passed
- JUnit-summary action suite: 7 passed
- v1beta1 package regeneration: no production client changes; generated tests and tox files remained absent
- v2beta1 package regeneration: no changes; generated tests and tox files remained present
- `bash -n backend/api/build_kfp_server_api_python_package.sh`
- `git diff --check`


zazulam (commented):
Given python 3.10 EoL is coming in October, I was assuming we would drop support for 3.9 then. I'm not sure if hardening validation around its support is something we want to move forward with. I'm open to hear what others think.

jeffspahr:
> Given python 3.10 EoL is coming in October, I was assuming we would drop support for 3.9 then. I'm not sure if hardening validation around its support is something we want to move forward with. I'm open to hear what others think.

It'll be pretty trivial to pull the 3.9 part of this out in October.  I'm happy to remove it from this PR now too if preferred.

The bigger driver for this PR is these two issues causes the osvscanner integration to fail on parts of the repo:

- fix the invalid ddef declaration and unbound kfp reference in the directory-output sample
- stop v1beta1 Python client generation from emitting unused API/model test stubs, remove the generated stubs, and remove their orphaned tox configuration

zazulam:
> > Given python 3.10 EoL is coming in October, I was assuming we would drop support for 3.9 then. I'm not sure if hardening validation around its support is something we want to move forward with. I'm open to hear what others think.
> 
> It'll be pretty trivial to pull the 3.9 part of this out in October. I'm happy to remove it from this PR now too if preferred.
> 
> The bigger driver for this PR is these two issues causes the osvscanner integration to fail on parts of the repo:
> 
> * fix the invalid ddef declaration and unbound kfp reference in the directory-output sample
> * stop v1beta1 Python client generation from emitting unused API/model test stubs, remove the generated stubs, and remove their orphaned tox configuration

@jeffspahr sounds good, we should then create a issue to track the python 3.10 EoL support and we can discuss in the maintainer sync on when should CI drop support for python 3.9.

/approve

PR #14022, opened 2026-08-12
chore: updated algorithm - AI Issue Triage - issue text samples by section

modichika:
Author: @alyssacgoins 
Co-author: @modichika, @zazulam, @sduvvuri1603 

Note: gh models have been deprecated so this pr highlights the algorithm we need to solve the tokens problem.

Instead of loading full GH issues into the prompt context (which is more expensive, token-wise) which is failing:
https://github.com/kubeflow/pipelines/pull/13372 in the main kubeflow/pipelines repo.
Depending on the type of issue identified, the script can compare each section of the issue to the the sample text.

**The Following Updated Workflow follows this algorithm:**

1. Script parses issue title to identify the issue type (bug, chore, or feature), and area (frontend, backend, sdk or testing), assuming the default format of <type>(<area>): <title contents>. (This is the pre-filled title format when you open a new issue, and it’s the format most users follow. This is also a way to limit AI slop - if a contributor does not care enough to follow the format, we don’t need to look at the issue). If the title doesn’t match the formatting, the script adds a comment on the issue - “issue title must follow the correct format”
2. Script validates each subsection of the issue body via individual, smaller prompts. Each issue type has the following subsections:
Bug: environment, steps to reproduce, expected result, materials & reference
Chore: description
Feature:  feature, use case, current workaround
	
Our current approach is failing because our prompts are too large and we’re hitting token limits too quickly. Dividing prompts by issue subsection will mitigate this risk. In addition, we spend a lot of tokens retrieving the text from the linked sample issues. We think it would make more sense to compare the direct text of each issue subsection against sample, approved text.
If a prompt fails to execute, it’s also worth adding a retry mechanism, potentially with exponential backoff.

3. Finally, the script outputs its analysis (no change here).

### EXAMPLES
- ### Bug
**Environment:**
A successful environment subsection should contain the KFP version number and/or context on how KFP was deployed.

**Steps to Reproduce:**
A successful environment subsection contains enough information where a mid-level engineer familiar with Kubeflow Pipelines can reasonably be expected to reproduce the issue within 30 minutes of trying. Characteristics of this section include specific inputs and/or function calls.

**Expected Result:**
A successful expected result subsection includes a program output that KFP is designed to produce, given a certain input.

**Materials & Reference:**
A successful materials & reference subsection includes ___.

- ### Chore

**Description:**
A successful chore description subsection includes enough detail on a small-scoped task that a mid-level engineer with moderate Kubeflow Pipelines experience can understand the task and create a fix within 1 working day.

**Checklist:**


Priyanshu-u07:
Hi @modichika, just wanted to flag something I ran into while reading this. I don't think the current failures are token limits. 
I checked a few recent runs of the workflow on master ([one here](https://github.com/kubeflow/pipelines/actions/runs/31583845356/job/94072953673)) and they're all returning:

```
Error: unexpected response from the server: 410 Gone
{"error":{"code":"github_models_retirement_brownout","message":"GitHub Models is temporarily unavailable as part of a scheduled retirement brownout."}}
```
GitHub Models was fully retired on 2026-07-30, the changelog says the playground, catalog and inference API are "no longer available to any customer, including existing customers with active usage": https://github.blog/changelog/2026-07-30-github-models-is-now-retired/
Two things that made me fairly confident it isn't size-related: it reproduces on short issues (#14023 has all four template sections filled and is nowhere near a token ceiling, same 410), and the run log shows `Models: read` is granted, so it doesn't look like permissions either.

Since this PR still uses `gh models run`, I think the retries and per-section prompts won't change the outcome on their own, though I may be missing something. Thanks!

modichika:
Hey @Priyanshu-u07 me and the other authors know about this we are working on it we just wrote this pr for kfp's community meeting to highlight the algorithm.

Thank you for mentioning it I forgot to write it in the pr's description!

.github/workflows/ai_analyzer.yml
```diff
@@ -3,35 +3,42 @@ on:
...
-      models: read
     steps:
-      - name: Checkout
+      - name: Checkout Code
         uses: actions/checkout@v4
 
+      - name: Set up Python
+        uses: actions/setup-python@v5
+        with:
+          python-version: '3.10'
```
zazulam: nit: I would bump to `3.13`

.github/scripts/ai_triage.py
```diff
@@ -0,0 +1,116 @@
...
+            with open(github_output_path, "a") as f:
+                f.write("analysis<<EOF\n")
+                f.write(error_msg + "\n")
+                f.write("EOF\n")
+        sys.exit(0)
+
+    print(f"✅ Title validated successfully. Type: {issue_type}, Area: {issue_area}")
+
+    system_instructions = """You are an expert open-source maintainer for Kubeflow Pipelines.
+You are an expert open-source maintainer for Kubeflow Pipelines.
```
zazulam: Looks like a duplication here

.github/workflows/ai_analyzer.yml
modichika: yeah done!

.github/scripts/ai_triage.py
modichika: Thanks for highlighting this, resolved it.

zazulam:
great work! Let's keep an eye on new issues for verifying the triage
/lgtm
/approve

PR #14026, opened 2026-08-12
fix(sdk): preserve pip_trusted_hosts when pip_index_urls is not set. Fixes #13806

### LINKED ISSUE #13806: [sdk] pip_trusted_hosts is silently ignored when pip_index_urls is not provided
### Environment

*  KFP version: 2.17.0 (master)
*  KFP SDK version: 2.17.0 (master)
*  All dependencies version:
```
kfp                      2.17.0
kfp-pipeline-spec        0.3.0
kfp-server-api           2.0.5
```

### Steps to reproduce

When authoring a lightweight Python component, if a user specifies `pip_trusted_hosts` without `pip_index_urls`, the `--trusted-host` argument is silently dropped from the generated `pip install` command. 

This causes failures when users configure a private PyPI mirror via a global `pip.conf` (e.g. baked into a custom `base_image`) and only need to pass `--trusted-host` for their corporate proxy/mirror.

You can reproduce this behavior directly in a Python shell by calling the underlying utility function that generates the pip command options:

```python
from kfp.dsl.component_factory import make_index_url_options

# Expected: '--trusted-host pypi.org '
# Actual: ''
print(repr(make_index_url_options(pip_index_urls=None, pip_trusted_hosts=['pypi.org'])))
```

### Expected result

The function should return the `--trusted-host` flags even when `--index-url` is not explicitly provided. The generated `pip install` command inside the container should include `--trusted-host pypi.org`.

### Materials and Reference

The bug originates in `sdk/python/kfp/dsl/component_factory.py`, within the `make_index_url_options` function. Due to an early return statement, the function exits immediately if `pip_index_urls` is falsy (which it is by default), completely ignoring the `pip_trusted_hosts` argument:

```python
# sdk/python/kfp/dsl/component_factory.py:102
def make_index_url_options(pip_index_urls: Optional[List[str]],
                           pip_trusted_hosts: Optional[List[str]]) -> str:
    # ... 
    if not pip_index_urls:
        return ''  # <--- BUG: Early return drops pip_trusted_hosts completely
```

This contradicts the intention of recent PR #13780, which aimed to make `--trusted-host` an explicit opt-in via `pip_trusted_hosts`. 

**Proposed Solution**
Refactor the function to handle both lists independently:
```python
def make_index_url_options(pip_index_urls: Optional[List[str]],
                           pip_trusted_hosts: Optional[List[str]]) -> str:
    options = []
    
    if pip_index_urls:
        options.append(f'--index-url {pip_index_urls[0]}')
        options.extend(f'--extra-index-url {url}' for url in pip_index_urls[1:])
        
    if pip_trusted_hosts:
        options.extend(f'--trusted-host {host}' for host in pip_trusted_hosts)
        
    if not options:
        return ''
        
    return ' '.join(options) + ' '
```

---

shivv23:

Fixes #13806.

When a lightweight Python component specifies `pip_trusted_hosts` without `pip_index_urls`, the `--trusted-host` arguments were silently dropped from the generated `pip install` command. `make_index_url_options` had an early `return ''` whenever `pip_index_urls` was empty or `None`, which discarded `pip_trusted_hosts` entirely.

This causes failures for users who configure a private PyPI mirror via a global `pip.conf` (e.g., baked into a custom `base_image`) and only need `--trusted-host` for their corporate proxy/mirror.

**Changes:**
- `sdk/python/kfp/dsl/component_factory.py`: refactor `make_index_url_options` to build `--index-url`/`--extra-index-url` and `--trusted-host` options independently, so `--trusted-host` flags are emitted even when `pip_index_urls` is not provided.
- `sdk/python/kfp/dsl/component_factory_test.py`: add `TestMakeIndexUrlOptions` unit tests covering the option combinations and an integration test for `_get_packages_to_install_command` with `pip_trusted_hosts` only.

**Reproducer (before/after):**
```python
from kfp.dsl.component_factory import make_index_url_options

# Before: ''
# After:  '--trusted-host pypi.org '
print(repr(make_index_url_options(pip_index_urls=None, pip_trusted_hosts=['pypi.org'])))
```

**Checklist:**

### DIFF
--- a/sdk/python/kfp/dsl/component_factory.py
+++ b/sdk/python/kfp/dsl/component_factory.py
@@ -92,30 +92,32 @@ def make_index_url_options(pip_index_urls: Optional[List[str]],
 
     Returns:
         str:
-            - An empty string if pip_index_urls is empty or None.
+            - An empty string if both pip_index_urls and pip_trusted_hosts are
+              empty or None.
             - '--index-url url ' if pip_index_urls contains 1 URL.
             - The above followed by '--extra-index-url url ' for each additional URL in pip_index_urls
             if pip_index_urls contains more than 1 URL.
             - If pip_trusted_hosts is None or an empty List:
                 - No --trusted-host information will be added.
             - If pip_trusted_hosts contains any hosts:
-                - The above followed by '--trusted-host host ' for each host in pip_trusted_hosts.
+                - '--trusted-host host ' for each host in pip_trusted_hosts, added
+                  regardless of whether pip_index_urls is provided.
     Note:
-        In case pip_index_urls is not empty, the returned string will contain a space at the end.
+        If any options are generated, the returned string will contain a space at the end.
     """
-    if not pip_index_urls:
-        return ''
+    options = []
 
-    index_url = pip_index_urls[0]
-    extra_index_urls = pip_index_urls[1:]
-
-    options = [f'--index-url {index_url}']
-    options.extend(f'--extra-index-url {extra_index_url}'
-                   for extra_index_url in extra_index_urls)
+    if pip_index_urls:
+        options.append(f'--index-url {pip_index_urls[0]}')
+        options.extend(f'--extra-index-url {extra_index_url}'
+                       for extra_index_url in pip_index_urls[1:])
 
     if pip_trusted_hosts:
         options.extend(f'--trusted-host {trusted_host}'
                        for trusted_host in pip_trusted_hosts)
+
+    if not options:
+        return ''
 
     return ' '.join(options) + ' '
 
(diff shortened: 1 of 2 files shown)


dobbydobap:
Heads up, this looks like it overlaps #13807, which fixes the same issue and predates it by about three weeks. #13806 is assigned to vishalmore90, who opened that PR the same day they took the issue.

The implementations come out the same: build `options` independently for index urls and trusted hosts, then return '' only when nothing was added. I read both at their current heads and the behaviour matches, including the empty-list cases. Test coverage is equivalent too, the difference is that you split the cases into named methods on `TestMakeIndexUrlOptions` while #13807 keeps them as assertions inside one method. The split is nicer when something fails, since one bad case there currently masks the ones after it, so that part might be worth carrying over regardless of which PR lands.

For context, #13807 has been open since 24 July with no review and three requests for one, and is still `needs-ok-to-test` and behind master, so the hold-up looks procedural rather than technical.

PR #14027, opened 2026-08-12
fix(backend): handle Kubernetes watch errors safely. Fixes #14025

### LINKED ISSUE #14025: [backend] Cache watcher panics when processing Kubernetes watch.Error events
###Steps to reproduce

In backend/src/cache/server/watcher.go, WatchPods performs an unchecked type assertion to *corev1.Pod before checking whether the event is a watch.Error:

```go 
     for event := range watcher.ResultChan() {
    pod := reflect.ValueOf(event.Object).Interface().(*corev1.Pod)
    if event.Type == watch.Error {
        continue
    }

    // ...
}
```
Kubernetes watch error events can contain a *metav1.Status instead of a *corev1.Pod.

I reproduced the behavior using watch.NewRaceFreeFake():

```go
w := watch.NewRaceFreeFake()

go func() {
    defer w.Stop()

    w.Action(watch.Added, &corev1.Pod{})
    w.Action(watch.Error, &metav1.Status{
        Status:  "Failure",
        Reason:  metav1.StatusReasonGone,
        Message: "too old resource version",
    })
}()

for event := range w.ResultChan() {
    fmt.Printf("Received event: %s, object: %T\n", event.Type, event.Object)

    // Same ordering as watcher.go.
    pod := event.Object.(*corev1.Pod)

    if event.Type == watch.Error {
        continue
    }

    fmt.Printf("Pod: %s\n", pod.Name)
}
```
###Output:

Received event: ADDED, object: *v1.Pod
Pod:
Received event: ERROR, object: *v1.Status
panic: interface conversion: runtime.Object is *v1.Status, not *v1.Pod

The panic occurs at:

pod := event.Object.(*corev1.Pod)

before the watch.Error check is reached.

###Expected result

The watcher should handle watch.Error events without panicking.

The event type should be checked before treating event.Object as a Pod, and the object should be type-asserted safely. For example:
```go
for event := range watcher.ResultChan() {
    if event.Type == watch.Error {
        continue
    }

    pod, ok := event.Object.(*corev1.Pod)
    if !ok {
        log.Printf("Unexpected watch object type: %T", event.Object)
        continue
    }

    // ...
}
```
A regression test should also verify that a watch.Error event containing a *metav1.Status does not cause a panic.

###Materials and Reference

-The affected code is backend/src/cache/server/watcher.go.

-The issue is caused by performing the unchecked Pod type assertion before checking event.Type.

-This appears related to the previously reported issue #3632, [Cache] Watcher needs to swallow periodic  errors. The current implementation still performs the unchecked type assertion before checking for watch.Error.

jahnavigajjala-3:

The cache server's pod watcher performs an unchecked type assertion of every Kubernetes watch event object to `*corev1.Pod` before checking the event type.

Kubernetes `watch.Error` events can contain a `*metav1.Status` object instead of a`*corev1.Pod`. This causes the cache watcher to panic when processing an error event.

This change:

- Checks for `watch.Error` events before accessing the event object.
- Replaces the unchecked type assertion with a safe type assertion.
- Removes the unnecessary `reflect` usage.
- Logs and skips unexpected Kubernetes watch object types instead of panicking.

The behavior was reproduced locally using `watch.NewRaceFreeFake()` with a `watch.Error` event containing a `*metav1.Status`.

Fixes #14025

**Checklist:**
- [x] The title for your pull request (PR) follows the repository title convention.


dobbydobap:
Fix looks right. Dropping `reflect` is a good call too, that call was only doing what a plain type assertion does, and nothing else in the file uses it so the import removal is safe.

Two things:

Worth adding a test. The package has `admission_test.go` and `mutation_test.go` but no `watcher_test.go`, and the repro you wrote in #14025 with `watch.NewRaceFreeFake` is most of one already.

There's also a second nil path in the same function this doesn't cover:

```go
watcher, err := k8sCore.PodClient(namespaceToWatch).Watch(ctx, listOptions)
if err != nil {
    log.Printf("%s", "Watcher error:"+err.Error())
}
for event := range watcher.ResultChan() {
```

Watch returns a nil `watch.Interface` on error and the log doesn't stop the fall-through, so `ResultChan` panics. I ran `WatchPods` against the repo's own `FakeClientManager` (`FakePodClient.Watch` returns `(nil, nil)`) and it panics with "invalid memory address or nil pointer dereference" before reading any event. Same class as what you're fixing here, so it might belong in this PR, or a separate issue if you'd rather keep this one tight.

jahnavigajjala-3:
I’ll add watcher_test.go with regression coverage for the watch.Error event containing a *metav1.Status, and also handle the case where Watch() returns an error with a nil watcher to prevent the ResultChan() panic. I’ll push the changes to this PR.

dobbydobap:
Nil guard and the type assertion both look right, and thanks for adding `watcher_test.go`.

One thing on the error path. `main.go:92` starts this as `go server.WatchPods(...)` once at startup and nothing restarts it, so `return` ends pod watching for the life of the process. A transient API error or a token refresh would silently disable caching with just that one log line. Before this PR the same path hit the nil watcher and panicked, which at least crashed the process and let Kubernetes restart it.

The outer `for` loop is already there to re-establish the watch, and your fake calling `runtime.Goexit()` on the second `Watch` shows it. Something like this keeps clean shutdown without losing the retry:

```go
if err != nil {
    log.Printf("Watcher error: %v", err)
    if ctx.Err() != nil {
        return
    }
    time.Sleep(retryDelay)
    continue
}
```

Small related thing: the test says "WatchPods failed to exit after context cancellation", but `WatchPods` never reads `ctx` today, it exits because the fake calls `runtime.Goexit()`. With the check above that message would actually be true.

jahnavigajjala-3:
thanks for pointing this out. You're right, returning from watchPods on a watch error would permanently stop the watcher since it is only started once. I'll update the error path to retry while respecting context cancellation, and I'll adjust the regression test accordingly.

droctothorpe:
Hi @jahnavigajjala-3, thank you for your PR! The linked issue (#14025) must be labeled `ready` before your contribution can be reviewed. Please check our [CONTRIBUTING.md](https://github.com/kubeflow/pipelines/blob/master/CONTRIBUTING.md) for details.

PR #14030, opened 2026-08-13
ci: make image artifacts architecture-specific

iknowmaaz-mz:
## Summary

- make CI image artifacts architecture-specific for amd64 and arm64
- build runtime base images on architecture-specific runners
- update Kind image loading to consume architecture-specific artifacts
- make image artifact waiting architecture-aware
- update runtime base image cache keys and pruning for each architecture

## Validation

- `git diff --check`
- YAML parsing passed for modified workflows/action
- `bash -n` passed for modified shell scripts
- no legacy image artifact paths remain
- architecture-specific artifact references verified

## Testing

GitHub Actions will provide the end-to-end CI validation for amd64 and arm64.


iknowmaaz-mz:
Hi @droctothorpe @nsingla,

Just following up on this PR as well. This change makes the CI image artifacts architecture-specific for AMD64 and ARM64 and updates the related runtime image and Kind image-loading logic.

I would appreciate it if you could review it when you have time. Please let me know if any changes or additional validation are required.

Thank you for your time and review.

jeffspahr (changes requested):
Thanks for continuing to iterate here. The current revision is a substantial improvement: both architecture builds complete, control-plane images are retagged to the manifest-compatible shared tag, Modelcar and Docker-mirror handling are restored, and the script test job now passes.

I'm requesting changes because the PR still has deterministic correctness problems and, despite building ARM images, does not exercise KFP on ARM.

Blocking issues:

- `.github/workflows/runtime-base-images.yml` computes the producer fingerprint from five inputs, including the Modelcar Dockerfile, but the cache-pruning job recomputes it from only four inputs. The pruner therefore looks for a cache key that cannot equal the key just saved. Because pruning is skipped on pull requests, the current checks do not exercise this failure.
- `.github/resources/scripts/wait-for-image-artifacts.sh` qualifies expected artifact names by architecture, but removes that architecture when matching producer jobs and returns the first image-name match. With amd64 and arm64 producers for every image, an ARM waiter can make readiness or failure decisions using the amd64 job.
- The same waiter accepts an unsuffixed legacy artifact as satisfying either architecture, while `.github/actions/deploy/action.yml` only opens architecture-qualified directories and tar names. The waiter can therefore report readiness for an artifact that deployment cannot load. The existing waiter tests still provide only unsuffixed names, so their success does not validate the new path.
- The reusable build workflow is invoked by nine test workflows. The observed revision created 118 ARM runner jobs and uploaded roughly 66 GB of artifact material, while every deployment and functional-test consumer still ran on amd64. At least one amd64 consumer downloaded both architecture sets. This multiplies build and artifact cost without providing ARM runtime coverage.
- `backend/metadata_writer/Dockerfile` now downloads Bazelisk without checksum verification, clones mutable tags, patches upstream `WORKSPACE` text during the build, and compiles with `-march=armv8.3-a`. That is not a generic `linux/arm64` baseline, is not runtime-tested here, remains excluded from the release ARM matrix, and still does not solve the amd64-only MLMD server required by the manifests.
- The branch also carries unrelated action-version regressions (`actions/checkout` v7 to v6, `docker/build-push-action` v7 to v6, and `actions/cache` v6 to v5). Those should be removed when the branch is brought current with master.

I recommend restructuring this as follows:

1. Build each architecture once in a centralized, path-filtered producer workflow and publish architecture-qualified CI artifacts.
2. Keep the broad functional matrix on amd64. Add one native ARM installation smoke test that runs a representative pipeline, thereby exercising both `kfp-driver` and `kfp-launcher`; broader ARM coverage can run nightly.
3. Make image plus architecture the producer identity throughout the waiter and downloader. Consumers should download only their architecture. Remove the legacy fallback, or normalize legacy artifacts into the exact layout required by deployment before reporting readiness.
4. Use one shared fingerprint definition for production and pruning, with regression coverage for both architectures and concurrent producer jobs.
5. Separate the metadata-writer/MLMD experiment from the CI artifact-plumbing change. If it remains in scope, pin source revisions immutably, verify downloaded tooling, and target the project's documented minimum ARM ISA rather than ARMv8.3.

Successfully compiling the ARM images—including `kfp-driver`, `kfp-launcher`, and `metadata-writer`—is valuable evidence, but it is not equivalent to ARM runtime coverage. I would be happy to re-review a revision that centralizes the builds and includes a native ARM pipeline smoke test.

jeffspahr:
State update since my review: #13986 has now merged to `master`, while `release-2.18` was cut immediately before that merge. Please keep this PR targeting `master`; I do not recommend retargeting or backporting it to `release-2.18`.

This makes one part of my September 4 review obsolete for a revision rebased onto current master: metadata-writer, metadata-envoy, and the deployed MLMD server have been removed from the active installation and image inventories. Please drop the `backend/metadata_writer/Dockerfile` workaround and obsolete matrix entries rather than resolving them back into the tree.

The other requested changes still apply: add a native ARM installation/pipeline consumer; build ARM once rather than through every amd64 test workflow; make producer selection and downloads architecture-aware; remove or correctly normalize the legacy artifact fallback; use identical runtime producer/pruner fingerprints; and remove the action-version regressions during the rebase.

The current head now conflicts with master, and its historical checks do not validate the post-MLMD tree. The clean next step is either a rewrite of this PR from current master or a focused replacement PR using the current ten-image inventory, a centralized/path-filtered ARM producer, architecture-scoped downloads, and one native ARM smoke pipeline. I am happy to re-review either form.

droctothorpe:
@iknowmaaz-mz: Hey, thank you for your PR! As you are not a member of Kubeflow organization, your PR must link to an issue with 'ready' label. Please read our [CONTRIBUTING.md](https://github.com/kubeflow/pipelines/blob/master/CONTRIBUTING.md), closing this PR

PR #14038, opened 2026-08-14
fix(backend): guard toModelRun against a malformed PipelineInfo in the pipeline spec

Atishyy27:
## Summary

`toModelRun` reads `specMap["PipelineInfo"]` directly out of the client-submitted `PipelineSpec` (a `structpb.Struct`, i.e. arbitrary JSON from the API request), then does two unguarded type assertions:

```go
if pName, ok := pv.(map[string]interface{})["Name"]; ok {
    resources := common.ParseResourceIdsFromFullName(pName.(string))
```

A v2 run submitted with a `PipelineInfo` that isn't an object, or whose `Name` isn't a string, panics the API server instead of returning an error, since neither assertion checks its `ok` result.

## Change

Guard both assertions with their two-value form, so a malformed `PipelineInfo`/`Name` is skipped the same way a *missing* one already is, instead of panicking:

```go
pvMap, pvIsMap := pv.(map[string]interface{})
if pName, ok := pvMap["Name"].(string); pvIsMap && ok {
    resources := common.ParseResourceIdsFromFullName(pName)
```

## Test plan

- Added `TestToModelRun_MalformedPipelineInfoDoesNotPanic` in `api_converter_test.go`, covering both a non-object `PipelineInfo` and a non-string `Name`.
- Could not run this package's Go test binary directly on this machine — `backend/src/apiserver/storage` depends on a CGO sqlite3 driver, and this box has no C compiler (`gcc` not found), so the build fails on an unrelated package before the test can run.
- Instead verified the fix at the logic level: extracted both the original and fixed type-assertion patterns into a standalone, dependency-free reproduction and ran it directly. Original panics on both malformed inputs; fixed does not, and still returns the correct name for a well-formed `PipelineInfo`. Noting this honestly rather than claiming a full test run I didn't do.
- `gofmt`-checked both changed files (on an LF-normalized copy — this repo checks out CRLF via `core.autocrlf`, which `gofmt` doesn't understand natively).

## Notes

Happy to have a maintainer (or CI) run the actual test if the logic-level verification isn't sufficient — I just don't have a working CGO toolchain here to confirm it myself.

### DIFF
--- a/backend/src/apiserver/server/api_converter.go
+++ b/backend/src/apiserver/server/api_converter.go
@@ -1331,8 +1331,9 @@ func toModelRun(r interface{}) (*model.Run, error) {
 
 		specMap := apiRunV2.GetPipelineSpec().AsMap()
 		if pv, ok := specMap["PipelineInfo"]; ok {
-			if pName, ok := pv.(map[string]interface{})["Name"]; ok {
-				resources := common.ParseResourceIdsFromFullName(pName.(string))
+			pvMap, pvIsMap := pv.(map[string]interface{})
+			if pName, ok := pvMap["Name"].(string); pvIsMap && ok {
+				resources := common.ParseResourceIdsFromFullName(pName)
 				if namespace == "" {
 					namespace = resources["Namespace"]
 				}
--- a/backend/src/apiserver/server/api_converter_test.go
+++ b/backend/src/apiserver/server/api_converter_test.go
@@ -4344,6 +4344,38 @@ func TestToModelRun(t *testing.T) {
 	}
 }
 
+func TestToModelRun_MalformedPipelineInfoDoesNotPanic(t *testing.T) {
+	// Regression: toModelRun reads specMap["PipelineInfo"] straight out of a
+	// client-submitted PipelineSpec (a structpb.Struct, i.e. arbitrary JSON), then
+	// asserts it's a map and asserts its "Name" is a string, both without checking
+	// ok. A pipeline spec where PipelineInfo isn't an object, or has a non-string
+	// Name, used to panic instead of returning an error.
+	tests := []struct {
+		name         string
+		pipelineInfo interface{}
+	}{
+		{"PipelineInfo is not an object", "not-an-object"},
+		{"PipelineInfo.Name is not a string", map[string]interface{}{"Name": 123}},
+	}
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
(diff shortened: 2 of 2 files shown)


dobbydobap:
Fix looks right. Worth noting for anyone reading it later that the reordering is safe on purpose: `pvMap["Name"]` runs before `pvIsMap` is checked, but a failed type assertion leaves `pvMap` nil, and indexing a nil map returns the zero value rather than panicking, so the `ok` on the second assertion is what actually guards it. `pvIsMap` is redundant given that, since `ok` can only be true when `pvMap` is a real map holding a string.

Both test cases genuinely panic on master, so the regression test earns its place.

The same shape appears seven more times in this file, and I think it is the more interesting half of what you found. `case X, *X:` followed by `s.(X)` does not narrow the variable, because Go only binds the case type when the case lists exactly one type. So the pointer half of every one of these panics if it is ever hit:

```
2432/2433  case string, *string:                     -> s.(string)
2444/2445  case apiv1beta1.Run_StorageState, *...    -> s.(apiv1beta1.Run_StorageState)
2446/2447  case apiv1beta1.Experiment_StorageState   -> s.(...)
2448/2449  case apiv2beta1.Run_StorageState          -> s.(...)
2450/2451  case apiv2beta1.Experiment_StorageState   -> s.(...)
2535/2536  case string, *string:                     -> s.(string)
2537/2538  case apiv2beta1.RuntimeState, *...        -> s.(...)
```

Minimal repro:

```
string     -> out="SUCCEEDED" err=<nil>
*string    -> out="" err=PANIC: interface conversion: interface {} is *string, not string
```

To be clear it is latent rather than live: every caller of `toModelStorageState` and `toModelRuntimeState` passes a proto enum or a plain string out of a `_name[...]` lookup, so nothing reaches the pointer branch today. The `*X` cases are effectively dead, and they read as though pointers are handled when they are not. Probably a separate change from this one, but it is the same defect you are fixing here.

Atishyy27:
Thanks, that reading is right and better stated than my description was. Happy to do the other seven in a follow-up PR once this one lands, so the regression test and the fix stay reviewable together here.

droctothorpe:
@Atishyy27: Hey, thank you for your PR! As you are not a member of Kubeflow organization, your PR must link to an issue with 'ready' label. Please read our [CONTRIBUTING.md](https://github.com/kubeflow/pipelines/blob/master/CONTRIBUTING.md), closing this PR

Issue #11930, opened 2025-05-25
[feature] Allow List[PipelineTask.output] or List[Component.output] for Fan-In Pipelines


marcocamilo:
### Feature Area

/area backend
/area sdk
/area components

### What feature would you like to see?

I would like to support passing a list of outputs from _multiple components_ as a single structured input (e.g., a List[Artifact]) to another component in Kubeflow Pipelines, as opposed to the current option of simply collecting the outputs of a single component iterating over multiple input parameters (e.g. `dsl.ParallelFor` / `dsl.Collected`).

This is conceptually different from what `dsl.ParallelFor` and `dsl.Collected` currently offer—those focus on collecting outputs from one component iterated over multiple inputs. The request here is to enable collection of outputs from different components, each invoked independently, into a single input for downstream components.

### What is the use case or pain point?

In many ML workflows—especially ensemble methods, boosting pipelines, or model selection tasks—multiple models are trained using different algorithms or configurations, but share the same downstream evaluation process.

For example, I may train several models using different components like `train_svc_op`, `train_xgb_op`, and `train_lr_op`, each producing the same type of `Model` artifact. These need to be evaluated together by a single component that accepts a `List[Model]`.

Currently, Kubeflow Pipelines does not provide a clean way to aggregate these outputs. Instead, I have to:

- Manually define a fixed number of input arguments in the evaluation component.
- Explicitly pass each model artifact individually.
- Lose flexibility in dynamically scaling the number of models.

This leads to verbose, brittle code that is difficult to generalize or maintain, especially when the number of models changes.

### Is there a workaround currently?

A workaround exists but is awkward and non-scalable. I must:

- Declare a fixed number of inputs in the downstream component.
- Pass each model artifact as a separate argument.
- Use `Input[Model]`, `Input[Model2]`, etc., making dynamic workflows difficult or impossible.

Example (conceptual psuedocode):

```
@component
def evaluate_models(
    model_1: Input[Model],
    model_2: Input[Model],
    model_3: Input[Model],
    test_set: Input[Dataset],
):
    # Evaluation logic that manually handles multiple inputs
    pass

@pipeline()
def ml_pipeline():
    svc_model = train_svc_op(
        train_set=prep_data_op.outputs["train_set"],
        val_set=prep_data_op.outputs["val_set"],
        mlflow_experiment_name=experiment_name,
    )

    xgb_model = train_xgb_op(
        train_set=prep_data_op.outputs["train_set"],
[...]

PR #14045, opened 2026-08-15
docs: KEP for static fan-in of task outputs

zazulam:

Adds KEP-11930 proposing explicit static fan-in of independently produced task outputs. The proposal recommends extending `dsl.Collected` while preserving existing `ParallelFor` collection behavior, and covers the required pipeline specification, SDK, backend, and KFP Local changes for KFP v3.

Part of #11930.

**Checklist:**
- [x] The title for your pull request (PR) follows the repository title convention.


proposals/11930-static-fan-in/README.md
```diff
@@ -0,0 +1,292 @@
...
+
+The single-channel form must retain its existing meaning for backward
+compatibility. Wrapping one channel in a sequence explicitly requests static
+fan-in.
+
+Static collection members must:
+
+- be pipeline channels;
+- all be parameters or all be artifacts; and
+- have compatible channel types.
```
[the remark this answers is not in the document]
zazulam: Addressed in 0303455ab. The KEP now limits static fan-in members to scalar channels and requires pipeline validation to reject nested list variants and references to list-valued outputs or component inputs, including directly submitted pipeline specifications.

proposals/11930-static-fan-in/README.md
```diff
@@ -0,0 +1,292 @@
...
+
+KFP Local resolves the new task-input variants using the same ordering as the
+remote driver. Both parameter and artifact static fan-in must work with local
+pipeline execution.
+
+### Frontend Considerations
+
+No frontend changes are required. Static fan-in only changes task-to-task
+wiring in the compiled pipeline. It does not add a pipeline-level runtime input
+or change the run creation experience.
```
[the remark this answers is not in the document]
zazulam: Addressed in 0303455ab. The frontend section now requires regenerating the TypeScript pipeline-spec bindings, parsing each list item in StaticFlow, preserving artifact-node edges, and adding parameter/artifact visualization tests.

proposals/11930-static-fan-in/README.md
```diff
@@ -0,0 +1,307 @@
...
+| `dsl.Collected([channel])` | One-element static fan-in |
+
+The single-channel form must retain its existing meaning for backward
+compatibility. Wrapping one channel in a sequence explicitly requests static
+fan-in.
+
+Static collection members must:
+
+- be pipeline channels;
+- represent scalar values rather than list-valued parameter or artifact
```
jeffspahr, line 152: **[P1] Define scalarity at the consumer's control-flow scope**

A producer output inside `dsl.ParallelFor` is still a scalar-typed SDK channel, but it has runtime multiplicity when referenced outside that loop. This rule therefore does not prevent static fan-in from becoming implicit loop flattening.

Please require each member to be independently legal as a scalar input at the consumer's scope: producer and consumer in the same iteration and outer-to-inner broadcast are valid; inner-to-outer and sibling-loop references are rejected; and the existing `If`/`ExitHandler` escape restrictions continue to apply. The test plan should cover those cases, including preservation of per-item `LoopArgumentVariable` selectors.

proposals/11930-static-fan-in/README.md
```diff
@@ -0,0 +1,307 @@
...
+
+  oneof kind {
+    // Existing kinds omitted.
+    ArtifactListSpec artifact_list = 6;
+  }
+}
+
+message InputParameterSpec {
+  message ParameterListSpec {
+    repeated InputParameterSpec items = 1;
```
jeffspahr, line 203: **[P1] Restrict recursive list items to supported scalar references**

Embedding the complete `InputParameterSpec` here also admits `runtime_value` (including an already-list-valued constant), `task_final_status`, another `parameter_list`, and `parameter_expression_selector`; the outer list spec can carry a selector as well. The public DSL contract only allows pipeline channels, and the existing scalar paths do not recursively infer all of those dependencies or apply item selectors. A directly submitted spec can therefore use shapes whose ordering and evaluation semantics are undefined.

Please either introduce non-recursive scalar item messages restricted to task-output and component-input references, or explicitly enumerate/reject every unsupported kind and selector location and define selector evaluation order. This needs direct-IR validation tests, not only SDK-constructor tests.

proposals/11930-static-fan-in/README.md
```diff
@@ -0,0 +1,307 @@
...
+
+  oneof kind {
+    // Existing kinds omitted.
+    ParameterListSpec parameter_list = 6;
+  }
+}
+```
+
+Reusing the existing input reference types preserves order and supports channels
+surfaced through nested DAG boundaries. Although the recursive protobuf shape
```
jeffspahr, line 214: **[P1] Define nested artifact fan-in against the post-MLMD execution path**

This promise requires artifact count, order, and duplicates to survive a graph-component input. The legacy path cannot currently provide that invariant: [`GetInputArtifactsByExecutionID`](https://github.com/kubeflow/pipelines/blob/bba33cdaf8b2696eee5db2850ea0aa1c1b92c2a4/backend/src/v2/metadata/client.go#L1150-L1190) assigns `inputs[name]` a new singleton for each same-name artifact, so only one member is retained.

Since MLMD is on the removal path, I would not add indexed-list persistence there. Please instead define the rollout contract: preserve members individually through each legacy boundary as a minimal transition, defer nested artifact fan-in until the replacement path preserves ordered duplicates, or make this topology an initial non-goal. A nested remote regression using `[a, b, a]` should run against the intended replacement path.

proposals/11930-static-fan-in/README.md
```diff
@@ -0,0 +1,307 @@
...
+can represent a list inside another list, nested lists are not valid for this
+feature. Pipeline validation rejects nested list variants and list items that
+reference list-valued task outputs or component inputs. This validation applies
+to directly submitted pipeline specifications as well as SDK-compiled
```
jeffspahr, line 218: **[P1] Validate the collection against the destination input contract**

The listed checks constrain item sources, but a directly submitted spec must also require that the destination exists, that `parameter_list` binds to `ParameterType_LIST`, that `artifact_list` binds to `is_artifact_list = true`, and that artifact member schemas are compatible with the destination. Otherwise the current driver can [convert a list bound to a `STRING` input into JSON text](https://github.com/kubeflow/pipelines/blob/bba33cdaf8b2696eee5db2850ea0aa1c1b92c2a4/backend/src/v2/driver/resolve.go#L202-L245), while the scalar artifact executor path can retain only the first artifact.

Please add destination existence, cardinality, source homogeneity, and artifact-type compatibility to a shared semantic validator, state whether it runs at upload or run compilation, and test missing, scalar, and incompatible destinations. For parameters, the KEP should acknowledge that IR erases the element type of `List[T]`, so backend validation can enforce source homogeneity but cannot recover `T`.

jeffspahr (commented):
Thanks for writing this KEP. Static fan-in fills a real SDK gap, and the explicit ordered-list API is a sensible direction. I found four behavioral contracts that need to be made precise before implementation, plus one frontend edge-identity issue. The inline comments cover control-flow scope, recursive item semantics, direct-IR destination validation, nested artifact behavior during the post-MLMD transition, and visualization coverage.

The main concern is that the proposal currently allows some specifications and topologies whose runtime meaning differs across the SDK, backend, Local, and nested graph execution. Tightening those contracts in the KEP and test plan should keep the implementation coordinated across all of those layers.

proposals/11930-static-fan-in/README.md
```diff
@@ -0,0 +1,307 @@
...
+KFP Local resolves the new task-input variants using the same ordering as the
+remote driver. Both parameter and artifact static fan-in must work with local
+pipeline execution.
+
+### Frontend Considerations
+
+The generated TypeScript pipeline-spec bindings must be regenerated for the new
+input variants. Static graph parsing must inspect each parameter-list and
+artifact-list item so every producer is connected to the consumer. Artifact
+members retain the existing producer-to-artifact-to-consumer visualization.
```
jeffspahr, line 240: **[P2] Make artifact edge IDs unique per producer and list position**

Traversing every item is not sufficient with the current edge-key contract. [`getArtifactToTaskEdgeKey`](https://github.com/kubeflow/pipelines/blob/bba33cdaf8b2696eee5db2850ea0aa1c1b92c2a4/frontend/src/lib/v2/StaticFlow.ts#L524-L527) uses only the output artifact key and consumer task, omitting the producer. Independent trainers commonly expose the same key such as `model`, so their edges collide; repeated members make that ambiguity even clearer.

Please require the producer task and list position in the artifact-input edge ID, and add a frontend test with multiple producers sharing one output key plus a repeated member, asserting distinct IDs and every expected source.

PR #14052, opened 2026-08-15
feat: require namespace for pipeline lookup in multi user mode

JerT33:

Currently, `GetPipelineByNameAndNamespace` allows namespace to be optional and falls back to kubeflow installation namespace. Normally this is fine, however in multi user environments this could allow user to receive pipeline information from an external namespace than what a user may have access to

This PR simply blocks pipeline lookup in multi user environments if a namespace is not provided, instead of falling back to installation namespace

Notes:
- Single-user mode is unchanged
- Shared SQL pipelines and the SDK's get_pipeline_id (which uses filtered ListPipelines) are unaffected

**Checklist:**

### DIFF
--- a/backend/src/apiserver/storage/pipeline_store_kubernetes_test.go
+++ b/backend/src/apiserver/storage/pipeline_store_kubernetes_test.go
@@ -235,6 +235,40 @@ func TestGetK8sPipeline_NotFoundError(t *testing.T) {
 	require.NotNil(t, err)
 }
 
+func TestGetK8sPipelineByNameAndNamespace_SingleUserDefaultsToPodNamespace(t *testing.T) {
+	podNamespace := viper.Get("POD_NAMESPACE")
+	viper.Set("POD_NAMESPACE", "Test")
+	defer viper.Set("POD_NAMESPACE", podNamespace)
+
+	store := NewPipelineStoreKubernetes(getClient())
+
+	// Single-user mode has one namespace, so the fallback stays correct.
+	pipeline, err := store.GetPipelineByNameAndNamespace("test-pipeline-3", "")
+	require.NoError(t, err)
+	assert.Equal(t, "test-pipeline-3", pipeline.Name)
+}
+
+func TestGetK8sPipelineByNameAndNamespace_MultiUserRequiresNamespace(t *testing.T) {
+	podNamespace := viper.Get("POD_NAMESPACE")
+	viper.Set("POD_NAMESPACE", "Test")
+	defer viper.Set("POD_NAMESPACE", podNamespace)
+
+	multiUser := viper.Get("MULTIUSER")
+	viper.Set("MULTIUSER", "true")
+	defer viper.Set("MULTIUSER", multiUser)
+
+	store := NewPipelineStoreKubernetes(getClient())
+
+	// "test-pipeline-3" lives in the pod namespace, which must not be reachable without naming it.
+	_, err := store.GetPipelineByNameAndNamespace("test-pipeline-3", "")
+	require.Error(t, err)
+	assert.Equal(t, codes.InvalidArgument, err.(*util.UserError).ExternalStatusCode())
+
+	pipeline, err := store.GetPipelineByNameAndNamespace("test-pipeline-3", "Test")
+	require.NoError(t, err)
+	assert.Equal(t, "test-pipeline-3", pipeline.Name)
+}
+
 func TestCreateK8sPipeline(t *testing.T) {
 	podNamespace := viper.Get("POD_NAMESPACE")
 	viper.Set("POD_NAMESPACE", "Test")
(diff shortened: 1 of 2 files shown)


backend/src/apiserver/storage/pipeline_store_kubernetes_test.go
```diff
@@ -235,6 +235,40 @@ func TestGetK8sPipeline_NotFoundError(t *testing.T) {
 	require.NotNil(t, err)
 }
 
+func TestGetK8sPipelineByNameAndNamespace_SingleUserDefaultsToPodNamespace(t *testing.T) {
+	podNamespace := viper.Get("POD_NAMESPACE")
```
jeffspahr, line 242: Please explicitly set and restore `MULTIUSER=false` in this single-user test, just as the multi-user case sets/restores true. Currently this case relies on the ambient global Viper setting, so it does not independently establish the mode named by the regression. The production fallback itself looks correct.

jeffspahr (commented):
@JerT33 We are considering this fix for the 2.18 release if you can get it across the line. Please let us know what remains before marking it ready for review; the intended path would be merge to master, then a targeted backport to `release-2.18` (#14409 / #14421).

Reviewed at `9b5fe711110a578963570673979f4be28aa11293`. The production change looks appropriate for the Kubernetes-backed store: reject an omitted namespace before using the installation namespace in multi-user mode, while preserving the single-user fallback. This matters because the API authorizes the requested namespace before the store resolves it, and an empty requested namespace follows the shared-read path. Explicit namespace lookups continue through the existing namespace authorization check. I did not find a production-code defect in this narrow patch.

Before release inclusion, I recommend:

- Add an API-level regression with multi-user enabled and `REQUIRE_NAMESPACE_FOR_PIPELINES=false`: an omitted namespace must not return an installation-namespace pipeline; explicit authorized namespace succeeds and unauthorized namespace is denied. The store tests cover the new guard but do not exercise that authorization boundary.
- Make the single-user regression explicitly set and restore `MULTIUSER=false` (inline note).
- Include a short compatibility note: multi-user Kubernetes-backed by-name callers must supply a namespace. This is not a blanket change to shared SQL pipelines or to SDK `get_pipeline_id`, which uses filtered ListPipelines. Include sample loading/restart behavior in the final release validation.

Scope note: the SQL V1/V2 by-name methods have a separate omitted-namespace scoping concern found during 2.18 review. That is tracked in #14421 and should be addressed in a separate focused fix; I am not asking this Kubernetes-only PR to absorb it.

Validation: full `go test ./backend/src/apiserver/storage -count=1` passes at the exact PR head. The patch also applies cleanly to release commit `75f3f4bbd9affbe6c3c80d6be3465ddeeb90face`, where both added lookup tests pass. This is not full API/cluster acceptance or approval to release.

JerT33:
@jeffspahr thanks for letting me know, i wasn't aware this was targeted for 2.18. I have updated PR with the feedback provided. Thanks!

backend/src/apiserver/server/pipeline_server_namespace_test.go
```diff
@@ -0,0 +1,151 @@
...
+	c := ctrlfake.NewClientBuilder().
+		WithScheme(scheme).
+		WithObjects(installPipeline, tenantPipeline).
+		Build()
+	return c, c
+}
+
+func multiUserPipelineServer(t *testing.T, authorized bool) (*PipelineServer, func()) {
+	t.Helper()
+	initEnvVars()
```
jeffspahr, line 73: [P3] Restore the installation namespace after initialization

`initEnvVars()` overwrites `POD_NAMESPACE=kubeflow` with `ns1`, while the fixture pipeline lives in `kubeflow`. Running the omitted-namespace regression against the old storage implementation therefore returns `NotFound` rather than exposing the seeded pipeline. The `InvalidArgument` assertion still detects removal of the guard, so this is non-blocking, but the intended unauthorized-read scenario is not exercised. Set `POD_NAMESPACE` to `installationNamespace` after `initEnvVars()`.

jeffspahr (commented):
Reviewed at `0b90acda32c9d7cbc2d63818c50dd6e1cf125fbf`. No production-code defects found in this focused Kubernetes-backed namespace guard. The explicit single-user regression and API authorization tests address the earlier coverage requests, with one non-blocking fixture issue noted inline.

Validation: full storage and server suites passed at this exact commit with `GOMAXPROCS=3 go test -p 2 ./backend/src/apiserver/storage ./backend/src/apiserver/server -count=1`. The re-review confirmed an unchanged head and diff, so these suites were not rerun. A baseline probe confirmed the fixture currently produces `NotFound` rather than returning the seeded installation-namespace pipeline. No live-cluster validation was performed.

Current CI reports 175 successful checks, six skipped checks, and Tide pending, with no reported failures. The PR targets `master`; any release backport should be validated on its target branch.

Issue #14028, opened 2026-08-12
chore(frontend): scale and harden native runtime metadata views after MLMD removal


jeffspahr:
## Chore description

PR #13986 replaces the frontend's MLMD-backed runtime metadata reads with the native V2 task and artifact APIs. The removal PR now covers the migration's direct correctness, security, compatibility, developer-experience, and bounded-rendering gaps, including final terminal-state synchronization, provider reconstruction, strict ownership configuration, iteration-aware graph binding, standalone logs, relationship direction, server-paginated artifact lineage, restored legacy visualizations, cross-run comparisons, per-run comparison failure isolation, and representative mock endpoints.

The remaining findings are scale, resiliency, API-boundary, and maintainability investments. They should be implemented immediately after #13986, but are intentionally separated because they require broader API design, authorization-cache policy, compatibility analysis, or performance measurement rather than a localized migration fix.

Related:

- #13986 — MLMD removal and native frontend migration
- #14020 — backend/runtime API resiliency, retry, deadline, and scale guarantees
- #13718 — superseded MLMD resiliency work

Resolved in #13986 and therefore not part of this follow-up:

- Launcher ConfigMap reads and provider YAML parsing are coalesced and cached for 30 seconds per namespace. Successful/not-found results expire normally; read and parse failures are evicted immediately; tests cover concurrency, expiration, and namespace isolation.
- Poll-time runtime-flow reconciliation no longer rebuilds and lays out the complete DAG merely to compare structure. It performs a cheap ParallelFor iteration-node check and invokes dagre only when that runtime structure changes.
- V2 run details render legacy `mlpipeline-ui-metadata` table, ROC-from-CSV, static HTML, TensorBoard, and web-app viewers through the native artifact path.
- CompareV2 provides a shared cross-run ROC overlay and two independently selectable confusion-matrix, HTML, and Markdown panels.
- CompareV2 loads each selected run and its task data under an independent query key. The queries start in parallel, preserve and render successful runs when another run is deleted, permission-denied, or unavailable, report the failed run explicitly, disable automatic retries, and reuse cached run data when the comparison set changes.
- Artifact Details → Related tasks requests and renders one server-backed page at a time through the existing `CustomTable` token interaction. It preserves relationship direction and task links, guards repeated tokens, ignores stale overlapping results, and no longer drains every page or mounts an unbounded relationship table.

## Why this is deferred until after #13986

[...]

Issue #14020, opened 2026-08-11
chore(backend): carry resiliency and scale guarantees into the post-MLMD architecture


jeffspahr:
## Chore description

PR #13986 removes MLMD from the KFP runtime metadata path and replaces it with run-scoped task and artifact APIs backed by the KFP API server and primary database. PR #13718 was opened to harden the MLMD client and driver behavior against MLMD restarts, hangs, and transient failures, but much of its implementation becomes obsolete once #13986 lands.

We need to determine which resiliency and scale guarantees should be included in #13986 and which can be delivered as explicit fast follows. Closing #13718 as merely “no longer needed” would lose track of several concerns that move to the replacement architecture rather than disappear.

Related:

- #13986 — MLMD removal
- #13718 — MLMD restart resilience and optional driver retries
- MLMD removal performance/test plan: `proposals/12147-mlmd-removal/test_plan.md`

### Executive summary

The earlier hardening work has three outcomes:

1. MLMD deployment-specific changes become obsolete because the corresponding pods, services, images, and routing are removed.
2. Several application-level retry and idempotency improvements are preserved or replaced by stronger native task/artifact invariants.
3. The method-aware transient RPC retry budget from #13179, bounded-call behavior discussed in #13718, and realistic scale/disruption validation are not currently carried into the new runtime API path.

The replacement architecture is structurally better in several respects: it removes an extra service/database hop, batches task and artifact mutations, creates artifacts and links transactionally, gives tasks deterministic logical identities, uses unique relationship constraints, and row-locks task merges. However, runtime pods now depend on the KFP API server and primary database for the same critical-path operations. API restarts, hung RPCs, database contention, ambiguous write responses, and saturation therefore still need an explicit policy and validation story.

### Disposition of earlier MLMD resiliency, scale, and retry work

| Earlier PR | Disposition after #13986 | Notes |
|---|---|---|
| #13179 — increase MLMD retry budget for MySQL deadlocks and `Unavailable` | **Not carried forward** | The replacement client configures connection backoff, but connection backoff does not replay failed RPCs. There is no per-method retry policy, retryable-code policy, bounded attempt budget, or per-attempt timeout for task/artifact RPCs. |
| #13276 — recover duplicate MLMD executions during workflow retries | **Captured and strengthened for tasks** | Tasks receive a deterministic logical key backed by a unique database index. Repeated or racing creates return the existing task. This is stronger than recognizing a duplicate MLMD error after insertion. |
[...]

PR #14061, opened 2026-08-17
fix(ci): retry issue analyzer on title edits

### LINKED ISSUE #14048: [aw] AI issue quality analyzer failed
### Workflow Failure

**Workflow:** [AI issue quality analyzer](https://github.com/kubeflow/pipelines/blob/master/.github/workflows/ai-analyzer.md)  
**Branch:** master  
**Run:** https://github.com/kubeflow/pipelines/actions/runs/31884537129

> [!WARNING]
> **Engine Failure**: The `copilot` engine terminated before producing output.

The engine exited immediately without producing any output. This often indicates a transient infrastructure issue (e.g., service unavailable, API rate limiting). If this failure recurs, check the GitHub Copilot status page and review the firewall audit logs.

### Action Required

**Assign this issue to an agent** to debug and fix the issue.

<details>
<summary>Debug with any coding agent</summary>

Use this prompt with any coding agent (GitHub Copilot, Claude, Gemini, etc.):

````
Debug the agentic workflow failure using https://raw.githubusercontent.com/github/gh-aw/main/debug.md

The failed workflow run is at https://github.com/kubeflow/pipelines/actions/runs/31884537129
````

</details>

<details>
<summary>Manually invoke the agent</summary>

Debug this workflow failure using your favorite Agent CLI and the `agentic-workflows` prompt.

- Start your agent
- Load the `agentic-workflows` skill from `.github/skills/agentic-workflows/SKILL.md` or <https://github.com/github/gh-aw/blob/main/.github/skills/agentic-workflows/SKILL.md>
- Type `debug the agentic workflow ai-analyzer failure in https://github.com/kubeflow/pipelines/actions/runs/31884537129`

</details>

> [!TIP]
> <details>
> <summary>Stop reporting this workflow as a failure</summary>
>
> To stop a workflow from creating failure issues, set `report-failure-as-issue: false` in its frontmatter:
> ```yaml
> safe-outputs:
>   report-failure-as-issue: false
> ```
>
> </details>

> Generated from [AI issue quality analyzer](https://github.com/kubeflow/pipelines/actions/runs/31884537129) · [◷](https://github.com/search?q=repo%3Akubeflow%2Fpipelines+is%3Aissue+%22gh-aw-workflow-id%3A+ai-analyzer%22&type=issues)
> - [x] expires  on Aug 22, 2026, 12:27 PM UTC

### PR DESCRIPTION
**Pull request by zazulam**
Fixes #14048

## Description

Recent issues exposed two independent failures in the AI issue quality analyzer.

### Reliable title correction retries

The regex evaluated the original `opened` event title correctly, but the workflow did not run again after contributors fixed the title:

- #14043 was opened as `[backend] UploadPipeline ...` and renamed to `bug(backend): UploadPipeline ...`.
- #14047 was opened as `[SDK] Add comprehensive ...` and renamed to `feat(sdk): add comprehensive ...`.

This change adds failure-aware event classification and retry behavior so:

- every newly opened issue receives deterministic title validation;
- invalid-title runs post guidance and attempt to self-cancel so corrections remain quota-free;
- a title edit retries analysis only when its current title is valid and no trusted prior comment contains the completed verdict section;
- skipped and validation notices never masquerade as completed reviews;
- comment lookup failures explicitly fail open to analysis rather than dropping retries;
- non-qualifying title edits attempt self-cancellation before inference, with visible warnings if cancellation loses a race;
- the original external-user limit remains three qualifying runs per hour;
- analyzed per-issue runs use `queue: max`, while body-only edits use unique concurrency groups and skip without provisioning the classifier runner;
- the classifier and validator share one required workflow-level title pattern.

### Fixed Copilot runtime path

#14046 passed title validation but its agent job failed because the v0.85.4-generated workflow installed Copilot under the runner tool cache and invoked nonexistent `/usr/local/bin/copilot` inside AWF.

The lock workflow is generated with gh-aw v0.86.2, the latest stable release. It contains the Copilot sandbox path fix and stages the resolved executable at `${RUNNER_TEMP}/gh-aw/bin/copilot`. Strict mode requires an explicit Bash policy with `min-integrity: none`; `bash: []` intentionally keeps shell access disabled.

## Verification

- `gh aw compile .github/workflows/ai-analyzer.md`
- `gh aw compile .github/workflows/ai-analyzer.md --no-emit --actionlint --shellcheck --validate`
- Tested opened, body-only, invalid-title, skipped-notice, trusted-verdict, missing-review, and corrected-title decisions.
- Enumerated recent analyzer comments: no historical completed verdict comments exist, only skipped notices and validation failures.
- Confirmed generated permissions, concurrency groups, queue behavior, three-run limit, and cancellation warnings.
- Confirmed canceled runs are excluded by gh-aw v0.86.2's rate limiter.
- Confirmed the generated workflow stages Copilot rather than using `/usr/local/bin/copilot`.
- `git diff --check`

### CODE REVIEW

File: .github/workflows/ai-analyzer.md [hunk 1]
```diff
@@ -3,22 +3,23 @@ description: Review the quality of new Kubeflow Pipelines issues
...
-    types: [opened]
+    types: [opened, edited]
   roles: all
   status-comment: false
   permissions:
     issues: write
   steps:
     - name: Validate and classify issue title
       id: validate_title
+      if: github.event.action == 'opened' || github.event.changes.title.from != ''
```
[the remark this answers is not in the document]
zazulam: Addressed. gh-aw v0.86.2 does not support `jobs.pre-activation.if` (the compiler only accepts `steps`, `pre-steps`, and `outputs` there), so I implemented the supported equivalent: a conditional `title_change` job listed in `on.needs`. Both `pre_activation` and `activation` now depend on that gate. Body-only edits skip `title_change`, so the rate-limit step never runs; opened issues and title edits proceed normally. The regenerated lock file shows `pre_activation: needs: title_change`.

jeffspahr:
Reviewed at `b8142a0`. The core gate is wired correctly — `title_change` is in `needs` for `pre_activation`, `activation`, `agent`, and `conclusion`, skip-propagation suppresses everything on body-only edits, and `conclusion`'s `always()` guard still evaluates false there. Comments below are about the surrounding behavior the new trigger exposes.

### Findings 1–3 compound

A contributor who opens an issue with a non-conforming title gets a stack of duplicate failure comments, and may hit the rate limit before the retry this PR exists to enable can fire.

**1. Duplicate "Validation Failed" comments stack on every invalid title edit** — `.github/workflows/ai-analyzer.md:22`

`validate_title` posts via raw `gh issue comment` rather than a safe output, so `safe-outputs.add-comment.hide-older-comments` does not apply to it. With `edited` in `types`, opening `Fix the thing` posts comment #1; renaming to `Fix the thing again` (still invalid) fires `title_change` → `pre_activation` → identical comment #2, and so on per rename. The `<!-- gh-aw-workflow-id: ai-analyzer -->` marker added in the same hunk only ever appears in the comment body — nothing in the lock file reads it, and `hide_older_comments` runs in the `safe_outputs` job, which is skipped whenever the title is invalid. So the pile only collapses once the title finally validates.

Suggest checking for an existing marker comment before posting, or posting only when `github.event.action == 'opened'`.

**2. The failure comment is posted even when the run is rate-limited** — `.github/workflows/ai-analyzer.md:22`

In the compiled job (`ai-analyzer.lock.yml:1541`) `validate_title` has no `if:` and runs unconditionally after `check_rate_limit`. When `rate_limit_ok` is false the analyzer is suppressed but the comment still goes out, telling a rate-limited user to rename the issue "to retry the quality review automatically" — and that retry is suppressed too. Gating on `steps.check_rate_limit.outputs.rate_limit_ok == 'true'` would fix it.

**3. Title edits now consume the per-user rate-limit budget** — `.github/workflows/ai-analyzer.md:74`

`user-rate-limit: max-runs-per-window: 3 / window: 60` with `GH_AW_RATE_LIMIT_EVENTS: "issues"` counts every triggering run; before this change only `opened` counted. Open the issue (run 1), correct the title (run 2), correct again (run 3), get it right on the third correction (run 4) → rate-limited → `pre_activation.outputs.activated == 'false'` → `activation`/`agent` skipped → no review is ever posted and nothing explains why. Worth raising the window budget or excluding `edited` from the counted events.

### Other findings

**4. Every title edit re-runs the full agent, not just edits that follow a failed validation** — `.github/workflows/ai-analyzer.md:6`

`GH_AW_RATE_LIMIT_IGNORED_ROLES: "admin,maintain,write"` exempts maintainers from rate limiting, so a triage session renaming existing issues to conform to the `<type>(<area>):` convention launches a full agent run (up to `max-ai-credits: 250`) plus threat detection (`max-ai-credits: 100`) per rename. Roughly 15 renames exhausts `max-daily-ai-credits: 5000`, after which genuinely new issues hit `daily_ai_credits_exceeded` and get no review at all.

The gate could additionally require that the *previous* title failed the pattern — `github.event.changes.title.from` is available and holds exactly the old title. That also matches the intent as described in `docs/agents/ci.md`.

**5. `github.event.changes.title.from != ''` relies on implicit coercion** — `.github/workflows/ai-analyzer.md:74` (nit)

On a body-only edit `github.event.changes.title` is null, so `.from` is null, and the comparison is true-negative only because Actions coerces both `null` and `''` to `0`. Correct today, but `github.event.changes.title != null` states the actual intent ("was the title among the changed fields?") and is less brittle.

zazulam:
Thanks @jeffspahr — addressed all five findings in `e32717f41`.

- Added an exact-regex event classifier and dependency gate. Opened issues still run, but edits now proceed only for invalid-to-valid title transitions; body-only, invalid-to-invalid, valid-to-valid, and valid-to-invalid edits skip before `pre_activation`.
- Made the deterministic validation comment idempotent with a dedicated marker and PATCH-on-repeat behavior.
- Gated validation on `steps.check_rate_limit.outputs.rate_limit_ok == 'true'`.
- Raised the external-user hourly limit from 3 to 5 to leave retry headroom under gh-aw's run-based accounting.
- Replaced the implicit `.from != ''` coercion with the explicit `github.event.changes.title != null` check.
- Updated `docs/agents/ci.md` to document the invalid-to-valid retry behavior precisely.

The source and generated lock workflow pass the v0.86.2 compile, actionlint, shellcheck, and validation checks.

jeffspahr:
Re-reviewed at `a1591c4b`. All five findings are resolved, and I verified the wiring at head rather than reading the summary:

- **1 (duplicate comments)** — resolved structurally by the new classifier; see note 1 below.
- **2 (comment while rate-limited)** — `if: steps.check_rate_limit.outputs.rate_limit_ok == 'true'` on `validate_title`.
- **3 (budget starvation)** — the classifier caps a user at roughly two runs per issue, and `GH_AW_RATE_LIMIT_MAX` moves 3 → 5.
- **4 (every edit re-runs the agent)** — `classify_title_event` now requires the current title to match and the previous title not to.
- **5 (implicit coercion)** — now `github.event.changes.title != null`.

Skip propagation checks out: `pre_activation`, `activation`, `agent`, and `conclusion` all carry both `classify_title_event` and `title_analysis_gate` in `needs`. `detection` (`always() && agent.result != 'skipped'`), `safe_outputs` (`detection.result == 'success'`), and `conclusion` (`always() && (agent.result != 'skipped' || <activation outputs>)`) all evaluate false when the gate skips, since a skipped `activation` yields empty outputs. Body-only, invalid→invalid, valid→valid, and valid→invalid edits all terminate before `pre_activation`.

Three non-blocking observations, all "consider" rather than "must fix":

**1. The idempotency fix is now unreachable.** Since the classifier only admits `opened` or invalid→valid, `validate_title` reaches the comment branch in exactly one situation: `opened` with an invalid title, where no prior marker comment exists. The invalid→valid path has a valid title and returns before commenting. So the `existing_comment_id` lookup and PATCH branch never fire, while adding a `gh api .../comments?per_page=100` request to every opened-with-invalid-title event. The fix for finding 4 subsumed the fix for finding 1 — reverting this back to the plain `gh issue comment` would lose nothing. If you'd rather keep it as defensive coding that's fine too; just note `per_page=100` would miss the marker on an issue with more than 100 comments.

**2. The title regex is now duplicated.** `title_pattern='^(bug|chore|feat)\(([a-z]+)\):[[:space:]]*([^[:space:]].*)$'` appears identically in both the `classify` step and the `validate_title` step. They agree today, but drift would fail in a confusing way: a title the classifier accepts and `validate_title` rejects spins up the agent job only to post a validation-failure comment. Worth a comment tying the two together, or hoisting to a single definition if gh-aw allows it.

**3. valid→invalid renames are now silent.** Renaming a conforming title to a non-conforming one produces no warning and no analysis. That's almost certainly intended, since the issue was already reviewed — but given `docs/agents/ci.md` now documents the retry path specifically, it may be worth a sentence there so the omission reads as deliberate.

Unrelated, for sequencing: this lock file is compiled with gh-aw v0.86.2, and Dependabot #14064 wants to bump the same generated file to v0.86.3 by editing the output directly. This PR should land first.

/approve

zazulam:
Thanks for the re-review, @jeffspahr. I addressed the three follow-up observations in `5da3cdb29`:

- Removed the unreachable marker lookup/PATCH branch; invalid-to-invalid edits are structurally gated out, so the plain validation comment cannot duplicate.
- Hoisted the title regex into one workflow-level `ISSUE_TITLE_PATTERN` used by both the event classifier and deterministic validator.
- Updated `docs/agents/ci.md` to state explicitly that valid-to-invalid title changes skip analysis.

I regenerated the lock workflow with gh-aw v0.86.2, reran compile/actionlint/shellcheck/schema validation, and synchronized with current `master`.

zazulam:
@jeffspahr One sequencing clarification: we are intentionally keeping this workflow on gh-aw **v0.86.2**, which is the latest stable release and contains the Copilot sandbox path fix required here. gh-aw v0.86.3 is currently marked as a prerelease, so we should not fold that compiler/action bump into this production workflow fix.

Dependabot PR #14064 should not merge while v0.86.3 remains prerelease. I have placed it on hold with the same rationale.

.github/workflows/ai-analyzer.md
```diff
@@ -1,24 +1,28 @@
 ---
 description: Review the quality of new Kubeflow Pipelines issues
 
+env:
+  ISSUE_TITLE_PATTERN: '^(bug|chore|feat)\(([a-z]+)\):[[:space:]]*([^[:space:]].*)$'
+
 on:
   issues:
-    types: [opened]
+    types: [opened, edited]
```
jeffspahr, line 17: **Gating before the rate limiter does not actually protect the budget - `edited` runs still count.**

The reordering here (title gate ahead of `pre_activation`) only helps if the rate limiter measures *analyses*. It measures **workflow runs**. `check_rate_limit.cjs` in `github/gh-aw` (`actions/setup/js/check_rate_limit.cjs`) calls `listWorkflowRuns` filtered by actor, time window, and `run.event`, and has no visibility into whether the run's jobs were gated off. Its only exclusions are:

- `conclusion === 'cancelled'`
- runs where `updated_at - created_at < 15s`

Any run in which `classify_title_event` actually executes provisions an `ubuntu-slim` runner and comfortably exceeds 15s, so it is counted even when `should_analyze=false`.

Concrete scenario, with `roles: all` and only `admin,maintain,write` exempt from rate limiting:

1. A non-collaborator opens `bug(sdk): foo` - counted run.
2. They tweak the already-valid wording four more times (typo, clarity, scope) - four more counted runs.
3. Their next genuine issue open hits `rate_limit_ok=false`.

Because of the `if:` added to `validate_title` (see the comment there), that final issue then gets no analysis *and* no comment at all. The `max-runs-per-window` 3 -> 5 bump on line 72 does not close this, since title edits now compete with issue creation for the same budget - it just moves where the cliff is.

If the goal is that gated-off runs stay free, the gate has to keep the run under the 15s exclusion (hard, given runner provisioning) or the limit needs to be applied somewhere that can see `should_analyze`. Worth confirming the intended semantics before merging, because the PR description reads as though gating early is what saves budget.

.github/workflows/ai-analyzer.md
```diff
@@ -1,24 +1,28 @@
...
+  needs: [classify_title_event, title_analysis_gate]
   roles: all
   status-comment: false
   permissions:
     issues: write
   steps:
     - name: Validate and classify issue title
       id: validate_title
+      if: steps.check_rate_limit.outputs.rate_limit_ok == 'true'
```
jeffspahr: **This turns rate-limited runs into complete silence, removing feedback master always gave.**

On master `validate_title` ran unconditionally, so a malformed title always got the "Validation Failed" comment even when the analyzer itself was rate-limited. With this `if:`, when `rate_limit_ok != 'true'`:

- `validate_title` is skipped, so `steps.validate_title.outputs.valid` is empty;
- `valid_title` is therefore empty, `activated` resolves to `'false'`, and `activation` is skipped;
- the author gets **nothing**: no quality review, no title-format comment, and no indication they were throttled.

That interacts badly with the new comment copy on line 25, which now promises "Rename the issue with a valid title to retry the quality review automatically". A user who follows that instruction triggers a run that (per the rate-limit comment above) is itself counted, and if they are over the limit it is throttled into total silence. The workflow makes a promise it can silently fail to keep, with no way for the author to tell the difference between "still thinking" and "dropped".

Simplest fix is to let title validation stay unconditional - it is a cheap `gh issue comment` and it is the only feedback channel the author has:

```suggestion
```

If the intent was specifically to stop rate-limited runs from re-posting the validation comment on every edit, the targeted fix is to make the comment idempotent instead - the `<!-- gh-aw-workflow-id: ai-analyzer -->` marker you added on line 25 is already there for exactly that kind of check.

.github/workflows/ai-analyzer.md
```diff
@@ -65,10 +69,42 @@ permissions:
...
+          CURRENT_TITLE: ${{ github.event.issue.title }}
+          ISSUE_ACTION: ${{ github.event.action }}
+          PREVIOUS_TITLE: ${{ github.event.changes.title.from || '' }}
+          TITLE_CHANGED: ${{ github.event.changes.title != null }}
+        run: |
+          should_analyze=false
+
+          if [[ "$ISSUE_ACTION" == "opened" ]]; then
+            should_analyze=true
+          elif [[ "$ISSUE_ACTION" == "edited" && "$TITLE_CHANGED" == "true" ]] &&
```
jeffspahr: **Invalid -> valid is the only retry path, so a run lost for any other reason is unrecoverable.**

`should_analyze` is true only for `opened`, or for an edit where `CURRENT_TITLE` matches and `PREVIOUS_TITLE` does not. That covers the case in the PR title, but it means an issue opened with an **already valid** title has exactly one chance at analysis. If that run produces no review because of:

- rate limiting (see the `validate_title` comment),
- an `agent` job failure,
- `daily_ai_credits_exceeded`,
- a threat-detection abort,

there is no way to ask for it again. A valid -> valid rename is explicitly excluded by the `PREVIOUS_TITLE` check on line 96, and a body edit does not even pass `classify_title_event`'s job `if` on line 77. The only workaround is renaming to a deliberately invalid title and back - unobvious, and it burns two more rate-limit slots.

Since the marker comment `<!-- gh-aw-workflow-id: ai-analyzer -->` is now written into the validation comment, "has this issue already been reviewed?" is answerable. Allowing a valid -> valid edit through when no prior review comment exists would make the retry path match what the new comment copy promises.

.github/workflows/ai-analyzer.md
```diff
@@ -65,10 +69,42 @@ permissions:
   copilot-requests: write
 
 user-rate-limit:
-  max-runs-per-window: 3
+  max-runs-per-window: 5
   window: 60
 
 jobs:
+  classify_title_event:
```
jeffspahr, line 90: **Per-issue concurrency has no `cancel-in-progress`, so a later non-qualifying run can drop the one run that mattered.**

The generated lock keeps `concurrency.group: gh-aw-${{ github.workflow }}-${{ github.event.issue.number || github.run_id }}` with `cancel-in-progress` unset (defaults to `false`). GitHub queues at most one *pending* run per group and cancels the previously pending one when a newer run arrives.

That group predates this PR, but `edited` is what makes the collision reachable. Sequence:

1. Issue opened with an invalid title - run A starts, still in progress.
2. Author renames invalid -> valid - run B queues as pending, `should_analyze=true`. This is the run that should produce the review.
3. Author edits the issue body a moment later - run C queues, cancelling pending run B. C's `classify_title_event` is skipped because `changes.title` is null.

Net result: the correction is never analyzed, and nothing reports why. Editing the body right after fixing the title is a very natural thing for someone cleaning up an issue to do.

Worth deciding explicitly whether the newest run should win here. If it should, the `classify_title_event` skip on body-only edits is the thing that makes the cancel lossy rather than harmless.

.github/workflows/ai-analyzer.md
```diff
@@ -85,6 +121,7 @@ engine:
 checkout: false
 
 tools:
+  bash: []
```
jeffspahr, line 167: **`bash: []` does not grant bash, and the tool-surface change it causes is undocumented.**

Together with the v0.85.4 -> v0.86.2 compiler bump, this line changed the compiled Copilot invocation from `--allow-all-tools` to `--allow-tool github --allow-tool safeoutputs --allow-tool write` (lock line 812). No shell or bash tool ends up in that allowlist, so the tool this line names is not actually available to the agent.

Nothing breaks today - the prompt body uses no shell - and moving off `--allow-all-tools` is a genuine improvement worth keeping. The problem is discoverability: the line reads as "grant bash" while doing the opposite, the tightening is conflated with a version bump, and it is mentioned in neither the PR description nor `docs/agents/ci.md`. The next person to touch this file has no way to know that deleting a line that appears to do nothing would widen the agent's tool surface back to allow-all.

```suggestion
  # Declaring any tool switches the compiler from --allow-all-tools to an
  # explicit allowlist. This grants no bash; removing it re-enables allow-all.
  bash: []
```

Please verify the "removing it re-enables allow-all" half by recompiling without the line before taking this suggestion verbatim - I inferred that from the lock diff rather than testing it, and the comment should not claim more than you have confirmed.

jeffspahr (commented):
Reviewed the source workflow, the generated lock, and `docs/agents/ci.md`, and pulled `check_rate_limit.cjs` from `github/gh-aw` to check what the rate limiter actually counts. The core mechanism works: the `on.needs:` key does wire `pre_activation`/`activation` to the two new gate jobs, and workflow-level `env.ISSUE_TITLE_PATTERN` is visible to both the classify step and the injected `validate_title` step.

Two things I would want resolved before this merges, and they compound:

1. **The gate-before-rate-limit ordering does not protect the budget.** The limiter counts workflow runs by actor and event; it cannot see `should_analyze`. Any title edit that reaches `classify_title_event` provisions a runner, exceeds the 15s exclusion, and is counted.
2. **Gating `validate_title` on `rate_limit_ok` makes throttled runs completely silent**, where master always posted the title-format comment. Together these mean a user can burn their budget on innocuous title tweaks and then get no response at all - including from the rename the new comment copy tells them to perform.

Below those, three lower-severity notes: the invalid->valid transition is the only retry path so any other kind of lost run is unrecoverable; the pre-existing per-issue concurrency group becomes lossy now that `edited` can queue competing runs; and `bash: []` tightens the agent's tool surface without granting the tool it names or documenting the change.

Things I checked that are correct, since they are the usual worries for this kind of change:

- **Injection safety.** `CURRENT_TITLE`, `PREVIOUS_TITLE`, and `ISSUE_TITLE` are all passed through step `env:` rather than interpolated into script bodies, so a hostile issue title cannot inject shell. `${{ }}` substitution happens after YAML parsing, so a title containing `:` or newlines cannot break the generated file.
- **Pattern plumbing.** `env.ISSUE_TITLE_PATTERN` resolves in both consumers (lock line 81); neither `pre_activation`'s nor `classify_title_event`'s job/step `env:` blocks shadow it. The single-quoted source form and the double-quoted `\\(` lock form yield the same ERE, and `[[:space:]]` works in bash `=~`.
- **Skip propagation.** `classify_title_event` skipped -> `title_analysis_gate` skipped -> `pre_activation`/`activation` skipped; none of those `if:` expressions use a status function, and `conclusion`'s `always()` is neutralized by `needs.agent.result != 'skipped'`.
- **Bash correctness.** The `elif` line-continuation chain parses, and an `if` with no matching branch returns 0, so the step will not trip `bash -e`.
- **Docs.** `docs/agents/ci.md:12` accurately describes the new trigger semantics, including that valid->invalid edits are intentionally skipped.

Thanks for tightening this up - the classify/gate split is the right shape, and the marker comment is a good addition.

.github/workflows/ai-analyzer.md
zazulam: Addressed in `16fd727ab`. The classifier now runs for edits and explicitly cancels non-qualifying runs with `actions: write`; gh-aw excludes runs whose conclusion is `cancelled`, so those edits do not consume the user quota. The limit is restored from 5 to the original 3 qualifying runs/hour. Title corrections are serialized with `queue: max`.

.github/workflows/ai-analyzer.md
zazulam: Addressed in `16fd727ab`: `validate_title` is unconditional again, restoring the deterministic malformed-title feedback path. A valid-title run that is throttled remains silent, but that behavior already existed on master and broad rate-limit notification redesign is outside this focused fix.

.github/workflows/ai-analyzer.md
zazulam: Addressed in `16fd727ab`. A valid title edit now retries whenever no prior successful quality review exists, covering agent failure, daily-credit rejection, and threat-detection failure. Detection uses gh-aw’s trusted post-sanitization footer plus the legacy verdict heading, rather than an agent-supplied marker that sanitization would remove.

.github/workflows/ai-analyzer.md
zazulam: Addressed in `16fd727ab` with explicit per-issue `queue: max`. GitHub now retains pending runs instead of replacing the meaningful title-correction run with a later edit. Body-only and other non-qualifying runs subsequently self-cancel before pre-activation.

.github/workflows/ai-analyzer.md
zazulam: Clarified in `16fd727ab`. With gh-aw v0.86.2 strict mode and `github.min-integrity: none`, omitting `tools.bash` is a compile error; it does not silently restore allow-all. The source now states the verified behavior: an explicit Bash policy is required, and `bash: []` intentionally disables shell access.

zazulam:
@jeffspahr Addressed the Opus review in `16fd727ab` and replied inline with the verified behavior. The material changes are: quota-exempt self-cancellation for non-qualifying edits, unconditional deterministic validation, retry on valid title edits with no prior review, `queue: max` for per-issue runs, trusted gh-aw footer detection, and explicit disabled-Bash documentation.

I pushed back only where the concern exceeded this PR's regression surface: valid-title rate-limit silence already exists on master, and removing `bash: []` under v0.86.2 strict mode fails compilation rather than restoring allow-all.

.github/workflows/ai-analyzer.md
```diff
@@ -69,6 +76,57 @@ user-rate-limit:
...
+          TITLE_CHANGED: ${{ github.event.changes.title != null }}
+        run: |
+          should_analyze=false
+
+          if [[ "$ISSUE_ACTION" == "opened" ]]; then
+            should_analyze=true
+          elif [[ "$ISSUE_ACTION" == "edited" && "$TITLE_CHANGED" == "true" ]] &&
+            [[ "$CURRENT_TITLE" =~ $ISSUE_TITLE_PATTERN ]]; then
+            quality_review_ids="$(
+              gh api --paginate "repos/$GITHUB_REPOSITORY/issues/$ISSUE_NUMBER/comments?per_page=100" \
```
jeffspahr, line 116: **This scan runs under `set -e`, so a transient API blip silently kills the retry it exists to enable.**

The command substitution sits in the body of an `if`, and GitHub's default shell is `bash -e`. If `gh api --paginate` exits non-zero - a 502, a secondary rate limit, a `GH_HOST` misresolution on the new GHES step - the step aborts *before* line 120 writes `should_analyze`. The job fails, `title_analysis_gate` is skipped, and the corrected title is never analyzed.

That is the worst possible failure mode for this particular code: the user did the one thing that triggers a retry, and the retry is dropped with no comment and no second chance. They would have to edit the title again to get another attempt.

Please make the failure explicit rather than letting `set -e` decide:

```bash
            quality_review_ids="$(
              gh api --paginate "repos/$GITHUB_REPOSITORY/issues/$ISSUE_NUMBER/comments?per_page=100" \
                --jq '...' 2>/dev/null
            )" || {
              echo "::warning::Could not list issue comments; assuming no prior review and analyzing."
              quality_review_ids=""
            }
```

Fail-open (analyze, risk a duplicate) is probably the right default here, since a duplicate review is recoverable and a dropped review is not - but either way it should be a decision in the file rather than an accident of `set -e`.

.github/workflows/ai-analyzer.md
```diff
@@ -69,6 +76,57 @@ user-rate-limit:
...
+                --jq '.[] | select(
+                  .user.type == "Bot" and (
+                    (
+                      ((.body // "") | contains("<!-- gh-aw-agentic-workflow: AI issue quality analyzer")) and
```
jeffspahr: **The "Automated Triage Skipped" notice matches this check, so issues that were never analyzed are permanently locked out of retrying.**

This clause matches *any* analyzer comment carrying the footer, and under v0.86.2 that includes the non-review fallback. Live examples in this repo: [#14020](https://github.com/kubeflow/pipelines/issues/14020) and [#14028](https://github.com/kubeflow/pipelines/issues/14028) both received

```
## 🤖 AI Issue Quality Review

### ⚠️ Automated Triage Skipped
```

That comment exists precisely to say *no analysis happened*. Treating it as a completed review means those issues can never be retried: a title correction finds a non-empty `quality_review_ids`, sets `should_analyze=false`, and self-cancels. The check inverts its own purpose on exactly the issues that most need a retry.

The `Overall Issue Quality Verdict` clause on line 110 is the one that actually identifies a real review. Requiring the verdict rather than accepting either signal would separate reviews from skip notices:

```jq
.[] | select(
  .user.type == "Bot"
  and ((.body // "") | contains("workflow_id: ai-analyzer"))
  and ((.body // "") | contains("Overall Issue Quality Verdict"))
) | .id
```

.github/workflows/ai-analyzer.md
```diff
@@ -69,6 +76,57 @@ user-rate-limit:
...
+                      ((.body // "") | contains("workflow_id: ai-analyzer"))
+                    ) or
+                    ((.body // "") | contains("Overall Issue Quality Verdict"))
```
jeffspahr: **No analyzer comment currently in the repo matches either marker, so every already-reviewed issue will get a duplicate review on its next title edit.**

Both signals are new. I dumped the raw body of the bot comment on [#14028](https://github.com/kubeflow/pipelines/issues/14028) with `od -c`: it ends immediately after the visible text, with no `<!-- gh-aw-agentic-workflow: ... -->` footer, because v0.85.4 did not emit one. And a repo-wide comment search for `Overall Issue Quality Verdict` returns **0 hits**.

So for the entire back catalogue of reviewed issues, `quality_review_ids` comes back empty, `should_analyze` flips to true, and a title edit re-runs the full analysis and posts a second review next to the first.

This is bounded - it self-heals as new reviews land with footers - so it may well be acceptable. But it is worth a line in the PR description either way, and if not, the cheap mitigation is to also treat the older comment shape as a prior review (the `## 🤖 AI Issue Quality Review` heading is present on all of them).

.github/workflows/ai-analyzer.md
```diff
@@ -69,6 +76,57 @@ user-rate-limit:
...
+                ) | .id'
+            )"
+            if [[ -z "$quality_review_ids" ]]; then
+              should_analyze=true
+            fi
+          fi
+
+          echo "should_analyze=$should_analyze" >> "$GITHUB_OUTPUT"
+          if [[ "$should_analyze" != "true" ]]; then
+            gh api --method POST "repos/$GITHUB_REPOSITORY/actions/runs/$GITHUB_RUN_ID/cancel" >/dev/null
```
jeffspahr: **Unguarded cancel: fails the step on 409, and silently bills the contributor when it loses the race.**

Two problems with the bare call:

1. The cancel endpoint returns **409** when the run is already completing. Under `set -e` that fails the step, leaving a red run on an issue whose only sin was needing no analysis.
2. If the cancel does not take effect, the run concludes `success`, not `cancelled`. `check_rate_limit.cjs` excludes only `conclusion === "cancelled"` runs and runs under 15s - so the run **does** consume one of the contributor's 3 runs per hour. The PR description's claim that gated-off runs are excluded holds only when the cancel wins the race.

Making it non-fatal but visible costs one line:

```suggestion
            gh api --method POST "repos/$GITHUB_REPOSITORY/actions/runs/$GITHUB_RUN_ID/cancel" >/dev/null || echo "::warning::Could not cancel run $GITHUB_RUN_ID; it will count against the author's rate limit."
```

That keeps the budget optimization while making the failure mode observable instead of silent.

.github/workflows/ai-analyzer.md
```diff
@@ -1,9 +1,17 @@
...
+  ISSUE_TITLE_PATTERN: '^(bug|chore|feat)\(([a-z]+)\):[[:space:]]*([^[:space:]].*)$'
+
+concurrency:
+  group: gh-aw-${{ github.workflow }}-${{ github.event.issue.number || github.run_id }}
+  queue: max
+
 on:
   issues:
-    types: [opened]
+    types: [opened, edited]
```
jeffspahr, line 17: **With the job-level `if` gone, every issue *body* edit now starts a run - and `queue: max` makes them serialize rather than collapse.**

Two changes compound here. `classify_title_event` lost its `if: github.event.action == 'opened' || github.event.changes.title != null`, so body-only edits now reach the job, and the new `queue: max` on line 9 replaces the previous collapse-the-pending-run behavior with genuine queueing.

Net effect: a contributor making N rapid body edits while an analysis is in flight produces N serialized queued runs, each waiting on the in-progress run before it can reach the classify step that would cancel it. Editing the body several times while cleaning up a freshly filed issue is normal behavior.

Restoring the job-level `if` would drop body-only events before a runner is provisioned at all, and costs nothing - the classify script already treats `TITLE_CHANGED != true` as `should_analyze=false`, so the `if` is pure short-circuit:

```yaml
  classify_title_event:
    if: github.event.action == 'opened' || github.event.changes.title != null
    runs-on: ubuntu-slim
```

Minor related note: the `conclusion` job carries `if: always()`, so it also runs on each cancelled run. Harmless - `handle_agent_failure.cjs` early-returns for non-`failure` conclusions - but it is extra runner time per edit.

.github/workflows/ai-analyzer.md
```diff
@@ -1,9 +1,17 @@
 ---
 description: Review the quality of new Kubeflow Pipelines issues
 
+env:
+  ISSUE_TITLE_PATTERN: '^(bug|chore|feat)\(([a-z]+)\):[[:space:]]*([^[:space:]].*)$'
```
jeffspahr, line 5: Nit, but it changes a failure mode: hoisting the regex to workflow-level `env` makes both consumers fail **open** rather than closed.

`[[ "$t" =~ $ISSUE_TITLE_PATTERN ]]` with an empty or unset variable matches *every* title. In `validate_title` that means a malformed title passes validation and `BASH_REMATCH[1]`/`[2]` come back empty, so the agent prompt renders with blank `issue_type`/`issue_area`. In `classify_title_event` it means every edit looks like a valid title. The old inline `title_pattern='...'` literal could not degrade this way, because it was defined in the same script that used it.

Nothing removes the variable today, so this is about keeping it that way. One line at the top of each `run:` block turns a silent-wrong into a loud failure:

```bash
: "${ISSUE_TITLE_PATTERN:?ISSUE_TITLE_PATTERN must be set}"
```

.github/workflows/ai-analyzer.md
```diff
@@ -69,6 +76,57 @@ user-rate-limit:
   window: 60
 
 jobs:
+  classify_title_event:
+    runs-on: ubuntu-slim
+    permissions:
+      actions: write
```
jeffspahr, line 95: `actions: write` is the broadest permission in this workflow, and it now sits on the least-gated job.

That scope can cancel or re-run any workflow in the repo and delete artifacts. `classify_title_event` triggers on `issues: opened|edited` with `roles: all`, ahead of any role check or rate limiting - and since the job-level `if` was dropped in this revision, it is now the first thing *every* issue event from *any* user reaches.

I do not think this is exploitable as written: the title is only ever the left operand of `[[ =~ ]]`, never evaluated as code, and it is passed via step `env:` rather than interpolated into the script. So this is a blast-radius observation, not a vulnerability report.

Still worth noting that the only thing the job needs the scope for is cancelling *its own* run. If gh-aw offers no narrower grant, a short comment recording why `actions: write` is here would help the next reader distinguish "deliberate, minimum viable" from "copied from a template".

jeffspahr (commented):
Re-reviewed at `b1465276f`. Every finding from my previous round is addressed - thanks for the quick turnaround:

- the `if: rate_limit_ok` on `validate_title` is gone, so malformed titles get feedback again even when throttled;
- the `PREVIOUS_TITLE` heuristic is replaced with a prior-review lookup, which is the right shape;
- an explicit `concurrency:` block with `queue: max` replaces the silent pending-run cancellation;
- the self-cancel exploits the `conclusion === "cancelled"` exclusion to keep gated-off runs off the budget, which is a genuinely clever fix to the accounting problem I raised;
- and `bash: []` now carries a comment - phrased more accurately than my suggestion, which I appreciate. Strict mode plus `min-integrity: none` is the real reason, not the allow-all inference I had guessed at.

The new findings below are almost all consequences of those fixes rather than fresh regressions. The pattern is that both new mechanisms - the comment scan and the self-cancel - are unguarded shell under `set -e`, and both have semantic gaps.

The one I would most want addressed is the comment scan, because it currently fails in **both** directions:

- it counts the "Automated Triage Skipped" notice as a completed review, so issues like #14020 and #14028 - which were never analyzed - are now permanently blocked from retrying, which is the exact opposite of the intent;
- and it cannot see any review posted before this PR, since v0.85.4 emitted no footer and `Overall Issue Quality Verdict` has 0 hits repo-wide, so every already-reviewed issue gets a duplicate on its next title edit.

Also worth a look: the self-cancel partly undermines its own purpose, since a cancel that loses the race leaves the run concluding `success` and still counting against the limit.

Verification notes, so you know what is checked rather than assumed: I confirmed `queue: max` is a real supported concurrency key and that `check_rate_limit.cjs` excludes `conclusion === "cancelled"` runs, both against gh-aw v0.86.2 sources. The Copilot staging fix is sound - the CLI is a self-contained release binary so `cp` of the resolved path works, and `${RUNNER_TEMP}/gh-aw` is mounted into the AWF sandbox. I also checked that the new `<!-- gh-aw-workflow-id: ai-analyzer -->` marker does **not** false-positive against the classifier's `contains("workflow_id: ai-analyzer")` test (hyphen vs underscore), and that it is load-bearing: `matchesWorkflowId` in `add_comment.cjs` uses it so `hide-older-comments` minimizes the stale failure notice once the retry posts its review. Nice detail.

.github/workflows/ai-analyzer.md
zazulam: Addressed in `4c8b2d13e`. The comment scan is now wrapped in an explicit fail-open branch under `bash -e`: API failures emit a workflow warning and allow analysis rather than dropping the retry.

.github/workflows/ai-analyzer.md
zazulam: Addressed in `4c8b2d13e`. Prior-review detection now requires the actual `Overall Issue Quality Verdict` section, so `Automated Triage Skipped` and validation notices do not block retry.

.github/workflows/ai-analyzer.md
zazulam: Verified and handled deliberately. A repository-wide comment enumeration since August 1 found zero successful verdict comments; the historical bot comments are only `Automated Triage Skipped` notices or validation failures. Letting those issues retry is correct and will not duplicate a completed review. Future/legacy verdicts are detected by the verdict heading.

.github/workflows/ai-analyzer.md
zazulam: Addressed in `4c8b2d13e`. Self-cancel failures are now nonfatal and produce an explicit warning that the run may count against the author’s quota.

.github/workflows/ai-analyzer.md
zazulam: Addressed in `4c8b2d13e`. The classifier job-level guard is restored. Body-only edits also receive unique run-ID concurrency groups, so they neither provision this runner nor queue behind/evict per-issue title analysis.

.github/workflows/ai-analyzer.md
zazulam: Addressed in `4c8b2d13e`. Both regex consumers now start with `${ISSUE_TITLE_PATTERN:?...}` so a missing workflow variable fails loudly instead of matching every title.

.github/workflows/ai-analyzer.md
zazulam: Documented in `4c8b2d13e`: GitHub exposes no narrower permission for canceling this workflow’s own run. The job remains injection-safe and is gated to opened/title-change events; `actions: write` is the minimum viable API scope for self-cancellation.

zazulam:
@jeffspahr Addressed the latest review in `4c8b2d13e` and replied inline. The API scan and cancel paths now have explicit failure semantics; skipped notices cannot satisfy the verdict check; body-only edits bypass the shared queue and runner; both regex consumers fail loudly if configuration disappears; and invalid-title feedback attempts to self-cancel so corrections remain quota-free.

One premise was checked rather than accepted: a repository-wide enumeration found no historical completed verdict comments—only skipped notices and validation failures—so retrying that backlog is correct and does not duplicate completed analysis.

.github/workflows/ai-analyzer.md
```diff
@@ -1,12 +1,26 @@
...
 on:
   issues:
-    types: [opened]
+    types: [opened, edited]
+  needs: [classify_title_event, title_analysis_gate]
   roles: all
   status-comment: false
   permissions:
+    # Invalid-title runs cancel after posting guidance so retries stay quota-free.
+    actions: write
```
jeffspahr, line 23: Only new thing I have not looked at before, and it is a note rather than an objection.

This makes two jobs carry `actions: write` - `pre_activation` (lock line 1577) and `classify_title_event` (lock line 990). Both are reachable by any user through `issues: opened|edited` under `roles: all`, ahead of any role gate.

To be explicit about what this is *not*, since it is the obvious worry: the `agent` job still holds only `issues: read` (lock line 398). The model does not get this scope. Both grants sit on plain shell jobs whose only use of the token is cancelling their own run, and the issue title reaches those scripts through step `env:` as a `[[ =~ ]]` operand, never as code.

So the trade looks right to me, and the comments explaining it are the correct instinct. The one thing I would add is a line in the PR description saying the workflow now holds `actions: write` in two jobs and why - a reviewer scanning permissions diffs will want that stated up front rather than inferred from two inline comments.

jeffspahr (commented):
Re-reviewed at `4c8b2d13e`. Everything from the previous round is addressed:

- the comment scan now fails open with a `::warning::` instead of dying under `set -e`;
- the footer clause is gone, so the "Automated Triage Skipped" notice no longer counts as a completed review;
- both cancel call sites are guarded and warn on failure;
- the job-level `if` is back, and the concurrency group now keys body-only edits to `run_id` - that second part goes beyond what I suggested and is the better fix, since it stops body edits from serializing on the issue group rather than just skipping them later;
- `: "${ISSUE_TITLE_PATTERN:?}"` is in both scripts;
- and `actions: write` carries an explanation.

On my "pre-existing reviews will be duplicated" point: that one is empty in practice and you can disregard it. A repo-wide comment search for `Overall Issue Quality Verdict` returns no analyzer output at all - the only comments this workflow has ever posted are the skip notices, which correctly no longer match. Since the prompt template emits that heading (line 225), matching on it is the right signal going forward. My apologies for the noise on that thread.

One new note inline on the `on.permissions` change. Nothing blocking from me.

PR #14088, opened 2026-08-18
chore(deps): bump @xyflow/react to 12.11.3

jeffspahr:
## Summary

Bump `@xyflow/react` from 12.10.1 to 12.11.3 and adapt `DagCanvas` to the updated `OnNodeDrag` signature.

Supersedes #14077.

## Why

React Flow 12.11 types the node drag event as the native `MouseEvent | TouchEvent` rather than React's synthetic `MouseEvent`. The existing annotation on `onNodeDragStop` therefore no longer matched, and #14077 failed `frontend-tests` with:

```
src/pages/v2/DagCanvas.tsx(117,13): error TS2322: Type '((_event: MouseEvent<Element, MouseEvent>, draggedNode: PipelineNode) => void) | undefined'
  is not assignable to type 'OnNodeDrag<PipelineNode> | undefined'.
```

Dependabot cannot make the accompanying source change, so its PR could not go green on its own.

Rather than widen the parameter annotation by hand, the callback is now typed as `useCallback<OnNodeDrag<PipelineNode>>`, so the parameter types come from the library. A future signature change then fails at the handler definition rather than at the `ReactFlow` call site.

The `package-lock.json` diff is larger than the single version bump because the updated `@xyflow/system` d3 ranges now overlap with `victory-vendor`'s, so npm hoists `d3-drag`, `d3-interpolate`, `d3-selection`, `d3-transition`, and `d3-zoom` to the top level.

## Verification

- `npx tsc --noEmit` — clean
- `npx prettier --check src/pages/v2/DagCanvas.tsx` — clean
- `npm test` — 149 files, 2050 tests, all passing
- Ran on node 24.14.0, matching `frontend/.nvmrc`

### DIFF
--- a/frontend/package-lock.json
+++ b/frontend/package-lock.json
@@ -17,7 +17,7 @@
         "@tanstack/react-query": "^5.91.2",
         "@types/lodash.groupby": "^4.6.9",
         "@types/pako": "^1.0.3",
-        "@xyflow/react": "^12.10.1",
+        "@xyflow/react": "^12.11.3",
         "ace-builds": "^1.44.0",
         "d3-dsv": "^3.0.1",
         "dagre": "^0.8.5",
@@ -3472,6 +3472,8 @@
     },
     "node_modules/@types/d3-drag": {
       "version": "3.0.7",
+      "resolved": "https://registry.npmjs.org/@types/d3-drag/-/d3-drag-3.0.7.tgz",
+      "integrity": "sha512-HE3jVKlzU9AaMazNufooRJ5ZpWmLIoc90A37WU2JMmeq28w1FQqCZswHZ3xR+SuxYftzHq6WU6KJHvqxKzTxxQ==",
       "license": "MIT",
       "dependencies": {
         "@types/d3-selection": "*"
@@ -3508,6 +3510,8 @@
     },
     "node_modules/@types/d3-selection": {
       "version": "3.0.11",
+      "resolved": "https://registry.npmjs.org/@types/d3-selection/-/d3-selection-3.0.11.tgz",
+      "integrity": "sha512-bhAXu23DJWsrI45xafYpkQ4NtcKMwWnAC/vKrd2l+nxMFuvOT3XMYTIj2opv8vq8AO5Yh7Qac/nSeP/3zjTK0w==",
       "license": "MIT"
     },
     "node_modules/@types/d3-shape": {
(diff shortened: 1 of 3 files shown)


frontend/package-lock.json
```diff
@@ -5354,8 +5317,26 @@
       }
     },
     "node_modules/d3-dispatch": {
-      "version": "1.0.5",
-      "license": "BSD-3-Clause"
+      "version": "3.0.1",
```
manaswinidas, line 5327: This hoist replaced top-level `d3-dispatch` / `d3-ease` / `d3-timer` 1.x with 3.x. `d3@5` is still a direct dependency and wants those modules at 1.x; app source never imports `d3` (only `d3-dsv`), so this is latent rather than a current break.

Please confirm nested 1.x copies remain under `node_modules/d3`, or drop unused `d3@5` so mixed majors cannot resolve later. Not blocking.

manaswinidas (commented):
Non-blocking: lockfile hoist replaced top-level d3 v1 modules with v3 while unused `d3@5` is still a direct dep.

frontend/package-lock.json
jeffspahr: Confirmed no mixed-major hazard here — though not for the reason assumed: there is no `d3` package in the tree at all.

`frontend/package.json` has exactly one d3 entry, `d3-dsv`. Neither this branch's lockfile nor master's contains a `d3` package entry or any `"d3"` version range, so there is no `node_modules/d3` for nested 1.x copies to live under, and no unused `d3@5` to drop.

On the hoist itself: nothing in the tree requires `d3-dispatch` / `d3-ease` / `d3-timer` at 1.x. The only ranges present are `1 - 3` (from `d3-drag`, `d3-transition`, `d3-zoom`) and `^3.0.1` (from `victory-vendor`). The 3.x hoist satisfies both, and dedupes the nested copies `victory-vendor` previously carried.

The top-level 1.x entries it replaces were stale orphans — `{"version": "1.0.5", "license": "BSD-3-Clause"}` with no `resolved` or `integrity` and no requirer anywhere in the tree. The new 3.x entries carry both.

After `npm ci`, every d3 module resolves to a single version:

```
d3-color 3.1.0   d3-dispatch 3.0.1   d3-drag 3.0.0        d3-dsv 3.0.1
d3-ease 3.0.1    d3-interpolate 3.0.1 d3-selection 3.0.0  d3-timer 3.0.1
d3-transition 3.0.1   d3-zoom 3.0.0
```

(`d3-format`, `d3-time`, `d3-time-format` are untouched by this PR.)

Verified on this branch rebased onto master, which picks up the `d3-dsv` 3.0.1 bump from #14087: `npm ci` clean, `tsc --noEmit` clean, 149 test files / 2050 tests passing.

manaswinidas (approved):
/lgtm
/approve

PR #14089, opened 2026-08-18
chore(deps): move frontend TypeScript to 6.0.3

jeffspahr:
## Summary

Move the frontend from TypeScript 4.9.5 to **6.0.3**, migrating the compiler options TypeScript 6 deprecates rather than only silencing them, and fix the type errors the upgrade exposes.

It also carries the `@tanstack/react-query` 5.101.4 bump that motivated the upgrade, superseding #14074.

## Why now

react-query 5.101.4's types use the `NoInfer<T>` utility, added in **TypeScript 5.4**. Under 4.9 it fails to resolve:

```
node_modules/@tanstack/query-core/build/legacy/_tsup-dts-rollup.d.ts(1302,203): error TS2304: Cannot find name 'NoInfer'.
```

`useQuery` then degrades to `any`, and because `skipLibCheck: true` suppresses the library errors, the only visible symptom is six unrelated-looking `TS7006 implicitly has an 'any' type` errors in application code. Annotating those parameters would have made CI green while leaving every react-query call site silently untyped.

The bump is included here as its own commit, and it needs no source changes at all: it touches only `package.json` and `package-lock.json`, and those six call sites are untouched and type correctly again.

## Why 6.0.3

6.0.3 is the newest release inside the `>=4.8.4 <6.1.0` window `typescript-eslint` declares, and that is the only upper bound on `typescript` anywhere in the dependency tree — every other constraint is an open-ended floor. There is no `typescript-eslint` release yet that supports 6.1+, so 6.0.3 is the ceiling today.

## Deprecated options

TypeScript 5.5 removed `suppressImplicitAnyIndexErrors`, which `tsconfig.json`, `mock-backend/tsconfig.json`, and `server/tsconfig.json` all set. TypeScript 6 additionally deprecates `baseUrl`, `target: es5`, `downlevelIteration`, and `moduleResolution: node10`, all of which stop working in TypeScript 7.

Migrated:

- **`target: es5` to `es2015`** — `noEmit` is set, so this only ever affected type checking, and `vite.config.mts` already builds with `build.target: 'es2015'`. The two were inconsistent.
- **`downlevelIteration` dropped** — only applies below es2015.
- **`moduleResolution: bundler`** in the app config, which is the correct mode for a Vite app.
- **Explicit `paths`** for the `src/` and `third_party/` import roots.

**All four are migrated — no deprecated compiler options remain.** Two of them initially had to stay because *runtime* tooling required them, and the fix was to replace that tooling rather than defer:

- **`baseUrl`** — `tsconfig-paths@3.x` resolves a config from the working directory when `npm run mock:api` starts and hard-fails with `Missing baseUrl in compilerOptions`. Removing `baseUrl` broke the mock backend while every typecheck still passed.
- **`moduleResolution: node10` in `mock-backend`** — `ts-node-dev` ran that config with `module=commonjs`, which is incompatible with `bundler`.

Both came from the same place: `mock:api` was started by `ts-node-dev` with a `-O '{"module":"commonjs"}'` override plus `-r tsconfig-paths/register`. `tsx` resolves tsconfig `paths` natively and selects the module format itself, so neither workaround is needed. The script now reads:

```
"mock:api": "tsx watch --tsconfig mock-backend/tsconfig.json mock-backend/mock-api-server.ts 3001"
```

`ts-node`, `ts-node-dev`, and `tsconfig-paths` are dropped entirely — `mock:api` was their only consumer, and `ts-node@10.9.2` and `ts-node-dev@2.0.0` have both been unmaintained for some time. That takes the lockfile from 878 to 842 package entries.

## Type fixes

Removing the suppression surfaced the implicit-any index accesses it was hiding. Each is now typed rather than re-suppressed:

- **Route params** — `PageProps` declares the params bag explicitly, and as `string | undefined` rather than `string`, since a param genuinely can be absent. The 18 read sites default to the empty string, matching what the surrounding code already did by hand. The catch-all 404 route narrows at the boundary, because `Route` forwards only its `Path` generic, not `Params`.
- **Injected globals** — `KFP_FLAGS` and the nonstandard Safari detection property are declared on `Window` via `declare global`, matching the existing idiom in `src/features.ts`.
- **Enum-keyed lookups** — `Trigger` and `VisualizationCreator` iterate `keyof typeof`.
- **`mlmd/Utils`** — the field name comes from the caller, so membership is now checked at runtime instead of asserted with `as keyof`.
- **Storybook** — better module resolution surfaced three stories declaring an `argType` for a `backgroundColor` prop none of those components accept, so the control was wired to nothing. Removed.

`frontend/server` moves to 6.0.3 as well and drops both `suppressImplicitAnyIndexErrors` and its unused `baseUrl`. It has no path mappings and no non-relative local imports, and compiles clean without the suppression.

## Mock backend sort keys

The v1 list handlers sorted by indexing the model with the raw sort key, so two keys the backend advertises silently did nothing:

- `ExperimentSortKeys.LAST_RUN_CREATED_AT` is a property of an experiment's runs, not of `ApiExperiment`.
- `PipelineSortKeys.DISPLAY_NAME` has no counterpart on the v1 models, which carry only `name`.

Both compared `undefined` against `undefined`, so the list came back unsorted. They now route through a projection mirroring the existing `getV2ResourceValue`/`sortV2Resources` pair, and the four duplicated comparator blocks collapse into one `sortV1Resources` helper. Sorting also copies before sorting, as the v2 path already did, so list requests no longer reorder the shared fixture data in place.

## Superseded Dependabot PR

#14074 is closed in favour of this branch. Dependabot could not make the accompanying TypeScript change, so that PR could never have gone green on its own.

## Verification

- `format:check`, `lint:ui`, `lint:server` — clean at `--max-warnings=0`
- `typecheck`, `typecheck:mock-backend` — clean
- `check:react-peers` — pass, 878 package entries
- `test:ui:coverage` — 149 files, 2050 tests, all passing
- `npm run build` — full production Vite build, exit 0
- `npm run mock:api` — starts and serves; this is what caught the `baseUrl` removal, since no CI job runs it
- `server`: `tsc --project .` builds clean
- Mock backend started and served requests; `last_run_created_at` and `display_name` verified ordering correctly in both directions, `name` and `created_at` unchanged, and repeated requests no longer perturb the default order

`test:server:coverage` was not run: `frontend/server`'s suite includes integration tests that require a live Kubernetes cluster.


frontend/mock-backend/mock-api-middleware.ts
```diff
@@ -253,6 +253,66 @@ function getComparableValue(value: unknown): number | string {
...
+    .filter((run) =>
+      RunUtils.getAllExperimentReferences(run).some((ref) => ref.key?.id === experimentId),
+    )
+    .map((run) => run.created_at)
+    .filter((createdAt): createdAt is Date => !!createdAt);
+
+  return createdAts.length
+    ? createdAts.reduce((latest, current) => (current > latest ? current : latest))
+    : undefined;
+}
```
manaswinidas: `getLastRunCreatedAt` walks every run on every comparator call, so experiment sort is O(n log n × runs). Build a `Map<experimentId, Date>` once per request and look it up from `getV1ResourceValue`. Mock fixtures make this cheap today; still the wrong complexity for a shared helper.

frontend/mock-backend/mock-api-middleware.ts
```diff
@@ -253,6 +253,66 @@ function getComparableValue(value: unknown): number | string {
...
+    let result = 0;
+    if (aValue < bValue) {
+      result = -1;
+    }
+    if (aValue > bValue) {
+      result = 1;
+    }
+    return result * (desc ? -1 : 1);
+  });
+}
```
manaswinidas: `sortV1Resources` and `sortV2Resources` (immediately below) are the same comparator besides the value getter. Collapse to one helper, e.g. `sortResources(items, defaultKey, query, getValue)`, so the copy-before-sort and comparison logic cannot drift.

manaswinidas (commented):
Non-blocking: mock-backend sort helpers are duplicated, and `getLastRunCreatedAt` rescans all runs on every comparison.

frontend/mock-backend/mock-api-middleware.ts
jeffspahr: Done in 9cf564d — agreed on the complexity, and it was worse than per-comparison rescanning alone: `getAllExperimentReferences` ran for every run on every comparison too.

Replaced with `buildLastRunCreatedAtMap()`, which makes one pass over `fixedData.runs` and keeps the max `created_at` per experiment id. `sortV1Resources` builds it lazily and holds it for the duration of the call, so it happens at most once per request, and only when the sort key actually needs it — sorting jobs, runs, or pipelines never builds it.

It stays per-call rather than module-level on purpose: `fixedData.runs` is mutated by the POST handlers, so a cached map would go stale.

Verified against the running mock backend that the resulting order matches the map computed independently from the fixtures:

```
ascending:  No Runs(-), Experiment with a very…(-), Experiment Number 2(2020-07-08),
            Kubeflow Pipelines Experiment(2020-09-08), KFP v2 Runs(2021-11-17)
descending: exact reverse of the dated three, undated last
```

frontend/mock-backend/mock-api-middleware.ts
jeffspahr: Done in 9cf564d, using the signature you suggested:

```ts
function sortResources<T>(
  resources: T[],
  defaultSortKey: string,
  queryParam: string | undefined,
  getValue: (resource: T, key: string) => unknown,
): T[]
```

`sortV1Resources` and `sortV2Resources` are now thin wrappers over it. I kept the two wrappers rather than calling `sortResources` directly from the 10 call sites so the `V1SortableResource` / `V2FilterableResource` constraints still apply at each call, and so the v1 wrapper has somewhere to own the `last_run_created_at` lookup — but there is only one comparator and one `.slice()` now, so the copy-before-sort and comparison logic can no longer drift.

manaswinidas (approved):
/lgtm
/approve

PR #14096, opened 2026-08-20
fix(backend): eliminate redundant artifactFileExists probe and fix nil error in ReadArtifact. Fixes #14094

### LINKED ISSUE #14094: bug(backend): ReadArtifact executes redundant ResolveArtifactPath and GetFileReader calls and emits nil error
### Environment

* How did you deploy Kubeflow Pipelines (KFP)? Standalone KFP deployment on Kubernetes
* KFP version: 2.4.0 / master (HEAD)
* KFP SDK version: 2.14.3

---

### Steps to reproduce

1. Start a pipeline run that produces artifacts.
2. Send a request to the API server HTTP artifact reading endpoint (`RunArtifactServer.ReadArtifact`):
   - For a non-existent artifact:
     ```bash
     curl -i "http://<kfp-apiserver>/apis/v1beta1/runs/{valid-run-id}/nodes/{node-id}/artifacts/nonexistent:read"
     ```
   - For an existing artifact:
     ```bash
     curl -i "http://<kfp-apiserver>/apis/v1beta1/runs/{valid-run-id}/nodes/{node-id}/artifacts/{artifact-name}:read"
     ```

3. Observe the response:
   - On non-existent artifact:
     ```http
     HTTP/1.1 404 Not Found
     Content-Type: application/json

     {"error_message":"artifact not found: \u003cnil\u003e","error_details":"artifact not found: \u003cnil\u003e"}
     ```
   - On existing artifact: Database logs and object storage metrics confirm `ResolveArtifactPath` (DB query + workflow manifest JSON unmarshaling) and `GetFileReader` (remote object storage connection) were executed twice in succession for a single request.

---

### Expected result

1. When an artifact is not found, the API server should return a clear, meaningful 404 error message identifying the missing resource (e.g. `artifact "nonexistent" not found for run ...`), rather than `"artifact not found: <nil>"`.
2. For every artifact read request, the API server should resolve the artifact path and obtain the remote reader in a single pass without performing redundant pre-flight probe lookups against the database and object storage.

---

#### Root Cause Analysis

In [`backend/src/apiserver/server/run_artifact_server.go`](https://github.com/kubeflow/pipelines/blob/master/backend/src/apiserver/server/run_artifact_server.go#L77-L96):

```go
func (s *RunArtifactServer) ReadArtifact(response http.ResponseWriter, r *http.Request) {
    ...
    artifactFileExists, err := s.artifactFileExists(r.Context(), runID, nodeID, artifactName)
    if err != nil {
        s.writeErrorToResponse(response, http.StatusInternalServerError, err)
        return
    } else if !artifactFileExists {
        s.writeErrorToResponse(response, http.StatusNotFound, fmt.Errorf("artifact not found: %v", err))
        return
    }

    artifactPath, err := s.resourceManager.ResolveArtifactPath(runID, nodeID, artifactName)
    if err != nil {
        s.writeErrorToResponse(response, http.StatusInternalServerError, err)
        return
    }

    reader, err := s.resourceManager.ObjectStore().GetFileReader(r.Context(), artifactPath)
    ...
}
```

And in `artifactFileExists`:
```go
func (s *RunArtifactServer) artifactFileExists(ctx context.Context, runID string, nodeID string, artifactName string) (bool, error) {
    artifactPath, err := s.resourceManager.ResolveArtifactPath(runID, nodeID, artifactName)
    if err != nil {
        if isNotFoundError(err) {
            return false, nil
        } else {
            return false, err
        }
    }
    reader, err := s.resourceManager.ObjectStore().GetFileReader(ctx, artifactPath)
    if err != nil {
        if isNotFoundError(err) {
            return false, nil
        } else {
            return false, err
        }
    }
    defer func() {
        if closeErr := reader.Close(); closeErr != nil {
            glog.Warningf("Failed to close artifact reader: %v", closeErr)
        }
    }()
    return true, nil
}
```

Two distinct defects occur here:

1. **Redundant I/O and CPU overhead on every successful read**:
   - `artifactFileExists` calls `ResolveArtifactPath`, which fetches the run from DB via `r.runStore.GetRun(runID)`, unmarshals the entire Argo workflow JSON string (`util.NewExecutionSpecJSON`), and parses node status to find the object store key.
   - `artifactFileExists` then opens a remote reader connection via `s.resourceManager.ObjectStore().GetFileReader` (S3/GCS/MinIO network request), immediately closes it, and returns `true`.
   - Immediately following that, `ReadArtifact` repeats both operations: it calls `ResolveArtifactPath` a **second time** (re-querying the DB and re-unmarshaling the workflow JSON) and calls `GetFileReader` a **second time** (initiating another remote storage connection).

2. **`<nil>` error formatting on 404 response**:
   - When an artifact does not exist, `artifactFileExists` returns `(false, nil)`.
   - Line 82 then constructs the error: `fmt.Errorf("artifact not found: %v", err)`. Since `err` is `nil`, the error message produced and returned in JSON to clients is literally:
     ```json
     {
       "error_message": "artifact not found: <nil>",
       "error_details": "artifact not found: <nil>"
     }
     ```

#### Affected Components

- `backend/src/apiserver/server/run_artifact_server.go`
- `backend/src/apiserver/server/run_artifact_server_test.go`

#### Acceptance Criteria

- [ ] `ReadArtifact` resolves the artifact path and retrieves the object store reader in a single pass without redundant pre-flight probe calls.
- [ ] 404 responses return descriptive error messages identifying the missing artifact/run instead of `<nil>`.
- [ ] Unit tests in `run_artifact_server_test.go` verify that 404 error responses contain meaningful error messages and that valid artifacts stream in a single attempt.

#### Proposed Solution Direction

Simplify `ReadArtifact` to resolve the path and obtain the reader once, checking for `isNotFoundError(err)` or `util.IsUserErrorCodeMatch(err, codes.NotFound)` directly on the initial lookup:

```go
artifactPath, err := s.resourceManager.ResolveArtifactPath(runID, nodeID, artifactName)
if err != nil {
    if isNotFoundError(err) {
        s.writeErrorToResponse(response, http.StatusNotFound, util.NewResourceNotFoundError("artifact", common.CreateArtifactPath(runID, nodeID, artifactName)))
    } else {
        s.writeErrorToResponse(response, http.StatusInternalServerError, err)
    }
    return
}

reader, err := s.resourceManager.ObjectStore().GetFileReader(r.Context(), artifactPath)
if err != nil {
    if isNotFoundError(err) {
        s.writeErrorToResponse(response, http.StatusNotFound, fmt.Errorf("artifact file not found at path %q: %w", artifactPath, err))
    } else {
        s.writeErrorToResponse(response, http.StatusInternalServerError, fmt.Errorf("failed to get file reader: %w", err))
    }
    return
}
defer reader.Close()
```

Remove the redundant `artifactFileExists` helper.

---

bhuvan-somisetty:

### Problem & Root Cause
In `RunArtifactServer.ReadArtifact` (`backend/src/apiserver/server/run_artifact_server.go`), every artifact read request called a pre-flight probe `artifactFileExists` before streaming the artifact. This probe called `ResolveArtifactPath` (performing a DB run lookup and unmarshaling the Argo workflow JSON manifest) and `GetFileReader` (initiating and immediately closing an object store connection), only for `ReadArtifact` to repeat both operations a second time.

Furthermore, when an artifact was missing, `artifactFileExists` returned `(false, nil)`, causing `ReadArtifact` to format `fmt.Errorf("artifact not found: %v", err)` with a nil error, emitting `"artifact not found: <nil>"` in 404 responses.

### Solution
- Resolve the artifact path and obtain the remote reader in a single pass in `ReadArtifact`.
- Handle not-found conditions directly on the primary lookup and reader retrieval with descriptive error messages.
- Remove the redundant `artifactFileExists` helper.
- Update and add regression tests in `run_artifact_server_test.go` verifying 404 error responses for missing runs, missing artifact names in workflow manifests, and missing files in object storage.

Fixes #14094

**Checklist:**

### DIFF
--- a/backend/src/apiserver/server/run_artifact_server.go
+++ b/backend/src/apiserver/server/run_artifact_server.go
@@ -74,24 +74,23 @@ func (s *RunArtifactServer) ReadArtifact(response http.ResponseWriter, r *http.R
 		return
 	}
 
-	artifactFileExists, err := s.artifactFileExists(r.Context(), runID, nodeID, artifactName)
-	if err != nil {
-		s.writeErrorToResponse(response, http.StatusInternalServerError, err)
-		return
-	} else if !artifactFileExists {
-		s.writeErrorToResponse(response, http.StatusNotFound, fmt.Errorf("artifact not found: %v", err))
-		return
-	}
-
 	artifactPath, err := s.resourceManager.ResolveArtifactPath(runID, nodeID, artifactName)
 	if err != nil {
-		s.writeErrorToResponse(response, http.StatusInternalServerError, err)
+		if isNotFoundError(err) {
+			s.writeErrorToResponse(response, http.StatusNotFound, util.Wrap(err, "Failed to read artifact"))
+		} else {
+			s.writeErrorToResponse(response, http.StatusInternalServerError, err)
+		}
 		return
 	}
 
 	reader, err := s.resourceManager.ObjectStore().GetFileReader(r.Context(), artifactPath)
 	if err != nil {
-		s.writeErrorToResponse(response, http.StatusInternalServerError, fmt.Errorf("failed to get file reader: %v", err))
+		if isNotFoundError(err) {
+			s.writeErrorToResponse(response, http.StatusNotFound, fmt.Errorf("artifact file not found at path %q: %w", artifactPath, err))
+		} else {
+			s.writeErrorToResponse(response, http.StatusInternalServerError, fmt.Errorf("failed to get file reader: %v", err))
+		}
 		return
 	}
 	defer func() {
@@ -129,31 +128,6 @@ func (s *RunArtifactServer) ReadArtifact(response http.ResponseWriter, r *http.R
 		glog.Errorf("Failed to write JSON closing: %v", err)
 		return
 	}
-}
-
-func (s *RunArtifactServer) artifactFileExists(ctx context.Context, runID string, nodeID string, artifactName string) (bool, error) {
-	artifactPath, err := s.resourceManager.ResolveArtifactPath(runID, nodeID, artifactName)
-	if err != nil {
-		if isNotFoundError(err) {
-			return false, nil
-		} else {
-			return false, err
-		}
-	}
-	reader, err := s.resourceManager.ObjectStore().GetFileReader(ctx, artifactPath)
-	if err != nil {
-		if isNotFoundError(err) {
-			return false, nil
-		} else {
-			return false, err
-		}
-	}
-	defer func() {
-		if closeErr := reader.Close(); closeErr != nil {
-			glog.Warningf("Failed to close artifact reader: %v", closeErr)
-		}
-	}()
-	return true, nil
 }
 
 // ReadArtifactV1 handles v1 artifact reading (delegates to v2 implementation)
--- a/backend/src/apiserver/server/run_artifact_server.go
+++ b/backend/src/apiserver/server/run_artifact_server.go
@@ -76,7 +76,7 @@ func (s *RunArtifactServer) ReadArtifact(response http.ResponseWriter, r *http.R
 	artifactPath, err := s.resourceManager.ResolveArtifactPath(runID, nodeID, artifactName)
 	if err != nil {
 		if isNotFoundError(err) {
-			s.writeErrorToResponse(response, http.StatusNotFound, util.Wrap(err, "Failed to read artifact"))
+			s.writeErrorToResponse(response, http.StatusNotFound, util.NewNotFoundError(err, "Failed to read artifact: %v", err))
 		} else {
 			s.writeErrorToResponse(response, http.StatusInternalServerError, err)
 		}
@@ -86,7 +86,7 @@ func (s *RunArtifactServer) ReadArtifact(response http.ResponseWriter, r *http.R
 	reader, err := s.resourceManager.ObjectStore().GetFileReader(r.Context(), artifactPath)
 	if err != nil {
 		if isNotFoundError(err) {
-			s.writeErrorToResponse(response, http.StatusNotFound, util.Wrap(err, "Failed to read artifact"))
+			s.writeErrorToResponse(response, http.StatusNotFound, util.NewNotFoundError(err, "Failed to read artifact: artifact file not found at path %q", artifactPath))
 		} else {
 			s.writeErrorToResponse(response, http.StatusInternalServerError, fmt.Errorf("failed to get file reader: %v", err))
 		}
--- a/backend/src/apiserver/server/run_artifact_server.go
+++ b/backend/src/apiserver/server/run_artifact_server.go
@@ -76,7 +76,7 @@ func (s *RunArtifactServer) ReadArtifact(response http.ResponseWriter, r *http.R
 	artifactPath, err := s.resourceManager.ResolveArtifactPath(runID, nodeID, artifactName)
 	if err != nil {
 		if isNotFoundError(err) {
-			s.writeErrorToResponse(response, http.StatusNotFound, util.NewNotFoundError(err, "Failed to read artifact"))
+			s.writeErrorToResponse(response, http.StatusNotFound, util.NewNotFoundError(err, "Failed to read artifact: resource not found"))
 		} else {
 			s.writeErrorToResponse(response, http.StatusInternalServerError, err)
 		}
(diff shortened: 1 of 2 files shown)


dobbydobap:
Heads up, this doesn't build as it stands. Removing `artifactFileExists` took the only use of `context` with it, so the import is now unused:

```
backend/src/apiserver/server/run_artifact_server.go:18:2: "context" imported and not used
```

Dropping that one import fixes it. Easy to miss, since the PR is still on needs-ok-to-test so CI has never actually built it.

The change itself reads right to me. Folding the probe into the real call removes a duplicate `ResolveArtifactPath` and a full open/close of the object store on every read, and handling not-found on each call directly is what gets rid of the `artifact not found: <nil>` body, since the old helper returned `(false, nil)` and left no error to format.

One small thing: the two not-found branches are worded differently, `util.Wrap` on one and `fmt.Errorf` on the other. Matching them would keep the 404 bodies consistent.

bhuvan-somisetty:
Thanks for catching that! Good catch on the unused import and the error response consistency. I've updated the PR to drop the unused `context` import and aligned both 404 branches to use `util.Wrap(err, "Failed to read artifact")`.

dobbydobap:
Build's fixed, thanks. One thing the message change broke though: `TestReadArtifactV1_ArtifactNotFound` still asserts "artifact file not found", but that branch now returns "Failed to read artifact: ...", so it fails:

```
"Failed to read artifact: InternalServerError: Failed to get file reader for test/nonexistent.txt: blob not found" does not contain "artifact file not found"
```

Loosening that assertion to "not found" like the other two would do it.

Also worth a look: `util.Wrap` keeps the inner error's classification, so the 404 body now literally reads "InternalServerError: ...". Not wrong, but a 404 that says InternalServerError is going to confuse someone.

bhuvan-somisetty:
Good points! Replaced `util.Wrap` with `util.NewNotFoundError` on both 404 paths so 404 response bodies return clean `NotFoundError` messages without `InternalServerError:` prefixed. Also updated `TestReadArtifactV1_ArtifactNotFound` to assert `"not found"` consistent with the other tests.

dobbydobap:
Tests pass at this head, nice. Two things the run turned up though.

The message comes out twice on the `ResolveArtifactPath` branch, since `err` is passed as both the cause and the `%v` arg:

```
NotFoundError: Failed to read artifact: ResourceNotFoundError: Run non-existent-run-id not found: ResourceNotFoundError: Run non-existent-run-id not found
```

`util.NewNotFoundError(err, "Failed to read artifact")` without the `: %v` would do it, since `Wrapf` already appends the cause. The `GetFileReader` branch is fine because it formats `artifactPath` rather than `err`.

Second, `InternalServerError` is still in the 404 body:

```
NotFoundError: Failed to read artifact: artifact file not found at path "test/nonexistent.txt": InternalServerError: Failed to get file reader for test/nonexistent.txt: blob not found
```

That one isn't the constructor's fault. `writeErrorToResponse` sets `ErrorMessage` to `err.Error()`, which for a `*UserError` is the internal chain rather than `ExternalMessage()`, so the inner classification gets serialised whichever constructor you pick. Reading `ExternalMessage()` there would give the clean body, though that's arguably its own change.

bhuvan-somisetty:
Good catches! Fixed both:
1. Updated the `ResolveArtifactPath` 404 branch to `util.NewNotFoundError(err, "Failed to read artifact")` so the cause isn't duplicated in the message.
2. Updated `writeErrorToResponse` to use `ExternalMessage()` when `err` is a `*util.UserError`, removing internal classification prefixes like `InternalServerError:` from external 404 response bodies.

dobbydobap:
Nice, `ExternalMessage()` is the right call there.

One knock-on though. `NewNotFoundError` sets `externalMessage` to exactly the format you pass, so with the shorter messages the 404 bodies are now "Failed to read artifact" and "Failed to read artifact file at path ...", neither of which contains "not found". Three tests still assert `Contains(ErrorMessage, "not found")`: `TestReadArtifactV1_RunNotFound` (168), `TestReadArtifactV1_ArtifactNotFound` (305) and `TestReadArtifactV1_ArtifactNameNotFoundInManifest` (343).

Either loosen those to match the new text or put "not found" back in the messages, whichever you prefer.

`TestReadArtifactV1_Unauthorized` and `TestReadArtifactV1_MissingParameters` should be fine: `NewPermissionDeniedError` already puts the user string in the external message, and the missing-parameter errors are plain `fmt.Errorf` so they never hit the new branch.

bhuvan-somisetty:
Good catch! I've included `not found` back into the `NewNotFoundError` messages (`Failed to read artifact: resource not found` and `Failed to read artifact: artifact file not found at path %q`). This keeps the 404 error bodies clean and descriptive while satisfying the test assertions.

bhuvan-somisetty:
@dobbydobap PTAL, whenever u get chance

PR #14107, opened 2026-08-21
fix: reject percent signs in pipeline root

### LINKED ISSUE #14101: Launcher builds artifact URIs without percent-escaping, making some object names unaddressable
### What happens

`Config.PrefixedBucket()` builds the persisted artifact URI with `path.Join(b.BucketName, b.Prefix)` and no URL escaping (`backend/src/v2/objectstore/config.go:113`). The resulting URI is stored on the artifact and is also the exact string used for authorization lookups.

Because the persisted key is unescaped, a key that *contains* a percent-escape sequence and a key that *is* the escaped form of another key are the same string. Readers cannot tell them apart:

| Object in the bucket | Persisted URI key | Resolves to |
|---|---|---|
| `root dir/x` | `root dir/x` | `root dir/x` — correct |
| `root%20dir/x` | `root%20dir/x` | `root dir/x` — **wrong object** |

The UI decodes the persisted key (`decodeArtifactUriKey`, `frontend/src/lib/v2/ArtifactFileUtils.ts:28`) and has no way to recover the distinction, so an object whose literal name is also exactly a canonical escaping is unaddressable — requests for it resolve to the decoded name instead.

### Scope

Narrow. It requires an object name containing a literal `%` followed by two hex digits that form a valid escape. Literal spaces, Unicode, `&`, and `?` all round-trip correctly today. Artifact key components are derived from pipeline, task, and output names that are already constrained, so the realistic source of a stray `%` is a user-supplied `pipeline_root` prefix.

### Why not fix it by escaping at the producer

Making the launcher percent-escape when building the URI is the correct long-term fix, but the URI is persisted and compared by exact match for authorization. Changing the convention orphans every existing artifact, so it needs a dual-read period or a migration. That is disproportionate to the size of the exposure.

### Proposed fix

Reject the ambiguous input instead of trying to represent it: validate that a `pipeline_root` prefix contains no `%`. This is validation-only — no schema change, no migration — and it does not break anything that works today, since such a root is already unreadable. It converts a silent wrong-object read into an explicit error.

Placement matters, because of where the error surfaces:

1. **Primary — validate early, at run submission and at `kfp-launcher` ConfigMap load.** This is the case worth optimizing for: it fails once, loudly, at configuration time, before any pod runs.
2. **Backstop — `ParseBucketPathToConfig` (`config.go:131`)**, alongside the existing encoded-query-delimiter rejection, which already has this shape. This catches every path, but on its own it surfaces poorly (see below).

### Failure behavior, if validation is added at the parse chokepoint only

`ParseBucketPathToConfig` errors propagate through `resolveArtifactBucketConfig` into `getBucket` (`component/interfaces.go:166`), the launcher's artifact I/O path:

- Input artifacts fail before the user container runs.
- Output artifacts fail *after* it has run to completion, so the task fails with the compute already spent.
- `resolveArtifactBucketConfig` calls `IsPathUnderDefaultPipelineRoot` unconditionally, and that parses `DefaultPipelineRoot()` (`config/env.go:168`) on every artifact operation. A bad prefix in a namespace's default pipeline root therefore fails **every artifact operation in that namespace**, reported once per task at runtime.

That blast radius already applies to the existing encoded-query-delimiter check, so this is not a new hazard — but it is the reason to put the primary validation at config-load and submission time rather than relying on the runtime chokepoint alone.

### Acceptance criteria

- A `pipeline_root` whose prefix contains `%` is rejected at run submission and at launcher ConfigMap load, with an error naming the offending value.
- Existing roots containing spaces, Unicode, `&`, or `?` continue to work unchanged.
- A regression test covers the `root%20dir/x` vs `root dir/x` pair.

### Context

Found while reviewing the frontend artifact-URI handling in #13986. The client-side work there resolves every other non-canonical spelling correctly; this pair is the one case that is undecidable without a producer-side escaping contract, so it belongs upstream rather than in the UI.

Sumitha007:

Reject percent signs in pipeline root paths to prevent invalid percent-encoded paths from being accepted by the object store configuration parser.

This adds validation in `ParseBucketPathToConfig` and validates the default pipeline root loaded from the Kubernetes ConfigMap.

**Tests:**

- Added a test to reject percent signs in pipeline root prefixes.
- Added tests to ensure valid pipeline roots containing spaces, Unicode, `&`, and `?` are accepted.
- `go test ./src/v2/objectstore ./src/v2/config` passes.
- `git diff --cached --check` passes.

**Checklist:**

- [x] You have signed off your commits.

Fixes #14101

### DIFF
--- a/backend/src/v2/config/env.go
+++ b/backend/src/v2/config/env.go
@@ -79,6 +79,16 @@ func FromConfigMap(ctx context.Context, clientSet kubernetes.Interface, namespac
 		}
 		return nil, err
 	}
+
+	if pipelineRoot := config.Data[configKeyDefaultPipelineRoot]; pipelineRoot != "" {
+		if _, err := objectstore.ParseBucketPathToConfig(pipelineRoot); err != nil {
+			return nil, fmt.Errorf(
+				"invalid default pipeline root %q: %w",
+				pipelineRoot,
+				err,
+			)
+		}
+	}
 	return &Config{data: config.Data}, nil
 }
 
--- a/backend/src/v2/config/env_test.go
+++ b/backend/src/v2/config/env_test.go
@@ -15,6 +15,7 @@
 package config
 
 import (
+	"context"
 	"fmt"
 	"os"
 	"testing"
@@ -22,6 +23,9 @@ import (
 	"github.com/kubeflow/pipelines/backend/src/v2/objectstore"
 	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
+	corev1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/client-go/kubernetes/fake"
 	"sigs.k8s.io/yaml"
 )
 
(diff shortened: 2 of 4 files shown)


droctothorpe:
Hi @Sumitha007, thank you for your PR! The linked issue (#14101) must be labeled `ready` before your contribution can be reviewed. Please check our [CONTRIBUTING.md](https://github.com/kubeflow/pipelines/blob/master/CONTRIBUTING.md) for details.

PR #14109, opened 2026-08-21
fix(ci): upgrade AI analyzer workflow runtime

zazulam:
## Summary

- upgrade the generated AI issue analyzer workflow from gh-aw v0.86.2 to v0.87.2
- disable the MCP CLI proxy explicitly while shell access is disabled, keeping safe outputs available as direct MCP tools
- carry GitHub event context into safeoutputs so comments target the triggering issue

This addresses the publication failure in [run 32340176973](https://github.com/kubeflow/pipelines/actions/runs/32340176973), where the review was generated but shell-based safeoutput calls were denied and the invocation budget was exhausted.

## Validation

- `gh aw compile ai-analyzer --strict --validate --actionlint`


zazulam:
gonna merge to address the broken agentic workflow to prevent wasted credits

Issue #14110, opened 2026-08-21
chore(ci): Auto-merge CI-passed Dependabot PRs without per-PR human approval


aholten:
## Chore description

Since #13676 was completed (dependabot added to trusted apps via GoogleCloudPlatform/oss-test-infra#2619), CI starts automatically on Dependabot PRs — the per-PR `/ok-to-test` tax is gone. The remaining toil is **merge approval**: every Dependabot PR still sits until a kubeflow org member applies `/lgtm` + `/approve`, even though CI has already verified it.

As weekly version updates continue (grouped `go-minor-and-patch` PRs spanning multiple modules, plus docker/npm/pip/GitHub Actions bumps), this per-PR human approval tax stays with us.

## Current state (verified 2026-08-21)

The safety machinery for unattended merges already exists in this repo:

- `.github/workflows/ci-checks.yml` ("CI Check") — polls all CI, records the outcome plus the PR head SHA, and uploads it as the `ci-check-result` artifact.
- `.github/workflows/add-ci-passed-label.yml` — a **SHA-bound classifier**: on `workflow_run` of CI Check it verifies the artifact's recorded SHA matches the workflow run head, the PR head is unchanged, and the PR is still open, then adds the `ci-passed` label. It removes `ci-passed` on `synchronize`/`reopen`, so a Dependabot rebase resets it until CI re-verifies the new head.

However, the classifier's eligibility check requires the `ok-to-test` label:

    const eligible = labels.has("ok-to-test") && !labels.has("needs-ok-to-test");

Trusted Dependabot PRs **never carry `ok-to-test`** — that is exactly what trusted-apps support removed. Live check today: 11 open Dependabot PRs (e.g. #14085, #14086), none labeled `ci-passed`, none labeled `ok-to-test`. **`ci-passed` is therefore never applied to Dependabot PRs in the current configuration**, and the existing Tide query (which requires `lgtm` + `approved` + `ci-passed`) keeps every bot PR waiting on a human.

### PR 1 (this repo): classifier eligibility for trusted Dependabot PRs

In `add-ci-passed-label.yml`, treat PRs authored by `dependabot[bot]` as eligible for `ci-passed` without the `ok-to-test` label. CI Check running at all is already the trust boundary (untrusted PRs cannot run it); the SHA-bound artifact/head checks remain the safety core and are unchanged.

- Effect: `ci-passed` is applied only after CI Check passes for the **current** head SHA, and is still stripped on `synchronize`/`reopen`.
[...]

PR #14111, opened 2026-08-21
fix(ci): make ci-passed a SHA-bound merge gate for Dependabot automation

aholten:
## Summary

Part of #14110: make exact Dependabot PRs eligible for unattended Tide merges when current-head CI passes, while preserving human review requirements.

CI Check is the sole publisher of the SHA-bound `ci-passed` commit status; the matching label is informational. Only trusted base/default-branch code executes. This PR also removes the default Dependabot creation holds introduced in #14249 across all six ecosystems, retains their dependency/ecosystem labels, and never removes existing PR-specific holds.

## Correctness and recovery

- Exact Dependabot and MEMBER/OWNER/COLLABORATOR authors are eligible without `ok-to-test`; `needs-ok-to-test` always rejects.
- Require all applicable workflows from the trusted trigger inventory to register and finish successfully. Missing runs, incomplete API evidence, unsupported filters, and stale inventory cannot produce success.
- Explicit constituent workflow selectors exclude CI Check itself. Reconcile requested/in-progress/completed runs using current attempts, including cancellation and failure.
- Read durable retarget history on every reconciliation. Only executions originally created after a base retarget can authorize success; labels, reopening, and rerunning old executions cannot erase invalidation.
- Serialize short reconciliations by head SHA with cancellation disabled. Verify head/base/open/eligibility and workflow evidence around publication; fail on drift or unverifiable evidence.
- Recover late external checks through a scheduled sweep every 15 minutes, using the same publisher and safety checks. Discovery selects eligible open PRs without successful `ci-passed`; green/ineligible PRs do not allocate recovery runners. At most four recovery jobs run in parallel. No long poll holds the writer lock. Scheduled retries preserve existing non-success and deduplicate unchanged status writes, avoiding exhaustion of GitHub’s per-SHA/context status limit. Status lookup is paginated.
- Do not rely on `check_run` delivery: GitHub documents suppression when the check suite SHA is associated with Actions. Scheduled execution may be delayed/dropped; this is a requested cadence, not a recovery SLA. Discovery explicitly fails above 256 recovery candidates.

## Latest review → change → evidence

| Finding | Change | Evidence |
| --- | --- | --- |
| External check finishes after final workflow reconciliation | Scheduled recovery through the same SHA-bound publisher | Final workflow with external check pending → failure; later sweep → success; failed checks, stale heads, retargets, missing CI, revocation and publication drift remain blocked |
| Include default creation-hold removal in this PR | Remove default holds in all six ecosystem entries; retain other labels | Dependabot configuration regressions; publisher label-write test proves existing holds are untouched |
| Coordinate hold removal with replacement enforcement | Concrete staged Tide containment and abort/rollback procedure | [Rollout contract](https://github.com/kubeflow/pipelines/blob/5588d899b/docs/agents/ci-passed.md); maintainer agreement and live deployment evidence still required |

Earlier review fixes remain covered: trigger validity/recursion, missing workflows, cancelled reruns, retarget invalidation, and publication races. Jeff's latest review acknowledged those implementation findings as addressed.

## Validation

- DCO passes on the final head. Signed follow-up head `5588d899b` builds on the contributor's master merge `0863d897b`; review history is preserved. Includes the latest master frontend fix (`03f470d9f`).
- Local full CI-script suite: 288 tests passed before adding the final status-read failure regression; the updated publisher suite (21 tests) also passes. Seven JUnit-summary tests pass.
- Actionlint 1.7.12, Python formatting checks, and `git diff --check` pass.
- Updated [hosted validation run](https://github.com/aholten/pipelines/actions/runs/34150657327) exercises the final scripts and a read-only upstream API probe. **Passed: 290 CI-script tests and 7 JUnit-summary tests.** The read-only probe rejected incomplete upstream CI as expected.
- [Hosted writer wiring validation](https://github.com/aholten/pipelines/actions/runs/34150506167) passed: skipped recovery discovery correctly allows the ordinary singleton writer to execute. Validation-only dispatch has no PR context; preparation succeeds without publishing a status. The observed matrix job name `check_ci_status (0)` is explicitly covered by the poller exclusion and a regression test.
- Adversarial review identified status-history exhaustion from repeated retries; the fix and regressions for repeated failures → late recovery, queued-success invalidation, and status-read failure passed follow-up review.
- Hosted fixtures and read-only API probes do **not** demonstrate live publisher webhook/scheduler behavior or Tide enforcement. Base-defined upstream `pull_request_target` checks also do not execute this PR's proposed publisher.

## Deployment: coordinate before merging

#14249 is already merged; #2674 supplies the final protection/Tide policy. Because this PR now removes creation holds, neither landing order alone is sufficient.

Proposed procedure for Jeff/designated infra maintainer to approve:

1. Temporarily remove review-free Dependabot matching from deployed Tide configuration while retaining human review requirements and hold exclusions. Inspect all overlapping queries; record the owner and deployed config SHA. Preserve existing PR-specific holds.
2. Record controlled live publisher success, invalidation, recovery, and stale-head rejection before deployment. If evidence fails, retain containment and fix this PR.
3. Review and land #14111 through the existing human path, before requiring its new status. Creation holds stop, but review-free matching remains disabled. Verify the deployed publisher lifecycle.
4. Apply #2674 protection/exclusions while retaining temporary removal of its Dependabot query. Inspect live `master` protection and Tide: required `ci-passed`, strict protection, pending/failing constituent contexts block (including queued reruns), and human/release behavior remains correct.
5. Restore #2674's master-only Dependabot query only after recording the evidence. Demonstrate an actual unheld Dependabot merge without reviews and negative cases for incomplete/failing CI, trust vetoes and PR-specific holds.

The temporary infra edit needs a maintainer owner and agreed deployment mechanism. If another PR is required, discuss it before creating one. Abort on failed lifecycle validation, uninspectable configuration, or bypassing queries; keep review-free matching disabled. Restore creation holds/containment before rollback, and coordinate required-status removal before removing its publisher.


jeffspahr (changes requested):
Requesting changes for two correctness blockers before this is coupled to the author-scoped Tide query.

1. This changes only the downstream classifier, but CI Check still rejects a Dependabot PR without ok-to-test. In ci-checks.yml the eligibility step leaves should_poll=false, the poll is skipped, and the artifact records should_add_label=false. add-ci-passed-label.yml returns on that value before reaching the new author-aware expression. The live Dependabot CI Check runs linked from #14110 show exactly this no-op path. Please extend the upstream CI Check eligibility as well, while continuing to reject needs-ok-to-test.

2. ci-passed is not fully SHA-bound at the point of mutation. fetch_data verifies the current head, then a later independent job adds the PR label without re-reading the head. A synchronize can occur between those jobs: its reset can run first, after which the old-head workflow_run adds ci-passed back. That is unsafe once the linked Tide change removes lgtm and approved. Prefer publishing a required commit status/check on the verified head SHA and use that as the merge gate; the label can remain informational.

Please also add a truth-table regression test covering human, exact dependabot[bot], another bot, ok-to-test, and needs-ok-to-test through both workflows, plus a synchronize-between-validation-and-publication case. add-ci-passed-label.yml should be included in the CI Scripts Tests path filter so this change actually runs that coverage.

The one-line author predicate itself is correctly narrow, but it is currently unreachable for normal trusted Dependabot runs and does not provide the freshness guarantee required for unattended merge.

aholten:
@jeffspahr thanks for the review! Tests added and changes made:

**1. Upstream eligibility reachable for Dependabot (ci-checks.yml)**
Eligibility reordered: `needs-ok-to-test` rejects unconditionally first, then `ok-to-test`, then a new `dependabot[bot]` author branch. A Dependabot PR now records `should_add_label=true` without the label, so the downstream author-aware expression is reachable. Commits: 4bd88fc0a.

**2. SHA-bound merge gate (add-ci-passed-label.yml + [oss-test-infra#2660](https://github.com/GoogleCloudPlatform/oss-test-infra/pull/2660))**
- The workflow publishes a `ci-passed` commit status on the verified head SHA after the CI poll passes (statuses permission added).
- The label job re-reads the PR head (`gh pr view --json headRefOid`) immediately before labeling and skips if it moved — the synchronize race between validation and publication is closed even without the status.
- The status is now the required merge gate: branch protection for kubeflow/pipelines requires the `ci-passed` status context, and the Dependabot Tide query no longer requires the label (label informational) — in the coupled config PR oss-test-infra#2660, commit [8ca666ca](https://github.com/aholten/oss-test-infra/commit/8ca666cabae85257e19ae1edb40fad7e6695baa1).

jeffspahr (changes requested):
Requesting changes at head `3387cdac712f5269ad01381739dabcdab7b8828d`.

Merging this head does not safely close the already-live rollout gap. As of 2026-08-28, [the review-free Dependabot Tide query is live](https://oss.gprow.dev/tide.js?var=tideData), but live `master` protection still requires only app-bound `DCO`, not `ci-passed`. An absent `ci-passed` status is therefore not currently a gate. 

1. **[P1] Trusted contributors cannot produce the proposed required status.** Both [CI Check eligibility](https://github.com/kubeflow/pipelines/blob/3387cdac712f5269ad01381739dabcdab7b8828d/.github/workflows/ci-checks.yml#L45-L58) and [downstream publication eligibility](https://github.com/kubeflow/pipelines/blob/3387cdac712f5269ad01381739dabcdab7b8828d/.github/workflows/add-ci-passed-label.yml#L142-L152) accept only `ok-to-test` or exact Dependabot. `MEMBER`/`OWNER`/`COLLABORATOR` PRs normally bypass `ok-to-test`; recent member PR #14163 followed that path. Once `ci-passed` is required, those PRs can never satisfy branch protection. Mirror the trusted-author-association rule in both stages while preserving `needs-ok-to-test` precedence.

2. **[P1] `ci-passed` is a one-way success latch.** The workflow [only writes `success`](https://github.com/kubeflow/pipelines/blob/3387cdac712f5269ad01381739dabcdab7b8828d/.github/workflows/add-ci-passed-label.yml#L154-L169), while reset and failure paths [only remove the label or return](https://github.com/kubeflow/pipelines/blob/3387cdac712f5269ad01381739dabcdab7b8828d/.github/workflows/add-ci-passed-label.yml#L171-L206). Adding `needs-ok-to-test`, removing `ok-to-test`, reopening, retargeting, or a later same-SHA CI failure leaves the required status green. The Dependabot Tide query also does not exclude `needs-ok-to-test`. Publish a pending/failing status on invalidation and revalidation, handle base-ref edits, serialize competing publications, and keep `needs-ok-to-test` as an independent Tide blocker.

3. **[P1] The coupled branch-protection rule is repository-scoped, not `master`-scoped.** [The merged configuration](https://github.com/GoogleCloudPlatform/oss-test-infra/blob/1b83c7c6b7ad134d6ab89f3176dff86e2ab49a87/prow/oss/config.yaml#L253-L260) inherits `ci-passed` across protected Pipelines branches. Release branches have older workflow definitions that do not publish this status, so reconciliation would block backport PRs. Scope the requirement under `branches: master`, or backport the complete publisher first.

4. **[P2] The label head check remains a read-then-write race.** [`gh pr view` and `gh pr edit`](https://github.com/kubeflow/pipelines/blob/3387cdac712f5269ad01381739dabcdab7b8828d/.github/workflows/add-ci-passed-label.yml#L221-L230) are separate operations. A synchronize can occur after the read, its reset can finish, and the stale run can then label the new head. Either remove label consumers and make it genuinely informational, or serialize/reconcile label mutations.

5. **[P2] The regression tests do not verify the claimed workflow ordering/race.** [`test_eligibility_step_author_branch_present_and_ordered`](https://github.com/kubeflow/pipelines/blob/3387cdac712f5269ad01381739dabcdab7b8828d/.github/resources/scripts/ci_checks_test.py#L146-L153) finds the first label occurrences in the outer job filter, not the executable eligibility branches, and never asserts `needs < ok`. The race test models only equality at read time, not an event between the production API calls. Scope assertions to the executable block and add state-transition/interleaving coverage.

Recommended rollout: temporarily restore the prior Tide/human gate; fix and merge the publisher while that gate remains; verify `ci-passed` is required on live `master` only; then relax Dependabot reviews. The informational-label race and structural test cleanup can follow only if no merge path consumes that label.

Verification: the focused 12-test suite passes, all three changed YAML files parse, and `git diff --check` passes. The current GitHub check rollup does not include the CI Scripts Tests workflow.

aholten:
Backporting to the release branches seems messy, opened https://github.com/GoogleCloudPlatform/oss-test-infra/pull/2674

jeffspahr (changes requested):
Re-review at head `6dc4cb42eedacb68d9a254f3783224b8033f1ce0`.

The prior trusted-author eligibility, `needs-ok-to-test` precedence, executable branch-order coverage, and post-label SHA-reconcile findings are addressed. Requesting changes for the remaining merge-gate blockers:

1. **[P1] Status publication and resets are not serialized.** The [concurrency key](https://github.com/kubeflow/pipelines/blob/6dc4cb42eedacb68d9a254f3783224b8033f1ce0/.github/workflows/add-ci-passed-label.yml#L20-L22) expects `workflow_run.pull_requests[0].number`, but the actual [CI Check run payload](https://api.github.com/repos/kubeflow/pipelines/actions/runs/33208738481) has `pull_requests: []`. A workflow-run receiver therefore falls back to its unique run ID while PR reset events use the PR number, so a stale run can publish success after a reset. Also, default GitHub concurrency retains only one pending run; a later no-op label event can replace a queued security reset. Use a key available identically to both event types, and either configure a real queue or make every surviving run reconcile current state.

2. **[P1] The failure-status path is unreachable.** Non-successful CI runs [return before reading the artifact](https://github.com/kubeflow/pipelines/blob/6dc4cb42eedacb68d9a254f3783224b8033f1ce0/.github/workflows/add-ci-passed-label.yml#L54-L59), and false artifact results [return before setting the PR/SHA outputs](https://github.com/kubeflow/pipelines/blob/6dc4cb42eedacb68d9a254f3783224b8033f1ce0/.github/workflows/add-ci-passed-label.yml#L111-L114). If parsing continues, `eligible == true` necessarily makes `should_add_label == true`, so the [failure condition](https://github.com/kubeflow/pipelines/blob/6dc4cb42eedacb68d9a254f3783224b8033f1ce0/.github/workflows/add-ci-passed-label.yml#L177-L192) can never fire. A failed same-SHA rerun can leave an older `ci-passed: success` green. Resolve the PR/SHA for unsuccessful runs and explicitly publish failure or pending.

3. **[P1] Base-retarget invalidation is malformed.** The [reset predicate](https://github.com/kubeflow/pipelines/blob/6dc4cb42eedacb68d9a254f3783224b8033f1ce0/.github/workflows/add-ci-passed-label.yml#L203-L224) interpolates `github.event.changes.base.ref`, which is an object containing `.from`, rather than a string. Use `github.event.changes.base.ref.from`; otherwise a retarget can retain success obtained against the prior base. After that reset is corrected, [CI Check still has no `edited` trigger](https://github.com/kubeflow/pipelines/blob/6dc4cb42eedacb68d9a254f3783224b8033f1ce0/.github/workflows/ci-checks.yml#L6-L9), so define a safe revalidation/recovery path for the new base rather than leaving the PR permanently failed or reusing stale results.

4. **[P1] The coupled configuration remains incomplete.** [oss-test-infra#2674](https://github.com/GoogleCloudPlatform/oss-test-infra/pull/2674) scopes branch protection to `master`, but leaves the review-free Dependabot Tide query repository-wide. Dependabot PRs targeting release branches would consequently have neither review nor `ci-passed` gating. Scope that query to `master`. The overlapping generic Kubeflow Tide query also omits `needs-ok-to-test`, so add the blocker to every query that can match Pipelines PRs.

The new focused tests pass, but their concurrency and failure checks are structural or hand-modeled and do not execute the event payload and output transitions above. Verification: 12 focused tests passed, all three changed workflow YAML files parsed, and `git diff --check` passed.

Given the already-live review-free Dependabot query, keep or restore a hard temporary Tide hold/review gate, land and verify the corrected publisher, confirm `ci-passed` is required on live `master` only, and only then enable unattended merges.

jeffspahr (changes requested):
Re-review at head `7090bdd18d8a10279ed1062a26d053991063858c`.

The sole-publisher rewrite resolves the prior cross-workflow concurrency, unreachable failure-status, and malformed base-edit-payload findings. The coupled Tide change also now scopes unattended Dependabot merges to `master` and adds `needs-ok-to-test` to every overlapping query. Requesting changes for the remaining merge-gate blockers:

1. **[P1] Later same-SHA CI reruns cannot invalidate an already-green aggregate.** [CI Check listens only to pull-request events](https://github.com/kubeflow/pipelines/blob/7090bdd18d8a10279ed1062a26d053991063858c/.github/workflows/ci-checks.yml#L7-L9). After it publishes `ci-passed: success`, rerunning a constituent workflow or job on the same SHA does not trigger the pending transition or reconciliation. If that rerun becomes pending or fails, branch protection can continue seeing the old aggregate success. Observe trusted workflow/check completion events and reconcile the same SHA, or retain the underlying required contexts instead of caching them indefinitely.

2. **[P1] Base-retarget invalidation is event-local and can be reversed without fresh CI.** [The `edited` event blocks polling only while that payload carries the old base](https://github.com/kubeflow/pipelines/blob/7090bdd18d8a10279ed1062a26d053991063858c/.github/workflows/ci-checks.yml#L35-L67). A later `ok-to-test` label or `needs-ok-to-test` removal on the unchanged head takes the normal eligible path, polls head-SHA checks produced for the prior base, and can restore success without a `synchronize`. GitHub's normal `pull_request` workflows [do not run on label events by default](https://docs.github.com/en/actions/reference/workflows-and-actions/events-that-trigger-workflows#pull_request), so those events do not prove the new merge result was tested. Persist the revalidation requirement until a head update/fresh run, or bind the evidence to the tested base/merge SHA. The test named `until_synchronize` currently evaluates two isolated inputs and does not model that persistent state.

3. **[P1] A base or eligibility change can race final success publication.** [Reconciliation reads live PR state once, then publishes success and adds the label](https://github.com/kubeflow/pipelines/blob/7090bdd18d8a10279ed1062a26d053991063858c/.github/workflows/ci-checks.yml#L162-L210). [The second read checks only `head.sha`](https://github.com/kubeflow/pipelines/blob/7090bdd18d8a10279ed1062a26d053991063858c/.github/workflows/ci-checks.yml#L211-L223), ignoring a changed base, closed state, or newly revoked eligibility. A same-head change after the first read can therefore leave stale success visible until the replacement run converges. The proposed Tide `needs-ok-to-test` blocker mitigates that label case after #2674 lands, but it does not validate base freshness. Reconcile the full state and explicitly fail the status when the post-mutation snapshot no longer matches.

4. **[P1] Coupled PR #2674 would drop strict branch protection.** Its [`required_status_checks` block](https://github.com/GoogleCloudPlatform/oss-test-infra/blob/852a7378751a559b44be3b93ed6836c436b64466/prow/oss/config.yaml#L258-L266) omits `strict: true`; [Prow renders an omitted value as `false`](https://github.com/kubernetes-sigs/prow/blob/69388feccbd9d4cdc62ea186db0fd50ac633bf69/cmd/branchprotector/request.go#L62-L71), whereas live `master` protection is currently strict. Add `strict: true` and verify the applied rule keeps that setting.

There is also a rollout blocker: [live Tide](https://oss.gprow.dev/tide.js?var=tideData) already has the review-free Dependabot query, while live `master` protection still requires only app-bound DCO, not `ci-passed`. This current unattended exposure is specific to Dependabot; the human Tide paths still require `lgtm` and `approved`. Merging this publisher alone does not make its status enforceable. Restore a temporary review/hold gate, deploy and exercise the publisher on a real eligible PR, verify an app-bound `ci-passed` status, apply and verify strict `master` protection, and relax Tide last.

Verification: the focused 12-test suite passes, relevant workflow YAML parses, and `git diff --check` passes. The hosted CI Scripts Tests run remains `action_required`, so this new workflow implementation has not executed on GitHub yet.

jeffspahr (changes requested):
Request changes at `3ba0168a18e69498917612f3e3dd4bd1e228bbbc`:

1. **[P1] Invalid workflow trigger prevents execution.** [`ci-checks.yml:12`](https://github.com/kubeflow/pipelines/blob/3ba0168a18e69498917612f3e3dd4bd1e228bbbc/.github/workflows/ci-checks.yml#L12) declares `workflow_run` without the required `workflows` selector. Actionlint v1.7.12 rejects it; the [hosted run](https://github.com/kubeflow/pipelines/actions/runs/33636571533) failed with zero jobs. Add the constituent workflow selector.

2. **[P1] Base-retarget invalidation remains unhandled.** The [PR event list](https://github.com/kubeflow/pipelines/blob/3ba0168a18e69498917612f3e3dd4bd1e228bbbc/.github/workflows/ci-checks.yml#L11) omits `edited`. Retargeting an already-green head therefore preserves success earned against the previous base. Restore invalidation and persist it until fresh CI validates the new base. This is an outstanding prior finding.

3. **[P1] Cancelled same-SHA reruns preserve stale success.** The [reconciliation guard](https://github.com/kubeflow/pipelines/blob/3ba0168a18e69498917612f3e3dd4bd1e228bbbc/.github/workflows/ci-checks.yml#L482) ignores cancelled completions, although the pinned poller treats cancelled checks as failures. Executing the actual script confirms no status update occurs. Re-poll these completions. This is a remaining case of the prior rerun finding.

Verification: 21 eligibility tests and 3 concurrency tests pass; whitespace checks pass. Workflow validation fails as above.

Live `master` protection still requires only DCO, and coupled PR #2674 remains open, so `ci-passed` is not yet enforced.

aholten:
I’ve pushed the gate revision and merged current master. The PR description maps each finding to its fix and validation. The prior revision passed workflow validation, 281 CI-script tests, and 7 JUnit tests in the linked hosted run. DCO now passes, and the coupled infrastructure PR’s CLA issue is fixed without changing its configuration.

The new head’s test workflows currently show action_required; could you approve them and re-review the updated gate contract? Dependabot holds remain until deployed publisher behavior and live merge enforcement are verified.

.github/workflows/ci-checks.yml
```diff
@@ -1,171 +1,115 @@
...
-          polling_interval: '5'
-          # Exclude checks dynamically created by the GitHub Copilot pull request reviewer.
-          # These checks fail with HTTP 403 on fork PRs due to permission limitations,
-          # causing false negatives in the status gate. If the Copilot reviewer adds or
-          # renames checks, update this list accordingly.
-          checks_exclude: '^Cleanup artifacts$,^Upload results$,^Agent$,^Prepare$'
+          commit_sha: ${{ steps.prepare.outputs.head_sha }}
+          # Completion events drive recovery. Never hold the shared writer
+          # for a long poll while a rerun or security reset waits behind it.
+          poll: 'false'
```
jeffspahr, line 131: **[P2] Restore recovery when external checks finish last.** With `poll: 'false'`, an external check such as DCO still pending at the final reconciliation causes `ci-passed` to fail. Its later success emits neither a selected `workflow_run` event nor a relevant PR event, so the gate remains failed despite all checks succeeding until someone manually reruns CI Check or causes another eligible event. Previously, polling could observe that completion. Please add external-check completion reconciliation or another bounded retry mechanism that does not hold the shared writer, with coverage for pending → success after the last constituent workflow event.

jeffspahr (changes requested):
Re-review at head `0863d897b7a1c54fbaa7b8df21a9e7f0c93d7012`.

One new P2 recovery finding is attached inline: external checks finishing after the final reconciliation cannot restore `ci-passed`. Earlier implementation findings appear addressed.

Verification: all 281 CI-script tests passed across runs, plus 7 JUnit-summary tests and actionlint 1.7.12 on both changed workflows. The finding was verified through production publisher/event wiring and pinned poller source inspection; the live lifecycle was not exercised. Upstream PR workflows still await approval. Deployed publisher behavior and live merge enforcement remain rollout prerequisites before lifting Dependabot holds.

jeffspahr:
@aholten Please include removal of the automatic Dependabot `do-not-merge/hold` creation gate introduced in #14249 in this PR's scope, rather than leaving it to an unspecified follow-up. The current diff retains that gate, so merging this alone would still leave new Dependabot PRs held.

Please update the implementation, regression coverage, and rollout documentation together. Coordinate activation of the hold removal with verification of the publisher lifecycle and the live Tide/branch-protection configuration from https://github.com/GoogleCloudPlatform/oss-test-infra/pull/2674 so the replacement gate is enforced before automatic holds stop. Preserve existing PR-specific holds for individual review.

PR #14136, opened 2026-08-23
fix(deployment): require edit access to start or delete TensorBoard viewers

herikwebb:

Viewer creation launches a caller-selected image from a caller-supplied PodTemplateSpec, so it is a workload-creation capability rather than a read-only operation. The pipelines view aggregate role currently grants Viewer create and delete even though it does not grant equivalent Pod or Workflow authoring access.

This change moves Viewer create and delete to the pipelines edit aggregate role, whose users can already create Argo Workflows. Read-only users retain Viewer get access. Existing custom images, pod templates, volume credentials, and Workload Identity behavior remain unchanged.

Testing:
- Rendered manifests/kustomize/base/installs/multi-user with Kustomize v5.2.1.
- Verified the rendered edit role grants Viewer create, delete, and get.
- Verified the rendered view role grants Viewer get only.
- git diff --check

**Checklist:**
- [x] The title for your pull request (PR) follows the project title convention.

### DIFF
--- a/manifests/kustomize/base/installs/multi-user/view-edit-cluster-roles.yaml
+++ b/manifests/kustomize/base/installs/multi-user/view-edit-cluster-roles.yaml
@@ -87,6 +87,13 @@ rules:
   resources:
   - scheduledworkflows
 - apiGroups:
+  - kubeflow.org
+  resources:
+  - viewers
+  verbs:
+  - create
+  - delete
+- apiGroups:
   - argoproj.io
   verbs:
   - '*'
@@ -131,9 +138,7 @@ rules:
   resources:
   - viewers
   verbs:
-  - create
   - get
-  - delete
 - apiGroups:
   - pipelines.kubeflow.org
   resources:


jeffspahr:
@herikwebb can you change the title so it's very clear in the release notes since this could be a breaking change for some folks:

> fix(deployment): require edit access to start or delete TensorBoard viewers, since fix titles automatically enter the changelog.

herikwebb:
@jeffspahr updated

dobbydobap:
Traced this end to end and it does what it says. The frontend authorizes per verb against the caller, not blanket-GET: `create` checks `AuthorizeRequestVerb.CREATE` and `deleteHandler` checks `DELETE` (tensorboard.ts:130 and :188). So a view-only user's SubjectAccessReview genuinely fails now, even though `ml-pipeline-ui` itself still holds `viewers: create/delete` and calls the API as its own service account.

Both hunks land where the description says: the added rule is under `aggregate-to-kubeflow-pipelines-edit`, the removal under `aggregate-to-kubeflow-pipelines-view`.

One small thing, not a blocker. `kubeflow-pipelines-view` carries the `aggregate-to-kubeflow-pipelines-edit: "true"` label, so edit already aggregates everything view has. The `get` you added to the edit rule is therefore redundant, edit picks it up from view anyway. Harmless, and arguably clearer being explicit, just noting it in case it was meant to be doing work.

herikwebb:
@dobbydobap Good catch—removed the redundant explicit `get` permission in 22723ed29. The edit role now grants only `create` and `delete` for Viewers and inherits `get` through the view-role aggregation.

PR #14138, opened 2026-08-23
fix(frontend): return TensorBoard errors as plain text

herikwebb:

TensorBoard handler errors can contain caller-derived or downstream text. Express string responses otherwise default to an HTML content type, allowing reflected markup in an error body to be interpreted as HTML by a browser.

This change gives lookup, create, malformed podtemplatespec, and delete errors an explicit text/plain content type. The integration tests cover each affected error path.

Testing:
- vitest run --root frontend/server integration-tests/tensorboard.test.ts (26 tests passed)
- tsc --project frontend/server
- prettier --check on both changed files
- git diff --check

**Checklist:**
- [x] The title for your pull request (PR) follows the project title convention.

### DIFF
--- a/frontend/server/handlers/tensorboard.ts
+++ b/frontend/server/handlers/tensorboard.ts
@@ -55,7 +55,7 @@ export const getTensorboardHandlers = (
         req,
       );
       if (authError) {
-        res.status(401).send(authError.message);
+        res.status(401).type('text/plain').send(authError.message);
         return;
       }
       const tensorboardInstance = await k8sHelper.getTensorboardInstance(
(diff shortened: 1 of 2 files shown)


jeffspahr (commented):
[P1] Authorization failures still render as HTML

All three authorization-failure branches still call res.status(...).send(authError.message), including GET at line 58 and the corresponding CREATE and DELETE paths. Express 5.2.1 defaults string responses to text/html, and getAuthorizeFn includes the parsed auth API error message verbatim. An auth backend rejection containing markup can therefore still be interpreted as HTML on direct navigation, leaving the reflected-markup fix incomplete.

Please set text/plain on all three authorization-error responses and extend the existing "rejects user requests when KFP auth api rejected" integration test with markup and a content-type assertion.

droctothorpe:
@herikwebb: Hey, thank you for your PR! As you are not a member of Kubeflow organization, your PR must link to an issue with 'ready' label. Please read our [CONTRIBUTING.md](https://github.com/kubeflow/pipelines/blob/master/CONTRIBUTING.md), closing this PR

PR #14145, opened 2026-08-24
fix(frontend): import produce as a named export for immer 11

jeffspahr:
## Summary

Bump `immer` from 9.x to 11.1.18 and switch every import to the named `produce` export.

Supersedes #14122.

## Why

immer 11 removed the default export. `produce` is now available only as a named export:

```ts
export { Draft, Immer, ..., produce, produceWithPatches, ... };
```

Every `immer` import in the frontend was a default import, so all six became invalid.

Three are in application code, and these are the errors reported on #14122:

```
src/components/ExperimentList.tsx(177,32): error TS2349: This expression is not callable.
src/components/ExperimentList.tsx(177,77): error TS7006: Parameter 'draft' implicitly has an 'any' type.
src/pages/ExperimentList.tsx(265,32): error TS2349: This expression is not callable.
src/pages/ExperimentList.tsx(265,77): error TS7006: Parameter 'draft' implicitly has an 'any' type.
src/pages/PipelineList.tsx(139,30): error TS2349: This expression is not callable.
src/pages/PipelineList.tsx(139,73): error TS7006: Parameter 'draft' implicitly has an 'any' type.
```

The remaining three are in test files:

- `src/pages/RecurringRunList.test.tsx`
- `src/pages/RunList.test.tsx`
- `src/pages/RunListsRouter.test.tsx`

Those did **not** surface as type errors, despite calling `produce` the same way. Fixing only the six reported errors would have left those three importing a binding that no longer exists, so they are included here.

Application call sites keep their local `immerProduce` name via `produce as immerProduce`, which keeps the diff to the import line in each file.

## Verification

- `format:check`, `lint:ui`, `typecheck` — clean, 0 errors
- 136 tests passing across the affected files and their component suites: `RunList`, `RecurringRunList`, `RunListsRouter`, `components/ExperimentList`, `pages/ExperimentList`, `pages/PipelineList`

Leaving the full suite to CI.

### DIFF
--- a/frontend/package-lock.json
+++ b/frontend/package-lock.json
@@ -24,7 +24,7 @@
         "google-protobuf": "^4.0.2",
         "grpc-web": "^2.1.1",
         "http-proxy-middleware": "^3.0.7",
-        "immer": "^9.0.6",
+        "immer": "^11.1.18",
         "js-yaml": "^4.3.1",
         "lodash": "^4.18.1",
         "lodash.debounce": "^4.0.8",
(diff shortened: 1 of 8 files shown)


manaswinidas (approved):
/lgtm
/approve

PR #14146, opened 2026-08-24
fix(frontend): narrow js-yaml load results for js-yaml 5

jeffspahr:
## Summary

Bump `js-yaml` from 4.x to 5.3.0 and narrow the parse results it now returns as `unknown`.

Supersedes #14083.

## Why

js-yaml 5 types `load` as returning `unknown` rather than `any`. That is a genuine improvement — the result of parsing arbitrary YAML really is unknown — but it does not type check against the existing call sites:

```
src/pages/NewRunV2.tsx(423,7): error TS2322: Type 'unknown' is not assignable to type 'object | undefined'.
src/pages/PipelineDetails.tsx(661,50): error TS2345: Argument of type 'unknown' is not assignable to parameter of type 'Workflow'.
src/pages/PipelineDetails.tsx(662,49): error TS2345: Argument of type '{}' is not assignable to parameter of type 'Workflow'.
src/pages/__tests__/newRunTestFixtures.ts(54,3): error TS2322: Type 'unknown' is not assignable to type 'object | undefined'.
```

## isArgoWorkflowTemplate

The `PipelineDetails` errors are worth fixing properly rather than casting away.

`isArgoWorkflowTemplate` was declared as taking a `Workflow`:

```ts
export function isArgoWorkflowTemplate(template: Workflow): boolean {
  if (template?.kind === 'Workflow' && template?.apiVersion?.startsWith('argoproj.io/')) {
```

but its body only does optional-chained property reads, so it was always safe against an arbitrary value. Two existing call sites already acknowledged that by working around the signature with `as Workflow`.

Redeclaring it as a type guard over `unknown` matches what the function actually does:

```ts
export function isArgoWorkflowTemplate(template: unknown): template is Workflow
```

This lets `PipelineDetails` pass the parse result straight in, and narrows `template` to `Workflow` for the `createGraph` call on the next line — so both errors are resolved by the guard rather than by an assertion. The two `as Workflow` casts in `WorkflowUtils` are now redundant and are removed.

The two `pipeline_spec` assignments take `object | undefined`, and there is no meaningful narrowing to do beyond that, so those results are asserted to `object` at the point of use.

## Verification

- `format:check`, `lint:ui`, `typecheck` — clean, 0 errors
- 86 tests passing across `WorkflowUtils`, `PipelineDetails`, and `NewRunV2`

Leaving the full suite to CI.

### DIFF
--- a/frontend/package.json
+++ b/frontend/package.json
@@ -23,7 +23,7 @@
     "grpc-web": "^2.1.1",
     "http-proxy-middleware": "^3.0.7",
     "immer": "^11.1.18",
-    "js-yaml": "^4.3.1",
+    "js-yaml": "^5.4.1",
     "lodash": "^4.18.1",
     "lodash.debounce": "^4.0.8",
     "lodash.flatten": "^4.4.0",
--- a/frontend/server/package.json
+++ b/frontend/server/package.json
@@ -13,7 +13,7 @@
     "google-protobuf": "^4.0.2",
     "gunzip-maybe": "^1.4.1",
     "http-proxy-middleware": "^2.0.10",
-    "js-yaml": "^4.3.0",
+    "js-yaml": "^5.3.0",
     "lodash": ">=4.18.1",
     "minio": "~8.0.7",
     "peek-stream": "^1.1.3",
--- a/frontend/src/components/PipelineSpecTabContent.tsx
+++ b/frontend/src/components/PipelineSpecTabContent.tsx
@@ -14,7 +14,8 @@
  * limitations under the License.
  */
 
-import { dump, load } from 'js-yaml';
+import { dump } from 'js-yaml';
+import { loadYaml } from 'src/lib/YamlLoad';
 import { isSafari } from 'src/lib/Utils';
 import Editor from './Editor';
 
@@ -27,7 +28,7 @@ const editorHeightWidth = isSafari() ? '640px' : '100%';
 export function PipelineSpecTabContent(props: PipelineSpecTabContentProps) {
   return (
     <Editor
-      value={dump(load(props.templateString || ''))}
+      value={dump(loadYaml(props.templateString || ''))}
       // Render the yaml-formatted string in <PipelineSpecTabContent>
       // V1(JSON-formatted):
       //    load() converts templateString to an object first,
--- a/frontend/src/lib/YamlLoad.ts
+++ b/frontend/src/lib/YamlLoad.ts
@@ -1,0 +1,30 @@
+/*
+ * Copyright 2026 The Kubeflow Authors
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import { CORE_SCHEMA, load as jsYamlLoad, mergeTag } from 'js-yaml';
+
+/**
+ * js-yaml 5 resolves merge keys only when `mergeTag` is present in the schema,
+ * where js-yaml 4 resolved them by default. KFP parses Argo and Kubernetes
+ * manifests that inherit fields through anchors, so without this a merged
+ * mapping keeps a literal `<<` property and the inherited fields are lost.
+ */
+const MANIFEST_SCHEMA = CORE_SCHEMA.withTags(mergeTag);
+
+/** Parses manifest YAML, preserving merge key (`<<`) resolution. */
+export function loadYaml(text: string): unknown {
+  return jsYamlLoad(text, { schema: MANIFEST_SCHEMA });
+}
--- a/frontend/src/lib/YamlLoad.ts
+++ b/frontend/src/lib/YamlLoad.ts
@@ -14,7 +14,7 @@
  * limitations under the License.
  */
 
-import { CORE_SCHEMA, load as jsYamlLoad, mergeTag } from 'js-yaml';
+import { CORE_SCHEMA, load as jsYamlLoad, loadAll as jsYamlLoadAll, mergeTag } from 'js-yaml';
 
 /**
  * js-yaml 5 resolves merge keys only when `mergeTag` is present in the schema,
@@ -27,14 +27,31 @@ const MANIFEST_SCHEMA = CORE_SCHEMA.withTags(mergeTag);
 /**
  * Parses manifest YAML, preserving merge key (`<<`) resolution.
  *
- * Returns undefined for input that holds no document. js-yaml 5 throws
- * `YAMLException: expected a document, but the input is empty` in that case,
- * where js-yaml 4 returned undefined. Call sites funnel optional template
- * strings through `|| ''`, so throwing would surface during render.
+ * Restores the js-yaml 4 result for input that carries no document. js-yaml 5
+ * throws `YAMLException: expected a document, but the input is empty` for a
+ * blank string, comment-only text, or a bare `...` marker, where js-yaml 4
+ * returned undefined for blank input and null otherwise. Call sites parse
+ * arbitrary template strings during render, so throwing is reachable.
+ *
+ * loadAll reports zero documents for exactly those inputs while still raising
+ * genuine syntax errors, which is what distinguishes the two cases without
+ * matching on exception messages. Input holding more than one document is
+ * handed back to `load` so it raises its own single-document error rather than
+ * silently yielding the first.
  */
 export function loadYaml(text: string): unknown {
   if (!text.trim()) {
     return undefined;
+  }
+
+  const documents: unknown[] = [];
+  jsYamlLoadAll(text, (document) => documents.push(document), { schema: MANIFEST_SCHEMA });
+
+  if (documents.length === 0) {
+    return null;
+  }
+  if (documents.length === 1) {
+    return documents[0];
   }
   return jsYamlLoad(text, { schema: MANIFEST_SCHEMA });
 }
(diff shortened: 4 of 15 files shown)


frontend/package.json
```diff
@@ -23,7 +23,7 @@
     "grpc-web": "^2.0.2",
     "http-proxy-middleware": "^3.0.7",
     "immer": "^9.0.6",
-    "js-yaml": "^4.3.1",
+    "js-yaml": "^5.3.0",
```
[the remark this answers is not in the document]
jeffspahr: Confirmed and fixed in ab8a6c1. This was a real runtime regression, not just a typing question — reproduced directly against js-yaml 5.3.0:

```js
yaml.load(`
base: &base
  keyFormat: "artifacts/{{workflow.name}}"
s3:
  <<: *base
  bucket: mybucket
`)
// { base: {...}, s3: { '<<': { keyFormat: ... }, bucket: 'mybucket' } }
// s3.keyFormat === undefined, and a literal '<<' key is present
```

`CORE_SCHEMA.withTags(mergeTag)` restores it.

Rather than repeat the schema at each call site, application parses now go through a shared `loadYaml` helper in `src/lib/YamlLoad.ts`. The affected production parses were:

- `src/lib/v2/WorkflowUtils.ts` (two)
- `src/pages/PipelineDetails.tsx`
- `src/pages/NewRunV2.tsx`
- `src/components/PipelineSpecTabContent.tsx`

The other `js-yaml` importers in `src/` only call `dump`, which is unaffected.

Regression coverage added in `YamlLoad.test.ts`: merge keys resolve, the merged mapping still overrides inherited keys, and documents without merge keys parse unchanged.

frontend/server/package.json
```diff
@@ -13,7 +13,7 @@
     "google-protobuf": "^4.0.2",
     "gunzip-maybe": "^1.4.1",
     "http-proxy-middleware": "^2.0.10",
-    "js-yaml": "^4.3.0",
+    "js-yaml": "^5.3.0",
```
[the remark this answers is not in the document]
jeffspahr: Confirmed and fixed in ab8a6c1.

`getKeyFormatFromArtifactRepositories` parses a user supplied configmap, so this was the more consequential of the two — an inherited `keyFormat` silently became `undefined`, and the function would fall through to its "missing one of [s3|gcs|oss|artifactory] fields" error or return nothing useful.

`frontend/server` has its own dependency tree, so it sets the schema locally at that parse rather than importing the frontend helper:

```ts
const artifactRepositoriesValue = jsYamlLoad(artifactRepositories, {
  schema: CORE_SCHEMA.withTags(mergeTag),
}) as PartialArtifactRepositoriesValue;
```

Added the anchored-configmap case you asked for to `workflow-helper.test.ts`, alongside the existing flat one: an `s3` entry inheriting `keyFormat` through `<<: *common`, asserting the inherited value is returned.

The server suite needs a live Kubernetes cluster, so I verified this change with `tsc --project server` and `lint:server` locally and left the test execution to CI.

frontend/src/lib/v2/WorkflowUtils.ts
```diff
@@ -45,8 +45,9 @@ export function isV2Pipeline(workflow: Workflow): boolean {
   return workflow?.metadata?.annotations?.['pipelines.kubeflow.org/v2_pipeline'] === 'true';
 }
 
-export function isArgoWorkflowTemplate(template: Workflow): boolean {
-  if (template?.kind === 'Workflow' && template?.apiVersion?.startsWith('argoproj.io/')) {
+export function isArgoWorkflowTemplate(template: unknown): template is Workflow {
+  const candidate = template as Workflow | undefined;
+  if (candidate?.kind === 'Workflow' && candidate?.apiVersion?.startsWith('argoproj.io/')) {
```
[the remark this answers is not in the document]
jeffspahr: Correct — fixed in ab8a6c1, and reproduced first:

```js
const wf = yaml.load('kind: Workflow\napiVersion: 1\n');
wf?.apiVersion?.startsWith('argoproj.io/');
// TypeError: wf?.apiVersion?.startsWith is not a function
```

Optional chaining only short-circuits on null and undefined, so a numeric `apiVersion` reached `startsWith` and threw. Widening the parameter to `unknown` is what made that reachable from parsed YAML, so it belongs in this PR.

Now checks the type before calling:

```ts
if (
  candidate?.kind === 'Workflow' &&
  typeof candidate.apiVersion === 'string' &&
  candidate.apiVersion.startsWith('argoproj.io/')
) {
```

Added tests covering a numeric `apiVersion`, a non-Argo `apiVersion`, a valid Argo manifest, and non-object input (`undefined` and a bare string).

frontend/src/components/PipelineSpecTabContent.tsx
```diff
@@ -27,7 +28,7 @@ const editorHeightWidth = isSafari() ? '640px' : '100%';
 export function PipelineSpecTabContent(props: PipelineSpecTabContentProps) {
   return (
     <Editor
-      value={dump(load(props.templateString || ''))}
+      value={dump(loadYaml(props.templateString || ''))}
```
jeffspahr, line 31: js-yaml 5 also changed empty-input behavior: loadYaml("") now throws YAMLException, whereas the previous v4 load/dump pair rendered an empty string. This component explicitly normalizes a falsy template to "", and PipelineDetailsV2 still passes its optional template through that fallback, so selecting/rendering the spec view without a template now throws during render. Please return an empty editor value for empty input (either here or in loadYaml) and add an empty-string regression test.

jeffspahr: Confirmed and fixed in 696a44a. Verified the behaviour change directly rather than assuming it:

| | `load('')` | `dump(load(''))` |
|---|---|---|
| js-yaml 4.3.1 | `undefined` | `""` |
| js-yaml 5.3.0 | throws `YAMLException: expected a document, but the input is empty` | throws |

Whitespace-only and newline-only input behave the same way.

Fixed in `loadYaml` rather than at the call site, since auditing the others turned up a second affected one you did not mention: `NewRunV2.tsx:424` uses the same `loadYaml(templateString || '')` shape, where the result becomes `pipeline_spec`. Handling it centrally keeps every parse on the js-yaml 4 contract:

```ts
export function loadYaml(text: string): unknown {
  if (!text.trim()) {
    return undefined;
  }
  return jsYamlLoad(text, { schema: MANIFEST_SCHEMA });
}
```

The remaining sites are unaffected: `PipelineDetails` already guards on `templateString`, and the two `WorkflowUtils` helpers index the parse result immediately, so empty input threw under js-yaml 4 as well — returning `undefined` keeps that unchanged rather than papering over it.

Regression coverage added at both levels:

- `YamlLoad.test.ts` — `undefined` for empty, whitespace, and newline-only input, plus `dump(loadYaml('')) === ''`
- `PipelineSpecTabContent.test.tsx` — renders an empty editor for both an absent and an empty template

Confirmed these actually pin the behaviour by removing the guard: all four fail, each with `YAMLException: expected a document, but the input is empty`.

100 tests passing across YamlLoad, PipelineSpecTabContent, WorkflowUtils, NewRunV2, and PipelineDetails.

frontend/src/lib/YamlLoad.ts
```diff
@@ -0,0 +1,40 @@
...
+/**
+ * Parses manifest YAML, preserving merge key (`<<`) resolution.
+ *
+ * Returns undefined for input that holds no document. js-yaml 5 throws
+ * `YAMLException: expected a document, but the input is empty` in that case,
+ * where js-yaml 4 returned undefined. Call sites funnel optional template
+ * strings through `|| ''`, so throwing would surface during render.
+ */
+export function loadYaml(text: string): unknown {
+  if (!text.trim()) {
```
jeffspahr, line 43: This restores empty/whitespace input, but not the full no-document behavior described above. With exact js-yaml 4.3.1, load("# comment\n") returns null without throwing; this v5 helper passes the nonblank text to jsYamlLoad and throws YAMLException: expected a document, but the input is empty. Multiple comment-only lines and a document-end-only input have the same compatibility gap. PipelineSpecTabContent parses arbitrary template strings during render, so this still leaves a reachable crash. Please cover comment-only input and preserve the previous non-throwing result without swallowing unrelated YAML syntax errors.

jeffspahr: Right on all counts — fixed in 414ff7e. My blank check was too narrow, and comment-only input still threw.

Mapped the actual js-yaml 4 contract first, which is two-valued rather than one:

| input | js-yaml 4 | js-yaml 5 `load` |
|---|---|---|
| `''`, whitespace | `undefined` | throws |
| `# comment` | `null` | throws |
| `# a\n# b` | `null` | throws |
| `...` | `null` | throws |
| `---` | `null` | `null` |

`loadAll` turns out to separate these cleanly without inspecting exception messages: it reports zero documents for exactly the no-document inputs, and still raises real syntax errors.

```ts
if (!text.trim()) {
  return undefined;
}
const documents: unknown[] = [];
jsYamlLoadAll(text, (document) => documents.push(document), { schema: MANIFEST_SCHEMA });
if (documents.length === 0) {
  return null;
}
if (documents.length === 1) {
  return documents[0];
}
return jsYamlLoad(text, { schema: MANIFEST_SCHEMA });
```

The multi-document branch matters for the "without swallowing errors" part: taking `documents[0]` there would have quietly accepted input that `load` rejects with `expected a single document in the stream`. Handing it back to `load` keeps that error.

Rather than assert my own expectations, I verified this as a differential against js-yaml 4.3.1 — same inputs through both, comparing value and throw/no-throw:

```
   empty      v4=OK undefined    loadYaml=OK undefined
   ws         v4=OK undefined    loadYaml=OK undefined
   nl         v4=OK undefined    loadYaml=OK undefined
   comment    v4=OK null         loadYaml=OK null
   comments   v4=OK null         loadYaml=OK null
   docstart   v4=OK null         loadYaml=OK null
   docend     v4=OK null         loadYaml=OK null
   valid      v4=OK {"a":1}      loadYaml=OK {"a":1}
   nested     v4=OK {"a":{"b":2}} loadYaml=OK {"a":{"b":2}}
   merge      v4=OK {"x":{"k":"v"},"y":{"k":"v"}} loadYaml=OK (same)
   invalid    v4=THROW           loadYaml=THROW
   multidoc   v4=THROW           loadYaml=THROW
   tabs       v4=THROW           loadYaml=THROW

PARITY: identical on all cases
```

Tests now cover blank versus no-document separately, the `dump` round trip for both (`''` and `'null\n'`), invalid syntax, tab indentation, and multi-document input. 103 tests passing across the affected suites.

manaswinidas (approved):
/lgtm
/approve

PR #14159, opened 2026-08-24
chore(frontend): upgrade markdown-to-jsx to v9

jeffspahr:
## Summary

- upgrade markdown-to-jsx from 7.7.17 to 9.10.2
- migrate React consumers to the supported markdown-to-jsx/react entry point
- reconcile intentional v9 wrapper and whitespace rendering changes in focused coverage
- replace the stale v7-specific inline-list comment

Supersedes #14123.

## Verification

- npm run format:check
- npm run lint
- npm run typecheck
- npm run check:react-peers
- npm run test:ui -- src/components/Description.test.tsx
- npm run test:ui -- src/components/viewers/MarkdownViewer.test.tsx
- npm run test:ui -- src/pages/GettingStarted.test.tsx
- npm run build

### DIFF
--- a/frontend/package-lock.json
+++ b/frontend/package-lock.json
@@ -31,7 +31,7 @@
         "lodash.flatten": "^4.4.0",
         "lodash.groupby": "^4.6.0",
         "lodash.isfunction": "^3.0.9",
-        "markdown-to-jsx": "^7.0.0",
+        "markdown-to-jsx": "^9.10.2",
         "pako": "^3.0.1",
         "proto3-json-serializer": "^4.0.2",
         "protobufjs": "~8.8.0",
@@ -7586,18 +7586,29 @@
       }
     },
     "node_modules/markdown-to-jsx": {
-      "version": "7.7.17",
-      "resolved": "https://registry.npmjs.org/markdown-to-jsx/-/markdown-to-jsx-7.7.17.tgz",
-      "integrity": "sha512-7mG/1feQ0TX5I7YyMZVDgCC/y2I3CiEhIRQIhyov9nGBP5eoVrOXXHuL5ZP8GRfxVZKRiXWJgwXkb9It+nQZfQ==",
+      "version": "9.10.2",
+      "resolved": "https://registry.npmjs.org/markdown-to-jsx/-/markdown-to-jsx-9.10.2.tgz",
+      "integrity": "sha512-iR9GadlIox0q1uXnpqdxpF02Vb1WDmZ/QIXWjBR5htzjUEEhyIWbM65LuVKavdwJaVo4q95C/F2OOyNjWe91ig==",
       "license": "MIT",
       "engines": {
-        "node": ">= 10"
+        "node": ">= 18"
       },
       "peerDependencies": {
-        "react": ">= 0.14.0"
+        "react": ">= 16.0.0",
+        "solid-js": ">=1.0.0",
+        "vue": ">=3.0.0"
       },
       "peerDependenciesMeta": {
         "react": {
+          "optional": true
+        },
+        "react-native": {
+          "optional": true
+        },
+        "solid-js": {
+          "optional": true
+        },
+        "vue": {
           "optional": true
         }
       }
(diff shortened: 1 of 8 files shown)


jeffspahr:
Reviewed. This closes the question that was left open when #14123 was parked, and I can retract the concern I raised there.

**The import is the right one, and the obvious alternative is wrong.** `markdown-to-jsx` v9 exports `Markdown as default` and `Markdown2 as Markdown` from the package root:

```ts
export { slugify, sanitizer, parser, Markdown as default, ..., Markdown2 as Markdown };
declare const Markdown: React.FC<...>;
/** @deprecated Use the `markdown-to-jsx/react` import instead */
declare var Markdown2: typeof Markdown;
```

So the named `Markdown` at the root is the **deprecated** alias — switching to `import { Markdown } from 'markdown-to-jsx'` would satisfy the lint rule while moving onto a deprecated binding. The `/react` subpath used here exports `Markdown` and `default` as the same non-deprecated symbol, which is also what the upstream deprecation note points at. Correct call.

**Keeping `preserveListMarkers` is right.** I checked whether v9's parser made it redundant: it has not. The bullet content in the snapshots is unchanged (`• abc` / `• def`), and the only snapshot delta is the removed wrapper. Rewording the comment to drop the v7-specific claim is the right adjustment.

**On dropping the `<span>` — I flagged this as a risk on #14123 and it does not hold up.** The worry was the five `forceInline` consumers, all of which are `CustomTable` cell renderers:

```
src/pages/NewPipelineVersion.tsx:120
src/pages/NewRun.tsx:124
src/pages/NewRunV2.tsx:82
src/pages/PipelineList.tsx:55
src/pages/PipelineVersionList.tsx:52
```

The concern was that truncation styling might depend on an element child. It does not — `CustomTable.tsx` applies `overflow: hidden`, `text-overflow: ellipsis`, and `white-space: nowrap` to the cell container itself (lines 86-88 and 96-98), and nothing in the table or shared CSS targets `& > *` or a child `span`. Ellipsis truncation on a block container works the same whether the content is a bare text node or an inline element, so removing the wrapper does not change the rendered result in those cells.

So accepting v9's DOM rather than forcing the old wrapper back with `wrapper: 'span'` / `forceWrapper: true` is the better of the two options: it drops a v7 artifact instead of carrying it forward, and the snapshot churn is confined to the wrapper.

jeffspahr:
Looked at a98f22f. This is a better answer than either option I laid out on #14123, and it also corrects my review above.

**What I got wrong.** I argued the wrapper could be dropped safely because `CustomTable` puts `overflow: hidden` / `text-overflow: ellipsis` on the cell container rather than a child. That was true as far as it went, but I only checked styling. The failures were in test queries: `PipelineDetailsV1`, `PrivateAndSharedPipelines`, and `PipelineVersionCard` locate description text with `getByText`, which matched the wrapper element and then could not find a bare text node:

```
TestingLibraryElementError: Unable to find an element with the text: test-pipeline-desc.
  This could be because the text is broken up by multiple elements...
```

Checking CSS but not queries was the gap.

**The fix is well targeted.** Wrapping only when `compiler` returns `null` or a bare string restores the stable element root exactly where v7 had one, without forcing a wrapper around links, formatting, or block content. That is why `GettingStarted` still correctly loses its span while `Description` keeps its DOM. Leaving the `Description` snapshots untouched is the strongest evidence available — they are the v7 baseline, so if they pass, DOM parity is proven rather than asserted.

**One difference worth recording.** `compiler()` and `<Markdown>` are not quite equivalent in v9. The component reads `MarkdownContext` and merges any provided options before compiling, and spreads leftover props into `wrapperProps`; the bare `compiler()` call does neither.

Nothing depends on that today — `MarkdownProvider` and `MarkdownContext` are new in v9 and unused in this repo, and `Description` passes no extra props. But it does leave the three call sites asymmetric: `MarkdownViewer` and `GettingStarted` keep `<Markdown>` and would honour a future `MarkdownProvider`, while `Description` would silently ignore it. Worth a brief comment at the `compiler` call so that asymmetry is deliberate rather than discovered later.

Also confirming `compiler` is a supported export of the `/react` subpath, so this stays clear of the deprecated root aliases.

manaswinidas (approved):
/lgtm
/approve

PR #14162, opened 2026-08-24
chore(deps): bump http-proxy-middleware to v4.2.0

jeffspahr:
## Summary

- update http-proxy-middleware to v4.2.0 in both frontend packages
- migrate proxy request callbacks to the v4 event API
- migrate the artifact proxy filter to pathFilter
- remove the obsolete logLevel option

## Verification

- frontend/server: npm run build
- frontend/server: npm run format:check
- frontend/server: npm test -- integration-tests/artifact-proxy.test.ts (13 tests passed)

Supersedes #14120.

### DIFF
--- a/frontend/package.json
+++ b/frontend/package.json
@@ -21,7 +21,7 @@
     "dagre": "^0.8.5",
     "google-protobuf": "^4.0.2",
     "grpc-web": "^2.1.1",
-    "http-proxy-middleware": "^3.0.7",
+    "http-proxy-middleware": "^4.2.0",
     "immer": "^11.1.18",
     "js-yaml": "^5.4.1",
     "lodash": "^4.18.1",
--- a/frontend/server/app.ts
+++ b/frontend/server/app.ts
@@ -254,8 +254,10 @@ function createUIServer(options: UIConfigs) {
       '/k8s/pod/logs',
       createProxyMiddleware({
         changeOrigin: true,
-        onProxyReq: (proxyReq) => {
-          console.log('Proxied log request: ', proxyReq.path);
+        on: {
+          proxyReq: (proxyReq) => {
+            console.log('Proxied log request: ', proxyReq.path);
+          },
         },
         headers: HACK_FIX_HPM_PARTIAL_RESPONSE_HEADERS,
         pathRewrite: (pathStr: string, req: any) => {
@@ -286,8 +288,10 @@ function createUIServer(options: UIConfigs) {
       '/k8s/pod/logs',
       createProxyMiddleware({
         changeOrigin: true,
-        onProxyReq: (proxyReq) => {
-          console.log('Proxied log request: ', proxyReq.path);
+        on: {
+          proxyReq: (proxyReq) => {
+            console.log('Proxied log request: ', proxyReq.path);
+          },
         },
         headers: HACK_FIX_HPM_PARTIAL_RESPONSE_HEADERS,
         pathRewrite: (pathStr: string, req: any) => {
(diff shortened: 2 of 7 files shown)


frontend/server/handlers/tensorboard-proxy.ts
```diff
@@ -342,7 +342,6 @@ export default function registerTensorboardProxy(
     proxyRoutes,
     createProxyMiddleware({
```
manaswinidas, line 343: v4 dropped `logLevel` here, but `frontend/server/proxy-middleware.ts` still sets it on the production `/_proxy/` middleware. Remove it there too.

frontend/server/app.ts
```diff
@@ -254,8 +254,10 @@ function createUIServer(options: UIConfigs) {
       '/k8s/pod/logs',
       createProxyMiddleware({
         changeOrigin: true,
-        onProxyReq: (proxyReq) => {
-          console.log('Proxied log request: ', proxyReq.path);
+        on: {
+          proxyReq: (proxyReq) => {
```
manaswinidas: `frontend/mock-backend/mock-api-server.ts` still uses `onProxyReq`. Switch to `on.proxyReq` like this.

frontend/package.json
```diff
@@ -21,7 +21,7 @@
     "dagre": "^0.8.5",
     "google-protobuf": "^4.0.2",
     "grpc-web": "^2.0.2",
-    "http-proxy-middleware": "^3.0.7",
+    "http-proxy-middleware": "^4.2.0",
```
manaswinidas: `mock-backend` resolves this package and still passes `logLevel` in `frontend/mock-backend/proxy-middleware.ts`. Drop it.

manaswinidas (changes requested):
Incomplete v4 migration: three call sites still use removed `logLevel` / `onProxyReq`.

frontend/server/handlers/tensorboard-proxy.ts
jeffspahr: Checked this and I don't think there's anything left to remove — `frontend/server/proxy-middleware.ts` doesn't exist in the tree. The only `createProxyMiddleware` call sites in the frontend are:

- `frontend/server/handlers/tensorboard-proxy.ts` (this one)
- `frontend/server/handlers/artifacts.ts`
- `frontend/server/app.ts` (×2)
- `frontend/mock-backend/mock-api-server.ts`

`grep -rn logLevel frontend/server frontend/mock-backend frontend/src` returns no proxy-related hits on this branch — the only `logLevel` occurrences anywhere are inside `node_modules`. So there's no production `/_proxy/` middleware still setting it.

Happy to take another look if you're seeing this on a different ref — could you point me at the file as it appears for you?

frontend/server/app.ts
jeffspahr: `frontend/mock-backend/mock-api-server.ts` is already on the v4 shape — this PR converts it. As of the current head:

```ts
// frontend/mock-backend/mock-api-server.ts:41
createProxyMiddleware({
  changeOrigin: true,
  on: {
    proxyReq: (proxyReq) => {
      console.log('Metadata proxied request: ', proxyReq.path);
    },
  },
  target: getAddress({ host: 'localhost', port: '9090' }),
  ...
```

There's no `onProxyReq` left in `frontend/server`, `frontend/mock-backend`, or `frontend/src` — the remaining `onError` matches are unrelated (generated OpenAPI runtime and an `app.test.ts` socket listener).

frontend/package.json
jeffspahr: `frontend/mock-backend/proxy-middleware.ts` doesn't exist either — `find frontend -name '*proxy*middleware*' -not -path '*/node_modules/*'` returns nothing, so the only mock-backend proxy setup is `mock-api-server.ts`, which this PR already migrated (see the reply on `app.ts`).

Worth noting `mock-backend` has no `http-proxy-middleware` entry of its own; it resolves the hoisted copy from `frontend/package.json`, so this bump covers it and there's no second `logLevel` to drop.

If you're seeing these files locally, they may be from a stale tree or a different branch — let me know and I'll recheck.

manaswinidas (approved):
/lgtm
/approve

manaswinidas (approved):
/lgtm

PR #14168, opened 2026-08-25
fix(backend): prevent retrying archived runs. Fixes #14164

### LINKED ISSUE #14164: bug(backend): RetryRun should reject archived runs to prevent GC lifecycle conflict
## Summary

The `RetryRun` API does not check whether a run's `StorageState` is `ARCHIVED` before allowing the retry. With the introduction of the run garbage collector in #13532, archived runs are candidates for permanent deletion after `ARCHIVED_RUNS_RETENTION_TIME` expires. Retrying an archived run creates a semantic conflict: the GC could delete a run that has been retried and is now actively running again.

## Current Behavior

`RetryRun` in [`resource_manager.go`](https://github.com/kubeflow/pipelines/blob/master/backend/src/apiserver/resource/resource_manager.go) checks:
- Workflow manifest is non-empty
- Workflow allows retries (`CanRetry`)
- Retry claim / generation logic (added in #13532)

It does **not** check `run.StorageState`. An archived run can be retried, which:
1. Resets the run to `PENDING` state
2. Creates a new workflow CR
3. But leaves `StorageState = ARCHIVED` — making it a deletion candidate for the GC

## Expected Behavior

`RetryRun` should return an error (e.g., `400 Bad Request`) if the run's `StorageState` is `ARCHIVED`. The user should unarchive the run first, then retry it. This preserves the invariant that archived runs are inactive and eligible for GC deletion.

Alternatively, `RetryRun` could automatically unarchive the run as part of the retry operation, clearing `StorageState` back to `AVAILABLE` and resetting `ArchivedAtInSec` to 0.

## Impact

Without this guard:
- A retried archived run could be deleted mid-execution by the GC
- The `ArchivedAtInSec` timestamp from the original archival would be used for deletion timing, not accounting for the retry
- The two-stage lifecycle invariant (`Active → Archived → Deleted`) is violated

## Context

- GC PR: https://github.com/kubeflow/pipelines/pull/13532
- jeffspahr's round-5 review established the retention lifecycle semantics
- The `ArchiveRun` / `UnarchiveRun` APIs exist and work correctly
- Follow-up from #13532 review discussion

## Suggested Fix

In `ResourceManager.RetryRun()`, after `GetRun()`, add:

```go
if run.StorageState.ToV2() == model.StorageStateArchived {
    return util.NewBadRequestError(
        util.NewInvalidInputError("Cannot retry an archived run. Unarchive the run first."),
        "Failed to retry run %s because it is archived", runId,
    )
}

krishkishore972:
Fixes #14164

This PR prevents retrying archived runs. `RetryRun` now returns a bad request error when the run is archived and tells users to unarchive the run before retrying.

**Checklist:**

**Tests:**
- [x] `go test -v ./backend/src/apiserver/resource -run TestRetryRun`

### DIFF
--- a/backend/src/apiserver/resource/resource_manager.go
+++ b/backend/src/apiserver/resource/resource_manager.go
@@ -1064,6 +1064,15 @@ func (r *ResourceManager) RetryRun(ctx context.Context, runId string) error {
 	if err != nil {
 		return util.Wrapf(err, "Failed to retry run %s due to error fetching the run", runId)
 	}
+	if run.StorageState.ToV2() == model.StorageStateArchived {
+		return util.NewBadRequestError(
+			util.NewInvalidInputError(
+				"Cannot retry an archived run. Unarchive the run first.",
+			),
+			"Failed to retry run %s because it is archived",
+			runId,
+		)
+	}
 	// TODO(gkcalat): consider using run.Namespace after migration logic will be available.
 	namespace, err := r.getNamespaceFromRunId(runId)
 	if err != nil {


dobbydobap:
Heads up on the status code: `NewBadRequestError` returns `codes.Aborted`, not `InvalidArgument`, so this ends up as 409 Conflict rather than the 400 the issue asked for. `NewFailedPreconditionError` is the one that maps to 400, and it's what `UnarchiveRun` uses for the same kind of refusal.

Also worth adding a test, since the guard is easy to regress and CONTRIBUTING asks for one on fix PRs.

I have #14164 assigned and pushed a branch with both of those before seeing this, so there's some overlap. Not fussed which one lands.

jeffspahr:
There is one response-contract issue to address if this change is reopened or superseded:

`NewBadRequestError` produces gRPC `codes.Aborted`, which grpc-gateway maps to HTTP 409, not the HTTP 400 described by this PR. The external message is also only `Failed to retry run <id> because it is archived`; the corrective instruction to unarchive first is nested in the internal cause rather than included in `ExternalMessage()`.

Please use `NewInvalidInputError` directly, or follow the nearby archived-lifecycle precedent with `NewFailedPreconditionError`, and include “Unarchive the run first” in the external message. This change should also add an archived-run regression test that asserts the exact status and message and verifies that no retry mutation occurs.

dobbydobap:
Good catch on the external message, that one caught me too. I had the unarchive instruction in the internal cause only, so `ExternalMessage()` didn't carry it and my test was asserting on `Error()`, which does. Both fixed in #14169.

It now uses `NewFailedPreconditionError` for the 400, puts "Unarchive the run first" in the external message, and the test asserts the exact code and message and checks the run is untouched: storage state, run state, retry generation, retry claim and workflow manifest all compared before and after.

PR #14169, opened 2026-08-25
fix(backend): reject retrying an archived run. Fixes #14164

### LINKED ISSUE #14164: bug(backend): RetryRun should reject archived runs to prevent GC lifecycle conflict

dobbydobap:

`RetryRun` never read `run.StorageState`, so an archived run could be retried. Since the run garbage collector landed in #13532 archived runs are deletion candidates, so the retry starts a workflow the collector is still free to delete, and the deletion timer keeps counting from the original `ArchivedAtInSec` rather than from the retry.

This adds a guard immediately after `GetRun`, so the request fails before any pods are deleted or a workflow is created:

```
Failed to retry run <id> as it is archived
Unarchive the run first to allow it to be retried
```

Rejecting rather than auto-unarchiving, as agreed on #14164. It also matches `UnarchiveRun`, the only other place in the backend that refuses this class of request: it tells you to unarchive the parent experiment rather than doing it for you.

Two details worth calling out:

`NewFailedPreconditionError` gives `codes.FailedPrecondition`, which grpc-gateway maps to 400 Bad Request. `NewBadRequestError` would look like the obvious choice from its name, but it returns `codes.Aborted`, which maps to 409 Conflict.

The check goes through `StorageState.ToV2()`, which is the same classification the collector uses: `archivedStorageStateStrings` in `run_store.go` is documented as the closed set of values `ToV2()` maps to ARCHIVED. So the guard and the GC agree on what counts as archived, legacy `STORAGESTATE_ARCHIVED` and `DISABLED` rows included.

Tests: one for the rejection and its status code, one covering unarchive-then-retry so the remedy the error message suggests is actually exercised. The existing 22 `RetryRun` and `Unarchive` tests still pass.

#14168 covered the same issue and was closed by its author. The review on it asked for `FailedPrecondition`, the unarchive instruction in the external message, and a regression test asserting both plus no retry mutation. All three are here.

**Checklist:**

### DIFF
--- a/backend/src/apiserver/resource/resource_manager.go
+++ b/backend/src/apiserver/resource/resource_manager.go
@@ -1064,6 +1064,13 @@ func (r *ResourceManager) RetryRun(ctx context.Context, runId string) error {
 	if err != nil {
 		return util.Wrapf(err, "Failed to retry run %s due to error fetching the run", runId)
 	}
+	// Archived runs are GC deletion candidates, so a retry would race the collector.
+	if run.StorageState.ToV2() == model.StorageStateArchived {
+		return util.NewFailedPreconditionError(
+			errors.New("Unarchive the run first to allow it to be retried"),
+			"Failed to retry run %s as it is archived", runId,
+		)
+	}
 	// TODO(gkcalat): consider using run.Namespace after migration logic will be available.
 	namespace, err := r.getNamespaceFromRunId(runId)
 	if err != nil {
--- a/backend/src/apiserver/resource/resource_manager.go
+++ b/backend/src/apiserver/resource/resource_manager.go
@@ -1065,10 +1065,7 @@ func (r *ResourceManager) RetryRun(ctx context.Context, runId string) error {
 		return util.Wrapf(err, "Failed to retry run %s due to error fetching the run", runId)
 	}
 	if run.StorageState.ToV2() == model.StorageStateArchived {
-		return util.NewFailedPreconditionError(
-			errors.New("Archived runs are garbage collection candidates, so retrying one would race the collector"),
-			"Failed to retry run %s as it is archived. Unarchive the run first to allow it to be retried", runId,
-		)
+		return storage.NewArchivedRunRetryError(runId)
 	}
 	// TODO(gkcalat): consider using run.Namespace after migration logic will be available.
 	namespace, err := r.getNamespaceFromRunId(runId)
--- a/backend/src/apiserver/resource/resource_manager.go
+++ b/backend/src/apiserver/resource/resource_manager.go
@@ -1064,11 +1064,10 @@ func (r *ResourceManager) RetryRun(ctx context.Context, runId string) error {
 	if err != nil {
 		return util.Wrapf(err, "Failed to retry run %s due to error fetching the run", runId)
 	}
-	// Archived runs are GC deletion candidates, so a retry would race the collector.
 	if run.StorageState.ToV2() == model.StorageStateArchived {
 		return util.NewFailedPreconditionError(
-			errors.New("Unarchive the run first to allow it to be retried"),
-			"Failed to retry run %s as it is archived", runId,
+			errors.New("Archived runs are garbage collection candidates, so retrying one would race the collector"),
+			"Failed to retry run %s as it is archived. Unarchive the run first to allow it to be retried", runId,
 		)
 	}
 	// TODO(gkcalat): consider using run.Namespace after migration logic will be available.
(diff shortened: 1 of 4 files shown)


juliusvonkohout:
@jeffspahr

backend/src/apiserver/resource/resource_manager.go
```diff
@@ -1064,6 +1064,12 @@ func (r *ResourceManager) RetryRun(ctx context.Context, runId string) error {
 	if err != nil {
 		return util.Wrapf(err, "Failed to retry run %s due to error fetching the run", runId)
 	}
+	if run.StorageState.ToV2() == model.StorageStateArchived {
```
jeffspahr, line 1067: **[P2] Recheck archival state inside the locked retry claim**

This validates only the earlier `GetRun` snapshot. `ArchiveExpiredRuns` can commit `ARCHIVED` after this read but before `ClaimRunForRetry` acquires its row lock; that claim currently reads only runtime/retry fields, so it accepts the newly archived terminal row, launches the retry, and leaves `StorageState` and `ArchivedAtInSec` unchanged. A deterministic interleaving that archives immediately before delegating the claim returns success and ends with a `RUNNING` + `ARCHIVED` row. The current `FinishedAtInSec > 0` predicates prevent deletion during execution, but the run remains archived and becomes deletable after its new finish time ages past retention. Please include `StorageState` in the claim's locked read, return the same `FailedPrecondition` before mutation, and cover the interleaving with a regression test.

jeffspahr (changes requested):
One concurrency issue remains in the archived-run rejection path. The submitted focused tests and the full affected resource package otherwise pass.

dobbydobap:
The pre-check only validated the GetRun snapshot, and the claim's locked read never looked at StorageState, so an archive committing in between would still get claimed.

Pushed a fix. StorageState is now part of the locked SELECT in `ClaimRunForRetry` and the claim returns the same `FailedPrecondition` before any mutation. It scans into `sql.NullString` since legacy rows can still have a NULL StorageState, which is what `ArchiveExpiredRuns` already assumes. I also pulled the error into one constructor that both the pre-check and the recheck call, so the two can't drift.

Two tests: a store-level one asserting the status, the exact external message, and that State, FinishedAtInSec, RetryGeneration and RetryClaimedAtInSec are untouched; and one that reproduces the interleaving by archiving the row immediately before the claim delegates.

jeffspahr (approved):
Reviewed at 87d29a33e. No actionable findings.

The locked StorageState recheck closes the earlier archive/claim race: if archival wins, the retry rejects before mutation; if the retry claim wins, FinishedAtInSec is cleared and the GC archival update no longer matches. FailedPrecondition and the external unarchive remedy are preserved across both paths.

Verification passed for the complete backend/src/apiserver/storage and backend/src/apiserver/resource package test suites, go vet for both packages, gofmt, and git diff --check.

PR #14177, opened 2026-08-26
feat(backend): add kubeflow-pipelines tag to S3 user agent

### LINKED ISSUE #13941: [feature] Append a kubeflow-pipelines tag to the API server S3 user agent

jahnavigajjala-3:

Fixes #13941.

This change appends a `kubeflow-pipelines/<TAG_NAME>` identifier to the API server's S3 `User-Agent` using the AWS SDK for Go v2's `AddUserAgentKeyValue` middleware.

### Implementation

- Read the existing API server `TAG_NAME` configuration.
- Fall back to `unknown` when `TAG_NAME` is unset or empty.
- Use AWS SDK v2 middleware to append:
  `kubeflow-pipelines/<TAG_NAME>`
- Preserve the default `aws-sdk-go-v2/...` User-Agent information.
- Scope the change to the API server S3 client only.

### Tests

Added a regression test that:
- Sends an actual S3 `HeadBucket` request to an `httptest.Server`.
- Verifies the standard `aws-sdk-go-v2/` User-Agent is preserved.
- Verifies `kubeflow-pipelines/2.4.0` is added when `TAG_NAME` is configured.
- Verifies `kubeflow-pipelines/unknown` is added when `TAG_NAME` is unset.

Tests run successfully:

```text
go test ./backend/src/apiserver/client_manager -run TestLoadAWSConfig_UserAgentHeader -v
go test ./backend/src/apiserver/client_manager/... -v
```

There is already a draft PR, #13942, associated with issue #13941. This PR is being opened to provide the completed implementation for the issue after the previous work was left as a draft.

### DIFF
--- a/backend/src/apiserver/client_manager/client_manager.go
+++ b/backend/src/apiserver/client_manager/client_manager.go
@@ -24,11 +24,13 @@ import (
 	"time"
 
 	awsv2 "github.com/aws/aws-sdk-go-v2/aws"
+	awsMiddleware "github.com/aws/aws-sdk-go-v2/aws/middleware"
 	awsv2cfg "github.com/aws/aws-sdk-go-v2/config"
 	awsv2creds "github.com/aws/aws-sdk-go-v2/credentials"
 	"github.com/aws/aws-sdk-go-v2/service/s3"
 	s3types "github.com/aws/aws-sdk-go-v2/service/s3/types"
 	"github.com/aws/smithy-go"
+	smithyMiddleware "github.com/aws/smithy-go/middleware"
 	"github.com/cenkalti/backoff"
 	mysqlStd "github.com/go-sql-driver/mysql"
 	"github.com/golang/glog"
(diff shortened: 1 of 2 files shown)


droctothorpe:
Hi @jahnavigajjala-3, thank you for your PR! The linked issue (#13941) must be labeled `ready` before your contribution can be reviewed. Please check our [CONTRIBUTING.md](https://github.com/kubeflow/pipelines/blob/master/CONTRIBUTING.md) for details.

PR #14194, opened 2026-08-28
fix(frontend): handle PAUSED state in hasFinishedV2 and statusToBgColorV2   Fixes #14193

### LINKED ISSUE #14193: bug(frontend): handle PAUSED state in  hasFinishedV2  and  statusToBgColorV2  to prevent runtime crash
### Description

V2beta1RuntimeState.PAUSED is a valid state in the v2beta1 API, but StatusUtils.ts does not handle it. If the API returned a run in the PAUSED state, hasFinishedV2() would fall through to its default branch and throw Error('Unexpected runtime state!'), taking down the Run Details page.

This is not currently reachable through the UI — there is no PauseRun RPC, and no backend path assigns model.RuntimeStatePaused. Filing it as hardening against enum drift: the backend already treats PAUSED as a valid, storable, active state (RuntimeState.IsValid, ToV2, and the run_store active-state query all include it), so the frontend would break as soon as anything starts emitting it.

### Root cause:
In [frontend/src/lib/StatusUtils.ts](https://github.com/kubeflow/pipelines/blob/master/frontend/src/lib/StatusUtils.ts), the hasFinishedV2() function does not have a case for V2beta1RuntimeState.PAUSED and throws an error in its default branch. The statusToBgColorV2() function is also missing a PAUSED case.

This is inconsistent because:

The API enum [V2beta1RuntimeState](https://github.com/kubeflow/pipelines/blob/master/frontend/src/apisv2beta1/run/models/V2beta1RuntimeState.ts#L39) defines PAUSED as a valid state.[statusProtoMap](https://github.com/kubeflow/pipelines/blob/master/frontend/src/lib/StatusUtils.ts#L54) in the same file maps PAUSED → 'Paused'.[StatusV2.tsx](https://github.com/kubeflow/pipelines/blob/master/frontend/src/pages/StatusV2.tsx#L66) correctly handles PAUSED for icon rendering.

Affected call sites:
* RunDetailsV2.tsx line 102: hasFinishedV2(run.state) — would throw, taking down the page.
* RunDetailsV2.tsx lines 375–376: "Finished at" and "Duration" fields also call hasFinishedV2.

### Materials and Reference

Proposed fix: Add V2beta1RuntimeState.PAUSED to the return false group in hasFinishedV2() (a paused run has not finished), and add a case in statusToBgColorV2() returning statusBgColors.warning.

// StatusUtils.ts — hasFinishedV2()
  ```case V2beta1RuntimeState.PENDING: // Fall through
  case V2beta1RuntimeState.RUNNING: // Fall through
  case V2beta1RuntimeState.CANCELING: // Fall through
+ case V2beta1RuntimeState.PAUSED: // Fall through
  case V2beta1RuntimeState.RUNTIME_STATE_UNSPECIFIED:
    return false;
```

// StatusUtils.ts — statusToBgColorV2()
  ```case V2beta1RuntimeState.CANCELING:   // fall through
  case V2beta1RuntimeState.RUNNING:
    return statusBgColors.running;
+ case V2beta1RuntimeState.PAUSED:
+   return statusBgColors.warning;
  case V2beta1RuntimeState.SUCCEEDED:
```

---

Aman5613:
Fixes #14193

Description of your changes:

- hasFinishedV2() had a default branch that threw new Error('Unexpected runtime state!'). Since state is optional on V2beta1Run and the API omits it for unspecified states, any unrecognized or missing state reaching this function would crash the Run Details page.

Changes made:

- hasFinishedV2() default branch — replaced throw new Error(...) with return false (keeps existing logger.warn). This matches the safe-fallback pattern used by all other status functions in the file and prevents unrecognized or undefined states from crashing the page.

- statusToBgColorV2() PAUSED case — returns statusBgColors.notStarted instead of statusBgColors.warning, matching StatusV2.tsx which already renders PAUSED with the same muted/pending icon as PENDING.

Tests added:

- hasFinishedV2(undefined)  does not throw
- statusToBgColorV2(PAUSED)  returns notStarted

**Checklist:**

### DIFF
--- a/frontend/src/lib/StatusUtils.test.tsx
+++ b/frontend/src/lib/StatusUtils.test.tsx
@@ -243,11 +243,15 @@ describe('StatusUtils', () => {
         expect(hasFinishedV2(state)).toBe(false);
       });
     });
+
+    it('does not throw for undefined state', () => {
+      expect(() => hasFinishedV2(undefined)).not.toThrow();
+    });
   });
 
   describe('statusToBgColorV2', () => {
-    it("returns 'warning' color for PAUSED state", () => {
-      expect(statusToBgColorV2(V2beta1RuntimeState.PAUSED)).toEqual(statusBgColors.warning);
+    it("returns 'notStarted' color for PAUSED state", () => {
+      expect(statusToBgColorV2(V2beta1RuntimeState.PAUSED)).toEqual(statusBgColors.notStarted);
     });
 
     it("returns 'running' color for RUNNING state", () => {
--- a/frontend/src/lib/StatusUtils.test.tsx
+++ b/frontend/src/lib/StatusUtils.test.tsx
@@ -17,12 +17,15 @@
 import {
   NodePhase,
   hasFinished,
+  hasFinishedV2,
   statusBgColors,
   statusToBgColor,
+  statusToBgColorV2,
   checkIfTerminated,
   parseNodePhase,
 } from './StatusUtils';
 import { NodeStatus, S3Artifact, Artifact } from 'third_party/argo-ui/argo_template';
+import { V2beta1RuntimeState } from 'src/apisv2beta1/run';
 
 describe('StatusUtils', () => {
   describe('hasFinished', () => {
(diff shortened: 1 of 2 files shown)


Priyanshu-u07 (changes requested):
Thanks for filing the issue and opening the PR, @Aman5613! I looked into the code and I think the fix needs to move down a level. Changes requested:

1.The thing that actually crashes the page is the `default` branch of `hasFinishedV2`. Adding `PAUSED` stops that one value from reaching it, but any other unrecognized state still takes down the Run Details page. 

Could you change the `default` to fall back instead of throwing? It's the only throw in the file. The other three status functions all fall back to a safe default, including statusToBgColorV2 right below it.
Keeping the existing logger.warn preserves the diagnostic:

```ts
    default:
      logger.warn('Unknown state:', state);
      return false;
```

Could you also add `expect(() => hasFinishedV2(undefined)).not.toThrow();`? state is optional on `V2beta1Run` and the API omits it for unspecified states, so `undefined` can reach this function.

2.The color in `statusToBgColorV2`. `StatusV2.tsx` already renders PAUSED exactly like PENDING, with the same pending icon and the same muted color, and PENDING maps to `statusBgColors.notStarted`. So `warning` would make the background disagree with the icon sitting right next to it, and it isn't used anywhere else in the frontend. 

Could you either drop that hunk or return `notStarted` instead? This one is more of a consistency preference than a bug, so if you think `warning` is the better choice, please tell me why.

3.Minor: PR description overstates both bugs. PAUSED isn't reachable yet, and `statusToBgColorV2` has no callers outside the new tests. Please update PR description.

Aman5613:
Hey @Priyanshu-u07, I've addressed all three points from your review:

- hasFinishedV2 default now returns false instead of throwing, with logger.warn preserved
- Added expect(() => hasFinishedV2(undefined)).not.toThrow() test
- PAUSED in statusToBgColorV2  now returns notStarted to match StatusV2.tsx
- PR description updated to reflect actual scope

Please take a look when you get a chance. Thanks!

Priyanshu-u07 (approved):
lgtm @Aman5613!

Priyanshu-u07:
@jeffspahr when you get a chance. Thanks!

jeffspahr (approved):
Reviewed a70606be45a222ea13a125cc67b5c241b90c0643. No actionable findings.

This is useful defensive hardening: treating PAUSED as unfinished matches backend semantics, and returning false for missing or unknown states avoids crashing Run Details. The explicit PAUSED color case preserves the existing fallback color. Since the backend currently has no path that emits PAUSED, this primarily guards against future state handling changes.

Verification: all 17 added test cases plus 3 additional assertions passed against the PR source in an isolated Node assertion harness with logging stubbed. Full Vitest and build checks were not run.

PR #14224, opened 2026-08-31
fix(frontend): supply DataTransfer items for react-dropzone 20

jeffspahr:
Supersedes #14222.

Carries the same `react-dropzone` bump Dependabot proposed (`^14.4.1` -> `^20.1.1`, pulling `file-selector` `^2.1.0` -> `^5.0.0`) plus the test change it needs to go green. #14222 failed `frontend-tests` on two files and could not fix itself.

## Why the bump fails on its own

`react-dropzone` 20 reads a drop's files through `file-selector` 5, which for a `drop` event takes them from `DataTransfer.items` and never consults `DataTransfer.files`:

```ts
async function getDataTransferFiles(dt: DataTransfer, type: string) {
  const items = fromList<DataTransferItem>(dt.items).filter(item => item.kind === "file");
  ...
}

function fromList<T>(items: DataTransferItemList | FileList | null): T[] {
  if (items === null) {   // guards null, not undefined
    return [];
  }
  return Array.from(items as unknown as ArrayLike<T>);
}
```

Our synthetic drop events supplied only `{ files, types }`, so `dt.items` was `undefined`, the `=== null` guard did not catch it, and `Array.from(undefined)` threw:

```
TypeError: undefined is not iterable (cannot read property Symbol(Symbol.iterator))
    at fromList             (file-selector/dist/index.js:111:15)
    at getDataTransferFiles (file-selector/dist/index.js:98:16)
    at fromEvent            (file-selector/dist/index.js:70:16)
```

Real browsers populate both `items` and `files` on a drop, so this only affects synthetic events in tests. `DropzoneArea` itself needs no change, and no production code is touched.

## The change

Adds `dataTransferWithFiles` and uses it at the three drop sites (two in `DropzoneArea.test.tsx`, one in `NewRun.test.tsx`). Each item carries the minimal shape `file-selector` needs: `kind: 'file'` so it survives the filter, and `getAsFile()` to read the `File`. Omitting `webkitGetAsEntry` and `getAsFileSystemHandle` deliberately keeps it on the plain-file path rather than directory traversal or the File System Access API.

The helper lives in `src/testUtils/` rather than `src/TestUtils.ts`: the latter imports the generated tailwind stylesheet, so putting it there would pull app-level scaffolding into an otherwise dependency-free atom test.

## Verification

- `vitest run` full suite: **149 files, 2054 tests, all passing**
- `DropzoneArea.test.tsx` 7/7 and `NewRun.test.tsx` 81/81 in isolation
- `tsc --noEmit` clean
- `eslint` 0 errors on the new file; `prettier --check` clean

## Note for reviewers

The bumped transitive dependencies (`file-selector` 5, and its own deps) now declare `"node": ">= 22"`. Worth confirming that matches the Node version the frontend builds against, though nothing in CI flagged it.

### DIFF
--- a/frontend/package-lock.json
+++ b/frontend/package-lock.json
@@ -39,7 +39,7 @@
         "react": "^19",
         "react-ace": "^15.0.0",
         "react-dom": "^19",
-        "react-dropzone": "^14.4.1",
+        "react-dropzone": "^20.1.1",
         "react-router": "^5.3.4",
         "react-router-dom": "^5.3.4",
         "react-textarea-autosize": "^8.5.9",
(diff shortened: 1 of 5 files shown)


manaswinidas (approved):
/lgtm
/approve

PR #14230, opened 2026-08-31
fix(backend): fail run creation on plugin runtime env injection failure

### LINKED ISSUE #14179: feat(backend): Modify MLflow plugin behavior to fail on InjectPluginRuntimeEnv failure
# Chore: Fix error handling for InjectPluginRuntimeEnv failures in OnBeforeRunCreation

## Feature Area

The `OnBeforeRunCreation` method in `dispatcher.go` currently swallows errors from `InjectPluginRuntimeEnv`, logging them as warnings instead of propagating them to cause run creation to fail. This creates a critical inconsistency in the plugin lifecycle that can leave orphaned resources.

**Location:** [`backend/src/apiserver/plugins/dispatcher.go:159-161`](backend/src/apiserver/plugins/dispatcher.go#L159-L161)

```go
if len(pluginRuntimeEnv) != 0 {
    if err := InjectPluginRuntimeEnv(executionSpec, pluginRuntimeEnv); err != nil {
        glog.Warningf("Failed to inject %s runtime env for run %q: %v", handler.Name(), run.RunID, err)
    }
}
```

### Current Behavior

1. `ResourceManager.CreateRun` calls `pluginDispatcher.OnBeforeRunCreation` ([`resource_manager.go:716`](backend/src/apiserver/resource/resource_manager.go#L716))
2. `OnBeforeRunCreation` iterates over all plugin handlers in a loop ([`dispatcher.go:124`](backend/src/apiserver/plugins/dispatcher.go#L124))
3. For each handler:
   - `handler.OnBeforeRunCreation` is called, which may create plugin resources (e.g., MLflow parent run)
   - If the handler returns `pluginRuntimeEnv`, `InjectPluginRuntimeEnv` is called to inject environment variables into the execution spec
   - **If `InjectPluginRuntimeEnv` fails, the error is only logged; the loop continues**
4. `OnBeforeRunCreation` returns `nil` (success) even though env var injection failed
5. `ResourceManager.CreateRun` proceeds to create the workflow ([`resource_manager.go:735`](backend/src/apiserver/resource/resource_manager.go#L735))

## What is the use case or pain point?

This creates several problems:

1. **Orphaned plugin resources:** If an earlier plugin handler in the loop successfully creates resources (e.g., an MLflow parent run), but a later plugin's `InjectPluginRuntimeEnv` fails, those resources are left orphaned because:
   - The KFP run may complete successfully despite missing environment variables
   - Or the run may fail, but the plugin resources aren't cleaned up because `OnRunEnd` doesn't know about the incomplete setup

2. **Silent failures:** Run creation appears to succeed, but the execution spec is incomplete, leading to runtime failures that are difficult to debug

3. **Inconsistent plugin state:** Plugin output indicates success, but the execution spec doesn't contain the required environment variables

## Is there a workaround currently?

No, there is not currently a workaround here

## Expected Behavior

1. When `InjectPluginRuntimeEnv` fails, `OnBeforeRunCreation` should return an error
2. This error should cause `ResourceManager.CreateRun` to fail before the workflow is created
3. The existing defer block in `ResourceManager.CreateRun` ([`resource_manager.go:727-733`](backend/src/apiserver/resource/resource_manager.go#L727-L733)) should trigger `OnRunEnd` to clean up any plugin resources created by handlers that executed before the failure

### 1. Propagate InjectPluginRuntimeEnv errors

**File:** `backend/src/apiserver/plugins/dispatcher.go`

Modify the `OnBeforeRunCreation` method to return an error when `InjectPluginRuntimeEnv` fails:

```go
if len(pluginRuntimeEnv) != 0 {
    if err := InjectPluginRuntimeEnv(executionSpec, pluginRuntimeEnv); err != nil {
        return fmt.Errorf("failed to inject %s runtime env for run %q: %w", handler.Name(), run.RunID, err)
    }
}
```

### 2. Ensure OnRunEnd handles partial creation scenarios

**File:** `backend/src/apiserver/plugins/dispatcher.go`

The `OnRunEnd` method should properly clean up plugin resources even when:
- The run was never persisted to the database
- Only some plugins in the handler loop successfully executed
- Plugin output indicates a parent run was created but the KFP run failed

Review the logic in `OnRunEnd` ([`dispatcher.go:174-226`](backend/src/apiserver/plugins/dispatcher.go#L174-L226)) to ensure it:
- Checks if plugin output exists before attempting cleanup
- Handles cases where `pluginOutput.State` indicates partial success
- Properly closes/terminates any parent runs created during the failed run creation

### 3. Update tests

Add test coverage for:
- `OnBeforeRunCreation` returning an error when `InjectPluginRuntimeEnv` fails
- Multiple plugins in the handler loop, where an earlier plugin creates resources and a later plugin's env var injection fails
- `OnRunEnd` being called via the defer block when run creation fails mid-flight
- Verification that orphaned plugin resources are cleaned up

### Feature Area

/area backend

---

Love this idea? Give it a 👍.

kr1xh:
Fixes #14179

### What was happening

When a plugin's `OnBeforeRunCreation` hook successfully created plugin-side resources but failed while injecting its runtime environment variables, the error from `InjectPluginRuntimeEnv` was only logged and not propagated.

As a result, run creation could continue even though the plugin setup was incomplete.

There was also a cleanup problem: if plugin setup failed before the KFP run was persisted, the existing `OnRunEnd` hook was not invoked, leaving partially-created plugin resources without the normal terminal cleanup path.

### What this changes

- Propagate `InjectPluginRuntimeEnv` failures from `OnBeforeRunCreation`.
- Register the `OnRunEnd` cleanup path before plugin setup begins, so it also runs when run creation fails before persistence.
- Preserve plugin output produced before `OnBeforeRunCreation` fails so `OnRunEnd` can use it for cleanup.
- Add regression tests covering both runtime environment injection failures and cleanup after partial plugin setup.

### Tests

- `go test ./backend/src/apiserver/plugins -run '^TestOnBeforeRunCreation_' -count=1 -timeout=120s`
- `go test ./backend/src/apiserver/resource -run '^TestCreateRun_' -count=1 -timeout=120s`
- `go test ./backend/src/apiserver/plugins ./backend/src/apiserver/resource -count=1 -timeout=180s`

### DIFF
--- a/backend/src/apiserver/plugins/dispatcher.go
+++ b/backend/src/apiserver/plugins/dispatcher.go
@@ -124,12 +124,18 @@ func (d *RunPluginDispatcherImpl) OnBeforeRunCreation(ctx context.Context, run *
 	if err != nil {
 		return fmt.Errorf("failed to retrieve launcher namespace plugin configs for run %q: %v", run.RunID, err)
 	}
+
 	for _, handler := range d.handlers {
-		func(ctx context.Context, run *PendingRun, executionSpec util.ExecutionSpec, handler RunPluginHandler) {
+		err := func(ctx context.Context, run *PendingRun, executionSpec util.ExecutionSpec, handler RunPluginHandler) error {
 			input := pluginInputs[handler.Name()]
 
 			launcherNamespacePluginCfgStr := launcherNamespacePluginCfgs[handler.Name()]
-			resolvedRunPluginCfg, err := handler.ResolveRunPluginConfig(ctx, d.kubeClients.GetClientSet(), launcherNamespacePluginCfgStr, run.Namespace)
+			resolvedRunPluginCfg, err := handler.ResolveRunPluginConfig(
+				ctx,
+				d.kubeClients.GetClientSet(),
+				launcherNamespacePluginCfgStr,
+				run.Namespace,
+			)
 			if err != nil {
 				message := fmt.Sprintf("%s config resolution failed; run creation will continue: ", handler.Name()) + err.Error()
 				glog.Warningf("MLflow OnBeforeRunCreation failed for run %q (%s)", run.RunID, message)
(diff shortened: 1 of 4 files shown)


droctothorpe:
Hi @kr1xh, thank you for contributing to Kubeflow Pipelines.

Issue #14245, opened 2026-09-02
bug(sdk): implicit one-hour retry window can suppress configured retries


herikwebb:
### Summary

`PipelineTask.set_retry(num_retries=N)` implicitly serializes `backoff_max_duration: 3600s`, even when the caller provides no retry-window limit.

KFP maps this to Argo Workflows `retryStrategy.backoff.maxDuration`. Argo treats it as a total wall-clock retry window beginning when the first retry child node is initialized, so scheduling, initialization, execution, and backoff can exhaust the window before `N` retries occur.

This issue proposes agreeing on the desired default and compatibility strategy before implementation.

### Environment

- KFP backend and SDK: current `master` at [`92f1147`](https://github.com/kubeflow/pipelines/commit/92f11478e0e86133dd5eb8e424c4c032c7d941e5)
- Argo Workflows: v4.1.2, as pinned by KFP `master`
- Deployment: reproducible in SDK/IR compilation and the Argo backend compiler

### Steps to reproduce

```python
from kfp import compiler, dsl


@dsl.component
def work():
    raise RuntimeError("fail")


@dsl.pipeline
def pipeline():
    work().set_retry(num_retries=3)


compiler.Compiler().compile(pipeline, "/tmp/pipeline.yaml")
```

The resulting IR contains:

```yaml
retryPolicy:
  backoffDuration: 0s
  backoffFactor: 2.0
  backoffMaxDuration: 3600s
  maxRetryCount: 3
```

The upstream fixture demonstrates the same behavior: [pipeline source](https://github.com/kubeflow/pipelines/blob/92f11478e0e86133dd5eb8e424c4c032c7d941e5/test_data/sdk_compiled_pipelines/valid/pipeline_with_retry.py#L24-L32) and [compiled IR](https://github.com/kubeflow/pipelines/blob/92f11478e0e86133dd5eb8e424c4c032c7d941e5/test_data/sdk_compiled_pipelines/valid/critical/pipeline_with_retry.yaml#L65-L69).

### Actual result

The implicit one-hour window can prevent retries requested through `num_retries`. For example, an initial attempt that spends more than an hour pending or running and then fails receives no retry.

[...]

PR #14248, opened 2026-09-02
fix(backend): apply retryStrategy to importer tasks in Argo compiler

### LINKED ISSUE #14246: bug(backend): set_retry is ignored for importer tasks
### Environment

- KFP v2 backend and SDK: current `master` at [`92f1147`](https://github.com/kubeflow/pipelines/commit/92f11478e0e86133dd5eb8e424c4c032c7d941e5)
- Execution mode: Argo Workflows backend

### Steps to reproduce

Compile and submit this pipeline:

```python
from kfp import compiler, dsl


@dsl.pipeline(name='importer-retry-repro')
def importer_retry_repro():
    task = dsl.importer(
        artifact_uri='s3://example-bucket/existing/data.csv',
        artifact_class=dsl.Dataset,
        reimport=False,
    )
    task.set_retry(
        num_retries=3,
        backoff_duration='10s',
        backoff_factor=2.0,
        backoff_max_duration='60s',
    )


compiler.Compiler().compile(
    pipeline_func=importer_retry_repro,
    package_path='importer-retry.yaml',
)
```

The SDK retains the policy in the pipeline IR:

```yaml
retryPolicy:
  backoffDuration: 10s
  backoffFactor: 2.0
  backoffMaxDuration: 60s
  maxRetryCount: 3
```

After the backend compiles the run into an Argo Workflow, inspect the importer task and `system-importer` template. The task references `system-importer`, but that template has no `retryStrategy`.

### Expected result

`PipelineTask.set_retry()` is documented as setting task retry parameters and does not exclude importer tasks. An importer configured with three retries should receive one initial attempt plus up to three retries using the requested backoff settings.

If importer retry is intentionally unsupported, the SDK should reject it with a clear error rather than retain a policy that the backend drops.

### Actual result

The retry policy survives SDK compilation but is ignored by the backend Argo compiler for importer tasks. A transient failure therefore receives only the independently configured platform behavior, if any; the requested task retry count and backoff are not honored.

The backend importer path accepts the complete `PipelineTaskSpec` but never reads `task.GetRetryPolicy()` or passes retry parameters:

- [`importerTask`](https://github.com/kubeflow/pipelines/blob/92f11478e0e86133dd5eb8e424c4c032c7d941e5/backend/src/v2/compiler/argocompiler/importer.go#L38-L64)
- [`addImporterTemplate`](https://github.com/kubeflow/pipelines/blob/92f11478e0e86133dd5eb8e424c4c032c7d941e5/backend/src/v2/compiler/argocompiler/importer.go#L66-L153)

By contrast, the container path forwards task retry parameters and adds `RetryStrategy` to the launcher template:

- [Container retry parameter forwarding](https://github.com/kubeflow/pipelines/blob/92f11478e0e86133dd5eb8e424c4c032c7d941e5/backend/src/v2/compiler/argocompiler/container.go#L327-L368)
- [Container `RetryStrategy`](https://github.com/kubeflow/pipelines/blob/92f11478e0e86133dd5eb8e424c4c032c7d941e5/backend/src/v2/compiler/argocompiler/container.go#L581-L592)

The local runner also applies the same IR retry policy before dispatching either an importer or container task, so local and cluster execution differ: [local importer retry loop](https://github.com/kubeflow/pipelines/blob/92f11478e0e86133dd5eb8e424c4c032c7d941e5/sdk/python/kfp/local/orchestrator/task_executor.py#L124-L161).

### Replay and idempotency considerations

Importer retries should preserve one logical task result even when a previous attempt committed work but its response was lost.

The current launcher creates a metadata execution before resolving and publishing the artifact:

- [Importer execution and publication](https://github.com/kubeflow/pipelines/blob/92f11478e0e86133dd5eb8e424c4c032c7d941e5/backend/src/v2/component/importer_launcher.go#L117-L165)
- [`reimport` artifact selection](https://github.com/kubeflow/pipelines/blob/92f11478e0e86133dd5eb8e424c4c032c7d941e5/backend/src/v2/component/importer_launcher.go#L168-L185)
- [Workspace download](https://github.com/kubeflow/pipelines/blob/92f11478e0e86133dd5eb8e424c4c032c7d941e5/backend/src/v2/component/importer_launcher.go#L296-L319)

The implementation should account for:

- `reimport=False`: replay must not leave duplicate logical executions or relationships when reusing an artifact.
- `reimport=True`: retry after an ambiguous publish must not accidentally create multiple artifacts for one logical attempt without an explicit decision that this is acceptable.
- `download_to_workspace=True`: repeated or interrupted downloads need deterministic overwrite, cleanup, or atomic replacement behavior.

### Proposed scope and acceptance criteria

- Consume importer `PipelineTaskSpec.retry_policy` in the backend Argo compiler, or explicitly reject unsupported configuration in the SDK.
- Apply the configured limit and backoff to the actual importer launcher pod.
- Preserve per-task isolation when a pipeline contains importers with different policies or no retry policy; do not unconditionally mutate the shared `system-importer` template.
- Cover both normal and workspace importer templates.
- Add compiler tests verifying the generated Argo `retryStrategy`.
- Add coverage for multiple importer tasks with distinct policies to prevent template-policy leakage.
- Verify bounded transient-failure recovery and document or test replay behavior for `reimport=False`, `reimport=True`, and workspace downloads.
- Keep general container-driver, DAG-driver, and task-group retry semantics out of this issue.

### Duplicate search

As of 2026-09-01, searches across open and closed Kubeflow Pipelines issues and pull requests for `importer retry`, `set_retry importer`, and `retryStrategy importer`, plus inspection of open PRs modifying the importer compiler, found no issue or PR directly addressing this behavior.

---

pujitha24:

`PipelineTask.set_retry()` compiles a retry policy into the pipeline IR for `dsl.importer()` tasks the same as for any other task, but the backend Argo compiler's importer path (`importerTask`/`addImporterTemplate` in `backend/src/v2/compiler/argocompiler/importer.go`) never read `PipelineTaskSpec.GetRetryPolicy()`, so the generated `system-importer` Argo template never had a `retryStrategy`. A transient importer failure was silently never retried, unlike an equivalent container task.

This mirrors the existing container-task retry pattern (`container.go`): when an importer task has a retry policy, its DAG task now resolves to a `"retry-"`-prefixed template (`retry-system-importer` / `retry-system-importer-workspace`) carrying the four retry input parameters and a `Template.RetryStrategy` built from them, while importer tasks without a retry policy keep sharing the plain, unaffected `system-importer`/`system-importer-workspace` template. Actual retry values are threaded per-task through the DAG task's arguments, so multiple importer tasks with different retry policies stay isolated even though they share the same parameterized template.

Scope note: this only wires up the Argo compiler's `retryStrategy` for the importer launcher pod. The issue also raises idempotency/replay behavior across retries (for `reimport=True/False` and workspace downloads) — that's a separate, larger concern in the importer launcher itself and is intentionally not addressed in this PR.

Validation:
- `go test ./backend/src/v2/compiler/...` passes, including new unit tests in `importer_test.go` covering: no retry policy (template unaffected), a retry policy (retry-prefixed template with `RetryStrategy`), the workspace variant, and multiple importer tasks with distinct retry policies keeping isolated templates.
- Added an end-to-end compiler golden-file fixture pair (`test_data/sdk_compiled_pipelines/valid/pipeline_with_importer_retry.yaml` and `test_data/compiled-workflows/pipeline_with_importer_retry.yaml`) exercised by the existing ginkgo suite; `go run github.com/onsi/ginkgo/v2/ginkgo ./backend/test/compiler` (from `backend/`) passes 305/309 specs (4 skipped need env vars unrelated to this change).
- Confirmed this is a real regression test: stashing only the `importer.go` change and re-running the focused ginkgo spec fails against the pre-fix code, and passes after restoring the fix.
- `go vet`, `gofmt`, and `golangci-lint run --new-from-rev=HEAD` on the changed package report no new issues.

Fixes #14246

**Checklist:**


dobbydobap:
Copilot's point about the missing input defaults is right, and I think it matters more than it looks.

`container.go:423` wraps its retry inputs in `addParameterDefault(..., "0")`, so all four tolerate a missing DAG argument. The importer template here declares the same four with no `Default`, while `getTaskRetryParametersWithValues` skips `backoff_duration` and `backoff_max_duration` when they're nil. A policy missing either one gives you a template requiring four inputs and a task supplying two.

It's masked today for anything the SDK compiles, since `RetryPolicy.to_proto` always fills all four fields, so only hand-written or API-submitted IR reaches it right now.

The reason I'd still fix it: #14245 proposes dropping exactly that SDK-side `backoff_max_duration` default. If that lands, ordinary `set_retry(num_retries=N)` pipelines start emitting IR without the field and this becomes reachable through the normal path. Matching `container.go`'s `addParameterDefault` now would make this PR independent of how #14245 is decided.

Rest reads well. The retry-prefixed template keeps non-retry importers untouched, and the per-task argument threading is the same shape `container.go` uses.

PR #14249, opened 2026-09-02
fix: harden dependency update validation and auto-merge safety

jeffspahr:
## Summary

- restore an explicit 100 MiB grpc-web response limit for the MLMD client, matching the deployment's server send limit
- run the hooks pinned in `.pre-commit-config.yaml` over each base-to-head change, with bounded representative coverage for pre-commit and golangci configuration updates
- trigger the frontend E2E/image-build workflow for metadata-envoy changes and pin that trigger chain with CI-script tests
- complete the mock backend's Express 5 route and type migration, and construct the standalone app with its own installed dependency graph in frontend CI
- fix Kubernetes platform proto import resolution and generate/typecheck both pipeline and platform specs in frontend CI
- start every new Dependabot PR with `do-not-merge/hold`, while explicitly retaining the standard dependency and ecosystem labels

## Why

Dependency updates exposed validation gaps: [#14208](https://github.com/kubeflow/pipelines/pull/14208) adopted grpc-web's 4 MiB default receive limit, [#14216](https://github.com/kubeflow/pipelines/pull/14216) updated pre-commit hooks without exercising them through the separately pinned golangci-lint workflow, [#14214](https://github.com/kubeflow/pipelines/pull/14214) changed metadata-envoy without triggering an image build, and the Express 5 update in [#14257](https://github.com/kubeflow/pipelines/pull/14257) passed while the standalone mock server crashed during route registration.

The ts-proto update in [#14290](https://github.com/kubeflow/pipelines/pull/14290) also exposed an untested generation path: the Kubernetes platform script omitted the include directory for `pipeline_spec.proto`. Both generation commands now support temporary output directories, and frontend CI generates and compiles their output with the installed generator and TypeScript versions. The smoke check does not rewrite committed generated clients.

The existing mock route tests constructed the middleware with the frontend server's compatibility-pinned router, so they did not exercise the package that `npm run mock:api` actually loads. The new app-construction test uses `frontend/mock-backend`'s nested Express tree and preserves representative route behavior across both router graphs.

The Dependabot Tide path has a separate auto-merge safety gap. Until Dependabot PRs can reliably earn `ci-passed` without `ok-to-test` ([#14111](https://github.com/kubeflow/pipelines/pull/14111)) and the external Tide configuration requires that label, repository-owned Dependabot labels give each newly raised PR a Tide-blocking starting state without relying on a repository `pull_request: opened` workflow. GitHub's [`labels` option](https://docs.github.com/en/code-security/reference/supply-chain-security/dependabot-options-reference#labels--) replaces Dependabot's defaults and applies to version and security updates, so every ecosystem explicitly retains `dependencies` and its ecosystem label alongside `do-not-merge/hold`.

This is intentionally a manual stopgap, not an automatic or SHA-bound gate. The hold is applied when a PR is created and is PR-scoped: this configuration is not guaranteed to retrofit existing or reopened PRs, and once the hold is removed it remains absent across Dependabot rebases and other head changes. Audit those PRs separately, remove the hold only immediately before merge after all expected presubmits pass for the current head, and reapply it immediately if that head changes. The label blocks Tide auto-merging; it does not prevent a privileged direct merge.

## Verification

- pinned Node 24.14.0/npm 11.17.0 frontend formatting, lint, root and mock-backend typechecks, and React peer validation
- UI coverage: 2,071 tests passed
- server coverage: 436 tests passed, 2 todo
- standalone plus compatibility-router mock suites: 49 tests passed, 2 todo
- complete CI-script suite: 229 tests passed
- focused Dependabot configuration suite: 9 tests passed
- JUnit summary action suite: 7 tests passed
- `.github/dependabot.yml` parses to six update entries with the expected ordered label lists and no `target-branch`
- YAPF 0.43.0 under Python 3.13, isort, pycln, docformatter, and `git diff --check`
- pipeline and Kubernetes platform generation smoke passes with ts-proto 2.12.1 and TypeScript 6.0.3; Bash syntax and workflow YAML validation pass
- the previous hosted CI run passed all 204 checks; the new revision requires a fresh hosted run


.pre-commit-config.yaml
```diff
@@ -61,21 +61,20 @@ repos:
...
+          if git diff --quiet "${base}" -- "*.go"; then
+            golangci-lint config verify;
+          else
+            golangci-lint run --new-from-rev "${base}" --fix;
+          fi'
         types: [go]
         language: golang
         require_serial: true
         pass_filenames: false
+      - id: golangci-lint-fmt
```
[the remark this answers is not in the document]
jeffspahr: Fixed in `3896b3202`. `golangci-lint-fmt` now overrides the upstream default with `pass_filenames: true`, so pre-commit forwards only its selected Go paths in both range and representative smoke runs. Added a hook-specific regression assertion; a two-file probe confirmed the selected file is formatted while the unselected file is untouched. The bounded hook smoke and exact PR range pass.

.pre-commit-config.yaml
```diff
@@ -61,21 +61,21 @@ repos:
...
-        language: golang
-        require_serial: true
-        pass_filenames: false
-      - id: golangci-lint
-        name: golangci-lint fmt
-        description: Formatter for Go.
-        entry: golangci-lint fmt
+        entry: >-
+          bash -c 'base="${PRE_COMMIT_FROM_REF:-HEAD}";
+          if git diff --quiet "${base}" -- "*.go"; then
```
jeffspahr, line 66: [P2] Exercise `golangci-lint run` during configuration smoke

For a configuration-only golangci update, the base-to-head diff contains no Go files, so this branch executes only `golangci-lint config verify`. Because filenames are disabled, the smoke test's representative Go file does not reach `golangci-lint run`. This leaves package loading and analyzers untested and could allow a broken hook version to auto-merge. Have the no-Go smoke path run a bounded, known-clean package and add coverage proving the `run` subcommand is invoked.

jeffspahr (commented):
Re-reviewed head `3896b3202`. The formatter-scoping issue is fixed and I found no other actionable findings, but one P2 remains and I would hold approval for it.

.pre-commit-config.yaml
jeffspahr: Fixed in c329489e7 and 3da778b14. For a no-Go diff, the hook now runs golangci-lint config verify and then the non-mutating bounded command golangci-lint run ./backend/src/common/plugins. The existing representative file remains stable and still activates the Go hooks. The behavioral regression executes the actual folded shell entry and proves that both subcommands run, a failed config verification short-circuits lint, and a failed lint run propagates. Verified with the focused 6-test suite, all 228 CI-script tests, the exact PR-range pre-commit run, the full configuration-smoke file set, actionlint, and the pinned linter against the package.

PR #14255, opened 2026-09-02
fix(sdk): honor enable_caching/cache_key when submitting by pipeline_…

### LINKED ISSUE #14253: bug(sdk): enable_caching and cache_key are ignored when submitting a run by pipeline_id/version_id
### Environment

* KFP version:
Not needed to reproduce this — it's a client-side bug you can see just by reading the code or printing the request object, no cluster required.

* KFP SDK version:
kfp==2.15.2 (current version on master). This has been broken since at least 2.0.0b14.

* All dependencies version:
kfp==2.15.2

### Steps to reproduce

`run_pipeline()` and `create_recurring_run()` both let you pass `enable_caching` to turn caching on/off for a run, and the docstring says it overrides whatever was set at compile time.

That only actually happens if you submit the pipeline as a local file (`pipeline_package_path`). If you instead point at a pipeline that's already uploaded to the server using `pipeline_id` + `version_id` — which is the normal way to trigger a pipeline once it's registered — `enable_caching` and `cache_key` are just thrown away. No error, no warning, nothing in the logs. The request goes out as if you never passed them.

You can see this directly in `_create_job_config()` in `sdk/python/kfp/client/client.py` (around line 978): it only calls the function that applies the caching override inside the `if pipeline_package_path:` block. The `pipeline_id`/`version_id` branch right below it never touches `enable_caching` at all.

Example that shows the problem:

    from kfp.client import Client

    client = Client(host="http://localhost:8888")

    run = client.run_pipeline(
        experiment_id="<experiment-id>",
        job_name="caching-repro",
        pipeline_id="<existing-pipeline-id>",
        version_id="<existing-version-id>",
        enable_caching=False,   # I want a fresh run, no cache
    )

If you inspect the request this sends, `enable_caching=False` never makes it into the payload — the run goes out with whatever caching was set when the pipeline was compiled, not what you just asked for.

Same problem in `create_recurring_run()`, since it uses the same helper function.

### Expected result

Either:
1. `enable_caching`/`cache_key` should actually work when you reference a pipeline by `pipeline_id`/`version_id`, not just when uploading a fresh file — most likely by fetching the pipeline version's spec, applying the override to it, and sending that.
2. Or, if that combination just isn't supported, the SDK should tell you clearly (raise an error) instead of quietly ignoring what you asked for.

Right now it does neither — it just silently does the wrong thing, which is the actual problem. Someone could think they've disabled caching for a run and have no idea it's still cached.

### Materials and Reference

* `sdk/python/kfp/client/client.py:688` — `run_pipeline()`
* `sdk/python/kfp/client/client.py:806` — `create_recurring_run()`
* `sdk/python/kfp/client/client.py:925` — `_create_job_config()`, where the bug actually is
* `sdk/python/kfp/client/client.py:1790` — `_override_caching_options()`, the function that never gets called for this case
* No existing test covers `enable_caching` combined with `pipeline_id`/`version_id`, which is probably why this slipped through.
* Related but not the same issue: #9196 shows this exact code path used to crash outright with `pipeline_id`/`version_id`. That crash got patched, but the fix just avoided the crash — it didn't restore the caching override, so this bug is what's left over from that patch.

---

aman21-droid:
Fixes #14253.

`_create_job_config()` only applied a submission-time `enable_caching`/`cache_key` override when the pipeline was submitted as a local package (`pipeline_package_path`). When referencing an already-uploaded pipeline by `pipeline_id`/`version_id` instead, the override was silently dropped: `enable_caching` and `cache_key` were passed to `run_pipeline()`/`create_recurring_run()` but never reached the request.

This fetches the pipeline version's spec via `get_pipeline_version()` when both `pipeline_id`/`version_id` and `enable_caching` are given, applies the existing `_override_caching_options()` to it, and submits the mutated spec alongside the existing `pipeline_version_reference`. When no caching override is requested, no extra fetch happens, so the common case stays a cheap reference-only request.

Added regression tests covering:
- `enable_caching` actually reaching the submitted spec when using `pipeline_id`/`version_id`, without losing the pipeline version reference.
- No extra API call when no caching override is requested.

Verified:
- `pytest sdk/python/kfp/client/client_test.py` — 53 passed
- `yapf --style .style.yapf -d` — clean
- `isort --settings-path .isort.cfg --check-only` — clean

**Checklist:**

### DIFF
--- a/sdk/python/kfp/client/client.py
+++ b/sdk/python/kfp/client/client.py
@@ -75,6 +75,33 @@ class _PipelineDoc:
             }
 
 
+def _pipeline_doc_from_dict(spec_dict: dict) -> _PipelineDoc:
+    """Builds a _PipelineDoc from a pipeline spec dict as returned by the
+
+    API server (e.g. via ``Client.get_pipeline_version()``).
+
+    The dict is either a flat PipelineSpec (the common case), or a
+    ``{'pipeline_spec': ..., 'platform_spec': ...}`` wrapper -- the shape
+    the API server returns when the pipeline was compiled with
+    platform-specific configuration (e.g. Kubernetes-specific config from
+    the ``kfp-kubernetes`` extension). This is the inverse of
+    ``_PipelineDoc.to_dict()``.
+    """
+    if 'pipeline_spec' in spec_dict:
+        raw_pipeline_spec = spec_dict['pipeline_spec']
+        raw_platform_spec = spec_dict.get('platform_spec', {})
+    else:
+        raw_pipeline_spec = spec_dict
+        raw_platform_spec = {}
+
+    pipeline_spec = pipeline_spec_pb2.PipelineSpec()
+    json_format.ParseDict(raw_pipeline_spec, pipeline_spec)
+    platform_spec = pipeline_spec_pb2.PlatformSpec()
+    json_format.ParseDict(raw_platform_spec, platform_spec)
+    return _PipelineDoc(
+        pipeline_spec=pipeline_spec, platform_spec=platform_spec)
+
+
 @dataclasses.dataclass
 class _JobConfig:
     pipeline_spec: dict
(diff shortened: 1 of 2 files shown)


droctothorpe:
Hi @aman21-droid, thank you for contributing to Kubeflow Pipelines.

PR #14262, opened 2026-09-03
fix(backend): reject cross-namespace workflow reports

herikwebb:

**Why:** Workflow reports previously selected their target using request-provided labels and owner references without proving the complete stored and live Kubernetes identity. A forged or stale report could therefore act on a run belonging to another namespace, delete a same-name replacement, overwrite a newly recreated run that reused an ID, or attach stale task records to that replacement.

**What:** This change accepts a workflow report only when it matches the persisted run or recurring-run owner and the live Kubernetes Workflow. Valid reports continue normally; mismatched, stale, concurrently superseded, or previous-incarnation reports fail before any run, task, or Workflow data is changed.

Technical details:

- Resolve and validate run and recurring-run namespaces before task persistence, database mutation, or Workflow deletion.
- Require the live Workflow UID, run label, namespace, and recurring-run owner to match the persisted identity.
- Treat live Workflow state and labels as authoritative instead of trusting the report body.
- Delete Workflows with UID and resource-version preconditions to protect same-name replacements.
- Fence every existing-run report write against the locked persisted workflow manifest and retry generation.
- For V2 runs before their first report, also fence on the UID-bearing pipeline-runtime manifest so a stale report cannot ABA-match a deleted and recreated run ID.
- Lock the parent run and verify its namespace, runtime manifests, and retry generation in the same transaction that reads and upserts tasks, preventing stale tasks from attaching to a replacement run.
- Reload authoritative state after a concurrent change and refresh the bounded identity cache after every successful report without allowing delayed writers to replace a newer manifest entry.
- Persist terminal state before cleanup and safely reconcile abandoned retry claims.
- Backfill legacy empty Workflow names and namespaces only from verified live or stored identity.
- Revalidate identity after concurrent insert conflicts and count permanently rejected reports by reason.
- Document the additional Kubernetes identity reads and operational cleanup behavior.

Validation:

- `GOTOOLCHAIN=go1.26.6 go test ./backend/src/apiserver/... -count=1`
- `go list ./backend/... | rg -v 'backend/test/(v2/api|integration|v2/integration|initialization|v2/initialization|compiler|end2end)' | xargs go test -p 1 -count=1`
- `go test ./backend/src/apiserver/storage ./backend/src/apiserver/resource -run 'TestCreateOrUpdateTasksIfRunUnchanged|TestCreateOrUpdateTasksForRun_RejectsTasksAfterRunIDRecreation|TestUpdateRunIfRuntimeManifestsUnchanged' -count=20`
- `go test -race ./backend/src/apiserver/storage ./backend/src/apiserver/resource -run 'TestCreateOrUpdateTasksIfRunUnchanged|TestCreateOrUpdateTasksForRun_RejectsTasksAfterRunIDRecreation|TestUpdateRunIfRuntimeManifestsUnchanged' -count=10`
- `GOTOOLCHAIN=go1.26.6 golangci-lint run --new --new-from-merge-base=upstream/master` with golangci-lint v2.10.1 (`0 issues`)
- `GOTOOLCHAIN=go1.26.6 go vet ./backend/src/apiserver/...`
- `gofmt` and `git diff --check`

The affected behavior and fix were already public in [herikwebb/pipelines#122](https://github.com/herikwebb/pipelines/pull/122).

**Checklist:**

- [x] The title for your pull request follows the repository title convention.

### DIFF
--- a/backend/src/apiserver/resource/resource_manager.go
+++ b/backend/src/apiserver/resource/resource_manager.go
@@ -15,7 +15,9 @@
 package resource
 
 import (
+	containerlist "container/list"
 	"context"
+	"crypto/sha256"
 	"encoding/json"
 	"fmt"
 	"io"
@@ -23,6 +25,7 @@ import (
 	"reflect"
 	"strconv"
 	"strings"
+	"sync"
 	"time"
 
 	"github.com/cenkalti/backoff"
@@ -54,6 +57,13 @@ import (
 	"k8s.io/client-go/util/retry"
 )
 
+const (
+	workflowReportRejectionIdentityMismatch    = "identity_mismatch"
+	workflowReportRejectionNamespaceMismatch   = "namespace_mismatch"
+	workflowReportRejectionOwnershipUnresolved = "ownership_unresolved"
+	storedWorkflowIdentityCacheCapacity        = 10_000
+)
+
 // Metric variables. Please prefix the metric names with resource_manager_.
 var (
 	extraLabels = []string{
@@ -69,6 +79,14 @@ var (
 		Name: "resource_manager_workflow_gc",
 		Help: "The number of garbage-collected workflows",
 	})
+
+	// Count reports rejected before they can mutate a run or delete a Workflow.
+	// The reason label is restricted to constants so the metric has
+	// bounded cardinality and can be used for operator alerts.
+	workflowReportRejectedCounter = promauto.NewCounterVec(prometheus.CounterOpts{
+		Name: "resource_manager_workflow_reports_rejected_total",
+		Help: "The number of workflow reports rejected before persistence or garbage collection",
+	}, []string{"reason"})
 
 	// Count the successful workflow runs
 	workflowSuccessCounter = promauto.NewGaugeVec(prometheus.GaugeOpts{
@@ -150,6 +168,130 @@ type ResourceManager struct {
 	authenticators            []kfpauth.Authenticator
 	options                   *ResourceManagerOptions
 	pluginDispatcher          apiserverPlugins.RunPluginDispatcher
+	storedWorkflowIdentities  storedWorkflowIdentityCache
+}
+
+type storedWorkflowIdentity struct {
+	name            string
+	namespace       string
+	uid             types.UID
+	retryGeneration int64
+	manifestDigest  [sha256.Size]byte
+}
+
+type cachedStoredWorkflowIdentity struct {
+	identity storedWorkflowIdentity
+	element  *containerlist.Element
+}
+
+type storedWorkflowIdentityCache struct {
+	mu      sync.Mutex
+	entries map[string]cachedStoredWorkflowIdentity
+	recency *containerlist.List
+}
+
+func (c *storedWorkflowIdentityCache) load(runID string) (storedWorkflowIdentity, bool) {
+	c.mu.Lock()
+	defer c.mu.Unlock()
+
+	entry, found := c.entries[runID]
+	if !found {
+		return storedWorkflowIdentity{}, false
+	}
+	c.recency.MoveToBack(entry.element)
+	return entry.identity, true
+}
+
+func (c *storedWorkflowIdentityCache) loadOrStore(
+	runID string,
+	identity storedWorkflowIdentity,
+) storedWorkflowIdentity {
+	c.mu.Lock()
+	defer c.mu.Unlock()
+
+	if entry, found := c.entries[runID]; found {
+		c.recency.MoveToBack(entry.element)
+		// Retry generations are monotonic. A report that loaded an older run
+		// row must never replace the identity cached by a completed RetryRun.
+		if entry.identity.retryGeneration > identity.retryGeneration {
+			return entry.identity
+		}
+		// Cache entries are valid only for the exact persisted manifest they
+		// were decoded from. Identity repairs can update that manifest without
+		// incrementing the retry generation, including from another replica.
+		if entry.identity.retryGeneration == identity.retryGeneration &&
+			entry.identity.manifestDigest == identity.manifestDigest {
+			return entry.identity
+		}
+		entry.identity = identity
+		c.entries[runID] = entry
+		return identity
+	}
+	if c.entries == nil {
+		c.entries = make(map[string]cachedStoredWorkflowIdentity)
+		c.recency = containerlist.New()
+	}
+	if len(c.entries) >= storedWorkflowIdentityCacheCapacity {
+		oldest := c.recency.Front()
+		delete(c.entries, oldest.Value.(string))
+		c.recency.Remove(oldest)
+	}
+	element := c.recency.PushBack(runID)
+	c.entries[runID] = cachedStoredWorkflowIdentity{identity: identity, element: element}
+	return identity
+}
+
+func (c *storedWorkflowIdentityCache) replaceAfterPersist(
+	runID string,
+	expectedManifestDigest [sha256.Size]byte,
+	identity storedWorkflowIdentity,
+) {
+	c.mu.Lock()
+	defer c.mu.Unlock()
+
+	if entry, found := c.entries[runID]; found {
+		if entry.identity.retryGeneration > identity.retryGeneration {
+			c.recency.MoveToBack(entry.element)
+			return
+		}
+		// A later report can commit and refresh the cache before an earlier
+		// reporter resumes after its own commit. Only advance an entry from the
+		// manifest this write replaced (or accept the exact new manifest), so
+		// delayed post-commit work cannot restore an older cache digest.
+		if entry.identity.retryGeneration == identity.retryGeneration &&
+			entry.identity.manifestDigest != expectedManifestDigest &&
+			entry.identity.manifestDigest != identity.manifestDigest {
+			c.recency.MoveToBack(entry.element)
+			return
+		}
+		entry.identity = identity
+		c.entries[runID] = entry
+		c.recency.MoveToBack(entry.element)
+		return
+	}
+	if c.entries == nil {
+		c.entries = make(map[string]cachedStoredWorkflowIdentity)
+		c.recency = containerlist.New()
+	}
+	if len(c.entries) >= storedWorkflowIdentityCacheCapacity {
+		oldest := c.recency.Front()
+		delete(c.entries, oldest.Value.(string))
+		c.recency.Remove(oldest)
+	}
+	element := c.recency.PushBack(runID)
+	c.entries[runID] = cachedStoredWorkflowIdentity{identity: identity, element: element}
+}
+
+func (c *storedWorkflowIdentityCache) delete(runID string) {
+	c.mu.Lock()
+	defer c.mu.Unlock()
+
+	entry, found := c.entries[runID]
+	if !found {
+		return
+	}
+	delete(c.entries, runID)
+	c.recency.Remove(entry.element)
 }
 
 func NewResourceManager(clientManager ClientManagerInterface, options *ResourceManagerOptions) *ResourceManager {
@@ -744,6 +886,7 @@ func (r *ResourceManager) CreateRun(ctx context.Context, run *model.Run) (*model
 		return nil, util.NewInternalServerError(err, "Failed to create a workflow for (%s)", executionSpec.ExecutionName())
 	}
 	// Update the run with the new scheduled workflow
+	run.Namespace = k8sNamespace
 	run.K8SName = newExecSpec.ExecutionName()
 	run.ServiceAccount = newExecSpec.ServiceAccount()
 	run.RunDetails.State = model.RuntimeState(string(newExecSpec.ExecutionStatus().Condition())).ToV2()
@@ -950,6 +1093,7 @@ func (r *ResourceManager) DeleteRun(ctx context.Context, runId string) error {
 	if err != nil {
 		return util.Wrapf(err, "Failed to delete a run %v", runId)
 	}
+	r.storedWorkflowIdentities.delete(runId)
 
 	if r.options.CollectMetrics {
 		if run.Conditions == string(exec.ExecutionSucceeded) {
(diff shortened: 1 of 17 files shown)


backend/src/apiserver/resource/resource_manager.go
```diff
@@ -2014,6 +2389,435 @@ func (r *ResourceManager) ReportWorkflowResource(ctx context.Context, execSpec u
...
+	}
+	if storedManifest == "" {
+		r.recordWorkflowReportRejection(workflowReportRejectionOwnershipUnresolved)
+		return storedWorkflowIdentity{}, util.NewInternalServerError(
+			errors.New("stored execution manifest is empty"),
+			"Failed to verify stored workflow identity before reporting run %s", run.UUID)
+	}
+	// Cache the persisted workflow identity in a bounded LRU so node-transition
+	// reports do not repeatedly decode a potentially large runtime manifest.
+	storedIdentity, found := r.storedWorkflowIdentities.load(run.UUID)
```
jeffspahr, line 2622: **[P1] Revalidate cached identity after same-generation DB changes**

This cache treats `(run UUID, retry_generation)` as sufficient freshness, but another API-server replica can update a legacy run's persisted workflow UID and namespace without changing its generation. A replica that previously cached the legacy empty UID then continues returning that entry here and rejects valid reports as `InvalidArgument`, which the persistence agent treats as permanent. This can leave run and task state stale until eviction, restart, or a later resync. Please validate cache entries using a persisted-manifest revision/hash, or avoid caching incomplete legacy identities; process-local `replace` and `delete` calls cannot keep replicas coherent.

jeffspahr (changes requested):
The new workflow-identity cache can retain stale data across API-server replicas, causing valid workflow reports to be rejected permanently. I reproduced the sequence with two resource managers sharing one store. Requesting changes because the cache needs cross-replica freshness semantics.

backend/src/apiserver/resource/resource_manager.go
herikwebb: Fixed in af276430b. Cached workflow identities are now bound to a SHA-256 digest of the exact persisted runtime manifest as well as the retry generation. A same-generation manifest repair therefore causes the replica to decode and replace its stale entry, while an older retry generation still cannot displace a newer cached identity. I added a two-ResourceManager regression using one shared store: the first manager caches a legacy empty UID/namespace, the second repairs the row, and the first then refreshes from the repaired manifest and accepts the valid report. The full resource suite, related client/server/storage suites, focused race tests, gofmt, go vet, and git diff checks pass.

backend/src/apiserver/resource/resource_manager.go
```diff
@@ -150,6 +168,112 @@ type ResourceManager struct {
...
+	identity storedWorkflowIdentity,
+) storedWorkflowIdentity {
+	c.mu.Lock()
+	defer c.mu.Unlock()
+
+	if entry, found := c.entries[runID]; found {
+		c.recency.MoveToBack(entry.element)
+		// Retry generations are monotonic. A report that loaded an older run
+		// row must never replace the identity cached by a completed RetryRun.
+		if entry.identity.retryGeneration > identity.retryGeneration {
```
jeffspahr, line 216: **[P1] Prevent stale same-generation identity rollback**

A digest mismatch proves only that manifests differ, not which one is newer. An overlapping request holding a UID-less legacy run can reach this branch after another report has adopted and cached UID `O`; because both rows have the same retry generation, the stale request replaces `O` with the old UID-less identity, reopens `legacyIdentityAdoption`, and can persist a live same-name replacement UID `X`. `UpdateRun` only fences on retry generation, so that overwrite becomes durable. A deterministic old-head/new-head comparison rejects this sequence at `2211304c` but accepts it and rewrites the stored UID at this head. Please use a monotonic persisted identity/revision or a database CAS followed by authoritative reload, and add coverage for the current-to-stale digest order.

backend/src/apiserver/resource/resource_manager.go
```diff
@@ -2014,6 +2399,440 @@ func (r *ResourceManager) ReportWorkflowResource(ctx context.Context, execSpec u
...
+		r.recordWorkflowReportRejection(workflowReportRejectionOwnershipUnresolved)
+		return storedWorkflowIdentity{}, util.NewInternalServerError(
+			errors.New("stored execution manifest is empty"),
+			"Failed to verify stored workflow identity before reporting run %s", run.UUID)
+	}
+	// Cache the persisted workflow identity in a bounded LRU so node-transition
+	// reports do not repeatedly decode a potentially large runtime manifest.
+	// The manifest digest keeps the cache coherent with repairs persisted by
+	// another API-server replica without requiring process-local invalidation.
+	manifestDigest := sha256.Sum256([]byte(storedManifest))
```
jeffspahr, line 2621: **[P2] Keep the cache synchronized after ordinary reports**

This digest covers the entire mutable Workflow manifest, including `resourceVersion`, status, and nodes, while the cached digest is refreshed after `UpdateRun` only when `workflowIdentityAdopted` is true. Every ordinary status transition therefore leaves the cache behind, so the next report hashes and JSON-decodes the full potentially large manifest again. That defeats the stated node-transition optimization and adds an extra full-manifest scan. Please refresh the cache after every successful manifest write, or key freshness to an immutable/monotonic identity revision.

jeffspahr (changes requested):
The manifest digest fixes the previously reported sequential cross-replica stale-cache case, but it introduces an unordered same-generation replacement path that can roll an adopted identity backward during overlapping reports. The full-manifest digest also leaves the cache ineffective for ordinary status transitions. Requesting changes for the identity rollback; the cache synchronization issue is included as a second inline finding.

backend/src/apiserver/resource/resource_manager.go
herikwebb: Fixed in 2a182654d. Every existing-run report now captures the exact persisted WorkflowRuntimeManifest, locks the run row, and compares both that manifest and RetryGeneration before writing. If another replica changed either value, the report stops before Workflow deletion or task persistence, invalidates and reloads the identity cache from the authoritative row, and returns retryable Unavailable. The new current-to-stale regression repairs legacy identity O through one ResourceManager, then gives another manager a stale UID-less snapshot and a same-name replacement X; the stale call is rejected, the database and cache retain O, and a fresh retry rejects X as InvalidArgument. A storage regression also proves a stale expected manifest cannot overwrite the repaired row.

backend/src/apiserver/resource/resource_manager.go
herikwebb: Fixed in 2a182654d. After every successful guarded report write, the cache is refreshed with the identity and SHA-256 digest of the exact manifest just persisted, not only after legacy adoption. The refresh compares against the prior digest and retry generation so delayed post-commit work cannot replace a newer cache entry. The regression uses a V2 run whose WorkflowRuntimeManifest is initially empty, verifies the PipelineRuntimeManifest fallback is cached, reports an ordinary nonterminal transition, and confirms the cache immediately matches the new persisted workflow manifest. Delayed M1-after-M2 and older-generation cache updates are covered separately.

backend/src/apiserver/resource/resource_manager.go
herikwebb: The sequential cross-replica cache repair remains covered, and 2a182654d adds the stronger database ordering required by the follow-up review: manifest plus retry-generation comparison under the run-row lock, authoritative reload on conflict, and ordered cache refresh after every committed report. This prevents both stale cache rejection and durable same-generation identity rollback.

backend/src/apiserver/storage/run_store.go
```diff
@@ -713,9 +721,64 @@ func (s *RunStore) GetRunByRecurringRunIDAndDisplayName(recurringRunID, displayN
...
+					"Failed to update run",
+				)
+			}
+			return false, util.NewInternalServerError(
+				lockError,
+				"Failed to lock run %s before updating",
+				run.UUID,
+			)
+		}
+		if currentWorkflowRuntimeManifest != *expectedWorkflowRuntimeManifest ||
```
jeffspahr: [P1] Fence updates against run-row reincarnation

`WorkflowRuntimeManifest` plus `RetryGeneration` is not a row-incarnation token: every newly created V2 run starts with `("", 0)`. Both API converters preserve a supplied run ID, and `CreateRun` only generates one when it is empty, so a deleted ID can be reused. If report processing preloads run A, A is deleted, and V2 run B is created under the same UUID, A's persisted-terminal report passes this check and overwrites B's name, namespace, state, and workflow manifest before cleanup returns NotFound for A. Please fence on a non-reusable row version/incarnation (or prevent ID reuse on every create path); at minimum, comparing the fallback `PipelineRuntimeManifest` would distinguish these pre-first-report rows, and this sequence needs regression coverage.

jeffspahr (changes requested):
The two previously reported issues are fixed, but the guarded write still permits a deleted-and-recreated V2 run to ABA-match a stale preloaded report. Requesting changes because this can overwrite the replacement run before the stale Workflow's NotFound is returned.

backend/src/apiserver/storage/run_store.go
herikwebb: Fixed in df93a9638. The guarded write now locks and compares RetryGeneration plus the exact persisted WorkflowRuntimeManifest. When that manifest is empty, as it is before the first V2 report, it also locks and compares PipelineRuntimeManifest. V2 CreateRun stores the post-Kubernetes-create Workflow there, including its server-assigned UID, and UpdateRun never changes that field, so deleting A and recreating B with the same run UUID cannot ABA-match even when both rows have empty workflow manifests and generation zero. The pipeline manifest is selected only for this empty-workflow case, avoiding an extra large-field read on ordinary reports. New storage and ResourceManager regressions delete V2 run A, recreate B with the same UUID and identical pipeline inputs, prove the stored live UIDs differ, submit the stale persisted-terminal report from A, and verify it returns Unavailable while the complete row and live Workflow for B remain unchanged. The full API-server suite and vet pass, and focused race regressions pass for 10 runs.

backend/src/apiserver/server/report_server.go
```diff
@@ -77,13 +81,32 @@ func (s *BaseReportServer) reportWorkflow(ctx context.Context, workflow string)
...
 
-	runId := newExecSpec.ExecutionObjectMeta().Labels[util.LabelKeyWorkflowRunId]
-	_, err = s.reportTasksFromExecution(newExecSpec, runId)
+	if newExecSpec.ExecutionStatus().HasNodes() && run == nil {
+		run, err = s.resourceManager.GetRun(runID)
+		if err != nil {
+			return nil, util.Wrap(err, "Failed to load run before reporting task details")
+		}
+	}
+	_, err = s.reportTasksFromExecution(newExecSpec, run)
```
jeffspahr, line 109: [P1] Fence task persistence against run-row reincarnation

`run` was loaded before `ReportWorkflowResourceWithRun`, but this call occurs after that method’s guarded run transaction has committed. If run A is deleted and run B is created with the same UUID in this window, `CreateOrUpdateTasksForRun` validates against stale A and then upserts A’s tasks under B’s UUID. Because it also trusts A’s namespace, B can belong to a different namespace. I reproduced this by deleting and recreating the row between these calls; the stale task insert succeeds. Please verify the immutable run incarnation in the same transaction as the task upsert (or otherwise make workflow and task persistence atomic), and add regression coverage.

jeffspahr (changes requested):
The previously reported run-row ABA is fixed, but task persistence retains a separate run-incarnation race. Requesting changes because a workflow report can attach task data to a deleted-and-recreated run that reuses the UUID.

backend/src/apiserver/server/report_server.go
herikwebb: Fixed in e8cfeaa25. Task persistence now begins a database transaction, locks the parent run row first with SELECT FOR UPDATE, and verifies the exact namespace, WorkflowRuntimeManifest, the UID-bearing PipelineRuntimeManifest fallback, and RetryGeneration before reading or upserting any tasks in that same transaction. This uses the same parent-to-children lock order as run deletion: if task persistence wins, deletion waits and then removes those tasks with run A; if deletion or recreation wins, the task transaction observes a missing or changed parent and returns retryable Unavailable without mutating task input or storage. The ResourceManager then invalidates and reloads its cached identity from the authoritative replacement row. New regressions reproduce the post-workflow-report delete/recreate window through a second ResourceManager, verify no stale task reaches cross-namespace run B, and isolate both identical-manifest namespace changes and same-namespace manifest changes. Matching guarded writes are covered as a positive control. The full API-server and serial non-integration backend unit suites, 20 repeated focused runs, 10 race-enabled runs, vet, gofmt, and diff checks pass.

jeffspahr:
Re-reviewed at e8cfeaa25bd105ea44790cfb84fa92b9f8c618cd. No actionable findings.

The previously reported task-persistence race is fixed. Task persistence now locks and verifies the parent run namespace, runtime manifests, and retry generation in the same transaction as task lookup and upsert. Run deletion uses the same parent-first lock order, preventing delete/recreate from interleaving before the task commit. Coverage includes both same-namespace UID changes and cross-namespace run reincarnation.

Verification:
- Affected storage, resource, server, client, and GC package tests pass.
- Focused run/task incarnation regressions pass under the race detector for 10 repetitions.
- Focused go vet and git diff --check pass.

jeffspahr:
CI is currently red because of one deterministic, PR-related lint error:

- The pre-commit job reports Staticcheck QF1008 at backend/src/apiserver/storage/task_store.go:453: the embedded DB field can be removed from the selector.
- Change s.db.DB.Begin() to s.db.Begin().
- check_ci_status is only red because it aggregates the pre-commit failure; it has no separate failure.

Other completed checks are passing, while several matrix jobs are still running.

Failed job: https://github.com/kubeflow/pipelines/actions/runs/33879963399/job/101050488775

herikwebb:
Fixed in 7c69ea6b4 by using the promoted s.db.Begin() method. The branch is also rebased onto current upstream master. The exact golangci-lint v2.10.1 lane on the project Go 1.26.6 toolchain now reports 0 issues; the full API-server test suite, focused storage/resource tests, vet, gofmt, and git diff checks pass.

jeffspahr:
Re-reviewed at 7c69ea6b480f3a06e59d0e0025b9e3b23fd98c13. No actionable findings.

The only PR-owned change from the previously reviewed head is s.db.DB.Begin() to s.db.Begin(). It resolves Staticcheck QF1008 while invoking the same promoted database/sql transaction method, so the task-fencing transaction and lock behavior are unchanged.

Verification:
- Affected storage, resource, server, client, and GC package tests pass.
- Focused run/task incarnation regressions pass under the race detector for 10 repetitions.
- Focused go vet, git diff --check, and the GitHub pre-commit check pass.

CI currently has no failed checks; one check remains pending.

PR #14263, opened 2026-09-03
fix(frontend): prevent artifacts from rendering as active content

herikwebb:

**Why:** Artifact and pod-log responses can contain untrusted content. If a browser interprets that content inline, active content can run in the Kubeflow Pipelines UI origin. Interrupted streams can also look like complete downloads, and downloads must remain byte-preserving while central and tenant services are on different versions.

**What:** This change makes those responses download as inert files, preserves safe filenames and original archive bytes, and ensures interrupted storage work fails visibly. Users can still preview and download artifacts normally, including during rolling upgrades, but tenant responses can no longer change browser-wide state through unsafe headers.

Technical details:

- Apply `Content-Disposition: attachment` and `X-Content-Type-Options: nosniff` before handling artifact, proxy, authentication, error, and pod-log responses.
- Allowlist safe tenant response headers, force proxied artifacts to `application/octet-stream`, and sanitize upstream filenames.
- Keep browser download URLs query-based, then translate them at the trusted proxy hop to the legacy raw-download route for old tenant services.
- Preserve binary GCS content and raw S3/MinIO archives without implicit extraction.
- Propagate storage, decompression, archive, volume, and proxy failures through controlled response handling.
- Detect already-failed pod-log streams and cancel upstream streams and directory scans when clients disconnect.
- Use the current `http-proxy-middleware` response hooks for both artifact and pod-log proxies.

Validation with Node.js 24.14.0:

- Full frontend server suite: 435 passed, 2 TODO.
- Full frontend UI suite: 2,071 passed.
- Focused feedback regressions: 63 server tests passed.
- Focused affected UI tests: 97 passed.
- Frontend production build passed, including ESLint and TypeScript.
- Frontend server TypeScript build passed.
- Prettier and `git diff --check` passed.

The affected behavior and fix were already public in [herikwebb/pipelines#116](https://github.com/herikwebb/pipelines/pull/116).

**Checklist:**

- [x] The title for your pull request follows the repository title convention.

### DIFF
--- a/frontend/server/app.ts
+++ b/frontend/server/app.ts
@@ -35,9 +35,22 @@ import { getClusterNameHandler, getProjectIdHandler } from './handlers/gke-metad
 import { getAllowCustomVisualizationsHandler } from './handlers/vis.js';
 import { getIndexHTMLHandler } from './handlers/index-html.js';
 
-import { Server } from 'http';
+import { IncomingMessage, Server } from 'http';
 import { HACK_FIX_HPM_PARTIAL_RESPONSE_HEADERS } from './consts.js';
 import registerTensorboardProxy from './handlers/tensorboard-proxy.js';
+
+function hardenPodLogsProxyResponse(proxyRes: IncomingMessage): void {
+  proxyRes.headers['x-content-type-options'] = 'nosniff';
+  proxyRes.headers['content-disposition'] = 'attachment';
+  proxyRes.headers['content-type'] = 'text/plain; charset=utf-8';
+}
+
+const hardenPodLogsProxyRequest: express.Handler = (_req, res, next) => {
+  res.setHeader('X-Content-Type-Options', 'nosniff');
+  res.setHeader('Content-Disposition', 'attachment');
+  res.type('text/plain');
+  next();
+};
 
 function getRegisterHandler(app: Application, basePath: string) {
   return (
@@ -252,12 +265,14 @@ function createUIServer(options: UIConfigs) {
   if (options.artifacts.streamLogsFromServerApi) {
     app.all(
       '/k8s/pod/logs',
+      hardenPodLogsProxyRequest,
       createProxyMiddleware({
         changeOrigin: true,
         on: {
           proxyReq: (proxyReq) => {
             console.log('Proxied log request: ', proxyReq.path);
           },
+          proxyRes: hardenPodLogsProxyResponse,
         },
         headers: HACK_FIX_HPM_PARTIAL_RESPONSE_HEADERS,
         pathRewrite: (pathStr: string, req: any) => {
@@ -286,12 +301,14 @@ function createUIServer(options: UIConfigs) {
   if (options.artifacts.streamLogsFromServerApi) {
     app.all(
       '/k8s/pod/logs',
+      hardenPodLogsProxyRequest,
       createProxyMiddleware({
         changeOrigin: true,
         on: {
           proxyReq: (proxyReq) => {
             console.log('Proxied log request: ', proxyReq.path);
           },
+          proxyRes: hardenPodLogsProxyResponse,
         },
         headers: HACK_FIX_HPM_PARTIAL_RESPONSE_HEADERS,
         pathRewrite: (pathStr: string, req: any) => {
--- a/frontend/server/handlers/artifacts.ts
+++ b/frontend/server/handlers/artifacts.ts
@@ -27,8 +27,12 @@ import {
   listObjectsUnderPrefix,
   summarizeDirectoryUnderPrefix,
 } from '../minio-helper.js';
+import type { MinioRequestConfig } from '../minio-helper.js';
 import * as tar from 'tar-stream';
 import * as zlib from 'zlib';
+import type { IncomingMessage } from 'http';
+import { Readable } from 'stream';
+import { pipeline as pipelinePromise } from 'stream/promises';
 import * as serverInfo from '../helpers/server-info.js';
 import { Handler, Request, Response, NextFunction } from 'express';
 import { createProxyMiddleware } from 'http-proxy-middleware';
@@ -64,6 +68,8 @@ interface ArtifactsQueryStrings {
   /** optional provider info to use to query object store */
   providerInfo?: string;
   namespace?: string;
+  /** return the artifact byte-for-byte without archive extraction */
+  download?: string;
 }
 
 type ArtifactSource = ArtifactsQueryStrings['source'];
@@ -76,6 +82,7 @@ const ARTIFACT_QUERY_PARAMETER_NAMES = [
   'providerInfo',
   'namespace',
   'peek',
+  'download',
 ] as const;
 
 export interface S3ProviderInfo {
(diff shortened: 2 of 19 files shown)


frontend/server/handlers/artifacts.ts
```diff
@@ -1030,6 +1342,19 @@ export function getArtifactsProxyHandler({
       proxyReq: (proxyReq) => {
         console.log('Proxied artifact request: ', proxyReq.path);
       },
+      // http-proxy-middleware copies upstream headers after this outer handler
+      // starts. Rewrite the proxy response itself so a tenant-side artifact
+      // service cannot replace the attachment guard with `inline` or remove
+      // nosniff while returning active HTML.
+      proxyRes: (proxyRes) => {
+        delete proxyRes.headers['set-cookie'];
```
jeffspahr: [P1] Strip origin-wide headers from tenant proxy responses. This denylist still forwards `Clear-Site-Data`. A namespaced artifact service can therefore return `Clear-Site-Data: "cookies", "storage"` through the central KFP origin; the same-origin ArtifactPreview fetch causes Chromium to process it and clear the central session cookie and local storage even though the response is an attachment with `nosniff`. Please strip this header at minimum, and preferably forward only an explicit allowlist of safe representation headers.

frontend/server/handlers/artifacts.ts
```diff
@@ -1030,6 +1342,19 @@ export function getArtifactsProxyHandler({
...
+        delete proxyRes.headers['access-control-allow-origin'];
+        delete proxyRes.headers['access-control-allow-credentials'];
+        proxyRes.headers['x-content-type-options'] = 'nosniff';
```
jeffspahr: [P2] Force an inert MIME type for untrusted proxied artifacts. This hook preserves the tenant-controlled `Content-Type`. Chromium still executes a response labeled `application/javascript` when loaded as a script subresource even when it also carries `Content-Disposition: attachment` and `X-Content-Type-Options: nosniff`; attachment is not an execution barrier for subresources. Please override the proxy response type with `application/octet-stream`.

frontend/src/lib/Apis.ts
```diff
@@ -346,10 +346,15 @@ export class Apis {
   }) {
     const { source, bucket, key } = path;
     if (isDownload) {
-      return `artifacts/${source}/${bucket}/${key}${buildQuery({
+      // Keep object keys in the query so browsers do not normalize standalone dot path segments.
+      return `artifacts/get${buildQuery({
```
jeffspahr, line 350: [P2] Preserve download behavior across central/tenant version skew. This now forwards the new `download=true` query protocol to each per-namespace artifact service. During a rolling upgrade an older service ignores that unknown flag and handles `/artifacts/get` in extraction mode, so a requested `.tar.gz` returns HTTP 200 with decompressed or first-member bytes instead of the original archive. The central proxy cannot recover those bytes. Please retain/translate to a backward-compatible upstream download route, negotiate capability/version, or otherwise ensure tenant services are upgraded before emitting this protocol; add a proxy regression test against the previous handler.

frontend/server/handlers/pod-logs.ts
```diff
@@ -120,7 +124,17 @@ export function getPodLogsHandler(
 
     try {
       const stream = await getPodLogsStream(podName, createdAt, podNamespace);
+      if (res.destroyed || res.writableEnded) {
```
jeffspahr, line 127: [P2] Handle streams that failed before this post-await listener is attached. `getObjectStream` installs an earlier default error listener and starts its pipeline before returning, so a fast storage/decompression error can leave `stream.errored` set before this handler resumes. The later listener never sees that event, and piping the already-failed stream leaves the HTTP response open indefinitely. Please install the request-specific callback before stream production starts, or attach a fail-once handler and atomically check `stream.errored`/destroyed state.

frontend/server/app.ts
```diff
@@ -258,6 +263,7 @@ function createUIServer(options: UIConfigs) {
           proxyReq: (proxyReq) => {
             console.log('Proxied log request: ', proxyReq.path);
           },
+          proxyRes: hardenPodLogsProxyResponse,
```
jeffspahr, line 275: [P3] Harden middleware-generated proxy failures too. `proxyRes` only runs after an upstream response; DNS, TLS, and refused-connection failures generated by `http-proxy-middleware` therefore return without either hardening header. Please set the headers on the Express response before invoking the proxy, while retaining this hook so an upstream cannot override them.

jeffspahr (changes requested):
Requesting changes for the four substantive issues below; I also left one lower-priority hardening comment.

Verification: 179 focused server tests, 97 focused UI tests, the full frontend build (ESLint, TypeScript, Tailwind, and Vite), Prettier, and `git diff --check` passed. Targeted Chromium and Node stream reproductions confirmed the browser-header and early-stream-error behavior.

frontend/server/handlers/artifacts.ts
herikwebb: Fixed in cebf85259: tenant artifact responses now pass through an explicit representation-header allowlist. Clear-Site-Data, Set-Cookie, CORS, CSP, Link, Location, Refresh, and arbitrary tenant headers are removed; the integration regression exercises these cases.

frontend/server/handlers/artifacts.ts
herikwebb: Fixed in cebf85259: successful tenant artifact responses are now forced to application/octet-stream with nosniff and a sanitized attachment disposition. The regression uses a tenant-controlled application/javascript response and verifies the inert output type.

frontend/src/lib/Apis.ts
herikwebb: Fixed in cebf85259: browser-facing downloads remain query-based, but the central proxy translates download=true at the final hop to the legacy path download route. The complete key is encoded as one segment, preserving slashes and dot segments. A previous-version handler regression verifies that a tar archive remains byte-for-byte rather than being extracted.

frontend/server/handlers/pod-logs.ts
herikwebb: Fixed in cebf85259: the pod-log handler attaches a fail-once listener and then synchronously checks stream.errored and stream.destroyed before piping. A regression supplies a stream whose error was emitted before the handler receives it and verifies a terminal 500 rather than an open response.

frontend/server/app.ts
herikwebb: Fixed in cebf85259: both streamed pod-log proxy registrations now run an Express hardening middleware before the proxy. The proxy-response hook remains in place, and a refused-connection regression verifies that middleware-generated failures retain attachment and nosniff headers.

jeffspahr:
Re-reviewed current head cebf85259d73b1bd716bf6237135e636961b4a6b. No actionable findings.

The five previously raised concerns are addressed in this revision. Verification included 218 focused server tests, the frontend server TypeScript build, git show --check, and an independent proxy-path check confirming that slash- and dot-segment-containing keys remain intact through the proxy hop.

PR #14264, opened 2026-09-03
workflow to gatekeep ci-workflows: Part of KEP  #13293

modichika:
This PR addresses the problem of overloading PRs that arrives in the repository of kubeflow/pipelines. It finishes part-2 of [KEP-13293](https://github.com/kubeflow/pipelines/pull/13293).

The workflow checks when a PR gets opened if the author is a member of Kubeflow org then the ci-checks will automatically be added otherwise if the author is not a member, and the PR has a issue linked with a /ready label (this signifies that the maintainers have agreed to work on the issue after discussions in meetings and channels that it's needed for the kfp architecture) then only the ci-checks will pass or else the PR will be closed automatically with a logical message of why it's getting closed.

Co-authored with: @alyssacgoins @zazulam 

**Checklist:**


droctothorpe:
Thanks for the work here — I reviewed the current diff and have a few concerns:

1. **Please update `docs/agents/ci.md` for the new workflow.**
   The repo CI guide explicitly says to update it when changing workflows / failure handling, and this PR adds a new workflow (`.github/workflows/ci-gatekeeper.yml`) without documenting it.

2. **The PR admission logic looks duplicated / split across multiple workflows.**
   We already have related gating logic in:
   - `.github/workflows/gh-workflow-approve.yml`
   - `.github/workflows/ci-checks.yml`

   Adding a third workflow with overlapping membership / label / CI-admission behavior makes the policy harder to reason about and maintain. It might make more sense to extend the existing path over introducing a parallel one.

3. **The `/ready` vs `ready` distinction looks wrong, and I can’t find either label in this repo today.**
   The workflow checks for a literal label named `/ready`.

   But at the moment, in `kubeflow/pipelines`:
   - issues with label `ready`: **0**
   - issues with label `/ready`: **0**

   So as written, the success path looks effectively unsatisfiable. Also, `/ready` reads more like a slash command than a normal GitHub label string, so I think this needs clarification / correction.

modichika:
@droctothorpe  I added the membership to explicitly check that if any PR is opened by the member of Kubeflow Pipelines then it will automatically pass the ci-checks, otherwise the PR needs a issue that has a /ready label to it by the maintainers after discussion if the issue is needed to be solved or not, as discussed in the KEP - 13293 part-2. Sorry for the confusion I'll write a description of this in the PR.

ci-checks and gh-workflows has membership but the main objective of gatekeeping-ci workflows is that PRs are flooded in the repo and we don't need that so this workflow is needed to check if the author of the PR is a member or not.

correct about the 3rd point the /ready label is not yet in the repo we need to add a label for this workflow to work as well.
cc: @alyssacgoins  PTAL

modichika:
I have a concern regarding the failure of pre-commit when it's passing in my local vs code but not in this PR branch, any suggestions?

PR #14311, opened 2026-09-07
fix(backend): prevent database connection leaks in storage layer

### LINKED ISSUE #14310: bug(backend): Database connection leaks across apiserver/storage due to delayed rows.Close()
### Environment

*  **How did you deploy Kubeflow Pipelines (KFP)?** N/A (Backend logic defect)
*  **KFP version:** `master`
*  **KFP SDK version:** N/A 

### Steps to reproduce

Across the `apiserver/storage` layer, there is a pervasive anti-pattern where `defer rows.Close()` and `defer sizeRow.Close()` are placed *after* error-handling logic (like `rows.Err()` or parsing operations) that can return early. 

In Go's `database/sql` package, if `*sql.Rows` is not explicitly closed, the underlying database connection remains held until the rows are fully consumed. When an error occurs before the `defer` is registered, the connection leaks, permanently exhausting the connection pool under load.

**Example from `run_store.go`:**
```go
	rows, err := tx.Query(rowsSql, rowsArgs...)
	if err != nil {
		return errorF(err)
	}
        // ---> IF AN ERROR HAPPENS BELOW, THE CONNECTION LEAKS!
	if err := rows.Err(); err != nil {
		return errorF(err)
	}
	runs, err := s.scanRowsToRuns(rows)
	if err != nil {
		return errorF(err)
	}
	defer rows.Close() // <--- REGISTERED TOO LATE

### PR DESCRIPTION
**Pull request by Ayush-kathil**
### What does this PR do?
Moves `defer rows.Close()` and `defer sizeRow.Close()` to execute immediately after `tx.Query` error checks across the `apiserver/storage` controllers. 

Previously, these defers were placed after `rows.Err()` or scanning operations. If any of those intermediate steps failed, the function returned early without closing the rows, leaking the connection from the `database/sql` pool. 

### Motivation and Context
Fixes #14310

Prevents the API server from freezing with `Too many connections` errors. When connections leak during early returns under heavy load or degraded DB performance, the connection pool rapidly exhausts.

### How Has This Been Tested?
- Verified with `go test ./backend/src/apiserver/storage/... -v`

### Type of changes
- [x] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)

### CODE REVIEW

**Comment by Ayush-kathil** [comment]
Hey @jeffspahr, this PR fixes the database connection leaks in the storage layer (Issue #14310). I relocated the `defer rows.Close()` statements immediately after the query checks so we don't permanently exhaust the connection pool on early returns. Let me know if this looks good to merge!

File: backend/src/apiserver/storage/pipeline_store.go [hunk 1]
```diff
@@ -387,14 +388,14 @@ func (s *PipelineStore) ListPipelines(filterContext *model.FilterContext, opts *
...
 	}
-	rows.Close()
 
 	// Count pipelines
 	sizeRow, err := tx.Query(sizeSQL, sizeArgs...)
 	if err != nil {
 		tx.Rollback()
 		return nil, 0, "", util.NewInternalServerError(err, "Failed to count pipelines")
 	}
+	defer sizeRow.Close()
```
jeffspahr, line 401: **[P1] Close count results before querying tags**

`ScanRowToTotalSize` calls `Next()` only once, leaving the count result unexhausted. Replacing the immediate `Close()` with a defer means nonempty pipeline lists execute their tag query on the same MySQL connection with unread result data, which can fail with a busy-buffer/connection error. The same regression occurs in `ListPipelineVersions` at line 1256.

Keep the new defer for early returns, but restore the explicit `sizeRow.Close()` before each tag query. Repeated `Close()` calls are supported by [database/sql](https://pkg.go.dev/database/sql#Rows.Close).

jeffspahr (changes requested):
Found one P1 blocker in the count-result lifecycle.

Verification: storage tests passed at this revision, and changed-file formatting checks passed. Hosted backend tests and pre-commit also passed. The finding is based on source and MySQL-driver inspection; a live MySQL reproduction was not run.

jeffspahr:
No new actionable findings at `dceac339`.

The [previous P1 finding](https://github.com/kubeflow/pipelines/pull/14311#discussion_r3951124224) is fixed: both pipeline-list methods close count results before querying tags while retaining deferred cleanup.

Verification:
- `go test ./backend/src/apiserver/storage/...` passed.
- All four changed files pass `gofmt`.
- No live MySQL test run.
- Hosted backend tests and pre-commit were still in progress at review time, so merge readiness remains provisional.

backend/src/apiserver/storage/pipeline_store.go
Ayush-kathil: Thanks @jeffspahr for catching this! You are absolutely right `ScanRowToTotalSize` leaving the `sizeRow` unexhausted causes the subsequent `getTags` query in the same transaction to hit a busy-buffer on the MySQL connection, which caused the E2E tests to fail.

I have amended the PR to restore the explicit `rows.Close()` and `sizeRow.Close()` calls immediately before the tag queries in `ListPipelines` and `ListPipelineVersions`, while keeping the safe `defer` placements at the top for early-return protection. The DCO is also formally signed. Thanks for the thorough review!

Ayush-kathil:
Awesome, thanks for the fast re-review and verification @jeffspahr! 

Glad to hear the structural fix cleanly resolves both the early-return leaks and the busy-buffer constraints. I'll keep an eye on the remaining CI workflows as they finish up. Let me know if you need anything else from my end once they turn green!

jeffspahr:
Could we also add `defer tx.Rollback()` immediately after the successful `Begin()` in `ListRuns`?

```go
tx, err := s.db.Begin()
if err != nil {
    glog.Error("Failed to start transaction to list runs")
    return errorF(err)
}
defer tx.Rollback()
```

The initial `tx.Query` error branch currently returns without ending the transaction. Deferring `rows.Close()` does not cover that branch because the query failed before returning rows. Even a read-only transaction holds a connection until it ends.

The existing commit and explicit rollback calls can stay; the deferred rollback harmlessly returns `sql.ErrTxDone` after either has completed.

Ayush-kathil:
Done! Good catch on that `tx.Query` error branch returning early without closing the transaction. 

I went ahead and added `defer tx.Rollback()` immediately after `s.db.Begin()` not just in `ListRuns`, but systematically across the entire `storage` package wherever a transaction is started. As you mentioned, it harmlessly catches any early returns and keeps the whole database layer perfectly leak-proof. 

DCO is signed, formatting is clean, and the local test suite is completely green. Let me know if we're good to merge once CI finishes!

jeffspahr:
No actionable findings at `84d15cc8`. Looks good to merge.

- The requested rollback guard is added; the broader transaction cleanup is sound.
- The previous count-result fix remains intact.
- A temporary regression test confirms the improvement: the prior revision retains one connection after query failure; this revision releases it.
- Storage tests and formatting pass. At review time, all 182 hosted checks had passed; only Tide remained pending.

The regression test is not included in the PR.

PR #14326, opened 2026-09-08
fix(frontend): avoid false artifact truncation markers

### LINKED ISSUE #14278: bug(frontend): artifact previews show truncation marker when no content is removed
### Environment
* How did you deploy Kubeflow Pipelines (KFP)?
Not deployment-specific, this is a frontend logic bug verified in the V2 artifact-preview paths (Input/Output tab, Runtime Node Details).
* KFP version: master

### Steps to reproduce
Ran into this while going through `ArtifactPreview.tsx`. The no-truncation check is:

```ts
if (data.length <= maxbytes && (!maxlines || data.split('\n').length < maxlines)) {
  return data;
}
```

The problem is at the `maxlines` boundary. If the content is within `maxbytes` and has exactly `maxlines` lines, this check fails and the early return gets skipped. The following `slice(0, maxlines)` doesn't actually remove anything since there's nothing left to cut, but by then it's already going down the truncation path and adds a `...` anyway.

`maxlines` defaults to 20, and that default is what's actually being used in the V2 UI right now. Input/Output tab and Runtime Node Details both call `ArtifactPreview` without overriding it, so content that fits within the byte limit and has exactly 20 lines can hit this case.

There's a similar bug in `MinioArtifactPreview.tsx`:

```ts
if (data.length <= maxbytes && !maxlines) {
  return { data, hasMore: false };
}
```

Here, just passing any `maxlines` value at all skips this early return, regardless of whether the content actually has more lines than that. So it can end up returning `hasMore: true` and appending `...` even though the full content already fits within both limits. Nothing in prod currently passes `maxlines` to this component, so this part appears to be latent at the moment.

To reproduce, for `ArtifactPreview`:
1. Use content that's within `maxbytes` and has exactly 20 lines.
2. Render it through a V2 artifact preview using the default `maxlines = 20`.
3. All 20 lines show up fine, but there's still a trailing `...`.

For `MinioArtifactPreview`, pass a `maxlines` value greater than the actual line count while keeping the content within `maxbytes`. The full content comes back but it's still marked as truncated.

### Expected result
`...` and `hasMore: true` should only show up when content was actually removed because of the byte or line limit. If everything fits within both limits, it should come back clean, no truncation marker.

### Materials and Reference
The existing tests only cover cases where content actually exceeds the line limit. There's nothing testing content that's exactly at `maxlines` or under it, which is how this slipped through. Both files end up with the same wrong result, just get there through slightly different conditions.

---
Impacted by this bug?

Amanyadav0556:

Fixes artifact previews showing a truncation marker even when no content is actually removed.

`ArtifactPreview` treated content exactly at the `maxlines` limit as truncated. `MinioArtifactPreview` could also report `hasMore: true` when the content fit within both the byte and line limits.

This PR fixes both boundary cases and adds regression tests for the line-limit behavior.

Fixes #14278

**Checklist:**

### DIFF
--- a/frontend/src/components/ArtifactPreview.test.tsx
+++ b/frontend/src/components/ArtifactPreview.test.tsx
@@ -123,6 +123,43 @@ describe('ArtifactPreview', () => {
     await waitFor(() => screen.getByText(`012 345 ...`));
   });
 
+  it('does not truncate when line count is below maxlines', async () => {
+    const data = `012\n345`;
+    vi.spyOn(Apis, 'readFile').mockResolvedValueOnce(data);
+    render(
+      <CommonTestWrapper>
+        <ArtifactPreview
+          value={'minio://bucket/key'}
+          namespace={'kubeflow'}
+          maxbytes={data.length}
+          maxlines={20}
+        />
+      </CommonTestWrapper>,
+    );
+    await waitFor(() => screen.getByText('minio://bucket/key'));
+    await waitFor(() => screen.getByText('012 345'));
+    expect(screen.queryByText(/\.\.\./)).not.toBeInTheDocument();
+  });
+
+  it('does not truncate when line count exactly equals maxlines', async () => {
+    const lines = Array.from({ length: 20 }, (_, i) => `line${i}`);
+    const data = lines.join('\n');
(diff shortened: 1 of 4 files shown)


droctothorpe:
Hi @Amanyadav0556, thank you for contributing to Kubeflow Pipelines.

PR #14331, opened 2026-09-08
fix(backend): fix broken pagination for several run and task sort fields

### LINKED ISSUE #14330: bug(backend): several run and task sort_by fields break pagination
### Environment

*  How did you deploy Kubeflow Pipelines (KFP)? Any deployment. The problem is in the API server.
*  KFP version: master (2.17.1)
*  KFP SDK version: not relevant, the fault is server side

### Steps to reproduce

List runs sorted by `experiment_id` with more runs than the page size:

```
GET /apis/v2beta1/runs?page_size=2&sort_by=experiment_id
```

The call fails with:

```
Invalid input error: cannot sort by field "ExperimentUUID" on type "Run"
```

The rows query already ran at that point, so the page is thrown away and the caller gets nothing back. The same request succeeds when the result set fits in one page, because a page token is only built when more rows remain. That makes the field look supported until a user has enough runs.

The same failure applies to:

| API | `sort_by` value | model field it maps to |
| --- | --- | --- |
| ListRuns | `experiment_id` | `ExperimentUUID` |
| ListRuns | `recurring_run_id` | `JobUUID` |
| ListTasks | `run_id`, `runId` | `RunUUID` |
| ListTasks | `start_time` | `StartedTimestamp` |
| ListTasks | `parent_task_id` | `ParentTaskUUID` |

### Cause

`list.Options.nextPageToken` reads the sorted value through `Listable.GetFieldValue`, and it passes the model field name that `GetField` took from `APIToModelFieldMap`. Those maps hold database column names. `Run.GetFieldValue` and `Task.GetFieldValue` switch on Go struct field names instead, so every field whose column name differs from its struct field name resolves to nil:

```go
"experiment_id": "ExperimentUUID",   // runAPIToModelFieldMap
...
case "ExperimentId":                 // Run.GetFieldValue
    return r.ExperimentId
```

`nextPageToken` turns that nil into `cannot sort by field ...`.

Filtering on the same fields is fine. Filters use the column name directly and never call `GetFieldValue`.

`state_history` also resolves to nil, but for a different reason. It maps to a serialized JSON column that neither `RunStore` nor `TaskStore` keeps in memory after parsing, so there is no value left to put in a page token.

### Expected result

Every value that `sort_by` accepts should paginate.

### Materials and Reference

- `backend/src/apiserver/model/run.go`: `runAPIToModelFieldMap`, `Run.GetFieldValue`
- `backend/src/apiserver/model/task.go`: `taskAPIToModelFieldMap`, `Task.GetFieldValue`
- `backend/src/apiserver/list/list.go`: `Options.nextPageToken`
- `TestTaskAPIFieldMap` already checks the field map against the SQL columns. There is no matching check against `GetFieldValue`.

---

Neilblaze:
Fixes #14330 🚀

This targets `release-2.18`, which still has the MLMD task model. On master #13986 replaced the task model, and the run half of this fix goes to master in #14400.

`Options.nextPageToken` looks up the sorted value with `Listable.GetFieldValue` and passes the model field name that `GetField` read out of `APIToModelFieldMap`. Those maps hold database column names, but `Run.GetFieldValue` and `Task.GetFieldValue` switched on Go struct field names. Any field where the two differ resolved to nil, and the whole list call failed with `cannot sort by field ...`. Only calls that need a page token reach that code, so the sort works until the result set outgrows one page.

Fields fixed:

| API | `sort_by` value | model field |
| --- | --- | --- |
| ListRuns | `experiment_id` | `ExperimentUUID` |
| ListRuns | `recurring_run_id` | `JobUUID` |
| ListTasks | `run_id` | `RunUUID` |
| ListTasks | `start_time` | `StartedTimestamp` |
| ListTasks | `parent_task_id` | `ParentTaskUUID` |

The switch labels now match the names in the field maps, and `Task` gains the missing `StartedTimestamp` case. The field maps themselves are unchanged, so filtering behaves exactly as before.

The camel case alias `runId` is not fixed by this. `NewOptions` lowercases the sort field to `runid`, which is not in the task field map, so it fails before paging starts.

`TestTaskAPIFieldMap` already asserts that every mapped field is a real SQL column. `TestGetFieldValue_ResolvesEveryMappedField` adds the other half of that check for all six listable models, so a new sort field cannot land without a `GetFieldValue` case behind it.

**What this PR does not cover:**

`state_history` stays unresolved. It maps to a serialized JSON column that both stores parse into a slice and then drop, so there is no value left to carry in a page token. The new test lists it explicitly and fails if it ever starts resolving.

**Testing:**

- `TestGetFieldValue_ResolvesEveryMappedField` (`backend/src/apiserver/list`)
- `TestListRuns_SortByExperimentIdAndRecurringRunIdPaginates` (`backend/src/apiserver/storage`)
- `TestListTasks_SortByStartTimeRunIdAndParentTaskIdPaginates` (`backend/src/apiserver/storage`)

All three fail on `release-2.18` with `cannot sort by field ...` and pass with this change. The `list`, `model` and `storage` suites are green.

**Checklist:**


Neilblaze:
CC'ing @jeffspahr for further visibility. Thanks!

Neilblaze:
@jeffspahr fixed the merge conflict, I think now it's good to go! :)

jeffspahr:
@Neilblaze, thanks for updating this. Now that #13986 has merged, could you take all three pieces below and link the new PRs here?

1. **Retarget this PR (#14331) to `release-2.18`.** That branch retains the MLMD-based task model, so the original run and task getter fixes are still needed there. Keep those fixes and their regression coverage, and run CI against `release-2.18`. The current diff against that branch is still limited to the five intended files. Please also correct the description's `runId` claim: the existing parser lowercases it to `runid`, which is absent from the legacy field map; `run_id` is fixed by this PR, but the camel-case alias still fails before pagination.

2. **Open a separate PR against `master` for the surviving run fixes and adapted tests.** `Run.GetFieldValue` still needs `ExperimentId` → `ExperimentUUID` and `RecurringRunId` → `JobUUID` in its switch labels. I reproduced both pagination errors at the #13986 merge, `b93fd252c`. Applying only those two corrections, with the run regression adapted to the new API, makes the complete `list`, `model`, and `storage` package suites pass.

   Drop the obsolete task getter changes: the native model already resolves `RunUUID`, `StartedInSec`, and `ParentTaskUUID`. Adapt the tests to the new model rather than restoring removed `RunID`, `ParentTaskId`, `StartedTimestamp`, `PodName`, or `MLMDExecutionID` fields. The run test now needs `ListRuns(filterContext, opts, false)` (`false` disables task hydration). Revise the mapping guard as well: nullable pointer/JSON fields can legitimately be nil, and native Task now returns `StateHistory`, so the old blanket exception/comment no longer describes it. Use suitable populated fixtures and token round-trip coverage for supported scalar sort fields; a nil optional value is not proof of a missing getter.

3. **Open a separate follow-up PR against `master` for native parentless-task pagination.** This is a distinct bug reproduced after #13986, and the old task getter changes in this PR do not address it.

   **Reproduction at `b93fd252c`:** create three tasks with `ParentTaskUUID == nil` under an existing run. At the storage layer, use `list.NewOptions(&model.Task{}, 2, "parent_task_id", nil)` and call `TaskStore.ListTasks`. Pass its returned token to `list.NewOptionsFromToken(token, 2)` and list again. The equivalent native API parameters are `order_by=parent_task_id`, `page_size=2`, and `page_token` for continuation. With UUIDs ending in `1011`, `1012`, and `1013`, page one returns `[1011, 1012]`; page two incorrectly returns `[1011, 1012]` again and still has a next-page token. Expected: page two returns `[1013]` and an empty token. The third task is otherwise never reached.

   **Root cause:** [native Task.ParentTaskUUID is a nullable `*string`](https://github.com/kubeflow/pipelines/blob/b93fd252cee648a93daa10c257e9f5c10a49cace/backend/src/apiserver/model/task.go#L147), and [GetFieldValue returns that pointer](https://github.com/kubeflow/pipelines/blob/b93fd252cee648a93daa10c257e9f5c10a49cace/backend/src/apiserver/model/task.go#L249). A typed nil pointer boxed in `interface{}` is not equal to nil, so [nextPageToken's nil check](https://github.com/kubeflow/pipelines/blob/b93fd252cee648a93daa10c257e9f5c10a49cace/backend/src/apiserver/list/list.go#L615) misses it. The value serializes as JSON `null`, but `SortByFieldIsNull` remains false. After decoding the token, `SortByFieldValue` is nil; neither the normal-value cursor branch nor the [NULL-cursor branch](https://github.com/kubeflow/pipelines/blob/b93fd252cee648a93daa10c257e9f5c10a49cace/backend/src/apiserver/list/list.go#L436) adds a continuation predicate, so the query repeats page one.

   **Fix/acceptance criteria:** represent legitimate nullable scalar cursors explicitly while preserving errors for unsupported fields. Preserve NULLS-last ordering, filters, existing metric-NULL behavior, and the UUID tie breaker. Tokens identify the first row of the next page: within the NULL segment, continuation must include that saved UUID (`>=` ascending / `<=` descending). Add storage regression coverage for all-parentless tasks and mixed parented/parentless tasks, ascending and descending, duplicate parent values, and transitions into the NULL segment. Traverse with a bounded page count until the token is empty, and assert every expected UUID appears exactly once. Also cover token serialization/deserialization and non-NULL parent values.

<details>
<summary>Minimal failing storage regression for item 3</summary>

This test can be added to `backend/src/apiserver/storage/task_store_test.go` at `b93fd252c`; it uses the existing helpers and imports in that file. `initializeTaskStore` seeds `run-1`. The deliberately omitted `ParentTaskUUID` produces real NULL values through `CreateTask`.

```go
func TestListTasks_ParentlessPagination(t *testing.T) {
	db, taskStore, _ := initializeTaskStore()
	defer db.Close()

	ids := []string{testUUID1, testUUID2, testUUID3}
	for i, id := range ids {
		taskStore.uuid = util.NewFakeUUIDGeneratorOrFatal(id, nil)
		_, err := taskStore.CreateTask(&model.Task{
			Namespace:        "ns1",
			RunUUID:          "run-1",
			Name:             fmt.Sprintf("parentless-%d", i),
			Pods:             createTaskPodsAsJSONSlice(createTaskPod("p1", "uid1", apiv2beta1.PipelineTask_EXECUTOR)),
			Fingerprint:      fmt.Sprintf("fp-%d", i),
			State:            1,
			StateHistory:     model.JSONSlice{},
			InputParameters:  model.JSONSlice{},
			OutputParameters: model.JSONSlice{},
			TypeAttrs:        model.JSONData{},
		})
		require.NoError(t, err)
	}

	opts, err := list.NewOptions(&model.Task{}, 2, "parent_task_id", nil)
	require.NoError(t, err)
	first, total, token, err := taskStore.ListTasks(&model.FilterContext{}, opts)
	require.NoError(t, err)
	require.Equal(t, 3, total)
	require.Len(t, first, 2)
	require.NotEmpty(t, token)

	opts, err = list.NewOptionsFromToken(token, 2)
	require.NoError(t, err)
	second, _, next, err := taskStore.ListTasks(&model.FilterContext{}, opts)
	require.NoError(t, err)

	var got []string
	for _, task := range append(first, second...) {
		got = append(got, task.UUID)
	}
	assert.Equal(t, ids, got)
	assert.Len(t, second, 1)
	assert.Empty(t, next)
}
```

Run the focused regression, then the affected suites:

```bash
go test ./backend/src/apiserver/storage -run '^TestListTasks_ParentlessPagination$' -count=1 -v
go test ./backend/src/apiserver/list ./backend/src/apiserver/model ./backend/src/apiserver/storage -count=1
```

</details>

Neilblaze:
Sure thing, on it!

Neilblaze:
@jeffspahr thanks for the detailed breakdown, all three are done.

1. This PR now targets `release-2.18`, still the same five files, and the description no longer claims `runId` works.
2. #14400 has the run getter fix for master. The tests are adapted to the native model, and the mapping guard now uses filled in fixtures plus a token round trip for scalar fields.
3. #14401 fixes paging through tasks with no parent, with storage coverage for the cases you listed.

Neilblaze:
@jeffspahr Could you please re-run the failed `MLflow` job and then `check_ci_status` when you get a chance? Thanks! :)

jeffspahr (approved):
Re-reviewed at `68602d496588376f94f01283f5fde79cb519ac67` against `release-2.18`. LGTM; no remaining code findings.

The run/task getter fixes and regression tests are appropriate for this MLMD-based release branch. The requested native-model follow-ups are separately scoped in #14400 and #14401.

Validation: `go test -count=1 ./backend/src/apiserver/list ./backend/src/apiserver/model ./backend/src/apiserver/storage` passed. Restoring only the original Run/Task getters makes the three new regression tests fail, confirming that they catch the defects fixed here.

The earlier PostgreSQL MLflow failure matches the existing MLMD concurrency issue #14353. Its rerun passed without a source change, and all 192 completed checks now pass, including the aggregate CI gate. The release-branch CI adjustment for unsupported PostgreSQL deployments is tracked separately in #14405.

jeffspahr:
Thanks for splitting these up @Neilblaze !

